PR #1588, opened 2019-12-19
Fix core IssueLinker for monorepo packages

rebelagentm:
For https://github.com/github/dependabot-api/issues/940


feelepxyz:
Looks like the issue is caused by this module so don't think we can use any of the tests/code from this PR: https://github.com/dependabot/dependabot-core/blob/master/common/lib/dependabot/pull_request_creator/message_builder/issue_linker.rb#L29-L44

Issue #352, opened 2018-04-18
Support sbt projects


albuch:
Hey, 
it would be awesome to support sbt dependencies and plugin dependencies as well.  

https://www.scala-sbt.org/1.x/docs/Library-Dependencies.html
https://www.scala-sbt.org/1.x/docs/Using-Plugins.html

Multi-projects are supported in sbt as well and is a common use case so that should be considered as well: https://www.scala-sbt.org/1.x/docs/Multi-Project.html

tyrcho:
I ran into this sbt plugin which should help a lot !
https://github.com/rtimush/sbt-updates

You can append `addSbtPlugin("com.timushev.sbt"    % "sbt-updates"         % "0.3.4")` to the end of `project/plugins.sbt` (creating it if not present) and then run `sbt dependencyUpdates` and parse the output in stdout. It looks like that:

```
[info] Found 10 dependency updates for notification-impl
[info]   com.datadoghq:dd-java-agent:dd-java-agent            : 0.10.0           -> 0.16.0
[info]   com.lightbend.lagom:lagom-logback                    : 1.4.6  -> 1.4.8           
[info]   com.lightbend.lagom:lagom-reloadable-server:dev-mode : 1.4.6  -> 1.4.8           
[info]   com.lightbend.lagom:lagom-scaladsl-dev-mode          : 1.4.6  -> 1.4.8           
[info]   com.lightbend.lagom:lagom-scaladsl-server            : 1.4.6  -> 1.4.8          
```

Note that in multi-project build you will have duplicated lines.

Even partial support for sbt in dependabot would be great (ie creating the PR to notify the library has been updated). It will probably be impossible to cover all ways in which versions are defined in sbt since you can code in your build files.

mark-dhl:
A project by @fthomas already does this. Hope you could maybe setup some sort of collaboration !

https://github.com/fthomas/scala-steward

anonymousRega:
Good from by Rega hawlery

frederikmars2653:
We are already using dependabot for python and ruby, but the biggest part of our codebase is scala.
Are there any updates on scala/sbt-support?

greysteil:
Not yet. We'd still love to add sbt support, but we're a small team and are currently focussed on scaling Dependabot so it can create automated security fixes for all GitHub repos.

ewolfe:
I was linked here from https://dependabot.com/java/ - I would like to help beta test Java Maven support

[...]

PR #1589, opened 2019-12-20
Initial attempt at support for sbt.

Grundlefleck:
Begins to address issue #352 

Versioning and update resolution is based heavily on the existing (beta) Gradle support.

This provides minimum support, it will:

- update build.sbt from root dir, and all *.scala and *.sbt files in the project/ dir
- update basic dependencies, of the form "org.someone" % "something" % "1.2.3"
- support updating scala version specific libs, iff the scalaVersion can be resolved,

     e.g. "org.someone" %% "something" % "1.2.3" will find
     org.someone:something_2.11:1.23 if val scalaVersion = "2.11" is set in the build.

It does not yet support:
- resolving versions from vals, e.g.
     val guavaVersion = "28.1-jre"
     ...
     libraryDependencies += "com.google.guava" % "guava" % guavaVersion
     because guavaVersion is not a string literal containing a version, it is ignored
- updating plugins from e.g. project/plugins.sbt, of the form:
     addSbtPlugin("com.typesafe.sbt" % "sbt-git" % "1.0.0")
- repositories other than maven central, which is hardcoded.
- GitHub dependencies
- dependabot's multi-dependency updates

I've manually tested with dry-run.sh against a half dozen or so scala repositories, and while a lot of updates are missed (for reasons described above) the support seems robust. I very much consider this PR a starting point for sbt support rather than the finished article, but hoped for some early feedback that I'm on the right path.


Grundlefleck:
The build_test_and_reorg build looks to have failed on unrelated infrastructure / networking issues. I can't trigger a repeat build without perms, but I have checked the initial failure has been fixed locally.

feelepxyz:
:wave: Thanks for opening this PR, back in 2019 😬 (sorry for the extreme delay!), we really appreciate the effort to make Dependabot better for the community.

Unfortunately, we are not accepting new ecosystems into core at the moment. We are currently focused on making some improvements into how we handle extensibility before we add any more ecosystems. We will still be improving and upgrading support for existing ecosystems.

Please check the [Contribution guidelines](https://github.com/dependabot/dependabot-core/blob/main/CONTRIBUTING.md) for more information.

Grundlefleck:
@feelepxyz thanks for letting me know. 

I am now far away from the Scala ecosystem, I'm unlikely to pick this up if and when you start accepting new ecosystems again. I doubt anything from this code would still be relevant so I'll close this PR.

PR #1599, opened 2019-12-26
Update config.yml

AnonaWolf:


.dependabot/config.yml
```diff
@@ -1,4 +1,4 @@
-version: 1
+version: 1
```
feelepxyz: Extra space shouldn't be needed

PR #1602, opened 2020-01-03
Docker: Add support for global ARG instructions

ffscl:
Evaluate any default values in global `ARG` instructions in a Dockerfile before parsing any `FROM` instructions.

See dependabot/feedback#813 for more details.


docker/spec/fixtures/docker/dockerfiles/multi_stage_global_arg
```diff
@@ -0,0 +1,26 @@
...
+
+# Install Ruby 2.4
+RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-keys C3173AA6 \
+    && echo "deb http://ppa.launchpad.net/brightbox/ruby-ng/ubuntu zesty main" > /etc/apt/sources.list.d/brightbox.list \
+    && apt-get update
+RUN apt-get install -y ruby2.4 ruby2.4-dev
+
+# This should have no impact on global ARG instructions used in FROM instructions
+ARG REGISTRY=docker.io
+FROM ${REGISTRY}/ubuntu:17.04
```
feelepxyz: Would docker use `docker.io` here? Think this will cause pain down the line if we diverge from Docker's behaviour. Could we rework this to use the last found ARG when matching a used arg in `FROM`?

MichaelScheetz-HPE:
lgtm. I would like to have this.

docker/spec/fixtures/docker/dockerfiles/multi_stage_global_arg
Fox32: Yes, `docker.io` is the default behavior in docker. On the other hand this is just a test example.

Fox32:
I would like to see this landed, is there anything I can do to help?

docker/spec/fixtures/docker/dockerfiles/multi_stage_global_arg
Fox32: Ok, I understand your comment the wrong way. No, docker would not use docker.io here, as there is a difference between global (before the first `FROM`) and local `ARG` instructions. The test is correct.

oswalya:
Hi there. Any timeline on finalizing / merging this? Looks good to me and we would really love to have this feature 👍

jeffwidman:
@ffscl should this be closed in favor of https://github.com/dependabot/dependabot-core/pull/4837? 

I haven't dug deeply into the implementation, so if there are things that this adds that are missing from that PR, please clarify.

Issue #1556, opened 2019-12-04
Poetry version support


ulgens:
Dependabot supports Poetry but it doesn't use given version in pyproject.toml file. When i use any version newer than 1.0.0a1, update check fails because of changed lock file syntax.

tommilligan:
Would also be interested in this. Just upgraded to the newly released poetry v1 and getting the error:

```log
updater | ERROR <job_17789943> Error processing tld (Dependabot::SharedHelpers::HelperSubprocessFailed)
updater | ERROR <job_17789943>                                   
updater | <job_17789943> [NonExistentKey]   
updater | <job_17789943> 'Key "hashes" does not exist.'  
updater | <job_17789943>                                   
updater | <job_17789943> update [--no-dev] [--dry-run] [--lock] [--] [<packages>]...
updater | <job_17789943> 
```

sobolevn:
I am all in into this change! 

Repo that has this problem: https://github.com/wemake-services/wemake-python-styleguide/blob/master/pyproject.toml

Poetry version is specified in the build file:

```
[build-system]
requires = ["poetry>=1.0"]
build-backend = "poetry.masonry.api"
```

But, it does not work. What website says:
<img width="670" alt="Снимок экрана 2019-12-18 в 11 27 18" src="https://user-images.githubusercontent.com/4660275/71069269-6bbce880-2189-11ea-9b06-268b693da59b.png">

Logs:

```
updater | ERROR <job_18040832> Error processing astboom (Dependabot::SharedHelpers::HelperSubprocessFailed)
updater | ERROR <job_18040832>                                   
updater | <job_18040832> [NonExistentKey]   
updater | <job_18040832> 'Key "hashes" does not exist.'  
updater | <job_18040832>                                   
updater | <job_18040832> update [--no-dev] [--dry-run] [--lock] [--] [<packages>]...
updater | <job_18040832> 
updater | ERROR <job_18040832> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-python-0.113.28/lib/dependabot/python/update_checker/poetry_version_resolver.rb:319:in `run_poetry_command'
updater | ERROR <job_18040832> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-python-0.113.28/lib/dependabot/python/update_checker/poetry_version_resolver.rb:85:in `block (2 levels) in fetch_latest_resolvable_version_string'
[...]

PR #1621, opened 2020-01-10
Support multiple poetry versions

tomzx:
Install poetry 0.12.17 for pre-1.0.0 lock files.
In the case that a poetry.lock file is discovered with a
metadata.hashes entry, we should use poetry 0.12.17 to update
the lockfile in order to preserve version compatibility. Otherwise
assume that the user wants to use the latest poetry version (1.0.0
at this time).

This attempts to tackle #1571. There are a few issues with this PR, but this is to get the discussion started. I'd like to add tests to confirm it works as expected, however I'm not a ruby/rspec expert so I'd need some help. I think that most of the tests in `python/spec/dependabot/python/update_checker/poetry_version_resolver_spec.rb` could be executed both under 0.12.7 and 1.0.0, which would require creating a separate set of fixtures for each version.

One thing I'd like to discuss is where the downgrading logic should go. In my tests, it seems it will do the downgrade when it checks the first package. I think however that it would make sense to execute the downgrade before that, however I'm unsure where that place is. I put the code there simply because it was the path where the crash was occurring and all the information necessary was there. There's also a bit of code about installing a version of python that may not have been installed yet and installing the requirements (poetry amongst them), so I thought it was not too bad.

### DIFF
--- a/python/helpers/requirements.txt
+++ b/python/helpers/requirements.txt
@@ -3,7 +3,7 @@ pip-tools==4.3.0
 hashin==0.14.6
 pipenv==2018.11.26
 pipfile==0.0.2
-poetry==0.12.17
+poetry==1.0.0
 
 # Some dependencies will only install if Cython is present
 Cython==0.29.14
--- a/python/lib/dependabot/python/update_checker/poetry_version_resolver.rb
+++ b/python/lib/dependabot/python/update_checker/poetry_version_resolver.rb
@@ -80,20 +80,42 @@ module Dependabot
                   )
                 end
 
+                update_poetry_binary_version
+
                 # Shell out to Poetry, which handles everything for us.
                 run_poetry_command(poetry_update_command)
 
-                updated_lockfile =
-                  if File.exist?("poetry.lock") then File.read("poetry.lock")
-                  else File.read("pyproject.lock")
-                  end
-                updated_lockfile = TomlRB.parse(updated_lockfile)
+                updated_lockfile = read_lockfile
 
                 fetch_version_from_parsed_lockfile(updated_lockfile)
               rescue SharedHelpers::HelperSubprocessFailed => e
                 handle_poetry_errors(e)
               end
             end
+        end
+
+        def update_poetry_binary_version
+          # TODO: I'm not sure if the case where there's no lockfile is
+          # already handled by dependabot.
+          lockfile = read_lockfile
+
+          # Before version 1.0.0, poetry used a metadata.hashes to store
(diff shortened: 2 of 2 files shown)


amureki:
Hey @tomzx !

Thanks for starting a work here. ✨ ✨ ✨ 
I switched from pipenv to Poetry (1.0+) and also struggling with dependabot now.

I noticed a comment from @sobolevn [here in #1556](https://github.com/dependabot/dependabot-core/issues/1556): 
>In case making a support for several versions is a hard thing, then I suggest to drop poetry@0.x support in favour of poetry@1.x support. Because update process is easy for developers.

And I am also wondering if support for _pre-release Poetry versions_ should be kept. 
This adds complexity of maintaining for a tiny value. Let me explain:
I do believe that Poetry was discovered and being used by developers who are actively maintaining their projects and got tired of pipenv/other tools. This group of people most likely switched to a stable 1.+ version already.
But even if they are not, instead of keeping compatibility, we should kindly suggest them to iterate and update versions (since those are users who are using _dependabot - tool to keep their stacks up-to-date, why packaging tooling should be an exception there?_)

I am happy to put some of my time in this and try to help here, though I am not yet familiar with Ruby stack. Maybe also with the help and opinion of @feelepxyz we can get this going!

sobolevn:
I sort of agree. There's no point in supporting `poetry<1.0` since for devs transition is just automatic. `poetry` itself converts old files to new ones.

tomzx:
Given that we might be upgrading someone's lockfile from a pre-1.0.0 format to a 1.0.0+ format, should there be some sort of warning as part of the PR to let them know or is it their responsibility to review the changes and notice that themselves?

I agree that supporting pre-1.0.0 and 1.x versions is probably not very useful for people using dependabot. I would thus suggest closing this and merging any of the automated PRs that bump to 1.0.x (1.0.3 was released today), which will take care of adding support for any 1.x versions, as well as upgrading pre-1.0.0 lock files.

sobolevn:
Yes, it makes sense.

jtbeach:
Is there anything I can do to help move this forward?

sobolevn:
@greysteil I would love to help with this one (but I don't know Ruby)! dependabot is not working for me for more than a month now.

greysteil:
I'm afraid I don't work on Dependabot anymore (I'm focussed on other security products at GitHub), but @feelepxyz and the team should be able to help!

feelepxyz:
Closing this out as we've been running poetry v1 for a good while now.

PR #1626, opened 2020-01-13
Support attributes in maven properties.

uri-canva:
Because we update the properties via string substitution instead of editing the dom with nokogiri we can fail to update valid POMs that don't don't match what we expect, for example POMs where properties have attributes.

This will result in no substitution being performed, which then triggers an assertion:

```
uri@MacBook-Pro-108 ~/g/d/dependabot-core> bin/dry-run.rb maven canva/canva --dir parent
=> fetching dependency files
=> dumping fetched dependency files: ./dry-run/canva/canva/parent
=> parsing dependency files
=> updating 192 dependencies
=== com.google.googlejavaformat:google-java-format (1.5)
 => checking for updates
 => latest version from registry is 1.7
 => latest resolvable version is 1.7
 => updating com.google.googlejavaformat:google-java-format from 1.5 to 1.7
	4: from bin/dry-run.rb:401:in `<main>'
	3: from bin/dry-run.rb:401:in `each'
	2: from bin/dry-run.rb:445:in `block in <main>'
	1: from bin/dry-run.rb:395:in `generate_dependency_files_for'
/Users/uri/github/dependabot/dependabot-core/maven/lib/dependabot/maven/file_updater.rb:34:in `updated_dependency_files': No files changed! (RuntimeError)
```

This PR adds support for properties with attributes, though it still uses string substitution, so it will still fail if the properties are not formatted as expected, for example if they're broken over multiple lines. To stay backwards compatible as much as possible it will still attempt to substitute using the regular expression first, and still fall back to not performing any substitution if neither the regular expression or the node matches.

### DIFF
--- a/maven/lib/dependabot/maven/file_updater/property_value_updater.rb
+++ b/maven/lib/dependabot/maven/file_updater/property_value_updater.rb
@@ -24,12 +24,22 @@ module Dependabot
           filename = declaration_details.fetch(:file)
 
           pom_to_update = dependency_files.find { |f| f.name == filename }
-          updated_content = pom_to_update.content.sub(
-            %r{<#{Regexp.quote(node.name)}>
-               \s*#{Regexp.quote(node.content)}\s*
-               </#{Regexp.quote(node.name)}>}xm,
-            "<#{node.name}>#{updated_value}</#{node.name}>"
-          )
+          property_re = %r{<#{Regexp.quote(node.name)}>
+            \s*#{Regexp.quote(node.content)}\s*
+            </#{Regexp.quote(node.name)}>}xm
+          property_text = node.to_s
+          if pom_to_update.content =~ property_re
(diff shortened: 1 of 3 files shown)


greysteil:
Thanks for the PR @uri-canva. I'm not a developer anymore and don't work on Dependabot, but @feelepxyz and the team should be able to help with review. The test looks 💯and makes your use here completely clear.

uri-canva:
GitHub suggested to add you as reviewer, sorry for the noise. 🙇

greysteil:
Absolutely no problem! :-)

SerCeMan:
Hi, @feelepxyz! Thank you for merging! Would you be able to say approximately when this change can be released https://dependabot.com/ so that we can plan when we can switch back to using property attributes?

feelepxyz:
@SerCeMan deploying the change now, should be live in the next 20 mins ✌️

PR #1631, opened 2020-01-17
Respect HTTP scheme for git dependencies

martijnversluis:
Even in these modern ages I do have this gem in a git repository that
does not support HTTPS (yet). Currently the uri_with_auth always adds
"https://", which makes the service pack URI invalid. This change makes
sure it uses `http` if and only if the original scheme is HTTP.


martijnversluis:
Thanks for accepting @feelepxyz! Not rushing anything, but do you by chance know within what timeframe a new version of `dependabot-omnibus` will be released containing this patch? 

Thanks! ✌️

feelepxyz:
@martijnversluis will get a new release out today in a few hours 👍

PR #1687, opened 2020-02-11
Fixes #1681 - Python repository URL's with environment variables are broken

### LINKED ISSUE #1681: Custom registry contining an environment variable appears to be broken
dependabot-omnibus version `0.116.6`

Hello,
I'm trying to use the [dependabot standalone script](https://github.com/dependabot/dependabot-script/blob/master/generic-update-script.rb) with a repository that contains a Pipfile with the following lines:

```toml
[[source]]
url = "https://${CREDENTIALS}@private.repo/repository/pypi/simple"
name = "some-name"
```

I've configured the credential like so:

```ruby
  {
    "type" => "python_index",
    "index-url" => "https://private.repo/repository/pypi/simple",
    "token" => ENV['CREDENTIALS'],
  }
```

**Note:** `CREDENTIALS` contains both a **username** and a **password**, i.e `user:pass`, not a token.

However this throws the following Ruby error:

```
x/dependabot/vendor/ruby/2.6.0/gems/dependabot-python-0.116.6/lib/dependabot/python/update_checker/index_finder.rb:183:in `clean_check_and_remove_environment_variables': 
The following source could not be reached as it requires authentication (and any provided details were invalid or lacked the required permissions): 
https://${CREDENTIALS}@private.repo/repository/pypi/simple/ (Dependabot::PrivateSourceAuthenticationFailure)
```

Now I've done some debugging and I cannot seem to see how this works at all, so I must be doing something really stupid:

1. [The clean_check_and_remove_environment_variables method](https://github.com/dependabot/dependabot-core/blob/224d29f57716d165c5624f208939362f8a691766/python/lib/dependabot/python/update_checker/index_finder.rb#L152) is called with the url `https://${CREDENTIALS}@private.repo/repository/pypi/simple/`
2. [The config_variable_urls](https://github.com/dependabot/dependabot-core/blob/224d29f57716d165c5624f208939362f8a691766/python/lib/dependabot/python/update_checker/index_finder.rb#L159-L165) variable contains the _correctly authenticated url_, i.e `https://username:password@private.repo/...`
3. The [regexp created](https://github.com/dependabot/dependabot-core/blob/224d29f57716d165c5624f208939362f8a691766/python/lib/dependabot/python/update_checker/index_finder.rb#L167-L171) strips out the environment variables, and so becomes a value like `https://private\.repo/repository/pypi/simple/`
4. This does not match any of the values in `config_variable_urls`
5. [`cleaned_url`](https://github.com/dependabot/dependabot-core/blob/224d29f57716d165c5624f208939362f8a691766/python/lib/dependabot/python/update_checker/index_finder.rb#L175) becomes a value `https://@private.repo/...`
6. [`credential_for`](https://github.com/dependabot/dependabot-core/blob/224d29f57716d165c5624f208939362f8a691766/python/lib/dependabot/python/update_checker/index_finder.rb#L189-L195) fails to match, as `https://@private.repo/` isn't in any defined python repository credential

Now if I set `index-url` like so: `https://@private.repo/repository/pypi/simple` it passes the authentication and makes a request, however the credential is **invalid** as it contains a percent-encoded `@` character - i.e the URL is created as `https://user:pass@@private.repo/...` with two `@` characters.

In total I've tried:
1. `https://@private.repo/repository/pypi/simple`
2. `https://private.repo/repository/pypi/simple`
3. `private.repo/repository/pypi/simple`
4. `@private.repo/repository/pypi/simple`
5. `https://USER:PASS@private.repo/repository/pypi/simple`
6. `https://${CREDENTIALS}@private.repo/repository/pypi/simple`

And nothing seems to work. Am I doing something completely wrong here?

orf:
Fixes #1681. Also I'm not a rubyist at all, so please let me know if I'm doing something stupid!

This method _nearly_ works, so I took the path of least resistance. `config_variable_urls` is a list of authenticated URLs, like `https://user:password@corp.com`, it's just that the current regular expression that's resolved doesn't match them at all.

So the change I made was to construct the regex differently. Rather than *stripping* out the entire `${env}@` structure I replace it with `.*@`, which will match any value. Technically this won't work if you define two separate repositories with the same path but with different credentials, but that's such a silly thing to do that I don't think we care about that.

All other environment variable interpolations are replaced with `.+`, which makes sense. I chose to use `.*@` instead of `.+@` for the credentials as it's _possible_ the credentials are empty, and it would be good if this worked in that case and selected the correct registry URL.

Here's how the regex is constructed:

```
> url = "https://${CREDS}@lol.${THING}.com"
> url_regex = url.sub(%r{(?<=://).+@}, "@")
                           .split(ENVIRONMENT_VARIABLE_REGEX). 
                           map { |part| Regexp.quote(part) }
                           .map { |part| part.sub("@", ".*@") }.join(".+")
https://.*@lol\..+\.com
```

This regex correctly matches against a fully authenticated URL:

```
> "https://user:pass@lol.my_place.com".match?(url_regex)
true
```
I'm not sure how to add tests for this, but I'll give it a go later.


python/spec/dependabot/python/update_checker/index_finder_spec.rb
```diff
@@ -309,5 +309,40 @@
         end
       end
     end
+
+    describe "#environment_variables" do
+      subject { finder.clean_check_and_remove_environment_variables(*arguments) }
```
orf: I've got no clue if this is a good idea or not, I copied it from here: https://zverok.github.io/blog/2017-11-01-rspec-method-call.html

python/spec/dependabot/python/update_checker/index_finder_spec.rb
```diff
@@ -309,5 +309,46 @@
         end
       end
     end
+
+    describe "#environment_variables" do
```
feelepxyz: The convention is to name these blocks to match the method being tested
```suggestion
    describe "#clean_check_and_remove_environment_variables" do
```

python/spec/dependabot/python/update_checker/index_finder_spec.rb
```diff
@@ -309,5 +309,46 @@
...
+      subject do
+        finder.clean_check_and_remove_environment_variables(*arguments)
```
feelepxyz:
```suggestion
        finder.clean_check_and_remove_environment_variables(url)
```

python/spec/dependabot/python/update_checker/index_finder_spec.rb
```diff
@@ -309,5 +309,46 @@
...
+              "type" => "python_index",
+              "index-url" => "https://${creds}@not.company.com/simple",
+              "token" => "user:password",
+              "replaces-base" => false
+            }
+          ]
+        end
+
+        it "selects the correct URL" do
+          let(:arguments) { "https://${creds}@company.com/simple" }
```
feelepxyz:
```suggestion
          let(:url) { "https://${creds}@company.com/simple" }
```

python/spec/dependabot/python/update_checker/index_finder_spec.rb
```diff
@@ -309,5 +309,46 @@
...
+
+          it do
+            is_expected.to eq "https://user:password@company.com/simple"
+          end
+        end
+
+        it "interpolates two variables" do
+          let(:arguments) { "https://${creds}@${domain}.company.com/simple" }
```
feelepxyz:
```suggestion
          let(:url) { "https://${creds}@${domain}.company.com/simple" }
```

python/spec/dependabot/python/update_checker/index_finder_spec.rb
```diff
@@ -309,5 +309,46 @@
...
+
+          it do
+            is_expected.to eq "https://user:password@not.company.com/simple"
+          end
+        end
```
feelepxyz: Thanks for adding these tests 🙌 Could you also add a couple that shouldn't match?

For example: `https://${creds}@other.com/simple` and `https://${creds}@other.company.com/simple`

orf: Sure thing!

python/spec/dependabot/python/update_checker/index_finder_spec.rb
```diff
@@ -309,5 +309,46 @@
         end
       end
     end
+
+    describe "#clean_check_and_remove_environment_variables" do
```
feelepxyz: Oh didn't spot that this is a private method. Could you move the `context "environment variable interpolation` block inside `describe "#index_urls` ?

I think you can then expect something like this instead

```ruby
context "environment variable interpolation" do
  let(:credentials) do
    [
      {
        "type" => "python_index",
        "index-url" => "https://${creds}@company.com/simple",
        "token" => "user:password",
        "replaces-base" => false
      },
      {
        "type" => "python_index",
        "index-url" => "https://${creds}@not.company.com/simple",
        "token" => "user:password",
        "replaces-base" => false
      }
    ]
  end

  it "cleans and removes the environment variables" do
    expect(index_urls).to match_array(
      [
        "https://user:password@company.com/simple",
        "https://user:password@not.company.com/simple"
      ]
    )
  end
end
```

orf: Thanks! I'll give it a go and see what failures I get. Thanks for your help with this 👍

python/lib/dependabot/python/update_checker/index_finder.rb
```diff
@@ -165,9 +165,10 @@ def clean_check_and_remove_environment_variables(url)
             map { |u| u.strip.gsub(%r{/*$}, "") + "/" }
 
           regexp = url.
-                   sub(%r{(?<=://).+@}, "").
+                   sub(%r{(?<=://).+@}, "@").
                    split(ENVIRONMENT_VARIABLE_REGEX).
                    map { |part| Regexp.quote(part) }.
+                   map { |part| part.sub("@", ".+") }.
                    join(".+")
           authed_url = config_variable_urls.find { |u| u.match?(regexp) }
```
SVilgelm: I tried to apply this patch but it doesn't help me with an issue is described here: #1772 
I added some debug prints and found that config_variable_urls is empty.

In my case I put the extra-index-url in the requirements.txt:
```pip
--extra-index-url https://${PYPI_USER}:${PYPI_PASSWORD}@<....>/artifactory/api/pypi/<....>/simple/
```
this patch gives a correct regexp, but it doesn't help because of empty config_variable_urls
@feelepxyz @orf Any thoughts or hints?

SVilgelm: Look like I found, I just added the index_url to credentials dict in dependabot-script. But I really thought it should be parsed automatically somehow

SVilgelm:
Any updates on this? I still need this patch to run script agains private pypi

dcaixinha:
Any updates on this one? Like @SVilgelm, I need this to run Dependabot against a private Pypi.

SVilgelm:
@hmarr @feelepxyz @greysteil @orf please

feelepxyz:
@SVilgelm sorry for dropping this! We've been heads down working on the new github integration. Will take a proper look next week. Want to understand the current behaviour a bit better as we have quite a complex auth setup involving multiple strategies.

orf:
I'm sorry - I likely will not have the time or energy to bring this MR to completion. @SVilgelm did you say that the current fix works for you?

SVilgelm:
Hi @orf, yes this fix works for me. It will be great If you can merge it

python/lib/dependabot/python/update_checker/index_finder.rb
hudelgado: @SVilgelm does the change i've added is needed to use this fix?
Can you tell me what do you had to change for it to work ok?
Thanks

SVilgelm:
@orf any news about this?

gnagel:
@orf / @SVilgelm - Is there any update on this one?  I'd love to see this merged in too.

jeffwidman:
Also I should add @orf thanks for the hard work on this. The team was very understaffed for a while and stuff slipped through the cracks. We are now staffed up a bit more, so although we can't debug every issue, if someone else is willing to take the time to debug and put together a PR we will do our best to be more responsive.

PR #1700, opened 2020-02-21
Ensures codecommit client actually creates a commit before opening a PR

userhas404d:
Currently, dependabot will happily open a PR against a codecommit repo that requires updates, but fails to actually include the updated file and associated commit in the PR.. this should fix that.

### DIFF
--- a/common/lib/dependabot/pull_request_creator/codecommit.rb
+++ b/common/lib/dependabot/pull_request_creator/codecommit.rb
@@ -50,6 +50,8 @@ module Dependabot
         branch = create_or_get_branch(base_commit)
         return unless branch
 
+        create_commit
+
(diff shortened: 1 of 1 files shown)


common/lib/dependabot/pull_request_creator/codecommit.rb
```diff
@@ -49,6 +49,7 @@ def base_commit_is_up_to_date?
       def create_pull_request
         branch = create_or_get_branch(base_commit)
         return unless branch
+        create_commit
```
feelepxyz: Looks like rubocop is complaining about this

```suggestion

        create_commit
```

userhas404d: fixed!

PR #1714, opened 2020-03-02
Fix Semver Build Identifier Support

a1flecke:
The version comparison logic was broken in the Maven Version.rb file.  The `PREFIXED_TOKEN_HIERARCHY` did not list `+` as a valid key.  The Maven documentation site listed did not provide any guidance on how to handle `+` so I chose some sane values for `qualifier` and `number`.
I added some tests to test comparison with plus signs.  Please let me know if additional testing is needed.


a1flecke:
@feelepxyz Is anything else needed for this PR?  I did fix the Rubocop issues that are currently on master.

PR #1715, opened 2020-03-03
Adds reverse_markdown to the codecommit client

userhas404d:
Unfortunately, codecommit doesn't support html rendering in PR details, making the resulting PR detail block difficult to read. This PR ensures that codecommit dependabot PR details are provided as markdown. 

without reverse_markdown: 
without_reverse_markdown

with reverse_markdown: 
with_reverse_markdown


feelepxyz:
@userhas404d actually, we'll change the message builder to not output html for codecommit. Didn't realise we broke this by introducing a change only intended for github links and mentions.

PR #1735, opened 2020-03-14
Update Python version to 3.8.2

ulgens:


jurre:
Thanks @ulgens, there are a few tests that also reference the version, I'm going to fix those up and then merge your changes 👍

PR #1786, opened 2020-04-07
Enable updates against codecommit hosted repositories

userhas404d:
Title says it all. 

Hoping this approach isn't too much of a hack.. 

I had to add `awscli` and update the `git` config of the base docker image in order to enable `git` operations against codecommit repositories. I tried to avoid this approach and query the cc api directly, but unfortunately it doesn't look like the `/info/refs?service=git-upload-pack` endpoint is exposed in a way where it is directly accessible.  Thankfully `git ls-remote <cc_repo_url>` returns the `git-upload-pack` without issue.

Obligatory I'm not a ruby dev, but I started to work on updated tests for this. Long story short, I'm struggling with mocking out the results of `Open3.capture3()`.  

It _feels_ like it should be simple[(ish)](https://stackoverflow.com/a/50152544/12031185) but given the amount of time I've wasted on it I'm thinking there's either a simpler approach to testing cc specifics here that I'm just not seeing, or what I've added needs a refactor. Any guidance on how to approach writing updated tests would be most appreciated.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -138,6 +138,14 @@ ENV RUSTUP_HOME=/opt/rust \
     PATH="${PATH}:/opt/rust/bin"
 RUN export CARGO_HOME=/opt/rust ; curl https://sh.rustup.rs -sSf | sh -s -- -y
 
+### CODECOMMIT
+
+# install the awscli and configure git to use the codecommit credential-helper
+RUN apt-get update \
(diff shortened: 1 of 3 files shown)


userhas404d:
Issue was addressed in https://github.com/dependabot/dependabot-core/pull/3934

PR #1790, opened 2020-04-09
Handle unauthorized pushes to protected branches

feelepxyz:
When you restrict who can push to protected branches a different error
message is returned compared to pushing to a branch that only disables
force-pushes.


feelepxyz:
Merging as `common` specs have passed

PR #1803, opened 2020-04-17
Nuget: Handle version requirements with prefix

feelepxyz:
Changed `gsub` to `sub`. I don't think it makes sense to `gsub` as there
shouldn't be multiple of the same version in the requirement.

Fixes https://github.com/dependabot/feedback/issues/906

### DIFF
--- a/nuget/lib/dependabot/nuget/update_checker/requirements_updater.rb
+++ b/nuget/lib/dependabot/nuget/update_checker/requirements_updater.rb
@@ -12,8 +12,6 @@ module Dependabot
   module Nuget
     class UpdateChecker
       class RequirementsUpdater
-        VERSION_REGEX = /[0-9a-zA-Z]+(?:\.[a-zA-Z0-9\-]+)*/.freeze
-
         def initialize(requirements:, latest_version:, source_details:)
           @requirements = requirements
           @source_details = source_details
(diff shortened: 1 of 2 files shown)


nuget/lib/dependabot/nuget/update_checker/requirements_updater.rb
```diff
@@ -12,7 +12,7 @@ module Dependabot
   module Nuget
     class UpdateChecker
       class RequirementsUpdater
-        VERSION_REGEX = /[0-9a-zA-Z]+(?:\.[a-zA-Z0-9\-]+)*/.freeze
+        VERSION_REGEX = /#{Nuget::Version::VERSION_PATTERN}/.freeze
```
jurre, line 15: Can we reference `Version::VERSION_PATTERN` directly in `initialize` or do you expect us to change it here separately from the one in `Version`?

feelepxyz: Ah yes you mean just reference where it's used? 👍

jurre: Yeah, bit of a nit, but we're not changing anything about the pattern right?

AraHaan: I do not think so. However not sure as I do not know much on dependabot. Only that if it works or not.

This pull request is to fix the issue where the nuget.config points to feeds that are used in the .net core projects themselves and uses packages in preview or beta, etc that dependabot manages to somehow screw up on. Hopefully this gets merged soon so that way the analyzer I use gets updated again realtime.

feelepxyz: Moved it inline, the only change is making it into a regex pattern whereas the `Version` pattern is a string.

PR #1450, opened 2019-10-18
Add support for Bitbucket Server

gvassallo:
Hi there. I've added support for Bitbucket Server. I've been using this code for quite a while now, and it's been working flawlessly. 

Please let me know if anything is unclear/missing. I'd be happy to help to provide support for Bitbucket Server :)

PR #1818, opened 2020-04-27
Add PR Creator for Bitbucket

tooppaaa:
What I Did: 

Use what has been done for Azure by replacing "Azure" with "Bitbucket"

Use API's to 
- Get Branch by name
- Get all PR's
- Create a commit 
- Create PR
- Get all commits

Please note that I'm not a Ruby developer, any feedbacks are welcome !

### DIFF
--- a/common/lib/dependabot/clients/bitbucket.rb
+++ b/common/lib/dependabot/clients/bitbucket.rb
@@ -15,7 +15,6 @@ module Dependabot
       #######################
 
       def self.for_source(source:, credentials:)
-       
         credential = credentials.
                      find do |cred|
           cred["type"] == "git_source" &&
@@ -129,15 +128,17 @@ module Dependabot
       def pull_requests(source_branch, target_branch)
         path = "#{source.repo}/pullrequests?"
         path += "state=OPEN&state=MERGED&state=SUPERSEDED&state=DECLINED"
-        path += "&q=source.branch.name = \"" + source_branch + "\" AND destination.branch.name = \"" + target_branch + "\""
+        path += "&q=source.branch.name = \"" + 
+          source_branch + "\" AND destination.branch.name = \"" + 
+          target_branch + "\""
 
         response = get(base_url + path)
 
         JSON.parse(response.body).fetch("values")
       end
 
-      def create_commit(branch_name, base_commit, commit_message, files,
-        author_details)
+      def create_commit(branch_name, _base_commit, commit_message, files,
+      author_details)
         path = "#{source.repo}/src"
 
         body = files.map { |file| [file.path, file.content] }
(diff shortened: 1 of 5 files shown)


tooppaaa:
@jurre I struggle with the formatting due to my ruby knowledge and my setup, any chance we can chat in order to see what's next for this PR ?

jarbey:
+1

infin8x:
Hi @tooppaaa!

Thank you for filing this pull request. Unfortunately, the Dependabot team are currently tied up with some time-sensitive initiatives and we're not able to give this request the attention it needs at the moment. We're hoping to circle back to this and other PRs sometime over the summer.

tooppaaa:
Awesome thanks
We'll keep the fork for now and get our changes back here from time to time !

common/lib/dependabot/pull_request_creator/pr_name_prefixer.rb
```diff
@@ -421,6 +451,14 @@ def codecommit_client_for_source
           )
       end
 
+      def bitbucket_client_for_source
+        @bitbucket_client_for_source ||=
+          Dependabot::Clients::BitbucketTS.for_source(
```
MMartyn, line 456: This looks like a typo maybe? Not seeing where BitbucketTS is defined.

tooppaaa: Indeed, BitbucketTS is the name I used to fork Bitbucket in the script we're using. Nice catch !

common/lib/dependabot/clients/bitbucket.rb
```diff
@@ -10,12 +10,27 @@ class NotFound < StandardError; end
...
+        new(source, credential)
+      end
+
       ##########
       # Client #
       ##########
 
-      def initialize(credentials:)
-        @credentials = credentials
+      def initialize(source, credential)
```
renatolond, line 31: This seems to break calls for existing methods (for instance the file fetcher in the generic script).
I think to keep compatibility, it should be something more like
```
def initialize(credentials:, source: nil)
```

jurre:
Hi, apologies on the lack of a response on this. I believe this is a duplicate of https://github.com/dependabot/dependabot-core/pull/1450. Let me paste my reply on that PR, because it applies here as well:

> Hi all, I appreciate the work that went into this PR and this definitely warrants a reply from the maintainers.
> 
> However, at this point I don't think that we can accept this patch into dependabot-core natively, as we don't have the capacity or expertise to provide support for Bitbucket. This is currently already a problem with the existing Azure DevOps and GitLab clients, where we're not positioned well to provide support and rely on other members in the community to do so. However, not having clear maintainers for those parts of the codebase makes this hard.
> 
> Where I would like to see support for (non GitHub) source control providers go in the future, is for them to be stand-alone gems that you can pull in next to dependabot-core. dependabot-core would provide a clear interface for those gems to implement. These gems can then be maintained by people who actually know and use these providers.
> 
> I think this should currently already be possible, but it might rely on monkey-patching some of the existing dependabot-core classes, which is not ideal. I'd be more than happy to accept patches that improve the interface for external providers if you decide to go down that route.

I appreciate the work that went into this, and if anyone wants to tackle releasing this as a standalone gem and there is any support I can provide to make it easier to integrate with dependabot-core I'm happy to help.

PR #1824, opened 2020-04-30
Readme

jsoref:
* Split `highlights` out
* Mention GitHub Actions
* Add some commas
* minor whitespace changes to make the above easier.


hmarr:
This looks like an automated pull request that makes the README harder to read in its raw form.

jsoref:
It might look like an automated pull request, but let me assure you that it isn't.

If I were writing automated things, I'd at least be consistent in writing my commit messages.

jsoref:
I'm happy to split the changes or drop the line splitting bit entirely... the reason for it is that it means that further changes to the list don't result in someone else taking blame for the entire line.

I'm a real human and my main goal is to improve content. Give me direction and I'll try to follow...

PR #1830, opened 2020-05-01
Sanitize Dir.chdir calls in gemspecs

haines:
Gems generated by `bundle gem` dynamically generate their `spec.files` by shelling out to `git ls-files`, inside a `Dir.chdir` block ([see here](
https://github.com/rubygems/rubygems/blob/9cfc95d9d9eeede71884a25ade8851b7ab722011/bundler/lib/bundler/templates/newgem/newgem.gemspec.tt#L25-L27)).

Dependabot was unable to update our dependencies because of this call.

```
There was an error while loading `our-private-project.gemspec`: No such file or directory @ dir_chdir
```

Currently, gemspec sanitization replaces dynamic `files` assignment with an empty array by searching for an AST node that looks like this:

```
s(:send,
  s(:lvar, :spec), :files=,
  s(:send, ...))
```

However, because `Dir.chdir` receives a [literal block](https://github.com/whitequark/parser/blob/68c21e247fde370fae7640fdc3bb6323b03fa8b7/doc/AST_FORMAT.md#passing-a-literal-block), the offending AST node looks like this instead:

```
s(:send,
  s(:lvar, :spec), :files=,
  s(:block,
    s(:send, ...),
    ...))
```

This PR adds a check to sanitize these types of dynamic `files` assignments as well.

### DIFF
--- a/bundler/lib/dependabot/bundler/file_updater/gemspec_sanitizer.rb
+++ b/bundler/lib/dependabot/bundler/file_updater/gemspec_sanitizer.rb
@@ -159,7 +159,15 @@ module Dependabot
             return false unless node.children.first&.type == :lvar
             return false unless node.children[1] == :files=
 
-            node.children[2]&.type == :send
+            node_dynamically_lists_files?(node.children[2])
+          end
+
+          def node_dynamically_lists_files?(node)
+            return false unless node.is_a?(Parser::AST::Node)
+
+            return true if node.type == :send
+
+            node.type == :block && node.children.first&.type == :send
           end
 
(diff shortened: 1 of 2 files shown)


haines:
The failing spec seems to be unrelated, it's broken on master too (the gem it's updating has a [new release](https://github.com/gocardless/business/releases/tag/v1.18.0) a couple of days ago).

jurre:
Thanks @haines! The change looks good, the offending error has since been fixed so if you rebase/merge master in your branch this should go green 🎉

Issue #1231, opened 2019-06-26
Support setting default NPM repo when no npmrc is present


keirlawson:
It appears that dependabot-core keeps the public NPM registry as the default in the case that there is no npmrc present in the repository in question, even when additional repos are specified via credentials.  This is causing dependabot-core to fail with a 404 when attempting to detect upgrades for internal libraries in our projects, which make the assumption that the default registry has been set to our internal one by the user/build environment.  I note that for python registries there is an option `replaces-base` to allow for setting the default registry, is there an equivalent for NPM registries? If not would you accept a PR for this functionality?

greysteil:
This sounds like something we should be able to fix without requiring configuration. Can you @-mention Dependabot in a PR or issue that it's creating? That will help me replicate and fix.

keirlawson:
Unfortunately I can't link to a PR as this was using dependabot-core on our internal Github Enterprise installation, any other way I can help investigate?

greysteil:
Ah, OK.

My first hunch here is that adding a `.npmrc` file or `.yarnrc` file that declared a global registry (with or without credentials) would stop this behaviour, as it's likely coming from the `RegistryFinder#global_registry` method. Do you have one committed?

keirlawson:
yes, adding such a `.npmrc` does indeed resolve the issue, the problem with this is I suspect we have a lot of repos that don't have this file.  We could of course do an exercise to add it to each of them, but I was wondering if it would be possible to set the default when invoking dependabot to avoid needing to do that.

greysteil:
Hmmm, Dependabot shouldn't be falling back to registry.npmjs.org anywhere else unless it can't find details of the dependency anywhere else. What line are you seeing the error from?

greysteil:
(I presume you're using the latest version of Dependabot Core, by the way?)

keirlawson:
I've been using the dependabot-core docker image, pulled 7 days ago and invoked via dependabot-script with our NPM repo added to the credentials block. The error is:

```
  - Updating debug (from )…Traceback (most recent call last):
	23: from ./generic-update-script.rb:126:in `<main>'
	22: from ./generic-update-script.rb:126:in `each'
	21: from ./generic-update-script.rb:164:in `block in <main>'
	20: from /workspace/vendor/ruby/2.6.0/gems/dependabot-npm_and_yarn-0.108.15/lib/dependabot/npm_and_yarn/file_updater.rb:39:in `updated_dependency_files'
[...]

PR #1847, opened 2020-05-08
#1231: Create .npmrc file in the current working directory

debo:
Many moons later, after quite some investigation and unfortunately after @keirlawson left the team, we managed to find out what the issue was and add a fix.

Basically what is/was happening is that in the absence of a `.npmrc` file (our default case in most repositories) dependabot creates a temporary one at runtime but it always creates it in the root folder of the project rather than in the working directory where the `package.json` is located and where the `npm:update` command is run which, in some cases, like in our setup, could be in subfolders.

I hope it helps.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
@@ -365,7 +365,9 @@ module Dependabot
         def write_temporary_dependency_files(lockfile_name,
                                              update_package_json: true)
           write_lockfiles(lockfile_name)
-          File.write(".npmrc", npmrc_content)
+
+          dir = Pathname.new(lockfile_name).dirname
+          File.write(File.join(dir, ".npmrc"), npmrc_content)
 
           package_files.each do |file|
             path = file.name


jurre:
Thanks so much for looking into this @debo 🎉 the explanation and changes make sense to me, would be ace if we could add a test to prevent this from happening again in the future

debo:
@jurre I didn't look too much into it because whilst fixing this the original tests where failing so I thought they were already good enough, but I'll double check and make sure we are not missing some weird edge case.

jurre:
For reference, this fixes #1231

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -365,11 +365,12 @@ def dependencies_in_error_message?(error_message)
         def write_temporary_dependency_files(lockfile_name,
                                              update_package_json: true)
           write_lockfiles(lockfile_name)
-          File.write(".npmrc", npmrc_content)
```
feelepxyz, line 368: It should also work if we write this into the directory where we write the lockfile instead of into each package json directory (there might be hundreds of package json files in a monorepo and we only run `npm install` in the directory where we write this lockfile). Something like:

```
lockfile_dir = Pathname.new(lockfile_name).dirname
npmrc_path = lockfile_dir.join(".npmrc")
File.write(npmrc_path, npmrc_content)
```

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -365,11 +365,12 @@ def dependencies_in_error_message?(error_message)
...
 
           package_files.each do |file|
             path = file.name
-            FileUtils.mkdir_p(Pathname.new(path).dirname)
+            working_directory = Pathname.new(path).dirname
+            FileUtils.mkdir_p(working_directory)
+            File.write("#{working_directory}/.npmrc", npmrc_content)
```
feelepxyz, line 373: Should be able to ✂️ this if the above suggestion works
```suggestion
```

debo: @feelepxyz I think I tried your suggestion as my first iteration but I seem to recall it didn't quite work, but I might be wrong, I'll give it another shot because I do also prefer to avoid writing files all the time within the loop.

feelepxyz:
@debo nice work on hunting this down, would be ace to add some tests for this in the file_updater_spec.rb. I had a look at adding these but ran into some weirdness around how the npmrc builder gets the registry from the lockfile. I think it might not work properly when there are nested lockfiles as it always reads the root lockfile and gets the dependency urls from that to check which global registry should be written to `npmrc`.

Could you share how you've structured your project which is failing? E.g. do you have a root package.json and npm lockfile along with a nested package.json and npm lockfile. Does the nested lockfile have a custom registry set on all dependencies or just some? What is the `resolved` URL in each lockfile? What registry URL are you providing in `credentials`? Feel free to substitute the custom URL with some dummy value, would be super helpful to get an idea how to reproduce your setup and create a test case.

debo:
@feelepxyz I'll try my best as soon as I have some additional time, maybe over the weekend or so.

Our project is structured something like the following:

```shell
./foobarbaz/client/package-lock.json
./foobarbaz/client/package.json
./foobarbaz/package-lock.json
./foobarbaz/package.json
./lerna.json
./package-lock.json
./package.json
./packages/client/package-lock.json
./packages/client/package.json
./packages/server/package-lock.json
./packages/server/package.json
```
Some of the packages are pulled from a private artifactory registry and the code is stored in our own instance of GitHub Enterprise.

I hope that helps.

debo:
@feelepxyz done, it worked like a charm. However, not being a ruby developer I genuinely have no idea how to start with the tests mostly because that specific file is not tested at all. Any pointer is welcome.

debo:
Hey @feelepxyz any news on this? I'd like to contribute further but I need some pointer 😄

feelepxyz:
@debo thanks for this! Sorry for the slow response.

debo:
Hey @feelepxyz don't be silly, it was a pleasure and I appreciate you people are busy during these times. I just wished I could provide also the tests as requested but I didn't know where to start from, next time maybe.

PR #1850, opened 2020-05-08
fix: Fix version types in gradle to allow matching postfixed version types

bfulco:
Fixes a bug in Gradle that would not match version types when additional context is provided.

For example: com.thoughtworks.xtream:xstream:1.4.11-java7

Now, the system will obey the version type that the user provides, maintaining the correct version types for future upgrades.


jurre:
Thanks @bfulco !

PR #1852, opened 2020-05-08
go modules: remove unnecessary requirements

hmarr:
Fixes an issue in our Go modules support where we'd fail to remove lines from the go.mod that because unnecessary due to a Dependabot version update.

Reviewer notes:
- To understand the problem this solves, it may be beneficial to start with the first commit, which adds a failing test.
- Reorganizes a bit of code (class constants, attr_reader), which creates some visual noise in the diff. Sorry about that.

Previously, we'd:
1. Add the new dependency versions to the go.mod
2. If a go.sum was present, run `go get` to update the go.sum

Sometimes the new dependency requirement should change other entries in the go.mod. For instance, say we use dependency x@v1, which depends on y@v1. We upgrade y to v2 manually in the go.mod, which promotes y@v2 to the go.mod. Dependabot kicks in and then updates x to v2, which includes a dependency on y@v2, meaning the go.mod entry for y@v2 is no longer necessary (according to MVS).

With the previous algorithm, we wouldn't update the go.mod correspondingly.

Now, we:
1. Add the new dependency versions to the go.mod
2. Run `go get` to produce a regenerated go.sum and go.mod
3. Figure out all the differences in `require`s between the new go.mod and the input go.mod
4. Update the original go.mod to reflect all these changes

This change also introduces an imperative `update_files` method, which walks through this process, and hopefully is easier to follow than the prior implementation.

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
@@ -2,7 +2,6 @@
 
 require "dependabot/shared_helpers"
 require "dependabot/errors"
-require "dependabot/go_modules/file_parser"
 require "dependabot/go_modules/file_updater"
 require "dependabot/go_modules/native_helpers"
 
--- a/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
@@ -25,8 +25,6 @@ module Dependabot
           /go: ([^@\s]+)(?:@[^\s]+)?: .* declares its path as: ([\S]*)/m
         ].freeze
 
-        attr_reader :dependencies, :go_mod, :go_sum, :credentials
-
         def initialize(dependencies:, go_mod:, go_sum:, credentials:)
           @dependencies = dependencies
           @go_mod = go_mod
@@ -43,6 +41,8 @@ module Dependabot
         end
 
         private
+
+        attr_reader :dependencies, :go_mod, :go_sum, :credentials
 
         def updated_files
           @updated_files ||= update_files
--- a/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
@@ -125,7 +125,8 @@ module Dependabot
           _, stderr, status = Open3.capture3(ENVIRONMENT, "go get -d")
           handle_subprocess_error(stderr) unless status.success?
 
-          { go_mod: File.read("go.mod"), go_sum: File.read("go.sum") }
+          updated_go_sum = go_sum ? File.read("go.sum") : nil
+          { go_mod: File.read("go.mod"), go_sum: updated_go_sum }
         end
 
         def parse_manifest_requirements(go_mod_content)
--- a/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
@@ -49,6 +49,7 @@ module Dependabot
           @updated_files ||= update_files
         end
 
+        # rubocop:disable Metrics/AbcSize
         def update_files
           substitutions = replace_directive_substitutions(go_mod.content)
           clean_go_mod = substitute_all(go_mod.content, substitutions)
@@ -86,6 +87,7 @@ module Dependabot
 
           { go_mod: output_go_mod, go_sum: regenerated_files[:go_sum] }
         end
+        # rubocop:enable Metrics/AbcSize
 
         def update_go_mod(go_mod_content, dependencies)
           File.write("go.mod", go_mod_content)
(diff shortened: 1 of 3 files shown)


go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -120,14 +202,15 @@ def local_replacements
...
-            body.sub(path, stub_path)
+        def substitute_all(file, substitutions)
+          substitutions.reduce(file) do |text, (a, b)|
+            text.sub(a, b)
           end
         end
 
-        def handle_subprocess_error(path, stderr)
+        def handle_subprocess_error(stderr)
+          stderr = stderr.gsub(Dir.getwd, "")
```
hmarr, line 212: The `path` argument was always intended to be the cwd, and was provided so we can strip it from the error message. It's simpler for the caller if we just look up the cwd directly.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -17,96 +36,159 @@ def initialize(dependencies:, go_mod:, go_sum:, credentials:)
...
-          /go: ([^@\s]+)(?:@[^\s]+)?: .* has non-.* module path "(.*)" at/,
-          /go: ([^@\s]+)(?:@[^\s]+)?: .* unexpected module path "(.*)"/,
-          /go: ([^@\s]+)(?:@[^\s]+)?: .* declares its path as: ([\S]*)/m
-        ].freeze
+        def updated_files
+          @updated_files ||= update_files
+        end
+
+        # rubocop:disable Metrics/AbcSize
+        def update_files
```
hmarr, line 53: While it's a relatively sizeable method, I think it makes the full update process clear, so I'd prefer to leave it as-is rather than subdiving it.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -2,13 +2,32 @@
...
 require "dependabot/go_modules/file_updater"
 require "dependabot/go_modules/native_helpers"
 
 module Dependabot
   module GoModules
     class FileUpdater
       class GoModUpdater
+        # Turn off the module proxy for now, as it's causing issues with
+        # private git dependencies
+        ENVIRONMENT = { "GOPRIVATE" => "*" }.freeze
```
hmarr, line 15: This appeared in a few methods below, so I've extracted it to a constant to reduce the duplication.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -2,13 +2,32 @@
...
+
+        RESOLVABILITY_ERROR_REGEXES = [
```
hmarr, line 17: These constants were below the `private` marker, which was confusing as constants are public in Ruby even if they follow the `private` marker.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -2,13 +2,32 @@
...
+          /build .*: cannot find module providing package/.freeze
+        ].freeze
+
+        MODULE_PATH_MISMATCH_REGEXES = [
+          /go: ([^@\s]+)(?:@[^\s]+)?: .* has non-.* module path "(.*)" at/,
+          /go: ([^@\s]+)(?:@[^\s]+)?: .* unexpected module path "(.*)"/,
+          /go: ([^@\s]+)(?:@[^\s]+)?: .* declares its path as: ([\S]*)/m
+        ].freeze
+
+        attr_reader :dependencies, :go_mod, :go_sum, :credentials
```
hmarr, line 29: This was at the bottom of the class. I've moved it here as it typically lives at the top of the class declaration.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -17,96 +36,159 @@ def initialize(dependencies:, go_mod:, go_sum:, credentials:)
         end
 
         def updated_go_mod_content
```
hmarr, line 38: It turned out that updating the `go.mod` and the `go.sum` required some common bits and pieces (such as running `go get`), so I've merged the implementations, and the original methods just reference the results.

go_modules/spec/dependabot/go_modules/file_updater/go_mod_updater_spec.rb
```diff
@@ -34,7 +34,7 @@
       requirements: requirements,
       previous_version: dependency_previous_version,
       previous_requirements: previous_requirements,
-      package_manager: "dep"
```
hmarr, line 37: This was a bad copy+paste in the tests that I caught while working on this.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jurre: Previously it was private and now it's public, if we don't need to access it from outside I'd typically declare it right after the `private` declaration, but not super bothered.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -17,96 +35,169 @@ def initialize(dependencies:, go_mod:, go_sum:, credentials:)
...
+
+          # Replace full paths with path hashes in the go.mod
+          clean_go_mod = substitute_all(go_mod.content, substitutions)
+
+          # Set the new dependency versions in the go.mod
+          updated_go_mod = in_temp_dir(stub_dirs) do
+            update_go_mod(clean_go_mod, dependencies)
+          end
+
+          # Then run `go get` to pick up other changes to the file cuased by
```
jurre, line 65:
```suggestion
          # Then run `go get` to pick up other changes to the file caused by
```

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jurre: I can follow along with the code as is so 👍 (the comments help)

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
hmarr: Good catch - fixed!

hmarr:
Thanks for the review @jurre 💞

PR #1863, opened 2020-05-18
Gradle: Add support for properties set as defaults

mcwarman:
Extract properties set as defaults, supports both the findProperty and hasProperty syntax styles:

```
findPropertyVersion = project.findProperty('findPropertyVersion') ?: '1.0.0'
hasPropertyVersion = project.hasProperty('hasPropertyVersion') ? project.getProperty('hasPropertyVersion') :'1.0.0'
```

### DIFF
--- a/gradle/lib/dependabot/gradle/file_parser/property_value_finder.rb
+++ b/gradle/lib/dependabot/gradle/file_parser/property_value_finder.rb
@@ -6,16 +6,19 @@ module Dependabot
   module Gradle
     class FileParser
       class PropertyValueFinder
+        # rubocop:disable Layout/LineLength
+        PROPERTY_DECLARATION_AS_DEFAULTS_REGEX =
+          /(?:\s*(?:project\.findProperty\(\s*['"][^\s]+['"]\s*\)\s*\?|project\.hasProperty\(\s*['"][^\s]+['"]\s*\)\s*\?\s*project\.getProperty\(\s*['"][^\s]+['"]\s*\)\s*):)?/.freeze
+
         SINGLE_PROPERTY_DECLARATION_REGEX =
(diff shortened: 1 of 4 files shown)


mcwarman:
If you do accept it, i'd consider squashing the commit as a lot of the changes are me battling with a CI of a language that i'm not familiar with.

gradle/lib/dependabot/gradle/file_parser/property_value_finder.rb
```diff
@@ -6,16 +6,19 @@ module Dependabot
   module Gradle
     class FileParser
       class PropertyValueFinder
+        # rubocop:disable Layout/LineLength
+        PROPERTY_DECLARATION_AS_DEFAULTS_REGEX =
+          /(?:\s*(?:project\.findProperty\(\s*['"][^\s]+['"]\s*\)\s*\?|project\.hasProperty\(\s*['"][^\s]+['"]\s*\)\s*\?\s*project\.getProperty\(\s*['"][^\s]+['"]\s*\)\s*):)?/.freeze
```
jurre, line 11: Is there any way we can break this up into smaller chunks? Maybe we could split out the `findProperty`, `hasProperty` and `getProperty` parts into separate variables and concatenate those?

The sheer length of this regex makes it very hard to understand what's going on, for me at least.

jurre:
Overall this looks like a great improvement @mcwarman, thanks! I'm just a bit worried about the complexity of the regexes, would we awesome if we can figure out a way to make those easier to understand.

mcwarman:
@jurre I've broken up the regex, hopefully its better.

In hindsight it should always looked like this.

jurre:
Thanks @mcwarman, really appreciate this fix 🎉

PR #1864, opened 2020-05-18
Handle protected branches enforcing linear history

feelepxyz:
You can enforce linear history for a protected branch preventing any
merge commits to be pushed to the branch, this currently prevents the
pull request from being updated. I'm not entirely sure where we're
creating a merge commit but might be because we're creating a new commit
and force pushing and this is the first rule that gets tripped up in
github.

Screen Shot 2020-05-18 at 15 03 45

`BranchProtected` is rescued in dependabot api resulting in a comment on
the pull request explaining why the update failed.

### DIFF
--- a/common/lib/dependabot/pull_request_updater/github.rb
+++ b/common/lib/dependabot/pull_request_updater/github.rb
@@ -163,7 +163,8 @@ module Dependabot
         return nil if e.message.match?(/Reference cannot be updated/i)
 
         if e.message.match?(/force\-push to a protected/i) ||
-           e.message.match?(/not authorized to push/i)
+           e.message.match?(/not authorized to push/i) ||
+           e.message.match?(/must not contain merge commits/)
           raise BranchProtected
         end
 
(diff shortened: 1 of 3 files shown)


common/lib/dependabot/pull_request_updater/github.rb
```diff
@@ -163,7 +163,8 @@ def update_branch(commit)
         return nil if e.message.match?(/Reference cannot be updated/i)
 
         if e.message.match?(/force\-push to a protected/i) ||
-           e.message.match?(/not authorized to push/i)
+           e.message.match?(/not authorized to push/i) ||
+           e.message.match?(/must not contain merge commits/)
```
brrygrdn, line 167: Since this case is not well understood, it makes sense to treat it as a protected branch error, but I wonder if we should instrument this specific outcome.

I guess the question is if this is the result of Dependabot, or a user modifying a Dependabot branch somehow?

feelepxyz: I think this will be a user change enforcing this branch protection rule on all branches, e.g. using `*`

feelepxyz: Could raise another error and instrument in the backend but inclined to leave it for now as it seems to be so niche

brrygrdn: Good call!

PR #1872, opened 2020-05-20
Always use exact dependencies label if one exists

feelepxyz:
Fixes https://github.com/dependabot/feedback/issues/846

This change makes sure an existing label named `dependencies` is always
used as the default over one that matches `*dependenc.`.

When a repo label exists that matches `*dependenc.` Dependabot will pick
and use that as the default label (unless you specifcy default/custom
labels).

### DIFF
--- a/common/lib/dependabot/pull_request_creator/labeler.rb
+++ b/common/lib/dependabot/pull_request_creator/labeler.rb
@@ -6,6 +6,8 @@ module Dependabot
   class PullRequestCreator
     class Labeler
       DEPENDENCIES_LABEL_REGEX = %r{^[^/]*dependenc[^/]+$}i.freeze
+      DEFAULT_DEPENDENCIES_LABEL = "dependencies"
+      DEFAULT_SECURITY_LABEL = "security"
 
       @package_manager_labels = {}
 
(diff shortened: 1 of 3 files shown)


common/lib/dependabot/pull_request_creator/labeler.rb
```diff
@@ -6,6 +6,8 @@ module Dependabot
   class PullRequestCreator
     class Labeler
       DEPENDENCIES_LABEL_REGEX = %r{^[^/]*dependenc[^/]+$}i.freeze
+      DEFAULT_DEPENDENCIES_LABEL = "dependencies"
+      DEFAULT_SECURITY_LABEL = "security"
```
brrygrdn, line 10: 👍🏻 Thanks for extracting these to constants

common/spec/fixtures/github/labels_with_custom_and_default.json
```diff
@@ -0,0 +1,58 @@
...
+      "id": 618335045,
+      "url": "https://api.github.com/repos/dependabot/dependabot-core/labels/bug",
+      "name": "bug",
+      "color": "ee0701",
+      "default": true
+  },
+  {
+      "id": 675690998,
+      "url": "https://api.github.com/repos/dependabot/dependabot-core/labels/dependencies",
+      "name": "Dependency: Gems",
```
brrygrdn, line 12: Should we add a lowercase `dependenci*` test fixture as well? I guess my concern is you could reverse the OR statement in `default_dependencies_label` and this test would pass if the match became case insensitive.

feelepxyz: Yeah good shout 👌

PR #1873, opened 2020-05-20
go_modules: multiline `go` output matching

thepwagner:
This PR updates the regular expressions used to parse `go` command output to handle errors encountered processing transitive dependencies.

When go encounters an error resolving a direct dependency, it outputs an error like:
```
go: github.com/hmarr/404@v1.4.0: reading github.com/hmarr/404/go.mod at revision v1.4.0: unknown revision v1.4.0
```

When go resolves a direct dependency, then fails resolving a transitive dependency-of-a-dependency, it outputs an error like:
```
go: github.com/dependabot/vgotest@v1.0.0 requires
	github.com/hmarr/404@v1.0.0: reading github.com/hmarr/404/go.mod at revision v1.0.0: unknown revision v1.0.0
```

That `\n\t` breaks dependabot's regexp, which is used to detect and retry on ephemeral errors (e.g. a 502 from a remote).
This results in a `DependencyFileNotParseable` being raised, which is incorrect if a valid `go.mod` was parsed with a problematic remote.

As observed https://github.com/dependabot/feedback/issues/938

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_parser.rb
+++ b/go_modules/lib/dependabot/go_modules/file_parser.rb
@@ -135,16 +135,16 @@ module Dependabot
         end
       end
 
-      GIT_ERROR_REGEX = /go: .*: git fetch .*: exit status 128/.freeze
+      GIT_ERROR_REGEX = /go: .*: git fetch .*: exit status 128/m.freeze
       def handle_parser_error(path, stderr)
         case stderr
-        when /go: .*: unknown revision/
+        when /go: .*: unknown revision/m
           line = stderr.lines.grep(/unknown revision/).first
           raise Dependabot::DependencyFileNotResolvable, line.strip
-        when /go: .*: unrecognized import path/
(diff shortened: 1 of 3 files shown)


go_modules/lib/dependabot/go_modules/file_parser.rb
```diff
@@ -135,16 +135,16 @@ def go_mod_content
...
-        when /go: .*: unknown revision/
+        when /go: .*: unknown revision/m
           line = stderr.lines.grep(/unknown revision/).first
           raise Dependabot::DependencyFileNotResolvable, line.strip
-        when /go: .*: unrecognized import path/
+        when /go: .*: unrecognized import path/m
           line = stderr.lines.grep(/unrecognized import/).first
           raise Dependabot::DependencyFileNotResolvable, line.strip
-        when /go: errors parsing go.mod/
+        when /go: errors parsing go.mod/m
```
jurre, line 147: TIL about `/m` multiline mode 🎉

PR #1882, opened 2020-05-27
Update tests to account for business 2.0 update

jurre:
We rely on the `business` gem for these tests, which recently got a [2.0
upgrade], which breaks these tests.

This is the simplest thing we can do to fix those tests, but I would
recommend that we swap out these public gems for dummy / cloned versions
that we can pin to a specific version.

[2.0 upgrade]: https://github.com/gocardless/business/blob/master/CHANGELOG.md#200---may-4-2020

### DIFF
--- a/bundler/spec/dependabot/bundler/file_updater/gemfile_updater_spec.rb
+++ b/bundler/spec/dependabot/bundler/file_updater/gemfile_updater_spec.rb
@@ -218,11 +218,12 @@ RSpec.describe Dependabot::Bundler::FileUpdater::GemfileUpdater do
     context "with a gem that has a git source" do
       let(:gemfile_fixture_name) { "git_source_with_version" }
       let(:lockfile_fixture_name) { "git_source_with_version.lock" }
+      let(:dependency_name) { "dependabot-test-ruby-package" }
       let(:dependency) do
         Dependabot::Dependency.new(
-          name: "business",
-          version: "c170ea081c121c00ed6fe8764e3557e731454b9d",
-          previous_version: "c5bf1bd47935504072ac0eba1006cf4d67af6a7a",
+          name: "dependabot-test-ruby-package",
+          version: "1c6331732c41e4557a16dacb82534f1d1c831848",
+          previous_version: "81073f9462f228c6894e3e384d0718def310d99f",
           requirements: requirements,
           previous_requirements: previous_requirements,
           package_manager: "bundler"
@@ -231,11 +232,12 @@ RSpec.describe Dependabot::Bundler::FileUpdater::GemfileUpdater do
       let(:requirements) do
         [{
           file: "Gemfile",
-          requirement: "~> 1.14.0",
+          requirement: "~> 1.1.0",
           groups: [],
           source: {
             type: "git",
-            url: "http://github.com/gocardless/business"
+            url: "http://github.com/dependabot-fixtures/"\
+            "dependabot-test-ruby-package"
           }
         }]
       end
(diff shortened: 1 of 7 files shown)


feelepxyz:
@jurre ah yeah should we swap this out for our own gem? Seems worth fixing these issues properly when they crop up, been trying to do it in the past but was lazy in this case 🙈

We've got this org that we've been using for test deps: https://github.com/orgs/dependabot-fixtures (just invited you as owner too)
And these rubygems:
- https://rubygems.org/gems/dummy-pkg-a 
- https://rubygems.org/gems/dummy-pkg-b

bundler/spec/dependabot/bundler/file_updater/gemfile_updater_spec.rb
```diff
@@ -231,11 +231,12 @@
       let(:requirements) do
```
feelepxyz, line 231: I think the `let(:dependency)` needs to be updated to `dependabot-test-ruby-package`? `version` possibly also changed to `1.1.0` and `previous_version` to `1.0.0`

bundler/spec/dependabot/bundler/file_updater/gemfile_updater_spec.rb
```diff
@@ -246,12 +247,17 @@
...
+            url: "http://github.com/dependabot-fixtures/"\
+            "dependabot-test-ruby-package"
           }
         }]
       end
 
-      it { is_expected.to include "\"business\", \"~> 1.14.0\", git" }
+      it do
+        is_expected.to include(
+          "\"dependabot-test-ruby-package\", \"~> 1.0.0\", git"
```
feelepxyz, line 258: Not entirely sure but looks like this spec was expecting an update matching the new requirements
```suggestion
          "\"dependabot-test-ruby-package\", \"~> 1.1.0\", git"
```

feelepxyz: Or maybe `1.0.1` https://github.com/dependabot-fixtures/dependabot-test-ruby-package/releases/tag/v1.0.1

bundler/spec/dependabot/bundler/file_updater/gemfile_updater_spec.rb
feelepxyz: oh I think git sources need the commit or tag as the version

bundler/spec/dependabot/bundler/file_updater/gemfile_updater_spec.rb
```diff
@@ -299,7 +305,11 @@
           }]
         end
 
-        it { is_expected.to include "\"business\", \"~> 1.8.0\"\n" }
+        it do
+          is_expected.to include(
+            "\"dependabot-test-ruby-package\", \"~> 1.0.0\", git"
```
feelepxyz, line 310: Hm looks like this test has changed, looks like it was testing whether the git source was removed from the lockfile if the new requirement was a published version where the previous_requirements was a git source. Might need to publish this package with a newer version to test it?

jurre: I messed up the versions and missed that all these tests were nested inside the same block, it should be fixed now

bundler/spec/dependabot/bundler/file_updater/gemfile_updater_spec.rb
```diff
@@ -282,9 +292,9 @@
       context "that should be removed" do
         let(:dependency) do
           Dependabot::Dependency.new(
-            name: "business",
-            version: "1.8.0",
-            previous_version: "c5bf1bd47935504072ac0eba1006cf4d67af6a7a",
+            name: "dependabot-test-ruby-package",
+            version: "1.0.0",
```
feelepxyz, line 296: Nit, should this be `1.1.0` to match the test expectation? Not sure if it's actually used but more for consistency, think `requirements` is the one used to update to the new version

bundler/spec/dependabot/bundler/file_updater/gemfile_updater_spec.rb
jurre: Ah ^ is a typo in the tagname, my bad

jurre:
Thanks, yeah that would be great. The deep nesting and heavy use of `let`s overwriting each other makes these quite hard to iterate on

PR #1891, opened 2020-06-02
Add error message to DependencyFileNotParseable

feelepxyz:
This should make these errors a lot easier to debug as we currently only show the file path that wasn't parseable.

We've done this in a few places already, e.g. go modules.

The error is presented in the backend. If the error has a `message` it's presented in the issue as `The error Dependabot encountered was: error message as code block`

## TODO 

- [ ] Add some tests 🤠
- [ ] Figure out if we will start leaking anything on public repos, e.g. errors containing config variables

### DIFF
--- a/bundler/lib/dependabot/bundler/file_fetcher.rb
+++ b/bundler/lib/dependabot/bundler/file_fetcher.rb
@@ -177,8 +177,10 @@ module Dependabot
             PathGemspecFinder.new(gemfile: file).path_gemspec_paths
           end.uniq
         end
-      rescue ::Bundler::LockfileError
-        raise Dependabot::DependencyFileNotParseable, lockfile.path
+      rescue ::Bundler::LockfileError => e
+        msg = e.message.strip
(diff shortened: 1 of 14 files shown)


bundler/lib/dependabot/bundler/file_fetcher.rb
```diff
@@ -177,8 +177,10 @@ def fetch_path_gemspec_paths
             PathGemspecFinder.new(gemfile: file).path_gemspec_paths
           end.uniq
         end
-      rescue ::Bundler::LockfileError
-        raise Dependabot::DependencyFileNotParseable, lockfile.path
+      rescue ::Bundler::LockfileError => e
+        msg = e.message.strip
```
jurre, line 181: Should we add that `strip` to the initializer of `DependencyFileNotParseable` instead so we don't have to duplicate it all over the place?

feelepxyz: Yeah good shout, we could maybe do it for all custom dependabot error messages.

PR #1894, opened 2020-06-03
Fix for GitHub Actions multiple sources

feelepxyz:
Fixes https://github.com/github/dsp-dependabot/issues/402

## TODO

- [x] Add review context and hints
- [x] Document changes
- [x] Check test cases where we go from a tag ref to no tag ref with a numeric version
- [x] Check test cases where we go from a previous version to a new tag ref instead of numeric version

## ☣️ Scary changes ahead

- 1. Changing a core assumption around dependency requirement sources
  - Previously, a **git** dependency was only assumed to have one unique requirement source and we bailed out early if this was not the cause because it becomes ambiguous to know what the current ref is
    - Previously, the *previous* and *new ref* for the dependency was found from the first requirement source, hence the above bailout
  - Now, we're instead checking if a dependency is a git dependency by checking if all requirement sources are git sources
    - Similarly, instead of getting the tag from the first requirement source, get the first one if there is **only one** unique one otherwise return `nil`
    - `nil` tags are already handled as this can already be the case
- 2. Changing the PR message builder to not require a previous version, where it will drop the `from x ` portion if there is no previous version
  - Previously this would either be left blank `from  to x` or raise when finding the previous requirement
  - When updating from multiple previous refs to a single new one we can't reliably figure out the previous version so should just drop `from x` but still say we're going `to v2.2.0`.

Handles the case when the same dependency has multiple sources:

```yml
on: [push]

name: Integration
jobs:
  chore:
    steps:
    - uses: actions/checkout@v2.1.0

  acceptance:
    steps:
    - uses: actions/checkout@v1.4.0
```

In the above case, we update both steps to the last version `v2.2.0` so the new tag ref is `v2.2.0`, but because we have two different previous requirement refs the previous tag will be nil.

This seems to be gracefully handled in all the places I've checked, e.g.
- Branch names only use the new ref if it exists
- The commit URL checker will only include the new tag if there is no previous tag, which includes all changes in that new tag
- The changelog finder and pruner will include all changes up to the new tag because we don't know the previous version


common/lib/dependabot/metadata_finders/base/changelog_finder.rb
```diff
@@ -319,19 +319,22 @@ def new_version
...
             r.dig(:source, "ref") || r.dig(:source, :ref)
-          end.compact.first
+          end.compact.uniq
+          return new_refs.first if new_refs.count == 1
         end
 
         def ref_changed?
-          previous_ref && new_ref && previous_ref != new_ref
+          # We could go from multiple previous refs (nil) to a single new ref
+          previous_ref != new_ref
```
feelepxyz, line 337: I've made this change in all places where we have the same logic but different places have slightly nuanced differences around which one is actually expected to exist so I've moved the presence check out of this method and made it local to wherever it's being used. e.g. `if new_ref && ref_changed? then new_ref else other`

common/lib/dependabot/metadata_finders/base/changelog_pruner.rb
```diff
@@ -139,19 +139,17 @@ def new_version
...
         end
 
         def new_ref
-          dependency.requirements.map do |r|
+          new_refs = dependency.requirements.map do |r|
             r.dig(:source, "ref") || r.dig(:source, :ref)
-          end.compact.first
-        end
-
-        def ref_changed?
```
feelepxyz, line 153: Unused

common/lib/dependabot/metadata_finders/base/commits_finder.rb
```diff
@@ -55,7 +55,7 @@ def new_tag
             return new_version
           end
 
-          return new_ref if git_source?(dependency.requirements) && ref_changed?
+          return new_ref if new_ref && ref_changed?
```
feelepxyz, line 58: `git_source?` is checked in `new_ref`

common/lib/dependabot/metadata_finders/base/release_finder.rb
```diff
@@ -252,36 +252,38 @@ def previous_version
           end
 
           if dependency.previous_version.match?(/^[0-9a-f]{40}$/) &&
-             ref_changed?
+             ref_changed? && previous_ref
```
feelepxyz, line 255: We could have gone to `nil` `previous_ref` here so we should return the `previous_version` instead

common/lib/dependabot/metadata_finders/base/release_finder.rb
```diff
@@ -252,36 +252,38 @@ def previous_version
...
             previous_ref
           else
             dependency.previous_version
           end
         end
 
         def new_version
-          if dependency.version.match?(/^[0-9a-f]{40}$/) && ref_changed?
+          if dependency.version.match?(/^[0-9a-f]{40}$/) &&
+              ref_changed? && new_ref
```
feelepxyz, line 264: Not sure this actually possible when the version is a git sha but seems like we could go from a `previous_ref` to `nil` `new_ref` here

common/lib/dependabot/pull_request_creator/message_builder.rb
```diff
@@ -64,7 +64,7 @@ def library_pr_name
         pr_name +
           if dependencies.count == 1
             "#{dependencies.first.display_name} requirement "\
-            "from #{old_library_requirement(dependencies.first)} "\
+            "#{from_version_msg(old_library_requirement(dependencies.first))}"\
```
feelepxyz, line 67: This is a pretty big change, we could try and figure out the previous version by taking the lowest previous ref if possible to compare (e.g. v1 and v2) otherwise falling back to the first one if there are multiple

common/spec/dependabot/metadata_finders/base/changelog_finder_spec.rb
```diff
@@ -71,12 +128,10 @@
 
       before do
         stub_request(:get, github_url).
-          with(headers: { "Authorization" => "token token" }).
```
feelepxyz, line 74: Messes with the `vcr` setup using `github_credentials`

common/spec/dependabot/pull_request_creator/branch_namer_spec.rb
```diff
@@ -444,5 +446,67 @@
...
+          source: {
+            type: "git",
+            url: "https://github.com/actions/checkout",
+            ref: "master",
+            branch: nil
+          }
+        }]
+      end
+
+      it "includes the new ref" do
```
feelepxyz, line 505:
```suggestion
      it "only includes the new ref" do
```

github_actions/spec/dependabot/github_actions/file_updater_spec.rb
```diff
@@ -148,6 +148,71 @@
...
+                ref: "v2.2.0",
+                branch: nil
+              }
+            }]
+          )
+        end
+
+        it "updates both sources" do
+          expect(subject.content).to include "actions/checkout@v2.2.0\n"
+          expect(subject.content).not_to include "actions/checkout@master\n"
```
feelepxyz, line 213:
```suggestion
          expect(subject.content).not_to include "actions/checkout@master\n"
          expect(subject.content).not_to include "actions/checkout@v2.1.0\n"
```

common/lib/dependabot/metadata_finders/base/changelog_finder.rb
```diff
@@ -343,10 +348,8 @@ def git_source?
           requirements = dependency.requirements
           sources = requirements.map { |r| r.fetch(:source) }.uniq.compact
           return false if sources.empty?
-          raise "Multiple sources! #{sources.join(', ')}" if sources.count > 1
 
-          source_type = sources.first[:type] || sources.first.fetch("type")
-          source_type == "git"
+          sources.all? { |s| s[:type] == "git" || s["type"] == "git" }
```
feelepxyz, line 352: Can we have multiple different source types for the same dependency, e.g. npm

common/spec/dependabot/pull_request_creator/message_builder_spec.rb
```diff
@@ -447,6 +497,16 @@ def commits_details(base:, head:)
           it { is_expected.to start_with("⬆️🔒 Bump business") }
         end
       end
+
+      context "with multiple git source requirements", :vcr do
+        include_context "with multiple git sources"
+
+        it do
+          is_expected.to eq(
+            "Update actions/checkout requirement to v2.2.0"
```
feelepxyz, line 506:
```suggestion
            "Update actions/checkout requirements to v2.2.0"
```

common/lib/dependabot/metadata_finders/base/changelog_finder.rb
```diff
@@ -314,24 +314,29 @@ def fetch_gitlab_file_list
         end
 
         def new_version
-          @new_version ||= git_source? ? new_ref : dependency.version
-          @new_version&.gsub(/^v/, "")
+          return @new_version if defined?(@new_version)
+
+          new_version = git_source? && new_ref ? new_ref : dependency.version
+          @new_version = new_version&.gsub(/^v/, "")
```
feelepxyz, line 320: Looks like the previous cached call wasn't doing `@new_version&.gsub(/^v/, "")` in any subsequent method calls

PR #1895, opened 2020-06-04
Handle cargo native dependencies

jurre:
Previously, when specifying a native dependency using `links =
"dependency"`, we would fail as we do not copy over the build.rs file,
which is needed to actually link to the binary. However, to run `cargo
update` we don't actually need to link the dependency, we just need
cargo not to fail. So adding a dummy build.rs file, like we do for
src/main.rs fixes the issue.

Fixes: https://github.com/dependabot/feedback/issues/943

### DIFF
--- a/cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
+++ b/cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
@@ -337,6 +337,8 @@ module Dependabot
 
             next if virtual_manifest?(file)
 
+            File.write(File.join(dir, "build.rs"), dummy_app_content)
+
             FileUtils.mkdir_p(File.join(dir, "src"))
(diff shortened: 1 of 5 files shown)


cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
```diff
@@ -372,6 +374,8 @@ def sanitized_manifest_content(content)
             object["package"].delete("default-run")
           end
 
+          # object["package"].delete("links") if object.dig("package", "links")
```
feelepxyz, line 377: ✂️ 
```suggestion
```

jurre:
Thnx for the reviews! As an aside, I notice there's a bunch of duplication between the two edited files here, making a mental note to look into if it makes sense to extract some of that, or if it's accidental duplication.

PR #1915, opened 2020-06-15
Fix bundler specs

feelepxyz:
This seems to fix the broken bundler tests on master, as seen here for example: https://circleci.com/gh/dependabot/dependabot-core/14271?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Fixed by updating the rubygems version index response and all `info/*` requests to rubygems. The update to some of the manifests using `dummy-pkg-*` was an initial attempt to fix this but didn't fix things. Kept these changes as using our own test gems should be more stable.

I attempted to switch to VCR for all the responses but I couldn't get it to record any cassettes, I think it's because Bundler uses a persistent `Net::HTTP` connection. Recording a plain `Net:HTTP` was working fine.

The other thing I spotted was that a bunch of rubygems package info requests where missing but not reported by VCR, e.g. https://index.rubygems.org/info/json 

My current theory is that when we see errors like this:

```
     Dependabot::SharedHelpers::ChildProcessFailed:
       Child process raised Bundler::VersionConflict with message: Bundler could not find compatible versions for gem "activesupport":
         In Gemfile:
           activesupport (= 5.2.0)
     
           rails_multitenant was resolved to 0.10.0, which depends on
             activerecord (>= 4.1, < 5.3) was resolved to 4.1.2.rc1, which depends on
               activesupport (= 4.1.2.rc1)
     
       Bundler could not find compatible versions for gem "arel":
         In Gemfile:
           rails_multitenant was resolved to 0.10.0, which depends on
             activerecord (>= 4.1, < 5.3) was resolved to 5.2.0, which depends on
               arel (>= 9.0)
     
       Could not find gem 'arel (>= 9.0)', which is required by gem 'activerecord (>= 4.1, < 5.3)', in any of the sources.
```

Specifically `Could not find gem 'arel (>= 9.0)', which is required by gem 'activerecord (>= 4.1, < 5.3)', in any of the sources.` it's because we haven't stubbed out the `arel` info request and this is either blocked in webmock but not picked up by VCR or bundler gets confused because the response it out of sync with what's in the cached versions index response. I have no idea why these tests just started failing on master 🤷‍♂️


bundler/spec/dependabot/bundler/file_updater_spec.rb
```diff
@@ -363,7 +363,7 @@
           end
 
           it "locks the updated gem to the latest version" do
-            expect(file.content).to include("business (1.13.0)")
+            expect(file.content).to include("business (1.18.0)")
```
feelepxyz, line 366: A bunch of the expected versions have changed after updating the rubygems index response stub and package info responses to match.

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
```diff
@@ -402,9 +402,9 @@
         }]
       end
 
-      it "doesn't just fall back to latest_version" do
+      it "unlocks the latest version" do
```
feelepxyz, line 405: Don't really understand the wording on the existing test, from my understanding and digging into to the version resolver, we'll try to unlock the latest version regardless of what's in the manifest currently because we pass `unlock_requirement = true`

PR #1918, opened 2020-06-18
Ensure we don't mention usernames with parens around them

jurre:
It seems like people are getting notifications from README's such as:
https://github.com/gong-boy/SSDIProject/pull/11

Which is generated from a CHANGELOG like:
https://github.com/angular/angular.js/blob/master/CHANGELOG.md#180-nested-vaccination-2020-06-01

Previously we would generate something like (<a href="https://github.com/feelepxyz">@feelepxyz</a>),
for which I would not expect a notification, but to be absolutely sure
we may as well wrap these is a code block.

### DIFF
--- a/common/spec/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer_spec.rb
+++ b/common/spec/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer_spec.rb
@@ -83,6 +83,15 @@ RSpec.describe namespace::LinkAndMentionSanitizer do
         it { is_expected.to eq("<p>Great work <code>@greysteil</code>!</p>\n") }
       end
 
+      context "that appears in parens" do
+        let(:text) { "by Esben Sparre Andreasen (@esbena)" }
+        it do
+          is_expected.to eq(
+            "<p>by Esben Sparre Andreasen (<code>@esbena</code>)</p>\n"
+          )
+        end
+      end
+
       context "with unmatched single code ticks previously" do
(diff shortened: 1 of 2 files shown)


common/spec/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer_spec.rb
```diff
@@ -83,6 +83,15 @@
         it { is_expected.to eq("<p>Great work <code>@greysteil</code>!</p>\n") }
       end
 
+      context "that appears in parens" do
+        let(:text) { "by Esben Sparre Andreasen (@esbena)" }
+        it do
+          is_expected.to eq(
+            "<p>by Esben Sparre Andreasen (<code>@esbena</code>)</p>\n"
```
feelepxyz, line 90: I think the current sanitizer will turn this into a link like `(<a href="https://github.com/esbena">@esbena</a>)`?

jurre: Correct, it seems like people are getting mentions from that. I'm going to test that a bit further and if it's the case, get rid of the linking entirely

jurre: Hmm, I don' t seem to get mentioned when I try this on a test account, I'm confused as to what's happening, but this change would actually make it worse as it won't wrap in codeblocks

PR #1922, opened 2020-06-19
Safely output markdown from link_and_mention_sanitizer

jurre:
When configured as UNSAFE, CommonMarker will leave in any HTML tags
present in the text, which can cause our markdown to break when rendered
by GitHub.

This change ensures that no unsafe attributes are included in the
output, but things like `<code>` blocks for usernames etc still continue
to work.

### DIFF
--- a/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
@@ -17,9 +17,8 @@ module Dependabot
         MENTION_REGEX = %r{(?<![A-Za-z0-9`~])@#{GITHUB_USERNAME}/?}.freeze
         # End of string
         EOS_REGEX = /\z/.freeze
-        # We rely on GitHub to do the HTML sanitization
         COMMONMARKER_OPTIONS = %i(
-          DEFAULT GITHUB_PRE_LANG FULL_INFO_STRING
+          GITHUB_PRE_LANG FULL_INFO_STRING
         ).freeze
         COMMONMARKER_EXTENSIONS = %i(
           table tasklist strikethrough autolink tagfilter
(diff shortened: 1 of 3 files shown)


feelepxyz:
Ouch lots of specs are failing because HTML 😖

common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
```diff
@@ -19,7 +19,7 @@ class LinkAndMentionSanitizer
         EOS_REGEX = /\z/.freeze
         # We rely on GitHub to do the HTML sanitization
         COMMONMARKER_OPTIONS = %i(
-          UNSAFE GITHUB_PRE_LANG FULL_INFO_STRING
+          DEFAULT GITHUB_PRE_LANG FULL_INFO_STRING
```
feelepxyz, line 22: Oh wait will this break all previous sections that have been turned into HTML bu this step?

feelepxyz: Yeah looks like we re-parse all the already sanitized markdown that's now HTML here https://github.com/dependabot/dependabot-core/blob/master/common/lib/dependabot/pull_request_creator/message_builder.rb#L326 - maybe ok if we get rid of that step? Not sure it's adding anything

jurre: Ah yes you're right, I think it should be safe to get rid of that step

feelepxyz: Looks like we sanitize all cascades except for `maintainer_changes_cascade`, might work as expected if we move `sanitize_links_and_mentions` from `metadata_cascades_for_dep` to `maintainer_changes_cascade`

jurre: I opted to do a final "unsafe" pass of the sanitization, to ensure we don't mention people by accident.

jurre: Also passed `maintainer_changes_cascade` through the sanitization to be sure, nice spot, ty!

jurre:
Yeah I inspected a few tests, and it seems to be mostly linebreaks etc that are added by the last `sanitize_links_and_mentions`.

PR #1924, opened 2020-06-19
    Bug Fixes: Maven Classifier, SemVer

a1flecke:
* [Add support for maven dependency classifiers](https://maven.apache.org/pom.html#dependencies)

```
    classifier:
    The classifier distinguishes artifacts that were built from the same POM but differ in content. It is some optional and arbitrary string that - if present - is appended to the artifact name just after the version number.

    As a motivation for this element, consider for example a project that offers an artifact targeting Java 11 but at the same time also an artifact that still supports Java 1.8. The first artifact could be equipped with the classifier jdk11 and the second one with jdk8 such that clients can choose which one to use.

    Another common use case for classifiers is to attach secondary artifacts to the project's main artifact. If you browse the Maven central repository, you will notice that the classifiers sources and javadoc are used to deploy the project source code and API docs along with the packaged class files.
```
   
 * [Build Identifiers should not be used in SemVer comparisions](https://semver.org/#spec-item-11)
      * Dependabot was not ignoring the build identifier when determing if a dependency update was a semver update


aaiezza:
This would be a great addition to dependabot! Thanks for making this @a1flecke!

a1flecke:
@feelepxyz All checks pass. My employer is encountering both of these bugs which is lowering the utility of dependabot for us.  Please let me know if there is something that I need to do for this PR.

maven/lib/dependabot/maven.rb
```diff
@@ -22,7 +22,7 @@
   register_display_name_builder(
     "maven",
     lambda { |name|
-      artifact_id = name.split(":").last
+      _, artifact_id, = name.split(":")
```
feelepxyz, line 25: I think something like this would be a bit clearer and shouldn't make rubocop complain
```suggestion
      _group_id, artifact_id, _classifier = name.split(":")
```

maven/lib/dependabot/maven/update_checker/version_finder.rb
```diff
@@ -257,15 +258,16 @@ def dependency_metadata_url(repository_url)
         end
 
         def dependency_files_url(repository_url, version)
-          group_id, artifact_id = dependency.name.split(":")
+          group_id, artifact_id, classifier = dependency.name.split(":")
```
feelepxyz, line 261: Looks like this also needs to change in `dependency_metadata_url`

feelepxyz: Looks like there's also quite a few instances where we split and get the last element as the artifact id in the MetadataFinder, e.g. https://github.com/dependabot/dependabot-core/blob/master/maven/lib/dependabot/maven/metadata_finder.rb#L26

a1flecke: @feelepxyz From what I can tell the classifier itself it not used in the [url for fetching maven metadata](https://github.com/dependabot/dependabot-core/blob/master/maven/lib/dependabot/maven/update_checker/version_finder.rb#L250).  Mulesoft dependencies use classifiers a lot, an example being [the pom file in this repo I created for a Mulesoft support ticket](https://github.com/a1flecke/mule-assert-equals-tester/blob/assert-equals/pom.xml#L58-L82).  I checked a [couple](https://repository.mulesoft.org/nexus/content/repositories/releases/org/mule/connectors/mule-http-connector/maven-metadata.xml) of the [metadata paths](https://repository.mulesoft.org/nexus/content/repositories/releases/org/mule/connectors/mule-jms-client/maven-metadata.xml) for those dependencies and only `group_id` and `artifact_id` were needed.

Thank you for catching the other `.last` calls.  I will fix those

a1flecke:
@feelepxyz I could use some assistance here.  I have been fighting what appears to be some flaky tests with [VCR and requests to rubygems](https://app.circleci.com/pipelines/github/dependabot/dependabot-core/353/workflows/13f46f9d-1b18-47c3-82e7-75a3d22d99cc/jobs/14518/steps)  Are these tests known to have issues?

Example results when running locally is VS Code devcontainer using the same rspec seed:
```
Finished in 2 minutes 38.6 seconds (files took 1.29 seconds to load)
680 examples, 0 failures

Randomized with seed 50232

[dependabot-core-dev] ~/dependabot-core/bundler $ pwd
/home/dependabot/dependabot-core/bundler
```

a1flecke:
@feelepxyz It seems like the backports gem is missing a spec fixture in `"fixtures/ruby/rubygems_responses/`.  Can you provide direction on how I generate the `info-backports` file?

I believe [this helper](https://github.com/dependabot/dependabot-core/blob/master/bundler/spec/shared_contexts.rb#L22-L31) will then setup the missing HTTP stub

The CircleCI failure.
```
     Failure/Error: raise VCR::Errors::UnhandledHTTPRequestError.new(vcr_request)
 
     
 
     Dependabot::SharedHelpers::ChildProcessFailed:
 
       Child process raised VCR::Errors::UnhandledHTTPRequestError with message: 
 
     
 
       ================================================================================
 
       An HTTP request has been made that VCR does not know how to handle:
 
         GET https://index.rubygems.org/info/backports
 ```

**Comment by feelepxyz** [comment]
> @feelepxyz It seems like the backports gem is missing a spec fixture in `"fixtures/ruby/rubygems_responses/`. Can you provide direction on how I generate the `info-backports` file?

We currently don't have any tooling but if you copy this https://index.rubygems.org/info/backports into a file named `info-backports` and add it https://github.com/dependabot/dependabot-core/blob/master/bundler/spec/fixtures/ruby/rubygems_responses/ it should work.

These stubs are pretty annoying as it seems like VCR/webmock don't always pick up that they are needed. Possibly because we run Bundler in a forked process 🤷‍♂️

File: maven/lib/dependabot/maven/metadata_finder.rb [hunk 3]
```diff
@@ -118,6 +115,12 @@ def dependency_pom_file
         @dependency_pom_file = Nokogiri::XML("")
       end
 
+      def dependency_artifact_id
+        _group_id, artifact_id, = dependency.name.split(":")
```
feelepxyz, line 119: Would be good to also include the classifier here to make it clear we expect the name to have three components in case someone missed this in a future refactor:

```suggestion
        _group_id, artifact_id, _classifier = dependency.name.split(":")
```

maven/lib/dependabot/maven/update_checker/version_finder.rb
```diff
@@ -248,7 +249,7 @@ def pom
         end
 
         def dependency_metadata_url(repository_url)
-          group_id, artifact_id = dependency.name.split(":")
+          group_id, artifact_id, = dependency.name.split(":")
```
feelepxyz, line 252:
```suggestion
          group_id, artifact_id, _classifier = dependency.name.split(":")
```

feelepxyz:
Looks like maven, common and bundler are passing so going to :shipit: this 🎉

a1flecke:
@feelepxyz   Thanks for your help with this.

honnix:
Sorry for digging into this old thread. This is a great improvement btw. Thanks for doing this.

We hit a corner case of this change so would like to clarify whether it is intended or not.

When checking whether a dependency is internal, meaning some submodule of a project, `classifier` is appended to the full name; however when building up the list of internal dependencies, there is no classifier appended because it is hard if not impossible to known. So the problem we have now is Dependabot would try to check newer version of a submodule if in parent pom.xml we have:
```pom.xml
<dependencyManagement>
  <dependencies>
    <groupId>foo</groupId>
    <artifactId>bar</artifactId>
    <version>${project.version}</version>
    <classifier>foobar</classifier>
  </dependencies>
</dependencyManagement>
```

Because it believes `foo:bar:${project.version}:foobar` is not an internal dependency.

Due to another no-so-standard setup we have, a submodule would have an auto-bumped version of artifacts published every time CI passes in main branch (without changing source code), and the result of that is the published artifacts have higher version than the project's. In this case, for this classified dependency, Dependabot would try to bump the project version in parent pom.xml because it is a property, and that is not what we expect since it results an invalid parent pom.xml because all children still refer to the old version in `parent` section.

My question is, whether this is intended change? Would it be better we exclude `classifier` when checking internal dependencies? A possible consequence of skipping `classifier` is that the classified artifact is indeed not published by a submodule of the current project, but the probability of this setup would be very low.

a1flecke:
@honnix Unfortunately a lot of time has passed since I created this PR.  Given the complexity of Maven I cannot be 100% certain, but you are likely accurate that using a classifier for intra-project dependencies is not going to work since the classifier comes into play when an artifact is published.

honnix:
@a1flecke Thanks for replying and the confirmation. I will try to work on a PR to address this.

PR #1936, opened 2020-06-23
Prefer exact match for 'security' label

qnighy:
We have a `No Security Risk` label and dependabot sometimes chooses it over the `security` label. This patch attempts to solve it by teaching dependabot the preference.


common/lib/dependabot/pull_request_creator/labeler.rb
```diff
@@ -193,7 +193,8 @@ def security_label_exists?
       end
 
       def security_label
```
feelepxyz, line 195:
```suggestion
       # Find the exact match first and then fallback to * security* label
      def security_label
```

common/lib/dependabot/pull_request_creator/labeler.rb
```diff
@@ -193,7 +193,8 @@ def security_label_exists?
...
-        labels.find { |l| l.match?(/security/i) }
+        labels.find { |l| l == DEFAULT_SECURITY_LABEL } ||
+          labels.find { |l| l.match?(/security/i) }
       end
```
feelepxyz, line 198: Could you also add a test case for this in the labeler_spec? You could do something similar to this: https://github.com/dependabot/dependabot-core/commit/503c77e1bcda22d9c06185afca0a8c61373307c4#diff-27da36e17c0a714c6d5db6824b28b4fe but adding `let(:includes_security_fixes) { true }` to the context.

PR #2273, opened 2020-07-01
Use Maven version ranges for ignored_versions in Maven and Gradle

### LINKED ISSUE #1845: Config file doesn't validate Maven versions.
**Package manage/ecosystem**

`java:maven`

When I use a correct [version requirement string](https://maven.apache.org/pom.html#dependency-version-requirement-specification) in Maven, I expect my comfig.yml file to be validated, however instead I get the following error (both as an issue in a repo and using the config validator online): 

```
The property '#/update_configs/0/ignored_updates/0/version_requirement' is an invalid version requirement for a java:maven ignore condition
```

If I were to remove the `version_requirement` line from that section of the config file, it works fine. Below is an example of the `ignored_updates` section of a config file I'm testing.

```yml
ignored_updates:
  - match:
      dependency_name: "org.springframework:spring-core"
      version_requirement: "[4.0.0.RELEASE,)"
```

_In this example, all updates greater than Spring 3 should be ignored._

One workaround to please the config validator is to remove the comma, but looking at the source code, I think Dependabot Core would treat that as an "equals" constraint as opposed to a "range" constraint".

I've taken a look at the [requirement.rb](https://github.com/dependabot/dependabot-core/blob/master/maven/lib/dependabot/maven/requirement.rb) file for the Maven part of the Dependabot core and tested the functions. The core handles version ranges as expected; it's just the config validator that is throwing issues. I tried to look at its code, but I couldn't find it. I am happy to take a look at this issue some more if someone can point me in the right direction.

eager:
Despite the [docs](https://dependabot.com/docs/config-file/#ignored_updates), which say that ignored updates should match the package manager, it looks like Maven & Gradle ranges have always been implemented using Gem’s pattern.

I’m not 100% confident that this will fix the actual issue with `config.yml`, since the implementation that parses that file into a Dependabot run doesn’t seem to be open source.

fixes #1845

### DIFF
--- a/gradle/lib/dependabot/gradle/update_checker/version_finder.rb
+++ b/gradle/lib/dependabot/gradle/update_checker/version_finder.rb
@@ -15,6 +15,8 @@ module Dependabot
         GRADLE_PLUGINS_REPO = "https://plugins.gradle.org/m2"
         TYPE_SUFFICES = %w(jre android java).freeze
 
+        GRADLE_RANGE_REGEX = /[\(\[].*,.*[\)\]]/.freeze
+
         def initialize(dependency:, dependency_files:, credentials:,
                        ignored_versions:, raise_on_ignored: false,
                        security_advisories:)
@@ -97,7 +99,7 @@ module Dependabot
           filtered = possible_versions
 
           ignored_versions.each do |req|
-            ignore_req = Gradle::Requirement.new(req)
+            ignore_req = Gradle::Requirement.new(parse_requirement_string(req))
             filtered =
               filtered.
               reject { |v| ignore_req.satisfied_by?(v.fetch(:version)) }
@@ -126,6 +128,12 @@ module Dependabot
           possible_versions.select do |v|
             v.fetch(:version) > version_class.new(dependency.version)
           end
+        end
+
+        def parse_requirement_string(string)
+          return string if string.match?(GRADLE_RANGE_REGEX)
+
+          string.split(",").map(&:strip)
         end
 
         def wants_prerelease?
--- a/gradle/lib/dependabot/gradle/update_checker/version_finder.rb
+++ b/gradle/lib/dependabot/gradle/update_checker/version_finder.rb
@@ -97,7 +97,7 @@ module Dependabot
           filtered = possible_versions
 
           ignored_versions.each do |req|
-            ignore_req = Gradle::Requirement.new(req.split(","))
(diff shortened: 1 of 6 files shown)


gradle/lib/dependabot/gradle/update_checker/version_finder.rb
```diff
@@ -97,7 +97,7 @@ def filter_ignored_versions(possible_versions)
           filtered = possible_versions
 
           ignored_versions.each do |req|
-            ignore_req = Gradle::Requirement.new(req.split(","))
+            ignore_req = Gradle::Requirement.new(req)
```
feelepxyz, line 100: @eager thanks for this! We'll still need to support the old-style ruby syntax which exists in a bunch of config files but we should be able to support both.

@jurre looks like we might need to include your fix for gradle/maven/nuget requirements in core too.

jurre: Ah nice one 🎉 I'll try to port it over, @eager if you're up for it, the code we have to handle this elsewhere is:

```ruby
GRADLE_RANGE_REGEX = /[\(\[].*,.*[\)\]]/.freeze

def self.parse_requirement_string(package_manager, string)
  if %w(nuget gradle maven).include?(package_manager) &&
     string.match?(GRADLE_RANGE_REGEX)
    return string
  end

 string.split(",").map(&:strip)
end
```

eager: @feelepxyz @jurre thanks for the feedback!

Just to be clear: should I update this PR to handle both styles, or is @jurre going to add a fix elsewhere in `dependabot-core`? I’m good either way 👍.

eager: It was pretty trivial to do, so I went ahead and got it started in `7845ad7`. _Not_ sharing code between package manager modules seems to be the convention, but let me know if anything should be DRY-er.

jurre: Thank you @eager 🎉 

> Not sharing code between package manager modules seems to be the convention

It's definitely something we're trying to improve where it makes sense but for now I think this looks good 👍

eager:
@feelepxyz 👍 done in ea10e6322

jurre:
Thanks so much @eager

PR #2277, opened 2020-07-02
Python: support binary path deps with requirements

feelepxyz:
Stop trying to remove imports from binary path dependencies when there
are a requirements files.

This bug only surfaced when you have both a requirements file and a
binary path dependency, where the FileParser tried to string match on a
biary file raising an `ArgumentError: invalid byte sequence in UTF-8`.


feelepxyz:
Ran tests locally instead 🙈

PR #2281, opened 2020-07-05
Parse dependencies from Python setup.cfg files

### LINKED ISSUE #2133: Add support for Python dependencies in `setup.cfg`
Hi !

Ever since `setuptools` started supporting setup configuration files, I've been using them instead of hardcoding configuration values in the `setup.py` script. However, it does not seem like these files are supported by Dependabot right now.

Adding support would be fairly simple: check for `setup.cfg`, parse it as an INI configuration file, and check the value in `setup_requires`, `tests_require`, and `install_requires` just like a `requirements.txt` file.

I've been using Dependabot extensively for my Rust projects but this is stopping me from extending that to my Python projects as well. I'd love to see `setup.cfg` support implemented!

akaihola:
Fixes #2133

Also discussed here:
https://github.community/t/dependency-graph-does-not-support-setup-cfg-for-python/2576

This implementation modifies the `parse_setup` Python helper to use
```python
distutils.core.run_setup(
    <path to unmodified setup.py file>,
    stop_after="init"
)
```
and the results are collected using the custom `setup()` function as before.

It also catches all exceptions. In case of errors or no collected dependencies, it reverts to the old method running a patched `setup.py` using `exec()` method.

I haven't yet added test cases for this. I'll need to install the Ruby environment and study how to run the test suite – pointers to documentation appreciated!

### DIFF
--- a/python/helpers/lib/parser.py
+++ b/python/helpers/lib/parser.py
@@ -1,3 +1,4 @@
+from distutils.core import run_setup
 from itertools import chain
 import glob
 import io
@@ -90,6 +91,21 @@ def parse_setup(directory):
             for key in extras_require_dict:
                 for req in extras_require_dict[key]:
                     parse_requirement(req, 'extras_require:{}'.format(key))
+
+        try:
+            # Use distutils.core.run_setup() - this will also parse setup.cfg
+            setup_result = run_setup(directory + "/setup.py", stop_after="init")
+        except Exception:
+            # Try with the original exec() method below
+            pass
+        else:
+            setup(
+                setup_requires=getattr(setup_result, "setup_requires", None),
+                install_requires=getattr(setup_result, "install_requires", None),
+                tests_requires=getattr(setup_result, "tests_requires", None),
+                extras_require=getattr(setup_result, "extras_require", None),
+            )
+
         setuptools.setup = setup
 
         def noop(*args, **kwargs):
@@ -132,7 +148,8 @@ def parse_setup(directory):
         # Run as main (since setup.py is a script)
         __name__ = '__main__'
 
-        # Exec the setup.py
-        exec(content) in globals(), locals()
+        # Exec the setup.py if distutils.core.run_setup() above didn't succeed
+        if not setup_packages:
+            exec(content) in globals(), locals()
 
     return json.dumps({ "result": setup_packages })


python/helpers/lib/parser.py
```diff
@@ -1,3 +1,4 @@
+from distutils.core import run_setup
```
pganssle, line 1: Best to avoid using distutils, we're trying to actively deprecate it.

akaihola: Ok, so I take it that I should close this PR?

Is there an alternative best practice for getting the dependencies of a Python package which uses `setuptools`, no matter whether the dependencies are listed in `setup.py` or `setup.cfg`?

pganssle: This question is actually a bit difficult to answer. Assuming you actually just want to get the dependencies, for `setuptools` projects, the biggest issue you'll have is that many projects unfortunately don't know about (or haven't updated to use) environment markers, so it's distressingly common to see stuff like this:

```python
requires = ["some_dependency"]
if sys.version_info[0] < 2:
     requires.append("some_py2_only_dependency")
```

E.g. from [s3transfer](https://github.com/boto/s3transfer/blob/97d46ec0f0e687f5ac71fca9a5ff6e1c7d68a675/setup.py#L13-L21). This means that depending on the machine where you run `setup.py`, you will get a different answer for the values of `install_requires`. The correct way to do this is to declare a fixed set of dependencies like so:

```
some_dependency
some_py2_only_dependency ; python_version < '2.7'
```

In which case even the conditional dependencies will be included in the wheel metadata.

Assuming that you are comfortable with accepting "whatever will be included in the wheel if I were to build it in whatever worker environment" (seems reasonable), then I would say that the best thing to do is to implement enough of a PEP 517/518 backend to execute [prepare_metadata_for_build_wheel](https://www.python.org/dev/peps/pep-0517/#prepare-metadata-for-build-wheel), and parse the dependency metadata from the `.dist-info/METADATA` file. The `pep517` library should make this easy enough.

If the project doesn't have a `pyproject.toml` file, I recommend defaulting to use `build-backend=setuptools.build_meta:__legacy__` and `requires = ["setuptools >= 40.8.0", "wheel"]`, as we're doing in [the python-build frontend](https://github.com/FFY00/python-build/blob/367009472af3d173e381625c62071ee204ab99b6/src/build/__init__.py#L27). I think the majority of setuptools projects will work properly with these defaults. You can also fall back to using `pip wheel` to generate an actual wheel file and then extract the metadata from there, though that will consume more resources than necessary.

Using PEP 517 also has the advantage that it works automatically with any PEP 517 backend, not just `setuptools`. (We in the PyPA would also probably not be unhappy if you said, "We only support packages that work well with PEP 517", as a further spur to get people to actively adopt the new build system).

That said, I'm not exactly sure why you want this sort of generic information about a package's depdencies — my understanding of dependabot is that it is supposed to update your dependencies, no? So don't you need to know not only what they are, but also exactly where they are defined, so that you can make a PR to automatically update them?

akaihola: >my understanding of dependabot is that it is supposed to update your dependencies,

GitHub's [Dependency Graph](https://docs.github.com/en/github/visualizing-repository-data-with-graphs/exploring-the-dependencies-and-dependents-of-a-repository) is implemented using dependabot. The way I arrived at this issue is that I wanted to simply see [my project's dependencies in the GitHub UI](https://github.com/akaihola/darker/network/dependencies) and was surprised to see this message instead:

> [**setup.py**](https://github.com/akaihola/darker/blob/master/setup.py) has no dependencies or is too large to display

So GitHub's Dependency Graph simply doesn't work for any Python packages which declare `setup_requires` in `setup.cfg` instead of `setup.py`.

I then found the [tip](https://stackoverflow.com/a/61754034/15770) to use `distutils.core.run_setup(..., stop_after="init").install_requires` and figured that dependabot could do that, and also as a side effect gain a possibly more robust way to extract dependencies from `setup.py` files than the current [parse_setup()](https://github.com/dependabot/dependabot-core/blob/v0.118.7/python/helpers/lib/parser.py#L60) method which modifies `setup.py` and `exec()`s it in a patched Python environment. `parse_setup()` could still be used as a fallback for those cases where `run_setup()` fails.

It sounds like a great idea to make GitHub's Dependency Graph only work on modern well-formed Python packages. It should indeed encourage authors to adopt good practices in packaging. The current situation actually does the reverse – it discourages adopting e.g. a declarative `setup.cfg` and `pyproject.toml` and favors old-style packages with a traditional executable `setup.py`

akaihola: I'm still eager to help make GitHub's [Dependency Graph](https://docs.github.com/en/github/visualizing-repository-data-with-graphs/exploring-the-dependencies-and-dependents-of-a-repository) work for modern Python repositories. @pganssle could you advise what would be the most efficient and acceptable way for me to contribute?

merwok: You could detect if a `pyproject.toml` file is present, then use `pep517` or `python-build` to create a wheel distribution from the source and inspect the metadata in that wheel (there are a few Python modules to do that, metadata and wheel are specified in standard PEPs).

nocarryr: @akaihola would the setuptools [Configuration API](https://setuptools.readthedocs.io/en/latest/setuptools.html#configuration-api) not be a viable solution here?

```python
from setuptools.config import read_configuration
# ...
cfg = read_configuration('setup.cfg')
cfg_setup_requires = cfg.get('options', {}).get('setup_requires', [])
# ...
```

The requirement specifiers could then just be merged with those collected from `setup.py`, correct?

akaihola: Thanks @merwok and @nocarryr, I'll carve out some time to explore those ideas.

gaborbernat: I'd avoid building wheels, all you need is to call https://www.python.org/dev/peps/pep-0517/#prepare-metadata-for-build-wheel to get the metadata files and then use [importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html#requirements):

```
from importlib.metadata import Distribution
requires = Distribution.at("/path/to/metadata").requires
```

akaihola: On Python 3.8, this script seems to work when run in the root of a package which uses `setuptools` as the build system:
```python
from importlib.metadata import Distribution
from pathlib import Path
from pprint import pprint
from setuptools.build_meta import __legacy__

dist_path = Path(__legacy__.prepare_metadata_for_build_wheel('.'))
dependencies = Distribution.at(dist_path / 'METADATA').requires
pprint(dependencies)
```

This of course won't work if another build system is used. Is there reusable code for getting the build backend from `pyproject.toml`? Could [python-build](https://github.com/FFY00/python-build) be added as a dependency for `dependabot` and utilized here?

gaborbernat: No reusable code at the moment, but would be fairly trivial to put something together with a toml parser, virtualenv and the above 👍

graingert: I've been using `pep517.meta.load` for this purpose

graingert: > No reusable code at the moment, but would be fairly trivial to put something together with a toml parser, virtualenv and the above +1

@gaborbernat it looks like `load` https://github.com/pypa/pep517/blob/master/pep517/meta.py#L62 is a nice bit of reusable code. I'm not sure what the deprecation status of it is though

mauvilsa: The deprecation means that people should use [python-build](https://pypi.org/project/python-build/) for building, see https://github.com/pypa/pep517/issues/91

graingert: Right so pep517.meta.load is still supported

jurre: > GitHub's [Dependency Graph](https://docs.github.com/en/github/visualizing-repository-data-with-graphs/exploring-the-dependencies-and-dependents-of-a-repository) is implemented using dependabot.

It's actually implemented separately. We hope to merge whatever is possible from these two implementations at some point, but we currently have no concrete plans to do that. Unfortunately any changes made to dependabot won't make it back into the dependency graph.

akaihola: @jurre ok so that significantly brings down my motivation... If at all possible, please be a good Open Source citizen, use the code base as such, and contribute back all your modifications. This way you'll also be able to benefit from volunteer work, like potentially in this case. Thanks!

jurre: Dependency Graph and Dependabot are completely separate projects and share no code between them, any improvements to dependabot-core are made in this repo which is publicly available, but this is not the case for Dependency Graph.

gaborbernat: @jurre considering Github is not interested in improving ``Dependency Graph`` (as you pointed out) at the moment, any chance it can be open-sourced (part of it at least?), so language communities can improve it themselves via PRs?

jurre: We're definitely interested in improving Dependency Graph, there's a whole team working on it, it's just not related to Dependabot is what I was saying.

gaborbernat: So how can we get the ball moving on improving python dependency discovery for ``Dependency Graph`` (in essence to use PPE-517 API to determine it, rather than setup.py/setup.cfg)?

jurre: I think reaching out to support is the easiest way to get in touch with the team working on Dependency Graph.

gaborbernat: Do you have a link to this support form? 🤔

jurre: https://support.github.com/

pganssle: Does the result of Dependency graph feed into something that matters? Why are y'all trying to help Github fix some broken and proprietary thing that they seem to not care about?

mauvilsa: There is now a feature request to have an official way to get the dependencies of a package https://github.com/pypa/build/issues/181

jurre: Thanks @mauvilsa!

nschloe:
Can probably be closed now that https://github.com/dependabot/dependabot-core/pull/3423 has landed.

jurre:
I think this has landed in #4323

python/helpers/lib/parser.py
wwuck: https://pypa-build.readthedocs.io/en/latest/api.html#build.util.project_wheel_metadata

sschuberth: > I would say that the best thing to do is to implement enough of a PEP 517/518 backend to execute [prepare_metadata_for_build_wheel](https://www.python.org/dev/peps/pep-0517/#prepare-metadata-for-build-wheel), and parse the dependency metadata from the `.dist-info/METADATA` file.

But that wouldn't work if one would like to just *determine* the transitive dependencies without actually building the project, or?

graingert: @sschuberth that's what `build.util.project_wheel_metadata` does

sschuberth: > I would say that the best thing to do is to implement enough of a PEP 517/518 backend to execute [prepare_metadata_for_build_wheel](https://www.python.org/dev/peps/pep-0517/#prepare-metadata-for-build-wheel), and parse the dependency metadata from the `.dist-info/METADATA` file.

Is there any complete example for doing that on a given `setup.py` / `requirements.txt` file?

sschuberth: Thanks for this, but I'm a complete Python noob, and I'm running into a bunch of missing import errors that I cannot seem to solve (on Python 3.6). So what I'd need it a *complete* example / standalone `.py` file that takes a directory / file as an argument and works out the box 🙄

graingert: @sschuberth what errors are you getting? You should try this on a recent version of python eg 3.9 or 3.10

merwok: For older versions, you have to install and use the backport `importlib_metadata` (notice the underscore)

PR #2307, opened 2020-07-11
Support docker for GitHub Actions

### LINKED ISSUE #5541: Bump `docker` image references in GitHub Actions workflow
Instead of running github workflows or gitlab-ci on a :latest docker image, using fixed versions provides better stability.
It would therefore be awesome if dependabot could create pull requests with a bumped (but not :latest) image for a ci-setup.

I guess getting the versions of docker images from dockerhub is already implemented for the "docker" mode which bumps Dockerfiles.

Cheers
APN

frezbo:
This is mostly some hacks I did to get `docker` dependency update support for `github-actions`. Due to the way the current code is organized it seems harder to use the existing `docker` *package_manager*. Are we going to support some kind of sub-dependecy for `github-actions` type?

Fix https://github.com/dependabot/dependabot-core/issues/5541

*DO NOT MERGE*

### DIFF
--- a/docker/lib/dependabot/docker/file_parser.rb
+++ b/docker/lib/dependabot/docker/file_parser.rb
@@ -1,61 +1,37 @@
 # frozen_string_literal: true
 
-require "docker_registry2"
-
 require "dependabot/dependency"
 require "dependabot/file_parsers"
 require "dependabot/file_parsers/base"
 require "dependabot/errors"
-require "dependabot/docker/utils/credentials_finder"
+require 'dependabot/docker/utils/dockerfile_parser'
 
 module Dependabot
   module Docker
     class FileParser < Dependabot::FileParsers::Base
       require "dependabot/file_parsers/base/dependency_set"
 
-      # Details of Docker regular expressions is at
-      # https://github.com/docker/distribution/blob/master/reference/regexp.go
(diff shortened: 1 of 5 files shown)


frezbo:
I did a code cleanup in the second commit, if it's the right direction, I can now think of fixing/adding specs and also update the `file_updater` for `docker` package in `github_actions`. It;s just a silly code now.

channelbeta:
Hi @frezbo , I'd love to see this code merged. Do you know if there is any open issue open about it as well?

jeffwidman:
👋 Thanks, and sorry for the horribly slow reply from the team. We're just now starting to catchup on some of the backlog that we couldn't get to while transitioning to GitHub's infra.

I appreciate the PR, but new ecosystem support is on pause for now (and has been for a while). Once we're ready to accept new ecosystems, then we'll be ready to have more discussions on this. We've also explored the idea of making Dependabot pluggable to support community contributions, but a number of bits are still unclear, so no idea if / when that might ship.

ssbarnea:
@jeffwidman there is one aspect here: github actions are as internal to github as they can be, so adding support for github own stuff,.... should maybe be a priority? Maybe it was not before the acquisition, but now?

Issue #2229, opened 2018-10-16
Run `go mod tidy` in Go projects


TheRedSpy15:
This would be awesome to have as a Go developer

Rather than just updating modules, running `go mod tidy` would add and remove them too

siavashs:
We use `-mod=readonly` in our workflows and get this error on the PRs opened by the bot:
```
go: updates to go.sum needed, disabled by -mod=readonly
```

aclevername:
> We use `-mod=readonly` in our workflows and get this error on the PRs opened by the bot:
> 
> ```
> go: updates to go.sum needed, disabled by -mod=readonly
> ```

same here! The only solution for us so far is to manually run a `go mod tidy` on the same branch as the PR, taking away from the lovely automation that we want 😢

jurre:
We're starting to roll out this feature on the GH native version of Dependabot, if anyone would like to preview this, please let me know which repo or account you'd like it enabled on.

jurre:
https://github.blog/changelog/2020-10-19-dependabot-go-mod-tidy-and-vendor-support/ 🎉

dfarrell07:
Is this expected to be enabled for all repositories now? As best I can tell, we're not seeing it yet.

For example, [this PR from Dependabot](https://github.com/submariner-io/submariner-operator/pull/776) didn't include [these `go mod tidy` updates](https://github.com/submariner-io/submariner-operator/pull/837). The same for [this version bump](https://github.com/submariner-io/lighthouse/pull/352) and [these `go mod tidy` changes](https://github.com/submariner-io/lighthouse/pull/372).

Thanks for this great tooling, we really find it useful and appreciate it!

dfarrell07:
> Is this expected to be enabled for all repositories now? As best I can tell, we're not seeing it yet.
> 
> For example, [this PR from Dependabot](https://github.com/submariner-io/submariner-operator/pull/776) didn't include [these `go mod tidy` updates](https://github.com/submariner-io/submariner-operator/pull/837) (or [this Dependabot PR from last week](https://github.com/submariner-io/submariner-operator/pull/7930)). The same for [this version bump](https://github.com/submariner-io/lighthouse/pull/352) and [these `go mod tidy` changes](https://github.com/submariner-io/lighthouse/pull/372).
> 
> Thanks for this great tooling, we really find it useful and appreciate it!

[...]

PR #2317, opened 2020-07-14
fix: go mod tidy instead of go get

matansh:
go get adds new dependencies but leaves the old, now unused, dependency versions in the go sum file.

this fixes: https://github.com/dependabot/dependabot-core/issues/2229

FYI im not a ruby dev or anything, just a dependabot user with broken pipelines :)
if this is unhelpful / wrong or anything like that feel free to close my PR, no harm done


matansh:
while investigating the failing test in circle-ci i found this little documentation line - 
```ruby
# This happens via `go mod tidy`, which we currently can't run, as we
# need to the whole source repo
pending "removes old entries from the go.sum" do
```
which explains the issue, sorry for the spam :sob:

PR #2324, opened 2020-07-16
Fix docker-dev-shell ruby/php build

feelepxyz:
When running docker dev shell it bails adding the keyserver. Adding the
repository seems to work but not sure if we're implicitly allowing any
key to be trusted if the repository gets pawned.

Update comment to say brightbox ruby 2.6 now installs 2.6.6.


feelepxyz:
Looks like rubocop has been updated to 0.88 and the new version no longer loads config files from the parent directory possibly because of this change: https://github.com/rubocop-hq/rubocop/pull/8239

.rubocop.yml
```diff
@@ -42,7 +42,7 @@ Metrics/ModuleLength:
   Max: 350
 
 Metrics/CyclomaticComplexity:
-  Max: 10
+  Max: 15
```
feelepxyz, line 45: Looks like the new algorithm was detecting a bunch of new warnings with 10 so upped to 15 to aim for least amount of changes. Keen to drop these as I feel like they just slow us down and don't actually add much values as we usually disable them and never review this.

Rakefile
```diff
@@ -41,7 +41,7 @@ namespace :ci do
     packages = changed_packages
     puts "Running rubocop on: #{packages.join(', ')}"
     packages.each do |package|
-      run_command("cd #{package} && bundle exec rubocop")
+      run_command("cd #{package} && bundle exec rubocop -c ../.rubocop.yml")
```
feelepxyz, line 44: I think this change broke the previous behaviour of checking for config files in the parent directory https://github.com/rubocop-hq/rubocop/pull/8239

feelepxyz: Although not sure this is true after looking at the code a bit closer

.rubocop.yml
jurre: Yeah I have mixed feelings about this, I find that codebases where we're strict about following this tend to be easier to grok and change over time, but at this point there are so many exceptions that it kind of defeats the point of having the rule.

feelepxyz: hm yeah true, should we be more strict and not introducing any new disables and getting rid of any we encounter? Up for that.

jurre: Yeah if you're up for it let's do it! 🎉

feelepxyz:
@jurre want to give this a re-review?

Dockerfile
```diff
@@ -98,9 +98,7 @@ RUN npm install elm@0.18.0 \
 
 # Install PHP 7.4 and Composer
 ENV COMPOSER_ALLOW_SUPERUSER=1
-RUN echo "deb http://ppa.launchpad.net/ondrej/php/ubuntu bionic main" >> /etc/apt/sources.list.d/ondrej-php.list \
-    && echo "deb-src http://ppa.launchpad.net/ondrej/php/ubuntu bionic main" >> /etc/apt/sources.list.d/ondrej-php.list \
-    && apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 4F4EA0AAE5267A6C \
+RUN add-apt-repository ppa:ondrej/php \
     && apt-get update \
     && apt-get install -y php7.4 php7.4-cli php7.4-xml php7.4-json php7.4-zip php7.4-mbstring php7.4-intl php7.4-common php7.4-gettext php7.4-curl php7.4-bcmath php7.4-gmp php7.4-imagick php7.4-gd php7.4-redis php7.4-soap php7.4-ldap php7.4-memcached php7.4-sqlite3 php7.4-apcu php7.4-tidy php7.4-mongodb php7.4-zmq php7.4-mysql php7.4-imap php7.4-geoip \
     && curl -sS https://getcomposer.org/installer | php \
```
feelepxyz, line 104: Looks like this is a better way https://getcomposer.org/download/

feelepxyz: Think I'll fix this in a new PR to not block red builds on master

PR #2341, opened 2020-07-20
Fix: Use directory instead of repo

localheinz:
This PR

* [x] fixes the wording of a message


localheinz:
Thank you, @feelepxyz!

PR #2342, opened 2020-07-20
Fix cargo requirement message in specs

jurre:
The layout of the error seems to have changed slightly in the latest
version of Cargo.

### DIFF
--- a/cargo/spec/dependabot/cargo/update_checker/version_resolver_spec.rb
+++ b/cargo/spec/dependabot/cargo/update_checker/version_resolver_spec.rb
@@ -131,7 +131,7 @@ RSpec.describe Dependabot::Cargo::UpdateChecker::VersionResolver do
           to raise_error do |error|
             expect(error).to be_a(Dependabot::DependencyFileNotResolvable)
             expect(error.message).
-              to include("version for the requirement `regex = \"= 99.0.0\"`")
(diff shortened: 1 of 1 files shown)


feelepxyz:
@jurre ops just deleted these specs 🙈 didn't get it to fail locally and the test case didn't seem valid, the version of regex isn't yanked and if I change it to a yanked version it doesn't raise at all so seemed like a bad/confusing spec 🤷‍♂️ 

Did you manage to reproduce the failure locally?

feelepxyz:
I think I've been bashing against core test suite a bit too much so just went for the 🔪 ...

jurre:
Yeah I did by installing the latest cargo version locally, I assume it would also work to rebuild the docker image (which CI probably does).

I ran this locally to install the latest rust:

```
RUSTUP_HOME=/opt/rust PATH="${PATH}:/opt/rust/bin"  curl https://sh.rustup.rs -sSf | sh -s -- -y
source $HOME/.cargo/env
```

feelepxyz:
> I ran this locally to install the latest rust and cargo versions:

Sweet! Reverted my change and will merge this in 🙇

feelepxyz:
Going to rename the spec from yanked to missing

jurre:
Thanks @feelepxyz <3

PR #2343, opened 2020-07-20
Fix: Configure phpstan level via phpstan.neon

localheinz:
This PR

* [x] configures the `phpstan` level via `phpstan.neon`

Slightly related to #2338.


localheinz:
@feelepxyz 

I'm unsure why a build has not been triggered.

localheinz:
Thank you, @feelepxyz!

PR #2344, opened 2020-07-20
Fix: Do not bother using array

localheinz:
This PR

* [x] uses a `string` instead of an `array` for a list of scripts that contains only a single entry


localheinz:
@feelepxyz 

I'm unsure why a build has not been triggered.

localheinz:
Thank you, @feelepxyz!

Issue #2293, opened 2020-07-07
[Private Registries] Unable to fetch repo metadata from private registry


milind009:
Recently, I encountered an issue where dependabot-core was unable to fetch repo metadata i.e releases text, changelog, commits, etc. from private registry in `metadata_finder` class for `npm_and_yarn`. Upon further investigation, I found out that while fetching the metadata it uses `Bearer` authorization to authenticate whereas I was using a `Basic` base64 encoded auth token. 

Is there a reason behind only using Bearer authorization in the above scenario whereas in other scenarios (i.e `getting latest version` of package from registry), it decides the authorization based on the token like here : https://github.com/dependabot/dependabot-core/blob/9f7b26453519e1b2a2db21d19078d4a65755bf6d/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/registry_finder.rb#L73

asciimike:
We [just launched support for private registries](https://github.blog/2021-03-15-dependabot-private-dependencies/) in beta, which may help fetch info from private registries. Closing this as it should provide the ability to grab that info if you provide auth info, but if it doesn't work, feel free to re-open.

PR #2348, opened 2020-07-21
NPM and Yarn: Fix for getting metadata for direct dependency update

milind009:
**Issue**
For `direct dependency update`, dependabot-core is unable to fetch metadata such as `release notes`, `changelog,`etc. when using a `base64 encoded token` for connecting to a `private` registry as mentioned in the issue: [Issue #2293](https://github.com/dependabot/dependabot-core/issues/2293). This results in a PR without `detailed description` relevant to the updated package.
In case of `subdependency`, it fetches the details from the `public` registry hence does not face this authentication issue.
Attaching some images for clear context:

****Image without detailed description****

****Image with detailed description****

**Cause**
Only `bearer` authorization is enabled for connecting to the `private` registry for fetching metadata whereas in other scenarios( like `getting latest version` of package from registry), it decides the authorization based on the auth token provided. Here is a code reference: https://github.com/dependabot/dependabot-core/blob/9f7b26453519e1b2a2db21d19078d4a65755bf6d/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/registry_finder.rb#L73

**Fix**
Added support for different types of auth tokens by deciding auth headers based on the auth token provided.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/metadata_finder.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/metadata_finder.rb
@@ -214,9 +214,18 @@ module Dependabot
       end
 
       def registry_auth_headers
-        return {} unless auth_token
+        token = auth_token
+        return {} unless token
 
-        { "Authorization" => "Bearer #{auth_token}" }
+        if token.include?(":")
+          encoded_token = Base64.encode64(token).delete("\n")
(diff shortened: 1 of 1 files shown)


jeffwidman:
Closing, per the comment https://github.com/dependabot/dependabot-core/issues/2293#issuecomment-799745626 this is _likely_ now working on `main` already.

If for some reasons it's not and you still plan to work on this, please comment and we can re-open.

PR #2349, opened 2020-07-21
Fix for PR description limit issue for AzureDevOps(ADO)

milind009:
**Issue**
While running dependabot-core on a `AzureDevOps repo`, it failed to create PR for dependency update with error **`PR description limit exceeded (Description limit cannot be more than 4000 characters)`**. However, to avoid this error, `create_pull_request` function in `azure_client` checks whether the `pr_description length` is greater than the `permissible limit ( i.e currently 4000 characters)` and accordingly trims the description.
https://github.com/dependabot/dependabot-core/blob/f4cf8cfdcc44fa4f309af2df264a3dabe7ee5b7b/common/lib/dependabot/clients/azure.rb#L160

**Cause**
The error was caused due to a `special unicode character` i.e 🎉 which occured multiple times in the PR description. In `UTF-8 encoding`( used in dependabot-core) this character has length `1` whereas in `UTF-16 encoding` (used in ADO), the size of the character is `2`. This diff in size resulted in the pr_description not getting trimmed properly and hence resulting in `pr description length > 4000` on ADO.

**Fix**
Modified `pr_description trimming` part to calculate the `PR description length using UTF-16 encoding` and find the `end index` upto which the description needs to be trimmed.

### DIFF
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -152,17 +152,21 @@ module Dependabot
           "/pushes?api-version=5.0", content.to_json)
       end
 
-      def create_pull_request(pr_name, source_branch, target_branch,
+     def create_pull_request(pr_name, source_branch, target_branch,
                               pr_description, labels)
-        # Azure DevOps only support descriptions up to 4000 characters
+        # Azure DevOps only support descriptions up to 4000 characters in UTF-16 encoding
         # https://developercommunity.visualstudio.com/content/problem/608770/remove-4000-character-limit-on-pull-request-descri.html
         azure_max_length = 3999
-        if pr_description.length > azure_max_length
+
+        if get_pr_description_length(pr_description) > azure_max_length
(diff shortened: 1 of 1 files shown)


common/lib/dependabot/clients/azure.rb
```diff
@@ -206,6 +208,32 @@ def post(url, json)
 
         response
       end
+      
+      # Get pr description length in UTF-16 encoding.
+      def get_pr_description_length(pr_description)
```
jurre, line 213: Maybe we can get the `bytesize` of the string instead, it's 4 chars for an emoji instead of 2, but I'm not super worried about truncating a bit earlier, and it'll simplify the code a lot.

Alternatively we could `pr_description.force_encoding(Encoding::UTF_16).size` and pr_description.force_encoding(Encoding::UTF_16).first(truncate_length).enforce_encoding(Encoding::UTF_8)`

milind009: Yes i think this would simplify the code! Thanks for the suggestion

PR #2352, opened 2020-07-21
[composer] Fix: Require PHP 7.4

localheinz:
This PR

* [x] requires PHP 7.4

💁‍♂️ For reference, see `Dockerfile`:

https://github.com/dependabot/dependabot-core/blob/dd510deaebce566f029adb7261d0d98fc65df49f/Dockerfile#L103.


localheinz:
Thank you, @jurre!

PR #2353, opened 2020-07-21
[composer] Fix: Explicitly require latest stable version of composer/composer

localheinz:
This PR

* [x] explicitly requires the latest stable version of `composer/composer`


jurre:
Thanks, this makes it more clear which version we're using 👍

localheinz:
Thank you, @jurre!

PR #2355, opened 2020-07-21
[composer] Enhancement: Use property type declarations where possible

localheinz:
This PR

* [x] uses property type declarations where possible

Follows #2352.

### DIFF
--- a/composer/helpers/src/DependabotInstallationManager.php
+++ b/composer/helpers/src/DependabotInstallationManager.php
@@ -13,9 +13,9 @@ use Composer\Repository\RepositoryInterface;
 
 class DependabotInstallationManager extends InstallationManager
 {
-    private $installed = [];
-    private $updated = [];
-    private $uninstalled = [];
+    private array $installed = [];
+    private array $updated = [];
+    private array $uninstalled = [];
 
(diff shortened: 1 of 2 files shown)


composer/helpers/src/DependabotInstallationManager.php
```diff
@@ -13,9 +13,9 @@
 
 class DependabotInstallationManager extends InstallationManager
 {
-    private $installed = [];
-    private $updated = [];
-    private $uninstalled = [];
+    private array $installed = [];
+    private array $updated = [];
+    private array $uninstalled = [];
```
jurre, line 18: I'm not very familiar with PHP, but can't it figure out that this is an array by looking at the declaration? Is this annotation only for humans?

localheinz: @jurre 

It's not necessary to use property type declarations, but I would consider it best practice when using PHP 7.4 and newer!

💁‍♂️ For reference, see https://www.php.net/manual/en/migration74.new-features.php#migration74.new-features.core.typed-properties.

localheinz:
Thank you, @feelepxyz and @jurre!

PR #2356, opened 2020-07-21
[composer] Fix: Mark classes as final

localheinz:
This PR

* [x] marks classes as `final`

💁‍♂️ There's probably no reason to extend these classes.


localheinz:
Thank you, @feelepxyz!

PR #2357, opened 2020-07-21
[composer] Fix: Wrap and sort packages in Dockerfile

localheinz:
This PR

* [x] wraps packages in `Dockerfile`
* [x] sorts packages in `Dockerfile`

💁‍♂️ Keeping this list wrapped and sorted makes it easier to spot packages, and also easier to decide where to add additional packages if need be.


jurre:
Thanks 👍

localheinz:
Thank you, @jurre!

Issue #1771, opened 2020-03-30
Dependabot incorrectly freezing Poetry path dependencies, causing schema validation failure


jtbeach:
Now that Dependabot has support for Poetry 1.x (🎉 ), some of my projects are now working, which is fantastic. I investigated the logs of the projects that are failing and they seem to contain the following logs:

```
updater | ERROR <job_27758849> Error processing typeguard (Dependabot::SharedHelpers::HelperSubprocessFailed)
updater | ERROR <job_27758849> 
updater | <job_27758849> [RuntimeError]
updater | <job_27758849> The Poetry configuration is invalid:
updater | <job_27758849>   - [dependencies.blaze_util] {'path': '../blaze_util/', 'version': '0.1.0'} is not valid under any of the given schemas
updater | <job_27758849> 
```

My hypoethesis is that Dependabot has generated a new file that violates the Poetry JSON schema. Looking at the current schema for path dependencies (https://github.com/python-poetry/poetry/blob/44c60c4caed0c4474ca5dde205fe92f9a95b3e26/poetry/json/schemas/poetry-schema.json#L368), you can see that the `version` property is not allowed on path dependencies.

I am not a Ruby programmer, but I think the problematic function is here: https://github.com/dependabot/dependabot-core/blob/master/python/lib/dependabot/python/file_updater/pyproject_preparer.rb#L39. It is trying to freeze the versions of every dependency except the dependency being updated. It is always called with a list of exactly one dependency, the dependency to be updated.

We need to stop this line of code being executed for path dependencies (and probably others as well): https://github.com/dependabot/dependabot-core/blob/master/python/lib/dependabot/python/file_updater/pyproject_preparer.rb#L62

I will raise a PR with want I think is a fix -- I think even my novice Ruby skills should suffice for a code fix, but might need help with writing tests in rspec, etc.

jtbeach:
Have one-line fix for this, but struggling to write the `rspec` tests

jurre:
Hi @jtbeach feel free to open a draft PR with those changes and ping me on it, I'm not sure when I'll have time to look into it but I'd love to see it resolved so I'll see what I can do

jtbeach:
> Hi @jtbeach feel free to open a draft PR with those changes and ping me on it, I'm not sure when I'll have time to look into it but I'd love to see it resolved so I'll see what I can do

Thanks @jurre -- that would be amazing! My draft PR is #1788

PR #1788, opened 2020-04-07
Don't freeze directory dependencies during Poetry update

jtbeach:
Fixes #1771 

Have added what I think is one-line code fix and a sample `pyproject.toml` for testing with a path dependency on `../toml`, but don't know how to write the rspec tests, so can't verify the fix.

For reference, the generated `poetry.lock` file for that test is attached (not in required format, which looks like JSON for the tests). As you can see, the `toml` dependency is versioned in the Lockfile, but we shouldn't carry across that version in the frozen dependencies. It has a `[package.source]` subsection that can be used to detect the type of dependency that it is. The code already handles git dependencies in a special way.

```
[[package]]
category = "main"
description = "Atomic file writes."
marker = "sys_platform == \"win32\""
name = "atomicwrites"
optional = false
python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
version = "1.3.0"

[[package]]
category = "main"
description = "Classes Without Boilerplate"
name = "attrs"
optional = false
python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
version = "19.3.0"

[package.extras]
azure-pipelines = ["coverage", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "zope.interface", "pytest-azurepipelines"]
dev = ["coverage", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "zope.interface", "sphinx", "pre-commit"]
docs = ["sphinx", "zope.interface"]
tests = ["coverage", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "zope.interface"]

[[package]]
category = "main"
description = "Cross-platform colored terminal text."
marker = "sys_platform == \"win32\""
name = "colorama"
optional = false
python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
version = "0.4.3"

[[package]]
category = "main"
description = "Read metadata from Python packages"
marker = "python_version < \"3.8\""
name = "importlib-metadata"
optional = false
python-versions = "!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,>=2.7"
version = "1.6.0"

[package.dependencies]
zipp = ">=0.5"

[package.extras]
docs = ["sphinx", "rst.linker"]
testing = ["packaging", "importlib-resources"]

[[package]]
category = "main"
[...]

PR #2361, opened 2020-07-23
Don't freeze poetry path dependency version

jurre:
This fix was proposed by @jtbeach in in https://github.com/dependabot/dependabot-core/pull/1788 and fixes https://github.com/dependabot/dependabot-core/issues/1771.

When specifying a path dependency in Poetry, we should not try to lock the version of the dependency.


jtbeach:
Thank you so much @jurre!

PR #2363, opened 2020-07-24
Set up actions for CI

jurre:
This is a very rough implementation to get our CI running on actions in parallel, lot's of room for improvement still.

The docker config etc is still on my personal account, will have to move that over.

PR #2372, opened 2020-07-28
Enhancement: Use job matrix

localheinz:
This PR

* [x] uses a job matrix

Follows #2363.

PR #2373, opened 2020-07-28
Fix: Run workflow on pull requests as well

localheinz:
This PR

* [x] runs the previously introduced workflow on pull request events as well

Follows #2363.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,7 @@
 name: CI
-on: [push]
+on:
+  - pull_request
+  - push
 
 jobs:
   ci-image:


localheinz:
@feelepxyz @jurre 

We cannot use secrets in workflows triggered by pushes from forks:

https://github.com/dependabot/dependabot-core/blob/0610325ea0d7044dfca68c89febec635078467b4/.github/workflows/ci.yml#L11-L13

feelepxyz:
> We cannot use secrets in workflows triggered by pushes from forks:

Oh yeah I think we'll need to switch to repo scoped github packages instead which should also work on the fork.

jurre:
> We cannot use secrets in workflows triggered by pushes from forks:

Will look into that and make sure that works 👍 

Regarding the proposed change, why do we need to also build on `pull_request`? Every push is already built, it would be the exact same build as for the latest push of a PR right?

feelepxyz:
> Regarding the proposed change, why do we need to also build on `pull_request`? Every push is already built, it would be the exact same build as for the latest push of a PR right?

Looks like `push` isn't triggered for forks: https://github.com/dependabot/dependabot-core/pull/2372

localheinz:
@jurre 

> Will look into that and make sure that works 👍

👍 

> Regarding the proposed change, why do we need to also build on pull_request? Every push is already built, it would be the exact same build as for the latest push of a PR right?

Perhaps it would make sense to build pushes to `main` only?

```diff
-  - pull_request
+  pull_request: null
-  - push
+  push:
+    - main  
```

jurre:
> Perhaps it would make sense to build pushes to main only?

A workflow that I've been using more recently is pushing a branch to get CI to run, but hold off with opening a pull request until it's in a shape that I'm happy with and CI is passing, so I'd like to keep building on `push` for all branches, I think building on both `push` and `pull_request` is probably fine, just a little bit wasteful.

localheinz:
Thank you, @feelepxyz and @jurre!

PR #2405, opened 2020-08-03
Encode '@' in python HTTP basic auth passwords

jurre:
If a password contains an `@` sign, when we construct a URL with it, it
will result in an invalid URL, since we now have multiple `@` signs,
resulting in the URL parser not being able to determine where the
password stops and the address begins.

### DIFF
--- a/python/lib/dependabot/python/authed_url_builder.rb
+++ b/python/lib/dependabot/python/authed_url_builder.rb
@@ -16,6 +16,8 @@ module Dependabot
           else token
           end
 
+        basic_auth_details = basic_auth_details.gsub("@", "%40")
+
         url.sub("://", "://#{basic_auth_details}@")
       end
     end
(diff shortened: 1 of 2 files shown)


python/lib/dependabot/python/authed_url_builder.rb
```diff
@@ -4,7 +4,7 @@ module Dependabot
   module Python
     class AuthedUrlBuilder
       def self.authed_url(credential:)
-        token = credential.fetch("token", nil)
+        token = credential.fetch("token", nil)&.gsub("@", "%40")
```
hmarr, line 7: Is there something special about the `@` or should we just be using `URI.escape`?

jurre: `URI.escape` and the likes will also encode the `:` which would make the Basic Auth scheme invalid, so I opted for this instead. I couldn't think of a library method that does exactly what we want.

feelepxyz: Ah yeah could we split on `:` and escape the parts?

hmarr: Ah got it. I also wonder if we want to move the substitution to `basic_auth_details` as I think technically this could also happen for base64-encoded tokens (that contain @ signs) too.

jurre: Ah, yes good call! 👍

PR #2410, opened 2020-08-04
Always return a Version object for indirect gomodule dependencies

jurre:
When dealing with an indirect gomodules dependency, we were retuning a
string value that came from the Dependency object. We use this version
to compare with a Dependency object in
common/lib/dependabot/update_checkers/base.rb:216, which resulted in an
ArgumentError. We already return a Version object when the dependency is
top_level.

### DIFF
--- a/go_modules/lib/dependabot/go_modules/update_checker.rb
+++ b/go_modules/lib/dependabot/go_modules/update_checker.rb
@@ -17,6 +17,8 @@ module Dependabot
         # dependency to the go.mod file forcing the resolver to pick this
         # version (possibly as # indirect)
         unless dependency.top_level?
+          return unless dependency.version
+
           return version_class.new(dependency.version)
         end
 
--- a/go_modules/lib/dependabot/go_modules/update_checker.rb
+++ b/go_modules/lib/dependabot/go_modules/update_checker.rb
@@ -16,7 +16,9 @@ module Dependabot
         # To update indirect dependencies we'll need to promote the indirect
         # dependency to the go.mod file forcing the resolver to pick this
         # version (possibly as # indirect)
-        return dependency.version unless dependency.top_level?
+        unless dependency.top_level?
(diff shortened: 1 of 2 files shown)


go_modules/lib/dependabot/go_modules/update_checker.rb
```diff
@@ -16,7 +16,9 @@ def latest_resolvable_version
         # To update indirect dependencies we'll need to promote the indirect
         # dependency to the go.mod file forcing the resolver to pick this
         # version (possibly as # indirect)
-        return dependency.version unless dependency.top_level?
+        unless dependency.top_level?
+          return version_class.new(dependency.version)
```
feelepxyz, line 20: Think the version can be nil, do we need to handle this when initializing the version class?

thepwagner:
Maybe related to https://github.com/dependabot/dependabot-core/pull/2312 ? I think that's the same error from another ecosystem's `UpdateChecker`?

WDYT about combining those approaches? Maybe emit a warning so we can defer fixing checkers for these ~known issues?

jurre:
> Maybe related to #2312 ? I think that's the same error from another ecosystem's UpdateChecker?

Yes, it seems like it is 👍 

> WDYT about combining those approaches? Maybe emit a warning so we can defer fixing checkers for these ~known issues?

Great idea! Should we reopen your PR so we can add it there?

go_modules/lib/dependabot/go_modules/update_checker.rb
jurre: The version class doesn't support it, when can this be nil exactly? I can add test coverage for it and handle that case 👍

jurre: I've added a naive `return unless dependency.version`, since I'm not sure how this happens exactly, but figured better safe than sorry

PR #2415, opened 2020-08-04
Build docker with cache

feelepxyz:
Push a `latest` tag to the ci image and build ci from that cache.

[Existing build time](https://github.com/dependabot/dependabot-core/runs/945555472):
- fetch cache: 1m 30s
- core: 12s
- ci: 9m 23s

[New build time](https://github.com/dependabot/dependabot-core/pull/2415/checks?check_run_id=945981526)  (when everything is already cached):
- fetch cache: 1m 53s
- core: 13s
- ci: 1s 🎉 🙌

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,27 +20,27 @@ jobs:
           echo ::set-env name=VERSION::$(grep -Eo "[0-9]+\.[0-9]+\.[0-9]+" common/lib/dependabot/version.rb)
       - name: Pull Docker base images & warm Docker cache
         run: |
-          docker pull ubuntu:18.04
+          docker login docker.pkg.github.com -u x -p ${{ secrets.GITHUB_TOKEN }}
           docker pull "dependabot/dependabot-core:latest"
+          docker pull "docker.pkg.github.com/${{ github.repository }}/dependabot-core-ci:latest"
       - name: Build dependabot-core image
         run: |
           docker build \
             -t "dependabot/dependabot-core:latest" \
-            --cache-from ubuntu:18.04 \
             --cache-from "dependabot/dependabot-core:latest" \
             .
       - name: Build dependabot-core-ci image
         run: |
           rm .dockerignore
           docker build \
+            -t "docker.pkg.github.com/${{ github.repository }}/dependabot-core-ci:latest" \
             -t "docker.pkg.github.com/${{ github.repository }}/dependabot-core-ci:${{ github.sha }}" \
             -f Dockerfile.ci \
-            --cache-from ubuntu:18.04 \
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -21,30 +21,22 @@ jobs:
...
-            .
-      - name: Push image to packages
-        run: |
-          docker login docker.pkg.github.com -u x -p ${{ secrets.GITHUB_TOKEN }}
-          docker push "docker.pkg.github.com/${{ github.repository }}/dependabot-core-ci:${{ github.sha }}"
+        uses: whoan/docker-build-with-cache-action@v5
+        with:
+          username: x
+          password: "${{ secrets.GITHUB_TOKEN }}"
+          image_name: "${{ github.repository }}/dependabot-core-ci"
```
thepwagner, line 36: Concern: the [sample](https://github.com/whoan/docker-build-with-cache-action#github-registry) made `${{github. repository}}` seem out of place, but I guess it matches the previous?

Edit: it may be injected https://github.com/whoan/docker-build-with-cache-action/blob/27d4a67a76fb54d11f7d1225e0ff0f122c682b9f/entrypoint.sh#L42-L43 ?

```suggestion
          image_name: "dependabot-core-ci"
```

.github/workflows/ci.yml
```diff
@@ -21,30 +21,22 @@ jobs:
...
-            --cache-from "dependabot/dependabot-core:latest" \
-            .
-      - name: Push image to packages
-        run: |
-          docker login docker.pkg.github.com -u x -p ${{ secrets.GITHUB_TOKEN }}
-          docker push "docker.pkg.github.com/${{ github.repository }}/dependabot-core-ci:${{ github.sha }}"
+        uses: whoan/docker-build-with-cache-action@v5
+        with:
+          username: x
+          password: "${{ secrets.GITHUB_TOKEN }}"
```
thepwagner, line 35: :tinfoil: WDYT about a `read:packages write:packages` PAT instead of `GITHUB_TOKEN`?

`GITHUB_TOKEN` is powerful - https://docs.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token#permissions-for-the-github_token , particularly `contents:write`.

feelepxyz:
> Concerns are paranoia about adding a 3rd party action, and getting confused in its implementation (I still don't get how `image_name` works...)

Good shout! Doesn't seem like it adds that much with some funky stuff around stage images. Changed it to just push and use ci:latest as the cache and seems to give similar time savings 🙌

jurre:
Woah that's such a huge improvement, awesome work ✨

PR #2421, opened 2020-08-06
Inline buildkit cache when building images

feelepxyz:
Hopefully improving caching, it seems docker is currently too agressive
in it's cache usage so we're not re-building native helpers.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,8 +6,8 @@ on:
       - "wip/**"
       - "main"
   pull_request_target:
-    branches-ignore:
-      - "actions/**"
+    branches:
+      - "**"
 jobs:
   ci-image:
     name: Build CI image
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -6,8 +6,8 @@ on:
       - "wip/**"
       - "main"
   pull_request_target:
-    branches-ignore:
```
feelepxyz, line 9: Turns out branches for PR events are the target branches, e.g. `main` and `branches-ignore` basically doesn't make any sense here as it would match all PRs that don't target `/actions/*`

.github/workflows/ci.yml
```diff
@@ -24,18 +24,24 @@ jobs:
...
       - name: Build dependabot-core-ci image
+        env:
+          DOCKER_BUILDKIT: 1
         run: |
           rm .dockerignore
           docker build \
             -t "docker.pkg.github.com/${{ github.repository }}/dependabot-core-ci:latest" \
             -t "docker.pkg.github.com/${{ github.repository }}/dependabot-core-ci:${{ github.sha }}" \
             -f Dockerfile.ci \
+            --build-arg BUILDKIT_INLINE_CACHE=1
```
feelepxyz, line 44: Looks like this is the recommended way from docs https://docs.docker.com/engine/reference/commandline/build/#specifying-external-cache-sources

Not sure if it will improve things yet

.github/workflows/docker.yml
```diff
@@ -11,10 +11,18 @@ jobs:
...
+          DOCKER_BUILDKIT: 1
+        run: |
+          docker build \
+            -t "dependabot/dependabot-core:latest" \
+            --build-arg BUILDKIT_INLINE_CACHE=1
+            --cache-from "dependabot/dependabot-core:latest" \
+            .
+      - name: Push image to packages
+        run: |
+          docker push "dependabot/dependabot-core:latest"
```
feelepxyz, line 28: This seems to be what docker-build-with-cache-action basically does

PR #2422, opened 2020-08-06
Properly escape both username and password in auth url

jurre:
Both the username and password may contain characters that are not valid
in HTTP basic auth, we want to escape both, and will assume that the
first colon is what should be split on.


feelepxyz:
@jurre CI should be fixed now 😅

PR #2428, opened 2020-08-07
Fix: Collapse jobs

localheinz:
This PR

* [x] collapses the `rubocop` and `rspec` into a `helper` job

💁‍♂️ Currently the [CI workflow](https://github.com/dependabot/dependabot-core/blob/6b5a42e35ef400f63544ed8d26030af6bd1b3fba/.github/workflows/ci.yml) consists of 38 jobs. [According to a recent workflow run](https://github.com/dependabot/dependabot-core/actions/runs/198691974), this takes  32 minutes and 45 seconds.
Most of that time is taken up by

- building the CI images (~10 minutes, once)
- pulling the CI images (~1.45 minutes, 37 times)

If we collapse the jobs, we could reduce the time it takes to

- building the CI images (~10 minutes, once)
- pulling the CI images (~1.45 minutes, 19 times)

### Before

Screen Shot 2020-08-07 at 08 30 35

### After

Looks like we can't observe the result as we use the newly introduced `pull_request_target` event.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,8 +44,8 @@ jobs:
           docker push "docker.pkg.github.com/${{ github.repository }}/dependabot-core-ci:latest"
           docker push "docker.pkg.github.com/${{ github.repository }}/dependabot-core-ci:${{ github.sha }}"
 
-  rubocop:
-    name: Rubocop
+  helper:
+    name: Helper
     runs-on: ubuntu-latest
     needs: ci-image
     strategy:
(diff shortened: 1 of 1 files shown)


feelepxyz:
@localheinz nice one! Yeah agreed also makes it easier to see which ecosystem fails instead of scrolling a massive list. Was going to test on a branch but accidentally pushed a cherry-pick to main 🙈

localheinz:
Thank you, @feelepxyz!

PR #2431, opened 2020-08-07
actions docker.yaml: publish tag ontag

thepwagner:
This appends the `docker.yaml` workflow to:
- be triggered when a tag is pushed
- if triggered from a tag, push `dependabot/dependabot-core:$VERSION` after `dependabot/dependabot-core:latest`

For reference, the release process pushes a commit to the default branch and a tag in rapid succession: https://github.com/dependabot/dependabot-core/blob/main/bin/bump-version.rb#L50-L54

We _could_ have just pushed `:$VERSION` on every build, but this approach keeps tagged images ~immutable.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,8 +3,9 @@ on:
   push:
     branches:
     - main
+  create:
     tags:
-    - '*'
+      - v[0-9]+.[0-9]+.[0-9]+
 
 jobs:
   push-core-image:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,6 +3,8 @@ on:
   push:
     branches:
     - main
+    tags:
+    - '*'
 
 jobs:
   push-core-image:
@@ -23,7 +25,15 @@ jobs:
             --build-arg BUILDKIT_INLINE_CACHE=1 \
             --cache-from "dependabot/dependabot-core:latest" \
             .
-      - name: Push image to packages
+      - name: Push image to packages (latest)
+        if: '!contains(github.ref, "refs/tags")'
         run: |
           docker login -u ${{ secrets.DOCKER_USERNAME }} -p ${{ secrets.DOCKER_PASSWORD }}
           docker push "dependabot/dependabot-core:latest"
+      - name: Push image to packages (tagged)
+        if: contains(github.ref, "refs/tags")
+        run: |
+          docker login -u ${{ secrets.DOCKER_USERNAME }} -p ${{ secrets.DOCKER_PASSWORD }}
+          VERSION="$(grep -Eo "[0-9]+\.[0-9]+\.[0-9]+" common/lib/dependabot/version.rb)"
+          docker tag "dependabot/dependabot-core:latest" "dependabot/dependabot-core:$VERSION"
+          docker push "dependabot/dependabot-core:$VERSION"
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -23,7 +25,15 @@ jobs:
             --build-arg BUILDKIT_INLINE_CACHE=1 \
             --cache-from "dependabot/dependabot-core:latest" \
             .
-      - name: Push image to packages
+      - name: Push image to packages (latest)
+        if: '!contains(github.ref, "refs/tags")'
```
thepwagner, line 29: I'm claiming this as an independent invention, but https://github.com/search?q=%22if%3A+contains%28github.ref%2C+%22tags%22%29%22&type=Code makes it seem ~known.

Maybe `refs/tags` is overkill given the masses' wisdom...

.github/workflows/docker.yml
```diff
@@ -3,6 +3,8 @@ on:
   push:
     branches:
     - main
+    tags:
+    - '*'
```
feelepxyz, line 7: Should we do the same as gems.yml and only run when creating a tag?
 
```suggestion
  create:
    tags:
      - v[0-9]+.[0-9]+.[0-9]+
```
```

File: .github/workflows/docker.yml (hunk 1 above)
Comment (feelepxyz) [inline .github/workflows/docker.yml:29]: Any harm in always pushing to latest?

File: .github/workflows/docker.yml [hunk 3]
```diff
@@ -23,7 +26,15 @@ jobs:
...
             --cache-from "dependabot/dependabot-core:latest" \
             .
-      - name: Push image to packages
+      - name: Push image to packages (latest)
+        if: '!contains(github.ref, "refs/tags")'
         run: |
           docker login -u ${{ secrets.DOCKER_USERNAME }} -p ${{ secrets.DOCKER_PASSWORD }}
           docker push "dependabot/dependabot-core:latest"
+      - name: Push image to packages (tagged)
+        if: contains(github.ref, "refs/tags")
```
feelepxyz, line 35: Think this is a nice improvement from circle days where we would just keep pushing to the tagged version with any new commits to main

.github/workflows/docker.yml
```diff
@@ -23,7 +26,15 @@ jobs:
...
+        run: |
+          docker login -u ${{ secrets.DOCKER_USERNAME }} -p ${{ secrets.DOCKER_PASSWORD }}
+          VERSION="$(grep -Eo "[0-9]+\.[0-9]+\.[0-9]+" common/lib/dependabot/version.rb)"
+          docker tag "dependabot/dependabot-core:latest" "dependabot/dependabot-core:$VERSION"
+          docker push "dependabot/dependabot-core:$VERSION"
```
feelepxyz, line 40: Added `docker logout` here https://github.com/dependabot/dependabot-core/pull/2430/files for fork PRs :neckbeard: 

```suggestion
          docker push "dependabot/dependabot-core:$VERSION"
          docker logout
```

.github/workflows/docker.yml
thepwagner: No _harm_, it's just making the release process race over `:latest` by design:

Since the workflow will trigger twice for the push to main+tag creation, and those workflows will produce ~equivalent but not `===` images.
The goal was to remove ambiguity over which workflow "won", and reduce churn.

.github/workflows/docker.yml
thepwagner: 👍  It's weird to disconnect the docker image tag from the repo tag, when equivalent values exist:
I expect `dependabot/dependabot-core:0.118.1` to be `refs/tags/v0.118.1`, not `refs/tags/v0.118.2~1`.

I'd dig that for `dependabot/dependabot-core:0.118` and `dependabot/dependabot-core:0` maybe (but YAGNI nobody wants these).

.github/workflows/docker.yml
feelepxyz: > Since the workflow will trigger twice for the push to main+tag creation, and those workflows will produce ~equivalent but not `===` images.

Ah yes nice one, forgot about the double runs 👍 💯

PR #2432, opened 2020-08-07
docker.yaml syntax fix

thepwagner:
This fixes a regression introduced by #2431 that blocks _all_ docker publishing.

The YAML editor didn't mind, and Actions apparently doesn't validate syntax until the workflow is _triggered_  (e.g. when the previous PR was merged to main, it [didn't work](https://github.com/dependabot/dependabot-core/actions/runs/199385282) ).

To verify this syntax, I forced this YAML to be evaluated and verified the runner interprets the workflow as expected: https://github.com/dependabot/dependabot-core/actions/runs/199413936

# Related
- Continues #2431

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -2,7 +2,7 @@ name: Push docker images
 on:
   push:
     branches:
-    - main
+    - '*'
   create:
     tags:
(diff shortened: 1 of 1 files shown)


feelepxyz:
Fixed the rubocop on main, snuck in because ci is broken for PRs 😕 would just merge this 👍

PR #2435, opened 2020-08-07
Add docker caching and fix ci for forks

feelepxyz:
We should be able to restore caching now and added a new branch-ref tag so the same branch should get quicker-rebuilds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,49 +9,9 @@ on:
     branches:
       - "**"
 jobs:
-  ci-image:
-    name: Build CI image
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v2
-      - name: Prepare environment variables
-        run: |
-          echo ::set-env name=VERSION::$(grep -Eo "[0-9]+\.[0-9]+\.[0-9]+" common/lib/dependabot/version.rb)
-      - name: Pull Docker base images & warm Docker cache
-        run: |
-          docker login docker.pkg.github.com -u x -p ${{ secrets.GITHUB_TOKEN }}
-          docker pull "dependabot/dependabot-core:latest"
-          docker pull "docker.pkg.github.com/${{ github.repository }}/dependabot-core-ci:latest"
-          docker logout
-      - name: Build dependabot-core image
-        env:
-          DOCKER_BUILDKIT: 1
-        run: |
-          docker build \
-            -t "dependabot/dependabot-core:latest" \
-            --build-arg BUILDKIT_INLINE_CACHE=1 \
-            .
-      - name: Build dependabot-core-ci image
-        env:
-          DOCKER_BUILDKIT: 1
-        run: |
-          rm .dockerignore
-          docker build \
-            -t "docker.pkg.github.com/${{ github.repository }}/dependabot-core-ci:latest" \
-            -t "docker.pkg.github.com/${{ github.repository }}/dependabot-core-ci:${{ github.sha }}" \
-            -f Dockerfile.ci \
-            --build-arg BUILDKIT_INLINE_CACHE=1 \
-            .
-      - name: Push image to packages
-        run: |
-          docker push "docker.pkg.github.com/${{ github.repository }}/dependabot-core-ci:latest"
-          docker push "docker.pkg.github.com/${{ github.repository }}/dependabot-core-ci:${{ github.sha }}"
-
   ci:
     name: CI
     runs-on: ubuntu-latest
-    needs: ci-image
     strategy:
       matrix:
         suite:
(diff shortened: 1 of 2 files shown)


.github/workflows/docker.yml
```diff
@@ -24,17 +24,16 @@ jobs:
...
             .
+      - name: Log in to the Docker registry
+        run: |
+          echo "${{ secrets.DOCKER_PASSWORD }}" | docker login -u ${{ secrets.DOCKER_USERNAME }} --password-stdin
       - name: Push image to packages (latest)
         if: "!contains(github.ref, 'refs/tags')"
         run: |
-          docker login -u ${{ secrets.DOCKER_USERNAME }} -p ${{ secrets.DOCKER_PASSWORD }}
           docker push "dependabot/dependabot-core:latest"
-          docker logout
```
feelepxyz, line 32: Got rid of these as we're not running on forks with creds

.github/workflows/ci.yml
```diff
@@ -73,32 +33,76 @@ jobs:
...
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Prepare environment variables
         run: |
-          docker login docker.pkg.github.com -u x -p ${{ secrets.GITHUB_TOKEN }}
-          docker pull "docker.pkg.github.com/${{ github.repository }}/dependabot-core-ci:${{ github.sha }}"
-          docker logout
-      - name: Run Rubocop linting
+          BRANCH_REF="$(echo '${{ github.ref }}' | sed -E 's/[^A-Za-z0-9]+/-/g')"
+          echo ::set-env name=BRANCH_REF::$(BRANCH_REF)
```
thepwagner, line 46:
```suggestion
          echo ::set-env name=BRANCH_REF::$(BRANCH_REF)
          echo ::set-env name=BASE_IMAGE::ubuntu:18.04
```

Idea: to de-dupe the later usages.
Maybe even `$(grep FROM Dockerfile | sed '...')`? Meh.

.github/workflows/ci.yml
```diff
@@ -73,32 +33,76 @@ jobs:
...
+      - name: Log in to the Docker registry
         run: |
-          docker run --rm docker.pkg.github.com/${{ github.repository }}/dependabot-core-ci:${{ github.sha }} bash -c "cd /home/dependabot/dependabot-core/${{ matrix.suite }} && bundle exec rubocop . -c ../.rubocop.yml"
-      - name: Run ${{ matrix.suite }} tests with rspec
+          if [ -n "$DOCKER_USERNAME" ] && [ -n "$DOCKER_PASSWORD" ]; then
```
thepwagner, line 49: This could maybe be slicker as an Actions-native conditional, but I'm not sure of the syntax...

https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions says `secrets` is in scope - maybe `if: ${{secrets. DOCKER_USERNAME && secrets. DOCKER_PASSWORD}}` ?

The only hit I could find is: https://github.com/Kegbot/kegbot-server/blob/54b02880e8b440168110985230a60b7390e10991/.github/workflows/docker.yml#L41 :

`if: ${{ '${{ secrets.DOCKER_HUB_PASSWORD }}' != '' && '${{ secrets.DOCKER_HUB_USERNAME }}' != '' }}` .

thepwagner: Idea: drop a breadcrumb like `echo ::set-env name=DOCKER_LOGGED_IN::true` for later steps to use, instead of duplicating this.

.github/workflows/ci.yml
```diff
@@ -73,32 +33,76 @@ jobs:
...
+      - name: Build dependabot-core-ci image
+        run: |
+          rm .dockerignore
+          docker build \
+            -t "dependabot/dependabot-core-ci:latest" \
+            -t "dependabot/dependabot-core-ci:ci--$BRANCH_REF" \
+            -f Dockerfile.ci \
+            --build-arg BUILDKIT_INLINE_CACHE=1 \
+            --cache-from ubuntu:18.04 \
+            --cache-from "dependabot/dependabot-core:latest" \
```
thepwagner, line 87:
```suggestion
            --cache-from "dependabot/dependabot-core-ci:latest" \
```

I think?  I'm extrapolating from :64 and :69.

.github/workflows/ci.yml
```diff
@@ -73,32 +33,76 @@ jobs:
...
+        run: |
+          if [ -n "$DOCKER_USERNAME" ] && [ -n "$DOCKER_PASSWORD" ]; then
+            docker push "dependabot/dependabot-core-ci:latest"
+            docker push "dependabot/dependabot-core-ci:ci--$BRANCH_REF"
+          else
+            echo "No Docker credentials, skipping push"
+          fi
+      - name: Run Rubocop linting
+        run: |
+          docker run --rm dependabot/dependabot-core-ci bash -c "cd /home/dependabot/dependabot-core/${{ matrix.suite }} && bundle exec rubocop . -c ../.rubocop.yml"
```
thepwagner, line 100: Aside: I was surprised to see this is a `matrix` build (e.g. that we're building+pushing these images so ~concurrently). 🏎️ .

.github/workflows/ci.yml
feelepxyz: > https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions says `secrets` is in scope - maybe `if: ${{secrets. DOCKER_USERNAME && secrets. DOCKER_PASSWORD}}` ?

Yeah tried that but got a syntax error saying it wasn't available 😕

.github/workflows/ci.yml
feelepxyz: Ah yeah could push images only when on `common`

feelepxyz: or I guess docker will discard most the pushes as layer already exists?

feelepxyz:
@thepwagner thanks for the feedback, pushed some cleanups 👍

PR #2452, opened 2020-08-14
Use one sentence per line for README

HonkingGoose:
**This PR changes:**

- All sentences are put on their own line.

**Benefits of change**

One sentence per line helps to ensure consistency when new lines are added.
Makes version control easier, because the changes are always on a line level, instead of spread around multiple lines.
Nudge writers to write shorter/simpler sentences to prevent excessive line length.

**Drawbacks of change**
It's a harder to see which lines belong to which number on a numbered list. (See lines 10 trough 13 in this pull request for an example of this).

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,14 +4,19 @@
 
 # Dependabot
 
-Welcome to the public home of Dependabot. This repository serves 2 purposes:
+Welcome to the public home of Dependabot.
+This repository serves 2 purposes:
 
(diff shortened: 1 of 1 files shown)


jurre:
Hey, I appreciate the effort to make the readme easier to read, but I think the downside of not being able to git blame on those lines is not worth it, so I'm going to close this out.

💛

PR #2459, opened 2020-08-19
Enable PR creator to handle deleted and binary files

jurre:
In order to delete files, we pass a `sha: nil` to the API, and to upload
binary files we create a blob first and pass it's sha to the tree.


feelepxyz (approved):
Ace! LGTM 💯

PR #2461, opened 2020-08-19
Add repo_contents_path attribute

feelepxyz:
Adding a unused attribute to be able to test multiple versions of core
from the updater.


feelepxyz:
Not sure this is needed, looking into bundling a unreleased core version in the updater

feelepxyz:
@jurre @thepwagner thoughts on adding this? If we release core with these attributes we could release a version of the updater that has the required changes for vendoring: https://github.com/dependabot/dependabot-updater/pull/1004 (we'll then just need to bump core in updater once it's ready)

jurre (approved):

PR #2472, opened 2020-08-23
Update to poetry to 1.0.10

tomzx:
A change in poetry 1.0.10 has changed how the `git clone` operation is performed. It now includes the `--recurse-submodules` option which wasn't there previously. This PR should superseed #2407.


jurre:
This is fantastic, thanks @tomzx

PR #2479, opened 2020-08-25
bundler vendoring: git sources

thepwagner:
This extends https://github.com/dependabot/dependabot-core/pull/2476 with a test case demonstrating updating a vendored git dependency.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,6 +7,7 @@ AllCops:
     - '*/tmp/**/*'
     - 'tmp/**/*'
     - '*/helpers/**/*'
+    - '*/spec/fixtures/**/*'
 
 Layout/DotPosition:
   EnforcedStyle: trailing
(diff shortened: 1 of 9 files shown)


bundler/spec/dependabot/bundler/file_updater_spec.rb
```diff
@@ -1607,6 +1607,73 @@
           expect(vendor_files).to include("statesman-7.2.0.gem")
         end
       end
+
+      context "with a git dependency" do
+        let(:repo_contents_path) { build_tmp_repo("vendored_git") }
+        let(:gemfile_fixture_name) { "git_source_with_version" }
+        let(:lockfile_fixture_name) { "git_source_with_version.lock" }
```
jurre, line 1614: In #2476 I've since restructured this to use a Gemfile and Gemfile.lock in the project path, would be great to do the same here

thepwagner: Awesomesauce! That disconnect [threw me for a loop](https://github.com/dependabot/dependabot-core/pull/2476#discussion_r476693529) until I found the indirection.

.rubocop.yml
```diff
@@ -7,6 +7,7 @@ AllCops:
     - '*/tmp/**/*'
     - 'tmp/**/*'
     - '*/helpers/**/*'
+    - '*/spec/fixtures/**/*'
```
jurre, line 10: 👌

jurre (approved):
Surprisingly simple 🎉

PR #2484, opened 2020-08-26
bundler vendoring: detect non-binary extensions

thepwagner:
This extends #2476 to avoid base64 encoding files that are probably not binary.

This is important because of how the `PullRequestCreator::GitHub` [creates PRs](https://github.com/dependabot/dependabot-core/blob/f9ee541691b91f76308a972bb535d40443bce33f/common/lib/dependabot/pull_request_creator/github.rb#L173-L182). Since binary files are created as blobs, a PR that updates many files (e.g. a vendored git dependency #2479 ) would have a slow sequential upload process.

We'd prefer to batch content as much as possible within the `POST /repos/:owner/:repo/git/trees` call.
The trade-off of this approach is the growing size of that API call, I couldn't find any documented limits just some [reports](https://github.community/t/creating-new-tree-with-git-data-api-returns-502/1320/5) of potential problems with >thousands of files.

I didn't see anything with this capability in scope, so ~added https://github.com/minad/mimemagic to avoid reinventing~ invented something.

### DIFF
--- a/bundler/lib/dependabot/bundler/file_updater.rb
+++ b/bundler/lib/dependabot/bundler/file_updater.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require "mimemagic"
-
 require "dependabot/file_updaters"
 require "dependabot/file_updaters/base"
 
@@ -108,11 +106,37 @@ module Dependabot
         end
       end
 
-      def binary_file?(path)
-        magic = MimeMagic.by_path(path)
-        return true unless magic
+      # notable filenames without a reliable extension:
+      TEXT_FILE_NAMES = [
+        "Gemfile",
+        "Gemfile.lock",
+        ".bundlecache",
+        ".gitignore"
+      ].freeze
 
-        !magic.text?
+      TEXT_FILE_EXTS = [
+        # code
+        ".rb",
+        ".erb",
+        ".gemspec",
+        ".js",
+        ".html",
+        # config
+        ".json",
+        ".xml",
+        ".toml",
+        ".yaml",
+        ".yml",
+        # docs
+        ".md",
+        ".txt"
+      ].freeze
+
+      def binary_file?(path)
+        return false if TEXT_FILE_NAMES.include?(File.basename(path))
+        return false if TEXT_FILE_EXTS.include?(File.extname(path))
+
+        true
       end
 
       def check_required_files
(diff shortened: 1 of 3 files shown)


jurre (approved):
Changes look 💯. Unsure about the dependency but keeping lists of extensions to date for every language that we support also doesn't seem fun. Should we move it to dependabot common gemspec perhaps?

bundler/lib/dependabot/bundler/file_updater.rb
```diff
@@ -102,6 +108,13 @@ def updated_vendor_cache_files(base_directory:)
         end
       end
 
+      def binary_file?(path)
+        magic = MimeMagic.by_path(path)
```
jurre, line 112: Unfortunately `MimeMagic` seems to think that `erb` files are binary for example, it only seems to know about `.rb` files explicitly, I think we may be better off keeping our own list in that case?

thepwagner: > I think we may be better off keeping our own list

Let's do it!

I started down that path - it's ~:100: if you match filenames (e.g. `Gemfile`) then a secondary pass on extensions.

I force-pushed it into oblivion while writing the PR description as:
* I didn't want to be on the hook to maintain it
* it felt a little not-invented-here

The result was the first google hit for a library I could delegate this too - thanks for vetting the implementation! ❤️ 

--- 

I pulled `find . -type f | grep -v .git | sed -e 'sF.*/FFg' -e 's/.*\.//g' | sort | uniq -c | sort -nr | head -n 25` from this here ruby monorepo:
```
 561 rb
 278 json
 156 lock - nope, these are probably Gemfile.lock
 135 xml
  67 txt
  35 toml - nope
  32 html
  31 md
  27 gradle - nope but moot for bundler ecosystem
  23 in - nope
  22 Gemfile - nope
  21 py
  19 js
  17 gemspec - nope, these are literally everywhere
  11 gem
--- i stopped checking here
  11 csproj
   9 config
   8 pom
   8 mod
   8 go
   7 build
   6 yml
   6 props
   6 php
   5 sum
```

thepwagner: https://github.com/dependabot/dependabot-core/pull/2484/commits/9a926d289dff1d6e0f35cdb0dc5b64131403c8a4

jurre (approved):
👍 we might eventually extract some of the binary file detection, but maybe YAGNI so leaving it in the Bundler::FileUpdater seems sane to me

Issue #2470, opened 2020-08-21
Update GitHub Actions to latest release commit hash


rosslh:
I am using Dependabot to update my GitHub Actions which is a very useful feature! There is one issue though: I want to specify the action version using commit hashes - not tag names. This is because tags can be deleted and republished with malicious code, whereas a commit at a specific hash is immutable. [More detail on this issue here](https://julienrenaux.fr/2019/12/20/github-actions-security-risk/).

Right now when updating my actions, Dependabot replaces the commit hash with the latest tag name (e.g. `actions/example@6c2f566` becomes `actions/example@v1.0.0`). It would be great if it could find the commit hash of the latest release and use that instead.

reitermarkus:
Might be as easy as changing

https://github.com/dependabot/dependabot-core/blob/main/github_actions/lib/dependabot/github_actions/update_checker.rb#L73

to return the `:commit_sha`, i.e. never “downgrading” a hash to a less secure tag.

Reading the code, I actually think these are swapped currently: Tag will be updated to a hash and hash will be updated to a tag.

jauderho:
FWIW, this still seems to be a problem. 

For example, see this pull request generated by dependabot: https://github.com/jauderho/psfiles/pull/1

It's trying update a hash to a version tag.

PR #2492, opened 2020-09-01
Don't downgrade a pinned commit to a tag.

reitermarkus:
If an action is pinned to a commit, this is most likely as a security precaution, so always update pinned commits to the commit hash corresponding to the latest tag.

Fixes https://github.com/dependabot/dependabot-core/issues/2470.

### DIFF
--- a/github_actions/lib/dependabot/github_actions/update_checker.rb
+++ b/github_actions/lib/dependabot/github_actions/update_checker.rb
@@ -90,11 +90,12 @@ module Dependabot
           return dependency_source_details.merge(ref: new_tag.fetch(:tag))
         end
 
-        # Update the git tag if updating a pinned commit
+        # Update the git commit if updating a pinned commit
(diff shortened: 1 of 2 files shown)


jurre:
This seems right to me, thanks! cc @infin8x double checking that you agree this is the desired behavior

infin8x:
Makes sense to me!

jurre:
I'm fixing those failing cargo specs on a different branch those are unrelated. Not sure why most of the suites are getting killed at 5m on here, but if the actions suite passes I'm going to merge this, thanks so much @reitermarkus

rosslh:
@jurre Just wondering, has this reached production? Because I'm still having actions updated from a commit hash to a tag.

jurre:
> It's just the PRs titles that still have the tag name which confused me. Sorry about that!

@rosslh was that because the old PR was still open, or did we create it after the fix went live with the wrong title?

PR #2502, opened 2020-09-04
Add OAuth support in azure client

milind009:
**Issue**
Currently, `azure_client` supports only `basic authentication` using `username` and `password`. For our service, which we have built on top of dependabot-core, we require support for `OAuth` tokens in `azure_client`. We found this as something which we could contribute back to dependabot-core hence creating this PR.

**Solution**
 For enabling authentication using tokens, it will be expecting a separate field named `'token'` in the `credentials` hash instead of `username` and `password`. Using the token, it will decide on the `authorization headers` for the `Excon` request in the same manner as it is done for connecting to registries in dependabot-core. For e.g 
https://github.com/dependabot/dependabot-core/blob/main/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/registry_finder.rb#L73-L85

If the user provides `username` and `password`, it will follow the current implementation as is. For the case of `'token' field`, it will set the `username` and `password` in the request to `nil` and will send the token in the `headers` with the appropriate `authorization`.

**Tests**
Added tests to verify the usage of authorization headers for get function in azure_client.

@jurre Let me know your thoughts on this PR

### DIFF
--- a/common/spec/dependabot/clients/azure_spec.rb
+++ b/common/spec/dependabot/clients/azure_spec.rb
@@ -3,8 +3,7 @@
 require "spec_helper"
 require "dependabot/clients/azure"
 
-RSpec.shared_examples "#get using auth headers" do
-|credential|
+RSpec.shared_examples "#get using auth headers" do |credential|
   before do
     stub_request(:get, base_url).
       with(headers: credential["headers"]).
--- a/common/spec/dependabot/clients/azure_spec.rb
+++ b/common/spec/dependabot/clients/azure_spec.rb
@@ -3,6 +3,22 @@
 require "spec_helper"
 require "dependabot/clients/azure"
 
+RSpec.shared_examples "Successfully completes get request with auth headers" do |credential|
+  before do
+    stub_request(:get, base_url).
+      with(headers: credential["headers"]).
+      to_return(status: 200, body: '{"result": "Success"}')
+  end
+
+
+  it "Using #{credential["token_type"]} token in credentials" do
+
+    client = described_class.for_source(source: source, credentials: credential["credentials"])
+    response = JSON.parse(client.get(base_url).body)
+    expect(response["result"]).to eq("Success")
+  end
+end
+
 RSpec.describe Dependabot::Clients::Azure do
   let(:username) { "username" }
   let(:password) { "password" }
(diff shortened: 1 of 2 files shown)


common/spec/dependabot/clients/azure_spec.rb
```diff
@@ -3,6 +3,22 @@
...
+RSpec.shared_examples "Successfully completes get request with auth headers" do |credential|
+  before do
+    stub_request(:get, base_url).
+      with(headers: credential["headers"]).
+      to_return(status: 200, body: '{"result": "Success"}')
+  end
+
+
+  it "Using #{credential["token_type"]} token in credentials" do
+
```
jurre, line 15: I think our linter is failing on this, if we get rid of the blank line it should pass 👍

common/spec/dependabot/clients/azure_spec.rb
```diff
@@ -109,4 +125,17 @@
...
     end
   end
+
+  describe "#get" do
+    context "Using auth headers" do
+        token = ":test_token"
+        encoded_token = Base64.encode64(":test_token").delete("\n")
+        bearer_token = "test_token"
+
+        include_examples 'Successfully completes get request with auth headers', {"token_type" => "basic non encoded", "credentials" => [{"type" => "git_source", "host" => "dev.azure.com", "token" => token}], "headers" => { "Authorization" => "Basic #{encoded_token}"}}
```
jurre, line 135: I think our linter is failing on the line-length here, there's a hard limit at 80 characters. I'd split up the hash over multiple lines and/or assign them to some variables outside of the example (using `let` seems most idiomatic in our codebase).

jurre (approved):
Code looks good, thanks for the contribution 👍 

There's a few linter failures on this branch that should be pretty easy to resolve. Once those are all green, I'll merge and release this. Lmk if I can help.

jurre:
Thanks @milind009, fyi, if you want to run the linter locally you can do this:

```
# from within the `common` directory
bundle exec rubocop . -c ../.rubocop.yml -a
```

The `-a` flag will `autofix` some cases for you

milind009:
> Thanks @milind009, fyi, if you want to run the linter locally you can do this:
> 
> ```
> # from within the `common` directory
> bundle exec rubocop . -c ../.rubocop.yml -a
> ```
> 
> The `-a` flag will `autofix` some cases for you

Thanks! I was able to run locally. Thanks for pointing out the locations that needed to be changed!

milind009:
@jurre is there anything else you need from my side or we can merge this PR as is?

PR #2509, opened 2020-09-07
Force encode nuspec files to utf-8 for regex matching

jurre:
Windows has a history of using UTF-16, recent versions of windows
default to UTF-8, but we sometimes encounter files that are encoded as
UTF-16. When trying to regex match on such a string, we get an error:

```
incompatible encoding regexp match (US-ASCII regexp with UTF-16 string)
```

By enforcing the encoding we should be able to prevent this.


feelepxyz (approved):
👍

Issue #1977, opened 2020-05-18
Python pip-compile "-e file:." requirement causes Dependabot to fail


davidism:
**Package manager/ecosystem**

python, pip-compile

**Manifest contents prior to update**

https://github.com/pallets/jinja/blob/master/requirements/dev.txt
<details>

```
#
# This file is autogenerated by pip-compile
# To update, run:
#
#    pip-compile requirements/dev.in
#
-e file:.                 # via -r requirements/dev.in, sphinx
alabaster==0.7.12         # via sphinx
appdirs==1.4.4            # via virtualenv
attrs==19.3.0             # via pytest
babel==2.8.0              # via sphinx
certifi==2020.4.5.1       # via requests
cfgv==3.1.0               # via pre-commit
chardet==3.0.4            # via requests
click==7.1.2              # via pip-tools
distlib==0.3.0            # via virtualenv
docutils==0.16            # via sphinx
filelock==3.0.12          # via tox, virtualenv
identify==1.4.15          # via pre-commit
idna==2.9                 # via requests
imagesize==1.2.0          # via sphinx
more-itertools==8.3.0     # via pytest
nodeenv==1.3.5            # via pre-commit
packaging==20.3           # via pallets-sphinx-themes, pytest, sphinx, tox
pallets-sphinx-themes==1.2.3  # via -r requirements/docs.in
pip-tools==5.1.2          # via -r requirements/dev.in
pluggy==0.13.1            # via pytest, tox
pre-commit==2.4.0         # via -r requirements/dev.in
py==1.8.1                 # via pytest, tox
pygments==2.6.1           # via sphinx
pyparsing==2.4.7          # via packaging
pytest==5.4.2             # via -r requirements/tests.in
pytz==2020.1              # via babel
pyyaml==5.3.1             # via pre-commit
requests==2.23.0          # via sphinx
six==1.14.0               # via packaging, pip-tools, tox, virtualenv
snowballstemmer==2.0.0    # via sphinx
sphinx-issues==1.2.0      # via -r requirements/docs.in
sphinx==2.4.4             # via -r requirements/docs.in, pallets-sphinx-themes, sphinx-issues, sphinxcontrib-log-cabinet
sphinxcontrib-applehelp==1.0.2  # via sphinx
sphinxcontrib-devhelp==1.0.2  # via sphinx
sphinxcontrib-htmlhelp==1.0.3  # via sphinx
sphinxcontrib-jsmath==1.0.1  # via sphinx
[...]

PR #2519, opened 2020-09-10
Fix python path dependencies with file

lfdebrux:
This should fix https://github.com/dependabot/dependabot-core/issues/1977

This PR updates the regex in Dependabot::Python::FileFetcher.parse_path_setup_paths to ignore if there is a `file:` at the beginning of the requirement path.


jurre (approved):
This makes sense to me, thanks for the contribution! 👍 (will merge this in once CI passes)

Mogost:
Is this fix already released?

PR #2523, opened 2020-09-11
Hex deep upgrade

thepwagner:
Previously when an update affected an indirect dependency, Dependabot could drop the `[:mix]` block from the lockfile.

This PR introduces a failing test case for this scenario, followed by resolution: fetching all dependencies (not just those targetted by the update) before reading the lockfile.


hex/helpers/lib/do_update.exs
```diff
@@ -31,6 +31,22 @@ dependency = String.to_atom(dependency_name)
 {dependency_lock, rest_lock} = Map.split(Mix.Dep.Lock.read(), [dependency])
 Mix.Dep.Fetcher.by_name([dependency_name], dependency_lock, rest_lock, [])
 
+System.cmd(
+  "mix",
+  [
+    "deps.get",
+    "--no-compile",
+    "--no-elixir-version-check",
```
thepwagner, line 39: Xref https://github.com/dependabot/dependabot-core/blob/main/hex/helpers/lib/run.exs#L54-L57 for the source.

I didn't think `--no-deps-check` was relevant as the intent is to... check the deps.

I thought `--no-start` may only be relevant to `mix run...`?

hex/spec/fixtures/lockfiles/minor_version
```diff
@@ -1,5 +1,7 @@
-%{"mime": {:hex, :mime, "1.2.0", "78adaa84832b3680de06f88f0997e3ead3b451a440d183d688085be2d709b534", [], [], "hexpm"},
-  "phoenix": {:hex, :phoenix, "1.2.5", "dbc45a5f8fb522aaa815b003f2d5598bc45e23102ae3e265768848ab23eafcb7", [], [{:cowboy, "~> 1.0", [hex: :cowboy, repo: "hexpm", optional: true]}, {:phoenix_pubsub, "~> 1.0", [hex: :phoenix_pubsub, repo: "hexpm", optional: false]}, {:plug, "~> 1.3.3 or ~> 1.2.4 or ~> 1.1.8 or ~> 1.0.5", [hex: :plug, repo: "hexpm", optional: false]}, {:poison, "~> 1.5 or ~> 2.0", [hex: :poison, repo: "hexpm", optional: false]}], "hexpm"},
-  "phoenix_pubsub": {:hex, :phoenix_pubsub, "1.0.2", "bfa7fd52788b5eaa09cb51ff9fcad1d9edfeb68251add458523f839392f034c1", [], [], "hexpm"},
-  "plug": {:hex, :plug, "1.3.5", "7503bfcd7091df2a9761ef8cecea666d1f2cc454cbbaf0afa0b6e259203b7031", [], [{:cowboy, "~> 1.0.1 or ~> 1.1", [hex: :cowboy, repo: "hexpm", optional: true]}, {:mime, "~> 1.0", [hex: :mime, repo: "hexpm", optional: false]}], "hexpm"},
-  "poison": {:hex, :poison, "2.0.1", "81248a36d1b602b17ea6556bfa8952492091f01af05173de11f8b297e2bbf088", [], [], "hexpm"}}
+%{
```
thepwagner, line 1: This is using `mix deps.get` with `Mix 1.10.4 (compiled with Erlang/OTP 23)` to reformat.

I'd done this with the hope that `[:mix]` would be dropped by the existing test case.
It didn't. We could revert this file but it's just prettier? 🥔 🥔 .

hex/helpers/lib/do_update.exs
jurre: This makes sense to me 👍

jurre (approved):
I don't fully understand _why_ fetching the dependencies resolves the lockfile issues, I guess they need to be there for mix to know if they use rebar etc? The added test makes me feel confident it works though so 👍

feelepxyz (approved):
Ace test case 👌 @thepwagner I think you are now the in-house hex expert 😀  :shipit:

PR #2533, opened 2020-09-18
Allow Azure client to set linked work item

JamieMagee:
This PR adds the ability to link work items when create a pull request in Azure DevOps.

The API is pretty forgiving, and allows the JSON `"workItemRefs": [{"id": null}]`

API docs: https://docs.microsoft.com/en-us/rest/api/azure/devops/git/pull%20requests/create?view=azure-devops-rest-5.0

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -86,7 +86,7 @@ Metrics/CyclomaticComplexity:
 
 Metrics/PerceivedComplexity:
   Max: 10 # default: 8
-  
+
 Metrics/AbcSize:
   Max: 35
 
@@ -101,7 +101,6 @@ Metrics/BlockLength:
     - '*/dependabot-*.gemspec'
 
 Metrics/ParameterLists:
-  Max: 6
   CountKeywordArgs: false
 
 Naming/FileName:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -101,6 +101,7 @@ Metrics/BlockLength:
     - '*/dependabot-*.gemspec'
 
 Metrics/ParameterLists:
+  Max: 6
   CountKeywordArgs: false
 
 Naming/FileName:
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -155,7 +155,7 @@ module Dependabot
 
       # rubocop:disable Metrics/ParameterLists
       def create_pull_request(pr_name, source_branch, target_branch,
-                              pr_description, labels, work_item)
+                              pr_description, labels, work_item = nil)
         # Azure DevOps only support descriptions up to 4000 characters
         # https://developercommunity.visualstudio.com/content/problem/608770/remove-4000-character-limit-on-pull-request-descri.html
         azure_max_length = 3999
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -154,7 +154,7 @@ module Dependabot
       end
 
       def create_pull_request(pr_name, source_branch, target_branch,
-                              pr_description, labels)
+                              pr_description, labels, work_item)
         # Azure DevOps only support descriptions up to 4000 characters
         # https://developercommunity.visualstudio.com/content/problem/608770/remove-4000-character-limit-on-pull-request-descri.html
         azure_max_length = 3999
@@ -169,7 +169,8 @@ module Dependabot
           targetRefName: "refs/heads/" + target_branch,
           title: pr_name,
           description: pr_description,
-          labels: labels.map { |label| { name: label } }
+          labels: labels.map { |label| { name: label } },
+          workItemRefs: [{id: work_item}]
         }
 
         post(source.api_endpoint +
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -153,6 +153,7 @@ module Dependabot
           "/pushes?api-version=5.0", content.to_json)
       end
 
+      # rubocop:disable Metrics/ParameterLists
       def create_pull_request(pr_name, source_branch, target_branch,
                               pr_description, labels, work_item)
         # Azure DevOps only support descriptions up to 4000 characters
@@ -163,6 +164,7 @@ module Dependabot
           truncate_length = azure_max_length - truncated_msg.length
           pr_description = pr_description[0..truncate_length] + truncated_msg
         end
+        # rubocop:enable Metrics/ParameterLists
 
         content = {
           sourceRefName: "refs/heads/" + source_branch,
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -170,7 +170,7 @@ module Dependabot
           title: pr_name,
           description: pr_description,
           labels: labels.map { |label| { name: label } },
-          workItemRefs: [{id: work_item}]
+          workItemRefs: [{ id: work_item }]
         }
 
         post(source.api_endpoint +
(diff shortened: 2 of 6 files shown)


JamieMagee:
Hi @jurre, another PR for you to review :smile:

.rubocop.yml
```diff
@@ -101,6 +101,7 @@ Metrics/BlockLength:
     - '*/dependabot-*.gemspec'
 
 Metrics/ParameterLists:
+  Max: 6
```
jurre, line 104: I'm not sure that we want to globally increase this, could we add inline exceptions for the methods that need it? Globally relaxing such rules is a bit of a slippery slope IMO, and we're already not in fantastic shape regarding linting in this project.

common/lib/dependabot/pull_request_creator.rb
```diff
@@ -33,7 +33,8 @@ def initialize(source:, base_commit:, dependencies:, files:, credentials:,
                    branch_name_separator: "/", branch_name_prefix: "dependabot",
                    label_language: false, automerge_candidate: false,
                    github_redirection_service: "github-redirect.dependabot.com",
-                   custom_headers: nil, require_up_to_date_base: false)
+                   custom_headers: nil, require_up_to_date_base: false,
+                   work_item: nil)
```
jurre, line 37: I wonder if it makes sense to start splitting up these PR creators based on the provider, there's more and more conditional logic / arguments, increasing the complexity and making it harder to add features and test things.

Not something that I'd expect in this PR, but more food for thought for ourselves cc @brrygrdn @feelepxyz

feelepxyz: Ah yeah would be ace to think about splitting these out. `work_item` here is slightly confusing as it's only for azure. Or could we add a generic `provider_metadata` hash that we pass to all pull request creators?

JamieMagee: @feelepxyz Totally agree with the idea. However, is this something you would like to see as part of this PR, or as a separate PR? It feels like a bit of a larger refactoring task in my opinion.

.rubocop.yml
JamieMagee: I do agree. I think this could be solved with the `provider_metadata` suggestion from @feelepxyz below.

common/lib/dependabot/pull_request_creator.rb
feelepxyz: @JamieMagee yeah lets leave a larger refactor out of this PR but thoughts on adding a new `provider_metadata` hash that gets passed in the Azure PR creator and then extract the `work_item` key?

JamieMagee: @feelepxyz I pushed an update that adds a `provider_metadata` hash currently just targeting the Azure specific properties, but this can easily be expanded to the other providers.

I think I can iterate on this a bit more. I think I'll pass `provider_metadata` directly through to the Azure client.

jurre: Maybe I misunderstood, but what I was thinking, was that it would be better to to pass the Azure specific `work_item` to `Dependabot::PullRequestCreator`, but passing it to the Azure client explicitly is fine.

So my thinking was that we change `Dependabot::PullRequestCreator#initialize` to:

```ruby
    def initialize(source:, base_commit:, dependencies:, files:, credentials:,
                   branch_name_separator: "/", branch_name_prefix: "dependabot",
                   label_language: false, automerge_candidate: false,
                   github_redirection_service: "github-redirect.dependabot.com",
                   custom_headers: nil, require_up_to_date_base: false)
                   custom_headers: nil, require_up_to_date_base: false,
                   provider_metadata: {})
```

We'd then pass the `work_item` in that metadata:

```ruby
Dependabot::PullRequestCreator.new(
# ...
provider_metadata: { work_item: "my-work-item" }
)
```

And then when creating the Azure client we pull it out:

```ruby
    Azure.new(
        source: source,
        branch_name: branch_namer.new_branch_name,
        base_commit: base_commit,
        credentials: credentials,
        files: files,
        commit_message: message_builder.commit_message,
        pr_description: message_builder.pr_message,
        pr_name: message_builder.pr_name,
        author_details: author_details,
        labeler: labeler,
        work_item: provider_metadata[:work_item]
      )
```

That way we don't pollute the public interface of the generic PR creator with implementation specific arguments. @feelepxyz is that also what you were thinking?

JamieMagee: Ah, okay. I thought you want to just pass in an object and extract properties as needed 🤦. Just to clarify you want the public interface to have all common properties, and then a has for platform specific properties?

common/lib/dependabot/pull_request_creator.rb
```diff
@@ -142,7 +144,8 @@ def azure_creator
         pr_description: message_builder.pr_message,
         pr_name: message_builder.pr_name,
         author_details: author_details,
-        labeler: labeler
+        labeler: labeler,
+        work_item: provider_metadata&.fetch(:work_item)
```
feelepxyz, line 148: Think this should default to nil otherwise it will raise

```suggestion
        work_item: provider_metadata&.fetch(:work_item, nil)
```

common/lib/dependabot/pull_request_creator/azure.rb
```diff
@@ -46,9 +47,7 @@ def azure_client_for_source
       end
 
       def branch_exists?
-        @branch_ref ||= azure_client_for_source.branch(branch_name)
-
-        @branch_ref
```
feelepxyz, line 51: Is this change intentional?

common/spec/dependabot/pull_request_creator_spec.rb
```diff
@@ -45,6 +46,7 @@
   let(:source) { Dependabot::Source.new(provider: "github", repo: "gc/bump") }
   let(:files) { [gemfile] }
   let(:base_commit) { "basecommitsha" }
+  let(:provider_metadata) { { work_item: 123 } }
```
feelepxyz, line 49: Maybe better to set this to nil here and only set it with a `work_item` in the `with an Azure source` context
```suggestion
  let(:provider_metadata) { nil }
```

common/lib/dependabot/pull_request_creator/azure.rb
JamieMagee: Yes, RubyMine highlighted this as an unnecessary store before return. It's now the same implementation as the GitHub client, however the GitLab and Code Commit clients still use this pattern.

feelepxyz: ah yes it's only used once from `create` so seems good to me 👍

common/lib/dependabot/clients/azure.rb
```diff
@@ -153,8 +153,9 @@ def create_commit(branch_name, base_commit, commit_message, files,
           "/pushes?api-version=5.0", content.to_json)
       end
 
+      # rubocop:disable Metrics/ParameterLists
       def create_pull_request(pr_name, source_branch, target_branch,
-                              pr_description, labels)
+                              pr_description, labels, work_item)
```
feelepxyz, line 158: Somewhat annoying but just thinking about existing users who might be calling these methods directly so should probably make these optional so we don't introduce breaking changes

```suggestion
                              pr_description, labels, work_item = nil)
```

common/lib/dependabot/pull_request_creator/azure.rb
```diff
@@ -8,11 +8,11 @@ class PullRequestCreator
     class Azure
       attr_reader :source, :branch_name, :base_commit, :credentials,
                   :files, :commit_message, :pr_description, :pr_name,
-                  :author_details, :labeler
+                  :author_details, :labeler, :work_item
 
       def initialize(source:, branch_name:, base_commit:, credentials:,
                      files:, commit_message:, pr_description:, pr_name:,
-                     author_details:, labeler:)
+                     author_details:, labeler:, work_item:)
```
feelepxyz, line 15:
```suggestion
                     author_details:, labeler:, work_item: nil)
```

feelepxyz (commented):
Looks good to me! Added some final comments about making the arguments optional so we don't accidentally introduce any breaking changes to existing users of the Azure PR creator.

jurre (approved):

PR #2536, opened 2020-09-20
Use <details> and <summary> tags for azure PR descriptions

GiriB:
Azure Devops now supports using html tags (`<details>`, `<summary>` and `<table>`) in the PR description. 

This is how it looks like:
image

Until now we used to skip using details and summary tags for azure. This PR removes that check and fixes https://github.com/dependabot/dependabot-core/issues/2537


GiriB:
I tried this locally and everything works as expected except this additional closing table tag. In few PRs I tried the description did not have an open `<table>` tag and the additional closing table tag that we added ended up in weird formatting (Azure Devops seems to handle this parsing different than Github, because in Github PR description works are expected even in presence of the extra mismatch open `table` tag)

common/lib/dependabot/pull_request_creator/message_builder.rb
```diff
@@ -454,7 +453,7 @@ def build_details_tag(summary:, body:)
       end
 
       def source_provider_supports_html?
-        !%w(azure codecommit).include?(source.provider)
+        !%w(codecommit).include?(source.provider)
```
jurre, line 456:
```suggestion
        source.provider != "codecommit"
```

jurre:
Thanks @GiriB, I know that @brrygrdn was working on a fix for an issue in GitHub with that closing table tag last week as well. In GH the `truncated` notice was not showing up at all, and it should be shown. Maybe he can chime in with how we're planning to fix that, and it might help here as well.

brrygrdn:
Hi @GiriB, I've just pushed a change to how we build the PR body that fixes the bug @jurre describes above: https://github.com/dependabot/dependabot-core/pull/2539

The tests would seem to indicate that for GitHub, we removed the closing table tag in Dependabot Core - likely in one of the sanitisation classes. I think we might need to make another change to where we test the provides somewhere to fix the issue you describe, I'll have a look.

common/lib/dependabot/pull_request_creator/message_builder.rb
```diff
@@ -442,7 +442,6 @@ def maintainer_changes_cascade(dep)
       end
 
       def build_details_tag(summary:, body:)
-        # Azure DevOps does not support <details> tag (https://developercommunity.visualstudio.com/content/problem/608769/add-support-for-in-markdown.html)
         # CodeCommit does not support the <details> tag (no url available)
         if source_provider_supports_html?
```
feelepxyz, line 446: This has been moved to a different file: https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/pull_request_creator/message_builder/metadata_presenter.rb#L147-L157 so there are some merge conflicts that need to be fixed

GiriB: Fixed conflicts.

GiriB:
@brrygrdn I created a PR using your latest changes, and unfortunately the `</tr></table>` still shows up. 

image

 You mentioned for Github "removed the closing table tag in Dependabot Core - likely in one of the sanitisation classes". Is this sanitisation happening outside of dependabot-core? 

Let me know if you want any help in validating this for azure. And thanks for taking a look!

GiriB:
Oops, I misread your comment. You say the sanitization happens within dependabot-core somewhere. Let me check on how we can get the same happen for azure client as well. :)

GiriB:
@brrygrdn Following the code, I don't see any Github specific sanitization happening - so whatever sanitization happens for Github also happens for azure.  Just to confirm, after your fix does the <table> tag gets removed from the description? I suspect that the `</tr></table>` does not get removed but instead just gets ignored by the Github UI rendering it.

brrygrdn:
@GiriB Ah, that's surprising. I'll have a look into the test case in #2539 and see why I thought it was being removed in `dependabot-core` - thanks for updating!

brrygrdn:
I put together a quick spike of this change plus some extra tests: https://github.com/dependabot/dependabot-core/compare/brrygrdn/html-closing-frag

It's really odd as [this test](https://github.com/dependabot/dependabot-core/compare/brrygrdn/html-closing-frag#diff-64234232d377c977f4afdc6f30393217R64-R66) on a GitHub-targetted PR shows the table tag is being removed by the dependabot code, but [this test](https://github.com/dependabot/dependabot-core/compare/brrygrdn/html-closing-frag#diff-64234232d377c977f4afdc6f30393217R79-R81) proves it isn't being removed for other sources.

I still can't see a point where this is happening - I'll have a look into this tomorrow morning, but I think this should be fixable with existing code, it's just finding a branch where we are testing the `source.provider` somewhere.

brrygrdn:
@GiriB I've updated my branch with a couple more commits that got the test I added passing:
https://github.com/dependabot/dependabot-core/compare/brrygrdn/html-closing-frag

I think this is probably the right direction, we'd need to add HTML-only sanitization for non-GitHub providers that also support HTML. I can see from looking at the sanitisation class that it handles markdown to <code> tags and a few other things that I'm not sure Azure supports, so I think we'd need to add a unit test for the HTML-only sanitisation that narrows down what should/shouldn't be supported.

Tizzit-max:
_Authicgoodjob

Tizzit-max:
8D @dependabot-rebase

jeffwidman:
I opened https://github.com/dependabot/dependabot-core/pull/6467 using Barry's branch mentioned ☝️ so we have a record for posterity...

PR #2541, opened 2020-09-21
Add temporary support for persistent_gems_after_clean in bundler

jurre:
This temporary feature allows people to list specific dependencies that
should never be pruned from vendor/cache.

This is a hack, and may be removed at any time without notice.


bundler/lib/dependabot/bundler/file_updater/lockfile_updater.rb
```diff
@@ -188,6 +188,13 @@ def cache_vendored_gems(definition)
           end
         end
 
+        # This is not officially supported and may be removed without notice.
+        def __keep_on_prune?(spec_name)
+          if (specs = ::Bundler.settings[:persistent_gems_after_clean])
```
feelepxyz, line 193: Will this get set from the `Gemfile`?

bundler/spec/dependabot/bundler/file_updater_spec.rb
```diff
@@ -1588,6 +1588,21 @@
         expect(file.deleted?).to eq true
       end
 
+      it "does not delete cached files marked as persistent" do
+        # This is not officially supported and may be removed without notice.
+        ::Bundler.settings.temporary(persistent_gems_after_clean: ["business"])
```
feelepxyz, line 1593: Should we try adding this config from the `Gemfile` using a file that's required? Unsure if bundler reads and sets this for us.

jurre: Yeah good shout

bundler/lib/dependabot/bundler/file_updater/lockfile_updater.rb
jurre: Yep, I've improved the tests to show this

feelepxyz (approved):
LGTM! Nice work on the specs 💯

Issue #2542, opened 2020-09-21
Dependabot Omnibus 0.119.5 throwing "undefined method `homepage_url'"


BobbyMcWho:
**Package manager/ecosystem**  
> ruby:bundler  

**Manifest contents prior to update**  
Gemfile  
```ruby
gem 'loofah', '2.3.0'
gem 'hashie', '~> 3'
gem 'rails', '~> 5.2.2'
gem 'sprockets', '< 4.0.0'
```
Gemfile.lock  
```ruby
GEM
  remote: https://gems.covermymeds.com/
  specs:
    actioncable (5.2.2)
      actionpack (= 5.2.2)
      nio4r (~> 2.0)
      websocket-driver (>= 0.6.1)
    actionmailer (5.2.2)
      actionpack (= 5.2.2)
      actionview (= 5.2.2)
      activejob (= 5.2.2)
      mail (~> 2.5, >= 2.5.4)
      rails-dom-testing (~> 2.0)
    actionpack (5.2.2)
      actionview (= 5.2.2)
      activesupport (= 5.2.2)
      rack (~> 2.0)
      rack-test (>= 0.6.3)
      rails-dom-testing (~> 2.0)
      rails-html-sanitizer (~> 1.0, >= 1.0.2)
    actionview (5.2.2)
      activesupport (= 5.2.2)
      builder (~> 3.1)
      erubi (~> 1.4)
      rails-dom-testing (~> 2.0)
      rails-html-sanitizer (~> 1.0, >= 1.0.3)
    activejob (5.2.2)
      activesupport (= 5.2.2)
      globalid (>= 0.3.6)
    activemodel (5.2.2)
      activesupport (= 5.2.2)
    activerecord (5.2.2)
      activemodel (= 5.2.2)
      activesupport (= 5.2.2)
      arel (>= 9.0)
    activestorage (5.2.2)
      actionpack (= 5.2.2)
      activerecord (= 5.2.2)
      marcel (~> 0.3.1)
    activesupport (5.2.2)
      concurrent-ruby (~> 1.0, >= 1.0.2)
      i18n (>= 0.7, < 2)
      minitest (~> 5.1)
      tzinfo (~> 1.1)
    arel (9.0.0)
    builder (3.2.4)
    concurrent-ruby (1.1.6)
    crass (1.0.5)
    erubi (1.9.0)
    globalid (0.4.2)
      activesupport (>= 4.2.0)
    hashie (3.6.0)
[...]

PR #2539, opened 2020-09-21
Fix missing truncation notice on GitHub PRs

brrygrdn:
We've encountered PRs where the Change Log notes have been truncated without any indication there is more content, leading to some confusion.

On investigation, we discovered that in an attempt to fix any table structures we have broken by truncating by line number we add a closing table table as part of the truncation line.

We rely on a later sanitisation set to remove this tag it if is not required, i.e. malformed, but this sanitisation currently removes the entire line including the truncation notice.

The fix itself is trivial, [we just insert a new line after the closing table tag](https://github.com/dependabot/dependabot-core/commit/7d1b1a7beb8d833c4baa0ca4ad3872cb19ca1fb1) and the sanitisation process works as intended.

The challenge of this PR was extracting a testable surface as the PR Builder class currently relies on a lot of webmocking to do its job. I've performed an extract-to-class on the `metadata_cascades_for_dep` class to make this easier to work with.

To keep this change somewhat understandable I've:

1. Avoided changing any existing tests, only added a new case for this bug
2. With the exception of the last three commits, I've kept the tests passing on each commit
3. I've tried to commit clear steps to make it easier to review-by-commit

Thanks to @jurre for pairing on the debugging and scoping of this work.

PR #2543, opened 2020-09-21
Revert an erroneously removed method in the PR message builder

brrygrdn:
See: https://github.com/dependabot/dependabot-core/issues/2542

This method was removed in https://github.com/dependabot/dependabot-core/pull/2539 as part of a class extraction of most of the `metadata_finder(dependency)` helpers.

Unfortunately, it seems there is no test coverage that hits this code line at time of writing.


jurre (approved):
Let's improve the coverage in a follow-up but get this fix out now 👍

common/lib/dependabot/pull_request_creator/message_builder.rb
```diff
@@ -327,6 +327,10 @@ def source_url(dependency)
         metadata_finder(dependency).source_url
       end
 
+      def homepage_url(dependency)
+        metadata_finder(dependency).homepage_url
+      end
+
```
BobbyMcWho, line 333: Does this warrant adding a test against removal of the method in the future?

brrygrdn:
Still reading through the class, there are a few other similar `_url` helper methods that have been ✂️  - will update with those and see if I can figure out a quick test case

common/lib/dependabot/pull_request_creator/message_builder.rb
brrygrdn: 👍🏻  I removed some self-stubbing on the unit-under test which was bypassing over these method calls.

common/spec/dependabot/pull_request_creator/message_builder_spec.rb
```diff
@@ -1699,16 +1699,23 @@ def commits_details(base:, head:)
...
   end
 
-  describe "#commit_message" do
+  describe "#commit_message", :vcr do
     subject(:commit_message) { builder.commit_message }
 
-    before do
-      allow(builder).to receive(:pr_name).and_return("PR name")
-      allow(builder).to receive(:commit_message_intro).and_return("Message")
-      allow(builder).to receive(:metadata_links).and_return("\n\nLinks")
```
brrygrdn, line 1708: I should have noticed these stubs meant the test wasn't exercising the real diff generation code, removing them and using VCR to capture a fixture ensures we hit the code that was broken and test for values on the `_url` helpers that were moved out of this class.

feelepxyz (approved):

PR #2546, opened 2020-09-23
PullRequestCreator::AnnotationError

thepwagner:
PullRequestCreators like `GitHub` separate the creation of PRs from
labelling, assigning reviewers, etc.

Since creating a pull request is a very noticeable side-effect for
customers, we'd prefer to throw an exception that attaches the 
half-completed PR, giving callers a hook for recovery.

An alternative would be _deleting_ the "failed" PR if annotation fails. Assuming PR creation is retried, I avoided that to squelch the `opened->closed->opened` noise.

### DIFF
--- a/common/lib/dependabot/pull_request_creator/github.rb
+++ b/common/lib/dependabot/pull_request_creator/github.rb
@@ -422,27 +422,38 @@ module Dependabot
       end
 
       def handle_error(err)
-        cause = case err
-                when AnnotationError
-                  err.cause
-                else
-                  err
-                end
+        case err
+        when AnnotationError
+          cause = err.cause
+          raiser = lambda do |type, message|
+            raise AnnotationError.new(
+              type.new(message),
+              err.pull_request
+            )
+          end
+        else
+          cause = err
+          raiser = ->(e, m) { raise e, m }
+        end
 
         case cause
         when Octokit::Forbidden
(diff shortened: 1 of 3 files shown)


common/lib/dependabot/pull_request_creator/github.rb
```diff
@@ -418,23 +423,42 @@ def github_client_for_source
 
       def handle_error(err)
         case err
+        when AnnotationError
+          cause = err.cause
+          raiser = lambda do |type, message|
```
thepwagner, line 428: The wrapping exception pattern, and this snippet for re-wrapping the delegated exceptions may be 🙃 .

In gopher terms, I was wanting to replace `return nil, err` with `return pr, err`.

jurre: It took me a few reads to grok this, wondering if it would it be easier to understand if this were a method instead of a lambda?

jurre (approved):

feelepxyz (approved):
Nice getting more context on these errors 👍  this could be a breaking change for some, would be good to call that out in the changelog and doing a minor bump.

Do we need to do anything in backend/api to handle this?

thepwagner:
@feelepxyz yep - backend changes are linked in the opposite direction (private -> public repo) :tinfoil:.

That PR mentions this is _probably_ `0.120.0`, so thanks for the 💯 !

PR #2551, opened 2020-09-24
Go mod tidy support

jurre:
This PR runs `go mod tidy` after updating go modules.

In order to do this, we rely on the full repo being checked out, and
because we now have access to the full source code, the implementation
has changed from a series of changes on temporary go modules, to running
the update in the context of the actual repo.

In order to properly test this, the test setup has changed. Instead of
updating the fixtures in tests, we set up a project folder for each
test, that mimics a minimal repository. This project is then copied over
and turned into a repository in the tests, and we perform the updates on
this repository.

There are a few things left to do:

- [x] No longer pass the `go.mod` and `go.sum` to `GoModUpdater`, it already has access to this via the repo_contents_path
- [x] Ensure we always pass the `repo_contents_path` and do a full checkout of the repo for go_modules, and represent this in all tests
  - [x] This means we may be able to get rid of the `FileFetcher`, or we can change its implementation to use the `repo_contents_path`. This optimization could then also be used in the Bundler vendoring implementation.
- [x] Beef up `go mod tidy` tests
- [x] ~Look into merging `GoModUpdater` into `FileUpdater`, it does not need to be its own class~ no


thepwagner:
> Look into merging GoModUpdater into FileUpdater, it does not need to be its own class

I'm fuzzy here; I think if we end up doing go module major version updates - the `FileUpdater` being composed of a `GoModUpdater` and `RestOfTheSourceCodeUpdater` (naming) is 💯 ?

jurre:
>  I think if we end up doing go module major version updates - the FileUpdater being composed of a GoModUpdater and RestOfTheSourceCodeUpdater (naming) is 💯 

Aye, good point. Although I'm not sure if we should split those out, I think we may want to run `go mod tidy` _after_ fixing the import paths? But as I haven't really thought it through we should probably leave it as is, PR is already big enough as is 👍

thepwagner (commented):
Awesome WIP, thanks for the early peek!
I reviewed with a lens of monorepos for some weird reason, all secondary concerns.

> PR is already big enough as is

Splitting the `repo_contents_path` + test refactor out of the feature improvements maybe a way to slice?

> I think we may want to run go mod tidy after fixing the import paths? 

💯 ; `GoModUpdater` -> `RestOfTheSourceCodeUpdater` -> `GoSumUpdater` -> `GoModVendorer` - so many factories :trollface:

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -25,11 +25,15 @@ class GoModUpdater
           /go: ([^@\s]+)(?:@[^\s]+)?: .* declares its path as: ([\S]*)/m
         ].freeze
 
-        def initialize(dependencies:, go_mod:, go_sum:, credentials:)
+        # TODO: No need to pass go_mod and go_sum anymore, we can grab them from
```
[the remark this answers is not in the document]
jurre: > Question: don't we need this path for repositories that aren't providing repo_contents_path?

I was building this under the assumption that we always pass the `repo_contents_path` for go_modules, it makes the implementation a lot simpler, especially we unlock major version updates and vendoring. We also wouldn't be able to run `go mod tidy` without the `repo_contents_path`, so we would essentially end up with two implementations that we'd have to maintain.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -42,64 +46,66 @@ def updated_go_sum_content
...
-          # undesirable ways (such as injecting the current Go version), so we
-          # need to update the original go.mod with the updated set of
-          # requirements rather than using the regenerated file directly
-          original_reqs = in_temp_dir(stub_dirs) do
-            parse_manifest_requirements(go_mod.content)
-          end
-          updated_reqs = in_temp_dir(stub_dirs) do
-            parse_manifest_requirements(regenerated_files[:go_mod])
-          end
+          SharedHelpers.in_a_temporary_repo_directory("/", repo_contents_path) do
```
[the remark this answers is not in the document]
jurre: The `repo_contents_path` is now set to the directory that's being updated: https://github.com/dependabot/dependabot-core/pull/2551/commits/77857ec217451287b017fc3b112be56cc119299b#diff-a619536a45c6fc217c2f5e4c56460121R76

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -108,78 +114,75 @@ def update_go_mod(go_mod_content, dependencies)
...
           _, stderr, status = Open3.capture3(ENVIRONMENT, "go get -d")
           handle_subprocess_error(stderr) unless status.success?
-
-          updated_go_sum = go_sum ? File.read("go.sum") : nil
-          { go_mod: File.read("go.mod"), go_sum: updated_go_sum }
         end
 
         def parse_manifest_requirements(go_mod_content)
-          File.write("go.mod", go_mod_content)
+          SharedHelpers.in_a_temporary_directory do
```
thepwagner, line 137: Nit: we can skip this temporary directory and IO.

AFAICT we read the file into ruby, then create a temporary directory and write the file back out there.
But since this doesn't change the base file, we could skip some IO by just parsing from the current directory?

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -73,6 +73,10 @@ def clone_repo_contents(target_directory: nil)
           _clone_repo_contents(target_directory: target_directory)
       end
 
+      def repo_contents_path
+        File.join(clone_repo_contents, directory)
```
thepwagner, line 77: Noice: I think I get how multi-module repos work 🎉

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -188,16 +191,14 @@ def in_temp_dir(stub_paths, &block)
...
-        def replace_directive_substitutions(go_mod_content)
+        def replace_directive_substitutions
           @replace_directive_substitutions ||=
-            SharedHelpers.in_a_temporary_directory do |path|
-              File.write("go.mod", go_mod_content)
-
+            begin
               # Parse the go.mod to get a JSON representation of the replace
               # directives
               command = "go mod edit -json"
```
thepwagner, line 199: Nit: iff the ` parse_manifest_requirements` comment above, it could be reused here too.

thepwagner: Nit: it's weird we're forking `go mod edit -json` at least 3 times during the update process...

AFAICT this is using the same output as we use for `original_paths` calculation and could be shared?

thepwagner (commented):
My biggest concern in this rev is around dropping support for `!repo_contents_path`.
Inline comments assume this is 👍 , but I'd assumed cloning and tidying would be opt-in features, perhaps even independent.

~AFAICT you can do an "alright" `go mod tidy` with generating source code from `go.mod`, right?
I'm assuming a null-import would make every path "used", so invoking `go mod tidy` will clean out old module versions from `go.sum`. That's <💯  but probably enough for some.~ 

Nope. I timeboxed this and came up with an internal counter-example easily.
* Project depends on `github.com/docker/libcontainer/netlink`
* that means go.mod contains: `github.com/docker/libcontainer v2.2.1+incompatible`
* using the "just stub from go.mod!" approach, we end up importing `_ "github.com/docker/libcontainer"`, which in turn needs [Sirupsen/logrus](https://github.com/docker-archive/libcontainer/blob/v2.2.1/init_linux.go#L12)
* Since the `./netlink` package _doesn't_ use that old version of logrus, my project doesn't have a `replace` setup to handle it.

common/lib/dependabot/utils.rb
```diff
@@ -29,5 +29,15 @@ def self.requirement_class_for_package_manager(package_manager)
     def self.register_requirement_class(package_manager, requirement_class)
       @requirement_classes[package_manager] = requirement_class
     end
+
+    @cloning_package_managers = []
```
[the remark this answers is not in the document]
jurre: Yeah good suggestion, should be a set :+1:

common/lib/dependabot/utils.rb
```diff
@@ -29,5 +29,15 @@ def self.requirement_class_for_package_manager(package_manager)
...
+
+    def self.always_clone_for_package_manager?(package_manager)
+      @cloning_package_managers.include?(package_manager)
+    end
+
+    def self.register_always_clone(package_manager)
```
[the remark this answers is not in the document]
jurre: This is mostly a workaround so we can adhere to the current API to stitch things together, but we need to take the `repo_contents_path` from the `FileFetcher` only if we are cloning, so we can pass it into the `FileParser`, `UpdateChecker` and `FileUpdater`. There's currently no clean API that those objects can use to communicate this to each other.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jurre: We pass in different versions of the `go.mod` file here, one before and one after the changes, for the `updated_reqs` we _could_ skip the IO, but we'd have to have to add a different helper I think?

jurre: Ah we call it at different times, yeah that works 👍

jurre:
The last two commits to use the go.mod path in the errors are two alternatives to accomplish the same thing, I _think_ I prefer passing it in (last approach). @thepwagner thoughts?

thepwagner:
@jurre I think https://github.com/dependabot/dependabot-core/pull/2551/commits/f2b1d1001f1ce25c188f4f8a6f672bb46a2a36f3

I don't like how the other uses `FileUpdater.go_mod`, which ends up relying on `dependency_files`? Seems further from our glorious `repo_contents_path`-only future.

jurre:
> maybe GoModuleUpdater should accept repo_contents_path + directory

Yeah that's a good shout, I like that 👍

go_modules/lib/dependabot/go_modules/file_updater.rb
```diff
@@ -60,9 +67,9 @@ def file_updater
         @file_updater ||=
           GoModUpdater.new(
             dependencies: dependencies,
-            go_mod: go_mod,
-            go_sum: go_sum,
-            credentials: credentials
+            credentials: credentials,
+            repo_contents_path: repo_contents_path,
+            directory: directory
```
feelepxyz, line 72: Maybe not the clearest but you could get the `directory` from any of the `dependencies`, should be fixed for all of them.

common/lib/dependabot/shared_helpers.rb
```diff
@@ -37,10 +37,12 @@ def initialize(error_class:, error_message:, error_backtrace:)
 
     def self.in_a_temporary_repo_directory(directory = "/",
                                            repo_contents_path = nil,
+                                           branch = nil,
                                            &block)
       if repo_contents_path
         path = Pathname.new(File.join(repo_contents_path, directory)).
                expand_path
+        checkout_git_branch(repo_contents_path, branch) if branch
```
feelepxyz, line 45: ☕ 🚗  will this not no-op with the current implementation of `checkout_git_branch`? https://github.com/dependabot/dependabot-core/blob/14ddb5b533f72a0b31c2398ef2c4ea2a5d9bf9a6/common/lib/dependabot/file_fetchers/base.rb#L436

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -231,28 +227,18 @@ def handle_subprocess_error(stderr)
...
 
-        def dummy_main_go
-          # If we use `main` as the package name, running `go get -d` seems to
-          # invoke the build systems, which can cause problems. For instance,
-          # if the go.mod includes a module that doesn't have a top-level
-          # package, we have no way of working out the import path, so the
-          # build step fails.
-          #
-          # In due course, if we end up fetching the full repo, it might be
-          # good to switch back to `main` so we can surface more errors.
```
feelepxyz, line 249: 😍

go_modules/lib/dependabot/go_modules/file_updater.rb
jurre: Ah, yeah good shout, let's do that 👍

common/lib/dependabot/shared_helpers.rb
jurre: I ran into this in some tests and while dry running, but those tests have since been changed to remove the cloned repo, and in a prod environment the repo should never be left behind, so you're right 👍

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -25,11 +25,13 @@ class GoModUpdater
           /go: ([^@\s]+)(?:@[^\s]+)?: .* declares its path as: ([\S]*)/m
         ].freeze
 
-        def initialize(dependencies:, go_mod:, go_sum:, credentials:)
+        def initialize(dependencies:, credentials:, repo_contents_path:,
+                       directory:, tidy:)
```
thepwagner, line 29: Question: should these (and assumed extensions: `vendor:`, etc) be individual booleans like this?

I think `!@repo_contents_stub` needs to be passed through to avoid things that assume there's a full clone,  🥔 🥔 as to how.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -108,78 +120,71 @@ def update_go_mod(go_mod_content, dependencies)
...
+          write_go_mod(body)
         end
 
-        def run_go_get(go_mod_content, go_sum)
-          File.write("go.mod", go_mod_content)
-          File.write("go.sum", go_sum.content) if go_sum
-          File.write("main.go", dummy_main_go)
+        def run_go_get
+          unless Dir.glob("*.go").any?
+            File.write("main.go", "package dummypkg\n")
```
thepwagner, line 135: Nit: this isn't a problem _yet_...

Once we're doing major version upgrades, we're affecting `**/*.go`, so we'll probably end up with something like `git status ...` (like `bundler`, right?).
I'm 😨 this temp go file will be caught then.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -212,10 +209,12 @@ def replace_directive_substitutions(go_mod_content)
             end
         end
 
-        def substitute_all(file, substitutions)
-          substitutions.reduce(file) do |text, (a, b)|
+        def substitute_all(substitutions)
+          body = substitutions.reduce(File.read("go.mod")) do |text, (a, b)|
```
thepwagner, line 213: Idea: keep passing `original_go_mod` to avoid reading this again? I don't think it's changed.

thepwagner (approved):
LGTM, comments are mostly seeding discussion about stuff I added and got lost in the mergedown.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jurre: Yeah, thanks for calling that out! I was thinking once we start doing that, maybe we can create a file with a random unique identifier here and filter that out when finding the changed files?

go_modules/lib/dependabot/go_modules/file_updater.rb
```diff
@@ -56,13 +75,18 @@ def go_sum
         @go_sum ||= get_original_file("go.sum")
       end
 
+      def directory
+        dependency_files.first.directory
```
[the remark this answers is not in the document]
jurre: Yeah, that's a good point, I think we'll keep passing in `dependency_files` for the foreseeable future though, at least as long as the majority of the ecosystems need them, should we re-evaluate once we move more package managers in this direction?

go_modules/lib/dependabot/go_modules/file_fetcher.rb
```diff
@@ -17,23 +17,30 @@ def self.required_files_message
...
+        # Ensure we always check out the full repo contents for go_module
+        # updates.
+        SharedHelpers.in_a_temporary_repo_directory(
+          directory,
+          clone_repo_contents
+        ) do
+          unless go_mod
+            raise(
+              Dependabot::DependencyFileNotFound,
+              File.join(directory, "go.mod")
```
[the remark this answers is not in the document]
jurre: Probably yes!

bin/dry-run.rb
```diff
@@ -301,16 +301,18 @@ def cached_dependency_files_read
   commit: $options[:commit]
 )
 
-$repo_contents_path = nil
+$repo_contents_path = Dir.mktmpdir if $options[:clone]
```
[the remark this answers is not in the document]
jurre: > is that better?

It's a little meh, because it also means that we won't fetch different branches or pull in updates if `origin main` has changed 🤔 I'm OK with the temp file every run

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
thepwagner: Could we just remove it in an `ensure` block after calling `go get`?

jurre: I'm worried that we'll delete actual `main.go` files that way, there's definitely a way to reliably do it, but do you think we should already worry about this now, or can we punt on it until we shell out to git to find changed files?

jurre: Scratch that, might as well do it now before we forget, I pushed a fix, lmk what you think!

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jurre: Looked into it, but we call this method twice, the second time with the inverted substitutions. The second time we haven't read the file yet, we can pull another `File.read` into the `update_files` right before calling this and avoid a single file read, but I think I prefer keeping the code simpler over avoiding a single file read. WDYT?

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jurre: For the time being I think that's 👍 would also keep the doors open to make `go_mod_tidy` a configuration option via the config file

thepwagner:
@jurre how can I help this land?

jurre:
> @jurre how can I help this land?

I think we just merge it?

feelepxyz (approved):
Epic work on this and all related PRs 🎉  would be cool to get a walk-through of the main changes, never really worked on go modules so don't actually know how it works 🙈

jurre:
> Epic work on this and all related PRs 🎉 would be cool to get a walk-through of the main changes, never really worked on go modules so don't actually know how it works 🙈

@feelepxyz  wanna do this before we merge it? I'm available all afternoon

feelepxyz:
> @feelepxyz wanna do this before we merge it? I'm available all afternoon

Yeah 👍 zoom in 30 mins ok?

Issue #1699, opened 2020-02-20
Support Bundler v2


xtreme-jason-smith:
For the past month or so, Dependepot has been opening and closing issues on our repo with the following issue:
```
Bundler::VersionConflict with message: Bundler could not find compatible versions for gem "bundler":
  In Gemfile:
    license_finder was resolved to 0.0.1, which depends on
      bundler (~> 2.1.0)

  Current Bundler version:
    bundler (1.17.3)
This Gemfile requires a different version of Bundler.
Perhaps you need to update Bundler by running `gem install bundler`?

Could not find gem 'bundler (~> 2.1.0)', which is required by gem 'license_finder', in any of the sources.

Bundler could not find compatible versions for gem "rack":
  In Gemfile:
    rack (> 1.6, ~> 2.0.7)

    capybara (~> 3.15.0) was resolved to 3.15.1, which depends on
      rack (>= 1.6.0)

    rack-test (> 0.7, ~> 1.1.0) was resolved to 1.1.0, which depends on
      rack (>= 1.0, < 3)
```
This aligns with our recent changes to lock our bundler version to ~> 2.1.0 in our gemspec file. However, it appears that Dependebot Docker image only has Bundler 1.17.3 installed.

It should be updated to also support Bundler 2.x

hadees:
I'm having the same issue, pretty frustrating since I don't think I can do anything about it.  We have no way to specify a bundler version.

feelepxyz:
Sorry about this, Dependabot currently doesn't support bundler v2. We have it on our backlog and will get to it but we're currently focusing on integrating Dependabot natively within GitHub so we've slowed down on package ecosystem support while we do this.

GiriB:
@feelepxyz  [**Bundler 1.x are prone to injection attack via gem name collisions**](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-7954). Shouldn't we prioritize moving to 2.x?

jamesbebbington:
Hey @feelepxyz Dependabot is a wonderful tool, and when combined with GitHub Actions, truly is a match made in heaven! Any update on when support for bundler 2.x might be added? Thanks.

feelepxyz:
[...]

PR #2552, opened 2020-09-24
Extract Bundler helpers to isolated native helpers

brrygrdn:
Lays the groundwork for supporting multiple Bundler versions (specifically Bundler 2) by extracting the helpers currently run in a forked process to isolated sub-process helpers.

Native helper scripts are currently run in the root directory with `bundle exec ../path/to/native-helpers` meaning native helper scripts could access gems in common but we're planning on running `bundle exec` in the tmp directory with the users `Gemfile.lock` when adding bundler 2. This will allow bundler to do the version switching for us based on what's set in the lockfile's `BUNDLED WITH`. 

Deprecating `SharedHelper.in_a_forked_process` used to run `Bundler` commands in an isolated way. All 'forked process' bundler helpers have been migrated to a sub-process where we send back a hash from the sub-process using `JSON.dump/load` instead of using `Marshal.dump/load`.

We've also spotted a potential reduction in unknown bundler errors. While testing dry-runs against preview the only discrepancy between the update job error created on main and this branch was that main raised an unknown error `marshal data too short` (this error is from using `Marshal.dump/load` to communicate between the forked process).

#### Spec changes
We can no longer stub native helper requests in core so these specs have been moved down to the native helper specs, we still have integration specs in core that hit the internet (would be neat to build/use a VCR-like proxy server that allows us to stub native helpers)

#### Changes to common
- `HelperSubprocessFailed` now accepts an optional `error_class` attribute used to match against the bundler exception class
- Added a `DEBUG_HELPERS` env var that `puts stdout` and `stderr` from the sub-process 

Related issue: https://github.com/dependabot/dependabot-core/issues/1699

#### Questions
- Are there risks in running `bundle exec` without clearing the current env? Could malicious packages inject ENV vars that are picked up by `bundler`?
  - We could set `unsetenv_others: true` option when executing the child-process to clear out existing ENV but a quick test using this broke all tests 🤔


bundler/lib/dependabot/bundler/file_parser.rb
```diff
@@ -1,9 +1,21 @@
...
+# The final version is refacor anywhere we use these classes to use
+# Dependabot::Dependency objects and instantiate them from the JSON object
+# retrieved via the shell-out to the native helper
+require "ostruct"
+
 require "dependabot/dependency"
 require "dependabot/file_parsers"
 require "dependabot/file_parsers/base"
 require "dependabot/bundler/file_updater/lockfile_updater"
+require "dependabot/bundler/native_helpers"
```
feelepxyz, line 18: Miss checking this in?

bundler/helpers/lib/functions.rb
```diff
@@ -0,0 +1,99 @@
...
+                           gemfile_name:, lockfile_name:, using_bundler_2:,
+                           dir:, credentials:)
+    VersionResolver.new(
+      dependency_name: dependency_name,
+      dependency_requirements: dependency_requirements,
+      gemfile_name: gemfile_name,
+      lockfile_name: lockfile_name,
+      using_bundler_2: using_bundler_2,
+      dir: dir,
+      credentials: @credentials
```
brrygrdn, line 96:
```suggestion
      credentials: credentials
```

bundler/helpers/lib/functions/version_resolver.rb
```diff
@@ -0,0 +1,206 @@
...
+      details
+    end
+
+    private
+
+    def using_bundler_2?
+      @using_bundler_2
+    end
+
+    def setup_bundler
```
brrygrdn, line 63: This method can probably move to `Functions::setup_bundler(dir:, credentials:, bundler_2:)` or similar as it's used in ~3 places.

bundler/helpers/lib/functions/lockfile_updater.rb
```diff
@@ -0,0 +1,261 @@
...
+      end
+
+      # Fall back to unlocking *all* sub-dependencies. This is required
+      # because Bundler's VersionConflict objects don't include enough
+      # information to chart the full path through all conflicts unwound
+      dependencies_to_unlock.append(*allowed_new_unlocks)
+    end
+    # rubocop:enable Metrics/PerceivedComplexity
+
+    def build_definition(dependencies_to_unlock)
```
brrygrdn, line 191: Several of the function classes implement this, there's some light variance in arguments but the gist of the implementation looks very close.

This is maybe a candidate to extract and unit test?

bundler/lib/dependabot/bundler/update_checker/force_updater.rb
```diff
@@ -42,154 +39,32 @@ def update_multiple_dependencies?
...
-          raise_unresolvable_error(e)
-        end
-
-        #########################
-        # Bundler context setup #
-        #########################
-
-        def in_a_temporary_bundler_context
           base_directory = dependency_files.first.directory
           SharedHelpers.in_a_temporary_repo_directory(base_directory,
```
brrygrdn: This can probably be switched over to use `in_a_native_bundler_context` as it is equivolent of lines 42-46 with unified bundler error handling, retrying etc.

bundler/lib/dependabot/bundler/update_checker/latest_version_finder/dependency_source.rb
```diff
@@ -47,27 +47,32 @@ def versions
...
                 )
-
-                # Tell Bundler we're fine with fetching the source remotely
-                source.instance_variable_set(:@allow_remote, true)
-
-                spec = source.specs.first
-                { version: spec.version, commit_sha: spec.source.revision }
               end
+            end.transform_keys(&:to_sym)
+          rescue Dependabot::SharedHelpers::HelperSubprocessFailed => e
```
brrygrdn, line 70: TODO: This should be refactored to align with the subprocess error handling in `VersionResolver`

bundler/helpers/lib/functions/version_resolver.rb
```diff
@@ -0,0 +1,206 @@
...
+      begin
+        definition = build_definition(dependencies_to_unlock)
+        definition.resolve_remotely!
+      rescue ::Bundler::GemNotFound => e
+        unlock_yanked_gem(dependencies_to_unlock, e) && retry
+      rescue ::Bundler::HTTPError => e
+        # Retry network errors
+        attempt ||= 1
+        attempt += 1
+        raise if attempt > 3 || !e.message.include?("Network error")
```
brrygrdn, line 65: NOTE: These retries are in addition to retries performed by the main process in `SharedBundlerHelpers#in_a_native_bundler_context` - I'm not sure this should be removed as it's pre-existing logic, but it might be worth a comment to acknowledge we are doing some additional low-level retries which might be multiplicative with the main process?

bundler/lib/dependabot/bundler/update_checker/latest_version_finder/dependency_source.rb
feelepxyz: 👍  yeah was running into some issues here were `error_class` wasn't set

bundler/helpers/lib/functions/version_resolver.rb
feelepxyz: Ah yeah looks like it's already handled in `in_a_native_bundler_context` > `RETRYABLE_ERRORS`

brrygrdn: I think the double retries may have pre-existed in master, I'll go have a look and verify.

bundler/helpers/lib/functions/lockfile_updater.rb
brrygrdn: We should punt on this for now as it means further transforms on a large/tricky extraction

bundler/helpers/lib/shared_native_helpers.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+require "excon"
+
+module SharedNativeHelpers
+  USER_AGENT = "dependabot-core/bundler-helper "\
```
brrygrdn, line 6: I wonder if we should pass the Dependabot::VERSION to the native script as an ENVVAR rather than lose the version string here.

common/lib/dependabot/shared_helpers.rb
```diff
@@ -136,6 +136,7 @@ def self.excon_headers(headers = nil)
       }.merge(headers)
     end
 
+    # Duplicated in bundler/helpers/lib/shared_native_helpers.rb
```
brrygrdn, line 139: 👌🏻

bundler/helpers/lib/shared_native_helpers.rb
feelepxyz: yeah good catch 👍

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
```diff
@@ -455,7 +432,7 @@
 
           it "ignores the minimum ruby version in the gemspec" do
             expect(resolver.latest_resolvable_version_details[:version]).
-              to eq(Gem::Version.new("7.2.0"))
+              to eq(Gem::Version.new("7.4.0"))
```
brrygrdn, line 435: I think this version number change was a naieve 'fix' for us by-passing stubs and hitting the internet for real.

We should probably walk this back and move this test into the native spec.

bundler/helpers/lib/functions/file_parser.rb
```diff
@@ -0,0 +1,104 @@
...
+        name: dependency.name,
+        requirement: dependency.requirement,
+        groups: dependency.groups,
+        source: source_for(dependency),
+        type: dependency.type
+      }
+    end
+
+    # Can't be a constant because some of these don't exist in bundler
+    # 1.15, which Heroku uses, which causes an exception on boot.
```
feelepxyz, line 92:
```suggestion
    # 1.15, which used to cause issues on Heroku (causing exception on boot).
    # TODO: Check if this will be an issue with multiple bundler versions
```

bundler/helpers/lib/functions/version_resolver.rb
```diff
@@ -0,0 +1,172 @@
...
+    def dependency_from_definition(unlock_subdependencies: true)
+      dependencies_to_unlock = [dependency_name]
+      dependencies_to_unlock += subdependencies if unlock_subdependencies
+      begin
+        definition = build_definition(dependencies_to_unlock)
+        definition.resolve_remotely!
+      rescue ::Bundler::GemNotFound => e
+        unlock_yanked_gem(dependencies_to_unlock, e) && retry
+      rescue ::Bundler::HTTPError => e
+        # Retry network errors
```
feelepxyz, line 62:
```suggestion
        # Retry network errors
        # Note: in_a_native_bundler_context will also retry `Bundler::HTTPError` errors
        # up to three times meaning we'll end up retrying this error up to six times
        # TODO: Could we get rid of this retry logic and only rely on 
        # SharedBundlerHelpers.in_a_native_bundler_context
```

bundler/helpers/lib/functions/version_resolver.rb
```diff
@@ -0,0 +1,172 @@
...
+
+      fetcher = dep.source.fetchers.first.fetchers.first
+
+      # It's only the old index we have a problem with
+      return false unless fetcher.is_a?(::Bundler::Fetcher::Dependency)
+
+      # If no Ruby version is specified, we don't have a problem
+      return false unless ruby_version
+
+      versions = Excon.get(
```
feelepxyz, line 138: How do we have access to Excon?

bundler/helpers/lib/shared_native_helpers.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+require "excon"
+
+module SharedNativeHelpers
+  # Duplicated in lib/dependabot/bundler/file_updater/lockfile_updater.rb
+  LOCKFILE_ENDING = /(?<ending>\s*(?:RUBY VERSION|BUNDLED WITH).*)/m.freeze
```
feelepxyz, line 7:
```suggestion
  # TODO: Stop sanitizing the lockfile once we have bundler 2 installed
  LOCKFILE_ENDING = /(?<ending>\s*(?:RUBY VERSION|BUNDLED WITH).*)/m.freeze
```

bundler/spec/native_helpers/functions/version_resolver_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+        stub_request(:get, gemfury_deps_url + "?gems=business,statesman").
+          to_return(status: 200, body: fixture("ruby", "gemfury_response"))
+        stub_request(:get, gemfury_deps_url + "?gems=business").
+          to_return(status: 200, body: fixture("ruby", "gemfury_response"))
+        stub_request(:get, gemfury_deps_url + "?gems=statesman").
+          to_return(status: 200, body: fixture("ruby", "gemfury_response"))
+      end
+
+      its([:version]) { is_expected.to eq(Gem::Version.new("1.9.0")) }
+      its([:fetcher]) { is_expected.to eq("Bundler::Fetcher::Dependency") }
```
brrygrdn, line 71: One thing I've realised in writing this test, if we get a version back from a _private_ registry, the fetcher will still show as `Bundler::Fetcher::Dependency`.

This means we will test the the gem's ruby requirement via this method: https://github.com/dependabot/dependabot-core/pull/2552/files#diff-3092b190ebf8948a50c8d770607f2417R172

My initial thought was that since the old method guards against the source being anything other than a `::Bundler::Source::Rubygems`, we should pass back the source as well and reinstate that check - which I think is a positive change - it doesn't fix this problem.

In the case where we are fetching from a private package manager the source will still be `::Bundler::Source::Rubygems` as proved by some of our gemfury specs elsewhere.

I think this means we have a pre-existing bug here where, if you have a private mirror of a gem that also exists in ruby gems, we will return update not possible if we find the rubygems version isn't compatible with your ruby version.

This feels like an extreme edge case and I'm reluctant to change existing behaviour as part of this diff but I just want to make sure my reasoning is correct.

brrygrdn: > My initial thought was that since the old method guards against the source being anything other than a ::Bundler::Source::Rubygems, we should pass back the source as well and reinstate that check - which I think is a positive change - it doesn't fix this problem.

The tests failed due to this check being missing - we can't get a fetcher from a non-rubygems source. This is fixed in 40a0464

bundler/lib/dependabot/bundler/update_checker/version_resolver.rb
```diff
@@ -228,55 +188,27 @@ def ruby_version_incompatible?(dep)
           # version details (could be that it's a private index, etc.)
           return false unless versions.status == 200
 
+          puts JSON.parse(versions.body)
```
feelepxyz, line 191:
```suggestion
```

bundler/spec/dependabot/bundler/update_checker/latest_version_finder_spec.rb
```diff
@@ -336,56 +385,23 @@
...
+                 }).
+            and_return(
+              []
             )
         end
 
         it { is_expected.to be_nil }
       end
-
-      context "that only implements the old Bundler index format..." do
```
feelepxyz, line 359: Moved to native helpers dependency_source_spec.rb

bundler/spec/dependabot/bundler/update_checker/latest_version_finder_spec.rb
```diff
@@ -336,56 +385,23 @@
...
-              status: 200,
-              body: fixture("ruby", "contribsys_old_index_prerelease_response")
-            )
-        end
-
-        its([:version]) { is_expected.to eq(Gem::Version.new("3.5.2")) }
-      end
     end
 
+    # TODO: Test appears to be hitting the internet
```
feelepxyz, line 404:
```suggestion
```

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
```diff
@@ -306,30 +311,6 @@
       end
     end
 
-    context "with a private gemserver source" do
```
feelepxyz, line 309: Moved to native helpers version_resolver_spec.rb

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
```diff
@@ -213,6 +211,7 @@
         its([:version]) { is_expected.to eq(Gem::Version.new("3.6.0")) }
       end
 
+      # TODO: Move to a native helper test spec
```
feelepxyz, line 214:
```suggestion
```

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
feelepxyz: 👍  thoughts on keeping it here for now and adding a TODO to move to native helper spec? Could do another pass at specs when we add bundler 2.

bundler/spec/native_helpers/functions/version_resolver_spec.rb
feelepxyz: > I think this means we have a pre-existing bug here where, if you have a private mirror of a gem that also exists in ruby gems (or if a private gem shadows the name of a 'real' gem in ruby gems ), we will return update not possible if we find the rubygems version isn't compatible with your ruby version as we think we've retrieved it from the old rubygems API.
>
> This feels like an extreme edge case and I'm reluctant to change existing behaviour as part of this diff but I just want to make sure my reasoning is correct.

Good catch! Sounds plausible. Does indeed feel like an extreme edge case 😅  maybe worth adding a comment explaining this potential issue in case we run into it in future? 👍  on holding off on a fix in this PR.

bundler/helpers/lib/shared_native_helpers.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+require "excon"
+
+module SharedNativeHelpers
+  # Duplicated in lib/dependabot/bundler/file_updater/lockfile_updater.rb
+  # TODO: Stop sanitizing the lockfile once we have bundler 2 installed
+  LOCKFILE_ENDING = /(?<ending>\s*(?:RUBY VERSION|BUNDLED WITH).*)/m.freeze
```
feelepxyz, line 8: Only used in tests, could move to native test helpers

bundler/lib/dependabot/bundler/update_checker/force_updater.rb
```diff
@@ -1,20 +1,20 @@
...
-require "dependabot/bundler/file_parser"
-require "dependabot/shared_helpers"
 require "dependabot/errors"
+require "dependabot/shared_helpers"
 
 module Dependabot
   module Bundler
     class UpdateChecker
       class ForceUpdater
+        require_relative "shared_bundler_helpers"
```
feelepxyz, line 15: Move to a regular `require`

bundler/spec/native_spec_helper.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  # We don't always need a lockfile, so define a nil value tests can override
+  let(:lockfile_fixture_name) do
+    nil
+  end
+
+  # TODO: Stop sanitizing the lockfile once we have bundler 2 installed
+  let(:lockfile_fixture) do
+    fixture("ruby", "lockfiles", lockfile_fixture_name).
+      gsub(SharedNativeHelpers::LOCKFILE_ENDING, "")
```
feelepxyz, line 32: Move `LOCKFILE_ENDING ` to spec helper

common/lib/dependabot/shared_helpers.rb
```diff
@@ -151,6 +136,7 @@ def self.excon_headers(headers = nil)
...
```
feelepxyz, line 139:
```suggestion
```

bundler/helpers/lib/shared_native_helpers.rb
brrygrdn: Removed ✂️

Issue #2550, opened 2020-09-24
Flaky auth problems with PyPI


webknjaz:
**Package manager/ecosystem**

python

**Manifest contents prior to update**

N/A

**Updated dependency**

N/A

**What you expected to see, versus what you actually saw**

Dependabot (the old GitHub App) has created a lot of issues like https://github.com/aio-libs/aiodocker/issues/502 across aio-libs org. Restarting some of the updates helped but this one hasn't succeeded.

Logs (https://app.dependabot.com/accounts/aio-libs/update-logs/48591655) are full of HTTP 403 responses:
```console
  proxy | 2020/09/24 09:52:24 [004] GET https://pypi.python.org:443/simple/tox-travis/
  proxy | 2020/09/24 09:52:24 [004] 403 https://pypi.python.org:443/simple/tox-travis/
  proxy | 2020/09/24 09:52:24 [006] GET https://pypi.python.org:443/simple/
  proxy | 2020/09/24 09:52:24 [006] 403 https://pypi.python.org:443/simple/
```
(there's this block of log lines per each dep)

I don't know anything about how dependabot is deployed but I can speculate that either Dependabot gets rate-limited (unlikely because then the HTTP code wouldn't be 403 unless there's a bug in https://github.com/pypa/warehouse) or your dependabot deployment is distributed and uses multiple API tokens in PyPI and one (some?) of them has been invalidated.

**Images of the diff or a link to the PR, issue or logs**

Example issue https://github.com/aio-libs/aiodocker/issues/502

di:
Hi! PyPI admin here. This doesn't seem like an issue with our backends, these endpoints should never return a 403.

Also, as an aside, I noticed that dependabot is configured to use `pypi.python.org` instead of `pypi.org`, which likely is resulting in lots of unnecessary redirects.

webknjaz:
Looks like all the links to this issue originate from issues created by the GitHub App version of Dependabot and not the new "native" one...

Bystroushaak:
Wtf is this? It spammed me with 12 issues and I have no idea what why. I never configured dependabot to have anything in common with pypi.

webknjaz:
@Bystroushaak those issues have logs, take a look at them if you are interested what Python package dependencies it detected. It is a GitHub App, Apps only have privileges to post in repos if you install them.

mjpieters:
[...]

PR #2557, opened 2020-09-25
Default to pypi.org instead of pypi.python.org

jurre:
This causes unnecessary redirects, as per https://github.com/dependabot/dependabot-core/issues/2550#issuecomment-698410412


brrygrdn (approved):
👍🏻

feelepxyz (approved):
🙌

PR #2558, opened 2020-09-25
Add a default User-Agent header on excon requests

feelepxyz:
Adding a `User-Agent` header to excon requests including the excon
version, ruby version and ruby platform:

```
dependabot/version excon/version ruby/version (ruby_platform) (+https://github.com/dependabot/dependabot-core)
```

Linking to dependabot-core repo for contact.


cargo/lib/dependabot/cargo/metadata_finder.rb
```diff
@@ -52,7 +52,6 @@ def crates_listing
 
         response = Excon.get(
           "https://crates.io/api/v1/crates/#{dependency.name}",
-          headers: { "User-Agent" => "Dependabot (dependabot.com)" },
```
jurre, line 55: Should we also add the new headers here?

feelepxyz: The new header should get added along with `**SharedHelpers.excon_defaults` - wonder if cargo have any hard-coded rues matching on this? Will try and see where this came from.

jurre: Ah yes ofc 👍

common/lib/dependabot/shared_helpers.rb
```diff
@@ -13,6 +13,9 @@ module SharedHelpers
     BUMP_TMP_FILE_PREFIX = "dependabot_"
     BUMP_TMP_DIR_PATH = "tmp"
     GIT_CONFIG_GLOBAL_PATH = File.expand_path("~/.gitconfig")
+    USER_AGENT = "dependabot/#{Dependabot::VERSION} #{Excon::USER_AGENT} "\
+                 "ruby/#{RUBY_VERSION} (#{RUBY_PLATFORM}) "\
+                 "(+https://github.com/dependabot/dependabot-core)"
```
thepwagner, line 18: Nit: this contact is fuzzy...

GitHub support would probably have better urgent response/escalation policies for managed Dependabot, but be _really_ confused if the problematic traffic is originating from someone using `dependabot-core` on a rando Jenkins.

So meh as this is the best idea, and any reference is better than "i am excon".

thepwagner (approved):
LGTM, just 🚴🏠 on the string.

The have+eat option maybe to allow injecting the user-agent via environment variable?
That way GitHub's hosted versions of dependabot-core can customize.

common/lib/dependabot/shared_helpers.rb
feelepxyz: Yeah also worried things might get lost/take a while through github support tbh

PR #2569, opened 2020-09-28
Handle empty pipfile requirement string

feelepxyz:
Dependabot doesn't allow the `Dependabot::Dependency`
requirements.requirement string to be a blank string.

Pip treats `""` equal to `"*"` so we can set this when building the
`Dependabot::Dependency` object.

### DIFF
--- a/python/spec/fixtures/pipfiles/empty_requirement
+++ b/python/spec/fixtures/pipfiles/empty_requirement
@@ -1,0 +1,8 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+tenacity = "*"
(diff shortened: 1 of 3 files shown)


brrygrdn (approved):
👍🏻

python/spec/fixtures/pipfiles/empty_requirement
```diff
@@ -0,0 +1,8 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+tenacity = "*"
+tensorflow-gpu = ""
```
feelepxyz, line 8: Tested a dry-run on this and works as expected only updating the lockfile

PR #2571, opened 2020-09-28
go_modules: stub repo_contents_path

thepwagner:
This is an idea to {have,eat} 🍰  in #2551 . Per discussion there, we think that cloning full repositories is our long term direction for `go_modules` updates.

This PR provides a migration path by bootstrapping a temporary git repository from `DependencyFiles[]` if the instantiator of `GoModules::FileUpdater` hasn't provided `repo_contents_path`.

This keeps our internal implementation forward-facing, but provides a migration path during transition.

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
@@ -26,11 +26,12 @@ module Dependabot
         ].freeze
 
         def initialize(dependencies:, credentials:, repo_contents_path:,
-                       directory:)
+                       directory:, tidy:)
           @dependencies = dependencies
           @credentials = credentials
           @repo_contents_path = repo_contents_path
           @directory = directory
+          @tidy = tidy
         end
 
         def updated_go_mod_content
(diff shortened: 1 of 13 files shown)


jurre:
I dig it! Will read up / test tomorrow but great idea IMO

jurre:
Does this assume that we'll revert the changes to the FileFetcher for now?

thepwagner:
> Does this assume that we'll revert the changes to the FileFetcher for now?

Yep ; we'd still want the `FileFetcher` to support grabbing `go.{mod,sum}` for the API, to keep the previous path available during the transition.
Should just be a few `repo_contents_path.nil?` checks? 🤞

thepwagner:
iteration objectives:
- [ ] merge conflicts
- [ ] resolve FileFetcher duality

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -262,6 +265,10 @@ def requirement_to_dependency_obj(req)
         def write_go_mod(body)
           File.write("go.mod", body)
         end
+
+        def tidy?
+          !!@tidy
```
jurre, line 270: Nit: we already pass in a bool right? 🤷‍♂️

jurre (approved):
It's slightly hard to review with this diff, but I think this looks good, nice work!

PR #2576, opened 2020-09-29
Add test coverage reports

brrygrdn:
We are working on some refactors of Dependabot Core and after I introduced  a bug into the 0.19.5 release in https://github.com/dependabot/dependabot-core/pull/2539, I realised the root cause was a test stub bypassing the affected code https://github.com/dependabot/dependabot-core/pull/2543#discussion_r492243627

This PR introduces [simplecov](https://github.com/simplecov-ruby/simplecov) to generate HTML coverage reports locally
Screenshot 2020-09-30 at 12 13 10

 Along with a summary in CI:
Screenshot 2020-09-30 at 12 13 56

It adds a few safety rails for work in progress:
- CI will fail if any builds drop below 80% line coverage / 70% branch coverage
- When working locally, CI will fail if any changes made between test runs drop the coverage

At time of writing, this does not implement a per-file coverage minimum and some specific test suites have their branch coverage lowered to provide a base-line.

The goal as we work is to bring up these values and improve the cover of any low-scoring files so we can turn on per-file coverage checks and bring up branch coverage to ~80%. It remains to be seen if this is a practical goal or will result in a lot of `:nocov:` blocks, so we can revise the targets as we go.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,4 +101,4 @@ jobs:
           docker run --rm "$CORE_CI_IMAGE" bash -c "cd /opt/npm_and_yarn && yarn test"
       - name: Run ${{ matrix.suite }} tests with rspec
         run: |
-          docker run --rm "$CORE_CI_IMAGE" bash -c "cd /home/dependabot/dependabot-core/${{ matrix.suite }} && bundle exec rspec spec"
+          docker run --env "CI=true" --rm "$CORE_CI_IMAGE" bash -c "cd /home/dependabot/dependabot-core/${{ matrix.suite }} && bundle exec rspec spec"
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ vendor
 /dry-run
 **/bin/helper
 /.core-bash_history
+coverage/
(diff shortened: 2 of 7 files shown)


feelepxyz (commented):
This is sweet! Tested this out locally and found the default coverage output is a bit annoying when testing single files/lines, e.g. if I have a failure here `bundle exec rspec spec/dependabot/bundler/file_fetcher_spec.rb:67` the failure will be obscured in console by all the coverage output which comes after the rspec output.

Thoughts on making coverage optional by setting a test `ENV` var on CI and only running coverage when this is set?

brrygrdn:
@feelepxyz That's a good point, yeah, we should swap formatter based on an ENVVAR instead of use both

thepwagner (approved):

jurre (approved):
Awesome 👍

feelepxyz (approved):
Nice one!

brrygrdn:
I've been meaning to ship this, however I noticed the summaries aren't appearing and hadn't had a chance to cycle back to why the `CI` envvar isn't being picked up from actions/passed in to the container

jurre:
> I've been meaning to ship this, however I noticed the summaries aren't appearing and hadn't had a chance to cycle back to why the `CI` envvar isn't being picked up from actions/passed in to the container

I think it's because we just run `docker run` for those tests, we could probably:

```diff
-          docker run --rm "$CORE_CI_IMAGE" bash -c "cd /home/dependabot/dependabot-core/${{ matrix.suite }} && bundle exec rspec spec"
+          docker run --rm "$CORE_CI_IMAGE" -env CI=true bash -c "cd /home/dependabot/dependabot-core/${{ matrix.suite }} && bundle exec rspec spec"
```

PR #2577, opened 2020-09-29
Use project based fixtures fo go_module specs

jurre:
This PR introduces a project based setups for go_modules, meaning that
every test points to a fixture project, which we then turn into a git
repository for each test.

This sets us up for `go mod tidy` support, in which we'll always assume
that the repository has been checked out.


go_modules/spec/dependabot/go_modules/file_updater/go_mod_updater_spec.rb
```diff
@@ -79,18 +81,14 @@
         it { is_expected.to include(%(rsc.io/quote v1.5.2\n)) }
 
         context "when a replace directive is present" do
-          let(:go_mod_body) do
-            fixture("go_mods", go_mod_fixture_name) +
-              "\nreplace github.com/fatih/Color => ../../../../../../foo"
```
thepwagner, line 84: Noice: baking these into the fixtures is magic--.

go_modules/spec/dependabot/go_modules/file_updater/go_mod_updater_spec.rb
```diff
@@ -201,14 +178,9 @@
           describe "a dependency that has no top-level package" do
             let(:dependency_name) { "github.com/prometheus/client_golang" }
             let(:dependency_version) { "v0.9.3" }
-            let(:go_mod_body) do
-              fixture("go_mods", go_mod_fixture_name).sub(
-                "rsc.io/quote v1.4.0",
-                "github.com/prometheus/client_golang v0.9.3"
-              )
-            end
+            let(:project_name) { "no_top_level_package" }
```
feelepxyz, line 181: So much nicer! 💯

thepwagner (approved):
Idea: maybe `go_modules/spec/fixtures/projects $ for i in *; do cd $i; go mod download; cd ..; done` 
AFAICT go_sum/go.sum maybe out of sync?

`for i in *; do cd $i; go build -o /dev/null; cd ..; done`, looks like the included `main.go`s are invalid too - do we gain anything from working examples?.

Noice: we can `foreach` to sanity check these! Could be a useful script (especially if it filters expected error cases).


Scope creep: every project has a `main.go` located in the project root. I'd dig a sample with `.../fixtures/not_root/cmd/main.go`, since we have some weirdness around detecting `main` go files in the root.

jurre:
> Scope creep: every project has a main.go located in the project root. I'd dig a sample with .../fixtures/not_root/cmd/main.go, since we have some weirdness around detecting main go files in the root.

Yeah good shout, I think we should add that in #2551, as the `main.go` files are actually not used in this iteration yet (we generate a dummy file for each update). I opted to already add them here, to make that diff smaller but I can also remove them here and introduce them in that PR.

> Idea: maybe go_modules/spec/fixtures/projects $ for i in *; do cd $i; go mod download; cd ..; done

We don't actually want a `go.sum` in all of these, based on the historical test setup at least, but we can probably check if it's present. Are you thinking we should do this in CI or add a convenience script?

feelepxyz (approved):
Nice improvement! ✨

thepwagner:
> Are you thinking we should do this in CI or add a convenience script?

I was thinking a convenience script. Easy punt as it's only really useful _once_.
tbh I just wanted a 🤖 way to verify `**/go.sum`s were ~sane in to  ✅ this PR.

> We don't actually want a go.sum in all of these,

💯 ; the above added some untracked `go.sum`s that I ignored - the out of date path `go_mod/go.sum` was the only 🤔 .

jurre:
> Scope creep: every project has a main.go located in the project root. I'd dig a sample with .../fixtures/not_root/cmd/main.go, since we have some weirdness around detecting main go files in the root.

@thepwagner I added a simple case here: https://github.com/dependabot/dependabot-core/pull/2551/commits/b555c7b6ce4ebe1ebd4f9894dfe98a236d95192a

PR #2578, opened 2020-09-29
FileFetcher::Base.repo_contents_path

thepwagner:
This extends the `repo_contents_path` attribute that was added to `FileParser`+`FileUpdater` to `FileFetcher`.

By making the `FileFetcher` aware of the target directory for a checkout at construction time:
* we can leverage this checkout in `#dependency_files`, to avoid double fetching (as demonstrated by _jurre in https://github.com/dependabot/dependabot-core/pull/2551/commits/77857ec217451287b017fc3b112be56cc119299b#diff-d76f068189ee62d3127d9dbc19dcf4d0R51-R65)
* avoid passing `target_directory` at runtime, as it's baked in

# Related
- Implements https://github.com/dependabot/dependabot-core/pull/2476#discussion_r476390866 
- Slice of https://github.com/dependabot/dependabot-core/pull/2551
- Continues #2461


common/spec/dependabot/file_fetchers/base_spec.rb
```diff
@@ -1313,4 +1319,82 @@ def fetch_files
...
+    # prepare a checkout to spoof `git clone`
+    let(:contents) { fixture("ruby", "gemfiles", "Gemfile") }
+    let(:mock_clone) do
+      `git init #{repo_contents_path}`
+      path = File.join(repo_contents_path, "requirements.txt")
+      File.write(path, contents)
+      `git -C #{repo_contents_path} add .`
+      `git -C #{repo_contents_path} commit -m'fake clone'`
+      repo_contents_path
+    end
```
jasonrudolph, line 1334: Since the `"file not found"` context and the `"symlink"` context define their own `let(:mock_clone`) block, I think that means that this block only applies to the `"with a GitHub source` context. How would you feel about moving this let block into the `"with a GitHub source` context to make it easier to see that it is specific to that context?

jasonrudolph, line 1334: From what I can tell, it seems like we're mostly using this let block for its side effects, as opposed to using it to set a variable. With that in mind, I recommend using a [`before(:example)`](https://rspec.info/documentation/3.9/rspec-core/RSpec/Core/Hooks.html#before-instance_method) instead:

```suggestion
    before(:example) do
      `git init #{repo_contents_path}`
      path = File.join(repo_contents_path, "requirements.txt")
      File.write(path, contents)
      `git -C #{repo_contents_path} add .`
      `git -C #{repo_contents_path} commit -m'fake clone'`
    end
```

That would also allow us to remove the `repo_contents_path` return value. What do you think?

jasonrudolph (commented):
Just a couple recommendations regarding the test setup. I hope this helps.

common/spec/dependabot/file_fetchers/base_spec.rb
thepwagner: I'd assumed `mock_clone` would be lazily evaluated, so `and_return(mock_clone)` below would mean this checkout doesn't exist until `FileFetcher#_clone_repo_contents` is called.

Unconfirmed assumption, but it felt pretty slick.

thepwagner: I didn't like the idea of the `before` block in this context referencing `mock_clone` that only existed in child contexts.

Per the other discussion thread though, `mock_clone` is attempted cleverness that should likely be reworked.

thepwagner: I've updated to DRY this: 
* `mock_clone` still exists here, in the context it's used
* `fill_repo` is introduced to do the custom bits between each context

`fill_repo` might be a little too close to the flagged pattern of abusing `let` for side effects though, what do you think?

jasonrudolph: I dig those updates, @thepwagner. 👍 

The changes overall look good to me, but I don't feel like I know enough about the big picture yet to offer a meaningful code review approval, so I'd like to give @jurre a chance to review it.

common/spec/dependabot/file_fetchers/base_spec.rb
```diff
@@ -1313,4 +1319,79 @@ def fetch_files
...
+    let(:mock_clone) do
+      `git init #{repo_contents_path}`
+      fill_repo
+      `git -C #{repo_contents_path} add .`
+      `git -C #{repo_contents_path} commit --allow-empty -m'fake clone'`
+      repo_contents_path
+    end
+    before do
+      allow(file_fetcher_instance).
+        to receive(:_clone_repo_contents).and_return(mock_clone)
```
jurre, line 1343: Slightly worried about stubbing the system under test, we could alternatively put this [in the `tmp` path expected by the FileFetcher](https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/file_fetchers/base.rb#L433), but it does make some assumptions about the implementation, so I'm not sure if it's a net win. Might also be an opportunity to extract an object (`RepoCloner` ?) that we could inject.

No need to change anything just wanted to bring it up and make sure we're OK stubbing the class being tested.

jurre (approved):
This is great! Few small nits.

One thing I was thinking about, is if we could (in the future) have a `FileFetchers::Git` class that doesn't inherit from `FileFetchers::Base`? The class is quite big and does a lot of things, potentially a class dedicated to just fetching files from git could be a lot simpler. This may just be a pipedream, I haven't really thought about how it could work.

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -91,7 +95,34 @@ def fetch_file_if_present(filename, fetch_submodules: false)
...
+      end
+
       def fetch_file_from_host(filename, type: "file", fetch_submodules: false)
+        unless @repo_contents_path.nil?
+          return load_cloned_file_if_present(filename)
+        end
+
         path = Pathname.new(File.join(directory, filename)).cleanpath.to_path
         content = _fetch_file_content(path, fetch_submodules: fetch_submodules)
         type = @linked_paths.key?(path.gsub(%r{^/}, "")) ? "symlink" : type
```
feelepxyz, line 128: Looks like `repo_contents` is also used from go_modules and other places to list a directories contents

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -91,7 +95,34 @@ def fetch_file_if_present(filename, fetch_submodules: false)
         raise Dependabot::DependencyFileNotFound, path
       end
 
+      def load_cloned_file_if_present(filename)
+        path = File.join(clone_repo_contents, filename)
```
feelepxyz, line 99: Might also need to join in the directory here if it's set, something like:

```suggestion
        path = File.join(clone_repo_contents, directory, filename)
```

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -91,7 +95,34 @@ def fetch_file_if_present(filename, fetch_submodules: false)
...
+        unless File.exist?(path)
+          raise Dependabot::DependencyFileNotFound, filename
```
feelepxyz, line 101:
```suggestion
          raise Dependabot::DependencyFileNotFound, path
```

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -91,7 +95,34 @@ def fetch_file_if_present(filename, fetch_submodules: false)
...
+        content = File.read(path)
+        type = if File.symlink?(path)
+                 symlink_target = File.readlink(path)
+                 "symlink"
+               else
+                 "file"
+               end
+
+        DependencyFile.new(
+          name: Pathname.new(filename).cleanpath.to_path,
```
feelepxyz, line 113:
```suggestion
          name: path
```

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -91,7 +95,34 @@ def fetch_file_if_present(filename, fetch_submodules: false)
...
 
+      def load_cloned_file_if_present(filename)
+        path = File.join(clone_repo_contents, filename)
+        unless File.exist?(path)
+          raise Dependabot::DependencyFileNotFound, filename
+        end
+
+        content = File.read(path)
+        type = if File.symlink?(path)
+                 symlink_target = File.readlink(path)
```
feelepxyz, line 106:
```suggestion
        content = File.read(repo_path)
        type = if File.symlink?(repo_path)
                 symlink_target = File.readlink(repo_path)
```

common/lib/dependabot/file_fetchers/base.rb
thepwagner: I think `filename` is correct here.

The user expects `requirements.txt`, not `/tmp/tmp-repo_contents_path/directory/requirements.txt`

I could dig `File.join(directory, path)`, but I'm not sure it's required? I'll add a test case (the current one is specific to `requirements.txt`)

common/lib/dependabot/file_fetchers/base.rb
thepwagner: Same as above: I think this should be relative to the repo root, but `path`  that we've assembled is absolute.

Only siths deal in absolutes. :trollface:

common/lib/dependabot/file_fetchers/base.rb
thepwagner: Can you expand on the implications of this?

feelepxyz: Oh the existing error case for `repo_contents` is really hard to parse. Submodule handing stands out as a painful thing, do we need to start calling something like `git submodule update --init --recursive` to get all the submodule content checked out?

common/spec/dependabot/file_fetchers/base_spec.rb
thepwagner: Alternatively, we could inject a `file://` based `source.url` and actually invoke the `git clone`.

That's how I did tests in $unlinkable - a local "upstream" + "downstream" clone for end-to-end. WDYT about that?

common/lib/dependabot/file_fetchers/base.rb
feelepxyz: Ah yes you're right!

common/lib/dependabot/file_fetchers/base.rb
feelepxyz: > Can you expand on the implications of this?

Should we `ls` the contents of the folder in the clone instead of fetching from github?

thepwagner:
> One thing I was thinking about, is if we could (in the future) have a FileFetchers::Git class that doesn't inherit from FileFetchers::Base?

💯 ; aligns with [this comment](https://github.com/dependabot/dependabot-core/pull/2551#discussion_r496049697) - it'd be noice if instead of registering a `FileFetcher`, then registering that `FileFetcher` as `always_clone`... you could just register a `FileFetcher` that does the needful.

common/lib/dependabot/file_fetchers/base.rb
feelepxyz: Looks like the existing cases of `Dependabot::DependencyFileNotFound` raise with `Pathname.new(File.join(directory, filename)).cleanpath.to_path`

common/lib/dependabot/file_fetchers/base.rb
feelepxyz: > Submodule handing stands out as a painful thing, do we need to start calling something like `git submodule update --init --recursive` to get all the submodule content checked out?

Looks like we don't fetch submodules from any of the go modules so maybe not worth implementing that now, would be good to add a comment explaining this functionality is missing from `load_cloned_file_if_present`.

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -68,14 +68,18 @@ def commit
       end
 
       # Returns the path to the cloned repo
-      def clone_repo_contents(target_directory: nil)
+      def clone_repo_contents
         @clone_repo_contents ||=
-          _clone_repo_contents(target_directory: target_directory)
+          _clone_repo_contents(target_directory: repo_contents_path)
```
feelepxyz, line 73: Looks like we are not be handing Octokit 404's when cloning the repo in `_clone_repo_contents`, might rarely end up there but wondering if we should handle it now that we'll be relying on it more widely. Think you want to rescue `SharedHelpers::HelperSubprocessFailed` and re-raise `Dependabot::RepoNotFound`

common/spec/dependabot/file_fetchers/base_spec.rb
jurre: I like it! @jasonrudolph and I discussed something similar for #2551 yesterday

common/spec/dependabot/file_fetchers/base_spec.rb
thepwagner: I think this concern (relying on the side-effect) is 💯 in the latest rev.

common/lib/dependabot/file_fetchers/base.rb
thepwagner: > would be good to add a comment explaining this functionality is missing from load_cloned_file_if_present

I aimed for that in https://github.com/dependabot/dependabot-core/pull/2578/commits/a98c4ddfe65ec25f82fe08990272cef612a1f6d8#diff-a619536a45c6fc217c2f5e4c56460121R35-R43 - note the limitations and punt.

I think the way [`Dir.glob`](https://github.com/dependabot/dependabot-core/pull/2551/files#diff-d76f068189ee62d3127d9dbc19dcf4d0R74) is used in the eventual downstream where this is to be enabled, the `GoModules::FileFetcher` can stop calling `repo_contents` as soon as it can `repo_contents_path.nil?` to know the new glob path will work.

feelepxyz: > GoModules::FileFetcher can stop calling repo_contents as soon as it can repo_contents_path.nil? to know the new glob path will work.

Gotcha, that's way nicer!

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -91,7 +104,35 @@ def fetch_file_if_present(filename, fetch_submodules: false)
         raise Dependabot::DependencyFileNotFound, path
       end
 
+      def load_cloned_file_if_present(filename)
+        path = Pathname.new(File.join(directory, filename)).cleanpath.to_path
+        repo_path = File.join(clone_repo_contents, path)
+        unless File.exist?(repo_path)
+          raise Dependabot::DependencyFileNotFound, filename
```
feelepxyz, line 111: Was this missed?
```suggestion
          raise Dependabot::DependencyFileNotFound, path
```

common/spec/dependabot/file_fetchers/base_spec.rb
```diff
@@ -1313,4 +1319,77 @@ def fetch_files
...
+
+    # `git clone` against a file:// URL that is filled by the test
+    let(:repo_path) { Dir.mktmpdir }
+    after { FileUtils.rm_rf(repo_path) }
+    let(:fill_repo) {}
+    before do
+      `git init #{repo_path}`
+      fill_repo
+      `git -C #{repo_path} add .`
+      `git -C #{repo_path} commit --allow-empty -m'fake clone source'`
```
feelepxyz, line 1335: Nit: Could wrap this in `Dir.chdir` and change to `File.write("requirements.txt", contents)` in fill_repo
```suggestion
      Dir.chdir(repo_path) do
        `git init .`
        fill_repo
        `git add .`
        `git commit --allow-empty -m'fake clone source'`
      end
```

common/spec/dependabot/file_fetchers/base_spec.rb
```diff
@@ -1313,4 +1319,77 @@ def fetch_files
...
+      end
+
+      context "file not found" do
+        it "raises DependencyFileNotFound" do
+          expect { subject }.
+            to raise_error(Dependabot::DependencyFileNotFound) do |error|
+            expect(error.file_path).to eq("requirements.txt")
+          end
+        end
+      end
```
feelepxyz, line 1370: Thoughts on adding the above test cases when specifying a `directory`?

feelepxyz (commented):
Nice work on this! Added some minor comments, looks good otherwise 💯  Can't wait for the day where we can use this by default in all package managers 🙌

common/spec/dependabot/file_fetchers/base_spec.rb
thepwagner: Noice! The only wrinkle here is that since `contents` is loaded from a fixture, changing the working directory before `fill_repo` loses `./fixtures/...`.

I'm going to patch by not using a fixture for `contents`;  that was just cargo culted from the `github` fixtures above.

feelepxyz: Ah yes missed that, you could also `let!(:contents)` having it execute before the `it` is run

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -91,7 +106,35 @@ def fetch_file_if_present(filename, fetch_submodules: false)
         raise Dependabot::DependencyFileNotFound, path
       end
 
+      def load_cloned_file_if_present(filename)
+        path = Pathname.new(File.join(directory, filename)).cleanpath.to_path
+        repo_path = File.join(clone_repo_contents, path)
```
jurre, line 111: Now that we automatically `clone_repo_contents` when `repo_contents_path` is set, in theory we don't need to call that method from outside anymore right?

I think it makes sense to keep that available as a public method for now (we depend on it in external systems), but it's nice to see that we're moving towards a model where the FileFetcher will just  do the right thing based on the configuration we pass in the constructor.

jurre (approved):

thepwagner:
:x: this changes the behaviour of `fetch_file_if_present()`. 

Previously it would list the repository, then return nil if not found.
As of this rev it raises `DependencyFileNotFound`  in the `repo_contents_path` case.

Caught when integrating with https://github.com/dependabot/dependabot-core/pull/2551 ; fetching `go.sum`s that aren't present.

common/lib/dependabot/file_fetchers/base.rb
thepwagner: 💯 ; I think it makes sense to keep around until `repo_contents` is supported from the clone?

It seems internally useful without `repo_contents`, subclasses want to know where the checkout is for `Dir.glob()`-ing.

feelepxyz (approved):
Great work on this 🙌  tests look 💯

PR #2579, opened 2020-09-29
re-entrant docker-dev-shell

thepwagner:
Modifies `bin/docker-dev-shell` helper to use a fixed container name, and reuse the container when it's already running.

This is a papercut: I'm usually `docker exec`-ing in to check out a tempdir while sitting in a breakpoint.


feelepxyz (approved):
Nice one 👍

jurre (approved):
Oh nice one, I've wanted to do this a few times as well 🙌

PR #2584, opened 2020-09-30
[Bundler] Extract a DependencySource class from LatestVersionFinder

brrygrdn:
This branch extracts a preamble refactor for https://github.com/dependabot/dependabot-core/pull/2552 that drops all the points where we call out to Bundler in `LatestVersionFinder` into a subclass.

This leaves `LatestVersionFinder` responsible for filtering versions returned from the `DependencySource` according to `ignored_versions` and `security_advisories` and creates a seam that will allow us to swap out the DependencySource implementation with one that shells out and ultimately returns the same output.

I've left tests as-is apart from adding some to document existing behaviour, I may extract some more specs on the `DependencySource` as a follow-up PR.

### DIFF
--- a/bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
+++ b/bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
@@ -108,10 +108,9 @@ module Dependabot
         end
 
         def dependency_source
-          @dependency_source = DependencySource.new(
+          @dependency_source ||= DependencySource.new(
             dependency: dependency,
             dependency_files: dependency_files,
-            repo_contents_path: repo_contents_path,
             credentials: credentials
           )
         end
--- a/bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
+++ b/bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
@@ -10,14 +10,13 @@ require "dependabot/bundler/update_checker"
 require "dependabot/bundler/requirement"
 require "dependabot/shared_helpers"
 require "dependabot/errors"
+require "dependabot/bundler/update_checker/latest_version_finder/" \
+        "dependency_source"
 
 module Dependabot
   module Bundler
     class UpdateChecker
       class LatestVersionFinder
-        require_relative "shared_bundler_helpers"
-        include SharedBundlerHelpers
-
         def initialize(dependency:, dependency_files:, repo_contents_path: nil,
                        credentials:, ignored_versions:, raise_on_ignored: false,
                        security_advisories:)
@@ -44,11 +43,11 @@ module Dependabot
                     :credentials, :ignored_versions, :security_advisories
 
         def fetch_latest_version_details
-          if dependency_source_type == "git"
-            return latest_git_version_details
+          if dependency_source.git?
+            return dependency_source.latest_git_version_details
           end
 
-          relevant_versions = registry_versions
+          relevant_versions = dependency_source.versions
           relevant_versions = filter_prerelease_versions(relevant_versions)
           relevant_versions = filter_ignored_versions(relevant_versions)
 
@@ -56,9 +55,9 @@ module Dependabot
         end
 
         def fetch_lowest_security_fix_version
-          return if dependency_source_type == "git"
+          return if dependency_source.git?
 
-          relevant_versions = registry_versions
+          relevant_versions = dependency_source.versions
           relevant_versions = filter_prerelease_versions(relevant_versions)
           relevant_versions = filter_vulnerable_versions(relevant_versions)
           relevant_versions = filter_ignored_versions(relevant_versions)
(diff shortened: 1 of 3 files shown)


bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
```diff
@@ -93,71 +92,6 @@ def filter_lower_versions(versions_array)
             select { |version| version > Gem::Version.new(dependency.version) }
         end
 
-        def registry_versions
```
brrygrdn, line 96: This is the main method I wanted to sink into a separate class as it requires strong internal contract on `dependency_source` that will change dramatically

bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
```diff
@@ -44,21 +43,21 @@ def lowest_security_fix_version
                     :credentials, :ignored_versions, :security_advisories
 
         def fetch_latest_version_details
-          if dependency_source.is_a?(::Bundler::Source::Git)
-            return latest_git_version_details
+          if dependency_source.git?
+            return dependency_source.latest_git_version_details
```
brrygrdn, line 47: This is a little bit of a smell, I couldn't find a better way to abstract this this method either directly returns a has obtained from the dependency file via Bundler, or a hash it creates via filtering an array of `Gem::Version` objects.

This seemed like the least obtrusive way to handle it.

bundler/lib/dependabot/bundler/update_checker/latest_version_finder/dependency_source.rb
```diff
@@ -0,0 +1,153 @@
...
+          def git?
+            source_type == GIT
+          end
+
+          private
+
+          def other?
+            source_type == OTHER
+          end
+
```
brrygrdn, line 80: ✂️  Not used

```suggestion
```

bundler/lib/dependabot/bundler/update_checker/latest_version_finder/dependency_source.rb
```diff
@@ -0,0 +1,149 @@
...
+                         dependency_files:,
+                         repo_contents_path: nil,
+                         credentials:)
+            @dependency          = dependency
+            @dependency_files    = dependency_files
+            @repo_contents_path  = repo_contents_path
+            @credentials         = credentials
+          end
+
+          def versions
```
jurre, line 29: How do you feel about adding a small snippet of yard documentation to the public methods that we're introducing? I've often found myself digging into the source code trying to figure out what our objects are meant to be doing, would be nice if we could gradually start introducing some of this:

```suggestion
          # The latest version details for a set of dependencies
          #
          # @return [Array<Hash<Symbol, String>>, Gem::Version]
          def versions
```

bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
jurre: Could we also call `dependency_source.versions` here? It seems like that uses the same `#git?` check to call this method? That way we may be able to hide some more of the implementation?

brrygrdn: Ah, I think that might be a hangover I meant to remove - passing back an array of one hash vs an array of gems means this method would have to check the contents of what it was given which is feels kind of gnarly.

I did briefly think about doing a double return of `versions, commit` to clarify the test in this class, but that seemed like a weird idiom.

bundler/lib/dependabot/bundler/update_checker/latest_version_finder/dependency_source.rb
brrygrdn: Yeah, that's a good call

bundler/lib/dependabot/bundler/update_checker/latest_version_finder/dependency_source.rb
```diff
@@ -0,0 +1,149 @@
...
+            return rubygems_versions if dependency.name == "bundler"
+            return rubygems_versions unless gemfile
+
+            case source_type
+            when OTHER
+              []
+            when GIT
+              latest_git_version_details
```
brrygrdn, line 37: As noted in my comment above, this method should actually return empty for `git` to keep the contract a reflection of `latest_git_version_details` - the methods should be mutually exclusive.

I do think there's a better way though, so will have a thinking

bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
```diff
@@ -174,18 +108,12 @@ def wants_prerelease?
...
-            in_a_temporary_bundler_context do
-              definition = ::Bundler::Definition.build(gemfile.name, nil, {})
-
-              specified_source =
-                definition.dependencies.
-                find { |dep| dep.name == dependency.name }&.source
-
-              specified_source || definition.send(:sources).default_source
-            end
+          @dependency_source = DependencySource.new(
```
feelepxyz, line 111: Memoized?
```suggestion
          @dependency_source ||= DependencySource.new(
```

bundler/lib/dependabot/bundler/update_checker/latest_version_finder/dependency_source.rb
```diff
@@ -0,0 +1,149 @@
...
+          attr_reader :dependency, :dependency_files, :repo_contents_path,
+                      :credentials
+
+          def initialize(dependency:,
+                         dependency_files:,
+                         repo_contents_path: nil,
+                         credentials:)
+            @dependency          = dependency
+            @dependency_files    = dependency_files
+            @repo_contents_path  = repo_contents_path
```
feelepxyz, line 25: ✂️ `repo_contents_path` unused?

feelepxyz (commented):
Nice cleanup! 👍  LGTM, added some minor comments

bundler/lib/dependabot/bundler/update_checker/latest_version_finder/dependency_source.rb
brrygrdn: I think it might be used in the SharedHelpers mixin methods, will double check

brrygrdn: It isn't it can go 👍🏻

bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
brrygrdn: I elected to stick with `#versions` returning an empty array for git to mirror `#latest_git_version_details` returning nil for !git. Adding some yard docs and comments help explain it better than a weird multiple return, etc.

feelepxyz (approved):
💯

PR #2586, opened 2020-10-01
Upgrade pip to 20.2.3

ulgens:


jurre:
It looks like some of the warnings/errors that pip outputs have changed and we're no longer parsing these correctly. I'll try to find some time to dig into those today

ulgens:
I didn't saw https://github.com/dependabot/dependabot-core/pull/2586 before opening this, sorry for duplication.

PR #2587, opened 2020-10-01
Install all python versions on CI

jurre:
Our python test suite takes a really long time, and the suspicion is
that most of that time is spent downloading python for versions that we
don't have installed locally.

Trying to keep the versions that we use in our tests up to date with the
pre-installed python versions is a losing battle, and bundling _all_ of
them in the container that we ship to users means it'll get huge.

But in our CI container it'd be really nice if we didn't have to try and
manually keep these aligned, so this hacky script will simply install
all supported python versions in our CI image.

### DIFF
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -61,5 +61,9 @@ RUN GREEN='\033[0;32m'; NC='\033[0m'; \
     cd $d && bundle install; \
   done
 
+COPY bin/install_python_versions bin/
+COPY python/lib/dependabot/python/python_versions.rb bin/
+RUN ./bin/install_python_versions
+
 RUN git config --global user.name dependabot-ci && git config --global user.email no-reply@github.com
 RUN cd omnibus && bundle install
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -8,6 +8,12 @@ ENV BUNDLE_PATH="/home/dependabot/.bundle" \
 
 COPY .rubocop.yml /home/dependabot/dependabot-core/
 
+# install all supported python versions
+# We do this before pulling in any code for efficient caching layers.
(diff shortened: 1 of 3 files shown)


bin/install_python_versions
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require_relative "./python_versions"
```
thepwagner, line 4: Nit: this is the gnarliest bit to me - the script is useless without the specific `COPY` instructions to stitch these files together in the same path.

Could this use the copy of `python_versions.rb` that we already [copied in](https://github.com/dependabot/dependabot-core/pull/2587/files#diff-5dc00173d2accff302746c5a02269225R47)?
`CODE_DIR` should be available, but the I'm not sure of the ruby incantation to suggest.

thepwagner (approved):
🏎️ as-is, and only affects CI.

bin/install_python_versions
jurre: Yeah good point, I tried getting that version that was copied in, but all my attempts failed. But you're right and I'll try to think of a better solution for this 👍

Dockerfile.ci
```diff
@@ -61,5 +61,9 @@ RUN GREEN='\033[0;32m'; NC='\033[0m'; \
     cd $d && bundle install; \
   done
 
+COPY bin/install_python_versions bin/
+COPY python/lib/dependabot/python/python_versions.rb bin/
```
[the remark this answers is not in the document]
jurre: Yeah I think that's a good suggestion, I don't mind mixing them tbh.

@thepwagner something else that crossed my mind, should we move this up in the Dockerfile to make caching more efficient?

PR #2594, opened 2020-10-02
Extract LatestVersionFinder bundler calls to a native helper

brrygrdn:
Implements the native call-outs for the bundler version finder.

### DIFF
--- a/bundler/helpers/lib/functions/dependency_source.rb
+++ b/bundler/helpers/lib/functions/dependency_source.rb
@@ -20,6 +20,23 @@ module Functions
       type_of(bundler_source)
     end
 
+    def latest_git_version(dependency_source_url:, dependency_source_branch:)
+      setup_bundler
+
+      source = ::Bundler::Source::Git.new(
+        "uri" => dependency_source_url,
+        "branch" => dependency_source_branch,
+        "name" => dependency_name,
+        "submodules" => true
+      )
+
+      # Tell Bundler we're fine with fetching the source remotely
+      source.instance_variable_set(:@allow_remote, true)
+
+      spec = source.specs.first
+      { version: spec.version, commit_sha: spec.source.revision }
+    end
+
     private
 
     def setup_bundler
--- a/bundler/helpers/lib/functions/dependency_source.rb
+++ b/bundler/helpers/lib/functions/dependency_source.rb
@@ -1,0 +1,77 @@
+module Functions
+  class DependencySource
+    attr_reader :gemfile_name, :dependency_name, :dir, :credentials
+
+    RUBYGEMS = "rubygems"
+    PRIVATE_REGISTRY = "private"
+    GIT = "git"
+    OTHER = "other"
+
+    def initialize(gemfile_name:, dependency_name:, dir:, credentials:)
+      @gemfile_name = gemfile_name
+      @dependency_name = dependency_name
+      @dir = dir
+      @credentials = credentials
+    end
+
+    def type
+      setup_bundler
+      bundler_source = specified_source || default_source
+      type_of(bundler_source)
+    end
+
+    private
+
+    def setup_bundler
+      ::Bundler.instance_variable_set(:@root, dir)
+
+      # TODO: DRY out this setup with Functions::LockfileUpdater
+      credentials.each do |cred|
+        token = cred["token"] ||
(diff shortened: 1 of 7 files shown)


bundler/lib/dependabot/bundler/update_checker/shared_bundler_helpers.rb
```diff
@@ -65,6 +77,17 @@ def in_a_temporary_bundler_context(error_handling: true)
           error_handling ? handle_bundler_errors(e) : raise
         end
 
+        # TODO: Use/remove argument
+        def in_a_native_bundler_context(_error_handling: true)
```
brrygrdn, line 81: My thinking on this is that `handle_bundler_errors` _should_ be used as part of this method but it will need the interface changes to accept a string, along with passing back the error class from the native helper. I've punted on this for now as resolving it isn't specific to this component.

bundler/lib/dependabot/bundler/update_checker/latest_version_finder/dependency_source.rb
```diff
@@ -98,46 +103,53 @@ def dependency_rubygems_uri
 
           def private_registry_versions
```
brrygrdn, line 104: Not pictured in the diff, there is a `rubygems_versions` which is the one method that doesn't call out to the native helper as we currently just use Excon to talk to it directly instead of using the `Bundler::Source#fetchers`.

I've left that as-is since moving it to the native helper doesn't give us any gain and leads to a huge test re-write where a lot of webstubbing needs to be replaced/moved. I figured it was better to minimise the diff for this pass.

More generally, I think it might be worth just using bundler's own fetchers to get information even from rubygems and some of this code might lean out, but that seems orthogonal to getting it working.

bundler/helpers/lib/functions/dependency_source.rb
```diff
@@ -0,0 +1,108 @@
...
+        end.
+        map(&:version)
+    end
+
+    private
+
+    def setup_bundler
+      ::Bundler.instance_variable_set(:@root, dir)
+
+      # TODO: DRY out this setup with Functions::LockfileUpdater
```
feelepxyz, line 59: Ah yeah we do this all over

bundler/lib/dependabot/bundler/update_checker/latest_version_finder/dependency_source.rb
```diff
@@ -98,46 +103,53 @@ def dependency_rubygems_uri
...
+              raise Dependabot::PrivateSourceAuthenticationFailure, source
+            elsif e.message.match(BundlerErrorPatterns::HTTP_ERR_REGEX)
+              source = Regexp.last_match(:source)
+              raise Dependabot::PrivateSourceTimedOut, source
+            end
 
-                specified_source =
-                  definition.dependencies.
-                  find { |dep| dep.name == dependency.name }&.source
+            # TODO: Add testing and further exception case handling
```
feelepxyz, line 132: Looks like this is the same handling as `handle_bundler_errors`, where you thinking pair on a refactor of this once this is merged into the bundler 2 spike?

bundler/lib/dependabot/bundler/update_checker/shared_bundler_helpers.rb
feelepxyz: Yeah nice, lets get this in and pair on refactoring the error handling Monday?

bundler/lib/dependabot/bundler/update_checker/latest_version_finder/dependency_source.rb
brrygrdn: Yeah, I was planning to leave that for now and only rescue the _tested_ error cases - we can either pair on it or branch out and do that

feelepxyz (approved):
Nice work on this! 👍  on merging into the spike, looks like the main thing reminding is cleaning up the error handling?

bundler/lib/dependabot/bundler/update_checker/latest_version_finder/dependency_source.rb
brrygrdn: I think when I wrote this I was thinking part of the error handling refactor could be [preparatory](https://martinfowler.com/articles/preparatory-refactoring-example.html) and we could ship (most of it) into master

PR #2596, opened 2020-10-05
Add vendoring support for go_modules

jurre:
This builds on the work we've done for go mod tidy, and adds support
for running `go mod vendor` and committing the results.

We now also expose a `vendor` flag in the `FileUpdater`, in the bundler
implementation this is not yet used, as we implicitly rely on the
`repo_contents_path` being available there. We should start using that
flag there as well, but leaving it as is for now to make upgrading
easier.

This relies on changes in:
- https://github.com/dependabot/dependabot-core/pull/2551
- https://github.com/dependabot/dependabot-core/pull/2598

And needs to be rebased once those changes land.

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater.rb
@@ -3,6 +3,7 @@
 require "dependabot/shared_helpers"
 require "dependabot/file_updaters"
 require "dependabot/file_updaters/base"
+require "dependabot/file_updaters/vendor_updater"
 
 module Dependabot
   module GoModules
@@ -52,6 +53,12 @@ module Dependabot
                 content: file_updater.updated_go_sum_content
               )
           end
+
+          vendor_updater.
+            updated_vendor_cache_files(base_directory: directory).
+            each do |file|
+            updated_files << file
+          end
         end
 
         raise "No files changed!" if updated_files.none?
@@ -79,6 +86,17 @@ module Dependabot
         dependency_files.first.directory
       end
 
+      def vendor_dir
+        File.join(repo_contents_path, directory, "vendor")
+      end
+
+      def vendor_updater
+        Dependabot::FileUpdaters::VendorUpdater.new(
+          repo_contents_path: repo_contents_path,
+          vendor_dir: vendor_dir
+        )
+      end
+
       def file_updater
         @file_updater ||=
           GoModUpdater.new(
@@ -86,8 +104,12 @@ module Dependabot
             credentials: credentials,
             repo_contents_path: repo_contents_path,
             directory: directory,
-            tidy: !@repo_contents_stub && options.fetch(:go_mod_tidy, false)
+            options: { tidy: tidy?, vendor: options.fetch(:vendor, false) }
           )
+      end
+
+      def tidy?
+        !@repo_contents_stub && options.fetch(:go_mod_tidy, false)
       end
     end
   end
--- a/go_modules/lib/dependabot/go_modules/file_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater.rb
@@ -114,7 +114,7 @@ module Dependabot
 
       def vendor?
         File.exist?(File.join(vendor_dir, "modules.txt")) &&
-          options.fetch(:vendor, false)
+          options.fetch(:go_mod_vendor, false)
       end
     end
   end
--- a/go_modules/lib/dependabot/go_modules/file_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater.rb
@@ -104,12 +104,17 @@ module Dependabot
             credentials: credentials,
             repo_contents_path: repo_contents_path,
             directory: directory,
-            options: { tidy: tidy?, vendor: options.fetch(:vendor, false) }
+            options: { tidy: tidy?, vendor: vendor? }
           )
       end
 
       def tidy?
         !@repo_contents_stub && options.fetch(:go_mod_tidy, false)
+      end
+
+      def vendor?
+        File.exist?(File.join(vendor_dir, "modules.txt")) &&
+          options.fetch(:vendor, false)
       end
     end
   end
--- a/go_modules/spec/dependabot/go_modules/file_updater/go_mod_updater_spec.rb
+++ b/go_modules/spec/dependabot/go_modules/file_updater/go_mod_updater_spec.rb
@@ -17,7 +17,7 @@ RSpec.describe Dependabot::GoModules::FileUpdater::GoModUpdater do
       }],
       repo_contents_path: repo_contents_path,
       directory: "/",
-      tidy: tidy
+      options: { tidy: tidy, vendor: false }
     )
   end
 
--- a/go_modules/spec/dependabot/go_modules/file_updater_spec.rb
+++ b/go_modules/spec/dependabot/go_modules/file_updater_spec.rb
@@ -22,6 +22,7 @@ RSpec.describe Dependabot::GoModules::FileUpdater do
   let(:files) { [go_mod, go_sum] }
   let(:project_name) { "go_sum" }
   let(:repo_contents_path) { build_tmp_repo(project_name) }
+  let(:vendor) { false }
 
   let(:credentials) do
     [{
@@ -31,7 +32,7 @@ RSpec.describe Dependabot::GoModules::FileUpdater do
       "password" => "token"
     }]
   end
-  let(:options) { {} }
+  let(:options) { { vendor: vendor } }
 
   let(:go_mod) do
     Dependabot::DependencyFile.new(name: "go.mod", content: go_mod_body)
@@ -110,7 +111,7 @@ RSpec.describe Dependabot::GoModules::FileUpdater do
             credentials: credentials,
             repo_contents_path: repo_contents_path,
             directory: "/",
-            tidy: true
+            options: { tidy: true, vendor: false }
           ).and_return(dummy_updater)
 
         updater.updated_dependency_files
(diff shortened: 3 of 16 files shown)


common/lib/dependabot/file_updaters/vendorable.rb
```diff
@@ -0,0 +1,77 @@
...
+  module FileUpdaters
+    module Vendorable
+      private
+
+      # Returns changed files in the vendor/cache folder
+      #
+      # @param base_directory [String] Update config base directory
+      # @return [Array<Dependabot::DependencyFile>]
+      def updated_vendor_cache_files(base_directory:)
+        return [] unless repo_contents_path && vendor_dir
```
feelepxyz, line 13: This is awesome! Nit style ☕ 🚗 review, thoughts on using this as a service-class/object or passing in `repo_contents_path` `vendor_dir` as args here? Get a bit nervous when mixins reference things from the parent class, so easy to miss

jurre: Yeah, good idea 👍  maybe something like `Dependabot::FileUpdaters::VendorUpdater.new(repo_contents_path:, vendor_dir:)`?

thepwagner:
Could we make the `FileUpdater` refactors separate from these other changes?

Per https://github.com/github/dependabot-updates/issues/673#issuecomment-696644582 , I found `github.com/pkg/errors` to be the perfect package for testing - this test project is pretty yuge: slowing tests down and bloating the repo.

Per https://github.com/github/dependabot-updates/issues/673#issuecomment-696644582 , I found checking `vendor/modules.txt` to be a good indicator that the project _wants_ to vendor. This may be a better default than requiring it be enabled. 
Manually enabling vendoring went out of style with [go1.14](https://golang.org/doc/go1.14#go-command).

Per https://github.com/github/dependabot-updates/issues/673#issuecomment-696644582 , the other features of the linked test case are worth verifying too:
* `modules.txt` should not only be changed, the version string should be updated
* maybe verify file contents were updated to the correct version? (you know the contents of `v1.9.0`, `v1.7.0` and can verify)

Maybe just steal the test case, it's 💯 .


This feels super rushed, ignoring the lessons from earlier experiments.
I'd really dig "go mod vendoring" being a scoped PR that targets `master`, without rejigging base+bundler implementations.

jurre:
> Could we make the FileUpdater refactors separate from these other changes?

Absolutely 👍 

> Per github/dependabot-updates#673 (comment) , I found github.com/pkg/errors to be the perfect package for testing - this test project is pretty yuge: slowing tests down and bloating the repo.

Thanks 💯 you're right

> Per github/dependabot-updates#673 (comment) , I found checking vendor/modules.txt to be a good indicator that the project wants to vendor. This may be a better default than requiring it be enabled.
> Manually enabling vendoring went out of style with go1.14.

Yeah, I was not sure if we should detect vendoring for go modules, as we explicitly have it as a config option for bundler.

> Per github/dependabot-updates#673 (comment) , the other features of the linked test case are worth verifying too:
>
> modules.txt should not only be changed, the version string should be updated
> maybe verify file contents were updated to the correct version? (you know the contents of v1.9.0, v1.7.0 and can verify)
>
> Maybe just steal the test case, it's 💯 .

Will do 👍 

> I'd really dig "go mod vendoring" being a scoped PR that targets master, without rejigging base+bundler implementations.

Targeting `main` was the plan, we just need the changed that are in the `go-mod-tidy` branch. I should have mentioned that this is not intended to be merged into that branch

thepwagner:
> not intended to be merged into that branch

😌 , that's what spooked me this morning: either snowball-ing risk into that branch, or making another diamond mess like #2578 from branch-juggling.

jurre:
@thepwagner I think I've addressed most of you concerns, could you have another :eyes:? Definitely not my intention to rush this through.

infin8x:
@jurre if possible, I'd expect that we check the config file for a `vendor: true` line, [same as Bundler](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates#vendor) - will that work? Presumably, once you set that line, we'd also check for a `vendor_dir` and show an appropriate error message if you don't have one - WDYT?

infin8x:
Also @jurre nice work getting this in!

go_modules/spec/fixtures/projects/vendor/go.mod
```diff
@@ -0,0 +1,7 @@
+module github.com/dependabot/vgotest
+
+go 1.12
+
+require github.com/pkg/errors v0.9.1
+
+replace rsc.io/qr => github.com/rsc/qr v0.2.0
```
thepwagner, line 7:
```suggestion
```

go_modules/spec/fixtures/projects/vendor/go.mod
```diff
@@ -0,0 +1,7 @@
+module github.com/dependabot/vgotest
+
+go 1.12
+
+require github.com/pkg/errors v0.9.1
```
thepwagner, line 5:
```suggestion
require github.com/pkg/errors v0.8.0
```

~Or maybe the `let(:dependency_previous_version) { "v0.8.0" }` lies? I think they should be consistent.~
Edit: more sure this should be `v0.8.0`

go_modules/spec/fixtures/projects/vendor/go.mod
thepwagner: Nit: I'm not sure why this is here.

go_modules/spec/fixtures/projects/vendor/vendor/github.com/pkg/errors/stack.go
```diff
@@ -0,0 +1,177 @@
...
+	if name == "unknown" {
+		return []byte(name), nil
+	}
+	return []byte(fmt.Sprintf("%s %s:%d", name, f.file(), f.line())), nil
+}
+
+// StackTrace is stack of Frames from innermost (newest) to outermost (oldest).
+type StackTrace []Frame
+
+// Format formats the stack of Frames according to the fmt.Formatter interface.
```
thepwagner, line 99: This confirms that `>0.8.0` is what's vendored here.

This makes the `"includes the new source code"` really moot: it's asserting the old code replaced with the new code, but the old code was never there.

thepwagner (approved):
Noice: this scoping with the extracted refactor 😍 .

I'm concerned the fixture is incorrect, :shipit: after that.

go_modules/spec/fixtures/projects/vendor/vendor/github.com/pkg/errors/stack.go
jurre: Nice catch, thanks!

common/lib/dependabot/file_updaters/vendor_updater.rb
```diff
@@ -0,0 +1,84 @@
...
+
+      # Returns changed files in the vendor/cache folder
+      #
+      # @param base_directory [String] Update config base directory
+      # @return [Array<Dependabot::DependencyFile>]
+      def updated_vendor_cache_files(base_directory:)
+        return [] unless repo_contents_path && vendor_dir
+
+        Dir.chdir(repo_contents_path) do
+          relative_dir = vendor_dir.sub("#{repo_contents_path}/", "")
```
jasonrudolph, line 48: Can we reply on `repo_contents_path` _not_ containing a trailing `/`? I ask because if it does include a trailing `/`, then I think this command would be a no-op, and I'm not sure if that's OK or not.

jasonrudolph (commented):
I'm still getting up to speed on these changes, but I noted a couple small questions inline below. I hope this helps.

common/lib/dependabot/file_updaters/vendor_updater.rb
```diff
@@ -0,0 +1,84 @@
...
+        ".json",
+        ".xml",
+        ".toml",
+        ".yaml",
+        ".yml",
+        # docs
+        ".md",
+        ".txt",
+        ".go"
+      ].freeze
```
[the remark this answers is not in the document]
jurre: That's the rebase that's still messed up, these should not be duplicated but moved to the extracted class

common/lib/dependabot/file_updaters/vendor_updater.rb
jurre: Good question, I think we can, this is a static directory that we configure in the runner: https://github.com/github/dependabot-update-job-runner/blob/13c121542a11ffeb7aaa557178f5b8a4568e4b94/guest/fetcher.go#L20

The diff for that refactor is better viewed here: https://github.com/dependabot/dependabot-core/pull/2598/files note that we've been using that relative dir for bundler vendoring support in prod for a while.

I agree it's a little brittle though and there's room for improvement 👍

jurre:
@thepwagner should we be checking for the presence of `vendor/modules.txt` in here as well? We now only look for the feature flag (and assume the vendor folder will be be in `.gitignore`), but I think we should make sure that we really need to be vendoring now that we decided to not make this configurable?

go_modules/lib/dependabot/go_modules/file_updater.rb
```diff
@@ -79,16 +86,36 @@ def directory
...
           )
       end
+
+      def tidy?
+        !@repo_contents_stub && options.fetch(:go_mod_tidy, false)
+      end
+
+      def vendor?
+        File.exist?(File.join(vendor_dir, "modules.txt")) &&
+          options.fetch(:vendor, false)
```
thepwagner, line 117:
```suggestion
          options.fetch(:go_mod_vendor, false)
```

Concern: I think we named this key differently

go_modules/spec/dependabot/go_modules/file_updater/go_mod_updater_spec.rb
```diff
@@ -17,7 +17,7 @@
       }],
       repo_contents_path: repo_contents_path,
       directory: "/",
-      tidy: tidy
+      options: { tidy: tidy, vendor: false }
```
thepwagner, line 20: Nit: test scoping feels a little off given where the new functionality is.

Meh as `FileUpdater` spec covers the right bits and they're still ~maybe combineable.

go_modules/spec/dependabot/go_modules/file_updater_spec.rb
```diff
@@ -148,5 +162,91 @@
...
+          groups: [],
+          source: {
+            type: "default",
+            source: "github.com/pkg"
+          }
+        }]
+      end
+
+      it "updates the version" do
+        expect(go_mod_body).to include("github.com/pkg/errors v0.8.0")
```
thepwagner, line 197: Noice: this pre-flight check.

go_modules/spec/dependabot/go_modules/file_updater_spec.rb
```diff
@@ -148,5 +162,91 @@
...
+          requirement: dependency_previous_version,
+          groups: [],
+          source: {
+            type: "default",
+            source: "github.com/pkg"
+          }
+        }]
+      end
+
+      it "updates the version" do
```
thepwagner, line 196:
```suggestion
      it "updates the go.mod" do
```

Nit: I think that's closer to what it's test is up to?

go_modules/spec/dependabot/go_modules/file_updater_spec.rb
```diff
@@ -148,5 +162,91 @@
...
+          )
+        )
+      end
+
+      it "updates the vendor/modules.txt file to the right version" do
+        modules_file = updater.updated_dependency_files.find do |file|
+          file.name == "vendor/modules.txt"
+        end
+
+        expect(modules_file.content).to include "github.com/pkg/errors v0.9.1"
```
thepwagner, line 229:
```suggestion
        expect(modules_file.content).to_not include "github.com/pkg/errors v0.8.0"
        expect(modules_file.content).to include "github.com/pkg/errors v0.9.1"
```

I think? (and I think this is independent of `vendoring`  - there can be only one)

thepwagner (approved):
Noooooice. I think the name of the `FileUpdater.new(options:)` key for vendoring needs to be fixed, everything else is 🚲🏠

jurre:
Ah yes, cheers for fixing that 🎉

PR #2597, opened 2020-10-05
Allow passing tidy as option to go_modules updater

jurre:
This PR targets #2551 

This allows us to pass configuration options to Updater classes,
starting with `tidy` for gomod. Current plans are for go mod tidy
support to be default and not configurable, but in order to roll these
changes out gradually we temporarily allow passing it as an option.

These options can be used to conditionally enable features in the
future, like for instance vendoring.

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater.rb
@@ -86,7 +86,7 @@ module Dependabot
             credentials: credentials,
             repo_contents_path: repo_contents_path,
             directory: directory,
-            tidy: options.fetch(:tidy, false)
+            tidy: !@repo_contents_stub && options.fetch(:tidy, false)
           )
       end
     end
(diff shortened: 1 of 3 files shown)


feelepxyz (approved):

go_modules/lib/dependabot/go_modules/file_updater.rb
```diff
@@ -86,7 +86,7 @@ def file_updater
             credentials: credentials,
             repo_contents_path: repo_contents_path,
             directory: directory,
-            tidy: !@repo_contents_stub
+            tidy: options.fetch(:tidy, false)
```
thepwagner, line 89: Concern: this should still respect `@repo_contents_stub`.

```suggestion
            tidy: options.fetch(:tidy, !@repo_contents_stub)
```

In the rejected rev from Friday this just set `#tidy` after the guard in the constructor, I think this is the equivalent?

jurre: Ah, good catch, thanks!

jasonrudolph (approved):
🚀

thepwagner (approved):

PR #2598, opened 2020-10-05
Refactor: reusable VendorDependencies object

jurre:
This pulls the vendoring logic out of the bundler updater, so we can
reuse it for other ecosystems as well.

Isolated changes from https://github.com/dependabot/dependabot-core/pull/2596


feelepxyz (approved):
Nice one!

thepwagner (approved):
Could any of https://github.com/dependabot/dependabot-core/blob/7782c22486b21f84acb04e9085e9d7d3ca9c17d9/bundler/spec/dependabot/bundler/file_updater_spec.rb#L1546-L1708 be reused to build a `vendor_updater_spec.rb`?

I'd dig unit testing this new unit.

jurre:
> Could any of https://github.com/dependabot/dependabot-core/blob/7782c22486b21f84acb04e9085e9d7d3ca9c17d9/bundler/spec/dependabot/bundler/file_updater_spec.rb#L1546-L1708 be reused to build a `vendor_updater_spec.rb`?
> 
> 
> 
> I'd dig unit testing this new unit.

Will try to add some unit tests 👍

common/spec/dependabot/file_fetchers/vendor_updater_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+  let(:project_name) { "vendor_gems" }
+  let(:repo_contents_path) { build_tmp_repo(project_name) }
+
+  let(:updated_files) do
+    updater.updated_vendor_cache_files(base_directory: repo_contents_path)
+  end
+
+  before do
+    in_cloned_repository(repo_contents_path) do
+      `mv vendor/cache/business-1.4.0.gem vendor/cache/business-1.5.0.gem`
```
jurre, line 24: Bit of cheating here, but I didn't want these tests to depend on any specific updater implementation, so just testing that updated files in the right folder are picked up

thepwagner:
```suggestion
      # change a vendor file like an updater would
      `mv vendor/cache/business-1.4.0.gem vendor/cache/business-1.5.0.gem`
```

Idea: encode that magic.

thepwagner (approved):

PR #2600, opened 2020-10-05
Extract VersionResolver to a native helper

brrygrdn:


brrygrdn:
There should be 4 test failures, all cases are where we were previously hitting webmocks but are now hitting rubygems for real.

The are mostly testing the exception block within `fetch_latest_resolvable_version_details` so I think they should be ok to mock out.

bundler/helpers/lib/functions/version_resolver.rb
```diff
@@ -0,0 +1,206 @@
...
+
+      # TODO: Rewrite this
+      #
+      # If the dependency wasn't found in the definition, but *is*
+      # included in a gemspec, it's because the Gemfile didn't import
+      # the gemspec. This is unusual, but the correct behaviour if/when
+      # it happens is to behave as if the repo was gemspec-only.
+      if dep.nil? && dependency_requirements.any?
+        return "latest"
+      end
```
brrygrdn, line 34: This is a bad signal to pass back, I took a shot at just passing back nil and then examining `dependency_requirements.any?` in the calling code, but it broke several tests surprisingly.

Will review now the rest of the refactor is relatively sane.

bundler/lib/dependabot/bundler/update_checker/shared_bundler_helpers.rb
```diff
@@ -77,15 +77,18 @@ def in_a_temporary_bundler_context(error_handling: true)
...
+        def in_a_native_bundler_context(error_handling: true)
           SharedHelpers.
             in_a_temporary_repo_directory(base_directory,
                                           repo_contents_path) do |_tmp_dir|
             write_temporary_dependency_files
 
             yield
           end
+        rescue Dependabot::SharedHelpers::HelperSubprocessFailed => e
+          # TODO: Add retrying
```
brrygrdn, line 89: This should probably implement the same retrying logic as the helper it is replacing, I just wanted to check the definition of retry-able errors still applied.

feelepxyz: I've got this mostly done locally so can merge my changes on top of this once it's merged

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
```diff
@@ -74,8 +74,6 @@
   describe "#latest_resolvable_version_details" do
     subject { resolver.latest_resolvable_version_details }
 
-    include_context "stub rubygems compact index"
```
feelepxyz, line 77: 💯

bundler/lib/dependabot/bundler/update_checker/version_resolver.rb
```diff
@@ -77,50 +75,42 @@ def fetch_latest_resolvable_version_details
...
-
-              # If the old Gemfile index was used then it won't have checked
-              # Ruby compatibility. Fix that by doing the check manually (and
-              # saying no update is possible if the Ruby version is a mismatch)
-              next nil if ruby_version_incompatible?(dep)
-
-              details = { version: dep.version }
-              if dep.source.instance_of?(::Bundler::Source::Git)
-                details[:commit_sha] = dep.source.revision
+            in_a_native_bundler_context(error_handling: false) do |tmp_dir|
```
feelepxyz, line 78: Any reason this isn't handling errors like before?

```suggestion
            in_a_native_bundler_context do |tmp_dir|
```

feelepxyz (approved):

bundler/helpers/lib/functions/version_resolver.rb
```diff
@@ -0,0 +1,206 @@
...
+      # running on, rather than the true latest resolvable version).
+      return nil if dep.name == "bundler"
+
+      # If the old Gemfile index was used then it won't have checked
+      # Ruby compatibility. Fix that by doing the check manually (and
+      # saying no update is possible if the Ruby version is a mismatch)
+      return nil if ruby_version_incompatible?(dep)
+
+      details = { version: dep.version }
+      if dep.source.instance_of?(::Bundler::Source::Git)
```
jurre, line 51: I suppose we don't need to force these to be top-level constants anymore?

```suggestion
      if dep.source.instance_of?(Bundler::Source::Git)
```

The reason I bring it up is because I was looking around for a local `Bundler` namespace in these helpers. But I'm also fine with keeping it as is.

Related: how come we have access to `Bundler` without requiring it? Do we always run this via `bundle exec` or similar?

feelepxyz: > Related: how come we have access to Bundler without requiring it? Do we always run this via bundle exec or similar?

Yeah running it here https://github.com/dependabot/dependabot-core/pull/2552/files#diff-8e18083192f4b551e74d7775819bb0a1R7

jurre:
Still reading through this a bit more carefully, but overall looks great

bundler/lib/dependabot/bundler/update_checker/version_resolver.rb
brrygrdn: Yeah, in this case there are a few errors the version resolver wants to handle first, if it doesn't handle them, it manually hands off to `handle_bundler_errors` on line [105](https://github.com/dependabot/dependabot-core/pull/2600/files/419713075e150d03c506c8f98fd29cde43f0f2f3#diff-3092b190ebf8948a50c8d770607f2417R105)

I was doing it differently up until [this commit](https://github.com/dependabot/dependabot-core/pull/2600/commits/419713075e150d03c506c8f98fd29cde43f0f2f3) but having the rescues inside the block lead to having to manage the temp manifest files being re-written after dropping a `.gemspec` which was hard to follow.

brrygrdn:
Merging into the refactor spike 🏀

PR #2604, opened 2020-10-06
Allow requirements.txt files of up to 200kb

jurre:
We have a limit on the size of .txt files that we fetch for Python
repositories, most likely because we try to fetch _all_ files with that
extension, and the GitHub contents API only allows fetching files up to
1mb.

Some users are now running into the 100kb size limit that we've set, and
are getting back a pretty opaque `DependencyFileNotFound` error.

In order to resolve the issue for the cases that have been reported,
increasing to 200kb seems like a reasonable first step.

@greysteil you added the original 100kb size limit: https://github.com/dependabot/dependabot-core/commit/86c769c096b57ca41b5c93f939f43ca0ca3b0db8. Do you happen to remember if what I've typed up here is indeed the reason, or is there another reason we cap at 100kb?

### DIFF
--- a/python/spec/fixtures/github/contents_python_large_requirements_txt.json
+++ b/python/spec/fixtures/github/contents_python_large_requirements_txt.json
@@ -1,0 +1,18 @@
+[
+  {
+    "name": "requirements.txt",
+    "path": "requirements.txt",
+    "sha": "88b4e0a1c8093fae2b4fa52534035f9f85ed0956",
+    "size": 5000000,
+    "url": "https://api.github.com/repos/CityOfNewYork/NYCOpenRecords/contents/requirements.txt?ref=develop",
+    "html_url": "https://github.com/CityOfNewYork/NYCOpenRecords/tree/develop/requirements.txt",
+    "git_url": "https://api.github.com/repos/CityOfNewYork/NYCOpenRecords/git/trees/88b4e0a1c8093fae2b4fa52534035f9f85ed0956",
+    "download_url": "https://raw.githubusercontent.com/CityOfNewYork/NYCOpenRecords/develop/requirements.txt",
+    "type": "file",
+    "_links": {
+      "self": "https://api.github.com/repos/CityOfNewYork/NYCOpenRecords/contents/requirements.txt?ref=develop",
+      "git": "https://api.github.com/repos/CityOfNewYork/NYCOpenRecords/git/trees/88b4e0a1c8093fae2b4fa52534035f9f85ed0956",
+      "html": "https://github.com/CityOfNewYork/NYCOpenRecords/tree/develop/requirements.txt"
+    }
+  }
+]
--- a/python/lib/dependabot/python/file_fetcher.rb
+++ b/python/lib/dependabot/python/file_fetcher.rb
@@ -166,7 +166,7 @@ module Dependabot
         repo_contents.
           select { |f| f.type == "file" }.
           select { |f| f.name.end_with?(".txt", ".in") }.
-          reject { |f| f.size > 100_000 }.
+          reject { |f| f.size > 200_000 }.
           map { |f| fetch_file_from_host(f.name) }.
           select { |f| requirements_file?(f) }.
           each { |f| @req_txt_and_in_files << f }
@@ -186,7 +186,7 @@ module Dependabot
         repo_contents(dir: relative_reqs_dir).
           select { |f| f.type == "file" }.
           select { |f| f.name.end_with?(".txt", ".in") }.
-          reject { |f| f.size > 100_000 }.
+          reject { |f| f.size > 200_000 }.
           map { |f| fetch_file_from_host("#{relative_reqs_dir}/#{f.name}") }.
           select { |f| requirements_file?(f) }
       end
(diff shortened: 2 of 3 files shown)


jurre:
I'll add a test for this to document the behavior a little more.

greysteil:
I just figured that it would be pretty unlikely an actual `requirements.txt` file would be over 100kb. Are we finding dependency files that are actually that big?! 🤯

jurre:
> I just figured that it would be pretty unlikely an actual `requirements.txt` file would be over 100kb. Are we finding dependency files that are actually that big?! 🤯

Yep 😄 found one that was 110kb via a support ticket

greysteil (approved):

python/spec/fixtures/github/contents_python_large_requirements_txt.json
```diff
@@ -0,0 +1,418 @@
...
+    "html_url": "https://github.com/CityOfNewYork/NYCOpenRecords/tree/develop/migrations",
+    "git_url": "https://api.github.com/repos/CityOfNewYork/NYCOpenRecords/git/trees/da206a9809330d01f4800718bca2a20c05038b44",
+    "download_url": null,
+    "type": "dir",
+    "_links": {
+      "self": "https://api.github.com/repos/CityOfNewYork/NYCOpenRecords/contents/migrations?ref=develop",
+      "git": "https://api.github.com/repos/CityOfNewYork/NYCOpenRecords/git/trees/da206a9809330d01f4800718bca2a20c05038b44",
+      "html": "https://github.com/CityOfNewYork/NYCOpenRecords/tree/develop/migrations"
+    }
+  },
```
thepwagner, line 353:
```suggestion
```

thepwagner, line 353:
```suggestion
```

thepwagner, line 353: Can we remove this other cruft from the fixture?
The test only needs a big `requirements.txt`, we're burying the important feature.

TIL: GitHub won't let me leave a >300 line suggested change  :trollface:

python/spec/fixtures/github/contents_python_large_requirements_txt.json
```diff
@@ -0,0 +1,418 @@
...
+    "html_url": "https://github.com/CityOfNewYork/NYCOpenRecords/blob/develop/update_definitions.sh",
+    "git_url": "https://api.github.com/repos/CityOfNewYork/NYCOpenRecords/git/blobs/fbde266ea5ebd519b94d18f8f78ab825b02766df",
+    "download_url": "https://raw.githubusercontent.com/CityOfNewYork/NYCOpenRecords/develop/update_definitions.sh",
+    "type": "file",
+    "_links": {
+      "self": "https://api.github.com/repos/CityOfNewYork/NYCOpenRecords/contents/update_definitions.sh?ref=develop",
+      "git": "https://api.github.com/repos/CityOfNewYork/NYCOpenRecords/git/blobs/fbde266ea5ebd519b94d18f8f78ab825b02766df",
+      "html": "https://github.com/CityOfNewYork/NYCOpenRecords/blob/develop/update_definitions.sh"
+    }
+  }
```
thepwagner, line 417:
```suggestion
  }
```

thepwagner (approved):
LGTM, just wondering if we need so much fixture to verify it.

python/spec/fixtures/github/contents_python_large_requirements_txt.json
jurre: Hehe yeah good shout, will remove the other entries locally 👍

Mogost:
> > I just figured that it would be pretty unlikely an actual `requirements.txt` file would be over 100kb. Are we finding dependency files that are actually that big?! 🤯
> 
> Yep 😄 found one that was 110kb via a support ticket. (It does seem like it was generated somehow though)

Yes, this is reality, that was my support ticket. Thanks for fixing this.
The `requirements.txt` file is so large, in part due to the fact that `generate-hashes` parameter is used to create them.
`pip-compile --generate-hashes`

jurre:
> Yes, this is reality, that was my support ticket. Thanks for fixing this.
> The `requirements.txt` file is so large, in part due to the fact that `generate-hashes` parameter is used to create them.
> `pip-compile --generate-hashes`

Ah, interesting, thanks for looping that back! Do you reckon the new 200kb limit would be sufficient for most projects that use this?

Mogost:
> Ah, interesting, thanks for looping that back! Do you reckon the new 200kb limit would be sufficient for most projects that use this?

I am not sure... My `requirements.in` file contains 111 deps, but `requirements.txt` file contains 1531 lines (including hashes for all deps).
I would increase the limit to a 1mb or at least 500kb. 
In any case, it is worth making the error text more understandable, so that if this happens again with someone, he will come here with a problem that will be clear from the very beginning.

jurre:
> I am not sure... My `requirements.in` file contains 111 deps, but `requirements.txt` file contains 1531 lines (including hashes for all deps).
> I would increase the limit to a 1mb or at least 500kb.
> In any case, it is worth making the error text more understandable, so that if this happens again with someone, he will come here with a problem that will be clear from the very beginning.

Yeah I agree and was thinking about that as well, but it's a little tricky due to how it's set up unfortunately 😞 will think about what we can do to improve it.

PR #2606, opened 2020-10-06
Release notes and version bump for v0.121.0

jasonrudolph:
Co-authored-by: Jurre Stender <jurre@github.com>
Co-authored-by: Pete Wagner <thepwagner@github.com>

PR #2605, opened 2020-10-06
Release notes for v0.120.5

jurre:
I liked @jasonrudolph's example of PR'ing these changes

PR #2589, opened 2020-10-01
Prepare CHANGELOG and version bump for 0.120.4

jasonrudolph:
This pull request includes a proposed changelog and version bump for a 0.120.4 release.

PR #2607, opened 2020-10-06
bump-version: push notes to branch for review

thepwagner:
We've taken to double-checking release notes before they're published - lower risk for everyone 🎉 .

This is an attempt at codifying the ~new(?) process.

# Related
- https://github.com/dependabot/dependabot-core/pull/2606
- https://github.com/dependabot/dependabot-core/pull/2605
- https://github.com/dependabot/dependabot-core/pull/2589


bin/bump-version.rb
```diff
@@ -49,7 +49,9 @@
 puts
 puts "git add CHANGELOG.md common/lib/dependabot/version.rb"
 puts "git commit -m 'v#{new_version}'"
-puts "git push origin main"
+puts "git push origin HEAD:v#{new_version}-release-notes"
+puts "... create PR, verify, merge"
```
thepwagner, line 53: A [gh](https://github.com/cli/cli) command to open the PR would be 😎 , but I wasn't 100% what it would be (or if it was a safe scripting assumption).

jurre: I'd dig it and I think `gh pr create` would be the command? Maybe:

```
puts "... create PR, verify, merge, for example:"
puts "gh pr create"
```

jurre (approved):

PR #2612, opened 2020-10-07
dry-run --updater-options=

thepwagner:
This accepts a comma separated list representing the keys in the `options` hash passed to `FileUpdater`s.
It's good enough while that's a bitmap, but won't play nicely if options accept values.

We can use this to toggle all the cool features masked behind `options` in the updater, without needing to turn each into a command line flag.
Maybe we should still accept the list via option instead of yanking it from environment?

# Related
- Continues https://github.com/dependabot/dependabot-core/pull/2597
- Formalized hack from testing https://github.com/dependabot/dependabot-core/pull/2596 + https://github.com/dependabot/dependabot-core/pull/2551

### DIFF
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -389,11 +389,18 @@ def peer_dependencies_can_update?(checker, reqs_to_unlock)
 end
 
 def file_updater_for(dependencies)
+  unless ENV["UPDATER_OPTIONS"].to_s.strip.empty?
+    options = ENV["UPDATER_OPTIONS"].split(",").
+                  map {|o| [o.downcase.to_sym, true] }
+    options = Hash[options]
+  end
+
   Dependabot::FileUpdaters.for_package_manager($package_manager).new(
     dependencies: dependencies,
     dependency_files: $files,
     repo_contents_path: $repo_contents_path,
-    credentials: $options[:credentials]
+    credentials: $options[:credentials],
+    options: options,
   )
 end
 
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -99,7 +99,8 @@ $options = {
   clone: false,
   lockfile_only: false,
   requirements_update_strategy: nil,
-  commit: nil
+  commit: nil,
+  updater_options: {},
 }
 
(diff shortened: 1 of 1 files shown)


jasonrudolph:
Looks promising, @thepwagner! Can you share a quick example usage and/or describe what steps we can take to verify that this is working as expected? :bow:

thepwagner:
@jasonrudolph  thanks! I'm not set on this implementation at all, but I wanted to breadcrumb this as it was handy for testing. 

As implemented, usage is: `UPDATER_OPTIONS=go_mod_tidy,go_mod_vendor bin/dry-run.rb go_modules thepwagner/my-awesome-repo --clone`

The alternative I considered as I was writing the PR description was using the existing [OptionParser](https://github.com/dependabot/dependabot-core/blob/main/bin/dry-run.rb#L121) for something like:
`bin/dry-run.rb go_modules thepwagner/my-awesome-repo --updater-opts=go_mod_tidy,go_mod_vendor --clone`.
Semantics but it looks like the pattern is to only use `ENV` for secrets?

Toggling the `go_mod_tidy` against `go_modules` is probably the best way to verify: does `go.sum` contain the previous version? If not, you're tidying.

thepwagner:
@jasonrudolph I talked myself into the options path since it provides better opportunity for discovery.

```
 bin/dry-run.rb go_modules thepwagner/my-awesome-repo --clone --updater-options=go_mod_tidy
```

I'm still hesitant to bake the available options into this script, but the help text hopefully leaves a breadcrumb that the ecosystem's file updater is what defines them.
We could add something like `FileUpdater::Base.available_options = []` to do it, doable but a fair bit of creep.

jasonrudolph (approved):
Nice work, @thepwagner! And thanks for describing how to try this out in https://github.com/dependabot/dependabot-core/pull/2612#issuecomment-705142923. I tried it out locally, and verified that it performs a `go mod tidy` when `--updater_options=go_mod_tidy` is given, and it doesn't perform `go mod tidy` otherwise. :clap:

feelepxyz (approved):
🙌

jurre (approved):
This is really nice 💯

PR #2615, opened 2020-10-08
Use `file` utility to check if files are binary

jurre:
Noticed a bunch of extensions that were incorrectly being submitted as
binary to the GitHub API.

Detecting if files are binary or not is hard, and lists of extensions are a pain to maintain, plus there can be any number of files that don't have an extension. This is especially problematic for `git` dependencies.

Currently, when a file is not marked correctly as binary, this causes issues with PR creation, as we need to [encode and upload those files differently](https://github.com/dependabot/dependabot-core/blob/jurre/add-extensions/common/lib/dependabot/pull_request_creator/github.rb#L182-L188)

This 🩹 should at least make sure that we do the right thing.

Longer term, we want to investigate creating compressed git packs, and uploading those packs using the API.


jurre:
I will add some tests to demonstrate pathological cases tomorrow, but would love other feedback in the meanwhile.

jasonrudolph (approved):
> I will add some tests to demonstrate pathological cases tomorrow, but would love other feedback in the meanwhile.

An additional test or two would be great, but regardless, this seems like a nice pragmatic choice. :zap:

jurre:
These cargo failures seem unrelated and I'm also getting them on `main`. Looking into it now

PR #2616, opened 2020-10-08
[Friction] If there isn't a backup gitconfig, remove the generated one

brrygrdn:
Minor thing I noticed during dev, I don't have a `~/.gitconfig` in my native dev environment so after running tests I frequently have one left behind by Dependabot Core.

This manifested in a slightly indirect way where I got 4xx errors trying to push my branch and had to poke around to figure out where the config was appearing from.

It just seems like an oversight we don't delete the config if we didn't move a backup aside, but it seems like the right outcome.


jurre (approved):

brrygrdn:
@jurre Thanks for the review - I added one more thing as an afterthought as we can still crash and leave the mystery config around, so I figured putting a fingerprint on the file would be useful.

jurre:
Nice one 👍

Issue #2618, opened 2020-10-09
go modules in subgroups cannot be resolved


hensur:
I recently deployed https://gitlab.com/dependabot-gitlab/dependabot on a private gitlab instance. I added a `.netrc` to the dependabot container which allow dependabot to fetch private repositories.

However, I noticed, that repositories inside gitlab subgroups could not be resolved.
The dependency `https://gitlab.example.com/test/subtest/dependency.git` causes this error:
```
remote: The project you were looking for could not be found.
    fatal: repository 'https://gitlab.example.com/test/subtest.git/' not found
```

I know that the `dependabot-gitlab` project might behave differently, but I was able to reproduce the issue with the helpers in `/go_modules/helpers` as well:
```
echo '{"function": "getUpdatedVersion","args": {"dependency": {"name": "gitlab.example.com/test/subtest/dependency", "version":"v0.1.5", "indirect": false}}}' | go_modules/helpers/helpers
```
This command returns the same error as described above.

The issue seems to be an outdated version of the golang internal module code which gets imported from https://github.com/dependabot/gomodules-extracted .

I re-extracted the code and was able to successfully fetch the latest version for the given package. The updated `gomodules-extracted` package requires some changes to the helpers in this repo as well, as some function signatures have changed.

I will create PRs with my changes and reference them here.
gomodules-extracted PR: https://github.com/dependabot/gomodules-extracted/pull/6
dependabot-core PR (depends on gomodules PR): https://github.com/dependabot/dependabot-core/pull/2619

**Package manager/ecosystem**
go mod

**Manifest contents prior to update**
private gitlab instance

**Updated dependency**
private gitlab instance

**What you expected to see, versus what you actually saw**
no errors while checking private go modules

**Images of the diff or a link to the PR, issue or logs**

jeffwidman:
Can this be closed now that the two PR's necessary to fix this have been merged?

hensur:
Yes, this can be closed!

PR #2619, opened 2020-10-09
bump gomodules-extracted

hensur:
depends on https://github.com/dependabot/gomodules-extracted/pull/6

This should resolve some issues I had with go modules in gitlab subgroups: https://github.com/dependabot/dependabot-core/issues/2618


jurre:
Thanks @hensur, I wonder, now that we're pulling in `golang.org/x/mod` (which wasn't available previously I believe) do we even need the `gomodules-extracted` dependency? Would be great to just rely on the native package

hensur:
I noticed this as well and checked if we can import all go mod dependencies from `golang.org/x/mod`. But there are still some functions in internal packages that we have to import from `gomodules-extracted`. Take a look at the imports in `go_modules/helpers/updatechecker/main.go`. I couldn't find a exported version of the internal imports in `golang.org/x/mod`.

go_modules/helpers/go.sum
```diff
@@ -4,3 +4,22 @@ github.com/dependabot/gomodules-extracted v0.0.0-20181020215834-1b2f850478a3 h1:
 github.com/dependabot/gomodules-extracted v0.0.0-20181020215834-1b2f850478a3/go.mod h1:+dRXSrUymjpT4yzKtn1QmeknT1S/yAHRr35en18dHp8=
 github.com/dependabot/gomodules-extracted v1.1.0 h1:k0Fcin3JZBfd/yW0JEeg38UtTiZl6E9uYkdqA4Dqbyw=
 github.com/dependabot/gomodules-extracted v1.1.0/go.mod h1:+dRXSrUymjpT4yzKtn1QmeknT1S/yAHRr35en18dHp8=
```
jurre, line 6: Would you mind running `go mod tidy` in this dir?

hensur: done!

PR #2626, opened 2020-10-12
Ignore changed  .gemspec from vendor/cache folder

feelepxyz:
When we vendor a bundler project we often end up with an updated `.gemspec` files in the vendor folder but we don't care about these when creating the PR branch or message as this incorrectly identifies the dependency as a library and attempts to generate a updated requirement.

The requirement sometimes doesn't change when only bumping the version in the lockfile.

This is a quick fix to keep the existing logic but we should refactor this to get it in line with the existing logic to determine if a project is a library as this varies between package manager.

### DIFF
--- a/common/lib/dependabot/pull_request_creator.rb
+++ b/common/lib/dependabot/pull_request_creator.rb
@@ -73,7 +73,7 @@ module Dependabot
     end
 
     def check_dependencies_have_previous_version
-      return if library? && dependencies.all? { |d| requirements_changed?(d) }
+      return if dependencies.all? { |d| requirements_changed?(d) }
       return if dependencies.all?(&:previous_version)
 
       raise "Dependencies must have a previous version or changed " \
@@ -212,12 +212,6 @@ module Dependabot
           label_language: label_language?,
           automerge_candidate: automerge_candidate?
         )
-    end
-
-    def library?
-      return true if files.any? { |file| file.name.end_with?(".gemspec") }
-
-      dependencies.any? { |d| !d.appears_in_lockfile? }
     end
 
     def includes_security_fixes?
--- a/common/lib/dependabot/pull_request_creator/branch_namer.rb
+++ b/common/lib/dependabot/pull_request_creator/branch_namer.rb
@@ -165,12 +165,12 @@ module Dependabot
         updated_reqs.first[:requirement]
       end
 
-      # TODO: Look into bringing this in line with existing library checks that
-      # we do in the update checkers, which are also overriden by passing an
-      # explicit `requirements_update_strategy`.
+      # TODO: Bring this in line with existing library checks that we do in the
+      # update checkers, which are also overriden by passing an explicit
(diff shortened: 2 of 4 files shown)


common/lib/dependabot/pull_request_creator.rb
```diff
@@ -214,12 +214,6 @@ def labeler
         )
     end
 
-    def library?
```
feelepxyz, line 217: Not sure this is needed here as we only seem to care if the requirement has changed or we have a previous version.

common/lib/dependabot/pull_request_creator/branch_namer.rb
```diff
@@ -165,12 +165,12 @@ def new_library_requirement(dependency)
...
-      # TODO: Look into bringing this in line with existing library checks that
-      # we do in the update checkers, which are also overriden by passing an
-      # explicit `requirements_update_strategy`.
+      # TODO: Bring this in line with existing library checks that we do in the
+      # update checkers, which are also overriden by passing an explicit
+      # `requirements_update_strategy`.
+      #
+      # TODO re-use in MessageBuilder
       def library?
-        return true if files.map(&:name).any? { |nm| nm.end_with?(".gemspec") }
```
feelepxyz, line 172: Not 💯  sure about this yet, might be better to only use the requirement if it's changed instead if inferring change based on this

jurre (approved):
I think this is OK as a stop-gap solution

common/lib/dependabot/pull_request_creator/message_builder.rb
```diff
@@ -459,8 +459,15 @@ def ref_changed?(dependency)
...
+      # TODO: Bring this in line with existing library checks that we do in the
+      # update checkers, which are also overriden by passing an explicit
+      # `requirements_update_strategy`.
+      #
+      # TODO re-use in BranchNamer
       def library?
-        return true if files.map(&:name).any? { |nm| nm.end_with?(".gemspec") }
+        gemspecs = files.map(&:name).select { |nm| nm.end_with?(".gemspec") }
+        # Reject changed gemspecs from the vendor cache folder (from git deps)
+        return true if gemspecs.reject { |nm| nm.start_with?("vendor/") }.any?
```
thepwagner, line 470: Confirming: it's OK that this only respects the default path?

thepwagner (approved):
Noice; clear explanation and 🩹 

Could we snapshot `library?` between the traditional updater and when we refresh the bundler cache? 🤷 about feasibility, just brainstorming ignoring vendored gemspecs is easy if they haven't changed yet.

common/lib/dependabot/pull_request_creator/message_builder.rb
jurre: Ooh good spot, might want to invert this and explicitly check in the root/update directory?

jurre (approved):

PR #2627, opened 2020-10-12
Gracefully handle gomod package import that has changed

jurre:
When a module exists, but it (no longer) defines the package of a given
name, we get an error like such:

```
go: finding module for package github.com/googleapis/gnostic/OpenAPIv2
github.com/dependabot/vgotest imports
	github.com/googleapis/gnostic/OpenAPIv2: module github.com/googleapis/gnostic@latest found (v0.5.1), but does not contain package github.com/googleapis/gnostic/OpenAPIv2
```

In this case, we should not treat that as a failed update, but rather as
a DependencyFileNotResolvable.


feelepxyz (approved):
💯

PR #2631, opened 2020-10-13
Handle `cannot find module` during go mod tidy

jurre:
Previously we would only catch this type of error during `go build`, by
explicitly matching for the `build` prefix. We also want to gracefully
handle this error when running `go mod tidy`.

I am not sure why this error sometimes arises during build time (likely
when running `go get`), and sometimes during other commands, and I also
cannot seem to replicate the error in a test scenario :/ it appears we're missing
test coverage for the existing error as well, so may be OK adding this as is?

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
@@ -16,6 +16,7 @@ module Dependabot
         RESOLVABILITY_ERROR_REGEXES = [
           /go: .*: git fetch .*: exit status 128/.freeze,
           /verifying .*: checksum mismatch/.freeze,
+          /cannot find module providing package .*: exit status 128/m.freeze,
           /build .*: cannot find module providing package/.freeze,
           /module .* found \(.*\), but does not contain package/m.freeze
         ].freeze


feelepxyz (approved):
Nice one! Yeah maybe worth adding a comment around it to explain when it happens as a test-substitute?

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -14,9 +14,15 @@ class GoModUpdater
         ENVIRONMENT = { "GOPRIVATE" => "*" }.freeze
 
         RESOLVABILITY_ERROR_REGEXES = [
+          # Private module could not be fetched
           /go: .*: git fetch .*: exit status 128/.freeze,
+          # The checksum in go.sum does not match the dowloaded content
           /verifying .*: checksum mismatch/.freeze,
+          # Private module could not be found during `go mod tidy`
+          /cannot find module providing package .*: exit status 128/m.freeze,
```
thepwagner, line 22: Idea: could this be combined with the below to just `/cannot find module providing package/`?
It's maybe the same root, just being wrapped differently.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -14,9 +14,15 @@ class GoModUpdater
...
+          # Private module could not be found during `go build`
           /build .*: cannot find module providing package/.freeze,
+          # Package in module was likely renamed or removed
```
thepwagner, line 25: Scope creep: https://github.com/facebook/ent/pull/660 caused 
```
go: github.com/facebookincubator/ent@v0.4.2: parsing go.mod:\n\tmodule declares its path as: github.com/facebook/ent\n\t        but was required as: github.com/facebookincubator/ent\n
```

Seems related to this block, but not what we're here for. (I was 🤞 it would be the missing reproduction).

thepwagner (approved):
LGTM; smells like existing path.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jurre: That would currently bubble up as a not_parseable_error I think, does that seem right?

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jurre: Yeah we probably can, I was slightly worried about some other error also reusing that string and us matching on it by accident but the fact that you suggest it makes me feel that's probably not likely. Will update 👍

thepwagner: My baseless assumption was this is something like `fmt.Errorf("cannot find module providing package: %s")` at the root. This seems [pretty close](https://github.com/golang/go/blob/ab2a5b48665eed6d670d719cdef5335bc3602359/src/cmd/go/internal/modload/import.go#L47) (and like the only hit for that string in the codebase?)

Then I assumed this was two different: `fmt.Errorf("build: %w", err)`  / `fmt.Errorf("tidy: %w", err)` paths wrapping - not sure where that's happening, but it seems 💯 that this error could be wrapped differently and should always be handled the same.

PR #2648, opened 2020-10-16
fix(docker): support platform option in from line

ttshivers:
Dockerfiles can specify the platform in the from line. This change adds support for parsing and updating FROM lines with the platform flag specified.

```Dockerfile
FROM [--platform=<platform>] <image> [AS <name>]
```

https://docs.docker.com/engine/reference/builder/#from


feelepxyz (approved):
Thanks for fixing this and for the added tests 👍  LGTM

PR #2667, opened 2020-10-21
Explain update not possible

jurre:
In order to better explain why a security update is not possible, we're exposing the  `lowest_security_fix_version` which returns the lowest available non-vulnerable version from the package registry not taking resolvability into account. We'll use this alongside the existing method `lowest_resolvable_security_fix_version` to explain which version is the latest possible version that can be installed and which version is the earliest non-vulnerable/fixed version.

Planned "update not possible" error message:

`Dependabot cannot update #{dependency_name} to a non-vulnerable version`

```
The latest possible version that can be installed is #{lowest_resolvable_security_fix_version},
another dependency is locking  #{dependency_name} to a vulnerable version range.
The earliest fixed version is #{lowest_security_fix_version}.
```

### DIFF
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -350,7 +350,9 @@ dependencies = cached_read("dependencies") { parser.parse }
 if $options[:dependency_name].nil?
   dependencies.select!(&:top_level?)
 else
-  dependencies.select! { |d| d.name == $options[:dependency_name] }
+  dependencies.select! do |d|
+    d.name.downcase == $options[:dependency_name].downcase
+  end
 end
 
 def update_checker_for(dependency)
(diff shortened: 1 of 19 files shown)


bin/dry-run.rb
```diff
@@ -350,7 +350,9 @@ def cached_dependency_files_read
 if $options[:dependency_name].nil?
   dependencies.select!(&:top_level?)
 else
-  dependencies.select! { |d| d.name == $options[:dependency_name] }
+  dependencies.select! do |d|
+    d.name.downcase == $options[:dependency_name].downcase
```
feelepxyz, line 354: Fix for maven/gradle, planning to port change to updater also

feelepxyz (approved):

PR #2686, opened 2020-10-27
Handle mixed case advisory dependency name & sec only dry-run

feelepxyz:
Handle mixed case dependency name when checking if updated dependency fixes security advisory.

## dry-run changes

- Add option to only do a security update and check if the updated dependency fixes any vulnerabilities
- Security only updates won't attempt to update any non-vulnerable versions

### DIFF
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -439,7 +439,7 @@ end
 
 def security_fix?(dependency)
   security_advisories.any? do |advisory|
-    advisory.fixes_advisory?(dependency)
+    advisory.fixed_by?(dependency)
   end
 end
 
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -101,7 +101,8 @@ $options = {
   requirements_update_strategy: nil,
   commit: nil,
   updater_options: {},
-  security_advisories: []
+  security_advisories: [],
+  security_updates_only: false,
 }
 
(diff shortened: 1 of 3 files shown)


bin/dry-run.rb
```diff
@@ -441,6 +455,17 @@ def generate_dependency_files_for(updated_dependencies)
...
   puts " => latest available version is #{checker.latest_version}"
 
+  if $options[:security_updates_only] && !checker.vulnerable?
+    if checker.version_class.correct?(checker.dependency.version)
+      puts "    (no security update needed as it's not vulnerable)"
+    else
+      puts "    (can't update vulnerable dependencies for "\
+           "projects without a lockfile as the currently "\
+           "installed version isn't known 🚨)"
+    end
```
feelepxyz, line 465: Planning to do the same in the updater

common/spec/dependabot/security_advisory_spec.rb
```diff
@@ -149,6 +149,11 @@
     end
 
     context "for a different dependency" do
+      let(:dependency_name) { "Rails" }
+      it { is_expected.to eq(true) }
+    end
+
+    context "when the name has a different case" do
```
jurre, line 156: Should these example names be swapped?

jurre (approved):
Looks great

common/spec/dependabot/security_advisory_spec.rb
feelepxyz: Ah yes 👍

PR #2691, opened 2020-10-28
Include backtrace in bundler native errors

jurre:
This forwards the backtrace from any native bundler helper, which makes it easier to figure out what happened in the native helper when ran from core.

I named this `trace` as to not clash with `StandardError#backtrace`, but open to alternatives.

### DIFF
--- a/common/lib/dependabot/shared_helpers.rb
+++ b/common/lib/dependabot/shared_helpers.rb
@@ -58,13 +58,14 @@ module Dependabot
     end
 
     class HelperSubprocessFailed < StandardError
-      attr_reader :error_class, :error_context
+      attr_reader :error_class, :error_context, :trace
 
-      def initialize(message:, error_context:, error_class: nil)
+      def initialize(message:, error_context:, error_class: nil, trace: nil)
         super(message)
         @error_class = error_class || ""
         @error_context = error_context
         @command = error_context[:command]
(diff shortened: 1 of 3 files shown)


common/lib/dependabot/shared_helpers.rb
```diff
@@ -58,13 +58,14 @@ def self.in_a_temporary_directory(directory = "/")
...
-      attr_reader :error_class, :error_context
+      attr_reader :error_class, :error_context, :trace
 
-      def initialize(message:, error_context:, error_class: nil)
+      def initialize(message:, error_context:, error_class: nil, trace: nil)
         super(message)
         @error_class = error_class || ""
         @error_context = error_context
         @command = error_context[:command]
+        @trace = trace
```
feelepxyz, line 68: Could we also set the backtrace here to have it rendered in sentry?

```suggestion
        @trace = trace
        set_backtrace(trace) if trace
```

jurre: We can, but it would obfuscate the backtrace of the calling context, I think it might be more useful to keep the backtrace of the calling code in tact as well.

feelepxyz: Good point 👍

feelepxyz (approved):
Nice one 👍

PR #2693, opened 2020-10-28
Handle unknown revision errors in go mod updater

jurre:
We get a `unknown revision` error when a certain version (git tag)
cannot be resolved. This was reported as a DependencyFileNotParseable
before, but the file is perfectly parseable, we just cannot resolve it.

When we do not know what happened, we currently assume the go.mod file
is not parseable, but that's very unlikely to be the case, as we already
parsed the file in the `UpdateChecker`. Instead raise a generic
`DependabotError` to indicate that we do not know what happened.


go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -21,7 +21,9 @@ class GoModUpdater
           # (Private) module could not be found
           /cannot find module providing package/.freeze,
           # Package in module was likely renamed or removed
-          /module .* found \(.*\), but does not contain package/m.freeze
+          /module .* found \(.*\), but does not contain package/m.freeze,
+          # Package does not exist or has been pulled
+          /go: .*: unknown revision/m.freeze
```
fatih, line 26: I think you also get this error if you're unauthenticated and the `go` tool can't get the module (either from Git directly or Go module proxy). Not sure if you want to update the comment in the code to make it clear.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -252,8 +254,9 @@ def handle_subprocess_error(stderr)
               new(go_mod_path, match[1], match[2])
           end
 
+          # We don't know what happened so we raise a generic error
           msg = stderr.lines.last(10).join.strip
-          raise Dependabot::DependencyFileNotParseable.
+          raise Dependabot::DependabotError.
             new(go_mod_path, msg)
```
feelepxyz, line 260: Looks like this only accepts on arg

```suggestion
            new(msg)
```

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
feelepxyz, line 26: Worth adding a spec for this case?

J-Vin:
Are additional headers typical? I keep running into this and asking, but apparently not understanding. 😕

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jurre: Yeah 👍

jurre: Thanks, added a note to clarify that 👍

feelepxyz (approved):

PR #2695, opened 2020-10-29
Handle missing directory in cloned repo

feelepxyz:
Handle missing directory in cloned repo by creating an empty one and
relying on file fetcher to raise because it can't find a manifest file.

Fixed up some issues around dry-run cloning and the go mod clone stub
(which might not be needed anymore?).

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater.rb
@@ -65,15 +65,23 @@ module Dependabot
       def use_repo_contents_stub
         @repo_contents_stub = true
         @repo_contents_path = Dir.mktmpdir
+
         Dir.chdir(@repo_contents_path) do
           dependency_files.each do |file|
-            File.write(file.name, file.content)
+            path = File.join(@repo_contents_path, directory, file.name)
+            path = Pathname.new(path).expand_path
+            FileUtils.mkdir_p(path.dirname) unless Dir.exist?(path.dirname)
+            File.write(path, file.content)
           end
-          `git config --global user.email "no-reply@github.com"`
-          `git config --global user.name "Dependabot"`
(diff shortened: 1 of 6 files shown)


jurre:
For context, when does this happen exactly?

feelepxyz:
> For context, when does this happen exactly?

Whenever the passed `directory` doesn't actually exist, e.g.

`bin/dry-run.rb go_modules "dependabot/dependabot-core" --dir="/whaaat" --cache=files`

Adding a spec for the file fetcher

go_modules/lib/dependabot/go_modules/file_updater.rb
```diff
@@ -67,7 +67,10 @@ def use_repo_contents_stub
         @repo_contents_path = Dir.mktmpdir
         Dir.chdir(@repo_contents_path) do
           dependency_files.each do |file|
-            File.write(file.name, file.content)
+            path = File.join(@repo_contents_path, directory, file.name)
+            path = Pathname.new(path).expand_path
+            FileUtils.mkdir_p(path.dirname) unless Dir.exist?(path.dirname)
+            File.write(path, file.content)
```
feelepxyz, line 73: Not sure this is used but was causing problems without the dry-run clone changes

jurre: It should not be used in prod, but people might not perform full clones when running standalone

jurre (approved):
Nice work

go_modules/lib/dependabot/go_modules/file_updater.rb
feelepxyz: Added a test case for this and fixed the global git config messing with my local config outside of docker, accidentally committed with `no-reply@github.com` 😅

PR #2696, opened 2020-10-29
Handle missing VCS when converting git_source path

feelepxyz:
Copied from go dep implementation:
https://github.com/dependabot/dependabot-core/blob/main/dep/lib/dependabot/dep/file_parser.rb#L142

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_parser.rb
+++ b/go_modules/lib/dependabot/go_modules/file_parser.rb
@@ -174,6 +174,15 @@ module Dependabot
           ref: git_revision(dep),
           branch: nil
         }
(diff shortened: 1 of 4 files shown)


go_modules/spec/fixtures/go_mods/source_unresolvable.mod
```diff
@@ -0,0 +1,11 @@
+module source/unresolved/cloud/go
+
+replace (
+	source/unresolved/lib/go => ../../lib/go
```
jurre, line 4: I think these need to be strings

```suggestion
	"source/unresolved/lib/go" => "../../lib/go"
```

Issue #2687, opened 2020-10-27
Dependabot vendors go modules into a nested directory


Kemosabert:
**Package manager/ecosystem**
gomod
**What you expected to see, versus what you actually saw**
We have a monorepo and enabled dependabot for a service in the monorepo by adding a `dependabot.yml` file at the root of the repo in the `.github` folder. Since the `go.mod` file of the service for which we want to have automatic bumps is in the `catalog` folder, we added the `directory: "/catalog"` property to the config file. We vendor our dependencies in this application.

When dependabot created a PR for the bump of a dependency, it automatically vendored it, which  is great, but it did so in the `/catalog/catalog/vendor` folder and not in `/catalog/vendor` folder like it should be. It seems like it's nesting the vendored dependencies wrongly somehow.

**Images of the diff or a link to the PR, issue or logs**
<img width="1874" alt="Screenshot 2020-10-27 at 17 31 14" src="https://user-images.githubusercontent.com/21224391/97332043-4ead7780-187a-11eb-9d36-2016b6ea3e59.png">

jurre:
Thanks for reporting this @Kemosabert, that definitely seems like a bug on our end, I'll look into it!

Kemosabert:
Thanks for checking this @jurre, let me know if you need more information or context about this.

PR #2697, opened 2020-10-29
Correctly handle vendored updates in nested directory

jurre:
Previously we would prepend the nested directory to the filename, but
this is normally stored separately in the `directory` field of a
DependencyFile.

This would result in PRs being opened with the vendored files in one
directory too deep (`custom_dir/custom_dir/vendor` instead of
`custom_dir/vendor`).

Fixes: https://github.com/dependabot/dependabot-core/issues/2687


common/lib/dependabot/file_updaters/vendor_updater.rb
```diff
@@ -31,8 +31,10 @@ def updated_vendor_cache_files(base_directory:)
               encoding = Dependabot::DependencyFile::ContentEncoding::BASE64
               encoded_content = Base64.encode64(encoded_content) unless deleted
             end
+
+            clean_base_directory = Pathname.new(base_directory).basename
```
feelepxyz, line 35: Is `basename` the right thing here? Looks like it just grabs the last component `Pathname.new("dir/nested").basename` > `nested` 

Should it instead be something like?

```
project_root  = Pathname.new(File.expand_path(File.join(Dir.pwd, base_directory)))
filename = Pathname.new(path).expand_path.relative_path_from(project_root)

Comment (jurre) [reply common/lib/dependabot/file_updaters/vendor_updater.rb]: Yeah great suggestion, I've improved the other relative path code as well

File: common/lib/dependabot/file_updaters/vendor_updater.rb [hunk 2]
```diff
@@ -18,7 +18,8 @@ def updated_vendor_cache_files(base_directory:)
         return [] unless repo_contents_path && vendor_dir
 
         Dir.chdir(repo_contents_path) do
-          relative_dir = vendor_dir.sub("#{repo_contents_path}/", "")
+          relative_dir = Pathname.new(vendor_dir).relative_path_from(Dir.pwd)
```
feelepxyz, line 21: 👍

feelepxyz (approved):
Nice one! LGTM 👍

PR #2699, opened 2020-10-29
Fix relative dir on mac where tmp is in /private

feelepxyz:
Get the relative vendor dir from the repo_contents_path which is what we
used to construct the vendor_dir.

When creating a tmp dir in mac it returns the directory as `/var/tmp`
but if you run `pwd` in the folder it's suddently in `/private/var/tmp`.

### DIFF
--- a/common/lib/dependabot/file_updaters/vendor_updater.rb
+++ b/common/lib/dependabot/file_updaters/vendor_updater.rb
@@ -18,7 +18,9 @@ module Dependabot
         return [] unless repo_contents_path && vendor_dir
 
         Dir.chdir(repo_contents_path) do
-          relative_dir = Pathname.new(vendor_dir).relative_path_from(Dir.pwd)
+          relative_dir = Pathname.new(vendor_dir).relative_path_from(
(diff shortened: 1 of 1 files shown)


common/lib/dependabot/file_updaters/vendor_updater.rb
```diff
@@ -18,7 +18,9 @@ def updated_vendor_cache_files(base_directory:)
         return [] unless repo_contents_path && vendor_dir
 
         Dir.chdir(repo_contents_path) do
-          relative_dir = Pathname.new(vendor_dir).relative_path_from(Dir.pwd)
+          relative_dir = Pathname.new(vendor_dir).relative_path_from(
+            repo_contents_path
+          )
```
feelepxyz, line 23: Not sure how to test this 🤔

jurre: Only seems to happen on mac, current tests are passing so 👍 I guess

jurre (approved):

PR #2700, opened 2020-10-29
Ignore go files tagged with `+build`

jurre:
`go get` will ignore any files marked with `// +build`, so if the
only files present at the root of the directory are tagged with this, we
do not have a valid module, and `go get` will fail.

We now check explicitly if go files are ignored or not, and build a
dummy package if needed.

Co-authored-by: Philip Harrison <feelepxyz@github.com>


go_modules/spec/dependabot/go_modules/file_updater/go_mod_updater_spec.rb
```diff
@@ -1,4 +1,4 @@
-# frozen_string_literal: true
+# .freeze frozen_string_literal: true
```
feelepxyz, line 1:
```suggestion
# frozen_string_literal: true
```

feelepxyz (approved):
LGTM 💯

go_modules/spec/fixtures/projects/ignored_go_files/.ignored.go
```diff
@@ -0,0 +1,5 @@
+// This file is ignored because it starts with a dot
```
feelepxyz, line 1: 👍

PR #2701, opened 2020-10-29
Go mod: handle missing package url meta tags

feelepxyz:
Raise DependencyFileNotResolvable error when checking for latest version
for a package that doesn't have any valid meta tags.

### DIFF
--- a/go_modules/lib/dependabot/go_modules/update_checker.rb
+++ b/go_modules/lib/dependabot/go_modules/update_checker.rb
@@ -10,6 +10,11 @@ require "dependabot/go_modules/version"
 module Dependabot
   module GoModules
     class UpdateChecker < Dependabot::UpdateCheckers::Base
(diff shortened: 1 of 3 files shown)


go_modules/lib/dependabot/go_modules/update_checker.rb
```diff
@@ -73,6 +78,18 @@ def find_latest_resolvable_version
...
+
+        handle_subprocess_error(e)
+      end
+
+      def handle_subprocess_error(error)
+        resolvability_error = RESOLVABILITY_ERROR_REGEXES.any? do |r|
+          error.message =~ r
+        end
+        if resolvability_error
+          raise Dependabot::DependencyFileNotResolvable.new, error.message
```
jurre, line 90:
```suggestion
          raise Dependabot::DependencyFileNotResolvable, error.message
```

jurre (approved):

PR #2705, opened 2020-10-30
Explain update not possible for bundler

jurre:
This adds a new method `blocking_parent_dependencies` to `UpdateChecker`, and implements it for bundler. It does this by traversing the dependency tree and finding all parent dependencies that are not satisfied by the target version.

It also outputs this in the dry-run script as follows:

```
=== kramdown (1.17.0) (vulnerable 🚨)
 => checking for updates
 => latest available version is 2.3.0
 => earliest available non-vulnerable version is 2.3.0
 => latest allowed version is 1.17.0
 => The update is not possible because of the following conflicting dependencies:
   github-pages (204) which requires:
     kramdown = 1.17.0
   jekyll (3.8.5) which requires:
     kramdown ~> 1.14
```


brrygrdn (approved):
Nice find, 👍🏻

bundler/helpers/lib/functions/force_updater.rb
```diff
@@ -1,5 +1,7 @@
 module Functions
   class ForceUpdater
+    class TransitiveDependencyError < StandardError; end
```
feelepxyz, line 3: Nice one, would be nice to also add a spec to cover this case so we don't accidentally break it in future.

Planning to rescue TransitiveDependencyError and raise a new update not possible error or similar?

jurre: This code is already sort of tested, we currently catch _any_ error here: https://github.com/dependabot/dependabot-core/blob/5cb4b67e15e19afc0eb6fc377fca7b8098417477/bundler/lib/dependabot/bundler/update_checker/force_updater.rb#L62

Functionally nothing has changed yet, but I plan to follow up on this by actually handling this error differently, my plan is to add some context to the message that describes how the transitive dependency was locked. Totally happy to fold those changes into one PR, as that change should be tested separately 👍

feelepxyz: Ah gotcha, that makes sense 👍

feelepxyz (approved):

bundler/helpers/lib/functions/parent_dependency_resolver.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+module Functions
+  class ParentDependencyResolver
+    def initialize(dependency_name:, target_version:, lockfile_name:)
+      @dependency_name = dependency_name
+      @target_version = target_version
+      @lockfile_name = lockfile_name
+
+      Bundler.settings.set_command_option("only_update_to_newer_versions", true)
```
thepwagner, line 10: Concern: this seems to modify static state from a constructor.
It seems ~safe given the usage, but heuristically I would never expect a constructor to modify static state.

Ideas:
* shift to the top of `blocking_parent_dependencies`
* refactor to a static method

bundler/helpers/lib/functions/parent_dependency_resolver.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+    # Finds any dependencies in the lockfile that have a subdependency on the
+    # given dependency that does not satisfly the target_version.
```
thepwagner, line 14:
```suggestion
    # given dependency that does not satisfy the target_version.
```

Nit: typo.

bundler/helpers/lib/functions/parent_dependency_resolver.rb
```diff
@@ -0,0 +1,55 @@
...
+    # @return [Array<Hash{String => String}]
```
thepwagner, line 15: Noice: these docs!

bundler/helpers/lib/functions/parent_dependency_resolver.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+module Functions
+  class ParentDependencyResolver
+    def initialize(dependency_name:, target_version:, lockfile_name:)
+      @dependency_name = dependency_name
+      @target_version = target_version
```
thepwagner, line 7:
```suggestion
      @target_version = Gem::Version.new(target_version)
```

Idea: the only usage is as a `Gem::Version`, just store as that.

bundler/lib/dependabot/bundler/update_checker/parent_dependency_resolver.rb
```diff
@@ -0,0 +1,48 @@
...
+        include SharedBundlerHelpers
+
+        def initialize(dependency_files:, repo_contents_path:, credentials:)
+          @dependency_files = dependency_files
+          @repo_contents_path = repo_contents_path
+          @credentials = credentials
+        end
+
+        # Finds any dependencies in the lockfile that have a subdependency on
+        # the given dependency that does not satisfly the target_version.
```
thepwagner, line 21:
```suggestion
        # the given dependency that does not satisfy the target_version.
```

bundler/spec/dependabot/bundler/update_checker/parent_dependency_resolver_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+require "dependabot/dependency_file"
+require "dependabot/bundler/update_checker/parent_dependency_resolver"
+
+RSpec.describe Dependabot::Bundler::UpdateChecker::ParentDependencyResolver do
+  include_context "stub rubygems compact index"
+
+  let(:resolver) do
+    described_class.new(
+      dependency_files: dependency_files,
+      repo_contents_path: repo_contents_path,
```
thepwagner, line 15:
```suggestion
```

Nit: this is always `nil`, skip.

bundler/spec/dependabot/bundler/update_checker/parent_dependency_resolver_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+      credentials: [{
+        "type" => "git_source",
+        "host" => "github.com",
+        "username" => "x-access-token",
+        "password" => "token"
+      }]
+    )
+  end
+  let(:dependency_files) { [gemfile, lockfile] }
+  let(:repo_contents_path) { nil }
```
thepwagner, line 25:
```suggestion
```

Per above, this never varies so rm

bundler/spec/dependabot/bundler/update_checker/parent_dependency_resolver_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+
+  describe "#blocking_parent_dependencies" do
+    subject(:blocking_parent_dependencies) do
+      resolver.blocking_parent_dependencies(
+        dependency: dependency,
+        target_version: target_version
+      )
+    end
+
+    it "returns the right array of updated dependencies" do
```
thepwagner, line 58:
```suggestion
    it "returns the right array of blocking dependencies" do
```

Nit: I didn't think this was updating anything.

common/lib/dependabot/update_checkers/base.rb
```diff
@@ -93,6 +93,16 @@ def latest_resolvable_version_with_no_unlock
         raise NotImplementedError
       end
 
+      # Finds any dependencies in the lockfile that have a subdependency on the
+      # given dependency that do not satisfly the target_version.
```
thepwagner, line 97:
```suggestion
      # given dependency that do not satisfy the target_version.
```

thepwagner (approved):
LGTM - comments are all style/naming nits.

bin/dry-run.rb
```diff
@@ -482,11 +482,23 @@ def security_fix?(dependency)
...
+                           end
   puts " => latest allowed version is #{latest_allowed_version || dep.version}"
 
+  parents = checker.blocking_parent_dependencies
+  if parents.any?
+    puts " => The update is not possible because of the following conflicting "\
+      "dependencies:"
+  end
+  parents.each do |parent|
+    puts "   #{parent['name']} (#{parent['version']}) which requires:"
```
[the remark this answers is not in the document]
jurre: I contemplated returning `Dependency` objects instead, but those are already used for lot's of different things and would end up having lots of unused fields. We could introduce a new class for this as well. I ended up opting for a hash instead, so that we can pass it through to the calling code without having to map over it, saving us having to transform the object back and forth. wdyt?

bin/dry-run.rb
```diff
@@ -482,11 +482,23 @@ def security_fix?(dependency)
...
-    checker.lowest_resolvable_security_fix_version :
-    checker.latest_resolvable_version
+  latest_allowed_version = if checker.vulnerable?
+                             checker.lowest_resolvable_security_fix_version
+                           else
+                             checker.latest_resolvable_version
+                           end
   puts " => latest allowed version is #{latest_allowed_version || dep.version}"
 
+  parents = checker.blocking_parent_dependencies
```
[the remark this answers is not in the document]
feelepxyz: > npm peer dependencies

👍  yeah might not be parents, slight leaning towards `conflicting_dependencies` for same reason

bundler/helpers/lib/functions/force_updater.rb
```diff
@@ -113,12 +115,15 @@ def build_definition(other_updates:)
 
       # Set the requirement for the gem we're forcing an update of
       new_req = Gem::Requirement.create("= #{target_version}")
```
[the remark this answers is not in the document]
jurre: Yep, good call 👍

bin/dry-run.rb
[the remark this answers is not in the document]
thepwagner: SGTM. the documented keys of the hash were 💯 for bridging the gap between "a completely untyped hash" and "a new class", awesome!

bin/dry-run.rb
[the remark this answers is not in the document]
jurre: I like `conflicting_dependencies` 👍

bin/dry-run.rb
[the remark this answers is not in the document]
feelepxyz: Do we need to check this every time or could we nest this under `if requirements_to_unlock == :update_not_possible`?

feelepxyz (approved):
Great work on this! This looks 💯 🙌

bin/dry-run.rb
```diff
@@ -482,11 +482,23 @@ def security_fix?(dependency)
...
+  if parents.any?
+    puts " => The update is not possible because of the following conflicting "\
+      "dependencies:"
+  end
+  parents.each do |parent|
```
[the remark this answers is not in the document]
jurre: Yeah, if it's enough to make you go "hey, what's going on here?" let's nest it in the guard 👍

PR #2706, opened 2020-10-30
Fix error handling for packages that 404

feelepxyz:
Bumping gomodules-extracted (https://github.com/dependabot/dependabot-core/pull/2619) caused the raised error to change for import paths that 404/don't include valid meta tags.

### DIFF
--- a/go_modules/lib/dependabot/go_modules/update_checker.rb
+++ b/go_modules/lib/dependabot/go_modules/update_checker.rb
@@ -11,8 +11,8 @@ module Dependabot
   module GoModules
     class UpdateChecker < Dependabot::UpdateCheckers::Base
       RESOLVABILITY_ERROR_REGEXES = [
-        # Package url/proxy doesn't include any redirect meta tags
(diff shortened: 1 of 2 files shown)


go_modules/lib/dependabot/go_modules/update_checker.rb
```diff
@@ -11,8 +11,8 @@ module Dependabot
   module GoModules
     class UpdateChecker < Dependabot::UpdateCheckers::Base
       RESOLVABILITY_ERROR_REGEXES = [
-        # Package url/proxy doesn't include any redirect meta tags
-        /no go-import meta tags/
```
jurre, line 15: We could still get a `no go-import meta tags` when the proxy does exist but doesn't include the required meta tags right? Should we add a spec for that as well?

feelepxyz: Yeah good shout, will merge this just to get main green and then follow up with a new test-case for this

PR #2708, opened 2020-10-30
Go mod: Add/fix specs for missing meta tag and 404

feelepxyz:
Add a spec for handling package url that don't include meta tags.

Both 404 and missing meta tag errors raised a error message starting
with `unrecognized import path`. I've split these out with two different
test cases.


jurre (approved):
Nice one, cheers

Issue #2532, opened 2020-09-18
Updating stale Cargo dependencies always fails with "Expected content to change!"


olson-sean-k:
**Package manager/ecosystem**
Cargo (Rust).

**Manifest contents prior to update**
I'm seeing this in multiple repositories, but I'll use [this one](https://github.com/olson-sean-k/plexus) as an example. See the manifest [here](https://github.com/olson-sean-k/plexus/blob/be7e5f06bf6481ae788bb24ae68be1c3c9c4f326/Cargo.toml) and the most interesting member of the workspace [here](https://github.com/olson-sean-k/plexus/blob/be7e5f06bf6481ae788bb24ae68be1c3c9c4f326/plexus/Cargo.toml).

**Updated dependency**
I'm seeing these errors when Dependabot examines multiple dependencies, but [`nalgebra`] always shows up in the logs and is stale (I expect Dependabot to open a PR to upgrade it). I see both [`nalgebra`] and [`ultraviolet`] show up consistently in another repository ([`ultraviolet`] is also stale).

**What you expected to see, versus what you actually saw**
I expect Dependabot to run successfully and open PRs for stale crates ([`nalgebra`] and [`ultraviolet`]). Instead, checking for updates always fails and produces logs similar to the following:

```
updater | ERROR <job_62440466> Error processing nalgebra (RuntimeError)
updater | ERROR <job_62440466> Expected content to change!
```

**Images of the diff or a link to the PR, issue or logs**
See the logs [here](https://github.com/olson-sean-k/plexus/network/updates/62440466) and [here](https://github.com/olson-sean-k/theon/network/updates/62440465) for examples. Attempting an update in either of these repositories results in similar errors each time. See the Dependabot configuration [here](https://github.com/olson-sean-k/plexus/blob/master/.github/dependabot.yml).

[`nalgebra`]: https://crates.io/crates/nalgebra
[`ultraviolet`]: https://crates.io/crates/ultraviolet

olson-sean-k:
I manually updated the affected packages described previously and I'm now seeing the same errors in the logs for other packages that have become stale (in this case `winit`, which can be upgraded from `^0.22.1` to `^0.23.0`). I'm curious if Cargo support is broken in general or if there's something unique in my repositories that is triggering this behavior...

lucab:
[...]

PR #2712, opened 2020-11-02
Cargo: handle caret version requirements

feelepxyz:
Handle `^` version requirements by escaping the previous requirement.

The old requirement (e.g. `^1.0.0`) was interpolated unescaped inside a
regex match so the caret `^` broke the match.

### DIFF
--- a/cargo/lib/dependabot/cargo/file_updater/manifest_updater.rb
+++ b/cargo/lib/dependabot/cargo/file_updater/manifest_updater.rb
@@ -100,8 +100,9 @@ module Dependabot
           if simple_declaration
             simple_declaration_regex =
               /(?:^|["'])#{Regexp.escape(simple_declaration)}/
+            old_req_escaped = Regexp.escape(old_req)
             content.gsub(simple_declaration_regex) do |line|
-              line.gsub(/.+=.*\K(#{old_req})/, new_req)
+              line.gsub(/.+=.*\K(#{old_req_escaped})/, new_req)
             end
           elsif content.match?(feature_declaration_version_regex(dep))
             content.gsub(feature_declaration_version_regex(dep)) do |part|
--- a/cargo/spec/dependabot/cargo/file_updater/manifest_updater_spec.rb
+++ b/cargo/spec/dependabot/cargo/file_updater/manifest_updater_spec.rb
@@ -264,6 +264,30 @@ RSpec.describe Dependabot::Cargo::FileUpdater::ManifestUpdater do
             to include(%([dependencies.pango]\nversion = "0.3.0"\n))
         end
       end
+
+      context "with a version requirement" do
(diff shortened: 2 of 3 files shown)


jurre:
I think this may resolve https://github.com/dependabot/dependabot-core/issues/2532?

jurre (approved):

cargo/lib/dependabot/cargo/file_updater/manifest_updater.rb
```diff
@@ -100,8 +100,9 @@ def update_manifest_req(content:, dep:, old_req:, new_req:)
           if simple_declaration
             simple_declaration_regex =
               /(?:^|["'])#{Regexp.escape(simple_declaration)}/
+            old_req_escped = Regexp.escape(old_req)
```
thepwagner, line 103:
```suggestion
            old_req_escaped = Regexp.escape(old_req)
```

Typo

cargo/lib/dependabot/cargo/file_updater/manifest_updater.rb
```diff
@@ -100,8 +100,9 @@ def update_manifest_req(content:, dep:, old_req:, new_req:)
...
             content.gsub(simple_declaration_regex) do |line|
-              line.gsub(/.+=.*\K(#{old_req})/, new_req)
+              line.gsub(/.+=.*\K(#{old_req_escped})/, new_req)
```
thepwagner, line 105:
```suggestion
              line.gsub(/.+=.*\K(#{old_req_escaped})/, new_req)
```

Typo. It wouldn't let me do a multi-line suggestion sorry

cargo/spec/dependabot/cargo/file_updater/manifest_updater_spec.rb
```diff
@@ -264,6 +264,30 @@
             to include(%([dependencies.pango]\nversion = "0.3.0"\n))
         end
       end
+
+      context "with a version requirement" do
+        context "with a target-specific dependency" do
```
thepwagner, line 269:
```suggestion
        context "with a caret requirement dependency" do
```

I'm borrowing the name from https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements .

thepwagner (approved):

cargo/spec/dependabot/cargo/file_updater/manifest_updater_spec.rb
feelepxyz: Ah yes nice one, folded into above context

lucab:
@feelepxyz @jurre I'm seeing new dependabot PRs flowing in multiple projects I maintain, so I think this PR (or another change that landed in the same timeframe) did indeed fix #2532, which can be closed now.

PR #2713, opened 2020-11-02
composer: increase Dependabot::OutOfMemory

thepwagner:
This aims to improve the behaviour of the `composer` updater while under memory pressure:

Composer updates shell to a helper process implemented in PHP, depending on the project this helper process might end up using several GB of memory.
Previously, there was some code to [enforce memory limits](https://github.com/dependabot/dependabot-core/blob/main/composer/helpers/bin/run#L45-L46) in PHP, that I don't think is used as we always run without specifying the [limit](https://github.com/dependabot/dependabot-core/search?q=memory_limit).

In practice, we're at the whims of the Linux Out of Memory (OOM) Killer, whose algorithm is roughly described [here](https://github.com/torvalds/linux/blob/7c2a69f610e64c8dec6a06a66e721f4ce1dd783a/Documentation/filesystems/proc.rst#31-procpidoom_adj--procpidoom_score_adj--adjust-the-oom-killer-score).
TLDR: since the helper process is likely using more memory than `dependabot-core`'s host process, it's the most likely to be killed. It would be nice to `oom_score_adjust` to encourage the oomkiller to target the helper, but that's not worth adding the permissions to make that possible (e.g. `SYS_RESOURCE` capability and running (at least initially) as root).

To address the OOM killer reality, I've made two adjustments:
* `Dependabot::SharedHelpers::ChildProcessFailed.error_context` has an added `termsig` field, forwarding [`Process::Status#termsig`](https://ruby-doc.org/core-2.7.2/Process/Status.html#method-i-termsig)
* If composer detects the helper process _was_ killed, assume it was the kernel's OOM killer and raise a more appropriate exception

### DIFF
--- a/composer/lib/dependabot/composer/update_checker/version_resolver.rb
+++ b/composer/lib/dependabot/composer/update_checker/version_resolver.rb
@@ -320,6 +320,9 @@ module Dependabot
           elsif error.message.start_with?("Allowed memory size") ||
                 error.message.start_with?("Out of memory")
             raise Dependabot::OutOfMemory
+          elsif error.error_context[:process_termsig] == 9
+            # If the helper was SIGKILL-ed, assume the OOMKiller did it
+            raise Dependabot::OutOfMemory
           elsif error.message.start_with?("Package not found in updated") &&
                 !dependency.top_level?
             # If we can't find the dependency in the composer.lock after an
--- a/composer/lib/dependabot/composer/update_checker/version_resolver.rb
+++ b/composer/lib/dependabot/composer/update_checker/version_resolver.rb
@@ -320,7 +320,8 @@ module Dependabot
           elsif error.message.start_with?("Allowed memory size") ||
                 error.message.start_with?("Out of memory")
             raise Dependabot::OutOfMemory
-          elsif error.error_context[:process_termsig] == 9
+          elsif error.error_context[:process_termsig] ==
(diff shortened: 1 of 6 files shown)


composer/lib/dependabot/composer/update_checker/version_resolver.rb
```diff
@@ -320,6 +320,9 @@ def handle_composer_errors(error)
           elsif error.message.start_with?("Allowed memory size") ||
                 error.message.start_with?("Out of memory")
             raise Dependabot::OutOfMemory
+          elsif error.error_context[:process_termsig] == 9
```
feelepxyz, line 323: Nit: could we add this to constant

```suggestion
          elsif error.error_context[:process_termsig] == Dependabot::SharedHelpers::SIGKILL
```

feelepxyz (approved):
Sweet! This looks 💯 to me and tests make me warm and fuzzy 🧖‍♂️

composer/lib/dependabot/composer/update_checker/version_resolver.rb
thepwagner: Noice! I was hopeful there might be a constant in the stdlib that I was failing to google, a la https://golang.org/pkg/syscall/ . I'll update to declare our own.

jurre (approved):
Cheers for the thorough writeup on this @thepwagner

PR #2725, opened 2020-11-05
Escape SharedHelpers.run_shell_command

feelepxyz:
Escape the command passed to `SharedHelpers.run_shell_command` with shellwords.


jurre:
Looks good, could we add a test to demonstrate how these will be escaped and to make sure we don't regress?

feelepxyz:
@jurre yeah great shout, pushed some tests.

common/lib/dependabot/shared_helpers.rb
```diff
@@ -262,17 +265,18 @@ def self.reset_global_git_config(backup_path)
       FileUtils.mv(backup_path, GIT_CONFIG_GLOBAL_PATH)
     end
 
-    def self.run_shell_command(command)
+    def self.run_shell_command(command, escape_command_str: true)
```
jurre, line 268: I think this is good as is, but I was wondering if a method name that's very explicit about this being unsafe would be even better?

```
def self.run_unsafe_shell_command(command)
```

jurre (approved):

common/lib/dependabot/shared_helpers.rb
brrygrdn: 👍🏻  to the above, I just wanted to follow up - is there ever any reason we should distinguish between safe/unsafe even in hard-coded internal calls?

My instinct would be to default to principal of least trust at all times _just in case_. We could indirectly allow unsafe strings to get into a 'trusted' `run_shell_command` as a parameter of another method without being able to spot it easily.

common/lib/dependabot/shared_helpers.rb
```diff
@@ -262,17 +265,18 @@ def self.reset_global_git_config(backup_path)
...
       start = Time.now
-      stdout, process = Open3.capture2e(command)
+      cmd = escape_command_str ? escape_command(command) : command
```
brrygrdn, line 270: As mentioned above, my brain pulled the handbrake on this conditional as escaping optionally makes me nervous.

common/spec/dependabot/file_fetchers/base_spec.rb
```diff
@@ -1518,5 +1518,41 @@ def fetch_files
...
+      end
+
+      context "with a branch name including bash command" do
+        let(:branch) do
+          "\"$(time)\""
+        end
+
+        it "clones the repo" do
+          clone_repo_contents
+          expect(`ls #{repo_contents_path}`).to include("time.md")
```
thepwagner, line 1543: Xref https://github.com/dependabot-fixtures/go-modules-app/tree/%22%24(time)%22
I wondered what was going on here before I looked up the fixture.

I thought it might be slick to verify the command is escaped locally, e.g. `let(:branch) { "$(touch #{repo_contents_path}/time.md)" }`, then verify the file is _not_ there.
But I like how this verifies the clone not only blocks the command from running, but clones! Noice!

thepwagner (approved):

common/spec/dependabot/file_fetchers/base_spec.rb
feelepxyz: Updated the spec copy a bit to make this clearer

common/lib/dependabot/shared_helpers.rb
feelepxyz: Pushed a commit to rename the argument `escape_command_str` to `allow_unsafe_shell_command`: https://github.com/dependabot/dependabot-core/pull/2725/commits/b36a5a0d972a51954b1a913c4c46275a5ee67a25

Thought on this?

> We could indirectly allow unsafe strings to get into a 'trusted' run_shell_command as a parameter of another method without being able to spot it easily. 

Not sure I follow, did you have an example in mind given the current design?

brrygrdn: I think the last commit solves the thing that made me itchy, I was looking at the calls in `configure_git_to_use_https`, e.g.

```
      run_shell_command(
        "git config --global --add url.https://github.com/."\
        "insteadOf git@github.com:"
      )
```

My thought was that if this method was ever parameterised (e.g. force https access to a private source ) the author/reviewer would need to be diligent to consider escaping these strings.

The change to break that method up into several calls to `run_shell_command` and making it escape by default limits the risk to configuring the credentials [here](https://github.com/dependabot/dependabot-core/pull/2725/files#diff-219bebe959e95ebf38d9a9fa9225f9df2cf12195eab2ee0e04054672ff178a0dR210)

That covers what I was concerned about well 👍🏻

brrygrdn (approved):

common/lib/dependabot/shared_helpers.rb
feelepxyz: > My thought was that if this method was ever parameterised

Yep, good call 👍 

> That covers what I was concerned about well 👍🏻

Thanks for checking 👌

common/lib/dependabot/shared_helpers.rb
```diff
@@ -176,15 +176,23 @@ def self.configure_git_to_use_https
...
+        "git config --global --add url.https://github.com/."\
+        "insteadOf git@github.com:"
+      )
+      run_shell_command(
+        "git config --global --add url.https://github.com/."\
+        "insteadOf git@github.com/"
+      )
+      run_shell_command(
+        "git config --global --add url.https://github.com/."\
         "insteadOf git://github.com/"
```
feelepxyz, line 196: Checked these commands work without quoting the url

PR #2726, opened 2020-11-05
Adjust rubocop max line length to 120

jurre:
As we've started doing this in other projects, I figured it would be
nice to stay consistent between them and also update the max line length
to 120 in dependabot-core. I've ran rubocop autocorrect on all files,
but have not gone in to make any manual changes.


feelepxyz (approved):
🙌

PR #2732, opened 2020-11-09
Explain update not possible for yarn and npm

jurre:
This uses [arborist](https://github.com/npm/arborist) to parse the
dependency tree, find any packages that depend on the dependency we're
trying to update, but have a version conflict.


npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/dependency_files_builder.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+
+module Dependabot
+  module NpmAndYarn
+    class UpdateChecker
+      class DependencyFilesBuilder
```
feelepxyz, line 6: Should we add a test for this?

jurre: Yep, I'll get on that 👍

npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker/conflicting_dependency_resolver_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+        target_version: target_version
+      )
+    end
+
+    it "returns the right array of blocking dependencies" do
+      expect(conflicting_dependencies).to match_array(
+        [
+          {
+            "name" => "objnest",
+            "version" => "^4.1.2",
```
feelepxyz, line 61: Should we set this to the installed version instead of the requirement? Did that for yarn and looks like we do the same for bundler

feelepxyz: Had a crack at this but not sure we have access to the installed version?

common/lib/dependabot/shared_helpers.rb
```diff
@@ -90,6 +90,11 @@ def self.run_helper_subprocess(command:, function:, args:, env: nil,
       stdin_data = JSON.dump(function: function, args: args)
       cmd = allow_unsafe_shell_command ? command : escape_command(command)
       env_cmd = [env, cmd].compact
+      if ENV["DEBUG_FUNCTION"] == function
```
feelepxyz, line 93: @jurre thoughts on this? A bit crude but found it helpful

common/lib/dependabot/shared_helpers.rb
```diff
@@ -90,6 +90,11 @@ def self.run_helper_subprocess(command:, function:, args:, env: nil,
...
+        escaped_stdin_data = stdin_data.gsub("\"", "\\\"")
+        puts "$ cd #{Dir.pwd} && echo \"#{escaped_stdin_data}\" | #{env_cmd.join(" ")}"
+        byebug
```
feelepxyz, line 96: Slightly scary adding this in here

npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker/conflicting_dependency_resolver_spec.rb
jurre: Oh, this is the version from the package.json 🤔 yeah I do think we should ideally show the installed version, will look into that as well

common/lib/dependabot/shared_helpers.rb
jurre: I've done something similar to test the native helpers, I think it's OK for now but ideally would love to have a better story for working on those helpers

npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker/conflicting_dependency_resolver_spec.rb
jurre: Pushed a fix, cheers for noticing!

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/dependency_files_builder.rb
```diff
@@ -0,0 +1,94 @@
+# frozen_string_literal: true
+
+require "dependabot/npm_and_yarn/file_updater/npmrc_builder"
+require "dependabot/npm_and_yarn/file_updater/package_json_preparer"
+
+module Dependabot
+  module NpmAndYarn
+    class UpdateChecker
+      class DependencyFilesBuilder
```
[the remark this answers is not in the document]
jurre: https://github.com/dependabot/dependabot-core/pull/2736

common/lib/dependabot/shared_helpers.rb
```diff
@@ -83,13 +83,20 @@ def self.escape_command(command)
...
       start = Time.now
       stdin_data = JSON.dump(function: function, args: args)
       cmd = allow_unsafe_shell_command ? command : escape_command(command)
       env_cmd = [env, cmd].compact
+      if ENV["DEBUG_FUNCTION"] == function
+        escaped_stdin_data = stdin_data.gsub("\"", "\\\"")
+        puts "$ cd #{Dir.pwd} && echo \"#{escaped_stdin_data}\" | #{env_cmd.join(' ')}"
+        # Pause execution so we can run helpers inside the temporary directory
+        byebug # rubocop:disable Lint/Debugger
+      end
```
feelepxyz, line 99: This is a bit of an eye-sore but seems useful to add debugging practices in code and keep iterating towards something better over time

feelepxyz (approved):
LGTM 👍

Tested a bunch of dry-run examples and didn't run into anything weird so think the test coverage is good enough to ship.

I think it's a good think that it fails gracefully and returns no conflicting deps if the native helper blows up. We could query the update job errors to see how many have conflicting deps set if we want to figure out if it's not performing as expected.

PR #2736, opened 2020-11-10
Extract DependencyFileBuilder to remove duplication

jurre:
Originally part of
https://github.com/dependabot/dependabot-core/pull/2732, but as that PR
was getting quite big, I've moved the refactor here. We need this class
in some of the code introduced in that PR, and instead of duplicating it
further, I've opted to extract it.


feelepxyz (approved):
LGTM 👍

PR #2738, opened 2020-11-11
Bugfix for Elixir Code.eval_file

baseballlover723:
Currently there is a bug where if you use `Code.eval_file` in your mix.exs, then dependabot will error out because the second argument is nil. A simple compact solves this issue by removing any nils that the regex catches. I have added a test as well.


jurre (approved):
This is great, thanks for the added test coverage!

PR #2739, opened 2020-11-11
Support Elixir Code.require_file

baseballlover723:
I have a project that requires another file as part of our mix.exs, and dependabot fails due to it not being able to find the required file when upgrading dependencies. This PR adds this functionality by adding another regex similar to how `Code.eval_file` is handled. I have also included tests fo this as well.

### DIFF
--- a/hex/lib/dependabot/hex/file_fetcher.rb
+++ b/hex/lib/dependabot/hex/file_fetcher.rb
@@ -8,9 +8,8 @@ module Dependabot
     class FileFetcher < Dependabot::FileFetchers::Base
       APPS_PATH_REGEX = /apps_path:\s*"(?<path>.*?)"/m.freeze
       STRING_ARG = %{(?:["'](.*?)["'])}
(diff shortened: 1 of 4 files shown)


hex/lib/dependabot/hex/file_fetcher.rb
```diff
@@ -73,6 +76,14 @@ def evaled_files
           fetch_file_from_host(path).tap { |f| f.support_file = true }
         end
       end
+
+      def required_files
```
jurre, line 80: There's a bit of duplication between this and `evaled_files` now, I wonder if we could either:

- extract that into a helper method
- or, update the regex to match both patterns, and update the method name to reflect that it deals with both cases

baseballlover723: I like the 2nd over the 1st, especially since there is already the concept of supporting files. so, if the method was named something like `support_files`, then it would fit quite well, with getting both of these cases. I'll check back tomorrow to see if theres any more discussion on it

jurre: I think I prefer that as well 👍 thanks for your work on this by the way 🎉

jurre (approved):

PR #2740, opened 2020-11-11
update ci jobs env variable assignment

baseballlover723:
I noticed that there were some github action warnings on the ci build of my previous PRs. I've made another PR to fix these warnings, as it appears they will break your CI on Monday.

the warnings (https://github.com/dependabot/dependabot-core/actions/runs/357182402): 
```The `set-env` command is deprecated and will be disabled on November 16th. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,15 +41,15 @@ jobs:
         uses: actions/checkout@v2
       - name: Prepare environment variables
         run: |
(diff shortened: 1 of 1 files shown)


localheinz (approved):
👍

jurre (approved):
Thanks!

PR #2682, opened 2020-10-26
Upgrade to composer v2 for PHP

WyriHaximus:
Resolves: #2668

PR #2741, opened 2020-11-11
Enhancement: Use scenario-specific fixtures

localheinz:
This PR

* [x] splits fixtures that are currently reused for different scenarios into scenario-specific fixtures

Follows https://github.com/dependabot/dependabot-core/pull/2682#issuecomment-722422269.

💁‍♂️ I will pick work on this up later!


jurre:
This is an amazing start, thank you @localheinz!

localheinz:
@jurre 

I have started a couple of times to reorganize the fixtures to stop reusing them across tests, but I have always come to the point where I feel that the tests fixtures are jus terribly organized, and given up. Adopting project-specific fixtures as for other package managers does not make sense to me either when these projects are then reused across different tests as well.

In my experience, the only thing that would make sense here would be to have a single set of files for every single test scenario.

The downside of this approach is that this will introduce a lot of duplicated files, but the upside is that changing fixtures in one place will not fail tests in unexpected places.

Currently, the fixtures are just terribly organized!

PR #2748, opened 2020-11-12
Support reading files in Elixir supporting files like in mixfiles

baseballlover723:
Currently you get an error if you try and read say a version file in a supporting file in an Elixir project. This is handled in the mixfile proper by stubbing out the call to read a file with just a standard string. This PR extends this behavior to supporting files as well. I have also added a test for this as well.

### DIFF
--- a/hex/spec/dependabot/hex/file_updater/lockfile_updater_spec.rb
+++ b/hex/spec/dependabot/hex/file_updater/lockfile_updater_spec.rb
@@ -220,6 +220,26 @@ RSpec.describe Dependabot::Hex::FileUpdater::LockfileUpdater do
       end
     end
 
+    context "with a mix.exs that opens another file in a required file" do
+      let(:mixfile_fixture_name) { "loads_file_with_require" }
+      let(:lockfile_fixture_name) { "exact_version" }
+      let(:files) { [mixfile, lockfile, support_file] }
+      let(:support_file) do
+        Dependabot::DependencyFile.new(
+          name: "module_version.ex",
+          content: fixture("support_files", "module_version"),
+          support_file: true
+        )
+      end
+
+      it "updates the dependency version in the lockfile" do
+        expect(updated_lockfile_content).to include %({:hex, :plug, "1.4.3")
+        expect(updated_lockfile_content).to include(
+          "236d77ce7bf3e3a2668dc0d32a9b6f1f9b1f05361019946aae49874904be4aed"
+        )
+      end
+    end
+
     context "with an umbrella app" do
       let(:mixfile_fixture_name) { "umbrella" }
       let(:lockfile_fixture_name) { "umbrella" }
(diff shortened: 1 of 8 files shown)


baseballlover723:
WIP because it wasn't working properly. I think I've found the fix (applying the sanitization in the version_resolver as well), but I'll update the PR tomorrow with some tests as well as seeing if there are any other places that this shows up.

baseballlover723:
the error gets swallowed here: https://github.com/dependabot/dependabot-core/blob/main/hex/lib/dependabot/hex/update_checker/version_resolver.rb#L93-L97. Basically if mix fails here (such as due to trying to read a file that isn't there) https://github.com/dependabot/dependabot-core/blob/main/hex/lib/dependabot/hex/update_checker/version_resolver.rb#L93-L97 then it just passes nil all the way back, and eventually it gets swallowed somewhere down the line, and it gets interpreted as "can't actually update this dependency". The error is never propagated back up to the user, which made this very difficult to find. I'll make a test and a fix for it tomorrow.

baseballlover723:
I have updated this PR that fixes the bug I ran into yesterday. unfortunately, it looks like there would be quite a bit of work needed to fix surfacing errors like these, so I'll make an issue for it and leave it for you all to fix (if you want to).

hex/spec/dependabot/hex/file_updater/lockfile_updater_spec.rb
```diff
@@ -220,6 +220,26 @@
...
+          name: "module_version.ex",
+          content: fixture("support_files", "module_version"),
+          support_file: true
+        )
+      end
+
+      it "updates the dependency version in the lockfile" do
+        expect(updated_lockfile_content).to include %({:hex, :plug, "1.4.3")
+        expect(updated_lockfile_content).to include(
+          "236d77ce7bf3e3a2668dc0d32a9b6f1f9b1f05361019946aae49874904be4aed"
```
jurre, line 238: Where does this string come from? Maybe we could assign it to a variable that explains what it represents?

```
let(:expected_something_something) { "236d77ce7bf3e3a2668dc0d32a9b6f1f9b1f05361019946aae49874904be4aed" }
```

jurre:
Thanks @baseballlover723, mostly looks great, thanks! Just had 1 question/suggestion

hex/spec/dependabot/hex/file_updater/lockfile_updater_spec.rb
baseballlover723: this was copied from the tests above it. I'll leave it to you all to decide if you want to extract it out into a variable though.

jurre: Ah, right, I think we should but we can follow up on that separately, thanks!

jurre (approved):

PR #2752, opened 2020-11-12
Check Azure PR description against utf-16 encoded length

jurre:
Azure Devops uses a character limit of 4000 characters in UTF-16
encoding, instead of Ruby's default UTF-8 encoding. This causes issues
with special characters like Emoji, which count as 2 characters in
UTF-8, but as 4 characters in UTF-16.

### DIFF
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -8,6 +8,8 @@ module Dependabot
     class Azure
       class NotFound < StandardError; end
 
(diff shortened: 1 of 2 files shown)


aldodfm:
thx @jurre , let's seee, when this will be merged, i will test with the last Dependabot core docker image verison. ;), just tell me when.

aldodfm:
:'(

aldodfm:
seems that now is ready for merge? =)

feelepxyz (approved):
👌

PR #2754, opened 2020-11-12
Include an explanation msg for conflicting dependencies

jurre:
In order to give users more context around which dependencies are
conflicting, we include a message with a human readable explanation.

For `npm`, we also include the top-level dependency that caused the conflict.


npm_and_yarn/helpers/lib/npm/conflicting-dependency-parser.js
```diff
@@ -18,22 +18,28 @@ async function findConflictingDependencies(directory, depName, targetVersion) {
...
+              parents.push({
+                name: edge.from.name,
+                version: edge.from.version,
+                requirement: edge.spec,
+              });
+            } else {
+              parents.push({
+                name: topLevel.to.name,
+                version: topLevel.to.version,
+                subdependency: {
```
feelepxyz, line 36: Should this also include `requirement` for the top level dep so we don't break the existing error rendering? I guess we'll have a period where we'll only show the top level parent depending on the vulnerable dependency which seems fine.

Small nit, wondering about the naming `subdependency` as we don't use that anywhere in core except for spec fixtures. Should it just be `dependency`?

jurre: > Should this also include requirement for the top level dep so we don't break the existing error rendering?

We can't use this structure as is in the existing rendering, if we would include the requirement there, it would be the requirement on the subdependency, but we don't always have that, since the edge between the top-level dependency and the immediate parent of the dependency we're trying to update could be nested deeply.

feelepxyz: > We can't use this structure as is in the existing rendering

Ah yes, in which case can we ship this without breaking the existing rendering?

jurre: I think we'd have to teach the rendering to deal with this change first. I'm looking into returning a flat message instead

jurre:
@feelepxyz I've pushed a change that adds an `explanation` string that we can display directly

npm_and_yarn/helpers/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+When working on these helpers, it's convenient to write some high level tests in
+JavaScript to make it easier to debug the code.
+
+In order to run an interactive debugger:
+
+- `node --inspect-brk node_modules/.bin/jest --runInBand test/npm/conflicting-dependency-parser.test.js`
+- In Chrome, nativate to chrome://inspect
+- Click `Open dedicated DevTools for Node`
+- You'll now be able to interactively debug using the chrome dev tools.
```
feelepxyz, line 21: Thanks for adding this! 🙌

feelepxyz (commented):
👍  on adding a `explanation` field with the string. My only hesitation is around the error message formatting having it nested without much explanation. Worth creating an issue and tagging `infin8x` and `mamuso` (designer) to get some feedback on the suggested format?

npm_and_yarn/helpers/lib/npm/conflicting-dependency-parser.js
```diff
@@ -43,4 +40,31 @@ async function findConflictingDependencies(directory, depName, targetVersion) {
...
+  if (directEdge.from === topLevelEdge.to) {
+    // The nodes parent is top-level
+    return `${directEdge.from.name}@${directEdge.from.version} requires ${directEdge.to.name}@${directEdge.spec}`;
+  } else if (topLevelEdge.to.edgesOut.has(directEdge.from.name)) {
+    // The nodes parent is a direct dependency of the top-level dependency
+    return `${topLevelEdge.to.name}@${topLevelEdge.to.version} requires ${directEdge.to.name}@${directEdge.spec} via ${directEdge.from.name}@${directEdge.from.version}`;
+  } else {
+    // The nodes parent is a transitive dependency of the top-level dependency
+    return `${topLevelEdge.to.name}@${topLevelEdge.to.version} requires ${directEdge.to.name}@${directEdge.spec} via a transitive dependency on ${directEdge.from.name}@${directEdge.from.version}`;
+  }
```
feelepxyz, line 53: Yeah I like how this focuses on the requirement on the vulnerable/target dependency first 💯

npm_and_yarn/helpers/lib/yarn/conflicting-dependency-parser.js
```diff
@@ -24,6 +24,7 @@ async function findConflictingDependencies(directory, depName, targetVersion) {
         if (subDepName === depName && !semver.satisfies(targetVersion, spec)) {
           const [_, parentDepName] = entry.match(LOCKFILE_ENTRY_REGEX);
           parents.push({
+            description: `${parentDepName}@${pkg.version} requires ${depName}@${spec}`,
```
feelepxyz, line 27:
```suggestion
            explanation: `${parentDepName}@${pkg.version} requires ${depName}@${spec}`,
```

feelepxyz: Worth adding the same js tests that you have for npm?

npm_and_yarn/helpers/lib/npm/conflicting-dependency-parser.js
```diff
@@ -43,4 +40,31 @@ async function findConflictingDependencies(directory, depName, targetVersion) {
...
+function buildExplanation(node, directEdge, topLevelEdge) {
+  if (directEdge.from === topLevelEdge.to) {
+    // The nodes parent is top-level
+    return `${directEdge.from.name}@${directEdge.from.version} requires ${directEdge.to.name}@${directEdge.spec}`;
+  } else if (topLevelEdge.to.edgesOut.has(directEdge.from.name)) {
+    // The nodes parent is a direct dependency of the top-level dependency
+    return `${topLevelEdge.to.name}@${topLevelEdge.to.version} requires ${directEdge.to.name}@${directEdge.spec} via ${directEdge.from.name}@${directEdge.from.version}`;
+  } else {
+    // The nodes parent is a transitive dependency of the top-level dependency
+    return `${topLevelEdge.to.name}@${topLevelEdge.to.version} requires ${directEdge.to.name}@${directEdge.spec} via a transitive dependency on ${directEdge.from.name}@${directEdge.from.version}`;
```
feelepxyz, line 52: Maybe cut up the line a bit?
```suggestion
    return `${topLevelEdge.to.name}@${topLevelEdge.to.version} requires ${directEdge.to.name}@${directEdge.spec} ` +
               `via a transitive dependency on ${directEdge.from.name}@${directEdge.from.version}`;
```

npm_and_yarn/helpers/lib/npm/conflicting-dependency-parser.js
```diff
@@ -43,4 +40,31 @@ async function findConflictingDependencies(directory, depName, targetVersion) {
   });
 }
 
+function buildExplanation(node, directEdge, topLevelEdge) {
+  if (directEdge.from === topLevelEdge.to) {
+    // The nodes parent is top-level
+    return `${directEdge.from.name}@${directEdge.from.version} requires ${directEdge.to.name}@${directEdge.spec}`;
+  } else if (topLevelEdge.to.edgesOut.has(directEdge.from.name)) {
+    // The nodes parent is a direct dependency of the top-level dependency
+    return `${topLevelEdge.to.name}@${topLevelEdge.to.version} requires ${directEdge.to.name}@${directEdge.spec} via ${directEdge.from.name}@${directEdge.from.version}`;
```
feelepxyz, line 49:
```suggestion
    return `${topLevelEdge.to.name}@${topLevelEdge.to.version} requires ${directEdge.to.name}@${directEdge.spec} ` +
           `via ${directEdge.from.name}@${directEdge.from.version}`;
```

package.json
```diff
@@ -0,0 +1,14 @@
+{
```
feelepxyz, line 1: Accidental commit? ✂️ this and `package-lock.json`?

npm_and_yarn/helpers/lib/npm/conflicting-dependency-parser.js
jurre: Yeah good call, prettier doesn't like that specific suggestion, but I've pushed an alternative 👍

bundler/helpers/lib/functions/conflicting_dependency_resolver.rb
```diff
@@ -20,6 +20,7 @@ def conflicting_dependencies
       parent_specs.map do |spec|
         req = spec.dependencies.find { |bd| bd.name == dependency_name }
         {
+          "explanation" => "#{spec.name} (#{spec.version}) requires #{dependency_name} (#{req.requirement})",
```
feelepxyz, line 23: Should we align these to npm/yarn? Any reason they are different?
```suggestion
          "explanation" => "#{spec.name}@#{spec.version} requires #{dependency_name}@#{req.requirement}",
```

feelepxyz: Hm might look weird with bundler notation though, e.g. `activesupport@= 5.0.0`, thoughts about aligning npm/yarn with bundler?

feelepxyz: npm semver could also get real weird, e.g. `objnest@^2 <2.2 || > 2.3`

feelepxyz: Leaning towards unifying around the bundler format

jurre: I tried to pick notations that are common/native to the ecosystem, although maybe the `@` notation is only common in yarn: `etag@^1.0.0`, and not so much in npm.

feelepxyz: Ah yeah good point, `@` is definitely more common to npm/yarn and "feels" more in line with those ecosystems. I'm up for keeping the current format as is 👍

feelepxyz: Another idea would be to always `@` exact versions and have requirements in brackets for all ecosystems to handle weird cases where the requirement is unruly

`dummy-pkg-b@1.0.0 requires dummy-pkg-a (< 2.0.0 || > 1.5.0) `

jurre: Might be more clear to just use the `(1.0.0)` format everywhere in that case?

feelepxyz (approved):
LGTM 💯  minor thought/comment about aligning the format between package managers but not blocking. Looks like there's a spec missing for yarn's conflicting dependency parser but  not blocking, wouldn't add any coverage over the existing ruby test.

bundler/helpers/lib/functions/conflicting_dependency_resolver.rb
feelepxyz: > Yarn does use the `@` notation for that though:

Good point, forgot it's in the lockfile format 👍  yeah seems fine to keep as is tbh, nice to align format to package manager

PR #2755, opened 2020-11-12
Fix: Use consistent names

localheinz:
This PR

* [x] uses consistent naming across specifications for `composer`

Related to #2741.

💁‍♂️ I believe using consistent naming here will make it easier to split out the scenarios - what do you think?

### DIFF
--- a/composer/spec/dependabot/composer/file_fetcher/path_dependency_builder_spec.rb
+++ b/composer/spec/dependabot/composer/file_fetcher/path_dependency_builder_spec.rb
@@ -10,13 +10,13 @@ RSpec.describe Dependabot::Composer::FileFetcher::PathDependencyBuilder do
     described_class.new(
       path: path,
       directory: directory,
-      lockfile: composer_lock
+      lockfile: lockfile
     )
   end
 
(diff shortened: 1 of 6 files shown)


localheinz:
@jurre 

If you like this, or have any alternative recommendations for consistent names, I will be happy to hear them, adjust this pull request (when necessary), and look forward to getting this merged!

I will then immediately continue to work on #2741!

jurre:
@localheinz, it seems the naming was already mostly consistent? We use `lockfile` and `manifest_file` across ecosystems and also in the code, so on the one hand it's kind of nice having that shared terminology. What were some of the inconsistencies you found?

localheinz:
> @localheinz, it seems the naming was already mostly consistent? We use lockfile and manifest_file across ecosystems and also in the code, so on the one hand it's kind of nice having that shared terminology. What were some of the inconsistencies you found?

Yes, almost - I'll be happy to use consistent naming, perhaps

- `manifest_file`
- `lock_file`

make sense, then?

localheinz:
@jurre 

I have adjusted to consistently use `manifest` and `lockfile`.

PR #2756, opened 2020-11-13
fix incorrect spec and mixfile fixture

baseballlover723:
I came across this spec that was setting the wrong let variables, and thus wasn't being run properly.

The mixfile was syntactically invalid, so I fixed it.


feelepxyz (approved):
👍  Thanks for fixing this 🙇

PR #2760, opened 2020-11-13
Optimize the CI dockerfile to let docker cache more aggressively

baseballlover723:
Moved the bundle installing to before copying over all of the code. This allows docker to cache the more expensive bundle install commands (in particular the common one, since its first). This is particularly helpful if you test code locally with the dockerfile, since you don't have to wait for the long bundle install commands to execute while change the source code.


baseballlover723:
seems like this is entirely ineffective, so I'll close this PR

Issue #1162, opened 2019-05-23
Tracking: New ecosystem requests


rurban:
There are many individual requests to support to new package managers. Since it's now integrated into github I made my own list TODO:
*(in order of importance)*

* cmake
* pkg-config
* perl:EUMM
* perl:cpanfile
* perl:MB
* ~~nuget~~
* c:ccan
* deb
* rpm
* c:clib

but there are many more. I'll start working on some as need arises

greysteil:
NuGet is already supported! 🎉 

I'm keen to add support for the others, too. But first I want to get PHP support into GitHub security alerts, so that we can start creating automated security fixes for it for all GitHub users.

rojepp:
For .Net, I'd like to see [Paket](https://fsprojects.github.io/Paket/) support. Most of my repos use Paket as a NuGet replacement.

greysteil:
@rojepp agreed! There's an old issue for it [here](https://github.com/dependabot/feedback/issues/363), but Dependabot-Core is the place to add it.

malramsay64:
There are a number of additional issues requesting support for new package managers which I thought  might be useful to collate here.

- Crystal #1237 
- Swift #1245 
- Conda #1249 
- Java:Bazel #1215 
- asdf #1033 
- appraisal #978 
- rpm #847 
- cake-build #733 WIP #958 
- CocoaPods #416 WIP #731 
- Docker compose #390

orj:
Swift package support would be great.

brl0:
Just a quick note that the issue listed for conda above was closed as a duplicate in favor of #2227.  

IMHO conda support, and possibly the derivative or related miniconda, miniforge, mamba as well, would be very helpful for the python ecosystem.

jeffwidman:
👋 Hey there! Sorry for the glacially slow response... we've been a bit buried.

I'm going to close this, as essentially it's identical to what you get by searching for the `T: new-ecosystem` label: https://github.com/dependabot/dependabot-core/labels/T%3A%20new-ecosystem

Having individual issues for each requested ecosystem makes it easier to track / discuss them individually.

Issue #572, opened 2018-07-04
Clojure support


greysteil:
Originally suggested in https://github.com/dependabot/feedback/issues/76.

camdez:
What's the latest here? Just looking for someone to do the work, or is something in flight? Happy to help!

devn:
Same, I'm new to dependabot, but I am happy to help in any way I can.

alexisvincent:
What is the process to get support for Clojure?

gphilipp:
https://github.com/dependabot/dependabot-core/blob/main/CONTRIBUTING.md#contributing-new-ecosystems
<img width="910" alt="image" src="https://user-images.githubusercontent.com/56603/120471365-86464a00-c39c-11eb-8d19-5c6290bbcf51.png">

steffan-westcott:
Congratulations to @0tza for getting [some funding from Clojurists Together](https://www.clojuriststogether.org/news/q3-2021-funding-announcement/) 😀 I hope the situation outlined in the post above improves and community contributions for Clojure support will be accepted in the near future.

greysteil:
Cc @asciimike @hmarr ☝️

asciimike:
Congrats on the funding, and thanks to the folks who picked Dependabot as an area of investment. We're glad to see the community invest in security!

As for the above, as stated, the main issue has been that we lack the domain expertise to support the full breadth of contributed ecosystems. We have been working with another popular ecosystem on a closer partnership experience where they can provide that level of support, and would consider a similar relationship with the Clojure community. At a minimum, we are asking for two folks who will commit to being active in this repo if/when Clojure questions come up (we will likely tag you); otherwise, we don't think we can provide an acceptable level of support for devs when they have questions.

danielcompton:
Hi @asciimike! I’m a recent ex-Hubber, admin for the Clojure community Maven repo Clojars, and the secretary of Clojurists Together which funded @0tza for this work. I’m happy to help on providing a close relationship with GitHub here 😀

asciimike:
@danielcompton @0tza I see that the project is listed as starting 1 Nov 2021; how firm is that date? We're starting to do quarterly planning and I want to get a rough idea of when we should start talking about this in more depth.

tobias:
Hello @asciimike. I also am an admin for Clojars, the Clojure community Maven repository. I would be happy to provide a second set of eyes on any Clojure (& Clojure dependency ecosystem) questions.

[...]

PR #2769, opened 2020-11-17
Clojure (Leiningen) Support

CGA1123:
👋  This code is not 100% but want to open this PR up for any feedback and to gauge whether there would be any appetite for getting `lein` support upstreamed into the dependabot-core and eventually the product!

We've been running this at work for the last couple weeks (via a version of [`dependabot-lein-runner`](https://github.com/CGA1123/dependabot-lein-runner)) and its been working a charm

This implementation mostly piggy-back off of the `maven` implementation, with some lein specific `FileFetcher` and `FileUpdater` (and also a `FileParser`, but only to set the `package_manager` as `lein` so that PRs have the proper label. 

Related: #572 #1162


lein/helpers/src/helpers/core.clj
```diff
@@ -0,0 +1,44 @@
...
+  (-> file
+      (zip/of-string)
+      (zip/find-value zip/next 'defproject)
+      (zip/find-value :dependencies)
+      (zip/right)
+      (#(zip/map (partial update-dependency dependencies) %))
+      (zip/root-string)))
+
+(defn generate-pom [{:keys [file]}]
+  (let [proj (project/read-raw (io/reader (char-array file)))]
```
danielcompton, line 33: Will this execute code that lives in the `project.clj` file? I think so? Not sure if that's safe in this context or not.

https://github.com/technomancy/leiningen/blob/830d19d5b9f9e2b0e3411428d83422791ca2f814/leiningen-core/src/leiningen/core/project.clj#L1094-L1110

jurre: I think that's mostly fine, as long as the code that's executed is from the source repository. The same thing is possible in a `setup.py`, `Gemfile` or `mix.exs` file. My main worry would be if an external package would be able to execute code in this context, but I don't think that's the case?

jurre:
Thanks for the PR @CGA1123, overall the approach looks solid, awesome work!

Currently the team is stretched a bit thin, and we won't be able to take on reviewing + supporting this new ecosystem, but we hope to have a bit more bandwidth for that in a few weeks.

One thing that would be great to see is some test coverage for the bits that are newly implemented, but that's something we can also look into once we have time to properly review + test this.

CGA1123:
@jurre No worries 😄 

And yes will be trying to carve out some time myself to get some test coverage

lein/helpers/src/helpers/core.clj
danielcompton: That's correct, downloading POMs (Maven/Clojure dependency manifest) and JARs (dependency files) doesn't execute any code. Only the code that exists in the project would be executed.

danielcompton:
Side note: If this approach works for Leiningen, you could do something very similar for `deps.edn` files. `clj -X:deps mvn-pom` will generate a POM given a `deps.edn` file.

J-Vin (approved):

brrygrdn:
:wave: Thanks for opening this PR, we really appreciate the effort to make Dependabot better for the community.

Unfortunately, we are not accepting new ecosystems into core at the moment. We are currently focused on making some improvements into how we handle extensibility before we add any more ecosystems. We will still be improving and upgrading support for existing ecosystems.

Please check the [Contribution guidelines](https://github.com/dependabot/dependabot-core/blob/main/CONTRIBUTING.md)
for more information.

CGA1123:
@brrygrdn Thanks for the update, makes sense. 😄

CGA1123:
Closing as part of some new years cleaning 😄 

If people are interested in running a Clojure dependabot for their org / themselves check out [dependabot-lein-runner](https://github.com/CGA1123/dependabot-lein-runner) which achieves that and feel free to open issues and i'll help out as much as time allows 🙂

colinphill-mdsol:
It looks like new ecosystems are now being accepted as of #8844.

My team would benefit greatly from first-class Leiningen support. Unfortunately, we're bound by a somewhat restrictive OSS contributions policy, and we don't have Ruby knowledge internally. If this PR were reopened, would it be mergeable as-is, or have the intervening 3.5 years made further changes necessary? I could try to get approval to take ownership of this and muddle my way through the Ruby, if the necessary changes are simply enough.

PR #2772, opened 2020-11-17
Yarn: Explain conflicting top-level dependency

feelepxyz:
Include the top-level dependency name and version when expalining
"update not possible" for yarn bringing the implementation in line with
npm and bundler.

Refactoring the existing logic for yarn  to recursively parse the
lockfile starting from the top-level dependency finding all conflicting
parents that prevent the target dependency from being installed.

Example when the top-level dependency has a direct dependency on the
conflicting parent dependency:

```
askconfig@4.0.4 requires abind@^1.0.4 via objnest@5.0.10
```

Example when the top-level dependency has a transitive dependency on the
conflicting parent dependency:

```
apass@1.1.0 requires abind@^1.0.0 via a transitive dependency on objnest@3.0.9
```


jurre (approved):
Awesome work 👍

PR #2780, opened 2020-11-20
Go modules: raise GitDependenciesNotReachable

thepwagner:
Sharing snapshot to discuss the approach:

When Dependabot encounters a Go module backed by a Git repository, especially one on `github.com`, it should raise `Dependabot::GitDependenciesNotReachable` with `dependency-urls` details describing the missing repository. This matters as the rich exception provides more detail about what went wrong and potential recovery options.

This PR accomplishes this by intercepting the error thrown from the "UpdateChecker", and detecting `github.com/.*` urls. If we detect the failed dependency was on `github.com`, we assume it's a "git dependency" and raise accordingly.

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_parser.rb
+++ b/go_modules/lib/dependabot/go_modules/file_parser.rb
@@ -139,8 +139,9 @@ module Dependabot
       def handle_parser_error(path, stderr)
         case stderr
         when /go: .*: unknown revision/m
-          line = stderr.lines.grep(/unknown revision/).first
-          raise Dependabot::DependencyFileNotResolvable, line.strip
+          line = stderr.lines.grep(/unknown revision/).first.strip
+          handle_github_unknown_revision(line) if line.start_with?("go: github.com/")
+          raise Dependabot::DependencyFileNotResolvable, line
         when /go: .*: unrecognized import path/m
           line = stderr.lines.grep(/unrecognized import/).first
           raise Dependabot::DependencyFileNotResolvable, line.strip
@@ -153,6 +154,26 @@ module Dependabot
         else
           msg = stderr.gsub(path.to_s, "").strip
           raise Dependabot::DependencyFileNotParseable.new(go_mod.path, msg)
+        end
+      end
+
+      GITHUB_REPO_REGEX = %r{github.com/[^@]*}.freeze
+      def handle_github_unknown_revision(line)
+        repo_path = line.scan(GITHUB_REPO_REGEX).first
+        return unless repo_path
+
+        # Query for _any_ version of this module, to know if it doesn't exist (or is private)
+        # or we were just given a bad revision by this manifest
+        SharedHelpers.in_a_temporary_directory do
+          SharedHelpers.with_git_configured(credentials: credentials) do
+            File.write("go.mod", "module dummy\n")
+
+            env = { "GOPRIVATE" => "*" }
+            _, _, status = Open3.capture3(env, SharedHelpers.escape_command("go get #{repo_path}"))
+            raise Dependabot::DependencyFileNotResolvable, line if status.success?
+
+            raise Dependabot::GitDependenciesNotReachable, [repo_path]
+          end
         end
       end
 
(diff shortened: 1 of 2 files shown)


go_modules/lib/dependabot/go_modules/update_checker.rb
```diff
@@ -86,6 +88,11 @@ def find_latest_resolvable_version
 
       def handle_subprocess_error(error)
         if RESOLVABILITY_ERROR_REGEXES.any? { |rgx| error.message =~ rgx }
+          if error.message.start_with?("module github.com/")
```
jurre, line 91: Should we even care about enterprise at this point? Probably YAGNI for now

go_modules/lib/dependabot/go_modules/update_checker.rb
```diff
@@ -86,6 +88,11 @@ def find_latest_resolvable_version
...
+            raise Dependabot::GitDependenciesNotReachable, [
+              error.message.scan(%r{github.com/[^:]*}).last
```
jurre, line 93: Nit: might pull the regex out into a constant

jurre (approved):
This seems like a reasonable approach to me 👍

go_modules/lib/dependabot/go_modules/update_checker.rb
thepwagner: AFAICT the only hook we have for that is the `api_endpoint` for octokit?: https://github.com/dependabot/dependabot-core/blob/05649b25af72d406b144965b6498167351547081/common/lib/dependabot/clients/github_with_retries.rb#L38

Getting the `Dependabot::Source` out of the `Dependabot::Dependency` that's passed here [seems complicated](https://github.com/dependabot/dependabot-core/blob/05649b25af72d406b144965b6498167351547081/common/lib/dependabot/git_commit_checker.rb#L233-L249), converting the API endpoint to the module path is another problem.

I'm also not sure about assuming GHES customers include the server URL in included module paths? 🤷 
For Enterprise-y reasons, I may want to host my server at `github.acme.com`, but namespace my modules as `acme.com/**`?

Definitely YAGNI.

fatih: This looks good for now, but I'm wondering the implications for modules hosted on GitHub but have a different module name by using a proxy.  We're probably going to found out once we start using this in production.

fatih (approved):
Looks good to me 👍🏼

go_modules/lib/dependabot/go_modules/file_parser.rb
```diff
@@ -156,6 +157,26 @@ def handle_parser_error(path, stderr)
         end
       end
 
+      GITHUB_REPO_REGEX = %r{github.com/[^@]*}.freeze
+      def handle_github_unknown_revision(line)
+        repo_path = line.scan(GITHUB_REPO_REGEX).first
+        return unless repo_path
```
thepwagner, line 163:
```suggestion
```

Nit: maybe too defensive? Given the guard on https://github.com/dependabot/dependabot-core/pull/2780/files#diff-5473cd4fd4f2f0f6b13968468fc3a0109a7769f9c729e4ad36a03ac8f419a660R143 this is never true.

go_modules/lib/dependabot/go_modules/file_parser.rb
```diff
@@ -139,8 +139,9 @@ def go_mod_content
       def handle_parser_error(path, stderr)
         case stderr
         when /go: .*: unknown revision/m
-          line = stderr.lines.grep(/unknown revision/).first
-          raise Dependabot::DependencyFileNotResolvable, line.strip
+          line = stderr.lines.grep(/unknown revision/).first.strip
+          handle_github_unknown_revision(line) if line.start_with?("go: github.com/")
```
thepwagner, line 143: Question: should this behaviour still be specific to `github.com/.*` urls?

Raising a typed exception for inaccessible module dependencies should be useful regardless of host, and the pattern seems portable.

I kept this filter for defensiveness.

jurre: In theory they could be mercurial or bazaar deps?

go_modules/lib/dependabot/go_modules/file_parser.rb
jurre: If we keep `line.start_with?("go: github.com/")` seems like we could get rid of it, but I don't mind the check either

jurre (approved):

PR #2791, opened 2020-11-25
go: verify GitDependenciesNotReachable from versioned modules

thepwagner:
No functional changes, but it may be worth verifying this behaviour with versioned modules?


fatih (approved):
Looks good to me 👍🏼

jurre (approved):

PR #2797, opened 2020-11-27
Pip compile: handle unresolvable errors

feelepxyz:
Check the cause exception when checking resolvability for pip compile.

When rescuing pip compile errors we retry the update on the initial
manifest files to see if the project is unresolvable to begin with, when
checking the exception parse the causing exception for resolvability
errors instead of the new exception that might have a different error
message.

Other changes:
- dry-run: add exception handling and re-raise on unknown errors
- Spec helper: add a project dependency files helper


bin/dry-run.rb
```diff
@@ -322,6 +322,96 @@ def cached_dependency_files_read
...
+          "go-mod": error.go_mod,
+        },
+      }
+    else
+      raise error
+    end
+
+  puts " => handled error whilst updating #{dependency.name}: #{error_details.fetch(:'error-type')} "\
+       "#{error_details.fetch(:'error-detail')}"
+end
```
brrygrdn, line 413: ✨

brrygrdn (commented):
🚀

bin/dry-run.rb
jurre: Might be time to extract some things out of this file soon 😄

jurre (approved):

brrygrdn (approved):

bin/dry-run.rb
feelepxyz: > Might be time to extract some things out of this file soon

Yeah! Will leave for now to avoid a big jak shave refactoring the dry-run script 😅

PR #2807, opened 2020-11-30
Add rubocop config for each language

feelepxyz:
At some point rubocop removed support for finding configs in parent
directories and only loads the .rubocop.yml defined in the current
folder so we used to run CI lint with `-c ../.rubocop.yml` but this
breaks editors (e.g. vscode) support as it needs to run rubocop in the
package manager folder.


jurre (approved):
Thanks, this is a great improvement! Not having my editor autocorrect changes has added some unnecessary churn on PRs

PR #2815, opened 2020-12-02
Fix building native python dependencies for pip-compile

jurre:
Some native python dependencies require to be built from scratch,
because no Wheel is available for them for the given version of Python.

If this is the case, when trying to install with the `--build-isolation`
flag, pip requires all the dependencies are available as wheels, but
this is often not the case.

Building with the `--no-build-isolation` ensures that any transitive
dependencies are also installed/compiled from source.


python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
```diff
@@ -91,10 +93,22 @@ def fetch_latest_resolvable_version_string(requirement:)
                 parse_updated_files
               end
             rescue SharedHelpers::HelperSubprocessFailed => e
+              retry_count ||= 0
+              retry_count += 1
+
+              if compilation_error?(e) && retry_count <= 2
```
feelepxyz, line 99: Did you spot cases that needed two retries or worth just doing one without build isolation?

jurre: Yeah, maybe one is just fine. I cargo culted this from places where we retry net requests, but this should probably just work the first time or something else is up 👍

jurre:
We'll probably want to do something similar for regular pip builds, but no reason we should hold off merging the fix for pip compile IMO.

feelepxyz (approved):
LGTM 👍

PR #2821, opened 2020-12-04
Add error message to dependency file not parseable

feelepxyz:
Adds the exception message to `DependencyFileNotParseable` errors and adds sanitisation on all `DependabotError` messages redacting anything that looks like a `http(s)` link, basic auth creds `//username:token@` and any path from private source exceptions.

We might end up redacting more than we want but seems ok to start strict and walk back changes if we get reports of error logs missing crucial debugging information.


feelepxyz:
Splitting the sanitization changes into a seprate PR: https://github.com/dependabot/dependabot-core/pull/2828

PR #2828, opened 2020-12-08
Sanitize creds from DependabotError

feelepxyz:
All errors now inherit from `DependabotError` and sanitize the message from http basic creds in urls.

### DIFF
--- a/common/lib/dependabot/errors.rb
+++ b/common/lib/dependabot/errors.rb
@@ -1,24 +1,45 @@
 # frozen_string_literal: true
 
-require "dependabot/shared_helpers"
+require "dependabot/utils"
 
 module Dependabot
   class DependabotError < StandardError
-    def initialize(msg = nil)
-      msg = sanitize_message(msg)
-      super(msg)
+    BASIC_AUTH_REGEX = %r{://(?<auth>[^:]*:[^@%\s]+(@|%40))}.freeze
+    # Remove any path segment from fury.io sources
+    FURY_IO_PATH_REGEX = %r{fury\.io/(?<path>.+)}.freeze
+
+    def initialize(message = nil)
+      super(sanitize_message(message))
     end
 
     private
 
     def sanitize_message(message)
-      return unless message
+      return message unless message.is_a?(String)
 
       path_regex =
-        Regexp.escape(SharedHelpers::BUMP_TMP_DIR_PATH) + "\/" +
-        Regexp.escape(SharedHelpers::BUMP_TMP_FILE_PREFIX) + "[^/]*"
+        Regexp.escape(Utils::BUMP_TMP_DIR_PATH) + "\/" +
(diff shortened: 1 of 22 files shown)


common/lib/dependabot/shared_helpers.rb
```diff
@@ -8,34 +8,19 @@
...
-    BUMP_TMP_FILE_PREFIX = "dependabot_"
-    BUMP_TMP_DIR_PATH = "tmp"
     GIT_CONFIG_GLOBAL_PATH = File.expand_path("~/.gitconfig")
     USER_AGENT = "dependabot-core/#{Dependabot::VERSION} "\
                  "#{Excon::USER_AGENT} ruby/#{RUBY_VERSION} "\
                  "(#{RUBY_PLATFORM}) "\
                  "(+https://github.com/dependabot/dependabot-core)"
     SIGKILL = 9
 
-    class ChildProcessFailed < StandardError
```
feelepxyz, line 24: Unused

common/lib/dependabot/errors.rb
```diff
@@ -15,10 +19,27 @@ def sanitize_message(message)
       return unless message
 
       path_regex =
-        Regexp.escape(SharedHelpers::BUMP_TMP_DIR_PATH) + "\/" +
-        Regexp.escape(SharedHelpers::BUMP_TMP_FILE_PREFIX) + "[^/]*"
+        Regexp.escape(Utils::BUMP_TMP_DIR_PATH) + "\/" +
+        Regexp.escape(Utils::BUMP_TMP_FILE_PREFIX) + "[^/]*"
```
feelepxyz, line 23: Changed this to not have a circular dependency between errors <> shared helpers

jurre (commented):
I'm wondering if we should also handle the errors' `cause` in the same way?

feelepxyz:
> I'm wondering if we should also handle the errors' `cause` in the same way?

Ah yes, how would you go about doing this? Should we create some kind of proxy object and sanitize the message on method call?

jurre:
> Ah yes, how would you go about doing this? Should we create some kind of proxy object and sanitize the message on method call?

Actually, since we sanitize in the initializer, I think this will already be handled 👍

common/lib/dependabot/errors.rb
```diff
@@ -12,13 +16,30 @@ def initialize(msg = nil)
     private
 
     def sanitize_message(message)
-      return unless message
+      return unless message && message.is_a?(String)
```
jurre, line 19: What could the message be, if not a string?

feelepxyz: We use some custom errors that raise with an array for some error context but rolled back the use of DependabotError in these cases but wondering if it's a good idea to keep the check in place in case we re-use in future. Seems sad to bomb out because the error serialiser raised an exception.

jurre (approved):

PR #2830, opened 2020-12-08
go_modules: don't raise error for go mod tidy

fatih:
`go mod tidy` makes sure that go.mod matches the source code in a Go
module. This also means that `dependabot` is doing a quassi linting on
the code, which prevents updating versions in a go.mod if the source
code doesn't got build. One particular use case is when a project is
missing generated files. Sometimes people don't submit and commit the
generated files of a code generator (i.e: `protobuf`) and do it in
pre-script in the CI. Because `dependabot` is not aware of this
pre-script, it'll run `go mod tidy` and fail for the user.

We should silently continue if `go mod tidy` fails and not raise an
error. This will allow dependabot still to run `go mod tidy`, but will
not block users for these kind of edge cases.

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
@@ -118,9 +118,12 @@ module Dependabot
         def run_go_mod_tidy
           return unless tidy?
 
+          # NOTE(arslan): use `go mod tidy -e` once Go 1.16 is out:
+          # https://github.com/golang/go/commit/3aa09489ab3aa13a3ac78b1ff012b148ffffe367
           command = "go mod tidy"
+
           _, stderr, status = Open3.capture3(ENVIRONMENT, command)
-          handle_subprocess_error(stderr) unless status.success?
+          handle_subprocess_error(stderr, true) unless status.success?
         end
 
         def run_go_vendor
@@ -243,7 +246,7 @@ module Dependabot
           write_go_mod(body)
         end
 
-        def handle_subprocess_error(stderr)
+        def handle_subprocess_error(stderr, istidy=false)
           stderr = stderr.gsub(Dir.getwd, "")
 
           error_regex = RESOLVABILITY_ERROR_REGEXES.find { |r| stderr =~ r }
@@ -258,6 +261,12 @@ module Dependabot
             raise Dependabot::GoModulePathMismatch.
               new(go_mod_path, match[1], match[2])
           end
+
+          # we explicitly don't raise an error for 'go mod tidy' and silently
+          # continue here. `go mod tidy` shouldn't block updating versions
+          # because there are some edge cases where it's ok to fail (such as
+          # generated files not available yet to us). 
+          return if istidy
 
           # We don't know what happened so we raise a generic error
           msg = stderr.lines.last(10).join.strip
--- a/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
@@ -122,8 +122,11 @@ module Dependabot
           # https://github.com/golang/go/commit/3aa09489ab3aa13a3ac78b1ff012b148ffffe367
           command = "go mod tidy"
 
-          _, stderr, status = Open3.capture3(ENVIRONMENT, command)
-          handle_subprocess_error(stderr, true) unless status.success?
+          # we explicitly don't raise an error for 'go mod tidy' and silently
+          # continue here. `go mod tidy` shouldn't block updating versions
+          # because there are some edge cases where it's OK to fail (such as
+          # generated files not available yet to us). 
+          Open3.capture3(ENVIRONMENT, command)
         end
 
         def run_go_vendor
@@ -246,7 +249,7 @@ module Dependabot
           write_go_mod(body)
         end
 
-        def handle_subprocess_error(stderr, istidy=false)
+        def handle_subprocess_error(stderr)
           stderr = stderr.gsub(Dir.getwd, "")
 
           error_regex = RESOLVABILITY_ERROR_REGEXES.find { |r| stderr =~ r }
@@ -261,12 +264,6 @@ module Dependabot
             raise Dependabot::GoModulePathMismatch.
               new(go_mod_path, match[1], match[2])
           end
-
-          # we explicitly don't raise an error for 'go mod tidy' and silently
-          # continue here. `go mod tidy` shouldn't block updating versions
-          # because there are some edge cases where it's ok to fail (such as
-          # generated files not available yet to us). 
-          return if istidy
 
           # We don't know what happened so we raise a generic error
           msg = stderr.lines.last(10).join.strip
--- a/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
@@ -125,7 +125,7 @@ module Dependabot
           # we explicitly don't raise an error for 'go mod tidy' and silently
           # continue here. `go mod tidy` shouldn't block updating versions
           # because there are some edge cases where it's OK to fail (such as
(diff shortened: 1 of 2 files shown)


go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -259,6 +262,12 @@ def handle_subprocess_error(stderr)
               new(go_mod_path, match[1], match[2])
           end
 
+          # we explicitly don't raise an error for 'go mod tidy' and silently
+          # continue here. `go mod tidy` shouldn't block updating versions
+          # because there are some edge cases where it's ok to fail (such as
+          # generated files not available yet to us). 
+          return if istidy
```
fatih, line 269: We still fail with `Dependabot::DependencyFileNotResolvable` and `Dependabot::GoModulePathMismatch` when running a `go mod tidy`. Initially I wanted to move up this line so it never fails, but it looks we have some test cases that cover these two specific error types. I know this you PR fixes one particular case, but wonder if we should also not take care of `Dependabot::DependencyFileNotResolvable` and `Dependabot::GoModulePathMismatch` ?

jasonrudolph: We run `go get` ([here](https://github.com/dependabot/dependabot-core/blob/499bc1adfaa583971c17d8f87edef0bc8719a9bf/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb#L84)) before we run `go mod tidy`. With that in mind, if there's something in the manifest that would trigger a `Dependabot::DependencyFileNotResolvable` error or a `Dependabot::GoModulePathMismatch` error, can we reasonably assume that those errors would have occurred during `go get`? If so, then we know that we won't even get to the `go mod tidy` step in that case. What do you think?

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -118,9 +118,12 @@ def update_files # rubocop:disable Metrics/AbcSize
         def run_go_mod_tidy
           return unless tidy?
 
+          # NOTE(arslan): use `go mod tidy -e` once Go 1.16 is out:
+          # https://github.com/golang/go/commit/3aa09489ab3aa13a3ac78b1ff012b148ffffe367
           command = "go mod tidy"
+
           _, stderr, status = Open3.capture3(ENVIRONMENT, command)
-          handle_subprocess_error(stderr) unless status.success?
+          handle_subprocess_error(stderr, true) unless status.success?
```
jasonrudolph, line 126: It seems to me like we're trying to silently ignore any problems that occur during `go mod tidy`. If that's accurate, then instead of adding a new arg to `handle_subprocess_error`, would it make sense to skip the invocation `handle_subprocess_error` entirely?

```suggestion
          # Attempt command, but ignore any errors
          Open3.capture3(ENVIRONMENT, command)
```

fatih: This makes sense and I think we should use this idiom. Initially I changed this line and removed handling the error, but I wasn't sure about some of the edge cases and whether we should return the specific error types described above. I'll update the tests as well to accommodate the changes 👍🏼

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
fatih: That's a good point Jason.  `go get` would definitely catch some of these (especially the `Dependabot::DependencyFileNotResolvable` errors).

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -118,9 +118,15 @@ def update_files # rubocop:disable Metrics/AbcSize
         def run_go_mod_tidy
           return unless tidy?
 
+          # NOTE(arslan): use `go mod tidy -e` once Go 1.16 is out:
+          # https://github.com/golang/go/commit/3aa09489ab3aa13a3ac78b1ff012b148ffffe367
```
feelepxyz, line 122: 💯

go_modules/spec/dependabot/go_modules/file_updater/go_mod_updater_spec.rb
```diff
@@ -225,10 +225,10 @@
...
             let(:dependency_version) { "v0.5.1" }
 
-            it "raises a DependencyFileNotResolvable error" do
+            it "does not raises a DependencyFileNotResolvable error" do
               error_class = Dependabot::DependencyFileNotResolvable
               expect { updater.updated_go_sum_content }.
-                to raise_error(error_class) do |error|
+                to_not raise_error(error_class) do |error|
                 expect(error.message).to include("googleapis/gnostic/OpenAPIv2")
               end
```
feelepxyz, line 233: Could change this spec to expect the update now that it is successful afaict:

```suggestion
              it "updates the go.mod" do
                expect(updater.updated_go_mod_content).to include(
                  %(github.com/googleapis/gnostic v0.5.1\n)
                )
              end
```

feelepxyz:
@fatih left a spec comment, otherwise LGTM! 🙌  The unrelated spec failures should be fixed on main now

go_modules/spec/dependabot/go_modules/file_updater/go_mod_updater_spec.rb
fatih: Thank you, this makes more sense Philip 👍🏼

go_modules/spec/dependabot/go_modules/file_updater/go_mod_updater_spec.rb
```diff
@@ -225,13 +225,19 @@
             # OpenAPIV2 has been renamed to openapiv2 in this version
             let(:dependency_version) { "v0.5.1" }
 
-            it "raises a DependencyFileNotResolvable error" do
+            it "does not raises a DependencyFileNotResolvable error" do
               error_class = Dependabot::DependencyFileNotResolvable
               expect { updater.updated_go_sum_content }.
-                to raise_error(error_class) do |error|
+                to_not raise_error(error_class) do |error|
```
jasonrudolph, line 231: 👋 @fatih: Can you help me understand why this behavior is changing in this pull request? Since we're only changing code related to `go mod tidy`, it's not yet clear to me why this test needs to change.

fatih: This particular change was caused previously by `go mod tidy`. Because we don't return an error anymore, this test fails as it expects an error. So I either had to remove it, or change it so it doesn't return an error.

jasonrudolph: Thanks, @fatih. Since https://github.com/dependabot/dependabot-core/pull/2627 introduced this test, and since that PR doesn't mention anything about `go mod tidy`, I'd like to ask @jurre to take a look before we ship this PR.

@jurre: Given your context on https://github.com/dependabot/dependabot-core/pull/2627, should we be changing this test here, or would it make more sense to delete it entirely?

feelepxyz: I believe this case is only caught by `go mod tidy` where the import path doesn't match the exported package name, `import` here: https://github.com/dependabot/dependabot-core/blob/7ce0ebf795d757c62893e60a47a53f07e2efaee0/go_modules%2Fspec%2Ffixtures%2Fprojects%2Frenamed_package%2Fmain.go#L4 and `package` export: https://github.com/google/gnostic/blob/master/openapiv2/OpenAPIv2.go#L17

I think we can delete this test case and rely on the below `updates the go.mod` as this would fail if we raised a resolvability error.

jurre: Yep, @feelepxyz is right 👍👍

feelepxyz (approved):
LGTM 👍  pushed a tiny change to the test, seemed like a good idea to keep it around for future.

PR #2838, opened 2020-12-09
Add Composer v2 support

feelepxyz:
Install composer v2 alongside v1 and default to v2 if there is no lockfile or the lockfile doesn't specify a api-plugin-version (matches the required composer version).

All spec fixtures have been converted to project fixtures.

## TODO

- [x] Fix update checker specs
- [x] Update fixtures to v2 composer
- [x] Add v1 specs for any regressions moving to v2
- [x] Check we have smoke tests covering v1 behaviour
  - [x] File parser
  - [x] Update checker
  - [x] File updater
- [x] Convert spec fixtures to project fixtures
- [x] Delete old fixtures (lockfiles and composer_files)
- [x] Investigate speeding up the update checker and file updater (disable downloading packages and just generate the lockfile)

Closes https://github.com/dependabot/dependabot-core/issues/2668


composer/spec/dependabot/composer/file_updater/lockfile_updater_spec.rb
```diff
@@ -273,23 +289,54 @@
...
-          name: "components/path_dep/composer.json",
-          content: fixture("composer_files", "path_dep")
-        )
+    context "that requires an environment variable (composer v2)" do
+      let(:project_name) { "env_variable" }
+
+      context "that hasn't been provided" do
+        it "raises a DependencyFileNotResolvable error" do
+          expect { updated_lockfile_content }.to raise_error do |error|
+            expect(error).to be_a(Dependabot::DependencyFileNotResolvable)
```
feelepxyz, line 304: TODO: figure out if we can improve this error message for v2, this used to raise (see v1 example)

```
expect(error).to be_a(Dependabot::MissingEnvironmentVariable)
expect(error.environment_variable).to eq("ACF_PRO_KEY")
```

feelepxyz: Fixed was caused by a plugin that only worked for composer v1

composer/helpers/v2/src/DependabotInstallationManager.php
```diff
@@ -0,0 +1,67 @@
...
+{
+    private array $installed = [];
+    private array $updated = [];
+    private array $uninstalled = [];
+
+    public function execute(RepositoryInterface $repo, array $operations, $devMode = true, $runScripts = true): void
+    {
+        foreach ($operations as $operation) {
+            $method = $operation->getOperationType();
+            // skipping download() step here for tests
```
feelepxyz, line 24:
```suggestion
            // skipping download() step here
```

jurre:
I'm going to prepare a small guide on how to review this, with notable diffs etc. Because as is this PR is very tough to dig into.

feelepxyz:
> I'm going to prepare a small guide on how to review this, with notable diffs etc. Because as is this PR is very tough to dig into.

Great shout! Filtering only ruby files could be useful: https://github.com/dependabot/dependabot-core/pull/2838/files?file-filters%5B%5D=.rb

jurre:
## What does this PR do?

It adds support for [composer v2](https://blog.packagist.com/composer-2-0-is-now-available/), while keeping support for composer v1.

We run these two versions side by side, and we look at the [plugin-api-version](https://getcomposer.org/doc/articles/plugins.md) to determine which version to use. The two versions are mostly compatible, and those plugins are the main reason for incompatibility between the two versions.

Not all projects use Composer plugins, and in that case, we default to Composer v2. Any updates we make in that case should be compatible with v1.

## How do we do it?

For this ecosystem, Dependabot relies fully on using Composer as an API, in native helpers, written in PHP. We never run the `composer` binary for updates.

This means that we install two versions of the composer api, which has changed between these two versions, so we also have two versions of the native helpers: [composer/helpers/v1](https://github.com/dependabot/dependabot-core/tree/composer-v2/composer/helpers/v1) and [composer/helpers/v2](https://github.com/dependabot/dependabot-core/tree/composer-v2/composer/helpers/v2). We'll get into some details on what changed in those helpers later on.

To be safe, we also [install two versions of the composer binary](https://github.com/dependabot/dependabot-core/blob/composer-v2/Dockerfile#L103-L104), which as mentioned before is not used for running the updates, but to install the dependencies of those two native helper scripts. We install the dependencies of each helper with their [respective composer version](https://github.com/dependabot/dependabot-core/blob/composer-v2/composer/helpers/v1/build).

There are two places in the ruby code where we decide which native helpers to use, the [LockfileUpdater](https://github.com/dependabot/dependabot-core/blob/composer-v2/composer/lib/dependabot/composer/file_updater/lockfile_updater.rb#L447) and the [VersionResolver](https://github.com/dependabot/dependabot-core/blob/composer-v2/composer/lib/dependabot/composer/update_checker/version_resolver.rb), we pass in the composer version (based on the `plugin-api-version` that we grab from the lockfile).

The rest of the changes we can make fully in Ruby, and we don't have to shell out for those, and since the format of both the `composer.json` and `composer.lock` has not changed, none of that code is touched in this PR, except for a few places where the error messages that we parse from Composer have been changed. In those cases we make sure to support both versions of the error.

Then finally, we've had to make some changed to the native PHP helpers, mostly to [satisfy the new API](https://github.com/dependabot/dependabot-core/blob/composer-v2/composer/helpers/v2/src/UpdateChecker.php#L54-L70), where some things are changed based on how the classes are set up and the options we need to pass in. 

And besides that, we've [updated our internal InstallationManager](https://github.com/dependabot/dependabot-core/pull/2838/commits/e2a32660a189c3a444f32031f1024d34733bf739) class, to no longer download any files. This object is used in the `Installer` (which runs most of the composer logic) to handle downloading and installing packages. We override it when checking which packages have changed, and count packages that are updated/deleted/changed. We've been able to gut out most of the actual work this does, and only keep track of which packages need to be updated.

## Testing

The Composer tests used to have two folders with `composer.json` and `composer.lock` files, and would mix and match these to account for certain scenario's. Those have all been moved, and we now have a dedicated folder for each scenario, which holds all of the relevant files for that test. So previously we'd have `fixtures/composer_files/git_source` and `fixtures/lockfiles/git_source`, where now we have `fixtures/projects/git_source`.

jurre:
Here is a diff between the native helpers:

```diff
diff --git a/composer/helpers/v1/src/DependabotInstallationManager.php b/composer/helpers/v2/src/DependabotInstallationManager.php
index a25fdc36d..cbc9de593 100644
--- a/composer/helpers/v1/src/DependabotInstallationManager.php
+++ b/composer/helpers/v2/src/DependabotInstallationManager.php
@@ -17,21 +17,27 @@ final class DependabotInstallationManager extends InstallationManager
     private array $updated = [];
     private array $uninstalled = [];
 
+    public function execute(RepositoryInterface $repo, array $operations, $devMode = true, $runScripts = true): void
+    {
+        foreach ($operations as $operation) {
+            $method = $operation->getOperationType();
+            // NOTE: skipping download() step
+            $this->$method($repo, $operation);
+        }
+    }
+
     public function install(RepositoryInterface $repo, InstallOperation $operation): void
     {
-        parent::install($repo, $operation);
         $this->installed[] = $operation->getPackage();
     }
 
     public function update(RepositoryInterface $repo, UpdateOperation $operation): void
     {
-        parent::update($repo, $operation);
         $this->updated[] = [$operation->getInitialPackage(), $operation->getTargetPackage()];
     }
 
     public function uninstall(RepositoryInterface $repo, UninstallOperation $operation): void
     {
-        parent::uninstall($repo, $operation);
         $this->uninstalled[] = $operation->getPackage();
     }
 
diff --git a/composer/helpers/v1/src/DependabotPluginManager.php b/composer/helpers/v2/src/DependabotPluginManager.php
index 67e1436a1..6b329b472 100644
--- a/composer/helpers/v1/src/DependabotPluginManager.php
+++ b/composer/helpers/v2/src/DependabotPluginManager.php
@@ -9,7 +9,7 @@ use Composer\Plugin\PluginManager;
 
 final class DependabotPluginManager extends PluginManager
 {
-    public function registerPackage(PackageInterface $package, $failOnMissingClasses = false): void
+    public function registerPackage(PackageInterface $package, $failOnMissingClasses = false, $isGlobalPlugin = false): void
     {
         // This package does some setup for PHP_CodeSniffer, but errors out the
         // install if Symfony isn't installed (which it won't be for a lockfile
@@ -18,6 +18,6 @@ final class DependabotPluginManager extends PluginManager
             return;
         }
 
-        parent::registerPackage($package, $failOnMissingClasses);
+        parent::registerPackage($package, $failOnMissingClasses, $isGlobalPlugin);
     }
 }
diff --git a/composer/helpers/v1/src/UpdateChecker.php b/composer/helpers/v2/src/UpdateChecker.php
index 16a37306b..9fa66ae49 100644
--- a/composer/helpers/v1/src/UpdateChecker.php
+++ b/composer/helpers/v2/src/UpdateChecker.php
@@ -4,9 +4,11 @@ declare(strict_types=1);
 
 namespace Dependabot\Composer;
 
+use Composer\DependencyResolver\Request;
 use Composer\Factory;
 use Composer\Installer;
 use Composer\Package\PackageInterface;
+use Composer\Util\Filesystem;
 
 final class UpdateChecker
 {
@@ -46,7 +48,14 @@ final class UpdateChecker
             $io->loadConfiguration($config);
         }
 
-        $installationManager = new DependabotInstallationManager();
+        $installationManager = new DependabotInstallationManager($composer->getLoop(), $io);
+
+        $fs = new Filesystem(null);
+        $binaryInstaller = new Installer\BinaryInstaller($io, rtrim($composer->getConfig()->get('bin-dir'), '/'), $composer->getConfig()->get('bin-compat'), $fs);
+
+        $installationManager->addInstaller(new Installer\LibraryInstaller($io, $composer, null, $fs, $binaryInstaller));
+        $installationManager->addInstaller(new Installer\PluginInstaller($io, $composer, $fs, $binaryInstaller));
+        $installationManager->addInstaller(new Installer\MetapackageInstaller($io));
 
         $install = new Installer(
             $io,
@@ -62,16 +71,19 @@ final class UpdateChecker
 
         // For all potential options, see UpdateCommand in composer
         $install
-            ->setDryRun(true)
             ->setUpdate(true)
             ->setDevMode(true)
-            ->setUpdateAllowList([$dependencyName])
-            ->setAllowListTransitiveDependencies(true)
-            ->setExecuteOperations(false)
+            ->setUpdateAllowTransitiveDependencies(Request::UPDATE_LISTED_WITH_TRANSITIVE_DEPS)
             ->setDumpAutoloader(false)
             ->setRunScripts(false)
             ->setIgnorePlatformRequirements(false);
 
+        // if no lock is present, we do not do a partial update as
+        // this is not supported by the Installer
+        if ($composer->getLocker()->isLocked()) {
+            $install->setUpdateAllowList([$dependencyName]);
+        }
+
         $install->run();
 
         $installedPackages = $installationManager->getInstalledPackages();
diff --git a/composer/helpers/v1/src/Updater.php b/composer/helpers/v2/src/Updater.php
index abb08d8b6..2ee2b5af5 100644
--- a/composer/helpers/v1/src/Updater.php
+++ b/composer/helpers/v2/src/Updater.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Dependabot\Composer;
 
+use Composer\DependencyResolver\Request;
 use Composer\Factory;
 use Composer\Installer;
 
@@ -79,8 +80,8 @@ final class Updater
             ->setUpdate(true)
             ->setDevMode(true)
             ->setUpdateAllowList([$dependencyName])
-            ->setAllowListTransitiveDependencies(true)
-            ->setExecuteOperations(false)
+            ->setUpdateAllowTransitiveDependencies(Request::UPDATE_LISTED_WITH_TRANSITIVE_DEPS)
+            ->setExecuteOperations(true)
             ->setDumpAutoloader(false)
             ->setRunScripts(false)
             ->setIgnorePlatformRequirements(false);

```

jurre:
@brrygrdn would you mind taking a look? We're pretty confident about these changes, but would be good to get some fresh eyes on it, at the very least we'll share a bit of context that way. Also happy to pair-review this

brrygrdn:
We have a couple of meetings come up, so just for clarity, I'm about 50% through code review. My PHP is rusty, but it's mostly just a few questions/nits so far. Will finish up after planning 🙇🏻

composer/spec/dependabot/composer/file_updater_spec.rb
```diff
@@ -113,22 +102,38 @@
...
+      let(:parsed_updated_lockfile_content) { JSON.parse(updated_lockfile_content) }
+      let(:updated_lockfile_entry) do
+        parsed_updated_lockfile_content["packages"].find do |package|
+          package["name"] == dependency.name
+        end
+      end
+
+      it "updates the dependency version and plugin-api-version (to match instaled composer) in the lockfile" do
+        expect(updated_lockfile_entry["version"]).to eq("1.22.1")
+        expect(parsed_updated_lockfile_content["prefer-stable"]).to be(false)
```
brrygrdn, line 117: ❓ Is this just a feature of v2 files or are we explicitly testing we are going from 'version unspecified to v2'?

composer/spec/dependabot/composer/update_checker/latest_version_finder_spec.rb
```diff
@@ -334,8 +313,7 @@
     end
 
     context "with an unreachable source (speccing we don't try to reach it)" do
-      let(:manifest_fixture_name) { "git_source_unreachable_git_url" }
-      let(:lockfile_fixture_name) { "git_source_unreachable_git_url" }
+      let(:project_name) { "git_source_unreachable_git_url" }
```
brrygrdn, line 316: Enjoying these test tidy ups, thanks 🎉

brrygrdn (approved):
This looks solid to me, nothing blocking just a few questions.

The test improvements/cleanups are 🥇

composer/helpers/v2/src/Updater.php
```diff
@@ -0,0 +1,99 @@
...
+    {
+        [$workingDirectory, $dependencyName, $dependencyVersion, $gitCredentials, $registryCredentials] = $args;
+
+        // Change working directory to the one provided, this ensures that we
+        // install dependencies into the working dir, rather than a vendor folder
+        // in the root of the project
+        $originalDir = getcwd();
+
+        if (!is_string($originalDir)) {
+            throw new \RuntimeException('Failed determining the current working directory.');
```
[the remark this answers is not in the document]
jurre: Yeah tbh this a 1-1 copy from the v1 implementation. I'm open to improving it, maybe in a separate PR?

composer/lib/dependabot/composer/file_updater/lockfile_updater.rb
```diff
@@ -425,7 +444,17 @@ def update_required_extensions(additional_extensions)
...
+          NativeHelpers.composer_helper_path(composer_version: composer_version)
+        end
+
+        def composer_version
+          @composer_version ||=
+            begin
+              return "v2" unless parsed_lockfile["plugin-api-version"]
+
+              version = Version.new(parsed_lockfile["plugin-api-version"])
+              version.canonical_segments.first == 1 ? "v1" : "v2"
```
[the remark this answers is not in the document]
jurre: Yeah good shout, we just don't have a great way of emitting warnings that I know of?

composer/spec/dependabot/composer/file_updater_spec.rb
feelepxyz: Don't actually know what `prefer-stable` is 🤔 copy pasted from the previous spec but added the spec to check the `plugin-api-version` matches the installed composer version (updates from 2.1.0 back to 2.0.0)

composer/lib/dependabot/composer/file_updater/lockfile_updater.rb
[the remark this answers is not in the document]
feelepxyz: Yeah lets resurrect the issue to add instrumentation to core, seems ok for now as a new major version is likely to break a bunch of specs.

PR #2848, opened 2020-12-11
go_modules: only stub modules outside the current checkout

thepwagner:
Some users have multiple go modules within a single repository, and use `replace` directives to reference local modules on disk. [sample](https://github.com/thepwagner/dependagot/blob/master/go/modules/go.mod#L5).

Previous to this PR, Dependabot would replace all local `replace` directives with stubbed modules, under the assumption that they weren't available.

This assumption changed when Dependabot started cloning full Go repositories (a requirement for `go mod vendor` and `go mod tidy` support)! This PR adjusts the logic to only stub modules that are:
* absolute paths
* relative paths that escape the current checkout (e.g. `../foo` from the root)
* relative paths that do not exist

cc https://github.com/dependabot/dependabot-core/pull/2551#discussion_r494289784

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
@@ -235,20 +235,31 @@ module Dependabot
               (manifest["Replace"] || []).
                 map { |r| r["New"]["Path"] }.
                 compact.
-                select { |p| external_path?(p) }.
+                select { |p| stub_replace_path?(p) }.
                 map { |p| [p, "./" + Digest::SHA2.hexdigest(p)] }.
                 to_h
             end
         end
 
-        # returns true if the provided path does not exist in the current checkout
-        def external_path?(path)
-          return true if path.start_with?("/")
-          return false unless path.start_with?(".")
+        # returns true if the provided path should be replaced with a stub
+        def stub_replace_path?(path)
+          return true if absolute_path?(path)
+          return false unless relative_replacement_path?(path)
 
-          !module_pathname.join(path).realpath.to_s.start_with?(repo_contents_path.to_s)
+          resolved_path = module_pathname.join(path).realpath
+          inside_repo_contents_path = resolved_path.to_s.start_with?(repo_contents_path.to_s)
+          !inside_repo_contents_path
         rescue Errno::ENOENT
           true
+        end
+
+        def absolute_path?(path)
+          path.start_with?("/")
+        end
+
+        def relative_replacement_path?(path)
+          # https://golang.org/ref/mod#go-mod-file-replace
+          path.start_with?("./") || path.start_with?("../")
         end
 
         def module_pathname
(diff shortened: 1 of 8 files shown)


go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -235,12 +235,26 @@ def replace_directive_substitutions(manifest)
...
+                select { |p| external_path?(p) }.
                 map { |p| [p, "./" + Digest::SHA2.hexdigest(p)] }.
                 to_h
             end
         end
 
+        # returns true if the provided path does not exist in the current checkout
+        def external_path?(path)
+          return true if path.start_with?("/")
+          return false unless path.start_with?(".")
```
fatih, line 247: I'm trying to understand these lines. How does these lines allow to understand whether the path is not inside the repository? I think a `/foo/bar/repo_path/qux` still could point into a folder that is inside the repository or `./foo` could point to a folder inside a repository. I wonder if `external_path` is meaning something else.

thepwagner: > I think a /foo/bar/repo_path/qux still could point into a folder that is inside the repository

This is unlikely. Since Dependabot clones the repository to a temporary directory, an absolute path inside the repository would have to be `/tmp/dependabot-${random}`, where `${random}` varies between runs. It _could_ happen, but is so unlikely we can save an `fstat()` by not even checking.

> ./foo could point to a folder inside a repository.

This should be catching that case. Per https://golang.org/ref/mod#go-mod-file-replace , we know only `./foo` and `../foo` are valid so the `path.start_with?(".")` check attempts to check that.
If a replacement path _isn't_ a relative path, then Dependabot shouldn't replace it - the `go.mod` file is invalid so the most useful thing to do is return the error from the `go` SDK:
```
go: errors parsing go.mod:
/Users/thepwagner/tmp/go_mod_test/go.mod:7: replacement module without version must be directory path (rooted or starting with ./ or ../)
```


> I wonder if external_path is meaning something else.

Good shout - I've renamed to more closely what this is doing: deciding if this path should be stubbed / `#stub_replace_path?`

feelepxyz (approved):
Nice improvement! Test case looks good to me 👍

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -240,12 +240,37 @@ def replace_directive_substitutions(manifest)
...
+          path.start_with?("/")
+        end
+
+        def relative_replacement_path?(path)
+          # https://golang.org/ref/mod#go-mod-file-replace
+          path.start_with?("./") || path.start_with?("../")
+        end
+
+        def module_pathname
+          @module_pathname ||= repo_contents_path.join(directory)
```
thepwagner, line 271: This was a bug: `repo_contents_path` is not a `Pathname`, fixed by https://github.com/dependabot/dependabot-core/pull/2863

go_modules/spec/dependabot/go_modules/file_updater/go_mod_updater_spec.rb
```diff
@@ -478,6 +479,29 @@
         end
       end
     end
+
+    context "for a monorepo" do
+      let(:project_name) { "monorepo" }
+      let(:directory) { "cmd" }
```
thepwagner, line 485: This was a bug: the directory is typically specified with a leading slash https://github.com/dependabot/dependabot-core/pull/2863

PR #2849, opened 2020-12-12
Add parsing of MSBuild SDK dependencies (NuGet)

### LINKED ISSUE #2839: Support MSBuild SDK packages specified in csproj
Currently, Dependabot only looks for `<PackageReference>`, `packages.config` and similar items used to consume packages to be referenced by a project.

However, it is also possible to consume a NuGet package containing an MSBuild SDK, and this does not use such references.
Instead, there are two ways of consuming `Awesome.Sdk 1.2.3`:
1. In the project file, use:

   ```xml
   <Project Sdk="Awesome.Sdk/1.2.3">
   ```

   the MSBuild SDK resolver will then look for a NuGet package `Awesome.Sdk` at version 1.2.3.
2. In the project file, use:

   ```xml
   <Project Sdk="Awesome.Sdk">
   ```

   and in `global.json`, have:

   ```json
   {
     "msbuild-sdks": {
       "Awesome.Sdk" : "1.2.3",
       "Another.Pretty.Good.Sdk": "5.0.0"
     }
   }
   ```

   (making it easier to synchronize the SDK versions across multiple projects).

Either way, the `Sdk` project attribute can contain multiple SDK names, separated by `;`, and while the version can be a pre-release (so `Awesome.Sdk/1.2.4-beta.do.not.use` is valid), it has to be a specific version (so no range syntax or `*`).

Zastai:
This adds the three ways to specify an MSBuild SDK reference inside a project file, including test cases for each (parsing and updating).

Intended to fix #2839 

Note: I'm new to Ruby, but do have access to the JetBrainz IDEs, which has helped. So please forgive novice code issues.


Zastai:
Hmm - looks like I missed that while the actual tests pass, the coverage is complaining about a 0.02% drop.
Will see what I can do about that.

Zastai:
Not a lot, it seems. Well, I could remove two existing-but-unused methods (`declaration_nodes` in the two finders), but that has nothing to do with what this PR is about, so seems like the wrong way to go.

Zastai:
Would be nice to get some feedback on this...

jurre:
Thanks for doing this work @Zastai, I'll try to make some time next week to thoroughly review this and get a bit more familiar with MSBuild SDK references. Just wanted to let you know that we appreciate the contribution.

japj:
Hello, great to see this (I recently “discovered” dependabot).
We have been building our own MSBuild Sdk for a while now and we actually use a different way to specify the sdk version (that is documented in the link you provided https://docs.microsoft.com/en-us/visualstudio/msbuild/how-to-use-project-sdk)

The global.json can have an msbuild-sdks section that can be modified to “upgrade” the sdk for all csproj files that use that sdk:

```
{
    "msbuild-sdks": {
        "My.Custom.Sdk": "5.0.0",
        "My.Other.Sdk": "1.0.0-beta"
    }
}

```

so in that case, just modifying the global.json is enough.
Could you perhaps add that behavior to this PR?

japj:
Sorry, I just read #2839 where you mention that global.json support is already in?

Zastai:
That was the plan - but then I found that global.json already had that behaviour: https://github.com/dependabot/dependabot-core/blob/3bca853184b6a345fbee4a4cf2e9d57072ebe90b/nuget/lib/dependabot/nuget/file_parser/global_json_parser.rb#L23

Zastai:
@jurre Any update?

Zastai:
@jurre
It has now been 8 months since I created this PR. Would be nice to get this properly reviewed.

Zastai:
We're rapidly approaching this PR's one year anniversary...
Any chance of a review, @jurre?

nuget/lib/dependabot/nuget/file_updater/project_file_declaration_finder.rb
```diff
@@ -95,6 +111,80 @@ def declaring_file
...
+          sdk_strings(SDK_IMPORT_REGEX, "Import", "Sdk", "Version")
+        end
+
+        # rubocop:disable Metrics/PerceivedComplexity
+        def sdk_project_strings
+          dep_name = dependency_name&.downcase
+          dep_version = declaring_requirement.fetch(:requirement)
+          strings = []
+          declaring_file.content.scan(SDK_PROJECT_REGEX).each do |string|
+            xml = string
```
jurre, line 125:
```suggestion
          declaring_file.content.scan(SDK_PROJECT_REGEX).each do |xml|
```

nuget/lib/dependabot/nuget/file_updater/project_file_declaration_finder.rb
```diff
@@ -95,6 +111,80 @@ def declaring_file
...
+        end
+
+        # rubocop:disable Metrics/CyclomaticComplexity
+        # rubocop:disable Metrics/PerceivedComplexity
+        def sdk_strings(regex, element_name, name_attribute, version_attribute)
+          dep_name = dependency_name&.downcase
+          dep_version = declaring_requirement.fetch(:requirement)
+          strings = []
+          declaring_file.content.scan(regex).each do |string|
+            xml = string
```
jurre, line 163:
```suggestion
          declaring_file.content.scan(regex).each do |xml|
```

nuget/lib/dependabot/nuget/file_updater/project_file_declaration_finder.rb
```diff
@@ -95,6 +111,80 @@ def declaring_file
...
+          dep_name = dependency_name&.downcase
+          dep_version = declaring_requirement.fetch(:requirement)
+          strings = []
+          declaring_file.content.scan(SDK_PROJECT_REGEX).each do |string|
+            xml = string
+            xml += "</Project>" unless string.end_with?("/>")
+            node = Nokogiri::XML(xml)
+            node.remove_namespaces!
+            element = node.at_xpath("/Project")
+            next unless element
```
jurre, line 130: There's some duplication between this and [the block in `#sdk_strings`](https://github.com/dependabot/dependabot-core/pull/2849/files#diff-fa02b4ba56ca051e40b58c9152d7796a54f9f2f017a304fc911a7edb41aee838R164-R168), could it be worth extracting a helper method?

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
```diff
@@ -50,6 +55,63 @@ def dependency_set(project_file:)
...
+
+        def add_sdk_refs_from_import_tags(doc, dependency_set, project_file)
+          doc.xpath("/Project/Import").each do |import_node|
+            next unless import_node.attribute("Sdk") && import_node.attribute("Version")
+
+            name = import_node.attribute("Sdk")&.value&.strip
+            version = import_node.attribute("Version")&.value&.strip
+
+            dependency =
+              build_dependency(name, version, version, nil, project_file)
```
jurre, line 88:
```suggestion
            dependency = build_dependency(name, version, version, nil, project_file)
```

We relaxed some linter rules a while ago. This is not important but I find it slightly easier on the eyes, feel free to ignore though.

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
```diff
@@ -50,6 +55,63 @@ def dependency_set(project_file:)
...
+
+        def add_sdk_refs_from_sdk_tags(doc, dependency_set, project_file)
+          doc.xpath("/Project/Sdk").each do |sdk_node|
+            next unless sdk_node.attribute("Version")
+
+            name = sdk_node.attribute("Name")&.value&.strip
+            version = sdk_node.attribute("Version")&.value&.strip
+
+            dependency =
+              build_dependency(name, version, version, nil, project_file)
```
jurre, line 110:
```suggestion
            dependency = build_dependency(name, version, version, nil, project_file)
```

We relaxed some linter rules a while ago. This is not important but I find it slightly easier on the eyes, feel free to ignore though.

nuget/lib/dependabot/nuget/file_updater/project_file_declaration_finder.rb
```diff
@@ -95,6 +111,80 @@ def declaring_file
...
+              next unless parts.length == 2
+              next unless parts[0]&.downcase == dep_name
+              next unless parts[1] == dep_version
+
+              strings << sdk_reference
+            end
+          end
+          # Drop duplicates, to avoid potential issues with repeated replacements (e.g. upgrading Foo/1.0.0-beta to
+          # Foo/1.0.0-beta.2). However, that turns the empty array into nil.
+          strings.uniq! || []
```
jurre, line 148:
```suggestion
          strings.uniq
```

The `uniq!` version of the method mutates the array in-place, but since we're returning the value it's fine to return a new instance, and this so called "non bang" version of the method does not return an empty array into nil.

nuget/lib/dependabot/nuget/file_updater/project_file_declaration_finder.rb
```diff
@@ -95,6 +111,80 @@ def declaring_file
...
+        end
+        # rubocop:enable Metrics/PerceivedComplexity
+
+        def sdk_sdk_strings
+          sdk_strings(SDK_SDK_REGEX, "Sdk", "Name", "Version")
+        end
+
+        # rubocop:disable Metrics/CyclomaticComplexity
+        # rubocop:disable Metrics/PerceivedComplexity
```
jurre, line 157: I think there might be some room to extract certain parts of this into separate methods, the linter things the complexity of this method is too high and I kind of agree that it's a little hard to follow.

I do also understand that you've submitted this PR a really long time ago and you might not have an appetite to dig back in and refactor things, but it would be appreciated if we can clean this up a little and reduce the complexity of the method a little.

jurre (commented):
Overall these changes look good, I have made some mostly superficial suggestions regarding readability and rubyisms.

I am not super familiar with how the newly added functionality is typically used in NuGet so I'm having somewhat of a hard time reviewing for correctness.

I think @xlgmokha has some more experience on this so I would love to get his eyes on this as well.

Do you happen to have a sample project I could test this end-to-end on?

Apologies for the slow response on this on my part, it definitely fell through the cracks a couple of times and going to try to land this ASAP.

Zastai:
I should be able to make a recipe for setting up an otherwise-empty project with each form of reference, if that suits.
All it would need is an installed .NET SDK (or .NET Core SDK).

nuget/lib/dependabot/nuget/file_updater/project_file_declaration_finder.rb
Zastai: Applied

nuget/lib/dependabot/nuget/file_updater/project_file_declaration_finder.rb
Zastai: Applied

nuget/lib/dependabot/nuget/file_updater/project_file_declaration_finder.rb
Zastai: Will have a look

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
Zastai: Applied (to all 3 such splits)

nuget/lib/dependabot/nuget/file_updater/project_file_declaration_finder.rb
Zastai: Applied

nuget/lib/dependabot/nuget/file_updater/project_file_declaration_finder.rb
Zastai: Actually, I'm having to revert this - `string` needs to be used unchanged afterwards, while `xml` potentially gets modified.

nuget/lib/dependabot/nuget/file_updater/project_file_declaration_finder.rb
Zastai: This has been refactored, with several ignore-complexity comments removed.

nuget/lib/dependabot/nuget/file_updater/project_file_declaration_finder.rb
Zastai: Done

Zastai:
I think I addressed all your remarks, @jurre.

Sample projects are present as fixtures, but in order to set up a repo for "real" testing, here are the recipes.

## Recipes

Each recipe starts from an empty directory and sets up a dependency on an old version of an SDK package I maintain (`MetaBrainz.Build.Sdk`).
Dependabot should suggest an update for it.

### A. Using global.json (already supported by dependabot)
1. Create a `.csproj` file (name can be chosen freely) containing
   ```
   <Project Sdk="MetaBrainz.Build.Sdk">
   </Project>
   ```
2. Create a `global.json` file containing
   ```
   {
     "msbuild-sdks": {
       "MetaBrainz.Build.Sdk" : "1.0.0"
     }
   }
   ```

### B. Using a version specifier in the Project tag (1)
1. Create a `.csproj` file (name can be chosen freely) containing
   ```
   <Project Sdk="MetaBrainz.Build.Sdk/1.0.0">
   </Project>
   ```

### C. Using a version specifier in the Project tag (2)
This is the same case as B, but it checks for support for multiple SDK specifications.
1. Create a `.csproj` file (name can be chosen freely) containing
   ```
   <Project Sdk="Microsoft.NET.Sdk;MetaBrainz.Build.Sdk/1.0.0">
   </Project>
   ```

### D. Using an Sdk tag
1. Create a `.csproj` file (name can be chosen freely) containing
   ```
   <Project>

     <Sdk Name="MetaBrainz.Build.Sdk" Version="1.0.0" />

   </Project>
   ```

### E. Using an Sdk attribute on an Import tag
1. Create a `.csproj` file (name can be chosen freely) containing
   ```
   <Project>

     <Import Sdk="MetaBrainz.Build.Sdk" Project="Defaults.props" Version="1.0.0" />
     <Import Project="CSharp.props" Version="1.0.0" Sdk="MetaBrainz.Build.Sdk" />

     <Target Name="Dummy" />

   </Project>
   ```

nuget/lib/dependabot/nuget/file_updater/project_file_declaration_finder.rb
jurre: fwiw string will also get modified:

```ruby
irb(main):004:0> string = "foo"
irb(main):005:0> xml = string
irb(main):006:0> xml << "x"
=> "foox"
irb(main):007:0> string
=> "foox"
```

Zastai: It would if I used `<<`, but I use `xml +=` and it's my understanding that that creates a new string.

jurre: Ah that's correct, but looking at the code we never refer to `string` after `xml` is changed, at least not now that the code is extracted to `parse_element`. Anyway, it's not really important I guess

Zastai: We do in `sdk_strings`: `strings << string`; that's why parse_element() retains the `+=`.

Zastai:
Bit confused that tests are failing on a part of Dependabot not touched by this PR (it tries to use `pandoc` which is not installed).
Is CI broken on main/master or something?

jurre:
> Bit confused that tests are failing on a part of Dependabot not touched by this PR (it tries to use `pandoc` which is not installed). Is CI broken on main/master or something?

I don't see this failures you're referring to? I do see a few linting errors for nuget that we'll need to address, and there are a few test suites that are unfortunately sometimes flaky, if a suite other than `nuget` fails in this PR I can retry/fix it on main though.

Zastai:
That's strange; I don't see it now either - only a style check, on a source I don't think I changed. I'll push an update for that to see if that fixes the checks.

xlgmokha:
> I think @xlgmokha has some more experience on this so I would love to get his eyes on this as well.

LGTM. I'm not sure where/why pandoc is used but I'll take a peek. Otherwise...  :shipit:

xlgmokha:
It looks like a new version of [pandoc-ruby](https://rubygems.org/gems/pandoc-ruby/versions/2.1.5) was released recently that includes a change in behaviour that is triggering a failure in the test suite. This repo doesn't seem to include a version controlled `Gemfile.lock` so the resolved version changed and caused this failure. I'll try to get this fixed up in the main branch so that we can get these changes released.

xlgmokha:
I'm working on a fix in https://github.com/dependabot/dependabot-core/pull/4412 that should restore the `common` build to a healthier state. Unfortunately, that means we'll need to update this branch which is something I don't have access to do. 
@Zastai Do you mind pulling in the latest when https://github.com/dependabot/dependabot-core/pull/4412 is merged? I'll take care of the rest.

Zastai:
Yeah, no worries - ping me here when it lands and I'll pull in main.

Zastai:
@xlgmokha Main merged; hopefully that will allow all the checks to pass.

xlgmokha (approved):

xlgmokha:
Thank you for your contribution @Zastai.

nuget/lib/dependabot/nuget/file_updater/project_file_declaration_finder.rb
```diff
@@ -95,6 +111,71 @@ def declaring_file
...
           raise "No file found with name #{filename}!"
         end
+
+        def sdk_import_strings
+          sdk_strings(SDK_IMPORT_REGEX, "Import", "Sdk", "Version")
+        end
+
+        def parse_element(string, name)
+          xml = string
+          xml += "</#{name}>" unless string.end_with?("/>")
```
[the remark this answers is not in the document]
Zastai: I do not know. It has been a long time since I created the PR, but I _think_ I copied what was already there for other processing (like `PackageReference`).

PR #2872, opened 2020-12-15
go_modules: fix test warning 🩹 

thepwagner:
Don't `expect().not_to raise_error(SpecificError)`, since that masks all other potential errors.


fatih (approved):
Looks good to me 👍🏼

jurre (approved):
Thanks! We should set up a linter for this

PR #2873, opened 2020-12-15
go_modules: GitDependenciesNotReachable.urls are repos, not module paths

thepwagner:
In https://github.com/dependabot/dependabot-core/pull/2791 we added a test for unreachable dependencies that aren't located at the repository root. 

At the time it was assumed that since the user wrote `github.com/thepwagner/my-awesome-repo/v2` that was the best "dependency URL" to return with errors. Since introducing [updates from private GitHub repositories](https://github.blog/changelog/2020-12-02-dependabot-version-updates-from-private-github-repositories/) it has become clear that recording the problematic repository is a better solution than the exact path.

This PR adjusts the behaviour so only the base repository URL (e.g. `github.com/${owner}/${repo}`) is raised in the error for unreachable git dependencies.


jurre (approved):
Nice one 👍

PR #2880, opened 2020-12-16
gomod: Do not attempt to parse transitive dependencies

jurre:
Previously we would run `go list -m -json all` to generate a list of
dependencies to attempt to update.

The way that this is implemented, it causes _every_ transitive
dependency to be listed, even if they are not required by this project.

This can happen when a module defines several packages, but only a
subset of those packages is used by the current repository.

It will also verify that each of those transitive dependencies is
available, and this causes issues when we rely on a dependency with
private transitive dependencies. We run into the scenario where we have
access to any of the transitive dependencies that are used, but since we
do not have access to an _unused_ transitive dependency, we run into a
`dependency_file_not_resolvable` error.

This is especially unfortunate, since we won't attempt to update _any_
indirect dependencies for go_modules anyway.

By using `go mod edit -json` instead, we get an easier to parse list of
dependencies, and we grab just the direct ones.

In order to replicate the previous behavior of raising a
`DependencyFileNotResolvable` error when some of the packages are not
available, we now also run a `go get`, but it's debatable if we even
need to do this, since this error will be raised in subsequent steps (in
the `FileUpdater`), and not raising it here means that we can
consolidate our error handling for this scenario in one place.

Co-authored-by: Fatih Arslan <fatih@github.com>

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_parser.rb
+++ b/go_modules/lib/dependabot/go_modules/file_parser.rb
@@ -16,17 +16,8 @@ module Dependabot
       def parse
         dependency_set = Dependabot::FileParsers::Base::DependencySet.new
 
-        i = 0
-        chunks = module_info.lines.
-                 group_by { |line| line == "{\n" ? i += 1 : i }
-        deps = chunks.values.map { |chunk| JSON.parse(chunk.join) }
-
-        deps.each do |dep|
-          # The project itself appears in this list as "Main"
-          next if dep["Main"]
-
-          dependency = dependency_from_details(dep)
-          dependency_set << dependency if dependency
+        required_packages.each do |dep|
+          dependency_set << dependency_from_details(dep)
         end
 
         dependency_set.dependencies
@@ -65,8 +56,8 @@ module Dependabot
         )
       end
 
-      def module_info
-        @module_info ||=
+      def required_packages
+        @required_packages ||=
           SharedHelpers.in_a_temporary_directory do |path|
             SharedHelpers.with_git_configured(credentials: credentials) do
               # Create a fake empty module for each local module so that
@@ -78,9 +69,10 @@ module Dependabot
               end
 
               File.write("go.mod", go_mod_content)
+              File.write("main.go", "package dummypkg\n func main() {}\n")
 
-              command = "go mod edit -print > /dev/null"
-              command += " && go list -m -json all"
+              command = "go get > /dev/null" # Verify the packages are installable
+              command += " && go mod edit -json"
 
               # Turn off the module proxy for now, as it's causing issues with
               # private git dependencies
@@ -88,7 +80,7 @@ module Dependabot
 
               stdout, stderr, status = Open3.capture3(env, command)
               handle_parser_error(path, stderr) unless status.success?
-              stdout
+              JSON.parse(stdout)["Require"]
             rescue Dependabot::DependencyFileNotResolvable
               # We sometimes see this error if a host times out.
               # In such cases, retrying (a maximum of 3 times) may fix it.
--- a/go_modules/lib/dependabot/go_modules/file_parser.rb
+++ b/go_modules/lib/dependabot/go_modules/file_parser.rb
@@ -17,7 +17,7 @@ module Dependabot
         dependency_set = Dependabot::FileParsers::Base::DependencySet.new
 
         required_packages.each do |dep|
-          dependency_set << dependency_from_details(dep)
+          dependency_set << dependency_from_details(dep) unless dep["Indirect"]
         end
 
         dependency_set.dependencies
--- a/go_modules/lib/dependabot/go_modules/file_parser.rb
+++ b/go_modules/lib/dependabot/go_modules/file_parser.rb
@@ -71,8 +71,7 @@ module Dependabot
               File.write("go.mod", go_mod_content)
               File.write("main.go", "package dummypkg\n func main() {}\n")
 
-              command = "go get > /dev/null" # Verify the packages are installable
-              command += " && go mod edit -json"
+              command = "go mod edit -json"
 
               # Turn off the module proxy for now, as it's causing issues with
               # private git dependencies
@@ -127,52 +126,9 @@ module Dependabot
         end
       end
 
-      GIT_ERROR_REGEX = /go: .*: git fetch .*: exit status 128/m.freeze
       def handle_parser_error(path, stderr)
-        case stderr
-        when /go: .*: unknown revision/m
-          line = stderr.lines.grep(/unknown revision/).first.strip
-          handle_github_unknown_revision(line) if line.start_with?("go: github.com/")
-          raise Dependabot::DependencyFileNotResolvable, line
-        when /go: .*: unrecognized import path/m
-          line = stderr.lines.grep(/unrecognized import/).first
-          raise Dependabot::DependencyFileNotResolvable, line.strip
-        when /go: errors parsing go.mod/m
-          msg = stderr.gsub(path.to_s, "").strip
-          raise Dependabot::DependencyFileNotParseable.new(go_mod.path, msg)
-        when GIT_ERROR_REGEX
-          lines = stderr.lines.drop_while { |l| GIT_ERROR_REGEX !~ l }
-          raise Dependabot::DependencyFileNotResolvable.new, lines.join
-        else
-          msg = stderr.gsub(path.to_s, "").strip
-          raise Dependabot::DependencyFileNotParseable.new(go_mod.path, msg)
-        end
-      end
-
-      GITHUB_REPO_REGEX = %r{github.com/[^@]*}.freeze
-      def handle_github_unknown_revision(line)
-        mod_path = line.scan(GITHUB_REPO_REGEX).first
-        return unless mod_path
-
-        # Query for _any_ version of this module, to know if it doesn't exist (or is private)
-        # or we were just given a bad revision by this manifest
-        SharedHelpers.in_a_temporary_directory do
-          SharedHelpers.with_git_configured(credentials: credentials) do
-            File.write("go.mod", "module dummy\n")
-
-            env = { "GOPRIVATE" => "*" }
-            _, _, status = Open3.capture3(env, SharedHelpers.escape_command("go get #{mod_path}"))
-            raise Dependabot::DependencyFileNotResolvable, line if status.success?
-
-            mod_split = mod_path.split("/")
-            repo_path = if mod_split.size > 3
-                          mod_split[0..2].join("/")
-                        else
-                          mod_path
-                        end
-            raise Dependabot::GitDependenciesNotReachable, [repo_path]
-          end
-        end
+        msg = stderr.gsub(path.to_s, "").strip
+        raise Dependabot::DependencyFileNotParseable.new(go_mod.path, msg)
       end
 
       def rev_identifier?(dep)
--- a/go_modules/lib/dependabot/go_modules/file_parser.rb
+++ b/go_modules/lib/dependabot/go_modules/file_parser.rb
@@ -60,7 +60,7 @@ module Dependabot
         @required_packages ||=
           SharedHelpers.in_a_temporary_directory do |path|
             # Create a fake empty module for each local module so that
-            # `go list` works, even if some modules have been `replace`d with
+            # `go mod edit` works, even if some modules have been `replace`d with
             # a local module that we don't have access to.
             local_replacements.each do |_, stub_path|
               Dir.mkdir(stub_path) unless Dir.exist?(stub_path)
--- a/go_modules/lib/dependabot/go_modules/file_parser.rb
+++ b/go_modules/lib/dependabot/go_modules/file_parser.rb
@@ -59,36 +59,33 @@ module Dependabot
       def required_packages
         @required_packages ||=
           SharedHelpers.in_a_temporary_directory do |path|
-            SharedHelpers.with_git_configured(credentials: credentials) do
-              # Create a fake empty module for each local module so that
-              # `go list` works, even if some modules have been `replace`d with
-              # a local module that we don't have access to.
-              local_replacements.each do |_, stub_path|
-                Dir.mkdir(stub_path) unless Dir.exist?(stub_path)
-                FileUtils.touch(File.join(stub_path, "go.mod"))
-              end
-
-              File.write("go.mod", go_mod_content)
-              File.write("main.go", "package dummypkg\n func main() {}\n")
-
-              command = "go mod edit -json"
-
-              # Turn off the module proxy for now, as it's causing issues with
-              # private git dependencies
-              env = { "GOPRIVATE" => "*" }
-
-              stdout, stderr, status = Open3.capture3(env, command)
-              handle_parser_error(path, stderr) unless status.success?
-              JSON.parse(stdout)["Require"]
-            rescue Dependabot::DependencyFileNotResolvable
-              # We sometimes see this error if a host times out.
-              # In such cases, retrying (a maximum of 3 times) may fix it.
-              retry_count ||= 0
-              raise if retry_count >= 3
-
-              retry_count += 1
-              retry
+            # Create a fake empty module for each local module so that
+            # `go list` works, even if some modules have been `replace`d with
+            # a local module that we don't have access to.
+            local_replacements.each do |_, stub_path|
+              Dir.mkdir(stub_path) unless Dir.exist?(stub_path)
+              FileUtils.touch(File.join(stub_path, "go.mod"))
             end
+
+            File.write("go.mod", go_mod_content)
+
+            command = "go mod edit -json"
+
+            # Turn off the module proxy for now, as it's causing issues with
+            # private git dependencies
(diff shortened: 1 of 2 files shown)


go_modules/spec/dependabot/go_modules/file_parser_spec.rb
```diff
@@ -37,7 +37,7 @@
   describe "parse" do
     subject(:dependencies) { parser.parse }
 
-    its(:length) { is_expected.to eq(6) }
+    its(:length) { is_expected.to eq(5) }
```
jurre, line 40: We no longer pull in `rsc.io/sampler` since it's not required

go_modules/lib/dependabot/go_modules/file_parser.rb
```diff
@@ -78,17 +69,18 @@ def module_info
               end
 
               File.write("go.mod", go_mod_content)
+              File.write("main.go", "package dummypkg\n func main() {}\n")
 
-              command = "go mod edit -print > /dev/null"
-              command += " && go list -m -json all"
+              command = "go get > /dev/null" # Verify the packages are installable
```
jurre, line 74: This does add a bit of overhead, and as discussed in the PR body, I doubt that it's necessary. The upside is that we fail early in the sad path, the downside is we have to duplicate error handling between this class and the `GoModUpdater`, and it adds some overhead in the happy path.

go_modules/lib/dependabot/go_modules/file_parser.rb
```diff
@@ -78,17 +69,18 @@ def module_info
               end
 
               File.write("go.mod", go_mod_content)
+              File.write("main.go", "package dummypkg\n func main() {}\n")
```
fatih, line 72: Why do we add this here? I think it would be nice to add some comment maybe explaining the reasoning.

go_modules/lib/dependabot/go_modules/file_parser.rb
fatih: Would this slow down the pull request opening latency? I assume beause we're running this twice, this means there will be significant overhead as you meant. I'm in favor of removing it. In fact maye this method's only concern should be `return the list of required packages`. If those packages are installable or valid should be handled probably somewhere else. Thoughts on that ?

fatih (approved):
LGTM 👍🏼  Just left a comment about `go get` behavior, otherwise I'm glad we changed this bit to more efficient. I think if we don't add `go get` and because we removed  `go list`, this will also significantly improve the time we open pull requests for Go modules.

go_modules/lib/dependabot/go_modules/file_parser.rb
jurre: > Would this slow down the pull request opening latency?

A bit, it's hard to say how significant it would be, we definitely perform some heavier operations in the actual updating step, but I noticed our tests for this this class increased by ~40 seconds after adding this `go get` vs `main`, and our test `go.mod` files are typically very small. When we remove the `go get`, we speed up this step significantly as we do not need to make any network calls.

> I'm in favor of removing it. In fact maye this method's only concern should be return the list of required packages. If those packages are installable or valid should be handled probably somewhere else. Thoughts on that ?

I tend to agree! 👍

go_modules/lib/dependabot/go_modules/file_parser.rb
jurre: Yeah good call, we add it so that we can run `go get`, we perform this step in a temp directory where we copy in just the `go.mod` and `go.sum` files, and without the `main.go` we get `go get .: path  is not a package in module rooted at`.

I agree that we should clarify this if we indeed decide to keep it, but if we get rid of `go get`, we can also nuke this step

hmarr:
I just want to make sure I understand the problem this solves. Does the following situation capture it correctly?

Project `a` depends on package `c` in module `b`. Module `b` includes both packages `c` and `d`. Package `d` in module `b` has a dependency on a private module `e`, which isn't accessible (e.g. because it's on a corporate network). Currently this causes update jobs to fail, but after this change those jobs would succeed?

jurre:
> I just want to make sure I understand the problem this solves. Does the following situation capture it correctly?
> 
> Project `a` depends on package `c` in module `b`. Module `b` includes both packages `c` and `d`. Package `d` in module `b` has a dependency on a private module `e`, which isn't accessible (e.g. because it's on a corporate network). Currently this causes update jobs to fail, but after this change those jobs would succeed?

Yes, I think that's exactly the scenario 👍 with the caveat that package `e` is not required in project `a`.

hmarr:
Sorry for all the questions - I just want to clarify this bit:

> The way that this is implemented, it causes every transitive dependency to be listed, even if they are not required by this project.

By "required", do you mean like "needed by the project", as opposed to `require`d in the `go.mod`? If that's the case, it sounds like we're still expecting "necessary" (but not necessarily `require`d) transitive dependencies to show up. I have a suspicion that this change removes all transitive dependencies from the list, perhaps other than the non-Go-modules `// indirect` ones, as they're listed in the top level `go.mod`. Is that what you're expecting?

I ask because we probably will want to support updating of transitive dependencies at some point. Though if it's causing problems right now, I'm happy to pull it out and we can re-add it down the line.

jurre:
> Sorry for all the questions - I just want to clarify this bit:

No worries!

> > The way that this is implemented, it causes every transitive dependency to be listed, even if they are not required by this project.
> 
> By "required", do you mean like "needed by the project", as opposed to `require`d in the `go.mod`? If that's the case, it sounds like we're still expecting "necessary" (but not necessarily `require`d) transitive dependencies to show up. I have a suspicion that this change removes all transitive dependencies from the list, perhaps other than the non-Go-modules `// indirect` ones, as they're listed in the top level `go.mod`. Is that what you're expecting?
> 
> I ask because we probably will want to support updating of transitive dependencies at some point. Though if it's causing problems right now, I'm happy to pull it out and we can re-add it down the line.

No, the scenario where this happens is actually a little different, apologies for not spotting that:

```
module a requires module b
  module b defines pkg c and pkg d
    pkg d requires module e

github/alive only requires pkg c from module b
```

In this case, module `a` has no transitive dependency on module `e`. It's not required in the `go.mod` and it is never imported anywhere in the code, it has an entry in the `go.sum`, but other than that it's not needed nor referenced.

We still grab all the `// indirect` dependencies by using `go mod edit -json`, even though we discard them in [the next step](https://github.com/dependabot/dependabot-core/blob/8fa6c1d17a56304d0dfcf751e45531e8d3343576/go_modules/lib/dependabot/go_modules/update_checker.rb#L26)

jurre:
So, the downside of this approach is that we do not parse any regular transitive dependencies, only those listed in the `go.mod` file, but not regular transitive deps for those that have a `go.mod` file.

Example provided by `@hmarr`:

```
gmtest $ cat go.mod
module foo
go 1.14
require github.com/stretchr/testify v1.6.1
gmtest $ go mod edit -json
{
	"Module": {
		"Path": "foo"
	},
	"Go": "1.14",
	"Require": [
		{
			"Path": "github.com/stretchr/testify",
			"Version": "v1.6.1"
		}
	],
	"Exclude": null,
	"Replace": null
}
gmtest $ go list -m -json all
{
	"Path": "foo",
	"Main": true,
	"Dir": "/tmp/gmtest",
	"GoMod": "/tmp/gmtest/go.mod",
	"GoVersion": "1.14"
}
{
	"Path": "github.com/davecgh/go-spew",
	"Version": "v1.1.0",
	"Time": "2016-10-29T20:57:26Z",
	"Indirect": true,
	"Dir": "/Users/harry/pkg/mod/github.com/davecgh/go-spew@v1.1.0",
	"GoMod": "/Users/harry/pkg/mod/cache/download/github.com/davecgh/go-spew/@v/v1.1.0.mod"
}
{
	"Path": "github.com/pmezard/go-difflib",
	"Version": "v1.0.0",
	"Time": "2016-01-10T10:55:54Z",
	"Indirect": true,
	"Dir": "/Users/harry/pkg/mod/github.com/pmezard/go-difflib@v1.0.0",
	"GoMod": "/Users/harry/pkg/mod/cache/download/github.com/pmezard/go-difflib/@v/v1.0.0.mod"
}
{
	"Path": "github.com/stretchr/objx",
	"Version": "v0.1.0",
	"Time": "2018-01-06T01:13:53Z",
	"Indirect": true,
	"Dir": "/Users/harry/pkg/mod/github.com/stretchr/objx@v0.1.0",
	"GoMod": "/Users/harry/pkg/mod/cache/download/github.com/stretchr/objx/@v/v0.1.0.mod"
}
{
	"Path": "github.com/stretchr/testify",
	"Version": "v1.6.1",
	"Time": "2020-06-05T10:48:45Z",
	"Dir": "/Users/harry/pkg/mod/github.com/stretchr/testify@v1.6.1",
	"GoMod": "/Users/harry/pkg/mod/cache/download/github.com/stretchr/testify/@v/v1.6.1.mod",
	"GoVersion": "1.13"
}
{
	"Path": "gopkg.in/check.v1",
	"Version": "v0.0.0-20161208181325-20d25e280405",
	"Time": "2016-12-08T18:13:25Z",
	"Indirect": true,
	"Dir": "/Users/harry/pkg/mod/gopkg.in/check.v1@v0.0.0-20161208181325-20d25e280405",
	"GoMod": "/Users/harry/pkg/mod/cache/download/gopkg.in/check.v1/@v/v0.0.0-20161208181325-20d25e280405.mod"
}
{
	"Path": "gopkg.in/yaml.v3",
	"Version": "v3.0.0-20200313102051-9f266ea9e77c",
	"Time": "2020-03-13T10:20:51Z",
	"Indirect": true,
	"Dir": "/Users/harry/pkg/mod/gopkg.in/yaml.v3@v3.0.0-20200313102051-9f266ea9e77c",
	"GoMod": "/Users/harry/pkg/mod/cache/download/gopkg.in/yaml.v3/@v/v3.0.0-20200313102051-9f266ea9e77c.mod"
}
```

Since we do not update transitive dependencies for go modules, I think we should update the FileParser to omit them entirely, and also remove the `go get` "safeguard".

Alternatively we could parse the `go.sum` file for all listed dependencies with a native helper, something along the lines of this will do the trick:

```go
package parser

import (
	"bytes"
	"encoding/json"
	"fmt"
	"io/ioutil"
	"path/filepath"
	"strings"

	"golang.org/x/mod/module"
)

type result struct {
	Packages []module.Version
}

func ParseSumfile() (interface{}, error) {
	data, err := ioutil.ReadFile("go.sum")
	if err != nil {
		return nil, err
	}

	file, err := filepath.Abs("go.sum")
	if err != nil {
		return nil, err
	}

	fmt.Println(file)

	m, err := readGoSum(file, data)

	res := result{Packages: m}

	p, err := json.Marshal(res)
	if err != nil {
		return nil, err
	}

	return string(p), nil
}

// emptyGoModHash is the hash of a 1-file tree containing a 0-length go.mod.
// A bug caused us to write these into go.sum files for non-modules.
// We detect and remove them.
const emptyGoModHash = "h1:G7mAYYxgmS0lVkHyy2hEOLQCFB0DlQFTMLWggykrydY="

// readGoSum parses data, which is the content of file,
// and adds it to goSum.m. The goSum lock must be held.
func readGoSum(file string, data []byte) ([]module.Version, error) {
	dst := []module.Version{}

	lineno := 0
	for len(data) > 0 {
		var line []byte
		lineno++
		i := bytes.IndexByte(data, '\n')
		if i < 0 {
			line, data = data, nil
		} else {
			line, data = data[:i], data[i+1:]
		}
		f := strings.Fields(string(line))
		if len(f) == 0 {
			// blank line; skip it
			continue
		}
		if len(f) != 3 {
			return dst, fmt.Errorf("malformed go.sum:\n%s:%d: wrong number of fields %v", file, lineno, len(f))
		}
		if f[2] == emptyGoModHash {
			// Old bug; drop it.
			continue
		}
		mod := module.Version{Path: f[0], Version: f[1]}
		dst = append(dst, mod)
	}
	return dst, nil
}
```

But I would suggest we cross that bridge when we do decide to start updating transitive dependencies.

cc @fatih wdyt?

fatih:
@jurre the current approach looks good for me if we don't want to return the transitive dependencies. It makes things easier in terms of readability as we only take care of listing the direct dependencies. As a bonus, we can later add a `transitive_dependencies` function that would return those if we're interested in those, which would make the code more readable in the future.

hmarr:
I'd rather get rid of all transitive dependencies than parse the `go.sum`. Particularly as `go.sum` files often contain old dependencies (if `go mod tidy` isn't run) and sometimes are missing altogether. Even when we do add support for transitive dependencies, I'd rather rely exclusively on the `go.mod`.

jurre:
> I'd rather get rid of all transitive dependencies than parse the `go.sum`. Particularly as `go.sum` files often contain old dependencies (if `go mod tidy` isn't run) and sometimes are missing altogether. Even when we do add support for transitive dependencies, I'd rather rely exclusively on the `go.mod`.

👍 I agree, that's what I've done in the latest commit FWIW

feelepxyz:
@jurre if you rebase/force push this branch ci should kick in, accidentally turned off actions while testing their repo select UI last week 🤦

jurre:
@thepwagner I would love to get your thoughts on the trade-offs made in this PR. We're currently still running `go get` in the FileParser to verify the packages exist, but I wonder if it makes more sense to let the FileUpdater worry about this

go_modules/lib/dependabot/go_modules/file_parser.rb
feelepxyz: Yeah 👍  on removing `go get` from this step, seems like we're overloading the parser by doing resolvability that should be handled by the update checker/file updater.

go_modules/lib/dependabot/go_modules/file_parser.rb
```diff
@@ -16,17 +16,8 @@ class FileParser < Dependabot::FileParsers::Base
...
-        deps = chunks.values.map { |chunk| JSON.parse(chunk.join) }
-
-        deps.each do |dep|
-          # The project itself appears in this list as "Main"
-          next if dep["Main"]
-
-          dependency = dependency_from_details(dep)
-          dependency_set << dependency if dependency
+        required_packages.each do |dep|
+          dependency_set << dependency_from_details(dep) unless dep["Indirect"]
```
thepwagner, line 20: Question: is skipping `Indirect` dependencies a requirement of this change?

I get that [they're filtered out](https://github.com/dependabot/dependabot-core/blob/b07eeda901af8ef769edb678edb31647d7bc0028/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb#L328) later, but is _that_ behaviour correct?

I don't think we should scope creep into supporting indirect dependencies in this PR - more surprised that we're not currently.

go_modules/lib/dependabot/go_modules/file_parser.rb
```diff
@@ -65,8 +56,8 @@ def dependency_from_details(details)
         )
       end
 
-      def module_info
-        @module_info ||=
+      def required_packages
+        @required_packages ||=
           SharedHelpers.in_a_temporary_directory do |path|
             SharedHelpers.with_git_configured(credentials: credentials) do
```
thepwagner, line 62: Idea: iff we don't `go list -m` any more, do we still need git configured here? I'd expect parsing the file and converting to json to be local-only...

thepwagner (approved):
I dig the `FileParser`'s reduction in scope to literally parsing the file.
We could probably do that without affecting transitive dependency support, but per discussion transitive dependencies will take out-of-scope follow-ups no matter what. This LGTM for what's supported today.

go_modules/lib/dependabot/go_modules/file_parser.rb
jurre: Yes, you're absolutely right! 👍 will get rid of this

go_modules/lib/dependabot/go_modules/file_parser.rb
jurre: > Question: is skipping Indirect dependencies a requirement of this change?

It's more of a consequence, really. `go list -m` only includes transitive dependencies that are included in the `go.mod` file, which apparently are only transitive dependencies that aren't go modules.

We filter indirect dependencies out much more explicitly in the [UpdateChecker](https://github.com/dependabot/dependabot-core/blob/b07eeda901af8ef769edb678edb31647d7bc0028/go_modules/lib/dependabot/go_modules/update_checker.rb#L21-L26) though, and seems to be a deliberate limitation of dependabot currently.

go_modules/lib/dependabot/go_modules/file_parser.rb
```diff
@@ -59,36 +59,33 @@ def dependency_from_details(details)
...
-            rescue Dependabot::DependencyFileNotResolvable
-              # We sometimes see this error if a host times out.
-              # In such cases, retrying (a maximum of 3 times) may fix it.
-              retry_count ||= 0
-              raise if retry_count >= 3
-
-              retry_count += 1
-              retry
+            # Create a fake empty module for each local module so that
+            # `go list` works, even if some modules have been `replace`d with
```
feelepxyz, line 63: Should this say `go mod edit`? sidenote: looks like line suggestions are missing from the editor  🤔

PR #2891, opened 2020-12-18
Verify if composer name is compatible with v2

jurre:
Composer v2 introduced a check on package names, and when no lockfile is
present to validate the plugin-api-version this causes us to attempt the
update with v2, which will fail because of this invalid name.

This includes an extra check to make sure the name is compatible with
the constraints of composer v2.

### DIFF
--- a/composer/spec/fixtures/projects/v1/invalid_v2_name/composer.json
+++ b/composer/spec/fixtures/projects/v1/invalid_v2_name/composer.json
@@ -1,0 +1,12 @@
+{
+  "name": "this name is no longer valid in composerv2",
+  "version": "3.1.7",
+  "dist": {
+    "url": "https://www.example.net/files/3.1.7.zip",
+    "type": "zip"
+  },
+  "require": {
+    "monolog/monolog" : "1.0.1",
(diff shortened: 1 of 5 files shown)


composer/spec/fixtures/projects/v1/invalid_v2_name/composer.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "this name is no longer valid in composerv2",
```
feelepxyz, line 2: woah this was valid in v1 🙈

feelepxyz (approved):
LGTM! Nice cleanup and thanks for referencing the cryptic regex 💯

composer/spec/fixtures/projects/v1/invalid_v2_name/composer.json
jurre: Yeah, it's been throwing warnings about these for a while though 😄 but we run into it reasonably often still, I'm guessing on older packages mostly

feelepxyz:
@jurre spec faillure looks like an unrelated flaky python spec, looking into a fix but this is probably good to merge

PR #1376, opened 2019-09-10
CircleCI: find Python syntax errors and undefined names

cclauss:
Like #1375 but without GitHub Actions.  Find syntax errors and undefined names in Python code like the ones fixed in #1219.

PR #2892, opened 2020-12-18
GitHub Action: Add Python flake8 linting

cclauss:
@feelepxyz as recommended at https://github.com/dependabot/dependabot-core/pull/1376#issuecomment-748133905

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,7 +90,7 @@ jobs:
       - name: Run Python flake8 linting
         if: matrix.suite == 'python'
         run: |
-          IGNORE=E127,E201,E202,E302,E306,F401
+          IGNORE=E126,E127,E201,E202,E302,E306,F401
           docker run --rm "$CORE_CI_IMAGE" bash -c "pyenv exec flake8 . --count --exclude=./.*,./python/spec/fixtures --ignore=$IGNORE --show-source --statistics"
       - name: Run Ruby Rubocop linting
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,7 +87,11 @@ jobs:
         run: |
           docker push "$CORE_CI_IMAGE:latest"
           docker push "$CORE_CI_IMAGE:ci--$BRANCH_REF"
-      - name: Run Rubocop linting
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -87,7 +87,11 @@ jobs:
         run: |
           docker push "$CORE_CI_IMAGE:latest"
           docker push "$CORE_CI_IMAGE:ci--$BRANCH_REF"
-      - name: Run Rubocop linting
+      - name: Run Python flake8 linting
+        run: |
```
jurre, line 91: Currently this will run for _every_ step in the matrix (18 times), I think we could limit it to only the python build by doing:

```suggestion
        if:  matrix.suite == 'python'
        run: |
```

.github/workflows/ci.yml
```diff
@@ -87,7 +87,12 @@ jobs:
         run: |
           docker push "$CORE_CI_IMAGE:latest"
           docker push "$CORE_CI_IMAGE:ci--$BRANCH_REF"
-      - name: Run Rubocop linting
+      - name: Run Python flake8 linting
+        if: matrix.suite == 'python'
+        run: |
+          sudo apt-get install python3-pip
```
jurre, line 93: Do we need to install python3 on the host image if we're running pip inside the container?

The container should already have `pip` installed via `pyenv` and it's available as `pyenv exec pip`, but I suggest we add flake8 to the requirements.txt so it gets cached in the docker build, and then we can run `pyenv exec flake8` here.

FYI if you want to test this out locally, the easiest way is to clone the project and run `bin/docker-dev-shell`, it will give you the actual container that it used in the `docker run` step below

jurre:
@cclauss the CI failures seem unrelated, would you mind trying to rebase these changes on the main branch?

jurre:
I'd say let's just fix those flake8 violations instead if ignoring them: https://github.com/dependabot/dependabot-core/commit/3796846fd141dbd2b3e4108e5b8ae58b390d558f

jurre (approved):

jurre:
Thanks so much for this improvement @cclauss, just waiting for CI to finish and will then go ahead and merge this 🎉

PR #2893, opened 2020-12-18
Resolve Python undefined name `hass_const`

cclauss:
@feelepxyz `import homeassistant.const as hass_const` as discussed in #1376

Also related to #2892

https://github.com/home-assistant/core/blob/dev/setup.py#L7 -->
https://github.com/home-assistant/core/blob/dev/homeassistant/const.py#L7


jurre:
This is a fixture that is deliberately broken, I don't think we'll want to "fix" it! We should probably only lint our native [python helpers](https://github.com/dependabot/dependabot-core/tree/b91ad6b0434d3f755704a2e0e6ee8323619c872e/python/helpers)

PR #2897, opened 2020-12-21
Upgrade Python version to 3.9.1

ulgens:


ulgens:
@jurre @feelepxyz :wave:

Dockerfile
```diff
@@ -68,10 +68,10 @@ RUN apt-get install -y software-properties-common \
 ENV PYENV_ROOT=/usr/local/.pyenv \
   PATH="/usr/local/.pyenv/bin:$PATH"
 RUN git clone https://github.com/pyenv/pyenv.git /usr/local/.pyenv \
-  && cd /usr/local/.pyenv && git checkout v1.2.21 && cd - \
-  && pyenv install 3.9.0 \
+  && cd /usr/local/.pyenv && git checkout 2bf6111fa0bad10b78c3130dfab497c8c3dcd2b6 && cd - \
```
jurre, line 71: Should we maybe wait until they've released a version of pyenv that includes this vs checking out an unreleased sha? We've ran into some issues with using a pre-release version in the past

ulgens: Hey @jurre , I would agree with this if pyenv is releasing versions faster :slightly_smiling_face: I could ping them to release a version but from my past experiences, it will take a while. Instead of waiting them for a release and blocking every Python 3.9.1 & Dependabot user, i think this is better. We already used that method for pyenv in this repo and i changed it back to the release tag when there is one. If you remember what was the issue in past, i can try to solve it.

jurre: I think it was just a bug that snuck into `pyenv` at the time, not something that was easy to detect from what I remember.

It would be awesome if we could rely on officially released versions, so I definitely do think it's worth bringing this up with the `pyenv` team. Having said that, we are also investigating alternative ways to install python versions, currently running a non-pre-installed version makes dependabot runs pretty slow.

I've compared the diff between the `v1.2.21` and this sha, and it looks pretty safe, so I'm going to merge this PR 👍

python/spec/dependabot/python/update_checker/pipenv_version_resolver_spec.rb
```diff
@@ -255,7 +255,7 @@
                 to start_with("Dependabot detected the following Python")
               expect(error.message).to include("3.4.*")
               expect(error.message).
-                to include("supported in Dependabot: 3.9.0, 3.8.6, 3.8.5")
+                to include("supported in Dependabot: 3.9.1, 3.8.6, 3.8.5")
```
jurre, line 258:
```suggestion
                to include("supported in Dependabot: 3.9.1, 3.9.0, 3.8.6")
```

jurre: I wish we could change this test so we don't have to update it every time we add a new python 🤔

feelepxyz (approved):

ulgens:
@jurre @feelepxyz Any ETA on this going live? Dependabot created 16 issues on only one of our repos, saying "Dependabot can't resolve your Python dependency files". It just feels like a spammer at this point.

PR #2904, opened 2020-12-22
Composer: use github token in test env 

feelepxyz:
Fixes flaky test failures where composer runs into gh rate limits:
`Could not authenticate against github.com`

Requires `DEPENDABOT_TEST_ACCESS_TOKEN` to be set in CI.

Flattened the git specs as they where pretty hard to reason about with
lots of nested lets.


jurre (approved):
Thanks for flattening these specs 🎉 this is much more readable ❤️

PR #2906, opened 2020-12-22
GoModUpdater: return updated go.mod directly

thepwagner:
Previously Dependabot would extract the changes detected in `go.mod`, then use `go mod edit -{drop,}require` to re-apply the updated paths on to the original `go.mod`.

This PR asserts that since Dependabot supports `go mod tidy`, customers expect `go.mod` to be tidied. That allows us to skip this additional complexity+compute of reapplying updates by returning the tidied contents directly.

Extra care is taken to ensure Dependabot won't inject a go language version (the `go` directive, e.g. `go 1.15`), to avoid regression. AFAICT this will be added by all versions since `go1.12 (released 2019/02/25)`. Meaning the potential for injection only affects modules that immediately upgrade to vgo when it became available, `go1.11 (released 2018/08/24)`, and haven't updated since. Likely rare.

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
@@ -193,15 +193,6 @@ module Dependabot
           end
         end
 
-        def add_requirements(requirements)
-          requirements.each do |r|
-            escaped_req = Shellwords.escape("#{r['Path']}@#{r['Version']}")
-            command = "go mod edit -require #{escaped_req}"
-            _, stderr, status = Open3.capture3(ENVIRONMENT, command)
-            handle_subprocess_error(stderr) unless status.success?
-          end
-        end
-
         def in_repo_path(&block)
           SharedHelpers.
             in_a_temporary_repo_directory(directory, repo_contents_path) do
--- a/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
@@ -98,13 +98,21 @@ module Dependabot
             updated_go_sum = original_go_sum ? File.read("go.sum") : nil
             updated_go_mod = File.read("go.mod")
 
-            # running "go get" may inject the current go version; remove
+            # running "go get" may inject the current go version, remove it
             original_go_version = original_go_mod.match(GO_MOD_VERSION)
             updated_go_version = updated_go_mod.match(GO_MOD_VERSION)
             if original_go_version != updated_go_version
-              updated_go_mod = updated_go_mod.lines.map do |l|
-                l.match?(GO_MOD_VERSION) ? original_go_version : l
-              end.compact.join("\n")
+              go_mod_lines = updated_go_mod.lines
+              go_mod_lines.each_with_index do |line, i|
+                next unless line&.match?(GO_MOD_VERSION)
+
+                # replace with the original version
+                go_mod_lines[i] = original_go_version
+                # avoid a stranded newline if there was no version originally
+                go_mod_lines[i + 1] = nil if original_go_version.nil?
+              end
+
+              updated_go_mod = go_mod_lines.compact.join
             end
 
             { go_mod: updated_go_mod, go_sum: updated_go_sum }
--- a/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
@@ -252,7 +252,7 @@ module Dependabot
         end
 
         def module_pathname
-          @module_pathname ||= Pathname.new(repo_contents_path).join(directory)
+          @module_pathname ||= Pathname.new(repo_contents_path).join(directory.sub(%r{^/}, ""))
         end
 
         def substitute_all(substitutions)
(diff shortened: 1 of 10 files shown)


go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -90,32 +92,29 @@ def update_files # rubocop:disable Metrics/AbcSize
...
+            # running "go get" may inject the current go version, remove it
+            original_go_version = original_go_mod.match(GO_MOD_VERSION)
+            updated_go_version = updated_go_mod.match(GO_MOD_VERSION)
+            if original_go_version != updated_go_version
+              go_mod_lines = updated_go_mod.lines
+              go_mod_lines.each_with_index do |line, i|
+                next unless line&.match?(GO_MOD_VERSION)
+
+                # replace with the original version
+                go_mod_lines[i] = original_go_version
```
thepwagner, line 110: Aside: I'm pretty sure this also only matters if `original_go_version` is nil.

What dependabot does with the `go` command won't ever change the required go version(?), only add it if it was missing.
I wrote it defensively just in case, to match the previous behaviour.

jurre:
Woah, this is a super sweet simplification!

> This PR asserts that since Dependabot supports go mod tidy, customers expect go.mod to be tidied. That allows us to skip this additional complexity+compute of reapplying updates by returning the tidied contents directly.

What happens when `go mod tidy` silently fails? https://github.com/dependabot/dependabot-core/blob/0811996de1fdde8ecdd70971d7142e38619b7e99/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb#L130-L134

thepwagner:
> What happens when go mod tidy silently fails?

https://github.com/dependabot/dependabot-core/blob/0811996de1fdde8ecdd70971d7142e38619b7e99/go_modules/spec/dependabot/go_modules/file_updater/go_mod_updater_spec.rb#L229-L239


<details><summary>full files via byebug</summary>
<p>

```
$ rspec spec/dependabot/go_modules/file_updater/go_mod_updater_spec.rb:224
[114, 123] in /home/dependabot/dependabot-core/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
=&gt; 119:             { go_mod: updated_go_mod, go_sum: updated_go_sum }

(byebug) puts original_go_mod
module github.com/dependabot/vgotest

go 1.15

require github.com/googleapis/gnostic v0.3.1
nil
(byebug) puts updated_go_mod
module github.com/dependabot/vgotest

go 1.15

require github.com/googleapis/gnostic v0.5.1 // indirect
```

</p>
</details>

jurre (approved):

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -90,32 +92,29 @@ def update_files # rubocop:disable Metrics/AbcSize
             run_go_vendor
             run_go_mod_tidy
```
kd7lxl, line 93: `go mod tidy` should run before `go mod vendor` to ensure any dependencies removed by `go mod tidy` are also removed from vendors.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -90,32 +92,29 @@ def update_files # rubocop:disable Metrics/AbcSize
...
-            # Put back the original content before we replace just the updated
-            # dependencies.
-            write_go_mod(original_go_mod)
-
-            remove_requirements(req_paths_to_remove)
-            deps = updated_reqs.map { |r| requirement_to_dependency_obj(r) }
-            update_go_mod(deps)
-
             # put the old replace directives back again
             substitute_all(substitutions.invert)
```
kd7lxl, line 96: The `replace` directives need to be restored _before_ calling `go mod tidy`, or the module graph will be calculated for the wrong versions of modules, which may cause different indirect dependencies.

thepwagner: 💯 , running `go mod {tidy,vendor}` while Dependabot has stubbed out external modules isn't going to work - I've added a failing test case.

I think instead of inverting the order, Dependabot should only call `go mod tidy` if no modules have been stubbed:
- `go mod tidy` will see different indirect dependencies, as noted
- `go mod vendor` will copy the stubbed module into `vendor/`, which will certainly break downstream CI

jurre (approved):

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -89,11 +88,15 @@ def update_files # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity
...
-            run_go_vendor
-            run_go_mod_tidy
 
-            # put the old replace directives back again
-            substitute_all(substitutions.invert)
+            # If we stubbed modules, don't run `go mod {tidy,vendor}` as
+            # dependencies are incomplete
+            if substitutions.empty?
+              run_go_mod_tidy
+              run_go_vendor
```
kd7lxl, line 96: In many projects, we have PR checks that run `go mod tidy && go mod vendor` and assert `git diff --exit-code` to ensure PRs that add/remove dependencies also update the go.mod and vendors. A Dependabot upgrade that removes an indirect dependency will fail this check now -- failing the check is a lot better than merging bad code into the repo, but would still be a nice case to support. That would be the benefit of running a final `go mod tidy && go mod vendor` on the new go.mod.

PR #2907, opened 2020-12-22
Gradle: Handle missing required manifest file

feelepxyz:
Raise DependencyFileNotFound when no build.gradle/build.gradle.kts file
is found.

Fixed spec fixtures to match repo response where we only fetch files
that are listed in the directory listing.


gradle/lib/dependabot/gradle/file_fetcher.rb
```diff
@@ -31,14 +31,15 @@ def fetch_files
...
         fetched_files += dependency_script_plugins
+        check_required_files_present
         fetched_files
       end
 
       def buildfile
         @buildfile ||= begin
           file = supported_build_file
           @buildfile_name ||= file.name if file
-          fetch_file_from_host(file.name) if file
```
feelepxyz, line 41: the file is already fetched

gradle/lib/dependabot/gradle/file_fetcher.rb
```diff
@@ -59,6 +60,8 @@ def subproject_buildfiles
       end
 
       def dependency_script_plugins
+        return [] unless buildfile
+
         dependency_plugin_paths =
           buildfile.content.
```
feelepxyz, line 66: This was raising if there's no buildfile

gradle/spec/fixtures/github/contents_java_with_subdir_settings.json
```diff
@@ -0,0 +1,742 @@
+[
+  {
+    "name": "build.gradle",
+    "path": "build.gradle",
+    "sha": "5dbaa74de419bb77b47d054a3374e309aa91eeab",
+    "size": 2354,
+    "url": "https://api.github.com/repos/VoxelGamesLib/dependencies/contents/build.gradle?ref=master",
```
thepwagner, line 7: Question: where did this fixture come from? Could it be reduced?

It reads like a mix of files from three different repositories:
* https://github.com/VoxelGamesLib/dependencies , a single module maven project
* https://github.com/Samourai-Wallet/samourai-wallet-android , a multimodule groovy gradle project
* https://github.com/togglz/togglz a multimodule maven project

I'd dig if we could reduce the franken-fixture to just the required files. It'd be extra nice if it were from a single functional repository.

thepwagner (approved):
`**/*.rb` LGTM, my biggest concern is potential fixture cruft - safe to ignore.

gradle/spec/fixtures/github/contents_java_with_subdir_settings.json
feelepxyz: Yeah great shout, the fixtures where real confusing. Updated to dependabot fixture repos: https://github.com/dependabot-fixtures/

PR #2908, opened 2020-12-22
[azure] Add retries for Azure client

GiriB:
Azure client today doesn't implement any retries. When working with bigger repos, we see (transient) ServiceUnavailable often. This PR adds retries for the standard 500, 502 and 503 errors.

### DIFF
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -7,7 +7,9 @@ module Dependabot
   module Clients
     class Azure
       class NotFound < StandardError; end
+      class ServiceNotAvailaible < StandardError; end
 
+      RETRYABLE_ERRORS = [Dependabot::Clients::Azure::ServiceNotAvailaible].freeze
       MAX_PR_DESCRIPTION_LENGTH = 3999
 
       #######################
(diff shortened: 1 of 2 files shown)


common/lib/dependabot/clients/azure.rb
```diff
@@ -8,6 +8,14 @@ module Clients
     class Azure
       class NotFound < StandardError; end
 
+      class InternalServerError < StandardError; end
```
GiriB, line 11: Defined these to throw better error messages.

GiriB, line 11: Added these to make up better stacktrace. Currently errors are show like `JSON::ParseError` while parsing failed response body.

GiriB:
@jurre @brrygrdn Can you help review this? Thanks!

common/lib/dependabot/clients/azure.rb
```diff
@@ -27,10 +35,11 @@ def self.for_source(source:, credentials:)
       # Client #
       ##########
 
-      def initialize(source, credentials)
+      def initialize(source, credentials, max_retries: 3)
         @source = source
         @credentials = credentials
         @auth_header = auth_header_for(credentials&.fetch("token", nil))
+        @max_retries = max_retries || 3
```
jurre, line 42: This is redundant as `max_retries` already has a default argument of 3, but I noticed the other retryable clients follow the same pattern so feel free to leave it in for now

jurre (approved):
Looks great 👍 I'm going to go ahead and merge this, but probably won't cut a new release until after the holidays, as we will be very short staffed and have a short release-freeze.

common/lib/dependabot/clients/azure.rb
GiriB: Thanks!

PR #2909, opened 2020-12-23
[minor] Fix typo in Azure client - Availaible -> Available

GiriB:
A typo went undetected in #2908 :(

### DIFF
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -7,9 +7,15 @@ module Dependabot
   module Clients
     class Azure
       class NotFound < StandardError; end
+
+      class InternalServerError < StandardError; end
+
       class ServiceNotAvailaible < StandardError; end
 
-      RETRYABLE_ERRORS = [Dependabot::Clients::Azure::ServiceNotAvailaible].freeze
+      class BadGateway < StandardError; end
+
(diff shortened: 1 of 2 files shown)


jurre (approved):
😄 totally missed this in the review

jurre:
Any idea why all the specs are failing? 🤔

GiriB:
@jurre Nope. :( Most of the breaks are unrelated to the change. Could it be a transient issue? Is there a way to re-run the specs?

GiriB:
@jurre Is `main` broken? See this PR - #2911 . I created a PR with a dummy text file and the still the CI checks are failing. 

Also, unrelated to this, I saw that my earlier PR was not squashed merge, which I think it should have.
image

jurre:
> Could it be a transient issue? Is there a way to re-run the specs? 

I've already tried to a few times, doesn't seem to do the trick.

> @jurre Is main broken? See this PR - #2911 . I created a PR with a dummy text file and the still the CI checks are failing.

It might be an issue with CI and forks, double checking..

> Also, unrelated to this, I saw that my earlier PR was not squashed merge, which I think it should have.

We typically don't squash merge, I personally try to squash commits before requesting a review so the commits make sense

jurre:
Something definitely seems up with our CI, going to look into it

jurre:
@GiriB if you rebase the suite should pass 🤞

GiriB:
@jurre Done!

PR #2913, opened 2020-12-24
pin faraday for octokit

thepwagner:
Per https://github.com/octokit/octokit.rb/issues/1315

[faraday-1.2.0](https://github.com/lostisland/faraday/releases/tag/v1.2.0) released recently, breaking Dependabot CI.
Specifically, errors are being returned as resources instead of raised as exceptions like `Octokit::NotFound`.

🩹 CI by pinning faraday to the previous version.

# Related
- discovered https://github.com/dependabot/dependabot-core/pull/2911
- discovered https://github.com/dependabot/dependabot-core/pull/2912


jurre (approved):
Nice find!

PR #2918, opened 2020-12-26
Add access token support for Bitbucket.

### LINKED ISSUE #2917: Add Bitbucket Access Token support
The Bitbucket credentials type doesn't support using access token authentication. I think that this can be done like the Azure access token support.

iinuwa:
Fixes #2917.

I'm not sure if this needs tests or how to add tests for this.

### DIFF
--- a/common/lib/dependabot/clients/bitbucket.rb
+++ b/common/lib/dependabot/clients/bitbucket.rb
@@ -18,6 +18,7 @@ module Dependabot
 
       def initialize(credentials:)
         @credentials = credentials
+        @auth_header = auth_header_for(credentials&.fetch("token", nil))
       end
 
       def fetch_commit(repo, branch)
@@ -72,7 +73,9 @@ module Dependabot
           user: credentials&.fetch("username", nil),
           password: credentials&.fetch("password", nil),
           idempotent: true,
-          **Dependabot::SharedHelpers.excon_defaults
+          **Dependabot::SharedHelpers.excon_defaults(
+            headers: auth_header
+          )
         )
         raise Unauthorized if response.status == 401
         raise Forbidden if response.status == 403
@@ -89,6 +92,13 @@ module Dependabot
 
       private
 
+      def auth_header_for(token)
+        return {} unless token
+
+        { "Authorization" => "Bearer #{token}" }
+      end
+
+      attr_reader :auth_header
       attr_reader :credentials
 
       def base_url


thepwagner:
👋 could you help me understand how to manually test this?
* How could I generate one of these tokens? I tried https://id.atlassian.com/manage-profile/security/api-tokens for Bitbucket cloud, but perhaps its limited to Bitbucket server? Could you link to the API reference?
* What would the configuration for credentials look like? A sample command using the `LOCAL_CONFIG_VARIABLES` format from the [dry-run](https://github.com/dependabot/dependabot-core/blob/main/bin/dry-run.rb#L117-L122) script would be awesome. 

For example, I can verify password authentication against Bitbucket cloud with:
```
LOCAL_CONFIG_VARIABLES='[{"type":"git_source","host":"bitbucket.org","username":"thepwagner","password":"REDACTED"}]' bin/dry-run.rb npm_and_yarn thepwagner/dependabot-test-npm
```

iinuwa:
Hello!

> How could I generate one of these tokens? 

Good question. For a particular workspace, you can create an OAuth consumer at https://bitbucket.org/{{ workspace }}/workspace/settings/api. (Can be accessed by going to workspace root > settings > OAuth Consumers).

The client ID and secret can be used to get an access token from this endpoint:  https://bitbucket.org/site/oauth2/access_token. Cf. [Bitbucket Cloud OAuth2 API reference](https://developer.atlassian.com/bitbucket/api/2/reference/meta/authentication#oauth-2). I used the client credentials grant flow with the `repository:write` and `pullrequest:write` scopes. E.g.,

```shell
$ curl -X POST -u "client_id:secret" \
  https://bitbucket.org/site/oauth2/access_token \
  -d "grant_type=client_credentials&scope=repository:write pullrequest:write"
```

returns

```json
{"scopes": "repository:write pullrequest", "access_token": "<access_token>", "expires_in": 7200, "token_type": "bearer", "state": "client_credentials", "refresh_token": "<refresh_token>"}
```

The value of the `access_token` field is what you'll need to set in the `token` field for the credentials object (see below).

> What would the configuration for credentials look like?

I modeled this after the Azure client, which seems to use the `token` field to specify a bearer token. So I think this is what you're looking for?

`LOCAL_CONFIG_VARIABLES='[{"type":"git_source","host":"bitbucket.org","token":"<access token here>"}]'`

I also have an example in my [PR to integrate this into dependabot-script](https://github.com/iinuwa/dependabot-script/blob/feature/bitbucket-support/generic-update-script.rb#L97-L114) if that helps.

> perhaps its limited to Bitbucket server

That's a good point. I tested this on Bitbucket Cloud; I don't have a Bitbucket Server instance, so I can't test it there.

thepwagner (approved):
@iinuwa thanks for the testing instructions, and for separating this change from the follow-up to add PR support.

With your instructions I was able to verify this change in isolation, LGTM!

iinuwa:
Thanks!

PR #2923, opened 2020-12-28
Support PR creation on Bitbucket

### LINKED ISSUE #2919: Add Bitbucket PR creation support
Bitbucket doesn't have an implementation for creating PRs. It would be great to have this, as I think this and #2918 are the only things keeping Bitbucket support from working in dependabot-script.

iinuwa:
This should fix #2919, but depends on #2918.

This gets us most of the way, but I'd appreciate some help polishing it off.

I was able to successfully create a PR with Bitbucket Cloud by modifying dependabot-script (cf. dependabot/dependabot-script#514), but I haven't added any automated tests because I don't really know where to start.

I also had a few questions as I was going through the code. I'll add review comments below.

### DIFF
--- a/common/lib/dependabot/clients/bitbucket.rb
+++ b/common/lib/dependabot/clients/bitbucket.rb
@@ -22,16 +22,14 @@ module Dependabot
           select { |cred| cred["type"] == "git_source" }.
           find { |cred| cred["host"] == source.hostname }
 
-        new(credentials: credential, source: source)
+        new(credentials: credential)
       end
 
       ##########
       # Client #
       ##########
 
-      # FIXME: I don't know if changing the constructor here is safe
-      def initialize(credentials:, source: nil)
-        @source = source
+      def initialize(credentials:)
         @credentials = credentials
         @auth_header = auth_header_for(credentials&.fetch("token", nil))
       end
--- a/common/lib/dependabot/clients/bitbucket.rb
+++ b/common/lib/dependabot/clients/bitbucket.rb
@@ -67,12 +67,9 @@ module Dependabot
       end
 
       def commits(repo, branch_name = nil)
-        commits_path = "#{repo}/commits/" + branch_name.to_s
-        commits_path += "?pagelen=100"
-
-        response = get(base_url + commits_path)
-
-        JSON.parse(response.body).fetch("values")
+        commits_path = "#{repo}/commits/#{branch_name}?pagelen=100"
+        next_page_url = base_url + commits_path
+        paginate({ "next" => next_page_url })
       end
 
       def branch(repo, branch_name)
@@ -85,16 +82,8 @@ module Dependabot
       def pull_requests(repo, source_branch, target_branch)
         pr_path = "#{repo}/pullrequests"
         pr_path += "?status=OPEN&status=MERGED&status=DECLINED&status=SUPERSEDED"
-        next_page = base_url + pr_path
-        pull_requests = []
-        loop do
-          response = get(next_page)
-          result = JSON.parse(response.body)
-          pull_requests.concat result.fetch("values")
-          break unless result.key?("next")
-
-          next_page = result.fetch("next")
-        end
+        next_page_url = base_url + pr_path
+        pull_requests = paginate({ "next" => next_page_url })
 
         pull_requests unless source_branch && target_branch
 
@@ -223,6 +212,31 @@ module Dependabot
         parameters.map do |key, value|
           URI.encode_www_form_component(key.to_s) + "=" + URI.encode_www_form_component(value.to_s)
         end.join("&")
+      end
+
+      # Takes a hash with optional `values` and `next` fields
+      # Returns an enumerator.
+      #
+      # Can be used a few ways:
+      # With GET:
+      #     paginate ({"next" => url})
+      # or
+      #     paginate(JSON.parse(get(url).body))
+      #
+      # With POST (for endpoints that provide POST methods for long query parameters)
+      #     response = post(url, body)
+      #     first_page = JSON.parse(repsonse.body)
+      #     paginate(first_page)
+      def paginate(page)
+        Enumerator.new do |yielder|
+          loop do
+            page.fetch("values", []).each { |value| yielder << value }
+            break unless page.key?("next")
+
+            next_page_url = page.fetch("next")
+            page = JSON.parse(get(next_page_url).body)
+          end
+        end
       end
 
       attr_reader :auth_header
--- a/common/lib/dependabot/clients/bitbucket.rb
+++ b/common/lib/dependabot/clients/bitbucket.rb
@@ -87,12 +87,11 @@ module Dependabot
 
         pull_requests unless source_branch && target_branch
 
-        pull_requests.
-          select do |pr|
-            pr_source_branch = pr.fetch("source").fetch("branch").fetch("name")
-            pr_target_branch = pr.fetch("destination").fetch("branch").fetch("name")
-            pr_source_branch == source_branch && pr_target_branch == target_branch
-          end
+        pull_requests.select do |pr|
+          pr_source_branch = pr.fetch("source").fetch("branch").fetch("name")
+          pr_target_branch = pr.fetch("destination").fetch("branch").fetch("name")
+          pr_source_branch == source_branch && pr_target_branch == target_branch
+        end
       end
 
       # rubocop:disable Metrics/ParameterLists
--- a/common/lib/dependabot/clients/bitbucket.rb
+++ b/common/lib/dependabot/clients/bitbucket.rb
@@ -18,6 +18,7 @@ module Dependabot
 
       def initialize(credentials:)
         @credentials = credentials
+        @auth_header = auth_header_for(credentials&.fetch("token", nil))
       end
 
       def fetch_commit(repo, branch)
@@ -72,7 +73,9 @@ module Dependabot
           user: credentials&.fetch("username", nil),
           password: credentials&.fetch("password", nil),
           idempotent: true,
-          **Dependabot::SharedHelpers.excon_defaults
+          **Dependabot::SharedHelpers.excon_defaults(
+            headers: auth_header
+          )
         )
         raise Unauthorized if response.status == 401
         raise Forbidden if response.status == 403
@@ -89,6 +92,13 @@ module Dependabot
 
       private
 
+      def auth_header_for(token)
+        return {} unless token
+
+        { "Authorization" => "Bearer #{token}" }
+      end
+
+      attr_reader :auth_header
       attr_reader :credentials
 
       def base_url
--- a/common/lib/dependabot/clients/bitbucket.rb
+++ b/common/lib/dependabot/clients/bitbucket.rb
@@ -12,11 +12,26 @@ module Dependabot
 
       class Forbidden < StandardError; end
 
+      #######################
+      # Constructor methods #
+      #######################
+
+      def self.for_source(source:, credentials:)
+        credential =
+          credentials.
+          select { |cred| cred["type"] == "git_source" }.
+          find { |cred| cred["host"] == source.hostname }
+
+        new(credentials: credential, source: source)
+      end
+
       ##########
       # Client #
       ##########
 
-      def initialize(credentials:)
+      # FIXME: I don't know if changing the constructor here is safe
+      def initialize(credentials:, source: nil)
+        @source = source
         @credentials = credentials
         @auth_header = auth_header_for(credentials&.fetch("token", nil))
       end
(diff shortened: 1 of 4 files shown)


common/lib/dependabot/clients/bitbucket.rb
```diff
@@ -12,12 +12,28 @@ class Unauthorized < StandardError; end
...
+        new(credentials: credential, source: source)
+      end
+
       ##########
       # Client #
       ##########
 
-      def initialize(credentials:)
+      # FIXME: I don't know if changing the constructor here is safe
+      def initialize(credentials:, source: nil)
```
iinuwa, line 33: Is this save to add an extra parameter here? (I hadn't done any Ruby before this, so I don't know about this.) It seems to match how the other constructors work.

common/lib/dependabot/clients/bitbucket.rb
```diff
@@ -52,6 +68,92 @@ def fetch_file_contents(repo, commit, path)
         response.body
       end
 
+      def commits(repo, branch_name = nil)
```
iinuwa, line 71: I copied this most of the new methods from the Azure client, but I had to add the `repo` parameter to all of them.

Is this supposed to be using `source.repo` internally for all these parameters?

common/lib/dependabot/clients/bitbucket.rb
```diff
@@ -52,6 +68,92 @@ def fetch_file_contents(repo, commit, path)
...
+        commits_path = "#{repo}/commits/" + branch_name.to_s
+        commits_path += "?pagelen=100"
```
iinuwa, line 73: This only returns the last 100 commits; should this go paginate through all commits instead?

common/lib/dependabot/clients/bitbucket.rb
```diff
@@ -52,6 +68,92 @@ def fetch_file_contents(repo, commit, path)
...
+        loop do
+          response = get(next_page)
+          result = JSON.parse(response.body)
+          pull_requests.concat result.fetch("values")
+          break unless result.key?("next")
+
+          next_page = result.fetch("next")
+        end
+
+        pull_requests unless source_branch && target_branch
```
iinuwa, line 101: Are we supposed to bail if either `source_branch` or `target_branch` is `nil`? Or should we let the provider determine defaults for us if not specified in the client?

common/lib/dependabot/clients/bitbucket.rb
```diff
@@ -52,6 +68,92 @@ def fetch_file_contents(repo, commit, path)
...
+            branch: {
+              name: source_branch
+            }
+          },
+          destination: {
+            branch: {
+              name: target_branch
+            }
+          },
+          description: pr_description,
```
iinuwa, line 148: Bitbucket doesn't like the HTML tags in the generated description; does that need to be fixed in this PR before merging?

common/lib/dependabot/pull_request_creator/bitbucket.rb
```diff
@@ -0,0 +1,98 @@
...
+        @work_item      = work_item
+      end
+
+      def create
+        return if branch_exists? && pull_request_exists?
+
+        # FIXME: Copied from Azure, but not verified whether this is true
+        # For Bitbucket we create or update a branch in the same request as creating
+        # a commit (so we don't need create or update branch logic here)
+        create_commit
```
iinuwa, line 35: As the comment says, this logic needs to be reviewed.

common/lib/dependabot/pull_request_creator/bitbucket.rb
```diff
@@ -0,0 +1,98 @@
...
+
+        create_pull_request
+      end
+
+      private
+
+      # FIXME: I don't know what the "for_source" methods are for
+      def bitbucket_client_for_source
```
iinuwa, line 43: I mentioned this above, but I'm not really sure what the `for_source` methods are for. Could someone explain this a little?

common/lib/dependabot/clients/bitbucket.rb
thepwagner: I suspect other clients do, but 100 commit seems like it might be "enough".
If it's working for you, works for me...

thepwagner, line 73:
```suggestion
        commits_path = "#{repo}/commits/#{branch_name}?pagelen=100"
```

Idea: simplify with string interpolation.

common/lib/dependabot/clients/bitbucket.rb
```diff
@@ -52,6 +68,92 @@ def fetch_file_contents(repo, commit, path)
...
+                        author_details)
+        parameters = {
+          message: commit_message, # TODO: Format markup in commit message
+          author: "#{author_details.fetch(:name)} <#{author_details.fetch(:email)}>",
+          parents: base_commit,
+          branch: branch_name
+        }
+
+        files.each do |file|
+          absolute_path = "/" + file.name unless file.name.start_with?("/")
```
thepwagner, line 122: Concern: what value should `absolute_path` take if `file.name.start_with?("/") == true`?
It might be `nil` currently, which would mean multiple absolute paths conflict.

common/lib/dependabot/clients/bitbucket.rb
```diff
@@ -52,6 +68,92 @@ def fetch_file_contents(repo, commit, path)
...
+          parameters[absolute_path] = file.content
+        end
+
+        body = encode_form_parameters(parameters)
+
+        commit_path = "#{repo}/src"
```
thepwagner, line 128: Xref https://developer.atlassian.com/bitbucket/api/2/reference/resource/repositories/%7Bworkspace%7D/%7Brepo_slug%7D/src#post

Not worth an inline comment, just sharing for fellow reviewers.

common/lib/dependabot/pull_request_creator/bitbucket.rb
```diff
@@ -0,0 +1,98 @@
...
+
+      def create_pull_request
+        bitbucket_client_for_source.create_pull_request(
+          source.repo,
+          pr_name,
+          branch_name,
+          source.branch || default_branch,
+          pr_description,
+          # TODO: Bitbucket doesn't yet support labels. Cf. https://jira.atlassian.com/browse/BCLOUD-11976
+          # labeler.labels_for_pr,
```
thepwagner, line 87:
```suggestion
          nil,
```

Comment SGTM, but since `labels` is a positional arg it needs to be provided so that `work_item` is passed correctly.

common/lib/dependabot/pull_request_creator/bitbucket.rb
thepwagner: I think their main goal is to instantiate the client once and only once (e.g. the `@bitbucket_client_for_source ||=` line is the key).

I'm not sure about the naming - AFAICT there is  only ever one `source` (and therefore one `bitbucket_client`).

common/lib/dependabot/clients/bitbucket.rb
thepwagner: Question: what is using this new `source` field? I skimmed and don't see any references.

common/lib/dependabot/pull_request_creator/bitbucket.rb
```diff
@@ -0,0 +1,98 @@
...
+      def bitbucket_client_for_source
+        @bitbucket_client_for_source ||=
+          Dependabot::Clients::Bitbucket.for_source(
+            source: source,
+            credentials: credentials
+          )
+      end
+
+      def branch_exists?
+        bitbucket_client_for_source.branch(source.repo, branch_name)
```
thepwagner, line 52: Nit: it's weird that we passed `source` to the constructor for `Dependabot::Clients::Bitbucket.for_source`, but still have to pass attributes like `source.repo` to individual methods.

If `source` isn't used by `Dependabot::Clients::Bitbucket` it may be cleanest to remove.

common/lib/dependabot/pull_request_creator/pr_name_prefixer.rb
```diff
@@ -307,6 +309,19 @@ def recent_azure_commit_messages
           map(&:strip)
       end
 
+      # FIXME: Should this paginate through all commits, or is 100 enough?
```
thepwagner, line 312:
```suggestion
```

I think this is expressed in the `commits()` function already.

common/lib/dependabot/pull_request_creator/pr_name_prefixer.rb
```diff
@@ -307,6 +309,19 @@ def recent_azure_commit_messages
...
+      def recent_bitbucket_commit_messages
+        @recent_bitbucket_commit_messages ||=
+          bitbucket_client_for_source.commits(source.repo)
+
+        @recent_bitbucket_commit_messages.
+          reject { |c| bitbucket_commit_author_email(c) == dependabot_email }.
+          reject { |c| c.fetch("message")&.start_with?("Merge") }.
+          map { |c| c.fetch("message") }.
+          compact.
```
thepwagner, line 321:
```suggestion
          map { |c| c.fetch("message") }.
          compact.
          reject { |m| m.start_with?("Merge") }.
```

Idea: reduce to commits with a `message` before filtering on the message contents.

common/lib/dependabot/clients/bitbucket.rb
thepwagner: It should be, since you're providing a default value.

thepwagner (commented):
LGTM, thanks for this!

I'm also confused about how `source` provided to these clients' constructor should interact with `source.repo` required by methods. This long predates me, @feelepxyz or @hmarr is this part of an ongoing refactor? Which should be preferred?

common/lib/dependabot/clients/bitbucket.rb
iinuwa: Meh. I went ahead and implemented pagination for both commits and pull requests so someone else doesn't have to do it later.

common/lib/dependabot/clients/bitbucket.rb
iinuwa: Ah, that was a mistake; I've corrected that.

common/lib/dependabot/clients/bitbucket.rb
iinuwa: I didn't add any references to it since I wasn't sure whether I should prefer a `repo` parameter passed to each method, or to use the instance variable. I went ahead and removed the parameter since I don't think `source` applies to Bitbucket anyway.

(There isn't really an alternative location for Bitbucket. Support for Bitbucket Server is going away in Feb 2024 and the ability to configure custom domain names for Bitbucket Cloud isn't on the Atlassian roadmap to be implemented any time soon, so `bitbucket.org`/`api.bitbucket.org` are the only endpoints that we need to consider—unless someone who can test Bitbucket Server requests the feature.)

common/lib/dependabot/clients/bitbucket.rb
```diff
@@ -12,12 +12,28 @@ class Unauthorized < StandardError; end
 
       class Forbidden < StandardError; end
 
+      #######################
+      # Constructor methods #
+      #######################
+
+      def self.for_source(source:, credentials:)
```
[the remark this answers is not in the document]
iinuwa: Yeah, I noticed that when I tried to remove the `for_source` method. It looks like all the clients do the same thing.

common/lib/dependabot/pull_request_creator/bitbucket.rb
iinuwa: Updating PRs is a [different endpoint](https://developer.atlassian.com/bitbucket/api/2/reference/resource/repositories/%7Bworkspace%7D/%7Brepo_slug%7D/pullrequests/%7Bpull_request_id%7D#put), so we either need to throw an error when attempting to update, or add the endpoint reference.

common/lib/dependabot/pull_request_creator/bitbucket.rb
iinuwa: Gotcha. I went ahead and removed the comment. I agree with the "only ever one source" part. 

> AFAICT there is only ever one `source`

Yeah, I agree. It seems to be the convention for the other clients, though, so I'll defer to you on the name.

common/lib/dependabot/pull_request_creator/pr_name_prefixer.rb
iinuwa: Done, with a modification.

iinuwa:
@thepwagner, I think I've addressed the suggestions you've made so far. I've put them in individual commits in case that's helpful for reviewing. I can squash and rebase after approval and before merging to clean up the history.

thepwagner (approved):
Thanks for this!
LGTM: I've regression tested with the `dry-run.rb` script in this repository.

dependabot/dependabot-script isn't supported by GitHub, so I haven't tested this there - follow-up PRs are welcome if more is needed.

meladRaouf:
I can't use dependabot with bitcucket I got this error 
labeler.rb:241:in `labels': Unsupported provider bitbucket (RuntimeError)

is there anyway to skip labels in bitbucket

PR #2931, opened 2020-12-31
dry-run: configurable source provider

thepwagner:
This modifies the `bin/dry-run.rb` testing script to accept a `--provider PROVIDER` argument, which can be used to select from [any available source providers](https://github.com/dependabot/dependabot-core/blob/f66290a0e341e656784fd1dae0d7e80ea913ebac/common/lib/dependabot/file_fetchers/base.rb#L200-L212).
This makes functionally testing authentication changes like #2918 easier.

To verify this change use a public repository. I searched `package.json site:bitbucket.org` and found [atlassian/atlasboard](https://bitbucket.org/atlassian/atlasboard/src/master/package.json):
```
bin/dry-run.rb --provider bitbucket npm_and_yarn atlassian/atlasboard
```

### DIFF
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -91,6 +91,7 @@ require "dependabot/terraform"
 
 $options = {
   credentials: [],
+  provider: "github",
   directory: "/",
   dependency_names: nil,
   branch: nil,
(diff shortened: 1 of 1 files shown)


feelepxyz (approved):
👌

PR #2942, opened 2021-01-04
actions: accept shortref hashes

thepwagner:
When dependabot updates an Actions workflow file, it prefers to replace versions with full length commit SHAs.
This is a great practice, reinforced by [Actions's best practices](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions).

This has an unfortunate side effect when the Action is already pinned to the short version of the SHA: Dependabot will propose an "upgrade" to the full SHA.
That's _is_ a good idea from a security standpoint, but not an expected feature of Dependabot. Customers only expect PRs for dependency updates.

This PR adjusts the `Actions` comparison so that short refs are considered equal to full refs when considering upgrades.
That means customers will be upgraded (to full SHAs) as new releases become available, but Dependabot won't open a PR that only replaces a short ref with the expanded full ref.

### DIFF
--- a/github_actions/lib/dependabot/github_actions/update_checker.rb
+++ b/github_actions/lib/dependabot/github_actions/update_checker.rb
@@ -28,6 +28,7 @@ module Dependabot
         updated = updated_source
         return dependency.requirements if updated == previous
 
+        # Maintain a short git hash only if it matches the latest
         if previous[:type] == "git" &&
            previous[:url] == updated[:url] &&
            updated[:ref]&.match?(/^[0-9a-f]{6,40}$/) &&
--- a/github_actions/lib/dependabot/github_actions/update_checker.rb
+++ b/github_actions/lib/dependabot/github_actions/update_checker.rb
@@ -23,10 +23,20 @@ module Dependabot
         dependency.version
       end
 
-      def updated_requirements
-        return dependency.requirements if updated_source == dependency_source_details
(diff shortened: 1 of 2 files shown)


github_actions/lib/dependabot/github_actions/update_checker.rb
```diff
@@ -23,10 +23,20 @@ def latest_resolvable_version_with_no_unlock
...
       end
 
-      def updated_requirements
-        return dependency.requirements if updated_source == dependency_source_details
+      def updated_requirements # rubocop:disable Metrics/PerceivedComplexity
+        previous = dependency_source_details
+        updated = updated_source
+        return dependency.requirements if updated == previous
+
+        if previous[:type] == "git" &&
```
feelepxyz, line 31: Thoughts on adding a comment explaining that this will maintain a short sha if latest?

feelepxyz (approved):
Added a minor nit comment, LGTM otherwise 👍

github_actions/lib/dependabot/github_actions/update_checker.rb
thepwagner:
```suggestion
        # Maintain a short git hash only if it matches the latest
        if previous[:type] == "git" &&
```
SGTM!

PR #2949, opened 2021-01-05
go_modules raise GitDependenciesNotReachable

thepwagner:
This PR restores the functionality introduced by https://github.com/dependabot/dependabot-core/pull/2780 , which was reverted by https://github.com/dependabot/dependabot-core/pull/2880 .

If Dependabot encounters a go module on `github.com` that can not be accessed, it should raise a `GitDependenciesNotReachable` error with details about the requested repositories.
Raising a `GitDependenciesNotReachable` is important as this is used by GitHub.com to recover from the error - https://github.blog/changelog/2020-12-02-dependabot-version-updates-from-private-github-repositories/

Since we no longer `go get` in the `FileParser`, there are two possible places this might be raised:
* In the `UpdateChecker`, if the private repository is a direct dependency
* In the `FlleUpdater`, if the private repository is an indirect dependency


jurre (approved):
Ah! I missed that this wasnt handled in those steps yet, my bad. The extracted module is a great improvement.

PR #2950, opened 2021-01-06
Support GitLab format npm registry

danoe:
Currently it will try to call `/api/v4/projects/229/packages/npm/@prefix%2Fpackage-name`
Which will return: HTTP/1.1 404 Not Found

The correct api endpoint for the package meta data is located at: `/api/v4/packages/npm/@prefix%2Fpackage-name`
While the endpoint for the tar.gz is `api/v4/projects/229/packages/npm/@prefix/package-name/-/@prefix/package-name-0.4.0.tgz`

### DIFF
--- a/npm_and_yarn/spec/dependabot/npm_and_yarn/file_parser_spec.rb
+++ b/npm_and_yarn/spec/dependabot/npm_and_yarn/file_parser_spec.rb
@@ -276,8 +276,6 @@ RSpec.describe Dependabot::NpmAndYarn::FileParser do
           end
 
           describe "the scoped gitlab dependency" do
-
-
             subject { top_level_dependencies[6] }
 
             it { is_expected.to be_a(Dependabot::Dependency) }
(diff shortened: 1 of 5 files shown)


jurre:
Thanks @danoe! The regex looks good, but to be sure, could you add a test for this scenario?

danoe:
Thanks @jurre, I will give it a try.

J-Vin (approved):

danoe:
@jurre, i used one of the other registry tests as a template and created one new test, it seems it covers npm and yarn both, if i understood it correctly. Gitlab Registry as far as i know only supports scoped packages, so i think only this one test is necessary.

npm_and_yarn/spec/dependabot/npm_and_yarn/file_parser_spec.rb
```diff
@@ -275,6 +275,30 @@
             end
           end
 
+          describe "the scoped gitlab dependency" do
+
+
+            subject { top_level_dependencies[6] }
```
jurre:
```suggestion
            subject { top_level_dependencies[6] }
```

jurre (approved):

jurre:
One small stylistic nit, other than that looks great, thanks!

danoe:
@jurre that last one should've been fixed with this, right?: https://github.com/dependabot/dependabot-core/pull/2950/commits/842efc912a56102a32525401cde99ac928c1c028

feelepxyz (approved):
👍

PR #2952, opened 2021-01-06
go_modules: FileParser accept go.mod without dependencies

thepwagner:
It's valid to declare a `go.mod` file that doesn't define any dependencies; when that happens `go mod edit -json` will return:

```
dependabot-core/go_modules/spec/fixtures/projects/no_dependencies $ go mod edit -json
{
	"Module": {
		"Path": "github.com/dependabot/vgotest"
	},
	"Go": "1.12",
	"Require": null,
	"Exclude": null,
	"Replace": null
}
```

This PR modifies the `FileParser` to return an empty array in this case, instead of the nil value.

# Related
- https://github.com/dependabot/dependabot-core/pull/2880


jurre (approved):
Nice find!

feelepxyz (approved):
🙌 Thanks for jumping on this! Specs look 👌

Issue #2910, opened 2020-12-24
"Couldn't find Ruby version!" occurs for Ruby 2.7


ybiquitous:
Hi, I have a `.gemspec` file with `required_ruby_version = ">= 2.7.2"` as follows:

https://github.com/sider/runners/blob/c272ab680b5bf2f1a4f0f268e154941c77085bdf/runners.gemspec#L13

But unfortunately, Dependabot updates fail. 😢 

**Package manager/ecosystem**
ruby:bundler

**Manifest contents prior to update**
`*.gemspec`, `Gemfile`, and `Gemfile.lock`

**Updated dependency**
All Ruby gems in my project

**What you expected to see, versus what you actually saw**
Some outdated packages should be updated. See below:

```console
$ bundle outdated

Gem             Current  Latest   Requested  Groups
aws-partitions  1.412.0  1.413.0
ffi             1.13.1   1.14.2
psych           3.2.1    3.3.0
regexp_parser   2.0.0    2.0.1
steep           0.37.0   0.38.0   = 0.37.0   development
```

**Images of the diff or a link to the PR, issue or logs**
The following is a part of the error log.

```
updater | INFO <job_78815922> Starting job processing
updater | INFO <job_78815922> Starting update job for sider/runners
updater | INFO <job_78815922> Checking if json 2.5.1 needs updating
updater | I, [2020-12-24T04:30:23.673900 #72]  INFO -- sentry: ** [Raven] Sending event 289c81fabf9b4d40816a9057f8c13884 to Sentry
  proxy | 2020/12/24 04:30:24 [042] POST https://sentry.io:443/api/1451818/store/
  proxy | 2020/12/24 04:30:24 [042] 200 https://sentry.io:443/api/1451818/store/
updater | ERROR <job_78815922> Error processing json (RuntimeError)
updater | ERROR <job_78815922> Couldn't find Ruby version!
updater | ERROR <job_78815922> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-bundler-0.129.1/lib/dependabot/bundler/file_updater/ruby_requirement_setter.rb:56:in `ruby_version'
updater | ERROR <job_78815922> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-bundler-0.129.1/lib/dependabot/bundler/file_updater/ruby_requirement_setter.rb:31:in `rewrite'
[...]

PR #2956, opened 2021-01-07
Bundler: support ruby 2.7 and 3.0 in gemspecs

feelepxyz:
Support gemspecs with `required_ruby_version` requirements matching 2.7
and 3.0. Also updated the other major versions to the latest available
version to match stricter requirements.

I'm not sure why we actually do this re-write from a requirement `>=
2.7.2` to `2.7.2`. Removing this logic seems to still work for some
updates that previously failed but fail with others so I don't yet feel
confident in re-writing this.

AFAICT we don't actually rely on these versions being installed, it only seems to be done for resolution where updates fail with exceptions about no compatible ruby version being found if removing the re-write.

Tested that this fix works for projects that require ruby 2.7+ but don't
rely on bundler 2 gems:

`bin/dry-run.rb bundler "HerbCSO/carstens-butler" --dir="/." --commit=331380d8f3e1b7d594405cb973d09f202c3ccfb6 --cache=files`

For projects that rely on gems that requireme bundler 2 the updater now
raises a `dependency_file_not_resolvable` error:

`bin/dry-run.rb bundler sider/runners --commit=c272ab680b5bf2f1a4f0f268e154941c77085bdf`

This should be fixed by upgrading bundler.

### DIFF
--- a/bundler/spec/fixtures/ruby/gemspecs/require_ruby_3
+++ b/bundler/spec/fixtures/ruby/gemspecs/require_ruby_3
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+Gem::Specification.new do |spec|
+  spec.name         = "example"
+  spec.version      = "0.9.3"
+  spec.summary      = "Automated dependency management"
+  spec.description  = "Core logic for updating a GitHub repos dependencies"
+
+  spec.author       = "Dependabot"
+  spec.email        = "support@dependabot.com"
+  spec.homepage     = "https://github.com/hmarr/example"
+  spec.license      = "MIT"
+
+  spec.require_path = "lib"
+  spec.files        = Dir["CHANGELOG.md", "LICENSE.txt", "README.md",
+                          "lib/**/*", "helpers/**/*"]
+
+  spec.required_ruby_version = ">= 3.0.0"
+  spec.required_rubygems_version = ">= 2.6.11"
+
+  spec.add_dependency 'business', '~> 1.0'
+end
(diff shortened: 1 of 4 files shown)


jurre (approved):
Thanks for digging into this, seems like a good step to make sure we support these configs first, but would be ace if we could eventually get rid of this.

feelepxyz:
> would be ace if we could eventually get rid of this.

💯 pushed a change to raise a custom error class instead making it easier to spot these sentries as they are currently rolled up in [this](https://sentry.io/organizations/github/issues/1395979033).

I have a feeling we might not need this anymore since doing the bundler refactor moving everything to native helpers but could be wrong. Let's look at it in more detail when doing the bundler 2 upgrade as one of the updates still fails with a bundler 2 error.

bundler/spec/fixtures/ruby/gemspecs/require_ruby_3
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
```
thepwagner, line 1:
```suggestion
# frozen_string_literal: true

```

Should there be a newline here? `grep -A 1 -r frozen_string .` says probably...
Meh, since when am I rubocop.

thepwagner (approved):
Closes https://github.com/dependabot/dependabot-core/issues/2910 ?

Question: in rspec-fu would it be better to `skip` this test instead of comment it out?
In other languages  I'd argue for skipping as it prevents compilation regressions.

feelepxyz:
> `skip` this test instead of comment it out?

Yes! Forgot `skip` was a thing, was thinking of `pending` but this only works if it consistently fails.

PR #2966, opened 2021-01-12
Use release version of pyenv

ulgens:
pyenv released a new version with Python 3.9.1 https://github.com/pyenv/pyenv/releases/tag/v1.2.22


feelepxyz (approved):
Thanks!

PR #2970, opened 2021-01-12
Use composer v1 when any of the requirements are invalid on v2

jurre:
Composer v2 has added some restrictions on package names. We already
validate that name for the package name itself, but v2 will _also_ raise
an error when any of the names listed in the require map are invalid.

We particularly notice this happens for invalid casing (v2 requires the
full name to be downcased), so that's what we're testing for here.


feelepxyz (approved):
Nice one 👍

feelepxyz (approved):

PR #2971, opened 2021-01-12
npm: move native helpers to npm6 namespace

feelepxyz:
Moving npm native helpers to npm6 namespace in preparation for adding
npm7 support.

The pattern is slightly different from composer because of the way we
invoke the helper with a `package_manager:function` rpc syntax.


jurre (approved):

npm_and_yarn/helpers/package.json
```diff
@@ -12,7 +12,7 @@
     "@dependabot/yarn-lib": "^1.21.1",
     "@npmcli/arborist": "^2.0.3",
     "detect-indent": "^6.0.0",
-    "npm": "6.14.11",
+    "npm6": "npm:npm@6.14.11",
```
feelepxyz, line 15: @jurre just pushed this change to alias npm to `npm6`

feelepxyz:
This should be good to go, will merge and release tomorrow. Just did some dry-runs with existing updates and working as expected.

jurre (approved):

npm_and_yarn/helpers/package.json
jurre: Nice!

caugner:
@feelepxyz Folks over at https://github.com/dependabot/dependabot-script/issues/526 (@timbru31 to be specific) have suggested that this PR here has caused a regression, which I happen to be affected by as well.

By any chance, would you have the opportunity to look into this?

PR #2974, opened 2021-01-13
Added require "set" to utils.rb

### LINKED ISSUE #2972: Missing require "set" in utils.rb
I am getting a runtime error in the dependabot-core docker image because the `/common/lib/dependabot/utils.rb` file doesn't have a `require "set"` but is using the `Set` class in line 36. I am not fluent in ruby I'm not sure if this should be needed or if it is a problem with how the docker image is made.

JohannesEH:
Fixes #2972

### DIFF
--- a/common/lib/dependabot/utils.rb
+++ b/common/lib/dependabot/utils.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "set"
+
 # TODO: in due course, these "registries" should live in a wrapper gem, not
 #       dependabot-core.
 module Dependabot


jurre (approved):
Yep this makes sense, thanks! 👍 I guess `Set` is already required in the project where we run Dependabot @ GitHub, but that shouldn't be a requirement.

feelepxyz (approved):
LGTM - I'm not sure why this doesn't cause any issues for us currently but seems like the right thing to do here so 👍 on adding it.

@jurre do you know if ruby loads this for us somehow?

JohannesEH:
> 
> 
> Yep this makes sense, thanks! 👍 I guess `Set` is already required in the project where we run Dependabot @ GitHub (and in the dry-run script), but that shouldn't be a requirement.

Thanks that makes sense. I'm not really home in Ruby but I figured there might be a more "global" way of including the Set class.

jurre:
> @jurre do you know if ruby loads this for us somehow?

I think one of our dependencies requires it, probably `bundler`. (the dry-run script has it after `Bundler.setup`). There may be something funky in how @JohannesEH's is setting up their code, but either way it would be good to explicitly require `set` there

JohannesEH:
@jurre maybe I'm using the dependabot-core docker image incorrectly. I'm trying to add it to our Azure DevOps pipeline, so I found [a usage example](https://github.com/chris5287/dependabot-for-azuredevops/) by @chris5287 that integrates with Azure DevOps and based my own work on that. So I'm basically clueless. 😉

jurre:
> @jurre maybe I'm using the dependabot-core docker image incorrectly. I'm trying to add it to our Azure DevOps pipeline, so I found [a usage example](https://github.com/chris5287/dependabot-for-azuredevops/) by @chris5287 that integrates with Azure DevOps and based my own work on that. So I'm basically clueless. 😉

If it works otherwise you should be good 👍

PR #2978, opened 2021-01-13
Install npm 6 and npm 7 side by side

jurre:
We plan to use npm v7 in a future update. This adds that, just to make
sure that having them side by side doesn't introduce any problems.


feelepxyz (approved):
Assuming tests pass 👍

PR #2979, opened 2021-01-13
npm: detect npm 7 lockfiles

feelepxyz:
WIP: detect npm 7 lockfiles and log the result without using it.

### DIFF
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -61,6 +61,7 @@ Bundler.setup
 require "optparse"
 require "json"
 require "byebug"
+require "logger"
 
 require "dependabot/file_fetchers"
 require "dependabot/file_parsers"
@@ -84,6 +85,8 @@ require "dependabot/npm_and_yarn"
 require "dependabot/nuget"
 require "dependabot/python"
 require "dependabot/terraform"
+
+Dependabot.logger = Logger.new($stdout)
 
 # GitHub credentials with write permission to the repo you want to update
 # (so that you can create a new branch, commit and pull request).
--- a/common/lib/dependabot/logger.rb
+++ b/common/lib/dependabot/logger.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+require "logger"
+
+module Dependabot
+  def self.logger
+    @logger ||= Logger.new(nil)
+  end
+
+  def self.logger=(logger)
+    @logger = logger
+  end
+end
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
@@ -4,6 +4,7 @@ require "dependabot/npm_and_yarn/file_updater"
 require "dependabot/npm_and_yarn/file_parser"
 require "dependabot/npm_and_yarn/update_checker/registry_finder"
 require "dependabot/npm_and_yarn/native_helpers"
+require "dependabot/npm_and_yarn/helpers"
 require "dependabot/shared_helpers"
 require "dependabot/errors"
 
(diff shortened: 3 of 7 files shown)


npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -129,25 +131,29 @@ def run_previous_npm_update(lockfile_name:)
...
+              run_npm_subdependency_updater(lockfile_name: lockfile_name, lockfile_content: lockfile_content)
             end
           end
         end
 
-        def run_npm_top_level_updater(lockfile_name:,
-                                      top_level_dependency_updates:)
+        def run_npm_top_level_updater(lockfile_name:, top_level_dependency_updates:, lockfile_content:)
+          npm_version = Dependabot::NpmAndYarn::Helpers.npm_version(lockfile_content)
+          puts npm_version
```
feelepxyz, line 155: @jurre not entirely happy with this duplication but seems worth doing it in ruby so we can emit stats on the version when we add instrumentation. Was also thinking we could do the detection in JS land but would make instrumentation a lot more painful.

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/subdependency_version_resolver.rb
```diff
@@ -107,9 +108,12 @@ def run_yarn_updater(path, lockfile_name)
           sleep(rand(3.0..10.0)) && retry
         end
 
-        def run_npm_updater(path, lockfile_name)
+        def run_npm_updater(path, lockfile_name, lockfile_content)
           SharedHelpers.with_git_configured(credentials: credentials) do
             Dir.chdir(path) do
+              npm_version = Dependabot::NpmAndYarn::Helpers.npm_version(lockfile_content)
+              puts npm_version
```
jurre, line 115: Is there anything more sophisticated we can/should do other than just `put`? 😄 Maybe we could introduce the first tiny step for actual logging support and set up a `Dependabot.logger`, it could just be:

```ruby
class Dependabot
  def self.logger
    @logger ||= Logger.new(STDOUT)
  end
end
```

And then we could `Dependabot.logger.info npm_version` here

feelepxyz: Yeah great shout!

common/lib/dependabot/logger.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+require "logger"
+
+module Dependabot
+  def self.logger
+    @logger ||= Logger.new(STDOUT)
```
feelepxyz, line 7: @jurre thoughts on setting this to nil and the configuring in the updater? Not entirely sure about the best practice here.  Would you add a method like `register_logger`?

jurre: What about:

```ruby
def self.logger=(logger)
  @logger = logger
end
```

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
jurre: Aye, yeah it might be slick to do it in the native helpers, but I can't think of a good way to report on it if we do that 🤔

bin/dry-run.rb
```diff
@@ -85,6 +86,8 @@
 require "dependabot/python"
 require "dependabot/terraform"
 
+Dependabot.logger = Logger.new($stdout)
```
jurre, line 89: That's already the default right?

jurre (approved):
One nit, otherwise great work 🎉

bin/dry-run.rb
feelepxyz: Set it to a nil logger by default here so existing users of dependabot-core don't get logging output without asking  for it: https://github.com/dependabot/dependabot-core/pull/2979/files#diff-ff1db9597b04517023f14d0ada4d0cb69112beb6b8ea697088cd15d3ce07d2fbR7

PR #2982, opened 2021-01-14
Sanitize mentions more aggressively

jurre:
The current sanitization works well when directly creating PRs, however,
when a user replies to an email notification containing that content, some
of the markup of the original message will be stripped out (either by
GitHub, or by email providers), which can result in a link like 
`<a href="https://github.com/jurre>@jurre</a>` being turned into just
`@jurre`, which will consequently mention theuser when that email reply 
is rendered.

This wraps the inner mention in a codeblock, which should prevent this
from happening. It looks _slightly_ funny:

<a href="https://github.com/jurre"><code>@jurre</code></a>

But I think this is worth it, getting unexpected mentions by Dependabot
is a source of frustration for users, and by showing that the mention is
wrapped in a codeblock, this should make it easier to spot if it was an
actual mention or not.

Alternatively we could insert some no-space character between the `@`
and the user login, but I personally prefer this more explicit approach.

Fixes https://github.com/dependabot/dependabot-core/issues/2820

### DIFF
--- a/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
@@ -45,9 +45,13 @@ module Dependabot
 
         def sanitize_mentions(doc)
           doc.walk do |node|
-            if !parent_node_link?(node) && node.type == :text &&
+            if node.type == :text &&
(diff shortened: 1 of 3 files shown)


feelepxyz (approved):
Thanks for fixing this! I like this approach, makes it a lot easier to spot when dependabot has sanitised the mention. Injecting no-width spaces can be super confusing when copy-pasting and trying to search using it.

PR #2986, opened 2021-01-14
Generate project based fixtures for npm and yarn

jurre:
This adds a script to automatically generate project based fixtures for npm6 and yarn.

The lockfiles are all generated by running `npm install` and `yarn install`, this may not always match what we're currently doing in tests. This also won't cover every scenario, since we tend to mix and match manifest files and lockfiles, and in some cases we set the test up to run without a lockfile.

However, this should still avoid a lot of noise in PRs once we gradually start moving over tests to use this setup.


bin/generate-npm-project-based-fixtures.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  private
+
+  def build_npm_project(filename)
+    project_name = File.basename(filename, ".json")
+    project_dir = FileUtils.mkdir_p(File.join(npm6_project_dir, project_name)).last
+    FileUtils.copy(filename, File.join(project_dir, "package.json"))
+    Dir.chdir(project_dir) do
+      _, _, status = Open3.capture3("npm install")
```
feelepxyz, line 32: Oh slight concern, I know we have some test fixtures where the package.json wouldn't generate the same lockfile that's currently being tested. I've committed examples where I had to pin the version in `package.json` to get a particular lockfile but then needed the version to not be pinned in `package.json` so changed the requirement after the lockfile was generated 😬

This should result in broken tests where things have changed so should be ok to continue with this and fix up the broken tests when we encounter them. Seems like a good time to try and change the manifest to generate the correct lockfile from the get go where possible.

bin/generate-npm-project-based-fixtures.rb
```diff
@@ -0,0 +1,70 @@
...
+  def yarn_project_dir
+    "npm_and_yarn/spec/fixtures/projects/yarn"
+  end
+
+  def npm_package_files
+    Dir.glob("npm_and_yarn/spec/fixtures/package_files/*.json")
+  end
+end
+
+ProjectBasedFixtureGenerator.new.run
```
feelepxyz, line 70: Nice work adding this! Seems like a good starting point but we'll need to verify each test case when using the project and restore the old lockfile if we needed some weird state that's not been reproduced. There might also be some cases missing where we re-use the same package.json with different lockfiles?

bin/generate-npm-project-based-fixtures.rb
jurre: Yeah I touched on this slightly in the PR description, there will probably be more cases where this doesn't fully cover what we need it to, I think the only thing we can hope for with this PR is to reduce noise in future PRs where we actually update the test. We'll still need to verify that the projects are set up correctly for what we're trying to test.

feelepxyz: Yeah I reckon we should get this in and go from there 👍

feelepxyz (approved):
Lets get this in and fix things as we start using them specs 👍

jurre:
We've figured out a smarter way to do this, it's not perfect yet, but going to close this out for now

PR #2987, opened 2021-01-14
Move logger to a Logger module

feelepxyz:
Changing the logger implementation to match what we have in the updater
so we can replace it with this.

### DIFF
--- a/common/lib/dependabot/logger.rb
+++ b/common/lib/dependabot/logger.rb
@@ -3,11 +3,17 @@
 require "logger"
 
 module Dependabot
-  def self.logger
-    @logger ||= Logger.new(nil)
-  end
+  module Logger
+    def logger
+      Dependabot::Logger.logger
+    end
 
-  def self.logger=(logger)
-    @logger = logger
+    def self.logger
+      @logger ||= ::Logger.new(nil)
+    end
+
+    def self.logger=(logger)
+      @logger = logger
+    end
   end
 end
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -86,7 +86,7 @@ require "dependabot/nuget"
 require "dependabot/python"
 require "dependabot/terraform"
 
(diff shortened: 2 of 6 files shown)


common/lib/dependabot/logger.rb
```diff
@@ -3,11 +3,26 @@
...
+    end
+
+    def self.destination
+      @destination ||= nil
+    end
 
-  def self.logger=(logger)
-    @logger = logger
+    def self.destination=(destination)
+      @destination = destination
```
jurre, line 25: I know we're just moving this code around, but I'm just noticing now that we need to set the destination before calling the logger or it will noop, and I'm wondering if that might be confusing

feelepxyz: Ah yes good catch, previous design of setting a logger might have been more straight forward as we don't need the formatter here and stuff here

common/lib/dependabot/logger.rb
```diff
@@ -3,11 +3,17 @@
 require "logger"
 
 module Dependabot
-  def self.logger
-    @logger ||= Logger.new(nil)
-  end
+  module Logger
+    def logger
+      Dependabot::Logger.logger
```
feelepxyz, line 8: @jurre thoughts on this pattern? Feels a bit weird

jurre: Yeah that makes sense 👍  I think we could simplify even further by removing `def logger` and adding a `module_function :logger` below `def self.logger`.

jurre (approved):

common/lib/dependabot/logger.rb
feelepxyz: getting a exception trying that `undefined method `logger'` - seems you have to do `def logger ... module_function :logger` but that doesn't seem to work when setting the instance variable from the setter, always uses the nil logger

Issue #2416, opened 2020-08-04
error="waiting for updater: waiting for container: context deadline exceeded" job_id=44344904


ElijahLynn:
Getting this message at end of logs, please advise: 

```
updater | time="2020-08-04T00:30:13Z" level=info msg="task complete" container_id=job-44344904-updater exit_code=137 job_id=44344904 step=updater
updater | time="2020-08-04T00:30:13Z" level=warning msg="timeout running job" error="waiting for updater: waiting for container: context deadline exceeded" job_id=44344904
```

**Package manager/ecosystem**

composer
**Manifest contents prior to update**

**Updated dependency**

**What you expected to see, versus what you actually saw**

**Images of the diff or a link to the PR, issue or logs**
Expect logs to run successfully.

ElijahLynn:
I opened an Enterprise Github support ticket for this but didn't get a confirmation because I wasn't logged in, and am having a tough time even after logging in getting an email to verify my email, our mail server must be blocking it.

jurre:
Hi @ElijahLynn, it looks like the updates are legitimately timing out, we currently have a hard limit of 30 minutes for an update job, and composer is known to sometimes be on the slow side.

Right now I would recommend that you retry the failed jobs from the UI, and I will create an issue on our backlog to see if we can give some ecosystems a longer timeout.

ElijahLynn:
Thanks @jurre, I have seen Packagist.org (composer's default source) hang sometimes and wonder if that may be it. Can you expose more info on why it is timing out? 

fwiw, A local `composer update` on this repo (with cached packages) takes under 3 minutes. The timeout is hanging on a step just after getting some package metadata. I don't think this needs a longer timeout, I think there is a bug here and something needs to be caught and bubbled up. 

proxy | 2020/08/08 00:30:10 [154] GET http://repo.packagist.org/p/provider-latest%24f2867f0b89850775aba29f04766de27c2ab9d10e9925545e0dbbebdd3185f1c1.json
  proxy | 2020/08/08 00:30:10 [154] 200 http://repo.packagist.org/p/provider-latest%24f2867f0b89850775aba29f04766de27c2ab9d10e9925545e0dbbebdd3185f1c1.json
**updater | time="2020-08-08T00:30:29Z" level=info msg="task complete" container_id=job-44795731-updater exit_code=137 job_id=44795731 step=updater
[...]

PR #2998, opened 2021-01-17
Prevent unnecessary downloads of library packages - just update the json/lock files

driskell:
I was investigating slow dependabot running on some Drupal projects and came across issue #2416 
Several things became clear as the cause, specifically around Drupal projects. One was that dependabot is slowed down by downloading actual packages when it does not need to.

This PR targets that by introducing a replacement for the `LibraryInstaller` that does not download or install anything.

Overall it shaves a few seconds off some of my local runs but might be worthwhile just for the bandwidth reductions.


driskell:
Referencing #2999

jurre:
Looking forward to giving this a spin! Thanks so much for digging into this @driskell

driskell:
@jurre No worries - I had set it as a personal challenge to dig into what's happening!
It's most noticeable if I remember on the patching - so once it finds the update, and then applies to get the lock file - that stage is noticeably faster. However, it may still be slow for things like Drupal where we'd be waiting on if Composer think my patches are acceptable, in which case once Dependabot receives those updates to Composer it should fix the main issue.

stof:
Composer v2 has a `composer update --no-install` feature, which allows to update the lock file without running the installation of new deps in the vendor folder. Configuring the code to run in this mode would be cleaner than replacing the LibraryInstaller IMO (and less code to maintain)

jurre:
> Composer v2 has a `composer update --no-install` feature, which allows to update the lock file without running the installation of new deps in the vendor folder. Configuring the code to run in this mode would be cleaner than replacing the LibraryInstaller IMO (and less code to maintain)

That's a good suggestion. I traced that setting back to setting `$installer->setInstall(false)`, but I get a bunch of test failures when doing that. I don't have time to spend a lot of cycles on this right now, but if there's something else we'd need to configure that you know of I'd love to hear!

stof:
> That's a good suggestion. I traced that setting back to setting `$installer->setInstall(false)`, but I get a bunch of test failures when doing that. I don't have time to spend a lot of cycles on this right now, but if there's something else we'd need to configure that you know of I'd love to hear!

I don't think there is anything else needed to configure that. And without seeing which test failure you have, I cannot give you an idea about a solution (note that even when seeing them, I don't guarantee that I will have an idea, given that I don't know the dependabot codebase, only the composer one)

driskell:
> > That's a good suggestion. I traced that setting back to setting `$installer->setInstall(false)`, but I get a bunch of test failures when doing that. I don't have time to spend a lot of cycles on this right now, but if there's something else we'd need to configure that you know of I'd love to hear!
> 
> I don't think there is anything else needed to configure that. And without seeing which test failure you have, I cannot give you an idea about a solution (note that even when seeing them, I don't guarantee that I will have an idea, given that I don't know the dependabot codebase, only the composer one)

Nice, that would definitely mean this PR is not needed. I did wonder though when I did this one if installers need to be installed still and loaded in case they impact the output. Would setInstall stop those installer plugins from being installed and loaded?

stof:
Custom installer can change the location of the the installed package on disk. But they cannot change the version being selected (and so cannot affect the lock file).

The only plugins that can affect the lock file are the ones which are hooking into the dependency solver itself (using the `pre-pool-create` event), like `symfony/flex` for instance. And for that to work during  a `composer update`, the plugin must be installed *before* running the update (so either as a global plugin, or by running a `composer install` first to have a vendor folder).

driskell:
@jurre Was the approach you suggested merged as the PR is showing as closed? 

I am now having issues in all my repositories where security updates are no longer appearing because dependabot simply can’t scan the repository in time anymore even though I’ve effectively set it to only check 3 packages. Investigating the logs from GitHub I can see it is literally downloading the ZIP for many packages each time and never gets to complete the second package... so for some reason something has made things a lot worse. I think it takes now 5 days or more to even get the updates for the 3 packages which isn’t sustainable.

I also worry how much CO2 this is burning unnecessarily it’s almost like mining bitcoins lol.

If there was an issue with the approach it be good to know as I’m keen to help get this working. The optimisation PRs at composer that are needed for extremely quick scans with little CO2 are stuck or might need some test attention (composer/composer#9620 composer/composer#9619 composer/composer#9261) but hopefully moving along as the maintainers find time but this issue here looks to me like a good win!

jurre:
> Was the approach you suggested merged as the PR is showing as closed

No, I closed it as I ran into a bunch of different test failures that I didn't really have time to investigate. The logs for those have since been garbage collected by Actions, so I've re-opened the PR temporarily to regenerate the CI failures, but I will close the PR again after that, as I'm trying to cut down pending PRs that I'm not actively working on to keep myself sane.

I'm surprised things have gotten that much worse, as nothing has really changed on our end 🤔 

It would be great if we could leverage changes to composer like the PR you linked to instead of building something bespoke for Dependabot, but if we can get the config flag working that seems reasonable to me.

mpdude:
Trying to migrate from `dependabot-preview` to `dependabot` for a whole bunch of PHP projects, I noticed that updates currently fail for me with the message

> Dependabot can't access your-org/your-repo
> You can fix the issue by granting Dependabot access, which will enable any repository in @your-org to get automatic updates from your-org/your-repo. Be sure that you want to share your-org/your-repo with everyone in your organization.

I am using a private package registry, but that seems to work find and access is granted (I configured it accordingly). Now, to my knowledge, Composer obtains all the information it needs from the registries, so I wonder why the above error occurs.

Am I right assuming that if the change suggested here were made, it would no longer be necessary to configure the additional permissions in https://github.com/organizations/my-org/settings/security_analysis, under "Grant Dependabot access to private repositories"?

jurre:
@mpdude that seems like a different issue, it appears that you're pulling in a git dependency and Dependabot cannot access it.

I don't think that these changes would affect access to private sources, Dependabot would still need to access them to find version info etc.

Could you open up a new issue for this and fill out the questions in the template? I'll be more than happy to help you sort it out 👍

mpdude:
@jurre Tried my very best over at #3628.

markdorison:
Is this PR still being considered? We are experiencing the issues described in https://github.com/dependabot/dependabot-core/issues/2416 & https://github.com/dependabot/dependabot-core/issues/2999, and while there are also some issues/PRs against Composer (https://github.com/composer/composer/pull/9261, https://github.com/composer/composer/pull/9620), I have not seen any recent traction on those. I know the sentiment seemed to be that if it could be fixed in Composer that would be ideal, but if this could be resolved here with little downside, I would be the first to cheer that on.

driskell:
> > Was the approach you suggested merged as the PR is showing as closed
> 
> No, I closed it as I ran into a bunch of different test failures that I didn't really have time to investigate. The logs for those have since been garbage collected by Actions, so I've re-opened the PR temporarily to regenerate the CI failures, but I will close the PR again after that, as I'm trying to cut down pending PRs that I'm not actively working on to keep myself sane.
> 
> I'm surprised things have gotten that much worse, as nothing has really changed on our end 🤔
> 
> It would be great if we could leverage changes to composer like the PR you linked to instead of building something bespoke for Dependabot, but if we can get the config flag working that seems reasonable to me.

I just tested the setInstall locally to take a look. Using `setInstall(false)` disables installation completely - it won't even hit the InstallationManager. So the `get_latest_resolvable_version` request and likely others will never find the packages as it relies on the InstallationManager to return the list of installed packages. I can think also this won't install any installer packages that might impact the output file.

driskell:
I fixed up the PHPCS for this and used NoopInstaller instead. I also tried to fix up tests but it warrants a bit of discussion - I've mostly repurposed tests for reporting nice errors into tests to ensure no error happens since those tests begin to fail because no attempt is made to download or install and so bad commits and references and variables aren't relevant.

driskell:
> Custom installer can change the location of the the installed package on disk. But they cannot change the version being selected (and so cannot affect the lock file).
> 
> The only plugins that can affect the lock file are the ones which are hooking into the dependency solver itself (using the `pre-pool-create` event), like `symfony/flex` for instance. And for that to work during a `composer update`, the plugin must be installed _before_ running the update (so either as a global plugin, or by running a `composer install` first to have a vendor folder).

Sorry I just reread this and took a further look into setInstall() - and ended up just removing the entire custom installation manager (DependabotInstallationManager) - since it occurred to me it's not really the ideal way to get Composer's list of installed packages - we can instead get it from the lock data by using the locker to get the locked packages repository and get the packages there. That is not quite "what did we install just now" but it definitely is the "what is installed overall" and should be equal if not richer to what was currently being used (since it likely the installation manager was receiving an install operation for every package anyway in the dependabot case...)

Looks like all tests still pass!

composer/helpers/v2/src/UpdateChecker.php
```diff
@@ -86,7 +77,7 @@ public static function getLatestResolvableVersion(array $args): ?string
 
         $install->run();
 
-        $installedPackages = $installationManager->getInstalledPackages();
+        $installedPackages = $composer->getLocker()->getLockedRepository(true)->getPackages();
```
driskell, line 80: This argument is "withDevReqs" to return all packages including dev requirements.

composer/spec/dependabot/composer/file_updater/lockfile_updater_spec.rb
```diff
@@ -609,12 +574,8 @@
         )
       end
 
-      it "raises a helpful errors" do
-        expect { updated_lockfile_content }.to raise_error do |error|
-          expect(error).to be_a Dependabot::GitDependencyReferenceNotFound
```
driskell, line 614: Do these exceptions potentially need cleaning up? I left as is as it seems it will need a higher discussion as this looks like the scope now increases as there’s potentially areas of code handing nice errors that won’t need handling anymore because installs aren’t attempted

driskell:
@jurre @stof Found an issue when using setInstall(false) with https://github.com/wikimedia/composer-merge-plugin

That plugin adds additional install requirements based on merges of composer.json from other installed packages. Thus I don't think you can not allow it to be installed, and potentially even you cannot disallow installation of libraries as any one could provide a merge...

In my test runs on one repository I have (I can't share) then during updates dependabot would start to remove the additional requirements provided by the merge plugin....

driskell:
I think just doing this in the UpdateChecker would be fine, as long as the Updater still installs. That would still give a speed benefit for where update not needed.

I accidentally force pushed now I can't reopen heh. Will raise separately once tested.

driskell:
New PR #4635

PR #3004, opened 2021-01-18
composer: fix check for platform deps in composer version detection

voda:
Hi,
after https://github.com/dependabot/dependabot-core/pull/2970 Dependabot PRs reverted to using composer v1 in my repo. The reason is that I have `"ext-zip": "*"`and similar in the `require` section.

I tested this by checking the value returned by `composer_version`.

### DIFF
--- a/composer/lib/dependabot/composer/helpers.rb
+++ b/composer/lib/dependabot/composer/helpers.rb
@@ -7,6 +7,11 @@ module Dependabot
     module Helpers
       # From composers json-schema: https://getcomposer.org/schema.json
       COMPOSER_V2_NAME_REGEX = %r{^[a-z0-9]([_.-]?[a-z0-9]+)*/[a-z0-9](([_.]?|-{0,2})[a-z0-9]+)*$}.freeze
+      # From https://github.com/composer/composer/blob/b7d770659b4e3ef21423bd67ade935572913a4c1/src/Composer/Repository/PlatformRepository.php#L33
+      PLATFORM_PACKAGE_REGEX = /
+        ^(?:php(?:-64bit|-ipv6|-zts|-debug)?|hhvm|(?:ext|lib)-[a-z0-9](?:[_.-]?[a-z0-9]+)*
+        |composer-(?:plugin|runtime)-api)$
+      /x.freeze
 
       def self.composer_version(composer_json, parsed_lockfile = nil)
         return "v1" if composer_json["name"] && composer_json["name"] !~ COMPOSER_V2_NAME_REGEX
@@ -21,7 +26,7 @@ module Dependabot
         return false unless composer_json.key?("require")
 
         composer_json["require"].keys.any? do |key|
-          key != "php" && key !~ COMPOSER_V2_NAME_REGEX
+          key != "php" && key !~ PLATFORM_PACKAGE_REGEX && key !~ COMPOSER_V2_NAME_REGEX
         end
       end
       private_class_method :invalid_v2_requirement?
--- a/composer/lib/dependabot/composer/helpers.rb
+++ b/composer/lib/dependabot/composer/helpers.rb
@@ -14,19 +14,22 @@ module Dependabot
       /x.freeze
 
       def self.composer_version(composer_json, parsed_lockfile = nil)
-        return "v1" if composer_json["name"] && composer_json["name"] !~ COMPOSER_V2_NAME_REGEX
-        return "v1" if invalid_v2_requirement?(composer_json)
-        return "v2" unless parsed_lockfile && parsed_lockfile["plugin-api-version"]
+        if parsed_lockfile && parsed_lockfile["plugin-api-version"]
+          version = Composer::Version.new(parsed_lockfile["plugin-api-version"])
+          return version.canonical_segments.first == 1 ? "v1" : "v2"
+        else
+          return "v1" if composer_json["name"] && composer_json["name"] !~ COMPOSER_V2_NAME_REGEX
+          return "v1" if invalid_v2_requirement?(composer_json)
+        end
 
-        version = Composer::Version.new(parsed_lockfile["plugin-api-version"])
-        version.canonical_segments.first == 1 ? "v1" : "v2"
+        "v2"
       end
 
       def self.invalid_v2_requirement?(composer_json)
         return false unless composer_json.key?("require")
 
         composer_json["require"].keys.any? do |key|
-          key != "php" && key !~ PLATFORM_PACKAGE_REGEX && key !~ COMPOSER_V2_NAME_REGEX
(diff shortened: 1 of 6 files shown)


composer/lib/dependabot/composer/helpers.rb
```diff
@@ -21,7 +26,7 @@ def self.invalid_v2_requirement?(composer_json)
         return false unless composer_json.key?("require")
 
         composer_json["require"].keys.any? do |key|
-          key != "php" && key !~ COMPOSER_V2_NAME_REGEX
+          key != "php" && key !~ PLATFORM_PACKAGE_REGEX && key !~ COMPOSER_V2_NAME_REGEX
```
jurre, line 29: We can remove the hard coded check for php, it's already in the regex 🎉

jurre (commented):
Thanks, this makes a lot of sense. I'll add a test to cover the scenario you described and get rid of the hard coded php check in a bit and try to roll this out today. 

On a side-note, I'm slightly worried about this version check growing further in complexity. We might want to consider making it configurable if it turns out there are even more edge cases. I'd prefer Dependabot to just do the right thing without having to configure anything, but if it doesn't work well configuring seems like a good option to have.

voda:
> On a side-note, I'm slightly worried about this version check growing further in complexity. We might want to consider making it configurable if it turns out there are even more edge cases. I'd prefer Dependabot to just do the right thing without having to configure anything, but if it doesn't work well configuring seems like a good option to have.

Maybe giving the `plugin-api-version` check a higher priority could also help. If the lockfile already has v2, then don't do the name checks, but directly return v2 from the `composer_version` function.

And the checks would be only to test if v2 can be used instead in cases where v1 was used previously.

jurre:
Yeah that's a good point, although if those checks fail, composer 2 will error so the update will definitely fail right?

voda:
The `"plugin-api-version": "2.0.0"` will be set the previous install/update was already done with composer v2, so I see no reason why it should fail on a later run.

jurre:
> The `"plugin-api-version": "2.0.0"` will be set the previous install/update was already done with composer v2, so I see no reason why it should fail on a later run.

Only if you're using a plugin, unfortunately not every project (especially libraries I think?) does

voda:
> Only if you're using a plugin

I think "plugin-api-version" is always present in the .lock file. I tested this with `require psr/log`. That should take care of repos with the .lock file present.

Not sure about the libraries/projects without a .lock file.

composer/spec/dependabot/composer/update_checker/version_resolver_spec.rb
```diff
@@ -201,6 +201,15 @@
       it { is_expected.to eq(Dependabot::Composer::Version.new("1.25.1")) }
     end
 
+    context "with a project that specifies a platform package" do
```
voda, line 204: I am not sure about this test. When I hardcode the return value in the Helpers.composer_version either v1 or v2 the test passes. Shouldn't it verify that the correct composer version is used?

jurre: You're right, thanks! I've added two different tests to make sure the new behavior is covered

composer/lib/dependabot/composer/helpers.rb
```diff
@@ -7,21 +7,29 @@ module Composer
...
+          return "v1" if composer_json["name"] && composer_json["name"] !~ COMPOSER_V2_NAME_REGEX
+          return "v1" if invalid_v2_requirement?(composer_json)
+        end
 
-        version = Composer::Version.new(parsed_lockfile["plugin-api-version"])
-        version.canonical_segments.first == 1 ? "v1" : "v2"
+        "v2"
       end
 
       def self.invalid_v2_requirement?(composer_json)
```
feelepxyz, line 28: Thoughts on adding a comment here to clarify the rules?

```suggestion
      # NOTE: Composer v2 validates `requires` keys to have the format `vendor/package-name` but
      # allows certain "platform" specific keys, starting with `php-`, `ext-` or `lib-`
      def self.invalid_v2_requirement?(composer_json)
```

composer/lib/dependabot/composer/helpers.rb
```diff
@@ -7,21 +7,29 @@ module Composer
...
+      PLATFORM_PACKAGE_REGEX = /
+        ^(?:php(?:-64bit|-ipv6|-zts|-debug)?|hhvm|(?:ext|lib)-[a-z0-9](?:[_.-]?[a-z0-9]+)*
+        |composer-(?:plugin|runtime)-api)$
+      /x.freeze
 
       def self.composer_version(composer_json, parsed_lockfile = nil)
-        return "v1" if composer_json["name"] && composer_json["name"] !~ COMPOSER_V2_NAME_REGEX
-        return "v1" if invalid_v2_requirement?(composer_json)
-        return "v2" unless parsed_lockfile && parsed_lockfile["plugin-api-version"]
+        if parsed_lockfile && parsed_lockfile["plugin-api-version"]
```
feelepxyz, line 17: 👍

feelepxyz (approved):
LGTM 👍  added a minor comment about clarifying the intention of `invalid_v2_requirement`

composer/lib/dependabot/composer/helpers.rb
```diff
@@ -7,21 +7,29 @@ module Composer
     module Helpers
       # From composers json-schema: https://getcomposer.org/schema.json
       COMPOSER_V2_NAME_REGEX = %r{^[a-z0-9]([_.-]?[a-z0-9]+)*/[a-z0-9](([_.]?|-{0,2})[a-z0-9]+)*$}.freeze
+      # From https://github.com/composer/composer/blob/b7d770659b4e3ef21423bd67ade935572913a4c1/src/Composer/Repository/PlatformRepository.php#L33
```
feelepxyz, line 10: 👍

composer/spec/fixtures/projects/platform_package/composer.json
```diff
@@ -0,0 +1,13 @@
...
+  "name": "valid/name",
+  "version": "3.1.7",
+  "dist": {
+    "url": "https://www.example.net/files/3.1.7.zip",
+    "type": "zip"
+  },
+  "require": {
+    "monolog/monolog" : "1.0.1",
+    "symfony/polyfill-mbstring": "1.0.1",
+    "php": "*"
```
lstrojny, line 11: How about requiring some more platform packages to exhaust the different options like `lib-*`, `ext-*` and so on?

jurre: The regex is already tested thoroughly in composer itself and we short circuit the logic most of the time by checking the lockfile, but I've added a bit more coverage in the helper tests

PR #3011, opened 2021-01-19
Add support for npm 7 lockfiles

feelepxyz:
Add support for updating npm 7 lockfiles alongside npm 6 based on the `lockfileVersion` set in the `package-lock.json` file. If this is set to `2` we used npm 7 otherwise fallback to npm 6.

The native helpers to update dependencies shell out to `npm install pkg-name` for direct/top-level and `npm update pkg-name` for indirect/sub-dependencies.

We made an initial attempt to use arborist programatically to do the update but abandoned this approach as we ran into a bug where [arborist](https://github.com/npm/arborist/) depends on [pacote](https://github.com/npm/pacote) which re-runs what it thinks is the main npm cli command on failures to retry, which in our setup lead to an infinite loop of retries. The bug was reproduced here: https://github.com/feelepxyz/npm7-hang-investigation and discussed with the npm team who advised us to no use arborist programatically in this way (yet).

## Review checklist/notable changes

- [x] Rebase on top of https://github.com/dependabot/dependabot-core/pull/3086
- [x] Review relevant files: https://github.com/dependabot/dependabot-core/pull/3011/files?file-filters%5B%5D=.js&file-filters%5B%5D=.rb&file-filters%5B%5D=dotfile
- [x] Changes to call out
  - [x] npm_and_yarn/lib/dependabot/npm_and_yarn/dependency_files_filterer.rb
    - Changes required to support npm 7 workspaces
  - [x] npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
    - Handle new error output
      - npm 7 now exits with a full error log so we can't do `starts_with?` on anything meaningful
    - handle updating the root lockfile when using workspaces in `top_level_dependency_update_not_required?`
    - Handle locked git dependencies, when updating `pkg-a` we go through and find any top-level git dependencies and lock them to the current resolved sha in `package.json` and then re-set the lockfile after update
    - Handle npm 7 lockfile `packages."".*dependencies` entries in `restore_locked_package_dependencies` which should mirror what's in your `package.json`, we need to reset this after update otherwise we'll have both locked git deps and updated dependencies new requirements in there even though we didn't update the `package.json`
      - If your package.json says `eslint: ^1.0.0` but we want to update to `1.1.1` without updating the package.json we run `npm install eslint@1.1.0` which updates the `packages."".*dependencies` entry to `1.1.0` but we want it to match the `package.json` so reset it to `^1.0.0` after update
  - `git_dependencies_to_lock` git dependencies `from` syntax in lockfiles has changed to include the dependency name and also always seem to default to `ssh://` protocol when not explicitly fetching from `https://github..` for example
    - It seems like a new default as it changes when converting a npm6 lockfile to npm 7 with `npm i --package-lock-only`

## Release plan

- [x] Review ⏫ 
- [x] Dry run existing npm 6 updates
- [x] Dry run npm 7 projects from around github, how to search for `"\"lockfileVersion\": 2"` with quotes?


npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -444,9 +460,14 @@ def git_dependencies_to_lock
               next unless details["version"]
               next unless details["version"].start_with?("git")
 
+              # npm 7 metadata
+              manifest_requirement = parsed_lockfile.dig("packages", "", "dependencies", nm)
```
feelepxyz, line 464: TODO: Handle other types, e.g. `devDependencies`

npm_and_yarn/helpers/lib/npm7/updater.js
```diff
@@ -0,0 +1,110 @@
...
+    // - `--dry-run=false` the updater sets a global .npmrc with dry-run: true to
+    //   work around an issue in npm 6, we don't want that here
+    // - `--force` ignores checks for platform (os, cpu) and engines
+    // - `--ignore-scripts` disables prepare and prepack scripts which are run
+    //   when installing git dependencies
+    await execa(
+      "npm",
+      [
+        "install",
+        ...args,
```
feelepxyz, line 45: TODO: escape these arguments

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -222,7 +232,7 @@ def handle_npm_updater_error(error, lockfile)
           # This happens if a new version has been published but npm is having
           # consistency issues and the version isn't fully available on all
           # queries
-          if error_message.start_with?("No matching vers") &&
+          if error_message.include?("No matching vers") &&
```
feelepxyz, line 235: Why don't we match on the full string here? `No matching version`

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -502,46 +521,104 @@ def git_ssh_requirements_to_swap
           @git_ssh_requirements_to_swap
         end
 
-        def post_process_npm_lockfile(original_content, updated_content)
-          updated_content =
-            replace_project_metadata(updated_content, original_content)
+        def post_process_npm_lockfile(original_content, updated_content, lockfile_name)
```
jurre, line 524: As discussed in pair-review: this might be a good candidate to extract out into its own class

npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater_spec.rb
```diff
@@ -197,202 +142,258 @@
...
+        let(:files) { project_dependency_files("#{npm_version}/nonexistent_dependency_yanked_version") }
+
+        it "raises a helpful error" do
+          expect { updated_npm_lock_content }.
+            to raise_error(Dependabot::PrivateSourceAuthenticationFailure)
+        end
       end
-    end
 
-    context "when a git src dependency doesn't have a valid package.json" do
```
feelepxyz, line 348: This was only removed from npm 7 as it takes absolutely ages and actually manages to do the update, the spec exists here for npm 6: https://github.com/dependabot/dependabot-core/blob/bfbc137cddb2da65f0e37f6ee4fae8c6dee75067/npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater_spec.rb#L79

npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater_spec.rb
```diff
@@ -197,202 +142,258 @@
...
-        expect { updated_npm_lock_content }.
-          to raise_error(Dependabot::DependencyFileNotResolvable)
-      end
-    end
+        it "has details of the updated item" do
+          parsed_lockfile = JSON.parse(updated_npm_lock_content)
+          expect(parsed_lockfile["dependencies"]["fetch-factory"]["version"]).
+            to eq("0.0.2")
 
-    context "with a corrupted npm lockfile (version missing)" do
```
feelepxyz, line 246: No longer failing in npm 7, npm 6 spec is here https://github.com/dependabot/dependabot-core/blob/bfbc137cddb2da65f0e37f6ee4fae8c6dee75067/npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater_spec.rb#L64

npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater_spec.rb
```diff
@@ -127,63 +127,8 @@
...
-      let(:npm_lock_fixture_name) { "package-lock.json" }
-
-      it "raises a helpful error" do
-        expect { updated_npm_lock_content }.
-          to raise_error(Dependabot::DependencyFileNotResolvable)
-      end
-    end
 
+    # NOTE: This no longer raises in npm 7
     context "when there is a private git dep we don't have access to" do
```
feelepxyz, line 187: No longer fails for npm 6, tests have moved here: https://github.com/dependabot/dependabot-core/blob/bfbc137cddb2da65f0e37f6ee4fae8c6dee75067/npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater_spec.rb#L131-L170

npm_and_yarn/lib/dependabot/npm_and_yarn/dependency_files_filterer.rb
```diff
@@ -51,8 +51,8 @@ def package_required_lockfile?(lockfile)
         end
       end
 
-      def yarn_workspaces_lockfile?(lockfile)
-        return false unless lockfile.name == "yarn.lock"
+      def workspaces_lockfile?(lockfile)
+        return false unless ["yarn.lock", "package-lock.json"].include?(lockfile.name)
```
feelepxyz, line 55: Note: support npm 7 workspace projects

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -153,28 +165,127 @@ def run_npm_updater(lockfile_name:, top_level_dependency_updates:, lockfile_cont
...
+            "npm",
+            "install",
+            *args,
+            "--force",
+            "--dry-run",
+            "false",
+            "--ignore-scripts",
+            "--package-lock-only"
+          ].join(" ")
+          SharedHelpers.run_shell_command(command)
```
feelepxyz, line 204: Note: command gets escaped by shellwords https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/shared_helpers.rb#L270

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -198,7 +309,7 @@ def handle_npm_updater_error(error, lockfile)
           # workspace project)
           sub_dep_local_path_error = "does not contain a package.json file"
           if error_message.match?(INVALID_PACKAGE) ||
-             error_message.start_with?("Invalid package name") ||
+             error_message.include?("Invalid package name") ||
```
feelepxyz, line 312: Note: npm 7 now exits with a full error log so we can't do starts_with? on anything meaningful

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
feelepxyz: Yeah agreed, there's a bunch of pre/post-processing that's intertwined and would probably be a lot easier to grok if it was in one place.

npm_and_yarn/lib/dependabot/npm_and_yarn/dependency_files_filterer.rb
jurre: To clarify, should we check to see if there are any workspaces defined in the package.json for npm7?

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -153,28 +165,127 @@ def run_npm_updater(lockfile_name:, top_level_dependency_updates:, lockfile_cont
...
+          SharedHelpers.run_shell_command(command)
+          { lockfile_name => File.read(lockfile_name) }
+        end
+
+        # TODO: Update the npm 6 updater to use these args as we currently do
+        # the same in the js updater helper, we've kept it seperate for the npm
+        # 7 rollout
+        def npm_top_level_updater_args(dependencies)
+          package_json_content = File.read("package.json")
+          manifest = JSON.parse(package_json_content)
```
jurre, line 248: Can we not grab this from `#updated_package_json_content_for_lockfile_name`? It does some memoization, maybe we explicitly don't want that here?

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -153,28 +165,127 @@ def run_npm_updater(lockfile_name:, top_level_dependency_updates:, lockfile_cont
...
+          else
+            SharedHelpers.run_helper_subprocess(
+              command: NativeHelpers.helper_path,
+              function: "npm6:updateSubdependency",
+              args: [Dir.pwd, lockfile_name, sub_dependencies.map(&:to_h)]
+            )
+          end
+        end
+
+        def run_npm_7_subdependency_updater(lockfile_name:)
```
jurre, line 222: Mentioned while pairing: might want to extract a separate `Npm7LockfileUpdater` in a follow-up PR.

npm_and_yarn/lib/dependabot/npm_and_yarn/dependency_files_filterer.rb
feelepxyz: Yep should already be covered by the line bellow 👇 `return false unless parsed_root_package_json["workspaces"]`

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
feelepxyz: Yeah good shout, updating

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -502,46 +631,109 @@ def git_ssh_requirements_to_swap
...
+        # requirements and the `packages.""` entry in npm 7 v2 lockfiles. These
+        # get out of sync because we lock git dependencies (that are not being
+        # updated) to a specific sha to prevent unrelated updates and the way we
+        # invoke the `npm install` cli, where we might tell npm to install a
+        # specific versionm e.g. `npm install eslint@1.1.8` but we keep the
+        # `package.json` requirement for eslint at `^1.0.0`, in which case we
+        # need to copy this from the manifest to the lockfile after the update
+        # has finished.
+        def restore_locked_package_dependencies(lockfile_name, lockfile_content)
+          npm_version = Dependabot::NpmAndYarn::Helpers.npm_version(lockfile_content)
```
jurre, line 662: We check for the `npm_version` 6 times in this file, I know we talked about extracting an npm 7 updater, but maybe we can start with:

```ruby
def npm7?
  return @npm7 if defined?(@npm7)

  @npm7 = Dependabot::NpmAndYarn::Helpers.npm_version(lockfile_content)
end
```

And using that everywhere? Might be a micro-optimization but I think it slightly clarifies the code and should be a pretty cheap change?

jurre (approved):
I think we should keep iterating on the code, but functionally I feel pretty confident about this

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
feelepxyz: Yeah this would be good but I think we need to refactor the lockfile updater a bit to get there as we currently pass around the lockfile name everywhere from here: https://github.com/dependabot/dependabot-core/blob/720e663fa95b952421f0883cdefe2f3048a6e797/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb#L26

Initialising the class with a single lockfile to be updated would make the class a lot simpler.

feelepxyz: Extracted it into a method but still passing around the content

feelepxyz:
Investigating python failures on main, seems unrelated to any recent changes

PR #3016, opened 2021-01-19
Fix JS debugging in vscode

feelepxyz:
This adds a couple launch scripts to debug npm and yarn tests when
running then using jest.

The launch scripts complained about a missing config file so added this
and changed the default env from jsdom to node as this is what we use.

Changed `eslint-plugin-prettier` to `eslint-config-prettier` as this
seems like the recommended extension now when using prettier with
eslint: https://prettier.io/docs/en/integrating-with-linters.html

This extracts these changes from the npm 7 branch:
#3011

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -15,7 +15,55 @@
         "${input:package_manager}",
         "${input:repository}"
       ]
-    }
+    },
+    {
+      "type": "node",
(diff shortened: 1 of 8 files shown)


jurre (approved):
Cheers for pulling this out into its own PR

Worker36ruS (approved):

Issue #3018, opened 2021-01-19
platform dependencies lead to wrong composer version being used


kasn:
**Package manager/ecosystem**
composer
**Manifest contents prior to update**
n/a
**Updated dependency**
n/a
**What you expected to see, versus what you actually saw**
correct composer version is used
**Images of the diff or a link to the PR, issue or logs**
n/a

Composer supports requiring [platform packages](https://getcomposer.org/doc/01-basic-usage.md#platform-packages) to verify the state of the system during install.

As per the example 
```
{
    "require" : {
        "php" : ">=7.4",
        "ext-mbstring": "*",
        "lib-libxml": "*"
    }
}
``` 
is valid. 

`Dependabot.Composer.Helpers.invalid_v2_requirement` checks all package names against the regex for valid composer v2 package names or if the name is php. 
Any required platform packages will cause dependabot to default to composer v1.

kasn:
I've implemented a fix in https://github.com/kasn/dependabot-core/commit/c233e52ae0555529d83f1312261263ddc7d8e7ee but my ruby/rspec experience is to limited to write meaningful tests. If anyone could point me into the right direction I would appreciate it.

kasn:
fixed with https://github.com/dependabot/dependabot-core/pull/3004

PR #3021, opened 2021-01-20
WIP #3018 allow ext-* and lib-* dependencies for composer v2

kasn:
this PR allows `ext-*` and `lib-*` system package requirements for composer v2


jurre:
This should be covered in https://github.com/dependabot/dependabot-core/pull/3004 👍 going to close this out in favor of that one

PR #3027, opened 2021-01-21
Automatically generate project based fixtures and updates tests

jurre:
This PR attempts to automatically convert a given test to use project based fixtures.

It roughly works by:
- Running the actual test, and hooking into the `let` definitions called in that test
- Comparing the `let` with a known list of fixture names
- Grabbing the files that we load from disk in those fixtures
- Create a folder using the original fixture names that were used and placing the files we grabbed earlier in there
- Parsing the tests AST and remove the offending `let` statements, and replace them with a single project based fixture setup

The code is a bit of a hot mess, but I've ran it on a few tests and it works well! 🎉


npm_and_yarn/spec/spec_helper.rb
```diff
@@ -9,3 +9,54 @@ def require_common_spec(path)
...
+
+  node.children.each do |child|
+    next unless child.is_a?(::Parser::AST::Node)
+
+    walk(child, &block)
+  end
+end
+
+RSpec.configure do |c|
+  c.after do
```
jurre, line 28: Doing this after each test is probably a little slower than it needs to be, we can run this in an `after(:all)` block instead, since we store the results per test. We can get get rid of storing the `current_group` also, since it's no longer used.

jurre:
I'd recommend keeping this as a branch and not merging it. We do nasty things to hook into `rspec` etc that I don't think should be in main

lastthyme (approved):

jurre:
This is a useful hack, but we're never going to merge this

PR #3032, opened 2021-01-24
Gradle Kotlin DSL: Add Support for Named URL Parameter in Maven Repository

hfhbd:
In Kotlin, a function call can have named parameters and Strings are only allowed to begin with `"`.

````kotlin
repositories {
    maven("foo.com")
    maven(url = "bar.com") // equals to maven("bar.com")
    maven('baz.com') // invalid, does not compile
}
````

This PR adds the optional `url = ` String to the regex as well as removing `'` as String delimiter, which is not allowed in Kotlin.

### DIFF
--- a/gradle/spec/fixtures/buildfiles/root_build.gradle.kts
+++ b/gradle/spec/fixtures/buildfiles/root_build.gradle.kts
@@ -16,7 +16,7 @@ buildscript {
 
     repositories {
         jcenter()
-        maven("https://dl.bintray.com/magnusja/maven")
+        maven(url = "https://dl.bintray.com/magnusja/maven")
         google()
     }
     dependencies {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: gradle/spec/fixtures/buildfiles/root_build.gradle.kts [hunk 1]
```diff
@@ -16,7 +16,7 @@ buildscript {
 
     repositories {
         jcenter()
-        maven("https://dl.bintray.com/magnusja/maven")
+        maven(url = "https://dl.bintray.com/magnusja/maven")
```
jurre, line 19: Could we add both a repository _with_ and without the `url` just to make sure we don't break this in the future?

jurre:
Thanks, this looks great, just 1 minor request

hfhbd:
@jurre added :)

jurre (approved):

PR #3037, opened 2021-01-25
Extract yarn/npm lockfile updater specs from FileUpdater specs

jurre:
These tests were really targeting the LockfileUpdaters through the
FileUpdater specs, by pulling them out into their own test files it
makes it easier to start adding test coverage for npm 7.

Nothing has been changed about the test structure, they've just been
pulled straight out, and the `for npm`/`for yarn` contexts have been
removed.

I'll follow this up with a few further cleanups like transforming those
files to use project based fixtures, in separate PRs.


feelepxyz (approved):
🙌  yes so much better splitting these specs into the respective pm! LGTM

PR #3040, opened 2021-01-25
Convert yarn and npm lockfile updater specs to use project based fixtures

jurre:
This converts the lockfile updater specs to used project based fixtures, making it easier to figure out what's going on in a given test.


feelepxyz (approved):
😻

PR #3045, opened 2021-01-26
Add python 3.8.7

### LINKED ISSUE #3019: Support Python 3.8.7
Python 3.8.7 was released on 21 Dec 2020

https://www.python.org/downloads/release/python-387/

Parnassius:
Closes #3019


jurre (approved):
This should be supported by our version of pyenv 👍 https://github.com/pyenv/pyenv/blob/master/CHANGELOG.md#1222

There might be some tests that fail (which is annoying and we should prevent from happening for a simple version added like this), but let's see what CI says

Parnassius:
The only one failing is go_modules, I guess it's unrelated?

PR #3046, opened 2021-01-26
Test NpmLockfileUpdater with npm7

jurre:
A few of the fixtures have not yet been updated to npm7, mainly those around private sources/auth failure


feelepxyz (approved):
Reviewed without whitespace changes and look 💯  looks like specs are still running though

jurre:
> Reviewed without whitespace changes and look 💯 looks like specs are still running though

Yeah, there are a few scenario's that I'd expect to fail since the lockfiles haven't been updated yet. It also seems like the tests timed out, not sure yet if that's related

npm_and_yarn/helpers/lib/npm7/subdependency-updater.js
```diff
@@ -2,33 +2,42 @@ const fs = require("fs");
...
+const install = async (directory, lockfileName, dependencies) => {
   await new Promise((resolve) => {
     npm.load(resolve);
   });
 
   const arb = new Arborist({
     ...npm.flatOptions,
     path: directory,
-    packageLockOnly: true,
+    packageLockOnly: false,
```
feelepxyz, line 15: This is a downgrade from what we have, results in downloading `node_modules` but seems to be the only way to get resolution errors from arborist.

The benefit of the new behaviour is that arborist only seems to bail when the updated dependencies are not resolvable and leaves existing lockfile entries as is. We'd need to rejig a bunch of test to account for this so leaving it disable for now.

feelepxyz:
@jurre think this is good to merge into the main npm 7 branch and keep working from there. There's one spec commented out for npm 7 here but planning to have another stab at that next week.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -55,6 +55,14 @@ def updated_lockfile_content(lockfile)
         MISSING_PACKAGE = %r{(?<package_req>[^/]+) - Not found}.freeze
         INVALID_PACKAGE = /Can't install (?<package_req>.*): Missing/.freeze
 
+        # TODO: look into fixing this in npm, seems like a bug in the git
+        # downloader introduced in npm 7
+        #
+        # NOTE: error message returned from arborist/npm 7 when trying to
+        # fetching a invalid/non-existent git ref
+        NPM7_MISSING_GIT_REF = /already exists and is not an empty directory/.freeze
+        NPM6_MISSING_GIT_REF = /did not match any file\(s\) known to git/.freeze
```
jurre, line 64: Thanks for pulling these out into meaningful constants ❤️

jurre:
> @jurre think this is good to merge into the main npm 7 branch and keep working from there. There's one spec commented out for npm 7 here but planning to have another stab at that next week.

Yep, agreed, this looks good!

Issue #2962, opened 2021-01-08
docker - incorrect update of manifest list sha to amd64 manifest sha


KnicKnic:
I built the latest master (0c3f58f3a33c93036a7e346f2ff42bb629d99e8e) and this repros 
**Package manager/ecosystem**
docker

**Manifest contents prior to update**
https://github.com/KnicKnic/myq-garage-server/blob/2f81d6dfac244bf06f669a5ef86d6b3b42949364/Dockerfile

```Dockerfile
FROM python:slim@sha256:4d92968b26bb6b7b62d957244de86fc1054f03793577d49e85c00864eb03ca07 as python-base
```

**Incorrect Updated dependency**
see incorrect pull request https://github.com/KnicKnic/myq-garage-server/pull/3
```Dockerfile
FROM python:slim@sha256:56d9bdc243bc53d4bb055305b58cc0be15b05cc09dcda9b9d5e224233889b61b as python-base
```

**What you expected to see, versus what you actually saw**
I did not expect to see an update as the original dockerfile referenced the most recent manifest list. I have dumped the manifest list at the time of the bug to show that it is going from the manifest list sha, to the linux/amd64 manifest sha.
```text
/mnt/c/Users/nmaliwa$ docker buildx imagetools inspect docker.io/library/python:slim
Name:      docker.io/library/python:slim
MediaType: application/vnd.docker.distribution.manifest.list.v2+json
Digest:    sha256:4d92968b26bb6b7b62d957244de86fc1054f03793577d49e85c00864eb03ca07

Manifests:
  Name:      docker.io/library/python:slim@sha256:56d9bdc243bc53d4bb055305b58cc0be15b05cc09dcda9b9d5e224233889b61b
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/amd64

  Name:      docker.io/library/python:slim@sha256:b1a7b49314a060e6e29cc0106c456f0006069d656afdc43ea0f822eae9578544
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/arm/v5

  Name:      docker.io/library/python:slim@sha256:2438965ab5800e6847d02c654abc87d8e1efc8b4b39ffcc1b120651a3943e50d
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/arm/v7

[...]

PR #3060, opened 2021-02-01
Use the correct Docker digest when checking for updates

jurre:
Previously, Dependabot would use the [Docker registry manifest
endpoint](https://docs.docker.com/registry/spec/api/#manifest) with the
`application/vnd.docker.distribution.manifest.v2+json` Media Type:

```
dt=$(curl "https://auth.docker.io/token?service=registry.docker.io&scope=repository:library/python:pull" | jq -r .access_token)
curl curl --get -v https://registry.hub.docker.com:443/v2/library/python/manifests/slim --header "Authorization: Bearer $dt" --header "Accept: application/vnd.docker.distribution.manifest.v2+json"
< HTTP/1.1 200 OK
< Content-Length: 1370
< Content-Type: application/vnd.docker.distribution.manifest.v2+json
< Docker-Content-Digest: sha256:af2d64e6de9f891bd5958a498bd2394389143749b3c57ffcc8c2aec054b28ad5
```

However, the digest that is returned in the header here, is actually
that of the first platform that's listed.

When we install that latest image, we see the digest should actually be:

```
docker images --digests | grep python
python slim sha256:50328b4efe5ff8ef8bb2e62e088a4cefeace39062c483c04987f5820e7cf73b1
```

When using the
`application/vnd.docker.distribution.manifest.list.v2+json` media type:

```
dt=$(curl "https://auth.docker.io/token?service=registry.docker.io&scope=repository:library/python:pull" | jq -r .access_token)
curl curl --get -v https://registry.hub.docker.com:443/v2/library/python/manifests/slim --header "Authorization: Bearer $dt" --header "Accept: application/vnd.docker.distribution.manifest.list.v2+json"

< HTTP/1.1 200 OK
< Content-Length: 1862
< Content-Type: application/vnd.docker.distribution.manifest.list.v2+json
< Docker-Content-Digest: sha256:50328b4efe5ff8ef8bb2e62e088a4cefeace39062c483c04987f5820e7cf73b1
```

We get the expected digest, and when inspecting the body, we can see
that we previously got the digest for amd64:

```json
{
  "manifests": [
    {
      "digest": "sha256:af2d64e6de9f891bd5958a498bd2394389143749b3c57ffcc8c2aec054b28ad5",
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      },
      "size": 1370
    },
```

The tests have remained unchanged in this PR, because (unfortunately)
all the requests and responses are stubbed out.

Fixes https://github.com/dependabot/dependabot-core/issues/2962

### DIFF
--- a/docker/lib/dependabot/docker/update_checker.rb
+++ b/docker/lib/dependabot/docker/update_checker.rb
@@ -9,6 +9,29 @@ require "dependabot/docker/version"
 require "dependabot/docker/requirement"
 require "dependabot/docker/utils/credentials_finder"
 
+module DockerRegistry2
+  class Registry
+    private
+
+    # By default the Docker Registry client sets the Accept header to
+    # `application/vnd.docker.distribution.manifest.v2+json`
+    # This results in the digest of a specific platform to be returned, we
+    # want to override this header so we can fetch the generic digest
+    # associated with the given repo/tag.
+    def headers(payload: nil, bearer_token: nil)
+      headers = {}
+      headers["Authorization"] = "Bearer #{bearer_token}" unless bearer_token.nil?
+      if payload.nil?
+        headers["Accept"] =
+          "application/vnd.docker.distribution.manifest.list.v2+json, application/json"
+      end
+      headers["Content-Type"] = "application/vnd.docker.distribution.manifest.v2+json" unless payload.nil?
+
+      headers
+    end
+  end
+end
+
 module Dependabot
   module Docker
     class UpdateChecker < Dependabot::UpdateCheckers::Base
@@ -24,7 +47,7 @@ module Dependabot
           #{VERSION_WITH_PFX}|
           #{VERSION_WITH_SFX}|
           #{VERSION_WITH_PFX_AND_SFX}
-        /x.freeze
+      /x.freeze
 
       def latest_version
         @latest_version ||= fetch_latest_version
@@ -188,6 +211,7 @@ module Dependabot
         @tags_from_registry ||=
           begin
             client = docker_registry_client
+
             client.tags(docker_repo_name, auto_paginate: true).fetch("tags")
           rescue *transient_docker_errors
             attempt ||= 1


docker/lib/dependabot/docker/update_checker.rb
```diff
@@ -9,6 +9,24 @@
 require "dependabot/docker/requirement"
 require "dependabot/docker/utils/credentials_finder"
 
+module DockerRegistry2
```
feelepxyz, line 12: Thoughts on adding a comment here explaining why we need to monkey patch?

jurre: Yeah, good call 👍

feelepxyz: looks like it's also used from the file parser, extract the monkey patch and require it in both places?

jurre: And actually, to be safe I'm going to override this on the instance instead of globally. We reuse the docker registry in another class as well, so this might have (but currently doesn't) unintended consequences

feelepxyz: Ah right, not sure I fully follow the impact on what we do in the file parser, will this work there too or not needed? Digest use in file parser: https://github.com/dependabot/dependabot-core/blob/577c4039d51e88e324f8be5e4de5ff8d8822ed96/docker/lib/dependabot/docker/file_parser.rb#L103

feelepxyz: seems this is used to figure out the "from" version of the dependency, not sure if this makes any difference?

feelepxyz (approved):
👍  should we be upstreaming this change to the gem?

jurre:
> 👍 should we be upstreaming this change to the gem?

Yeah, good shout. I'll open an issue to see what they think

PR #1388, opened 2019-09-18
Adds terraform 0.12 support

userhas404d:
Fixes https://github.com/dependabot/dependabot-core/issues/1176

I opted for both [hcl2json](https://github.com/tmccombs/hcl2json) and [terraform-config-inspect](https://github.com/hashicorp/terraform-config-inspect). 

`hcl2json` for terragrunt and 
`terraform-config-inspect` for tf 0.12

I wanted to go with `terraform-config-inspect` for both, but it didn't play nice with terragrunt :disappointed:

Notes: 
- I created my own fork of [hcl2json](https://github.com/userhas404d/hcl2json/releases) in order to publicly host the binaries for it
- opting for `terraform-config-inspect` adds go as a dependency to being able to support this language. Not sure if that's kosher or not..

PR #3063, opened 2021-02-02
Terraform 12

jmahowald:
This is an attempt to get "terraform 12" hcl (should be good for future versions) working.  Most notably, you do NOT need to wrap up values from other terraform resources in interpolation if there is no other string to include.


jmahowald:
I am aware that not all tests are passing yet, there are some reorderings still necessary (the newer tool used for parsing returns modules in alphabetical order) in tests, as well as terragrunt fixes I still have to investigate.

jurre:
Thanks for diving into this @jmahowald! Would this approach support both versions? I don't think so, right? Typically that's what we've been trying to do with these major version updates, but in this case since HCL 1 doesn't really seem to be used that much anymore, we might be able to drop support for it when we upgrade. I'll try to pull some usage data on this, but also curious to hear your thoughts.

jmahowald:
This should work with both, most of the spec tests are working unchanged, the two parsers simply return a different ordering for modules.    I've used hcl2json on both versions of terraform in other contexts as well.

userhas404d:
😢 Unless I'm missing something this is a dupe of https://github.com/dependabot/dependabot-core/pull/1388

@jmahowald time might be better spent getting tf014 support added to 1388 or to @patrickjahns' [fork](https://github.com/patrickjahns/dependabot-core/tree/feature_terraform_12). I would, but I don't have the cycles

jmahowald:
> 😢 Unless I'm missing something this is a dupe of #1388
> 
> @jmahowald time might be better spent getting tf014 support added to 1388 or to @patrickjahns' [fork](https://github.com/patrickjahns/dependabot-core/tree/feature_terraform_12). I would, but I don't have the cycles

@jurre  after having to watch team members update 20 some repos with simple version updates I'm certainly willing to move my energy to seeing if I can help along #1388 but that issue appears stalled. I'm not sure if that stalling is because of the issue that @userhas404d  has found, or if it's just stalled for other reasons. I understand you can't make a commitment, but do you think that if we can resolve the issues there you'd likely be able to look at it in the nearish future?

jmahowald:
> 😢 Unless I'm missing something this is a dupe of #1388
> 
> @jmahowald time might be better spent getting tf014 support added to 1388 or to @patrickjahns' [fork](https://github.com/patrickjahns/dependabot-core/tree/feature_terraform_12). I would, but I don't have the cycles

There weren't meaningful language changes that I'm aware of that should be different from 12 to 14 from an HCL parsing perspective, though it certainly would make sense for dependabot in terraform 13+ to handle the new ways of declaring provider versions.  Are there language changes you're thinking that should be there for 14 that would require other changes than what you made to support HCL 2?

userhas404d:
> Are there language changes you're thinking that should be there for 14 that would require other changes than what you made to support HCL 2?

thinking specifically about the new dependency lock file: https://www.terraform.io/docs/language/dependency-lock.html lock file specific update logic either needs to be added to the terraform update workflow or ignored outright.

jurre:
> @jurre after having to watch team members update 20 some repos with simple version updates I'm certainly willing to move my energy to seeing if I can help along #1388 but that issue appears stalled. I'm not sure if that stalling is because of the issue that @userhas404d has found, or if it's just stalled for other reasons. I understand you can't make a commitment, but do you think that if we can resolve the issues there you'd likely be able to look at it in the nearish future?

It's mostly stalled because we haven't been able to get to it yet. We have planned to add support for Terraform HCL 2 this quarter, but have committed to a few other ecosystem upgrades (npm/bundler) before that. Given that the team working on these updates is pretty tiny, we're focussing on one upgrade at a time.

Any issues that you're able to tackle are definitely very much appreciated, we just might be a little slow to respond/review. 💛

jurre:
Terraform 12 (HCL 2) support is now available, thank you so much for this PR, it was instrumental in landing this.

PR #3070, opened 2021-02-03
Nuget: Parse floating notation when used in range

jurre:
Typically you’d define a Nuget range as `[2.0, 3.0)` which translates to: "higher than
or equal to 2.0 but lower than 3.0". You can also say `2.0.*` which means any
patch version of 2.0.

However, it's valid to combine these like `[2.1.*,3.0.0)`.

The floating notation (`*`) seems like it’s redundant in this case,
as the `3.0.0)` bit would allow 2.2 to be installed.

I could not find any documentation on this, so I made a best guess
effort to do what seems right, previously a notation like this would
fail and cause the update to error. I've verified that this resolves
those errors when dry-running.

### DIFF
--- a/nuget/lib/dependabot/nuget/requirement.rb
+++ b/nuget/lib/dependabot/nuget/requirement.rb
@@ -49,13 +49,18 @@ module Dependabot
         return convert_dotnet_range_to_ruby_range(req_string) if req_string&.start_with?("(", "[")
 
         return req_string.split(",").map(&:strip) if req_string.include?(",")
+
         return req_string unless req_string.include?("*")
 
         convert_wildcard_req(req_string)
       end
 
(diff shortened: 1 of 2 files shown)


feelepxyz (approved):
Seems like a good fix to me 👍  don't know that much about nuget version ranges but seems sensible to treat `*` as `0`.

Is it valid do something like this in nuget? `[*, 2.0.0])`

jurre:
Good question, I think so? But it's kind of pointless, since it would be the same as writing `[, 2.0.0)`

feelepxyz:
> Good question, I think so? But it's kind of pointless, since it would be the same as writing `[, 2.0.0)`

Good point 👍  :shipit:

PR #3071, opened 2021-02-03
Nuget: Ignore `Remove` ProjectReferences

jurre:
Nuget allows us to reference nested projects using the
`<ProjectReference Include="some\project.csproc">` notation. You can
also explicitly remove references as shown in the fixture added in this
PR.

In that case we used to error, now we just ignore that reference
instead.


feelepxyz (approved):
LGTM 👍  nice having a targeted unit spec

PR #3075, opened 2021-02-04
Composer: handle unreachable path vcs source

jurre:
When a Composer repository is specified as a local path source which is
not available we should raise a DependencyFileNotResolvable error.

### DIFF
--- a/composer/lib/dependabot/composer/update_checker/version_resolver.rb
+++ b/composer/lib/dependabot/composer/update_checker/version_resolver.rb
@@ -255,7 +255,8 @@ module Dependabot
               named_captures.fetch("url")
             raise Dependabot::GitDependenciesNotReachable, dependency_url
           elsif error.message.start_with?("Could not parse version") ||
-                error.message.include?("does not allow connections to http://")
+                error.message.include?("does not allow connections to http://") ||
+                error.message.match?(/The `url` supplied for the path .* does not exist/)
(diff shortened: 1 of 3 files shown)


composer/lib/dependabot/composer/update_checker/version_resolver.rb
```diff
@@ -255,7 +255,8 @@ def handle_composer_errors(error)
               named_captures.fetch("url")
             raise Dependabot::GitDependenciesNotReachable, dependency_url
           elsif error.message.start_with?("Could not parse version") ||
-                error.message.include?("does not allow connections to http://")
+                error.message.include?("does not allow connections to http://") ||
+                error.message.match?("The `url` supplied for the path .* does not exist")
```
feelepxyz, line 259: Should this use `include?` to match the partial string? Assumed `match?` was for regexes

feelepxyz (approved):
👍  added a nit comment about using `include?` over `match?`

composer/lib/dependabot/composer/update_checker/version_resolver.rb
jurre: Yeah it is, it should actually be `error.message.match?(/The `url` supplied for the path .* does not exist/)`, I guess it's being coerced into a regex but it's definitely not what I intended, cheers

PR #3077, opened 2021-02-04
Composer: Handle version constraints with both caret and dev postfix

jurre:
The `.x-dev` postfix is a way to tell composer to check out a branch
instead of a tag. For the purpose of defining the version range, we can
ignore this notation and replace it with a `0` to ensure we match the
right range.


feelepxyz (approved):
Nice one 👍

PR #3078, opened 2021-02-04
Don't raise when adding temp platform extensions

jurre:
It's not clear to me why we need to raise in this scenario, those errors
are not handled. It seems better to attempt the update without adding
the extensions, for some cases that I've tried locally that seems to
update dependencies just fine.

Maybe this behavior was improved with composer 2?

### DIFF
--- a/composer/lib/dependabot/composer/update_checker/version_resolver.rb
+++ b/composer/lib/dependabot/composer/update_checker/version_resolver.rb
@@ -159,7 +159,7 @@ module Dependabot
           json = JSON.parse(content)
 
           composer_platform_extensions.each do |extension, requirements|
-            raise "No matching version for #{requirements}!" unless version_for_reqs(requirements)
+            next unless version_for_reqs(requirements)
 
             json["config"] ||= {}
             json["config"]["platform"] ||= {}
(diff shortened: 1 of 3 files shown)


composer/lib/dependabot/composer/update_checker/version_resolver.rb
```diff
@@ -159,7 +159,7 @@ def add_temporary_platform_extensions(content)
           json = JSON.parse(content)
 
           composer_platform_extensions.each do |extension, requirements|
-            raise "No matching version for #{requirements}!" unless version_for_reqs(requirements)
+            next unless version_for_reqs(requirements)
```
feelepxyz, line 162: Oh yeah no idea why we did this 🤔 is there a test we could add for this new behaviour? 

I think we've been too eager to raise in the past instead of trying to find ways of gracefully degrading/continuing so 👍  on this direction.

jurre: Yeah I think I can add a meaningful test based on the dry-run I just did, I'd initially opened the PR to check if it wouldn't introduce any failures in CI.

feelepxyz (approved):

PR #3079, opened 2021-02-04
Composer: handle invalid version string

jurre:
This would previously bubble up as an unknown error, but we should
inform the user their manifest file is not resolvable.

We handle a slightly different variant of this error already.


feelepxyz (approved):
✨

PR #3081, opened 2021-02-04
v0.131.1

jurre:
- Composer: handle invalid version string
- Composer: Don't raise when adding temp platform extensions
- Composer: Handle version constraints with both caret and dev postfix
- Docker: Use the correct Docker digest when checking for updates


feelepxyz (approved):
🙌

PR #3083, opened 2021-02-05
Maven: Raise DependencyFileNotResolvable when invalid repo is specified

jurre:
Previously an invalid URI would result in an unhandled error. When one
is specified, we cannot resolve any of the dependencies, so we should
raise an error indicating such.


feelepxyz (approved):
👍

PR #3087, opened 2021-02-05
Nuget: handle version ranges in VersionFinder

jurre:
Previously, a version specified as a range would result in a `undefined
method `any?' for "[5.1.0, 6.0.0)":String` error. This prevents that
from happening.


feelepxyz (approved):
🙌

PR #3098, opened 2021-02-08
Dependabot::UnexpectedExternalCode

thepwagner:
This introduces `Dependabot::UnexpectedExternalCode` as a new error, to be raised when Dependabot
determines it may need to evaluate external code to update a dependency.

This exception will be part of an **opt-in** hardening, to
be used in conjunction with features that extend the resources
accessible to Dependabot (e.g. private packages and sources).

The included sample and test cases centre around the Bundler ecosystem,
where Git sources which may result in a evaluating the Ruby code contained in the the dependency's
`.gemspec`, e.g. https://github.com/dependabot/dependabot-core/blob/main/common/dependabot-common.gemspec

### DIFF
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -103,6 +103,7 @@ $options = {
   write: false,
   clone: false,
   lockfile_only: false,
+  reject_external_code: false,
   requirements_update_strategy: nil,
   commit: nil,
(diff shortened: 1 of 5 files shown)


jasonrudolph (approved):
Nice addition, @thepwagner. ✨

PR #3106, opened 2021-02-09
npm: fix npm workspace bug

feelepxyz:
Fixes one of the bugs discovered in this update:
https://github.com/AviVahl/ts-tools/pull/152

This particular update no raises when dry-running: `bin/dry-run.rb npm_and_yarn AviVahl/ts-tools --cache files --dep yargs` - we'll need to do some further work in the file parser and update checker to not attempt this update in the first place.

Dependabot will no longer attempt to add the nested dependency to the root level lockfile when doing updates to npm workspace projects.

This is done by checking if the dependency being updated is defined in a `package.json` in the same folder as the lockfile being updated. If it's a nested package we run `npm install` without any package@version arguments letting npm update the nested package references which have already been updated in the package.json by the `PackageJsonUpdater`.

Additionally, I've added a workaround to stop adding the `name` attribute to the lockfile `packages.""` section when running `npm install` with `--package-lock-only`. This seems to be a bug with npm but without this fix dependabot will always add this attribute and it will most likely never exist in a users repo so adding confusion about why it's added.

We also sanitize invalid `package.json` names, e.g. `{{ name }}` to `something` during the update, we don't want this to bleed into the updated lockfile.

TODO
- [x] Add more specs covering these cases

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
@@ -17,35 +17,22 @@ module Dependabot
         require_relative "npmrc_builder"
         require_relative "package_json_updater"
 
-        def initialize(dependencies:, dependency_files:, credentials:)
+        def initialize(lockfile:, dependencies:, dependency_files:, credentials:)
+          @lockfile = lockfile
           @dependencies = dependencies
           @dependency_files = dependency_files
           @credentials = credentials
         end
 
-        def updated_lockfile_content(lockfile)
-          return lockfile.content if npmrc_disables_lockfile?
-          return lockfile.content if updatable_dependencies(lockfile).empty?
-
-          @updated_lockfile_content ||= {}
-          @updated_lockfile_content[lockfile.name] ||=
-            SharedHelpers.in_a_temporary_directory do
-              path = Pathname.new(lockfile.name).dirname.to_s
-              lockfile_name = Pathname.new(lockfile.name).basename.to_s
-              write_temporary_dependency_files(lockfile.name)
-              updated_files = Dir.chdir(path) do
-                run_current_npm_update(lockfile_name: lockfile_name, lockfile_content: lockfile.content)
-              end
-              updated_content = updated_files.fetch(lockfile_name)
-              post_process_npm_lockfile(lockfile.content, updated_content, lockfile.name)
-            end
-        rescue SharedHelpers::HelperSubprocessFailed => e
-          handle_npm_updater_error(e, lockfile)
+        def updated_lockfile
+          updated_file = lockfile.dup
+          updated_file.content = updated_lockfile_content
+          updated_file
         end
 
         private
 
-        attr_reader :dependencies, :dependency_files, :credentials
+        attr_reader :lockfile, :dependencies, :dependency_files, :credentials
 
         UNREACHABLE_GIT = /fatal: repository '(?<url>.*)' not found/.freeze
         FORBIDDEN_GIT = /fatal: Authentication failed for '(?<url>.*)'/.freeze
@@ -63,6 +50,21 @@ module Dependabot
         NPM7_MISSING_GIT_REF = /already exists and is not an empty directory/.freeze
         NPM6_MISSING_GIT_REF = /did not match any file\(s\) known to git/.freeze
 
+        def updated_lockfile_content
+          return lockfile.content if npmrc_disables_lockfile?
+          return lockfile.content unless updatable_dependencies.any?
+
+          @updated_lockfile_content ||=
+            SharedHelpers.in_a_temporary_directory do
+              write_temporary_dependency_files
+              updated_files = Dir.chdir(lockfile_directory) { run_current_npm_update }
+              updated_lockfile_content = updated_files.fetch(lockfile_basename)
+              post_process_npm_lockfile(updated_lockfile_content)
+            end
+        rescue SharedHelpers::HelperSubprocessFailed => e
+          handle_npm_updater_error(e)
+        end
+
         def top_level_dependencies
           dependencies.select(&:top_level?)
         end
@@ -71,16 +73,14 @@ module Dependabot
           dependencies.reject(&:top_level?)
         end
 
-        def updatable_dependencies(lockfile)
+        def updatable_dependencies
           dependencies.reject do |dependency|
-            dependency_up_to_date?(lockfile, dependency) ||
-              top_level_dependency_update_not_required?(dependency, lockfile)
+            dependency_up_to_date?(dependency) || top_level_dependency_update_not_required?(dependency)
           end
         end
 
-        def lockfile_dependencies(lockfile)
-          @lockfile_dependencies ||= {}
-          @lockfile_dependencies[lockfile.name] ||=
+        def lockfile_dependencies
+          @lockfile_dependencies ||=
             NpmAndYarn::FileParser.new(
               dependency_files: [lockfile, *package_files],
               source: nil,
@@ -88,9 +88,8 @@ module Dependabot
             ).parse
         end
 
-        def dependency_up_to_date?(lockfile, dependency)
-          existing_dep = lockfile_dependencies(lockfile).
-                         find { |dep| dep.name == dependency.name }
+        def dependency_up_to_date?(dependency)
+          existing_dep = lockfile_dependencies.find { |dep| dep.name == dependency.name }
 
           # If the dependency is missing but top level it should be treated as
           # not up to date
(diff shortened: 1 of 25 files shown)


npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater_spec.rb
```diff
@@ -2063,6 +2063,36 @@
           end
         end
 
+        context "with a dependency that's actually up-to-date but has the wrong previous version" do
```
feelepxyz, line 2066: @jurre added a test case for the bogus update attempt

npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater_spec.rb
```diff
@@ -167,6 +168,33 @@
           to include("19c4dba3bfce7574e28f1df2138d47ab4cc665b3")
       end
     end
+
+    context "when the packages name needs sanitizing" do
+      let(:files) { project_dependency_files("npm7/simple") }
+
+      it "restores the packages name attribute" do
+        parsed_lockfile = JSON.parse(updated_npm_lock_content)
+        expect(parsed_lockfile.dig("packages", "", "name")).to eq("{{ name }}")
```
feelepxyz, line 177: This is sad, we currently re-write this to `something` because https://github.com/dependabot/dependabot-core/blob/f7d4608eed73741fa0d0a759f7e93bf74b9d2c9e/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb#L786

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -637,14 +615,40 @@ def post_process_npm_lockfile(original_content, updated_content, lockfile_name)
...
           # Switch back npm 7 lockfile "pacakages" requirements from the package.json
-          updated_content = restore_locked_package_dependencies(lockfile_name, updated_content)
+          updated_content = restore_locked_package_dependencies(updated_content)
 
           # Switch back the protocol of tarball resolutions if they've changed
           # (fixes an npm bug, which appears to be applied inconsistently)
           replace_tarball_urls(updated_content)
         end
 
+        def restore_packages_name(updated_content)
```
feelepxyz, line 629: @jurre this is new to handle resetting `packages."".name`

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -17,35 +17,22 @@ class NpmLockfileUpdater
         require_relative "npmrc_builder"
         require_relative "package_json_updater"
 
-        def initialize(dependencies:, dependency_files:, credentials:)
```
feelepxyz, line 20: Most of this is a re-write because we no longer need pass around `lockfile`

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -106,93 +105,77 @@ def dependency_up_to_date?(lockfile, dependency)
...
               ]
             )
           end
         end
 
-        def run_npm_7_top_level_updater(lockfile_name:, top_level_dependency_updates:)
-          # - `--dry-run=false` the updater sets a global .npmrc with dry-run: true to
-          #   work around an issue in npm 6, we don't want that here
+        def run_npm_7_top_level_updater(top_level_dependencies:)
+          is_dependency_in_current_package_json = top_level_dependencies.any? do |dependency|
```
feelepxyz, line 156: Adding this to handle the bug where we add nested workspace package dependencies to the root lockfile.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
jurre: That's a great improvement 👍

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -637,14 +619,43 @@ def post_process_npm_lockfile(original_content, updated_content, lockfile_name)
...
+          # TODO: Submit a patch to npm fixing this issue making `npm install`
+          # consistent with `npm install --package-lock-only`
+          #
+          # NOTE: This is a workaround for npm adding a `name` attribute to the
+          # packages section in the lockfile because we install using
+          # `--package-lock-only`
+          current_name = parsed_updated_lockfile.dig("packages", "", "name")
+          original_name = parsed_original_package["name"]
+          if !parsed_original_lockfile.dig("packages", "", "name")
+            updated_content = updated_content.gsub(/"": {[\n\s]+"name": "#{current_name}",/, '"": {')
```
jurre, line 649: Could we pull out this regex into a constant/method with an explanatory name?

npm_and_yarn/spec/fixtures/projects/npm7/packages_name_missing/package-lock.json
```diff
@@ -0,0 +1,163 @@
+{
+    "name": "{{ name }}",
+    "version": "1.0.0",
+    "lockfileVersion": 2,
+    "requires": true,
+    "packages": {
```
jurre, line 6: Could we use a more minimal fixture to test this? Makes it a bit easier to spot what's going on

jurre (approved):
Left some stylistic nits inline, but overall seems like the right thing to do 👍

npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater_spec.rb
```diff
@@ -168,6 +168,81 @@
...
       end
     end
+
+    context "when the packages name needs sanitizing" do
+      let(:files) { project_dependency_files("npm7/simple") }
+
+      it "restores the packages name attribute" do
+        parsed_lockfile = JSON.parse(updated_npm_lock_content)
+        expected_updated_npm_lock_content = fixture("updated_projects", "npm7", "simple", "package-lock.json")
+        expect(updated_npm_lock_content).to eq(expected_updated_npm_lock_content)
```
feelepxyz, line 178: Added a bunch of "updated" fixtures to make sure we don't mess up whitespace etc. Could be cool to automate this at some point like jest snapshot testing https://jestjs.io/docs/en/snapshot-testing

npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater_spec.rb
```diff
@@ -2090,6 +2120,8 @@
             expect(updated_files.map(&:name)).
               to match_array(%w(package-lock.json packages/package1/package.json))
             expect(parsed_root_lockfile.dig("dependencies", "etag", "version")).to eq("1.8.1")
+            # Doesn't install etag as a top-level dependency
+            expect(parsed_root_lockfile.dig("packages", "", "dependencies", "etag")).to be_nil
```
feelepxyz, line 2124: This isn't the most robust spec, would be better to add a updated lockfile fixture and expect the content matches.

npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater_spec.rb
```diff
@@ -2085,20 +2125,30 @@
...
+            expected_updated_npm_lock_content = fixture(
+              "updated_projects", "npm7", "workspaces_dev", "package-lock.json"
+            )
+            parsed_npm_lockfile = JSON.parse(updated_npm_lock_content.content)
             expect(updated_files.map(&:name)).
-              to match_array(%w(package-lock.json packages/package1/package.json))
-            expect(parsed_root_lockfile.dig("dependencies", "etag", "version")).to eq("1.8.1")
+              to match_array(%w(package-lock.json other_package/package.json packages/package1/package.json))
+            expect(parsed_npm_lockfile.dig("dependencies", "etag", "version")).to eq("1.8.1")
+            expect(updated_npm_lock.content).to eq(expected_updated_npm_lock_content)
```
feelepxyz, line 2136: added an expected updated fixture to make this spec a bit more robust, created it by manually updating the nested package.json requirement and running `npm install` in the root folder

feelepxyz: FYI this was previously adding `etag` as a top-level dependency but wasn't covered by this spec

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -627,13 +642,51 @@ def replace_project_name(updated_lockfile_content, parsed_updated_lockfile_conte
...
             /"name":\s"#{current_name}"/,
             "\"name\": \"#{original_name}\""
           )
         end
 
+        def replace_lockfile_packages_name_attribute(current_name, original_name, updated_lockfile_content)
+          packages_key_line = '"": {'
+          updated_lockfile_content.sub(
+            /(#{packages_key_line}[\n\s]+"name":\s)"#{current_name}"/,
+            '\1"' + original_name + '"'
```
feelepxyz, line 681: Was surprised that you can't use backreferences in a double quoted string, e.g. `"\1\"#{original_name}\""`

jurre (approved):

PR #3108, opened 2021-02-09
v0.133.0

thepwagner:
This PR:
* https://github.com/dependabot/dependabot-core/pull/3098

Tiny feature to a single ecosystem, but semver "minor" since it adds to the `FileParser` interface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## v0.133.0, 9 February 2021
+
+- Bundler: Raise UnexpectedExternalCode if `reject_external_code: true`
+
 ## v0.132.0, 8 February 2021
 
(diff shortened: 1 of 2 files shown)


jurre (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## v0.133.0, 9 February 2021
+
+- Bundler: Raise UnexpectedExternalCode if `reject_external_code: true`
```
jasonrudolph, line 3: Should we say a bit more about this? Maybe something like this?

```suggestion
- Bundler: Raise UnexpectedExternalCode if `reject_external_code: true` and the update involves external code
```

PR #3111, opened 2021-02-09
Npm7: correctly parse version from package dependencies

jurre:
Previously, when using npm 7 workspaces, when a dependency was installed
from a workspaces' `package.lock` but a _different_ version of that
dependency was installed via the top-level `package.json`, we would grab
the top-levels version, resulting in incorrect behavior.

This changes the implementation to look in the lockfiles' `packages`
object for the actual dependency from that workspace, and grab the we
need there.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
@@ -2,6 +2,7 @@
 
 require "dependabot/dependency_file"
 require "dependabot/npm_and_yarn/file_parser"
+require "dependabot/npm_and_yarn/helpers"
 
 module Dependabot
   module NpmAndYarn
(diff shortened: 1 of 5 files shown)


npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
```diff
@@ -65,6 +52,43 @@ def potential_lockfiles_for_manifest(manifest_filename)
...
         end
 
+        def npm_lockfile_details(lockfile, dependency_name, manifest_name)
+          parsed_lockfile = parse_package_lock(lockfile)
+
+          if Helpers.npm_version(lockfile.content) == "npm7"
+            parsed_lockfile.dig(
+              "packages",
+              node_modules_path(manifest_name, dependency_name)
+            )&.slice("version", "resolved", "integrity", "dev")
```
jurre, line 62: @feelepxyz this is the actual change to grab the right data from the lockfile in this scenario

npm_and_yarn/spec/dependabot/npm_and_yarn/file_parser/lockfile_parser_spec.rb
```diff
@@ -307,5 +307,20 @@
         )
       end
     end
+
+    context "for an npm7 workspace project with a direct dependency in the workspace" do
+      let(:dependency_files) { project_dependency_files("npm7/workspace_nested_package") }
```
feelepxyz, line 312: Should we add some more specs here for npm 7 lockfiles? e.g. not in a workspace

feelepxyz (commented):
Nice one! Fix looks straight forward to me 👍 would be nice with a few more test cases for npm 7 as it now uses a different code path and only the workspace looks tested?

npm_and_yarn/spec/dependabot/npm_and_yarn/file_parser/lockfile_parser_spec.rb
jurre: Yeah maybe! I figured most of this was covered already in the `FileParser` specs, do you have any scenario's in mind?

feelepxyz (approved):
LGTM 👍

npm_and_yarn/spec/dependabot/npm_and_yarn/file_parser/lockfile_parser_spec.rb
feelepxyz: Yeah the added spec looks good 👍

PR #3112, opened 2021-02-10
Refactor NpmLockfileUpdater

feelepxyz:
Extracting a refactor from
https://github.com/dependabot/dependabot-core/pull/3106

Initialize the lockfile updater with a lockfile being updated removing
the need to pass around the lockfile everywhere.

A few spec changes have snuck in from the above pr but these should be
cosmetic as no fuctionality has changed.


npm_and_yarn/spec/fixtures/projects/npm7/simple/package-lock.json
```diff
@@ -1,11 +1,11 @@
 {
-    "name": "{{ name }}",
+    "name": "project-name",
```
feelepxyz, line 2: This is more realistic example where the lockfile has a "valid" generated name by some build step, we make sure to restore this name, not the invalid name in the `package.json`, the code to restore this hasn't changed

jurre (approved):
This looks great, thanks for pulling it out!

PR #3117, opened 2021-02-10
Convert LockfileParserSpec to use project based fixtures

jurre:
A lot of these projects only have a lockfile in there, but I think it's still nice for consistency to structure them this way


feelepxyz (approved):
🙌

Issue #3074, opened 2021-02-03
[ruby:bundler] Failing to create PRs with JSON::GeneratorError


filipkis:
For about 3 weeks now we're not getting any new dependancy updates for our **ruby:bundler** ecosystem. This correlates to the moment we turned the `vendor: true` option, but I didn't yet confirm causation (i.e. if reverting that change would fix it, but we would really like to have vendoring on).

Here is the relevant part of the log for one dependency. All that need to be updated fail at same place with the same error.

```
updater | INFO <job_86767312> Checking if newrelic_rpm 6.14.0 needs updating
  proxy | 2021/02/03 16:58:57 [148] GET https://rubygems.org:443/api/v1/versions/newrelic_rpm.json
  proxy | 2021/02/03 16:58:57 [148] 200 https://rubygems.org:443/api/v1/versions/newrelic_rpm.json
updater | INFO <job_86767312> Latest version is 6.15.0
  proxy | 2021/02/03 16:58:58 [150] GET https://index.rubygems.org:443/versions
  proxy | 2021/02/03 16:58:58 [150] 206 https://index.rubygems.org:443/versions
updater | INFO <job_86767312> Requirements to unlock own
updater | INFO <job_86767312> Requirements update strategy bump_versions
updater | INFO <job_86767312> Updating newrelic_rpm from 6.14.0 to 6.15.0
  proxy | 2021/02/03 16:59:13 [152] GET https://index.rubygems.org:443/versions
  proxy | 2021/02/03 16:59:13 [152] 206 https://index.rubygems.org:443/versions
  proxy | 2021/02/03 16:59:24 [154] GET https://rubygems.org:443/gems/newrelic_rpm-6.15.0.gem
  proxy | 2021/02/03 16:59:24 [154] 200 https://rubygems.org:443/gems/newrelic_rpm-6.15.0.gem
updater | INFO <job_86767312> Submitting newrelic_rpm pull request for creation
updater | I, [2021-02-03T16:59:31.805722 #72]  INFO -- sentry: ** [Raven] Sending event e3c9ae4f7724410f8c9a2afe16318955 to Sentry
  proxy | 2021/02/03 16:59:31 [156] POST https://sentry.io:443/api/1451818/store/
  proxy | 2021/02/03 16:59:31 [156] 200 https://sentry.io:443/api/1451818/store/
updater | ERROR <job_86767312> Error processing newrelic_rpm (JSON::GeneratorError)
[...]

PR #3124, opened 2021-02-11
common: assume vendored resources are binary

thepwagner:
Previously Dependabot included a list of binary mime types and encoded
only binary files.
Invert to include a list of acceptable text mime types to leave
un-encoded.

Per the included test, the edge case here is `iso-8859-1` strings, which
aren't binary but also aren't valid `utf-8`.

# Related
- Closes https://github.com/dependabot/dependabot-core/issues/3074

### DIFF
--- a/common/lib/dependabot/file_updaters/vendor_updater.rb
+++ b/common/lib/dependabot/file_updaters/vendor_updater.rb
@@ -53,7 +53,7 @@ module Dependabot
 
       private
 
-      BINARY_ENCODINGS = %w(application/x-tarbinary binary).freeze
+      TEXT_ENCODINGS = %w(us-ascii utf-8).freeze
 
       attr_reader :repo_contents_path, :vendor_dir
 
@@ -62,7 +62,7 @@ module Dependabot
 
         encoding = `file -b --mime-encoding #{path}`.strip
 
-        BINARY_ENCODINGS.include?(encoding)
+        !TEXT_ENCODINGS.include?(encoding)
       end
     end
   end
--- a/common/spec/dependabot/file_updaters/vendor_updater_spec.rb
+++ b/common/spec/dependabot/file_updaters/vendor_updater_spec.rb
@@ -89,6 +89,31 @@ RSpec.describe Dependabot::FileUpdaters::VendorUpdater do
       )
     end
 
+    context "with iso-8859 files" do
+      before do
+        in_cloned_repository(repo_contents_path) do
+          File.write("vendor/cache/utf8.txt", "special ü".encode("utf-8"))
+          File.write("vendor/cache/iso8859.txt", "special ü".encode("iso-8859-1"))
+        end
+      end
+
+      it "marks binary files as such" do
+        file = updated_files.find do |f|
+          f.name == "vendor/cache/iso8859.txt"
+        end
+
+        expect(file.binary?).to be_truthy
+      end
+
+      it "does not mark all files as binary" do
+        file = updated_files.find do |f|
+          f.name == "vendor/cache/utf8.txt"
+        end
+
+        expect(file.binary?).to be_falsy
+      end
+    end
+
     context "in a directory" do
       let(:project_name) { "nested_vendor_gems" }
       let(:directory) { "nested" }


jurre (approved):
Nice one! I actually thought we'd already done this at some point 😄

filipkis:
Thank you @thepwagner for this fix. Any estimate on when this will be deployed?

thepwagner:
@filipkis sorry for the delay, this slipped over a long weekend here!

This is now live, as of: `0.133.3-23c9359e16a1afa23c9e39b780870bd72a429895 (since 2021-02-16 16:10:48 +0000)`

filipkis:
No worries, not a big delay :)

I run the dependabot on the test repo and while PRs do get create they have many files (instead of just one or couple of gems). It seems that gems somehow get unpacked. [Here is an example](https://github.com/filipkis/test-dependabot-bundler/pull/2/files)

thepwagner:
@filipkis isn't that expected? 

Since [json_vat](https://github.com/filipkis/test-dependabot-bundler/blob/c29cbbdf7a6907609d5f86daa008d492b1c4be33/Gemfile#L49) uses a git dependency, "vendoring" means copying that repository's contents at the target commit.

```
$ docker run --rm -it ruby:2.5.5 bash
root@25d83f8cdd02:/# git clone https://github.com/filipkis/test-dependabot-bundler.git
root@25d83f8cdd02:/# cd test-dependabot-bundler/

# ... hack to remove bundler2 dependency, otherwise unchanged
root@25d83f8cdd02:/test-dependabot-bundler# git diff Gemfile.lock
diff --git a/Gemfile.lock b/Gemfile.lock
index b9c039f..d3911f0 100644
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -689,6 +689,3 @@ DEPENDENCIES

 RUBY VERSION
    ruby 2.5.5p157
-
-BUNDLED WITH
-   2.1.4

# `bundle cache --all` vendors git dependencies, regardless of Dependabot:
root@25d83f8cdd02:/test-dependabot-bundler# bundle cache --all
NOTE: Gem::Specification#has_rdoc= is deprecated with no replacement. It will be removed on or after 2018-12-01.
Gem::Specification#has_rdoc= called from /usr/local/bundle/bundler/gems/premailer-214543a3b214/premailer.gemspec:8.
Updating files in vendor/cache
/test-dependabot-bundler/vendor/cache/premailer-214543a3b214/lib/premailer/version.rb:3: warning: already initialized constant Premailer::VERSION
/usr/local/bundle/bundler/gems/premailer-214543a3b214/lib/premailer/version.rb:3: warning: previous definition of VERSION was here
NOTE: Gem::Specification#has_rdoc= is deprecated with no replacement. It will be removed on or after 2018-12-01.
Gem::Specification#has_rdoc= called from /test-dependabot-bundler/vendor/cache/premailer-214543a3b214/premailer.gemspec:8.
root@25d83f8cdd02:/test-dependabot-bundler# git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   Gemfile.lock

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	vendor/cache/json-vat-6f6f3420f7ce/
	vendor/cache/libv8-8.4.255.0-x86_64-linux.gem
	vendor/cache/nokogiri-1.11.0-x86_64-linux.gem
	vendor/cache/premailer-214543a3b214/
	vendor/cache/spring-commands-rspec-434b4e7616b9/

no changes added to commit (use "git add" and/or "git commit -a")
```

Treating vendored git dependencies differently than vendored gems is a separate concern. Dependabot assumes _all_ dependencies are fully vendored, or none are.

filipkis:
Well, maybe :)

In our workflow I was before relying on only things that were added to `vendor/cache` on regular `bundle install` and that didn't use to add git dependencies. And then what confused me here is that these git dependencies are not related to this PR (but since they were not cached before, they appear in every PR created by dependabot). 

So yes, I guess this is correct behaviour. And we'll just need to do the initial commit of these (or maybe add them to `.gitignore`). 

Btw. is there any chance down the road this will be configurable?

PR #3125, opened 2021-02-11
 npm: Use CLI for peer dep conflicts, and default to it without lockfile

jurre:
TLDR: Replace the native peer dependency conflict helper with the npm 7
CLI, and use that version when there is no lockfile.

I've been investigating a slow update scenario for what seems like a
fairly normal scenario. A project with 14 dependencies[1], and no
lockfile present.

From the logs I noticed a ton of requests being made to yarnpkg.com, and
at first I couldn't really place those.

Although there seems nothing weird about this setup, running an update on my
local machine is pretty slow: `7m2.660s`

My laptop is a lot beefier than the average dependabot worker, so it seems like
this update is indeed slow. It's not clear to me at all why that's the case
though, it's a completely normal setup. None of these packages even seem to
need an update, so taking 7 minutes to tell us that on a beefy laptop seems
excessive.

To figure out what's going on, I ran [stackprof](https://github.com/tmm1/stackprof)
on the dryrun for a single update (`axios`) and noticed that a _lot_ of time is
spent in `open3`, which runs our native helpers:

Screenshot 2021-02-11 at 22 26 05

From tracing the code I noticed that we're running the yarn specific
`yarn:checkPeerDependencies` when we don't have a lockfile. That explains all
the requests to yarnpkg.com. I didn't quite understand why we use the yarn
resolver and it seemed pretty slow, I tried replacing it with the npm 6
native helper, but it misses some edgecases when there is no lockfile. I
then tried to replace it with the npm 7 native helper, which runs
Arborist, but this was roughly as slow.

I wondered why this is so slow, and ran some comparisons against using
the (npm7) CLI to just perform an `npm install`, and this seemed a bit
faster and always correct.

This PR replaces the npm 7 native helper with a call to the CLI, and it
also defaults to using this version when there is no lockfile present.

This seems to consistently shave off ~20 seconds from a 54 second update for a single dependency with no lockfile:

Before:

```
real	0m54.637s
user	0m29.511s
sys	0m19.109s
```

Screenshot 2021-02-16 at 14 22 25

After:

```
real	0m35.279s
user	0m19.014s
sys	0m5.420s
```

Screenshot 2021-02-16 at 14 22 38

[1]: `axios, change-case, dayjs, jstz, logrocket, react, react-dom,
react-ga, react-scripts, sass, title-case, babel-eslint,
eslint-config-prettier, prettier`


feelepxyz (commented):
This seems like a good fix! Do you have a test case we could add for this?

jurre:
> This seems like a good fix! Do you have a test case we could add for this?

Not really sure how, the old implementation _works_, it's just slow. Since the actual work happens in the native helper, we also can't easily check for redundant network calls etc. There are tests that fail when using the npm6 implementation, so the behavior in question is covered

jurre:
Unfortunately it seems that the npm7 implementation is roughly as slow as the yarn one, and I missed this when defaulting to the npm7 version. The curious thing is that it seems roughly as slow with or without a lockfile present so I'm going to look into some generic optimizations

jurre:
This seems like a safer alternative: https://github.com/dependabot/dependabot-core/pull/3142

jurre:
I've updated the PR to use the `npm` cli vs running arborist, and parsing its error output.

This seems to consistently shave off ~20 seconds from a 54 second update for a single dependency with no lockfile:

Before:

```
real	0m54.637s
user	0m29.511s
sys	0m19.109s
```

Screenshot 2021-02-16 at 14 22 25

After:

```
real	0m35.279s
user	0m19.014s
sys	0m5.420s
```

Screenshot 2021-02-16 at 14 22 38

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -47,6 +47,15 @@ class VersionResolver
             (?<required_dep>.+?)\sbut\snone\sis\sinstalled.
           /x.freeze
```
feelepxyz, line 48: Thoughts on renaming `NPM_PEER_DEP_ERROR_REGEX` > `NPM6_PEER_DEP_ERROR_REGEX`?

npm_and_yarn/helpers/package.json
```diff
@@ -13,7 +13,6 @@
     "@npmcli/arborist": "^2.2.2",
     "detect-indent": "^6.0.0",
     "npm6": "npm:npm@6.14.11",
-    "npm7": "npm:npm@7.4.0",
```
feelepxyz, line 16: 💯  nice not having two different versions of npm 7 installed 🙌

We could also change `npm6` back to `npm` which would allow dependabot to keep it up to date as I don't think it currently handles aliased packages

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -274,7 +287,7 @@ def error_details_from_captures(captures)
             requirement_name:
               captures.fetch("required_dep").sub(/@[^@]+$/, ""),
             requirement_version:
-              captures.fetch("required_dep").split("@").last,
+              captures.fetch("required_dep").split("@").last.gsub('"', ""),
```
feelepxyz, line 290: What's the gsub fixing?

feelepxyz: ah the requirement is quoted `react@"^16.14.0"`

npm_and_yarn/helpers/package.json
jurre: Yeah, good call 👍

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
jurre: Yeah, this is a bit gnarly but it does the job

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -436,6 +450,15 @@ def run_npm_checker(path:, version:)
           end
         end
 
+        def run_npm7_checker(version:)
+          SharedHelpers.run_shell_command(
+            "npm install #{dependency.name}@#{version} --package-lock-only --ignore-scripts"
```
feelepxyz, line 455: I wonder if we need to run `--dry-run=false`? it will be set to `true` in the updater container, I've set it before on my machine or in the container and re-run to see if tests still pass, e.g. `npm config set dry-run true`

feelepxyz, line 455: Wondering if `{dependency.name}@#{version}` works for git dependencies, where we used to do this:

```js
  const source = (reqs.find((req) => req.source) || {}).source;

  if (source && source.type === "git") {
    return [`${depName}@${source.url}#${desiredVersion}`];
  } else {
    return [`${depName}@${desiredVersion}`];
  }
```

Did this in the lockfile updater: https://github.com/dependabot/dependabot-core/blob/b51a95558cd4a3dfc0b7825e964136571df304f8/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb#L248-L275

Wondering if it's worth extracting that to a helper and share the same logic to build install args?

jurre: I'd previously set it to `dry-run=true` explicitly and that worked fine to check peer deps, should we add to for clarity?

jurre: Yeah, that's a good shout 👍

feelepxyz: maybe a bit gnarly extracting it currently as it relies on `updated_version_requirement_for_dependency` which pulls stuff out of the `package.json`, maybe bring back the same logic we had in JS to build git install args and add a todo to fix?

feelepxyz: Ah right, yeah think it would be a good idea to add it explicitly as I imagine we'll drop the npm config in the updater at some point when we deprecate npm 6

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -436,6 +450,25 @@ def run_npm_checker(path:, version:)
...
+          nil
+        rescue SharedHelpers::HelperSubprocessFailed => e
+          raise if e.message.match?(NPM7_PEER_DEP_ERROR_REGEX)
+        end
+
+        def version_install_arg(version:)
+          source = dependency.requirements.find { |req| req[:source] && req[:source][:type] == "git" }
+
+          if source && source[:type] == "git"
+            "#{dependency.name}@#{source.url}##{version}"
```
feelepxyz, line 466: Won't always be a git requirement or nil? 

```suggestion
          git_req = dependency.requirements.find { |req| req[:source] && req[:source][:type] == "git" }

          if git_req
            "#{dependency.name}@#{git_req[:source][:url]}##{version}"
```

jurre: Yeah, you're right 🎉

feelepxyz, line 466: Looks like you maybe didn't push changes for this? 🤔

npm_and_yarn/helpers/package.json
feelepxyz: Not blocking to get this out, seems like a good follow up to change `npm6` across the helpers

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
feelepxyz: Wondering if it's worth setting up a test cases for this? Seems like it's not tested atm but this code-path will run for all git dependencies in npm 7 projects

jurre: 🤔 I thought I pulled in your change

jurre:
@feelepxyz applied the last review comments and rewrote the PR body to better represent the actual change

feelepxyz (approved):
🙌

PR #3129, opened 2021-02-13
elixir: support projects using Nerves extensions

fhunleth:
Nerves provides tooling for using Elixir in embedded systems. Using
Dependabot on these projects currently fails with the following error
message:

```
Dependabot can't evaluate your Elixir dependency files.
As a result, Dependabot couldn't check whether any of your dependencies
are out-of-date.

The error Dependabot encountered was:
** (Mix) Archive "nerves_bootstrap" could not be found. Please make sure the archive is installed locally.
```

Adding the `nerves_bootstrap.archive` fixes this issue. This does not
affect projects that do not use the Nerves tooling since the code in the
archive is not called in those cases.

### DIFF
--- a/hex/spec/fixtures/mixfiles/nerves
+++ b/hex/spec/fixtures/mixfiles/nerves
@@ -1,0 +1,67 @@
+defmodule NervesApp.MixProject do
+  use Mix.Project
+
+  @app :nerves_app
+  @version "0.1.0"
+  @all_targets [:rpi, :rpi0, :rpi2, :rpi3, :rpi3a, :rpi4, :bbb, :osd32mp1, :x86_64]
+
+  def project do
+    [
+      app: @app,
+      version: @version,
+      elixir: "~> 1.9",
+      archives: [nerves_bootstrap: "~> 1.10"],
+      start_permanent: Mix.env() == :prod,
+      build_embedded: true,
+      deps: deps(),
+      releases: [{@app, release()}],
+      preferred_cli_target: [run: :host, test: :host]
+    ]
+  end
+
+  # Run "mix help compile.app" to learn about applications.
+  def application do
+    [
+      mod: {NervesApp.Application, []},
+      extra_applications: [:logger, :runtime_tools]
+    ]
+  end
+
+  # Run "mix help deps" to learn about dependencies.
+  defp deps do
+    [
+      # Dependencies for all targets
+      {:nerves, "~> 1.7.4", runtime: false},
+      {:shoehorn, "~> 0.7.0"},
+      {:ring_logger, "~> 0.8.1"},
+      {:toolshed, "~> 0.2.13"},
+
+      # Dependencies for all targets except :host
+      {:nerves_runtime, "~> 0.11.3", targets: @all_targets},
+      {:nerves_pack, "~> 0.4.0", targets: @all_targets},
+
+      # Dependencies for specific targets
+      {:nerves_system_rpi, "~> 1.13", runtime: false, targets: :rpi},
+      {:nerves_system_rpi0, "~> 1.13", runtime: false, targets: :rpi0},
+      {:nerves_system_rpi2, "~> 1.13", runtime: false, targets: :rpi2},
+      {:nerves_system_rpi3, "~> 1.13", runtime: false, targets: :rpi3},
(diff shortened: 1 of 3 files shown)


feelepxyz (commented):
This looks looks like a decent fix @fhunleth! Do you have a  simple test case we could add for this? 

This would be the place to add it: https://github.com/dependabot/dependabot-core/blob/main/hex/spec/dependabot/hex/file_parser_spec.rb

We've started using a new pattern for fixtures, adding them to the `spec/fixtures/projects` folder, e.g. `spec/fixtures/projects/nerves`

Then reference it from specs like so:
```ruby
context "with a nerves project" do
  let(:files) { project_dependency_files("nerves") }

  it "parses the dependencies correctly" do
    expect(dependencies).to include(
      Dependabot::Dependency.new(
        name: "nerves",
        version: "1.7.4",
        requirements: [{
          requirement: "~> 1.7.4",
          file: "mix.exs",
          groups: [],
          source: nil
        }],
        package_manager: "hex"
      )
    )
  end
end      
```

J-Vin (approved):
😨

hex/spec/fixtures/mixfiles/nerves
```diff
@@ -0,0 +1,67 @@
+defmodule NervesApp.MixProject do
```
jurre, line 1: If you move this file to `hex/spec/fixtures/projects/nerves/mix.exs` the test should pass

fhunleth: Thanks - I missed your comment and fixed it a slightly different. I think this is consistent with the other tests in the directory, but I'll certainly change if that way is not preferred.

fhunleth:
I have an updated test, but I'm still getting the original error like the `mix archive.install hex nerves_bootstrap` isn't happening:

```
   # JSON::ParserError:
     #   767: unexpected token at '** (Mix) Archive "nerves_bootstrap" could not be found. Please make sure the archive is installed 
```

I thought that I added the line to the right place, but maybe not? Is there a way that I can verify that it took?

cblavier:
I guess the Docker image you updated to install nerves_bootstrap is not used during tests.
you should update `hex/helpers/build` 

```
diff --git a/hex/helpers/build b/hex/helpers/build
index 2f1f8f4be..bc075ac8e 100755
--- a/hex/helpers/build
+++ b/hex/helpers/build
@@ -10,7 +10,7 @@ fi

 mkdir -p $install_dir

-mix local.hex --force
+mix local.hex --force && mix archive.install hex nerves_bootstrap --force

 helpers_dir="$(dirname "${BASH_SOURCE[0]}")"
```

fhunleth:
Thanks - I just added your change. Let's see what happens...

fhunleth:
@cblavier The tests passed! Thanks!

I rebased and squashed the intermediate work.

hex/spec/fixtures/mixfiles/nerves
jurre: Yeah, we're moving away from this test setup, and prefer to set up projects that mimic an actual repo. But this is fine for now and we'll migrate all these hex specs in one go at some point

Dockerfile
```diff
@@ -167,7 +167,8 @@ RUN wget https://packages.erlang-solutions.com/erlang-solutions_1.0_all.deb \
   && echo "$ELIXIR_CHECKSUM  Precompiled.zip" | sha512sum -c - \
   && unzip -d /usr/local/elixir -x Precompiled.zip \
   && rm -f Precompiled.zip erlang-solutions_1.0_all.deb \
-  && mix local.hex --force
+  && mix local.hex --force \
+  && mix archive.install hex nerves_bootstrap --force
```
jurre, line 171: Since we're already doing this in the build script, I think we can get rid of it here right?

cblavier: if this build script is used during regular builds as well (and not only test ones) I think it is safe to remove it 👍

jurre: Yep, it is 👍 https://github.com/dependabot/dependabot-core/blob/6a682e7381893edc7df03961662db5494cd0dc2c/Dockerfile#L202

fhunleth: Updated.

feelepxyz (approved):
Thanks for the contribution and added specs 🙇

PR #3138, opened 2021-02-15
Restore npm6/7 yanked version spec

feelepxyz:
This was previously not working for npm 7 using arborist so we removed
the spec but seems to be working as expected using the cli.


npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater_spec.rb
```diff
@@ -469,6 +469,35 @@
             to raise_error(Dependabot::DependencyFileNotResolvable)
         end
       end
+
+      context "with a dependency version that can't be found" do
+        let(:files) { project_dependency_files("#{npm_version}/yanked_version") }
```
feelepxyz, line 474: Test case is from the yarn lockfile updater: https://github.com/dependabot/dependabot-core/blob/main/npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater_spec.rb#L62-L69

jurre (approved):
Nice!

PR #3139, opened 2021-02-15
Raise Azure::NotFound error when branch does not exist

milind009:
## Issue
Currently for fetch_commit function in azure_client, if the given branch does not exist, then dependabot-core raises error KeyError: Key not found 'commit' instead of Azure::NotFound error.

## Cause
The ADO api used to fetch branch details returns a status code of 400(Bad Request) instead of 404 because of which it does not raise a Azure::NotFound error.

## Solution
Added logic in fetch_commit method in Azure client class to raise Azure::NotFound whenever the incoming response is missing commit details.

## Testing
Tested this locally. Added unit test to verify this behaviour

### DIFF
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -48,11 +48,9 @@ module Dependabot
           "/_apis/git/repositories/" + source.unscoped_repo +
           "/stats/branches?name=" + branch)
 
-        commit = JSON.parse(response.body).fetch("commit", nil)
+        raise NotFound if response.status == 400
 
-        raise NotFound unless commit
-
-        commit.fetch("commitId")
+        JSON.parse(response.body).fetch("commit").fetch("commitId")
       end
 
       def fetch_default_branch(_repo)
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -48,7 +48,11 @@ module Dependabot
           "/_apis/git/repositories/" + source.unscoped_repo +
           "/stats/branches?name=" + branch)
 
-        JSON.parse(response.body).fetch("commit").fetch("commitId")
+        commit = JSON.parse(response.body).fetch("commit", nil)
+
+        raise NotFound unless commit
+
+        commit.fetch("commitId")
       end
 
       def fetch_default_branch(_repo)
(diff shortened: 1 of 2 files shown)


common/lib/dependabot/clients/azure.rb
```diff
@@ -48,7 +48,11 @@ def fetch_commit(_repo, branch)
           "/_apis/git/repositories/" + source.unscoped_repo +
           "/stats/branches?name=" + branch)
 
-        JSON.parse(response.body).fetch("commit").fetch("commitId")
+        commit = JSON.parse(response.body).fetch("commit", nil)
+
+        raise NotFound unless commit
```
jurre, line 53: Could we instead check if the `response.status` is 400 instead of checking if the commit is present?

common/spec/dependabot/clients/azure_spec.rb
```diff
@@ -66,6 +66,18 @@
         expect { subject }.to raise_error(Dependabot::Clients::Azure::NotFound)
       end
     end
+
+    context "when response body does not contain commit details" do
+      before do
+        stub_request(:get, branch_url).
+          with(basic_auth: [username, password]).
+          to_return(status: 400, body: fixture("azure", "branch_not_found.json"))
```
jurre, line 74: Is that the actual response we get back? I'd expect the body to be empty in this case? I'd also expect the API to return a 404 though so this might very well be what happens 😄

milind009: If the branch does not exist, the body actually contains exception details. Yeah agreed its kind of weird that it returns 400 and not 404. Let me update the json response file to reflect the actual response body that we get

jurre: Might be easier to just go by the status code instead of inspecting the body?

milind009: Agreed status code would be easier but i thought that checking for commit and then raising exception if not found made more sense since the function is trying to fetch commit details. Thoughts?

jurre: I think that's fine, the branch name is part of the request we're making 👍

milind009: Cool then i'll update it to raise an error when status code is 400

common/lib/dependabot/clients/azure.rb
milind009: Updated to raise error when response.status == 400

jurre (approved):

PR #3141, opened 2021-02-15
Add option to profile dry-run using Stackprof

jurre:
This makes it easy to figure out performance bottlenecks by analyzing
where time is spent in a run. The trace can be visualized by running
`stackprof --d3-flamegraph tmp/stackprof-2021-02-15-13:00.dump > tmp/flamegraph.html`
and viewing that in a browser. For some reason the built in viewer did
not work on my machine, but the d3-generated flamegraph has been clear
enough for me.

### DIFF
--- a/common/dependabot-common.gemspec
+++ b/common/dependabot-common.gemspec
@@ -42,6 +42,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "rubocop", "~> 1.9.0"
   spec.add_development_dependency "simplecov", "~> 0.21.0"
   spec.add_development_dependency "simplecov-console", "~> 0.9.1"
+  spec.add_development_dependency "stackprof", "~> 0.2.16"
   spec.add_development_dependency "vcr", "6.0.0"
   spec.add_development_dependency "webmock", "~> 3.4"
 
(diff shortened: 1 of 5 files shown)


common/dependabot-common.gemspec
```diff
@@ -42,6 +42,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "rubocop", "~> 1.9.0"
   spec.add_development_dependency "simplecov", "~> 0.21.0"
   spec.add_development_dependency "simplecov-console", "~> 0.9.1"
+  spec.add_development_dependency "stackprof", "~> 0.2.16"
```
feelepxyz, line 45: Looks like this is tripping up the python linter, might need to exclude `.bundle` when running `pyenv exec flake8`?

feelepxyz (approved):
Thanks for adding this 👍  

Python spec failure seems like a legit issue from `flake8` trying to lint stackprof files in `.bundle`

common/dependabot-common.gemspec
jurre: Ah, we run flake8 on all folders. Should be fixed 🤞

PR #3142, opened 2021-02-15
Do not run peerDependencyChecker when version has no peer deps

jurre:
For each version of each dependency that we consider upgrading, we
currently perform a fairly expensive check to see if that would result
in any peer dependency conflicts.

However, in the majority of the cases this check is not needed, as the
dependency in question has no peer dependencies.

We can run `npm info react-dom@1.0.0 peerDependencies` to figure out
which peer-dependencies the given dependency@version has. If there are
none, the command will simply return nothing.

If we _do_ have any peer dependencies, we still perform the recursive
check to figure out if the peer-dep requirements are met.

This roughly cuts the time it takes to check a `package.json` with 14
dependencies in half when running locally.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
@@ -385,6 +385,10 @@ module Dependabot
         end
 
         def run_checker(path:, version:)
+          # If there are no peer dependencies for the dependency, we don't need
+          # to run the checker, which is pretty slow.
+          return unless dependency_has_peer_dependencies?(version: version)
+
           # If there are both yarn lockfiles and npm lockfiles only run the
           # yarn updater, yarn is also used when only a package.json exists
           if lockfiles_for_path(lockfiles: dependency_files_builder.yarn_locks, path: path).any? ||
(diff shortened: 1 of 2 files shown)


npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -395,6 +399,13 @@ def run_checker(path:, version:)
           run_npm_checker(path: path, version: version)
         end
 
+        def dependency_has_peer_dependencies?(version:)
+          command = "npm info #{dependency.name}@#{version} peerDependencies"
```
feelepxyz, line 403: Just having a quick poke in npm 7 lockfiles and looks like we have `peerDependencies` in there for all existing versions (not the case for npm 6), so if there's a lockfile we could check if any dependencies have `peerDependencies` defined and run this command for the new version and short-circuit.

https://github.com/dependabot/dependabot-core/blob/e5faa799769de92975587c8bb2b65c0c374c2f6d/npm_and_yarn/spec/fixtures/projects/npm7/peer_dependency/package-lock.json#L135-L137

jurre:
Hmm, I guess this won't work, since peer dependencies are not bi-directional, meaning that `react-dom` has a peer-dependency on `react`, but not the other way around. This means that when we check if we can update `react`, this check will tell us it's fine because it has no peer-dependencies, but upgrading it would result in a conflict for `react-dom`.

There are some other potential improvements we can make, but I'm going to close out this PR

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
jurre: Yeah good call. It's _slightly_ tricky though, in that if you have an existing project and you upgrade by simply running `npm install` with npm 7, it won't actually add this `peerDependencies` info :/

PR #3153, opened 2021-02-17
Pull request updater for azure client

milind009:
# Feature
Added support for updating pull requests for `azure client (AzureDevOps)`. For updating pull requests (i.e pushing new file changes) implemented the following logic - 

- Check if the `PR & PR source branch` exists. If no, return nil else continue
- Push the new file changes onto the PR source branch.

`Caveat` - Currently the AzureDevOps PR files diff UI shows the diff in commits instead of the actual diff in the file content. Hence, there is a temporary workaround to push file changes to the source branch to ensure the reviewer sees the exact files changes. The workaround is as follows - 

- Create a new temp branch from the PR target branch and push the new file changes in that branch.
- Update the PR source branch head commit to this newly created branch head commit(hence preserving the commit history and pushing the new file changes)
- Delete the temp branch

# Testing
Added unit tests for the pull request updater implementation for azure client as well as the additional methods added in the azure client class.

### DIFF
--- a/common/dependabot-common.gemspec
+++ b/common/dependabot-common.gemspec
@@ -33,6 +33,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "pandoc-ruby", "~> 2.0"
   spec.add_dependency "parser", ">= 2.5", "< 4.0"
   spec.add_dependency "toml-rb", ">= 1.1.2", "< 3.0"
+  spec.add_dependency "uuid", "~> 2.3.9"
 
   spec.add_development_dependency "byebug", "~> 11.0"
   spec.add_development_dependency "gpgme", "~> 2.0"
--- a/common/dependabot-common.gemspec
+++ b/common/dependabot-common.gemspec
@@ -33,7 +33,6 @@ Gem::Specification.new do |spec|
   spec.add_dependency "pandoc-ruby", "~> 2.0"
   spec.add_dependency "parser", ">= 2.5", "< 4.0"
   spec.add_dependency "toml-rb", ">= 1.1.2", "< 3.0"
-  spec.add_dependency "uuid", "~> 2.3.9"
 
   spec.add_development_dependency "byebug", "~> 11.0"
   spec.add_development_dependency "gpgme", "~> 2.0"
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -183,6 +183,28 @@ module Dependabot
           "/_apis/git/repositories/" + source.unscoped_repo +
           "/pullrequests?api-version=5.0", content.to_json)
       end
+
+      def fetch_pull_request(pull_request_id)
+        response = get(source.api_endpoint +
+          source.organization + "/" + source.project +
+          "/_apis/git/pullrequests/" + pull_request_id)
+
+        JSON.parse(response.body)
+      end
+
+      def update_ref(branch_name, old_commit, new_commit)
+        content = [
+          {
+            name: "refs/heads/" + branch_name,
+            oldObjectId: old_commit,
+            newObjectId: new_commit
+          }
+        ]
+
+        post(source.api_endpoint + source.organization + "/" + source.project +
+          "/_apis/git/repositories/" + source.unscoped_repo +
+          "/refs?api-version=5.0", content.to_json)
+      end
       # rubocop:enable Metrics/ParameterLists
 
       def get(url)
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -184,7 +184,7 @@ module Dependabot
           "/pullrequests?api-version=5.0", content.to_json)
       end
 
-      def fetch_pull_request(pull_request_id)
+      def pull_request(pull_request_id)
         response = get(source.api_endpoint +
           source.organization + "/" + source.project +
           "/_apis/git/pullrequests/" + pull_request_id)
--- a/common/lib/dependabot/pull_request_updater/azure.rb
+++ b/common/lib/dependabot/pull_request_updater/azure.rb
@@ -61,7 +61,7 @@ module Dependabot
 
       def pull_request
         @pull_request ||=
-          azure_client_for_source.fetch_pull_request(pull_request_number.to_s)
+          azure_client_for_source.pull_request(pull_request_number.to_s)
       end
 
       def source_branch_name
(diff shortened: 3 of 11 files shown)


J-Vin (approved):

common/lib/dependabot/pull_request_updater/azure.rb
```diff
@@ -0,0 +1,110 @@
...
+          commit_message,
+          files,
+          nil
+        )
+
+        JSON.parse(response.body).fetch("refUpdates").first.fetch("newObjectId")
+      end
+
+      def temp_branch_name
+        "#{source_branch_name}-temp"
```
feelepxyz, line 84: Would it be worth generating a short id, e.g. `#{source_branch_name}-temp-Digest::SHA1.hexdigest(source_branch_name)[0..6]`?

Thinking of the case where two jobs try to update the same PR or someone has stashed some changes in a `-temp` version of the current branch, in which case it would be deleted. Seems unlikely but have a feeling I've created branches with temp suffixes in the past when wrangling a rebase for example.

common/lib/dependabot/clients/azure.rb
```diff
@@ -183,6 +183,28 @@ def create_pull_request(pr_name, source_branch, target_branch,
           "/_apis/git/repositories/" + source.unscoped_repo +
           "/pullrequests?api-version=5.0", content.to_json)
       end
+
+      def fetch_pull_request(pull_request_id)
```
feelepxyz, line 187: Could rename this to align with the github client:

```suggestion
      def pull_request(pull_request_id)
```

feelepxyz (commented):
Thanks for adding tests 👍  added a few minor comments, otherwise looks good to me.

milind009:
> Thanks for adding tests 👍 added a few minor comments, otherwise looks good to me.

Cool Thanks for reviewing it so promptly. Will make the suggested changes

common/lib/dependabot/pull_request_updater/azure.rb
J-Vin: Feeling your thoughts here as potential across the spectrum in various ways. You probably have a magic hat full of solutions... let me know if I can help with checklists/more specific contribution.... It's a lot for sure. Extra atd (attn to detail) is always a wise investment.

milind009: @feelepxyz Digest::SHA1.hexdigest(source_branch_name) produces consistent hashing. So for the case where we have two jobs trying to update the same PR, this might still result in issues. I was thinking of using uuid ([documentation](https://www.rubydoc.info/gems/uuid/2.3.1/UUID)) for uniquely identifying the temp branch. Thoughts?

feelepxyz: Ah yes you're right, SHA1 is stable, uuid seems like a good candidate 👍

milind009: Cool made an update accordingly

common/lib/dependabot/pull_request_updater/azure.rb
```diff
@@ -0,0 +1,113 @@
+# frozen_string_literal: true
+
+require "dependabot/clients/azure"
+require "uuid"
```
feelepxyz, line 4: You can use this module from the standard library instead:

```suggestion
require "securerandom"
```

common/dependabot-common.gemspec
```diff
@@ -33,6 +33,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "pandoc-ruby", "~> 2.0"
   spec.add_dependency "parser", ">= 2.5", "< 4.0"
   spec.add_dependency "toml-rb", ">= 1.1.2", "< 3.0"
+  spec.add_dependency "uuid", "~> 2.3.9"
```
feelepxyz, line 36: See bellow using `securerandom` instead

```suggestion
```

common/lib/dependabot/pull_request_updater/azure.rb
```diff
@@ -0,0 +1,113 @@
...
+
+      def initialize(source:, files:, base_commit:, old_commit:,
+                     credentials:, pull_request_number:)
+        @source = source
+        @files = files
+        @base_commit = base_commit
+        @old_commit = old_commit
+        @credentials = credentials
+        @pull_request_number = pull_request_number
+        @uuid = UUID.new
```
feelepxyz, line 22:
```suggestion
        @uuid = SecureRandom.uuid
```

common/lib/dependabot/pull_request_updater/azure.rb
```diff
@@ -0,0 +1,113 @@
...
+          files,
+          nil
+        )
+
+        JSON.parse(response.body).fetch("refUpdates").first.fetch("newObjectId")
+      end
+
+      def temp_branch_name
+        @temp_branch_name ||=
+          "#{source_branch_name}-temp-#{@uuid.generate[0..6]}"
```
feelepxyz, line 87:
```suggestion
          "#{source_branch_name}-temp-#{SecureRandom.uuid[0..6]}"
```

common/lib/dependabot/pull_request_updater/azure.rb
milind009: Cool Thanks for the suggestion!

common/lib/dependabot/clients/azure.rb
```diff
@@ -183,6 +183,28 @@ def create_pull_request(pr_name, source_branch, target_branch,
           "/_apis/git/repositories/" + source.unscoped_repo +
           "/pullrequests?api-version=5.0", content.to_json)
       end
+
+      def pull_request(pull_request_id)
+        response = get(source.api_endpoint +
+          source.organization + "/" + source.project +
+          "/_apis/git/pullrequests/" + pull_request_id)
+
+        JSON.parse(response.body)
```
jurre, line 192: Not blocking for now, but we might want to handle non-200 responses here

milind009: I see we have some sort of handling in the get method of the azure client. Can you specify the scenarios that we might want to cover here?

milind009:
Hi, can we complete this PR? I actually want to use this code further in our service. Let me know if i need to add anything else.

feelepxyz (approved):
LGTM 👍  we don't expose this functionality ourselves so can't really test it further.

milind009:
Cool Thanks!

PR #3157, opened 2021-02-17
Insert zero-width space in @mentions

feelepxyz:
Insert zero-width space in @mentions to prevent sending notifications to
users tagged in changelogs and release notes when email-replying to pull
requests from dependabot.

When sending an email reply to a pull request from dependabot the
contents of the pull request description including changelogs can be
included in the body of the email causing notifications to be sent to
the @mentioned users.

We've previously attempted to fix this by changing @mentions into code
blocks but it seems like the email content is treated as plain text
removing any formatting before detecting usernames.

Ref: https://github.com/dependabot/dependabot-core/pull/2982

### DIFF
--- a/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
@@ -119,7 +119,7 @@ module Dependabot
 
         def build_mention_link_text_nodes(text)
           code_node = CommonMarker::Node.new(:code)
-          code_node.string_content = text
+          code_node.string_content = insert_zero_width_space_in_mention(text)
           [code_node]
         end
 
@@ -127,9 +127,17 @@ module Dependabot
           link_node = CommonMarker::Node.new(:link)
           code_node = CommonMarker::Node.new(:code)
           link_node.url = url
-          code_node.string_content = text
+          code_node.string_content = insert_zero_width_space_in_mention(text)
           link_node.append_child(code_node)
           link_node
+        end
+
+        # NOTE: Add a zero-width space between the @ and the username to prevent
+        # email replies on dependabot pull requests triggering notifications to
+        # users who've been mentioned in changelogs etc. PR email replies parse
+        # the content of the pull request body in plain text.
+        def insert_zero_width_space_in_mention(mention)
+          mention.sub("@", "@\u200B").encode("utf-8")
         end
 
         def parent_node_link?(node)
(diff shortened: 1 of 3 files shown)


common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
```diff
@@ -118,20 +118,28 @@ def build_mention_nodes(text)
...
+          link_node.append_child(text_node)
           link_node
         end
 
+        # NOTE: Add a zero-width space between the @ and the username to prevent
+        # email replies on dependabot pull requests triggering notifications to
+        # users who've been mentioned in changelogs etc. PR email replies parse
+        # the content of the pull request body in plain text.
+        def insert_zero_width_space_in_mention(mention)
+          mention.sub("@", "@\u200B").encode("utf-8")
```
feelepxyz, line 140: Attempted this to make it more obvious we're inserting this zero-width space and make it somewhat greppable, would like to confirm this encoding won't cause any issues when creating the pr.

common/spec/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer_spec.rb
```diff
@@ -22,7 +22,7 @@
       it "sanitizes the text" do
         expect(sanitize_links_and_mentions).
           to eq("<p>Great work <a href=\"https://github.com/greysteil\">"\
-            "<code>@greysteil</code></a>!</p>\n")
+            "@​greysteil</a>!</p>\n")
```
jurre, line 25: Is the character in this string as well?

Also; wdyt about keeping the code block in there, to signal that we changed these from regular mentions?

jurre (approved):
Nice one, asked a question inline but nothing blocking

common/spec/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer_spec.rb
feelepxyz: Ah yeah good point 👍  will pull back the code block.

Yeah wen't through and pasted in a empty space in each one to make sure 😄

J-Vin: 👍

J-Vin (approved):

common/spec/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer_spec.rb
jurre: Aye yeah, maybe we can also use the `@\u200B` form for the specs? Kinda hard to spot what's going on now 😄

feelepxyz: > Aye yeah, maybe we can also use the `@\u200B` form for the specs? Kinda hard to spot what's going on now 😄

Yes great shout 👍

jurre (approved):

PR #3161, opened 2021-02-18
JS: Switch from yarn to npm for helper deps

feelepxyz:
Update npm from 7.5.3 to 7.5.4 and remove yarn as a system
dependency (was only used to install helper deps), yarn is still
installed as a helper in `npm_and_yarn/helpers/package.json`.

Use `npm ci` to install helper dependencies, this command will bail
unless package.json and package-lock.json are in sync.


jurre (approved):
Yes, nice one! I think we just need to patch up the `.github/workflows/ci.yaml` JS linting

feelepxyz:
> .github/workflows/ci.yaml

Yeah just pushed that up 👍

PR #3180, opened 2021-02-22
Convert version resolver specs to projects

jurre:
This was mostly an automated pass with some manual cleanup.

The project names are slightly gnarly, and I think we'll want to name
them ourselves eventually, but I figured that keeping them as is until
we've converted all specs will make it easier to do a single rename pass
once we're done.

### DIFF
--- a/bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
+++ b/bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
@@ -22,7 +22,6 @@ RSpec.describe Dependabot::Bundler::UpdateChecker::VersionResolver do
       latest_allowable_version: latest_allowable_version
     )
   end
-  let(:dependency_files) { [gemfile, lockfile] }
   let(:ignored_versions) { [] }
   let(:latest_allowable_version) { nil }
   let(:unlock_requirement) { false }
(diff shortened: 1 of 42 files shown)


bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
```diff
@@ -47,68 +46,42 @@
...
   describe "#latest_resolvable_version_details" do
     subject { resolver.latest_resolvable_version_details }
 
     context "with a rubygems source" do
       context "with a ~> version specified constraining the update" do
-        let(:gemfile_fixture_name) { "Gemfile" }
-        let(:lockfile_fixture_name) { "Gemfile.lock" }
         let(:requirement_string) { "~> 1.4.0" }
 
+        let(:dependency_files) { project_dependency_files("bundler1/Gemfile_Gemfile") }
```
feelepxyz, line 58: Would it work if we uniqued these names so this ends up as `Gemfile`?

feelepxyz (commented):
Nice one getting it working for bundler! Added a comment about uniquing manifest and lockfile so we don't end up with duplicates as this seems like the most common pattern, otherwise 🙌 💯

feelepxyz:
> The project names are slightly gnarly, and I think we'll want to name
> them ourselves eventually, but I figured that keeping them as is until
> we've converted all specs will make it easier to do a single rename pass
> once we're done.

Seems ok to pass through and re-name after but wondering if a `uniq` would work in most cases?

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
jurre: Yeah, but we need to also handle the scenario where there's only a Gemfile too

brrygrdn:
+1 to keeping the renaming until after, as long as we're using a consistent project naming strategy that seems fine, we _could_ add the extension as an underscore to make them a little easier to read, but ultimately I don't think it has much value.

feelepxyz (approved):
👍 :shipit:

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
jurre: Maybe we can figure out if there's a lockfile, and if not we append `_no_lockfile`

jurre:
@brrygrdn @feelepxyz I've improved the fixture names, thanks for calling it out, this is much nicer and it wasn't much effort

brrygrdn (approved):

PR #3188, opened 2021-02-24
Bundler: Convert FileParser specs to projects

jurre:
Mostly using https://github.com/dependabot/dependabot-core/pull/3027
with some manual fixing afterwards.

### DIFF
--- a/bundler/spec/dependabot/bundler/file_parser_spec.rb
+++ b/bundler/spec/dependabot/bundler/file_parser_spec.rb
@@ -9,16 +9,9 @@ require_common_spec "file_parsers/shared_examples_for_file_parsers"
 RSpec.describe Dependabot::Bundler::FileParser do
   it_behaves_like "a dependency file parser"
 
-  let(:files) { [gemfile, lockfile] }
-  let(:gemfile) do
-    Dependabot::DependencyFile.new(name: "Gemfile", content: gemfile_body)
-  end
-  let(:lockfile) do
-    Dependabot::DependencyFile.new(name: "Gemfile.lock", content: lockfile_body)
-  end
   let(:parser) do
     described_class.new(
-      dependency_files: files,
+      dependency_files: dependency_files,
       source: source,
       reject_external_code: reject_external_code
     )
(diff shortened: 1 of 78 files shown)


feelepxyz:
Looks like there are a couple spec failures and seeing this in the error output `ruby -rubygems -e 'require "jekyll-import";` is this new? https://github.com/dependabot/dependabot-core/pull/3188/checks?check_run_id=1969526284

bundler/spec/dependabot/bundler/file_parser_spec.rb
```diff
@@ -153,17 +136,7 @@
...
-      end
-      let(:lockfile) do
-        Dependabot::DependencyFile.new(
-          name: "gems.locked",
-          content: lockfile_body
-        )
-      end
-      let(:gemfile_fixture_name) { "version_specified" }
-      let(:lockfile_fixture_name) { "bundler_2.lock" }
+      let(:dependency_files) { project_dependency_files("bundler1/version_specified_bundler_2") }
```
feelepxyz, line 139: Not for now but adding a mental note to go through fixtures once we add support for bundler 2 and move to `bundler2/..`

bundler/spec/dependabot/bundler/file_parser_spec.rb
```diff
@@ -373,14 +335,7 @@
...
       context "that comes from a .specification file" do
-        let(:files) { [gemfile, lockfile, specification] }
-        let(:specification) do
-          Dependabot::DependencyFile.new(
-            name: "plugins/example/.specification",
-            content: fixture("ruby", "specifications", "statesman"),
-            support_file: true
-          )
-        end
+        let(:dependency_files) { project_dependency_files("bundler1/version_specified_gemfile_specification") }
```
feelepxyz, line 338: does this not need the same `support_file = true` setup?

bundler/spec/dependabot/bundler/file_parser_spec.rb
```diff
@@ -507,44 +446,20 @@
     end
 
     context "with a Gemfile that includes a file with require_relative" do
-      let(:files) { [gemfile, lockfile, required_file] }
-      let(:gemfile_fixture_name) { "includes_require_relative" }
-      let(:lockfile_fixture_name) { "Gemfile.lock" }
-      let(:required_file) do
-        Dependabot::DependencyFile.new(
-          name: "../some_other_file.rb",
```
feelepxyz, line 515: Looks like this was dropped from the fixture folder but not sure how to do this without creating the dependency file like this. Assuming this is returned from the file fetcher if setting up updates in a nested directory with a Gemfile that requires a file from the parent directory.

feelepxyz (approved):
Added some questions around the spec failures. LGTM once specs are 🟢

bundler/spec/dependabot/bundler/file_parser_spec.rb
jurre: Yeah this now lives in the `bundler1` folder which isn't great but works

bundler/spec/dependabot/bundler/file_parser_spec.rb
jurre: It might? I thought all tests were passing locally but it seems I may have missed one or two, this might be one of them!

bundler/spec/dependabot/bundler/file_parser_spec.rb
feelepxyz: Could you nest the project within the project and move the required file down?

feelepxyz: e.g. `project_dependency_files("bundler1/includes_require_relative_gemfile/nested")`

jurre: Yeah that's what I was trying to get at, I think that'll work 👍

PR #3191, opened 2021-02-24
PullRequestCreator::Message

thepwagner:
Introduce `Dependabot::PullRequestCreator::Message` as an alternative to the `Dependabot::PullRequestCreator::MessageBuilder` used to generate pull request bodies. It's intended that `Dependabot::PullRequestCreator::Message` be a memoized `Dependabot::PullRequestCreator::MessageBuilder` result, but that's not enforced.

Allow `Dependabot::PullRequestCreator` to accept a `Message` argument, and if provided prefer that static result to invoking the `MessageBuilder` for dynamic calculation.

Add `FileUpdater.pr_message` as a means of resolving a `Message` corresponding to a dependency update that was processed. This is retrofitted with `Dependabot::PullRequestCreator::MessageBuilder` so that updater implementations are unchanged, but can customize it in the future (cc @jurre).

Finally, add a `--pull-request` argument to the `dry-run.rb` script, so that developers can easily exercise this shiny new path, and get used to having PR metadata derived in the same environment as the update job.

The theme of this change is allowing the `PullRequestCreator` to use a precalculated PR message, e.g. so when a common package is updated Dependabot doesn't need to recalculate this N times.

### DIFF
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -700,7 +700,7 @@ dependencies.each do |dep|
       files: updated_files,
       credentials: $options[:credentials],
       source: $source,
-    )
+    ).message
     puts "Pull Request Title: #{msg.pr_name}"
     puts "--description--\n#{msg.pr_message}\n--/description--"
     puts "--commit--\n#{msg.commit_message}\n--/commit--"
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -109,7 +109,8 @@ $options = {
   commit: nil,
   updater_options: {},
   security_advisories: [],
-  security_updates_only: false
+  security_updates_only: false,
+  pull_request: false
 }
 
 unless ENV["LOCAL_GITHUB_ACCESS_TOKEN"].to_s.strip.empty?
@@ -206,6 +207,11 @@ option_parse = OptionParser.new do |opts|
   opts.on("--profile",
           "Profile using Stackprof. Output in `tmp/stackprof-<datetime>.dump`") do
     $options[:profile] = true
+  end
+
+  opts.on("--pull-request",
+    "Output pull request information: title, description") do
+    $options[:pull_request] = true
   end
 end
 
@@ -430,7 +436,7 @@ end
 
 StackProf.start(raw: true) if $options[:profile]
 
-source = Dependabot::Source.new(
+$source = Dependabot::Source.new(
   provider: $options[:provider],
   repo: $repo_name,
   directory: $options[:directory],
@@ -446,7 +452,7 @@ if $options[:clone] || always_clone
 end
 
 fetcher = Dependabot::FileFetchers.for_package_manager($package_manager).
-          new(source: source, credentials: $options[:credentials],
+          new(source: $source, credentials: $options[:credentials],
               repo_contents_path: $repo_contents_path)
 $files = if $options[:clone] || always_clone
            fetcher.clone_repo_contents
@@ -462,7 +468,7 @@ puts "=> parsing dependency files"
 parser = Dependabot::FileParsers.for_package_manager($package_manager).new(
   dependency_files: $files,
   repo_contents_path: $repo_contents_path,
-  source: source,
+  source: $source,
   credentials: $options[:credentials],
   reject_external_code: $options[:reject_external_code],
 )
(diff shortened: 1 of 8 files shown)


common/lib/dependabot/file_updaters/base.rb
```diff
@@ -25,6 +26,21 @@ def updated_dependency_files
         raise NotImplementedError
       end
 
+      def pr_message
+        Dependabot::PullRequestCreator::MessageBuilder.new(
+          dependencies: dependencies,
+          files: updated_dependency_files,
+          credentials: credentials,
+          source: @source,
+          # TODO: from config file:
```
thepwagner, line 35: Question: how could `FileUpdater::Base` locate itself in the config file here?

Loading the config file seems doable: if we tell `FileFetcher::Base` that `.github/dependabot.ya?ml` is a dependency it will just appear in the `dependency_files` argument - right?

But within that file, we need to find the `.config[]` entry with the right `package-ecosystem` and `directory` that matches what's currently running:
* Mapping `directory` is `@source.directory`, easy peasy.
* Mapping `package-ecosystem` is _gnarly_ as the string in the config file doesn't correspond to `dependabot-core` (e.g. `gomod` in the config vs `go_modules` in dependabot-core).

Is config file parsing a pipe dream? We may want to find another way of sourcing these values, then accept `commit_message_options` as an initialization argument...

feelepxyz: Yeah seems like a heavy lift to pull in config file parsing to get these options pulled in, could we pass them in as arguments for now when calling `pr_message(prefix: "prod", prefix_development: "dev", include: "scope")`?

common/lib/dependabot/file_updaters/base.rb
```diff
@@ -25,6 +26,21 @@ def updated_dependency_files
         raise NotImplementedError
       end
 
+      def pr_message
+        Dependabot::PullRequestCreator::MessageBuilder.new(
```
feelepxyz, line 30: Slight smell nesting this inside the FileUpdater as it doesn't update any files, thoughts on just calling this class directly instead?

feelepxyz: That way also easier to pass the options in directly from the updater until we figure out how to dismantle the updater and move more stuff into core.

common/lib/dependabot/pull_request_creator.rb
```diff
@@ -197,7 +198,7 @@ def codecommit_creator
     end
 
     def message_builder
-      @message_builder ||
+      @message ||
```
feelepxyz, line 201: Pretty sure this should have been memoized with `||=`, when it's not set we'll initialise and build 3 times

Maybe also clearer if we renamed the method:

```
def message
  @message ||=
    MessageBuilder.new...
end
```

Not sure if it's worth it but we could also add a `message` method on the Builder which wraps the three arguments in the `Dependabot::PullRequestCreator::Message` class so we always access things through the same class (interfaces would be neat..)

common/lib/dependabot/file_updaters/base.rb
thepwagner: SGTM, this was based on @jurre's  intuition that while it's _cleaner_ to instantiate this class in our closed-source wrapper, that doesn't leave any sense in the open-source `dependabot-core` how the components are linked together.

It's also plausible that the `FileUpdater` could share state from the update to build the PR message: e.g. parsing `node_modules/foo/package.json` instead of re-fetching `foo` from the registry via the `MetadataFinder`.

common/lib/dependabot/pull_request_creator.rb
thepwagner: > Not sure if it's worth it but we could also add a message method on the Builder which wraps the three arguments

I'd done that, but I didn't dig amending the `PullRequestCreator::MessageBuilder` tests enough to keep it around. I'll bring it back.

common/lib/dependabot/file_updaters/base.rb
thepwagner: Yep - even easier if we decouple from `FileUpdater` per https://github.com/dependabot/dependabot-core/pull/3191#discussion_r582810432

common/lib/dependabot/pull_request_creator.rb
thepwagner: > Pretty sure this should have been memoized with ||=, when it's not set we'll initialise and build 3 times
> rename the method

Good shout, done and tested in  https://github.com/dependabot/dependabot-core/pull/3191/commits/b55cbc8f3c46b7e4a1e75163c45f857845c939b8

> add a message method on the Builder 

Added in https://github.com/dependabot/dependabot-core/pull/3191/commits/ea09d928d1ca0760039dd23da1e81b128e358aae , incl the weak test that initially made me question the value.

common/lib/dependabot/file_updaters/base.rb
thepwagner: @jurre : I've reverted the changes to `FileUpdater::Base` in https://github.com/dependabot/dependabot-core/pull/3191/commits/40a357565cc70b0fdc4cbf65f72ee82a5a17597b .

I dig this more, I think _feelepxyz digs this more, but it does leave the `MessageBuilder` as a bit of detached magic: WDYT?

common/lib/dependabot/pull_request_creator.rb
```diff
@@ -187,17 +188,17 @@ def codecommit_creator
...
         labeler: labeler,
         require_up_to_date_base: require_up_to_date_base?
       )
     end
 
-    def message_builder
-      @message_builder ||
+    def message
+      @message ||=
         MessageBuilder.new(
```
feelepxyz, line 202: Should this return `MessageBuilder.new(...).message`? Looks like that's currently unused

bin/dry-run.rb
```diff
@@ -691,6 +693,18 @@ def security_fix?(dependency)
...
     end
   end
+
+  if $options[:pull_request]
+    msg =  Dependabot::PullRequestCreator::MessageBuilder.new(
+      dependencies: updated_deps,
+      files: updated_files,
+      credentials: $options[:credentials],
+      source: $source,
+    )
```
feelepxyz, line 703:
```suggestion
    ).message
```

common/lib/dependabot/pull_request_creator.rb
thepwagner: It doesn't need to, since `Message` and `MessageBuilder` satisfy the same interface loose typing does the rest? 🦆==🦆 ?

I'm impartial but this suggestion appears twice so I'll take it as a Rubyism.

feelepxyz (approved):
LGTM 👍

PR #3195, opened 2021-02-25
haskell (wip) - fixes #2745

### LINKED ISSUE #2745: Haskell support
The Haskell ecosystem often suffers from outdated package bounds. Various attempts have been made to improve the situation, but ultimately developers need a convenient method to update their bounds.

Haskell uses [PVP](https://pvp.haskell.org/) for versioning (which is somewhat similar to semver). It then uses [Cabal](https://github.com/haskell/cabal/) for package and version management (similar to rusts cargo). Packages are shared and uploaded on [hackage](http://hackage.haskell.org/).

I think the main hurdle of adding support here is that the cabal file format is custom. Obviously there exists a haskell parser in the [Cabal library](http://hackage.haskell.org/package/Cabal). There also seems to be *some* [documentation on the file format](https://cabal.readthedocs.io/en/latest/buildinfo-fields-reference.html). I'm not sure if that's enough to re-implement a parser or if a regexp solution would be possible.

KiaraGrouwstra:
I tried to start on adding support for Haskell.
I'm mainly still working on the `update_checker` and `file_updater` parts.


feelepxyz:
@KiaraGrouwstra thank you for taking the time to start adding support for Haskell. We really appreciate the effort to make Dependabot better for the community.

Unfortunately, we are not accepting new ecosystems into core at the moment. We are focused on making some improvements into how we handle extensibility before we add any more ecosystems. We will still be improving and upgrading support for existing ecosystems.

Please check the [Contribution guidelines](https://github.com/dependabot/dependabot-core/blob/main/CONTRIBUTING.md#contributing-new-ecosystems) for more information.

KiaraGrouwstra:
'until at least June 2021' huh. that sounds manageable so far in the grand scheme of things -- thanks!
as this PR will not require your further attention at this point, I'll close it for the time being then.

PR #3196, opened 2021-02-25
Run Bundler v1 native helpers with explicit version

feelepxyz:
This runs existing bundler 1 native helpers by explicitly setting a `BUNDLER_VERSION` env var as well as pointing the `BUNDLE_GEMFILE` to `bundler/helpers/v1/Gemfile` which has been installed with `BUNDLER_VERSION=1 bundle install` setting `BUNDLED WITH` to the latest installed v1 version.

I've verified this setup works with bundler 2 installed in the core container and installing core's gems with v2 and running helpers with v1, I've got the native helper tests working with a setup that extends on this PR: https://github.com/dependabot/dependabot-core/pull/3211/files

A follow on ship from this would be running native helper specs with bundler 1 followed by actually installing bundler 2 and installing core's gems with it.

This also moves the native helpers to `v1` folder making room for a `v2` folder.

## Release plan

- [ ] Deploy to staging and verify a couple of different updates, including private registry/repo


Dockerfile
```diff
@@ -61,6 +61,8 @@ RUN apt-get install -y software-properties-common \
   && gem update --system 3.0.3 \
   && gem install bundler -v 1.17.3 --no-document
 
+# NOTE: For some reason this isn't succeeding if installed in the above RUN
+RUN gem install bundler -v 2.2.11 --no-document
```
feelepxyz, line 65: Installed and used to `bundle install` core's dependencies but not used in updates.

bundler/helpers/Gemfile.lock
```diff
@@ -0,0 +1,9 @@
+GEM
+  remote: https://rubygems.org/
+  specs:
+
+PLATFORMS
+  ruby
+  x86_64-darwin-19
+
+DEPENDENCIES
```
feelepxyz, line 9: This is funky, we need to commit this without a `BUNDLED WITH` as it's loaded from setting `BUNDLE_GEMFILE` to the above `Gemfile` the `when running `bundle exec` on customers manifests, if we don't set this bundler complains that the gems defined in the customers gemfile are not installed.

Similarly, `require "file"` used to happen here https://github.com/dependabot/dependabot-core/blob/b53df52e75d5eb1c585d6339b90f03e676e5ca73/common/dependabot-common.gemspec#L3 so worked if `File` was used in a customers `.gemspec` but not required 😑

bundler/helpers/Gemfile
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
```
feelepxyz, line 1: nest this under `helpers/v1` to make it explicit which version is `BUNDLED WITH`

bundler/lib/dependabot/bundler/native_helpers.rb
```diff
@@ -1,10 +1,36 @@
...
+            # Required to git clone to a writeable folder
+            "GEM_HOME" => ENV["GEM_HOME"],
+            # Required by git fetch
+            "SSH_AUTH_SOCK" => ENV["SSH_AUTH_SOCK"]
+          }
+        )
+      end
+
+      def self.helper_path(bundler_version: "v1")
+        "ruby #{File.join(native_helpers_root, bundler_version, 'run.rb')}"
```
feelepxyz, line 33: @jurre running without `bundle exec` worked and removed the need for an empty `Gemfile` 🎉

bundler/helpers/v1/build
```diff
@@ -8,6 +8,10 @@ if [ -z "$install_dir" ]; then
   exit 1
 fi
 
+if [ ! -d "$install_dir" ]; then
+  mkdir -p "$install_dir"
```
feelepxyz, line 12: creates the nested `v1` in `/opt/bundler`

bundler/lib/dependabot/bundler/native_helpers.rb
```diff
@@ -1,10 +1,36 @@
...
 
+require "dependabot/shared_helpers"
+
 module Dependabot
   module Bundler
     module NativeHelpers
-      def self.helper_path
-        "bundle exec ruby #{File.join(native_helpers_root, 'run.rb')}"
+      def self.run_bundler_subprocess(function:, args:, bundler_version:)
+        bundler_env_version = bundler_version.tr("v", "")
```
feelepxyz, line 9: This feels a bit funky, might be easier to do the version detection here instead

feelepxyz:
@jurre @brrygrdn got specs passing with just running the native helpers with a re-set ENV and setting `BUNDLER_VERSION`, this worked with bundler 2 installed but the native helper specs broke so thinking it makes more sense to ship this incrementally and then follow up with a PR  to install bundler 2 but not use and fix the native helper specs.

brrygrdn (commented):
Thanks for breaking this change out, it's very easy to follow the restructuring 👍🏻 

I only have a couple of questions

common/lib/dependabot/shared_helpers.rb
```diff
@@ -69,20 +69,24 @@ def self.escape_command(command)
...
+                                   unsetenv_others: false)
       start = Time.now
       stdin_data = JSON.dump(function: function, args: args)
       cmd = allow_unsafe_shell_command ? command : escape_command(command)
       env_cmd = [env, cmd].compact
       if ENV["DEBUG_FUNCTION"] == function
         escaped_stdin_data = stdin_data.gsub("\"", "\\\"")
-        puts "$ cd #{Dir.pwd} && echo \"#{escaped_stdin_data}\" | #{env_cmd.join(' ')}"
+        env_keys = env ? env.map { |k, v| "#{k}=#{v}" } : ""
+        puts "$ cd #{Dir.pwd} && echo \"#{escaped_stdin_data}\" | #{env_keys} #{cmd}"
```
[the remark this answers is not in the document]
feelepxyz: This path should only be used when you pass `DEBUG_FUNCTION=somefunc` as env var when running specs/dry-run and puts out the command that's being executed by ruby so you can cd into the temp dir and run the same thing in bash. The previous way of concatenating `env` and `cmd` didn't work when actually passing in a `env` hash.

The env_cmd arg still works for `Open3` but now the bash version flattens the `env` hash to `ENV_KEY=value ./func`.

bundler/lib/dependabot/bundler/native_helpers.rb
```diff
@@ -1,10 +1,35 @@
...
+            "GEM_HOME" => ENV["GEM_HOME"],
+            # Required by git fetch
+            "SSH_AUTH_SOCK" => ENV["SSH_AUTH_SOCK"]
+          }
+        )
+      end
+
+      def self.helper_path(bundler_version:)
+        native_helper_version = "v#{bundler_version}"
+        "ruby #{File.join(native_helpers_root, native_helper_version, 'run.rb')}"
```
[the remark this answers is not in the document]
feelepxyz: 👍  intentionally removing as `bundle exec` needs to be run in a folder with a `Gemfile` source (or point to one with a env var) that has resolved dependencies, we currently don't rely on any dependencies in the native helpers so this just got in the way of running the native helpers with a custom bundler version. Bundler refuses to run `BUNDLER_VERSION=1 cmd` if the `Gemfile.lock` that we're pointing to has been resolved with `BUNDLED WITH 2.x` and vice versa.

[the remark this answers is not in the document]
feelepxyz: Previously, when we executed `bundle exec cmd` it inherited all existing `ENV` values so ended up pointing to  `./bundler/Gemfile` as we ran the specs with `bundle exec rspec spec` and inherited the same `BUNDLER_VERSION` `ENV` var which I couldn't re-set without setting `unsetenv_others: true`.

Similarly, when run from the updater it would inherit the `BUNDLED WITH` from the updaters `Gemfile.lock`: https://github.com/dependabot/dependabot-updater/blob/main/bin/run#L21

bundler/lib/dependabot/bundler/native_helpers.rb
```diff
@@ -1,10 +1,35 @@
...
+            # Bundler will pick the matching installed major version
+            "BUNDLER_VERSION" => bundler_version,
+            # Required to find the ruby bin
+            "PATH" => ENV["PATH"],
+            # # Requried to create tmp directories in a writeable folder
+            "HOME" => ENV["HOME"],
+            # Required to git clone to a writeable folder
+            "GEM_HOME" => ENV["GEM_HOME"],
+            # Required by git fetch
+            "SSH_AUTH_SOCK" => ENV["SSH_AUTH_SOCK"]
```
feelepxyz, line 25: Oh we probably need for forward some env from the runner, e.g.

SSL_CERT_FILE
http_proxy
HTTP_PROXY
https_proxy
HTTPS_PROXY

common/lib/dependabot/shared_helpers.rb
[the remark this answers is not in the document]
feelepxyz: I attempted to clarify this a bit with the above comment and moving `env_cmd` down to where it's used

common/lib/dependabot/shared_helpers.rb
```diff
@@ -75,13 +75,17 @@ def self.run_helper_subprocess(command:, function:, args:, env: nil,
...
       cmd = allow_unsafe_shell_command ? command : escape_command(command)
-      env_cmd = [env, cmd].compact
+
+      # NOTE: For debugging native helpers in specs and dry-run: outputs the
+      # bash command to run in the tmp directory created by
+      # in_a_temporary_directory
       if ENV["DEBUG_FUNCTION"] == function
-        escaped_stdin_data = stdin_data.gsub("\"", "\\\"")
-        puts "$ cd #{Dir.pwd} && echo \"#{escaped_stdin_data}\" | #{env_cmd.join(' ')}"
+        puts helper_subprocess_bash_command(stdin_data: stdin_data, command: cmd, env: env)
```
feelepxyz, line 83: @brrygrdn moved this into it's own method

Example output, if you run `DEBUG_FUNCTION=parsed_gemspec bundle exec rspec spec/dependabot/bundler/file_parser_spec.rb:685` you'll see this in console:

```
$ cd /Users/feelepxyz/code/dependabot-core/bundler/tmp/dependabot_20210302-86366-1xt144c && echo "{\"function\":\"parsed_gemspec\",\"args\":{\"gemspec_name\":\"example.gemspec\",\"lockfile_name\":null,\"dir\":\"/Users/feelepxyz/code/dependabot-core/bundler/tmp/dependabot_20210302-86366-1xt144c\"}}" | BUNDLER_VERSION=1 BUNDLE_GEMFILE=/Users/feelepxyz/code/dependabot-core/bundler/lib/dependabot/bundler/../../../helpers/v1/Gemfile bundle exec ruby /Users/feelepxyz/code/dependabot-core/bundler/lib/dependabot/bundler/../../../helpers/v1/ru
```

bundler/helpers/v1/Gemfile
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+# TODO: Look into removing this. "file" used to get required from common's
+# gemspec so we've added it here for backwards compatability during bundler 2
+# rollout.
+#
+# NOTE: If we don't require it and a customers `.gemspec` uses File without
```
brrygrdn, line 7:
```suggestion
# NOTE: If we don't require it and a customers `.gemspec` uses Find without
```

bundler/helpers/v1/Gemfile
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+# TODO: Look into removing this. "file" used to get required from common's
```
brrygrdn, line 3:
```suggestion
# TODO: Look into removing this. "find" used to get required from common's
```

bundler/lib/dependabot/bundler/native_helpers.rb
```diff
@@ -1,10 +1,39 @@
...
+            # Force bundler to use the helper Gemfile that has been bundled with
+            # v1, otherwise it will point to core's bundler/Gemfile which will
+            # be bundled with v2 once it's installed
+            "BUNDLE_GEMFILE" => File.join(versioned_helper_path(bundler_version: bundler_version), "Gemfile"),
+            # Unset ruby env set by running dependabot-core with bundle exec,
+            # forcing bundler to reset them from helpers/v1
+            "RUBYLIB" => nil,
+            "RUBYOPT" => nil,
+            "GEM_PATH" => nil,
+            "GEM_HOME" => nil
```
feelepxyz, line 25: This is kinda funky but seems to work, verified that these env vars are not set in the container so get set by bundler/ruby when running core with `bundle exec rspec/updater run`

brrygrdn, line 25: This seems like a fair workaround, it feels like the cost of doing business since we're running bundler to run core to run bundler ♻️  😅

common/lib/dependabot/shared_helpers.rb
brrygrdn: Nice, thanks for clarifying this 🙇🏻

brrygrdn (commented):
A couple of copy edits, otherwise this looks 💯 

I don't feel bad unsetting specific ruby/bundler envvars given the circumstances. It's well called out and doesn't feel brittle in the sense those are pretty fixed points in the ecosystem.

brrygrdn (approved):
🚀

PR #3197, opened 2021-02-25
Set bitbucket labeler to nil

kobellem:
As bitbucket does not support the labeler, @iinuwa tried to fix it in https://github.com/dependabot/dependabot-core/commit/52db87ebd750452d500e713bc85b745bc84f4352 by setting the default labeler in the constructor to nil. However from my testing this does not fix it, as we're still injecting the default labeler here.

### DIFF
--- a/common/lib/dependabot/pull_request_creator.rb
+++ b/common/lib/dependabot/pull_request_creator.rb
@@ -175,7 +175,7 @@ module Dependabot
         pr_description: message_builder.pr_message,
         pr_name: message_builder.pr_name,
         author_details: author_details,
-        labeler: labeler,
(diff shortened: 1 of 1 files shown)


xlgmokha:
Thank you for this submission.

It would be helpful to capture the runtime characteristics that produce this defect so that we don't re-introduce a regression. Is it possible to reproduce this defect in a unit test?

bmc08gt:
Hitting this as well attempting to use the generic-update-script

jeffwidman:
Closing as this was handled in https://github.com/dependabot/dependabot-core/pull/4608 (many thanks @stefangr!)

PR #3211, opened 2021-03-01
Bundler: Run v1 native helpers with bundler v1

feelepxyz:
This runs existing bundler v1 native helpers with `BUNDLER_VERSION=1`, not falling back to the default version, which will be v2 once it's installed.


.github/workflows/ci.yml
```diff
@@ -102,6 +102,10 @@ jobs:
         run: |
           docker run --rm "$CORE_CI_IMAGE" bash -c "cd /opt/npm_and_yarn && npm run lint"
           docker run --rm "$CORE_CI_IMAGE" bash -c "cd /opt/npm_and_yarn && npm test"
+      - name: Run bundler v1 native helper specs
+        if: matrix.suite == 'bundler'
+        run: |
+          docker run --rm "$CORE_CI_IMAGE" bash -c "cd /opt/bundler/v1 && bundle exec rspec spec"
```
feelepxyz, line 108: We probably want to explicitly set `BUNDLER_VERSION=1 bundle exec rspec spec` to make sure we haven't installed the gems with a different bundler version

bundler/lib/dependabot/bundler/native_helpers.rb
```diff
@@ -11,31 +11,28 @@ def self.run_bundler_subprocess(function:, args:, bundler_version:)
...
-            "HTTP_PROXY" => ENV["HTTP_PROXY"],
-            "https_proxy" => ENV["https_proxy"],
-            "HTTPS_PROXY" => ENV["HTTPS_PROXY"]
+            # Force bundler to use the helper gemfile that has been bundled with v1
+            "BUNDLE_GEMFILE" => File.join(versioned_helper_path(bundler_version: bundler_version), "Gemfile"),
+            # Unset ruby env set by running bundle exec from dependabot-core/bundler
+            "RUBYLIB" => nil,
+            "RUBYOPT" => nil,
+            "GEM_PATH" => nil,
+            "GEM_HOME" => nil
```
feelepxyz, line 24: @brrygrdn think this might work instead of unsetting the env, had to unset the the above, otherwise bundler got real confused about where it was loading gems and complained the helper gems hand't been installed when running `bundle exec`

Installed the gems with an explicit bundler version here: https://github.com/dependabot/dependabot-core/pull/3211/files#diff-a52f9842f7c3e55a30d022d6bc436a99d82b59aa42b14ccc61ca8aa8c3c2312bR27

This is slightly wasteful atm as they are only used for tests 😕

Dockerfile
```diff
@@ -54,12 +54,13 @@ RUN apt-get update \
...
 ENV BUNDLE_SILENCE_ROOT_WARNING=1
-RUN apt-get install -y software-properties-common \
+RUN apt-get update && apt-get install -y software-properties-common \
   && apt-add-repository ppa:brightbox/ruby-ng \
   && apt-get update \
   && apt-get install -y ruby2.6 ruby2.6-dev \
   && gem update --system 3.0.3 \
-  && gem install bundler -v 1.17.3 --no-document
+  && gem install bundler -v 1.17.3 --no-document \
+  && gem install bundler -v 2.2.13 --no-document
```
feelepxyz, line 63: @brrygrdn looks like all specs pass with v2 installed 🎉  going to jank it out from here so we can deploy it separately to this change

bundler/helpers/v1/build
```diff
@@ -21,4 +21,4 @@ cd "$install_dir"
 
 # NOTE: Sets `BUNDLED WITH` to match the installed v1 version in Gemfile.lock
 # forcing specs and native helpers to run with the same version
```
feelepxyz, line 23:
```suggestion
# forcing native helpers to run with the same version
```

brrygrdn (approved):
LGTM :shipit: !

.github/workflows/ci.yml
```diff
@@ -102,6 +102,11 @@ jobs:
         run: |
           docker run --rm "$CORE_CI_IMAGE" bash -c "cd /opt/npm_and_yarn && npm run lint"
           docker run --rm "$CORE_CI_IMAGE" bash -c "cd /opt/npm_and_yarn && npm test"
+      - name: Run bundler v1 native helper specs
+        if: matrix.suite == 'bundler'
```
brrygrdn, line 106: Not sure if it's a big deal right for this PR, but I guess we should add `bundler2` to the matrix rather than mess with this existing key?

feelepxyz: Ah yes, was also thinking we could run them in in the same suite a second step as I think they run quickly. Each matrix ends up building the entire image as it's set up currently.

brrygrdn: Good call, that makes sense!

PR #3220, opened 2021-03-02
Fix gemspec sanitization bug when heredoc has methods chained onto it

jasonrudolph:
When dependabot is processing a gemspec, it performs some sanitization of the gemspec's content. In most situations, it handles a heredoc by replacing the heredoc's content as seen in [this test](https://github.com/dependabot/dependabot-core/blob/e25a21c558ba3c52d497f3536fa66ea15451511e/bundler/spec/dependabot/bundler/file_updater/gemspec_sanitizer_spec.rb#L123-L138).

Given a gemspec like this:

```rb
Spec.new do |s|
  s.version = "0.1.0"
  s.post_install_message = <<~DESCRIPTION
    My description
  DESCRIPTION
end
```

The sanitized output should look like this:

```rb
Spec.new do |s|
  s.version = "0.1.0"
  "sanitized"
end
```

That's the current behavior, and it meets our needs. However, if the heredoc has methods chained onto it, the sanitzation currently produces invalid Ruby code.

Given a gemspec like this:

```rb
Spec.new do |s|
  s.version = "0.1.0"
  s.post_install_message = <<~DESCRIPTION.strip.downcase
    My description
  DESCRIPTION
end
```

We should get the same sanitized output that we got in the previous example, but we currently get the following output instead:

```rb
Spec.new do |s|
  s.version = "0.1.0"
  "sanitized"
    My description
  DESCRIPTION
end
```

Later on in the dependabot-core workflow, when we ask bundler to evaluate the sanitized gemspec, it fails with a parser error because the output above is invalid Ruby code.

### TODO

- [x] Add a failing test to demonstrate the bug
- [x] Fix the bug


feelepxyz:
Looks like this is where we try and do the sanitisation: https://github.com/dependabot/dependabot-core/blob/main/bundler/lib/dependabot/bundler/file_updater/gemspec_sanitizer.rb#L237-L242 haven't worked on this code but would be up for spending a short block on trying to fix it.

bundler/lib/dependabot/bundler/file_updater/gemspec_sanitizer.rb
```diff
@@ -249,6 +246,30 @@ def remove_unnecessary_assignments(node)
...
+              next unless child.is_a?(Parser::AST::Node)
+
+              return child.location.heredoc_end if child.location.respond_to?(:heredoc_end)
+
+              range = find_heredoc_end_range(child)
+              return range if range
+            end
+
+            nil
+          end
```
jasonrudolph, line 271: We (@feelepxyz and I) suspect that there's a more readable way to implement this. We explored a few options, but the other options were a bit harder to follow in our opinion. So, if you have ideas for an alternative implementation that would be easier to read/maintain, definitely let us know. :bow:

jasonrudolph:
The npm_and_yarn CI job is failing due to a flaky test. That flaky test is fixed in https://github.com/dependabot/dependabot-core/pull/3229. Once https://github.com/dependabot/dependabot-core/pull/3229 is merged, I'll merge `main` into this branch, and that should give us a green build.

feelepxyz:
Just merged #3229 👌

xlgmokha (approved):

PR #3222, opened 2021-03-03
Revert: Run Bundler v1 native helpers with explicit version

feelepxyz:
This reverts https://github.com/dependabot/dependabot-core/pull/3196

We've run into some issues in production where bundle installs fail when fetching git dependencies as they try to write to a read-only file path: 

```
There was an error while trying to write to`/var/lib/gems/2.6.0/cache/bundler/git`. It is likely that you need to grant write permissions for that path.
```

We're rolling this back to put core in good state while we figure out an approach that will allow us to run multiple bundler versions.


brrygrdn (approved):
👍🏻

PR #3229, opened 2021-03-04
Fix flaky npm 6 peer dependency test

feelepxyz:
It looks like this spec has become flaky recently, not clear why but
seems to have fixed it by making the expectation compare the dependency
hash instead of the instance.

Also removed the alias on `npm6` as we no longer install npm7 as a npm package.


npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker/version_resolver_spec.rb
```diff
@@ -1319,48 +1319,49 @@
       let(:latest_allowable_version) { Gem::Version.new("16.3.1") }
 
       it "gets the right list of dependencies to update" do
-        expect(resolver.dependency_updates_from_full_unlock).
```
feelepxyz, line 1322: Spotted this flake here https://github.com/dependabot/dependabot-core/runs/2030784173

jasonrudolph (approved):
Nice fix! This does indeed seem like it will be more deterministic. :zap:

PR #3230, opened 2021-03-04
Add an options argument to the bundler_version helper method that is passed from the public API down

brrygrdn:
In order to start testing bundler 2 support, we want to be able to _optionally_ use it in Dependabot core by passing an `options[:bundler_2_available]` flag into the Fetcher, Parser and Updater APIs from the running environment.

This PR adds this configuration layer which currently results in the Bundler native helper using a stubbed V2 library which returns a `NotImplemented` error for the existing method signatures.

This error is this used as a testing proof that our top-level classes fail due to this error if we pass in the options to verify we have passed the configuration correctly. This is further verified by making `options` a required argument of the `bundler_version` helper method, ensuring we get an `ArgumentError` in any cases we've missed.

### DIFF
--- a/bundler/helpers/v2/build
+++ b/bundler/helpers/v2/build
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+set -e
+
+install_dir=$1
+if [ -z "$install_dir" ]; then
+  echo "usage: $0 INSTALL_DIR"
+  exit 1
+fi
+
+helpers_dir="$(dirname "${BASH_SOURCE[0]}")"
+cp -r \
+  "$helpers_dir/.bundle" \
+  "$helpers_dir/lib" \
+  "$helpers_dir/monkey_patches" \
+  "$helpers_dir/run.rb" \
+  "$helpers_dir/Gemfile" \
+  "$install_dir"
+
+cd "$install_dir"
+
+# NOTE: Sets `BUNDLED WITH` to match the installed v1 version in Gemfile.lock
+# forcing specs and native helpers to run with the same version
+#
+# TODO: Change to BUNDLER_VERSION=2 once it is installed
+BUNDLER_VERSION=1 bundle install
--- a/bundler/helpers/v2/build
+++ b/bundler/helpers/v2/build
@@ -12,7 +12,6 @@ helpers_dir="$(dirname "${BASH_SOURCE[0]}")"
 cp -r \
   "$helpers_dir/.bundle" \
   "$helpers_dir/lib" \
-  "$helpers_dir/monkey_patches" \
   "$helpers_dir/run.rb" \
   "$helpers_dir/Gemfile" \
   "$install_dir"
--- a/bundler/helpers/v2/build
+++ b/bundler/helpers/v2/build
@@ -21,6 +21,4 @@ cd "$install_dir"
 
 # NOTE: Sets `BUNDLED WITH` to match the installed v1 version in Gemfile.lock
 # forcing specs and native helpers to run with the same version
-#
-# TODO: Change to BUNDLER_VERSION=2 once it is installed
-BUNDLER_VERSION=1 bundle install
+BUNDLER_VERSION=2 bundle install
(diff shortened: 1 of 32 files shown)


feelepxyz (commented):
Nice one, are there other places left for this or should we get this in?

brrygrdn:
I think there's a few more, my plan was to take this PR right up to having a null adapter that just raises if the feature is on

feelepxyz:
> I think there's a few more, my plan was to take this PR right up to having a null adapter that just raises if the feature is on

Ah yes that makes sense 👌

bundler/helpers/v2/build
```diff
@@ -0,0 +1,26 @@
...
+  "$helpers_dir/Gemfile" \
+  "$install_dir"
+
+cd "$install_dir"
+
+# NOTE: Sets `BUNDLED WITH` to match the installed v1 version in Gemfile.lock
+# forcing specs and native helpers to run with the same version
+#
+# TODO: Change to BUNDLER_VERSION=2 once it is installed
+BUNDLER_VERSION=1 bundle install
```
feelepxyz, line 26: On `main` now
```suggestion
BUNDLER_VERSION=2 bundle install
```

bundler/lib/dependabot/bundler/update_checker/shared_bundler_helpers.rb
```diff
@@ -237,10 +237,6 @@ def using_bundler2?
 
           lockfile.content.match?(/BUNDLED WITH\s+2/m)
         end
-
-        def bundler_version
-          @bundler_version ||= Helpers.bundler_version(lockfile)
-        end
```
brrygrdn, line 243: I've tentatively removed this as I think all the classes mixing it in are overriding it locally, but CI may prove me wrong

bundler/lib/dependabot/bundler/file_updater/lockfile_updater.rb
```diff
@@ -33,11 +33,12 @@ def gemspec_sources
         end
 
         def initialize(dependencies:, dependency_files:,
-                       repo_contents_path: nil, credentials:)
+                       repo_contents_path: nil, credentials:, options: {})
```
brrygrdn, line 36:
```suggestion
                       repo_contents_path: nil, credentials:, options:)
```

I've tried to follow the rule this is a required argument everywhere that isn't a top-level class, so this should be required here.

bundler/helpers/v2/run.rb
```diff
@@ -0,0 +1,30 @@
+require "bundler"
+require "json"
+
+$LOAD_PATH.unshift(File.expand_path("./lib", __dir__))
+$LOAD_PATH.unshift(File.expand_path("../v1/monkey_patches", __dir__))
```
feelepxyz, line 5:
```suggestion
$LOAD_PATH.unshift(File.expand_path("../v2/monkey_patches", __dir__))
```

feelepxyz: oh wait this is intentional

brrygrdn: No worries, I flip flopped on that, I was nervous about hard forking monkey patches. It might make sense to go ahead and fork them but for now I just wanted to avoid breaking the loading.

feelepxyz (approved):
Thanks for testing this! LGTM 💯

feelepxyz:
Looks like quite a few classes require `options` to be passed in

brrygrdn:
Aha, I made it a required argument and didn't update the unit tests 🤕  Should be a relatively quick fix.

brrygrdn:
The CI failures seem to be limited to the new tests that use the v2 native helper, but I'm not seeing them locally, I suspect the issue is the v2 helper just isn't set up properly for CI yet. I've committed `DEBUG_HELPERS=true` to verify this while I go through the dockerfile and ci workflow.

brrygrdn:
Yep, that was the issue, I forgot to add v2 to the Dockerfile and had some build script tweaks to make 😅

feelepxyz (approved):
🙌  just did a release of core so could merge this in now and release after?

PR #3235, opened 2021-03-04
test: update assertions to pass

xlgmokha:
# Why is this needed?

To unblock merging work into the default branch.

# How was this fixed?

By updating the expected bundler version in the assertions
related to the failing tests.


thepwagner (approved):
🎉 

cc @feelepxyz I think this is an alternate method to the patch proposed in https://github.com/dependabot/dependabot-core/pull/3239

I _think_ #3239 may be a more general fix, which I dig. But CI passes here, and I can't argue with a ✅ .

feelepxyz:
> can't argue with a ✅ .

Agreed 👍 :shipit:

PR #3239, opened 2021-03-05
Fix flaky bundler spec by pinning business gem

feelepxyz:
Pin the `business` gem version to 2.1.0 to keep specs from flaking when
new versions are released.

Fixes these test failures:
https://github.com/dependabot/dependabot-core/runs/2034039927


thepwagner (approved):
Pinning SGTM, but tests still fail atm 😿 .

Issue #3173, opened 2021-02-21
bug: dependabot incorrectly changed from one docker tag to another


scyto:
Dependabot should not have changed `FROM node:erbium-buster AS local-copy-src` to `FROM node:15.9.0-alpine AS local-copy-src` in the file docker/Dockerfile.contrib in this commit.

https://github.com/zwave-js/zwavejs2mqtt/commit/1bb27519c0b92a63621f15116c83f118e7e8949a#diff-4ddc2e10398d5c169795b7937f893ec72d2619b395259c592156a5dd0b476c9f

We are unclear why it did this.

Note: the change to the other file in this commit was ok.

Sarke:
Same for me.  It looks like it just grabs the first tag and applies it to all the Dockerfiles.

https://github.com/Sarke/docker-php-fiber/pull/4/commits/66a5b4d74a38b7c9be9445979f281a7ce050c894

```
- FROM php:8.0.1-apache
+ FROM php:8.0.2-apache
```
```
- FROM php:8.0.1-cli
+ FROM php:8.0.2-apache
```
```
- FROM php:8.0.1-zts
+ FROM php:8.0.2-apache
```

feelepxyz:
@mctofu did your PR fix this issue? https://github.com/dependabot/dependabot-core/pull/3277

mctofu:
Yup, the fix has been released and I tested that it covers both of these reports. We now preserve the distinction between tags like `php:8.0.1-apache` and `php:8.0.1-cli` as long as they are in separate files. The zwave-js/zwavejs2mqtt example did include `node:erbium-buster` and `node:erbium-buster-slim` in the same file but since they don't reference a version number they won't receive updates.

PR #3251, opened 2021-03-09
[WIP] docker: don't ignore version tag when consolidating dependencies

mctofu:
Some early work to try and resolve https://github.com/dependabot/dependabot-core/issues/3173.

This causes the docker file parser to consider the tag in addition to
the image name when extracting dependencies. Previously php:8.0.1-apache and
php:8.0.1-cli would be treated as the same dependency which caused them
to both get updated to php:8.0.3-apache. With the change they are
treated as separate dependencies and get updated to php:8.0.3-apache and
php:8.0.3-cli in separate PRs.

Does this seem like the right place to address this or should we try to handle this when doing updates?

One downside of this is that if we had a reference to `php:8.0.1-cli` and `php:8.0.2-cli` they would be treated as 2 different dependencies and you'd end up with an update PR for each. We could address this later by making the deduplication smarter and allowing `php:8.0.1-cli` and `php:8.0.2-cli` to be combined but not `php:8.0.1-apache`.

With this change it parsed these dependencies from my test fixture:
```
[
  {
    "name": "php",
    "version": "8.0.1-apache",
    "requirements": [
      {
        "requirement": null,
        "groups": [

        ],
        "file": "Dockerfile",
        "source": {
          "tag": "8.0.1-apache"
        }
      },
      {
        "requirement": null,
        "groups": [

        ],
        "file": "Dockerfile.ci",
        "source": {
          "tag": "8.0.1-apache"
        }
      }
    ],
    "package_manager": "docker"
  },
  {
    "name": "php",
    "version": "8.0.1-cli",
    "requirements": [
      {
        "requirement": null,
        "groups": [

        ],
        "file": "Dockerfile.cli",
        "source": {
          "tag": "8.0.1-cli"
        }
      }
    ],
    "package_manager": "docker"
  }
]
```
and resulted in this output:
```
=> updating 2 dependencies: php, php

=== php (8.0.1-apache)
 => checking for updates 1/2
 => latest available version is 8.0.3-apache
 => latest allowed version is 8.0.3-apache
 => updating php from 8.0.1-apache to 8.0.3-apache

    ± Dockerfile
    ~~~
    1c1
    < FROM php:8.0.1-apache
    ---
    > FROM php:8.0.3-apache
    ~~~

    ± Dockerfile.ci
    ~~~
    1c1
    < FROM php:8.0.1-apache
    ---
    > FROM php:8.0.3-apache
    ~~~

=== php (8.0.1-cli)
 => checking for updates 2/2
 => latest available version is 8.0.3-cli
 => latest allowed version is 8.0.3-cli
 => updating php from 8.0.1-cli to 8.0.3-cli

    ± Dockerfile.cli
    ~~~
    1c1
    < FROM php:8.0.1-cli
    ---
    > FROM php:8.0.3-cli
    ~~~
```

Prior to the change the dependences parsed as:
```
[
  {
    "name": "php",
    "version": "8.0.1-apache",
    "requirements": [
      {
        "requirement": null,
        "groups": [

        ],
        "file": "Dockerfile",
        "source": {
          "tag": "8.0.1-apache"
        }
      },
      {
        "requirement": null,
        "groups": [

        ],
        "file": "Dockerfile.ci",
        "source": {
          "tag": "8.0.1-apache"
        }
      },
      {
        "requirement": null,
        "groups": [

        ],
        "file": "Dockerfile.cli",
        "source": {
          "tag": "8.0.1-cli"
        }
      }
    ],
    "package_manager": "docker"
  }
]
```
and resulted in this output:
```
=> updating 1 dependencies: php

=== php (8.0.1-apache)
 => checking for updates 1/1
 => latest available version is 8.0.3-apache
 => latest allowed version is 8.0.3-apache
 => updating php from 8.0.1-apache to 8.0.3-apache

    ± Dockerfile
    ~~~
    1c1
    < FROM php:8.0.1-apache
    ---
    > FROM php:8.0.3-apache
    ~~~

    ± Dockerfile.ci
    ~~~
    1c1
    < FROM php:8.0.1-apache
    ---
    > FROM php:8.0.3-apache
    ~~~

    ± Dockerfile.cli
    ~~~
    1c1
    < FROM php:8.0.1-cli
    ---
    > FROM php:8.0.3-apache
    ~~~
```

### DIFF
--- a/docker/lib/dependabot/docker/file_parser.rb
+++ b/docker/lib/dependabot/docker/file_parser.rb
@@ -35,7 +35,7 @@ module Dependabot
       AWS_ECR_URL = /dkr\.ecr\.(?<region>[^.]+).amazonaws\.com/.freeze
 
       def parse
-        dependency_set = DependencySet.new
+        dependencies = []
 
         dockerfiles.each do |dockerfile|
           dockerfile.content.each_line do |line|
@@ -47,7 +47,7 @@ module Dependabot
             version = version_from(parsed_from_line)
             next unless version
 
-            dependency_set << Dependency.new(
+            current =  Dependency.new(
               name: parsed_from_line.fetch("image"),
               version: version,
               package_manager: "docker",
@@ -58,10 +58,16 @@ module Dependabot
                 source: source_from(parsed_from_line)
               ]
             )
+            existing = dependencies.find {|d| d.name == current.name && d.version == current.version }
+            if existing
+              existing.requirements.push(*current.requirements).uniq!
+            else
+              dependencies << current
+            end
           end
         end
 
-        dependency_set.dependencies
+        dependencies
       end
 
       private


feelepxyz (commented):
This is a nice and simple fix! I'm slightly worried of changes that introduce more PRs as [grouped updates](https://github.com/dependabot/dependabot-core/issues/1190) is also one the most requested features.

Reckon the file updater is right place for this fix if we want to keep the grouped update aspect. Maybe we could use the same approach and expand the dependency into two at the file updating stage and apply each each on in turn on the same manifest?

mctofu:
> Reckon the file updater is right place for this fix if we want to keep the grouped update aspect. Maybe we could use the same approach and expand the dependency into two at the file updating stage and apply each each on in turn on the same manifest?

That was a good idea @feelepxyz! I spent a little time experimenting with that and ended up with https://github.com/dependabot/dependabot-core/pull/3277. One thing I'm struggling with there is that if we only return a single dependency in `FileParser` then `UpdateChecker.latest_version` has difficulty with situations where only some of the tags need updating. For ex, if it found `php:8.0.3-apache` and `php:8.0.1-cli` it sees that `php:8.0.3-apache` is up to date and stops there. Is there a way we've worked around that in other ecosystems?

mctofu:
We ended up going with https://github.com/dependabot/dependabot-core/pull/3277 to keep the updates in a single PR.

Issue #2106, opened 2019-09-25
Support multiple lock files for Bundler


paulsturgess:
The [BootBoot](https://github.com/Shopify/bootboot) gem makes it very easy to run tests against against future versions of Rails. 

When running `bundle install` it produces a `Gemfile_next.lock` alongside the usual `Gemfile.lock`. At the moment Dependabot only checks in the `Gemfile.lock` and I guess it ignores the `Gemfile_next.lock` as this is a non-standard file. The result is an error is raised when the CI attempts run bundler before running the tests.

Here's an example of the error output:

```
Using bundler 2.0.2
Installed plugin bootboot

You are trying to install in deployment mode after changing
your Gemfile. Run `bundle install` elsewhere and add the
updated Gemfile_next.lock to version control.

The dependencies in your gemfile changed
You have added to the Gemfile:
* activeadmin (~> 2.3.1)
You have deleted from the Gemfile:
* activeadmin (~> 2.2.0)

The command "eval bundle install --jobs=3 --retry=3 --deployment --path=${BUNDLE_PATH:-vendor/bundle} " failed.
```

It would be great if there was some way to specify any custom lock files (or maybe just any non-standard output) that should be committed by Dependabot. Does this sound like a sane idea?

infin8x:
Looks like you've got this working via multiple `update_configs`, so closing for now.

ashkulz:
@infin8x it doesn't work for us, see my latest comment in #2231

davidwessman:
I solved this problem by adding a Github Action workflow which updates `Gemfile.next.lock` everytime Dependabot updates `Gemfile.lock`. It does not have as nice update logic as Dependabot but help keeping the `Gemfile.next` deployable!

I wrote it up in a blog post:
https://wessman.co/rails-dual-boot-with-dependabot

davidwessman:
@infin8x Can we open up this issue again?
Now my solution and my tests fail since the change to only read-permissions from the 1st of March.
https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/

Or should I create a new PR?

infin8x:
Unfortunately I'm not on the Dependabot team anymore. @hmarr or @feelepxyz may be able to get you in the right direction.

Issue #375, opened 2018-04-27
Support multiple gemfiles


greysteil:
_From @deivid-rodriguez on November 19, 2017 18:44_

Some projects use multiple gemfiles. The main use case I know for this is to be able to test libraries against different major versions of dependencies.

Does dependabot support this kind of setups?

_Copied from original issue: dependabot/feedback#51_

greysteil:
Slightly absurdly, the answer is "yes on the backend, no on the frontend". We've never figured out a good way to display information about which directory (or files) within a project an update setup is targeting, so while the backend supports have multiple update setups of the same language for the same project, there's no way to add them in the frontend.

If you've got a particular repo in mind, let me know and I'll take a look at whether I can get things set up manually on our side.

greysteil:
_From @deivid-rodriguez on November 21, 2017 0:5_

> Slightly absurdly, the answer is "yes on the backend, no on the frontend". We've never figured out a good way to display information about which directory (or files) within a project an update setup is targeting, so while the backend supports have multiple update setups of the same language for the same project, there's no way to add them in the frontend.

I think mentioning the specific gemfile being updated in the description & body of PR's could be enough? Like "Bump <gem_name> from <old_version> to <new_version> in <path_to_gemfile>"?

> If you've got a particular repo in mind, let me know and I'll take a look at whether I can get things set up manually on our side.

Yes, this one: https://github.com/activeadmin/activeadmin!

Dependabot would need to open PR's to keep the `gemfiles/rails_{42,50,21}.gemfile` up to date automatically. We're not currently committing the lock files to version control, but I think we can change that if we can get our dependencies handled automatically in exchange :)

greysteil:
OK, that links in with the work I'm doing on https://github.com/dependabot/feedback/issues/27 and https://github.com/dependabot/dependabot-core/pull/135 - will take a few days but I'll have something for you relatively soon. And I'd love to have Dependabot running on https://github.com/activeadmin/activeadmin so doubly motivated to get it done!

greysteil:
_From @deivid-rodriguez on November 21, 2017 0:17_

Thanks @greysteil, let me know if you need anything from our side! :heart:

greysteil:
@deivid-rodriguez - I've been thinking about this one for a bit:
[...]

PR #3262, opened 2021-03-11
[Bundler] FileParser: Supports multiple gemfiles

davidwessman:
## Goal

Support updating multiple Gemfiles at the same time, there have been two bigger discussions that I have seen before:

- #375 Discussing having one Gemfile for testing a gem against each supported Rails version.
  - ActiveAdmin has multiple Gemfile and Gemfile.lock in separate folders, for example `./gemfiles/rails_52/Gemfile`, `./gemfiles/rails_52/Gemfile.lock`.
  - They fixed their situation by configuring dependabot to update each folder separately: [config.yml](https://github.com/activeadmin/activeadmin/blob/master/.dependabot/config.yml).
  - Multiple Gemfiles can be updated separately and therefore the solution with many folders work.
  - **Good:** They will receive one PR for each update to Rails.
  - **Bad, but works:** They will receive the same PR for each gemfile if a shared dependency is updated.
  
- #2106 Discusses multiple Gemfiles for dual-booting a Rails application, see more information from [fastruby.io](https://www.fastruby.io/blog/upgrade-rails/dual-boot/dual-boot-with-rails-6-0-beta.html) or [Shopify's bootboot](https://github.com/Shopify/bootboot).
  - An example dual-booted with fastruby's setup is https://github.com/davidwessman/synka.
  - One Gemfile, one Gemfile.next which is just a symlink to the main one, then two separate lockfiles Gemfile.lock and Gemfile.next.lock (For bootboot the files are called Gemfile_next and Gemfile_next.lock).
  - Some of the listed dependencies and versions are only used for Gemfile.lock and some for Gemfile.next.lock, but as many as possible are shared.
  - **Bad**: When an update is done in the Gemfile with a specified requirement, the Gemfile.next.lock gets out of sync with its Gemfile and the application is no longer deployable in both versions.

My solution to the second situation have been running a special Github Action Workflow everytime there is a PR with changes to `Gemfile.lock` and just run `BUNDLE_GEMFILE=Gemfile.next bundle update --minor --conservative` and commit the change. But this no longer works for PRs triggered by Dependabot since the read-only update on the 1st of March, [blog post](https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/).

## Changes

- I started working on the FileParser to allow multiple gemfiles and lockfiles, I have made my way through the parsers to get the test to work, but not quite happy with the results.
- [x] FileParser: Allow multiple gemfiles and lockfiles.
- [ ] FileParser: Write tests for multiple Gemfiles.
- [ ] Next steps?

Would appreciate some feedback on what to do next, and I understand if this is not interesting or possible to change right now.

### DIFF
--- a/bundler/lib/dependabot/bundler/file_parser.rb
+++ b/bundler/lib/dependabot/bundler/file_parser.rb
@@ -45,23 +45,25 @@ module Dependabot
       def gemfile_dependencies
         dependencies = DependencySet.new
 
-        return dependencies unless gemfile
+        return dependencies unless gemfiles.any?
 
-        [gemfile, *evaled_gemfiles].each do |file|
-          parsed_gemfile.each do |dep|
+        all_gemfiles = gemfiles + evaled_gemfiles
+
+        all_gemfiles.each do |gemfile|
+          parsed_gemfile(gemfile).each do |dep|
             gemfile_declaration_finder =
-              GemfileDeclarationFinder.new(dependency: dep, gemfile: file)
+              GemfileDeclarationFinder.new(dependency: dep, gemfile: gemfile)
             next unless gemfile_declaration_finder.gemfile_includes_dependency?
 
             dependencies <<
               Dependency.new(
                 name: dep.fetch("name"),
-                version: dependency_version(dep.fetch("name"))&.to_s,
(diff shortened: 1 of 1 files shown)


davidwessman:
After more investigation I think this requires larger changes :/.
See #3266

PR #3271, opened 2021-03-15
Install bundler 2.2.11 and update gem

feelepxyz:
Install bundler 2 and update gem from 3.0.3 to 3.2.14

Bundler 2 is only used to install core's gems when building the image, bundler 1 is still used for all updates.

Once we ship this we can enable bundler v2 support here https://github.com/dependabot/dependabot-core/blob/main/bundler/lib/dependabot/bundler/helpers.rb#L10

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,12 +54,13 @@ RUN apt-get update \
 
 # Install Ruby 2.6.6, update RubyGems, and install Bundler
 ENV BUNDLE_SILENCE_ROOT_WARNING=1
-RUN apt-get install -y software-properties-common \
+RUN apt-get update && apt-get install -y software-properties-common \
   && apt-add-repository ppa:brightbox/ruby-ng \
   && apt-get update \
   && apt-get install -y ruby2.6 ruby2.6-dev \
-  && gem update --system 3.0.3 \
-  && gem install bundler -v 1.17.3 --no-document
+  && gem update --system 3.2.14 \
+  && gem install bundler -v 1.17.3 --no-document \
+  && gem install bundler -v 2.2.11 --no-document
 
 
 ### PYTHON


Dockerfile
```diff
@@ -54,12 +54,13 @@ RUN apt-get update \
 
 # Install Ruby 2.6.6, update RubyGems, and install Bundler
 ENV BUNDLE_SILENCE_ROOT_WARNING=1
-RUN apt-get install -y software-properties-common \
+RUN apt-get update && apt-get install -y software-properties-common \
```
feelepxyz, line 57: This step failed without this initial update complaining that `software-properties-common` wasn't installed, not sure why

Dockerfile
```diff
@@ -54,12 +54,13 @@ RUN apt-get update \
...
   && apt-add-repository ppa:brightbox/ruby-ng \
   && apt-get update \
   && apt-get install -y ruby2.6 ruby2.6-dev \
-  && gem update --system 3.0.3 \
-  && gem install bundler -v 1.17.3 --no-document
+  && gem update --system 3.2.14 \
```
feelepxyz, line 61: Have dug through changes on rubygems for the version, requirement and specifications class. There are some changes to sorting of the requirements class: https://github.com/rubygems/rubygems/pull/2652/files

Test suite looks green so seems we don't rely on this here. I know we rely on it somewhat in api which prevented us from upgrading to ruby 2.7 but don't think this change will affect api as `gem` is installed from the `ruby2.7` docker image there.

feelepxyz:
Planning to test this on staging, dry-runs look 👌

jurre (approved):

feelepxyz:
Looking good on staging 🎉  https://github.com/dsp-testing/feelepxyz-testing-git-bundler/pull/9/files

Issue #2992, opened 2021-01-15
D-bot encountered an unknown error: undefined method `zero?' for nil:NilClass


pzygielo:
**Package manager/ecosystem**
maven

**What you expected to see, versus what you actually saw**
Expected - clean run at least, finally "error prone needs update to [2.5.1](https://github.com/google/error-prone/releases/tag/v2.5.1)" to be discovered.

Actually seen:
>  Dependabot encountered an unknown error
>
>  Dependabot failed to update your dependencies because an unexpected error occured. See the logs for more details.
```
...
INFO <job_83400252> Checking if com.google.errorprone:javac 9+181-r4173-1 needs updating
  proxy | 2021/01/15 08:19:47 [026] GET https://jitpack.io:443/com/google/errorprone/javac/maven-metadata.xml
  proxy | 2021/01/15 08:19:47 [026] 404 https://jitpack.io:443/com/google/errorprone/javac/maven-metadata.xml
  proxy | 2021/01/15 08:19:47 [028] GET https://repo.maven.apache.org:443/maven2/com/google/errorprone/javac/maven-metadata.xml
  proxy | 2021/01/15 08:19:47 [028] 200 https://repo.maven.apache.org:443/maven2/com/google/errorprone/javac/maven-metadata.xml
updater | I, [2021-01-15T08:19:47.183815 #73]  INFO -- sentry: ** [Raven] Sending event e5a5f5d4896249929c2d41823ca6f0d7 to Sentry
  proxy | 2021/01/15 08:19:47 [030] POST https://sentry.io:443/api/1451818/store/
  proxy | 2021/01/15 08:19:47 [030] 200 https://sentry.io:443/api/1451818/store/
updater | ERROR <job_83400252> Error processing com.google.errorprone:javac (NoMethodError)
updater | ERROR <job_83400252> undefined method `zero?' for nil:NilClass
updater | ERROR <job_83400252> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-maven-0.130.0/lib/dependabot/maven/version.rb:83:in `block in <=>'
updater | ERROR <job_83400252> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-maven-0.130.0/lib/dependabot/maven/version.rb:76:in `times'
[...]

PR #3274, opened 2021-03-15
Maven: fix error when comparing string and integer versions

jurre:
When comparing a string version with one that can be coerced to an
integer, we would call `<=>` on two different types, resulting in `nil`
values that would raise errors further up the call stack.

This introduces a change to only coerce these values to ints if we can
do this on both, and perform a string comparison otherwise.

Fixes: https://github.com/dependabot/dependabot-core/issues/2992

### DIFF
--- a/maven/lib/dependabot/maven/version.rb
+++ b/maven/lib/dependabot/maven/version.rb
@@ -175,8 +175,11 @@ module Dependabot
 
         return 1 if NAMED_QUALIFIERS_HIERARCHY[other_token]
 
-        token = token.to_i if token.match?(/^\d+$/)
-        other_token = other_token.to_i if other_token.match?(/^\d+$/)
+        if token.match?(/\A\d+\z/) && other_token.match?(/\A\d+\z/)
+          token = token.to_i
+          other_token = other_token.to_i
+        end
+
         token <=> other_token
       end
     end
--- a/maven/spec/dependabot/maven/version_spec.rb
+++ b/maven/spec/dependabot/maven/version_spec.rb
@@ -307,6 +307,12 @@ RSpec.describe Dependabot::Maven::Version do
           let(:other_version) { described_class.new("1-alpha-1") }
           it { is_expected.to eq(0) }
         end
+
+        context "comparing string versions with integer ones" do
+          let(:version) { described_class.new("181") }
+          let(:other_version) { described_class.new("dev") }
+          it { is_expected.to eq(1) }
+        end
       end
     end
   end


maven/spec/dependabot/maven/version_spec.rb
```diff
@@ -297,6 +297,12 @@
           let(:other_version) { described_class.new("1-alpha-1") }
           it { is_expected.to eq(0) }
         end
+
+        context "comparing string versions with integer ones" do
+          let(:version) { described_class.new("181") }
+          let(:other_version) { described_class.new("dev") }
+          it { is_expected.to eq(1) }
+        end
```
xlgmokha, line 305: Is it worth including a test for "hello\r\n181\r\nworld"?

maven/lib/dependabot/maven/version.rb
```diff
@@ -175,8 +175,11 @@ def compare_token(token:, other_token:)
 
         return 1 if NAMED_QUALIFIERS_HIERARCHY[other_token]
 
-        token = token.to_i if token.match?(/^\d+$/)
-        other_token = other_token.to_i if other_token.match?(/^\d+$/)
+        if token.match?(/^\d+$/) && other_token.match?(/^\d+$/)
```
xlgmokha, line 178: Would it be safer to use `\A` and `\z` to match start and end of string to make sure the token doesn't have additional line endings injected in? It's not clear to me if token and/or other_token is considered user input or not.

WDYT?

xlgmokha (approved):
I added a couple of suggestions. Feel free to ignore if it's overly defensive. Otherwise, this LGTM.

maven/spec/dependabot/maven/version_spec.rb
jurre: I don't think it's even a valid version string, should we need to worry about it?

xlgmokha: > I don't think it's even a valid version string, should we need to worry about it?

No it's not a valid version string but I don't trust the input until I know where it's coming from. I ask because exploiting regex matching on `^` and `&` is common.

maven/lib/dependabot/maven/version.rb
jurre: Yeah, seems safer 👍

maven/spec/dependabot/maven/version_spec.rb
feelepxyz: > exploiting regex matching on ^ and & is common

Good shout but this might be tricky to test as the version class validates the input afaik

jurre: We can instantiate a `Maven::Version` with that string, but I'm not sure what we should be testing for in that case. If you've specified that as the version, we either compare the string `"hello\r\n181\r\nworld"` to whatever the other version is (as a string), or we compare `181` (int) with another int, but it's pretty unlikely your update will pass either way so I'm not sure if it really matters which one we use in the comparison. Am I missing the point, could this be exploited some way?

xlgmokha: I'm sorry I didn't do a very good job of explaining my thinking very well.

When I first read the code the regex that was used was `/^\d+$/`. I wanted to suggest adding a test that uses the string `hello\n181\nworld` to indicate how this regex match could be bypassed. By adding that test I thought it would lead to the understanding of why `\A` and `\z` is preferred and prevent a regression from entering the code base.

```irb
irb(main):001:0> "hello\n181\nworld".match(/^\d+$/)
=> #<MatchData "181">
irb(main):002:0> "hello\n181\nworld".match(/\A\d+\z/)
=> nil
irb(main):003:0> "181".match(/\A\d+\z/)
=> #<MatchData "181">
```

We ended up making the suggested regex change without adding a test to prove why it's favoured. I still like the idea of adding a test to document why we would use `\A` over `^` and etc. This is also not a blocker for me. I hope this helps clarify my thinking process.

jurre: Thanks for clarifying, I understand how the different regexes parse those values, I just don't get how that could lead to potential issues in this case. 

This regex is not used for parsing the version, it's there to check if we should be doing a string<>string or int<>int comparison.

@feelepxyz was actually right that it's not possible to instantiate a version with that string:

```ruby
Dependabot::Maven::Version.new("hello\n181\nworld")
*** ArgumentError Exception: Malformed version number string hello
181
world
```

I made the change to the regex because I thought it more clearly communicates what the regex intends to do, but I don't think that there is a potential bug that we could regress to.

PR #3277, opened 2021-03-16
docker: consider each requirement when updating dependencies

### LINKED ISSUE #3173: bug: dependabot incorrectly changed from one docker tag to another

mctofu:
This change works toward resolving #3173 with an alternate approach to https://github.com/dependabot/dependabot-core/pull/3251. It maintains grouping of updates so that we can update `php:8.0.1-apache` and `php:8.0.1-cli` in a single PR instead of two as suggested in https://github.com/dependabot/dependabot-core/pull/3251#pullrequestreview-612037713.

Prior to this change if there were multiple tags with different prefixes/suffixes we'd only resolve the latest version for one of them and would end up updating all of the tags to a common version:
```
=== php (8.0.1-apache)
 => checking for updates 1/1
 => latest available version is 8.0.3-apache
 => latest allowed version is 8.0.3-apache
 => updating php from 8.0.1-apache to 8.0.3-apache

    ± Dockerfile
    ~~~
    1c1
    < FROM php:8.0.1-apache
    ---
    > FROM php:8.0.3-apache
    ~~~

    ± Dockerfile.cli
    ~~~
    1c1
    < FROM php:8.0.1-cli
    ---
    > FROM php:8.0.3-apache
    ~~~
```

This change resolves the latest version for each tag we can correctly preserve the prefix/suffix in the update:
```
=== php (8.0.1-apache)
 => checking for updates 1/1
 => latest available version is 8.0.3-apache
 => latest allowed version is 8.0.3-apache
 => updating php from 8.0.1-apache to 8.0.3-apache

    ± Dockerfile
    ~~~
    1c1
    < FROM php:8.0.1-apache
    ---
    > FROM php:8.0.3-apache
    ~~~

    ± Dockerfile.cli
    ~~~
    1c1
    < FROM php:8.0.1-cli
    ---
    > FROM php:8.0.3-cli
    ~~~
```

One caveat is this only works if the tags are in different files. If a Dockerfile includes two tags with different prefixes/suffixes then we still end up replacing both of them with a single tag. Fixing that requires splitting out to multiple dependencies and updating the file updater to handle multiple dependencies. I started going down that road (https://github.com/dependabot/dependabot-core/commit/b5b80df63869bd54186dfd5afd027b8f49f2889d) but that's going to be a bigger chunk of work and I think it would be better to release this initial fix first.

### DIFF
--- a/docker/lib/dependabot/docker/update_checker.rb
+++ b/docker/lib/dependabot/docker/update_checker.rb
@@ -97,17 +97,22 @@ module Dependabot
 
       def version_up_to_date?
         # If the tag isn't up-to-date then we can definitely update
-        return false if version_tag_up_to_date? == false
+        return false if version_tag_up_to_date?(dependency.version) == false
+        return false if dependency.requirements.any? do |req|
+                          version_tag_up_to_date?(req.fetch(:source, {})[:tag]) == false
+                        end
 
         # Otherwise, if the Dockerfile specifies a digest check that that is
         # up-to-date
         digest_up_to_date?
       end
 
-      def version_tag_up_to_date?
-        return unless dependency.version.match?(NAME_WITH_VERSION)
+      def version_tag_up_to_date?(version)
+        return unless version&.match?(NAME_WITH_VERSION)
 
-        old_v = numeric_version_from(dependency.version)
+        latest_version = fetch_latest_version(version)
+
+        old_v = numeric_version_from(version)
         latest_v = numeric_version_from(latest_version)
 
         return true if version_class.new(latest_v) <= version_class.new(old_v)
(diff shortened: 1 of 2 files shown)


feelepxyz (approved):
Thanks for splitting this out of the file updater WIP, shipping this first makes lots of sense to me 👍 

Changes LGTM 💯

PR #3279, opened 2021-03-16
Update dev and ci Dockerfile to match updater

feelepxyz:
Updating the dev and ci Dockerfile to match the user permissions set by the updater image.

This reproduces the error seen production when updating a vendored bundler project.

The [failing bundler specs](https://github.com/dependabot/dependabot-core/runs/2122196759) are expected and fixed here https://github.com/dependabot/dependabot-core/pull/3280

### DIFF
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -1,12 +1,29 @@
 FROM dependabot/dependabot-core
-ARG CODE_DIR=/home/dependabot/dependabot-core
+
+USER root
+
+ARG USERNAME=dependabot
+ARG DEPENDABOT_HOME=/home/${USERNAME}
+ARG CODE_DIR=${DEPENDABOT_HOME}/dependabot-core
+
+# Install project dependencies as an unprivileged user
+RUN useradd -m ${USERNAME} && mkdir ${CODE_DIR}
+RUN chown -R ${USERNAME}:${USERNAME} \
+  ${CODE_DIR} \
+  /usr/local/.pyenv \
+  /opt/go/gopath \
+  /opt/rust/ \
+  /opt/bundler/
+
 WORKDIR ${CODE_DIR}
 
-ENV BUNDLE_PATH="/home/dependabot/.bundle" \
-  BUNDLE_BIN=".bundle/binstubs" \
-  PATH=".bundle/binstubs:$PATH:/home/dependabot/.bundle/bin"
+USER ${USERNAME}
 
-COPY .rubocop.yml /home/dependabot/dependabot-core/
+ENV BUNDLE_PATH="${CODE_DIR}/.bundle" \
+  BUNDLE_BIN=".bundle/bin"
+ENV PATH="$BUNDLE_BIN:$PATH:$BUNDLE_PATH/bin"
+
+COPY .rubocop.yml ${CODE_DIR}/
 
 RUN mkdir -p \
   ${CODE_DIR}/bundler \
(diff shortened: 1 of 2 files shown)


bundler/lib/dependabot/bundler/native_helpers.rb
```diff
@@ -17,7 +17,9 @@ def self.run_bundler_subprocess(function:, args:, bundler_version:)
               # Bundler will pick the matching installed major version
               "BUNDLER_VERSION" => bundler_version,
               "BUNDLE_GEMFILE" => File.join(versioned_helper_path(bundler_version: bundler_version), "Gemfile"),
-              "BUNDLE_PATH" => File.join(versioned_helper_path(bundler_version: bundler_version), ".bundle")
+              "BUNDLE_PATH" => File.join(versioned_helper_path(bundler_version: bundler_version), ".bundle"),
+              # Prevent the GEM_HOME from being set to a folder owned by root
+              # "GEM_HOME" => File.join(versioned_helper_path(bundler_version: bundler_version), ".bundle")
```
feelepxyz, line 22: bundler tests for vendoring should go red without this

Dockerfile.ci
```diff
@@ -1,12 +1,29 @@
...
+  /opt/rust/ \
+  /opt/bundler/
+
+ARG CODE_DIR=${DEPENDABOT_HOME}/dependabot-core
 WORKDIR ${CODE_DIR}
 
-ENV BUNDLE_PATH="/home/dependabot/.bundle" \
-  BUNDLE_BIN=".bundle/binstubs" \
-  PATH=".bundle/binstubs:$PATH:/home/dependabot/.bundle/bin"
+USER ${USERNAME}
```
feelepxyz, line 20: This setup is the same as Dockerfile.development and the updater, previously ci would run as root having access to `/var/lib` which is the default cache directory for rubygems when `GEM_HOME` isn't set

Dockerfile.ci
```diff
@@ -1,12 +1,29 @@
...
+
+ENV BUNDLE_PATH="${CODE_DIR}/.bundle" \
+  BUNDLE_BIN=".bundle/bin"
+ENV PATH="$BUNDLE_BIN:$PATH:$BUNDLE_PATH/bin"
```
brrygrdn, line 24: `$BUNDLE_BIN` and `$BUNDLE_PATH/bin` are now the same value of `.bundle/bin` - do we still need both?

Dockerfile.ci
```diff
@@ -67,23 +84,23 @@ RUN cd python && bundle install
...
+COPY --chown=${USERNAME}:${USERNAME} github_actions/ ${CODE_DIR}/github_actions/
+COPY --chown=${USERNAME}:${USERNAME} go_modules/ ${CODE_DIR}/go_modules/
+COPY --chown=${USERNAME}:${USERNAME} gradle/ ${CODE_DIR}/gradle/
+COPY --chown=${USERNAME}:${USERNAME} hex/ ${CODE_DIR}/hex/
+COPY --chown=${USERNAME}:${USERNAME} maven/ ${CODE_DIR}/maven/
+COPY --chown=${USERNAME}:${USERNAME} npm_and_yarn/ ${CODE_DIR}/npm_and_yarn/
+COPY --chown=${USERNAME}:${USERNAME} nuget/ ${CODE_DIR}/nuget/
+COPY --chown=${USERNAME}:${USERNAME} python/ ${CODE_DIR}/python/
+COPY --chown=${USERNAME}:${USERNAME} terraform/ ${CODE_DIR}/terraform/
+COPY --chown=${USERNAME}:${USERNAME} omnibus/ ${CODE_DIR}/omnibus/
```
feelepxyz, line 104: Seems `COPY` defaults to the file being owned by `root`, this was failing when running rspec and loading the `vcr` configuration that attempted to create a folder here: https://github.com/dependabot/dependabot-core/blob/main/common/spec/spec_helper.rb#L38 ([vcr codes](https://github.com/vcr/vcr/blob/master/lib/vcr/cassette/persisters/file_system.rb#L15))

Dockerfile.ci
```diff
@@ -1,12 +1,29 @@
 FROM dependabot/dependabot-core
-ARG CODE_DIR=/home/dependabot/dependabot-core
+
+USER root
+
+ARG DEPENDABOT_HOME=/home/dependabot
+ARG USERNAME=dependabot
```
brrygrdn, line 6:
```suggestion
ARG USERNAME=dependabot
ARG DEPENDABOT_HOME=/home/${USERNAME}
```

Minor, matches `Dockerfile.development`

Dockerfile.ci
feelepxyz: Yeah this tripped me up several times, the `BUNDLE_BIN` is set to a relative path but the `BUNDLE_PATH` isn't

Dockerfile.ci
brrygrdn: Yeah, the default `COPY` behaviour is surprising.

To reduce repetition we need to update this, we could just `chown -r ${CODE_DIR}` after the copy block?  I don't feel strongly about making that change.

Dockerfile.ci
brrygrdn: 🦅 👁️ 

Good catch!

brrygrdn (approved):
Nits only, change LGTM - thanks for making this change as that was definitely causing a lot of subtle issues.

Dockerfile.ci
feelepxyz: Hm attempted to run a `chwon -r ${CODE_DIR}` after the copy but looks like its failing and running out of disk space, reverting to the inline option https://github.com/dependabot/dependabot-core/runs/2122700298

PR #3280, opened 2021-03-16
Bundler: Fix permission error when vendoring gems

feelepxyz:
This fixes a permissions error when vendoring gems, when `GEM_HOME` was
unset it defaulted to `/var/lib` which isn't writeable.

Attempting to fix the discrepancy between ci and updater here:
https://github.com/dependabot/dependabot-core/pull/3279

### DIFF
--- a/bundler/lib/dependabot/bundler/native_helpers.rb
+++ b/bundler/lib/dependabot/bundler/native_helpers.rb
@@ -17,7 +17,9 @@ module Dependabot
               # Bundler will pick the matching installed major version
               "BUNDLER_VERSION" => bundler_version,
               "BUNDLE_GEMFILE" => File.join(versioned_helper_path(bundler_version: bundler_version), "Gemfile"),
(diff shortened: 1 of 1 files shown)


jurre (approved):
Could we somehow test for this?

feelepxyz:
> Could we somehow test for this?

Yes attempted to do it here https://github.com/dependabot/dependabot-core/pull/3279 but looks like it's breaking a whole tonne of stuff in the process so keen to get this out to fix the issue in prod and then try and get that PR working.

I can reproduce it in the dev container and the vendor specs also fail in the dev container but pass in the ci container because they run as root.

feelepxyz:
@jurre reproduced the failure here https://github.com/dependabot/dependabot-core/runs/2122196759 🎉

feelepxyz:
Tested successfully on staging https://github.com/dsp-testing/feelepxyz-testing-git-bundler/pull/12/files

PR #3283, opened 2021-03-16
Instrument bundler version

jurre:
Instrument the version of bundler that a package has specified.

This way we can collect metrics about usage of these gems.

This PR implements just a single ecosystem, but we've tried to take into account changes that might be required for other ecosystems. For python, for example, it is possible to have multiple package managers configured, so we instrument a hash of package manager => version.

Planning to have the consumer side of this set up and reporting actual data before adding more ecosystems to the mix.

### DIFF
--- a/bundler/lib/dependabot/bundler/helpers.rb
+++ b/bundler/lib/dependabot/bundler/helpers.rb
@@ -11,6 +11,13 @@ module Dependabot
       def self.bundler_version(_lockfile)
         V1
       end
+
+      def self.detected_bundler_version(lockfile)
+        return "unknown" unless lockfile
+        return V2 if lockfile.content.match?(/BUNDLED WITH\s+2/m)
+
+        V1
+      end
     end
   end
 end
--- a/common/dependabot-common.gemspec
+++ b/common/dependabot-common.gemspec
@@ -21,6 +21,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = ">= 2.5.0"
   spec.required_rubygems_version = ">= 2.7.3"
 
+  spec.add_dependency "activesupport", ">= 6.0.0"
   spec.add_dependency "aws-sdk-codecommit", "~> 1.28"
   spec.add_dependency "aws-sdk-ecr", "~> 1.5"
   spec.add_dependency "bundler", ">= 1.16", "< 3.0.0"
--- a/common/lib/dependabot/notifications.rb
+++ b/common/lib/dependabot/notifications.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+require "active_support/notifications"
+
+module Dependabot
+  module Notifications
+    FILE_PARSER_PACKAGE_MANAGER_VERSION_PARSED = "dependabot.file_parser.package_manager_version_parsed"
+  end
+
+  def self.instrument(name, payload = {})
+    ActiveSupport::Notifications.instrument(name, payload)
+  end
+
+  def self.subscribe(pattern = nil, callback = nil, &block)
+    ActiveSupport::Notifications.subscribe(pattern, callback, &block)
+  end
+end
--- a/bundler/lib/dependabot/bundler/file_parser.rb
+++ b/bundler/lib/dependabot/bundler/file_parser.rb
@@ -23,6 +23,7 @@ module Dependabot
         dependency_set += gemspec_dependencies
         dependency_set += lockfile_dependencies
         check_external_code(dependency_set.dependencies)
+        instrument_package_manager_version
         dependency_set.dependencies
       end
 
@@ -40,6 +41,17 @@ module Dependabot
         dependencies.any? do |dep|
           dep.requirements.any? { |req| req.fetch(:source)&.fetch(:type) == "git" }
         end
+      end
+
+      def instrument_package_manager_version
+        version = Helpers.detected_bundler_version(lockfile)
+        Dependabot.instrument(
+          Notifications::FILE_PARSER_PACKAGE_MANAGER_VERSION_PARSED,
+          ecosystem: "bundler",
+          package_managers: {
+            "bundler" => version
+          }
+        )
       end
 
       def gemfile_dependencies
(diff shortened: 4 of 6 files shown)


python/lib/dependabot/python/file_parser.rb
```diff
@@ -104,6 +106,38 @@ def requirement_dependencies
...
+      def pipenv_version
+        "unknown"
+      end
+
+      def pip_version
+        "unknown"
+      end
+
+      def setup_py_version
+        "unknown"
```
jurre, line 138: I'm not sure if we can even get a version for this?

common/lib/dependabot/notifications.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require "active_support/notifications"
+
+module Dependabot
+  def self.instrument(name, payload = {})
```
jurre, line 6: I didn't want to spill `ActiveSupport::Notifications` all over the place, but not sure if this adds much value looking at it again, if we ever decide to replace this with some other system we most likely will need to change the interface anyway, and it just hides information from a reader of the code.

feelepxyz: Seems ok to wrap it up, feel like you always end up wanting to wrap stuff for different reasons in the end

python/lib/dependabot/python/file_parser.rb
feelepxyz: yeah not sure there is a version?

feelepxyz: but this would be valuable even without version information, just knowing which ones are actually used

common/lib/dependabot/notifications.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require "active_support/notifications"
+
+module Dependabot
```
feelepxyz, line 5: Thoughts on keeping track of event names here as constants?

jurre: I dig it!

jurre: Hmm, it does make it harder to remove the instrumentation. When this is just a string, we can keep the `subscribe` call in the calling code while we remove all of the instrumentation code, the calling code will just noop. Now we need to be careful to not remove the constant because it would break upstream usage

jurre: In this single scenario I don't think ^ matters much, but when thinking about a generic system that lets us instrument all sort of things, I feel like optimizing to make it easy to add/delete notifications without friction is something to strive for. Thoughts?

bundler/lib/dependabot/bundler/helpers.rb
```diff
@@ -11,6 +11,14 @@ module Helpers
       def self.bundler_version(_lockfile)
         V1
       end
+
+      # TODO: Replace with bundler_version once it is implemented
+      def self.actual_bundler_version(lockfile)
+        return V1 unless lockfile
```
jurre, line 17: Should we return `unknown` here?

I guess the question ultimately is: should we instrument the _version used_ or the _version specified_

feelepxyz: Ah yes, I think it would be useful to gauge the usage of current users so returning `V1` makes sense to me 👍

feelepxyz: Hm I just re-read this, good point when there's no lockfile, I guess we'll eventually default to `V2` here so maybe a good idea to return `uknown` actually so we can get a sense for how many projects would be impacted by flipping the default, thoughts?

common/lib/dependabot/notifications.rb
feelepxyz: Good point, maybe doesn't make sense to add all notifications as constants but seems ok for ones that we care more about routing to specific places, for this one we'll probably want to end up persisting on update jobs directly whereas other ones we might just want to pass straight through to datadog?

common/dependabot-common.gemspec
```diff
@@ -21,6 +21,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = ">= 2.5.0"
   spec.required_rubygems_version = ">= 2.7.3"
 
+  spec.add_dependency "activesupport", ">= 6.0.0"
```
feelepxyz, line 24: Should we pin this to the major version or thinking the core API probably won't change so safe to allow a future v7?

```suggestion
  spec.add_dependency "activesupport", "~> 6.0.0"
```

feelepxyz (approved):
Added a few minor comments, otherwise LGTM 👍

bundler/lib/dependabot/bundler/helpers.rb
jurre: Yeah, agreed 👍 we'll probably want to keep this separate helper so I'll think of a more permanent name

common/dependabot-common.gemspec
jurre: Since activesupport is such a widely used library, I figured that pinning it to a major version might be annoying for other projects that want to pull in core as a dependency

feelepxyz: Yeah makes sense 👍

common/lib/dependabot/notifications.rb
brrygrdn: I agree that ideally it would be easier to add more, but I'm very mildly paranoid that core, as a public library, could have instrumentation added to it that we might not want to relay through our runner to instrumentation services.

Having a list of constants in Dependabot Core allows us to programmatically filter events as we can, by default relay all constants in Dependabot::Notifications. 

We'd have to consciously bump dependabot-core in our runner to allow new notification to relay which would decouple a core upgrade from a sudden surge of new events hitting our instrumentation layer.

`</tinfoil-hat>`

brrygrdn (approved):
🚀  Nice - I'm happy to merge this into https://github.com/dependabot/dependabot-core/pull/3289 if you want to roll up into a single release

common/lib/dependabot/notifications.rb
jurre: Good point @brrygrdn, my thinking was that we would only subscribe to specific events and not relay anything, but I can definitely see us moving towards that approach

PR #3296, opened 2021-03-18
Add support for disablePackageSources in NuGet.Config

### LINKED ISSUE #3295: Disabled package sources are used in NuGet.Config
For my repository, I have a disabled package source in my `NuGet.Config`---used only for CD builds where it is enabled.

This causes Dependabot to think they are active package sources:

```log
updater | I, [2021-03-17T20:05:49.459298 #7]  INFO -- sentry: ** [Raven] Raven 3.1.1 ready to catch errors
updater | INFO <job_96808390> Starting job processing
updater | INFO <job_96808390> Starting update job for AshleighAdams/Verlite
updater | INFO <job_96808390> Checking if xunit 2.4.1 needs updating
  proxy | 2021/03/17 20:05:53 [036] GET https://nuget.pkg.github.com:443/AshleighAdams/index.json
  proxy | 2021/03/17 20:05:53 [036] 401 https://nuget.pkg.github.com:443/AshleighAdams/index.json
updater | INFO <job_96808390> Handled error whilst updating xunit: private_source_authentication_failure {:source=>"https://nuget.pkg.github.com/AshleighAdams/index.json"}
```

My `NuGet.Config` looks as follows:

```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <add key="github" value="https://nuget.pkg.github.com/AshleighAdams/index.json" />
  </packageSources>
  <disabledPackageSources>
    <add key="github" value="true" />
  </disabledPackageSources>
</configuration>
```

I would expect Dependabot to ignore the disabled `github` package source, and thus require no secrets.

The repository in question is https://github.com/AshleighAdams/Verlite

AshleighAdams:
This prevents disabled package sources, such as those used for doing deploys to private package registries, from being used.

Will read, and disable, any repositories specified in `disabledPackageSources` in `NuGet.Config`, for example, the following `github` package source will not be used:

```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <add key="github" value="https://nuget.pkg.github.com/AshleighAdams/index.json" />
  </packageSources>
  <disabledPackageSources>
    <add key="github" value="true" />
  </disabledPackageSources>
</configuration>
```

https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file#disabledpackagesources

~~Though I haven't tested, I strongly suspect that without this change, if a disabled source has a newer dependency version, that the PR dependabot creates would fail to build.~~

Edit: Confirmed, does trigger the bug, see https://github.com/AshleighAdams/dependabot-bug-poc/pull/1

Fix #3295

### DIFF
--- a/nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
+++ b/nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
@@ -151,7 +151,6 @@ module Dependabot
         # rubocop:disable Metrics/CyclomaticComplexity
         # rubocop:disable Metrics/PerceivedComplexity
         # rubocop:disable Metrics/AbcSize
-        # rubocop:disable Metrics/MethodLength
         def repos_from_config_file(config_file)
           doc = Nokogiri::XML(config_file.content)
           doc.remove_namespaces!
@@ -167,16 +166,7 @@ module Dependabot
               }
             end
 
-          disabled_sources =
-            doc.css("configuration > disabledPackageSources > add").map do |node|
-              value = node.attribute("value")&.value&.strip&.downcase ||
-                      node.at_xpath("./value")&.content&.strip&.downcase
-              if value == "true"
-                node.attribute("key")&.value&.strip ||
-                  node.at_xpath("./key")&.content&.strip
-              end
-            end
-
+          disabled_sources = disabled_sources(doc)
           sources.reject! do |s|
             disabled_sources.include?(s[:key])
           end
@@ -197,7 +187,6 @@ module Dependabot
 
           sources
         end
-        # rubocop:enable Metrics/MethodLength
         # rubocop:enable Metrics/AbcSize
         # rubocop:enable Metrics/PerceivedComplexity
         # rubocop:enable Metrics/CyclomaticComplexity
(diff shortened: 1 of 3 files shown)


nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
```diff
@@ -166,6 +167,20 @@ def repos_from_config_file(config_file)
               }
             end
 
+          disabled_sources =
+            doc.css("configuration > disabledPackageSources > add").map do |node|
```
jurre, line 171: I think we could extract a method for this, the `repos_from_config` method is already pretty complext:

```ruby
        def disabled_sources # rubocop:disable Metrics/PerceivedComplexity
          @disabled_sources ||=
            doc.css("configuration > disabledPackageSources > add").map do |node|
              value = node.attribute("value")&.value ||
                      node.at_xpath("./value")&.content

              if value&.strip&.downcase == "true"
                node.attribute("key")&.value&.strip ||
                  node.at_xpath("./key")&.content&.strip
              end
            end
        end
```

I also made a small optimization by calling `&.strip&.downcase` on `value`

jurre:
Looks good 👍 I made one small suggestion inline

nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
AshleighAdams: Yes, I agree, that is much better, updated the PR with your suggested changes.

jurre (approved):

PR #3308, opened 2021-03-20
Fix the retries counter for github pull request

codisart:
Refactor tree creation in its own class

### DIFF
--- a/common/lib/dependabot/pull_request_creator/github.rb
+++ b/common/lib/dependabot/pull_request_creator/github.rb
@@ -3,11 +3,12 @@
 require "octokit"
 require "securerandom"
 require "dependabot/clients/github_with_retries"
+require "dependabot/pull_request/github"
 require "dependabot/pull_request_creator"
 require "dependabot/pull_request_creator/commit_signer"
+
 module Dependabot
   class PullRequestCreator
-    # rubocop:disable Metrics/ClassLength
     class Github
       MAX_PR_DESCRIPTION_LENGTH = 65_536 # characters (see #create_pull_request)
 
@@ -37,6 +38,9 @@ module Dependabot
         @assignees               = assignees
         @milestone               = milestone
         @require_up_to_date_base = require_up_to_date_base
+
+        @commit_creation = 0
+        @tree_creation = 0
       end
 
       def create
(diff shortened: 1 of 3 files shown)


common/lib/dependabot/pull_request_creator/github.rb
```diff
@@ -35,6 +36,9 @@ def initialize(source:, branch_name:, base_commit:, credentials:,
         @assignees               = assignees
         @milestone               = milestone
         @require_up_to_date_base = require_up_to_date_base
+
+        @commit_creation = 0
+        @tree_creation = 0
```
jurre, line 41: Great catch on this bug, but I still don't think this will actually ever increment the counter.

Number are immutable, so we can't update the value in place, so the `raise_or_increment_retry_counter` will never update the instance variable `@commit_creation`, we assign a new value to the local `counter` variable instead.

```ruby
irb(main):001:0> def raise_or_increment_retry_counter(counter:, limit:)
        counter ||= 0
        counter += 1
        raise if counter > limit
      end
=> :raise_or_increment_retry_counter
irb(main):007:0> @commit_creation = 0
=> 0
irb(main):008:0> raise_or_increment_retry_counter(counter: @commit_creation, limit: 3)
=> nil
irb(main):009:0> raise_or_increment_retry_counter(counter: @commit_creation, limit: 3)
=> nil
irb(main):010:0> raise_or_increment_retry_counter(counter: @commit_creation, limit: 3)
=> nil
irb(main):011:0> @commit_creation
=> 0
```

Let me try to show what happens:

```ruby
irb(main):001:0> def what_is_my_obj_id(counter:)
irb(main):002:1>   puts counter.object_id
irb(main):003:1>   counter += 1
irb(main):004:1>   puts counter.object_id
irb(main):005:1> end
=> :what_is_my_obj_id
irb(main):006:0> @counter = 0
=> 0
irb(main):007:0> @counter.object_id
=> 1
irb(main):008:0> what_is_my_obj_id(counter: @counter)
1
3
```

There are some "clever" things we could do by extracting a class that holds this state that we _can_ increment, but I think it would be better to just keep track of the retries in-line in the respective methods.

Lmk if you're up for fixing this, otherwise I'll put it on our backlog.

It'd be awesome to add a test for this, I'd be happy to help with that.

codisart: I up for this, it will be a good way for me to try to work with rspec.

jeffwidman:
@codisart are you still interested in tackling this?

codisart:
Hi @jeffwidman 

I will try to get back to it.

PR #3311, opened 2021-03-21
Improve maintainability of labeler class using polymorphism

codisart:
The more frequent code smells in the dependabot core code base are shotgun surgery and duplication. There is a lot of duplicate provider clients initialisation. I have tried to reduce this duplication since 2018 but it was very difficult to find a simple thread to pull without taking all the codebase in. I think this PR is a good example of what could be done to reduce the complexity and the duplication. 
After an evening tinkering with the common/pull_request_creator/labeler class, I think this polymorphism proposal is a good way to tackle the duplication from multiple providers. I hope this PR to be a base and to be able to find my way trough all the provider duplication.

### DIFF
--- a/common/lib/dependabot/pull_request_creator.rb
+++ b/common/lib/dependabot/pull_request_creator.rb
@@ -12,6 +12,7 @@ module Dependabot
     require "dependabot/pull_request_creator/message_builder"
     require "dependabot/pull_request_creator/branch_namer"
     require "dependabot/pull_request_creator/labeler"
+    require "dependabot/pull_request_creator/labelers/factory"
 
     class RepoNotFound < StandardError; end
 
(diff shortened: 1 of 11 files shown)


xlgmokha:
We love the enthusiasm to improve the codebase. Dependabot code has some areas that can benefit from improvement and we appreciate the time you took to provide this as a suggestion.

Please consider opening an issue in the [issue tracker](https://github.com/dependabot/dependabot-core/issues/new) so that we can get a better understanding of why these sorts of changes are necessary. We have varying levels of skill and providing the answer to `why?` over `what?` tends to yield better discussion in our experience.

After weighing the risk vs. value we've decided that we can't accept this contribution at this time.

PR #3315, opened 2021-03-22
Bundler 2: [Prerelease] Add native helpers for file parsing

feelepxyz:
Add native helpers for bundler 2 file parsing with testing setup to run bundler specs with both bundler 1 and 2.

The current setup runs all bundler tests twice, sequentially which will add around ~4 mins to the bundler test suite run time.

We could split and run these in parallel or make the bundler 2 tests optional somehow but this also complicates the test setup.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,24 +16,25 @@ jobs:
       fail-fast: false
       matrix:
         suite:
-          - bundler
-          - cargo
-          - common
-          - composer
-          - dep
-          - docker
-          - elm
-          - git_submodules
-          - github_actions
-          - go_modules
-          - gradle
-          - hex
-          - maven
-          - npm_and_yarn
-          - nuget
-          - omnibus
-          - python
-          - terraform
+        - { path: bundler, name: bundler1 }
+        - { path: bundler, name: bundler2 }
+        - { path: cargo, name: cargo }
+        - { path: common, name: common }
+        - { path: composer, name: composer }
+        - { path: dep, name: dep }
+        - { path: docker, name: docker }
+        - { path: elm, name: elm }
+        - { path: git_submodules, name: git_submodules }
+        - { path: github_actions, name: github_actions }
+        - { path: go_modules, name: go_modules }
+        - { path: gradle, name: gradle }
+        - { path: hex, name: hex }
+        - { path: maven, name: maven }
+        - { path: npm_and_yarn, name: npm_and_yarn }
+        - { path: nuget, name: nuget }
+        - { path: omnibus, name: omnibus }
+        - { path: python, name: python }
+        - { path: terraform, name: terraform }
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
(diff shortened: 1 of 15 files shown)


bundler/helpers/v2/lib/functions/file_parser.rb
```diff
@@ -0,0 +1,106 @@
...
+        dependencies.
+        map(&method(:serialize_bundler_dependency))
+    end
+
+    private
+
+    def lockfile
+      return @lockfile if defined?(@lockfile)
+
+      @lockfile =
```
codisart, line 27: je crois qu'on peut remplacer le return par 
`@lockfile ||= `

feelepxyz: We actually want to keep the existing behaviour to memoize a `nil` value, `@lockfile ||=` won't memoize nil

.github/workflows/ci.yml
```diff
@@ -16,24 +16,25 @@ jobs:
...
-          - gradle
-          - hex
-          - maven
-          - npm_and_yarn
-          - nuget
-          - omnibus
-          - python
-          - terraform
+        - { path: bundler, name: bundler1 }
+        - { path: bundler, name: bundler2 }
```
feelepxyz, line 20: @brrygrdn @jurre looks like you can add hashes as the suite argument, thoughts on doing something like this?

jurre: Yep, I dig it!

bundler/helpers/v2/build
```diff
@@ -20,4 +19,5 @@ cd "$install_dir"
 
 # NOTE: Sets `BUNDLED WITH` to match the installed v1 version in Gemfile.lock
 # forcing specs and native helpers to run with the same version
-BUNDLER_VERSION=2 bundle install
+BUNDLER_VERSION=2 bundle config set --local path ".bundle"
```
jurre, line 22: Nice!

bundler/spec/spec_helper.rb
```diff
@@ -9,3 +9,24 @@ def require_common_spec(path)
 end
 
 require "#{common_dir}/spec/spec_helper.rb"
+
+def bundler_project_dependency_files(project)
+  bundler_version = bundler_2_available? ? "bundler2" : "bundler1"
+  project_dependency_files(File.join(bundler_version, project))
+end
```
jurre, line 16: Should we remove this until we actually need it?

jurre (approved):

.github/workflows/ci.yml
brrygrdn: 🎉 Nice

brrygrdn: Ah, the only downside of this approach is it seems to mess with our configured required builds, we'll need to flip the required builds from the old set to the new ones, merge this and then force all other PRs to merge master.

Not a big deal overall, as I think this is ultimately a better approach, just a one-time pain to switch over.

brrygrdn (approved):

.github/workflows/ci.yml
feelepxyz: Good catch! Will fix that up 👍

PR #3317, opened 2021-03-22
Attempt to fix builds running out of disk space

feelepxyz:
Looks like most builds on main are now running out of disk space when
`chown`ing: https://github.com/dependabot/dependabot-core/runs/2168114315

This is a quick fix to try and get builds green. We should look into
removing the need to chown files.


feelepxyz:
Merging as a hotfix to fix broken builds on master, we use this pattern in the updater with success

PR #3319, opened 2021-03-22
Bundler 2: [Prerelease] Add version resolver

feelepxyz:
Add native helpers for bundler 2 version resolver.

Added two `pending` specs as they require the `LatestVersionFinder`.

## TODO
- [x] Add specs when bundler 2 version is specified
- [x] Add specs when updating gems that require a specific major version of bundler, e.g. guard-bundler


bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
```diff
@@ -130,6 +130,67 @@
...
+        let(:dependency_files) { project_dependency_files("bundler2/bundler_specified") }
+        its([:version]) { is_expected.to eq(Gem::Version.new("1.4.0")) }
+
+        context "attempting to update Bundler" do
+          let(:dependency_name) { "bundler" }
+          let(:requirement_string) { "~> 2.2.0" }
+
+          let(:dependency_files) { project_dependency_files("bundler2/bundler_specified") }
+
+          it "returns nil as resolution returns the bundler version installed by core" do
```
feelepxyz, line 145: Ref: https://github.com/dependabot/dependabot-core/pull/3319/files#diff-8619130a1efd74403d2096bc77c13c862073216cecadb1ae2e00ec4a62070675R34

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
```diff
@@ -130,6 +130,67 @@
...
+        end
+      end
+
+      context "with a dependency that requiers bundler v2", :bundler_v2_only do
+        let(:dependency_name) { "guard-bundler" }
+        let(:requirement_string) { "3.0.0" }
+
+        let(:dependency_files) { project_dependency_files("bundler2/requires_bundler") }
+
+        pending "resolves version" do
```
feelepxyz, line 177: These specs require the LatestVersionFinder and all the dependency source native helpers to be added so left these as pending until we have that merged in.

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
```diff
@@ -384,18 +445,29 @@
             to eq(Gem::Version.new("2.0.1"))
         end
 
-        context "that isn't satisfied by the dependencies" do
+        context "that isn't satisfied by the dependencies", :bundler_v1_only do
           let(:dependency_files) do
             project_dependency_files("bundler1/imports_gemspec_version_clash_old_required_ruby_no_lockfile")
           end
-          let(:gemfile_fixture_name) { "imports_gemspec_version_clash" }
```
feelepxyz, line 391: Unused

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
```diff
@@ -384,18 +445,29 @@
...
           end
         end
+
+        context "that isn't satisfied by the dependencies", :bundler_v2_only do
+          let(:dependency_files) do
+            project_dependency_files("bundler2/imports_gemspec_version_clash_old_required_ruby_no_lockfile")
+          end
+          let(:current_version) { "3.0.1" }
+
+          it "raises a DependencyFileNotResolvable error" do
```
feelepxyz, line 466: Previously it looked like bundler v1 `ignores the minimum ruby version in the gemspec` (see spec above), I think this new behaviour is correct as the currently installed version of `statesman` doesn't satisfy the ruby version in the project gemspec.

bundler/spec/dependabot/bundler/update_checker_spec.rb
```diff
@@ -2074,98 +2074,4 @@
       it { is_expected.to eq(false) }
     end
   end
-
-  context "with bundler 2 support enabled" do
```
feelepxyz, line 2078: All these specs fail because the v2 version resolver is used and returns errors.

bundler/spec/dependabot/bundler/file_updater_spec.rb
```diff
@@ -1708,29 +1708,4 @@
       end
     end
   end
-
-  context "with bundler 2 support enabled" do
```
feelepxyz, line 1712: These specs fail because the v2 version resolver is used and returns errors.

bundler/helpers/v2/monkey_patches/definition_ruby_version_patch.rb
```diff
@@ -8,11 +8,11 @@ def index
       if ruby_version
         requested_version = ruby_version.to_gem_version_with_patchlevel
         sources.metadata_source.specs <<
-          Gem::Specification.new("ruby\0", requested_version)
+          Gem::Specification.new("Ruby\0", requested_version)
```
feelepxyz, line 11: Looks like it changed here: https://github.com/rubygems/rubygems/commit/4e950f56d61fc0c276651cefe59ccd87b8a465ad

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
feelepxyz: This is a bummer, it doesn't seem like we can currently update bundler at all.

When running a dry-run to update a gemfile with bundler v1 it tries to go to v2 even though the project has been bundled with v1, it looks like it falls through here atm: https://github.com/dependabot/dependabot-core/blob/main/bundler/helpers/v1/lib/functions/version_resolver.rb#L24 because `bundler` isn't listed as one of the gems in the definition.

When running the v2 helpers, `bundler` is listed as a gem but we [return nil if the gem name is `bundler`](https://github.com/dependabot/dependabot-core/pull/3319/files#diff-8619130a1efd74403d2096bc77c13c862073216cecadb1ae2e00ec4a62070675R34), I think this is the best thing for now as the "resolved version" is core's bundler version, not the latest resolvable version.

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
feelepxyz: Should be able to remove `pending` once this is in https://github.com/dependabot/dependabot-core/pull/3327/

jurre (approved):

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
jurre: I'll do this downstream in a separate PR 👍

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
```diff
@@ -130,6 +130,67 @@
...
+        let(:requirement_string) { "3.0.0" }
+
+        let(:dependency_files) { project_dependency_files("bundler2/requires_bundler") }
+
+        pending "resolves version" do
+          expect(subject.version).to eq(Gem::Version.new("3.0.0"))
+        end
+      end
+
+      context "when bundled with v2 and requesting a version that requires bundler v1", :bundler_v2_only do
```
feelepxyz, line 182: @jurre looks like this spec is not actually raising, not sure there's anything wrong with that though so probably just a bogus test case that can be removed.

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
```diff
@@ -130,6 +130,67 @@
...
+      end
+
+      context "with a dependency that requiers bundler v2", :bundler_v2_only do
+        let(:dependency_name) { "guard-bundler" }
+        let(:requirement_string) { "3.0.0" }
+
+        let(:dependency_files) { project_dependency_files("bundler2/requires_bundler") }
+
+        pending "resolves version" do
+          expect(subject.version).to eq(Gem::Version.new("3.0.0"))
```
feelepxyz, line 178: This should be `expect(subject[:version]).to eq(Gem::Version.new("3.0.0"))`

jurre: Cheers 👍

PR #3323, opened 2021-03-23
Reverts #3320 bringing back the changes from #3279

feelepxyz:
WIP: Attempting to redo the docker build without `chown`ing a bunch of files which was causing the image build to run out of disk space.


feelepxyz:
We'll probably not need to do this once https://github.com/dependabot/dependabot-core/pull/3398 lands

feelepxyz:
This is now done here https://github.com/dependabot/dependabot-core/pull/3398/files

PR #3326, opened 2021-03-23
Bundler 2: Add native helpers for LockfileUpdater

jurre:
Add native helpers for bundler 2 lockfile updater.

### DIFF
--- a/bundler/helpers/v2/lib/functions.rb
+++ b/bundler/helpers/v2/lib/functions.rb
@@ -1,30 +1,41 @@
-require "functions/file_parser"
+# frozen_string_literal: true
+
 require "functions/conflicting_dependency_resolver"
+require "functions/file_parser"
+require "functions/lockfile_updater"
 
 module Functions
   class NotImplementedError < StandardError; end
 
   def self.parsed_gemfile(lockfile_name:, gemfile_name:, dir:)
     set_bundler_flags_and_credentials(dir: dir, credentials: [],
-      using_bundler2: false)
+                                      using_bundler2: false)
     FileParser.new(lockfile_name: lockfile_name).
       parsed_gemfile(gemfile_name: gemfile_name)
   end
 
   def self.parsed_gemspec(lockfile_name:, gemspec_name:, dir:)
     set_bundler_flags_and_credentials(dir: dir, credentials: [],
-      using_bundler2: false)
+                                      using_bundler2: false)
     FileParser.new(lockfile_name: lockfile_name).
       parsed_gemspec(gemspec_name: gemspec_name)
   end
 
   def self.vendor_cache_dir(dir:)
-    raise NotImplementedError, "Bundler 2 adapter does not yet implement #{__method__}"
+    set_bundler_flags_and_credentials(dir: dir, credentials: [],
+                                      using_bundler2: false)
+    Bundler.app_cache
   end
 
   def self.update_lockfile(dir:, gemfile_name:, lockfile_name:, using_bundler2:,
                            credentials:, dependencies:)
-    raise NotImplementedError, "Bundler 2 adapter does not yet implement #{__method__}"
+    set_bundler_flags_and_credentials(dir: dir, credentials: credentials,
+                                      using_bundler2: using_bundler2)
+    LockfileUpdater.new(
+      gemfile_name: gemfile_name,
+      lockfile_name: lockfile_name,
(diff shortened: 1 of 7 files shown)


bundler/spec/dependabot/bundler/file_updater_spec.rb
```diff
@@ -878,6 +881,56 @@
...
+              source: {
+                type: "git",
+                url: "https://github.com/elastic/elasticsearch-ruby.git",
+                branch: "5.x",
+                ref: "5.x"
+              }
+            }]
+          end
+
+          it "updates the dependency's revision" do
```
feelepxyz, line 913: Is this true? 😅 seems like the test is checking if we keep the same revision as the old lockfile? 🤔 

Looks like the requirements are not being updated at all so not clear what's being tested. What happens if we do end up updating the requirement? Do we need to unlock the other gems also?

bundler/spec/dependabot/bundler/file_updater_spec.rb
```diff
@@ -1709,27 +1762,23 @@
...
-      )
-    end
-
+  context "with bundler 2 support enabled", :bundler_v2_only do
     describe "updated_dependency_files" do
-      it "fails as the native helper is not yet implemented" do
-        expect { updater.updated_dependency_files }.
-          to raise_error(Dependabot::NotImplemented,
-                         "Bundler 2 adapter does not yet implement update_lockfile")
+      it "updates the dependencies" do
```
feelepxyz, line 1767: Is this case already tested elsewhere for bundler v2? Could we ✂️ ?

bundler/spec/dependabot/bundler/file_updater_spec.rb
```diff
@@ -1709,27 +1762,23 @@
...
+
+      context "with a dependency that requiers bundler v2" do
+        let(:dependency_files) { project_dependency_files("bundler2/requires_bundler") }
+        let(:dependency_name) { "guard-bundler" }
+        let(:requirements) do
+          [{ file: "Gemfile", requirement: "3.0.0", groups: [], source: nil }]
+        end
+
+        it "updates the version" do
+        end
```
feelepxyz, line 1781: Looks like the expectation is missing

bundler/helpers/v2/lib/functions.rb
```diff
@@ -1,29 +1,40 @@
+# frozen_string_literal: true
```
feelepxyz, line 1: Thanks for fixing the rubocops ✨

bundler/spec/dependabot/bundler/file_updater_spec.rb
jurre: This was mostly just adapted from the v1 version, but I think what this is supposed to test is that we keep the same remote after the update, but that it does not have the same revision:

```
            expect(new_remote_line).to eq(original_remote_line)
            expect(new_revision_line).to_not eq(original_revision_line)
```

This seems right to me?

bundler/spec/dependabot/bundler/file_updater_spec.rb
jurre: I think so, yes! ✨

bundler/spec/dependabot/bundler/file_updater_spec.rb
feelepxyz: Oh yes you're right, totally missed the `to_not eq` bit! 👍

feelepxyz (approved):
LGTM 👍  think you also need to yank out these specs https://github.com/dependabot/dependabot-core/blob/main/bundler/helpers/v2/spec/functions_spec.rb#L8-L9

bundler/spec/dependabot/bundler/file_updater_spec.rb
jurre: Let's chance it to `not_to` 👍

bundler/helpers/v2/lib/functions.rb
feelepxyz, line 1: Looks like my merge on main created a merge conflict in this file

jurre:
The python failures seem unrelated, and with that suite taking 60m at the minute (💀), I'm gonna go ahead and merge this. If those tests continue to fail and are not a flake, I'll open a separate PR to fix

PR #3327, opened 2021-03-23
Bundler 2: [Prerelease] Implement v2  native functions for LatestVersionFinder

brrygrdn:
Adds the DependencySource class back into the v2 helpers without any modifications, tests that it works with bundler v2.

### Notes

This may be better merging into https://github.com/dependabot/dependabot-core/pull/3319 since it is a downstream component

### TODOs

- [ ] ~Update specs to use project-based fixtures.~

As per comment below, we'll follow up with those separately to unblock #3319


feelepxyz (commented):
Sweet! Do all specs pass with v2? I reckon we should get this in so I can remove the `pending` specs in https://github.com/dependabot/dependabot-core/pull/3319/files 

Thoughts on doing a follow up PR to move to project based fixtures?

bundler/helpers/v2/lib/functions/dependency_source.rb
```diff
@@ -0,0 +1,86 @@
+module Functions
```
feelepxyz, line 1: Could also copy these native helper specs over for v2 https://github.com/dependabot/dependabot-core/blob/main/bundler/helpers/v1/spec/functions/dependency_source_spec.rb

brrygrdn:
@feelepxyz Yeah, that sounds good, I'll add the native helper specs I missed and make sure those pass, then we should merge I think

bundler/helpers/v2/spec/functions/dependency_source_spec.rb
```diff
@@ -0,0 +1,187 @@
...
+          with(basic_auth: ["SECRET_CODES", ""]).
+          to_return(status: 401)
+        stub_request(:get, registry_url + "specs.4.8.gz").
+          with(basic_auth: ["SECRET_CODES", ""]).
+          to_return(status: 401)
+      end
+
+      it "blows up with a useful error" do
+        error_class = Bundler::Fetcher::AuthenticationRequiredError
+        error_message = "Authentication is required for repo.fury.io"
```
feelepxyz, line 76: I have a vague memory of these specs failing because the error_class has changed

Looks like that's handled here https://github.com/dependabot/dependabot-core/blob/cbd62f2611af242457227d49add1f1a88ec38542/bundler/lib/dependabot/bundler/update_checker/shared_bundler_helpers.rb#L130

Also tested here https://github.com/dependabot/dependabot-core/blob/cbd62f2611af242457227d49add1f1a88ec38542/bundler/spec/dependabot/bundler/update_checker/latest_version_finder_spec.rb#L286-L297

Kinda annoying that we do regex parsing on the message, which probably means handling the new error outside of the native helper. Maybe one to revisit and refactor, would be neat if the native helper returned a formatted error instead.

brrygrdn: Yeah, that's odd as it passed for me locally in a container but not on CI. It's failing for me on my native rubyenv now too, will fix.

brrygrdn: Actually, I think I mispoke, running the tests natively on my Mac seems to result in bundler 1 being used in the v2 native helper 🤔 

I see `spec/dependabot/bundler/update_checker/latest_version_finder_spec.rb` fail two tests on:

```ruby
     Failure/Error:
       expect { finder.latest_version_details }.
         to raise_error do |error|
           expect(error).to be_a Dependabot::GitDependencyReferenceNotFound
           expect(error.dependency).to eq("business")
         end

       expected #<Dependabot::SharedHelpers::HelperSubprocessFailed: private method `URI' called for Bundler:Module> to be a kind of Dependabot::GitDependencyReferenceNotFound
```

I think this is likely working as intended as my native Mac hasn't been provisioned properly to run different bundler versions in a contained way but it is getting kind of tricky to reason about the 'right' test environment.

feelepxyz: Ahh yes it's a bit of a minefield, I find it works locally if I've installed both, e.g. `gem install bundler:2.2.15 && gem install bundler:1.17.3`

brrygrdn: I put together a small patch to the native helpers after hitting the above:
https://github.com/dependabot/dependabot-core/pull/3330

I think this is a safe sanity measure _but_ it also proved that the failures in `latest_version_finder_spec.rb` I am seeing locally don't hit this new error, meaning the native helper is using Bundler v2. This failure hasn't turned up in CI as it's blocked by the dependency source failure in the v2 native helpers, so I think there's actually two 'real' bugs.

brrygrdn: 👍🏻  Yeah, it is.

feelepxyz: Ah but from bundler v1? I changed that monkey patch slightly from the v1 implementation but might still be borked as I only tested it in a dry-run https://github.com/dependabot/dependabot-core/blob/main/bundler/helpers/v1/monkey_patches/git_source_patch.rb#L17

brrygrdn: 👍🏻  I've walked that back in https://github.com/dependabot/dependabot-core/pull/3327/commits/18690c8aaca57ef8e4b1078bb7bf01797941f119 and it passes for me locally

brrygrdn: Ok, figured out what was going on here - I started using my native environment while waiting for Docker to rebuild which seemed find apart from the failure above. It turns out my native environment had Bundler 2.0.x installed from previous testing, Bundler::URI works in 2.1.x and above.

I'll open a separate PR to update `v2/run.rb` to raise if the Bundler version is lower than 2.1.x since it seems that is the actual API we are dependent on. It might be worth pushing that into the Gemfile itself by depending on bundler but that could get gnarly fast 😬

common/lib/dependabot/errors.rb
```diff
@@ -4,7 +4,7 @@
 
 module Dependabot
   class DependabotError < StandardError
-    BASIC_AUTH_REGEX = %r{://(?<auth>[^:]*:[^@%\s]+(@|%40))}.freeze
+    BASIC_AUTH_REGEX = %r{://(?<auth>[^@%\s]+(@|%40))}.freeze
```
brrygrdn, line 7: See: https://github.com/dependabot/dependabot-core/pull/3327/files#diff-8310fd74ffea4abeb1ff784b465b75bc8e6b549319d4038b1516a0dc957e08b0R352

For Bundler 1, if we get either `Bundler::Fetcher::AuthenticationRequiredError` or `Bundler::Fetcher::BadAuthenticationError`, we re-wrap them as `Dependabot::PrivateSourceAuthenticationFailure` with the `source` extracted from the Bundler error.

In Bundler v2, in either scenario, we get a `Bundler::Fetcher::BadAuthenticationError` and can strip the `source` out using `BundlerErrorPatterns::BAD_AUTH_REGEX`.

The only caveat is that if the user set a username without a password, i.e. `https://my-malformed-creds@example.com/username` we would leave the `my-malformed-creds` in the error message since there was not a `:` preceding it.

I figured the least surprising behaviour is just to strip anything before @ in a URI, which results in the new test case above.

bundler/helpers/v2/spec/functions/dependency_source_spec.rb
feelepxyz: Oh nice find! Yeah sounds like a good idea to check for >= 2.1 👍   nervous of adding `bundler` to the Gemfile, spotted here that dependabot can't upgrade it currently: https://github.com/dependabot/dependabot-core/pull/3319/files#r599508512

common/lib/dependabot/errors.rb
feelepxyz: > I figured the least surprising behaviour is just to strip anything before @ in a URI, which results in the new test case above.

This makes sense to me 👍

bundler/spec/dependabot/bundler/update_checker/latest_version_finder_spec.rb
```diff
@@ -318,6 +321,39 @@
...
+      context "that we don't have authentication details for", :bundler_v2_only do
+        let(:error_message) do
+          <<~ERR
+            Bad username or password for https://SECRET_CODES@repo.fury.io/greysteil/.
+            Please double-check your credentials and correct them.
+          ERR
+        end
+
+        let(:error_class) do
+          "Bundler::Fetcher::BadAuthenticationError"
```
feelepxyz, line 333: We'll also need to handle it the new error here https://github.com/dependabot/dependabot-core/blob/cbd62f2611af242457227d49add1f1a88ec38542/bundler/lib/dependabot/bundler/update_checker/shared_bundler_helpers.rb#L130 but that's probably integration tested in the update checker so maybe ok to punt on that change here?

feelepxyz (approved):
LGTM 👍

bundler/spec/dependabot/bundler/update_checker/latest_version_finder_spec.rb
brrygrdn: Good point, let me double check as I think my mocking is off.

brrygrdn: Ah, we handle this error already in that helper
https://github.com/dependabot/dependabot-core/blob/cbd62f2611af242457227d49add1f1a88ec38542/bundler/lib/dependabot/bundler/update_checker/shared_bundler_helpers.rb#L134

The diff between bundler 1 and bundler 2 is 'creds are missing' is now just the same as 'creds are bad', so we get the behaviour different handled for free with a very slight change in `error_details[:source]` for bundler 2 that is non-breaking.

feelepxyz: Yes totally missed that, nice one 👍

common/lib/dependabot/errors.rb
feelepxyz: Looks like this change broke a spec in common

feelepxyz: Pushed up a spec change

brrygrdn:
The NPM test failures do seem to be local to this branch, I suspect the change to Dependabot::Error might be involved

common/lib/dependabot/errors.rb
brrygrdn: I've walked back this change and just updated the test case to use a properly formed basic auth token.

This is a bit of a distraction to getting this merged as it seems this updated schema has the potential to mutate NPM dependencies which is why the npm tests start to fail.

PR #3330, opened 2021-03-23
Avoid subtle runtime failures by raising if a native helper is invoked with the wrong bundler version

brrygrdn:
I've ran into a few subtle bugs running tests where my environment wasn't set up properly and my native helpers were actually failing over to an unexpected bundler version.

Since a native helper running with the wrong bundler version is likely to fail in subtle ways that might be hard to diagnose, let's force a clear failure if/when it happens.

### DIFF
--- a/bundler/helpers/v1/run.rb
+++ b/bundler/helpers/v1/run.rb
@@ -11,11 +11,25 @@ require "git_source_patch"
 
 require "functions"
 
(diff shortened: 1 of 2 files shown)


feelepxyz (approved):
Great idea, thanks for adding this 🙇 👌

PR #3329, opened 2021-03-23
Bundler [Prerelease]: Add force updater helpers

feelepxyz:
Adding Bundler v2 native helpers for the ForceUpdater.

PR #3338, opened 2021-03-24
Attempt to fix npm install timing out ci builds

feelepxyz:
Looks like npm installing the npm_and_yarn helpers is timing out a fair
bit on ci:
https://github.com/dependabot/dependabot-core/pull/3329/checks?check_run_id=2183869598

Seeing errors like:

```
npm ERR! code ERR_SOCKET_TIMEOUT
npm ERR! errno ERR_SOCKET_TIMEOUT
npm ERR! request to https://registry.npmjs.org/@jest/reporters/-/reporters-26.6.2.tgz failed, reason: Socket timeout
```

I also disabled `npm audit` when isntalling which we never look at from
the build output.

### DIFF
--- a/npm_and_yarn/helpers/build
+++ b/npm_and_yarn/helpers/build
@@ -20,4 +20,4 @@ cp -r \
   "$install_dir"
 
 cd "$install_dir"
-npm ci --no-audit --fetch-timeout=600000
+npm ci --no-audit --fetch-timeout=600000 --fetch-retries=5
--- a/npm_and_yarn/helpers/build
+++ b/npm_and_yarn/helpers/build
@@ -20,4 +20,4 @@ cp -r \
   "$install_dir"
 
 cd "$install_dir"
-npm ci
+npm ci --no-audit --fetch-timeout=600000


npm_and_yarn/helpers/build
```diff
@@ -20,4 +20,4 @@ cp -r \
   "$install_dir"
 
 cd "$install_dir"
-npm ci
+npm ci --no-audit --fetch-timeout=600000
```
feelepxyz, line 23: Don't know if this will actually fix things, docs say the timeout is already 5 mins which I'm not sure is true from the build output. https://docs.npmjs.com/cli/v7/using-npm/config#fetch-timeout

thepwagner: https://docs.npmjs.com/cli/v7/using-npm/config#fetch-retries seems interesting too - we may want to raise that from the default of `2`.

thepwagner (approved):

npm_and_yarn/helpers/build
feelepxyz: Yeah will try `5` and see if the build failure goes away 👍

PR #3339, opened 2021-03-24
Make code coverage optional via an envvar

brrygrdn:
Calculating code coverage reports isn't free and our CI builds are quite long running as-is.

Let's turn this off for now to save some time as we are not actively pushing coverage right now - individual developers can still generate these reports via:

`COVERAGE=true bundle exec rspec spec`

We can review the status of coverage reporting after making some optimisation passes on builds.


feelepxyz (approved):
Makes sense for now 👍

PR #3344, opened 2021-03-24
Bundler req replacer not equal

thepwagner:
This adjusts the `Dependabot::Bundler::FileUpdater::RequirementReplacer` to adequately ignore `!=` operators within requirements.
Previously the implementation would produce invalid requirements (e.g. for the included test case: `~> 2.0.1, ! 2.0.3, ! 2.0.4`).

This also addresses a nit in the  `Dependabot::Bundler::FileUpdater::RequirementUpdater`: settling on `twiddle` instead of a mix of `twidle`/`twiddle`.

### DIFF
--- a/bundler/lib/dependabot/bundler/file_updater/requirement_replacer.rb
+++ b/bundler/lib/dependabot/bundler/file_updater/requirement_replacer.rb
@@ -167,6 +167,8 @@ module Dependabot
             req_string.include?(" ")
           end
 
+          EQUALITY_OPERATOR = /(?<![<>!])=/.freeze
+
           def use_equality_operator?(requirement_nodes)
             return true if requirement_nodes.none?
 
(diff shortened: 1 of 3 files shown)


bundler/lib/dependabot/bundler/file_updater/requirement_replacer.rb
```diff
@@ -203,7 +203,7 @@ def serialized_req(req, use_equality_operator)
             # Gem::Requirement serializes exact matches as a string starting
             # with `=`. We may need to remove that equality operator if it
             # wasn't used originally.
-            tmp_req = tmp_req.gsub(/(?<![<>])=/, "") unless use_equality_operator
+            tmp_req = tmp_req.gsub(/(?<![<>!])=/, "") unless use_equality_operator
```
jurre, line 206: Should we pull that regex out into a const? I notice we had to update it in two places 🤷

jurre (approved):
👌

PR #3349, opened 2021-03-25
Bundler 2 [pre-release]: Add UpdateChecker

jurre:
Implements remaining native helpers to get the UpdateChecker to work on v2

### DIFF
--- a/bundler/helpers/v2/spec/functions_spec.rb
+++ b/bundler/helpers/v2/spec/functions_spec.rb
@@ -5,9 +5,7 @@ require "native_spec_helper"
 RSpec.describe Functions do
   # Verify v1 method signatures are exist, but raise as NYI
   {
-    private_registry_versions: [:gemfile_name, :dependency_name, :dir, :credentials ],
-    jfrog_source: [:dir, :gemfile_name, :credentials, :using_bundler2],
-    git_specs: [:dir, :gemfile_name, :credentials, :using_bundler2],
+    jfrog_source: %i(dir gemfile_name credentials using_bundler2)
   }.each do |function, kwargs|
     describe "::#{function}" do
       let(:args) do
--- a/bundler/spec/dependabot/bundler/update_checker_spec.rb
+++ b/bundler/spec/dependabot/bundler/update_checker_spec.rb
@@ -16,7 +16,10 @@ RSpec.describe Dependabot::Bundler::UpdateChecker do
       dependency_files: dependency_files,
       credentials: credentials,
       ignored_versions: ignored_versions,
-      security_advisories: security_advisories
+      security_advisories: security_advisories,
+      options: {
+        bundler_2_available: bundler_2_available?
+      }
     )
   end
   let(:credentials) do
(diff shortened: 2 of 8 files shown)


bundler/spec/dependabot/bundler/update_checker_spec.rb
```diff
@@ -218,6 +221,47 @@
...
+      let(:gemfury_business_url) do
+        "https://repo.fury.io/greysteil/api/v1/dependencies?gems=business"
+      end
+      before do
+        # We only need to stub out the version callout since it would
+        # otherwise call out to the internet in a shell command
+        allow(Dependabot::Bundler::NativeHelpers).
+          to receive(:run_bundler_subprocess).
+          with({
+                 bundler_version: "2",
```
brrygrdn, line 245: Is this bundler version set in the stub the only difference between test version? WDYT think about just setting the expectation using something like:

```
bundler_version: bundler_2_available? ? "2" : "1"
```

We don't have the fixture swapping wired up based on `SUITE_NAME` but when we do, the gemfiles uses should be automagically pulled from the right folder, so it might save some test duplication.

bundler/spec/dependabot/bundler/update_checker_spec.rb
```diff
@@ -1282,14 +1326,22 @@
...
                 expect(error).to be_a(Dependabot::GitDependenciesNotReachable)
                 expect(error.dependency_urls).
                   to eq(["git@github.com:fundingcircle/prius"])
               end
           end
+
+          context "bundler v2", :bundler_v2_only do
+            let(:dependency_files) { project_dependency_files("bundler2/private_git_source") }
+
+            it "updates the dependency" do
```
brrygrdn, line 1341: 🎉

bundler/spec/fixtures/ruby/lockfiles/specified_source.lock
```diff
@@ -1,6 +1,6 @@
 GEM
   remote: https://rubygems.org/
-  remote: https://wxuokzLuQTRgMGtEYMPJ@repo.fury.io/greysteil/
+  remote: https://SECRET_CODES@repo.fury.io/greysteil/
```
brrygrdn, line 3: 😬  Thanks for ✂️  this

brrygrdn (approved):
LGTM, I have only one question and it's something we can punt on as we have other things to do to settle the test suite structure, otherwise 🚀

bundler/spec/dependabot/bundler/update_checker_spec.rb
jurre: The fixtures are also different between these tests, but I re-checked and it turns out they don't need to be, so let's go with your suggestion 💯

bundler/spec/dependabot/bundler/update_checker_spec.rb
feelepxyz: Nice!

bundler/helpers/v2/spec/functions_spec.rb
```diff
@@ -5,9 +5,7 @@
 RSpec.describe Functions do
   # Verify v1 method signatures are exist, but raise as NYI
   {
-    private_registry_versions: [:gemfile_name, :dependency_name, :dir, :credentials ],
-    jfrog_source: [:dir, :gemfile_name, :credentials, :using_bundler2],
-    git_specs: [:dir, :gemfile_name, :credentials, :using_bundler2],
+    jfrog_source: %i(dir gemfile_name credentials using_bundler2)
```
feelepxyz, line 8: I wonder if this function isn't tested anywhere?

feelepxyz (approved):
🎉

bundler/helpers/v2/spec/functions_spec.rb
brrygrdn: 😬  Possible, we should have cleared out this file by now, right?

jurre: I don't think it is, I'll add some coverage 👍

PR #3350, opened 2021-03-25
Bundler 2 [pre-release]: Detect bundler version to use when bundler 2 is available

brrygrdn:
Depends on https://github.com/dependabot/dependabot-core/pull/3406

This modifies the `bundler_2_available` option so it no longer forces use of the bundler 2 native helpers but rather decides based on the lockfile if present.

The primary change is to expand and test the `Dependabot::Bundler::Helper` to clearly specify scenarios such as defaulting to the newest when no Lockfile is present and failing over to v1 if a file is present but appears to pre-date the `BUNDLED_WITH` notation.

This also clarifies how Dependabot will behave if a new/malformed Gemfile.lock version were to appear.

### DIFF
--- a/bundler/helpers/v2/run.rb
+++ b/bundler/helpers/v2/run.rb
@@ -11,7 +11,7 @@ require "git_source_patch"
 
 require "functions"
 
-MIN_BUNDLER_VERSION = "2.0.0"
+MIN_BUNDLER_VERSION = "2.1.0"
 
 def validate_bundler_version!
   return true if correct_bundler_version?
--- a/bundler/lib/dependabot/bundler/helpers.rb
+++ b/bundler/lib/dependabot/bundler/helpers.rb
@@ -5,23 +5,36 @@ module Dependabot
     module Helpers
       V1 = "1"
       V2 = "2"
+      # If we are updating a project with no Gemfile.lock, we default to the
+      # newest version we support
+      DEFAULT = V2
+      # If we are updating a project with a Gemfile.lock that does not specify
+      # the version it was bundled with, with failover to V1 on the assumption
+      # it was created with an old version that didn't add this information
+      FAILOVER = V1
+
+      BUNDLER_MAJOR_VERSION_REGEX = /BUNDLED WITH\s+(?<version>\d+)\./m
 
       # NOTE: options is a manditory argument to ensure we pass it from all calling classes
-      def self.bundler_version(_lockfile, options:)
-        # For now, force V2 if bundler_2_available
-        return V2 if options[:bundler_2_available]
+      def self.bundler_version(lockfile, options:)
+        # TODO: Remove once bundler 2 is fully supported
+        return V1 unless options[:bundler_2_available]
+        return DEFAULT unless lockfile
 
-        # TODO: Add support for bundler v2 based on lockfile
-        # return V2 if lockfile.content.match?(/BUNDLED WITH\s+2/m)
-
-        V1
+        if matches = lockfile.content.match(BUNDLER_MAJOR_VERSION_REGEX)
+          matches[:version].to_i >= 2 ? V2 : V1
+        else
+          FAILOVER
+        end
       end
 
       def self.detected_bundler_version(lockfile)
         return "unknown" unless lockfile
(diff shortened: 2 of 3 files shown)


bundler/spec/spec_helper.rb
```diff
@@ -10,15 +10,30 @@ def require_common_spec(path)
 
 require "#{common_dir}/spec/spec_helper.rb"
 
-def bundler_2_available?
-  ENV["SUITE_NAME"] == "bundler2"
+module PackageManagerHelper
```
brrygrdn, line 13: This is an opinionated change, I tend to depart from the house style on preferring not to have test helper method defined globally.

That said, now there are a few helper methods related/derived from checking `ENV["SUITE_NAME"]`, I thought having a wrapper made much more sense.

bundler/spec/dependabot/bundler/update_checker/conflicting_dependency_resolver_spec.rb
```diff
@@ -19,7 +19,7 @@
         "username" => "x-access-token",
         "password" => "token"
       }],
-      options: { bundler_2_available: bundler_2_available? }
+      options: { bundler_2_available: PackageManagerHelper.use_bundler_2? }
```
feelepxyz, line 22: Won't we need to change all specs to have `BUNDLED WITH 2` when this is true? Maybe worth adding an override just for tests until we've converted specs?

brrygrdn: Yeah, my plan is to add a temporary patch to [this method](https://github.com/dependabot/dependabot-core/pull/3350/files#diff-1f5350b0c5588ffb399b6384bb37abc35ea941b5897989c74f2ed70c83a0c6b6R27) that rewrites the lockfile to `BUNDLED WITH 2` when this is true and get the tests to call project fixtures via that.

This probably cannot merge until we've sorted out the project fixtures, but I wanted to lay out the helper logic + tests now since the behaviour has a few quirks around defaulting/failing back.

brrygrdn: Oh I should mention, that rewrite should eventually just be replaced with straight up forking the project fixtures for clarity since there may be other tweaks/changes.

brrygrdn:
Odd, I didn't expect `bundler2` to pass, and it isn't passing for me locally 🤔

bundler/spec/dependabot/bundler/update_checker/conflicting_dependency_resolver_spec.rb
brrygrdn: I've ended up adding the v2 project fixtures as part of this PR as it turned out easy to implement using this toggle to find the specs that failed due to missing fixtures. It might need broken out for review purposes, but the guts of it should be done.

brrygrdn:
The only things failing on this PR now should be a handful of tests in the bundler2 build which aren't using project-based fixtures. Since that will make them easy to find, my plan is a follow up commit which fixes this we can optionally cherry-pick unto the main branch.

Almost there!

bundler/spec/spec_helper.rb
jurre: I also prefer this, thoughts on adding them to the rspec context though?

```ruby
RSpec.configure do |c|
  c.include PackageManagerHelper
end
```

bundler/helpers/v2/run.rb
```diff
@@ -11,7 +11,7 @@
 
 require "functions"
 
-MIN_BUNDLER_VERSION = "2.0.0"
+MIN_BUNDLER_VERSION = "2.1.0"
```
brrygrdn, line 14: We actually rely on the Bundler v2.1.0 API for Bundler::URI, which was private in 2.0.x

feelepxyz (approved):
🎉 :shipit:

feelepxyz:
🎉  Looks like all bundler tests are ✅  I would probably just merge and start the release over waiting for python to complete 👌

PR #3355, opened 2021-03-25
Bundler v2 [pre-release]: Add and test jfrog source helper

jurre:
Adds a basic test for the jfrog_source method. I tried to test this from
the core codebase where it's used as well, but due to the current setup
it's currently very hard to do that without connecting to an actual
jfrog server. Since we had no coverage for this before, I'm calling it
an improvement.

### DIFF
--- a/bundler/lib/dependabot/bundler/update_checker/shared_bundler_helpers.rb
+++ b/bundler/lib/dependabot/bundler/update_checker/shared_bundler_helpers.rb
@@ -187,7 +187,9 @@ module Dependabot
         end
 
         def jfrog_source
-          in_a_native_bundler_context(error_handling: false) do |dir|
+          return @jfrog_source unless defined?(@jfrog_source)
+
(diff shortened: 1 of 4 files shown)


bundler/lib/dependabot/bundler/update_checker/shared_bundler_helpers.rb
```diff
@@ -187,7 +187,9 @@ def inaccessible_git_dependencies
         end
 
         def jfrog_source
-          in_a_native_bundler_context(error_handling: false) do |dir|
+          return @jfrog_source unless defined?(@jfrog_source)
```
feelepxyz, line 190:
```suggestion
          return @jfrog_source if defined?(@jfrog_source)
```

jurre: 😮 thanks 😄

feelepxyz (approved):

brrygrdn (approved):
👍🏻  Thanks for adding this

PR #3354, opened 2021-03-25
v0.138.4

thepwagner:
Patch release containing [these commits](https://github.com/dependabot/dependabot-core/compare/v0.138.3...v0.138.4-release-notes)

And these PRs:
* https://github.com/dependabot/dependabot-core/pull/3326
* https://github.com/dependabot/dependabot-core/pull/3337
* https://github.com/dependabot/dependabot-core/pull/3301
* https://github.com/dependabot/dependabot-core/pull/3327
* https://github.com/dependabot/dependabot-core/pull/3319
* https://github.com/dependabot/dependabot-core/pull/3329
* https://github.com/dependabot/dependabot-core/pull/3338
* https://github.com/dependabot/dependabot-core/pull/3339
* https://github.com/dependabot/dependabot-core/pull/3344
* https://github.com/dependabot/dependabot-core/pull/3349
* https://github.com/dependabot/dependabot-core/pull/3331
* https://github.com/dependabot/dependabot-core/pull/3352

PR #3356, opened 2021-03-25
bump-version: re-tag the release notes ref

thepwagner:
Following the discussion in https://github.com/dependabot/dependabot-core/pull/3354#issuecomment-807056613 , let's prefer tagging the commit from the `-release-notes` branch  instead of tagging the merge commit.

This avoids potential races if the `main` branch changes outside the release notes PR's context (e.g. another PR is merged).


bin/bump-version.rb
```diff
@@ -52,7 +52,7 @@
 puts "git push origin HEAD:v#{new_version}-release-notes"
 puts "# ... create PR, verify, merge, for example:"
 puts "gh pr create"
-puts "git pull"
-puts "git tag 'v#{new_version}'"
-puts "git push --tags origin main"
+puts "# tag the approved release notes:"
+puts "git tag 'v#{new_version}' 'v#{new_version}-release-notes'"
```
thepwagner, line 56:
```suggestion
puts "git fetch"
puts "git tag 'v#{new_version}' 'origin/v#{new_version}-release-notes'"
```

This seemed right enough when I wrote it, but because of :52 this ref doesn't exist locally. I think referencing the remote branch is the best idea to keep this as something you can cleanly paste.

I also figured `git pull` wasn't necessary, but maybe the release notes were appended during review (e.g. via a suggestion comment like this!).

feelepxyz (approved):
👍

PR #3362, opened 2021-03-26
Update Rust to 1.51.0

CryZe:
With Rust 1.51.0 cargo now supports resolver = "2", which dependabot currently fails to parse.

https://github.com/LiveSplit/livesplit-core/issues/417

```
error: failed to parse manifest at `/home/dependabot/dependabot-updater/dependabot_tmp_dir/Cargo.toml`

Caused by:
  feature `resolver` is required

  this Cargo does not support nightly features, but if you
  switch to nightly channel you can add
  `cargo-features = ["resolver"]` to enable this feature
  ````

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -170,11 +170,11 @@ RUN wget https://packages.erlang-solutions.com/erlang-solutions_1.0_all.deb \
 
 ### RUST
 
-# Install Rust 1.47.0
+# Install Rust 1.51.0
 ENV RUSTUP_HOME=/opt/rust \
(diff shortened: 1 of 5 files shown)


jurre:
Thanks! There are two failing tests that look like they may just be things that are resolved in this version of the toolchain ( didn't dive into it yet), and I think it'd be good to add a case for ` resolver = "2"`. @CryZe is that something you'd be willing to take on?

CryZe:
Yeah I'll take a look

jurre (approved):
Thanks for tackling this!

PR #3363, opened 2021-03-26
Repurpose custom_headers for github client

honnix:
According to https://github.com/octokit/octokit.rb/blob/d56af021d21d9a8f623e0b41cac073c272864ec2/lib/octokit/client/pull_requests.rb#L52, github API expects `options` as general purpose JSON body extension, passing down `headers` won't make any impact. This PR repurposes `custom_headers` to pass additional JSON body fields.

### DIFF
--- a/common/lib/dependabot/pull_request_creator/github.rb
+++ b/common/lib/dependabot/pull_request_creator/github.rb
@@ -353,7 +353,7 @@ module Dependabot
           branch_name,
           pr_name,
           pr_description,
-          headers: custom_headers || {}
+          options: custom_headers || {}
         )
       rescue Octokit::UnprocessableEntity => e
         return handle_pr_creation_error(e) if e.message.include? "Error summary"


jurre:
The naming might be a bit confusing, but the `options` parameter in the GH client is not a keyword argument, but rather a regular param with a default empty hash. The `headers: ` that we pass is a hash argument, and the `headers` key is pulled out of there to add custom headers to the request here: https://github.com/octokit/octokit.rb/blob/d56af021d21d9a8f623e0b41cac073c272864ec2/lib/octokit/connection.rb#L152

jurre:
Actually, did you run into any issues passing in custom headers? Might be something else going on, feel free to reopen in that case

honnix:
@jurre Maybe it's easier to explain what I wanted to achieve. I would like dependabot-core to create a draft PR, so to achieve that I will need to pass down an additional `draft` flag in POST body.

honnix:
Thanks for explaining `options` btw, I think now I understand its purpose and usage.

honnix:
I guess I can patch it in this way?

```ruby
module Dependabot
  class PullRequestCreator
    class Github
      private

      def create_pull_request
        github_client_for_source.create_pull_request(
          source.repo,
          target_branch,
          branch_name,
          pr_name,
          pr_description,
          headers: custom_headers || {},
          draft: 'true'
        )
      rescue Octokit::UnprocessableEntity => e
        return handle_pr_creation_error(e) if e.message.include? 'Error summary'

        # Sometimes PR creation fails with no details (presumably because the
        # details are internal). It doesn't hurt to retry in these cases, in
        # case the cause is a race.
        retrying_pr_creation ||= false
        raise if retrying_pr_creation

        retrying_pr_creation = true
        retry
      end
    end
  end
end
```

jurre:
> @jurre Maybe it's easier to explain what I wanted to achieve. I would like dependabot-core to create a draft PR, so to achieve that I will need to pass down an additional `draft` flag in POST body.

Ah, yeah I'm afraid you'd have to patch the `#create_pull_request` method to add it, there are a few ways to do it, your suggestion would work, you could also subclass (might be a bit cleaner, easier to figure out what's going on for other devs in the future, it can be hard to spot that a class was patched), but it would require a bit more setup to wire up the right class

honnix:
Thanks. Do you think it is possible to support this in dependabot-core? For example introducing `options` the same way as octokit. To keep backward compatibility, `custom_headers` can still be kept.

honnix:
For example, something like: https://github.com/dependabot/dependabot-core/compare/main...honnix:pass-options?expand=1

jurre:
> For example, something like: https://github.com/dependabot/dependabot-core/compare/main...honnix:pass-options?expand=1

Yeah I'd be open to that. WDYT about making it a regular arument with a default of false? I'm a little hesitant having a catch all `other_options`, although it does provide a lot of flexibility, it's not super clear what it is or how it should be used.

honnix:
Right, I was thinking about that approach too, but it could be a bit weird to pass in this argument from the factory class because maybe not all underlying systems support it (github and gitlab support drafting I think, haven't checked others).

PR #3366, opened 2021-03-26
Add support for updating Maven extensions.xml files

britter:
These files contain dependency coordinates for Maven extensions that
want to receive the Maven afterSessionStart event. The file is always
located in the .mvn folder in the project root and is optional.

See also
  - https://maven.apache.org/guides/mini/guide-using-extensions.html
  - https://maven.apache.org/examples/maven-3-lifecycle-extensions.html

### DIFF
--- a/maven/lib/dependabot/maven/file_fetcher.rb
+++ b/maven/lib/dependabot/maven/file_fetcher.rb
@@ -37,11 +37,11 @@ module Dependabot
         return @extensions if defined?(@extensions)
         return @extensions if defined?(@extensions)
 
-          begin
-            fetch_file_if_present(".mvn/extensions.xml")
-          rescue Dependabot::DependencyFileNotFound
-            nil
-          end
+        begin
+          fetch_file_if_present(".mvn/extensions.xml")
+        rescue Dependabot::DependencyFileNotFound
+          nil
+        end
       end
 
       def child_poms
--- a/maven/lib/dependabot/maven/file_fetcher.rb
+++ b/maven/lib/dependabot/maven/file_fetcher.rb
@@ -34,7 +34,15 @@ module Dependabot
       end
 
       def extensions
-        @extensions ||= fetch_file_if_present(".mvn/extensions.xml")
+        return @extensions if defined?(@extensions)
+        
+        @extensions =
+          begin
+            fetch_file_if_present(".mvn/extensions.xml")
+          rescue Dependabot::DependencyFileNotFound
+            nil
+          end
+
       end
 
       def child_poms
--- a/maven/lib/dependabot/maven/file_fetcher.rb
+++ b/maven/lib/dependabot/maven/file_fetcher.rb
@@ -34,10 +34,7 @@ module Dependabot
       end
 
       def extensions
-        @extensions ||= fetch_file_from_host(".mvn/extensions.xml")
-      rescue Dependabot::DependencyFileNotFound
-        # extensions.xml is optional
-        nil
+        @extensions ||= fetch_file_if_present(".mvn/extensions.xml")
       end
 
       def child_poms
--- a/maven/lib/dependabot/maven/file_fetcher.rb
+++ b/maven/lib/dependabot/maven/file_fetcher.rb
@@ -35,7 +35,6 @@ module Dependabot
 
       def extensions
         return @extensions if defined?(@extensions)
-        
         @extensions =
           begin
             fetch_file_if_present(".mvn/extensions.xml")
--- a/maven/lib/dependabot/maven/file_fetcher.rb
+++ b/maven/lib/dependabot/maven/file_fetcher.rb
@@ -42,7 +42,6 @@ module Dependabot
           rescue Dependabot::DependencyFileNotFound
             nil
           end
-
       end
 
       def child_poms
--- a/maven/lib/dependabot/maven/file_fetcher.rb
+++ b/maven/lib/dependabot/maven/file_fetcher.rb
@@ -25,7 +25,7 @@ module Dependabot
         fetched_files << pom
         fetched_files += child_poms
         fetched_files += relative_path_parents(fetched_files)
-        fetched_files << extensions unless extensions.nil?
+        fetched_files << extensions if extensions
         fetched_files.uniq
       end
 
--- a/maven/lib/dependabot/maven/file_fetcher.rb
+++ b/maven/lib/dependabot/maven/file_fetcher.rb
@@ -25,9 +25,7 @@ module Dependabot
         fetched_files << pom
         fetched_files += child_poms
         fetched_files += relative_path_parents(fetched_files)
-        unless extensions.nil?
-          fetched_files << extensions
-        end
+        fetched_files << extensions unless extensions.nil?
         fetched_files.uniq
       end
 
--- a/maven/lib/dependabot/maven/file_fetcher.rb
+++ b/maven/lib/dependabot/maven/file_fetcher.rb
@@ -25,11 +25,21 @@ module Dependabot
         fetched_files << pom
         fetched_files += child_poms
         fetched_files += relative_path_parents(fetched_files)
+        unless extensions.nil?
+          fetched_files << extensions
+        end
         fetched_files.uniq
       end
 
       def pom
         @pom ||= fetch_file_from_host("pom.xml")
+      end
+
+      def extensions
+        @extensions ||= fetch_file_from_host(".mvn/extensions.xml")
+      rescue Dependabot::DependencyFileNotFound
+        # extensions.xml is optional
+        nil
       end
 
       def child_poms
--- a/maven/lib/dependabot/maven/file_fetcher.rb
+++ b/maven/lib/dependabot/maven/file_fetcher.rb
@@ -35,7 +35,8 @@ module Dependabot
 
       def extensions
         return @extensions if defined?(@extensions)
-        @extensions =
+        return @extensions if defined?(@extensions)
+
           begin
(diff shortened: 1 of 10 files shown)


maven/lib/dependabot/maven/file_updater.rb
```diff
@@ -11,24 +11,27 @@ class FileUpdater < Dependabot::FileUpdaters::Base
       require_relative "file_updater/property_value_updater"
 
       def self.updated_files_regex
-        [/^pom\.xml$/, %r{/pom\.xml$}]
+        [
+          /^pom\.xml$/, %r{/pom\.xml$},
+          /^extensions.\.xml$/, %r{/extensions\.xml$},
```
britter, line 16: I'm not sure this is correct

maven/lib/dependabot/maven/file_updater.rb
```diff
@@ -11,24 +11,27 @@ class FileUpdater < Dependabot::FileUpdaters::Base
...
-            pomfiles: updated_files,
+          updated_files = update_files_for_dependency(
+            original_files: updated_files,
             dependency: dependency
           )
         end
 
-        updated_files.select! { |f| f.name.end_with?("pom.xml") }
+        updated_files.select! { |f| f.name.end_with?("pom.xml") || f.name.end_with?("extensions.xml") }
         updated_files.reject! { |f| original_pomfiles.include?(f) }
```
britter, line 35: Not sure the reject part is correct. Should this also look for extensions.xml files?

maven/lib/dependabot/maven/file_updater.rb
```diff
@@ -40,18 +43,18 @@ def updated_dependency_files
       private
 
       def check_required_files
-        raise "No pom.xml!" unless get_original_file("pom.xml")
+        raise "No pom.xml!" unless get_original_file("pom.xml") || get_original_file("extensions.xml")
```
britter, line 46: This should probably be done differently.

britter (commented):
I'm just a Java developer and really have no idea what I'm doing here 😅 The tests helped me find the right places to modify and I tried to cover the new functionality with some new tests. Hope this is going in the right direction and I'm looking for somebody with more Ruby knowledge to polish this.

britter:
@jurre can you find the right person to provide some feedback on this?

maven/lib/dependabot/maven/file_fetcher.rb
```diff
@@ -25,13 +25,21 @@ def fetch_files
...
 
       def pom
         @pom ||= fetch_file_from_host("pom.xml")
       end
 
+      def extensions
+        @extensions ||= fetch_file_from_host(".mvn/extensions.xml")
+      rescue Dependabot::DependencyFileNotFound
+        # extensions.xml is optional
+        nil
```
feelepxyz, line 40:
```suggestion
        @extensions ||= fetch_file_if_present(".mvn/extensions.xml")
```

maven/lib/dependabot/maven/file_fetcher.rb
```diff
@@ -25,13 +25,21 @@ def fetch_files
         fetched_files << pom
         fetched_files += child_poms
         fetched_files += relative_path_parents(fetched_files)
+        fetched_files << extensions unless extensions.nil?
```
feelepxyz, line 28: Might be clearer as an `if`
```suggestion
        fetched_files << extensions if extensions
```

maven/lib/dependabot/maven/file_updater.rb
feelepxyz: Can you have a `extensions.xml` file without a `pom.xml`? Wondering if this should be left as it was, and only requiring `pom.xml` leaving extensions optional

maven/lib/dependabot/maven/file_updater.rb
feelepxyz: Ah yeah, it looks like we reject all files that haven't been updated, we might just be able to do this:
```suggestion
        updated_files.select! { |f| f.name.end_with?("pom.xml") || f.name.end_with?("extensions.xml") }
        updated_files.reject! { |f| dependency_files.include?(f) }
```

feelepxyz (commented):
@britter thanks for the contribution and sorry for the incredibly slow review! This looks good to me, left a few minor comments 👍

maven/lib/dependabot/maven/file_updater.rb
britter: You could have an extension that replaces `pom.xml` with something else. [polyglot-maven](https://github.com/takari/polyglot-maven) for example lets you write the build with languages other than xml. So could have a `pom.groovy` instead.

britter:
@feelepxyz thanks for the review! I've applied all but one change. PTAL.

britter:
@feelepxyz what's missing for this to get merged?

maven/lib/dependabot/maven/file_updater.rb
feelepxyz: Gotcha, although dependabot only supports `pom.xml` atm. I think it would be better to fail in the case where there's a `pom.groovy` and a `estensions.xml` as dependabot would ignore the groovy file and update the extensions file, which is probably not what you expect.

I think we should revert this change unless there's a way to have a `extensions.xml` file without a `pom.*` file.

maven/lib/dependabot/maven/file_fetcher.rb
```diff
@@ -25,13 +25,18 @@ def fetch_files
...
+        fetched_files << extensions if extensions
         fetched_files.uniq
       end
 
       def pom
         @pom ||= fetch_file_from_host("pom.xml")
       end
 
+      def extensions
+        @extensions ||= fetch_file_if_present(".mvn/extensions.xml")
```
feelepxyz, line 37: Spotted this fails when the repo doesn't include a `.mvn` directory
```suggestion
        return @extensions if defined?(@extensions)
        
        @extensions =
          begin
            fetch_file_if_present(".mvn/extensions.xml")
          rescue Dependabot::DependencyFileNotFound
            nil
          end

```

feelepxyz:
> @feelepxyz what's missing for this to get merged?

Sorry for dropping the ball, left a couple of comments.

britter:
@feelepxyz resolved your comment. Thanks for catching that one.

maven/lib/dependabot/maven/file_updater.rb
feelepxyz: @britter thoughts on reverting this to always require a `pom.xml`?

britter: @feelepxyz sorry, I missed this comment. I'm fine with reverting. I think would be possible to write an extension that makes a Maven build work without any other build file, but I don't think it's a relevant use case for this feature.

britter:
@feelepxyz I think the last comment was resolved. 👍🏻

feelepxyz:
Nice one! Forgot to approve the action, so just waiting for CI, there might be a few rubocop linting errors that pop up.

maven/lib/dependabot/maven/file_fetcher.rb
```diff
@@ -25,13 +25,26 @@ def fetch_files
...
+      def extensions
+        return @extensions if defined?(@extensions)
+        
+        @extensions =
+          begin
+            fetch_file_if_present(".mvn/extensions.xml")
+          rescue Dependabot::DependencyFileNotFound
+            nil
+          end
+
```
feelepxyz, line 45:
```suggestion
```

feelepxyz (approved):

maven/lib/dependabot/maven/file_fetcher.rb
```diff
@@ -25,13 +25,26 @@ def fetch_files
...
         fetched_files.uniq
       end
 
       def pom
         @pom ||= fetch_file_from_host("pom.xml")
       end
 
+      def extensions
+        return @extensions if defined?(@extensions)
+
```
feelepxyz, line 38:
```suggestion
```

britter:
@feelepxyz would you might triggering the CI pipeline again? Thanks!

xlgmokha:
It looks like the terraform failures have been fixed in the default branch. Pulling in the latest from main should fix the failing tests for the terraform build.

xlgmokha:
The maven job appears to be failing with a linter error:

```bash
Offenses:

lib/dependabot/maven/file_fetcher.rb:37:9: C: [Correctable] Layout/EmptyLineAfterGuardClause: Add empty line after guard clause.
        return @extensions if defined?(@extensions)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Give `$ bundle exec rubocop -a` a try and that should resolve the issue.

maven/lib/dependabot/maven/file_fetcher.rb
```diff
@@ -25,13 +25,24 @@ def fetch_files
...
         fetched_files.uniq
       end
 
       def pom
         @pom ||= fetch_file_from_host("pom.xml")
       end
 
+      def extensions
+        return @extensions if defined?(@extensions)
+        @extensions =
```
xlgmokha, line 38:
```suggestion
        return @extensions if defined?(@extensions)

```

xlgmokha (commented):
It looks like this PR is almost there.

1. Pull in the latest changes from `main`
2. Fix the linter error.
3. :shipit:

britter:
@xlgmokha @feelepxyz fixed the linter error and rebased on main branch. Thanks for the help!

PR #3368, opened 2021-03-26
Handle both ruby and package manager specific version requirements from ignore conditions

feelepxyz:
This is an attempt to handle both ruby and package manager specific version requirements from ignore conditions in the package manager specific requirement class.

We currently create ruby style ignore requirement when commenting on dependabot PRs with `£dependabot ignore this major version` for example. We also [say to use the package manager specific requirement syntax](https://docs.github.com/en/code-security/supply-chain-security/configuration-options-for-dependency-updates#specifying-dependencies-and-versions-to-ignore) when adding `ignore`s to the config file.

We also support specifying comma-separated ignore conditions as a single string, `> 1.2.3, <= 2.0.0` and split these when parsing the ignore conditions. This works for most ecosystems except for gradle/maven that have special range syntax, we used to handle this specifically when parsing ignore conditions: https://github.com/dependabot/dependabot-core/pull/3368/files#diff-3107de50e63836063ddedc0b51b50e47d1dba7d517f060e051b1b4d9494dd2abL131-L135

All requirement classes now support splitting comma-separated strings, whereas only some did this in the past and those who didn't would raise from `Gem::Requirement`.

Using the `requirements_array` method also means we'll properly deal with languages that support `OR` syntax, e.q. https://github.com/dependabot/dependabot-core/blob/main/maven/lib/dependabot/maven/requirement.rb#L28

### DIFF
--- a/bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
+++ b/bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
@@ -70,7 +70,7 @@ module Dependabot
 
         def filter_ignored_versions(versions_array)
           filtered = versions_array.
-                     reject { |v| ignore_reqs.any? { |r| r.satisfied_by?(v) } }
+                     reject { |v| ignore_requirements.any? { |r| r.satisfied_by?(v) } }
           raise AllVersionsIgnored if @raise_on_ignored && filtered.empty? && versions_array.any?
 
           filtered
@@ -110,8 +110,8 @@ module Dependabot
           )
         end
 
-        def ignore_reqs
-          ignored_versions.map { |req| Gem::Requirement.new(req.split(",")) }
+        def ignore_requirements
+          ignored_versions.flat_map { |req| requirement_class.requirements_array(req) }
         end
 
         def gemfile
--- a/docker/lib/dependabot/docker/update_checker.rb
+++ b/docker/lib/dependabot/docker/update_checker.rb
@@ -358,17 +358,11 @@ module Dependabot
           candidate_tags.
           reject do |tag|
             version = version_class.new(numeric_version_from(tag))
-            ignore_reqs.any? { |r| r.satisfied_by?(version) }
+            ignore_requirements.any? { |r| r.satisfied_by?(version) }
           end
         raise AllVersionsIgnored if @raise_on_ignored && filtered.empty? && candidate_tags.any?
 
         filtered
-      end
-
-      def ignore_reqs
-        # NOTE: we use Gem::Requirement here because ignore conditions will
-        # be passed as Ruby ranges
-        ignored_versions.map { |req| Gem::Requirement.new(req.split(",")) }
       end
     end
   end
--- a/elm/lib/dependabot/elm/update_checker.rb
+++ b/elm/lib/dependabot/elm/update_checker.rb
@@ -80,7 +80,7 @@ module Dependabot
 
       def candidate_versions
         filtered = all_versions.
-                   reject { |v| ignore_reqs.any? { |r| r.satisfied_by?(v) } }
+                   reject { |v| ignore_requirements.any? { |r| r.satisfied_by?(v) } }
 
         raise AllVersionsIgnored if @raise_on_ignored && filtered.empty? && all_versions.any?
 
@@ -117,12 +117,6 @@ module Dependabot
           map { |r| r.fetch(:requirement) }.
           map { |r| requirement_class.new(r) }.
           all? { |r| r.satisfied_by?(latest_version) }
-      end
-
-      def ignore_reqs
-        # NOTE: we use Gem::Requirement here because ignore conditions will
-        # be passed as Ruby ranges
-        ignored_versions.map { |req| Gem::Requirement.new(req.split(",")) }
       end
     end
   end
--- a/gradle/lib/dependabot/gradle/requirement.rb
+++ b/gradle/lib/dependabot/gradle/requirement.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "dependabot/utils"
+require "dependabot/maven/requirement"
 require "dependabot/gradle/version"
 
 module Dependabot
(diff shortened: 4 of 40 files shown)


bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
```diff
@@ -111,7 +111,7 @@ def dependency_source
         end
 
         def ignore_reqs
-          ignored_versions.map { |req| Gem::Requirement.new(req.split(",")) }
+          ignored_versions.flat_map { |req| requirement_class.requirements_array(req) }
```
feelepxyz, line 114: Tested https://github.com/dependabot/dependabot-core/blob/c325d573c3fde3ab4476d605e0ff213d26943fc5/bundler/spec/dependabot/bundler/update_checker/latest_version_finder_spec.rb#L107

python/lib/dependabot/python/update_checker/latest_version_finder.rb
```diff
@@ -227,7 +227,7 @@ def registry_index_response(index_url)
         end
 
         def ignore_reqs
-          ignored_versions.map { |req| requirement_class.new(req.split(",")) }
+          ignored_versions.flat_map { |req| requirement_class.requirements_array(req) }
```
feelepxyz, line 230: Tested: https://github.com/dependabot/dependabot-core/blob/473c8813c58d6dcb1dcd1706a8f64d013a83ebe7/python/spec/dependabot/python/update_checker/latest_version_finder_spec.rb#L197

maven/lib/dependabot/maven/update_checker/version_finder.rb
```diff
@@ -95,23 +93,17 @@ def filter_ignored_versions(possible_versions)
           filtered = possible_versions
 
           ignored_versions.each do |req|
-            ignore_req = Maven::Requirement.new(parse_requirement_string(req))
+            ignore_reqs = Maven::Requirement.requirements_array(req)
             filtered =
               filtered.
-              reject { |v| ignore_req.satisfied_by?(v.fetch(:version)) }
+              reject { |v| ignore_reqs.any? { |r| r.satisfied_by?(v.fetch(:version)) } }
```
feelepxyz, line 99: We currently test both maven and ruby ranges are supported https://github.com/dependabot/dependabot-core/blob/b0416b0f5287620dd40683f2860576f2b0b989fa/gradle/spec/dependabot/gradle/update_checker/version_finder_spec.rb#L118-L128

gradle/lib/dependabot/gradle/update_checker/version_finder.rb
```diff
@@ -99,10 +97,10 @@ def filter_ignored_versions(possible_versions)
           filtered = possible_versions
 
           ignored_versions.each do |req|
-            ignore_req = Gradle::Requirement.new(parse_requirement_string(req))
+            ignore_reqs = Gradle::Requirement.requirements_array(req)
             filtered =
               filtered.
-              reject { |v| ignore_req.satisfied_by?(v.fetch(:version)) }
+              reject { |v| ignore_reqs.any? { |r| r.satisfied_by?(v.fetch(:version)) } }
```
feelepxyz, line 103: We test both ruby and gradle requirement syntax here: https://github.com/dependabot/dependabot-core/blob/b0416b0f5287620dd40683f2860576f2b0b989fa/gradle/spec/dependabot/gradle/update_checker/version_finder_spec.rb#L118-L128

elm/lib/dependabot/elm/update_checker.rb
```diff
@@ -118,12 +118,6 @@ def requirements_up_to_date?
           map { |r| requirement_class.new(r) }.
           all? { |r| r.satisfied_by?(latest_version) }
       end
-
-      def ignore_reqs
```
feelepxyz, line 122: Defined in update checker base

docker/lib/dependabot/docker/update_checker.rb
```diff
@@ -364,12 +364,6 @@ def filter_ignored(candidate_tags)
 
         filtered
       end
-
-      def ignore_reqs
```
feelepxyz, line 368: Defined in update checker base

maven/lib/dependabot/maven/requirement.rb
```diff
@@ -32,7 +34,11 @@ def self.requirements_array(requirement_string)
 
       def initialize(*requirements)
         requirements = requirements.flatten.flat_map do |req_string|
-          convert_java_constraint_to_ruby_constraint(req_string)
+          if RUBY_VERSION_RESTRICTIONS.any? { |vr| req_string.include?(vr) }
```
feelepxyz, line 37: I attempted to reverse this and match on `MAVEN_RANGE_REGEX = /[\(\[].*,.*[\)\]]/` but we have a bunch of tests for version that don't match this that still get special maven/gradle treatment so opted for checking against known ruby version range restrictions: https://docs.ruby-lang.org/en/2.6.0/Gem/Requirement.html

thepwagner (approved):
LGTM: I wondered about reuse and one edge case, neither is particularly important.

gradle/lib/dependabot/gradle/requirement.rb
```diff
@@ -32,7 +33,13 @@ def self.requirements_array(requirement_string)
 
       def initialize(*requirements)
         requirements = requirements.flatten.flat_map do |req_string|
-          convert_java_constraint_to_ruby_constraint(req_string)
+          # NOTE: Support ruby-style version requirements that are created from
+          # PR ignore conditions
+          if Gem::Requirement::OPS.keys.any? { |vr| req_string.include?(vr) }
```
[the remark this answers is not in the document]
feelepxyz: > Do we have Requirement#PATTERN

Yes great suggestion, incorporated this 👍 

> Related, do we care about versions without an opcode? (1.2.3 is interpreted as = 1.2.3):

I don't think so, should be turned into a requirement that matches the exact version. Don't think this is changing from previous implementation.

PR #3369, opened 2021-03-26
Bump version: Remove merge commits

jurre:
I typically remove those manually as they don't add value to the changelog, but we may as well let the script do it for us


feelepxyz (approved):
💯

PR #3370, opened 2021-03-26
Convert the spec for Bundler::UpdateChecker to use project based fixtures.

brrygrdn:
This PR extracts from https://github.com/dependabot/dependabot-core/pull/3350

It prepares the way for bundler version detection by ensuring we can readily produce builds against lockfiles with the correct `BUNDLED WITH` values based on test parameters.

### Methodology

In order to verify the change over, I used a helper shim to force any tests that called `fixture` to load gemfiles to raise, e.g.

```
# bundler/spec/spec_helper.rb
alias non_project_fixture fixture

def fixture(*name)
  if name.any? { |folder| %w(gemfiles gemspecs lockfiles).include? folder }
    raise "Non-Project Fixture Loaded: '#{File.join(name)}'."
  end

  non_project_fixture(*name)
end
```

A large number of the `bundler1` fixtures already existed, so it was just a case of generating any missing files, checking their naming lined up and then finally grepping the file for `_fixture_name` to ensure no test cases were quietly failing over to the top-level dependency files.

### DIFF
--- a/bundler/helpers/v1/spec/native_spec_helper.rb
+++ b/bundler/helpers/v1/spec/native_spec_helper.rb
@@ -27,6 +27,9 @@ LOCKFILE_ENDING = /(?<ending>\s*(?:RUBY VERSION|BUNDLED WITH).*)/m.freeze
 
 def project_dependency_files(project)
   project_path = File.expand_path(File.join("../../spec/fixtures/projects/bundler1", project))
+
+  raise "Fixture does not exist for project: '#{project}'" unless Dir.exist?(project_path)
+
   Dir.chdir(project_path) do
     # NOTE: Include dotfiles (e.g. .npmrc)
     files = Dir.glob("**/*", File::FNM_DOTMATCH)
--- a/bundler/spec/spec_helper.rb
+++ b/bundler/spec/spec_helper.rb
@@ -14,6 +14,11 @@ def bundler_2_available?
   ENV["SUITE_NAME"] == "bundler2"
 end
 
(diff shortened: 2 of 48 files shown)


brrygrdn:
Aha, I forgot I'll need to commit a stubbed version of `bundler_project_dependency_files` for these tests to pass, will add that now

bundler/spec/spec_helper.rb
```diff
@@ -14,6 +14,11 @@ def bundler_2_available?
   ENV["SUITE_NAME"] == "bundler2"
 end
 
+# Load project files prepended with the bundler version, which is currently only ever bundler1
+def bundler_project_dependency_files(project)
+  project_dependency_files(File.join("bundler1", project))
+end
```
brrygrdn, line 20: This method provides a forward way of loading any fixtures modified in this PR, I left any existing calls to `project_dependency_files` alone to avoid bloating the diff.

jurre (approved):
Right on 🤙

I've reviewed the changes to the tests, and based on them passing trust that the fixtures are 💯

bundler/helpers/v1/spec/native_spec_helper.rb
```diff
@@ -27,6 +27,9 @@
 
 def project_dependency_files(project)
   project_path = File.expand_path(File.join("../../spec/fixtures/projects/bundler1", project))
+
+  raise "Fixture does not exist for project: '#{project}'" unless Dir.exist?(project_path)
```
feelepxyz, line 31: 👍

feelepxyz (approved):
🙌

PR #3383, opened 2021-03-29
Go mod: Handle multi-line error messages

feelepxyz:
Changes the go mod updater error handling to fall back to matching the
whole string if no lines where matched against the regex.

Some errors where surfacing with an empty error message as we initially
match on the whole string but filter per line.

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
@@ -269,22 +269,20 @@ module Dependabot
           write_go_mod(body)
         end
 
-        # rubocop:disable Metrics/AbcSize
-        # rubocop:disable Metrics/PerceivedComplexity
         def handle_subprocess_error(stderr)
           stderr = stderr.gsub(Dir.getwd, "")
 
           # Package version doesn't match the module major version
           error_regex = RESOLVABILITY_ERROR_REGEXES.find { |r| stderr =~ r }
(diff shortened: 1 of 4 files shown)


jurre (approved):

go_modules/spec/fixtures/projects/unknown_revision_version/go.sum
```diff
@@ -0,0 +1,13 @@
+github.com/fatih/Color v1.7.0 h1:GPsoW9Vx62R0G4RLveWNf9++7g1zBbzTiyuqVCNHhyw=
```
thepwagner, line 1: Uber-nit: this `go.sum` feels mighty invalid - I'm guessing it was copied from another fixture.

A couple of ideas:
* Since this project is invalid (e.g. `go mod tidy` will just exit without writing a `go.sum` file), just remove this file to align with that.
* Get this project into a working state with another version of `github.com/deislabs/oras`, or using a `require` directive to replace the invalid dependency. Store _that_ `go.sum`, then re-break the package.

I'd lean towards the (significantly faster) first.

thepwagner (approved):
Error handling change LGTM.

I flagged a smell in the new test harness, but am fine to ignore that.

go_modules/spec/fixtures/projects/unknown_revision_version/go.sum
feelepxyz: Good shout, removing `go.sum` 👍

PR #3384, opened 2021-03-29
Dry-run: cache clones in tmp/ and add a trace opt

feelepxyz:
Changes the clone destination to `./tmp/repo/name` and re-uses this if
`--cache=files` is passed in.

Adding a `--trace` option that prints the backtrace of rescued errors.

### DIFF
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -110,7 +110,8 @@ $options = {
   updater_options: {},
   security_advisories: [],
   security_updates_only: false,
-  pull_request: false
+  pull_request: false,
+  trace: false
 }
 
 unless ENV["LOCAL_GITHUB_ACCESS_TOKEN"].to_s.strip.empty?
@@ -212,6 +213,11 @@ option_parse = OptionParser.new do |opts|
   opts.on("--pull-request",
     "Output pull request information: title, description") do
     $options[:pull_request] = true
+  end
+
+  opts.on("--trace",
+    "Print error backtrace") do
+    $options[:trace] = true
   end
 end
 
@@ -432,6 +438,7 @@ def handle_dependabot_error(error:, dependency:)
 
   puts " => handled error whilst updating #{dependency.name}: #{error_details.fetch(:'error-type')} "\
        "#{error_details.fetch(:'error-detail')}"
+  puts error.backtrace if $options[:trace]
 end
 
 StackProf.start(raw: true) if $options[:profile]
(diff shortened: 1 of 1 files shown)


bin/dry-run.rb
```diff
@@ -447,15 +454,19 @@ def handle_dependabot_error(error:, dependency:)
 always_clone = Dependabot::Utils.
                always_clone_for_package_manager?($package_manager)
 if $options[:clone] || always_clone
-  $repo_contents_path = Dir.mktmpdir
-  puts "=> cloning into #{$repo_contents_path}"
+  $repo_contents_path = File.expand_path(File.join("tmp", $repo_name.split("/")))
```
thepwagner, line 457: Concern: since the `FileFetcher` [exits](https://github.com/dependabot/dependabot-core/blob/e7c1929d7f850f8516311cc41753958cb6b980d2/common/lib/dependabot/file_fetchers/base.rb#L476) if the path exists, this flow won't:
* fetch the latest from the repository
* update if `--branch` or `--commit` flags are provided

Is that expected?

bin/dry-run.rb
```diff
@@ -447,15 +454,19 @@ def handle_dependabot_error(error:, dependency:)
...
 end
 
 fetcher = Dependabot::FileFetchers.for_package_manager($package_manager).
           new(source: $source, credentials: $options[:credentials],
               repo_contents_path: $repo_contents_path)
 $files = if $options[:clone] || always_clone
-           fetcher.clone_repo_contents
+           unless $options[:cache_steps].include?("files") && Dir.exist?(File.join($repo_contents_path, ".git"))
```
thepwagner, line 464: Nit: it's a weird smell that this duplicates the check that `clone_repo_contents` does internally https://github.com/dependabot/dependabot-core/blob/e7c1929d7f850f8516311cc41753958cb6b980d2/common/lib/dependabot/file_fetchers/base.rb#L476

Maybe `clone_repo_contents` should return a status?

thepwagner (approved):
LGTM for the reuse!

I'm worried the `FileFetcher` should be able to switch references/branches, or we might end up with subtle bugs where we're not running against the expected code.

I'd dig if the `FileFetcher` would `git fetch` when the path exists: so _most_ blobs are reused, but we still know the repo is up to date. Would it need to `git checkout` too?
I only see  https://github.com/dependabot/dependabot-core/blob/49d179ef1ebbfd17adad7ab360a48eca41467f66/common/lib/dependabot/shared_helpers.rb#L246-L247

bin/dry-run.rb
feelepxyz: Oh good spot, forgot we did that. Will have a think about improving the cloning behaviour, we currently always ignore the `--commit` option for example.

PR #3389, opened 2021-03-29
Dockerfile: promote software-properties-common

thepwagner:
This moves `software-properties-common`  to earlier in the `Dockerfile`.

This package provides the `apt-add-repository` used by setup PPAs used by several ecosystems: `ruby`, `php`. Since the package is part of baseline ubuntu, we can install it in the first layer to avoid a duplicate `apt-get update`.

I opted to place it before `make`, so the preceding "things after `make` are for python" comment was still valid.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,6 +29,7 @@ RUN apt-get update \
     unzip \
     locales \
     openssh-client \
+    software-properties-common \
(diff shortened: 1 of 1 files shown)


feelepxyz (approved):
🙌

PR #3398, opened 2021-03-29
Dockerfile: create "dependabot" user

thepwagner:
Previously, the root `Dockerfile` would run all instructions as the default user of the parent container (specifically: `root`). This leaves several paths where dynamic SDK versions might be installed (e.g. `pyenv`, `rustup`) owned by root.
In `Dockerfile.development`, and hopefully other downstream containers, we [create a user grant that user ownership of these paths](https://github.com/dependabot/dependabot-core/blob/a7fd3a37f963b41bc5477786558a6609c832a977/Dockerfile.development#L19-L24). Those ownership changes make for a very expensive docker layer.

This PR migrates the creation of the user account to the base `Dockerfile`, so that dependencies can be owned by the correct user in the layer they're initially added. This will reduce build times, and the size of final images.

The ability for `bin/docker-dev-shell --rebuild` to synchronize the `dependabot` UID/GID for consistency across volume mounts is retained, but now developers are expected to rebuild the `dependabot/dependabot-core` image locally (previously it could be pulled from the registry). The `--rebuild` flag already does this, so it shouldn't be any additional toil 🤞 .

~I introduced a `bin/cibuild` script that chains the `Dockerfile`->`Dockerfile.ci` steps similar to Actions. This was helpful for this iteration - it could be adopted by `.github/workflows/ci.yml` or abandoned. I don't think we should start maintaining this flow in two places.~ We rolled that back for now.

Using that harness, I verified the CI script can be simplified by just bulk adding the source code in the `Dockerfile.ci` build. This avoids the scripting complexity (and for Macs: hefty performance penalty) of volume mounting all the individual directories. That script complexity is retained in `bin/docker-dev-shell`, because sharing with an external editor is worth it there. This avoids synchronizing the UID/GID with the host in CI, so the UID/GID of `dependabot/dependabot-core` is a reliable `1000:1000`.


Dockerfile.development
```diff
@@ -4,42 +4,33 @@ FROM dependabot/dependabot-core
...
 ENV PATH="$BUNDLE_BIN:$PATH:$BUNDLE_PATH/bin"
 
-COPY common/Gemfile common/dependabot-common.gemspec ${CODE_DIR}/common/
-COPY common/lib/dependabot/version.rb ${CODE_DIR}/common/lib/dependabot/
-RUN cd common && bundle install
+COPY --chown=dependabot:dependabot common/Gemfile common/dependabot-common.gemspec ${CODE_DIR}/common/
+COPY --chown=dependabot:dependabot common/lib/dependabot/version.rb ${CODE_DIR}/common/lib/dependabot/
+
+WORKDIR ${CODE_DIR}
+RUN chown -R dependabot:dependabot $CODE_DIR
```
feelepxyz, line 31: Do we need to chown the code dir? Could we `mkdir` before copying as the `dependabot` user?

Dockerfile
```diff
@@ -173,8 +178,12 @@ RUN wget https://packages.erlang-solutions.com/erlang-solutions_1.0_all.deb \
 # Install Rust 1.51.0
 ENV RUSTUP_HOME=/opt/rust \
   PATH="${PATH}:/opt/rust/bin"
+RUN mkdir -p "$RUSTUP_HOME" && \
+  chown -R dependabot:dependabot "$RUSTUP_HOME"
```
feelepxyz, line 182: Do we need this `chown` if we move up the `dependabot` user to before creating the dir?

feelepxyz:
Thanks for starting on this! I reckon this might end up speeding up updater image builds a fair bit and reduce the size 🙌 also means we can test core with the same permissions as the updater without running out of disk space in actions ✨

Dockerfile
```diff
@@ -50,6 +50,13 @@ RUN apt-get update \
     python3-enchant \
   && locale-gen en_US.UTF-8
 
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+ARG OSTYPE=Linux
+
+RUN bash -c "[ \"$OSTYPE\" = \"Linux\" ] && groupadd --gid \"${USER_GID}\" dependabot" \
```
thepwagner, line 57: Question: why do we only want to create the group if `OSTYPE = Linux`?

When `OSTYPE != Linux`, how does the `chown` on L59 work? We haven't created the `dependabot` group...

Dockerfile
```diff
@@ -173,25 +185,29 @@ RUN wget https://packages.erlang-solutions.com/erlang-solutions_1.0_all.deb \
 # Install Rust 1.51.0
 ENV RUSTUP_HOME=/opt/rust \
   PATH="${PATH}:/opt/rust/bin"
+USER dependabot
+RUN mkdir -p "$RUSTUP_HOME"
```
thepwagner, line 189: Noice: since `/opt` is already owned by `dependabot`, doing the `USER` instruction first means we can skip the `chown` part of the `mkdir && chown` from other ecosystems.

Dockerfile.ci
```diff
@@ -1,10 +1,35 @@
 FROM dependabot/dependabot-core
 
+ARG CODE_DIR=/home/dependabot/dependabot-core
+
+USER dependabot
+
```
thepwagner, line 6:
```suggestion
```

Nit: since L209 in `Dockerfile`, this doesn't do anything.

Dockerfile
```diff
@@ -68,11 +75,14 @@ RUN apt-add-repository ppa:brightbox/ruby-ng \
 # Install Python 2.7 and 3.9 with pyenv. Using pyenv lets us support multiple Pythons
 ENV PYENV_ROOT=/usr/local/.pyenv \
   PATH="/usr/local/.pyenv/bin:$PATH"
+RUN mkdir -p /usr/local/.pyenv && chown dependabot:dependabot /usr/local/.pyenv
```
thepwagner, line 78:
```suggestion
RUN mkdir -p "$PYENV_ROOT" && chown dependabot:dependabot "$PYENV_ROOT"
```

Nit: leverage existing variable.

thepwagner (commented):
LGTM! I wondered about implications of `OSTYPE`

If it's necessary, should the later chowns use `dependabot:$USER_GID` instead of the `dependabot:dependabot` (e.g. the group by name)?

Dockerfile
feelepxyz: This was my attempt at fixing an issue on mac where the `USER_GID` already existed so the `groupadd` failed, but you're right, the chown should fail as the group is surely called something other than `dependabot`.

Could we do something like `chown ${USER_UID}:${USER_GID}` instead? There might also be a better way to deal with this, not sure why I've never run into it before with the `docker-dev-shell`

thepwagner: > Could we do something like chown ${USER_UID}:${USER_GID} instead?

SGTM.

Alternatively, do we need to align `USER_GID` with the host? As long as `USER_UID` matches the permissions will work out...

Dockerfile
```diff
@@ -50,6 +50,14 @@ RUN apt-get update \
     python3-enchant \
   && locale-gen en_US.UTF-8
 
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+
+RUN GROUP_NAME=$(getent group $USER_GID | awk -F':' '{print $1}') \
+  && if [ -z $GROUP_NAME ]; then groupadd --gid $USER_GID dependabot ; \
+     else groupmod -n dependabot $GROUP_NAME ; fi \
```
feelepxyz, line 58: @thepwagner updated it to rename the group if it exists, seems to work in `docker-dev-shell`

Dockerfile
```diff
@@ -173,25 +186,29 @@ RUN wget https://packages.erlang-solutions.com/erlang-solutions_1.0_all.deb \
...
+COPY --chown=dependabot:dependabot hex/helpers /opt/hex/helpers
+COPY --chown=dependabot:dependabot npm_and_yarn/helpers /opt/npm_and_yarn/helpers
+COPY --chown=dependabot:dependabot python/helpers /opt/python/helpers
+COPY --chown=dependabot:dependabot terraform/helpers /opt/terraform/helpers
 
 ENV DEPENDABOT_NATIVE_HELPERS_PATH="/opt" \
   PATH="$PATH:/opt/terraform/bin:/opt/python/bin:/opt/go_modules/bin:/opt/dep/bin" \
   MIX_HOME="/opt/hex/mix"
 
+USER dependabot
```
thepwagner, line 210: With deeper reflection, I wondered about the implications of this bit.

Previously, helpers would be installed and owned by the `root` user. If a downstream consumer of the image were to add a `dependabot` user to drop privileges, that user would not have write access to this code.

Now we've removed the burden of adding a `dependabot` user from the downstream consumers, but may have unintentionally made these helpers writable!

If we take [Dockerfile.development](https://github.com/dependabot/dependabot-core/blob/ac276c959e8fa193061e47420152b7e4fa178b0a/Dockerfile.development#L21-L24) as indicative of what permissions the delegated user actually needs:
* `/usr/local/.pyenv` - we address this earlier in the Dockerfile
* `/opt/rust` - we address this earlier in the Dockerfile
* `/opt/go/gopath` - this requires consideration


These make sense: python and rust may install toolchains to these locations, and Go may fetch sources to `${GOPATH}/pkg`.

---

I'm wondering if we should return to installing helpers as the `root` user, then append a `chown /opt/go/gopath` to fixup the problematic paths.

I'm not 100% sold it's necessary, just flagging the change.

Dockerfile.development
```diff
@@ -4,46 +4,27 @@ FROM dependabot/dependabot-core
...
-
-ENV BUNDLE_PATH="/home/$USERNAME/.bundle" \
+ENV BUNDLE_PATH="/home/dependabot/.bundle" \
   BUNDLE_BIN=".bundle/binstubs"
 ENV PATH="$BUNDLE_BIN:$PATH:$BUNDLE_PATH/bin"
 
-COPY common/Gemfile common/dependabot-common.gemspec ${CODE_DIR}/common/
-COPY common/lib/dependabot/version.rb ${CODE_DIR}/common/lib/dependabot/
-RUN cd common && bundle install
+WORKDIR ${CODE_DIR}
```
thepwagner, line 22:
```suggestion
```

Concern: this is a problem if `DOCKER_BUILDKIT=0`, and I'm not sure we should require that...

I had to check, but before BuildKit the `WORKDIR` command ignores the current `USER`:
```
$ cat Dockerfile
FROM debian:buster
WORKDIR /tmp/a-workdir-pre
USER nobody
WORKDIR /tmp/a-workdir-post
RUN touch /tmp/a-file

$ DOCKER_BUILDKIT=0 docker build -t foo . && docker run --rm foo ls -l /tmp
total 8
-rw-r--r-- 1 nobody nogroup    0 Apr  7 17:01 a-file
drwxr-xr-x 2 root   root    4096 Apr  7 17:01 a-workdir-post
drwxr-xr-x 2 root   root    4096 Apr  7 17:01 a-workdir-pre

$ DOCKER_BUILDKIT=1 docker build -t foo . && docker run --rm foo ls -l /tmp
total 8
-rw-r--r-- 1 nobody nogroup    0 Apr  7 17:01 a-file
drwxr-xr-x 2 nobody nogroup 4096 Apr  7 17:01 a-workdir-post
drwxr-xr-x 2 root   root    4096 Apr  7 17:01 a-workdir-pre
```

This manifests as L26 failing: since `$CODE_DIR` is owned by `root:root` the `mkdir` lacks permission to write it.

We could:
* Move it after the `mkdir` block
* Prefer absolute paths (e.g. `cd "${CODE_DIR}/common && bundle install`)

Dockerfile
xlgmokha: One small observation. I tried running strace on a process to see why it was taking so long to finish which I was working on a bug. strace requires root privileges and without sudo and /etc/sudoers.d this made strace unusable AFAIK.

Dockerfile
```diff
@@ -60,19 +68,22 @@ RUN apt-add-repository ppa:brightbox/ruby-ng \
...
 # Install Python 2.7 and 3.9 with pyenv. Using pyenv lets us support multiple Pythons
 ENV PYENV_ROOT=/usr/local/.pyenv \
   PATH="/usr/local/.pyenv/bin:$PATH"
+RUN mkdir -p "$PYENV_ROOT" && chown dependabot:dependabot "$PYENV_ROOT"
+USER dependabot
 RUN git clone https://github.com/pyenv/pyenv.git /usr/local/.pyenv \
   && cd /usr/local/.pyenv && git checkout 1.2.26 && cd - \
   && pyenv install 3.9.4 \
   && pyenv install 2.7.18 \
   && pyenv global 3.9.4
```
xlgmokha, line 85: I have come to greatly appreciate [asdf](https://asdf-vm.com/#/).

If we create a `.tool-versions` file with the following:

```plaintext
python 3.9.4 2.7.18
ruby 2.7.2
golang 1.16.2
rust 1.50.0
```

Then we can install multiple versions of tools via `asdf install` assuming we've installed the appropriate plugins. Writing plugins is very easy as well so we could swap out the default plugins for our own custom versions that are optimized for this runtime environment.

Some of the plugins also support `.default-<packages` configs to install packages for each version of that tool.

E.g.

```bash
モ cat ~/.default-gems
bundler
interactive_editor
jive
neovim
solargraph
spandx

モ cat ~/.default-python-packages
awscli
docker-compose
lookatme
pass-git-helper
pip
pipenv
turses
twine
wheel
```

xlgmokha (approved):
Is there any chance you prepare a [dive](https://github.com/wagoodman/dive) report or hook up `dive` in the `cibuild` script via [this](https://github.com/wagoodman/dive#ci-integration).

Dockerfile
feelepxyz: Nice! I haven't had a chance to play with that yet but looks promising! Seems like a good follow up to investigate 🙌  do you know if it can install things from binaries or builds from source?

Dockerfile
thepwagner: > I tried running strace on a process 

How did that work before? This PR doesn't change the dev shell dropping to `dependabot` [previous](https://github.com/dependabot/dependabot-core/blob/3656c3de267921c038e6d23c3055941f071e87a8/Dockerfile.development#L25)

AFAICT this capability was introduced [2018-11-21](https://github.com/dependabot/dependabot-core/commit/c7e9a9cd8715157cd86ebc7a5963ef13330dddd1) and has been broken since [2018-27-28](https://github.com/dependabot/dependabot-core/commit/63fe436b9706a63be6ef32e0cef392510028a40a)?
If this is an issue, could we address it in a separate PR?

thepwagner:
> a dive report
> asdf

@xlgmokha I dig it! I've opened internal tracking issues for both of these improvements, I don't think they should slip into this PR's scope.

Dive reporting on `dependabot-core` would only half-solve the problem - we care most about the final image in production.

bin/cibuild
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -e
```
feelepxyz, line 1: Should we bring this out into a separate PR/change? I like the direction but might be a bit more work to get it to do all the caching things we do in the ci build.

Dockerfile.ci
```diff
@@ -7,4 +7,9 @@ RUN pyenv install -s 3.6.13
 RUN pyenv install -s 3.7.1
 RUN pyenv install -s 3.7.10
 
-RUN git config --global user.name dependabot-ci && git config --global user.email no-reply@github.com
+RUN git config --global user.name dependabot-ci && \
+  git config --global user.email no-reply@github.com
+
+ARG CODE_DIR=/home/dependabot/dependabot-core
+WORKDIR ${CODE_DIR}
+COPY --chown=dependabot:dependabot . ${CODE_DIR}
```
feelepxyz, line 15: Thoughts on splitting this up to more granular COPYs? This step takes over 2 mins on my machine and I guess would always run in full if any file changes?

feelepxyz: Tried this but ran into a `max depth exceeded` error after 104 layers, if we copy all folders inside each pm we get 117 😬

bin/cibuild
feelepxyz: Ah you just using this for testing for now? 👍 on keeping for that. I was modifying the docker-dev-shell but ended up committing those changes once

Dockerfile
feelepxyz: I attempted to undo this change and install helpers as root but was running into weird issues with composer failing during updates, so backpedalled to get it green again and test on staging

bin/cibuild
thepwagner: SGTM, let's remove it.

Dockerfile.ci
thepwagner: SGTM!

feelepxyz: Even after reducing the layers to 76 it looked like actions gave up and killed the process so went back to copying it all in one go https://github.com/dependabot/dependabot-core/runs/2297918989

Dockerfile
feelepxyz: I think this permission change is the only thing left to figure out, I ran into loads of trouble trying to untangle this without ending up with the same `chown` we set out to remove from the updater image. As some of these helpers need to have write access to the helpers anyways, e.g. `bundler`, I'm wondering if it's worth trying to restrict the rest?

omnibus/script/ci-test
```diff
@@ -2,4 +2,3 @@
 
 bundle install
 bundle exec rubocop .
-bundle exec rspec spec
```
feelepxyz, line 5: There are no omnibus specs

bin/cibuild
feelepxyz: Actually finding this pretty useful for dev testing

Dockerfile
feelepxyz: I think I got this back to where previously so the native helper scripts should not be writeable by dependabot but the `gopath` and `.bundle` folders will be without the big `chown`. This setup feels a bit gnarly but tests should fail hard if this breaks so not too worried about it accidentally breaking.

.github/workflows/ci.yml
```diff
@@ -85,126 +85,8 @@ jobs:
...
-            -v "$(pwd)/bundler/spec:$CODE_DIR/bundler/spec" \
-            -v "$(pwd)/bundler/helpers:/opt/bundler" \
-            -v "$(pwd)/bundler/helpers:/opt/bundler/helpers" \
-            -v "$(pwd)/bundler/script:$CODE_DIR/bundler/script" \
-            -v "$(pwd)/omnibus/.rubocop.yml:$CODE_DIR/omnibus/.rubocop.yml" \
-            -v "$(pwd)/omnibus/Gemfile:$CODE_DIR/omnibus/Gemfile" \
-            -v "$(pwd)/omnibus/dependabot-omnibus.gemspec:$CODE_DIR/omnibus/dependabot-omnibus.gemspec" \
-            -v "$(pwd)/omnibus/lib:$CODE_DIR/omnibus/lib" \
-            -v "$(pwd)/omnibus/spec:$CODE_DIR/omnibus/spec" \
-            -v "$(pwd)/omnibus/script:$CODE_DIR/omnibus/script" \
```
feelepxyz, line 203: 💥

feelepxyz:
@thepwagner I think this is close to done now 🎉  could I get your 👀  on the native helper permission changes? I'm going to do a test on staging with all ecosystems and leave it until Monday before thinking about deploying.

thepwagner:
>  could I get your 👀 on the native helper permission changes?

I'm concerned about the changes for rust permissions.

Referencing the [chown workaround](https://github.com/dependabot/dependabot-core/blob/fc12b554eaf835858e1cf9252da39bfc7243258f/Dockerfile.development#L22-L24) we're trying to replace, I think there are 3 important postconditions to test in the image:
* If the update requires custom python, can it be installed? `pyenv install 3.9.3`
* If the update downloads go modules, can they be installed? `go get github.com/sirupsen/logrus`
* If the update requires a custom rust toolchain, can it be installed? `rustup install 1.50.0`

b35c591 passes the first 2, but can't install a custom rust version:
```
dependabot@6f91d58b997c:/opt/go/gopath$ rustup install 1.50.0
info: syncing channel updates for '1.50.0-x86_64-unknown-linux-gnu'
error: could not create temp file: /opt/rust/tmp/sqi0qzbcwhefp9b3_file
```

I didn't work step by step, but the final result also writes to `/opt/rust/toolchains` and `/opt/rust/update-hashes`:
```
dependabot@ce69f0013b03:/opt/rust$ find /opt/rust -maxdepth 3 -name "*1.50.0*"
/opt/rust/toolchains/1.50.0-x86_64-unknown-linux-gnu
/opt/rust/update-hashes/1.50.0-x86_64-unknown-linux-gnu
```

We could either `chown` those paths so the delegated user can write them, or use a `USER dependabot` directive to perform the entire install as the `dependabot` user.

feelepxyz:
> [b35c591](https://github.com/dependabot/dependabot-core/commit/b35c591f0c02aa7e58b1c4a6a95fa81f674c8390) passes the first 2, but can't install a custom rust version:

Oh good spot! I'm trying to find any references of installing custom rust versions from updates but can't find any but looks like we fetch a `rust-toolchain` file if present. https://rust-lang.github.io/rustup/overrides.html

feelepxyz:
We might be able to test against this repo https://github.com/rust-lang/rust-clippy

Dockerfile
```diff
@@ -173,32 +189,38 @@ RUN wget https://packages.erlang-solutions.com/erlang-solutions_1.0_all.deb \
...
-  bash /opt/python/helpers/build /opt/python && \
-  bash /opt/dep/helpers/build /opt/dep && \
-  mkdir -p /opt/bundler/v1 && bash /opt/bundler/helpers/v1/build /opt/bundler/v1 && \
-  mkdir -p /opt/bundler/v2 && bash /opt/bundler/helpers/v2/build /opt/bundler/v2 && \
-  bash /opt/go_modules/helpers/build /opt/go_modules && \
-  bash /opt/npm_and_yarn/helpers/build /opt/npm_and_yarn && \
-  bash /opt/hex/helpers/build /opt/hex && \
-  bash /opt/composer/helpers/v2/build /opt/composer/v2 && \
-  bash /opt/composer/helpers/v1/build /opt/composer/v1
+USER dependabot
```
feelepxyz, line 209: @thepwagner thoughts on changing these helpers to be owned by dependabot making them writeable?

I've done a many loops on this trying to get it to work and run into upstream failures that I don't understand so went back to what was working previously.

feelepxyz: Wondering if it's worth the added mental overhead to button these down further, I've also spent countless hours trying to get it to work so seems brittle and easy to break even if we get it into shape.

feelepxyz:
Looks like rust updates with a custom toolchain are indeeed failing, tested on https://github.com/feelepxyz/rust-clippy and seeing: `info: syncing channel updates for 'nightly-2021-04-08-x86_64-unknown-linux-gnu'
updater | <job_109124542> error: could not create temp file: /opt/rust/tmp/ec6c4k1xq74aa07p_file`

thepwagner:
@xlgmokha now that we're ~stable, here's a (manual for now :/ ) `dive` report:

Pre, https://github.com/dependabot/dependabot-core/commit/fc12b554eaf835858e1cf9252da39bfc7243258f :
```
Image Source: docker://docker.io/dependabot/dependabot-core:fc12b554eaf835858e1cf9252da39bfc7243258f
Fetching image... (this can take a while for large images)
Analyzing image...
  efficiency: 98.7559 %
  wastedBytes: 93762386 bytes (94 MB)
  userWastedPercent: 2.1754 %
Inefficient Files:
Count  Wasted Space  File Path
    2        8.2 MB  /var/lib/apt/lists/archive.ubuntu.com_ubuntu_dists_bionic-updates_main_binary-amd64_Packages.lz4
    2        7.2 MB  /var/lib/apt/lists/archive.ubuntu.com_ubuntu_dists_bionic-updates_universe_binary-amd64_Packages.lz4
    2        6.8 MB  /var/lib/apt/lists/security.ubuntu.com_ubuntu_dists_bionic-security_main_binary-amd64_Packages.lz4
    6        5.6 MB  /var/cache/debconf/templates.dat
    2        4.6 MB  /var/lib/apt/lists/security.ubuntu.com_ubuntu_dists_bionic-security_universe_binary-amd64_Packages.lz4
    3        3.1 MB  /var/cache/debconf/templates.dat-old
    2        3.1 MB  /root/.npm/_cacache/content-v2/sha512/b7/61/15994ac668f42f5e7d0591bcbb5b6967f5b678f0a64d5a5144a4731786c17a0bee115f6fff2a93520c7a395b09347f7812145c1e0336051609a6c3a86716
    6        2.2 MB  /var/lib/dpkg/status
    6        2.2 MB  /var/log/dpkg.log
    5        2.2 MB  /var/lib/dpkg/status-old
    5        1.2 MB  /var/lib/apt/lists/archive.ubuntu.com_ubuntu_dists_bionic_InRelease
```

Post, https://github.com/dependabot/dependabot-core/pull/3398/commits/86c2b45012053f4e8f23ee53c38f37474c3916e2
```
Image Source: docker://dependabot/dependabot-core:86c2b45012053f4e8f23ee53c38f37474c3916e2
Fetching image... (this can take a while for large images)
Analyzing image...
  efficiency: 98.6085 %
  wastedBytes: 98797996 bytes (99 MB)
  userWastedPercent: 2.2888 %
Inefficient Files:
Count  Wasted Space  File Path
    3         12 MB  /var/lib/apt/lists/archive.ubuntu.com_ubuntu_dists_bionic-updates_main_binary-amd64_Packages.lz4
    3         11 MB  /var/lib/apt/lists/archive.ubuntu.com_ubuntu_dists_bionic-updates_universe_binary-amd64_Packages.lz4
    2        6.8 MB  /var/lib/apt/lists/security.ubuntu.com_ubuntu_dists_bionic-security_main_binary-amd64_Packages.lz4
    6        5.6 MB  /var/cache/debconf/templates.dat
    2        4.6 MB  /var/lib/apt/lists/security.ubuntu.com_ubuntu_dists_bionic-security_universe_binary-amd64_Packages.lz4
    3        3.1 MB  /var/cache/debconf/templates.dat-old
    6        2.2 MB  /var/lib/dpkg/status
    6        2.2 MB  /var/log/dpkg.log
    5        2.2 MB  /var/lib/dpkg/status-old
    5        1.2 MB  /var/lib/apt/lists/archive.ubuntu.com_ubuntu_dists_bionic_InRelease
    2        590 kB  /usr/local/.pyenv/versions/2.7.18/lib/python2.7/site-packages/pip/_vendor/idna/uts46data.pyc
```

Pretty boring: no regressions - still some wins to be squeaked out with `&& rm  /var/lib/apt/lists/*` but that wasn't the intent.

---

The expected wins are in images derived from `dependabot/dependabot-core`, e.g. the `Dockerfile.development` from inside this repo:

Pre:
```
Image Source: docker://dependabot/dependabot-core-development:fc12b554eaf835858e1cf9252da39bfc7243258f
Fetching image... (this can take a while for large images)
Analyzing image...
  efficiency: 65.1342 %
  wastedBytes: 4670635698 bytes (4.7 GB)
  userWastedPercent: 67.5946 %
Inefficient Files:
Count  Wasted Space  File Path
    2        251 MB  /opt/rust/toolchains/1.51.0-x86_64-unknown-linux-gnu/lib/librustc_driver-7ea116e55de24565.so
    2        251 MB  /opt/rust/toolchains/stable-x86_64-unknown-linux-gnu/lib/librustc_driver-7ea116e55de24565.so
    2        186 MB  /opt/rust/toolchains/stable-x86_64-unknown-linux-gnu/lib/libLLVM-11-rust-1.51.0-stable.so
    2        186 MB  /opt/rust/toolchains/1.51.0-x86_64-unknown-linux-gnu/lib/libLLVM-11-rust-1.51.0-stable.so
    2        139 MB  /opt/rust/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/libcore-3aaf8f932781f33e.rlib
    2        139 MB  /opt/rust/toolchains/1.51.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/libcore-3aaf8f932781f33e.rlib
    2         72 MB  /usr/local/.pyenv/versions/3.9.4/lib/python3.9/config-3.9-x86_64-linux-gnu/libpython3.9.a
    2         72 MB  /usr/local/.pyenv/versions/3.9.4/lib/libpython3.9.a
    3         48 MB  /home/dependabot/.bundle/cache/compact_index/rubygems.org.443.29b0360b937aa4d161703e6160654e47/versions
    2         44 MB  /opt/rust/toolchains/1.51.0-x86_64-unknown-linux-gnu/bin/cargo
    2         44 MB  /opt/rust/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo
```

Post:
```
Image Source: docker://docker.io/dependabot/dependabot-core-development:86c2b45012053f4e8f23ee53c38f37474c3916e2
Fetching image... (this can take a while for large images)
Analyzing image...
  efficiency: 97.5930 %
  wastedBytes: 165345180 bytes (165 MB)
  userWastedPercent: 3.6150 %
Inefficient Files:
Count  Wasted Space  File Path
    4         64 MB  /home/dependabot/.bundle/cache/compact_index/rubygems.org.443.29b0360b937aa4d161703e6160654e47/versions
    3         12 MB  /var/lib/apt/lists/archive.ubuntu.com_ubuntu_dists_bionic-updates_main_binary-amd64_Packages.lz4
    3         11 MB  /var/lib/apt/lists/archive.ubuntu.com_ubuntu_dists_bionic-updates_universe_binary-amd64_Packages.lz4
    2        6.8 MB  /var/lib/apt/lists/security.ubuntu.com_ubuntu_dists_bionic-security_main_binary-amd64_Packages.lz4
    7        6.6 MB  /var/cache/debconf/templates.dat
    2        4.6 MB  /var/lib/apt/lists/security.ubuntu.com_ubuntu_dists_bionic-security_universe_binary-amd64_Packages.lz4
    3        3.1 MB  /var/cache/debconf/templates.dat-old
    7        2.8 MB  /var/lib/dpkg/status
    7        2.7 MB  /var/log/dpkg.log
    6        2.7 MB  /var/lib/dpkg/status-old
    5        1.2 MB  /var/lib/apt/lists/archive.ubuntu.com_ubuntu_dists_bionic_InRelease
```

That's exactly the expected result: we've saved [this super expensive layer](https://github.com/dependabot/dependabot-core/blob/fc12b554eaf835858e1cf9252da39bfc7243258f/Dockerfile.development#L21-L24) 🏎️ 🎉 .

feelepxyz:
Sweet!

feelepxyz:
Rust updates with custom toolchains are working now 🎉  https://github.com/feelepxyz/rust-clippy/pull/1

feelepxyz (approved):
I've tested all ecosystems on staging are working, including installing custom rust toolchains and ruby git vendoring 🎉 

Screenshot 2021-04-12 at 12 29 58

thepwagner:
> feelepxyz requested a review from hmarr 3 hours ago

Per this, I'm considering this blocked in @hmarr 's review.

hmarr (approved):
@feelepxyz and I chatted about this separately, I'm 👍  on the change.

PR #3401, opened 2021-03-29
📝 Improve guidance for opening actionable bug reports

jasonrudolph:
When investigating a bug report, there are a handful of details that we often end up manually gathering in order to diagnose the problem. This pull request updates the template to prompt the issue author to provide these details, which gives us a better chance of being able to respond sooner.


feelepxyz (approved):
Nice one! 🙇 ✨

PR #3403, opened 2021-03-29
Added option to use Gitlab access token for authentication against maven repositories

gringostar:
An option to use custom header for maven repositories is added, to allow for alternative authentication on Gitlab maven repos.

Motivation: When using GitLabs maven repository, a token header must be provided as authentication as one of the following headers:
- Private-Token
- Job-Token
- Deploy-Token
as specified [here](https://docs.gitlab.com/ee/user/packages/maven_repository/#authenticate-to-the-package-registry-with-maven).

The change is implemented by adding the access token provided for a git_source in the 'Private-token' header, if no username/password is provided for the maven_repository, and the host of the git_source and maven_repository is the same.


jurre:
Thanks for the contribution @gringostar!

I wonder if we could alternatively detect that we're dealing with a GitLab maven repo, and pass in the `token` from the configuration in the right header.

The reason I'm suggesting that, is that I would prefer to keep the configuration options for Dependabot as simple as possible, ideally users would have to know as little as possible about authenticating against different kinds of maven repo's. WDYT?

gringostar:
> Thanks for the contribution @gringostar!
> 
> I wonder if we could alternatively detect that we're dealing with a GitLab maven repo, and pass in the `token` from the configuration in the right header.
> 
> The reason I'm suggesting that, is that I would prefer to keep the configuration options for Dependabot as simple as possible, ideally users would have to know as little as possible about authenticating against different kinds of maven repo's. WDYT?

I considered going with that solution as well, but in our own case, we are dealing with a privately hosted gitlab instance, and I am not sure how to detect the repo is a Gitlab maven repo. Nothing in the hostname reveals that this is a GitLab server. It is of course an option to be more Gitlab specific, by defining a set of maven_gitlab_repository credentials, that includes an tokentype/name and the token, perhaps even limiting the type option to the 3 defined by Gitlab. I can give that a shot, if that is preferable, over the generic header solution.

gringostar:
The basic issue I faced, was being able to specify other headers, than what was already defined, when looking up dependencies on private repositories. I am not aware if there is a similar issue, when using gitlab hosted repos. The type of token has to correspond to the type of token used, otherwise gitlab will throw a 401.

gringostar:
@jurre I changed the implementation to look for a git_source with a matching host, and use the provided password as 'Private-token' header, if a username/password is not defined for the maven_repository. I was unsure how to look up defined sources, and check the provider is gitlab, which would probably make sense.

gringostar:
Hi @jurre. I have made a change to only apply the gitlab header, if the maven repository path contains the path `/api/v4/.../packages/maven` which is the current structure of gitlab maven repositories. This should ensure it is not applied to other repos. I have extracted some repeated logic into separate classes, one for Gradle and one for Maven. If you have a suggestion on how to refactor this into a single class, let me know. I will add tests for the two classes, but for now they should be tested by the existing tests.

maven/lib/dependabot/maven/utils/auth_details_finder.rb
```diff
@@ -0,0 +1,58 @@
...
+              c.fetch("password", nil)
+            end
+
+          return {} unless cred
+
+          { "Private-Token" => cred.fetch("password") }
+        end
+
+        def gitlab_maven_repo?(maven_repo_path)
+          gitlab_maven_repo_reg = %r{^(/api/v4).*(/packages/maven)/?$}.freeze
```
xlgmokha, line 51: It doesn't look like we use the capture groups. Do this need them?

Should `.*` become `.+/-`. I think this is meant to capture the group hierarchy in gitlab projects. E.g `org/group/subgroup/project`. Is that right?

xlgmokha (commented):
Thanks for working on this @gringostar  I left a few comments to consider. I hope that they are helpful and that I'm not nitpicking too much.

My two main blockers are:

1. Defaulting to gitlab for all http authentication schemes except for basic auth.
2. Usage of the non standard `Authorization: Private-Token` HTTP authentication scheme. I would be okay with supporting the schemes registered with [IANA](https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml) like `Basic`, `Bearer`, `OAuth`.
  
WDYT @jurre?

gringostar:
Thank you for your feedback @xlgmokha. To clarify a few things:
- The PR aims to enable scanning for possible version updates on gitlabs maven repositories, either hosted by gitlab or privately hosted. Gitlab uses a [custom header for authentication](https://docs.gitlab.com/ee/user/packages/maven_repository/#authenticate-to-the-package-registry-with-maven), rather than any standardized authentication scheme. The changes only concerns maven repositories.
- The intention is to only set this custom header, if:
1. The path to the maven repository matches the current pattern known for gitlab maven repositories. There might be a need to add additional, if other cases are known. 
2. There is a set of credentials for a git_source with a matching host, could be gitlab.com or my-hosted-gitlab.private.com
- For all other cases the logic is left as is.

> Thanks for working on this @gringostar I left a few comments to consider. I hope that they are helpful and that I'm not nitpicking too much.
> 
> My two main blockers are:
> 
> 1. Defaulting to gitlab for all http authentication schemes except for basic auth.

The solution should not default to gitlab, only in case we know this is a gitlab maven repository. 
The priority will be to:  
1. use basic auth (as is), if a username/password was provided for the maven_repository 
2. if no username/password is provided for the maven_repository, and above mentioned criteria is met, set a header named 'Private-token' with the value provided as access-token for the git_source
3. provide no headers for authentication (as is)

> 2. Usage of the non standard `Authorization: Private-Token` HTTP authentication scheme. I would be okay with 
supporting the schemes registered with [IANA](https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml) like `Basic`, `Bearer`, `OAuth`.

The header is not an 'Authorization' header, but a custom. Had they only used that :). As mentioned this is only to enable dependency updates specifically from gitlab maven repositories. 

> 
> WDYT @jurre?

gradle/lib/dependabot/gradle/update_checker/version_finder.rb
```diff
@@ -184,10 +185,8 @@ def dependency_metadata(repository_details)
             begin
               response = Excon.get(
                 dependency_metadata_url(repository_details.fetch("url")),
-                user: repository_details.fetch("username"),
-                password: repository_details.fetch("password"),
                 idempotent: true,
-                **SharedHelpers.excon_defaults
+                **Dependabot::SharedHelpers.excon_defaults(headers: repository_details.fetch("auth_details"))
```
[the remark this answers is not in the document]
gringostar: The Dependabot::SharedHelpers sets up default header 'User-agent', and merges the headers given as argument in the excon_defaults method. In case of an empty array (if no auth details were found) none will be added, but there will still be headers on the request made.

gradle/lib/dependabot/gradle/utils/auth_details_finder.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+module Dependabot
+  module Gradle
+    module Utils
+      class AuthDetailsFinder
```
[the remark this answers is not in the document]
gringostar: The responsibility should be to find auth details of maven repositories. Perhaps the the detail that it is specifically for maven repositories should be included in the name. I would like to have used the auth_details_finder located in the maven module, as this class is dedicated to finding auth details for maven repositories when using Gradle. But I would assume that the auth_details_finder class should be placed in the common module for that to be OK. And being very specific for Maven repositories, I left it in the two modules for now.

gradle/lib/dependabot/gradle/update_checker/version_finder.rb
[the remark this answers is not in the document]
gringostar: Perhaps `auth_headers` would be a better term.

gradle/lib/dependabot/gradle/utils/auth_details_finder.rb
```diff
@@ -0,0 +1,58 @@
...
+        def initialize(credentials)
+          @credentials = credentials
+        end
+
+        def auth_details(maven_repo_url)
+          cred =
+            credentials.select { |c| c["type"] == "maven_repository" }.
+            find do |c|
+              cred_url = c.fetch("url").gsub(%r{/+$}, "")
+              next false unless cred_url == maven_repo_url
```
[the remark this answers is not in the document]
gringostar: This logic have been taken from metadata_finder.rb 1:1. This could probably be written cleaner.

gradle/lib/dependabot/gradle/utils/auth_details_finder.rb
```diff
@@ -0,0 +1,58 @@
...
+
+              c.fetch("username", nil)
+            end
+
+          return gitlab_auth_details(maven_repo_url) unless cred
```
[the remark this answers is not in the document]
gringostar: The code might be unclear, the first thing done in the gitlab_auth_details method is to check if the maven repository path matches a gitlab maven repository, otherwise `{}` is returned.

My first approach actually was to add an option to provide custom headers when defining maven_repositories, but a change would still be needed to allow the header being added to the request, as the current solution defaults to basic auth.

gradle/lib/dependabot/gradle/utils/auth_details_finder.rb
```diff
@@ -0,0 +1,58 @@
...
+
+          token = cred.fetch("username") + ":" + cred.fetch("password")
+          encoded_token = Base64.encode64(token).delete("\n")
+          { "Authorization" => "Basic #{encoded_token}" }
```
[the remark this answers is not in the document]
gringostar: The User-agent header is being set up in the Dependabot::SharedHelpers.excon_defaults. Here we only deal with auth headers, which will subsequently be merged these headers.

gradle/lib/dependabot/gradle/utils/auth_details_finder.rb
```diff
@@ -0,0 +1,58 @@
...
+            find do |c|
+              cred_host = c.fetch("host").gsub(%r{/+$}, "")
+              next false unless URI(maven_repo_url).host == cred_host
+
+              c.fetch("password", nil)
+            end
+
+          return {} unless cred
+
+          { "Private-Token" => cred.fetch("password") }
```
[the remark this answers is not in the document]
gringostar: If only Gitlab would use an IANA recognized scheme, but for some reason they use custom headers, as described [here](https://docs.gitlab.com/ee/user/packages/maven_repository/#authenticate-with-a-ci-job-token-in-gradle).

gradle/spec/dependabot/gradle/update_checker/version_finder_spec.rb
```diff
@@ -213,6 +213,41 @@
         is_expected.to eq("https://private.registry.org/repo")
       end
 
+      context "that is a gitlab maven repository" do
+        let(:credentials) do
+          [
+            {
+              "type" => "maven_repository",
+              "url" => "https://private.registry.org/api/v4/groups/-/packages/maven/"
```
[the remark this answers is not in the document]
gringostar: Once a v5 is supported, it would make sense to extend to support this, given that they do not change the authorization scheme (one could hope). I can see other parts of the code also defaults to v4, like [source.rb](https://github.com/dependabot/dependabot-core/blob/3656c3de267921c038e6d23c3055941f071e87a8/common/lib/dependabot/source.rb#L131), so this would probably be part of a larger update.

There is no guarantee that the gitlab maven repository is hosted on gitlab.com, it could be any host, which is why the path is being checked.

maven/lib/dependabot/maven/utils/auth_details_finder.rb
gringostar: You are right, I will remove them.

The intention was to focus on the start and the end of the path, and allow anything or nothing in between, as the path for the same repo can be on group level, project level or instance level: https://docs.gitlab.com/ee/user/packages/maven_repository/#use-the-gitlab-endpoint-for-maven-packages

gradle/lib/dependabot/gradle/update_checker/version_finder.rb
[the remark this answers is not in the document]
gringostar: I have made an update renaming auth_details auth_headers, I hope that better reflects what it is.

gradle/lib/dependabot/gradle/utils/auth_details_finder.rb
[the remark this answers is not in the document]
gringostar: The class have been renamed AuthHeadersFinder, let me know if you have a suggestion you find more fitting.

gradle/lib/dependabot/gradle/update_checker/version_finder.rb
```diff
@@ -333,6 +330,14 @@ def central_repo_urls
         def version_class
           Gradle::Version
         end
+
+        def auth_headers_finder
+          @auth_headers_finder ||= Utils::AuthHeadersFinder.new(credentials)
+        end
+
+        def auth_headers(maven_repo_url)
+          auth_headers_finder.auth_headers(maven_repo_url)
```
jurre, line 339: Can we memoize the result of the finder instead of the finder itself? I'm assuming it doesn't change for the lifespan of this object?

maven/lib/dependabot/maven/utils/auth_headers_finder.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+module Dependabot
+  module Maven
+    module Utils
+      class AuthHeadersFinder
```
jurre, line 6: This is currently the same between the Maven and Gradle implementation right? Wondering if it makes sense to hoist this up to the `common` package and reuse. Should we ever expect these implementations to deviate?

gradle/lib/dependabot/gradle/update_checker/version_finder.rb
gringostar: In this case, the list of maven_repositories will be iterated and mapped, so I assume (with my minimal Ruby experience) that the method will be called multiple times with varying maven_repo_urls, why it cannot be memoized. Let me know if I am wrong. In case of the metadata_finder.rb it should be possible.

maven/lib/dependabot/maven/utils/auth_headers_finder.rb
gringostar: That was my thought as well, but I was not sure if this was common enough, as it is very specific for maven repos. There is a natural dependency from Gradle to Maven, as Gradle uses Maven dependencies. Would it make sense to let the gradle module depend on the maven module?
Otherwise it should probably be moved to the common package. Implementation should not deviate, as long as it concerns maven repositories.

gradle/lib/dependabot/gradle/update_checker/version_finder.rb
gringostar: I was wrong 😑 The change have been made.

maven/lib/dependabot/maven/utils/auth_headers_finder.rb
gringostar: I have taken the liberty of adding maven as a dependency on the gradle module. I will move it to the common module if this is preferred. In that case suggestions for where it should be placed are welcome.

maven/lib/dependabot/maven/update_checker/version_finder.rb
```diff
@@ -287,6 +283,14 @@ def central_repo_urls
...
           %w(http:// https://).map { |p| p + central_url_without_protocol }
         end
+
+        def auth_headers_finder
+          @auth_headers_finder ||= Utils::AuthHeadersFinder.new(credentials)
+        end
+
+        def auth_headers(maven_repo_url)
+          auth_headers_finder.auth_headers(maven_repo_url)
+        end
```
jurre, line 293: We can do the same thing here right?

```suggestion
        def auth_headers
          @auth_headers ||= Utils::AuthHeadersFinder.new(credentials).auth_headers(maven_repo_url)
        end
```

jurre (approved):
Overall looks great, I left 1 inline suggestion

maven/lib/dependabot/maven/utils/auth_headers_finder.rb
```diff
@@ -0,0 +1,56 @@
...
+        def initialize(credentials)
+          @credentials = credentials
+        end
+
+        def auth_headers(maven_repo_url)
```
jurre, line 11: This feels like an absolute nit-pick, but since we don't expect to pass different `maven_repo_url`s to this method, it seems slightly more clear to pass the url in the constructor and calling `auth_headers` without arguments. WDYT?

maven/lib/dependabot/maven/update_checker/version_finder.rb
gringostar: It actually turns out it was the tests not being thorough enough, since it only tested with a single defined maven_repository at the time. With multiple repositories defined, several requests will be made, with different maven_repo_urls. I have added tests and reverted the chang on the gradle version_finder.rb.

maven/lib/dependabot/maven/utils/auth_headers_finder.rb
gringostar: As mentioned, there will be cases where multiple calls are made, with different URLs, tests did not catch this.

jurre: Ah I missed that 👍

maven/lib/dependabot/maven/update_checker/version_finder.rb
jurre: Ah yes that makes sense 👍

maven/lib/dependabot/maven/utils/auth_headers_finder.rb
jurre: I think adding maven as a dependency is fine 👍 . I wonder if there are more things we can reuse between the two, but let's leave that as an exercise for the future

gringostar:
Hi @jurre and @xlgmokha. Let me know if there is more I need to do, for this to be merged.

jurre (approved):

gradle/Gemfile
```diff
@@ -3,5 +3,6 @@
 source "https://rubygems.org"
 
 gem "dependabot-common", path: "../common"
+gem "dependabot-maven", path: "../maven"
```
xlgmokha, line 6: I think we may need to add this dependency to the `dependabot-gradle` gemspec.

E.g.
```ruby
spec.add_dependency "dependabot-maven", Dependabot::VERSION
```

gradle/spec/dependabot/gradle/update_checker/version_finder_spec.rb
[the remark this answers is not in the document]
JLDixon2022: It all looks good to me

Issue #3378, opened 2021-03-28
Remove `@` from PR description intro


shu-mutou:
Not to send lots of mentions, our community does not allow using `@` in commit message. But some packages has `@` in its name, so commit message in PR from dependabot includes `@` like following.

```
Bump @angular/platform-browser from 11.2.6 to 11.2.7

Bumps [@angular/platform-browser](https://github.com/angular/angular/tree/HEAD/packages/platform-browser) from 11.2.6 to 11.2.7.
- [Release notes](https://github.com/angular/angular/releases)
- [Changelog](https://github.com/angular/angular/blob/master/CHANGELOG.md)
- [Commits](https://github.com/angular/angular/commits/11.2.7/packages/platform-browser)

Signed-off-by: dependabot[bot] <support@github.com>
```

Can we remove `@` from commit message?

markcellus:
We have something similar. What setup are you using to prevent merging PRs with `@` in commit messages? A required Github status check?

I-Cat:
I can confirm this is an issue look at my code as i tested on my own repo

shu-mutou:
Our community checks `@` in commit message by own script via GitHub status check. They worry about lots of mentions not only `@user` but also `@org/team` due to commits that include `@`.

jurre:
We escape `@user` mentions pretty aggressively, you check out the code [here](https://github.com/dependabot/dependabot-core/blob/e7c1929d7f850f8516311cc41753958cb6b980d2/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb)

Did you see an instance where this caused a user to be mentioned? If so we'd love to know so we can fix it ASAP, it is incredibly annoying to be bombarded by dependabot mentions. As far as I can tell `@angular/platform-browser` should not lead to a mention, but you're saying sometimes there is a team in the org with the same name and then it does cause mentions?

FWIW what we do instead of remove the `@`, is insert a zero-width space, and wrap the mention in a code block, maybe we need to include teams in the regex that does this escaping

shu-mutou:
@jurre, thank you for showing the check inside dependabot!

And our community is checking mention by [this regex](https://github.com/kubernetes/test-infra/blob/55dab213bff9ed663013072c14f17c19dc9c625f/prow/plugins/invalidcommitmsg/invalidcommitmsg.go#L66), and this check prevents merging PR.

[...]

PR #3404, opened 2021-03-30
Pull Requests: Sanitize team mentions

jurre:
When we pull in changelogs and commits etc we sanitize user mentions
(`@jurre`), by wrapping them in a codeblock and inserting a zero-width
space between the `@` and the username, and wrapping that in a link to
the profile, to prevent users from getting notifications from Dependabot PRs.

Something that we didn't cover before yet were team-mentions, e.q:
@dependabot/reviewers. These are slightly tricky in the sense that there
are ecosystems (npm for example) that have package names that follow the
same pattern. There is no way to distinguish between a package name and
a team mention, so we treat them the same and sanitize them but do not
attempt to link, as this is expected to lead to a lot of false
positives.

Fixes https://github.com/dependabot/dependabot-core/issues/3378

### DIFF
--- a/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
@@ -15,6 +15,8 @@ module Dependabot
           (?:issue|pull)s?/(?<number>\d+)
         }x.freeze
         MENTION_REGEX = %r{(?<![A-Za-z0-9`~])@#{GITHUB_USERNAME}/?}.freeze
+        # regex to match a team mention on github
+        TEAM_MENTION_REGEX = %r{(?<![A-Za-z0-9`~])@(?<org>#{GITHUB_USERNAME})/(?<team>#{GITHUB_USERNAME})/?}.freeze
         # End of string
         EOS_REGEX = /\z/.freeze
         COMMONMARKER_OPTIONS = %i(
@@ -35,8 +37,10 @@ module Dependabot
             text, :LIBERAL_HTML_TAG, COMMONMARKER_EXTENSIONS
           )
 
+          sanitize_team_mentions(doc)
           sanitize_mentions(doc)
           sanitize_links(doc)
+
           mode = unsafe ? :UNSAFE : :DEFAULT
           doc.to_html(([mode] + COMMONMARKER_OPTIONS), COMMONMARKER_EXTENSIONS)
         end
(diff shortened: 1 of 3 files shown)


common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
```diff
@@ -117,6 +141,27 @@ def build_mention_nodes(text)
...
+          until scan.eos?
+            line = scan.scan_until(TEAM_MENTION_REGEX) ||
+                   scan.scan_until(EOS_REGEX)
+            line_match = line.match(TEAM_MENTION_REGEX)
+            mention = line_match&.to_s
+            text_node = CommonMarker::Node.new(:text)
+            text_node.string_content = line
+            next unless mention
+
+            text_node.string_content = line_match.pre_match
```
feelepxyz, line 157: Looks like the text node isn't used unless there's a mention? Would something like this work?

```suggestion
            next unless mention

            text_node = CommonMarker::Node.new(:text)
            text_node.string_content = line_match.pre_match
```

common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
```diff
@@ -117,6 +141,27 @@ def build_mention_nodes(text)
...
+            nodes << text_node
+            nodes << build_mention_link_text_nodes(mention.to_s).first
```
feelepxyz, line 159: Maybe? 
```suggestion
            nodes += build_mention_link_text_nodes(mention.to_s)
```

common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
feelepxyz: Or should we also do what we do above in `build_mention_nodes` and always add the text_node if there is no mention?

e.g.

```
if mention
  text_node.string_content = line_match.pre_match
  nodes << text_node
  nodes << build_mention_link_text_nodes(mention.to_s).first
else
  text_node.string_content = line
  nodes << text_node
end
```

I can't actually remember what that case does for regular mentions but maybe we have some weird test cases covering it?

feelepxyz: @jurre just thought of above, might not be needed but maybe worth checking the existing test suite to see when we end up in the `else` for mentions?

jurre: > Or should we also do what we do above in build_mention_nodes and always add the text_node if there is no mention?

Ah yes we do, otherwise we'll end up dropping lines without mentions. Will add a test to cover this

jurre: Ah, no we don't. The reason that case is there for regualar mentions is because @feelepxyz/ doesn't need to be sanitized, that's the `&& !mention.end_with?("/")` case

feelepxyz: Ah gotcha! Thanks for verifying 👍

common/spec/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer_spec.rb
```diff
@@ -200,6 +195,26 @@
...
+        let(:text) { "Thanks @dependabot/reviewers" }
+
+        it "sanitizes the team mention" do
+          expect(sanitize_links_and_mentions).to eq(
+            "<p>Thanks <code>@\u200Bdependabot/reviewers</code></p>\n"
+          )
+        end
+      end
+
+      context "multiple team mentions" do
```
feelepxyz, line 209: Worth still adding another test case where there's regular text between the team mention or modifying this test case?

feelepxyz (approved):
Thanks for tackling this 💯  looks like there's an integration test failing where this is now sanitising a team mention.

PR #3406, opened 2021-03-30
Bundler 2 [pre-release]: Run the entire bundler test suite with v2 Gemfile.lock fixtures

brrygrdn:
Extracts from #3350 

This PR adds a helper method with checks the correct project folder depending on whether the test suite is targeting bundler1 or bundler2.

This ensures that we use Gemfile.lock fixtures that have appropriate `BUNDLED WITH` strings to ensure we correctly detect and handle them with the correct wrapped native bundler in future.


brrygrdn:
I think the last commit fixes some test failures I have locally in `bundler2`, will confirm.

feelepxyz (approved):
💥💯

LGTM! Skimmed all lockfiles and didn't spot any missing `BUNDLED WITH` statements.

PR #3408, opened 2021-03-30
Bundler 2: Add test for bundler dependency

jurre:
On v1, gems with a new version with a dependency on bundler 2 would fail to update,
this has since been fixed as demonstrated by this test.


feelepxyz (approved):
Nice! 💯

bundler/spec/dependabot/bundler/file_updater_spec.rb
```diff
@@ -1596,6 +1596,33 @@
...
+        [{
+          file: "Gemfile",
+          requirement: "~> 2.2.1",
+          groups: [],
+          source: nil
+        }]
+      end
+
+      it "raises an error", :bundler_v1_only do
+        expect { updated_gemfile.content }.to raise_error(/Bundler could not find compatible versions for gem "bundler"/)
```
feelepxyz, line 1618: Nit: calling a method on a raise feels a bit weird but not blocking 

```suggestion
        expect { updated_gemfile }.to raise_error(/Bundler could not find compatible versions for gem "bundler"/)
```

bundler/spec/dependabot/bundler/file_updater_spec.rb
```diff
@@ -1443,6 +1443,33 @@
...
+      let(:requirements) do
+        [{
+          file: "Gemfile",
+          requirement: "~> 2.2.1",
+          groups: [],
+          source: nil
+        }]
+      end
+
+      it "raises an error", :bundler_v1_only do
```
jurre, line 1464: I don't think there's a way to trigger this error anymore, it'll just not propose an update I think

brrygrdn (approved):
🚀

bundler/spec/dependabot/bundler/file_updater_spec.rb
jurre: That was my thought as well, but turns out the UpdateChecker isn't currently smart enough, demonstrated here: https://github.com/dependabot/dependabot-core/pull/3418

PR #3413, opened 2021-03-30
If SUITE_NAME is not specified, default to bundler2

brrygrdn:
This branch inverts our current default testing behaviour to use bundler2 by default since it is the latest release.

This mostly impacts developer workflows where it is anticipated they won't set the SUITE_NAME flag unless specifically testing a bunder 1-specific change.

This will have no affect on CI but just sets the expectation that in future, the builder1 CI build is a regression test and all new work will focus the latest release.


feelepxyz (approved):

feelepxyz:
Seems like that composer spec fails intermittently so probably fine to ignore until we fix it

PR #3414, opened 2021-03-30
Implement delete/create action in GitLab Client

### LINKED ISSUE #3409: Azure/Gitlab/Bitbucket clients do not handle file deletion + file create is missing
**Package manager/ecosystem**
go_modules
**Manifest contents prior to update**
github.com/buger/jsonparser v1.0.0
**Updated dependency**
github.com/buger/jsonparser v1.1.1
The vendoring pulls in the dependencies, it is possible that vendoring either creates new files, or deletes files, besides updating them, which works for most implementations. I think the following things are missing:

- The Azure, Gitlab and Bitbucker client implementations miss the `file.deleted` check to see if a file has been deleted to send a delete action in the pull request
- When new files are created, the creation of the pull request for Gitlab fails, since everything is [added as update](https://github.com/dependabot/dependabot-core/blob/488259938c04c846b9943d501d5107243665b6ab/common/lib/dependabot/pull_request_creator/gitlab.rb#L97), and Gitlab recognizes that the file doesn't exist so there is nothing to update
- `DependencyFile` should probably get a `created` property for new files so clients that see create/update as a different action are also supported

The following error is raised when new files are created by vendoring:
```
WARN: Gitlab::Error::BadRequest: Server responded with code 400, message: A file with this name doesn't exist
```

jerbob92:
Fixes #3409


jerbob92:
@jurre added tests on this one too

common/lib/dependabot/file_updaters/vendor_updater.rb
```diff
@@ -27,7 +27,12 @@ def updated_vendor_cache_files(base_directory:)
           )
           changed_paths = status.split("\n").map { |l| l.split(" ") }
           changed_paths.map do |type, path|
+            # The following types are possible to be returned:
+            # M = Modified = Default for DependencyFile
+            # D = Deleted
+            # ?? = Untracked = Created
             deleted = type == "D"
+            created = type == "??"
```
jurre, line 35: This is starting to look like we might need a `type` or `mode` struct/variable that can be either `:modified, :deleted, :created`. Fine to leave this in for now but since these are mutually exclusive it seems like a bit of a smell

jurre:
Looks good! Should we test the changes to `common/lib/dependabot/pull_request_updater/gitlab.rb` as well? Would be nice to have a test to verify that works as intended

common/lib/dependabot/file_updaters/vendor_updater.rb
jerbob92: Nah I think you're right. I just added operation to DependencyFile and removed deleted/created.

jerbob92:
> Looks good! Should we test the changes to `common/lib/dependabot/pull_request_updater/gitlab.rb` as well? Would be nice to have a test to verify that works as intended

For some reason I did not see the tests for gitlab before, I'll update them :)

common/lib/dependabot/file_updaters/vendor_updater.rb
jurre: Ah that's a great change but I'm slightly worried that this api is used externally, I'm going to check out what we need to keep, might be just the `#deleted?` flag. Apologies, I should have mentioned that before

jerbob92: Yeah probably only deleted, I'll make sure its still there

common/lib/dependabot/dependency_file.rb
```diff
@@ -5,23 +5,29 @@
...
+    class Operation
+      UPDATE = "update"
+      CREATE = "create"
+      DELETE = "delete"
+    end
+
     def initialize(name:, content:, directory: "/", type: "file",
                    support_file: false, symlink_target: nil,
-                   content_encoding: ContentEncoding::UTF_8, deleted: false)
+                   content_encoding: ContentEncoding::UTF_8, operation: Operation::UPDATE)
```
jurre, line 23: Unfortunately this initializer is used publicly, so this would result in a breaking change that I'm slightly uncomfortable with making right now. Adding a new param with a default value is fine, but we'll need to keep the `deleted:` param around to avoind breaking downstream dependants of dependabot-core.

We should start adding this to inline-docs to make that more clear (also for ourselves)

jerbob92: Ok, I'll make sure deleted will stay and work

jerbob92: @jurre Do you also want to keep the `attr_accessor`? Or just the param in the initializer?

jerbob92:
@jurre should be good now. I removed the `attr_accessor` and added a `deleted` method so at least the logic about the operation still comes from the operation attribute.

I extended the unit tests, for both the DependencyFile and GitLab, however, I noticed that the GitLab tests don't actually check the posted body (unlike the GitHub test), do you want me to add that?

common/lib/dependabot/dependency_file.rb
```diff
@@ -75,8 +86,12 @@ def support_file?
       @support_file
     end
 
+    def deleted
```
jurre, line 89: This is not exactly the same as having an `attr_accessor`, the accessor also acts as a setter, we'll need to keep that behavior around unfortunately. Maybe we should reimplement that by setting the `@operation` to `DELETE`:

```ruby
def deleted=(deleted)
  operation = deleted ? Operation::DELETE : Operation::UPDATE
end
```

We should eventually deprecate the `deleted` field and rely on the `operation` instead, but we currently don't have a formal deprecation policy so I'll need to give that a think

jerbob92: Ah, my bad. Not that much Ruby experience. 
That sounds like a good solution to me.

Can you tell me whether this one also need to be kept:
```
def deleted?
  deleted
end
```

Is that something different compared to `def deleted` now that it's not an `attr_accessor` anymore?

jerbob92: Done :) And added a test for the setter.
Do we need to add any deprecation notices? I don't really see any through the codebase.

jurre: Yeah we've historically never really deprecated many things, and when we did we've sort of pushed through breaking changes and dealt with them on our own side. I'll check-in with the team if we want to introduce a more thorough process

jurre:
Got around to testing this on our internal projects just now, there is a small "breaking" change, in that the serialization for the DependencyFile objects has an extra attribute, but I think that just requires us updating some tests internally but it should not cause any further issues.

cc @feelepxyz could I get a pair of second eyes on this?

jurre (approved):

PR #3417, opened 2021-03-30
Generate alternatives for every git source

### LINKED ISSUE #3416: Add https alternative for every git_source
At this moment Dependabot only creates ssh alternatives in the git config (insteadOf) for github.com, but I think it would make more sense to do this for every `git_source` credential that's configured in the credentials.

jerbob92:
Fixes #3416


common/lib/dependabot/shared_helpers.rb
```diff
@@ -234,13 +205,39 @@ def self.configure_git_credentials(credentials)
...
         git_store_content += authenticated_url + "\n"
+        configure_git_to_use_https(cred.fetch('host'))
       end
 
       # Save the file
       File.write("git.store", git_store_content)
     end
     # rubocop:enable Metrics/PerceivedComplexity
 
+    def self.configure_git_to_use_https(host)
```
jurre, line 216: Feel free to:

```suggestion
    def self.configure_git_to_use_https(host) # rubocop:disable Metrics/AbcSize
```

I don't think that abstracting any of this code is useful or even an improvement

common/lib/dependabot/shared_helpers.rb
```diff
@@ -234,13 +205,39 @@ def self.configure_git_credentials(credentials)
           "@#{cred.fetch('host')}"
 
         git_store_content += authenticated_url + "\n"
+        configure_git_to_use_https(cred.fetch('host'))
```
jurre, line 208:
```suggestion
        configure_git_to_use_https(cred.fetch("host"))
```

Should fix the rubocop violation

jerbob92:
@jurre Sure, will add one tomorrow that validates the content of the git config.
Can you explain me why the tests fail? Looks like the default setup doesn't have a github credential. 

Do you want me to make a fallback, something like this:
```
      # Make sure we always have https alternatives for github.com.
      configure_git_to_use_https("github.com") if github_credential.nil?
```

jerbob92:
@jurre tests have been added, let me know what you think

common/spec/dependabot/shared_helpers_spec.rb
```diff
@@ -314,4 +314,131 @@
...
     end
   end
+
+  describe ".with_git_configured" do
+    # rubocop:disable Layout/LineLength
+    config_header = "Generated by dependabot/dependabot-core"
+    credentials_helper = "[credential]\n\thelper = !#{Dependabot::SharedHelpers.credential_helper_path} --file #{Dir.pwd}/git.store"
+
+    def alternatives(host)
+      "[url \"https://#{host}/\"]\n\tinsteadOf = ssh://git@#{host}/\n\tinsteadOf = ssh://git@#{host}:\n\tinsteadOf = git@#{host}:\n\tinsteadOf = git@#{host}/\n\tinsteadOf = git://#{host}/"
```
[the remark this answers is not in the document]
jerbob92: Yes, you're right. I was looking for something like this :) Updated

common/spec/dependabot/shared_helpers_spec.rb
```diff
@@ -314,4 +314,140 @@
...
+    end
+
+    let(:git_config) { -> { `cat ~/.gitconfig` } }
+    let(:git_credentials) { -> { `cat #{Dir.pwd}/git.store` } }
+
+    context "when providing no extra credentials" do
+      let(:credentials) { [] }
+
+      it "creates a .gitconfig that contains the Dependabot header" do
+        expect(with_git_configured[0]).to include(config_header)
```
jurre, line 352: It's not immediately clear to be what the first element of `with_git_configured` is, could we pull out a `let` that describes it?

```ruby
let(:configured_git_config) { with_git_configured.first }
let(:configured_git_credentials) { with_git_configured.last }
```

Or even two separate `let`s, they don't _need_ to be a `subject`

```ruby
let(:configured_git_config) do
  Dependabot::SharedHelpers.with_git_configured(credentials: credentials) do
    `cat ~/.gitconfig`
  end
end
```

jurre (approved):
This looks great, I'm commenting in-line with a few suggestions to make the tests a bit more readable but awesome work

common/spec/dependabot/shared_helpers_spec.rb
jerbob92: Yeah makes sense. What do you think of this:

```go
def with_git_configured(&block)
  Dependabot::SharedHelpers.with_git_configured(credentials: credentials) { block.call }
end

let(:configured_git_config) { with_git_configured { `cat ~/.gitconfig` } }
let(:configured_git_credentials) { with_git_configured { `cat #{Dir.pwd}/git.store` } }
```

jurre: I dig it 👍

brrygrdn (approved):

jurre:
✨ Thanks @jerbob92!

PR #3418, opened 2021-03-30
Bundler 2: Test UpdateChecker with dependencies on bundler itself

jurre:
Testing the UpdateChecker with dependencies that have a dependency on Bundler. This doesn't currently take the bundler version into account, so the v1 spec is disabled. Ideally we would detect this in the `VersionResolver`, so I left the test in.


bundler/spec/dependabot/bundler/update_checker_spec.rb
```diff
@@ -363,6 +363,36 @@
...
+          file: "Gemfile",
+          requirement: "~> 2.2.1",
+          groups: [],
+          source: nil
+        }]
+      end
+
+      context "bundler v1", :bundler_v1_only do
+        # Surprisingly, this returns the incompatible version 3.0.0.
+        xit { is_expected.to eq(Gem::Version.new("2.2.1")) }
```
feelepxyz, line 389: Looks like this is testing `latest_version`, think it should be `latest_resolvable_version` but also seeing weird failures on my branch where the test cases are inverted 🤔 https://github.com/dependabot/dependabot-core/pull/3424/checks?check_run_id=2244656544

PR #3420, opened 2021-03-30
Improve maintainability of labeler class with labels class extract

codisart:


jurre:
Thanks for contributing to Dependabot @codisart, but I'm a little hesitant to make cosmetic changes like this to parts of the codebase that we basically never need to change. I agree that this code looks cleaner, but the apparent benefit is not totally clear to me. Unfortunately things like the label setup etc are not always tested as thoroughly in the code that we use to integrate Dependabot with GitHub, and the risk of potentially breaking something is a little higher than I'd like.

I hope that doesn't sound harsh or puts you off from opening more contributions.

It might be a good idea to open an issue to discuss things to improve before going in and doing the work, if you ping me I'm happy to engage there.

💛

xlgmokha:
It looks like there hasn't been much action on this PR for a few months. I'm going to close for now. Feel free to re-open when the discussion continues.

PR #3426, opened 2021-04-01
CI: Only bundle install for ecosystem under test

jurre:
Currently in the CI container we bundle install for _every_ package,
this ends up taking 4 to 5 minutes on every build for every ecosystem.

Since the tests are split up per ecosystem, we only need to install the
gems for the package that is being tested in this matrix step.

This also introduces a `script/ci-test` per ecosystem, it currently runs
both rubocop and rspec for all ecosystems, but the expectation is that
these scripts might diverge per ecosystem.

Running both rspec and rubocop in the same script is done to prevent us
from having to `bundle install` for each of those steps.

We could take this approach a step further and split out a CI container
per ecosystem, and that would let us run rubocop and rspec as separate
steps, but I don't think this is currently worth it.

This should shave about 5 minutes of every matrix build.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,9 +102,6 @@ jobs:
         if: matrix.suite.name == 'python'
         run: |
           docker run --rm "$CORE_CI_IMAGE" bash -c "pyenv exec flake8 python/helpers/. --count --exclude=./.*,./python/spec/fixtures --show-source --statistics"
-      - name: Run Ruby Rubocop linting
-        run: |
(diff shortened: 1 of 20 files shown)


brrygrdn (approved):
💥  🚀 

This is awesome

.github/workflows/ci.yml
```diff
@@ -123,4 +120,4 @@ jobs:
       - name: Run ${{ matrix.suite.name }} tests with rspec
```
feelepxyz, line 120:
```suggestion
      - name: Run ${{ matrix.suite.name }} tests
```

feelepxyz (approved):
Nice! Makes a tonne of sense 👍

PR #3428, opened 2021-04-01
CI: Use buildkit caching when building Dockerfile.ci

feelepxyz:
Leverage Docker's buildkit caching when building `Dockerfile.ci`.

Using docker volumes instead of copying all the files over to the ci image as the buildkit cache was previously hanging when building the ci image.

The caching change was previously [reverted in this commit](a89aca9#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,7 @@ jobs:
           echo "BASE_IMAGE=ubuntu:18.04" >> $GITHUB_ENV
           echo "CORE_IMAGE=dependabot/dependabot-core" >> $GITHUB_ENV
           echo "CORE_CI_IMAGE=dependabot/dependabot-core-ci" >> $GITHUB_ENV
+          echo "CODE_DIR=/home/dependabot/dependabot-core" >> $GITHUB_ENV
       - name: Log in to Docker registry
         run: |
           if [ -n "${{ secrets.DOCKER_USERNAME }}" ] && [ -n "${{ secrets.DOCKER_PASSWORD }}" ]; then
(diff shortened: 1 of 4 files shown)


jurre:
I think this needs to be rebased on main before it'll work?

feelepxyz:
This can be simplified once https://github.com/dependabot/dependabot-core/pull/3430 lands

.github/workflows/ci.yml
```diff
@@ -43,11 +43,12 @@ jobs:
...
           echo "CORE_IMAGE=dependabot/dependabot-core" >> $GITHUB_ENV
-          echo "CORE_CI_IMAGE=dependabot/dependabot-core-ci" >> $GITHUB_ENV
+          echo "CORE_CI_IMAGE=ghcr.io/${{ github.repository_owner }}/dependabot-core/dependabot-core-ci" >> $GITHUB_ENV
+          echo "CODE_DIR=/home/dependabot/dependabot-core" >> $GITHUB_ENV
       - name: Log in to Docker registry
         run: |
-          if [ -n "${{ secrets.DOCKER_USERNAME }}" ] && [ -n "${{ secrets.DOCKER_PASSWORD }}" ]; then
-            echo ${{ secrets.DOCKER_PASSWORD }} | docker login -u ${{ secrets.DOCKER_USERNAME }} --password-stdin
+          if [[ -n "${{ secrets.GHCR_PAT }}" ]]; then
+             echo "${{ secrets.GHCR_PAT }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
```
feelepxyz, line 51: Tried to use the `GITHUB_TOKEN` here but doesn't seem to work even after giving the repo write access to the org package

.github/workflows/ci.yml
```diff
@@ -103,14 +84,147 @@ jobs:
...
+            --env "BUNDLE_PATH=/home/dependabot/.bundle" --env "BUNDLE_BIN=.bundle/binstubs" \
+            --env "PATH=.bundle/binstubs:$PATH:/home/dependabot/.bundle/bin" \
+            -w "/home/dependabot/dependabot-core"
+            "$CORE_CI_IMAGE" bash -c \
             "cd /home/dependabot/dependabot-core/bundler/helpers/v2 && BUNDLER_VERSION=2 bundle install && BUNDLER_VERSION=2 bundle exec rspec spec"
       - name: Run ${{ matrix.suite.name }} tests
         run: |
-          docker run --env "CI=true" --env "DEPENDABOT_TEST_ACCESS_TOKEN=$DEPENDABOT_TEST_ACCESS_TOKEN" --env "SUITE_NAME=${{ matrix.suite.name }}" --rm "$CORE_CI_IMAGE" bash -c \
+          docker run \
+            -v "$(pwd)/.rubocop.yml:$CODE_DIR/.rubocop.yml" \
```
feelepxyz, line 109: Would be nice to share this between the dev-shell setup

Dockerfile.ci
```diff
@@ -1,74 +1,7 @@
...
-ARG CODE_DIR=/home/dependabot/dependabot-core
-WORKDIR ${CODE_DIR}
-
-ENV BUNDLE_PATH="/home/dependabot/.bundle" \
-  BUNDLE_BIN=".bundle/binstubs" \
-  PATH=".bundle/binstubs:$PATH:/home/dependabot/.bundle/bin"
 
 RUN pyenv install -s 3.6.13
 RUN pyenv install -s 3.7.1
 RUN pyenv install -s 3.7.10
```
jurre, line 5: Unrelated to this PR, but I noticed that these still aren't being cached currently. Do you have any idea why that could be?

feelepxyz: Not sure, are they being installed again in the tests?

feelepxyz: First hunch was some permission issue but I think these, as the tests, are running as root so should be fine.

jurre: No, they are being installed in the docker-ci build step in the action actually, when IMO all these layers should be cached?

feelepxyz: Oh wait you mean on this PRs build? I busted all the docker images used here so everything needs to rebuild.

feelepxyz: Ah yeah I see, I wonder if the change in this PR to use the inline buildkit cache would help with this?

jurre: Curious to find out! It might 🤞

.github/workflows/ci.yml
```diff
@@ -44,6 +44,7 @@ jobs:
           echo "BASE_IMAGE=ubuntu:18.04" >> $GITHUB_ENV
           echo "CORE_IMAGE=dependabot/dependabot-core" >> $GITHUB_ENV
           echo "CORE_CI_IMAGE=dependabot/dependabot-core-ci" >> $GITHUB_ENV
```
feelepxyz, line 46: Went back to using the docker registry as it seemed like ghcr doesn't do well with the inline caching, possibly not storing the cache properly so was re-building every time

.github/workflows/ci.yml
feelepxyz: Going to try and clean this up in a follow up, with a sahred `docker-shell` script we can use from here and `docker-dev-shell`

.github/workflows/docker.yml
```diff
@@ -13,27 +13,33 @@ jobs:
...
       - name: Log in to the Docker registry
         run: |
           echo "${{ secrets.DOCKER_PASSWORD }}" | docker login -u ${{ secrets.DOCKER_USERNAME }} --password-stdin
-      - name: Push image to packages (latest)
-        if: "!contains(github.ref, 'refs/tags')"
+      - name: Push latest image
         run: |
-          docker push "dependabot/dependabot-core:latest"
-      - name: Push image to packages (tagged)
+          docker push "$CORE_IMAGE:latest"
```
feelepxyz, line 39: Changed this to always push the `latest` tag, can't see a reason where we don't want to push latest when tagging a release

jurre (approved):
I think the changes look good and so far the logs here look promising 👍

PR #3429, opened 2021-04-01
CI: Use GitHub Container Registry for CI images

feelepxyz:
Attempting to use ghrc.io for CI images. This will probably not work on
forks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,15 +50,10 @@ jobs:
           echo "BRANCH_REF=$(echo '${{ github.ref }}' | sed -E 's/[^A-Za-z0-9]+/-/g')" >> $GITHUB_ENV
           echo "BASE_IMAGE=ubuntu:18.04" >> $GITHUB_ENV
           echo "CORE_IMAGE=dependabot/dependabot-core" >> $GITHUB_ENV
-          echo "CORE_CI_IMAGE=dependabot/dependabot-core-ci" >> $GITHUB_ENV
+          echo "CORE_CI_IMAGE=ghcr.io/${{ github.repository_owner }}/dependabot/dependabot-core-ci" >> $GITHUB_ENV
       - name: Log in to Docker registry
         run: |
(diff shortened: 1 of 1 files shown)


jurre:
Do we assume this will be faster, or are we trying to find out?

feelepxyz:
@jurre yeah mostly trying, assuming it would be a lot faster to pull if also hosted on azure next to actions runners somewhere 🤞

.github/workflows/ci.yml
```diff
@@ -50,14 +50,14 @@ jobs:
...
       - name: Log in to Docker registry
         run: |
-          if [ -n "${{ secrets.DOCKER_USERNAME }}" ] && [ -n "${{ secrets.DOCKER_PASSWORD }}" ]; then
-            echo ${{ secrets.DOCKER_PASSWORD }} | docker login -u ${{ secrets.DOCKER_USERNAME }} --password-stdin
-            echo "DOCKER_LOGGED_IN=true" >> $GITHUB_ENV
-          else
+          if [[ -z "${{ secrets.GHCR_PAT }}" ]]; then
             echo "No Docker credentials, skipping login"
+          else
+            echo "${{ secrets.GHCR_PAT }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
```
thepwagner, line 59: Question: could this be `secrets.GITHUB_TOKEN`? https://github.blog/changelog/2021-03-24-packages-container-registry-now-supports-github_token/

It looks like you may have started there and fell back to a PAT.

thepwagner: Question: who does this PAT belong to? It may be worth documenting here so access can be managed.

thepwagner (approved):
Noice! LGTM, though I think we should avoid the PAT if possible.

feelepxyz:
Tried this here https://github.com/dependabot/dependabot-core/pull/3428 and didn't see any improvements, if anything the downloads where slower (around ~2mins vs 1.5min) and not sure it's storing inline caching as builds where often re-building from scratch even though there should have built from an up-to-date cache.

feelepxyz:
Looks like buildkit support is on the radar but not done yet https://github.slack.com/archives/CJL5ATGUC/p1607096013102400

PR #3431, opened 2021-04-01
CI: Pre-install python versions used in tests

jurre:
I hacked together a little script to figure out which python versions
we're actually installing in our test suite, turns out it's not that
many:

https://github.com/dependabot/dependabot-core/pull/3425/commits/9f7251b423dccd474077da9227c3dbfb14f3535f

But this leads to almost 10 minutes of time spent _every python run_.

This change runs the install steps in the CI Dockerfile so they can be
cached.

It does nothing to prevent newer versions from python being added in new
tests, but it's a fairly safe way to shave 10 minutes of the build.

### DIFF
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -6,6 +6,10 @@ ENV BUNDLE_PATH="/home/dependabot/.bundle" \
   BUNDLE_BIN=".bundle/binstubs" \
   PATH=".bundle/binstubs:$PATH:/home/dependabot/.bundle/bin"
 
+RUN pyenv install -s 3.6.13
+RUN pyenv install -s 3.7.1
+RUN pyenv install -s 3.7.10
+
 COPY .rubocop.yml /home/dependabot/dependabot-core/
 
 RUN mkdir -p \


Dockerfile.ci
```diff
@@ -6,6 +6,10 @@ ENV BUNDLE_PATH="/home/dependabot/.bundle" \
   BUNDLE_BIN=".bundle/binstubs" \
   PATH=".bundle/binstubs:$PATH:/home/dependabot/.bundle/bin"
 
+RUN pyenv install -s 3.6.13
```
brrygrdn, line 9: Worth a comment on why we've added these for future us?

hmarr:
Epic! Is there some way we could catch regressions (i.e. new versions being installed)?

brrygrdn (approved):
👍🏻  This is a solid improvement, and a good trade off for CI time compared to fine-tooth test tuning which we can do over time.

hmarr:
What will this do to non-python CI jobs, btw? Will it slow them down as we now need to install python 3 times?

jurre:
> Epic! Is there some way we could catch regressions (i.e. new versions being installed)?

Maybe, but it's a little tricky. Versions are currently installed all over the place. We might be able to DRY that up and add a way to check for uninstalled versions in tests, but it'll be a bigger investment than this PR so I punted on it for now.

> What will this do to non-python CI jobs, btw? Will it slow them down as we now need to install python 3 times?

It should be cached in the CI docker image and have basically no impact once it's in the cache.

Dockerfile.ci
jurre: Yeah, good call.

jurre: :/ I forgot to commit this, will open a separate pr

PR #3432, opened 2021-04-01
[ci skip] Explain pre-installed pythons in Dockerfile.ci

jurre:
In https://github.com/dependabot/dependabot-core/pull/3431 we introduced
some pre-installed versions of Python in an effort to speed up our test.

@brrygrdn rightfully pointed out it'd be good to add a note as to why
this is here. This is said note.

### DIFF
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -6,6 +6,9 @@ ENV BUNDLE_PATH="/home/dependabot/.bundle" \
   BUNDLE_BIN=".bundle/binstubs" \
   PATH=".bundle/binstubs:$PATH:/home/dependabot/.bundle/bin"
 
+# We pre-install these versions of python as they're known to be used in the
+# test suite. These used to be installed in-line, and pre-installing them helps
+# with the performance of the python test suite.
 RUN pyenv install -s 3.6.13
 RUN pyenv install -s 3.7.1
 RUN pyenv install -s 3.7.10


Dockerfile.ci
```diff
@@ -6,6 +6,9 @@ ENV BUNDLE_PATH="/home/dependabot/.bundle" \
   BUNDLE_BIN=".bundle/binstubs" \
   PATH=".bundle/binstubs:$PATH:/home/dependabot/.bundle/bin"
 
+# We pre-install these versions of python as they're known to be used in the
+# test suite. These used to be installed in-line, and pre-installing them helps
+# with the performance of the python test suite.
```
feelepxyz, line 11: Where did you find these versions? Could be useful with a ref if we end up wanting to change. Not related 
 to this PR but wondering if we could reduce this to one non-default version?

jurre: I wrote some code to figure it out which I referenced in the previous PR: https://github.com/dependabot/dependabot-core/commit/9f7251b423dccd474077da9227c3dbfb14f3535f

It's a little gnarly and building a general-purpose version of it would require some cleanup in the python codebase.

Agree it'd be nice to reduce that to a single version, but unsure how much work it would be, and that cutting down the time to install them on CI seemed like a good first step

feelepxyz: Ah gotcha, maybe something we can hook into notifications that we set up for recording the bundler version?

Wonder where those versions even get extracted from, can't find them in the codebase by grepping 🤔

feelepxyz (approved):
Lets get this in as is and do any improvements to tracked versions as follow ups 👍

PR #3437, opened 2021-04-02
Sanitize mentions for merge requests in Gitlab

andrcuns:
I have encountered an issue that quite often handles in gitlab will be the same as on github. This creates an issue where PR descriptions in various projects can send notifications to people mentioned in release notes.

I'm not entirely sure if same issue could exist on bitbucket as well and maybe this check should be removed all together. For now it seemed safer to explicitly just add gitlab to conditional.

### DIFF
--- a/common/lib/dependabot/pull_request_creator/message_builder/metadata_presenter.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder/metadata_presenter.rb
@@ -245,7 +245,7 @@ module Dependabot
         end
 
         def sanitize_links_and_mentions(text, unsafe: false)
-          return text unless source.provider == "github"
+          return text unless source.provider == "github" || source.provider == "gitlab"
 
           LinkAndMentionSanitizer.
             new(github_redirection_service: github_redirection_service).
--- a/common/spec/dependabot/pull_request_creator/message_builder_spec.rb
+++ b/common/spec/dependabot/pull_request_creator/message_builder_spec.rb
@@ -976,8 +976,8 @@ RSpec.describe Dependabot::PullRequestCreator::MessageBuilder do
             Dependabot::Source.new(provider: "gitlab", repo: "gocardless/bump")
           end
 
-          it "does not sanitize github links" do
-            expect(pr_message).not_to include(github_redirection_service)
+          it "sanitizes github links" do
+            expect(pr_message).to include(github_redirection_service)
           end
         end
 


feelepxyz (commented):
I'm not sure this is what you want, this will just sanitize links to github on pull requests created on GitLab. GitLab MergeRequests shouldn't create issue/PR references on github. 

Are you looking at sanitizing links to GitLab MergeRequests and Issues?

andrcuns:
Yes, exactly, you need to sanitize mentions in Gitlab Merge Requests as well.

There is currently an issue, that if dependabot creates some merge request where release notes contain a mention of @andrcuns for example, it would send a notification to this user because it is also a valid handle in Gitlab. This could result in quite a lot of spam to active contributors who have same username on Github and Gitlab.

feelepxyz (changes requested):
@andrcuns yeah makes sense 👍  we won't be able to rely on the `LinkAndMentionSanitizer` for this though as this only works for github users/links/PRs and re-writes `@userhandles` to github.com links. I think the way to go here would be adding a `GitLabLinkAndMentionSanitizer` and using this instead when the source is gitlab.

We're love a contribution to fix this and happy to review it.

andrcuns:
> @andrcuns yeah makes sense 👍 we won't be able to rely on the `LinkAndMentionSanitizer` for this though as this only works for github users/links/PRs and re-writes `@userhandles` to github.com links. I think the way to go here would be adding a `GitLabLinkAndMentionSanitizer` and using this instead when the source is gitlab.
> 
> We're love a contribution to fix this and happy to review it.

Well, technically it works as expected, because all users/links/PR's always reference Github, I don't think dependabot fetches any info from non Github sources, does it?

Here is an example: https://gitlab.com/dependabot-gitlab/dependabot/-/merge_requests/567. The changelog actually points to Github user profiles, which is who actually contributed the changes. I don't think it would make much sense to sanitize these user profiles and create links that point to Gitlab profile counter parts.

Do You think it makes sense to implement `GitLabLinkAndMentionSanitizer` which pretty much inherit's the `LinkAndMentionSanitizer`. At the moment I don't see what custom logic it might have given PR descriptions contain references only to Github related objects (users/PR's).

feelepxyz:
> Well, technically it works as expected, because all users/links/PR's always reference Github, I don't think dependabot fetches any info from non Github sources, does it?

It depends on the dependency, if the package is hosted on gitlab it will.

I'm still not following exactly what the problem is. The MergeRequest you link to has two user mentions, `@​honnix` and `
@​jerbob92`, these look like github usernames afaikt, but those users should not receive any notifications on github from these merge requests. They should only receive notifications on gitlab if they also exist there. The existing `LinkAndMentionSanitizer` won't help with this as it will start linking these users to github instead of gitlab.

Is the problem not with gitlab usernames getting mentions on gitlab?

andrcuns:
> > Well, technically it works as expected, because all users/links/PR's always reference Github, I don't think dependabot fetches any info from non Github sources, does it?
> 
> It depends on the dependency, if the package is hosted on gitlab it will.
> 
> I'm still not following exactly what the problem is. The MergeRequest you link to has two user mentions, `@​honnix` and ` @​jerbob92`, these look like github usernames afaikt, but those users should not receive any notifications on github from these merge requests. They should only receive notifications on gitlab if they also exist there. The existing `LinkAndMentionSanitizer` won't help with this as it will start linking these users to github instead of gitlab.
> 
> Is the problem not with gitlab usernames getting mentions on gitlab?

The problem has nothing to do with Github, it's Gitlab only :) I must be bad at explaining it.

The problem is exactly that these 2 users would receive a notification on Gitlab about a merge request in a project that has nothing to do with them simply because they are mentioned in the changelog.

The example MR works correctly and did not ping those 2 users, because it was created by a script that has the proposed changes as monkey patch and it transformed the usernames to urls to github.

I need to dig deeper in the code then, if description can reference release notes or changelog that comes from gitlab, then indeed it would need it's own sanitizer.

feelepxyz:
Ah gotcha, think I follow now. So it will work as expected if the mentioned users are always github users. I was thinking of the case where these users are only on gitlab, in which case this change would suddenly link them to non-existent github users.

Hm not entirely sure what the best approach is here. Will have a poke at the code to see if we know where we are fetching changelogs/release notes from and see if we can toggle based on this instead of the updated repo.

andrcuns:
Yes, exactly what I was thinking now. If it's possible to know where release notes come from, it should be pretty straight forward to implement the proper sanitation. I was not aware that release notes can even come from Gitlab so didn't think that linking to non existent Github accounts might be an issue. I guess similar problem can happen with referencing issues and PR's/MR's in changelog since it could reference to incorrect platform.

Let me know what you find. I'm willing to try and implement it.

common/lib/dependabot/pull_request_creator/message_builder/metadata_presenter.rb
```diff
@@ -244,7 +244,7 @@ def source_provider_supports_html?
         end
 
         def sanitize_links_and_mentions(text, unsafe: false)
-          return text unless source.provider == "github"
+          return text unless source.provider == "github" || source.provider == "gitlab"
```
xlgmokha, line 247: If I understand correctly we need something like...

```ruby
return text if running_on?(:github) && source.provider == 'github'
return text if running_on?(:gitlab) && source.provider == 'gitlab'
```

Does that capture the reason that this PR is blocked?

andrcuns: From what I understand, the issue is determining the source of `Release notes` and `Changelog` so we do not sanitize links pointing natively to Gitlab as Github links (I have not encountered such a case myself though).

If I understand correctly, `source.provider` is not that, it is basically a pointer where the dependabot-core runs and not where changelog comes from.

jeffwidman:
👋 Sorry for the delay here. Both @xlgmokha and @feelepxyz have transitioned off the core team, so this got dropped for a bit. 

> I was not aware that release notes can even come from Gitlab so didn't think that linking to non existent Github accounts might be an issue.

I haven't dug deeply into that code, but if we don't support querying non-GitHub sources today, I'm sure we'll support it in the future. Having correct/useful metadata all in one place in the PR description is one of the value-adds of Dependabot--I absolutely loved that about Dependabot when using it at my previous employer. So if we query a registry like PyPI for that data and it points us at GitLab, then we _should_ be grabbing this data from GitLab. Again, not sure if we do this today, but given the importance we'll do it in the future so I'm a bit hesitant to extend the current "hack" of assuming that if we're running on GitHub that all links point to GitHub.

>  I guess similar problem can happen with referencing issues and PR's/MR's in changelog since it could reference to incorrect platform.

Exactly.

> If it's possible to know where release notes come from, it should be pretty straight forward to implement the proper sanitation. 

Yeah, that's a much more sustainable / better solution IMO. I opened https://github.com/dependabot/dependabot-core/issues/5729 to track that. 

This probably isn't something we can get to anytime soon, but if you want to take a crack at it I'll try to be responsive on the PR reviews to keep it moving.

Given all this, I'm going to close this PR as "the right problem but wrong solution". And again, my apologies for the delay here!

jeffwidman:
Actually, I think I jumped the gun on closing this. 

The reality is that _most_ open source projects are hosted on GitHub, so extending the hack to include doing the sanitization on GitLab is probably fine in the short term.

Long term we'd still want to address this through looking at the metadata source (#5729), but in the meantime this PR probably improves the user experience for anyone running on other platforms.

I'm going to get a second opinion internally before merging, just to be sure I'm not missing something.

jeffwidman (approved):
After noodling on it and getting a second opinon, going to merge this!

I'm going to also open a follow-up PR to remove the guard altogether, as [we probably should sanitize on all platforms](https://github.com/dependabot/dependabot-core/pull/3437#issuecomment-1247597678). And then address the underlying bug in #5729.

PR #3445, opened 2021-04-06
cargo: support build annotations

thepwagner:
This PR amends the cargo updater to handle target versions with a build annotation, such as `1.5.0+build.1`.

First: there are several calls to `Cargo::Version#correct?` that probably expect to be validated against the regexp in `Cargo::Version`, but are actually evaluated against `Gem::Version`. The biggest implication of this bug is that the `RequirementsUpdater` will silently ignore the upgrade - so `Cargo.lock` will be updated but `Cargo.toml` will be unaffected:
* https://github.com/dependabot/dependabot-core/blob/2e5dd943912f005208ea1b43a5e4ad164029d96e/cargo/lib/dependabot/cargo/update_checker/requirements_updater.rb#L31
* https://github.com/dependabot/dependabot-core/blob/2e5dd943912f005208ea1b43a5e4ad164029d96e/cargo/lib/dependabot/cargo/update_checker/requirements_updater.rb#L42

Next up, the `RequirementsUpdater` would include the build annotation in the `Cargo.toml` update. This is accepted, but noise that users may not be expecting:
```
warning: /src/Cargo.toml: version requirement `0.7.0+std` for dependency `zstd` includes semver metadata which will be ignored, removing the metadata is recommended to avoid confusion
```

Finally, another place was relying on `Gem::Version#correct?` to reject Git commits as invalid versions. To address that I changed the `Cargo::Version::VERSION_PATTERN` to reject alphanumeric major versions.

Changing the `VERSION_PATTERN` is the biggest risk.

### DIFF
--- a/cargo/lib/dependabot/cargo/update_checker/requirements_updater.rb
+++ b/cargo/lib/dependabot/cargo/update_checker/requirements_updater.rb
@@ -94,7 +94,7 @@ module Dependabot
         end
 
         def update_version_string(req_string)
-          new_target_parts = target_version.to_s.sub(/\+.*/, '').split(".")
+          new_target_parts = target_version.to_s.sub(/\+.*/, "").split(".")
           req_string.sub(VERSION_REGEX) do |old_version|
             # For pre-release versions, just use the full version string
             next target_version.to_s if old_version.match?(/\d-/)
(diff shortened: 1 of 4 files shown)


jurre (approved):
Thanks for the clear PR description, this seems right to me

PR #3454, opened 2021-04-07
Bundler: Delete unused fixture files

feelepxyz:
Deleting all unused fixture files now that specs are using project based
fixtures.

Also moved `rubygems_responses` into the fixture folder.


jurre (approved):
Nice one, I was just doing the same thing 😄

PR #3459, opened 2021-04-07
Disabled poetry experimental new installer

honnix:
Workaround of https://github.com/python-poetry/poetry/issues/3010


jurre (approved):
Thanks!

PR #3463, opened 2021-04-08
v0.140.1

feelepxyz:
## v0.140.1, 8 April 2021

- Python: Disabled poetry experimental new installer @honnix
- GitLab: Implement delete/create action in client @jerbob92

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## v0.140.1, 8 April 2021
+
+- Python: Disabled poetry experimental new installer @honnix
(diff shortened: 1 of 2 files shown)


jurre:
@feelepxyz this should have been a minor version bump. The GitLab delete/create action required a breaking change to the DependencyFile object (it now has an extra attribute, meaning that serializing it to JSON no longer behaves the same)

feelepxyz:
@jurre thanks for flagging! Missed that change, will roll with this as it seems like it should only break tests but good shout nonetheless 👍

PR #3466, opened 2021-04-09
Final pass on bundler1 test cruft

brrygrdn:
This PR collects a few small changes to the bundler1 tests to ensure we aren't accidentally testing using bundler v2 native helpers anywhere


brrygrdn:
I had expected some test breakages given a few fixtures were actually sending the test suite down the bundler2 code line, but it seems the affected fixtures were all testing common functions, so I think this is all the change that is required.

bundler/spec/fixtures/projects/bundler1/version_specified_bundler_2/gems.locked
```diff
@@ -12,4 +12,4 @@ DEPENDENCIES
   statesman (~> 1.2.0)
```
feelepxyz, line 12: Should we move this project to bundler2? Looks like it's testing some bundler2 specific stuff?

brrygrdn: Good point, I suspect we should delete this fixture, assuming there is a bundler 2 version, and tag the test as bundler_v2_only.

I've went down a bit of a rabbit hole forking off from this as I notice we still have `using_bundler2?` in both the native helpers and it defaults to `false` in some places. I don't think it's harmful, but it is weird.

feelepxyz: Or should we update the project name? Was just confused at seeing `version_specified_bundler_2` in `bundler1` project folder

feelepxyz: > Good point, I suspect we should delete this fixture, assuming there is a bundler 2 version, and tag the test as bundler_v2_only.

Ah yes, makes sense 👍  didn't see your reply when replying

brrygrdn: Actually, I think you're right, there's nothing about the test that is actually bundler2 specific, I think the fixture name was just misleading. I've reinstated the test for bundler1 and renamed the fixture for both versions.

feelepxyz (approved):
👍

PR #3470, opened 2021-04-09
fix: detect No Space errors

xlgmokha:
# Why is this needed?

To detect out of disk errors and fail gracefully.

## What does this do?

It rescues `Errno::ENOSPC` and converts it to a `Dependabot::OutofDisk` error.

Example:

From job id `109063142`.

```plaintext
ERROR <job_109063142> Error processing github.com/sirupsen/logrus (Errno::ENOSPC)
ERROR <job_109063142> No space left on device @ fptr_finalize_flush - /home/dependabot/.gitconfig
ERROR <job_109063142> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-common-0.140.2/lib/dependabot/shared_helpers.rb:171:in `close'
ERROR <job_109063142> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-common-0.140.2/lib/dependabot/shared_helpers.rb:171:in `open'
ERROR <job_109063142> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-common-0.140.2/lib/dependabot/shared_helpers.rb:171:in `configure_git_to_use_https_with_credentials'
ERROR <job_109063142> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-common-0.140.2/lib/dependabot/shared_helpers.rb:158:in `with_git_configured'
ERROR <job_109063142> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-go_modules-0.140.2/lib/dependabot/go_modules/update_checker.rb:60:in `block in find_latest_resolvable_version'
ERROR <job_109063142> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-common-0.140.2/lib/dependabot/shared_helpers.rb:45:in `block (2 levels) in in_a_temporary_directory'
ERROR <job_109063142> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-common-0.140.2/lib/dependabot/shared_helpers.rb:45:in `chdir'
ERROR <job_109063142> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-common-0.140.2/lib/dependabot/shared_helpers.rb:45:in `block in in_a_temporary_directory'
ERROR <job_109063142> /usr/lib/ruby/2.6.0/tmpdir.rb:93:in `mktmpdir'
[...]

PR #3477, opened 2021-04-12
Add dependency to dependabot-maven from dependabot-gradle

xlgmokha:
In https://github.com/dependabot/dependabot-core/pull/3403 we added a dependency on `dependabot-maven` from the `dependabot-gradle` gem.
To ensure that the `dependabot-maven` gem is installed when we install the `dependabot-gradle` gem we need to declare `dependabot-maven` as a runtime dependency.

PR #3481, opened 2021-04-12
v0.141.0

thepwagner:
Release of [these commits](https://github.com/dependabot/dependabot-core/compare/v0.140.3...v0.141.0-release-notes)

As noted in the changelog, this is a potentially breaking change for consumers of the `dependabot/dependabot-core` docker image. You could work around it with `USER root` at the top of the consuming `Dockerfile`, but we encourage updating images to follow least privilege.

* https://github.com/dependabot/dependabot-core/pull/3472
* https://github.com/dependabot/dependabot-core/pull/3473
* https://github.com/dependabot/dependabot-core/pull/3470
* https://github.com/dependabot/dependabot-core/pull/3398

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## v0.141.0, 12 April 2021
+
+- Dockerfile: create a `dependabot` user and drop privileges
+  This is a potentially BREAKING change for consumers of the `dependabot/dependabot-core` docker image.
+- Maven/Gradle: Add option to use Gitlab access token for authentication against maven repositories @gringostar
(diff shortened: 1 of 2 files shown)


mctofu (approved):

xlgmokha:
Before you release this you may also want to include https://github.com/dependabot/dependabot-core/pull/3477. Otherwise, the `dependabot-gradle` dependency will be published without a runtime dependency on `dependabot-maven`.

thepwagner:
@xlgmokha will do! As it's a follow-up to something mentioned I won't append the changelog.

PR #3484, opened 2021-04-13
Stop installing custom python versions

feelepxyz:
Janked out the custom python version installs from tests as the ci image often fails to build from cache and takes 7+ mins.

We install python versions as `dependabot` in the [Dockerfile](https://github.com/dependabot/dependabot-core/blob/main/Dockerfile#L85-L86) so that gives us some confidence `pyenv install` works.

I've also split up the big `COPY` to hopefully build more from cache as any file change would trigger a full `COPY` which often took > 2 mins on my machine.

I've also rejigged the `.dockerignore` to ignore more things from `.gitignore`, spotted there was a mysterious `gitignored` line which I can't see documented anywhere so not sure it does what we think.

Question: Are we comfortable removing all custom python versions? 🤔


.dockerignore
```diff
@@ -1,8 +1,21 @@
...
+/.core-bash_history
+/.env
+/.envrc
 /.git
-/tmp/
-/helpers
-/spec
-gitignored
-/bundler/helpers/helpers/
-/python/helpers/helpers/
```
feelepxyz, line 8: Not sure what this was ignoring

jurre (approved):
Yeah, I think I'm OK doing this. We should technically still hit the `pyenv install` code, it will just use the already installed python.

thepwagner (approved):

bundler/helpers/v1/build
```diff
@@ -21,5 +20,6 @@ cd "$install_dir"
 
 # NOTE: Sets `BUNDLED WITH` to match the installed v1 version in Gemfile.lock
 # forcing native helpers to run with the same version
+BUNDLER_VERSION=2 bundle config set --local path ".bundle"
```
feelepxyz, line 23: Copying setup from v2

PR #3486, opened 2021-04-13
Adding update PR elements in azure PR updater

milind009:
This change allows to update PR elements such as `title`, `description`, `status`, set autocomplete on an existing PR,etc. for any PR in `azure devops`. 
@feelepxyz @jurre Can you guys review this PR and verify in this change makes sense to keep in dependabot-core?

### DIFF
--- a/common/lib/dependabot/pull_request_updater/azure.rb
+++ b/common/lib/dependabot/pull_request_updater/azure.rb
@@ -27,6 +27,21 @@ module Dependabot
         update_source_branch
       end
 
+      def update_pr_elements(elements = {})
+        return unless elements
+
+        azure_client_for_source.update_pull_request(
+          pull_request_id: pull_request_number,
+          status: elements[:status],
+          pr_name: elements[:pr_name],
+          description: elements[:description],
+          completion_options: elements[:completion_options],
+          merge_options: elements[:merge_options],
+          auto_complete_setby_user_id: elements[:auto_complete_setby_user_id],
+          target_branch: elements[:target_branch]
+        )
+      end
+
       private
 
       def azure_client_for_source
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -184,6 +184,28 @@ module Dependabot
           "/pullrequests?api-version=5.0", content.to_json)
       end
 
+      def update_pull_request(pull_request_id:, status: nil, pr_name: nil, description: nil,
+                              completion_options: nil, merge_options: nil,
+                              auto_complete_setby_user_id: nil, target_branch: nil)
+        content = {
+          autoCompleteSetBy: ({ id: auto_complete_setby_user_id } unless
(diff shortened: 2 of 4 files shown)


common/lib/dependabot/pull_request_updater/azure.rb
```diff
@@ -27,6 +27,21 @@ def update
         update_source_branch
       end
 
+      def update_pr_elements(elements = {})
```
jurre, line 30: What's the idea behind updating certain elements sometimes vs the entire PR other times?

A concern is that we now no longer follow the same API for different updaters.

I don't know if that's necessarily problematic, but up until now we've kept the interfaces the same between these different clients. At this point I think it's pretty unlikely that someone will agnostically open PRs against different providers, so deviating might be fine. @feelepxyz wdyt?

I've been thinking that it might even make sense to completely split off the Azure and GitLab clients from core and finding long-term maintainers for those, as we're probably not the best folks to maintain this code.

feelepxyz: @milind009 would be good to understand the use case for this and why the existing update method is not working for you? Slightly worried we're adding support for use-cases we don't understand and we end up with several slightly different ways to do the same thing.

Keen to figure if we can improve the existing update method instead or leave this as something that's done outside of dependabot if it's specific to your workflow.

feelepxyz: Hm just re-reading your PR description and sounds like we won't be able to easily fix the existing update method to do what you want. Looks like we only pass in `pr_name` in the creator for example and assuming you want to be able to update this?

I'm inclined to not merge this into core and encourage you to write a wrapper for this functionality in your dependabot setup as it seems specific to what you are doing.

milind009: @feelepxyz Currently the update function that we have for the azure client is used force push new changes on to the PR. However, we have a requirement to set a PR to autocomplete (azure create PR API does have an option to set PR to autocomplete but unfortunately it does not work, hence we need to create a PR and then set it to autocomplete)/abandon existing PR for which we need to use the azure [update PR API](https://docs.microsoft.com/en-us/rest/api/azure/devops/git/pull%20requests/update?view=azure-devops-rest-6.0#:~:text=URI%20Parameters,-URI%20Parameters&text=The%20name%20of%20the%20Azure%20DevOps%20organization.&text=ID%20of%20the%20pull%20request%20to%20update.&text=The%20repository%20ID%20of%20the%20pull%20request's%20target%20branch.&text=Version%20of%20the%20API%20to%20use.). So I created to separate method which can be used for updating other PR aesthetics like title, description,etc.

jurre: @milind009 is this something that could maybe live in the project that you use to run dependabot? It doesn't seem to rely on anything dependabot-specific?

milind009: Cool makes sense. I'll add this in our project. Thanks for the prompt replies!

PR #3491, opened 2021-04-14
npm/yarn: Always use registry source when available

jurre:
Previously there was some logic in place to determine if a package came
from a public package repo or not, and if it came from a registry
considered central/public, we would not parse its source.

However, in the case of private packages that specify yarnpkg.com as
their source, this resulted in some calls expecting that package to
exist on npmjs.org, which it technically is (yarnpkg.com is a proxy for
it), but we would have no credentials configured for npmjs.org, so those
requests will fail.

This PR changes that behavior to always parse a package's source and add
it to the requirement if it exists. We previously only did this for
registries we considered "private", hence the name `private_registry`,
but since we cannot reliably determine if a package hosted on
`yarnpkg.com` or `npmjs.org` is private or not, this has been renamed to
just `registry`.

This means we no longer fall back to `npmjs.org` for private packages
that go through `yarnpkg.com`, so any credentials that are configured
can be reliably used.

I considered always configuring creds for `npmjs.org` for `yarnpkg.com`
and vice versa, but it seemed a little gnarly, and I think this change
is more clear.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser.rb
@@ -252,10 +252,9 @@ module Dependabot
 
         return unless resolved_url
         return unless resolved_url.start_with?("http")
-        return if CENTRAL_REGISTRIES.any? { |u| resolved_url.start_with?(u) }
         return if resolved_url.match?(/(?<!pkg\.)github/)
 
-        private_registry_source_for(resolved_url, name)
+        registry_source_for(resolved_url, name)
       end
 
       def requirement_for(requirement)
@@ -287,7 +286,7 @@ module Dependabot
         }
       end
 
-      def private_registry_source_for(resolved_url, name)
+      def registry_source_for(resolved_url, name)
         url =
           if resolved_url.include?("/~/")
             # Gemfury format
@@ -305,7 +304,7 @@ module Dependabot
           else resolved_url.split("/")[0..2].join("/")
           end
 
-        { type: "private_registry", url: url }
+        { type: "registry", url: url }
       end
 
       def url_for_relevant_cred(resolved_url)
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/metadata_finder.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/metadata_finder.rb
@@ -39,7 +39,7 @@ module Dependabot
 
         case source_type
         when "git" then find_source_from_git_url
-        when "private_registry" then find_source_from_registry
+        when "registry" then find_source_from_registry
         else raise "Unexpected source type: #{source_type}"
         end
       end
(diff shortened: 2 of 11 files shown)


thepwagner (approved):
LGTM, appreciate the `{private_,}registry` cleanup!

npm_and_yarn/lib/dependabot/npm_and_yarn/metadata_finder.rb
```diff
@@ -39,7 +39,7 @@ def look_up_source
 
         case source_type
         when "git" then find_source_from_git_url
-        when "private_registry" then find_source_from_registry
+        when "registry" then find_source_from_registry
```
feelepxyz, line 42: I think we only run this during the update now while creating the pull request so shouldn't run in backend with old attribute 👌

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/registry_finder.rb
```diff
@@ -62,8 +62,8 @@ def first_registry_with_dependency_details
 
         def registry_url
           protocol =
-            if private_registry_source_url
-              private_registry_source_url.split("://").first
+            if registry_source_url
+              registry_source_url.split("://").first
```
feelepxyz, line 66: Nice, rename fits in perfectly 💯

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser.rb
```diff
@@ -252,10 +252,9 @@ def source_for(name, requirement, manifest_name)
 
         return unless resolved_url
         return unless resolved_url.start_with?("http")
-        return if CENTRAL_REGISTRIES.any? { |u| resolved_url.start_with?(u) }
         return if resolved_url.match?(/(?<!pkg\.)github/)
```
feelepxyz, line 255: I wonder why we return here? Seems ok to keep but maybe one to clean up at some point

feelepxyz (approved):

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser.rb
jurre: Yes, I was also wondering about that. I can dig around the history for a bit and maybe find out why it's there

jurre: So, this seems to be there to prevent a subdependency that's a git dependency from getting a `github.com` registry source:

https://github.com/dependabot/dependabot-core/commit/96980b412ba77290f0aef0402c66768e9ef36fe2

It explicitly excludes `pkg.github.com`, so I think it's safe (and necessary) to keep. There's test coverage for this so should be 👍

feelepxyz: Thanks for digging that out 👍

PR #3492, opened 2021-04-14
pyenv: shallow clone

thepwagner:
Continuing Dockerfile cleanup, let's do a shallow clone of `pyenv`.
That means instead of fetching the repository and checking out the branch we'd like to use, we'll fetch only the branch we'd like to use.

<details><summary>this only trims 4MB, but is trivial</summary>
<p>

Screen Shot 2021-04-14 at 8 34 47 AM

</p>
</details>


jurre (approved):
Nice

feelepxyz (approved):
I'll take any space savings 🙌✨

PR #3493, opened 2021-04-14
Add error handling for failure while updating Pull request in azure

milind009:
Added logic in the azure PR updater to raise an error of type `Dependabot::PullRequestUpdateFailed` when the update ref call to update the source branch is unsuccessful. Currently the update_ref API in azure returns a 200 response code even when the update operation is unsuccessful hence need to check the response body to verify if the update operation is successful
@feelepxyz @jurre Does this way of error handling makes sense? Apologies for creating a separate PR for this small change

### DIFF
--- a/common/lib/dependabot/errors.rb
+++ b/common/lib/dependabot/errors.rb
@@ -71,15 +71,6 @@ module Dependabot
     end
   end
 
-  class PullRequestUpdateFailed < DependabotError
-    attr_reader :pull_request_id
-
-    def initialize(pull_request_id, msg = nil)
-      @pull_request_id = pull_request_id
-      super(msg)
-    end
-  end
-
   #####################
   # File level errors #
   #####################
--- a/common/lib/dependabot/errors.rb
+++ b/common/lib/dependabot/errors.rb
@@ -71,7 +71,7 @@ module Dependabot
     end
   end
 
-  class PullRequestUpdatefailed < DependabotError
+  class PullRequestUpdateFailed < DependabotError
     attr_reader :pull_request_id
 
     def initialize(pull_request_id, msg = nil)
--- a/common/lib/dependabot/errors.rb
+++ b/common/lib/dependabot/errors.rb
@@ -71,6 +71,15 @@ module Dependabot
     end
   end
 
+  class PullRequestUpdatefailed < DependabotError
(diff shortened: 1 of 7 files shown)


common/lib/dependabot/pull_request_updater/azure.rb
```diff
@@ -55,9 +55,16 @@ def update_source_branch
...
         new_commit = create_temp_branch
         # 2) Update PR source branch to point to the temp branch head commit.
-        update_branch(source_branch_name, old_source_branch_commit, new_commit)
+        response = update_branch(source_branch_name, old_source_branch_commit, new_commit)
         # 3) Delete temp branch
         update_branch(temp_branch_name, new_commit, OBJECT_ID_FOR_BRANCH_DELETE)
+
+        unless response.fetch(
+          "success", false
+        )
```
jurre, line 64: Could we check the http status code instead?

milind009: Unfortunately yes, it returns 200 for the failure case too. The success parameter in the response body contains true or false to denote whether the update succeeded or not. I think this might due to the fact that the API we are using here is used to update multiple branches hence it returns 200 even when one of the branch update failed

jurre, line 64: Nitty, but this is easier to read on a single line IMO

```suggestion
        unless response.fetch("success", false)
```

jurre (approved):
I left an inline suggestion, but overall this seems like an improvement to me 👍

common/lib/dependabot/errors.rb
```diff
@@ -71,6 +71,15 @@ def initialize(source, msg = nil)
     end
   end
 
+  class PullRequestUpdateFailed < DependabotError
```
feelepxyz, line 74: Sorry for the slow review. Do we want to add this as a top-level error? Not sure we want to backport this to the other updaters, for github there are so many different reasons for a PR update failing so will be hard to reuse there. Maybe better to just add it in the Azure class for now?

milind009: Cool makes sense! I'll add this in azure pr updater class

milind009:
@jurre @feelepxyz can we merge this PR?

Issue #2879, opened 2020-12-16
Missing keyword using_bundler_2


zhaomengru2015:
When I  used `dependabot/dependabot-core` v0.126.1 to update ruby dependencies, I got `missing keyword: using_bundler_2` error. But it works well in v0.125.6.

jurre:
Could you give us a bit more context on how you're using dependabot-core, where you're seeing this error and what you've already tried to fix it so far?

zhaomengru2015:
we are using dependabot in Github Enterprise, since it didn't support Github Enterprise yet, we've implemented our own dependabot to update dependencies. It used the dependabot-core as the base image. recently we've upgraded the dependabot-core from 0.125.6 to 0.126.1 and all ruby dependencies were failed to update. So we temporarily downgraded our dependabot-core to 0.125.6 and the problem fixed

zhaomengru2015:
could it be caused by we were using bundler v1?

jurre:
No, bundler v1 should work fine. I would really need to see your code and a stacktrace of the error etc to be able to help here.

feelepxyz:
Are you running it inside the core docker container? This needs to be updated aswell.

feelepxyz:
Closing as stale

PR #3505, opened 2021-04-16
Remove unused bundler v2 codepath 

feelepxyz:
Removing the bundler 2 config changes that we prevously did in the v1
helpers. This codepath should no longer be hit as we invoke the v2
helpers if the lockfile uses bunder 2.

I've also changed the native helper args to take a hash making it less brittle when removing attributes and you run a different version of the gem to the docker image, which was reported in the past https://github.com/dependabot/dependabot-core/issues/2879

### DIFF
--- a/bundler/helpers/v1/lib/functions.rb
+++ b/bundler/helpers/v1/lib/functions.rb
@@ -7,29 +7,25 @@ require "functions/conflicting_dependency_resolver"
 
 module Functions
   def self.parsed_gemfile(lockfile_name:, gemfile_name:, dir:)
-    set_bundler_flags_and_credentials(dir: dir, credentials: [],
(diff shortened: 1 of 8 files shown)


jurre:
> making it less brittle when removing attributes and you run a different version of the gem to the docker image

Since we're calling `.fetch(:key)` on the args hash, it could still happen, right?

feelepxyz:
> Since we're calling `.fetch(:key)` on the args hash, it could still happen, right?

Yeah true for adding, but I think we probably want to explode if we don't have a required attribute. I think it should improve removing attributes as it just gets ignored if it's passed in, and this one is already not used in core since quite a few releases.

jurre (approved):

brrygrdn (approved):

PR #3510, opened 2021-04-19
Refactoring new_branch_name function in branch_namer

milind009:
Added a separate method name `branch_version_suffix` which basically gets the version related suffix for the branch name. We need to use this function in our own service hence created a separate method for this. @feelepxyz @jurre Does adding this refactoring in dependabot-core make sense?

### DIFF
--- a/common/lib/dependabot/pull_request_creator/branch_namer.rb
+++ b/common/lib/dependabot/pull_request_creator/branch_namer.rb
@@ -34,15 +34,7 @@ module Dependabot
                   tr("@", "")
               end
 
-            dep = dependencies.first
-
-            if library? && ref_changed?(dep) && new_ref(dep)
-              "#{dependency_name_part}-#{new_ref(dep)}"
-            elsif library?
-              "#{dependency_name_part}-#{sanitized_requirement(dep)}"
-            else
-              "#{dependency_name_part}-#{new_version(dep)}"
-            end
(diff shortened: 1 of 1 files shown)


jurre:
Yeah this makes sense to me, but currently that method is private, is that intended?

milind009:
> Yeah this makes sense to me, but currently that method is private, is that intended?

Yes i wanted to keep that as private

jurre:
> Yes i wanted to keep that as private

You mentioned you wanted to use it in your service, are you planning to use `send` to access it?

milind009:
> > Yes i wanted to keep that as private
> 
> You mentioned you wanted to use it in your service, are you planning to use `send` to access it?

We actually have a custom implementation of this branch_namer which inherits from this class i.e Dependabot::PullRequestCreator::BranchNamer so i would be using super() call in our custom implementation

jurre (approved):

jurre:
Trying to figure out what's up with the CI failures, they're not related to the changes here

milind009:
> Trying to figure out what's up with the CI failures, they're not related to the changes here

yeah cool. We are also getting these failures while downloading rubygems in our CI pipeline

jurre:
Ah, seems like they're having some issues: https://status.rubygems.org/ caused by https://dnsimple.statuspage.io/

milind009:
@jurre Can we merge this PR?

Issue #1957, opened 2020-06-21
Can't validate version 2 of configuration files


gnarea:
The [online validator](https://dependabot.com/docs/config-file/validator/) doesn't support version 2 files. Here's a file that Dependabot itself [created](https://github.com/relaycorp/relaynet-pong/pull/85):

![dependabot](https://user-images.githubusercontent.com/369863/85227191-cd2d7080-b3d3-11ea-8b90-0a1adb9741f8.png)

EmaSuriano:
@feelepxyz is it possible to add this link into the official [Github Documentation](https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#commit-message), thanks to this tool I found the typo I was having! I think it will help other people too 😄

GMartinez-Sisti:
> @gnarea we have one here: https://dependabot.com/docs/config-file-beta/validator/
> We also add a status check to the commit when you change the file in github validating the config file.

Hi @feelepxyz, the status check is not being executed with the MR. Can you please confirm, or let me know where I can configure this?

Thanks

asciimike:
Closing this, as there is a validator. I'm going to see about linking it from https://docs.github.com (we won't be able to embed it, but can link).

If there are other issues (e.g. status checks), please create a new issue with an MCVE and we can triage!

Issue #1927, opened 2020-06-20
Please publish full dependabot.yml schema


jongio:
When I was researching this #1926, I wan't to find the full schema for the dependabot.yml file to see if an option was available.  

I found docs here: 
https://dependabot.com/docs/config-file/

But that doesn't include all options, for example it doesn't include all the options available for "allowed_updates".

I found this schema:
https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/dependabot.json

But it is unclear if that is maintained by dependabot.

I would like to see a reference to the full dependabot.yml schema somewhere in the dependabot docs or source.  

Thanks!
Jon

azsdke2e

jeffwidman:
Related:
* https://github.com/dependabot/dependabot-core/issues/1919
* https://github.com/dependabot/dependabot-core/issues/4605

JamieMagee:
Linking the community provided JSON schema from [Schema Store][1]:

- https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/dependabot-2.0.json
- https://json.schemastore.org/dependabot-2.0.json

[1]: https://www.schemastore.org/json/

dmitris:
> [Schema Store](https://www.schemastore.org/json/)

This link (the first one in the comment) gives a 404 page now, but https://www.schemastore.org does work (and has a link to the [dependabot-v2.json](https://www.schemastore.org/dependabot-2.0.json) schema file).

PR #3512, opened 2021-04-19
dry-run: Dependabot::ConfigFile

thepwagner:
This introduces the `Dependabot::Config` module, starring the `Dependabot::Config::File` class.

This PR wraps the configuration stored in `dependabot.yaml`, with helpers to locate the configuration relevant to the active ecosystem/directory.

This is incomplete, only extracting the features to provide:
* use the `commit-message` options, if `--pull-request` is enabled
* use the `ignore:` conditions, if `IGNORE_CONDITIONS` is not set (following #3507)


feelepxyz (commented):
Nice! I also started on some of these primitives a while ago and wondering if we want to crib some stuff from there?

I tried to implement the config file fetcher as another file fetcher: https://github.com/dependabot/dependabot-core/compare/feelepxyz/cli-experiment?expand=1#diff-7f068f9118efc60b552a73767f52e71a60020c0690b239c292b4fc050d421782R8-R42

I also copied over the whole schema and a lot of the parsing logic which we probably don't want to get this over the line.

thepwagner:
> I tried to implement the config file fetcher as another file fetcher

SGTM, I'll make this change.

It's definitely a chicken/egg problem: the `FileFetcher` instance is informed by the `package_manager`, but the active `package_manager` s are contained the config file.

> copied over the whole schema and a lot of the parsing logic which we probably don't want to get this over the line

I wasn't keen to bring "real" validation, it seems harmless scope creep but would have the maintenance cost of requiring the schema be maintained in two places.
There's some demand, e.g. #1957 #1927  - I just didn't want to go there (yet).

common/lib/dependabot/config/update_config.rb
```diff
@@ -0,0 +1,49 @@
...
+        {
+          prefix: commit_message[:prefix],
+          prefix_development: commit_message[:"prefix-development"],
+          include_scope: commit_message[:include] == "scope"
+        }
+      end
+
+      def interval
+        return unless @config[:schedule]
+        return unless @config[:schedule][:interval]
```
feelepxyz, line 37: Should we raise here if it's not set? Or drop it until we need it? 🤷‍♂️

common/lib/dependabot/config/file_fetcher.rb
```diff
@@ -0,0 +1,52 @@
...
+      end
+
+      def self.required_files_message
+        "Repo must contain either a #{CONFIG_FILE_PATHS.join(' or a ')} file"
+      end
+
+      def config_file
+        @config_file ||= begin
+          cfg = files.first
+          Dependabot::Config::File.parse(cfg.content) if cfg
```
feelepxyz, line 21: Nit: I wonder if it would be clearer to do this parsing step after fetching the file in dry-run? Mostly to line up with existing core concepts.

common/lib/dependabot/config/file.rb
```diff
@@ -0,0 +1,48 @@
...
+
+      def initialize(updates:, registries: nil)
+        @updates = updates || []
+        @registries = registries || []
+      end
+
+      def update_config(package_manager, directory: nil)
+        dir = directory || "/"
+        package_ecosystem = PACKAGE_MANAGER_LOOKUP.invert.fetch(package_manager)
+        cfg = updates.find { |u| u[:"package-ecosystem"] == package_ecosystem && u[:directory] == dir }
```
feelepxyz, line 16: Thoughts on something like?

```suggestion
        cfg = updates.find { |u| u[:"package-ecosystem"] == package_ecosystem && u[:directory] == dir }
        raise InvalidConfigError, "no update config found for package-ecosystem #{package_ecosystem} and directory: #{dir}" unless cfg
```

feelepxyz (approved):
LGTM overall, left a few minor comments, feel free to merge when you've folded in what makes sense 👍

common/lib/dependabot/config/file.rb
thepwagner: I wasn't keen on that, and [explicitly test for the opposite](https://github.com/dependabot/dependabot-core/pull/3512/files#diff-3a8a97257f0436102272ecec2896ea6f42c8794accda9c7cc65cd2b07ed46d24R37).

My thinking is:
* I may want to test (e.g. `bin/dry-run.rb`) a repository that doesn't currently have Dependabot Version Updates enabled - that should be OK.
* In a hypothetical future where this infrastructure applies to Dependabot Security Updates, an `updates:` configuration may not exist.

I don't feel super strongly, but the idea was we should return default values instead of throwing an exception.

feelepxyz: Ah yes of course that's a great shout, didn't think of the security update case where you don't have a config file 👌

common/spec/dependabot/config/file_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+        expect(update_config.interval).to eq("monthly")
+      end
+
+      it "matches target-branch" do
+        update_config = config.update_config("npm_and_yarn", directory: "/target", target_branch: "the-awesome-branch")
+        expect(update_config).to be_a(Dependabot::Config::UpdateConfig)
+        expect(update_config.interval).to eq("daily")
+      end
+
+      it "returns empty when not found" do
```
feelepxyz, line 43: 👍

feelepxyz (approved):
🙌

PR #3513, opened 2021-04-19
experiment: generate ignored versions from "update-type"

thepwagner:
Continuing https://github.com/dependabot/dependabot-core/pull/3512 , this amends the `Dependabot::ConfigFile::UpdateConfig#ignored_versions_for` to dynamically generate a semver range based on the current version of the dependency.

This allows the (as of now, hypothetical) syntax:
```yaml
updates:
  - package-ecosystem: "npm"
    directory: "/"
    schedule:
      interval: "monthly"
    ignore:
      - dependency-name: "lodash"
        update-type: "patch"
```

This obviously won't work in dependabot proper, but since it reads the config file directly is a useful experiment 🔬 .


common/lib/dependabot/config.rb
```diff
@@ -54,10 +60,22 @@ def initialize(config)
       def ignored_versions_for(dep)
         return [] unless @config[:ignore]
 
-        @config[:ignore].
-          select { |ic| ic[:"dependency-name"] == dep.name }. # FIXME: wildcard support
-          map { |ic| ic[:versions] }.
-          flatten
+        ignore_configs = @config[:ignore].select { |ic| ic[:"dependency-name"] == dep.name } # FIXME: wildcard support
```
feelepxyz, line 63: I started adding a new interface to be able to do this into the update checker https://github.com/dependabot/dependabot-core/pull/3521 

Thinking we probably want to transform the attributes to be `snake_case` ruby symbols so we turn `:'dependency-name'` into `:dependency_name`.

thepwagner: > Thinking we probably want to transform the attributes to be snake_case ruby symbols so we turn :'dependency-name' into :dependency_name.

We're using [`symbolize_names`](https://github.com/dependabot/dependabot-core/blob/0352d543c2ed09ed24e42a72d6b0477b23a55dcf/common/lib/dependabot/config/file.rb#L46), is there a similar option to snake case the keys? (I could not find it).

thepwagner: We chatted about this and landed on https://github.com/dependabot/dependabot-core/pull/3525 ; I'm going to resolve this thread and force-push a branch that re-does this experiment against that model.

common/lib/dependabot/config/file.rb
```diff
@@ -60,7 +60,8 @@ def ignore_conditions(cfg)
         ignores.map do |ic|
           Dependabot::Config::UpdateConfig::IgnoreCondition.new(
             name: ic[:"dependency-name"],
-            versions: ic[:versions]
+            versions: ic[:versions],
+            update_type: ic[:"update-type"]
```
feelepxyz, line 64: I think this will end up being an array
```suggestion
            update_types: ic[:"update-types"]
```

common/spec/dependabot/config/ignore_condition_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+      Dependabot::Dependency.new(
+        name: dependency_name,
+        requirements: [],
+        package_manager: "npm_and_yarn",
+        version: "1.2.3"
+      )
+    end
+
+    def expect_allowed(*versions)
+      req = Gem::Requirement.new(ignored_versions.flat_map { |s| s.split(",").map(&:strip) })
```
feelepxyz, line 27: This should work but might be wrong?
```suggestion
      req = Gem::Requirement.new(*ignored_versions) })
```

feelepxyz: Oh wait it's `Gem::Requirement` so won't work 👍 

We could get the requirement class for the dependency?

common/lib/dependabot/config/ignore_condition.rb
```diff
@@ -0,0 +1,57 @@
...
+        when :ignore_minor_versions
+          [ignore_version(dep.version, 3)]
+        when :ignore_major_versions
+          [ignore_version(dep.version, 2)]
+        else
+          []
+        end
+      end
+
+      def ignore_version(version, precision)
```
feelepxyz, line 39: Do we want to crib some tests from find or create ignore condition in api? Looks like there's also some codes to add `.a`, assuming this is to include pre-releases. https://github.com/github/dependabot-api/blob/29bce7f4a03d42a67c5c528110b98cc7db9aec8b/app/actions/find_or_create_ignore_condition.rb#L55-L75

common/spec/dependabot/config/ignore_condition_spec.rb
thepwagner: I figured `Gem::Requirement` would be good enough for a test?

My goal was to write tests as "X should be ignored, Y should not be", as I found that easier to reason about than `"> a.b.c, < d.e.f"`.

common/lib/dependabot/config/ignore_condition.rb
thepwagner: That link 404s for our external friends :/
I think the equivalent `"a" for pre-release` is on L46, I'll include pre-release considerations in the test suite.

feelepxyz:
Ops thanks for flagging!

common/spec/dependabot/config/ignore_condition_spec.rb
thepwagner: I threw in a comment aiming to clarify these are limited tools for testing.

TBH they were only really necessary early in development when the ranges were blowing my mind - do you think we should just remove the tests?

common/lib/dependabot/config/ignore_condition.rb
```diff
@@ -0,0 +1,79 @@
...
+
+      def ignored_versions(dependency)
+        return [ALL_VERSIONS] if @versions.empty? && @update_types.empty?
+
+        versions_by_type(dependency) + @versions
+      end
+
+      private
+
+      def effective_update_type
```
thepwagner, line 30: Question: should `update_type` be a single value?

I assumed we don't need to care about every range, and just need to focus on the most restrictive (e.g. major > minor > patch).
It might make sense for the parser to select the "effective" `update_type`, and let `IgnoreCondition` deal with a single value. 🤷

common/lib/dependabot/config/update_config.rb
```diff
@@ -1,32 +1,25 @@
...
         @commit_message_options = commit_message_options
       end
 
-      def ignored_versions_for(dep)
+      def ignored_versions_for(dependency)
         @ignore_conditions.
-          select { |ic| ic.dependency_name == dep.name }. # FIXME: wildcard support
-          map(&:versions).
+          select { |ic| ic.dependency_name == dependency.name }. # FIXME: wildcard support
+          map { |ic| ic.ignored_versions(dependency) }.
```
thepwagner, line 18: Concern: as I as alluding to in https://github.com/dependabot/dependabot-core/pull/3536/files#r617772124 , this isn't going to behave when multiple conditions match the target dependency...

thepwagner (commented):
I marked RFR as this is code complete and roughly what I was going for.

As noted, the way conditions are aggregated needs some attention:
* What if there are multiple  matches with identical `update-types:`?
* What if there are multiple  matches with different `update-types:`?
* What if there are `update-types:` and `versions:`?

This definitely isn't shippable, but is a worthy iteration.

common/lib/dependabot/config/ignore_condition.rb
mctofu: I think we do need to care about each range? For the case of ignoring `version-update:semver-major` we were assuming the user still wants to get the minor/patch releases unless they also opt out of those types specifically. To be consistent in how it works the user could also specify something like `[version-update:semver-major, version-update:semer-patch]` to only get minor releases but not major/patch.

Sorry, I should have spotted this earlier!

common/spec/dependabot/config/ignore_condition_spec.rb
mctofu: > My goal was to write tests as "X should be ignored, Y should not be", as I found that easier to reason about than `"> a.b.c, < d.e.f"`.

👍 I agree this makes it easier to understand the tests. I think they are worth keeping!

common/spec/dependabot/config/ignore_condition_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+
+    def expect_ignored(*versions)
+      req = Gem::Requirement.new(ignored_versions.flat_map { |s| s.split(",").map(&:strip) })
+      versions.map do |v|
+        expect(req.satisfied_by?(Gem::Version.new(v))).
+          to eq(true), "Expected #{v} to be ignored, but was allowed"
+      end
+    end
+
+    context "without versions or update_types" do
```
mctofu, line 39: nice!

common/spec/dependabot/config/ignore_condition_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+
+        it "returns the expected range" do
+          expect(ignored_versions).to eq([">= 1.2.a, < 1.3"])
+        end
+
+        context "and a non-semver dependency" do
+          let(:dependency_version) { "Finchley.SR3" }
+
+          it "returns the expected range" do
+            expect(ignored_versions).to eq([">= Finchley.SR3.a, < Finchley.SR3.999999"])
```
mctofu, line 79: Should we return nothing in this case instead? I'm wondering if this would get rejected by an ecosystem as an invalid requirement.

thepwagner: 🤷 it's definitely a garbage-in-garbage-out case (which was the core reason for wanting to include it!).

`Finchley.SR3` as a version string to test was ported from the backend: e.g. if you `@dependabot ignore this major version` on a pull request for https://spring.io/blog/2019/02/26/spring-cloud-finchley-sr3-now-available , Dependabot will ignore `">= Finchley.a", "< Finchley.999999"`.
That makes sense, right? You don't want https://spring.io/blog/2019/06/14/spring-cloud-finchley-sr4-released .

> I'm wondering if this would get rejected by an ecosystem

I _think_ that as long as Dependabot's requirement can parse it, we're good:
```
irb(main):002:0> Dependabot::Maven::Requirement.new(">= Finchley.SR3.a, < Finchley.SR3.999999")
=> #<Dependabot::Maven::Requirement:0x000055b2411889d8 @requirements=[[">=", #<Dependabot::Maven::Version "Finchley.SR3.a">], ["<", #<Dependabot::Maven::Version "Finchley.SR3.999999">]]>
```

> Should we return nothing in this case instead

Yeah... let's!

common/lib/dependabot/config/update_config.rb
feelepxyz: Not sure I follow, do we have a test case for this that will fail currently?

I was looking at moving the dependency name checking into the IgnoreCondition class but would mean returning an empty `ignored_versions` if the dependency name doesn't match the `ic.dependency_name` which seems confusing to me but unsure if I'm missing something else?

common/spec/dependabot/config/ignore_condition_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+    end
+
+    def expect_ignored(versions)
+      reqs = ignored_versions.map { |v| Gem::Requirement.new(v.split(",").map(&:strip)) }
+      versions.each do |v|
+        version = Gem::Version.new(v)
+        ignored = reqs.any? { |req| req.satisfied_by?(version) }
+        expect(ignored).to eq(true), "Expected #{v} to be ignored, but was allowed"
+      end
+    end
```
feelepxyz, line 40: Nice test helpers 👍

common/lib/dependabot/config/update_config.rb
feelepxyz: > this isn't going to behave when multiple conditions match the target dependency

Is this case tested here?: https://github.com/dependabot/dependabot-core/pull/3513/files#diff-1bee1e376554d0d150b3a22f7b1ca87ac8d9de442cebf942c2f0e3f7fcc50c04R56

common/lib/dependabot/config/ignore_condition.rb
```diff
@@ -0,0 +1,102 @@
...
+        upper_bound = "< #{upper_parts.join('.')}"
+
+        ["#{lower_bound}, #{upper_bound}"]
+      end
+
+      def numeric_version?(version)
+        return false if version == ""
+
+        Gem::Version.correct?(version)
+      end
```
feelepxyz, line 99: Unused?
```suggestion
```

feelepxyz (commented):
This is looking good! Tests are real clear and easy to follow 💯 

I've been testing this on dry-runs and it's working well 🎉  Should we ship this and start incorporating in updater? Not sure if it's worth refactoring the wildcard matching as I see it possibly being more confusing having the dependency name filtering done in the individual IgnoreCondition but might have missed some edge-case that's not working.

feelepxyz:
> * What if there are multiple  matches with identical `update-types:`?
> * What if there are multiple  matches with different `update-types:`?
> * What if there are `update-types:` and `versions:`?

Tried to add some tests around this. Added a uniq on the generated version requirements but seems to work? 🤔

feelepxyz (approved):
I've been testing a bunch of dry-runs with different ignore setups and looking good so far! 🚢

PR #3514, opened 2021-04-19
dry-run: rubocop

thepwagner:
This is the result of `cd common && rubocop -A ../bin/dry-run.rb`, plus a few manual fixups:
* Annotation methods/blocks with `rubocop:enable` / `rubocop:disable` as needed
* Removing a duplicate rescue clause

I didn't go so far as to include this script in the CI process to prevent further regressions.

I just formatted this file in #3512 and the resulting diff was gnarly: let's consider that cleanup in isolation.

# Related
- Discovered https://github.com/dependabot/dependabot-core/pull/3512


feelepxyz (approved):
✨

PR #3520, opened 2021-04-20
Gomod: Run `go mod tidy` with flag to allow errors

jurre:
Go 1.16 added an option to instruct `go mod tidy` to make a best effort
to update the `go.sum` file in spite of errors for specific packages as
described in the commit the old comment pointed to:  https://github.com/golang/go/commit/3aa09489ab3aa13a3ac78b1ff012b148ffffe367

This adds said flag, now that we're on go 1.16

Edit: should we add a test that demonstates the new behavior?


thepwagner (approved):
LGTM, I'm not sure a test is worth it.

Thanks _fatih!

PR #3521, opened 2021-04-20
Add ignore conditions to update checker

feelepxyz:
Adding ignore conditions to the update checker to make it possible to
pass further types to core and filter relevant ones based on the
dependency name.

I've added this new attribute in a backwards compatible way. Once we're
passing it in we can drop `ignored_versions` from the interface.

### DIFF
--- a/common/lib/dependabot/update_checkers/base.rb
+++ b/common/lib/dependabot/update_checkers/base.rb
@@ -8,12 +8,11 @@ module Dependabot
   module UpdateCheckers
     class Base
       attr_reader :dependency, :dependency_files, :repo_contents_path,
-                  :credentials, :ignored_versions, :raise_on_ignored,
-                  :security_advisories, :requirements_update_strategy,
-                  :options
+                  :credentials, :raise_on_ignored, :security_advisories,
+                  :requirements_update_strategy, :options
 
       def initialize(dependency:, dependency_files:, repo_contents_path: nil,
-                     credentials:, ignored_versions: [],
+                     credentials:, ignore_conditions: [], ignored_versions: [],
                      raise_on_ignored: false, security_advisories: [],
                      requirements_update_strategy: nil,
                      options: {})
@@ -22,6 +21,7 @@ module Dependabot
         @repo_contents_path = repo_contents_path
         @credentials = credentials
         @requirements_update_strategy = requirements_update_strategy
+        @ignore_conditions = ignore_conditions
         @ignored_versions = ignored_versions
         @raise_on_ignored = raise_on_ignored
         @security_advisories = security_advisories
@@ -139,6 +139,23 @@ module Dependabot
 
         version = version_class.new(dependency.version)
         security_advisories.any? { |a| a.vulnerable?(version) }
+      end
+
+      def ignore_conditions
+        @ignore_conditions.select do |ic|
+          Utils.match_wildcard?(
+            name_normaliser.call(ic[:dependency_name]),
+            name_normaliser.call(dependency.name)
+          )
+        end
+      end
+
+      def ignored_versions
+        if ignore_conditions.any?
+          ignore_conditions.flat_map { |ic| ic[:versions] }
+        else
+          @ignored_versions
+        end
       end
 
       private
@@ -295,6 +312,10 @@ module Dependabot
         return false if changed_requirements.none?
 
         changed_requirements.none? { |r| r[:requirement] == :unfixable }
+      end
+
+      def name_normaliser
+        Dependency.name_normaliser_for_package_manager(dependency.package_manager)
       end
 
       def ignore_reqs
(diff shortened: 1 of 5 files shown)


common/lib/dependabot/update_checkers/base.rb
```diff
@@ -8,12 +8,11 @@ module Dependabot
...
       attr_reader :dependency, :dependency_files, :repo_contents_path,
-                  :credentials, :ignored_versions, :raise_on_ignored,
-                  :security_advisories, :requirements_update_strategy,
-                  :options
+                  :credentials, :raise_on_ignored, :security_advisories,
+                  :requirements_update_strategy, :options
 
       def initialize(dependency:, dependency_files:, repo_contents_path: nil,
-                     credentials:, ignored_versions: [],
+                     credentials:, ignore_conditions: [], ignored_versions: [],
```
thepwagner, line 15: Concern: accepting both `ignore_conditions` and `ignored_versions` here is a weird smell.  Should `ignored_versions` be considered deprecated?

Since `ignore_conditions` is an array of items that is filtered by `dependency`, I don't get the value of including the entire list (including entries not relevant to `dependency` of interest).

common/lib/dependabot/update_checkers/base.rb
```diff
@@ -141,6 +141,23 @@ def vulnerable?
         security_advisories.any? { |a| a.vulnerable?(version) }
       end
 
+      def ignore_conditions
+        @ignore_conditions.select do |ic|
```
thepwagner, line 145: This seems eerily similar to https://github.com/dependabot/dependabot-core/blob/0352d543c2ed09ed24e42a72d6b0477b23a55dcf/common/lib/dependabot/config/update_config.rb#L21 , but has wildcard support!

thepwagner (commented):
I'm concerned that accepting `ignore_conditions` and `ignore_versions` leaves an API with two ways to accomplish the same thing. Can you expand on what the updater will do with `ignore_conditions` that it can't do with `ignore_versions`?

If there isn't a use case, what do you think about keeping `Dependabot::Config::UpdateConfig` as the translation layer between the conditions->versions?

feelepxyz:
Yeah agreed, I think it would be a lot easier to reason about this if we did this in the `UpdateConfig` class, will close this once we've copied over the wildcard things into https://github.com/dependabot/dependabot-core/pull/3513

feelepxyz:
Ported wildcard filtering: https://github.com/dependabot/dependabot-core/pull/3536

craigta:
@feelepxyz  wondering why this one was closed? I was using Dependabot-Script and wanted to pass IgnoreConditions to the UpdateChecker, and it looks like this change was on the path to do that. Right now, UpdateChecker supports IgnoreVersions, but I'd like to skip major/minor -- did I miss that there is another way to do that via Dependabot-Script? I took a copy of the update script so happy to modify that to skip major/minor, just trying to figure out best way to do so.. open to ideas!!

feelepxyz:
@craigta yes this should be possible. If you have something like this in your setup:

```ruby
#########################################
# Get update details for the dependency #
#########################################
checker = Dependabot::UpdateCheckers.for_package_manager(package_manager).new(
  dependency: dep,
  dependency_files: files,
  credentials: credentials,
)
```

Try changing it to:

```ruby
ignore_conditions = [{
  "dependency-name" => "dependency-name*", # glob wildcards can be applied to the dependency name
  "update-types" => ["version-update:semver-minor"] # supports: version-update:semver-patch, version-update:semver-minor, version-update:semver-major
}]

def ignore_conditions_for(dep)
  ignore_conditions = ignore_conditions.map do |ic|
    Dependabot::Config::IgnoreCondition.new(
      dependency_name: ic["dependency-name"],
      versions: [ic["version-requirement"]].compact,
      update_types: ic["update-types"]
    )
  end
  Dependabot::Config::UpdateConfig.
    new(ignore_conditions: ignore_conditions).
    ignored_versions_for(dep)
end

checker = Dependabot::UpdateCheckers.for_package_manager(package_manager).new(
  dependency: dep,
  dependency_files: files,
  ignored_versions: ignore_conditions_for(dep),
  credentials: credentials,
)

```

craigta:
Amazing! Thank you that did the trick! Although I just ended up putting the ignore condition directly into the "ignore_conditions_for", rather than defining it outside as the key value pair

def ignore_conditions_for(dep)
  ignore_conditions = [
    Dependabot::Config::IgnoreCondition.new(
      dependency_name: "*",
      update_types: ["version-update:semver-major", "version-update:semver-minor"]
    )
  ]

  Dependabot::Config::UpdateConfig.
    new(ignore_conditions: ignore_conditions).
    ignored_versions_for(dep)
end


Now to figure out why just my packages.config got updated, and not my project files... but it did stick to the patch version for the packages.config, which is what I want to start with

PR #3525, opened 2021-04-20
UpdateConfig ruby-er model

thepwagner:
This shifts the `Dependabot::Config::UpdateConfig` class introduced by https://github.com/dependabot/dependabot-core/pull/3512 from being a shallow facade over an entry in the `updates:` configuration, to a plain old ruby object.

Parsing is migrated to the `Dependabot::Config::File` class, which makes sense as the owner of parsing+translation.

I also removed `Dependabot::Config::UpdateConfig#interval`. As discussed in the earlier PR, it only existed to provide a unique attribute on `.updates[]` parsed from the config file. By using the `#commit_message_options` this can be accomplished with less cruft.

# Related
- Continues https://github.com/dependabot/dependabot-core/pull/3512
- Discussed https://github.com/dependabot/dependabot-core/pull/3521
- Conflicts https://github.com/dependabot/dependabot-core/pull/3513

### DIFF
--- a/common/lib/dependabot/config/file.rb
+++ b/common/lib/dependabot/config/file.rb
@@ -20,7 +20,10 @@ module Dependabot
           u[:"package-ecosystem"] == package_ecosystem && u[:directory] == dir &&
             (target_branch.nil? || u[:"target-branch"] == target_branch)
         end
-        Dependabot::Config::UpdateConfig.new(cfg)
+        Dependabot::Config::UpdateConfig.new(
+          cfg,
+          commit_message_options: commit_message_options(cfg)
+        )
       end
 
       PACKAGE_MANAGER_LOOKUP = {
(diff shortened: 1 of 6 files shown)


common/spec/fixtures/configfile/npm-weekly.yml
```diff
@@ -1,15 +1,22 @@
+# NOTE: prefix: values are used by file_spec.rb
```
thepwagner, line 1:
```suggestion
# NOTE: prefix: values are used by common/spec/dependabot/config/update_config_spec.rb
```

Full path and real path.

common/lib/dependabot/config/file.rb
```diff
@@ -49,6 +52,27 @@ def self.parse(config)
...
         File.new(updates: parsed[:updates], registries: parsed[:registries])
       end
+
+      private
+
+      def ignore_conditions(cfg)
+        ignores = cfg&.dig(:ignore) || []
+        ignores.map do |ic|
+          Dependabot::Config::UpdateConfig::IgnoreCondition.new(
+            name: ic[:"dependency-name"],
```
feelepxyz, line 62: Nit but wondering if `dependency_name` would be clearer once we add other things like update-types, dependency-types
```suggestion
            dependency_name: ic[:"dependency-name"],
```

feelepxyz (approved):
Added a minor comment about renaming `name` to `dependency_name`, otherwise LGTM 👍

PR #3528, opened 2021-04-21
Fix typo to clarify meaning

jeffwidman:
There's clearly a typo here, but I'm less clear on what the intended wording was:
* we either use the commit sha or the tag name
* we use the commit sha of the tag name

These have very different meanings... I picked one, but I may be wrong, as I've never written Ruby, so kinda spitballin' here.

Let me know if I should reword different.

### DIFF
--- a/go_modules/lib/dependabot/go_modules/update_checker.rb
+++ b/go_modules/lib/dependabot/go_modules/update_checker.rb
@@ -121,7 +121,7 @@ module Dependabot
       end
 
       def version_from_tag(tag)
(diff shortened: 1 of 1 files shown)


jurre (commented):
Thanks! Yeah, That `of` should be an or, either the commit sha _or_ the tag name. 

The method returns early it the commit sha is found, else it returns the tag name

jeffwidman:
That makes more sense. I just updated the PR via the GitHub web UI, so please merge via squashing.

jurre (approved):
Thanks!

PR #3531, opened 2021-04-21
Restrict permissions for actions workflows

feelepxyz:
Following this change: https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/

We only need to clone repos in `ci.yml` tests so `contents: read` should
be enough.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -5,7 +5,8 @@ on:
       - main
     tags:
       - v[0-9]+.[0-9]+.[0-9]+
-permissions: none
+permissions:
+  contents: read
 jobs:
   push-core-image:
     name: Push dependabot-core image to docker hub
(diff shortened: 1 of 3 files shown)


.github/workflows/docker.yml
```diff
@@ -5,7 +5,7 @@ on:
       - main
     tags:
       - v[0-9]+.[0-9]+.[0-9]+
-
+permissions: none
```
thepwagner, line 8: Question: how does the `actions/checkout@v2` on L15 work without `contents: read`? I expected that to be via the automatically generated `GITHUB_TOKEN`.

thepwagner (approved):
`ci.yml` LGTM.
I'm concerned the other workflows may still need `contents:read` as well, but since `docker.yml` triggers `on: push: main`, we'll find out soon enough 🔬 .

.github/workflows/docker.yml
feelepxyz: Ah yes, I think you're right, will set `contents:read`

PR #3533, opened 2021-04-21
Fix regressions from ignore version refactor

feelepxyz:
Fixes a few bugs discovered from https://github.com/dependabot/dependabot-core/pull/3368

For reasons I don't yet understand, the bundler latest version finder doesn't declare `requirement_class` but passes specs testing ignored versions but if I `byebug` inside the method during the test it explodes with an exception saying it's undefined.

I also discovered we didn't properly handle for the case where a maven/gradle version requirement is created from a pull request ignore with ruby style requirements but with a maven style version, e.g. `>= Finchley.a, < Finchley.999999`.

### DIFF
--- a/gradle/lib/dependabot/gradle/requirement.rb
+++ b/gradle/lib/dependabot/gradle/requirement.rb
@@ -33,14 +33,7 @@ module Dependabot
 
       def initialize(*requirements)
         requirements = requirements.flatten.flat_map do |req_string|
-          # NOTE: Support ruby-style version requirements that are created from
-          # PR ignore conditions
-          version_reqs = req_string.split(",").map(&:strip)
-          if version_reqs.all? { |s| Gem::Requirement::PATTERN.match?(s) }
-            version_reqs
-          else
-            convert_java_constraint_to_ruby_constraint(req_string)
-          end
+          convert_java_constraint_to_ruby_constraint(req_string)
         end
 
         super(requirements)
@@ -72,9 +65,14 @@ module Dependabot
           raise "Can't convert multiple Java reqs to a single Ruby one"
         end
 
-        return convert_java_range_to_ruby_range(req_string) if req_string&.include?(",")
-
-        convert_java_equals_req_to_ruby(req_string)
+        # NOTE: Support ruby-style version requirements that are created from
+        # PR ignore conditions
+        version_reqs = req_string.split(",").map(&:strip)
+        if req_string.include?(",") && !version_reqs.all? { |s| PATTERN.match?(s) }
+          convert_java_range_to_ruby_range(req_string) if req_string.include?(",")
+        else
+          version_reqs.map { |r| convert_java_equals_req_to_ruby(r) }
+        end
       end
 
       def convert_java_range_to_ruby_range(req_string)
--- a/gradle/lib/dependabot/gradle/version.rb
+++ b/gradle/lib/dependabot/gradle/version.rb
@@ -28,7 +28,7 @@ module Dependabot
       VERSION_PATTERN =
         "[0-9a-zA-Z]+"\
         '(?>\.[0-9a-zA-Z]*)*'\
-        '([_-][0-9A-Za-z_-]*(\.[0-9A-Za-z_-]*)*)?'
+        '([_\-\+][0-9A-Za-z_-]*(\.[0-9A-Za-z_-]*)*)?'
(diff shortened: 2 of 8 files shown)


gradle/lib/dependabot/gradle/requirement.rb
```diff
@@ -36,7 +36,7 @@ def initialize(*requirements)
           # NOTE: Support ruby-style version requirements that are created from
           # PR ignore conditions
           version_reqs = req_string.split(",").map(&:strip)
-          if version_reqs.all? { |s| Gem::Requirement::PATTERN.match?(s) }
+          if req_string.include?(",") && version_reqs.all? { |s| PATTERN.match?(s) }
```
feelepxyz, line 39: I missed that `PATTERN` includes the Gem::Requirement::OPS` and the `PATTERN` from `Gradle::Version` so does exactly what we want here.

gradle/lib/dependabot/gradle/version.rb
```diff
@@ -28,7 +28,7 @@ class Version < Gem::Version
       VERSION_PATTERN =
         "[0-9a-zA-Z]+"\
         '(?>\.[0-9a-zA-Z]*)*'\
-        '([_-][0-9A-Za-z_-]*(\.[0-9A-Za-z_-]*)*)?'
+        '([_\-\+][0-9A-Za-z_-]*(\.[0-9A-Za-z_-]*)*)?'
```
feelepxyz, line 31: This looks like an omission, brings this pattern to par with `maven`. This is currently tested in the requirement class but without this change it's not `correct?` so you can't initialise it as a version: https://github.com/dependabot/dependabot-core/blob/1ba226ad7fb22f22fa3eef73843e6f6c956573f5/gradle/spec/dependabot/gradle/requirement_spec.rb#L55

maven/spec/dependabot/maven/requirement_spec.rb
```diff
@@ -74,6 +74,21 @@
...
+
+      context "that specifies a minimum" do
+        let(:requirement_string) { "1.5+" }
+        its(:to_s) { is_expected.to eq(Gem::Requirement.new("~> 1.5").to_s) }
+      end
+
+      context "that is just a +" do
+        let(:requirement_string) { "+" }
+        its(:to_s) { is_expected.to eq(Gem::Requirement.new(">= 0").to_s) }
+      end
```
feelepxyz, line 86: Copied these from the gradle requirement specs https://github.com/dependabot/dependabot-core/blob/main/gradle/spec/dependabot/gradle/requirement_spec.rb#L54-L67

gradle/spec/dependabot/gradle/requirement_spec.rb
```diff
@@ -64,6 +64,11 @@
         let(:requirement_string) { "+" }
         its(:to_s) { is_expected.to eq(Gem::Requirement.new(">= 0").to_s) }
       end
+
+      context "with a comman-separated dynamic version requirements" do
```
thepwagner, line 68:
```suggestion
      context "with a comma-separated dynamic version requirements" do
```

Nit: typo.

maven/lib/dependabot/maven/requirement.rb
```diff
@@ -105,10 +103,11 @@ def convert_java_equals_req_to_ruby(req_string)
...
 
       def convert_wildcard_req(req_string)
-        version = req_string.gsub(/(?:\.|^)\+/, "")
-        return ">= 0" if version.empty?
+        version = req_string.split("+").first
+        return ">= 0" if version.nil? || version.empty?
 
-        "~> #{version}.0"
+        version += "0" if version.end_with?(".")
+        "~> #{version}"
```
feelepxyz, line 110: Copied implementation from gradle https://github.com/dependabot/dependabot-core/blob/main/gradle/lib/dependabot/gradle/requirement.rb#L107-L113

maven/spec/dependabot/maven/requirement_spec.rb
```diff
@@ -74,6 +74,21 @@
...
+
+      context "with a comman-separated dynamic version requirements" do
```
thepwagner, line 88:
```suggestion
      context "with a comma-separated dynamic version requirements" do
```

thepwagner (approved):
LGTM, just getting picky about typos in specs.

PR #3536, opened 2021-04-21
UpdateConfig: Match wildcards in dependency names

feelepxyz:
Add wildcard dependency name filtering when finding ignore conditions
for a given dependency.

### DIFF
--- a/common/lib/dependabot/config/update_config.rb
+++ b/common/lib/dependabot/config/update_config.rb
@@ -13,10 +13,32 @@ module Dependabot
 
       def ignored_versions_for(dep)
         @ignore_conditions.
-          select { |ic| ic.dependency_name == dep.name }. # FIXME: wildcard support
+          select do |ic|
+            self.class.wildcard_match?(
+              name_normaliser_for(dep).call(ic.dependency_name),
+              name_normaliser_for(dep).call(dep.name)
+            )
+          end.
           map(&:versions).
           flatten.
           compact
+      end
+
+      def self.wildcard_match?(wildcard_string, candidate_string)
+        return false unless wildcard_string && candidate_string
+
+        regex_string = "a#{wildcard_string.downcase}a".split("*").
+                       map { |p| Regexp.quote(p) }.
+                       join(".*").gsub(/^a|a$/, "")
+        regex = /^#{regex_string}$/
+        regex.match?(candidate_string.downcase)
+      end
+
+      private
+
+      def name_normaliser_for(dep)
+        name_normaliser ||= {}
+        name_normaliser[dep] ||= Dependency.name_normaliser_for_package_manager(dep.package_manager)
       end
 
       class IgnoreCondition
--- a/common/spec/dependabot/config/update_config_spec.rb
+++ b/common/spec/dependabot/config/update_config_spec.rb
@@ -17,7 +17,7 @@ RSpec.describe Dependabot::Config::UpdateConfig do
       )
     end
     let(:ignore_conditions) { [] }
-    let(:config) { Dependabot::Config::UpdateConfig.new(ignore_conditions: ignore_conditions) }
+    let(:config) { described_class.new(ignore_conditions: ignore_conditions) }
 
     it "returns empty when not defined" do
       expect(ignored_versions).to eq([])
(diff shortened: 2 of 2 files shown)


common/lib/dependabot/config/update_config.rb
```diff
@@ -12,13 +12,18 @@ def initialize(ignore_conditions: nil, commit_message_options: nil)
       end
 
       def ignored_versions_for(dep)
+        name_normaliser = Dependency.name_normaliser_for_package_manager(dep.package_manager)
         @ignore_conditions.
-          select { |ic| ic.dependency_name == dep.name }. # FIXME: wildcard support
+          select { |ic| name_match?(name_normaliser.call(ic.dependency_name), name_normaliser.call(dep.name)) }.
```
thepwagner, line 17: Idea: `name_normaliser.call(dep.name)` will always have the same value, we could save a few nanos by memoizing outside of the loop.

thepwagner: Idea: should the `IgnoreCondition` be what decides if the ignore condition matches this dependency?
Something like `IgnoreCondition#applies_to?(dependency)` or `IgnoreCondition#applies_to?(dependency_name)` 

Meh, just brainstorming.

common/lib/dependabot/utils.rb
```diff
@@ -44,5 +44,15 @@ def self.always_clone_for_package_manager?(package_manager)
     def self.register_always_clone(package_manager)
       @cloning_package_managers << package_manager
     end
+
+    def self.wildcard_match?(wildcard_string, candidate_string)
```
thepwagner, line 48: Question: what else is this used by/do we anticipate using this?

Could we reduce its visibility? Maybe it only needs to exist in `IgnoreCondition`

common/spec/dependabot/utils_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+    context "with a wildcard" do
+      context "at the start" do
+        let(:wildcard_string) { "*bus" }
+
+        context "with a matching string" do
+          let(:candidate_string) { wildcard_string }
+          it { is_expected.to eq(true) }
+        end
+
+        context "with a matching string (except the wildcard" do
```
thepwagner, line 62:
```suggestion
        context "with a matching string (except the wildcard)" do
```

Nit: balanced parens.

common/spec/dependabot/utils_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+
+      context "at the end" do
+        let(:wildcard_string) { "bus*" }
+
+        context "with a matching string" do
+          let(:candidate_string) { wildcard_string }
+          it { is_expected.to eq(true) }
+        end
+
+        context "with a matching string (except the wildcard" do
```
thepwagner, line 91:
```suggestion
        context "with a matching string (except the wildcard)" do
```

common/spec/dependabot/utils_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+
+      context "in the middle" do
+        let(:wildcard_string) { "bu*s" }
+
+        context "with a matching string" do
+          let(:candidate_string) { wildcard_string }
+          it { is_expected.to eq(true) }
+        end
+
+        context "with a matching string (except the wildcard" do
```
thepwagner, line 120:
```suggestion
        context "with a matching string (except the wildcard)" do
```

common/lib/dependabot/utils.rb
```diff
@@ -44,5 +44,15 @@ def self.always_clone_for_package_manager?(package_manager)
...
+      return false unless wildcard_string && candidate_string
+
+      regex_string = "a#{wildcard_string.downcase}a".split("*").
```
thepwagner, line 51: Question: what's up with these `a` characters? Is there a test case that fails without them?

thepwagner (approved):
LGTM, just some brainstorming about separating these concerns.

I'm biased by the work in https://github.com/dependabot/dependabot-core/pull/3513 , which is striving to make `Dependabot::Config::IgnoreCondition` "a thing". :shrug:.

common/spec/dependabot/config/update_config_spec.rb
```diff
@@ -33,6 +33,17 @@
         expect(ignored_versions).to eq([">= 14.14.x, < 15"])
       end
     end
+
+    context "with a wildcard dependency name" do
```
thepwagner, line 37: Idea: I'd dig a test where the target dependency matches multiple conditions.
I think the expectation is we'll `flat_map` them out, and I think it will work... it just wasn't possible before, but is now.

The behaviour of multiple matches gets more interesting in https://github.com/dependabot/dependabot-core/pull/3513 , but that branch doesn't have wildcards yet! 🐔🥚

common/lib/dependabot/utils.rb
feelepxyz: Yeah looks like there's a few test cases for this fail, e.g. `wildcard_match?("*", "bus")`, looks like we just add a random character so we end up with multiple parts when splitting by `*` so we can join `.*` on the parts

common/spec/dependabot/config/update_config_spec.rb
feelepxyz: Added 👍

common/lib/dependabot/utils.rb
feelepxyz: Good shout, folded into UpdateConfig for now, thought it might be easier to refactor into IgnoreCondition once we've merged it into your branch as I don't want to create too many merge conflicts

common/lib/dependabot/config/update_config.rb
feelepxyz: Noted bellow, going to merge it into UpdateConfig and then apply to your branch and refactor from there

Issue #3529, opened 2021-04-21
[go] Drop support for `dep`, as everyone is now moving to `go mod`


jeffwidman:
I noticed several references in the source to `dep` which was a package manager that was a precursor to `go mod`. 

Now that `go mod` is finalized and built into the standard lib, the ecosystem is migrating to that, so it's probably worth dropping support for `dep` completely. 

That would simplify the code for both maintainers and drive-by readers (such as myself) since I could safely assume that the code works according to expected `go mod` paradigm, rather than being afraid there are edge cases I don't realize that may be causing subtle bugs when `dep` behavior is accidentally mis-applied to `go mod`... 

Examples of `dep` references:
* https://github.com/dependabot/dependabot-core/blob/760b69a68962c4692d5d7dd8fef8fa36e6f2d9f8/go_modules/lib/dependabot/go_modules/requirement.rb#L135
* https://github.com/dependabot/dependabot-core/blob/760b69a68962c4692d5d7dd8fef8fa36e6f2d9f8/go_modules/lib/dependabot/go_modules/update_checker.rb#L113

jurre:
`dep` is already deprecated in dependabot already, it's not supported in the GH native version of Dependabot, and it's marked as deprecated in dependabot-preview. We're planning to remove the codepath entirely, I think the references in the `go_modules` implementation are leftovers from when in was ported from the `dep` code, there are two top level folders: `go_modules` and `dep` which each house the implementation for that package manager.

Those references should probably be cleaned up though, thanks for pointing them out.

jeffwidman:
I put up https://github.com/dependabot/dependabot-core/pull/3537 to remove the outdated references. 

I hadn't even seen the separate `dep` folder... yeah, makes sense to just clear that out completely.

jeffwidman:
I also put up https://github.com/dependabot/dependabot-core/pull/3538 to remove the `dep` folder... no need to keep maintaining it IMO. Anyone still stuck with `go dep` is likely also going to be stuck with an older version of dependabot so they are unlikely to be impacted...

AraHaan:
Unless they use the github native dependabot.

jeffwidman:
>> Anyone still stuck with go dep is likely also going to be stuck with an older version of dependabot so they are unlikely to be impacted...
> Unless they use the github native dependabot.

Except that [github-native dependabot doesn't support `dep`](https://github.com/dependabot/dependabot-core/issues/3529#issuecomment-823837664).

[...]

PR #3537, opened 2021-04-21
Remove dep references from go modules code

jeffwidman:
This was likely leftover from when the `go dep` code was ported to `go
mod` support.

See also https://github.com/dependabot/dependabot-core/issues/3529

This only removes the references in `go mod`, it doesn't remove `go dep`
support.


go_modules/lib/dependabot/go_modules/update_checker.rb
```diff
@@ -112,6 +112,8 @@ def updated_dependencies_after_full_unlock
       # Override the base class's check for whether this is a git dependency,
       # since not all dep git dependencies have a SHA version (sometimes their
       # version is the tag)
+      #
+      # TODO should this be removed entirely? I'm not sure what it referred to originally, and this functionality doesn't ring a bell for `go mod`...
```
jeffwidman, line 116: I never used `go dep`, so unclear what the original thing this was trying to guard against... as a result, I'm unclear if this is still applicable to `go mod`... @jurre any idea?

go_modules/lib/dependabot/go_modules/requirement.rb
```diff
@@ -132,13 +130,6 @@ def ruby_range(req_string)
         "~> #{parts.join('.')}"
       end
 
-      # NOTE: Dep's caret notation implementation doesn't distinguish between
```
jeffwidman, line 135: I removed this entire function because as best I can tell `go mod` has no special treatment for `^`...

go_modules/lib/dependabot/go_modules/update_checker.rb
jurre: In `go mod` versions can also be a tag AFAIK so we'll want to keep this code in

go_modules/lib/dependabot/go_modules/requirement.rb
jurre: I think you're right, but we'll also need to remove the calling code and the tests in `go_modules/spec/dependabot/go_modules/requirement_spec.rb` that fail because of this removal

go_modules/lib/dependabot/go_modules/update_checker.rb
jeffwidman: ah, I see what it's getting at... updated the comment

go_modules/lib/dependabot/go_modules/requirement.rb
jeffwidman: So I started looking at this, and it looks like there's a _lot_ of tests in `go_modules/spec/dependabot/go_modules/requirement_spec.rb` that are simply inaccurate... Since `go mod` uses [MVS](https://golang.org/ref/mod#minimal-version-selection), most of the more traditional requirements ideas like ranges, wildcards, etc simply doesn't apply (IIUC).

So should the following be removed from the tests?
* `!`
* `^`
* `>`
* `*`
* `||`
* `,` (comma, separated list of versions)

Any of the above should result in a parsing error IMO, as they are not valid for `go.mod` syntax. 

Also, I'm confused why it looks like this is re-duplicating in ruby the `go mod` syntax checker? Why not just run `go mod tidy` and if it complains about syntax, then bomb out? 

Given the above, should all these helpers be removed?
* https://github.com/dependabot/dependabot-core/blob/80acab859af6322f6f160c1d30bc7e1777836a5e/go_modules/lib/dependabot/go_modules/requirement.rb#L63
* https://github.com/dependabot/dependabot-core/blob/80acab859af6322f6f160c1d30bc7e1777836a5e/go_modules/lib/dependabot/go_modules/requirement.rb#L67-L69
* https://github.com/dependabot/dependabot-core/blob/80acab859af6322f6f160c1d30bc7e1777836a5e/go_modules/lib/dependabot/go_modules/requirement.rb#L75-L119

I'm sure there's some stuff I'm ignorant of, so if there's a good reason to leave this all in, please educate me.

jurre: Yeah, I noticed it as well, tbh I think you might be right and these all stem from the original `dep` implementation and are redundant for `go mod`, but just haven't bitten us because the regexes simply never match in real world scenario's.

cc @hmarr, it looks like you were the last person to really touch this code, could you confirm that we can nuke all those conversions?

jeffwidman: Nudge @hmarr any feedback? 

I'd like to push this across the line.

hmarr: It [looks like](https://github.com/dependabot/dependabot-core/commit/a23c650564ce168128f36acd6cf990cf97bbcf66) I did indeed just copy this file directly across from the dep implementation, which means it won't be correct for Go modules.

I'm definitely 👍  on making this code reflect how Go modules actually work. We might need to double check this won't break ignore conditions (which you can add to a config file, and also get generated when you say e.g. `@dependabot ignore this minor version`).

jeffwidman: > We might need to double check this won't break ignore conditions (which you can add to a config file, and also get generated when you say e.g. @dependabot ignore this minor version).

@jurre do you know if there are tests already covering this? I wasn't sure if there was a higher-level test spec that contains generic scenarios that all package managers need to satisfy, or if it needs to be done on a per-implementation level...

jeffwidman: nudge @jurre ☝️

PR #3538, opened 2021-04-21
Remove go dep entirely

jeffwidman:
This has been deprecated for a while, both here in dependabot and also
in the `go` ecosystem. The ecosystem has moved to `go mod` (which we
already support).

So remove the `go dep` code entirely. No need to keep supporting it.

Anyone who wants this old of code is likely not keeping up with their
dependencies anyway, so probably will just keep using an old version of
`dependabot`.

See also https://github.com/dependabot/dependabot-core/issues/3529.


Dockerfile
```diff
@@ -164,20 +164,15 @@ USER root
 
 ### GO
 
-# Install Go and dep
+# Install Go
 ARG GOLANG_VERSION=1.16.2
 ARG GOLANG_CHECKSUM=542e936b19542e62679766194364f45141fde55169db2d8d01046555ca9eb4b8
-ENV PATH=/opt/go/bin:$PATH \
-  GOPATH=/opt/go/gopath
```
jeffwidman, line 171: This one I wasn't entirely sure of... there were other references to `gopath` folder, but from a quick skim they looked unrelated to this particular folder... it's tricky because `gopath` is also a convenient name to use for a generic folder of where `go` stuff is.

At a higher level the `GOPATH` concept is going away completely in the upcoming `go 1.17`: https://blog.golang.org/go116-module-changes so it'd be best to make it so dependabot works w/o needing `GOPATH` (as traditionally used by `go`) to be set.

I think we should just see what CI says, and if CI greenlights this, then we are probably good to go (assuming CI uses this Dockerfile).

jeffwidman:
I think I got all the references, but not sure... grep'ing for `dep` doesn't suffice because it's also used as a variable name in some of the other code. 😀

jurre (commented):
We can't delete this just yet, we still have users that rely on this in dependabot-preview, we'll need to handle this a little more gracefully than just deleting the code.

jeffwidman:
No problem, if you need to run a more formal migration process I understand.

xlgmokha:
Let's revisit this PR after [August 3](https://github.blog/2021-04-29-goodbye-dependabot-preview-hello-dependabot/). I'm looking forward to merging this.

jeffwidman:
Rebased to fix merge conflicts now that August 3rd is upons us. @xlgmokha @jurre this is ready to _go_. 😜

jeffwidman:
Also, any reviewers please grep the remaining code for `"dep"`... as best I could tell the remaining references were unrelated, but I'd really appreciate a second set of eyes there.

The risk is that tests could pass but we leave dead code in that isn't caught.

jeffwidman:
Awesome! Thanks @jurre

PR #3541, opened 2021-04-21
docker: handle versions generated with `git describe --tags --long`

### LINKED ISSUE #3539: docker: not able to upgrade tags formatted with `git describe --tags --long`
**Package ecosystem**

docker

```Dockerfile
FROM infoblox/grafana-operator:v3.9.0-177-ged5bcde
```

**dependabot.yml content**

```yaml
version: 2
updates:
  - package-ecosystem: docker
    directory: "/.docker"
    schedule:
      interval: daily
```
**Updated dependency**

`infoblox/grafana-operator:v3.9.0-177-ged5bcde` => `infoblox/grafana-operator:v3.10.0-169-gfe040d3`

I expect Dependabot to be able to suggest an upgrade from `infoblox/grafana-operator:v3.9.0-177-ged5bcde` to `infoblox/grafana-operator:v3.10.0-169-gfe040d3`. Instead, Dependabot reports that the image is already up to date.
```
$ bin/dry-run.rb --dir '.docker' docker private-org/athena.grafana
=> fetching dependency files
=> dumping fetched dependency files: ./dry-run/private-org/athena.grafana/.docker
=> parsing dependency files
=> updating 1 dependencies: infoblox/grafana-operator
=== infoblox/grafana-operator (v3.9.0-177-ged5bcde)
 => checking for updates 1/1
 => latest available version is v3.9.0-177-ged5bcde
 => latest allowed version is v3.9.0-177-ged5bcde
```

These tags are generated with `git describe --tags --long`, which outputs a format like `<latest tag>-<number of commits since tag>-g<commit sha>`. This is a common way to generate version strings/tags, so I'd like it to be supported.

```Dockerfile
FROM infoblox/grafana-operator:v3.9.0-177-ged5bcde
```

## more detail

I am trying to debug this myself, but figured I should open this ticket to document my work.

The reason this upgrade is currently not suggested is because it fails the matching suffix check:
https://github.com/dependabot/dependabot-core/blob/b1f81765c32de788182ae204c802c3c83e081efd/docker/lib/dependabot/docker/update_checker.rb#L176

It thinks the commit sha is the suffix:
```
(byebug) suffix_of(tag)
"-gfe040d3"
(byebug) original_suffix
"-ged5bcde"
```

Also, the regex to detect a commit sha doesn't detect this format (needs to add `g?`):
https://github.com/dependabot/dependabot-core/blob/b1f81765c32de788182ae204c802c3c83e081efd/docker/lib/dependabot/docker/update_checker.rb#L193

But if it did, it would get rejected here anyway:
https://github.com/dependabot/dependabot-core/blob/b1f81765c32de788182ae204c802c3c83e081efd/docker/lib/dependabot/docker/update_checker.rb#L178

It seems like it shouldn't be too hard to rework this logic to support tags formatted by `git describe --tags --long`.

kd7lxl:
In #3539 I noted that I wanted to be able to upgrade docker version tags formatted with `git describe --tags --long`. The format for this is `<latest tag>-<number of commits since tag>-g<commit sha>`, meaning the first two parts are sortable and the last part should be ignored.

The docker updater has special handling for commit shas, but it doesn't seem to be needed. To make my case work, I removed the line that filters out tags with commit shas: `reject { |tag| commit_sha_suffix?(tag) }`. Surprisingly, this didn't break any test cases. Please review this thoroughly!

`numeric_from_version()` discards commit shas before comparison, so it's safe to leave tags with commits shas in the `candidate_tags` set:
```
(byebug) version
"v3.9.0-177-ged5bcde"
(byebug) numeric_version_from(version)
"3.9.0-177"
```

In this PR, I've changed the `comparable_tags_from_registry` to select tags that look like a commit sha, since they will be discarded later by `numeric_from_version`. This adds support for upgrading from tags like `v3.9.0-177-ged5bcde` to `v3.10.0-169-gfe040d3`.

Fixes #3539


drewwells (approved):

docker/spec/fixtures/docker/registry_tags/git_describe.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "infoblox/grafana-operator",
```
thepwagner, line 2: Noice: thanks for including a public image that behaves like this!

docker/spec/fixtures/docker/registry_tags/git_describe.json
```diff
@@ -0,0 +1,40 @@
...
+    "v3.7.0-13-g02d4c9ef",
+    "v3.7.0-14-g051e1a97",
+    "v3.7.0-15-g5c997230",
+    "v3.7.0-16-ga9d63825",
+    "v3.7.0-16-gd5a70ddc",
+    "v3.7.0-18-ga7fdf65c",
+    "v3.9.0-173-g4d84d01",
+    "v3.9.0-177-ged5bcde",
+    "v3.9.0-181-gbabf17f",
+    "v3.10.0-169-gfe040d3"
```
thepwagner, line 38:
```suggestion
    "v3.10.0-17-gfefefef",
    "v3.10.0-169-gfe040d3"
```

Idea: verify this bit of the version is being compared as integers and not strings. I think it is - but I wondered for long enough to double check.

```
irb(main):001:0> [17,169].sort
=> [17, 169]
irb(main):002:0> ["17","169"].sort
=> ["169", "17"]
```

thepwagner (approved):
LGTM, one idea for extending the test case (but making the change locally - the test still passes so it is potentially moot).

PR #3550, opened 2021-04-23
Transform update_types in IgnoreCondition

feelepxyz:
Moving the update_types transform from the config file parser to the
ignore condition making it easier to initialize IgnoreCondition without
fetching a config file.

Working towards being able to do this:

```ruby
ignore_conditions = ignore_conditions.map do |ic|
  Dependabot::Config::IgnoreCondition.new(
    dependency_name: ic["dependency-name"],
    versions: [ic["version-requirement"]],
    update_types: ic["update-types"],
  )
end
Dependabot::Config::UpdateConfig.
  new(ignore_conditions: ignore_conditions).
  ignored_versions_for(dependency)
```

### DIFF
--- a/common/lib/dependabot/config/ignore_condition.rb
+++ b/common/lib/dependabot/config/ignore_condition.rb
@@ -4,15 +4,14 @@ module Dependabot
   module Config
     # Filters versions that should not be considered for dependency updates
     class IgnoreCondition
-      UPDATE_TYPES = %i(
-        ignore_major_versions
-        ignore_minor_versions
-        ignore_patch_versions
-      ).freeze
+      PATCH_VERSION_TYPE = "version-update:semver-patch"
+      MINOR_VERSION_TYPE = "version-update:semver-minor"
+      MAJOR_VERSION_TYPE = "version-update:semver-major"
 
       ALL_VERSIONS = ">= 0"
 
       attr_reader :dependency_name, :versions, :update_types
+
       def initialize(dependency_name:, versions: nil, update_types: nil)
         @dependency_name = dependency_name
         @versions = versions || []
@@ -20,21 +19,25 @@ module Dependabot
       end
 
       def ignored_versions(dependency)
-        return [ALL_VERSIONS] if @versions.empty? && @update_types.empty?
+        return [ALL_VERSIONS] if versions.empty? && transformed_update_types.empty?
 
-        versions_by_type(dependency) + @versions
+        versions_by_type(dependency) + versions
       end
 
       private
 
+      def transformed_update_types
+        update_types.map(&:downcase).map(&:strip).compact
+      end
+
       def versions_by_type(dependency)
-        @update_types.flat_map do |t|
+        transformed_update_types.flat_map do |t|
           case t
-          when :ignore_patch_versions
+          when PATCH_VERSION_TYPE
             ignore_patch(dependency.version)
-          when :ignore_minor_versions
+          when MINOR_VERSION_TYPE
             ignore_minor(dependency.version)
-          when :ignore_major_versions
+          when MAJOR_VERSION_TYPE
             ignore_major(dependency.version)
           else
             []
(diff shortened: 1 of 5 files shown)


common/lib/dependabot/config/ignore_condition.rb
```diff
@@ -4,31 +4,40 @@ module Dependabot
...
-      UPDATE_TYPES = %i(
-        ignore_major_versions
-        ignore_minor_versions
-        ignore_patch_versions
-      ).freeze
+      UPDATE_TYPE_LOOKUP = {
+        "version-update:semver-patch" => :ignore_patch_versions,
+        "version-update:semver-minor" => :ignore_minor_versions,
+        "version-update:semver-major" => :ignore_major_versions
+      }.freeze
```
thepwagner, line 11: Question: what does this remapping accomplish now?

If we're happy assigning meaning to these strings, mapping them to string-agnostic symbols is a little moot.

thepwagner (approved):
LGTM: I'd assumed these exact strings only mattered when parsed directly from the config file.

If we're going to pass the verbatim strings through the external parser+storage and back into here, this works. I'd assumed it would be converted at some point, and intended to use the symbol.

What I really wanted is an enum in the JVM or an `iota` constant in Golang. Symbols where my guess at how Ruby might follow the same pattern...
With similar biases, I'd dig if these magic strings were defined and referenced as constants instead of duplicated across the codebase. The pattern in https://github.com/dependabot/dependabot-core/pull/3513/commits/7375421618e2ffd4380f0c767f453a2d5bb5bb9f#diff-67a26681133898ab39f6dac25dad58b8bfa751fd7245ec723c196f1184f93101R9 seemed pretty neat!

common/lib/dependabot/config/ignore_condition.rb
feelepxyz: Ah yes good point, will rejig this. I didn't think of persisting a transformed value but seems easier to just pass the string along as is? Typed enum would have been real nice!

feelepxyz:
> If we're going to pass the verbatim strings through the external parser+storage and back into here, this works. I'd assumed it would be converted at some point, and intended to use the symbol.

It would be real nice to iterate towards moving all config file validation and parsing to core and using that code from api, this would make it easier to consolidate on transforms in one place. We could then fail jobs if the config is currently invalid which would increase visibility of these issues.

common/lib/dependabot/config/ignore_condition.rb
```diff
@@ -4,37 +4,40 @@ module Dependabot
...
+        return [ALL_VERSIONS] if versions.empty? && transformed_update_types.empty?
 
-        versions_by_type(dependency) + @versions
+        versions_by_type(dependency) + versions
       end
 
       private
 
+      def transformed_update_types
+        update_types.map(&:downcase).map(&:strip).compact
```
thepwagner, line 30:
```suggestion
        update_types.compact.map(&:downcase).map(&:strip)
```

Nit: this `compact` does nothing, as nil values have already raised?
```
irb(main):002:0> [nil].map(&:downcase)
(irb):2:in `map': undefined method `downcase' for nil:NilClass (NoMethodError)
```

I'm not sure if it should be refactored to be a guard for the `map()`s (the suggestion), or just removed

thepwagner: Question: should this transform happen in the `ConfigFile` / parsing logic?

We might be able to simplify by assuming the caller of `IgnoreCondition` has normalized these values: either through `ConfigFile` or the mysterious server-side parsing process.

thepwagner (approved):
LGTM! Just a nit about the transform.

common/lib/dependabot/config/ignore_condition.rb
feelepxyz: Good catch! Yeah true, might be better to move to the `File` class and rely on the api to have done this already

PR #3552, opened 2021-04-23
IgnoreCondition: severity

thepwagner:
Continuing https://github.com/dependabot/dependabot-core/pull/3513 and https://github.com/dependabot/dependabot-core/pull/3550 , as the string values like `version-updates:semver-*` suggest, "update types" should be bypassed when Dependabot is creating an update in response to a security alert.

This PR introduces the vapourware concept of `security_severity` to the lookup of ignored versions for a dependency: if set to a non-nil value, the current `version-updates:*` types will be ignored.
All we really need here is a boolean flag, like `security_updates_only` (as suggested by @feelepxyz). The idea of building the API around severity is to empower some fast-follow features like: `update-types: ["security-updates:low-severity"]`.

### DIFF
--- a/common/spec/dependabot/config/ignore_condition_spec.rb
+++ b/common/spec/dependabot/config/ignore_condition_spec.rb
@@ -10,7 +10,7 @@ RSpec.describe Dependabot::Config::IgnoreCondition do
   let(:ignore_condition) { described_class.new(dependency_name: dependency_name) }
 
   describe "#versions" do
-    subject(:ignored_versions) { ignore_condition.ignored_versions(dependency) }
+    subject(:ignored_versions) { ignore_condition.ignored_versions(dependency, security_severity) }
     let(:dependency) do
       Dependabot::Dependency.new(
         name: dependency_name,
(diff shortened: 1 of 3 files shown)


common/spec/dependabot/config/ignore_condition_spec.rb
```diff
@@ -202,6 +203,15 @@ def expect_ignored(versions)
           end
         end
       end
+
+      context "with a security update" do
+        let(:security_severity) { "low" }
+        let(:update_types) { ["version-update:semver-patch"] }
+
+        it "ignores ignores update-types" do
```
thepwagner, line 211:
```suggestion
        it "ignores update-types" do
```

feelepxyz:
I'm wondering if this is where we'll end up using `security_severity` with `security-updates:low-severity`. Would it make more sense to skip running the job altogether if the severity is ignored in api as all versions will need to be ignored? Although, if we set the ignored version to `>= 0`, it would return immediately here: https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/update_checkers/base.rb#L41

Leaning towards a simpler interface of `security_updates_only` and punt on the decision to figure out how severity ignores will work. Thoughts?

I might be getting into the weeds here, either interface would be straight forward to change.

PR #3553, opened 2021-04-23
UpdateConfigSpec: verify name normalization

thepwagner:
Extends https://github.com/dependabot/dependabot-core/pull/3536 with a test that exercises `Dependency.name_normaliser_for_package_manager` is applied to both dependency names and ignore condition strings.

In practice, the only registered name normaliser is for Python: https://github.com/dependabot/dependabot-core/search?q=register_name_normaliser
Since that's not available in `/common`, the test introduces a locally scoped `fake-package-manager`.

### DIFF
--- a/common/spec/dependabot/config/update_config_spec.rb
+++ b/common/spec/dependabot/config/update_config_spec.rb
@@ -106,6 +106,40 @@ RSpec.describe Dependabot::Config::UpdateConfig do
         expect(ignored_versions).to eq([">= 13.a, < 14", ">= 12.13.a, < 13"])
       end
     end
+
+    context "with an dependency that must be name normalized" do
+      let(:dependency) do
+        Dependabot::Dependency.new(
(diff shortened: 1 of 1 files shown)


xlgmokha (approved):

thepwagner:
@feelepxyz this was supposed to reveal a bug, do the tests seem legit?

feelepxyz (approved):
Nice one, I think adding this test is a good shout but not sure this is testing the issue we where seeing in api as the python one didn't seem to have been registered there. 

It might just be a `require` issue as it seems this class wasn't loaded: https://github.com/dependabot/dependabot-core/blob/main/python/lib/dependabot/python.rb

thepwagner:
>  we where seeing in api as the python one didn't seem to have been registered there

Gotcha. That was presented to me as "name normalization doesn't work", I didn't know we'd debugged that deep.

PR #3562, opened 2021-04-26
IgnoreCondition: support `!!` sha ignores

thepwagner:
Amend `Dependabot::Config::IgnoreCondition` to support `!!` requirements for git dependencies.

This is a syntax previously used internally by Dependabot's backing service: if you `@dependabot ignore` an update from one Git SHA to another, one of these was created.

Since the syntax isn't supported by `Gem::Requirement`, care is taken to avoid parsing it as such: it's not returned in `ignored_versions,` but _is_ included in a new helper: `Dependabot::Config::IgnoreCondition#ignored?`.

### DIFF
--- a/common/lib/dependabot/config/ignore_condition.rb
+++ b/common/lib/dependabot/config/ignore_condition.rb
@@ -136,7 +136,7 @@ module Dependabot
                                 end
       end
 
-      # Custom requirement for ignoring specific git commit SHAs
+      # Custom requirement for ignoring all versions expect a specific SHA
       class ShaRequirement < Gem::Requirement
         def initialize(*requirements)
           requirements = requirements.flatten.flat_map do |req_string|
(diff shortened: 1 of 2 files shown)


common/lib/dependabot/config/ignore_condition.rb
```diff
@@ -95,6 +123,45 @@ def ignore_major(version)
...
+            return [GIT_SHA_IGNORE_PREFIX, obj.gsub(GIT_SHA_IGNORE_REGEX, "").strip]
+          end
+
+          super
+        end
+
+        def satisfied_by?(version)
+          if requirements.any? { |op, _| op == GIT_SHA_IGNORE_PREFIX }
+            rv = requirements.find { |op, _| op == GIT_SHA_IGNORE_PREFIX }.last
+            return rv == version.to_s
```
thepwagner, line 160: Concern: this is the opposite of the internal implementation, but given my assumptions is what _sort of_ made sense... But then, why would the requirement not just use `= ${commitSHA}` for that?

I'm leaning towards the base assumption being wrong: `!! ${commitSHA}` means: ignore versions other than `commitSHA`.
I'm not sure about the use case for that - why wouldn't you just ignore all versions?

thepwagner:
@feelepxyz also did this in https://github.com/dependabot/dependabot-core/pull/3564 , let's take that.

PR #3564, opened 2021-04-26
WIP: Add ignored_update? to update config

feelepxyz:
Attempting to add a new `ignored_update?` method to the update config that will allow checking if an updated dependency is ignored.

Adding support for `!!` prefix denoting a special git sha ignore syntax used by dependabot.

Thoughts on this approach?

### DIFF
--- a/bundler/lib/dependabot/bundler/requirement.rb
+++ b/bundler/lib/dependabot/bundler/requirement.rb
@@ -1,10 +1,11 @@
 # frozen_string_literal: true
 
 require "dependabot/utils"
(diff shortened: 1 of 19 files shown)


thepwagner:
This is https://github.com/dependabot/dependabot-core/pull/3562 , which I'd committed to adding in an internal comment on Friday (skipping link due to OSS repo).

feelepxyz:
Closing after discussing with @thepwagner - we'll keep the current handling of this outside of core for now

PR #3577, opened 2021-04-28
Drop GO111MODULE env var now that it defaults to `on`

jeffwidman:
`go` `1.16` flipped the default for `GO111MODULE` to `on`. So now that
ci targets `go` `1.16`, this is no longer needed.

Similar cleanup could be done for `dep`, but that is deprecated and will
soon be deleted, so I skipped it.

### DIFF
--- a/go_modules/helpers/Makefile
+++ b/go_modules/helpers/Makefile
@@ -3,7 +3,7 @@
 all: darwin linux
 
 darwin:
-	GO111MODULE=on GOOS=darwin GOARCH=amd64 go build -o go-helpers.darwin64 .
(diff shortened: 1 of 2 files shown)


jeffwidman-ns:
Test failure is on `npm_and_yarn` which is completely unrelated to this PR...

jeffwidman:
Test failure is on `npm_and_yarn` which is completely unrelated to this PR...

thepwagner (approved):

jeffwidman:
@thepwagner this is another reason to bump `go.mod` to `1.16` ;-) Again technically not necessary and somewhat unlikely to happen, but if someone tries to crib the build infra against an older `go` version, this could cause problems.

PR #3576, opened 2021-04-28
Cleanup go.mod handling of helpers

jeffwidman:
Since `go.mod` in the root of helpers declares itself as:
```
module github.com/dependabot/dependabot-core/go_modules/helpers
```

then my understanding is there's no need to have custom replace
directives... `go` is smart enough to realize that these are local pkgs
and should be imported locally.

As a result, we also no longer need a custom `go.mod` file within the
subpackage.

Which makes life easier as this subpackage declared an older version of
the extracted go modules repo... which works, but can be confusing since
there are two `go.mod` files with slightly different minimum supported
versions. Avoid all that by treating this as a package of the parent
modules rather than a standalone module.

PR #3578, opened 2021-04-28
Bump the go directive to 1.16

jeffwidman:
This isn't strictly required, but it makes life simpler in some edge
cases... and also later someone is considering updating, it's easier if
this has been kept up to date all along so the delta is smaller down the
road. I ran `go mod tidy` and it didn't update the `go.sum` file here
(sometimes it does when you bump this directive).

Note: This depends on https://github.com/dependabot/dependabot-core/pull/3576, as that reduced the number of `go.mod` files that needed this update. 
If that is rejected, then this change should be applied to the other `go.mod` files that weren't removed.

### DIFF
--- a/go_modules/helpers/go.mod
+++ b/go_modules/helpers/go.mod
@@ -1,6 +1,6 @@
 module github.com/dependabot/dependabot-core/go_modules/helpers
 
-go 1.13
+go 1.16
 
 require (
 	github.com/Masterminds/vcs v1.13.1
--- a/go_modules/helpers/go.mod
+++ b/go_modules/helpers/go.mod
@@ -4,13 +4,6 @@ go 1.13
 
 require (
 	github.com/Masterminds/vcs v1.13.1
-	github.com/dependabot/dependabot-core/go_modules/helpers/updater v0.0.0
 	github.com/dependabot/gomodules-extracted v1.2.0
 	golang.org/x/mod v0.4.2
 )
-
-replace github.com/dependabot/dependabot-core/go_modules/helpers/importresolver => ./importresolver
-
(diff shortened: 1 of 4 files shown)


thepwagner (approved):
👋 can you expand on the affected edge cases for posterity?

As the go directive refers to the [minimum-go-version](https://golang.org/doc/modules/gomod-ref#go), it's tempting to leave as-is (the helpers code base _is_ theoretically compatible with `go1.13`, right?).

But we only compile and test with [go1.16](https://github.com/dependabot/dependabot-core/pull/3233) - I like the idea of aligning the directive with what we're testing to prevent regressions!

jeffwidman:
@thepwagner so I'm actually ignorant of the specific edge cases, but when I was investigating some other stuff, I've seen comments from bcmills and others saying that in every version of `go` they've fixed bugs in how `go mod tidy` works, so the behavior on some edge cases is very version-specific.

That was enough for me and I didn't dig further, because IMO if you've got CI already running at `go 1.16` then even though technically the helpers code bases _is_ compatible with older go versions, this short-circuits risks of users reporting bugs that are specific to older versions of `go` (which you wouldn't realize until you've spent a lot of time investigating).

Additionally, as I mentioned, this makes it simpler when you later upgrade to some newer go version where you absolutely have to bump because you only have to check the delta between 1.16 behavior and the newer behavior, not from 1.13 onwards. Small thing, but I've found things like that cut down on maintenance time in other projects I maintain.

Anyway, sorry it's not a smoking gun, but esp the edge cases idea left me saying "I think it's simpler/less time going forward to keep this in lock step with CI for now".

PR #3456, opened 2021-04-07
Handle repo not found errors for go modules

xlgmokha:
This change reproduces an error produced by the `dry-run` script and resolves it by detecting `repository not found` error messages.

Before:

```bash
./bin/dry-run.rb go_modules "DavidSpek/pytorch-operator" --commit=3ca1300e834b15e6f37e576f2db6254ce17a7efd --dep github.com/go-openapi/spec
=> cloning into /tmp/d20210407-53-17pq0yq
=> parsing dependency files
=> updating 1 dependencies: github.com/go-openapi/spec

=== github.com/go-openapi/spec (0.19.2)
 => checking for updates 1/1
 => latest available version is 0.20.3
 => latest allowed version is 0.20.3
 => requirements to unlock: own
 => requirements update strategy:
 => updating github.com/go-openapi/spec from 0.19.2 to 0.20.3
Traceback (most recent call last):
        16: from ./bin/dry-run.rb:567:in `<main>'
        15: from ./bin/dry-run.rb:567:in `each'
        14: from ./bin/dry-run.rb:655:in `block in <main>'
        13: from /home/dependabot/dependabot-core/go_modules/lib/dependabot/go_modules/file_updater.rb:34:in `updated_dependency_files'
        12: from /home/dependabot/dependabot-core/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb:63:in `updated_go_mod_content'
        11: from /home/dependabot/dependabot-core/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb:76:in `updated_files'
        10: from /home/dependabot/dependabot-core/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb:80:in `update_files'
         9: from /home/dependabot/dependabot-core/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb:197:in `in_repo_path'
         8: from /home/dependabot/dependabot-core/common/lib/dependabot/shared_helpers.rb:34:in `in_a_temporary_repo_directory'
         7: from /home/dependabot/dependabot-core/common/lib/dependabot/shared_helpers.rb:34:in `chdir'
         6: from /home/dependabot/dependabot-core/common/lib/dependabot/shared_helpers.rb:34:in `block in in_a_temporary_repo_directory'
         5: from /home/dependabot/dependabot-core/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb:198:in `block in in_repo_path'
[...]

PR #3580, opened 2021-04-28
Sync copy/pasted private methods with upstream

jeffwidman:
`git blame` looks like the only change was https://github.com/golang/mod/commit/788aebd06792f833c387fcee8647edac94108a98

Added a "sync date/commit sha" so that it's easier to see when these were last updated.

### DIFF
--- a/go_modules/helpers/updater/helpers.go
+++ b/go_modules/helpers/updater/helpers.go
@@ -6,7 +6,9 @@ import (
 	"golang.org/x/mod/modfile"
 )
 
-// Private methods lifted from the `modfile` package
+// Private methods lifted from the `modfile` package.
+// Last synced: 4/28/2021 from:
+// https://github.com/golang/mod/blob/858fdbee9c245c8109c359106e89c6b8d321f19c/modfile/rule.go
 
 // setIndirect sets line to have (or not have) a "// indirect" comment.
 func setIndirect(line *modfile.Line, indirect bool) {
@@ -20,13 +22,17 @@ func setIndirect(line *modfile.Line, indirect bool) {
 			line.Suffix = []modfile.Comment{{Token: "// indirect", Suffix: true}}
 			return
 		}
-		// Insert at beginning of existing comment.
+
 		com := &line.Suffix[0]
-		space := " "
-		if len(com.Token) > 2 && com.Token[2] == ' ' || com.Token[2] == '\t' {
-			space = ""
+		text := strings.TrimSpace(strings.TrimPrefix(com.Token, string(slashSlash)))
+		if text == "" {
+			// Empty comment.
+			com.Token = "// indirect"
+			return
 		}
-		com.Token = "// indirect;" + space + com.Token[2:]
+
+		// Insert at beginning of existing comment.
+		com.Token = "// indirect; " + text
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


go_modules/helpers/updater/helpers.go
```diff
@@ -6,7 +6,11 @@ import (
...
 
-// Private methods lifted from the `modfile` package
+// Private methods lifted from the `modfile` package.
+// Last synced: 4/28/2021 from:
+// https://github.com/golang/mod/blob/858fdbee9c245c8109c359106e89c6b8d321f19c/modfile/rule.go
+
+var slashSlash = []byte("//")
 
 // setIndirect sets line to have (or not have) a "// indirect" comment.
 func setIndirect(line *modfile.Line, indirect bool) {
```
thepwagner, line 16: Question (maybe for @jurre): do we still need to do this?

Since https://github.com/dependabot/dependabot-core/pull/2551 , I'd expect `go mod tidy` to detect and handle `// indirect` in the modfile - and probably do a better job than this...

Is this potentially dead code?

thepwagner (approved):
Thanks for adding the extraction information to the code, and linking to the affected PR - that made reviewing this update a breeze!

go_modules/helpers/updater/helpers.go
jurre: Yeah I think it's mostly safe to drop this, there are (in theory at least?) some edge-cases in which `go mod tidy` will silently fail and we ignore that, I'm not quite sure what those scenario's are though. In that case this code would still add the `// indirect` comment, but it's unclear to me how valuable that is if `go mod tidy` has already failed

thepwagner: I took at timeboxed look at this: [this test](https://github.com/dependabot/dependabot-core/blob/552be5d0236078c18574bba4baa59aa9284da56e/go_modules/spec/dependabot/go_modules/file_updater_spec.rb#L147-L149) started failing.

That harness under test is a little garbage-in-garbage-out. Since https://github.com/dependabot/dependabot-core/pull/3456 , it's invalid:
```
~/git/github/dependabot-core/go_modules/spec/fixtures/projects/repo_not_found $ go run .
go: github.com/go-openapi/spec@v0.19.2 requires
	github.com/go-openapi/jsonpointer@v0.19.2: missing go.sum entry; to add it:
	go mod download github.com/go-openapi/jsonpointer
```

Given the expected error caused by https://github.com/kubernetes/kubernetes/blob/2e808b7cb054ee242b68e62455323aa783991f03/go.mod#L104 , maybe that harness should use `k8s.io/kubernetes` and not worry about `github.com/go-openapi/spec`? 🤷 

I created https://github.com/dependabot-fixtures/go-modules-public to potentially simplify the harness to 3 dependencies, but that fails with a different error (the `PrivateSourceAuthenticationFailure` error is raised in both cases).

I'm just going to merge this update PR and retreat.

jeffwidman: @thepwagner Do you want to at least file an issue tracking this failure and including a link to your investigation so far? No sense making the next person have to duplicate your initial work here...

thepwagner: @jeffwidman I'm not sure it's a change we should pursue: I suspected the code isn't necessary, but it seems to be.

jeffwidman: Ah right, I misread your comment as investigating a test failure on my _original_ change, didn't realize the test failure was on your experiment. Makes complete sense.

PR #3583, opened 2021-04-28
Check the err returned by AddRequire

jeffwidman:
Looking at the source of `AddRequire`, it looks like it never actually returns a non-`nil` error: https://github.com/golang/mod/blob/858fdbee9c245c8109c359106e89c6b8d321f19c/modfile/rule.go#L828-L846

However, they wrote the func signature to return an `err` to give themselves future flexibility.

So we should respect that and check for any unexpected errors. This way future updates that
suddenly start returning errors will be caught rather than insidiously slipping by and causing
difficult-to-debug issues elsewhere.

### DIFF
--- a/go_modules/helpers/updater/main.go
+++ b/go_modules/helpers/updater/main.go
@@ -28,7 +28,10 @@ func UpdateDependencyFile(args *Args) (interface{}, error) {
 	}
 
 	for _, dep := range args.Dependencies {
-		f.AddRequire(dep.Name, dep.Version)
+		err := f.AddRequire(dep.Name, dep.Version)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	for _, r := range f.Require {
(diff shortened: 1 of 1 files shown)


jeffwidman:
Also, I found this by running `golangci-lint run` against the `go_modules/helpers` directory. I don't have time to do so myself, but it might be useful for the maintainers to set that up as a dedicated linter through GitHub Actions to prevent issues like this from creeping in down the road...

go_modules/helpers/updater/main.go
```diff
@@ -28,7 +28,10 @@ func UpdateDependencyFile(args *Args) (interface{}, error) {
 	}
 
 	for _, dep := range args.Dependencies {
-		f.AddRequire(dep.Name, dep.Version)
+		err := f.AddRequire(dep.Name, dep.Version)
+		if err != nil {
```
thepwagner, line 32:
```suggestion
		if err := f.AddRequire(dep.Name, dep.Version); err != nil {
```

Nit: minimize scope of `err`.

thepwagner (approved):

go_modules/helpers/updater/main.go
jeffwidman: Yeah, I'm always torn on this one because it's less LOC, and tighter scope, but I also find it less readable. Happy to go either way.

PR #3588, opened 2021-04-28
IgnoreCondition: don't ignore current version when ignoring patches and handle multi-length semver ranges

mctofu:
When handling the relative `version-update:semver-patch` ignore condition we were generating an ignored version range that included the current version. This was causing the latest allowed version to resolve to a prior version when running an update:
```
=== ssri (8.0.0)
 => checking for updates 1/1
 => latest available version is 7.1.0
 => latest allowed version is 7.1.0
```

This tweaks the starting range of the ignore condition so that the current version is excluded:
```
=== ssri (8.0.0)
 => checking for updates 1/1
 => latest available version is 8.0.0
 => latest allowed version is 8.0.0
```

I also added some tests to verify we didn't exclude the current version when ignoring minor or major ranges.

### DIFF
--- a/common/lib/dependabot/config/ignore_condition.rb
+++ b/common/lib/dependabot/config/ignore_condition.rb
@@ -62,7 +62,7 @@ module Dependabot
         return [] unless rubygems_compatible?(version)
 
         parts = version.split(".")
-        version_parts = parts.fill(0, parts.length...3)
+        version_parts = parts.fill(0, parts.length...2)
         lower_parts = version_parts.first(1) + [version_parts[1].to_i + 1] + ["a"]
         upper_parts = version_parts.first(0) + [version_parts[0].to_i + 1]
         lower_bound = ">= #{lower_parts.join('.')}"
@@ -74,8 +74,7 @@ module Dependabot
       def ignore_major(version)
         return [] unless rubygems_compatible?(version)
 
-        parts = version.split(".")
-        version_parts = parts.fill(0, parts.length...2)
+        version_parts = version.split(".")
         lower_parts = [version_parts[0].to_i + 1] + ["a"]
         upper_parts = [version_parts[0].to_i + 2]
         lower_bound = ">= #{lower_parts.join('.')}"
--- a/common/lib/dependabot/config/ignore_condition.rb
+++ b/common/lib/dependabot/config/ignore_condition.rb
@@ -50,10 +50,10 @@ module Dependabot
         parts = version.split(".")
         return [] unless parts.size > 2
 
-        lower_parts = parts.first(2) + ["a"]
+        lower_parts = parts.first(3)
         upper_parts = parts.first(2)
         upper_parts[1] = upper_parts[1].to_i + 1
-        lower_bound = ">= #{lower_parts.join('.')}"
+        lower_bound = "> #{lower_parts.join('.')}"
         upper_bound = "< #{upper_parts.join('.')}"
         ["#{lower_bound}, #{upper_bound}"]
       end
--- a/common/lib/dependabot/config/ignore_condition.rb
+++ b/common/lib/dependabot/config/ignore_condition.rb
@@ -47,60 +47,45 @@ module Dependabot
       end
 
       def ignore_patch(version)
+        return [] unless rubygems_compatible?(version)
+
         parts = version.split(".")
         version_parts = parts.fill(0, parts.length...4)
-        lower_parts = if numeric_version?(version)
-                        version_parts.first(3) + [version_parts[3].to_i + 1] + ["a"]
-                      else
-                        version_parts.first(2) + [version_parts[2].to_i + 1] + ["a"]
-                      end
-        upper_parts = if numeric_version?(version)
-                        version_parts.first(1) + [version_parts[1].to_i + 1]
-                      else
-                        version_parts.first(2) + [999_999]
-                      end
+        lower_parts = version_parts.first(3) + [version_parts[3].to_i + 1] + ["a"]
+        upper_parts = version_parts.first(1) + [version_parts[1].to_i + 1]
         lower_bound = ">= #{lower_parts.join('.')}"
         upper_bound = "< #{upper_parts.join('.')}"
+
         ["#{lower_bound}, #{upper_bound}"]
       end
 
       def ignore_minor(version)
+        return [] unless rubygems_compatible?(version)
+
         parts = version.split(".")
         version_parts = parts.fill(0, parts.length...3)
-        lower_parts = if numeric_version?(version)
-                        version_parts.first(1) + [version_parts[1].to_i + 1] + ["a"]
-                      else
-                        version_parts.first(1) + ["a"]
-                      end
-        upper_parts = if numeric_version?(version)
-                        version_parts.first(0) + [version_parts[0].to_i + 1]
-                      else
-                        version_parts.first(1) + [999_999]
-                      end
+        lower_parts = version_parts.first(1) + [version_parts[1].to_i + 1] + ["a"]
+        upper_parts = version_parts.first(0) + [version_parts[0].to_i + 1]
         lower_bound = ">= #{lower_parts.join('.')}"
         upper_bound = "< #{upper_parts.join('.')}"
+
         ["#{lower_bound}, #{upper_bound}"]
       end
 
       def ignore_major(version)
+        return [] unless rubygems_compatible?(version)
+
         parts = version.split(".")
         version_parts = parts.fill(0, parts.length...2)
-        lower_parts = if numeric_version?(version)
-                        [version_parts[0].to_i + 1] + ["a"]
-                      else
-                        version_parts.first(1) + [999_999]
-                      end
-        upper_parts = if numeric_version?(version)
-                        [version_parts[0].to_i + 2]
-                      else
-                        [999_999]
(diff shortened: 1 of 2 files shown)


common/lib/dependabot/config/ignore_condition.rb
```diff
@@ -50,10 +50,10 @@ def ignore_patch(version)
...
 
-        lower_parts = parts.first(2) + ["a"]
+        lower_parts = parts.first(3)
         upper_parts = parts.first(2)
         upper_parts[1] = upper_parts[1].to_i + 1
-        lower_bound = ">= #{lower_parts.join('.')}"
+        lower_bound = "> #{lower_parts.join('.')}"
         upper_bound = "< #{upper_parts.join('.')}"
         ["#{lower_bound}, #{upper_bound}"]
       end
```
feelepxyz, line 59: Nice catch on the current version being ignored 👍 

I'm wondering if we also want to be able to handle longer and shorter versions here, we currently bail if the version has less than 3 parts which prevents us creating patch ignores for `1.2`, which we could assume means `1.2.0` (we do this in api when creating ignore conditions from PR commands).

Something like this might fix the current bug and work for `1.2` and `1.2.3.1`:

```ruby
def ignore_patch(version)
    parts = version.split(".")
    version_parts = parts.fill(0, parts.length...4)
    lower_parts = version_parts.first(3) + [version_parts.last.to_i + 1] + ["a"]
    upper_parts = version_parts.first(1) + [version_parts[1].to_i + 1]
    lower_bound = ">= #{lower_parts.join('.')}"
    upper_bound = "< #{upper_parts.join('.')}"
    ["#{lower_bound}, #{upper_bound}"]
  end
```

But might need some special handling for non-numeric versions, e.g. `Finchley.SR3`

common/lib/dependabot/config/ignore_condition.rb
```diff
@@ -50,10 +50,10 @@ def ignore_patch(version)
         parts = version.split(".")
         return [] unless parts.size > 2
 
-        lower_parts = parts.first(2) + ["a"]
+        lower_parts = parts.first(3)
```
thepwagner, line 53: Question: can we drop this `a`?

For context, it's a metacharacter recognized by [Gem::{Version.Requirement}](https://docs.ruby-lang.org/en/3.0.0/Gem/Version.html), which we is the implementation of semver used for most ecosystems.

> If any part contains letters (currently only a-z are supported) then that version is considered prerelease. Versions with a prerelease part in the Nth part sort less than versions with N-1 parts. Prerelease parts are sorted alphabetically using the normal Ruby string sorting rules. If a prerelease part contains both letters and numbers, it will be broken into multiple parts to provide expected sort behavior (1.0.a10 becomes 1.0.a.10, and is greater than 1.0.a9).

I ask because I just cargo culted it along in the base implementation, the change makes sense but I didn't dig into the implications.

thepwagner (approved):
Doh; first pass was suppose to be an approval: LGTM I'd just like more details about the `a`.

common/lib/dependabot/config/ignore_condition.rb
feelepxyz: I went down a bit of a yak shave here while trying to get my head around some test cases: https://github.com/dependabot/dependabot-core/pull/3593

Do these test cases seem legit to you?

cc @mctofu and @thepwagner

common/lib/dependabot/config/ignore_condition.rb
mctofu: Good question! I haven't yet thought of a case where it would matter but @feelepxyz figured out what it would look like to keep the `a` here:
https://github.com/dependabot/dependabot-core/blob/2dddae6b97cbef2846d31e4ceecc9f4fbacf0714/common/spec/dependabot/config/ignore_condition_spec.rb#L89

feelepxyz (approved):
🙌

common/lib/dependabot/config/ignore_condition.rb
```diff
@@ -48,53 +48,63 @@ def versions_by_type(dependency)
...
-          lower_parts = parts.first(1) + ["a"]
-          upper_parts = parts.first(1)
-          begin
-            upper_parts[0] = Integer(upper_parts[0]) + 1
-          rescue ArgumentError
-            upper_parts.push(999_999)
-          end
-        end
-
+        version_parts = parts.fill(0, parts.length...3)
```
thepwagner, line 69: Nit: this seems an awful lot like `ignore_patch(version)` now, but with `version_parts = parts.fill(0, parts.length...4)` is now `version_parts = parts.fill(0, parts.length...3)`.

Is it possible we could fold this back to a single function that accepts a `precision` parameter? https://github.com/dependabot/dependabot-core/pull/3513/commits/7375421618e2ffd4380f0c767f453a2d5bb5bb9f#diff-67a26681133898ab39f6dac25dad58b8bfa751fd7245ec723c196f1184f93101R54-R68

common/spec/dependabot/config/ignore_condition_spec.rb
```diff
@@ -121,6 +124,53 @@ def expect_ignored(versions)
         it "ignores expected versions" do
           expect_allowed(minor_upgrades)
           expect_ignored(patch_upgrades + major_upgrades)
+          expect_allowed([dependency_version])
+        end
+      end
+
+      context "with a `major.minor.patch.patch` semver version" do
```
thepwagner, line 131:
```suggestion
      context "with a `major.minor.patch.extra` version" do
```

Nit: `patch.patch` is not a thing, https://semver.org/ . Could we set another name for this non-semver string?

common/spec/dependabot/config/ignore_condition_spec.rb
```diff
@@ -121,6 +124,53 @@ def expect_ignored(versions)
...
+          it "returns the expected range" do
+            expect(ignored_versions).to eq([">= 1.3.a, < 2"])
+          end
+        end
+
+        context "with ignore_patch_versions" do
+          let(:update_types) { ["version-update:semver-patch"] }
+
+          it "ignores expected updates" do
+            expect_ignored(["1.2.3.2", "1.2.4.0"])
```
thepwagner, line 166:
```suggestion
            expect_ignored(patch_upgrades + ["1.2.3.2", "1.2.4.0"])
```

Idea: include the base cases.

thepwagner (approved):
I understood the original intent of not ignoring the current version.

I'm less sure about the continued drift from "semver" to "semver-ish strings" that variable length versions bring along. It's tempting to keep the semver-filtering features focussed on semver™️ for clarity's sake.

common/lib/dependabot/config/ignore_condition.rb
```diff
@@ -48,53 +48,63 @@ def versions_by_type(dependency)
 
       def ignore_patch(version)
         parts = version.split(".")
-        return [] unless parts.size > 2
-
-        lower_parts = parts.first(2) + ["a"]
-        upper_parts = parts.first(2)
-        upper_parts[1] = upper_parts[1].to_i + 1
+        version_parts = parts.fill(0, parts.length...4)
+        lower_parts = if numeric_version?(version)
```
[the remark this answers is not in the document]
mctofu: Good question! Digging into this is making me feel like this is more complicated than it needs to be. I'm going to take another pass and look at simplifying. I think we don't need to be too concerned about the behavior when non semver versions are provided.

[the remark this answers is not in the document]
mctofu: Ok, I've simplified this a bit. Instead of trying to deal with non ruby gem compatible versions we bail early with no results. We'd still do weird things with a version like `1.banana` but I don't think that should end up breaking anything.

I also found the attempt to use `.a` in the ignore-patch version ranges was causing issues with pre-release semver versions like `1.2.3-alpha.1`. I've reverted back to the original attempt which used `>` instead of `>=` & `.a` in the ignore range.

common/spec/dependabot/config/ignore_condition_spec.rb
```diff
@@ -172,24 +180,85 @@ def expect_ignored(versions)
...
 
           it "returns the expected range" do
-            expect(ignored_versions).to eq([])
+            expect(ignored_versions).to eq(["> 1, < 1.1"])
+          end
+        end
+      end
+
+      context "with a pre-release semver version" do
+        let(:dependency_version) { "1.2.3-alpha.1" }
```
thepwagner, line 221: Noice: I dig this additional test case, especially because `parts[2] == "3-alpha"`. That's a spicy meatball to verify!

common/spec/dependabot/config/ignore_condition_spec.rb
```diff
@@ -207,7 +276,14 @@ def expect_ignored(versions)
         context "with ignore_minor_versions" do
           let(:update_types) { ["version-update:semver-minor"] }
           it "returns the expected range" do
-            expect(ignored_versions).to eq([">= Finchley.a, < Finchley.999999"])
+            expect(ignored_versions).to eq([])
```
thepwagner, line 279: Aside: this is a regression from how `@dependabot ignore` behaves on such a request.
I'm cool with that, as it seems consistent for the `semver-*` to enforce semver semantics and not sweat the edge cases.

thepwagner (approved):

PR #3590, opened 2021-04-29
Switch to letting `go get` mutate `go.mod`

### LINKED ISSUE #3526: Unfixable failure when the dep tree includes a dependency circle and one of those deps at one time referenced a broken module
A particular `go` library has a very convoluted dep tree:

* A > B > A
* A > C > D > A & B (and B of course depends on A)

Unfortunately C also depends on another repo that released a version of their library that they later removed, so it broke the `go.sum` checks: https://github.com/grpc/grpc-go/issues/3945

Breaking this cycle of dependabot death has been a royal pain. It's easy to bump the broken version within C. But then when dependabot tries to walk the deps of the current (fixed) C, it finds D, then finds A, which relies on an older version of C which is broken. 

Literally I'm seeing error messages in repo A that look like:
```
go: github.com/(redacted)/C@v0.5.0 requires
	github.com/(redacted)/D@v0.4.0 requires
	github.com/(redacted)/A@v0.17.0 requires
	github.com/(redacted)/C@v0.1.5 requires
	github.com/(redacted)/D@v0.3.0 requires
	github.com/(redacted)/B@v0.0.57 requires
	github.com/(redacted)/A@v0.0.92 requires
	google.golang.org/grpc@v1.33.0: reading google.golang.org/grpc/go.mod at revision v1.33.0: unknown revision
```

I've been manually unrolling the loop by slowly bumping things and cutting new versions, but I was wondering if there was a more efficient way to resolve this?

I was a bit surprised in the above error message to see Dependabot ping-ponging between older and older versions of the same library... shouldn't it be smart enough to cache the first time it sees a library version pin and if it later sees older pins of that library to ignore them and eliminate that subtree as completely irrelevant?

Additionally, `go.mod` supports the concept of `// indirect` whereby I can specify a minimum version for a dep of a dep... If dependabot prunes subtrees for older versions, then in the top-level repo I could add some `// indirect` pins which would force dependabot's tree walker to simply skip old/broken versions of underlying deps.

In addition to making it faster to unbreak things, it'd also speed up the dep runs significantly.

I am unclear if this is a bug, or desired behavior... 

The only reason I could see it being expected behavior is to guard against an edge case where a dep pins a dep which is incompatible with another dep... but again, if the parent dep has already pinned a library version, then somewhere a human has already explicitly said "this is okay". I suppose there may be some really weird edge case where an incompatibility is added after the fact to a newer release of an underlying dep... but there's still got to be some cleaner way to workaround this.

jeffwidman:
Previously, we were manually modifying `go.mod` for the dep we wanted to
bump, then running `go get` to update `go.sum` and run the MVS algorithm
to determine if any other changes needed to be made to `go.mod`.

However, this caused a few problems:
* the `go` docs/comments on github issues consistently recommend _not_
editing `go.mod` directly, but instead use `go get` to update it.
* We have to copy/paste in private methods from upstream, and it's a
pain to keep those in sync: https://github.com/dependabot/dependabot-core/pull/3580
* When `go get` sees that `go.mod` is in a state that has a version
which it can't find in `go.sum`, then it _appears_ (although I'm not
100% certain) to try to recover by doing a bunch of additional checks to
ensure it fixes the dep graph correctly. I saw @bcmills mention
something along these lines in a github issue although I'm afraid I
can't find the exact reference now. Compare to if `go get` is run
against a `go.mod` that has a matching (`tidy`'d) `go.sum`, it appears
to skip those checks. That seems to be the root cause of #3526 where
those checks result in an impossible-to-recover-from-situation. (Side
note: that problem looks like it will get resolved in `go 1.17` via the
new [lazy module loading](http://golang.org/design/36460-lazy-module-loading).)
* In some cases, we're doing an unecessary write to disk since `go.mod`
gets written by us and then re-written by `go get`. That could be a
single write.

Instead of doing all this munging, we can simply pass our desired
version to `go get -d <dep>@<version>` and it will handle updating
`go.mod` and `go.sum`.

Experimenting locally, this fixes #3526 which I was previously hitting
in multiple repos. They all work now.

It also lets us do some code cleanup, skip an uneccessary write-to-disk,
make all deeply nested dep trees process much faster, becomes more
idiomatic, and reduces the risk of our logic for updating `go.mod` from
diverging from upstream.


go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -91,16 +91,17 @@ def update_files # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity
...
             substitute_all(substitutions)
 
-            # Set the stubbed replace directives
-            update_go_mod(dependencies)
-
-            # Then run `go get` to pick up other changes to the file caused by
-            # the upgrade
-            run_go_get
+            # TODO as far as I can tell, this array will only ever have one value, but we should assert that somehow
+            # and blow up if it's ever not true... I'm not a rubyist, so not sure how best to handle?
```
jeffwidman, line 95: Help please?

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -91,16 +91,17 @@ def update_files # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity
...
+            dep = dependencies.first
+            # `go get` will generate the required go.mod/go.sum updates for the new dep version
+            run_go_get(dep)
 
             # If we stubbed modules, don't run `go mod {tidy,vendor}` as
             # dependencies are incomplete
             if substitutions.empty?
+              # go mod tidy should run before go mod vendor to ensure any
+              # dependencies removed by go mod tidy are also removed from vendors.
```
jeffwidman, line 104: I ran across this comment on an old PR while researching my change... the comment seems very useful/easy to miss, so thought I should add it here. If needed I can break it out as a separate PR.

go_modules/lib/dependabot/go_modules/update_checker.rb
```diff
@@ -64,6 +64,18 @@ def find_latest_resolvable_version
...
+            # we could invert this and run `go get dep@latest` and then
+            # inspect the go.mod/go.sum file to see what it was bumped
+            # to, and then flow from there. It would remove the need for
+            # https://github.com/dependabot/gomodules-extracted.
+            # But currently `go get` is a bit slower than our fast
+            # version checker. That may change in go 1.17 with
+            # lazy-loading: https://github.com/golang/go/issues/36460
+            # Also it's unclear if this would work with the pipeline
+            # approach of dependabot which naturally leads to splitting
+            # the finding of updates from actually updating to them.
```
jeffwidman, line 78: So this is a much, much bigger design conversation. Def out of scope for the current PR (so possibly I should drop the code comment from this PR).

But I wanted to at least point out that there's a lot of custom code whose sole goal is to duplicate what `go get -d <dep>@latest` would do... That's the _only_ reason we need https://github.com/dependabot/dependabot-core/blob/main/go_modules/helpers/updatechecker/main.go which is the _only_ reason we need an entire repo https://github.com/dependabot/gomodules-extracted/

I realize that dependabot needs to know what the updated version was, so you'd invert the flow by then checking `go.mod` to see what the new version is and then returning it out to the code that opens the PR. And that may be to much of an inversion and simply not work with the dependabot pipeline flow. Or the perf hit from running `go get` may simply be too costly compared to running this really simple updater code (since `go get -d` not only checks for latest but also downloads the module code). But the code maintainability savings/lowered chance of bugs (at least on the `go get` logic) are quite large too. So it's a big tradeoff either way.

go_modules/helpers/updater/helpers.go
```diff
@@ -1,65 +0,0 @@
...
-
-import (
-	"strings"
-
-	"golang.org/x/mod/modfile"
-)
-
-// Private methods lifted from the `modfile` package.
-// Last synced: 4/28/2021 from:
-// https://github.com/golang/mod/blob/858fdbee9c245c8109c359106e89c6b8d321f19c/modfile/rule.go
```
jeffwidman, line 11: @thepwagner @jurre this nicely solves this morning's discussion here: https://github.com/dependabot/dependabot-core/pull/3580#discussion_r622096596

jeffwidman:
I realize tests are failing, but before spending any time on those, can I get a yes/no on the overall direction of this PR?

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jurre: Tbh I'm not entirely sure if this is only ever 1 dependency or multiple in some cases. Can we run `go get` for each dependency in the array? That seems safest.

jurre:
In general I think the approach is an improvement 👍 , I'm not entirely sure if there is a technical reason that Dependabot manually updates the `go.mod` file though 🤔, the test-failures _seem_ like they are just errors that manifest slightly differently now, but they may reveal the true reason behind the way things are?

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -179,7 +161,12 @@ def run_go_get
 
           File.write(tmp_go_file, "package dummypkg\n") unless package
 
-          _, stderr, status = Open3.capture3(ENVIRONMENT, "go get -d")
+          # Use version pinning rather than `latest` just in case
+          # a new version gets released in the middle of our run.
+          version = "v" + dep.version.sub(/^v/i, "")
+          # TODO: go 1.18 will make `-d` the default behavior, so remove the flag then
+          command = "go get -d #{dep.name}@#{version}"
```
thepwagner, line 168: I'm pondering the behaviour of this when there's a replacement applied to the module, e.g.
```
go 1.15

require github.com/pkg/errors v0.8.0
replace github.com/pkg/errors => github.com/thepwagner/errors v0.8.0
```


--- 

In upgrading to `0.8.1` (a version available in both repos), I'd expect Dependabot to output:
```
require github.com/pkg/errors v0.8.0
replace github.com/pkg/errors => github.com/thepwagner/errors v0.8.1
```



But `go get -d github.com/pkg/errors@v0.8.1` produces:
```
require github.com/pkg/errors v0.8.1
replace github.com/pkg/errors => github.com/thepwagner/errors v0.8.0
```

That has no effect, since the `github.com/thepwagner/errors v0.8.0` replacement is still applied.


And `go get -d github.com/thepwagner/errors@v0.8.1` produces:
```
require (
	github.com/pkg/errors v0.8.0
	github.com/thepwagner/errors v0.8.1 // indirect
)

replace github.com/pkg/errors => github.com/thepwagner/errors v0.8.0
```

That has no effect, since my code doesn't use that package.

---

FWIW, `updateDependencyFile` doesn't handle this correctly today either 😿 . I think https://github.com/thepwagner/action-update-go/blob/2ba2b9a3d49a3285b327622fbd5daccb4fe64e19/gomodules/applyupdate.go#L101-L136 is an implementation that does - and it looks closer to `updateDependencyFile` than `go get -d`.

Have you looked at `go mod edit` as an alternative to `updateDependencyFile`? It's not quite plug and play, but at least it's provided+supported by the SDK.

thepwagner (commented):
I like the idea of reducing the scope of these helpers, and I respect any non-rubyist crawling through this code ❤️ .

My concerns are around things the `go_modules` plugin doesn't do correctly today (module replacements, ignored versions), and wondering if this change makes that harder/easier in the future.

In particular with [the recent-ish focus on ignoring versions](https://github.com/dependabot/dependabot-core/pulls?q=is%3Apr+is%3Aclosed+ignore+condition+created%3A%3E2021-04-01+), the latter may be improved shortly...

go_modules/lib/dependabot/go_modules/update_checker.rb
jeffwidman: > Dependabot can be configured to ignore certain versions of a dependency (e.g. >= 1.0.0).

That makes sense. I had a feeling there were going to be some use cases that make this non-tenable.

> I think something like `go list -m -mod=mod -versions -json <dep>` is a better solution than the current getUpdatedVersion helper.

Oh, I hadn't even considered that. Regardless, this is way out of scope for the current PR, so I filed https://github.com/dependabot/dependabot-core/issues/3599 and we can continue the conversation there.

I'll remove this comment from this PR.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jeffwidman: > In upgrading to 0.8.1 (a version available in both repos), I'd expect Dependabot to output:
>```
> require github.com/pkg/errors v0.8.0
> replace github.com/pkg/errors => github.com/thepwagner/errors v0.8.1
>```

I actually disagree. My understanding of the _intent_ of the `replace` directive is that it's basically a "user override pin" that shouldn't be changed by automated tooling. I realize that some repos use it otherwise, but it's often either a [programmer mistake](https://github.com/dependabot/dependabot-core/pull/3576) or used to workaround specific problems with automated tooling... so if your automated tooling starts munging your workaround, then pretty soon you'll need a workaround for your workaround. 😀 

So in fact, your examples are exactly how I'd want and expect (based on the go docs/comments in the issue tracker) `go get` (and by extension Dependabot) to work. 

In fact, my mental model of Dependabot is that it merely automates the work of manually running `go get -u ./... && go mod tidy`. I get that it has a few additional tweaks such as ignore certain version updates, but in general the more we diverge from the typical `go` module update patterns, the more user confusion it'll probably create...

> Have you looked at go mod edit as an alternative to updateDependencyFile? It's not quite plug and play, but at least it's provided+supported by the SDK.

I haven't. And I doubt it will solve the root problem this PR is trying to solve where `go get` sees a mismatch between `go.mod` / `go.sum` and walks the dep tree and then blows up on the missing grpc module. Some of that may no longer be applicable when go lands lazy module loading, but very unclear at this point. 

Again, I say all this with less than 100% confidence in which outcomes of `go` machinery are expected vs bugs/inconsistencies that haven't been ironed out yet, but it's been very painful to debug this... The problematic repo that I am using for testing results in `go get -d` taking 20+ minutes for its initial cache population. So even to get this far has already taken more than a full day to narrow down what was happening in a semi-reproducible way.

Note also that I did see a big speed improvement when doing `go get -d <specific dep>` vs `go get -d`... I suspect the latter downloads all the modules specified in `go.mod`, vs the former only grabs a subset. So continuing with `go get -d` has some downsides.

jeffwidman:
> I'm not entirely sure if there is a technical reason that Dependabot manually updates the go.mod file though

I also wondered, and before I put up this PR I spent two hours spelunking through all the old commits to this folder. I never saw commit messages indicating it was written this way to workaround specific issues/bugs. Instead, I _suspect_ the following all played a role:
* this code first landed as a port of `dep`.
* the `go get`/`go mod` commands were pretty immature at the time of the port.
* the idiomatic ways to manage modules weren't very clear in the community, so the author sometimes just had to hack his way through it (leading to some surprising code)[https://github.com/dependabot/dependabot-core/pull/3576].
* this pipeline originally did _not_ git clone the repo, which I suspect resulted in more of a mindset of "surgically modify specific lines in specific files to replicate go".

Again, I'm just spitballin' here, so may be wrong.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -179,7 +161,12 @@ def run_go_get
 
           File.write(tmp_go_file, "package dummypkg\n") unless package
 
-          _, stderr, status = Open3.capture3(ENVIRONMENT, "go get -d")
+          # Use version pinning rather than `latest` just in case
+          # a new version gets released in the middle of our run.
+          version = "v" + dep.version.sub(/^v/i, "")
```
jeffwidman, line 166: Any idea what the `version.sub(/^v/i, "")` is protecting against? Never having written Ruby before, I'm not even quite sure what this is doing... 

I copied it from https://github.com/dependabot/dependabot-core/pull/3590/files#diff-491ccbbf0ae3cfe37dca8fba172104cc5466631d946daf9639deac37b014692eL158 but I noticed it's not used over in https://github.com/dependabot/dependabot-core/blob/7a2085bbee797ba7ffd0c09c83fc4d5481c859c2/go_modules/lib/dependabot/go_modules/update_checker.rb#L74

I'm unclear why the inconsistency between these two. It it a bug or intentional?

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jeffwidman: Oh, it turns out `go get` accepts multiple package arguments, so that makes life much simpler/more intuitive: https://golang.org/pkg/cmd/go/#hdr-Add_dependencies_to_current_module_and_install_them

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -179,7 +158,16 @@ def run_go_get
...
           File.write(tmp_go_file, "package dummypkg\n") unless package
 
-          _, stderr, status = Open3.capture3(ENVIRONMENT, "go get -d")
+          # TODO: go 1.18 will make `-d` the default behavior, so remove the flag then
+          command = "go get -d"
+          # `go get` accepts multiple packages, each separated by a space
+          dependencies.each do |dep|
+            # Use version pinning rather than `latest` just in case
+            # a new version gets released in the middle of our run.
+            version = "v" + dep.version.sub(/^v/i, "")
```
jeffwidman, line 167: Any idea what the `version.sub(/^v/i, "")` is protecting against? Never having written Ruby before, I'm not even quite sure what this is doing... 

I copied it from https://github.com/dependabot/dependabot-core/blob/7a2085bbee797ba7ffd0c09c83fc4d5481c859c2/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb#L158 but it's not used over in https://github.com/dependabot/dependabot-core/blob/7a2085bbee797ba7ffd0c09c83fc4d5481c859c2/go_modules/lib/dependabot/go_modules/update_checker.rb#L74

I'm unclear why the inconsistency between these two. It it a bug or intentional?

go_modules/spec/dependabot/go_modules/file_updater/go_mod_updater_spec.rb
```diff
@@ -421,7 +421,7 @@
       let(:previous_requirements) { [] }
 
       it "raises the correct error" do
-        error_class = Dependabot::DependencyFileNotResolvable
+        error_class = Dependabot::DependabotError
```
jeffwidman, line 424: So this makes the tests pass, but I kinda wonder if the problem isn't the test, but rather some code needs to be added to transform the `go get` failure from a catch-all `DependabotError` to a `DependencyFileNotResolvable` error... Any idea where that should happen?

jeffwidman: NVM, I found the problematic regex that needed updating, so I put back the test to before.

go_modules/spec/dependabot/go_modules/file_updater/go_mod_updater_spec.rb
```diff
@@ -236,12 +236,23 @@
...
+            end
+
+            it "updates the go.mod" do
+              # this is failing, but I'm not sure why.
+              # The code runs `go get -d github.com/googleapis/gnostic@v0.5.1`
+              # and then later runs `go mod tidy -e`.
+              # So I manually ran those in the test fixture, and it resulted in
+              # this line appearing. But when the test executes, this line is missing.
+              # I'm not sure why, and not sure how to run the Ruby debugger
+              # to step through it.
```
jeffwidman, line 252: The test was failing, and after looking into it, I understand why. So I replicated the steps manually, and this is what I expect.

But the test is still failing, so at this point unclear if the test is just plain wrong, or if there's some wrong code, or what.

Anyone else want to take a stab at it?

go_modules/spec/dependabot/go_modules/file_updater/go_mod_updater_spec.rb
```diff
@@ -515,7 +526,7 @@
         expect { updater.updated_go_sum_content }.
           to raise_error(error_class) do |error|
           expect(error.message).to include(
-            "go: github.com/deislabs/oras@v0.10.0 requires\n"\
+            "go: github.com/deislabs/oras@v0.9.0 requires\n"\
```
jeffwidman, line 529: This is confusing, but as best I can tell the error message is reported slightly differently by `go get`.

I'd feel a bit more comfortable if someone else can take a look at it too.

go_modules/spec/dependabot/go_modules/file_updater_spec.rb
```diff
@@ -104,9 +104,14 @@ module declares its path as: go.etcd.io/bbolt
...
         allow(Open3).to receive(:capture3).and_call_original
-        allow(Open3).to receive(:capture3).with(anything, "go get -d").and_return(["", stderr, exit_status])
+        allow(Open3).to receive(:capture3).with(anything, "go get -d github.com/etcd-io/bbolt@v1.3.5").and_return(["", stderr, exit_status])
       end
 
+      # This is failing and I'm not sure why...
+      # See also discussion here where it seems the author/reviewers agreed on one solution
+      # but then (confusingly) merged a different one w/o explaining why:
+      # https://github.com/dependabot/dependabot-core/pull/3482#discussion_r612343046
+      # So I'm unclear if this needs the regex, fixture, test, or something else to be updatd.
```
jeffwidman, line 114: Help?

I'm not really sure what needs to happen here...

jeffwidman:
I've taken this about as far as I can, need some help here on the tests side. I am hesitant to make them go green until we are comfortable that they're validating what we actually want 😀 .

jurre:
> > I'm not entirely sure if there is a technical reason that Dependabot manually updates the go.mod file though
> 
> I also wondered, and before I put up this PR I spent two hours spelunking through all the old commits to this folder. I never saw commit messages indicating it was written this way to workaround specific issues/bugs. Instead, I _suspect_ the following all played a role:
> 
> * this code first landed as a port of `dep`.
> * the `go get`/`go mod` commands were pretty immature at the time of the port.
> * the idiomatic ways to manage modules weren't very clear in the community, so the author sometimes just had to hack his way through it [leading to some surprising code](https://github.com/dependabot/dependabot-core/pull/3576).
> * this pipeline originally did _not_ `git clone` the repo, which I suspect resulted in more of a mindset of "surgically modify specific lines in specific files to replicate go".
> 
> Again, I'm just spitballin' here, so may be wrong.

Yep, that sounds like the most likely case to me as well! 👍

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jurre: It's a bit of defensive coding to make sure we never accidentally end up with a version like `vv1.1`.

Internally we store versions without the `v` prefix, so it's prepended to the version here, and in theory there might be cases where the version was parsed from a dependency file or the registry and the `v` hasn't been stripped out. I couldn't find any immediate cases where that would happen by following the code that is fed into this, but all our test-data does have the `v` prefix. 

When we run ` UpdateChecker#find_latest_resolvable_version`, that data comes from the `FileParser`, which already [strips out the `v` prefix](https://github.com/dependabot/dependabot-core/blob/7a2085bbee797ba7ffd0c09c83fc4d5481c859c2/go_modules/lib/dependabot/go_modules/file_parser.rb#L42) so it seems OK to omit it there.

It seems like the precaution it's not actually needed (anymore?), especially since `Dependabot::GoModules::Version` [strips out the prefix](https://github.com/dependabot/dependabot-core/blob/c3f01dff836d2031203c3870f772e42a42668b25/go_modules/lib/dependabot/go_modules/version.rb#L26). But, I'm always slightly nervous making these changes, as we don't have a thorough integration test-suite, and `dependency.version` can also be a straight-up string (yay, duck typing!)

I would opt to keep it in for now, the code is fairly simple and should have no meaningful performance implications. But we can investigate if we can get rid of it separately IMO.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -17,7 +17,7 @@ class GoModUpdater
         RESOLVABILITY_ERROR_REGEXES = [
           # The checksum in go.sum does not match the downloaded content
           /verifying .*: checksum mismatch/.freeze,
-          /go: .*: go.mod has post-v\d+ module path/
+          /go get: .*: go.mod has post-v\d+ module path/
```
jurre, line 20: Do we only care about the case where `go get` specifically raises this error? It seems OK to be more lenient when matching on the error?

jurre:
Thanks @jeffwidman! One of us will can dive into the failing tests but we'll need to make a bit of time for it, I hope to have a few slots next week, or maybe someone else from the team can get to it sooner. Really appreciate your thoroughness on this though 🎉

Small25 (approved):

Small25 (approved):

lastthyme (approved):

lastthyme (approved):

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jeffwidman: I'm fine either way, but the inconsistency between the two usages bothers me, because it's defensive but not... or at least appears that way. 

If I had to choose, I'd probably pick dropping it, because if Dependabot is broken for 24 hours it's not _that_ big of a deal... this isn't nuclear or hospital or banking code... and then if something does break you can track down the root cause rather than band-aiding it here.

Alternatively you could add an error log if the condition is ever true so that you defensively recover from the error, but also get notified that there's an error likely indicating a bug elsewhere in the code. But that gets a little complex, and given the impacts aren't that big of deal (as long as you've got observability metrics that alert when there's a spike in the percent of your `go_modules` jobs that are failing).

Regardless, agreed that it's out of scope for this PR, so I filed https://github.com/dependabot/dependabot-core/issues/3612 to track it.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jeffwidman: Agree! It should match whether emitted by `go mod` tooling or `go get` tooling.

I assume this would match both in ruby? But still tight enough to only match _either_ `go:` or `go get:` and nothing else?

```suggestion
          /go (?:get)?: .*: go.mod has post-v\d+ module path/
```

jeffwidman:
> Thanks @jeffwidman! One of us will can dive into the failing tests but we'll need to make a bit of time for it, I hope to have a few slots next week, or maybe someone else from the team can get to it sooner. Really appreciate your thoroughness on this though 🎉

Fantastic! I'll still be around to help out as needed, but will let you folks drive it across the line. And to be clear--I am _very_ okay if you choose to go a different route with the design after investigating the test failures, my feelings will not be hurt. 😀  

I just want the problem that we're hitting in #3526 to be resolved, w/o having to wait for lazy-loading in `go` `1.17` (especially as I'm not even sure that will fix it!). Another week or two is fine though.

One other thing that I'm curious on is benchmarks... I'd be very curious if this speeds things up quite a bit. I'm certainly seeing that in my local runs, but that's also running against a dep graph that has circular loops, which may simply be a special case that `go get -d <dep@version>` seems to short-circuit.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
thepwagner: My use case here centres around long-lived forks:

As a worked example, let's say I want to add a feature to GitHub's API. To do this, I'm going to make a private fork of  https://github.com/google/go-github , and add my new functionality.
I don't want to leak the new feature into `go-github` before it's announced, so in the interim I'm going to use a `replace` directory to prefer my fork as the source of this dependency.

With my project in this setup, what happens if `google/go-github` makes a release?

Today, Dependabot is going to see the release of `google/go-github` and propose a PR performing that upgrade. Since it doesn't understand the `replace` directive, the resulting PR will be syntactically valid but semantically meaningless: merge or close, my project will still be using `thepwagner/go-github` version specified in the replace directive.

I don't think this PR will affect that behaviour.

---

What I'd prefer [in the scenario of a non-local replacement](https://github.com/thepwagner/action-update-go/tree/main/gomodules/testdata/replace) is:
1. Dependabot should consider the replaced dependency, (e.g. [`github.com/thepwagner/go-github`](https://github.com/thepwagner/action-update-go/blob/2ba2b9a3d49a3285b327622fbd5daccb4fe64e19/gomodules/dependencies_test.go#L33))
2. Dependabot should continue to propose updates, if available, from the fork. Ignoring the upstream.
3. Dependabot should update the [`replace` directive](https://github.com/thepwagner/action-update-go/blob/2ba2b9a3d49a3285b327622fbd5daccb4fe64e19/gomodules/applyupdate_test.go#L138-L155) and leave the `require` directives untouched (since they have no impact).

This isn't what `go get -u` does, but as a maintainer of a long-lived for it's what I'd like.

While this PR doesn't affect this scenario, switching from hand-updating the file to `go get -d` means (3) is potentially further than it was before. That was the flag I wanted to raise.

> go get -d <dep>@<version> is a clearly recommended way to bump to a particular version, so why not just use it if it solves the problems?

Agreed, SGTM!

jeffwidman:
Circling back on this--any chance someone on the team has time to look into the failing / confusing tests? I just fixed the merge conflicts... 

I've got a couple of teams at my day job who are eager to see this land because dependabot is failing on half their repos due to #3526 (which would be fixed by this PR).

jeffwidman:
Gentle nudge on this one. Anything I can do to help move this forward?

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jurre: Yeah that'd work, or maybe even just this?:

```ruby
/go *. go.mod has post-v\d+ module path/
```

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -180,7 +159,16 @@ def run_go_get
...
+          # TODO: go 1.18 will make `-d` the default behavior, so remove the flag then
+          command = +"go get -d"
+          # `go get` accepts multiple packages, each separated by a space
+          dependencies.each do |dep|
+            # Use version pinning rather than `latest` just in case
+            # a new version gets released in the middle of our run.
+            version = "v" + dep.version.sub(/^v/i, "")
+            command << " #{dep.name}@#{version}"
+          end
+          _, stderr, status = Open3.capture3(ENVIRONMENT, command)
```
jurre, line 171: `spec/dependabot/go_modules/file_updater/go_mod_updater_spec.rb:245` is failing because we now no longer run `go get -d` on all the modules, so basically we're only running the code for `update_go_mod` (but now using `go get instead of our custom helper), we still need to run a bare `go get -d` to verify that the packages we have downloaded have the module that we expect to find etc.

I verified the tests pass after adding a `go get -d` after this command, but what might be nice is keeping the original `def run_go_get` method, and moving the contents of this method into `def update_go_mod`

I think the tests should pass after that

go_modules/spec/dependabot/go_modules/file_updater_spec.rb
jurre: I think this is the same problem as https://github.com/dependabot/dependabot-core/pull/3590/files#r632456405

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jeffwidman: For now, I'd probably lean toward the slightly more restrictive regex, as `*.` has too often bitten me unexpectedly. It's easy enough to relax it later, but this at least communicates to the reader/dev what the expected matching is.

go_modules/spec/dependabot/go_modules/file_updater_spec.rb
```diff
@@ -104,9 +104,14 @@ module declares its path as: go.etcd.io/bbolt
...
         exit_status = double(success?: false)
         allow(Open3).to receive(:capture3).and_call_original
-        allow(Open3).to receive(:capture3).with(anything, "go get -d").and_return(["", stderr, exit_status])
+        allow(Open3).to receive(:capture3).with(anything, "go get -d github.com/etcd-io/bbolt@v1.3.5").and_return(["", stderr, exit_status])
       end
 
+      # This is failing and I'm not sure why...
+      # See also discussion here where it seems the author/reviewers agreed on one solution
+      # but then (confusingly) merged a different one w/o explaining why:
+      # https://github.com/dependabot/dependabot-core/pull/3482#discussion_r612343046
```
jeffwidman, line 113: This PR issue got resolved in https://github.com/dependabot/dependabot-core/pull/3660/commits/5bfdf42fb0ce6655f0fad6f93afdcc82dab01ebe#diff-491ccbbf0ae3cfe37dca8fba172104cc5466631d946daf9639deac37b014692eR44

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jeffwidman: I broke this out as https://github.com/dependabot/dependabot-core/pull/3732 for simplicity.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jeffwidman: Ah, so that was intentional... 

I'd thought the generic `go get -d` wasn't needed because we now rely on `go get lib@version` to _be smart enough to know how to upgrade a particular library_ and if some other part of the dep tree is hosed, then we let `go get -d lib@version` decide whether that is relevant or not. I don't expect (or want) dependabot to validate my _existing_ dep tree, I only expect it to validate that version bumps don't put the dep tree in a worse state.

I still think that actually. 

So let's do this, I'll break this PR in two. The first will keep this generic `go get -d` command, and simply swap out the custom helper for `go get` and then the second will be a discussion of whether this generic command is within the scope of what Dependabot should be doing. 

That should keep the majority of this PR in a simple, non-controversial PR, and then the second PR (which will be more controversial) can be much more easily reasoned about/researched/discussed. 

I'm also reasonably certain this first PR will _not_ solve #3526 which was my original impetus for all this work. That'd require the second PR.

jeffwidman:
Converting to Draft for now... I'm breaking this out as 3 PRs:
* Add the helpful inline comment https://github.com/dependabot/dependabot-core/pull/3732
* The implementation swap: https://github.com/dependabot/dependabot-core/pull/3733
* Removing the generic `go get -d` call, since arguably it's not needed and currently causes more harm than good (this PR will come later).

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jurre: Yeah I get what you're saying, however:

>  I only expect it to validate that version bumps don't put the dep tree in a worse state.

I think that running the generic `go get -d` still protects against this, for the case the failing test was validating, it prevents updating to a version when the package has been renamed, right?

Maybe it is an edge-case we should not worry about, assume peoples CI will fail and they'll figure out what the issue is and update the import statements etc manually, but I definitely appreciate you pulling these out into separate PRs so we can discuss more focussed 🙇

go_modules/spec/dependabot/go_modules/file_updater/go_mod_updater_spec.rb
jeffwidman: Yep, this is def just an error message getting reported differently, because the same change was necessary over in https://github.com/dependabot/dependabot-core/pull/3733/files#diff-983a4baf78264a5fd352fd21d2ba73a5a6a594bbb6f7bee2c912c56f10df772cR518

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jeffwidman: Also, a few notes to myself when I get time to dig into this:
1. IIRC, the other reason this was originally here was because we'd manually munged `go.mod` w/o touching `go.sum`, so we needed this `go get` to pickup those `go.sum` changes. (This could have also been done via `go mod tidy`, but `go get` was handling it well enough). This is why I originally thought this was no longer needed.
2. The transitory resolving stuff will likely get cleaned up a lot more as part of [lazy loading in `go 1.17`](https://tip.golang.org/doc/go1.17#go-command).
3. There's still the open question though of should dependabot fail when something deep upstream fails... eg, this started failing yesterday 4 steps up our dep chain: https://github.com/hashicorp/go-discover/issues/172. I can see valid arguments both pro/con on this.

jeffwidman:
Closing for now, this will need complete re-evaluation once `go 1.17` drops later this month with the new lazy-loading feature.

PR #3593, opened 2021-04-29
WIP: proposal to handle multi-length semver ranges

feelepxyz:
I ended up shaving the yak while reviewing and getting my head around variable length semver ranges: https://github.com/dependabot/dependabot-core/pull/3588

Opening to gather some feedback on this and to check weather these test cases are actually legit and should be covered. I'm less sure about the non-numeric handling.

### DIFF
--- a/common/spec/dependabot/config/ignore_condition_spec.rb
+++ b/common/spec/dependabot/config/ignore_condition_spec.rb
@@ -227,6 +227,12 @@ RSpec.describe Dependabot::Config::IgnoreCondition do
         context "with ignore_major_versions" do
           let(:update_types) { ["version-update:semver-major"] }
 
+          # TODO: We probably want to ignore all major version, e.g. v2, v3 etc
+          it "ignores expected versions" do
+            expect_ignored(["2.0.0", "2.0.0.a", "2.9.9"])
+            expect_allowed([dependency_version, "1.2.2", "1.2.0.a", "3.0.0"])
+          end
+
           it "returns the expected range" do
             expect(ignored_versions).to eq([">= 2.a, < 3"])
           end
--- a/common/spec/dependabot/config/ignore_condition_spec.rb
+++ b/common/spec/dependabot/config/ignore_condition_spec.rb
@@ -88,38 +88,6 @@ RSpec.describe Dependabot::Config::IgnoreCondition do
         it "returns the expected range" do
           expect(ignored_versions).to eq([">= 1.2.3.1.a, < 1.3"])
         end
-
-        context "with a minor dependency version" do
-          let(:dependency_version) { "1.2" }
-
-          it "ignores expected updates" do
-            expect_ignored(["1.2.1.a", "1.2.1.1"])
-            expect_allowed(minor_upgrades + major_upgrades)
-            expect_allowed([dependency_version, "1.2.0"])
-          end
-        end
-
-        context "with a single major dependency version" do
-          let(:dependency_version) { "1" }
-          let(:patch_upgrades) { %w(1.0.0.1 1.0.2 1.0.5 1.0.4-rc0) }
-          let(:minor_upgrades) { %w(1.1 1.2.0 1.3 1.4.0) }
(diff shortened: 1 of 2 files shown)


common/spec/dependabot/config/ignore_condition_spec.rb
```diff
@@ -217,6 +259,13 @@ def expect_ignored(versions)
             expect(ignored_versions).to eq([">= Finchley.a, < Finchley.999999"])
           end
         end
+
+        context "with ignore_major_versions" do
+          let(:update_types) { ["version-update:semver-major"] }
+          it "returns the expected range" do
+            expect(ignored_versions).to eq([">= Finchley.999999, < 999999"])
```
feelepxyz, line 266: I don't really know what this means tbh

common/spec/dependabot/config/ignore_condition_spec.rb
```diff
@@ -86,7 +86,39 @@ def expect_ignored(versions)
...
+            expect_allowed([dependency_version, "1.2.0"])
+          end
+        end
+
+        context "with a single major dependency version" do
+          let(:dependency_version) { "1" }
+          let(:patch_upgrades) { %w(1.0.0.1 1.0.2 1.0.5 1.0.4-rc0) }
+          let(:minor_upgrades) { %w(1.1 1.2.0 1.3 1.4.0) }
+
+          it "does not attempt to ignore any versions" do
```
mctofu, line 107:
```suggestion
          it "ignores expected updates" do
```

This didn't seem right since we are testing that patches are ignored.

common/spec/dependabot/config/ignore_condition_spec.rb
```diff
@@ -86,7 +86,39 @@ def expect_ignored(versions)
...
+            expect_ignored(patch_upgrades)
+            expect_allowed(minor_upgrades + major_upgrades)
+            expect_allowed([dependency_version])
+          end
+        end
+
+        context "with a patched patch dependency version" do
```
mctofu, line 114: 👍

common/spec/dependabot/config/ignore_condition_spec.rb
```diff
@@ -116,6 +148,15 @@ def expect_ignored(versions)
         it "returns the expected range" do
           expect(ignored_versions).to eq([">= 2.a, < 3"])
         end
+
+        context "with a single major dependency version" do
+          let(:dependency_version) { "1" }
+
+          it "ignores patch updates" do
```
mctofu, line 155:
```suggestion
          it "ignores expected versions" do
```

common/spec/dependabot/config/ignore_condition_spec.rb
```diff
@@ -116,6 +148,15 @@ def expect_ignored(versions)
...
+            expect_ignored(["2.0.0", "2.0.0.a"])
+            expect_allowed([dependency_version, "1.2.2", "1.2.0.a", "3.0.0"])
```
mctofu, line 157: This looks like existing behavior but do we actually want to allow 3? I'd expect if I ignore major versions I won't get v2, v3, v4, etc.

mctofu (approved):
Nice! These test cases look right to me. Had a minor few nits about test names. Also a question about how the ignore major works but I think that can be a separate issue if we want to adjust it.

common/spec/dependabot/config/ignore_condition_spec.rb
feelepxyz: Good spot! I just spotted this when testing in the wild: https://github.com/github/dependabot-updates/issues/1350#issuecomment-829374042

I think you're right, we should ignore all major versions. We have the current behaviour when ignore from PRs as you're talking about a specific major version.

feelepxyz:
@mctofu shall I merge into your branch?

mctofu:
> @mctofu ok to merge into your branch?

@feelepxyz 👍 Let's do it!

PR #3610, opened 2021-04-30
`version-update:semver-major` ignores all major version updates

mctofu:
Removes the upper bound when generating ranges for `version-update:semver-major` so that we ignore all upcoming major versions instead of only the next one.

### DIFF
--- a/common/lib/dependabot/config/ignore_condition.rb
+++ b/common/lib/dependabot/config/ignore_condition.rb
@@ -76,11 +76,9 @@ module Dependabot
 
         version_parts = version.split(".")
         lower_parts = [version_parts[0].to_i + 1] + ["a"]
-        upper_parts = [version_parts[0].to_i + 2]
         lower_bound = ">= #{lower_parts.join('.')}"
-        upper_bound = "< #{upper_parts.join('.')}"
 
-        ["#{lower_bound}, #{upper_bound}"]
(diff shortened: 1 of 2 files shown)


common/lib/dependabot/config/ignore_condition.rb
```diff
@@ -76,11 +76,9 @@ def ignore_major(version)
 
         version_parts = version.split(".")
         lower_parts = [version_parts[0].to_i + 1] + ["a"]
-        upper_parts = [version_parts[0].to_i + 2]
         lower_bound = ">= #{lower_parts.join('.')}"
-        upper_bound = "< #{upper_parts.join('.')}"
 
-        ["#{lower_bound}, #{upper_bound}"]
+        [lower_bound.to_s]
```
thepwagner, line 81:
```suggestion
        [lower_bound]
```

Nit: the `.to_s` seems superflous, we initialized the string on L79 so it's probably still a string.

Noice: I dig keeping the `lower_bound` naming+structure for consistency.

thepwagner (approved):

common/lib/dependabot/config/ignore_condition.rb
mctofu: Nice catch! That was an auto rubocop fix from `["#{lower_bound}"]` so I didn't think about it.

Issue #3619, opened 2021-05-01
setup.cfg inline comments are not ignored


adehad:
**Package ecosystem**
pip
**Package manager version**
21.0.1
**Language version**
Python 3.x
**Manifest location and content prior to update**
https://github.com/pycontribs/jira/blob/7fa3a454c08a14e2d7d7670fcfa87482e16936ba/setup.cfg 
**dependabot.yml content**
```yaml
version: 2
updates:
  - package-ecosystem: pip
    directory: "/"
    schedule:
      interval: daily
    open-pull-requests-limit: 10
    target-branch: master
```
**Updated dependency**
N/A
**What you expected to see, versus what you actually saw**
Saw:
```
Dependabot can't evaluate your Python dependency files.
As a result, Dependabot couldn't check whether any of your dependencies are out-of-date.
The error Dependabot encountered was:
InstallationError("Invalid requirement: 'pre-commit  # MIT'")
``` 

Expected:
Ignores comments successfully

**Native package manager behavior**
Ignores comments successfully
**Images of the diff or a link to the PR, issue or logs**
N/A
**🕹 Bonus points: Smallest manifest that reproduces the issue**

setup.py
```python
# !/usr/bin/env python
import setuptools


if __name__ == "__main__":
    setuptools.setup(use_scm_version=True)
```

setup.cfg
```ini
[metadata]
name = jira
author = Ben Speakmon
author_email = ben.speakmon@gmail.com
maintainer = Sorin Sbarnea
maintainer_email = sorin.sbarnea@gmail.com
summary = Python library for interacting with Jira via REST APIs.
long_description = nothing
url = https://github.com/pycontribs/jira
project_urls =
    Bug Tracker = https://github.com/pycontribs/jira/issues
requires_python = >=3.5
platforms = any
license = BSD
classifiers =
    Development Status :: 5 - Production/Stable
keywords = api, atlassian, jira, rest, web

[files]
packages =
    jira

[options]
use_scm_version = True
python_requires = >=3.5
packages = find:
include_package_data = True
zip_safe = False
# These are required during `setup.py` run:
setup_requires =
    setuptools_scm >= 1.15.0
    setuptools_scm_git_archive >= 1.0
install_requires =
    defusedxml
    keyring
    requests-oauthlib>=1.1.0
    requests>=2.10.0
    requests_toolbelt
    setuptools>=20.10.1  # probably will break too

[options.extras_require]
test =
    pre-commit  # MIT
```

## Details
[...]

PR #3624, opened 2021-05-03
Python: allow comments when parsing setup.cfg

jurre:
Currently any comments in a setup.cfg file will raise an error in
Dependabot. Pip handles this fine, so let's make sure Dependabot does as
well.

Fixes https://github.com/dependabot/dependabot-core/issues/3619


feelepxyz (approved):
LGTM 👍  thanks for adding the pip internal ref

feelepxyz (approved):
LGTM 👍  thanks for adding the pip internal ref

PR #3637, opened 2021-05-04
Ignored update types: handle nil dep version

feelepxyz:
Handle a `nil` dependency version when computing ignored versions from
ignored update types.

When we can't reliably figure out the exact version of a dependency
(e.g. it doesn't have a lockfile), we just leave the
`dependency.version` as nil.

### DIFF
--- a/common/spec/dependabot/config/update_config_spec.rb
+++ b/common/spec/dependabot/config/update_config_spec.rb
@@ -148,6 +148,26 @@ RSpec.describe Dependabot::Config::UpdateConfig do
         end
       end
     end
+
+    context "when the dependency version isn't known" do
+      let(:dependency) do
+        Dependabot::Dependency.new(
+          name: "actions/checkout",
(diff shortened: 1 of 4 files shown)


common/spec/dependabot/config/update_config_spec.rb
```diff
@@ -148,6 +148,26 @@
...
       end
     end
+
+    context "when the dependency version isn't known" do
+      let(:dependency) do
+        Dependabot::Dependency.new(
+          name: "actions/checkout",
+          requirements: [],
+          version: nil,
+          package_manager: "go_modules"
```
mctofu, line 158:
```suggestion
          package_manager: "github_actions"
```

Doesn't really matter but we could make this match the dependency name unless it specifically needs to be `go_modules`?

mctofu (approved):

thepwagner (approved):

PR #3638, opened 2021-05-04
Elm: Drop support for Elm 0.18

jurre:
This drops support for Elm 0.18, and therefore also removes dealing with
`elm-package.json` files.

Elm 0.18 is 4 years old now, and it no longer installs using `npm`, because the
binary is hosted on bintray, which is no more.

Instead of patching up the installation for this, let's remove support for this
old version, it is not widely used at all, keeping support for it around does not
seem worth the overhead.

The test coverage for the new `elm.json` files is actually not
fantastic, I have converted a few tests for `elm-package.json` over, but
I found it hard to justify spending more time on this, as this code has
been running in production for a while so should be stable enough for
the low number of users it has.


thepwagner (approved):
Awesome cleanup - I probably would have dropped the binary but left `elm-package.json` lingering, this is much better!

I wondered about identifying an elm expert from `git log --pretty="%ae" elm | sort | uniq` for feedback, but this seems to be mostly internal. :shipit:

Dockerfile
```diff
@@ -99,17 +99,16 @@ RUN curl -sL https://deb.nodesource.com/setup_14.x | bash - \
...
+RUN wget "https://github.com/elm/compiler/releases/download/0.19.0/binaries-for-linux.tar.gz" \
   && tar xzf binaries-for-linux.tar.gz \
   && mv elm /usr/local/bin/elm19 \
   && rm -f binaries-for-linux.tar.gz \
   && rm -rf ~/.npm
 
-# NOTE: This is a hack to get around the fact that elm 18 fails to install with
-# npm 7, we should look into deprecating elm 18
+# NOTE: This was a hack to get around the fact that elm 18 failed to install with
+# npm 7, we should look into installing the latest version of node + npm
```
feelepxyz, line 111: Nice! We could move this `RUN` up to the `JAVASCRIPT` section but can follow up with this and update npm

jurre: Yeah, good call, I'll move it over 👍

elm/spec/dependabot/elm/file_updater_spec.rb
```diff
@@ -69,58 +62,6 @@
...
-    it "doesn't store the files permanently" do
-      expect { updated_files }.to_not(change { Dir.entries(tmp_path) })
-    end
-
-    it "returns DependencyFile objects" do
-      updated_files.each { |f| expect(f).to be_a(Dependabot::DependencyFile) }
-    end
-
-    it { expect { updated_files }.to_not output.to_stdout }
-    its(:length) { is_expected.to eq(1) }
```
feelepxyz, line 81: Should we bring these back? Look like they are just testing the shape of `updated_dependency_files`

jurre: They weren't testing the code that we've been running for most users in production. Moving tests over proved to be fairly painful after having done a few for the `UpdateChecker`, I mostly considered it as "also wasn't tested before" and was unsure it was worth spending more time on this given the low usage.

jurre: Aye, yeah this was pretty easy to patch up, good call-out 👍

feelepxyz (approved):
:shipit:

PR #3640, opened 2021-05-04
GitHub PullRequestCreator: prepend refs/

thepwagner:
This is a workaround to the proposed upstream fix, https://github.com/octokit/octokit.rb/pull/1336.

Consider a dependency named `github.com/football-refs/league-rules`.
As part of an upgrade, Dependabot might attempt a branch like `dependabot/go_modules/github.com/football-refs/league-rules/2.0.0`.

Since that matches `refs/`, it won't automatically have `refs/` prepended, ultimately resulting in a `Octokit::UnprocessableEntity`.

### DIFF
--- a/common/lib/dependabot/pull_request_creator/github.rb
+++ b/common/lib/dependabot/pull_request_creator/github.rb
@@ -226,12 +226,12 @@ module Dependabot
       end
 
       def create_branch(commit)
-        ref = "heads/#{branch_name}"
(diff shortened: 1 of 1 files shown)


jurre (approved):
Given this is tested in octokit and will be patched there, seems overkill to repeat that test-case here. :shipit:

brrygrdn (approved):
👍🏻

PR #3642, opened 2021-05-05
Create changelog from merge commits

feelepxyz:
Build the changelog entries from merge commit PR details instead of the individual commits, automating attribution to non-team members and including a link to the PR.

Example:

```
## v0.146.0, 5 May 2021

- fix(Go mod): detect when remote end hangs up [#3469](https://github.com/dependabot/dependabot-core/pull/3469)
- Final pass on bundler1 test cruft [#3466](https://github.com/dependabot/dependabot-core/pull/3466)
- Handle repo not found errors for go modules [#3456](https://github.com/dependabot/dependabot-core/pull/3456)
- Disabled poetry experimental new installer (@honnix) [#3459](https://github.com/dependabot/dependabot-core/pull/3459)
- Implement delete/create action in GitLab Client (@jerbob92) [#3414](https://github.com/dependabot/dependabot-core/pull/3414)
```

### DIFF
--- a/bin/bump-version.rb
+++ b/bin/bump-version.rb
@@ -9,7 +9,7 @@ component = ARGV[0].to_sym
 
 unless `which gh` && $?.success?
   puts "Please install the gh cli: brew install gh"
-  exut 1
+  exit 1
 end
 
 unless `gh auth status > /dev/null 2>&1` && $?.success?
--- a/bin/bump-version.rb
+++ b/bin/bump-version.rb
@@ -49,12 +49,20 @@ merge_subjects = commit_subjects.select { |s| s.start_with?("Merge pull request 
 pr_numbers = merge_subjects.map { |s| s.match(/#(\d+)/)[1].to_i }
 puts "⏳ fetching pull request details"
 pr_details = pr_numbers.map do |pr_number|
-  `gh pr view #{pr_number} --json title,author --jq ".title,.author.login"`
-end.map { |s| s.split("\n") }
+  pr_details = `gh pr view #{pr_number} --json title,author --jq ".title,.author.login"`
+  title, author = pr_details.split("\n").map(&:strip)
+  {
+    title: title,
+    author: author,
+    number: pr_number,
+    link: "https://github.com/dependabot/dependabot-core/pull/#{pr_number}"
+  }
+end
 
-proposed_changes = pr_details.map do |title, author|
-  line = "- #{title}"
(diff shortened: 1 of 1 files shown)


thepwagner:
> wanted to get feedback on the idea of using gh to do this,

👍 

This (PR titles) is what I tend to do manually when I'm not familiar enough with a diff to summarize.

Aside: https://github.com/atlassian/changesets / https://github.com/changesets/action provides some neat tooling here too, but that's a much bigger leap than "CHANGELOG.md, but easier". :shipit: 

--- 

Alternatively, what about a process that encourages `CHANGELOG.md` be modified as part of each PR? That solves the summarization problem, and lets contributors decide themselves if they want a shoutout, etc...

Meh, that's manual and tedious - this is ✨magic✨ .

feelepxyz:
> Aside: https://github.com/atlassian/changesets / https://github.com/changesets/action provides some neat tooling here too, but that's a much bigger leap than "CHANGELOG.md, but easier". :shipit:

This looks cool! Up for testing it out.

Yeah keen to improve the current flow a bit as I frequently reach for the PRs to reach for author/title.

bin/bump-version.rb
```diff
@@ -7,6 +7,19 @@
 end
 component = ARGV[0].to_sym
 
+unless `which gh` && $?.success?
```
jurre, line 10: Should we also check if jq is installed? I noticed it was used further down in the script

jurre (approved):
I dig it!

bin/bump-version.rb
feelepxyz: Looks like this isn't needed as it's bundled in `gh` somehow

jurre: Sweet

brrygrdn:
👍🏻  I like this approach, as _thepwagner notes this is automating what I already do when I don't have context on the change. I think it's a good forcing function to encourage expressive Pull Request naming by authors and their reviewers over time.

bin/bump-version.rb
```diff
@@ -7,6 +7,19 @@
...
+  puts "Please install the gh cli: brew install gh"
+  exut 1
```
Nishnha, line 12: nit typo

feelepxyz: nice catch!

feelepxyz, line 12:
```suggestion
  exit 1
```

PR #3644, opened 2021-05-05
Log when all versions are ignored

feelepxyz:


feelepxyz:
Ehh not sure about this approach, it looks like call these version methods a lot so end up with duplicate logs:

```
=== numpy (1.11.0)
 => checking for updates 1/1
I, [2021-05-05T17:38:20.414118 #98032]  INFO -- : All versions for numpy were ignored
 => latest available version is
I, [2021-05-05T17:38:21.329798 #98032]  INFO -- : All versions for numpy were ignored
 => latest allowed version is 1.11.0
I, [2021-05-05T17:38:21.438065 #98032]  INFO -- : All versions for numpy were ignored
I, [2021-05-05T17:38:21.553390 #98032]  INFO -- : All versions for numpy were ignored
I, [2021-05-05T17:38:21.695491 #98032]  INFO -- : All versions for numpy were ignored
I, [2021-05-05T17:38:21.811033 #98032]  INFO -- : All versions for numpy were ignored
    (no update possible 🙅‍♀️)
```

Also for some package managers (python) `the latest_version` becomes nil if we filter out lower versions 🤦‍♂️

PR #3648, opened 2021-05-05
Tidy up the terraform tests

xlgmokha:
# Why is this needed?

To update the terraform tests to follow patterns used by the other eco-system
projects.

## What does this do?

It updates the terraform unit tests to use the `project` fixtures pattern
used in the other eco system unit test suites.


xlgmokha:
@nishnha do you mind reviewing this change?

terraform/.rubocop.yml
```diff
@@ -1 +1,7 @@
 inherit_from: ../.rubocop.yml
+
+Layout/FirstArrayElementIndentation:
```
thepwagner, line 3: Question: should we just do this in `../.rubocop.yml`? `wc -l */.rubocop.yml` suggests modules tend not to customize.
If doing so would make a big diff of lint fixes, maybe handle as a separate cleanup PR?

terraform/spec/dependabot/terraform/file_parser_spec.rb
```diff
@@ -9,336 +9,275 @@
...
+
+  describe "#parse" do
+    subject { parser.parse }
+
+    context "with an invalid registry source" do
+      let(:files) do
+        [
+          Dependabot::DependencyFile.new(
+            name: "main.tf",
+            content: fixture("projects", "invalid_registry", "main.tf")
```
thepwagner, line 25: Noice: project fixtures!

Idea: maybe `project_name` could be a scoped variable?
I think the contents of `files` could be derived from something like `Dir.glob("fixtures/#{project_name}/*").map{ |p| DependencyFile.new(...) }`. That's consistent with what the [FileFetcher would do](https://github.com/dependabot/dependabot-core/blob/5e3d19ed811c76f70dda0b2ff3e6aeeaf856fe9e/terraform/lib/dependabot/terraform/file_fetcher.rb#L35) - return all the `.tf` files.

thepwagner (approved):
:shipit: as spec-only and the specs pass.

I'm a little concerned about the rubocop deviations: I _think_ we prefer consistency across modules, but I'm not sure 🤷 .

The idea about generating `let(:files)` based on the project name is just an idea - safe to ignore.

terraform/.rubocop.yml
xlgmokha: Updated here https://github.com/dependabot/dependabot-core/pull/3649

terraform/spec/dependabot/terraform/file_parser_spec.rb
xlgmokha: We found a helper called `project_dependency_files` that seems to accomplish this same goal. Thank you for calling out this suggestion.

PR #3652, opened 2021-05-05
Update providers versions in terraform lock files

xlgmokha:
# Why is this needed?

To be able to update providers to the latest version.

## What does this do?

Nothing yet. It adds a test that reproduces the desired outcome.

## Type of change

- [X] New feature (non-breaking change which adds functionality)

### DIFF
--- a/terraform/spec/dependabot/terraform/file_updater_spec.rb
+++ b/terraform/spec/dependabot/terraform/file_updater_spec.rb
@@ -50,7 +50,7 @@ provider "registry.terraform.io/hashicorp/aws" {
   version     = "3.38.0"
   constraints = "~> 3.27"
   hashes = [
-    "h1:qKEjN/EM56XT46vGY33eoq7nD6JuGqRqFp7tkzTrRM0=",
+    "h1:ARuS11ThIcUfmAQKWNXGPLOa1GheaIwkeCnMh9Mjvao=",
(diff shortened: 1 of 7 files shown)


terraform/spec/dependabot/terraform/file_updater_spec.rb
```diff
@@ -42,29 +42,7 @@
...
-    "zh:86536598796ba65539816f08351ac0ab32988ab84fa8f100049579996fafc800",
-    "zh:b9985c64f0f0b5bafb7067a60381fd807f7c3dd952c5d9f531385e464867bdd5",
-    "zh:c19c692896469724c6320fa7d87532ec3935e14e0e0fa0a8a0f1cf28ae7a0b0a",
-    "zh:cb8b14f246953a275ada562f5275a0d1a4938b7d20597e62fabe264012410f53",
-    "zh:cdbfa0ad87ff4d7451cfb89e53692a651d4c9cadece6845e60d986fd454b52b3",
-    "zh:ed5c4c8ae5adda37942bb15ef058c0811a95cb4c87259ae822627756dcb90efc",
-  ]
-}
-        LOCKFILE
+        expect(lock_file.content).to eql(fixture("projects/aws/.terraform.lock.hcl.expected"))
```
jurre, line 45: Tbh I preferred having this inlined in the test file, it makes it much clearer what we're testing.

jurre:
@xlgmokha I think we can build off of this PR for our implementation after a rebase? The code and approach overall looks good to me!

PR #3658, opened 2021-05-06
Filter lower versions when checking ignored versions

feelepxyz:
Changes the condition that raises `AllVersionsIgnored` when all new
versions are ignored when there are some new versions, filtering out
lower versions.

I opted to do this in the condition to not introduce any breaking
changes in the version checkers.

I've also explored always filtering lower version before filtering
ignored versions but this has been implemented inconsistently across
package managers so could cause a bunch of subtle bugs.

Most package managers only filter lower versions when checking for the
lowest security fix version, and rely on the latest version being set,
even if it's the current version. 🤦‍♂️

I wonder if it would make more sense to do a separate refactor to start
using the `UpdateConfig` and `IgnoreCondition` classes in a more generic
way in the base update checker instead of doing the same filtering in each
package manager.

## TODO:
- [x] Add some test cases


hex/lib/dependabot/hex/update_checker.rb
```diff
@@ -221,13 +221,23 @@ def latest_release_from_hex_registry
...
+            end
 
             filtered.max
           end
       end
       # rubocop:enable Metrics/PerceivedComplexity
 
+      def filter_lower_versions(versions_array)
+        versions_array.select do |version|
+          next unless current_version
```
thepwagner, line 235:
```suggestion
        return [] unless current_version

        versions_array.select do |version|
```

Nit: `current_version` isn't going to change inside the loop, so check it once and exit early.

thepwagner: Nit: should we care about the memoization of `current_version`? We could shave some nanos by storing it outside the loop.

hex/lib/dependabot/hex/update_checker.rb
```diff
@@ -246,13 +256,14 @@ def hex_registry_response
...
+      end
+
       def wants_prerelease?
-        current_version = dependency.version
-        if current_version &&
-           version_class.correct?(current_version) &&
-           version_class.new(current_version).prerelease?
-          return true
-        end
+        return true if current_version&.prerelease?
```
thepwagner, line 266: Noice: this cleanup!

thepwagner (approved):
Diff LGTM.

Can you walk me through a case where this is new behaviour is desirable?
Maybe we're ignoring `>= 0`, but are currently on the latest version.

feelepxyz:
@thepwagner I've attempted to fix this issue where currently we won't raise `AllVersionsIgnored ` when ignoring update types as these only filter new versions:

Pre:
Screenshot 2021-05-06 at 14 25 51

Post:
Screenshot 2021-05-06 at 14 25 03

So now, if the ignore conditions end up rejecting all new versions we raise.

common/spec/dependabot/git_commit_checker_spec.rb
```diff
@@ -944,6 +963,18 @@
           end
         end
 
+        context "all later versions ignored" do
```
mctofu, line 966: This test is currently not passing because `current_version` ends up being nil

mctofu:
> [ ] Add some test cases

I made it through about half of the ecosystems. The `GitCommitChecker` is not yet passing but I haven't dug in yet to see if my test setup is wrong or if target code needs some adjustment.

jasonrudolph (approved):
> I added three test cases to each of the ecosystems to check that when raise_on_ignored is enabled...

I reviewed all the tests and they look good.

Thanks so much for taking the time to [commit the tests separately for each ecosystem](https://github.com/dependabot/dependabot-core/compare/f034763fa7938ffe039768fd3390e2175265d19a...db79fbe667bd53c3cf09e30b5430cb2f2b4fff82). For me, that made the review nice and approachable, as I was able to review small, focused commits instead of facing the whole diff at once. 😅

PR #3660, opened 2021-05-06
ci-test: exit on error

thepwagner:
The most important change of the PR is to each ecosystem's `script/ci-test` scripts, introduced in https://github.com/dependabot/dependabot-core/pull/3426 . We want these shells to exit on the first non-zero status, not to return the status of their _last_ command.

The previous version of these scripts masked problems:
* Most commonly, the output of `rubocop` was ignored. I noticed this as https://github.com/dependabot/dependabot-core/pull/3649 changed rules and passed CI - but produced a diff when auto-applying locally.
* In the `python` ecosystem, the `pylint` status was ignored. I fixed a whitespace issue
* In the `bundler` ecosystem, the status of `rspec` is ignored - we'd removed some required fixtures and were looking for some fixtures in the wrong place.

Sorry for the meandering scope: it's fixing the CI script to detect problems, then trying to find a minimal resolution to problems that have crept in.


thepwagner:
In this draft, I expect the `go_modules` build to fail: I think it had issues the automated fix couldn't resolve, and I wanted to demonstrate lint failures are build failures now.

mctofu (approved):
nice catch!

PR #3661, opened 2021-05-06
go_modules: Gracefully handle +incompatible versions when checking for updates

jasonrudolph:
🚧 [TODO - Fill in meaningful description before marking pull request as "ready for review"] 🚧


go_modules/spec/dependabot/go_modules/update_checker/latest_version_finder_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      Dependabot::DependencyFile.new(
+        name: "go.mod",
+        content: go_mod_content
+      )
+    ]
+  end
+
+  describe "#latest_version" do
+    context "when the latest version is an '+incompatible' version" do # https://golang.org/ref/mod#incompatible-versions
+      let(:dependency_name) { "github.com/go-yaml/yaml" }
```
jasonrudolph, line 45: @mctofu: Do you think this is a situation where we should create our own fixture in the @dependabot-fixtures org?

mctofu: That would guard against any future changes to `github.com/go-yaml/yaml` from breaking the test although that seems less likely in this case. I think we'd just need to setup a go library repo that didn't have a `go.mod` but did have a version tag to reproduce.

jasonrudolph: Thanks, @mctofu. I'll work on getting a fixture for this in @dependabot-fixtures.

go_modules/spec/dependabot/go_modules/update_checker/latest_version_finder_spec.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+require "dependabot/dependency"
+require "dependabot/dependency_file"
+require "dependabot/go_modules/native_helpers"
+require "dependabot/go_modules/update_checker/latest_version_finder"
+
+RSpec.describe Dependabot::GoModules::UpdateChecker::LatestVersionFinder do
+  let(:dependency_name) { "github.com/dependabot-fixtures/go-modules-lib" }
```
mctofu, line 9: For context: we are planning to move the `#latest_resolvable_version` tests from https://github.com/dependabot/dependabot-core/blob/c639755a862939c8d103898daadf8b63a2e85a10/go_modules/spec/dependabot/go_modules/update_checker_spec.rb#L55 to here so it makes sense to have this as the default even though it's not used.

mctofu (approved):
Fix looks good!

jurre (approved):
Fix seems good! I wonder if we should log when this happens?

jasonrudolph:
> I wonder if we should log when this happens?

@jurre: Can you say more about the kind of logging you have in mind?

I'll probably ship this fix as-is to get things for the folks that are affected by https://github.com/dependabot/dependabot-core/issues/3664, but I'm happy to circle back afterwards and add some logging if that seems appropriate.

mctofu:
I think it's worth revisiting later to see if we should change the behavior in this case from "No updated needed" to "Update not possible".

It seems like the cause of this was when you were using an incompatible version but other compatible versions exist. Dependabot is unable to update you to a newer version in that case so "Update not possible" would be a better message to signal that a manual intervention is needed. The user would need to manually update to a compatible version to get updates back on track.

jeffwidman:
> Dependabot is unable to update you to a newer version

For clarification, why is this? Is it because Dependabot can't figure out the newer version, or because it knows a newer version exists but isn't sure how to apply it? 

From the PR description I assume the former (in which case logging an "Unable to determine update" might be appropriate), but from @mctofu 's comment above, it seems he thinks it might be the latter. If the latter, I wonder if #3590 might fix it.... don't know, just thinking aloud.

mctofu:
> For clarification, why is this? Is it because Dependabot can't figure out the newer version, or because it knows a newer version exists but isn't sure how to apply it?

It's because Dependabot can't figure out the newer version. We traced this to the `getVersions` helper not getting any results when querying for versions here: https://github.com/dependabot/dependabot-core/blob/eb4f4779f9fa48553df060c5f580da14ce46b9fb/go_modules/helpers/updatechecker/main.go#L35

That is different behavior than if we were using `go list` to get the versions:
```
% go list -m -versions github.com/go-yaml/yaml
github.com/go-yaml/yaml v2.0.0+incompatible v2.1.0+incompatible
```

That doesn't give any indication that a compatible version exists since it would be a major version upgrade but with those results Dependabot would be able to upgrade to a new incompatible version if available.

PR #3662, opened 2021-05-06
Actions: improve semver support

thepwagner:
This improves the handling of GitHub Actions workflow files when the Action is fetched from a GitHub repository using semver-compatible tags, as [suggested by Actions' best practices](https://docs.github.com/en/actions/creating-actions/about-actions#using-tags-for-release-management).

Our goal is to increase the likelihood that `Depedendency#version` in the `github_actions` package manager is a valid semver tag.  That is what allows the [update types](https://github.com/dependabot/dependabot-core/pull/3550) feature to derive relative versions.

Changes focus on the `FileParser`:
* If the Action's referenced version is a semver string (e.g. `uses: actions/checkout@v2.3.3`, consider that the `Dependency.version == "2.3.3"`
* If the Action's referenced version is a pinned commit (e.g.  `uses: actions/checkout@a81bbbf8298c0fa03ea29cdc473d45769f953675`, AND that commit corresponds to a valid semver tag, consider the tag to be [`Dependency.version == "2.3.3"`](https://github.com/actions/checkout/commits/v2.3.3).
Preferring full length commit SHA is [another GitHub best practice](https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions) that Dependabot should expect and encourage.

<details><summary>Pinned to a sha</summary>
<p>

Pre:
```
=== actions/checkout ()
 => checking for updates 1/1
 => latest available version is 2.3.4
 => latest allowed version is 2.3.4
 => updating actions/checkout to 2.3.4

    ± .github/workflows/pinned.yaml
    ~~~
    7c7
    <     - uses: actions/checkout@aabbfeb2ce60b5bd82389903509092c4648a9713
    ---
    >     - uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
    ~~~
```

Post:
```
=== actions/checkout (2.2.0)
 => checking for updates 1/1
 => latest available version is 2.3.4
 => latest allowed version is 2.3.4
 => updating actions/checkout from 2.2.0 to 2.3.4

    ± .github/workflows/pinned.yaml
    ~~~
    7c7
    <     - uses: actions/checkout@aabbfeb2ce60b5bd82389903509092c4648a9713
    ---
    >     - uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
    ~~~
```

</p>
</details> 

<details><summary>Pinned to version</summary>
<p>

Pre:
```
=== actions/checkout ()
 => checking for updates 1/1
 => latest available version is 5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
 => latest allowed version is 5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
 => updating actions/checkout to 5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f

    ± .github/workflows/pinned.yaml
    ~~~
    7c7
    <     - uses: actions/checkout@v2.2.0
    ---
    >     - uses: actions/checkout@v2.3.4
    ~~~
```

Post:
```
=== actions/checkout (2.2.0)
 => checking for updates 1/1
 => latest available version is 2.3.4
 => latest allowed version is 2.3.4
 => updating actions/checkout from 2.2.0 to 2.3.4

    ± .github/workflows/pinned.yaml
    ~~~
    7c7
    <     - uses: actions/checkout@v2.2.0
    ---
    >     - uses: actions/checkout@v2.3.4
    ~~~
```

</p>
</details>


common/lib/dependabot/git_commit_checker.rb
```diff
@@ -60,10 +60,10 @@ def pinned_ref_looks_like_version?
     end
 
     def pinned_ref_looks_like_commit_sha?
-      return false unless pinned?
-
       ref = dependency_source_details.fetch(:ref)
-      return false unless ref.match?(/^[0-9a-f]{6,40}$/)
+      return false unless ref&.match?(/^[0-9a-f]{6,40}$/)
+
+      return false unless pinned?
```
thepwagner, line 66: This is performance: `pinned?` "does things" - but a regexp check is local and free!

common/lib/dependabot/dependency.rb
```diff
@@ -35,10 +35,12 @@ def self.register_name_normaliser(package_manager, name_builder)
       @name_normalisers[package_manager] = name_builder
     end
 
-    attr_reader :name, :version, :requirements, :package_manager,
+    attr_reader :name, :requirements, :package_manager,
                 :previous_version, :previous_requirements,
                 :subdependency_metadata
 
+    attr_accessor :version
```
thepwagner, line 42: This is pretty out there eh?

It stems from the new `resolve_git_tags` helper - I wanted that to be called after the `DependencySet` had already de-duplicated entries, to minimize the amount of work it would have to do.
The YOLO solution was to make `Dependency#version` mutable, so the "resolve" step can re-assign it.

thepwagner (commented):
Just a checkpoint as it's EOD for me, feedback welcome!

common/lib/dependabot/git_commit_checker.rb
```diff
@@ -115,6 +115,16 @@ def local_tag_for_latest_version
...
     # rubocop:enable Metrics/PerceivedComplexity
 
+    def local_tag_for_pinned_version
+      return unless pinned?
+
+      ref = dependency_source_details.fetch(:ref)
+      tags = local_tags.select { |t| t.commit_sha == ref }
+      return if tags.empty?
+
+      tags.first.name
```
thepwagner, line 125: This probably needs a sort first - e.g. if a single commit is tagged with `v2.3.4`, `v2.3` and `v2`.

That's a common pattern.

common/lib/dependabot/dependency.rb
thepwagner: I rolled this back to instead rely on [DependencySet#<<](https://github.com/dependabot/dependabot-core/blob/eb4f4779f9fa48553df060c5f580da14ce46b9fb/common/lib/dependabot/file_parsers/base/dependency_set.rb#L30-L35) to combine a `Dependency` with the resolved version with the actual `Dependency`.

It's a little gnarly, but it feels better than opening the Pandora's Box of mutability.

thepwagner:
A neat side effect of this change, Actions pinned to a commit SHA get a more clear PR title:

* Pre: `Bump actions/checkout from aabbfeb2ce60b5bd82389903509092c4648a9713 to 2.3.4`
* Post: `Bump actions/checkout from 2.2.0 to 2.3.4`

common/spec/dummy_package_manager/version.rb
```diff
@@ -4,6 +4,21 @@
 
 module DummyPackageManager
   class Version < Gem::Version
+    def initialize(version)
```
thepwagner, line 7: Why is this the same as `Dependabot::GithubActions::Version` now?

In the tests for `local_tag_for_pinned_version`, I'm using a `git-upload-pack` captured from `actions/cache` https://github.com/actions/checkout/tags

Having the active `version_class` (which in that test - is `DummyPackageManager::Version`) understand those tags was the path of least resistance.
Alternatively I could setup and capture a cleaner fixture - I liked the idea of the test case being "real".

github_actions/spec/dependabot/github_actions/update_checker_spec.rb
```diff
@@ -165,11 +165,11 @@
 
     context "given a dependency with a tag reference" do
       let(:reference) { "v1.0.1" }
-      it { is_expected.to eq("5273d0df9c603edc4284ac8402cf650b4f1f6686") }
+      it { is_expected.to eq(Gem::Version.new("1.1.0")) }
```
mctofu, line 168: Is it intentional to use `Gem::Version` here vs `GithubActions::Version`?

mctofu (approved):
I'm not very familiar with this updater but the changes make sense to me!

Since we strip the "v" when populating the dependency version the relative ignore conditions should also start working without any other changes?

github_actions/spec/dependabot/github_actions/update_checker_spec.rb
thepwagner: Nope, it came from the exception:
```
     Failure/Error: it { is_expected.to eq("some hash i replaced") }

       expected: "some hash i replaced"
            got: #<Dependabot::GithubActions::Version "1.1.0">

       (compared using ==)

       Diff:
       @@ -1 +1 @@
       -"some hash i replaced"
       +Gem::Version.new("1.1.0")
```

I was lazy enough to copy/paste the logged string :)
I dig `Dependabot::GithubActions::Version` for clarity, switching.

aaronfranke:
@thepwagner @mctofu Could this PR have caused #3704?

PR #3672, opened 2021-05-07
Spike: tf12+ support

jurre:


jurre:
Most of this work has since landed, there are a few bits for provider updates that I'll pull in separately

PR #3685, opened 2021-05-10
Terraform: Document and improve coverage for RequirementsUpdater

jurre:
I found myself having to dig a bit to figure out what this class does,
and felt some docs would have been helpful.


jurre:
While working on this I found myself wish we had a `DependencyRequirement` class or something that takes the place of this hash format that we're passing around. It's a bit of a bigger conversation though, since we use this pattern in all the ecosystems (but actually do not rely on it outside of the package boundary, meaning there is no common functionality in `common` built on-top of this abstraction)

xlgmokha (approved):

PR #3688, opened 2021-05-10
Run all pip-compile commands with options

JimNero009:
I believe I have stumbled across a bug in check_original_requirements_resolvable in the pip-compile version resolver.

I think the error is that the pip-compile command runs here without the relevant options that it runs with in all other contexts. This can mean, for example, that this command can fail when requirements contain libraries that are present only in a private repository and not on the public PyPi.

I can see three ways to resolve this:
1. Add the call to the relevant function when running pip-compile. This is the simplest and what I've opened initially, for discussion.
2. Make a function that returns the full pip-compile-with-options command that should be used everywhere, so you don't have to remember to add the option
3. Write a local pip.conf file and ensure it's picked up by the relevant commands to avoid having to pass them in as arguments each time.

WDYT, in terms of whether this is indeed a bug and what the fix should be?

### DIFF
--- a/python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
+++ b/python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
@@ -169,7 +169,7 @@ module Dependabot
 
               filenames_to_compile.each do |filename|
                 run_pip_compile_command(
-                  "pyenv exec pip-compile --allow-unsafe #{filename}"
+                  "pyenv exec pip-compile #{pip_compile_options(filename)} --allow-unsafe #{filename}"
                 )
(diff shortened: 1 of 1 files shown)


jurre:
Yeah, it looks like a bug at first glance. I'm going to wait for the tests to confirm that for us 😬 

It they indeed pass with this change (I suspect they will), I think having a function that returns the pip command with options sounds good, we have some logic in place that swaps them out and I think it'll be hard to do with a `pip.conf` file

Thanks for looking into this @JimNero009

jurre (approved):
Going to merge this as is for now, we can extract a method separately

PR #3700, opened 2021-05-11
Add HCL2 tests to file_updater

Nishnha:
- Modified git_tags_012/main.tf to keep ref=tags/version formatting
- Add test for no dependency updates


jurre (approved):
Nice work!

PR #3721, opened 2021-05-14
Terraform: remove legacy terraform feature flag

jurre:
The legacy terraform parser is currently not used in production, this
removes the feature flag to use it in local dev, that we left behind to
make it easier to revert.

Once we're confident we won't need to do that, we can remove the old
implementation by merging this in.

This PR also removes the codepath for legacy terragrunt files.

Our new parser does not support legacy terragrunt files, and the legacy
terragrunt file format is very old and not widely used at all.

Support for it was already removed, and only worked behind the legacy
feature flag which is not enabled


terraform/lib/dependabot/terraform/file_fetcher.rb
```diff
@@ -10,7 +10,7 @@ class FileFetcher < Dependabot::FileFetchers::Base
       include FileSelector
 
       def self.required_files_in?(filenames)
-        filenames.any? { |f| f.end_with?(".tf", ".tfvars") }
+        filenames.any? { |f| f.end_with?(".tf", ".hcl") }
```
jurre, line 13: If we plan to hold off merging this, we'll want to backport this change cc @xlgmokha @Nishnha, as otherwise we'd raise an error if there is only a `.hcl` file in the repo

xlgmokha (approved):
🔥

jurre:
The hex failure is unrelated and also happening on other branches, so I'm going to merge this in and will look at that separately.

PR #3728, opened 2021-05-17
Hex: Fix version resolver specs

feelepxyz:
Fixes broken hex specs: https://github.com/dependabot/dependabot-core/pull/3719/checks?check_run_id=2581264410

Handle hex errors that return valid json, e.g. `{"error":"Dependency resolution failed: Can't continue due to errors on dependencies"}`

Previously we only checked if the sub-process output was parseable and tried to get the `result` key, but this will just return nil if the json is valid.

For context, valid results look like `{"result":"1.3.6"}`

### DIFF
--- a/hex/lib/dependabot/hex/update_checker/version_resolver.rb
+++ b/hex/lib/dependabot/hex/update_checker/version_resolver.rb
@@ -101,8 +101,7 @@ module Dependabot
           result = error.message&.split("\n")&.last
           return false unless result
 
-          JSON.parse(error.message&.split("\n")&.last)["result"]
-          true
+          JSON.parse(error.message&.split("\n")&.last).key?("result")
         rescue JSON::ParserError
           false
         end
--- a/hex/lib/dependabot/hex/update_checker/version_resolver.rb
+++ b/hex/lib/dependabot/hex/update_checker/version_resolver.rb
@@ -74,8 +74,10 @@ module Dependabot
             raise Dependabot::PrivateSourceAuthenticationFailure, org if org
           end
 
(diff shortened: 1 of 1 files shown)


jurre (approved):
Nice find 🎉

jurre (changes requested):
I cheered to quickly 😄 looks like the tests are still failing: https://github.com/dependabot/dependabot-core/pull/3728/checks?check_run_id=2601941563#step:9:295

feelepxyz:
Ah yes looks like something else failed now, will take a look

hex/lib/dependabot/hex/update_checker/version_resolver.rb
```diff
@@ -122,6 +123,12 @@ def check_original_requirements_resolvable
 
           true
         rescue SharedHelpers::HelperSubprocessFailed => e
+          # TODO: Catch the warnings as part of the Elixir module. This happens
+          # when elixir throws warnings from the manifest files that end up in
+          # stdout and cause run_helper_subprocess to fail parsing the result as
+          # JSON.
+          return error_result(e) if includes_result?(e)
```
feelepxyz, line 130: Will have a poke at the underlying issue, as this is caused by warnings like: 

```
warning: a term is constructed, but never used
  lib/encoder.ex
```

Don't think this is from our code so might be internal hex or customer manifests

feelepxyz: This change is consistent with the error result handling in `handle_hex_errors`

jurre (approved):

PR #3729, opened 2021-05-17
Terraform: Support provider updates

### LINKED ISSUE #2054: Support Terraform provider version bump
Dependabot still does not support version bump in the provider (specially `aws` provider). Issue is that `aws` release verify often and it is quite hard to keep up with its pace there. It would be amazing if provider version bump for Terraform module could be provided.

jurre:
This PR implements version updates for Terraform [providers](https://www.terraform.io/docs/language/providers/requirements.html)

Fixes #2054


terraform/lib/dependabot/terraform/file_parser.rb
```diff
@@ -106,17 +119,23 @@ def source_from(details_hash)
...
         source_details[:proxy_url] = raw_source if raw_source != bare_source
         source_details
       end
 
-      def registry_source_details_from(source_string)
+      def registry_source_details_from(source_string, provider)
         parts = source_string.split("//").first.split("/")
 
-        if parts.count == 3
+        if provider && parts.count == 2
```
jurre, line 132: I'm not happy having to pass down this 'provider' boolean this deeply, but I didn't find a clean way yet without having a bunch of duplication 🤔 

Would love suggestions on how to make this suck less

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -47,7 +47,7 @@ def updated_terraform_file_content(file)
           case new_req[:source][:type]
           when "git"
             update_git_declaration(new_req, old_req, content, file.name)
-          when "registry"
+          when "registry", "provider"
```
jurre, line 50: The naming of these still bugs me, they should probably be `registry_module` and `registry_provider`

terraform/lib/dependabot/terraform/update_checker.rb
```diff
@@ -65,34 +67,35 @@ def latest_version_for_registry_dependency
...
+        RegistryClient.new(hostname: hostname).all_module_versions(identifier: identifier)
+      end
 
-        JSON.parse(response.body).
-          fetch("modules").first.fetch("versions").
-          map { |release| version_class.new(release.fetch("version")) }
+      def all_provider_versions
+        hostname = dependency_source_details.fetch(:registry_hostname)
+        identifier = dependency_source_details.fetch(:module_identifier)
+        RegistryClient.new(hostname: hostname).all_provider_versions(identifier: identifier)
```
jurre, line 98: We could "reuse" this client, but I don't think we actually call it more than twice, so it would just DRY up the code a tiny bit

jurre:
I'm planning to pull out the `RegistryClient` into a separate PR to keep this changeset a bit smaller

terraform/lib/dependabot/terraform/file_parser.rb
xlgmokha: 🤔 I took a peek at https://www.terraform.io/docs/language/providers/requirements.html#source-addresses which says:

> `[<HOSTNAME>/]<NAMESPACE>/<TYPE>`

Here's an idea:

```ruby
host, namespace, type = source_address_from(source_string)

def source_address_from(source_string)
  parts = source_string.split('/')
  parts.length == 2 ? ['registry.terraform.io', parts[0], parts[1]] : parts
end
```

Feel free to ignore.

terraform/lib/dependabot/terraform/file_updater.rb
xlgmokha: It's a bit unclear that `registry` refers to modules. I'm not sure what the impact vs. effort is to address this. Perhaps, if it's important enough we can consider it in a follow up PR. WDYT?

terraform/lib/dependabot/terraform/registry_client.rb
```diff
@@ -46,7 +64,13 @@ def source(dependency:)
...
         return unless hostname == PUBLIC_HOSTNAME
 
-        endpoint = "modules/#{dependency.name}/#{dependency.version}"
+        endpoint = if dependency.requirements.first[:source][:type] == "registry"
+                     "modules/#{dependency.name}/#{dependency.version}"
+                   elsif dependency.source[:type] == "provider"
+                     "providers/#{dependency.name}/#{dependency.version}"
+                   else
+                     raise "Invalid source type"
+                   end
```
xlgmokha, line 73: I kind wish we could write something like:

```ruby
[pluralize(dependency.source[:type]), dependency.name, dependency.version].join('/')
```

to collapse the conditional logic. I think this suggestion would be a bigger change so I'm not asking for it.

xlgmokha (approved):
LGTM from what I can see. Eventually I would love to see some fixtures with source addresses that include an optional `[HOSTNAME/]` section described in https://www.terraform.io/docs/language/providers/requirements.html#source-addresses. I don't think we need it yet.

terraform/lib/dependabot/terraform/file_parser.rb
jurre: Yeah I had considered a similar approach, but the downside is that if someone has an invalid module configured, we would end up trying to treat it as a provider. We *know* if we're dealing with a module or a provider, so trying to assume it seems slightly off? 🤔

terraform/lib/dependabot/terraform/file_updater.rb
jurre: Agreed!

terraform/lib/dependabot/terraform/file_parser.rb
xlgmokha: > Yeah I had considered a similar approach, but the downside is that if someone has an invalid module configured, we would end up trying to treat it as a provider. We _know_ if we're dealing with a module or a provider, so trying to assume it seems slightly off? 🤔

I don't quite see where modules fit into this. Is it possible that this comment was meant for [this suggestion](https://github.com/dependabot/dependabot-core/pull/3729#discussion_r634666201) or is there more that I'm not seeing yet.

PR #3732, opened 2021-05-18
Add helpful comment

jeffwidman:
I ran across this comment on an old PR while researching https://github.com/dependabot/dependabot-core/pull/3590/.

The comment seems very useful/easy to overlook, so thought I should add it here.


feelepxyz (approved):
Thanks! 👍

PR #3733, opened 2021-05-18
Replace custom helper with `go get -d lib@version`

jeffwidman:
This is a more bite-sized take on #3590. That PR had a functionality change which needs further discussion.

So this PR only includes the implementation swap from manually modifying
`go.mod` to using `go get lib@version` for bumping deps.

Previously for the dep we wanted to bump, we'd manually munge `go.mod`,
then run `go get` to update `go.sum` and run the MVS algorithm
to determine if any other changes needed to be made to `go.mod`.

However, this caused a few problems:
* the `go` docs/comments on github issues consistently recommend _not_
editing `go.mod` directly, but instead use `go get` to update it.
* We have to copy/paste in private methods from upstream, and it's a
pain to keep those in sync: #3580

Instead of doing all this munging, we can simply pass our desired
version to `go get -d <dep>@<version>` and it will handle updating
`go.mod` and `go.sum`.

This is both more idiomatic and reduces the risk of our logic for
updating `go.mod` from diverging from upstream.

The slight test changes are because the error message changed slightly.


go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -152,22 +151,27 @@ def run_go_vendor
...
+          dependencies.each do |dep|
+            # Use version pinning rather than `latest` just in case
+            # a new version gets released in the middle of our run.
+            version = "v" + dep.version.sub(/^v/i, "")
+            command << " #{dep.name}@#{version}"
+          end
+          _, stderr, status = Open3.capture3(ENVIRONMENT, command)
+          handle_subprocess_error(stderr) unless status.success?
+        ensure
+          File.delete(tmp_go_file) if File.exist?(tmp_go_file)
```
jeffwidman, line 174: If we go this route, there's some DRY'ing up that could happen (maybe for the final call use `run_go_get()` with no args? I need to check how Ruby would handle this to ensure the generated command would simply be `go get -d`... this is purely my notes to self as I need to head to dinner+helping my wife with the kids.

jeffwidman:
The `gradle` test failure is unrelated, I filed https://github.com/dependabot/dependabot-core/issues/3735 to address it.

@jurre please reboot that particular test and it should pass.

jurre (approved):
Thanks for splitting this out into separate PRs 🎉

jeffwidman:
Looks like CI might need another jumpstart.

jurre:
> Looks like CI might need another jumpstart.

Yeah there is currently an [incident with GitHub Actions](https://www.githubstatus.com/incidents/m16jzl31gnqt) so we'll have to wait that out :(

PR #3740, opened 2021-05-18
npm: Handle workspace deps installed top-level

feelepxyz:
Handle nested workspace dependencies that have been installed in the
top-level `node_modules` folder instead of the nested
`workspace/node_modules` folder.

While this case returned `nil` previously, we still ended up extracting
lockfile details for the dependency by checking all instances of the
same dependency and picking the lowest version, meaning the `from`
version in updates was incorrect.

It seems this happens when the package is only installed once, so
doesn't end up having several versions across different pacakges.


jurre (approved):
👍

PR #3756, opened 2021-05-19
Parse version from terraform lock file

xlgmokha:
# Why is this needed?

To parse the version from the [terraform lock file][lock_file] so that
the correct version is compared against the eligible versions.

fixes:

```bash
updater | INFO <job_139154747> Starting job processing
updater | INFO <job_139154747> Starting update job for dsp-testing/dependabot-updates-1480
updater | INFO <job_139154747> Checking if hashicorp/aws ~> 3.0 needs updating
  proxy | 2021/05/25 19:31:39 [012] GET https://registry.terraform.io:443/v1/providers/hashicorp/aws/versions
  proxy | 2021/05/25 19:31:39 [012] 200 https://registry.terraform.io:443/v1/providers/hashicorp/aws/versions
updater | INFO <job_139154747> Latest version is 3.42.0
updater | I, [2021-05-25T19:31:39.299128 #7]  INFO -- sentry: ** [Raven] Sending event 2930bd6eaf7e4716bbf3ed064fb224f7 to Sentry
  proxy | 2021/05/25 19:31:39 [014] POST https://sentry.io:443/api/1451818/store/
  proxy | 2021/05/25 19:31:39 [014] 200 https://sentry.io:443/api/1451818/store/
updater | ERROR <job_139154747> Error processing hashicorp/aws (ArgumentError)
updater | ERROR <job_139154747> Malformed version number string ~> 3.0
updater | ERROR <job_139154747> /usr/lib/ruby/2.6.0/rubygems/version.rb:212:in `initialize'
updater | ERROR <job_139154747> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-terraform-0.148.8/lib/dependabot/terraform/version.rb:14:in `initialize'
updater | ERROR <job_139154747> /usr/lib/ruby/2.6.0/rubygems/version.rb:201:in `new'
updater | ERROR <job_139154747> /usr/lib/ruby/2.6.0/rubygems/version.rb:201:in `new'
updater | ERROR <job_139154747> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-common-0.148.8/lib/dependabot/update_checkers/base.rb:238:in `numeric_version_up_to_date?'
updater | ERROR <job_139154747> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-common-0.148.8/lib/dependabot/update_checkers/base.rb:189:in `version_up_to_date?'
updater | ERROR <job_139154747> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-common-0.148.8/lib/dependabot/update_checkers/base.rb:33:in `up_to_date?'
updater | ERROR <job_139154747> /home/dependabot/dependabot-updater/lib/dependabot/updater.rb:205:in `check_and_create_pull_request'
updater | ERROR <job_139154747> /home/dependabot/dependabot-updater/lib/dependabot/updater.rb:80:in `check_and_create_pr_with_error_handling'
updater | ERROR <job_139154747> /home/dependabot/dependabot-updater/lib/dependabot/updater.rb:56:in `block in run'
updater | ERROR <job_139154747> /home/dependabot/dependabot-updater/lib/dependabot/updater.rb:56:in `each'
updater | ERROR <job_139154747> /home/dependabot/dependabot-updater/lib/dependabot/updater.rb:56:in `run'
updater | ERROR <job_139154747> /home/dependabot/dependabot-updater/lib/dependabot/update_files_job.rb:17:in `perform_job'
updater | ERROR <job_139154747> /home/dependabot/dependabot-updater/lib/dependabot/base_job.rb:28:in `run'
updater | ERROR <job_139154747> bin/update_files.rb:21:in `<main>'
updater | INFO <job_139154747> Checking if namespace/name ~> 0.1 needs updating
updater | INFO <job_139154747> Latest version is 
updater | INFO <job_139154747> Requirements to unlock update_not_possible
updater | INFO <job_139154747> Requirements update strategy 
updater | INFO <job_139154747> No update possible for namespace/name ~> 0.1
updater | INFO <job_139154747> Finished job processing
```

## What does this do?

It parses the version from the [lock file][lock_file].

/xref: https://github.com/github/dependabot-updates/issues/1480

## Type of change

## Verification Plan

- [ ] `.deploy ` this PR
- [ ] Re-run update job for [test repo](https://github.com/dsp-testing/dependabot-updates-1480/network/updates/)

[source_address]: https://www.terraform.io/docs/language/providers/requirements.html#source-addresses
[lock_file]: https://www.terraform.io/docs/language/dependency-lock.html

### DIFF
--- a/terraform/lib/dependabot/terraform/file_parser.rb
+++ b/terraform/lib/dependabot/terraform/file_parser.rb
@@ -20,6 +20,9 @@ module Dependabot
       include FileSelector
 
       ARCHIVE_EXTENSIONS = %w(.zip .tbz2 .tgz .txz).freeze
+      DEFAULT_REGISTRY = "registry.terraform.io".freeze
+      # https://www.terraform.io/docs/language/providers/requirements.html#source-addresses
+      PROVIDER_SOURCE_ADDRESS = /\A((?<hostname>.+)\/)?(?<namespace>.+)\/(?<name>.+)\z/
 
       def parse
         dependency_set = DependencySet.new
@@ -34,7 +37,11 @@ module Dependabot
             required_providers = terraform.fetch("required_providers", {})
             required_providers.each do |provider|
               provider.each do |name, details|
-                dependency_set << build_terraform_dependency(file, name, details, true)
+                dependency_set << build_provider_dependency_from(
+                  file.name,
+                  details.fetch("source"),
+                  details["version"]&.strip
+                )
               end
             end
           end
@@ -53,6 +60,26 @@ module Dependabot
       end
 
       private
+
+      def build_provider_dependency_from(file_name, source_address, version)
+        hostname, namespace, name = provider_source_from(source_address)
+
+        Dependency.new(
+          name: "#{namespace}/#{name}",
+          version: version,
+          package_manager: "terraform",
+          requirements: [
+            requirement: version,
+            groups: [],
+            file: file_name,
+            source: {
+              type: "provider",
+              registry_hostname: hostname,
+              module_identifier: "#{namespace}/#{name}"
+            }
+          ]
+        )
+      end
 
       def build_terraform_dependency(file, name, details, provider)
         details = details.is_a?(Array) ? details.first : details
@@ -150,6 +177,15 @@ module Dependabot
           msg = "Invalid registry source specified: '#{source_string}'"
           raise DependencyFileNotEvaluatable, msg
         end
+      end
+
+      def provider_source_from(source_address)
+        matches = source_address.match(PROVIDER_SOURCE_ADDRESS)
+        [
+          matches[:hostname] || DEFAULT_REGISTRY,
+          matches[:namespace],
+          matches[:name]
+        ]
       end
 
       def git_source_details_from(source_string)
--- a/terraform/lib/dependabot/terraform/file_parser.rb
+++ b/terraform/lib/dependabot/terraform/file_parser.rb
@@ -184,15 +184,6 @@ module Dependabot
         end
       end
 
-      def provider_source_from(source_address)
-        matches = source_address.match(PROVIDER_SOURCE_ADDRESS)
-        [
-          matches[:hostname] || DEFAULT_REGISTRY,
-          matches[:namespace],
-          matches[:name]
-        ]
-      end
-
       def git_source_details_from(source_string)
         git_url = source_string.strip.gsub(/^git::/, "")
         git_url = "https://" + git_url unless git_url.start_with?("git@") || git_url.include?("://")
--- a/terraform/lib/dependabot/terraform/file_parser.rb
+++ b/terraform/lib/dependabot/terraform/file_parser.rb
@@ -20,9 +20,9 @@ module Dependabot
       include FileSelector
 
       ARCHIVE_EXTENSIONS = %w(.zip .tbz2 .tgz .txz).freeze
-      DEFAULT_REGISTRY = "registry.terraform.io".freeze
+      DEFAULT_REGISTRY = "registry.terraform.io"
       # https://www.terraform.io/docs/language/providers/requirements.html#source-addresses
-      PROVIDER_SOURCE_ADDRESS = /\A((?<hostname>.+)\/)?(?<namespace>.+)\/(?<name>.+)\z/
+      PROVIDER_SOURCE_ADDRESS = %r{\A((?<hostname>.+)/)?(?<namespace>.+)/(?<name>.+)\z}.freeze
 
       def parse
         dependency_set = DependencySet.new
@@ -36,12 +36,8 @@ module Dependabot
           parsed_file(file).fetch("terraform", []).each do |terraform|
             required_providers = terraform.fetch("required_providers", {})
             required_providers.each do |provider|
-              provider.each do |name, details|
-                dependency_set << build_provider_dependency_from(
-                  file.name,
-                  details.fetch("source"),
-                  details["version"]&.strip
-                )
+              provider.each do |_name, details|
+                dependency_set << build_provider_dependency_from(file, details)
               end
             end
           end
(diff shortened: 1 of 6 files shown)


jurre (approved):
Nice one!

I don't think we'd currently actually make those requests, we bail out on anything but a public module or provider: https://github.com/dependabot/dependabot-core/blob/ec05a0b7a02bb2fdd0bd4da091757a156033d240/terraform/lib/dependabot/terraform/registry_client.rb#L27 so I wouldn't expect to see a GET to that resource yet.

Also, might suffice to run a dry-run to verify this instead of deploying the PR (assuming you meant pre-merge, which is a hassle).

And another question, I think we would also expect credentials to have been configured for the repo and for those to be used when referencing a private source, right? Is that something we should add here, or in a follow-up PR?

terraform/lib/dependabot/terraform/file_parser.rb
```diff
@@ -54,6 +57,28 @@ def parse
...
+            source: {
+              type: "provider",
+              registry_hostname: hostname,
+              module_identifier: "#{namespace}/#{name}"
+            }
+          ]
+        )
+      end
+
       def build_terraform_dependency(file, name, details, provider)
```
feelepxyz, line 82: Can the `provider` arg be removed here? Looks like it might always be `false` now

xlgmokha:
> I don't think we'd currently actually make those requests, we bail out on anything but a public module or provider:
> 
> https://github.com/dependabot/dependabot-core/blob/ec05a0b7a02bb2fdd0bd4da091757a156033d240/terraform/lib/dependabot/terraform/registry_client.rb#L27
> 
> so I wouldn't expect to see a GET to that resource yet.

Good to know. I'll take a look at that.

> Also, might suffice to run a dry-run to verify this instead of deploying the PR (assuming you meant pre-merge, which is a hassle).

Good call. Trying a dry-run produced the following:

```bash
[dependabot-core-dev] ~/dependabot-core $ ./bin/dry-run.rb terraform dsp-testing/dependabot-updates-1480
=> fetching dependency files
=> dumping fetched dependency files: ./dry-run/dsp-testing/dependabot-updates-1480/
=> parsing dependency files
=> updating 2 dependencies: hashicorp/aws, namespace/name

=== hashicorp/aws (~> 3.0)
 => checking for updates 1/2
 => latest available version is 3.41.0
 => latest allowed version is 3.41.0
        9: from ./bin/dry-run.rb:599:in `<main>'
        8: from ./bin/dry-run.rb:599:in `each'
        7: from ./bin/dry-run.rb:646:in `block in <main>'
        6: from /home/dependabot/dependabot-core/common/lib/dependabot/update_checkers/base.rb:33:in `up_to_date?'
        5: from /home/dependabot/dependabot-core/common/lib/dependabot/update_checkers/base.rb:189:in `version_up_to_date?'
        4: from /home/dependabot/dependabot-core/common/lib/dependabot/update_checkers/base.rb:238:in `numeric_version_up_to_date?'
        3: from /usr/lib/ruby/2.6.0/rubygems/version.rb:201:in `new'
        2: from /usr/lib/ruby/2.6.0/rubygems/version.rb:201:in `new'
        1: from /home/dependabot/dependabot-core/terraform/lib/dependabot/terraform/version.rb:14:in `initialize'
/usr/lib/ruby/2.6.0/rubygems/version.rb:212:in `initialize': Malformed version number string ~> 3.0 (ArgumentError)
```

So it looks like there's a little more to do to handle versions.

> And another question, I think we would also expect credentials to have been configured for the repo and for those to be used when referencing a private source, right? Is that something we should add here, or in a follow-up PR?

I think I need to follow-up on this. From what I can tell the credentials are managed by the dependency proxy side but I'll make sure to take a closer look.

Thanks @jurre.

terraform/lib/dependabot/terraform/file_parser.rb
xlgmokha: I think it can. Would it be okay for me to do that in this PR or would you prefer that to be a separate one?

jurre:
> From what I can tell the credentials are managed by the dependency proxy side but I'll make sure to take a closer look.
 
So, the _actual_ credential does get passed in from the proxy (for the way we run dependabot at GitHub), but our code still needs to set the headers etc, and when dry-running (or when folks use dependabot-core standalone) there is no proxy, an example in npm is here:
 
https://github.com/dependabot/dependabot-core/blob/jurre/upgrade-bundler/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/registry_finder.rb#L74-L86
 
But, we do need to implement a handler on the proxy side as well 👍

xlgmokha:
> > From what I can tell the credentials are managed by the dependency proxy side but I'll make sure to take a closer look.
> 
> So, the _actual_ credential does get passed in from the proxy (for the way we run dependabot at GitHub), but our code still needs to set the headers etc, and when dry-running (or when folks use dependabot-core standalone) there is no proxy, an example in npm is here:
> 
> https://github.com/dependabot/dependabot-core/blob/jurre/upgrade-bundler/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/registry_finder.rb#L74-L86
> 
> But, we do need to implement a handler on the proxy side as well 👍

Sweet. I added the link above next to ` x: verify if any changes need to be made to the dependency proxy to support this new registry type.` in https://github.com/github/dependabot-updates/issues/1480 so that I don't lose that helpful note. Thanks @jurre.

jurre:
Hey, I had to do a double take on this since the entire PR changed after approving it, but I don't think this will fully resolve the issue that is now described in the PR body, those errors happen because we pull in a requirement that is not a valid version, and in many cases we simply can't determine a version, so that field needs to be `nil`.

This would indeed fix projects that have a lockfile present, but the majority of those projects do not have a lockfile present, so parsing this won't resolve the error for those cases.

To fix this for projects without a lockfile I opened: https://github.com/dependabot/dependabot-core/pull/3792

jurre:
I don't think this will currently work, because we're not fetching the lockfiles. Luckily it'll still work for the majority of cases where the is no lockfile, as it'll correctly set the version to nil

PR #3763, opened 2021-05-21
Terraform: Improve updating provider requirements

jurre:
Previously we would match for any `version` string, and this would end
up matching on `required_version`, and we'd fail to update these
configs.


brrygrdn (approved):
👍🏻

PR #3766, opened 2021-05-21
Terraform: Add lockfile update support

Nishnha:
## Overview

Lockfiles were added to Terraform v0.14 and only track Terraform `providers`.  The lockfile entry for a provider will contain the version, the version constraints, and the dependency hash.

This pull request adds in Dependabot updater support for lockfiles.

## What it does

When a provider dependency is updated by Dependabot, we update the mainfest vile (`main.tf` or `versions.tf`) dependency version and update the lockfile entry.

The implementation for terraform lockfile updates is based on bundler lockfile updates.

## Known issues

** These will be addressed in future PRs **

- Lockfiles will not update if the provider dependency version is pinned. 
  - This is because lockfile entries are updated independently of the manifest file. When the lockfile update it is not aware of the version change in the manifest file.
- Rspec tests for the lockfile currently match against the *latest at the time* dependency version. If there is a new version for a dependency after the test was updated, the test will fail.
   - Fix might require stubbing the network call for `terraform providers lock`
   
 ## Testing the Pull Request
 
 Currrently testing against Terraform rspec tests and 3 GitHub internal repositories. From inside the docker-dev-shell with `LOCAL_GITHUB_ACCESS_TOKEN` set:
 ```bash 
 cd terraform && rspec spec
 ```
 
 ```bash
 bin/dry-run.rb terraform dsp-testing/dependabot-terraform-v0.15-test --cache=files --dir="/." 
 ```
 
 ```bash
 bin/dry-run.rb terraform dsp-testing/dependabot-terraform-lockfile-test --cache=files --dir="/."
 ```
 
 ```bash
 bin/dry-run.rb terraform dsp-testing/dependabot-terraform-lockfile-test --cache=files --dir="/." --branch=pinned-versions
 ```
   
## Update log

- [x] Add lockfiles as a dependency source
- [x] Update lockfile dependencies individually via `providers lock update`
- [x] Refactor: remove lockfile dependencies, update lockfile when provider dependency matches lockfile entry.
- [x] Fix: lockfile detected changes when dependency hash opportunistically updated. Only update when lockfile version changes.
- [x] Delete "dead" commits from PR
- [x] Rebase onto main
- [x] Memoize network calls from `terraform provider lock`
- [ ] Update lockfile when updating from a pinned version
- [ ] Pin rspec tests dependency versions! (stub network call?)
- [ ] Merge PR, cut release


terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -80,6 +86,21 @@ def update_registry_declaration(new_req, old_req, updated_content)
...
+      def update_lockfile_declaration(new_req, old_req, updated_content)
+        return unless lock_file?(filename)
+
+        provider_source = new_req[:source][:registry_hostname] + "/" + new_req[:source][:module_identifier]
+        declaration_regex = lockfile_declaration_regex(provider_source)
+
+        copy_dir_to_temporary_directory do
+          updated_content.sub!(declaration_regex, "")
+          
+          system "terraform providers lock #{provider_source}"
```
feelepxyz, line 98: Small drive by comment: we've got this util to run shell commands with escaped arguments

```suggestion
          `SharedHelpers.run_shell_command("terraform providers lock #{provider_source}")
```

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -39,16 +41,20 @@ def updated_terraform_file_content(file)
...
           case new_req[:source][:type]
           when "git"
             update_git_declaration(new_req, old_req, content, file.name)
-          when "registry", "provider"
+          when "registry" 
             update_registry_declaration(new_req, old_req, content)
+          when "provider"
+            update_registry_declaration(new_req, old_req, content)
+          when "lockfile"
+            update_lockfile_declaration(new_req, old_req, content)
```
feelepxyz, line 57: I wonder if we should refactor these methods to return the updated content instead of mutating the `content` argument, this might make it easier to reason about `update_lockfile_declaration`, e.g.

```
when "provider"
  content = update_registry_declaration(new_req, old_req, content)
when "lockfile"
  content = update_lockfile_declaration(new_req, old_req, content)
...
```

terraform/lib/dependabot/terraform/file_parser.rb
```diff
@@ -53,6 +53,15 @@ def parse
...
 
+        lock_file.each do |file|
+          lockfile_provider = parsed_file(file).fetch("provider", {})
+          lockfile_provider.each do |provider|
+            provider.each do |details, hashes|
+              dependency_set << build_lockfile_dependency(file, details, hashes)
+            end
+          end
+        end
+
```
jurre, line 64: I think we can get rid of this as these files are already parsed and pull out the resolved version (the only thing we care about in this step) [here](https://github.com/dependabot/dependabot-core/blob/2be3d7c81db4b5944f770c0dbb5af7b4b16b9e80/terraform/lib/dependabot/terraform/file_parser.rb#L336-L341). That means we can also get rid of `build_lockfile_dependency`. It'd be sweet to keep `def lock_file` in the file fetcher, so we'd get rid of that method in the `file_parser`.

terraform/lib/dependabot/terraform/file_selector.rb
```diff
@@ -14,4 +14,12 @@ def terragrunt_files
   def terragrunt_file?(file_name)
     file_name != ".terraform.lock.hcl" && file_name.end_with?(".hcl")
```
jurre, line 15: Maybe we can reuse the new `lock_file?` method here now:

```suggestion
    !lock_file?(file_name) && file_name.end_with?(".hcl")
```

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -126,6 +153,26 @@ def git_declaration_regex(filename)
...
+          provider\s*["']#{Regexp.escape(provider_source)}["']\s*\{
+          (?:(?!^\}).)*}
+        /mx
+      end
+
+      def copy_dir_to_temporary_directory
+        SharedHelpers.in_a_temporary_directory do
+          dependency_files.each do |file|
+            # Do not include the .terraform directory
+            next if file.name.include?(".terraform/")
```
jurre, line 169: I would not expect any of the `dependency_files` to be this directory, when does this happen?

terraform/lib/dependabot/terraform/file_parser.rb
Nishnha: Initially the lockfile didn't build dependencies, but then `FileUpdater` would skip over the lockfile since no dependencies changed: 

```ruby
[*terraform_files, *terragrunt_files, *lock_file].each do |file|
      next unless file_changed?(file)
```
Where `file_changed?` calls the `FileUpdaters/base#requirements_changed?` function on each dependency of the file

```ruby
def requirement_changed?(file, dependency)
    changed_requirements = dependency.requirements - dependency.previous_requirements
    changed_requirements.any? { |f| f[:file] == file.name }
end
```

If we get rid of lockfile dependencies we would need to either remove the `file_changed?(file)` check and fix the test suite or create exceptions for when the file being updated is a lockfile.

terraform/lib/dependabot/terraform/file_updater.rb
Nishnha: Dependency files are copied to a temporary directory to provide context for the `terraform providers lock` command. The files are copied over in the `FileUpdater#update_lockfile_declaration` function but I'm not sure when the `dependency_files` array itself is populated.

jurre: It's what we grab in the `FileFetcher`, I don't think we ever pull in those files so we probably don't need to worry about it?

terraform/lib/dependabot/terraform/file_parser.rb
```diff
@@ -163,10 +163,30 @@ def provider_source_from(source_address, name)
...
+        lockfile_requirement = dependency.requirements.select{ |requirement| requirement[:source][:type] == "lockfile" }.first
+        provider_requirement = dependency.requirements.select{ |requirement| requirement[:source][:type] == "provider" }.first
+
+        if lockfile_requirement[:source][:registry_hostname] == provider_requirement[:source][:registry_hostname] &&
+           lockfile_requirement[:source][:module_identifier] == lockfile_requirement[:source][:module_identifier]
+          return dependency.requirements.select!{ |requirement| requirement == lockfile_requirement }
+        end
+
+        dependency
+      end
```
jurre, line 178: I think this is no longer used so we can 🔥 it?

terraform/lib/dependabot/terraform/file_parser.rb
```diff
@@ -163,10 +163,30 @@ def provider_source_from(source_address, name)
...
+
       def registry_source_details_from(source_string)
         parts = source_string.split("//").first.split("/")
 
-        if parts.count == 3
+        if parts.count == 2
+          {
+            type: "provider",
```
jurre, line 185: I think this is now handled in a different method https://github.com/dependabot/dependabot-core/blob/0a9377f61e31f3ecac9bf88612504a8fa1f1d260/terraform/lib/dependabot/terraform/file_parser.rb#L89-L110 so I don't think we need to account for it here

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -16,38 +17,47 @@ def self.updated_files_regex
 
       def updated_dependency_files
         updated_files = []
-
+
```
jurre, line 20:
```suggestion

```

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -16,38 +17,47 @@ def self.updated_files_regex
 
       def updated_dependency_files
         updated_files = []
-
+        
         [*terraform_files, *terragrunt_files].each do |file|
           next unless file_changed?(file)
-
+
```
jurre, line 23:
```suggestion

```

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -81,6 +91,42 @@ def update_registry_declaration(new_req, old_req, updated_content)
         end
       end
 
+      def update_lockfile_declaration(dependency)
```
jurre, line 94: We're calling this twice, redoing all that work each time, we can solve this in two ways, the first one is to memoize the result of this method, in that case we need to account for the fact that we pass it a `dependency` arg, so:

```ruby
def update_lockfile_declaration(dependency)
  @updated_lockfile_declarations[dependency] ||=
    begin
      # .. the current implementation
    end
end
```

Or we can make sure that we only call it once, by assigning the results of the method in `#updated_dependency_files`, and passing it as an argument to `lockfile_changed?`

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -81,6 +91,42 @@ def update_registry_declaration(new_req, old_req, updated_content)
...
+        new_req = dependency.requirements.first
+        lockfile = lock_file.first
+        content = lockfile.content.dup
+
+        provider_source = new_req[:source][:registry_hostname] + "/" + new_req[:source][:module_identifier]
+        declaration_regex = lockfile_declaration_regex(provider_source)
+        lockfile_dependency_removed = content.sub(declaration_regex, "")
+
+        copy_dir_to_temporary_directory do
```
jurre, line 103: This is a little nitty, but I think this would be slightly more idiomatic for the codebase:

```ruby
  write_dependency_files

  File.write(".terraform.lock.hcl", lockfile_dependency_removed)
  # ...

def write_dependency_files
  dependency_files.each do |file|
    # Do not include the .terraform directory or .terraform.lock.hcl
    next if file.name.include?(".terraform")

    File.write(file.name, file.content)
```

Dockerfile
```diff
@@ -208,6 +208,18 @@ RUN mkdir -p "$RUSTUP_HOME" && chown dependabot:dependabot "$RUSTUP_HOME"
...
+
+USER root
+RUN curl -fsSL https://apt.releases.hashicorp.com/gpg | apt-key add -
+RUN apt-add-repository "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main" \
+  && apt-get update -y \
+  && apt-get install terraform \
+  && terraform -help
+
+
 USER root
```
feelepxyz, line 223:
```suggestion
```

common/lib/dependabot/file_updaters/base.rb
```diff
@@ -40,6 +40,8 @@ def file_changed?(file)
       end
 
       def requirement_changed?(file, dependency)
+        return true if lock_file?(file.name)
+
```
feelepxyz, line 44: Partial review here. It looks like this common class method is changing for a single ecosystem which worries me as it might have unintended consequences in other package managers and make refactoring this harder in future when we've lost this context.

I think it would be better to re-implementing this method change in the terrraform lockfile updater instead.

Althought, is this change required? Looks like it might not be needed anymore now that you're updating the lockfile outside the loop that updates the terraform and terragrunt files.

Dockerfile
```diff
@@ -208,6 +208,18 @@ RUN mkdir -p "$RUSTUP_HOME" && chown dependabot:dependabot "$RUSTUP_HOME"
 USER dependabot
 RUN curl https://sh.rustup.rs -sSf | sh -s -- -y \
   && rustup toolchain install 1.51.0 && rustup default 1.51.0
+
+
+### Terraform
+
+USER root
```
jurre, line 215: Can we run this as `USER dependabot`? We might run into issues using the binary otherwise?

feelepxyz: I think root is required for `apt-` stuff, should be fine as long as we don't need to modify it

common/lib/dependabot/file_updaters/base.rb
Nishnha: I'm also worried about the unintended side-effects of this, so reimplementing it in the terraform ecosystem sounds like the way to go.

The requirement_changed? function is called indirectly in FileUpdater through file_changed?. Without a terraform lockfile dependency the function returns false, so this change resolves that check.

Dockerfile
Nishnha: Looks like root is required for both `apt-key add` and `apt-get`

Screen Shot 2021-06-07 at 11 30 07 AM

Screen Shot 2021-06-07 at 11 47 48 AM

terraform/lib/dependabot/terraform/file_updater.rb
Nishnha: Implemented in https://github.com/dependabot/dependabot-core/pull/3766/commits/41930eff1c4a1b2ca7e37f3ad57591a30d5f5d38.

terraform/spec/dependabot/terraform/file_parser_spec.rb
```diff
@@ -27,7 +27,7 @@
 
       it "raises a helpful error" do
         expect { subject }.to raise_error(Dependabot::DependencyFileNotEvaluatable) do |boom|
-          expect(boom.message).to eq("Invalid registry source specified: 'consul/aws'")
+          expect(boom.message).to eq("Invalid registry source specified: 'invalid/source/hashicorp/consul/aws'")
```
jurre, line 30: I think we'll still want to make sure that a module with `consul/aws` doesn't parse as it's not valid. I think we can safely revert this

jurre, line 30:
```suggestion
          expect(boom.message).to eq("Invalid registry source specified: 'consul/aws'")
```

terraform/spec/fixtures/projects/invalid_registry/main.tf
```diff
@@ -1,5 +1,5 @@
 module "consul" {
 
-  source = "consul/aws"
+  source = "invalid/source/hashicorp/consul/aws"
```
jurre, line 3:
```suggestion
  source = "consul/aws"
```

jurre (approved):
This looks great, I made one suggestion to revert a test change that no longer seems necessary

terraform/lib/dependabot/terraform/file_fetcher.rb
```diff
@@ -45,6 +46,13 @@ def terragrunt_files
           select { |f| f.type == "file" && terragrunt_file?(f.name) }.
           map { |f| fetch_file_from_host(f.name) }
       end
+
+      def lock_file
+        @lock_file ||=
```
feelepxyz, line 51: Looks like this is returning an array
```suggestion
      def lock_files
        @lock_files ||=
```

terraform/lib/dependabot/terraform/file_selector.rb
```diff
@@ -12,6 +12,14 @@ def terragrunt_files
...
-    file_name != ".terraform.lock.hcl" && file_name.end_with?(".hcl")
+    !lock_file?(file_name) && file_name.end_with?(".hcl")
+  end
+
+  def lock_file?(filename)
+    filename == ".terraform.lock.hcl"
+  end
+
+  def lock_file
+    dependency_files.select { |f| lock_file?(f.name) }
```
feelepxyz, line 23: Could use find to just get the first one matching the method name
```suggestion
    dependency_files.find { |f| lock_file?(f.name) }
```

terraform/lib/dependabot/terraform/registry_client.rb
```diff
@@ -86,6 +86,20 @@ def service_url_for(service_key)
 
       attr_reader :hostname, :tokens
 
+      def get(endpoint:)
```
feelepxyz, line 89: Can't see any references to this method, could we ✂️  it?

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -39,15 +44,17 @@ def updated_terraform_file_content(file)
...
-        # Loop through each changed requirement and update the files
+        # Loop through each changed requirement and update the files and lockfile
         reqs.each do |new_req, old_req|
           raise "Bad req match" unless new_req[:file] == old_req[:file]
           next unless new_req.fetch(:file) == file.name
 
           case new_req[:source][:type]
           when "git"
             update_git_declaration(new_req, old_req, content, file.name)
-          when "registry", "provider"
```
feelepxyz, line 50: Could we revert this change?

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -18,14 +19,18 @@ def updated_dependency_files
         updated_files = []
 
         [*terraform_files, *terragrunt_files].each do |file|
-          next unless file_changed?(file)
+          next unless file_changed?(file) || lock_file?(file.name)
 
           updated_content = updated_terraform_file_content(file)
-          raise "Content didn't change!" if updated_content == file.content
+
+          raise "Content didn't change!" unless lockfile_changed? || updated_content != file.content
```
feelepxyz, line 26: Are these changes required? Looks like this `each` loop won't loop over any lockfiles now that the `terragrunt_files` method exlcudes these so wondering if we can revert the above changes and ✂️  the `lockfile_changed?` method?

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -18,14 +19,18 @@ def updated_dependency_files
...
 
           updated_files << updated_file(file: file, content: updated_content)
         end
 
+        updated_files << update_lockfile_declaration(dependency) unless lock_file.empty?
```
feelepxyz, line 31: Small nit but might be clearer to not pass any arguments to this method as both `dependency` and `lock_file` are available in the class

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -81,6 +88,58 @@ def update_registry_declaration(new_req, old_req, updated_content)
         end
       end
 
+      def update_lockfile_declaration(dependency)
+        @updated_lockfile_declarations ||= {}
+        @updated_lockfile_declarations[dependency] ||=
```
feelepxyz, line 93: Does this need to be memoized? The file updater should only have a single dependency set so should always be the same one that gets memoized

jurre: Yeah, it's called twice, so if we don't memoize we'd perform the work twice. It's a relatively slow method as well, so seems worth to memoize.

We might be able to get rid of the `dependency` argument, but it's a bit cumbersome with how the class is set up with inheriting from base

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -81,6 +88,58 @@ def update_registry_declaration(new_req, old_req, updated_content)
...
+              write_dependency_files
+
+              File.write(".terraform.lock.hcl", lockfile_dependency_removed)
+              SharedHelpers.run_shell_command("terraform providers lock #{provider_source}")
+
+              updated_lockfile = File.read(".terraform.lock.hcl")
+              updated_dependency = updated_lockfile.scan(declaration_regex).first
+
+              # Terraform will occasionally update h1 hashes without updating the version of the dependency
+              # Here we make sure the dependency's version actually changes in the lockfile
```
feelepxyz, line 113: This comment implies this only happens occasionally but it looks like we only update the original lockfile content if the version has changed in the updated lockfile. Have I interpreted this correctly?

```suggestion
              # Update the original lockfile content if the version of the dependency has changed in the updated 
              # lockfile along with the h1 hashes. This prevents updating unrelated dependencies.
```

terraform/lib/dependabot/terraform/file_updater.rb
feelepxyz: Gotcha but looks like this is guarded by `file_changed?` in line `22` so shouldn't get this far, as this block only loops over manifests. We then check if we have no updated files again on line `34` after the lockfile update so should cover a lockfile only update scenario.

terraform/lib/dependabot/terraform/file_fetcher.rb
```diff
@@ -45,6 +46,13 @@ def terragrunt_files
...
+          repo_contents(raise_errors: false).
+          select { |f| f.type == "file" && lock_file?(f.name) }.
+          map { |f| fetch_file_from_host(f.name) }
```
feelepxyz, line 54: I think this can be simplified to something like:
```
@lockf_file ||= fetch_file_if_present(".terraform.lock.hcl")
```

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -81,6 +88,58 @@ def update_registry_declaration(new_req, old_req, updated_content)
         end
       end
 
+      def update_lockfile_declaration(dependency)
```
feelepxyz, line 91: It looks like this method can return an unmodified lockfile in which case we'll still add it to the `updated_files` array. I wonder if this method should instead return the updated lockfile content or nil instead and then conditionally build the `updated_file(file: lockfile, content: content)` and add it to the `updated_files` array on line `31` to prevent adding it in cases where it doesn't update.

I'm thinking of the case where we no manifests change and we attempt a lockfile update but for some reason it doesn't update anything so we end up creating a PR for an unchanged lockfile.

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -18,14 +19,18 @@ def updated_dependency_files
         updated_files = []
 
         [*terraform_files, *terragrunt_files].each do |file|
-          next unless file_changed?(file)
+          next unless file_changed?(file) || lock_file?(file.name)
```
jurre, line 22: I don't think this will ever hit a lockfile, right?

terraform/lib/dependabot/terraform/file_fetcher.rb
```diff
@@ -23,6 +23,7 @@ def fetch_files
         fetched_files = []
         fetched_files += terraform_files
         fetched_files += terragrunt_files
+        fetched_files += lock_file
```
feelepxyz, line 26:
```suggestion
        fetched_files += lock_file if lock_file
```

terraform/lib/dependabot/terraform/file_updater.rb
feelepxyz: I had a poke locally and seems like tests pass without these changes. From my understanding we never loop over any lockfiles here so we're only checking if a manifest's requirements have changed. I remember seeing a past commit where the `each` block included lockfiles which would have required these changes.

terraform/lib/dependabot/terraform/registry_client.rb
Nishnha: This was pulled in from a rebase off main. Not sure when it stopped being referenced but tests pass without it.

terraform/lib/dependabot/terraform/update_checker.rb
```diff
@@ -174,6 +174,12 @@ def provider_dependency?
         dependency_source_details.fetch(:type) == "provider"
       end
 
+      def lockfile_dependency?
```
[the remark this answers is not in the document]
Nishnha: Removed

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -21,11 +22,15 @@ def updated_dependency_files
           next unless file_changed?(file)
 
           updated_content = updated_terraform_file_content(file)
-          raise "Content didn't change!" if updated_content == file.content
+
+          raise "Content didn't change!" unless updated_content != file.content
```
feelepxyz, line 26: nit
```suggestion
          raise "Content didn't change!" if updated_content == file.content
```

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -21,11 +22,15 @@ def updated_dependency_files
...
 
           updated_files << updated_file(file: file, content: updated_content)
         end
 
+        updated_files << update_lockfile_declaration
```
feelepxyz, line 31: Nice cleanup! I'm still wondering if this should return the content instead of the `updated_file(file..)`? That way you could only append it to `updated_files` if the content has changed. 

Something like

```
updated_lockfile_content = update_lockfile_declaration
if (updated_lockfile_content && lock_file.content != update_lockfile_content)
  updated_files << updated_file(file: lock_file, content: update_lockfile_content)
end
```

terraform/lib/dependabot/terraform/file_updater.rb
Nishnha: Originally, the `each` block looped over `*lock_files`--it no longer does, so the lockfile guards are no longer necessary.

We now attempt to update the lockfile outside of the `each` block irrespective of changes to the manifest file. Thus, if a new version satisfies the constraints of the manifest file, the manifest files will not not update but the lockfile still will.

terraform/lib/dependabot/terraform/file_updater.rb
Nishnha: Removed the dependency argument and moved the `lock_file.nil?` check to inside the `update_lockfile_declaration` method.

terraform/lib/dependabot/terraform/file_updater.rb
Nishnha: The guards for the lockfile in `FileUpdater#update_dependency_files` were removed including the check for whether the lockfile will update. Now the `update_lockfile_declaration` method is only called once, so I removed the memoization for the `update_lockfile_declaration` method.

terraform/lib/dependabot/terraform/file_updater.rb
feelepxyz: It would be good to add a test case to make sure `updated_dependency_files` doesn't return a lockfile if it doesn't change.

terraform/lib/dependabot/terraform/file_updater.rb
Nishnha: From the [dependency lock docs](https://www.terraform.io/docs/language/dependency-lock.html):

> Due to the limited scope of the zh: scheme, Terraform will opportunistically add in the corresponding h1: checksums as it learns of them

You are correct that we only update the original lockfile content if the version has changed in the updated lockfile. However, the reason the check is there is to prevent dependabot from updating the lockfile hash when the version has not changed.

feelepxyz (commented):
This is looking very close to done! Only have one comment remaining about checking the lockfile content has changed 🙌

terraform/lib/dependabot/terraform/file_updater.rb
Nishnha: Good catch!

terraform/lib/dependabot/terraform/file_updater.rb
Nishnha: Implemented @feelepxyz's  suggestion and created a test/fixture to make sure `updated_dependency_files` doesn't return a lockfile if it doesn't change.

The test uses a terraform project fixture that is fully up-to-date. However, since no files change after running dependabot on the fixture, it raises a "No files changed!" error. 

The error makes the function quit early and prevents the test from directly checking whether the lockfile is unmodified or appended to the `updated_files` array.

So instead, I am testing for the error, which indicates that the lockfile didn't change.

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -21,10 +22,18 @@ def updated_dependency_files
           next unless file_changed?(file)
 
           updated_content = updated_terraform_file_content(file)
+
           raise "Content didn't change!" if updated_content == file.content
 
           updated_files << updated_file(file: file, content: updated_content)
         end
+        updated_lockfile_content = update_lockfile_declaration
+
```
feelepxyz, line 31: small style nit, would usually leave a empty line after `end` and remove the empty line here
```suggestion
        
        updated_lockfile_content = update_lockfile_declaration
```

feelepxyz (approved):
🎉  Thanks for incorporating the feedback and for adding the test case! This looks good to go to me :shipit:

terraform/lib/dependabot/terraform/file_updater.rb
Nishnha: Rubocop did not like this

Dockerfile
```diff
@@ -208,6 +208,18 @@ RUN mkdir -p "$RUSTUP_HOME" && chown dependabot:dependabot "$RUSTUP_HOME"
...
+RUN curl -fsSL https://apt.releases.hashicorp.com/gpg | apt-key add -
+RUN apt-add-repository "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main" \
+  && apt-get update -y \
+  && apt-get install terraform \
```
jurre, line 219: Should we pin the version here?

jurre: Meh, whatevs let's do it in a follow-up PR

Nishnha: Pinned to 1.0.0 https://github.com/dependabot/dependabot-core/pull/3766/commits/81a424eac4bd9086a86fe9cd8d18ccf09d5bf2f2

PR #3718, opened 2021-05-13
Allow multiple js sources

xlgmokha:
# Why is this needed?

To support js projects that have configured both `registry.npmjs.org` and
`registry.yarnpkg.com` sources.

## What does this do?

It removes a check that was raising an error when multiple
non-git sources were detected while determining if the
package comes from a git source.

```plaintext
Checking if x 10.5.12 needs updating
Error processing x (RuntimeError)
Multiple sources! {:type=>"registry", :url=>"https://registry.yarnpkg.com"}, {:type=>"registry", :url=>"https://registry.npmjs.org"}
vendor/ruby/2.6.0/gems/dependabot-common-0.146.0/lib/dependabot/git_commit_checker.rb:277:in `dependency_source_details'
vendor/ruby/2.6.0/gems/dependabot-common-0.146.0/lib/dependabot/git_commit_checker.rb:36:in `git_dependency?'
vendor/ruby/2.6.0/gems/dependabot-npm_and_yarn-0.146.0/lib/dependabot/npm_and_yarn/update_checker.rb:244:in `git_dependency?'
vendor/ruby/2.6.0/gems/dependabot-npm_and_yarn-0.146.0/lib/dependabot/npm_and_yarn/update_checker.rb:20:in `latest_version'
lib/dependabot/updater.rb:459:in `log_checking_for_update'
lib/dependabot/updater.rb:147:in `check_and_update_pull_request'
lib/dependabot/updater.rb:106:in `check_and_update_existing_pr_with_error_handling'
lib/dependabot/updater.rb:53:in `run'
lib/dependabot/update_files_job.rb:17:in `perform_job'
lib/dependabot/base_job.rb:28:in `run'
bin/update_files.rb:21:in `<main>'
Finished job processing
```

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

PR #3780, opened 2021-05-24
npm: Handle multiple sources in the update checker

feelepxyz:
Handle multiple sources in the update checker, following on from https://github.com/dependabot/dependabot-core/pull/3718 where we started doing this.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb
@@ -297,9 +297,8 @@ module Dependabot
 
       def dependency_source_details
         sources =
-          dependency.requirements.map { |r| r.fetch(:source) }.uniq.compact
-
-        raise "Multiple sources! #{sources.join(', ')}" if sources.count > 1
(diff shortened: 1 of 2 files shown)


Nishnha:
Defaulting to a private registry source makes sense 🆗

Is the purpose of multiple sources to have a backup registry in case the main one goes down? If that is the case, what happens if the private registry does not return a 200 status code?
Maybe we should add a test where the private registry fails and a secondary source is used.

Other than that, LGTM 👍

feelepxyz:
> Is the purpose of multiple sources to have a backup registry in case the main one goes down? If that is the case, what happens if the private registry does not return a 200 status code?

It's a bit weird in npm/yarn land as you can have a monorepo setup that links several `package.json` files together and some might be installed from a different registry, e.g. one using yarn's npm mirror and another the main npm registry and a third a private npm mirror where each package has been vetted internally. It would be strange to have the same package fetched from both private and public sources in the same repo but it _could_ happen in theory so seems best to always prefer the private one.

I'm not sure we want to fall back to public registry as it could mean we fetch the wrong package if you've misconfigured your project to use both a public and private source but only wanted the private one. We could revisit if we get feedback on this causing us to do the wrong thing.

jurre (approved):

PR #3781, opened 2021-05-24
npm: handle latest version requirement

feelepxyz:
Handle `latest` version requirements which are valid in npm:

```
"dependencies": {
  "etag" : "latest"
}
```

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/requirement.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/requirement.rb
@@ -8,6 +8,7 @@ module Dependabot
     class Requirement < Gem::Requirement
       AND_SEPARATOR = /(?<=[a-zA-Z0-9*])\s+(?:&+\s+)?(?!\s*[|-])/.freeze
       OR_SEPARATOR = /(?<=[a-zA-Z0-9*])\s*\|+/.freeze
+      LATEST_REQUIREMENT = "latest"
 
       # Override the version pattern to allow a 'v' prefix
       quoted = OPS.keys.map { |k| Regexp.quote(k) }.join("|")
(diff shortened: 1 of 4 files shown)


common/lib/dependabot/update_checkers/base.rb
```diff
@@ -293,6 +293,8 @@ def version_from_requirements
           reject { |req_array| req_array.first.start_with?("<") }.
           map(&:last).
           max
+      rescue requirement_class::BadRequirementError
```
jurre, line 296: Should we rescue *any* `BadRequirementError`? I'd expect some more specific check on `latest` being set, but maybe nothing else is possible in this case?

But looking at the code, it seems like this would now no longer raise:

```json
"dependencies": {
  "etag" : "some-invalid-string-that-npm-doesnt-know"
}
```

feelepxyz: Oh yes good catch! Will have a think about how to handle this better in the npm update checker 👍

npm_and_yarn/lib/dependabot/npm_and_yarn/requirement.rb
```diff
@@ -17,6 +18,7 @@ class Requirement < Gem::Requirement
       PATTERN = /\A#{PATTERN_RAW}\z/.freeze
 
       def self.parse(obj)
+        return ["=", nil] if obj.is_a?(String) && obj.strip == LATEST_REQUIREMENT
```
feelepxyz, line 21: This is a bit obscure but seems to work and not blow up, we can't actually compare `latest` to anything

feelepxyz: `satisfied_by?` always returns false

jurre (approved):
Nice work 👍

PR #3786, opened 2021-05-25
Terraform: handle implicit (v0.12 style) provider sources

jurre:
When leaving out a `source = ..` declaration for a Terraform provider,
it falls back to assuming the default `hashicorp` namespace and the
declared local name as the package name.

This is unrecommended, but we do come across it.

Previously Dependabot would fail, with this change it parses + updates
these cases correctly.

### DIFF
--- a/terraform/lib/dependabot/terraform/file_parser.rb
+++ b/terraform/lib/dependabot/terraform/file_parser.rb
@@ -20,6 +20,10 @@ module Dependabot
       include FileSelector
 
       ARCHIVE_EXTENSIONS = %w(.zip .tbz2 .tgz .txz).freeze
+      DEFAULT_REGISTRY = "registry.terraform.io"
+      DEFAULT_NAMESPACE = "hashicorp"
+      # https://www.terraform.io/docs/language/providers/requirements.html#source-addresses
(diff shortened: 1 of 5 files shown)


terraform/lib/dependabot/terraform/file_parser.rb
```diff
@@ -54,10 +58,10 @@ def parse
 
       private
 
-      def build_terraform_dependency(file, name, details, provider)
+      def build_terraform_dependency(file, name, details)
         details = details.is_a?(Array) ? details.first : details
```
feelepxyz, line 62: When is this an array? Is the same not true for `build_provider_dependency`?

terraform/lib/dependabot/terraform/file_parser.rb
```diff
@@ -82,8 +86,31 @@ def build_terraform_dependency(file, name, details, provider)
         )
       end
 
+      def build_provider_dependency(file, name, details = {})
```
feelepxyz, line 89: 👍  on splitting this out

feelepxyz (approved):
LGTM 👍  only had a non-blocking q about the previous method arg

terraform/lib/dependabot/terraform/file_parser.rb
jurre: I don't think it is anymore, will yank it out 👍

PR #3790, opened 2021-05-25
feat(Terraform): Inject authorization header

xlgmokha:
# Why is this needed?

We need this to be able to connect to private terraform registries that are
secured with a bearer token.

## What does this do?

This change searches for credentials that match the provided host and injects
that as an [Authorization header](https://www.terraform.io/docs/cloud/api/index.html#authentication).

## Type of change

- [X] New feature (non-breaking change which adds functionality)

### DIFF
--- a/terraform/spec/dependabot/terraform/registry_client_spec.rb
+++ b/terraform/spec/dependabot/terraform/registry_client_spec.rb
@@ -33,35 +33,6 @@ RSpec.describe Dependabot::Terraform::RegistryClient do
     )
   end
 
-  let(:provider_dependency) do
-    Dependabot::Dependency.new(
-      name: "hashicorp/aws",
-      version: "0.9.3",
-      package_manager: "terraform",
-      previous_version: "3.19.0",
-      requirements: [{
-        requirement: "3.40.0",
-        groups: [],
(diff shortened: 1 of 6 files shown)


terraform/spec/dependabot/terraform/registry_client_spec.rb
```diff
@@ -33,73 +35,108 @@
...
+
   it "fetches module versions", :vcr do
-    client = described_class.new(hostname: described_class::PUBLIC_HOSTNAME)
     response = client.all_module_versions(identifier: "hashicorp/consul/aws")
     expect(response.max).to eq(Gem::Version.new("0.9.3"))
   end
 
+  it "fetches module versions from a custom registry" do
+    hostname = "registry.example.org"
+    stub_request(:get, "https://#{hostname}/v1/modules/hashicorp/consul/aws/versions").
```
jurre, line 83: Do we not VCR these because we don't have access to an actual private registry?

xlgmokha: > Do we not VCR these because we don't have access to an actual private registry?

Yes, that is one reason.

PR #3796, opened 2021-05-26
Yarn: use .yarnrc file if present

feelepxyz:
Fixes a bug when trying to force yarn to use `registry.npmjs.org`
instead of `registry.yarnpkg.com`.

Previously if a .yarnrc file was present with registry `registry
"https://registry.npmjs.org"` it would be ignored during updates as we
run yarn programtically and didn't instruct it to use the config file.

I've also tested a dry-run without a `.yarnrc` file present.


jurre (approved):
🎉 nice one!

npm_and_yarn/helpers/lib/yarn/subdependency-updater.js
```diff
@@ -39,6 +39,7 @@ async function getLockfileEntryForUpdate(depName, depVersion) {
     cwd: directory,
     nonInteractive: true,
     enableDefaultRc: true,
+    extraneousYarnrcFiles: [".yarnrc"],
```
feelepxyz, line 42: This is the only documentation I've found for this https://github.com/yarnpkg/yarn/blob/25d5526c8995443b03c113c44f9c8e9a79c2c074/src/config.js#L57

PR #3802, opened 2021-05-27
Tests: avoid squatted repositories

jurre:
A few repositories we expect not to exist on npm have been squatted.
Where possible this updates those tests to use a namespace that we own,
and for one fixture we explicitly tested against an unscoped package,
those tests have been removed as we do not expect behavior to differ
between scoped and unscoped packages.


feelepxyz (approved):
This seems good to me 👍  I'm up for removing this yarn spec, it's for v1 which is unlikely to change much going forward

Issue #2963, opened 2021-01-08
"Illformed requirement" error when using `Gem::Requirement` for `spec.required_ruby_version`


ybiquitous:
**Package manager/ecosystem**
ruby:bundler

**Manifest contents prior to update**
`*.gemspec` file

**Updated dependency**
All Ruby gems in my project

**What you expected to see, versus what you actually saw**
I have a `*.gemspec` file including the following `spec.required_ruby_version`:

```ruby
spec.required_ruby_version = Gem::Requirement.new(">= 2.7.2", "< 3.0.0")
```

https://github.com/sider/runners/blob/495a312a369dd34bf227211fa40b60e559753ac4/runners.gemspec#L15

This is a valid specifier, but Dependabot does not seem to parse it. 😢 

For example, the latest version of Bundler generates a `*.gemspec` file including `Gem::Requirement`:

```console
$ bundle -v
Bundler version 2.2.4

$ bundle gem foo
...
Gem 'foo' was successfully created. For more information on making a RubyGem visit https://bundler.io/guides/creating_gem.html

$ grep required_ruby_version foo/foo.gemspec
  spec.required_ruby_version = Gem::Requirement.new(">= 2.3.0")
```

**Images of the diff or a link to the PR, issue or logs**
The following is a part of the log in <https://github.com/sider/runners/network/updates/80287287>:

```
updater | ERROR <job_80287287> Illformed requirement [#<Gem::Requirement:0x0000556ac7d92ab0 @requirements=[[">=", #<Gem::Version "2.7.2">], ["<", #<Gem::Version "3.0.0">]]>]
updater | ERROR <job_80287287> /usr/lib/ruby/2.6.0/rubygems/requirement.rb:105:in `parse'
updater | ERROR <job_80287287> /usr/lib/ruby/2.6.0/rubygems/requirement.rb:135:in `block in initialize'
updater | ERROR <job_80287287> /usr/lib/ruby/2.6.0/rubygems/requirement.rb:135:in `map!'
updater | ERROR <job_80287287> /usr/lib/ruby/2.6.0/rubygems/requirement.rb:135:in `initialize'
updater | ERROR <job_80287287> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-bundler-0.129.5/lib/dependabot/bundler/file_updater/ruby_requirement_setter.rb:52:in `new'
[...]

PR #3809, opened 2021-05-28
Bundler: handle gemspec required ruby version ranges

feelepxyz:
Handle `required_ruby_version` ranges in gemspecs:
`required_ruby_version = ">= 2.2, < 4.0"` and
`required_ruby_version = Gem::Requirement.new(">= 2.1.8", "< 4.0.0")`

This would previously raise a `Gem::Requirement::BadRequirementError`.

Note: The `Bundler::Requirement` class splits requirements by `,`.

I've done an audit of other uses of `Gem::Requirement` but these seem ok
as we control the range we create so shouldn't run into similar bugs
there.

Fixes https://github.com/dependabot/dependabot-core/issues/2963

### DIFF
--- a/bundler/lib/dependabot/bundler/file_updater/ruby_requirement_setter.rb
+++ b/bundler/lib/dependabot/bundler/file_updater/ruby_requirement_setter.rb
@@ -2,6 +2,7 @@
 
 require "parser/current"
 require "dependabot/bundler/file_updater"
+require "dependabot/bundler/requirement"
 
 module Dependabot
   module Bundler
@@ -49,7 +50,11 @@ module Dependabot
         end
 
         def ruby_version
-          requirement = Gem::Requirement.new(ruby_requirement)
+          requirement = if !ruby_requirement.is_a?(Gem::Requirement)
+                          Dependabot::Bundler::Requirement.new(ruby_requirement)
+                        else
+                          ruby_requirement
+                        end
 
           ruby_version =
             RUBY_VERSIONS.
--- a/bundler/lib/dependabot/bundler/update_checker/version_resolver.rb
+++ b/bundler/lib/dependabot/bundler/update_checker/version_resolver.rb
@@ -197,8 +197,8 @@ module Dependabot
           # required Ruby version.
           return false unless ruby_requirement
 
-          ruby_requirement = Gem::Requirement.new(ruby_requirement)
-          current_ruby_version = Gem::Version.new(details[:ruby_version])
+          ruby_requirement = Dependabot::Bundler::Requirement.new(ruby_requirement)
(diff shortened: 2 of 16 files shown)


bundler/lib/dependabot/bundler/file_updater/ruby_requirement_setter.rb
```diff
@@ -49,7 +50,11 @@ def declares_ruby_version?(node)
         end
 
         def ruby_version
-          requirement = Gem::Requirement.new(ruby_requirement)
+          requirement = if !ruby_requirement.is_a?(Gem::Requirement)
+                          Bundler::Requirement.new(ruby_requirement)
```
jurre, line 54: Let's make it super clear that this is not `::Bundler`?

```suggestion
                          Dependabot::Bundler::Requirement.new(ruby_requirement)
```

Alternatively we could drop the namespaces altogether since we're already in `Dependabot::Bundler`, but I prefer being explicit here

bundler/lib/dependabot/bundler/update_checker/version_resolver.rb
```diff
@@ -197,8 +197,8 @@ def ruby_version_incompatible?(details)
           # required Ruby version.
           return false unless ruby_requirement
 
-          ruby_requirement = Gem::Requirement.new(ruby_requirement)
-          current_ruby_version = Gem::Version.new(details[:ruby_version])
+          ruby_requirement = Bundler::Requirement.new(ruby_requirement)
+          current_ruby_version = Bundler::Version.new(details[:ruby_version])
```
jurre, line 201: Same thing here:

```suggestion
          ruby_requirement = Dependabot::Bundler::Requirement.new(ruby_requirement)
          current_ruby_version = Dependabot::Bundler::Version.new(details[:ruby_version])
```

bundler/lib/dependabot/bundler/file_updater/ruby_requirement_setter.rb
feelepxyz: Yes good point!

jurre (approved):
Makes sense 👍 thanks for the thorough test coverage on it

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
```diff
@@ -238,50 +238,62 @@
         # or greater.
         let(:dependency_files) { bundler_project_dependency_files("legacy_ruby") }
         its([:version]) { is_expected.to eq(Gem::Version.new("1.4.6")) }
+      end
```
feelepxyz, line 241: Review here is easier with https://github.com/dependabot/dependabot-core/pull/3809/files?diff=unified&w=1

PR #3819, opened 2021-05-31
Bundler tests: Use git sources for organizations that we own

jurre:
Bundler git dependencies have the possibility to execute code during the
installation phase, to ensure this does not happen in our test suite,
this updates any tests that pull in a git dependency to reference an
organization/account that we own.

This has the added benefit of protecting us against those repositories
being renamed, moved or deleted.


jurre:
I'm expecting CI to fail an initial run, there are a few things I still need to fix up

bundler/spec/dependabot/bundler/file_parser_spec.rb
```diff
@@ -19,7 +19,7 @@
   let(:source) do
     Dependabot::Source.new(
       provider: "github",
-      repo: "gocardless/bump",
+      repo: "dependabot-fixtures/bump",
```
thepwagner, line 22: Noice: line 1 and I'm already digging this diff!

bundler/spec/dependabot/bundler/update_checker_spec.rb
```diff
@@ -1160,7 +1160,7 @@
           before do
             stub_request(
               :get,
-              "https://github.com/fundingcircle/prius.git/info/refs"\
+              "https://github.com/no-exist-sorry/prius.git/info/refs"\
```
thepwagner, line 1163: Question: could this be `dependabot-fixtures/sorry-missing-prius.git`?

I'd dig if we could minimize the number of GitHub orgs that we need to maintain, `dependabot-fixtures` is _pretty_ close to this use case.

Meh as there isn't much "maintenance" required in the `no-exist-sorry` organization: it just need to exist and not be pwned.

thepwagner (approved):
`s/${org}/dependabot-fixtures/g` :shipit: , feel free to dismiss if changes get deeper than that.

I wondered if varying the repo within `dependabot-fixtures` would be easier than introducing the `no-exist-sorry` org, but don't feel strongly either way.

bundler/spec/dependabot/bundler/update_checker_spec.rb
jurre: Right, yeah I registered that org last week since we had some prior tests using it, and registering the org was easier than changing the tests, in this case `fundingcircle/prius` is meant to be _private_, but as far as the rubygems client knows, non-existing and private are the same thing.

I think changing to a non existing repo name should be fine here, I'll make this change 👍

jurre: I ended up keeping on test on `no-exist-sorry` (which we own) because I couldn't get our tests to pass for the life of me with a different repo name. I tried _everything_ that I could think of and kept getting `The following git URLs could not be retrieved: git@github.com:dependabot-fixtures/private-prius`.

This suspects me to believe that the test in question is not entirely correct and this might be an edge case in bundler, but in the interest of keeping the set of changes here limited I ended up using the `no-exist-sorry` org.

PR #3835, opened 2021-06-02
Hex tests: Use git dependencies of repositories we own

jurre:
To prevent these repositories from being squatted or from otherwise changing
under our feet in unexpected ways, prefer using repositories that we own
as git dependencies in our tests.


xlgmokha (approved):
LGTM

Issue #3813, opened 2021-05-30
Issue when updating npm github packages


ImRodry:
**Package ecosystem**

npm

**Package manager version**

 6.14.13

**Language version**

Node 14.17.0

**Manifest location and content prior to update**

You can find the diff on [this](https://github.com/Hypixel-Translators/hypixel-translators-bot/pull/327) PR.

```json
{
  "name": "hypixel-translators",
  "version": "3.0.0",
  "main": "dist/index",
  "license": "MIT",
  "scripts": {
    "start": "node --trace-warnings dist/index.js",
    "test": "node --unhandled-rejections=strict dist/index.js",
    "compile": "rimraf dist/ && tsc -w",
    "postinstall": "tsc"
  },
  "repository": {
    "type": "git",
    "url": "git://github.com/Hypixel-Translators/hypixel-translators-bot.git"
  },
  "dependencies": {
    "canvas": "^2.8.0",
    "country-emoji": "^1.5.6",
    "countryjs": "^1.8.0",
    "discord.js": "github:discordjs/discord.js",
    "dotenv": "^9.0.2",
    "mongodb": "^3.6.6",
    "node-fetch": "^2.6.1",
    "puppeteer": "^9.1.1",
    "source-map-support": "^0.5.19",
    "typescript": "^4.2.4",
    "uuid": "^8.3.2"
  },
  "engines": {
    "node": "14.x"
  },
  "devDependencies": {
    "@types/countryjs": "^1.8.1",
    "@types/mongodb": "^3.6.12",
    "@types/node-fetch": "^2.5.10",
    "@types/uuid": "^8.3.0",
    "@types/ws": "^7.4.2",
    "discord-api-types": "^0.18.1",
    "rimraf": "^3.0.2"
  }
}
```

[package.json](https://github.com/Hypixel-Translators/hypixel-translators-bot/blob/738da28e765e5fc8b47b6da12b7efd481d520925/package.json)

**dependabot.yml content**

```yaml
version: 2
updates:
- package-ecosystem: npm
  directory: "/"
  schedule:
    interval: daily
    time: "04:00"
    timezone: Europe/Lisbon
  open-pull-requests-limit: 10
  assignees:
  - imrodry
```

[dependabot.yml](https://github.com/Hypixel-Translators/hypixel-translators-bot/blob/8d41b4071abdf27fff397d36ebdaace9223687af/.github/dependabot.yml)

**Updated dependency**

[...]

PR #3837, opened 2021-06-02
fix(npm): Prevent unnecessary hash pinning in lock file constraint

xlgmokha:
# Why is this needed?

To make sure that Dependabot does not pin to a specific git hash when the `package.json` has not pinned to one.

## What does this do?

It updates the npm helper to check for a pinned hash in git based version constraints.

Fixes https://github.com/dependabot/dependabot-core/issues/3813

### Screenshots

Before:

```bash
[dependabot-core-dev] ~/dependabot-core $ ./bin/dry-run.rb npm_and_yarn xlgmokha/hypixel-translators-bot --dep discord.js --branch core-3813
Resolving dependencies...
=> fetching dependency files
=> dumping fetched dependency files: ./dry-run/xlgmokha/hypixel-translators-bot/core-3813/
=> parsing dependency files
=> updating 1 dependencies: discord.js

=== discord.js (ab82cafcde0ee259a32ef14303c1b4a64dea8fae)
 => checking for updates 1/1
 => latest available version is 7b85a7259f563ab14ae6c0a665a3dd43c486fde4
 => latest allowed version is 7b85a7259f563ab14ae6c0a665a3dd43c486fde4
 => updating discord.js from ab82cafcde0ee259a32ef14303c1b4a64dea8fae to 7b85a7259f563ab14ae6c0a665a3dd43c486fde4

    ± package-lock.json
    ~~~
    37a38,42
    >     "@sapphire/async-queue": {
    >       "version": "1.1.2",
    >       "resolved": "https://registry.npmjs.org/@sapphire/async-queue/-/async-queue-1.1.2.tgz",
    >       "integrity": "sha512-NkR7AzC9uyb++tMIZgG4X0ci8JM1rnvNmKbLwY42RgotRV8JGUntZ7ZpR7MN7p5zPlVdKo/YmOqcCCsBJ6LNuw=="
    >     },
    88d92
    <       "dev": true,
    308,309c312
    <       "integrity": "sha512-hNC38R9ZF4uaujaZQtQfm5CdQO58uhdkoHQAVvMfIL0LgOSZeW575W8H6upngQOuoxWd8tiRII3LLJm9zuQKYg==",
    <       "dev": true
    ---
    >       "integrity": "sha512-hNC38R9ZF4uaujaZQtQfm5CdQO58uhdkoHQAVvMfIL0LgOSZeW575W8H6upngQOuoxWd8tiRII3LLJm9zuQKYg=="
    312,313c315,316
    <       "version": "github:discordjs/discord.js#ab82cafcde0ee259a32ef14303c1b4a64dea8fae",
    <       "from": "github:discordjs/discord.js",
    ---
    >       "version": "github:discordjs/discord.js#7b85a7259f563ab14ae6c0a665a3dd43c486fde4",
    >       "from": "github:discordjs/discord.js#7b85a7259f563ab14ae6c0a665a3dd43c486fde4",
    316a320,321
    >         "@sapphire/async-queue": "^1.1.2",
    >         "@types/ws": "^7.4.4",
    317a323
    >         "discord-api-types": "^0.18.1",
    319c325
    <         "prism-media": "^1.2.2",
    ---
    >         "prism-media": "^1.2.9",
    321c327,334
    <         "ws": "^7.3.1"
    ---
    >         "ws": "^7.4.6"
    >       },
    >       "dependencies": {
    >         "ws": {
    >           "version": "7.4.6",
    >           "resolved": "https://registry.npmjs.org/ws/-/ws-7.4.6.tgz",
    >           "integrity": "sha512-YmhHDO4MzaDLB+M9ym/mDA5z0naX8j7SIlT8f8z+I0VtzsRbekxEutHSme7NPS2qE8StCYQNUnfWdXta/Yu85A=="
    >         }
    ~~~
```

After:

```bash
$ ./bin/dry-run.rb npm_and_yarn xlgmokha/hypixel-translators-bot --dep 'discord.js' --branch core-3813
=> fetching dependency files
=> dumping fetched dependency files: ./dry-run/xlgmokha/hypixel-translators-bot/core-3813/
=> parsing dependency files
=> updating 1 dependencies: discord.js

=== discord.js (ab82cafcde0ee259a32ef14303c1b4a64dea8fae)
 => checking for updates 1/1
 => latest available version is 51551f544b80d7d27ab0b315da01dfc560b2c115
 => latest allowed version is 51551f544b80d7d27ab0b315da01dfc560b2c115
 => updating discord.js from ab82cafcde0ee259a32ef14303c1b4a64dea8fae to 51551f544b80d7d27ab0b315da01dfc560b2c115

    ± package-lock.json
    ~~~
    37a38,42
    >     "@sapphire/async-queue": {
    >       "version": "1.1.2",
    >       "resolved": "https://registry.npmjs.org/@sapphire/async-queue/-/async-queue-1.1.2.tgz",
    >       "integrity": "sha512-NkR7AzC9uyb++tMIZgG4X0ci8JM1rnvNmKbLwY42RgotRV8JGUntZ7ZpR7MN7p5zPlVdKo/YmOqcCCsBJ6LNuw=="
    >     },
    88d92
    <       "dev": true,
    308,309c312
    <       "integrity": "sha512-hNC38R9ZF4uaujaZQtQfm5CdQO58uhdkoHQAVvMfIL0LgOSZeW575W8H6upngQOuoxWd8tiRII3LLJm9zuQKYg==",
    <       "dev": true
    ---
    >       "integrity": "sha512-hNC38R9ZF4uaujaZQtQfm5CdQO58uhdkoHQAVvMfIL0LgOSZeW575W8H6upngQOuoxWd8tiRII3LLJm9zuQKYg=="
    312c315
    <       "version": "github:discordjs/discord.js#ab82cafcde0ee259a32ef14303c1b4a64dea8fae",
    ---
    >       "version": "github:discordjs/discord.js#fdad14099779e61cb84dcd1cb2497e0e853a6144",
    316a320,321
    >         "@sapphire/async-queue": "^1.1.2",
    >         "@types/ws": "^7.4.4",
    317a323
    >         "discord-api-types": "^0.18.1",
    319c325
    <         "prism-media": "^1.2.2",
    ---
    >         "prism-media": "^1.2.9",
    321c327,334
    <         "ws": "^7.3.1"
    ---
    >         "ws": "^7.4.6"
    >       },
    >       "dependencies": {
    >         "ws": {
    >           "version": "7.4.6",
    >           "resolved": "https://registry.npmjs.org/ws/-/ws-7.4.6.tgz",
    >           "integrity": "sha512-YmhHDO4MzaDLB+M9ym/mDA5z0naX8j7SIlT8f8z+I0VtzsRbekxEutHSme7NPS2qE8StCYQNUnfWdXta/Yu85A=="
    >         }
    ~~~
```

## Type of change

### DIFF
--- a/npm_and_yarn/helpers/lib/npm6/updater.js
+++ b/npm_and_yarn/helpers/lib/npm6/updater.js
@@ -137,6 +137,8 @@ function installArgs(
     // requirement when installing the new version
     if (existingVersionRequirement.match(desiredVersion)) {
       return `${depName}@${existingVersionRequirement}`;
+    } else if (!existingVersionRequirement.includes("#")) {
+      return `${depName}@${existingVersionRequirement}`;
     } else {
       return `${depName}@${existingVersionRequirement.replace(
         /#.*/,
--- a/npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater_spec.rb
+++ b/npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater_spec.rb
@@ -500,4 +500,44 @@ RSpec.describe Dependabot::NpmAndYarn::FileUpdater::NpmLockfileUpdater do
       end
     end
   end
+
+  context "when updating a git source dependency that is not pinned to a hash" do
+    subject { JSON.parse(updated_npm_lock_content) }
+
+    let(:files) { project_dependency_files("npm6/ghpr_no_hash_pinning") }
+    let(:dependency_name) { "discord.js" }
+    let(:version) { "HEAD" }
+    let(:previous_version) { "ab82cafcde0ee259a32ef14303c1b4a64dea8fae" }
+    let(:requirements) do
+      [{
+        file: "package.json",
+        requirement: nil,
+        groups: ["dependencies"],
+        source: {
+          type: "git",
+          url: "https://github.com/discordjs/discord.js",
+          branch: nil,
+          ref: "master"
+        }
+      }]
+    end
+    let(:previous_requirements) do
+      [{
+        file: "package.json",
+        requirement: nil,
+        groups: ["dependencies"],
+        source: {
+          type: "git",
+          url: "https://github.com/discordjs/discord.js",
+          branch: nil,
+          ref: "master"
+        }
+      }]
+    end
+
+    it "pins the version to a hash and ensures that the `from` field matches the original constraint" do
+      expect(subject["dependencies"]["discord.js"]["version"]).to match(%r{github:discordjs/discord.js#[0-9a-z]{40}})
+      expect(subject["dependencies"]["discord.js"]["from"]).to eq("github:discordjs/discord.js")
+    end
+  end
 end
--- a/npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater_spec.rb
+++ b/npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater_spec.rb
@@ -335,7 +335,7 @@ RSpec.describe Dependabot::NpmAndYarn::FileUpdater do
           parsed_package_lock = JSON.parse(updated_npm_lock.content)
           expect(parsed_package_lock["dependencies"]["is-number"]["version"]).
             to eq("github:jonschlinkert/is-number#"\
-                  "0c6b15a88bc10cd47f67a09506399dfc9ddc075d")
+                  "98e8ff1da1a89f93d1397a24d7413ed15421c139")
 
           expect(updated_yarn_lock.content).to include(
             "is-number@jonschlinkert/is-number:"
@@ -357,7 +357,7 @@ RSpec.describe Dependabot::NpmAndYarn::FileUpdater do
             parsed_package_lock = JSON.parse(updated_npm_lock.content)
             expect(parsed_package_lock["dependencies"]["is-number"]["version"]).
               to eq("git+https://github.com/jonschlinkert/is-number.git#"\
-                    "0c6b15a88bc10cd47f67a09506399dfc9ddc075d")
+                    "98e8ff1da1a89f93d1397a24d7413ed15421c139")
 
             expect(updated_yarn_lock.content).to include("is-number")
             expect(updated_yarn_lock.content).to include("0c6b15a88b")
@@ -436,7 +436,7 @@ RSpec.describe Dependabot::NpmAndYarn::FileUpdater do
                 parsed_package_lock["dependencies"]["slick-carousel"]["version"]
               expect(npm_lockfile_version).
                 to eq("git://github.com/brianfryer/slick.git#"\
-                      "a2aa3fec335c50aceb58f6ef6d22df8e5f3238e1")
+                      "fc6f7d860844ad562df5b94b5918b58bab067751")
 
               expect(updated_yarn_lock.content).
                 to include('slick-carousel@git://github.com/brianfryer/slick":')
@@ -457,7 +457,7 @@ RSpec.describe Dependabot::NpmAndYarn::FileUpdater do
                 parsed_package_lock["dependencies"]["is-number"]["version"]
               expect(npm_lockfile_version).
                 to eq("git+ssh://git@github.com/jonschlinkert/is-number.git#"\
-                      "0c6b15a88bc10cd47f67a09506399dfc9ddc075d")
+                      "98e8ff1da1a89f93d1397a24d7413ed15421c139")
 
               expect(updated_yarn_lock.content).to include("is-number")
               expect(updated_yarn_lock.content).to include("0c6b15a88bc")
@@ -1264,7 +1264,7 @@ RSpec.describe Dependabot::NpmAndYarn::FileUpdater do
           parsed_package_lock = JSON.parse(updated_npm_lock.content)
           expect(parsed_package_lock["dependencies"]["is-number"]["version"]).
             to eq("git+https://github.com/jonschlinkert/is-number.git#"\
-                  "0c6b15a88bc10cd47f67a09506399dfc9ddc075d")
+                  "98e8ff1da1a89f93d1397a24d7413ed15421c139")
         end
       end
 
--- a/npm_and_yarn/spec/fixtures/projects/npm6/ghpr_no_hash_pinning/package.json
+++ b/npm_and_yarn/spec/fixtures/projects/npm6/ghpr_no_hash_pinning/package.json
@@ -1,0 +1,10 @@
+{
+  "name": "example",
+  "version": "1.0.0",
+  "dependencies": {
+    "discord.js": "github:discordjs/discord.js"
+  },
+  "engines": {
+    "node": "14.x"
+  }
+}
--- a/npm_and_yarn/spec/fixtures/projects/npm6/ghpr_no_hash_pinning/package-lock.json
+++ b/npm_and_yarn/spec/fixtures/projects/npm6/ghpr_no_hash_pinning/package-lock.json
@@ -1,0 +1,100 @@
+{
+  "name": "example",
+  "version": "1.0.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@discordjs/collection": {
+      "version": "0.1.6",
+      "resolved": "https://registry.npmjs.org/@discordjs/collection/-/collection-0.1.6.tgz",
+      "integrity": "sha512-utRNxnd9kSS2qhyivo9lMlt5qgAUasH2gb7BEOn6p0efFh24gjGomHzWKMAPn2hEReOPQZCJaRKoURwRotKucQ=="
+    },
+    "@discordjs/form-data": {
+      "version": "3.0.1",
+      "resolved": "https://registry.npmjs.org/@discordjs/form-data/-/form-data-3.0.1.tgz",
+      "integrity": "sha512-ZfFsbgEXW71Rw/6EtBdrP5VxBJy4dthyC0tpQKGKmYFImlmmrykO14Za+BiIVduwjte0jXEBlhSKf0MWbFp9Eg==",
+      "requires": {
+        "asynckit": "^0.4.0",
+        "combined-stream": "^1.0.8",
+        "mime-types": "^2.1.12"
+      }
+    },
+    "abort-controller": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/abort-controller/-/abort-controller-3.0.0.tgz",
+      "integrity": "sha512-h8lQ8tacZYnR3vNQTgibj+tODHI5/+l06Au2Pcriv/Gmet0eaj4TwWH41sO9wnHDiQsEj19q0drzdWdeAHtweg==",
+      "requires": {
+        "event-target-shim": "^5.0.0"
+      }
+    },
+    "asynckit": {
+      "version": "0.4.0",
+      "resolved": "https://registry.npmjs.org/asynckit/-/asynckit-0.4.0.tgz",
+      "integrity": "sha1-x57Zf380y48robyXkLzDZkdLS3k="
+    },
+    "combined-stream": {
+      "version": "1.0.8",
+      "resolved": "https://registry.npmjs.org/combined-stream/-/combined-stream-1.0.8.tgz",
+      "integrity": "sha512-FQN4MRfuJeHf7cBbBMJFXhKSDq+2kAArBlmRBvcvFE5BB1HZKXtSFASDhdlz9zOYwxh8lDdnvmMOe/+5cdoEdg==",
+      "requires": {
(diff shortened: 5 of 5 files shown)


ImRodry:
Thanks for this! I see that you committed our package and package-lock files, is this intentional?

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser.rb
```diff
@@ -256,7 +257,7 @@ def requirement_for(requirement)
         return requirement unless git_url?(requirement)
 
         details = requirement.match(GIT_URL_REGEX).named_captures
-        details["semver"]
+        details["semver"] || (requirement.start_with?("github:") ? requirement : nil)
```
feelepxyz, line 260: Do we need to change this? Wondering if we should leave this as is, being `nil` if it doesn't match semver. Doesn't look like we actually rely on this change anywhere and might cause issues upstream issues when creating pull requests.

common/lib/dependabot/file_parsers/base/dependency_set.rb
```diff
@@ -21,6 +21,7 @@ def initialize(dependencies = [], case_sensitive: false)
         attr_reader :dependencies
 
         def <<(dep)
+          return if dep.nil?
```
feelepxyz, line 24: Is this change required? Less sure about changing the common classes unless we need to and make use of the change across each package manager to not introduce subtle differences in implementation.

npm_and_yarn/helpers/lib/npm6/updater.js
```diff
@@ -138,10 +138,10 @@ function installArgs(
...
       return `${depName}@${existingVersionRequirement}`;
     } else {
-      return `${depName}@${existingVersionRequirement.replace(
-        /#.*/,
-        ""
-      )}#${desiredVersion}`;
+      if (existingVersionRequirement.includes("#")) {
+        return `${depName}@${existingVersionRequirement.replace(/#.*/, "")}#${desiredVersion}`;
+      }
+      return `${depName}@${existingVersionRequirement}`;
```
feelepxyz, line 144: Nice detective work on finding this fix! 🙌

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser.rb
```diff
@@ -55,21 +55,22 @@ def manifest_dependencies
...
-              requirement = "*" if requirement == ""
-              dep = build_dependency(
-                file: file, type: type, name: name, requirement: requirement
+              dependency_set << build_dependency(
+                file: file,
+                type: type,
+                name: name,
+                requirement: requirement.empty? ? "*" : requirement
               )
-              dependency_set << dep if dep
```
feelepxyz, line 72: Related to the change to the dependency_set, thoughts on keeping this behaviour here so we don't need to change the common class?

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
```diff
@@ -151,11 +151,12 @@ def recursively_fetch_npm_lock_dependencies(object_with_dependencies)
 
           object_with_dependencies.
             fetch("dependencies", {}).each do |name, details|
-              next unless semver_version_for(details["version"])
+              version = semver_version_for(details["version"])
```
feelepxyz, line 154: 👍

npm_and_yarn/spec/fixtures/projects/npm6/ghpr_no_hash_pinning/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "hypixel-translators",
```
feelepxyz, line 2: This would be good to strip down to basic example, maybe just including `discord.js` dependency. An easy way of generating an example with npm is running `yes "" | npm init && npm install github:discordjs/discord.js#ab82cafcde0ee259a32ef14303c1b4a64dea8fae --save --lockfile-only`, if you then remove the `#sha` from `package.json` and re-run `npm install` it should then update the lockfile `from` to match keeping the old sha `version`

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser.rb
feelepxyz: I think this steer is a good idea but unsure of just changing it here for requirements that specifically match `github:`. Might be a good idea to figure out if we could always insert the requirement here but feels out of scope for this change.

xlgmokha:
> Thanks for this! I see that you committed our package and package-lock files, is this intentional?

No, this is temporary until I could wrap my head around the problem. I'll tidy up the fixture file before submitting for review.

common/lib/dependabot/file_parsers/base/dependency_set.rb
xlgmokha: > Is this change required?

Nope. I'll revert it.

npm_and_yarn/spec/fixtures/projects/npm6/ghpr_no_hash_pinning/package.json
ImRodry: You shouldn't need to install with the #sha to begin with though?

feelepxyz: > You shouldn't need to install with the #sha to begin with though?

I think we want to make sure the installed sha is old and can be updated

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser.rb
xlgmokha: > Do we need to change this?

Yes. I believe this this is the root cause of the issue. A `nil` constraint doesn't provide `npm` with enough information. More info here https://github.com/dependabot/dependabot-core/issues/3813#issuecomment-853458082.

I would prefer to change this line to:

```ruby
  details["semver"] || requirement
```

This would ensure that we always pass the original constraint down to npm to complete an install using the constraint provided in the manifest. However, I'm unsure of the other consequences of this change. Because of that I chose to be defensive and try to restrict this to only [GitHub repository style URLs](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#repository) for the time being. Without this change the defect is consistently reproduced.

npm_and_yarn/helpers/lib/npm6/updater.js
xlgmokha: Thanks. This was a fun side quest.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
xlgmokha: I noticed a few places where we could memoize work in the npm code. Since npm seems to be one of our slower ecosystems, it felt like a safe and beneficial change to make. Although, I recognize that this change isn't necessary for resolving this particular defect.

npm_and_yarn/spec/fixtures/projects/npm6/ghpr_no_hash_pinning/package.json
xlgmokha: I updated the fixture to make sure that the declared constraint isn't pinned to a sha and that the `from` field in the lockfile matches this constraint. Please let me know if I've missed anything else.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser.rb
feelepxyz: Ah not sure this gets used in the npm update, the requirement is extracted from the `package.json` file here https://github.com/dependabot/dependabot-core/blob/0b514b24ff8a201e379157a22a50107f7d86af27/npm_and_yarn/helpers/lib/npm6/updater.js#L52

Confusingly we also do this here https://github.com/dependabot/dependabot-core/blob/main/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb#L252-L257 🤔

I think this might need to broader refactors to clean up, I definitely think it would be a lot better to pass this from the file parser instead of re-extracting it in the lockfile updater and native helper.

xlgmokha: I am looking for more context around how the requirement is used in the downstream processing. Please do tell me what you think and if you have any alternative ideas that I can/should pursue.

feelepxyz: Does the dry-run update work if you revert this change? If so would be up for shipping this, I have a feeling this change doesn't have an impact on the update so might just cause issues in the pr creator where we use the requirement to build the PR title/body

feelepxyz: > Without this change the defect is consistently reproduced

I tested this dry-run without this change and seems to work but might be missing something else?

```
bin/dry-run.rb npm_and_yarn Hypixel-Translators/hypixel-translators-bot --dep=discord.js --cache=files --commit=e2bfc6768492d32f22bdc6a732237315b6b22dd2
```

npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater_spec.rb
```diff
@@ -500,4 +500,42 @@
...
+          type: "git",
+          url: "https://github.com/discordjs/discord.js",
+          branch: nil,
+          ref: "master"
+        }
+      }]
+    end
+
+    specify { expect(subject["dependencies"]["discord.js"]["version"]).to match(/github:discordjs\/discord.js#[0-9a-z]{40}/) }
+    specify { expect(subject["dependencies"]["discord.js"]["from"]).to eq("github:discordjs/discord.js") }
```
feelepxyz, line 539: Nit, looks like this would run the update twice to check each subject. Thoughts on using a single `it` here instead and check both attributes?

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser.rb
xlgmokha: > Ah not sure this gets used in the npm update, the requirement is extracted from the `package.json` file here
> 
> https://github.com/dependabot/dependabot-core/blob/0b514b24ff8a201e379157a22a50107f7d86af27/npm_and_yarn/helpers/lib/npm6/updater.js#L52
> 
> Confusingly we also do this here https://github.com/dependabot/dependabot-core/blob/main/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb#L252-L257 🤔
> 
> I think this might need to broader refactors to clean up, I definitely think it would be a lot better to pass this from the file parser instead of re-extracting it in the lockfile updater and native helper.

You're right. We don't need this. I just re-tested with and without and both appear to leave the `from` field in the desired state. I'll remove this change. Thank you for spotting this @feelepxyz. 🙇

feelepxyz (approved):
🙌 🚢

feelepxyz (approved):

npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater_spec.rb
```diff
@@ -316,7 +316,7 @@
         }]
       end
       let(:previous_version) { "d5ac0584ee9ae7bd9288220a39780f155b9ad4c8" }
-      let(:version) { "0c6b15a88bc10cd47f67a09506399dfc9ddc075d" }
+      let(:version) { "98e8ff1da1a89f93d1397a24d7413ed15421c139" }
```
feelepxyz, line 319: Oh why did this change?

feelepxyz: Ah this is the latest commit? hm not sure we should change this variable for all tests as we then loose code paths testing updating to specific commits vs latest when there's no reference. I think we still want some tests that try to update to a specific old commit, e.g. pointing to a tag.

xlgmokha: > Oh why did this change?

The fixtures in those tests do not pin to a hash (e.g. `"is-number": "jonschlinkert/is-number"`) so the npm install step fetches the latest hash for the default branch of that repo.

> Ah this is the latest commit?

Yep. 

> hm not sure we should change this variable for all tests as we then loose code paths testing updating to specific commits vs latest when there's no reference

🤔 Can I update the constraint in the package.json for those fixtures to pin to a fixed version?

xlgmokha:
@feelepxyz would you mind double checking the latest commit to this PR?

feelepxyz (approved):
sgtm 👍 seems ok to punt on copying the git dep over to dependabot-fixtures, could just do it when/if it ever flakes

PR #3871, opened 2021-06-09
Upgrade pip to 21.1.2 and pip-tools to 6.1.0

jurre:
https://github.com/jazzband/pip-tools/blob/master/CHANGELOG.md#610-2021-04-14

https://pip.pypa.io/en/stable/news/#v21-1-2

### DIFF
--- a/python/lib/dependabot/python/file_updater/pip_compile_file_updater.rb
+++ b/python/lib/dependabot/python/file_updater/pip_compile_file_updater.rb
@@ -410,7 +410,7 @@ module Dependabot
         def pip_compile_options_from_compiled_file(requirements_file)
           options = ["--output-file=#{requirements_file.name}"]
 
-          options << "--no-index" unless requirements_file.content.include?("index-url http")
+          options << "--no-emit-index-url" unless requirements_file.content.include?("index-url http")
 
           options << "--generate-hashes" if requirements_file.content.include?("--hash=sha")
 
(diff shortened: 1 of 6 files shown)


python/lib/dependabot/python/file_updater/pip_compile_file_updater.rb
```diff
@@ -410,7 +410,7 @@ def pip_compile_options(filename)
         def pip_compile_options_from_compiled_file(requirements_file)
           options = ["--output-file=#{requirements_file.name}"]
 
-          options << "--no-index" unless requirements_file.content.include?("index-url http")
+          options << "--no-emit-index-url" unless requirements_file.content.include?("index-url http")
```
jurre, line 413: This option was renamed

python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
```diff
@@ -141,18 +142,18 @@ def handle_pip_compile_errors(error)
...
                    named_captures.fetch("name")
             raise GitDependencyReferenceNotFound, name
           end
 
+          if error.message.match?(GIT_DEPENDENCY_UNREACHABLE_REGEX)
+            url = error.message.match(GIT_DEPENDENCY_UNREACHABLE_REGEX).
+                  named_captures.fetch("url")
+            raise GitDependenciesNotReachable, url
+          end
+
```
jurre, line 156: Had to swap these around, as the message for `GIT_DEPENDENCY_UNREACHABLE_REGEX` matches both

xlgmokha (approved):

PR #3873, opened 2021-06-09
v0.152.0

Nishnha:


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## v0.152.0, 9 June 2021
+
+- Upgrade and pin Terraform to version 1.0.0
+- Add lockfile support for Terraform
+- Pipenv upgrade
```
jurre, line 5: Would be good to mention the version here, and we also upgraded pip-tools and pip

jurre:
Opened https://github.com/dependabot/dependabot-core/pull/3879 instead since that's not on a fork, will make updating, tagging/releasing a little easier

Issue #3881, opened 2021-06-10
Terraform Unknown Error when searching for updates


ElvenSpellmaker:
**Package ecosystem**
Terraform

**Manifest location and content prior to update**

https://github.com/amido/stacks-infrastructure-aks/blob/master/.github/dependabot.yml

Pointing to this folder: https://github.com/amido/stacks-infrastructure-aks/tree/master/deploy/azure/infra/stacks-aks

We have a constraints file specifying the versions of things we need: https://github.com/amido/stacks-infrastructure-aks/blob/master/deploy/azure/infra/stacks-aks/constraints.tf

Error Output from the logs:
```
...
updater | INFO <job_156766197> Starting update job for amido/stacks-infrastructure-aks
updater | ERROR <job_156766197> undefined method `fetch' for "~> 2.5":String
updater | ERROR <job_156766197> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-terraform-0.152.0/lib/dependabot/terraform/file_parser.rb:90:in `build_provider_dependency'
updater | ERROR <job_156766197> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-terraform-0.152.0/lib/dependabot/terraform/file_parser.rb:41:in `block (4 levels) in parse'
updater | ERROR <job_156766197> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-terraform-0.152.0/lib/dependabot/terraform/file_parser.rb:40:in `each'
updater | ERROR <job_156766197> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-terraform-0.152.0/lib/dependabot/terraform/file_parser.rb:40:in `block (3 levels) in parse'
updater | ERROR <job_156766197> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-terraform-0.152.0/lib/dependabot/terraform/file_parser.rb:39:in `each'
updater | ERROR <job_156766197> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-terraform-0.152.0/lib/dependabot/terraform/file_parser.rb:39:in `block (2 levels) in parse'
updater | ERROR <job_156766197> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-terraform-0.152.0/lib/dependabot/terraform/file_parser.rb:37:in `each'
[...]

PR #3883, opened 2021-06-10
Terraform: Helpful error message for old provider syntax

Nishnha:
This PR adds a helpful error message to failing dependencies using the deprecated provider syntax.﻿
https://github.com/dependabot/dependabot-core/issues/3881


jurre (commented):
Would be great to add a test for this

terraform/lib/dependabot/terraform/file_parser.rb
```diff
@@ -109,6 +111,21 @@ def build_provider_dependency(file, name, details = {})
...
+          "This terraform provider syntax is now deprecated.\n"\
+          "See https://www.terraform.io/docs/language/providers/requirements.html "\
+          "for the new Terraform v0.13+ provider syntax."
+        )
+      end
+
+      def deprecated_provider?(details)
+        # The old syntax for terraform providers v0.12- looked like
+        # "tls ~> 2.1" which gets parsed as a string instead of a hash
+        !details.respond_to?(:fetch)
```
jurre, line 126: Should we check if this actually uses the old style syntax, or at least if it's a string instead of checking it responds to fetch?

terraform/spec/dependabot/terraform/file_parser_spec.rb
```diff
@@ -292,6 +292,20 @@
       end
     end
 
+    context "hcl1 provider syntax" do
+      let(:files) { project_dependency_files("hcl1_provider") }
```
jurre, line 296: I don't think this is hcl1 is it? Should we rename it?

jurre (approved):
I had one question about the fixture name, but other than that looks great

terraform/spec/dependabot/terraform/file_parser_spec.rb
Nishnha: Replaced hcl1 with deprecated. Now it reads as "deprecated provider syntax" and "deprecated_provider"

jurre (approved):
:shipit:

PR #3884, opened 2021-06-11
Tests: Allow profiling tests with stackprof when tagged

jurre:
This adds the ability to tag a spec with `:profile`:

```ruby
it "profiles with rspec", :profile do
  # ..
end
```

And will then generate a stackprof dump in the `tmp` directory.

This is useful when debugging performance issues.

### DIFF
--- a/common/spec/spec_helper.rb
+++ b/common/spec/spec_helper.rb
@@ -6,6 +6,7 @@ require "vcr"
 require "byebug"
 require "simplecov"
 require "simplecov-console"
+require "stackprof"
 
 require "dependabot/dependency_file"
 require_relative "dummy_package_manager/dummy"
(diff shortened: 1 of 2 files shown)


common/spec/spec_helper.rb
```diff
@@ -34,6 +35,18 @@
   config.order = :rand
   config.mock_with(:rspec) { |mocks| mocks.verify_partial_doubles = true }
   config.raise_errors_for_deprecations!
+
+  config.around do |example|
+    if example.metadata[:profile]
```
brrygrdn, line 40: 👍🏻

brrygrdn (approved):
Thanks of adding this 💯

feelepxyz (approved):
🙌

PR #3887, opened 2021-06-11
Update Ruby from 2.6 to 2.7

feelepxyz:
Updates Ruby from 2.6 to 2.7.

Ruby 2.7 ships a version of rubygems where the `Gem::Requirement` class is sorted. I don't think we rely on sort order in dependabot-core, we loop over requirements in turn to check if they match.

I had to change how we run the native helpers, to use `bundle exec` to make it pick up the `BUNDLER_VERSION` env var. Ruby 2.7 started falling back to bundler 2.1.4 for both bundler v1 and v2 without `bundle exec`.

## TODO
- [x] Test on staging

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,15 +64,19 @@ RUN if ! getent group $USER_GID; then groupadd --gid $USER_GID dependabot ; \
 
 ### RUBY
 
-# Install Ruby 2.6.6, update RubyGems, and install Bundler
+# Install Ruby 2.7, update RubyGems, and install Bundler
 ENV BUNDLE_SILENCE_ROOT_WARNING=1
+# Allow gem installs as the dependabot user
+ENV BUNDLE_PATH=".bundle" \
+    BUNDLE_BIN=".bundle/bin"
+ENV PATH="$BUNDLE_BIN:$PATH:$BUNDLE_PATH/bin"
 RUN apt-add-repository ppa:brightbox/ruby-ng \
   && apt-get update \
-  && apt-get install -y ruby2.6 ruby2.6-dev \
+  && apt-get install -y ruby2.7 ruby2.7-dev \
   && gem update --system 3.2.20 \
   && gem install bundler -v 1.17.3 --no-document \
   && gem install bundler -v 2.2.20 --no-document \
-  && rm -rf /var/lib/gems/2.6.0/cache/* \
+  && rm -rf /var/lib/gems/2.7.0/cache/* \
   && rm -rf /var/lib/apt/lists/*
 
 
@@ -253,10 +257,6 @@ RUN bash /opt/terraform/helpers/build /opt/terraform
 RUN bash /opt/composer/helpers/v1/build /opt/composer/v1
 RUN bash /opt/composer/helpers/v2/build /opt/composer/v2
 
-# Allow further gem installs as the dependabot user
 ENV HOME="/home/dependabot"
-ENV BUNDLE_PATH="$HOME/.bundle" \
-    BUNDLE_BIN=".bundle/bin"
-ENV PATH="$BUNDLE_BIN:$PATH:$BUNDLE_PATH/bin"
 
 WORKDIR ${HOME}
--- a/bundler/helpers/v2/lib/functions/lockfile_updater.rb
+++ b/bundler/helpers/v2/lib/functions/lockfile_updater.rb
@@ -9,6 +9,7 @@ module Functions
         not\sfind\s(?<name>[^\s]+)-\d|
         has\s(?<name>[^\s]+)\slocked\sat
       /x.freeze
+      DEPENDENCY_DROPPED = "_dependency_dropped_"
 
     def initialize(gemfile_name:, lockfile_name:, dependencies:)
       @gemfile_name = gemfile_name
@@ -36,7 +37,7 @@ module Functions
 
         old_reqs.each do |dep_name, old_req|
           d_dep = definition.dependencies.find { |d| d.name == dep_name }
-          if old_req == :none then definition.dependencies.delete(d_dep)
+          if old_req.to_s == DEPENDENCY_DROPPED then definition.dependencies.delete(d_dep)
           else
             d_dep.instance_variable_set(:@requirement, old_req)
           end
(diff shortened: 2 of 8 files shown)


bundler/helpers/v2/lib/functions/lockfile_updater.rb
```diff
@@ -36,7 +37,7 @@ def generate_lockfile
 
         old_reqs.each do |dep_name, old_req|
           d_dep = definition.dependencies.find { |d| d.name == dep_name }
-          if old_req == :none then definition.dependencies.delete(d_dep)
+          if old_req.to_s == DEPENDENCY_DROPPED then definition.dependencies.delete(d_dep)
```
feelepxyz, line 40: This started to blow up as the new gem requirement sorting tried to call `_requirements_sorted?` on `:none` https://github.com/rubygems/rubygems/blob/32ac0ba0cdad8ffe522865196d1a2ec3bbf62e08/bundler/lib/bundler/rubygems_ext.rb#L137

bundler/lib/dependabot/bundler/native_helpers.rb
```diff
@@ -17,7 +17,6 @@ def self.run_bundler_subprocess(function:, args:, bundler_version:)
               # Bundler will pick the matching installed major version
               "BUNDLER_VERSION" => bundler_version,
               "BUNDLE_GEMFILE" => File.join(versioned_helper_path(bundler_version: bundler_version), "Gemfile"),
-              "BUNDLE_PATH" => File.join(versioned_helper_path(bundler_version: bundler_version), ".bundle"),
```
feelepxyz, line 20: This causes issues when running `bundle exec`, pretty sure we added this before `GEM_HOME` which ended up being the real fix for gem caches not being writeable by the `dependabot` user the default location was outside of the home directory

bundler/lib/dependabot/bundler/native_helpers.rb
```diff
@@ -36,7 +35,7 @@ def self.versioned_helper_path(bundler_version:)
       end
 
       def self.helper_path(bundler_version:)
-        "ruby #{File.join(versioned_helper_path(bundler_version: bundler_version), 'run.rb')}"
+        "bundle exec ruby #{File.join(versioned_helper_path(bundler_version: bundler_version), 'run.rb')}"
```
feelepxyz, line 38: This is needed to force the `BUNDLER_VERSION` env to take effect for bundler v1 and force bundler to pick the lastest v2 `2.2.20` instead of always falling back to the pre-installed version in ruby 2.7 (`2.1.4`)

python/spec/dependabot/python/requirement_spec.rb
```diff
@@ -119,7 +119,7 @@
     context "with an array" do
       let(:requirement_string) { ["== 1.3.*", ">= 1.3.1"] }
       its(:to_s) do
-        is_expected.to eq(Gem::Requirement.new([">= 1.3.1", "~> 1.3.0.a"]).to_s)
+        is_expected.to eq(Gem::Requirement.new(["~> 1.3.0.a", ">= 1.3.1"]).to_s)
```
feelepxyz, line 122: Gem requirements are now sorted in ruby 2.7

Dockerfile
```diff
@@ -236,6 +237,11 @@ ENV DEPENDABOT_NATIVE_HELPERS_PATH="/opt" \
   PATH="$PATH:/opt/terraform/bin:/opt/python/bin:/opt/go_modules/bin:/opt/dep/bin" \
   MIX_HOME="/opt/hex/mix"
 
+# Allow further gem installs as the dependabot user
+ENV BUNDLE_PATH="/home/dependabot/.bundle" \
+    BUNDLE_BIN=".bundle/bin"
+ENV PATH="$BUNDLE_BIN:$PATH:$BUNDLE_PATH/bin"
```
feelepxyz, line 243: When fixing the typo in `bundler/helpers/v1/build` `bundle installs` started failing and this fixed it, I'm pretty sure we want to set these env vars before building the v1 v2 helpers. Not sure why this was working previously.

Dockerfile
```diff
@@ -63,15 +63,16 @@ RUN GROUP_NAME=$(getent group $USER_GID | awk -F':' '{print $1}') \
...
 
-# Install Ruby 2.6.6, update RubyGems, and install Bundler
+# Install Ruby 2.7, update RubyGems, and install Bundler
 ENV BUNDLE_SILENCE_ROOT_WARNING=1
 RUN apt-add-repository ppa:brightbox/ruby-ng \
   && apt-get update \
-  && apt-get install -y ruby2.6 ruby2.6-dev \
+  && apt-get install -y ruby2.7 ruby2.7-dev \
   && gem update --system 3.2.20 \
+  && gem uninstall bundler --version 2.1.4 \
```
jurre, line 72: Is this still needed now that we `bundle exec`?

bundler/lib/dependabot/bundler/native_helpers.rb
jurre: Seems like a good change to run this in the bundle environment regardless 👍

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
```diff
@@ -142,6 +142,11 @@
 
           let(:dependency_files) { bundler_project_dependency_files("bundler_specified") }
 
+          before do
+            stub_request(:get, rubygems_url + "versions/bundler.json").
+              to_return(status: 200, body: fixture("rubygems_responses", "versions-bundler.json"))
```
jurre, line 147: Does this mean we make an extra API call for every update now?

jurre (approved):
Few questions/suggestions, but none blocking

Dockerfile
feelepxyz: Probably not 👍

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
feelepxyz: Nah, this just seemed to pop up in this test which specifically updates `bundler`, but going to revert to see if this still fails with the `bundle exec` fix

jurre (approved):
Exciting!

Issue #946, opened 2019-01-22
Docker image is huge


LeoColomb:
[dependabot/dependabot-core](https://hub.docker.com/r/dependabot/dependabot-core/) image from [Dockerfile](https://github.com/dependabot/dependabot-core/blob/426fff73823cd27c6a65e1f220e936262cd07a96/Dockerfile) is amazing!

But it now exceeds [1 GB](https://hub.docker.com/r/dependabot/dependabot-core/tags), which is starting to be huge. ([why](https://developers.redhat.com/blog/2016/03/09/more-about-docker-images-size/) it [matters](https://semaphoreci.com/blog/2018/03/14/docker-image-size.html))

Some suggestions to optimize Dependabot's image (sorted by difficulty):
- [ ] Follow basic `apt` recommandations:
  * Use of `apt-get update` should be paired with `rm -rf /var/lib/apt/lists/*` in the same layer.
  * Use of `apt-get upgrade` must be avoided in a Dockerfile. ([ref](http://crosbymichael.com/dockerfile-best-practices-take-2.html)).
  * Consider using a `--no-install-recommends` when `apt-get` installing packages.
- [ ] Exclude development dependencies when distributing "production" image (#851).
  * Can be optimized easily with `build` scripts and an environment variable.
- [ ] Use [stages](https://docs.docker.com/develop/develop-images/multistage-build/) to build helpers and avoid excessive binaries distribution.
  * For example (if I'm correct) `golang` is not required to _run_ the helpers once built.
- [ ] Switch to a smaller base image for the last stage, commonly to Alpine.

LeoColomb:
Coming back to this issue (my system regularly complains this is taking too much disk space! 😱).

Actually I wonder if dividing the docker image wouldn't be even cleaner:
* Native helper isolation;
* Much (much) smaller image.

The only downside is that running all managers updaters will not be possible in the same container. (Is this a real downside? Are you used to run multiple manager-type updater in the same container?)

If interested, I can make PRs to iterate over this idea.

greysteil:
I know @hmarr has some plans for this - can you share?

hmarr:
@LeoColomb yes, we're planning exactly that! This is actually one of the main motivations for splitting up each language / package manager into independent top-level gems in dependabot-core. Once that big refactor is finished (we're super close), the next stage is to produce much leaner Docker images for each package manager, and use multi-stage builds to further trim the unnecessary fat.

[...]

PR #3896, opened 2021-06-14
Add missing apt lists clean up after in Dockerfile

PeterDaveHello:
Just a couple places missed the apt lists clean up thing 😄 

cc #946


jurre (approved):
Thank you!

PR #3897, opened 2021-06-14
Fix npm cache clean up in Dockerfile

PeterDaveHello:
The clean up of `~/.npm` should be in the same layer with `npm install`

cc #946 #3638


feelepxyz (approved):
Nice catch, looks like the previous `rm` was around when we supported elm 18 which was installed using `npm`

PR #3900, opened 2021-06-14
Add missing `-y` for apt-get install in Dockerfile

PeterDaveHello:
Should always use `-y` with `apt-get install` in Dockerfile to prevent manual input interrupt the image build. (This is the only on missing in the Dockerfile)


jurre (approved):
👍

jurre (approved):

PR #3904, opened 2021-06-15
Load bundler2 fixtures

thepwagner:
As a new contributor to the `bundler` ecosystem, I was surprised to discover that none of the included [bundler2 project fixtures](https://github.com/dependabot/dependabot-core/tree/main/bundler/spec/fixtures/projects/bundler2) are used.
* https://github.com/dependabot/dependabot-core/search?q=%22projects%2Fbundler1%22
* https://github.com/dependabot/dependabot-core/search?q=%22projects%2Fbundler2%22

This PR modifies the spec helper to load these fixtures, and plugs any gaps that have only been implemented in `bundler1` fixtures with copy/paste.
An alternative approach would be to remove the unused fixtures.


jurre (approved):
Thank you, the tests in `bundler/spec` do use those fixtures, but the native helpers seem to have been running against the old fixtures until now 😨

bundler/helpers/v2/spec/native_spec_helper.rb
```diff
@@ -26,8 +26,7 @@
 LOCKFILE_ENDING = /(?<ending>\s*(?:RUBY VERSION|BUNDLED WITH).*)/m.freeze
```
feelepxyz, line 26: Non blocking: We might be able to follow up and yank this out now

PR #3905, opened 2021-06-15
Add version_filter method to common, remove filter_vulnerable_version from ecosystems' update_checker

Nishnha:
Creates a new `version_filter` module and removes the `filter_vulnerable_version` method from ecosystems' `update_checker`.

### DIFF
--- a/bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
+++ b/bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
@@ -14,7 +14,6 @@ module Dependabot
   module Bundler
     class UpdateChecker
       class LatestVersionFinder
-        include Dependabot::UpdateCheckers::VersionFilters
 
         def initialize(dependency:, dependency_files:, repo_contents_path: nil,
                        credentials:, ignored_versions:, raise_on_ignored: false,
--- a/bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
+++ b/bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
@@ -14,7 +14,6 @@ module Dependabot
   module Bundler
     class UpdateChecker
       class LatestVersionFinder
-
         def initialize(dependency:, dependency_files:, repo_contents_path: nil,
                        credentials:, ignored_versions:, raise_on_ignored: false,
                        security_advisories:, options:)
--- a/bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
+++ b/bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
@@ -60,7 +60,8 @@ module Dependabot
           relevant_versions = filter_prerelease_versions(relevant_versions)
           relevant_versions = filter_ignored_versions(relevant_versions)
           relevant_versions = filter_lower_versions(relevant_versions)
-          relevant_versions = Dependabot::UpdateCheckers::VersionFilters.filter_vulnerable_versions(relevant_versions, security_advisories)
+          relevant_versions = Dependabot::UpdateCheckers::VersionFilters.filter_vulnerable_versions(relevant_versions,
+                                                                                                    security_advisories)
 
           relevant_versions.min
         end
--- a/bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
+++ b/bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
@@ -58,9 +58,9 @@ module Dependabot
 
           relevant_versions = dependency_source.versions
           relevant_versions = filter_prerelease_versions(relevant_versions)
-          relevant_versions = filter_vulnerable_versions(relevant_versions)
           relevant_versions = filter_ignored_versions(relevant_versions)
           relevant_versions = filter_lower_versions(relevant_versions)
+          relevant_versions = Dependabot::UpdateCheckers::VersionFilters.filter_vulnerable_versions(relevant_versions, security_advisories)
 
           relevant_versions.min
         end
--- a/bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
+++ b/bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
@@ -3,6 +3,7 @@
 require "excon"
 
 require "dependabot/bundler/update_checker"
+require "dependabot/update_checkers/version_filters"
 require "dependabot/bundler/requirement"
 require "dependabot/shared_helpers"
 require "dependabot/errors"
@@ -13,6 +14,8 @@ module Dependabot
   module Bundler
     class UpdateChecker
       class LatestVersionFinder
+        include Dependabot::UpdateCheckers::VersionFilters
+
         def initialize(dependency:, dependency_files:, repo_contents_path: nil,
                        credentials:, ignored_versions:, raise_on_ignored: false,
                        security_advisories:, options:)
@@ -76,11 +79,6 @@ module Dependabot
           end
 
           filtered
-        end
-
-        def filter_vulnerable_versions(versions_array)
-          versions_array.
-            reject { |v| security_advisories.any? { |a| a.vulnerable?(v) } }
         end
 
         def filter_lower_versions(versions_array)
--- a/cargo/lib/dependabot/cargo/update_checker/latest_version_finder.rb
+++ b/cargo/lib/dependabot/cargo/update_checker/latest_version_finder.rb
@@ -2,11 +2,14 @@
 
 require "excon"
 require "dependabot/cargo/update_checker"
+require "dependabot/update_checkers/version_filters"
 
 module Dependabot
   module Cargo
     class UpdateChecker
       class LatestVersionFinder
+        include Dependabot::UpdateCheckers::VersionFilters
+
         def initialize(dependency:, dependency_files:, credentials:,
                        ignored_versions:, raise_on_ignored: false,
                        security_advisories:)
@@ -61,11 +64,6 @@ module Dependabot
           end
 
           filtered
-        end
-
-        def filter_vulnerable_versions(versions_array)
-          versions_array.
-            reject { |v| security_advisories.any? { |a| a.vulnerable?(v) } }
         end
 
         def filter_lower_versions(versions_array)
--- a/cargo/lib/dependabot/cargo/update_checker/latest_version_finder.rb
+++ b/cargo/lib/dependabot/cargo/update_checker/latest_version_finder.rb
@@ -8,7 +8,6 @@ module Dependabot
   module Cargo
     class UpdateChecker
       class LatestVersionFinder
-
         def initialize(dependency:, dependency_files:, credentials:,
                        ignored_versions:, raise_on_ignored: false,
                        security_advisories:)
--- a/cargo/lib/dependabot/cargo/update_checker/latest_version_finder.rb
+++ b/cargo/lib/dependabot/cargo/update_checker/latest_version_finder.rb
@@ -44,9 +44,10 @@ module Dependabot
         def fetch_lowest_security_fix_version
           versions = available_versions
           versions = filter_prerelease_versions(versions)
-          versions = filter_vulnerable_versions(versions)
           versions = filter_ignored_versions(versions)
           versions = filter_lower_versions(versions)
+          versions = Dependabot::UpdateCheckers::VersionFilters.filter_vulnerable_versions(versions, security_advisories)
+
           versions.min
         end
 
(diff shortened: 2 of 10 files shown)


common/lib/dependabot/update_checkers/version_filters.rb
```diff
@@ -0,0 +1,17 @@
+module Dependabot
+  module UpdateCheckers
+    module VersionFilters
+      def filter_vulnerable_versions(versions_array)
+        versions_array.reject do |v|
+          security_advisories.any? do |a|
+            if a.is_a?(Gem::Version)
```
feelepxyz, line 7: should this check `v`?
```suggestion
            if v.is_a?(Gem::Version)
```

common/lib/dependabot/update_checkers/version_filters.rb
```diff
@@ -0,0 +1,17 @@
+module Dependabot
+  module UpdateCheckers
+    module VersionFilters
+      def filter_vulnerable_versions(versions_array)
```
feelepxyz, line 4: This might be a personal style but I always feel nervous about mixins as a pattern. I find they often cause subtle bugs and can get hard to debug. Thoughts on exposing a class method and calling that directly instead?

e.g. `Dependabot::UpdateCheckers::VersionFilters.filter_vulnerable_versions(relevant_versions)`

common/lib/dependabot/update_checkers/version_filters.rb
```diff
@@ -0,0 +1,17 @@
...
+        versions_array.reject do |v|
+          security_advisories.any? do |a|
+            if v.is_a?(Gem::Version)
```
feelepxyz, line 7: I wonder if we should transform the argument passed in here to always be a Gem::Version in the package managers that don't expose this?

jurre: The tricky thing is we'd then also _return_ a `Gem::Version`, but in some cases we're filtering a set of "version details" hashes. We _could_ then filter those using the array `Gem::Version` returned, but that feels a little off?

common/lib/dependabot/update_checkers/version_filters.rb
jurre: Yeah, it'd require us to pass in the `security_advisories` as well, but that'd work 👍

common/lib/dependabot/update_checkers/version_filters.rb
feelepxyz: Aye, seems find as is 👍

common/lib/dependabot/update_checkers/version_filters.rb
feelepxyz: Hm good point, at which point the method is pretty basic and the signature is almost the same number of characters 😀 

Mixing in all these methods also feels hairy, `security_advisories` is currently exposed in the base update checker, and the dependency is very implicit 🤔

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/latest_version_finder.rb
```diff
@@ -12,6 +13,8 @@ module NpmAndYarn
     class UpdateChecker
       class LatestVersionFinder
         class RegistryError < StandardError
+          include Dependabot::UpdateCheckers::VersionFilters
+
```
feelepxyz, line 17:
```suggestion
      class LatestVersionFinder
        include Dependabot::UpdateCheckers::VersionFilters

        class RegistryError < StandardError
```

common/lib/dependabot/update_checkers/version_filters.rb
jurre: Yeah I think it'd be an improvement 👍

feelepxyz: I definitely prefer the class method/functional pattern where we just pass in the args we need over mixins but realise it's not a common pattern in ruby. It's a shame we need to pass all this global state around.

I'm not strongly adverse to getting this in as is. This might be a good refactor candidate in future if we figure out a `context` pattern ala react where we have a immutable store that includes context like update config options, ignore conditions and sec advisories.

Nishnha: I think for the sake of a [dependent PR](https://github.com/dependabot/dependabot-core/pull/3910) we should keep this as is and consider options like a global `context` for down the road.

jurre: @Nishnha agreed for sure on punting on the `context` part, but we could still implement the suggested stateless method:

```ruby
Dependabot::UpdateCheckers::VersionFilters.filter_vulnerable_versions(relevant_versions, security_advisories)
```

The downside of using a mixin like this, is that it relies on whatever class we include it in to define the `security_advisories` state. This can lead to bugs when a class doesn't, or implements it slightly differently, and it's not a patter we use a ton in the rest of the codebase.

In order to make the suggested change, we would change the method definition to:

```ruby
def self.filter_vulnerable_versions(versions_array, security_advisories)
```

Remote all the `include Dependabot::UpdateCheckers::VersionFilters`, and change the method calls to call the method directly on the module:

```diff
- filter_vulnerable_versions(relevant_versions)
+ Dependabot::UpdateCheckers::VersionFilters.filter_vulnerable_versions(relevant_versions, security_advisories)
```

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/latest_version_finder.rb
```diff
@@ -11,6 +12,8 @@ module Dependabot
   module NpmAndYarn
     class UpdateChecker
       class LatestVersionFinder
+        include Dependabot::UpdateCheckers::VersionFilters
```
jurre, line 15:
```suggestion
```

nuget/lib/dependabot/nuget/update_checker/version_finder.rb
```diff
@@ -14,6 +15,8 @@ class UpdateChecker
       class VersionFinder
         require_relative "repository_finder"
 
+        include Dependabot::UpdateCheckers::VersionFilters
```
jurre, line 18:
```suggestion
```

python/lib/dependabot/python/update_checker/latest_version_finder.rb
```diff
@@ -14,6 +15,8 @@ module Dependabot
   module Python
     class UpdateChecker
       class LatestVersionFinder
+        include Dependabot::UpdateCheckers::VersionFilters
```
jurre, line 18:
```suggestion
```

bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
```diff
@@ -13,6 +14,8 @@ module Dependabot
   module Bundler
     class UpdateChecker
       class LatestVersionFinder
+        include Dependabot::UpdateCheckers::VersionFilters
```
jurre, line 17:
```suggestion
```

cargo/lib/dependabot/cargo/update_checker/latest_version_finder.rb
```diff
@@ -2,11 +2,14 @@
 
 require "excon"
 require "dependabot/cargo/update_checker"
+require "dependabot/update_checkers/version_filters"
 
 module Dependabot
   module Cargo
     class UpdateChecker
       class LatestVersionFinder
+        include Dependabot::UpdateCheckers::VersionFilters
```
jurre, line 11:
```suggestion
```

jurre (changes requested):
I made some suggestions to remove no-op includes, other than that I think this looks good

bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
```diff
@@ -55,9 +57,10 @@ def fetch_lowest_security_fix_version
 
           relevant_versions = dependency_source.versions
           relevant_versions = filter_prerelease_versions(relevant_versions)
-          relevant_versions = filter_vulnerable_versions(relevant_versions)
           relevant_versions = filter_ignored_versions(relevant_versions)
           relevant_versions = filter_lower_versions(relevant_versions)
+          relevant_versions = Dependabot::UpdateCheckers::VersionFilters.filter_vulnerable_versions(relevant_versions,
+                                                                                                    security_advisories)
```
feelepxyz, line 63: You'll want to filtering vulnerable versions before filtering ignored versions as this code expects the set to be empty if all versions have been filtered out.

```suggestion
          relevant_versions = Dependabot::UpdateCheckers::VersionFilters.filter_vulnerable_versions(relevant_versions,
                                                                                                    security_advisories)
          relevant_versions = filter_ignored_versions(relevant_versions)
          relevant_versions = filter_lower_versions(relevant_versions)
```

feelepxyz: This is probably related to the npm spec failure too

feelepxyz: Ref: https://github.com/dependabot/dependabot-core/blob/4de04ae1967590b34f611192f4c788d6f834f7f3/bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb#L77

Nishnha: Nice spot! 
@xlgmokha and I considered this fix but we preferred a fix that wasn't order-dependent. We pulled out the `Dependabot::AllVersionsIgnored` error from `#filter_ignored_versions` [into its own method](https://github.com/dependabot/dependabot-core/pull/3905/commits/1606f2c5d97551ac6e57cb5cee6d3acd4df6c998#diff-fa6fc731ba81068ee1096165004258eb11a1ff282b214ff79622dbd9d18f0ef2R89). 

We called the new error raising method in [#latest_version_from_registry](https://github.com/dependabot/dependabot-core/pull/3905/commits/1606f2c5d97551ac6e57cb5cee6d3acd4df6c998#diff-fa6fc731ba81068ee1096165004258eb11a1ff282b214ff79622dbd9d18f0ef2R45) and [#lowest_security_fix_version](https://github.com/dependabot/dependabot-core/pull/3905/commits/1606f2c5d97551ac6e57cb5cee6d3acd4df6c998#diff-fa6fc731ba81068ee1096165004258eb11a1ff282b214ff79622dbd9d18f0ef2R81)

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/latest_version_finder.rb
```diff
@@ -58,22 +67,29 @@ def latest_version_with_no_unlock
...
             else possible_versions(filter_ignored: false)
             end
 
-          secure_versions = filter_vulnerable_versions(versions_array)
           secure_versions = filter_ignored_versions(secure_versions)
           secure_versions = filter_lower_versions(secure_versions)
+          secure_versions = Dependabot::UpdateCheckers::VersionFilters.filter_vulnerable_versions(secure_versions,
+                                                                                                  security_advisories)
+
+          ensure_secure_version_available!(secure_versions)
```
feelepxyz, line 81: I don't think we can move this method out of `filter_ignored_versions` as the condition has now changed to raise when there are no versions left regardless of where they where filtered out.

We only want to raise `AllVersionsIgnored` when there are some versions available and the ignore conditions have filtered them all out.

I think the previous implementation should work as expected if you moved the vulnerable version filter to be before the ignored versions filter:

```
          secure_versions = Dependabot::UpdateCheckers::VersionFilters.filter_vulnerable_versions(secure_versions,
                                                                                                  security_advisories)
          secure_versions = filter_ignored_versions(secure_versions)
          secure_versions = filter_lower_versions(secure_versions)
```

Nishnha: Reordered the filers like you suggested @feelepxyz which seems to do the job for this pull request 💯 

Opening a new PR to consider refactoring the filters to be order-independent. That would be outside the scope of this PR since it makes significant changes across ecosystems and we may not have test coverage for the order dependence of the filters.

bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
```diff
@@ -55,9 +56,10 @@ def fetch_lowest_security_fix_version
...
```
feelepxyz, line 62: @Nishnha quick flag, I think we should revert the order of these filters across all ecosystems to be `filter_vulnerable_versions` then `filter_ignored_versions`. I think we've got spotty test coverage covering this order so might have just failed in a couple.

Nishnha: Since this PR is already merged in I opened a new one to address the issue 
https://github.com/dependabot/dependabot-core/pull/3962

feelepxyz: Thanks appreciate the quick fix!

PR #3906, opened 2021-06-15
enable rubocop on bundler/helpers

thepwagner:
Dependabot's logic for the bundler ecosystem is split between "normal" dependabot-core, and helpers specific to bundler1 and bundler2.

I'm not sure why the ruby code in the helpers was being excluded from rubocop, so this PR enables it and fixes all outstanding issues that have accumulated. This maintains styling across dependabot-core's ruby codebase.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,9 +4,7 @@ AllCops:
   Exclude:
     - '*/vendor/**/*'
     - '*/bin/**/*'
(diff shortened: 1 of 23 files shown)


feelepxyz (approved):
Nice catch! Not sure why these where excluded, it might have been an old hangup pre bundler v1/v2 days

PR #3910, opened 2021-06-15
Security advisories for Go

Nishnha:
Currently tested against manual security advisories. 
Uses [github.com/dependabot-fixtures/go-modules-lib][test-fixture] as a test fixture and passes a dry-run.

Based off of https://github.com/dependabot/dependabot-core/pull/3905

TODO:
- [x] Add the lowest_security_fix_version and lowest_resolvable_security_fix_version methods to the `update checker` and the `version finder`
- [x] Add more releases to the [test fixture][test-fixture]
- [x] `LatestVersionFinderSpec#lowest_security_fix_version` tests
    - [x] supports/ignores pre-releases
    - [x] doesn't downgrade
    - [x] fixed versions
    - [x] not latest version
    - [x] not ignored
    - [x] Removed: not retracted release
- [x] `UpdateChecker#lowest_security_fix_version` tests

[test-fixture]: https://github.com/dependabot-fixtures/go-modules-lib

### DIFF
--- a/go_modules/lib/dependabot/go_modules/update_checker.rb
+++ b/go_modules/lib/dependabot/go_modules/update_checker.rb
@@ -42,7 +42,6 @@ module Dependabot
       end
 
       def lowest_resolvable_security_fix_version
-        raise "Dependency not vulnerable!" unless vulnerable?
 
         return latest_resolvable_version if git_dependency?
 
--- a/go_modules/lib/dependabot/go_modules/update_checker.rb
+++ b/go_modules/lib/dependabot/go_modules/update_checker.rb
@@ -24,14 +24,7 @@ module Dependabot
           return version_class.new(dependency.version)
         end
 
-        @latest_resolvable_version ||=
-          LatestVersionFinder.new(
-            dependency: dependency,
-            dependency_files: dependency_files,
-            credentials: credentials,
-            ignored_versions: ignored_versions,
-            raise_on_ignored: raise_on_ignored
-          ).latest_version
+        latest_version_finder.latest_version
       end
 
       # This is currently used to short-circuit latest_resolvable_version,
@@ -42,12 +35,19 @@ module Dependabot
       end
 
       def lowest_resolvable_security_fix_version
+        raise "Dependency not vulnerable!" unless vulnerable?
 
-        return latest_resolvable_version if git_dependency?
+        unless dependency.top_level?
+          return unless dependency.version
 
-        return unless lowest_security_fix_version
+          return version_class.new(dependency.version)
+        end
 
-        resolvable?(lowest_fix) ? lowest_fix : latest_resolvable_version
+        lowest_security_fix_version
+      end
+
+      def lowest_security_fix_version
+        latest_version_finder.lowest_security_fix_version
       end
 
       def latest_resolvable_version_with_no_unlock
@@ -62,6 +62,18 @@ module Dependabot
       end
 
       private
+
+      def latest_version_finder
+        @latest_version_finder ||=
+          LatestVersionFinder.new(
+            dependency: dependency,
+            dependency_files: dependency_files,
+            credentials: credentials,
+            ignored_versions: ignored_versions,
+            security_advisories: security_advisories,
+            raise_on_ignored: raise_on_ignored
+          )
+      end
 
       def latest_version_resolvable_with_full_unlock?
         # Full unlock checks aren't implemented for Go (yet)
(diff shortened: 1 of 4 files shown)


go_modules/lib/dependabot/go_modules/update_checker.rb
```diff
@@ -41,6 +41,16 @@ def latest_version
         latest_resolvable_version
       end
 
+      def lowest_resolvable_security_fix_version
```
jurre, line 44: We should add some test coverage for this method as well, is `resolvable?` available in the go update checker? I thought it wasn't

Nishnha: You're correct, looks like I jumped the gun on this PR.
`lowest_security_fix_version` is also not available to UpdateChecker.

go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb
```diff
@@ -50,6 +55,19 @@ def fetch_latest_version
...
 
+        def fetch_lowest_security_fix_version
+          return dependency.version if dependency.version =~ PSEUDO_VERSION_REGEX
+
+          relevant_versions = available_versions
+          relevant_versions = filter_prerelease_versions(relevant_versions)
+          relevant_versions = filter_ignored_versions(relevant_versions)
+          relevant_versions = filter_lower_versions(relevant_versions)
+          relevant_versions = Dependabot::UpdateCheckers::VersionFilters.filter_vulnerable_versions(relevant_versions,
+                                                                                                    security_advisories)
```
feelepxyz, line 66:
```suggestion
          relevant_versions = Dependabot::UpdateCheckers::VersionFilters.filter_vulnerable_versions(relevant_versions,
                                                                                                    security_advisories)
          relevant_versions = filter_ignored_versions(relevant_versions)
          relevant_versions = filter_lower_versions(relevant_versions)
```

go_modules/spec/dependabot/go_modules/update_checker/latest_version_finder_spec.rb
```diff
@@ -238,4 +241,56 @@ module foobar
...
+      it "returns newest pre-release" do
+        expect(finder.lowest_security_fix_version).to eq(Dependabot::GoModules::Version.new("1.2.0-pre2"))
+      end
+    end
+
+    context "when on a stable release and a newer prerelease is available" do
+      let(:current_version) { "1.1.0" }
+
+      it "doesn't return pre-release" do
+        expect(finder.lowest_security_fix_version).to_not eq(Dependabot::GoModules::Version.new("1.2.0-pre2"))
```
feelepxyz, line 292: Could this test the positive case instead and expect it to `eq` a version? Can be hard to catch regressions when only checking the negative case, as a new pre-release version could still make this pass

go_modules/spec/dependabot/go_modules/update_checker_spec.rb
```diff
@@ -75,4 +79,58 @@ module foobar
...
+      it "updates to the least new supported version" do
+        is_expected.to eq(Dependabot::GoModules::Version.new("1.0.5"))
+      end
+    end
+  end
+
+  describe "#lowest_resolvable_security_fix_version" do
+    subject(:lowest_resolvable_security_fix_version) { checker.lowest_resolvable_security_fix_version }
+
+    let(:dependency_version) { "1.0.1" }
```
feelepxyz, line 108: Is there a test case that could be added to cover the `unless dependency.top_level?` code-path?

feelepxyz: Looks like this will be an `indirect` dependency?

feelepxyz (commented):
Nice! Left a few minor comments otherwise looks ace to me 💯 

Had a quick play with `SECURITY_ADVISORIES='[{"dependency-name":"github.com/miekg/dns","patched-versions":[],"unaffected-versions":[],"affected-versions":["<= 1.1.35"]}]' bin/dry-run.rb go_modules v2ray/v2ray-core --cache=files --dep=github.com/miekg/dns` 🙌

go_modules/spec/dependabot/go_modules/update_checker_spec.rb
feelepxyz:
```
context "with an indirect dependency"
  let(:requirements) { [] }

  it "updates to the least new supported version" do
      is_expected.to eq(Dependabot::GoModules::Version.new(dependency_version))
  end
end
```

PR #3913, opened 2021-06-16
Raise on ruby 2.7 deprecation warnings

feelepxyz:
Break tests if we introduce any ruby 2.7 deprecation warnings.

### DIFF
--- a/common/spec/spec_helper.rb
+++ b/common/spec/spec_helper.rb
@@ -10,6 +10,7 @@ require "stackprof"
 
 require "dependabot/dependency_file"
 require_relative "dummy_package_manager/dummy"
+require_relative "warning_monkey_patch"
 
 if ENV["COVERAGE"]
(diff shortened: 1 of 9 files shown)


jurre (approved):
Nice! (I guess we're expecting this to fail for a bit and still need to clean a few things up?)

We might need to allow _some_ warnings, not all will be ruby 2.7 related?

common/spec/spec_helper.rb
```diff
@@ -11,19 +11,22 @@
 require "dependabot/dependency_file"
 require_relative "dummy_package_manager/dummy"
 
+ALLOW_PATTERNS = [
+  # Ignore parser warnings for ruby 2.7 minor version mismatches
+  %r{parser/current is loading parser/ruby27}
```
jurre, line 16: 👍

PR #3921, opened 2021-06-16
Github actions: Handle no latest version found

feelepxyz:
When no latest version can be found return nil to prevent the update
checker from comparing the version with the current version:

```
ArgumentError: comparison of String with Dependabot::GithubActions::Version failed
```

We previously returned the current `dependency.version` which is a string. This change means `can_update?` returns false when there is no latest version found.


github_actions/spec/fixtures/vcr_cassettes/Dependabot_GithubActions_UpdateChecker/_can_update_/given_a_dependency_with_a_tag_reference/with_a_dependency_that_has_a_latest_requirement_and_a_valid_version/1_2_2_6_1.yml
```diff
@@ -0,0 +1,91 @@
...
+        refs/pull/86/head\n003f2c8f71882f7df483f8c60077e839dcd77c4bb110 refs/pull/87/head\n003f22098c348e61d788892de2c5a2bd3dc3046cfd35
+        refs/pull/88/head\n003f52f35f70f8db19ff5850f8a454bf87acb2af92d6 refs/pull/89/head\n0040eb0228e41914eb033762292e266e6ca5ffe5ab75
+        refs/pull/89/merge\n003f70321bad373c97665e2d7e463a4373ea17586c22 refs/pull/90/head\n003faa8d5ff2a35e819db0332788c64c03f00abf73ca
+        refs/pull/93/head\n0040ed3a08a68b8d83e5f69b4fe612dc1d60129de4e0 refs/pull/93/merge\n003f5a98c8fe6feae865a4da24d3042e5813c6c0688a
+        refs/pull/94/head\n00406a10cbc405cf4a95a328e41b382379c015605a6f refs/pull/94/merge\n003ff60d8eac1114608aba7746762ea1575ab4313dbd
+        refs/pull/95/head\n004048a8b21925e962d8a7738d323d915b0cecfe1f2b refs/pull/95/merge\n003f8868c043d3b524cc5bda7bfb5f1c205aee14b8b3
+        refs/pull/96/head\n003f7132fc2fa62693d51d29209dfac800bb3a92c358 refs/pull/97/head\n0040b581cd874bd08b678bc8cac29e2538bdb766e22b
+        refs/pull/97/merge\n003fc1929ead8889aba340647d4ae8a388c61b198424 refs/pull/98/head\n003e0cb9c9b65d5d1901c1f53e5e66eaf4afd303e70e
```
xlgmokha, line 83: To make it easier to spot the defect would it be helpful to delete the refs listed above. I think the only relevant refs are the ones listed below because the test indicates a ref of `@v1`. 

I'm also a bit confused by the data below, it looks like there is a `refs/tags/v1\n` with a sha1 (003e8d93430eddafb926c668181c71f579556f68668c) that matches the `refs/tags/latest` sha1.

github_actions/spec/fixtures/vcr_cassettes/Dependabot_GithubActions_UpdateChecker/_latest_version/with_a_dependency_that_has_a_latest_requirement_and_a_valid_version/1_3_5_1.yml
```diff
@@ -0,0 +1,91 @@
...
+    method: get
+    uri: https://github.com/actions/create-release.git/info/refs?service=git-upload-pack
+    body:
+      encoding: US-ASCII
+      string: ''
+    headers:
+      User-Agent:
+      - dependabot-core/0.154.1 excon/0.82.0 ruby/2.7.3 (x86_64-darwin19) (+https://github.com/dependabot/dependabot-core)
+      Authorization:
+      - Basic eC1hY2Nlc3MtdG9rZW46Z2hwX1ZveHh3TVZVWDBCMXZQdTdzd2N5eDEwVWxsbndwQjJhTmVzTA==
```
xlgmokha, line 13:
```bash
echo eC1hY2Nlc3MtdG9rZW46Z2hwX1ZveHh3TVZVWDBCMXZQdTdzd2N5eDEwVWxsbndwQjJhTmVzTA== | base64 -D
```

I think we should remove any non-essential headers. The request headers aren't really necessary and the only response header that we need is probably `Content-Type`.

github_actions/spec/fixtures/vcr_cassettes/Dependabot_GithubActions_UpdateChecker/_can_update_/given_a_dependency_with_a_tag_reference/with_a_dependency_that_has_a_latest_requirement_and_a_valid_version/1_2_2_6_1.yml
```diff
@@ -0,0 +1,91 @@
...
+    method: get
+    uri: https://github.com/actions/create-release.git/info/refs?service=git-upload-pack
+    body:
+      encoding: US-ASCII
+      string: ''
+    headers:
+      User-Agent:
+      - dependabot-core/0.154.1 excon/0.82.0 ruby/2.7.3 (x86_64-darwin19) (+https://github.com/dependabot/dependabot-core)
+      Authorization:
+      - Basic eC1hY2Nlc3MtdG9rZW46Z2hwX1ZveHh3TVZVWDBCMXZQdTdzd2N5eDEwVWxsbndwQjJhTmVzTA==
```
xlgmokha, line 13: See my related comment below.

github_actions/spec/fixtures/vcr_cassettes/Dependabot_GithubActions_UpdateChecker/_can_update_/given_a_dependency_with_a_tag_reference/with_a_dependency_that_has_a_latest_requirement_and_a_valid_version/1_2_2_6_1.yml
feelepxyz: Hm I'm not sure we want to start editing vcr records as they'll all be regenerated from actual requests by running `VCR=true rspec spec`.

I think `v1` == `latest` is a common pattern, means you can just do `uses: actions/create-release@v1` to get the latest v1 release, so `v1 == v1.1.4 == latest` are all pointing to the same sha.

common/spec/spec_helper.rb
```diff
@@ -59,6 +60,14 @@
     config.debug_logger = File.open(ENV["DEPENDABOT_TEST_DEBUG_LOGGER"], "w")
   end
 
+  # Prevent auth headers and username:password params being written to VCR cassets
+  config.before_record do |interaction|
+    interaction.response.headers.transform_keys!(&:downcase).delete("set-cookie")
+    interaction.request.headers.transform_keys!(&:downcase).delete("authorization")
+    uri = URI.parse(interaction.request.uri)
+    interaction.request.uri.sub!(%r{:\/\/.*#{Regexp.escape(uri.host)}}, "://#{uri.host}")
+  end
```
feelepxyz, line 69: Snuck this in part of this PR as a separate commit as the VCR cassette included an authorization header with a PAT.

jurre: Nice one, did you check to make sure there are no other PATs in the VCR records?

Do we _need_ these requests to be authenticated? I think we should still keep this in, but we can maybe also record them without being auth'ed?

jurre (approved):

PR #3927, opened 2021-06-17
Terraform: Handle 401 registry responses

feelepxyz:
Raise a `PrivateSourceAuthenticationFailure` with the provided hostname
if the registry responds with a `401`.


jurre (approved):
Thanks!

PR #3930, opened 2021-06-17
Terraform: handle dependencies without a namespace

feelepxyz:
Default providers without a namespace to `hashicorp`.

### DIFF
--- a/terraform/lib/dependabot/terraform/file_parser.rb
+++ b/terraform/lib/dependabot/terraform/file_parser.rb
@@ -170,13 +170,11 @@ module Dependabot
       end
 
       def provider_source_from(source_address, name)
-        return [DEFAULT_REGISTRY, DEFAULT_NAMESPACE, name] unless source_address
-
-        matches = source_address.match(PROVIDER_SOURCE_ADDRESS)
+        matches = source_address&.match(PROVIDER_SOURCE_ADDRESS)
         [
-          matches[:hostname] || DEFAULT_REGISTRY,
-          matches[:namespace],
-          matches[:name] || name
+          matches.try(:[], :hostname) || DEFAULT_REGISTRY,
+          matches.try(:[], :namespace) || DEFAULT_NAMESPACE,
+          matches.try(:[], :name) || name
         ]
       end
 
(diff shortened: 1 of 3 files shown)


feelepxyz:
@xlgmokha tagged you for more input on this change. I think we should ignore request failures here but also wondering if I'm missing something about how terraform providers work? 

Looks like this might be the `random` provider https://registry.terraform.io/providers/hashicorp/random/latest/docs

terraform/lib/dependabot/terraform/file_parser.rb
```diff
@@ -92,7 +92,7 @@ def build_provider_dependency(file, name, details = {})
         source_address = details.fetch("source", nil)
         version_req = details["version"]&.strip
         hostname, namespace, name = provider_source_from(source_address, name)
-        dependency_name = source_address ? "#{namespace}/#{name}" : name
+        dependency_name = source_address && namespace ? "#{namespace}/#{name}" : name
```
xlgmokha, line 95:
```suggestion
        dependency_name = [namespace, name].compact.join('/')
```

xlgmokha (approved):
LGTM

feelepxyz:
@xlgmokha ops I just realised what was going on, just needed to default the namespace to `hashicorp`. I think it might still be worth handling request errors but will push that in a separate PR.

terraform/lib/dependabot/terraform/file_parser.rb
```diff
@@ -170,13 +170,11 @@ def source_from(details_hash)
...
-        return [DEFAULT_REGISTRY, DEFAULT_NAMESPACE, name] unless source_address
-
-        matches = source_address.match(PROVIDER_SOURCE_ADDRESS)
+        matches = source_address&.match(PROVIDER_SOURCE_ADDRESS)
         [
-          matches[:hostname] || DEFAULT_REGISTRY,
-          matches[:namespace],
-          matches[:name] || name
+          matches.try(:[], :hostname) || DEFAULT_REGISTRY,
+          matches.try(:[], :namespace) || DEFAULT_NAMESPACE,
```
xlgmokha, line 176: TIL: you can omit `hashicorp` as that becomes the default namespace.

xlgmokha (approved):
:shipit:

PR #3931, opened 2021-06-17
Returns basename and relative path in CodeCommit file fetcher

lorengordon:
Previously, the `name` and `path` attributes were both the absolute
path for all files retrieved from CodeCommit sources. This caused
problems in `fetch_file_from_host`, which joins the `directory`
from the config with the `filename`. When `filename` is the absolute
path, this essentially duplicated the directory in the result, e.g.
`directory/directory/relative path`.

This change returns the basename and relative path instead, matching
the implemenation for azure devops, and allowing the join in `fetch_file_from_host`
to return `directory/relative path.

### DIFF
--- a/common/lib/dependabot/file_fetchers/base.rb
+++ b/common/lib/dependabot/file_fetchers/base.rb
@@ -330,8 +330,8 @@ module Dependabot
 
         response.files.map do |file|
           OpenStruct.new(
-            name: file.absolute_path,
-            path: file.absolute_path,
+            name: File.basename(file.relative_path),
+            path: file.relative_path,
             type: "file",
             size: 0 # file size would require new api call per file..
           )


common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -136,7 +136,7 @@ def load_cloned_file_if_present(filename)
       def fetch_file_from_host(filename, type: "file", fetch_submodules: false)
         return load_cloned_file_if_present(filename) unless repo_contents_path.nil?
 
-        path = Pathname.new(File.join(directory, filename)).cleanpath.to_path
+        path = Pathname.new(File.join(directory, File.basename(filename))).cleanpath.to_path
```
thepwagner, line 139: Could you provide a failing test case for this?

I'm concerned about the impact of this change on ecosystems that fetch files from directories other than `directory`. Consider for example the GitHub Actions plugin:
* GitHub Actions workflows are stored in `.github/workflows`. By convention, Dependabot treats this as `directory == "/`" - https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#directory
* When the Actions file fetcher requests an individual workflow file, e.g. `.github/workflows/ci.yml`, it passes that full string to `fetch_file_from_host()` https://github.com/dependabot/dependabot-core/blob/e2ad9b9a7014dac6d8d06db3fc9df78efc24a2c4/github_actions/lib/dependabot/github_actions/file_fetcher.rb#L43

Given that scenario where `directory == "/"` and `filename == ".github/workflows/ci.yaml"`, I think this change means Dependabot will no longer fetch the file from the right path:
```
irb(main):001:0> directory = "/"
=> "/"
irb(main):002:0> filename = ".github/workflows/ci.yml"
=> ".github/workflows/ci.yml"
irb(main):003:0> File.join(directory, filename) # looks good
=> "/.github/workflows/ci.yml"
irb(main):004:0> File.join(directory, File.basename(filename)) # 404 file not found
=> "/ci.yml"
```

---

FWIW, this issue may be specific to the CodeCommit fetchers - since Dependabot was acquired by GitHub, the GitHub paths get much more attention. Maybe it is returning `directory` when it shouldn't be?

lorengordon: You're probably right. We're using a script based on dependabot-script to run Dependabot from a CodeBuild job, so we don't see all the GitHub stuff.

I did see the failing tests and realized there was likely more to what was going on than I understood at first. I assumed `directory` was always present/set, but that's obviously not correct. I'm now trying to understand what is different about the CodeCommit fetcher, or whether it is some artifact of the dependabot-script...

lorengordon: @thepwagner You were totally right. I found the root of the problem in `_codecommit_repo_contents()`, where it was returning the absolute path. Went ahead and updated the PR accordingly!

thepwagner (approved):
This would be a stronger PR if it included a test case, but given that it aligns the code_commit's mapping with:
* the Azure Repos mapping - https://github.com/dependabot/dependabot-core/blob/297f50e2267e70bf35cad7914fb02d3e1f4174dc/common/lib/dependabot/file_fetchers/base.rb#L293-L294
* the BitBucket mapping - https://github.com/dependabot/dependabot-core/blob/297f50e2267e70bf35cad7914fb02d3e1f4174dc/common/lib/dependabot/file_fetchers/base.rb#L316-L317

This really does seem like a bug that nobody noticed - thanks for the great catch!

lorengordon:
Thanks @thepwagner!

PR #3917, opened 2021-06-16
Double quote variables in shellscript

PeterDaveHello:
This will help prevent globbing and word splitting. Most of them are
already quoted, just few of them are missed.

PR #3932, opened 2021-06-17
Run linter against shell scripts

xlgmokha:
# Why is this needed?

To ensure that we follow the best practices for shell scripting described in
the Shellcheck wiki. e.g. [SC2006](https://github.com/koalaman/shellcheck/wiki/SC2006).

## What does this do?

It adds a `./bin/lint` script that can be used to run the shellcheck linter
on your machine. It also adds a job to the CI workflow to ensure that the
linter checks are passing.

xref: https://github.com/dependabot/dependabot-core/pull/3917#issuecomment-862823068

### Screenshots

Before:

```bash
モ ./bin/lint

In ./bin/docker-dev-shell line 14:
OPTS=`getopt -o hr --long help,rebuild -n 'parse-options' -- "$@"`
^--^ SC2034: OPTS appears unused. Verify use (or export if used externally).
     ^-- SC2006: Use $(...) notation instead of legacy backticked `...`.

Did you mean:
OPTS=$(getopt -o hr --long help,rebuild -n 'parse-options' -- "$@")


In ./bin/docker-dev-shell line 15:
if [ $? != 0 ]; then
     ^-- SC2181: Check exit code directly with e.g. 'if mycmd;', not indirectly with $?.


In ./bin/docker-dev-shell line 61:
echo $RUNNING
     ^------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean:
echo "$RUNNING"


In ./hex/helpers/build line 18:
case `uname` in
     ^-----^ SC2006: Use $(...) notation instead of legacy backticked `...`.

Did you mean:
case $(uname) in

For more information:
  https://www.shellcheck.net/wiki/SC2034 -- OPTS appears unused. Verify use (...
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
  https://www.shellcheck.net/wiki/SC2006 -- Use $(...) notation instead of le...
```

After:

```bash
モ ./bin/lint
~/src/github.com/xlgmokha/dependabot-core [turtles-in-a-half-shellcheck]
モ echo $?
0
```

## Type of change

### DIFF
--- a/bin/docker-dev-shell
+++ b/bin/docker-dev-shell
@@ -11,7 +11,7 @@ REBUILD=false
 # Enable docker buildkit with inline cache builds
 export DOCKER_BUILDKIT=1
 
-OPTS=`getopt -o hr --long help,rebuild -n 'parse-options' -- "$@"`
+OPTS=$(getopt -o hr --long help,rebuild -n 'parse-options' -- "$@")
 if [ $? != 0 ]; then
   echo "failed parsing options" >&2
   exit 1
--- a/bin/docker-dev-shell
+++ b/bin/docker-dev-shell
@@ -58,7 +58,7 @@ fi
 set +e
 RUNNING=$(docker ps --format '{{.Names}}' | grep "$CONTAINER_NAME")
 set -e
-echo $RUNNING
+echo "$RUNNING"
 if [ -n "$RUNNING" ]; then
   if [ -z "$BUILT_IMAGE" ]; then
     # image was not rebuilt - can we reuse existing?
--- a/bin/docker-dev-shell
+++ b/bin/docker-dev-shell
@@ -11,7 +11,9 @@ REBUILD=false
 # Enable docker buildkit with inline cache builds
 export DOCKER_BUILDKIT=1
 
+# shellcheck disable=SC2034
 OPTS=$(getopt -o hr --long help,rebuild -n 'parse-options' -- "$@")
+# shellcheck disable=SC2181
 if [ $? != 0 ]; then
   echo "failed parsing options" >&2
   exit 1
--- a/bin/lint
+++ b/bin/lint
@@ -6,4 +6,5 @@ shellcheck \
   ./bin/docker-dev-shell \
   ./bin/lint \
   ./*/helpers/build \
+  ./*/script/* \
   "$@"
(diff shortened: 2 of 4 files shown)


bin/docker-dev-shell
```diff
@@ -11,7 +11,9 @@ REBUILD=false
 # Enable docker buildkit with inline cache builds
 export DOCKER_BUILDKIT=1
 
-OPTS=`getopt -o hr --long help,rebuild -n 'parse-options' -- "$@"`
+# shellcheck disable=SC2034
+OPTS=$(getopt -o hr --long help,rebuild -n 'parse-options' -- "$@")
+# shellcheck disable=SC2181
```
xlgmokha, line 16: I added the ignores above to skip the following:

```bash
モ ./bin/lint

In ./bin/docker-dev-shell line 14:
OPTS=$(getopt -o hr --long help,rebuild -n 'parse-options' -- "$@")
^--^ SC2034: OPTS appears unused. Verify use (or export if used externally).


In ./bin/docker-dev-shell line 15:
if [ $? != 0 ]; then
     ^-- SC2181: Check exit code directly with e.g. 'if mycmd;', not indirectly with $?.

For more information:
  https://www.shellcheck.net/wiki/SC2034 -- OPTS appears unused. Verify use (...
  https://www.shellcheck.net/wiki/SC2181 -- Check exit code directly with e.g...
```

bin/lint
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+set -e
+
+shellcheck \
```
jurre, line 5: Should we add this dependency to the development container as well, so we can run it locally and don't need to rely on CI when editing scripts?

jurre (approved):
I had one question about installing the new dependency in the dev container

jurre:
Ah, ok automerge was enabled. I suppose we can add shellcheck to the dev container in a separate PR

bin/lint
xlgmokha: Ah yes, good call. I'm sorry about the auto-merge. I'll fire up another PR to add this.

xlgmokha: I opened this: https://github.com/dependabot/dependabot-core/pull/3938

PR #3935, opened 2021-06-17
Remove headers from HTTP fixture recordings

xlgmokha:
# Why is this needed?

To remove non-essential data from test fixtures.

## What does this do?

To remove headers that aren't needed.

### DIFF
--- a/github_actions/spec/fixtures/vcr_cassettes/Dependabot_GithubActions_UpdateChecker/_updated_requirements/with_multiple_requirement_sources/1_5_4_1.yml
+++ b/github_actions/spec/fixtures/vcr_cassettes/Dependabot_GithubActions_UpdateChecker/_updated_requirements/with_multiple_requirement_sources/1_5_4_1.yml
@@ -3,37 +3,13 @@
   - request:
       method: get
       uri: https://github.com/actions/checkout.git/info/refs?service=git-upload-pack
-      body:
-        encoding: US-ASCII
-        string: ''
-      headers:
-        User-Agent:
-        - excon/0.73.0
-        Accept:
-        - "*/*"
-        Authorization:
-        - Basic eC1hY2Nlc3MtdG9rZW46NDFlZTdiODA0MmEwZWNjMmFiYmNkNDViNzkwZTc1NWM0NWRkMDI5Yw==
     response:
       status:
         code: 200
         message: OK
       headers:
-        Server:
-        - GitHub Babel 2.0
         Content-Type:
         - application/x-git-upload-pack-advertisement
-        Expires:
-        - Fri, 01 Jan 1980 00:00:00 GMT
-        Pragma:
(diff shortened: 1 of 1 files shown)


Nishnha (approved):
LGTM 👍🏾

jurre:
@xlgmokha I don't think we should manually modify these fixtures, those fixtures are recorded by [vcr](https://github.com/vcr/vcr), this means that any time we re-record them that data will be put back.

VCR has [some hooks](https://relishapp.com/vcr/vcr/v/1-10-1/docs/configuration/filter-sensitive-data) to scrub sensitive data, but other than that we should leave these generated files alone IMO.

feelepxyz:
I added some hooks to scrub headers from recordings but missed this old one.

xlgmokha:
Larger recordings make it easier to miss things like accidental inclusion of sensitive data. We purposefully commit manifest files with the minimum data required to isolate defects/features and I think this change aligns with that convention.

Is it possible to specify an allow list of headers rather than a deny list?

Nishnha:
It looks like this VCR recording included an Authorization token but it should have been removed via https://github.com/dependabot/dependabot-core/blob/1a9e69f78066ed135d30fbfc57736e762e95d3eb/common/spec/spec_helper.rb#L66

As far as I can tell, VCR recordings are automated--do we have a way of debugging to find out how this key snuck into the file?

xlgmokha:
> As far as I can tell, VCR recordings are automated--do we have a way of debugging to find out how this key snuck into the file?

A couple of thoughts come to mind. Is it possible that the header was committed before we added the filter? We could try deleting the generated recording and re-running the test and debug the hook to see how it works.

feelepxyz:
Yeah the fixture with the auth header was created just over a year ago: https://github.com/xlgmokha/dependabot-core/commit/f3ef00bbe610998408cb51809b5e09f122b3cf45

I only added the code to strip these out 4 days ago https://github.com/dependabot/dependabot-core/commit/c35256e3b3cd41945aa1b8063081d1cc9bd52dfb

We shouldn't see these added in future recordings.

jurre:
> We purposefully commit manifest files with the minimum data required to isolate defects/features and I think this change aligns with that convention.

For any files that we manually add I think this is a great default, but for generated files it seems needlessly painful and error-prone, we might end up removing headers that turn out to be relevant and future changes might break production without being covered by our tests.

I tend to ~never read these generated files, if we're worried about committing auth-headers maybe a better approach would be to add a linter to catch these?

feelepxyz:
I think there's some miscommunication here? This PR only strips out an old header that was added before this code to automatically strip them out was added: https://github.com/dependabot/dependabot-core/commit/c35256e3b3cd41945aa1b8063081d1cc9bd52dfb

Pretty sure we shouldn't need to worry about this again.

PR #3936, opened 2021-06-17
Enable available rubocops

thepwagner:
This is an idle hands PR: I upgraded my system and wanted to verify this toolchain wasn't broken.

The changeset begins by adding a `rake rubocop:sort` task, which does a recursive sort of `.rubocop.yml` for 💄 .
From there, new but not enabled cops are enabled in 3 phases with autofix+sort applied at each round.

When an autofix wasn't available, the rule was disabled (I didn't write any code):
* [`Lint/NoReturnInBeginEndBlocks`](https://rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/NoReturnInBeginEndBlocks)
* [`Lint/DuplicateBranch`](https://rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/DuplicateBranch)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,205 +1,143 @@
+---
 AllCops:
   DisplayCopNames: true
-  TargetRubyVersion: 2.5
   Exclude:
-    - '*/vendor/**/*'
-    - '*/bin/**/*'
-    - '**/tmp/**/*'
-    - '*/spec/fixtures/**/*'
-
+  - "*/vendor/**/*"
+  - "*/bin/**/*"
+  - "**/tmp/**/*"
+  - "*/spec/fixtures/**/*"
+  TargetRubyVersion: 2.5
 Layout/DotPosition:
   EnforcedStyle: trailing
-
 Layout/EmptyLinesAroundAttributeAccessor:
   Enabled: false
-
 Layout/FirstArrayElementIndentation:
   EnforcedStyle: consistent
-
 Layout/FirstHashElementIndentation:
   EnforcedStyle: consistent
-
 Layout/LineLength:
   Max: 120
-
 Layout/RescueEnsureAlignment:
   Enabled: false
-
 Layout/SpaceAroundMethodCallOperator:
   Enabled: false
-
-Lint/DeprecatedOpenSSLConstant:
-  Enabled: false
-
-Lint/MixedRegexpCaptureTypes:
-  Enabled: false
-
-Lint/RaiseException:
-  Enabled: false
-
-Lint/StructNewOverride:
-  Enabled: false
-
 Lint/BinaryOperatorWithIdenticalOperands:
   Enabled: true
-
-Lint/DuplicateRequire:
-  Enabled: true
-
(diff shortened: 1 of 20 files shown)


.rubocop.yml
```diff
@@ -1,205 +1,212 @@
...
-    - '*/vendor/**/*'
-    - '*/bin/**/*'
-    - '**/tmp/**/*'
-    - '*/spec/fixtures/**/*'
-
+  - "*/vendor/**/*"
+  - "*/bin/**/*"
+  - "**/tmp/**/*"
+  - "*/spec/fixtures/**/*"
+  TargetRubyVersion: 2.5
```
xlgmokha, line 9: unrelated: Ruby 2.5 is now EOL. Should we consider updating this to 2.6? https://www.ruby-lang.org/en/downloads/branches/

Rakefile
```diff
@@ -86,6 +87,25 @@ namespace :gems do
...
 
+class Hash
+  def sort_by_key(recursive = false, &block)
+    keys.sort(&block).each_with_object({}) do |key, seed|
+      seed[key] = self[key]
+      seed[key] = seed[key].sort_by_key(true, &block) if recursive && seed[key].is_a?(Hash)
+      seed
+    end
+  end
+end
```
xlgmokha, line 98: I have never used Ruby's refinements feature but this might actually fit it's intended use case. https://docs.ruby-lang.org/en/2.4.0/syntax/refinements_rdoc.html. Also, I would like to add... https://www.youtube.com/watch?v=plzw1NWVYSM&t=8s

xlgmokha (approved):
I added a few comments but none of them are urgent so please feel free to ignore.

Rakefile
thepwagner: Disclosure: I typed "rubocop.yml sort" into google and built this from https://gist.github.com/ream88/e510ca628d35515e5cbbcc18e2ea3238 _ream88 is the real MVP!

Being critical: I think `&block` could be trimmed. I don't mind [cargo-culting](https://en.wikipedia.org/wiki/Cargo_cult_programming) it along, but it's not needed.

.rubocop.yml
thepwagner: I tried `2.7` when I saw it in the diff, but that nets a warning that it should match [spec.required_ruby_version](https://github.com/dependabot/dependabot-core/blob/dadd9bb5408604c633590de6675f751e0f0b46d0/common/dependabot-common.gemspec#L21)

```
dependabot-common.gemspec:21:32: C: Gemspec/RequiredRubyVersion: required_ruby_version (2.5, declared in dependabot-common.gemspec) and TargetRubyVersion (2.7, which may be specified in .rubocop.yml) should be equal.
  spec.required_ruby_version = ">= 2.5.0"
                               ^^^^^^^^^^
```

I was nervous about impacting https://rubygems.org/gems/dependabot-common for the linter. I'd also like this to be `2.7`, but that seemed like a big enough change to consider separately from this PR's scope of non-structural robot patches.

PR #3937, opened 2021-06-18
Fix the terraform build

xlgmokha:
# Why is this needed?

To fix the failing build caused by the release of new upstream software used in a test.

## What does this do?

It updates the fixture constraints to be loose enough to allow for updates but
also ensure that the test can be deterministic.


jurre (approved):
Thanks Mo, nice one 🚀

PR #3954, opened 2021-06-21
Terraform: Handle "terraform-providers" namespace

feelepxyz:
Handle dependencies that have been declared with the
`terraform-providers` namespace, .e.g "terraform-providers/fastly".

This seems to always map to packages where the namespace matches the
name and there's also a package with the same name. Not all providers
where a package matches the namespace supports this but they also can't
be installed with `terraform-providers` as the namespace. e.g. 
`terraform-providers/checkly` doesn't work but `checkly/checkly` works.

Only changing the dependency name as this is used to later fetch the
dependency source.

### DIFF
--- a/terraform/lib/dependabot/terraform/file_parser.rb
+++ b/terraform/lib/dependabot/terraform/file_parser.rb
@@ -91,12 +91,16 @@ module Dependabot
 
         source_address = details.fetch("source", nil)
         version_req = details["version"]&.strip
-        hostname, namespace, name = provider_source_from(source_address, name)
-        dependency_name = source_address ? "#{namespace}/#{name}" : name
+        hostname, file_namespace, name = provider_source_from(source_address, name)
+        # NOTE: terraform-providers/name is a shorthard for the provider name
+        # matching the namespace, we separate this from the file namespace
+        # needed to find the provider in the parsed file
+        dependency_namespace = file_namespace == "terraform-providers" ? name : file_namespace
+        dependency_name = source_address ? "#{dependency_namespace}/#{name}" : name
 
         Dependency.new(
           name: dependency_name,
-          version: determine_version_for(hostname, namespace, name, version_req),
+          version: determine_version_for(hostname, file_namespace, name, version_req),
           package_manager: "terraform",
           requirements: [
             requirement: version_req,
@@ -105,7 +109,7 @@ module Dependabot
             source: {
               type: "provider",
               registry_hostname: hostname,
-              module_identifier: "#{namespace}/#{name}"
+              module_identifier: "#{file_namespace}/#{name}"
             }
           ]
         )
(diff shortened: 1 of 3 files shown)


terraform/spec/dependabot/terraform/file_parser_spec.rb
```diff
@@ -666,6 +666,16 @@
         expect { subject }.to raise_error(Dependabot::PrivateSourceAuthenticationFailure) do |boom|
           expect(boom.source).to eq("artifactory.dependabot.com")
         end
+    end
+
+    context "with a provider that uses the terraform-providers namespace" do
+      let(:files) { project_dependency_files("terraform_providers_namespace") }
+
+      it "has the right details" do
+        dependency = dependencies.find { |d| d.name == "fastly/fastly" }
```
feelepxyz, line 675: maps to https://registry.terraform.io/providers/fastly/fastly/latest

terraform/spec/fixtures/projects/terraform_providers_namespace/main.tf
```diff
@@ -0,0 +1,9 @@
+terraform {
+  required_providers {
+    fastly = {
+      source = "terraform-providers/fastly"
```
feelepxyz, line 4: I wonder if this is a legacy mapping that only works with some set of old providers?

jurre (approved):

terraform/spec/fixtures/projects/terraform_providers_namespace/main.tf
feelepxyz: It's also not a shorthand for `hashicorp` as this does not work:

```
terraform {
  required_providers {
    consul = {
      source = "terraform-providers/consul"
    }
  }
}
```

Although, `source = "hashicorp/consul"` and `source = "hashicorp/consul"` works.

feelepxyz: Found this reference, looks like a legacy thing https://github.com/hashicorp/terraform/blob/d35bc0531255b496beb5d932f185cbcdb2d61a99/internal/getproviders/didyoumean_test.go#L48-L54

terraform/lib/dependabot/terraform/file_parser.rb
```diff
@@ -91,12 +91,16 @@ def build_provider_dependency(file, name, details = {})
 
         source_address = details.fetch("source", nil)
         version_req = details["version"]&.strip
-        hostname, namespace, name = provider_source_from(source_address, name)
-        dependency_name = source_address ? "#{namespace}/#{name}" : name
+        hostname, file_namespace, name = provider_source_from(source_address, name)
+        # NOTE: terraform-providers/name is a shorthard for the provider name
+        # matching the namespace, we separate this from the file namespace
+        # needed to find the provider in the parsed file
+        dependency_namespace = file_namespace == "terraform-providers" ? name : file_namespace
```
feelepxyz, line 98: Hm this isn't right, I think we need to make a request to get the correct namespace, `terraform-providers/postgresql` does not map to `postgresql/postgresql`.

PR #3958, opened 2021-06-21
Terraform: handle nested module sources

feelepxyz:
Relax the regex that matches module sources to include nested modules.

Nested modules mentioned in docs here: https://www.terraform.io/docs/language/modules/develop/structure.html

### DIFF
--- a/terraform/lib/dependabot/terraform/file_updater.rb
+++ b/terraform/lib/dependabot/terraform/file_updater.rb
@@ -157,7 +157,11 @@ module Dependabot
         %r{
           (?<=\{)
           (?:(?!^\}).)*
-          source\s*=\s*["'](#{Regexp.escape(registry_host_for(dependency))}/)?#{Regexp.escape(dependency.name)}["']
+          source\s*=\s*["']
+            (#{Regexp.escape(registry_host_for(dependency))}/)?
+            #{Regexp.escape(dependency.name)}
+            (//modules/\S+)?
+            ["']
           (?:(?!^\}).)*
         }mx
       end
--- a/terraform/spec/fixtures/projects/nested_modules/main.tf
+++ b/terraform/spec/fixtures/projects/nested_modules/main.tf
@@ -1,0 +1,4 @@
+module "github_terraform" {
+  source  = "terraform-aws-modules/iam/aws//modules/iam-assumable-role"
+  version = "4.0.0"
+}
--- a/terraform/spec/dependabot/terraform/file_updater_spec.rb
+++ b/terraform/spec/dependabot/terraform/file_updater_spec.rb
@@ -791,5 +791,52 @@ RSpec.describe Dependabot::Terraform::FileUpdater do
         )
       end
     end
+
+    describe "for a nested module" do
+      let(:files) { project_dependency_files("nested_modules") }
+      let(:dependencies) do
+        [
+          Dependabot::Dependency.new(
+            name: "terraform-aws-modules/iam/aws",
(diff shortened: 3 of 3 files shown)


terraform/spec/fixtures/projects/nested_modules/main.tf
```diff
@@ -0,0 +1,5 @@
+module "github_terraform_read_only" {
+  source  = "terraform-aws-modules/iam/aws//modules/iam-assumable-role"
```
xlgmokha, line 2: Is the `//` intentional?

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -157,7 +157,8 @@ def registry_declaration_regex
         %r{
           (?<=\{)
           (?:(?!^\}).)*
-          source\s*=\s*["'](#{Regexp.escape(registry_host_for(dependency))}/)?#{Regexp.escape(dependency.name)}["']
+          source\s*=\s*["'](#{Regexp.escape(registry_host_for(dependency))}/)?
+            #{Regexp.escape(dependency.name)}(//modules/[^"']+)?["']
```
xlgmokha, line 161: > (//modules/[^"']+)?

A couple of questions:

* Is the `^` needed here?
* I was expecting to see `\/` to escape the slash but I think that `%r` removes the need for escaping. Is the extra leading `/` needed?
* What is the optional `[^"'` mean to handle? Examples if possible pretty please.

Is this the equivalent of?

```ruby
          source\s*=\s*["']
            (#{Regexp.escape(registry_host_for(dependency))}/)?
            #{Regexp.escape(dependency.name)}
            (/modules/\S*)?
            ["']
```

terraform/spec/fixtures/projects/nested_modules/main.tf
```diff
@@ -0,0 +1,5 @@
...
+  version = "4.0.0"
+  role_name        = "github-terraform-read-only"
```
xlgmokha, line 4: nit: I think this is non-essential. Is it possible to remove it?

xlgmokha (approved):
I left a couple of questions for my own clarification. Otherwise, LGTM.

terraform/spec/fixtures/projects/nested_modules/main.tf
feelepxyz: Yeah this seems like a convention for getting nested modules, there are some examples here: https://registry.terraform.io/modules/terraform-aws-modules/iam/aws/latest

Using `source  = "terraform-aws-modules/iam/aws/modules/iam-assumable-role"` raised an error from terraform:  `Cannot apply a version constraint to module "github_terraform_read_only" (at main.tf:9) because it has a non Registry URL.`

terraform/spec/fixtures/projects/nested_modules/main.tf
feelepxyz: Yup removing!

terraform/lib/dependabot/terraform/file_updater.rb
feelepxyz: The optional part is attempting to match the nested module path.
Input: `source  = "terraform-aws-modules/iam/aws//modules/iam-assumable-role"` match `//modules/iam-assumable-role`, terraform enforces the `//` in the path.

I was attempting to match everything after `//modules/` up until the the ending quote with `[^'"]+` but `\S+` would be equivalent 👍

PR #3911, opened 2021-06-15
bundler: ignore invalid auth_uri

thepwagner:
This PR handles what I'm assuming is a garbage-in-garbage-out case of misconfigured `git` sources in the bundle ecosystem.

The goal is to avoid raising the following exception, which contains potentially private information - the misconfigured dependency.

<details><summary>exception of interest</summary>
<p>

```
Dependabot::SharedHelpers::HelperSubprocessFailed: Git error: command `git clone OWNER/REPO /opt/bundler/v2/.bundle/ruby/2.6.0/cache/bundler/git/REPO-0d7c3906c195220fefe0cae81e8a14e493ab554c --bare --no-hardlinks --quiet` in directory /home/dependabot/dependabot-updater/dependabot_tmp_dir has failed.

If this error persists you could try removing the cache directory '/home/dependabot/dependabot-updater/dependabot_tmp_dir'
  from dependabot/shared_helpers.rb:115:in `run_helper_subprocess'
  from dependabot/bundler/native_helpers.rb:12:in `block in run_bundler_subprocess'
  from bundler.rb:309:in `block in with_original_env'
  from bundler.rb:562:in `with_env'
  from bundler.rb:309:in `with_original_env'
  from dependabot/bundler/native_helpers.rb:11:in `run_bundler_subprocess'
  from dependabot/bundler/update_checker/version_resolver.rb:82:in `block (2 levels) in fetch_latest_resolvable_version_details'
  from dependabot/bundler/update_checker/shared_bundler_helpers.rb:54:in `block in in_a_native_bundler_context'
  from dependabot/shared_helpers.rb:45:in `block (2 levels) in in_a_temporary_directory'
  from dependabot/shared_helpers.rb:45:in `chdir'
  from dependabot/shared_helpers.rb:45:in `block in in_a_temporary_directory'
  from tmpdir.rb:93:in `mktmpdir'
  from dependabot/shared_helpers.rb:42:in `in_a_temporary_directory'
  from dependabot/shared_helpers.rb:36:in `in_a_temporary_repo_directory'
  from dependabot/bundler/update_checker/shared_bundler_helpers.rb:50:in `in_a_native_bundler_context'
  from dependabot/bundler/update_checker/version_resolver.rb:81:in `block in fetch_latest_resolvable_version_details'
  from dependabot/shared_helpers.rb:159:in `with_git_configured'
  from dependabot/bundler/update_checker/version_resolver.rb:77:in `fetch_latest_resolvable_version_details'
[...]

PR #3959, opened 2021-06-21
bin/ci-test

thepwagner:
This brings back the helper added in https://github.com/dependabot/dependabot-core/pull/3398 . I'd re-written it for https://github.com/dependabot/dependabot-core/pull/3911 , then @Nishnha asked about the same thing.

The goal of this script is to replicate the build performed by Actions in CI, https://github.com/dependabot/dependabot-core/blob/main/.github/workflows/ci.yml by bypass remote Docker caching.

It's still not a good idea to have this build logic defined in two places; I just wanted feedback looks from Actions environment (which isn't _quite_ the same as the development shell) without pushing.

### DIFF
--- a/bin/ci-test
+++ b/bin/ci-test
@@ -1,0 +1,56 @@
+#!/bin/bash -e
+#/ usage: ci-test SUITE
+#/ simulate the process followed by ~/.github/workflows/ci.yml
+
+function print_usage() {
+  grep ^#/ "$0" | cut -c4-
+}
+
+function handle_args() {
+  export SUITE=$1
+  case $SUITE in
+    bundler1 | bundler2)
+      export MODULE=bundler
+      ;;
+    "")
+      print_usage
+      exit 1
+      ;;
+    *)
+      export MODULE=$SUITE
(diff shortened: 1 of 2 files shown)


Nishnha:
Bless'ed be!

Nishnha (approved):

bin/ci-test
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash -e
+#/ usage: ci-test SUITE
+#/ simulate the process followed by ~/.github/workflows/ci.yml
```
xlgmokha, line 3: 💯

bin/lint
```diff
@@ -3,6 +3,7 @@
 set -e
 
 shellcheck \
+  ./bin/ci-test \
```
xlgmokha, line 6: 🎉

xlgmokha (approved):
:shipit:

thepwagner:
I still don't like this solution: it's a hack, but a valuable one. 🩹 :shipit:

PR #3961, opened 2021-06-21
Dev shell gemspecs

thepwagner:
`bin/docker-dev-shell` and the `devcontainer.json` configuration remap some ecosystems' gemspec to `dependabot-core.gemspec`.  This impacts: `cargo`, `git_submodules`, `maven`, `nuget`.
That means in the development environment, two gemspecs are defined: `cargo/dependabot-cargo.gemspec`, `cargo/dependabot-core.gemspec`. The latter doesn't seem to be used by anything.

This PR assumes `dependabot-core.gemspec` is an error, and updates scripts to volume mount `dependabot-cargo.gemspec` into development containers.


mctofu (approved):
This looks better! Was this breaking anything in the dev container?

thepwagner:
> Was this breaking anything in the dev container?

Nothing broke, it's just a weird path to nowhere that might have sent someone else down the same "wait, why is dependabot doing this?!" chain.

It manifested as duplicate entries when using gemspecs to list modules, like https://github.com/dependabot/dependabot-core/pull/3959/files#diff-d446418a630555d896ab07a4ab73474abfef93c362d3020703812b055cf6d581R26-R29

PR #3971, opened 2021-06-22
GemspecSanitizer: replace interpolated strings

thepwagner:
The previous impl only replaced the components of the interpolated
string that looked like versions.

When this is called with a `replacement_version` read from
`Gemfile.lock`, the replacement value is based on the interpolated string. That means replacing the entire string is more correct than swapping individual components.


feelepxyz (approved):
Nice fix, LGTM 👍

PR #3976, opened 2021-06-23
Terraform: fix module updates with a lockfile

feelepxyz:
This is the other side of https://github.com/dependabot/dependabot-core/pull/3968

This fixes module updates when there's a lockfile present. Modules are
not included in the lockfile so adding a guard to only attempt lockfile
updates for providers.

Fixed some test setup to match what the parsed files would look like.

### DIFF
--- a/terraform/spec/dependabot/terraform/file_updater_spec.rb
+++ b/terraform/spec/dependabot/terraform/file_updater_spec.rb
@@ -593,25 +593,25 @@ RSpec.describe Dependabot::Terraform::FileUpdater do
       let(:dependencies) do
         [
           Dependabot::Dependency.new(
-            name: ".terraform.lock.hcl",
-            version: "3.44.0",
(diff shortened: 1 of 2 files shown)


terraform/spec/dependabot/terraform/file_updater_spec.rb
```diff
@@ -691,21 +691,6 @@ module "consul" {
             provider "registry.terraform.io/hashicorp/aws" {
               version     = "3.45.0"
               constraints = ">= 3.37.0, < 3.46.0"
-              hashes = [
-                "h1:LKU/xfna87/p+hl5yTTW3dvOqWJp5JEM+Dt3nnvSDvA=",
```
feelepxyz, line 695: This expectation was flaking on my machine as the hash is changing

jurre (approved):

PR #3978, opened 2021-06-23
Terraform: always clone repository contents

feelepxyz:
Terraform projects can include local path modules that are currently not
fetched in the file fetching step so any lockfile update will fail
because these files are missing.

The easiest fix seemed to be to always start cloning terraform projects
as this is what we want to end up doing for all ecosystems.

The file fetcher will still hit the gh api when getting the [repo contents](https://github.com/dependabot/dependabot-core/blob/40baefd1e812feca9712daa1e359583a29b5fdc1/common/lib/dependabot/file_fetchers/base.rb#L171-L178)
but this would require changes in common so keen to make that change
separately as it currently works.

### DIFF
--- a/terraform/lib/dependabot/terraform.rb
+++ b/terraform/lib/dependabot/terraform.rb
@@ -17,3 +17,6 @@ Dependabot::PullRequestCreator::Labeler.
 require "dependabot/dependency"
 Dependabot::Dependency.
   register_production_check("terraform", ->(_) { true })
+
+require "dependabot/utils"
+Dependabot::Utils.register_always_clone("terraform")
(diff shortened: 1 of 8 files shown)


terraform/lib/dependabot/terraform.rb
```diff
@@ -17,3 +17,6 @@
 require "dependabot/dependency"
 Dependabot::Dependency.
   register_production_check("terraform", ->(_) { true })
+
+require "dependabot/utils"
+Dependabot::Utils.register_always_clone("terraform")
```
feelepxyz, line 22: I'm pretty sure this is all that needs to be done, in addition to the file updater change to call `SharedHelpers.in_a_temporary_repo_directory`. Tested with a dry-run and looks like the updater checks this flag to turn on cloning.

feelepxyz: Any reason we shouldn't be doing this?

jurre: I think this is it! 👍

jurre (approved):

PR #3983, opened 2021-06-24
Terraform: Configure git for `terraform init` and capture errors

jurre:
When attempting to `terraform init` when there are private module
sources present, currently we raise a `HelperSubprocessFailed`.

Terraform will attempt to use any git credentials that are configured,
so passing these along should help in some cases when Dependabot is
configured to access these, and when it isn't we should raise a
PrivateSourceAuthenticationFailure to communicate more clearly that we
cannot reach a certain source.

It's a little unfortunate that we even need to do this, because this
happens when we attempt to update a lockfile, and these don't support
modules. However, `terraform providers lock` will still complain about
modules needing to be installed, and this should do a better job of
that.

### DIFF
--- a/terraform/lib/dependabot/terraform/file_updater.rb
+++ b/terraform/lib/dependabot/terraform/file_updater.rb
@@ -11,6 +11,8 @@ module Dependabot
     class FileUpdater < Dependabot::FileUpdaters::Base
       include FileSelector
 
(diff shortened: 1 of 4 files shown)


feelepxyz (approved):
Nice one! LGTM 👍

jurre:
The bundler1 failure is a known flake, gonna go ahead and merge this. We can look into the flakey test separately

Issue #3909, opened 2021-06-15
Terraform lockfile update failed when using modules


hosht:
**Package ecosystem**
Terraform

**Language version**
1.0.0

**Manifest location and content prior to update**
https://github.com/hosht/dependabot-terraform/tree/main/example5

**dependabot.yml content**
https://github.com/hosht/dependabot-terraform/blob/main/.github/dependabot.yml

**What you expected to see, versus what you actually saw**

When the root module has child modules, terraform providers lock command failed.
To fix it, running terraform init before.

Error output
```
updater | INFO <job_161564816> Starting job processing
updater | INFO <job_161564816> Starting update job for hosht/dependabot-terraform
updater | INFO <job_161564816> Checking if hashicorp/null 3.0.0 needs updating
  proxy | 2021/06/15 15:40:19 [012] GET https://registry.terraform.io:443/.well-known/terraform.json
  proxy | 2021/06/15 15:40:19 [012] 200 https://registry.terraform.io:443/.well-known/terraform.json
  proxy | 2021/06/15 15:40:19 [014] GET https://registry.terraform.io:443/v1/providers/hashicorp/null/versions
  proxy | 2021/06/15 15:40:19 [014] 200 https://registry.terraform.io:443/v1/providers/hashicorp/null/versions
updater | INFO <job_161564816> Latest version is 3.1.0
updater | INFO <job_161564816> Requirements to unlock own
updater | INFO <job_161564816> Requirements update strategy 
updater | INFO <job_161564816> Updating hashicorp/null from 3.0.0 to 3.1.0
updater | I, [2021-06-15T15:40:19.571809 #7]  INFO -- sentry: ** [Raven] Sending event b6f835a1f2864b80b01fcb292b587a8b to Sentry
  proxy | 2021/06/15 15:40:19 [017] POST https://sentry.io:443/api/1451818/store/
  proxy | 2021/06/15 15:40:19 [015] WARN: Cannot handshake client checkpoint-api.hashicorp.com:443 EOF
  proxy | 2021/06/15 15:40:19 [017] 200 https://sentry.io:443/api/1451818/store/
updater | ERROR <job_161564816> Error processing hashicorp/null (Dependabot::SharedHelpers::HelperSubprocessFailed)
updater | ERROR <job_161564816> ╷
  proxy | 2021/06/15 15:40:19 [021] GET https://registry.terraform.io:443/.well-known/terraform.json
[...]

PR #3986, opened 2021-06-25
Terraform: Prevent `terraform init` from initializing backends

hfurubotten:
Some terraform setups have backend (where the state is stored) configurations present in the code. When this is the case the `terraform init` command will fail trying to initialize the backend. Most backend types can't be initialized by dependabot as where they are pointing should be well protected or require user input to continue. Example of error message: https://github.com/dependabot/dependabot-core/issues/3909#issuecomment-866239727

I have added the following options to the `terraform init` command:

- `-backend=false` to ignore any backend configuration
- `-input=false` to immediately fail if it needs user input
- `-no-color` to prevent color characters from printing, which means `strip_terminal_colors` could be removed

Hopefully I have gotten things correct set up here. First time I'm writing in Ruby 😄

### DIFF
--- a/terraform/lib/dependabot/terraform/file_updater.rb
+++ b/terraform/lib/dependabot/terraform/file_updater.rb
@@ -133,7 +133,10 @@ module Dependabot
 
       def run_terraform_init
         SharedHelpers.with_git_configured(credentials: credentials) do
-          SharedHelpers.run_shell_command("terraform init")
+          # -backend=false option used to ignore any backend configuration, as these won't be accessible
+          # -input=false option used to immediately fail if it needs user input
+          # -no-color option used to prevent any color characters being printed in the output 
+          SharedHelpers.run_shell_command("terraform init -backend=false -input=false -no-color")
         rescue SharedHelpers::HelperSubprocessFailed => e
           output = strip_terminal_colors(e.message)
 
--- a/terraform/lib/dependabot/terraform/file_updater.rb
+++ b/terraform/lib/dependabot/terraform/file_updater.rb
@@ -135,19 +135,15 @@ module Dependabot
         SharedHelpers.with_git_configured(credentials: credentials) do
           # -backend=false option used to ignore any backend configuration, as these won't be accessible
           # -input=false option used to immediately fail if it needs user input
-          # -no-color option used to prevent any color characters being printed in the output 
+          # -no-color option used to prevent any color characters being printed in the output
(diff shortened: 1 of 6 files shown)


jurre:
This is fantastic, I had been meaning to figure this out!

Since you added the `-no-color` flag, we can get rid of the `strip_terminal_colors` method, this seems much better. Feel free to do it in this PR or I'm happy to do it as a follow-up

hfurubotten:
I removed the `strip_terminal_colors` method, and fixed my lint violation. Good to do the whole update in the same PR.

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -133,20 +133,19 @@ def update_lockfile_declaration # rubocop:disable Metrics/AbcSize
...
 
           if output.match?(PRIVATE_MODULE_ERROR)
             raise PrivateSourceAuthenticationFailure, output.match(PRIVATE_MODULE_ERROR).named_captures.fetch("repo")
           end
         end
       end
 
-      def strip_terminal_colors(output)
-        output.gsub(/\e\[(\d+)m/, "")
-      end
```
feelepxyz, line 148: Nice!

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -133,20 +133,19 @@ def update_lockfile_declaration # rubocop:disable Metrics/AbcSize
 
       def run_terraform_init
         SharedHelpers.with_git_configured(credentials: credentials) do
-          SharedHelpers.run_shell_command("terraform init")
+          # -backend=false option used to ignore any backend configuration, as these won't be accessible
```
feelepxyz, line 136: 🎉

hfurubotten:
Look like the `azurerm` provider has been updated between my CI runs here, which makes the tests fail. The azure provider releases a new version at least once a week, so might become a problem using them as test cases.. 

Should I just update the tests to match the new version released today, or do you guys have any ideas for how to fix this a bit more permanently?

jurre:
> Should I just update the tests to match the new version released today, or do you guys have any ideas for how to fix this a bit more permanently?

Ha, glad we're catching this now! We've previously resolved this by pinning to a range: https://github.com/dependabot/dependabot-core/pull/3937/files

hfurubotten:
Nice! Will update the test fixtures to contain pinned versions

jurre (approved):

PR #3987, opened 2021-06-25
poetry: ignore path dependencies in UpdateChecker

thepwagner:
This PR updates the python `UpdateChecker` to ignore dependencies specified as a poetry [path dependency](https://python-poetry.org/docs/dependency-specification/#path-dependencies).

There are two points of concern:
* When querying for the latest versions, available indexes were queried. This is a privacy concern: customers with monorepos will leak every submodule's name to PyPI. (In my test setup, my internal module that I named  `nested` was compared to https://pypi.org/project/Nested/ 😨 ).
* When invoking `poetry` to query for resolvable versions, dependabot [injects a `version` string to dependencies](https://github.com/dependabot/dependabot-core/blob/dce1b23793935cc0080729e90d8f3ba0bd52b5c3/python/lib/dependabot/python/update_checker/poetry_version_resolver.rb#L271-L275).

The latter causes a manifest like:
```toml
[tool.poetry.dependencies]
python = "^3.9"
nested = { path = "nested" }
```

To be re-written by Dependabot and fed to poetry as:
```toml
[tool.poetry.dependencies]
python = "^3.9"
nested = { path = "nested", version = ">= 0.1.0, <= 1.0.1" }
```

Which results in:
```
rsion_resolver.rb:88:in `block (2 levels) in fetch_latest_resolvable_version_string'
/home/dependabot/dependabot-core/python/lib/dependabot/python/update_checker/poetry_version_resolver.rb:338:in `run_poetry_command': RuntimeError (Dependabot::SharedHelpers::HelperSubprocessFailed)

  The Poetry configuration is invalid:
    - [dependencies.nested] {'path': 'nested', 'version': '>= 0.1.0, <= 1.0.1'} is not valid under any of the given schemas
  

  at /usr/local/.pyenv/versions/3.9.5/lib/python3.9/site-packages/poetry/core/factory.py:43 in create_poetry
       39│             message = ""
       40│             for error in check_result["errors"]:
       41│                 message += "  - {}\\n".format(error)
       42│ 
    →  43│             raise RuntimeError("The Poetry configuration is invalid:\\n" + message)
       44│ 
       45│         # Load package
       46│         name = local_config["name"]
       47│         version = local_config["version"]
```

Splitting between the `PoetryVersionResolver` and `UpdateChecker` is an unintentional byproduct of implementation order - I'd intended to only change the `PoetryVersionResolver`, but I didn't like seeing the PyPI results in the dry-run script output.

### DIFF
--- a/python/lib/dependabot/python/update_checker.rb
+++ b/python/lib/dependabot/python/update_checker.rb
@@ -252,6 +252,8 @@ module Dependabot
       end
 
       def fetch_latest_version
+        return nil if PoetryVersionResolver.path_dependency?(pyproject.content, dependency)
+
         latest_version_finder.latest_version
       end
 
--- a/python/lib/dependabot/python/update_checker.rb
+++ b/python/lib/dependabot/python/update_checker.rb
@@ -252,7 +252,8 @@ module Dependabot
       end
 
       def fetch_latest_version
-        return nil if PoetryVersionResolver.path_dependency?(pyproject.content, dependency)
+        # NOTE: Prevent leaking potentially private path dependency names to PyPi
+        return nil if pyproject && PoetryVersionResolver.path_dependency?(pyproject.content, dependency)
 
         latest_version_finder.latest_version
       end
--- a/python/lib/dependabot/python/update_checker/poetry_version_resolver.rb
+++ b/python/lib/dependabot/python/update_checker/poetry_version_resolver.rb
@@ -44,6 +44,8 @@ module Dependabot
         end
 
         def latest_resolvable_version(requirement: nil)
+          return nil if PoetryVersionResolver.path_dependency?(pyproject.content, dependency)
+
           version_string =
             fetch_latest_resolvable_version_string(requirement: requirement)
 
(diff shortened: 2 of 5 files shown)


python/lib/dependabot/python/update_checker.rb
```diff
@@ -252,6 +252,8 @@ def updated_version_req_lower_bound
       end
 
       def fetch_latest_version
+        return nil if PoetryVersionResolver.path_dependency?(pyproject.content, dependency)
```
feelepxyz, line 255: Should this also check if the `resolver_type` is poetry and/or `pyproject` exists? Looks like this will be called for all python ecosystems where this file might not exist so would call `content` on nil.

Or what happens if we don't return early and rely on the resolver returning nil?

feelepxyz: > What happens if we don't return early and rely on the resolver returning nil?

Ah just spotted your comment about this.

feelepxyz, line 255:
```suggestion
        # NOTE: Prevent leaking potentially private path dependency names to PyPi
        return nil if pyproject && PoetryVersionResolver.path_dependency?(pyproject.content, dependency)
```

python/lib/dependabot/python/update_checker/poetry_version_resolver.rb
```diff
@@ -65,6 +67,21 @@ def resolvable?(version:)
           @resolvable[version] = false
         end
 
+        def self.path_dependency?(pyproject_content, dependency)
+          pyproject_object = TomlRB.parse(pyproject_content)
+          poetry_object = pyproject_object.dig("tool", "poetry")
+
+          %w(dependencies dev-dependencies).each do |type|
```
feelepxyz, line 74: Could we share this array as a constant between the two uses?

thepwagner:
I'm going to re-work and resubmit this.

I'd assumed skipping these dependencies in the `FileParser` wasn't viable (or expected that simplification in PR feedback).

But there's [existing code](https://github.com/dependabot/dependabot-core/blob/5042e9c39132b07d6253cfa531ef4f1514d35e9e/python/lib/dependabot/python/file_parser/poetry_files_parser.rb#L43) to skip git dependencies - leveraging that to skip path dependencies will be a cleaner one liner!

PR #3991, opened 2021-06-25
poetry: skip path and url dependencies

thepwagner:
This PR updates the python `FileParser` to ignore dependencies specified as a poetry [path dependency](https://python-poetry.org/docs/dependency-specification/#path-dependencies) or [url dependency](https://python-poetry.org/docs/dependency-specification/#url-dependencies), by following the existing pattern for ignoring `git` dependencies.

This is an alternative to https://github.com/dependabot/dependabot-core/pull/3987 , which would allow these dependencies to be parsed but not updated. Since they'll never be updated, it's easier just to skip them in the parse. It addresses the same concerns:
* When querying for the latest versions, available indexes were queried. This is a privacy concern: customers with monorepos will leak every submodule's name to PyPI. (In my test setup, my internal module that I named  `nested` was compared to https://pypi.org/project/Nested/ 😨 ).
* When invoking `poetry` to query for resolvable versions, dependabot [injects a `version` string to dependencies](https://github.com/dependabot/dependabot-core/blob/dce1b23793935cc0080729e90d8f3ba0bd52b5c3/python/lib/dependabot/python/update_checker/poetry_version_resolver.rb#L271-L275).

The latter causes a manifest like:
```toml
[tool.poetry.dependencies]
python = "^3.9"
nested = { path = "nested" }
```

To be re-written by Dependabot and fed to poetry as:
```toml
[tool.poetry.dependencies]
python = "^3.9"
nested = { path = "nested", version = ">= 0.1.0, <= 1.0.1" }
```

Which results in:
```
rsion_resolver.rb:88:in `block (2 levels) in fetch_latest_resolvable_version_string'
/home/dependabot/dependabot-core/python/lib/dependabot/python/update_checker/poetry_version_resolver.rb:338:in `run_poetry_command': RuntimeError (Dependabot::SharedHelpers::HelperSubprocessFailed)

  The Poetry configuration is invalid:
    - [dependencies.nested] {'path': 'nested', 'version': '>= 0.1.0, <= 1.0.1'} is not valid under any of the given schemas
  

  at /usr/local/.pyenv/versions/3.9.5/lib/python3.9/site-packages/poetry/core/factory.py:43 in create_poetry
       39│             message = ""
       40│             for error in check_result["errors"]:
       41│                 message += "  - {}\\n".format(error)
       42│ 
    →  43│             raise RuntimeError("The Poetry configuration is invalid:\\n" + message)
       44│ 
       45│         # Load package
       46│         name = local_config["name"]
       47│         version = local_config["version"]
```

### DIFF
--- a/python/lib/dependabot/python/file_parser/poetry_files_parser.rb
+++ b/python/lib/dependabot/python/file_parser/poetry_files_parser.rb
@@ -72,8 +72,7 @@ module Dependabot
           dependencies = Dependabot::FileParsers::Base::DependencySet.new
 
           parsed_lockfile.fetch("package", []).each do |details|
-            dep_type = details.dig("source", "type")
-            next if %w(git directory).include?(dep_type)
+            next if %w(directory git url).include?(details.dig("source", "type"))
 
             dependencies <<
               Dependency.new(
--- a/python/lib/dependabot/python/file_parser/poetry_files_parser.rb
+++ b/python/lib/dependabot/python/file_parser/poetry_files_parser.rb
@@ -15,6 +15,9 @@ module Dependabot
       class PoetryFilesParser
         POETRY_DEPENDENCY_TYPES = %w(dependencies dev-dependencies).freeze
 
+        # https://python-poetry.org/docs/dependency-specification/
+        UNSUPPORTED_DEPENDENCY_TYPES = %w(git path url).freeze
+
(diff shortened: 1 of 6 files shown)


feelepxyz (approved):
Nice one, this is a neat fix! LGTM 👍

PR #3998, opened 2021-06-28
PR Updater: Handle required status checks

jurre:
As per [the docs][docs] when branch protection rules are enabled, force
pushing to them is disabled.

This handles errors when this happens and raises a `BranchProtected`
error, which will allow us to handle this gracefully and inform the user
of this.

[docs]: https://docs.github.com/en/github/administering-a-repository/defining-the-mergeability-of-pull-requests/about-protected-branches#about-branch-protection-rules

### DIFF
--- a/common/lib/dependabot/pull_request_updater/github.rb
+++ b/common/lib/dependabot/pull_request_updater/github.rb
@@ -173,7 +173,8 @@ module Dependabot
 
         if e.message.match?(/protected branch/i) ||
            e.message.match?(/not authorized to push/i) ||
-           e.message.match?(/must not contain merge commits/)
+           e.message.match?(/must not contain merge commits/) ||
+           e.message.match?(/required status check/i)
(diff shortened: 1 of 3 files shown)


common/lib/dependabot/pull_request_updater/github.rb
```diff
@@ -173,7 +173,8 @@ def update_branch(commit)
 
         if e.message.match?(/protected branch/i) ||
            e.message.match?(/not authorized to push/i) ||
-           e.message.match?(/must not contain merge commits/)
+           e.message.match?(/must not contain merge commits/) ||
+           e.message.match?(/required status check/i)
```
jurre, line 177: There are a few variations of the error message, depending on wether there is 1 required check or more enabled, it didn't seem pertinent to test every variation, but that's why the check is not case sensitive

thepwagner (approved):

PR #4011, opened 2021-06-30
Fix for Azure client trying to parse 401 responses

wolf-cola:
Currently azure.rb does not have any logic to catch 401 error codes and will attempt to parse the response if one is encountered, causing auth exceptions for this client to be surfaced to the user as a JSON::ParserError.

This fix adds an Unauthorized exception to azure.rb and raises it when calls to Excon.get and Excon.post return a 401 status.


wolf-cola:
added commits from different account; abandoning and re-opening to avoid confusion

PR #4016, opened 2021-07-01
Gomod: Handle unrecognized import path error

jurre:
When referencing a go proxy that 404s, we see this error. Previously
this would not be handled, but it should result in a
DependencyFileNotResolvable error, so users have a sense of what's
wrong.


Nishnha (approved):

Nishnha:
I'm slightly confused about the wording in this PR.
Is it an unrecognizable Go `proxy` or Go `package` that 404s?

The solution looks correct either way, but I think Go proxies refer to [something else](https://proxy.golang.org/)?

jurre:
> I'm slightly confused about the wording in this PR.
> Is it an unrecognizable Go `proxy` or Go `package` that 404s?
> 
> The solution looks correct either way, but I think Go proxies refer to [something else](https://proxy.golang.org/)?

In cases I've found so far this has always been a proxy, pointing to a non-existing repo on GitHub will give a different error

PR #4029, opened 2021-07-06
Add Forbidden error type to Azure client

wolf-cola:
Adds a Forbidden error type to Dependabot:Clients:Azure to raise when a request returns 403

This fixes an issue where Dependabot:PullRequestCreator:Azure.create returns a 400 status to the caller when the root issue is a 403 being returned by Dependabot:PullRequestCreator:Azure.create_commit (since Azure.create creates both the commit and pull request it only ever returns the response code for the latter).


feelepxyz (approved):
Thanks for adding this 👍

PR #4044, opened 2021-07-13
Terraform: handle mixed case providers

feelepxyz:
Provider names can include uppercase characters but the resulting
lockfile downcases the name in the provider url.


brrygrdn (approved):
👍🏻  Good catch

Issue #4043, opened 2021-07-13
terraform: lock.hcl file not updated if provider version with '~>' constraint


mwieczorek:
**Package ecosystem**
 
terraform

**Package manager version**
 
terraform 1.0.2

**Language version**
 
N/A

**Manifest location and content prior to update**

**dependabot.yml content**

**Updated dependency**

**What you expected to see, versus what you actually saw**

I'd expect dependabot will update terraform.lock.hcl file, not (or not only) .tf file with provider config. 

**Native package manager behavior**

**Images of the diff or a link to the PR, issue or logs**

https://github.com/mwieczorek/dependabot-test/pull/2/files

**🕹 Bonus points: Smallest manifest that reproduces the issue**

PR #4048, opened 2021-07-14
Terraform: support updating local path modules

feelepxyz:
Add support for fetching and updating local path modules:

```
module "project" {
  source = "./project"
}
```

Terraform will resolve all local path modules when updating the
lockfile, so we need to make sure all requirements for the same
dependency have been updated, and written to the temporary directory
before we run `terraform providers lock _provider_`.

This also updates the file fetcher to not make an api request when
checking folder contents.

Fixes: https://github.com/dependabot/dependabot-core/issues/4043

### DIFF
--- a/terraform/lib/dependabot/terraform/file_fetcher.rb
+++ b/terraform/lib/dependabot/terraform/file_fetcher.rb
@@ -9,6 +9,9 @@ module Dependabot
     class FileFetcher < Dependabot::FileFetchers::Base
       include FileSelector
 
+      # https://www.terraform.io/docs/language/modules/sources.html#local-paths
+      LOCAL_PATH_SOURCE = %r{source\s*=\s*['"](?<path>..?\/[^'"]+)}.freeze
+
       def self.required_files_in?(filenames)
         filenames.any? { |f| f.end_with?(".tf", ".hcl") }
       end
@@ -23,6 +26,7 @@ module Dependabot
         fetched_files = []
         fetched_files += terraform_files
         fetched_files += terragrunt_files
+        fetched_files += local_path_module_files(terraform_files)
         fetched_files += [lock_file] if lock_file
 
         return fetched_files if fetched_files.any?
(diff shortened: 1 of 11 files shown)


terraform/spec/dependabot/terraform/file_updater_spec.rb
```diff
@@ -698,7 +698,7 @@ module "consul" {
           <<~DEP
             provider "registry.terraform.io/hashicorp/aws" {
               version     = "3.45.0"
-              constraints = ">= 3.37.0, < 3.46.0"
+              constraints = ">= 3.42.0, < 3.46.0"
```
feelepxyz, line 701: Lockfile constraints now mirror the updated provider constraints.

terraform/spec/dependabot/terraform/file_fetcher_spec.rb
```diff
@@ -14,68 +14,30 @@
       directory: directory
```
feelepxyz, line 14: Terraform is set up to always clone the repo so I've updated the specs to use a clone when fetching files

terraform/spec/fixtures/projects/provider_with_multiple_local_path_modules/loader/project/providers.tf
```diff
@@ -0,0 +1 @@
+../providers.tf
```
feelepxyz, line 1: This seems to be some kind of terraform magic reference, vscode also renders it as the referenced file, not sure how this works yet

feelepxyz: Looks like this when viewed in vscode 🤔

Screenshot 2021-07-14 at 15 12 46

terraform/spec/fixtures/projects/provider_with_multiple_local_path_modules/.terraform.lock.hcl
```diff
@@ -0,0 +1,20 @@
+# This file is maintained automatically by "terraform init".
+# Manual edits may be lost in future updates.
+
+provider "registry.terraform.io/mongey/confluentcloud" {
+  version     = "0.0.6"
+  constraints = "0.0.6"
```
Nishnha, line 6: You might run into regressions here if a new version is released
Adding version constraints solves this https://github.com/dependabot/dependabot-core/pull/3937/files#diff-3c5654bd9cc5ac3a940be7e9806a42b2378031143810aa1f632d53492ef76763R678

terraform/spec/fixtures/projects/provider_with_multiple_local_path_modules/loader/project/providers.tf
xlgmokha: magic

xlgmokha (approved):

terraform/lib/dependabot/terraform/file_fetcher.rb
```diff
@@ -47,6 +51,33 @@ def terragrunt_files
...
+            nested_terraform_files =
+              repo_contents(dir: base_path).
+              select { |f| f.type == "file" && f.name.end_with?(".tf") }.
+              map { |f| fetch_file_from_host(File.join(base_path, f.name)) }
+            terraform_files += nested_terraform_files
+            terraform_files += local_path_module_files(nested_terraform_files, dir: path)
+          end
+        end
+
+        terraform_files.tap { |fs| fs.each { |f| f.support_file = true } }
```
Nishnha, line 69: Are `support_files` handled differently?
I don't see how this value is used.

feelepxyz: Yeah we don't use it differently for terraform but we tag path dependency files this way for other ecosystems so I was only thinking of maintaining that data model. Will add a comment explaining this.

terraform/spec/fixtures/projects/provider_with_multiple_local_path_modules/.terraform.lock.hcl
feelepxyz: Oh will check this, i was thinking this would be fine as the constraint forces an exact version

PR #4050, opened 2021-07-15
Dry-run: cache cloned repos

feelepxyz:
Cache the cloned repo when `--cache=files` is passed to the dry-run
script.

This also allows setting up a test repo in `tmp/repo/name` and passing
`--clone` and `--cache=files` to load it for any package manager.

The codes in the dry-run scripts are not pretty but punting on a wider
refactor as this probably means re-writing it.

### DIFF
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -112,7 +112,7 @@ $options = {
   security_advisories: [],
   security_updates_only: false,
   ignore_conditions: [],
-  pull_request: false,
+  pull_request: false
 }
 
 unless ENV["LOCAL_GITHUB_ACCESS_TOKEN"].to_s.strip.empty?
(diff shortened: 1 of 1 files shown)


bin/dry-run.rb
```diff
@@ -476,8 +475,20 @@ def handle_dependabot_error(error:, dependency:)
...
 fetcher = Dependabot::FileFetchers.for_package_manager($package_manager).new(**fetcher_args)
-$files = if $options[:clone] || always_clone
+$files = if $repo_contents_path
+           if $options[:cache_steps].include?("files") && Dir.exist?($repo_contents_path)
+            puts "=> reading cloned repo from #{$repo_contents_path}"
+           else
+            puts "=> cloning into #{$repo_contents_path}"
+            FileUtils.rm_rf($repo_contents_path)
+           end
            fetcher.clone_repo_contents
```
feelepxyz, line 485: FYI: This method short circuits if the repo exists https://github.com/dependabot/dependabot-core/blob/40baefd1e812feca9712daa1e359583a29b5fdc1/common/lib/dependabot/file_fetchers/base.rb#L477

bin/dry-run.rb
```diff
@@ -476,8 +475,20 @@ def handle_dependabot_error(error:, dependency:)
...
+           if $options[:commit]
```
feelepxyz, line 486: Note: the branch is handled in `clone_repo_contents` but `commit` is a dry-run only option so handled here

jurre (approved):
My only concern is that bin/dry-run.rb keeps getting bigger and bigger and it's pretty hard to grasp what's going on

feelepxyz:
> My only concern is that bin/dry-run.rb keeps getting bigger and bigger and it's pretty hard to grasp what's going on

Yup good call out, would be worth trying to do a series of smaller refactors to split it up. Will see if I can find a good starting point.

PR #4053, opened 2021-07-15
Terraform: ignore symlinks in file fetcher

feelepxyz:
Fix a bug with symlinks being fetched in the file fetcher, resulting in
attempted updates to these files, which fail when creating a pull
request.

### DIFF
--- a/common/lib/dependabot/file_fetchers/base.rb
+++ b/common/lib/dependabot/file_fetchers/base.rb
@@ -233,10 +233,19 @@ module Dependabot
         Dir.entries(repo_path).map do |name|
           next if [".", ".."].include?(name)
 
+          absolute_path = File.join(repo_path, name)
+          type = if File.symlink?(absolute_path)
+                   "symlink"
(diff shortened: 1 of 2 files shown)


common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -233,10 +233,19 @@ def _cloned_repo_contents(relative_path)
...
           next if [".", ".."].include?(name)
 
+          absolute_path = File.join(repo_path, name)
+          type = if File.symlink?(absolute_path)
+                   "symlink"
+                 elsif Dir.exist?(absolute_path)
+                   "dir"
+                 else
+                   "file"
+                 end
```
feelepxyz, line 243: This private method is indirectly tested through the file fetcher spec, only `file` types are selected here https://github.com/dependabot/dependabot-core/blob/main/terraform/lib/dependabot/terraform/file_fetcher.rb#L62

This `type` mirrors what we get back from github.

jurre (approved):

PR #4056, opened 2021-07-16
Extend tests for Azure.create_pull_request(..., reviewers)

trejjam:
This PR adds support for reviewers in the create_pr.

Azure doc: https://docs.microsoft.com/en-us/rest/api/azure/devops/git/pull%20requests/create?view=azure-devops-rest-6.0#request-body

### DIFF
--- a/common/spec/dependabot/clients/azure_spec.rb
+++ b/common/spec/dependabot/clients/azure_spec.rb
@@ -41,7 +41,7 @@ RSpec.describe Dependabot::Clients::Azure do
   end
 
   describe "#fetch_commit" do
-    subject { client.fetch_commit(nil, branch) }
+    subject(:fetch_commit) { client.fetch_commit(nil, branch) }
 
     context "when response is 200" do
       before do
@@ -50,7 +50,7 @@ RSpec.describe Dependabot::Clients::Azure do
           to_return(status: 200, body: fixture("azure", "master_branch.json"))
       end
 
-      specify { expect { subject }.to_not raise_error }
+      specify { expect { fetch_commit }.to_not raise_error }
 
       it { is_expected.to eq("9c8376e9b2e943c2c72fac4b239876f377f0305a") }
     end
@@ -63,7 +63,7 @@ RSpec.describe Dependabot::Clients::Azure do
       end
 
       it "raises a helpful error" do
-        expect { subject }.to raise_error(Dependabot::Clients::Azure::NotFound)
+        expect { fetch_commit }.to raise_error(Dependabot::Clients::Azure::NotFound)
       end
     end
 
@@ -75,7 +75,7 @@ RSpec.describe Dependabot::Clients::Azure do
       end
 
       it "raises a helpful error" do
-        expect { subject }.to raise_error(Dependabot::Clients::Azure::Forbidden)
+        expect { fetch_commit }.to raise_error(Dependabot::Clients::Azure::Forbidden)
       end
     end
 
@@ -87,7 +87,7 @@ RSpec.describe Dependabot::Clients::Azure do
       end
 
       it "raises a helpful error" do
-        expect { subject }.to raise_error(Dependabot::Clients::Azure::Unauthorized)
+        expect { fetch_commit }.to raise_error(Dependabot::Clients::Azure::Unauthorized)
       end
     end
 
@@ -99,7 +99,7 @@ RSpec.describe Dependabot::Clients::Azure do
       end
 
       it "raises a helpful error" do
-        expect { subject }.to raise_error(Dependabot::Clients::Azure::NotFound)
+        expect { fetch_commit }.to raise_error(Dependabot::Clients::Azure::NotFound)
       end
     end
   end
@@ -129,7 +129,7 @@ RSpec.describe Dependabot::Clients::Azure do
       end
 
       it "raises a helpful error" do
-        expect { subject }.to raise_error(Dependabot::Clients::Azure::Forbidden)
+        expect { create_commit }.to raise_error(Dependabot::Clients::Azure::Forbidden)
       end
     end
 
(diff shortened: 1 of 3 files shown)


jurre:
Thanks for the contribution!

I'm going to trigger CI, but I notice some things that'll definitely hit the linter, and I expect most tests for this class will now fail.

Speaking of tests, it'd be great to add some test cases for this that show how it works, it wasn't immediately clear to be from the docs what structure we need to pass to `reviewers`, currently it'll pass an array of hashes (objects in JSON) with an `id` key, is that right?

trejjam:
Executing tests on Windows 10 was not really smooth ... But I manage it (with a few fixes)

I did add tests for creating the Azure PR

jurre:
> Executing tests on Windows 10 was not really smooth

Yeah I'd recommend using the docker dev environment to run tests, there are quite a few things that need to be set up in order to run it outside of docker. I personally always run inside the docker env

> I did add tests for creating the Azure PR

Thanks!

bin/docker-dev-shell
```diff
@@ -71,6 +72,11 @@ if [ -n "$RUNNING" ]; then
   fi
 fi
 
+if [ -z "$PROJECT_LOCATION" ]; then
+  # a helper for window subsystem for linux
+  PROJECT_LOCATION=$(pwd)
```
jurre, line 77: I'm confused why we need to set `PROJECT_LOCATION=$(pwd)` again here, when it's already set on line 5? Why can't we directly use `$(pwd)` on windows, but we _can_ set assign it to `PROJECT_LOCATION`?

common/lib/dependabot/clients/azure.rb
```diff
@@ -170,7 +170,7 @@ def create_commit(branch_name, base_commit, commit_message, files,
 
       # rubocop:disable Metrics/ParameterLists
       def create_pull_request(pr_name, source_branch, target_branch,
-                              pr_description, labels, work_item = nil)
+                              pr_description, labels, work_item = nil, reviewers)
```
jurre, line 173: I assume the linter will complain about this, but the optional argument (`work_item = nil`) should go last in the argument list, this ensures that we don't accidentally pass the wrong value when leaving it out: https://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Style/OptionalArguments

bin/docker-dev-shell
trejjam: Thanks, I forgot the 5th line

common/lib/dependabot/clients/azure.rb
trejjam: Make sense

common/spec/dependabot/clients/azure_spec.rb
```diff
@@ -181,6 +181,107 @@
...
+    let(:work_item) { nil }
+    let(:create_pull_request_url) do
+      base_url + "/_apis/git/repositories/#{source.unscoped_repo}/pullrequests?api-version=5.0"
+    end
+
+    context "when response is 200" do
+      response_body = fixture("azure", "create_pull_request_details.json")
+
+      before do
+        @request_body = "{\"sourceRefName\":\"refs/heads/#{source_branch}\","
```
jurre, line 207: I don't think this needs to be an instance variable, it's only referenced in this block, so we can just make it local to the block (drop the `@`)

common/spec/dependabot/clients/azure_spec.rb
```diff
@@ -181,6 +181,107 @@
...
+        + "\"targetRefName\":\"refs/heads/#{target_branch}\",\"title\":\"#{pr_name}\","
+        + "\"description\":\"#{pr_description}\",\"labels\":[],\"workItemRefs\":[],\"reviewers\":[]}"
+
+        stub_request(:post, create_pull_request_url).
+          with(basic_auth: [username, password], body: @request_body).
+          to_return(status: 200, body: response_body)
+      end
+
+      specify { expect { subject }.to_not raise_error }
```
jurre, line 216: We already implicitly test this doesn't raise an error in the next test, so I think it's safe to drop this 👍

common/spec/dependabot/clients/azure_spec.rb
```diff
@@ -181,6 +181,107 @@
     end
   end
 
+  describe "#create_pull_request" do
+    subject do
```
jurre, line 185: It'd be nice if we gave this subject a name, so we can reference it in the tests:

```suggestion
    subject(:create_pull_request) do
```

common/spec/dependabot/clients/azure_spec.rb
```diff
@@ -181,6 +181,107 @@
...
+        + "\"description\":\"#{pr_description}\",\"labels\":[],\"workItemRefs\":[],\"reviewers\":[]}"
+
+        stub_request(:post, create_pull_request_url).
+          with(basic_auth: [username, password], body: @request_body).
+          to_return(status: 200, body: response_body)
+      end
+
+      specify { expect { subject }.to_not raise_error }
+
+      it { is_expected.to eq(JSON.parse(response_body)) }
```
jurre, line 218: It'd be nice to give this expectation a descriptive name:

```suggestion
      it "returns the parsed response" do
        expect(create_pull_request).to eq(JSON.parse(response_body)
      end
```

But also, looking at this closer, currently this doesn't test a whole lot 🤔 we just test that the mocked response is returned.

Might be better to follow the pattern seen here: https://github.com/dependabot/dependabot-core/pull/4056/files#diff-e13fc24cac609b0c56689577fb64b4dc27386d0ee684b58c1e2a32700e64dc6aR169-R178 and verify the request we made is right?

trejjam: I think it tests what is necessary. The mock returns the expected response only when the request matches the request_body. And it tests that the response body is parsed.

jeffwidman:
@trejjam for this one, two requests:
1. Can you rebase to fix merge conflicts?
2. Can you split out the Windows-compatibility fixes as a separate PR? It's entirely unrelated to the logic change for adding Azure reviewers... To be clear, I'm not opposed to them, but they shouldn't be shoehorned into this PR.

common/lib/dependabot/pull_request_creator.rb
```diff
@@ -176,6 +176,7 @@ def azure_creator
         pr_name: message.pr_name,
         author_details: author_details,
         labeler: labeler,
+        reviewers: reviewers || [],
```
jeffwidman, line 179: This was the only thing that gave me pause... IMO we need to do one of two things:
1. decide we want to support reviewers as a top-level arg across all the clients that support it... planning to eventually also add plumbing support for it in github/gitlab/bitbucket even if they're no-ops for now.
2. Not let this leak from Azure upwards... not even sure this is possible with the current code layout. But if we're only doing it for Azure, I'd rather not add an arg globally.

At first glance I think option 1 is the better route to go, but I'm going to reach out to a teammate to get a second opinion. In the meantime, no need to modify this PR, just leave as-is for now.

common/lib/dependabot/clients/azure.rb
```diff
@@ -172,7 +172,7 @@ def create_commit(branch_name, base_commit, commit_message, files,
 
       # rubocop:disable Metrics/ParameterLists
       def create_pull_request(pr_name, source_branch, target_branch,
-                              pr_description, labels, work_item = nil)
+                              pr_description, labels, reviewers, work_item = nil)
```
jeffwidman, line 175: Note to self: We could make this an optional arg instead of required here... depends whether we make `reviewers` a global required arg vs a global optional or local to azure arg.

I want to noodle on this API a little...

jeffwidman:
Thanks for the fast rebase and splitting out as separate commits. Give me a little bit of time to think on the API design.

jeffwidman:
After noodling on this more, "requested reviewers" is a pretty ubiquitous concept... I checked GitHub, BitBucket, GitLab, and they all support it. So I'm 👍 on the path this PR took of adding `reviewers` this as an optional arg at the top level... with the expectation that over time, other folks will probably submit PR's to plumb this through for other platforms...

@trejjam can you fix merge conflicts?

The code in this PR currently all looks good to me, so unless another maintainer feels differently, once those conflicts are fixed this should be good to merge.

trejjam:
Duplicate #5988

trejjam:
In the meantime, someone else implemented the feature...

So this simply becomes tests extension...

trejjam:
And the tests are failing, and I do not have the will to fix them

jeffwidman:
Oh man, I'm so sorry! I completely forgot about #5988, my apologies! 

And when looking at that PR, we should have pursued this one first... but we just weren't caught up on our backlog so completely overlooked it.

Again, my sincere apologies.

PR #4059, opened 2021-07-18
bundler: optimize gemfile parsing

skipkayhil:
# Summary
After profiling `FileParser#parse` for bundler, I found that the ruby
Parser was extremely hot. The same Gemfile was being parsed twice for
each dependency, when it only needed to be parsed a single time per
file.

The performance improvement depends on the size of the Gemfile being
parsed, but the following example shows the speedup for the
[discourse/discourse](https://github.com/discourse/discourse) repo.

Before:
```
Warming up --------------------------------------
               parse     1.000  i/100ms
Calculating -------------------------------------
               parse      0.121  (± 0.0%) i/s -   1.000  in   8.262804s
```

After:
```
Warming up --------------------------------------
               parse     1.000  i/100ms
Calculating -------------------------------------
               parse      6.319  (±15.8%) i/s -  32.000  in   5.099269s
```

# Other Information

I wasn't sure what the policy around breaking API changes was, especially for a relatively internal class like `GemfileDeclarationFinder`. Any feedback would be appreciated, I believe I can accomplish something similar without breaking the method signatures if necessary.

## Benchmark Script
```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "dependabot-common", path: "./common"
  gem "dependabot-bundler", path: "./bundler"
  gem "benchmark-ips"
  # This was helping some DNS issues I was having but is optional
  gem "resolv-replace"
end

package_manager = "bundler"
repo_name = "discourse/discourse"

credentials = [
  {
    "type" => "git_source",
    "host" => "github.com",
    "password" => ENV["GITHUB_ACCESS_TOKEN"] # A GitHub access token with read access to public repos
  }
]

source = Dependabot::Source.new(
  provider: "github",
  repo: repo_name,
  directory: "/",
)

puts "Fetching #{package_manager} dependency files for #{repo_name}"
fetcher = Dependabot::FileFetchers.for_package_manager(package_manager).new(
  source: source,
  credentials: credentials,
)

puts "Parsing dependencies information"
parser = Dependabot::FileParsers.for_package_manager(package_manager).new(
  dependency_files: fetcher.files,
  source: source,
  credentials: credentials,
)

Benchmark.ips do |x|
  x.report("parse") { parser.parse }
  x.compare!
end
```

### DIFF
--- a/bundler/lib/dependabot/bundler/file_parser/gemfile_declaration_finder.rb
+++ b/bundler/lib/dependabot/bundler/file_parser/gemfile_declaration_finder.rb
@@ -8,20 +8,20 @@ module Dependabot
     class FileParser
       # Checks whether a dependency is declared in a Gemfile
       class GemfileDeclarationFinder
-        def initialize(dependency:, gemfile:)
-          @dependency = dependency
+        def initialize(gemfile:)
           @gemfile = gemfile
+          @declaration_nodes = {}
         end
 
-        def gemfile_includes_dependency?
-          !declaration_node.nil?
+        def gemfile_includes_dependency?(dependency)
+          !declaration_node(dependency).nil?
         end
 
-        def enhanced_req_string
-          return unless gemfile_includes_dependency?
+        def enhanced_req_string(dependency)
+          return unless gemfile_includes_dependency?(dependency)
 
           fallback_string = dependency.fetch("requirement")
-          req_nodes = declaration_node.children[3..-1]
+          req_nodes = declaration_node(dependency).children[3..-1]
           req_nodes = req_nodes.reject { |child| child.type == :hash }
 
           return fallback_string if req_nodes.none?
@@ -39,31 +39,35 @@ module Dependabot
 
         private
 
-        attr_reader :dependency, :gemfile
+        attr_reader :gemfile
 
-        def declaration_node
-          return @declaration_node if defined?(@declaration_node)
-          return unless Parser::CurrentRuby.parse(gemfile.content)
-
-          @declaration_node = nil
-          Parser::CurrentRuby.parse(gemfile.content).children.any? do |node|
-            @declaration_node = deep_search_for_gem(node)
-          end
-          @declaration_node
+        def parsed_gemfile
+          @parsed_gemfile ||= Parser::CurrentRuby.parse(gemfile.content)
         end
 
-        def deep_search_for_gem(node)
-          return node if declares_targeted_gem?(node)
+        def declaration_node(dependency)
+          return @declaration_nodes[dependency] if @declaration_nodes.key?(dependency)
+          return unless parsed_gemfile
+
+          @declaration_nodes[dependency] = nil
+          parsed_gemfile.children.any? do |node|
+            @declaration_nodes[dependency] = deep_search_for_gem(node, dependency)
+          end
+          @declaration_nodes[dependency]
+        end
+
+        def deep_search_for_gem(node, dependency)
+          return node if declares_targeted_gem?(node, dependency)
           return unless node.is_a?(Parser::AST::Node)
 
           declaration_node = nil
           node.children.find do |child_node|
-            declaration_node = deep_search_for_gem(child_node)
+            declaration_node = deep_search_for_gem(child_node, dependency)
           end
           declaration_node
         end
 
-        def declares_targeted_gem?(node)
+        def declares_targeted_gem?(node, dependency)
           return false unless node.is_a?(Parser::AST::Node)
           return false unless node.children[1] == :gem
 
--- a/bundler/lib/dependabot/bundler/file_parser.rb
+++ b/bundler/lib/dependabot/bundler/file_parser.rb
@@ -60,17 +60,17 @@ module Dependabot
         return dependencies unless gemfile
 
         [gemfile, *evaled_gemfiles].each do |file|
+          gemfile_declaration_finder = GemfileDeclarationFinder.new(gemfile: file)
+
           parsed_gemfile.each do |dep|
-            gemfile_declaration_finder =
-              GemfileDeclarationFinder.new(dependency: dep, gemfile: file)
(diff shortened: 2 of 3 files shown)


bundler/lib/dependabot/bundler/file_parser/gemfile_declaration_finder.rb
```diff
@@ -8,20 +8,20 @@ module Bundler
     class FileParser
       # Checks whether a dependency is declared in a Gemfile
       class GemfileDeclarationFinder
-        def initialize(dependency:, gemfile:)
-          @dependency = dependency
```
Nishnha, line 12: @dependency is used by a Dependabot runner internally, and unfortunately removing it would break much of how Dependabot runs.

bundler/lib/dependabot/bundler/file_parser/gemfile_declaration_finder.rb
```diff
@@ -8,20 +8,20 @@ module Bundler
...
         end
 
-        def enhanced_req_string
-          return unless gemfile_includes_dependency?
+        def enhanced_req_string(dependency)
+          return unless gemfile_includes_dependency?(dependency)
 
           fallback_string = dependency.fetch("requirement")
-          req_nodes = declaration_node.children[3..-1]
+          req_nodes = declaration_node(dependency).children[3..-1]
```
Nishnha, line 24: Dependencies are fed into FileParser one-at-a-time by the FileFetcher with the @dependency variable, so you shouldn't have to pass dependency as a parameter to the functions

Nishnha:
Hello @SkipKayhil,

Thank you for this contribution! 
Indeed, many of Dependabot Core's `FileParsers` read the same dependencies multiple times.

> I wasn't sure what the policy around breaking API changes was, especially for a relatively internal class like `GemfileDeclarationFinder`

Generally we try to avoid breaking the existing API. I've made a couple of comments in the code pointing out breaking changes (e.g. removing the @dependency variable).

Have you been able to run the test suite to make sure these changes pass? There are instruction in the README on how to run the test suite.
Once you have passing tests we can run your PR against the CI 😄 

These speedups are very welcome since they have an even bigger effect at scale! Let's try to find a solution that doesn't break the existing API together.

P.S., the Dependabot team is currently short-handed so we may take a while to respond.

bundler/lib/dependabot/bundler/file_parser/gemfile_declaration_finder.rb
feelepxyz: Looks like `GemfileDeclarationFinder` is only used once from the file parser so might be ok changing this.

feelepxyz:
Thanks! Following on from @Nishnha comment. I don't this change would break any APIs as we don't expose this module externally and don't think anyone is relying on it directly.

I also think the interface change makes sense given to goal of only parsing a gemfile once, as there are usually multiple dependencies per file 👍 

I've kicked off CI, new contributions need the actions run approved.

jurre:
Thanks for looking into this @SkipKayhil, I think it'd be good to instrument this in a full dry-run rather than in isolation, as the parsing step only happens once during an entire update, so even though it's not very fast in isolation it might not make a huge difference overall. That's not to say this isn't worth doing, would just good to get a sense of the overall impact

SkipKayhil:
> 
> 
> Thanks for looking into this @SkipKayhil, I think it'd be good to instrument this in a full dry-run rather than in isolation, as the parsing step only happens once during an entire update, so even though it's not very fast in isolation it might not make a huge difference overall. That's not to say this isn't worth doing, would just good to get a sense of the overall impact

That's a pretty good point, I hadn't thought about the other steps. So far I've been using dependabot as a library to gather dependency information but haven't yet used it to update things. In that scenario this change makes a pretty big impact, but after running the dry run script it looks like it doesn't end up being as much of an improvement in the whole process. I'll probably be looking into the `UpdateChecker` process in the future because that seems to be where the bulk of time is spent.

On my (not so powerful) laptop:
```bash
# main
$ time bin/dry-run.rb bundler discourse/discourse
...
real    43m9.203s
user    41m10.464s
sys     1m47.502s
```

And swapping out `Benchmark.ips` for `Benchmark.measure` in my original script:
```bash
# main
$ ruby benchmark_script.rb
14.032599   0.035577  14.623058 ( 14.636967)

# skipkayhil:gemfile-parse-optimization
$ ruby benchmark_script.rb
0.395320   0.000374   0.901548 (  0.899722)
```

jurre:
>  I'll probably be looking into the UpdateChecker process in the future because that seems to be where the bulk of time is spent.

Nice 👍 I think most of the time is spent doing networking, but if we can find any optimizations there, that'd be fantastic!

bundler/lib/dependabot/bundler/file_parser/gemfile_declaration_finder.rb
jeffwidman: I assume this code is now public with https://github.com/dependabot/dependabot-core/pull/5608? In that case, should be easier to identify the downstream impacts of this change...

jeffwidman:
@skipkayhil thanks again for putting up this PR... are you still interested in driving it to completion? Given the millions of jobs that we run, we're always interested in perf improvements. 👍

skipkayhil:
> are you still interested in driving it to completion?

Would definitely still love to see this merged! I just gave it a rebase and will update if anything fails CI

jurre (approved):

Nishnha (approved):

Nishnha:
I had some initial concerns with merging this PR incase it changed some underlying behavior in how we parse dependencies.

Seeing this PR is passing our e2e smoke tests, which tests for consistent output, is reassuring!

Thanks for the massive speed up and benchmark script @skipkayhil - definitely going to try this out on our other ecosystems!

PR #4061, opened 2021-07-19
npm: prevent scoped registries becoming global

feelepxyz:
Fixes a bug where all packages (public and private) are forced through a
private registry, where it's only intended to be used for scoped
packages so fails for public packages hosted on npm.

Added a check to the method that attempts to extract a global registry,
checking if credentials registry host has already been configured in
`.npmrc` as a scoped registry.

This prevents dependabot from re-writing the `.npmrc` global `registry`
with the scoped registry:

```
registry=registry.npmjs.org
@Private=private.example.com
registry=private.example.com
auth..
```

Keeping the global and scoped config intact:

```
registry=registry.npmjs.org
@Private=private.example.com
auth..
```


npm_and_yarn/spec/fixtures/projects/yarn/scoped_private_source_with_npmrc copy/package.json
```diff
@@ -0,0 +1,30 @@
+{
```
brrygrdn, line 1: Unintentional check-in?

npm_and_yarn/spec/fixtures/projects/yarn/scoped_private_source_with_npmrc copy/yarn.lock
```diff
@@ -0,0 +1,39 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
brrygrdn, line 1: Same as above?

brrygrdn (changes requested):
Change looks good to me, one question about the `foo copy` files I think might be cruft, otherwise 👍🏻

npm_and_yarn/spec/fixtures/projects/yarn/scoped_private_source_with_npmrc copy/package.json
feelepxyz: Yep good catch!

jurre (approved):
Thanks for tackling this

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npmrc_builder.rb
```diff
@@ -150,6 +153,16 @@ def credential_lines_for_npmrc
           ["always-auth = true"] + lines
         end
 
+        def npmrc_scoped_registries
+          return [] unless npmrc_file
+
+          @npmrc_scoped_registries ||=
+            npmrc_file.content.lines.select { |line| line.match?(/^\s*@/) && line.include?(":registry") }.
+            map do |line|
+              line.match(/:registry\s*=\s*(?<registry>\S+)/)&.named_captures&.fetch("registry")
```
jurre, line 162: We could maybe extract a constant for the regex here, but not blocking for review

jurre (approved):

brrygrdn (approved):

PR #4071, opened 2021-07-21
Draft: Added support for bitbucket-server

oleg-andreyev:
Yet another Bitbucket Server (on-prem) implementation.


common/lib/dependabot/clients/bitbucket.rb
```diff
@@ -19,140 +19,266 @@ class Forbidden < StandardError; end
...
         @credentials = credentials
         @auth_header = auth_header_for(credentials&.fetch("token", nil))
       end
 
       def fetch_commit(repo, branch)
-        path = "#{repo}/refs/branches/#{branch}"
-        response = get(base_url + path)
-
-        JSON.parse(response.body).fetch("target").fetch("hash")
+        if @source.provider == "bitbucket_server"
```
jurre, line 39: Given that most of these methods include this conditional, I think it makes more sense to extract a separate class for bitbucket server?

jurre (changes requested):
Thanks for contributing to dependabot!

We won't be able to actively maintain this implementation, but I'm OK with merging it if it doesn't break API compatibility. 

We'll need to have sufficient test coverage for this new functionality though, and from the looks of it I think this server implementation should live in its own class, rather than have a conditional in each method. 

I'll trigger CI on the PR so far, but I think it needs some more work.

common/lib/dependabot/clients/bitbucket.rb
oleg-andreyev: @jurre I've initially implemented it as a separate class, but later saw a TODO

> TODO: When self-hosted Bitbucket is supported this should use `Bitbucket.for_source`

and decided well if you want to have it in a single class okay

oleg-andreyev:
> I'll trigger CI on the PR so far, but I think it needs some more work.

@jurre thanks for such a fast response, I'm in the middle of testing dependabot locally on our BitBucket Server and so far it's working
image
image

common/lib/dependabot/clients/bitbucket.rb
```diff
@@ -12,38 +12,23 @@ class Unauthorized < StandardError; end
...
-      #######################
-
-      def self.for_source(source:, credentials:)
-        credential =
-          credentials.
-          select { |cred| cred["type"] == "git_source" }.
-          find { |cred| cred["host"] == source.hostname }
-
-        new(credentials: credential)
-      end
```
jurre, line 26: I'm slightly worried that removing this might break folks using this method, can we leave it in?

oleg-andreyev: will return.

bin/bump-version.rb
```diff
@@ -7,12 +7,12 @@
 end
 component = ARGV[0].to_sym
 
-unless `which gh` && $?.success?
+unless `which gh` && $CHILD_STATUS.success?
```
oleg-andreyev, line 10: rubocop

jurre: I think we can leave this change out of the PR as it's unrelated, and the scripts are currently not checked by rubocop on CI

common/lib/dependabot/clients/bitbucket.rb
```diff
@@ -139,8 +138,8 @@ def create_pull_request(repo, pr_name, source_branch, target_branch,
         pr_path = "#{repo}/pullrequests"
         post(base_url + pr_path, content.to_json)
       end
-      # rubocop:enable Metrics/ParameterLists
 
+      # rubocop:enable Metrics/ParameterLists
```
jurre, line 142: This should go back under the `end`, it closes out the disable pragma from line 119 and wraps that method

jurre:
This is starting to look good, but we'll need decent automated test coverage for all this new functionality and for the new class before we can consider pulling this in

monwolf:
@oleg-andreyev I'm interested in seeing this PR merged, are you planning to add the coverage requested?

oleg-andreyev:
@monwolf I'm quite busy at the moment, but any help is be appreciated.

monwolf:
> @monwolf I'm quite busy at the moment, but any help is be appreciated.

I could try to do somethimg but I don't know anything about Ruby. 

Maybe some contributor could help/guide me on how to start this

oleg-andreyev:
> > @monwolf I'm quite busy at the moment, but any help is be appreciated.
> 
> I could try to do somethimg but I don't know anything about Ruby.
> 
> Maybe some contributor could help/guide me on how to start this

Lol, I also don't have any experience with Ruby :)

jurre:
Copy/pasting my reply from #1450, because the same thing applies.

> Hi all, I appreciate the work that went into this PR and this definitely warrants a reply from the maintainers.
> 
> However, at this point I don't think that we can accept this patch into dependabot-core natively, as we don't have the capacity or expertise to provide support for Bitbucket. This is currently already a problem with the existing Azure DevOps and GitLab clients, where we're not positioned well to provide support and rely on other members in the community to do so. However, not having clear maintainers for those parts of the codebase makes this hard.
> 
> Where I would like to see support for (non GitHub) source control providers go in the future, is for them to be stand-alone gems that you can pull in next to dependabot-core. dependabot-core would provide a clear interface for those gems to implement. These gems can then be maintained by people who actually know and use these providers.
> 
> I think this should currently already be possible, but it might rely on monkey-patching some of the existing dependabot-core classes, which is not ideal. I'd be more than happy to accept patches that improve the interface for external providers if you decide to go down that route.

I know that I encouraged adding test coverage to this implementation in order to get it merged, but I don't believe that the dependabot-core maintainers are the right people to maintain this implementation.

I would recommend implementing this as a standalone gem that can be pulled in next to dependabot-core. Most of the code and tests in this PR should still be applicable for that approach and I am happy to help land any changes to make such a gem easier to implement.

mieliespoor:
@jurre has then been any progress since this issue was closed to get us closer to moving the bitbucket server support into it's own gem? I'm no expert in Ruby, but would most certainly like to see the changes of this PR become reality, even if through a new gem alongside dependabot-core.

jeffwidman:
@mieliespoor I think it might be as simple as extending the generic wrapper script: https://github.com/dependabot/dependabot-script/pull/675

Not sure, but any further discussions should probably happen in that `dependabot-script` repo and not here as they're unlikely to be seen here.

Issue #2547, opened 2020-09-23
Bumps kotlinx-coroutines-core from 1.3.1 to 1.3.9-native-mt-2.


ColtonIdle:
Installed dependabot and I keep getting this PR in my Android project that uses kotlin coroutines.

`Bumps kotlinx-coroutines-core from 1.3.1 to 1.3.9-native-mt-2.`

**Package manager/ecosystem**
gradle/maven

**What you expected to see, versus what you actually saw**
It would have been nice for this PR to be to 1.3.9 directly instead of this experimental native-mt-2 dep from kotlin.

jurre:
Hmm, yeah it seems like our version resolver doesn't correctly recognize that as a prerelease, I'm not super familiar with Gradle prerelease/experimental releases. Is it common to append a tag like that? We may be comparing lexically when we shouldn't.

ColtonIdle:
@jurre I don't know. Maybe this is a one off thing that is just best for me to skip this.

I asked about this on kotlin slack since I've never heard of native-mt-2 before and this is what someone said

This user https://github.com/gildor said this:

This is a special branch with developer preview of support multithreading coroutines for Kotlin native
if you are not interested in multithreading on native, use standard 1.3.9
More details https://github.com/Kotlin/kotlinx.coroutines/issues/462

gildor:
> Is it common to append a tag like that

I wouldn't say that it's very widespread, but definitely not unique

I think depedabot shouldn't suggest update if 1.3.9 and 1.3.9-something versions are available to update on 1.3.9-something from 1.3.8

Fryie:
This is unfortunately rather annoying and could also lead to someone who's not paying close attention (or has auto-merge enabled for patch releases) accidentally merging this.

ashughes:
I keep running into this as well. Can [`ignore`](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates#ignore) be used as a workaround? I'm not sure what a proper value for `versions` would be such that it means *ignore all versions ending with \*-native-mt\**.

I also have the added complexity that I'm using a `def coroutinesVersion` in my `build.gradle` to define the coroutine version once for multiple artifacts, so even with the correct `versions` above, I'm not entirely sure whether the `dependency-name` should be `coroutinesVersion` or `org.jetbrains.kotlinx:kotlinx-coroutines-*` or something else. Any guidance there would be appreciated!

hiqua:
[...]

PR #4077, opened 2021-07-23
Make native-mt and agp version types for Maven and Gradle

greysteil:
Fixes https://github.com/dependabot/dependabot-core/issues/2547.

Version comparison in Dependabot uses the [Maven specification](https://maven.apache.org/pom.html#Version_Order_Specification) when looking at Maven / Gradle versions. Whilst technically correct, this can cause problems where a popular dependency has chosen to append their own suffix onto their versions to distinguish between types. The problems are [not unique to Dependabot](https://issuetracker.google.com/issues/171369798?pli=1#comment9).

The pragmatic solution is to add some special cases to our version fetching logic. We already treat `jre`, `android` and `java` as "types" when they appear in a version string - this PR extends that logic to do the same for `native-mt` and `agp` (short for "Android Gradle Plugin" and [referenced here](https://github.com/Triple-T/gradle-play-publisher/releases/tag/3.5.0)).

Note: the `native-mt` type is a particular pain, as it includes a token separator character (the `-`). I've added extra special casing to deal with that, and a test for it.


greysteil:
@hiqua FYI

xlgmokha (approved):

hiqua:
@greysteil nice, thanks a lot!

I think this doesn't cover this case though:

* https://github.com/iSoron/uhabits/pull/745

or does it?

It's unfortunate, since, from my reading of https://semver.org/spec/v2.0.0.html#spec-item-9, this should be considered a pre-release. From my limited experience, it would be better to default to considering these x.y.z-$SUFFIX as prereleases, and whitelist a few of these $SUFFIX that could mean something else (like a specific flavor of the library).

greysteil:
It won't fix that, no, but it's not clear to me that it should - as far as I'm aware the Maven ecosystem uses [this specification](https://maven.apache.org/pom.html#Version_Order_Specification) for version comparison, **not** SemVer 2.0, and I presume Gradle does the same for consistency. Dependabot would be doing the wrong thing in even more cases if it followed SemVer 2.0 for Maven/Gradle.

(FWIW, here's the [version comparison logic Dependabot uses for Gradle](https://github.com/dependabot/dependabot-core/blob/main/gradle/lib/dependabot/gradle/version.rb. We have different version comparison logic for each language because they're all different - here's [python's](https://github.com/dependabot/dependabot-core/blob/main/python/lib/dependabot/python/version.rb) and here's [npm's](https://github.com/dependabot/dependabot-core/blob/main/npm_and_yarn/lib/dependabot/npm_and_yarn/version.rb) (the easy one, as it's just SemVer - they wrote the SemVer spec).)

Are you using other tooling that sort `1.0` ahead of `1.0-20050927.133100`? I don't use Maven or Gradle myself, so am happy to be corrected, but my hunch here is that in this case it's the package author who has made a mistake, releasing alphas / betas with versions that in fact treat them as full releases.

hiqua:
> It won't fix that, no, but it's not clear to me that it should - as far as I'm aware the Maven ecosystem uses this specification for version comparison, not SemVer 2.0, and I presume Gradle does the same for consistency. Dependabot would be doing the wrong thing in even more cases if it followed SemVer 2.0 for Maven/Gradle.

Unless I'm misreading it, your link indicates that Maven should mostly follow Semver 1.0.0 (which seems similar for these cases: https://semver.org/spec/v1.0.0.html#spec-item-4): 

> If version strings are syntactically correct Semantic Versioning 1.0.0 version numbers, then in almost all cases version comparison follows the precedence rules outlined in that specification. These versions are the commonly encountered alphanumeric ASCII strings such as 2.15.2-alpha.

This example (`2.15.2-alpha`) seems to indicate that they indeed treat a `-alpha` version as a prerelease.

> Are you using other tooling that sort 1.0 ahead of 1.0-20050927.133100?

Android Studio (and I guess vanilla Jetbrains' IDEs) correctly suggest to upgrade from the prerelease to the release for this example (and not the other way around):

2021-07-26-074631_1400x144_scrot

Of course they could have a specific exception for this library but that's unlikely.

PR #4081, opened 2021-07-24
dry-run: earlier next when dependency up to date

SkipKayhil:
# Summary

The dry run script currently does a lot of unnecessary work checking
vulnerability status and checking sub dependencies when a dependency is
already on the latest version. This commit moves the check for a
dependency being up to date immediately after fetching the latest
version so that the unnecessary work is skipped.

This also aligns the behavior with dependabot-script:
https://github.com/dependabot/dependabot-script/blob/4330ff7043b6fe2bb009005e2f5b0ca9985f32f2/generic-update-script.rb#L179

Before:
```
$ time bin/dry-run.rb bundler discourse/discourse
real    43m9.203s
user    41m10.464s
sys     1m47.502s
```

After:
```
$ time bin/dry-run.rb bundler discourse/discourse
real    22m48.066s
user    21m36.573s
sys     0m52.919s
```

### DIFF
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -613,6 +613,11 @@ dependencies.each do |dep|
   puts " => checking for updates #{checker_count}/#{dependencies.count}"
   puts " => latest available version is #{checker.latest_version}"
 
+  if checker.up_to_date?
+    puts "    (no update needed as it's already up-to-date)"
+    next
+  end
+
   if $options[:security_updates_only] && !checker.vulnerable?
     if checker.version_class.correct?(checker.dependency.version)
       puts "    (no security update needed as it's not vulnerable)"
@@ -648,11 +653,6 @@ dependencies.each do |dep|
     conflicting_dependencies.each do |conflicting_dep|
       puts "   #{conflicting_dep['explanation']}"
     end
-  end
-
-  if checker.up_to_date?
-    puts "    (no update needed as it's already up-to-date)"
-    next
   end
 
   requirements_to_unlock =


jurre:
I think some cases would now be missed, like:

```
puts " => there is no available non-vulnerable version"
```

This is valuable information for how we run Dependabot at GitHub, because if this is the case, we'll want to inform people that they're vulnerable but there was nothing we could do.

I'm hesitant to change the behavior of the dry-run script from how we run it in production. It's a real bummer that we don't have any tests for this script, as those would really help communicating what's intended and what's not in this code.

SkipKayhil:
> 
> 
> I think some cases would now be missed, like:
> 
> ```
> puts " => there is no available non-vulnerable version"
> ```
> 
> This is valuable information for how we run Dependabot at GitHub, because if this is the case, we'll want to inform people that they're vulnerable but there was nothing we could do.
> 
> I'm hesitant to change the behavior of the dry-run script from how we run it in production. It's a real bummer that we don't have any tests for this script, as those would really help communicating what's intended and what's not in this code.

Ah, I wasn't sure about that. Would it be possible to move the `up_to_date?` block up to line 640 to put it after the vulnerability check but before the conflicting dependencies check? If not, I understand and this can probably be closed.

jurre:
@SkipKayhil yes, I _think_ that would be good. cc @feelepxyz for a quick double-check

feelepxyz:
> Would it be possible to move the `up_to_date?` block up to line 640 to put it after the vulnerability check but before the conflicting dependencies check?

Yeah, although I think it would make more sense to move the conflicting dependency check down instead, and nest it inside the `if requirements_to_unlock == :update_not_possible` block instead. We only want to know which dependencies are conflicting when the update was not possible.

jurre:
@SkipKayhil thank you for looking into this, but I don't believe this currently ready to ship, and the PR has been stale for a bit, in an effort to make our backlog on this project a bit more manageable, I'm going to close it out for now.

Feel free to re-open this or a new PR if you want to iterate on this change.

PR #4089, opened 2021-07-26
Elixir version upgrades

baseballlover723:
updates for

- Erlang deb package: 1.0 -> 2.0
- Elixir: 1.11.4 -> 1.12.2
- Erlang: 23.3.1 -> 24.0.4

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -187,8 +187,8 @@ RUN cd /tmp \
 # Install Erlang, Elixir and Hex
 ENV PATH="$PATH:/usr/local/elixir/bin"
 # https://github.com/elixir-lang/elixir/releases
-ARG ELIXIR_VERSION=v1.11.4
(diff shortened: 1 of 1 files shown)


Nishnha (approved):
LGTM 👍🏾

jurre:
I think @mctofu and I actually ran into some certificate errors when running this in production

Nishnha:
@baseballlover723 Thank you for your contribution!

Newer versions of Elixir are causing issues with the way we run Dependabot in production. For this reason we've pinned it to the latest working version. Unfortunately we have to avoid upgrading Elixir at the moment.

Closing the PR.

jurre:
FYI we're having issues with the OTP version, so that needs to stay at 23, I think the rest of this should be safe to update.

I'm documenting the pinned version for future reference: https://github.com/dependabot/dependabot-core/pull/4092

PR #4093, opened 2021-07-27
Upgrade elixir to v1.12.2

jurre:
This was previously done here: https://github.com/dependabot/dependabot-core/pull/4089

But that was reverted after being released due to an issue with the
erlang upgrade in that PR. This bumps only the elixir version, which
should be safe.


brrygrdn (approved):
👍🏻  Makes sense to do an isolated update

PR #4108, opened 2021-08-02
go_modules: Handle invalid version error from transferred package

jurre:
Quite a few projects depend on k8s, of which older versions depend on
github.com/mholt/caddy. That repo has been transferred though, and when
`go mod` tries to resolve the old path, it fails, likely because that
repo has been replaced with a private repository, which stops github
from automatically redirecting to the new location.

These errors would result in an "unknown error", this catches them and
marks them as a resolvability error, which seems more descriptive.


go_modules/spec/fixtures/projects/transferred_private_package/go.sum
```diff
@@ -0,0 +1,573 @@
+bitbucket.org/bertimus9/systemstat v0.0.0-20180207000608-0eeff89b0690/go.mod h1:Ulb78X89vxKYgdL24HMTiXYHlyHEvruOj1ZPlqeNEZM=
```
jurre, line 1: This is a hot mess, because I yanked this from an actually failing (public) repo, but since this project doesn't build, I can't `go mod tidy` this go.sum file. Leaving it as is seemed better than trying to manually clean it up.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -33,6 +33,8 @@ class GoModUpdater
           # Package pseudo-version does not match the version-control metadata
           # https://golang.google.cn/doc/go1.13#version-validation
           /go: .*: invalid pseudo-version/m.freeze,
+          # Package has been transferred and the original location is now private (mholt/caddy)
+          /go: .*: invalid version/m.freeze,
```
mctofu, line 37: You can also get an invalid version message if you specify a version that doesn't exist but that seems in line with the errors around not finding the package in the repo.

```
go get github.com/caddyserver/caddy/v2@v2.9.9
go get: github.com/caddyserver/caddy/v2@v2.9.9: invalid version: unknown revision v2.9.9
```

mctofu (approved):

jurre:
I think this is covered better in #4130

PR #4115, opened 2021-08-03
Terraform: Pin fixture to range to prevent tests failing

jurre:
This test was currently set up in a way that caused new releases of
Mongey/confluentcloud to make our tests fail. Now we pin it to a range,
in line with other tests


brrygrdn (approved):
Thanks for fixing this 👍🏻

PR #4117, opened 2021-08-03
Skip classifier when checking for internal dep

honnix:
When building up the list of internal dependencies, it is not possible
to know the classifiers, so we skip appending classifier to a dependency
name when checking whether it is an internal dependency or not.

The details of this proposed change is described by https://github.com/dependabot/dependabot-core/pull/1924#issuecomment-891981747https://github.com/dependabot/dependabot-core/pull/1924#issuecomment-891981747

### DIFF
--- a/maven/spec/fixtures/poms/multimodule_pom.xml
+++ b/maven/spec/fixtures/poms/multimodule_pom.xml
@@ -49,6 +49,12 @@
                 <artifactId>util</artifactId>
                 <version>${project.version}</version>
             </dependency>
+            <dependency>
+                <groupId>net.evenh.multimodule</groupId>
+                <artifactId>util</artifactId>
+                <version>${project.version}</version>
+                <classifier>sources</classifier>
+            </dependency>
         </dependencies>
     </dependencyManagement>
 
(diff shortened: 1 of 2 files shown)


honnix:
cc @a1flecke

maven/spec/fixtures/poms/multimodule_pom.xml
```diff
@@ -49,6 +49,12 @@
                 <artifactId>util</artifactId>
                 <version>${project.version}</version>
             </dependency>
+            <dependency>
+                <groupId>net.evenh.multimodule</groupId>
+                <artifactId>util</artifactId>
+                <version>${project.version}</version>
+                <classifier>sources</classifier>
```
feelepxyz, line 56: Could you add a test case in the filer_parser spec covering this case? Does this change mean we correctly detect internal dependencies?

honnix: There is already a test case for this. Adding this classified dependency should cover the change. And yes, without the patch this added dependency will fail a test case where it asserts all dependencies to look for updates.

honnix: https://github.com/dependabot/dependabot-core/blob/main/maven/spec/dependabot/maven/file_parser_spec.rb#L725 should be the test case.

feelepxyz: Oh gotcha! Thanks for linking 👍

feelepxyz (approved):

PR #4130, opened 2021-08-11
Handle errors where go module dependencies are unreachable

mctofu:


go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -239,6 +239,10 @@ def handle_subprocess_error(stderr) # rubocop:disable Metrics/AbcSize
             raise Dependabot::DependencyFileNotResolvable, error_message
           end
 
+          if (matches = stderr.match(/Authentication failed for '(?<url>.+)'/))
```
mctofu, line 242: I moved this up to fix this test which was providing invalid credentials: https://github.com/dependabot/dependabot-core/blob/main/go_modules/spec/dependabot/go_modules/file_updater_spec.rb#L137

I think it also makes sense to prioritize reporting that your credentials are invalid since that could be the root cause that the git deps aren't reachable.

go_modules/spec/dependabot/go_modules/file_updater/go_mod_updater_spec.rb
```diff
@@ -9,12 +9,7 @@
   let(:updater) do
     described_class.new(
       dependencies: [dependency],
-      credentials: [{
-        "type" => "git_source",
-        "host" => "github.com",
-        "username" => "x-access-token",
-        "password" => "token"
-      }],
+      credentials: empty_credentials,
```
mctofu, line 12: It's more realistic to have empty or valid credentials here. Fake credentials will result in `PrivateSourceAuthenticationFailure` in some of the cases. We have a test for that in another file but I think we could move those cases over here as a future cleanup.

jurre (approved):
This is fantastic, thanks!

jerbob92:
Could it be that this makes my test fail?
https://github.com/dependabot/dependabot-core/pull/4140/checks?check_run_id=3342250845

PR #4140, opened 2021-08-16
Ignore replaced dependencies

jerbob92:
This pull request makes sure that replaced modules are not being updated. I saw this behavior with one of my own projects where Dependabot updated the original import directive, but since the dependency itself was replaced by another directive, the actual update had no effect. 

Basically the go.mod looked like this:

```
require (
	github.com/dghubble/oauth1 v0.6.0
)

replace github.com/dghubble/oauth1 => github.com/klippa-app/oauth1 v0.0.0-20190731095211-0d6b37cb52ba
```

And the PR that Dependabot created resulted in this:
```
-	github.com/dghubble/oauth1 v0.6.0
+	github.com/dghubble/oauth1 v0.7.0
```	

While technically correct, since that is the latest version of `github.com/dghubble/oauth1`, this PR doesn't do anything, since the import is replaced. I updated the `FileParser` to be aware of these situations, and it now mimics the behaviour of `go get -u`.

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_parser.rb
+++ b/go_modules/lib/dependabot/go_modules/file_parser.rb
@@ -52,7 +52,7 @@ module Dependabot
         Dependency.new(
           name: details["Path"],
           version: version,
-          requirements: details["Indirect"] ? [] : reqs,
+          requirements: details["Indirect"] || dependency_is_replaced(details) ? [] : reqs,
           package_manager: "go_modules"
         )
       end
@@ -92,6 +92,15 @@ module Dependabot
 
       def local_replacements
         @local_replacements ||=
+          # Find all the local replacements, and return them with a stub path
+          # we can use in their place. Using generated paths is safer as it
+          # means we don't need to worry about references to parent
+          # directories, etc.
+          ReplaceStubber.new(repo_contents_path).stub_paths(manifest, go_mod.directory)
+      end
+
+      def manifest
+        @manifest ||=
           SharedHelpers.in_a_temporary_directory do |path|
             File.write("go.mod", go_mod.content)
 
@@ -106,12 +115,7 @@ module Dependabot
             stdout, stderr, status = Open3.capture3(env, command)
             handle_parser_error(path, stderr) unless status.success?
 
-            # Find all the local replacements, and return them with a stub path
-            # we can use in their place. Using generated paths is safer as it
-            # means we don't need to worry about references to parent
-            # directories, etc.
-            manifest = JSON.parse(stdout)
-            ReplaceStubber.new(repo_contents_path).stub_paths(manifest, go_mod.directory)
+            JSON.parse(stdout)
           end
       end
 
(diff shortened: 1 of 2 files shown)


jerbob92:
I'd like some advice by the team, because this made a lot of tests fail, because these dependencies aren't returned anymore.
I think there are 2 options:

- I change the logic to return the dependency but leave the requirements empty, this will also leave the monorepo working intact
- I fix up all the unit tests

jerbob92:
I don't know why this test fails, it doesn't really seem to be related to my changes?

jerbob92:
Tests are fixed :)

go_modules/lib/dependabot/go_modules/file_parser.rb
```diff
@@ -172,6 +176,24 @@ def skip_dependency?(dep)
...
+        # Mark dependency as replaced if the requested dependency has a
+        # "replace" directive and that either has the same version, or no
+        # version mentioned. This mimics the behaviour of go get -u, and
+        # prevents that we change dependency versions without any impact since
+        # the actual version that is being imported is defined by the replace
+        # directive.
+        if manifest["Replace"]
+          dep_replace = manifest["Replace"].find do |replace|
+            replace["Old"]["Path"] == details["Path"] &&
+              (!replace["Old"]["Version"] || replace["Old"]["Version"] == details["Version"])
```
Nishnha, line 190: This is a bit odd to read but the comment above clarifies this check 👍🏾

go_modules/lib/dependabot/go_modules/file_parser.rb
```diff
@@ -92,6 +92,15 @@ def required_packages
...
         @local_replacements ||=
+          # Find all the local replacements, and return them with a stub path
+          # we can use in their place. Using generated paths is safer as it
+          # means we don't need to worry about references to parent
+          # directories, etc.
+          ReplaceStubber.new(repo_contents_path).stub_paths(manifest, go_mod.directory)
+      end
+
+      def manifest
+        @manifest ||=
```
Nishnha, line 103: +1 for caching the manifest

go_modules/spec/dependabot/go_modules/file_parser_spec.rb
```diff
@@ -47,7 +47,7 @@
         parser.parse.select(&:top_level?)
       end
 
-      its(:length) { is_expected.to eq(3) }
+      its(:length) { is_expected.to eq(2) }
```
Nishnha, line 50: The fixture file is located at https://github.com/dependabot/dependabot-core/blob/main/go_modules/spec/fixtures/go_mods/go.mod

There are three direct dependencies in the fixture, but rsc.io/qr uses a replace directive, so it is not counted as a top-level anymore with this change.

Nishnha (approved):
Hi @jerbob92 thank you for submitting this pull request and rebasing it. 
Sorry we took so long to review this--the Dependabot team is currently focused on another project.

This PR checks out and the test looks good to me. 
Thank you for your contribution 😄

PR #4142, opened 2021-08-17
Handle unreachable go module dependencies in LatestVersionFinder

mctofu:
Previously we improved handling of unreachable go module dependencies when performing an update in https://github.com/dependabot/dependabot-core/pull/4130

This updates the `LatestVersionFinder` to detect unreachable dependency errors as well. Previously we'd return an `UnknownError` in this case where Dependabot checks if an unreachable dependency needs an update.

I've also continued replacing the default credentials used in the test with empty credentials. This is a more realistic use case since using made up credentials can result in other unexpected errors: https://github.com/dependabot/dependabot-core/pull/4141


brrygrdn (approved):
👍🏻

Issue #2138, opened 2019-06-20
Support "apply from:" in Gradle


wtanaka:
It's possible to use "apply from" in gradle to include other .gradle files:

https://github.com/wtanaka/ciconfiggen/blob/cc8bfc43f7e9e4109310c9fb48af6ad83cdacf13/build.gradle#L2-L4

Based on what's shown in the web dashboard, it looks like dependabot might not be scanning dependencies in these included files:

https://app.dependabot.com/accounts/wtanaka/repos/178891901

tunderwood:
Following since I would also find this feature useful. I was actually just looking at the code and it appears that this is actually semi implemented for a very specific filename. 

It wouldn't solve this exact use case but it looks like they will scan dependencies in a file located at `<default dir>/gradle/dependencies.gradle` without any extra configuration

This seems to only work if you include the line:  `apply from: 'gradle/dependencies.gradle'` in your `build.gradle` file. Since the functionality is already there for this specific file, it might not be too hard to extend it to work with other custom file names.

Issue #1164, opened 2019-05-23
Fully Featured Gradle Support with the Gradle Tooling API


JLLeitschuh:
*Foreward* I'm not intimately familiar with how Dependabot works so please excuse me if I wildly misrepresent something about Dependabot.

## Problem

Currently, Dependabot is attempting to parse `build.gradle` files as text files instead of fully understanding the complexities of Gradle Builds.

This is because the dependencies declared in a `build.gradle` file doesn't tell the full story.

Additionally, since Gradle now supports the Gradle Kotlin DSL, there are now `build.gradle.kts` files. On top of that, users can also change the name of their build files so that they call the files `my-project.gradle.kts` and Gradle will handle that fine.

This current methodology completely misses any dependencies added by plugins, transitive dependencies, and dependencies added using Gradle in abnormal but valid ways.

Fundamentally this issue arises because unlike NPM and Yarn, where dependency declaration files are static, Gradle & Maven both require code execution to fully resolve the dependency graph.

## Solution

Gradle offers something called the [Gradle Tooling API](https://docs.gradle.org/current/userguide/embedding.html). This tooling API is already in use in IntelliJ, Eclipse, and the Gradle Test Kit.

The Tooling API exposes the entire dependency graph for the repositories build in a way that can be transformed into a format Dependabot can understand.

The Tooling API executes the Gradle Build and can be used to extract metadata about the build for consumption by external tools.

## Risks

There are certain risks that are taken on by moving forward with this plan.

### Security

In order to use the Tooling API, Dependabot will be required to execute the user's code in their repository. Fundamentally, this introduces untrusted code execution into the same environment where Dependabot is being executed. I don't know what the security model of Dependabot's execution is, but if executing user code is a non-starter, using the Tooling API won't work.

### Performance

Sometimes developers have really, really weird builds. Sometimes they also have really slow builds. Calling the ToolingAPI will bring the Dependabot execution time from being predictable with respect to the number of files being parsed, to a non-deterministic amount of execution time.

### Enviromental

Certain versions of Gradle only work with certain versions of the JVM. Certain builds only work with certain versions of the JVM. Implementing this feature will require maintainers of repositories to explicitly declare what version of the JVM their build depends upon.

## Rewards

Keeping the Java Ecosystem Safe.

[...]

PR #4155, opened 2021-08-19
Improved support `apply from` in gradle files

zbynek:
For now work in progress, should fix https://github.com/dependabot/dependabot-core/issues/2138

I still need to add some tests etc, for now I'm just wondering if there is any interest in getting this merged as a stop-gap solution before the Gradle Tooling is integrated as proposed in https://github.com/dependabot/dependabot-core/issues/1164

### DIFF
--- a/gradle/lib/dependabot/gradle/file_parser.rb
+++ b/gradle/lib/dependabot/gradle/file_parser.rb
@@ -49,6 +49,20 @@ module Dependabot
         dependency_set.dependencies
       end
 
+      def self.find_include_names(buildfile)
+        return [] unless buildfile
+
+        buildfile.content.
+          scan(/apply(\(| )\s*from(\s+=|:)\s+['"]([^'"]+)['"]/).
+          map { |match| match[2] }
+      end
+
+      def self.find_includes(buildfile, dependency_files)
+        FileParser.find_include_names(buildfile).
+          map { |f| dependency_files.find { |bf| bf.name == f } }.
+          compact
+      end
+
       private
 
       def map_value_regex(key)
@@ -296,13 +310,6 @@ module Dependabot
         @buildfiles ||= dependency_files.select do |f|
           f.name.end_with?(*SUPPORTED_BUILD_FILE_NAMES)
         end
-      end
-
-      def self.find_includes(buildfile, dependency_files)
-        buildfile.content.
-          scan(/apply from(\s+=|:)\s+['"]([^'"]+)['"]/).flatten.
-          map { |f| dependency_files.find { |bf| bf.name == f } }.
-          compact
       end
 
       def script_plugin_files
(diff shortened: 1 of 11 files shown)


zbynek:
@greysteil do you think it's worth it to finish and merge this? I understand #1164 is a better approach, but I don't have the knowledge or time needed to implement that and there seems to be no progress on that issue.

greysteil:
Hey @zbynek, I'm not the right person to ask! @asciimike / @hmarr are likely to have a better view.

asciimike:
@JLLeitschuh might be a good candidate to review this, if available; otherwise, someone on our eng team will take a look at this when free.

JLLeitschuh:
A set of unit/integration tests would help me better understand what you're trying to do here

gradle/lib/dependabot/gradle/file_parser.rb
```diff
@@ -298,13 +298,17 @@ def buildfiles
         end
       end
 
+      def self.find_includes(buildfile, dependency_files)
+        buildfile.content.
+          scan(/apply from(\s+=|:)\s+['"]([^'"]+)['"]/).flatten.
```
JLLeitschuh, line 303: This won't work for the Kotlin DSL as the syntax is different. Also, I don't think this covers the full Groovy syntax that's supported for this declaration.

JLLeitschuh:
You may want to check for recursion and make sure that your logic doesn't get stuck checking two files that are potentially self-referential.

zbynek:
@JLLeitschuh thanks for checking this! I've added some tests. 

There aim of this PR is not to implement full support for `apply`, only to add support for some usecases.

Improvements in this PR:
* resolve property defined in included file and used in (sub)project build file
* resolve repositories defined in a file included from the main project file
* basic Kotlin syntax `apply(from = "dependencies.gradle.kts")`

Existing limitations not addressed by this PR:
* no support for more advanced syntaxes (e.g. filename defined by property, Koltin `apply` with `mapOf` etc.)
* no recursive resolution of includes
* no support for `apply to:` (`to` is ignored)
* all included files that are actually fetched must contain `dependencies` in the filename

JLLeitschuh:
> all included files that are actually fetched must contain dependencies in the filename

That will require some kind of documentation somewhere too

JLLeitschuh:
I don't think you should drop this PR, but I do want to point you towards the more idiomatic way of doing this with Gradle.

We're encouraging our users to move away from using `apply from:` to using either plugins, or using platforms.

https://docs.gradle.org/current/userguide/platforms.html

JLLeitschuh:
The test look sane, so a 👍 for me from here.

zbynek:
@JLLeitschuh thanks. For the docs I created https://github.com/github/docs/pull/10041/files .

zbynek:
> The test look sane, so a 👍 for me from here.

@asciimike given the approval from @JLLeitschuh , what are the next steps to get this merged?

jurre (approved):
Thanks for the contribution, the code looks good to me with the updated docs 👍

PR #4170, opened 2021-08-25
added php mcrypt extension 

oleg-andreyev:
Before
```bash
dependabot@d0e8d7a84fce:~$ php -i | grep mcrypt
dependabot@d0e8d7a84fce:~$
```

After
```bash
dependabot@102a62523fb5:~# php -i | grep mcrypt
/etc/php/7.4/cli/conf.d/20-mcrypt.ini,
Registered Stream Filters => zlib.*, string.rot13, string.toupper, string.tolower, string.strip_tags, convert.*, consumed, dechunk, convert.iconv.*, mcrypt.*, mdecrypt.*
mcrypt
mcrypt support => enabled
mcrypt_filter support => enabled
mcrypt.algorithms_dir => no value => no value
mcrypt.modes_dir => no value => no value
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -156,6 +156,7 @@ RUN add-apt-repository ppa:ondrej/php \
     php7.4-xml \
     php7.4-zip \
     php7.4-zmq \
+    php7.4-mcrypt \
   && rm -rf /var/lib/apt/lists/*
 USER dependabot
 # Perform a fake `composer update` to warm ~/dependabot/.cache/composer/repo


jurre:
Why do we need it @oleg-andreyev?

oleg-andreyev:
Some dependencies may require "ext-mcrypt", so dependabot won't be able to upgrade them, because system does not have mcrypt.

oleg-andreyev:
Proprietary project but there is an example like https://packagist.org/packages/yiisoft/yii + custom composer.json with require ext-mcrypt

jurre:
> Proprietary project but there is an example like https://packagist.org/packages/yiisoft/yii + custom composer.json with require ext-mcrypt

Interesting, I've definitely seen updates for yii work. What would need to be in the composer.json to trigger this error? Do you think it's worth adding a test for it?

jurre (approved):
I guess it can't hurt, there's already a big list of these dependencies that we pull in, what's one more

PR #4179, opened 2021-08-30
fix: Parse multiple requirements from a poetry dependency

xlgmokha:
# Why is this needed?

Poetry projects can specify multiple requirements for a dependency.

e.g.

```toml
numpy = [
	{ version = "^1.19", python = "^3.6,<3.7" },
	{ version  = "^1.20", python = "^3.7" }]
scipy = [
	{ version = "^1.5", python = "^3.6,<3.7" },
	{ version = "^1.6.0", python = "^3.7" }
]
```

This change is needed to allow the parser to parse dependencies with multiple
requirements.

## What does this do?

This change handles poetry dependencies with multiple requirements by looping
through each requirement specified for a dependency.

## Type of change

### DIFF
--- a/python/lib/dependabot/python/file_parser/poetry_files_parser.rb
+++ b/python/lib/dependabot/python/file_parser/poetry_files_parser.rb
@@ -56,8 +56,9 @@ module Dependabot
           dependencies
         end
 
+        # @param req can be an Array, Hash or String that represents the constraints for a dependency
         def parse_requirements_from(req, type)
-          Array(req).compact.map do |requirement|
+          [req].flatten.compact.map do |requirement|
             next if requirement.is_a?(Hash) && (UNSUPPORTED_DEPENDENCY_TYPES & requirement.keys).any?
             check_requirements(requirement)
 
--- a/python/lib/dependabot/python/file_parser/poetry_files_parser.rb
+++ b/python/lib/dependabot/python/file_parser/poetry_files_parser.rb
@@ -43,22 +43,26 @@ module Dependabot
 
             deps_hash.each do |name, req|
               next if normalise(name) == "python"
-              next if req.is_a?(Hash) && UNSUPPORTED_DEPENDENCY_TYPES.any? { |t| req.key?(t) }
 
-              check_requirements(req)
+              requirements = Array(req).compact.map do |detected_requirement|
+                next if detected_requirement.is_a?(Hash) && UNSUPPORTED_DEPENDENCY_TYPES.any? { |t| detected_requirement.key?(t) }
 
-              dependencies <<
-                Dependency.new(
-                  name: normalise(name),
-                  version: version_from_lockfile(name),
-                  requirements: [{
-                    requirement: req.is_a?(String) ? req : req["version"],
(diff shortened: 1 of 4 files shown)


python/lib/dependabot/python/file_parser/poetry_files_parser.rb
```diff
@@ -43,28 +43,38 @@ def pyproject_dependencies
...
             end
           end
 
           dependencies
         end
 
+        # @param req can be an Array, Hash or String that represents the constraints for a dependency
+        def parse_requirements_from(req, type)
+          [req].flatten.compact.map do |requirement|
+            next if requirement.is_a?(Hash) && (UNSUPPORTED_DEPENDENCY_TYPES & requirement.keys).any?
```
jurre, line 65: Just double-checking, this is functionally the same as `next if req.is_a?(Hash) && UNSUPPORTED_DEPENDENCY_TYPES.any? { |t| req.key?(t) }` right?

python/lib/dependabot/python/file_parser/poetry_files_parser.rb
```diff
@@ -43,28 +43,38 @@ def pyproject_dependencies
...
+
+            check_requirements(requirement)
+
+            {
+              requirement: requirement.is_a?(String) ? requirement : requirement["version"],
```
jurre, line 70: Slightly paranoid that we might get something that's not a String but also does not respond to `#[]` in a way that we'd expect but since no tests break that might not be a concern. Should we handle the type of `req` more explicitly?

Totally fine to keep as is, just thinking out loud here.

python/lib/dependabot/python/file_parser/poetry_files_parser.rb
xlgmokha: Yep. It's a set intersect operation.

```ruby
irb(main):001:0> [1, 2, 3] & [2, 3]
=> [2, 3]
irb(main):002:0> [1, 3, 4] & [5, 6]
=> []
```

jurre (approved):
Nice one! I slow-chatted in the review-comments about a possibility to make the code a little more explicit, but it's totally fine to keep as is.

python/lib/dependabot/python/file_parser/poetry_files_parser.rb
xlgmokha: > Slightly paranoid that we might get something that's not a String but also does not respond to #[] in a way that we'd expect but since no tests break that might not be a concern.

I think this is a valid concern. Matching on type signature feels icky to me. I *feel* like there's a missing abstraction/interface here but I'm not sure what that is.

> Should we handle the type of req more explicitly?

I'm not against this idea. I worked on this issue during my on-call shift last week so I've lost some of the context due to mo's limited memory but I can circle back to this. I think I would prefer to ship this small change as is because there were some jobs failing with this issue so this would at least unblock those. WDYT?

jurre: > WDYT?

👍

PR #4184, opened 2021-08-31
Fix failing builds

xlgmokha:
Pinning to an older version of faraday appears to improve the condition of the builds. A new version of faraday was released yesterday. 

* https://rubygems.org/gems/faraday/versions
* https://github.com/lostisland/faraday/releases/tag/v1.7.1
* https://github.com/lostisland/faraday/pull/1306

A golang test is failing and I was able to reproduce it with:

```bash
~/src/github.com/dependabot/dependabot-core/go_modules/helpers [fix-build]
モ echo "{\"function\":\"getVcsRemoteForImport\",\"args\":{\"import\":\"code.cloudfoundry.org/bytefmt\"}}" | go run main.go
{"error":"Cannot detect VCS"}exit status 1
```

I was able to resolve this by falling back to http after attempting https resolution for paths that do not have a scheme specified.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,6 +227,7 @@ Dependabot and Dependabot Core started life as [Bump][bump] and
 GoCardless in helping make Dependabot possible - if you need to collect
 recurring payments from Europe, check them out.
 
+
 [dependabot]: https://dependabot.com
 [dependabot-status]: https://api.dependabot.com/badges/status?host=github&identifier=93163073
 [dependabot-script]: https://github.com/dependabot/dependabot-script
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -227,6 +227,7 @@ Dependabot and Dependabot Core started life as [Bump][bump] and
 GoCardless in helping make Dependabot possible - if you need to collect
 recurring payments from Europe, check them out.
 
+
```
xlgmokha, line 230: Intentional `noop` added to highlight the build failure before and after pinning to a specific version of faraday. I'll see if I can work upstream to help identify the root cause.

Nishnha:
Looks like there are a couple Go tests that have regressed from version updates. I think that pinning the Faraday version for stability here is the right call, though, and we can fix the Go tests in another PR.

Nishnha (approved):

xlgmokha:
Thanks for approving this @Nishnha.

@jurre, @brrygrdn as per our conversation earlier I've disabled the flaky test for the time being and reverted the change to fallback to http. Do you mind double checking that I haven't included/excluded anything that wasn't discussed?

Issue #4181, opened 2021-08-31
Terraform module versions with 'v' prefix


declan-fitzpatrick:
**Package ecosystem**
terraform

**Language version**
Terraform v1.0.5

**dependabot.yml content**

```yaml
version: 2
registries:
  terraform-cloud:
    type: terraform-registry
    url: https://app.terraform.io
    token: ${{secrets.TERRAFORM_API_TOKEN}}

updates:
  - package-ecosystem: "terraform"
    registries:
      - terraform-cloud
    directory: "/"
    target-branch: "release"
    labels:
      - "terraform"
    schedule:
      interval: "daily"

```
**Description**

Terraform registry allows for semantic version, which can [optionally be prefixed with a `v`](https://www.terraform.io/docs/registry/modules/publish.html#requirements). 
Dependabot throws a `BadRequirementError` when module versions are prefixed with a `v`. The full error: 

**Steps to reproduce**
1. set requirement_string to v1.0.0 in `terraform/spec/dependabot/terraform/requirement_spec.rb` 
2. run `bundle exec rspec spec`
output looks like: 
```shell
Randomized with seed 59800
....................................................F........................................................................................................................

Failures:

  1) Dependabot::Terraform::Requirement.new 
     Failure/Error: raise BadRequirementError, msg
     
     Gem::Requirement::BadRequirementError:
       Illformed requirement ["v1.0.0"]
     # ./lib/dependabot/terraform/requirement.rb:15:in `parse'
     # ./lib/dependabot/terraform/requirement.rb:37:in `initialize'
     # ./spec/dependabot/terraform/requirement_spec.rb:7:in `new'
     # ./spec/dependabot/terraform/requirement_spec.rb:7:in `block (2 levels) in <top (required)>'
     # ./spec/dependabot/terraform/requirement_spec.rb:11:in `block (3 levels) in <top (required)>'
     # /home/declan/Documents/repo/projects/dependabot-core/common/spec/spec_helper.rb:49:in `block (2 levels) in <top (required)>'
     # /var/lib/gems/2.7.0/gems/webmock-3.14.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'

Finished in 32.72 seconds (files took 0.56344 seconds to load)
173 examples, 1 failure

Failed examples:

rspec ./spec/dependabot/terraform/requirement_spec.rb:11 # Dependabot::Terraform::Requirement.new 

Randomized with seed 59800
```

[...]

PR #4191, opened 2021-09-02
handle terraform module versions with a 'v' prefix

declan-fitzpatrick:
Possible workaround for #4181 

This adds `v?` to the `PATTERN` regex used in Gem::Requirement to support Terraform configuration where releases are tagged with a `v` prefix.

This `v` prefix will be dropped if dependabot does do an update. This will force a different formatting that users may not desire. However, Terraform forces consistent (and opinionated) formatting already with [`terraform fmt`](https://www.terraform.io/docs/cli/commands/fmt.html), and considering the Terraform registry returns a semantically correct version from the [API](https://www.terraform.io/docs/cloud/api/modules.html) regardless of whether or not the publisher uses the prefix, it seems reasonable to force this consistency.


terraform/spec/dependabot/terraform/requirement_spec.rb
```diff
@@ -16,3 +16,17 @@
...
+
+  describe ".new" do
+    it { is_expected.to be_a(described_class) }
+
+    context "with a comma-separated string" do
+      let(:requirement_string) { "~> v4.2.5, >= v4.2.5.1" }
+      it { is_expected.to eq(described_class.new("~> 4.2.5", ">= 4.2.5.1")) }
+    end
+  end
+end
```
jurre, line 32:
```suggestion
end

```

terraform/spec/dependabot/terraform/requirement_spec.rb
```diff
@@ -16,3 +16,17 @@
...
   end
 end
+
+RSpec.describe Dependabot::Terraform::Requirement do
+  subject(:requirement) { described_class.new(requirement_string) }
+  let(:requirement_string) { ">=v1.0.0" }
+
+  describe ".new" do
+    it { is_expected.to be_a(described_class) }
+
```
jurre, line 26: No need to repeat all of this, we can nest the code in the existing block and it will inherit all of the setup of the previous tests 👍 

```suggestion

```

terraform/lib/dependabot/terraform/requirement.rb
```diff
@@ -7,6 +7,13 @@
 module Dependabot
   module Terraform
     class Requirement < Gem::Requirement
+      # Override regex PATTERN from Gem::Requirement to add support for the
+      # optional 'v' prefix to release tag names, which Terraform supports.
+      # https://www.terraform.io/docs/registry/modules/publish.html#requirements
+      quoted = OPS.keys.map { |k| Regexp.quote k }.join "|"
+      PATTERN_RAW = "\\s*(#{quoted})?\\s*v?(#{Gem::Version::VERSION_PATTERN})\\s*"
```
jurre, line 14:
```suggestion
      OPERATORS = OPS.keys.map { |k|ey Regexp.quote(key) }.join("|").freeze
      PATTERN_RAW = "\\s*(#{OPERATORS})?\\s*v?(#{Gem::Version::VERSION_PATTERN})\\s*"
```

jurre (commented):
I made some suggestions for things that rubocop will complain about once I trigger CI (which I'll do right now).

I think the change makes sense, but it'd be nice to have a test for the [`FileUpdater`](https://github.com/dependabot/dependabot-core/blob/dd1fba76003434639209627676febc595787ba7c/terraform/spec/dependabot/terraform/file_updater_spec.rb#L9) that shows an existing `v` prefix being stripped out after the update.

Thanks for contributing!

terraform/lib/dependabot/terraform/requirement.rb
declan-fitzpatrick: updated this with commit 99a4d42, because typo in `|k|ey`

declan-fitzpatrick:
Hi @jurre

I have made the changes that you asked for. 
I only added the tests for modules to show the `v` being stripped, because provider blocks with `required_providers` blocks and the `v` prefix are actually invalid Terraform, for example: 
```hcl
terraform {
  required_version = "~> v1.0"

  required_providers {
    confluentcloud = {
      source  = "Mongey/confluentcloud"
      version = "v0.0.12"
    }
  }
}
```
gives the following output from `terraform init`: 
```shell
Initializing the backend...
╷
│ Error: Invalid version constraint
│ 
│   on pro.tf line 5, in terraform:
│    5:     confluentcloud = {
│    6:       source  = "Mongey/confluentcloud"
│    7:       version = "v0.0.12"
│    8:     }
│ 
│ Incorrect version constraint syntax: a "v" prefix should not be used.

```

terraform/lib/dependabot/terraform/requirement.rb
jurre: Ah yeah my bad!

jurre:
@Nishnha would love if you could take a look at this as well

terraform/lib/dependabot/terraform/requirement.rb
```diff
@@ -7,6 +7,13 @@
 module Dependabot
   module Terraform
     class Requirement < Gem::Requirement
+      # Override regex PATTERN from Gem::Requirement to add support for the
+      # optional 'v' prefix to release tag names, which Terraform supports.
+      # https://www.terraform.io/docs/registry/modules/publish.html#requirements
```
xlgmokha, line 12: Nice! :heart:

Nishnha (approved):
Everything here looks great to me! 
Thank you for adding in tests for the 'v' prefix being removed after the update.

Given that Terraform modules support the `v` prefix while providers do not, and that the prefix is dropped by `terraform fmt`, I suspect the Terraform authors added support as a workaround for old modules already using the prefix.

It's difficult to measure the impact of a compatibility change like this, but this will definitely make a difference in the security of the Terraform community, so thank you for this PR @declan-fitzpatrick  :)

Issue #3676, opened 2021-05-09
git checkout fails due to some really long filenames on Windows


SingingBush:
I decided to take a look at making changes to support D this morning but couldn't even work on the repo on Windows due to insanely long spec filenames in common:

```
Cloning into 'dependabot-core'...
remote: Enumerating objects: 108774, done.
remote: Counting objects: 100% (1807/1807), done.
remote: Compressing objects: 100% (896/896), done.
remote: Total 108774 (delta 959), reused 1579 (delta 773), pack-reused 106967
Receiving objects: 100% (108774/108774), 80.01 MiB | 3.00 MiB/s, done.
Resolving deltas: 100% (73984/73984), done.
error: unable to create file common/spec/fixtures/vcr_cassettes/Dependabot_MetadataFinders_Base_ChangelogFinder/_changelog_text/with_a_github_repo/with_a_changelog/for_a_git_dependency_with_multiple_sources/when_the_ref_has_changed/1_2_1_1_6_1_1.yml: Filename too long
warning: unable to access 'common/spec/fixtures/vcr_cassettes/Dependabot_MetadataFinders_Base_ReleaseFinder/_releases_text/with_a_github_repo/with_releases/when_the_release_is_present/updating_from_no_previous_release_to_new_release/.gitattributes': Filename too long
error: unable to create file common/spec/fixtures/vcr_cassettes/Dependabot_MetadataFinders_Base_ReleaseFinder/_releases_text/with_a_github_repo/with_releases/when_the_release_is_present/updating_from_no_previous_release_to_new_release/1_2_1_1_1_4_1.yml: Filename too long
fatal: cannot create directory at 'common/spec/fixtures/vcr_cassettes/Dependabot_MetadataFinders_Base_ReleaseFinder/_releases_text/with_a_github_repo/with_releases/when_the_release_is_present/updating_from_uknown_previous_release_to_new_release': Filename too long
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry the checkout with 'git checkout -f HEAD'
```

Windows can be set to allow long paths by setting *LongPathsEnabled* in the registry but I'd still go with perhaps shortening those paths a bit.

edit:
[...]

PR #4198, opened 2021-09-03
Describe workaround for cloning with long file names on Windows

Marcono1234:
Relates to #3676

On Windows cloning the repository might fail with "Filename too long". This pull request adds a section to the README explaining how to solve this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,21 @@ In addition to this library, you may be interested in:
   GitLab or Azure DevOps
 - The [API docs][api-docs] for Dependabot's hosted instance (dependabot.com)
 
+## Cloning the repository
+Clone the repository with Git using:
+
+```
+git clone https://github.com/dependabot/dependabot-core.git
+```
+
+On Windows this might fail with "Filename too long". To solve this, run the
+following commands in the cloned Git repository:
+
+1. `git config core.longpaths true`
+2. `git restore --source=HEAD :/`
+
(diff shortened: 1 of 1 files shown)


Marcono1234:
Hmm `git restore --source=HEAD :/` is suggested to me by Git after the checkout failed:
Git error message

However, it does not seem to work. `git status` still reports changes afterwards.
Using `git reset --hard` works.

Have marked this pull request as draft for now, until it is clear what the proper command is (and have [asked on StackOverflow about this](https://stackoverflow.com/q/69050583)).

jurre:
Thanks @Marcono1234, this would be a great addition to the docs. None of the maintainers work on Windows so the contribution is very much appreciated, lmk if I can help in any way.

Marcono1234:
I would like to wait a few days to get an answer on the StackOverflow question or from the Git for Windows maintainers whether `git restore --source=HEAD :/` (as currently suggested by Git) is indeed the "correct" command and there is a bug in Git for Windows.
In case I don't receive a definite response there I will edit this pull request to recommend `git reset --hard` instead.

Marcono1234:
I did receive a reply and with some delay on my side posted this issue to the Git mailing list. However, it probably takes some time until this is resolved, so the easiest solution might be to recommend `git reset --hard` for now (which I changed with the latest commit).

Sorry that I did not inform you about the progress.

PR #4215, opened 2021-09-10
bin/dry-run.rb requires a development container to run

brrygrdn:
While it is possible to run `bin/dry-run.rb` on a native machine, it requires the user to correctly install native dependencies, such as the right versions of package managers and so on to be guaranteed to match a 'real' Dependabot run.

Since we use this tool as a first line debugging step, it often adds friction to the process of reproducing builds if a native environment is used.

In order to reduce context and make this process more confident, let's make it a requirement the script is ran using the developer shell so we default to reproducible outcomes.

### DIFF
--- a/Dockerfile.development
+++ b/Dockerfile.development
@@ -2,7 +2,6 @@ FROM dependabot/dependabot-core
 
 # Temporarily switch to root user in order to install packages
 USER root
-RUN touch /etc/dependabot-core-dev
 RUN apt-get update \
   && apt-get install -y vim strace ltrace gdb shellcheck \
   && rm -rf /var/lib/apt/lists/*
--- a/Dockerfile.development
+++ b/Dockerfile.development
@@ -2,6 +2,7 @@ FROM dependabot/dependabot-core
 
 # Temporarily switch to root user in order to install packages
 USER root
+RUN touch /etc/dependabot-core-dev
 RUN apt-get update \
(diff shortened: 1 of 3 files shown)


Dockerfile.development
```diff
@@ -1,5 +1,7 @@
 FROM dependabot/dependabot-core
 
+RUN touch /etc/dependabot-core-dev
```
brrygrdn, line 3: This means the script effectively requires a development container specifically. I'm not sure if there's an argument to put this fingerprint in the base image so the dry run script can be used in production images, so it seemed best to be specific for now.

bin/dry-run.rb
```diff
@@ -35,6 +35,15 @@
 
 # rubocop:disable Style/GlobalVars
 
+unless File.exist?("/etc/dependabot-core-dev")
```
jurre, line 38: WDYT about checking if the user is `dependabot` instead?

```
[dependabot-core-dev] ~/dependabot-core $ whoami
dependabot
```

jurre (approved):
Nice one! I made a suggestion that doesn't require touching a file, but this totally works also

Dockerfile.development
jurre: Pretty sure the script is not included in the prod image 👍

bin/dry-run.rb
brrygrdn: Yeah, I like that, it saves the hassle of having to rebuild images

PR #4250, opened 2021-09-21
update Elixir from 1.12.2 -> 1.12.3

baseballlover723:
This also will update hex, since it will install the newer `0.21.3` (from `0.21.2`) when hex is installed in the docker images.


baseballlover723:
seems like the CI is failing because of https://github.com/nodesource/distributions/issues/1266 being cached or something

PR #4255, opened 2021-09-23
Update NuGet to fetch packages using SemVer 2

Nishnha:
Dependabot is currently failing to search/fetch NuGet packages using SemVer 2 versioning.
The API query we use for package versions seems to only fetch package versions using SemVer 1

For example, on https://www.nuget.org/packages/Xamarin.Forms/, the current version is `5.0.0.2125`,
but Dependabot's current query for NuGet packages, https://api-v2v3search-0.nuget.org/query?q=xamarin.forms&prerelease=true, only shows `4.x.x` and lower releases

To query for newer packages that use SemVer 2, we add `&semVerLevel=2.0.0` to the query string:
https://api-v2v3search-0.nuget.org/query?q=xamarin.forms&prerelease=true&semVerLevel=2.0.0 correctly shows the `5.x.x` releases.

---

This PR also changes the server we use for NuGet queries. 

https://api.nuget.org/v3/index.json returns a list of `SearchQueryService` servers, and api-v2v3search-0.nuget.org is no longer on that list. 
The query server has been changed to azuresearch-usnc.nuget.org, which should return the same results as the previous server.


nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
```diff
@@ -59,7 +59,7 @@ def build_url_for_details(repo_details)
           end
           if search_url
             details[:search_url] =
-              search_url + "?q=#{dependency.name.downcase}&prerelease=true"
+              search_url + "?q=#{dependency.name.downcase}&prerelease=true&semVerLevel=2.0.0"
```
jurre, line 62: Would this still include semver level 1 releases?

Nishnha: Yes it does

Nishnha:
There's a VCR recording in the NuGet test suite causing the CI to fail on this PR. The recording is of fetching a package from a local Artifactory install. 
Unfortunately, I don't seem to be able to re-record this fixture. Artifactory fails to install on my Mac via docker-compose and it apparently requires Linux to run. 

The original VCR fixture was recorded on Dependabot 0.149.5 -- do we know how to re-record this?

jurre (approved):

PR #4257, opened 2021-09-26
Treat tokens after underscore as numeric if possible

### LINKED ISSUE #1636: Gradle (Java) - Dependabot doesn't recognize new version
When checking for new versions does dependabot not recognize a new update version, even tho it is set in the metadata.xml's `<latest>` tags.

The dependency in general is JDA (`net.dv8tion:JDA`) on bintray.
The bot does retrieve the metadata.xml from the site, to check what the latest version is.
And even tho the file has `<latest>4.1.0_100</latest>` and the current version in the repo is set to `4.1.0_99` does the bot not recognize that the version has changed and sees it as "up to date"

Here's the output of the relevant logs:
```
updater | INFO <job_20942363> Checking if net.dv8tion:JDA 4.1.0_99 needs updating
  proxy | 2020/01/20 09:52:15 GET https://repo.maven.apache.org:443/maven2/net/dv8tion/JDA/maven-metadata.xml
  proxy | 2020/01/20 09:52:15 200 https://repo.maven.apache.org:443/maven2/net/dv8tion/JDA/maven-metadata.xml
  proxy | 2020/01/20 09:52:15 GET https://jcenter.bintray.com:443/net/dv8tion/JDA/maven-metadata.xml
  proxy | 2020/01/20 09:52:15 200 https://jcenter.bintray.com:443/net/dv8tion/JDA/maven-metadata.xml
updater | INFO <job_20942363> Latest version is 4.1.0_99
updater | INFO <job_20942363> No update needed for net.dv8tion:JDA 4.1.0_99
```

zbynek:
Fixes #1636

Treat `-` and `_` the same in Gradle version names. Since `1.0.0-10` > `1.0.0-9`, it makes sense to also have `1.0.0_10` > `1.0.0_9`.

### DIFF
--- a/gradle/lib/dependabot/gradle/version.rb
+++ b/gradle/lib/dependabot/gradle/version.rb
@@ -14,7 +14,8 @@ module Dependabot
       NULL_VALUES = %w(0 final ga).freeze
       PREFIXED_TOKEN_HIERARCHY = {
         "." => { qualifier: 1, number: 4 },
-        "-" => { qualifier: 2, number: 3 }
+        "-" => { qualifier: 2, number: 3 },
+        "_" => { qualifier: 2, number: 3 }
       }.freeze
       NAMED_QUALIFIERS_HIERARCHY = {
         "a" => 1, "alpha"     => 1,
@@ -132,7 +133,7 @@ module Dependabot
       end
 
       def split_into_prefixed_tokens(version)
-        ".#{version}".tr("_", "-").split(/(?=[\-\.])/)
+        ".#{version}".split(/(?=[_\-\.])/)
       end
 
       def pad_for_comparison(prefixed_tokens, other_prefixed_tokens)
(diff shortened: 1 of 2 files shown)


gradle/lib/dependabot/gradle/version.rb
```diff
@@ -132,7 +132,7 @@ def convert_dates(version, other_version)
       end
 
       def split_into_prefixed_tokens(version)
-        ".#{version}".split(/(?=[\-\.])/)
+        ".#{version}".tr("_", "-").split(/(?=[\-\.])/)
```
jurre, line 135: I think we could just include it in the regex as one of the values to split on, instead of converting to a dash and then matching on that later on?

```suggestion
        ".#{version}".split(/(?=[\-\.\_])/)
```

zbynek: I did that in the last commit. It requires another change in the file.

jurre: Yep, that makes sense to me, thanks for making the change!

jurre (approved):

Issue #4069, opened 2021-07-21
Support `gradlePluginPortal()` as a Gradle repository


bigdaz:
In addition to `mavenCentral()`, `jcenter()` and `google()`, Gradle supports `gradlePluginPortal()` as an alias for `https://plugins.gradle.org/m2`.
This is added by default when resolving plugins, but if a project uses a plugin jar as an implementation dependency, then `gradlePluginPortal()` can be added to the `repositories` block.

PR #4268, opened 2021-09-29
Add support for gradlePluginPortal()

zbynek:
Fixes https://github.com/dependabot/dependabot-core/issues/4069

Also extracts the URLs into constants, note that trailing slashes are stripped in `repositories_finder.rb` line 115, so excluding trailing slashes from the constants does not change the result.

No constant introduced for jcenter because it's deprecated (both the method and the service).


zbynek:
@jurre thanks for the fast feedback on my previous PR. Will you have time to check this one too?

PR #4277, opened 2021-10-01
fix: allow up to 120 seconds to complete an update for a single dependency

xlgmokha:
# Why is this needed?

To prevent spending too much time trying to update a single dependency during a
scheduled update.

## What does this do?

It sets a timeout of 120 seconds to ensure that the sequential dependency
updates aren't blocked by a single dependency that is having difficulty being
updated.

### DIFF
--- a/bundler/helpers/v2/run.rb
+++ b/bundler/helpers/v2/run.rb
@@ -2,6 +2,7 @@
 
 require "bundler"
 require "json"
+require "timeout"
 
 $LOAD_PATH.unshift(File.expand_path("./lib", __dir__))
 $LOAD_PATH.unshift(File.expand_path("./monkey_patches", __dir__))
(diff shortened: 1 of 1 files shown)


jurre:
I'm not sure if this is the right place to enforce such a timeout.

I have some doubts about this being hard coded into core and for bundler 2 only. It seems like something we should make configurable per ecosystem, and should maybe belong in the system we use to run dependabot-core. 

Not sure if this is doing what it's supposed to do either, the way it's set up now is for every native helper function that we call; there's a 2 minute timeout, and we call these helpers quite a few times during an update run. 

would you mind holding off on this a little bit so we can think it through some more?

xlgmokha:
> would you mind holding off on this a little bit so we can think it through some more?

Sure.

xlgmokha:
Closing in favour of more discussion.

PR #4293, opened 2021-10-07
[WIP] Nodesource cert verification issue

jurre:
As per https://github.com/nodesource/distributions/issues/1266 the
LetsEncrypt DST Root CA X3 has expired last week, and this prevents us
from building the docker image.

Installing `libgnutls30` as suggested in that thread seems to fix it,
but to be entirely honest I am not sure why.

Until the nodesource maintainers suggest a solution, this at least
unblocks container builds, but I'd recommend that we only use this in
local development and not publish it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,6 +48,7 @@ RUN apt-get update \
     libxml2-dev \
     libxmlsec1-dev \
     libgeos-dev \
+    libgnutls30 \
     python3-enchant \
   && locale-gen en_US.UTF-8 \
   && rm -rf /var/lib/apt/lists/*


baseballlover723:
it may be worth clearing the docker cache. I had some issues with it in https://github.com/dependabot/dependabot-core/pull/4250. And I found that a clean docker build or building without the caching enabled worked, so clearing the docker cache might be able to fix the image without having to add anything to the image. Adding `libgnutls30` might be working because it forces the docker to download a fresh copy of nodesource. Not 100% sure if this will work, but its worth a shot

PR #4306, opened 2021-10-11
Upgrade OTP to latest minor: 23.3.4.5

tomaspinho:


jurre:
I'm somewhat hesitant to merge this change, as we've hit a few headaches w/ OTP upgrades in prod, but given this is a minor update it should be safe. I can monitor and revert if needed.

PR #4317, opened 2021-10-14
Add license to image and gemspec

Nishnha:
This adds the LICENSE to the home directory of the Docker image.

Screen Shot 2021-10-14 at 11 30 09 AM

The [dependabot-* gems on rubygems](https://rubygems.org/gems/dependabot-omnibus) currently have the license set as 'nonstandard' and do not include a license file.
~~TBD:~~ `rake gems:release` will **not** include the LICENSE in the `dependabot-*` gems in this PR.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -43,9 +43,6 @@ namespace :gems do
     pkg_path = File.join(root_path, "pkg")
     Dir.mkdir(pkg_path) unless File.directory?(pkg_path)
 
-    license_path = File.join(root_path, "LICENSE")
-    FileUtils.cp(license_path, pkg_path) if File.file?(license_path)
-
     GEMSPECS.each do |gemspec_path|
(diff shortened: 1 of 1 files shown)


Nishnha:
The scope of this PR has been reduced to only including the LICENSE file in the Dependabot-Core Docker image. We will punt on adding the LICENSE to the `dependabot-*` Gems themselves since it's non-trivial with this Nonstandard license.

Since the LICENSE will not be included in the Gems, we will also have to omit the `dependabot-*` Gems from being vendored with `licensed`, but that isn't that big of an issue since Core is owned by us and anyone can look at our required Gems for licensing info.

djrising1234:
Thank you 4 being sk helpful. Truly appreciate it

djrising1234 (commented):
Thanks

PR #4323, opened 2021-10-15
Fix poetry secondary source bug

### LINKED ISSUE #4026: Private repository name gets overwritten in poetry.lock update (even when the update is for a PyPI package)
**Package ecosystem**: Python / poetry
**Package manager version**: Poetry 1.1.7
**Language version**: Python 3.8
**Manifest location and content prior to update**:

_Shortened and redacted pypoetry.toml_:

```
[tool.poetry.dependencies]
python = "^3.8"
environ-config = "^20.1.0"
private-package = { version = "=0.17.0", source = "private-repository" }

[[tool.poetry.source]]
name = "pypi-default"
url = "https://pypi.org/simple/"

[[tool.poetry.source]]
name = "private-repository"
url = "https://redacted-pypi.example.com/simple/"
secondary = true

...
```

_Shortened and redacted poetry.lock_:

```
[[package]]
name = "environ-config"
version = "20.1.0"
description = "Boilerplate-free configuration with env variables."
category = "main"
optional = false
python-versions = "*"

[package.source]
type = "legacy"
url = "https://pypi.org/simple"
reference = "pypi-default"

[[package]]
name = "private-package"
version = "0.17.0"
description = "A package on our private PyPI-alike server."
category = "main"
optional = false
python-versions = ">=3.7,<4.0"

[package.source]
type = "legacy"
url = "https://redacted-pypi.example.com/simple/"
reference = "private-repository"

...
```

**Updated dependency**: environ-config, from 20.1.0 to 21.2.0

**What you expected to see, versus what you actually saw**:
_Expected_
* environ-config and its dependencies have their version numbers updated in their `[[package]]` sections and in `[metadata.files]` in poetry.lock, and also in `[tool.poetry.dependencies]` in pyproject.toml.
* The content-hash in the `[metadata]` section is updated.
* Possibly Poetry finds some other file changes, but nothing else outside `[metadata.files]` or the appropriate `[[package]]` sections are changed.
 
_Observed_
* environ-config and its dependencies have version numbers updated as expected.
* content-hash is updated as expected.
* However, the name of the private repository in poetry.lock is also changed to a 4-digit hex string in the `[package.source]` section for `private-package`:

```
[package.source]
type = "legacy"
url = "https://redacted-pypi.example.com/simple/"
reference = "1bd7"
```

but is not changed in pyproject.toml, resulting in an error in our tests of `Repository "1bd7" does not exist.`.

When run on the pyproject.toml file that Dependabot produces, `poetry lock --no-update` does not change the reference name to a hex string.

----

The hex string looks like it comes from [config_variable_sources in PyprojectPreparer](https://github.com/dependabot/dependabot-core/blob/5ed3f7fcb086b1c71fa3b846bebdaa76b0add879/python/lib/dependabot/python/file_updater/pyproject_preparer.rb#L109), but I'm afraid I don't know Ruby well enough to have searched any further.

Apologies for the heavily redacted manifests - I hope these are useful enough on their own, but can try to produce actual poetry.lock and pyproject.toml files that are small but still reproduce this behaviour if the excerpts above are not enough.

isobelhooper:
Fixes #4026 (hopefully).

`PyprojectPreparer` made sure to get all of the originally-defined Poetry sources (with their original names), and any sources configured based on `python-index` sources in Dependabot's config (with short random hex strings as names).

However, it then used these to generate `poetry.lock` without trying to combine them, so packages in private repos would end up with the short random hex string as their reference in `poetry.lock`, which would not work.

This PR merges any configured sources into the originally-defined Poetry sources in `PyprojectPreparer.replace_sources`, so that the final sources should have the original name they had in `pyproject.toml`.

----

Please let me know if there are any code style or other changes you'd like made to this. Ruby isn't my main language, but I'm happy to change whatever's needed.


jurre:
> Is this something that I should address with more tests, or is it something that you'd have to run yourself?

@isobelhooper apologies for the slow response on this! It would be great to add a few tests to `python/lib/dependabot/python/file_updater/poetry_file_updater.rb`, as we call the changed code [from there](https://github.com/dependabot/dependabot-core/blob/f6e28cd4597ab416fc1fdf7e45df2c2e40c0e830/python/lib/dependabot/python/file_updater/poetry_file_updater.rb#L153-L157). We want to make sure that the updated files do not contain hard-coded secrets after that code has been called. If you happen to have a project I could manually test this against, that'd also be much appreciated!

isobelhooper:
> > Is this something that I should address with more tests, or is it something that you'd have to run yourself?
> 
> @isobelhooper apologies for the slow response on this! It would be great to add a few tests to `python/lib/dependabot/python/file_updater/poetry_file_updater.rb`, as we call the changed code [from there](https://github.com/dependabot/dependabot-core/blob/f6e28cd4597ab416fc1fdf7e45df2c2e40c0e830/python/lib/dependabot/python/file_updater/poetry_file_updater.rb#L153-L157). We want to make sure that the updated files do not contain hard-coded secrets after that code has been called. If you happen to have a project I could manually test this against, that'd also be much appreciated! I can set one up if not though

I'm afraid I don't have a project I could test this against - setting one up would be much appreciated!

I had a try at writing tests for `poetry_file_updater.rb` using Webmock for the PyPI responses, but ran into the problem that the updater calls out to `poetry update` and Webmock can't mock out the requests that it makes. A test project would be good for this.

isobelhooper:
Apologies for the slow response on this - I've pushed another commit with a test in `python/lib/dependabot/python/file_updater/poetry_file_updater.rb` that will, unfortunately, currently not work. It relies on having a Python repository set up to allow username / password `test` / `test` and that hosts the `luigi` package.

I did this using pypiserver and following the instructions in https://pypi.org/project/pypiserver/ to set up a server running on `localhost:8080` that serves no local packages and instead redirects all queries to `pypi.org` (which would then serve the `luigi` package). My test setup just ran pypiserver as a separate process, but there's a [Docker image](https://hub.docker.com/r/pypiserver/pypiserver) that could be used to do the same.

The new test should pass once there's a Python repository like this (with auth credentials) that it can access - though I wouldn't expect it to be on `localhost`, and the credentials could be passed in as secrets

jurre:
> Would it be possible for the Dependabot CI to either set up a temporary Python repository like this just for testing, or use an internal private Python repository?

I'm not sure, both approaches seem fairly brittle and involved 🤔 is there a way that we could mock out those calls instead?

isobelhooper:
> > Would it be possible for the Dependabot CI to either set up a temporary Python repository like this just for testing, or use an internal private Python repository?
> 
> I'm not sure, both approaches seem fairly brittle and involved 🤔 is there a way that we could mock out those calls instead?

Unfortunately those calls are made as part of the `poetry update` command that's called by `PoetryFileUpdater.updated_lockfile_content_for` - we could mock out how Poetry updates lock files, but that also seems brittle and involved.

I'm thinking that it might be possible to move the test to a little earlier in the file updating, and avoid having to make an external call at all.

Since the only thing that `updated_lockfile_content_for` changes about the lockfile once Poetry has generated it is the hash, my thought is that if we make sure that the project file from `PoetryFileUpdater.prepared_pyproject` (which is then used to generate the lockfile) doesn't contain any credentials, then any leaking of credentials into the lock file would be a Poetry bug.

I'll have a go at making a test for `PoetryFileUpdater.prepared_pyproject` in the case where we have auth credentials.

jurre (approved):

jurre:
Thanks for all your work on this!

PR #4324, opened 2021-10-17
Gitlab: add support for creating merge requests for forks

andrcuns:
This PR adds support for creating merge requests from a forked project in to the upstream project.

Because Gitlab doesn't prevent running the pipeline for merge request created within the project like it happens with workflows created by dependabot in github repositories, a malicious dependency would gain access to all environment variables, including keys, secrets etc.

For this purpose users might want to setup a fork of a project and use a less privileged user to run the dependabot-core and create merge requests from forks. This way the pipeline is not gaining access to secrets of the upstream project until user manually reviews the updates and triggers the pipeline as a more privileged user.


jurre:
Overall sounds like this approach makes sense for gitlab, I made some suggestions to avoid adding provider-specific keys to the (already large 😅) list of arguments we pass to these classes

PR #4334, opened 2021-10-20
Do not freeze file-based Poetry dependency version

### LINKED ISSUE #4333: Cannot use Poetry file-based dependencies
I am almost certain that this is the same problem detailed in #1771. The difference is rather than using a path to a directory, I am using a path to a file. Here is the relevant part of my `pyproject.toml`:

```toml
[tool.poetry.dependencies]
geocoding = { file = "here_api/geocoding-7.65.tar.gz" }
```

This produces the following in `poetry.lock`:

```toml
[package.source]
type = "file"
url = "here_api/geocoding-7.65.tar.gz"
```

Dependabot errors with almost the exact same error as #1771:

```
pdater | ERROR <job_223616772> Error processing types-requests (Dependabot::SharedHelpers::HelperSubprocessFailed)
updater | ERROR <job_223616772> RuntimeError
updater | <job_223616772> 
updater | <job_223616772>   The Poetry configuration is invalid:
updater | <job_223616772>     - [dependencies.geocoding] {'file': 'here_api/geocoding-7.65.tar.gz', 'version': '7.65'} is not valid under any of the given schemas
updater | <job_223616772>   
updater | <job_223616772> 
updater | <job_223616772>   at /usr/local/.pyenv/versions/3.8.7/lib/python3.8/site-packages/poetry/core/factory.py:43 in create_poetry
updater | <job_223616772>        39│             message = ""
updater | <job_223616772>        40│             for error in check_result["errors"]:
updater | <job_223616772>        41│                 message += "  - {}\n".format(error)
updater | <job_223616772>        42│ 
updater | <job_223616772>     →  43│             raise RuntimeError("The Poetry configuration is invalid:\n" + message)
updater | <job_223616772>        44│ 
updater | <job_223616772>        45│         # Load package
updater | <job_223616772>        46│         name = local_config["name"]
updater | <job_223616772>        47│         version = local_config["version"]
```

I believe the fix is almost the exact same one-line fix applied in #2361 but also needs to special case `type=="file"`.

**Package ecosystem**
`pip`
**Package manager version**
Poetry 1.1.8
**Language version**
Python 3.8.7
poetry.lock, pyproject.toml

**dependabot.yml content**

```yaml
version: 2

registries:
  private-pypi:
    type: python-index
    url: https://pypi.myrepo.com
    username: user
    password: ${{secrets.PYPI_PASS}}

updates:
  - package-ecosystem: "pip"
    directory: "/"
    insecure-external-code-execution: allow
    registries:
      - private-pypi
    schedule:
      interval: daily
    reviewers:
      - "someteam"
```

**Updated dependency**
N/A
Poetry updating to work
N/A, not exactly doing an update

phillipuniverse:
Fixes #4333. This was modeled after #2361 with the same type of tests.

I also renamed the test files to differentiate between directory-based and file-based dependencies, both of which can be specified by a path.


jurre:
> Do you think we need to do anything about the url dependencies or do you think those will be covered?

Tbh, I am not sure! Maybe we can add a test for it?

> What's the process/timeline for this fix being included in the official GitHub Dependabot?

Once it's merged I can cut a release fairly quickly and can roll it out usually the same or next day.

jurre:
There's a couple of failing python tests on CI @phillipuniverse, would you mind taking a look at those?

phillipuniverse:
ok! @jurre I feel pretty good about this now, sorry again for how long this whole thing has taken.

New updates:

- URL dependencies had the same problem and were incorrectly adding a `version` attribute, I added a test and whitelisted it
- Added more positive assertions for directory and file dependencies -- essentially make sure the dependency is actually there but the dependency + version is not

Thanks for your help getting this over the line! We should probably squash merge this when it's all done.

phillipuniverse:
@jurre last thing here - I have 1 failing test locally and I'm not sure why:

image

But even with my change in `python/lib/dependabot/python/file_updater/pyproject_preparer.rb` commented out it still fails. And I even checked out `main` and ran it and still fails so... 🤷 hopefully it won't fail in CI!

jurre:
>  I have 1 failing test

I am currently working on upgrading pipenv so I think I can tackle that failure, it doesn't seem related to these changes

PR #4335, opened 2021-10-20
Treat GHES hosted sources as github sources

jurre:
When a dependency is hosted on GHES, previously it was not treated as a
GitHub source, meaning that we would not check for releases/changelogs
etc when requesting Metadata for the PR.

This fixes that, by first parsing the URL, and then making a request to
`<host>/status`, and checking for a `X-GitHub-Request-Id` header, which
we return from GitHub Enterprise Server.


jurre:
I need to fix up a bunch of tests that now attempt to check if the unknown host is GHES, on it 👍

PR #4342, opened 2021-10-22
Cargo: pass permit_copy_rename environment variable

jurre:
Resolves an issue with rustup copying files during updates.

https://rust-lang.github.io/rustup/environment-variables.html:

When set, allows rustup to fall-back to copying files if attempts to rename
result in an cross-device link errors. These errors occur on OverlayFS,
which is used by Docker.

### DIFF
--- a/cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
+++ b/cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
@@ -19,6 +19,7 @@ module Dependabot
         OBJECT_PATTERN = /object not found - no match for id \(.*\)/.freeze
         REF_NOT_FOUND_REGEX =
           /#{UNABLE_TO_UPDATE}.*(#{REVSPEC_PATTERN}|#{OBJECT_PATTERN})/m.freeze
+        CARGO_DEFAULT_ENV = { "RUSTUP_PERMIT_COPY_RENAME" => "1" }.freeze
 
         def initialize(dependency:, credentials:,
                        original_dependency_files:, prepared_dependency_files:)
@@ -133,7 +134,7 @@ module Dependabot
         def run_cargo_command(command)
           start = Time.now
           command = SharedHelpers.escape_command(command)
-          stdout, process = Open3.capture2e(command)
+          stdout, process = Open3.capture2e(CARGO_DEFAULT_ENV, command)
           time_taken = Time.now - start
 
           # Raise an error with the output from the shell session if Cargo


jorendorff:
Some background here: I noticed Dependabot seems to miss most dependency updates for github/blackbird.

@jurre and I looked into this a few weeks ago and it seems the bug is something like this:

-    Dependabot grabs the user code and tries various Cargo commands to see what can be updated.
-    But that project has a `rust-toolchain` file that specifies `stable`, so every `cargo` command checks to make sure we have a reasonable recent Rust stable toolchain, which I guess Dependabot's build image doesn't.
-    So `cargo` ends up trying to install the latest Rust from the `stable` channel, as if by `rustup update stable`.
-    But this fails because `rustup` requires a [special magic environment variable](https://rust-lang.github.io/rustup/environment-variables.html) in order to work in Docker (because of Docker OverlayFS specialness).

I don't know if this will work because I don't know what it means in the Rustup docs that this environment variable is "unstable".

jurre:
> make sure we have a reasonable recent Rust stable toolchain

@jorendorff this is what I am still a little confused about, because:

```
[dependabot-core-dev] ~/dependabot-core $ rustup --version
rustup 1.24.3 (ce5817a94 2021-05-31)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: The currently active `rustc` version is `rustc 1.51.0 (2fd73fabe 2021-03-23)`
[dependabot-core-dev] ~/dependabot-core $ rustup update stable
info: syncing channel updates for 'stable-x86_64-unknown-linux-gnu'

  stable-x86_64-unknown-linux-gnu unchanged - rustc 1.56.1 (59eed8a2a 2021-11-01)

info: checking for self-updates
```

Also looking at https://github.com/rust-lang/rustup/tree/stable is seems we _are_ on the stable version

jorendorff:
Hmm. Did this change recently? Our repo now shows [a different error](https://github.com/github/blackbird/network/updates/234948484), which I can troubleshoot separately.

No matter what Dependabot's build image has, it's likely that `cargo` commands will want to install other versions sometimes, because

- a `rust-toolchain` file can specify any version of the compiler;
- probably after a new version of Rust is released, there is some delay before a new Dependabot build image is created.

jorendorff:
The new error turned out to be easy to fix. Blackbird is back on the Dependabot train.

jurre:
Seems to have been resolving by updating the cargo version, so going to close this out

PR #4352, opened 2021-10-29
[Azure] Check & Raise TagsCreationForbidden Exception

AlekhyaYalla:
**Issue:** 
Today, while creating pull requests(PR), labels are passed for Azure [create_pull_request method](https://github.com/dependabot/dependabot-core/blob/d734dc6e4fe0f7903ee97e0079ac9f5362e408c1/common/lib/dependabot/pull_request_creator/azure.rb#L81) & created by default.(may it be default/security etc)
In one of our repo, the user who tries to create a dependency update PR, has permissions to create a PR in that repo, but not for the tags. So the PR creation fails, as dependabot-core tries to create tags first and then PR. As the tags creation step is failing with 403 forbidden errors, PR creation also failed. 

**Fix:**
- As Tags creation is failed and the same has been clearly mentioned in the error message, instead of raising a regular 403: Forbidden exception, we can raise TagsCreationForbidden exception. For this, we can validate the error code in the message & raise the same. This exception is only seen in Azure, so handled in the Azure only.

### DIFF
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -266,8 +266,10 @@ module Dependabot
         end
 
         raise Unauthorized if response.status == 401
+
         if response.status == 403
           raise TagsCreationForbidden if tags_creation_forbidden?(response)
+
           raise Forbidden
         end
         raise NotFound if response.status == 404
@@ -317,8 +319,9 @@ module Dependabot
 
       def tags_creation_forbidden?(response)
         return if response.body.empty?
-        message = JSON.parse(response.body).fetch('message', nil)
-        message && message.include?("TF401289: The current user does not have permissions to create tags")
+
+        message = JSON.parse(response.body).fetch("message", nil)
+        message&.include?("TF401289: The current user does not have permissions to create tags")
       end
 
       attr_reader :auth_header
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -321,7 +321,7 @@ module Dependabot
         return if response.body.empty?
 
         message = JSON.parse(response.body).fetch("message", nil)
-        message&.include?("TF401289: The current user does not have permissions to create tags")
+        message&.include?("TF401289")
       end
 
       attr_reader :auth_header
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -18,6 +18,8 @@ module Dependabot
 
       class Forbidden < StandardError; end
 
+      class TagsCreationForbidden < StandardError; end
+
       RETRYABLE_ERRORS = [InternalServerError, BadGateway, ServiceNotAvailable].freeze
 
       MAX_PR_DESCRIPTION_LENGTH = 3999
@@ -264,7 +266,10 @@ module Dependabot
         end
 
         raise Unauthorized if response.status == 401
-        raise Forbidden if response.status == 403
+        if response.status == 403
+          raise TagsCreationForbidden if tags_creation_forbidden?(response)
+          raise Forbidden
+        end
         raise NotFound if response.status == 404
 
         response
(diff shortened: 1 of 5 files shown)


jurre:
> So instead of creating a new object of pull_request_creator, I thought azure_creator can be created based on a new field/param that would talk about if this creation need labels or not.

I'm not sure that I really understand the difference or at least why one is better over the other?

> And also, I thought custom_labels is meant for passing if the user has any custom labels for the PR. Using it in this case may sound a little ambigous

While I agree that the naming is not as clear as it could be, the functionality is already there to create PRs without any labels, I think having both the option of passing an empty array to `custom_labels` _and_ also having the `labels_required` option leaves us worse-off?

AlekhyaYalla:
Hi @jurre 
Based on the discussion we had on chat, concluding the behaviour to attempt labels creation in PullRequestCreator can be achieved by passing `[]` to the customLabels. Removed that part.
Added `TagsCreationForbidden` in the azure client as it is a valid case.

common/lib/dependabot/pull_request_creator.rb
```diff
@@ -60,7 +60,7 @@ def initialize(source:, base_commit:, dependencies:, files:, credentials:,
                    label_language: false, automerge_candidate: false,
                    github_redirection_service: DEFAULT_GITHUB_REDIRECTION_SERVICE,
                    custom_headers: nil, require_up_to_date_base: false,
-                   provider_metadata: {}, message: nil)
+                   provider_metadata: {}, message: nil, labels_required: true)
```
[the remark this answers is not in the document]
AlekhyaYalla: Removed the param `labels_required`

common/spec/dependabot/clients/azure_spec.rb
```diff
@@ -181,6 +181,39 @@
...
+                                 "", [], nil)
+    end
+
+    let(:pull_request_url) { repo_url + "/pullrequests?api-version=5.0" }
+
+    context "when response is 403 & tags creation is forbidden" do
+      before do
+        stub_request(:post, pull_request_url).
+          with(basic_auth: [username, password]).
+          to_return(status: 403, body: JSON.unparse({ message: "TF401289: The current user does not have permissions to create tags" }))
```
jurre, line 196:
```suggestion
          to_return(
            status: 403,
            body: { message: "TF401289: The current user does not have permissions to create tags" }.to_json
          )
```

The linter is currently failing because this line is too long, and the `.to_json` is a bit more in line with the rest of the codebase.

common/lib/dependabot/clients/azure.rb
```diff
@@ -310,6 +317,13 @@ def truncate_pr_description(pr_description)
         pr_description.force_encoding(Encoding::UTF_8)
       end
 
+      def tags_creation_forbidden?(response)
+        return if response.body.empty?
+
+        message = JSON.parse(response.body).fetch("message", nil)
+        message&.include?("TF401289: The current user does not have permissions to create tags")
```
jurre, line 324: Do you think it would be sufficient to just match on `TF401289`, that way if the copy in that error ever changes we don't need to patch this code, and it seems like this is a code that specifically represents this tags permission error?

common/spec/dependabot/clients/azure_spec.rb
AlekhyaYalla: Sure will be addressing this.

common/lib/dependabot/clients/azure.rb
AlekhyaYalla: Seems like that works, as `TF401289` is the message code for the permission tags issue. Will update the fix

AlekhyaYalla:
> @AlekhyaYalla would you mind updating the PR title / description to reflect the changes we landed on in this PR, that way, if we ever go back to figure out why this code was changed, the PR will reflect the relevant background info

Updated @jurre My apologies for not updating.

jurre (approved):

PR #4356, opened 2021-11-01
Ensure we cleanup tmp directories after use

brrygrdn:
In very long-running jobs with a lot of dependencies, it is feasible we can loop through the `in_a_temporary_directory` step hundreds of times.

This change ensures we clean up these directories after use in an effort to reclaim disk space in all cases and to avoid problems in persistent run environments such as the `bin/docker-dev-shell` environment.


jurre (approved):
Nice one 👍

PR #4362, opened 2021-11-02
Add timeout per operation

xlgmokha:
This change introduces an experiment that allows configuring a timeout per
operation to prevent spawned processes from running indefinitely.

### DIFF
--- a/bundler/lib/dependabot/bundler/native_helpers.rb
+++ b/bundler/lib/dependabot/bundler/native_helpers.rb
@@ -6,21 +6,46 @@ require "dependabot/shared_helpers"
 module Dependabot
   module Bundler
     module NativeHelpers
+      class BundleCommand
+        attr_reader :timeout_seconds
+
+        def initialize(timeout_seconds)
+          @timeout_seconds = timeout_seconds
+        end
+
+        def build(script_path)
+          [
+            timeout_command,
+            :bundle, :exec, :ruby,
+            script_path
+          ].compact.join(" ")
+        end
+
+        private
+
+        def timeout_command
+          "timeout -s HUP #{timeout_seconds}" if timeout_seconds
+        end
+      end
+
       def self.run_bundler_subprocess(function:, args:, bundler_version:, options: {})
         # Run helper suprocess with all bundler-related ENV variables removed
         bundler_major_version = bundler_version.split(".").first
+        helpers_path = versioned_helper_path(bundler_version: bundler_major_version)
         ::Bundler.with_original_env do
+          command = BundleCommand.
+                    new(options[:timeout_per_operation_seconds]).
+                    build(File.join(helpers_path, "run.rb"))
           SharedHelpers.run_helper_subprocess(
-            command: helper_path(bundler_version: bundler_major_version,
-                                 timeout_seconds: options[:timeout_per_operation_seconds]),
+            command: command,
             function: function,
             args: args,
             env: {
               # Bundler will pick the matching installed major version
               "BUNDLER_VERSION" => bundler_version,
-              "BUNDLE_GEMFILE" => File.join(versioned_helper_path(bundler_version: bundler_major_version), "Gemfile"),
+              "BUNDLE_GEMFILE" => File.join(helpers_path, "Gemfile"),
               # Prevent the GEM_HOME from being set to a folder owned by root
-              "GEM_HOME" => File.join(versioned_helper_path(bundler_version: bundler_major_version), ".bundle")
+              "GEM_HOME" => File.join(helpers_path, ".bundle")
             }
           )
         rescue SharedHelpers::HelperSubprocessFailed => e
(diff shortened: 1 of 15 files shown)


bundler/lib/dependabot/bundler/native_helpers.rb
```diff
@@ -35,8 +36,9 @@ def self.versioned_helper_path(bundler_version:)
         File.join(native_helpers_root, native_helper_version)
       end
 
-      def self.helper_path(bundler_version:)
-        "bundle exec ruby #{File.join(versioned_helper_path(bundler_version: bundler_version), 'run.rb')}"
+      def self.helper_path(bundler_version:, timeout_seconds: nil)
+        prefix = "timeout -s HUP #{timeout_seconds} " if timeout_seconds
+        "#{prefix}bundle exec ruby #{File.join(versioned_helper_path(bundler_version: bundler_version), 'run.rb')}"
```
xlgmokha, line 41: How about?

```ruby
      def self.helper_path(bundler_version:, timeout_seconds: nil)
        [
          timeout_seconds ? "timeout -s HUP #{timeout_seconds}" : nil,
          "bundle exec ruby",
          File.join(versioned_helper_path(bundler_version: bundler_version),
          "run.rb"
        ].compact.join(" ")
      end
```

xlgmokha: or...

```ruby
      def self.helper_path(bundler_version:, timeout_seconds:)
        [
          "timeout -s HUP #{timeout_seconds || MAX_TIMEOUT_SECONDS}",
          "bundle exec ruby",
          File.join(versioned_helper_path(bundler_version: bundler_version),
          "run.rb"
        ].compact.join(" ")
      end
```

bundler/spec/dependabot/bundler/native_helpers_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    context "with a timeout provided" do
+      it "terminates the spawned process when the timeout is exceeded" do
+        allow(Dependabot::SharedHelpers).to receive(:run_helper_subprocess)
+
+        subject.run_bundler_subprocess(
+          function: "noop",
+          args: [],
+          bundler_version: "2.0.0",
+          options: {
+            timeout_per_operation_seconds: 1
```
brrygrdn, line 19: I think historically we have used the experiments hash as effectively a feature flag, so I was surprised to see this passed as an integer value.

I think this approach is fine though, but in the spirit of checking your inputs, should we place a hard cap on the value passed in the event the calling code makes a magnitude error and passes in 'milliseconds' instead of 'seconds' or something like that?

xlgmokha: > should we place a hard cap on the value passed in the event the calling code makes a magnitude error and passes in 'milliseconds' instead of 'seconds' or something like that?

Good point. How does an upper bound of 30 minutes and a lower bound of 1 minute sound?

bundler/lib/dependabot/bundler/native_helpers.rb
```diff
@@ -24,7 +27,17 @@ def build(script_path)
...
-          "timeout -s HUP #{timeout_seconds}" if timeout_seconds
+          "timeout -s HUP #{timeout_seconds}" unless timeout_seconds.zero?
+        end
+
+        def adjust(seconds)
+          return 0 unless seconds
+
+          seconds = seconds.to_i
+          return MAX_SECONDS if seconds > MAX_SECONDS
+          return MIN_SECONDS if seconds < MIN_SECONDS
```
landongrindheim, line 34: You might dig [`Comparable#clamp`](https://docs.ruby-lang.org/en/2.5.0/Comparable.html#method-i-clamp)

```plaintext
irb(main):001:0> max = 100
=> 100
irb(main):002:0> min = 1
=> 1
irb(main):003:0> 150.clamp(min, max)
=> 100
irb(main):004:0> 0.clamp(min, max)
=> 1
irb(main):005:0> 37.clamp(min, max)
=> 37
```

It's not the most popular feature. If it feels unintuitive, what you're doing is probably preferable 🙂

xlgmokha: TIL: [Comparable#clamp](https://ruby-doc.org/core-3.0.2/Comparable.html#method-i-clamp)

landongrindheim (commented):
The code looks good to me. I hesitate to approve as I could use more context re: the overall workflow. I'm happy to run through this with you to get to that point 😄

xlgmokha:
> The code looks good to me. I hesitate to approve as I could use more context re: the overall workflow. I'm happy to run through this with you to get to that point 😄

Thanks Landon.

lseppala (approved):
Did a pair-review of this work with @xlgmokha. Covered the following topics:

* How the timeout options flows (and options in general)
* Whether `SIGHUP` was safe to use a timeout signal
  * Alternatives discussed. No other HUP signal handler was found (for example, like a configuration reload handler) in any dependencies or processes. Process is not run in a terminal session which is expected to close & for the process to continue.
* Min & max timeout values

Given this timeout is disabled by default & tests pass as expected, approving this work. Nice stuff! 🎉

PR #4393, opened 2021-11-10
Bump to go 1.17.3

jeffwidman:
There were some minor fixes to the `go` command in `1.17.2`/`1.17.3`... 

I don't think any of them directly affect dependabot, but it
doesn't hurt to bump this and makes it so the next person has a smaller
diff to look at when they consider updating.

I also ran the `script/extract` in `gomodules-extracted` against `1.17.3` locally, and didn't see any changes other than a very minor build file change which I suspect is simply due to building on my Mac, so that should keep working as-is. It's important to keep the code in `gomodules-extracted`  in-sync with the version of `go` used here.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -167,8 +167,8 @@ USER root
 ### GO
 
 # Install Go
-ARG GOLANG_VERSION=1.17.1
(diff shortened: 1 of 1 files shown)


jeffwidman:
Test failure looks completely unrelated?

mctofu (approved):
Thanks! I fixed the failing test in https://github.com/dependabot/dependabot-core/pull/4399.

PR #4401, opened 2021-11-11
Ignore errors from Source enterprise check and ignore known failures

jurre:
We check if a potential Source is GitHub enterprise by making a request
to a `/status` endpoint against the root URL and checking some headers.

We've observed this check failing in some cases when the source is not
enterprise, and we get rate limited, or otherwise the request fails with
an error.

In this case we do not want to block creating a PR, but instead we
should assume the source is not Enterprise.

This also adds a list of known hosts that we come across often that
definitely are not GitHub Enterprise instances, and we ignore those and
don't bother making a request to them.

### DIFF
--- a/common/lib/dependabot/source.rb
+++ b/common/lib/dependabot/source.rb
@@ -45,6 +45,8 @@ module Dependabot
       (?:#{AZURE_SOURCE})
     /x.freeze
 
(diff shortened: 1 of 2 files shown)


brrygrdn (approved):
I paired with @jurre on debugging this issue so I would normally refrain from approving, however this change is fairly non-controversial and addresses a bug that is actively affecting customers.

PR #4405, opened 2021-11-12
Terraform: Improve file updater and registry request error handling

jurre:
This PR addresses two issues:

Dependabot sometimes comes across terraform modules that need to be installed, where terraform will hint at this with a message like: `This module is not yet installed. Run "terraform init" to install all modules`. We will then dutifully run that command and retry, and typically the `terraform providers lock` command will succeed at that point.

However, when that command fails we would previously try to run `terraform providers lock` again, and spit out the full error to users, which on our end would be handled as an unknown error.

The change introduced here is to report any failures for `terraform init` as resolvability failures, and the same goes if the command exits successfully, but the next run of `terraform providers lock` gives us the same `This module is not yet installed` error.

The other error we address is networking failures when attempting to reach a dependencies provider (terraform registry), we rescue these and raise an appropriate message.

### DIFF
--- a/terraform/lib/dependabot/terraform/file_updater.rb
+++ b/terraform/lib/dependabot/terraform/file_updater.rb
@@ -12,6 +12,7 @@ module Dependabot
       include FileSelector
 
       PRIVATE_MODULE_ERROR = /Could not download module.*code from\n.*\"(?<repo>\S+)\":/.freeze
+      MODULE_NOT_INSTALLED_ERROR =  /Module not installed.*module\s*\"(?<mod>\S+)\"/m.freeze
 
       def self.updated_files_regex
         [/\.tf$/, /\.hcl$/]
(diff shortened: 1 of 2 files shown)


terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -122,6 +123,10 @@ def update_lockfile_declaration(updated_manifest_files) # rubocop:disable Metric
             content.sub!(declaration_regex, updated_dependency)
           end
         rescue SharedHelpers::HelperSubprocessFailed => e
+          if @retrying_lock && e.message.match?(MODULE_NOT_INSTALLED_ERROR)
+            mod = e.message.match(MODULE_NOT_INSTALLED_ERROR).named_captures.fetch("mod")
```
landongrindheim, line 127: Great use of named captures here ✨

landongrindheim (approved):
:shipit:

Issue #4288, opened 2021-10-06
Dependabot takes hours for a common dependency upgrade in a mono repo.


AlekhyaYalla:
**Issue:**
One of the issues we observed in dependency updates, is Dependabot taking a very huge amount of time in a mono repo for updating a dependency which is populated in more no of packages. In our case, we have a mono repo with yarn as a package manager and ~600 packages in it. _React_ is one of the commonly added dependency in ~350 packages in this repo (both dependency & devDependency). While we try to update _react_ using dependabot in this repo, it is taking nearly ~4hrs to finish the update task. The usual time by dependabot for doing an update of an irregular dependency in this repo is ~15-20min. The below are the observations from the investigation we did for common dependency.

**Observations:**
-	While generating the updated lock file content in this [method](https://github.com/dependabot/dependabot-core/blob/main/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb#:~:text=def-,updated_yarn_lock,-(yarn_lock)), it is passing the dependencies to-be-updated along with the requirements (These are the package.jsons that has the dependency in it, either devDependency or dependency).
-	This [method ](https://github.com/dependabot/dependabot-core/blob/main/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb)calls the respective package manager function (in our case, it is yarn) to update the dependencies. This [method ](https://github.com/dependabot/dependabot-core/blob/main/npm_and_yarn/helpers/lib/yarn/updater.js#:~:text=async%20function-,updateDependencyFiles,-(directory%2C%20dependencies)%20%7B)runs a loop to run the commands add & install in each and every package.json to autogenerate the lock file content. 
-	With our investigations, it is evident, that the time taken to update a dependency in the mono repo is directly proportional to the number of package.jsons that dependency presents in.

**Wanted to know:**
-	In the case of mono repo, all these packages.jsons would be updating to the same version and having a common lock file in the repo level, is it okay to run a single command in lock file level/repo level and autogenerate the content of it?
-	Are there anything else contributing to the longer runs in this area (apart from investigations)?
-	If this is a known issue, are there any plans/planned to accommodate this?
-	Is it possible to optimize these commands in Dependabot-core using  “yarn workspace” or any mono repo tool, that would help in running these commands efficiently at the repo level, instead of running them in each and every package level?

@jurre @hmarr @feelepxyz Please share your thoughts on this

[...]

PR #4409, opened 2021-11-15
Run YarnUpdate only once for a version requirement

AlekhyaYalla:
Issue: https://github.com/dependabot/dependabot-core/issues/4288

Currently, For a mono repo, when [UpdateDependencyFiles ](https://github.com/dependabot/dependabot-core/blob/d734dc6e4fe0f7903ee97e0079ac9f5362e408c1/npm_and_yarn/helpers/lib/yarn/updater.js#L103)is called for generating the lock file content, dependencies param is passed to it. Dependencies param contains the dep that needs to be updated in all required workspaces in the repo. 

As far as observed, the multiple package.json (multiple requirements) for a dependency are getting passed for a mono repo., where it does have only a single yarn.lock file. 
In this case, it's not necessary to run `yarn update` for each and every package.json or requirement if all are referring to same version.
For a single version, if run once, yarn.lock would be generated, and running the same version `yarn update` would result in repetition of the task.
In case if each package.json is referring to different versions of that dependency, this code would handle that by running `yarn update` accordingly.

### DIFF
--- a/npm_and_yarn/helpers/lib/yarn/updater.js
+++ b/npm_and_yarn/helpers/lib/yarn/updater.js
@@ -107,8 +107,7 @@ async function updateDependencyFiles(directory, dependencies) {
   let requiredVersions = [];
   for (let dep of dependencies) {
     for (let reqs of dep.requirements) {
-      if(requiredVersions.indexOf(reqs.requirement) > -1)
-      {
+      if (requiredVersions.indexOf(reqs.requirement) > -1) {
           continue;
       }
       updateRunResults = Object.assign(
--- a/npm_and_yarn/helpers/lib/yarn/updater.js
+++ b/npm_and_yarn/helpers/lib/yarn/updater.js
@@ -104,16 +104,20 @@ async function updateDependencyFiles(directory, dependencies) {
   const readFile = (fileName) =>
     fs.readFileSync(path.join(directory, fileName)).toString();
   let updateRunResults = { "yarn.lock": readFile("yarn.lock") };
-
+  let requiredVersions = [];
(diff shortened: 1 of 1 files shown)


jurre:
Thanks @AlekhyaYalla, could you add some context on what this change does and how it works exactly?

npm_and_yarn/helpers/lib/yarn/updater.js
```diff
@@ -104,16 +104,20 @@ async function updateDependencyFiles(directory, dependencies) {
   const readFile = (fileName) =>
     fs.readFileSync(path.join(directory, fileName)).toString();
   let updateRunResults = { "yarn.lock": readFile("yarn.lock") };
-
+  let requiredVersions = [];
   for (let dep of dependencies) {
     for (let reqs of dep.requirements) {
+      if(requiredVersions.indexOf(reqs.requirement) > -1)
+      {
```
jurre, line 111:
```suggestion
      if (requiredVersions.indexOf(reqs.requirement) > -1) {
```

We should probably set up a linter that checks this, but this is a but more in line with the style of the rest of the file.

AlekhyaYalla:
> Thanks @AlekhyaYalla, could you add some context on what this change does and how it works exactly?

@jurre Updated the description of the PR. My bad I missed it earlier. I hope the description is setting some context about the fix

> Is there any additional test coverage that you think we could introduce around this?

Yes, I will add additional unit test cases for this change.

AlekhyaYalla:
@jurre Could you please review this change and suggest if any changes are required on this?

jurre (approved):

PR #4417, opened 2021-11-16
Maven: Correctly handle nested declarations

jurre:
We've observed errors around pom files that have a nested `plugin`
section inside an outer plugin declaration.

Previously we used a regex to recursively scan the XML document, which
caused the inner `plugin` declarations closing tag to match the regex,
resulting in an incomplete XML section.

To demonstrate, given the following XML:

```xml
<plugins>
  <plugin>
    <configuration>
      <jvmTarget>11</jvmTarget>
      <compilerPlugins>
        <plugin>spring</plugin>
      </compilerPlugins>
    </configuration>
    <groupId>org.jetbrains.kotlin</groupId>
    <artifactId>kotlin-maven-plugin</artifactId>
    <version>${kotlin.version}</version>
  </plugin>
</plugin>
```

The `<plugin>spring</plugin>` declaration would cause a regex match,
resulting in the required information (the `version` in this case) to be
omitted from the XML snippet.

This is resolved by using Nokogiri to traverse the XML instead of using
regular expressions, and selecting the nodes by name.

### DIFF
--- a/maven/lib/dependabot/maven/update_checker/property_updater.rb
+++ b/maven/lib/dependabot/maven/update_checker/property_updater.rb
@@ -105,7 +105,7 @@ module Dependabot
             dependency: dep,
             declaring_requirement: declaring_requirement,
             dependency_files: dependency_files
-          ).declaration_nodes.first.at_css("version")&.content
+          ).declaration_nodes.first.at_xpath("./*/version")&.content
         end
 
         def pom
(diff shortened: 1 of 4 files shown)


maven/lib/dependabot/maven/update_checker/property_updater.rb
```diff
@@ -105,7 +105,7 @@ def version_string(dep)
             dependency: dep,
             declaring_requirement: declaring_requirement,
             dependency_files: dependency_files
-          ).declaration_nodes.first.at_css("version")&.content
+          ).declaration_nodes.first.at_xpath("./*/version")&.content
```
jurre, line 108: Using `at_css` will match the lexical first instance of `version`, which can match a nested property

brrygrdn (approved):
Very good find, this is definitely a preferable approach 🚀

PR #4421, opened 2021-11-16
Use GitHub-flavored Markdown as the output from Pandoc

landongrindheim:
Per Pandoc's docs[^1]
> Pandoc’s parsers can exhibit pathological performance on some corner
  cases. It is wise to put any pandoc operations under a timeout, to avoid
  DOS attacks that exploit these issues. If you are using the pandoc
  executable, you can add the command line options +RTS -M512M -RTS (for
  example) to limit the heap size to 512MB. Note that the commonmark
  parser (including commonmark_x and gfm) is much less vulnerable to
  pathological performance than the markdown parser, so it is a better
  choice when processing untrusted input.

[^1]: https://pandoc.org/MANUAL.html

### DIFF
--- a/common/lib/dependabot/metadata_finders/base/changelog_finder.rb
+++ b/common/lib/dependabot/metadata_finders/base/changelog_finder.rb
@@ -48,7 +48,7 @@ module Dependabot
             PandocRuby.convert(
               pruned_text,
               from: :rst,
-              to: :markdown,
+              to: :gfm,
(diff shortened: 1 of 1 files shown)


landongrindheim:
Closing this as we're going to move away from relying on `PandocRuby` ([#4427](https://github.com/dependabot/dependabot-core/pull/4427#issuecomment-974136609))

PR #4427, opened 2021-11-18
Add pandoc to dockerfile

landongrindheim:
## Summary

PandocRuby is used to convert RestructuredText (rst), a markdown-like
format widely used in the Python ecosystem, to markdown. We recently
noticed new errors surfacing around Pandoc and started to investigate.
This led to the discovery that Pandoc was not installed in the Docker
container GitHub is using to run Dependabot against repositories.

There was error handling to make this hard to discover, and rst is
similar enough to markdown that there were no reported problems.
Since the conversion was intentional at some point in the past, and
since we can't _know_ that users of `dependabot-core` rely on Pandoc one
way or another, I'm opting to add it to the Docker image.

---

### Cleanup

I made two additional changes while adding Pandoc to the image:
* Localizing the Python-specific comment: It doesn't seem that any packages were included to the wrong place since the comment was added, but I thought the potential for it was high.
* Sorting the list of packages to install: Actually, I chose to sort the lists separately. This builds off the previous idea of keeping Python-specific packages grouped. Sorting lists is included as a [Dockerfile best practice](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#sort-multi-line-arguments). I don't see the downside outside of surprises that might come from install order. I would expect our test suite to surface errors related to that, but it's not guaranteed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,9 +8,6 @@ ENV DEBIAN_FRONTEND="noninteractive" \
   LC_ALL="en_US.UTF-8" \
   LANG="en_US.UTF-8"
 
-# Everything from `make` onwards in apt-get install is only installed to ensure
-# Python support works with all packages (which may require specific libraries
-# at install time).
 RUN apt-get update \
   && apt-get upgrade -y \
   && apt-get install -y --no-install-recommends \
@@ -32,6 +29,8 @@ RUN apt-get update \
     openssh-client \
     software-properties-common \
     make \
+# All packages from here onward are installed to ensure the Python ecosystem
+# is properly supported as Python packages may depend on them.
     libpq-dev \
     libssl-dev \
     libbz2-dev \
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,6 +49,7 @@ RUN apt-get update \
     libxmlsec1-dev \
     libgeos-dev \
     python3-enchant \
+    pandoc \
   && locale-gen en_US.UTF-8 \
   && rm -rf /var/lib/apt/lists/*
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,43 +12,43 @@ RUN apt-get update \
   && apt-get upgrade -y \
   && apt-get install -y --no-install-recommends \
     build-essential \
-    dirmngr \
-    git \
     bzr \
-    mercurial \
-    gnupg2 \
     ca-certificates \
     curl \
+    dirmngr \
     file \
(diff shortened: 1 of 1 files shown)


jurre:
I am not sure that we should do this:

```
root@64ee852953ab:/# apt install pandoc
# snip
Need to get 15.5 MB of archives.
After this operation, 125 MB of additional disk space will be used.
```

Our image is already pretty big, and we've been running the service without converting anything to pandoc for a long time now and have not seen any issues with changelogs not being rendered nicely.

landongrindheim:
> I am not sure that we should do this:
> 
> ```
> root@64ee852953ab:/# apt install pandoc
> # snip
> Need to get 15.5 MB of archives.
> After this operation, 125 MB of additional disk space will be used.
> ```
> 
> Our image is already pretty big, and we've been running the service without converting anything to pandoc for a long time now and have not seen any issues with changelogs not being rendered nicely.

@jurre I totally get where you're coming from. I'm trying to balance two concerns, about which I'd like to hear your thoughts:
* Keeping our image from bloating further
* Not breaking current functionality

If we favor keeping our image from bloating, the move that I see as appropriate is removing the calls to `PandocRuby`. The downside is that I don't know whether anyone relies on this being present. Further, I don't know what the created diffs are rendered like outside of GitHub. I'm wary of breaking users' workflows and providing a degraded experience. 

If we favor not breaking functionality, which I've done here, the downside is what you've called out: we're increasing the image size by a non-trivial amount. This probably affects more users, and may provide a degraded experience. Unfortunately, I don't think either approach gives us a clear win.

I know we're at a pre-1.0 release, but I want to be cautious. I'd love to hear your thoughts. How do you think about these trade-offs?

jurre:
@landongrindheim yeah that's a fair concern, the way I see it is we're not really breaking functionality given this has never worked for the GH-native version of dependabot, we've been silently ignoring those errors.

I would also assume that no-one else is relying on the functionality, since running dependabot without the dependabot-core image is pretty badly supported, so unless folks were inheriting from that image in order to add pandoc, they're likely not using the functionality.

Given that the conversion from rst to HTML happens entirely within dependabot-core using CommonMark, I think we're mostly safe by removing the calls to PandocRuby.

landongrindheim:
> I would also assume that no-one else is relying on the functionality, since running dependabot without the dependabot-core image is pretty badly supported

@jurre Thanks for this context. I'll close this PR in favor of removing the reliance on `PandocRuby` 😄

PR #4432, opened 2021-11-19
Add support for custom commit message trailers

andrcuns:
Some platforms like Gitlab use git trailers to generate changelog entries from commit messages.

This PR adds option to pass in custom git trailer via `commit_message_options` hash in a completely backwards compatible way.

### DIFF
--- a/common/lib/dependabot/pull_request_creator/message_builder.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder.rb
@@ -142,6 +142,19 @@ module Dependabot
       end
 
       def message_trailers
+        return unless signoff_trailers || custom_trailer
+
+        [signoff_trailers, custom_trailer].compact.join("\n")
+      end
+
+      def custom_trailer
+        trailer = commit_message_options[:trailer]
+        return unless trailer.is_a?(String)
+
+        trailer
+      end
+
+      def signoff_trailers
         return unless on_behalf_of_message || signoff_message
 
         [on_behalf_of_message, signoff_message].compact.join("\n")
--- a/common/lib/dependabot/pull_request_creator/message_builder.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder.rb
@@ -142,16 +142,17 @@ module Dependabot
       end
 
       def message_trailers
-        return unless signoff_trailers || custom_trailer
+        return unless signoff_trailers || custom_trailers
 
-        [signoff_trailers, custom_trailer].compact.join("\n")
+        [signoff_trailers, custom_trailers].compact.join("\n")
       end
 
-      def custom_trailer
-        trailer = commit_message_options[:trailer]
-        return unless trailer.is_a?(String)
+      def custom_trailers
+        trailers = commit_message_options[:trailers]
+        return if trailers.nil?
+        raise("Commit trailers must be a Hash object") unless trailers.is_a?(Hash)
 
-        trailer
+        trailers.compact.map { |k, v| "#{k}: #{v}" }.join("\n")
       end
 
       def signoff_trailers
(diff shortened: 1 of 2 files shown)


common/lib/dependabot/pull_request_creator/message_builder.rb
```diff
@@ -142,6 +142,19 @@ def suffixed_pr_message_header
...
 
       def message_trailers
+        return unless signoff_trailers || custom_trailer
+
+        [signoff_trailers, custom_trailer].compact.join("\n")
+      end
+
+      def custom_trailer
+        trailer = commit_message_options[:trailer]
+        return unless trailer.is_a?(String)
```
jurre, line 152: When would it not be a string?

jurre:
I think this makes sense 👍 

One thing I am wondering though, not necessarily for this pull request but more in general going forward; should we maybe start thinking about moving some of this logic to GitLab specific (sub)classes?

common/lib/dependabot/pull_request_creator/message_builder.rb
```diff
@@ -142,6 +142,19 @@ def suffixed_pr_message_header
       end
 
       def message_trailers
+        return unless signoff_trailers || custom_trailer
+
+        [signoff_trailers, custom_trailer].compact.join("\n")
+      end
+
+      def custom_trailer
```
jurre, line 150: Should we maybe. mention this is GitLab specific?

andrcuns: In this case, having some custom git trailer is not necessarily a Gitlab specific logic. I think many workflows on any platform could do some actions based on presence of git trailer. It's just I happened to come by this usecase using Gitlab's changelog generation functionality 😄

common/lib/dependabot/pull_request_creator/message_builder.rb
andrcuns: It's mostly a guard clause in case someone would pass say and array of trailers by mistake since the `commit_message_options` object is a plain has that doesn't exactly document which parameters it supports.

Similar like current implementation of sign off details forces it to be a Hash

common/lib/dependabot/pull_request_creator/message_builder.rb
andrcuns: A bit more context on what other types of trailers someone might use from the git documentation itself: https://git-scm.com/docs/git-interpret-trailers

jurre: Makes sense, cheers 👍

common/lib/dependabot/pull_request_creator/message_builder.rb
jurre: Should we add a test for that? Not worried it won't work, mostly to prevent regressions in the future

andrcuns: Makes sense. Or maybe expand a bit and allow both String and Array and raise for anything else? I believe `git commit` command also allows passing `--trailer "token: value"` option multiple times which would add multiple trailers.

jurre: Yep that makes sense to me. Do you think we should still raise if an invalid option is passed, for example someone might pass a hash thinking it'd be `{ token: "value" }`?

andrcuns: I think it makes sense and it is essentially just one additional line of code which helps debugging if someone would have some code that creates trailers dynamically and it would be added as a hash by mistake for example.

andrcuns: Or we could go another way and just allow the hash, it maps better to passing one or multiple trailers and automatically forces to not pass multiple trailers with same `key`

jurre: That also works for me 👍

andrcuns:
I changed the format of custom trailers to be Hash only but it is still allowed to be `nil` to allow a bit more flexibility and easer unit test setup.

jurre (approved):

andrcuns:
@jurre Thanks for such a quick review 🎉

PR #4433, opened 2021-11-19
Remove reliance on `PandocRuby`

landongrindheim:
PandocRuby has been used to convert RestructuredText (rst), a
markdown-like format widely used in the Python ecosystem, to markdown.
We recently noticed new errors surfacing around Pandoc and started to
investigate.  This led to the discovery that Pandoc was not installed in
the Docker container GitHub is using to run Dependabot against
repositories.

I'm opting to remove this dependency as PandocRuby is effectively
unused.

Note: There is the possibility that some users rely on this
functionality. As has been noted in a recent PR-review, non-Docker usage
of dependabot-core is poorly supported, so this seems unlikely.

### DIFF
--- a/common/spec/dependabot/metadata_finders/base/changelog_finder_spec.rb
+++ b/common/spec/dependabot/metadata_finders/base/changelog_finder_spec.rb
@@ -798,18 +798,9 @@ RSpec.describe Dependabot::MetadataFinders::Base::ChangelogFinder do
             "* Fixed `issue <https://github.com/pytest-dev/"\
             "pytest-selenium/issues/216>`_ with TestingBot local tunnel."
           end
-          let(:converted_text) do
-            "1.16.0 (2019-02-12)\n"\
-            "===================\n"\
-            "\n"\
-            "-   `pytest-selenium` now requires pytest 3.6 or later.\n"\
-            "-   Fixed [issue](https://github.com/pytest-dev/"\
-            "pytest-selenium/issues/216) with TestingBot local tunnel.\n"
-          end
 
-          it "converts the rst properly (or falls back)" do
-            expect([converted_text, unconverted_text]).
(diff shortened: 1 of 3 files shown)


jurre (approved):
Thanks for tackling this!

jurre:
Oh fwiw there's a few tests that we can simplify with this change: https://github.com/dependabot/dependabot-core/blob/11d12da3ce8c646d7cbdc95eea512e06bd70ddbb/common/spec/dependabot/metadata_finders/base/changelog_finder_spec.rb#L779-L814

common/spec/dependabot/metadata_finders/base/changelog_finder_spec.rb
```diff
@@ -798,18 +798,9 @@
...
-            "-   `pytest-selenium` now requires pytest 3.6 or later.\n"\
-            "-   Fixed [issue](https://github.com/pytest-dev/"\
-            "pytest-selenium/issues/216) with TestingBot local tunnel.\n"
-          end
 
-          it "converts the rst properly (or falls back)" do
-            expect([converted_text, unconverted_text]).
-              to include(changelog_text)
+          it "does not convert the rst" do
+            expect(changelog_text).to eq(unconverted_text)
```
landongrindheim, line 803: I suspect this is not needed, but thought it would be useful to leave some breadcrumbs in our history. Is it better to remove this test altogether?

lseppala: I kind of like this. If tests are there to assert behaviors or constraints, and we are explicitly deciding on the behavior of _not_ converting RST, it makes sense to assert as much. Also, if someone asks if RST conversion is broken, we have a test to point to :)

jurre (approved):

PR #3630, opened 2021-05-04
Draft: Go modules ignore conditions using `go list -m -versions`

mctofu:
This is an initial exploration of moving go_module latest version selection out of the [`updatechecker`](https://github.com/dependabot/dependabot-core/blob/c3c87bb90042cf5b3d3af50d42b0e83fe3267041/go_modules/helpers/updatechecker/main.go#L34) helper and into the ruby code. The goal is to support the upcoming `update_type` ignore conditions.

Instead of modifying the `updatechecker` helper we started down the route of using `go list -m -versions --json <module>` to fetch the list of available versions and then added any extra filtering that was needed to make the tests pass.

```
% go list -m -versions --json github.com/dependabot-fixtures/go-modules-lib
{
	"Path": "github.com/dependabot-fixtures/go-modules-lib",
	"Version": "v1.1.0",
	"Versions": [
		"v1.0.0",
		"v1.0.1",
		"v1.1.0",
		"v1.2.0-pre1",
		"v1.2.0-pre2"
	],
	"Time": "2018-10-18T21:48:48Z"
}
```

This appeared to work pretty well and had some benefits:
- The result omits major version upgrades (since they would have a different module path) so we didn't need to reimplement the [major version filter](https://github.com/dependabot/dependabot-core/blob/c3c87bb90042cf5b3d3af50d42b0e83fe3267041/go_modules/helpers/updatechecker/main.go#L67)
- The command is aware of retracted go modules so we gained filtering of them for free
- No longer need to maintain the `updatechecker` code

However, we hit one snag. While pairing with @jasonrudolph we discovered that his dev shell was out of date and using go 1.15. When I tried this under 1.16 the `go list` command started to report this error:
```
go: github.com/dependabot-fixtures/go-modules-lib@v1.0.0: missing go.sum entry; to add it:
    go mod download github.com/dependabot-fixtures/go-modules-lib
```

A workaround I came up with was to change the command to `go list -m -mod=mod -versions --json github.com/dependabot-fixtures/go-modules-lib`. This works but takes much more time as it ends up downloading the module into the module cache. For `update_checker_spec.rb` this increased the runtime from ~2s to ~10s when run with an empty module cache. In practice this change would mean we need to download every dependency we check for updates instead of only downloading dependencies that need to be updated.

[...]

PR #4434, opened 2021-11-19
Use `go list -m -versions` to determine available versions of a go module

mctofu:
This removes usage of the [native helper](https://github.com/dependabot/dependabot-core/blob/417dd1de9ac6ee9f8e366031a8efb288f6484291/go_modules/helpers/updatechecker/main.go) to resolve available go module versions. Instead we'll use the existing go tooling which has added [support for this](https://golang.org/ref/mod#go-list-m). This has a few benefits:
- No longer need to maintain https://github.com/dependabot/gomodules-extracted or make calls to internal go libs
- No longer need to manually handle exclude directives
- Retract directives are now supported so Dependabot won't propose updates to retracted versions

We previously attempted this change in https://github.com/dependabot/dependabot-core/pull/3630 but the approach was abandoned after we found the change was negatively impacting performance. It was [further noted](https://github.com/dependabot/dependabot-core/pull/3630#issuecomment-832678444) that using `go list -m -versions` was causing the go tooling to download all of the modules referenced in `go.mod` and this accounted for the bulk of the performance hit. This appears to only happen when `GOPRIVATE=*` is set which we currently need to do to support private modules.

As a workaround I found that providing a `go.mod` with no requirements was a way to sidestep the performance issue. We still end up getting the version information we are looking for and no extra modules are downloaded! The performance is close to the native version as well.

### Example output / performance
<details>
  <summary>Using native helper</summary>
  
```
[dependabot-core-dev] /tmp/go $ rm -rf /home/dependabot/go/pkg/mod/cache
[dependabot-core-dev] /tmp/go $ cat go.mod
module foo

require github.com/dependabot-fixtures/go-modules-lib v1.0.0

go 1.15
[dependabot-core-dev] /tmp/go $ time echo '{"function":"getVersions","args":{"dependency":{"name":"github.com/dependabot-fixtures/go-modules-lib","version":"v1.0.0"}}}'  | GOPRIVATE='*' /opt/go_modules/bin/helper
{"result":["v1.0.0","v1.0.1","v1.0.5","v1.0.6","v1.1.0","v1.2.0-pre1","v1.2.0-pre2"]}
real    0m1.017s
user    0m0.105s
sys     0m0.087s
[dependabot-core-dev] /tmp/go $ du -sk /home/dependabot/go/pkg/mod/cache
176     /home/dependabot/go/pkg/mod/cache
```

</details>

<details>
  <summary>Using `go list -m -versions` with project `go.mod`</summary>
  
```
[dependabot-core-dev] /tmp/go $ rm -rf /home/dependabot/go/pkg/mod/cache
[dependabot-core-dev] /tmp/go $ cat go.mod
module foo

require github.com/dependabot-fixtures/go-modules-lib v1.0.0

go 1.15
[dependabot-core-dev] /tmp/go $ time GOPRIVATE='*' go list -m -versions -json github.com/dependabot-fixtures/go-modules-lib
{
        "Path": "github.com/dependabot-fixtures/go-modules-lib",
        "Version": "v1.0.0",
        "Versions": [
                "v1.0.0",
                "v1.0.1",
                "v1.0.5",
                "v1.0.6",
                "v1.1.0",
                "v1.2.0-pre1",
                "v1.2.0-pre2"
        ],
        "Time": "2018-10-18T21:48:48Z",
        "GoMod": "/home/dependabot/go/pkg/mod/cache/download/github.com/dependabot-fixtures/go-modules-lib/@v/v1.0.0.mod"
}

real    0m8.828s
user    0m1.882s
sys     0m1.465s
[dependabot-core-dev] /tmp/go $ du -sk /home/dependabot/go/pkg/mod/cache
24844   /home/dependabot/go/pkg/mod/cache
```

</details>

<details>
  <summary>Using `go list -m -versions` with stub `go.mod` (this PR)</summary>
  
```
[dependabot-core-dev] /tmp/go $ rm -rf /home/dependabot/go/pkg/mod/cache
[dependabot-core-dev] /tmp/go $ cat go.mod
module foo
[dependabot-core-dev] /tmp/go $ time GOPRIVATE='*' go list -m -versions -json github.com/dependabot-fixtures/go-modules-lib
{
        "Path": "github.com/dependabot-fixtures/go-modules-lib",
        "Versions": [
                "v1.0.0",
                "v1.0.1",
                "v1.0.5",
                "v1.0.6",
                "v1.1.0",
                "v1.2.0-pre1",
                "v1.2.0-pre2"
        ]
}

real    0m1.705s
user    0m0.140s
sys     0m0.135s
[dependabot-core-dev] /tmp/go $ du -sk /home/dependabot/go/pkg/mod/cache
228     /home/dependabot/go/pkg/mod/cache
```

</details>

### Other notes
- This change revealed that the [v2](https://github.com/dependabot-fixtures/go-modules-lib/blob/v2.0.0/go.mod) release of our test fixture wasn't actually valid. I created a valid [v3](https://github.com/dependabot-fixtures/go-modules-lib/blob/v3.0.0/go.mod) release and added handling for invalid version errors.
- I believe the usage of `SharedHelpers.run_shell_command` will escape the args being passed to `go mod list -m -versions` and `go mod edit` to prevent command injection. I'd appreciate a 2nd look at this though.


jeffwidman:
> As a workaround I found that providing a go.mod with no requirements was a way to sidestep the performance issue. We still end up getting the version information we are looking for and no extra modules are downloaded!

Very clever!

jurre (approved):
This looks great, so glad we're able to get rid of `gomodules-extracted`!

> I believe the usage of SharedHelpers.run_shell_command will escape the args being passed to go mod list -m -versions and go mod edit to prevent command injection. I'd appreciate a 2nd look at this though.

That's right, unless we explicitly tell `run_shell_command` to `allow_unsafe_shell_command` it runs any `command` we pass through `Shellwords` which should protect against any command injection 👍

weitzj:
Looking forward to this release.

common/lib/dependabot/shared_helpers.rb
```diff
@@ -280,10 +280,11 @@ def self.reset_global_git_config(backup_path)
       FileUtils.mv(backup_path, GIT_CONFIG_GLOBAL_PATH)
     end
 
-    def self.run_shell_command(command, allow_unsafe_shell_command: false)
+    def self.run_shell_command(command, allow_unsafe_shell_command: false, env: nil)
       start = Time.now
       cmd = allow_unsafe_shell_command ? command : escape_command(command)
-      stdout, process = Open3.capture2e(cmd)
+      env_cmd = [env, cmd].compact
+      stdout, process = Open3.capture2e(*env_cmd)
```
mattt, line 287: Could we rephrase this to make the intent of this a bit clearer? It took me a few minutes to understand what was happening here.

At first, I thought _"`env` is something like `TEST_ENV=prefix`, so this concatenates the original command"_. But then I realized that `env` is a hash, and had to consult the [docs for `Process#spawn`](https://docs.ruby-lang.org/en/2.0.0/Process.html#method-c-spawn) to understand how `compact` is called here to mitigate the awkwardness of `#spawn` taking `env` as an optional first argument.

Would something like this work instead?

```ruby
def self.run_shell_command(command, allow_unsafe_shell_command: false, env: {})
      start = Time.now
      stdout, process = Open3.capture2e(env, cmd)
```

landongrindheim: I think using a default of `env: {}` in the method signature might clear this up as well. We don't have a guarantee against `nil` being passed in, but we catch the cases where it is not supplied 🙂

jurre: 👍 the optional first argument has thrown me off a few times as well, it's such a weird API

mctofu: Love these ideas! Updated in https://github.com/dependabot/dependabot-core/pull/4434/commits/16395d30cedb5554090873ece20d99321cea4fc7.

There is another usage of compact at https://github.com/dependabot/dependabot-core/blob/ebe338ed2bd1d09b2c86510f9e44cc0cf642e18b/common/lib/dependabot/shared_helpers.rb#L93-L94 but I'm going to leave that for a future PR.

mattt (approved):

go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb
```diff
@@ -73,23 +75,22 @@ def fetch_lowest_security_fix_version
         def available_versions
           SharedHelpers.in_a_temporary_directory do
             SharedHelpers.with_git_configured(credentials: credentials) do
-              File.write("go.mod", go_mod.content)
+              manifest = parse_manifest
+
+              # Set up an empty go.mod so 'go list -m' won't attempt to download dependencies. This
```
mctofu, line 80: One other observation from testing is that we can also run `go list -m -versions` without a `go.mod` and it works the same way. However, it won't know about the exclude directives and won't be able to apply them for us so creating a `go.mod` with the excludes is still preferred.

go_modules/helpers/main.go
```diff
@@ -32,10 +31,6 @@ func main() {
 		funcErr error
 	)
 	switch helperParams.Function {
-	case "getVersions":
-		var args updatechecker.Args
-		parseArgs(helperParams.Args, &args)
-		funcOut, funcErr = updatechecker.GetVersions(&args)
 	case "getVcsRemoteForImport":
 		var args importresolver.Args
 		parseArgs(helperParams.Args, &args)
```
jeffwidman, line 36: May be able to delete this as well, I filed https://github.com/dependabot/dependabot-core/issues/4448 for that

jeffwidman:
Nicely done! 

I opened https://github.com/dependabot/gomodules-extracted/pull/17 and https://github.com/dependabot/gomodules-extracted/issues/18 as followups to this.

mctofu:
I noticed one unanticipated side effect where Dependabot now updates major versions for `+incompatible` versions: https://github.com/dependabot/dependabot-core/issues/4453

PR #4439, opened 2021-11-22
Python: Honour `--strip-extras` flag of `pip-compile`

NicolasT:
The `--strip-extras` flag of `pip-compile` is useful to generate
contstraints files (instead of the more traditional requirements files)
from some `constraints.in` file.

Constraints files can't have extras markers on the packages (this
wouldn't make much sense indeed), hence they need to be stripped. Since
`pip-tools` 6.2.0 this is supported by using the `--strip-extras` flag.

This commit adds this as a 'recognized' flag: if it's found in an
existing `foo.txt` file, it will be passed to `pip-compile` when some
`foo.in` file is updated.

See: https://github.com/dependabot/dependabot-core/pull/3974#issuecomment-866660246
See: https://github.com/dependabot/dependabot-core/pull/3974#issuecomment-975953752
See: https://github.com/jazzband/pip-tools/issues/1300#issuecomment-860020431


NicolasT:
(Note: I have no clue whether this is the right approach to add tests, and how to actually write proper tests in the given framework. Not a Ruby'ist either. So this is very much hit-and-miss)

jurre:
I think this makes sense, letting CI run to make sure there aren't any unexpected side-effects, but thanks for contributing!

NicolasT:
Not sure how this change can cause a Bundler test to fail :thinking:

jurre:
> Not sure how this change can cause a Bundler test to fail 🤔

That's not related, it's a known flaky test unfortunately

jurre (approved):

NicolasT:
Thanks for merging!

Are there any docs on how one can know whether/when this change landed in the Github-hosted Dependabot service?

NicolasT:
Looks like it's in prod, and it's achieving the desired result: I now get 'correct' (without extras) Dependabot PRs for my `constraints.txt`. Sweet!

jurre:
> Are there any docs on how one can know whether/when this change landed in the Github-hosted Dependabot service?

No docs on this currently, but typically once a PR is included in a release (you should get a mention in the release notes), we'll roll that out in production the same or next day.

Thanks again for the contribution!

PR #4442, opened 2021-11-23
Hex: Handle additional dynamic version reading approaches

landongrindheim:
I _think_ `#sanitized_content` is meant to make a common Elixir pattern
safe for parsing (rather, to prevent code execution while we dependency
management is being performed). Unfortunately, there are no hints in
either Git commit history or in the pull requests that were opened
around Elixir package management.

The pattern that this is handling (at least in the fixtures in this
repo) is the case where there's a top-level file titled VERSION, which
is read in in the Mixfile's project metadata. We're handling a couple of
cases here:

```elixir
String.trim(File.read("VERSION"))  # String.trim("0.0.1")
String.trim(File.read!("VERSION")) # String.trim({:ok, "0.0.1"})
```

A lot of Elixir code relies on piping output from one function to
another, which can cause the above pattern to read as:

```elixir
"VERSION" |> File.read() |> String.trim()
"VERSION"
|> File.read()
|> String.trim()
```

We've not been handling these pipelines properly, which has led to errors like:
```plaintext
(ArgumentError) cannot pipe "VERSION" into "0.0.1", can only pipe (snip)
```

This pull request is meant to enable us to handle both the nested and piped approaches.


hex/lib/dependabot/hex/file_updater/mixfile_sanitizer.rb
```diff
@@ -11,16 +11,40 @@ def initialize(mixfile_content:)
           @mixfile_content = mixfile_content
         end
 
+        FILE_READ      = /File.read\(.*?\)/.freeze
+        FILE_READ_BANG = /File.read!\(.*?\)/.freeze
```
xlgmokha, line 15: question (non-blocking): it looks like we're trying to match `File.read`. Do we also need to match `IO.read` and/or `Pathname#read`?

xlgmokha (approved):

hex/lib/dependabot/hex/file_updater/mixfile_sanitizer.rb
landongrindheim: We shouldn't. These are matching Elixir functions what share a name with Ruby methods. They _do_ have a very familiar feel as the language's creator came out of the Ruby community 😄

PR #4444, opened 2021-11-23
python: Upgrade to pip 21.3.1

rouge8:
This patch contains #4345 plus a fix for the failing test.

Closes #4345.


jurre (approved):
Thanks!

PR #4447, opened 2021-11-23
Bump minimum to 1.17

jeffwidman:
This isn't strictly necessary, but since the rest of the infra bumped to 1.17,
might as well bump it here too. Esp since the `go.mod` behavior changed
a bit in `1.17`... so if more libs get added ever, this makes it so `go.mod`/`go.sum`
will follow the new behavior...


lseppala (approved):
Excellent, thanks for helping keep this up-to-date

PR #4450, opened 2021-11-24
Limit the GitHub PR description message to API max

lseppala:
Limit the PR description message length to the maximum supported by GitHub API (65,536 characters).

Strictly speaking, the API can support up to 262,144 _bytes_. As Ruby strings are UTF-8 encoded, which represents each character with 1 to 4 bytes, a longer message could be sent. This is pessimistic and assumes 4-bytes per character. However, the API responds with an error message saying the limit is 65536 characters. This is easier to support--we don't have to figure out what to do if the truncation occurs at the end of a multi-byte character, which can be tricky.

### DIFF
--- a/common/lib/dependabot/pull_request_creator/github.rb
+++ b/common/lib/dependabot/pull_request_creator/github.rb
@@ -9,6 +9,9 @@ module Dependabot
   class PullRequestCreator
     # rubocop:disable Metrics/ClassLength
     class Github
+
+      MAX_PR_DESCRIPTION_LENGTH = 65536
+
       attr_reader :source, :branch_name, :base_commit, :credentials,
                   :files, :pr_description, :pr_name, :commit_message,
                   :author_details, :signature_key, :custom_headers,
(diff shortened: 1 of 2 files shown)


common/lib/dependabot/pull_request_creator/github.rb
```diff
@@ -347,6 +350,18 @@ def add_milestone_to_pull_request(pull_request)
       end
 
       def create_pull_request
+        # Limit PR description to MAX_PR_DESCRIPTION_LENGTH (65,536) characters
+        # and truncate with message if over. The API limit is 262,144 bytes
+        # (https://github.community/t/maximum-length-for-the-comment-body-in-issues-and-pr/148867/2).
+        # As Ruby strings are UTF-8 encoded, this is a pessimistic limit: it
+        # presumes the case where all characters are 4 bytes.
+        pr_description = @pr_description.dup
+        if pr_description && pr_description.length > MAX_PR_DESCRIPTION_LENGTH
```
lseppala, line 359: This is similar to a block in `common/lib/dependabot/clients/azure.rb`, which also has to deal with encoding issues. Chose to copy in preference of a [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) refactor.

mattt (approved):
Nice work! I think this is a reasonable approach to sidestep complex encoding edge cases. ("65k ought to be enough", amiright?)

common/lib/dependabot/pull_request_creator/github.rb
```diff
@@ -9,6 +9,9 @@ module Dependabot
   class PullRequestCreator
     # rubocop:disable Metrics/ClassLength
     class Github
+
+      MAX_PR_DESCRIPTION_LENGTH = 65536
```
mattt, line 13: Since the constant declaration comes before `#create_pull_request`, you might want to attach the comment here instead.

Or, alternatively:

```suggestion
      MAX_PR_DESCRIPTION_LENGTH = 65_536 # characters (see #create_pull_request)
```

jurre:
I think a quick `bundle exec rubocop -A -c ../rubocop.yaml` in the `common` directly should fix up CI

mattt (approved):

common/lib/dependabot/pull_request_creator/github.rb
djrising1234: Thank you

PR #4064, opened 2021-07-20
Add Azure autocomplete_pull_request

trejjam:
This PR adds support for Azure PR autocompletion.

Azure doc: https://docs.microsoft.com/en-us/rest/api/azure/devops/git/pull-requests/update?view=azure-devops-rest-6.0

edit: fix link

PR #4451, opened 2021-11-24
[WIP] Allow Azure client to set auto complete

yufeih:
Azure DevOps has an option to auto merge a PR after review and tests, this is to allow an option to set auto complete by dependabot, to save 1 click for the merge. The `auto_complete` property is the user id that enabled it. 

Screen Shot 2021-11-24 at 8 23 07 PM

API docs: https://docs.microsoft.com/en-us/rest/api/azure/devops/git/pull-requests/create?view=azure-devops-rest-5.0

### DIFF
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -172,7 +172,7 @@ module Dependabot
 
       # rubocop:disable Metrics/ParameterLists
       def create_pull_request(pr_name, source_branch, target_branch,
-                              pr_description, labels, work_item = nil)
+                              pr_description, labels, work_item = nil, auto_complete = nil)
         pr_description = truncate_pr_description(pr_description)
 
         content = {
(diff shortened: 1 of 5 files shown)


jurre:
@yufeih could you add some context in the PR description as to why this change is needed?

@AlekhyaYalla given that you've contributed to the Azure client a bunch recently, would you mind reviewing this code?

yufeih:
@jurre Definitely! I just realize that `autoCompleteSetBy` only works when updating an existing pull request, so I mark this as WIP before it is ready.

jeffwidman:
@yufeih are you still planning to work on this?

jeffwidman:
It looks like there's currently two open PR's for this functionality, but they're implemented quite differently:
* https://github.com/dependabot/dependabot-core/pull/4451
* https://github.com/dependabot/dependabot-core/pull/4064

We are a small team and won't be able to test this or spend a whole lot of time beyond basic code review, so need input from folks who use Dependabot on Azure for which approach to take...

yufeih:
#4064 looks like the correct implementation. Autocomplete needs an additional API call after a pull request has been created, so I'll close this one.

jeffwidman:
Thanks for the fast reply!

Issue #4361, opened 2021-11-02
package-lock files with version 3 are not handled correctly by Dependabot


ferferga:
We recently upgraded to Node 16 and the lockfile version 3 of package-lock.json (the version without backwards-compatible features of the lockfile, see docs [here](https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json#lockfileversion), as we want to enforce always the most current features with the latest LTSs).

Lockfiles proposed by dependabot have massive changes, even when recreated or rebased

**Package ecosystem**
**npm**
**Package manager version**
**npm 8.1.0**
**Language version**
- Node 16

**Relevant PRs**

- https://github.com/jellyfin/jellyfin-vue/pull/1585
- https://github.com/jellyfin/jellyfin-vue/pull/1584
- https://github.com/jellyfin/jellyfin-vue/pull/1583
- https://github.com/jellyfin/jellyfin-vue/pull/1582
- https://github.com/jellyfin/jellyfin-vue/pull/1581
- https://github.com/jellyfin/jellyfin-vue/pull/1580
- https://github.com/jellyfin/jellyfin-vue/pull/1579
- https://github.com/jellyfin/jellyfin-vue/pull/1578
- https://github.com/jellyfin/jellyfin-vue/pull/1577

cedric-anne:
I confirm the issue. Same problem on this PR: https://github.com/glpi-project/glpi/pull/10035 .

jeffwidman:
Closing, as I think this was fixed by the bump to NPM 8:
* https://github.com/dependabot/dependabot-core/pull/4763

Please comment if I'm misreading/misunderstanding something, and we can re-open.

cedric-anne:
Indeed, it seems to be OK now.

PR #4454, opened 2021-11-25
Dependabot config: ignore npm lib major version updates

jurre:
These helpers exist to support npm 6, so we require that specific major version and should not update it.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,6 +20,10 @@ updates:
     directory: "/npm_and_yarn/helpers"
     schedule:
       interval: "daily"
+    ignore:
+      - dependency-name: "npm"
+        # We explicitly need npm 6 here, in order to support npm 6.
+        versions: ["version-update:semver-major"]
   - package-ecosystem: "pip"
     directory: "/python/helpers"
     schedule:


brrygrdn (approved):

ferferga:
@jurre Updating npm shouldn't fix this? https://github.com/dependabot/dependabot-core/issues/4361

Can you please detail why npm needs to be pinned to 6? I'm interested in make a PR to fix that issue, as it's bothering us too much and I want it to be solved as soon as possible, but I need to understand all the casuistics of it :)

jurre:
> Can you please detail why npm needs to be pinned to 6?

We support npm 6 and npm 7, which have different native helpers. The helpers for npm 6 depend on that specific version :)

You can see both helpers side by side [here](https://github.com/dependabot/dependabot-core/tree/a5efcdaba81d5d4e1034e773f565b248f0346df8/npm_and_yarn/helpers/lib), where the `npm` folder currently holds the helpers for npm 7.

In order to resolve the linked issue, we'd have to either figure out if the helpers in the `npm` folder can be augmented to support both versions, and also if there are no other code changes required throughout the codebase. If we cannot support both versions, we'll have to split out another set of helpers specifically for npm 8.

ferferga:
@jurre Thank you very much for pointing me to the specifics of the code.

I created a PR at https://github.com/dependabot/dependabot-core/pull/4518. It should fix the issue, while avoiding a full upgrade to npm 8 (which shouldn't break anything as npm 8 is just npm 7 with dropped node 12 support), as I understand that such huge library updates are problematic and takes more time to ship :).

Hopefully we can get that one merged and deployed soon, so all of us who are using v3 lockfiles can start using dependabot again.

Thank you very much again for your input!

PR #4460, opened 2021-11-26
Update npm6 to allow for better `semver` handling

devpow112:
Often when using **git**, **github** or **gitlab** style `semver` dependencies with **npm** (ex, `github:Brightspace/d2l-activity-alignments#semver:^2`) the tags/branches associated with the repository can be for the form `v[major].[minor].[patch]` (common if following standard `semver` set-up s and automated tooling such as `semantic-release`) or `[major].[minor].[patch]`. The `semver` keyword allows for variance of branch name allowing `v3.0.0` and `3.0.0` to be treated as the same version allowing these dependency checks to succeed. When not using the `semver` keyword in the dependency install command the resolved version will be treated as needing to be an exact match for a branch/tag name. This change adjust the install command used to update package lock file to include the `semver` keyword whenever it's present in the existing requirement which allows this variance in branch naming to be allowed. Once this change is done we also need to repair the `from` properties of the installed dependencies since they will no longer match the existing requirement in the `package.json` file (repair `github:Brightspace/d2l-activity-alignments#semver:3.0.0` to `github:Brightspace/d2l-activity-alignments#semver:^3`).

### Example of something that wouldn't work before

The `github:Brightspace/d2l-activity-alignments#semver:^2` dependency is attempting to be updated to `github:Brightspace/d2l-activity-alignments#semver:^3` which results in the install command of
```
npm install d2l-activity-alignments@github:Brightspace/d2l-activity-alignments#3.0.0
```
This causes a git error since the branch on the target repository is `v3.0.0`.

### After change
The `github:Brightspace/d2l-activity-alignments#semver:^2` dependency is attempting to be updated to `github:Brightspace/d2l-activity-alignments#semver:^3` which results in the install command of
```
npm install d2l-activity-alignments@github:Brightspace/d2l-activity-alignments#semver:3.0.0
```
This succeeds even though the tag name is `v3.0.0`.

#### Notes
Repository used for testing of changes along side existing tests https://github.com/devpow112/dependabot-core-test. This would just throw errors when trying to run dependabot


lseppala (commented):
Thanks for your contribution! This seems reasonable and seems to do as you describe. Because this is change that could have significant impact, we'll want to test this on a repo before merging—thanks again for providing a test repository we can do this with. I'll report back with results later.

I think it might make sense to test this behavior by adding a test to the [updater test file](https://github.com/dependabot/dependabot-core/blob/main/npm_and_yarn/helpers/test/npm6/updater.test.js). Would you mind doing so? I may be able to help later.

devpow112:
@lseppala for sure. I'll work on adding some tests to the updater file stuff as well as looking why the latest test runs failed. Thanks for taking a look.

lseppala:
With regards to the test failure, there was an external dependency we used in testing that had drifted. We corrected this in #4466. Update your branch with the latest in `main` (or click the "Update branch" button below) to fix :)

Afsoning:
#4485 #4482 #4460 #4483 #4482

Issue #3611, opened 2021-04-30
Add an issue category that isn't bug/feature request


jeffwidman:
I've been filing a number of tickets that aren't bugs--the code works find today--but also aren't feature requests. They're more like tech debt/code improvements. I never know which category to use. So maybe add another category like "tech debt"? With the idea that it _improves_ dependabot w/o actually adding any new features?

Examples of these types of tickets:
* https://github.com/dependabot/dependabot-core/issues/4448
* https://github.com/dependabot/dependabot-core/issues/4154
* https://github.com/dependabot/dependabot-core/issues/3612

jeffwidman:
cc @brrygrdn for visibility since I see you do a lot of labeling these days 😀

jurre:
That makes sense, I've added a `T: tech-debt` label 👍

jeffwidman:
Added a issue template here: https://github.com/dependabot/dependabot-core/pull/4470

PR #4470, opened 2021-11-30
Add tech-debt issue template

jeffwidman:
Create a issue template for the new `T: tech-debt` label 
added as part of https://github.com/dependabot/dependabot-core/issues/3611.


.github/ISSUE_TEMPLATE/tech-debt.md
```diff
@@ -0,0 +1,15 @@
+---
+name: "\U00002699 Tech Debt"
```
jeffwidman, line 2: @jurre I wasn't sure which fun emoji you thought would be best... this gear icon looked good. If you like it, maybe add it to the label itself just like is on the bug label?

jurre: I dig it 👍

PR #4474, opened 2021-12-01
Convert updated content to XML before diffing

landongrindheim:
Recently we made a change that updated our file parsing to use Nokogiri
in place of Regexp. This solved the problem at hand, but introduced a
new class of issues. One issue that we've encountered is in handling
whitespace in self-closing tags. Example:

```ruby
my_tag = "<myTag />"
=> "<myTag />"       # note the extra space
Nokogiri::XML(my_tag).to_xml
=> "<myTag/">        # note the space has been removed
```

We use `String#gsub` to update the pom file in the Maven FileUpdater.
When processing the converted XML and the original file, there is no
match, which causes errors reading "Expected content to change!".


landongrindheim:
Closing this as we're not ready to commit to a method of handling pom.xml files.

PR #4477, opened 2021-12-01
Test for updating version of a parent pom

mctofu:
This adds a test case for the regression fixed by https://github.com/dependabot/dependabot-core/pull/4479


maven/spec/fixtures/poms/pom_with_parent.xml
```diff
@@ -13,7 +13,7 @@
     <groupId>org.springframework.boot</groupId>
     <artifactId>spring-boot-starter-parent</artifactId>
     <version>1.5.9.RELEASE</version>
-    <relativePath/> <!-- lookup parent from repository -->
+    <relativePath /> <!-- lookup parent from repository -->
```
landongrindheim, line 16: This makes sense to me. Since the test using this is specific to preserving formatting, I think it would be useful to call that out in some way.

I understand we're effectively testing the whole stack in the test, so I'm not sure the best way to make it obvious. If you can think of a way, without spending a lot of time, I think it would be beneficial 🙂

landongrindheim (approved):

maven/spec/fixtures/poms/pom_with_parent.xml
mattt: > I understand we're effectively testing the whole stack in the test, so I'm not sure the best way to make it obvious.

Maybe change the filename to something like `pom_with_parent_and_insignificant_whitespace.xml`?

mattt (approved):

landongrindheim (approved):
🎉

maven/spec/fixtures/poms/pom_with_parent.xml
mctofu: Thanks for the ideas! I had been debating if it was worth creating a new fixture for the whitespace case and I think this shows that it would. I added a new fixture/test case for the whitespace issue and checked locally that the test still fails under the original conditions: https://github.com/dependabot/dependabot-core/pull/4477/commits/3fc608c61a90d3e0f3637efd4033d8f7af49a7e6

PR #4487, opened 2021-12-05
Maven: Handle nested plugin declarations

jurre:
This was originally resolved in
https://github.com/dependabot/dependabot-core/pull/4417 but later
reverted, because that changed introduced a regression due to parsing
the XML using Nokogiri, which caused formatting of the XML to be lost.

We've observed errors around pom files that have a nested plugin
section inside an outer plugin declaration.

Given the following XML:

```xml
<plugins>
  <plugin>
    <configuration>
      <jvmTarget>11</jvmTarget>
      <compilerPlugins>
        <plugin>spring</plugin>
      </compilerPlugins>
    </configuration>
    <groupId>org.jetbrains.kotlin</groupId>
    <artifactId>kotlin-maven-plugin</artifactId>
    <version>${kotlin.version}</version>
  </plugin>
</plugin>
```

The <plugin>spring</plugin> declaration would cause a regex match,
resulting in the required information (the version in this case) to be
omitted from the XML snippet.

This is resolved by matching on an additional nested plugin block in the
regex, so that we always end up matching until the next closing tag.

This does mean that if _another_ level of nesting is present we'd still
end up with an incomplete XML section. This also does not resolve the
issue for other elements than the `plugin` one. If we do observe the
same issue with those other elements we can introduce the same type of
nesting there.

### DIFF
--- a/maven/lib/dependabot/maven/file_updater/declaration_finder.rb
+++ b/maven/lib/dependabot/maven/file_updater/declaration_finder.rb
@@ -11,7 +11,7 @@ module Dependabot
       class DeclarationFinder
         DECLARATION_REGEX =
           %r{<parent>.*?</parent>|<dependency>.*?</dependency>|
-             <plugin>.*?</plugin>|<extension>.*?</extension>}mx.freeze
+             <plugin>.*?(?:<plugin>.*?</plugin>.*)?</plugin>|<extension>.*?</extension>}mx.freeze
 
         attr_reader :dependency, :declaring_requirement, :dependency_files
 
--- a/maven/spec/dependabot/maven/file_updater/declaration_finder_spec.rb
+++ b/maven/spec/dependabot/maven/file_updater/declaration_finder_spec.rb
@@ -420,5 +420,32 @@ RSpec.describe Dependabot::Maven::FileUpdater::DeclarationFinder do
           to eq("org.springframework")
       end
     end
+
+    context "with a plugin that contains a nested plugin configuration declaration" do
+      let(:pom) do
+        Dependabot::DependencyFile.new(name: "pom.xml", content: fixture("poms", "nested_plugin.xml"))
+      end
+      let(:dependency_name) { "org.jetbrains.kotlin:kotlin-maven-plugin" }
+      let(:dependency_version) { "1.4.30" }
+      let(:declaring_requirement) do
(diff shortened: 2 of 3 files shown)


maven/lib/dependabot/maven/file_updater/declaration_finder.rb
```diff
@@ -11,7 +11,7 @@ class FileUpdater
       class DeclarationFinder
         DECLARATION_REGEX =
           %r{<parent>.*?</parent>|<dependency>.*?</dependency>|
-             <plugin>.*?</plugin>|<extension>.*?</extension>}mx.freeze
+             <plugin>.*?(?:<plugin>.*?</plugin>.*)?</plugin>|<extension>.*?</extension>}mx.freeze
```
landongrindheim, line 14: I saw that Ruby has the ability to express [recursive regex](https://ruby-doc.org/core-2.7.0/Regexp.html#class-Regexp-label-Subexpression+Calls). I'm in the process of wrapping my mind around that, but I'm wondering if we could use that functionality. Are you familiar?

If you think recursive regex is a promising solution, I think it's worth spending some time to get it right (and to make it as readable as possible). If not, or if it seems too esoteric, I think what you have added here will solve the problem that is in front of us 👍

jurre: I've not had to use it, could you show an example of what the regex could look like and what edgecases it would catch over the current implementation? It doesn't have to fully work, but just to express your thinking?

landongrindheim: > could you show an example of what the regex could look like

I've been trying to grok the syntax for recursive regex. I'm not great at recursive thinking, and handling multiple tag types makes that more of a challenge. After trying this out in Rubular for a bit, I'm ready to say the investment isn't worth it for now.

>  what edgecases it would catch over the current implementation

My thinking was that we'd be able to possibly catch the edge case we saw with nested `<plugin>` declarations for each of the tags we're looking at. I was less concerned with edge cases and more concerned with not having to handle other issues as they pop up (e.g. nested `<extension>` tags).

---

Conclusion here is that I don't think the investment is worth it for now.

landongrindheim (approved):

PR #4490, opened 2021-12-06
feat(:git): do not decline valid git@... source

sebbrandt87:
Now it will be possible to also pass the source_type(source_string) call 
with a terraform valid git@... source property being set.

The function [git_source_details_from(source_string)](https://github.com/dependabot/dependabot-core/blob/main/terraform/lib/dependabot/terraform/file_parser.rb#L225-L234) is already prepared 
to handle this type.

### DIFF
--- a/terraform/lib/dependabot/terraform/file_parser.rb
+++ b/terraform/lib/dependabot/terraform/file_parser.rb
@@ -289,7 +289,7 @@ module Dependabot
         return :path if source_string.start_with?(".")
         return :github if source_string.include?("github.com")
         return :bitbucket if source_string.start_with?("bitbucket.org/")
-        return :git if source_string.start_with?("git::")
+        return :git if source_string.start_with?("git::") || source_string.start_with?("git@")
         return :mercurial if source_string.start_with?("hg::")
         return :s3 if source_string.start_with?("s3::")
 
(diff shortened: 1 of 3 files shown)


jurre:
This looks good to me, cc @Nishnha wdyt?

Nishnha (changes requested):
I did not know that we had previously excluded `git@` sources!

Nit: Can we reuse an existing test module with the `git@` protocol? Perhaps https://github.com/dependabot/dependabot-core/blob/0146d0f2572124d337563e9a667a1ffe1e579fd5/terraform/spec/fixtures/projects/git_tags_012/main.tf#L199-L210 ?

Otherwise, this looks good to me!

sebbrandt87:
😎
The reason I initial did not want to reuse an existing test module, is that there is also an extra catch just for direct github.com sources. So I decided for now I don't want to touch that piece and instead extend the `:git` catch.

Nishnha:
Yep, that makes sense to me @sebbrandt87 
Can I request that you change the test folder name, then? 💟 

We originally used the `git_tags_0xx` folder naming scheme to test changes between Terraform 0.11 and 0.12 😅 
Maybe a more appropriate folder name would be `git_protocol`? What do you think?

sebbrandt87:
I am doing it right now @Nishnha 👍

sebbrandt87:
Btw @Nishnha I force pushed to keep a clean history. Hope this was okay in this case here.

Nishnha (approved):
LGTM 👍🏾 
We will merge this in and release it with our next Dependabot-Core release 😄 
If it's okay with you, we will mention you in the release notes!

PR #4498, opened 2021-12-07
chore(terraform/helpers): added Darwin checksum and if clause to switch

sebbrandt87:
Added so that we can also depend on the helpers script to install 
hcl2json for all the MacOS users out there 😉

### DIFF
--- a/terraform/helpers/build
+++ b/terraform/helpers/build
@@ -14,9 +14,14 @@ fi
 
 os="$(uname -s | tr '[:upper:]' '[:lower:]')"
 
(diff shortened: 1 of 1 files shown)


jurre:
Not sure that we should do this, typically we strongly encourage everyone to use the development container that we ship, as setting up all the dependencies for dependabot-core is non-trivial and that mirrors what we use in production 🤔 

For terraform specifically we don't require that many things right now, I believe, but developers might still be running a different version locally etc, so ideally they'd also use the dev container. Wdyt, does it sound reasonable?

sebbrandt87:
Indeed sounds reasonable, in this case we only would need to change the Readme.md 😇.
As long as it is easy for everyone to start developing and running tests I am all in.

I close this PR and will open a new one later.

PR #4501, opened 2021-12-08
Test PR for supporting complex workspace regexes

milind009:
This PR deals with changes required for supporting complex workspace regexes like packages/**/* in dependabot-core. This solution currently only works for AzureDevOps

### DIFF
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -170,6 +170,51 @@ module Dependabot
           "/pushes?api-version=5.0", content.to_json)
       end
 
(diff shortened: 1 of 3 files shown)


jurre:
@milind009 it seems like this PR would require quite a bit of work in order for it to be ready to be merged (make sure it doesn't break things for other providers than ADO, add tests, the code could use some cleaning up etc), is this still something you're interesting in getting over the line? If yes, please lmk how I can help!

jeffwidman:
Closing as it it looks like this was fleshed out more thoroughly by @milind009 but in a different PR: https://github.com/dependabot/dependabot-core/pull/4549

PR #4510, opened 2021-12-10
Dart/Flutter pub support

sigurdm:
This is a MVP preview of the pub support ready for first comments.

The pub-side of this is implemented here: https://github.com/dart-lang/pub/tree/dependency_services

Known limitations (others feel free to add by updating this description):
 - Doesn't support updating git dependencies. 
 - No support for vulnerabilities.
 - If the found version is ignored (by dependabot config) no update will happen (an earlier version is not used)
 - Very limited metadata support (just retrieves the github link)
 - Only stable versions of dart and flutter supported.
 - No support for private package repositories.
 - `updated_dependencies_after_full_unlock` only allows updating to a later version, if the latest version that is mutually compatible with other dependencies is the latest version of the said package. This is a dependabot limitation.


Dockerfile
```diff
@@ -14,49 +14,49 @@ ENV DEBIAN_FRONTEND="noninteractive" \
 RUN apt-get update \
   && apt-get upgrade -y \
   && apt-get install -y --no-install-recommends \
-    build-essential \
```
jonasfj, line 17: I'm guess my formatter in vscode did a few ticks here...

jurre: I don't feel strongly about this 🤷

sigurdm: Sorry - some unrelated changes sneaked in.

I did an update to remove them

Dockerfile
```diff
@@ -224,6 +224,38 @@ RUN apt-add-repository "deb [arch=amd64] https://apt.releases.hashicorp.com $(ls
...
+  && chown -R dependabot:dependabot "$PUB_CACHE" \
+  && chown -R dependabot:dependabot /opt/dart/pub
+
+# Install Flutter
+ARG FLUTTER_VERSION=2.5.3
+RUN curl --connect-timeout 15 --retry 5 "https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_${FLUTTER_VERSION}-stable.tar.xz" > "/tmp/flutter.xz" \
+  && tar xf "/tmp/flutter.xz" -C /opt/dart \
+  && chmod -R o+rx "/opt/dart/flutter" \
+  && chown -R dependabot:dependabot "/opt/dart/flutter" \
+  && runuser -l dependabot -c "/opt/dart/flutter/bin/flutter --version"
```
jurre, line 257: Do y'all have any indication what this means for the size of the resulting docker image? It's already fairly large, and we might want to start thinking about splitting things out (which will require some work on our end)

jurre:
Thanks! This is pretty exciting. I haven't done an in-depth review, but one thing I notice is that we still end up with a fair bit of dependabot-specific ruby code. I think a lot of that comes down to the fairly chatty API that dependabot-core currently exposes.

Now that we're relying on the package manager much more heavily, I'm sure we can come up with a much simpler interface, and end up with less boiler-plate-ish code when integrating a new package manager.

Changing the interface that dependabot-core exposes would require some changes on our end with the way we run dependabot internally, but I think it would ultimately be worth it. We're currently researching what that could look like, and taking this PR as a starting point to see if we can indeed simplify things a bunch. Realizing that's fairly vague and hand-wavy so I'll update here as we learn more!

Anything specific that you'd like us to dig into right now?

ted-marozzi:
You are hero's @sigurdm and the dependabot team!

Dockerfile
sigurdm: Locally on my linux machine this expands to a 831M folder.

If we wanted to try to optimize for space:

The only things we need from flutter is:
* the version
* the "sdk-packages" (< 50 megs)

If we could obtain those  'out-of-band' we could use the dart-sdk only (~481M decompressed)... Not sure it is worth it...

sigurdm:
> Anything specific that you'd like us to dig into right now?

I think the biggest thing is that we would prefer versions and requirements to be "black-box" strings seen from dependabot.
In that way we would not have to have ad-hoc implementations (probably full of bugs) of these for each ecosystem.
I don't know how much logic dependabot needs for these (maybe it need to test for inclusions in ranges etc.).

A few other things come to mind:

* The update_checker interface could be simplified significantly, I think we could have an interface like:

  * `latest_version `
  * `updated_requirements(requirements_to_unlock)`  returns `nil` if no upgrade available for the given 
     `requirements_to_unlock` otherwise it returns a list of updated requirements.

* If we go towards always cloning the whole repo every time, the Dependabot::DependencyFile might be rendered redundant (just use the path).

* Another possible restructuring would be to not make one update-checker for each dependency, but instantiate it once, and query it for each dependency. It makes sense to do resolutions for all dependencies at once (shares version-listings between all resolutions) - we would not have to cache the output of the report.

@jonasfj might have more ideas.

jurre:
> I think the biggest thing is that we would prefer versions and requirements to be "black-box" strings seen from dependabot. In that way we would not have to have ad-hoc implementations (probably full of bugs) of these for each ecosystem.

Yeah that's exactly what I was getting at with the simplified interface as well 👍 

> The update_checker interface could be simplified significantly, I think we could have an interface like:

I am not sure we even need an `UpdateChecker`, as long as we can pass in the right configuration, that separate step is merely an implementation detail of how Dependabot currently runs. Ultimately we only care that an update is performed and we get either the results or a meaningful error back.

Dockerfile
jonasfj: Or we could delete everything else after extracting the Flutter SDK :D

mctofu: I built this and found that these additions increased the image size by about 3GB which is almost doubling the size. If it's possible to optimize this it would be very helpful as we're already concerned about the size of the dependabot-core image and would need to look into splitting the image by ecosystem to accommodate that much growth.

The dart sdk was around 600MB:
```
/bin/bash -o pipefail -c curl --connect-timeout 15 --retry 5 "https://storage.googleapis.com/dart-archive/channels/stable/release/${DART_VERSION}/sdk/dartsdk-linux-x64-release.zip" > "/tmp/dart-sdk.zip"   && mkdir -p "$PUB_CACHE"   && chown dependabot:dependabot "$PUB_CACHE"   && unzip "/tmp/dart-sdk.zip" -d "/opt/dart" > /dev/null   && chmod -R o+rx "/opt/dart/dart-sdk"   && rm "/tmp/dart-sdk.zip"   && dart --version # buildkit                                                                                                                    573MB
```

Flutter was around 2.3GB
```
/bin/bash -o pipefail -c curl --connect-timeout 15 --retry 5 "https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_${FLUTTER_VERSION}-stable.tar.xz" > "/tmp/flutter.xz"   && tar xf "/tmp/flutter.xz" -C /opt/dart   && chmod -R o+rx "/opt/dart/flutter"   && chown -R dependabot:dependabot "/opt/dart/flutter"   && runuser -l dependabot -c "/opt/dart/flutter/bin/flutter --version" # buildkit                                                                                                    2.33GB
```

Ideally we wouldn't want to add more than a few hundred MB to our current monolithic image if we were to include pub support. Would it be possible to trim it down that much?

pub/README.md
```diff
@@ -0,0 +1,188 @@
...
+
+Dart (pub) support for [`dependabot-core`][core-repo].
+
+### Limitations
+
+ - No support for updating git-dependencies
+   * `dart pub` in general doesn't read versions from git, so upgrade logic is limited.
+   * Some variant of support for updating git-dependencies could be added in the future.
+ - No support for security advisory integration.
+ - If the version found is ignored (by dependabot config) no update will happen (even if, an earlier version could be used)
```
mctofu, line 11: This is the right behavior. We don't expect Dependabot to ever downgrade a dependency.

pub/lib/dependabot/pub/helpers.rb
```diff
@@ -0,0 +1,134 @@
...
+      end
+
+      def dependency_services_report
+        sha256 = Digest::SHA256.new
+        dependency_files.each do |f|
+          sha256 << f.path + "\n" + f.content + "\n"
+        end
+        hash = sha256.hexdigest
+
+        cache_file = "/tmp/report-#{hash}-pid-#{Process.pid}.json"
```
mctofu, line 27: @dependabot/reviewers: Is this the right spot to maintain a cache that persists across dependency updates within an update job? This looks like it would give a performance boost to version updates by not having to re-resolve this for each dependency whereas `SharedHelpers.in_a_temporary_directory` would get reset between each dependency.

pub/spec/dependabot/pub/file_updater_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+require "dependabot/pub/file_updater"
+require "webrick"
+
+require_common_spec "file_updaters/shared_examples_for_file_updaters"
+
+RSpec.describe Dependabot::Pub::FileUpdater do
+  it_behaves_like "a dependency file updater"
+
+  before(:all) do
+    # Because we do the networking in dependency_services we have to run an
```
mctofu, line 15: This is great that we are able to setup fixtures for these tests!

mctofu (commented):
Sharing some early feedback after a first pass through this. It's looking pretty solid and has good test coverage :zap:. I think we will need to address the size issue before merging the Dockerfile changes.

pub/lib/dependabot/pub/file_updater.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      def self.updated_files_regex
+        [
+          /^pubspec\.yaml$/,
+          /^pubspec\.lock$/
+        ]
+      end
+
+      def updated_dependency_files
+        dependency_services_apply(@dependencies)
```
mctofu, line 32: Is it possible for a dependency update to trigger executing code provided by that dependency? I think the answer is currently no but wanted to confirm. If it is we'll put some additional restrictions in place regarding private repo/registry access (which won't affect anything here since it's not supported yet).

Examples of this behavior in other ecosystems:
- https://medium.com/ochrona/arbitrary-code-execution-during-python-package-installation-3a60990350ef
- https://medium.com/@kyle_martin/understanding-and-protecting-against-malicious-npm-package-lifecycle-scripts-8b6129619d7c

pub/lib/dependabot/pub/helpers.rb
jurre: Yeah I think so, since the folder is owned by `root` I was worried the `dependabot` user might not have write access, but looks like it does:

```
w[dependabot-core-dev] ~/dependabot-core $ whoami
dependabot
[dependabot-core-dev] ~/dependabot-core $ irb
irb(main):001:0> File.write("/tmp/permissions-test", "testing 123, is this thing on?")
=> 30
irb(main):002:0> exit
[dependabot-core-dev] ~/dependabot-core $ cat /tmp/permissions-test
testing 123, is this thing on?
```

👍

pub/lib/dependabot/pub/helpers.rb
```diff
@@ -0,0 +1,134 @@
...
+        report = JSON.parse(run_dependency_services("report"))["dependencies"]
+        File.write(cache_file, JSON.generate(report))
+        report
+      end
+
+      def dependency_services_apply(dependency_changes)
+        run_dependency_services("apply", stdin_data: dependencies_to_json(dependency_changes)) do
+          dependency_files.map do |f|
+            updated_file = f.dup
+            updated_file.content = File.read(f.name)
```
jurre, line 39: Maybe I'm missing something obvious, but what's the reason for duplicating the file and reading the contents of the original file into it? Could we not return the actual file instead?

pub/lib/dependabot/pub/metadata_finder.rb
```diff
@@ -0,0 +1,35 @@
...
+  module Pub
+    class MetadataFinder < Dependabot::MetadataFinders::Base
+      private
+
+      def look_up_source
+        repo = pub_listing.dig("latest", "pubspec", "repository")
+        return nil unless repo
+
+        Source.from_url(repo)
+      end
```
jurre, line 18: I think this will actually give us a bunch of Dependabot's default behavior out of the box, so this is 👍

pub/lib/dependabot/pub/version.rb
```diff
@@ -0,0 +1,48 @@
...
+  module Pub
+    class Version < Gem::Version
+      VERSION_PATTERN = Gem::Version::VERSION_PATTERN + "(\\+[0-9a-zA-Z\\-.]+)?"
+      ANCHORED_VERSION_PATTERN = /\A\s*(#{VERSION_PATTERN})?\s*\z/.freeze
+
+      attr_reader :build_info
+
+      def initialize(version)
+        @version_string = version.to_s
+        version, @build_info = version.to_s.split("+") if version.to_s.include?("+")
```
jurre, line 26: I don't see this `build_version` referenced anywhere, if we don't need it, maybe we can just drop the attr_reader and defining it?

```suggestion
        version, _ = version.to_s.split("+") if version.to_s.include?("+")
```

pub/lib/dependabot/pub/helpers.rb
sigurdm: Hmm, I think we are trying to create a Dependabot::DependencyFile that is exactly as the original just with updated content without knowing about the details of Dependabot::DependencyFile. Alternatively we should do something like:

```suggestion
            updated_file = DependencyFile.new(
              name: f.name,
              content: File.read(f.name),
              directory: f.directory,
              support_file: f.support_file?
            )
```

This is more fragile to changes in DependencyFile, but they will probably be rare...

pub/lib/dependabot/pub/update_checker.rb
```diff
@@ -0,0 +1,98 @@
...
+  module Pub
+    class UpdateChecker < Dependabot::UpdateCheckers::Base
+      include Dependabot::Pub::Helpers
+
+      def initialize(dependency:, dependency_files:, repo_contents_path: nil,
+                     credentials:, ignored_versions: [],
+                     raise_on_ignored: false, security_advisories: [],
+                     requirements_update_strategy: nil,
+                     options: {}, pub_hosted_url: nil)
+        @pub_hosted_url = pub_hosted_url
```
jurre, line 17: This is a bit of a nit, but could we fold the `pub_hosted_url` into the `options` hash? That way we don't need to alter the initializer for this class:

```suggestion
                     options: {})
        @pub_hosted_url = options.fetch(:pub_hosted_url, nil)
```

pub/lib/dependabot/pub/requirement.rb
```diff
@@ -0,0 +1,122 @@
...
+    class Requirement < Gem::Requirement
+      quoted = OPS.keys.map { |k| Regexp.quote(k) }.join("|")
+      version_pattern = Pub::Version::VERSION_PATTERN
+
+      PATTERN_RAW = "\\s*(#{quoted})?\\s*(#{version_pattern})\\s*"
+      PATTERN = /\A#{PATTERN_RAW}\z/.freeze
+
+      # Use Pub::Version rather than Gem::Version to ensure that
+      # pre-release versions aren't transformed.
+      def self.parse(obj)
```
jurre, line 22: It's unfortunate that we need to do this at all, I quickly looked if there's a way to avoid doing this for pub, since most of the logic is handled by pub itself, but I'm worried that lot's of things will break in dependabot's current form. Eventually though I think we should be able to get rid of this, once we have a higher level API that we can implement against.

jurre (commented):
@jonasfj @sigurdm overall this looks pretty good, @mctofu and I have left some questions in-line, the only thing that I'm a little concerned about before running this on customer workloads is the size of the image, other than that it's mostly just some questions and optional suggestions.

If there's no way to reduce the size of the resulting image drastically we have a few workarounds that we can implement on our end and we can prioritize those soon.

I'll carve out some time to test this against a few real world repositories, and then see if we can give it a spin on our staging environment.

pub/lib/dependabot/pub/helpers.rb
jurre: Ah gotcha, we've read the file contents from disk already into memory, but they have changed on disk and we need to pull in the updated contents 👍 this approach seems good, I think it's the most concise way to write it so let's keep as is!

sigurdm:
>  the only thing that I'm a little concerned about before running this on customer workloads is the size of the image, other than that it's mostly just some questions and optional suggestions.

I'm looking into shrinking the image... however I get an error when I build the image, even from `main`:
```
> docker build -f Dockerfile -t dependabot/dependabot-core .
Sending build context to Docker daemon  72.18MB
Step 1/65 : FROM ubuntu:18.04
 ---> 886eca19e611
Step 2/65 : SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 ---> Using cache
 [...]
Removing intermediate container d6d858b09572
 ---> 31a8c9b3dbcc
Step 55/65 : RUN bash /opt/bundler/helpers/v1/build
 ---> Running in bfe798d8dbc5
mkdir: cannot create directory ‘/opt/bundler/v1’: Permission denied
```
Am I holding it wrong?

pub/lib/dependabot/pub/version.rb
jonasfj: Yeah, I guess we could throw it away..

Though if instances of `Pub::Version` are every compared or ordered, we ought to consider `build_info`.
I guess we should consider overriding `eql?`: https://docs.ruby-lang.org/en/2.6.0/Gem/Version.html#method-i-eql-3F

jurre:
> I'm looking into shrinking the image... however I get an error when I build the image, even from `main`:

> Am I holding it wrong?

Taking a look now, what you're doing looks ok 🤔

jurre:
So, this worked for me:

```
dependabot-core pub-support % docker build . -t dependabot/dependabot-core
[+] Building 918.9s (47/47) FINISHED

docker --version
Docker version 20.10.11, build dea9396
```

Trying the command as you typed it now, but it takes a minute to complete 😆/😭

sigurdm:
> dependabot-core pub-support % docker build . -t dependabot/dependabot-core

That gives me the same error - something seems off on my side.

If I change the last instance:
```diff
-USER dependabot
+USER root
```
It builds. Not sure why it would behave differently for me.

pub/lib/dependabot/pub/file_updater.rb
sigurdm: No, that would be a bug in pub.

jurre:
Interesting, the folder that it's complaining about on your end seems to be owned by dependabot on my image:

```
drwxr-xr-x   1 dependabot dependabot 4096 Jan 31 16:52 opt
```

For testing the image size it's totally fine to change that line though, so I'd just go with that for now

pub/README.md
sigurdm: This is not about downgrading. The scenario is as follows:
Suppose we have 3 versions of a dependency `dep`:

* dep 1.0.0 <- current
* dep 1.5.0 <- available
* dep 1.7.0 <- ignored by dependabot

One could hope dependabot would suggest upgrading to 1.5.0 as it is not ignored. But because `dependency_services` doesn't know about the ignored versions it will suggest 1.7.0, and that will get filtered away by this implementation. So by ignoring 1.7.0 you also don't get 1.5.0.

sigurdm:
I pushed a commit that deletes everything in the flutter install not needed for pub resolution.
```
> docker image inspect dependabot/dependabot-core:latest --format='{{.Size}}'
4877462066
```

pub/lib/dependabot/pub/update_checker.rb
sigurdm: I was considering that, but sure I was "allowed"

Done

pub/README.md
jurre: Something I was wondering, would it make sense for `dependency_services` to know about the ignored versions? For the initial beta release I think this would be fine, but it'd be nice if we could push the ignored versions into pub and have it resolve things using that.

This would also go some way towards dependabot not having to worry about `pub` requirements, since it could just forward the `ignored_versions` as strings.

I can also imagine this makes sense for end users when updating a package via the CLI, I've definitely been in situations where I _know_ version x of some dependency is not compatible for whatever reason, and if I can tell the package manager "update my dependency but not to 1.7.0" that would be nice.

Alternatively,  could `dependency_services` return a list of possible versions to update to, which dependabot could then filter to exclude any ignored versions?

Not expecting this to be fixed right now, just riffing on possible solutions for the future.

pub/lib/dependabot/pub/version.rb
jurre: Yeah that makes sense 👍 I'd say we can always pull it in when we need it, but I don't feel strongly about removing it or keeping it for now, I'll leave it up to y'all

jurre:
@sigurdm @jonasfj this PR is still marked as draft, is there anything we can do from our end to mark it as ready for review?

pub/README.md
sigurdm: Yes I think it would absolutely make sense.

The pub version solver currently doesn't have a concept of ignored versions, so without changes to the solver itself pub would not be able to use that information thats why we left it out of this initial MVP.

It could certainly be done.

Dockerfile
```diff
@@ -271,8 +271,6 @@ RUN curl --connect-timeout 15 --retry 5 "https://storage.googleapis.com/flutter_
     ! -path /opt/dart/flutter \
     -delete
 
-USER root
```
jurre: I am a not entirely sure if this change is desirable, I know it helped unblock some things for local development on your machine, but I think we want these files to be owned by root in production. I was able to build the image successfully from this branch without needing to make this change, even after clearing all my docker caches 🤔 

Could you try rebasing onto main and see if that resolves the issue? Otherwise I can dig into it and see if I can figure out what happened here.

sigurdm: This commit is actually (at least that was the purpose) undoing an unwanted change from this pr. If you look at the global view (https://github.com/dependabot/dependabot-core/pull/4510/files) you see that this pr now neither adds or removes USER statements.

jurre: Ah apologies I missed that 👍

sigurdm:
We now have merged the dependency-services into the mainline pub.
I updated the dockerfile to use that commit.

I think this is good for review.

jurre:
Testing some dry-run updates, and I ran into this scenario which is unexpected to me:

```
[dependabot-core-dev] ~/dependabot-core $ bin/dry-run.rb pub dart-lang/pub --requirements-update-strategy=lockfile-only --dep=analyzer
=> cloning into /home/dependabot/dependabot-core/tmp/dart-lang/pub
=> parsing dependency files
=> updating 1 dependencies: analyzer

=== analyzer (2.8.0)
 => checking for updates 1/1
 => latest available version is 3.2.0
 => latest allowed version is 3.2.0
 => requirements update strategy: lockfile-only
 => updating analyzer from 2.8.0 to 3.2.0

    ± pubspec.yaml
    ~~~
    <   analyzer: ^2.7.0
    ---
    >   analyzer: ^3.2.0
    ~~~
```

I would not expect an update using the `lockfile-only` strategy to perform a breaking major update, or touch the pubspec.

fwiw `lockfile-only` performs a `can_update?(requirements_to_unlock: :none)`  on the `UpdateChecker`.

This example is a little contrived as `dart-lang/pub` does not use a lockfile, but the same behavior can be observed when a lockfile does exist, for example:


```
# pubspec.yaml

name: dependabot_testcase
environment:
  sdk: '>=2.12.0 <3.0.0'
dependencies:
  collection: ^1.14.13 # Locked to 1.14.13, can update with no unlock.
  retry: ^2.0.0 # Can update with updated constraint, no further constraints.
  protobuf: 1.1.4 # Can update with updated constraint, only together with fixnum.
  fixnum: 0.10.11
  path: 1.8.1 # Already at latest
```

```
# pubspec.lock

# Generated by pub
# See https://dart.dev/tools/pub/glossary#lockfile
packages:
  collection:
    dependency: "direct main"
    description:
      name: collection
      url: "https://pub.dartlang.org"
    source: hosted
    version: "1.14.13"
  fixnum:
    dependency: "direct main"
    description:
      name: fixnum
      url: "https://pub.dartlang.org"
    source: hosted
    version: "0.10.11"
  path:
    dependency: "direct main"
    description:
      name: path
      url: "https://pub.dartlang.org"
    source: hosted
    version: "1.8.1"
  protobuf:
    dependency: "direct main"
    description:
      name: protobuf
      url: "https://pub.dartlang.org"
    source: hosted
    version: "1.1.4"
  retry:
    dependency: "direct main"
    description:
      name: retry
      url: "https://pub.dartlang.org"
    source: hosted
    version: "2.0.0"
sdks:
  dart: ">=2.12.0 <3.0.0"
```

```
bin/dry-run.rb pub dsp-testing/dependabot-dart --requirements-update-strategy=lockfile-only --dep=retry
=> cloning into /home/dependabot/dependabot-core/tmp/dsp-testing/dependabot-dart
=> parsing dependency files
=> updating 1 dependencies: retry

=== retry (2.0.0)
 => checking for updates 1/1
 => latest available version is 3.1.0
 => latest allowed version is 3.1.0
 => requirements update strategy: lockfile-only
 => updating retry from 2.0.0 to 3.1.0

    ± pubspec.yaml
    ~~~
    6c6
    <   retry: ^2.0.0 # Can update with updated constraint, no further constraints.
    ---
    >   retry: ^3.1.0 # Can update with updated constraint, no further constraints.
    ~~~

    ± pubspec.lock
    ~~~
    38c38
    <     version: "2.0.0"
    ---
    >     version: "3.1.0"
    ~~~
```

I would expect to see an update to the latest version that respects the pubspec constraint, which in this case would be no update, as it looks there's only one 2.x.x release of `retry`: https://pub.dev/packages/retry/versions

pub/lib/dependabot/pub/update_checker.rb
```diff
@@ -0,0 +1,85 @@
...
+        # Version we can get if we're not allowed to change pubspec.yaml, but we
+        # allow changes in the pubspec.lock file.
+        entry = current_report["compatible"].find { |d| d["name"] == dependency.name }
+        return nil unless entry
+
+        new_version = Dependabot::Pub::Version.new(entry["version"])
+        # We ignore this solution, if any of the requirements in
+        # ignored_versions satisfy the version we're proposing as an upgrade
+        # target.
+        return nil if ignore_requirements.any? { |r| r.satisfied_by(new_version) }
```
jurre, line 26:
```suggestion
        return nil if ignore_requirements.any? { |r| r.satisfied_by?(new_version) }
```

jurre (changes requested):
I noticed an error when trying to use an ingored dependency, here's the dry-run I used to surface this:

```
IGNORE_CONDITIONS='[{"dependency-name":"sqflite","version-requirement":"2.0.2"}]' bin/dry-run.rb pub alibaba/flutter-go --dep=sqflite
```

pub/lib/dependabot/pub/version.rb
sigurdm: I think we'll leave this for now.

jurre:
@sigurdm at this point I have two open questions: 

- None of the versioning-strategies seem supported, maybe this is fine? Is this by design on your end? We'd need to document  it.
- I'd like to disable metric collection, any concerns about that? Would you like me to push a commit that does this?

Thanks for pushing the typo fix, I'll double check that ignored versions work as expected now. 

With those last questions addressed I think things are looking good to merge.

Dockerfile
```diff
@@ -222,6 +222,54 @@ RUN apt-add-repository "deb [arch=amd64] https://apt.releases.hashicorp.com $(ls
...
+  && dart --version
+# HACK: We use a patched version of "dart pub" that is not in the Dart SDK yet.
+#       Hence, we clone from git and globally install it from local folder.
+#       Now it can be used with: 'dart pub global run pub ...'
+RUN git clone https://github.com/dart-lang/pub.git /opt/dart/pub \
+  && git -C /opt/dart/pub checkout 941191f7f83ad60259348860197cfcdd83bb8e6f \
+  && dart pub global activate --source path /opt/dart/pub \
+  && chmod -R o+r "/opt/dart/pub" \
+  && chown -R dependabot:dependabot "$PUB_CACHE" \
+  && chown -R dependabot:dependabot /opt/dart/pub
```
jurre, line 248: Now that the dependency services API has landed, can we clean this up, or do we still need the manual checkout?

jurre:
Still somewhat confused by how ignored dependencies work, we don't crash now (🎉), but:

```
IGNORE_CONDITIONS='[{"dependency-name":"sqflite","version-requirement":"2.0.2"}]' bin/dry-run.rb pub alibaba/flutter-go --dep=sqflite
=> cloning into /home/dependabot/dependabot-core/tmp/alibaba/flutter-go
=> parsing dependency files
=> updating 1 dependencies: sqflite

=== sqflite (1.3.2+4)
 => checking for updates 1/1
 => latest available version is 2.0.2
 => latest allowed version is 1.3.2+4
 => requirements to unlock: all
    (no update possible, peer dependency can be updated)
```

There is a version `2.0.1` available, so I'd expect that we'd update to that version, but instead it assumes no update is possible.

mctofu:
> There is a version `2.0.1` available, so I'd expect that we'd update to that version, but instead it assumes no update is possible.

I think that is what was called out in the [limitations](https://github.com/dependabot/dependabot-core/pull/4510/files#r798526444):
> If the found version is ignored (by dependabot config) no update will happen (an earlier version is not used)

sigurdm:
> I think that is what was called out in the [limitations](https://github.com/dependabot/dependabot-core/pull/4510/files#r798526444):

Yes - that is exactly what we try to express. I'll try to write a test of this behavior.

sigurdm:
> I'd like to disable metric collection, any concerns about that? Would you like me to push a commit that does this?

Metric collection from the pub client? I think it would be fine to disable those. 

> None of the versioning-strategies seem supported, maybe this is fine? Is this by design on your end? We'd need to document it.

I don't remember that we took an active decision here.

I think in general we recommend upgrading the lower bound instead of widening (https://dart.dev/tools/pub/dependencies#best-practices) For library packages this has to with how unlikely it is that tests are run across major versions of dependencies. So I think at least for the first release supporting only that strategy is ok.

Where should we document this?

jurre:
> Where should we document this?

We can take care of this on our end 👍

Dockerfile
jurre: Cool, thanks for clarifying!

jurre:
I think this looks good and is ready to merge! I'll follow up with a small PR to disable metrics collection. I'm going to do this + release a new version of dependabot-core on Monday, after which we will run a short private beta internally to make sure everything integrates well. I'll reach out to y'all when this is available if you want to enable it on some of your repositories @sigurdm @jonasfj.

Once it looks good there, we can open this up for everyone to test.

jonasfj:
>  I'll follow up with a small PR to disable metrics collection.

I don't think our [`dependency_services`](https://github.com/dart-lang/pub/blob/master/bin/dependency_services.dart) will send any metrics.. It'll make requests to `pub.dev` and those request will might have headers indicating that they come from dependabot. IMO, that's fine, but we can double check if they even send anything, since those headers are often disabled when `CI=true`.

I guess we could minimize analytics from `dart pub global run` by using `dart --disable-analytics pub global run ...` instead. We don't really want those analytics anyways (I think they are also disabled by `CI=true`, but I'd have to check).

jurre:
> > I'll follow up with a small PR to disable metrics collection.
> 
> I don't think our [`dependency_services`](https://github.com/dart-lang/pub/blob/master/bin/dependency_services.dart) will send any metrics.. It'll make requests to `pub.dev` and those request will might have headers indicating that they come from dependabot. IMO, that's fine, but we can double check if they even send anything, since those headers are often disabled when `CI=true`.
> 
> I guess we could minimize analytics from `dart pub global run` by using `dart --disable-analytics pub global run ...` instead. We don't really want those analytics anyways (I think they are also disabled by `CI=true`, but I'd have to check).

Thanks for clarifying 👍 

I agree headers make sense and that's fine. Can we disable things globally by running `dart --disable-analytics` in the Dockerfile, or should we do it in the code where we call out to `dart` like you suggested?

I don't think we set `CI=true` for Dependabot jobs currently (outside of in our CI environment that is), any other places this could lead to unexpected behavior?

jurre (approved):

jonasfj:
> Can we disable things globally by running `dart --disable-analytics` in the Dockerfile

Yeah, but we're changing user from `root` to `dependabot`, and I think the file is stored in `$HOME` somewhere... So probably best we just pass a flag for every invocation.

PR #4517, opened 2021-12-13
Reinstate the dev container build, pushing to GHCR

brrygrdn:
Follows up on #4511, #4513 

This reinstates the matrix build to push the development image of core, using GHCR as a target instead of docker.io.

Since this image is a development tool, it makes less sense to have it on `docker.io` for general consumption, GHCR is a better home for this - and likely our CI image which I will migrate in a follow up.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -8,12 +8,12 @@ on:
       - main
     tags:
       - v[0-9]+.[0-9]+.[0-9]+
-permissions:
-  contents: read
 jobs:
   push-core-image:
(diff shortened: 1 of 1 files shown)


mattt (approved):
Nice work! Once this is merged and runs for the first time, we can make the development container image publicly readable following [these instructions](https://docs.github.com/en/packages/learn-github-packages/configuring-a-packages-access-control-and-visibility).

brrygrdn:
It seems the container image defaulted to publicly readable, presumably via the inherited access 👍🏻

PR #4518, opened 2021-12-13
Use npm 7 for lockfileVersion 3

### LINKED ISSUE #4361: package-lock files with version 3 are not handled correctly by Dependabot

ferferga:
This should fix #4361 

Ideally, Dependabot should be updated to npm 8 (as it's the version that comes with node 16, the latest LTS version of node, contrary to npm 7). npm 8 isn't that different from npm 7 (besides a few small fixes and the deprecation of node 12, which should not be a problem for this).

However, this PR is much simpler and straightforward fix that directly upgrading to npm 8 and it shouldn't break anything, while keeping proper compatibility with the people that already wanted to have their setups upgraded to the v3 ``lockfileVersion``

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
@@ -5,6 +5,7 @@ module Dependabot
     module Helpers
       def self.npm_version(lockfile_content)
         return "npm7" unless lockfile_content
+        return "npm7" if JSON.parse(lockfile_content)["lockfileVersion"] == 3
         return "npm7" if JSON.parse(lockfile_content)["lockfileVersion"] == 2
 
         "npm6"


jurre:
Thanks, I think this should fix things 🎉

It should also be safe to update npm in that case, but let's do it in a separate PR

ferferga:
Yes, that's in my roadmap :)

I wanted to get this one up as-is so it can be deployed to production quickly and have most projects that rely on this up and running again.
 
The npm 8 update can come once I'm more knowledgeable of the codebase and no urgency is needed for deployment and testing :)

jurre:
The bundler and cargo CI failures are unrelated, but it'd be great if we can add a few tests that uses the new lockfile version! Are you comfortable adding one?

ferferga:
I've been looking at how to acomplish it, but sadly this is my first contact with Ruby and so far no luck at all. I've searched for a test that verifies that npm7 is used when the lockfileVersion key is set to 2 (in order to use it as a template), but I couldn't find it in any of the test specs (which, if I'm not wrong, are located under ``npm_and_yarn\spec\dependabot\npm_and_yarn``).

jurre:
No worries @ferferga I can look into adding some test coverage 👍

jurre:
It looks like this isn't quite working yet, when I tried to run an update using this new patch, it still ends up downgrading the lockfile from v3 to v2:


```
 => updating jest from 26.6.3 to 27.4.5

    ± package.json
    ~~~
    124c124
    <     "jest": "^26.6.3",
    ---
    >     "jest": "^27.4.5",
    ~~~

    ± package-lock.json
    ~~~
    4c4
    <   "lockfileVersion": 3,
    ---
    >   "lockfileVersion": 2,

# lot's of changes omitted
```

I'm running out of the time that I've set out to help get this landed, I'll push the tests etc that I have so far but I'll need to come back to it later.

Fwiw here's what I'd been running for a full update:

```
bin/dry-run.rb npm_and_yarn jellyfin/jellyfin-vue --dep=jest
```

It looks like we need to make some changes to `npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb` and add test coverage around it ensuring that we don't downgrade v3 lockfiles.

As a last step I'm going to try installing npm 8 in the docker container and see how that changes things, but I don't think I'll be able to land that change today.

[this](https://github.com/dependabot/dependabot-core/blob/c8350f1d93fc16918a55ba29d00c4e2ee42aeb9a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb#L183-L192) is the npm command that we end up running, and I was able to verify that the `.npmrc` with the `lockfile-version=3` is present in the directory that we run the command in.

jurre:
From a quick search I can only seem to find a handful of projects on GitHub using this v3 lockfile format, so I'll have to prioritize this accordingly

jurre:
Looks like installing npm@v8 did help, but will possibly break a bunch of things.

My changes are here: https://github.com/dependabot/dependabot-core/pull/4531

ferferga:
Thank you very much for taking the time to do all this testing. I added some small feedback there to potentially make things easier for the future. Dependabot doesn't really use node for anything else besides npm (from what I could tell looking at the codebase) so upgrading to Node 16 (or always stick to LTS) should not be a problem at all.

As of the handful of projects, yes, there are not many right now. We upgraded because installing dependencies is much faster and in (probably) the next version, [lockfileVersion 3 will be the default](https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json#lockfileversion) so I think it makes sense to support it as soon as possible (but without neglecting bugs or other critical stuff).

As of npm 8 upgrade [here you can see how it's basically an npm 7 in steroids](https://github.com/npm/cli/releases/tag/v8.0.0).

I'm closing this PR in favor of yours. I'll try to help you as much as I can in the node side of things, as unfortunately Ruby is not my strong point 😅.

PR #4524, opened 2021-12-14
Update the CI workflow to use GHPR for branch images and as a mirror for CI images

brrygrdn:
This PR follows on from #4523 to increase our use of GHPR.

The high level intent of this change is to move our development images out of dockerhub as they aren't relevant to downstream consumers of Dependabot.

Rather than a breaking change where everyone's builds are suddenly much slower, this is a gradual switch over that just mirrors the CI image to GHPR for now, we'll follow up with a change to consume from this image in a few days once it has an opportunity to propagate on a few WIP PRs.

## Changes to branch builds

One major change is that we _immediately_ start pushing and consuming the branch-specific build of `dependabot-core` to and from GHPR instead of the main `dependabot/dependabot-core` image.

This is due to a decision the team has made that these branch builds should live in a separate project to avoid any confusion over which refs of `dependabot/dependabot-core` are properly tested for distribution.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,16 +36,16 @@ jobs:
         - { path: omnibus, name: omnibus }
         - { path: python, name: python }
         - { path: terraform, name: terraform }
+    env:
+      BASE_IMAGE: ubuntu:18.04
+      CORE_IMAGE: dependabot/dependabot-core
+      CORE_CI_IMAGE: dependabot/dependabot-core-ci
+      CODE_DIR: /home/dependabot/dependabot-core
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
-      - name: Prepare environment variables
-        run: |
-          echo "BRANCH_REF=$(echo '${{ github.ref }}' | sed -E 's/[^A-Za-z0-9]+/-/g')" >> $GITHUB_ENV
-          echo "BASE_IMAGE=ubuntu:18.04" >> $GITHUB_ENV
-          echo "CORE_IMAGE=dependabot/dependabot-core" >> $GITHUB_ENV
-          echo "CORE_CI_IMAGE=dependabot/dependabot-core-ci" >> $GITHUB_ENV
-          echo "CODE_DIR=/home/dependabot/dependabot-core" >> $GITHUB_ENV
+      - name: Prepare BRANCH_REF environment variable
+        run: echo "BRANCH_REF=$(echo '${{ github.ref }}' | sed -E 's/[^A-Za-z0-9]+/-/g')" >> $GITHUB_ENV
       - name: Log in to Docker registry
         run: |
           if [ -n "${{ secrets.DOCKER_USERNAME }}" ] && [ -n "${{ secrets.DOCKER_PASSWORD }}" ]; then
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,8 +2,6 @@ name: CI
 on:
   push:
     branches:
-      - "actions/**"
-      - "wip/**"
       - "main"
   pull_request:
     branches:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,7 @@ jobs:
       CORE_IMAGE: dependabot/dependabot-core
       CORE_BRANCH_IMAGE: ghcr.io/dependabot/dependabot-core-branch
       CORE_CI_IMAGE: dependabot/dependabot-core-ci
+      CORE_CI_MIRROR: ghcr.io/dependabot/dependabot-core-ci
       CODE_DIR: /home/dependabot/dependabot-core
     steps:
       - name: Checkout code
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -2,16 +2,15 @@ name: CI
 on:
   push:
     branches:
-      - "actions/**"
-      - "wip/**"
```
brrygrdn, line 6: We don't have any branches that meet these patterns are present. I elected to remove this to avoid spending CI time on anything that isn't either on main or merging to main.

We can reinstate these if and when we have reason to start building long-lived branches in future.

.github/workflows/ci.yml
```diff
@@ -2,16 +2,15 @@ name: CI
...
       - "main"
   pull_request:
     branches:
-      - "**"
+      - "main"
```
brrygrdn, line 8: As above, our CI build time is quiet long so this is a conservation measure that we only build things with intent to merge to main. Branch-to-branch merging is not a common workflow for us.

.github/workflows/ci.yml
```diff
@@ -56,23 +57,28 @@ jobs:
...
-          DOCKER_BUILDKIT=1 docker build \
-            -t "$CORE_IMAGE:latest" \
-            -t "$CORE_IMAGE:branch--$BRANCH_REF" \
+          docker build \
+            -t "$CORE_BRANCH_IMAGE:$BRANCH_REF" \
             --build-arg BUILDKIT_INLINE_CACHE=1 \
             --cache-from "$BASE_IMAGE" \
-            --cache-from "$CORE_IMAGE:branch--$BRANCH_REF" \
             --cache-from "$CORE_IMAGE:latest" \
+            --cache-from "$CORE_BRANCH_IMAGE:$BRANCH_REF" \
```
brrygrdn, line 72: This is an intentional format change since the `branch` name has folded into the container repository.

Old: `dependabot/dependabot-core:branch--refs-pull-4511-merge`
New: `ghcr.io/dependabot/dependabot-core-branch:refs-pull-4511-merge`

.github/workflows/ci.yml
```diff
@@ -85,6 +91,12 @@ jobs:
         run: |
           docker push "$CORE_CI_IMAGE:latest"
           docker push "$CORE_CI_IMAGE:branch--$BRANCH_REF"
+      - name: Push dependabot-core-ci image to GHCR mirror
+        run: |
+          docker tag "$CORE_CI_IMAGE:latest" "$CORE_CI_MIRROR:latest"
+          docker push "$CORE_CI_MIRROR:latest"
+          docker tag "$CORE_CI_IMAGE:latest" "$CORE_CI_MIRROR:branch--$BRANCH_REF"
+          docker push "$CORE_CI_MIRROR:branch--$BRANCH_REF"
```
brrygrdn, line 99: I've left it alone for now, but one thing of note is that I believe pushing `latest` or `branch--$BRANCH_REF` should be mutually exclusive depending on the ref being built.

We should only really be pushing `latest` when building `main` as otherwise the `latest` image is actually juddering between CI builds, meaning it's not clear if each build is caching from `main` or a sibling CI build on another PR. I think this may mean our builds are currently subject to some fuzzing 🤔

.github/workflows/ci.yml
```diff
@@ -38,16 +37,18 @@ jobs:
...
         uses: actions/checkout@v2
-      - name: Prepare environment variables
-        run: |
-          echo "BRANCH_REF=$(echo '${{ github.ref }}' | sed -E 's/[^A-Za-z0-9]+/-/g')" >> $GITHUB_ENV
-          echo "BASE_IMAGE=ubuntu:18.04" >> $GITHUB_ENV
-          echo "CORE_IMAGE=dependabot/dependabot-core" >> $GITHUB_ENV
-          echo "CORE_CI_IMAGE=dependabot/dependabot-core-ci" >> $GITHUB_ENV
-          echo "CODE_DIR=/home/dependabot/dependabot-core" >> $GITHUB_ENV
+      - name: Prepare BRANCH_REF environment variable
+        run: echo "BRANCH_REF=$(echo '${{ github.ref }}' | sed -E 's/[^A-Za-z0-9]+/-/g')" >> $GITHUB_ENV
```
mattt, line 51: Taking another look at this, are we accounting for the possibility of multiple PRs coming from different forks with the same downstream branch name? Looking at [the docs](https://docs.github.com/en/actions/learn-github-actions/contexts), it's not obvious to me that `github.ref` captures this...

mattt: Oh wait, never mind. Please disregard my last comment. `github.ref` uses the PR number (not branch name), which is indeed unique.

jurre (approved):
🥇

.github/workflows/ci.yml
Nishnha: I noticed that a [fork of Dependabot-Core](https://github.com/dependabot/dependabot-core/runs/4644267200?check_suite_focus=true#step:6:118) is _correctly_ not able to push its branch but the error message is unclear on how to fix this.

I'm not sure whether this error is produced from a failing to push to `dockerhub` or `ghcr`, so maybe we should be more explicit about disallowing forked repo/branch CI runs and exit early.

Nishnha: Resolved by https://github.com/dependabot/dependabot-core/pull/4569

PR #4525, opened 2021-12-14
Disable `docker` workflow on forks

mattt:
Related to #4524 

When a fork of `dependabot-core` pushes a new commit, it triggers the ["Push docker images"](https://github.com/dependabot/dependabot-core/blob/main/.github/workflows/docker.yml) workflow. However, that workflow fails unless `DOCKER_USERNAME` and `DOCKER_PASSWORD` secrets are configured, which will be the case for pretty much everyone.

Screen Shot 2021-12-14 at 06 27 54

This PR conditionalizes the `push-core-image` step to run only in the `dependabot/dependabot-core` repository.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -13,6 +13,7 @@ jobs:
   push-core-image:
     name: Push dependabot-core image to docker hub
     runs-on: ubuntu-latest
+    if: github.repository == "dependabot/dependabot-core"
     permissions:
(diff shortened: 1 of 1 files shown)


jurre (approved):

brrygrdn (approved):
👍🏻  Good call, this makes sense.

landongrindheim (approved):

PR #4528, opened 2021-12-14
Allow whitespace at beginning of Cargo dependency declarations

Nishnha:
Allows `cargo.toml` manifest files to have arbitrary whitespace at the beginning of dependency declarations.

Includes a new fixture file that contains dependencies with tabs or spaces at the start of their declaration.

### DIFF
--- a/cargo/lib/dependabot/cargo/file_updater/manifest_updater.rb
+++ b/cargo/lib/dependabot/cargo/file_updater/manifest_updater.rb
@@ -139,7 +139,7 @@ module Dependabot
         end
 
         def declaration_regex(dep)
-          /(?:^|["'])#{Regexp.escape(dep.name)}["']?\s*=.*$/i
+          /(?:^|^\s*|^\t*|["'])#{Regexp.escape(dep.name)}["']?\s*=.*$/i
         end
 
         def feature_declaration_version_regex(dep)
--- a/cargo/lib/dependabot/cargo/file_updater/manifest_updater.rb
+++ b/cargo/lib/dependabot/cargo/file_updater/manifest_updater.rb
@@ -139,7 +139,7 @@ module Dependabot
         end
 
         def declaration_regex(dep)
-          /(?:^|^\s*|^\t*|["'])#{Regexp.escape(dep.name)}["']?\s*=.*$/i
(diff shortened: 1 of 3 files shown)


cargo/lib/dependabot/cargo/file_updater/manifest_updater.rb
```diff
@@ -139,7 +139,7 @@ def update_manifest_pin(content:, dep:, old_pin:, new_pin:)
         end
 
         def declaration_regex(dep)
-          /(?:^|["'])#{Regexp.escape(dep.name)}["']?\s*=.*$/i
+          /(?:^|^\s*|^\t*|["'])#{Regexp.escape(dep.name)}["']?\s*=.*$/i
```
jurre, line 142: Does `\s` (any whitespace character) not include `\t` (tab character)?

Nishnha: Looks like it should. Will update

cargo/spec/fixtures/manifests/whitespace_names
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "dependabot" # the name of the package
+version = "0.1.0"    # the current version, obeying semver
+authors = ["support@dependabot.com"]
+
+[dependencies]
+    time = "0.1.12"
+  regex = "0.1.41"
```
jurre, line 8: Maybe make one of these a tab?

jurre (approved):
Had an optional test-improvement suggestion but not blocking and looks good 👍

cargo/spec/fixtures/manifests/whitespace_names
Nishnha: Hmm the time dependency should be a tab but it might have been converted to a space by git or your editor.
Is there a way to check what it is here? Maybe via the online editor?

Nishnha: It looks like the online editor lets you choose between either tabs or spaces and converts them automagically for you 👎🏾

jurre: Looks like it was spaces, but I converted it to a tab and pushed the change.

PR #4531, opened 2021-12-15
Jurre/npm lockfile v3

jurre:
This change instructs our npm code to use npm8 for lockfiles using the
v2 or v3 format. Npm 8 is mostly backwards compatible but drops support for node 12 (which we don't run).

Npm 7 fully supports this new format, (which is not backwards
compatible with npm 6), and prior to this change they were treated as
npm 6 lockfiles, which would result in the entire lockfile being
rewritten.

Not sure what this will break so opening as a draft

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -97,11 +97,11 @@ USER root
 
 ### JAVASCRIPT
 
-# Install Node 14.0 and npm v7
+# Install Node 14.0 and npm v8
 RUN curl -sL https://deb.nodesource.com/setup_14.x | bash - \
   && apt-get install -y --no-install-recommends nodejs \
   && rm -rf /var/lib/apt/lists/* \
-  && npm install -g npm@v7.21.0 \
+  && npm install -g npm@v8.3.0 \
   && rm -rf ~/.npm
 
 
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
@@ -5,7 +5,7 @@ module Dependabot
     module Helpers
       def self.npm_version(lockfile_content)
         return "npm7" unless lockfile_content
-        return "npm7" if JSON.parse(lockfile_content)["lockfileVersion"] == 2
+        return "npm7" if JSON.parse(lockfile_content)["lockfileVersion"] >= 2
 
         "npm6"
       rescue JSON::ParserError
--- a/npm_and_yarn/spec/dependabot/npm_and_yarn/file_parser/lockfile_parser_spec.rb
+++ b/npm_and_yarn/spec/dependabot/npm_and_yarn/file_parser/lockfile_parser_spec.rb
@@ -310,5 +310,51 @@ RSpec.describe Dependabot::NpmAndYarn::FileParser::LockfileParser do
         )
       end
     end
+
+    context "npm8 with a v3 lockfile-version" do
+      context "workspace project with a direct dependency that's installed in the workspace's node_modules" do
+        let(:dependency_files) { project_dependency_files("npm8/workspace_nested_package_lockfile_v3") }
+        let(:dependency_name) { "yargs" }
+        let(:manifest_name) { "packages/build/package.json" }
+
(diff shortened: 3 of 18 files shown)


Dockerfile
```diff
@@ -97,11 +97,11 @@ USER root
 
 ### JAVASCRIPT
 
-# Install Node 14.0 and npm v7
+# Install Node 14.0 and npm v8
 RUN curl -sL https://deb.nodesource.com/setup_14.x | bash - \
```
ferferga, line 101: Maybe this should be switched to node 16? This way we stick to the current LTS version, which also always include well-tested npm versions (no new features, only bugfixes).

You can use ``https://deb.nodesource.com/setup_lts.x`` to always point to the latest LTS or ``https://deb.nodesource.com/setup_16.x`` to just point to node 16.

This also removes the additional step of installing npm afterwards, which makes the docker building faster :)

Dockerfile
```diff
@@ -97,11 +97,11 @@ USER root
 
 ### JAVASCRIPT
 
-# Install Node 14.0 and npm v7
+# Install Node 14.0 and npm v8
```
shu-mutou, line 100:
```suggestion
# Install Node 16.0 and npm v8
```

Dockerfile
shu-mutou, line 101:
```suggestion
RUN curl -sL https://deb.nodesource.com/setup_16.x | bash - \
```

Dockerfile
```diff
@@ -97,11 +97,11 @@ USER root
...
 RUN curl -sL https://deb.nodesource.com/setup_14.x | bash - \
   && apt-get install -y --no-install-recommends nodejs \
   && rm -rf /var/lib/apt/lists/* \
-  && npm install -g npm@v7.21.0 \
+  && npm install -g npm@v8.3.0 \
```
shu-mutou, line 104:
```suggestion
```

shu-mutou (commented):
As @ferferga said, normally, npm8 will be installed at the same time as nodejs v16.

npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
```diff
@@ -5,7 +5,7 @@ module NpmAndYarn
     module Helpers
       def self.npm_version(lockfile_content)
         return "npm7" unless lockfile_content
-        return "npm7" if JSON.parse(lockfile_content)["lockfileVersion"] == 2
+        return "npm7" if JSON.parse(lockfile_content)["lockfileVersion"] >= 2
```
feelepxyz, line 8: Do we reference this in places?
```suggestion
        return "npm8" if JSON.parse(lockfile_content)["lockfileVersion"] >= 2
```

Was also wondering if it would make sense to ship an incremental fix with this changed to `>= 2` but keep npm7 installed, if it's backwards compatible?

Dockerfile
feelepxyz: 👍  on this change but might be worth splitting this up into a separate PR to make it easier to isolate potential regressions.

Dockerfile
feelepxyz: The docker win is nice but I think we still want to keep this explicit so we can update to new patch releases. Otherwise we're at the mercy of docker caching the node + npm version.

npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
jurre: Yeah that was my thinking. Npm 7 should already support the v3 format

feelepxyz: I had a go at downgrading to npm7 with your tests, looks 🍏  https://github.com/dependabot/dependabot-core/pull/4630 - thougts on shipping this first then figuring out the test failures for npm8?

jurre: Yeah I think that's a great call, let's do it!

jurre:
Going to close this out for now

PR #4542, opened 2021-12-17
Dependency updaters

mattt:


common/lib/dependabot/dependency_updaters/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+| Argument       | Type                     | Description                      |
+| -------------- | ------------------------ | -------------------------------- |
+| `dependency`   | `Dependabot::Dependency` | The package dependency to update |
+| `requirements` | `VersionRange`           | The updated version constraints  |
+
+#### Return
+
+Returns `true` if the dependency was successfully updated 
+to satisfy the new requirements.
```
jurre, line 28: I think it could be super useful if this returned some data like:

- the dependencies that were updated with their `from` and `to` values, wether or not they were top-level or transitive maybe?
- files that were changed

common/lib/dependabot/dependency_updaters/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+Updates a dependency according to the provided requirements.
+This method is called within the root directory of a project.
+
+#### Arguments
+
+| Argument       | Type                     | Description                      |
+| -------------- | ------------------------ | -------------------------------- |
+| `dependency`   | `Dependabot::Dependency` | The package dependency to update |
+| `requirements` | `VersionRange`           | The updated version constraints  |
```
jurre, line 23: For security updates, would we pass in the inverted `affected_versions` of the advisory?

common/lib/dependabot/dependency_updaters/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+Dependabot::DependencyUpdaters.register("pub", Dart::Pub)
+
+# Updating a dependency
+
+dependency = Dependency.new(package_url: "pkg://pub/http_parser")
+requirements = VersionRange.parse(">=4.0.0")
+
+dependency_updater = Dependabot::DependencyUpdaters.for(dependency.package_manager)
+dependency_updater.update(dependency: dependency, requirements: requirements)
```
jurre, line 62: I'm curious how we'd get the information that we need to feed into a PR, some of could be answered by https://github.com/dependabot/dependabot-core/pull/4542/files#r771474851, but there might be other options

jurre: Maybe we should include the other (existing or fictional) classes that make up a complete PR in the doc here?

common/lib/dependabot/dependency_updaters/README.md
mattt: I'm with you. Why settle for `true` when we could return a more useful "truthy" value, right?

My only concern would be adding complexity for a hypothetical benefit. How would we use "from" and "to" values?

As far as file changes, this method is performed within the context of a Git worktree in a filesystem, so we can easily calculate that afterward.

common/lib/dependabot/dependency_updaters/README.md
mattt: Maybe. It depends on which perspective we want to favor. The advisory database gives us ranges affected versions, but package managers express positive requirement.

Any range or enumeration of affected versions can be trivially inverted, so it may not be an issue. I'd be interested to hear what implementors have to say once we have better primitives for requirements.

common/lib/dependabot/dependency_updaters/README.md
mattt: Great call. We should definitely think in terms of the full end-to-end functionality.

A few options to consider:

1. Have this method return diff information, as suggested above
2. Add a new required method (say, `#dependencies`); to get a diff, we run once before `#update` and again afterward and compare.
2a. Alternatively, a `#manifest_files` that returns all manifest files in the project, and a `#dependencies(manifest_file:)` that returns dependencies scoped to a particular manifest file.

common/lib/dependabot/dependency_updaters/README.md
jurre: > My only concern would be adding complexity for a hypothetical benefit. How would we use "from" and "to" values?

In the PR title/description for example

> As far as file changes, this method is performed within the context of a Git worktree in a filesystem, so we can easily calculate that afterward.

Slightly worried we might end up picking up unintended changes, but we can figure out and learn if that's actually a problem or me just being paranoid

common/lib/dependabot/dependency_updaters/README.md
jurre: > Any range or enumeration of affected versions can be trivially inverted, so it may not be an issue. I'd be interested to hear what implementors have to say once we have better primitives for requirements.

👍 

I think we'll also need a way to pass in ignored ranges from the [config file](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#ignore)

mattt: What do you think about passing a `Configuration` object as an argument?

common/lib/dependabot/dependency_updaters/README.md
mattt: > In the PR title/description for example

Let's game that out: How important is it for a PR to say "Update `vulnerable` to version 2.0.1"? Or would something like "Fix <CVE> on `vulnerable` 2.0.0" be sufficient with the raw diff? (I have no idea! That sounds like a product / UX question, so we can ask the relevant stakeholders)

> Slightly worried we might end up picking up unintended changes, but we can figure out and learn if that's actually a problem or me just being paranoid.

Conversely, taking the results verbatim from the package manager operating on the filesystem ensures that the PR doesn't leave out details. For example, a Go update that updates checksum values and a `vendor/` subdirectory, or a theoretical package manager that automatically generates an SBOM.

jurre: > How important is it for a PR to say "Update vulnerable to version 2.0.1"?

From a product perspective not sure but we can ask around.

We do tend to use that information to show metadata about the PR like relevant changelog entries, commits between the versions etc etc. I think that information is very useful. 

I'm more concerned about not having to update all the code that's downstream from these changes that expects that information to be present. 

My thinking is that the more we're able to isolate the changes to a portion of the system the easier it'll be to integrate, and also to move existing ecosystems over.

common/lib/dependabot/dependency_updaters/README.md
jurre: I like it, that also gives us an opportunity to pass in things like credentials for registries.

Even though we don't pass credentials in directly when running dependabot at GitHub (we run a proxy that adds auth details to requests), downstream users of the library often do depend on being able to pass the credentials into dependabot-core and having core make authenticated requests that way.

common/lib/dependabot/dependency_updaters/README.md
mattt: What do you think about taking a progressive enhancement approach? 

Plugins with a basic level of support return a "truthy" return value to indicate success and a "falsey" return value (`nil` or `false` (or raising an exception)) to indicate failure. The resulting PR may not have all of the information to populate a full description, but it works all the same.

Plugins that want to support additional functionality  integration could construct a response object with additional information to fill out the description with changelog entries, commits, and all the rest.

You're more familiar with that side of the codebase. Do you see any opportunities to take a tiered / relaxed / progressive approach like this to generating PRs?

@exvuma Thoughts about the product impact here? How important do you think it is for all PRs generated by Dependabot to have the same level of information for all ecosystems (trading off against the ease of supporting more ecosystems)?

common/lib/dependabot/dependency_updaters/README.md
mattt: I just sketched this out in 8a0a19b. Take a look and let me know what you think 😄

common/lib/dependabot/dependency_updaters/README.md
jurre: > Do you see any opportunities to take a tiered / relaxed / progressive approach like this to generating PRs?

I think it'll be pretty hard to integrate with the existing PR creation code if we don't have that information. Do you foresee many problems getting the before/after versions of a dependency? It seems like it should be somewhat doable to grab those, I'd assume the package manager would definitely have that information in order to be able to resolve the dependency tree.

For example, in the case of `npm` you could do:

```bash
$ npm ls etag --json
{
  "version": "1.0.0",
  "name": "name",
  "dependencies": {
    "etag": {
      "version": "1.8.1",
      "resolved": "https://registry.npmjs.org/etag/-/etag-1.8.1.tgz"
    }
  }
}
```

feelepxyz: > I think it'll be pretty hard to integrate with the existing PR creation code if we don't have that information.

👍 

I like the push to progressively enhancing the PR metadata, also good if sources are down but I also have alarm bells around the state of the codebase for PR creation. Branch creation and rebasing would likely break if we omitted the to version for example as we use this to create unique branch names per pm/package/version. For example: https://github.com/dependabot/dependabot-core/tree/dependabot/pip/python/helpers/hashin-0.17.0 checked here https://github.com/dependabot/dependabot-core/blob/dependabot/pip/python/helpers/hashin-0.17.0/common/lib/dependabot/pull_request_creator/github.rb#L43

common/lib/dependabot/dependency_updaters/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Public API
+
+Each `Dependabot::DependencyUpdaters` class inherits from `Base` 
+and implements the following methods:
+
+### `#update(dependency:, requirements:)`
+
+Updates a dependency according to the provided requirements.
+This method is called within the root directory of a project.
```
feelepxyz, line 16: We currently support setting the `directory` so running commands in some sub-project, do we want to support this?

jurre: Yes, I think we should, so maybe:

```suggestion
This method is called within the directory specified in `configuration.directory` or the root directory of a project if `configuration.directory` is `nil`.
```

Or alternatively, make the default `configuration.directory` be the root, and ensure it's never nil because nil sucks and then we can say:

```suggestion
This method is called within the directory specified in `configuration.directory`, which is the root directory of the project by default.
```

common/lib/dependabot/dependency_updaters/README.md
```diff
@@ -0,0 +1,72 @@
...
+| `configuration` | `Dependabot::Config::UpdateConfig` | The configuration options for performing updates |
+
+#### Return
+
+Returns `true` if the dependency was successfully updated 
+to satisfy the new requirements.
+
+#### Raise
+
+This method should raise an exception to communicate underlying errors
```
mctofu, line 33: Is there a case where `#update` would return false and not raise an error? In our current implementation the specific errors returned are pretty important and can dictate how we present the error to the user, whether we should attempt to retry later or could trigger alerts. We probably need to provide some kinda of guidance on what errors to raise so we don't need to rely on parsing error strings?

common/lib/dependabot/dependency_updaters/base.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Dependabot
+  module DependencyUpdaters
+    class Base
+      def update(dependency:, requirements:, configuration:)
+        raise NotImplementedError
```
mctofu, line 7: Would it be useful to create an adapter implementation that would allow taking our existing ecosystem updaters and fitting it into this interface? If we could quickly get something going end to end with a working update it could help us discover if there's anything additional we need to pass in or return to get a successful PR.

jurre: Yeah I think that's definitely something we should explore, it _seems_ pretty feasible to retrofit our existing implementations

PR #4549, opened 2021-12-20
Enabling parsing of complex workspace glob patterns (Azure client)

milind009:
Currently, dependabot only understand workspaces defined in `package.json` using simpler regexes like `packages/*`, `packages/folder/*`, `abcd`. However, some large monorepos define workspaces with regexes like `packages/**/*`,`*/packages/*` which with the current logic would not be supported and hence not included in dependency update process and the created PR.

Solution (**This currently only works for AzureDevOps repos**) - 
For repos having complex workspace regexes, using this ADO code search REST API: https://docs.microsoft.com/en-us/rest/api/azure/devops/search/code-search-results/fetch-code-search-results?view=azure-devops-rest-6.1 (works in a similar that we use the code search feature in AzureDevOps UI), we can get the references for “package.json” string in the entire repo in a given repo path. For every code path result obtained in this search, we can do these two checks – 
1)	Check if the path is for a “package.json” file and lies in the given repo source directory
2)	Check if the path satisfies any of the workspace regexes mentioned in the package.json

If both of the conditions are met, then we will include this path while updating any dependency and also fetch its `package.json` in the file fetcher phase.

**Drawback** - The Code search API by default works only on the repo default branch unless the code search feature is enabled for a particular branch in the repo policies in AzureDevOps

Therefore, in case of repo clients other than ADO and for branches on which code search feature is not enabled in case of ADO repos, we will fallback to the old logic of condensing the workspace regex to simpler form in case of complex glob pattern

For repos which don't have any complex workspace regexes and all workspace patterns are of the form packages/folder1/folder2.../* or simpler (i.e exact workspace directory is specified), we will still use the old logic present in dependabot

### DIFF
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -2,6 +2,7 @@
 
 require "dependabot/shared_helpers"
 require "excon"
+require "cgi"
 
 module Dependabot
   module Clients
@@ -19,6 +20,8 @@ module Dependabot
       class Forbidden < StandardError; end
 
       class TagsCreationForbidden < StandardError; end
+
+      class BadRequest < StandardError;
 
       RETRYABLE_ERRORS = [InternalServerError, BadGateway, ServiceNotAvailable].freeze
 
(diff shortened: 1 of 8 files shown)


jeffwidman:
So essentially this PR handles parsing complex workspace patterns by calling out to the Azure DevOps code search API... clever, but I don't think this is something we can move forward with. 

It's too invasive of a change into `dependabot-core` for a feature that is very specific to a single code hosting platform. Even though GitHub has a similar code search API, the implementation would probably be a bit different.

Stepping back, the whole idea of using a platform's code search API to retrieve specific files feels like it breaks our general abstraction of handling target repos as generic git repos that we run package managers against... I don't know how closely we follow that abstraction everywhere, but I'm very hesitant to move further away from it without a much larger conversation, especially because there are other ways to solve/optimize for this feature w/o using custom code search APIs.

If you really want to implement this feature via this API, I suggest running a fork of `dependabot-core` (which I think your team at Microsoft already does), and then submitting PR's against `dependabot-core` adding whatever hooks you need to make it easy to implement this functionality in your fork. We are _very_ open to PR's making the library more generic/extensible for others to build on top of.

@deivid-rodriguez if you can give this a second look I'd appreciate it. If you feel differently, happy to re-evaluate. Otherwise we should close this IMO.

deivid-rodriguez:
@jeffwidman I 100% agree, this is a too invasive change for us to include in core. And I'm not even sure it's the best solution to support this kind of manifest. If you want to use this, please fork or monkeypatch dependabot-core 👍.

But this is a real problem for all providers, right? We should probably have a ticket for it because I would be happy to include a solution that we do fine acceptable.

jeffwidman:
> But this is a real problem for all providers, right? We should probably have a ticket for it because I would be happy to include a solution that we do find acceptable.

I did what I should have done before and searched old issues. 😉 

Looks like this was already at least partially (and maybe completely) fixed by:
* https://github.com/dependabot/dependabot-core/pull/5231

We can probably leave it alone for now until someone who's actually affected files an issue illustrating a gap in current functionality (and proving it's valuable enough to someone that they took the time to file an issue)

deivid-rodriguez:
Interesting, I actually found that but I didn't think it was the same since it did not seem to address "**" globs. But maybe it did, since the PR is newer than this PR. @milind009 who knows, maybe your problem is already fixed?

PR #4554, opened 2021-12-21
Consistency pass on composer native helper build

brrygrdn:
As part of streamlining how we mount the native helpers into the development shell, I noticed that the composer native helper install path is inconsistent with our other native helpers.

In all other cases we follow the pattern of:

```Dockerfile
COPY --chown=dependabot:dependabot $ECOSYSTEM/helpers /opt/$ECOSYSTEM/helpers

RUN bash /opt/$ECOSYSTEM/helpers/build /opt/$ECOSYSTEM/
```

Which results in the relevant contents of the `helpers/` folder being copied up a level in the directory structure.

In the case of Bundler, this results in a directory with `helpers, v1, v2` folders.

### The problem

My expectation is that the composer helpers should be in the same paths as bundler, but in reality we are using them in-situ without copying any files or actually using the `$install_dir` argument in the script.

### Rationale

My understanding for not using the native helpers in situ is to avoid artefacts such as tests and development dependencies being pulled into the 'production' code we use in the real image.

Although composer works without this step, it being inconsistent is surprising and breaks any convention we could use to automate generation of the relevant paths.

### ⚠️ Caution

One thing of note is that this makes a change to the base image directory structure _and_ the `dependabot-composer` gem that much be updated in unison or it will be a breaking change.

I'm not sure how we've managed this kind of change in the past.

### DIFF
--- a/composer/helpers/v1/build
+++ b/composer/helpers/v1/build
@@ -9,7 +9,16 @@ if [ -z "$install_dir" ]; then
 fi
 
 helpers_dir="$(dirname "${BASH_SOURCE[0]}")"
-cd "$helpers_dir"
+cp -r \
+  "$helpers_dir/bin" \
+  "$helpers_dir/src" \
+  "$helpers_dir/.php_cs" \
+  "$helpers_dir/composer.json" \
+  "$helpers_dir/composer.lock" \
+  "$helpers_dir/phpstan.neon" \
+  "$install_dir"
+
+cd "$install_dir"
 
 composer1 validate --no-check-publish
 composer1 install
--- a/composer/lib/dependabot/composer/native_helpers.rb
+++ b/composer/lib/dependabot/composer/native_helpers.rb
@@ -8,7 +8,7 @@ module Dependabot
       end
 
       def self.composer_helpers_dir
-        File.join(native_helpers_root, "composer/helpers")
+        File.join(native_helpers_root, "composer")
       end
 
       def self.native_helpers_root
--- a/Dockerfile
+++ b/Dockerfile
@@ -240,13 +240,14 @@ USER dependabot
 RUN mkdir -p /opt/bundler/v1 /opt/bundler/v2
 RUN bash /opt/bundler/helpers/v1/build /opt/bundler/v1
 RUN bash /opt/bundler/helpers/v2/build /opt/bundler/v2
+RUN mkdir -p /opt/composer/v1 /opt/composer/v2
+RUN bash /opt/composer/helpers/v1/build /opt/composer/v1
+RUN bash /opt/composer/helpers/v2/build /opt/composer/v2
 RUN bash /opt/go_modules/helpers/build /opt/go_modules
 RUN bash /opt/hex/helpers/build /opt/hex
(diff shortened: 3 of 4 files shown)


composer/lib/dependabot/composer/native_helpers.rb
```diff
@@ -8,7 +8,7 @@ def self.composer_helper_path(composer_version: "v2")
       end
 
       def self.composer_helpers_dir
-        File.join(native_helpers_root, "composer/helpers")
+        File.join(native_helpers_root, "composer")
```
brrygrdn, line 11: Looking at this with fresh eyes, perhaps we should modify this to check if `composer/vX` folders exist and if not fall back to the original path to maintain compatibility between image and gem for a few releases?

composer/helpers/v1/build
```diff
@@ -9,7 +9,16 @@ if [ -z "$install_dir" ]; then
...
 
 helpers_dir="$(dirname "${BASH_SOURCE[0]}")"
-cd "$helpers_dir"
+cp -r \
+  "$helpers_dir/bin" \
+  "$helpers_dir/src" \
+  "$helpers_dir/.php_cs" \
+  "$helpers_dir/composer.json" \
+  "$helpers_dir/composer.lock" \
+  "$helpers_dir/phpstan.neon" \
```
jurre, line 18: This is linting configuration, I don't think we need it in the prod build?

brrygrdn: 👍🏻  Good point, will remove

composer/lib/dependabot/composer/native_helpers.rb
jurre: I'm mostly OK with ensuring we deploy this side by side, our process is set up to do so any way for this exact reason.

We can bump the minor version and mention it's a breaking change in the CHANGELOG?

composer/helpers/v1/build
brrygrdn: It turns out we do actually use this as part of the build script, we run the install and then 

```
composer run lint -- --dry-run
composer run stan
```

It's slightly unexpected as it means we'd experience a breakage at image build time rather than in a CI test, but I'm going to leave it as-is for now.

brrygrdn:
Since this change will require a lock-step upgrade of `gem dependabot-omnibus` and `docker pull dependabot/dependabot-core` I plan to:

1. Release v1.70.1 to ship any minor patches on `main`: https://github.com/dependabot/dependabot-core/pull/4615
2. Immediately merge this PR
3. Release v1.71.0 which includes _just_ this change
4. Merge https://github.com/dependabot/dependabot-core/pull/4556 and prepare a tracking PR for v1.72.0

The rationale for this is:
- v1.171.0 includes only the breaking change
- v1.171.x is reserved in the event we need to quickly withdraw the change **without** any other merges to main
- v.1.72.0 resumes business as usual patch updates

jurre (approved):

brrygrdn:
Looks like the python failures are legitimate on master, working on that now.

The failures in composer look legit and like the new composer paths aren't being picked up properly. My initial suspicion is that CI isn't ending up with the image in the right state, will investigate.

PR #4556, opened 2021-12-21
Refresh the workflow for native helpers within the docker dev shell, Consistency pass on native dev

brrygrdn:
Follow on from https://github.com/dependabot/dependabot-core/pull/4554, which should merge first

Our `docker-dev-shell` script currently uses volume mounting to ensure that changes made to core's code on the host machine are reflected in the container without rebuilding, and vice versa.

As part of this volume mounting strategy we are inconsistent on how we handle the 'native helpers' used by some languages.

These helpers are installed into `/opt/` in the base `dependabot/dependabot-core` image via `build` scripts which ensure they have any dependencies available in a path _outside_ the working directory.

For bundler and python, we currently volume mount the native helpers from the host machine into `/opt/` like so:

```bash
  -v "$(pwd)/bundler/helpers:/opt/bundler" \
  -v "$(pwd)/bundler/helpers:/opt/bundler/helpers" \
  -v "$(pwd)/python/helpers:/opt/python" \
  -v "$(pwd)/python/helpers:/opt/python/helpers" \
```

The intent of this is so to make it easier to debug/iterate on these helpers without rebuilding everything from `dependabot/dependabot-core` up -or- editing `/opt/` files in place before transcribing the changes back to the host machine to commit.

### The problem

The main issue with this approach is that by volume mounting the helpers directory from the host, **the host version of the directory must be properly built first**.

This creates a chicken and the egg problem, where before using the development container - to avoid installing the native dependencies for the relevant language on your host - you need to manually run some instructions from the relevant `build` script.

Secondly, we don't provide this volume mounting for all native helpers so contributors are left having to work out how to get native helpers in the right state and the documentation for each dependabot package has gone stale in some cases.

### Proposed changes

This PR aims to do two things:
- Avoid trampling the installed native helpers in `/opt/` with host versions that may not be in the correct state
- Provide a consistent workflow for _all_ native helpers to repeatably apply host changes without rebuilding everything

### The approach

Our current build scripts accept an argument for an `install_dir` that we use in the `Dockerfile` to place them in `/opt/` - but this argument is somewhat redundant as they will _always_ be installed in the same place, which is determined conventionally by the value of `DEPENDABOT_NATIVE_HELPERS` - which is set early in the `Dockerfile` itself.

I decided to remove the `install_dir` input from all scripts and just have it target this location as:
- Removing the need for a directory input means you can just re-run `$foo/helpers/build` to apply the latest changes to `/opt` without knowing anything about Dependabot's installation internals
- Making the entire `build` script repeatable ensures your local changes are correctly placed into `/opt/` by repeating the precise install commands ad verbatim, avoiding transcription errors or missed steps ( e.g. bundling the Ruby native helper gems into the wrong path )
- `DEPENDABOT_NATIVE_HELPERS` is already used by the Ruby code to determine where the native helpers should be on disk, so this means both the installer and calling code are aligned on the same convention

Finally, I added a statement to each script to `mkdir -p` their conventional directory as we were doing this for _some_ native helpers, but by making it consistent we ensure the convention for working on native helpers on your host machine follows the convention of placing an install in `$foo/helpers/install-dir` is consistent.

### Proposed new workflow

I've updated the README to:
1. Emphasise developing inside the container is the preferred approach
2. Describe pulling our pre-built developer container as a speed boost for this workflow
3. Describe how to make and apply native helper changes via the `$foo/helpers/build` script

### Out of scope

I deliberately left our `.devcontainer.json` alone as part of this pass. It duplicates some of the configuration we use for `bin/docker-dev-shell`, but it's use-case is a distinct contribution to support a VS Code automation that wraps our `bin/dry-run.rb` script.

I think there are improvements to be made here, such as using the prebuilt image, but I'd rather address those separately to solicit feedback from developers more familiar with this workflow as it was a community contribution.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -244,7 +244,7 @@ RUN mkdir -p /opt/composer/v1 /opt/composer/v2
 RUN bash /opt/composer/helpers/v1/build
 RUN bash /opt/composer/helpers/v2/build
 RUN bash /opt/go_modules/helpers/build
-RUN bash /opt/hex/helpers/build /opt/hex
+RUN bash /opt/hex/helpers/build
 RUN bash /opt/npm_and_yarn/helpers/build /opt/npm_and_yarn
 RUN bash /opt/python/helpers/build /opt/python
 RUN bash /opt/terraform/helpers/build /opt/terraform
--- a/Dockerfile
+++ b/Dockerfile
@@ -243,7 +243,7 @@ RUN bash /opt/bundler/helpers/v2/build
 RUN mkdir -p /opt/composer/v1 /opt/composer/v2
 RUN bash /opt/composer/helpers/v1/build
 RUN bash /opt/composer/helpers/v2/build
-RUN bash /opt/go_modules/helpers/build /opt/go_modules
+RUN bash /opt/go_modules/helpers/build
 RUN bash /opt/hex/helpers/build /opt/hex
 RUN bash /opt/npm_and_yarn/helpers/build /opt/npm_and_yarn
 RUN bash /opt/python/helpers/build /opt/python
--- a/Dockerfile
+++ b/Dockerfile
@@ -247,7 +247,7 @@ RUN bash /opt/go_modules/helpers/build
 RUN bash /opt/hex/helpers/build
 RUN bash /opt/npm_and_yarn/helpers/build
 RUN bash /opt/python/helpers/build
-RUN bash /opt/terraform/helpers/build /opt/terraform
+RUN bash /opt/terraform/helpers/build
 
 ENV HOME="/home/dependabot"
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -239,7 +239,6 @@ ENV DEPENDABOT_NATIVE_HELPERS_PATH="/opt" \
 USER dependabot
 RUN bash /opt/bundler/helpers/v1/build
 RUN bash /opt/bundler/helpers/v2/build
-RUN mkdir -p /opt/composer/v1 /opt/composer/v2
 RUN bash /opt/composer/helpers/v1/build
 RUN bash /opt/composer/helpers/v2/build
 RUN bash /opt/go_modules/helpers/build
--- a/Dockerfile
+++ b/Dockerfile
@@ -237,7 +237,6 @@ ENV DEPENDABOT_NATIVE_HELPERS_PATH="/opt" \
   MIX_HOME="/opt/hex/mix"
 
 USER dependabot
-RUN mkdir -p /opt/bundler/v1 /opt/bundler/v2
 RUN bash /opt/bundler/helpers/v1/build
 RUN bash /opt/bundler/helpers/v2/build
 RUN mkdir -p /opt/composer/v1 /opt/composer/v2
--- a/Dockerfile
+++ b/Dockerfile
@@ -238,8 +238,8 @@ ENV DEPENDABOT_NATIVE_HELPERS_PATH="/opt" \
 
 USER dependabot
 RUN mkdir -p /opt/bundler/v1 /opt/bundler/v2
-RUN bash /opt/bundler/helpers/v1/build /opt/bundler/v1
-RUN bash /opt/bundler/helpers/v2/build /opt/bundler/v2
+RUN bash /opt/bundler/helpers/v1/build
+RUN bash /opt/bundler/helpers/v2/build
 RUN mkdir -p /opt/composer/v1 /opt/composer/v2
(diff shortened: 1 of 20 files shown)


README.md
```diff
@@ -65,30 +62,73 @@ To run all of Dependabot Core, you'll need Ruby, Python, PHP, Elixir, Node, Go,
...
+Several Dependabot packages make use of 'native helpers', small executables in their host language.
+
+**Changes to these files are not automatically reflected inside the development container**
+
+Once you have made any edits to the helper files, run the appropriate build script to update the
+installed version with your changes like so:
+
+```bash
+bundler/helpers/v1/build
+bin/dry-run.rb bundler dependabot/demo --dir="/ruby"
```
brrygrdn, line 125:
```suggestion
$ bin/docker-dev-shell
=> running docker development shell
$ bundler/helpers/v1/build
$ bin/dry-run.rb bundler dependabot/demo --dir="/ruby"
```

README.md
```diff
@@ -65,30 +62,73 @@ To run all of Dependabot Core, you'll need Ruby, Python, PHP, Elixir, Node, Go,
...
 
-Run the tests by running `rspec spec` inside each of the packages. Style is
-enforced by RuboCop. To check for style violations, simply run `rubocop` in
-each of the packages.
+Run the tests by running `rspec spec` inside each of the packages, e.g.
 
-### Running with Docker
+```bash
+cd go_modules
+bundle exec rspec spec
```
brrygrdn, line 103:
```suggestion
$ cd go_modules
$ bundle exec rspec spec
```

README.md
```diff
@@ -65,30 +62,73 @@ To run all of Dependabot Core, you'll need Ruby, Python, PHP, Elixir, Node, Go,
...
 
-While you can run Dependabot Core without Docker, we also provide a development
-Dockerfile. In most cases, you'll be better off running Dependabot in the
-development Docker container as it bakes in all required dependencies.
+Style is enforced by RuboCop. To check for style violations, simply run `rubocop` in
+each of the packages, e.g.
+
+```bash
+cd go_modules
+bundle exec rubocop
```
brrygrdn, line 111:
```suggestion
$ cd go_modules
$ bundle exec rubocop
```

jurre (approved):
This looks great, thanks for tackling it 🎉

Dockerfile
```diff
@@ -240,17 +240,15 @@ ENV DEPENDABOT_NATIVE_HELPERS_PATH="/opt" \
...
-RUN bash /opt/bundler/helpers/v1/build /opt/bundler/v1
-RUN bash /opt/bundler/helpers/v2/build /opt/bundler/v2
-RUN bash /opt/go_modules/helpers/build /opt/go_modules
-RUN bash /opt/hex/helpers/build /opt/hex
-RUN bash /opt/npm_and_yarn/helpers/build /opt/npm_and_yarn
-RUN bash /opt/python/helpers/build /opt/python
-RUN bash /opt/terraform/helpers/build /opt/terraform
-RUN bash /opt/composer/helpers/v1/build /opt/composer/v1
-RUN bash /opt/composer/helpers/v2/build /opt/composer/v2
+RUN bash /opt/bundler/helpers/v1/build
```
lseppala, line 243: Could it be worthwhile to collapse these into a single `RUN` command and thus create a single FS layer? Admittedly they're not large layers, but may save time when downloading image & building.

Also, I don't think the `bash` is necessary because each is an executable file with a hash-bang?

jurre: I guess the downside to that would be that they're not cached individually, so changing a single helper requires reinstalling all of them right? Not sure what's the best trade-off in that case

brrygrdn: Yeah, I was kind of leaving this as-is, some of these installs do more than we'd ideally like - the build script is a bit of a black box in terms of core overall, it _could_ feasibly install some large dependencies ( for example, pulling in and vendoring an old binary for python2, or similar ).

I'd be inclined to keep these layers separate even if it is currently a little ceremonial, but I'd like to do an image optimisation pass more generally as well.

lseppala: > I guess the downside to that would be that they're not cached individually, so changing a single helper requires reinstalling all of them right?

True. With separate layers, this may also cause unneccessary reinstalling since layers are rebuilt if previous layers change. So kinda caught either way in building. The gain for combining would be less in building time and more in downloading time.

> I'd be inclined to keep these layers separate even if it is currently a little ceremonial, but I'd like to do an image optimisation pass more generally as well.

I'm fine saving this for later 👍

lseppala (approved):

PR #4558, opened 2021-12-22
Python: Quarantine slow tests into their own CI run

jurre:
These tests have not been wearing masks, and to avoid them infecting the
other tests we need to quarantine them.

Jokes aside, splitting these slow tests off into their own CI run should
reduce the overall time we have to wait on CI to pass, and it gives us
some indication on which tests we might want to try to speed up going
forward.

So far I've split off the ~10 slowest tests, which according to rspec
take 40% of the total time to run these tests!

We can move more things into the slow suite, but I'd like to avoid
having the majority of the tests live there, because then we'd still end
up with one very slow suite.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,7 @@ jobs:
         - { path: nuget, name: nuget }
         - { path: omnibus, name: omnibus }
         - { path: python, name: python }
+        - { path: python, name: python_slow }
         - { path: terraform, name: terraform }
     env:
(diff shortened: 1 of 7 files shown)


brrygrdn (approved):
I like this approach, I think it's a really pragmatic way to deal with the issue 👍🏻 

I also think this is a better approach than using a build-matrix balancing `rspec` extension as that would just result in us running an increasingly large number of workers in parallel which doesn't get to the root of the problem.

landongrindheim (approved):
🙌

PR #4559, opened 2021-12-22
Enable support for ActiveSupport 7

landongrindheim:
We encountered an issue recently where upgrading to ActiveSupport 7 caused our builds to fail. Upon investigation, I found that `byebug` was no longer a dependency of ActiveSupport. The library now depends on `debug`, which is a gem now but will ship with Ruby 3.1.

### DIFF
--- a/common/lib/dependabot/notifications.rb
+++ b/common/lib/dependabot/notifications.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "active_support"
 require "active_support/notifications"
 
 module Dependabot
--- a/common/lib/dependabot/shared_helpers.rb
+++ b/common/lib/dependabot/shared_helpers.rb
@@ -87,7 +87,7 @@ module Dependabot
       if ENV["DEBUG_FUNCTION"] == function
         puts helper_subprocess_bash_command(stdin_data: stdin_data, command: cmd, env: env)
         # Pause execution so we can run helpers inside the temporary directory
-        byebug # rubocop:disable Lint/Debugger
+        binding.break # rubocop:disable Lint/Debugger
       end
 
       env_cmd = [env, cmd].compact
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -68,7 +68,7 @@ Bundler.setup
 
 require "optparse"
 require "json"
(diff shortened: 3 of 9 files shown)


common/lib/dependabot/notifications.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "active_support"
```
landongrindheim, line 3: Per a [comment](https://github.com/rails/rails/issues/43889#issuecomment-995364395) on rails/rails, 

> Requiring individual files from Active Support without first requiring the top-level active_support is not supported.

This played out in the test suite for the previous commit, with errors of the following type:

```plaintext
      Failure/Error:
        Dependabot.instrument(
          Notifications::FILE_PARSER_PACKAGE_MANAGER_VERSION_PARSED,
          ecosystem: "bundler",
          package_managers: {
            "bundler" => version
          }
        )

      NameError:
        uninitialized constant ActiveSupport::IsolatedExecutionState
```

jurre (approved):
Thanks!

common/lib/dependabot/notifications.rb
jurre: Nice find!

landongrindheim:
For posterity's sake: The switch from `byebug` to `debug` wasn't necessary in every case here, but we otherwise would be maintaining two dependencies.

common/lib/dependabot/shared_helpers.rb
```diff
@@ -87,7 +87,7 @@ def self.run_helper_subprocess(command:, function:, args:, env: nil,
       if ENV["DEBUG_FUNCTION"] == function
         puts helper_subprocess_bash_command(stdin_data: stdin_data, command: cmd, env: env)
         # Pause execution so we can run helpers inside the temporary directory
-        byebug # rubocop:disable Lint/Debugger
+        binding.break # rubocop:disable Lint/Debugger
```
jurre, line 90: Could have used the generic `debugger` here which is aliased in most (all?) debuggers, but 🤷

landongrindheim: Today I learned!

PR #4567, opened 2021-12-27
Cleanup after byebug removal

landongrindheim:
### `.byebug_history` removals

These artifacts are remnants of a [recently removed gem](https://github.com/dependabot/dependabot-core/pull/4559). Now that we're
using `debug`, we no longer need to ignore `.byebug_history`.

### `#debugger` aliasing

I wasn't aware that the major Ruby debuggers implement `debugger` as an
alias to their `binding` calls in source code. While I don't see this as
a necessary or important change, I think it's worth making.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -16,6 +16,5 @@
 **/node_modules
 !**/spec/fixtures/*
 git.store
-.byebug_history
 .DS_Store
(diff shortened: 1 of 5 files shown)


lseppala (approved):
👍

PR #4569, opened 2021-12-28
Check for access before ghcr push

mctofu:
Our CI workflow currently fails on PRs opened from forks or Dependabot. Workflows for those PRs run with a read-only access token and no secrets so the push to ghcr fails.

Previously we avoided this failure by [checking](https://github.com/dependabot/dependabot-core/commit/c9314d37da7ca29c94fa63b28ed7cab9fa2e91f3#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL91) if the docker credentials were available to the workflow and skipping the push step if not. I've recreated that mechanism by adding a dummy secret, `ACCESS_CANARY`, that the workflow can check for before pushing. It's currently set to a random guid. The secret won't be available to fork PRs or Dependabot PRs so it should cause them to skip the push step.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,6 +64,9 @@ jobs:
             --cache-from "$CORE_BRANCH_IMAGE:$BRANCH_REF" \
             .
       - name: Push dependabot-core-branch image to GHCR
+        env:
+          ACCESS_CANARY: ${{ secrets.ACCESS_CANARY }}
+        if: env.ACCESS_CANARY != ''
         run: |
           docker push "$CORE_BRANCH_IMAGE:$BRANCH_REF"
       - name: Build dependabot-core-ci image
@@ -79,6 +82,9 @@ jobs:
             --cache-from "$CORE_CI_IMAGE:branch--$BRANCH_REF" \
             .
       - name: Push dependabot-core-ci image to GHCR
+        env:
+          ACCESS_CANARY: ${{ secrets.ACCESS_CANARY }}
+        if: env.ACCESS_CANARY != ''
         run: |
           docker push "$CORE_CI_IMAGE:latest"
           docker push "$CORE_CI_IMAGE:branch--$BRANCH_REF"


.github/workflows/ci.yml
```diff
@@ -64,6 +64,9 @@ jobs:
             --cache-from "$CORE_BRANCH_IMAGE:$BRANCH_REF" \
             .
       - name: Push dependabot-core-branch image to GHCR
+        env:
```
landongrindheim, line 67: This seems like a very sensible approach. In addition to protecting the GHCR, this should surface cases of misuse ✨

I've done a little googling, but I'd like to hear your thoughts on the name `ACCESS_CANARY`. My assumption is that it relates to pairing failed workflow runs to users. Is that correct?

landongrindheim (approved):
🚀

.github/workflows/ci.yml
mctofu: I'm not concerned about tracking the failures as that's currently the expectation for any PRs from external contributors or Dependabot. The GHCR push steps aren't required for those PRs so this is more about letting the rest of the workflow run so we can see if the tests are passing. Here's an example of how they are currently failing: https://github.com/dependabot/dependabot-core/runs/4644267200?check_suite_focus=true (sorry, should have linked that to start!)

I chose canary since it's an indirect indicator of whether the workflow will have the appropriate access to do the GHCR push. Sort of like the https://en.wiktionary.org/wiki/canary_in_a_coal_mine. I'd be open changing the name to something better but I'd like to merge this first to make sure it works.

landongrindheim: 🚢🇮🇹

mctofu: Confirmed this worked! CI is passing now in this fork PR and we can see the GHCR push steps were skipped: https://github.com/dependabot/dependabot-core/runs/4653634875?check_suite_focus=true

Another option would have been to try and detect forks/dependabot directly like:
`if: github.event.pull_request.head.repo.full_name == github.repository && github.actor != 'dependabot[bot]'` ([source](https://github.community/t/how-to-detect-a-pull-request-from-a-fork/18363/4))

I was worried there were more cases we'd need to detect and deal with so the secret check seemed like the simpler way to go. That alternative would make it more a little more clear which cases we expect to skip doing the GHCR push in.

PR #4584, opened 2022-01-05
Simplify Dockerfile a little bit

PeterDaveHello:
`USER root` was set already on above, just few lines before, and there's no need to use two `mkdir` together, just use it once, just like like Dockerfile.development: https://github.com/dependabot/dependabot-core/blob/84367a4ee86fe17bc22e7cfa1b0b4a1abbe9e156/Dockerfile.development#L56


landongrindheim (approved):
Thanks!

PR #4585, opened 2022-01-05
Bump cargo from 1.51.0 to 1.57.0

jurre:
https://blog.rust-lang.org/2021/12/02/Rust-1.57.0.html

PR #4588, opened 2022-01-05
Revert "Bump cargo from 1.51.0 to 1.57.0"

jurre:
Reverts dependabot/dependabot-core#4585


mattt (approved):

jurre:
It seems like the docker image is outdated in these PRs, the tests pass for me locally on `1.51.0` but fail on `1.57.0`. I will dig into that separately, but going to assume that merging this will fix the builds for now

PR #4598, opened 2022-01-06
Docker: Support ARG FROM

### LINKED ISSUE #4597: Docker: Support ARG FROM
**Package ecosystem**
docker
**Package manager version**
"Dockerfile"
**Language version**
docker
```docker
ARG HUB=docker.io
FROM ${HUB}/docker:20.10.7-dind
```
**dependabot.yml content**
```yml
version: 2
updates:
  - package-ecosystem: "docker"
    directory: "/"
    schedule:
      interval: "daily"
```
**Updated dependency**
```docker
ARG HUB=docker.io
FROM ${HUB}/docker:20.10.12-dind
```
Actually, this line does not support ARGs parsing including replacing: https://github.com/dependabot/dependabot-core/blob/b7ad648e8f97d273f6e74650644fa808d5731bb0/docker/lib/dependabot/docker/file_parser.rb#L42
Dependabot should put all `ARG`s in a map and replace the values before executing the regex.
This syntax is valid: https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

n/a
n/a, because Dependabot does not detect this line as valid `FROM` syntax, so no dependency was detected.

hfhbd:
Fixes #4597

### DIFF
--- a/docker/lib/dependabot/docker/file_parser.rb
+++ b/docker/lib/dependabot/docker/file_parser.rb
@@ -16,15 +16,16 @@ module Dependabot
       # Details of Docker regular expressions is at
       # https://github.com/docker/distribution/blob/master/reference/regexp.go
       DOMAIN_COMPONENT =
-        /(?:[[:alnum:]]|[[:alnum:]][[[:alnum:]]-]*[[:alnum:]])/.freeze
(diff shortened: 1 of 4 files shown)


jeffwidman:
Related: https://github.com/dependabot/dependabot-core/pull/1602

xlgmokha (approved):

PR #4617, opened 2022-01-10
Revert hashin upgrade for python native helper

brrygrdn:
Manually reverts https://github.com/dependabot/dependabot-core/pull/4610

CI failed on master and PRs subsequent to this merging, which is unexpected as the branch itself _did_ pass.

I built a base image of the previous [SHA on main](https://github.com/dependabot/dependabot-core/commit/4a33928a9c13fe665f24556c52f3a9033049f90e) locally and confirmed the failing python spec now passes:

```bash
$ bundle exec rspec spec/dependabot/python/file_updater/requirement_file_updater_spec.rb:177
Run options: include {:locations=>{"./spec/dependabot/python/file_updater/requirement_file_updater_spec.rb"=>[177]}}

Randomized with seed 4982
.

Finished in 2.41 seconds (files took 2.22 seconds to load)
1 example, 0 failures

Randomized with seed 4982
```

I want to see if CI agrees with me but there are two unknowns:
- why did the `hashin` bump break a ~3 year old test using a ~4 year old fixture, it doesn't seem like any out of band issue is involved ( e.g. remote fixture or real package changed in the wild )
- why did the branch CI not detect the problem?

PR #4618, opened 2022-01-10
[CI] Ensure CI builds pick up native helper changes

brrygrdn:
See: https://github.com/dependabot/dependabot-core/pull/4617

We've determined we have a green + green problem in CI where changes are made to native helpers are not fully tested on branches.

It turns out this was a result of my refactor of this file in https://github.com/dependabot/dependabot-core/pull/4524, where I removed `-t "$CORE_IMAGE:latest"` from the `docker build` that generated the 'branch image'.

Without this branch image tagged as `dependabot/dependabot-core:latest`, the `Dockerfile.ci` will pick up the base image pulled in, which is effectively `main`.

This results in us running Dependabot Core using the `native helpers` installed on `main` absent any changes made on branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,7 @@ jobs:
           DOCKER_BUILDKIT: 1
         run: |
           docker build \
+            -t "$CORE_IMAGE:latest" \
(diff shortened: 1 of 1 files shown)


jurre (approved):
This setup is so complex, I definitely also missed that this would break things.

PR #4626, opened 2022-01-12
Set two flaking tests to pending to clear up CI signals

brrygrdn:
I'm currently investigating these two tests, but they've been consistently red for the last two days and we are have accepted this risk as:
- Bundler 1 support is essentially on LTS at this point and we'd like to deprecate it in the mid-term
- They are known flakes which started to be consistently failing on an unrelated merge; i.e. they do not appear to be legitimate breakages.

I'm concerned that while we remain in the state of accepting `bundler1` being red for merges to main we risk missing other breakages, so I'd like to ensure CI is back to strongly signalling new regressions and reinstate these tests once they are confidently passing again.

### DIFF
--- a/bundler/spec/dependabot/bundler/file_updater_spec.rb
+++ b/bundler/spec/dependabot/bundler/file_updater_spec.rb
@@ -633,7 +633,13 @@ RSpec.describe Dependabot::Bundler::FileUpdater do
         context "that specifies the dependency using github:" do
           let(:dependency_files) { bundler_project_dependency_files("github_source") }
 
-          pending "doesn't update the git dependencies" do
+          # TODO: Reinstate test for bundler1
+          #
+          # This test flakes exclusively for bundler v1 builds, so we've excluded it form that build for now
+          # it should be reinstated once we've resolved the issue
+          it "doesn't update the git dependencies" do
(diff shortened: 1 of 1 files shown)


jurre (approved):
Yeah, good call, thanks for disabling these 👍

landongrindheim (approved):
Very reasonable. Thanks @brrygrdn 😄

bundler/spec/dependabot/bundler/file_updater_spec.rb
```diff
@@ -633,7 +633,13 @@
         context "that specifies the dependency using github:" do
           let(:dependency_files) { bundler_project_dependency_files("github_source") }
 
+          # TODO: Reinstate test for bundler1
+          #
+          # This test flakes exclusively for bundler v1 builds, so we've excluded it form that build for now
```
jeffwidman, line 638: typo: `form` -> `from`

xlgmokha:
```suggestion
          # This test flakes exclusively for bundler v1 builds, so we've excluded it from that build for now
```

deivid-rodriguez:
Flakies were fixed, so this can be closed now 🎉.

PR #4627, opened 2022-01-12
Attempt to run tests in parallel

jurre:
Will it work? No idea but let's find out!


jurre:
The python suite, which is our slowest, still seems to take just as long when running in parallel, so even though some suites see a 2x improvement, I don't think it's worth trying to resolve the concurrency issues in the failing tests, as it won't improve our overall build times :(

Issue #2999, opened 2021-01-17
Drupal project dependabot runs are very slow


driskell:
**Package manager/ecosystem**
Composer v2

**Manifest contents prior to update**
```
{
    "name": "drupal/recommended-project",
    "description": "Project template for Drupal 8 projects with a relocated document root",
    "type": "project",
    "license": "GPL-2.0-or-later",
    "homepage": "https://www.drupal.org/project/drupal",
    "support": {
        "docs": "https://www.drupal.org/docs/user_guide/en/index.html",
        "chat": "https://www.drupal.org/node/314178"
    },
    "repositories": [
        {
            "type": "composer",
            "url": "https://packages.drupal.org/8"
        }
    ],
    "require": {
        "composer/installers": "^1.2",
        "drupal/block_permissions": "1.0.0",
        "drupal/core-composer-scaffold": "^8.8",
        "drupal/core-project-message": "^8.8",
        "drupal/core-recommended": "^8.8"
    },
    "require-dev": {
        "drupal/core-dev": "^8.8"
    },
    "conflict": {
        "drupal/drupal": "*"
    },
    "minimum-stability": "dev",
    "prefer-stable": true,
    "config": {
        "sort-packages": true
    },
    "extra": {
        "drupal-scaffold": {
            "locations": {
                "web-root": "web/"
            }
        },
        "installer-paths": {
            "web/core": [
                "type:drupal-core"
            ],
            "web/libraries/{$name}": [
                "type:drupal-library"
            ],
            "web/modules/contrib/{$name}": [
                "type:drupal-module"
            ],
            "web/profiles/contrib/{$name}": [
                "type:drupal-profile"
            ],
            "web/themes/contrib/{$name}": [
                "type:drupal-theme"
            ],
            "drush/Commands/contrib/{$name}": [
                "type:drupal-drush"
            ],
            "web/modules/custom/{$name}": [
                "type:drupal-custom-module"
            ],
            "web/themes/custom/{$name}": [
                "type:drupal-custom-theme"
            ]
        },
        "drupal-core-project-message": {
            "include-keys": [
                "homepage",
                "support"
            ],
            "post-create-project-cmd-message": [
                "<bg=blue;fg=white>                                                         </>",
                "<bg=blue;fg=white>  Congratulations, you’ve installed the Drupal codebase  </>",
                "<bg=blue;fg=white>  from the drupal/recommended-project template!          </>",
                "<bg=blue;fg=white>                                                         </>",
                "",
[...]

PR #4635, opened 2022-01-15
Prevent unnecessary composer package downloads

driskell:
This PR prevents all installations of packages during composer updates. It significantly increases speed of updates.

This was discussed in #2999 and should help greatly with #2416.

Previous PR was #2998 and I closed it as I thought there was issues with this approach, as https://github.com/wikimedia/composer-merge-plugin stopped working. After rechecking with existing dependabot-core main branch I did see setScripts is false already and was able to reproduce the merge-plugin issue there. Having thought about it I think it's unreasonable to support that plugin in dependabot - even if the packages merged in weren't removed there would still be incompatibility in that packages merged in won't get dependabot updates. So I've produced this PR again as it was for discussion.

Credit should also go to @jurre here as it's a progression from #3009.


feelepxyz (commented):
Amazing work on this! 🎉  I think this looks good, will give it a quick QA test locally to verify.

Thanks for updating the tests, not raising for these cases is a real nice improvement.

jurre:
This seems very promising, thanks @driskell 🎉 

@feelepxyz do you need a hand w/ QAing?

feelepxyz (approved):
Tested this out a few test repos and working nicely! Double win here, speeding up composer v2 updates and removing a class of update errors where downloads needed authentication 🎉  

This is in line with other package managers so great to see this working in composer now.

PR #4646, opened 2022-01-19
Upgrade Python version to 3.10.2

ulgens:


Dockerfile
```diff
@@ -83,14 +83,17 @@ RUN apt-add-repository ppa:brightbox/ruby-ng \
...
   PATH="/usr/local/.pyenv/bin:$PATH"
 RUN mkdir -p "$PYENV_ROOT" && chown dependabot:dependabot "$PYENV_ROOT"
 USER dependabot
-RUN git clone https://github.com/pyenv/pyenv.git --branch v2.2.2 --single-branch --depth=1 /usr/local/.pyenv \
-  && pyenv install 3.10.0 \
-  && pyenv global 3.10.0 \
+# FIXME: Use tagged release when there is release with Python 3.10.2 support
+# RUN git clone https://github.com/pyenv/pyenv.git --branch v2.2.3 --single-branch --depth=1 /usr/local/.pyenv \
+RUN git clone https://github.com/pyenv/pyenv.git /usr/local/.pyenv \
+  && cd /usr/local/.pyenv && git checkout a55dd92436863cae0e2af1c1e051ee5f8853c005 && cd - \
```
landongrindheim, line 94: As noted in your comment, I think it's best if we use a tagged release of `pyenv`. Rather than checking out an unreleased commit, let's hold off until `v2.2.4` is released.

landongrindheim (changes requested):
Thanks for the contribution! 🙇

This looks like a solid approach, and I appreciate your updating comments 😄 I'm requesting that we update the `pyenv` version in this PR to `v2.2.4`, which has yet to be released. Hopefully we don't have to wait long 🤞

Dockerfile
ulgens: For previous pyenv releases, it took some time to create a new release with the latest Python. I wouldn't prefer to wait them if that's up to me, but I created a issue on their side to hopefully speed up the process https://github.com/pyenv/pyenv/issues/2224

ulgens: Pyenv released 2.2.4 and i updated the PR.

landongrindheim (approved):
Thanks for following up with this 😄

PR #4661, opened 2022-01-25
Update GitHub Actions label description

xlgmokha:
This change updates the label description for the `github_actions` label

from:

> Pull requests that update Github_actions code

to:

> Pull requests that update GitHub action code


common/lib/dependabot/pull_request_creator/labeler.rb
```diff
@@ -351,23 +351,24 @@ def create_gitlab_security_label
...
         github_client_for_source.add_label(
           source.repo,
-          langauge_name,
-          self.class.label_details_for_package_manager(package_manager).
-            fetch(:colour),
-          description: "Pull requests that update #{langauge_name.capitalize} "\
-                       "code",
+          language_name,
+          label.fetch(:colour),
+          description: label.fetch(:description) { default_description_for(language_name) },
```
xlgmokha, line 360: I chose to introduce an optional key `:description` key in the Hash to allow overriding the default description that is generated.

github_actions/lib/dependabot/github_actions.rb
```diff
@@ -15,6 +15,7 @@
   register_label_details(
     "github_actions",
     name: "github_actions",
+    description: "Pull requests that update GitHub action code",
     colour: "000000"
   )
```
xlgmokha, line 20: I wanted to test this registration code directly but it looks like we do some work to skip or remove this registration in the tests in favour of https://github.com/dependabot/dependabot-core/blob/44168cffbace107ee1021713fb71eed53ebc6e56/common/spec/dummy_package_manager/dummy.rb#L9. Because of this I had to stub out this registration in the tests added below.

github_actions/spec/dependabot/pull_request_creator/labeler_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+  subject(:labeler) do
+    described_class.new(
+      source: source,
+      credentials: [],
+      custom_labels: nil,
+      includes_security_fixes: false,
+      dependencies: [dependency],
+      label_language: true,
+      automerge_candidate: false
+    )
```
xlgmokha, line 15: I would love to add sane default values for most of these attributes because the constructor requires a lot of parameters and they seem to distract from the test and make it difficult to use this class for very simple setups.

github_actions/spec/dependabot/pull_request_creator/labeler_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+          })
+
+        stub_request(:get, "https://api.github.com/repos/#{source.repo}/labels?per_page=100").
+          to_return(status: 200, headers: { "Content-Type" => "application/json" }, body: JSON.generate([]))
+        stub_request(:post, "https://api.github.com/repos/#{source.repo}/labels").
+          to_return(
+            status: 201,
+            headers: { "Content-Type" => "application/json" },
+            body: JSON.generate({ id: 1, name: "github_actions", color: "000000" })
+          )
```
xlgmokha, line 41: Duplication was intended because it makes the test easier to read IMO. It also cuts down on the need to jump up and down in the file to understand the moving parts. I hope that's okay.

github_actions/lib/dependabot/github_actions.rb
```diff
@@ -15,6 +15,7 @@
   register_label_details(
     "github_actions",
     name: "github_actions",
+    description: "Pull requests that update GitHub action code",
```
mctofu, line 18:
```suggestion
    description: "Pull requests that update GitHub Actions code",
```

Should we use "GitHub Actions" instead? That seems to be the official way they are referred to: https://docs.github.com/en/actions

github_actions/spec/dependabot/pull_request_creator/labeler_spec.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+RSpec.describe Dependabot::PullRequestCreator::Labeler do
```
mctofu, line 5: There is a test for this class in `common/spec/dependabot/pull_request_creator/labeler_spec.rb`. Maybe these tests should go there?

mctofu (commented):
This approach makes sense! Just had a few questions.

github_actions/spec/dependabot/pull_request_creator/labeler_spec.rb
xlgmokha: I have moved the test code in https://github.com/dependabot/dependabot-core/pull/4661/commits/a9de16bf24e1550c889ffe61a5a1070cfc1a8e3f.

common/spec/dependabot/pull_request_creator/labeler_spec.rb
```diff
@@ -228,6 +228,39 @@
...
+          end
+
+          it "creates a label" do
+            labeler.create_default_labels_if_required
+
+            expect(WebMock).to have_requested(:post, "https://api.github.com/repos/#{source.repo}/labels").
+              with(body: {
+                name: "github_actions",
+                color: "000000",
+                description: "Pull requests that update GitHub Actions code"
```
exvuma, line 260: Looks good to me. Totally 💯 on copy as is 

```suggestion
                description: "Pull requests that update GitHub Actions (e.g code in .github/workflows folder"
```


nit/idea  something about the word "code" when talking about actions I feel like could be confusing. very stretch goal, but could we be more descriptive and include workflows, or .github files to be specific about what's happening?

mctofu (approved):

PR #4664, opened 2022-01-25
npm7: Fix subdependency versoion resolver 

feelepxyz:
Fixes the npm 7 subdependency version resolver by doing an update using
the npm cli, similar to how we do subdependency updates in the lockfile
updater.

Dependabot currently says no update is possible for npm 7 subdependency
updates.

I ran into a few issues testing npm 8 so leaving this out of this PR.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/subdependency_version_resolver.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/subdependency_version_resolver.rb
@@ -114,11 +114,16 @@ module Dependabot
             Dir.chdir(path) do
               npm_version = Dependabot::NpmAndYarn::Helpers.npm_version(lockfile_content)
 
-              SharedHelpers.run_helper_subprocess(
-                command: NativeHelpers.helper_path,
-                function: "#{npm_version}:updateSubdependency",
-                args: [Dir.pwd, lockfile_name, [dependency.to_h]]
-              )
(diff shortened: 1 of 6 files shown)


npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/subdependency_version_resolver.rb
```diff
@@ -114,11 +114,32 @@ def run_npm_updater(path, lockfile_name, lockfile_content)
...
+                command = [
+                  "npm",
+                  "update",
+                  dependency.name,
+                  "--force",
+                  "--dry-run",
+                  "false",
+                  "--ignore-scripts",
+                  "--package-lock-only"
+                ].join(" ")
```
jurre, line 133: Any point in DRYing this with the update checker?

jurre (approved):
This makes a lot of sense to me, thanks for tackling!

Should we get rid of the unused npm8 fixtures?

feelepxyz:
> Should we get rid of the unused npm8 fixtures?

👍

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/subdependency_version_resolver.rb
jurre: I like it 👍

jurre (approved):

PR #4692, opened 2022-02-01
Add support for Python 3.9.10

Dresdn:
Adds python 3.9.10 to the python supported versions matrix


Nishnha:
Hi @Dresdn,

Thanks for adding Python v3.9.10 to our supported versions matrix, we must have skipped over it!

It looks like the release is [mostly bug fixes](https://www.python.org/downloads/release/python-3910/) and we already support Python v3.9.9, so I think this will be okay to merge once CI passes.

Nishnha (approved):

PR #4693, opened 2022-02-01
Fix peer dependency regex for cases with 2 semver constraints

FaHeymann:
At the moment the case
```
npm ERR! Could not resolve dependency:
npm ERR! peer opentelemetry/api@">=1.0.0 <1.1.0" from @opentelemetry/context-async-hooks@1.0.1
``` seems not to be recognized as a peer dependency conflict because the semver constraint `>=1.0.0 <1.1.0` contains a whitespace and then does not match the regex.


FaHeymann:
Hi, thanks for looking at this!

It took a while for me to get the tests running locally and it seems like the case I wrote does not work yet anyways. I'll try to improve it.

FaHeymann:
@Nishnha could we maybe run CI for this branch so that I can be sure that it passes there too? Thanks!

FaHeymann:
Hello again :)

What are the next steps? Is there something I should do to get this reviewed / merged?
Thanks!

xlgmokha (approved):

PR #4702, opened 2022-02-05
Remove rust toolchains to save space in the docker image

mctofu:
The rust toolchains take around 2GB of space.

```
[dependabot-core-dev] ~/dependabot-core $ du -sk /opt/rust/toolchains/*
1067576 /opt/rust/toolchains/1.58.0-x86_64-unknown-linux-gnu
1067648 /opt/rust/toolchains/stable-x86_64-unknown-linux-gnu
```

Removing them slims down the image from 4.3GB to 2.3GB!

This is an experiment to see if we can still do cargo updates.


mctofu:
This worked but I could see that the cargo tests were taking longer to compete than normal. Taking a closer look I found that the necessary toolchain would be downloaded on demand so we are incurring a performance hit by not including a toolchain in the image. I took another stab at this in https://github.com/dependabot/dependabot-core/pull/4711 which installs a single minimal toolchain rather the the two standard toolchains we currently install.

PR #4716, opened 2022-02-08
Verify the right version of composer is installed in helper and docker

jurre:
Composer versions in the native helpers and dockerfile need to match up. Currently they don't, but there's also no way for us to know right now, so let's add a simple test for it to remind ourselves


brrygrdn (approved):
👍🏻  On this approach

Nishnha (approved):

PR #4729, opened 2022-02-10
Use correct file paths in Bitbucket commits

rimas-kudelis:
If a dependency file is not in the repository root, it's currently being committed as a new file in the root directory when creating a Bitbucket commit.
This change should ensure that it will be committed in the right place.

I removed the `starts_with("/")` check, assuming it's no longer necessary and seeing how dumped file path starts with the `/` character. I haven't tested all possible scenarios (or indeed any) though, Hopefully the CI will make sure that everything works as it should.

### DIFF
--- a/common/lib/dependabot/clients/bitbucket.rb
+++ b/common/lib/dependabot/clients/bitbucket.rb
@@ -106,8 +106,7 @@ module Dependabot
         }
 
         files.each do |file|
-          absolute_path = file.name.start_with?("/") ? file.name : "/" + file.name
-          parameters[absolute_path] = file.content
+          parameters[file.path] = file.content
         end
 
         body = encode_form_parameters(parameters)


Nishnha:
> If a dependency file is not in the repository root, it's currently being committed as a new file in the root directory when creating a Bitbucket commit

This is definitely not the behavior we intended!

I'll run the CI here, but I'm not sure what our test coverage around file paths is; we _may_ need to add testing around this scenario.

jurre:
Thanks @rimas-kudelis, could you add a test that covers this?

rimas-kudelis:
> Thanks @rimas-kudelis, could you add a test that covers this?

No I can't, sorry.

jurre:
> No I can't, sorry.

Is there anything I can do to help you write the tests? Not really comfortable merging this in without knowing that it'll work or won't break in the future.

rimas-kudelis:
Hi, 
the thing is twofold:
1. I'm not familiar with the Ruby or the testing frameworks you're using. Hacking up a simple bugfix is one thing, but setting up everything for the tests to work and learning how to add them is a totally different task to undertake.
2. I actually switched to Renovate. There were multiple reasons why I did so, and I doubt I can remember them all now (or indeed any), but at least one was even being communicated by your team as something not currently supported by Dependabot, but supported by Renovate.

I'd be glad if someone else could take care of the test.

landongrindheim:
Since this isn't actionable as is, and we don't have great insight into BitBucket, I'm going to close this. I _do_ think there's value in this PR existing, even if it's just inspiration for a future (tested) fix.

If someone with BitBucket familiarity wants to run with this and add some tests to prevent regressions in the future, please feel re-open and push it forward 😄

PR #4736, opened 2022-02-14
Add `--dry-run` flag to `bump-version` script

jurre:
I often find myself not really knowing if I should be doing a patch or
minor bump for dependabot-core, and what I end up doing is just running
the update with either, looking at the changelog, discarding the changes
and running it again with the right flag.

This adds a `--dry-run` option to the script instead, that allows you to
peek at what would be generated, which can help inform if we should be
doing a patch or minor release.


brrygrdn (approved):
Good call 👍🏻

PR #4738, opened 2022-02-14
Support depname.version = "1.2.3" syntax in TOML

roblabla:
With cargo, it's possible to define a dependency using this syntax:

```
sentry.version = "1.2.3"
sentry.features = ["a", "b", "c"]
```

This PR adds support for this in dependabot-core's manifest updater.

### DIFF
--- a/cargo/lib/dependabot/cargo/file_updater/manifest_updater.rb
+++ b/cargo/lib/dependabot/cargo/file_updater/manifest_updater.rb
@@ -139,7 +139,7 @@ module Dependabot
         end
 
         def declaration_regex(dep)
-          /(?:^|^\s*|["'])#{Regexp.escape(dep.name)}["']?\s*=.*$/i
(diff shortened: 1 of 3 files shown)


jurre:
Thanks! That makes sense to me, would you mind adding a testcase for it? Lmk if you need guidance on where to add it etc

roblabla:
I added a test case. It fails before the fix, and passes after.

jeffwidman (commented):
Should the test fixture file name be `doted_key_version` or `dotted_key_version`?

I would expect the latter, but I've only played with Cargo a little, so maybe I'm just missing something... ?

roblabla:
Ah, damn, you're right it's dotted. I copy pasted the typo :sweat_smile: . Fixed.

jurre:
One thing I just quickly want to verify manually is if the FileParser also knows how to deal with this format (I'd expect it does from glancing over the code), so setting up a test repo and running an update against it before merging this in

jurre (approved):

PR #4744, opened 2022-02-15
Docker: ARG Support: Only replace args outside build stage

hfhbd:
First PR of fixing Docker ARG support...

https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact
`
An ARG declared before a FROM is outside of a build stage, so it can’t be used in any instruction after a FROM. 
To use the default value of an ARG declared before the first FROM use an ARG instruction without a value inside of a build stage:
`


mctofu (approved):
Thanks! I'm merging this to https://github.com/dependabot/dependabot-core/pull/4837

PR #4755, opened 2022-02-21
github_actions: Add support for Composite Actions

### LINKED ISSUE #4178: Check and update dependencies in GitHub Actions action.yml
When using `composite` actions steps are defined in `action.yml` and we can have dependency to another actions.

Currently dependencies aren't checked in `action.yml`

JonasAlfredsson:
Composite Actions follow a very similar structure to the "normal"
GitHub Actions, with the exception that they must be located in a
file called action.yml (or .yaml) in the root of the repository.

Because of this similarity the file_parser and the file_updater does
not need any tweaking, and it is only the file_fetcher that needs to
be able to search beyond the .github/workflows folder.

Since GitHub only looks for a single file in the root directory of
the repository we can limit the expansion to the search to the same
strict parameters so we don't accidentally find a lot of other stuff.

Resolves #4178


JonasAlfredsson:
Hi,

This is my first time coding in Ruby, so I will gratefully accept any feedback to my style of coding in case it does not adhere to best practices in this language.

I also tested this against my composite action repository, and it appears to work:

```
bin/dry-run.rb github_actions JonasAlfredsson/checkout-qemu-buildx
```

jurre:
Thanks for tackling this! I'm "out of office" today but looking forward to taking a peek tomorrow! Overall it looks good, let's see what CI comes back with

github_actions/lib/dependabot/github_actions/file_fetcher.rb
```diff
@@ -7,11 +7,11 @@ module Dependabot
   module GithubActions
     class FileFetcher < Dependabot::FileFetchers::Base
       def self.required_files_in?(filenames)
-        filenames.any? { |f| f == ".github" }
+        filenames.any? { |f| f.match?("^(\.github|action.ya?ml)$") }
```
jurre, line 10: Little bit of a nit, but:

I was convinced that `match` takes a regular expression as an argument, but it turns out that strings are converted to regular expressions. However, I still feel like we could save a few computation cycles and improve clarity a little bit by passing a regex:

```suggestion
        filenames.any? { |f| f.match?(/^(\.github|action.ya?ml)$/) }
```

And while we're at it, could we extract this into a constant? Something like `FILENAME_PATTERN`, it typically goes at the top of the file in ruby so:

```ruby
  module GithubActions
    class FileFetcher < Dependabot::FileFetchers::
      FILENAME_PATTERN = /^(\.github|action.ya?ml)$/.freeze

      # ...
      filenames.any? { |f| f.match?(FILENAME_PATTERN) }
```

github_actions/spec/dependabot/github_actions/file_fetcher_spec.rb
```diff
@@ -31,13 +31,20 @@
...
-  context "with a workflow file" do
+  context "with workflow files" do
     before do
+      stub_request(:get, url + "?ref=sha").
+        with(headers: { "Authorization" => "token token" }).
+        to_return(
+          status: 200,
+          body: fixture("github", "contents_githubaction_repo_base_basic.json"),
+          headers: { "content-type" => "application/json" }
+        )
```
jurre, line 42: Why is this change important for adding composite actions?

github_actions/lib/dependabot/github_actions/file_fetcher.rb
```diff
@@ -40,7 +40,8 @@ def workflow_files
         @workflow_files ||=
           repo_contents(dir: ".github/workflows", raise_errors: false).
           select { |f| f.type == "file" && f.name.match?(/\.ya?ml$/) }.
-          map { |f| fetch_file_from_host(".github/workflows/#{f.name}") }
+          map { |f| fetch_file_from_host(".github/workflows/#{f.name}") } \
+          + [fetch_file_if_present("action.yml"), fetch_file_if_present("action.yaml")].compact
```
jurre, line 44: Why do we use `fetch_file_from_host` for the workflow files but `fetch_file_if_present` for the action.yaml? I wonder if we should filter out the action.yaml from `repo_contents(dir: "/", raise_errors: false)` and then `fetch_file_from_host`, or if that would just needlessly complicate things. Not sure that I'm getting at a suggestion but the code stood out to me a little so maybe you can share a bit of context that I'm missing.

JonasAlfredsson: No you are on the right track, I had the same thought initially. 
But in the top case we know that there exists a folder named `.github/workflows", and we should iterate through all the files in that one to find all `.ya?ml` files so there is a good reason to go through the entire folder.

In this new case we are only interested in looking at one file (however, with two possible file endings) so I thought it would be unnecessary to iterate through the entire root directory to find it so I looked in the other file_fetcher examples and found these functions that were used when we were only interested in one file. 

I have no issue with changing it to keep it similar to the first one, but it will be quite verbose compared to this one.

github_actions/spec/dependabot/github_actions/file_fetcher_spec.rb
JonasAlfredsson: The code is now looking for files in the root of the directory, and an error is returned by the mocking library in case a "request" is made to this URL and nothing is there to respond. 
Using an actual example of how a root of a repository would look like was just bonus.

github_actions/lib/dependabot/github_actions/file_fetcher.rb
JonasAlfredsson: I took this from how it looked in other file_fetchers, but I can update this according to the suggestion :)

JonasAlfredsson:
Made an `--amend` and force push of the suggested change, so I guess you have to approve the workflow again :)

github_actions/lib/dependabot/github_actions/file_fetcher.rb
jurre: Sounds reasonable to me, thanks for elaborating 👍

github_actions/spec/dependabot/github_actions/file_fetcher_spec.rb
jurre: Perfect, thank you!

github_actions/lib/dependabot/github_actions/file_fetcher.rb
jurre: Ah, interesting that the other fetchers are done this way, very understandable you carried over that style that's probably what everyone implementing these has done 😄. I'll have a look at those separately because I do think this is slightly preferable and thanks for updating it 🙇

jurre:
Gonna set some time aside today or tomorrow to manually test this on a few projects before merging it, but things look good, cheers!

JonasAlfredsson:
The failed check reference something about an old version of ruby. Is it something I need to be concerned about or is "my" test green?

jurre:
> The failed check reference something about an old version of ruby. Is it something I need to be concerned about or is "my" test green?

It's an unfortunate flaky test that you won't need to worry about

jurre:
Apologies for the delay on this, but looks good on my end, have rebased things and will merge once CI passes

JonasAlfredsson:
Sounds good!

landongrindheim (approved):
[Transferring approval](https://github.com/dependabot/dependabot-core/pull/4755#issuecomment-1061655182)

:shipit:

PR #4756, opened 2022-02-21
Fix references to changelog path/contents in bump_version.rb

brrygrdn:
Fixes a small bug in the refactored script


jurre (approved):
Thanks!

PR #4763, opened 2022-02-24
Update to npm 8

jurre:
This updates npm to version 8, which is mostly backwards compatible with v7, but drops support for an old version of node.

We've left the references to `npm7` in the codebase (test fixtures and npm version that we reference) as is, because it would become a fairly noisy change and functionally it doesn't change anything. We do plan to follow this up with a PR to update all of those references to `npm8`.

~There is currently one issue that we're digging into which is that the formatting of lockfiles is not respected between updates, and this seems to be caused by a change in npm where in version 8 it bases the formatting (indentation) on the `package.json` instead. When Dependabot performs updates to the `package-lock.json` we strip formatting, so in our case this always results in a large diff in the lockfile, however, I don't think we should assume consistent formatting between the package.json and lockfile either, so ideally this is something that can be resolved in npm.~

There is one functional change here:

On npm7 (due to a bug in npm), the lockfile would keep it's original formatting (indentation depth, tabs vs spaces) between updates, even if it was different from the package.json. In npm 8, the `package-lock.json` will always use the JSON formatting from the `package.json`.

Because of how dependabot updates lockfiles for npm, we modify the JSON in ruby and then spit it back out in a temporary folder. This causes us to lose any formatting, so I've added some code that attempts to figure out how the file was formatted, but the approach is fairly naive, so if someone mixes tabs and spaces, we assume it's tabs, if they mix indentation depth, we use the shallowest indentation. npm will default to indenting with 2 spaces in cases where the is no indentation in the package.json, which I think is preferable in most cases.

I don't expect this to be a huge issue, but it might trip up a few users that rely on this behavior and still use npm 7. We should advise them to update their version of npm because v7 is not officially supported anymore.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
@@ -476,6 +476,7 @@ module Dependabot
             updated_content = lock_deps_with_latest_reqs(updated_content)
 
             updated_content = sanitized_package_json_content(updated_content)
+
             File.write(file.name, updated_content)
           end
         end
@@ -495,6 +496,16 @@ module Dependabot
           end
         end
 
+        # Takes a JSON string and detects if it is spaces or tabs and how many
+        # levels deep it is indented.
+        def detect_indentation(json)
+          indentation = json.scan(/^\s+/).min_by(&:length)
+          indentation_size = indentation.length
+          indentation_type = indentation.scan(/\t/).any? ? "\t" : " "
+
+          indentation_type * indentation_size
+        end
+
         def lock_git_deps(content)
           return content if git_dependencies_to_lock.empty?
 
@@ -508,7 +519,8 @@ module Dependabot
             end
           end
 
-          json.to_json
+          indent = detect_indentation(content)
+          JSON.pretty_generate(json, indent: indent)
         end
 
         def git_dependencies_to_lock
@@ -549,7 +561,8 @@ module Dependabot
             end
           end
 
-          json.to_json
+          indent = detect_indentation(content)
+          JSON.pretty_generate(json, indent: indent)
         end
 
         def replace_ssh_sources(content)
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
@@ -56,8 +56,8 @@ module Dependabot
         # npm ERR! peer @opentelemetry/api@">=1.0.0 <1.1.0" from @opentelemetry/context-async-hooks@1.0.1
         NPM7_PEER_DEP_ERROR_REGEX =
           /
-            npm\sERR!\sCould\snot\sresolve\sdependency:\n
-            npm\sERR!\speer\s(?<required_dep>\S+@\S+(\s\S+)?)\sfrom\s(?<requiring_dep>\S+@\S+)
+            npm\s(?:WARN|ERR!)\sCould\snot\sresolve\sdependency:\n
+            npm\s(?:WARN|ERR!)\speer\s(?<required_dep>\S+@\S+(\s\S+)?)\sfrom\s(?<requiring_dep>\S+@\S+)
           /x.freeze
 
         def initialize(dependency:, credentials:, dependency_files:,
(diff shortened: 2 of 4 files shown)


Nishnha (approved):
This is sweet 🍬 
Maybe we should do a `minor` core release for this by itself?
We can suggest users update their version of npm in the release notes.

jurre:
> We can suggest users update their version of npm in the release notes.

Things _should_ work for npm 7, but let's explicitly mention that we're running npm 7 updates using npm 8. Not sure who actually reads our release notes but can't hurt.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -495,6 +496,16 @@ def write_lockfiles
...
 
+        # Takes a JSON string and detects if it is spaces or tabs and how many
+        # levels deep it is indented.
+        def detect_indentation(json)
+          indentation = json.scan(/^\s+/).min_by(&:length)
+          indentation_size = indentation.length
+          indentation_type = indentation.scan(/\t/).any? ? "\t" : " "
+
+          indentation_type * indentation_size
+        end
```
brrygrdn, line 507: 👏🏻  This is one of those functions that's just so satisfyingly terse, I love it

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -56,8 +56,8 @@ class VersionResolver
         # npm ERR! peer @opentelemetry/api@">=1.0.0 <1.1.0" from @opentelemetry/context-async-hooks@1.0.1
         NPM7_PEER_DEP_ERROR_REGEX =
           /
-            npm\sERR!\sCould\snot\sresolve\sdependency:\n
-            npm\sERR!\speer\s(?<required_dep>\S+@\S+(\s\S+)?)\sfrom\s(?<requiring_dep>\S+@\S+)
+            npm\s(?:WARN|ERR!)\sCould\snot\sresolve\sdependency:\n
+            npm\s(?:WARN|ERR!)\speer\s(?<required_dep>\S+@\S+(\s\S+)?)\sfrom\s(?<requiring_dep>\S+@\S+)
```
brrygrdn, line 60: I'm assuming this is handling an internal change between npm 7/8?

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -458,10 +458,13 @@ def run_npm_checker(path:, version:)
...
-          SharedHelpers.run_shell_command(
+          cmd =
             "npm install #{version_install_arg(version: version)} --package-lock-only --dry-run=true --ignore-scripts"
-          )
-          nil
+          output = SharedHelpers.run_shell_command(cmd)
+          if output.match?(NPM7_PEER_DEP_ERROR_REGEX)
+            error_context = { command: cmd, process_exit_value: 1 }
+            raise SharedHelpers::HelperSubprocessFailed.new(message: output, error_context: error_context)
+          end
```
brrygrdn, line 467: Just to ensure I understand the change, this is an improvement to attach the command being ran to the error context so it's less of a mystery what has happened?

brrygrdn (approved):
This approach makes sense to me, I'm fairly sure the 'guessing' of JSON format will have pretty much the same limitations within NPM itself so this implementation seems great! 🚀

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
jurre: Yeah it is, in 8 now some peer dependency errors have been turned into warnings

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
jurre: Copilot did about half of it to be fair 😄

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
jurre: In npm 8 some peer dependency conflicts don't result in an error, but they print a warning out instead and return a `0` exit code. In order to keep the change as minimal as possible this detects whenever such a warning is printed and turns it into an error.

I initially looked into handling those warnings separately but the change ended up being much more involved and harder to follow IMO.

brrygrdn: Ah, I get it, yeah that makes sense. I agree with this approach of keeping our existing behaviour 👍🏻

PR #4771, opened 2022-02-28
Update npm 7 references to npm 8

jurre:
We recently upgraded npm to use npm 8 instead of 7, but had left
references in the code pointing to npm 7 to minimize the diff on that
change.

This aligns our code with the version we're actually using.


landongrindheim (approved):
Looks good to me! 🥇

@jurre If I had made this change, it would have involved using grep and/or sed. 💡 Could you leave the commands you used to find/replace so they can be referred to in the future?

I think that could be helpful for:
* The next time we make a change like this
* Finding any references that might have been missed (I know Ruby's not very tool-friendly for this kind of task)

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -195,7 +195,7 @@ def run_npm_7_top_level_updater(top_level_dependencies:)
         end
 
         def run_npm_subdependency_updater
-          if npm7?
+          if npm8?
             run_npm_7_subdependency_updater
```
landongrindheim, line 199: Should this be updated as well?

I pulled this branch locally and ran a couple commands to see if I could expand that search:
```plaintext
...
# changelog entries (shouldn't be updated)
...
npm_and_yarn/helpers/lib/npm6/updater.js:// NOTE: Re-used in npm 7 updater
npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb:        # downloader introduced in npm 7
npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb:        # proj). npm 7 introduces workspace support so we explitly want to
npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb:            run_npm_7_top_level_updater(top_level_dependencies: top_level_dependencies)
npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb:        def run_npm_7_top_level_updater(top_level_dependencies:)
npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb:            # the npm 7 rollout
npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb:            run_npm_7_subdependency_updater
npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb:        def run_npm_7_subdependency_updater
npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb:            # TODO: Figure out if we need to lock git deps for npm 7 and can
npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb:              # NOTE: The `from` syntax has changed in npm 7 to inclued the dependency name
```

jurre: Oh good call, the methods I think so, the comments I left as is because they refer to things introduced in npm 7 or the rollout

yeikel: Should this method `run_npm_7_subdependency_updater` be updated?

PR #4778, opened 2022-03-01
pub support for requirement_update_strategy

sigurdm:


pub/lib/dependabot/pub/helpers.rb
```diff
@@ -112,6 +115,21 @@ def to_dependency(json)
         Dependency.new(**params)
       end
 
+      # expects "auto" to already have been resolved to one of the other
+      # strategies.
+      def constraint_field_from_update_strategy(requirements_update_strategy)
+        case requirements_update_strategy
+        when nil
+          "constraint"
```
jonasfj, line 123: This isn't a field anymore. I think `auto` should be `"constraintBumpedIfNeeded"`

pub/lib/dependabot/pub/update_checker.rb
```diff
@@ -78,6 +78,31 @@ def report
...
+
+        if requirements_update_strategy.nil? 
+          # Check for a version field in the pubspec.yaml. If it is present
+          # we assume the package is a library, and the requirement update 
+          # strategy is widening. Otherwise we assume it is an application, and
+          # go for "bump_versions".
+          pubspec = dependency_files.find { |d| d.name == "pubspec.yaml" }
+          begin
+            parsed_pubspec = YAML.safe_load(pubspec.content, aliases: false)
+          rescue ScriptError
```
jurre, line 94: I can't find any references to `YAML.safe_load` raising `ScriptError`, but either way it might be safer to rescue `StandardError` here, do we every want this to prevent updates running with `bump_versions` if the yaml loading fails?

jurre (approved):
This looks great 🎉 

I left some questions in-line, but nothing concerning. Going to run a few dry-runs, but other than that looks good to merge.

Thanks for tackling this!

pub/lib/dependabot/pub/update_checker.rb
```diff
@@ -78,6 +78,31 @@ def report
...
+            return "bump_versions"
+          end
+          if parsed_pubspec["version"].nil?
+            "bump_versions"
+          else
+            "widen_ranges"
+          end
+        else
+          requirements_update_strategy
+        end
```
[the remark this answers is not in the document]
sigurdm: Did this!

pub/lib/dependabot/pub/update_checker.rb
sigurdm: I went by the explanation here https://stackoverflow.com/a/14676819/4809566. SyntaxError doesn't inherit from StandardError...

I don't know the ruby yaml parser well enough, so I thought it would be better to just have a default case here... maybe we should just fail

pub/lib/dependabot/pub/helpers.rb
sigurdm: This is. because we overload this to also parse the "dependency_services list" output. I have split this function in two one for each case. Hope that is clearer.

pub/lib/dependabot/pub/update_checker.rb
jurre: Aha, interesting TIL! Yeah maybe we _should_ actually fail because in this scenario the YAML is invalid and there's probably not much we can do. It would be good to raise a `Dependabot::DependencyFileNotParseable` error in this case though, so that our internal systems know how to handle the job error correctly.

Fwiw, in other places we do:

```ruby
rescue Psych::SyntaxError, Psych::DisallowedClass, Psych::BadAlias
```

Issue #4042, opened 2021-07-13
terraform: add h1 hashes for multiple os_arch in terraform.lock.hcl file


mwieczorek:
## Problem

Dependabot updates terraform.lock.hcl file with h1 hash for new provider version, but only for default platform it runs on. Example PR: https://github.com/mwieczorek/dependabot-test/pull/1

In above example I run locally command

```
terraform providers lock \                      
  -platform=linux_arm64 \
  -platform=linux_amd64 \
  -platform=darwin_amd64 \
  -platform=windows_amd64
```

which produced 4 h1 hashes in file. When depedabot run, it added only one h1 hash (linux_amd64). 

If I merge that PR and then clone the repository locally I may run into one of the issues:

1. ([cached providers](https://www.terraform.io/docs/cli/config/config-file.html#plugin_cache_dir): plugin_cache_dir set in .terraformrc file) - during local `terraform init` I'll get an error like `Error while importing hashicorp/vault v2.21.0 from the shared cache directory: the provider cache at .terraform/providers has a copy of registry.terraform.io/hashicorp/vault 2.21.0 that doesn't match`
2. (no cached providers) -  local`terraform init` will update terraform.lock.hcl file with new h1

## Proposed solution

Add most common os_arch 'platforms' to ['terraform providers lock' command](https://github.com/dependabot/dependabot-core/blob/5042e9c39132b07d6253cfa531ef4f1514d35e9e/terraform/lib/dependabot/terraform/file_updater.rb#L110). F.e. linux_amd64, darwin_amd64, windows_amd64

I understand some providers may not have all of those os_arch combinations, so dependabot should gracefully ignore if one of them is not available.

billinghamj:
I wonder if this could be worked around temporarily with a Github Action workflow which adds to the dependabot PR 🤔

moritzheiber:
Right now, all of dependabot's PRs for Terraform are useless to us because they only include the hashes for `linux_amd64`, and not any of the other architectures. I'd love to see this either turned into an option (e.g. `architectures = [linux_amd64, darwin_amd64]`) or default to "all platforms" by default.

sebbrandt87:
It would be valuable to have to option to extend the platform providers locking.

https://github.com/dependabot/dependabot-core/blob/f6acb90326f54ad7b19d6045ad9d8f9960248855/terraform/lib/dependabot/terraform/file_updater.rb#L115

[...]

PR #4784, opened 2022-03-02
feat(terraform): defined nearly all supported Terraform platforms for locking

sebbrandt87:
For the purpose of Terraform locking, we now have defined nearly all supported Terraform platforms in the `terraform providers lock` command.

Except linux_arm64, as not all providers seem to be present yet for that platform.

This makes sure that we have all h1 sums in the `.terraform.lock.hcl` so that we keep compatability and do not run in errors on runs at other platforms than Linux.

Quote Terraform docs

> If you wish to avoid ongoing additions of new h1: hashes as you work with your configuration on new target platforms, or if you are installing providers from a mirror that therefore can't provide official signed checksums, you can ask Terraform to pre-populate hashes for a chosen set of platforms using [the terraform providers lock command](https://www.terraform.io/cli/commands/providers/lock):

```
terraform providers lock \
  -platform=linux_arm64 \
  -platform=linux_amd64 \
  -platform=darwin_amd64 \
  -platform=windows_amd64
```

> The above command will download and verify the official packages for all of the required providers across all four of the given platforms, and then record both zh: and h1: checksums for each of them in the lock file, thus avoiding the case where Terraform will learn about a h1: equivalent only at a later time. See the terraform providers lock documentation for more information on this command.

resolves: #4042


sebbrandt87:
Hey @jurre 

I wouldn't expect merge conflicts, but for sure larger diffs initially.
We might see if we really get concerns about merge conflicts. I have not seen once in the past, as most of the time a user is not explicitly locking the providers but terraform just does for us automatically depending on the platform we work on.

billinghamj:
FWIW, the current situation is likely far worse, in terms of conflicts with changes made locally

Currently every Windows + Mac user has dependabot continuously removing the right architectures, resulting in a lot of people always having uncommitted changes from running `terraform init` locally

Screenshot 2022-03-04 at 13 51 40

So every time there's a change and they pull, they have to stash/discard those changes

With this, people will no longer need to routinely change their local `.terraform.lock.hcl` files, so merge conflicts are far less likely

sebbrandt87:
correct @billinghamj 😄

sebbrandt87:
@jurre anything still open to be clarified?

Nishnha:
Hi @sebbrandt87 and @billinghamj,

This is something that's been on our radar for a bit. I agree that Dependabot continuously removing the right architectures is pretty frustrating, so I'm open to merging this change in for the time being.

I ran some numbers on the time it took for Terraform to download hashes for mac, linux, and windows (x64) versus just one platform, and it looks like it takes ~60% longer, which makes sense, but isn't ideal. It also results in more outbound calls (where it could error) and bloats the lockfile, like mentioned before.

We discussed a couple solutions:

1. We include a `platform` key in dependabot.yml for Terraform to specify which architecture(s) to fetch. We can leave the default as is (i.e. linux x64)

```yaml
version: 2
updates:
  - package-ecosystem: 'terraform'
    platform: 'linux'
    directory: '/'
    schedule:
      interval: 'daily'
```

2. We guess what architecture hashes to include based on the existing hashes in the lockfile. This would be slightly slower than the first option and is more prone to breaking.

What do y'all think?

sebbrandt87:
Hey @Nishnha 
thank you very much 🥰
I am all in for option 1.

billinghamj:
I'd be very, very happy with either of those options :)

That said, in the temporary absence of either of those - if they wouldn't be realistic in the near term, I'd very much appreciate having this more "brute force" approach in the meantime

Nishnha:
Sounds like consensus.

I'll work with the Dependabot team to track this work and make a release plan. 
Unfortunately, I will have to ship *this* PR next week since it's Friday, and this introduces a significant change to most Terraform users's lockfiles. 

Hope we can resolve this pain point soon!

jurre:
Could we maybe talk to the terraform folks and see if we can get something like a `.tfrc` (like `.npmrc`, but for `tf`) config file that you can commit to the repo? That way you can configure terraform to do the right thing, and Dependabot just reads that config file?

billinghamj:
@jurre I think that is _very_ unlikely to be considered. Terraform is developed quite slowly, and having now reached 1.0, is no longer changing significantly in how it works. There is also already good support for this use-case, and it automatically chooses the right options for the architecture you're using. Adding a feature just for a team inside GitHub seems unlikely

jurre:
> @jurre I think that is _very_ unlikely to be considered. Terraform is developed quite slowly, and having now reached 1.0, is no longer changing significantly in how it works. There is also already good support for this use-case, and it automatically chooses the right options for the architecture you're using. Adding a feature just for a team inside GitHub seems unlikely

Surely users run into this running terraform locally as well, right? 😕

sebbrandt87:
> > @jurre I think that is _very_ unlikely to be considered. Terraform is developed quite slowly, and having now reached 1.0, is no longer changing significantly in how it works. There is also already good support for this use-case, and it automatically chooses the right options for the architecture you're using. Adding a feature just for a team inside GitHub seems unlikely
> 
> Surely users run into this running terraform locally as well, right? 😕 If you run on `linux_arm64` and `darwin_amd64` now you need to always pass those config options to the CLI, having this stored in a config file seems sensible to me for all users, not just us

I bet we can also create a PR to Terraform, so that we maybe get something like a .`tfrc` implemented. But it could be that it is not considered to be a valid point as this is more a workflow / convention thing in each team and if once the .terraform.lock.hcl is written, it is respected.

But as dependabot does atomic updates to the providers, without a options for this convention to be set and respected, we still have this issue.

Can't we proceed as @Nishnha said a few days ago for now, and we come back to remove this "hard" locking in the next iterations of releases / features being implemented? Please ❤️  😊

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -14,6 +14,7 @@ class FileUpdater < Dependabot::FileUpdaters::Base
       PRIVATE_MODULE_ERROR = /Could not download module.*code from\n.*\"(?<repo>\S+)\":/.freeze
       MODULE_NOT_INSTALLED_ERROR =  /Module not installed.*module\s*\"(?<mod>\S+)\"/m.freeze
       GIT_HTTPS_PREFIX = %r{^git::https://}.freeze
+      TF_PLATFORMS = "-platform=linux_amd64 -platform=darwin_amd64 -platform=windows_amd64"
```
billinghamj, line 17: Could we make sure to include Apple Silicon support too?
```suggestion
      TF_PLATFORMS = "-platform=linux_amd64 -platform=darwin_amd64 -platform=darwin_arm64 -platform=windows_amd64"
```

sebbrandt87: I am unsure if we can handle this atm, we would need to make sure that for all basic terraform providers we have the `darwin_arm64` build available. I can tell that we don't have the `linux_arm64` nor `darwin_arm64` available for all, that is why I excluded it for now from the static TF_PLATFORMS. To check some you can have a look at https://releases.hashicorp.com

billinghamj: Mmm this probably needs to be configurable rather than hardcoded then I think - unless e.g. it can just add all available platforms

billinghamj: The number of people using M1 MacBooks now is pretty huge - so I don't think this prevents the pain for "most" at this point really

I think I'll look into making a quick workflow which just updates the dependabot PRs to add on the other providers

billinghamj:
For the benefit of anyone remaining frustrated at this problem still not getting any traction, you can make a workflow to deal with this short-term:

https://gist.github.com/billinghamj/2a127b9cbd475e6b210d10431ccdde30

@sebbrandt87 in case you find it helpful

To GitHub - this still needs to be fixed. It's very disappointing that you agreed to work with us to figure this out, that you said you'd release this PR, and that a month later you have gone silent. As time goes on, it continues to get increasingly difficult to recommend Dependabot to other engineers.

Nishnha:
Hi all, sorry for my general lack of response on this issue. I should have posted this update sooner.

I wanted to try out automatically detecting platform architectures by checking against the existing hashes in `.terraform.lock.hcl`. This work is in https://github.com/dependabot/dependabot-core/pull/4905, and it's proving to be to a reasonable solution to this problem--since it would "just work" for users, and it only slows down the update job by a few seconds (updates usually operate on the scale of minutes).

I realized, however, that if we merge in this PR, we would have already written other architectures to a user's lockfile, and we would need to reverse that if we decide to automatically detect architectures. The only way I can think of reversing this is if we force users manually specify them again with `--platform`. 

Given the "just work" nature that we are going for with Dependabot, we will need to think of a way to reverse this PR or we may unintentionally break Terraform projects for many users. Please let me know if you all think of a solution so we can get this PR merged in.

---

The code for automatically detecting Terraform architectures (https://github.com/dependabot/dependabot-core/pull/4905) is still failing one of its tests and hasn't been fully optimized, but the Dependabot team would like to move forward with that approach if possible.

We also would like to eventually work with Hashicorp to bring a `.tfr` config file to Terraform like Jurre stated, since it would solve the problem for the greater community. If you know any Terraform maintainers, we would love to work with them on scoping out this feature!

Sorry for changing things from what I had originally stated.
Please let me know what you think.

billinghamj:
Thank you @Nishnha - it's great to understand how the work has changed, and makes sense that this could conflict with it. Will follow the other PR longer term :)

Nishnha:
Opened https://github.com/dependabot/dependabot-core/pull/4905 for review

sebbrandt87:
Closing in favor of #4905

PR #4786, opened 2022-03-02
Bump bundler to 2.3.8

jurre:
https://github.com/rubygems/rubygems/blob/master/bundler/CHANGELOG.md#238-february-23-2022


brrygrdn (commented):
I think we need to upgrade the native helpers as well, and just to verify we are targeting 2.3.8, yeah?

bundler/helpers/v2/spec/functions/dependency_source_spec.rb
```diff
@@ -115,6 +115,9 @@
         stub_request(:get, registry_url + "specs.4.8.gz").
           with(basic_auth: ["SECRET_CODES", ""]).
           to_return(status: 400)
+        stub_request(:get, registry_url + "info/business").
```
landongrindheim, line 118: For the sake of posterity (aka future me 😅), could you give some context on why this stub is needed?

bundler/helpers/v2/spec/functions/version_resolver_spec.rb
```diff
@@ -29,6 +29,15 @@
...
   let(:rubygems_url) { "https://index.rubygems.org/api/v1/" }
   let(:old_index_url) { rubygems_url + "dependencies" }
+  let(:gemfury_url) { "https://repo.fury.io/greysteil/" }
+
+  before do
+    stub_request(:get, "https://rubygems.org/quick/Marshal.4.8/statesman-1.2.1.gemspec.rz").
+      to_return(status: 200, body: fixture("rubygems_responses", "statesman-1.2.1.gemspec.rz"))
+
+    stub_request(:get, %r{quick/Marshal.4.8/business-.*.gemspec.rz}).
+      to_return(status: 200, body: fixture("rubygems_responses", "business-1.0.0.gemspec.rz"))
```
landongrindheim, line 39: Same here, I might wonder in a future Bundler update why these had to change here and whether I should be updating them, too. I'd benefit from a quick comment re: why 🙂

landongrindheim (approved):
🥇

brrygrdn (approved):

bundler/helpers/v2/spec/functions/version_resolver_spec.rb
jurre: Same thing here, bundler internally now makes these calls, I didn't try to figure out why it does this because there's not much we can do about it anyway, I just added the requests being made to the list of stubbed requests

PR #4794, opened 2022-03-03
Pub: use system root certificates

jurre:
By default Dart uses it's own certificate store. This is problematic for
us in production where we send traffic through a proxy to add
authentication headers to requests being made, which uses a randomly
generated certificate that we add to the systems cert store, which will
then be ignored by dart.

This instructs dart to use the system certificates instead.

cc @jonasfj, did I explain that correctly?

### DIFF
--- a/pub/lib/dependabot/pub/helpers.rb
+++ b/pub/lib/dependabot/pub/helpers.rb
@@ -60,6 +60,7 @@ module Dependabot
               stdout, stderr, status = Open3.capture3(
                 env.compact,
                 "dart",
+                "--root-certs-file=/etc/ssl/certs/ca-certificates.crt",
                 "--no-analytics",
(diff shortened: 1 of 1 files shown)


landongrindheim (approved):
This looks good to me! I'd love to have a second pair of 👀 from someone on the Dart team if possible 😄

jurre:
The approach was suggested by _jonasfj via email, so going to try this out in our staging environment, given that pub is not available publicly I'm not worried about breaking anything and we can always roll it back.

jonasfj:
Using `--root-certs-file=` is right... and as long as this runs in an ubuntu docker container it should work.

I filed https://github.com/dart-lang/sdk/issues/48506 to make do some better detection in the SDK.

Issue #4329, opened 2021-10-19
Request: Update both the software version and the os version where both are present (in Docker image tags)


amrishparmar:
We have Docker images which use node and look like the following.
```dockerfile
FROM node:14.18.0-alpine3.12
```

We use Dependabot to keep said image up-to-date. So for example, Dependabot will create a PR that updates this to:
```dockerfile
FROM node:14.18.1-alpine3.12
```
since that is the latest version of node 14 (we have an ignore rule for major versions greater than v14).

However, at present Dependabot doesn't appear to update the base OS. In this case, we would like Dependabot to also bump alpine when new releases become available. In this case I would want Dependabot to update `alpine3.12` to `alpine3.14` (the current release). The end result would be
```dockerfile
FROM node:14.18.1-alpine3.14
```

(Apologies if this has already been requested, I couldn't find it when I searched)

jeffwidman:
Related: https://github.com/dependabot/dependabot-core/issues/5538 which was resolved in https://github.com/dependabot/dependabot-core/pull/5560

jeffwidman:
As reported in #5861, this needs to cover:
1. when both node and OS are bumped
2. only node bumped
3. only OS bumped

VKTB:
Any updates on this one?

Issue #2692, opened 2020-10-28
non semver docker tags like python:3.9.0-alpine3.12


vchrisb:
There are many docker image tags that are in the form of `python:3.9.0-alpine3.12` or `golang:1.15-alpine3.12`
In this case `3.9.0` is the python version and `alpine3.12` the actual image version.
I currently don't see a way to use this or similar formats with dependabot.

Maybe having an option to specify a regex for the image tag could be a solution.

missedone:
I have similar maybe more complex request, we use openjdk images, Ex. `openjdk:11.0.2-jre-stretch`

* works from `openjdk:11.0.2-jre-stretch` to `openjdk:11.0.3-jre-stretch`
* works from `openjdk:11.0.9-jre-buster` to `openjdk:11.0.10-jre-buster`
* does NOT work `openjdk:11.0.3-jre-stretch` to `openjdk:11.0.10-jre-buster`, note that both the jre and os version gets updated

jeffwidman:
Closing as a dupe of https://github.com/dependabot/dependabot-core/issues/2692. I realize that one is newer, but it better describes the problem, so let's use that to track the conversation.

PR #4801, opened 2022-03-04
Docker: Handle versions with underscore characters

RodrigoPetter:
Allows the underscore character to be present in the prefix, root or
suffix.

Also fixes a bug where numbers were not allowed to be present in the
prefix or in the suffix when prefix and suffix were present at the
same time.

### DIFF
--- a/docker/lib/dependabot/docker/update_checker.rb
+++ b/docker/lib/dependabot/docker/update_checker.rb
@@ -43,12 +43,12 @@ module Dependabot
   module Docker
     class UpdateChecker < Dependabot::UpdateCheckers::Base
       VERSION_REGEX =
-        /v?(?<version>[0-9]+(?:(?:\.[a-z0-9]+)|(?:-(?:kb)?[0-9]+))*)/i.freeze
-      VERSION_WITH_SFX = /^#{VERSION_REGEX}(?<suffix>-[a-z0-9.\-]+)?$/i.freeze
-      VERSION_WITH_PFX = /^(?<prefix>[a-z0-9.\-]+-)?#{VERSION_REGEX}$/i.freeze
-      VERSION_WITH_PFX_AND_SFX =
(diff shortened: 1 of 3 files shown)


docker/lib/dependabot/docker/update_checker.rb
```diff
@@ -316,7 +316,7 @@ def prerelease?(tag)
       def numeric_version_from(tag)
         return unless tag.match?(NAME_WITH_VERSION)
 
-        tag.match(NAME_WITH_VERSION).named_captures.fetch("version").downcase
+        tag.match(NAME_WITH_VERSION).named_captures.fetch("version").downcase.gsub(/_/i, ".")
```
RodrigoPetter, line 319: Probably a better place to handle this change would be in the `version.rb` file, as has been done for other languages like Gradle, but I wouldn't know how to do that.

deivid-rodriguez: Thank you @RodrigoPetter! I didn't see your PR in time and created a separate one at #5734. Coincidentally, I used the approach you mentioned here used for other ecosystems 👍.

Anyways, this PR was very useful and I reused part of the specs and added attribution to you in the commit ❤️.

RodrigoPetter: Wow @deivid-rodriguez, thank you! I'm glad I could help. 

Probably your PR fixes the original problem that I was facing. My PR doesn't seems necessary anymore.

I would prefer that, for more complex cases of non semver, it would be taken into consideration the cases in the issues #2692 and #4329.

With that in mind, I'm closing my PR.

deivid-rodriguez: Thank you! We'll try to keep improving things and you're welcome to keep sending us patches. We'll try to be more responsive next time!

Issue #4811, opened 2022-03-08
`dry-run` script complains on repo that has dependency files in subdirectories


jeffwidman:
Trying to locally debug the root cause of https://github.com/dependabot/dependabot-core/issues/4625 using a private repo at my day job. 

The `dry-run.rb` script is bombing out with: `DependencyFileNotFound` error.

This particular repo is a bit of a mish-mash, with multiple `go.mod` files in various subdirectories.  I suspect something about the subdirectories is causing the problem.

I verified my local setup is correct by running it against:
1. another repo which doesn't have subdirectories - works perfectly.
2. a non-existent repo - throws a `RepoNotFound` error.

Additionally, the "real" dependabot has no issues finding the subdirectories / `go.mod` files.

This is either a bug in the `dry-run` setup, or if I need to execute the dry run script differently, then that could be documented on the readme.

Here's the full traceback:
```
[dependabot-core-dev] ~/dependabot-core $ bin/dry-run.rb go_modules <redacted>/<redacted>
=> cloning into /home/dependabot/dependabot-core/tmp/<redacted>/<redacted>
Traceback (most recent call last):
	6: from bin/dry-run.rb:508:in `<main>'
	5: from /home/dependabot/dependabot-core/common/lib/dependabot/file_fetchers/base.rb:65:in `files'
	4: from /home/dependabot/dependabot-core/go_modules/lib/dependabot/go_modules/file_fetcher.rb:22:in `fetch_files'
	3: from /home/dependabot/dependabot-core/common/lib/dependabot/shared_helpers.rb:35:in `in_a_temporary_repo_directory'
	2: from /home/dependabot/dependabot-core/common/lib/dependabot/shared_helpers.rb:35:in `chdir'
	1: from /home/dependabot/dependabot-core/common/lib/dependabot/shared_helpers.rb:35:in `block in in_a_temporary_repo_directory'
/home/dependabot/dependabot-core/go_modules/lib/dependabot/go_modules/file_fetcher.rb:27:in `block in fetch_files': Dependabot::DependencyFileNotFound (Dependabot::DependencyFileNotFound)
```

jeffwidman:
Looks like this is simply a case of ignorance on my part... I opened https://github.com/dependabot/dependabot-core/pull/4812 to document this better for folks doing drive-by PRs.

jurre:
[...]

PR #4812, opened 2022-03-08
Note that the `--dir` flag is required for subdirs

jeffwidman:
When I tried to run `dry-run.rb` against a repo where the
dependency files were in a subdirectory, I got a very confusing error.

It took me _way_ too long to realize that parsing `.github/dependabot.yml`
wasn't part of `dependabot-core` therefore I needed to manually pass in
the subdirectory. So document this better for folks doing drive-by PRs.

Temporarily addresses https://github.com/dependabot/dependabot-core/issues/4811, at least until a better runtime error can be emitted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,9 @@ $ bin/dry-run.rb go_modules rsc/quote
 ...
 ```
 
+Note: If the dependency files are not in the top-level directory, then you must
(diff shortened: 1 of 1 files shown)


jurre (approved):
Thanks!

PR #4815, opened 2022-03-08
Improve file_fetcher for GitHub Actions

JonasAlfredsson:
In an earlier [commit][1] we introduced the feature where dependabot
also looked for "action.ya?ml" files in the root directory in addition
to what could be found in the ".github/workflows/" folder. However,
what was not known at the time was that the action.yml files do not
need to reside directly in the root directory, but may be nested in
subdirectories.

While the code (unintentionally) worked with subdirectories it has now
been updated to be a bit more clear in what it will look for.

[1]: https://github.com/dependabot/dependabot-core/commit/c4f3e6eae592c5c42332719b9e597c8bd47a2db9

### DIFF
--- a/github_actions/lib/dependabot/github_actions/file_fetcher.rb
+++ b/github_actions/lib/dependabot/github_actions/file_fetcher.rb
@@ -13,7 +13,7 @@ module Dependabot
       end
 
       def self.required_files_message
-        "Repo must contain a .github/workflows directory with YAML files or an action.yml in the root"
+        "Repo must contain a .github/workflows directory with YAML files or an action.yml file"
       end
 
       private
@@ -28,7 +28,7 @@ module Dependabot
         if incorrectly_encoded_workflow_files.none?
           raise(
             Dependabot::DependencyFileNotFound,
-            File.join(directory, ".github/workflows/<anything>.yml")
+            File.join(directory, "action.yml") + " or /.github/workflows/<anything>.yml"
           )
         else
           raise(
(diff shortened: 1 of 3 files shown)


JonasAlfredsson:
This was created after the discussion that arose here: https://github.com/dependabot/dependabot-core/issues/4178

I also once again just want to highlight that this is my second time writing anything in Ruby, so any comments regarding best practices are welcome :)

JonasAlfredsson:
I am not super fond of having the special case of also scanning the workflows folder when the directory you define is actually just "/", but the alternative is to make a special case where we catch all .yml files in case the directory given is ".github/workflows". 
However, that would break all current dependabot setups which I guess would upset more people. 

A more drastic solution is to split this out into a separate composite_action scanner/ecosystem. But that would cause a lot of duplicate code.

github_actions/lib/dependabot/github_actions/file_fetcher.rb
```diff
@@ -39,11 +39,16 @@ def fetch_files
...
 
       def workflow_files
-        @workflow_files ||=
+        @workflow_files ||= [fetch_file_if_present("action.yml"), fetch_file_if_present("action.yaml")].compact
+
+        # In the special case where the root directory is defined we also scan
+        # the .github/workflows/ folder.
+        return @workflow_files unless directory == "/"
+
+        @workflow_files +=
```
jurre, line 48: I think this will now keep appending these files to the array every time this method is called, where the expected behavior is that all the files are fetched once, and if we have them in memory just return them, what I'd suggest is write something like this:

```ruby
return @workflow_files if defined? @workflow_files

@workflow_files = [fetch_file_if_present("action.yaml") #.. etc]

@workflow_files += ...
```

jurre (commented):
Thanks for improving this! I left one suggestion where I think the code is doing something unintended. 

Overall this direction seems fine to me, I wouldn't split out a whole ecosystem for this

github_actions/lib/dependabot/github_actions/file_fetcher.rb
JonasAlfredsson: Oh, that is why the ||= operator was there before right?
I have changed it, but I still kept the "unless directory == /" check since we do not want to scan that one in case a specific path is given.

jurre (approved):

PR #4817, opened 2022-03-08
Enable codeql

mctofu:
This is the auto suggested workflow.


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,70 @@
...
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main ]
```
Nishnha, line 16: I feel like we might want this to run on all non-Dependabot branches?

Nishnha (approved):
The way CodeQL split their steps into different actions is interesting. I wonder how they handle versioning across them.

Curious to see what this reports in Dependabot-Core. If we do happen to uncover a code vuln, how do we ensure it won't be exposed to non-maintainers?

mctofu:
> If we do happen to uncover a code vuln, how do we ensure it won't be exposed to non-maintainers?

There's not much additional risk there since it's already possible for anyone to clone the repo and run codeql against it on their own.

.github/workflows/codeql-analysis.yml
mctofu: The pull_request trigger should cover other branches but it will be interesting to see how this interacts with Dependabot PRs.

jeffwidman:
@mctofu do you recall where you got this code from? Was it a copy/paste/edit from their readme or do they have some sort of code generator or repo of example files? I've got a go-based repo that I'd like to update the codeQL action on, so like to diff what already exists against latest...

PR #4820, opened 2022-03-09
Fix Ruby version conflict error detection for new Bundler versions

deivid-rodriguez:
Hello!

I'm having some issues running specs locally, and I won't be able to finish this work (and possibly write some specs) today, but I want to open a WIP PR if that's ok, so that I don't lose the context and that I can check whether the current version of my patch is ok with your CI.

On some of my repositories I'm getting mysterious Ruby version conflict errors that make dependabot updates fail. See for example: https://github.com/activeadmin/arbre/security/dependabot/9, which shows this error:

```
Bundler::VersionConflict with message: Bundler found conflicting requirements for the Ruby version:
  In Gemfile:
    Ruby (~> 2.6.7.0)

    rails (~> 7.0.2) was resolved to 7.0.2.2, which depends on
      actionpack (= 7.0.2.2) was resolved to 7.0.2.2, which depends on
        Ruby (>= 2.7.0)
```

However, my library does not impose any Ruby version requirements, so I wasn't sure where this was coming from.

After investigation, I noticed that dependabot is automatically adding this Ruby version constraint to the resolution when updating libraries.

After further investigation, I noticed that dependabot also wants to retry without this constraint, in case adding it fails.  It does this by checking the content of the specific error message given by Bundler in this case. In particular

> Bundler could not find compatible versions __for gem "ruby\0"__

However, in https://github.com/rubygems/bundler/pull/6647 (released with Bundler 2.1.0.pre.1), this message was changed to

> Bundler found conflicting requirements for the __Ruby\0__ version

That means that when using recent Bundler versions, the error message not longer matches, so dependabot leaves the extra Ruby version constraint there, and update fails.

This PR fixes the issue by also checking for the new error message in order to remove the extra Ruby version constraint.

I tried `dependabot-script` and it worked just fine with this change, but I want to check full CI too.

### DIFF
--- a/bundler/lib/dependabot/bundler/file_updater/lockfile_updater.rb
+++ b/bundler/lib/dependabot/bundler/file_updater/lockfile_updater.rb
@@ -16,7 +16,6 @@ module Dependabot
         require_relative "gemspec_updater"
         require_relative "gemspec_sanitizer"
         require_relative "gemspec_dependency_name_finder"
-        require_relative "ruby_requirement_setter"
 
         LOCKFILE_ENDING =
           /(?<ending>\s*(?:RUBY VERSION|BUNDLED WITH).*)/m.freeze
@@ -80,23 +79,10 @@ module Dependabot
               )
             end
           post_process_lockfile(lockfile_body)
-        rescue SharedHelpers::HelperSubprocessFailed => e
-          raise unless ruby_lock_error?(e)
-
-          @dont_lock_ruby_version = true
-          retry
-        end
-
-        def ruby_lock_error?(error)
-          return false unless error.error_class == "Bundler::VersionConflict"
-          return false unless error.message.include?(" for gem \"ruby\0\"")
-          return false if @dont_lock_ruby_version
-
-          dependency_files.any? { |f| f.name.end_with?(".gemspec") }
         end
 
         def write_temporary_dependency_files
-          File.write(gemfile.name, prepared_gemfile_content(gemfile))
+          File.write(gemfile.name, updated_gemfile_content(gemfile))
           File.write(lockfile.name, sanitized_lockfile_body)
 
           top_level_gemspecs.each do |gemspec|
@@ -235,21 +221,6 @@ module Dependabot
           spec&.version || gemspec_specs.first&.version || "0.0.1"
         end
         # rubocop:enable Metrics/PerceivedComplexity
-
-        def prepared_gemfile_content(file)
-          content =
-            GemfileUpdater.new(
-              dependencies: dependencies,
-              gemfile: file
-            ).updated_gemfile_content
-          return content if @dont_lock_ruby_version
-
-          top_level_gemspecs.each do |gs|
-            content = RubyRequirementSetter.new(gemspec: gs).rewrite(content)
-          end
-
-          content
-        end
 
         def updated_gemfile_content(file)
           GemfileUpdater.new(
--- a/bundler/lib/dependabot/bundler/update_checker/version_resolver.rb
+++ b/bundler/lib/dependabot/bundler/update_checker/version_resolver.rb
@@ -139,7 +139,8 @@ module Dependabot
         end
 
         def ruby_lock_error?(error)
-          return false unless error.message.include?(" for gem \"ruby\0\"")
+          return false unless error.message.include?(" for the Ruby\0 version") || # Bundler 2
+                              error.message.include?(" for gem \"ruby\0\"") # Bundler 1
           return false if @gemspec_ruby_unlocked
 
           dependency_files.any? { |f| f.name.end_with?(".gemspec") }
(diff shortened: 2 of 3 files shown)


deivid-rodriguez:
I noticed that when Bundler v1 and Bundler v2 behavior differ, you extract interface methods that then get their own version specific implementation at `bundler/helper/v1` and `bundler/helper/v2`. I was not sure it was worth doing that here, but I'm happy to do it if that's the way to go!

jurre:
> I noticed that when Bundler v1 and Bundler v2 behavior differ, you extract interface methods that then get their own version specific implementation at `bundler/helper/v1` and `bundler/helper/v2`. I was not sure it was worth doing that here, but I'm happy to do it if that's the way to go!

No this is fine! Those extracted interface methods are programs that we run as standalone CLIs and parse the results back to the main process. This is mostly used for when we use bundler as a library in order to avoid mismatches between the bundler used to manage dependabot-core itself and the one used to perform updates

As for running locally, are you using `bin/docker_dev_shell`? If not I'd recommend trying that route, running the project stand-alone is very hard and not well supported

deivid-rodriguez:
> No this is fine! Those extracted interface methods are programs that we run as standalone CLIs and parse the results back to the main process. This is mostly used for when we use bundler as a library in order to avoid mismatches between the bundler used to manage dependabot-core itself and the one used to perform updates

Ah, I see!

> As for running locally, are you using bin/docker_dev_shell? If not I'd recommend trying that route, running the project stand-alone is very hard and not well supported

No, indeed I tried running what's in the `script/ci-test` directly. I actually got it to work, at least for bundler 2, with a minor modification that I will contribute next. But I will use `bin/docker_dev_shell` from now on!

> Also, thanks for contributing!

You're welcome, thanks to you for dependabot :)

deivid-rodriguez:
Turns out that while my changes fix my use case, breaks a different use case. The difference between our use cases is the presence of a lockfile. In my case, there's a `Gemfile.lock` file so dependabot should not add any extra magic about Ruby version requirements. If there's no lockfile though, it's totally reasonable that dependabot tries to make sure resolution supports the oldest Ruby supported by the library, and that's the spec this is breaking.

I'll dig deeper.

deivid-rodriguez:
I fixed the broken spec in the Bundler 2 suite, and added a spec for this fix. But now I have a broken spec in the Bundler 1 suite 😬. Will keep digging.

deivid-rodriguez:
I think the current failure is some flaky failure, but other than that seems green.

However, I need to look closer at the consequences of doing this. It fixes my use case but I think it might break other cases.

I imagine now if a library provides a Gemfile + Gemfile.lockfile + gemspec, then dependabot could potentially change a gemspec dependencies to a range that no longer resolves on the minimum ruby supported, and that would be bad since it would mean unintentionally dropping old rubies support (although it would most likely be detected by CI running on the minimum ruby since the new lockfile would no longer be installable there).

It seems like a very rare case, but I want to double check the new behavior.

jurre:
> I think the current failure is some flaky failure, but other than that seems green.

Yes, we unfortunately have a few flaky tests in our bundler suite, one of which is disabled [here](https://github.com/dependabot/dependabot-core/pull/4823)

jurre:
@deivid-rodriguez is this ready for review?

deivid-rodriguez:
Still a WIP and I'm also not even sure this makes sense in general, or just for my use case. In my libraries I use a `lockfile_only` versioning strategy, so it definitely makes sense in that context. But I'm not sure if ignoring this ruby requirement dependabot adds is too aggressive for other cases.

So, not ready for a review, but happy to get feedback about the idea.

deivid-rodriguez:
Alright, so I had another look at this and my conclusion is that this is a bug introduced when Bundler 2 was added. I did my best to explain the context, and the relevant commits/PRs both in the dependabot and bundler side. The explanation is at https://github.com/dependabot/dependabot-core/pull/4820/commits/a2def5453fbe47806466e7b073d390917a45c06e.

However, I tried this PR against my repository and while it did successfully generate a PR (see https://github.com/activeadmin/activeadmin/pull/7400), it updated the `Gemfile` which I guess should be "forbidden" given that we use the `lockfile-only` versioning strategy in our configuration. And I think that issue has been introduced by this PR, because an analogous and recent PR did not have this issue (see https://github.com/activeadmin/activeadmin/pull/7390).

So... I need to investigate that latter issue more.

jurre (approved):

jeffwidman:
> So, not an issue related to this PR, so I'm marking this as ready!

While outside the scope of this PR, it sounds like this is a real issue that needs addressing. Are you planning to tackle this issue in a separate PR or file an issue so all your research work doesn't get forgotten about?

jurre:
Maybe we should just retire dependabot-script, no one is really maintaining it and it really only serves as an example of how one might use dependabot-core. I wish the code from our dry-run script was a little more composable/reusable and folks could just use that instead, but alas.

deivid-rodriguez:
I'm not sure what are the differences between the two, but they look very similar. I guess (from naming) one difference is that dry-run.rb is not supposed to actually create PR's.

I guess it makes sense to merge them and retire eventually retire dependabot-script. I'm happy to contribute my fix to dependabot-script to start reducing the delta between the two scripts and open an issue to eventually merge them. How about that?

PR #4822, opened 2022-03-09
Switch VCR record mode to `:none`

Nishnha:
This avoids recording cassettes when we run our tests in CI in case the PR author didn't already record them. This also helps us catch instances where the cassette itself might be poorly recorded; a test with the :vcr annotation can have its cassette change in subsequent runs if the test doesn't contain a description

This may reveal that we have other poorly recorded tests in our codebase


mctofu:
I cleaned up a few more unpinned cassettes in https://github.com/dependabot/dependabot-core/pull/4953

jeffwidman (approved):
This seems reasonable to me. Curious what others think though?

> We might want to set up a cron job to try running this once a week so we can keep our tests up-to-date after this merges

IMO we should definitely do that.

deivid-rodriguez:
Maybe `:once` is still useful locally to get cassettes automatically generated when writing new specs. Then CI can have a more strict mode.

Periodically regenerating cassettes also sounds good.

Finally, we could also add a CI check that rejects unused cassettes. For example, when you write a spec, generate cassette, then rename the spec, a differently named cassette is generated, and you commit everything, including an empty cassettes. Here's a blog post on that: https://www.jdeen.com/blog/detecting-unused-vcr-cassettes.

Nishnha:
> Maybe `:once` is still useful locally to get cassettes automatically generated when writing new specs.

The `:once` VCR mode can still be set locally with an [environment variable](https://github.com/dependabot/dependabot-core/pull/4822/files#diff-a2f6f0e566271a72a272c6fd258923a9f0dd0130ea86374986b52794aca76c54R96)!

When running a spec you can do something like `VCR=once bundle exec rspec spec/...` and it will capture a recording.

> Periodically regenerating cassettes also sounds good.

> Finally, we could also add a CI check that rejects unused cassettes. For example, when you write a spec, generate cassette, then rename the spec, a differently named cassette is generated, and you commit everything, including an empty cassettes. Here's a blog post on that: https://www.jdeen.com/blog/detecting-unused-vcr-cassettes.

I like these ideas, but I think we should do them as a follow up PR?

deivid-rodriguez:
Yep, that sounds great!

PR #4833, opened 2022-03-11
fixes comparison between prerelease identifiers for NuGet

### LINKED ISSUE #1972: Semver 2.0 versions not handled correctly for nuget pre-release dependencies
When the bot open PRs for pre-releases the version parsing doesn't work properly for semver 2.0 versions.

https://github.com/Particular/NServiceBus.Gateway.Sql/pull/34 shows this where `3.2.0-alpha.66` is a higher version than `3.2.0-alpha0014` so no PR should have been created.

jakecoffman:
Fixes #1972

The pre-release section of the version string was being compared as a single string. Now it is correctly breaking up the dot-separated pre-release identifier and comparing those parts numerically when possible per the semver spec.

Verified this was the issue by running `bin/dry-run.rb nuget "ecoAPM/SrcSet" --dir="/." --commit=d2ec33dd1ea80b40a4f163780281cd4678d5db12 --cache=files` and seeing it no longer bumps Statiq.Web from 1.0.0-beta.33 to 1.0.0-beta.9, instead it chooses 1.0.0-beta.46.

### DIFF
--- a/nuget/lib/dependabot/nuget/version.rb
+++ b/nuget/lib/dependabot/nuget/version.rb
@@ -88,8 +88,8 @@ module Dependabot
       end
 
       def compare_dot_separated_part(lhs, rhs)
-        return 1 if lhs.nil?
-        return -1 if rhs.nil?
+        return -1 if lhs.nil?
+        return 1 if rhs.nil?
 
         return lhs.to_i <=> rhs.to_i if lhs.match?(/^\d+$/) && rhs.match?(/^\d+$/)
 
--- a/nuget/lib/dependabot/nuget/version.rb
+++ b/nuget/lib/dependabot/nuget/version.rb
@@ -74,8 +74,26 @@ module Dependabot
 
         return -1 if prerelease_string && !other_prerelease_string
         return 1 if !prerelease_string && other_prerelease_string
+        return 0 if !prerelease_string && !other_prerelease_string
 
-        prerelease_string.<=>(other_prerelease_string)
+        split_prerelease_string = prerelease_string.split(".")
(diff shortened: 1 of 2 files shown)


nuget/lib/dependabot/nuget/version.rb
```diff
@@ -74,8 +74,26 @@ def compare_prerelease_part(other)
...
+        0
+      end
+
+      def compare_dot_separated_part(lhs, rhs)
+        return -1 if lhs.nil?
+        return 1 if rhs.nil?
+
+        return lhs.to_i <=> rhs.to_i if lhs.match?(/^\d+$/) && rhs.match?(/^\d+$/)
+
+        lhs.<=>(rhs)
```
jurre, line 96: *Total nit that you're more than free to ignore!*

I know this is following the pattern present in this file, but it's a pretty unidiomatic way to write this, operators in ruby (`<`, `==` and also the `<=>` "spaceship" operator) are all actually "just" methods, which is why calling them like `foo.<=>(bar)` is possible. However, typically in ruby you wouldn't write code like `if 1.==(2)` even though it's syntactically valid, you would write `if 1 == 2`. Same goes here.

```suggestion
        lhs <=> rhs
```

jurre (approved):
🎉 this makes sense to me! I left one pretty nitty code-style comment, and I do worry a little about the `#compare_prerelease_part` method growing even bigger than it is now (we have linting for complexity disabled for that method), but I couldn't think of an obvious refactor that would make the code easier to read, so I think this is good as is 👍

jakecoffman:
Still learning Ruby so style feedback is always welcome.

Also forgot to say thanks to @Nishnha for the help on this PR!

Issue #2057, opened 2020-06-29
Add global ARG support for Dockerfiles


ffscl:
**Feature request**: Dependabot to support `ARG` with default values in any `FROM` statement when resolving dependencies.

Docker supports `ARG` instructions before the first `FROM` instruction. (https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact).  The `ARG` instruction should be easy to parse and evaluate.

Example:
```
ARG REGISTRY=docker.io
FROM ${REGISTRY}/ubuntu:18.04@sha256:2695d3e10e69cc500a16eae6d6629c803c43ab075fa5ce60813a0fc49c47e859
```
Before checking for new versions, the `FROM` instruction should be evaluated as:
```
FROM docker.io/ubuntu:18.04@sha256:2695d3e10e69cc500a16eae6d6629c803c43ab075fa5ce60813a0fc49c47e859
```

**Use case**: Using a registry mirror that dependabot doesn't have access to.  With this feature in your CI you could override the registry as follows without breaking dependabot:
```
docker build --build-arg REGISTRY=hub.mirror.example.com .
```

ruffsl:
Another example where this would helpful is when the `ARG` directive is used `FROM` multistage Dockerfiles:

``` Dockerfile
ARG FROM_IMAGE=osrf/ros2:testing-20210605003201

# multi-stage for caching
FROM $FROM_IMAGE AS cacher
...

# multi-stage for building
FROM $FROM_IMAGE AS builder
...

```

joebowbeer:
Another example where this would helpful is when the ARG defines the version:

```dockerfile
ARG NODE_VERSION=14.20

FROM node:${NODE_VERSION}-alpine as code
```

joebowbeer:
@ruffsl thanks for the reality check. I realize that renovate doesn't handle that directly either.

I'll simplify my test case to:

```dockerfile
ARG NODE_BASE=node:14.19-alpine

FROM ${NODE_BASE} as code
```

I would like dependabot to create one PR updating to `node:14.20-alpine` and another updating to `node:18.10-alpine`

ralftar:
How about doing something like this:

```
ARG ARCHITECTURE
FROM eclipse-mosquitto:2.0.14-openssl AS amd64
FROM arm64v8/eclipse-mosquitto:2.0.14-openssl AS arm64
FROM ${ARCHITECTURE} AS final
```

 Does the trick for me. Bumps both architectures.

[...]

PR #4837, opened 2022-03-12
Docker: Support ARG FROM

mctofu:
This was originally introduced by @hfhbd in https://github.com/dependabot/dependabot-core/pull/4598.

@jurre added a fix in https://github.com/dependabot/dependabot-core/pull/4723.

In testing we found we'd merged this pre-maturely as we are missing updates to the file updater (https://github.com/dependabot/dependabot-core/pull/4723#issuecomment-1034082693) so we temporarily reverted in https://github.com/dependabot/dependabot-core/pull/4759.

This PR re-stages those changes but we still need to implement the file updater part.

Fix https://github.com/dependabot/dependabot-core/issues/2057


aanm:
@mctofu any update on this topic? Thank you!

PR #4841, opened 2022-03-13
Get bundler specs green when run on Ruby 3.1

deivid-rodriguez:
I was getting 3 spec failures when running dependabot bundler specs on my system.

I debugged it, and the cause turned out to be that I'm using Ruby 3.1.

This PR makes bundler specs fully pass on Ruby 3.1.

Only the first commit is necessary, the second one only fixes some annoying VCR warnings.

### DIFF
--- a/bundler/helpers/v1/Gemfile
+++ b/bundler/helpers/v1/Gemfile
@@ -7,6 +7,6 @@ group :test do
   gem "debug", ">= 1.0.0"
   gem "rspec", "~> 3.8"
   gem "rspec-its", "~> 1.2"
-  gem "vcr", "6.0.0"
+  gem "vcr", "6.1.0"
   gem "webmock", "~> 3.4"
 end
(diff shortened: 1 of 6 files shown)


deivid-rodriguez:
The VCR update makes Ruby 2.6 the minimum version to develop dependabot, since VCR 6.1.0 no longer supports Ruby 2.5.

Since dependabot runs Ruby 2.7 since https://github.com/dependabot/dependabot-core/commit/2a6611f69e5dc0880f5e353512a59cdaa75a4427, I guess we could even explicitly bump the required_ruby_version to 2.7?

jurre (approved):
Great, thanks! We're not quite ready to run on ruby 3.1 just yet (there's a few systems that we use to run dependabot at GitHub that we need to update together), but this will be helpful once we are ready!

deivid-rodriguez:
I see, yeah, it should make a future upgrade easier. About bumping the `required_ruby_version` to 2.7, do you think it makes sense?

PR #4843, opened 2022-03-14
Unify RSpec requirements in helpers Gemfiles to match dependabot-common

deivid-rodriguez:
I run into a similar issue as #4838, because `bundle install` had installed RSpec versions compatible with `Dependabot-common` dependencies (RSpec 3.11), but these didn't match the specification in these Gemfiles (RSpec 3.10.2).

This PR proposes the same solution as in #4838. Now all development dependencies are consistent.


landongrindheim (approved):
Thanks!

deivid-rodriguez:
You're welcome!

PR #4853, opened 2022-03-15
Bump to `go` `1.18`

jeffwidman:
From https://golang.org/dl/:
* go1.18.linux-amd64.tar.gz
* `e85278e98f57cdb150fe8409e6e5df5343ecb13cebf03a5d5ff12bd55a80264f`

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
@@ -162,8 +162,7 @@ module Dependabot
 
           File.write(tmp_go_file, "package dummypkg\n") unless package
 
-          # TODO: go 1.18 will make `-d` the default behavior, so remove the flag then
-          command = +"go get -d"
+          command = +"go get"
           # `go get` accepts multiple packages, each separated by a space
           dependencies.each do |dep|
             version = "v" + dep.version.sub(/^v/i, "")
@@ -195,7 +194,7 @@ module Dependabot
 
         def build_module_stubs(stub_paths)
           # Create a fake empty module for each local module so that
-          # `go get -d` works, even if some modules have been `replace`d
+          # `go get` works, even if some modules have been `replace`d
           # with a local module that we don't have access to.
           stub_paths.each do |stub_path|
             Dir.mkdir(stub_path) unless Dir.exist?(stub_path)
--- a/Dockerfile
+++ b/Dockerfile
@@ -167,8 +167,8 @@ USER root
 ### GO
 
 # Install Go
-ARG GOLANG_VERSION=1.17.7
-ARG GOLANG_CHECKSUM=02b111284bedbfa35a7e5b74a06082d18632eff824fd144312f6063943d49259
+ARG GOLANG_VERSION=1.18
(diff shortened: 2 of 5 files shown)


jeffwidman:
Still needs a test fixture or two that explicitly lists `go 1.18`.

But I'm semi-blocked on locally running `go 1.18` until https://github.com/Homebrew/homebrew-core/pull/91369 lands... based on `1.17`, it took a couple of weeks.

jurre:
I'll chalk this up to a race condition: https://github.com/dependabot/dependabot-core/pull/4855 😄 

But let's go with your PR 👍

jeffwidman:
>> Also need to update [dependabot/gomodules-extracted](https://github.com/dependabot/gomodules-extracted/)

> I don't think it's used anymore right?

Yep, my bad.

go_modules/spec/fixtures/projects/nonexisting_proxy/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/dependabot/vgotest
 
-go 1.18
+go 1.18 // TODO why was this already at 1.18?
```
jeffwidman, line 3: @jurre any idea? Was this accident or on-purpose? Just wondering if it's supposed to be always bumped to a non-existing version of go (which means the test needs changing because there's no way this will get remembered on every bump of go).

jurre: I'm not sure 🤔 I see that I added this file, but I don't think it's relevant for the test. I think the main thing that we need to make sure of is that we don't change the  go version specified in the go.mod file. I don't think what is in there is used, we strip it out when copying over the file: https://github.com/dependabot/dependabot-core/blob/add0e471c1920e2a726273aee7f84a37429b31c6/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb#L115-L127

jeffwidman: Ah, I see what this test is actually trying to test--that a non-existing proxy (`dependabot.com/nonexistingdep`) error gets handled correctly. This has nothing to do with the `go 1.18`. But should be safe enough to leave as-is (I'll remove the TODO note of course).

jeffwidman:
The `go_modules` test failure is fixed in https://github.com/dependabot/dependabot-core/pull/4868

I broke that out separate because I think the bug is not version specific, it just
happens to only get exposed by the test suite when updating to `go 1.18`. So
keeping as two separate PR's means that if this version upgrade has to be
reverted, that change won't also be reverted.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -162,8 +162,7 @@ def run_go_get(dependencies = [])
 
           File.write(tmp_go_file, "package dummypkg\n") unless package
 
-          # TODO: go 1.18 will make `-d` the default behavior, so remove the flag then
-          command = +"go get -d"
+          command = +"go get"
```
mctofu, line 165: 🎉 Thanks for cleaning that up!

mctofu (commented):
Looks good but lets rebase after merging https://github.com/dependabot/dependabot-core/pull/4868 so we can see the go tests pass?

mctofu (approved):
👍

PR #4855, opened 2022-03-15
Bump go to 1.18

jurre:
https://tip.golang.org/doc/go1.18, it's a pretty generic release.


Nishnha (approved):

Nishnha:
Thanks for linking to where we can get the SHA.
We don't have comments like that in other services (looking at you, job-runner) and so I'll open a patch to include it there

PR #4394, opened 2021-11-10
Bump Ubuntu to 20.04

jeffwidman:
`18.04` is getting a bit long in the tooth... `20.04` is the most recent LTS, and `22.04` isn't too far in the future.

I'm sure this will break a bunch of things, so let's try it and see what the tests say...

PR #26, opened 2017-08-01
Revert "Bump fs-extra from 3.0.1 to 4.0.1 in /helpers/javascript"

greysteil:
Reverts dependabot/dependabot-core#25

PR #4858, opened 2022-03-15
Dockerfile: arm64 support

thepwagner:
This PR amends the `Dockerfile` and `bin/docker-dev-shell` commands to support native builds on arm64, as provided by an M1 Mac.

* Initial commits walk the Dockerfile and provide arm64-native alternatives when binaries are fetched.
   * There is no Elm, sorry.
   * There are no Terraform Debian packages, so I switched to fetching zips.
   * In places where the digest of downloaded files is verified, you have two digests to maintain now.
* No attempt is made to build `arm64` in CI, or publish `arm64` releases.

The resulting build stands up OK to light testing: I could dry-run a golang, node and gradle repository.

I hit an error in the bundler that I didn't try to debug:
```
	 1: from /home/dependabot/dependabot-core/bundler/lib/dependabot/bundler/update_checker/shared_bundler_helpers.rb:54:in `block in in_a_native_bundler_context'
/home/dependabot/dependabot-core/bundler/lib/dependabot/bundler/update_checker/conflicting_dependency_resolver.rb:44:in `block in conflicting_dependencies': undefined method `name' for nil:NilClass (NoMethodError)
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -219,13 +219,13 @@ RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain 1.58.0 --pr
 
 USER root
 ARG TERRAFORM_VERSION=1.1.6
-RUN curl -fsSL https://apt.releases.hashicorp.com/gpg | apt-key add -
-RUN apt-add-repository "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main" \
-  && apt-get update -y \
-  && apt-get install -y --no-install-recommends terraform=${TERRAFORM_VERSION} \
-  && terraform -help \
-  && rm -rf /var/lib/apt/lists/*
-
+ARG TERRAFORM_AMD64_CHECKSUM=3e330ce4c8c0434cdd79fe04ed6f6e28e72db44c47ae50d01c342c8a2b05d331
+ARG TERRAFORM_ARM64_CHECKSUM=a53fb63625af3572f7252b9fb61d787ab153132a8984b12f4bb84b8ee408ec53
+RUN cd /tmp \
+  && curl -o terraform-${TARGETARCH}.tar.gz https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_${TARGETARCH}.zip \
+  && printf "$TERRAFORM_AMD64_CHECKSUM terraform-amd64.tar.gz\n$TERRAFORM_ARM64_CHECKSUM terraform-arm64.tar.gz\n" | sha256sum -c --ignore-missing - \
+  && unzip -d /usr/local/bin terraform-${TARGETARCH}.tar.gz \
+  && rm terraform-${TARGETARCH}.tar.gz
 
 ### DART
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -108,11 +108,15 @@ RUN curl -sL https://deb.nodesource.com/setup_16.x | bash - \
 ### ELM
 
 # Install Elm 0.19
+# This is amd64 only, see:
+# - https://github.com/elm/compiler/issues/2007
+# - https://github.com/elm/compiler/issues/2232
 ENV PATH="$PATH:/node_modules/.bin"
-RUN curl -sSLfO "https://github.com/elm/compiler/releases/download/0.19.0/binaries-for-linux.tar.gz" \
+RUN [ "$TARGETARCH" != "amd64" ] \
+  || (curl -sSLfO "https://github.com/elm/compiler/releases/download/0.19.0/binaries-for-linux.tar.gz" \
   && tar xzf binaries-for-linux.tar.gz \
   && mv elm /usr/local/bin/elm19 \
-  && rm -f binaries-for-linux.tar.gz
+  && rm -f binaries-for-linux.tar.gz)
 
 
 ### PHP
--- a/Dockerfile
+++ b/Dockerfile
@@ -234,7 +234,9 @@ ENV PUB_CACHE=/opt/dart/pub-cache \
   PUB_ENVIRONMENT="dependabot" \
   PATH="${PATH}:/opt/dart/dart-sdk/bin"
 ARG DART_VERSION=2.15.1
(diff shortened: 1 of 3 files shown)


jeffwidman:
I opened a PR a while back for the `focal` upgrade: https://github.com/dependabot/dependabot-core/pull/4394

Although at this point `jammy` gets released next month, so if it's hard for the core team to pull in [due to scheduling](https://github.com/dependabot/dependabot-core/pull/4394#issuecomment-969275077), they may want to jump straight to that--better to test one release well than to test two releases less rigorously.

jurre:
Thanks @thepwagner! Bumping ubuntu seems reasonable to me, although I'd like to roll that out separately, which I might as well do now, and so will.

@jeffwidman yeah it's not a bad call, although I kind of don't want to wait even longer, and we should just prioritize rolling that out when it's released. We have a little bit more slack on the team now so we should be able to prioritize this type of work quicker

Dockerfile
```diff
@@ -250,6 +260,8 @@ RUN curl --connect-timeout 15 --retry 5 "https://storage.googleapis.com/flutter_
   && tar xf "/tmp/flutter.xz" -C /opt/dart \
   && rm "/tmp/flutter.xz" \
   && chmod -R o+rx "/opt/dart/flutter" \
+  && rm -Rf /opt/dart/flutter/bin/cache/dart-sdk \
```
mctofu, line 263: I'm curious about these changes since they seem unrelated to arm support? I think this ends up getting deleted anyway below in the `find ... -delete` command?

pavera: Not an expert, but it appears flutter ships with a dart-sdk, these 2 lines replace the x86 dart sdk with the arm sdk downloaded in the previous step.  I say this because without these 2 lines I get the following error during docker build:
`#26 47.08 qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory`

thepwagner: Everyone is correct 🎉 :

The flutter SDK (even after https://github.com/dependabot/dependabot-core/pull/4904/files), embeds an `x86_64` dart sdk:

```
$ file /opt/flutter/new/flutter/bin/cache/dart-sdk/bin/dart 
/opt/flutter/new/flutter/bin/cache/dart-sdk/bin/dart: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, stripped
```

And the files are deleted by the find, so they aren't part of the final image.

IIUC the issue is L266 where `flutter --version` is called before the delete happens - that uses the bundled `dart-sdk` and fails, so I ninja-patched the architecture-appropriate `dart-sdk` in.

If `flutter --version` is just meant to "hello world"/verify the installation, removing that step is an alternative to this hack.

mctofu: Ah, got it, thanks! I don't think we need to make it conditional, I was just curious why it was necessary. This is worth adding a comment about?

cc: @jonassf @sigurdm to see if you have any feedback here.

sigurdm: > If flutter --version is just meant to "hello world"/verify the installation, removing that step is an alternative to this hack.

I think so! When you clone flutter from github the first invocation will create the `version` file (that pub relies on) - that is why we did this step, but the packaged flutter contains it already - so I think it can just be removed!

mctofu (approved):
Thanks!

jurre:
Just running it on my old, old non-arm64 laptop to make sure that still works, but assume it will after which I'll merge this in. Thanks @thepwagner!

jurre:
Hmm, running into a little snag:

```
 => ERROR [stage-0 13/38] RUN cd /tmp   && curl --http1.1 -o go-x86_64.tar.gz https://dl.google.com/go/go1.18.linux-x86_64.tar.gz   && printf "e85278e98f57cdb150fe8409e6e5df5343ecb13cebf03a5d5ff12bd55a80264f go-amd64.tar.gz\n7ac7b396a691e588c5fb57  0.6s
------
 > [stage-0 13/38] RUN cd /tmp   && curl --http1.1 -o go-x86_64.tar.gz https://dl.google.com/go/go1.18.linux-x86_64.tar.gz   && printf "e85278e98f57cdb150fe8409e6e5df5343ecb13cebf03a5d5ff12bd55a80264f go-amd64.tar.gz\n7ac7b396a691e588c5fb57687759e6c4db84a2a3bbebb0765f4b38e5b1c5b00e go-arm64.tar.gz\n" | sha256sum -c --ignore-missing -   && tar -xzf go-x86_64.tar.gz -C /opt   && rm go-x86_64.tar.gz:
#19 0.344   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
#19 0.344                                  Dload  Upload   Total   Spent    Left  Speed
100  1449  100  1449    0     0   7748      0 --:--:-- --:--:-- --:--:--  7748
#19 0.537 sha256sum: 'standard input': no file was verified
------
```

I can dig into that later

thepwagner:
@jurre I think the issue is with how `bin/docker-dev-shell` is setting `TARGET_ARCH`.

From `-o go-x86_64.tar.gz` in the output, we can reason `TARGETARCH=x86_64` for you, but from `printf "... go-amd64.tar.gz"`, I expected `TARGETARCH=amd64`.

In CI, the `bin/docker-dev-shell` script isn't executed - so it uses the built-in default of `amd64` and works great.
If `uname -m` returns `x86_64` for you, then we probably just need a little extra script in `docker-dev-shell` to map that to `amd64`.

I'm confident that `TARGET_ARCH=arm64` works as expected on M1:
```
$ uname -a
Darwin trolololol.local 21.4.0 Darwin Kernel Version 21.4.0: Mon Feb 21 20:35:58 PST 2022; root:xnu-8020.101.4~2/RELEASE_ARM64_T6000 arm64
$ uname -m
arm64
```

pavera:
Yes @thepwagner I also have confirmed that the same error occurs on amd64 and arm64.  The dry run script for bundler expects/requires a lock file in the repo used for testing.  I've also confirmed that bundler works on both amd64 and arm64 as long as the repo has a lock file.

jurre:
image

PR #4862, opened 2022-03-16
NuGet versioning does not consider the "build part" in comparisons

### LINKED ISSUE #4845: Nuget Unnecessarily updates SemVer2 build component
Dependabot is unnecessarily creating updates/PRs for Nuget where the before and after refer to the **exact same underlying package version**, but with added SemVer2 data. 

**Package ecosystem**
Nuget
**Package manager version**
6.1.0
**Language version**
C#10/.NET6.0 

`csproj` file with the following dependencies (see below for minimal example) 

- Dependency: Serilog.Exceptions
  - Initial Version: `8.1.0`
  - Updated to: `8.1.0+build.52`
- Dependency: Serilog.Enrichers.Span
  - Initial Version: `2.2.0`
  - Updated to: `2.2.0+build.276`

Dependabot is updating package versions that are already technically correct, but substituting in the full SemVer2 version. This creates an unnecessary pull request and is not in line with the instructions on nuget.org or Package Manager behavior. 

For example, the current version of [Serilog.Exceptions](https://www.nuget.org/packages/Serilog.Exceptions/) on nuget.org is listed as `8.1.0` but with a SemVer2 version of  `8.1.0+build.52`. Similarly, [Serilog.Enrichers.Span](https://www.nuget.org/packages/Serilog.Enrichers.Span/) is listed officially as `2.2.0` but with a SemVer2 of `2.2.0+build.276`. 

 My project file was already referencing `8.1.0` and `2.2.0` respectively but dependabot created an update for each with the following diffs:

```diff
-        <PackageReference Include="Serilog.Exceptions" Version="8.1.0" />
+        <PackageReference Include="Serilog.Exceptions" Version="8.1.0+build.526" />
```
and
```diff
-        <PackageReference Include="Serilog.Enrichers.Span" Version="2.2.0" />
+        <PackageReference Include="Serilog.Enrichers.Span" Version="2.2.0+build.276" />
```

IMO this is incorrect for three reasons. First, they technically are the **same version**. Second, per nuget.org, the *official*/listed versions do not include the build component. Finally, the instructions for installing he packages also do not include the extra data.

For example, the instructions for Serilog.Exceptions show  the project file:
```xml
<PackageReference Include="Serilog.Exceptions" Version="8.1.0" />
```
And the CLI:
```bash
dotnet add package Serilog.Exceptions --version 8.1.0
```
Note that the extra SemVer2 build information is absent in both cases. If you add the package via Nuget Package Manager it will not have the added build info either. 

Updating with the native package manager does not add the extra SemVer2 build data. 

Note however that this is **not** an error case. The version *is* respected (or ignored?) during restore without generating an error. 

See above

```xml
<Project Sdk="Microsoft.NET.Sdk.Web">
    <PropertyGroup> 
        <TargetFramework>net6.0</TargetFramework>
    </PropertyGroup> 
    <ItemGroup>
        <PackageReference Include="Serilog.Exceptions" Version="8.1.0" />
        <PackageReference Include="Serilog.Enrichers.Span" Version="2.2.0" />
    </ItemGroup>
</Project>
```
**Notes**
I am running a customized version of the dependabot script on my Azure DevOps repo. I run inside of a container built from the `Dockerfile` at the root of this repository (so I can customize the UID and GID).

My current version of `dependabot-omnibus` is `0.178.0` though this exhibits in `0.177.0` as well. Likely related to #4255

jakecoffman:
The [NuGet docs](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#semantic-versioning-200) specify they use SemVer 2.0 [which says](https://semver.org/#spec-item-10):

> Build metadata MUST be ignored when determining version precedence.

Also the NuGet docs say [a little further down](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#where-nugetversion-diverges-from-semantic-versioning) that they depart from SemVer 2.0 in a few ways:

> 1. NuGetVersion supports a 4th version segment, Revision, to be compatible with, or a superset of, [System.Version](https://docs.microsoft.com/en-us/dotnet/api/system.version). Therefore, excluding prerelease and metadata labels, a version string is Major.Minor.Patch.Revision. As per version normalization described above, if Revision is zero, it is omit from the normalized version string.
> 2. NuGetVersion only requires the major segment to be defined. All others are optional, and are equivalent to zero. This means that 1, 1.0, 1.0.0, and 1.0.0.0 are all accepted and equal.
> 3. NuGetVersion uses case insenstive string comparisons for pre-release components. This means that 1.0.0-alpha and 1.0.0-Alpha are equal.

So I wrote a quick C# app to verify the behaviors:

```c#
using NuGet.Versioning;

Console.WriteLine(SemanticVersion.Parse("1.0.0") == SemanticVersion.Parse("1.0.0+1")); // True
Console.WriteLine(SemanticVersion.Parse("1.0.0-alpha") == SemanticVersion.Parse("1.0.0-ALPHA")); // True
//SemanticVersion.Parse("1.0.0.1"); // ArgumentException
//SemanticVersion.Parse("1") // ArgumentException
```

Curiously `1.0.0.1` and `1` didn't parse even though the docs call them out, so I've only fixed the first two in the code for now.

fixes #4845

### DIFF
--- a/nuget/lib/dependabot/nuget/version.rb
+++ b/nuget/lib/dependabot/nuget/version.rb
@@ -10,8 +10,6 @@ require "rubygems_version_patch"
 module Dependabot
   module Nuget
     class Version < Gem::Version
-      attr_reader :build_info
-
       VERSION_PATTERN = Gem::Version::VERSION_PATTERN + '(\+[0-9a-zA-Z\-.]+)?'
       ANCHORED_VERSION_PATTERN = /\A\s*(#{VERSION_PATTERN})?\s*\z/.freeze
 
@@ -24,7 +22,7 @@ module Dependabot
       def initialize(version)
         @version_string = version.to_s
 
-        version, @build_info = version.to_s.split("+") if version.to_s.include?("+")
+        version = version.to_s.split("+").first if version.to_s.include?("+")
 
         super
       end
@@ -41,10 +39,7 @@ module Dependabot
         version_comparison = compare_release(other)
         return version_comparison unless version_comparison.zero?
 
-        version_comparison = compare_prerelease_part(other)
-        return version_comparison unless version_comparison.zero?
-
-        compare_build_info(other)
+        compare_prerelease_part(other)
       end
 
       def compare_release(other)
@@ -99,25 +94,7 @@ module Dependabot
 
         return lhs.to_i <=> rhs.to_i if lhs.match?(/^\d+$/) && rhs.match?(/^\d+$/)
 
-        lhs <=> rhs
-      end
-
-      def compare_build_info(other)
-        return build_info.nil? ? 0 : 1 unless other.is_a?(Nuget::Version)
-
-        # Build information comparison
-        lhsegments = build_info.to_s.split(".").map(&:downcase)
-        rhsegments = other.build_info.to_s.split(".").map(&:downcase)
-        limit = [lhsegments.count, rhsegments.count].min
-
-        lhs = ["1", *lhsegments.first(limit)].join(".")
-        rhs = ["1", *rhsegments.first(limit)].join(".")
-
-        local_comparison = Gem::Version.new(lhs) <=> Gem::Version.new(rhs)
-
-        return local_comparison unless local_comparison.zero?
-
-        lhsegments.count <=> rhsegments.count
+        lhs.upcase <=> rhs.upcase
       end
     end
   end
--- a/nuget/spec/dependabot/nuget/version_spec.rb
+++ b/nuget/spec/dependabot/nuget/version_spec.rb
@@ -85,7 +85,7 @@ RSpec.describe Dependabot::Nuget::Version do
 
         context "but our version has build information" do
           let(:version_string) { "1.0.0+gc.1" }
-          it { is_expected.to eq(1) }
+          it { is_expected.to eq(0) }
         end
       end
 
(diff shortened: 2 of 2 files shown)


Danielku15:
Seems you've been a bit faster than me for opening the PR 😁 I also added some additional unit tests on my branch. Maybe you want to take them over: 

https://github.com/dependabot/dependabot-core/compare/main...Danielku15:nuget-build-metadata-comparison-fix

mctofu (approved):

pinkfloydx33:
What's not clear to me (I opened #4845) and this is perhaps because my ruby isn't the greatest and I'm not up to date on dependabot internals:

While you don't compare versions using the build metadata, the build is still there in the nuget SemVer2 feed.

Suppose the current version listed in nuget is officially `1.0.0` but with additional metadata making it `1.0.0+abc`. My project file would actually reference `1.0.0`. If you're not comparing the build component and no updates have been pushed, dependabot should not change my project file. That seems to be the case here so far and would solve the bug as it exhibited (no new version pushed but my reference was updated to include the build data of the existing version) 

But suppose now a new version `1.0.1+xyz` is pushed to nuget. What is dependabot going to update my project file to? The major/minor/patch is greater so there should of course be an update. But any version number other than `1.0.1` would be incorrect (or rather extremely unexpected)

jakecoffman:
@Danielku15 Thanks! I went ahead and added a test for pre-release with build info which I think I missed. 

@pinkfloydx33 I'm new to this codebase as well, but I don't think that this project does any stripping of build info when updating dependencies. So in your example I think it will set `1.0.1+xyz` in the manifest. 

In a new project I ran `dotnet add package Serilog.Enrichers.Span --version 2.2.0+build.276` which added this to the manifest `<PackageReference Include="Serilog.Enrichers.Span" Version="2.2.0+build.276" />`. So I don't think this behavior is _incorrect_. 

If it is _unexpected_, maybe file another issue to discuss whether the build info should be stripped? If the dotnet community largely discards the build number then that would be a good feature!

jakecoffman:
@pinkfloydx33 Ah ok, thanks for pointing that out. I'll add a follow-up issue to track that work on not adding the build identifier to the manifest.

PR #4866, opened 2022-03-16
getting more info around a ci build failure

jakecoffman:


deivid-rodriguez:
Yeah, I was thinking that we could also print what `DEBUG_HELPERS=true` prints by default when failures happen.

jurre:
To reproduce locally, run the tests like this:

```
SUITE_NAME=bundler1 rspec ./spec/dependabot/bundler/file_updater_spec.rb:796
```

We should document this much better, but our suites run against both versions of bundler by having fixtures that require either bundler 1 or bundler 2, and that `SUITE_NAME` variable determines which fixtures are pulled in

deivid-rodriguez:
I think I figured it out, this patch should do the trick I believe!

```diff
diff --git a/bundler/helpers/v1/lib/functions.rb b/bundler/helpers/v1/lib/functions.rb
index fb69fdb37..dd24d1d4b 100644
--- a/bundler/helpers/v1/lib/functions.rb
+++ b/bundler/helpers/v1/lib/functions.rb
@@ -144,6 +144,9 @@ module Functions
         token.gsub("@", "%40F").gsub("?", "%3F")
       )
     end
+
+    # Use HTTPS for GitHub if lockfile
+    Bundler.settings.set_command_option("github.https", "true")
   end
 
   def self.relevant_credentials(credentials)
```

If it looks ok, feel free to add it here, or I can open a separate PR.

jurre:
The failures seem to happen because the bundler1 fixtures for those tests reference git dependencies over the `git://` protocol which GitHub no longer supports. The bundler2 fixtures reference them using `https`, which is preferred.

jakecoffman:
Hmm @deivid-rodriguez that doesn't work for me testing locally. Feel free to put up a PR, perhaps I am missing something. Thanks!

deivid-rodriguez:
I swear it passed @jurre's reproducer command line when I first tried it, but I just retried and I'm now getting a different error 😲. I'll double check tomorrow in more detail.

bundler/spec/fixtures/projects/bundler1/git_source_with_multiple_deps/Gemfile.lock
```diff
@@ -1,5 +1,5 @@
 GIT
-  remote: git://github.com/dependabot-fixtures/elasticsearch-rails.git
+  remote: https://github.com/dependabot-fixtures/elasticsearch-rails.git
```
jurre, line 2: Not sure that we should just change the lockfile, users might specify theirs this way, and if there is a way to make this work that would be better 🤔

jakecoffman:
closing in lieu of #4869

PR #4869, opened 2022-03-17
Fix Bundler 1 spec failures

deivid-rodriguez:


jurre:
@deivid-rodriguez I think this new behavior will now update the source in the lockfile from `git://` to `https://`, but I actually think this is desirable behavior, since the old source wouldn't work anymore even on someones local machine. We'd just have to update the specs to account for this. WDYT?

deivid-rodriguez:
Sure, I'd just like to figure out why Bundler 2 does not have this behavior, because if that's the desired behavior, then Bundler 2 should be updated to behave like that too, right?

jurre:
> Sure, I'd just like to figure out why Bundler 2 does not have this behavior, because if that's the desired behavior, then Bundler 2 should be updated to behave like that too, right?

I think it already does, the fixtures used for bundler 2 already use https, since that's what bundler will do by default when generating the lockfile I believe:

```
# from bundler/spec/fixtures/projects/bundler2/git_source_with_multiple_deps/Gemfile.lock
GIT
  remote: https://github.com/dependabot-fixtures/elasticsearch-rails.git
  revision: 212b37a1e927015f2a095e0f841c897d377ad1c7
  branch: 5.x
  specs:
    elasticsearch-model (5.1.0)
      activesupport (> 3)
      elasticsearch (~> 5)
      hashie
    elasticsearch-rails (5.1.0)
```

deivid-rodriguez:
Oh, you're right! I missed that fixtures are also bundler version specific. Let me update the specs then!

jurre:
Do you think we should add a test that demonstrates the lockfile gets updated when it specifies `git://` as well? Happy to tackle that separately from this PR

deivid-rodriguez:
Sounds good, either here or as a separate PR!

deivid-rodriguez:
Back to green!! 🎉

jakecoffman (approved):
Still failing locally for us but it CI is working and the change seems sound. We can continue to chase this down. Thanks!

deivid-rodriguez:
Mmmm weird, because your (dockerized) environment should be more similar to CI than mine, since I'm not yet using docker. Hopefully we'll figure it out.

Thanks!

Issue #4878, opened 2022-03-18
metadata_finder for pub should use "homepage" when "repository" is not present


jonasfj:
https://github.com/dependabot/dependabot-core/blob/416f1958f287f2d620da80a944cda692480aee12/pub/lib/dependabot/pub/metadata_finder.rb#L14

Should use `homepage` when repository is not present.

**Example PRs:**
 * https://github.com/dart-lang/pana/pull/1040 has metadata, because [`pubspec.yaml`](https://github.com/simolus3/tar/blob/7cdb563c9894600c6a739ec268f8673d6122006f/pubspec.yaml#L4) has `repository`.
 * https://github.com/dart-lang/pana/pull/1041 does not have metadata, because [`pubspec.yaml`](https://github.com/dart-lang/yaml/blob/1ee508c2f052aebff9df3c12abb9ca08aed29738/pubspec.yaml#L5) only has `homepage`.

**Context:** Historically there was only `homepage` in `pubspec.yaml`, but since some projects had separate repositories the `repository` key was introduced. Patterns on pub.dev detect if `homepage` contains a github URL and then derive a repository link from there, so most packages don't care to add `repository` if they already have `homepage` and it works fine.

-----------

The simplest option is probably to just fallback to `homepage` if `repository` is not defined. It looks as if patterns in dependabot can easily handle github URLs with directories and such:
https://github.com/dependabot/dependabot-core/blob/416f1958f287f2d620da80a944cda692480aee12/common/lib/dependabot/source.rb#L5-L10

--------------

me and @sigurdm will look into making a PR for this.

PR #4879, opened 2022-03-18
pub metadata_finder: Fall back to `homepage` if `repository` is not present

sigurdm:
Fixes: https://github.com/dependabot/dependabot-core/issues/4878


jurre:
Yep, that makes sense to me! I'd spotted some PRs without a link already and wondered if we could fix it, so appreciate this! Do you think we should add a test for it, mostly just to avoid breaking this going forward?

jurre (approved):

sigurdm:
I'll add a test.

PR #4881, opened 2022-03-18
fixed a bug and improved testing around NuGet version compare

jakecoffman:
I found the NuGet version spec too verbose and hard to read, so I tried switching to a table-driven approach. I find this to be much more readable and maintainable for this particular test.

Also while implementing this, I found another bug in the comparisons!

The bug was due to `zip` returning a shorter array in this case `[1].zip([1, 2])`. I thought it would pad out with nils as in this case `[1, 2].zip([1])` but it does not.

Let me know any Ruby tips to make the style better or if you spot a test that I might have missed.

### DIFF
--- a/nuget/spec/dependabot/nuget/version_spec.rb
+++ b/nuget/spec/dependabot/nuget/version_spec.rb
@@ -71,189 +71,39 @@ RSpec.describe Dependabot::Nuget::Version do
   end
 
   describe "#<=>" do
-    subject { version <=> other_version }
-
-    context "compared to a Gem::Version" do
-      context "that is lower" do
-        let(:other_version) { Gem::Version.new("0.9.0") }
-        it { is_expected.to eq(1) }
-      end
-
-      context "that is equal" do
-        let(:other_version) { Gem::Version.new("1.0.0") }
-        it { is_expected.to eq(0) }
-
-        context "but our version has build information" do
-          let(:version_string) { "1.0.0+gc.1" }
-          it { is_expected.to eq(0) }
-        end
-      end
-
-      context "that is greater" do
-        let(:other_version) { Gem::Version.new("1.1.0") }
-        it { is_expected.to eq(-1) }
+    sorted_versions = [
+      "",
+      "0.9.0",
+      "1.0.0-alpha",
+      "1.0.0-alpha.1",
+      "1.0.0-alpha.beta",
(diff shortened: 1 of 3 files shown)


nuget/spec/dependabot/nuget/version_spec.rb
```diff
@@ -71,189 +71,39 @@
...
+      "1.0.0-beta",
+      "1.0.0-beta.2",
+      "1.0.0-beta.11",
+      "1.0.0-beta.11.1",
+      "1.0.0-beta-extra-hyphens",
+      "1.0.0-rc.1",
+      "1.0.0",
+      "1.1.0"
+    ]
+    sorted_versions.combination(2).each do |lhs, rhs|
```
Nishnha, line 89: This is a cool approach!

Nishnha (approved):
I think you captured most of the different versioning quirks. 

Since nuget supports a [revision number](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#where-nugetversion-diverges-from-semantic-versioning) (`major.minor.patch.revision`) maybe we should also include that? I would expect a version with a revision identifier to be greater than a version with only a patch identifier.

Otherwise this looks great!

nuget/spec/dependabot/nuget/version_spec.rb
jurre: A slightly more idiomatic and performant way to write this is:

```suggestion
    sorted_versions.each_slice(2) do |lhs, rhs|
```

Performance-wise this doesn't matter at all when n = 12, but it might read slightly more familiar to rubyists. Will leave it up to you wether or not to make the change though, I don't feel strongly about this

jakecoffman: It looks like `each_slice` does something different than `combination`, in irb:

```ruby
irb(main):001:0> [1,2,3].combination(2).to_a
=> [[1, 2], [1, 3], [2, 3]]
irb(main):002:0> [1,2,3].each_slice(2).to_a
=> [[1, 2], [3]]
```

In this case I think we want to stick with `combination` so we can test every version against the others to catch mistakes like the one I made with `zip` 🙃 .

jakecoffman:
@Nishnha Great suggestion, added a few with revision numbers.

PR #4882, opened 2022-03-20
Reenable commented out failing specs

deivid-rodriguez:
The specs are still failing, but I think it's better to explicit mark them as pending than to keep them commented out.

### DIFF
--- a/bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
+++ b/bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
@@ -85,17 +85,16 @@ RSpec.describe Dependabot::Bundler::UpdateChecker::VersionResolver do
         let(:dependency_files) { bundler_project_dependency_files("subdependency") }
         its([:version]) { is_expected.to eq(Gem::Version.new("0.7.0")) }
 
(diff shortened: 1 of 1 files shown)


bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
```diff
@@ -85,17 +85,16 @@
...
-        #     expect(subject.version).to eq(Gem::Version.new("1.10.9"))
-        #   end
-        # end
+        context "that will be removed if other sub-dependencies are updated" do
+          let(:gemfile_fixture_name) { "subdependency_change" }
+          let(:lockfile_fixture_name) { "subdependency_change.lock" }
+          let(:dependency_name) { "nokogiri" }
+          let(:requirements) { [] }
+
+          pending "is updated" do
```
deivid-rodriguez, line 94: Well, I didn't reenable anything since the specs are still pending 🤣, let me see if they actually pass now.

deivid-rodriguez: Alright, specs are still failing but I think it still has some value to actually run them but mark them as pending, so I will keep this PR with that goal.

jurre (approved):

jurre:
Thanks!

PR #4884, opened 2022-03-21
Bump bundler from 2.3.8 to 2.3.9 and automate it

jurre:
This bumps the bundler version to the latest release, and adds a quick
'n dirty script to do it for us going forward.

This solution isn't ideal as places where we reference the version might
change, but I think it's an improvement over manually going through the
files. Also, if we happen to reference the same value as the bundler version
to mean something else in the same file, it'll get updated.

### DIFF
--- a/bundler/lib/dependabot/bundler/helpers.rb
+++ b/bundler/lib/dependabot/bundler/helpers.rb
@@ -4,7 +4,7 @@ module Dependabot
   module Bundler
     module Helpers
       V1 = "1.17.3"
-      V2 = "2.3.8"
+      V2 = "2.3.9"
       # If we are updating a project with no Gemfile.lock, we default to the
       # newest version we support
       DEFAULT = V2
--- a/bundler/spec/dependabot/bundler/helper_spec.rb
+++ b/bundler/spec/dependabot/bundler/helper_spec.rb
@@ -41,7 +41,7 @@ RSpec.describe Dependabot::Bundler::Helpers do
   end
 
   let(:v1) { "1.17.3" }
-  let(:v2) { "2.3.8" }
+  let(:v2) { "2.3.9" }
 
   describe "#bundler_version" do
     def described_method(lockfile)
--- a/Dockerfile
+++ b/Dockerfile
@@ -76,7 +76,7 @@ RUN apt-add-repository ppa:brightbox/ruby-ng \
   && apt-get install -y --no-install-recommends ruby2.7 ruby2.7-dev \
   && gem update --system 3.2.20 \
   && gem install bundler -v 1.17.3 --no-document \
-  && gem install bundler -v 2.3.8 --no-document \
+  && gem install bundler -v 2.3.9 --no-document \
   && rm -rf /var/lib/gems/2.7.0/cache/* \
   && rm -rf /var/lib/apt/lists/*
 
--- a/bin/update-bundler.rb
+++ b/bin/update-bundler.rb
@@ -1,0 +1,26 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+# This script bumps the bundler version used, since we reference it in a few
+# different places.
+
+require "excon"
+require "json"
+
+LATEST_VERSION = JSON.parse(Excon.get("https://rubygems.org/api/v1/gems/bundler.json").body)["version"]
+CURRENT_VERSION = File.read("Dockerfile").match(/gem install bundler -v (2.\d+\.\d+)/)[1]
+
+def update_file(filename)
+  File.open(filename, "r+") do |f|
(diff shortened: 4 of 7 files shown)


bundler/lib/dependabot/bundler/helpers.rb
```diff
@@ -4,7 +4,7 @@ module Dependabot
   module Bundler
     module Helpers
       V1 = "1.17.3"
-      V2 = "2.3.8"
+      V2 = "2.3.9"
```
deivid-rodriguez, line 7: Then Bundler version selection behavior implemented in this file should be the default as long as there's single v1 and v2 versions installed in the running environment. If that's the case I think it should be possible to completely remove this logic (don't pass `BUNDLER_VERSION` at all to native helpers) and let RubyGems do the right thing. But it seems something to consider for the future, definitely not for this PR.

bundler/spec/dependabot/bundler/helper_spec.rb
```diff
@@ -41,7 +41,7 @@
   end
 
   let(:v1) { "1.17.3" }
-  let(:v2) { "2.3.8" }
+  let(:v2) { "2.3.9" }
```
deivid-rodriguez, line 44: I think this spec could be relaxed to not hardcode exact versions, but instead just check the major version selected.

deivid-rodriguez (commented):
Looks good, and unlikely to have false positives I believe. I think it's only missing updates in `bundler/script/ci-test`.

I left a couple of comments with suggestions on how to reduce the number of places that need updates in the future, just something to consider in future improvements.

bundler/lib/dependabot/bundler/helpers.rb
jurre: Aye, yeah we should definitely try that and see if it doesn't break anything!

deivid-rodriguez:
Looks good to me!

bundler/spec/dependabot/bundler/helper_spec.rb
jurre: I'm not sure, right now I think the exact version is relevant, if we do [this](https://github.com/dependabot/dependabot-core/pull/4884#discussion_r830971299) then we can maybe relax it?

deivid-rodriguez: These specs are unit tests for the `Bundler::Helpers.bundler_version` method, which selects an exact bundler version according to the lockfile contents, using the following criteria:

* No lockfile -> The exact v2 version dependabot has available.
* Lockfile with no BUNDLED WITH section -> The exact v2 version dependabot has available.
* Lockfile with BUNDLED WITH 1.x section -> The exact v1 version dependabot has available.
* Lockfie with BUNDLED WITH 2.x section or higher -> The exact v2 version dependabot has available.

Namely, a unit test for this method

https://github.com/dependabot/dependabot-core/blob/ea8fbaf64af5af477d5b69d0b985782340ca3b36/bundler/lib/dependabot/bundler/helpers.rb#L18-L26

In my opinion, testing that the result is a valid 3 component version number, and that the first segment is correct (1.x or 2.x) is enough. The exact version itself that dependabot is using is "configuration" and should not need a test, because it's not useful for detecting bugs and it means essentially testing the specific value of a constant, which defeats one of the points of using constants (keep things DRY).

deivid-rodriguez: That's said that just my personal opinion but it's no big deal, and indeed if https://github.com/dependabot/dependabot-core/pull/4884#discussion_r830971299 works, all this code will go away, so these tests will go too anways.

landongrindheim (approved):
Thanks for checking in the automation! Looks good to me 🚀 

Outside the scope of this change: Do you think we could define the Bundler version in one place as a constant and reference it throughout? I'm thinking we could have Dependabot keep it up-to-date   😄

jurre:
> Thanks for checking in the automation! Looks good to me 🚀 
> 
> Outside the scope of this change: Do you think we could define the Bundler version in one place as a constant and reference it throughout? I'm thinking we could have Dependabot keep it up-to-date   😄

I think it would be tricky, because we reference it in the Dockerfile, in Ruby source and bash scripts, the only things I can think of are real hacks 🤔

landongrindheim:
> > Thanks for checking in the automation! Looks good to me 🚀
> > Outside the scope of this change: Do you think we could define the Bundler version in one place as a constant and reference it throughout? I'm thinking we could have Dependabot keep it up-to-date   😄
> 
> I think it would be tricky, because we reference it in the Dockerfile, in Ruby source and bash scripts, the only things I can think of are real hacks 🤔

Same here. If I think of something I'll ping you 😄

PR #4886, opened 2022-03-21
Add Ruby RBS Type Signatures 

mattt:
This PR explores the use of [RBS](https://github.com/ruby/rbs) and [Steep](https://github.com/soutaro/steep) to improve our Ruby source code.

Run the following commands in this branch to try it out:

```terminal
$ cd common
$ bundle install
$ bundle exec steep check
```

### DIFF
--- a/common/lib/dependabot/file_fetchers/base.rb
+++ b/common/lib/dependabot/file_fetchers/base.rb
@@ -25,10 +25,15 @@ module Dependabot
         Dependabot::Clients::CodeCommit::NotFound
       ].freeze
 
+      # Checks an array of filenames (string) and returns a boolean
+      # describing whether the language-specific dependency files
+      # required for an update run are present.
       def self.required_files_in?(_filename_array)
         raise NotImplementedError
       end
 
+      # Returns a static error message which can be displayed to a user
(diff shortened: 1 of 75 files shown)


mattt:
After doing the work to implement type signatures manually for a small ecosystem, I've concluded that it's too much effort. If we could get RBS / Steep to infer types automatically (perhaps by co-mingling code that calls ecosystem plugins directly, like in [dependabot-script](https://github.com/dependabot/dependabot-script)), then it might be worth taking another look.

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -61,10 +66,17 @@ def target_branch
...
+      # the repo this instance was created with.
+      # Returns an array of `Dependabot::DependencyFile` instances.
       def files
         @files ||= fetch_files
       end
 
+      # Returns the commit SHA-1 hash at the time
+      # the dependency files were fetched.
+      # If called before files, the returned value
+      # will be used in subsequent calls to files.
```
jeffwidman, line 79: @mattt Do you want to pull these docstrings out into a separate PR and add them? They look useful...

PR #4889, opened 2022-03-21
Npm: Handle `.tar.gz` extensions for path dependencies

jurre:
Dependabot has support for loading file dependencies for npm, and when
these files are larger than 1mb, we have some logic in place to retry
using the Git Data API.

We ran into a case where this would surface a `Blob too large` error,
indicating that the retry via the Git Data API wasn't working as expected,
and it turns out it's because we did not support the `.tar.gz` extension
in our code.

Previously this would cause us to assume this was a nested package.json,
(see [here][npm path dependency ref]) so we'd tack `package.json` onto the
filename, which in turn would cause a 404, which in turn would cause us to
retry without `package.json` attached, only this time using
`_github_repo_contents` instead of `_fetch_file_content_from_github`,
which does not have logic to retry using the Git Data API.

[npm path dependency ref]: https://github.com/dependabot/dependabot-core/blob/ea8fbaf64af5af477d5b69d0b985782340ca3b36/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb#L124-L128

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
@@ -123,7 +123,7 @@ module Dependabot
           filename = path
           # NPM/Yarn support loading path dependencies from tarballs:
           # https://docs.npmjs.com/cli/pack.html
-          filename = File.join(filename, "package.json") unless filename.end_with?(".tgz", ".tar")
+          filename = File.join(filename, "package.json") unless filename.end_with?(".tgz", ".tar", ".tar.gz")
           cleaned_name = Pathname.new(filename).cleanpath.to_path
           next if fetched_files.map(&:name).include?(cleaned_name)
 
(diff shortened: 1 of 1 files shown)


npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -123,7 +123,7 @@ def path_dependencies(fetched_files)
           filename = path
           # NPM/Yarn support loading path dependencies from tarballs:
           # https://docs.npmjs.com/cli/pack.html
-          filename = File.join(filename, "package.json") unless filename.end_with?(".tgz", ".tar")
+          filename = File.join(filename, "package.json") unless filename.end_with?(".tgz", ".tar", ".tar.gz")
```
jurre, line 126: Maybe this could be a constant

mctofu (approved):
Makes sense! I like the constant idea so we don't have to manage the list in both places but it also doesn't seem like it will change very frequently.

jurre:
> Makes sense! I like the constant idea so we don't have to manage the list in both places but it also doesn't seem like it will change very frequently.

Yeah I think in this case maybe being able to see the values in the place where they are used is more valuable

Issue #3165, opened 2021-02-18
Dependabot doesn't properly handle python versions with epoch


daviddavis:
In python, if you have a version requirement such as `pulp-smash>=1!0.12.0`, dependabot will throw this error: 

```
Illformed requirement [">=1!0.12.0"]
```

This is a perfectly valid version though according to [PEP440](https://www.python.org/dev/peps/pep-0440/). The `1!` being the epoch segment.

kohder:
We're running into this too because we use Astronomer's Airflow packages. Example for the GitHub provider package: https://pip.astronomer.io/simple/apache-airflow-providers-github/

PR #4891, opened 2022-03-22
Add support for Python packages with epoch version

kohder:
Dependabot currently considers a Python package version with an epoch to be illformed. This change adds support for parsing them and using epochs in version comparisons. When an epoch is not specified, the epoch is treated as 0.

Epoch version definition: https://peps.python.org/pep-0440/#version-epochs
Example of packages with epoch in version: https://pip.astronomer.io/simple/apache-airflow/


kohder:
Related: https://github.com/dependabot/dependabot-core/issues/3165

jakecoffman:
Moved changes to #4928

PR #4894, opened 2022-03-23
Add raise_on_ignore support to Pub

mctofu:
When running on GitHub, Dependabot sets `raise_on_ignore` to true in order to determine when a version update is not possible because the user has ignored all possible updates. It's expected that calling the updater's `latest_version` method will raise `AllVersionsIgnored` when that is the case.

cc: @sigurdm, @jonasfj

### DIFF
--- a/pub/lib/dependabot/pub/update_checker.rb
+++ b/pub/lib/dependabot/pub/update_checker.rb
@@ -10,7 +10,13 @@ module Dependabot
       include Dependabot::Pub::Helpers
 
       def latest_version
-        Dependabot::Pub::Version.new(current_report["latest"])
+        version = Dependabot::Pub::Version.new(current_report["latest"])
+
+        return version if ignore_requirements.none? { |r| r.satisfied_by?(version) }
+        return version if version == version_class.new(dependency.version)
+        return nil unless @raise_on_ignored
+
+        raise AllVersionsIgnored
       end
 
       def latest_resolvable_version_with_no_unlock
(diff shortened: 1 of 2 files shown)


jurre (approved):

jonasfj:
We shouldn't also do this in `latest_resolvable_version` and `latest_resolvable_version_with_no_unlock`.

Both of these can return `nil` because something was ignored.

---------------------------------

Also `latest_version` is generally the latest version of the package, regardless of whether or not said version is mutually compatible with the other dependencies.

mctofu:
> We shouldn't also do this in `latest_resolvable_version` and `latest_resolvable_version_with_no_unlock`? Both of these can return `nil` because something was ignored.
> 
> I'll admit I don't fully understand the interface required by `Dependabot::UpdateCheckers::Base`.

Currently it's only expected from `latest_version` but I think that may be an oversight since you're right, it's possible you haven't ignored the `latest_version` but you have ignored the `latest_resolvable_version`. It's also possible that you ignore the `latest_version` but not the `latest_resolvable_version` and won't get an update since `latest_version` doesn't currently consider all available versions. If that becomes a common issue we could perhaps work around it by also considering `latest_resolvable_version` if the `latest_version` is ignored.

mctofu:
I tested this out and this is the log output now when ignore conditions prevent an update:
```
updater | INFO <job_325709714> Checking if collection 1.14.13 needs updating
updater | INFO <job_325709714> Ignored versions:
updater | INFO <job_325709714>   version-update:semver-minor - from .github/dependabot.yml
  proxy | 2022/03/24 20:36:40 [015] GET https://pub.dartlang.org:443/api/packages/collection
  proxy | 2022/03/24 20:36:40 [015] 200 https://pub.dartlang.org:443/api/packages/collection
updater | INFO <job_325709714> All updates for collection were ignored
```

PR #4905, opened 2022-03-24
Automate Terraform Platform Detection for Lockfile Hashes

Nishnha:
Terraform supports a few different architectures and, depending on the one(s) being used in the project, will save a hash for each in the `.terraform.lock.hcl` lockfile.  

Previously, Dependabot would use the default architecture `linux_amd64` for every Terraform update, which caused issues for users who were not using that platform (see https://github.com/dependabot/dependabot-core/pull/4784).

This PR automates Terraform architecture detection by comparing the existing lockfile hashes against the possible platforms. Once the architecture(s) are determined, Dependabot will update the hashes for the actual architectures in that project to the new version.

The architectures that we check are (in order):
- linux_amd64
- darwin_amd64
- windows_amd64
- darwin_arm64
- linux_arm64


pavera (approved):
I tried to think through any way to further optimize this, but I think this is as good as it gets with the tools TF currently provides.

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -94,25 +94,110 @@ def update_registry_declaration(new_req, old_req, updated_content)
...
+
+      def lockfile_details(new_req)
+        content = lock_file.content.dup
+        provider_source = new_req[:source][:registry_hostname] + "/" + new_req[:source][:module_identifier]
+        declaration_regex = lockfile_declaration_regex(provider_source)
+
+        [content, provider_source, declaration_regex]
+      end
+
+      def lookup_hash_architecture # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, Metrics/PerceivedComplexity
```
jurre, line 112: Whew that's a big one 😄 are there maybe any methods that we can extract out of here to make this a little easier to grok? Happy to take a stab at it (tomorrow) also

jurre (approved):
Overall the change looks good to me, I think it might be worth trying to extract some logic from `#lookup_hash_architecture`, rubocop also complains about this and we need to disable 3 linting rules in order to satisfy it, but don't want to block on it. We can either tackle it in this PR or follow up later, lmk if I can help with this!

Nishnha:
> Overall the change looks good to me, I think it might be worth trying to extract some logic from `#lookup_hash_architecture`, rubocop also complains about this and we need to disable 3 linting rules in order to satisfy it, but don't want to block on it. We can either tackle it in this PR or follow up later, lmk if I can help with this!

Thanks for the review! 🎉 

I think it would be best for the community to get this fix merged in since it seems better than the status quo described in https://github.com/dependabot/dependabot-core/pull/4784

I decided to open https://github.com/dependabot/dependabot-core/pull/5013 as a refactor pass on this PR. Maybe we can pair up on it together sometime this week?

Nishnha:
This PR was shipped in https://github.com/dependabot/dependabot-core/pull/5014

billinghamj:
This has instantly removed the need for the action, and made it immediately do exactly the right thing for virtually everyone using Terraform - thank you for getting this sorted :)

billinghamj:
Just to gently flag, it seems like this has stopped working... not quite sure why

Screenshot 2022-05-03 at 19 07 36

Nishnha:
We had a brief period where we had to roll back this PR.
Commenting `@dependabot recreate` on a Terraform PR that showed the old behavior and will recreate it with the updated behavior.

Sorry about that!

Flydiverny:
Thank you very much for this! It enabled us to finally start using dependabot for our terraform dependencies ❤️ 
We are in some cases still seeing PRs removing hashes, I've opened a related issue here https://github.com/dependabot/dependabot-core/issues/5225 with a reproduction repo, recreating does not solve it :(

PR #4912, opened 2022-03-26
Added Ruby 3.1.1 to RubyRequirementSetter

schinery:


jeffwidman:
cc @deivid-rodriguez

deivid-rodriguez (commented):
Looks good to me!

jurre (approved):

schinery:
@jurre I've just force pushed to add the missing Bundler 1 Ruby 3.1 fixtures which caused their specs to fail. Sorry for missing these originally.

jurre:
Thanks @schinery, I will, but it seems like Actions is having a bad time so it's not quite working right now :/

schinery:
Sorry @jurre hopefully third time lucky 🤞🏻. I ran the tests after the last "fix" locally but hadn't set the `SUITE_NAME` env var that gets set when the CI runs so the `bundler1` version didn't run and I missed my copy + paste error. 

I've resolved this and pushed the new fix up. Below are the tests run as both `bundler1` and `bunder2` just to prove I've actually tested it properly this time.

dependabot-bundler-specs
y

PR #4913, opened 2022-03-26
Add defensive checks to prevent inscrutable error messages

jeffwidman:
Part of the reason
https://github.com/dependabot/dependabot-core/issues/4625 was so
difficult to debug was the original error messages were a total red
herring. They appeared to be related to IPs, but digging deeper realized
they were actually due to errors happening earlier in the processing
pipeline but not caught until several steps down the line.

So this adds some defensive checks so that if there's an upstream error,
we catch it sooner and report the proper error message(s).

### DIFF
--- a/go_modules/lib/dependabot/go_modules/resolvability_errors.rb
+++ b/go_modules/lib/dependabot/go_modules/resolvability_errors.rb
@@ -12,7 +12,19 @@ module Dependabot
 
       sig { params(message: String, goprivate: T.untyped).void }
       def self.handle(message, goprivate:)
+        # TODO: currently this matches last. Instead, if more than one match, and they
+        # aren't identical, then don't try to be clever with GitDependenciesNotReachable
+        # but instead raise DependencyFileNotResolvable and report the whole error.
+        # This would have resulted in a more obvious error message for #4625
         mod_path = message.scan(GITHUB_REPO_REGEX).last
+        if mod_path
+          # TODO: if mod_path doesn't look like a URL, don't continue, but instead raise
+          # DependencyFileNotResolvable and report the whole error.
+          # This would have resulted in a more obvious error message for #4625
+          # How to implement this though?
+          # * Ruby has no built-in URL parsing, and no great alternatives in https://stackoverflow.com/q/1805761/770425...
+          # Not sure what Dependabot team policy is on using 3rd-party gems?
+          # Alternatively a basic sanity check of "it should not contain whitespace" may suffice for now... ??
         unless mod_path && message.include?("If this is a private repository")
           raise Dependabot::DependencyFileNotResolvable, message
         end


go_modules/lib/dependabot/go_modules/resolvability_errors.rb
```diff
@@ -6,7 +6,19 @@ module ResolvabilityErrors
...
+        # This would have resulted in a more obvious error message for #4625
         mod_path = message.scan(GITHUB_REPO_REGEX).last
+        if mod_path
+        # TODO: if mod_path doesn't look like a URL, don't continue, but instead raise
+        # DependencyFileNotResolvable and report the whole error.
+        # This would have resulted in a more obvious error message for #4625
+        # How to implement this though?
+        # * Ruby has no built-in URL parsing, and no great alternatives in https://stackoverflow.com/q/1805761/770425...
+        # Not sure what Dependabot team policy is on using 3rd-party gems?
+        # Alternatively a basic sanity check of "it should not contain whitespace" may suffice for now... ??
```
jeffwidman, line 21: @jurre @mctofu any feedback on implementing this? Maybe there's a gem that's already being used elsewhere in another check that could be leveraged here? Otherwise, given that this is a simple defensive check, a regex match on any whitespace (especially spaces and newlines) is probably sufficient.

go_modules/lib/dependabot/go_modules/resolvability_errors.rb
```diff
@@ -6,7 +6,19 @@ module ResolvabilityErrors
       GITHUB_REPO_REGEX = %r{github.com/[^:@]*}.freeze
 
       def self.handle(message, credentials:, goprivate:)
+        # TODO: currently this matches last. Instead, if more than one match, and they
+        # aren't identical, then don't try to be clever with GitDependenciesNotReachable
+        # but instead raise DependencyFileNotResolvable and report the whole error.
+        # This would have resulted in a more obvious error message for #4625
```
jeffwidman, line 12: I actually wasn't 100% sure on this... I could see a case where a multi-line error string may have multiple URLs, and the last one is indeed the proper one to match on... But given that `GitDependenciesNotReachable` tells the user that Dependabot can't reach the repo and to update their access creds, that's a fairly opinionated solution... IMO I'd only want to report that if 110% sure that's the issue, otherwise surface the whole error message (I assume that's what `DependencyFileNotResolvable` does... if it just hides it in logs that only Dependabot devs can see, then that's not so helpful either.)

mctofu: I see these cases as most useful to raise `GitDependenciesNotReachable`.

1. You have a direct dependency on a private github repo but Dependabot doesn't have access to it.
2. You have an indirect dependency on a github private repo (from another direct private github dependency) but Dependabot doesn't have access to it.

Do we have examples of what multiple URLs look like in this error message handy? I imagine this could get tricky if you had a failing indirect dependency on another host that came from one of your direct github dependencies.

jeffwidman: Here's an example of multiple URLs in the error message:
```
"go: downloading github.com/<redacted>/<private-repo-that-we-are-directly-running-dependabot-against> v0.39.0\ngo get: reading google.golang.org/grpc/go.mod at revision v1.33.0: unknown revision v1.33.0\n"
```

jeffwidman:
I'm not planning to spend more time on this anytime soon because the value isn't there for the time it'll require... Someone else is welcome to pick this up and run with it.

jeffwidman:
Re-opening as AI makes tackling this much cheaper... not urgent in the least, just probably sometime down the line I'll handle it.

jeffwidman:
I had Copilot spin on this, and there's enough risk fo false positives/negatives that I'm hesitant to proceed without investigating more deeply.

This doesn't happen frequently enough to spend more cycles on it at this point.

PR #4916, opened 2022-03-27
Metadetafinders: Use the default branch instead of master for gitlab

### LINKED ISSUE #4897: Gitlab metadata does not respect other branches than master
Heey, 

I noticed that in `Dependabot::MetadataFinders::Base::ChangelogFinder` the branch for gitlab is hard-coded to `master`.
See: https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/metadata_finders/base/changelog_finder.rb#L284 though as far as I can see we can use `source.branch`

For [Bitbucket](https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/metadata_finders/base/changelog_finder.rb#L263) this seems to be correct, even though we don't use `source.branch` directly

dennisvandehoef:
This pull request Fixes #4897 by replacing the hard-coded `master` with the default branch of the repository

### DIFF
--- a/common/lib/dependabot/metadata_finders/base/changelog_finder.rb
+++ b/common/lib/dependabot/metadata_finders/base/changelog_finder.rb
@@ -271,6 +271,7 @@ module Dependabot
         end
 
         def fetch_gitlab_file_list
+          branch = default_gitlab_branch
           gitlab_client.repo_tree(source.repo).map do |file|
             type = case file.type
                    when "blob" then "file"
@@ -281,8 +282,8 @@ module Dependabot
               name: file.name,
               type: type,
               size: 100, # GitLab doesn't return file size
-              html_url: "#{source.url}/blob/master/#{file.path}",
-              download_url: "#{source.url}/raw/master/#{file.path}"
+              html_url: "#{source.url}/blob/#{branch}/#{file.path}",
+              download_url: "#{source.url}/raw/#{branch}/#{file.path}"
             )
           end
         rescue Gitlab::Error::NotFound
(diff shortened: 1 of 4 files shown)


common/lib/dependabot/metadata_finders/base/changelog_finder.rb
```diff
@@ -271,6 +271,7 @@ def fetch_bitbucket_file_list
         end
 
         def fetch_gitlab_file_list
+          branch = default_gitlab_branch
```
jurre, line 274: Not really important, but since `default_gitlab_branch` is already memoized we don't really need to assign it here and could use `default_gitlab_branch` directly below? 🤷

jurre (approved):

common/lib/dependabot/metadata_finders/base/changelog_finder.rb
dennisvandehoef: You are absolutely right. The reason I did it like this is so that it is the same as in the [fetch_bitbucket_file_list](https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/metadata_finders/base/changelog_finder.rb#L252)-method. I would be happy to change this as suggested, please let me know 😄

dennisvandehoef:
Heey,

The 2 checks that failed, seem to have failed due to API limits not related to my change. Though sadly, I don't have the needed access to re-start them. Can someone with access, please restart them, if that is blocking this from getting merged?

common/lib/dependabot/metadata_finders/base/changelog_finder.rb
jurre: It's fine :D

PR #4928, opened 2022-03-29
Add support for Python packages with epoch version

### LINKED ISSUE #3165: Dependabot doesn't properly handle python versions with epoch

jakecoffman:
@kohder thanks for PR #4891!

While I was reviewing that PR I noticed we would also need to add epoch support to the `requirement_parser`, so I went ahead and did that in this branch on top of your changes. I also made a few simplifications to the code and brought in more thorough tests like those in the nuget spec.

Let me know what you think, and I will also get another @dependabot/reviewers to take a look at these changes as well!

Fixes #3165

### DIFF
--- a/python/lib/dependabot/python/requirement_parser.rb
+++ b/python/lib/dependabot/python/requirement_parser.rb
@@ -6,7 +6,7 @@ module Dependabot
       NAME = /[a-zA-Z0-9](?:[a-zA-Z0-9\-_\.]*[a-zA-Z0-9])?/.freeze
       EXTRA = /[a-zA-Z0-9\-_\.]+/.freeze
       COMPARISON = /===|==|>=|<=|<|>|~=|!=/.freeze
-      VERSION = /[0-9]+[a-zA-Z0-9\-_\.*]*(\+[0-9a-zA-Z]+(\.[0-9a-zA-Z]+)*)?/.
+      VERSION = /([1-9][0-9]*!)?[0-9]+[a-zA-Z0-9\-_.*]*(\+[0-9a-zA-Z]+(\.[0-9a-zA-Z]+)*)?/.
                 freeze
       REQUIREMENT =
         /(?<comparison>#{COMPARISON})\s*\\?\s*(?<version>#{VERSION})/.freeze
--- a/python/lib/dependabot/python/version.rb
+++ b/python/lib/dependabot/python/version.rb
@@ -68,7 +68,7 @@ module Dependabot
       private
 
       def epoch_comparison(other)
-        epoch.to_i - other.epoch.to_i
+        epoch.to_i <=> other.epoch.to_i
       end
 
       def post_version_comparison(other)
@@ -78,7 +78,7 @@ module Dependabot
 
         return -1 if post_release_version.nil?
 
-        post_release_version.to_i - other.post_release_version.to_i
+        post_release_version.to_i <=> other.post_release_version.to_i
       end
 
       def local_version_comparison(other)
--- a/python/lib/dependabot/python/version.rb
+++ b/python/lib/dependabot/python/version.rb
@@ -14,7 +14,6 @@ module Dependabot
       attr_reader :local_version
       attr_reader :post_release_version
 
-
       # See https://peps.python.org/pep-0440/#appendix-b-parsing-version-strings-with-regular-expressions
       VERSION_PATTERN = 'v?([1-9][0-9]*!)?[0-9]+[0-9a-zA-Z]*(?>\.[0-9a-zA-Z]+)*' \
                         '(-[0-9A-Za-z-]+(\.[0-9a-zA-Z-]+)*)?' \
--- a/python/lib/dependabot/python/version.rb
+++ b/python/lib/dependabot/python/version.rb
@@ -14,7 +14,7 @@ module Dependabot
       attr_reader :local_version
       attr_reader :post_release_version
 
-      VERSION_PATTERN = 'v?([0-9]+!)?[0-9]+[0-9a-zA-Z]*(?>\.[0-9a-zA-Z]+)*' \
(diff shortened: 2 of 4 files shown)


python/spec/dependabot/python/version_spec.rb
```diff
@@ -70,126 +75,46 @@
...
+      "",
+      "0.9",
+      "1.0.0-alpha",
+      "1.0.0-a.1",
+      "1.0.0-beta",
+      "1.0.0-b.2",
+      "1.0.0-beta.11",
+      "1.0.0-rc.1",
+      "1",
+      # "1.0.0.post", TODO fails comparing to 1
```
jakecoffman, line 88: The more thorough testing is showing here that dependabot thinks `1 > 1.0.0.post`, but yet `1 < 1.0.0.post1` so I think there's a bug here. 

According to [PEP 440](https://peps.python.org/pep-0440/#post-releases), post release identifiers are for non-material changes like fixing release notes. So I assume if we fix that then dependabot will be bumping a lot of major versions to post release versions.

But I want to look into this more after we fix epoch support so I'll make a followup issue.

python/spec/dependabot/python/version_spec.rb
```diff
@@ -70,126 +75,46 @@
...
+      "1.post2+gc1.11",
+      "1.0.0.post11",
+      "1.0.2",
+      "1.0.11",
+      "2016.1",
+      "1!0.1.0",
+      "2!0.1.0",
+      "10!0.1.0"
+    ]
+    sorted_versions.combination(2).each do |lhs, rhs|
```
kohder, line 102: I'm jealous. I've never had an occasion to use `combination`.

python/lib/dependabot/python/version.rb
```diff
@@ -14,7 +14,7 @@ class Version < Gem::Version
       attr_reader :local_version
       attr_reader :post_release_version
 
-      VERSION_PATTERN = 'v?([0-9]+!)?[0-9]+[0-9a-zA-Z]*(?>\.[0-9a-zA-Z]+)*' \
+      VERSION_PATTERN = 'v?([1-9][0-9]*!)?[0-9]+[0-9a-zA-Z]*(?>\.[0-9a-zA-Z]+)*' \
```
kohder, line 17: I guess this makes sense because there wouldn't be a version like `0!0.0.1`.
Probably doesn't matter but we could also make the new group a non-capturing group like the other one by adding `?>`.

kohder (commented):
Nice cleanup! Love the test simplifications.

python/lib/dependabot/python/version.rb
jakecoffman: Oh I forgot to mention this! I grabbed the epoch portion of the [official regex](https://peps.python.org/pep-0440/#appendix-b-parsing-version-strings-with-regular-expressions). This might be a little too pedantic, having a leading 0 wouldn't matter, maybe I should switch it back to be more forgiving 🤔

kohder: Ah! I didn't know that was available. Personally, I think what you have is fine.

python/lib/dependabot/python/version.rb
```diff
@@ -63,39 +65,23 @@ def <=>(other)
...
         end
 
         return -1 if post_release_version.nil?
 
-        # Post release versions should only ever be a single number, so we can
-        # just string-comparison them.
-        return 0 if post_release_version.to_i == other.post_release_version.to_i
-
-        post_release_version.to_i > other.post_release_version.to_i ? 1 : -1
+        post_release_version.to_i - other.post_release_version.to_i
```
jurre, line 81: Can you elaborate on this change a little bit? It seems to work from the tests passing but I don't quite understand how it will always result in the same outcome 🙈

jakecoffman: Ah I was treating it like other languages' comparison functions, but I see now that Ruby strictly defines `<=>` to only return -1, 0, or 1. I'll switch this to use `<=>` which looks idiomatic.

python/lib/dependabot/python/version.rb
jakecoffman: Yeah we should definitely adopt the official regexes. 

I started pulling in the Python regex with capture groups so we can get away from all the splitting and just use the groups, but saw it was going to change current behavior. I thought it would be best to fix epochs on the first pass.

jurre (approved):
Looks good to me, awesome work on this @jakecoffman

jakecoffman:
Thanks to @kohder for the initial fix!

python/lib/dependabot/python/requirement_parser.rb
```diff
@@ -6,7 +6,7 @@ class RequirementParser
       NAME = /[a-zA-Z0-9](?:[a-zA-Z0-9\-_\.]*[a-zA-Z0-9])?/.freeze
       EXTRA = /[a-zA-Z0-9\-_\.]+/.freeze
       COMPARISON = /===|==|>=|<=|<|>|~=|!=/.freeze
-      VERSION = /[0-9]+[a-zA-Z0-9\-_\.*]*(\+[0-9a-zA-Z]+(\.[0-9a-zA-Z]+)*)?/.
+      VERSION = /([1-9][0-9]*!)?[0-9]+[a-zA-Z0-9\-_.*]*(\+[0-9a-zA-Z]+(\.[0-9a-zA-Z]+)*)?/.
```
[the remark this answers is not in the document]
Nishnha: Ah I see I wasn't the only one who noticed this. Thanks for the explanation, makes sense why you chose the regex match you did.

Just curious, what happens if there is an epoch that starts with 0! (because someone decided not to read the PEP 😛) Hopefully we raise an error?

PR #4950, opened 2022-04-04
v0.180.3

jakecoffman:
https://github.com/dependabot/dependabot-core/compare/v0.180.2...498d850

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,21 @@
+## v0.180.3, 4 April 2022
+
+- Add support for Python packages with epoch version [#4928](https://github.com/dependabot/dependabot-core/pull/4928)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,21 @@
+## v0.180.3, 4 April 2022
+
+- Add support for Python packages with epoch version [#4928](https://github.com/dependabot/dependabot-core/pull/4928)
```
jurre, line 3: We could consider this a new feature and do a `minor` bump, but we could also consider it a bugfix 🤷

jurre (approved):

CHANGELOG.md
jakecoffman: Yeah I was leaning towards patch but it's definitely arguable 🤷

PR #4952, opened 2022-04-04
replace lint workflow

willianpaixao:
The current `bin/lint` seemed to not lint at all but instead check for shellcheck vulnerabilities in a set of files.

This PR removes that file and its reference in the `.github/workflows/ci.yml`, then replacing with the official supported [github/super-linter](https://github.com/github/super-linter).
This action is not a linter per se, but an umbrella that [calls different linters](https://github.com/github/super-linter#supported-linters) according to the language within each file.

Also note that the flag `VALIDATE_ALL_CODEBASE = false` means that only new and changed code will be scanned. This is by design due to the huge existing codebase in this repo.


jurre:
Linting happens on a per-suite basis, we run `script/ci-test` for each ecosystem and it will lint the files for that test suite: https://github.com/dependabot/dependabot-core/blob/main/bundler/script/ci-test

This uses rubocop for linting which is the preferred linter for our ruby code. I think we'll want to keep this per-suite behavior because it gives us faster feedback on where any violating code lives, and we get to parallelize the work.

jurre:
I'm going to close this out for now, as I don't think this change is going to do what was intended, but I'm definitely open to improving our linting setup, feel free to open up an issue to discuss it and tag me in it.

Issue #4768, opened 2022-02-26
GitHub Actions - full patch versions are bumped to minimal new major versions


adamralph:
**Package ecosystem**
GitHub Actions
**Package manager version**
n/a
**Language version**
n/a
**Manifest location and content prior to update**
https://github.com/adamralph/bullseye/blob/bbf7aef61a19ab8a78af7d11f36aabc4d5d0acf4/.github/workflows/ci.yml#L28
**dependabot.yml content**
https://github.com/adamralph/bullseye/blob/bbf7aef61a19ab8a78af7d11f36aabc4d5d0acf4/.github/dependabot.yml
**Updated dependency**
actions/setup-dotnet, from `1.9.1` to `2`
**What you expected to see, versus what you actually saw**
I expected to see the dependency update from `1.9.1` to `2.0.0`. The convention with GitHub Actions is to continually move tags like `1` and `2` to match the latest patch version, e.g. `1.9.1`, `2.3.4` etc. I am deliberately _not_ using tags like `1` and `2`. I am using the full patch version, e.g. `1.9.1` for better build reproducibility. If the current version is a full version, e.g. `1.9.1` then dependabot should not update that version to `2`. It should update it to `2.0.0`.
**Native package manager behavior**
n/a
**Images of the diff or a link to the PR, issue or logs**
![image](https://user-images.githubusercontent.com/677704/155833820-a89d80e2-d8c9-4a51-af9d-2c2fc101cef5.png)
**🕹 Bonus points: Smallest manifest that reproduces the issue**

xt0rted:
This looks like an inconsistent regression. My `actions/setup-dotnet` PRs all came in going from `v1.9.1` to `v2`, while my `actions/setup-node` PRs came in for [`v2.5.1` to `v3.0.0`](https://github.com/xt0rted/markdownlint-problem-matcher/pull/398) and [`v2.5.1` to `v3`](https://github.com/xt0rted/tailwindcss-tag-helpers/pull/101).

xt0rted:
Just had a bunch more of these types of PRs opened. Wish there was a way to force dependabot to recreate them using the full version number.

Here's the relevant logs for the above PRs incase they help at all.

```
updater | INFO <job_304023390> Checking if actions/setup-node 2.5.1 needs updating
  proxy | 2022/02/28 03:01:35 [018] GET https://github.com:443/actions/setup-node.git/info/refs?service=git-upload-pack
[...]

PR #4953, opened 2022-04-05
Actions: Consider precision of current version when selecting latest

mctofu:
This attempts to address https://github.com/dependabot/dependabot-core/issues/4768

This has become noticeable recently with the release of the v3 versions of many github actions. The existing code to find the latest version did not deal with situations where multiple tags represented the latest version (v3, v3.0.0). Users that had pinned to a patch version were getting swapped to a major version (v2.1.5 -> v3) and some saw the opposite (v2 -> v3.0.0). Also, when v3.0.1 was released users pinned to a major version got [updates](https://github.com/mctofu/deadbeef-library/pull/9) to a patch version (v2 -> v3.0.1)

To work around this I've added a new method, `GitCommitChecker.local_tags_for_latest_version_commit_sha`, which returns all the version tags attached to the commit for the latest version. I then updated the Actions `UpdateChecker` to try and choose the version tag that matches the precision of the current version.


github_actions/lib/dependabot/github_actions/update_checker.rb
```diff
@@ -87,13 +85,39 @@ def fetch_latest_version_for_git_dependency
...
       end
 
+      def latest_version_tag
+        @latest_version_tag ||= begin
+          return git_commit_checker.local_tag_for_latest_version if dependency.version.nil?
+
+          latest_tags = git_commit_checker.local_tags_for_latest_version_commit_sha
+
+          # Find the latest version with the same precision as the pinned version.
+          # Falls back to a version with the closest precision if no exact match.
```
mctofu, line 95: I wasn't sure what to do here but assumed we'd probably run into this somewhere in the wild. A simpler option would be to refuse to update if we couldn't find a match?

github_actions/spec/dependabot/github_actions/update_checker_spec.rb
```diff
@@ -133,7 +133,7 @@
         let(:reference) { "1c24df3" }
 
         let(:repo_url) { "https://api.github.com/repos/actions/setup-node" }
-        let(:comparison_url) { repo_url + "/compare/v1.1.0...1c24df3" }
+        let(:comparison_url) { repo_url + "/compare/v1.1...1c24df3" }
```
mctofu, line 136: This was a side effect of adding a `v1.1` version to the test fixture.

mctofu (commented):
I think the cargo failure is unrelated?

jurre:
> I think the cargo failure is unrelated?

Yeah they look related to git dependencies, but checked on a different branch and locally that the same errors exist, might be testing against some public deps that have changed

common/lib/dependabot/git_commit_checker.rb
```diff
@@ -86,20 +86,32 @@ def head_commit_for_current_branch
...
-      if @raise_on_ignored && filter_lower_versions(filtered).empty? && filter_lower_versions(tags).any?
-        raise Dependabot::AllVersionsIgnored
+    def local_tags_for_latest_version_commit_sha
+      tags = allowed_tags
+      max_tag = tags.
+                max_by do |t|
+        version = t.name.match(VERSION_REGEX).named_captures.
+                  fetch("version")
+        version_class.new(version)
       end
```
brrygrdn, line 96: This seems to be a duplication of lines 114-119, is it maybe worth extracting this to a method?

It took me a few minutes for the need to do the transform from the version back to the tags ( which are versions effectively and likely why I did a double-take 😅  ) on the commit sha to click, isolating this as a method might allow us to make it a little more expressive.

github_actions/lib/dependabot/github_actions/update_checker.rb
brrygrdn: I think this approach is good, I think both outcomes are mildly surprising but failing over to _some_ kind of update at least alerts the user to the new version and highlights that the tagging strategy on the library has changed. 

Cases where we don't do anything are harder to rely to the user without them reading the logs closely, so I think this is the better outcome.

github_actions/spec/dependabot/github_actions/update_checker_spec.rb
```diff
@@ -235,13 +237,37 @@
...
+
+      context "using the major minor version" do
+        let(:reference) { "v1.0" }
+        it { is_expected.to eq(Dependabot::GithubActions::Version.new("2.1")) }
+      end
+
+      context "using the full version" do
+        let(:reference) { "v1.0.0" }
+        it { is_expected.to eq(Dependabot::GithubActions::Version.new("2.1.3")) }
+      end
```
brrygrdn, line 263: 🎉

github_actions/lib/dependabot/github_actions/update_checker.rb
brrygrdn: Do we have a test that covers the tie break behaviour? 

I haven't gone over the existing tests outside the diff, but it occurs to me we might already be testing this as a variation on the current behaviour but I couldn't spot it.

brrygrdn (commented):
I have a couple of questions, but overall this looks great 🥇

common/lib/dependabot/git_commit_checker.rb
mctofu: I tried to improve this in https://github.com/dependabot/dependabot-core/pull/4953/commits/9d1dafbd16fe3eae46a447d007a6f3747814a1ff but open to any other naming suggestions!

github_actions/lib/dependabot/github_actions/update_checker.rb
mctofu: Nope, there were not. It's a bit of work to setup test fixtures for each case so I wanted to be sure we were OK with the approach first. I added some tests in https://github.com/dependabot/dependabot-core/pull/4953/commits/aaa77031e5ae9a33bf23f3d4b7a6c53198d19383 by stubbing out the call to GitCommitChecker so the examples are easier to read.

brrygrdn (approved):
🚀  Thanks for addressing those comments, this looks great!

pzygielo:
The https://github.com/dependabot/dependabot-core/pull/4953#issuecomment-1095125413 and https://github.com/dependabot/dependabot-core/pull/4953#issuecomment-1095151523 happened in the short time while setup-java `3` tag was still pointing at the previous release. Once this `3` was moved to the latest release new PRs _consider precision_ again indeed.

I still do not think, that the case of major-only version not advanced in the same time as point-release should cause observed behaviour.

mctofu:
Thanks @pzygielo, that's a good catch! A mismatch in the `v3` and latest `v3.x.x` tag would definitely cause an issue when upgrading from a previous major version.

One argument for keeping the current behavior is that creating a PR for the `v3` tag would not be updating you to the latest version. Could this cause problems for actions which don't follow the recommend practices for tagging?

An argument against the current behavior is it's inconsistent with the behavior for updates within a major version. If you are currently on the `v3` tag which happens to be behind the latest `v3.x.x` tag you won't get an update. An update from `v2` -> `v3.x.x` is only triggered as part of a major version upgrade.

I think it does make sense to try and preserve consistency and stick to upgrading from `v2 -> v3`, even when `v3` isn't the latest release. We could do that by reworking the version selection to not require the `v3` tag is from the same commit as the latest `v3.x.x` tag.

pzygielo:
> An update from v2 -> v3.x.x is only triggered as part of a major version upgrade.

That's what happened in my cases, indeed.

Should I accept the change "Bump actions/setup-java from 2 to 3.1.1" I'd be again in the loop of PRs for every bugfix release. (As x.x.x precision would be preferred then.)
While for some, let's say, more trustworthy actions, that are not project's dependencies but CI env only, I'd prefer to implicitly upgrade as the tag is or isn't moved by action vendor.

Thanks for checking.

Issue #4887, opened 2022-03-21
Need to update the go build tag exclusion


jeffwidman:
https://github.com/dependabot/dependabot-core/blob/ea8fbaf64af5af477d5b69d0b985782340ca3b36/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb#L160 specifies `// +build` but this is now replaced by `//go:build`. Summary of the change in https://go.dev/doc/go1.18 and more details in https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md.

There's also a few places in tests that this shows up, so need to check if those need updating, or if they need to stay as-is to ensure dependabot handles the deprecated form in addition to the newer form. IE, may need tests of both styles.

PR #4954, opened 2022-04-05
Handle the new `go` build tag syntax

jeffwidman:
The build syntax changed from `1.16`-`1.18`:
* https://go.dev/design/draft-gobuild
* https://go.dev/doc/go1.18#go-command
* old syntax: `// +build`
* new syntax: `//go:build`

The `go fix` tool will update files in `1.17` to use both tags
simultaneously as a transitional version, but after that will replace
old tag styles with the new tag styles.

However, the old tag style continues to be supported for backwards
compatibility reasons, even though it's deprecated.

Fix https://github.com/dependabot/dependabot-core/issues/4887


go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -157,7 +157,12 @@ def run_go_get(dependencies = [])
           tmp_go_file = "#{SecureRandom.hex}.go"
 
           package = Dir.glob("[^\._]*.go").any? do |path|
-            !File.read(path).include?("// +build")
+            # The build syntax changed from 1.16-1.18:
+            # https://go.dev/design/draft-gobuild
+            # https://go.dev/doc/go1.18#go-command
+            # old syntax: // +build
+            # new syntax: //go:build
+            !File.read(path).include?("/(\/\/ \+build)|(\/\/go:build)/".freeze)
```
jeffwidman, line 165: I found this regex a bit inscrutable, so open to suggestions... 

I could pull this out into an array of regexes like most other string matching, or maybe Ruby has something as simple as "include?(x).or(y)"??

go_modules/spec/fixtures/projects/ignored_go_files_1.17/_ignored.go
```diff
@@ -0,0 +1,6 @@
+// This file is ignored because it starts with an underscore
```
jeffwidman, line 1: These files are copied from the other spec, and aren't really relevant to this particular test, but I thought they didn't hurt to test that the behavior doesn't change in the newer go versions.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
jurre: What about this?

```ruby
!File.read(path).match?(%r((// \+build)|(//go:build)))
```

jurre: Could also pull out the regular expression into a constant with a name but 🤷

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -157,7 +157,12 @@ def run_go_get(dependencies = [])
           tmp_go_file = "#{SecureRandom.hex}.go"
 
           package = Dir.glob("[^\._]*.go").any? do |path|
-            !File.read(path).include?("// +build")
+            # The build syntax changed from 1.16-1.18:
+            # https://go.dev/design/draft-gobuild
+            # https://go.dev/doc/go1.18#go-command
+            # old syntax: // +build
+            # new syntax: //go:build
+            !File.read(path).include?("/(\/\/ \+build)|(\/\/go:build)/")
```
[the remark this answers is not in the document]
mattt: > I am wondering though, since we're still calling `File.read`, does this make a difference in terms of memory usage? You may have meant to write `File.open`.

Indeed I did. `File.open` is the right method here. Thanks for catching that.

jakecoffman:
I was curious about this so I was looking into why it drops a dummy file in the package to run a blank `go get`. [PR 2700](https://github.com/dependabot/dependabot-core/pull/2700) introduces it because a blank `go get` will fail if there are no go files in the directory.

A `// +build` or `//go:build` can be used to exclude files from the build making `go get` fail, but they may also be used to include conditionally based on the architecture or OS. For example: `// +build arm64` or `//go:build arm64 && amd64`. So the code here is not actually doing enough to check to see if the file is going to be excluded on the system it runs on.

However, dropping a go file with a `package` line in it will always make `go get` succeed... even if the package name doesn't match the rest of the files in the directory! I assume this is because it doesn't actually compile anything when it runs.

So I propose we simplify this and just always drop the file.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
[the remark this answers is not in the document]
jeffwidman: Thanks @mattt / @jurre for not only the code suggestions, but also the explanations.

jeffwidman:
I updated the logic with @jakecoffman 's suggestion--he's right that the previous logic was both brittle and missed edge cases.

I wonder what the perf hit will be here though... we go from _rarely_ writing/deleting the temp file to _always_ writing/deleting the temp file.

Are these files being written to memory or disk? I realize the [SSD-to-memory gap is closing rapidly](https://itnext.io/modern-storage-is-plenty-fast-it-is-the-apis-that-are-bad-6a68319fbc1a), but memory is probably still faster / cheaper.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -154,13 +154,15 @@ def run_go_vendor
         end
 
         def run_go_get(dependencies = [])
+          # `go get` will fail if there are no go files in the directory.
+          # For example, if a `//go:build` tag excludes all files when run
+          # on a particular architecture. However, dropping a go file with
+          # a `package ...` line in it will always make `go get` succeed...
+          # Even when the package name doesn't match the rest of the files
+          # in the directory! I assume this is because it doesn't actually
+          # compile anything when it runs.
```
jeffwidman, line 163: FTR: I'm channeling @jakecoffman 's knowledge here, not mine.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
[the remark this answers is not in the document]
jeffwidman: Shouldn't `^//\s*build` actually be `^//\s*\+build`? To capture the `+` in the old syntax?

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -154,13 +154,15 @@ def run_go_vendor
         end
 
         def run_go_get(dependencies = [])
+          # `go get` will fail if there are no go files in the directory.
+          # For example, if a `//go:build` tag excludes all files when run
+          # on a particular architecture. However, dropping a go file with
+          # a `package ...` line in it will always make `go get` succeed...
+          # Even when the package name doesn't match the rest of the files
```
jurre, line 161: Interesting! I wonder if this behavior could break on us in the future, because it does seem a bit counter-intuitive, but I guess we can cross that bridge when we get there

jurre (approved):

jakecoffman:
@jeffwidman if you want to leave the previous behavior, that is ok with me. We don't know if it's really causing any problems, I was just looking for ways to simplify the code.

Also consider that files can exclude based on filename as well, e.g. `blah_linux.go` would only build on Linux.

jeffwidman:
> Also consider that files can exclude based on filename as well, e.g. blah_linux.go would only build on Linux.

Oh, I didn't realize that. Agreed then that the only thing that will guarantee correctness (at least right now) is to drop the `fake.go` file. I'm a little concerned about the perf impacts, so I'm going to do a little more digging into cost of writing/deleting a file on every `go` job.

mattt (approved):

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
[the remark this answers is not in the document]
jeffwidman: Oh silly me, this regex disappeared with the refactor to always write the temp file, so this doesn't even matter.

PR #4958, opened 2022-04-05
Restore python 3.6 support

pavera:
This PR restores python 3.6 support by allowing version ranges for pip and pip-tools.  The latest versions of these dependencies no longer support python 3.6, so we need to allow earlier versions.

Looking into an issue creating PRs for a python repo found that if you specify a python version, IE 3.6.8 for your repo, then when we attempt to create the 3.6.8 pyenv and pip install -r python/helpers/requirements.txt it fails because some dependencies in this helper have removed python 3.6 support in their latest versions.

I'm not clear on whether dependabot will just immediately try to bump these versions again?  If it will what is the correct way to pin these versions back?

I'd like to add tests as well, but need a pointer or 2 about where to add such "integration" type tests.


python/helpers/requirements.txt
```diff
@@ -1,5 +1,5 @@
-pip==22.0.4
-pip-tools==6.5.1
+pip>=21.3.1
+pip-tools>=6.4.0
```
jakecoffman, line 2: When `pip install -r` runs won't `>=` just pick the latest version and not restore Python 3.6 support?

If so, I think we should pin it with `==` and tell dependabot `@dependabot ignore this minor version` if it tries to bump it. 

When `6.6.0` releases, dependabot will make a PR and we can re-evaluate whether to update to `6.5.X`. 

Maybe also add a comment here as to why we're lagging behind on these releases, so when dependabot makes a PR we don't just merge it right on in!

Does that sound like it would work? 🤔 Do we have a policy on what older versions we support?

pavera: I tested this setup on ubuntu 18.04 (python 3.6.9) and it does the correct thing.  Pip detects that the latest "compatible" version is 21.3.1 and 6.4.0 and correctly installs those on python 3.6, also correctly installs the actual latest version on python >= 3.7.

Comments incoming :)

jakecoffman:
This makes sense now that I learned the Python helpers are run with the target version of Python.

Many of the spec are integration style tests already, they make calls to GitHub and compile external code! So maybe stick a new test in `pip_compile_file_updater_spec.rb`?

Also, I noticed this file https://github.com/dependabot/dependabot-core/blob/48f4df6b37e686f95fd9a71681e1c0976cc03cea/python/lib/dependabot/python/python_versions.rb#L12-L19

Should we test back to 3.5.3? 🤔

pavera:
Yeah, I also had read the version file, going back to 3.5 pulls these two packages back quite a ways.  I figured I'd wait and see on that.  3.6 is already eol, but ubuntu 18.04 is still supported and that is the official python version there.  Not sure if there are any supported distros that are keeping 3.5 alive still (rhel/centos maybe?)

Good to know about the tests, I will endeavor to add one for this case.

jakecoffman (approved):
Looks good, nice work! 👏

PR #4962, opened 2022-04-06
code is unsused: main packages have no particular effect on dependencies

jakecoffman:
While looking at #4954 I noticed the code was gathering main files for some reason, so I pulled the thread and it wasn't attached to anything!

I assume this code was from the days of `dep`. Go modules don't give main files any special treatment, so removing this code.


mctofu (approved):
✂️

jeffwidman:
Nice! Yeah, there's quite a bit of leftover `dep` stuff, some evening when I have some free time I am hoping to pickup  https://github.com/dependabot/dependabot-core/pull/3537 again.

PR #4969, opened 2022-04-07
handle multiple python manifests containing the same dependency

jakecoffman:
Repositories containing multiple Python dependency management manifests in the same directory that specify the same dependency are seeing "dependency not vulnerable" messages.

This PR fixes one of those cases.

See [dsp-testing/python-multiple-manifests](https://github.com/dsp-testing/python-multiple-manifests) for two examples. Currently neither of those directories can do an update, but let's start with the one this PR fixes:

## scenario 1

In [/lockfile](https://github.com/dsp-testing/python-multiple-manifests/tree/main/lockfile) of that repo, we have a Pipfile.lock that requires the latest version of `pipenv` version 2022.3.28, but requirements.txt has an older version 2022.1.8. The reason we don't get an update currently is due to the code that this PR removes where it specifically skips when the dependency exists in another manifest already.

So once we remove those lines, it works:

```
[dependabot-core-dev] ~/dependabot-core $ ruby bin/dry-run.rb pip dsp-testing/python-multiple-manifests --dir=/lockfile --cache=files
=> failed to read all dependency files from cache manifest: ./dry-run/dsp-testing/python-multiple-manifests/lockfile/cache-manifest-pip.json
=> fetching dependency files
=> dumping fetched dependency files: ./dry-run/dsp-testing/python-multiple-manifests/lockfile
=> parsing dependency files
=> updating 1 dependencies: pipenv

=== pipenv (2022.1.8)
 => checking for updates 1/1
 => latest available version is 2022.3.28
 => latest allowed version is 2022.3.28
 => updating pipenv from 2022.1.8 to 2022.3.28

    ± requirements.txt
    ~~~
    1c1
    < pipenv==2022.1.8
    ---
    > pipenv==2022.3.28
    ~~~
```

## scenario 2

Now looking at [/pipfile](https://github.com/dsp-testing/python-multiple-manifests/tree/main/pipfile) of that repo to see a case this doesn't fix. If the Pipfile has the latest version (instead of only the Pipfile.lock!) then this doesn't change anything:

```
[dependabot-core-dev] ~/dependabot-core $ ruby bin/dry-run.rb pip dsp-testing/python-multiple-manifests --dir=/pipfile --cache=files
=> failed to read all dependency files from cache manifest: ./dry-run/dsp-testing/python-multiple-manifests/pipfile/cache-manifest-pip.json
=> fetching dependency files
=> dumping fetched dependency files: ./dry-run/dsp-testing/python-multiple-manifests/pipfile
=> parsing dependency files
=> updating 1 dependencies: httplib2

=== httplib2 (0.20.4)
 => checking for updates 1/1
 => latest available version is 0.20.4
 => latest allowed version is 0.20.4
```

I think this is because of this line in dependency_set: https://github.com/dependabot/dependabot-core/blob/eaf091f6c4fcf0a67e8f6c44e5827ea578fd838b/common/lib/dependabot/file_parsers/base/dependency_set.rb#L66

Since it's not in the requirements of the previous example, it is able to choose the lowest version to update. However in this example it does exist and chooses the larger version instead. 

## Conclusion

Merging this will fix *some* of the failures to update, but we might want to revisit that line in dependency_set and check to see if it's in BOTH of the requirements and if so pick the lowest. 

I can't find the original reasoning this was skipping when the dependency existed in another manifest. The comment seems to indicate FileUpdater just wasn't ready for it? If anyone has another idea of a situation to test against let me know!


python/lib/dependabot/python/file_parser.rb
```diff
@@ -72,12 +72,6 @@ def poetry_dependencies
       def requirement_dependencies
         dependencies = DependencySet.new
         parsed_requirement_files.each do |dep|
-          # This isn't ideal, but currently the FileUpdater won't update
```
jurre, line 75: I wonder what has changed to make this statement no longer hold true, maybe we should add a test to the FileUpdater that verifies this keeps working in the future?

jakecoffman: I added two tests. 

[One](https://github.com/dependabot/dependabot-core/pull/4969/files#diff-92c4c5a8f4322dade1159b654421d41742e07913bcb3b07db7c065fc48363f00R1328-R1346) shows that when the lockfile + requirement.txt files share a dependency what the DependencySet combiner does.  Since there's only one requirement, the rest of the update will be standard and is already tested in other specs.

[The next test](https://github.com/dependabot/dependabot-core/pull/4969/files#diff-8384d4d3db066e463df0d4c95a00929220ee4f27ee7a97d2858d6dfa3e3501d5R171-R218) shows that having two requirements is ok with the FileUpdater. It doesn't replace the dependency in both files, but it didn't before this change either! 😄

jurre (approved):
I _think_ this is an overall improvement, although the experience of only a single manifest being updated might be frustrating for users. I don't have a clear sense as to why you'd even have this sort of setup so it's a little hard to determine what the right thing to do is, so I'll trust your judgement on this 👍

jakecoffman:
I think the right thing to do is update every manifest... and this is a small step in that direction 😄

Issue #4834, opened 2022-03-11
dependabot not detecting latest version of github action


rethab:
**Package ecosystem**

- github actions

**dependabot.yml content**

```yaml
version: 2
updates:
  - package-ecosystem: 'github-actions'
    directory: '/'
    schedule:
      interval: 'weekly'
    commit-message:
      prefix: 'feat(deps)'
```
**Updated dependency**

- nothing was updated

**What you expected to see, versus what you actually saw**

workflow references action like so:
```yaml
        uses: actions/download-artifact@v2
```

In the logs from dependabot I see:

```text
updater | INFO <job_31347D_N_T> Checking if actions/download-artifact 2 needs updating
  proxy | 2022/03/11 14:27:39 [040] GET https://github.com:443/actions/download-artifact.git/info/refs?service=git-upload-pack
  proxy | 2022/03/11 14:27:39 [040] * authenticating git server request (host: github.com)
  proxy | 2022/03/11 14:27:39 [040] 200 https://github.com:443/actions/download-artifact.git/info/refs?service=git-upload-pack
updater | INFO <job_31347D_N_T> Latest version is f023be2c48cc18debc3bacd34cb396e0295e2869
updater | INFO <job_31347D_N_T> No update needed for actions/download-artifact 2
```

but this is wrong: https://github.com/actions/download-artifact/releases there is a version `v3.0.0` as well as a major tag `v3`.

the latest version f023be2c48cc18debc3bacd34cb396e0295e2869 mentioned in the logs points to `v2`.

So I would expect dependabot to create a PR and update this action to `v3`.

mctofu:
Thanks for this report! This is a problem in https://github.com/dependabot/dependabot-core/blob/bb84f6c03bd6283c7c83774d41272fd3b07fbbe6/common/lib/dependabot/git_commit_checker.rb#L52-L53

It's incorrectly matching the https://github.com/actions/download-artifact/tree/v2-preview branch.

rethab:
Thanks for your response @mctofu . Are you going to fix this? If not, would you accept a PR? In the latter case, could you please give some more hints?

mctofu:
@rethab: The fix was pretty simple so I went ahead and opened a PR for it.

mctofu:
We should keep this open until the fix is released and confirmed to work

rethab:
[...]

PR #4972, opened 2022-04-07
Capture failing testcase

mctofu:
WIP to fix https://github.com/dependabot/dependabot-core/issues/4834

actions/download-artifact contains refs/heads/v2-preview which causes Dependabot to believe that version v2 is a branch


common/lib/dependabot/git_commit_checker.rb
```diff
@@ -50,7 +50,7 @@ def pinned?
       return true if dependency.version&.start_with?(ref)
 
       # Check the specified `ref` isn't actually a branch
-      !local_upload_pack.match?("refs/heads/#{ref}")
+      !local_upload_pack.match?("\srefs/heads/#{ref}$")
```
landongrindheim, line 53: It took me a minute to reason through this change, specifically the `\s` inclusion. I _think_ having the argument be a regex would make it a bit easier for **me** to parse. What do you think of the following?

````suggestion
      !local_upload_pack.match?(%r{ refs/heads/#{ref}$})

landongrindheim (approved):
Left a comment that you can do with as you please 🙂

🚀

PR #4973, opened 2022-04-07
Reduce bundler version hardcoding

deivid-rodriguez:
This is an experiment to reduce the number of places that need to be updated when Bundler version is bumped to just the `Dockerfile` and the `bundler/helpers/v2/Gemfile.lock` files.


mattt:
Thanks for working on this, @deivid-rodriguez! I just opened #5024, which I hope to help DRY up a lot of our infrastructure. Do you see any potential issues / opportunities with this new approach?

deivid-rodriguez:
@mattt Unfortunately my idea here didn't work as well as I expected, because of what I think it's a chicken-and-egg issue. I wanted to let RubyGems (required first thing when booting Ruby) choose the appropriate Bundler version according to the version in the `Gemfile.lock` file, instead of having to reimplement that logic ourselves. But it turns out we need to know the Bundler version beforehand, so that we can apply the proper monkey patches to Bundler, so that dependabot works properly. I left the PR opened because I want to have another look, but it my initial approach didn't really work well.

I'll have a look at the new infra to see if I come up with new dry up ideas, but don't worry at all about how it could affect this cleanup since it's already not a successful attempt :)

deivid-rodriguez:
Alright, I found a way to do this.

It doesn't allow to remove Dependabot helpers to parse lockfiles, which was one original motivation from https://github.com/dependabot/dependabot-core/pull/4884#discussion_r830971299.

But it does accomplish the goal of having a single source of truth for the Bundler version run by Dependabot (well, two, actually, but much better than before :)).

bundler/lib/dependabot/bundler/helpers.rb
```diff
@@ -21,7 +17,7 @@ def self.bundler_version(lockfile)
         if (matches = lockfile.content.match(BUNDLER_MAJOR_VERSION_REGEX))
           matches[:version].to_i >= 2 ? V2 : V1
         else
-          FAILOVER
+          DEFAULT
```
deivid-rodriguez, line 20: This is a change over how dependabot currently works. I made it when trying to figure out how to completely remove this file and delegate the logic to RubyGems, but finally gave up so the change is no longer necessary I believe.

I will remove https://github.com/dependabot/dependabot-core/pull/4973/commits/3c1cfcdab1d4165a5341c191736a58b6a087bd39 from this PR, and propose it separately if I feel it's useful.

deivid-rodriguez:
I got one more idea to potentially further improve this and leave the Bundler version in a single place, and a much simpler setup. Will give it a try tomorrow.

deivid-rodriguez:
Ok, I put together a small patch that leaves the exact Bundler version at a single place (the Dockerfile). The removal of barcoded versions is based on two ideas:

* Bundler by default runs the highest version installed the image. That's 2.3.x. There's no need to explicit pass `BUNDLER_VERSION` for running tests and building helpers since that's default default behavior (it's still needed though to force Bundler 1.17.3).
* When shelling out to native helpers, we don't need to know the exact version, we just need to know the major version to figure out the path to native helpers `run.rb` script and everything else. Proper activation of bundler can be done inside the `run.rb` script without needing the full version either.

jeffwidman:
@deivid-rodriguez is this ready for review or still a work-in-progress?

deivid-rodriguez:
Yes! I only changed some commit messages to try to better explain the approach I ended up taking, but other than that I think it's ready!

jeffwidman (approved):
This all makes sense to me, thank you for the helpful commit messages, they made it much easier to follow the logic.

I'm 👍 for merging this, but I don't have time to cut a release and deploy it just yet, and I don't want to push that risk onto a teammate w/o their permission in case there's a problem and they need to revert. I'll see if someone else can take a look, otherwise will merge when I have the space to also deploy it.

deivid-rodriguez:
That sounds super wise @jeffwidman 👍, thanks!

jurre:
Yes, I think this should work, and is a super welcome simplification! 🎉

deivid-rodriguez:
Let me know of any issues 🤞

jeffwidman:
Thx again for this @deivid-rodriguez, a really nice simplification.

PR #4981, opened 2022-04-08
pub: Bump when publish to none

sigurdm:
Apps created from a flutter template will have a 'version' field, though they are not library-packages, they will however have publish_to: none to indicate they are not libraries.

The template is here: https://github.com/flutter/flutter/blob/master/packages/flutter_tools/templates/app/pubspec.yaml.tmpl


jurre (approved):
Makes sense, thanks!

PR #4988, opened 2022-04-12
Python: unwrap arbitrary parentheses from requirement string

jakecoffman:
When encountering a Python requirement containing parentheses, dependabot-core would throw `Illformed requirement`. 

This fixes that error by unwrapping simple parenthesis patterns. 

I tried looking in the [packaging](https://github.com/pypa/packaging/blob/main/tests/test_requirements.py) and [poetry](https://github.com/python-poetry/poetry-core/blob/master/tests/version/test_requirements.py) tests for more complex examples but could only find simple ones. Hopefully that means this will catch most of the use cases for parentheses inside of a requirement string!


brrygrdn (approved):
:shipit:

jurre (approved):
Python requirement syntax is wild!

PR #4992, opened 2022-04-13
Adds failing tests for Typescript @types updates

pavera:
This work is the start of adding definitely typed @types package updates in sync with the original package.

The skipped tests added here are skipped as they are failing tests awaiting implementation of underlying support.


mattt (approved):
Great work! Really happy to have these in place for when we build out this functionality.

PR #5000, opened 2022-04-14
Updating a package and its related @types package simultaneously

pavera:
This PR updates both a javascript package and its corresponding definitely typed @types package in a single PR.

We followed the same mechanism used by peer dependency updates to unlock all dependencies and then detect that there is an update to both the original package and @types package and submit a PR with both dependencies updated.


npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -222,6 +225,33 @@ def updated_monorepo_dependencies
           updates
         end
 
+        def types_update_available?
+          # TODO: This should probably actually check if there is an updated version
+          #      along with just the existence of the dependency.
+          types_name = PackageName.new(dependency.name).types_package_name
+          types_dep_to_update = top_level_dependencies.find { |d| d.name == types_name.to_s }
+          return true if types_dep_to_update
```
landongrindheim, line 233: I think we can take advantage of `Enumerable#any?` for this and remove the explicit return 🙂 What do you think of?
````suggestion
          types_package_name = PackageName.new(dependency.name).types_package_name.to_s
          top_level_dependencies.any? { |d| d.name == types_package_name }
````

File: npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb [hunk 2]
```diff
@@ -222,6 +225,33 @@ def updated_monorepo_dependencies
...
+        end
+
+        def updated_types_dependencies
+          types_name = PackageName.new(dependency.name).types_package_name
+          types_dep_to_update = top_level_dependencies.find { |d| d.name == types_name.to_s }
+          return unless types_dep_to_update
```
landongrindheim, line 239: 💡 Since this logic is referenced twice, we could extract it into a private method. Maybe something like 💭

```
def types_package
  types_package_name = PackageName.new(dependency.name).types_package_name.to_s
  top_level_dependencies.find { |d| d.name == types_package_name }
end
```

Then we could reference it in `#types_update_available?` as

```ruby
def types_update_available?
  !types_package.nil?
end
```

And in `#updated_types_dependencies` as:

```ruby
def updated_types_dependencies
  return unless types_update_available?
  # omitted
```

What do you think?

npm_and_yarn/spec/fixtures/npm_responses/types_jquery.json
```diff
@@ -0,0 +1 @@
...ain":"","types":"index","repository":{"type":"git","url":"https://github.com/DefinitelyTyped/DefinitelyTyped.git","directory":"types/jquery"},"scripts":{},"dependencies":{"@types/sizzle":"*"},"typesPublisherContentHash":"df62dc4dca38a444be365d9e0268e088a19ae08d0a4464c56083e0aea9df2c54","typeScriptVersion":"2.3","_id":"@types/jquery@3.3.30","dist":{"integrity":"sha512-chB+QbLulamShZAFcTJtl8opZwHFBpDOP6nRLrPGkhC6N1aKWrDXg2Nc71tEg6ny6E8SQpRwbWSi9GdstH5VJA==","shasum":"af4ad612d86d954d74664b2b0ec337a251fddb5b","tarball":"https://registry.npmjs.org/@types/jquery/-/jquery-3.3.30.tgz","fileCount":10,"unpackedSize":1099089,"npm-signature":"-----BEGIN PGP SIGNATURE-----\r\nVersion: OpenPGP.js v3.0.4\r\nComment: https://openpgpjs.org\r\n\r\nwsFcBAEBCAAQBQJdESoWCRA9TVsSAnZWagAABPwP/iDJYNJv02A06PbtuX2L\nX8gyF+bfBkt7I0fTsA3EuGghiWUeVFc6ANxx8RzE6WLPnuoy7w8GNCanUhjd\n6n8FtxUpKRbZbEumFmrM7uQmBE4rz+7e/TXndXsDd9sCYo5by1hSET6xjLNf\n3QV3FzSjjgVZP7ou0REHT4jMFNowbY7bqcsW2eN6fDOs57L2X9+UwmpOczP/\nYpi1TLD8Qdkso0RAT7oFpG4FiRXhHe1kZtshwfB9Ndnrnwd5O/yLLQtkMWbV\nye250DVz1fNmrMEyKBVM3D2t3XuMJt9TmgqxKpmEPmUdNOSNvnBfxMK4egCM\nrKuM9KNLbYpI1WHYV7M8udfULUrjIKYxjiTra7cy0KnjMovAbFEe3cbbD4Ia\n7wiw3GOWtsSSJDnj3T7Kqi3Vg
```
landongrindheim, line 1: Should this be empty?

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -74,6 +75,7 @@ def initialize(dependency:, credentials:, dependency_files:,
         def latest_resolvable_version
           return latest_allowable_version if git_dependency?(dependency)
           return if part_of_tightly_locked_monorepo?
+          return if types_update_available?
```
landongrindheim, line 78: Please pardon my ignorance here 😇 Do we return early here because we have more than one dependency to return and will need to do a full unlock?

pavera: Yes, at least I think :) I'm also mostly ignorant, but that seemed to be the pattern for peer deps and the "tightly locked" deps, so I followed it.  Would love input from @jurre on this.

npm_and_yarn/spec/fixtures/npm_responses/types_jquery.json
pavera: What are you referring to here?  Just not clear on the context of this comment.

landongrindheim: Sorry for the confusion 😅 

I was referring to the fixture file only having a new line. I noticed the other fixture did as well. I didn't delve deeply into the code, so I'm not sure how the file is being parsed/processed. I think I had expected to find some data in the fixture.

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -222,6 +225,33 @@ def updated_monorepo_dependencies
           updates
         end
 
+        def types_package
+          types_package_name = PackageName.new(dependency.name).types_package_name.to_s
+          top_level_dependencies.find { |d| d.name == types_package_name }
```
mattt, line 230: You could memoize this to avoid computation on each call.

```suggestion
          @types_package ||= begin
            types_package_name = PackageName.new(dependency.name).types_package_name.to_s
            top_level_dependencies.find { |d| d.name == types_package_name }
          end
```

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -222,6 +225,33 @@ def updated_monorepo_dependencies
...
+        end
+
+        def types_update_available?
+          return false if types_package.nil?
+
+          latest_version = latest_version_finder(types_package).latest_version_from_registry
+          (version_class.new(types_package.version) < latest_version && # there is a newer version of types package
+          latest_version.segments[0] <= latest_allowable_version.segments[0]) # and not greater than original package
```
mattt, line 238: Possible suggestion to improve readability. At a glance, I missed that the last two lines were part of the same compound return value expression.

```suggestion
          latest_version = latest_version_finder(types_package).latest_version_from_registry
          return false unless latest_version.segments.first <= latest_allowable_version.segments.first

          types_package_version = version_class.new(types_package.version)
          return false unless types_package_version < latest_version

          true
```

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -222,6 +225,33 @@ def updated_monorepo_dependencies
...
+
+          updated_version =
+            latest_version_finder(types_package).latest_version_from_registry
+          {
+            dependency: types_package,
+            version: updated_version,
+            previous_version: resolve_latest_previous_version(
+              types_package, updated_version
+            )
+          }
```
mattt, line 252: Style nitpick: Separate implicit return expression from other statements with a blank line. Otherwise, this hash literal can be confused for a block argument for the preceding method call.

```suggestion
          updated_version = latest_version_finder(types_package).latest_version_from_registry

          {
            dependency: types_package,
            version: updated_version,
            previous_version: resolve_latest_previous_version(
              types_package, updated_version
            )
          }
```

npm_and_yarn/lib/dependabot/npm_and_yarn/package_name.rb
```diff
@@ -37,12 +37,16 @@ def to_s
         end
       end
 
-      def <=>(other)
-        to_s.casecmp(other.to_s)
+      def eql?(other)
+        self.class == other.class && self == other
```
mattt, line 41: See https://apidock.com/ruby/Object/eql%3F

> The [eql?](https://apidock.com/ruby/Object/eql%3F) [method](https://apidock.com/ruby/Object/method) returns true if obj and other refer to the same [hash](https://apidock.com/ruby/Object/hash) key. This is used by [Hash](https://apidock.com/ruby/Hash) to test members for equality. For objects of [class](https://apidock.com/ruby/Object/class) [Object,](https://apidock.com/ruby/Object) [eql?](https://apidock.com/ruby/Object/eql%3F) is synonymous with ==. Subclasses normally continue this tradition by aliasing [eql?](https://apidock.com/ruby/Object/eql%3F) to their overridden == method, but there are exceptions. [Numeric](https://apidock.com/ruby/Numeric) [type](https://apidock.com/ruby/Object/type)s, for example, perform type conversion across ==, but not across eql?, so:
> ```ruby
> 1 == 1.0     #=> true
> 1.eql? 1.0   #=> false
> ```

npm_and_yarn/lib/dependabot/npm_and_yarn/version.rb
```diff
@@ -34,6 +34,27 @@ def initialize(version)
...
+      end
+
+      def patch
+        @patch ||= segments[2] || 0
+      end
+
+      def backwards_compatible_with?(other)
+        case major
+        when 0
+          self == other
```
mattt, line 52: According to the [SemVer spec](https://semver.org)

> 4. Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

> 6. Patch version Z (x.y.Z | x > 0) MUST be incremented if only backwards compatible bug fixes are introduced. A bug fix is defined as an internal change that fixes incorrect behavior.

I can't tell if the carve-out for `x > 0` means that we should match with strict equality here. Should we instead do this?

```suggestion
          major == other.major && minor == other.minor
```

mattt: Then again, npm advises releases start numbering at `1.0.0`, and there only appears to be a [single `@types` package with a major zero version](https://www.npmjs.com/search?q=%40types%200.) So maybe this is fine as is?

landongrindheim: Dependabot's [version update docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates) don't promise compatibility (emphasis mine):

> Dependabot takes the effort out of maintaining your dependencies. You can use it to ensure that your repository automatically keeps up with **the latest releases** of the packages and applications it depends on. That's sort of a different concern though (direct dependencies).

We *should* ship compatible packages together. Given items `4` and `6` from the SemVer spec, I think the current code is the only one in which we can guarantee backward compatibility.

---

💡 I wonder if this method would benefit from being a little more specific in its name or argument. In our specific use case, we're considering two packages backward compatible if their versioning schemes could safely resolve to each others. I don't think that's necessarily true of other packages (like peer dependencies). I'll try to think of something and update this comment if I do 😄

mattt (approved):

npm_and_yarn/lib/dependabot/npm_and_yarn/version.rb
```diff
@@ -34,6 +34,27 @@ def initialize(version)
...
+
+      def minor
+        @minor ||= segments[1] || 0
+      end
+
+      def patch
+        @patch ||= segments[2] || 0
+      end
+
+      def backwards_compatible_with?(other)
```
landongrindheim, line 49: My 🪙🪙 -- I think this method name is a little misleading _on this object_. I understand this to be specific to `@types` packages, where we're treating `@types/foo@3.5` as compatible with `foo@3.5`. However, I don't think this is generalizable to other use cases.

Since this method is on `NpmAndYarn::Version`, I'd like to see us adopt a name that _is_ generalizable and true in all cases. I'll try to think of some names and leave them here in a follow-up comment 🙂

landongrindheim: My thinking here is that the breaking changes in SemVer refer to the package _itself_, while this is (at least how we're using it so far) comparing two distinct packages. Does that make sense @pavera @mattt 

Maybe something like `shares_stable_version_with?`? or `eqivalent_stable_version?`

mattt: As I understand it, this `Version` object represents only the version of a package, independent of any particular package release. Given two releases of the same package, this method _should_ provide the correct answer for whether they're backwards compatible — at least as I understand how that's defined in the [SemVer 2 spec](https://semver.org/)

> Given a version number MAJOR.MINOR.PATCH, increment the:
>
> * MAJOR version when you make incompatible API changes,
> * MINOR version when you add functionality in a backwards compatible manner, and
> * PATCH version when you make backwards compatible bug fixes

So we should be able to use `backwards_compatible_with?` to compare versions of either the same package or a `@types` package and its corresponding package. The fact that this corresponds with how `@types` packages are associated is a happy coincidence.

landongrindheim: > Do you see any particular test case that you find surprising / breaks your expectation of how this method should behave?

I don't. It's odd to me that our fist use case is comparing versions of separate packages. Your reply highlights that I don't have a good understanding of my own reservations. Since I don't have a better way to say it, let's move forward with what we have 😄

landongrindheim (approved):

npm_and_yarn/lib/dependabot/npm_and_yarn/version.rb
mattt: > My thinking here is that the breaking changes in SemVer refer to the package _itself_, while this is (at least how we're using it so far) comparing two distinct packages. Does that make sense @pavera @mattt
> 
> Maybe something like `shares_stable_version_with?`? or `eqivalent_stable_version?`

(Sorry, my last comment came right after yours)

OK, that really helped me understand your concern, I think. 

It sounds to me that your objection is more with how this method is used than its name — that we're saying "3.5.0" is backwards compatible with "3.6.1", even though we're talking about two different packages. Would you agree with that?

If we change the name, we'd still be comparing 🍎s to 🟠s. It just happens that, in the case of `@types` packages, it's 🍎 to 🍏 , so it works out.

We _could_ raise up this API to the level of comparing a tuple of `(Package, Version)`, which could tell you that you're comparing 🍎s to 🟠s. But for the one place that we're making this inter-package comparison, I'm personally alright with leaving things as they are.

PR #5001, opened 2022-04-14
Map Package Name to Types Package Name

landongrindheim:
This is work that is being done to support updating TypeScript types
packages at the same time that we update the packages that consume them.
There are a couple of assumptions baked into this work:

* Types packages will share the same name as the libraries consuming
  them
* Libraries will follow the rules put forward by
  https://github.com/npm/validate-npm-package-name

**Note:** This code is not integrated anywhere yet. The intention is to translate between packages to make sure we can update them in tandem.


npm_and_yarn/lib/dependabot/npm_and_yarn/package_name.rb
```diff
@@ -0,0 +1,58 @@
...
+      def to_s
+        if scoped?
+          "@#{@scope}/#{@name}"
+        else
+          @name.to_s
+        end
+      end
+
+      def <=>(other)
+        to_s <=> other.to_s
```
mattt, line 34: We should either be more discerning on what we accept (lowercase only), or make this comparison case insensitive.

```suggestion
        to_s.casecmp(other.to_s)
```

I slightly favor the latter, since we care more about matching declared intent more than enforcing [npm moniker rules](https://blog.npmjs.org/post/168978377570/new-package-moniker-rules.html)

mattt: Oh, actually — that's I forgot that's what npm calls them. _moniker_. I'm fine with `PackageName` as a class name, but `PackageMoniker` would also be a reasonable alternative, I think.

npm_and_yarn/lib/dependabot/npm_and_yarn/package_name.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      def types_package
+        return self if types_package?
+
+        if scoped?
+          "@types/#{@scope}__#{@name}"
+        else
+          "@types/#{@name}"
+        end
+      end
```
mattt, line 45: Sorry, I misspoke when I gave feedback about this earlier. What I meant to suggest was that this method return class instances instead of string:

```suggestion
      def types_package_name
        return nil if types_package?

        if scoped?
          PackageName.new("@types/#{@scope}__#{@name}")
        else
          PackageName.new("@types/#{@name}")
        end
      end
```

A few possible advantages:

- You can do a direct equality check to find the types package for a package (e.g. `devDependencies.include?(package.types_package_name)`
- Running through initializer again does second round of validation, which is unlikely, but possible (e.g. `@a/really_long_package_name(...up to 214 limit...)` would produce an invalid package name, because the resulting name is too long).

mattt: ...but that may be not worth the effort. In which case `return self` should be `return nil`.

```suggestion
      def types_package
        return nil if types_package?

        if scoped?
          "@types/#{@scope}__#{@name}"
        else
          "@types/#{@name}"
        end
      end
```

npm_and_yarn/lib/dependabot/npm_and_yarn/package_name.rb
landongrindheim: I addressed this (and your other suggestions) in [997cade](https://github.com/dependabot/dependabot-core/pull/5001/commits/997cadecd2b823f4b4c8ece5bae0a7b7c4777c9d) 🙂 

I opted to keep `PackageName` as that npm blog post used "moniker" only in the title, then used "name" throughout the contents of the post. I had also referenced a package wherein npm was validating "names". The term "name" seems less esoteric and I couldn't think of any semantic meaning that's lost, so I ran with it.

npm_and_yarn/lib/dependabot/npm_and_yarn/package_name.rb
landongrindheim: Addressed a combination of these two comments in [997cade](https://github.com/dependabot/dependabot-core/pull/5001/commits/997cadecd2b823f4b4c8ece5bae0a7b7c4777c9d) 😄

mattt (approved):

npm_and_yarn/lib/dependabot/npm_and_yarn/package_name.rb
```diff
@@ -0,0 +1,63 @@
...
+
+module Dependabot
+  module NpmAndYarn
+    class PackageName
+      DEFINITELY_TYPED_SCOPE = /types/i.freeze
+      PACKAGE_NAME_REGEX     = %r{
+          \A                                         # beginning of string
+          (?=.{1,214}\z)                             # enforce length (1 - 214)
+          (@(?<scope>[a-z0-9\-~][a-z0-9\-\._~]*)\/)? # capture 'scope' if present
+          (?<name>[a-z0-9\-~][a-z0-9\-._~]*)         # capture package name
```
bdragon, line 11: Drive-by comment, completely ignorable: inside a character class you don't need to escape `-` if it appears first or last (e.g., `[-a-z0-9~]`) and you don't have to escape the `.` (not on this line but the one above). Just a thought to -5 characters 🙂

Also, I love the clarity of the multi-line expression with comments.

PR #5002, opened 2022-04-14
nuget: fix PR missing commits in message when using private registry

jakecoffman:
## Context

Dependabot PRs are missing the release notes and commits sections that we expect them to have when using a private NuGet registry like nuget.pkg.github.com.

This is because the GitHub NuGet registry doesn't support downloading `.nuspec` files currently, and [this is the way we're finding the source repo of the project](https://github.com/dependabot/dependabot-core/blob/bdbf87a1f64594997092fc74f3998b17b75d8ec4/nuget/lib/dependabot/nuget/metadata_finder.rb#L58-L77).

The `.nuspec` URL is part of the [NuGet Server spec](https://docs.microsoft.com/en-us/nuget/api/package-base-address-resource#download-package-manifest-nuspec).

## A solution

This PR adds a fallback to try to find the `projectUrl` or `licenseUrl` in the search data, which ***might*** contain the repo. The search URL is supported by GitHub Packages and the official NuGet server. 

This will require users to enter a repo URL in the `projectUrl`, instead of the `repositoryUrl` field, since `repositoryUrl` isn't exposed in the search endpoint. The only way to get to that `repositoryUrl` seems to be in that `.nuspec` file.

Another possible solution is downloading the `.nupkg` and extracting the `.nuspec`, but I think that could have an effect on billing and that is probably undesirable. Also there's a risk the `.nupkg` is quite large since it contains binaries.

We could also ask GitHub Packages to implement that endpoint, this PR is meant as a stopgap or fallback when the `nuspec` is not available.

**No tests yet: Wanted to get buy-in from the approach here.**

### DIFF
--- a/nuget/spec/dependabot/nuget/metadata_finder_spec.rb
+++ b/nuget/spec/dependabot/nuget/metadata_finder_spec.rb
@@ -197,7 +197,7 @@ RSpec.describe Dependabot::Nuget::MetadataFinder do
             stub_request(:get, nuget_url).to_return(status: 500)
             # it falls back to get search URL from the index, but it fails too
             stub_request(:get, "https://www.myget.org/F/exceptionless/api/v3/index.json").
-              to_return(status: 500, body: 'internal server error')
+              to_return(status: 500, body: "internal server error")
           end
 
           it { is_expected.to be_nil }
@@ -212,7 +212,7 @@ RSpec.describe Dependabot::Nuget::MetadataFinder do
               to_return(status: 200, body: fixture("nuspecs", "index.json"))
             # oops, we're a little overloaded
             stub_request(:get, "https://azuresearch-usnc.nuget.org/query?prerelease=true&q=microsoft.extensions.dependencymodel&semVerLevel=2.0.0").
-              to_return(status: 503, body: '')
+              to_return(status: 503, body: "")
           end
 
           it { is_expected.to be_nil }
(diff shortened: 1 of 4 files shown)


mctofu (commented):
This approach makes sense to me! Thanks for the detailed explanation!

Agree that it's best to avoid downloading the `.nupkg`.

jakecoffman:
@dependabot/reviewers This is ready for review now.

nuget/spec/dependabot/nuget/metadata_finder_spec.rb
```diff
@@ -171,6 +175,22 @@
...
+      context "that doesn't support .nuspec routes" do
+        before do
+          # registry doesn't support .nuspec route, so returns 404
+          stub_request(:get, nuget_url).to_return(status: 404)
+          # fallback begins by getting the search URL from the index
+          stub_request(:get, "https://www.myget.org/F/exceptionless/api/v3/index.json").
+            to_return(status: 200, body: fixture("nuspecs", "index.json"))
+          # next query for the package at the search URL returned
+          stub_request(:get, "https://azuresearch-usnc.nuget.org/query?prerelease=true&q=microsoft.extensions.dependencymodel&semVerLevel=2.0.0").
+            to_return(status: 200, body: fixture("nuspecs", "microsoft.extensions.depdencymodel-results.json"))
```
brrygrdn, line 188: ❓  Should we add some examples to cover these calls turning non-200 responses?

jakecoffman: Good idea, I added a couple. How do those look?

brrygrdn (approved):
👍🏻  I like this approach, :shipit:

PR #5018, opened 2022-04-20
Update Bundler from 2.3.10 to 2.3.12

deivid-rodriguez:
Includes a small change in monkeypatches to catch up with
https://github.com/rubygems/rubygems/commit/21c145cb232c5266b330d4f6df2c8c635930e23c.

https://github.com/rubygems/rubygems/blob/master/bundler/CHANGELOG.md#2311-april-7-2022
https://github.com/rubygems/rubygems/blob/master/bundler/CHANGELOG.md#2312-april-20-2022

Dependabot should give slightly better errors on Ruby version conflicts with this version.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -78,7 +78,7 @@ RUN apt-add-repository ppa:brightbox/ruby-ng \
   && apt-get install -y --no-install-recommends ruby2.7 ruby2.7-dev \
   && gem update --system 3.2.20 \
   && gem install bundler -v 1.17.3 --no-document \
-  && gem install bundler -v 2.3.10 --no-document \
+  && gem install bundler -v 2.3.12 --no-document \
   && rm -rf /var/lib/gems/2.7.0/cache/* \
   && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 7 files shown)


jurre (approved):
Thanks!

deivid-rodriguez:
Oooops, something actually broke here. Let me have a look.

deivid-rodriguez:
@jurre This one should be ok now. It actually complicated the monkey patch a bit, but I now understand better why you need to do this, so I hope to support this use case better upstream in the future. Also, and related to this, I introduced a regression that could affect Dependabot in RubyGems 3.3.12, but Dependabot still uses 3.2.20 and I'm fixing it ASAP, so it should be ok.

jurre:
> @jurre This one should be ok now. It actually complicated the monkey patch a bit, but I now understand better why you need to do this, so I hope to support this use case better upstream in the future. Also, and related to this, I introduced a regression that could affect Dependabot in RubyGems 3.3.12, but Dependabot still uses 3.2.20 and I'm fixing it ASAP, so it should be ok.

Cheers! We should really have a bit of documentation on those monkey patches because I also forget why we have them until they break and I need to fix them again :(

PR #5030, opened 2022-04-22
Bump underlying `ubuntu` to `22.04` LTS

jeffwidman:
`22.04` LTS dropped yesterday: https://ubuntu.com/blog/ubuntu-22-04-lts-released

I'm not sure of the GitHub internal infra, there may be security/infra reasons blocking
this upgrade that are not visible to external contributors, but thought I'd at least open
this PR to see what CI thinks.

For the Erlang issue, I had to install Erlang from Ubuntu Universe rather than Erlang Solutions.
Unfortunately that means we will probably lag on Erlang versions in the future.

While this isn't ideal from an Erlang perspective, this unblocks the
`22.04` upgrade, which is a bigger deal for the other ecosystems.

I'm hopeful this is only a temporary state of affairs and someone from
the community steps up to provide a PR switching to installing from
either the Erlang Solutions PPA or the RabbitMQ PPA.

Related:
* https://github.com/esl/packages/issues/15
* https://github.com/dependabot/dependabot-core/pull/7865
* https://erlangforums.com/t/erlang-solutions-apt-package-for-otp-25/1552/1
* https://erlangforums.com/t/the-eef-is-looking-for-volunteers-to-take-over-esls-build-packages/2238/1

### DIFF
--- a/Dockerfile.updater-core
+++ b/Dockerfile.updater-core
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04
+FROM ubuntu:22.04
 
 LABEL org.opencontainers.image.source="https://github.com/dependabot/dependabot-core"
 
@@ -97,7 +97,7 @@ COPY --from=ruby:3.1.4-bullseye --chown=dependabot:dependabot /usr/local /usr/lo
 # This way other projects that import this library don't have to futz around with installing new / unexpected bundler versions.
 ARG BUNDLER_V2_VERSION=2.4.17
 
-# We had to explicitly bump this as the bundled version `0.2.2` in ubuntu 20.04 has a bug.
+# We had to explicitly bump this as the bundled version `0.2.2` in ubuntu 22.04 has a bug.
 # Once Ubuntu base image pulls in a new enough yaml version, we may not need to
 # explicitly manage this. However, if we do opt to pull it back out, see all changes
 # required in https://github.com/dependabot/dependabot-core/pull/7112
--- a/Dockerfile.updater-core
+++ b/Dockerfile.updater-core
@@ -90,7 +90,7 @@ WORKDIR $DEPENDABOT_HOME/dependabot-updater
 
 # Install Ruby from official Docker image
 # When bumping Ruby minor, need to also add the previous version to `bundler/helpers/v{1,2}/monkey_patches/definition_ruby_version_patch.rb`
-COPY --from=ruby:3.1.4-bullseye --chown=dependabot:dependabot /usr/local /usr/local
+COPY --from=ruby:3.1.4-bookworm --chown=dependabot:dependabot /usr/local /usr/local
 
 # When bumping Bundler, need to also regenerate `updater/Gemfile.lock` via `bundle update --lock --bundler`
 # Generally simplest to match the bundler version to the one that comes by default with whatever Ruby version we install.
--- a/swift/Dockerfile
+++ b/swift/Dockerfile
@@ -25,7 +25,7 @@ USER dependabot
 
 # https://www.swift.org/download/
 ARG SWIFT_VERSION=5.8.1
-ARG SWIFT_UBUNTU_VERSION=ubuntu20.04
+ARG SWIFT_UBUNTU_VERSION=ubuntu22.04
 
 RUN if [ "$TARGETARCH" = "arm64" ]; then SWIFT_UBUNTU_VERSION="${SWIFT_UBUNTU_VERSION}-aarch64"; fi \
   && SWIFT_SHORT_UBUNTU_VERSION=$(echo $SWIFT_UBUNTU_VERSION | tr -d .) \
--- a/hex/Dockerfile
+++ b/hex/Dockerfile
@@ -3,22 +3,29 @@ RUN apt-get update \
   && apt-get upgrade -y \
   && apt-get install -y --no-install-recommends \
     gnupg2
-# Install Erlang and Elixir
-ENV PATH="$PATH:/usr/local/elixir/bin"
+
+# Install Erlang
+# Note: Currently we install from Ubuntu PPA to unblock upgrading to Ubuntu 22.04, but we'd be happy to accept a PR
+# switching to a more up to date PPA. See also:
+#  * https://github.com/esl/packages/issues/15
+#  * https://github.com/dependabot/dependabot-core/pull/7865
+#  * https://erlangforums.com/t/erlang-solutions-apt-package-for-otp-25/1552/1
+#  * https://erlangforums.com/t/the-eef-is-looking-for-volunteers-to-take-over-esls-build-packages/2238/1
+ARG ERLANG_MAJOR_VERSION=24
+ARG ERLANG_VERSION=1:${ERLANG_MAJOR_VERSION}.2.1+dfsg-1ubuntu0.1
+RUN apt-get update \
+  && apt-get install -y --no-install-recommends \
+    erlang=${ERLANG_VERSION}
+
+# Install Elixir
 # https://github.com/elixir-lang/elixir/releases
 ARG ELIXIR_VERSION=v1.14.4
 ARG ELIXIR_CHECKSUM=b5705275d62ce3dae172d25d7ea567fffdabb45458abeb29c0189923b907367c
-ARG ERLANG_MAJOR_VERSION=24
-ARG ERLANG_VERSION=1:${ERLANG_MAJOR_VERSION}.2.1-1
-RUN curl -sSLfO https://packages.erlang-solutions.com/erlang-solutions_2.0_all.deb \
-  && dpkg -i erlang-solutions_2.0_all.deb \
-  && apt-get update \
-  && apt-get install -y --no-install-recommends esl-erlang=${ERLANG_VERSION} \
-  && curl -sSLfO https://github.com/elixir-lang/elixir/releases/download/${ELIXIR_VERSION}/elixir-otp-${ERLANG_MAJOR_VERSION}.zip \
+RUN curl -sSLfO https://github.com/elixir-lang/elixir/releases/download/${ELIXIR_VERSION}/elixir-otp-${ERLANG_MAJOR_VERSION}.zip \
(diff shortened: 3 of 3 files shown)


jeffwidman:
Looks like this is all blocked until https://ppa.launchpadcontent.net/brightbox/ruby-ng/ubuntu adds support for `jammy`.

mattt:
@jeffwidman I just merged #5356, which should unblock us from upgrading to LTS. Can you rebase and try again with this PR?

jeffwidman:
Thanks @mattt, just saw your comment.

Rebased!

.github/workflows/docker.yml
```diff
@@ -1,6 +1,6 @@
 name: Push docker images
 env:
-  BASE_IMAGE: "ubuntu:18.04"
```
jeffwidman, line 3: I bumped this to `20.04` in #5432 as an intermediate step until this lands.

I wish there was a way to programmatically pull this from the `Dockerfile`, but not coming up with any easy solutions... everything I think of seems very brittle/hacky. Anyone else got an idea?

jeffwidman:
CI failures are because Ubuntu 22.04 uses OpenSSL 3.0, which requires Ruby >= 3.1...

The workaround is to install `openssl` lib manually before building Ruby.

Further background info here:
* https://github.com/ruby/setup-ruby/issues/333#issuecomment-1138406377
* https://github.com/rbenv/ruby-build/pull/1974 - note that we build Ruby manually, not using rbenv, but I found this thread useful for describing the problem.

Dockerfile
```diff
@@ -35,6 +35,8 @@ RUN apt-get update \
     unzip \
     locales \
     openssh-client \
+    # openssl is required to build Rubies < 3.1 on Ubuntu 22.04
+    openssl \
```
jeffwidman, line 39: This will fail because it simply installs `openssl` 3.x... when we need an older version of `openssl`, which simply isn't available via `apt`

jeffwidman:
So at this point, due to the problems with Ruby < 3.1 requiring an older OpenSSL than Ubuntu 22.04 ships, we have several options:
1. manually compile the relevant OpenSSL lib, and then somehow tell our downloaded Ruby how to use it... example: https://github.com/rvm/rvm/issues/5209#issuecomment-1192650802
2. Manually add a PPA with a pre-compiled OpenSSL ([example](https://github.com/rvm/rvm/issues/5209#issuecomment-1134927685)), downgrade libSSL, then use that... feels very brittle though... esp given https://bugs.ruby-lang.org/issues/18658#note-2.
3. bump to Ruby 3.1, we need to do this anyway at some point, not sure how much work it is, or if we'd run into problems with GitHub internal infra... but we might be okay on the infra part since we're silo'd into our own docker image, IDK.
4. switch to using `rbenv/ruby-build` since at first glance it will "just work" due to https://github.com/rbenv/ruby-build/pull/1974. However, we originally went with `ruby-install` instead of `ruby-build` for [some good reasons](https://github.com/dependabot/dependabot-core/pull/5356).
5. Wait a while, and see what shakes out further in the Ruby ecosystem... most of the open source discussions I've seen are within the past month, and I have a hunch waiting another 3-6 months will result in a bit more clarity/standardization on how to deal with this problem.

Everything I've read so far makes it seem very brittle/painful to manage old OpenSSL versions, so I vote for working towards option 3 which will implicitly result in also picking option 5.

Thoughts?

jurre:
I'm all for bumping the ruby version, @mattt's work on using ruby-install unblocks us, I think that's the main reason that we haven't done it yet, it should be a doable upgrade from my experience.

jeffwidman:
Sounds good to me, I opened https://github.com/dependabot/dependabot-core/pull/5447 to bump to 3.1.

I think my main concern was that something internal within GitHub tooling that we may use now or in the near future may not be ready to support Ruby 3.1... but I suppose we can always try it and if we run into problems we can revert or push those teams to start providing support for 3.1... Anyway, we'll see how that PR goes.

jurre:
> GitHub tooling that we may use now or in the near future may not be ready to support Ruby 3.

Most internal things are on the latest or even pre-release versions of ruby and rails, dependabot, a bot that updates your dependencies, being on older versions is a cruel paradox

jeffwidman:
Current CI [failure](https://github.com/dependabot/dependabot-core/runs/7731672757?check_suite_focus=true#step:3:6690):
```
https://github.com/dependabot/dependabot-core/runs/7731672757?check_suite_focus=true
```

Found this [relevant issue](https://github.com/elixir-lang/elixir-lang.github.com/issues/1575#issuecomment-1147535797), but no obvious easy solution.

deivid-rodriguez:
I'm going to tentatively reopen this because I think it was closed automatically by mistake!

jeffwidman:
Note to self: Also bump https://github.com/dependabot/dependabot-core/blob/9e15f164370b9baa68acfbb23cd5729a8b26c8b6/.github/workflows/docker-fork-releases.yml#L4 and any other workflows where we started pinning the image since this PR was originally opened...

jeffwidman:
I requested an update on the upstream issue:
https://github.com/esl/packages/issues/15#issuecomment-1296378191

jeffwidman:
> I requested an update on the upstream issue: [esl/packages#15 (comment)](https://github.com/esl/packages/issues/15#issuecomment-1296378191)

We may be able to continue using the Erlang PPA after all:
* https://github.com/esl/packages/issues/15#issuecomment-1446612343

deivid-rodriguez (approved):
Awesome job!

Dockerfile.updater-core
```diff
@@ -90,14 +90,14 @@ WORKDIR $DEPENDABOT_HOME/dependabot-updater
 
 # Install Ruby from official Docker image
 # When bumping Ruby minor, need to also add the previous version to `bundler/helpers/v{1,2}/monkey_patches/definition_ruby_version_patch.rb`
-COPY --from=ruby:3.1.4-bullseye --chown=dependabot:dependabot /usr/local /usr/local
+COPY --from=ruby:3.1.4-bookworm --chown=dependabot:dependabot /usr/local /usr/local
```
jakecoffman, line 93: We could base the image off of the ruby image and delete this line, and then get Dependabot bumps for it.

jakecoffman:
Should we update the [Swift Dockerfile](https://github.com/dependabot/dependabot-core/blob/3fca7d25f98d6a881e5e5968b63d4dac855d75a8/swift/Dockerfile#L28) to install the 22.04 version?

Dockerfile.updater-core
```diff
@@ -90,14 +90,14 @@ WORKDIR $DEPENDABOT_HOME/dependabot-updater
...
 
 # When bumping Bundler, need to also regenerate `updater/Gemfile.lock` via `bundle update --lock --bundler`
 # Generally simplest to match the bundler version to the one that comes by default with whatever Ruby version we install.
 # This way other projects that import this library don't have to futz around with installing new / unexpected bundler versions.
 ARG BUNDLER_V2_VERSION=2.4.17
 
-# We had to explicitly bump this as the bundled version `0.2.2` in ubuntu 20.04 has a bug.
+# We had to explicitly bump this as the bundled version `0.2.2` in ubuntu 22.04 has a bug.
```
abdulapopoola, line 100: Does the new version of Ubuntu handle the yaml parsing now? And can we therefore undo this [PR](https://github.com/dependabot/dependabot-core/pull/7112/files)

abdulapopoola:
[Would this need to be updated too?](https://github.com/dependabot/dependabot-core/blob/c46977e196c9ebc7395c716e7ce2eeb068ac9f3e/docker/spec/dependabot/docker/file_parser_spec.rb#L999)

Dockerfile.updater-core
jeffwidman: I wish we could, but we'll need Ubuntu `23.04` before we get a new enough version:
 * https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=libyaml-0-2&searchon=names

We have a :dependabot: PR to bump to `23.04`, but probably should land this PR first, and then also have a conversation about do we even want to bump to non-LTS versions:
 * https://github.com/dependabot/dependabot-core/pull/6561

swift/Dockerfile
```diff
@@ -25,7 +25,7 @@ USER dependabot
 
 # https://www.swift.org/download/
 ARG SWIFT_VERSION=5.8.1
-ARG SWIFT_UBUNTU_VERSION=ubuntu20.04
+ARG SWIFT_UBUNTU_VERSION=ubuntu22.04
```
jeffwidman, line 28: Great catch @jakecoffman , this crept in the last few months, and I completely forgot to re-grep for any new `20.04` references.

jeffwidman:
> [Would this need to be updated too?](https://github.com/dependabot/dependabot-core/blob/c46977e196c9ebc7395c716e7ce2eeb068ac9f3e/docker/spec/dependabot/docker/file_parser_spec.rb#L999)

No, this is just testing a docker tag. In fact better if it test for a tag that isn't our image just to ensure there's no coupling.

> [And this spec file too?](https://github.com/dependabot/dependabot-core/blob/c46977e196c9ebc7395c716e7ce2eeb068ac9f3e/docker/spec/fixtures/kubernetes/yaml/multiple-resources.yaml#L18)

Same as above, this is just a spec of a user's manifest... so _their_ Ubuntu version shouldn't matter relative to _our_ Ubuntu version.

That said, I _really_ appreciate you and Jake for grep'ing for other references and double-checking this. I completely forgot to look for any others that crept in over the past year.

Dockerfile.updater-core
jeffwidman: It's an interesting idea for sure, but I'd rather do it after this PR lands. I considered switching to `--from=ruby:3.1.4` since today it's the same as `ruby:3.1.4-bookworm` but we used to do that and ran into problems when it switched its base debian image under our feet to one that conflicted with our version of Ubuntu used elsewhere... it'd probably be okay if it were the base image, but still let's postpone looking at that to a later PR and here just catch us up to a newer Ubuntu.

jeffwidman:
And a year and a half later I can finally hit the 🟢 button! 😁

PR #5031, opened 2022-04-22
Clarify comment and remove TODO

jeffwidman:
The current version of the warning:
```
warning: parser/current is loading parser/ruby27, which recognizes2.7.6-compliant syntax, but you are running 2.7.5.
```

So I removed the TODO comment because it's outdated. 
But I left the actual exclusion in because it will be a recurring issue from time to time because https://ppa.launchpadcontent.net/brightbox/ruby-ng/ubuntu can be a little slow to update...
for example see various discussions on the mailing list that sometimes get no response from
the maintainers: https://groups.google.com/g/brightbox-ruby-ubuntu-packaging

### DIFF
--- a/common/spec/warning_monkey_patch.rb
+++ b/common/spec/warning_monkey_patch.rb
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
 ALLOW_PATTERNS = [
-  # Ignore parser warnings for ruby 2.7 minor version mismatches
-  # TODO: Fix these by upgrading to ruby 2.7.3 (requires ubuntu upgrade)
+  # Ignore parser warnings for ruby 2.7 minor version mismatches.
+  # This is a recurring issue that occurs whenever the parser gets
+  # ahead of our installed ruby version.
   %r{parser/current is loading parser/ruby27},
   /2.7.\d-compliant syntax, but you are running 2.7.\d/,
   %r{whitequark/parser}


common/spec/warning_monkey_patch.rb
```diff
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
 ALLOW_PATTERNS = [
-  # Ignore parser warnings for ruby 2.7 minor version mismatches
-  # TODO: Fix these by upgrading to ruby 2.7.3 (requires ubuntu upgrade)
+  # Ignore parser warnings for ruby 2.7 minor version mismatches.
+  # This is a recurring issue that occurs whenever the parser gets
+  # ahead of our installed ruby version.
   %r{parser/current is loading parser/ruby27},
   /2.7.\d-compliant syntax, but you are running 2.7.\d/,
```
jeffwidman, line 8: I am unclear why these first two lines are separate patterns, when the matching error line is a single line:
```
parser/current is loading parser/ruby27, which recognizes2.7.6-compliant syntax, but you are running 2.7.5.
```

landongrindheim: Just did a quick Google search for "compliant syntax, but you are running " and saw [an issue](https://github.com/whitequark/parser/issues/231) wherein someone had the message displayed as:

```ruby
warning: parser/current is loading parser/ruby22, which recognizes
warning: 2.2.3-compliant syntax, but you are running 2.2.4.
```

I'd expect `ALLOW_PATTERNS.none? { |pattern| pattern =~ message }` to nicely handle that case and the [case you quoted](https://github.com/dependabot/dependabot-core/pull/5031/files#r856532794) 🙂

landongrindheim (approved):

common/spec/warning_monkey_patch.rb
jeffwidman: Ah, that makes perfect sense, thank you @landongrindheim!

deivid-rodriguez:
Looks good to me!

I'm wondering, given how long it takes for brightbox to package ruby, whether it would make sense to move to something else.

jeffwidman:
> I'm wondering, given how long it takes for brightbox to package ruby, whether it would make sense to move to something else.

@deivid-rodriguez I had the same question... but didn't know offhand where else to pick it up from. Beyond speed of updates, from a security perspective, we need to be especially careful where we pick this up from.

deivid-rodriguez:
@jeffwidman Absolutely!

If I recall correctly, brightbox Ruby was born at the time because Linux distributions were too slow updating Ruby. However, now the situation seems exactly the opposite, for example, Ruby 3.0 is already available in Ubuntu 22.04.

PR #4976, opened 2022-04-07
Upgrade Python to 3.10.4 and support newer releases

henrikhorluck:

PR #5033, opened 2022-04-24
Support python 3.9.{11,12} and 3.10.4

JacobHayes:
Basing on https://github.com/dependabot/dependabot-core/commit/7a27f952a9148fe8a4ed9fa20011c43dd3b84130 - are these all I need to change?

- https://www.python.org/downloads/release/python-3911/
- https://www.python.org/downloads/release/python-3912/
- https://www.python.org/downloads/release/python-3104/


henrikhorluck:
Duplicate of https://github.com/dependabot/dependabot-core/pull/4976

It might be worth considering changing the dependency on `pyenv`, e.g. changing to `master` instead of specific version, as they _have_ already added support for [`3.10.4`](https://github.com/pyenv/pyenv/pull/2295), just not released it. They also have an [issue](https://github.com/pyenv/pyenv/issues/2319) regarding the lack of release[^0] with 23 thumbs-up at time of writing.

I ran into the exact same problem as you @JacobHayes which was why I created the existing PR, and is hindering us from getting dependabot PR's to fix e.g. recent [Django CVEs](https://www.djangoproject.com/weblog/2022/apr/11/security-releases/).

[^0]: I am not familiar with `python-build`, so that is purely by assumption

JacobHayes:
Thanks for handling this one so thoroughly! I'll go ahead and close.

> BTW: The update instruction were documented in the python-README

Wow, that would have been easier than trawling the old commits. 😅

PR #4132, opened 2021-08-12
Bump bundler from 2.2.20 to 2.2.25

jurre:
https://github.com/rubygems/rubygems/blob/master/bundler/CHANGELOG.md#2225-july-30-2021

Something we _might_ get some support questions on is:
https://github.com/rubygems/rubygems/pull/4647 but overall I think
that's a very good change! The GEM sources not being split has bitten
our users in the past, so happy to see that change.

In a recent version of bundler, passing a major version like
`BUNDLER_VERSION=2 bundle --version` no longer works.

I suspect this may have happened here: https://github.com/rubygems/rubygems/pull/4750/files

We now need to pass the exact version of bundler we're using, which is a
bit cumbersome as you can see from the diff, and it also means we'll
have to change a few more lines every time we update a version of
bundler.

I don't think the current behavior we were relying on was documented, so
I'm not sure if we can open an issue with the Bundler team.

This seems like the most straight forward fix to me, but I am definitely
open to other suggestions

PR #5044, opened 2022-04-25
Set BUNDLER_VERSION to the specific installed version in native helpers

brrygrdn:
We've encountered an issue in the move forward to bundler 2.3.12 where our approach to setting the `BUNDLER_VERSION` in native helpers now behaves differently.

Using `v0.182.0`, you can observe the behaviour like so:

```
$ cd bundler/helpers/v1
$ bundle install
$ BUNDLER_VERSION=1.14.6 bundler -v
> Bundler version 1.17.3
```

On main the behaviour is like so:

```
$ cd bundler/helpers/v1
$ bundle install
$ BUNDLER_VERSION=1.14.6 bundler -v
> Bundler version 2.3.12
```

We updated this behaviour in https://github.com/dependabot/dependabot-core/pull/4132 when we noted that passing in just the major version, i.e. `BUNDLER_VERSION=1` no longer worked and we started using the _detected_ version instead, which bundler would previously 'guess' at a near match.

### The change

This modifies our `native_helper` function to set the BUNDLER_VERSION to the actual version of v1.x or v2.x we have installed according to the detected major version.

In retrospect, this feels a little bit like it was working by accident and we were relying on bundler doing some magic for us, so I think this is a more correct approach.

### DIFF
--- a/bundler/lib/dependabot/bundler/native_helpers.rb
+++ b/bundler/lib/dependabot/bundler/native_helpers.rb
@@ -36,7 +36,7 @@ module Dependabot
       def self.run_bundler_subprocess(function:, args:, bundler_version:, options: {})
         # Run helper suprocess with all bundler-related ENV variables removed
         bundler_major_version = bundler_version.split(".").first
-        helpers_path = versioned_helper_path(bundler_version: bundler_major_version)
+        helpers_path = versioned_helper_path(bundler_major_version)
         ::Bundler.with_original_env do
           command = BundleCommand.
                     new(options[:timeout_per_operation_seconds]).
@@ -47,7 +47,7 @@ module Dependabot
             args: args,
             env: {
               # Bundler will pick the matching installed major version
-              "BUNDLER_VERSION" => bundler_version,
+              "BUNDLER_VERSION" => installed_bundler_version(major_bundler_version),
               "BUNDLE_GEMFILE" => File.join(helpers_path, "Gemfile"),
               # Prevent the GEM_HOME from being set to a folder owned by root
               "GEM_HOME" => File.join(helpers_path, ".bundle")
@@ -61,8 +61,15 @@ module Dependabot
         end
       end
 
-      def self.versioned_helper_path(bundler_version:)
-        File.join(native_helpers_root, "v#{bundler_version}")
+      def self.versioned_helper_path(bundler_major_version)
+        File.join(native_helpers_root, "v#{bundler_major_version}")
+      end
+
+      # Maps the major version unto the specific version we have installed
+      def self.installed_bundler_version(bundler_major_version)
+        return Helpers::V1 if bundler_major_version == '1'
+
+        Helpers::V2
       end
 
       def self.native_helpers_root
--- a/bundler/lib/dependabot/bundler/native_helpers.rb
+++ b/bundler/lib/dependabot/bundler/native_helpers.rb
@@ -67,7 +67,7 @@ module Dependabot
 
       # Maps the major version unto the specific version we have installed
       def self.installed_bundler_version(bundler_major_version)
-        return Helpers::V1 if bundler_major_version == '1'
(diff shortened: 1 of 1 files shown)


bundler/lib/dependabot/bundler/native_helpers.rb
```diff
@@ -61,8 +61,15 @@ def self.run_bundler_subprocess(function:, args:, bundler_version:, options: {})
...
 
-      def self.versioned_helper_path(bundler_version:)
-        File.join(native_helpers_root, "v#{bundler_version}")
+      def self.versioned_helper_path(bundler_major_version)
+        File.join(native_helpers_root, "v#{bundler_major_version}")
+      end
+
+      # Maps the major version unto the specific version we have installed
+      def self.installed_bundler_version(bundler_major_version)
+        return Helpers::V1 if bundler_major_version == '1'
```
landongrindheim, line 70: 🤖🚓 

````suggestion
        return Helpers::V1 if bundler_major_version == "1"
````

Comment (brrygrdn) [inline bundler/lib/dependabot/bundler/native_helpers.rb:70]:
```suggestion
        return Helpers::V1 if bundler_major_version == "1"
```

landongrindheim (approved):
Looks reasonable to me!

bundler/lib/dependabot/bundler/native_helpers.rb
```diff
@@ -47,7 +47,7 @@ def self.run_bundler_subprocess(function:, args:, bundler_version:, options: {})
             args: args,
             env: {
               # Bundler will pick the matching installed major version
-              "BUNDLER_VERSION" => bundler_version,
+              "BUNDLER_VERSION" => installed_bundler_version(major_bundler_version),
```
landongrindheim, line 50:
````suggestion
              "BUNDLER_VERSION" => installed_bundler_version(bundler_major_version)
````

Nishnha (approved):
Nice spot! This fix seems fine to me.

I think in a future PR we can improve on how we use use the env variable that we set here:

The "BUNDLER_VERSION" env variable is passed in when executing a the `run_helper_subprocess` command, [but it's not an officially supported/documented feature](https://github.com/rubygems/rubygems/issues/3978).

Maybe we should swap it for using [rubygem's `bundle` version prefixes instead](https://github.com/rubygems/rubygems/issues/3978#issuecomment-701285069)?

brrygrdn:
@Nishnha yeah, this is very much a minimal patch possible - I think https://github.com/dependabot/dependabot-core/pull/4973 represents a good candidate for a 'proper' fix.

deivid-rodriguez:
Hi folks, sorry about this issue.

I believe it must've been [the rubygems upgrade](https://github.com/dependabot/dependabot-core/pull/5035/commits/34233698350c0b954f7fd3dbf3d48d743d3f4799) that actually caused this, since it's where the code to activate the Bundler gem lives.

In particular, I reviewed the changelog and it must've been https://github.com/rubygems/rubygems/pull/5181. Indeed, although the goal was to remove hard errors when activating Bundler, this "magic" was also removed from rubygems and now `BUNDLER_VERSION` only activates the exact version if present but applies other filtering using the major version only.

This change looks good to adapt to that, and I think #4973 should also fix this.

brrygrdn:
@deivid-rodriguez Thanks for checking in - we've been testing this fix out but it seems like while it changes behaviour it doesn't remediate the issue downstream.

Part of the issue is a lack of an integration test in core that checks the functionality thoroughly so we couldn't really detect this until it was added to an internal integration suite. We're going to rollback the rubygems upgrade as we've confirmed that does resolve the issue and then regroup on how we can roll forward.

The long tail while be figuring out how we can move some testing around this earlier in the feedback loop so we don't have to revert contributions, which feels bad.

deivid-rodriguez:
Don't worry about reverting, I'm glad you have further checks down to the road to detect this kind of issue on time.

PR #5048, opened 2022-04-26
Rollback RubyGems from 3.3.11 to 3.2.20

mctofu:
We attempted to resolve an issue around Bundler version detection for projects using Bundler 1.x in https://github.com/dependabot/dependabot-core/pull/5044, but this wasn't sufficient to fix the issue.

We've verified that the issue is resolved by reverting our recent RubyGems update so we're reverting #5035 to unblock releasing forward and we will reintroduce the RubyGems upgrade once we have determined root cause.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -76,7 +76,7 @@ ENV PATH="$BUNDLE_BIN:$PATH:$BUNDLE_PATH/bin"
 RUN apt-add-repository ppa:brightbox/ruby-ng \
   && apt-get update \
   && apt-get install -y --no-install-recommends ruby2.7 ruby2.7-dev \
-  && gem update --system 3.3.11 --no-document \
+  && gem update --system 3.2.20 \
   && gem install bundler -v 1.17.3 --no-document \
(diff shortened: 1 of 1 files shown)


landongrindheim (approved):

brrygrdn:
On hold on merging this while we perform some additional verification.

Dockerfile
```diff
@@ -76,7 +76,7 @@ ENV PATH="$BUNDLE_BIN:$PATH:$BUNDLE_PATH/bin"
 RUN apt-add-repository ppa:brightbox/ruby-ng \
   && apt-get update \
   && apt-get install -y --no-install-recommends ruby2.7 ruby2.7-dev \
-  && gem update --system 3.3.11 --no-document \
```
jeffwidman, line 79: Per https://github.com/dependabot/dependabot-core/pull/5035#issue-1213736168, the `--no-document` flag is faster, so we should try adding that back in even with the held-back version.

I'll open a PR.

PR #5065, opened 2022-04-28
fix unsafe directory git issues during updates

jakecoffman:
When `with_git_configured` is executed it stashes the existing `.gitconfig` and loses any `safe.directory` settings which is causing update issues in production.

This fix grabs the `safe.directory` settings before stashing, and writes it in the new file so the update succeeds.


common/lib/dependabot/shared_helpers.rb
```diff
@@ -190,6 +191,15 @@ def self.configure_git_to_use_https_with_credentials(credentials)
         allow_unsafe_shell_command: true
       )
 
+      # see https://github.blog/2022-04-12-git-security-vulnerability-announced/
+      safe_directories ||= []
+      safe_directories.each do |path|
+        run_shell_command(
+          "git config --global --add safe.directory #{path}",
+          allow_unsafe_shell_command: true
```
mctofu, line 199: I see we tend to set this when using path parameters. Do you know what breaks if we don't allow this?

mctofu (approved):

common/lib/dependabot/shared_helpers.rb
jakecoffman: It seems to be good without and will be safer to avoid potential shell injection attacks, so I removed it.

brrygrdn (approved):

PR #5072, opened 2022-04-29
nuget: fix PR missing commits in message when using private registry

jakecoffman:
We reverted #5002 because it introduced some new errors, so this is #5002 plus some additional error handling and tests.

I left in some TODOs to ask the question: should we log? I haven't seen any other log messages in core yet so I wanted to see if we could log at all. If we are eating exceptions and don't log, we are going to spend more time investigating issues that end up to be working as expected.

Ruby style feedback is welcome as well!


nuget/lib/dependabot/nuget/metadata_finder.rb
```diff
@@ -12,7 +12,73 @@ class MetadataFinder < Dependabot::MetadataFinders::Base
...
+
+        # Fallback to getting source from the search result's projectUrl or licenseUrl.
+        # GitHub Packages doesn't support getting the `.nuspec`, switch to getting
+        # that instead once it is supported.
+        src_repo_from_project
+      rescue StandardError
+        # At this point in the process the PR is ready to be posted, we tried to gather commit
+        # and release notes, but have encountered an exception. So let's eat it and log it
+        # since it's better to have a PR with no info than error out.
+        # TODO how do we log this?
```
brrygrdn, line 26: We don't have a pattern for logging / metrics in core right now as we don't have a way to inject our backend services for either into core.

This is something we should add to improve observability though, but for now swallowing an error mysteriously is consistent behaviour 😅

nuget/lib/dependabot/nuget/metadata_finder.rb
```diff
@@ -12,7 +12,73 @@ class MetadataFinder < Dependabot::MetadataFinders::Base
...
+            source = Source.from_url(search_result.fetch("projectUrl"))
+            return source if source
+          end
+          if search_result.key?("licenseUrl")
+            source = Source.from_url(search_result.fetch("licenseUrl"))
+            return source if source
+          end
+        end
+        # failed to find a source URL
+        nil
```
brrygrdn, line 81: Should we add a `rescue JSON::ParserError` to this method or move the rescue up into `src_repo_from_project` to handle both `extract_` methods?

I'm just thinking that while a registry switching from JSON to not-JSON between endpoints would be surprising, it could happen if we hit some kind of error that returned arbitrary content from the remote stack.

brrygrdn (commented):
One question, otherwise this LGTM

brrygrdn (approved):
🚀

PR #5073, opened 2022-04-30
bin/docker-dev-shell: Teach `DOCKER` env variable

bb010g:
`DOCKER=podman bin/docker-dev-shell` now works.

`LOCAL_CODE_DIR` eases running `bin/docker-dev-shell` with MSYS2 Bash.

### DIFF
--- a/bin/docker-dev-shell
+++ b/bin/docker-dev-shell
@@ -2,6 +2,7 @@
 
 set -e
 
+DOCKER="${DOCKER:=docker}"
 IMAGE_NAME="${IMAGE_NAME:=dependabot/dependabot-core-development}"
 CONTAINER_NAME="${CONTAINER_NAME:=dependabot-core-development}"
 DOCKERFILE="Dockerfile.development"
@@ -15,7 +16,7 @@ export DOCKER_BUILDKIT=1
 OPTS=$(getopt -o hr --long help,rebuild -n 'parse-options' -- "$@")
 # shellcheck disable=SC2181
 if [ $? != 0 ]; then
-  echo "failed parsing options" >&2
+  printf '%s\n' "failed parsing options" >&2
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


jeffwidman:
👋 Thanks for this and sorry for the delay reviewing. It would be nice to support `podman` etc.

However, I see three unrelated changes here:
1. `$DOCKER`
2. switch `echo` -> `printf`
3. `$LOCAL_CODE_DIR`

Since they are unrelated and do impact , so can you break them out into separate PR's? It's just easier to discuss them that way, as I had a couple of questions about items 2 and 3.

bb010g:
`echo "$msg"` to `printf '%s\n' "$msg"` shouldn't result in different functionality under Bash (`echo`'s behavior with regard to escape sequences isn't portable), except that when `$msg` starts with `-` it'll print as expected and not try to be parsed as an option. It's good shell script hygiene. (See [Bash Pitfall #14](https://mywiki.wooledge.org/BashPitfalls#echo_.24foo).)

The introduction of `$LOCAL_CODE_DIR` to override `$(pwd)` is because MSYS2's Bash (which is an easy way to run this without WSL2) prints a path on `pwd` that Podman's native Windows build can't deal with. I'm guessing this is similar to what that other pull request ran into. I don't know of a way to solve this without giving the user the ability to directly override this path.

jeffwidman:
Thanks, I'm actually aware of most of ☝️... but they're still unrelated changes, so at the very least they should be separate commits, and by that time it's easier to review/discuss as separate PR's.

Additionally, from time to time we have to revert commits/PRs for various unexpected reasons, and it's a lot nicer if we can keep that revert scoped to a single logical change so that we don't have a side effect of rolling back other changes that are still fine to keep.

To be clear, I'm 👍 on all these changes so far as I can tell from a quick glance at this PR... so should be straightforward, but they do need to be broken up.

jeffwidman:
@bb010g Did you see my comment ☝️ ?

Any interest in moving this forward by splitting it up?

PR #5078, opened 2022-05-02
fix debase build errors

jakecoffman:
Reverts #4971

@pavera Unfortunately debase started failing to build in Codespaces. 

I noticed this on my machine a couple weeks ago as well and managed to work around it by changing debase versions and ruby versions. It seems others are having this issue as well: https://github.com/ruby-debug/debase/issues/98

I tried a couple different versions of debase and they all failed, so I guess we should roll this back for now? Any ideas we can try before we do this?

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -17,15 +17,6 @@
       ]
     },
     {
-      "name": "Debug Tests",
-      "type": "Ruby",
-      "request": "launch",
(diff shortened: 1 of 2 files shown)


pavera:
I don't really have any ideas.  Odd that it seems somewhat intermittent, but yeah, try again later.

pavera (approved):
🚀

PR #5085, opened 2022-05-03
remove multiplicative retries to improve system health

jakecoffman:
Excon has a builtin retry of 4 when using idempotent: true. The retries removed by this PR were multiplying the total retries done. So for NPM, with 2 retries a total of 12 retries are made.

We think 4 retries are probably enough! This will probably make some jobs succeed that were hitting the maximum wait limit since doing 12 retries can take quite a while. 

*Implementation note:*
BitBucketWIthRetries delegates to BitBucket and it seems like the intent was to not retry using BitBucket, so there I set idempotent to false with a note as to why. The other option is to remove BitBucketWithRetries entirely which I am not opposed to. Any opinions there?

### DIFF
--- a/cargo/lib/dependabot/cargo/update_checker/latest_version_finder.rb
+++ b/cargo/lib/dependabot/cargo/update_checker/latest_version_finder.rb
@@ -90,12 +90,6 @@ module Dependabot
           )
 
           @crates_listing = JSON.parse(response.body)
-        rescue Excon::Error::Timeout
-          retrying ||= false
-          raise if retrying
-
-          retrying = true
-          sleep(rand(1.0..5.0)) && retry
         end
 
         def wants_prerelease?
--- a/common/lib/dependabot/clients/bitbucket.rb
+++ b/common/lib/dependabot/clients/bitbucket.rb
@@ -160,7 +160,8 @@ module Dependabot
           url,
           user: credentials&.fetch("username", nil),
           password: credentials&.fetch("password", nil),
-          idempotent: true,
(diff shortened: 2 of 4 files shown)


cargo/lib/dependabot/cargo/update_checker/latest_version_finder.rb
```diff
@@ -90,12 +90,6 @@ def crates_listing
           )
 
           @crates_listing = JSON.parse(response.body)
-        rescue Excon::Error::Timeout
-          retrying ||= false
-          raise if retrying
-
-          retrying = true
-          sleep(rand(1.0..5.0)) && retry
```
brrygrdn, line 98: I'm 👍🏻  on removing this as I think a strategy of retrying a Timeout with a random backoff isn't terribly compelling - if it was a rate limit or something else I think that would make more sense.

brrygrdn (approved):
I think this change makes good sense in all modified cases, and we're still rescuing the things we want to rescue without accidental multiplication.

common/lib/dependabot/clients/bitbucket.rb
```diff
@@ -160,7 +160,8 @@ def get(url)
           url,
           user: credentials&.fetch("username", nil),
           password: credentials&.fetch("password", nil),
-          idempotent: true,
+          # Setting to false to prevent Excon retries, use BitbucketWithRetries for retries.
+          idempotent: false,
```
brrygrdn, line 164: 👍🏻  I'm good with the approach of making this version of the client _really_ not retry to be consistent with the naming of `BitbucketWithRetries`.

PR #5087, opened 2022-05-03
Prevent Dependabot from downgrading Dockerfiles if the repository doesn't have the tag yet

brrygrdn:
We've had some issues with Dependabot opening PRs to downgrade Dockerfiles if the repository's `tags/list` does not include the current version.

This is generally down to a bug in the Docker repository, or a project changing which repository it uses, but it does highlight that our logic for removing 'downgrades' when determining the `latest_version` isn't working as intended.


docker/lib/dependabot/docker/update_checker.rb
```diff
@@ -146,8 +146,7 @@ def fetch_latest_version(version)
           # Prune out any downgrade tags before checking for pre-releases
           # (which requires a call to the registry for each tag, so can be slow)
           candidate_tags = comparable_tags_from_registry(version)
-          non_downgrade_tags = remove_version_downgrades(candidate_tags, version)
-          candidate_tags = non_downgrade_tags if non_downgrade_tags.any?
+          candidate_tags = remove_version_downgrades(candidate_tags, version)
```
brrygrdn, line 149: ❓ None of the tests fail when I removed the logic to use _all_ `candidate_tags` if the `non_downgrade_tags` fail, is there a reason we would want to fail over to using all tags I'm not seeing?

Do we have a missing test case?

docker/spec/dependabot/docker/update_checker_spec.rb
```diff
@@ -415,6 +415,7 @@
 
     context "when the dependency has a namespace" do
       let(:dependency_name) { "moj/ruby" }
+      let(:version) { "2.4.0" }
```
brrygrdn, line 418: This test was working by accident due to the downgrade behaviour the `version` value being used was `17.10`, which is the Ubuntu version set at the top level in the test that remained a mystery guest here.

This should be a 2.4.0 to 2.4.2 upgrade to align with our other Ruby specs in this file.

docker/lib/dependabot/docker/update_checker.rb
mctofu: I can imagine there could be some tricky tags that might trip this up but it would be better to identify & handle those rather than let it continue without filtering?

mctofu (approved):

PR #5120, opened 2022-05-10
Validate before creating version objects in `Dependabot::NpmAndYarn::UpdateChecker::VersionResolver`

mattt:
The current implementation of `original_package_update_available?` and `types_update_available?` in `Dependabot::NpmAndYarn::UpdateChecker::VersionResolver` calls `Version.new` without first validating, which raises an exception.

This PR adds missing calls to `Version.correct?` and returns from these methods early.


pavera (approved):
👍

PR #1013, opened 2019-02-20
NPM: Handle 403 Forbidden when updating lockfiles

feelepxyz:
Seems to handle the case when we auth against artifactory and get
"Unable to authenticate, need: Basic realm="Artifactory Realm""

This seems to be caused by a trailing slash in the registry host as set
in config variables. When we write this out in npmrc we add another
trailing slash and this seems to cause the above error. Changing to a
single trailing slash returns a 403 Forbidden error.

PR #5124, opened 2022-05-10
Fix Dependabot ignoring scoped registries if the lockfile resolved uri contains a port number

brrygrdn:
Assuming a project has the following config:

**dependabot.yml**
```yml
version: 2
registries:
  private-npm:
    type: npm-registry
    url: example.com/artifactory/api/npm/example
    username: ${{secrets.USER}}
    password: ${{secrets.PASS}}
updates:
  - package-ecosystem: "npm"
    registries:
      - private-npm
    schedule:
      interval: "daily"
```

In their package.json they have a dependency from their private registry, which has a `resolved` URL in the form:

```
https://example.com:443/artifactory/api/npm/example/@example/example-client/-/@example-corp/example-client-1.0.0.tgz
```

This will result in confusing behaviour where we will correctly find the latest version in the private registry but fail to actually update the package with a `private_source_authentication_failure`.

### The problem

The `private_source_authentication_failure` is actually a misdirection, using the debugger I was able to output the raw error which was coming from npm [here](https://github.com/dependabot/dependabot-core/blob/main/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb#L57-L63)

```
npm ERR! code E404
npm ERR! 404 Not Found - GET https://registry.npmjs.org/@example%2fexample-client - Not found
npm ERR! 404 
npm ERR! 404  '@example/example-client@1.0.1' is not in this registry.
npm ERR! 404 You should bug the author to publish it (or use the name yourself!)
npm ERR! 404 
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.
```

This indicates that **we are actually calling the _public_ npm registry** and getting a 404 but then assume that it's a private registry problem [here](https://github.com/dependabot/dependabot-core/blob/main/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb#L409-L426) as this line is false:

```
UpdateChecker::RegistryFinder.central_registry?(reg)
```

### The fix

After investigating I discovered that no 'scoped registry' line was being added to the `.npmrc` file we generate on the fly when performing the update.

It should have had a line like so:
```
@example=https://example.com/artifactory/api/npm/example
```

The reason this wasn't being added is that due to a fix for a duplicate trailing slash in the past (https://github.com/dependabot/dependabot-core/pull/1013/files), we started adding a trailing slash to the `registry` value used in this method unintentionally:
https://github.com/dependabot/dependabot-core/blob/3428ef7720a4b68e5238f4299a1d096f59da81e6/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npmrc_builder.rb#L174-L198

The fix was to add the compulsory trailing slash later in the method, I've added test cases that verify this resolves the issue.

### Unrelated changes

I fixed the bug in fd4fb97 to make the red/green pass as clear as possible. The final commit cleans up some test churn where we expected the scoped registry lines to have the trailing `/` - this isn't necessary and we tolerate it as a side-effect of the original bug fix.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npmrc_builder.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npmrc_builder.rb
@@ -194,12 +194,7 @@ module Dependabot
           # Registry used for unscoped packages
           return if scopes.include?(nil)
 
-          # TODO: Remove this line
-          #
-          # We do not require the trailing slash for scopes, but we are currently added one
-          # by accident. This avoids us having to churn the tests to ensure the bug fix passes
-          registry_with_trailing_slash = registry.sub(%r{\/?$}, "/")
-          scopes.map { |scope| "@#{scope}:registry=https://#{registry_with_trailing_slash}" }
+          scopes.map { |scope| "@#{scope}:registry=https://#{registry}" }
         end
         # rubocop:enable Metrics/PerceivedComplexity
 
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npmrc_builder.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npmrc_builder.rb
@@ -48,7 +48,9 @@ module Dependabot
         end
 
         def global_registry # rubocop:disable Metrics/PerceivedComplexity
-          @global_registry ||=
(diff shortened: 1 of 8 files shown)


jakecoffman (approved):

pavera (approved):
🚀

PR #5132, opened 2022-05-11
[Maven] Fix some malformed memoizations

brrygrdn:
This PR collects up a pass on any methods that make `Excon.get` calls which attempt to memoize the outcome of requests.

I found a couple of cases where the pattern used wouldn't correctly memoize falsy values and fall through to making the request a second time.

From walking through the code I don't think these will have any performance implications unfortunately, I don't believe these methods are called more than once in any of our runtimes, but it could catch us out in future.

### DIFF
--- a/maven/lib/dependabot/maven/update_checker/version_finder.rb
+++ b/maven/lib/dependabot/maven/update_checker/version_finder.rb
@@ -156,7 +156,7 @@ module Dependabot
 
         def dependency_metadata(repository_details)
           repository_key = repository_details.hash
-          return @dependency_metadata[repository_key] if @dependency_metadata.has_key?(repository_key)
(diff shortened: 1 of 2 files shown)


maven/lib/dependabot/maven/update_checker/version_finder.rb
```diff
@@ -154,25 +155,28 @@ def released?(version)
         end
 
         def dependency_metadata(repository_details)
-          @dependency_metadata ||= {}
-          @dependency_metadata[repository_details.hash] ||=
```
brrygrdn, line 158: This memoization pattern works, but I found it hard to read so this is an opinionated change just to improve the readability and make this a little easier to maintain in the event that the `fetch_dependency_metadata` code was ever to return a falsy value rather than a blank string.

landongrindheim (approved):

PR #5139, opened 2022-05-13
skip CI workflow when creating release notes

jakecoffman:
The release process is already pretty lengthy, so this allows us to skip the CI step when cutting release notes. We could use admin powers to merge before CI is complete but that is a bad habit.


brrygrdn (approved):
This makes sense to me 👍🏻

mattt (approved):

mattt:
Good call, @jakecoffman!

PR #5143, opened 2022-05-16
v0.188.0

bdragon:
## v0.188.0, 16 May 2022

- Dwc0011/fix codecommit error (@dwc0011) [#4926](https://github.com/dependabot/dependabot-core/pull/4926)
- Upgrade Python to 3.10.4 and support newer releases (@henrikhorluck) [#4976](https://github.com/dependabot/dependabot-core/pull/4976)
- build(deps): bump npm from 6.14.16 to 6.14.17 in /npm_and_yarn/helpers [#5067](https://github.com/dependabot/dependabot-core/pull/5067)
- build(deps-dev): bump eslint from 8.13.0 to 8.15.0 in /npm_and_yarn/helpers [#5106](https://github.com/dependabot/dependabot-core/pull/5106)
- build(deps-dev): bump jest from 27.5.1 to 28.1.0 in /npm_and_yarn/helpers [#5105](https://github.com/dependabot/dependabot-core/pull/5105)
- build(deps-dev): bump phpstan/phpstan from 1.6.4 to 1.6.8 in /composer/helpers/v2 [#5126](https://github.com/dependabot/dependabot-core/pull/5126)
- build(deps-dev): bump phpstan/phpstan from 1.6.4 to 1.6.8 in /composer/helpers/v1 [#5127](https://github.com/dependabot/dependabot-core/pull/5127)
- build(deps-dev): update rubocop requirement from ~> 1.28.2 to ~> 1.29.1 in /common [#5136](https://github.com/dependabot/dependabot-core/pull/5136)
- skip CI workflow when creating release notes (@jakecoffman) [#5139](https://github.com/dependabot/dependabot-core/pull/5139)
- Strip UTF-8 BOM from file contents (@mattt) [#5129](https://github.com/dependabot/dependabot-core/pull/5129)

PR #5144, opened 2022-05-16
fix required checks not run during a release

jakecoffman:
We have "required checks" configured for the CI, and so if the CI never runs then the build will never pass. So this creates a workflow which runs during a release with identical names to the real CI workflow, but doesn't run the real tests.

We also thought it would be a good idea to skip the CodeQL step during a release since it probably won't detect any issues. The CodeQL isn't currently required so no codeql-analysis-release.yml required yet!

See https://github.com/dependabot/dependabot-core/pull/5143

### DIFF
--- a/.github/workflows/ci-release.yml
+++ b/.github/workflows/ci-release.yml
@@ -1,0 +1,38 @@
+name: CI
+on:
+  pull_request:
+    paths:
+      - 'CHANGELOG.md'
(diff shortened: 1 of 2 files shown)


brrygrdn (approved):
Makes sense to me, thanks for adding this 👍🏻

jakecoffman:
It worked! https://github.com/dependabot/dependabot-core/pull/5155

PR #5157, opened 2022-05-17
Bump Ruby to 2.7.6

yeikel:


Dockerfile
```diff
@@ -75,11 +75,11 @@ ENV BUNDLE_PATH=".bundle" \
 ENV PATH="$BUNDLE_BIN:$PATH:$BUNDLE_PATH/bin"
 RUN apt-add-repository ppa:brightbox/ruby-ng \
   && apt-get update \
-  && apt-get install -y --no-install-recommends ruby2.7 ruby2.7-dev \
+  && apt-get install -y --no-install-recommends ruby2.7.6 ruby2.7.6-dev \
```
deivid-rodriguez, line 78: I don't think Brightbox nor Ubuntu provides these packages. I think the way to upgrade would be to migrate to ubuntu 22.04, which provides Ruby 3. Alternatively, copy installation commands from the official docker Ruby images, which would give more flexibility on the version that we use.

jeffwidman: FWIW, the `22.04` upgrade is already tracked in https://github.com/dependabot/dependabot-core/pull/5030, although even once CI passes there may be some internal-to-GitHub issues like packaging or security that may delay upgrading.

Ruby 3 is an interesting idea, but probably a bigger conversation. I'll try to remember to check with the rest of the team this coming week to see what the current plan is re: moving to Ruby 3.

deivid-rodriguez: I see.

Regardless of the update to 22.04 and/or to Ruby 3, it would seem like a good move to migrate to Ruby 2.7.6, [the latest security release in the 2.7 series](https://www.ruby-lang.org/en/news/2022/04/12/ruby-2-7-6-released/).

In order to achieve that, I think the best way would be to move away from Brightbox. The alternatives I can think of are:

* Build from source like the official Ruby docker images do: https://github.com/docker-library/ruby/blob/ac24ae0a894606d3ac91c28095db39e69771cac4/2.7/slim-bullseye/Dockerfile. I think this is the most flexible but least maintainable option.
* Use [ruby-build](https://github.com/rbenv/ruby-build). This might be a good middleground?

yeikel: Apologies. I meant to add a comment explaining that this was meant fo open the conversation about the upgrade. I wanted to see the feedback from the CI

mattt:
Thanks for your contribution with this PR, @yeikel. We recently merged #5356, which updates our image to use `ruby-install` instead of `apt`. This allowed us to update to 2.7.6 with #5364.

Issue #5150, opened 2022-05-17
Dependabot update only first terraform dependency in a single file


andrewozh:
Terraform
If there is multiple dependencies located in a single file dependabot will update only first one!

I reproduced a problem here (Check the Pull Requests ):
https://github.com/andrewozh/dependabot-bug-check

**Example:**

Here's a fake .tf file with a two modules! 
0.42.1 version is available here

```
module "first-dep" {
  source  = "cloudposse/elasticache-redis/aws"
  version = "0.42.0"

  name = "it supposed to be updated by dependabot"
}

module "second-dep" {
  source  = "cloudposse/elasticache-redis/aws"
  version = "0.42.0"

  name = "it supposed not to be updated by dependabot"

}
```

I use github-native dependabot, here's my `.github/dependabot.yaml`

```
version: 2
updates:
  - package-ecosystem: "terraform"
    directory: "/"
    schedule:
      interval: "daily"
```

**Expected:**

New version of dependency causes dependabot to create a PRs on updating all dependencies, no matter they located in a single file or not

**Reality:**

Got PR only for first dependency in a file

dwc0011:
TL/DR;  the provider value is used as the dependency name, ie not unique resulting in only one dependency being created
Workaround: (if possible) use git url for source values instead, that appears to be working fine

I believe this is due to the naming for provider sources not being unique, the provider value is used as the name for the dependency as set in [ terraform/file parser.rb](https://github.com/dependabot/dependabot-core/blob/57136896e10834f87a647f99b6d7601e95a45a08/terraform/lib/dependabot/terraform/file_parser.rb#L79) 

Since its name is not unique it is found as an existing dependency and is not added again, thus will only process once. [base/dependency_set](https://github.com/dependabot/dependabot-core/blob/57136896e10834f87a647f99b6d7601e95a45a08/common/lib/dependabot/file_parsers/base/dependency_set.rb#L26)

In order to fix this and add a unique PR for each module referencing the same provider, the naming must change as well as the regex in the [terraform/file_updater.rb method update_registry_declaration](https://github.com/dependabot/dependabot-core/blob/57136896e10834f87a647f99b6d7601e95a45a08/terraform/lib/dependabot/terraform/file_updater.rb#L88) 

Let me know if you are able to address this or not, I am currently working another unrelated task but also within the terraform module and can look at this after that if you are unable to fix this.

dwc0011:
[...]

Issue #5128, opened 2022-05-11
Dependabot raises RuntimeError instead of opening PR for new version 


AtzeDeVries:
**Package ecosystem**
terraform
**Package manager version**
1.x
**Language version**
1.x
**Manifest location and content before the Dependabot update**
reference to a private terraform module
**dependabot.yml content**
```yaml
---
version: 2

registries:
  coop-terraform:
    type: terraform-registry
    url: https://terraform.coop.no
    token: ${{ secrets.TFE_TOKEN }}
  coop-ghcr:
    type: docker-registry
    url: ghcr.io
    username: CoopGithubServiceaccount
    password: ${{ secrets.DEPENDABOT_GHCR_PULL }}

updates:
  - package-ecosystem: "terraform"
    directory: "/azure"
    registries:
      - coop-terraform
    schedule:
      interval: "weekly"
      day: "sunday"
  - package-ecosystem: "terraform"
    directory: "/management"
    registries:
      - coop-terraform
    schedule:
      interval: "weekly"
      day: "sunday"
  - package-ecosystem: "terraform"
    directory: "/projects"
    registries:
      - coop-terraform
    schedule:
      interval: "weekly"
      day: "sunday"
  - package-ecosystem: "terraform"
    directory: "/sandboxes"
    registries:
      - coop-terraform
    schedule:
      interval: "weekly"
      day: "sunday"
  - package-ecosystem: "terraform"
    directory: "/teams"
    registries:
      - coop-terraform
    schedule:
      interval: "weekly"
      day: "sunday"
  - package-ecosystem: "terraform"
    directory: "/tfe-workspaces"
    registries:
      - coop-terraform
    schedule:
      interval: "weekly"
      day: "sunday"
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
      day: "sunday"
  - package-ecosystem: "docker"
    directory: "/docker-compose"
    registries:
      - coop-ghcr
    schedule:
      interval: "daily"
```
**Updated dependency**
from 0.8.0 to 1.0.5 (from old to new version model
**What you expected to see, versus what you actually saw**
Expected: A PR with an update for the new version
Actually: Dependabot throws an error: 
```
updater | ERROR <job_362442582> Error processing coopnorge/repos/github (RuntimeError)
updater | ERROR <job_362442582> Content didn't change!
```
**Native package manager behavior**
[...]

PR #5168, opened 2022-05-18
Check if updated file exists before adding to list of updated files

### LINKED ISSUE #5167: Terraform: Two or more modules with same relative source in the same file results in multiple attempts to create the same PR
Terraform referencing the same relative source more than once in the same file results in the following error when creating commit and PR in CodeCommit, it works with Github as that processes duplicate files in the commit differently, unable to test with Bitbucket and Azure

```
Updating iam_policy_document::github::plus3it/terraform-aws-tardigrade-iam-principals::11.0.1 from 11.0.1..to 12.0.0
--
986 | /home/dependabot/dependabot-script/vendor/ruby/2.7.0/gems/aws-sdk-core-3.131.0/lib/seahorse/client/plugins/raise_response_errors.rb:17:in `call': 
Cannot put the file at the specified path because there are one or more conflicting file change requests for that file. 
Either resolve the conflicts locally, or put the files you want to add in different paths, and then try again.. 
Path: /modules/iam_policy_document/main.tf (Aws::CodeCommit::Errors::SamePathRequestException)
```

**Package manager version**
Terraform

**Updated dependency**
:github::plus3it/terraform-aws-tardigrade-iam-principals::11.0.1 from 11.0.1..to 12.0.0
Expected: the child module iam_policy_document to be updated once since the module is a relative source, it should only need to be updated once
Actual: the updated_files list contains the file multiple times referenced for the same 

Proposed solution:  [terraform/file_updater.rb](https://github.com/dependabot/dependabot-core/blob/780bcdb0b6f81659a071c0220632a98265625bfc/terraform/lib/dependabot/terraform/file_updater.rb#L28) not add duplicate files if they exist already.

** Minimal Config to reproduce **
./main.tf:
```
module "child1" {
  source = "./child"
}

module "child2" {
  source = "./child"
}

```
./child/main.tf
```
module "dns" {
  source = "git::https://github.com/cloudposse/terraform-aws-route53-cluster-zone.git//some/dir?ref=tags/0.2.5"
}
```

I am working this PR already.

dwc0011:
Fixes #5167 by not adding duplicate files to the updated files list

### DIFF
--- a/terraform/lib/dependabot/terraform/file_updater.rb
+++ b/terraform/lib/dependabot/terraform/file_updater.rb
@@ -29,7 +29,9 @@ module Dependabot
 
           raise "Content didn't change!" if updated_content == file.content
 
-          updated_files << updated_file(file: file, content: updated_content)
+          updated_file = updated_file(file: file, content: updated_content)
+
+          updated_files << updated_file unless updated_files.include?(updated_file)
         end
         updated_lockfile_content = update_lockfile_declaration(updated_files)
 


dwc0011:
I have updated issue to provide a sampel config to illustrate the issue

jeffwidman:
Code change looks reasonable to me, with the caveat that I've got minimal terraform expertise.

However, would it be possible to add an rspec test that includes a repo fixture simulating the failure condition of duplicate files? No need to simulate the codecommit response, and I realize GitHub already doesn't fail on this, but it'd still be good to have a test repo to ensure no regressions in the future. Just create a test fixture repo copying your setup from https://github.com/dependabot/dependabot-core/issues/5167#issue-1240553985.

dwc0011:
> Code change looks reasonable to me, with the caveat that I've got minimal terraform expertise.
> 
> However, would it be possible to add an rspec test that includes a repo fixture simulating the failure condition of duplicate files? No need to simulate the codecommit response, and I realize GitHub already doesn't fail on this, but it'd still be good to have a test repo to ensure no regressions in the future. Just create a test fixture repo copying your setup from [#5167 (comment)](https://github.com/dependabot/dependabot-core/issues/5167#issue-1240553985).

I added in a test fixture and updated the rspec, I am still new to ruby and especially the testing stuff so please review, I am sure it will need some expanding on but wanted to see if this was the right start for it.  

Thanks ahead of time.

dwc0011:
After investigating the test further I may need some guidance on how to test it properly.  The terraform file fetcher gathers the files to [process based relative paths etc](https://github.com/dependabot/dependabot-core/blob/50fd5475565065e8e6a1421460b78093e9cc53e8/terraform/lib/dependabot/terraform/file_fetcher.rb#L54)., this is how the duplicate file gets in there.  
But the rspec loads the files differently, by using the [helper](https://github.com/dependabot/dependabot-core/blob/main/bundler/helpers/v2/spec/native_spec_helper.rb) to read all the files in the fixture, this does not duplicate the file.  
I am not sure on the best path for this.  Do we edit the [file fetcher](https://github.com/dependabot/dependabot-core/blob/main/terraform/lib/dependabot/terraform/file_fetcher.rb) instead and prevent duplicate files there or allow the updater to filter out the duplicates to ensure that we do not exclude something prematurely in the file fetcher.  Any thoughts on this would be appreciated.

dwc0011:
Thinking about this more, the current solution gets this working,  maybe use this patch as a quick fix and address the bigger issues as different issue or in one of the related issues such as, #5128 & #5150.
A do think review of how the terraform file fetcher works (processing directories listed in dependabot.yml in addition to resolving relative paths), naming of terraform dependencies for modules and submodules (including git sources, registry and provider sources), and just an overall design review is probably warranted.

Nishnha:
Hi! 👋🏾  

Sorry for the silence around this PR--it looks good to me. Thank you for adding appropriate tests.

As for the other linked Terraform issues:
I don't expect this PR to resolve either of them since I think https://github.com/dependabot/dependabot-core/issues/5128 is due to Dependabot not comparing the `variables.tf` file or other similar "metadata" files between versions, and https://github.com/dependabot/dependabot-core/issues/5150 is due to Dependabot having inconsistent in how it treats different Terraform source.

I'll try to take a look at those issues separately sometime this week.

Nishnha (approved):

dwc0011:
> Hi! 👋🏾
> 
> Sorry for the silence around this PR--it looks good to me. Thank you for adding appropriate tests.
> 
> As for the other linked Terraform issues: I don't expect this PR to resolve either of them since I think #5128 is due to Dependabot not comparing the `variables.tf` file or other similar "metadata" files between versions, and #5150 is due to Dependabot having inconsistent in how it treats different Terraform source names/versions.
> 
> I'll try to take a look at those issues separately sometime this week.

Nish, thanks for approving this PR.  I didn't mean that this PR would resolve those other PRs rather that those PRs would address the bigger issues of how dependencies are named and files are fetched would fix this issue more appropriately.  
Thanks!

PR #5181, opened 2022-05-23
Remove versions from comments

jeffwidman:
There's no need for the comments to specify the versions when
the code with the actual version pin is just a couple of lines below.

Removing the version from the comment removes the chance that the comment diverges from the code.

It also makes future PR's simpler because they have one less thing to remember to update.


landongrindheim (approved):
Makes sense to me 😄

PR #5190, opened 2022-05-25
Install Ruby with rbenv

mattt:
Related to https://github.com/dependabot/dependabot-core/pull/5031#issuecomment-1107498561

Currently, we use the [Brightbox](https://www.brightbox.com/docs/ruby/ubuntu/) PPA to install Ruby with `apt`. However, this limits our ability to update to a more recent patch release of Ruby 2.7 (or later, Ruby 3).

This PR updates this step in `Dockerfile` to instead use [rbenv](https://github.com/rbenv/rbenv) and [ruby-install](https://github.com/rbenv/ruby-install) to install and manage ruby versions. I used Ruby 2.7.5 to match the version installed now. As a follow-up, we should upgrade to [2.7.6](https://www.ruby-lang.org/en/news/2022/04/12/ruby-2-7-6-released/).


Dockerfile
```diff
@@ -73,15 +76,24 @@ ENV DEBIAN_DISABLE_RUBYGEMS_INTEGRATION=true
 ENV BUNDLE_PATH=".bundle" \
     BUNDLE_BIN=".bundle/bin"
 ENV PATH="$BUNDLE_BIN:$PATH:$BUNDLE_PATH/bin"
-RUN apt-add-repository ppa:brightbox/ruby-ng \
-  && apt-get update \
-  && apt-get install -y --no-install-recommends ruby2.7 ruby2.7-dev \
```
Nishnha, line 78: I noticed we used to install `ruby-dev` which is a package with "header files for compiling extension modules for Ruby".

AFAIK RVM installs ruby as source for the machine so this might not be a concern?

We might also want to check if this change makes the Docker build take significantly longer in case we are actually building from source instead of pulling a binary

Nishnha (approved):
LGTM

Dockerfile
mattt: > I noticed we used to install ruby-dev which is a package with "header files for compiling extension modules for Ruby".
>
> AFAIK RVM installs ruby as source for the machine so this might not be a concern?

The first commit (bcf435f), I messed up the `PATH` and ended up running whatever system Ruby was available. And CI failed when attempting to install native extensions because of those missing headers.

The next commit (b591c76) fixed the `PATH` and everything worked as expected. So I think we should be arlight.

> We might also want to check if this change makes the Docker build take significantly longer in case we are actually building from source instead of pulling a binary

That shouldn't be as much of a concern on subsequent builds if the layer is cached. But yeah, it'll be interesting to see what difference that makes.

deivid-rodriguez:
I'm totally in favor of going this route instead of keep using Ruby packaged by the OS 👍.

Also, just to make sure, there's another popular software called [ruby-installer](https://github.com/postmodern/ruby-install), but this PR intends to use [ruby-build](https://github.com/rbenv/ruby-build), correct?

Finally, do we need `rbenv` at all? Rbenv is normally used when you need to frequently switch between Ruby versions, but in this case, there's a single Ruby version. It could be fine, only gotcha I see is that since every Ruby executable has to go through a shim to "redirect" to the one from the right version, but it's probably not a big deal for us. Anyways, if wanted I think we may be able to get rid of `rbenv` by directly configuring `PATH`?

DavidTurner-N:
Is there a plan for when this PR is going to be merged in please?  And will it unblock #5030 ?

Thanks!

Nishnha:
I did some more reading on this topic and it seems like what this PR proposes, using [rbenv](https://github.com/rbenv/rbenv) + [ruby-build](https://github.com/rbenv/ruby-build), covers all the bases for installing Ruby from source and changing the necessary paths for our Docker shell. 
It's also in-line with our recommended setup instructions in the [Dependabot contributing doc](https://github.com/dependabot/dependabot-core/blob/e1029333896ef268226b90f8ad794d055a2f2caf/CONTRIBUTING.md#setup-instructions)

There are a couple of alternatives to `rbenv` and `ruby-build` though:

- We could use [chruby](https://github.com/postmodern/chruby), an alternative to `rbenv` that doesn't use shims. This might be be worth looking into since shimming can be expensive. @deivid-rodriguez what do you think? Bundler jobs already have a longer average runtime than other ecosystem, could this make a difference? As far as I can tell, Dependabot doesn't actually change its Ruby version after the Docker image is built.

- We could also use [ruby-install](https://github.com/postmodern/ruby-install) instead of `ruby-build`, which means a `ruby-install ruby 2.7` would pull the latest patch version (instead of manually having to set the exact version with `ruby-build`). We already have to manually increment the supported Ruby versions in https://github.com/dependabot/dependabot-core/blob/e1029333896ef268226b90f8ad794d055a2f2caf/bundler/lib/dependabot/bundler/file_updater/ruby_requirement_setter.rb#L13-L15 anyway though, so it's not a big deal.


One potential side effect we should check before deploying is that `rbenv` won't try to [change the local ruby version if it detects a `.ruby-version` file](https://github.com/rbenv/rbenv#choosing-the-ruby-version) in a project, which might happen when we use SharedHelpers to execute commands in a temporary project directory. 
Since we already set the `BUNDLER_VERSION` environment variable in our Helpers when calling bundler, we could likewise set the `RBENV_VERSION`?

deivid-rodriguez:
@Nishnha Yes, it's true that chruby could be better since it doesn't use shims, but my idea was to avoid installing and using any ruby switcher at all since we don't really need to switch rubies. I'm happy to further investigate this idea.

mattt:
Thanks for sharing your thoughts, @Nishnha @deivid-rodriguez! I agree that the additional switching behavior of `rbenv` could be problematic for our particular use case, so I opened #5356 as an alternative.

PR #5187, opened 2022-05-24
Instrument npm version

Nishnha:
Instruments the package manager version for npm and yarn similar to [how we instrument bundler versions](https://github.com/dependabot/dependabot-core/blob/e96b546e8d6bbe723732ba9657f9008c75c0424a/bundler/lib/dependabot/bundler/file_parser.rb#L46-L55)

Note: 
This PR instruments in the FileFetcher step so it should only run once.
The dry-runner shows this too

```shell
[dependabot-core-dev] ~/dependabot-core $ SECURITY_ADVISORIES='[{"dependency-name":"node-forge","patched-versions":[],"unaffected-versions":[],"affected-versions":["< 0.10.0"]},{"dependency-name":"node-forge","patched-versions":[],"unaffected-versions":[],"affected-versions":["< 1.0.0"]},{"dependency-name":"node-forge","patched-versions":[],"unaffected-versions":[],"affected-versions":["< 1.3.0"]}]' bin/dry-run.rb npm_and_yarn <redacted> --dir=/node-forge-npm8 --dep=node-forge --security-updates-only
warning: parser/current is loading parser/ruby27, which recognizes2.7.6-compliant syntax, but you are running 2.7.5.
Please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
=> fetching dependency files
{:ecosystem=>"npm", :package_managers=>{"npm"=>8}} # <----- This is the only time the instrumentation is called
=> dumping fetched dependency files: ./dry-run/<redacted>
=> parsing dependency files
=> updating 1 dependencies: node-forge

=== node-forge (0.10.0) (vulnerable 🚨)
 => checking for updates 1/1
🌍 https//registry.npmjs.org:443/node-forge
🌍 https//registry.npmjs.org:443/node-forge/1.3.1
 => latest available version is 1.3.1
🌍 https//registry.npmjs.org:443/node-forge/1.3.0
 => earliest available non-vulnerable version is 1.3.0
 => latest allowed version is 0.10.0
 => The update is not possible because of the following conflicting dependencies:
   webpack-dev-server@3.11.3 requires node-forge@^0.10.0 via selfsigned@1.10.14
 => requirements to unlock: update_not_possible
🌍 https//registry.npmjs.org:443/forge-npm8
 => requirements update strategy: bump_versions
    (no security update possible 🙅‍♀️)
🌍 Total requests made: '4'
🎈 Package manager version log: {:ecosystem=>"npm", :package_managers=>{"npm"=>8}}

```

PR #5194, opened 2022-05-25
Print package manager version instrumentation events in the dry runner

Nishnha:
Follow up to https://github.com/dependabot/dependabot-core/pull/5187

Prints package manager version instrumentation events in the dry runner

Previously in this PR, the dry runner would print out the package manager version log when the notification event occurred, but really it should only be logged once per ecosystem. 

Since the additional logging was polluting, I modified the dry runner to instead print out the logging event(s) at the end of the run.

The dry runner will still show all the package manager version log events in case it really is logged more than once per ecosystem

```
🌍 Total requests made: '7'
🎈 Package manager version log: {:ecosystem=>"npm", :package_managers=>{"npm"=>8, "yarn"=>1}}
```

### DIFF
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -469,11 +469,9 @@ ActiveSupport::Notifications.subscribe(/excon.request/) do |*args|
   puts "🌍 #{payload[:scheme]}//#{payload[:host]}:#{payload[:port]}#{payload[:path]}"
 end
 
-$package_manager_version_events_count = 0
+$package_manager_version_log = []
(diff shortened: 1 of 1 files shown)


jakecoffman (approved):
🎈

jeffwidman (approved):

Nishnha:
I changed the base branch on this PR because CI is stuck. It only updates `bin/dry-run.rb` so it shouldn't need to run the ecosystem tests to pass.

Not sure what to do 😮‍💨

PR #5208, opened 2022-05-31
Update test cases with new hashes returned from npm registry

mctofu:
We are getting some [npm ci failures](https://github.com/dependabot/dependabot-core/runs/6665124687?check_suite_focus=true) from tests that run against the actual npm registry. Some dependencies have been returning sha512 hashes instead of sha1. This updates the failing test cases to expect the sha512 hashes.


jeffwidman (approved):
Thanks for handling this.

PR #5220, opened 2022-06-02
Fix Yarn ignored path checking

FunkyloverOne:
The previous implementation looked like this:
```ruby
ignored_paths = path.scan(/!\((.*?)\)/).flatten
# ...
reject { |fn| ignored_paths.any? { |p| fn.include?(p) } }
```

Which easily could reject something by mistake. Let's say we have a workspace glob pattern like this: `packages/!(pack)` (this is the one used in the spec), therefore: 
```ruby
ignored_paths #=> ["pack"]
["package2"].reject { |fn| ignored_paths.any? { |p| fn.include?(p) } } #=> []
```
It won't happen anymore. :slightly_smiling_face:


pavera (approved):
Thanks for the PR @FunkyloverOne!

PR #5221, opened 2022-06-02
[feature] Support NPM Transitive Dependency Security Updates

Nishnha:
Adds NPM transitive dependency support for Dependabot security updates.

This PR leverages the npm Arborist library to figure out which other dependencies need to be updated in order to upgrade a vulnerable dependency to the lowest non-vulnerable version.

We run `Arborist.audit` on a copy of the project and filter out the API request for all vulnerabilities so that only vulnerabilities for the dependency receiving the security update is considered.

Arborist then uses this filtered list of vulnerabilities to determine which group of dependencies need to be updated.

Dependabot takes that group of dependencies and updates them all in a single pull request.


npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker/vulnerability_auditor_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    context "when the native helper raises" do
+      let(:dependency_files) { project_dependency_files("npm8/simple") }
+
+      it "returns fix_available => false and logs the failure" do
+        # Stub native helper path with the `false` builtin in order to get a
+        # non-zero exit status from the helper subprocess and cause
+        # `Dependabot::SharedHelpers::HelperSubprocessFailed` to be raised.
+        allow(Dependabot::NpmAndYarn::NativeHelpers).
+          to receive(:helper_path).and_return("false")
```
bdragon, line 84: I'm not crazy about this test knowing that `#audit` uses `Dependabot::NpmAndYarn::NativeHelpers.helper_path` to run the helper subprocess, but this was the best I could come up with for simulating an exception raised by the helper.

Issue #4797, opened 2022-03-04
Dependabot generates faulty poetry.lock when using a private package repository


itsdani:
**Package ecosystem**
poetry (python)

**Package manager version**
poetry 1.1.13 (and pip 22.0.3)

**Language version**
Python 3.9.6

**Manifest location and content prior to update**
[/pyproject.toml](https://github.com/turbine-ai/dependabot-poetry-lock-bug/blob/main/pyproject.toml)
<details>
  <summary>Original pyproject.toml contents</summary>

  ```toml
[tool.poetry]
name = "dependabot-poetry-lock-bug"
version = "0.1.0"
description = "Minimal example to reproduce dependabot messing up poetry.lock repository references"
authors = ["Daniel Segesdi <daniel.segesdi@turbine.ai>"]
license = "MIT"

[[tool.poetry.source]]
name = "privaterepo"
url = "https://turbinetest.jfrog.io/artifactory/api/pypi/privaterepo/simple"

[tool.poetry.dependencies]
python = "^3.9"
toml = "==0.9.6"

[tool.poetry.dev-dependencies]

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
  ```
</details>

[/poetry.lock](https://github.com/turbine-ai/dependabot-poetry-lock-bug/blob/main/poetry.lock)

<details>
  <summary>Original poetry.lock contents</summary>

```toml
[[package]]
name = "toml"
version = "0.9.6"
description = "Python Library for Tom's Obvious, Minimal Language"
category = "main"
optional = false
python-versions = "*"

[package.source]
type = "legacy"
url = "https://turbinetest.jfrog.io/artifactory/api/pypi/privaterepo/simple"
reference = "privaterepo"

[metadata]
lock-version = "1.1"
python-versions = "^3.9"
content-hash = "652d570962ca3eb788b15045d1cb3d205b1f22e3e813fc8763346d6742e72e5e"

[metadata.files]
toml = [
    {file = "toml-0.9.6-py2.py3-none-any.whl", hash = "sha256:a7901919d3e4f92ffba7ff40a9d697e35bbbc8a8049fe8da742f34c83606d957"},
    {file = "toml-0.9.6.tar.gz", hash = "sha256:380178cde50a6a79f9d2cf6f42a62a5174febe5eea4126fe4038785f1d888d42"},
]
```
</details>

[...]

Issue #2487, opened 2020-08-27
Dependabot is leaking private registry credentials in poetry.lock


omartrigui:
**Setup**
- dependabot-core v0.118.16
- 2 private python registries were configured

**Package manager/ecosystem**
pip

**Manifest contents prior to update** (pyproject.toml)
![Screenshot from 2020-08-27 17-18-08](https://user-images.githubusercontent.com/7053612/91461599-7cca0780-e889-11ea-9da5-750e6ca6f4d5.png)

**Updated dependency**

PR title: `Bump pytest-cov from 2.8.1 to 2.10.1`

**What you expected to see, versus what you actually saw**

Expectation (poetry.lock):
- `poetry.lock` gets updated without leaking secrets (USERNAME & PASSWORD) and without altering the url (as I already mentioned, 2 python registries are used)

Reality (poetry.lock):
![Screenshot from 2020-08-27 17-17-38](https://user-images.githubusercontent.com/7053612/91461797-b733a480-e889-11ea-9949-2aafb1e1688a.png)

jurre:
Thanks @omartrigui, could you maybe share a bit more context about how your project is configured? A sample poetry.lock configuration from before the update for example would be super helpful.

omartrigui:
@jurre Thanks for your quick reply. As the existence of a private package triggers this issue, I have added all the possible details to the description (_pyproject.toml_ & _poetry.lock_).

jurre:
Thanks! Just to verify, where it says `__OMITTED__`, that has a different value in the actual file checked into git, right?

When using `url = "https://__OMITTED_HOST__/pypi/simple"` locally, have you configured the credentials using `POETRY_HTTP_BASIC_PYPI_USERNAME` and `POETRY_HTTP_BASIC_PYPI_PASSWORD`?

Also, I'm assuming you're using dependabot.com / dependabot-preview for these updates, is that correct?

omartrigui:
> Also, I'm assuming you're using dependabot.com / dependabot-preview for these updates, is that correct?

Actually no, I am using the open source version of dependabot: `gem "dependabot-omnibus", "~> 0.118.16"`

> where it says __OMITTED__, that has a different value in the actual file checked into git, right?

Actually not in all cases, `__OMITTED__` was referring to sensitive values that cannot be exposed publicly. I have updated the description and have added `# changed by dependabot but should not` comment where Dependabot shouldn't have suggested changes.

[...]

PR #5222, opened 2022-06-02
poetry: fixes for broken lockfile and leaked credentials

jakecoffman:
fixes for #2487 and #4797

These issues are caused by the current method for adding credentials to private registries which is to add them to the pyproject.toml file as a new source with a dynamically generated name and credentials in the url.

Unfortunately this is causing Poetry to add the name and url with credentials (when present) to the lockfile.

GitHub hosted Dependabot uses a proxy to inject credentials into the request for extra security, so this file writing to add credentials is unnecessary. 

For users who host Dependabot themselves, or for when we want to run the dry-run script without a proxy, there are other ways to get Poetry the credentials.

So, in this PR I removed the pyproject.toml `add_private_sources` code and added credentials via environment variables. This should fix the issue for both use-cases of Dependabot.


pavera (approved):
I think this looks good :)

PR #5237, opened 2022-06-07
Only report conflicting dependencies when update is not possible

### LINKED ISSUE #4867: Dependabot can never fix Rails vulnerabilities
**Package ecosystem:** bundler
**Updated dependency:** `rails`
**What you expected to see, versus what you actually saw:**
`rails` is a collection of sub-gems that are all updated in lockstep. Our `Gemfile` just has `gem "rails"`. Anytime a Rails vulnerability is found, it's reported under the respective sub-gem (like `activestorage`). We'll get a GitHub security Dependabot alert for it but Dependabot will be unable to make a PR to fix the vulnerability. The security alert (example [here](https://github.com/panorama-ed/rainbow-backend/security/dependabot/6)) looks something like this:

> **Dependabot cannot update activestorage to a non-vulnerable version**

> The latest possible version that can be installed is `6.1.4.6` because of the following conflicting dependencies:

```
rails (6.1.4.6) requires activestorage (= 6.1.4.6) via actionmailbox (6.1.4.6)
rails (6.1.4.6) requires activestorage (= 6.1.4.6) via actiontext (6.1.4.6)
rails (6.1.4.6) requires activestorage (= 6.1.4.6)
```

The earliest fixed version is `6.1.4.7`.

This happens all the time, since Rails vulnerabilities are unfortunately somewhat common and we have many repos that use Rails as a dependency.

Would it be possible for Dependabot to try updating a dependency's parent when updating the dependency fails for a security fix? (Alternatively, given the popularity of Rails, perhaps a Rails-specific fix would be justified and simpler to implement?)

deivid-rodriguez:
I extracted this commit from https://github.com/dependabot/dependabot-core/pull/5221, because it's also useful for me to (fix) #4867.

If I understand correctly, both https://github.com/rubygems/rubygems/pull/5520 and #5221 are fixing the same issue but for different ecosystems, and we both run into the same issue where the dependency is properly updated but `bin/dry-run.rb` still reports conflicts.

### DIFF
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -674,16 +674,6 @@ dependencies.each do |dep|
                            end
   puts " => latest allowed version is #{latest_allowed_version || dep.version}"
 
-  conflicting_dependencies = checker.conflicting_dependencies
-  if conflicting_dependencies.any?
-    puts " => The update is not possible because of the following conflicting "\
-      "dependencies:"
-
-    conflicting_dependencies.each do |conflicting_dep|
-      puts "   #{conflicting_dep['explanation']}"
-    end
-  end
-
   if checker.up_to_date?
     puts "    (no update needed as it's already up-to-date)"
     next
@@ -712,6 +702,17 @@ dependencies.each do |dep|
     else
       puts "    (no update possible 🙅‍♀️)"
     end
+
+    conflicting_dependencies = checker.conflicting_dependencies
+    if conflicting_dependencies.any?
+      puts " => The update is not possible because of the following conflicting "\
+        "dependencies:"
+
+      conflicting_dependencies.each do |conflicting_dep|
+        puts "   #{conflicting_dep['explanation']}"
+      end
+    end
+
     next
   end
 


Nishnha:
Hey @deivid-rodriguez, 

Thanks for your work on https://github.com/rubygems/rubygems/pull/5520!

It should make a big impact on the Ruby PRs Dependabot can open! Once it lands I think we will also need to modify the [bundler ConflictingDependencyResolver](https://github.com/dependabot/dependabot-core/blob/main/bundler/helpers/v2/lib/functions/conflicting_dependency_resolver.rb) so it no longer flags locked parent dependencies as conflicting.

And yes, your work is similar to what we are doing with `npm` in #5221--we are leverage Arborist and `npm audit fix` to identify and update conflicting dependencies.

We had closed out #5221 in favor of #5239 so our PR was more up-to-date. The new PR contains the same commits, and it should be shipped in the next few weeks. 

I think we will also need to apply this PR's patch to our internal Updater system for it to work in production--the `dry-run.rb` script essentially emulates that system for local testing.

mctofu (approved):
This would be fine to pull in early. It's purely cosmetic as the message doesn't prevent the update from happening.

@Nishnha I've already checked our internal updater and it already has this behavior of only reporting conflicts when the update fails.

PR #5239, opened 2022-06-07
[npm] Initial support for locked transitive dependency updates

mctofu:
This adds support for updating locked transitive dependencies.

TODO: Provide example

## Key points
- The support is gated by a feature flag for testing
- We rely on the audit features of the `@npmcli/arborist` library to resolve which dependencies to update. This is similar to running `npm audit fix --force`.
- An `@npmcli/arborist` audit will normally audit dependencies against the full list of vulnerabilities in the npm audit registry. When Dependabot performs a security update we want to target a particular vulnerable dependency. To achieve this we've provided a custom audit registry to `@npmcli/arborist` which returns only the vulnerabilities Dependabot intends to fix.
- This new feature only activates for vulnerabilities that Dependabot is not able to fix (update not possible). For instances where Dependabot is able to provide a fix (top level dependency or transitive dependency that is not locked) the existing logic applies.

## Limitations / Known issues / Notes
- The vulnerable dependency is returned as a dependency to update so it will be included in the PR. However, in most cases we don't need to attempt to update it since it would be updated along with its parent's update. We may try to optimize this in the future.
- Dependabot normally tries to update to the lowest fixed version of the dependencies. This implementation defers to the updated versions from the arborist audit which favor the latest allowed versions.
- There are edge cases such as a vulnerable dependency that is both a top level & locked transitive dependency that won't be properly resolved yet.
- While the `@npmcli/arborist` audit will produce an updated `package.json` and `package-lock.json` we discard this result and rely on our npm `FileUpdater` to reproduce the same changes. This may lead to issues where the `FileUpdater` isn't able to apply the changes but is the most pragmatic route for now as changing that process would require extensive changes.


npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
```diff
@@ -77,6 +80,19 @@ def audit(dependency:, security_advisories:)
...
+          # we only need to check results that indicate a fix is available
+          return true unless audit_result["fix_available"]
+
+          return false if vulnerable_dependency_removed?(audit_result)
+
+          true
+        end
+
+        def vulnerable_dependency_removed?(audit_result)
+          !audit_result.key?("target_version")
```
bdragon, line 93: We might want to handle a falsy value, too, just to be safe.

mctofu: Addressed in https://github.com/dependabot/dependabot-core/pull/5239/commits/ff7fe3a8358cee699494d42a71e0dc22b1346dd6. Is that what you were thinking?

bdragon: Yup!

Nishnha:
Opened https://github.com/github/dependabot-updates/issues/2531 as a shipping checklist for this PR

bin/dry-run.rb
```diff
@@ -592,6 +592,8 @@ def security_advisories
 end
 
 def peer_dependencies_can_update?(checker, reqs_to_unlock)
+  return false if $options[:security_updates_only]
```
feelepxyz, line 595: This makes sense on a third take.

Initially I was thinking you could get duplicate PRs if a security update also needs to update a peer dependency. If another update tries to update the peer dep you would end up with two PRs that have an identical diff but the title/body would say PR 1 updates: A + B and PR 2 updates B + A.

I think my initial assumption is incorrect as you only get into the identical diff case if you're doing version updates, trying to update every dependency in turn, whereas security updates are only ever trying to update one dependency.

There might be a slightly odd case if two security advisories exist, one for a transitive dep and one for a top-level dep that needs to be unlocked. In this case you would not get an identical diff, but the two PRs would conflict as they would both update the top level dep but this doesn't seem worth handing yet..

I was wondering if it would be worth documenting some of this weirdness here, in case someone runs into this case and comes along looking for explanations in future?

npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker_spec.rb
```diff
@@ -1205,6 +1265,146 @@
...
+              version: "1.0.1",
+              package_manager: "npm_and_yarn",
+              previous_version: "1.0.0",
+              requirements: [],
+              previous_requirements: []
+            )
+          ])
+      end
+
+      context "when a transitive dependency is locked by multiple top-level dependencies" do
```
feelepxyz, line 1327: Cool seeing this case handled!

npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker/vulnerability_auditor_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+            "fix_available" => true,
+            "target_version" => "1.0.0"
+          })
+
+        expect(subject.audit(dependency: dependency, security_advisories: security_advisories)).
+          to include("fix_available" => false)
+      end
+    end
+
+    context "when a fix would downgrade a dependency" do
```
feelepxyz, line 110: Nice!

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
```diff
@@ -0,0 +1,136 @@
...
+        end
+
+        # Finds any dependencies in the `package-lock.json` or `npm-shrinkwrap.json` that have
+        # a subdependency on the given dependency that is locked to a vuln version range.
+        #
+        # NOTE: yarn is currently not supported.
+        #
+        # @param dependency [Dependabot::Dependency] the dependency to check
+        # @param security_advisories [Array<Dependabot::SecurityAdvisory>] advisories for the dependency
+        # @return [Hash<String, [String, Array<Hash<String, String>>]>] the audit results
```
feelepxyz, line 30: This return object is a nice pattern 👍

npm_and_yarn/helpers/lib/npm/vulnerability-auditor.js
```diff
@@ -0,0 +1,207 @@
...
+    auditRegistry: 'http://localhost:9999',
+    ca: caCerts,
+    force: true,
+    dryRun: true,
+  })
+
+  const scope = nock('http://localhost:9999')
+    .persist()
+    .post('/-/npm/v1/security/advisories/bulk')
+    .reply(200, convertAdvisoriesToRegistryBulkFormat(advisories))
```
feelepxyz, line 49: Nice hack 🤠  would be nice if you could just pass advisories to Arborist somehow? 🤔

npm_and_yarn/helpers/lib/npm/vulnerability-auditor.js
```diff
@@ -0,0 +1,207 @@
...
+ */
+
+const Arborist = require('@npmcli/arborist')
+const nock = require('nock')
+const { promisify } = require('util');
+const exec = promisify(require('child_process').exec)
+
+async function findVulnerableDependencies(directory, advisories) {
+  const npmConfig = await loadNpmConfig()
+  const caCerts = loadCACerts(npmConfig)
```
feelepxyz, line 36: Would be neat if there was lib or something in npm that could be re-used to load the config for arborist? Can't see this but maybe @nlf or @wraithgar might know?

feelepxyz (approved):
Awesome work on this 🙌  This looks really thorough and well thought out to me 💯  The use of arborist makes sense and the test cases look good afaict. 

The use of nock for advisories and having to reconstruct the npm config as well as loading ca files stood out as something that we could maybe improve upstream?

npm_and_yarn/helpers/lib/npm/vulnerability-auditor.js
wraithgar: This is a known pain point on the cli, and it's due to the fact that the config is functionally split between the cli itself (where config definitions live) and `@npmcli/config` (where the config logic itself lives).  There is no better way today to do this in tests right now, but it's on our roadmap for improvement.

Until config is pulled out into a more reproducible format, getting it straight from the npm cli is absolutely the best place to build from.

npm_and_yarn/helpers/lib/npm/vulnerability-auditor.js
wraithgar: It's not tied to the main arborist export but you could try using the same [method](https://github.com/npm/cli/blob/f2340f816f5a137038945229bf8a0d30160e76e8/workspaces/arborist/lib/audit-report.js#L275) that arborist uses to do this:

```js
> AuditReport = require('@npmcli/arborist/lib/audit-report')
[class AuditReport extends Map]
> AuditReport.auditToBulk
[Function: auditToBulk]
```

The arborist method would use the `id` from the advisory itself.  You'd have to move the `nextAdvisoryId` counter logic somewhere else still (if you still need it, I'm not 100% sure what problem that's solving).

bdragon: > You'd have to move the nextAdvisoryId counter logic somewhere else still (if you still need it, I'm not 100% sure what problem that's solving).

Early on we observed that if we returned multiple vuln version ranges for a single dependency, e.g.,

```json
{
  "node-forge": [
    {"vulnerable_versions":"<0.10.0"},
    {"vulnerable_versions":"<1.0.0"},
    {"vulnerable_versions":"<1.3.0"}
  ]
}
```

Arborist would only pick up the first range (`<0.10.0` here) and ignore the rest. We eventually figured out that internally it was uniquing these objects by an `id` field so we added the counter. Thanks for the callout—we could make this clearer in the code comments.

npm_and_yarn/helpers/lib/npm/vulnerability-auditor.js
bdragon: @wraithgar thanks for confirming—we weren't sure if this was the best way to do this. We did look at pulling in `@npmcli/config` but we felt a lot of friction in trying to get it configured correctly so ultimately we chose this.

npm_and_yarn/helpers/lib/npm/vulnerability-auditor.js
wraithgar: > Early on we observed that if we returned multiple vuln version ranges for a single dependency, e.g.,

That's not how advisories return multiple ranges in the npm advisory database. A single semver string is generated that includes all the versions.  Where are the multiple ranges being sourced from? Is this something you're generating to cover use cases? If that's the case I'd highly recommend trying to get your fixture data to mirror that of the npm advisory endpoint.  Having to generate your own ids here seems like a solution to a problem rooted in patterning your fixtures in a way that doesn't match the real advisory data.

Again I admittedly don't know a lot about the internals here, but having an array of `vulnerable_versions` for a single advisory doesn't match what I recall from the npm advisory endpoints.

bdragon: I went back and reproduced the debugging that led us down this path in the first place, by logging the response from the bulk audit endpoint (`/-/npm/v1/security/advisories/bulk`) in Arborist using a local npmcli checkout (`@v8.5.1`).

```javascript
{
  'node-forge': [
    {
      id: 1070354,
      url: 'https://github.com/advisories/GHSA-2r2c-g63r-vccr',
      title: 'Improper Verification of Cryptographic Signature in `node-forge`',
      severity: 'moderate',
      vulnerable_versions: '<1.3.0',
      cwe: [Array],
      cvss: [Object]
    },
    {
      id: 1070355,
      url: 'https://github.com/advisories/GHSA-x4jg-mjrx-434g',
      title: 'Improper Verification of Cryptographic Signature in node-forge',
      severity: 'high',
      vulnerable_versions: '<1.3.0',
      cwe: [Array],
      cvss: [Object]
    },
    {
      id: 1070356,
      url: 'https://github.com/advisories/GHSA-cfm4-qjh2-4765',
      title: 'Improper Verification of Cryptographic Signature in node-forge',
      severity: 'high',
      vulnerable_versions: '<1.3.0',
      cwe: [Array],
      cvss: [Object]
    },
    {
      id: 1067452,
      url: 'https://github.com/advisories/GHSA-8fr3-hfg3-gpgp',
      title: 'Open Redirect in node-forge',
      severity: 'moderate',
      vulnerable_versions: '<1.0.0',
      cwe: [Array],
      cvss: [Object]
    },
    {
      id: 1067471,
      url: 'https://github.com/advisories/GHSA-5rrq-pxf6-6jx5',
      title: 'Prototype Pollution in node-forge debug API.',
      severity: 'low',
      vulnerable_versions: '<1.0.0',
      cwe: [Array],
      cvss: [Object]
    },
    {
      id: 1067473,
      url: 'https://github.com/advisories/GHSA-gf8q-jrpm-jvxq',
      title: 'URL parsing in node-forge could lead to undesired behavior.',
      severity: 'low',
      vulnerable_versions: '<1.0.0',
      cwe: [],
      cvss: [Object]
    }
  ]
}
```

@wraithgar is this not what you're expecting?

wraithgar: Ah yes ok I think I understand.  The question now becomes "do the advisories being passed into `findVulnerableDependencies` have their own id attribute?".  If they do then the `auditToBulk` method should be a drop-in replacement for `convertAdvisoriesToRegistryBulkFormat`. If not, then that's another matter altogether.

bdragon: Currently the advisories that get passed to the script do not have an `id` but I wonder we couldn't fix that easily... I'll investigate. Thanks!

wraithgar: If you're sourcing them from the github advisory db and not the npm one that would explain it.  Those ids are added by the npm database and are generally unrelated to the advisory itself (in fact currently those ids change if any data in the advisory itself changes).  Arborist et al have no limits on what that id can be, so you may be able to get away with setting the `ghsaId` as the id of the object.

wraithgar: fwiw [this is the code](https://github.com/npm/security-microservice/blob/master/lib/utils/github-gql-adapter.js) that pulls from github to npm for advisories.  I think the decision to not tether ids to the ghsaId was not the best, in hindsight, and if this works here (using the ghsaId as id) it could be the beginning of a conversation w/ the registry team to get that rectified.

nlf:
i agree with @feelepxyz, this looks great! good test cases, and the usage looks appropriate (aside from the comments @wraithgar brought up)

we _definitely_ have some work to do to make reading npm configs a more reasonable process, but what you've done here is likely about as good as you can expect today. we're reaching the point where we'll be thinking about what our public api surface area looks like, and it feels pretty likely that configs will be one of the first things on deck.

in addition to that, i see absolutely no reason we couldn't extend arborist to allow you to provide it with a set of advisories that it can use instead of retrieving them via the bulk request

npm_and_yarn/helpers/lib/npm/vulnerability-auditor.js
bdragon: We recognize and agree that mocking the bulk advisory db response is not ideal and we would like to find a better approach as soon as that's feasible. To hopefully satisfy the question for now, I want to provide some additional explanation for our decision.

First, we're trying to resolve at a single dependency at a time. It looks like the default behavior of `npm audit [fix]` is to report/fix advisories on all dependencies in the tree. By mocking the bulk audit response, we're able to narrow the scope of the report/fix to just the dependency we care about and its effects.

Second, the security advisories passed to the script come from the GitHub advisory db. We want to avoid any risk of the npm db being out of sync with ours, and controlling the bulk audit response does give us that.

bin/dry-run.rb
mctofu: @feelepxyz I've tried to improve the docs around that call in https://github.com/dependabot/dependabot-core/pull/5239/commits/a70b5435889ead31f8de739319e76d2c6945b9b0. Does that sound accurate to you?

I also spotted a small optimization in https://github.com/dependabot/dependabot-core/pull/5239/commits/2b2cc38665b037dd07e9b18d492d85a2789595bc. Is there any reason we'd need to make a 2nd call to `checker.updated_dependencies(requirements_to_unlock: reqs_to_unlock)`?

Liar1975:
@nlf

npm_and_yarn/helpers/lib/npm/vulnerability-auditor.js
mctofu: It looks like we'll also need to populate any scoped/global registry settings from the npm config here to support private registries.

Nishnha:
Thank you for the review @feelepxyz, @wraithgar, @nlf 🌟 

We noticed a bug involving npm private registries that we're working to iron out, but other than that, we seem to be in good shape on this PR!

If we find that the GHSA ID can be used as the internal ID in Arborist, we'll start a discussion with the package registry/AdvisoryDB teams to see how the npm + GHSA DBs can be more in parity.

We would also love to work with the npm-cli team to extend Arborist so that Dependabot could provide it with a set of advisories (instead of Arborist retrieving them via the bulk request endpoint). I think the start of that discussion will probably have to be pushed off to a later date though as it falls outside of the scope of this PR.


Looking forward to continue improving the npm ecosystem together in the future!

wraithgar:
> If we find that the GHSA ID can be used as the internal ID in Arborist, we'll start a discussion with the package registry/AdvisoryDB teams to see how the npm + GHSA DBs can be more in parity.

This would make quite a few people in the nodejs community very happy, as the changing ids breaks most of the tooling around ignoring advisories.

bdragon (approved):

Nishnha (approved):

npm_and_yarn/helpers/lib/npm/vulnerability-auditor.js
mctofu: I followed up with populating registry settings from the npm config in https://github.com/dependabot/dependabot-core/pull/5313. This ended up being a bit awkward as to get the credentials we needed to bypass the npm cli's result (which omit credentials) and read them directly from `.npmrc`

PR #5250, opened 2022-06-09
cargo: handle patch with path that is a submodule

jakecoffman:
In Cargo.toml a user can add this to override a dependency:

```
[patch.crates-io]
uuid = { path = './path/to/uuid' }
```

This doesn't work in Dependabot if `path/to/uuid` is a submodule since the file doesn't exist in this repository. So I've added the "fetch_submodules: true" flag which handles this case.

### DIFF
--- a/cargo/spec/dependabot/cargo/file_fetcher_spec.rb
+++ b/cargo/spec/dependabot/cargo/file_fetcher_spec.rb
@@ -410,6 +410,34 @@ RSpec.describe Dependabot::Cargo::FileFetcher do
       end
     end
 
+    context "that is in a submodule" do
+      before do
+        # This file doesn't exist because sub_crate is a submodule, so returns a 404
+        stub_request(:get, url + "lib/sub_crate/Cargo.toml?ref=sha").
+          with(headers: { "Authorization" => "token token" }).
+          to_return(status: 404, headers: json_header)
+        # This returns type: submodule, we're in the common submodule logic now
+        stub_request(:get, url + "lib/sub_crate?ref=sha").
+          with(headers: { "Authorization" => "token token" }).
+          to_return(status: 200, headers: json_header, body: fixture("github", "contents_cargo_submodule.json"))
+        # Attempt to find the Cargo.toml in the submodule's repo.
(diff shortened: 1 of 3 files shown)


cargo/spec/dependabot/cargo/file_fetcher_spec.rb
```diff
@@ -403,6 +410,35 @@
...
+        stub_request(:get, submodule_root + "/contents/Cargo.toml?ref=453df4efd57f5e8958adf17d728520bd585c82c9 ").
+          with(headers: { "Authorization" => "token token" }).
+          to_return(status: 200, headers: json_header, body: fixture("github", "contents_cargo_manifest.json"))
+      end
+
+      it "places the found Cargo.toml in the correct directories" do
+        expect(file_fetcher_instance.files.map(&:name)).
+          to match_array(%w(Cargo.toml lib/sub_crate/Cargo.toml))
+        expect(file_fetcher_instance.files.map(&:path)).
+          to match_array(%w(/Cargo.toml /lib/sub_crate/Cargo.toml))
```
Nishnha, line 438: these seem like they both to the same sub_crate?

cargo/spec/dependabot/cargo/file_fetcher_spec.rb
```diff
@@ -395,6 +395,13 @@
         stub_request(:get, url + "lib/sub_crate/Cargo.toml?ref=sha").
           with(headers: { "Authorization" => "token token" }).
           to_return(status: 404, headers: json_header)
+        # additional requests due to submodule searching
```
Nishnha, line 398: I noticed that these tests are nested under the "workspace dependency" context. Does the path override only apply to cargo workspaces?

I might be misunderstanding what workspaces are for this ecosystem though if they differ from how npm implements them

jakecoffman: Yes I think once there are multiple Cargo.tomls involved it's a "workspace" but nothing else really changes, whereas with npm it looks like a workspace has to explicitly be created with `npm init -w`?

Issue #5199, opened 2022-05-27
Pub error: Unsupported operation: Trying to serialize a root package description.


mctofu:
**Package ecosystem**
Pub

**What you expected to see, versus what you actually saw**
Expected Dependabot to run successful.

Actual:
```
=> parsing dependency files
Traceback (most recent call last):
        14: from bin/dry-run.rb:533:in `<main>'
        13: from bin/dry-run.rb:282:in `cached_read'
        12: from bin/dry-run.rb:533:in `block in <main>'
        11: from /home/dependabot/dependabot-core/pub/lib/dependabot/pub/file_parser.rb:17:in `parse'
        10: from /home/dependabot/dependabot-core/pub/lib/dependabot/pub/file_parser.rb:30:in `list'
         9: from /home/dependabot/dependabot-core/pub/lib/dependabot/pub/helpers.rb:17:in `dependency_services_list'
         8: from /home/dependabot/dependabot-core/pub/lib/dependabot/pub/helpers.rb:46:in `run_dependency_services'
         7: from /home/dependabot/dependabot-core/common/lib/dependabot/shared_helpers.rb:49:in `in_a_temporary_directory'
         6: from /home/dependabot/dependabot-core/common/lib/dependabot/shared_helpers.rb:49:in `chdir'
         5: from /home/dependabot/dependabot-core/common/lib/dependabot/shared_helpers.rb:49:in `block in in_a_temporary_directory'
         4: from /home/dependabot/dependabot-core/pub/lib/dependabot/pub/helpers.rb:52:in `block in run_dependency_services'
         3: from /home/dependabot/dependabot-core/common/lib/dependabot/shared_helpers.rb:168:in `with_git_configured'
         2: from /home/dependabot/dependabot-core/pub/lib/dependabot/pub/helpers.rb:59:in `block (2 levels) in run_dependency_services'
         1: from /home/dependabot/dependabot-core/pub/lib/dependabot/pub/helpers.rb:59:in `chdir'
/home/dependabot/dependabot-core/pub/lib/dependabot/pub/helpers.rb:71:in `block (3 levels) in run_dependency_services': dart pub failed: Unsupported operation: Trying to serialize a root package description. (Dependabot::DependabotError)
package:pub/src/source/root.dart 69:5                    ResolvedRootDescription.serializeForLockfile
package:pub/src/command/dependency_services.dart 241:24  _source
[...]

PR #5272, opened 2022-06-16
Update pub revision

sigurdm:
This should fix: https://github.com/dependabot/dependabot-core/issues/5199 again.

### DIFF
--- a/pub/helpers/pubspec.lock
+++ b/pub/helpers/pubspec.lock
@@ -138,8 +138,8 @@ packages:
     dependency: "direct main"
     description:
       path: "."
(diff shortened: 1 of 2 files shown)


sigurdm:
cc @mctofu

sigurdm:
@mctofu  Oh - I see you already fixed this - thanks!

mctofu:
@sigurdm: Would it be worth bumping up to this version still? I only bumped it to the version that fixed the missing lockfile issue but would be happy to pull in this update too.

sigurdm:
There is nothing new for dependabot, but it wouldn't hurt.

mctofu (approved):

PR #5278, opened 2022-06-16
Revert "build(deps-dev): bump phpstan/phpstan from 1.6.8 to 1.7.14 in /composer/helpers/v2"

jakecoffman:
Reverts dependabot/dependabot-core#5266

May have broken CI


Nishnha:
I think the failing test is requesting a malformed URL? Looks like it has an `:` tacked on at the end.
https://github.com/dependabot/dependabot-core/runs/6920582472?check_suite_focus=true#step:9:304

Not sure what changed in the phpstan to cause that one

Issue #4080, opened 2021-07-23
Terraform JSON Syntax is not supported


captn3m0:
[Terraform also supports an alternative syntax that is JSON-compatible](https://www.terraform.io/docs/language/syntax/json.html). Terraform expects native syntax for files named with a .tf suffix, and JSON syntax for files named with a .tf.json suffix.

However, this syntax is not supported by dependabot. As an example:

```json
{
  "module": [
    {
      "kernel": [
        {
          "source": "git::https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git?ref=v5.0"
        }
      ]
    }
  ]
}
```

is a simple terraform module in JSON syntax that should get upgraded. I've tried both `.json`, and `.tf.json` extensions: https://github.com/captn3m0/dependabot-github-test/tree/main/json, but it doesn't get picked up. However, the plain `.tf` syntax does: https://github.com/captn3m0/dependabot-github-test/blob/main/kernel/main.tf (PR at https://github.com/captn3m0/dependabot-github-test/pull/7)

xlgmokha:
TIL: Thank you @captn3m0!

captn3m0:
not stale

Frulfump:
> 👋 This issue has been marked as stale because it has been open for 2 years with no activity. You can comment on the issue to hold stalebot off for a while, or do nothing. If you do nothing, this issue will be closed eventually by the stalebot. Please see CONTRIBUTING.md for more policy details.

Not stale, waiting for maintainer feedback.

PR #5293, opened 2022-06-20
Add Terraform JSON support

melendezd:
This PR addresses #4080

Adding support for updating Terraform JSON files with `.tf.json` file extension. Most of the tests present for `.tf` files are duplicated for `.tf.json` files (with some restructuring of `file_updater_spec.rb` to make this easier and less repetitious) since the functionality should be essentially the same for the two formats.

Currently, the modified file updater pulls the JSON into an object, modifies the object, and `JSON.pretty_generate`'s it out to the file. This may change the formatting of the original JSON file (tab width, for example). I'm not sure if this is okay or if there is a better approach to avoid this issue.


huajie2020 (commented):
- [ ]

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -73,27 +87,86 @@ def updated_terraform_file_content(file)
...
+
+        module_name = dependency.name.split("::").first
+        tf = JSON.parse(updated_content)
+        tf_module = tf["module"][module_name] || []
+        tf_module.each do |obj|
+          obj["source"].sub!(url_regex) do |url_match|
+            url_match.sub(old_req[:source][:ref], new_req[:source][:ref])
+          end
+        end
+        JSON.pretty_generate(tf)
```
pavera, line 117: As you mentioned in your PR comment this may reformat the json which I think we'd prefer to avoid to limit noisy PRs at the very least.  I'm not a regex expert, but I think we should take a shot at building a regex to match/sub for json.  This would be the final piece to prevent branching as we could just handle this new regex the same as we currently do for terragrunt/terraform in `git_declaration_regex`

pavera (commented):
Thanks for the PR @melendezd!  I look forward to working through this PR with you.

terraform/lib/dependabot/terraform/file_updater.rb
melendezd: Thanks for the feedback! I think you got to the bottom of it. Really the only reason for the distinction between .tf and .tf.json files in the fetcher was because I foresaw having to make the distinction in the updater. I didn't initially think about the messy diffs that this approach could create -- I could give the json regex a shot.

melendezd:
I've updated the file updater to use regex substitutions in the JSON files. The provider declaration regex was quite difficult because the source and version can be declared in either order, but I believe my approach works as long as there aren't any references to named values (like `${var.name}`) after the version and before the source. It seems like the case of the version coming before the source isn't handled in the hcl format updater, so I'm wondering how important this is.

jeffwidman:
@melendezd do you mind rebasing and also squashing down some of the smaller commits? We merge PR's with a merge strategy currently, not a squash strategy, so what's in the PR history is what eventually gets committed.

To be clear: No need to squash to only a single commit--if there are multiple logical changes, feel free to keep them separate for easier review, but I suspect there's not 32 separate logical change sets here... 😀

jeffwidman:
gentle nudge @melendezd

melendezd:
> gentle nudge @melendezd

My bad, this got away from me. I'm a bit busy at the moment but will let you know when I get back on this.

jeffwidman:
@melendezd any chance things have slowed down enough to get this across the finish line?

melendezd:
How's this look? :)

deivid-rodriguez:
I started to have a look:

* Some fixtures seem unused: `unparseable_json`, `provider_implicit_source_json`, `git_tags_013_json`.
* Needs a rebase since this conflicts with #6537.

abdulapopoola:
@njculver , are you still available to help see this over the line? We'd be interested in merging it.

PR #5294, opened 2022-06-21
Refactor CI Workflow for cache re-use

mattt:
Our current CI workflow attempts to build and push multiple images in parallel before running tests. Not only does this create extra work, but we have reason to believe that it's creating race conditions that result in unpredictable runs.  

This PR makes the following changes:

- [x] Moves the build and push steps into a new `build` job
- [x] Moves the subsequent test steps into a [dependent](https://docs.github.com/en/actions/using-workflows/about-workflows#creating-dependent-jobs) `test` job
- [x] Stops tagging branch image as `dependabot/dependabot-core:latest` (this isn't currently pushed anyway)
- [x] Derives CI container images from `github.repository`, so that forks can run CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,19 +5,18 @@ on:
       - "main"
   pull_request:
     paths-ignore:
-      - 'CHANGELOG.md'
-      - 'common/lib/dependabot/version.rb'
+      - "CHANGELOG.md"
+      - "common/lib/dependabot/version.rb"
     branches:
       - "main"
   schedule:
     - cron: "0 0 * * *"
-permissions:
-  contents: read
-  packages: write
+
 jobs:
-  ci:
-    name: CI
+  test:
+    name: Test
     runs-on: ubuntu-latest
+
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -87,12 +65,44 @@ jobs:
...
             .
       - name: Push dependabot-core-ci image to GHCR
-        env:
-          ACCESS_CANARY: ${{ secrets.ACCESS_CANARY }}
-        if: env.ACCESS_CANARY != ''
         run: |
           docker push "$CORE_CI_IMAGE:latest"
           docker push "$CORE_CI_IMAGE:branch--$BRANCH_REF"
+
+  test:
```
jakecoffman, line 72: Why break out test into another job?

mattt: At a high-level, the idea is to build a single CI image, and then run tests for each ecosystem on that image. The idea being that building once is better than building separately for each ecosystem.

brrygrdn (commented):
This is a big improvement, our current workflow is way too baroque.

The only potential issue is that we lose caching the `dependabot/dependabot-core-ci` itself to avoid building it additively from `dependabot/dependabot-core`, but I'm kind of ambivalent about that and I think we can effectively bring it back buy building the 'golden' image of `dependabot/dependabot-core-ci` on the `main` branch only and caching from it if we need the speed boost.

This should also remove some of the churn around what we push to ghcr.io as we only build golden images off main and therefore your CI in most cases would just use them.

PR #5305, opened 2022-06-24
Rebuild the core container before building ci

mctofu:


mattt:
I opened #5307 with equivalent changes targeting the `main` branch.

PR #5310, opened 2022-06-24
fix cargo errors in dependencies that have submodules

jakecoffman:
## Problem

When using a dependency that has a submodule such as this:

```
glommio = { git = "ssh://git@github.com/DataDog/glommio.git", tag = "v0.6.0" }
```

We see strange DNS-related errors in the Dependabot logs:

```
Updating crates.io index
    Updating git repository `https://github.com/DataDog/glommio.git`
    Updating git submodule `git://git.kernel.dk/liburing`
warning: spurious network error (2 tries remaining): failed to resolve address for git.kernel.dk: Temporary failure in name resolution; class=Net (12)
warning: spurious network error (1 tries remaining): failed to resolve address for git.kernel.dk: Temporary failure in name resolution; class=Net (12)
error: failed to get `glommio` as a dependency of package `git-submodule-unresolvable v0.1.0 (/home/dependabot/dependabot-core/dependabot_tmp_dir)`

Caused by:
  failed to load source for dependency `glommio`

Caused by:
  Unable to update https://github.com/DataDog/glommio.git?tag=v0.6.0

Caused by:
  failed to update submodule `glommio/liburing`

Caused by:
  failed to fetch submodule `glommio/liburing` from git://git.kernel.dk/liburing

Caused by:
  network failure seems to have happened
  if a proxy or similar is necessary `net.git-fetch-with-cli` may help here
  https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli

Caused by:
 failed to resolve address for git.kernel.dk: Temporary failure in name resolution; class=Net (12)
```

I found that the issue was that `cargo` uses git (or rather a git library) to get the submodule like so: `git fetch --tags --force --update-head-ok 'git://git.kernel.dk/liburing' '+refs/heads/*:refs/remotes/origin/*' '+HEAD:refs/remotes/origin/HEAD'` which uses the git protocol and not https.

Normally we'd configure git to use the https protocol for each host, but here since the host isn't known to Dependabot at this point it isn't configured.

## Solution

I thought the simplest solution is to make a more general git configuration that isn't dependent on knowing every host ahead of time. So now no matter which host is present, git will be able to clone it. 

This is a wider reaching change than I would want to make but it seems like it might be ok! Let me know if you think of any downsides to this approach as I couldn't think of any.

### DIFF
--- a/cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
+++ b/cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
@@ -352,7 +352,7 @@ module Dependabot
           FileUtils.mkdir_p(".cargo")
           # Use git for git operations instead of the built-in libgit
           # so that it picks up our configuration.
-          File.write("[net]\ngit-fetch-with-cli = true")
+          File.write(".cargo/config.toml", "[net]\ngit-fetch-with-cli = true")
         end
 
         def git_dependency_version
--- a/cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
+++ b/cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
@@ -152,7 +152,7 @@ module Dependabot
 
         def write_temporary_dependency_files(prepared: true)
           write_manifest_files(prepared: prepared)
-          write_cargo_config
+
           File.write(lockfile.name, lockfile.content) if lockfile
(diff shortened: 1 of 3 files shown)


jakecoffman:
Ah, the composer error is showing the `:` in `git@` style uris will not work with this approach.

PR #5311, opened 2022-06-24
Prioritize target dependency in transitive dependency updates

mctofu:
When responding to rebase requests Dependabot will re-run the job using the first dependency in the PR. Since we aren't enforcing the target dependency is at the front of the list the rebase could run against a different dependency and produce a different result.

I only wanted to re-order the results with https://github.com/dependabot/dependabot-core/commit/7c2c02ae85762d8b7463feae5c098a34dd2c0374 but had to do a little bit more to satisfy Rubocop's complexity requirements.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb
@@ -178,7 +178,9 @@ module Dependabot
           )
         end
 
-        updated_deps
+        # Target dependency should be first in the result to support rebases
+        updated_deps.select { |dep| dep.name == dependency.name } +
+          updated_deps.reject { |dep| dep.name == dependency.name }
       end
 
       def build_updated_dependency(update_details)
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb
@@ -142,18 +142,12 @@ module Dependabot
       end
 
       def conflicting_updated_dependencies
-        top_level_dependencies = FileParser.new(
-          dependency_files: dependency_files,
-          credentials: credentials,
-          source: nil
-        ).parse.select(&:top_level?)
-
-        top_level_dependency_lookup = top_level_dependencies.map { |dep| [dep.name, dep] }.to_h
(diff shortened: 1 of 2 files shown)


npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb
```diff
@@ -178,7 +172,19 @@ def conflicting_updated_dependencies
           )
         end
 
-        updated_deps
+        # Target dependency should be first in the result to support rebases
```
Nishnha, line 175: Nice approach

Nishnha (approved):

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb
```diff
@@ -178,7 +172,19 @@ def conflicting_updated_dependencies
...
+        updated_deps.select { |dep| dep.name == dependency.name } +
+          updated_deps.reject { |dep| dep.name == dependency.name }
```
bdragon, line 177: This could be simplified to

```ruby
updated_deps.partition {|d| d.name == dependency.name }.inject(:+)
```
but that's probably too clever.

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb
```diff
@@ -178,7 +172,19 @@ def conflicting_updated_dependencies
...
+      end
+
+      def top_level_dependency_lookup
```
bdragon, line 180: Thought: this seems kind of expensive, maybe we should memoize it? But we can always cross that bridge if we use this method elsewhere.

bdragon (approved):

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb
mctofu: Yup, I figured we'd wait to see if it gets used elsewhere first.

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb
mctofu: Neat! I did have to google to figure that one out so I'm leaning towards sticking with this as a more beginner friendly option.

PR #5313, opened 2022-06-24
Configure arborist registries from npm config

mctofu:
To support dependencies in private registries we need to forward the credentials we've passed to npm along to arborist.

I was able to extract registry scopes from the existing npm config we parse from the `npm config list --json` command. Unfortunately that doesn't return credentials so I needed to do a second pass to read them directly from the project's `.npmrc` file.

In the future we hope to be able to replace this with a npm maintained library to load the configuration.


Nishnha (approved):
This looks good to me. I cross-checked the keys being searched against https://www.npmjs.com/package/@npmcli/arborist and it seems like this will check everything including scoped registries/tokens.

PR #5321, opened 2022-06-28
[npm] Prevent cycles when traversing vuln effects in helper

bdragon:


npm_and_yarn/helpers/lib/npm/vulnerability-auditor.js
```diff
@@ -148,21 +149,44 @@ function convertAdvisoriesToRegistryBulkFormat(advisories) {
...
+
+    // The item that will represent this dependency in this chain.
+    const item = { name, version }
+
+    // Set containing the names of all dependencies in this chain up to and including this one;
+    // used for preventing cycles.
+    const seen = new Set([name, ...chain.items.map(i => i.name)])
+
+    // Array of effects, excluding cycles.
+    const effects = [...vuln.effects].filter(effect => !seen.has(effect.name))
```
Nishnha, line 168: Could an effect have the same name but a different vulnerability? We might need to compare the vulns as well if that's the case.

bdragon: Since we're starting from a single dependency, it won't matter whether it has 1 vuln associated with it or 10—its `effects` are simply the names of other dependencies that are vulnerable as a result of depending on it, and so on up the chain.

npm_and_yarn/helpers/lib/npm/vulnerability-auditor.js
```diff
@@ -153,21 +154,56 @@ function convertAdvisoriesToRegistryBulkFormat(advisories) {
...
+      // We currently throw when this happens. Ultimately we want to gracefully handle
+      // cycles and still return the recommended fix updates.
+      const source = chain.items[chain.items.length-1]
+      const message = `Cycle detected while traversing effects from ` +
+        `${source.name}@${source.version}: ` +
+        inspect([name, ...visited], {
+          breakLength: Infinity,
+          depth: 1,
+          maxStringLength: 255,
+        })
```
Nishnha, line 182: What does this message look like when it's rendered?

Nishnha (approved):
This and the test look good. Only concern is around what the error message looks like when it's rendered since it could be 10s of lines for a big cycle

mctofu (approved):

Issue #4483, opened 2021-12-03
Dependabot trying to download setup.py after move to pyproject.toml


Panaetius:
**Package ecosystem**
poetry (pip)
**Package manager version**
1.1.11
**Language version**
Python 3.7
**Manifest location and content prior to update**
https://github.com/SwissDataScienceCenter/renku-python/blob/dependabot-updates/pyproject.toml

**dependabot.yml content**
https://github.com/SwissDataScienceCenter/renku-python/blob/dependabot-updates/.github/dependabot.yml

**What you expected to see, versus what you actually saw**
I expect PRs to be opened with version updates, but instead Dependabot errors with 
```
Dependabot couldn't fetch all your path-based dependencies

The affected dependencies were setup.py.

To use path-based dependencies with Dependabot the paths must be relative and resolve to a directory in this project's source code.
```

```

  proxy | time="2021-12-03T09:16:26Z" level=info msg="proxy starting" commit=d5f262668736016da1a91e42cb4fba36a081bddf
  proxy | 2021/12/03 09:16:26 Listening (:1080)
updater | 2021-12-03T09:16:26.805063597 [anonymous-instance:main:WARN:src/firecracker/src/main.rs:370] You are using a deprecated parameter: --seccomp-level 2, that will be removed in a future version.
updater | 2021-12-03T09:16:26.904453390 [239653965:main:WARN:src/devices/src/legacy/serial.rs:432] Detached the serial input due to peer close/error.
updater | time="2021-12-03T09:16:30Z" level=info msg="guest starting" commit=8e918e4cf121d74a5b43e170ec4a717c1df98819
updater | time="2021-12-03T09:16:30Z" level=info msg="starting job..." fetcher_timeout=5m0s job_id=239653965 updater_timeout=45m0s updater_version=0.169.3-388839f8e94e2e09f180f781975462a82ae23d04
updater | I, [2021-12-03T09:16:33.533694 #6]  INFO -- sentry: ** [Raven] Raven 3.1.2 ready to catch errors
updater | warning: parser/current is loading parser/ruby27, which recognizes
updater | warning: 2.7.5-compliant syntax, but you are running 2.7.1.
updater | warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
updater | INFO <job_239653965> Starting job processing
[...]

PR #5324, opened 2022-06-29
Allow Python projects that only have a ``setup.py``

DanielNoord:
Closes https://github.com/dependabot/dependabot-core/issues/4483
Based on the solution provided by @jurre in https://github.com/dependabot/dependabot-core/issues/4483#issuecomment-1035199337.

As I said in https://github.com/dependabot/dependabot-core/issues/4483#issuecomment-1078832206 if I could get some guidance I'd happily provide tests or additional information, but I have no experience with Ruby so I don't really know how to.

However, as the Python ecosystem is moving away from `setup.py` I really think this change should be landed. Hopefully by providing the PR we can get some traction on the issue. If this is not the preferred way to do so, please feel free to close.

### DIFF
--- a/python/lib/dependabot/python/file_fetcher.rb
+++ b/python/lib/dependabot/python/file_fetcher.rb
@@ -290,7 +290,7 @@ module Dependabot
               fetch_submodules: true
             ).tap { |f| f.support_file = true }
           rescue Dependabot::DependencyFileNotFound
-            raise unless allow_pyproject
+            raise unless allow_pyproject || setup_cfg_file
 
             fetch_file_from_host(
               path.gsub("setup.py", "pyproject.toml"),


jurre:
Thanks @DanielNoord, I can look into adding a test for this ^^

jurre:
While trying to reproduce this in order to write a test, and I can't seem to 😅 do yo have a repo handy where this is happening?

jurre:
Thanks! I can just point a local version dependabot at those public repo's to debug further, this should get me unstuck 👍

jurre:
@DanielNoord should the title of this PR be: `Allow Python projects that have a pyproject.toml + setup.cfg`? That seems to be what causes the error I think? Only a pyproject is fine, only a setup.cfg is fine also, but the combination causes an issue?

DanielNoord:
> @DanielNoord should the title of this PR be: `Allow Python projects that have a pyproject.toml + setup.cfg`? That seems to be what causes the error I think? Only a pyproject is fine, only a setup.cfg is fine also, but the combination causes an issue?

I haven't tested that, but that could be the case.

To give some background. The ecosystem is moving towards putting everything in `pyproject.toml` but many of the build tools don't fully support that yet. The traditional setup was to have `setup.py` or `setup.cfg` or both. Because of the slow move towards `pyproject.toml` I think now all 6 possible combinations of the three files should now be allowed.

The title should probably be `Allow Python projects that do not have a "setup.py"` as 1) `pyproject.toml`, 2) `setup.cfg` and 3) `pyproject.toml` + `setup.cfg` should all be allowed. I'm not sure about the current support of case 1 and 2, but I think after this PR all 3 should be okay.

jurre:
Thanks for that context! It seems like only a pyproject.toml and only a setup.cfg are already working as expected, there's tests for those cases as well:

https://github.com/dependabot/dependabot-core/blob/c0c5fc256831519c43395281e6fa30921f9bb70c/python/spec/dependabot/python/file_fetcher_spec.rb#L290

https://github.com/dependabot/dependabot-core/blob/c0c5fc256831519c43395281e6fa30921f9bb70c/python/spec/dependabot/python/file_fetcher_spec.rb#L219

jurre:
Ah ok so for your repo it seems Dependabot wants to fetch a setup.py because there is a `requirements-test.txt` requirements file, and it assumes this needs a setup.py, if I understand correctly this is not or no longer true?

DanielNoord:
> Ah ok so for your repo it seems Dependabot wants to fetch a setup.py because there is a `requirements-test.txt` requirements file, and it assumes this needs a setup.py, if I understand correctly this is not or no longer true?

No! I think there is no case where `setup.py` is necessary any more.

See:
https://setuptools.pypa.io/en/latest/userguide/quickstart.html#setup-py

This is the tool that originally created the need for `setup.py`. They actively discourage creating a `setup.py` for any new project and only recommend using `setup.py` whenever you need "editable installs". This last case is almost moot as well because of this beta feature that is almost ready for release: https://discuss.python.org/t/help-testing-pep-660-support-in-setuptools/16904/

Dependabot shouldn't really care about whether somebody wants "editable installs" so as far as I can see there is no need for any type of checking whether a `setup.py` file exists. Based on my own experience and the documentation there is no case where it is required for dependency resolvement or installation of a package.

jurre:
> there is no need for any type of checking whether a setup.py file exists

Yeah I agree, however this code is currently lumped in with the code that pulls in path dependencies, so it's a little tricky to get it to always behave correctly, and I don't think my proposed fix really cuts it, because it assumes there is always a `setup.py` _or_ a `setup.cfg` when there is a `requirements.*.txt` present, and if I understand correctly this is also not correct?

DanielNoord:
> > there is no need for any type of checking whether a setup.py file exists
> 
> Yeah I agree, however this code is currently lumped in with the code that pulls in path dependencies, so it's a little tricky to get it to always behave correctly, and I don't think my proposed fix really cuts it, because it assumes there is always a `setup.py` _or_ a `setup.cfg` when there is a `requirements.*.txt` present, and if I understand correctly this is also not correct?

No, that's indeed not correct. I'm trying to find an example of a package that doesn't have this but I can't quickly find one. Probably because many project have kept either of those files around to make sure they keep getting dependabot updates 😄

jurre:
@DanielNoord I had to dig a bit deeper and eventually was able to fix it here: https://github.com/dependabot/dependabot-core/pull/5392

Would you mind taking a look at this, especially interested if I've described the problem correctly and if my assumptions on the pip side are correct.

PR #5332, opened 2022-06-30
use a "shim" to fix inability to rewrite git commands to use HTTPS URLs

jakecoffman:
Follow-up to https://github.com/dependabot/dependabot-core/pull/5310

Summary of the issue: When invoking `cargo update`, cargo can make git calls to non-HTTP endpoints which fail because our container only allows connections through the HTTP proxy.

My first attempt was to use git's `.insteadOf` feature, but it isn't advanced enough to handle every host using the SCP-style URLs such as `git@example.com:org/user` because of the colon.

So I created a git "shim" which sits on the path ahead of the real git, rewrites the arguments it receives for the clone and fetch commands, and then invokes the real git with the same arguments.

Initially this seems a bit hacky, but with this in place we'll be able to remove our [monkey patches of Bundler](https://github.com/dependabot/dependabot-core/blob/c1e159a29ff4c817f37a690aa33335278d1291d0/bundler/helpers/v1/monkey_patches/git_source_patch.rb), and code that does similar URL rewriting in several of the ecosystems. This made me feel better about doing this as it will clean up a lot of our code and centralize the URL rewrite logic. 

The shim code is here, please review it as well: https://github.com/dependabot/git-shim

With regards to testing this functionality in CI, I think the cargo issue this addresses is currently untestable as our CI isn't executing in a network that only allows HTTP access through a proxy. So even without the shim it would pass. 🤔


jurre (approved):
🎲 😄

Dockerfile
```diff
@@ -292,3 +292,10 @@ RUN bash /opt/terraform/helpers/build
 ENV HOME="/home/dependabot"
 
 WORKDIR ${HOME}
+
+# Place a git shim ahead of git on the path to rewrite git arguments to use HTTPS.
+ARG SHIM="https://github.com/dependabot/git-shim/releases/download/v1.4.0/git-v1.4.0-linux-amd64.tar.gz"
+RUN curl -sL $SHIM -o git-shim.tar.gz && mkdir -p ~/bin && tar -xvf git-shim.tar.gz -C ~/bin && rm git-shim.tar.gz
+ENV PATH="$HOME/bin:$PATH"
+# Configure cargo to use git CLI so the above takes effect
+RUN mkdir -p ~/.cargo && printf "[net]\ngit-fetch-with-cli = true\n" >> ~/.cargo/config.toml
```
jeffwidman, line 301: You mentioned in the PR description that although this is a bit of a hack, it'll allow us to replace rewriting in multiple places in our code, which is a definite 🏆 . But to do that, all native helpers will need to be reconfigured to query git via cli and not native wire libraries... from those that you saw, did that option look possible in those helpers? Otherwise it'll prevent some of that 🏅...

cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
```diff
@@ -188,7 +188,8 @@ def handle_cargo_errors(error)
           end
 
           if error.message.include?("authenticate when downloading repo") ||
-             error.message.include?("HTTP 200 response: got 401")
+             error.message.include?("HTTP 200 response: got 401") ||
+             error.message.include?("fatal: Authentication failed for")
```
jeffwidman, line 192: Was this error not possible to be thrown by Cargo before?

jeffwidman (approved):
This looks reasonable enough to to me. Or rather, I completely agree with your assessment that this is pretty hacky, but at least it DRY's up the hack location so it's not scattered across all our ecosystems.

Long term, IMO we should encourage the upstream package managers to support this git-over-https feature natively, so we'd only need to pass them an arg rather than having to write a custom shim. Or do I misunderstand and there's some reason that we'd still need the shim in that case?

Dockerfile
```diff
@@ -292,3 +292,10 @@ RUN bash /opt/terraform/helpers/build
 ENV HOME="/home/dependabot"
 
 WORKDIR ${HOME}
+
+# Place a git shim ahead of git on the path to rewrite git arguments to use HTTPS.
+ARG SHIM="https://github.com/dependabot/git-shim/releases/download/v1.4.0/git-v1.4.0-linux-amd64.tar.gz"
+RUN curl -sL $SHIM -o git-shim.tar.gz && mkdir -p ~/bin && tar -xvf git-shim.tar.gz -C ~/bin && rm git-shim.tar.gz
```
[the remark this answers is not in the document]
jeffwidman: Looks like `go install` can be tweaked to support custom directory for just that one invocation: https://github.com/golang/go/issues/50553#issuecomment-1046046678

Dockerfile
jeffwidman: Totally out of scope of this PR, but maybe we should file an issue to eventually reach out to all the native helpers that we need to shim to ask them to add an option for natively use https urls... then we wouldn't need the shim here, but would instead simply pass `[net]\ngit-fetch-with-https = true` here.

If you don't have time / interest, I can reach out to those ecosystems to get the conversations started, but if you could at least file the dependabot-side issue listing all the ecosystems that we are currently patching that'd give me a checklist to start with and something I could cross-link.

jakecoffman: As a first pass I was going to get this cargo issue fixed, and then we can start removing hacks in a controlled manner per-ecosystem. I did search a good amount for an option in cargo to force https but there isn't one currently. 

That's a great idea to reach out to ask for that option in the package managers. My other thought was we could see about improving `insteadOf` functionality in Git. We have some Get contributors here at GitHub who might be able to help. However I struggled with how that would work exactly, so I went this shim route.

cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
jakecoffman: Since Cargo uses libgit by default, I think they were generating their own error. Now with `git-fetch-with-cli = true` it's using Git CLI's error message. So yes, it wasn't possible, and I decided to leave the previous error in just in case for now.

Dockerfile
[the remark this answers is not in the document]
jakecoffman: This quarter we're hoping to break the core image into smaller ecosystem-specific images. Thus in the future the Bundler image won't have Go installed, so the download method makes more sense here.

cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
jeffwidman: Did you want to add a TODO then about eventually removing the previous error? Otherwise it will likely just cause confusion/cruft for a long time...

Dockerfile
jeffwidman: > My other thought was we could see about improving insteadOf functionality in Git. We have some Get contributors here at GitHub who might be able to help. 

Yeah, that'd be an even better idea.

> However I struggled with how that would work exactly, so I went this shim route.

To clarify, are you unclear how to discover/work with the git contributors at GitHub to fix this upstream? Or even if `insteadOf` were improved to cover all these cases, do you still think we'd need the shim?

cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
jeffwidman: I added a TODO in the code, and then opened https://github.com/dependabot/dependabot-core/pull/5967 to eventually remove both the code comment and the error message...

Issue #3107, opened 2021-02-09
Dependabot::PullRequestCreator.new creates branch name that is too long


cyril-s:
**Package manager/ecosystem**
ruby:bundler

**What you expected to see, versus what you actually saw**
When dependabot tried to create a merge request in gitlab, API responded with:
```json
{
   "api_error" : [
      "{\"message\":\"Invalid reference name: 6cca8a5a83a143598d5acd494f099fcdab4105ae\"}"
   ],
   "redis_calls" : 4,
   "redis_write_bytes" : 545,
   "duration_s" : 0.31665,
   "username" : "builder",
   "correlation_id" : "OBo4uFA7Zu4",
   "db_duration_s" : 0.01291,
   "meta.caller_id" : "/api/:version/projects/:id/repository/branches",
   "route" : "/api/:version/projects/:id/repository/branches",
   "gitaly_duration_s" : 0.248659,
   "queue_duration_s" : 0.047001,
   "meta.user" : "builder",
   "redis_cache_write_bytes" : 545,
   "severity" : "INFO",
   "redis_cache_read_bytes" : 220,
   "redis_read_bytes" : 220,
   "host" : "******",
   "remote_ip" : "******",
   "redis_cache_duration_s" : 0.001662,
   "redis_duration_s" : 0.001662,
   "meta.project" : "******",
   "time" : "2021-02-09T12:22:28.100Z",
   "ua" : "Gitlab Ruby Gem 4.16.1",
   "view_duration_s" : 0.30374,
   "user_id" : 6,
   "status" : 400,
   "params" : [
      {
         "value" : "dependabot/bundler/rails-and-draper-and-rails-controller-testing-and-rspec-rails-and-font-awesome-rails-and-activeadmin-and-responders-and-sass-rails-and-devise-and-jquery-rails-and-jquery-tablesorter-and-factory_bot_rails-and-ransack-and-delayed_job_active_record-and-activerecord-partitioning-and-active_record_extended-and-active_admin_import-6.1.1",
         "key" : "branch"
      },
      {
         "key" : "ref",
         "value" : "6cca8a5a83a143598d5acd494f099fcdab4105ae"
      }
   ],
   "method" : "POST",
   "path" : "/api/v4/projects/******/repository/branches",
   "gitaly_calls" : 1,
   "meta.root_namespace" : "******",
   "redis_cache_calls" : 4
}

```

[...]

Issue #396, opened 2018-04-27
Provide option to configure format of branch name


greysteil:
_From @edkellena on January 9, 2018 15:40_

Hey,
It would be super awesome if we could specify some format for the branch name created.

Use-case:
Our CI system sees a PR raised and will run through a build process that generates a Docker image and tags it with the branch name.  Unfortunately Docker tags have more limited scope than Git branches...
`A tag name must be valid ASCII and may contain lowercase and uppercase letters, digits, underscores, periods and dashes. A tag name may not start with a period or a dash and may contain a maximum of 128 characters.`

The `/` in the Dependabot branches is causing that tagging to fail.  We can work around this with some awk/shell wizardry I guess, but would be nice not to have to do this.

_Copied from original issue: dependabot/feedback#72_

greysteil:
Thanks for the feedback @edkellena.

It's a totally legit request, but I'm not keen on changing things on Dependabot's side for a couple of reasons:
- branch name generation is already [quite complicated](https://github.com/dependabot/dependabot-core/blob/master/lib/dependabot/pull_request_creator/branch_namer.rb) and I don't want to add more considerations to it
- there's no obvious place to put the setting in the frontend at the moment

Do you have control over the build process on your side? I'm hoping it would be straightforward for you to sanitize the branch name when generating a tag.

greysteil:
_From @edkellena on January 9, 2018 22:12_

Fair enough 👍 I did wonder whether it would be a fruitless ask!   Would be happy to attempt a PR for the generation, but without the UI setting available, that would be a bit pointless.

I do have some very hacky workaround at the moment, which I'll try to unhack soon.  Cheers though!

greysteil:
Always worth asking! Thanks for understanding 🙂 .

SanjiKir:
Is it still a problem to implement? We have exactly the same issue

BobbyMcWho:
You could always monkey-patch it, now whether or not you _should_... 
```ruby
module Dependabot
  class PullRequestCreator
    class BranchNamer
      def new_branch_name
        super.gsub(TO_REMOVE, TO_REPLACE)
      end
    end
  end
end
```

ritaly:
if you use dependabot v2  just add to your `dependabot.yml` 

```
pull-request-branch-name:
      separator: "-"
```

https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#pull-request-branch-nameseparator

[...]

PR #5338, opened 2022-07-01
Add max length option to BranchNamer

TomNaessens:
This PR adds a max length option to the BranchNamer and addresses https://github.com/dependabot/dependabot-core/issues/3107. Also relevant for the "... and may contain a maximum of 128 characters" part of https://github.com/dependabot/dependabot-core/issues/396.

This is a backwards compatible change and requires no configuration: when no option is passed, no truncating is performed on the branch name.

As this is my first contribution here, let me know if I missed some conventions, or if there are additional changes expected!

### DIFF
--- a/common/lib/dependabot/pull_request_creator/branch_namer.rb
+++ b/common/lib/dependabot/pull_request_creator/branch_namer.rb
@@ -1,20 +1,23 @@
 # frozen_string_literal: true
 
+require "digest"
+
 require "dependabot/metadata_finders"
 require "dependabot/pull_request_creator"
 
 module Dependabot
   class PullRequestCreator
     class BranchNamer
-      attr_reader :dependencies, :files, :target_branch, :separator, :prefix
+      attr_reader :dependencies, :files, :target_branch, :separator, :prefix, :max_length
 
       def initialize(dependencies:, files:, target_branch:, separator: "/",
-                     prefix: "dependabot")
+                     prefix: "dependabot", max_length: nil)
         @dependencies  = dependencies
         @files         = files
         @target_branch = target_branch
         @separator     = separator
         @prefix        = prefix
+        @max_length    = max_length
       end
 
       def new_branch_name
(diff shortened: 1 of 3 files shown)


common/lib/dependabot/pull_request_creator/branch_namer.rb
```diff
@@ -37,7 +38,14 @@ def new_branch_name
           end
 
         # Some users need branch names without slashes
-        sanitize_ref(File.join(prefixes, @name).gsub("/", separator))
+        sanitized_name = sanitize_ref(File.join(prefixes, @name).gsub("/", separator))
+
+        # Shorten the ref in case users refs have length limits
+        if @max_length
+          sanitized_name[0, @max_length]
```
jurre, line 45: I'm worried that we'd end up with duplicate branch names in this scenario, as we'd end up truncating the bit that makes them unique (the version), and I think that could cause issues

nudded: That's a valid concern.

Would you be okay with taking a hash of the full-length branch name and taking the first `max_length` characters from that if `max_length` is specified? (At some point we'd have to choose the lesser of 2 evils, we cannot deal with going over the limit, however then you do lose pretty branch names)

jurre: Yeah a hash could also work, it's a tad safer I guess (if for some reason the combination of dependency names changes, it's still unique)

jeffwidman:
Approving the CI runs for first-time contributor as the PR seems headed in a generally useful direction.

TomNaessens:
Sorry, took a while to get the feedback processed. Let me know what you think of the changes!

SimonSomlai:
@TomNaessens , thanks for opening this, we've also ran into a similar problem. Seems like there are just some linter issues left (most of which can be autocorrected);

Screenshot 2022-09-13 at 10 11 37

@jeffwidman / @jurre, does this need more work, or do the changes look good? :)

TomNaessens:
> @TomNaessens , thanks for opening this, we've also ran into a similar problem. Seems like there are just some linter issues left (most of which can be autocorrected);

Thanks! Should be fixed now 🙇‍♂️

common/lib/dependabot/pull_request_creator/branch_namer.rb
jeffwidman: Very clean solution @TomNaessens 👍

jeffwidman (approved):

jeffwidman:
Rebased, once CI passes I'll deploy to production and assuming that looks good will merge. It's getting late here, so may not finish that til tomorrow though.

PR #5341, opened 2022-07-01
fix(terraform): correctly keep platform hashes when multiple providers are installed

### LINKED ISSUE #5225: Terraform updates sometimes fails to add lockfile hashes for included platforms
**Package ecosystem**
terraform

**Package manager version**
terraform 1.0.0 and terraform 1.2.2 (not entirely sure what dependabot runs)
`required_version = "1.0.0"` ([repro](https://github.com/Flydiverny/dependabot-tf-repro-1.0.0/pull/2/files))  and `required_version = ">=1.1.0, <2.0.0"` ([repro](https://github.com/Flydiverny/dependabot-tf-repro/pull/1)) both exhibit the same problem 

**Language version**
terraform 1.0.0 and terraform 1.2.2 (not entirely sure what dependabot runs)
`required_version = "1.0.0"` ([repro](https://github.com/Flydiverny/dependabot-tf-repro-1.0.0/pull/2/files))  and `required_version = ">=1.1.0, <2.0.0"` ([repro](https://github.com/Flydiverny/dependabot-tf-repro/pull/1)) both exhibit the same problem 

**Manifest location and content before the Dependabot update**
`.terraform.lock.hcl`

**dependabot.yml content**
https://github.com/Flydiverny/dependabot-tf-repro/blob/3b5e9aa2024a5a3d5cef34edea3a7b6d9fdab71e/.github/dependabot.yml#L1-L12

**Updated dependency**
`registry.terraform.io/hashicorp/aws` from 4.13.0 to 4.17.0

Expected
```terraform
  version = "4.17.0"
  hashes = [
    "h1:0i4o/1AEBOIW/fr3C45vJlG350FfRzSldJcWfViOFLc=",
    "h1:5NfNH5iQ3VGpMwjlSOkQyBYh8mvKH3QMp0UHB71eTqc=",
    "h1:FtkENM8QDK6CLhBr4k97kx26G2CqHZk9q5sNhHcnYRc=",
    "zh:2cc932fb0af13850de3c60a5318b695c82973489c140ca4f13218f69136c36e5",
    "zh:4018884d66acfa8273f7100ef0334004ed8a3790ffc7621eaef65d1d9c3fab43",
    "zh:6a7769e5c81e543f5deaaa8596e45f92244a61f5026c8c66d5bf55f2a7fd4801",
    "zh:7956c1e17ec7647af3612cc98cbcd21d50b2d9f5e41c676b62ee214f5610c29f",
    "zh:833d9d608dbffda7da565004ef592a8a364e96b5c13cacf873f5d32714e197ff",
    "zh:9b12af85486a96aedd8d7984b0ff811a4b42e3d88dad1a3fb4c0b580d04fa425",
    "zh:a55b8b72e47999d9c7aecaa009797ed7eb3f669a719d3f6127ee5e0f1b91ecc2",
    "zh:a6f2377d71dfba9669f060e687498e589b490366026821bd83451ac9ef0cd9e8",
    "zh:b006aa281097b3db27a62ea3c8cfaf4c4d979d57f4a6c180bd2da3d0ab4bd61a",
    "zh:d6a6d29256fee6c3b35384719c84c19b13dcccc53bffce5f529023607d130d0b",
    "zh:edc155e147883872e1227aa6a4ef3205fa9de421475d96c20a34a5eaff3df01f",
    "zh:f25773dcc00dead0412e222cf3891ac6228dcb4d69da9bacfca305a0a2a1db56",
  ]
}
```

Result ([see dependabot PR here](https://github.com/Flydiverny/dependabot-tf-repro/pull/1))
```terraform
  version = "4.17.0"
  hashes = [
    "h1:0i4o/1AEBOIW/fr3C45vJlG350FfRzSldJcWfViOFLc=",
    "zh:2cc932fb0af13850de3c60a5318b695c82973489c140ca4f13218f69136c36e5",
    "zh:4018884d66acfa8273f7100ef0334004ed8a3790ffc7621eaef65d1d9c3fab43",
    "zh:6a7769e5c81e543f5deaaa8596e45f92244a61f5026c8c66d5bf55f2a7fd4801",
    "zh:7956c1e17ec7647af3612cc98cbcd21d50b2d9f5e41c676b62ee214f5610c29f",
    "zh:833d9d608dbffda7da565004ef592a8a364e96b5c13cacf873f5d32714e197ff",
    "zh:9b12af85486a96aedd8d7984b0ff811a4b42e3d88dad1a3fb4c0b580d04fa425",
    "zh:a55b8b72e47999d9c7aecaa009797ed7eb3f669a719d3f6127ee5e0f1b91ecc2",
    "zh:a6f2377d71dfba9669f060e687498e589b490366026821bd83451ac9ef0cd9e8",
    "zh:b006aa281097b3db27a62ea3c8cfaf4c4d979d57f4a6c180bd2da3d0ab4bd61a",
    "zh:d6a6d29256fee6c3b35384719c84c19b13dcccc53bffce5f529023607d130d0b",
    "zh:edc155e147883872e1227aa6a4ef3205fa9de421475d96c20a34a5eaff3df01f",
    "zh:f25773dcc00dead0412e222cf3891ac6228dcb4d69da9bacfca305a0a2a1db56",
  ]
}
```

Using `terraform init -upgrade` and `terraform providers lock -platform=darwin_amd64 -platform=darwin_arm64 -platform=linux_amd64` will give the expected output as pasted above.
https://github.com/Flydiverny/dependabot-tf-repro/pull/1

The referenced repository has the smallest reproduction I was able to make.
Having just the `registry.terraform.io/hashicorp/aws` provider to upgrade the pull request from dependabot actually did include the platform hashes, and reading the log, see attached 
[dependabotlog.txt](https://github.com/dependabot/dependabot-core/files/8834074/dependabotlog.txt), it looks like it does download the platforms so I'm not sure why they are not in the PR in the end.

Flydiverny:
Fixes #5225

### DIFF
--- a/terraform/lib/dependabot/terraform/file_updater.rb
+++ b/terraform/lib/dependabot/terraform/file_updater.rb
@@ -103,6 +103,11 @@ module Dependabot
           select { |h| h&.match?(/^h1:/) }
       end
 
+      def remove_provider_h1_hashes(content, declaration_regex)
+        content.match(declaration_regex).to_s.
+          sub(hashes_object_regex, "")
+      end
+
       def lockfile_details(new_req)
         content = lock_file.content.dup
         provider_source = new_req[:source][:registry_hostname] + "/" + new_req[:source][:module_identifier]
@@ -131,7 +136,7 @@ module Dependabot
         )
 
         base_dir = dependency_files.first.directory
-        lockfile_hash_removed = content.sub(hashes_object_regex, "")
+        lockfile_hash_removed = remove_provider_h1_hashes(content, declaration_regex)
 
         # This runs in the same directory as the actual lockfile update so
         # the platform must be determined before the updated manifest files
@@ -265,7 +270,7 @@ module Dependabot
       end
 
       def hashes_object_regex
-        /hashes\s*=\s*.*\]/m
+        /hashes\s*=\s*[^\]]*\]/m
       end
 
       def hashes_string_regex


terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -131,7 +136,7 @@ def lookup_hash_architecture # rubocop:disable Metrics/AbcSize, Metrics/MethodLe
         )
 
         base_dir = dependency_files.first.directory
-        lockfile_hash_removed = content.sub(hashes_object_regex, "")
+        lockfile_hash_removed = remove_provider_h1_hashes(content, declaration_regex)
```
Flydiverny, line 139: before this would remove the hashes of the _first_ provider in the lockfile. now it removes the hashes from the provider that we are trying to update so it can properly deduce the hashes for each platform and correctly match which platforms are in use.

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -265,7 +270,7 @@ def check_required_files
       end
 
       def hashes_object_regex
-        /hashes\s*=\s*.*\]/m
+        /hashes\s*=\s*[^\]]*\]/m
```
Flydiverny, line 273: This makes the regex less greedy so it will stop within each provider block and not match over all provider blocks at the same time.

terraform/lib/dependabot/terraform/file_updater.rb
Flydiverny: I think typically we would have the same platforms in all providers, only reason I've seen different ones is because people manually update a dependency and forget to run `terraform providers lock -platform=darwin_amd64 -platform=darwin_arm64 -platform=linux_amd64` :) 

So as you say the example in my reproduction would not be what you want, but doesn't seem unlikely to happen

Flydiverny: it would be great if it could auto recover somehow, as in you could specify which platforms you actually want.
Or if it could be deduced by looking at the provider that has the most h1 hashes 🤔 
Tho I suppose typically if you depend on multiple platforms it would be something that's solved before a dependency update happens again anyway. so maybe its not really an issue.
Its unfortunate that the platform lock is so awkward for terraform (imo) 😄

Nishnha: > it would be great if it could auto recover somehow, as in you could specify which platforms you actually want

We thought about adding this feature to Dependabot in https://github.com/dependabot/dependabot-core/pull/4784#issuecomment-1064313966 but decided it would make more sense to be added upstream

I tried reaching out to the Terraform Core team about this but haven't heard back. Opening an issue in https://github.com/hashicorp/terraform for visibility might help?

The idea we proposed is similar - 

> Terraform could include something like [a Terraform resource configuration file `.tfrc` that lets users specify their platforms in plain text](https://github.com/dependabot/dependabot-core/pull/4784#issuecomment-1072700971). 
It would be helpful for tools like Dependabot to keep Terraform projects up-to-date, and also for Terraform users to more plainly define and read what architectures their projects run on.

---

> Or if it could be deduced by looking at the provider that has the most h1 hashes 🤔

This sounds like a reasonable middle ground. It might lead to some confusion though since additional platforms could be added to their provider lockfiles

FYI, right now we default to using `linux_amd64` if we can't detect the platform being used.

Nishnha (approved):

Nishnha:
Hi @Flydiverny, sorry for the late approval/merge. 

I tried to add some tests around this and found that our Terraform file_updater tests only update a single dependency (provider) at a time.

So even if we specified additional dependencies specified in https://github.com/dependabot/dependabot-core/blob/main/terraform/spec/dependabot/terraform/file_updater_spec.rb#L779, testing the updated behavior against multiple providers is non-trivial. 

I instead manually checked that the updated regex has the correct captures against a lockfile with multiple providers. The previously failing cases you reproduced in https://github.com/dependabot/dependabot-core/issues/5225 also work in a dry-run on this branch, so I'm comfortable merging this fix.

Thanks!

Flydiverny:
Cheers @Nishnha thanks merging and taking a look at the tests! :)

Issue #2217, opened 2018-11-13
Update Docker image on kubernetes deployment files


AndreaGiardini:
It would be nice to have the possibility to have the docker image update logic also for Kubernetes deployment files.

Dependabot will have to scan all yml/yaml files, get the `image` field and check for updates of the docker image.

jglick:
See also #2237.

JesperTerkelsen:
+1

jeffwidman:
Support for this was merged in https://github.com/dependabot/dependabot-core/pull/5348 🎉

Raboo:
@jeffwidman is it the `package-ecosystem: "docker"` that is used for Kubernetes YAML files?

dimisjim:
@Raboo seems to be so yeah, based on: https://github.com/dependabot/dependabot-core/pull/5348#issue-1296588620 

> It is added onto the Docker package.

as it's also the only one in the doc right now: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem

jeffwidman:
Currently this is buried behind a feature flag for folks on GitHub.com...

We'd like to start slowly rolling it out for everyone, but before I do that, I'd like to test on a few repos to ensure no unexpected issues. I can / will setup a fake test repo, but there's nothing like live data to confirm things.

Anyone watching this issue care to volunteer to have their repo(s) be part of the beta test?

Repos must match the following criteria:

* on GitHub.com
* public, because for security reasons we can't access private repos w/o jumping through some hoops
* have k8s deployment yaml files... ideally with a combination of up-to-date and outdated docker image tags

If interested, please comment on this thread with the URLs of any repos you own/maintain. Once I get a couple of volunteers, no need to keep volunteering... I expect/hope the rollout should be pretty quick, so you won't be waiting too long.

chadlwilson:
Hi @jeffwidman Is this intended to support Helm templates? Or only raw K8s YAML?

[...]

PR #5348, opened 2022-07-06
Add a dependabot implementation for Kubernetes YAML files.

brendandburns:
This addresses #2217 

~This is heavily based off of the Dockerfile updater. I basically copied the code + tests and then edited as necessary.~

~There's a lot of duplication, so if there's a better way to factor this (make it part of the docker package?) I'm open to suggestions.~

It is added onto the Docker package. If there is a better way to structure the code, let me know.

Thanks!


docker/lib/dependabot/docker/file_parser.rb
```diff
@@ -61,15 +64,18 @@ def parse
           end
         end
 
+        workflow_files.each do |file|
```
jurre, line 67: Is "workflow_file" typically how you'd refer to these config files? If that's typical for folks in the community then 👍, I've heard "manifest" I think? https://kubernetes.io/docs/reference/glossary/?fundamental=true#term-manifest

You obv have a much better idea on this, so curious to hear your thoughts

jurre (commented):
Just some general initial questions, one scenario that I'm wondering about (and will at least test manually) is if there are multiple images defined in a single manifest, for example with a sidecar deployment?

docker/lib/dependabot/docker/file_fetcher.rb
```diff
@@ -7,7 +7,8 @@ module Dependabot
   module Docker
     class FileFetcher < Dependabot::FileFetchers::Base
       def self.required_files_in?(filenames)
-        filenames.any? { |f| f.match?(/dockerfile/i) }
+        filenames.any? { |f| f.match?(/dockerfile/i) } or
+          filenames.any? { |f| f.match?(/^[^\.]+\.ya?ml/i) }
```
bdragon, line 11: nit pick: match end of line to avoid matching something like `deployment.yml.template`?

`/^[^.]+\.ya?ml$/i`

docker/lib/dependabot/docker/file_parser.rb
```diff
@@ -154,6 +160,68 @@ def check_required_files
 
         raise "No Dockerfile!"
       end
+
+      def workfile_file_dependencies(file)
```
bdragon, line 164: A k8s yaml template file may contain [multiple resource definitions](https://kubernetes.io/docs/concepts/cluster-administration/manage-deployment/#organizing-resource-configurations) separated by `---`, so we should make sure we can correctly handle that. By my testing, when given such a file Psych will parse the first one and ignore all the rest. That's acceptable for the purposes of the heuristic above, but here I think we'll want to make sure we handle all of them with something like

```ruby
def workfile_file_dependencies(file)

  resources = file.content.split(/^---$/).map(&:strip).reject(&:empty?) # assuming a yaml file
  resources.flat_map do |resource|
    json = YAML.safe_load(resource, aliases: true)
    # ...
```

docker/lib/dependabot/docker/file_parser.rb
bdragon: What about `kubernetes_resource_template_files`?

jurre:
We've talked about turning this functionality off with a feature-flag initially, so some pointers on how to do it:

- Most top-level base classes take an optional `options` argument, wish accepts a Hash:
https://github.com/dependabot/dependabot-core/blob/f614de480d65a4c110815c1c13df8374f6c47f86/common/lib/dependabot/update_checkers/base.rb#L19
https://github.com/dependabot/dependabot-core/blob/f614de480d65a4c110815c1c13df8374f6c47f86/common/lib/dependabot/file_updaters/base.rb#L14

We can pass some sort of `kube_manifests_enabled` flag to them, and check for it in the code. Here's another piece of functionality (in the npm ecosystem) that's shielded off in that way:

https://github.com/dependabot/dependabot-core/blob/f614de480d65a4c110815c1c13df8374f6c47f86/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb#L132

PS: I noticed the `FileFetcher::Base` class currently doesn't take this arg, feel free to add it, as long as it takes a default empty hash `def initialize(some:, other:, :args, options: {})` it won't break any code not passing it yet

docker/lib/dependabot/docker/file_parser.rb
brendandburns: switched to manifests, this was a copy and paste left over.

docker/lib/dependabot/docker/file_parser.rb
brendandburns: done.

brendandburns:
Comments addressed. Please take another look.

docker/lib/dependabot/docker/file_fetcher.rb
```diff
@@ -7,26 +7,44 @@ module Dependabot
...
+      def kubernetes_enabled?
+        options.key?(:kubernetes_updates) && options[:kubernetes_updates]
+      end
+
       def fetch_files
         fetched_files = []
         fetched_files += correctly_encoded_dockerfiles
+        if kubernetes_enabled?
+          fetched_files += correctly_encoded_yamlfiles
+        end
```
jurre, line 29: 👍

jurre:
@brendandburns thanks! the linting errors should be auto-correctable using `bundle exec rubocop -A -c ../.rubocop.yaml` from the `docker` directory, taking a look at the rest now

docker/lib/dependabot/docker/file_fetcher.rb
```diff
@@ -7,26 +7,44 @@ module Dependabot
   module Docker
     class FileFetcher < Dependabot::FileFetchers::Base
       def self.required_files_in?(filenames)
-        filenames.any? { |f| f.match?(/dockerfile/i) }
+        filenames.any? { |f| f.match?(/dockerfile/i) } or
+          filenames.any? { |f| f.match?(/^[^\.]+\.ya?ml$/i) }
```
jurre, line 11: Bit pedantic, but, I notice we check for this regex in a few different places, sometimes with the `$` and sometimes not, would it make sense to extract it into a constant and reference that everywhere?

docker/lib/dependabot/docker/file_fetcher.rb
```diff
@@ -7,26 +7,44 @@ module Dependabot
...
       end
 
       def self.required_files_message
-        "Repo must contain a Dockerfile."
+        "Repo must contain a Dockerfile or Kubernetes YAML files."
       end
 
       private
 
+      def kubernetes_enabled?
```
jurre, line 20: I guess since we only fetch manifests when this is true, we don't need to add an option further down the chain in the FileUpdater etc, as they will simply not be considered?

jurre:
Here's a little change I had to make to get the dry-run script to accept this new arg:

```
diff --git a/bin/dry-run.rb b/bin/dry-run.rb
index 96c792f06..af1c1e70d 100755
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -489,7 +489,8 @@ $repo_contents_path = File.expand_path(File.join("tmp", $repo_name.split("/")))
 fetcher_args = {
   source: $source,
   credentials: $options[:credentials],
-  repo_contents_path: $repo_contents_path
+  repo_contents_path: $repo_contents_path,
+  options: $options[:updater_options]
 }
 $config_file = begin
   cfg_file = Dependabot::Config::FileFetcher.new(**fetcher_args).config_file
```

But with that change, it seems to work:

```
[dependabot-core-dev] ~/dependabot-core $ bin/dry-run.rb docker dsp-testing/dependabot-all-updates-test --dir=/config/kubernetes/default/deployments --updater-options=kubernetes_updates
=> fetching dependency files
=> dumping fetched dependency files: ./dry-run/dsp-testing/dependabot-all-updates-test/config/kubernetes/default/deployments
=> parsing dependency files
=> updating 1 dependencies: nginx

=== nginx (1.14.2)
 => checking for updates 1/1
 => latest available version is 1.23.1
 => latest allowed version is 1.23.1
 => updating nginx from 1.14.2 to 1.23.1

    ± nginx.yaml
    ~~~
    19c19
    <         image: nginx:1.14.2
    ---
    >         image: nginx:1.23.1
    ~~~
```

We have a set of (still) internal integration tests, so I'm setting those up now to make it a bit easier to keep verifying any changes made to this, but overall it looks pretty good

jurre:
@brendandburns I'll extract the changes to pass options to the FileFetcher into a separate PR so I can make some changes to our internal systems that run Dependabot to properly deal with this, do you mind if I push a few small changes to this branch and rebase it on the latest main? I _think_ I should have access to your fork

brendandburns:
@jurre please go ahead and push your commits to this branch. I will address the style/constant extraction comments after that

docker/lib/dependabot/docker/file_fetcher.rb
brendandburns: Yes, I believe that is correct, since we won't fetch the manifests, they won't get parsed etc.

docker/lib/dependabot/docker/file_fetcher.rb
brendandburns: done (and for `/dockerfile/i` also)

brendandburns:
I actually went ahead and addressed your comments since they were so small. Please add your commits and then hopefully we can merge this!

jurre:
Looks like I don't have commit access to the fork actually, all I think we need to do for now is rebase this on main with the changes you've already made, and the bits I pulled out into a separate PR.

After that I'll get those e2e tests set up, test a few more edge cases and I think that's it!

brendandburns:
@jurre rebased, ptal. Thanks!

jurre:
Did some extra testing against https://github.com/dsp-testing/dependabot-all-updates-test/blob/main/config/kubernetes/default/multiple-resources.yaml (thanks @bdragon for setting that up), which it seems to handle fine:

```
 bin/dry-run.rb docker dsp-testing/dependabot-all-updates-test --dir=/config/kubernetes/default --updater-options=kubernetes_updates
To use retry middleware with Faraday v2.0+, install `faraday-retry` gem
=> fetching dependency files
=> dumping fetched dependency files: ./dry-run/dsp-testing/dependabot-all-updates-test/config/kubernetes/default
=> parsing dependency files
=> updating 3 dependencies: ruby, nginx, busybox

=== ruby (3.1)
 => checking for updates 1/3
 => latest available version is 3.1.2
 => latest allowed version is 3.1.2

=== nginx (1.23)
 => checking for updates 2/3
 => latest available version is 1.23.1
 => latest allowed version is 1.23.1

=== busybox (1.35)
 => checking for updates 3/3
 => latest available version is 1.35.0
 => latest allowed version is 1.35.0
🌍 Total requests made: '0'
```

Given that this is behind a feature-gate, I'm inclined to merge it. @bdragon do you see any reason not to?

We'll want to be mindful when writing release notes for dependabot-core, since this is not enabled by default

bdragon (approved):

jurre (approved):

jurre:
The python spec failure was addressed in https://github.com/dependabot/dependabot-core/pull/5546 and is not related, I'm going to go ahead and merge this change, thanks @brendandburns!

PR #5356, opened 2022-07-11
Install Ruby with ruby-install

mattt:
Alternative to #5190. 
Related to https://github.com/dependabot/dependabot-core/pull/5031#issuecomment-1107498561. 

Currently, we use the [Brightbox](https://www.brightbox.com/docs/ruby/ubuntu/) PPA to install Ruby with `apt`. However, this limits our ability to update to a more recent patch release of Ruby 2.7 (or later, Ruby 3).

This PR updates this step in `Dockerfile` to instead use [ruby-install](https://github.com/postmodern/ruby-install) to install Ruby versions. I used Ruby 2.7.5 to match the version installed now. As a follow-up, we should upgrade to [2.7.6](https://www.ruby-lang.org/en/news/2022/04/12/ruby-2-7-6-released/).


deivid-rodriguez (commented):
Nice! 😍

jurre:
I think this is the right approach, thanks for taking this on, I've actually been meaning to do this for so long! I believe there's a little syntax issue with the Dockerfile that we need to patch up, but other than that looks good!

Dockerfile
```diff
@@ -65,23 +67,34 @@ RUN if ! getent group "$USER_GID"; then groupadd --gid "$USER_GID" dependabot ;
...
-ENV BUNDLE_SILENCE_ROOT_WARNING=1
+ARG RUBY_VERSION=2.7.5
+ARG RUBY_INSTALL_VERSION=0.8.3
+
+ARG RUBYGEMS_SYSTEM_VERSION=3.2.20
 # Disable the outdated rubygems installation from being loaded
 ENV DEBIAN_DISABLE_RUBYGEMS_INTEGRATION=true
+
+ARG BUNDLER_V1_VERSION=1.17.3
+ARG BUNDLER_V2_VERSION=2.3.13
```
landongrindheim, line 78: I really like this update 🥇

💡 Seeing this change makes me wonder if we could take it further. What do you think about making these `ENV` variables, then pointing the references in our code ([example](https://github.com/dependabot/dependabot-core/blob/main/bundler/helpers/v1/build#L25-L27)) to `BUNDER_Vx_VERSION`? If you're interested, I think a followup PR would be a great fit for the change 🙂

landongrindheim (approved):
Great improvement 🏆

Dockerfile
```diff
@@ -27,6 +27,8 @@ RUN apt-get update \
     file \
     zlib1g-dev \
     liblzma-dev \
+    libyaml-dev \
+    libgdbm-dev \
```
jurre, line 31:
```suggestion
    libgdbm-dev \
    bison \
```

(and then order it) seems to get things to install for me locally

Dockerfile
deivid-rodriguez: I don't fully recall the whole thing from my work on #4973, but I'm pretty sure if `BUNDLER_VERSION` is used is because RubyGems actually uses that ENV variable to activate the proper Bundler version. So if I understood what you mean, I think it would not work as expected. But if we introduced something like #4973 it wouldn't be needed anyways.

jurre (approved):

PR #5358, opened 2022-07-12
Use new pip resolver for pip-compile

jurre:
As of recent the [new(-ish) pip resolver](https://pip.pypa.io/en/latest/user_guide/#changes-to-the-pip-dependency-resolver-in-20-3-2020),
which includes many improvements, can be used by pip-compile.

This enables that option, and includes a tiny bit of cleanup.

Mostly just curious to see how CI responds to this.


python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
```diff
@@ -32,6 +32,8 @@ class PipCompileVersionResolver
           "pip._internal.exceptions.InstallationSubprocessError: Command errored out with exit status 1:"
         # See https://packaging.python.org/en/latest/tutorials/packaging-projects/#configuring-metadata
         PYTHON_PACKAGE_NAME_REGEX = /[A-Za-z0-9_\-]+/.freeze
+        CONFLICTING_DEPENDENCIES_REGEX =
+          /(Cannot install.*conflicting dependencies.)|(No matching distribution found for.*$)/.freeze
```
jurre, line 36: This feels a little brittle to be honest, going to study the error output some more and figure out if there's a better way

PR #5366, opened 2022-07-12
fix: set correct url for terraform module version download 

umglurf:
According to the docs (https://www.terraform.io/internals/module-registry-protocol)
the correct download url for a terraform module registry is
/v1/modules/:namespace/:name/:system/:version/download
The source is returned in the X-Terraform-Get header and not body.
Also since there is no body, the return code is 204 and not 200.

Tested with a private module registry and with this change 
I get a changelog for module updates.


Nishnha:
Nice spot

It looks like these HTTP sources can also return a meta element https://www.terraform.io/language/modules/sources#http-urls

We might want to account for that and return the value in `content`?

umglurf:
> Nice spot
> 
> It looks like these HTTP sources can also return a meta element https://www.terraform.io/language/modules/sources#http-urls
> 
> We might want to account for that and return the value in `content`?

Yes, good catch. It was already handled in file_parser, so moved out the get_proxied_source function to a help class, so it can be reused.

terraform/lib/dependabot/terraform/file_parser.rb
```diff
@@ -168,7 +168,7 @@ def build_terragrunt_dependency(file, details)
       # Full docs at https://www.terraform.io/docs/modules/sources.html
       def source_from(details_hash)
         raw_source = details_hash.fetch("source")
-        bare_source = get_proxied_source(raw_source)
+        bare_source = Helpers::get_proxied_source(raw_source)
```
jurre, line 171:
```suggestion
        bare_source = Helpers.get_proxied_source(raw_source)
```

Ruby 🤷 (it's idiomatic to only use `::` for constant lookup and not methods), a linter might also complain about this, in which case apologies for the noise

umglurf: Fixed, it's been a while since I've done much ruby :)

umglurf:
@Nishnha / @jurre anything else that needs to be changed or fixed, or can this be merged?

Nishnha:
Hi @umglurf, the functionality in this PR looks good, but I would like to avoid creating a catch-all Helpers file.

Maybe we can move `#get_proxied_source` into `Terraform::RegistryClient` and require the RegistryClient in the FilerParser?

umglurf:
> Hi @umglurf, the functionality in this PR looks good, but I would like to avoid creating a catch-all Helpers file.
> 
> Maybe we can move `#get_proxied_source` into `Terraform::RegistryClient` and require the RegistryClient in the FilerParser?

Hi, that sounds good, I've changed it now.

Nishnha (approved):
LGTM
Thanks!

PR #5373, opened 2022-07-13
[NPM/YARN] Cache client-side timeouts when a remote host is unreachable

brrygrdn:

This PR modifies the timeout caching strategy we used for Maven into a generic 'Dependabot::RegistryClient' and then applies it to the `npm_and_yarn` gem to ship this strategy for another ecosystem.

NPM is currently the ecosystem experiencing this problem most often, so let's address it next.

### Alternatives considered

I originally approached this as an Excon middleware, but there were a few issues with that approach in my opinion:
- I'd prefer to roll out the change gradually one ecosystem at a time to avoid missing any unexpected interactions with package managers
- I'd prefer to limit the change in Excon behaviour to code that is calling registries rather than _all_ remote calls.
  - We could do this by parameterising the SharedHelpers to incorporate the middleware for some callers but not others, but that feels somewhat brittle
- Since we are now caching state I'd prefer to do it within a concrete class, a middleware could be made to call out to a data store class but the lifecycle is slightly harder to reason about when it is part of an Excon stack


common/lib/dependabot/registry_client.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+require "dependabot/shared_helpers"
+
+# This class provides a thin wrapper around our normal usage of Excon as a simple HTTP client in order to
+# provide some minor caching functionality.
+#
+# This is not used to support full response caching currently, we just use it to ensure we detect unreachable
+# hosts and fast-fail on any subsequent requests to them to avoid excessive use of retries and connect- or
+# read-timeouts as Maven jobs tend to be sensitive to exceeding our overall 45 minute timeout.
```
jurre, line 10:
```suggestion
# read-timeouts as some jobs tend to be sensitive to exceeding our overall 45 minute timeout.
```

🤷

maven/lib/dependabot/maven/metadata_finder.rb
```diff
@@ -104,7 +105,7 @@ def source_from_anywhere_in_pom(pom)
       def dependency_pom_file
         return @dependency_pom_file unless @dependency_pom_file.nil?
 
-        response = RegistryClient.get(
+        response = Dependabot::RegistryClient.get(
```
jurre, line 108: This code could stay unchanged right? As we're already in the `Dependabot` module. I don't mind if you prefer this for clarity though

brrygrdn: Yeah, I could have left this but there is a Terraform::RegistryClient which will likely need to extend this class, so I err'd on the side of the full namespace in case we ever insert other extensions

jurre: Aye, that makes sense 👍

PR #5374, opened 2022-07-13
[Python] Cache client-side timeouts when a remote host is unreachable

brrygrdn:
See: https://github.com/dependabot/dependabot-core/pull/5373

This PR modifies the timeout caching strategy we used for Maven into a generic 'Dependabot::RegistryClient' and then applies it to the `python` gem to ship this strategy for another ecosystem.

Python is currently in the two three ecosystems experiencing this issue, but it's a little lower frequency than NPM so it might be worth shipping this first.


Nishnha (approved):
I reviewed this against https://github.com/dependabot/dependabot-core/pull/5373 and it LGTM

jeffwidman (approved):

PR #5392, opened 2022-07-18
Pip: Do not raise PathDependenciesNotReachable for missing setup.py

jurre:
Previously, Dependabot would assume that for every requirements.txt with
a self-reference (`-e .`), we should expect a `setup.py` to be present. The
self-reference is considered a path dependency that needs a setup.py.

However, nowadays it's valid to have a requirements.txt without a
setup.py. This change accounts for that by still attempting to pull in a
setup.py if it exists, but to not raise and error when it doesn't.

Closes https://github.com/dependabot/dependabot-core/pull/5324
Resolves https://github.com/dependabot/dependabot-core/issues/4483

### DIFF
--- a/python/spec/dependabot/python/file_fetcher_spec.rb
+++ b/python/spec/dependabot/python/file_fetcher_spec.rb
@@ -1037,25 +1037,44 @@ RSpec.describe Dependabot::Python::FileFetcher do
           end
         end
       end
+    end
 
-      context "that has an unfetchable path" do
-        before do
-          stub_request(:get, url + "setup.py?ref=sha").
-            with(headers: { "Authorization" => "token token" }).
-            to_return(status: 404)
-          stub_request(:get, url + "setup.cfg?ref=sha").
-            with(headers: { "Authorization" => "token token" }).
-            to_return(status: 404)
-        end
+    context "with a pyproject.toml and a requirements.txt file that does not use setup.py" do
+      let(:repo_contents) do
+        fixture("github", "contents_python_pyproject_and_requirements_without_setup_py.json")
+      end
+      before do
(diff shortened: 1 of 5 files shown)


python/spec/dependabot/python/file_fetcher_spec.rb
```diff
@@ -1037,25 +1037,44 @@
...
+            headers: { "content-type" => "application/json" }
+          )
+        stub_request(:get, url + "pyproject.toml?ref=sha").
+          with(headers: { "Authorization" => "token token" }).
+          to_return(
+            status: 200,
+            body: fixture("github", "contents_python_pyproject.json"),
+            headers: { "content-type" => "application/json" }
+          )
+        stub_request(:get, url + "setup.cfg?ref=sha").
```
rouge8, line 1061: There should probably be a test for no `setup.cfg` as well, just a `requirements.txt` with `-e .` and `pyproject.toml`

jurre: I don't think another test would add much extra coverage, I would not assume that to be able to break individually from this current test, or am I missing something?

DanielNoord (commented):
As far as I can see this should work. Thanks for all the work you did here @jurre 😄

jakecoffman (approved):

python/spec/dependabot/python/file_fetcher_spec.rb
jurre: Thanks for checking this out though, always appreciate another set of eyes! 🙇

pavera (approved):

DanielNoord:
Thanks @jurre this is now working as expected 😄

PR #5399, opened 2022-07-20
[Nuget] Cache client-side timeouts when a remote host is unreachable

brrygrdn:

This PR switches Nuget over to use the `Dependabot::RegistryClient` so it benefits from caching of unreachable hosts.


jurre (approved):
LGTM 👍

PR #5407, opened 2022-07-21
[terraform] Cache client-side timeouts when a remote host is unreachable

brrygrdn:

This PR switches Terraform over to use the `Dependabot::RegistryClient` so it benefits from caching of unreachable hosts.

~~Since this gem implements `Dependabot::Terraform::RegistryClient`, I incorporated the caching client as a base class. I have one question about the existing implementation that might require some extra changes~~ I walked that back as it just creates problems with the use of class variables and there's not a lot gained by inheritance.


terraform/lib/dependabot/terraform/file_parser.rb
```diff
@@ -270,11 +270,7 @@ def get_proxied_source(raw_source) # rubocop:disable Metrics/AbcSize
         url = raw_source.split(%r{(?<!:)//}).first + "?terraform-get=1"
         host = URI.parse(raw_source).host
 
-        response = Excon.get(
-          url,
-          idempotent: true,
-          **SharedHelpers.excon_defaults
-        )
+        response = Dependabot::RegistryClient.get(url: url)
```
brrygrdn, line 273: I found it curious that the file parser doesn't already use `Dependabot::Terraform::RegistryClient`, but at a cursory read it seems like it is always going to be making unauthenticated calls so this is likely working as intended.

@jurre am I correct in thinking we intentionally do not use the terraform client here?

jurre: It's been a minute since I've worked on this, but I do think you're right yes

landongrindheim (approved):

terraform/lib/dependabot/terraform/file_parser.rb
```diff
@@ -257,6 +257,35 @@ def version_from_ref(ref)
         ref.match(version_regex).named_captures.fetch("version")
       end
 
+      # rubocop:disable Metrics/PerceivedComplexity
+      # See https://www.terraform.io/docs/modules/sources.html#http-urls for
+      # details of how Terraform handle HTTP(S) sources for modules
+      def get_proxied_source(raw_source) # rubocop:disable Metrics/AbcSize
```
Nishnha, line 263: The Terraform::RegistryClient actually already [has a copy of #get_proxied_source](https://github.com/dependabot/dependabot-core/blob/994075e9975e399fa0a7384b4e6ece9b872ae501/terraform/lib/dependabot/terraform/registry_client.rb#L27) in `main`. 
It's being used in the FileParser on [line 171](https://github.com/dependabot/dependabot-core/pull/5407/files#diff-ae8ba34c4ac15d002de6506c383c15096471b36b1e5642f8dfd6ce1825ba06b1R171)

It was moved from the Terraform::FileParser after this PR was created by https://github.com/dependabot/dependabot-core/pull/5366/ 

So we can either move this method back into the FileParser, undoing the changes in #5366, or we can remove it from this PR

Nishnha (changes requested):
`#get_proxied_source` is being duplicated here since it was moved to the `RegistryClient` in a more recent PR

terraform/lib/dependabot/terraform/file_parser.rb
Nishnha: Addressed in [c2c188f](https://github.com/dependabot/dependabot-core/pull/5407/commits/c2c188ffdf2768d7ace7fc3f64b2263eddd94d0f)

landongrindheim (approved):
Thanks 🎉

Nishnha (approved):

PR #5408, opened 2022-07-21
Cache client-side timeouts when a remote host is unreachable for remaining ecosystems

brrygrdn:

This PR collects up the final swap over from using `Excon` directly to using `Dependabot::RegistryClient` to cache unreachable hosts and provide an interface to start adding more logging/caching behaviour in future.


landongrindheim (approved):
I've been following along with these changes. Great improvement! 🥇

PR #5413, opened 2022-07-22
Tweak bump-version script to just push the newly-created tag

brrygrdn:
Small nitpick, it makes me nervous to use `git push --tags` in the event I've accidentally created all release-like tags locally and accidentally push something to RubyGems, etc


jakecoffman (approved):
That has always made me nervous as well, thanks for this.

PR #5428, opened 2022-07-26
Add dependabot updates for Dockerfiles

jeffwidman:
I was surprised I had to manually open #5030 until I realized
that the docker ecosystem was never enabled for Dependabot.

So this fixes that.


mctofu (approved):
👍

jeffwidman:
Nice to see this is now working: https://github.com/dependabot/dependabot-core/pull/5431

PR #5432, opened 2022-07-27
Bump base image in docker push workflow to `20.04`

jeffwidman:
We've been running on `20.04` since https://github.com/dependabot/dependabot-core/pull/4394, I just didn't realize this actions workflow also pinned the version. So this bumps to our current base image.

Unfortunately, this will likely be outdated in the future again, once #5030 lands... and likely to be an ongoing issue of drift whenever we bump Ubuntu versions.

I wish there was an easy way to extract this value from the Dockerfile, but everything I come up with seems quite hacky/brittle.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,6 +1,6 @@
 name: Push docker images
 env:
-  BASE_IMAGE: "ubuntu:18.04"
+  BASE_IMAGE: "ubuntu:20.04"
   CORE_IMAGE: "dependabot/dependabot-core"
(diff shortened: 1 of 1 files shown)


jurre (approved):
We could do a similarly hacky thing that we do for bundler which is to write a little script that does all of them in one go (`bin/update-bundler.rb`), and maybe a workflow that checks that we didn't forget, but it's not great

jeffwidman:
True I didn't even think of that as an option. A workflow to verify would be great, but again, tough to keep it up to date. I guess for now just ignore it unless something pops to mind suddenly.

PR #5434, opened 2022-07-27
Move comment to relevant section of code

jeffwidman:
Previously, it was easy to lose track of this comment because it's
located quite a ways from the code. Moving it makes it less likely that
packages will listed in the wrong category.

Although I'd happily bet an ice cream cone that the comment is
partially incorrect and some of the listed packages are not just for
python.


jurre (approved):
Not entirely sure we've never added any non-python-specific packages below `make`, but seems like a better place for the comment either way

PR #5440, opened 2022-07-28
Allow to mark files as executable

NobodysNightmare:
So far all files have been assumed to be non-executable. For most dependencies this is fine, because only metadata files are being touched.

However, custom `FileUpdater` implementations may have the need to create files that are marked as executable.

### What's included?

This PR contains the necessary changes to `DependencyFile` to express the intent of marking a file as executable, as well as changes to the PR creators and updaters of GitHub and GitLab to act on those intentions.

### Limitations

The most notable limitation at this point is, that the support has not been added for all code hosting platforms. If that's required, I will go through them one-by-one and add support if possible as per docs. I will not be able to actually test the change on all related platforms though.

Also noteworthy is the fact, that for `GitLab` an extra action has to be added to determine the filemode of a file. This PR is keeping the list of actions smaller by only indicating a `chmod` action on executable files. Non-executable files receive no indication of a filemode. In case an updater wanted to _change_ the filemode of an existing file, this would not work. On the other hand it seemed like an unlikely use-case to me, so I thought it would be preferable to keep the requests shorter.
Alternatively I could implement it so that there is a `chmod` action for every created or updated file.

### Use case

We have an internal tool that helps keeping parts of our CI consistent. This includes adding/maintaining script files that are under version control. A custom `FileUpdater` for this tool allows us to update the tool version and regenerate files automatically within a pull request, however when creating a new script file this will eventually fail the CI, because the script is no executable.

The GitLab implementation has been tested on a local dependabot installation already through the help of monkey patching.

### DIFF
--- a/common/lib/dependabot/dependency_file.rb
+++ b/common/lib/dependabot/dependency_file.rb
@@ -5,7 +5,7 @@ require "pathname"
 module Dependabot
   class DependencyFile
     attr_accessor :name, :content, :directory, :type, :support_file,
-                  :symlink_target, :content_encoding, :operation
+                  :symlink_target, :content_encoding, :operation, :execute_filemode
 
     class ContentEncoding
       UTF_8 = "utf-8"
@@ -20,7 +20,8 @@ module Dependabot
 
     def initialize(name:, content:, directory: "/", type: "file",
                    support_file: false, symlink_target: nil,
-                   content_encoding: ContentEncoding::UTF_8, deleted: false, operation: Operation::UPDATE)
+                   content_encoding: ContentEncoding::UTF_8, deleted: false, operation: Operation::UPDATE,
+                   execute_filemode: false)
       @name = name
       @content = content
       @directory = clean_directory(directory)
@@ -28,6 +29,7 @@ module Dependabot
       @support_file = support_file
       @content_encoding = content_encoding
       @operation = operation
+      @execute_filemode = execute_filemode
 
       # Make deleted override the operation. Deleted is kept when operation
       # was introduced to keep compatibility with downstream dependants.
@@ -55,7 +57,8 @@ module Dependabot
         "support_file" => support_file,
         "content_encoding" => content_encoding,
         "deleted" => deleted,
-        "operation" => operation
+        "operation" => operation,
+        "execute_filemode" => execute_filemode
       }
 
       details["symlink_target"] = symlink_target if symlink_target
@@ -96,6 +99,10 @@ module Dependabot
 
     def deleted?
       deleted
+    end
+
+    def execute_filemode?
+      execute_filemode
     end
 
     def binary?
--- a/common/lib/dependabot/pull_request_creator/github.rb
+++ b/common/lib/dependabot/pull_request_creator/github.rb
@@ -193,7 +193,7 @@ module Dependabot
             {
               path: (file.symlink_target ||
                      file.path).sub(%r{^/}, ""),
-              mode: "100644",
+              mode: file.execute_filemode? ? "100755" : "100644",
               type: "blob"
             }.merge(content)
           end
--- a/common/lib/dependabot/pull_request_creator/gitlab.rb
+++ b/common/lib/dependabot/pull_request_creator/gitlab.rb
@@ -114,6 +114,14 @@ module Dependabot
           }
         end
 
+        files.select(&:execute_filemode?).each do |file|
+          actions << {
+            action: "chmod",
+            file_path: file.path,
+            execute_filemode: true
+          }
+        end
+
         gitlab_client_for_source.create_commit(
           source.repo,
           branch_name,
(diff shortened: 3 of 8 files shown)


jurre:
> However, custom FileUpdater implementations may have the need to create files that are marked as executable.

Could you share a concrete example of this maybe? There are a bunch of scenario's in which Dependabot does already execute code, and we've never really ran into this limitation, so I'm trying to better understand what this is aiming to achieve.

NobodysNightmare:
> > However, custom FileUpdater implementations may have the need to create files that are marked as executable.
> 
> Could you share a concrete example of this maybe? There are a bunch of scenario's in which Dependabot does already execute code, and we've never really ran into this limitation, so I'm trying to better understand what this is aiming to achieve.

I tried outlining this in the use case above, but I will be more detailed. Our internal tool allows us to keep boilerplate across a medium amount of applications in sync. Think of:
* Dockerfiles (they look very similar with few differences)
* entrypoints for Dockerfiles (usually few, but equal boilerplate) -> this file is executable, but usually only created once
* CI scripts for deployment etc

The last one is where we currently have this need. I'll try to expand on this a little as well: We generate `.gitlab-ci.yml` files. While GitLab offers a few ways to share CI code across different repos, they all come with certain drawbacks and we chose none of them (yet). The latest release of our tool extracts some of the code you would have found inside `.gitlab-ci.yml` into shell scripts, since that makes the generated code easier to read. However that also means updating to that new version will have dependabot create new shell scripts.

Did that explanation help?

jurre:
Thanks, yes that does help! Would it be possible to perform this step in whatever code you have running these updates? Either by extending the current base class, or by iterating over the files after they've been pulled in?

Another thing that I'm curious about, if you tell your FileFetcher to use the git clone option ([like so](https://github.com/dependabot/dependabot-core/blob/644c1434e3fef878156a763a5830bb94beea120e/go_modules/lib/dependabot/go_modules.rb#L22)) would that also solve your issue? I think it would pull in the files including whatever permissions are set on the files in the source repo?

NobodysNightmare:
> I think it would pull in the files including whatever permissions are set on the files in the source repo?

I think that would not be possible, because my files are not present in the source repo, but will be created during the update step.

Let me quickly explain how I (on a high level) expect a dependabot script to work (based on examples provided in `dependabot-script`):
1. the file fetcher gets the current state of the repo
2. the update checker figures out whether an update is possible (returning updated dependencies if needed)
3. the file updater performs the actual _change_ to files (returning `DependencyFile` objects that need a change)
4. the PR creator/updater takes the files and turns them into a pull request

My custom `FileUpdater` in step 3 wants to create new files (i.e. the final PR created by dependabot will add these as new files) that have the executable flag set. For the file updater to be able to express "this file needs to be executable", we need the extension to `DependencyFile`.

Then in step 4 the commits for the PR get created and these commits also need to specify the files as executable that are marked as such. These are my changes to `PullRequestCreator::*` and `PullRequestUpdater::*`.

> Either by extending the current base class, or by iterating over the files after they've been pulled in?

Extending the `DependencyFile` base class is working pretty fine. I can just add an `attr_accessor` and that's good enough. What is really hard to override properly is the `create_commit` method (in my case of the `GitLab` versions of the creator/updater):

As you can see the change I am performing is in the middle of the method, by adding extra actions to the action list. I _can_ do this in a overwritten or patched method (and we actually do this right now), but it is not nice, because I am forced to copy original content of these methods. This is how it looks for me right now without the changes upstreamed):

```ruby
      def create_commit
        return create_submodule_update_commit if files.count == 1 && files.first.type == "submodule"

        actions = files.map do |file|
          {
            action: file_action(file),
            file_path: file.type == "symlink" ? file.symlink_target : file.path,
            content: file.content
          }
        end

        # start of intended diff
        files.select(&:filemode_executable).each do |file|
          actions << {
            action: 'chmod',
            file_path: file.path,
            execute_filemode: true
          }
        end
        # end of intended diff

        gitlab_client_for_source.create_commit(
          source.repo,
          branch_name,
          commit_message,
          actions
        )
      end

```

As you can see most of this is not from my change, but from original source code. In case you are concerned with how to make this code easier to extend (and thus maybe avoid adding the execute-flag as proposed here), an alternative could be to extract the `actions` variable to a method like so:

```ruby
def file_actions
  files.map do |file|
    {
      action: file_action(file),
      file_path: file.type == "symlink" ? file.symlink_target : file.path,
      content: file.content
    }
  end
end
```

This would make overwriting much easier:

```ruby
#overwritten in subclass
def file_actions
  actions = super
  # add extra actions here
  actions
end
```

All that said, I would naturally prefer having the full feature integrated, but I kinda understand if you think it is too edge-casey.

jurre:
> an alternative could be to extract the actions variable to a method

I like this, that way we don't need to upstream/maintain any code that we currently don't need on our end but we do offer the flexibility to extend it in this way. That change would effectively be a no-op on our end, but make the code a bit easier to read, I'd be happy to accept it.

NobodysNightmare:
Then let's do that.

I created a separate PR for this to keep things clear (after all, the code change is a different one): https://github.com/dependabot/dependabot-core/pull/5448

Closing this PR, assuming that it will not be accepted. Thanks for your time and effort.

PR #5447, opened 2022-07-29
Bump Ruby to 3.1

jeffwidman:
This is necessary to unblock https://github.com/dependabot/dependabot-core/pull/5030, because Ubuntu
`22.04` doesn't provide an older version of `openssl`, so need a new
enough Ruby to work with the newer version of `openssl`. See https://github.com/dependabot/dependabot-core/pull/5030#issuecomment-1196371879 for
the gory details.

And besides, we want to move to Ruby 3.1 anyway...


common/dependabot-common.gemspec
```diff
@@ -35,8 +35,6 @@ Gem::Specification.new do |spec|
   spec.add_dependency "parser", ">= 2.5", "< 4.0"
   spec.add_dependency "toml-rb", ">= 1.1.2", "< 3.0"
 
-  spec.add_development_dependency "debase", "0.2.3"
-  spec.add_development_dependency "debase-ruby_core_source", "0.10.16"
```
jeffwidman, line 39: This dependency was added in https://github.com/dependabot/dependabot-core/pull/5086/

Had to remove it because `debase` doesn't currently support Ruby 3 or 3.1: https://github.com/ruby-debug/ruby-debug-ide/issues/223#issuecomment-1009609192

Ruby 3 ships with a new `debug` gem? don't have many details on it... 

Not sure if we should also remove `ruby-debug-ide` or at least bump the version pin? https://github.com/dependabot/dependabot-core/pull/5086/files#diff-e843e646a7311fb218cf05845818b31d84bc04c0742f0567395c0c289cacf72cR46

cc @pavera as he may have more context since he wrote the PR originally adding this.

pub/dependabot-pub.gemspec
```diff
@@ -24,6 +24,8 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency "dependabot-common", Dependabot::VERSION
 
+  spec.add_development_dependency "webrick", ">= 1.7"
```
jeffwidman, line 27: `webrick` was removed from the `ruby` `3.0` stdlib, so have to manually add it as a dependency.

.rubocop.yml
```diff
@@ -154,6 +154,8 @@ Style/HashExcept:
   Enabled: true
 Style/HashLikeCase:
   Enabled: false
+Style/HashSyntax:
+  EnforcedShorthandSyntax: either
```
jeffwidman, line 158: This allows but doesn't require the new shorthand syntax. See detailed notes in the relevant commit message.

jurre: Yeah this makes sense to me. I'm not quite sure how I feel about the new syntax, in js/ts I think it's fine so probably just takes a bit of getting used to. Either way I probably would avoid major syntax style changes in this upgrade, and once everything works, re-evaluate on that stuff

common/dependabot-common.gemspec
landongrindheim: > Ruby 3 ships with a new debug gem? I don't have many details on it...

It's in the [ruby org](https://github.com/ruby/debug), and we use it in the Bundler ecosystem ([Bundler 1](https://github.com/dependabot/dependabot-core/blob/main/bundler/helpers/v1/Gemfile#L7) and [Bundler 2](https://github.com/dependabot/dependabot-core/blob/main/bundler/helpers/v2/Gemfile#L7)). I think it makes sense to keep it (I see it's declared on the next line down from this diff).

pavera: I added this to add support for debugging rspec tests in vscode.  We saw issues where later versions of debase had issues, I believe with missing pre-compiled binary blobs, hence the version pin.  It's easy enough to move forward and see if things behave so I'd say bump it forward and I'll look at any resulting errors.

.rubocop.yml
jeffwidman: @jurre good thought on separating out the major syntax style changes, I added a commit temporarily doing that for some freezing style changes... that keeps this PR's diff very focused.

common/dependabot-common.gemspec
```diff
@@ -18,8 +18,8 @@ Gem::Specification.new do |spec|
   spec.require_path = "lib"
   spec.files        = []
 
-  spec.required_ruby_version = ">= 2.7.0"
-  spec.required_rubygems_version = ">= 2.7.3"
+  spec.required_ruby_version = ">= 3.1.0"
+  spec.required_rubygems_version = ">= 3.3.3"
```
jeffwidman, line 22: This is the minimum rubygems version required by ruby `3.1`.

omnibus/dependabot-omnibus.gemspec
```diff
@@ -17,7 +17,7 @@ Gem::Specification.new do |spec|
   spec.homepage     = common_gemspec.homepage
   spec.license      = common_gemspec.license
 
-  spec.required_ruby_version = ">= 2.7.0"
+  spec.required_ruby_version = ">= 3.1.0"
```
jeffwidman, line 20: I'm a little surprised we need to pin this here, I would have thought we'd inherit this pin from https://github.com/dependabot/dependabot-core/pull/5447/files#diff-e843e646a7311fb218cf05845818b31d84bc04c0742f0567395c0c289cacf72cR21

Can anyone who's more familiar with ruby gemspec inheritance comment on whether we need to keep this pin? 

cc @deivid-rodriguez @jurre

jeffwidman: Ah NVM, I see it was added to pacify Rubocop here: https://github.com/dependabot/dependabot-core/commit/37cff08adfa60a6cfa251b678b815f0c046543d1

I verified that removing the pin causes Rubocop to complain again:
```
Offenses:

dependabot-omnibus.gemspec:1:1: C: Gemspec/RequiredRubyVersion: required_ruby_version should be specified.
```

common/dependabot-common.gemspec
jeffwidman: Thanks guys, I have a few followup questions:
1. Does this `debug` gem have to be pinned now that it's part of the stdlib?
2. Should I remove the `ruby-debug-ide` gem altogether? It's unclear to me from the gem readme if it's used by the new stdlib `debug` gem or if it's only used for `debase`... and if the latter, then should be removed altogether since `debase` was removed.

I'll reach out on slack/zoom for clarification... as I had some broader surrounding context questions due to my ruby ignorance.

.rubocop.yml
```diff
@@ -210,3 +212,11 @@ Style/StringLiterals:
...
 Style/SwapValues:
   Enabled: true
+
+# TODO these were temporarily disabled during the Ruby 2.7 -> 3.1 upgrade
+# in order to keep the upgrade diff small, they will be enabled/fixed in
+# a follow-on PR.
+Style/MutableConstant:
+  Enabled: false
+Style/RedundantFreeze:
+  Enabled: false
```
jeffwidman, line 222: Opened https://github.com/dependabot/dependabot-core/pull/5468 to re-enable these checks and fix the existing violations.

common/dependabot-common.gemspec
jurre: I think we'd want to bump it in the Dockerfile in that case? https://github.com/dependabot/dependabot-core/blob/8cdbf1e37096c0ac06ab192fbc07b3f5fc5c85e7/Dockerfile#L74

common/dependabot-common.gemspec
```diff
@@ -18,8 +18,8 @@ Gem::Specification.new do |spec|
   spec.require_path = "lib"
   spec.files        = []
 
-  spec.required_ruby_version = ">= 2.7.0"
-  spec.required_rubygems_version = ">= 2.7.3"
+  spec.required_ruby_version = ">= 3.1.0"
```
jurre, line 21: Do we need it to be the _minimum_ version? It'll require all consumers of the gem to also update, and if our changes are backwards compatible (which they seem to be?) we wouldn't need to force the upgrade on them

common/dependabot-common.gemspec
jeffwidman: Yep, already bumped it there: https://github.com/dependabot/dependabot-core/pull/5447/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L74

The reason it's bumped here is because otherwise `Rubocop` complains that `spec.required_ruby_version` is not >= to [`TargetRubyVersion`](https://github.com/dependabot/dependabot-core/pull/5447/files#diff-4f894049af3375c2bd4e608f546f8d4a0eed95464efcdea850993200db9fef5cR9)

common/dependabot-common.gemspec
jeffwidman: I'm not sure if we _need_ it to be bumped, I am simply obeying the `Rubocop`: https://github.com/dependabot/dependabot-core/pull/5447#discussion_r936824160

We could disable that specific cop, but given that other folks can keep using the previous release of Dependabot, I think it's okay (🤞) if we bump this. That way if we introduce backwards breaking changes later (such as [the new shorthand syntax](https://github.com/dependabot/dependabot-core/pull/5447#discussion_r935202378)), then we don't start silently breaking code for others.

That said, IIUC, currently our internal service that consumes this lib isn't on 3.1 yet, so holding back this part of the change allows us to merge this rather than having to first update the internal consumer... but again IMO, I'd kinda rather just update the internal service since it should be fairly straightforward... from what I could tell, the `2.7`->`3.1` changes were mostly straightforward despite the major version bump.

common/dependabot-common.gemspec
jeffwidman: Just  noticed your note about the past update... thx! We'll need to figure out what the root cause was as we _have_ to bump rubygems in order to bump ruby... looks like https://github.com/dependabot/dependabot-core/pull/5044#issuecomment-1109348793 and following discussion is a good place to start.

common/dependabot-common.gemspec
deivid-rodriguez: Yes, definitely. I think either bumping both or bumping none is fine. But bumping the RuboCop config only has the risk @jeffwidman mentioned.

common/dependabot-common.gemspec
pavera: My recollection is that `ruby-debug-ide` brought in `debase` as a dependency.  `ruby-debug-ide` is the "magic" that allows vscode debugging.  Since debase is not yet Ruby 3.1, and ruby-debug-ide hasn't had any updates lately, I think we'll have to remove all of these and figure out the new debugger.

common/dependabot-common.gemspec
deivid-rodriguez: My guess is that with #4973 these issues might no longer happen after upgrading RubyGems.

The rationale is that, if I understood correctly, those issues (and the RubyGems changes that caused them) were related to the implicit Bundler activation mechanism in RubyGems. However, with #4973 Dependabot will no longer rely on that, but will activate Bundler explicitly instead (through `gem "bundler", "~> 1.17"`, for example).

I guess we need to try and see.

common/dependabot-common.gemspec
jeffwidman: Deleted `ruby-debug-ide` per ☝️.

Will sync with Landon shortly to see if we can drop the pin for the `debug` gem now that it's in the 3.1 `stdlib`.

jeffwidman:
Note to self: need to also bump the ruby version that we'll start pinning here: https://github.com/dependabot/dependabot-core/pull/5433/files#diff-d982fa1f231ea7f77f270a05e3e93c9fee61101b3612945bb372fcb3fbce6d95R17

.rubocop.yml
```diff
@@ -6,7 +6,7 @@ AllCops:
   - "*/bin/**/*"
   - "**/tmp/**/*"
   - "*/spec/fixtures/**/*"
-  TargetRubyVersion: 2.7
+  TargetRubyVersion: 3.1
```
jeffwidman, line 9: If we had a `.ruby-version` file, this wouldn't be necessary: https://docs.rubocop.org/rubocop/configuration.html#setting-the-target-ruby-version

common/dependabot-common.gemspec
```diff
@@ -35,15 +35,11 @@ Gem::Specification.new do |spec|
   spec.add_dependency "parser", ">= 2.5", "< 4.0"
   spec.add_dependency "toml-rb", ">= 1.1.2", "< 3.0"
 
-  spec.add_development_dependency "debase", "0.2.3"
-  spec.add_development_dependency "debase-ruby_core_source", "0.10.16"
-  spec.add_development_dependency "debug", ">= 1.0.0"
```
jeffwidman, line 40: This pin is removed because it's now part of stdlib as of ruby 3.1.

jeffwidman: Weirdly, the tests are now failing complaining about unable to find `"debug"`, but I verified it's using Ruby 3.1.2, so not sure why it can't find `"debug"`...

deivid-rodriguez: Hei @jeffwidman!

Stdlib (default) gems don't need to be explicitly added as dependencies, although in my opinion, it's a good practice to add them anyways. But they do need to be explicitly required.

In this case, though, `debug` is not a "default gem", but a "bundled gem". The distinction is subtle but matters here. You can find more about it at https://stdgems.org.

While in the case of default gems you can always use them because they are "part of Ruby", in the case of bundled gems, that depends on your gem setup. In our case, we use a customized `GEM_PATH` (the path to install and look for gems), so the `debug` gem bundled with Ruby by default is no longer found since it lives in the default `GEM_PATH`.

So, in conclusion, I believe we need to keep the pin.

jeffwidman: Thank you @deivid-rodriguez that's very helpful.

jeffwidman: I just put back the pin and updated the relevant commit message, so it should pass CI now🤞.

jeffwidman:
I think the [remaining CI failure](https://github.com/dependabot/dependabot-core/runs/7731180550?check_suite_focus=true) might be related to https://github.com/rubygems/rubygems/pull/5812...

jeffwidman:
A bunch more failures in the updater code... I noticed `updater/Gemfile` specifies some older library versions, so I opened https://github.com/dependabot/dependabot-core/pull/5697

jeffwidman:
It looks like [the failing test for bundler](https://github.com/dependabot/dependabot-core/actions/runs/3062786504/jobs/4944148496#step:6:359) is because `dependency_rubygems_uri` now points at an unstubbed URL: `"https://rubygems.org/api/v1/versions/bundler.json"`

Heading to bed now, but tomorrow I'll check what this URL used to point at on `main` and why/when the change.

deivid-rodriguez:
I think that one will be fixed by https://github.com/dependabot/dependabot-core/pull/5509/commits/629d1f085d1e1bca7dd8740881e84662d5517a70!

jeffwidman:
I think you're on the right track, but unfortunately just cherry-picking that commit didn't fix it... I'll have to dig a little deeper.

jeffwidman:
Some of the updater specs that are failing appear to be related to Bundler v1 missing under Ruby 3.1... 

```
(rdbg) stderr
"/usr/local/lib/ruby/site_ruby/3.1.0/rubygems/dependency.rb:313:in `to_specs': Could not find 'bundler' (~> 1.17) - did find: [bundler-2.3.22] (Gem::MissingSpecVersionError)\nChecked in 'GEM_PATH=/opt/bundler/v1/.bundle' , execute `gem env` for more information\n\tfrom /usr/local/lib/ruby/site_ruby/3.1.0/rubygems/dependency.rb:323:in `to_spec'\n\tfrom /usr/local/lib/ruby/site_ruby/3.1.0/rubygems/core_ext/kernel_gem.rb:62:in `gem'\n\tfrom /opt/bundler/v1/run.rb:3:in `<main>'\n"
```

Then `response = JSON.parse(stdout)` fails since `stdout` is empty...

common/dependabot-common.gemspec
jeffwidman: I think the root cause of the updater specs failing is this bumping of Ruby gems... more specifically as outlined in https://github.com/dependabot/dependabot-core/pull/5513:
> When upgrading RubyGems to 3.3+, RubyGems gets confused about which Bundler version to use when shelling out to native helpers.

Which is exactly the problem I observe in https://github.com/dependabot/dependabot-core/pull/5447#issuecomment-1263989537.

jeffwidman:
Rebased onto https://github.com/dependabot/dependabot-core/pull/5513 to see if that solves the "can't find correct bundler issue". If that fixes it, we'll still want to merge that PR first, then rebase this onto `main` before merging this one.

common/dependabot-common.gemspec
jeffwidman: I split the ruby gems upgrade out into a separate PR that needs to land before this one: https://github.com/dependabot/dependabot-core/pull/5823

That way it's easier for us to debug any possible issues as Rubygems vs Ruby issues.

pavera (approved):
FWIW, LGTM :)

jurre (approved):

PR #5450, opened 2022-07-29
[npm] Prefer setting `process.exitCode` to `process.exit()` with pending writes

bdragon:
Writes to `process.stdout` may happen asynchronously, during the next tick (or several) of the event loop. Calling `process.exit()` immediately after calling `process.stdout.write()` may result in lost writes to stdout.

See: https://nodejs.org/api/process.html#processexitcode

### DIFF
--- a/npm_and_yarn/helpers/run.js
+++ b/npm_and_yarn/helpers/run.js
@@ -25,6 +25,6 @@ process.stdin.on("end", () => {
     })
     .catch((error) => {
       output({ error: error.message });
-      process.exit(1);
+      process.exitCode = 1;
     });
 });


jeffwidman:
@bdragon curious why you closed this? I'm sure you have a good reason, but documenting it briefly would be helpful in case someone else starts to go down this road in the future.

bdragon:
@jeffwidman I was testing behavior around exceptions thrown by the JS native helper and I was encountering cases where the helper was not writing anything to stdout/stderr when it should have been writing a well-formed JSON object describing the error. I looked into it briefly and found this recommendation in the docs. I only closed it because I haven't had time to research the implications/test it properly; I should have left it as a Draft PR instead.

bdragon:
Superseded by #6146

PR #5451, opened 2022-07-29
Testing new ruby

jeffwidman:
doing some testing against CI for #5447

### DIFF
--- a/common/dependabot-common.gemspec
+++ b/common/dependabot-common.gemspec
@@ -35,8 +35,6 @@ Gem::Specification.new do |spec|
   spec.add_dependency "parser", ">= 2.5", "< 4.0"
   spec.add_dependency "toml-rb", ">= 1.1.2", "< 3.0"
 
-  spec.add_development_dependency "debase", "0.2.3"
-  spec.add_development_dependency "debase-ruby_core_source", "0.10.16"
   spec.add_development_dependency "debug", ">= 1.0.0"
   spec.add_development_dependency "gpgme", "~> 2.0"
   spec.add_development_dependency "rake", "~> 13"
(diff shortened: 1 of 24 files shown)


common/dependabot-common.gemspec
```diff
@@ -35,8 +35,6 @@ Gem::Specification.new do |spec|
   spec.add_dependency "parser", ">= 2.5", "< 4.0"
   spec.add_dependency "toml-rb", ">= 1.1.2", "< 3.0"
 
-  spec.add_development_dependency "debase", "0.2.3"
-  spec.add_development_dependency "debase-ruby_core_source", "0.10.16"
```
jeffwidman, line 39: Note to myself when I put together the "real" PR:
* added in https://github.com/dependabot/dependabot-core/pull/5086/files
* debase doesn't currently support Ruby 3 or 3.1: https://github.com/ruby-debug/ruby-debug-ide/issues/223#issuecomment-1009609192
* ruby 3 ships with a new `debug` gem? don't have many details on it... 
* not sure if we should also remove `ruby-debug-ide` or at least bump the version pin: https://github.com/dependabot/dependabot-core/pull/5086/files#diff-e843e646a7311fb218cf05845818b31d84bc04c0742f0567395c0c289cacf72cR46

but we shouldn't need debase, it's only for debugging and we can always add it back later when they add support...

jeffwidman:
Had to temp drop the ubuntu 22.04 image in my testing as I ran into an erlang problem, and the first thing we want to do is get ruby working, then will deal with erlang etc 22.04 issues... hopefully I don't get a chicken-and-egg where they all have to bump simultaneously...

jeffwidman:
Closing as I got this passing other than Rubocop... the remaining work to make this mergable will be handled in a less scattered manner over in #5447

Issue #5388, opened 2022-07-17
Unable to authenticate to a private python package registry


andrcuns:
**Package ecosystem**
python

**Package manager version**
pipenv

**Language version**
python 3.10

**Manifest location and content before the Dependabot update**
Pipfile: https://github.com/andrcuns/dependabot-private-python/blob/main/Pipfile

**dependabot.yml content**
https://github.com/andrcuns/dependabot-private-python/blob/main/.github/dependabot.yml

**What you expected to see, versus what you actually saw**
When using private python package registry, if registry url contains environment variable and dependabot.yml contains credentials for this registry, dependabot will fail to properly match authed url to use.

Interestingly enough I am not able to reproduce the issue with native Github dependabot integration, registry is resolved correctly, yet I'm encountering this problem when using `dependabot-core` directly.

I was able to reproduce the problem through unit tests [here](https://github.com/andrcuns/dependabot-core/runs/7373288264?check_suite_focus=true#step:5:390) by simply adding token to [credentials](https://github.com/dependabot/dependabot-core/blob/main/python/spec/dependabot/python/update_checker/index_finder_spec.rb#L293).

It seems that the issue is with [regexp](https://github.com/dependabot/dependabot-core/blob/c462624656ffecb9e2789ff45b506b51a2b6e877/python/lib/dependabot/python/update_checker/index_finder.rb#L170) construction. The problem is that the regexp contains the protocol portion of the url so it will fail to match authed url. For example: `https://username:password@private.registry.com` from registries credentials will be matched using `https://private.registry.com` regexp and thus fail to match. It seems that a simple fix would be to simply remove protocol portion in constructed regexp.

**🕹 Bonus points: Smallest manifest that reproduces the issue**
https://github.com/andrcuns/dependabot-private-python

PR #5452, opened 2022-07-30
Python private registry authentication fix

andrcuns:
This should fix: https://github.com/dependabot/dependabot-core/issues/5388


pavera (approved):
Thanks for the fix @andrcuns!

Issue #5461, opened 2022-08-01
Pin the version of parser/current to whatever Ruby version we're using


jeffwidman:
Currently, we pull in whatever the latest version of `parser/current` is... if that is ahead of our actual Ruby version, then we can get warnings, which we currently silence: https://github.com/dependabot/dependabot-core/blob/18bb480e09cd99ae37b0366b394b8c7a98fcb09e/common/spec/warning_monkey_patch.rb#L4-L9

However, there's another way to handle this--by having the parser version match whatever version of Ruby we're using.

See how the Rubocop project did it:
* Original issue: https://github.com/rubocop/rubocop/issues/9700 
* Hacky solution of silencing the warning using `$VERSOSE = nil`: https://github.com/marcandre/rubocop-ast/commit/7081f7cae0326150b648c54a99e124998ab2bffb
* Cleaner solution of loading the parser version that corresponds to the ruby version: https://github.com/rubocop/rubocop/pull/9702

jeffwidman:
Actually, looking at https://github.com/whitequark/parser/blob/492871419b298e233bb788ec1793b7af7004e23e/lib/parser/current.rb#L95-L99 the warning can still be thrown even if we pin the parser to the current ruby minor version... if the patch versions are outdated. Since the parser doesn't appear to have patch versions--those regexes will catch all patch versions under the same comparison with the latest ruby patch version.

So closing as I don't think there's anything we can do here.

PR #5462, opened 2022-08-01
Add TODO note about removing parser warning

jeffwidman:
I originally thought we were only hitting this warning due to Rubocop,
and tried removing it, but it turns out we pull in this library for
other code. I don't have time to dig deeper right now, so I filed
https://github.com/dependabot/dependabot-core/issues/5461 explaining
what I learned so far and this PR adds a TODO note in the code so the
information doesn't get lost in the issue tracker.


jeffwidman:
Closing due to https://github.com/dependabot/dependabot-core/issues/5461#issuecomment-1201594998

PR #5469, opened 2022-08-03
Allow Actions to be fetched from non-GitHub source

landongrindheim:
I don't think we realized that we lacked the ability to fetch Actions
from non-GitHub.com sources. With this change, we're now able to fetch
Actions from GitHub Enterprise Server instances.

### DIFF
--- a/github_actions/lib/dependabot/github_actions/file_parser.rb
+++ b/github_actions/lib/dependabot/github_actions/file_parser.rb
@@ -55,7 +55,7 @@ module Dependabot
       def build_github_dependency(file, string)
         details = string.match(GITHUB_REPO_REFERENCE).named_captures
         name = "#{details.fetch('owner')}/#{details.fetch('repo')}"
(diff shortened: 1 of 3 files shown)


github_actions/lib/dependabot/github_actions/file_parser.rb
```diff
@@ -55,7 +55,7 @@ def workfile_file_dependencies(file)
       def build_github_dependency(file, string)
         details = string.match(GITHUB_REPO_REFERENCE).named_captures
         name = "#{details.fetch('owner')}/#{details.fetch('repo')}"
-        url = "https://github.com/#{name}"
+        url = "https://#{source.hostname}/#{name}"
```
jurre, line 58: I was wondering if we should also grab the scheme from the source, but looks like it isn't available, and people should really use ssl so 🤷

jurre (approved):

PR #5471, opened 2022-08-03
Update ruby gems using the `--no-document` flag

jeffwidman:
The `--no-document` flag [skips installing documentation](https://guides.rubygems.org/command-reference/#installupdate-options), so is faster.

This was originally proposed by @deivid-rodriguez as part of
https://github.com/dependabot/dependabot-core/pull/5035, but was lost when https://github.com/dependabot/dependabot-core/pull/5048 reverted the version bump.

So this adds back only the `--no-document` flag, but doesn't touch the
version.


deivid-rodriguez (commented):
👍 💪

landongrindheim (approved):

PR #5382, opened 2022-07-14
Bump bundler from 2.3.13 to 2.3.18 and fix script

jurre:
[Changelog](https://github.com/rubygems/rubygems/blob/db374bc25bbe024d02787336e557d9e31efc9f6d/bundler/CHANGELOG.md#2318-july-14-2022)

PR #5479, opened 2022-08-04
Downgrade bundler to 2.3.14

jurre:
The latest versions of bundler has an issue where incomplete lockfiles that are
locked to "ruby" but not including exclusively gems NOT platform specific.

This causes an issue with existing lockfiles, generating a missing gem
error.

This is explained in more detail in: https://github.com/rubygems/rubygems/pull/5807

Since this is currently preventing Dependabot updates for our users,
especially around sorbet dependencies, and since we have no good way to
signal these failures to them with an explanation on how to resolve it,
I think it's best that we downgrade bundler to a version that does not
include this, until https://github.com/rubygems/rubygems/pull/5807 is
merged and we can bump to the latest version again.


jurre:
Seems like there are some regressions in this version? The latest version that seemed to work was 2.3.14 so going to try that

deivid-rodriguez:
Yeah, I recall some issues around Ruby requirements, that's probably why we went [from 2.3.13 to 2.3.18 directly](https://github.com/dependabot/dependabot-core/pull/5382) last time.

landongrindheim (approved):

PR #5482, opened 2022-08-04
Skip installing docs when installing Ruby

### LINKED ISSUE #5475: Check if possible to build/install Ruby w/o Rdoc support
https://github.com/dependabot/dependabot-core/runs/7646741836?check_suite_focus=true#step:3:5062

```
#7 463.0 Generating RDoc documentation
#7 463.2 Parsing sources...
#7 463.2   0% [ 1/917]  /usr/local/src/ruby-3.1.2/doc/NEWS-1.8.7
```

Notice this is step 1 of 917... so there's 917 steps to build docs, which probably no one looks at, but we build on every CI run.

We should check to see if there's a flag we can pass to skip building Rdoc when we build Ruby...

jeffwidman:
I noticed that our Ruby installs included generating docs, which takes
~30 seconds:
https://github.com/dependabot/dependabot-core/runs/7646741836?check_suite_focus=true#step:3:5996

But we don't need docs in our docker images--humans can always look them
up online.

So skip installing docs and cut 30s off our CI times.

Thanks to @deivid-rodriguez for the tip on [how to pass the arg through
`ruby-install` to the underlying `ruby`](https://github.com/dependabot/dependabot-core/issues/5475#issuecomment-1204742884).

Fix #5475

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -90,7 +90,7 @@ RUN mkdir -p /tmp/ruby-install \
  && tar -xzvf ruby-install-$RUBY_INSTALL_VERSION.tar.gz \
  && cd ruby-install-$RUBY_INSTALL_VERSION/ \
  && make \
- && ./bin/ruby-install --system ruby $RUBY_VERSION \
+ && ./bin/ruby-install --system ruby $RUBY_VERSION -- --disable-install-doc \
  && gem update --system $RUBYGEMS_SYSTEM_VERSION --no-document \
  && gem install bundler -v $BUNDLER_V1_VERSION --no-document \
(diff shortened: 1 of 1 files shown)


jurre (approved):
Yes, good call!

landongrindheim (approved):

jeffwidman:
I verified the new build doesn't include building the docs. Compare:
* With docs: https://github.com/dependabot/dependabot-core/runs/7646741836?check_suite_focus=true#step:3:5996
* Without docs: https://github.com/dependabot/dependabot-core/runs/7677091557?check_suite_focus=true#step:3:4000

PR #5483, opened 2022-08-04
Add a `.ruby-version` file

jeffwidman:
We're starting to pin the desired ruby version in multiple places:
* Rubocop
* the new `ruby/setup-ruby` action that we're adding in https://github.com/dependabot/dependabot-core/pull/5433

We used to not ship the `.ruby-version` file, with the thought that this
is a library... but this file isn't included in the gemfile spec, so
we're not tying the hands of our users. This merely sets the default
version of Ruby that we're using for development. We will keep this in
sync with the Dockerfile, so it'll target the version of Ruby we use in
production at GitHub. But since not included in the gem, others are free
to use a different Ruby version.


jeffwidman:
Test failure [is weird](https://docs.rubocop.org/rubocop/configuration.html#setting-the-target-ruby-version):
```
dependabot-common.gemspec:21:32: C: Gemspec/RequiredRubyVersion: required_ruby_version and TargetRubyVersion (2.6, which may be specified in .rubocop.yml) should be equal.
```

The weird part is that according to the Rubocop docs, it honors `.ruby-version`, so should be reporting `2.7` not `2.6`... I wonder if the problem is that `.ruby-version` is in the top-level directory and all the rubocops are in the child directories?

I wonder if it might make more sense to switch to either:
1. running Rubocop from the top-level directory (not sure this is even possible given that we have separate gemspecs for each ecosystem)
2. running the github super linter... but again, may run into the problems mentioned above of needing to configure it to run w/in the ecosystem subdirs.

deivid-rodriguez:
@jeffwidman I think the issue is that you need to make sure to copy the `.ruby-version` file into the CI image, like this

```diff
diff --git a/Dockerfile.ci b/Dockerfile.ci
index c094ee0e7..da8e22857 100644
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -7,6 +7,7 @@ ARG CODE_DIR=/home/dependabot/dependabot-core
 WORKDIR ${CODE_DIR}
 
 COPY --chown=dependabot:dependabot .rubocop.yml ${CODE_DIR}/.rubocop.yml
+COPY --chown=dependabot:dependabot .ruby-version ${CODE_DIR}/.ruby-version
 COPY --chown=dependabot:dependabot omnibus ${CODE_DIR}/omnibus
 COPY --chown=dependabot:dependabot git_submodules ${CODE_DIR}/git_submodules
 COPY --chown=dependabot:dependabot terraform ${CODE_DIR}/terraform
```

jeffwidman:
Thanks @deivid-rodriguez that worked perfectly!

jurre (approved):

PR #5484, opened 2022-08-04
Add support for multi-level wildcard paths in composer

pavera:
A customer reported that nested wildcard resolution fails in Dependabot.  Given a composer.json repository like:
```{
"repositories": [
   {
     "type": "path", 
     "url": "components/*/*", 
     "options": {"symlink": false}
   }
 ], 
 "require": {
   "monolog/monolog": "1.0.*", 
   "symfony/polyfill-mbstring": "1.0.*"
 }
}
```
We would fail to properly handle both `*` and would attempt to retrieve `components/*` from the source repository.  When this attempt returns an error we fail the update job.

This PR adds support to traverse the path and expand the directory list at each level of the path which is represented by a wildcard.  It also adds a test for this added path traversal support.

### DIFF
--- a/composer/lib/dependabot/composer/file_fetcher.rb
+++ b/composer/lib/dependabot/composer/file_fetcher.rb
@@ -103,9 +103,25 @@ module Dependabot
       end
 
       def expand_path(path)
-        repo_contents(dir: path.gsub(/\*$/, "")).
-          select { |file| file.type == "dir" }.
-          map { |f| path.gsub(/\*$/, f.name) }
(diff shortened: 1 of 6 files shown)


honeyankit (approved):

honeyankit:
Thank you for explaining the changes during our 1-1.

honeyankit:
It would be good if you attach the test cases result here as well and thank you for working on this issue!

jurre (approved):

PR #5487, opened 2022-08-06
Support transitive dependency removal

mctofu:


mctofu:
Going with https://github.com/dependabot/dependabot-core/pull/5549 instead

PR #5490, opened 2022-08-07
Bump to `go` `1.19`

jeffwidman:
Announcement: https://go.dev/blog/go1.19

Full release notes: https://go.dev/doc/go1.19

I read through both, and nothing looks relevant for us, other than
picking up possible bug fixes / perf improvements. Potentially we could
leverage the new ["soft hint" to the garbage collector](https://go.dev/doc/go1.19#runtime), but I'm not aware
that we run out of memory in our containers.


Dockerfile
```diff
@@ -189,14 +189,14 @@ USER root
 ### GO
 
 # Install Go
-ARG GOLANG_VERSION=1.18.4
+ARG GOLANG_VERSION=1.19
```
jeffwidman, line 192: For major releases, the `.0` should be omitted for package download URLs, otherwise it 404's.

landongrindheim (approved):

PR #5492, opened 2022-08-08
Workaround thread safety issue by limiting worker count

jeffwidman:
The original intent of this code was to workaround a thread safety
issue. However, according to the `poetry` maintainer, the more "future
proof" workaround is [to limit the worker count rather than disable the
new installer](https://github.com/python-poetry/poetry/issues/3010#issuecomment-1076518050).

The underlying root issue is supposedly resolved in [the upcoming `1.2`
release of `poetry`](https://github.com/python-poetry/poetry/issues/3010#issuecomment-1076518050).

But until then let's use the cleaner workaround.

### DIFF
--- a/python/helpers/build
+++ b/python/helpers/build
@@ -20,7 +20,8 @@ cp -r \
 cd "$install_dir"
 PYENV_VERSION=3.10.5 pyenv exec pip --disable-pip-version-check install --use-pep517 -r "requirements.txt"
 
-# Workaround of https://github.com/python-poetry/poetry/issues/3010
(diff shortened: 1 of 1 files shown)


jeffwidman:
Oof, the maintainer suggested this workaround, but unfortunately the setting isn't available til `1.2`, by which time the underlying root cause should be fixed... more details here: https://github.com/python-poetry/poetry/issues/3010#issuecomment-1207683271

jeffwidman:
Closing in favor of https://github.com/dependabot/dependabot-core/pull/5838

PR #5503, opened 2022-08-08
Bump Erlang to `25.0.3`

jeffwidman:
I verified this version was available here: http://binaries.erlang-solutions.com/ubuntu/dists/focal/contrib/binary-amd64/Packages

Bumping some of the erlang/elixir stuff as part of trying to resolve https://github.com/dependabot/dependabot-core/pull/5030#issuecomment-1208508201

See also: #5502


jurre (approved):

jeffwidman:
Weird, so this was passing CI no problem yesterday: https://github.com/dependabot/dependabot-core/actions/runs/2820713431

Then I merged #5502 to bump Elixir, and now CI is consistently failing.

I'm going to open a separate PR to see if holding back Elixir and bumping Erlang still passes...

jeffwidman:
As suspected, #5507 passes no problem.

Stepping back, there's already issues with our erlang PPA in #5030, so probably best to focus on finding a solution for that so I don't waste time fixing it here only to have to then switch the PPA...

jeffwidman:
The problem here was Elixir `1.13` series added a function call that was removed from Erlang OTP 25... and the `1.13` series was pre-compiled against Erlang OTP 24. See #5991 which resolved the issue by bumping both together.

PR #5505, opened 2022-08-09
github_actions ecosystem was missing for debugging

honeyankit:
github_actions ecosystem was missing for debugging in vscode

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -86,6 +86,7 @@
         "dep",
         "docker",
         "elm",
+        "github_actions",
         "go_modules",
         "gradle",
         "hex",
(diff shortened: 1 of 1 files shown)


jeffwidman (approved):
Man, I wish the JSON spec allowed trailing commas so we could stop polluting `git blame`...

.vscode/launch.json
```diff
@@ -115,7 +116,8 @@
         "npm_and_yarn",
         "nuget",
         "python",
-        "terraform"
+        "terraform",
+        "github_actions"
```
landongrindheim, line 120: This makes perfect sense 😄 What do you think of putting this ecosystem in its place alphabetically? I think it'd be a nice touch and would slim the diff down 🤏

landongrindheim (approved):

.vscode/launch.json
honeyankit: Let me do that!

PR #5506, opened 2022-08-09
Set version range for pipenv to maintain python 3.6 support

pavera:
A dependabot update attempt is failing CI due in part to pipenv removing python 3.6 support.  This PR will maintain that support and resolve any other issues with the latest pipenv version.

### DIFF
--- a/python/helpers/requirements.txt
+++ b/python/helpers/requirements.txt
@@ -2,7 +2,7 @@ pip>=21.3.1,<22.2.3  # Allow earlier versions to retain python 3.6 support
 pip-tools>=6.4.0,<6.8.1  # Allow earlier versions to retain python 3.6 support
 flake8==5.0.4
 hashin==0.17.0
-pipenv==2022.4.8
+pipenv>=2022.4.8,<=22.8.5  # Allow earlier versions to retain python 3.6 support
 pipfile==0.0.2
 poetry==1.1.14
 wheel==0.37.1
--- a/python/helpers/requirements.txt
+++ b/python/helpers/requirements.txt
@@ -2,7 +2,7 @@ pip>=21.3.1,<22.2.3  # Allow earlier versions to retain python 3.6 support
 pip-tools>=6.4.0,<6.8.1  # Allow earlier versions to retain python 3.6 support
 flake8==5.0.4
 hashin==0.17.0
-pipenv>=2022.4.8,<=22.8.5  # Allow earlier versions to retain python 3.6 support
+pipenv>=2022.4.8,<=2022.8.5  # Allow earlier versions to retain python 3.6 support
 pipfile==0.0.2
 poetry==1.1.14
 wheel==0.37.1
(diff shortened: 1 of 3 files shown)


python/helpers/requirements.txt
```diff
@@ -2,7 +2,7 @@ pip>=21.3.1,<22.2.3  # Allow earlier versions to retain python 3.6 support
 pip-tools>=6.4.0,<6.8.1  # Allow earlier versions to retain python 3.6 support
 flake8==5.0.4
 hashin==0.17.0
-pipenv==2022.4.8
+pipenv>=2022.4.8,<=2022.8.5  # Allow earlier versions to retain python 3.6 support
```
jurre, line 5: We'll probably want to tell dependabot to ignore this dependency, because I think it'll override the range here?

pavera: It didn't on the 2 previous ones (pip and pip-tools) that I added in April.  I think it handles this correctly.

pavera:
The latest versions of pipenv have changed the error output quite a bit.  I can fix the 3 remaining failing tests but with much less specificity about the error that occurred.

- The [bad git ref](https://github.com/dependabot/dependabot-core/blob/main/python/spec/dependabot/python/update_checker/pipenv_version_resolver_spec.rb#L292) output no longer contains any mention of the dependency name, only the tag we were trying to checkout.
- The 2 other failures appear to have been consolidated into a new error `ERROR: metadata generation failed`.  Again this provides no mention of the dependency name or version which is causing the failure.

Because these errors lack any information that could help the user fix the problem, I'm inclined to say we pin pipenv to `2022.4.8` and have dependabot ignore updates for this dependency going forward.

jeffwidman:
> Because these errors lack any information that could help the user fix the problem, I'm inclined to say we pin pipenv to 2022.4.8 and have dependabot ignore updates for this dependency going forward.

IIUC, the root problem is due to `python 3.6`... that was EOL'd 7 months ago, but we've chosen to still support it because Ubuntu `18.04` shipped with it, and `18.04` isn't EOL til April.

At that point, hopefully we can drop support for `3.6`, which should unblock these upgrades, right?

So maybe add a TODO comment explaining ☝️.

To be clear, this merely reserves the right for us to drop support at tha ttime, it isn't a commitment that we'll drop support... I'm sure we'll have more discussions internally where we can pull metrics to see how many users of hosted Dependabot are still using 3.6...

jurre:
Could we chat to the pipenv maintainers about improving the error messages? It would be a bummer to be stuck on an old version for a while

pavera:
3.6 support is what necessitates adding a version range for various python packages as the latest packages have dropped 3.6 support.  However, the difficulty here is that pipenv has greatly reduced the specificity of their error messages in the later releases of pipenv.  As @jurre mentioned, I was working on writing up an issue for pipenv, I'll finish that out later today.

As to adding a TODO for revisiting 3.6 support in April 2023, I agree.  Since this PR is most likely going to be abandoned, I'll open a new PR that simply adds that comment in the helper requirements.txt file.

PR #5507, opened 2022-08-09
Test bumping erlang while holding back elixir

jeffwidman:
Debug PR to figure out cause of CI failures on #5503


jeffwidman:
I suspect the 1.12 series of Elixir was compiled against OTP 23 or 24, but didn't include a call that was removed from OTP 25 so it ran fine against the newer OTP. Then the 1.13 series added that new call, so started blowing up on OTP 25.

Long term, we want to solve this by using a version of Elixir that was compiled against OTP 25, so downgrading Elixir simply so we can upgrade Erlang isn't the right way to go...

PR #4818, opened 2022-03-09
Pin nuget vcr cassette names

mctofu:
When we have VCR enabled tests without a provided description like
```
its([:version]) { is_expected.to eq(version_class.new("7.3.0")) }
```

The VCR library will name the cassette [based on the `:scoped_id`](https://github.com/vcr/vcr/blob/5c0ded824dd8b1f9d50ab5d48a53be065b94ece8/lib/vcr/test_frameworks/rspec.rb#L10-L15) of the test. That id will change if the test is moved around or other tests are added before it. Our VCR config allows recording a new cassette which means a test with a changed id is silently fetching the latest data and will eventually fail when the latest data no longer matches our test expectations.

This PR fixes occurrences of this problem in the nuget tests (which are currently causing a failure) by adding a description to the test. We can address other ecosystems & the VCR config in a later PR.

PR #5508, opened 2022-08-09
Pin MessageBuilder cassette names

mctofu:
Following up on https://github.com/dependabot/dependabot-core/pull/4818 this assigns names to the tests & vcr cassettes so the tests don't fail if they are moved to a different position.

I also found we were missing cassettes for some test cases so I've added them and I've removed an orphaned cassette.


jurre (approved):

jeffwidman (approved):
😍 thanks for cleaning this up!

jeffwidman:
Would https://github.com/dependabot/dependabot-core/pull/4822 have caught the missing cassettes sooner?

mctofu:
> Would #4822 have caught the missing cassettes sooner?

@jeffwidman Yes, the tests would fail in that mode if there were missing cassettes.

Issue #5420, opened 2022-07-25
Dependabot has stopped opening PRs - breaks with sorbet dependencies


vseguin:
**Package ecosystem**
bundler
**Package manager version**
unapplicable
**Language version**
Ruby 3.1.2
**Manifest location and content before the Dependabot update**
Relevant section in our Gemfile:
```
# Type checking
gem 'sorbet', '~> 0.5', group: :development
gem 'sorbet-runtime', '~> 0.5'
```

Resulting in the following in the Gemfile.lock:
```
    sorbet (0.5.10206)
      sorbet-static (= 0.5.10206)
    sorbet-runtime (0.5.10206)
    sorbet-static (0.5.10206-universal-darwin-14)
    sorbet-static (0.5.10206-universal-darwin-15)
    sorbet-static (0.5.10206-universal-darwin-16)
    sorbet-static (0.5.10206-universal-darwin-17)
    sorbet-static (0.5.10206-universal-darwin-18)
    sorbet-static (0.5.10206-universal-darwin-19)
    sorbet-static (0.5.10206-universal-darwin-20)
    sorbet-static (0.5.10206-universal-darwin-21)
    sorbet-static (0.5.10206-universal-darwin-22)
    sorbet-static (0.5.10206-x86_64-linux)
    sorbet-static-and-runtime (0.5.10206)
      sorbet (= 0.5.10206)
      sorbet-runtime (= 0.5.10206)
```

**dependabot.yml content**

```
version: 2
registries:
  capdesk-ruby:
    type: rubygems-server
    url: https://rubygems.pkg.github.com/capdesk
    token: ${{ secrets.DEPENDABOT_WORKFLOW_TOKEN }}
updates:
  - package-ecosystem: bundler
    directory: "/"
    schedule:
      interval: weekly
    insecure-external-code-execution: allow
    open-pull-requests-limit: 10
    labels:
      - ruby
    registries:
      - capdesk-ruby
  - package-ecosystem: npm
    directory: "/"
    schedule:
      interval: weekly
    open-pull-requests-limit: 10
    labels:
      - javascript
    ignore:
      - dependency-name: ng-table
      - dependency-name: webui-popover
```

**Updated dependency**
sorbet + sorbet_static
**What you expected to see, versus what you actually saw**
Best here is to see the error link at the bottom of this report - would expect it to work.
**Native package manager behavior**
Works well ✅ 
**Images of the diff or a link to the PR, issue, or logs**

[...]

PR #5509, opened 2022-08-10
Upgrade Bundler to 2.3.22

deivid-rodriguez:
We had to downgrade it recently due to https://github.com/dependabot/dependabot-core/issues/5420.

Now this is fixed upstream, although with the slight difference that if Dependabot follows the package manager behavior here strictly (I think it does), PRs will also remove the "ruby" platform from the `Gemfile.lock` file in these cases. I understand that this might surprise some users, but these lockfiles can no longer by bundled in newer Bundler versions otherwise, so I think it's a good thing after all.


deivid-rodriguez:
Mmm, CI is red, I'll have a look!

deivid-rodriguez:
I this ends up being another issue in Bundler, I think I'm going to need to start running dependabot tests before Bundler releases, since it's so effective at caching bugs 😅.

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
```diff
@@ -142,11 +141,6 @@
 
           let(:dependency_files) { bundler_project_dependency_files("bundler_specified") }
 
-          before do
-            stub_request(:get, rubygems_url + "versions/bundler.json").
-              to_return(status: 200, body: fixture("rubygems_responses", "versions-bundler.json"))
-          end
-
```
deivid-rodriguez, line 149: This code was meant to never actually make it to the repository, it seems just an oversight that it landed. See https://github.com/dependabot/dependabot-core/pull/3887#discussion_r651000907.

bundler/helpers/v2/lib/functions/version_resolver.rb
```diff
@@ -28,11 +33,6 @@ def version_details
       # subdependency that was removed when attempting to update it.
       return nil if dep.nil?
 
-      # If the dependency is Bundler itself then we can't trust the
-      # version that has been returned (it's the version Dependabot is
-      # running on, rather than the true latest resolvable version).
-      return nil if dep.name == "bundler"
-
```
deivid-rodriguez, line 35: I explain things in the commit message, but there's a small inconsistency in Bundler 2.3.20 created by https://github.com/rubygems/rubygems/pull/5784/commits/2777e79b8efe1bf77589496632c74464022f04f3 which makes the previous early `return "latest"` trigger before this one, which is the expected one. I'm not a fan of that inconsistency, but regardless of what upstream does, we can special case Bundler earlier to get the expected outcome.

landongrindheim (approved):
Thanks for this! And big thanks for the information-rich commit messages 🥇

jeffwidman (approved):
Man it's nice to have your depth of knowledge on this ecosystem!

deivid-rodriguez:
Unfortunately there's [another regression](https://github.com/rubygems/rubygems/issues/5830) that it's going to affect Dependabot so let's postpone the upgrade one more release... :(

deivid-rodriguez:
041415e426ef0f1a91987aa05b5b8da1ca5d23bc and f64b5cfa5b3daa76fba2db2128e26c9b742404ca could be merged straight away though.

deivid-rodriguez:
And of course, this time that I decided to try the upgrading before actually releasing, no regressions were found 😄.

deivid-rodriguez:
This one should be ready now!

Just noting that in the particular case that caused us to downgrade in the first place (https://github.com/dependabot/dependabot-core/issues/5420), Bundler will now create the update successfully BUT will include removing the "ruby" platform from the `Gemfile.lock` file with the update. This is because these lockfiles were incorrectly and were working just but chance in previous versions, so Bundler "auto-heals" them.

I understand this may cause some confusion to some users, but I think there's no much that can be done. Maybe include a note in the pull request?

landongrindheim (approved):

landongrindheim:
> > I understand this may cause some confusion to some users, but I think there's no much that can be done. Maybe include a note in the pull request?
> 
> @landongrindheim So the above is not a blocker for this? Do you think it's worth doing though?

@deivid-rodriguez I think I misread this question at first. Are you suggesting a temporary message to bundler PRs? I think that would be a nice touch, and it'd be a bonus if you have a good idea of how to implement it 🙂 

I don't know that I'd consider it a blocker if we aren't breaking builds. I'd welcome the change, especially if we can do so just when the Ruby platform is removed from `Gemfile.lock`.

deivid-rodriguez:
Yes, that was my suggestion. Add a note to the PR in case Bundler removes a platform like here.

That said, I think I have to block this PR again due to another regression: https://github.com/rubygems/rubygems/issues/5871.

Oh boy.

deivid-rodriguez:
Technically is a musl (alpine) specific issue, so it shouldn't affect us. But I want to have a better understanding of the issue before moving this forward.

deivid-rodriguez:
Normally when Bundler releases break things we get reports quite early, so it sounds that Bundler 2.3.22 is finally going to be a "quiet release" 😅. And on top of that Dependabot test suite hasn't detected any issues either.

As mentioned earlier, the only change that I know of that will affect Dependabot PS is that, when dealing with lock files including the sorbet family of gems, the new version of Bundler will remove the "ruby" platform from it, and add "x86_64-linux", creating an usual diff that may surprise users.

I suggested before to note this in the PR message when it happens. I don't know how difficult implementing that will be, but I can give it a try. Also, not sure whether it's worth blocking this PR on that, as discussed previously with @landongrindheim, but also I don't think upgrading Bundler is urgent either.

jurre:
> the only change that I know of that will affect Dependabot PRs is that, when dealing with lock files including the sorbet family of gems, the new version of Bundler will remove the "ruby" platform from it, and add "x86_64-linux"

I think that's fine, what happens when users then bundle with an older version manually? Does it get put back?

deivid-rodriguez:
That's a good question, because since dependabot currently does not respect "BUNDLED WITH", but uses its own version, users are indeed likely to run an older version locally! The answer is no, it's a one time change, and older versions should deal with it fine and stick with the new platform 👍.

jeffwidman (approved):
Maybe squash your two commits that bump bundler into one? No need for separate "bump 21 to 22" and then "bump 22 to 23" IMO...

deivid-rodriguez:
I can do that, and I'm a big fan of keeping git history clean and meaningful but I think this may be fine. It shows the real story of this PR, getting little issues on each version I tried. The advantage of leaving this as is (in addition to "I don't have to do anything" 😆) is that if some problem is found on a specific version, it can be more easily bisected and reverted.

deivid-rodriguez:
By the way, I haven't merged this because I wanted to use this as an excuse to have a look at how PRs are built and try to add a note about the potential platform removal. But not a blocker at all as discussed above many times :)

jeffwidman (approved):
I think we should move forward with this, without spending any time on adding a temp note to the PR about the platform change... in general, most users won't notice/care, and the few do may file an issue in core and it's trivial to point them to this PR for details.

This way we land this and keep moving forward, and don't wonder "is that because of an older version of Bundler or something else??"

jeffwidman (approved):

PR #5510, opened 2022-08-10
Monkey patch bundler EndpointSpecification

pavera:
We have had a number of customer escalations reporting timeouts and excessive requests to private gem servers.

This is a known performance regression in bundler documented here: https://github.com/rubygems/rubygems/issues/5385

Through discussions with @deivid-rodriguez he is working on an upstream fix, but in the meantime we can patch the EndpointSpecification class so that it no longer retrieves gemspec.rz files from private servers.

I've confirmed through manual tests against a private gem server and a repo that consumes said registry that this patch greatly improves performance (~10 minutes to run dry-run.rb down to 30-45s)

### DIFF
--- a/bundler/helpers/v2/lib/functions.rb
+++ b/bundler/helpers/v2/lib/functions.rb
@@ -143,7 +143,7 @@ module Functions
 
       Bundler.settings.set_command_option(
         cred.fetch("host"),
-        token.gsub("@", "%40F").gsub("?", "%3F")
+        token.gsub("@", "%40").gsub("?", "%3F")
       )
     end
 
--- a/bundler/helpers/v2/monkey_patches/endpoint_specification_patch.rb
+++ b/bundler/helpers/v2/monkey_patches/endpoint_specification_patch.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+require "bundler/endpoint_specification"
+
+module EndpointSpecificationPatch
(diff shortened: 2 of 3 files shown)


bundler/helpers/v2/lib/functions.rb
```diff
@@ -143,7 +143,7 @@ def self.set_bundler_flags_and_credentials(dir:, credentials:)
 
       Bundler.settings.set_command_option(
         cred.fetch("host"),
-        token.gsub("@", "%40F").gsub("?", "%3F")
+        token.gsub("@", "%40").gsub("?", "%3F")
```
deivid-rodriguez, line 146: Is this related to the fix?

deivid-rodriguez (commented):
The patch looks perfect! 👍 Just a question about a part of the commit that did not seem related.

bundler/helpers/v2/lib/functions.rb
pavera: Only in that testing private gem server access via dry-run.rb was broken because of that spurious F.

deivid-rodriguez: I see, so this bug does not affect end users at all, only `dry-run.rb`, correct?

pavera: Yes.

deivid-rodriguez (approved):

PR #5512, opened 2022-08-11
Log reasons why npm audit can't succeed

mctofu:
This will make it easier to see why an update wasn't possible by viewing the job logs. Ideally we'd surface these in update error message but until then this is better than needing to reproduce the failure with the debugger.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
@@ -88,7 +88,7 @@ module Dependabot
 
         def viable_audit_result?(audit_result, security_advisories)
           validation_result = validate_audit_result(audit_result, security_advisories)
-          return true if validation_result == :valid
+          return true if validation_result == :viable
 
           Dependabot.logger.info("VulnerabilityAuditor: audit result not viable: #{validation_result}")
           false
@@ -100,7 +100,7 @@ module Dependabot
           return :dependency_still_vulnerable if dependency_still_vulnerable?(audit_result, security_advisories)
           return :downgrades_dependencies if downgrades_dependencies?(audit_result)
 
-          :valid
+          :viable
         end
 
         def vulnerable_dependency_removed?(audit_result)
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
@@ -73,7 +73,7 @@ module Dependabot
               function: "npm:vulnerabilityAuditor",
               args: [Dir.pwd, vuln_versions]
             )
-            return fix_unavailable unless valid_audit_result?(audit_result, security_advisories)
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):
Looks good!

I only wonder about the change in terminology from `valid` to `viable` in the method name, while keeping the method a predicate that returns `true` only if the result is `:valid`.

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
```diff
@@ -86,15 +86,21 @@ def audit(dependency:, security_advisories:)
 
         attr_reader :dependency_files, :credentials
 
-        def valid_audit_result?(audit_result, security_advisories)
-          # we only need to check results that indicate a fix is available
-          return true unless audit_result["fix_available"]
+        def viable_audit_result?(audit_result, security_advisories)
```
landongrindheim, line 89: 👏 Great name! It feels honest and is understandable at a glance 😄

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
```diff
@@ -86,15 +86,21 @@ def audit(dependency:, security_advisories:)
...
-          return false if downgrades_dependencies?(audit_result)
+          Dependabot.logger.info("VulnerabilityAuditor: audit result not viable: #{validation_result}")
+          false
+        end
+
+        def validate_audit_result(audit_result, security_advisories)
+          return :fix_unavailable unless audit_result["fix_available"]
+          return :vulnerable_dependency_removed if vulnerable_dependency_removed?(audit_result)
+          return :dependency_still_vulnerable if dependency_still_vulnerable?(audit_result, security_advisories)
+          return :downgrades_dependencies if downgrades_dependencies?(audit_result)
```
landongrindheim, line 101: I really like these reasons, especially the last three! I these will be really helpful for users and maintainers 😄 

🤔 Is `:fix_unavailable` as granular as we can get? It's a big improvement, but I wonder if we could send a signal like `:currently_unpatched` or `:not_fixed`.

landongrindheim (approved):
📈 🙌

mctofu:
> I only wonder about the change in terminology from `valid` to `viable` in the method name, while keeping the method a predicate that returns `true` only if the result is `:valid`.

Good eye! I renamed that because `valid_audit_result?` returned `true` when `audit_result["fix_available"] == false`. In that case it was technically valid but later on we wouldn't do the update because there wasn't a fix available. `viable_audit_result?` returns `false` when `audit_result["fix_available"] == false` but this doesn't change the end behavior of not doing the update.

I think I'll change the result from `:valid` to `:viable` to be more consistent?

deivid-rodriguez:
Makes sense!

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
mctofu: I think there may be a few edge cases we'd need to detect and expose in the `audit_result` to give a clearer reason when there isn't a fix available. We're planning to look at improving the error message in that case so can dig into that more then.

PR #5515, opened 2022-08-11
Remove ENV no longer necessary

deivid-rodriguez:
Thankfully we moved away from Debian-packaged Ruby, so we no longer need this.


landongrindheim (commented):
:shipit: 🩸

jeffwidman (approved):

PR #5516, opened 2022-08-11
Allow diverged-commit pinned Actions to be updated

landongrindheim:
The established wisdom for versioning actions is to syn major version
tags to the most recent corresponding release/tag (eg. v2 == v2.2.3).
However, there doesn't seem to be any recommendations for versioning via
commit SHA.

This commit makes the assumption that we should look up the _version_ of
diverged commits and make the update based on the version rather than
the commit. Until now, we've been bailing when commits have diverged
(aka no longer exist as part of a branch's lineage).

### DIFF
--- a/github_actions/lib/dependabot/github_actions/update_checker.rb
+++ b/github_actions/lib/dependabot/github_actions/update_checker.rb
@@ -130,8 +130,8 @@ module Dependabot
 
         # Update to the latest tag if pinned to a commit that has diverged
         if git_commit_checker.pinned_ref_looks_like_commit_sha? &&
-          !git_commit_checker.branch_or_ref_in_release?(latest_tag[:version])
-          return dependency_source_details.merge(ref: latest_tag[:tag])
+           !git_commit_checker.branch_or_ref_in_release?(latest_tag[:version])
+          return dependency_source_details.merge(ref: latest_tag.fetch(:commit_sha))
         end
 
         # Otherwise return the original source
--- a/github_actions/lib/dependabot/github_actions/update_checker.rb
+++ b/github_actions/lib/dependabot/github_actions/update_checker.rb
@@ -121,16 +121,9 @@ module Dependabot
 
         latest_tag = git_commit_checker.local_tag_for_latest_version
 
-        # Update the git commit if updating a pinned commit
+        # Update the pinned git commit if one is available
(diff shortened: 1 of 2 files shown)


github_actions/lib/dependabot/github_actions/update_checker.rb
```diff
@@ -122,14 +119,21 @@ def updated_source
...
+
         # Update the git commit if updating a pinned commit
         if git_commit_checker.pinned_ref_looks_like_commit_sha? &&
-           (latest_tag = git_commit_checker.local_tag_for_latest_version) &&
            git_commit_checker.branch_or_ref_in_release?(latest_tag[:version]) &&
            (latest_commit = latest_tag.fetch(:commit_sha)) != current_commit
           return dependency_source_details.merge(ref: latest_commit)
         end
 
+        # Update to the latest tag if pinned to a commit that has diverged
```
mctofu, line 131: I'd still expect this to update to the commit sha of the latest version rather than switch to a version here. Maybe we just need to remove the `git_commit_checker.branch_or_ref_in_release?(latest_tag[:version])` above?

If we had v1 (sha: a) and latest version v2 (sha: b) I'd expect:
1. workflow referencing v1 = PR that updates to v2
2. workflow referencing sha:a = PR that updates to sha: b

The #2 case isn't working with diverged commits now but it should.

mctofu (approved):

github_actions/lib/dependabot/github_actions/update_checker.rb
deivid-rodriguez: In my opinion, there's no easy way of knowing what the user wants in this case. I think there's two use cases for direct sha references:

> I want to pick up a bug fix or feature not yet released, because I really need it, but once released I want to stay on stable versions.

In the above case, user will expect dependabot to bring her back to a regular version when possible.

> I love what this action is doing and I really want to be on top of the latest changes as soon as they hit the default branch.

In the above case, user will expect dependabot to keep using a commit reference and keep it up to date.

So, I guess the optimal solution would be to make it configurable, but probably too much of an edge case for that.

deivid-rodriguez: In any case, what we are doing after this PR feels better because we keep the same style in the workflow, so whatever the end user expectation is, this is likely to be less surprising than what we were doing previously 👍.

PR #5519, opened 2022-08-11
Run shellcheck on all native helper build scripts

deivid-rodriguez:
It was not being run for Bundler and Composer due to having a slightly different structure.

### DIFF
--- a/bin/lint
+++ b/bin/lint
@@ -6,6 +6,7 @@ shellcheck \
   ./bin/ci-test \
   ./bin/docker-dev-shell \
   ./bin/lint \
+  ./*/helpers/*/build \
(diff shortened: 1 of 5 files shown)


mctofu (approved):
nice catch!

jeffwidman (approved):
I was actually just chatting with @mctofu earlier today about how I'd like to setup the github superlinter to handle linting all these non-core app things like yamllint, hadolint, etc... 

I didn't realize shellcheck was already setup!

deivid-rodriguez:
I pushed a small tweak to better follow existing patterns and potentially adapt to changes in the structure of other ecosystems.

landongrindheim (approved):

PR #5520, opened 2022-08-11
Option to install gems for only target ecosystem in dev shell

mctofu:
This speeds up build time of the dev shell if you only need to work with a particular ecosystem. Installing gems for all ecosystems takes a long time so this flag allows you to only install them for a particular ecosystem. You'll be able to dry-run/test the target ecosystem but you'll need to rebuild or manually install gems to work with others.

```
TARGET_ECOSYSTEM=go_modules ./bin/docker-dev-shell --rebuild
```

My bash skills are a bit rusty so open to any tips on improving this as well!

### DIFF
--- a/Dockerfile.development
+++ b/Dockerfile.development
@@ -41,7 +41,7 @@ RUN GREEN='\033[0;32m'; NC='\033[0m'; \
   for d in `find ${CODE_DIR} -type f -mindepth 2 -maxdepth 2 \
     -not -path "${CODE_DIR}/common/Gemfile" \
     -name 'Gemfile' | xargs dirname`; do \
-    if [ -z $TARGET_ECOSYSTEM ] || [ "$d" = "${CODE_DIR}/$TARGET_ECOSYSTEM" ]; then \
+    if [ -z "$TARGET_ECOSYSTEM" ] || [ "$d" = "${CODE_DIR}/$TARGET_ECOSYSTEM" ]; then \
       echo && \
       echo "---------------------------------------------------------------------------" && \
       echo "Installing gems for ${GREEN}$(realpath --relative-to=${CODE_DIR} $d)${NC}..." && \
--- a/Dockerfile.development
+++ b/Dockerfile.development
@@ -36,15 +36,18 @@ COPY --chown=dependabot:dependabot nuget/Gemfile nuget/dependabot-nuget.gemspec 
 COPY --chown=dependabot:dependabot python/Gemfile python/dependabot-python.gemspec ${CODE_DIR}/python/
 COPY --chown=dependabot:dependabot pub/Gemfile pub/dependabot-pub.gemspec ${CODE_DIR}/pub/
 COPY --chown=dependabot:dependabot terraform/Gemfile terraform/dependabot-terraform.gemspec ${CODE_DIR}/terraform/
+ARG TARGET_ECOSYSTEM=
 RUN GREEN='\033[0;32m'; NC='\033[0m'; \
   for d in `find ${CODE_DIR} -type f -mindepth 2 -maxdepth 2 \
     -not -path "${CODE_DIR}/common/Gemfile" \
     -name 'Gemfile' | xargs dirname`; do \
-    echo && \
-    echo "---------------------------------------------------------------------------" && \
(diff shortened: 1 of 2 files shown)


Dockerfile.development
```diff
@@ -36,15 +36,18 @@ COPY --chown=dependabot:dependabot nuget/Gemfile nuget/dependabot-nuget.gemspec
 COPY --chown=dependabot:dependabot python/Gemfile python/dependabot-python.gemspec ${CODE_DIR}/python/
 COPY --chown=dependabot:dependabot pub/Gemfile pub/dependabot-pub.gemspec ${CODE_DIR}/pub/
 COPY --chown=dependabot:dependabot terraform/Gemfile terraform/dependabot-terraform.gemspec ${CODE_DIR}/terraform/
+ARG TARGET_ECOSYSTEM=
```
jeffwidman, line 39: Maybe add a code comment right above that says "If specified, will only build that ecosystem. Blank will build for all ecosystems. Speeds up any debugging that requires recompiling an ecosystem".

I realize that increase the risk of the comment going out of sync with the code, but in this case it's worth it IMO simply for discoverability if someone is skimming the source it's super easy to miss that this flag exists.

jeffwidman (approved):
It's neat and we should merge it, but one nit is that it's not really discoverable for new users of the dev shell.

Since this isn't the dry-run script, I'm not exactly sure how to make this discoverable... if we should be printing potential args at the beginning fo the script run, IDK... and potentially part of this may be replaced (at least partially) by the upcoming work of splitting out the docker images (at least if we end up going down that route) so right now it's not worth investing additional time into docs/discoverability... but something to keep in mind on whenever we add more dev options

jurre (approved):
Nice one! Maybe stick a note in the README?

deivid-rodriguez (approved):
I run into this just yesterday and it was indeed painfully slow! So thanks!! :)

Maybe a silly question, but I see that all these Gemfiles only install the dependencies in `common/`. Would it be a bad idea to share this gem location accross ecosystems? Just saying because that would make this fast by default.

deivid-rodriguez:
This is what I was thinking: https://github.com/dependabot/dependabot-core/pull/5528. Not sure if it works but makes sense to me in principle?

mctofu:
> Maybe a silly question, but I see that all these Gemfiles only install the dependencies in `common/`. Would it be a bad idea to share this gem location accross ecosystems? Just saying because that would make this fast by default.

Oh interesting, I didn't realize the gems aren't customized for any of the ecosystems! I think sharing them would make more sense although it would make it harder to add a gem to a single ecosystem in the future. That may not be a problem given it hasn't happened yet.

deivid-rodriguez:
Yeah!

It that happened, I think it might still work, just the other environments would have one unused gem available. But we could split out that single bundle anyways (partially rolling this back indeed, so with some extra work).

Anyways, I think we can give #5528 a try and see if it has any unexpected side effects!

jurre:
FYI: there are some places where we do install dependencies outside of the common gemspec, not sure if that matters in this context https://github.com/dependabot/dependabot-core/blob/10d3ba4b4cddbea14173275a07707eceb6bbd85f/git_submodules/dependabot-git_submodules.gemspec#L26

mctofu:
> FYI: there are some places where we do install dependencies outside of the common gemspec, not sure if that matters in this context

Aha, nice find! I think that means the submodule updater would be missing that gem with https://github.com/dependabot/dependabot-core/pull/5528? If that's the only case I wouldn't be opposed to moving it to common as that makes the dev containers faster to build/easier to use for everyone.

Does it make sense to install the gems for the omnibus first since that should include dependencies from all the sub-projects? ex: https://github.com/dependabot/dependabot-core/pull/5529

deivid-rodriguez:
I verified that with #5528 the submodule updater is indeed missing the `parseconfig` gem, nice catch @jurre! So I'm going to close this in favor of #5529!

deivid-rodriguez:
I actually meant to close my own PR, not this one, sorry 😳.

But I guess #5529 looks promising enough to close this one too. We can reopen if we end up not merging #5529, sounds good?

Issue #4448, opened 2021-11-23
Cleanup duplicate code by standardizing on one way - maybe eliminate native helper for `go`?


jeffwidman:
Now that https://github.com/dependabot/dependabot-core/pull/4434/ landed,
the only reason for the native `go` helpers is https://github.com/dependabot/dependabot-core/blob/9060bc4a10d5a8254e674358e72db0f23095c5e0/go_modules/lib/dependabot/go_modules/path_converter.rb#L17-L20

However, looking at that file, I noticed that right after that function there's
another variant of `git_url_for_path` that is implemented in pure ruby so doesn't
require the native `go` helper:
https://github.com/dependabot/dependabot-core/blob/9060bc4a10d5a8254e674358e72db0f23095c5e0/go_modules/lib/dependabot/go_modules/path_converter.rb#L25-L28

Weirdly it says the `dependabot` backend doesn't have access to `go`, which
certainly isn't true, otherwise the code removed by #4434 wouldn't have been
added in the first place.

So it'd be best to standardize on one of these two functions and remove the other to reduce confusion.

The tradeoff here appears to be that the pure ruby func would be simpler and allow
completely eliminating the native go helpers, but according to [the code comment is less robust](https://github.com/dependabot/dependabot-core/blob/9060bc4a10d5a8254e674358e72db0f23095c5e0/go_modules/lib/dependabot/go_modules/path_converter.rb#L59-L60). However, the `robust` comment may be a red-herring, since it is a leftover from the [`go dep` days](https://github.com/dependabot/dependabot-core/commit/59fd2071dd4f0fbd09a60fc58cf105e68cbe7d6f), before `go modules` was the "one true way".

I have no opinion on which is the better way to go here, just that having two methods doing
the same thing with misleading code comments is definitely not the right way! 😄

jeffwidman:
Also, if chosen solution is to remove the native `go` helper, then https://github.com/dependabot/vcs can be archived or even deleted (since it's an unused fork).

jurre:
Yeah, agreed. I _think_ that the native version handles more than just git repo's (mercurial/svn etc) and the ruby version does only git. I am doubtful that Dependabot currently handles non-git sources well, and I wonder if switching to the pure ruby version would break any existing functionality but it's definitely worth a short investigation

jeffwidman:
[...]

PR #5521, opened 2022-08-12
Remove the native `go` helper

jeffwidman:
There are two functions that do the same thing, one in `ruby` and one in
`go`.

TODO explain why we had them both and why safe to remove the `go` one
once I finish researching it.

So I removed the one written in `go`.

It turns out this function was the only reason we had native `go`
helpers, so I was able to completely remove the helpers infra altogether.

Fix https://github.com/dependabot/dependabot-core/issues/4448


jeffwidman:
This needs to be rebased to the new architecture, but before I do that, I want to investigate the underlying problem a bit more, as when I last left it I was starting to think we actually _should_ prefer using the native `go` helper function and instead drop the `ruby` implementation. But I need to take another look.

jeffwidman:
We went the other way and dropped the ruby implementation instead:
* https://github.com/dependabot/dependabot-core/pull/6723

PR #5524, opened 2022-08-12
Optional options arg for FileFetcher

jurre:
Originally introduced by @brendandburns in
https://github.com/dependabot/dependabot-core/pull/5348,

This allows our FileFetcher to accept an optional `options` hash, that
we can use for feature-flagging, like we do in the
FileUpdater/UpdateChecker etc.

This does not make use of the functionality yet, but it will enable us
to start passing the argument in our internal systems.

No tests are added or changed, because right now the behavior is not
altered, as long as the existing tests keep working, we're good!

### DIFF
--- a/common/lib/dependabot/file_fetchers/base.rb
+++ b/common/lib/dependabot/file_fetchers/base.rb
@@ -15,7 +15,7 @@ require "dependabot/shared_helpers"
 module Dependabot
   module FileFetchers
     class Base
-      attr_reader :source, :credentials, :repo_contents_path
+      attr_reader :source, :credentials, :repo_contents_path, :options
 
       CLIENT_NOT_FOUND_ERRORS = [
         Octokit::NotFound,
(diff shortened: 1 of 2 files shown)


common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -42,11 +42,14 @@ def self.required_files_message
...
       # by repo_contents_path and still use an API trip.
-      def initialize(source:, credentials:, repo_contents_path: nil)
+      #
+      # options supports custom feature enablement
+      def initialize(source:, credentials:, repo_contents_path: nil, options: {})
         @source = source
         @credentials = credentials
         @repo_contents_path = repo_contents_path
         @linked_paths = {}
+        @options = {}
```
deivid-rodriguez, line 52: Did you mean to actually set it or is it deliberate to keep in empty for now?

deivid-rodriguez (commented):
One small comment, but looks good!

common/lib/dependabot/file_fetchers/base.rb
jurre: No this was a typo, thanks 😄

landongrindheim (approved):

deivid-rodriguez (approved):

PR #5525, opened 2022-08-12
DRY up CI workflow and make it easier to run locally

deivid-rodriguez:
Whenever I get to a new repository, one of the first things I check is the GitHub workflows files, since that gives me a nice hint of what I need to run locally to get my PRs green.

And that's what I did in this project, but since everything is docker based, it's a bit painful to copy-paste all those docker commands spread across multiple lines.

Until I found `bin/ci-test`!

So this change makes it so that CI configuration can reuse `bin/ci-test`, making `bin/ci-test` more discoverable I believe.

Also I maintained backwards compatibility with the previous way of running the script (build + run one suite), since that's useful too and I don't want to break anybody's workflow.

Note that building images with this script does not really work on M1, but that was also the case before. The same handling of `TARGET_PLATFORM` that `bin/docker-dev-shell` does needs to be added here. But I'll leave that for another time!

### DIFF
--- a/bin/ci-test
+++ b/bin/ci-test
@@ -1,12 +1,11 @@
 #!/bin/bash -e
-#/ usage: ci-test SUITE
-#/ simulate the process followed by ~/.github/workflows/ci.yml
+#/ usage: ci-test build-core|build-core-ci|<SUITE>|run-suite <SUITE>
 
 function print_usage() {
   grep ^#/ "$0" | cut -c4-
 }
 
-function handle_args() {
+function handle_suite_args() {
   export SUITE=$1
   case $SUITE in
     bundler1 | bundler2)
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
ℹ️ If this is accepted, then the required branches need to be changed in UI configuration since job names have changed.

jeffwidman (approved):
Makes sense to me!

Given what this touches, probably best to get a second review from someone else on the team who's been around a bit longer and may have more historical context for why it was originally wired up this way.

bin/ci-test
```diff
@@ -31,27 +33,72 @@ function handle_args() {
...
-    -e "DEPENDABOT_TEST_ACCESS_TOKEN=${LOCAL_GITHUB_ACCESS_TOKEN}" \
-    -it $CI \
-    bash -c \
+
+  test -t 1 && USE_TTY="-it"
+
+  docker run $USE_TTY \
+    --env "CI=true" \
+    --env "RAISE_ON_WARNINGS=true" \
+    --env "DEPENDABOT_TEST_ACCESS_TOKEN=${LOCAL_GITHUB_ACCESS_TOKEN}" \
```
Nishnha, line 70: I think we would want to do `env.` since the variable is set as a workflow environment variable in https://github.com/dependabot/dependabot-core/pull/5525/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL69

This would probably break if you tried to run `run_suite` locally, though. Maybe we could test if the `LOCAL_GITHUB_ACCESS_TOKEN` is set first and use `env.` if it isn't?

```suggestion
    --env "DEPENDABOT_TEST_ACCESS_TOKEN=${env.LOCAL_GITHUB_ACCESS_TOKEN}" \
```


see https://docs.github.com/en/actions/learn-github-actions/environment-variables

bin/ci-test
```diff
@@ -1,17 +1,19 @@
 #!/bin/bash -e
-#/ usage: ci-test SUITE
-#/ simulate the process followed by ~/.github/workflows/ci.yml
+#/ usage: ci-test build-core|build-core-ci|<SUITE>|run-suite <SUITE>
```
Nishnha, line 2: This usage is kind of difficult to read. I think we can leave more extensive comments

```suggestion
#/ usages: 
#/  - ci-test <SUITES>
#/  - ci-test build-core
#/  - ci-test build-core-ci
#/  - ci-test run-suite <SUITE>
```

bin/ci-test
Nishnha: > Do the environment variables get passed to the workflow run commands?

> and the variables are interpolated on the runner machine that runs your workflow

Oh yep they totally are, disregard!

Nishnha (approved):
I have a minor suggestion around documentation, but overall this looks like a nice improvement. 

Glad to see that the CI for this branch is indeed using the updated workflow file from this PR:
https://github.com/dependabot/dependabot-core/runs/7805915696?check_suite_focus=true#step:3:2

deivid-rodriguez:
Thanks for having a look @Nishnha, your suggestion seems totally on point 👍!

This PR somewhat conflicts with https://github.com/dependabot/dependabot-core/pull/5536, so I guess we can wait until that PR is merged, and then decide what to do here.

PR #5528, opened 2022-08-12
Experimenting with speeding up development Dockerfile build

deivid-rodriguez:

PR #5529, opened 2022-08-13
Share omnibus gems with all projects to reduce install time

mctofu:
Similar to https://github.com/dependabot/dependabot-core/pull/5528/files but re-using the omnibus gems instead of common to make sure we have the full set.

A few issues/changes I've noticed:
- Due to the path change I need to `bundle exec rspec` in a dir rather than being able to just run `rspec`.
-  Rspec didn't work on the first try and complained of
```
Could not find gem 'rubocop (~> 1.33.0)' in locally installed gems.

The source contains the following gems matching 'rubocop':
  * rubocop-1.35.0
```
A `bundle install` resulted `Installing rubocop 1.33.0` which cleared up the issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -77,9 +77,7 @@ ARG BUNDLER_V1_VERSION=1.17.3
 ARG BUNDLER_V2_VERSION=2.3.14
 ENV BUNDLE_SILENCE_ROOT_WARNING=1
 # Allow gem installs as the dependabot user
-ENV BUNDLE_PATH=".bundle" \
-    BUNDLE_BIN=".bundle/bin"
-ENV PATH="$BUNDLE_BIN:$PATH:$BUNDLE_PATH/bin"
+ENV BUNDLE_PATH=".bundle"
 
 # Install Ruby, update RubyGems, and install Bundler
 RUN mkdir -p /tmp/ruby-install \
--- a/Dockerfile.development
+++ b/Dockerfile.development
@@ -16,9 +16,6 @@ ARG CODE_DIR=${HOME}/dependabot-core
 
 COPY --chown=dependabot:dependabot common/Gemfile common/dependabot-common.gemspec ${CODE_DIR}/common/
 COPY --chown=dependabot:dependabot common/lib/dependabot/version.rb ${CODE_DIR}/common/lib/dependabot/
-WORKDIR ${CODE_DIR}
-RUN cd common \
-  && bundle install
 
 COPY --chown=dependabot:dependabot bundler/Gemfile bundler/dependabot-bundler.gemspec ${CODE_DIR}/bundler/
 COPY --chown=dependabot:dependabot cargo/Gemfile cargo/dependabot-cargo.gemspec ${CODE_DIR}/cargo/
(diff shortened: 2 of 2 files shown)


deivid-rodriguez:
> Due to the path change I need to bundle exec rspec in a dir rather than being able to just run rspec.

Oh, l see, that's inconvenient. Let me see if this can be fixed.

deivid-rodriguez (commented):
@mctofu I think the following changes should fix the issues you were having

```diff
diff --git a/Dockerfile b/Dockerfile
index 4edb14d8c..1a6020098 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -77,7 +77,8 @@ ARG BUNDLER_V1_VERSION=1.17.3
 ARG BUNDLER_V2_VERSION=2.3.14
 ENV BUNDLE_SILENCE_ROOT_WARNING=1
 # Allow gem installs as the dependabot user
-ENV BUNDLE_PATH=".bundle"
+ENV BUNDLE_PATH=".bundle" \
+    BUNDLE_BIN=".bundle/bin"
 
 # Install Ruby, update RubyGems, and install Bundler
 RUN mkdir -p /tmp/ruby-install \
diff --git a/Dockerfile.development b/Dockerfile.development
index 5401eb6c0..d3bcb943c 100644
--- a/Dockerfile.development
+++ b/Dockerfile.development
@@ -42,6 +42,8 @@ RUN cd omnibus \
 # Make omnibus gems available to bundler in root directory
 RUN echo 'eval_gemfile File.join(File.dirname(__FILE__), "omnibus/Gemfile")' > Gemfile
 
+ENV PATH="${CODE_DIR}/omnibus/$BUNDLE_BIN:$PATH"
+
 RUN cd common \
   && bundle config path ${CODE_DIR}/omnibus/.bundle
 
```

mctofu:
> A few issues/changes I've noticed:
> 
> * Due to the path change I need to `bundle exec rspec` in a dir rather than being able to just run `rspec`.
> * Rspec didn't work on the first try and complained of
> 
> ```
> Could not find gem 'rubocop (~> 1.33.0)' in locally installed gems.
> 
> The source contains the following gems matching 'rubocop':
>   * rubocop-1.35.0
> ```
> 
> A `bundle install` resulted `Installing rubocop 1.33.0` which cleared up the issue.

These were both cleared up by @deivid-rodriguez's suggested changes. I also undid the changes we'd done to the `Dockerfile` since I don't think they were needed for this to work (removing the path changes in https://github.com/dependabot/dependabot-core/pull/5529/commits/d440c632af8bb5b33d130a22a69d2bd1211da50f).

deivid-rodriguez:
Nice! Yes, the changes in the `Dockerfile` are not needed for this work 👍.

deivid-rodriguez (approved):
This looks good to me, assuming CI failures are unrelated!

(I would probably squash commits, but I'm not sure if you usually rewrite git history or prefer not to do that)

honeyankit:
Wow this is huge! From 19m down to 1.3m. Good work! thanks

jurre (approved):
🙌

jakecoffman (approved):
I was able to do a dry-run with this, seems good. ✅

Issue #5037, opened 2022-04-25
Consider publishing arm64 architecture docker images


andrcuns:
Now that https://github.com/dependabot/dependabot-core/pull/4858 is merged, it should be possible to build and publish arm64 architecture images. It would be most useful for anyone that uses core image as base image or just for local development on M1 macs.

I can prepare a PR updating the docker workflow if you are up for it.

jeffwidman:
@andrcuns sorry for the slow response!

This is a great idea, I'd happily merge a PR updating the docker publishing workflow.

jeffwidman:
Whenever this ships it'll be the easiest way to resolve this problem:
* https://github.com/github/roadmap/issues/528

There is movement on this internally, so MacOS runners for GitHub actions _will_ be happening, it's just unclear on when. 

If you're interested in this topic, I suggest subscribing to ☝️ issue to be notified of updates.

andrcuns:
Until official support lands, in case someone needs arm images, I have set up a small project that builds them for releases: https://github.com/dependabot-gitlab/core-images

Frulfump:
> Whenever this ships it'll be the easiest way to resolve this problem:
> * https://github.com/github/roadmap/issues/528
> 
> There is movement on this internally, so MacOS runners for GitHub actions _will_ be happening, it's just unclear on when. 
> 
> If you're interested in this topic, I suggest subscribing to ☝️ issue to be notified of updates.

That has shipped as well as this https://github.com/github/roadmap/issues/960 so should be no blockers to easily add this now.

PR #5536, opened 2022-08-15
Update workflows to support multi architecture image builds

andrcuns:
Implements: https://github.com/dependabot/dependabot-core/issues/5037

This pr updates workflows to support multi-arch docker image builds.

This is done by adding https://github.com/docker/build-push-action github action to build images.

This pr implements following changes:

* Use [buildx](https://github.com/docker/buildx#building-multi-platform-images) with QEMU to build multi-arch images. Benefit of this approach is minimal modifications to Dockerfile. Downside is increased build time since emulation is not very fast.
* Build ci docker image once and pass it to test jobs as an artifact. It can potentially reduce flakiness as I observed that docker build can sometimes fail when fetching dependencies etc.

Since we can't fully test this workflow in scope of pull request, I ran a few builds in my fork. Full time of core image build takes around 2h when no cache is present: https://github.com/andrcuns/dependabot-core/runs/7818717934?check_suite_focus=true. It produces correct images for both architectures: https://github.com/andrcuns/dependabot-core/pkgs/container/dependabot-core. I have also tried to build another image using dependabot-core as base and it seems to work correctly on M1 macbook now.

Some potential follow ups to improve build speeds:

* Switch from inline cache to [github](https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#cache-backend-api) cache. This also allows to use `max` cache mode which is useful for multi-stage dockerfiles. It will also add branch scoped caches so changes in pr's would be cached properly instead of always fetching inline cache from latest published image.
* Split dockerfile in multiple stages. It should improve caching since update in some of the first instructions wouldn't invalidate whole Dockerfile cache. Though change is less trivial for some of the ecosystems since they also install system packages and it's not possible to simply copy produced executable from the separate stage. But I believe there was a closed pr which already attempted to split the core image in to multiple dockerfiles, it could be useful to use as a base

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -15,6 +15,12 @@ jobs:
     permissions:
       contents: read
       packages: write
+    strategy:
+      fail-fast: false
+      matrix:
+        arch:
+          - amd64
+          - arm64
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
@@ -40,14 +46,15 @@ jobs:
           tags: |
             type=raw,value=latest
             type=semver,pattern={{version}}
-      - name: Build and push dependabot-core image
+      - name: Build and push ${{ matrix.arch }} dependabot-core image
         uses: docker/build-push-action@v3.1.1
         with:
           push: true
-          platforms: linux/amd64,linux/arm64
+          platforms: linux/${{ matrix.arch }}
           cache-from: type=registry,ref=dependabot/dependabot-core:latest
           cache-to: type=inline
           tags: ${{ steps.tags.outputs.tags }}
+
   push-development-image:
     name: Push dependabot-core-development image to GHCR
     runs-on: ubuntu-latest
@@ -76,12 +83,12 @@ jobs:
           tags: |
             type=raw,value=latest
             type=semver,pattern={{version}}
-      - name: Build dependabot-core image
+      - name: Build and push ${{ matrix.arch }} dependabot-core-development image
         uses: docker/build-push-action@v3.1.1
         with:
           push: true
           file: Dockerfile.development
-          platforms: linux/amd64,linux/arm64
+          platforms: linux/${{ matrix.arch }}
           cache-from: type=registry,ref=ghcr.io/dependabot/dependabot-core-development:latest
           cache-to: type=inline
           tags: ${{ steps.tags.outputs.tags }}
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,14 +1,12 @@
 name: Push docker images
-env:
-  BASE_IMAGE: "ubuntu:20.04"
-  CORE_IMAGE: "dependabot/dependabot-core"
-  CORE_IMAGE_MIRROR: "ghcr.io/dependabot/dependabot-core"
+
 on:
   push:
     branches:
       - main
     tags:
       - v[0-9]+.[0-9]+.[0-9]+
+
 jobs:
   push-core-image:
     name: Push dependabot-core image to docker hub
@@ -20,35 +18,36 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
-      - name: Build dependabot-core image
-        env:
-          DOCKER_BUILDKIT: 1
-        run: |
-          docker build \
-            -t "$CORE_IMAGE:latest" \
-            --build-arg BUILDKIT_INLINE_CACHE=1 \
-            --cache-from "$BASE_IMAGE" \
-            --cache-from "$CORE_IMAGE:latest" \
-            .
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2.0.0
       - name: Log in to the Docker registry
         run: |
           echo "${{ secrets.DOCKER_PASSWORD }}" | docker login -u ${{ secrets.DOCKER_USERNAME }} --password-stdin
       - name: Log in to GHCR
         run: |
           echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
-      - name: Push latest image
-        run: |
-          docker push "$CORE_IMAGE:latest"
-          docker tag "$CORE_IMAGE:latest" "$CORE_IMAGE_MIRROR:latest"
-          docker push "$CORE_IMAGE_MIRROR:latest"
-      - name: Push tagged image
-        if: "contains(github.ref, 'refs/tags')"
-        run: |
-          VERSION="$(grep -Eo "[0-9]+\.[0-9]+\.[0-9]+" common/lib/dependabot/version.rb)"
-          docker tag "$CORE_IMAGE:latest" "$CORE_IMAGE:$VERSION"
-          docker push "$CORE_IMAGE:$VERSION"
-          docker tag "$CORE_IMAGE:latest" "$CORE_IMAGE_MIRROR:$VERSION"
-          docker push "$CORE_IMAGE_MIRROR:$VERSION"
+      - name: Docker tags
+        id: tags
+        uses: docker/metadata-action@v4
+        with:
+          images: |
+            dependabot/dependabot-core
+            ghcr.io/dependabot/dependabot-core
+          flavor: |
+            latest=false
+          tags: |
+            type=raw,value=latest
+            type=semver,pattern={{version}}
+      - name: Build and push dependabot-core image
+        uses: docker/build-push-action@v3.1.1
+        with:
+          push: true
+          platforms: linux/amd64,linux/arm64
+          cache-from: type=registry,ref=dependabot/dependabot-core:latest
+          cache-to: type=inline
+          tags: ${{ steps.tags.outputs.tags }}
   push-development-image:
     name: Push dependabot-core-development image to GHCR
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM ubuntu:20.04
 
-ARG TARGETARCH=amd64
+ARG TARGETARCH
```
andrcuns, line 3: Default value is removed because this is a special argument which is populated by buildx plugin when --platform argument is used. If it has a default value assigned, it's not overwritten properly.

.github/workflows/docker.yml
```diff
@@ -20,67 +18,70 @@ jobs:
...
-      - name: Push latest image
-        run: |
-          docker push "$DEV_IMAGE:latest"
-      - name: Push tagged image
-        if: "contains(github.ref, 'refs/tags')"
-        run: |
-          VERSION="$(grep -Eo "[0-9]+\.[0-9]+\.[0-9]+" common/lib/dependabot/version.rb)"
-          docker tag "$DEV_IMAGE:latest" "$DEV_IMAGE:$VERSION"
-          docker push "$DEV_IMAGE:$VERSION"
+      - name: Docker tags
```
andrcuns, line 68: Metadata action correctly creates tags for docker images. It will automatically assign just the `latest` tag when workflow is triggered by push to default branch and will add semver version for tag pushes.

jurre:
Thanks for digging into this! I'm concerned about introducing any changes that makes our CI even slower than it already is. Or, to put it more strongly, we can't accept any changes that make our builds slower.

I noticed the `build` step takes 23 minutes in this setup, I presume that's without being able to leverage any sort of caching? I'm not sure how quick it would be normally, but having a setup where we need to wait relatively long before we get any sort of meaningful test signals seems less than ideal. Do you have any data on how fast we should expect workflows to finish counting from a PR being opened?

I remember that we've experienced using artifact caching, but landed on rebuilding the image for every workflow, because downloading the image was still slower than doing a cached build. Maybe this has changed, but just wanted to share what I remember.

The staged builds seem promising, but it would probably still be faster to use those in conjunction without the QEMU setup?

Would it be possible to split this build out into a separate optional workflow to try and keep the required workflows as fast as possible?

andrcuns:
>I noticed the build step takes 23 minutes in this setup, I presume that's without being able to leverage any sort of caching?

Yes, the current CI build step is slower due to completely invalidated cache, essentially it builds the image from scratch (whole cache got invalidated by changing the TARGETARCH argument in Dockerfile which is at the very top). You would get roughly the same result when updating anything ruby related with existing setup. As far as CI job goes, actually nothing changed, it is using exactly the same setup as before, only instead of plain docker with manually enabled buildkit it uses the build and push action and buildx (which is essentially also a wrapper around buildkit with more options).

>I remember that we've experienced using artifact caching, but landed on rebuilding the image for every workflow, because downloading the image was still slower than doing a cached build. Maybe this has changed, but just wanted to share what I remember.

I suspected it might be so. If we check the job output, core image build itself took 14 minutes to build but it takes 6 minutes to upload to cache, since it's a fairly large image. I can simply revert the last commit and it will bring back the ci workflow to exactly how it was before, just instead of plain docker commands it uses the build and push action.

>Do you have any data on how fast we should expect workflows to finish counting from a PR being opened?

I can try and run the workflow with proposed change to github cache type, because it's the only way to populate cache without actually publishing latest core image. Right now cache is essentially locked to whatever we have in latest core image.

>The staged builds seem promising, but it would probably still be faster to use those in conjunction without the QEMU setup?

Splitting in to stages would just improve docker caching. The way Dockerfile is structured right now as single stage means that update to [rubygems](https://github.com/dependabot/dependabot-core/blob/main/Dockerfile#L74) version for example invalidates the whole docker cache since it is all one sequential stage. Separate stages would simply allow docker to understand that changes in ruby stage do not affect all other stages.

Getting rid of QEMU however is rather non trivial since it requires substantial changes to how different types of ecosystems are compiled (the ones that are compiled during the build and not just fetched as a pre-compiled binary). Here is a good read about it for more context: https://medium.com/@tonistiigi/faster-multi-platform-builds-dockerfile-cross-compilation-guide-part-1-ec087c719eaf

>Would it be possible to split this build out into a separate optional workflow to try and keep the required workflows as fast as possible?

I think the best course of action to not introduce too many changes to existing workflow could be, wdyt?

* Remove the change that passes docker image as artifact and essentially revert setup to the way it was before (I think it's fairly safe to leave the changes that introduce build-and-push action since it doesn't change actual behaviour, it just introduces more options compared to plain docker command)
* Extract arm64 build to separate workflow of job for now, so the feedback from the existing release jobs remains intact

jurre:
> wdyt?

Yeah that sounds good to me! I hope we can dedicate some time to improve this setup across the board

andrcuns:
Some build time comparisons, just for visibility:

Bundler update (invalidates cache for most of the Dockerfile) with existing setup: [12m](https://github.com/dependabot/dependabot-core/runs/7773113392?check_suite_focus=true)
This PR with completely invalidated cache: [13.5m](https://github.com/dependabot/dependabot-core/runs/7834945793?check_suite_focus=true)

Difference just due to system dependency still being cached when bundler version is being updated. So this PR shouldn't introduce any additional delay in CI feedback loop.

andrcuns:
@jurre I have reverted all the changes that changed existing behaviour in any way. The arm builds are simply started as separate jobs and CI setup should be building image in exactly the same way with the difference of using buildx github action instead of plain docker command.

There are a few CI failures but I'm fairly certain these are not related to the changes.

jurre:
Yeah they don't seem related, let me dig into those failures for a bit, these changes look good to me, thanks @andrcuns!

andrcuns:
I pushed one more small fix, I noticed that git-shim isn't using TARGETARCH argument to download the correct executable

deivid-rodriguez:
@andrcuns Is it possible to keep things easily runnable locally? We have this `bin/ci-test` script for simulating what CI does locally, and I had just refactored our CI workflows to use it at #5525, so that it never becomes out of date. We should consider what to do about it. I think it's neat to be able to easily run what the CI runs locally, any ideas?

andrcuns:
@deivid-rodriguez What that `build-and-push` action does, is basically creates a docker command, so for sure it is possible to run locally.

Essentially what it runs now is:

`docker buildx build --cache-to=type=inline --cache-from=type=registry,ref=ghcr.io/dependabot/dependabot-core --platform=linux/amd64 --load -t tag-to-push .`

I would simply update the commands that can build locally, to also use buildx plugin.

Buildx plugin comes preinstalled with docker for windows/mac out of the box. Linux users will have to install it manually.

deivid-rodriguez:
Yeah, that's what I figured. My thinking is that, if we have the commands needed refactored in a script that we can run locally, maybe the actions don't buy us anything? How about updating the commands here to use `docker buildx` instead of `docker build` and not use any custom actions. I'm pretty sure it won't be as simple, right? My main concern is that `bin/ci-test` will get out of date constantly unless we actually exercise it in CI.

andrcuns:
It's obviously not my call, but my rationale has always been CI first and everything else second.

If you would opt using plain shell scripts for the sake of sometimes running similar docker command locally, you would end up essentially reimplementing all the niceties that github action provides in plain shell scripts. It would also require either pretty complicated argument parser or duplication in workflow files to correctly push multiple tags to multiple registries plus reimplementing arguments that can be passed to the build command, like platform, cache type etc.

Some flows wouldn't even be compatible with running locally, like using github cache type, which I would suggest to explore over inline cache as it opens up possibilities for properly caching multi-stage docker builds and adds capability of having branch scoped caches as well.

In my humble opinion there is no real need to necessarily have 1:1 docker build command for local use. The setup is rather simple (for now at least) in terms of building core docker image. Simple `docker build .` will provide the same final image anyway. Some specific options used in CI flow is nothing more than performance and build time improvements.

deivid-rodriguez:
> If you would opt using plain shell scripts for the sake of sometimes running similar docker command locally, you would end up essentially reimplementing all the niceties that github action provides in plain shell scripts. It would also require either pretty complicated argument parser or duplication in workflow files to correctly push multiple tags to multiple registries plus reimplementing arguments that can be passed to the build command, like platform, cache type etc.

I think this wouldn't be too bad, but of course someone needs to do the work. I normally like using actions in CI, but in this particular case I prefer the way it's done now with plain docker commands that can be easily extracted to a script and run locally.

> Some flows wouldn't even be compatible with running locally, like using github cache type, which I would suggest to explore over inline cache as it opens up possibilities for properly caching multi-stage docker builds and adds capability of having branch scoped caches as well.

Even if not a current concern, but a potential future concern, this does not seem too bad either. The script can handle both CI and local usages. For example, at #5525 I'm only passing the `-it` flag when running locally, but not in CI.

> In my humble opinion there is no real need to necessarily have 1:1 docker build command for local use. The setup is rather simple (for now at least) in terms of building core docker image. Simple docker build . will provide the same final image anyway. Some specific options used in CI flow is nothing more than performance and build time improvements.

You certainly know that, but a newcomer to the code base might not know the real differences and it can be reassuring to make sure you're reproducing locally what the CI is doing as exactly as possible.

All in all, my take here is that the benefits this PR is bringing are superior to the ones in #5525. So I'd say, green light to the approach in this PR, and if I feel it's worth merging everything back to `bin/ci-test`, I can take on that work later. But I'm happy of course to hear the opinions of other maintainers.

pharindoko:
Hey guys, 
nice idea ...
what`s missing that this PR can be merged ?

jeffwidman:
Actually chatted internally on this, and we may postpone this until after those guys finish splitting up the docker image by ecosystem... there's enough moving parts and some questions about the emulation slowness that it seems more straightforward to let that work finish and then re-evaluate this.

That said, with the popularity of Apple's M1 laptops, I expect this will continue to be a popular request and so this will happen at some point, it just might not be next week.

Also, for the build slowness, we'll have to figure out whether we gate our PR's on this, or do it as a side-car job that runs but doesn't block merges to `main`. That's probably more likely.

Anyway, some open questions... but even though I don't yet have answers, wanted to post a status update.

andrcuns:
With the split in to separate images, the arm builds actually should be more feasible because it will be much faster overall since the Dockerfiles are much smaller and it's a lot better in terms of caching since updates in some ecosystem that was higher up in Dockerfile will not be invalidating all the steps after it.

jeffwidman:
Related:
* https://github.com/dependabot/dependabot-core/pull/6588

Also, I haven't lost track of this... there's some other stuff we're waiting for internally that means we should probably keep this floating until that lands.

andrcuns:
Maybe someone finds this useful, for now I have set up arm builds in my own fork using approach that was proposed in this pr: https://github.com/andrcuns/dependabot-core/blob/arm-images/.github/workflows/images-arm.yml

It's quite slow, especially the python updater image, so I wouldn't build arm images in pr's, probably only releases. I think the main problems come from not being able to properly cache things which comes from the Dockerfile setup.

The main issue with caching I believe is that [this line](https://github.com/dependabot/dependabot-core/blob/main/Dockerfile.updater-core#L67) for all non ecosystem specific changes will invalidate docker cache. That however basically invalidates cache for all ecosystem specific images, because they all have `FROM core image` instruction, and since that has changed, docker will have to rebuild the whole ecosystem image. Most of the ecosystem images are not compilation intensive, so it's not a huge issue except the python one. Installation of python versions with arm emulation is very slow.

The other issue is that `inline` cache is not very useful for building `core` image. Inline cache supports only `min` mode which means it can't cache all stages of multi-stage docker build.

PR #5539, opened 2022-08-15
Updating yarn-lib to 1.22.19

honeyankit:
Updating @dependabot/yarn-lib to v1.22.19 

Main issue: https://github.com/github/dependabot-updates/issues/2657


jurre:
@honeyankit probably also need to `npm install` so the lockfile is also updated?

honeyankit:
After chatting with @jurre. I will let the dependabot to create this PR while executing the dependabot version update manually.

PR #5540, opened 2022-08-15
The new pip resolver is not backwards compatible

### LINKED ISSUE #5405: Unknown option --resolver in pip-compile for python3.6 + piptools-6.4.0
**Package ecosystem**

python
**Package manager version**

**Language version**
python3.6 
pip-compile 6.4.0

**dependabot.yml content**

```
version: 2

updates:
  - package-ecosystem: "pip"
    directory: "/"
    schedule:
      interval: "weekly"
      day: "monday"
    labels:
      - "dependabot"
      - "pip dependencies"
    open-pull-requests-limit: 5
    pull-request-branch-name:
      separator: "/"
```

Using dependabot on github project fails due to pip-compile trying to use `--resolver` option which is not available in pip-compile 6.4.0, the latest vailable version for python3.6.
error log:
```
updater | ERROR <job_424387440> Error processing boto3 (Dependabot::SharedHelpers::HelperSubprocessFailed)
updater | ERROR <job_424387440> Usage: pip-compile [OPTIONS] [SRC_FILES]...
updater | <job_424387440> Try 'pip-compile -h' for help.
updater | <job_424387440> 
updater | <job_424387440> Error: No such option: --resolver
```

this commit seems to be the one adding the `--resolver` option https://github.com/dependabot/dependabot-core/commit/7969ac892231c5138d0697bc085b0bcd89292c35

any requirement file will do as long as you use python3.6 via `runtime.txt` file or other method to force python3.6 version.

pavera:
Fixes #5405

Adding error handling and a retry if the pip-compile is not new enough to have the --resolver argument.

### DIFF
--- a/python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
+++ b/python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
@@ -31,7 +31,6 @@ module Dependabot
         NATIVE_COMPILATION_ERROR =
           "pip._internal.exceptions.InstallationSubprocessError: Command errored out with exit status 1:"
         # See https://packaging.python.org/en/latest/tutorials/packaging-projects/#configuring-metadata
-        UNPINNED_SETUPTOOLS = "ModuleNotFoundError: No module named 'distutils.msvccompiler'"
         PYTHON_PACKAGE_NAME_REGEX = /[A-Za-z0-9_\-]+/.freeze
         RESOLUTION_IMPOSSIBLE_ERROR = "ResolutionImpossible"
         ERROR_REGEX = /(?<=ERROR\:\W).*$/.freeze
@@ -107,7 +106,7 @@ module Dependabot
         end
 
         def compilation_error?(error)
-          error.message.include?(NATIVE_COMPILATION_ERROR) || error.message.include?(UNPINNED_SETUPTOOLS)
+          error.message.include?(NATIVE_COMPILATION_ERROR)
         end
 
         # rubocop:disable Metrics/AbcSize
--- a/python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
+++ b/python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
@@ -216,7 +216,7 @@ module Dependabot
         end
 
         def new_resolver_supported?
-          python_version.to_s.start_with?("3.6")
+          !python_version.to_s.start_with?("3.6")
         end
 
         def pip_compile_options(filename)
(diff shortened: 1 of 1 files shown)


jurre:
Instead of having to rely on a relatively slow retry, could we somehow detect that it's not available and get it right the first time? 🤔

pavera:
Possibly, I took the python approach of "ask forgiveness not permission" initially.  I contemplate using the python version which we have at hand though that might not be specific enough for all cases, it would be best if we could use the pip-compile or pip-tools version, but I didn't see a quick/easy way to get that other than shelling out, so went this direction.  I'll make another attempt.

python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
```diff
@@ -216,10 +219,15 @@ def run_command(command, env: python_env)
           )
         end
 
+        def backwards_compat?
```
jurre, line 222: Total nit, would this name make things clearer?

```suggestion
        def new_resolver_supported?
```

python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
```diff
@@ -110,6 +109,10 @@ def compilation_error?(error)
           error.message.include?(NATIVE_COMPILATION_ERROR)
         end
 
+        def unknown_option_error?(error)
+          error.message.include?(UNKNOWN_OPTION_ERROR)
+        end
+
```
jurre, line 115: Is this code now dead?

jurre (approved):
Overall direction looks great to me, I made a readability suggestion and what I think might be an unused method

python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
pavera: Probably :)

python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
```diff
@@ -216,10 +215,15 @@ def run_command(command, env: python_env)
           )
         end
 
+        def new_resolver_supported?
+          python_version.to_s.start_with?("3.6")
```
jurre, line 219: Should this now be:

```suggestion
          !python_version.to_s.start_with?("3.6")
```

Or what about:

```suggestion
          python_version >= Python::Version.new("3.6")
```

pavera: I think we'd have to do `python_version >= Python::Version.new("3.7")`.  This is the best option I think.

jurre (approved):
⭐

PR #5542, opened 2022-08-16
Add more helpful error messaging when a vulnerable dependency cannot be upgraded

Nishnha:
This PR adds a new method, `#explain_fix_unavailable`, which adds an `explanation` key to the `fix_unavailable` object when the UpdateChecker returns an `update_not_possible` error.

This `explanation` key is also used in the ConflictingDependencies check which is printed out after a failed job run and also shows up in the Dependabot security alert view.

The UpdateChecker should always raise with an `:update_not_possible` by this point since the call from [`#latest_version_resolvable_with_full_unlock?`](https://github.com/dependabot/dependabot-core/blob/7d717862f7c441f3fafd265e516e9f5a26e1c90f/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb#L121) will not have any [`viable audit results`](https://github.com/dependabot/dependabot-core/blob/7d717862f7c441f3fafd265e516e9f5a26e1c90f/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb#L76) and thus no `fix available`

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb
@@ -105,8 +105,7 @@ module Dependabot
         )
 
         vulnerable = vulnerability_audit.select do |v|
-          v["fix_available"] == false &&
-            v["explanation"]
+          !v["fix_available"] && v["explanation"]
         end
 
         conflicts + vulnerable
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb
@@ -96,13 +96,20 @@ module Dependabot
       end
 
       def conflicting_dependencies
-        ConflictingDependencyResolver.new(
+        conflicts = ConflictingDependencyResolver.new(
           dependency_files: dependency_files,
           credentials: credentials
         ).conflicting_dependencies(
           dependency: dependency,
           target_version: lowest_security_fix_version
         )
+
+        vulnerable = vulnerability_audit.select do |v|
+          v["fix_available"] == false &&
+            v["explanation"]
+        end
+
+        conflicts + vulnerable
       end
 
       private
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
@@ -98,8 +98,8 @@ module Dependabot
             when :fix_unavailable, :dependency_still_vulnerable, :downgrades_dependencies
               "No patched version available for #{fix_unavailable['dependency_name']}"
             when :vulnerable_dependency_removed
-              "#{fix_unavailable['dependency_name']} was removed in the update. Dependabot is unable to " \
-                "open a pull request. Please upgrade manually"
+              "#{fix_unavailable['dependency_name']} was removed in the update. Dependabot is not able to " \
+                "deal with this yet, but you can still upgrade manually."
             end
 
           fix_unavailable
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
@@ -78,7 +78,8 @@ module Dependabot
 
             validation_result = validate_audit_result(audit_result, security_advisories)
             unless viable_audit_result?(validation_result)
-              return explain_fix_unavailable(validation_result, fix_unavailable)
+              fix_unavailable["explanation"] = explain_fix_unavailable(validation_result, dependency)
+              return fix_unavailable
             end
 
             audit_result
@@ -92,17 +93,14 @@ module Dependabot
 
         attr_reader :dependency_files, :credentials
 
-        def explain_fix_unavailable(validation_result, fix_unavailable)
-          fix_unavailable["explanation"] =
-            case validation_result
-            when :fix_unavailable, :dependency_still_vulnerable, :downgrades_dependencies
-              "No patched version available for #{fix_unavailable['dependency_name']}"
-            when :vulnerable_dependency_removed
-              "#{fix_unavailable['dependency_name']} was removed in the update. Dependabot is not able to " \
-                "deal with this yet, but you can still upgrade manually."
-            end
-
-          fix_unavailable
+        def explain_fix_unavailable(validation_result, dependency)
+          case validation_result
+          when :fix_unavailable, :dependency_still_vulnerable, :downgrades_dependencies
+            "No patched version available for #{dependency.name}"
+          when :vulnerable_dependency_removed
+            "#{dependency.name} was removed in the update. Dependabot is not able to " \
+              "deal with this yet, but you can still upgrade manually."
+          end
         end
 
         def viable_audit_result?(validation_result)
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
@@ -41,6 +41,7 @@ module Dependabot
         #       dependency on the blocking dependency
         #   * :top_level_ancestors [Array<String>] the names of all top-level dependencies with a transitive
         #     dependency on the dependency
+        #   * :explanation [String] an explanation for why the project failed the vulnerability auditor run
         def audit(dependency:, security_advisories:)
           fix_unavailable = {
             "dependency_name" => dependency.name,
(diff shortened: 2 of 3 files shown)


npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
```diff
@@ -90,6 +90,13 @@ def viable_audit_result?(audit_result, security_advisories)
           validation_result = validate_audit_result(audit_result, security_advisories)
           return true if validation_result == :viable
 
+          case validation_result
```
landongrindheim, line 93: Could this be placed after the logging on line 100 so that we can get/share the insight? As it is, I'd expect execution to stop with `raise`

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
```diff
@@ -90,6 +90,13 @@ def viable_audit_result?(audit_result, security_advisories)
...
+          when :fix_unavailable, :dependency_still_vulnerable, :downgrades_dependencies
+            raise "No patched version available"
+          when :vulnerable_dependency_removed
+            raise "Unable to upgrade vulnerable dependency. Upgrade manually"
+          end
```
bdragon, line 98: It feels off to raise from a predicate method (yes/no question) like this. Perhaps a minor refactoring could help?

Nishnha:
🙃 
I'm getting an error from the "behaves like an update checker" test:
https://github.com/dependabot/dependabot-core/runs/7982211556?check_suite_focus=true#step:5:267

I'm pretty sure the `UpdateChecker#vulnerability_audit` method this PR introduces will need to be `public` to be accessed from the Updater (we include Dependabot Core as a gem there, so we should only access the public interface)

For `#conflicting_dependencies`, we added as a method to the UpdateChecker base class even though it's only used in the `bundler` and `npm_and_yarn` ecosystems
https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/update_checkers/base.rb#L97-L105

I think the best course of action for the `#vulnerability_audit` method is to also add it to the UpdateChecker base class and note that it's used for npm transitive dependency updates.
 
Does anyone have other suggestions?

mctofu:
> 🙃 I'm getting an error from the "behaves like an update checker" test: https://github.com/dependabot/dependabot-core/runs/7982211556?check_suite_focus=true#step:5:267
> 
> I'm pretty sure the `UpdateChecker#vulnerability_audit` method this PR introduces will need to be `public` to be accessed from the Updater (we include Dependabot Core as a gem there, so we should only access the public interface)
> 
> For `#conflicting_dependencies`, we added as a method to the UpdateChecker base class even though it's only used in the `bundler` and `npm_and_yarn` ecosystems https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/update_checkers/base.rb#L97-L105
> 
> I think the best course of action for the `#vulnerability_audit` method is to also add it to the UpdateChecker base class and note that it's used for npm transitive dependency updates.
> 
> Does anyone have other suggestions?

Can you represent the error message in the `#conflicting_dependencies` result? `#conflicting_dependencies` is fairly generic and could be implemented by more ecosystems if we had the time to do so. `#vulnerability_audit` seems much more specific to our npm implementation so it doesn't seem as likely we'd expose that for any other ecosystem.

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
```diff
@@ -73,21 +75,40 @@ def audit(dependency:, security_advisories:)
...
         private
 
         attr_reader :dependency_files, :credentials
 
-        def viable_audit_result?(audit_result, security_advisories)
-          validation_result = validate_audit_result(audit_result, security_advisories)
+        def explain_fix_unavailable(validation_result, fix_unavailable)
+          fix_unavailable["explanation"] =
+            case validation_result
+            when :no_lockfile
```
mctofu, line 99: I think it's unlikely we'd get here without a lockfile (no transitive deps to alert on in that case). We could make this simpler and just worry about the 'No patched version available" response since the removed condition is going away soon 🤞

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
```diff
@@ -87,8 +92,20 @@ def audit(dependency:, security_advisories:)
...
-        def viable_audit_result?(audit_result, security_advisories)
-          validation_result = validate_audit_result(audit_result, security_advisories)
+        def explain_fix_unavailable(validation_result, fix_unavailable)
+          fix_unavailable["explanation"] =
+            case validation_result
+            when :fix_unavailable, :dependency_still_vulnerable, :downgrades_dependencies
+              "No patched version available for #{fix_unavailable['dependency_name']}"
+            when :vulnerable_dependency_removed
+              "#{fix_unavailable['dependency_name']} was removed in the update. Dependabot is unable to " \
+                "open a pull request. Please upgrade manually"
```
deivid-rodriguez, line 102: If I understood correctly, this case is not currently supported, but we do plan to eventually support it? If that's the case, should we make it more clear to the user? I'm thinking something along the lines of 

> Dependabot is not currently able to deal with this situation, but you can still upgrade manually.

deivid-rodriguez (approved):
I'm not really well versed into how npm updates work, but this looks good to me!

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
Nishnha, line 102: That's correct and eventually we will remove this case.

Thanks for the suggestion!

```suggestion
              "#{fix_unavailable['dependency_name']} was removed in the update. Dependabot is not able to " \
                "deal with this yet, but you can still upgrade manually."
```

Nishnha: That's correct and eventually we will remove this case.

deivid-rodriguez: That's even better :)

npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker/vulnerability_auditor_spec.rb
```diff
@@ -110,7 +110,11 @@
 
           expect(Dependabot.logger).to receive(:info).with(/audit result not viable: vulnerable_dependency_removed/i)
           expect(subject.audit(dependency: dependency, security_advisories: security_advisories)).
-            to include("fix_available" => false)
+            to include(
+              "fix_available" => false,
+              "explanation" => "#{dependency.name} was removed in the update. "\
+                "Dependabot is unable to open a pull request. Please upgrade manually"
```
Nishnha, line 116:
```suggestion
                "Dependabot is not able to deal with this yet, but you can still upgrade manually."
```

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb
```diff
@@ -96,13 +96,20 @@ def requirements_update_strategy
...
+        conflicts = ConflictingDependencyResolver.new(
           dependency_files: dependency_files,
           credentials: credentials
         ).conflicting_dependencies(
           dependency: dependency,
           target_version: lowest_security_fix_version
         )
+
+        vulnerable = vulnerability_audit.select do |v|
+          v["fix_available"] == false &&
```
bdragon, line 108: I'd make the `fix_available` check a truthy test, too, just to be safe in case the key is missing somehow

```ruby
!v["fix_available"] && v["explanation"]
```

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
```diff
@@ -74,7 +75,11 @@ def audit(dependency:, security_advisories:)
               function: "npm:vulnerabilityAuditor",
               args: [Dir.pwd, vuln_versions]
             )
-            return fix_unavailable unless viable_audit_result?(audit_result, security_advisories)
+
+            validation_result = validate_audit_result(audit_result, security_advisories)
+            unless viable_audit_result?(validation_result)
+              return explain_fix_unavailable(validation_result, fix_unavailable)
```
bdragon, line 81: Subjective/non-blocking: if `explain_fix_unavailable` accepted `validation_result` as its only argument and simply returned the explanation as a string, it wouldn't need to know/restrict how the explanation was going to be used. This is just a helper method, so no big deal.

bdragon (approved):

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/vulnerability_auditor.rb
Nishnha: Good suggestion. I went with that approach in https://github.com/dependabot/dependabot-core/pull/5542/commits/620491327d8a6e9bf9624ffec771108d53e16ee3

Nishnha:
Reverting this in https://github.com/dependabot/dependabot-core/pull/5613 because of a TypeError. Will fix, add tests, and re-deploy.

PR #5544, opened 2022-08-16
fix Python tests

jakecoffman:
We test using numpy and that was recently broken https://github.com/pypa/setuptools/pull/3505

This solution is to pin setuptools to before that PR was released.


jurre (approved):

bdragon:
Should we add an issue to remind us to un-skip this test once the issue is fixed?

jakecoffman:
Yes, I'm also looking at a potential fix rather than skipping the test: https://github.com/dependabot/dependabot-core/pull/5545

jakecoffman:
Skipping is not the right answer, probably going with #5546

PR #5545, opened 2022-08-16
disable build isolation to allow custom setuptools versions

jakecoffman:


pavera:
@jurre let me know https://github.com/dependabot/dependabot-core/pull/5540#issuecomment-1215486261 that this retry is relatively expensive, so I would think we wouldn't want to always retry if it is an error we know we can't recover from

jurre:
Yeah, I guess it's better than a broken update, but if we can avoid it that would probably be preferable

jakecoffman:
#5546 makes the most sense at this point

PR #5547, opened 2022-08-16
Work around the bad setuptools release

pavera:
Setuptools removed some modules https://github.com/pypa/setuptools/pull/3505 which caused a breaking API change which other projects were not expecting, esp numpy.

Due to this, dependabot fails to create PRs if the project is using pip-compile and depends on `numpy<1.3.2`.  If we see an uptick in user complaints, this PR should alleviate the issue.


pavera:
setuptools rolled back the breaking change.

Issue #1158, opened 2019-05-19
Go module update indirect dependencies


saschagrunert:
I would take dependencies with `// indirect` into account when checking for updates, WDYT?

wwuck:
Any update on this?  It would be a useful feature for dependabot when there are security vulnerabilities in the indirect dependencies.

mctofu:
Support for this was added in https://github.com/dependabot/dependabot-core/pull/6361.

For indirect dependencies in your `go.mod` file you'll now get security updates. You can also enable version updates for them by [configuring](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#allow) your `dependabot.yml` to allow all dependency types (default is direct only).

PR #5555, opened 2022-08-18
WIP: Surface indirect deps from go.mod for security updates

mctofu:
This will allow us to propose updates to vulnerable transitive dependencies.

When the `go.mod` is in a recent format (go 1.17+) we can expect that any transitive dependencies that are used by the project are included in the `go.mod` but labeled as `// indirect`. By surfacing these as dependencies during a security update we can update them like any other dependency.

When the `go.mod` is in an older format, the vulnerable transitive dependency might only be in the `go.sum` file and we won't propose an update. This should become less common over time but we could revisit this in the future if we find that adoption of the newer format isn't widespread.

An alternative to updating the transitive dependency directly would be to look for an update to the parent dependency which requires a fixed version of the transitive dependency. I don't think that's something we currently want to maintain but may adopt in the future if the go tooling provided support for it.

Fix https://github.com/dependabot/dependabot-core/issues/1158

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_parser.rb
+++ b/go_modules/lib/dependabot/go_modules/file_parser.rb
@@ -53,7 +53,7 @@ module Dependabot
         Dependency.new(
           name: details["Path"],
           version: version,
-          requirements: details["Indirect"] || dependency_is_replaced(details) ? [] : reqs,
+          requirements: dependency_is_replaced(details) ? [] : reqs,
           package_manager: "go_modules"
         )
       end
@@ -152,7 +152,7 @@ module Dependabot
       end
 
       def skip_dependency?(dep)
-        return true if dep["Indirect"]
+        return true if dep["Indirect"] && !security_update?
 
         begin
           path_uri = URI.parse("https://#{dep['Path']}")
@@ -160,6 +160,11 @@ module Dependabot
         rescue URI::InvalidURIError
           false
         end
+      end
+
+      def security_update?
+        # TODO: Implement this
+        return true
       end
 
       def dependency_is_replaced(details)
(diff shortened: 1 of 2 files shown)


go_modules/lib/dependabot/go_modules/file_parser.rb
```diff
@@ -53,7 +53,7 @@ def dependency_from_details(details)
         Dependency.new(
           name: details["Path"],
           version: version,
-          requirements: details["Indirect"] || dependency_is_replaced(details) ? [] : reqs,
+          requirements: dependency_is_replaced(details) ? [] : reqs,
```
mctofu, line 56: This doesn't 100% fit in with Dependabot's model. The details for a direct or indirect dependency are the same in the `go.mod` so we do have "requirements" for it. However, Dependabot treats a Dependency with requirements as "top_level?" which usually means a direct dependency.

Not supplying requirements for indirect dependencies will require this PR gets more complicated as there are several places that will bail out in the update process if requirements aren't present or changed.

jeffwidman:
Interesting idea for sure. We'll obviously encounter this problem in other ecosystems as well, so we might want to start to think about what needs to change in Dependabot generically so that we don't assume all deps are top-level deps.

go_modules/lib/dependabot/go_modules/file_parser.rb
mctofu: Just to add, this would also cause transititive dependencies to be labeled as direct in the commit metadata. That's used by some users via https://github.com/dependabot/fetch-metadata in determining further actions for a Dependabot PR.

So I think we should fix this correctly and maintain the direct/indirect distinction if possible.

mctofu:
Closing in favor of https://github.com/dependabot/dependabot-core/pull/6361

PR #5557, opened 2022-08-19
Set the author details when updating an Azure pull request.

JManou:
Add a new parameter to set the author details when updating an Azure pull request.
Like in [PullRequestCreator](https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/pull_request_creator/azure.rb) set the author details when creating a commit.
Otherwise azure API returns an error when [commit author email validation](https://docs.microsoft.com/en-us/azure/devops/repos/git/repository-settings?view=azure-devops&tabs=browser#commit-author-email) is enabled.


pavera:
@JManou Thanks for the PR! Would it be possible for you to add a test for this in [azure_spec.rb](https://github.com/dependabot/dependabot-core/blob/main/common/spec/dependabot/pull_request_updater/azure_spec.rb)?

pavera (approved):
Thanks for the contribution and test!

Issue #4909, opened 2022-03-25
`SharedHelpers#with_git_configured` is non-thread safe 


mattt:
The `SharedHelpers#with_git_configured` method performs the passed block within a context where the global Git configuration file is swapped out for one generated by Dependabot: 

https://github.com/dependabot/dependabot-core/blob/e50d44cf5464c1ae1a7db1ce918c350459ed67a5/common/lib/dependabot/shared_helpers.rb#L161

This method is called by many of the ecosystem adapters, including `bundler`, `cargo`, `composer`, `go_modules`, `hex`, `npm_and_yarn`, `pub`, `python`, `terraform`.

As a result, Dependabot cannot be run safely from multiple processes at the same time. Because update jobs are performed in isolated, single-threaded environments, this isn't normally an issue. However, I've been looking at ways to speed up our test suite, and this came up as a blocker.

Using the [`parallel_tests` gem](https://github.com/grosser/parallel_tests) to run our rspec test suite could improve our runtime by a factor of 2–4x. But because of how this is currently implemented, a bunch of tests fail with the Git error `"unable to auto-detect email address"`.

Is there another way that we could execute Git commands with an ephemeral configuration?

mattt:
It looks like `--config-env=` might give us what we need: https://git-scm.com/docs/git#Documentation/git.txt---config-envltnamegtltenvvargt

jurre:
Thanks for opening this @mattt! I've also ran into this when trying to use `parallel_tests` (I've pushed the [branch](https://github.com/dependabot/dependabot-core/compare/jurre/parallel-rspec?expand=1) from when I experimented with this), but I gave up when I ran into this threading issue. _Maybe_ we can use a local git config? We'd just have to make sure to ignore it in cases where we create a temp directory in order to run some package manager commands, and then inspect the files in that directory for changes.

mattt:
@jurre What if we set `GIT_CONFIG_NOSYSTEM=1`? According to [the Git docs](https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables), that _should_ ensure that only the local configuration is read.

jurre:
Yeah that sounds right, let's give it a spin? Would be pretty sweet if we could run things in parallel

mattt:
@jakecoffman Could we replace this helper with a `git` shim like how you're doing with to resolving SSH URLs for Cargo?

jakecoffman:
[...]

PR #5568, opened 2022-08-22
Split slow ecosystems across multiple CI jobs

jurre:
The python tests are too damn slow. This is something that we _should_
fix at the core, by making them faster, but until we do that, let's not
waste precious time waiting on those tests to pass, let's throw more
computers at the problem.

This splits the tests up across multiple CI jobs using the `parallel_test` gem.

You'll notice the builds are still all fairly slow in this PR, but that's because the docker build step isn't being cached here, because we've passed some extra arguments to that step. It will be once this PR is merged.

You'll also see a few required checks now no longer pass, because their names have been changed. I'll update those once this PR is accepted.

There's a few other suites that take more than 5 minutes as well, so I've opted to also split those up.


jurre:
Based on just glancing at the test speeds in the actions outputs of a few recent builds, I would recommend we do this for a few more suites:

- [ ] bundler1: split into 2 jobs
- [ ] bundler2: split into 2 jobs
- [ ] composer: split into 2 jobs
- [ ] go_modules: split into 3 jobs
- [ ] hex: split into 3 jobs
- [ ] npm_and_yarn: split into 5 jobs
- [ ] pub: split into 2 jobs

At that point it might be worth looking if we can somehow reuse the docker images across all those suites instead of building from scratch for each one, but previous experimentation has shown that building for every job was slightly faster because copying the image from storage was relatively slow. It's a lot of wasted compute though, and might have gotten faster?

python/script/parallel-test
```diff
@@ -0,0 +1,14 @@
...
+
+tests = Dir["spec/**/*_spec.rb"].
+        sort.
+        shuffle(random: Random.new(ENV["GITHUB_SHA"].to_i(16))).
+        select.
+        with_index do |_el, i|
+  i % ENV["CI_NODE_TOTAL"].to_i == ENV["CI_NODE_INDEX"].to_i
+end
+
+puts "Running the following tests: #{tests.join(' ')}"
```
jurre, line 12: I don't see this in the job output, why?

landongrindheim: I was wondering the same thing

mattt:
Nice work, @jurre! Adding some links for posterity:

- #4564: Created separate suite for slow Python tests) 
- #4627: Attempt to parallelize Python test suite)
- #4909: Why we can't run tests in parallel automatically with `parallel_rspec` (if we fix this, we might be able to keep 1:1 between ecosystems and CI matrix)

python/script/parallel-test
deivid-rodriguez: No idea either!

landongrindheim:
I'm seeing some pretty big variation between what's being run. I wonder if we could sort the tests being run into buckets before shuffling (though I understand that introduces some brittleness), so that we could make sure they're somewhat evenly distributed?


### Suite 0
image

**Tests run:** 7

### Suite 1
image

**Tests run:** 0

### Suite 2

image

**Tests run:** 4

### Suite 3

image

**Tests run:** 12

### Suite 4

image

**Tests run:** 0

jurre:
@landongrindheim I can't think of a way to do it easily, we'd have to keep track of every files average runtime somewhere or manually tag the files which is annoying and very easy to forget when adding new files.

The `python_slow` suite should already pull out some excessively slow tests into it's own suite, we can re-evaluate that and see if we can move things there.

But with all those approaches you always keep running into the fact that some _files_ are just much slower than others, and you could keep track of every test individually and try to match them up as best as possible etc, but it seems very involved.

There is a service that offers something like this: https://knapsackpro.com/, but I'm not sure if we can get it approved or if it's even worth it. 

If you have an idea on how to do it, I'd suggest we put it on a backlog somewhere to pick up relatively soon but move forward with this approach for now.

deivid-rodriguez:
The `parallel_tests` gem also has that feature of balancing test times. But there's the blockers to use this gem linked above, and also as @jurre mentioned it's tricky to implement even if the balancing itself is built-in, since it indeed requires keeping the file of recorded runtimes in sync. I guess it'd need to be kept as a build artifact or something, upload it after runs on pushes to main, and download it before running tests on PRs.

deivid-rodriguez:
Something else I'm thinking, related to @landongrindheim's comment, is: do we need the shuffling at all? We cannot yet balance this properly, but maybe we can at least make test times stable by always running the same files on each runner?

landongrindheim (approved):
Nice work on this @jurre!

I think the conversation that's coming out of this change is definitely worth having, but I'd be disappointed if we hesitated on shipping this improvement because we hoped we could do better.

Let's keep talking about further improvements while benefiting from this one :shipit:

jurre:
> The `parallel_tests` gem also has that feature of balancing test times. But there's the blockers to use this gem linked above, and also as @jurre mentioned it's tricky to implement even if the balancing itself is built-in, since it indeed requires keeping the file of recorded runtimes in sync. I guess it'd need to be kept as a build artifact or something, upload it after runs on pushes to main, and download it before running tests on PRs.

Yeah, and with `parallel_tests` we're also limited by the number of cores available in the actions VM's (which is [2 I think](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources)), so there are limits to what you can achieve with that approach.

Another thing I hope to look at soon is figuring out how much time in those tests is spent installing a new version of python, listing those versions out somewhere and preinstalling them in the CI image. That way we can cache them in between many runs. There must be more things like that where we can improve the actual runtime of the tests themselves, which seems like that would be the ultimate fix, but it's much more time consuming.

deivid-rodriguez:
> Let's keep talking about further improvements while benefiting from this one :shipit:

Absolutely, my suggestions are definitely not blockers for shipping this, it's an awesome improvement already!

jurre:
> Something else I'm thinking, related to @landongrindheim's comment, is: do we need the shuffling at all? We cannot yet balance this properly, but maybe we can at least make test times stable by always running the same files on each runner?

Yeah we can try it and see if it happens to be a good distribution :D

deivid-rodriguez:
It's going to change over time anyways, but I think it's better to have something stable (even if there's an unlucky machine that doesn't run anything like the example copied above) than something unstable.

deivid-rodriguez:
Just for future reference, apparently parallel_tests also supports this exact use case: https://github.com/grosser/parallel_tests/wiki/Distributed-Parallel-Tests-on-CI-systems. By default it distributes tests by file size.

.github/workflows/ci.yml
```diff
@@ -68,6 +77,9 @@ jobs:
             --env "RAISE_ON_WARNINGS=true" \
             --env "DEPENDABOT_TEST_ACCESS_TOKEN=${{ secrets.GITHUB_TOKEN }}" \
             --env "SUITE_NAME=${{ matrix.suite.name }}" \
+            --env "GITHUB_SHA=${{ github.sha }}" \
```
deivid-rodriguez, line 80: This one is no longer used.

.github/workflows/ci.yml
```diff
@@ -68,6 +77,9 @@ jobs:
...
+            --env "CI_NODE_TOTAL=${{ matrix.suite.ci_node_total }}" \
```
deivid-rodriguez, line 81: I wonder if ` --env "CI_NODE_TOTAL=${{ matrix.suite.ci_node_total || 1 }}"` and similarly below will work, so that we don't need to change the whole matrix, just python, and as a result we don't manually need change all required jobs for CI to be green in UI configuration.

deivid-rodriguez (commented):
I noticed that required jobs are pending because all required job names are now changed. I left an idea that I'm not sure it works, but would leave us with more readable job names, and without needing to change all required jobs in UI configuration manually.

andrcuns:
>At that point it might be worth looking if we can somehow reuse the docker images across all those suites instead of building from scratch for each one, but previous experimentation has shown that building for every job was slightly faster because copying the image from storage was relatively slow. It's a lot of wasted compute though, and might have gotten faster?

@jurre just an FYI, when I was experimenting with arm image builds, I did just that, build image only once and transfer via github artifacts. The image is pretty beefy, it takes roughly 6 minutes to upload the image tar to gh artifacts and something like 30 seconds to download it in the next job.

It is also possible to simply publish dev docker images in to the registry, it will definitely be faster since it will allow reusing existing layers but will create a fair amount of dev tags bloat. Some kind of cleanup setup will be required.

.github/workflows/ci.yml
jurre: We can give it a try, but first I just want to get the builds passing before I clean things up. Keep the suggestions coming though, appreciate it!

python/script/ci-test
```diff
@@ -5,4 +5,4 @@ set -e
 bundle install
 pyenv exec flake8 helpers/. --count --exclude=./.*,./python/spec/fixtures --show-source --statistics
 bundle exec rubocop .
-bundle exec rspec spec
+bundle exec parallel_test spec/ -n $"CI_NODE_TOTAL" --only-group $"CI_NODE_INDEX" --group-by filesize --type rspec
```
deivid-rodriguez, line 8: There may be typos here, did you mean `$CI_NODE_TOTAL` and `$CI_NODE_INDEX`?

jurre: No I meant `"$CI_NODE_TOTAL" and `"$CI_NODE_INDEX`

deivid-rodriguez: Oh, I see, the first quote was off by one :)

jurre:
It _seems_ a bit slower using `parallel_test`, but maybe that's a coincidence.

Now that it's this easy, should I split up a couple other ecosystems as well?

.github/workflows/ci.yml
jurre: I think we don't even need to pass a default value, we just don't use them in those scripts 🤷

.github/workflows/ci.yml
```diff
@@ -20,26 +20,43 @@ jobs:
...
+          - { path: python, name: python, ci_node_total: 5, ci_node_index: 3 }
+          - { path: python, name: python, ci_node_total: 5, ci_node_index: 4 }
+          - { path: python, name: python_slow, ci_node_total: 5, ci_node_index: 0 }
+          - { path: python, name: python_slow, ci_node_total: 5, ci_node_index: 1 }
+          - { path: python, name: python_slow, ci_node_total: 5, ci_node_index: 2 }
+          - { path: python, name: python_slow, ci_node_total: 5, ci_node_index: 3 }
+          - { path: python, name: python_slow, ci_node_total: 5, ci_node_index: 4 }
+          - { path: pub, name: pub, ci_node_total: 2, ci_node_index: 0 }
+          - { path: pub, name: pub, ci_node_total: 2, ci_node_index: 1 }
+          - { path: terraform, name: terraform, ci_node_total: 1, ci_node_index: 0 }
```
landongrindheim, line 58: Please forgive my YAML syntax ignorance 😇

Is it possible to use space inside these curly braces to align these arguments? Ex:
```yaml
 - { path: pub,       name: pub,       ci_node_total: 2, ci_node_index: 1 }
 - { path: terraform, name: terraform, ci_node_total: 1, ci_node_index: 0 }
```

It'd make it easier for _me_ to read.

landongrindheim (approved):
⚡

.github/workflows/ci.yml
jeffwidman: In VS Code, I tend to just right-click in YAML files and select "format document" and it auto-spaces things... whatever it spits out might be a sensible default? 

Also, I filed:
* https://github.com/dependabot/dependabot-core/issues/5572

abu2rki:
> The python tests are too damn slow. This is something that we _should_
> fix at the core, by making them faster, but until we do that, let's not
> waste precious time waiting on those tests to pass, let's throw more
> computers at the problem.
> 
> This splits the tests up across multiple CI jobs using the `parallel_test` gem.
> 
> You'll notice the builds are still all fairly slow in this PR, but that's because the docker build step isn't being cached here, because we've passed some extra arguments to that step. It will be once this PR is merged.
> 
> You'll also see a few required checks now no longer pass, because their names have been changed. I'll update those once this PR is accepted.
> 
> There's a few other suites that take more than 5 minutes as well, so I've opted to also split those up.

bundler/script/ci-test
```diff
@@ -4,11 +4,12 @@ set -e
 
 bundle install
 bundle exec rubocop .
-bundle exec rspec spec
+bundle exec parallel_test spec/ -n "$CI_NODE_TOTAL" --only-group "$CI_NODE_INDEX" --group-by filesize --type rspec
 
 # NOTE: Don't use `if` branches without `else` part, since the code in some of
 # then seems to not abort the script regardless of `set -e`
 
+# Should we only run these on one of the CI_NODE_INDEX's?
```
jeffwidman, line 12: Is this a future TODO that should be merged in or does it need addressing/removing before merge?

jeffwidman (approved):
Very clever!

bundler/script/ci-test
jurre: I was treating it as a future TODO. Should be easy to tackle, but I don't really have more spare cycles to burn on this

ilyazub:
> It _seems_ a bit slower using `parallel_test`, but maybe that's a coincidence.

CI specs run using a single core even when `ci_node_total` is greater than 1.

Examples:
* `1 process for 8 specs, ~ 8 specs per process`: https://github.com/dependabot/dependabot-core/actions/runs/4661988395/jobs/8251872425
* `1 process for 16 specs, ~ 16 specs per process`: https://github.com/dependabot/dependabot-core/actions/runs/4751708654/jobs/8441161997
* `1 process for 16 specs, ~ 16 specs per process`: https://github.com/dependabot/dependabot-core/actions/runs/4669826594/jobs/8268798949#step:5:168

PR #5587, opened 2022-08-25
python/helpers/build: fix a pip warning related to pipfile installation

### LINKED ISSUE #5478: Warning `Using legacy 'setup.py install' for pipfile, since package 'wheel' is not installed.` while building python helper
Noticed [this warning](https://github.com/dependabot/dependabot-core/runs/7646741836?check_suite_focus=true#step:3:9195):

```
Using legacy 'setup.py install' for pipfile, since package 'wheel' is not installed.
```

I'm a little confused, since `wheel` _is_ [already installed in a previous step](https://github.com/dependabot/dependabot-core/runs/7646741836?check_suite_focus=true#step:3:9073).

Perhaps the problem is that both are installed during the same `pip` run and instead need to first install `wheel` and then re-run pip install with the other stuff?

Need to dig further... Looks like there might be some additional info here: https://github.com/pypa/pip/issues/8102

SpecLad:
The `pipfile` package doesn't have published wheels, so pip has to build it from source. By default, it tries to do that in the current environment, but it lacks the `wheel` package, so pip falls back to the legacy `setup.py install` method, which generates a build warning. The `--use-pep517` option lets pip do the build in an isolated environment, automatically downloading the necessary dependencies.

Fixes #5478.

### DIFF
--- a/python/helpers/build
+++ b/python/helpers/build
@@ -18,7 +18,7 @@ cp -r \
   "$install_dir"
 
 cd "$install_dir"
-PYENV_VERSION=3.10.5 pyenv exec pip --disable-pip-version-check install -r "requirements.txt"
+PYENV_VERSION=3.10.5 pyenv exec pip --disable-pip-version-check install --use-pep517 -r "requirements.txt"
 
 # Workaround of https://github.com/python-poetry/poetry/issues/3010
 # By default poetry config file is stored under ~/.config/pypoetry


deivid-rodriguez (approved):
Nice!

jeffwidman (approved):
Thanks @SpecLad !

I confirmed that with this PR the warning is now gone from the CI logs: https://github.com/dependabot/dependabot-core/runs/8015360667?check_suite_focus=true#step:3:7062

We have a temporary code freeze on the build scripts / build infra through the end of this week, so someone will merge this once that freeze is lifted.

Issue #4860, opened 2022-03-16
Resolve rubocop warning - either enable or disable these cops, or default to enabling all new cops


jeffwidman:
I'm seeing this noise in the CI log output that looks like it'd be easy to handle: https://github.com/dependabot/dependabot-core/runs/5559703308?check_suite_focus=true

Does the core team prefer to explicitly enable / disable each linter, or opt-in by default and then disable ones that cause problems?

```
The following cops were added to RuboCop, but are not configured. Please set Enabled to either `true` or `false` in your `.rubocop.yml` file.

Please also note that you can opt-in to new cops by default by adding this to your config:
  AllCops:
    NewCops: enable
Gemspec/RequireMFA: # new in 1.23
  Enabled: true
Layout/LineEndStringConcatenationIndentation: # new in 1.18
  Enabled: true
Lint/AmbiguousOperatorPrecedence: # new in 1.21
  Enabled: true
Lint/AmbiguousRange: # new in 1.19
  Enabled: true
Lint/IncompatibleIoSelectWithFiberScheduler: # new in 1.21
  Enabled: true
Lint/RequireRelativeSelfPath: # new in 1.22
  Enabled: true
Lint/UselessRuby2Keywords: # new in 1.23
  Enabled: true
Security/IoMethods: # new in 1.22
  Enabled: true
Style/NumberedParameters: # new in 1.22
  Enabled: true
Style/NumberedParametersLimit: # new in 1.22
  Enabled: true
Style/OpenStructUse: # new in 1.23
  Enabled: true
Style/RedundantSelfAssignmentBranch: # new in 1.19
  Enabled: true
Style/SelectByRegexp: # new in 1.22
  Enabled: true
For more information: https://docs.rubocop.org/rubocop/versioning.html
```

deivid-rodriguez:
Yes, that's a quite verbose warning.

I would go with enabling or disabling each cop separately, and to setup RuboCop to raise on warnings so that this decision of whether to enable/disable new cops is handled every time Rubocop is upgraded, and warnings no longer don't pile up like this.

PR #5588, opened 2022-08-25
Enable new RuboCop cops

mattt:
Fixes https://github.com/dependabot/dependabot-core/issues/4860

This PR updates our RuboCop configuration to turn on new cops, disables a couple of them,  fixes auto-correctable offenses, and disables a few offenses on a per-file basis.


bundler/dependabot-bundler.gemspec
```diff
@@ -42,4 +42,5 @@ Gem::Specification.new do |spec|
       spec.files << path unless File.directory?(path)
     end
   end
+  spec.metadata["rubygems_mfa_required"] = "true"
```
Nishnha, line 45: I don't think we would want to enable MFA here 
https://guides.rubygems.org/mfa-requirement-opt-in/

Nishnha: This line is also included in the other `dependabot-*.gemspec` files

mattt: Nice catch! I'll remove that and disable the relevant check.

mattt: Fixed by 1cc36ec31.

Nishnha (approved):

.rubocop.yml
```diff
@@ -2,13 +2,16 @@
...
   Exclude:
-  - "*/vendor/**/*"
-  - "*/bin/**/*"
-  - "**/tmp/**/*"
-  - "*/spec/fixtures/**/*"
+    - "*/vendor/**/*"
+    - "*/bin/**/*"
+    - "**/tmp/**/*"
+    - "*/spec/fixtures/**/*"
+  NewCops: enable
```
jeffwidman, line 9: 👍

jeffwidman:
Could you squash [Revert "Fix Style/SelectByRegexp offenses"](https://github.com/dependabot/dependabot-core/pull/5588/commits/a33608815626ac9097ddb2f36b4765abec6fd189)?

Same with:
* https://github.com/dependabot/dependabot-core/pull/5588/commits/fe7244db06886861ec53153748b7bbf147effb52
* https://github.com/dependabot/dependabot-core/pull/5588/commits/e6f2293d48496c10a58310099149808fd570a58b

as it looks like these last two are just reverting the changes from https://github.com/dependabot/dependabot-core/pull/5588/commits/fc46074365b893633c48110a65e73eb0735afa18 so all three commits can be simply dropped from the commit log IIUC?

jeffwidman (approved):
Thanks so much for doing this!

Be nice (but not required) to cleanup the git changelog just to minimize `git blame` spelunking pain for future readers of various files.

The sooner you can land this the better IMO in order to minimize merge pain from other PRs.

mattt:
> Be nice (but not required) to [cleanup the git changelog](https://github.com/dependabot/dependabot-core/pull/5588#issuecomment-1230742352) just to minimize `git blame` spelunking pain for future readers of various files.

I was planning the merge commit to a [`.git-blame-ignore-revs`](https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/). Sound good?

deivid-rodriguez:
My question is, why caring about splitting each cop to its own commit, if the whole thing will be ignored? Is the idea to still allow fine grained `git blame` with `git` CLI of the real changes, but ignore it by default in GitHub blame view?

deivid-rodriguez:
Or maybe to make reviewing easier?

jeffwidman:
> I was planning the merge commit to a [.git-blame-ignore-revs](https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/). Sound good?

Oh neat! Didn't know that existed! That said, I'd personally rather avoid it... I would find it surprising/confusing if I'm spelunking through old code changes (which I do a _lot_ of here in core) and suddenly the code changes unexpectedly w/o showing the commit between them. 

I'm very okay with the stream of commits that you have, and personally slightly prefer the one-commit-per-cop workflow as I find it more readable than if they're squashed into a single commit... other than the reverting commits, those I'd squash so they disappear... but one commit vs many isn't something I feel strongly on.

jeffwidman:
I have a _slight_ preference for default enable new cops. I realize it can cause stuff to suddenly get flagged / break the rubocop version updates, but it's still convenient to start flagging new things w/o having to manually fix and then be like "oh, I wonder if there's now a cop for this?" and then waste time researching...

That said, not something I feel strongly about, so if others prefer default disable, I'm happy to work with that too.

mattt:
Rebased onto latest `main` post-merge without any conflicts, and `rubocop updater` succeeds, so I'm feeling good about this.

mattt:
> Oh neat! Didn't know that existed! That said, I'd personally rather avoid it... I would find it surprising/confusing if I'm spelunking through old code changes (which I do a _lot_ of here in core) and suddenly the code changes unexpectedly w/o showing the commit between them.

Alright, we can hold off on ignoring any commits by default until we deem them to be too noisy.


> My question is, why caring about splitting each cop to its own commit, if the whole thing will be ignored? Is the idea to still allow fine grained `git blame` with `git` CLI of the real changes, but ignore it by default in GitHub blame view?

Splitting up changes across individual commits provides the flexibility to choose how to manage history until the end of the process. It also makes the PR easier to reason about for reviewers.

Issue #4652, opened 2022-01-20
`@dependabot recreate` does not actually recreate the commit


jeffwidman:
I had a dependabot PR that I was munging around with trying to see what additional changes were required to get CI to pass.

However, I accidentally fat-fingered some git commands and dropped the original commit from Dependabot, both locally and on the remote.

So I did a `@dependabot recreate` since the description of that command says:
```
@dependabot recreate will recreate this PR, overwriting any edits that have been made to it
```

However, when Dependabot recreated the PR, it didn't reset the commit title or description. It merely reset the code changes of the commit.

I think this is a bug because recreating the PR should recreate the commit metadata, title, description, etc. Honestly, that's a big part of the usefulness of Dependabot, is the detailed commit messages listing what changed and links to more info. Very useful for later `git blame` spelunking when trying to figure out what may have caused a problem.

mctofu:
I think this happens because a recreate triggers an update of the PR and we currently expect that we can re-use the message from the previous commit: https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/pull_request_updater/github.rb#L186

In this case the original commit was lost so it probably used the description from another commit in the PR?

jeffwidman:
> In this case the original commit was lost so it probably used the description from another commit in the PR?

Yep, that's exactly what happened.

And that code section is what needs updating. But before opening a PR to fix, wanted to make sure we were all in agreement on what the desired behavior here should be.

mctofu:
I do think the ideal behavior here would have been to have the proper message on the commit. However, I'm seeing that in the spot that we are calling the PR updater internally we don't have the metadata available... We'd need to track down why that is the case and adjust it first.

solvaholic:
> > In this case the original commit was lost so it probably used the description from another commit in the PR?
> 
> Yep, that's exactly what happened.
> 
> And that code section is what needs updating. But before opening a PR to fix, wanted to make sure we were all in agreement on what the desired behavior here should be.

In what scenarios do y'all envision @dependabot would re-use a commit message from a _different_ pull request?

My Ruby fu is weak. Unless Ruby does something surprising with [this bit of code](https://github.com/dependabot/dependabot-core/blob/ed74216913c17803e5175645bab9ba8c2b0257a7/common/lib/dependabot/pull_request_updater/github.rb#L201-L204), tho, I think any commit `commit_being_updated` returns should be from the same pull request where `@dependabot recreate` was run:

```ruby
            commit =
[...]

PR #5595, opened 2022-08-29
More descriptive PR message when multiple dependencies are fixed in a security update

Nishnha:
This PR adds a `#transitive_multidependency_intro` method to provide a more descriptive PR intro when multiple dependencies are updated in a transitive dependency update.

Some caveats:

- Avoids using the transitive multidependency message in lieu of the [removed dependency message](https://github.com/dependabot/dependabot-core/pull/5549/files#diff-c56495d54a4d5d788fbf17bf39e34b0330baee9d52d77117a5ce1cc4eb71f494) if any dependency is removed in an update.

- Still uses the old PR message when a single transitive dependency is updated (see https://github.com/dsp-testing/dependabot-reproduce-update-not-possible/pull/66 for an example)

---

An example PR after this change would read as:

> Bumps [node-forge](https://github.com/digitalbazaar/forge) to 1.3.1 and updates ancestor dependency [webpack-dev-server](https://github.com/webpack/webpack-dev-server). These dependencies need to be updated together. 

Also see the image below:

Screen Shot 2022-09-15 at 3 13 09 PM


common/lib/dependabot/pull_request_creator/message_builder.rb
```diff
@@ -239,6 +242,20 @@ def multidependency_intro
         "dependencies needed to be updated together."
       end
 
+      def vulnerable_multidependency_intro
+        msg = "Bumps #{dependency_links[0]} to a non-vulnerable version "
```
mctofu, line 246: There's a few other cases here:
1. The vulnerable dependency is removed rather than bumped.
2. Multi dependency PRs that are updating multiple direct dependencies

Not necessarily a blocker but it be useful to have some examples of what these PRs would look like.

Nishnha:
Another heuristic for transitive dependency updates that @mctofu mentioned is whether the dependencies being updated have a mix of `top_level?` and non top-level dependencies.

Since it is possible to have a transitive dependency update that only updates the transitive dependency, I think we can get away with checking if any of the dependencies being updated is not a `top_level?` dep

common/lib/dependabot/pull_request_creator/message_builder.rb
```diff
@@ -192,14 +192,22 @@ def requirement_commit_message_intro
...
+      # rubocop:disable Metrics/AbcSize
       def version_commit_message_intro
+        dependency = dependencies.first
+
         return multidependency_property_intro if dependencies.count > 1 && updating_a_property?
 
         return dependency_set_intro if dependencies.count > 1 && updating_a_dependency_set?
 
+        return transitive_multidependency_intro if dependencies.count > 1 &&
+                                                   dependencies.any?(&:top_level?) &&
```
mctofu, line 205: We might be able to avoid disabling the rubocop rules & make this more consistent with the other checks if you extracted these to another method. Maybe `updating_top_level_and_transitive_dependencies?`?

mctofu (approved):
Looks good! Just had some minor cleanup suggestions.

common/lib/dependabot/pull_request_creator/message_builder.rb
Nishnha: Fixed in https://github.com/dependabot/dependabot-core/pull/5595/commits/306413f012f903f8dc23721b1e73a044aac45d05

That allowed me to remove the rubocop:disable CyclomaticComplexity line!

common/lib/dependabot/pull_request_creator/message_builder.rb
```diff
@@ -192,14 +192,22 @@ def requirement_commit_message_intro
       end
 
       # rubocop:disable Metrics/PerceivedComplexity
+      # rubocop:disable Metrics/CyclomaticComplexity
+      # rubocop:disable Metrics/AbcSize
       def version_commit_message_intro
+        dependency = dependencies.first
```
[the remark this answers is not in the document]
Nishnha: Fixed in https://github.com/dependabot/dependabot-core/pull/5595/commits/946eacc5aeb82355051f696e62bd50d7f0dab457

Nishnha:
I deployed this but haven't been able to generate a PR that uses the updated copy.

Not sure if it's due to how `@dependabot recreate` works, but I would expect the recreated PR to use the new the PR message.

Oddly enough, the dry-runner is providing the updated copy, so maybe this will only work for newly created PRs.

Nishnha:
Here's an example in production using the updated copy https://github.com/dsp-testing/erinhav-update-not-possible/pull/8

> Bumps [node-forge](https://github.com/digitalbazaar/forge) to 1.3.1 and updates ancestor dependency [webpack-dev-server](https://github.com/webpack/webpack-dev-server). These dependencies need to be updated together.

jeffwidman:
Yeah, from https://github.com/dependabot/dependabot-core/issues/4652:
> However, when Dependabot recreated the PR, it didn't reset the commit title or description. It merely reset the code changes of the commit.

Implementation details in https://github.com/dependabot/dependabot-core/issues/4652#issuecomment-1021724508, hopefully we can fix that at some point...

PR #5597, opened 2022-08-29
Why is wheel needed?

jeffwidman:
WIP for debugging https://github.com/dependabot/dependabot-core/issues/5478#issuecomment-1227107158...

### DIFF
--- a/python/helpers/requirements.txt
+++ b/python/helpers/requirements.txt
@@ -5,7 +5,6 @@ hashin==0.17.0
 pipenv==2022.4.8
 pipfile==0.0.2
 poetry==1.1.15
-wheel==0.37.1
 
 # Some dependencies will only install if Cython is present
 Cython==0.29.32


deivid-rodriguez:
Interesting, no broken specs. @SpecLad where did you find wheel being depended on, do you remember?

SpecLad:
I don't know exactly how it's used, I just looked at the commit it was added in (21ed214f), and figured it was for a reason.

jurre:
I think nothing happens because it is still installed by [pip-tools](https://github.com/jazzband/pip-tools), which also requires it:

https://github.com/jazzband/pip-tools/blob/8d4a30ff02fd736553f10a5565168b3d03558500/setup.cfg#L39

At the time of that commit, I don't think it did.

deivid-rodriguez (approved):
That makes total sense @jurre, since https://github.com/jazzband/pip-tools/commit/511a7a10fe11a082586103a787c4aea76a39aec1 happened much later than https://github.com/dependabot/dependabot-core/commit/21ed214fee0b4be6ee4704def7d60f662dd5a111 👍.

And the answer to "why is wheel needed?" would be because dependabot-core sometimes needs to fallback to running `pip-tools` in `--no-build-isolation` mode, which requires `wheels` preinstalled.

So I think this PR should be good to go too!

deivid-rodriguez:
BUT, I though `requirements.txt` was something like a lockfile for Python, right? However, if I understand correctly, with this change, dependabot will no longer use a fixed `wheel` version, but a rolling one, right? If that's the case I think I'd rather leave things as they are?

SpecLad:
FWIW, `requirements.txt` only works as a lockfile if you specify versions for all (transitive) dependencies, which this one doesn't.

deivid-rodriguez:
I see, thanks for that information!

I generally prefer locked dependencies so I say let's close this one now that we understood why `wheel` is needed.

jurre:
Yeah so ironically pip-tools is actually what lets you use the requirement.txt as a lockfile, you specify your top-level dependencies in a `requirements.in` file, and it dumps everything including transitive dependencies as locked versions in a `requirements.txt` file, which can then be installed by pip. We should probably investigate locking our python deps in some way, although there are many options (pip-compile, poetry etc), and it doesn't seem like the community has picked a clear winner yet

deivid-rodriguez:
When I searched about this a few days ago, the most appealing choice to me seemed `pyproject.toml` + `poetry.lock`.

jurre:
> When I searched about this a few days ago, the most appealing choice to me seemed `pyproject.toml` + `poetry.lock`.

I'm open to it! We have a few experienced python people on our team, so let's ask them to chime in as well

pavera:
This may already be addressed in newer python best practices, but an impediment I've consistently run into with using different package managers in python is you almost always still end up with a bare requirements.txt to at least specify the package manager + pin some package manager dependencies in order to bootstrap the package manager with pip.

In our case where we're just trying to bootstrap a few package managers it seems overkill to have a requirements.txt to specify poetry + pin a couple dependencies, and then define the rest of the package managers in pyproject.toml

I would also be worried (quite possibly unnecessarily) about our version of poetry clashing with a requested version in a pyproject.toml from a repo we're trying to update.  This may already be an issue in our requirements.txt approach as well, but as we attempt to implement proper PEP 518 support I think choosing a top level package manager for the helper might  complicate things.

deivid-rodriguez:
I agree that a migration to a package manager that locks dependencies to manage our own Python dependencies is overkill at this point.

I would personally close this PR too, because keeping `wheel` locked still seems a good idea.

jeffwidman:
Yeah, def agree with you @deivid-rodriguez thanks for digging into this a bit.

PR #5604, opened 2022-08-31
Propagate author details when initializing PullRequestUpdater for Azure.

JManou:
Propagate author details when initializing PullRequestUpdater for Azure.
Follow up of #5557 .


jeffwidman (approved):
Thanks!

We've got an internal code freeze through the end of the week, so will merge once that's lifted.

PR #5606, opened 2022-08-31
Configure RSpec to allow running `rspec --only-failures`

deivid-rodriguez:
When I make changes and break stuff, I like running `rspec --only-failures` to focus on fixing what I broke. This allows RSpec to keep track of what's currently failing, so that this flag can actually be used.

What do you think?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ vendor
 coverage/
 .ruby-gemset
 .tool-versions
+.rspec_status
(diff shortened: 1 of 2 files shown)


Nishnha (approved):
Just tested this out on my machine and confirmed it makes a separate `.rspec_status` file for each ecosystem directory you run rspec in

Also, thanks for sharing a cool rspec feature!

jeffwidman (approved):

PR #5607, opened 2022-08-31
Run `rspec` with `--profile` flag by default in Python

deivid-rodriguez:
This flag prints the slowest 10 specs and their run times after each rspec run.

Because our Python test suite is so slow, It just occurred to me that maybe using it by default in Python would raise awareness on the areas that need more love with regards to test speed.

### DIFF
--- a/python/spec/spec_helper.rb
+++ b/python/spec/spec_helper.rb
@@ -26,4 +26,6 @@ RSpec.configure do |config|
       example.skip
     end
   end
+
+  config.profile_examples = 10
 end


jeffwidman (approved):

jurre (approved):

deivid-rodriguez:
@jurre I just noticed that updater does this same thing, but enabling profiling through configuration instead (`config.profile_examples`), so I changed this PR to do just that too for consistency 👍. I will merge after CI if that's ok!

PR #5608, opened 2022-09-01
Merge previously private updater code into core

jurre:
This adds a new component to dependabot-core that is only used and useful to GitHub internally, it's part of how we run Dependabot in our internal systems. By merging this into the main codebase we hope to be able to simplify both it and how core works going forward, and it will open up some simplifications in our deployment and development processes.

This code is not useful to anyone outside of GitHub, other than maybe informative in how to run dependabot-core in a service setting. It's also explicitly _not_ a goal for this code to be useful or reusable outside of its current context.

One thing that _is_ useful to outside contributors, is that this change also includes a set of smoke tests that run against both the core ecosystems and the updater. This will give us higher confidence when reviewing + merging PRs.

The code in this PR was contributed to by many people, and it has a rich history available in an internal repository. Since those commits were made in a private context, we've squashed all of them in a single commit and omitted any non GitHub-owned email-addresses.


jakecoffman (approved):
LGTM!

PR #5610, opened 2022-09-01
publish main branch images

jakecoffman:
This should publish a tagged image for commits on main too. This way we can deploy a SHA from a main merge commit.


honeyankit (approved):
LGTM!

PR #5614, opened 2022-09-01
only use GHCR for deploys

jakecoffman:
We created two ways to deploy Dependabot Updater internally: 

1. Push the Updater to GHCR
1. Export an image of the Updater and save it as an artifact for more processing later

Due to layer caching and the size of the image, saving it as an artifact takes a significant amount of time compared to just using GHCR. So we're going to remove it and just use GHCR.

### DIFF
--- a/.github/workflows/docker-branch-releases.yml
+++ b/.github/workflows/docker-branch-releases.yml
@@ -44,7 +44,7 @@ jobs:
             -t "$UPDATER_IMAGE:$TAG" \
             --build-arg BUILDKIT_INLINE_CACHE=1 \
             --cache-from "$BASE_IMAGE" \
-            --cache-from "$UPDATER_IMAGE:latest" \
+            --cache-from "$UPDATER_IMAGE_MIRROR" \
             --build-arg OMNIBUS_VERSION=$TAG \
             -f Dockerfile.updater \
             .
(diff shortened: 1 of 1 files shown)


jeffwidman (approved):
Can you please squash before merging? I'm not opposed to multiple commits, just these mostly look like fixups so rather squash them down into the primary commit

.github/workflows/docker-branch-releases.yml
```diff
@@ -56,13 +56,14 @@ jobs:
...
-          name: updater-${{ github.sha }}.tar
-          path: dependabot-updater.tar
+          docker tag "$UPDATER_IMAGE:$TAG" "$UPDATER_IMAGE_MIRROR:latest"
+          docker push "$UPDATER_IMAGE_MIRROR:latest"
       - name: Set summary
-        run: echo "updater uploaded \`ghcr.io/dependabot/dependabot-updater/dependabot-updater:$TAG\`" >> $GITHUB_STEP_SUMMARY
+        run: |
+          echo "updater uploaded with tag \`$TAG\`" >> $GITHUB_STEP_SUMMARY
+          echo "\`\`\`" >> $GITHUB_STEP_SUMMARY
+          echo "ghcr.io/dependabot/dependabot-updater:$TAG" >> $GITHUB_STEP_SUMMARY
```
mattt, line 68:
```suggestion
          echo "$UPDATER_IMAGE_MIRROR:$TAG" >> $GITHUB_STEP_SUMMARY
```

mattt (approved):

PR #5615, opened 2022-09-01
To prevent dependabot-core from failing when the incorrect release tag is created for a release, adding a rescue statement

honeyankit:
When the user creates a wrong release [tag](https://github.com/aio-libs/yarl/releases/tag/v1.8.0--failed-release-attempt), the dependabot is not handling the exception and failing. The fix in this PR will catch such exception and will return nothing for such cases.

Error:
```
updater | ERROR <job_446384276> Error processing yarl (ArgumentError)
updater | ERROR <job_446384276> Malformed version number string 1.8.0..failed.release.attempt
```

- Error logs before fix: [failed-argument-error.log](https://github.com/dependabot/dependabot-core/files/9473822/failed-argument-error.log)
- Logs after fix: [fixed-agrement-error.log](https://github.com/dependabot/dependabot-core/files/9473829/fixed-agrement-error.log)

------ 
Original issue: https://github.com/github/dependabot-updates/issues/2776


deivid-rodriguez (commented):
Hei @honeyankit! Nice to get rid of all these Python errors 🎉.

However I believe this `rescue` might not be the best solution.

While it correctly ignores this case, it may ignore many other things, including programming mistakes!

I think even if with decide to go with a silencing rescue, it should be more scoped to the exact place where this issue is happening.

How about trying to fix the actual culprit instead? I think you already identified the issue, this tag is getting incorrectly flagged as `correct?` by our version regexp. And that's why we later try to instantiate a `Python::Version` out of it.

I think the double dash is suspicious, I think the pattern may be unintentionally allowing that. Does this patch fix things?

```diff
diff --git a/python/lib/dependabot/python/version.rb b/python/lib/dependabot/python/version.rb
index 525be66d6..0290de9f0 100644
--- a/python/lib/dependabot/python/version.rb
+++ b/python/lib/dependabot/python/version.rb
@@ -16,7 +16,7 @@ module Dependabot
 
       # See https://peps.python.org/pep-0440/#appendix-b-parsing-version-strings-with-regular-expressions
       VERSION_PATTERN = 'v?([1-9][0-9]*!)?[0-9]+[0-9a-zA-Z]*(?>\.[0-9a-zA-Z]+)*' \
-                        '(-[0-9A-Za-z-]+(\.[0-9a-zA-Z-]+)*)?' \
+                        '(-[0-9A-Za-z]+(\.[0-9a-zA-Z]+)*)?' \
                         '(\+[0-9a-zA-Z]+(\.[0-9a-zA-Z]+)*)?'
       ANCHORED_VERSION_PATTERN = /\A\s*(#{VERSION_PATTERN})?\s*\z/.freeze
 
```

mattt:
> How about trying to fix the actual culprit instead? I think you already identified the issue, this tag is getting incorrectly flagged as `correct?` by our version regexp. And that's why we later try to instantiate a `Python::Version` out of it.

The [linked PEP440 doc](https://peps.python.org/pep-0440/#appendix-b-parsing-version-strings-with-regular-expressions) provides a regex pattern for matching version identifiers, but it's not obvious to me how that relates to the `VERSION_PATTERN` we use here. You might try copying that verbatim to see if that solves the problem at hand.

mattt:
> I had the same thought, both regexp don't look the same to me. Anyways, I suggested a focused patch to what we currently have, because it seems "mostly" correct already (as per the different tests cases in the `Version` class unit specs).

A focused change sounds good to me. And thanks for verifying the existing behavior in Python.

If we do change the regex, I'd really like to see a comment explaining that our regex has diverged from PEP440 (bonus points if we can articulate why).

deivid-rodriguez:
For what it's worth the Python regexp never includes dashes in alphanumeric character classes, so this change at least brings us closer to the "upstream regexp".

honeyankit:
@deivid-rodriguez thanks for reviewing the PR and providing the suggestion. My initial thought was to modify the regex and remove the double-dash, since single dash is acceptable with for the python version. I tried version check with some other python module and it came out correct making me believe that version is correct, so changed my direction and use rescue  to catch the exception.

python/lib/dependabot/python/version.rb
```diff
@@ -16,7 +16,7 @@ class Version < Gem::Version
 
       # See https://peps.python.org/pep-0440/#appendix-b-parsing-version-strings-with-regular-expressions
       VERSION_PATTERN = 'v?([1-9][0-9]*!)?[0-9]+[0-9a-zA-Z]*(?>\.[0-9a-zA-Z]+)*' \
-                        '(-[0-9A-Za-z-]+(\.[0-9a-zA-Z-]+)*)?' \
+                        '(-[0-9A-Za-z]+(\.[0-9a-zA-Z]+)*)?' \
```
pavera, line 19:
```suggestion
                        '([0-9A-Za-z]+(\.[0-9a-zA-Z]+)*)?' \
```
Does this address the issue you mentioned internally?  [PEP 440](https://peps.python.org/pep-0440/#examples-of-compliant-version-schemes) has no valid examples with dashes so I think this gets even closer to the spec.

deivid-rodriguez: From trying to figure https://peps.python.org/pep-0440/#appendix-b-parsing-version-strings-with-regular-expressions out, dash is only allowed to precede an explicitly list of pre|post|dev release identifiers. For example, "1.8-alpha" would be a valid version string while "1.8-arfa" would not.

honeyankit: @pavera : We need to have a dash in the pre/post/dev as per https://peps.python.org/pep-0440/#pre-release-separators

```
Pre-releases should allow a ., -, or _ separator between the release segment and the pre-release segment.
The normal form for this is without a separator. This allows versions such as 1.1.a1 or 1.1-a1 which would be 
normalized to 1.1a1. It should also allow a separator to be used between the pre-release signifier and the numeral. 
This allows versions such as 1.0a.1 which would be normalized to 1.0a1.
```

pavera (approved):

PR #5618, opened 2022-09-02
break ci into files to use paths: so we can skip unneeded test runs

jakecoffman:
Dependabot rebases PRs and kicks off every CI job so we're running out of Actions runners. This reduces the amount of CI jobs run by only running ones that are applicable to the change.


landongrindheim (dismissed):
I ❤️ this change!

jakecoffman:
Unfortunately I think we're going to have to remove the required checks for the individual CI runs for this to work. I thought about making `ci-bundler-skip.yml` that did a `paths-ignore: [bundler]`, but if you edit `bundler/` plus common it would still run the skip I think 🤔

.github/workflows/ci-bundler.yml
```diff
@@ -0,0 +1,36 @@
+name: CI Bundler
+on:
+  push:
+    branches:
+      - "main"
+    paths:
+      - Dockerfile
+      - common
+      - bundler
```
landongrindheim, line 9: I wanted to highlight to anyone concerned with this change that _all_ workflows will run when either `common/`  or `Dockerfile` change.

jakecoffman: Assuming I can get the pattern right!

jakecoffman:
closing for #5620

PR #5620, opened 2022-09-02
skip CI if no changes effect the ecosystem

jakecoffman:
Dependabot rebases PRs and kicks off every CI job so we're running out of Actions runners. This reduces the amount of CI jobs run by only running ones that are applicable to the change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,9 +58,88 @@ jobs:
           - { path: terraform, name: terraform }
 
     steps:
+      - uses: dorny/paths-filter@v2
+        id: changes
+        with:
+          filters: |
+            bundler:
+              - Dockerfile
+              - 'common/**'
+              - 'bundler/**'
+            cargo:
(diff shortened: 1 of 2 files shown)


mattt (approved):

.github/workflows/ci.yml
```diff
@@ -58,9 +58,88 @@ jobs:
...
+              - 'common/**'
+              - 'github_actions/**'
+            go_modules:
+              - Dockerfile
+              - 'common/**'
+              - 'go_modules/**'
+            gradle:
+              - Dockerfile
+              - 'common/**'
+              - 'gradle/**'
```
jurre, line 103: Gradle also pulls in some files from the Maven implementation: https://github.com/dependabot/dependabot-core/blob/cafef062f5bb155e0735d6421a872f36d0240c19/gradle/lib/dependabot/gradle/requirement.rb#L4

There might be a few more cases like this but this is one I know of.

PR #5623, opened 2022-09-02
branch naming for removed dependencies

mctofu:
Fixes an error when generating the branch name for a PR with a removed transitive dependency.

In a multi-dependency PR we'll include the version of the first dependency in the branch name.

ex: `dependabot/dummy/business-and-statesman-1.5.0`

For a removed transitive dependency PR the version of the first dependency is nil. Based on a suggestion in https://github.com/dependabot/dependabot-core/pull/5623#discussion_r962064279 I'm converting this to `-removed`
ex: `dependabot/dummy/business-and-statesman--removed`

Are there other options you'd prefer?

1. Omit the trailing dash.
ex: `dependabot/dummy/business-and-statesman`
I like this option but was wondering if anyone might depend on there being an extra dash?

2. Use a placeholder value without an extra dash.
ex: `dependabot/dummy/business-and-statesman-removed`
It could collide if "removed" happens to be a valid version of the dependency although this is likely pretty rare? 

3. Use a blank version
ex: `dependabot/dummy/business-and-statesman-`
Maintains the current format but looks like something went wrong IMO.


common/spec/dependabot/pull_request_creator/branch_namer_spec.rb
```diff
@@ -247,6 +247,43 @@
...
+          previous_requirements: [{
+            file: "Gemfile",
+            requirement: "~> 1.4.0",
+            groups: [],
+            source: nil
+          }]
+        )
+      end
+
+      it { is_expected.to eq("dependabot/dummy/business-and-statesman-") }
```
landongrindheim, line 284: What do you think about `dependabot/dummy/business-and-statesman--removed`? I think using two dashes is a nice middleground that would signal the dependency as removed _and_ prevent conflicts.

landongrindheim (approved):
I think this is good to merge. I like both of your suggestions for further improvement! I suggested a _third_ option 😵‍💫  but I don't think you can go really wrong with any of them 🙂

common/spec/dependabot/pull_request_creator/branch_namer_spec.rb
mctofu: I like it as that makes a conflict much less likely. Updated in https://github.com/dependabot/dependabot-core/pull/5623/commits/e0dabbcab9a9f81a3a892b3185f96c230e39c015.

PR #5624, opened 2022-09-05
Speedup rebuilding native helper changes

deivid-rodriguez:
I was working on improving #5617, and while changing the `pub` native helper script, I noticed that rebuilding would rebuild native helpers for all ecosystems.

I think reordering things this way should make things a bit more efficient.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -268,12 +268,12 @@ USER dependabot
 ENV DEPENDABOT_NATIVE_HELPERS_PATH="/opt"
 
 COPY --chown=dependabot:dependabot composer/helpers /opt/composer/helpers
-RUN bash /opt/composer/helpers/v1/build
-RUN bash /opt/composer/helpers/v2/build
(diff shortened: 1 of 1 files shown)


jurre:
Change is 👍 to me, but did you verify it's actually faster? 😄 I've been surprised by Docker caching behavior a few times

deivid-rodriguez:
Yeah, I did this while making changes to `pub` native helpers and after this when rebuilding you would no longer rebuild composer helpers, which are quite slow to rebuild, and others, making things faster!

Here, worst case scenario is changing `composer` helpers. No speedup there because it's the first in the list so changing any native helper file there triggers a rebuild of everything.

Best case scenario is changing `terraform` helpers. Since they come last in the list, changes in them will only cause rebuilding terraform native helpers but nothing else from other ecosystems.

jurre (approved):

Nishnha:
This is really surprising behavior. Thanks for the speed up!

PR #5636, opened 2022-09-06
fix latest docker image not being pushed on merge to main

jakecoffman:


deivid-rodriguez (approved):
How did I miss this?! Nice catch! 😍

PR #5638, opened 2022-09-06
pull request message is always built, removing experiment

jakecoffman:


landongrindheim (approved):

updater/spec/dependabot/integration_spec.rb
```diff
@@ -37,6 +37,9 @@
     allow(Dependabot::Environment).to receive(:token).and_return("some_token")
     allow(Dependabot::Environment).to receive(:job_id).and_return(job_id)
     allow(Dependabot.logger).to receive(:info).and_call_original
+    message_builder = double(Dependabot::PullRequestCreator::MessageBuilder)
+    allow(Dependabot::PullRequestCreator::MessageBuilder).to receive(:new).and_return(message_builder)
+    allow(message_builder).to receive(:message).and_return(nil)
```
landongrindheim, line 42: ❓ Since this is an integration spec, do we want to return `nil` here? I could see value coming from ensuring that this is exercised.

I'm fine with this being added in a separate PR if you think it's valuable. I'll trust your judgment on whether it should be 🙂

landongrindheim (approved):

.github/workflows/smoke.yml
```diff
@@ -41,60 +41,74 @@ jobs:
...
     - uses: actions/checkout@v3
     - uses: dorny/paths-filter@v2
+      if: github.event_name != 'workflow_dispatch'
       id: changes
       with:
         filters: |
           actions:
             - Dockerfile
             - 'common/**'
+            - 'updater/**'
```
landongrindheim, line 51: I wonder if we could [re-use](https://stackoverflow.com/a/56171053) the shared filters:

```yaml
fruit:
  - &banana
    name: banana
    colour: yellow

food:
  - *banana
  - name: carrot
    colour: orange
```

updater/spec/dependabot/integration_spec.rb
jakecoffman: We're testing it in the smoke tests now pretty extensively. I'm leaning towards not testing it here too 🤔

.github/workflows/smoke.yml
jakecoffman: I was reading about this the other day and wondered if it was too weird to introduce. I'll give it a shot.

landongrindheim: Maybe good to do in its own PR?

jakecoffman: It's good here, already exposed a few copy paste errors!

jakecoffman: Actually it looks like this isn't possible with arrays, there's no way to flatten it, so it's placing the shared array in the outer array and I'm assuming that won't work well with this action.

PR #5640, opened 2022-09-06
push docker images after approval

jakecoffman:
We're working on improving our deployment process!

### DIFF
--- a/.github/workflows/docker-branch-releases.yml
+++ b/.github/workflows/docker-branch-releases.yml
@@ -10,13 +10,15 @@ on:
     paths-ignore:
       - "CHANGELOG.md"
       - "common/lib/dependabot/version.rb"
-  pull_request:
+  pull_request_review:
+    types: [submitted]
     paths-ignore:
       - "CHANGELOG.md"
       - "common/lib/dependabot/version.rb"
 
 jobs:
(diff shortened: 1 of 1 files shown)


honeyankit (approved):

.github/workflows/docker-branch-releases.yml
```diff
@@ -10,13 +10,15 @@ on:
...
-  pull_request:
+  pull_request_review:
+    types: [submitted]
     paths-ignore:
       - "CHANGELOG.md"
       - "common/lib/dependabot/version.rb"
 
 jobs:
   push-updater-image:
+    if: github.event.review.state == 'approved'
```
mctofu, line 21: It doesn't look like the review is included in a push payload? https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#push

mctofu: `github.event_name == 'push'` might work here? 

https://docs.github.com/en/actions/learn-github-actions/contexts#example-contents-of-the-github-context

jakecoffman: True, I just broke the push again that I got working this morning. 😅

Issue #5639, opened 2022-09-06
github links in release notes cause timeline notification spam


ljharb:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

npm

### What you expected to see, versus what you actually saw

expected no timeline notifications

saw thousands of notifications, and it seems to be infinitely increasing: https://twitter.com/bitandbang/status/1567066927893303296 → https://twitter.com/NicoloRibaudo/status/1567065351522848768 → https://github.com/tc39/ecma262/pull/2819

deivid-rodriguez:
Thanks for letting us know about this @ljharb! We're aware how disrupting this kind of timeline references can be.

We have identified the issue as specific to releases including the `org/repo#pr` shortcut in their changelog or release notes, and we're already working on a fix.

I will post further updates once this is resolved!

ljharb:
Thanks!

deivid-rodriguez:
Just confirming that the fix was deployed! :)

jnyrup:
Did this regress?

E.g. https://github.com/microsoft/vstest/pull/15720 now has +19000 items, which practically makes the timetable unusable.

Issue #3276, opened 2021-03-15
Escape links in release notes


asciimike:
Currently we escape @ mentions of users in Dependabot PRs by adding a ZWJ between the @ and the username, so that users don't constantly get notified of everything Dependabot is doing.

We should likely do something similar with links in release notes, since GitHub linkifies the release note links and shows them in the maintainer's repos.

Without this, maintainers are crushed under thousands of useless back-references from Dependabot.

Examples:
 - https://twitter.com/MrJamesHenry/status/1371492139180392459 (and https://github.com/eslint/eslint/pull/14178)
 - https://github.com/laravel/framework/pull/36480

brrygrdn:
We have identified this as a regression in the Dependabot service itself, and a fix will go out shortly to prevent this behaviour in future, with a few follow-ups to avoid this kind of problem in future.

Apologies for the noise! 🙉

brrygrdn:
Closing this out as the core PR is just a follow up.

bradzacher:
Just flagging that there are fresh reports of this regressing again

https://twitter.com/NicoloRibaudo/status/1567065351522848768

nicolo-ribaudo:
Thank you @bradzacher!

This is the spammed PR: https://github.com/tc39/ecma262/pull/2819

jeffwidman:
☝️  was tracked in https://github.com/dependabot/dependabot-core/issues/5639

PR #5643, opened 2022-09-06
Remove links to reduce timeline noise

landongrindheim:
### 🖼️ Context

In an attempt to reduce noise in timelines, I'm removing links to PRs opened by Dependabot. I've opted to leave non-bot contributors for now.

I also inserted a zero-width space between `@` and `handle` in the release notes so as to limit notification noise. If this is unwelcome, please let me know 🙂 

Closes https://github.com/dependabot/dependabot-core/issues/5639


deivid-rodriguez:
@landongrindheim Not sure if this is helpful, but this is the original report with links with the fix that was implemented at the time: https://github.com/dependabot/dependabot-core/issues/3276.

deivid-rodriguez:
It seems that we have regressed somewhere but I have no idea where or how. This PR seems like a workaround for now to stop the bleeding? I mean we always showed links, and GitHub didn't create cross references until now. With this solution, we would lose the links?

landongrindheim:
Thanks @deivid-rodriguez! I misread this report. Looking closer, it appears that we started spamming yesterday (Sept 5th):
image

That should help track this down.

jeffwidman:
Could this be at all related to internal work on the monolith or api back end? I know some of the team was recently touching code around PR generation there... perhaps a side effect?

bakkot:
> Looking closer, it appears that we started spamming yesterday (Sept 5th):

If you're referring to https://github.com/tc39/ecma262/pull/2819, it's not that something in Dependabot changed at that time; rather, that's when [the babel release which mentioned said PR went out](https://github.com/babel/babel/releases/tag/v7.19.0).

PR #5645, opened 2022-09-07
Add more helpful error messaging when a vulnerable dependency cannot be upgraded

Nishnha:
This PR fixes the TypeError that was present in https://github.com/dependabot/dependabot-core/pull/5542 and re-introduces the commits from the original PR since they were reverted.


npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker_spec.rb
```diff
@@ -1491,6 +1491,180 @@
...
+                             "@dependabot-fixtures/npm-intermediate-dependency@0.0.1",
+            "name" => "@dependabot-fixtures/npm-intermediate-dependency",
+            "requirement" => "1.0.0",
+            "version" => "0.0.1"
+          )
+
+        expect(conflicting_dependencies_result.last).
+          to eq(
+            "dependency_name" => "@dependabot-fixtures/npm-transitive-dependency",
+            "explanation" => "No patched version available for @dependabot-fixtures/npm-transitive-dependency",
```
mctofu, line 1606: What would the explanation be if it's one of the parents that doesn't have an update to a non locking version? I think that'd be the primary use case of this error as we might not get here if the vulnerable dependency doesn't have a fix available.

bdragon (approved):

npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker_spec.rb
```diff
@@ -1491,6 +1491,180 @@
...
+            "fix_available" => false,
+            "fix_updates" => [],
+            "top_level_ancestors" => []
+          )
+      end
+    end
+
+    context "with an unsatisfiable vulnerablity" do
+      let(:dependency_files) { project_dependency_files("npm8/locked_transitive_dependency") }
```
Nishnha, line 1615: This should be `npm8/transitive_dependency_unlocked`.

I will adjust the test so that the parent dependency can't update to a non-locking version so then the transitive dependency can't be updated to a non-vulnerable version.

npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker_spec.rb
Nishnha: I think that's captured by this test in the vulnerability auditor https://github.com/dependabot/dependabot-core/pull/5645/files#diff-f0c4f0847c642ca1ba8627d73ce6dfc76f107816ac9916ef12d3ae5c4edce665R122-R149

The error message is the same though. Would you expect the message to list out the locking parent?

npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker_spec.rb
```diff
@@ -1491,6 +1491,165 @@
...
+        expect(conflicting_dependencies_result.first).
+          to eq(
+            "dependency_name" => "@dependabot-fixtures/npm-transitive-dependency",
+            "explanation" => "No patched version available for @dependabot-fixtures/npm-transitive-dependency",
+            "fix_available" => false,
+            "fix_updates" => [],
+            "top_level_ancestors" => []
+          )
+      end
+    end
```
Nishnha, line 1650: For the record, I'm not sure why this test fails in the Vulnerability Auditor.

The project, `npm8/transitive-dependency-unlocked` requires `npm-parent-dependency@^2.0.0` so the dependency chain should be 

`npm-parent-dependency@2.0.2` --> `npm-intermediate-dependency@0.0.2` --> `npm-transitive-dependency@1.0.1`

With `npm-transitive-depenency` updated to v1.0.1 it should no longer be vulnerable 🤷🏾

Nishnha: Turns out this is an issue with the lockfiles in `npm8/transitive-dependency-unlocked`

npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker_spec.rb
Nishnha: I removed the test for "with an unsatisfiable vulnerability" because I don't think we can reach that code without also having conflicting dependencies.

I updated the test with both "conflicting dependencies and an unsatisfiable vulnerability" to use an invalid version range for the vulnerability in https://github.com/dependabot/dependabot-core/pull/5645/commits/a8d6b3a1f28a5ea5c1759e01ad22737bfe73024c. 

Our @dependabot-fixtures setup is not complex enough to test this scenario using valid vulnerability ranges, but it does seem like a case that can occur in a project, so I want to keep test coverage around it.

npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker_spec.rb
Nishnha: updated in https://github.com/dependabot/dependabot-core/pull/5645/commits/a23a3efab927f80d7cc6478a59464e29c66f7197

npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker_spec.rb
Nishnha: I opened https://github.com/dependabot/dependabot-core/pull/5672 to create and test a scenario where a valid vulnerable version range is blocked by a locking parent dependency

PR #5646, opened 2022-09-07
Adding code tags around any nwo#number text string

pavera:
Attempting to improve our link sanitation to prevent `<org>/<repo>#<pr number>` text strings from being hydrated into full github.com links by PR creation machinery.

I found that a number of our tests produce redirect links, however the link text itself is in the format above, and adding those links to a PR comment or description results in a valid github.com link and creates a mention on the targeted PR timeline.

This should also break up any text string in the above format to prevent flood of mentions experienced in the below issue:
Closes https://github.com/dependabot/dependabot-core/issues/5639

I also added common tests to the test debug dropdown in vscode and fixed a path issue with the vscode debugger.

~~This is still a WIP as there is 1 failing test which I need to fix, and the approach may be overly heavy handed.  I'm not sure my tests which resulted in lots of valid links were valid, so feedback there would be good.~~

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -20,7 +20,7 @@
       "name": "Debug Tests",
       "type": "Ruby",
       "request": "launch",
-      "program": "${workspaceRoot}/${input:ecosystem}/.bundle/bin/rspec",
+      "program": "${workspaceRoot}/omnibus/.bundle/bin/rspec",
       "cwd": "${workspaceRoot}/${input:ecosystem}",
       "useBundler": true,
       "args": ["${input:test_path}"],
@@ -106,6 +106,7 @@
       "options": [
         "bundler",
         "cargo",
+        "common",
         "composer",
         "docker",
         "elm",
--- a/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
@@ -42,7 +42,7 @@ module Dependabot
           sanitize_mentions(doc)
           sanitize_links(doc)
           sanitize_nwo_text(doc)
-          
+
           mode = unsafe ? :UNSAFE : :DEFAULT
           doc.to_html(([mode] + COMMONMARKER_OPTIONS), COMMONMARKER_EXTENSIONS)
         end
--- a/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
@@ -107,7 +107,8 @@ module Dependabot
                   node.string_content.match?(GITHUB_REF_REGEX)
               node.string_content = replace_github_host(node.string_content)
             elsif node.type == :text &&
-                  node.string_content.match?(GITHUB_NWO_REGEX)
+                  node.string_content.match?(GITHUB_NWO_REGEX) &&
+                  !parent_node_link?(node)
               match = node.string_content.match(GITHUB_NWO_REGEX)
               repo = match.named_captures.fetch("repo")
               number = match.named_captures.fetch("number")
--- a/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
@@ -41,7 +41,8 @@ module Dependabot
           sanitize_team_mentions(doc)
           sanitize_mentions(doc)
           sanitize_links(doc)
-
+          sanitize_nwo_text(doc)
+          
           mode = unsafe ? :UNSAFE : :DEFAULT
           doc.to_html(([mode] + COMMONMARKER_OPTIONS), COMMONMARKER_EXTENSIONS)
         end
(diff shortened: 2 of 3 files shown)


pavera:
I've now limited this to only find text strings and not link text.  I am concerned as copy/pasting a link such as: https://github.com/dependabot/dependabot-core/blob/main/common/spec/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer_spec.rb#L278

results in:
<p><a href=\"https://github-redirect.com/rust-num/num-traits/" \
          "pull/144\">rust-num/num-traits#144</a></p>\n

I feel like we should probably be escaping the link text as well.

honeyankit (approved):

common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
```diff
@@ -14,6 +14,7 @@ class LinkAndMentionSanitizer
           github\.com/(?<repo>#{GITHUB_USERNAME}/[^/\s]+)/
           (?:issue|pull)s?/(?<number>\d+)
         }x.freeze
+        GITHUB_NWO_REGEX = %r{(?<repo>#{GITHUB_USERNAME}/[^/\s#]+)#(?<number>\d+)}.freeze
```
Nishnha, line 17: Could you explain this regex?

common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
```diff
@@ -109,6 +111,25 @@ def sanitize_links(doc)
...
+              replace_nwo_node(node)
+            end
+          end
+        end
+
+        def replace_nwo_node(node)
+          match = node.string_content.match(GITHUB_NWO_REGEX)
+          repo = match.named_captures.fetch("repo")
+          number = match.named_captures.fetch("number")
+          new_node = build_nwo_text_node("#{repo}##{number}")
```
Nishnha, line 128: Does this still just render as `NWO#number`?

Nishnha (approved):

common/spec/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer_spec.rb
```diff
@@ -281,6 +281,17 @@
       end
     end
 
+    context "with a GitHub NWO and PR number" do
+      let(:text) do
+        "dsp-testing/dependabot-ts-definitely-typed#25"
+      end
+      it do
+        is_expected.to eq(
+          "<p><code>dsp-testing/dependabot-ts-definitely-typed#25</code></p>\n"
```
Nishnha, line 290: Do we need to include the newline at the end of the line?

common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
pavera: I cribbed it together from the above regex that grabs the repo and number from a full github pr or issue link.  Given a string like `myorg/myrepo#123` the regex will match myorg based on the GITHUB_USERNAME regex, then will match nonwhitespace and non # chars after a / as the repo name, then grab the digits after the # as the issue or PR number.

common/spec/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer_spec.rb
pavera: the [call here](https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb#L45) renders the doc and adds the newline, so the test check has to expect it as far as I know.

common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
honeyankit: @pavera : could you put the regex explanation in the comment as well. It will be easier for future debugging.

mattt: I'd recommend a positive lookahead instead of matching ("not character class" then "character class") 

```suggestion
        GITHUB_NWO_REGEX = %r{(?<repo>#{GITHUB_USERNAME}/.+)(?=#(?<number>\d+))}.freeze
```

deivid-rodriguez (approved):
Looks good to me!

What does nwo stand for?

pavera:
> What does nwo stand for?
"name with owner" shorthand for `org/repo`

jakecoffman (approved):
Seems like it should work, and it didn't break any of the smoke tests. Nicely done!

mattt (approved):

.vscode/launch.json
```diff
@@ -20,7 +20,7 @@
       "name": "Debug Tests",
       "type": "Ruby",
       "request": "launch",
-      "program": "${workspaceRoot}/${input:ecosystem}/.bundle/bin/rspec",
+      "program": "${workspaceRoot}/omnibus/.bundle/bin/rspec",
```
jeffwidman, line 23: What was the rationale for this? I don't see it mentioned in the commit messages / PR description, so was surprised to see it... did I just miss it somehow?

pavera: This change was necessitated by https://github.com/dependabot/dependabot-core/pull/5529

PR #5647, opened 2022-09-07
Fix incomplete clean up of odd python requirements

deivid-rodriguez:
In 45f5b77a776cf8bada4c262af9e39ff492ab6ba9, we started handling odd python requirements, with multiple constraints not separated by colons.

Native behavior is to ignore anything after the first requirement, yet we were only ignoring the first one.

### DIFF
--- a/python/spec/dependabot/python/requirement_spec.rb
+++ b/python/spec/dependabot/python/requirement_spec.rb
@@ -116,6 +116,17 @@ RSpec.describe Dependabot::Python::Requirement do
       end
     end
 
+    context "with multiple operators after the first" do
+      let(:requirement_string) { ">=2.0<2.1<2.2" }
+      # Python ignores operators after the first!
+      it { is_expected.to eq(Gem::Requirement.new(">=2.0")) }
+
+      context "separated with a comma" do
+        let(:requirement_string) { ">=2.0,<2.1,<2.2" }
+        it { is_expected.to eq(Gem::Requirement.new(">=2.0", "<2.1", "<2.2")) }
+      end
+    end
+
     context "with an array" do
       let(:requirement_string) { ["== 1.3.*", ">= 1.3.1"] }
       its(:to_s) do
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Aha! I had recently learnt that requirement format for Poetry is different than the standard, so I had assumed that we were not parsing those in the same way. But maybe we are? I will review this!

pavera (approved):
🚀

python/spec/dependabot/python/requirement_spec.rb
```diff
@@ -116,6 +116,17 @@
       end
     end
 
+    context "with multiple operators after the first" do
+      let(:requirement_string) { ">=2.0<2.1<2.2" }
+      # Python ignores operators after the first!
+      it { is_expected.to eq(Gem::Requirement.new(">=2.0")) }
```
jeffwidman, line 122: Curious why you added a new test case rather than modifying the existing test case from https://github.com/dependabot/dependabot-core/commit/45f5b77a776cf8bada4c262af9e39ff492ab6ba9#diff-606e746adf06604fa9039dd903280cffd0c3ee1485e2d54ad5e13932279a019aR40-R44 ?

Is there a scenario you envision where it's useful to test 2 operators that wouldn't be caught by testing 3 operators?

deivid-rodriguez: Normally I prefer adding separate specs over modifying existing ones, mostly because I don't want to lose any coverage and many times I don't feel I can tell for sure whether that would be the case if I modify the spec. I also think it makes patches more focused, although it normally introduces some duplication and overhead. In this case it would probably be fine to change the existing spec, although it would be a bit weird to only test three operators instead of two? I think testing both cases illustrates that there's a bit more complexity when handling more than two.

jeffwidman: Sounds reasonable.

PR #5648, opened 2022-09-07
move devcontainer to allow debugging updater with the default devcontainer

jakecoffman:
Now that updater is in core, there's no way to use Codespaces to run the updater tests since the devcontainer.json we specified drops us in docker-dev-shell where updater doesn't exist. 

So I've moved devcontainer.json to a subdirectory which gives us the ability to choose between the devcontainer.json we specified and the default one Codespaces uses:

Screen Shot 2022-09-07 at 9 29 19 AM

Bonus: I noticed there was no cacheFrom specified, so I've added one and it seems to be using it when building. This should save some time!

### DIFF
--- a/.devcontainer/core-dev/devcontainer.json
+++ b/.devcontainer/core-dev/devcontainer.json
@@ -1,6 +1,9 @@
 {
-  "name": "Ruby",
-  "dockerFile": "../Dockerfile.development",
+  "name": "core-dev",
+  "build": {
+    "dockerfile": "../../Dockerfile.development",
+    "cacheFrom": "ghcr.io/dependabot/dependabot-core-development"
+  },
 
   "workspaceFolder": "/home/dependabot/dependabot-core",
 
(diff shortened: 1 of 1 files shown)


.devcontainer/core-dev/devcontainer.json
```diff
@@ -120,11 +123,11 @@
...
-		"misogi.ruby-rubocop",
-		"groksrc.ruby",
-		"hoovercj.ruby-linter",
-		"miguel-savignano.ruby-symbols"
+    "rebornix.ruby",
+    "castwide.solargraph",
+    "misogi.ruby-rubocop",
+    "groksrc.ruby",
+    "hoovercj.ruby-linter",
+    "miguel-savignano.ruby-symbols"
```
jakecoffman, line 131: I guess my editor didn't like this spacing.

jakecoffman: It's enforcing the .editorconfig, so I think this is ok.

Nishnha (approved):
Nice!

PR #5649, opened 2022-09-07
fix tagged push overwriting previous tags

jakecoffman:
I noticed that [in GHCR](https://github.com/dependabot/dependabot-core/pkgs/container/dependabot-core/versions) we're tagging the latest tagged version 0.212.0 repeatedly,  I think this is because this if statement is a string:

```workflow
if: "contains(github.ref, 'refs/tags')"
```

I also changed the updater push to not push latest since that's being done in `docker-branch-releases.yml` now. I moved the if statement up on that job so it only runs to tag if the version changes.


Nishnha (approved):
Seems correct to me

PR #5650, opened 2022-09-07
deploy commits made after approval

jakecoffman:
A downside to #5640 was commits pushed after approval aren't deployed.

So to fix that, I've restored the pull_request trigger, but use `gh` to get the review state. Thus if a commit comes in after approval it will still be able to see if was approved. Commits before an approval should be skipped.

### DIFF
--- a/.github/workflows/docker-branch-releases.yml
+++ b/.github/workflows/docker-branch-releases.yml
@@ -41,7 +41,10 @@ jobs:
         # when not on main, sets DECISION to the PR's review decision, handling the push-after-approval case
       - name: Check if pull request is approved
         if: github.ref != 'refs/heads/main'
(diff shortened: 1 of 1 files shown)


jakecoffman:
Hold off on approving so I can test this. Also I seem to have broken the workflow.

jakecoffman:
Great, it looks like it's working now: 

```
Review decision is: REVIEW_REQUIRED
```

pavera (approved):
Test away :)

jakecoffman:
The approval trigger worked: https://github.com/dependabot/dependabot-core/runs/8236295929?check_suite_focus=true

jakecoffman:
Update trigger worked too: https://github.com/dependabot/dependabot-core/runs/8236356928?check_suite_focus=true

Only thing left is to test on main 🤞

PR #5657, opened 2022-09-08
Fix updater specs on M1

deivid-rodriguez:
I had not been able to get all updater specs passing on my laptop. I was getting two spec failures, which I describe in each commit message, together with how I got them passing.


pavera (approved):
🚀

landongrindheim (approved):
🙌 Thanks for the context-filled commit messages 😍

PR #5658, opened 2022-09-08
Speed up dealing with non-reachable git repos

deivid-rodriguez:
While working on a spec failure, I observed the failing spec was also soooooo slow:

```
Top 1 slowest examples (48.37 seconds, 100.0% of total time):
  Dependabot::EndToEndJob bundler git dependencies updates dependencies correctly
    48.37 seconds ./spec/dependabot/integration_spec.rb:296
```

Further investigation revealed that this spec was trying to do the following up to 76 times, on a non-existing github repository:

* Try fetch `https://github.com/dependabot/dummy-git-dependency.git/info/refs?service=git-upload-pack`.
* Since the previous fetch failed, try run `git ls-remote https://github.com/dependabot/dummy-git-dependency.git` locally.

This is done every time the update checker needs to fetch remote git tags, and if this fails, the checker considers that there are no remote tags, but will keep on retrying the same thing and failing once and again.

This commit changes the commit fetcher to memoize the tags the first time they are fetched. With this change, the spec is now much faster:

```
Top 1 slowest examples (4.26 seconds, 100.0% of total time):
  Dependabot::EndToEndJob bundler git dependencies updates dependencies correctly
    4.26 seconds ./spec/dependabot/integration_spec.rb:296
```

I expect this to also improve things in production, of course.

### DIFF
--- a/common/lib/dependabot/git_commit_checker.rb
+++ b/common/lib/dependabot/git_commit_checker.rb
@@ -365,17 +365,19 @@ module Dependabot
     def listing_tags
       return [] unless listing_source_url
 
-      tags = listing_repo_git_metadata_fetcher.tags
+      @listing_tags ||= begin
+        tags = listing_repo_git_metadata_fetcher.tags
 
-      if dependency_source_details&.fetch(:ref, nil)&.start_with?("tags/")
-        tags = tags.map do |tag|
-          tag.dup.tap { |t| t.name = "tags/#{tag.name}" }
+        if dependency_source_details&.fetch(:ref, nil)&.start_with?("tags/")
+          tags = tags.map do |tag|
+            tag.dup.tap { |t| t.name = "tags/#{tag.name}" }
+          end
         end
-      end
 
-      tags
-    rescue GitDependenciesNotReachable
-      []
+        tags
+      rescue GitDependenciesNotReachable
+        []
+      end
     end
 
     def listing_upload_pack


landongrindheim (approved):
I think this makes a lot of sense! I'm assuming we're okay with not picking up the (likely) small number of cases where a tag is found on retry.

common/lib/dependabot/git_commit_checker.rb
```diff
@@ -365,17 +365,19 @@ def listing_tag_for_version(version)
     def listing_tags
       return [] unless listing_source_url
 
-      tags = listing_repo_git_metadata_fetcher.tags
+      @listing_tags ||= begin
```
pavera, line 368: I believe we've been swapping `||=` memoization to using `if defined?` to handle valid nil values.

deivid-rodriguez:
My understanding is that `GitDependenciesNotReachable` would be a non-retryable error, but we could make the logic more fine grained to make sure: let tag fetcher raise the original exception if its a `RepoNotFound` error, and memoize only in that case.

However, I'm not sure what the result will be if the update checker gets inconsistent results about the available remote tags 😬.

common/lib/dependabot/git_commit_checker.rb
deivid-rodriguez: Hehe, @mattt's cop would've caught me here 😅. I did check that this should always return an array, so I think standard memoization is safe. But I can add a `|| []` if wanted just in case, or a `defined?` check.

landongrindheim: I think in this case we'll always be dealing with a collection, though that collection may be empty. I'm not sure about our conventions for non-nil value memoization.

deivid-rodriguez: I think standard memoization should play nice with empty arrays. Here I followed the style in `git_commit_checker.rb` where several collections are memoized in the standard way.

landongrindheim: Sorry if my comment added confusion. I was responding at the same time as you @deivid-rodriguez and your comment didn't render before I clicked "comment".

I agree with you that we should be fine to use standard memoization (without `|| []`)

pavera (approved):

PR #5660, opened 2022-09-08
Initial yarn berry support

jurre:
This adds support for yarn-berry behind a feature flag, meaning that it won't run on repo's by default, but we can start using it to test targeted repo's.

There are a few notable known limitations:
- Support for private and non-default registries is still WIP
- Setups that rely on Git LFS are not supported
- Repo's with existing peer-dependency issues are not handled gracefully


npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -64,6 +65,22 @@ def instrument_package_manager_version
...
 
+      def yarn_version
+        @yarn_version ||= begin
+          package = JSON.parse(package_json.content)
+          if (pkgmanager = package.fetch("packageManager", nil))
+            @yarn_version = get_yarn_version_from_path(pkgmanager)
+          elsif yarn_lock
+            @yarn_version = 1
+          end
+        end
```
landongrindheim, line 76: Noting that this memoization pattern is prone to causing re-evaluation since `if`/`elseif` will fall back to `nil`. We could fall back to `return @yarn_version if defined?(@yarn_version)`.

Also noting that we don't need the `@yarn_version` assignments inside the `if` block.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -64,6 +65,22 @@ def instrument_package_manager_version
...
+      end
+
+      def get_yarn_version_from_path(path)
+        version_match = path.match(/yarn@(?<version>\d+.\d+.\d+)/)
+        version_match.named_captures.fetch("version")
```
landongrindheim, line 81: ❓ What kinds of guarantees do we have about `path` here? I can foresee some `NoMethodError`s being thrown if we don't have a full match, and a `KeyError` if the named captures aren't properly saturated.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
```diff
@@ -188,7 +190,11 @@ def semver_version_for(version_string)
         end
 
         def alias_package?(requirement)
-          requirement.include?("@npm:")
+          Experiments.enabled?(:yarn_berry) ? requirement.match?(/@npm:(.+@(?!npm))/) : requirement.include?("@npm:")
```
landongrindheim, line 193: This is really a matter of personal preference, but I find this line a little harder to parse than normal due to the Regexp syntax combining with the ternary syntax. What do you think of 👇?

```suggestion
          if Experiments.enabled?(:yarn_berry)
            requirement.match?(/@npm:(.+@(?!npm))/)
          else  
            requirement.include?("@npm:")
          end
```

jurre:
Thanks @brrygrdn / @landongrindheim, I've addressed those issues I think 🙇

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
```diff
@@ -190,7 +190,11 @@ def semver_version_for(version_string)
         end
 
         def alias_package?(requirement)
-          Experiments.enabled?(:yarn_berry) ? requirement.match?(/@npm:(.+@(?!npm))/) : requirement.include?("@npm:")
+          if Experiments.enabled?(:yarn_berry)
+            requirement.match?(/@npm:(.+@(?!npm))/)
+          else
```
landongrindheim, line 195: Sorry, I had a space in my suggestion and it's making Rubocop :rage4: 
```suggestion
          else
```

landongrindheim (approved):
I'd love to see another 👍 from someone with better JS chops. Given the assurance that this is behind a feature flag, I'm happy to sign off 🚀

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -53,8 +54,8 @@ def fetch_files
       def instrument_package_manager_version
         package_managers = {}
 
-        package_managers["npm"] =  Helpers.npm_version_numeric(package_lock.content) if package_lock
-        package_managers["yarn"] = 1 if yarn_lock
+        package_managers["npm"] = Helpers.npm_version_numeric(package_lock.content) if package_lock
+        package_managers["yarn"] = yarn_version if yarn_version
```
brrygrdn, line 58: ❓ Should this change to yarn detection be moved behind the experiment flag?

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater.rb
```diff
@@ -53,11 +55,56 @@ def updated_dependency_files
           )
         end
 
+        if Experiments.enabled?(:yarn_berry)
+          # There's probably a way to avoid this duplication between the
+          # vendor_updater and install_state_updater, but yeah.
```
brrygrdn, line 60:
```suggestion
          # There's probably a way to avoid this duplication between the
          # vendor_updater and install_state_updater, but yeah.
```

😄 FWIW, I think this is fine - I always feel bad when I can't figure out some Ruby golf to compact a statement but this is clear and works so 👍🏻

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
brrygrdn: Oh nevermind, this is an instrumentation method so I'm less worried about this

brrygrdn (approved):
Looks good to me 💯

Issue #3290, opened 2021-03-17
Standard Python support


flying-sheep:
There’s exactly two standard ways of specifying dependencies, none of which dependabot supports.

1. [PEP 517’s `prepare_metadata_for_build_wheel()`](https://www.python.org/dev/peps/pep-0517/#prepare-metadata-for-build-wheel).

   The legacy way to specify Python dependencies is `requirements.txt`. The standard way is having a PEP 517 build system configured. That build system can optionally define a function called `prepare_metadata_for_build_wheel`. If that function exists, the dependencies can be obtained (from the metadata) without building a wheel, otherwise `build_wheel()` needs to be called.

   Code to just do the above would look like:

   ```py
   import build
   deps = build.ProjectBuilder("path/to/project").get_dependencies("sdist")
   # or
   deps = build.ProjectBuilder("path/to/project").get_dependencies("wheel") - {"wheel"}
   ```

2. [PEP 631](https://www.python.org/dev/peps/pep-0631/) a new standardized way to specify dependencies in a Python project (now merged into [PEP 621](https://www.python.org/dev/peps/pep-0621/)).

   It’s supported by [`flit`](https://flit.readthedocs.io/en/latest/) and a bunch of other new build backends. So if you want to continue your (incomplete) way of just parsing files and not calling into Python, this needs to be supported.

flying-sheep:
This should be pretty straightforward (if one knows ruby, which I don’t). You already have

1. a format preserving TOML parser (for poetry support)
2. a parser and serializer for the [PEP 508](https://www.python.org/dev/peps/pep-0508/) dependency specifier language (for requirements.txt support)

So you just have to read the TOML tables `project.dependencies` and `project.optional-dependencies.*` instead of `tool.poetry.dependencies` and `tool.poetry.extras`, and then use the PEP 508 handling logic instead of the poetry version specifier logic

dreinon:
I just got a dependabot PR of my project using pdm! Hope this has been finally accepted :slightly_smiling_face:

drewcassidy:
now that setuptools supports PEP621, It would be nice to see this

drewcassidy:
> So you just have to read the TOML lists `project.dependencies` and `project.optional-dependencies.*` instead of the toml tables `tool.poetry.dependencies` and `tool.poetry.extras.*`, and then use the PEP 508 handling logic instead of the poetry version specifier logic

[...]

PR #5661, opened 2022-09-08
Initial work on standard Python support

deivid-rodriguez:
This adds initial support for `pyproject.toml` files. 

Dependabot will be able to parse `[project]` `dependencies` and provide version and security updates. ~, but will not yet be able to parse `[project.optional-dependencies]`.~ EDIT: We also added support for updating dependencies under `[project.optional-dependencies]`.

Since poetry can also uses pyproject.toml files, this also support parsing poetry dependencies. Using poetry's version constraint syntax was added in https://github.com/dependabot/dependabot-core/pull/5735

<details>
<summary>Here's an example PR generated with the `standard_python.toml` file
</summary>
</br>

Pull Request Title: Bump ansys-templates from 0.3.0 to 0.3.1

Bumps [ansys-templates](https://github.com/pyansys/pyansys-template) from 0.3.0 to 0.3.1.
<details>
<summary>Commits</summary>
<ul>
<li><a href="https://github.com/ansys/ansys-templates/commit/5175dd82520a2af0cf316fd31441c18ea69c0edf"><code>5175dd8</code></a> Bump 0.3.1 version</li>
<li>See full diff in <a href="https://github.com/pyansys/pyansys-template/compare/v0.3.0...v0.3.1">compare view</a></li>
</ul>
</details>
<br />

</details>

```
=> parsing dependency files
=> updating 1 dependencies: ansys-templates

=== ansys-templates (0.3.0)
 => checking for updates 1/1
🌍 https://pypi.org/simple/ansys-templates/
 => latest available version is 0.3.1
🌍 https://pypi.org/simple/ansys-templates/
 => latest allowed version is 0.3.1
 => updating ansys-templates from 0.3.0 to 0.3.1

    ± pyproject.toml
    ~~~
    12c12
    <    "ansys-templates==0.3.0",
    ---
    >    "ansys-templates==0.3.1",
    ~~~
🌍 https://pypi.org/pypi/ansys-templates/json
🌍 https://github.com/pyansys/pyansys-template.git/info/refs
🌍 https://github.com/pyansys/pyansys-template.git/info/refs
```


python/lib/dependabot/python/file_parser/pyproject_files_parser.rb
```diff
@@ -59,6 +67,38 @@ def pyproject_dependencies
           dependencies
         end
 
+        def pep621_dependencies
+          dependencies = Dependabot::FileParsers::Base::DependencySet.new
+
+          parsed_pep621_dependencies.each do |dep|
+            # If a requirement has a `<` or `<=` marker then updating it is
+            # probably blocked. Ignore it.
+            next if dep["markers"].include?("<")
```
deivid-rodriguez, line 76: Not fully sure what to do with markers. If I understand correctly, this is what we currently do:

* For requirements.txt: https://github.com/dependabot/dependabot-core/blob/7b6f72eddeee3a4cf43ef61e13782b855010bcc0/python/lib/dependabot/python/file_parser.rb#L120-L126
* For setup.py: https://github.com/dependabot/dependabot-core/blob/7b6f72eddeee3a4cf43ef61e13782b855010bcc0/python/lib/dependabot/python/file_parser/setup_file_parser.rb#L30-L32
* For poetry I think we don't ignore them but maybe because we delegate to poetry itself which does the right thing.

I guess it makes sense to ignore marked dependencies in a similar way here too.

Nishnha: > If a requirement has a `<`, `<=` or '==' marker then updating it is probably blocked. Ignore it.

I'm surprised we just ignore these dependencies for Python. For npm we check whether we have conflicting dependencies by passing them into the same library that the CLI uses https://github.com/dependabot/dependabot-core/blob/main/npm_and_yarn/helpers/lib/npm/conflicting-dependency-parser.js

Fixing this in Python is probably outside of the scope of this PR, but I wonder how many updates are blocked because we ignore certain dependency requirements

python/lib/dependabot/python/file_parser/pyproject_files_parser.rb
```diff
@@ -75,6 +115,14 @@ def parse_requirements_from(req, type)
           end
         end
 
+        def using_poetry?
+          !parsed_pyproject.dig("tool", "poetry").nil?
```
Nishnha, line 119: nice!

python/lib/dependabot/python/file_parser/pyproject_files_parser.rb
deivid-rodriguez: Note that these are not requirements though, these are something Python calls ["environment markers"](https://peps.python.org/pep-0496/), which set additional constraints on a dependency in addition to version, like platform or python version. The idea (I believe) is that if there's a marker explicitly holding the dependency is that the user probably don't want an update because it won't work? Not sure.

Nishnha (approved):

python/lib/dependabot/python/file_fetcher.rb
```diff
@@ -24,7 +24,7 @@ def self.required_files_in?(filenames)
         # If this repo is using a Pipfile return true
         return true if filenames.include?("Pipfile")
 
-        # If this repo is using Poetry return true
+        # If this repo is using pryproject.toml return true
```
mctofu, line 27:
```suggestion
        # If this repo is using pyproject.toml return true
```

python/spec/dependabot/python/file_parser_spec.rb
```diff
@@ -1212,18 +1212,18 @@
...
     end
 
-    context "with a pyproject.toml and pyproject.lock" do
+    context "with a pyproject.toml in poetry format and pyproject.lock legacy poetry lock file" do
       let(:files) { [pyproject, pyproject_lock] }
       let(:pyproject) do
         Dependabot::DependencyFile.new(
           name: "pyproject.toml",
-          content: fixture("pyproject_files", "pyproject.toml")
+          content: fixture("pyproject_files", "poetry.toml")
```
mctofu, line 1220: Does the file name matter here? It's looks a little strange to load content from `poetry.toml` here but have the DependacyFile name `pyproject.toml` in this test and a few others.

mctofu (commented):
I know very little about Poetry or pep621 but I gave this a first pass. Nice job keeping these changes to a minimum!

Is there any test coverage for doing a standard python update? I see from the PR description that it can perform updates but I'm not seeing an added test case for the updater. It'd be great to have one (or more) if it's not there!

Also, is there anything else missing besides the optional-dependencies? Would we want to feature flag this till we have more complete or tested support?

deivid-rodriguez:
Thanks for having a look @mctofu!

> Is there any test coverage for doing a standard python update? I see from the PR description that it can perform updates but I'm not seeing an added test case for the updater. It'd be great to have one (or more) if it's not there!

That's the one task I had pending before considering this PR finished, I will add those soon :)

> Also, is there anything else missing besides the optional-dependencies? Would we want to feature flag this till we have more complete or tested support?

We could also potentially support updating build dependencies. See [this comment](https://github.com/dependabot/dependabot-core/issues/3290#issuecomment-1237306729) for useful links. But that sounds much trickier and I'm not sure how many people would find it useful, so I would not add support for now. Regarding feature flagging this, I'm not sure. It seems a simple update but it's true that Dependabot will start considering many files it didn't consider before so maybe it makes sense, yeah.

python/lib/dependabot/python/update_checker.rb
```diff
@@ -261,22 +269,23 @@ def latest_version_finder
         )
       end
 
+      def poetry_based?
+        pyproject && !poetry_details.nil?
+      end
+
       def poetry_library?
```
deivid-rodriguez, line 276: I've been adding a bunch of specs here and I think I ended up finding a bug. We only detect libraries in the case of poetry, so if a Python library is using standard python and has a requirement in `pyproject.toml` like this:

```toml
dependencies = [
  "requests~=1.0.0"
]
```

Then with this PR dependabot will propose an update like this:

```toml
dependencies = [
  "requests~=2.19.1"
]
```

instead of widening the requirement, like we do for Poetry

```toml
dependencies = [
  "requests>=1,<3"
]
```

I'll push all the new specs, including a failing one for this, and will try to wrap this up tomorrow.

deivid-rodriguez:
I'm finding a few more things missing, so I'll recap those here. I believe all of them are suitable for follow up work, and not necessarily need to block this PR.

* Optional dependencies.
* Build dependencies.
* Dependencies aliases. As per https://peps.python.org/pep-0621/#dependencies-optional-dependencies, different package managers use different aliases for the `project.dependencies` field, like `requires` for flit, `install_requires` for setuptools.

davidism:
It seems like the focus is on `pyproject.toml` itself, but bumping versions there might not make sense in many cases.

For example, pip-tools separates `requirements.in`, the template listing of direct dependencies and broad constraints, from `requirements.txt` that is the full dependency tree with exact versions. Dependabot updates the pins, not the template.

For pyproject.toml based tools, I'd expect the same thing: the pyproject file is the template with constraints, and tools generate a lock file with specific versions that I'd want bumped. For example, PDM generates `pdm.lock`, another TOML file. Unfortunately [lock files are not standardized](https://discuss.python.org/t/how-should-a-lockfile-pep-665-successor-look-like/17690), but it sounds like you're already considering multiple behaviors for multiple tools. You can

python/lib/dependabot/python/update_checker.rb
deivid-rodriguez: Fixed by https://github.com/dependabot/dependabot-core/pull/5661/commits/ea1ad25ef843b0d641ac931e78c2dafab754a8bc!

python/lib/dependabot/python/file_parser/pyproject_files_parser.rb
deivid-rodriguez: We'll be leaving this as is for now, marking as solved.

deivid-rodriguez:
@davidism The behavior you're suggesting is implemented by the [`lockfile-only` versioning strategy](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy), but of course that only makes sense if there's a lock file. In the case of Standard Python, lockfiles are not defined as you point out, so we only focus on the manifest and not on the lock file. Hope this makes sense!

I'm now satisfied with this PR, by the way! 😄

davidism:
I'm worried that only supporting pins in `pyproject.toml`, rather than waiting for lockfiles, will encourage users to put pins there. That's probably fine for applications, but would be disastrous if libraries started pinning their dependencies like that. Please reconsider supporting some common lock file formats, since you already have custom support for Poetry.

deivid-rodriguez:
> I'm worried that only supporting pins in pyproject.toml, rather than waiting for lockfiles, will encourage users to put pins there. That's probably fine for applications, but would be disastrous if libraries started pinning their dependencies like that. 

Dependabot does not encourage any way of handling dependencies, it respects what users are already doing, and just tries to help them keep things up to date. So it will only use pins in `pryproject.toml` files if they are already using pins. If they are using fully unrestricted dependencies, then dependabot will not propose any updates.

> Please reconsider supporting some common lock file formats, since you already have custom support for Poetry.

This is just a first iteration, we will be happy to add support for lock file formats using standard Python in the future.

deivid-rodriguez:
How will users "see" that we support pins, if they are not already using pins? I'm not following your logic of how Dependabot will encourage a bad thing here.

davidism:
Presumably by seeing the announcement that Dependabot supports pins in pyproject.toml, or reading the Dependabot documentation about what pins it updates.

ofek:
Everyone is excited for this 😄

PR #5665, opened 2022-09-08
fix smoke tests from forks by using public sources

jakecoffman:
Our smoke tests were broken for forks since PATs aren't available there. So I've moved the tests, caches, and CLI binary to public places so we can use GITHUB_TOKEN which is available as a read-only source.


honeyankit (approved):
LGTM!

PR #5668, opened 2022-09-08
deploy from a fork using a workflow

jakecoffman:
Since fork PRs can only get read access from GITHUB_TOKEN, we can't use the `docker-branch-releases.yml` workflow to deploy a fork to GHCR before merging it.

This works around that with a manually dispatched workflow that pulls in the PR code, builds, and deploys it. Since it was kicked off by the team it _should_ have the correct permissions.

> **Note**
> While implementing this, I tried to simply `git merge main` after the `gh pr checkout` but git errors with `fatal: refusing  to merge unrelated histories`. So I assume the `gh` command is pulling in the PR as a different tree. To work around that, I squashed the changes and cherry picked them over to main. If anyone has a better or easier solution to that let me know!

### DIFF
--- a/.github/workflows/docker-fork-releases.yml
+++ b/.github/workflows/docker-fork-releases.yml
@@ -81,6 +81,7 @@ jobs:
 
       - name: Set summary
         run: |
+          echo "generated for PR ${{ github.event.inputs.pr }}" > $GITHUB_STEP_SUMMARY
           echo "updater uploaded with tag \`$TAG\`" >> $GITHUB_STEP_SUMMARY
           echo "\`\`\`" >> $GITHUB_STEP_SUMMARY
           echo "$UPDATER_IMAGE_MIRROR:$TAG" >> $GITHUB_STEP_SUMMARY
--- a/.github/workflows/docker-fork-releases.yml
+++ b/.github/workflows/docker-fork-releases.yml
@@ -1,0 +1,75 @@
+name: Push docker fork images
+env:
+  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  BASE_IMAGE: "ubuntu:20.04"
+  UPDATER_IMAGE: "dependabot/updater"
+  UPDATER_IMAGE_MIRROR: "ghcr.io/dependabot/dependabot-updater"
+on:
+  workflow_dispatch:
+    inputs:
+      pr:
+        required: true
+        type: string
+        description: PR number
+
+jobs:
+  push-fork-image:
+    name: Build and push fork image
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    env:
+      TAG: ${{ github.sha }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+        # The PR must be approved. This is mostly to ensure you've typed the correct PR.
+      - name: Check if pull request is approved
+        run: |
+          DECISION=$(gh pr view ${{ github.event.inputs.pr }} --json reviewDecision -t {{.reviewDecision}})
+          echo "Review decision is: $DECISION"
+          [[ $DECISION == "APPROVED" ]] || exit 1
+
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-fork-releases.yml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Check if pull request is approved
+        # The PR must be approved. This is mostly to ensure you've typed the correct PR.
+        # Note: forks will have a blank review decision without approval. Not NEEDS_REVIEW.
+        run: |
+          DECISION=$(gh pr view ${{ github.event.inputs.pr }} --json reviewDecision -t {{.reviewDecision}})
+          echo "Review decision is: $DECISION"
+          [[ $DECISION == "APPROVED" ]] || exit 1
```
Nishnha, line 34: Does this exit the job or just the step?

.github/workflows/docker-fork-releases.yml
```diff
@@ -0,0 +1,87 @@
...
+    permissions:
+      contents: read
+      packages: write
+    env:
+      TAG: ${{ github.sha }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Check if pull request is approved
```
Nishnha, line 28: Is this step necessary?
see https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks#about-workflow-runs-from-public-forks

> To help prevent this, workflows on pull requests to public repositories from some outside contributors will not run automatically, and might need to be approved first. By default, all first-time contributors require approval to run workflows.

Nishnha: Ah the workflow approval is different from an approval on a PR. Disregard

.github/workflows/docker-fork-releases.yml
```diff
@@ -0,0 +1,87 @@
...
+            .
+
+      - name: Build dependabot-updater image
+        env:
+          DOCKER_BUILDKIT: 1
+        run: |
+          docker build \
+            -t "$UPDATER_IMAGE:$TAG" \
+            --build-arg BUILDKIT_INLINE_CACHE=1 \
+            --cache-from "$BASE_IMAGE" \
```
Nishnha, line 67: cache from ubuntu:20.04 seems a bit strange. Any reason we include this line?

.github/workflows/docker-fork-releases.yml
```diff
@@ -0,0 +1,87 @@
...
+            --cache-from "$UPDATER_IMAGE_MIRROR" \
```
Nishnha, line 68: double checked that this is public at https://github.com/dependabot/dependabot-core/pkgs/container/dependabot-updater

.github/workflows/docker-fork-releases.yml
```diff
@@ -0,0 +1,87 @@
...
+            --build-arg OMNIBUS_VERSION=$TAG \
```
Nishnha, line 69: This really just pulls the tagged GHCR `dependabot/dependabot-core` image in https://github.com/dependabot/dependabot-core/blob/main/Dockerfile.updater#L2

Nishnha (approved):

.github/workflows/docker-fork-releases.yml
jakecoffman: This will fail the workflow.

.github/workflows/docker-fork-releases.yml
jakecoffman: Yeah I agree, I copied this from another workflow.

.github/workflows/docker-fork-releases.yml
jakecoffman: Well, uses the image built in the `Build dependabot-updater image` step above, but yes!

PR #5671, opened 2022-09-09
Include removed dependency flag when creating a pull request

mctofu:
This is required to support removed transitive dependencies.


landongrindheim (approved):

updater/spec/dependabot/api_client_spec.rb
```diff
@@ -101,6 +101,34 @@
             end)
       end
     end
+
+    context "with a removed dependency" do
```
Nishnha, line 105: What happens if somehow we pass in `removed: false`?

mctofu: It defaults to false so should be the same: https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/dependency.rb#L42-L44

Nishnha (approved):
LGTM

PR #5674, opened 2022-09-10
Increase docker registry client timeout

pavera:
DockerRegistry2 has a default read timeout of 5s.  We have seen issues where loading the list of docker tags times out if that list is very large.  This increases the timeout to 10s which in my testing locally is sufficient to load hexpm/elixir with > 150k tags.

### DIFF
--- a/docker/lib/dependabot/docker/update_checker.rb
+++ b/docker/lib/dependabot/docker/update_checker.rb
@@ -347,7 +347,8 @@ module Dependabot
           DockerRegistry2::Registry.new(
             "https://#{registry_hostname}",
             user: registry_credentials&.fetch("username", nil),
-            password: registry_credentials&.fetch("password", nil)
+            password: registry_credentials&.fetch("password", nil),
+            read_timeout: 10,
           )
       end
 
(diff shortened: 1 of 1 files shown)


docker/lib/dependabot/docker/update_checker.rb
```diff
@@ -347,7 +347,8 @@ def docker_registry_client
           DockerRegistry2::Registry.new(
             "https://#{registry_hostname}",
             user: registry_credentials&.fetch("username", nil),
-            password: registry_credentials&.fetch("password", nil)
+            password: registry_credentials&.fetch("password", nil),
+            read_timeout: 10,
```
jurre, line 351:
```suggestion
            read_timeout: 10
```

To appease the RuboCop

jurre (approved):

docker/lib/dependabot/docker/update_checker.rb
jeffwidman: 🚲 shed: 
I'd also be open to disabling this rubocop instead... I'm personally a fan of trailing commas because it preserves `git blame` history/smaller diffs if someone later appends another method arg to the list. Yes, I realize this is a [contentious issue](https://github.com/rubocop/ruby-style-guide/issues/273), and no, it's not something I feel strongly about.

jurre: I'm fine either way, but I kind of like being able to run `rubocop -A` and have it produce something consistent 🤷

PR #5685, opened 2022-09-12
Basic support for yarn berry workspaces

jurre:
Adds basic support for parsing + updating yarn berry projects using workspaces.


pavera (approved):
A question this brings to mind is: will this make the "no longer vulnerable" issue worse somehow?  Not asking for a solution, but we already have the issue of a package being mentioned multiple times tripping up the security update logic, should we think about trying to move in a direction to improve that?

What happens if the workspaces need to maintain different versions of the specific dependency?

PR #5691, opened 2022-09-12
Rename `.php_cs` -> `.php-cs-fixer.dist.php`

jeffwidman:
Fix deprecation warning by renaming the config file:
```
  15.95 > php-cs-fixer fix --diff --verbose '--dry-run'
  16.10 You are running PHP CS Fixer v2, which is not maintained anymore. Please update to v3.
  16.10 You may find an UPGRADE guide at https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.3.0/UPGRADE-v3.md .
  16.10 If you need help while solving warnings, ask at https://gitter.im/PHP-CS-Fixer, we will help you!
  16.10
  16.10 PHP CS Fixer 2.19.3 Testament by Fabien Potencier and Dariusz Ruminski
  16.10 Runtime: PHP 7.4.30
  16.10 Loaded config default from "/opt/composer/v1/.php_cs".
  16.14 ......
  16.26 Legend: ?-unknown, I-invalid file syntax (file ignored), S-skipped (cached or empty file), .-no changes, F-fixed, E-error
  16.26
  16.26 Checked all files in 0.134 seconds, 14.000 MB memory used
  16.26
  16.26 Detected deprecations in use:
  16.26 - Configuration file `.php_cs` is deprecated, rename to `.php-cs-fixer.php`.
```

From https://github.com/dependabot/dependabot-core/actions/runs/3040792095/jobs/4897247001#step:6:6736

The warning suggests to rename to `php-cs-fixer.php`. However, looking
at https://github.com/keradus/PHP-CS-Fixer/blob/master/UPGRADE-v3.md I
see that is the recommended file name for the _local development_ config file. The
distributed one should be `.php-cs-fixer.dist.php`, so I used that
instead. More details on the distinction between the two types of files
here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/5607

PR #5692, opened 2022-09-12
Rename `phpstan.neon` -> `phpstan.dist.neon`

jeffwidman:
Per https://phpstan.org/config-reference#config-file:
> The usual practice is to have `phpstan.neon.dist` or `phpstan.dist.neon`
> under version control, and allow the user to override certain settings
> in their environment (on their own computer or on a continuous
> integration server) by creating `phpstan.neon` that’s present in
> `.gitignore` file.

I picked `phpstan.dist.neon` for consistency with
`.php-cs-fixer.dist.php` config (#5691).

I didn't add the file to `.gitignore` because I actually want to know if there starts being a new file under that name and see it in git...


jeffwidman:
Confirmed this new config file location is used in https://github.com/dependabot/dependabot-core/actions/runs/3041184101/jobs/4898047993#step:6:6746:
```
 Note: Using configuration file /opt/composer/v1/phpstan.dist.neon.
```

jakecoffman (approved):

PR #5698, opened 2022-09-13
Bump `rspec` and `rspec-its` to latest versions

jeffwidman:
I noticed these are a bit behind, and they use `~>` so will stay within their local minor version ranges. This also brings `rspec` in-sync with the version used in `updater/Gemfile`: https://github.com/dependabot/dependabot-core/blob/4e854a5933035eea4b9f22f9a049712d368b8944/updater/Gemfile#L31

I'm unclear why Dependabot didn't open PR's to bump them up automatically... perhaps something related to us not pinning the full patch version?

There are some other pins that we may want to consider bumping, like ActiveSupport, Webmock, etc... But let's start with this and see what CI thinks, plus if others on the team can explain Dependabot's surprising behavior to me.


jeffwidman:
I noticed that we've also got https://github.com/dependabot/dependabot-core/blob/main/bundler/helpers/v1/Gemfile and https://github.com/dependabot/dependabot-core/blob/main/bundler/helpers/v2/Gemfile which don't appear to be being managed by Dependabot... so may want to manually bump those as well... and check whether we want Dependabot opening PR's to bump Gemfiles or not... since these are apps, not libraries.

deivid-rodriguez (approved):
The reason why no updates where proposed is, I believe, because Dependabot's default strategy for libraries is to widen requirements, and no RSpec 4 version has been released, so no updates are necessary.

In my opinion, this kind of update in general does not buy us much, although in this particular case it seems nice to synchronize the duplicated dependency.

We should look into ways of not having to keep these dependencies duplicated though.

PR #5706, opened 2022-09-13
Add test for peer dependency resolution in yarn berry

jurre:
Add a basic test for peer dep resolution with a yarn berry lockfile. It seems like we don't have any parsing or updating specific tests for peer dependencies and these appear to work as expected.


jeffwidman (commented):
I noticed there's nearly 1 MB of zip files pulled in as part of the `.yarn/cache/*.zip`:
1. Does the `.yarn/cache` dir even need to be committed to the repo or should it be `.gitignore`'d?
2. If it does need to be committed for reproducability, is there another package that could be used for testing that has a smaller number of peer dependencies (at least in bytes size)?

jeffwidman (approved):

jurre:
> Does the .yarn/cache dir even need to be committed to the repo

Good call, we might be able to axe them for this specific test! In others where we test that the yarn cache is updated we do unfortunately need them, so we can't easily `.gitignore` globally

PR #5714, opened 2022-09-13
Relax the composer pin to make it less confusing

jeffwidman:
The previous pin technically allowed any version of composer >= 1.10.9, < 2.0... but because it included the full version pin, it was easy to overlook the `^` and think it was pulling in that older version of composer.

Over in the `v2/composer.json` file, we only pin to `"^2", so I made this pin less explicit for both improved consistency and less confusion.

There is no functional change to `composer.lock`, so this shouldn't have any impact beyond clearer documentation.

### DIFF
--- a/composer/helpers/v1/composer.json
+++ b/composer/helpers/v1/composer.json
@@ -2,7 +2,7 @@
     "require": {
         "php": "^7.4",
         "ext-json": "*",
-        "composer/composer": "^1.10.9"
+        "composer/composer": "^1"
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.9",
--- a/composer/helpers/v1/composer.lock
+++ b/composer/helpers/v1/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "7f2c53b67946a00c067c0930ea168be9",
+    "content-hash": "a0dd40524c184de01243997f04394e71",
     "packages": [
         {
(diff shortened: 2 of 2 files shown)


composer/helpers/v1/composer.json
```diff
@@ -2,7 +2,7 @@
     "require": {
         "php": "^7.4",
         "ext-json": "*",
-        "composer/composer": "^1.10.9"
+        "composer/composer": "^1"
```
bdragon, line 5: Since we're going from `>=1.10.9, <2.0.0` to `>=1.0.0, <2.0.0`, the only thing I wonder about is whether we were relying on anything specific about `1.10.9` as the lower bound?

bdragon (approved):

composer/helpers/v1/composer.json
jeffwidman: Yeah, I checked `git blame` for that when I was researching this... it looked like folks were simply bumping it every so often when they remembered to keep it in sync with the version of Composer in the Dockerfile... nothing special. It just creates confusion though because we've got a very explicit number here, and then another one in `composer.lock` and then another one in the Dockerfile... and the other two are the ones that actually matter for us, but it's easy for someone less familiar with the `composer` ecosystem to think that this pin is affecting things. That was effectively me a few hours ago before I read up on this. 😀 

I also observed the surprising behavior that historically Dependabot opens version bump PRs in this _same_ `composer.json` file for other libraries that are pinned to explicit versions like this... so not sure why it didn't for this version? the only difference was that it was in `require-dev` section and that it used a `~`. Given that this ecosystem is mostly EOL'd, I didn't dig into which of the above two it was.

Issue #5695, opened 2022-09-12
Address these `phpstan` warnings in `composer/v1/*`


jeffwidman:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code improvement description

Saw these warnings in https://github.com/dependabot/dependabot-core/actions/runs/3040792095/jobs/4897247001:

```
16.51 > phpstan analyse
16.82 Note: Using configuration file /opt/composer/v1/phpstan.neon.
18.22  0/6 [░░░░░░░░░░░░░░░░░░░░░░░░░░░░]   0% 6/6 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%
18.31 
18.31 
18.31  [OK] No errors                                                                 
18.31 
18.73 ./composer.json is valid for simple usage with Composer but has
18.73 strict errors that make it unable to be published as a package
18.73 See https://getcomposer.org/doc/04-schema.md for details on the schema
18.73 # General warnings
18.73 - No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
18.73 # Publish warnings
18.73 - name : The property name is required
18.73 - description : The property description is required
```

As noted, these are not strictly needed given that we aren't publishing packages / libraries, but it's always nice to silence warnings for less visual noise.

PR #5716, opened 2022-09-14
Add composer fields to silence PHPStan

jeffwidman:
Addresses the following warnings:
```
16.51 > phpstan analyse
16.82 Note: Using configuration file /opt/composer/v1/phpstan.neon.
18.22  0/6 [░░░░░░░░░░░░░░░░░░░░░░░░░░░░]   0% 6/6 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%
18.31
18.31
18.31  [OK] No errors
18.31
18.73 ./composer.json is valid for simple usage with Composer but has
18.73 strict errors that make it unable to be published as a package
18.73 See https://getcomposer.org/doc/04-schema.md for details on the schema
18.73 # General warnings
18.73 - No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
18.73 # Publish warnings
18.73 - name : The property name is required
18.73 - description : The property description is required
```

As noted, these are not strictly needed given that we aren't publishing packages / libraries, but it's always nice to silence warnings for less visual noise.

After editing the `composer.json` file, `composer.lock` was regenerated
using `composer install && composer update --lock`. I made sure to use
`composer1` for the `v1` helpers, and `composer` for the `v2` helpers.

Fix https://github.com/dependabot/dependabot-core/issues/5695

### DIFF
--- a/composer/helpers/v1/composer.json
+++ b/composer/helpers/v1/composer.json
@@ -1,4 +1,7 @@
 {
+    "name": "dependabot/composer-v1-helper",
+    "description": "A helper package for Dependabot to perform updates using Composer",
+    "license": "The Prosperity Public License 2.0.0",
     "require": {
         "php": "^7.4",
         "ext-json": "*",
--- a/composer/helpers/v2/composer.lock
+++ b/composer/helpers/v2/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "27bae1c74f03f67fba9fa88beed4b83c",
+    "content-hash": "6753b4220fb20aa0fab7c0de277385ef",
     "packages": [
         {
             "name": "composer/ca-bundle",
@@ -61,6 +61,11 @@
                 "ssl",
                 "tls"
             ],
+            "support": {
+                "irc": "irc://irc.freenode.org/composer",
+                "issues": "https://github.com/composer/ca-bundle/issues",
+                "source": "https://github.com/composer/ca-bundle/tree/1.3.2"
+            },
             "funding": [
                 {
                     "url": "https://packagist.com",
@@ -166,6 +171,11 @@
                 "dependency",
                 "package"
             ],
+            "support": {
+                "irc": "ircs://irc.libera.chat:6697/composer",
+                "issues": "https://github.com/composer/composer/issues",
+                "source": "https://github.com/composer/composer/tree/2.3.9"
+            },
             "funding": [
                 {
                     "url": "https://packagist.com",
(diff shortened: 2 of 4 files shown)


jeffwidman:
Complains that I need to regenerate the `composer.lock` file, but I can't find a way to do that w/o bumping a bunch of libs... when all we need is a single hash updated. 🤦 

I'll open a separate PR to run `composer update` to pickup the other changes first.

jurre (approved):

stefangr:
@jeffwidman use `composer update --lock` to only update the lockfile, but not the libraries.

See: https://getcomposer.org/doc/03-cli.md#update-u

jeffwidman:
Thanks, I did try `composer update --lock`, but it still resulted in bumping the transitive dependencies which are only listed in the lockfile... what i want is to only regenerate the hash that composer uses to validate that `composer.json` and `composer.lock` are in-sync... w/o bumping any library versions at all, not even transitive ones.

stefangr:
@jeffwidman
With a fresh checkout of this branch and a `composer install` to install the dependencies from the current composer.lock file, the result of a `composer update --lock` is as I expected.

**v1/composer.lock**
- Only the content-hash is changed

**v2/composer.lock**
- The content-hash is changed
- A support section is added to some of the dependencies (when run with the composer v2 executable)

In the vendor/composer directory composer keeps a file (installed.json) that contains the actual versions of the installed dependencies.
If you have updated dependencies in a different branch, than that is in the installed.json.
If you perform a `composer install` the installed.json file is updated to the situation from the composer.lock.

My guess is that you did not perform a `composer install` before executing the `composer update --lock` while the installed.json from the vendor directory was updated in one of the other composer related branches.

jeffwidman:
> My guess is that you did not perform a composer install before executing the composer update --lock while the installed.json from the vendor directory was updated in one of the other composer related branches.

Thanks @stefangr that was exactly it!

composer/helpers/v2/composer.lock
```diff
@@ -2643,5 +2780,5 @@
         "ext-json": "*"
     },
     "platform-dev": [],
-    "plugin-api-version": "1.1.0"
+    "plugin-api-version": "2.3.0"
```
jeffwidman, line 2783: I did notice this change also crept in with `composer update --lock`. But it should be fine given that we don't use any plugins, and even if we did I suspect we'd want to be on the v2 API given that this is `composer` `v2`.

composer/helpers/v1/composer.json
```diff
@@ -1,4 +1,7 @@
 {
+    "name": "dependabot/composer-v1-helper",
+    "description": "A helper package for Dependabot to perform updates using Composer",
+    "license": "The Prosperity Public License 2.0.0",
```
jeffwidman, line 4: PHPStan will still complain about this, since it's not a SPDX identifier.

I looked into submitting this license for inclusion into SPDX index, but after reading through https://github.com/spdx/license-list-XML/pull/960 I decided that wasn't worth pursuing further.

 I could have used `"proprietary"`, but decided to just ignore that warning for now.

PR #5717, opened 2022-09-14
Update `v1/composer.lock` using `composer1 update`

jeffwidman:
Used `composer1 update` to update the `v1/composer.lock` file:
```
[dependabot-core-dev] ~/dependabot-core/composer/helpers/v1 $ composer1 update
Loading composer repositories with package information
Warning from https://repo.packagist.org: Support for Composer 1 is deprecated and some packages will not be available. You should upgrade to Composer 2. See https://blog.packagist.com/deprecating-composer-1-support/
Info from https://repo.packagist.org: #StandWithUkraine
Updating dependencies (including require-dev)
Nothing to install or update
Package php-cs-fixer/diff is abandoned, you should avoid using it. No replacement was suggested.
Writing lock file
Generating autoload files
28 packages you are using are looking for funding.
Use the `composer fund` command to find out more!
```
I suspect the reason that Dependabot hasn't opened PR's to bump these versions is because they're transitive dependencies.

I opened a sister PR for `composer` v2, but keeping them separate in case we need to revert anything:
https://github.com/dependabot/dependabot-core/pull/5718

### DIFF
--- a/composer/helpers/v1/composer.lock
+++ b/composer/helpers/v1/composer.lock
@@ -8,16 +8,16 @@
     "packages": [
         {
             "name": "composer/ca-bundle",
-            "version": "1.3.1",
+            "version": "1.3.3",
             "source": {
                 "type": "git",
                 "url": "https://github.com/composer/ca-bundle.git",
-                "reference": "4c679186f2aca4ab6a0f1b0b9cf9252decb44d0b"
+                "reference": "30897edbfb15e784fe55587b4f73ceefd3c4d98c"
             },
             "dist": {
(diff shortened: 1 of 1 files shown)


deivid-rodriguez (approved):
Seems fine! Not feel super confident when reviewing PRs for ecosystems I don't know since I don't know how stable this kind of thing is there, but I think our standard practice is to try keep everything up to date and mainly trust CI, so should be fine!

jurre (approved):
It's probably fine, if tests are passing let's just roll it out and keep an eye on production metrics etc

PR #5718, opened 2022-09-14
Update `v2/composer.lock` using `composer update`

jeffwidman:
Used `composer update` to update the `v2/composer.lock` file.

I first temporarily pinned `composer` to `2.4.1` to match what's
currently on `main` as I want to isolate `composer` bumps to distinct
PR's, but didn't actually commit that pin so that it wouldn't block
Dependabot, only pinned while running `composer update`.

### DIFF
--- a/composer/helpers/v2/composer.lock
+++ b/composer/helpers/v2/composer.lock
@@ -4,20 +4,20 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "367ec5f822a3d8d251c7188331b495d5",
+    "content-hash": "6e0414ea165f045f95b239ca36ddca93",
     "packages": [
         {
             "name": "composer/ca-bundle",
-            "version": "1.3.3",
+            "version": "1.3.5",
             "source": {
                 "type": "git",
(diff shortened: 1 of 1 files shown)


jeffwidman:
Marking as Draft until https://github.com/dependabot/dependabot-core/pull/5577 is merged as that way we do the biggest risk bump of composer separate from the others...

composer/helpers/v2/composer.lock
```diff
@@ -80,36 +80,111 @@
...
+                    "url": "https://tidelift.com/funding/github/packagist/composer/composer",
+                    "type": "tidelift"
+                }
+            ],
+            "time": "2022-06-19T11:31:27+00:00"
         },
         {
             "name": "composer/composer",
-            "version": "2.3.9",
+            "version": "2.4.2",
```
jeffwidman, line 160: Marking as Draft until https://github.com/dependabot/dependabot-core/pull/5577 is merged as that way we do this biggest risk bump of composer separate from the others...

jeffwidman: I backed out the `composer` bump so that it only bumps the non-composer stuff.

jeffwidman:
So this only bump `composer/ca-bundle`, which isn't a big deal, we can pick these up when we bump `composer` itself...

PR #5732, opened 2022-09-15
Fix URIs logged by dry-run

deivid-rodriguez:
On the one hand, it was missing ":" after the colon. On the other hand, including the port makes the url not browsable, which I think it's the point of logging this.

Please trust me and don't even try to visit `https//pypi.org`, which is how I got here 🤣.


jeffwidman (approved):
Oof! Good catch!

PR #5733, opened 2022-09-15
Make `script/debug` a simple wrapper over the CLI

deivid-rodriguez:
Not sure if this makes sense to you, but today I wanted to use a full CLI run, with my development copy of dependabot-core code, but without a debugging shell. And the current `script/debug` script doesn't let you do that.

I thought about removing the hardcoded `--debug` flag, and if you want a debugger console, you can pass the `--debug` flag explicitly. Of course, that's more typing, so alternatively we could, for example:

* Use `dep` for the script name.
* Keep `script/debug` and do what I'm doing here in a separate script.

Does this make sense?

### DIFF
--- a/script/dependabot
+++ b/script/dependabot
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-dependabot "$@" --debug \
+dependabot "$@" \
   -v "$(pwd)"/.core-bash_history:/home/dependabot/.bash_history \
   -v "$(pwd)"/updater/bin:/home/dependabot/dependabot-updater/bin \
   -v "$(pwd)"/updater/lib:/home/dependabot/dependabot-updater/lib \


jeffwidman (approved):
👍 The only thing we lose with this is discoverability. 

I'm a fan of few scripts where possible, so prefer not to have separate run + debug script (esp if only for one flag!)

My one question since I've never played with this script yet is where is this `--debug` flag even get passed to / kick in? Is there something in the dockerfile that is dropping the `dependabot` CLI `go` binary into this folder? I checked but didn't see anything... maybe I'm just blind though.

jakecoffman:
Makes sense 👍

jakecoffman:
@jeffwidman You have to install the CLI manually for now, it's not used anywhere beside the [smoke tests](https://github.com/dependabot/dependabot-core/blob/f68191fad0b0bf7a1b4222f28a13c64ca6cc33c7/.github/workflows/smoke.yml#L167-L169) where it gets downloaded on-demand each time.

deivid-rodriguez:
How about `script/dep update bundler ...`? Reads nice and short!

jakecoffman:
I'm up for anything! You can also add an alias in your shell to make it even shorter `dep update bundler ...` 😍

jeffwidman:
`script/dep` to me is ambiguous... is that short for "dependabot" or "dependency"?? Rather keep it a little longer and use tab-completion for speed, or if you're paranoid of carpal tunnel 😀  use a custom local alias like Jake suggested.

alternatively, given the names of the [other scripts in this folder](https://github.com/dependabot/dependabot-core/tree/218f0af2354390c2f908d8d1963cdcd1e855728e/script), could do `script/run` as that'd make it clear it's the "normal" entrypoint within this folder. I actually prefer that as `script/dependabot` is a little ambiguous of what it's doing (unless you happen to remember that the CLI is generally called `dependabot`)

deivid-rodriguez:
Renamed to `dep`!

deivid-rodriguez:
> script/dep to me is ambiguous... if I parachuted into this folder as a new dev, I'd wonder is that short for "dependabot" or "dependency" or is it some other random Ubuntu tool that we needed to script or something

I think both abbreviations work and read great 😅.

But I'll update it back since I don't think it matters too much!

deivid-rodriguez:
`dependabot` matches the name chosen for the CLI, so makes sense to stick to that!

jeffwidman (approved):

jeffwidman:
I went ahead and merged since this one was green and up to date, that way it's one less PR that we have to play the "rebase" game on.

Issue #5363, opened 2022-07-12
Dependabot fails to parse a more complex poetry dependency


mvadari:
**Package ecosystem**
pip
**Package manager version**
poetry 
**Language version**
Python 3.7-3.10
**Manifest location and content before the Dependabot update**
https://github.com/XRPLF/xrpl-py/blob/master/pyproject.toml
**dependabot.yml content**
```
version: 2
updates:
- package-ecosystem: pip
  directory: "/"
  schedule:
    interval: monthly
    time: "15:00"
  open-pull-requests-limit: 10
```
**What you expected to see, versus what you actually saw**
Dependabot hasn't been running at all on this repo, but it has on a repo with an identical dependabot.yml file. I went to Insights -> Dependency graph -> Dependabot and found this `Illformed requirement` error:
![image](https://user-images.githubusercontent.com/8029314/178514662-fd993341-e991-4c93-bd47-68b3fa47bd73.png)
I'm pretty sure it's coming from this part of the `pyproject.toml` file:
```
websockets = [
    {version = ">=9.0.1 <11.0", python = ">= 3.7, < 3.10"},
    {version = "^10.0", python = "^3.10"}
]
```
This is completely valid in poetry, but Dependabot is struggling.
**🕹 Bonus points: Smallest manifest that reproduces the issue**
See above.

deivid-rodriguez:
Hei @mvadari, thanks for the report!

This is a weird one because while Poetry supports this format, standard Python does not. I should've probably got back to you earlier when I first looked at this, to at least mention the easy workaround to you: change the requirements to be separated by commas instead of spaces.

Anyways, if Poetry supports this, we probably should too, so I opened a PR to fix it! See #5735.

mvadari:
Thanks!

I'm unfamiliar with how new features are added, will dependabot just automatically be resolved at some future point or will I need to change something in the config to update?

deivid-rodriguez:
No problem!

I don't think you need to change anything, it should work next time Dependabot tries to update your repo after the fix is deployed!

mvadari:
Confirming that it is indeed working, thanks! 💯

jeffwidman:
Great! Thanks for circling back to let us know.

PR #5735, opened 2022-09-15
Fix multiple Python requirements separated by whitespace

deivid-rodriguez:
Standard Python does not support this, but Poetry does, so when they appear on Poetry dependency files, they make dependabot crash.

Fixes https://github.com/dependabot/dependabot-core/issues/5363.

### DIFF
--- a/python/lib/dependabot/python/requirement.rb
+++ b/python/lib/dependabot/python/requirement.rb
@@ -60,6 +60,9 @@ module Dependabot
         requirements = requirements.flatten.flat_map do |req_string|
           next if req_string.nil?
 
+          # Standard python doesn't support whitespace in requirements, but Poetry does.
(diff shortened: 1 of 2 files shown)


jeffwidman (approved):
Given that these are non-standard and only supported by poetry, I'd suggest adding code comments for each of these three code blocks. Something simple like:
```
# Standard python doesn't support whitespace in requirements, but Poetry does.
```

I realize it's in the commit log, but still it's too easy to not understand _why_ this is needed in the logic / tests otherwise.

jeffwidman:
also, why did poetry go this route? 🤦

deivid-rodriguez:
I have no idea, but seems much better than Python ignoring (but not erroring out) all requirements after the first if you don't split them with commas 🤯.

Issue #5729, opened 2022-09-15
Use the metadata source when determining how to sanitize mentions


jeffwidman:
In Dependabot PR's, we sanitize team / user mentions in the PR metadata (changelog etc) to avoid notification spam.

However, as @andrcuns very patiently pointed out in https://github.com/dependabot/dependabot-core/pull/3437, today that logic is a bit of a hack: the "sanitization"  assumes that all mentions should be rewritten as GitHub links.

While that works the majority of the time, that's only because GitHub happens to be the most popular place for open source projects.

Instead, we should be looking at the _metadata source_, rather than the _Dependabot running environment_. In other words, if we fetched metadata from GitHub, we should sanitize the mentions to GitHub links. And if we fetched the metadata from GitLab we should sanitize the mentions to GitLab links. 

There are certainly places we fetch from that aren't on those hosting platforms, such as PyPI, Crates.IO, etc and for those it's probably fine to assume GitHub.

That will probably cover the vast majority of cases, but I expect the URL linking structure is pretty generic across platforms so wouldn't be hard to tag on BitBucket, Azure Devops, etc since in most cases the only difference between platforms will likely be the hostname...

PR #5739, opened 2022-09-16
Sanitize metadata links on all platforms

jeffwidman:
As pointed out in https://github.com/dependabot/dependabot-core/issues/5729, we should be using the "source" of the metadata for deciding how to sanitize, not the "place where Dependabot is hosted".

This hack was annoying the folks who use GitLab to the point that [they monkey-patched this code to also run the sanitizer on GitLab](https://github.com/dependabot/dependabot-core/pull/3437#issuecomment-816582168), and then contributed that back in #3437.

If it's true on GitLab, it's even more true on other platforms as well for the reasons explained in https://github.com/dependabot/dependabot-core/pull/3437#issuecomment-1247597678.

So remove this guard altogether.

### DIFF
--- a/common/lib/dependabot/pull_request_creator/message_builder/metadata_presenter.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder/metadata_presenter.rb
@@ -245,8 +245,6 @@ module Dependabot
         end
 
         def sanitize_links_and_mentions(text, unsafe: false)
-          return text unless source.provider == "github" || source.provider == "gitlab"
-
           LinkAndMentionSanitizer.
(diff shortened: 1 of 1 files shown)


Nishnha (approved):
Kind of wonder why we put up the guard in the first place?

Looks like it was added in https://github.com/dependabot/dependabot-core/pull/2539, specifically in https://github.com/andrcuns/dependabot-core/commit/8fb45b2c950d9bf0c87058ad3090982ee564f289 where we switched out using `source.provider` for just `source`. 🤷🏾 

Seems fine to remove the guard and just sanitize regardless, though.

PR #5744, opened 2022-09-18
Fix commitlint message style detection

### LINKED ISSUE #1666: Cannot configure commit message case to match commitlint
My projects' commitlint specifications require both a prefix & scope as well as sentence-case for the actual message. It seems dependabot will use sentence-case when no prefix is used or lower case when it is, but I can't seem to get it to mix and match. [This comment](https://github.com/dependabot/dependabot-core/issues/406#issuecomment-385024411) on a related issue seems to suggest is should be possible, but I can't figure out how.

I'm configuring commitlint using the YAML file option (`.commitlintrc.yaml`) rather than `packages.json` if that makes a difference.

Is there a way to set up the correct casing?

deivid-rodriguez:
We should not try to be too smart since there's too many different commitlint setups. Instead, it seems more sound to use the style of previous commits.

Also, I think commit message should default to use the configured prefix for development dependencies if `prefix` is configured, but `prefix-development` is not.

This is just a naive approach that works with `bin/dry-run.rb` but probably specs need fixing, etc.

Fixes #1666.

### DIFF
--- a/common/spec/dependabot/pull_request_creator/message_builder_spec.rb
+++ b/common/spec/dependabot/pull_request_creator/message_builder_spec.rb
@@ -15,7 +15,7 @@ RSpec.describe Dependabot::PullRequestCreator::MessageBuilder do
       credentials: credentials,
       pr_message_header: pr_message_header,
       pr_message_footer: pr_message_footer,
-      commit_message_options: { signoff_details: signoff_details, trailers: trailers },
+      commit_message_options: commit_message_options,
       vulnerabilities_fixed: vulnerabilities_fixed,
       github_redirection_service: github_redirection_service
     )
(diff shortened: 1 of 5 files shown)


common/spec/dependabot/pull_request_creator/message_builder_spec.rb
```diff
@@ -446,6 +447,29 @@ def commits_details(base:, head:)
...
+                headers: json_header
+              )
+          end
+
+          it do
+            is_expected.to eq("chore(deps): Bump business from 1.4.0 to 1.5.0")
+          end
+
+          context "and with commit messages explicitly configured" do
+            let(:commit_message_options) { super().update(prefix: "chore(dependencies)") }
```
landongrindheim, line 465: I appreciate your use of `super()` here because it signals that I need to look up the chain. I'm a fan of RSpec, but one of my least favorite parts of it is that it can lead to an explosion of mystery guests. Because of this, I wonder if we could use `#merge`, which I think of as being a `Hash`-specific method. I think it gives 🤏 more context to the variable, and takes away a little bit of ambiguity _for me_.

```suggestion
            let(:commit_message_options) { super().merge(prefix: "chore(dependencies)") }
```

If this goes against your instincts, please feel free to ignore 😄

landongrindheim (approved):

common/spec/dependabot/pull_request_creator/message_builder_spec.rb
deivid-rodriguez: Using `#merge` feels better to me too, not sure why I used `update`!

Issue #5396, opened 2022-07-19
[pub] Log the Dart / Flutter SDK version selected when running dependabot


jonasfj:
Diving through logs I found it hard to debug what Dart/Flutter version dependabot had used.
It might be nice if we can write that to log, that would help people trying to find out why their dependabot setup is acting up.

This is easily  the case if you don't have sufficient SDK constraints, in which case dependabot will also pick latest stable SDK.

jonasfj:
While doing this we could also log a few other useful stings.

And check if it's running the flutter install script too many times, at-least we should double check logs, the one I looked at seemed a bit weird.

cc @sigurdm,

jeffwidman:
Care to submit a PR?

jeffwidman:
Addressed by https://github.com/dependabot/dependabot-core/pull/5748

@sigurdm Thanks for jumping on this so fast!

PR #5748, opened 2022-09-19
[pub] Log the Dart / Flutter SDK version selected

sigurdm:
Fix of https://github.com/dependabot/dependabot-core/issues/5396

### DIFF
--- a/pub/lib/dependabot/pub/helpers.rb
+++ b/pub/lib/dependabot/pub/helpers.rb
@@ -109,52 +109,74 @@ module Dependabot
       def ensure_right_flutter_release
         @ensure_right_flutter_release ||= begin
           versions = Helpers.run_infer_sdk_versions url: options[:flutter_releases_url]
-          if versions
-            Dependabot.logger.info(prefixed_log_message("Installing the Flutter SDK version: #{versions['flutter']} from channel #{versions['channel']} with Dart #{versions['dart']}"))
-          else
-            Dependabot.logger.info(prefixed_log_message("Failed to infer the flutter version. Attempting to use latest stable release."))
-          end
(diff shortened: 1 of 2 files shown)


pub/lib/dependabot/pub/helpers.rb
```diff
@@ -138,9 +144,12 @@ def ensure_right_flutter_release
           end
 
           parsed = JSON.parse(stdout)
+          flutter_version = parsed["frameworkVersion"]
+          dart_version = parsed["dartSdkVersion"].split.first
```
jurre, line 148:
```suggestion
          dart_version = parsed["dartSdkVersion"]&.split&.first
```

Should we add this to safe-guard against `parsed["dartSdkVersion"]` being nil? This would now raise an error if it's nil.

jurre:
Thanks @sigurdm! I noticed there are a few linter violations that you'll see in the CI output, they should be easy to fix up but lmk if I can help.

I also notice this breaks a (new) end-to-end test:

```
2022/09/19 09:05:04 [010] POST http://host.docker.internal:40845/update_jobs/cli/record_update_job_error
2022/09/19 09:05:04 type was unexpected: expected update_dependency_list got record_update_job_error
```

Which hints that this might be causing the job to fail, I left an in-line comment about what could potentially cause an issue here, but I may be off.

pub/lib/dependabot/pub/helpers.rb
sigurdm: We should always have a dart version. I made the check such that the error message will become nicer.

jurre (approved):
This looks good to me, thanks @sigurdm!

PR #5749, opened 2022-09-19
Gate yarn berry changes behind feature-flag

jurre:
In order to merge these changes into main without disrupting existing behavior, gate the changes behind a feature flag, so we can gradually roll them out and iron out issues without building up a huge set of changes.

I've opted to explicitly _enable_ the feature in our tests, so they run against the new situation, we can add a few tests that ensure the new behavior doesn't run _without_ the flag present, but I'm a little concerned about how invasive this feature-flag change is to be honest, needing to pass the feature flag around etc. and would like to possibly rethink the approach first.

This also still includes the changes to always clone the full repo, which we should flag in updater/api instead.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser.rb
@@ -3,6 +3,7 @@
 # See https://docs.npmjs.com/files/package.json for package.json format docs.
 
 require "dependabot/dependency"
+require "dependabot/experiments"
 require "dependabot/file_parsers"
 require "dependabot/file_parsers/base"
 require "dependabot/shared_helpers"
@@ -327,7 +328,7 @@ module Dependabot
               dependency_files.
               select { |f| f.name.end_with?("package.json") }.
               reject { |f| f.name == "package.json" }.
-              reject { |f| f.name.include?("node_modules/") }.
+              reject { |f| f.name.include?("node_modules/") if Experiments.enabled?(:yarn_berry) }.
               reject(&:support_file?)
 
             [
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
@@ -96,8 +96,8 @@ module Dependabot
             parse_yarn_lock(yarn_lock).each do |req, details|
               next unless semver_version_for(details["version"])
               next if alias_package?(req)
-              next if workspace_package?(req)
-              next if req == "__metadata"
+              next if Experiments.enabled?(:yarn_berry) && workspace_package?(req)
+              next if Experiments.enabled?(:yarn_berry) && req == "__metadata"
 
               # NOTE: The DependencySet will de-dupe our dependencies, so they
               # end up unique by name. That's not a perfect representation of
@@ -190,7 +190,7 @@ module Dependabot
         end
 
         def alias_package?(requirement)
-          requirement.match?(/@npm:(.+@(?!npm))/)
+          Experiments.enabled?(:yarn_berry) ? requirement.match?(/@npm:(.+@(?!npm))/) : requirement.include?("@npm:")
         end
 
         def workspace_package?(requirement)
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "dependabot/experiments"
 require "dependabot/file_updaters"
 require "dependabot/file_updaters/base"
 require "dependabot/file_updaters/vendor_updater"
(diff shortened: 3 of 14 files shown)


common/lib/dependabot/experiments.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Dependabot
+  module Experiments
+    def self.register(name, value)
+      @experiments ||= {}
+      @experiments[name.to_sym] = value
+    end
```
landongrindheim, line 8: I really like this move to globalize the Experiment! Feels much safer than passing around a hash that can be mutated accidentally 😅

This interface reminds me somewhat of Flipper. In that case, `#register` is [scoped to groups](https://github.com/jnunemaker/flipper/blob/2dc8f6da8d75bc889cf262e7197b2d26a07a4f6d/lib/flipper.rb#L71-L91), which isn't really what we're after here. Of course, we don't have to follow that interface, but it might be worth considering. I could see `.toggle(name, should_enable)`, `.enable(name)` and `.disable(name)` or something along those lines.

I know this is a PR to be merged into a feature branch, but I thought it was worth mentioning 😇

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser.rb
```diff
@@ -328,7 +328,7 @@ def package_files
               dependency_files.
               select { |f| f.name.end_with?("package.json") }.
               reject { |f| f.name == "package.json" }.
-              reject { |f| f.name.include?("node_modules/") if options[:yarn_berry] }.
+              reject { |f| f.name.include?("node_modules/") if Experiments.enabled?(:yarn_berry) }.
```
landongrindheim, line 331: I think this is a great example of the improvement being added in this PR ✨

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater.rb
```diff
@@ -54,7 +55,7 @@ def updated_dependency_files
           )
         end
 
-        if options[:yarn_berry]
+        if Experiments.enabled?(:yarn_berry)
```
landongrindheim, line 58: This gives me so much more confidence as I read it 👌

landongrindheim (approved):
Great improvement! I left a couple comments, but I think this is a nice addition to the codebase 🥇

common/lib/dependabot/experiments.rb
```diff
@@ -0,0 +1,14 @@
...
+
+    def self.enabled?(name)
+      @experiments[name.to_sym]
```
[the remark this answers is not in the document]
jurre: Would this be equivalent to `@experiments.key?(name)`?  It wouldn't let us pass a `false` value to the experiments, but I guess that's kind of pointless anyway?

jurre:
Thanks @landongrindheim, I was planning on extracting this and opening a separate PR against main, so I appreciate your comments!

common/lib/dependabot/experiments.rb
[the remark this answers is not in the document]
landongrindheim: I think of `@experiments.key?(name)` as being a little surprising because you could be storing `false` or `nil` in the value. FWIW, I don't think this change is _necessary_, but the `true`/`false` from a predicate is something I think is worth sticking to.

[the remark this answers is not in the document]
jurre: The same goes for what you're proposing though, right?

```ruby
irb(main):001:0> @experiments = { test: false }
=> {:test=>false}
irb(main):002:0> !@experiments[:test].nil?
=> true
```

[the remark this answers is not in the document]
landongrindheim: 🤦‍♂️ Yeah, you're right. Maybe `!!@experiments[name]`?

common/lib/dependabot/experiments.rb
jurre: Yeah I'm definitely open to adding those methods, but I was thinking let's add them when we actually need the functionality?

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser.rb
```diff
@@ -327,7 +328,7 @@ def package_files
               dependency_files.
               select { |f| f.name.end_with?("package.json") }.
               reject { |f| f.name == "package.json" }.
-              reject { |f| f.name.include?("node_modules/") }.
+              reject { |f| f.name.include?("node_modules/") if Experiments.enabled?(:yarn_berry) }.
```
Nishnha, line 331: I think we would want to continue ignoring `node_modules/` for projects without yarn_berry enabled too?

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
```diff
@@ -96,8 +96,8 @@ def yarn_lock_dependencies
             parse_yarn_lock(yarn_lock).each do |req, details|
               next unless semver_version_for(details["version"])
               next if alias_package?(req)
-              next if workspace_package?(req)
-              next if req == "__metadata"
+              next if Experiments.enabled?(:yarn_berry) && workspace_package?(req)
+              next if Experiments.enabled?(:yarn_berry) && req == "__metadata"
```
Nishnha, line 100: This has a side effect of no longer skipping over `workspace_packages` or `__metadata requirements` anymore unless the project has yarn_berry enabled. 

I don't think that's the intended effect though?

pavera: This PR is targeting the yarn_berry branch these statements were added in that branch, so this just puts them behind a feature flag.

pavera (approved):

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser.rb
jurre: That's a new change made specifically for the yarn berry changes, so we want to exclude it when the flag is disabled to preserve the old behavior

PR #5755, opened 2022-09-20
Add new interface to register experiments

jurre:
Currently experiments are required to be passed in as an optional
`options` flag to the individual classes like the FileFetcher and
FileUpdater. This is fine, but it becomes a bit cumbersome when more
deeply nested objects need to access this flag, as we need to pass it
down to all of them.

This introduces a global `Dependabot::Experiments` interface that we can
use to register experiments and access them where needed.

The idea is that these are set in the Updater before running an update,
and will be available for the duration of the job.

Right now most of the existing experiments are left as is, using the `options` flags,
but I've ported the `kubernetes_enabled` feature flag for the Docker ecosystem to
demonstrate how this would work.

There is more functionality we could add to this interface, but for now it just adds what
is needed.

### DIFF
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -224,6 +224,10 @@ option_parse = OptionParser.new do |opts|
         [o.strip.downcase.to_sym, true]
       end
     end
+
+    $options[:updater_options].each do |name, val|
+      Dependabot::Experiment.register(name, val)
+    end
   end
 
   opts.on("--security-updates-only",
--- a/common/lib/dependabot/experiments.rb
+++ b/common/lib/dependabot/experiments.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module Dependabot
+  module Experiments
+    @experiments = {}
+
+    def self.reset!
+      @experiments = {}
+    end
+
+    def self.register(name, value)
+      @experiments[name.to_sym] = value
+    end
+
+    def self.enabled?(name)
+      !!@experiments[name.to_sym]
+    end
+  end
+end
--- a/common/spec/dependabot/experiments_spec.rb
+++ b/common/spec/dependabot/experiments_spec.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+require "dependabot/experiments"
+
+RSpec.describe Dependabot::Experiments do
+  before do
+    described_class.reset!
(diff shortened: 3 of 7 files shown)


bin/dry-run.rb
```diff
@@ -224,6 +224,10 @@
         [o.strip.downcase.to_sym, true]
       end
     end
+
+    $options[:updater_options].each do |name, val|
+      Dependabot::Experiment.register(name, val)
+    end
```
landongrindheim, line 230: Do we use `:updater_options` for more than experiments? Maybe it's not such a big deal for `dry-run`, but the name doesn't really fit the purpose.

landongrindheim (approved):
I *really* like this change! I find it much easier to reason about and I'm guessing it's less bug-prone than the previous use of `options` being passed around everywhere 🙌

bin/dry-run.rb
jurre: Yeah we use a few different names in different places, we ended up with an `options` arg for the `FileUpdater` with the idea that it could be used for more things than just experiments. I'm not sure that we ever have, and going forward we should probably rename things where it makes sense, but I was aiming to keep the change as small as possible

pavera (approved):
This feels much cleaner than before. 🚀

jurre:
cc @brendandburns this change affects how the `kubernetes_enabled` flag is checked in the Docker ecosystem, so when you update Dependabot Core on your end you'll need to make a small change to make things work, it would be very similar to https://github.com/dependabot/dependabot-core/pull/5755/files#diff-b18ee444afdafa5177657bddd586b0097d8d8f631729850ed2afd5b6eb573e5aR65-R67. Please let me know if I can help with that in any way!

jeffwidman:
Thanks @jurre, funny having never looked at the actual code, I'd always assume feature/experiment flags were implemented this way already. 😆 

Is there any cleanup that needs to be done in a follow-on PR to remove code / feautres that this PR makes superfluous? Based on the conversation ☝️, I wasn't clear if the `options` arg should be kept or removed since we never use it?

jurre:
> Is there any cleanup that needs to be done in a follow-on PR to remove code / feautres that this PR makes superfluous?

Yeah we _can_ migrate the existing flags to use this functionality, but we can also keep them as is and wait for them to naturally be removed as those experiments are wrapped up.

PR #5763, opened 2022-09-21
allow interactive debugging in the CLI

jakecoffman:
When running in the CLI with `script/dependabot`, this allows us to add `debugger` in the code and then invoke `bin/run fetch_files --debug` to start a debugging session once that line is hit.

### DIFF
--- a/updater/bin/fetch_files.rb
+++ b/updater/bin/fetch_files.rb
@@ -7,6 +7,7 @@ $stdout.sync = true
 require "raven"
 require "dependabot/setup"
 require "dependabot/file_fetcher_job"
+require "debug" if ENV["DEBUG"]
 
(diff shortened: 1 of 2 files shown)


jeffwidman (commented):
I think `byebug` goes away in Ruby 3.1 with their new `debug` gem?

If you can afford to wait a day or two, let's use that instead as the breakpoint and I'll use this as a reminder to push https://github.com/dependabot/dependabot-core/pull/5447 over the line.

Happy to have a quick call if I'm misunderstanding the relationship between `byebug` and `debug`.

jeffwidman:
#5447 is merged, so this should be unblocked.

landongrindheim (approved):

PR #5767, opened 2022-09-21
Remove label filter for azure

mburumaxwell:
Azure does not have centralized labels, as such none are fetched. Filtering labels in the `Labeler` results in no labels if `custom_labels` is not null. This PR removes the filtering if `source.provider == "azure"`.

Fix: #6231 

Related to: 
- https://github.com/tinglesoftware/dependabot-azure-devops/pull/348
- https://github.com/tinglesoftware/dependabot-azure-devops/issues/317

### DIFF
--- a/common/lib/dependabot/pull_request_creator/labeler.rb
+++ b/common/lib/dependabot/pull_request_creator/labeler.rb
@@ -174,7 +174,11 @@ module Dependabot
       end
 
       def default_labels_for_pr
-        if custom_labels then custom_labels & labels
+        if custom_labels
+          # Azure does not have centralised labels
+          return custom_labels if source.provider == "azure"
+
+          custom_labels & labels
         else
           [
             default_dependencies_label,
--- a/common/spec/dependabot/pull_request_creator/labeler_spec.rb
+++ b/common/spec/dependabot/pull_request_creator/labeler_spec.rb
@@ -336,6 +336,40 @@ RSpec.describe Dependabot::PullRequestCreator::Labeler do
       end
     end
 
+    context "with Azure details" do
+      let(:source) do
+        Dependabot::Source.new(provider: "azure", repo: "gocardless/bump")
+      end
+
+      context "when the 'dependencies' label doesn't yet exist" do
+        it "creates a 'dependencies' label" do
+          labeler.create_default_labels_if_required
+
+          expect(labeler.labels_for_pr).to include("dependencies")
+        end
+      end
+
+      context "for an update that fixes a security vulnerability" do
+        let(:includes_security_fixes) { true }
+
+        context "when the 'security' label doesn't yet exist" do
+          it "creates a 'security' label" do
+            labeler.create_default_labels_if_required
+
+            expect(labeler.labels_for_pr).to include("security")
+          end
+        end
+
+        context "when a 'security' label already exist" do
+          it "does not creates a 'security' label" do
+            labeler.create_default_labels_if_required
+
+            expect(labeler.labels_for_pr).to include("security")
+          end
+        end
+      end
+    end
+
     context "with GitLab details" do
       let(:source) do
         Dependabot::Source.new(provider: "gitlab", repo: "gocardless/bump")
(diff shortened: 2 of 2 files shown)


mburumaxwell:
@jeffwidman could you kindly have a look at this?

mmoreno79:
Would really like to see this PR merged to get labels/tags working with azure repositories.

jeffwidman (commented):
So I looked at this a bit a few weeks ago... while this change is clearly low risk in terms of LOC, I'm not excited that it's letting `"azure"` infiltrate the general interface...

I'll try to take another look later this week and see if there's a cleaner way to do it from an architecture perspective...

jeffwidman:
And to be clear, I 100% support what you're trying to do here...  just want to try to avoid special cases leaking out from specific client implementations if possible.

mburumaxwell:
I understand the concern here. Open to suggestions of a more general approach even though my ruby experience is bare minimum as of now.

mburumaxwell:
> just want to try to avoid special cases leaking out from specific client implementations if possible.

Does this check https://github.com/dependabot/dependabot-core/blob/c5cd618812b07ece4a4b53ea18d80ad213b077e7/common/lib/dependabot/source.rb#L111 also qualify for a "special case"?

jeffwidman:
👋 I haven't forgotten about this, just haven't had time to really sit down and poke at the code a bit to see if there's other ways of restructuring this. I hope to get to it in January after the holidays.

jeffwidman:
Like https://github.com/dependabot/dependabot-core/pull/6320, I'd also like to get input from the ADO folks on this PR before merging it... just see what they're thinking re: labels.

jeffwidman (commented):
Didn't this PR used to be more complicated or am I mis-remembering things?

I'm still not at all excited about special casing things within the platform-agnostic code, but this is a pretty simple exception such that I'm making a mountain out of a molehill to block it on purity points alone. 🤦‍♂️ 

Long term, I'd like to extract the "pull request creators" code out of Dependabot into separate libraries that could be more easily maintained by folks who know ADO/Bitbucket/GitLab etc w/o needing ✅ from us at GitHub who simply don't have an easy way to test/dev against those other platforms. So letting this implementation leak makes that a little more difficult, but OTOH it also makes it clear how the generic interface might need tweaking in order to support the library idea.

On the product side, I'd like to get input from the ADO folks on whether centralized labels are on their roadmap... if so, we can later add a TODO comment here to pull this out once that ships. And if they did that long term, it'd keep the generic interface simple.

Anyway... CI isn't happy... I think it's just a ruby syntax problem and the logic won't change... if so, I'm 👍 on merging once you fix that.

mburumaxwell:
CI issue resolved.

common/lib/dependabot/pull_request_creator/labeler.rb
```diff
@@ -174,7 +174,10 @@ def create_default_language_label_if_required
       end
 
       def default_labels_for_pr
-        if custom_labels then custom_labels & labels
+        if custom_labels
+          # Azure does not have centralised labels
+          custom_labels if source.provider == "azure"
```
deivid-rodriguez, line 179: Did you try this after the last push to fix CI? I think with the current logic the above line is going be ignored. It should be 

```suggestion
          return custom_labels if source.provider == "azure"
```

If the above is correct, we should probably add a spec at `common/spec/dependabot/pull_request_creator/labeler_spec.rb` that catches the bug.

deivid-rodriguez (commented):
I believe the implementation is now incorrect, can you double check @mburumaxwell?

jeffwidman:
Can you add a test to ensure we don't regress on this in the future?

mburumaxwell:
> Can you add a test to ensure we don't regress on this in the future?

Done

deivid-rodriguez (approved):
Looks awesome, thank you!

mburumaxwell:
> Looks awesome, thank you!

I see one E2E test failing. I cannot seem to see the re-run button from my end. What can be done about it? Or will you just merge?

deivid-rodriguez:
It does not seem related to your changes, so no worries, we'll take care of it.

jeffwidman (approved):
Thanks for the thoroughness on the tests, well done!

jeffwidman (approved):

PR #5785, opened 2022-09-22
[npm] Only return a chain if a node matches a vulnerable version

mctofu:
Without this check we can surface already fixed versions of the dependency as needing to be updated which confuses Dependabot.


honeyankit (approved):

bdragon (approved):
Nice fix. Returning an empty array if the node's not vulnerable looks right to me 👍

PR #5786, opened 2022-09-23
Update all versions of the same private module in single terraform file

szemek:
Hello,
I noticed that dependabot running for one our repositories had issues with updating one dependency.

```
updater | ERROR <job_468187453> No files changed!
updater | ERROR <job_468187453> /home/dependabot/terraform/lib/dependabot/terraform/file_updater.rb:44:in `updated_dependency_files'
updater | ERROR <job_468187453> /home/dependabot/dependabot-updater/lib/dependabot/updater.rb:746:in `generate_dependency_files_for'
updater | ERROR <job_468187453> /home/dependabot/dependabot-updater/lib/dependabot/updater.rb:311:in `check_and_create_pull_request'
updater | ERROR <job_468187453> /home/dependabot/dependabot-updater/lib/dependabot/updater.rb:109:in `check_and_create_pr_with_error_handling'
updater | ERROR <job_468187453> /home/dependabot/dependabot-updater/lib/dependabot/updater.rb:80:in `block in run'
updater | ERROR <job_468187453> /home/dependabot/dependabot-updater/lib/dependabot/updater.rb:80:in `each'
updater | ERROR <job_468187453> /home/dependabot/dependabot-updater/lib/dependabot/updater.rb:80:in `run'
updater | ERROR <job_468187453> /home/dependabot/dependabot-updater/lib/dependabot/update_files_job.rb:17:in `perform_job'
updater | ERROR <job_468187453> /home/dependabot/dependabot-updater/lib/dependabot/base_job.rb:50:in `run'
updater | ERROR <job_468187453> bin/update_files.rb:22:in `<main>'
```

After investigation I found out that when single terraform file uses the same private module multiple times, but with different versions, then only first occurrence is found and other are not updated.
This pull request is a bug fix for this situation.

### DIFF
--- a/terraform/lib/dependabot/terraform/file_updater.rb
+++ b/terraform/lib/dependabot/terraform/file_updater.rb
@@ -52,15 +52,17 @@ module Dependabot
       # To detect any changes in dependencies we need to overwrite an implementation from the base class
       #
       # Example (for simplicity other parameters are skipped):
-      # previous_requirements  = [{requirement: "0.9.1"}, {requirement: "0.11.0"}]
+      # previous_requirements = [{requirement: "0.9.1"}, {requirement: "0.11.0"}]
       # requirements = [{requirement: "0.11.0"}, {requirement: "0.11.0"}]
       #
       # Simple difference between arrays gives:
-      # requirements - previous_requirements = []
+      # requirements - previous_requirements
+      #  => []
       # which loses an information that one of our requirements has changed.
       #
       # By using symmetric difference:
-      # (requirements - previous_requirements) | (previous_requirements - requirements) = [{requirement: "0.9.1"}]
+      # (requirements - previous_requirements) | (previous_requirements - requirements)
+      #  => [{requirement: "0.9.1"}]
       # we can detect that change.
       def requirement_changed?(file, dependency)
         changed_requirements =
--- a/terraform/lib/dependabot/terraform/file_updater.rb
+++ b/terraform/lib/dependabot/terraform/file_updater.rb
@@ -89,7 +89,7 @@ module Dependabot
 
       def update_registry_declaration(new_req, old_req, updated_content)
         regex = new_req[:source][:type] == "provider" ? provider_declaration_regex : registry_declaration_regex
-        updated_content.sub!(regex) do |regex_match|
+        updated_content.gsub!(regex) do |regex_match|
           regex_match.sub(/^\s*version\s*=.*/) do |req_line_match|
             req_line_match.sub(old_req[:requirement], new_req[:requirement])
           end
--- a/terraform/lib/dependabot/terraform/file_updater.rb
+++ b/terraform/lib/dependabot/terraform/file_updater.rb
@@ -46,6 +46,14 @@ module Dependabot
         updated_files
       end
 
+      def requirement_changed?(file, dependency)
+        changed_requirements =
+          (dependency.requirements - dependency.previous_requirements) |
+          (dependency.previous_requirements - dependency.requirements)
+
+        changed_requirements.any? { |f| f[:file] == file.name }
+      end
+
       private
 
       def updated_terraform_file_content(file)
--- a/terraform/lib/dependabot/terraform/file_updater.rb
+++ b/terraform/lib/dependabot/terraform/file_updater.rb
@@ -48,6 +48,20 @@ module Dependabot
 
       private
 
+      # Terraform allows to use a module from the same source multiple times
+      # To detect any changes in dependencies we need to overwrite an implementation from the base class
+      #
+      # Example (for simplicity other parameters are skipped):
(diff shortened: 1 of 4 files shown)


Nishnha:
Hi @szemek thanks for this contribution! 

The symmetric difference check for `requirement_changed?` was surprising!

Subtracting two arrays does not do what I expected. For two arrays A and B, `A-B` will remove all instances of elements in array B from array A.

So for two arrays
```
A = [{:requirement=>"0.9.1", :groups=>[], :file=>"main.tf", :source=>{:type=>"registry", :registry_hostname=>"app.terraform.io", :module_identifier=>"example-org-5d3190/s3-webapp/aws"}},
   {:requirement=>"0.11.0", :groups=>[], :file=>"main.tf", :source=>{:type=>"registry", :registry_hostname=>"app.terraform.io", :module_identifier=>"example-org-5d3190/s3-webapp/aws"}}]
```

and 

```
B = [{:requirement=>"0.11.0", :groups=>[], :file=>"main.tf", :source=>{:type=>"registry", :registry_hostname=>"app.terraform.io", :module_identifier=>"example-org-5d3190/s3-webapp/aws"}},
   {:requirement=>"0.11.0", :groups=>[], :file=>"main.tf", :source=>{:type=>"registry", :registry_hostname=>"app.terraform.io", :module_identifier=>"example-org-5d3190/s3-webapp/aws"}}]
```

subtracting the two from each other will give two different results:

```
A - B = []
```

while 

```
B - A =
{:requirement=>"0.9.1", :groups=>[], :file=>"main.tf", :source=>{:type=>"registry", :registry_hostname=>"app.terraform.io", :module_identifier=>"example-org-5d3190/s3-webapp/aws"}}
```


I think this is a special case that only applies to Terraform registry dependencies with the same source and different versions. Most other ecosystems will force you to use a single version of a dependency.

Also, this may resolve https://github.com/dependabot/dependabot-core/issues/5128,
but https://github.com/dependabot/dependabot-core/issues/5150 is an issue with Terraform Git dependencies not having unique identifiers.

Nishnha (approved):

szemek:
> Hi @szemek thanks for this contribution!
> 
> [...]
>
> I think this is a special case that only applies to Terraform registry dependencies with the same source and different versions. Most other ecosystems will force you to use a single version of a dependency.
> 
> Also, this may resolve #5128, but #5150 is an issue with Terraform Git dependencies not having unique identifiers.

Thanks @Nishnha for checking my changes!
Yes, I noticed that this behaviour is Terraform-specific. At the beginning I was trying to update `requirement_changed?` method in base class, but it broke some tests for other ecosystem, so I moved it to subclass.

When it comes to Git dependencies, I can see that there is a different method for that 
https://github.com/dependabot/dependabot-core/blob/12bc6f133174fa171b89428734859990e4a37f23/terraform/lib/dependabot/terraform/file_updater.rb#L76-L88
so I'll prepare a new pull request with tests.

My pull request is approved but my branch is out-of-date with the `main` branch. Should I merge/rebase my branch, or leave that to you?

jeffwidman:
> My pull request is approved but my branch is out-of-date with the main branch. Should I merge/rebase my branch, or leave that to you?

We can handle that since you've given maintainers edit permission on your PR branch. We generally follow a deploy-then-merge process, so whoever gets to it will handle rebasing right before deploying.

terraform/lib/dependabot/terraform/file_updater.rb
```diff
@@ -48,6 +48,14 @@ def updated_dependency_files
 
       private
 
+      def requirement_changed?(file, dependency)
```
jeffwidman, line 51: Can you add a code comment explaining why Terraform requires overriding the base class implementation?
Esp given that the implementation is a little surprising... it may be enough to very briefly summarize https://github.com/dependabot/dependabot-core/pull/5786#issuecomment-1262533433 and then link back to it. I realize that's also accessible via `git blame`, but I think a dedicated code comment for the _why_ would be really useful.

Nishnha:
> so I'll prepare a new pull request with tests.

Thanks!

> When it comes to Git dependencies, I can see that there is a different method for that

Here's how we handle that Git dependency vs "native" dependency in NPM

https://github.com/dependabot/dependabot-core/blob/main/npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser.rb#L41-L50

terraform/lib/dependabot/terraform/file_updater.rb
szemek: @jeffwidman sure, I added an explanation in the newest changes. Feel free to add any suggestions and perform any updates to it on your own.

jeffwidman: This was a fantastic comment, great example / wording. ❤️

PR #5796, opened 2022-09-26
[npm] Only shortcut search when non-vuln version of advisory dep is found

mctofu:
There can be other deps marked as vulnerable that aren't the advisory dep (locking parents) and we can't be sure that a non-vulnerable version of those deps doesn't still have a vulnerable version of the advisory dep as a child.

For ex for an advisory on C:
```
A -> B (vuln) -> C (vuln)
| -> D -> B(not vuln) -> C (vuln)
```

I haven't found real examples of this occurring yet but this is worth fixing since it could lead to PRs that don't fully fix an alert.


bdragon (approved):
Took me a second to wrap my brain around this, but makes sense 👍

PR #5800, opened 2022-09-28
[npm] Only include VulnerabilityAuditor explanation if previously invoked

mctofu:
Minor optimization to only invoke the VulnerabilityAuditor if it's been previously used by the job. This is to avoid unnecessary invocation of the VulnerabilityAuditor if not required. For example, a security update to a direct dependency with no patched version available.


jeffwidman (approved):

deivid-rodriguez (approved):
Makes sense to me! 👍

PR #5801, opened 2022-09-28
[npm] Consider all installed versions when checking if a dependency is still vulnerable

bdragon:
## Context

[`DependencySet`](https://github.com/dependabot/dependabot-core/blob/91112975d6748eb0d211dccd23ca3f97934b6d00/common/lib/dependabot/file_parsers/base/dependency_set.rb#L10) is used ubiquitously during the manifest file-parsing phase of an update. It offers set semantics on the basis of dependency name and it has the property that when multiple instances of the same dependency (by name) are added, each is combined into the one held by the set and the latter's version is set according to an [algorithm](https://github.com/dependabot/dependabot-core/blob/91112975d6748eb0d211dccd23ca3f97934b6d00/common/lib/dependabot/file_parsers/base/dependency_set.rb#L59-L87) that roughly favors the lowest version.

One issue with this behavior is that when multiple versions of a dependency are used in a project, it's possible for versions other than the one divined by the dependency set (usually the lowest version) to be affected by a security vulnerability. As a result, Dependabot [will report that an update is not needed](https://github.com/dependabot/dependabot-core/blob/91112975d6748eb0d211dccd23ca3f97934b6d00/updater/lib/dependabot/updater.rb#L225-L233) because the dependency is not vulnerable, which is of course incorrect and confusing for users.

npm/yarn is an ecosystem in which it is possible to have multiple installed versions of a single dependency, and we have seen the misleading update-not-needed behavior occur. The goal of this PR is to stop this behavior for npm. A separate PR will follow to actually attempt the update when the vulnerability is on a version other than the one decided by the dependency set.

## What's changing

[ea27677](https://github.com/dependabot/dependabot-core/pull/5801/commits/ea276774cbdb2b1ff97259ebd50b9298c57e1ac3) I refactored `DependencySet` so that internally it retains all added versions for each dependency it holds. To achieve this I changed the internal data structure from an array to a hash, and introduced a private helper class to store the versions and present a combined representation of the dependency using the same algorithm linked above. The existing public API and behavior remains unchanged and two new public methods are added to expose the new functionality, `#dependency_for_name` and `#all_versions_for_name`.

[63b2569](https://github.com/dependabot/dependabot-core/pull/5801/commits/63b2569085b561fa621e9172b9b33e75d9f193bb) When combining two dependency sets (a very common operation) care was taken to combine all tracked versions in the correct order to ensure that the combined representation of each dependency in the combined set would be the same as it is today.

[7967e2f](https://github.com/dependabot/dependabot-core/pull/5801/commits/7967e2fe7c0514aef28f3f483d39cdc6598cc371) To make it possible to associate a dependency with all installed versions of it detected during the file-parsing phase, I added a new `#metadata` field (not included in `#to_h` serialization) to `Dependency`.

[3ed012c](https://github.com/dependabot/dependabot-core/pull/5801/commits/3ed012c388a6f40a7ae07ba2d7afe28e1e2dd538) and [d8bcee7](https://github.com/dependabot/dependabot-core/pull/5801/commits/d8bcee73899ec9427025fa68cf50249c706ef4c9) Before returning the list of dependencies from the file parser, all installed versions of each dependency are injected into its metadata.

[c57b5d7](https://github.com/dependabot/dependabot-core/pull/5801/commits/c57b5d79c04b86045c1a690d586cd0033719ffeb) With the preceding changes in place, I modified the update checker so that it looks at all versions of a dependency when checking if it's vulnerable.

## Risks

`DependencySet` is a fairly central type, which makes this a fairly risky change. I'm placing a good deal of faith in our test suite and I appreciate your careful review!


common/lib/dependabot/file_parsers/base/dependency_set.rb
```diff
@@ -50,41 +60,101 @@ def case_sensitive?
...
-
-          # If we already have a requirement use the existing version
-          # (if present). Otherwise, use whatever the lowest version is
-          new_version =
-            if old_dep.requirements.any? then old_dep.version || new_dep.version
-            elsif !v_cls.correct?(new_dep.version) then old_dep.version
-            elsif !v_cls.correct?(old_dep.version) then new_dep.version
-            elsif v_cls.new(new_dep.version) > v_cls.new(old_dep.version)
-              old_dep.version
+        class DependencySlot
```
jurre, line 71: Can we add a short bit of documentation about what a `DependencySlot` is here? From my understanding right now something like:

```ruby
# DependencySlot represents a single Dependency with one or more versions of that dependency.
# This exists for ecosystems (like npm_and_yarn) where it's possible to pull in multiple versions of
# a dependency in a project. This way we can keep track of all those versions while performing an
# update.
```

jurre: Also could be in it's own file, but I don't feel super strongly about that

common/lib/dependabot/file_parsers/base/dependency_set.rb
```diff
@@ -50,41 +60,101 @@ def case_sensitive?
...
+          attr_reader :combined
+
+          def initialize
+            @all_versions = []
+            @combined = nil
+          end
+
+          def all_versions(sort:)
+            sort ? sorted : @all_versions.dup
```
jurre, line 80: Why do we need to dup all_versions only if we're not sorting? 🤔

common/lib/dependabot/file_parsers/base/dependency_set.rb
```diff
@@ -50,41 +60,101 @@ def case_sensitive?
...
+                        old_dep.version || new_dep.version
+                      elsif !version_class.correct?(new_dep.version)
+                        old_dep.version
+                      elsif !version_class.correct?(old_dep.version)
+                        new_dep.version
+                      elsif version_class.new(new_dep.version) > version_class.new(old_dep.version)
+                        old_dep.version
+                      else
+                        new_dep.version
+                      end
```
jurre, line 124: Could you add a short comment about what this combine "algorithm" does and (more importantly) why? It's not immediately clear to me why this works the way it does.

jurre:
Overall this approach looks great! If I understand this correctly, with this change we will detect that a given dependency is still vulnerable, but will we also be able to then update it to a non-vulnerable version, or is that something that we'd need to tackle as a follow-up?

common/lib/dependabot/file_parsers/base/dependency_set.rb
bdragon: Sorting produces a new array, so I wanted to mirror that behavior and make a shallow copy. Helps reduce the risk of a caller will accidentally mutating `@all_versions`.

common/lib/dependabot/file_parsers/base/dependency_set.rb
bdragon: It's not immediately clear to me either 😂 but yes, absolutely.

I carried this over verbatim from the [existing code](https://github.com/dependabot/dependabot-core/blob/91112975d6748eb0d211dccd23ca3f97934b6d00/common/lib/dependabot/file_parsers/base/dependency_set.rb#L63-L73) in order not to break anything. One thing I don't like about it is the fact that it will have different outcomes depending on the order in which versions of dependency are added—for example, if the first added has `#requirements`, its version will always win out even if it's the highest version, otherwise the lowest version will always win out—but not new behavior (and things broke in various ecosystems if I tried to modify that).

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser.rb
```diff
@@ -36,7 +37,8 @@ def parse
         dependency_set = DependencySet.new
         dependency_set += manifest_dependencies
         dependency_set += lockfile_dependencies
-        dependencies = dependency_set.dependencies
+
+        dependencies = Helpers.dependencies_with_all_versions_metadata(dependency_set)
 
         # TODO: Currently, Dependabot can't handle dependencies that have both
         # a git source *and* a non-git source. Fix that!
```
Nishnha, line 44: I think this is also an issue with the Terraform ecosystem because Dependencies using a Git source have a different internal identifier than Terraform registry dependencies.

Another thing that they may share in common now is, because npm and Terraform Dependencies can have multiple version requirements, the `requirements_changed?` method in Common needs to be overrided so that if one required version is already up-to-date, the `requirements_changed?` method still notices that the other required versions have been updated.

See https://github.com/dependabot/dependabot-core/pull/5786#issuecomment-1262533433 for more info

(this should be private)

```ruby
def requirement_changed?(file, dependency)
  changed_requirements =
    (dependency.requirements - dependency.previous_requirements) |
    (dependency.previous_requirements - dependency.requirements)

  changed_requirements.any? { |f| f[:file] == file.name }
end
```

common/lib/dependabot/file_parsers/base/dependency_set.rb
bdragon: Updated in ab9036b3cff2294b0f5306f4e4f563aa77317c87

common/lib/dependabot/file_parsers/base/dependency_set.rb
bdragon: Documentation added in ab9036b3cff2294b0f5306f4e4f563aa77317c87

common/lib/dependabot/file_parsers/base/dependency_set.rb
mctofu: > if the first added has `#requirements`, its version will always win out even if it's the highest version, otherwise the lowest version will always win out—but this isn't new behavior (and things broke in various ecosystems if I tried to modify that).

I think the intention there is that the version of a direct dependency should be preferred over transitive versions.

bdragon: Ah, thanks for that context. This should be using `Dependency#top_level?` then so it's clear that that's the intent. Otherwise it's not really clear (to me at least) that direct vs. transitive is what the code is concerned with here.

mctofu (approved):
Nice work!

To successfully get a PR I think theres a few more checks we need to investigate/fix in the updater like https://github.com/dependabot/dependabot-core/blob/12bc6f133174fa171b89428734859990e4a37f23/updater/lib/dependabot/updater.rb#L274-L282.

That work would be for a future PR but if we find that this change converts jobs that previously return "not vulnerable" to an "unknown error" or a PR that doesn't fully fix the vulnerability we should get those to return "security_update_not_possible" instead before shipping this.

common/lib/dependabot/file_parsers/base/dependency_set.rb
```diff
@@ -14,34 +14,44 @@ def initialize(dependencies = [], case_sensitive: false)
...
+            all_versions = other.all_versions_for_name(name)
+            all_versions.each { |dep| self << dep }
           end
 
           self
         end
 
-        def +(other)
-          raise ArgumentError, "must be a DependencySet" unless other.is_a?(DependencySet)
+        def all_versions_for_name(name, sort: false)
```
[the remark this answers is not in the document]
bdragon: Now that I'm looking at it, this still calls `#sort!` rather than `#sort` because the first pass wanted to mutate the array. This is a bug but it hasn't come up yet since nothing is using this code and is more than enough to convince me to delete the code!

common/spec/dependabot/file_parsers/base/dependency_set_spec.rb
```diff
@@ -234,4 +234,140 @@
...
+            foo_v1.requirements +
+            foo_sha.requirements +
+            foo_v1_1.requirements
+          ).uniq,
+          package_manager: "dummy"
+        )
+      )
+    end
+
+    it "returns all versions in the order they were added by default" do
```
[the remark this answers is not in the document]
bdragon: It does! although that's not exactly what's happening here—in this case the hash has one entry and the array of versions on that entry is returned in insertion order.

common/lib/dependabot/file_parsers/base/dependency_set.rb
bdragon: I removed the sort option and made `all_versions` an attr reader so this is moot now.

common/spec/dependabot/file_parsers/base/dependency_set_spec.rb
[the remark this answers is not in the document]
mctofu: 👍 Yup, this is a misplaced comment! I was wondering about the dependency order since it changed from an array to hash and incorrectly read this test description as checking for that.

No concerns about the dependency order though since the Hash is ordered!

PR #5805, opened 2022-09-28
Fix syntax error in Actions workflow file

bdragon:
Moves the if condition to the job.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -22,13 +22,13 @@ on:
       - 'CHANGELOG.md'
       - 'common/lib/dependabot/version.rb'
   schedule:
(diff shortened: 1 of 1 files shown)


honeyankit (approved):

jeffwidman (commented):
Thanks!

Curious btw, is there any way to test these? I couldn't figure out how to test it locally when I originally put up #5784 ... I guess I could have done it on my local repo using a much shorter cron period temporarily in my draft PR to see what got triggered...

bdragon:
I'm not sure if we have something more official but you could lint it using a JSON Schema validator and [this schema](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/github-workflow.json).

PR #5806, opened 2022-09-28
Update README with instructions for using the `dependabot` CLI for testing and development

mattt:
The [`dependabot` CLI](https://github.com/dependabot/cli) is a tool for testing and debugging Dependabot update jobs.

We've been using it internally for the past couple months, and it's made a big improvement to how quickly we can iterate and how confident we are making changes to `dependabot-core`.

If you've developed in the `dependabot-core` codebase, you're no doubt familiar with the `bin/dry-run` script. Until recently, that script provided the best approximation for understanding how a change in core affected the overall behavior of Dependabot update jobs. But it was always just that — an approximation. Whereas the dry-run script is Ruby code that approximates how Dependabot performs update jobs, the `dependabot` CLI uses the same containers we run in production, which lets us perfectly reproduce behavior in any environment. 

This PR updates the README to recommend the CLI as the primary way to test and debug `dependabot-core`. It's still a WIP as I go through and get a better understanding of current developer workflows, and how they can best be served by the new CLI tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,27 +58,53 @@ following commands in the cloned Git repository:
 
 You can read more about this in the [Git for Windows wiki](https://github.com/git-for-windows/git/wiki/Git-cannot-create-a-file-or-directory-with-a-long-path).
 
-## Setup
+## Running with the `dependabot` CLI
 
-To run all of Dependabot Core, you'll need Ruby, Python, PHP, Elixir, Node, Go,
-Elm, and Rust installed. However, if you just wish to run it for a single
-language you can get away with just having that language and Ruby.
+Some ecosystems in Dependabot Core have specific system dependencies.
+For example, running the `composer` ecosystem requires
+a particular version of `php` and other libraries to be installed.
+As such, we use containers to build and run Dependabot Core.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -58,28 +58,54 @@ following commands in the cloned Git repository:
...
+You can use the [`dependabot` CLI][dependabot-cli]
+to simulate a Dependabot update job for the provided ecosystem and repo.
+To test out the behavior of a development version of `dependabot-core`,
+build `Dockerfile.updater` and
+run the `dependabot update` subcommand
+passing the image tag to the `--updater-image` option.
+
+```console
+$ docker build . -f Dockerfile.updater \
+                 -t dependabot-updater-dev
```
Bo98, line 81: Needs `--build-arg OMNIBUS_VERSION=latest` or similar.

It's probably worth adding a step to build `dependabot-core` so that local changes can be reflected, particularly given that's how `docker-dev-shell` worked.

Bo98 (commented):
Tried this out to try reproduce an issue with a repo I had that I couldn't reproduce locally.

PR #5807, opened 2022-09-28
Set volumes as delegated mounts

Nishnha:
https://docker-docs.netlify.app/docker-for-mac/osxfs-caching/

Sets docker volume mounts for the dev-shell and dependabot CLI as `delegated` mounts, which uses the Container as the source of truth.

From the docs:
> When any container sharing a delegated mount terminates, changes to the mount must be written back to the host file system. If this writeback fails, the container’s execution must fail via exit code and/or Docker event channels.


Nishnha:
I haven't observed any real speed-ups since using a dependabot core container with delegated mounts. Closing this out

PR #5813, opened 2022-09-29
add dependabot CLI dev container

jakecoffman:
This allows us to quickly launch a Codespace and execute `script/dependabot` without having to figure out how to download the CLI.


honeyankit (commented):
Looks good to me!

PR #5817, opened 2022-09-30
Ensure always_clone is enabled for yarn_berry during file_fetching

jurre:
Previously we set the `always_clone` flag in the updater, but in prod we run file fetching as a separate step, and as a result the flag was not set at that point.

This change sets all the experiments on Dependabot::Experiment whenever we initialize a job. That way we can ensure that all the experiments are globally available during all steps.


brrygrdn (approved):
🚀

PR #5822, opened 2022-09-30
[npm] Allow updates with both top level and sub dependencies

mctofu:
Previously we wouldn't update sub dependencies if a top level dependency was also in the update list. With our new ability to update locking parents of a vulnerable transitive dependency it's now possible for an update to require updating both. This is resulting in incomplete PRs that include fixes for top level dependencies (and their child dependencies) but not fixes for separate sub dependencies that need updates.


honeyankit (approved):
Looks good to me!

bdragon (approved):

Issue #4761, opened 2022-02-22
Label for security PRs


JacobEvelyn:
I've reported this several times in the past, such as [here](https://github.com/dependabot/feedback/issues/493#issuecomment-497760182), [here](https://github.com/dependabot/dependabot-core/issues/1998), and [here](https://news.ycombinator.com/item?id=27933085).

We'd like a label for all PRs that Dependabot makes to address security fixes. The label could be customizable or just hardcoded to `security` as pre-GitHub-native Dependabot used to do. That way we could do a GitHub Issues search like `is:open is:pr label:security org:<our-org>` and see/merge them all quickly. (We also have a lot of automation we wrote for `security` PRs from Dependabot from before GitHub-native removed this functionality.) This is a pretty big pain point for us.

The impression I had from the conversation in the third link is that some solution here should be possible. Is that still the case?

Nishnha:
Hi @JacobEvelyn,

Thank you for your feedback!
We are now tracking this feature as part of a larger one internally around custom label creation for Dependabot.

Right now, the Dependabot team is unable to take on this work, but we will try to prioritize this and other community feature requests in the next quarter!

jeffwidman:
Related:
* #5825

jeffwidman:
We had some internal discussion on this, and the summary was:
> If users opt into adding labels on their PRs, that's on them and we don't mind supporting it. We just don't want to add that functionality by default because it can have the effect of disclosing vulnerabilities before maintainers are ready to address them.

However, from an implementation perspective, we want to avoid the business of managing labels wherever possible... it is a [morass of edge cases](https://github.com/dependabot/dependabot-core/issues?q=is%3Aopen+is%3Aissue+label%3A%22F%3A+labels%22).

It also makes it harder to keep PR creation logic separate from "generate the update" logic... rather keep that interface simple if possible (although it may be impossible here) so that we can keep it pluggable... long term, I'd like for us to extract out the GitLab, BitBucket, ADO, etc PR creators to separate plugins that the community can maintain. They are the ones who are impacted if a feature is missing/broken, so they'll best know what to add/how to test and we don't have the cycles to do that.

A better solution IMO for security labels is if we could figure out a way for https://github.com/dependabot/fetch-metadata/ to post-apply the label... But I don't think we expose the necessary data in a way that `fetch-metadata` can access it?

Perhaps we could wire in some metadata to:
* https://github.com/dependabot/dependabot-core/issues/4680

[...]

PR #5825, opened 2022-10-02
Mark security pull requests with respective vulnerability severity labels

andrcuns:
It would be rather useful to be able to see from the labels, what type of security vulnerability is being fixed by respective pr. It would allow to prioritise reviews and merging.

This PR adds ability for `Labeler` class to use hash of fixed vulnerabilities, fetch severities from it and add as labels to PR.


common/lib/dependabot/pull_request_creator.rb
```diff
@@ -248,7 +248,8 @@ def labeler
           includes_security_fixes: includes_security_fixes?,
           dependencies: dependencies,
           label_language: label_language?,
```
andrcuns, line 250: Should this feature be behind similar toggle?

common/lib/dependabot/pull_request_creator.rb
```diff
@@ -248,7 +248,8 @@ def labeler
           includes_security_fixes: includes_security_fixes?,
```
andrcuns, line 248: Technically this isn't needed since it can be inferred from `vulnerabilities` fixed, but I did not want to add backwards incompatible changes by removing it

common/lib/dependabot/pull_request_creator/labeler.rb
```diff
@@ -4,10 +4,16 @@
 require "dependabot/pull_request_creator"
 module Dependabot
   class PullRequestCreator
-    class Labeler
+    class Labeler # rubocop:disable Metrics/ClassLength
```
andrcuns, line 7: Not sure if we should try and split or shorten the class to not trigger this violation in this PR

sgrzemski:
+150 for this feature!

jeffwidman:
Thanks! 

At first glance this seems very useful/intuitive, but I want to doublecheck the issue backlog / discuss the product change with a few folks internally before merging as how we currently handle labels works for the majority of folks but causes some edge case issues: https://github.com/dependabot/dependabot-core/issues?q=is%3Aopen+is%3Aissue+label%3A%22F%3A+labels%22

At some point, I'm hoping we can re-evaluate how we handle labels, but that might be a bit down the road... so for now just want to ensure this doesn't paint ourselves into a corner... but I expect no problems and that this will be a welcome addition.

andrcuns:
One thing to note, it is probably safer to add a toggle, same as language label, and set it to false by default. This way it would minimise impact and allow for the feature to be used by external users right away.

jeffwidman:
We chatted about this internally, and I added a summary of the discussion here:
* https://github.com/dependabot/dependabot-core/issues/4761#issuecomment-1468970993

Let's continue the discussion there, in particular exploring if we could get Dependabot to expose metadata such that we can keep Dependabot itself out of the labels business... Let's keep this PR open because we might not be able to use metadata, in which case we'd move forward with this PR, but it's at least worth an exploration...

PR #5827, opened 2022-10-03
Yarn Berry: Run commands in `update-lockfile` mode

jurre:
Yarn Berry commands related to performing updates can run in a mode called `update-lockfile`, which was purpose-built for exactly what Dependabot does:

```
- `update-lockfile` will skip the link step altogether, and only fetch packages
  that are missing from the lockfile (or that have no associated checksums).
  This mode is typically used by tools like Renovate or Dependabot to keep a
  lockfile up-to-date without incurring the full install cost.
```

This makes yarn berry updates _significantly_ more performant, but it has the downside that when the yarn cache is committed to the repo, old entries in there are not cleaned up. Everything still works as expected when this happens, but over time it could lead to that cache folder getting larger than is necessary.

I've opened an issue about this upstream:
https://github.com/yarnpkg/berry/issues/4886, not entirely sure if this is a bug or just a consequence of how this works.

For now I think it's worth the trade-off, as many projects don't commit the cache folder to repo, and the performance benefits are very significant.

I'll try to find some time to address this upstream.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
@@ -161,7 +161,7 @@ module Dependabot
             # updated to a single new version, so we just pick the first one.
             "#{dep[:name]}@#{dep[:requirements].first[:requirement]}"
           end
-          command = "yarn add #{updates.join(' ')}"
+          command = "yarn add #{updates.join(' ')} --mode=update-lockfile"
           Helpers.run_yarn_commands(command)
           { yarn_lock.name => File.read(yarn_lock.name) }
         end
@@ -171,9 +171,9 @@ module Dependabot
           update = "#{dep.name}@#{dep.version}"
 
           Helpers.run_yarn_commands(
-            "yarn add #{update}",
-            "yarn dedupe #{dep.name}",
-            "yarn remove #{dep.name}"
+            "yarn add #{update} --mode=update-lockfile",
+            "yarn dedupe #{dep.name} --mode=update-lockfile",
+            "yarn remove #{dep.name} --mode=update-lockfile"
           )
           { yarn_lock.name => File.read(yarn_lock.name) }
         end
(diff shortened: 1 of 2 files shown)


npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
```diff
@@ -171,9 +171,9 @@ def run_yarn_berry_subdependency_updater(yarn_lock:)
           update = "#{dep.name}@#{dep.version}"
 
           Helpers.run_yarn_commands(
-            "yarn add #{update}",
-            "yarn dedupe #{dep.name}",
-            "yarn remove #{dep.name}"
+            "yarn add #{update} --mode=update-lockfile",
+            "yarn dedupe #{dep.name} --mode=update-lockfile",
+            "yarn remove #{dep.name} --mode=update-lockfile"
```
jurre, line 176: Opted to pass the argument vs automatically adding it in `run_yarn_commands`, since not all commands take that option and I figured it would lead to confusing errors down the line when using `run_yarn_commands` with one of those commands.

brrygrdn: Yeah, this seems like the right trade-off 👍🏻

jurre:
On a large-ish repo:

```
time yarn add @actions/core@1.10.0
yarn add @actions/core@1.10.0  9.02s user 3.63s system 51% cpu 24.357 total
```

vs

```
time yarn add @actions/core@1.10.0 --mode=update-lockfile
yarn add @actions/core@1.10.0 --mode=update-lockfile  2.50s user 0.25s system 74% cpu 3.674 total
```

Another option could be to detect if there is a yarn cache folder present, and if there is not, then we use the `update-lockfile` mode. From the logs it does appear that the most costly step is fetching all the missing packages, which we would not have to do when there is a cache folder committed to the repo.

brrygrdn (approved):

brrygrdn:
> Another option could be to detect if there is a yarn cache folder present, and if there is not, then we use the update-lockfile mode.

I like this idea, I'm not sure we need to block on adding this since we are in beta and as you say, committing the cache is potentially not something many projects do 🤔

jurre:
> > Another option could be to detect if there is a yarn cache folder present, and if there is not, then we use the update-lockfile mode.
> 
> I like this idea, I'm not sure we need to block on adding this since we are in beta and as you say, committing the cache is potentially not something many projects do 🤔

It is [recommended](https://yarnpkg.com/features/offline-cache#sharing-the-cache) by yarn:

> Starting from Yarn v2, Yarn will by default configure the cache to be local to your project. This is done to make it easier for you to store it as part of your repository, which we believe is the best way to ensure that your projects can still be installed just fine regardless of the availability of your package registries.

I don't have any numbers, but anecdotally I've seen more repo's that don't have it in version control

PR #5833, opened 2022-10-05
Yarn Berry: run git lfs pull before running yarn

jtbandes:
cc @jurre  @defunctzombie
Run `git lfs pull --include .yarn` before using yarn for dependency updating. Yarn plugins stored with LFS can now be used.

Resolves LFS issues described in https://github.com/dependabot/dependabot-core/issues/1297, see example Dependabot PR at https://github.com/foxglove/studio/pull/4528

Tested with `bin/dry-run.rb npm_and_yarn foxglove/studio --clone --dep rc-tree`, which now updates yarn.lock instead of failing with an error.

Ran unit test with `cd npm_and_yarn && bundle exec rspec spec/dependabot/npm_and_yarn/file_fetcher_spec.rb -e "with .yarn data stored in git-lfs"`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -317,4 +317,5 @@ ENV PATH="$HOME/bin:$PATH"
 # Configure cargo to use git CLI so the above takes effect
 RUN mkdir -p ~/.cargo && printf "[net]\ngit-fetch-with-cli = true\n" >> ~/.cargo/config.toml
 # Disable automatic pulling of files stored with Git LFS
+# This avoids downloading large files not necessary for the dependabot scripts
 ENV GIT_LFS_SKIP_SMUDGE=1
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,7 @@ RUN apt-get update \
     build-essential \
     dirmngr \
     git \
+    git-lfs \
     bzr \
     mercurial \
     gnupg2 \
@@ -315,3 +316,5 @@ RUN curl -sL $SHIM -o git-shim.tar.gz && mkdir -p ~/bin && tar -xvf git-shim.tar
 ENV PATH="$HOME/bin:$PATH"
 # Configure cargo to use git CLI so the above takes effect
 RUN mkdir -p ~/.cargo && printf "[net]\ngit-fetch-with-cli = true\n" >> ~/.cargo/config.toml
+# Disable automatic pulling of files stored with Git LFS
+ENV GIT_LFS_SKIP_SMUDGE=1
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
@@ -30,6 +30,19 @@ module Dependabot
         "Repo must contain a package.json."
       end
 
+      # Overridden to pull any yarn data or plugins which may be stored with Git LFS.
+      def _clone_repo_contents(target_directory:)
+        path = super(target_directory: target_directory)
+        begin
+          SharedHelpers.with_git_configured(credentials: credentials) do
+            SharedHelpers.run_shell_command("git -C #{path} lfs pull --include .yarn")
+            path
+          end
+        rescue StandardError
+          path
+        end
+      end
+
       private
 
       def fetch_files
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
@@ -38,11 +38,7 @@ module Dependabot
         begin
           SharedHelpers.with_git_configured(credentials: credentials) do
             Dir.chdir(path) do
-              cache_dir = if File.exist?(".yarnrc.yml")
-                            YAML.load_file(".yarnrc.yml").fetch("cacheFolder", "./.yarn/cache")
-                          else
-                            "./.yarn/cache"
-                          end
+              cache_dir = Helpers.fetch_yarnrc_yml_value("cacheFolder", "./yarn/cache")
               SharedHelpers.run_shell_command("git lfs pull --include .yarn,#{cache_dir}")
             end
             path
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
@@ -30,20 +30,27 @@ module Dependabot
         "Repo must contain a package.json."
       end
 
+      private
+
       # Overridden to pull any yarn data or plugins which may be stored with Git LFS.
       def _clone_repo_contents(target_directory:)
         path = super(target_directory: target_directory)
         begin
           SharedHelpers.with_git_configured(credentials: credentials) do
-            SharedHelpers.run_shell_command("git -C #{path} lfs pull --include .yarn")
+            Dir.chdir(path) do
+              cache_dir = if File.exist?(".yarnrc.yml")
+                            YAML.load_file(".yarnrc.yml").fetch("cacheFolder", "./.yarn/cache")
+                          else
+                            "./.yarn/cache"
+                          end
+              SharedHelpers.run_shell_command("git lfs pull --include .yarn,#{cache_dir}")
+            end
             path
           end
         rescue StandardError
           path
         end
       end
-
-      private
 
       def fetch_files
         fetched_files = []
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
@@ -30,23 +30,25 @@ module Dependabot
         "Repo must contain a package.json."
       end
 
-      private
-
       # Overridden to pull any yarn data or plugins which may be stored with Git LFS.
-      def _clone_repo_contents(target_directory:)
-        path = super(target_directory: target_directory)
+      def clone_repo_contents
+        return @git_lfs_cloned_repo_contents_path if defined?(@git_lfs_cloned_repo_contents_path)
+
+        @git_lfs_cloned_repo_contents_path = super
         begin
           SharedHelpers.with_git_configured(credentials: credentials) do
-            Dir.chdir(path) do
+            Dir.chdir(@git_lfs_cloned_repo_contents_path) do
               cache_dir = Helpers.fetch_yarnrc_yml_value("cacheFolder", "./yarn/cache")
               SharedHelpers.run_shell_command("git lfs pull --include .yarn,#{cache_dir}")
             end
-            path
+            @git_lfs_cloned_repo_contents_path
           end
         rescue StandardError
-          path
+          @git_lfs_cloned_repo_contents_path
         end
       end
+
+      private
 
       def fetch_files
         fetched_files = []
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
@@ -55,6 +55,8 @@ module Dependabot
         def updated_yarn_lock(yarn_lock)
           base_dir = dependency_files.first.directory
           SharedHelpers.in_a_temporary_repo_directory(base_dir, repo_contents_path) do
+            # Pull any yarn data or plugins which may be stored with Git LFS
+            SharedHelpers.run_shell_command("git lfs pull --include .yarn")
             write_temporary_dependency_files
             lockfile_name = Pathname.new(yarn_lock.name).basename.to_s
             path = Pathname.new(yarn_lock.name).dirname.to_s
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
@@ -55,8 +55,6 @@ module Dependabot
         def updated_yarn_lock(yarn_lock)
           base_dir = dependency_files.first.directory
           SharedHelpers.in_a_temporary_repo_directory(base_dir, repo_contents_path) do
-            # Pull any yarn data or plugins which may be stored with Git LFS
-            SharedHelpers.run_shell_command("git lfs pull --include .yarn")
             write_temporary_dependency_files
             lockfile_name = Pathname.new(yarn_lock.name).basename.to_s
             path = Pathname.new(yarn_lock.name).dirname.to_s
--- a/npm_and_yarn/spec/dependabot/npm_and_yarn/file_fetcher_spec.rb
+++ b/npm_and_yarn/spec/dependabot/npm_and_yarn/file_fetcher_spec.rb
@@ -70,8 +70,8 @@ RSpec.describe Dependabot::NpmAndYarn::FileFetcher do
         directory: directory
       )
     end
-    let(:url) { "https://api.github.com/repos/jtbandes/dependabot-yarn-lfs-fixture/contents/" }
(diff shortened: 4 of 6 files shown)


Dockerfile
```diff
@@ -312,3 +313,5 @@ RUN curl -sL $SHIM -o git-shim.tar.gz && mkdir -p ~/bin && tar -xvf git-shim.tar
 ENV PATH="$HOME/bin:$PATH"
 # Configure cargo to use git CLI so the above takes effect
 RUN mkdir -p ~/.cargo && printf "[net]\ngit-fetch-with-cli = true\n" >> ~/.cargo/config.toml
+# Disable automatic pulling of files stored with Git LFS
```
defunctzombie, line 316:
```suggestion
# Disable automatic pulling of files stored with Git LFS
# This avoids downloading large files not necessary for the dependabot scripts
```

defunctzombie:
Besides storing pre-built plugins using LFS. Another good use case for LFS with yarn projects is the offline cache: https://yarnpkg.com/features/offline-cache

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
```diff
@@ -56,6 +56,8 @@ def sub_dependencies
         def updated_yarn_lock(yarn_lock)
           base_dir = dependency_files.first.directory
           SharedHelpers.in_a_temporary_repo_directory(base_dir, repo_contents_path) do
+            # Pull any yarn data or plugins which may be stored with Git LFS
+            SharedHelpers.run_shell_command("git lfs pull --include .yarn")
```
jurre, line 60: We wouldn't want to perform this step here, as this gets called for every dependency that gets updated, which would result in a lot of unnecessary work.

If we wanted to do this we'd have to do it here: https://github.com/dependabot/dependabot-core/blob/86b4fdf9054552c0ffc4b8ff145f0e65cf4a7473/common/lib/dependabot/file_fetchers/base.rb#L507, or override that method for the `Dependabot::NpmAndYarn::FileFetcher`.

We would probably want to inspect the yarn config in order to determine what path to pull, as the cache and plugin folders are configurable I believe

jurre:
Thanks for opening a PR so we have a place to discuss the approach.

I am curious if this setup is a common one, do you know of more projects following this approach of storing yarn assets in LFS?

jtbandes:
> I am curious if this setup is a common one, do you know of more projects following this approach of storing yarn assets in LFS?

You can check out https://github.com/renovatebot/renovate/issues/6842 for some other people who are interested in LFS support for bumping dependencies.

> We would probably want to inspect the yarn config in order to determine what path to pull, as the cache and plugin folders are configurable I believe

This is a good point. However, it seems like it would add a lot of complexity to this otherwise simple patch. Do you think it would make more sense to add a config option to dependabot.yml that allows the user to specify which paths need to be `git lfs pull`ed? I would guess that for most yarn users, the default of pulling only `.yarn` would suffice. Would it be reasonable to only pull `.yarn` for now and wait to see if users actually request the behavior of pulling arbitrary custom paths?

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
jtbandes: I've updated to override `_clone_repo_contents` instead of doing this inside of `updated_yarn_lock`. See below for a question about the yarn config.

defunctzombie:
Not having a cache does seem to allow yarn to continue unlike not having the plugins.

FWIW there are open issues on the yarn repo around installing plugins automatically: https://github.com/yarnpkg/berry/issues/4464 but no PR has been accepted yet.

jurre:
> Do you think it would make more sense to add a config option to dependabot.yml that allows the user to specify which paths need to be git lfs pulled?

No we're really trying to keep the config file as simple as possible and avoid any ecosystem specific configuration, especially if like in this case the setting can be inferred from the repo.

We have a bit of code that checks the config file for the cache path already, so I _think_ we could just run some similar code in the repo content path we just fetched, parse the yaml and check what the correct path to pull is?

https://github.com/dependabot/dependabot-core/blob/ca5756f24954c47c180b7ac2b7ce456c9cbec997/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater.rb#L76

> Would it be reasonable to only pull .yarn for now and wait to see if users actually request the behavior of pulling arbitrary custom paths?

That doesn't sound unreasonable, but it would be nice if we can make it work out of the box?

> Not having a cache does seem to allow yarn to continue unlike not having the plugins.

And plugins always go in `.yarn/plugins`? Not pulling the yarn cache correctly would cause some broken behavior I think though, because Dependabot now wouldn't add the right entry to the yarn cache.

Overall I think the approach here seems reasonable and I can't see it leading to any problems so I'd be inclined to accept this , it would be really nice if you could take a look at inferring the yarn cache path from the config file

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -31,6 +31,19 @@ def self.required_files_message
         "Repo must contain a package.json."
       end
 
+      # Overridden to pull any yarn data or plugins which may be stored with Git LFS.
+      def _clone_repo_contents(target_directory:)
```
jurre, line 35: This should go under the `private` keyword just below

jtbandes:
@jurre any further thoughts on this change? I've implemented some of your feedback, please let me know if there are any more changes you'd like to se..

jurre:
Hey, so first of all thanks for those changes, I think this should do the trick!

One last thing I would like to request is that we add some tests around this, here is an example of the existing tests we have around cloning: https://github.com/dependabot/dependabot-core/blob/4bd71272f49df0bddf439f588e5d2aadb8aff424/common/spec/dependabot/file_fetchers/base_spec.rb#L1323-L1348

Let me know if you need any help setting that up!

jtbandes:
Yeah, some guidance around this would be helpful, since I am not familiar with rspec or this repo's test organization. For instance, are there any existing tests for file fetchers of specific package ecosystems? If I were to add this one would you imagine adding it to the base_spec.rb you linked, or making a new spec file specific to npm_and_yarn? Would you like to see a fixture that actually includes data stored in LFS or just a test that verifies that the `git lfs pull` command is run?

jtbandes:
@jurre Yeah, some guidance around this would be helpful, since I am not familiar with rspec or this repo's test organization. For instance, are there any existing tests for file fetchers of specific package ecosystems? If I were to add this one would you imagine adding it to the base_spec.rb you linked, or making a new spec file specific to npm_and_yarn? Would you like to see a fixture that actually includes data stored in LFS or just a test that verifies that the `git lfs pull` command is run?

jurre:
> For instance, are there any existing tests for file fetchers of specific package ecosystems?

Yep, they all have ecosystem specific file fetcher specs, they'll be in `<ecosystem>/spec/dependabot/<ecosystem>/file_fetcher_spec.rb`. The go_mod ecosystem is somewhat relevant since it also uses a cloning strategy (vs using the GitHub API which we do for most ecosystems), but we don't perform any checks around this.

> If I were to add this one would you imagine adding it to the base_spec.rb you linked, or making a new spec file specific to npm_and_yarn? 

I'd add it to the npm_and_yarn one: https://github.com/dependabot/dependabot-core/blob/8d60d39e19bca9ba738e497fae31fbe0323386f3/npm_and_yarn/spec/dependabot/npm_and_yarn/file_fetcher_spec.rb

> Would you like to see a fixture that actually includes data stored in LFS or just a test that verifies that the git lfs pull command is run?

I think the best way to go would be to set up a fixture that includes lfs data, and then verify that it is included in the fetched files

Here's a snippet of useful context on when the file fetcher uses the cloning strategy: https://github.com/dependabot/dependabot-core/blob/4bd71272f49df0bddf439f588e5d2aadb8aff424/common/lib/dependabot/file_fetchers/base.rb#L38-L42, so that means that when passing the FileFetcher a `repo_contents_path`, it'll clone the repo, otherwise use the API. In the case of Yarn Berry, we already make sure to always set that when actually running Dependabot, but for these tests you'll have to pass it a value (a tmpdir works best) manually

jtbandes:
Thanks for the pointers. Is it okay for the test to require network access? Otherwise I'm not sure if it will be possible to set up a repo that uses LFS but the files are hosted fully locally (probably possible, maybe not easy).

jurre:
>  Is it okay for the test to require network access?

If we can avoid it then that'll be preferred, but if there's no way around it we do have some other tests accessing the network. Our existing cloning specs do use a local repo I believe, so hopefully we can repurpose that!

jtbandes:
Since the whole point of LFS is that git is supposed to fetch the large files from a file server, making it work offline may require a custom adapter like this: https://github.com/sinbad/lfs-folderstore

jurre:
Ah I see, well, if it's not feasible to do it locally then we can pull from a remote. I'll move the repo into a dedicated org we have for these sort of fixtures once you have a working setup.

Thanks for working on this by the way! 🎉 🙇

jtbandes:
@jurre I've created a fixture repo at https://github.com/jtbandes/dependabot-yarn-lfs-fixture and added the test. Please critique! 😃

jurre:
Looks good at a cursory glance! It's getting a bit late on my end so I'll give this a more thorough look tomorrow 🙇

jtbandes:
@jurre just a quick ping on this, let me know if there is anything else I can do. Thanks for your repeated reviews!

npm_and_yarn/spec/dependabot/npm_and_yarn/file_fetcher_spec.rb
```diff
@@ -62,6 +62,41 @@
...
+  context "with .yarn data stored in git-lfs" do
+    let(:source) do
+      Dependabot::Source.new(
+        provider: "github",
+        repo: repo,
+        directory: directory
+      )
+    end
+    let(:url) { "https://api.github.com/repos/jtbandes/dependabot-yarn-lfs-fixture/contents/" }
+    let(:repo) { "jtbandes/dependabot-yarn-lfs-fixture" }
```
jurre, line 74:
```suggestion
    let(:repo) { "dependabot-fixtures/yarn-berry-lfs-fixture" }
```

jtbandes:
@jurre The new repo does not have the files properly tracked with Git LFS, so they're just text files: https://github.com/dependabot-fixtures/yarn-berry-lfs-fixture/blob/main/.yarn/releases/yarn-3.2.4.cjs (compare with https://github.com/jtbandes/dependabot-yarn-lfs-fixture/blob/main/.yarn/releases/yarn-3.2.4.cjs)

  Luckily the test caught it 😃  I'll re-initiate the transfer so you can just use my repo.

npm_and_yarn/spec/dependabot/npm_and_yarn/file_fetcher_spec.rb
```diff
@@ -62,6 +62,41 @@
...
+  context "with .yarn data stored in git-lfs" do
+    let(:source) do
+      Dependabot::Source.new(
+        provider: "github",
+        repo: repo,
+        directory: directory
+      )
+    end
+    let(:url) { "https://api.github.com/repos/dependabot-fixtures/yarn-berry-lfs-fixture/contents/" }
+    let(:repo) { "dependabot-fixtures/yarn-berry-lfs-fixture" }
```
jurre, line 74:
```suggestion
    let(:url) { "https://api.github.com/repos/dependabot-fixtures/dependabot-yarn-lfs-fixture/contents/" }
    let(:repo) { "dependabot-fixtures/dependabot-yarn-lfs-fixture" }
```

jtbandes:
Looks like the rest of the test failures are some intermittent network issues.

jtbandes:
Still looks like sporadic network failures to me: 
> curl: (56) OpenSSL SSL_read: Connection reset by peer, errno 104

However, if you think any of these errors are related to my change, please let me know and I can try to help debug.

I would definitely recommend [retrying only the jobs that failed](https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs#re-running-failed-jobs-in-a-workflow) if possible, so the ones that succeeded don't fail later 😅

jtbandes:
@pavera Would you have some time to take a look at this and provide some feedback? Thank you 😄  🙌

jurre:
> However, if you think any of these errors are related to my change, please let me know and I can try to help debug.

I don't think they are!

jtbandes:
Finally 🟢  😁

jurre:
Have asked @pavera to take a look at this with fresh eyes to make sure I didn't miss anything

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -32,6 +32,26 @@ def self.required_files_message
...
+              cache_dir = if File.exist?(".yarnrc.yml")
+                            YAML.load_file(".yarnrc.yml").fetch("cacheFolder", "./.yarn/cache")
+                          else
+                            "./.yarn/cache"
+                          end
+              SharedHelpers.run_shell_command("git lfs pull --include .yarn,#{cache_dir}")
+            end
+            path
+          end
+        rescue StandardError
```
bdragon, line 50: Do we want to return the path and carry on for _any_ error?

jtbandes: What other options do you have in mind? I’m not sure what meaningful recovery could be done here…

bdragon: Let me first better understand, which errors are you anticipating?

Here are the ones I can think of:
* `Dependabot::SharedHelpers::HelperSubprocessFailed` raised by `run_shell_command` in `super` or here
* `NoMethodError` if `.yarnrc.yml` is empty, in which case `YAML.load_file` will return false, which does not respond to `fetch`
* `Psych::ParseError` if the YAML in the `.yarnrc.yml` is invalid

If we encounter an exception outside of these, I think we'd want to let it fail? @jurre might know better 😅

jurre: I think it's reasonable to continue without `lfs pull` in those cases rather than fail the update entirely?

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -32,6 +32,26 @@ def self.required_files_message
 
       private
 
+      # Overridden to pull any yarn data or plugins which may be stored with Git LFS.
+      def _clone_repo_contents(target_directory:)
```
bdragon, line 36: At least at one time in history, methods like this one in `Dependabot::FileFetchers::Base` were not meant to be overridden:

https://github.com/dependabot/dependabot-core/blob/6a9c745252f7ff8389751b58133c851df6207b79/common/lib/dependabot/file_fetchers/base.rb#L176-L178

This would be the first instance of a file fetcher subclass overriding it. Could we override public method `#clone_repo_contents` instead?

jurre:
I'll try to find a bit of time to make the suggestions @bdragon shared today and then hopefully we can get this merged. Thanks for your work and patience on this @jtbandes!

bdragon (approved):

pavera (approved):

jurre:
Thanks, this is now live

jtbandes:
Thanks for all your help!

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
jeffwidman: Related:
* https://github.com/dependabot/dependabot-core/pull/6583#issuecomment-1416056516

It took a bit to track down why this started failing on `main` since the error gets swallowed... that means users who are hitting this failure scenario won't be told why...

PR #5837, opened 2022-10-05
Consider all dependency versions in Job.vulnerable?

mctofu:
In testing https://github.com/dependabot/dependabot-core/pull/5801 via the updater we found there's an additional spot we need to consider all versions of a dependency when doing a vulnerability check. This is to prevent Dependabot from returning a confusing error message that all versions have been ignored.

This code path is invoked from the updater at https://github.com/dependabot/dependabot-core/blob/97ae98c8f7fbb308ff317be20f644d700f612258/updater/lib/dependabot/updater.rb#L231

Part of the allowed update check is to see if the dependency is actually still vulnerable:
https://github.com/dependabot/dependabot-core/blob/97ae98c8f7fbb308ff317be20f644d700f612258/updater/lib/dependabot/job.rb#L77-L78

I considered removing this check as it's a duplicate of this check in the create pr flow:
https://github.com/dependabot/dependabot-core/blob/97ae98c8f7fbb308ff317be20f644d700f612258/updater/lib/dependabot/updater.rb#L221-L222

However, it's also checked in the update pr flow so it would have taken more changes to try and sort that out so I'm leaving that for a future cleanup:
https://github.com/dependabot/dependabot-core/blob/97ae98c8f7fbb308ff317be20f644d700f612258/updater/lib/dependabot/updater.rb#L140-L141

### DIFF
--- a/common/lib/dependabot/dependency.rb
+++ b/common/lib/dependabot/dependency.rb
@@ -106,6 +106,15 @@ module Dependabot
       display_name_builder.call(name)
     end
 
+    # Returns all detected versions of the dependency. Only ecosystems that
+    # support this feature will return more than the current version.
+    def all_versions
+      all_versions = metadata[:all_versions]
+      return [version].compact unless all_versions
+
+      all_versions.filter_map(&:version)
+    end
+
     def ==(other)
       other.instance_of?(self.class) && to_h == other.to_h
     end
--- a/common/spec/dependabot/dependency_spec.rb
+++ b/common/spec/dependabot/dependency_spec.rb
@@ -303,6 +303,22 @@ RSpec.describe Dependabot::Dependency do
       )
       expect(dependency.to_h.keys).not_to include("metadata")
     end
+
+    it "isn't utilized by the equality operator" do
+      dependency1 = described_class.new(
+        name: "dep",
+        requirements: [],
+        package_manager: "dummy",
+        metadata: { foo: 42 }
+      )
+      dependency2 = described_class.new(
+        name: "dep",
+        requirements: [],
+        package_manager: "dummy",
+        metadata: { foo: 43 }
+      )
+      expect(dependency1).to eq(dependency2)
+    end
   end
 
   describe "#all_versions" do
--- a/common/spec/dependabot/dependency_spec.rb
+++ b/common/spec/dependabot/dependency_spec.rb
@@ -304,4 +304,54 @@ RSpec.describe Dependabot::Dependency do
       expect(dependency.to_h.keys).not_to include("metadata")
     end
   end
+
+  describe "#all_versions" do
+    it "returns an empty array by default" do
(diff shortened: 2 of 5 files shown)


common/lib/dependabot/dependency.rb
```diff
@@ -106,6 +106,15 @@ def display_name
...
     end
 
+    # Returns all detected versions of the dependency. Only ecosystems that
+    # support this feature will return more than the current version.
+    def all_versions
+      all_versions = metadata[:all_versions]
+      return [version].compact unless all_versions
+
+      all_versions.filter_map(&:version)
+    end
```
bdragon, line 116: 👌

updater/lib/dependabot/job.rb
```diff
@@ -112,8 +112,9 @@ def vulnerable?(dependency)
         version_class_for_package_manager(dependency.package_manager)
       return false unless version_class.correct?(dependency.version)
 
-      version = version_class.new(dependency.version)
-      security_advisories.any? { |a| a.vulnerable?(version) }
+      all_versions = dependency.all_versions.
+                     filter_map { |v| version_class.new(v) if version_class.correct?(v) }
+      security_advisories.any? { |a| all_versions.any? { |v| a.vulnerable?(v) } }
```
bdragon, line 117: I've noticed this code

```ruby
version_class = Utils.version_class_for_package_manager(package_manager)
version_class.new(version) if version_class.correct?(version)
```

in a lot of places. Looking at it once again here, I'm wondering if it wouldn't make sense to put this in `Dependency`. It would be a reasonable argument to say it's too much for that class to know about, but it would remove some boilerplate. Maybe something like this (but probably with better naming)

```ruby
def correct_version
  version_class = Utils.version_class_for_package_manager(package_manager)
  version_class.new(version) if version_class.correct?(version)
end

def all_correct_versions
  version_class = Utils.version_class_for_package_manager(package_manager)
  all_versions.filter_map { |v| version_class.new(v) if version_class.correct?(v) }
end
```

common/spec/dependabot/dependency_spec.rb
```diff
@@ -303,5 +303,71 @@
...
+        metadata: { foo: 42 }
+      )
+      dependency2 = described_class.new(
+        name: "dep",
+        requirements: [],
+        package_manager: "dummy",
+        metadata: { foo: 43 }
+      )
+      expect(dependency1).to eq(dependency2)
+    end
```
bdragon, line 321: Good thinking

bdragon (approved):
This looks great!

updater/lib/dependabot/job.rb
mctofu: Yeah, that's a good thought! There's still some future changes needed to get a successful PR through. If that ends up duplicating this again I think that'd be good opportunity to dry it up?

PR #5840, opened 2022-10-06
use configured global registry for library lookup

jakecoffman:
## Context

Dependabot honors an `.npmrc` or `.yarnrc` file to set the global registry. When a global registry is present in one of these files, Dependabot shouldn't make calls to registry.npmjs.org. This library check was doing just that!

## Approach

At first I tried using the RegistryFinder's `registry` method, but that method assumes you're giving it a dependency. The library check is checking the registry for the project itself, and thus won't be present in the dependencies of the project when RegistryFinder goes to look for it.

So rather than change the `registry` method, I exposed the `global_registry` method so I could use it in the library check. I also modified it to stop stripping the protocol, since it is valid for enterprises to have an internal registry that isn't secured by TLS. I put the stripping in the `first_registry_with_dependency_details` which was the only place it was used.


jakecoffman:
Hmm it's also worth considering scoped dependencies in this case. I might have to do another pass at this to search the rc files for scopes that match. There doesn't appear to be any scoped registry handling in the RegistryFinder.

honeyankit:
Just wanted to confirm that the global registry == private registry ?

jakecoffman:
@honeyankit yes typically, although conceivably it could point to anything. In npm and yarn, you can configure a registry globally or scoped. In npm that looks like:

```
registry=https://example.com               #global
@dependabot:registry=https://example.com   #scoped
```

pavera:
Are we concerned with changing the default behavior for all dependabot users? I think this seems like the more correct behavior in general, but this does change how dependabot currently behaves in a way users might not expect.  Should we put this behind a feature flag?

jakecoffman:
@pavera I'm not concerned. This code path only influences whether the update strategy is [widen ranges or bump versions](https://github.com/dependabot/dependabot-core/blob/ba241764a479746e044ac2afd1812ffbe2b9545b/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb#L95). Dependabot is already using the private registry for all the other calls.

So folks who don't use an .npmrc won't see a difference. Those that have a registry defined are likely to see more correct behavior, especially those who've published to the private registry, as we can now correctly identify a library that has been published there.

PR #5858, opened 2022-10-09
Regenerate `updater/Gemfile.lock`

jeffwidman:
I noticed the following log lines in [the output of building the `updater`](https://github.com/dependabot/dependabot-core/actions/runs/3212248230/jobs/5250946524) Docker image:
```
32 sha256:563a023320753425444761bd82772c641463b33ea716db65998c70bdc931dd95
32 1.167 Bundler 2.3.22 is running, but your lockfile was generated with 2.2.20. Installing Bundler 2.2.20 and restarting using that version.
32 2.192 Fetching gem metadata from https://rubygems.org/.
32 2.217 Fetching bundler 2.2.20
32 2.297 Installing bundler 2.2.20
32 2.511 Calling `DidYouMean::SPELL_CHECKERS.merge!(error_name => spell_checker)' has been deprecated. Please call `DidYouMean.correct_error(error_name, spell_checker)' instead.
32 3.308 Fetching gem metadata from https://rubygems.org/.........
```

That's probably because of
https://github.com/dependabot/dependabot-core/pull/5509 which updated `bundler` to `2.3.22` which doesn't match the version of Bundler specified in `Gemfile.lock`:
* https://bundler.io/blog/2022/01/23/bundler-v2-3.html
* https://github.com/dependabot/dependabot-core/blob/2ba96fcc09564675d94140b7f8ee6fefa32de935/updater/Gemfile.lock#L333

So I built the docker updater image (couldn't use the core image unfortunately because it doesn't mount `/updater`, and for good reason).

And then within that docker image ran:
```
$ bundler update --bundler
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -74,6 +74,7 @@ ARG RUBY_INSTALL_VERSION=0.8.3
 ARG RUBYGEMS_SYSTEM_VERSION=3.3.22
 
 ARG BUNDLER_V1_VERSION=1.17.3
+# When bumping Bundler, need to also regenerate `updater/Gemfile.lock` via `bundle update --bundler`
 ARG BUNDLER_V2_VERSION=2.3.22
 ENV BUNDLE_SILENCE_ROOT_WARNING=1
 # Allow gem installs as the dependabot user
--- a/updater/Gemfile.lock
+++ b/updater/Gemfile.lock
@@ -330,4 +330,4 @@ DEPENDENCIES
   webmock (~> 3.18)
 
 BUNDLED WITH
-   2.2.20
+   2.3.22


Dockerfile
```diff
@@ -74,6 +74,7 @@ ARG RUBY_INSTALL_VERSION=0.8.3
 ARG RUBYGEMS_SYSTEM_VERSION=3.3.22
 
 ARG BUNDLER_V1_VERSION=1.17.3
+# When bumping Bundler, need to also regenerate `updater/Gemfile.lock` via `bundle update --bundler`
```
jeffwidman, line 77: This command resulted in bumping a bunch of lines beyond the Bundler version... I wasn't sure if that's because those are what our current Gemfile's resolve to, or because those happen to be the newest versions on RubyGems and regardless of whether we've upgraded Bundler still tries to pull those in? IE, is there a command that _only_ regenerates w/o pulling in new versions of other stuff?

Also, I added a code comment because I couldn't come up with a simple programmatic way to trigger regenerating the lockfile whenever we bump Bundler... since we can't do it in the `Dockerfiler.updater` build step since then it never makes its way back to the git repo. And we don't want to do it in the `Dockerfile` because that (purposefully) doesn't mount `/updater` code. 

We also (AFAIK) don't have a check that fails the build if any of the Dockerfile builds result in an unexpected `git diff`... that'd probably be useful to add so that we aren't silently having code mutate under our feet before deploy w/o us realizing it. That would have caught this problem actually.

updater/Gemfile.lock
```diff
@@ -1,19 +1,19 @@
 PATH
   remote: ../bundler
   specs:
-    dependabot-bundler (0.211.0)
-      dependabot-common (= 0.211.0)
+    dependabot-bundler (0.212.0)
+      dependabot-common (= 0.212.0)
```
jeffwidman, line 5: This is a bit scary... `Gemfile.lock` was way back on `0.211.0`, yet we released `0.212.0` on Sept 5th: https://github.com/dependabot/dependabot-core/releases/tag/v0.212.0

Ideally we're actually using the _local_ copy of `dependabot-common`, so I'm unclear why this is even pinned here? Or at least I hope we're using the _local_ copy available as part of the build and not fetching it off `RubyGems`?? 😱 

But if it does need to be pinned for some reason, then we ought to at least have it keep up with the latest released version... Maybe that's a secondary step in our release script after we tag a new version?

updater/Gemfile.lock
```diff
@@ -127,19 +127,19 @@ GEM
...
+    aws-sdk-core (3.159.0)
       aws-eventstream (~> 1, >= 1.0.2)
       aws-partitions (~> 1, >= 1.525.0)
       aws-sigv4 (~> 1.1)
       jmespath (~> 1, >= 1.6.1)
     aws-sdk-ecr (1.56.0)
       aws-sdk-core (~> 3, >= 3.127.0)
       aws-sigv4 (~> 1.1)
-    aws-sigv4 (1.5.1)
+    aws-sigv4 (1.5.2)
```
jeffwidman, line 142: I'm unclear why Dependabot hasn't opened a PR to bump these? 
https://github.com/dependabot/dependabot-core/pulls?q=is%3Apr+aws+author%3Aapp%2Fdependabot+

Maybe it's because we've got Dependabot setup to watch this as a library rather than an app? But for `/updater` it should be treating it as an app and bumping these... Need to investigate...

updater/Gemfile.lock
```diff
@@ -330,4 +330,4 @@ DEPENDENCIES
   webmock (~> 3.18)
 
 BUNDLED WITH
-   2.2.20
+   2.3.22
```
jeffwidman, line 333: I think once the other issues are fixed, this becomes the only line that actually changes in this lockfile under the new Bundler version...

updater/Gemfile.lock
jurre: These are pulled in as path dependencies, not from rubygems. The version shows whatever the version is in the branch you're working in, but so does the rest of the code, it always has all the latest changes from disk

Dockerfile
deivid-rodriguez: I think the reason is as @jurre pointed out. These dependencies are pulled from the sibling `dependabot-*` folders, because they are path dependencies. This kind of dependencies are never locked and always use what's there in the filesystem.

updater/Gemfile.lock
deivid-rodriguez: I guess we could have a CI step that runs `bundle update --bundler` and reminds us to commit the result whenever this lock file is about to get out of date (or directly synchronizes it).

updater/Gemfile.lock
deivid-rodriguez: I think it's because dependabot does not update indirect dependencies by default?

updater/Gemfile.lock
jeffwidman: IIUC, https://github.com/dependabot/dependabot-core/pull/5972 solved this by adding `bundle config set --local frozen 'true'` so that CI will blow up if this lock file gets out of date...

Is my understanding correct @deivid-rodriguez ? Just want to double-check before I mark this conversation as resolved... if not, I'll look into adding a CI step in a follow-on PR.

updater/Gemfile.lock
jeffwidman: Indeed, after rebasing on top of #5972 this is the only change left

bdragon (approved):

updater/Gemfile.lock
jeffwidman: Gotcha, then for now we're probably good enough with the comment I added in https://github.com/dependabot/dependabot-core/pull/5858/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R77.

It may get overlooked by someone bumping bundler, but every so often at least someone will see that and remember to do it hopefully so we shouldn't get too far behind.

It'd certainly be nice to have a CI enforcement step, but I'm not coming up with a trivial way to do that that isn't a bit brittle (regex comparisons of the two versions?), and the delta generally won't cause breakage so IMO the comment can suffice for now.

PR #5859, opened 2022-10-10
Fix typos

jeffwidman:


landongrindheim (approved):

updater/spec/dependabot/updater_spec.rb
```diff
@@ -378,8 +378,7 @@
             to receive(:info).with(
               "<job_1> Dependabot can't update vulnerable dependencies for " \
               "projects without a lockfile or pinned version requirement as " \
-              "as the currently installed version of " \
```
jeffwidman, line 381: the second `"as"` was deleted so the test matches the typo fix above

PR #5863, opened 2022-10-10
Yarn Berry: Ensure registry config is respected

jurre:
When a global/scoped registry is configured in the yarnrc, Dependabot should honor this for all requests, also for Yarn Berry, which prior to this change wasn't the case.

This ensures we correctly parse it etc.


npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/latest_version_finder.rb
```diff
@@ -371,7 +371,8 @@ def registry_finder
             dependency: dependency,
             credentials: credentials,
             npmrc_file: npmrc_file,
-            yarnrc_file: yarnrc_file
+            yarnrc_file: yarnrc_file,
+            yarnrc_yml_fle: yarnrc_yml_file
```
jakecoffman, line 375:
```suggestion
            yarnrc_yml_file: yarnrc_yml_file
```

jakecoffman (approved):

yarinsa:
+1 Thanks very helpful!

pavera (approved):

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/latest_version_finder.rb
S1983tt: ,🌠💲👍👍

jurre:
> +1 Thanks very helpful!
> 
> How did you figure out the npmAuthIdent?
> 
> #5870

I don't think it's related? Let's debug your problem in the issue you've opened for it though 👍

PR #5865, opened 2022-10-10
fix: correctly find workspace.dependencies-declared dependencies and their Cargo.toml

### LINKED ISSUE #5864: Rust: Cargo.toml discovery fails if in a workspace and said manifest is not a member but used in members with workspace dependencies

### Package ecosystem

rust, cargo

### Package manager version

cargo 1.64.0+

### Language version

rust 1.64.0+

The situation appeared after my PR https://github.com/dependabot/dependabot-core/pull/5794, and this use case is a little weird so I didn't think about it.

1) Have a Cargo workspace, e.g:

```
root/
- Cargo.toml (C1)
- crate_in_workspace/Cargo.toml (C2)
- crate_outside_workspace/Cargo.toml (C3)
```

```toml
# In (C1)
[workspace.package]
version  = "1.0.0"

[workspace]
members = [ "crate_in_workspace" ]

[workspace.dependencies]
crate_outside_workspace = { version = "1", path = "./crate_outside_workspace" }
```

```toml
# In (C2)
[package]
version.workspace = true

[dependencies]
crate_outside_workspace.workspace = true
```

### dependabot.yml content

_No response_

### Updated dependency

_No response_

Dependabot will currently fail on the config presented earlier with the error that it cannot find `crate_outside_workspace/Cargo.toml`.

This happens because the `workspace.dependencies` table is not looked at by the file filter and I forgot to add it when adding the support for workspace dependencies in #5794.

_No response_

_No response_

_No response_

poliorcetics:
I would like help for writing a test, I have found `cargo/file_fetcher_spec.rb`, but I don't
understand how it works.

Closes #5864

### DIFF
--- a/cargo/lib/dependabot/cargo/file_fetcher.rb
+++ b/cargo/lib/dependabot/cargo/file_fetcher.rb
@@ -264,6 +264,16 @@ module Dependabot
           end
         end
 
+        # Paths specified for workspace-wide dependencies
+        workspace = parsed_file(file).fetch("workspace", {})
+        workspace.fetch("dependencies", {}).each do |_, details|
+          next unless details.is_a?(Hash)
+          next unless details["path"]
+          next unless path == File.join(details["path"], "Cargo.toml")
+
+          return true if details["git"].nil?
+        end
+
         # Paths specified as replacements
         parsed_file(file).fetch("replace", {}).each do |_, details|
           next unless details.is_a?(Hash)


poliorcetics:
I will two need `Cargo.toml` at least from a public repo with the problem then ?

jurre:
> I will two need `Cargo.toml` at least from a public repo with the problem then ?

You can also bas64 encode the contents of a cargo.toml that you have locally and stick it in the fixtures manually

jeffwidman:
@poliorcetics Anything else we can do to unblock you here?

jeffwidman:
gentle nudge @poliorcetics

jeffwidman:
Checking in one last time @poliorcetics are you still interested in working on this?

jeffwidman:
Let me see if I can read up on `cargo` a little and put something together...

jeffwidman:
@poliorcetics I did spend some time on this back in January, and ran into some issues with creating mocks of test code in Rust. The code seems straightforward, but the UT I'd added kept failing--the ruby looked fine, so I think I was misunderstanding something about how the rust cargo files should look. Which is probably on me, as your example in #5864 makes sense.

Any chance you could push up an example repo or mock to help clarify?

PR #5869, opened 2022-10-11
raise when a path dependency is absolute

jakecoffman:
According to [the NPM docs](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#local-paths), a path dependency can be relative or absolute.

Absolute path dependencies can never be resolved by Dependabot and are resulting in EACCES errors when it tries to create files in various places around the update container that it doesn't have access to. For instance there are a considerable amount of errors trying to create `/Users` which is probably a macOS path in the manifest.

So I think our best option here is to throw an appropriate error to bring it to the user's attention.


jurre (approved):
Should we only do this when `path.start_with?("/")` or for any path that we cannot resolve?

jakecoffman:
@jurre True, maybe I'll end up removing this with a more general solution. 

We do already have [this code](https://github.com/dependabot/dependabot-core/blob/a9decb73267e103f49cf8d3c94ba79c10e9be934/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb#L174-L185) which tries to fill out any dependencies it couldn't fetch. It seems like I could just add a raise there when there's no lockfile present.

I'll find an example that hits that case and take another pass at it and see if I can remove the special case in this PR.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -159,6 +159,8 @@ def path_dependencies(fetched_files)
 
         path_dependency_details(fetched_files).each do |name, path|
           path = path.gsub(PATH_DEPENDENCY_CLEAN_REGEX, "")
+          raise PathDependenciesNotReachable, name if path.start_with?("/")
```
landongrindheim, line 162: What do you think of leaving a little more evidence for debugging?

```suggestion
          raise PathDependenciesNotReachable, "#{name} at #{path}" if path.start_with?("/")
```

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -234,6 +236,8 @@ def path_dependency_details_from_manifest(file)
           select { |_, v| v.is_a?(String) && v.start_with?(*path_starts) }.
           map do |name, path|
             path = path.gsub(PATH_DEPENDENCY_CLEAN_REGEX, "")
+            raise PathDependenciesNotReachable, name if path.start_with?("/")
```
landongrindheim, line 239: Same as above. This would help _me_, so I'm suggesting it 😇 If you think it's noise, or if it's going to log information that shouldn't be, feel free to 🗑️ it.

```suggestion
            raise PathDependenciesNotReachable, "#{name} at #{path}" if path.start_with?("/")
```

jakecoffman:
Screen Shot 2022-10-11 at 9 31 29 AM

Issue #4031, opened 2021-07-08
Too many PRs: Rebase new PR before open a new outdated PR


hfhbd:
If your code contains more updatable dependencies as configured in the settings (default 5), dependabot won't create more PRs. After merging one PR, dependabot will open a new PR. 
Unfortunately, this new PR is every time out of date. Although dependabot automatically rebases the PR after a few minutes, it still creates a (email) notification and uses your CI minutes for a useless commit. I would expect a rebased PR when dependabot creates a new PR. 

As the issue happens when merging one of 5 PRs, it would be easy to reproduce by using an out of date repo with more than 5 updatable dependencies, start dependabot, and merge 1 PR.

Sample: checkout this out of date [commit](https://github.com/hfhbd/ComposeTodo/commit/cdda7e509a0131b0064a12ddf3037da94ebe791a) and upload to to a new GitHub repository

**Package ecosystem**
Gradle
**Package manager version**
Gradle 7.1
**Language version**
Kotlin
**Manifest location and content prior to update**
build.gradle.kts

**dependabot.yml content**
Nothing special, https://github.com/hfhbd/ComposeTodo/blob/main/.github/dependabot.yml
**Images of the diff or a link to the PR, issue or logs**
https://github.com/hfhbd/ComposeTodo/pull/303

I hope, this issue does affect the _updating logic_, otherwise I will ask GitHub support.

saiichihashimoto:
I agree. I reduced dependabot down to one PR at a time, but the second PR it creates is always instantly out of date and requires a rebase.

dave-v:
I am having the same problem, except that it doesn't rebase automatically, so every PR that gets raised this way requires a manual rebase command before dependabot rebases it.

orien:
We're experiencing the same. It's causing confusion amongst our engineering team and reducing trust in Dependabot.

1. Dependabot raises 5 PRs (`open-pull-requests-limit` default value).
2. One or more of these are merged.
3. Dependabot creates more PRs, each:
   -  include changes from merged Dependabot PRs (looks like multiple libraries are being upgraded).
   -  the base commit is outdated

@asciimike You mention something similar in https://github.com/dependabot/dependabot-core/issues/4163#issuecomment-908415614. Is this issue being tracked? Is it considered  defect?

melink14:
[...]

PR #5871, opened 2022-10-11
Log when the FileFetcher can't determine the base_commit_sha

jurre:
We've been observing some PRs with diffs that seem to indicate that the PR is opened against a different sha then the one that was used to fetch the files.

In order to determine if that's related to not being able to determine the base_commit_sha, I'd like to start logging this.


mctofu (approved):

jeffwidman:
Issues that are _possibly_ caused by the SHA difference:
* #4031 
* #5234

deivid-rodriguez (approved):

PR #5872, opened 2022-10-11
maven: stop querying repositories once one returns a result

jakecoffman:
Per the [Maven docs](https://maven.apache.org/guides/mini/guide-multiple-repositories.html):

> Remote repository URLs are queried in the following order for artifacts _**until one returns a valid result**_

(emphasis mine)

So when Dependabot checks for updated versions, if it finds some it should stop querying. If the user has outdated dependencies in a registry that may be on purpose, and we should honor those versions presented and not try to find more.

I found "that augment the central repo" a good test of this as it failed once I made this change, so I fixed it and didn't feel the need to write another test.


maven/spec/dependabot/maven/update_checker/version_finder_spec.rb
```diff
@@ -707,9 +707,12 @@
             fixture("maven_central_metadata", "with_date_releases.xml")
           stub_request(:get, maven_central_metadata_url).
             to_return(status: 200, body: body)
+          # 404 causes Dependabot to fall back to the central repo
+          stub_request(:get, jboss_metadata_url).
+            to_return(status: 404)
         end
 
-        its(:count) { is_expected.to eq(87) }
+        its(:count) { is_expected.to eq(17) }
```
landongrindheim, line 715: 🤯

maven/lib/dependabot/maven/update_checker/version_finder.rb
```diff
@@ -57,8 +57,10 @@ def versions
           version_details =
             repositories.map do |repository_details|
               url = repository_details.fetch("url")
-              dependency_metadata(repository_details).
-                css("versions > version").
+              xml = dependency_metadata(repository_details)
+              next [] unless xml
```
landongrindheim, line 61: I spent more time parsing this than I was expecting. Part of this is that I'm less used to seeing return values with `next` (and `break`), but that's 100% valid. The other part, I think, is the way it reads as a line. 👇 is easier for _me_ to parse. What do you think of it?

```suggestion
              next [] if xml.blank?
```

landongrindheim (approved):
Huge improvement!

mctofu (commented):
Looks good! One question is whether the repos are currently in the right order where the first to return is the one we want?

jakecoffman:
@mctofu It seems to be, but I noticed the specs are using `match_array` which doesn't check order. I'll switch to using eq since that's important now.

jakecoffman:
I'll just state here that the rules from the docs linked above are:

1. effective settings:
    A. Global settings.xml
    B. User settings.xml
2. local effective build POM:
   A. Local pom.xml
   B. Parent POMs, recursively
   C. Super POM
3. effective POMs from dependency path to the artifact

1 and 3 don't apply because Dependabot doesn't support settings.xml, and runs in a clean environment.

Rule 2 is captured pretty succinctly in this recursive [section of code](https://github.com/dependabot/dependabot-core/blob/42e4ac21eee8a82235289ad3603ddea19fa38680/maven/lib/dependabot/maven/file_parser/repositories_finder.rb#L47-L53).

I've updated the tests and they look correct.

mctofu (approved):
🎉

olamy:
Hi,
I came into this PR.
I will give you my POV as Maven committer since 2007, Maven PMC member and been chair of the project few years.
 
You are right regarding

>  Remote repository URLs are queried in the following order for artifacts until one returns a valid result

But this applies when Maven is querying artifact `foo` in version x.
So if Maven finds repositories x, y, and z and finds artifact `foo` in version 1 in repository x, Maven will stop here, of course.
Now the Maven build needs artifact `foo` in version 2, but this one is not available in repository x, but Maven finds it in repository y. Maven will use it from repository y. And then artifact `foo` in version 3 can be in repository z.
This means Maven can search multiple repositories for a given version.
As dependabot have a goal of proposing an upgrade of version and most of the time the highest version, this highest version can be in another repository than one containing the current (old?) version.
Ideally, as Maven repository managers are working (see [https://maven.apache.org/repository-management.html](https://maven.apache.org/repository-management.html), dependabot should scan all repositories and aggregate all the available metadata for an artifact and so propose the highest version for an artifact available in all repositories.
Please let me know if this makes sense.

PR #5873, opened 2022-10-11
[npm] Flag indirect transitive updates to be ignored by the FileUpdater

mctofu:
When we need to update a parent dependency to fix an alert on a transitive child dependency we currently return both the parent & child as dependencies to update. Only the parent actually needs to be directly updated as its child dependencies will be automatically updated as part of that process. Still, we include the child dependency in the updated dependency list so it will be included in the PR title/description for context.

Since introducing https://github.com/dependabot/dependabot-core/pull/5822 that means we'll attempt to do a sub-dependency update of the child which has no effect.

This adds a flag to indicate dependencies returned by the updater that don't need to be directly updated and then filters them out before sending them to the FileUpdater. This also works for removed dependencies so we no longer need to filter them explicitly.

I'm making use of the new dependency metadata field that was added in https://github.com/dependabot/dependabot-core/pull/5801. It's currently meant for temporary data that doesn't need to be persisted or used in equals checks so I ended up working on some custom matchers to help validate the metadata output in tests. Let me know if theres a better way I could have done that!

### DIFF
--- a/npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker_spec.rb
+++ b/npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker_spec.rb
@@ -1363,18 +1363,18 @@ RSpec.describe Dependabot::NpmAndYarn::UpdateChecker do
 
     # Dependency doesn't consider metadata as part of equality checks
     # so this allows us to check that the metadata is updated in tests.
-    RSpec::Matchers.define :with_metadata do |expected|
+    RSpec::Matchers.define :including_metadata do |expected|
       match do |actual|
         actual == expected && actual.metadata == expected.metadata
       end
     end
 
-    def contain_exactly_with_metadata(*expected)
-      contain_exactly(*expected.map { |e| with_metadata(e) })
+    def contain_exactly_including_metadata(*expected)
+      contain_exactly(*expected.map { |e| including_metadata(e) })
     end
 
-    def eq_with_metadata(expected_array)
-      eq(expected_array).and contain_exactly_with_metadata(*expected_array)
+    def eq_including_metadata(expected_array)
+      eq(expected_array).and contain_exactly_including_metadata(*expected_array)
     end
 
     context "for a security update for a locked transitive dependency" do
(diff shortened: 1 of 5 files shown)


updater/lib/dependabot/updater.rb
```diff
@@ -733,9 +733,10 @@ def generate_dependency_files_for(updated_dependencies)
         logger_info("Updating #{dependency_names.join(', ')}")
       end
 
-      # Removal is only supported for transitive dependencies which are removed as a
-      # side effect of the parent update
-      deps_to_update = updated_dependencies.reject(&:removed?)
+      # Ignore dependencies that are tagged as information_only. These will be
+      # updated indirectly as a result of a parent dependency update and are
+      # only included here to be included in the PR info.
+      deps_to_update = updated_dependencies.reject { |d| d.metadata[:information_only] }
```
deivid-rodriguez, line 739: I had the thought of whether it makes sense to move this to a dependency method like `#informational_only?` or something, but I'm not sure if other ecosystems will need something like this so probably premature. I just thought that `deps_to_update = updated_dependencies.reject(&:informational_only?)` would read a bit better.

npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker_spec.rb
```diff
@@ -1426,14 +1443,15 @@
         let(:registry_listing_url) { "https://registry.npmjs.org/transitive-dependency-locked-by-intermediate" }
 
         it "correctly updates the transitive dependency" do
-          expect(checker.send(:updated_dependencies_after_full_unlock)).to eq([
+          expect(checker.send(:updated_dependencies_after_full_unlock)).to eq_with_metadata([
```
deivid-rodriguez, line 1446: Nice, I would maybe have chosen `eq_including_metadata` but I'm not great at English 😛.

deivid-rodriguez (approved):
Nice optimization!

updater/lib/dependabot/updater.rb
mctofu: Thanks! I was on the fence about adding that but you've persuaded me!

npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker_spec.rb
mctofu: I like that wording better as well so updated to use it!

PR #5874, opened 2022-10-11
Fixed disabledPackageSources for nuget.org

honeyankit:
In the nuget.config, adding the default public Nuget registry to `disabledPackageSources` section was not disabling it. Dependabot was still able to access the default public registry (nuget.org). 

With this modification, the dependabot won't contact the default public registry if a user disable the public nuget registry in the `disabledPackageSources` section even without using the `<clear />` tag in the nuget.config. [[ref](https://learn.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#example-nugetdefaultsconfig-and-application)]

### DIFF
--- a/nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
+++ b/nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
@@ -11,6 +11,7 @@ module Dependabot
     class UpdateChecker
       class RepositoryFinder
         DEFAULT_REPOSITORY_URL = "https://api.nuget.org/v3/index.json"
+        DEFAULT_REPOSITORY_API_KEY = "nuget.org"
 
         def initialize(dependency:, credentials:, config_files: [])
           @dependency  = dependency
@@ -30,7 +31,7 @@ module Dependabot
           @find_dependency_urls ||=
             known_repositories.flat_map do |details|
               if details.fetch(:url) == DEFAULT_REPOSITORY_URL
-                # Save a request for the default URL, since we already how
+                # Save a request for the default URL, since we already know how
                 # it addresses packages
                 next default_repository_details
               end
@@ -151,25 +152,24 @@ module Dependabot
         def repos_from_config_file(config_file)
           doc = Nokogiri::XML(config_file.content)
           doc.remove_namespaces!
-          sources =
-            doc.css("configuration > packageSources > add").map do |node|
-              {
-                key:
-                  node.attribute("key")&.value&.strip ||
-                    node.at_xpath("./key")&.content&.strip,
-                url:
-                  node.attribute("value")&.value&.strip ||
-                    node.at_xpath("./value")&.content&.strip
-              }
-            end
+          # analogous to having a root config with the default repository
+          base_sources = [{ url: DEFAULT_REPOSITORY_URL, key: "nuget.org" }]
 
+          sources = []
+          doc.css("configuration > packageSources").children.each do |node|
+            if node.name == "clear"
+              sources.clear
+              base_sources.clear
+            else
+              key = node.attribute("key")&.value&.strip || node.at_xpath("./key")&.content&.strip
+              url = node.attribute("value")&.value&.strip || node.at_xpath("./value")&.content&.strip
+              sources << { url: url, key: key }
+            end
+          end
+          sources += base_sources # TODO: quirky overwrite behavior
           disabled_sources = disabled_sources(doc)
           sources.reject! do |s|
             disabled_sources.include?(s[:key])
-          end
-
-          unless doc.css("configuration > packageSources > clear").any?
-            sources << { url: DEFAULT_REPOSITORY_URL, key: nil }
           end
 
           sources.reject! do |s|
@@ -202,7 +202,7 @@ module Dependabot
 
         # rubocop:disable Metrics/PerceivedComplexity
         def disabled_sources(doc)
-          doc.css("configuration > disabledPackageSources > add").map do |node|
+          doc.css("configuration > disabledPackageSources > add").filter_map do |node|
             value = node.attribute("value")&.value ||
                     node.at_xpath("./value")&.content
 
--- a/nuget/spec/dependabot/nuget/update_checker/repository_finder_spec.rb
+++ b/nuget/spec/dependabot/nuget/update_checker/repository_finder_spec.rb
@@ -228,6 +228,84 @@ RSpec.describe Dependabot::Nuget::UpdateChecker::RepositoryFinder do
         )
       end
 
+      context "include the default repository" do
+        let(:config_file_fixture_name) { "include_default_disable_ext_sources.config" }
+
+        it "with disable external source" do
+          expect(dependency_urls).to match_array(
+            [{
+              repository_url: "https://www.myget.org/F/exceptionless/api/v3/" \
+                              "index.json",
+              versions_url: "https://www.myget.org/F/exceptionless/api/v3/" \
+                            "flatcontainer/microsoft.extensions." \
+                            "dependencymodel/index.json",
+              search_url: "https://www.myget.org/F/exceptionless/api/v3/" \
+                          "query?q=microsoft.extensions.dependencymodel" \
+                          "&prerelease=true&semVerLevel=2.0.0",
+              auth_header: { "Authorization" => "Basic bXk6cGFzc3cwcmQ=" },
+              repository_type: "v3"
+            }, {
+              repository_url: "https://api.nuget.org/v3/index.json",
+              versions_url: "https://api.nuget.org/v3-flatcontainer/" \
+                            "microsoft.extensions.dependencymodel/index.json",
+              search_url: "https://azuresearch-usnc.nuget.org/query" \
+                          "?q=microsoft.extensions.dependencymodel" \
+                          "&prerelease=true&semVerLevel=2.0.0",
+              auth_header: {},
+              repository_type: "v3"
+            }]
+          )
+        end
+      end
+
+      context "that overides the default package sources" do
+        let(:config_file_fixture_name) { "override_def_source_with_same_key.config" }
+
+        before do
+          repo_url = "https://www.myget.org/F/exceptionless/api/v3/index.json"
+          stub_request(:get, repo_url).
+            to_return(
+              status: 200,
+              body: fixture("nuget_responses", "myget_base.json")
+            )
+        end
+
+        it "when the default api key of defaut registry is provided without clear" do
+          expect(dependency_urls).to match_array(
+            [{
+              repository_url: "https://www.myget.org/F/exceptionless/api/v3/" \
+                              "index.json",
+              versions_url: "https://www.myget.org/F/exceptionless/api/v3/" \
+                            "flatcontainer/microsoft.extensions." \
+                            "dependencymodel/index.json",
+              search_url: "https://www.myget.org/F/exceptionless/api/v3/" \
+                          "query?q=microsoft.extensions.dependencymodel" \
+                          "&prerelease=true&semVerLevel=2.0.0",
+              auth_header: {},
+              repository_type: "v3"
+            }]
+          )
+        end
+
+        let(:config_file_fixture_name) { "override_def_source_with_same_key_default.config" }
+        it "when the default api key of defaut registry is provided with clear" do
+          expect(dependency_urls).to match_array(
+            [{
+              repository_url: "https://www.myget.org/F/exceptionless/api/v3/" \
+                              "index.json",
+              versions_url: "https://www.myget.org/F/exceptionless/api/v3/" \
+                            "flatcontainer/microsoft.extensions." \
+                            "dependencymodel/index.json",
+              search_url: "https://www.myget.org/F/exceptionless/api/v3/" \
+                          "query?q=microsoft.extensions.dependencymodel" \
+                          "&prerelease=true&semVerLevel=2.0.0",
+              auth_header: {},
+              repository_type: "v3"
+            }]
+          )
+        end
+      end
+
       context "that doesn't include the default repository" do
         let(:config_file_fixture_name) { "excludes_default.config" }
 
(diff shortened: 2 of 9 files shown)


jakecoffman (dismissed):
Makes sense to me!

nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
```diff
@@ -163,15 +163,15 @@ def repos_from_config_file(config_file)
               }
             end
 
+          unless doc.css("configuration > packageSources > clear").any?
+            sources << { url: DEFAULT_REPOSITORY_URL, key: nil }
```
mctofu, line 167: Since the key is `nil` here I'm wondering how it could possible be removed by the disabled sources which rejects by key. Is there something else that would filter this out? Are we de-duping the the url somewhere?

jakecoffman: Good catch, it looks like `disabled_sources` is returning nil because of the `value="false"` in the test which happens to match the default registry key here and removes it.

honeyankit: Good point, I looked into it. The function `disabled_sources` returns an array of the same number of elements it is provided as input because it is using map method inside `(doc.css("configuration > disabledPackageSources > add").select do |node|)` . So based on the number of entires in the`<disabledPackageSources>` section of nuget.config file it will return same number of elements .

So for the fixture which I have written it had two entries and the output of `disabled_sources` function is `[nil, 'nuget.org']`.

```
    <disabledPackageSources>
        <add key="MyRepo - ES" value="false" />
        <add key="nuget.org" value="true" />
    </disabledPackageSources>
```

Since the output of the `disabled_sources is [nil, 'nuget.org']` the below code is able to reject the key with nil value.
```
  sources.reject! do |s|
    disabled_sources.include?(s[:key])
  end
```

I can see where this will fail, when dependabot has access to public internet, no `<clear />` tag is provided in nuget.config file and the `<disabledPackageSources>` contains more then one entry with value as 'false'. In this case `disabled_sources` function will return `[nil]` and will remove the default `nuget.org` because we are setting the sources with key as nil `sources << { url: DEFAULT_REPOSITORY_URL, key: nil }`. A simple fix will be to add a key 'nuget.org' instead of `nil` to code `sources << { url: DEFAULT_REPOSITORY_URL, key: nil }`

```
    <disabledPackageSources>
        <add key="MyRepo - ES" value="false" />
    </disabledPackageSources>
```

honeyankit: Push the changes.

honeyankit: A simple fix is that a key 'nuget.org' instead of `nil` is added the to code `sources << { url: DEFAULT_REPOSITORY_URL, key: nil }`

nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
```diff
@@ -163,15 +163,15 @@ def repos_from_config_file(config_file)
               }
             end
 
+          unless doc.css("configuration > packageSources > clear").any?
+            sources << { url: DEFAULT_REPOSITORY_URL, key: "nuget.org" }
```
mctofu, line 167: This goes beyond the scope of this original change but thinking about this:
1. Should we also skip adding this if there is already a "nuget.org" key defined? Is that a valid way to override the default nuget source by defining your own package source with the same key?
2. Is "nuget.org" the key used by the default source? I think it is but haven't been able to confirm that.

nuget/spec/fixtures/configs/disabled_default_sources.config
```diff
@@ -0,0 +1,69 @@
...
+
+    <!--
+        Used to specify the default Sources for list, install and update.
+        See: nuget.exe help list
+        See: nuget.exe help install
+        See: nuget.exe help update
+    -->
+    <packageSources>
+        <add key="MyRepo - ES" value="https://www.myget.org/F/exceptionless/api/v3/index.json" />
+        <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
```
mctofu, line 43: It might be better to remove this line? The goal of the change is to make sure you can disable the default package source but by having this here we are introducing an entry that isn't the default.

nuget/spec/dependabot/nuget/update_checker/repository_finder_spec.rb
```diff
@@ -228,6 +228,36 @@
         )
       end
 
+      context "that include the default repository with disable external source" do
```
mctofu, line 231: 💯 

We could also add a `.compact` to `disabled_sources` to remove the nil keys from disabled sources.

```
        def disabled_sources(doc)
          doc.css("configuration > disabledPackageSources > add").map do |node|
            value = node.attribute("value")&.value ||
                    node.at_xpath("./value")&.content
            if value&.strip&.downcase == "true"
              node.attribute("key")&.value&.strip ||
                node.at_xpath("./key")&.content&.strip
            end
          end.compact     <--
        end
```

nuget/spec/fixtures/configs/disabled_default_sources.config
honeyankit: I see, what you are saying. Thanks.

nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
honeyankit: The default key is `nuget` in the nuget.config from the nuget command line, so changing the key from the nuget.org to nuget. 

```
$ dotnet new nugetconfig
The template "NuGet Config" was created successfully.
Ankits-MacBook-Pro:test anhoney$ ls
nuget.config
Ankits-MacBook-Pro:test anhoney$ cat nuget.config 
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <!--To inherit the global NuGet package sources remove the <clear/> line below -->
    <clear />
    <add key="nuget" value="https://api.nuget.org/v3/index.json" />
  </packageSources>
</configuration>
```

nuget/spec/dependabot/nuget/update_checker/repository_finder_spec.rb
honeyankit: The problem is the key used as `nil`. If we use default source url's key as `nil` and compact in `disabled_sources` then the below code will never be remove the default resource. 
```
  sources.reject! do |s|
    disabled_sources.include?(s[:key])
  end
```

honeyankit: After this change, we should update nuget ecosystem doc as well  that to disable default nuget registy without `<clear />`, user can add key 'nuget' to the `disabledPackageSources` in their nuget.config file.

nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
honeyankit: > Should we also skip adding this if there is already a "nuget.org" key defined? Is that a valid way to override the default nuget source by defining your own package source with the same key?

No need to skip. If the user specify the same key as default registry, the user specified key should override [[ref](https://learn.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#settings-walkthrough)]. This should also be true if users is using the default registry key to override with own package source. I think users needs to know via docs that dependabot uses 'nuget' as a key to set the default registry in nuget.config file.

nuget/spec/dependabot/nuget/update_checker/repository_finder_spec.rb
honeyankit: If we agreed to keep default registry key as `nuget` or `nuget.org` or some value then we will not need compact, though test case is still valid.

nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
honeyankit: Reading through the [nuget.config settings walk through ](https://learn.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#settings-walkthrough), there is another bug where `<clear />` is not working as intended. Currently clear is just removing the default registry when used, but in reality it should remove all the registries which are mentioned before it in the `packageSources` section. I will open a separate bug for it.

honeyankit: As discussed, added the fix to override the default registry url with some other url when same key (nuget) is specified in the nuget.config.

nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
```diff
@@ -11,6 +11,7 @@ module Nuget
     class UpdateChecker
       class RepositoryFinder
         DEFAULT_REPOSITORY_URL = "https://api.nuget.org/v3/index.json"
+        DEFAULT_REPOSITORY_API_KEY = "nuget"
```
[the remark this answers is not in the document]
honeyankit: Thanks for reviewing. I used `dotnet new nugetconfig` and it create a default nuget.config with `nuget` as key. But in many examples I can see nuget.org used as key. I will switch the default key to `nuget.org`

```
Ankits-MacBook-Pro:test anhoney$ dotnet new nugetconfig
The template "NuGet Config" was created successfully.
Ankits-MacBook-Pro:test anhoney$ cat nuget.config 
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <!--To inherit the global NuGet package sources remove the <clear/> line below -->
    <clear />
    <add key="nuget" value="https://api.nuget.org/v3/index.json" />
  </packageSources>
</configuration>
Ankits-MacBook-Pro:test anhoney$ 
```

nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
```diff
@@ -163,15 +164,20 @@ def repos_from_config_file(config_file)
...
 
+          # If the clear tag is missing and the user is using "nuget" as a key in the
+          # the packageSources section of nuget.config for its private registry url,
+          # the user-specified url will take precedence over the default repository url
+          is_default_repo_key_exist = doc.css("configuration > packageSources > add").find \
+            { |node| node.attribute("key")&.value&.strip == DEFAULT_REPOSITORY_API_KEY }
+          if doc.css("configuration > packageSources > clear").none? && is_default_repo_key_exist.nil?
+            sources << { url: DEFAULT_REPOSITORY_URL, key: DEFAULT_REPOSITORY_API_KEY }
+          end
+
```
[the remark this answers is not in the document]
honeyankit: You are right, anything coming before `clear` should be ignored. This makes totally sense to me.

nuget/spec/fixtures/configs/disabled_default_sources.config
```diff
@@ -0,0 +1,69 @@
...
+        <add key="MyRepo - ES" value="https://www.myget.org/F/exceptionless/api/v3/index.json" />
+        <add key="nuget" value="https://api.nuget.org/v3/index.json" />
+    </packageSources>
+
+    <!-- Used to store credentials -->
+    <packageSourceCredentials>
+        <MyRepo_x0020_-_x0020_ES>
+            <add key="Username" value="my" />
+            <add key="ClearTextPassword" value="passw0rd" />
+        </Test_x0020_Source>
```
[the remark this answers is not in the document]
honeyankit: I copied the template from an existing one and it seems all the incorrect. Fixing it now.

nuget/spec/dependabot/nuget/update_checker/repository_finder_spec.rb
```diff
@@ -228,6 +228,83 @@
...
+                            "flatcontainer/microsoft.extensions." \
+                            "dependencymodel/index.json",
+              search_url: "https://www.myget.org/F/exceptionless/api/v3/" \
+                          "query?q=microsoft.extensions.dependencymodel" \
+                          "&prerelease=true&semVerLevel=2.0.0",
+              auth_header: {},
+              repository_type: "v3"
+            }]
+          )
+        end
```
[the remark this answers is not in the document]
honeyankit: Yes, good catch. Fixed it. Thanks

nuget/spec/fixtures/configs/disabled_default_sources.config
```diff
@@ -0,0 +1,69 @@
+<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
honeyankit: Noted and fixed in all other spec file as well.

nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
```diff
@@ -162,16 +163,26 @@ def repos_from_config_file(config_file)
                     node.at_xpath("./value")&.content&.strip
               }
             end
```
jakecoffman, line 165: You can delete this whole sources block since we're handling that below.

nuget/spec/dependabot/nuget/update_checker/repository_finder_spec.rb
```diff
@@ -228,6 +228,83 @@
...
+          repo_url = "https://www.myget.org/F/exceptionless/api/v3/index.json"
+          stub_request(:get, repo_url).
+            to_return(
+              status: 200,
+              body: fixture("nuget_responses", "myget_base.json")
+            )
+        end
+
+        it "when the default api key of defaut registry is provided without clear" do
+          expect(dependency_urls).to match_array(
```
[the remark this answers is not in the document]
jakecoffman: We discussed that this behavior is probably a bug and we're going to ignore it and just let it overwrite which is what the documentation says it should do.

nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
honeyankit: Completed. Good catch. Thanks!

jakecoffman (approved):
Looks great!

PR #5878, opened 2022-10-12
Fix Maven inability to overwrite repository urls by ID

jakecoffman:
In Maven, it is possible to overwrite repository URLs defined elsewhere by defining a repository with the same ID, as long as the file has precedence according to the [order defined in the docs](https://maven.apache.org/guides/mini/guide-multiple-repositories.html#repository-order).

This is often used to define a different central repository to replace the default one inherited from the Super POM:

```xml
<repositories>
    <repository>
      <id>central</id>
      <url>https://example.com</url>
      <snapshots>
        <enabled>false</enabled>
      </snapshots>
    </repository>
  </repositories>
```

In this PR, I've tried to honor that behavior by gathering the IDs along with the URL and deduplicate by ID according to that precedence.

One caveat is that Maven differentiates between repository and pluginRepository. Our code does not, so if central is overwritten in a repository it will also overwrite it in pluginRepository. It seems somewhat unlikely that only one or the other is overwritten, so I think this is probably ok for now.

### DIFF
--- a/maven/lib/dependabot/maven/file_parser/repositories_finder.rb
+++ b/maven/lib/dependabot/maven/file_parser/repositories_finder.rb
@@ -24,6 +24,7 @@ module Dependabot
         # The Central Repository is included in the Super POM, which is
         # always inherited from.
         CENTRAL_REPO_URL = "https://repo.maven.apache.org/maven2"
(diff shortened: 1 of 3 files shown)


mctofu (approved):
Nice!

PR #5883, opened 2022-10-13
Yarn Private Registry Support

Nishnha:
This PR aims to fix how Dependabot handles private registries for Yarn Classic (Yarn v1).

Previously, Dependabot would make public registry calls to `registry.yarnpkg.com`, even if a private registry was configured, because the logic for identifying private registries was copied over from npm.

With this change, Dependabot can use dedicated Yarn registry builder methods to make authenticated private registry requests, and should not make any more public registry requests if a private registry is configured in `.yarnrc` or `dependabot.yml`.

For reviewers: can you double check that the test cases listed here are comprehensive?


npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
```diff
@@ -336,11 +337,15 @@ def dependencies_in_error_message?(message)
           end
         end
 
-        def write_temporary_dependency_files(update_package_json: true)
+        def write_temporary_dependency_files(yarn_lock, update_package_json: true)
```
Nishnha, line 340: I required a yarn lock for this method because of the `yarn_berry?` check

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
```diff
@@ -539,11 +544,16 @@ def yarnrc_specifies_npm_reg?
 
           return false unless yarnrc_global_registry
 
-          URI(yarnrc_global_registry).host == "registry.npmjs.org"
+          UpdateChecker::RegistryFinder::CENTRAL_REGISTRIES.any? do |r|
+            r.include?(URI(yarnrc_global_registry).host)
+          end
```
Nishnha, line 549: Is there a better way to check for whether a config file lists private registries?
We could filter out by `CENTRAL_REGISTRIES` and check whether the output has any entries?

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/dependency_files_builder.rb
```diff
@@ -69,6 +74,24 @@ def write_lock_files
...
+            lines.find { |line| line.match?(regex) }&.
+            match(regex)&.
+            named_captures&.
+            fetch("registry")
+
+          return false unless yarnrc_global_registry
+
+          UpdateChecker::RegistryFinder::CENTRAL_REGISTRIES.none? do |r|
+            r.include?(URI(yarnrc_global_registry).host)
+          end
```
Nishnha, line 92: Ditto for this

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npmrc_builder.rb
```diff
@@ -34,6 +35,20 @@ def npmrc_content
           ([initial_content] + credential_lines_for_npmrc).compact.join("\n")
         end
 
+        # PROXY WORK
+        # Yarn allows registries to be defined either in an .npmrc or .yarnrc
+        # so we need to parse both files for registry keys
+        def yarnrc_content
+          initial_content =
+            if npmrc_file then complete_yarnrc_from_credentials
+            elsif yarnrc_file then build_yarnrc_from_yarnrc
```
Nishnha, line 44: we should flip these 2 so we check for a `.yarnrc` first

jakecoffman (approved):
I tested this with our private registry integration tests and it seems to work flawlessly with Yarn v1. The Yarn Berry tests do still make calls to registry.yarnpkg.com but it succeeds in making the two PRs. 

So this is a huge improvement! 👏 

@pavera anything you want to look at before we ship this?

pavera (approved):

Issue #5523, opened 2022-08-12
Maven package ecosystem not checking dependencies from my private repo


gianielsevier:
**Package ecosystem**
maven
**Package manager version**
3.8.6
**Language version**
java11, java17
**Manifest location and content before the Dependabot update**
pom.xml

If using GitHub-native Dependabot, attach your dependabot.yml file or provide a link to it
```yaml
version: 2
registries:
  my-artifactory:
    type: maven-repository
    url: https://my.internal.repo
    username: MY_USER
    password: ${{secrets.MY_SECRET}}
updates:
  - package-ecosystem: "maven"
    directory: "/"
    registries:
      - my-artifactory
    schedule:
      interval: "daily"
```
**What you expected to see, versus what you actually saw**
I was expecting dependabot check dependencies from my private maven repo. But it is going to repo.maven.apache
**Native package manager behavior**
```
proxy | time="2022-08-12T09:55:35Z" level=info msg="proxy starting" commit=d1feaa99254076b1f2be38bedebf2e3fe49ef1c6
  proxy | 2022/08/12 09:55:35 Listening (:1080)
updater | 2022-08-12T09:55:35.249798454 [anonymous-instance:main:WARN:src/firecracker/src/main.rs:370] You are using a deprecated parameter: --seccomp-level 2, that will be removed in a future version.
updater | 2022-08-12T09:55:35.273297317 [439162729:main:WARN:src/devices/src/legacy/serial.rs:432] Detached the serial input due to peer close/error.
updater | time="2022-08-12T09:55:36Z" level=info msg="guest starting" commit=03c82863413a174048b5d5d081fe33c52016da94
updater | time="2022-08-12T09:55:36Z" level=info msg="starting job..." fetcher_timeout=5m0s job_id=439162729 updater_timeout=45m0s updater_version=0.207.0-8bdfd36a8d3a47a9944976c94af2e0105382e08f
updater | I, [2022-08-12T09:55:38.142880 #7]  INFO -- sentry: ** [Raven] Raven 3.1.2 ready to catch errors
updater | To use retry middleware with Faraday v2.0+, install `faraday-retry` gem
updater | INFO <job_439162729> Starting job processing
  proxy | 2022/08/12 09:55:40 [002] GET https://api.github.com:443/repos/internal-github-repo/my-app
  proxy | 2022/08/12 09:55:40 [002] * authenticating github api request
[...]

PR #5884, opened 2022-10-13
add Maven credential metadata to the URLs it searches for POM files

jakecoffman:
Currently Dependabot uses any repository or pluginRepository URL it finds in the POM to search for parent POM files. However, there is no reason a child POM needs to have any repositories defined at all. It's fairly common for private/custom registries to only be defined in the parent which is published to the registry. 

The reason why this works locally for Maven users is they often have a `settings.xml` which defines what registry to use, but that file isn't typically checked in to git.

Dependabot users commonly try to put the registry in `dependabot.yml` as a hint to where it can find the parent POM.

This PR takes that hint and puts Maven registries defined in `dependabot.yml` at the front of the list of registries to check.

### DIFF
--- a/maven/lib/dependabot/maven/file_parser/repositories_finder.rb
+++ b/maven/lib/dependabot/maven/file_parser/repositories_finder.rb
@@ -26,8 +26,9 @@ module Dependabot
         CENTRAL_REPO_URL = "https://repo.maven.apache.org/maven2"
         SUPER_POM = { url: CENTRAL_REPO_URL, id: "central" }
 
-        def initialize(dependency_files:, evaluate_properties: true)
+        def initialize(dependency_files:, credentials: [], evaluate_properties: true)
           @dependency_files = dependency_files
+          @credentials = credentials
 
           # We need the option not to evaluate properties so as not to have a
           # circular dependency between this class and the PropertyValueFinder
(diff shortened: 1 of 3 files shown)


jakecoffman:
This should fix https://github.com/dependabot/dependabot-core/issues/5523

honeyankit (approved):
This looks good to me.

maven/lib/dependabot/maven/file_parser/repositories_finder.rb
```diff
@@ -155,6 +156,12 @@ def remote_pom_url(group_id, artifact_id, version, base_repo_url)
             "#{artifact_id}-#{version}.pom"
         end
 
+        def urls_from_credentials
+          @credentials.
+            select { |cred| cred["type"] == "maven_repository" }.
```
mctofu, line 161: I had been thinking we'd only do this for creds with `replaces-base: true`. I can see now that you might want this but still allow a public registry though. Is there any potential for breaking existing jobs if we start to automatically move the registry sources to the front?

Alternatives might be:
1. Introducing a new "global" key on the registry to indicate it comes from a global configuration not checked into the repo so we should insert it
2. Use `replaces-base` which would also remove automatic usage of the default registry but allow the default registry to added back via inclusion in `dependabot.yml` registries.

But if it doesn't seem likely this would break anything we could try this out first.

jakecoffman: My thought was `replaces-base` would remove the default registry if it is true in any of the maven creds. I was going to follow this up by adding that in the next bite-sized PR. 

This approach does allow fallback to Maven Central. For users of GitHub Registry that's probably what they want. It's also possible GHR users would want to check Central first to avoid 404ing first. If we drop the username and password requirement, then they have complete control of the order by specifying in dependabot.yml.

PR #5891, opened 2022-10-16
Never change version precision of actions chosen by users

### LINKED ISSUE #2304: Allow bumping GH actions to rolling releases
Currently dependabot will bump GitHub actions to whatever the latest `vX.Y.Z` tag is. However, actions also usually have a `vX` tag that points to the latest X.Y.Z (and gets updated). I'd like to be able to select the bumping behavior and prefer `vX` tags.

Example PR that was opened with full version and then manually changed to `v2`: https://github.com/theforeman/releasetool/pull/19

### LINKED ISSUE #5887: Github Actions - allow updating major versions only

### Feature description

GitHub Actions follow docker-like tag pattern, where a new version updates 3 tags simultaneously:

- full version: v1.2.3
- minor version: v1.2
- major version: v1

A common practice then is to use a major version tag (`action@v1`) to allow automatic minor version drift, and update major versions manually (as that often implies breaking changes)

At the moment, Dependabot is not aware of this convention, and unconditionally suggests the "max" version. This leads to situation when an action updates an action v2 to v3.0.0 [example](https://github.com/DarkWanderer/ClickHouse.Client/pull/203), and that leads to a lot of extraneous PRs after (updating 3.0.0=>3.0.1, 3.0.1=>3.0.2 etc.)

Suggestion: add `versioning-strategy` value for GitHub actions (e.g. `major`) which will instruct Dependabot to exclusively use tags where only major version is specified

deivid-rodriguez:
Never changing the existing version precision is our default behavior, however there's one edge case where Dependabot may actually propose a PR changing the precision.

This is only in the interim between the time where a new major version of the action has been released, but the action's author has not yet created a new major version tag pointing to the new release.

If Dependabot happens to run between the above two events, it will create a PR changing the precision from vX to vX.Y.Z, just because it's unable to find any versions with the same precision.

I think this in never what users expect, and it introduces inconsistent behavior in Dependabot, not dependent of the state of the manifest files, but on external circumstances which is confusing. On top of that, the issue will be automatically "fixed" next time Dependabot runs, once the actions author has created the proper major version tag.

I'm aware that I'm changing expectations of existing specs here, but I don't think this is intended after all. These specs where initially added [here](https://github.com/dependabot/dependabot-core/pull/4953), precisely to fix "precision being changed" issues. The few specs that I'm tweaking I think may have been added to have a nice fallback behavior when tags with the same precision are not available. However, in practice users still don't want this since the valid edge cases where this happens are only temporary.

So I think it's best to simplify this code to filter out versions not fully matching existing precision, and then choose the latest over what's left.

Fixes #5887.
Closes #2304.

### DIFF
--- a/github_actions/lib/dependabot/github_actions/update_checker.rb
+++ b/github_actions/lib/dependabot/github_actions/update_checker.rb
@@ -98,23 +98,13 @@ module Dependabot
           latest_tags = git_commit_checker.local_tags_for_latest_version_commit_sha
 
           # Find the latest version with the same precision as the pinned version.
-          # Falls back to a version with the closest precision if no exact match.
-          current_dots = dependency.version.split(".").length
-          latest_tags.max do |a, b|
-            next a[:version] <=> b[:version] unless shortened_semver_version_eq?(a[:version], b[:version])
-
-            a_dots = a[:version].to_s.split(".").length
-            b_dots = b[:version].to_s.split(".").length
-            a_diff = (a_dots - current_dots).abs
-            b_diff = (b_dots - current_dots).abs
-            next -(a_diff <=> b_diff) unless a_diff == b_diff
-
-            # preference to a less specific version if we have a tie
-            next 1 if a_dots < current_dots
-
-            -1
-          end
+          current_precision = precision(dependency.version)
+          latest_tags.select { |tag| precision(tag[:version].to_s) == current_precision }.max_by { |tag| tag[:version] }
         end
+      end
+
+      def precision(version)
+        version.split(".").length
       end
 
       def updated_source
@@ -191,13 +181,6 @@ module Dependabot
         return false unless base_split.length <= other_split.length
 
         other_split[0..base_split.length - 1] == base_split
-      end
-
-      def shortened_semver_version_eq?(base_version, other_version)
-        base = base_version.to_s
-        other = other_version.to_s
-
-        shortened_semver_eq?(base, other) || shortened_semver_eq?(other, base)
       end
 
       def find_container_branch(sha)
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Ok, this discussion actually happened here: https://github.com/dependabot/dependabot-core/pull/4953#discussion_r842257936. @mctofu was considering both the approach in this PR and the one used currently, but in the end it was decided after discussing with @brrygrdn to go with the current one. However, I think the situation described here was not considered at the time, namely, author not intentionally changing the version policy, just taking some time to create the new major version tag, enough for Dependabot to run in the middle and create an unexpected update.

So I'd like to reevaluate our approach, I think it's more consistent and simpler this way, and less subject to surprise users.

I think changing existing behavior should require strong agreement, so if not everyone agrees here I'm happy to close and possibly think of something else.

mctofu (approved):
:+1: Thanks for fixing this! The precision change when there's a delay in publishing the version was definitely not expected or intended.

brrygrdn (approved):
TL;DR - I think this behaviour is preferable 👍🏻 

Reading over the original conversation, I think my bias towards creating _any PR at all_ was based on the idea that we (currently) aren't prompted to try to create a PR again when a new version of the Action is released, so if we miss the opportunity to create a PR when the alert goes out we likely won't create one without user intervention (i.e. they push the workflow / dependabot config / press a button on the Alert )

Looking at it with fresh eyes, I think this behaviour is probably better as:
- The alert that the repo is vulnerable will still exist so if the user wants to change precision to get the fix asap, they can
- I also overlooked the convention in actions to 'move' the major tag to the new minor version so this was likely causing a lot more unhelpful PRs than I considered at the time.

PR #5904, opened 2022-10-17
Install composer in a way that does not use `COPY --from`

deivid-rodriguez:
I observed that many times buildkit cache is not being hit, triggering a rebuild of the whole dependabot-core image, in situations where it would not seem necessary.

One hypothesis is that this is due to the `COPY --from` usage when installing composer, since I recall issues with that not playing well with buildkit cache.

I want to try a different way of installing composer, in order to verify whether this is the culprit and this speeds up our CI.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -141,8 +141,6 @@ RUN [ "$TARGETARCH" != "amd64" ] \
 
 # Install PHP and Composer
 ENV COMPOSER_ALLOW_SUPERUSER=1
-COPY --from=composer:1.10.26 /usr/bin/composer /usr/local/bin/composer1
-COPY --from=composer:2.3.9 /usr/bin/composer /usr/local/bin/composer
 RUN add-apt-repository ppa:ondrej/php \
   && apt-get update \
   && apt-get install -y --no-install-recommends \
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
I don’t think the problem is related to this after another look. When cache is not hit, I thought the first not hit line was the composer one, hence my hypothesis, but I see that I just got that wrong, the apt-get update line above does not hit the cache either.

However, each of the new ways of installing seems to take ~1second while the old way of pulling file installed file from external docker images seems considerably slower (~10 seconds), so maybe still worth doing. And both installations can be compressed to a single layer using this approach.

Nishnha (approved):
Glad the composer installations get cached with this change. Thanks!

deivid-rodriguez:
Maybe it's too early to celebrate but I've seen consistent CACHE hits when rebuilding on main image since I merged this PR 🎉

jeffwidman:
One potential downside of this PR is that the composer version is now sitting in a random bash command which Dependabot is unlikely to know how to monitor. Versus the docker image tags--while we don't process them today, we are more likely to process them in the future... I've seen a POC floating around on us eventually doing this.

If this allows us to have faster builds/hit caches better, than I'm still 👍 on this change, but wanted to mention it.

deivid-rodriguez:
That's a good point! Still I haven't seen any unepexted cache misses since merging this so I think this is worth keeping though it also makes things consistent because it's the only place in the Dockerfile where we use `COPY --from`, other ecosystems are installed in a way more similar way to what we do now.

PR #5905, opened 2022-10-17
Only call pip compile once

### LINKED ISSUE #5903: pip-compile is called twice per dependency

I noticed some of our Python pip-compile projects taking a long time to be processed by Dependabot (as in 30+ minutes for a cached install). When looking into it I noticed that pip-compile just seems to be slow in general, it will install all dependencies for every dependency to update (which is slow, even if they are cached). 

Both the file updater as the version resolver will run pip-compile twice:
https://github.com/dependabot/dependabot-core/blob/main/python/lib/dependabot/python/file_updater/pip_compile_file_updater.rb#L84

https://github.com/dependabot/dependabot-core/blob/main/python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb#L83

This was introduced in #961 because pip-compile added something that made the header added by pip-compile be changed by Dependabot. 

The header without the change in #961 will have looked something like this:

```
pip-compile --allow-unsafe --output-file=requirements.txt --resolver=backtracking -P {dependency-name} requirements.in
```

Where `{dependency-name}` is the name of the dependency it just tried to update.

I think running pip-compile twice just to fix the header is a bit overkill, as it basically doubles the runtime that is needed for dependabot, and when there is an update it takes even longer.

Suggested fix: pip-compile adds ` -P {dependency-name}` to the header, we can just remove ` -P {dependency-name}` from the resulting .txt file with a replace.

Another fix could be to copy the header from the old .txt, and build the new .txt with the `--no-header` option and then prepend the old header in the resulting .txt file, in that way we don't mess with the header at all, but that requires more work and might be a ticket on it's own.

jerbob92:
Avoid unnecessary `pip-compile` processes, and explain why we need to run an extra one sometimes.

Closes #5903.

### DIFF
--- a/python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
+++ b/python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
@@ -80,9 +80,17 @@ module Dependabot
                   run_pip_compile_command(
                     "pyenv exec pip-compile -v #{pip_compile_options(filename)} -P #{dependency.name} #{filename}"
                   )
-                  # Run pip-compile a second time, without an update argument,
-                  # to ensure it handles markers correctly
-                  write_original_manifest_files unless dependency.top_level?
+
+                  next if dependency.top_level?
+
+                  # Run pip-compile a second time for transient dependencies
+                  # to make sure we do not update dependencies that are
+                  # superfluous. pip-compile does not detect these when
+                  # updating a specific dependency with the -P option.
+                  # Running pip-compile a second time will automatically remove
+                  # superfluous dependencies. Dependabot then marks those with
+                  # update_not_possible.
+                  write_original_manifest_files
                   run_pip_compile_command(
                     "pyenv exec pip-compile #{pip_compile_options(filename)} #{filename}"
                   )
--- a/python/lib/dependabot/python/file_updater/pip_compile_file_updater.rb
+++ b/python/lib/dependabot/python/file_updater/pip_compile_file_updater.rb
@@ -81,12 +81,6 @@ module Dependabot
                 "#{SharedHelpers.escape_command(version_part)}",
                 allow_unsafe_shell_command: true
               )
-              # Run pip-compile a second time, without an update argument, to
-              # ensure it resets the right comments.
-              run_pip_compile_command(
-                "pyenv exec pip-compile #{pip_compile_options(filename)} " \
-                "#{filename}"
-              )
             end
 
             # Remove any .python-version file before parsing the reqs


jurre:
Thanks for digging into this, would be interesting to get some data on how this effects real-world workloads!

jerbob92:
@jurre I have just ran the change on a real-world project of ours. It has 3 .in/.txt combinations (ci, dev and prod). It has a total of 33 direct dependencies and ~150 indirect dependencies. When running the dry run for one dependency where there is a newer version available, these are the run times:

 - On current main branch: 4m21.778s
 - With my changes: 1m52.702s
 
 As you can see the improvement is quite significant. This is also because it has to run pip-compile for every .in/.txt combination. So you can imagine the total time profit if it's 2.5m faster for every dependency that has a newer version available (it always run this logic, even if the rules don't allow going to this new version, it only doesn't run it when the latest version available is your current version).

In this case the update was not allowed, due to requirements, so the total time profit could be even bigger because it also runs pip-compile twice in the updater itself.

deivid-rodriguez:
I think my suggestion from https://github.com/dependabot/dependabot-core/issues/5903#issuecomment-1284185425 should hopefully get things green!

deivid-rodriguez (approved):
Awesome speed up! 🎉

python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
```diff
@@ -80,9 +80,13 @@ def fetch_latest_resolvable_version_string(requirement:)
...
+
+                  next if dependency.top_level?
+
+                  # Run pip-compile a second time for transient dependencies
+                  # to make sure we do not update dependencies that are
+                  # superfluous.
+                  write_original_manifest_files
                   run_pip_compile_command(
                     "pyenv exec pip-compile #{pip_compile_options(filename)} #{filename}"
                   )
```
Nishnha, line 92: Based on your comment in https://github.com/dependabot/dependabot-core/issues/5903#issuecomment-1282101954 it looks like we should also be able to get rid of this second call to `pip-compile`. Is there a reason we keep it here but remove it from the file updater?

jerbob92: If we remove it, Dependabot will create PR's for dependencies that are superfluous when the config option `dependency-type` is set to `indirect`. It's a minor inconvenience imho but I think it's better to keep the output/results of Dependabot the same as before.

We could remove it from the file updater because it appears that the reason it ran pip-compile twice in both files had a different reasons. When I removed it, it had no effect on any tests nor on the output of any of my manual tests. There is already logic to copy the old header text of the generated requirements file so there isn't really anything to "correct" it seems. It could also be that this was implemented to work around some specific behavior of pip-compile that has since been fixed.

Nishnha: Thanks for the explanation! 

Here's a guess at what's happening: in the second run, because `pip-compile` doesn't use the `-P {dependency}` flag, it's able to consider all dependencies and see that there are superfluous ones after the update. With the `-P` flag pip can't know whether other dependencies are using them.

Nishnha (approved):
Approving based on the discussion in https://github.com/dependabot/dependabot-core/issues/5903

python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
jerbob92: Yes that's exactly what happens :)

deivid-rodriguez: It also makes sense to me, and agree on including that in the comment 👍. Can you amend the commit to improve the explanation and rebase the PR?

jerbob92: @deivid-rodriguez done!

deivid-rodriguez:
Thanks so much for this @jerbob92! 🎉

PR #5906, opened 2022-10-17
Remove CI hack much less needed now

deivid-rodriguez:
We added this because our CI was pretty heavy and felt unnecessary to
run the whole thing every time ruby gems were released.

However, things have changed now so I'm not sure this is necessary.
Because:

* We release ruby gems much less often now.
* We now have other changed files filters in place, so that we only run
  ecosystem specs when files in those ecosystems change.

So, now, we would only run common and updater specs when releasing
ruby gems. This seems acceptable to me.

The reason I'd like to remove this is:

* No longer having to keep this matrix duplicated.
* Don't get duplicated jobs in the sidebar of [our actions
  page](https://github.com/dependabot/dependabot-core/actions), which is
  makes it harder to access the one you actually want (the non dummy
  one).


jakecoffman (approved):

deivid-rodriguez:
Thanks for having a look @jakecoffman!

PR #5908, opened 2022-10-17
maven: implement replaces-base to avoid calling central

jakecoffman:
It's often the case with proxy-style registries that users do not want Dependabot to try to reach Maven Central.

With this PR I've implemented the "replaces-base" functionality first used in the Python ecosystem. This means the registry with that setting will be used everywhere that Maven Central was used previously by Dependabot.

The order by which Dependabot chooses to query will still be the order the registries appear in dependabot.yml. I think this makes more sense that making the "replaces-base" registry go last as it mirrors what the Maven executable would do with a settings.xml present with the dependabot.yml registries defined, overwriting central by ID.


honeyankit (approved):
This make sense to me.

jurre:
> mirrors what the Maven executable would do with a settings.xml present with the dependabot.yml registries defined

Should we try and pull that `settings.xml` file in to determine wether to do this?

jakecoffman:
@jurre The settings.xml lives outside of the repository so we can't pull it in: [reference](https://maven.apache.org/settings.html)

I only mention it here because I think users will expect that we're taking the registries defined in dependabot.yml and placing them in a setting.xml on the system so the Maven executable will use them in the [typical order](https://maven.apache.org/guides/mini/guide-multiple-repositories.html#repository-order). Thus the "replaces-base" functionality should not actually replace the base (e.g. go to the end of the list), but just remove base and preserve the order in dependabot.yml. 

If we were to one day use the Maven executable, to achieve this same functionality we'd drop the settings.xml so Maven would pick it up, and hopefully the behavior would not change.

Issue #5888, opened 2022-10-14
PR title repeats dependency name


bdragon:
When multiple versions of a dependency are being updated, the dependency name is repeated in the PR title. Example:
`Bump ansi-regex and ansi-regex in /npm_and_yarn/helpers` in #5879 

Proposed change for this example:
`Bump ansi-regex in /npm_and_yarn/helpers`

For PRs in which multiple versions of multiple dependencies are being updated, names should be similarly de-duped:
`Bump foo, bar and bar in /directory` -> `Bump foo and bar in /directory`

Issue #5880, opened 2022-10-13
Incorrect version specified in PR metadata


jeffwidman:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code improvement description

The PR description for https://github.com/dependabot/dependabot-core/pull/5879 says it's bumping to `5.0.1`, but it's actually bumping to `3.0.1` and `4.1.1`.

The string "Updates ansi-regex from 3.0.0 to 5.0.1" should probably say "Updates ansi-regex from 3.0.0 to 3.0.1 and 4.1.0 to 4.1.1".

I'm not sure how we pick which one to show in the changelog.

More details:
https://github.com/dependabot/dependabot-core/pull/5879#issuecomment-1277097898

mctofu:
I happened to be looking at the PR title code and cleaned part of this up in https://github.com/dependabot/dependabot-core/pull/5915

> The string "Updates ansi-regex from 3.0.0 to 5.0.1" should probably say "Updates ansi-regex from 3.0.0 to 3.0.1 and 4.1.0 to 4.1.1".
> 
> I'm not sure how we pick which one to show in the changelog.

I think if we keep them as their own entries in the PR body then we could show the correct changelog for both? That would be the easiest change assuming we can get the right source/target versions for each one.

Nishnha:
I think the issue with the PR body is because `ansi-regex` is being parsed as a git registry and pulling the latest version from the registry https://github.com/dependabot/dependabot-core/blob/f28709ac6c59cb0f76c35adf3946719a4aef3f21/common/lib/dependabot/pull_request_creator/labeler.rb#L124-L131

I think we would want to have a special case for security updates?

PR #5915, opened 2022-10-18
Don't repeat dependency names in PR title

mctofu:
This improves the PR title to only mention a dependency once if it's being updated to multiple versions.

ex: https://github.com/dependabot/dependabot-core/issues/5880

Fix https://github.com/dependabot/dependabot-core/issues/5888

### DIFF
--- a/common/lib/dependabot/pull_request_creator/message_builder.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder.rb
@@ -78,8 +78,12 @@ module Dependabot
               "#{from_version_msg(old_library_requirement(dependencies.first))}" \
               "to #{new_library_requirement(dependencies.first)}"
           else
-            names = dependencies.map(&:name)
-            "requirements for #{names[0..-2].join(', ')} and #{names[-1]}"
+            names = dependencies.map(&:name).uniq
+            if names.count == 1
+              "requirements for #{names.first}"
+            else
+              "requirements for #{names[0..-2].join(', ')} and #{names[-1]}"
+            end
           end
       end
 
@@ -104,8 +108,12 @@ module Dependabot
               "#{from_version_msg(previous_version(dependency))}" \
               "to #{new_version(dependency)}"
           else
-            names = dependencies.map(&:name)
-            "#{names[0..-2].join(', ')} and #{names[-1]}"
+            names = dependencies.map(&:name).uniq
+            if names.count == 1
+              names.first
+            else
+              "#{names[0..-2].join(', ')} and #{names[-1]}"
+            end
           end
       end
 
--- a/common/lib/dependabot/pull_request_creator/message_builder.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder.rb
@@ -87,6 +87,7 @@ module Dependabot
           end
       end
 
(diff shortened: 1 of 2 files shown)


common/spec/dependabot/pull_request_creator/message_builder_spec.rb
```diff
@@ -619,20 +656,55 @@ def commits_details(base:, head:)
...
 
           it "includes both dependencies" do
             expect(pr_name).
-              to eq("Update requirements for business and business")
+              to eq("Update requirements for business and business2")
           end
         end
 
+        context "with two dependencies with the same name" do
+          let(:dependencies) { [dependency, dependency] }
```
Nishnha, line 678: Maybe we should test this with 2 different dependencies with the same name?

Nishnha (approved):
Generally this looks good to me. I left one comment on a test case but the code change itself looks fine.

common/lib/dependabot/pull_request_creator/message_builder.rb
```diff
@@ -78,11 +78,16 @@ def library_pr_name
...
               "to #{new_library_requirement(dependencies.first)}"
           else
-            names = dependencies.map(&:name)
-            "requirements for #{names[0..-2].join(', ')} and #{names[-1]}"
+            names = dependencies.map(&:name).uniq
+            if names.count == 1
+              "requirements for #{names.first}"
+            else
+              "requirements for #{names[0..-2].join(', ')} and #{names[-1]}"
+            end
```
landongrindheim, line 86: 💡 Since we have ActiveSupport available, we could use `#to_sentence` for this case ([docs](https://guides.rubyonrails.org/active_support_core_extensions.html#to-sentence))

common/lib/dependabot/pull_request_creator/message_builder.rb
```diff
@@ -104,10 +109,15 @@ def application_pr_name
...
               "to #{new_version(dependency)}"
           else
-            names = dependencies.map(&:name)
-            "#{names[0..-2].join(', ')} and #{names[-1]}"
+            names = dependencies.map(&:name).uniq
+            if names.count == 1
+              names.first
+            else
+              "#{names[0..-2].join(', ')} and #{names[-1]}"
+            end
```
landongrindheim, line 117: Same thought here re: `#to_sentence` 🙂

common/lib/dependabot/pull_request_creator/message_builder.rb
bdragon: I had this change stashed so I just pushed it up here in case you want to use any of it: https://github.com/dependabot/dependabot-core/commit/ce8eb68a4ad41dd4940d68f79dc4123e0866813b

mctofu: Thanks for the suggestions! `#to_sentence` looks ideal but because it requires adding some new active_support imports I think it'd be better to make that change as a separate PR focused on adding active_support to simplify this class.

bdragon (approved):

mctofu:
This is what the example PR looks like now:
image

We could do some additional de-duping at the start of the description but I'll leave that for a future PR.

PR #5920, opened 2022-10-18
Pyproject.toml optional-dependencies

Nishnha:
Refactors the `parse_pep621_dependencies` funciton to also parse optional-dependencies.

Currently not passing tests - the shape of the result is wrong. It should be a single list with a multiple objects but it's being returned as a list with lists of objects inside of it.

The `DEPENDENCY_TYPES` constant could be moved into the Ruby code so that `dependencies` and `optional-dependencies` can be parsed in different steps and so any additional dependency keys that need to be parsed can be appended.


deivid-rodriguez:
@Nishnha I added some changes a few days ago on #5661 to ignore pdm.lock files, I guess the conflicts are about that. I'm happy to fix those conflicts myself while I start having a look at your changes, sounds good?

PR #5928, opened 2022-10-19
fix calling npm.org when there's no npmrc with replaces-base

jakecoffman:
## Context 

Currently, when there is no `.npmrc` file, but a registry has been specified in the credentials, Dependabot generates an `.npmrc` file with a global registry, assuming all of the [resolved URLs match](https://github.com/dependabot/dependabot-core/blob/main/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npmrc_builder.rb#L92), and the update goes well. 

However, once it gets to the library check it calls out to the default registry. 

## Replaces-base

The code for checking the lockfile seems to be npm specific, and its private in another class. So it will take a bit of a refactor to get that common code out.

I thought this was a good first step to give the user a way to set a global registry with the `replaces-base` flag. 

We should probably circle back and make that npmrc generator code apply to the library check as well. Also use `replaces-base` in the npmrc generator when present.

This PR also fixes a few places where we assume the registry doesn't have a protocol. I think it's important to honor the protocol as customers may have private internal registries serving on http since they aren't exposed to the internet.

Let me know if you disagree with introducing replaces-base like this!


npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker/registry_finder_spec.rb
```diff
@@ -46,6 +46,18 @@
...
+    context "with no rc and with credentials" do
+      let(:credentials) do
+        [{
+          "type" => "npm_registry",
+          "registry" => "http://example.com",
+          "replaces-base" => true
+        }]
+      end
+
+      it { is_expected.to eq("http://example.com") }
```
mctofu, line 58: It makes sense to preserve the protocol but it currently will have limited use. On dotcom/ghes:
- the protocol is stripped on the backend
- the proxy won't attach credentials to http requests

mctofu (approved):
Makes sense to have another option to specify the registry if .npmrc isn't used!

npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker/registry_finder_spec.rb
jakecoffman: Ah, that's too bad. It will be there if we need it.

PR #5930, opened 2022-10-19
Yarn Berry: Fixes subdependency security updates

pavera:
## Context
Version resolution for security updates was broken for yarn berry if there were any packages sourced from private registries, even if the private registry was correctly configured. Previously we were not performing version resolution with yarn berry for security updates, this still worked as long as yarn 1 could talk to all of the necessary registries.

## Approach
This PR adds specific version resolution calls for yarn berry to the subdependency version resolver so that it doesn't attempt to use yarn 1 when performing this resolution.

## Ongoing
This still requires that any package mentioned in yarn.lock or package.json is reachable by Dependabot. This means if there are packages sourced from private registries and Dependabot does not have access to those registries due to lack of configuration or incorrect configuration, a subdependency security update will still fail.


honeyankit (approved):
Looks good.

PR #5932, opened 2022-10-20
Fix crash when updating Python libraries with multiple manifest types

deivid-rodriguez:
I discovered this issue while wrapping up #5661. However, the issue is already there so I think it makes sense to extract it to its own PR.

If a library has both a pyproject.toml file and a standard requirements.txt file, we'd end up using the `:widen` strategy for the dependencies in the `requirements.txt` file if the project is a library, and eventually crashing with an error like the following:

```
/home/dependabot/dependabot-core/common/lib/dependabot/update_checkers/base.rb:266:in `block in preferred_version_resolvable_with_unlock?': undefined method `[]' for nil:NilClass (NoMethodError)
    
       updated_requirements.none? { |r| r[:requirement] == :unfixable }
                                              ^^^^^^^^^^^^^^
        from /home/dependabot/dependabot-core/common/lib/dependabot/update_checkers/base.rb:266:in `none?'
        from /home/dependabot/dependabot-core/common/lib/dependabot/update_checkers/base.rb:266:in `preferred_version_resolvable_with_unlock?'
        from /home/dependabot/dependabot-core/common/lib/dependabot/update_checkers/base.rb:249:in `numeric_version_can_update?'
        from /home/dependabot/dependabot-core/common/lib/dependabot/update_checkers/base.rb:199:in `version_can_update?'
        from /home/dependabot/dependabot-core/common/lib/dependabot/update_checkers/base.rb:44:in `can_update?'
        from bin/dry-run.rb:709:in `block in <main>'
        from bin/dry-run.rb:661:in `each'
        from bin/dry-run.rb:661:in `<main>'
```
    
I think the crash happens because the requirements.txt file updater does not supoort the `:widen` strategy. So my fix is to fallback to `increase` in this case, since requirements.txt files usually include pinned dependencies so widening probably doesn't make much sense there.

### DIFF
--- a/python/lib/dependabot/python/update_checker.rb
+++ b/python/lib/dependabot/python/update_checker.rb
@@ -259,7 +259,7 @@ module Dependabot
       end
 
       def poetry_library?
-        return false unless pyproject
+        return false unless updating_pyproject?
 
         # Hit PyPi and check whether there are details for a library with a
         # matching name and description
--- a/python/lib/dependabot/python/update_checker.rb
+++ b/python/lib/dependabot/python/update_checker.rb
@@ -89,7 +89,7 @@ module Dependabot
 
       def updated_requirements
         RequirementsUpdater.new(
-          requirements: dependency.requirements,
+          requirements: requirements,
           latest_resolvable_version: preferred_resolvable_version&.to_s,
           update_strategy: requirements_update_strategy,
           has_lockfile: !(pipfile_lock || poetry_lock || pyproject_lock).nil?
@@ -133,8 +133,7 @@ module Dependabot
       end
 
       def resolver_type
-        reqs = dependency.requirements
-        req_files = reqs.map { |r| r.fetch(:file) }
+        reqs = requirements
 
         # If there are no requirements then this is a sub-dependency. It
         # must come from one of Pipenv, Poetry or pip-tools, and can't come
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Setting as draft because I'm not sure this is desired behavior. Nobody has complained about this so maybe best to not create the PR and just skip the crash.

deivid-rodriguez:
I reviewed the repository where I observed this, and repo maintainer was [happily merging dependabot PRs](https://github.com/Textualize/rich/pull/2423) before this bug appeared. So I say let's go with this approach and restore those PRs.

python/lib/dependabot/python/update_checker.rb
```diff
@@ -283,6 +280,26 @@ def poetry_library?
         false
       end
 
+      def updating_pipfile?
+        requirement_files.any?("Pipfile")
+      end
+
+      def updating_pyproject?
```
pavera, line 287: Would it be better/more clear if we used `poetry.lock` for this determination and renamed this method `updating_poetry`? Since `pyproject.toml` is used by other package managers like flit and hatch I think tying poetry to pyproject.toml may be misleading.

deivid-rodriguez: Right, I think this will make more sense with #5661, where this logic will apply to all ways of updating `pyproject.toml`. For now it doesn't make a difference because Poetry is the only supported way of updating a `pyproject.toml` file. In conclusion, I'm happy to change it but I'll probably change it back in #5661 so probably not worth it?

pavera (approved):

PR #5937, opened 2022-10-20
fix race and updating local mounted repositories

jakecoffman:
Currently the Updater:
1. Gets the base commit
2. Fetches manifest files
3. Clones the repo (for ecosystems that clone)

This seems like a race. To get the base commit it makes a call to get the default branch (if one isn't specified) and then another call to get the HEAD commit, then it clones. That leaves some space where a user could have pushed up a change, and so the base commit and cloned repo will be different.

It makes more sense to me to clone, then get the base commit and fetch the files from the locally cloned repo.

This also fixes the issue that we have when testing with local repositories with the Dependabot CLI. We were having to pass the `vendor_dependencies` flag to trick Dependabot into using the local repo. Now with `already_cloned?` Dependabot can tell it can use the local repo.


bdragon (approved):
Makes sense to me 👍

jakecoffman:
When deploying this I found a new error pop up caused by trying to get the base_commit_sha after the clone failed because the branch doesn't exist. Some users have a branch that does not exist in their dependabot.yml. 

We report the SHA based on the branch, and so if the clone failed then we can't really tell what branch we're on. So to prevent from hiding the root cause I added this check to see if this was really a git repo before trying to use it.

This is somewhat minor, the job used the fail and it fails now too, but at least it reports what the issue is more accurately.

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -473,6 +474,16 @@ def _fetch_file_content_from_github(path, repo, commit)
       end
       # rubocop:enable Metrics/AbcSize
 
+      def cloned_commit
+        return if repo_contents_path.nil? || !File.directory?(File.join(repo_contents_path, ".git"))
```
deivid-rodriguez, line 478: I noticed that this logic is now duplicated with `Job#already_cloned?`. That makes me wonder (other than whether that shows any refactoring opportunities), does it make sense to also add this extra check to other places in the fetcher that check for `repo_contents_path.nil?`?

jakecoffman: Yeah it was an awkward refactor too since we need to check before constructing a file_fetcher, and the file_fetcher doesn't get the job so I can't just call the method on the job. I punted and opted for the duplication for now.

I think since this is just an error path any other checks for repo_contents_path is safe? If we get past the clone step without throwing we can safely assume the repo exists there now.

PR #5939, opened 2022-10-20
Create a merge-queue script

deivid-rodriguez:
To ease creation of a "Current merge train" PR.

I used it to generate #5938.

To use it:

```
$ bin/merge-queue.rb <pr1> ... <prN>
```

If a merge queue PR already exists, PRs are added to it. Otherwise, it creates a new merge queue PR.

What do you think?

### DIFF
--- a/bin/merge-queue.rb
+++ b/bin/merge-queue.rb
@@ -1,0 +1,64 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require "open3"
+
+module MergeQueue
+  extend self
+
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Actually this is not working as intended because the PR was generated out of date with latest main 🤔.

deivid-rodriguez:
Whenever you want to do something like this, I think creating a separate branch, merging the PRs to be batched into it, and creating a new PR from the resulting branch works well enough I believe. So I'll close this.

PR #5948, opened 2022-10-21
No public url call when public registry is disabled

honeyankit:
## Context

When pubic registry is disabled in the Nuget, it still tries to make a call to public registry url (`https://api.nuget.org/v3-flatcontainer/nuget.versioning/6.2.1/nuget.versioning.nuspec`) to get the nuspec file if no nuspec file url is provided.

## Solution

Earlier, in the `dependency_nuspec_url` function of the metadata finder, a default public nuspec url was added if the source was missing or if the source was present but had no nuspec file. I have removed the injection of the default public nuspec url when there is either no source or no nuspec url specified in the source.

Even without my change from this PR, when nuspec url was nil in source, the metadata finder was calling the default nuspec url and returning 404.

## Note
Nuspec file: A .nuspec file is an XML manifest that contains package metadata. This manifest is used both to build the package and to provide information to consumers. The manifest is always included in a package. [[ref](https://learn.microsoft.com/en-us/nuget/reference/nuspec)]

### DIFF
--- a/nuget/lib/dependabot/nuget/metadata_finder.rb
+++ b/nuget/lib/dependabot/nuget/metadata_finder.rb
@@ -122,10 +122,7 @@ module Dependabot
                  find { |r| r&.fetch(:source) }&.fetch(:source)
 
         if source&.key?(:nuspec_url)
-          source.fetch(:nuspec_url) ||
-            "https://api.nuget.org/v3-flatcontainer/" \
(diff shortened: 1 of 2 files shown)


jakecoffman (approved):
Makes sense, if the URL is empty we shouldn't automatically fall back to the public registry.

PR #5951, opened 2022-10-22
Update version comments for SHA-pinned GitHub Actions

### LINKED ISSUE #4691: For actions that are pinned-by-hash, bump the human readable version number in the code comment
Hello!

As good security practice and guided by the code scanning alert for 'Pinned-Dependencies' from the [ossf/scorecard](https://github.com/ossf/scorecard) project, users are encouraged to pin GitHub workflow actions by hash. The [example provided by the scorecard repo](https://github.com/ossf/scorecard/blob/f55b86d6627cc3717e3a0395e03305e81b9a09be/.github/workflows/main.yml#L27)includes a comment following the action+pinned-hash, example below:
```yaml
     - name: Clone the code
       uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f # v2.3.4
```

Subsequent updates by dependabot do not bump the version in the comment, leading to confusion and incorrect information. An example of this can be found here: ChrisCarini/environment-variable-settings-summary-intellij-plugin#36

Below is a before & after example of the above linked PR

### Before

Screen Shot 2022-01-31 at 22 53 54

### After

Screen Shot 2022-01-31 at 22 52 13

### Summary

Dependabot updated `ossf/scorecard-action` from version `v1.0.1` to `v1.0.2`.

In the "Before", the hash was updated correctly, however the trailing comment with the tag version, was not.

| Hash                                     | Tag Version |
|------------------------------------------|-------------|
| e3e75cf2ffbf9364bbff86cdbdf52b23176fe492 | v1.0.1      |
| c8416b0b2bf627c349ca92fc8e3de51a64b005cf | v1.0.2      |

I believe this change would help GitHub workflow owners to have a better security posture w.r.t. using hash-pinned GitHugb action dependencies, while also having improved ergonomics to be able to quickly verify they are on the correct version.

This idea shares similar sentiment of dependabot/dependabot-core#3699

_(I would consider trying to make a code change for this, however I lack expertise in Ruby and have thus far struggled to get the project opened for development.)_

jproberts:
GitHub encourages pinning third-party GitHub Actions to a full length commit SHA. It's common for actions pinned by commit SHA to include a comment specifying the version associated with the commit. For example:

    - uses: actions/checkout@01aecc # v2.1.0

This change updates the GitHub Actions manager to bump versions in comments that follow SHA-pinned actions, so the comment stays up-to-date with the SHA being updated.

The file_updater now searches the comment string for all references to the previous version and replaces them with the new version. To avoid changing unrelated comments, the comment updater only updates dependencies that pin SHA refs.

---

> **Note**
> I'm not an experienced Ruby developer and I welcome any suggestions to improve the code. My main goal was to make a minimal change and follow the existing code conventions.

Closes  #4691

### DIFF
--- a/github_actions/lib/dependabot/github_actions/file_updater.rb
+++ b/github_actions/lib/dependabot/github_actions/file_updater.rb
@@ -87,7 +87,9 @@ module Dependabot
 
       def updated_version_comment(comment, new_req)
         raise "No comment!" unless comment
+        comment = comment.rstrip
         return unless dependency.previous_version && dependency.version
+        return unless comment.end_with? dependency.previous_version
 
         git_checker = Dependabot::GitCommitChecker.new(dependency: dependency, credentials: credentials)
         return unless git_checker.ref_looks_like_commit_sha?(new_req.fetch(:source).fetch(:ref))
--- a/github_actions/lib/dependabot/github_actions/file_updater.rb
+++ b/github_actions/lib/dependabot/github_actions/file_updater.rb
@@ -65,16 +65,34 @@ module Dependabot
             gsub(/@.*+/, "@#{new_req.fetch(:source).fetch(:ref)}")
 
           # Replace the old declaration that's preceded by a non-word character
-          # and followed by a whitespace character (comments) or EOL
+          # and followed by a whitespace character (comments) or EOL.
+          # If the declaration is followed by a comment, attempt to update
+          # any version comments associated with SHA source refs.
           updated_content =
             updated_content.
             gsub(
-              /(?<=\W|"|')#{Regexp.escape(old_declaration)}(?=\s|"|'|$)/,
-              new_declaration
-            )
+              /(?<=\W|"|')#{Regexp.escape(old_declaration)}(?<comment>\s+#.*)?(?=\s|"|'|$)/
+            ) do |match|
+              comment = Regexp.last_match(:comment)
+              match.gsub!(old_declaration, new_declaration)
+              if comment && (updated_comment = updated_version_comment(comment, new_req))
+                match.gsub!(comment, updated_comment)
+              end
+              match
+            end
         end
 
         updated_content
+      end
+
+      def updated_version_comment(comment, new_req)
+        raise "No comment!" unless comment
+        return unless dependency.previous_version && dependency.version
+
+        git_checker = Dependabot::GitCommitChecker.new(dependency: dependency, credentials: credentials)
+        return unless git_checker.ref_looks_like_commit_sha?(new_req.fetch(:source).fetch(:ref))
+
+        comment.gsub(dependency.previous_version, dependency.version)
       end
     end
   end
--- a/github_actions/lib/dependabot/github_actions/file_updater.rb
+++ b/github_actions/lib/dependabot/github_actions/file_updater.rb
@@ -90,6 +90,7 @@ module Dependabot
 
       def updated_version_comment(comment, new_req)
         raise "No comment!" unless comment
+
         comment = comment.rstrip
         return unless dependency.previous_version && dependency.version
         return unless comment.end_with? dependency.previous_version
--- a/github_actions/lib/dependabot/github_actions/file_updater.rb
+++ b/github_actions/lib/dependabot/github_actions/file_updater.rb
@@ -66,8 +66,11 @@ module Dependabot
 
           # Replace the old declaration that's preceded by a non-word character
           # and followed by a whitespace character (comments) or EOL.
-          # If the declaration is followed by a comment, attempt to update
-          # any version comments associated with SHA source refs.
+          # If the declaration is followed by a comment that lists the version associated
+          # with the SHA source ref, then update the comment to the human-readable new version.
+          # However, if the comment includes additional text beyond the version, for safety
+          # we skip updating the comment in case it's a custom note, todo, warning etc of some kind.
+          # See the related unit tests for examples.
           updated_content =
             updated_content.
             gsub(
--- a/github_actions/spec/fixtures/workflow_files/pinned_sources_version_comments.yml
+++ b/github_actions/spec/fixtures/workflow_files/pinned_sources_version_comments.yml
@@ -11,7 +11,20 @@ jobs:
     - uses: actions/checkout@01aecccf739ca6ff86c0539fbc67a7a5007bbc81 # tag=v2.1.0
     - uses: actions/checkout@01aecccf739ca6ff86c0539fbc67a7a5007bbc81  #   v2.1.0
     - uses: actions/checkout@01aecccf739ca6ff86c0539fbc67a7a5007bbc81 #v2.1.0
+    # The comment on the next line has a trailing tab. The version should still be updated.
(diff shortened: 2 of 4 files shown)


deivid-rodriguez:
@jproberts Thanks so much for doing this! The PR looks pretty neat and seems to have extensive test coverage 💪. I'll hope to get to do a proper review of this PR soon-ish unless someone else beat me to it, but I wanted to drop a big THANKS for now. Also pretty nice detail to open a PR to your fork first and get a colleague to review it 🥇.

github_actions/spec/fixtures/workflow_files/pinned_sources_version_comments.yml
```diff
@@ -0,0 +1,17 @@
...
+  chore:
+    steps:
+    - uses: actions/checkout@01aecccf739ca6ff86c0539fbc67a7a5007bbc81 # v2.1.0
+    - uses: actions/checkout@01aecccf739ca6ff86c0539fbc67a7a5007bbc81 # 2.1.0
+    - uses: actions/checkout@01aecccf739ca6ff86c0539fbc67a7a5007bbc81 # @v2.1.0
+    - uses: actions/checkout@01aecccf739ca6ff86c0539fbc67a7a5007bbc81 # pin @v2.1.0
+    - uses: actions/checkout@01aecccf739ca6ff86c0539fbc67a7a5007bbc81 # tag=v2.1.0
+    - uses: actions/checkout@01aecccf739ca6ff86c0539fbc67a7a5007bbc81  #   v2.1.0
+    - uses: actions/checkout@01aecccf739ca6ff86c0539fbc67a7a5007bbc81 #v2.1.0
+    - uses: actions/checkout@01aecc # v2.1.0
```
jeffwidman, line 14: Major 💖 for phenom set of test cases.

Two more cases:
1. What about the case where a comment includes the version, but then some text?

```
- uses: actions/checkout@01aecccf739ca6ff86c0539fbc67a7a5007bbc81 # Versions older than v2.1.0 have a security vulnerability
```

2. What if the version string is repeated twice?

```
- uses: actions/checkout@01aecccf739ca6ff86c0539fbc67a7a5007bbc81 # v2.1.0 - Versions older than v2.1.0 have a security vulnerability
```

I don't want to throw up roadblocks on this awesome PR, so if for simplicity neither case is bumped, that's perfectly acceptable. So one simple heuristic could be _if there's any text after the version string, then don't bump the version string_... Users that still want it bumped can instead leave their comment as a full-line comment above the actual code line:
```
# Versions older than v2.1.0 have a security vulnerability
- uses: actions/checkout@01aecccf739ca6ff86c0539fbc67a7a5007bbc81 # v2.1.0
```

But we do need to ensure we never auto-bump the comment to start incorrectly saying `"Versions older than v2.2.0 have a security vulnerability"`...

So can you add these two test cases to ensure no future regressions?

jeffwidman (commented):
This looks fantastic to me, very well done.

Well done writing this flexibly enough to handle all the permutations people use for specifying version strings.

github_actions/spec/fixtures/workflow_files/pinned_sources_version_comments.yml
jproberts: Thanks for the feedback. I was worried the simple search-and-replace was too broad but I couldn't come up with a sample comment where we wouldn't want to bump the version. The cases you mentioned are exactly what I was looking for!

I've updated the PR with the simple fix to only change comments if there's no text after the version string.

github_actions/lib/dependabot/github_actions/file_updater.rb
```diff
@@ -65,17 +65,37 @@ def updated_workflow_file_content(file)
             gsub(/@.*+/, "@#{new_req.fetch(:source).fetch(:ref)}")
 
           # Replace the old declaration that's preceded by a non-word character
-          # and followed by a whitespace character (comments) or EOL
+          # and followed by a whitespace character (comments) or EOL.
+          # If the declaration is followed by a comment, attempt to update
+          # any version comments associated with SHA source refs.
```
jeffwidman, line 70: What do you think about tweaking this slightly?

Something like:
```
# If the declaration is followed by a comment that lists the version associated
# with the SHA source ref, then update the comment to the human-readable new version.
# However, if the comment has additional text beyond the version, for safety
# we skip updating it in case it's a custom note, todo, warning etc of some kind.
```

jeffwidman (approved):
Looks great to me.

Had one minor pedantic suggestion about tweaking the description of intended behavior, but I'm also fine to merge as-is... up to you.

We should be able to get this deployed/merged in the next few days...

jeffwidman:
For the CI failure here, it's probably best to just disable the Rubocop line-length check for that block of code:
https://github.com/dependabot/dependabot-core/actions/runs/3357784436/jobs/5564093603

Example of how to do that here:
https://github.com/dependabot/dependabot-core/commit/08c679d9742ec87a74021fb737fa177c700e5dc4#diff-ee98e028c59b193d58fde56ab4daf54d43c486ae674e63d50ddf300b07943e0fR57

Just don't forget to re-enable it after the block as demonstrated in the above example as well.

jeffwidman (approved):

jeffwidman:
FYI @jproberts this got a nice shout out on the blog: https://github.blog/changelog/2022-10-31-dependabot-now-updates-comments-in-github-actions-workflows-referencing-action-versions/

jproberts:
I just saw the post and I'm amazed. Thanks so much for your support @jeffwidman ❤️

laurentsimon:
Congrats on the post!

Is this feature compatible with renovatebot's way of handling comments? Renovatebot expects a comment of the form `# tag=v1.2.3`

It would be useful for the various update tools to be compatible on this point. Wdut?

/cc @rarkins from renovatebot

rarkins:
I have created https://github.com/renovatebot/renovate/issues/18640 recently to align Renovate with this approach. As described in that issue, I have concerns about assuming _any_ comment represents a tag (e.g. `# do not change this`) so we will make sure it's version-ish before treating it as a tag. I think it would have been better to adopt the explicit `# tag=` approach for everyone but it's likely too late for that.

jeffwidman:
Thanks for swinging by. Agreed that getting the update tools aligned on the implicit format spec would be great.

The `# v2.1.0` format is so intuitive that I personally prefer it, despite the ambiguity compared to `# tag=v2.1.0`.

Thankfully @jproberts was thorough and already included a unit test for both these formats:
https://github.com/dependabot/dependabot-core/blob/45f52b300ae163d6fb221fceb7965e016eaf84de/github_actions/spec/fixtures/workflow_files/pinned_sources_version_comments.yml#L7-L30

@rarkins you might find the whole set of examples useful to test against, particularly the negative tests. If you come up with additional edge cases we missed, feel free to circle back and let us know.

kuhnroyal:
Thanks for this, saves a lot of time!

crenshaw-dev:
Do I need to do anything to opt in to this? Not seeing the version update: https://github.com/argoproj/argo-cd/pull/11480/files

Issue #5017, opened 2022-04-20
Dependabot not detecting the latest version of a Github action


tyrann0us:
**Package ecosystem**
* github-actions

**Manifest location and content before the Dependabot update**
* e.g. https://github.com/inpsyde/reusable-workflows/blob/main/.github/workflows/assets-compilation.yml#L64-L67

**dependabot.yml content**
```yml
version: 2
updates:
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "monthly"
```

**Updated dependency**
* nothing was updated

**What you expected to see, versus what you actually saw**
The workflow references the [`ramsey/composer-install`](https://github.com/ramsey/composer-install) action like so:
```yml
        uses: ramsey/composer-install@v1
```

In the Dependabot [logs](https://github.com/inpsyde/reusable-workflows/network/updates/346693526) I see:
```
updater | INFO <job_346693526> Checking if ramsey/composer-install 1 needs updating
  proxy | 2022/04/14 20:12:36 [044] GET https://github.com:443/ramsey/composer-install.git/info/refs?service=git-upload-pack
  proxy | 2022/04/14 20:12:36 [044] * authenticating git server request (host: github.com)
  proxy | 2022/04/14 20:12:36 [044] 200 https://github.com:443/ramsey/composer-install.git/info/refs?service=git-upload-pack
updater | INFO <job_346693526> Latest version is 994bb194a4fefcf39449ccf0f7766a4318f1ac76
updater | INFO <job_346693526> No update needed for ramsey/composer-install 1
```

There are a couple of things off:
* The latest version is `v2` (or to be correct [`v2.1.0`](https://github.com/ramsey/composer-install/releases/tag/2.1.0) at the time of writing)
* The latest version's commit hash is [`f680dac46551dffb2234a240d65ae806c2999dd6`](https://github.com/ramsey/composer-install/commit/f680dac46551dffb2234a240d65ae806c2999dd6)
* The mentioned commit hash [`994bb194a4fefcf39449ccf0f7766a4318f1ac76`](https://github.com/ramsey/composer-install/commit/994bb194a4fefcf39449ccf0f7766a4318f1ac76) is the last commit on the [`v1`](https://github.com/ramsey/composer-install/commits/v1) _branch_
* But it's actually part of the [`v2`](https://github.com/ramsey/composer-install/compare/1.3.0...2.0.0) _tag_

I would expect Dependabot to create a PR and update this action to `v2`.

[...]

PR #5963, opened 2022-10-24
Fix updating GitHub actions to major tags that are branches

deivid-rodriguez:
If users have something like in the workflow file

```yaml
    - name: Install vcpkg
      uses: lukka/run-vcpkg@v7
```

And the given action, in this case [`lukka/run-vcpkg`](https://github.com/lukka/run-vcpkg) keeps the major version reference as a branch (as opposed to a tag), then Dependabot would not be able to create updates for that.

This PR fixes the problem.

Closes https://github.com/dependabot/dependabot-core/issues/5017.


deivid-rodriguez:
This one is ready now.

It fixes a pretty common case Dependabot could not yet update! 🎉

common/lib/dependabot/git_commit_checker.rb
```diff
@@ -52,17 +52,17 @@ def pinned?
...
 
-      # If the specified `ref` is actually a branch, we're NOT pinned
-      return false if local_upload_pack.match?(%r{ refs/heads/#{ref}$})
+      # Assume we're pinned unless the specified `ref` is actually a branch
+      return true unless local_upload_pack.match?(%r{ refs/heads/#{ref}$})
 
-      # Otherwise, assume we're pinned
-      true
+      # If the specified `ref` is actually a branch, we're pinned if the branch looks like a version
+      version_tag?(ref)
```
brrygrdn, line 59: ❓ Could there be any unintended consequences of making this change in the `common/lib`?

The release branch pattern is very common to `github-actions` and this makes total sense for that ecosystem, but I'm a little concerned that this change in behaviour may mean that in other languages where people are tracking a version branch for other reasons we will start suggesting upgrades where we haven't before 🤔

deivid-rodriguez: 🤔 good point!

Seems a very edge case in other ecosystems, and I'm not sure what user's expectation is, but probably they don't want an update? Not really sure...

I made a similar change about a month ago: https://github.com/dependabot/dependabot-core/commit/e1a8c49f24e41b4ad368aff3a7d35b76aa8c96cb, by the way 😬.

What should we do? Should I override `pinned?` with this logic just for Actions?

deivid-rodriguez: I believe https://github.com/dependabot/dependabot-core/commit/e1a8c49f24e41b4ad368aff3a7d35b76aa8c96cb probably makes sense everywhere (and no issues reported so far, which is good).

Something else I'm observing is that this method returns `true` if an explicit branch argument is given. So I believe this should not affect Bundler, which has an explicit `:branch` option when it comes to dealing with `git` sources. In other words, I think this should only potentially affect ecosystems with an "ambiguous notation" for git refs.

deivid-rodriguez: Hei @brrygrdn! Does it look ok now? I'd like to move this one forward if so.

common/lib/dependabot/git_commit_checker.rb
```diff
@@ -52,17 +54,17 @@ def pinned?
...
       return true if local_upload_pack.match?(%r{ refs/tags/#{ref}$})
 
-      # If the specified `ref` is actually a branch, we're NOT pinned
-      return false if local_upload_pack.match?(%r{ refs/heads/#{ref}$})
+      # Assume we're pinned unless the specified `ref` is actually a branch
+      return true unless local_upload_pack.match?(%r{ refs/heads/#{ref}$})
 
-      # Otherwise, assume we're pinned
-      true
+      # TODO: Research whether considering branches that look like versions pinned makes sense for all ecosystems
```
brrygrdn, line 60: 👍🏻

common/lib/dependabot/git_commit_checker.rb
```diff
@@ -23,13 +23,15 @@ class GitCommitChecker
...
-                   requirement_class: nil, version_class: nil)
+                   requirement_class: nil, version_class: nil,
+                   consider_version_branches_pinned: false)
       @dependency = dependency
       @credentials = credentials
       @ignored_versions = ignored_versions
       @raise_on_ignored = raise_on_ignored
       @requirement_class = requirement_class
       @version_class = version_class
+      @consider_version_branches_pinned = consider_version_branches_pinned
```
brrygrdn, line 34: I like the approach of parameterising this rather than overriding it 🥇

brrygrdn (approved):
Thanks for the update, this looks great! I like that we can now progressively roll out the behaviour to ecosystems where we determine it makes sense rather than having to figure it all out centrally 😄

deivid-rodriguez:
Awesome, thanks so much @brrygrdn!

PR #5966, opened 2022-10-25
Fix Rubocop warnings about mutable constants that should be frozen

jeffwidman:
Rubocop is warning about mutable constants that should be frozen.

These warnings started showing up as part of the Ruby 3.1 upgrade, and were excluded from that PR in order to minimize noise.

So this PR fixes those warnings.


jeffwidman:
Not sure why this is passing... locally this fails all over the place for me... 😕 ?? May need https://github.com/dependabot/dependabot-core/pull/5959 to land first, as locally I'm running `1.37.1` already in the docker dev shell somehow.

jeffwidman:
This throws a lot fewer warnings for me now, not sure if we've changed our code or upstream fixed some bugs... regardless, the changes from `rubocop -A` still look reasonable to me, so marking this as ready for review/merge.

landongrindheim (approved):

PR #5968, opened 2022-10-25
Remove Current User From List of Default Reviewer

Kimor-hello:
Creating a pull request on bitbucket automatically fetch all default reviewer and add it to the reviewer list. However if the user logged in to use the dependabot script on bitbucket is one of the default reviewer, the pull request creation will fail.
To avoid this, this fetch the current user uuid and remove it from the default reviewer list if necessary.


common/.ruby-version
```diff
@@ -0,0 +1 @@
+3.1.2
```
jeffwidman, line 1: Is there a reason for this? We already have https://github.com/dependabot/dependabot-core/blob/efaca437c99258fdedf61ec7d6e498346aa59f44/.ruby-version

jeffwidman (commented):
Looks fine to me other than I am unclear why add the `.ruby-version` file?

To be clear, I have no way to integration test the client, it's very much a "we're happy to sanity check the code, but otherwise up to the community to maintain". Thank you for the unit test, that certainly helps.

common/.ruby-version
Kimor-hello: Hello, 

Sorry no, I was testing some options in you read me in order to test it in my own custom usage with dependabot script, will remove it.

jeffwidman (approved):
Looks good, thanks for including the unit test!

PR #5969, opened 2022-10-25
Actually consider development dependencies

deivid-rodriguez:
This fixes the issue reported by @jeffwidman at  https://github.com/dependabot/dependabot-core/pull/5966#issuecomment-1289837387.

Namely, we have a `~> 1.36.0` constraint on RuboCop, and still with the `1.37.0` release of it, now we're getting it installed in our development image.

The problem is fixed by 266038a6e63a4f2a49587f57feb72093d45cbf54, by I figured if development dependencies for subgems don't buy us much, so we can unify everything in the omnibus Gemfile.


deivid-rodriguez:
I reverted the additional commit to drop development dependencies from subgems, since CI showed it's actually needed.

jurre (approved):

PR #5972, opened 2022-10-25
Keep updater lockfile in sync with subgem changes

deivid-rodriguez:
This is not a big problem, specially now that we're not pushing new gems to rubygems.org. But the lock file could still get out of sync unintentionally and this should prevent that.

This should clear #5858 from any changes other than the Bundler upgrade.

### DIFF
--- a/Dockerfile.updater
+++ b/Dockerfile.updater
@@ -33,6 +33,7 @@ COPY --chown=dependabot:dependabot common ${DEPENDABOT_HOME}/common
 WORKDIR $DEPENDABOT_HOME/dependabot-updater
 
 RUN bundle config set --local path 'vendor' && \
+bundle config set --local frozen 'true' && \
 bundle config set --local without 'development' && \
 bundle install
 
--- a/updater/Gemfile.lock
+++ b/updater/Gemfile.lock
@@ -1,19 +1,19 @@
 PATH
   remote: ../bundler
   specs:
-    dependabot-bundler (0.211.0)
-      dependabot-common (= 0.211.0)
+    dependabot-bundler (0.212.0)
+      dependabot-common (= 0.212.0)
 
 PATH
   remote: ../cargo
   specs:
-    dependabot-cargo (0.211.0)
-      dependabot-common (= 0.211.0)
+    dependabot-cargo (0.212.0)
+      dependabot-common (= 0.212.0)
 
 PATH
   remote: ../common
   specs:
-    dependabot-common (0.211.0)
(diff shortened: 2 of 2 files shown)


Dockerfile.updater
```diff
@@ -33,6 +33,7 @@ COPY --chown=dependabot:dependabot common ${DEPENDABOT_HOME}/common
 WORKDIR $DEPENDABOT_HOME/dependabot-updater
 
 RUN bundle config set --local path 'vendor' && \
+bundle config set --local frozen 'true' && \
```
jeffwidman, line 36: So IIUC this will ensure the built docker image has the updated `Gemfile.lock`, but is there a chance that copy of the `Gemfile.lock` will start to drift from the committed copy? IE, what ensures that as this version increases that the increase also makes it back to the committed code?? Should we also add a `git status --porcelain` somewhere?

jeffwidman (approved):
Thanks!

Had one question about enforcing no drift, but if needed that can also be done as a separate PR, so approving this one.

Dockerfile.updater
deivid-rodriguez: I didn't understand the question, which version do you mean when you say "this version increases", the one in `version.rb` files of every gem?

What the `frozen` configuration ensures is that whenever Bundler runs, it will abort if the `Gemfile.lock` is out of date with the `Gemfile`. In other words, that running Bundler won't cause any changes in the `Gemfile.lock` file.

deivid-rodriguez: Another way to explain this, we're already using `aws-sdk-core 3.164.0` and other dependencies "upgraded" here, as can be seen in the [updater build logs](https://github.com/dependabot/dependabot-core/actions/runs/3319986944/jobs/5485825132) of the last build in the main branch:

```
Installing rubocop-ast 1.23.0
Fetching aws-sdk-core 3.164.0 (was 3.137.0)
Installing aws-sdk-core 3.164.0 (was 3.137.0)
Fetching webmock 3.18.1
```

This PR just makes this explicit and prevents this kind of divergence from happening in the future, but it shouldn't cause the updater to use any more up to date gems than the ones it's using now.

jeffwidman: Perfect, this code does exactly what I want than... I didn't understand the `frozen` flag... I was thinking that this caused bundler to regenerate the lockfile during the build step (aka `pip freeze` which i'm more familiar with). If it just means it regenerates the lock file and compares and flags divergence and throws an error, then it is exactly what we want of preventing drift.

PR #5974, opened 2022-10-25
Unify development dependencies

deivid-rodriguez:
Just curious to see what breaks if we unify our development dependencies.

### DIFF
--- a/updater/Gemfile
+++ b/updater/Gemfile
@@ -27,12 +27,24 @@ gem "sentry-raven", "~> 3.1"
 gem "terminal-table", "~> 3.0.2"
 
 group :test do
-  gem "debug", "~> 1.7.1"
-  gem "rspec", "~> 3.12"
-  gem "rubocop", "~> 1.42.0"
-  gem "rubocop-performance", "~> 1.15.2"
-  gem "vcr", "~> 6.1"
-  gem "webmock", "~> 3.18"
+  common_gemspec = File.expand_path("../common/dependabot-common.gemspec", __dir__)
+
+  deps_shared_with_core = %w(
+    debug
+    rspec
+    rubocop
+    rubocop-performance
+    vcr
+    webmock
+  )
+
+  Dir.chdir(File.dirname(common_gemspec)) do
+    Gem::Specification.load(common_gemspec).development_dependencies.each do |dep|
+      next unless deps_shared_with_core.include?(dep.name)
+
+      gem dep.name, *dep.requirement.as_list
+    end
+  end
 end
 
 group :development do
--- a/common/dependabot-common.gemspec
+++ b/common/dependabot-common.gemspec
@@ -35,19 +35,19 @@ Gem::Specification.new do |spec|
   spec.add_dependency "parser", ">= 2.5", "< 4.0"
   spec.add_dependency "toml-rb", ">= 1.1.2", "< 3.0"
 
-  spec.add_development_dependency "debug", ">= 1.0.0"
+  spec.add_development_dependency "debug", "~> 1.7.1"
   spec.add_development_dependency "gpgme", "~> 2.0"
   spec.add_development_dependency "parallel_tests", "~> 4.1.0"
   spec.add_development_dependency "rake", "~> 13"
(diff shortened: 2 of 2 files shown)


updater/Gemfile
```diff
@@ -2,39 +2,30 @@
...
-group :test do
-  gem "debug", "~> 1.6.2"
-  gem "rspec", "~> 3.11"
-  gem "rubocop", "~> 1.37.1"
-  gem "rubocop-performance", "~> 1.15.0"
-  gem "vcr", "~> 6.1"
-  gem "webmock", "~> 3.18"
-end
-
 group :development do
```
jeffwidman, line 29: The PR description says unifying the development dependencies, but looks like this is only unifying the `test` deps? As `dev` still has `licensed`?

Also, there's been internal discussion about potentially adding license management to the rest of `dependabot-core` beyond `updater`, in which case this would probably also go away too?

jeffwidman (approved):
Looks like nothing breaks. 😀

deivid-rodriguez:
The downside of this PR is that some dependencies are no longer locked in CI now. That means CI is less deterministic and potentially could get affected by issues in dependencies. For example, I guess there was a reason why `vcr` was pinned to 6.1.0. Now it's more unlocked.

But this may be acceptable.

deivid-rodriguez:
After a second look, the only dependency that got relaxed with these changes is `vcr`. I'm happy to change this to keep that pin.

deivid-rodriguez:
I was not too happy about the previous approach, because it was adding dependencies to the updater that it didn't really need. I changed the approach so that the set of dependencies used by `/updater` is exactly the same.

It's probably a bit too hacky, but essentially the same thing other gemspecs are doing by pulling development dependencies from `common`.

Let me know what you think!

updater/Gemfile
```diff
@@ -27,12 +27,15 @@ gem "sentry-raven", "~> 3.1"
...
-  gem "rspec", "~> 3.12"
-  gem "rubocop", "~> 1.37.1"
-  gem "rubocop-performance", "~> 1.15.0"
-  gem "vcr", "~> 6.1"
-  gem "webmock", "~> 3.18"
+  common_gemspec = File.expand_path("../common/dependabot-common.gemspec", __dir__)
+
+  Dir.chdir(File.dirname(common_gemspec)) do
+    Gem::Specification.load(common_gemspec).development_dependencies.each do |dep|
+      next unless %(debug rspec rubocop rubocop-performance vcr webmock).include?(dep.name)
```
jeffwidman, line 34: it's really easy to overlook that this is _the_ list of gems pulled in here.

A stupid simple change that would make it visually stand out in the source is to make it a multi-line list:
```
debug
rspec
rubocop
rubocop-performance
vcr
webmock
```

dunno if they need commas, but I'm sure you'll know the answer to that. 😄

jeffwidman (approved):
This way of doing it makes sense to me, totally fine with the hack, since these two libs/apps are effectively joined at the hip into a monorepo.

updater/Gemfile
deivid-rodriguez: Thanks, updated!

updater/Gemfile
```diff
@@ -27,12 +27,22 @@ gem "sentry-raven", "~> 3.1"
...
+
+  Dir.chdir(File.dirname(common_gemspec)) do
+    Gem::Specification.load(common_gemspec).development_dependencies.each do |dep|
+      next unless %(
+        debug
+        rspec
+        rubocop
+        rubocop-performance
+        vcr
+        webmock
```
Nishnha, line 40: maybe we should name this list something more descriptive?

Nishnha (approved):
I think this is an okay approach. I agree with @jeffwidman that this is effectively a monorepo of core and updater anyway, so it makes sense to share dev dependencies.

I had one minor nit but it's not blocking

updater/Gemfile
deivid-rodriguez: Updated, what do you think?

Nishnha: Thanks that looks good to me

deivid-rodriguez:
It worked! 🥳

Issue #3844, opened 2021-06-04
Consider creating the PR even if pulling the metadata fails


yeikel:
While the additional metadata that Dependabot brings with the PRs are nice to see and add a lot of value, generally  what we truly care about is the update itself and not necessarily the metadata associated with it. 

https://github.com/dependabot/dependabot-core/blob/9e6d16c1bc7375b5cf13966908249c555c0e271d/common/lib/dependabot/pull_request_creator/message_builder.rb#L56

https://github.com/dependabot/dependabot-core/blob/9e6d16c1bc7375b5cf13966908249c555c0e271d/common/lib/dependabot/pull_request_creator/message_builder.rb#L289

I believe that we should have an option to make the metadata details optional rather than to fail the PR entirely when it fails to pull it.

jeffwidman:
Do you have any logs of a Dependabot run where it worked successfully enough that it _should_ have created the PR, but instead bombed out the metadata step failed?

I haven't stepped through the code, but my hunch is that at least _some_ of the metadata has to be fetched successfully for us to be certain it's okay to create the PR... but I agree that once it's to that point, it's probably okay to finish creating the PR even if further metadata fetching throws errors... 

But tackling this isn't really tractable without concrete examples.

yeikel:
> Do you have any logs of a Dependabot run where it worked successfully enough that it should have created the PR, but instead bombed out the metadata step failed?

I run `dependabot-script` within a Corporate environment so my logs won't be as useful. 


Essentially, the PR builder fails when it tries to fetch the metadata. In particular when it tries to fetch the commits and/or release notes as it reads`scm` configuration(for maven) and then tries to connect to github.com when the firewall is blocking external calls

In my opinion, all the metadata information are "nice to have" that should not become a blocker while trying to build the pull requests

jeffwidman:
> In my opinion, all the metadata information are "nice to have" that should not become a blocker while trying to build the pull requests
> To me, the actual upgrade is what matters and if metadata information is not available we can fetch it from other sources even if we need to do it manually.

Agreed. 

Unfortunately, given our current priorities, we're unlikely to get to this this anytime soon. 

Want to submit a PR? Should be relatively straightforward given you're already familiar with where in the code this is bombing out...

PR #5980, opened 2022-10-26
prevent failing to create a PR due to metadata gathering errors

jakecoffman:
Dependabot gathers useful information to populate the PR body and commit message. If Dependabot encounters a network error, the PR won't be posted since that extra information can't be obtained.

This PR catches those errors and populates the PR body and commit message with a minimal safe amount of info, and also logs the error so it's not a mystery as to why the message is sparse.


jeffwidman:
This will fix https://github.com/dependabot/dependabot-core/issues/3844

jakecoffman:
I have this up as a draft because it's unclear to me why the `pr_name` test is failing, it seems to be ignoring the rescue. 🤷 

I'll look into this more tomorrow, but if anyone has a tip let me know!

jakecoffman:
That was odd, there was a second exception but it was showing the first rescued exception in the output which made it look like the rescue wasn't working. Once I stepped through the code I could see what the issue was.

common/lib/dependabot/pull_request_creator/message_builder.rb
```diff
@@ -38,7 +39,12 @@ def initialize(source:, dependencies:, files:, credentials:,
...
 
       def pr_name
-        pr_name = pr_name_prefixer.pr_name_prefix
+        begin
+          pr_name = pr_name_prefixer.pr_name_prefix
+        rescue StandardError => e
+          Dependabot.logger.error("Error while generating PR name: #{e.message}")
+          pr_name = ""
+        end
         pr_name += library? ? library_pr_name : application_pr_name
```
honeyankit, line 48: Just for the understanding what is the library here?

honeyankit (approved):

common/lib/dependabot/pull_request_creator/message_builder.rb
jakecoffman: Dependabot differentiates between library and application in a few places. For NPM, if you publish to the registry it's considered a library, otherwise it's an application. I actually didn't look at how this code changes depending on the type, but in another place it decides whether to [increase the version vs widen the range](https://github.com/dependabot/dependabot-core/blob/6fdb1e09bfc8d4f61d2f66ae6eb684f7edf4baa4/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb#L110).

jeffwidman (approved):

PR #5981, opened 2022-10-26
prevent trying to get a commit that can't exist

jakecoffman:
The change in #5937 caused Dependabot to begin raising an exception inside of the rescue for repositories that had a branch that didn't exist inside of their dependabot.yml.

What is happening is:
- The clone raises a HelperSubprocessFailed
- It gets turned into RepoNotFound
- The file_fetcher_job's rescue tries to call `mark_job_as_processed`
- But first it tries to get the `base_commit_sha` which... doesn't exist! So it also raises.

This fixes two issues:
- By setting @base_commit_sha to "unknown" when we know the clone failed, we can circumvent the second raise.
- By checking the error message for "remote brach X not found" we can raise the correct error so the UI is more helpful

### DIFF
--- a/updater/lib/dependabot/file_fetcher_job.rb
+++ b/updater/lib/dependabot/file_fetcher_job.rb
@@ -57,6 +57,10 @@ module Dependabot
       return unless job.clone?
 
       file_fetcher.clone_repo_contents
+    rescue Dependabot::RepoNotFound, Dependabot::BranchNotFound
+      # base_commit_sha can't be reached, don't attempt to fetch it in the rescue
+      @base_commit_sha = "unknown"
+      raise
     end
 
     def base64_dependency_files
--- a/common/lib/dependabot/file_fetchers/base.rb
+++ b/common/lib/dependabot/file_fetchers/base.rb
@@ -85,7 +85,11 @@ module Dependabot
       def clone_repo_contents
         @clone_repo_contents ||=
           _clone_repo_contents(target_directory: repo_contents_path)
-      rescue Dependabot::SharedHelpers::HelperSubprocessFailed
+      rescue Dependabot::SharedHelpers::HelperSubprocessFailed => e
+        if e.message.include?("fatal: Remote branch #{target_branch} not found in upstream origin")
+          raise Dependabot::BranchNotFound, target_branch
+        end
+
         raise Dependabot::RepoNotFound, source
       end
 
--- a/common/spec/dependabot/file_fetchers/base_spec.rb
+++ b/common/spec/dependabot/file_fetchers/base_spec.rb
@@ -1553,6 +1553,16 @@ RSpec.describe Dependabot::FileFetchers::Base do
           expect { subject }.to raise_error(Dependabot::RepoNotFound)
         end
       end
+
+      context "when the branch can't be found" do
+        let(:branch) do
+          "notfound"
+        end
+
+        it "raises a not found error" do
+          expect { subject }.to raise_error(Dependabot::BranchNotFound)
+        end
+      end
     end
   end
 end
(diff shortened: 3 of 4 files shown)


Nishnha (approved):
Looks correct to me

Definitely an improvement over the error cascades we had earlier - reminds me of `throw null` in Java

updater/lib/dependabot/file_fetcher_job.rb
```diff
@@ -57,6 +57,10 @@ def clone_repo_contents
       return unless job.clone?
 
       file_fetcher.clone_repo_contents
+    rescue Dependabot::RepoNotFound, Dependabot::BranchNotFound
+      # base_commit_sha can't be reached, don't attempt to fetch it in the rescue
```
deivid-rodriguez, line 61: I found this comment slightly confusing because we're inside a `rescue` clause ourselves, but it's easy enough to figure out you meant the `rescue` clause in the caller (`#perform`).

Having to reference other parts of the code in comments though makes me wonder if this needs a refactoring, so I was thinking of a more explicit approach, like this:

```diff
diff --git a/updater/lib/dependabot/file_fetcher_job.rb b/updater/lib/dependabot/file_fetcher_job.rb
index 415221ee5..58cb2c948 100644
--- a/updater/lib/dependabot/file_fetcher_job.rb
+++ b/updater/lib/dependabot/file_fetcher_job.rb
@@ -8,12 +8,15 @@ require "octokit"
 module Dependabot
   class FileFetcherJob < BaseJob
     def perform_job
+      @base_commit_sha = nil
+
       begin
         connectivity_check if ENV["ENABLE_CONNECTIVITY_CHECK"] == "1"
         clone_repo_contents
-        base_commit_sha
+        @base_commit_sha = file_fetcher.commit || "unknown"
         dependency_files
       rescue StandardError => e
+        @base_commit_sha ||= "unknown"
         if Octokit::RATE_LIMITED_ERRORS.include?(e.class)
           remaining = rate_limit_error_remaining(e)
           logger_error("Repository is rate limited, attempting to retry in " \
@@ -22,14 +25,14 @@ module Dependabot
           logger_error("Error during file fetching; aborting")
         end
         handle_file_fetcher_error(e)
-        service.mark_job_as_processed(job_id, base_commit_sha)
+        service.mark_job_as_processed(job_id, @base_commit_sha)
         clear_repo_contents_path
         return
       end
 
       File.write(Environment.output_path, JSON.dump(
                                             base64_dependency_files: base64_dependency_files.map(&:to_h),
-                                            base_commit_sha: base_commit_sha
+                                            base_commit_sha: @base_commit_sha
                                           ))
 
       save_job_details
@@ -40,7 +43,7 @@ module Dependabot
 
       File.write(Environment.job_path, JSON.dump(
                                          base64_dependency_files: base64_dependency_files.map(&:to_h),
-                                         base_commit_sha: base_commit_sha,
+                                         base_commit_sha: @base_commit_sha,
                                          job: job_definition["job"]
                                        ))
     end
@@ -57,10 +60,6 @@ module Dependabot
       return unless job.clone?
 
       file_fetcher.clone_repo_contents
-    rescue Dependabot::RepoNotFound, Dependabot::BranchNotFound
-      # base_commit_sha can't be reached, don't attempt to fetch it in the rescue
-      @base_commit_sha = "unknown"
-      raise
     end
 
     def base64_dependency_files
@@ -88,15 +87,6 @@ module Dependabot
       @job ||= Job.new(attrs)
     end
 
-    def base_commit_sha
-      @base_commit_sha ||= file_fetcher.commit || "unknown"
-    rescue StandardError
-      # If an error occurs, set the commit SHA instance variable (so that we
-      # don't raise when recording the error later) and re-raise
-      @base_commit_sha = "unknown"
-      raise
-    end
-
     def file_fetcher
       return @file_fetcher if defined? @file_fetcher
 
```

Issue #5975, opened 2022-10-25
Enable fetching git submodules in cloning file fetchers


bdragon:
Scenario: Go modules project using the `replace` directive to point to a dependency that is a git submodule of the project repo:

```
$ cat .gitmodules
[submodule "my-dep"]
    path = "my-dep"
    url = "https://github.com/my-org/my-dep.git"
```

```
$ cat go.mod
module github.com/my-org/my-repo

go 1.18

require (
	github.com/my-org/my-dep v1.0.0
)

replace (
	github.com/my-org/my-dep => ./my-dep
)
```

Current behavior: Dependabot currently fails to read `my-dep/go.mod` because the file fetcher clones the repository with `--no-recurse-submodules`.

Desired behavior: cloning file fetchers should be able to specify `--recurse-submodules` when appropriate.

PR #5982, opened 2022-10-27
Allow file fetchers to opt into loading git submodules

bdragon:
## Context

Closes https://github.com/dependabot/dependabot-core/issues/5975

Repos are always git-cloned with `--no-recurse-submodules`. This can be problematic if a dependency lives in a git submodule and dependency resolution requires reading from it. For my use case, this is showing up when using go modules.

## What's Changing

This PR adds `#recurse_submodules_when_cloning?` to `Dependabot::FileFetchers::Base`. If it returns a truthy value, repos are git-cloned with [`--recurse-submodules`](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---recurse-submodulesltpathspecgt) and [`--shallow-submodules`](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---no-shallow-submodules); if it returns a falsy value, `--no-recurse-submodules` is used (the current behavior).

The default implementation of the method returns false, preserving the existing behavior for all file fetchers. Subclasses of `Dependabot::FileFetchers::Base` may override to opt into the behavior, and this PR does so for the go modules file fetcher.

The change also extends the behavior to the git-fetch and git-reset operations if `source.commit` is present (i.e., for testing). The relevant options used are [`git fetch --recurse-submodules=on-demand`](https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt---recurse-submodulesyeson-demandno) and [`git reset --recurse-submodules`](https://git-scm.com/docs/git-reset#Documentation/git-reset.txt---no-recurse-submodules).

## How to Review

I recommend reviewing this PR by commit:

0b1f0978c93a890174a68731bf220dd7d6745f15: I noticed that `Dependabot::FileFetchers::Base` declares some methods "private" (ruby `private`, prefixed with underscore) under a comment heading that says they should not be used by subclasses, but some "protected" (ruby `private`, no underscore) methods were mixed in with these under the same comment. This commit moves the protected methods above the comment, which makes the overall diff appear larger than what's actually changing.

41bac21d56c5cfd2c35adec5030ad2987442ea32: adds `#recurse_submodules_when_cloning?` and integrates it into `#_clone_repo_contents`.

 2f8a624d74de0ee214954894e5a64ac5f1ce0e05: overrides `#recurse_submodules_when_cloning?` for the `go_modules` file fetcher in order to opt into the behavior.

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_fetcher.rb
+++ b/go_modules/lib/dependabot/go_modules/file_fetcher.rb
@@ -47,6 +47,10 @@ module Dependabot
       def go_sum
         @go_sum ||= fetch_file_if_present("go.sum")
       end
+
+      def recurse_submodules_when_cloning?
+        true
+      end
     end
   end
 end
--- a/common/lib/dependabot/file_fetchers/base.rb
+++ b/common/lib/dependabot/file_fetchers/base.rb
@@ -169,6 +169,93 @@ module Dependabot
                                 end
       end
 
+      def cloned_commit
+        return if repo_contents_path.nil? || !File.directory?(File.join(repo_contents_path, ".git"))
+
+        SharedHelpers.with_git_configured(credentials: credentials) do
+          Dir.chdir(repo_contents_path) do
+            return SharedHelpers.run_shell_command("git rev-parse HEAD")&.strip
+          end
+        end
+      end
+
+      def default_branch_for_repo
+        @default_branch_for_repo ||= client_for_provider.
+                                     fetch_default_branch(repo)
+      rescue *CLIENT_NOT_FOUND_ERRORS
+        raise Dependabot::RepoNotFound, source
+      end
+
+      def update_linked_paths(repo, path, commit, github_response)
+        case github_response.type
+        when "submodule"
+          sub_source = Source.from_url(github_response.submodule_git_url)
+          return unless sub_source
+
+          @linked_paths[path] = {
+            repo: sub_source.repo,
+            provider: sub_source.provider,
+            commit: github_response.sha,
+            path: "/"
+          }
+        when "symlink"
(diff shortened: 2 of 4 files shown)


jakecoffman (approved):
Makes sense to me, great work!

thepwagner:
👋 hey friends! This broke our updates 😿 

We build a `go_modules` app that uses a git submodule for some non-code resources. This submodule is irrelevant to our updates.
The submodule repository is owned by a different GitHub organization, so providing a token with appropriate access is inconvenient (e.g. requires a Personal Access Token?)

Before this PR, dependabot worked great.
After this PR was deployed, Dependabot can no longer clone our `go_modules` repository and therefore can not give us any updates.

Is there any way for us to disable this new functionality via configuration?

go_modules/lib/dependabot/go_modules/file_fetcher.rb
```diff
@@ -47,6 +47,10 @@ def go_mod
       def go_sum
         @go_sum ||= fetch_file_if_present("go.sum")
       end
+
+      def recurse_submodules_when_cloning?
+        true
```
Nishnha, line 52: Something we can try here to mitigate the issue @thepwagner brought up is only install submodules if the initial dependency resolution step failed to run

We do something of that sorts for Terraform modules in https://github.com/dependabot/dependabot-core/blob/95df3a4d92ffc938adc1ece2942a7e9887387573/terraform/lib/dependabot/terraform/file_updater.rb#L242-L253

jakecoffman: Another option might be we leave the submodule clone by default, but ignore any errors to attempt the update anyway? That way if it didn't matter then we still do the update.

PR #5984, opened 2022-10-27
Bump Node.js in bug report description field

HonkingGoose:
## Changes:

- Bump from Node.js `14.1` to `16.18`

## Context:

`16.18` is the latest Node.16 maintenance release.

You use Node `16` in your Dockerfile, so it seems good to match that in the bug report description.

https://github.com/dependabot/dependabot-core/blob/6da4eadd68fea6c0c9f4975ea17be2e957514576/Dockerfile#L114-L121


jeffwidman (approved):
Thank you!

PR #5985, opened 2022-10-27
Centralize pyenv install logic

pavera:
## Context
As part of the work to allow Dependabot to work reliably when isolated from the internet some code cleanup has been identified.  This is the beginning of the work to limit the places where we make attempts to talk to the public internet

## Approach
I've added a new `helpers.rb` for python which will contain various methods which are currently copy/pasted between the various python package manager implementations.  The first method I've moved over is the method that installs new python versions via `pyenv install`.

## Ongoing
I am pretty sure the `python_version` and the methods it calls in the various implementations can also be moved to the new `helpers.rb` and I will be working on that next.


honeyankit (approved):
Looks good to me

pavera:
@mctofu will this log as you are expecting to allow us to search in a central place?

PR #5986, opened 2022-10-27
Sanitize file names in `Dependabot::DependencyFile` initializer

mattt:
Most package ecosystems have code to write dependency files into a temporary directory. Here's a representative example from `npm_and_yarn`:

https://github.com/dependabot/dependabot-core/blob/6a9c745252f7ff8389751b58133c851df6207b79/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/dependency_files_builder.rb#L16-L26

You can [find a few dozen instances of this pattern](https://github.com/search?q=repo%3Adependabot%2Fdependabot-core%20%22FileUtils.mkdir_p(Pathname.new(path).dirname)%22&type=code) in our codebase.

When `file.name` has a leading `/`, those calls to `FileUtils#mkdir_p` and `File.write` write to the root directory. This causes an `Errno::EACCES`. For example, when Dependabot runs an update job for a JavaScript project with a local path dependency it produces the error `"Permission denied @ rb_sysopen - /package.json"`.

This PR attempts to address this problem by sanitizing `Dependabot::DependencyFile` names in the initializer. With this change, any file names with an absolute path are translated into relative paths that will be written into temporary directories instead of attempting to be written to system root.

### DIFF
--- a/common/lib/dependabot/dependency_file.rb
+++ b/common/lib/dependabot/dependency_file.rb
@@ -21,7 +21,7 @@ module Dependabot
     def initialize(name:, content:, directory: "/", type: "file",
                    support_file: false, symlink_target: nil,
                    content_encoding: ContentEncoding::UTF_8, deleted: false, operation: Operation::UPDATE)
-      @name = name
+      @name = sanitize_name(name)
       @content = content
       @directory = clean_directory(directory)
       @symlink_target = symlink_target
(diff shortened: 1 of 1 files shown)


jakecoffman:
This has been an issue with the file_fetchers (#5866, #5869), but it manifests in the update_checker when it tries to write the files out. I'm wondering if it's causing other subtle issues, and if we sanitize here we might see more issues elsewhere that will be harder to diagnose.

Also if users are checking in absolute paths we should just be rejecting the job since we can't see some of the dependencies involved, the update could be invalid. 

So I think we need to dig down into each of these and see where we're going wrong in the file_fetcher rather than fix it up later.

jeffwidman:
Related:
* https://github.com/dependabot/dependabot-core/pull/6229

PR #5990, opened 2022-10-29
Cast source_url to string in terraform provider

andrcuns:
It seems that https://github.com/dependabot/dependabot-core/commit/b4d66fa5ec88d6456a63fd1c04504ea9b325e586 introduced a bug where `URI` object would be passed in to `get_proxied_source` method. This results in raising `undefined method 'start_with?' for #<URI::HTTPS>`.

Convert `URI` to `String` for it to work correctly

### DIFF
--- a/terraform/lib/dependabot/terraform/registry_client.rb
+++ b/terraform/lib/dependabot/terraform/registry_client.rb
@@ -103,7 +103,7 @@ module Dependabot
           return nil unless response.status == 204
 
           source_url = response.headers.fetch("X-Terraform-Get")
-          source_url = URI.join(download_url, source_url) if
+          source_url = URI.join(download_url, source_url).to_s if
             source_url.start_with?("/", "./", "../")
           source_url = RegistryClient.get_proxied_source(source_url) if source_url
         when "provider", "providers"


jeffwidman (commented):
Thank you, this looks like a straightforward fix.

However, I find it concerning that this isn't caught by the unit tests... yet when I grep https://github.com/dependabot/dependabot-core/blob/46aecc93aaf79dfe9a2c55574fbf4644d27f70ed/terraform/spec/dependabot/terraform/registry_client_spec.rb it's littered with `https` references.

So what is it about this particular scenario that isn't represented in our unit tests? What is `source_url = response.headers.fetch("X-Terraform-Get")` returning? It looks like it has to return a relative URL to hit this code path... but we've got a unit test for relative urls here:
https://github.com/dependabot/dependabot-core/blob/46aecc93aaf79dfe9a2c55574fbf4644d27f70ed/terraform/spec/dependabot/terraform/registry_client_spec.rb#L284-L290

I'd just feel a lot more comfortable if we can add a test case for the broader scenario that is missing to ensure no future regressions.

andrcuns:
@jeffwidman Totally makes sense to add a spec for that and I was thinking about it but was a bit lazy on a Saturday morning to do so 😄 

The spec You linked, only tests `service_url_for` method, so it doesn't get to the code in question where the problem happens.

The issue is basically branch coverage. 
https://github.com/dependabot/dependabot-core/blob/46aecc93aaf79dfe9a2c55574fbf4644d27f70ed/terraform/spec/dependabot/terraform/registry_client_spec.rb#L150
this spec is executing the right code, but it does not return `X-Terraform-Get` as path to trigger condition `source_url.start_with?("/", "./", "../")`.

To add a spec, I would need to find a way to record a proper vcr cassette. I ran in to this issue when using private `Gitlab` terraform registry, not sure what other cases might reproduce this.

deivid-rodriguez:
Hei @andrcuns!

Even if not a real repo, what I would do is modify the recorded VCR responses that we have already in place to simulate this case, namely, copy the existing responses by with a `X-Terraform-Get` relative to the download endpoint.

Do you think that would work?

andrcuns:
@deivid-rodriguez 

Yeah, I was thinking about just copying the existing one, did not have time to do it yet.

One note though, after the latest release, it started working again and that line of code is not triggered anymore for some reason 😅

deivid-rodriguez:
🤣 That's surprising! Research on why appreciated 😅.

jeffwidman:
gentle nudge @andrcuns any chance of figuring out why this suddenly started working?

andrcuns:
@jeffwidman Sorry, I have been a bit busy lately and haven't had time to do any investigations. Given I will be "on the road" for several upcoming months, I'll probably wont have much time to investigate it anytime soon.

I could try and add a proper spec, regardless of why it actually started working (I suspect changes in tf version or tf registry itself), the actual described bug is still there as it's fairly clear from just looking at the code.

deivid-rodriguez:
Our point is it would be useless to fix a bug that nobody is running into because the code is no longer hit. That would be the case if this was a change in the terraform registry, right? And even it it's some change in terraform itself, the bug would be gone once you upgrade.

That's why I believe it's more rewarding to spend time on investigating this and potentially deleting this now dead code, than on writing a spec.

andrcuns:
>Our point is it would be useless to fix a bug that nobody is running into because the code is no longer hit

I kind of don't agree with such approach when there is a clearly visible wrong code execution path 😄 

Basically, I won't have time to do such investigation, so feel free to close this then.

deivid-rodriguez:
> I kind of don't agree with such approach when there is a clearly visible wrong code execution path 😄

But don't you agree it would be better to delete the code if it's no longer needed? I think it would be not good that you spend time writing a test for this if we will delete both the test and the code some time later once we verify this is no longer needed in the versions of terraform that we support.

> Basically, I won't have time to do such investigation, so feel free to close this then.

It's alright, let's leave it open for a bit, hopefully we can look into it ourself!

Thanks for all the help in this and other PRs, by the way ❤️

andrcuns:
>But don't you agree it would be better to delete the code if it's no longer needed?

100% agree on that, but I don't know terraform well enough and can't really determine if it's safe to delete this code even if usecase I encountered this problem personally, is not valid anymore.

PR #5991, opened 2022-10-30
Bump elixir and erlang so they don't conflict

jeffwidman:
The `1.13.x` series of Elixir wasn't precompiled to work with Erlang OTP 25.

The `1.14.x` series changed from a single pre-compiled binary to releasing separate versions compatible with OTP 23, 24, and 25.

So bump to Erlang OTP 25, and also to the Elixir version that's compiled with OTP 25.

Related:
* #5503 
* #5507 

Also, we may need to switch away from using the Erlang Solutions PPA to install Erlang via the default PPA, but I'm hopeful we won't need to do this...
* https://github.com/dependabot/dependabot-core/pull/5030#issuecomment-1296379546
* https://github.com/esl/packages/issues/15#issuecomment-1296378191
* https://github.com/elixir-lang/elixir-lang.github.com/pull/1637
* https://packages.ubuntu.com/jammy/erlang

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -211,17 +211,18 @@ RUN cd /tmp \
 # Install Erlang, Elixir and Hex
 ENV PATH="$PATH:/usr/local/elixir/bin"
 # https://github.com/elixir-lang/elixir/releases
-ARG ELIXIR_VERSION=v1.13.4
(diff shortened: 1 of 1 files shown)


jeffwidman:
I'm going to hold off on this because due to https://github.com/dependabot/dependabot-core/pull/5030/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R216-R223 we may have to hold back to Erlang OTP 24... and I extracted the Elixir bump separately into #6004.

PR #6006, opened 2022-10-31
fix hex native helpers: v2.0.0 removed Hex.Version.InvalidRequirementError

jakecoffman:
Our native helpers reference Hex.Version.InvalidRequirementError which comes from Hex itself, and since 2.0.0 was released it seems that struct is gone causing a compiler error. So I've pinned Dependabot to 1.0.1 until we can fix the issue and upgrade.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -315,3 +315,6 @@ RUN curl -sL $SHIM -o git-shim.tar.gz && mkdir -p ~/bin && tar -xvf git-shim.tar
 ENV PATH="$HOME/bin:$PATH"
 # Configure cargo to use git CLI so the above takes effect
 RUN mkdir -p ~/.cargo && printf "[net]\ngit-fetch-with-cli = true\n" >> ~/.cargo/config.toml
+
+# Pin to an earlier version of Hex. This must be run as dependabot
+RUN mix hex.install 1.0.1


honeyankit (approved):

Dockerfile
```diff
@@ -315,3 +315,6 @@ RUN curl -sL $SHIM -o git-shim.tar.gz && mkdir -p ~/bin && tar -xvf git-shim.tar
 ENV PATH="$HOME/bin:$PATH"
 # Configure cargo to use git CLI so the above takes effect
 RUN mkdir -p ~/.cargo && printf "[net]\ngit-fetch-with-cli = true\n" >> ~/.cargo/config.toml
+
+# Pin to an earlier version of Hex. This must be run as dependabot
+RUN mix hex.install 1.0.1
```
jeffwidman, line 320: Is there a reason you pinned it here and not up next to `RUN bash /opt/hex/helpers/build` or even more intuitively next to where erlang/elixir are installed? I assume the latter was a no-go because needing to run as `dependabot`?

Also, why does it have to run as `dependabot`?

jeffwidman: SG, I added moving the pin to a more intuitive location as part of https://github.com/dependabot/dependabot-core/issues/6009.

PR #6010, opened 2022-11-01
Downgrade RubyGems to 3.3.7 to match Ruby 3.1.2

jeffwidman:
We use `dependabot-core` as a library in a few internal projects. After the bump to `3.3.22`, they started complaining with the following error:
```
dependabot-common-0.213.0 requires rubygems version >= 3.3.22, which is
incompatible with the current version, 3.3.7
```

RubyGems `3.3.7` is the version shipped with `ruby:3.1.2`.

While we could patch all those places to update their system RubyGems, it's a bit of a game of whack-a-mole, and adds minimal value.

So instead, relax the constraint slightly so that those versions work. I don't mind forcing folks to upgrade to the newest Ruby, but as long as they do that, then there's no need for us to be forcing them to futz around further.

Downgrading the version used in the Dockerfile was optional, but I thought better to keep that in sync with what most consumers of the library will be using.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -70,8 +70,8 @@ RUN if ! getent group "$USER_GID"; then groupadd --gid "$USER_GID" dependabot ; 
 
 ARG RUBY_VERSION=3.1.2
 ARG RUBY_INSTALL_VERSION=0.8.3
-
-ARG RUBYGEMS_SYSTEM_VERSION=3.3.22
+# Generally simplest to pin RUBYGEMS_SYSTEM_VERSION to the version that default ships with RUBY_VERSION.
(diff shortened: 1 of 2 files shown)


jeffwidman:
Once this lands, we'll need to cut a new version here in `dependabot-core`...

pavera:
> Once this lands, we'll need to cut a new version here in `dependabot-core`...

We'll also need to revert some of the changes we started making to account for this error.

pavera (approved):

jeffwidman:
Agreed. I'll cut a release later today. And my apologies for the hassle here.

PR #6016, opened 2022-11-01
Removes `YarnLockfileParser`

bdragon:
`YarnLockfileParser` is a vestige from an era in which it may have been safer to parse a yarn.lock ourselves rather than shelling out to yarn. Elsewhere Dependabot uses a native helper, `yarn:parseLockfile` (which in turn uses Yarn's own [lockfile parser](https://github.com/yarnpkg/yarn/blob/v1.22.19/src/lockfile/parse.js)) to parse yarn lockfiles. Since `YarnLockfileParser` is no longer needed, and in order to parse lockfiles consistently, this PR removes it.


mctofu (approved):
✂️

PR #6025, opened 2022-11-03
Implement replaces-base in docker ecosystem

honeyankit:
## Context

With this PR I've implemented the "replaces-base" functionality which is already used in the Python and Maven ecosystem. This means that registry with replaces-base as true will be used as mirror. More details are [here](https://docs.docker.com/registry/recipes/mirror/)

## Solution

In file_parser.rb and in update_checker.rb, during the call of `docker_registry_client`, earlier if the registry was nil then default docker hub registry (`https://registry.hub.docker.com`) url was set. With my changes, if registry is nil, then I also check if replaces_base is set to true via function `credentials_finder.replaces_base?`. If it is true, then I fetch the registry details from the credentials instead of returning the default docker hub registry (`https://registry.hub.docker.com`) url.


docker/lib/dependabot/docker/file_parser.rb
```diff
@@ -124,15 +128,23 @@ def docker_repo_name(image, registry)
         "library/#{image}"
```
honeyankit, line 128: @jakecoffman I needed your suggestion here, for the ecr container registry (AWS), if the replaces-base is set to true and Dockerfile contains the below, then the function `docker_repo_name` will return the repo name as  "library/docker-php". I guess adding "library" to the repo name is a docker hub thing. Should I need to add a check here to check if the registry passed with replaces-base == true is ecr or not and if replaces-base == true and the registry is ecr then simply return `docker-php` instead of appending `library` to it?
 
```
FROM docker-php@sha256:18305429afa14ea462f810146ba44d4363ae76e4c8dfc38288cf73aa07485005
```

jakecoffman: Yeah that makes sense, check the URL if the host contains `.ecr.` I guess? That makes sense I think!

docker/lib/dependabot/docker/file_parser.rb
```diff
@@ -103,6 +103,13 @@ def version_from_digest(registry:, image:, digest:)
 
         repo = docker_repo_name(image, registry)
         client = docker_registry_client(registry)
+        # Here, the client's registry_hostname is collected for use in passing to
+        # the RestClient::Exceptions exception for display. When the FROM statement in
+        # the Dockerfile does not any contain registry information but replaces-bases is set
+        # to true and contains the registry information, this is required because in such cases
+        # the registry extarcted from the Dockerfile will be nil which is getting passed
+        # to docker_registry_client method as well as to the RestClient::Exceptions exception
+        registry_hostname = client.instance_variable_get(:@base_uri).sub(/^https?\:\/\/(www.)?/, "")
```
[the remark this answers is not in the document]
honeyankit: Good call. I will make the necessary changes.

docker/lib/dependabot/docker/file_parser.rb
honeyankit: I thought more on this and I think this is also applicable for all private registries passed with replaces-bases. Only when registry is default `https://registry.hub.docker.com` then we should append `library/` to the repo?

jakecoffman:
I put up a suggestion on how to simplify the code a bit [in this branch](https://github.com/dependabot/dependabot-core/compare/honeyankit/implement-replace-base-in-docker-ecosystem...jakecoffman/suggestion?expand=1).

jakecoffman (approved):

Issue #6009, opened 2022-11-01
Bump Hex package manager to `2.0`


jeffwidman:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code improvement description

We had to hold back the Hex package manager to `1.0.1` in https://github.com/dependabot/dependabot-core/pull/6006.

At some point we need to update `Hex` to `2.0.0` and fix that compiler error.

tiagoefmoraes:
You can use `Version.InvalidRequirementError` directly from [elixir](https://hexdocs.pm/elixir/Version.InvalidRequirementError.html) instead of `Hex.Version.InvalidRequirementError`

sorentwo:
@jeffwidman Seconding the comment from @tiagoefmoraes, Hex doesn't have a public API, and you should use the official [InvalidRequirementError](https://hexdocs.pm/elixir/Version.InvalidRequirementError.html).

jeffwidman:
Thanks!

Do either of you want to submit a PR? Since it's just the location of a type, should be very straightforward... the bumping Hex version and changing the pinned location can be done as a separate PR.

No worries if not interested.

jeffwidman:
Related fixes:
* https://github.com/dependabot/dependabot-core/pull/6028
* https://github.com/dependabot/dependabot-core/pull/6037
* https://github.com/dependabot/dependabot-core/pull/6026

isaacsanders:
I think I have all the current tests passing: https://github.com/dependabot/dependabot-core/pull/7890

isaacsanders:
Got the last few stragglers

Issue #2757, opened 2020-11-13
Propagate mixfile errors to the user better.


baseballlover723:
Currently if there is an error that causes dependabot to be unable to run the mixfile (such as using a file) while trying to update a dependency, then dependabot will swallow the error, making it very difficult to debug why the error is happening. I encountered this when I hadn't yet sanitized all of the supporting files (#2748), and it silently failed while trying to update the lockfiles. This particular error is fixed, but this could happen in the future with perhaps a more complex mixfile setup, or if future updates are not applied to all the places they need to.

The core cause for me was that the error was being swallowed [here](https://github.com/dependabot/dependabot-core/blob/main/hex/lib/dependabot/hex/update_checker/version_resolver.rb#L93-L97) and then nil being passed all the way back up and is eventually treated as not being upgradeable instead of being an error that the mixfile couldn't be executed. On top of that, the "error" return by the mix helpers, was just the empty string, so theres something not being handled correctly for propagating errors that occur in the hex helpers.

This probably requires a revamping of how errors are handled in the helpers in general, so I'll leave it to you all to decide if this is worth fixing or not.

jeffwidman:
👋 Hey there! Sorry for the glacially slow response... we've been a bit buried the past few years, and only just now starting to dig through some of our backlog.

This issue is pretty old, and we've pushed a lot of bugfixes since then... is this still a valid issue?

baseballlover723:
Hi, I don't work with Elixir anymore, so I'm not sure if its still an issue. I would presume so, if there haven't been major modifications to the hex part since I made this issue. Since the crux of it was that it would swallow the error message, and give back a generic one.

PR #6026, opened 2022-11-03
Bump Hex to `2.0.1`

jeffwidman:
Bump Hex to `2.0.1`.

Fix https://github.com/dependabot/dependabot-core/issues/6009

### DIFF
--- a/hex/Dockerfile
+++ b/hex/Dockerfile
@@ -25,7 +25,7 @@ USER dependabot
 COPY --chown=dependabot:dependabot hex/helpers /opt/hex/helpers
 ENV MIX_HOME="/opt/hex/mix"
 # https://github.com/hexpm/hex/releases
-ENV HEX_VERSION="1.0.1"
+ENV HEX_VERSION="2.0.1"
 RUN bash /opt/hex/helpers/build
 
 COPY --chown=dependabot:dependabot hex /home/dependabot/hex


hex/helpers/lib/check_update.exs
```diff
@@ -74,7 +74,7 @@ case UpdateChecker.run(dependency_name, credentials) do
     version = :erlang.term_to_binary({:ok, version})
     IO.write(:stdio, version)
 
-  {:error, %Version.InvalidRequirementError{} = error}  ->
+  {:error, %Hex.Version.InvalidRequirementError{} = error}  ->
```
jeffwidman, line 77: This will fail on Hex 2.0, which will prove this install method is working as expected. I'll remove it before final merge.

hex/helpers/build
```diff
@@ -10,7 +10,14 @@ fi
 install_dir="$DEPENDABOT_NATIVE_HELPERS_PATH/hex"
 mkdir -p "$install_dir"
 
-mix local.hex --force
```
jeffwidman, line 13: This is the second time we're calling `mix local.hex --force` on `main`, so need to either remove it from here or remove it from the Dockerfile... that's why I opted to centralize the logic in here.

hex/helpers/build
```diff
@@ -10,7 +10,14 @@ fi
...
+# Initial Hex install
+mix local.hex --force --if-missing
```
jeffwidman, line 14: `--if-missing` skips installing if already installed... ie, if someone re-running the build script locally for some reason.

hex/helpers/build
```diff
@@ -10,7 +10,14 @@ fi
...
+# Annoyingly, the Hex version cannot be specified during the initial install.
+# It must be downgraded later.
+if [ -n "$HEX_VERSION" ]; then
+  mix hex.install "$HEX_VERSION"
```
jeffwidman, line 18: This will fail if `mix local.hex` isn't called beforehand... and that command updates to the latest Hex... so if it's broken out of the gate, we will be hosed. 😢 

That's why we want the pinned install to happen immediately after the `mix local.hex` install.

Dockerfile
```diff
@@ -320,6 +321,3 @@ RUN mkdir -p ~/.cargo && printf "[net]\ngit-fetch-with-cli = true\n" >> ~/.cargo
 # Disable automatic pulling of files stored with Git LFS
 # This avoids downloading large files not necessary for the dependabot scripts
 ENV GIT_LFS_SKIP_SMUDGE=1
-
-# Pin to an earlier version of Hex. This must be run as dependabot
```
jeffwidman, line 324: > This must be run as dependabot

This is the other reason I opted to install hex within the helper builder script and not up when we install Erlang / Elixir, as they're installed as the root user.

jakecoffman (approved):

jeffwidman:
I spent six hours on this yesterday/this morning before I finally understood why the tests are failing... and now I know a _lot_ more about Elixir than I used to. 😁 

Here's the flow of this code for this failing unit test: https://github.com/dependabot/dependabot-core/blob/dd2c5dba914782e3acbd2b9f052af1b973347c88/hex/spec/dependabot/hex/update_checker_spec.rb#L711

1. Within the Elixir native helper, `run.exs` spawns a subprocess here to call `check_update.exs`:
https://github.com/dependabot/dependabot-core/blob/dd2c5dba914782e3acbd2b9f052af1b973347c88/hex/helpers/lib/run.exs#L50
2. `check_update.exs` calls into Hex to handle the dep resolution here: 
https://github.com/dependabot/dependabot-core/blob/dd2c5dba914782e3acbd2b9f052af1b973347c88/hex/helpers/lib/check_update.exs#L37
3. And then returns the output from Hex here back to `stdio`:
https://github.com/dependabot/dependabot-core/blob/dd2c5dba914782e3acbd2b9f052af1b973347c88/hex/helpers/lib/check_update.exs#L58-L59
4. The parent `run.exs` then reads that output and attempts to translate it from binary back to Erlang native terms:
https://github.com/dependabot/dependabot-core/blob/dd2c5dba914782e3acbd2b9f052af1b973347c88/hex/helpers/lib/run.exs#L11
5. Which then eventually passes through `handle_result` > `encode_and_write` to be JSON encoded:
https://github.com/dependabot/dependabot-core/blob/dd2c5dba914782e3acbd2b9f052af1b973347c88/hex/helpers/lib/run.exs#L36
6. The JSON result is then parsed by our ruby calling code and more stuff happens... but the problem is within the above.

Both Hex 1 and Hex 2 result in the same error:
```
{:error, %Mix.Error{message: "Hex dependency resolution failed", mix: 1}}
```

However, Hex 2 also directly logs the reason for the resolution failure to stdio. This is also mentioned in the [Hex 2.0 announcement blog post](https://hex.pm/blog/hex-v2.0-released-with-new-version-solver):
> Recording the conflict causes has the added benefit that we can use them to create human readable error messages that step-by-step explain why solving failed if no solution can be found.

This is an awesome feature as seen in the screenshot in the linked blog post above.

Unfortunately, those human readable reasons get included between steps 2 and 3 above... so the output that step 4 is trying
to parse is not only the machine-readable error, but also the human-readable error messages... so it blows up because
those human readable messages don't map back to native Erlang terms.

There's two approaches to solve this:
1. Disable the human readable error message output by passing a config to Hex. I don't know if Hex 2 currently has an API to do this.
2. Change the message passing architecture of the native helper so that the returned value never has to translate to/from stdio/binary.
   As I investigated this native helper, I was surprised that it internally spawns a subprocess... why not do all the work within the same process? There may be a valid reason that has to happen, or it may be legacy. If it's legacy and we can get rid of it, that makes for a much cleaner solution because we can still expose the human-readable error messages, while handling the error appropriately and without the overhead of a subprocess.

To reproduce the failure for debugging:
1. Within the docker dev shell built from this branch (since we're using a native helper, we can't use the pre-built image) by calling: 
  ```shell
  bin/docker-dev-shell
  # wait forever while docker image is built
  # drops into the container
  ```
2. Ran the following unit test:
```shell
$ cd hex && DEBUG_HELPERS=true DEBUG_FUNCTION="get_latest_resolvable_version" rspec ./spec/dependabot/hex/update_checker_spec.rb:711
```
This effectively runs the unit test until it shells out to the native helper, then prints the command to run and pauses so that I can `docker exec` into the container and re-run the native helper call directly against the temp dir fixture created as part of the test...  It printed out:
```shell
cd /home/dependabot/dependabot-core/hex/tmp/dependabot_20221203-12922-hmumqj && echo "{\"function\":\"get_latest_resolvable_version\",\"args\":[\"/home/dependabot/dependabot-core/hex/tmp/dependabot_20221203-12922-hmumqj\",\"phoenix\",[]]}" | MIX_EXS=/opt/hex/mix.exs MIX_LOCK=/opt/hex/mix.lock MIX_DEPS=/opt/hex/deps MIX_QUIET=1 mix run /opt/hex/lib/run.exs
```
Leave this running because if you exit the paused state it will cleanup the temp dir fixture.

3. In another terminal I did `docker exec -it <container sha> bash`, followed by running the above command.

4. At this point you can play with the native helper code... just be sure to rebuild it if you edit it using a text editor
   outside the container. Also note that the subprocess script `check_update.exs` is copied to the test's temp dir, and
   so it's rebuilt when using the normal helper rebuild. I found it fastest to directly edit it with `nvim`. And also
   to run it directly when I was debugging... otherwise since it's run in a subprocess, its stdio wasn't exposed.

jeffwidman:
Another thing to double-check is whether the `mix hex.*` CLI interface might now be advanced enough that we could use it directly w/o needing a custom native helper script... ala what we did in `go` ecosystem.

See for example:
* https://blog.appsignal.com/2022/03/15/a-guide-to-secure-elixir-package-updates.html
* https://tosbourn.com/keeping-elixir-packages-updated/
* https://elixirforum.com/t/finding-out-hex-package-version-numbers/51654/5

jeffwidman:
Looks like `Mix.Shell.Process` may provide a way to do IPC w/o having to capture all of stio:
* https://elixirforum.com/t/get-mix-task-output-so-far-for-tests/44518/4
* https://hexdocs.pm/mix/Mix.Shell.Process.html#content

But I'd still rather simply drop the subprocess if possible and run the subtask within the task:
* https://stackoverflow.com/questions/41656482/how-do-i-run-a-mix-task-from-within-a-mix-task
* https://stackoverflow.com/questions/36995398/how-to-run-multiple-mix-tasks-in-one-command

Or even get away from using the private Hex API's entirely if we can get the Hex CLI API to give us the info we need:
* https://hexdocs.pm/hex/Mix.Tasks.Hex.Outdated.html

jeffwidman:
Related since Hex 2 changes result in more helpful errors that we may wish to display to users:
* https://github.com/dependabot/dependabot-core/issues/2757

jeffwidman:
If anyone from the community is interested in picking this up, please feel free.

isaacsanders:
Where are we at with it?

jeffwidman:
As mentioned above, this most likely needs someone from the community to step forward and carry it across the finish line. 

Only a small fraction of our users run on this ecosystem... so from a user impact perspective it's tough for the core team to justify investing time into this compared to our more popular ecosystems.

Additionally, as I'm not an Erlang / Hex / Mix expert it'll take a bit more work to figure out how to workaround the problem mentioned here: https://github.com/dependabot/dependabot-core/pull/6026#issuecomment-1336276793

Ideally we do something like this: https://github.com/dependabot/dependabot-core/pull/6026#issuecomment-1336473731

But again, it'll need someone who has time/interest in digging deeper into this ecosystem to carry it forward. 

If I get time, I could see myself doing it for fun, but I can't justify this from a business or user-impact perspective.

isaacsanders:
Where is the PR for those changes referred to in that code? All I see in this PR is the change to the ENV. I’d be willing to provide feedback to someone who already has all the dev tooling set up, but it’s a big task to use all the dev tools.

jeffwidman:
> I’d be willing to provide feedback to someone who already has all the dev tooling set up, but it’s a big task to use all the dev tools.

Totally up to you. Like any other open source project, so until someone from the community steps up, this is unlikely to move forward.

The readme has pretty good step-by-step instructions if you're interested in learning what it'd take to get a dev environment going to pursue this further: https://github.com/dependabot/dependabot-core#development-guide

isaacsanders:
I've fixed the tests and updated to a later patch version: #7890

jeffwidman:
Superseded by #7890.

There's some great notes in here about ways to improve the format of the underlying IO passing between the native helper and Dependabot. However, I've moved off the official Dependabot team and also putting my open source free time elsewhere right now, so I don't foresee myself getting back to this.

Issue #5926, opened 2022-10-19
Unrelated libraries are updated with bundler


t-tiger:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

Bundler

### Package manager version

bundler `2.2.32`

### Language version

Ruby `2.7.2`

### Manifest location and content before the Dependabot update

/Gemfile

### dependabot.yml content

```yaml
- package-ecosystem: bundler
  directory: "/"
  schedule:
    interval: daily
  open-pull-requests-limit: 2
  target-branch: develop
```

### Updated dependency

When updating [warning](https://github.com/jeremyevans/ruby-warning) from 1.2.1 to 1.3.0, the followings are also updated while only warning should be updated.

- [apollo-gateway](https://github.com/Gusto/apollo-federation-ruby) from 3.0.0 to 3.3.0
- [ddtrace](https://github.com/DataDog/dd-trace-rb) from 0.54.2 to 1.1.0.
- [google-protobuf](https://rubygems.org/gems/google-protobuf) from 3.20.0 to 3.21.7
- [haml-rails](https://github.com/haml/haml-rails) from 2.0.1 to 2.1.0
- [listen](https://github.com/guard/listen) from 3.2.1 to 3.7.1
- [nokogiri](https://github.com/sparklemotion/nokogiri/releases) from 1.36.6 to 1.36.8

Above libraries are not related to the original gem, they should be updated with other PRs.

### What you expected to see, versus what you actually saw

**Expected**
Only [warning](https://github.com/jeremyevans/ruby-warning) is updated from 1.2.1 to 1.3.0, which is the result when I manually execute `bundle update warning`.

**Actual**
Other libraries are also updated as `Updated dependency` section shows.

It seems to start last week. Our team has been using dependabot for a long time.

deivid-rodriguez:
Thanks for reporting.

I guess related to the upgrade of Bundler at https://github.com/dependabot/dependabot-core/pull/5509.

Can you share a public repository that reproduces this?

t-tiger:
Thank you for quick response.

> Can you share a public repository that reproduces this?

Sorry, this case is inside private repository at my work 🙇

t-tiger:
I made a public repository that extracts a subset of libraries. As a result, dependabot works as expected 🤔 
https://github.com/t-tiger/dependabot-test/pull/1

After I created a PR, I found other repositories in the company don't have the same problem. 
[...]

PR #6030, opened 2022-11-03
Fix accidental downgrades in Bundler PRs

deivid-rodriguez:
Dependabot monkeypatches Bundler to allow resolving for arbitrary rubies, not just for the running Ruby. However, when no Ruby information is provided via either `gemspec` files or `ruby` Gemfile directive, Dependabot was only allowing to resolve either to Ruby 2.5, or to the running Ruby 3.1.

In some situations, it can happen that resolving on Ruby 3.1 is not possible, while resolving on Ruby 2.5 is possible but requires downgrading some lock file dependencies.

To avoid this problem, let's add the missing rubies to our list of rubies Dependabot can resolve to.

NOTE: I spent some time trying to write specs for this, but it turned out pretty tricky. In the end, I added a note in the Dockerfile to avoid falling out of sync again here. Also, I will look into porting this functionality to Bundler so that Dependabot does not need to monkey patch it, and thus leads to more consistent results.

Fixes https://github.com/dependabot/dependabot-core/issues/5926.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,6 +69,7 @@ RUN if ! getent group "$USER_GID"; then groupadd --gid "$USER_GID" dependabot ; 
 
 ### RUBY
 
+# When bumping Ruby minor, need to also add the previos version to `bundler/helpers/v{1,2}/monkey_patches/definition_ruby_version_patch.rb`
 ARG RUBY_VERSION=3.1.2
 ARG RUBY_INSTALL_VERSION=0.8.3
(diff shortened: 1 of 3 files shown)


jurre (approved):

deivid-rodriguez:
Thanks for having a look @jurre!

etiennebarrie:
> Also, I will look into porting this functionality to Bundler so that Dependabot does not need to monkey patch it, and thus leads to more consistent results.

Is there any issue I can track about this feature? I'd love Bundler to be able to bundler for other Ruby versions than the one it's currently running in.

deivid-rodriguez:
No, sorry, I never opened an upstream issue about it!

Issue #1303, opened 2019-08-09
Support for NuGet package lock files


xt0rted:
In NuGet 4.9 I think it is they added support for package lock files using the setting `RestorePackagesWithLockFile` which is part of your `.csproj` or `.props` file chain. This creates a file called `packages.lock.json` alongside your project file. The spec for this can be found at https://github.com/NuGet/Home/wiki/Enable-repeatable-package-restore-using-lock-file

greysteil:
Super keen to do this, but it's going to require some work - at the moment Dependabot does all of its .NET updating in Ruby, and we'll need to change that. We might want to continue this over at https://github.com/dependabot/dependabot-core/issues/541?

gitfool:
@jerriep. given your great work with [dotnet-outdated](https://github.com/jerriep/dotnet-outdated), do you have any advice on how Dependabot can also update any associated [dependency lock](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#locking-dependencies) files in its PR?

Is the best way to run [`dotnet restore`](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-restore), perhaps with [`--force-evaluate`](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#lock-file-extensibility), or are there other better ways?

jerriep:
@gitfool No, I was not even aware of this new feature until you just pointed it out. You may have to resort to using some of the [NuGet Client libraries](https://github.com/NuGet/NuGet.Client) to achieve this.

Maybe look at [this PR](https://github.com/NuGet/NuGet.Client/pull/2342) and the linked specifications to see whether there are some hints to guide you.

vmcbaptista:
is there any news regarding this topic?

edumserrano:
Would also be interested in support for this feature. Any updates?

majastrz:
Would also be interested in support for this.

flcdrg:
I've written up a workaround for this at https://david.gardiner.net.au/2021/05/dependabot-nuget-lockfiles.html, but would much prefer it was handled natively given that dependabot stops maintaining the PR once a 'foreign' commit has been added.

nogic1008:
As a workaround, I define a GitHub action in my repository that kicks `dotnet restore`.
[...]

PR #5902, opened 2022-10-17
Fix `lockfile-only` versioning strategy not creating some updates that are expected (v2)

deivid-rodriguez:
Trying to retry #5581, but in a correct way now.

Fixes #5569.

PR #6031, opened 2022-11-03
Support NuGet lockfiles

anthony-c-martin:
First draft at addressing #1303

I'm super unfamiliar with this project as well as Ruby, so guidance is very welcome.


ldeluigi:
Any news on this?

deivid-rodriguez:
Hei @ldeluigi, you just got an update yesterday at #1303. Let's try to not be too pushy please, we get a lot of notifications around here 😅.

ldeluigi:
> Hei @ldeluigi, you just got an update yesterday at #1303. Let's try to not be too pushy please, we get a lot of notifications around here 😅.

whops: I'm sorry!

jeffwidman (commented):
👋 Sorry for the delay, I've been wanting to get to this and finally had a little time this evening.

A few thoughts:
1. First, a big caveat, I've never worked with Dotnet/Nuget myself, so any comments may simply be due to misunderstanding.
2. This looks really solid from a general "dependabot" structure perspective, nice job! 🎉 
3. We split out the dockerfiles by ecosystem, so you'll need to rebase to install `dotnet` into https://github.com/dependabot/dependabot-core/blob/d0bd472495eda2eabb71040a59dbbfeb4dfb4931/nuget/Dockerfile
4. I _really_ like that you're leveraging the native tooling... we want to migrate this direction as much as we can.
5. Today we do all our dotnet updating via ruby/regex text parsing hackery... now that we'd have the native tooling available, does it also support updating the manifest itself? Such that we could start to leverage that to migrate away from our existing text hackery? Obviously something we'd handle in a different PR, but just want to know if we should be creating a new ticket to track opportunities for migration.
7. CI looks unhappy, mind fixing?

If you can rebase and fix that last CI failure, then I'm 👍 on merging.

I'll also ask a teammate who's more familiar with Nuget to take a look just in case I'm overlooking something, but this looks pretty straightforward so I don't expect any issues. Even if we don't cover _all_ use cases, even adding support for some of them would be a great start for others to build out further.

Thanks again! And sorry took so long to review.

Dockerfile
```diff
@@ -289,6 +289,13 @@ RUN DART_ARCH=${TARGETARCH} \
   && rm "/tmp/${DART_EXE}" \
   && dart --version
 
+### DOTNET
+ENV DOTNET_NOLOGO=true \
+  DOTNET_CLI_TELEMETRY_OPTOUT=true
+
+RUN curl -sSL https://dot.net/v1/dotnet-install.sh | bash /dev/stdin --channel 6.0 --install-dir /usr/share/dotnet \
```
jeffwidman, line 296: I see both `6.0` and `7.0` listed on https://dotnet.microsoft.com/en-us/download

I'm totally ignorant of dotnet / nuget versioning expectations.

Did anything change from a manifest / lockfile perspective between them? Is `7.0` backwards compatible with `6.0` files? If so, should we use that? Or is it better to stick with `6.0`  since it's LTS?

jeffwidman: Also, is it typical to only pin to the latest minor version, or should we pin to the full patch release? Not sure how frequent/stable the patch releases tend to be...

nuget/spec/fixtures/csproj/lockfiles_basic
jeffwidman, line 1: I'm curious: why does this fixture not have a file extension? No other fixture in this folder is extension-less.

I was also surprised that it's plural "lockfiles" when the content appears to be only a single lockfile?

jeffwidman:
Nudge @anthony-c-martin any chance of addressing my questions?☝️

Be great to move this forward!

Dockerfile
lizalc: .NET versions are backwards compatible, so yes, `7.0` works with `6.0`. For updating packages and lockfiles the only universal option is to use the latest preview SDK version as older SDK versions cannot work with projects that specify newer SDK versions. Thus, this PR should not hardcode the `6.0` version as it will break when working with projects using `7.0+`. Granted I'm not super familiar with how dependabot manages toolchain versions.

Patch releases are released once a month. Most projects I've seen only pin to the latest minor version.

Relevant docs:

- [Overview of how .NET is versioned](https://learn.microsoft.com/en-us/dotnet/core/versions/)
- [Breaking changes in .NET](https://learn.microsoft.com/en-us/dotnet/core/compatibility/breaking-changes)

jeffwidman:
Note to self: Since this adds lockfile support, we'll need to ensure it supports the `lockfile-only` strategy, see how other ecosystems are wired up here:
* https://github.com/dependabot/dependabot-core/pull/5902

Dockerfile
jeffwidman: Thank you, that's very helpful.

Typically for most ecosystems we pin the toolchain and then manually bump it as needed (ironic, I know 😁, we'll get :dependabot: there over time).

Pinning to minor rather than patch is probably fine, we can always get more specific if needed.

So we'd hardcode to `7.0` at the start, and bump to `7.1` or `8.0` once those get released.

lizalc: Maybe outside of the scope of this PR, but could dependabot read and use the highest .NET version specified in the project configuration? I believe Mend Renovate does something like this.

This would ensure dependabot uses the version used by the developer and avoids potential breaking changes due to using a newer version. It would also allow dependabot to work with preview SDK versions which at least I'd find useful.

jeffwidman: Thanks for the feedback, but definitely outside the scope of this PR. 

If an issue already exists for these, it's probably tagged with the `L: dotnet:nuget` label: https://github.com/dependabot/dependabot-core/issues?q=is%3Aopen+is%3Aissue+label%3A%22L%3A+dotnet%3Anuget%22

Otherwise please file new issues.

Porges:
I'm not entirely sure you want `--force-evaluate` since that could affect packages other than the one that :dependabot: is trying to update? At least for me it is sufficient to run the basic `dotnet restore` command to update the lockfile after :dependabot: has updated the `.csproj` files with the new version.

anthony-c-martin:
> 4. I _really_ like that you're leveraging the native tooling... we want to migrate this direction as much as we can.
> 5. Today we do all our dotnet updating via ruby/regex text parsing hackery... now that we'd have the native tooling available, does it also support updating the manifest itself? Such that we could start to leverage that to migrate away from our existing text hackery? Obviously something we'd handle in a different PR, but just want to know if we should be creating a new ticket to track opportunities for migration.

These are great points, and something I was wondering about. The `dotnet add <package>` will take care of both updating the `.csproj` file, and then updating lockfiles afterwards. There's also [dotnet outdated](https://github.com/dotnet-outdated/dotnet-outdated), but it's not built-in tooling, so you may want to understand the implications of taking a dependency.

A caveat that I can think of that needs investigating - You've already got some pretty sophisticated tooling here (for example, you're looking inside `Directory.Build.props` files for updates). I'm not sure whether the native tools actually support equivalent functionality for some of these edge cases.

ryanbrandenburg:
> I'm not sure whether the native tools actually support equivalent functionality for some of these edge cases.

I can confirm that `dotnet add package <PackageName> --version <version>` does not work in the `Directory.Build.props` scenario (it replaced `Version="$(PackageVersion)"` with `Version="1.2.3"`).

neotrow:
is there any news on this? Would love to see depandabot to support nuget lock files :-)

anthony-c-martin:
@jeffwidman sorry for the delay, I've merged and fixed up the CI.

nuget/lib/dependabot/nuget/file_updater.rb
```diff
@@ -174,6 +186,34 @@ def updated_declaration(old_declaration, previous_req, requirement)
...
         )
       end
+
+      def update_lock_files(files)
+        files = files.dup
+
+        lock_files = files.select { |f| lock_file?(f) }
+        lock_files.each do |lock_file|
+          project_file = files.find { |f| project_file?(f) && File.dirname(f.name) == File.dirname(lock_file.name) }
+          next if project_file.nil?
```
Nishnha, line 196: Can you explain this check? 

It seems like there can be more than one lockfile (`packages.lock.json`) per project?

anthony-c-martin: In general there's a lockfile for every project in a solution, and transitive dependencies are represented in each lockfile.

So for example, you could have a structure like the following:
* dirs.sln (solution file, contains refs to projA & projB)
* projA/projA.csproj (proj file, has nuget reference to nugetX)
* projB/projB.csproj (proj file, has project reference to projA)

If dependabot runs for the solution file (dirs.sln), and decides to bump nugetX, you would expect to see:
* projA/packages.lock.json has a direct dependency, and is updated
* projB/packages.lock.json has a transitive dependency, and is also updated

Here's a concrete example of what this looks like in my project:
* Bicep.sln [references Bicep.Core & Bicep.Cli](https://github.com/Azure/bicep/blob/e620efb6c8fa43f4f180938177e2db4777330ee0/Bicep.sln#L6-L11)
* Bicep.Core has a [direct dependency on SharpYaml](https://github.com/Azure/bicep/blob/e620efb6c8fa43f4f180938177e2db4777330ee0/src/Bicep.Core/Bicep.Core.csproj#L37)
* Bicep.Cli has a [project dependency on Bicep.Core](https://github.com/Azure/bicep/blob/e620efb6c8fa43f4f180938177e2db4777330ee0/src/Bicep.Cli/Bicep.Cli.csproj#L25C67-L25C67) (and thus a transitive dependency on SharpYaml)
* Both of these lockfiles should be updated if SharpYaml is upgraded:
    * [Bicep.Core](https://github.com/Azure/bicep/blob/e620efb6c8fa43f4f180938177e2db4777330ee0/src/Bicep.Core/packages.lock.json#L207-L212)
    * [Bicep.Cli](https://github.com/Azure/bicep/blob/e620efb6c8fa43f4f180938177e2db4777330ee0/src/Bicep.Cli/packages.lock.json#L675-L679)

anthony-c-martin:
@jeffwidman, @Nishnha - out of interest, is there any way to manually test this against a repo prior to merging?

Nishnha:
> @jeffwidman, @Nishnha - out of interest, is there any way to manually test this against a repo prior to merging?

You could set up a smoke test https://github.com/dependabot/smoke-tests

anthony-c-martin:
> > @jeffwidman, @Nishnha - out of interest, is there any way to manually test this against a repo prior to merging?
> 
> You could set up a smoke test https://github.com/dependabot/smoke-tests

Thanks! I've done this, and temporarily updated this branch to point to my fork, to demonstrate that it's working.

milkshakeuk:
@jeffwidman @anthony-c-martin @Nishnha are there any updates on this?

na1307:
Since the last commit, nuget's dependabot logic has been rewritten in C#. I think this PR should (probably) be updated.

JamieMagee:
Closing this for now, as NuGet support has largely been rewritten in C# and this PR has diverged a lot.

Thank you @anthony-c-martin for the implementation, and thank you @na1307 for picking it up.

PR #6037, opened 2022-11-03
Move location of `hex` version pin

jeffwidman:
Moves the location of the `hex` version pin to a more intuitive location.

This also centralizes the install logic within the helper builder script so that we don't call `mix local.hex` twice.

We could do it externally, but Hex requires being installed as the Dependabot user, and so we end up with the Hex install commands sitting right next to the build script... so rather than scattering the logic across the Dockerfile and build script, I opted to keep it all in the build script. This way if a developer is using the build script for local development w/o rebuilding the docker image, it's a repeatable experience.

### DIFF
--- a/hex/helpers/build
+++ b/hex/helpers/build
@@ -10,7 +10,14 @@ fi
 install_dir="$DEPENDABOT_NATIVE_HELPERS_PATH/hex"
 mkdir -p "$install_dir"
 
-mix local.hex --force
+# Initial Hex install - will always be the latest available version
+mix local.hex --force --if-missing
+# Annoyingly, a specific Hex version cannot be specified during the initial install.
+# The only way to pin is to re-install.
+if [ -n "$HEX_VERSION" ]; then
+  mix hex.install "$HEX_VERSION"
+fi
+
 mix archive.install hex nerves_bootstrap --force
 
 helpers_dir="$(dirname "${BASH_SOURCE[0]}")"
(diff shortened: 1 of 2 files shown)


jakecoffman (approved):
makes sense!

hex/helpers/build
```diff
@@ -10,7 +10,14 @@ fi
...
-mix local.hex --force
+# Initial Hex install - will always be the latest available version
+mix local.hex --force --if-missing
+# Annoyingly, a specific Hex version cannot be specified during the initial install.
+# The only way to pin is to re-install.
+if [ -n "$HEX_VERSION" ]; then
+  mix hex.install "$HEX_VERSION"
+fi
+
 mix archive.install hex nerves_bootstrap --force
```
jeffwidman, line 21: I'm a little uncertain on this... From what I read, the `mix archive.install` is a global install, so doing `mix archive.install hex` seems like it might be once again installing/overwriting the hex install? 🤷 I'll open an issue against the Hex repo and ask for their input...

jeffwidman:
CI Failure https://github.com/dependabot/dependabot-core/actions/runs/3388321608/jobs/5630170173 looks possibly related as the error is a `hex` build error, despite being on a `go` test... 

Since the others passed fine, I wonder if this is a race in Hex related stuff? I will re-run and merge if passes, and report it upstream to Hex team.

hex/helpers/build
jeffwidman: https://github.com/hexpm/hex/issues/962

PR #6038, opened 2022-11-03
ignore - purely WIP CI testing: bump hex without changing location

jeffwidman:
WIP purely for CI testing to see if somehow #6037 unexpectedly broke this upgrade causing the CI failures in #6026

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -323,4 +323,4 @@ RUN mkdir -p ~/.cargo && printf "[net]\ngit-fetch-with-cli = true\n" >> ~/.cargo
 ENV GIT_LFS_SKIP_SMUDGE=1
 
 # Pin to an earlier version of Hex. This must be run as dependabot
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
In case it has any relation to your investigation, I got a potentially related failure here: https://github.com/dependabot/dependabot-core/actions/runs/3393923229/jobs/5641835537.

jeffwidman:
Thanks for stopping by. That one doesn't match any failures I've observed previously, but I have seen intermittent download failures from the hex servers while trying to build the docker images, and that test failure looks very related... ie, a network flake could cause it. Given that its' a one-off failure, I suspect that's all it was in that case.

jeffwidman:
Root cause had nothing to do with this, it's explained here: https://github.com/dependabot/dependabot-core/pull/6026#issuecomment-1336276793

PR #6040, opened 2022-11-04
Yarn Berry: Subdependency fixes

pavera:
I noticed during some of my performance testing for peer dependency resolution that we were no longer updating subdependencies successfully because the yarn commands were not being run in the context of the checked out repo.

This PR addresses that issue by running inside the repo directory.


jeffwidman:
How hard is it to add a unit test to prevent future regressions?

jeffwidman (approved):

pavera:
> How hard is it to add a unit test to prevent future regressions?

I'll look at that. I won't merge until it's tested.

jurre (approved):

PR #6042, opened 2022-11-04
Upgrade Bundler to 2.3.25

Bo98:
Recently I've been experiencing an issue where Dependabot has stopped working and always [times out](https://github.com/Homebrew/brew/network/updates/499334214).

I've looked into why that was happening and it's a Bundler bug that was [fixed](https://github.com/rubygems/rubygems/pull/5698) in 2.3.24. Dependabot is however still using 2.3.22.

I've tested this change locally and it fixed the issue for me. `dependabot update` always hung before but now succeeeds.

The fix upstream changed some methods a bit, so I've updated a monkey-patch accordingly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -77,7 +77,7 @@ ARG RUBYGEMS_SYSTEM_VERSION=3.3.7
 
 ARG BUNDLER_V1_VERSION=1.17.3
 # When bumping Bundler, need to also regenerate `updater/Gemfile.lock` via `bundle update --bundler`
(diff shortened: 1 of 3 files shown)


jeffwidman (approved):
LGTM, but let's get @deivid-rodriguez to take a quick look since he will know full context on the monkeypatch change

deivid-rodriguez (approved):
Looks good, yeah!

deivid-rodriguez:
Thanks @Bo98!

PR #6043, opened 2022-11-04
Add note about why `3.9.3` omitted

jeffwidman:
Add note about why `3.9.3` excluded

### DIFF
--- a/python/lib/dependabot/python/python_versions.rb
+++ b/python/lib/dependabot/python/python_versions.rb
@@ -9,6 +9,9 @@ module Dependabot
 
       # Due to an OpenSSL issue we can only install the following versions in
       # the Dependabot container.
+      # NOTE: When adding one version, always doublecheck for additional releases: https://www.python.org/downloads/
+      #
+      # WARNING: 3.9.3 is purposefully omitted as it was recalled: https://www.python.org/downloads/release/python-393/
       SUPPORTED_VERSIONS = %w(
         3.11.0
         3.10.8 3.10.7 3.10.6 3.10.5 3.10.4 3.10.3 3.10.2 3.10.1 3.10.0


python/lib/dependabot/python/python_versions.rb
```diff
@@ -9,14 +9,15 @@ module PythonVersions
 
       # Due to an OpenSSL issue we can only install the following versions in
       # the Dependabot container.
```
jeffwidman, line 11: Here's the commit that added this: https://github.com/dependabot/dependabot-core/commit/e43c80f6e4521fbf59b01c6b97ea8071c359dead

So when this note was added, the vulnerable versions were in the 3.5.x series... 3.6 is now EOL from upstream Python, although we're unofficially made a number of choices to continue to support it until Ubuntu 18.04 goes EOL this coming April.

But at that point, we may want to refactor and drop this note.

jeffwidman: I filed https://github.com/dependabot/dependabot-core/issues/6044 to track this, as not something I have time to look into right now. For now IMO we should merge this to unblock ourselves/get correct, then later circle back and clean this up.

deivid-rodriguez (approved):

python/lib/dependabot/python/python_versions.rb
```diff
@@ -9,14 +9,15 @@ module PythonVersions
...
       SUPPORTED_VERSIONS = %w(
         3.11.0
-        3.10.7 3.10.6 3.10.5 3.10.4 3.10.3 3.10.2 3.10.1 3.10.0
-        3.9.14 3.9.13 3.9.12 3.9.11 3.9.10 3.9.9 3.9.8 3.9.7 3.9.6 3.9.5 3.9.4 3.9.2 3.9.1 3.9.0
-        3.8.14 3.8.13 3.8.12 3.8.11 3.8.10 3.8.9 3.8.8 3.8.7 3.8.6 3.8.5 3.8.4 3.8.3 3.8.2 3.8.1 3.8.0
-        3.7.14 3.7.13 3.7.12 3.7.11 3.7.10 3.7.9 3.7.8 3.7.7 3.7.6 3.7.5 3.7.4 3.7.3 3.7.2 3.7.1 3.7.0
-        3.6.15 3.6.14 3.6.13 3.6.12 3.6.11 3.6.10 3.6.9 3.6.8 3.6.7 3.6.6 3.6.5 3.6.4 3.6.3
-        3.6.2 3.6.1 3.6.0 3.5.10 3.5.8 3.5.7 3.5.6 3.5.5 3.5.4 3.5.3
+        3.10.8 3.10.7 3.10.6 3.10.5 3.10.4 3.10.3 3.10.2 3.10.1 3.10.0
+        3.9.15 3.9.14 3.9.13 3.9.12 3.9.11 3.9.10 3.9.9 3.9.8 3.9.7 3.9.6 3.9.5 3.9.4 3.9.3 3.9.2 3.9.1 3.9.0
```
ulgens, line 16: 3.9.3 is recalled, I don't think we should add it there: https://www.python.org/downloads/release/python-393/

jeffwidman: Thanks @ulgens I updated this PR with a note to that effect.

PR #6051, opened 2022-11-04
Run `go mod vendor` with `-e` flag

mattt:
I've seen some customer reports of Go updates failing with a generic `Dependabot::DependabotError` error in `GoModUpdater#update_files`. I noticed that we run `go mod tidy -e` (with `-e` flag) and then `go mod vendor` (without the `-e` flag). From `go help mod vendor`:

> The -e flag causes vendor to attempt to proceed despite errors
encountered while loading packages.

This PR adds the `-e` flag to match the behavior of our `go mod tidy` invocation in the hopes of resolving the underlying issue of failed update jobs.

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
@@ -148,7 +148,7 @@ module Dependabot
         def run_go_vendor
           return unless vendor?
 
-          command = "go mod vendor"
+          command = "go mod vendor -e"
           _, stderr, status = Open3.capture3(environment, command)
           handle_subprocess_error(stderr) unless status.success?
         end


jurre (approved):
That makes sense to me, thanks

jakecoffman:
Does running with `-e` mean Dependabot would potentially be creating invalid PRs?

mctofu:
> Does running with `-e` mean Dependabot would potentially be creating invalid PRs?

Yeah, I think so? The consequence of `go mod tidy` failing is you could have some cruft leftover in your mod files. If `go mod vendor` fails the PR wouldn't have the right files in the vendor dir which should break your build.

jeffwidman:
I've looked into the `go mod tidy -e` behavior before, and I think we:
1. want `go mod tidy -e`
2. don't want `go mod vendor -e`

I personally wouldn't ship this as a general improvement for consistency purposes. If there was a smoking gun bug that was definitively happening due to `go mod vendor` not having the `-e` flag, I'd still be very hesitant because of the side-effects on other repos, and instead would look for workarounds, or even asking upstream to reconsider what gets ignored by the `-e` flag.

PR #6056, opened 2022-11-05
Added useful error message when image not found

honeyankit:
## Context

When an image is not found in the private registry, it throws an exception of `DockerRegistry2::NotFound` and lot of stack trace but this info is not useful for the end user. 
```
updater | ERROR <job_500122728> Error processing myreg/ubuntu (DockerRegistry2::NotFound)
updater | ERROR <job_500122728> 404 Not Found
updater | ERROR <job_500122728> /home/dependabot/dependabot-updater/vendor/ruby/3.1.0/gems/docker_registry2-1.12.0/lib/registry/registry.rb:285:in `rescue in doreq'
updater | ERROR <job_500122728> /home/dependabot/dependabot-updater/vendor/ruby/3.1.0/gems/docker_registry2-1.12.0/lib/registry/registry.rb:269:in `doreq'
updater | ERROR <job_500122728> /home/dependabot/dependabot-updater/vendor/ruby/3.1.0/gems/docker_registry2-1.12.0/lib/registry/registry.rb:26:in `doget'

```
Error logs: https://github.com/dsp-testing/docker-firewalled-testing/network/updates/500122728

## Approach
I'm using a rescue block to catch the "DockerRegistry2::NotFound" exception and throwing it with useful error message when a Dependabot tries to retrieve an image from the registry but it doesn't exist there.

### DIFF
--- a/docker/lib/dependabot/docker/update_checker.rb
+++ b/docker/lib/dependabot/docker/update_checker.rb
@@ -224,6 +224,7 @@ module Dependabot
         @tags_from_registry ||=
           begin
             client = docker_registry_client
+
             client.tags(docker_repo_name, auto_paginate: true).fetch("tags")
           rescue *transient_docker_errors
             attempt ||= 1
--- a/docker/lib/dependabot/docker/update_checker.rb
+++ b/docker/lib/dependabot/docker/update_checker.rb
@@ -234,7 +234,7 @@ module Dependabot
             retry
           end
       rescue DockerRegistry2::NotFound
-        raise DockerRegistry2::NotFound, "404 Not Found. Image https://#{registry_hostname}/#{repo} not found"
+        raise DockerRegistry2::NotFound, "Image not found: https://#{registry_hostname}/#{docker_repo_name}"
       rescue DockerRegistry2::RegistryAuthenticationException,
              RestClient::Forbidden
         raise PrivateSourceAuthenticationFailure, registry_hostname
--- a/docker/lib/dependabot/docker/update_checker.rb
+++ b/docker/lib/dependabot/docker/update_checker.rb
@@ -225,7 +225,7 @@ module Dependabot
           begin
             client = docker_registry_client
             repo = docker_repo_name
-            client.tags(repo, auto_paginate: true).fetch("tags")
+            client.tags(docker_repo_name, auto_paginate: true).fetch("tags")
           rescue *transient_docker_errors
             attempt ||= 1
             attempt += 1
--- a/docker/lib/dependabot/docker/update_checker.rb
+++ b/docker/lib/dependabot/docker/update_checker.rb
@@ -224,8 +224,8 @@ module Dependabot
         @tags_from_registry ||=
           begin
             client = docker_registry_client
-
-            client.tags(docker_repo_name, auto_paginate: true).fetch("tags")
+            repo = docker_repo_name
+            client.tags(repo, auto_paginate: true).fetch("tags")
           rescue *transient_docker_errors
             attempt ||= 1
             attempt += 1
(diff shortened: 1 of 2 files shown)


docker/lib/dependabot/docker/update_checker.rb
```diff
@@ -224,15 +224,17 @@ def tags_from_registry
         @tags_from_registry ||=
           begin
             client = docker_registry_client
-
-            client.tags(docker_repo_name, auto_paginate: true).fetch("tags")
+            repo = docker_repo_name
+            client.tags(repo, auto_paginate: true).fetch("tags")
```
deivid-rodriguez, line 228: Is the above necessary?

docker/lib/dependabot/docker/update_checker.rb
```diff
@@ -224,15 +224,17 @@ def tags_from_registry
...
           rescue *transient_docker_errors
             attempt ||= 1
             attempt += 1
             raise if attempt > 3
 
             retry
           end
+      rescue DockerRegistry2::NotFound
+        raise DockerRegistry2::NotFound, "404 Not Found. Image https://#{registry_hostname}/#{repo} not found"
```
deivid-rodriguez, line 237: Would it make sense to improve this upstream as well, so we can eventually remove this? https://github.com/deitch/docker_registry2/blob/970b4aafd1259e4867293f9a6be180a46730b366/lib/registry/registry.rb#L284-L285. Just thinking that the enhancement may be useful for all users of the `docker_registry2` gem.

deivid-rodriguez (commented):
Thanks for improving error messages! I just added two minor comments.

docker/lib/dependabot/docker/update_checker.rb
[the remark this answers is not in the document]
honeyankit: This looks good as well. I was eliminating to have two function calls to `docker_repo_name` so implemented repo = docker_repo_name as similar thing was done in [file_parser.rb](https://github.com/dependabot/dependabot-core/blob/main/docker/lib/dependabot/docker/file_parser.rb#L105)

docker/lib/dependabot/docker/update_checker.rb
honeyankit: This looks good as well. I was eliminating to have two function calls to docker_repo_name (one here and another in the rescue block )so implemented repo = docker_repo_name as similar thing was done in [file_parser.rb](https://github.com/dependabot/dependabot-core/blob/main/docker/lib/dependabot/docker/file_parser.rb#L105)

docker/lib/dependabot/docker/update_checker.rb
honeyankit: I agree with you. I will create an issue to improve this upstream. Thank you for pointing it out.

honeyankit: Issue created in upstream: https://github.com/deitch/docker_registry2/issues/67

jeffwidman:
IMO let's hold off merging this for a couple of days to see if we can land a fix in upstream, as it should be straightforward and that'd be less work than merging this then backing it out when upstream lands a fix. I pinged the maintainer on @honeyankit 's issue, hopefully we get a quick response.

jeffwidman:
Pending fix upstream: https://github.com/deitch/docker_registry2/pull/68

deivid-rodriguez:
For what it's worth, I'm looking into actually deleting all of this code at https://github.com/dependabot/dependabot-core/pull/6150 since I believe it's problematic.

jeffwidman:
IMO we should not merge this because even if if the underlying code doesn't get deleted in #6150, the upstream lib will still get patched in https://github.com/deitch/docker_registry2/pull/68.

So not worth the trouble to add it here, then pulling it back out once upstream fix arrives...

honeyankit:
This PR is not relevant as the fix is coming from upstream: https://github.com/deitch/docker_registry2/pull/68.
So closing this PR.

PR #6068, opened 2022-11-07
docker support for Containerfile name

### LINKED ISSUE #6067: Support for files named `Containerfile` rather than `Dockerfile`

### Feature description

OCI tooling like podman and buildah supports and suggests the more neutral name `Containerfile`. These use the same syntax as `Dockerfile`. It would be nice if the `docker` ecosystem supported this alternate file name.

andrewazores:
Fixes #6067


andrewazores:
Not really a Rubyist here and not sure how to update the VCR fixtures properly.

jeffwidman:
👋 Thanks for putting this up. I keep meaning to review it but just haven't had time yet. One of us will try to get to it in the next few days.

dhirschfeld:
Any interest in allowing a full filepath since neither docker nor buildah *require* any particular naming convention for the file - you can call it whatever you want and point docker/buildah to it with `--file`

https://docs.docker.com/engine/reference/commandline/build/#specify-a-dockerfile--f

dhirschfeld:
My interest in this feature is that I often have multiple dockerfiles alongside each other in the same repository.

I use a `meaningful-name.docker` convention and name my images after the filename they were created from. It would be great if I could somehow tell dependabot to look for `meaningful-name.docker` instead of hardcoding `Dockerfile`.

e.g. maybe something like
```yaml
updates:
  - package-ecosystem: "docker"
    directory: "/build/docker"
    manifest-name: "meaningful-name.docker"
```
This could be made backward compatible by having a `package-ecosystem` default value for `manifest-name` which is just the existing hardcoded value.

ssbarnea:
That also made me curious, is there any YAML file on which dependabot github app can update container versions? 

I see that the source code mentions kubernetes YAML files but there is nothing inside the documentation for that.

sid-maddy:
Could this PR be picked up for review? AFAICT, it only needs a rebase onto `main`.

Also, from the comments here and in [the linked issue](https://github.com/dependabot/dependabot-core/issues/6067), there seems to be interest in allowing users to provide the manifest names via the Dependabot config file.  
AFAICT, this would need changes in `docker`'s file fetcher & updater ([my WIP attempt](https://github.com/sid-maddy/dependabot-core/tree/sid/feature/docker/custom-manifests)) _as well as_ the logic (not sure where) that parses package manager specific options from the config file.

Could that be included in this PR as well?

andrewazores:
Done by https://github.com/dependabot/dependabot-core/pull/11141 instead

abdulapopoola:
Closing since this has now been fixed

PR #6052, opened 2022-11-04
Fix GitHub actions updater crashing and cloning the wrong repository

deivid-rodriguez:
This PR wants to fix two issues:

* After #5951, we seem to have started using the `GitCommitChecker#head_commit_for_current_branch` method when `dependency.version` is an actual numbered version and not a ref. This is probably because with #5951 we have to deal with both numeric and ref versions in the same update. The current implementation of this method does not support this. I expect 90612fdda8a2471b66e78f341a649b013f6309a5 to fix this issue.

* Even with the above fix, updates for pinned refs still don't work as expected and we either get a different crash or a downgrade. The reason for this seems to lie in #5576. In that PR, in order to be able to properly detect whether actions pinned to arbitrary references are out of date, we started cloning the whole repository. However, I think the implementation there was incorrect and it started cloning the repository being updated (the one that includes the workflow files), not the repository of the dependency being updated (the repository of the action itself). I expect 721bd2df6cdd8462ddb6b3ab0adfd76978487831 to fix this issue.

With this I get proper updates of repositories pinned to specific references.

Fixes #6014.
Fixes #6045.

PR #6071, opened 2022-11-07
Add security update support for Github Actions

deivid-rodriguez:
It currently sits on top of #6052, because they were updating the same part of the code and I didn't want conflicts.

Still needs tests but just open a WIP PR for now.


deivid-rodriguez:
This one is ready for a review now! :)

github_actions/lib/dependabot/github_actions/update_checker.rb
```diff
@@ -82,6 +92,27 @@ def fetch_latest_version_for_git_dependency
...
+
+      def fetch_lowest_security_fix_version_for_git_dependency
+        lowest_security_fix_version_tag.fetch(:version)
+      end
+
+      def lowest_security_fix_version_tag
+        @lowest_security_fix_version_tag ||= begin
+          relevant_tags = git_commit_checker.local_tags_for_allowed_versions
+
+          relevant_tags = Dependabot::UpdateCheckers::VersionFilters.filter_vulnerable_versions(relevant_tags,
```
Nishnha, line 110: We may also filter out ignored versions here - might be best to check another ecosystem to confirm.

common/lib/dependabot/git_commit_checker.rb
```diff
@@ -404,19 +402,26 @@ def wants_prerelease?
...
-      version_class.new(version).prerelease?
+      version_from_tag(tag).prerelease?
+    end
+
+    def version_from_tag(tag)
+      version_from_ref(tag.name)
+    end
+
+    def version_from_ref(name)
+      version = name.match(VERSION_REGEX).named_captures.fetch("version")
```
Nishnha, line 423: the version could return nil here and error out on the next line

github_actions/lib/dependabot/github_actions/update_checker.rb
deivid-rodriguez: Good point @Nishnha. Looking at other ecosystems, some filter out ignored versions, some don't. And also, some filter out lower versions, and some don't. Not sure about ignored versions, but for lower versions I believe the logic in those other ecosystems is wrong because even if it seems ok to make a best effort to not downgrade the dependency, we should still downgrade if strictly necessary, and I don't think other ecosystems are doing that... 🤔.

Anyways, I was thinking on punting on this now and maybe return to it later if issues are reported.

common/lib/dependabot/git_commit_checker.rb
deivid-rodriguez: I believe this method works on "version tags" only, i.e., tags that look line versions and where we know this regex will match.

github_actions/lib/dependabot/github_actions/update_checker.rb
Nishnha: when we were working on npm transitive dependency support we made it a point to filter out any lower versions to avoid downgrading dependencies - we would rather fail the job than downgrade. So I think we would want to do that here too

cc/ @mctofu

deivid-rodriguez: Ok, so we prefer to leave the alert unresolved over proposing a downgrade, correct? I'm not sure I agree, but I'm happy to mimic that behavior! I think in the current state a downgrade would be proposed, so it needs some changes.

deivid-rodriguez: Actually, I'm looking into it now and all ecosystems seem consistent here both in terms of ignored versions and not downgrading. I will update this.

deivid-rodriguez: I reviewed this now. Ignored versions were already being ignored because that logic is actually built-in to `GitCommitChecker`.

Regarding lower versions, I actually learned that the logic to avoid downgrade PRs is actually very necessary because if there's both fixed versions higher and lower than the current version, we would be choosing a downgrade over an upgrade otherwise. I had actually missed this so that was a pretty bad bug.

I fixed the lower versions case and added specs for both ignored and lower versions 👍.

deivid-rodriguez:
Alright, so I addressed review comments, let me update the status of this PR.

The PR currently implements the standard "lowest fixed version" logic of other ecosystems with the following particularities:
  
* It tries to update to the lowest fixed version that keeps the existing version precision.  So a vulnerable `v2` would go to `v3` if possible. This seems ✅ to me.

* If a fixed version keeping precision does not exist, it changes precision in favor of fixing the vulnerability. So say `v2` is vulnerable, because it points to `v2.7.4` and the lowest fixed version is `2.7.5`. And say there's no fixed `v3` yet. In that case we propose an update from `v2` to `v2.7.5`. Maybe this one is too smart, not fully sure if users will like this change in precision or if it's best to not create an update at all ❓.

* If pinned to a major version tag which belongs to the vulnerable series but points to a non vulnerable tag, it does not propose an update. So in the case above, if pinned to `v2` pointing to `v2.7.5` (the latest patch, most common case), then everything is ok and we don't create an update. This one also seems ✅, just noting that this is currently a bug in DG, so although we may receive false positive alerts due to it, we won't create alerts for them.

Finally, I went a bit overboard an added a few cross-ecosystem refactoring that are not directly related to this PR. May I split them to a separate PR?

deivid-rodriguez:
I totally agree with your points!

> The main problem, I think, is in communicating why Dependabot is suggesting a particular change. I don't expect most users to understand the idiosyncrasies of Actions versioning, but I'm not sure what we can really do about that. In the interest of shipping to learn, I'd be happy going with this behavior and seeing what kind of reception it gets from users.

Yes, that was my biggest hesitation here, that users may not understand why we're proposing this update changing precision. I guess we could add a note about this to the PR, but I'm not fully sure how to go about implementing that.

> Yeah, let's evaluate those separately.

Sure, let me split them!

jurre (approved):

PR #6074, opened 2022-11-08
Fix multiple constraint Poetry dependencies getting bad updates

### LINKED ISSUE #5862: Poetry multiple-constraint dependencies broken by unrelated dependency update

### Package ecosystem

pip

### Package manager version

Poetry (version 1.2.1)

### Language version

python = ">=3.7,<3.10"

`pyproject.toml`: https://github.com/greschd/dependabot-poetry-multiple-constraints/blob/4c051f7216cbb70decb1b206fe8de7ff060c188a/pyproject.toml
`poetry.lock`: https://github.com/greschd/dependabot-poetry-multiple-constraints/blob/4c051f7216cbb70decb1b206fe8de7ff060c188a/poetry.lock

### dependabot.yml content

https://github.com/greschd/dependabot-poetry-multiple-constraints/blob/4c051f7216cbb70decb1b206fe8de7ff060c188a/.github/dependabot.yml

### Updated dependency

`wheel` updated from `0.37.0` to `0.37.1`

Only the `wheel` related entries in `poetry.lock` should be updated. Instead, the multiple-constraint `numpy` dependency is partially removed.

It seems that dependabot doesn't understand the [multiple-constraint dependency](https://python-poetry.org/docs/dependency-specification/#multiple-constraints-dependencies) (different `numpy` version for different `python` versions)

```toml
numpy = [
    {version = ">=1.15,<1.22", python = ">=3.7,<3.8"},
    {version = ">=1.22", python = ">=3.8,<3.11"}
]
```

This is mentioned in https://github.com/dependabot/dependabot-core/issues/2715#issuecomment-777435053, but I couldn't find an issue specific to this incompatibility.

In the `poetry.lock`, the multiple-constraint dependency is added as multiple `numpy` entries (AFAICT, the last matching one is selected for a given Python version). Dependabot updates keep only the first entry, and discard the rest.

PR created manually, using `poetry lock`: https://github.com/greschd/dependabot-poetry-multiple-constraints/pull/2

https://github.com/greschd/dependabot-poetry-multiple-constraints/pull/1

Almost minimal repository: https://github.com/greschd/dependabot-poetry-multiple-constraints

deivid-rodriguez:
Before trying to generate a lockfile with an upgraded dependency, Dependabot first explicitly updates the pyproject.toml file like the following:
* It pins the dependency to be upgraded to the resolved versions.
* It pins all other dependencies to the locked version.

In this particular case, the numpy dependency is getting locked to 1.21.6, but this is incorrect, because it's a multiple constraint dependency, so locking it explicitly will cause some versions to be removed from the lock file.

In general, the second step does not seem necessary to me. The command we run to update the lock file, `poetry update <dependency> --lock` should deal with it fine and keep the locked dependencies not being upgraded untouched.

However, when removing all that logic, some specs fail since they let other dependencies be upgraded and I'm not sure why. So as an alternative solution for the time being, I'm just explicitly avoiding pinning multiple constraint dependencies.

Fixes #5862.

### DIFF
--- a/python/lib/dependabot/python/file_updater/pyproject_preparer.rb
+++ b/python/lib/dependabot/python/file_updater/pyproject_preparer.rb
@@ -72,6 +72,10 @@ module Dependabot
                 }
               elsif poetry_object[key][dep_name].is_a?(Hash)
                 poetry_object[key][dep_name]["version"] = locked_version
+              elsif poetry_object[key][dep_name].is_a?(Array)
+                # if it has multiple-constraints, locking to a single version is
+                # going to result in a bad lockfile, ignore
+                next
               else
                 poetry_object[key][dep_name] = locked_version
               end
--- a/python/spec/dependabot/python/file_updater/pyproject_preparer_spec.rb
+++ b/python/spec/dependabot/python/file_updater/pyproject_preparer_spec.rb
@@ -136,6 +136,19 @@ RSpec.describe Dependabot::Python::FileUpdater::PyprojectPreparer do
       it { is_expected.to include("geopy = \"^1.13\"\n") }
     end
 
+    context "with a multiple constraint dependency" do
+      let(:dependencies) { [] }
+
+      let(:pyproject_lock_fixture_name) { "multiple_constraint_dependency.lock" }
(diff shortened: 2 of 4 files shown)


pavera (approved):
lgtm.

Issue #6041, opened 2022-11-04
"Malformed version number string 929ab5b" updating nixos/nix image


l0b0:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

Docker

### Package manager version

N/A

### Language version

N/A

### Manifest location and content before the Dependabot update

```
FROM nixos/nix:2.11.0
```

### dependabot.yml content

```
---
version: 2
updates:
  - package-ecosystem: docker
    directory: /
    schedule:
      interval: daily
    commit-message:
      prefix: build(deps)
```

### What you expected to see, versus what you actually saw

The following output from Dependabot:

```
updater | I, [2022-11-02T14:18:29.936751 #7]  INFO -- sentry: ** [Raven] Sending event 340d10dfd574458c9322b6aee565024e to Sentry
  proxy | 2022/11/02 14:18:30 [026] POST https://sentry.io:443/api/1451818/store/
  proxy | 2022/11/02 14:18:30 [026] 200 https://sentry.io:443/api/1451818/store/
updater | ERROR <job_498258213> Error processing nixos/nix (ArgumentError)
updater | ERROR <job_498258213> Malformed version number string 929ab5b
updater | ERROR <job_498258213> /usr/local/lib/ruby/3.1.0/rubygems/version.rb:217:in `initialize'
updater | ERROR <job_498258213> /usr/local/lib/ruby/3.1.0/rubygems/version.rb:208:in `new'
updater | ERROR <job_498258213> /usr/local/lib/ruby/3.1.0/rubygems/version.rb:208:in `new'
updater | ERROR <job_498258213> /home/dependabot/docker/lib/dependabot/docker/version.rb:18:in `initialize'
updater | ERROR <job_498258213> /usr/local/lib/ruby/3.1.0/rubygems/version.rb:206:in `new'
updater | ERROR <job_498258213> /usr/local/lib/ruby/3.1.0/rubygems/version.rb:206:in `new'
updater | ERROR <job_498258213> /home/dependabot/docker/lib/dependabot/docker/update_checker.rb:179:in `block in remove_version_downgrades'
updater | ERROR <job_498258213> /home/dependabot/docker/lib/dependabot/docker/update_checker.rb:178:in `select'
updater | ERROR <job_498258213> /home/dependabot/docker/lib/dependabot/docker/update_checker.rb:178:in `remove_version_downgrades'
[...]

PR #6075, opened 2022-11-08
Fix processing docker image names ending with `_<sha>`

deivid-rodriguez:
We accept as part of the version because of being sortable/comparable, images ending with `_[0-9]`, like the update segment in Java versions. But anything other than that should not be considered part of the numeric version.

Fixes https://github.com/dependabot/dependabot-core/issues/6041.


deivid-rodriguez:
Now ready!

jeffwidman (approved):

PR #6077, opened 2022-11-08
Yarn Berry: Adding a peer dependency checker

pavera:
## Context
Previous to this PR we were still running yarn classic to check for peer dependency issues created by a proposed version update. This was slow and probably could result in incorrect answers.

## Approach
I've added a yarn berry specific checker based on the output I could see after adding a dependency at a version that breaks a peer dependency request. This may be naive, and almost certainly does not catch all cases so I'd appreciate feedback around how to make this check more accurate. One specific shortcoming is that without making additional calls to `yarn explain` the versions of the the peer dependency mismatch are not available. In the short term this may cause us to fail to create a possible PR if we could have resolved an update to an earlier version.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
@@ -50,11 +50,7 @@ module Dependabot
         end
       end
 
-      # Run any number of yarn commands while ensuring that `enableScripts` is
-      # set to false. Yarn commands should _not_ be ran outside of this helper
-      # to ensure that postinstall scripts are never executed, as they could
-      # contain malicious code.
-      def self.run_yarn_commands(*commands)
+      def self.setup_yarn_berry
         # Always disable immutable installs so yarn's CI detection doesn't prevent updates.
         SharedHelpers.run_shell_command("yarn config set enableImmutableInstalls false")
         # We never want to execute postinstall scripts, either set this config or mode=skip-build must be set
@@ -74,7 +70,19 @@ module Dependabot
             SharedHelpers.run_shell_command("yarn config set caFilePath #{ca_file_path}")
           end
         end
+      end
+      # Run any number of yarn commands while ensuring that `enableScripts` is
+      # set to false. Yarn commands should _not_ be ran outside of this helper
+      # to ensure that postinstall scripts are never executed, as they could
+      # contain malicious code.
+      def self.run_yarn_commands(*commands)
+        setup_yarn_berry
         commands.each { |cmd| SharedHelpers.run_shell_command(cmd) }
+      end
+
+      def self.run_yarn_command(command)
+        setup_yarn_berry
+        SharedHelpers.run_shell_command(command)
       end
 
       def self.dependencies_with_all_versions_metadata(dependency_set)
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
@@ -44,9 +44,9 @@ module Dependabot
         if yarn_major_version == 2
           ""
         elsif yarn_major_version >= 3 && yarn_zero_install?
-          " --mode=skip-build"
+          "--mode=skip-build"
         else
-          " --mode=update-lockfile"
+          "--mode=update-lockfile"
         end
       end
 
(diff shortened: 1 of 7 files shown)


npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
```diff
@@ -74,9 +70,21 @@ def self.run_yarn_commands(*commands)
...
+      # Run any number of yarn commands while ensuring that `enableScripts` is
+      # set to false. Yarn commands should _not_ be ran outside of this helper
+      # to ensure that postinstall scripts are never executed, as they could
+      # contain malicious code.
+      def self.run_yarn_commands(*commands)
+        setup_yarn_berry
         commands.each { |cmd| SharedHelpers.run_shell_command(cmd) }
       end
 
+      def self.run_yarn_command(command)
```
jurre, line 83: I think this is functionally the same as the `run_yarn_commands` method we already have, where we can pass one or more arguments, if we're worried about this reading weird, we could maybe alias the method, or delegate to the original method, to avoid the two from diverging in functionality over time?

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -468,13 +481,33 @@ def lockfiles_for_path(lockfiles:, path:)
...
         end
 
+        def run_yarn_berry_checker(path:, version:)
+          SharedHelpers.with_git_configured(credentials: credentials) do
+            Dir.chdir(path) do
+              output = Helpers.run_yarn_command("yarn add #{dependency.name}@#{version}#{Helpers.yarn_berry_args}")
+              if output.include?("YN0060")
+                raise SharedHelpers::HelperSubprocessFailed.new(
+                  message: output,
+                  error_class: "JSON::ParserError",
```
jurre, line 503: Why are we raising a `JSON::ParseError` in this scenario?

pavera: Lack of knowledge about other error times that might be more appropriate :)

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -468,13 +481,33 @@ def lockfiles_for_path(lockfiles:, path:)
...
+            end
           end
 
           run_npm_checker(path: path, version: version)
         end
 
+        def run_yarn_berry_checker(path:, version:)
+          SharedHelpers.with_git_configured(credentials: credentials) do
+            Dir.chdir(path) do
+              output = Helpers.run_yarn_command("yarn add #{dependency.name}@#{version}#{Helpers.yarn_berry_args}")
```
Nishnha, line 499: This is a really clever solution but it took me a while to figure it out. Might be helpful to include a comment explaining how it works

npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
```diff
@@ -74,9 +70,21 @@ def self.run_yarn_commands(*commands)
             SharedHelpers.run_shell_command("yarn config set caFilePath #{ca_file_path}")
           end
         end
+      end
+      # Run any number of yarn commands while ensuring that `enableScripts` is
+      # set to false. Yarn commands should _not_ be ran outside of this helper
+      # to ensure that postinstall scripts are never executed, as they could
+      # contain malicious code.
+      def self.run_yarn_commands(*commands)
+        setup_yarn_berry
```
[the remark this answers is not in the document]
pavera: The 2 methods `#run_yarn_command` and `#run_yarn_commands` are not dependent on each other, in a given temp repo directory one or the other might be called without calling the other, so they both need to configure the yarn berry environment. I agree with @jurre's comment above this is not clear so open to ideas about how to improve. Maybe I should just pay the piper and return a hash from `#run_yarn_commands`

npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
jurre: Ah, I see, yeah that makes sense. I wish the difference between these methods was a bit more evident but I'm not sure what we can reasonably change to make it so

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
jurre: https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/errors.rb is where all the top-level errors that are handled in the backend service are defined. But looking at this closer, it seems like the `error_class` is not really used here, since we catch all `HelperSubprocessFailed` errors in the calling code. Seems like we can just leave it out here?

jurre:
This is looking great! Are there any test-cases that we could add for things that were previously failing, or is it mainly just a performance pass that we're doing by using the new version of yarn here?

agrobbin:
@pavera is it still true, even after this change, that Dependabot will not attempt to update 2 dependencies that are peers of each other if they are *not* `devDependencies`? That's what's alluded to here:

https://github.com/dependabot/dependabot-core/blob/a766bf403492309d7e532d791873a4bccf6f9598/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/package_json_updater.rb#L209-L212

I'm thinking of the `@datadog/rum` and `@datadog/logs` packages, which are peers of each other, and must be updated together. As of right now, when those packages are updated, Dependabot creates 2 separate PRs which have to be manually combined.

jakecoffman (approved):

PR #6080, opened 2022-11-08
Fix docker version regexp matching too much

deivid-rodriguez:
This was detected by CodeQL in #6075 as inefficient Regexp.

I extracted it from that PR since it's an unrelated fix.

I had a look and yeah, I believe the Regexp was too unspecific and can be made more specific and thus easier to process.

### DIFF
--- a/docker/lib/dependabot/docker/update_checker.rb
+++ b/docker/lib/dependabot/docker/update_checker.rb
@@ -42,8 +42,7 @@ end
 module Dependabot
   module Docker
     class UpdateChecker < Dependabot::UpdateCheckers::Base
-      VERSION_REGEX =
-        /v?(?<version>[0-9]+(?:\.[_a-z0-9]+|-(?:kb)?[0-9]+)*)/i
+      VERSION_REGEX = /v?(?<version>[0-9]+(?:\.[_a-z0-9]+|-(?:kb)?[0-9]+)*)/i
       VERSION_WITH_SFX = /^#{VERSION_REGEX}(?<suffix>-[a-z][a-z0-9.\-]*)?$/i
       VERSION_WITH_PFX = /^(?<prefix>[a-z0-9][a-z0-9.\-]*-)?#{VERSION_REGEX}$/i
       VERSION_WITH_PFX_AND_SFX = /^(?<prefix>[a-z\-]+-)?#{VERSION_REGEX}(?<suffix>-[a-z\-]+)?$/i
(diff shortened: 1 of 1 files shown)


docker/lib/dependabot/docker/update_checker.rb
```diff
@@ -43,9 +43,9 @@ module Dependabot
   module Docker
     class UpdateChecker < Dependabot::UpdateCheckers::Base
       VERSION_REGEX =
-        /v?(?<version>[0-9]+(?:(?:\.[_a-z0-9]+)|(?:-(?:kb)?[0-9]+))*)/i
-      VERSION_WITH_SFX = /^#{VERSION_REGEX}(?<suffix>-[a-z0-9.\-]+)?$/i
-      VERSION_WITH_PFX = /^(?<prefix>[a-z0-9.\-]+-)?#{VERSION_REGEX}$/i
+        /v?(?<version>[0-9]+(?:\.[_a-z0-9]+|-(?:kb)?[0-9]+)*)/i
```
jeffwidman, line 46:
```suggestion
      VERSION_REGEX = /v?(?<version>[0-9]+(?:\.[_a-z0-9]+|-(?:kb)?[0-9]+)*)/i
```

No need for the newline now that it's shorter, and the newline is making it more painful to review than it needs to be...

jeffwidman (approved):

docker/lib/dependabot/docker/update_checker.rb
jeffwidman, line 46:
```
      VERSION_REGEX = /v?(?<version>[0-9]+(?:\.[_a-z0-9]+|-(?:kb)?[0-9]+)*)/i
```

No need for the newline now that it's shorter, and the newline is making it more painful to review than it needs to be... 

Unfortunately I can't do this as a `suggestion` since it thinks I'm trying to replace other lines too...

deivid-rodriguez: Happy to remove the newline but I don't understand how introducing less changes makes the PR harder to review? Is it because the new line distracts you? To me limiting the diff to what I want to fix is helpful, GitHub also displays in darker green or red the specific changes:

Captura de pantalla 2022-11-08 a las 21 21 30

Anyways, let me remove the new line now. This kind of artificial line breaks is why I never set any hard limits on line length.

deivid-rodriguez: To complete my point, this is the new diff, in my eyes it's harder to review

Captura de pantalla 2022-11-08 a las 21 26 38

jeffwidman: * I was using `diff view: split` setting
* I wasn't so much reviewing the delta between old/new, as reviewing the flow of how the regexes chain their inheritance to be sure the new code's logic made sense... having them stacked vertically such that all three lines begin at nearly the same column makes it easier.

jeffwidman (approved):

docker/lib/dependabot/docker/update_checker.rb
jeffwidman: Yeah, I forget the "unified" setting even exists because I use it so rarely... the split diff allows me to trace new code flow w/o getting interuptted by the dead code... here's what I see now:

image

PR #6082, opened 2022-11-08
Fix updating GitHub Actions & Dockerfiles with mixed versions

### LINKED ISSUE #3750: GitHub Actions aren't bumped if two files have different versions of an action
**Package ecosystem**
GitHub Actions
**Package manager version**
Latest on github.com
**Language version**
Latest on github.com

`.github/workflows/plantuml.yml`
````
[...]
    - name: checkout
      uses: actions/checkout@v1
[...]
````
`.github/workflows/build.yml`
````
[...]
    - name: checkout
      uses: actions/checkout@v2
[...]
````
**dependabot.yml content**
````yaml
version: 2
updates:
- package-ecosystem: nuget
  directory: "/"
  schedule:
    interval: daily
  open-pull-requests-limit: 10
- package-ecosystem: github-actions
  directory: "/"
  schedule:
    interval: daily
  open-pull-requests-limit: 10

````
**Updated dependency**
None
I expect to see a PR that bumps the checkout action to version 2 in `.github/workflows/plantuml.yml`, but there isn't one.

The only occurence of `checkout` in the log: 
````
updater | INFO <job_135451773> Checking if actions/checkout 2 needs updating
  proxy | 2021/05/19 07:01:27 [014] GET https://github.com:443/actions/checkout.git/info/refs?service=git-upload-pack
  proxy | 2021/05/19 07:01:27 [014] * authenticating git server request (host: github.com)
  proxy | 2021/05/19 07:01:27 [014] 200 https://github.com:443/actions/checkout.git/info/refs?service=git-upload-pack
updater | INFO <job_135451773> Latest version is 2
updater | INFO <job_135451773> No update needed for actions/checkout 2
````

However both files are monitored:
grafik

When sorted lexically `build.yml` comes bofore `plantuml.yml`. Since this repo is private, I cannot link to it.

### LINKED ISSUE #5633: Updates in projects with mixed versions

### Package ecosystem

NuGet

### Package manager version

_No response_

### Language version

_No response_

_No response_

### dependabot.yml content

```
version: 2

updates:
    directory: "/"
    schedule:
      interval: "weekly"
      
  - package-ecosystem: "nuget"
    directory: "/ProtocolEditor/"
    schedule:
      interval: "weekly"
    ignore:
      - dependency-name: "*"
        # ignore all major updates
        update-types: ["version-update:semver-major"]
```

### Updated dependency

Microsoft.VSSDK.BuildTools

When two projects use a different version (range) of a certain NuGet package, and one of the packages have an update available, a PR is opened that updates both projects.

Solution with two (csproj) projects that use a different version of a nuget:
- project A: NuGet x 1.0
- project B: NuGet x 2.0

Both ranges of the NuGet receive an update: 1.1 and 2.1

PR is opened that updates the projects as follows:
- project A: NuGet x 1.1
- project B: NuGet x 1.1

This is a problem for us because we develop a Visual Studio extension that targets two versions of Visual Studio. Each version has its own project. The problematic NuGet is Microsoft.VSSDK.BuildTools NuGet.
16.11.65 -> 16.11.67
17.3.2093 -> 17.3.2094

_No response_

image

_No response_

deivid-rodriguez:
Fixes #3750.
Fixes #5633.

### DIFF
--- a/common/lib/dependabot/file_parsers/base/dependency_set.rb
+++ b/common/lib/dependabot/file_parsers/base/dependency_set.rb
@@ -117,19 +117,10 @@ module Dependabot
 
           # Produces a new dependency by merging the attributes of `old_dep` with those of
           # `new_dep`. Requirements and subdependency metadata will be combined and deduped.
-          # The version of the combined dependency is determined by the logic below.
+          # The version of the combined dependency is determined by the
+          # `#combined_version` method below.
           def combined_dependency(old_dep, new_dep)
-            version = if old_dep.top_level? # Prefer a direct dependency over a transitive one
-                        old_dep.version || new_dep.version
-                      elsif !version_class.correct?(new_dep.version)
-                        old_dep.version
-                      elsif !version_class.correct?(old_dep.version)
-                        new_dep.version
-                      elsif version_class.new(new_dep.version) > version_class.new(old_dep.version)
-                        old_dep.version
-                      else
-                        new_dep.version
-                      end
+            version = combined_version(old_dep, new_dep)
             requirements = (old_dep.requirements + new_dep.requirements).uniq
             subdependency_metadata = (
               (old_dep.subdependency_metadata || []) +
@@ -143,6 +134,22 @@ module Dependabot
               package_manager: old_dep.package_manager,
               subdependency_metadata: subdependency_metadata
             )
+          end
+
+          def combined_version(old_dep, new_dep)
+            if old_dep.version.nil? ^ new_dep.version.nil?
+              [old_dep, new_dep].find(&:version).version
+            elsif old_dep.top_level? ^ new_dep.top_level? # Prefer a direct dependency over a transitive one
+              [old_dep, new_dep].find(&:top_level?).version
+            elsif !version_class.correct?(new_dep.version)
+              old_dep.version
+            elsif !version_class.correct?(old_dep.version)
+              new_dep.version
+            elsif version_class.new(new_dep.version) > version_class.new(old_dep.version)
+              old_dep.version
+            else
+              new_dep.version
+            end
           end
 
           def version_class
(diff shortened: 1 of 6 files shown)


deivid-rodriguez:
Interesting.

If I understand correctly, when the same dependency is defined in multiple places, the intention is to choose the lowest version so that nothing is misinterpreted as up to date.

And that's what I think it's the culprit for #5633 and what I'm trying to fix here.

However, there are broken specs on several ecosystems, but I'm not really convinced they signal an issue with my fix.

If someone else can have a look, I appreciate it!

deivid-rodriguez:
I rebased this PR after noticing yet another issue it would fix.

In https://github.com/dependabot/dependabot-core/issues/6336, Dependabot confusingly seems to be trying to upgrade the base Docker image to `python:3.11` even if it's already at `python:3.11`. The update is actually correct thought, because there's a secondary Dockerfile in the same directory that's actually pinned to `python:3.9`. So Dependanbot fails to detect what the previous version is. With this PR, it properly documents the update as from `python:3.9` to `python:3.11`. This is a separate problem from the crash presented in that PR, but wanted to notice it.

I'm going to have another look at the failing tests to confirm that the issue is with the tests asserting the wrong thing and not with the implementation proposed here.

deivid-rodriguez:
This PR would also partially fix #6700.

deivid-rodriguez:
I worked through the test failures and they all looked as "bad expectations" to me, so I went ahead and updated the tests. I will add inline comments on the updated tests to elaborate.

mctofu (approved):
Makes sense! We weren't expecting a top level dep to have multiple versions but it can happen for ecosystems where we process top level deps from multiple files at once.

deivid-rodriguez:
Thanks for the review @mctofu!

PR #6084, opened 2022-11-08
Don't try to update path dependencies

### LINKED ISSUE #5989: Ignore path-based gems

### Feature description

In a monorepo, you can have a number of gems that are linked to an app via symlink.

Since these gems are effectively unversioned, checking for updates is a waste of time. Given a large enough number of these, checking for updates on these can lead to Dependabot timing out and not getting to more versioned gems.

This can be improved by manually adding each gem to the ignore file, but ideally the code would be smart enough to figure out it doesn't need to waste time checking these gems.

deivid-rodriguez:
Since by definition they're at the latest version.

There were actually specs positively checking for this, but I believe they may have been an oversight. They were introduced when adding support for updating subdpendencies at
b8010392313fa97dbbdb74792e3fafc7deb35d67, and I don't think this change regresses there.

I'm not sure if this could have any bad side effect, just opening it as a draft.

Fixes #5989.

### DIFF
--- a/bundler/helpers/v1/lib/functions/file_parser.rb
+++ b/bundler/helpers/v1/lib/functions/file_parser.rb
@@ -13,7 +13,7 @@ module Functions
     def parsed_gemfile(gemfile_name:)
       Bundler::Definition.build(gemfile_name, nil, {}).
         dependencies.select(&:current_platform?).
-        reject { |dep| dep.source.is_a?(Bundler::Source::Gemspec) }.
+        reject { |dep| local_sources.include?(dep.source.class) }.
         map { |dep| serialize_bundler_dependency(dep) }
     end
 
@@ -103,9 +103,15 @@ module Functions
         NilClass,
         Bundler::Source::Rubygems,
         Bundler::Source::Git,
-        Bundler::Source::Path,
-        Bundler::Source::Gemspec,
+        *local_sources,
         Bundler::Source::Metadata
+      ]
+    end
+
+    def local_sources
+      [
+        Bundler::Source::Path,
+        Bundler::Source::Gemspec
       ]
     end
   end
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
CI seems fine with it? I'll set it as ready.

jeffwidman (commented):
So the code looks fine, but a question on the behavioral change:
I realize path deps by definition are local imports so no need to import... but I also recall my confusion in https://github.com/dependabot/dependabot-core/pull/5858#discussion_r990817179 where I was unfamiliar with path deps.

1. Before this change, does Dependabot try to update the version listed in the path deps to whatever the latest tagged version is?
2. And then after will that not happen?

Just wondering if the current behavior, even though functionally a no-op due to the local import, still results in less dev confusion for those unfamiliar with local path deps??

Even if above is true, it still may be worth doing this change to avoid useless network I/O etc but wanted to raise the question.

deivid-rodriguez:
Good question. Maybe Dependabot had not updated our lockfile to get in sync with our local gems because of this exclusion?

I believe this change just brings path sourced gems behavior inline with gemspec sourced gems.

In Bundler, the only difference between `Source::Path` and `Source::Gemspec` is that `Source::Path` is created through the `gem "<foo>", :path => "<path>"` Gemfile directive, while `Source::Gemspec` is created through the `gemspec` Gemfile directive. But they are functionally the same thing. However, only gems from `Source::Gemspec` were currently excluded from being checked for updates, and I think that was just an oversight.

But if checking local gems can actually help bringing lockfiles back in sync with the local versions of their path gems maybe it's worth keeping.

I'll mark this as a draft and eventually give it another look.

deivid-rodriguez:
> 1. Before this change, does Dependabot try to update the version listed in the path deps to whatever the latest tagged version is?
> 2. And then after will that not happen?

According to my testing, this doesn't happen before or after this change, so we should be good here. Setting this back to ready.

jeffwidman (approved):
This looks super straightforward :shipit:

PR #6094, opened 2022-11-09
Add `values-other.yaml` & `other_values.yaml` to test cases

jeffwidman:
Noticed this while debugging a related issue. We should test all variants here.

### DIFF
--- a/docker/spec/dependabot/docker/file_fetcher_spec.rb
+++ b/docker/spec/dependabot/docker/file_fetcher_spec.rb
@@ -273,8 +273,12 @@ RSpec.describe Dependabot::Docker::FileFetcher do
       matching_filenames = [
         "other-values.yml",
         "other-values.yaml",
+        "other_values.yml",
+        "other_values.yaml",
         "values.yml",
         "values.yaml",
+        "values-other.yml",
+        "values-other.yaml",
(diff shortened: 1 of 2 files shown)


docker/spec/dependabot/docker/file_fetcher_spec.rb
```diff
@@ -273,8 +273,10 @@
 
       matching_filenames = [
         "other-values.yaml",
+        "other_values.yaml",
         "values.yaml",
         "values2.yaml",
+        "values-other.yaml",
```
jeffwidman, line 279: @brendandburns since you originally wrote these test cases here: https://github.com/dependabot/dependabot-core/commit/7046a3c6668912d64eaf9d1f8c44f5a0ac05f4dd#diff-f117121fabe9e441088126488f34fa0b148aced3b0b99393850768cc6d9534eeR321

Was there a reason you didn't test these two variants as well?

I'm chasing a regex-related bug so wanted to doublecheck whether this was intentional omission or not...

jeffwidman: After digging some more, pretty sure this wasn't an intentional omission, so safe to add these.

bdragon (approved):

PR #6100, opened 2022-11-10
Remove unused mappings

deivid-rodriguez:
The `omnibus/spec` folder does not exist.


jakecoffman (approved):

jeffwidman (approved):
Merging, as this is dev/test only no need to deploy/wait for deploy freeze to lift.

jeffwidman:
Actually, on second thought, is this causing any problems / errors / warnings?

Because if we do ever add a `spec` folder, fairly likely no one will remember these need to be added here until they debug some obscure issue with their dev environment. So if there's no warnings, maybe better to just let it float for now?

deivid-rodriguez:
Yep, it was causing errors for me when running rubocop from the root of the development container.

deivid-rodriguez:
I would copy the error but I can't reproduce it anymore :/ It was happening while rubocop was discovering the list of files it needs to analyze, and it was complaining about the missing `omnibus/spec` path.

I think it's unlikely that we'll add specs to omnibus, and if we did, the development container should give a clear error if the mapping was missing.

jeffwidman:
Perfect... as long as we'd get a clear error if we needed it, then we're good.

PR #6101, opened 2022-11-10
NPM: infer the global registry based on package-lock resolved property

jakecoffman:
It is common in the NPM ecosystem to have a global .npmrc not checked into the repository which defines a private registry. When this happens Dependabot calls the public registry even though it can be inferred that there exists an .npmrc pointing to this private registry.

This PR adds a an inferred .npmrc with a global registry when there exists a package-lock.json containing resolved properties that are all from the same host. This makes Dependabot behave correctly as if an .npmrc had been checked into the repository.

I'm putting this up as a draft, wanted to get some feedback before writing a test since this is a little unusual of a solution.


pavera (commented):
I think this seems sane. Should we expand this support to yarn.lock as well for `.yarnrc` and `.yarnrc.yml`? 

For yarn classic it should be the same resolved field in yarn.lock, for yarn berry it's slightly different but: https://github.com/dependabot/dependabot-core/blob/main/npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser.rb#L252. 

Since yarn classic respects .npmrc I think you'd just need to get the data from yarn.lock instead of package_lock and populate .npmrc, for yarn berry you'd need to update provide a .yarnrc.yml.

jakecoffman:
@pavera Yeah we should cover the Yarn situation as well. I'll follow up this PR with more PRs to implement each of those cases. Thanks for the info on Yarn, I mostly know NPM so that's very helpful!

honeyankit:
> It is common in the NPM ecosystem to have a global .npmrc not checked into the repository which defines a private registry. When this happens Dependabot calls the public registry even though it can be inferred that there exists an .npmrc pointing to this private registry.

Based on the above comment, I think this pr make sense.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -9,6 +10,7 @@
 
 module Dependabot
   module NpmAndYarn
+    # rubocop:disable Metrics/ClassLength
```
jakecoffman, line 13: I'll follow this up with PRs for Yarn so I'll break out a new class to get rid of this.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -65,9 +67,42 @@ def fetch_files
...
 
+      # If every entry in the lockfile uses the same registry, we can infer
+      # that there is a global .npmrc file, so add it here as if it were in the repo.
+      def inferred_npmrc
+        return @inferred_npmrc if defined?(@inferred_npmrc)
+        return @inferred_npmrc = nil unless npmrc.nil? && package_lock
+
+        known_registries = []
+        JSON.parse(package_lock.content).fetch("dependencies", {}).each do |_name, details|
+          resolved = details.fetch("resolved", "https://registry.npmjs.org")
```
honeyankit, line 83: I liked the logic of resolved!

honeyankit (approved):

Israphel:
according to my logs, dependabot is infering the wrong registry.

It should be something like:

https://company.jfrog.io/artifactory/api/npm/repo/

What it gets:

https://company.jfrog.io

jeffwidman:
@Israphel comments on merged PR's generally aren't monitored, you'll need to open a new issue, and feel free to link to this PR in your issue.

PR #6104, opened 2022-11-10
Bump pipenv to 2022.11.5

### LINKED ISSUE #6091: Dependabot flip-flop with python_full_version/python_version

### Package ecosystem

pip, pipenv

### Package manager version

_No response_

### Language version

_No response_

_No response_

### dependabot.yml content

_No response_

### Updated dependency

_No response_

It's becoming ridiculous already how dependabot fight with itself changing `python_version` to `python_full_version` and vice versa:

- https://github.com/Taxel/PlexTraktSync/commits/69dc149838db5c4f3cc36d886ad827ce9988a725/requirements.txt
  - `python_full_version` to `python_version`: https://github.com/Taxel/PlexTraktSync/commit/69dc149838db5c4f3cc36d886ad827ce9988a725
  - `python_version` to `python_full_version` https://github.com/Taxel/PlexTraktSync/commit/dc86ef1a3c277804393d92440f4e1354c61ca0e7 
  - `python_version < '4'` to `python_version < '4.0'`, drops `python_version >= '3.6'` https://github.com/Taxel/PlexTraktSync/commit/31516d211b291331a6b4786bddc95e5e1242f6ba
  - drops `python_full_version < '4.0.0'`, `python_version >= '3.6'` https://github.com/Taxel/PlexTraktSync/commit/bd3501dae2f631710ad9e29ccc3cbba66d1d6eb8

....and so on!

and typically the changes are not even for packages that are being bumped.

the same changes are visible in `Pipfile.lock` in same commits as well.

_No response_

_No response_

_No response_

deivid-rodriguez:
I think this PR would fix #6091 if it was acceptable, because it includes https://github.com/pypa/pipenv/pull/5373 which I believe fixes exactly the issue reported at #6091.

However, it seems to introduce several other noise in manifest file diffs, although at least it's deterministic.

In addition to this, I _think_ it no longer works with Python 3.6 according to https://github.com/pypa/pipenv/issues/5406, so I guess we can't bump it just yet.

Fixes #6091.

### DIFF
--- a/python/helpers/requirements.txt
+++ b/python/helpers/requirements.txt
@@ -2,7 +2,7 @@ pip>=21.3.1,<22.4.0  # Range maintains py36 support TODO: Review python 3.6 supp
 pip-tools>=6.4.0,<6.9.1  # Range maintains py36 support TODO: Review python 3.6 support in April 2023 (eol ubuntu 18.04)
 flake8==5.0.4
 hashin==0.17.0
-pipenv==2022.4.8
+pipenv==2022.11.5
 pipfile==0.0.2
 poetry>=1.1.15,<1.3.0
 wheel==0.37.1
--- a/python/lib/dependabot/python/file_updater/pipfile_file_updater.rb
+++ b/python/lib/dependabot/python/file_updater/pipfile_file_updater.rb
@@ -246,7 +246,7 @@ module Dependabot
         def run_command(command, env: {})
           start = Time.now
           command = SharedHelpers.escape_command(command)
-          stdout, process = Open3.capture2e(env, command)
+          stdout, _, process = Open3.capture3(env, command)
           time_taken = Time.now - start
 
           # Raise an error with the output from the shell session if Pipenv
--- a/python/lib/dependabot/python/file_updater/pipfile_file_updater.rb
+++ b/python/lib/dependabot/python/file_updater/pipfile_file_updater.rb
@@ -233,12 +233,12 @@ module Dependabot
 
         def generate_updated_requirements_files
           req_content = run_pipenv_command(
-            "pyenv exec pipenv lock -r"
+            "pyenv exec pipenv requirements"
           )
           File.write("req.txt", req_content)
 
           dev_req_content = run_pipenv_command(
-            "pyenv exec pipenv lock -r -d"
+            "pyenv exec pipenv requirements --dev"
           )
           File.write("dev-req.txt", dev_req_content)
         end


python/helpers/requirements.txt
```diff
@@ -2,7 +2,7 @@ pip>=21.3.1,<22.4.0  # Range maintains py36 support TODO: Review python 3.6 supp
 pip-tools>=6.4.0,<6.9.1  # Range maintains py36 support TODO: Review python 3.6 support in April 2023 (eol ubuntu 18.04)
 flake8==5.0.4
 hashin==0.17.0
-pipenv==2022.4.8
+pipenv==2022.11.5
```
pavera, line 5: To preserve python 3.6 support you can try:
```suggestion
pipenv>=2022.4.8,<2022.11.11
```

pavera: It appears we've got a good number of tests that fail using the latest pipenv, and the change to the lockfile format could be problematic, we'll almost have to match pipenv version with the project/repo I think to avoid these flip/flop issues.

glensc: why would you need to match python version of the project? just run your pipenv with latest python version only?

matteius: @glensc I think the issue is with markers and resolution right, because if your project is on say python 3.7 and a package you depend on drops support for it, then a lock file generated by python 3.11 is going to be much different than the one generated by 3.7.

matteius:
There were only a few validation constraints that mattered when we switched from Pipfile to plette -- from the test failures I can see one of them identified:  the sources have to have a `name` attribute now (not sure why they could ever be unnamed before, it was probably an oversight) -- anyway if there is anything I can do or questions I can help answer to help you get to a reasonable state with pipenv please let me know.

python/helpers/requirements.txt
glensc: perhaps install different versions based on python version then?

for python 3.7 install whatever is last pipenv is installable there, and for newer python go on with current version

glensc:
so what's the plan here? drop supporting older python? create fork and make pipenv that's compatible with your python versions? use different versions as noted [here](https://github.com/dependabot/dependabot-core/pull/6104#discussion_r1054358137)?

had to fix yet another bug caused by the ecosystem:
- https://github.com/Taxel/PlexTraktSync/commit/9ecd9d71532b2eafec551b16f6e652a795a91e43

jeffwidman:
> so what's the plan here?

We're planning to drop support for python 3.6 which will unblock this:
* https://github.com/dependabot/dependabot-core/issues/6299

Timing is still TBD, see that issue for details.

python/lib/dependabot/python/file_updater/pipfile_file_updater.rb
```diff
@@ -233,20 +233,20 @@ def post_process_lockfile(updated_lockfile_content)
...
+            "pyenv exec pipenv requirements --dev"
           )
           File.write("dev-req.txt", dev_req_content)
         end
 
         def run_command(command, env: {})
           start = Time.now
           command = SharedHelpers.escape_command(command)
-          stdout, process = Open3.capture2e(env, command)
+          stdout, _, process = Open3.capture3(env, command)
```
jeffwidman, line 249: This was extracted out separately into:
* https://github.com/dependabot/dependabot-core/pull/6103

python/lib/dependabot/python/file_updater/pipfile_file_updater.rb
```diff
@@ -233,20 +233,20 @@ def post_process_lockfile(updated_lockfile_content)
 
         def generate_updated_requirements_files
           req_content = run_pipenv_command(
-            "pyenv exec pipenv lock -r"
+            "pyenv exec pipenv requirements"
```
jeffwidman, line 236: I think we'll instead want to long-term go with:
* https://github.com/dependabot/dependabot-core/issues/6836

python/helpers/requirements.txt
jeffwidman: We're dropping python 3.7 support on August 17th:
* #7702

jeffwidman:
I'm going to close this PR in favor of:
* https://github.com/dependabot/dependabot-core/pull/7715

I suspect we'll hit the same test failures, but we can work through them over there... and may end up migrating to the new `upgrade` command as part of that:
* https://github.com/dependabot/dependabot-core/issues/6836

PR #6097, opened 2022-11-09
Update helm charts beyond `values.yaml`

jeffwidman:
The docker updater for helm charts was hardcoded to `"values.yaml"`.

But we should support the full range of filenames supported in the file fetcher... for example, in the wild we saw this failing to update `"arc-values.yaml"` because of this hardcoding.

PR #6105, opened 2022-11-10
Specs for helmfiles with different names

jeffwidman:
Unit tests for #6097.

Broke out as a separate PR because I'm still learning the Rspec DSL.

### DIFF
--- a/docker/spec/dependabot/docker/file_updater_spec.rb
+++ b/docker/spec/dependabot/docker/file_updater_spec.rb
@@ -1139,7 +1139,7 @@ RSpec.describe Dependabot::Docker::FileUpdater do
   let(:helmfile) do
     Dependabot::DependencyFile.new(
       content: helmfile_body,
-      name: "values.yaml"
+      name: "values.yaml" # TODO not sure if I need to replace the filenames here or if I can override later in a specific test
     )
   end
   let(:helmfile_body) { fixture("helm", "yaml", "values.yaml") }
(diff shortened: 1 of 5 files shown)


docker/spec/dependabot/docker/file_updater_spec.rb
```diff
@@ -1139,7 +1139,7 @@
   let(:helmfile) do
     Dependabot::DependencyFile.new(
       content: helmfile_body,
-      name: "values.yaml"
+      name: "values.yaml" # TODO not sure if I need to replace the filenames here or if I can override later in a specific test
```
jeffwidman, line 1142: note to self: this fixture seems to be for all the helm chart tests, so will likely need to refactor things a bit... but no need to fully test that the other filename permutations have exhaustive descriptions as that is the kind of thing that will slow down CI over time... just ensure some basic sanity checking to ensure the updater is picking it up, and then rely on the existing tests to verify that the updater is doing the right thing once it picks it up...

docker/spec/dependabot/docker/utils/helpers_spec.rb
```diff
@@ -0,0 +1,33 @@
+# # frozen_string_literal: true
+
+require "dependabot/dependency_file"
+require "dependabot/docker/utils/helpers"
+require "spec_helper"
+
+RSpec.describe Dependabot::Docker::Utils.likely_helm_chart? do
```
jeffwidman, line 7: Since this is a static method on the class, not sure that this is the proper way to call it? Writing this was a reminder I _really_ need to spend some time on basic ruby syntax tutorials... 😀

PR #6106, opened 2022-11-11
Use correct package manager name

jeffwidman:
These tests were originally written by copy/pasting the `"docker"` ecosystem into an entirely new ecosystem called `"kubernetes"`. However, given the amount of duplication between them, we worked with the original contributor to instead extend the `"docker"` ecosystem itself.

It looks like these tests were never updated from the original `"kubernetes"` package manager name back to `"docker"`, so this fixes that.


deivid-rodriguez (approved):

deivid-rodriguez:
🤣 Perfect sync!

PR #6111, opened 2022-11-11
Fix NuGet packages.config parsing

brandonaut:
Since pattern whitespace is being ignored via the `x` flag, the string `<packages>` incorrectly matches the second part of the regex (`<package\s[^>]*?[^/]>.*?</package>`) when "package" isn't a self-closing element.

For example:

```xml
<packages>
  <package id="mypackage" version="0.0.0"></package>
</packages>
```

will match:

```
<packages>
  <package id="mypackage" version="0.0.0"></package>
```

This commit makes the space explicit so that `<packages>` no longer matches.

### DIFF
--- a/nuget/spec/fixtures/packages_configs/packages.config
+++ b/nuget/spec/fixtures/packages_configs/packages.config
@@ -9,5 +9,5 @@
   <package id="NuGet.Server" version="2.11.2" targetFramework="net46" />
   <package id="RouteMagic" version="1.3" targetFramework="net46" />
   <package id="WebActivatorEx" version="2.1.0" targetFramework="net46" />
+  <package id="NUnit" version="3.13.0" targetFramework="net46"></package>
 </packages>
-
--- a/nuget/spec/fixtures/packages_configs/packages.config
+++ b/nuget/spec/fixtures/packages_configs/packages.config
@@ -10,3 +10,4 @@
   <package id="RouteMagic" version="1.3" targetFramework="net46" />
   <package id="WebActivatorEx" version="2.1.0" targetFramework="net46"></package>
 </packages>
+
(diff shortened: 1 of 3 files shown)


jurre:
Makes sense to me, thanks! Would you mind adding a test to verify the new behavior and to make sure we don't regress in the future?

brandonaut:
Working on it! I'm not very familiar with Ruby and RSpec, so it's taking me a bit...

jeffwidman:
I approved the CI run, let us know when it's ready for another round of review or if you need CI runs approved again.

brandonaut:
I'm ready for another CI run. Sorry, I'm not set up to run the tests locally :(

jeffwidman:
No problem at all, just approved. I also for now changed our CI rules so that every run from a first-time contributor doesn't require pre-approval, but only if they're new to GitHub. That should make your life easier because you won't have to wait for us to approve anymore, sorry for the hassle.

Also for running tests locally it's actually super simple since you can do it directly within docker... see https://github.com/dependabot/dependabot-core#running-with-docker and then also see the "tests" section slightly down. Within a failing test you can drop a `binding.b` at any point to drop yourself into a debugger.

Holler if you have questions.

nuget/spec/fixtures/packages_configs/packages.config
```diff
@@ -8,6 +8,6 @@
   <package id="NuGet.Core" version="2.11.1" targetFramework="net46" />
   <package id="NuGet.Server" version="2.11.2" targetFramework="net46" />
   <package id="RouteMagic" version="1.3" targetFramework="net46" />
-  <package id="WebActivatorEx" version="2.1.0" targetFramework="net46" />
+  <package id="WebActivatorEx" version="2.1.0" targetFramework="net46"></package>
```
jeffwidman, line 11: I assume this test fails w/o this patch?

jeffwidman (approved):
👍 nice job thank you!

nuget/spec/fixtures/packages_configs/packages.config
brandonaut: Correct

brandonaut: I got the tests running locally and was able to confirm :)

PR #6112, opened 2022-11-12
Log `git clone` output when we fail to find base commit sha

deivid-rodriguez:


jurre (approved):

jeffwidman (approved):
This looks harmless, but I'm curious what the motivation was here? Is it just so that when things go sideways we have greater observability or is there a specific error you're trying to track down?

deivid-rodriguez:
This should no longer be necessary.

PR #6113, opened 2022-11-13
Update licensed cache

deivid-rodriguez:
I'm not sure whether it makes sense for us to keep using licensed, given that the updater is fully open source.

If yes, this PR would still need to add licensed to a CI step, probably the recently added `dependency-review.yml` would be a good place?

If no, this PR just needs to be amended to delete all the things.


jeffwidman:
There's some internal discussion related to this, I'll cross link it to you privately so you have context.

deivid-rodriguez:
I understand that removing the license cache needs some discussion, but fixing the existing license check should be a no-brainer, so setting this as ready!

jeffwidman (approved):

PR #6114, opened 2022-11-13
Remove `faraday-retry` warning

deivid-rodriguez:
This warning is printed every time dependabot is required.

To fix it, I think we can actually add the gem and make sure it's used by Octokit internally, instead of rolling our own implementation.


landongrindheim (approved):

deivid-rodriguez:
Yeah, in my opinion, this warning is fundamentally wrong because there's no way out of it if you don't want to use retries. And all warnings should be fixable.

Luckily we already use retries so I think it does make sense for us to use the extracted gem.

PR #6127, opened 2022-11-14
Tolerate running `dry-run.rb` without `docker-dev-shell`

deivid-rodriguez:
I actually run `dry-run.rb` without `docker-dev-shell` quite often, specifically in those ecosystem where I know I have everything I need, or when I'm confident I can troubleshoot any issues.

With the current situation I have to keep stashing and unstashing a patch to remove this whole block when I need to. Instead I could live with a warning every time I run this, but that allows me to avoid all this dance.

I also think this is less important now that our main entry point to reproduce issues is the [CLI](https://github.com/dependabot/cli).

Thoughts?

### DIFF
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -37,7 +37,7 @@
 # rubocop:disable Style/GlobalVars
 
 require "etc"
-unless Etc.getpwuid(Process.uid).name == "dependabot"
+unless Etc.getpwuid(Process.uid).name == "dependabot" || ENV["ALLOW_DRY_RUN_STANDALONE"] == "true"
   puts <<~INFO
     bin/dry-run.rb is only supported in a development container.
 


honeyankit (approved):

Nishnha (approved):
Makes sense to me

honeyankit:
> I also think this is less important now that our main entry point to reproduce issues is the [CLI](https://github.com/dependabot/cli).

I think it is still worth having dry-run script as well and any improvement to it good. Once we will completely switch to CLI then we can obsolete dry-run.

deivid-rodriguez:
> I think it is still worth having dry-run script as well and any improvement to it good. Once we will completely switch to CLI then we can obsolete dry-run.

Totally agreed! I'm not even sure I want to sunset it at all, it's useful when you want to test the update logic in isolation and don't need the "full system" to run.

bin/dry-run.rb
```diff
@@ -39,11 +39,10 @@
 require "etc"
 unless Etc.getpwuid(Process.uid).name == "dependabot"
   puts <<~INFO
-    bin/dry-run.rb is only supported in a development container.
+    bin/dry-run.rb is best supported in a development container.
 
-    Please use bin/docker-dev-shell first.
+    Consider using bin/docker-dev-shell first.
```
jeffwidman, line 44:
```suggestion
    bin/dry-run.rb is only supported in a development container.
    
    Please use bin/docker-dev-shell first unless you have a specific reason not to.
```

jeffwidman (commented):
I'm +1 to making this a warning rather than a hard `exit`, but given that this warning is targeting users who lack context on `dry-run` and `docker-dev-shell`, I suggest tweaking it...

We don't need someone to start using it and then complain about some hiccup they hit saying "it's best supported, not only supported"... 😀

jeffwidman:
Also, I like having a hook at a lower level than the CLI, the fewer layers in the way the faster the iteration speed if you know exactly which section you want to work on.

My preference is eventually merge dry-run script and updater and possibly even dependabot-script so that there's a single main entrypoint that offers a dry-run flag... how we do that requires some thought from a design standpoint to handle the fact that the payload that we pass to the internal dependabot API is different than when it's running as a library creating PR's without the API. But IMO that's all a different discussion for a different time.

bin/dry-run.rb
deivid-rodriguez: May I suggest an ENV instead that communicates the same thing but also allows me to run this without warnings? Something like `FORCE_UNSUPPORTED_USE_OF_DRY_RUN_WITHOUT_DOCKER_DEV_SHELL`?

deivid-rodriguez: So, keep the current wording and keep this a hard error, but skip it if this ENV is set?

jeffwidman: Yeah, that's probably a better move so that someone has to explicitly opt into bypassing the `exit`.

An env var or a CLI arg would be fine too... I'm not too picky, whatever is smoothest for your workflow... Although a shorter env var name might be better 😆 Maybe something like `ALLOW_DRY_RUN_STANDALONE`? Or a cli arg of `--allow-standalone`... up to you.

deivid-rodriguez: For my workflow ENV var works best, because it's silent and I can set it permanently 👍.

Now onto naming. My verbose proposal was precisely to accommodate your suggestion of not giving the impression that this running mode is supported in any way. But if you think `ALLOW_DRY_RUN_STANDALONE` is fine, that works for me!

I think a flag does give the impression of support so I would avoid that in favor of an ENV variable.

deivid-rodriguez: I updated it to use `ALLOW_DRY_RUN_STANDALONE`.

jeffwidman (approved):

PR #6128, opened 2022-11-14
Remove unused `appdir` arg

jeffwidman:
These args appear unused. `git blame` shows they landed as part of the merge of updater-into-core, so possibly they're a relic from that.

### DIFF
--- a/script/_common
+++ b/script/_common
@@ -16,8 +16,6 @@ function docker_build() {
 
 function docker_exec() {
   docker_build
-
-  appdir=$(cd $(dirname "$0")/../updater && pwd)
   docker run --env "DEPENDABOT_TEST_ACCESS_TOKEN=$DEPENDABOT_TEST_ACCESS_TOKEN" \
   --rm \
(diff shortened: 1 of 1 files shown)


script/_common
```diff
@@ -16,8 +16,6 @@ function docker_build() {
 
 function docker_exec() {
   docker_build
-
-  appdir=$(cd $(dirname "$0")/../updater && pwd)
```
jeffwidman, line 20: I have a funny feeling that setting the `appdir` var isn't needed, but that the `cd $(dirname "$0")/../updater` is... 

Let's see what CI thinks, although that probably doesn't cover all workflows that use this...

deivid-rodriguez: My understand is that this is to support running this script from outside of the `updater/` folder. Not sure if we need that.

deivid-rodriguez (approved):
👍 Nice find!

script/_common
jeffwidman: ah right, I missed that!

deivid-rodriguez (approved):

PR #6129, opened 2022-11-14
Bump `Dart` to `2.18.4`

### LINKED ISSUE #6126: Install Dart 2.18

https://github.com/dependabot/dependabot-core/blob/235cd665d940bb1cb200bc2cabf8ec3feaf1c54b/Dockerfile#L261

Dart 2.18 dropped a bit ago, we should bump to it.

jeffwidman:
Bump to the latest release. I started to add the SHA256 validation, but it'll require refactoring how the unzipping works, so leaving that for later as the version bump is a step in the right direction for now.

Fix #6126

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -258,7 +258,9 @@ ENV PUB_CACHE=/opt/dart/pub-cache \
   PUB_ENVIRONMENT="dependabot" \
   PATH="${PATH}:/opt/dart/dart-sdk/bin"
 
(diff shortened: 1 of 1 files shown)


jeffwidman:
cc @sigurdm @jonasfj I assume no currently known blocker bugs in the `2.18.4` release that should prevent merging this?

mctofu (approved):

PR #6135, opened 2022-11-15
Make specs time zone independent

deivid-rodriguez:
I was being a bad boy and running `common/` specs outside of the development container and got a couple of failures.

Failures were due to my specific timezone.

This PR makes specs timezone independent.


jurre (approved):
I wish we had `Time.zone.now` instead, but then we'd have to pull in the ActiveSupport extensions etc and seems not worth it 👍

PR #6137, opened 2022-11-15
test: add test for private registry module with dir suffix

stigok:
I am suspecting that module sources containing a dir suffix is not properly handled. Unable to test locally due to the size of the Docker developer image.

### DIFF
--- a/terraform/spec/fixtures/projects/private_module_with_dir_suffix/main.tf
+++ b/terraform/spec/fixtures/projects/private_module_with_dir_suffix/main.tf
@@ -1,0 +1,4 @@
+module "my-module" {
+  source  = "registry.example.com/org/name/provider//sub/dir/module"
+  version = "1.2.3"
+}
--- a/terraform/spec/dependabot/terraform/file_parser_spec.rb
+++ b/terraform/spec/dependabot/terraform/file_parser_spec.rb
@@ -871,6 +871,34 @@ RSpec.describe Dependabot::Terraform::FileParser do
       end
     end
 
+    context "with a private module with directory suffix" do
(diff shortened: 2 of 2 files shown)


stigok:
Thanks! I've fixed my lint errors now. Could you please take a look at my test to see if it I am testing what I think I'm testing? I am trying to see if dependabot actually ends up with the actual module and registry URL even when the `source = "<src>"` contains a `//sub/dir/in/module` suffix.

```terraform
module "foo" {
    source = "registry.example.com/my/module/generic//sub-dir"
    version = "1.2.3"
}
```

stigok:
This actually seems to work out fine. We may have some other problems in our setup outside of Dependabot. Feel free to merge this if you care about this extra test.

terraform/spec/fixtures/projects/private_module_with_dir_suffix/main.tf
```diff
@@ -0,0 +1,4 @@
+module "my-module" {
+  source  = "registry.example.com/org/name/provider//sub/dir/module"
```
jeffwidman, line 2: I don't know `terraform` very well... is the double slash `//` in this path expected/desired?

jeffwidman (commented):
Always happy to add more tests, had a quick question though before merging...

terraform/spec/fixtures/projects/private_module_with_dir_suffix/main.tf
stigok: Yeah, this is to select a subfolder within the module source. https://developer.hashicorp.com/terraform/language/modules/sources#modules-in-package-sub-directories

jeffwidman (approved):

PR #6138, opened 2022-11-15
NPM: reject changes to resolved in package-lock

jakecoffman:
Dependabot should never change the resolved property in `package-lock.json`. If it does this is an indicator of a bug or a mistake in configuration, like a missing .npmrc. In this case it's better to fail the dependency's update rather than put up a PR to avoid name-squatting or other attacks. 

So here I've parsed the pre-update and post-update lockfile to try to see if the resolved property has changed. If so it throws a recognizable error. In the case of a bug or something we can handle this will make it easier to track down those cases. 

Setting this PR as Draft again to get early feedback on this approach.


jakecoffman:
This is a bit of a can of worms, closing for now.

Issue #6081, opened 2022-11-08
Yarn v2/v3 zero-install file permissions are changed


agrobbin:
When doing zero-installs on Yarn v2/v3, the `.pnp.cjs` file permission is changed by Dependabot from `755` to `644`.

@pavera to ensure this subtle bug is not lost, I thought it'd make sense to create a separate issue for it.

_Originally posted by @agrobbin in https://github.com/dependabot/dependabot-core/issues/5946#issuecomment-1303986157_

pavera:
Some but not all? That would seem to indicate that our detection of zero installs is non-deterministic somehow. If we detect zero-install (by the presence of a .pnp.cjs file in the repo) we use `skip-build` otherwise we do fall back to using the `enableScripts` setting. I'll need to investigate further how this might be failing.

agrobbin:
Sorry, what I meant to say was "in all Dependabot PRs issued this morning". My use of the word "some" was meant to indicate that I got several Dependabot PRs, all of which had this issue. Sorry for the confusion!

pavera:
OK, so `skip-build` doesn't seem to preserve the permissions in that case. I'll have to test more and try to find a solution, thanks for creating this issue so it's not lost.

pavera:
I've found the issue here, we set the mode on all files as part of the [PR creation](https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/pull_request_creator/github.rb#L196). 

Working on a solution now.

pavera:
I've deployed the fix for this and verified it works in our test repos. Closing this issue, but if this is still occurring please reopen.

agrobbin:
Thanks @pavera! I will confirm the next time Dependabot updates a Yarn dependency.

agrobbin:
@pavera just to confirm, this seems to have done the trick! Really appreciate all of your follow-up work on Yarn Berry Dependabot support.

agrobbin:
@pavera I'm sorry to resurrect this issue, but I have started seeing this issue again on one of my repositories. I haven't been able to deduce a pattern, yet, but if there's anything I can do to help debug it, please let me know!

PR #6141, opened 2022-11-15
Yarn Berry: Make a best effort to preserve file permissions

pavera:
## Context
As reported in https://github.com/dependabot/dependabot-core/issues/6081 we are overwriting file permissions with `644`. This causes issues for yarn berry which expects some files to be executable.

## Approach
I've added an instance variable to DependencyFile which will attempt to gather the original mode of the file. We then will use this mode if we were able to successfully gather it, falling back to `644` if we cannot.

## Next Steps
The 2 failing e2e tests are expected as this PR changes the PR creation content. Once this is merged I will fix the tests to expect the new values.


honeyankit (approved):
🚀

common/lib/dependabot/pull_request_creator/github.rb
```diff
@@ -193,7 +193,7 @@ def create_tree
             {
               path: (file.symlink_target ||
                      file.path).sub(%r{^/}, ""),
-              mode: "100644",
+              mode: (file.mode || "100644"),
```
honeyankit, line 196: I was looking at the fallback logic and found it here.

PR #6142, opened 2022-11-15
Raise `DependencyFileNotFound` for missing files

jeffwidman:
`DependencyFileNotFound` is treated by our internal monitoring as a user configuration error, and not something we need to act on. Additionally, we surface a nice clean error banner to the user.

`DependabotError` is treated as an unknown error that requires further investigation, so it's causing our internal alerts to flap when users don't have a manifest file... for example, users sometimes assume the `docker` check is recursive so they specify it as `"/"`, but it's not so it raises this error.

This is a super quick PR that we can get deployed to stop our alerts flapping. I'm working on a larger refactor of this section of code that also removes the feature flag and adds a unit test to ensure the proper exception is raised.

### DIFF
--- a/docker/lib/dependabot/docker/file_fetcher.rb
+++ b/docker/lib/dependabot/docker/file_fetcher.rb
@@ -40,8 +40,9 @@ module Dependabot
           )
         elsif incorrectly_encoded_dockerfiles.none? && incorrectly_encoded_yamlfiles.none?
           raise(
-            Dependabot::DependabotError,
-            "Found neither Kubernetes YAML nor Dockerfiles in #{directory}"
+            Dependabot::DependencyFileNotFound,
+            File.join(directory, "Dockerfile"),
+            "Found neither Dockerfiles nor Kubernetes YAML in #{directory}"
           )
         elsif incorrectly_encoded_dockerfiles.none?
           raise(


honeyankit (approved):

docker/lib/dependabot/docker/file_fetcher.rb
```diff
@@ -40,8 +40,9 @@ def fetch_files
           )
         elsif incorrectly_encoded_dockerfiles.none? && incorrectly_encoded_yamlfiles.none?
           raise(
-            Dependabot::DependabotError,
-            "Found neither Kubernetes YAML nor Dockerfiles in #{directory}"
+            Dependabot::DependencyFileNotFound,
+            File.join(directory, "Dockerfile"),
```
jeffwidman, line 44: I copied this from the implementation above where the feature flag is disabled... however they're both wrong because we look for any file matching the regex `DOCKER_REGEXP = /dockerfile/i`... 

I considered returning just the directory, but I noticed over in https://github.com/dependabot/dependabot-core/blob/b6784e245a6636982a5a66f764cf04cb0d3c2335/common/lib/dependabot/errors.rb#L86 that both `file_name` and `directory` methods assume a file name is tacked onto the path... 

I'm not sure where those methods are called, but I have a hunch they're displayed to the user in the error UI banner that we display... since we have custom handling of this error so that users don't have to grep through the logs to see why Dependabot is failing. So for now I just stuck with what exists and will investigate whether this is used at all later.

Issue #1985, opened 2020-04-30
Missing current version field in GitHub action output


jsoref:
**Package manage/ecosystem**
GitHub Actions
**Manifest contents prior to update**
An outdated version of `actions/checkout`

**Updated dependency**

**What you expected to see, versus what you actually saw**
I saw this:
```
updater | INFO <job_31581840> Checking if actions/checkout  needs updating
  proxy | 2020/04/30 22:26:29 GET https://github.com:443/actions/checkout.git/info/refs?service=git-upload-pack
  proxy | 2020/04/30 22:26:29 * authenticating git server request (host: github.com)
  proxy | 2020/04/30 22:26:29 200 https://github.com:443/actions/checkout.git/info/refs?service=git-upload-pack
updater | INFO <job_31581840> Latest version is 01aecccf739ca6ff86c0539fbc67a7a5007bbc81
updater | INFO <job_31581840> Requirements to unlock own
updater | INFO <job_31581840> Requirements update strategy 
updater | INFO <job_31581840> Updating actions/checkout from  to 01aecccf739ca6ff86c0539fbc67a7a5007bbc81
updater | INFO <job_31581840> Submitting actions/checkout pull request for creation
```
There are two things that strike me as suspicious:

* `if actions/checkout  needs`
* `Updating actions/checkout from  to`

Specifically the two spaces (` ` and `  `) in a row.

I half expect the first one to include some form of a version.
I definitely expect a version between `from` and `to`.

**Images of the diff or a link to the PR, issue or logs**

deivid-rodriguez:
Hi, sorry for the delay.

This is probably working fine now, please double check.

If not, what's the original content of the `actions/checkout` line in the workflow?

jsoref:
(I'm pretty sure I actively lost the identity of the original request's repository, and I'm not sure I could find one that's currently behind for that action...)

Hi @deivid-rodriguez, here's an example from the last 24 hours (I think):
```
updater | INFO <job_511354460> Checking if google-github-actions/auth 0 needs updating
  proxy | 2022/11/14 05:01:37 [066] GET https://github.com:443/google-github-actions/auth.git/info/refs?service=git-upload-pack
  proxy | 2022/11/14 05:01:37 [066] * authenticating git server request (host: github.com)
  proxy | 2022/11/14 05:01:37 [066] 200 https://github.com:443/google-github-actions/auth.git/info/refs?service=git-upload-pack
[...]

PR #6143, opened 2022-11-15
Customize dependency full name

deivid-rodriguez:
This was suggested at https://github.com/dependabot/dependabot-core/issues/1985#issuecomment-1315815054.

Not sure if allowing customizing this per ecosystem is overkill, or if the `name@versio` notation makes sense for every ecosystem, or if it's just best to leave this as is.

Just opening a WIP for discussion.

### DIFF
--- a/common/lib/dependabot/dependency.rb
+++ b/common/lib/dependabot/dependency.rb
@@ -6,6 +6,7 @@ module Dependabot
   class Dependency
     @production_checks = {}
     @display_name_builders = {}
+    @display_full_name_builders = {}
     @name_normalisers = {}
 
     def self.production_check_for_package_manager(package_manager)
@@ -25,6 +26,14 @@ module Dependabot
 
     def self.register_display_name_builder(package_manager, name_builder)
       @display_name_builders[package_manager] = name_builder
+    end
+
+    def self.display_full_name_builder_for_package_manager(package_manager)
+      @display_name_builders[package_manager]
+    end
+
+    def self.register_display_full_name_builder(package_manager, full_name_builder)
+      @display_name_builders[package_manager] = full_name_builder
     end
 
     def self.name_normaliser_for_package_manager(package_manager)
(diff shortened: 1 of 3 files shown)


jsoref:
Personally, I like this. Note that for some other ecosystems, using `:` as the version separator would be a nice improvement.

jsoref:
Oh, it'd need the `v` to be restored as well for actions.

jsoref:
Imagining this wasn't a github action (for which `google-github-actions/auth@v0` would be the preferred output for the first line) and that there isn't some other pretty printer available (although, I expect most ecosystems should provide a pretty printer for at least the first thing), I think this change would be helpful:
```diff
-updater | INFO <job_511354460> Checking if google-github-actions/auth 0 needs updating
-updater | INFO <job_511354460> Updating google-github-actions/auth from 0 to 1
+updater | INFO <job_511354460> Checking if google-github-actions/auth "0" needs updating
+updater | INFO <job_511354460> Updating google-github-actions/auth from "0" to "1"
```

Or something similar (possibly backticks instead of double-quotes).

A stray `0` is just kinda odd, and it's generally helpful to highlight the portions of text that are dynamic...

deivid-rodriguez:
All good points!

How about being a bit more verbose?

> updater | INFO <job_511354460> Checking if google-github-actions/auth at version 0 needs updating
> updater | INFO <job_511354460> Updating google-github-actions/auth from version 0 to version 1

jsoref:
Maybe?

I'd probably go with:
> updater | INFO <job_511354460> Checking if google-github-actions/auth (version 0) needs updating
> updater | INFO <job_511354460> Updating google-github-actions/auth from version 0 to version 1

I can't quite explain why I really don't like `google-github-actions/auth version 0`. I think it's because a version itself doesn't need updates, it's the object `google-github-actions/auth` that needs an update. 

Fwiw, I'm not sure that `Updating` is quite right. I mean, technically a branch is created with an update. But, I'd half argue in favor of `Proposing updating ...` because that's what it'd mean for the average user...

deivid-rodriguez:
@jsoref Honestly, all your suggestions seem sensible and improvements over the current wording to me. And I also think they make sense in all ecosystems. Would you like to open a PR?

I'll close this one since it does feel overkill and not better than what you just proposed.

jsoref:
I'm currently fighting a GitLab pipeline (I just signed up for the importer beta because I'd really rather not have to fight GitLab). And your test failures here w/ Ruby remind me that it isn't my strongest language.

I certainly could, but if you're up for doing it, I'd rather leave it to you.

➡️ 🛌 💤

deivid-rodriguez:
Please don't be scared about test failures here, this was just some code I hacked without running anything so not surprised I broke something 😅.

But no worries, I can take care of it. Good night!

PR #6144, opened 2022-11-15
Remove the `kubernetes_updates` feature flag

jeffwidman:
This has been rolled out 100%, so safe to remove.

### DIFF
--- a/docker/lib/dependabot/docker/file_fetcher.rb
+++ b/docker/lib/dependabot/docker/file_fetcher.rb
@@ -22,27 +22,18 @@ module Dependabot
 
       private
 
-      def kubernetes_enabled?
-        Experiments.enabled?(:kubernetes_updates)
-      end
-
       def fetch_files
         fetched_files = []
         fetched_files += correctly_encoded_dockerfiles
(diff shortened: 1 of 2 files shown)


docker/lib/dependabot/docker/file_fetcher.rb
```diff
@@ -22,27 +22,18 @@ def self.required_files_message
...
 
-      def kubernetes_enabled?
-        Experiments.enabled?(:kubernetes_updates)
-      end
-
       def fetch_files
         fetched_files = []
         fetched_files += correctly_encoded_dockerfiles
-        fetched_files += correctly_encoded_yamlfiles if kubernetes_enabled?
+        fetched_files += correctly_encoded_yamlfiles
```
jeffwidman, line 28: Is there some ruby idiom for why it's instantiated as an empty array rather than `fetched_files = correctly_encoded_dockerfiles + correctly_encoded_yamlfiles`? Is it just trying to guarantee `fetched_files` will never be `nil`?

bdragon: I think this is harmless. I would guess it was stylistic—it accommodates adding in another array on one line without having to reformat.

Issue #4459, opened 2021-11-26
Considering main branch as a default one for bundler


CrazyPit:
**Package ecosystem**

Bundler.

**Package manager version**
2.2.29

**Language version**
Ruby 3.0.2

**Dependabot config**

```yaml
version: 2
registries:
  toptal-github:
    type: "git"
    url: "https://github.com"
    username: "x-access-token"
    password: "${{ secret }}"
updates:
  - package-ecosystem: "bundler"
    directory: "/"
    insecure-external-code-execution: allow
    registries:
      - toptal-github
    schedule:
      interval: "daily"
    labels:
      - "no-jira"
    allow:
      - dependency-type: "all"
  - package-ecosystem: "docker"
    directory: "/"
    schedule:
      interval: "daily"
    labels:
      - "no-jira"
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "daily"
    labels:
      - "no-jira"
```

**Updated dependency**
yard

`gem 'yard', github: 'lsegal/yard'`

**What you expected to see, versus what you actually saw**

I'm getting this message: 
<img width="926" alt="Screenshot 2021-11-24 at 15 01 04" src="https://user-images.githubusercontent.com/2117/143579631-432d1601-2b21-47d4-88a2-f1f4086ee4a5.png">

The issue is that yard is using `main` branch instead of `master` and dependbot is not considering main branch as default. 
I've fixed it by adding `branch: 'main'`. But I think that since main is a default branch name better to update dependabot.

brrygrdn:
I think this may affect git sources for more ecosystems than bundler, I've labelled this for further triage. Just to confirm my read, when you say:

> I've fixed it by adding `branch: 'main'`

You are referring to the Gemfile, i.e.

> gem 'yard', github: 'lsegal/yard', branch: 'main'

CrazyPit:
> You are referring to the Gemfile, i.e.
> gem 'yard', github: 'lsegal/yard', branch: 'main'

correct

ojab:
Related part of dependabot logs
```

  proxy | 2021/11/23 12:46:20 [049] GET https://github.com:443/lsegal/yard.git/info/refs?service=git-upload-pack
  proxy | 2021/11/23 12:46:20 [049] * authenticating git server request (host: github.com)
  proxy | 2021/11/23 12:46:20 [049] 200 https://github.com:443/lsegal/yard.git/info/refs?service=git-upload-pack
[...]

PR #6149, opened 2022-11-16
Remove hardcoded master branch

deivid-rodriguez:
Found this while looking into something else, and I believe it could be the fix for #4459?

I'll run CI to see if this breaks something.


deivid-rodriguez:
@jakecoffman I think this change is correct, but revealed an error in a smoke test, so I believe the smoke test should be updated.

If I'm right and this change is correct, then I should open a PR to https://github.com/dependabot/smoke-tests and then get this one merged right after that other PR?

Just thinking that merging the change to https://github.com/dependabot/smoke-tests will make all other PRs fail until this one is merged, so just wondering how you handled this situation other times.

jakecoffman:
I would update the smoke test first to get the ✅ here. It's not a big deal they're broken in the meantime. Only changes to bundler or the updater would trigger the Action anyway so no one may notice.

deivid-rodriguez:
Ok, I'll wait for a thumbs up here, in case I'm missing something and we should be actually changing this. Once I get approval I'll open a PR to smoke_tests to get the PR to green.

jakecoffman (approved):
This makes sense!

PR #6150, opened 2022-11-16
Don't try to resolve pinned sha's to versions when updating docker images

### LINKED ISSUE #2997: The docker `update_checker` method; `version_of_latest_tag` can be very slow
The 'version_of_latest_tag' method in the docker 'update_checker' implementation is really slow when it doesn't match the 'latest' tag digest in the first few tags.

Currently the implementation downloads a list of tags and tries to match the 'latest' tag's digest to another tag's digest by fetching digests one-by-one for all tags in the docker repo (note it short-circuits if a match if found early). The problem is that a single loop of the fetching a digest can take about a second, because it has to fetch an auth token first. Now couple this with a repo containing hundreds of tags, none of which is maching the digest of 'latest' (like dependabot-core), and you have a process taking tens of minutes.

I'm not sure how to deal with this problem though. I've looked at the Docker API, and searched online, to see if I could find a good solution to get a list of tags for a specific digest from a docker repo, but I found nothing.

So at the very least I wanted to raise a flag, of a potential performance bottleneck in the current implementation. Maybe you guys have some ideas on how to make this fster. I mean, maybe its possible to parallelize fetching the digests or something.

Let me know what you think.

**Code ref:** https://github.com/dependabot/dependabot-core/blob/6d2bc8ce93c9efd64f7f1d6359e109ab8703b099/docker/lib/dependabot/docker/update_checker.rb#L158-L171

### LINKED ISSUE #4419: No pull request is created if a digest SHA is used to identify a docker-image
**Package ecosystem**
docker

**Package manager version**
n/a

**Language version**
n/a

[Dockerfile.txt](https://github.com/dependabot/dependabot-core/files/7546744/Dockerfile.txt)
> FROM openjdk:11-jre-slim@sha256:9223f8e5033f47c86323d31aa98a3779c4edd89d92194287f04e2f5391fd9ac1
>
> ENV JAVA_TOOL_OPTIONS=""
> ENV SERVER_PORT=8080
>
> COPY ${project.build.finalName}.jar /${project.build.finalName}.jar
>
> EXPOSE 8080
> 
> RUN adduser --system exec
> 
> USER exec
>
> ENTRYPOINT ["java", "-Djava.security.egd=file:/dev/./urandom", "-jar", "/${project.build.finalName}.jar"]

**dependabot.yml content**
[dependabot.yml.txt](https://github.com/dependabot/dependabot-core/files/7546769/dependabot.yml.txt)
> version: 2
> updates:
> \- package-ecosystem: docker
> directory: "/src/assembly/docker"
> schedule:
> interval: daily
> time: "01:00"
> open-pull-requests-limit: 10

**Updated dependency**
Based on the information on [hub.docker-com](https://hub.docker.com/_/openjdk?tab=tags&page=1&name=11-jre-slim) from Nov 16, 2021
> openjdk:11-jre-slim@sha256:5352ab50afd260ab429b2debf82b3a5ba52b0785acff983f0f847798463871ab

I would expect that a pull request is created which is changing the SHA digest to the version that is available through hub.docker.com.

n/a

>   proxy | time="2021-11-16T01:08:56Z" level=info msg="proxy starting" commit=d5f262668736016da1a91e42cb4fba36a081bddf
>   proxy | 2021/11/16 01:08:56 Listening (:1080)
> updater | 2021-11-16T01:08:56.459542996 [anonymous-instance:main:WARN:src/firecracker/src/main.rs:370] You are using a deprecated parameter: --seccomp-level 2, that will be removed in a future version.
> updater | 2021-11-16T01:08:56.482486950 [233159402:main:WARN:src/devices/src/legacy/serial.rs:432] Detached the serial input due to peer close/error.
> updater | time="2021-11-16T01:08:58Z" level=info msg="guest starting" commit=aca5609d0a3f160d4065a3c391d9a0e3c36b620d
> updater | time="2021-11-16T01:08:58Z" level=info msg="starting job..." fetcher_timeout=5m0s job_id=233159402 updater_timeout=45m0s updater_version=0.166.0-6eea28a15de585bcd28ef1877265e46598e3a339
> updater | I, [2021-11-16T01:08:59.451688 #7]  INFO -- sentry: ** [Raven] Raven 3.1.2 ready to catch errors
> updater | warning: parser/current is loading parser/ruby27, which recognizes
> updater | warning: 2.7.4-compliant syntax, but you are running 2.7.1.
> updater | warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
> updater | INFO <job_233159402> Starting job processing
>   proxy | 2021/11/16 01:09:01 [002] GET https://api.github.com:443/repos/some-org/some-repo
>   proxy | 2021/11/16 01:09:01 [002] * authenticating github api request
>   proxy | 2021/11/16 01:09:02 [002] 200 https://api.github.com:443/repos/some-org/some-repo
>   proxy | 2021/11/16 01:09:02 [004] GET https://api.github.com:443/repos/some-org/some-repo/git/refs/heads/development
>   proxy | 2021/11/16 01:09:02 [004] * authenticating github api request
>   proxy | 2021/11/16 01:09:02 [004] 200 https://api.github.com:443/repos/some-org/some-repo/git/refs/heads/development
>   proxy | 2021/11/16 01:09:02 [006] GET https://api.github.com:443/repos/some-org/some-repo/contents/src/assembly/docker?ref=5940b0c8c19679a7f65db392be25b5d87791a25e
>   proxy | 2021/11/16 01:09:02 [006] * authenticating github api request
>   proxy | 2021/11/16 01:09:02 [006] 200 https://api.github.com:443/repos/some-org/some-repo/contents/src/assembly/docker?ref=5940b0c8c19679a7f65db392be25b5d87791a25e
>   proxy | 2021/11/16 01:09:02 [008] GET https://api.github.com:443/repos/some-org/some-repo/contents/src/assembly/docker/Dockerfile?ref=5940b0c8c19679a7f65db392be25b5d87791a25e
>   proxy | 2021/11/16 01:09:02 [008] * authenticating github api request
>   proxy | 2021/11/16 01:09:03 [008] 200 https://api.github.com:443/repos/some-org/some-repo/contents/src/assembly/docker/Dockerfile?ref=5940b0c8c19679a7f65db392be25b5d87791a25e
> updater | INFO <job_233159402> Finished job processing
> updater | time="2021-11-16T01:09:03Z" level=info msg="task complete" container_id=job-233159402-file-fetcher exit_code=0 job_id=233159402 step=fetcher
> updater | I, [2021-11-16T01:09:04.161896 #9]  INFO -- sentry: ** [Raven] Raven 3.1.2 ready to catch errors
> updater | warning: parser/current is loading parser/ruby27, which recognizes
> updater | warning: 2.7.4-compliant syntax, but you are running 2.7.1.
> updater | warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
> updater | INFO <job_233159402> Starting job processing
> updater | INFO <job_233159402> Starting update job for some-org/some-repo
> updater | INFO <job_233159402> Checking if openjdk 11-jre-slim needs updating
>   proxy | 2021/11/16 01:09:06 [012] GET https://registry.hub.docker.com:443/v2/library/openjdk/tags/list
>   proxy | 2021/11/16 01:09:06 [012] 401 https://registry.hub.docker.com:443/v2/library/openjdk/tags/list
>   proxy | 2021/11/16 01:09:06 [014] GET https://auth.docker.io:443/token?service=registry.docker.io&scope=repository%3Alibrary%2Fopenjdk%3Apull
>   proxy | 2021/11/16 01:09:06 [014] 200 https://auth.docker.io:443/token?service=registry.docker.io&scope=repository%3Alibrary%2Fopenjdk%3Apull
>   proxy | 2021/11/16 01:09:06 [016] GET https://registry.hub.docker.com:443/v2/library/openjdk/tags/list
>   proxy | 2021/11/16 01:09:08 [016] 200 https://registry.hub.docker.com:443/v2/library/openjdk/tags/list
>   proxy | 2021/11/16 01:09:08 [018] HEAD https://registry.hub.docker.com:443/v2/library/openjdk/manifests/latest
>   proxy | 2021/11/16 01:09:08 [018] 401 https://registry.hub.docker.com:443/v2/library/openjdk/manifests/latest
>   proxy | 2021/11/16 01:09:08 [020] GET https://auth.docker.io:443/token?service=registry.docker.io&scope=repository%3Alibrary%2Fopenjdk%3Apull
>   proxy | 2021/11/16 01:09:08 [020] 200 https://auth.docker.io:443/token?service=registry.docker.io&scope=repository%3Alibrary%2Fopenjdk%3Apull
>   proxy | 2021/11/16 01:09:08 [022] HEAD https://registry.hub.docker.com:443/v2/library/openjdk/manifests/latest
>   proxy | 2021/11/16 01:09:08 [022] 200 https://registry.hub.docker.com:443/v2/library/openjdk/manifests/latest
>   proxy | 2021/11/16 01:09:08 [024] HEAD https://registry.hub.docker.com:443/v2/library/openjdk/manifests/18
>   proxy | 2021/11/16 01:09:08 [024] 401 https://registry.hub.docker.com:443/v2/library/openjdk/manifests/18
>   proxy | 2021/11/16 01:09:08 [026] GET https://auth.docker.io:443/token?service=registry.docker.io&scope=repository%3Alibrary%2Fopenjdk%3Apull
>   proxy | 2021/11/16 01:09:08 [026] 200 https://auth.docker.io:443/token?service=registry.docker.io&scope=repository%3Alibrary%2Fopenjdk%3Apull
>   proxy | 2021/11/16 01:09:08 [028] HEAD https://registry.hub.docker.com:443/v2/library/openjdk/manifests/18
>   proxy | 2021/11/16 01:09:08 [028] 200 https://registry.hub.docker.com:443/v2/library/openjdk/manifests/18
>   proxy | 2021/11/16 01:09:08 [030] HEAD https://registry.hub.docker.com:443/v2/library/openjdk/manifests/17.0.1
>   proxy | 2021/11/16 01:09:08 [030] 401 https://registry.hub.docker.com:443/v2/library/openjdk/manifests/17.0.1
>   proxy | 2021/11/16 01:09:08 [032] GET https://auth.docker.io:443/token?service=registry.docker.io&scope=repository%3Alibrary%2Fopenjdk%3Apull
>   proxy | 2021/11/16 01:09:08 [032] 200 https://auth.docker.io:443/token?service=registry.docker.io&scope=repository%3Alibrary%2Fopenjdk%3Apull
>   proxy | 2021/11/16 01:09:08 [034] HEAD https://registry.hub.docker.com:443/v2/library/openjdk/manifests/17.0.1
>   proxy | 2021/11/16 01:09:08 [034] 200 https://registry.hub.docker.com:443/v2/library/openjdk/manifests/17.0.1
> updater | INFO <job_233159402> Latest version is 11-jre-slim
>   proxy | 2021/11/16 01:09:08 [036] HEAD https://registry.hub.docker.com:443/v2/library/openjdk/manifests/11-jre-slim
>   proxy | 2021/11/16 01:09:08 [036] 401 https://registry.hub.docker.com:443/v2/library/openjdk/manifests/11-jre-slim
>   proxy | 2021/11/16 01:09:08 [038] GET https://auth.docker.io:443/token?service=registry.docker.io&scope=repository%3Alibrary%2Fopenjdk%3Apull
>   proxy | 2021/11/16 01:09:08 [038] 200 https://auth.docker.io:443/token?service=registry.docker.io&scope=repository%3Alibrary%2Fopenjdk%3Apull
>   proxy | 2021/11/16 01:09:09 [040] HEAD https://registry.hub.docker.com:443/v2/library/openjdk/manifests/11-jre-slim
>   proxy | 2021/11/16 01:09:09 [040] 200 https://registry.hub.docker.com:443/v2/library/openjdk/manifests/11-jre-slim
> updater | INFO <job_233159402> No update needed for openjdk 11-jre-slim
> updater | INFO <job_233159402> Finished job processing
> updater | time="2021-11-16T01:09:09Z" level=info msg="task complete" container_id=job-233159402-updater exit_code=0 job_id=233159402 step=updater

https://github.com/gbtec-ag/dependabot-docker-sha-digest-issue

deivid-rodriguez:
This may seem too aggressive, and I'm probably missing something, and I haven't yet run full tests BUT, I think this is incorrect and not necessary.

When a base image is pinned to a specific sha, Dependabot does a very heavy operation of requesting the list of tags from the registry, and then asking for the sha of each of them until one tag that matches the current sha is found.

However, this is not necessary, because Dependabot will end up updating to the latest sha anyways, since it defaults to always keeping the current style/precision.

Also, I believe this is incorrect, because docker tags are mutable, so the registry will only return the latest sha associated to each tag. So this loop is likely to run forever to end up not finding any matches if the current sha present in the Dockerfile has been overwritten.

Finally, this should be a big performance boost because this operation is very very slow.

Fixes #2997.
Fixes #4419.

Just opening a WIP PR to check CI for now.


Nishnha:
> However, this is not necessary, because Dependabot will end up updating to the latest sha anyways, since it defaults to always keeping the current style/precision.

I think the reason we fetch versions and try to match them against the SHA is to determine the version precision. Otherwise, we can't know what version the image needs to be upgraded to

Not sure if I'm misunderstanding something here though

Nishnha:
Oh I think I was missing something.... Docker doesn't support any other `versioning strategies` other than default https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy

So we really might just be updating all Docker SHAs to `latest`. This code might actually be redundant!

Can you confirm that matches what your reasoning for removing this feature?

deivid-rodriguez:
Let me try explain what happens when Dependabot tries to update a `Dockerfile` using a `sha256` reference before and after this PR through an example.

Let's take for example the declaration `FROM ghcr.io/dependabot/dependabot-updater@sha256: 654abc3436d0c99896df14d3d6bb450774fd5639a97e8a2bf88bc813151ef7bd`.

This is what happens without this PR:

* Dependabot tries to find a tag with sha matching `sha256: 654abc3436d0c99896df14d3d6bb450774fd5639a97e8a2bf88bc813151ef7bd`.
* This sometimes works and something don't because while today a specific SHA may be associated with a tag, docker tags are usually moved and previous SHAs pointed to the tag get "orphaned". Also, when this works, it's very slow because it needs to make an API request for every tag checked, and some images have a lot of tags.
* Assuming the previous point works, Dependabot finds a tag associated to the SHA. In this case, it's [v2.0.20221207081814](https://github.com/dependabot/dependabot-core/pkgs/container/dependabot-updater/57575794?tag=v2.0.20221207081814).
* Now it uses the that tag to find another tag to update. In this case, since the tag uses patch level precision, it will find an up to date patch level tag. In this case, it would figure out it needs to update to the latest tag: [v2.0.20221213172033](https://github.com/orgs/dependabot/packages/container/dependabot-updater/58658754?tag=v2.0.20221213172033).
* Finally it finds the SHA of the updated tag: `sha256:e95a9f539fa58bc6a2e57fe6333cdcb08d4da9774bddc7d7b66e07bd748a89dd`.
* The final result is Dependabot would create a PR with title "Update ghcr.io/dependabot/dependabot-updater from v2.0.20221207081814 to v2.0.20221213172033" and that updates the Dockerfile from `FROM ghcr.io/dependabot/dependabot-updater@sha256: 654abc3436d0c99896df14d3d6bb450774fd5639a97e8a2bf88bc813151ef7bd` to `FROM ghcr.io/dependabot/dependabot-updater@sha256:e95a9f539fa58bc6a2e57fe6333cdcb08d4da9774bddc7d7b66e07bd748a89dd`.

This is the logic with this PR:

* Dependabot sees that the `Dockerfile` is locked to a `sha256` reference.
* Dependabot finds the `sha256` reference of the `latest` tag.
* Dependabot creates a PR with title "Update ghcr.io/dependabot/dependabot-updater from 654abc3436d0c99896df14d3d6bb450774fd5639a97e8a2bf88bc813151ef7bd to e95a9f539fa58bc6a2e57fe6333cdcb08d4da9774bddc7d7b66e07bd748a89dd" and the same diff as with the previous logic.

deivid-rodriguez:
> So we really might just be updating all Docker SHAs to latest. The feature seems to be redundant!

> Can you confirm that matches what your reasoning for removing this feature? The only exception I can think of is security updates, but GitHub doesn't publish any Docker advisories 🤔

Yes, that's exactly right by the way! Indeed, if we wanted a "lowest fixed version" logic for Security Updates, we wouldn't have that, but that does not seem a concern now, and we don't have that logic in place at the moment anyways.

docker/lib/dependabot/docker/update_checker.rb
```diff
@@ -52,6 +52,7 @@ class UpdateChecker < Dependabot::UpdateCheckers::Base
           #{VERSION_WITH_SFX}|
           #{VERSION_WITH_PFX_AND_SFX}
       /x
+      DIGEST = /^sha256:[0-9a-f]{64}$/
```
Nishnha, line 55: Calling this the digest makes sense, and we use the same term in the FileParser, but they have different definitions
https://github.com/dependabot/dependabot-core/blob/1cd6b45919e19c6b2aa719015629804302724e18/docker/lib/dependabot/docker/file_parser.rb#L27

Can we try to make the regex consistent between the two?

Nishnha (approved):
Thanks for walking me through the logic on this ❤️ 

Awesome change. I had one minor nit but this LGTM

stevehipwell:
@deivid-rodriguez do you have a status on this?

deivid-rodriguez:
Hei! There's no ETA but the PR is already approved and just needs a rebase and a final look to get shipped. It should not take too long.

docker/lib/dependabot/docker/update_checker.rb
deivid-rodriguez: @Nishnha I took my a long time to loop back to this, but I addressed your comment and this PR is now ready for another review!

deivid-rodriguez:
This should be now ready for another review unless CI proves me wrong. It's a big PR, but mostly deletes a lot of code!

docker/lib/dependabot/docker/update_checker.rb
```diff
@@ -234,20 +241,18 @@ def digest_of(tag)
...
-              nil
-            else
-              raise if attempt > 3
+        @digests[tag] = fetch_digest_of(tag)
+      end
 
-              retry
-            end
-          end
+      def fetch_digest_of(tag)
```
Nishnha, line 247: Nice cleanup!

common/spec/dependabot/pull_request_creator/message_builder_spec.rb
```diff
@@ -441,7 +441,7 @@ def commits_details(base:, head:)
                 groups: [],
                 source: {
                   type: "digest",
-                  digest: "sha256:2167a21baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" \
+                  digest: ":2167a21baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" \
```
Nishnha, line 444: Is the leading colon here a typo?

Nishnha (commented):
I notice this can match tags with different prefixes. It seems like these are both valid?

* `sha265:18305429afa14ea462f810146ba44d4363ae76e4c8dfc38288cf73aa07485005` 
* `:sha265:18305429afa14ea462f810146ba44d4363ae76e4c8dfc38288cf73aa07485005` 

But most tags will be in the form of `FROM ghcr.io/dependabot/dependabot-updater@sha256: 654abc3436d0c99896df14d3d6bb450774fd5639a97e8a2bf88bc813151ef7bd`, right?

Is there a reason you support the other prefix with a leading `:`? Do they also appear in Dockerfiles?

docker/spec/dependabot/docker/file_parser_spec.rb
```diff
@@ -259,197 +259,11 @@
           it "has the right details" do
             expect(dependency).to be_a(Dependabot::Dependency)
             expect(dependency.name).to eq("ubuntu")
-            expect(dependency.version).to eq("12.04.5")
+            expect(dependency.version).to eq("18305429afa14ea462f810146ba44d4363ae76e4c8dfc38288cf73aa07485005")
             expect(dependency.requirements).to eq(expected_requirements)
           end
         end
 
-        context "for a private registry" do
```
Nishnha, line 267: Were these private registry tests moved elsewhere or no longer necessary?

docker/spec/dependabot/docker/file_parser_spec.rb
```diff
@@ -228,7 +228,7 @@
             and_return(status: 200, body: "", headers: digest_headers)
         end
 
-        its(:length) { is_expected.to eq(0) }
+        its(:length) { is_expected.to eq(1) }
```
Nishnha, line 231: 🎉

docker/spec/dependabot/docker/file_parser_spec.rb
deivid-rodriguez: I removed them because I think they are no longer necessary since the code they validated was removed.

Previously, when an image reference in a Dockerfile included only a digest, we would need to access the registry to figure out the corresponding version for that digest. Now we set the digest as the version directly and don't try to look up a numeric equivalent for the digest.

Basically, we now use exclusively the information provided in the Dockerfile for parsing dependencies, so we no longer need any registry access here.

common/spec/dependabot/pull_request_creator/message_builder_spec.rb
deivid-rodriguez: Typo indeed, nice catch! I'll fix it.

deivid-rodriguez (commented):
@Nishnha Regarding your comment about leading `:`, is it because the regex is now

```ruby
IMAGE_SPEC = %r{^(#{REGISTRY}/)?#{IMAGE}#{TAG}?(?:@sha256:#{DIGEST})?#{NAME}?}x
```

If that's it, the `?:` part before `sha256` is meant to mean "match this group, but don't include it as a named match". So in other words, that leading `:` is not supposed to be matched, it's just a special character when preceded by `?`.

See this: https://rubular.com/r/FXdZeMjcFcMH03.

common/spec/dependabot/pull_request_creator/message_builder_spec.rb
deivid-rodriguez: Fixed!

Nishnha:
> @Nishnha Regarding your comment about leading `:`, is it because the regex is now
> 
> ```ruby
> IMAGE_SPEC = %r{^(#{REGISTRY}/)?#{IMAGE}#{TAG}?(?:@sha256:#{DIGEST})?#{NAME}?}x
> ```
> 
> If that's it, the `?:` part before `sha256` is meant to mean "match this group, but don't include it as a named match". So in other words, that leading `:` is not supposed to be matched, it's just a special character when preceded by `?`.
> 
> See this: https://rubular.com/r/FXdZeMjcFcMH03.

TIL! 😮 

The only other question I have is why remove the private registry tests in https://github.com/dependabot/dependabot-core/pull/6150/commits/666c1004dac1cdad5120d977a1e10d1a32cd94f6 ?

deivid-rodriguez:
Sorry @Nishnha, not sure what happened there, I was sure I had answered the question about the removed tests 😅.

I removed those because the logic they were testing has been removed.

Previously, if we found an image reference with only a SHA-version (no numeric version), we would talk to the registry to try figure out whether there's a numeric tag associated to that SHA. Those tests were testing that the registry requests to do this happen correctly.

With the new approach, we don't mess with trying to find a numeric version for a given SHA, and rely exclusively to the information present in the Dockerfile itself. So we don't talk to the registry anymore when parsing Dockerfile dependencies.

This is why those tests were removed, I hope it makes sense!

Nishnha (approved):

PR #6155, opened 2022-11-17
prevent workflow from pushing on closed PRs

jakecoffman:
I noticed when we comment on a closed PR it was trying to run this workflow. I guess it considers even a simple comment to be a pull request review? 

So to prevent accidentally deploying to the latest tag, I put it in its own workflow with just a push to main trigger.

I also added the PR state to the "DECISION" variable in the branch release workflow, so we only consider open PRs now, others will just skip all steps. Also removed the pull_request trigger completely since we should only deploy when approval happens, so that could never be the case for that trigger.


deivid-rodriguez (approved):
Notice the same weirdness too some days ago, thanks for fixing it!

jakecoffman:
testing

jakecoffman:
Oh it won't trigger for me because I created the PR. @deivid-rodriguez can you comment on this PR?

deivid-rodriguez:
Please ignore this dear workflow 🙏

jakecoffman:
I was expecting it to run but skip, but it didn't run at all. 🤷

Issue #6157, opened 2022-11-17
Not detecting old packages with pipenv and private repository


m3h:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

pipenv

### Language version

Python 3.8

### Manifest location and content before the Dependabot update

[/Pipfile](https://github.com/m3h/test_dependabot/blob/main/Pipfile)
[/Pipfile.lock](https://github.com/m3h/test_dependabot/blob/main/Pipfile.lock)

### dependabot.yml content

[dependabot.yml](https://github.com/m3h/test_dependabot/blob/main/.github/dependabot.yml)

### Updated dependency

`example-package-m3h` from `0.0.1` to `0.0.2` 

https://test.pypi.org/project/example-package-m3h/

### What you expected to see, versus what you actually saw

I expected Dependabot to submit a PR to upgrade `example-package-m3h` from `0.0.1` to `0.0.2`

What I saw was an error in the Dependabot logs:

```
 proxy | time="2022-11-17T14:48:13Z" level=info msg="proxy starting" commit=843c0928fee6f7a20d835e4f6c436aa7c1567af9
  proxy | 2022/11/17 14:48:13 Listening (:1080)
updater | 2022-11-17T14:48:14.146720909 [515921304:main:WARN:src/devices/src/legacy/serial.rs:214] Detached the serial input due to peer close/error.
updater | time="2022-11-17T14:48:16Z" level=info msg="guest starting" commit=c919e3df5491e18828a7a1d5fc6d8a0d7aad90f5
updater | time="2022-11-17T14:48:16Z" level=info msg="starting job..." fetcher_timeout=10m0s job_id=515921304 updater_timeout=45m0s updater_version=e320eddfbb9c71d4788279231be43846c48383a0
updater | I, [2022-11-17T14:48:17.353685 #7]  INFO -- sentry: ** [Raven] Raven 3.1.2 ready to catch errors
updater | INFO <job_515921304> Starting job processing
  proxy | 2022/11/17 14:48:19 [002] GET https://api.github.com:443/repos/m3h/test_dependabot
  proxy | 2022/11/17 14:48:19 [002] * authenticating github api request
  proxy | 2022/11/17 14:48:22 [002] 200 https://api.github.com:443/repos/m3h/test_dependabot
  proxy | 2022/11/17 14:48:22 [004] GET https://api.github.com:443/repos/m3h/test_dependabot/git/refs/heads/main
[...]

PR #6158, opened 2022-11-17
only replace url for private pipenv source

m3h:
Instead of filtering out private pipenv sources completely, and inserting the urls from the Dependabot credentials configuration (which removes other parts of the pipenv source configuration, e.g. the source name), only replace the url with the new authed url.

This should fix https://github.com/dependabot/dependabot-core/issues/6157

### DIFF
--- a/python/lib/dependabot/python/file_updater/pipfile_preparer.rb
+++ b/python/lib/dependabot/python/file_updater/pipfile_preparer.rb
@@ -21,7 +21,7 @@ module Dependabot
           pipfile_object = TomlRB.parse(pipfile_content)
 
           pipfile_object["source"] =
-            pipfile_sources.reject { |h| h["url"].include?("${") } +
+            pipfile_sources.filter_map { |h| sub_auth_url(h, credentials) } +
             config_variable_sources(credentials)
 
           TomlRB.dump(pipfile_object)
@@ -112,6 +112,22 @@ module Dependabot
           @pipfile_sources ||=
             TomlRB.parse(pipfile_content).fetch("source", []).
             map { |h| h.dup.merge("url" => h["url"].gsub(%r{/*$}, "") + "/") }
+        end
+
+        def sub_auth_url(source, credentials)
+          if source["url"].include?("${")
+            base_url = source["url"].sub(/\${.*}@/, "")
+
+            source_cred = credentials.
+                          select { |cred| cred["type"] == "python_index" }.
+                          find { |c| c["index-url"].sub(/\${.*}@/, "") == base_url }
+
+            return nil if source_cred.nil?
+
+            source["url"] = AuthedUrlBuilder.authed_url(credential: source_cred)
+          end
+
+          source
         end
 
         def config_variable_sources(credentials)
--- a/python/spec/dependabot/python/file_updater/pipfile_preparer_spec.rb
+++ b/python/spec/dependabot/python/file_updater/pipfile_preparer_spec.rb
@@ -158,5 +158,31 @@ RSpec.describe Dependabot::Python::FileUpdater::PipfilePreparer do
           to include("https://username:password@pypi.posrip.com/pypi/")
       end
     end
+
+    context "with auth details provided in Pipfile" do
+      let(:credentials) do
+        [{
+          "type" => "git_source",
+          "host" => "github.com",
+          "username" => "x-access-token",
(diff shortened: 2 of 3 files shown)


deivid-rodriguez (commented):
Nice!

This makes sense to me, and the implementation looks good, you just need a few style fixes.

It would be nice though to add a spec to cover this fix, do you think you could do that?

m3h:
Cool, I've given it a go - let me know if that works for you.

deivid-rodriguez:
Nice, thank you! There's a spec that broke and looks related to the changes in here, can you have a look?

m3h:
I've made some changes, but I'm not 100% I've handled it correctly.
I've now 
* added in the auth from the credentials to existing Pipfile sources if the repository URLs can be matched,
* removed existing Pipfile sources that require auth but don't have a match in the credentials
* added all the sources from the credentials.

I have a suspicion that the old approach is based on pipenv scanning _all_ repository sources, which doesn't happen in newer versions of pipenv (see the note [here](https://pipenv.pypa.io/en/latest/advanced/#specifying-package-indexes)).

But I don't really have the historical background with pipenv to comment more on that.

deivid-rodriguez:
Interesting, so what you mean is that your initial work that was making one spec fail may be correct in newer piping versions, and we may want to instead change the spec to the new "correct behavior"?

In any case, this PR looks good to me, pending a closer review.

jeffwidman:
@m3h can you comment on ☝️ ?

m3h:
> Interesting, so what you mean is that your initial work that was making one spec fail may be correct in newer piping versions, and we may want to instead change the spec to the new "correct behavior"?
> 
> In any case, this PR looks good to me, pending a closer review.

No, sorry, I don't think I was being clear. I do think the code needs updating due to the new `pipenv` behaviour. See my comments in https://github.com/dependabot/dependabot-core/issues/6157

The current dependabot behaviour is to replace this:
```
[[source]]
url = "https://${TEST_USER}:${TEST_PASS}@test.pypi.org/simple"
verify_ssl = true
name = "testindex"
```
with this:
```
[[source]]
url = "https://my_test_user:my_test_password@test.pypi.org/simple"
```

stripping the `name` and `verify_ssl` parameters in the process. I was just commenting that this probably used to work when `pipenv` searched *all* sources for a matching package, whereas current versions of `pipenv` will only search sources with a *matching* `name` parameter (which is being stripped by the current dependabot code).


The intent of the PR is to have dependabot instead replace this:
```
[[source]]
url = "https://${TEST_USER}:${TEST_PASS}@test.pypi.org/simple"
verify_ssl = true
name = "testindex"
```
with this:
```
[[source]]
url = "https://my_test_user:my_test_password@test.pypi.org/simple"
verify_ssl = true
name = "testindex"
```

keeping the `verify_ssl` and `name` (and other) parameters intact.

python/lib/dependabot/python/file_updater/pipfile_preparer.rb
```diff
@@ -114,6 +114,24 @@ def pipfile_sources
...
+
+            source_cred = credentials.
+                          select { |cred| cred["type"] == "python_index" }.
+                          find { |c| c["index-url"].sub(/\${.*}@/, "") == base_url }
+
+            return nil if source_cred.nil?
+
+            source["url"] = AuthedUrlBuilder.authed_url(credential: source_cred)
+            return source
+
```
deivid-rodriguez, line 129: The above two lines (`return source` + blank line) are redundant.

deivid-rodriguez (commented):
Looks good to me!

I made one minor comment, and I would squash commits into one (I think the second one was just amending the first commit because it broke some specs).

python/lib/dependabot/python/file_updater/pipfile_preparer.rb
m3h: Thanks, I've removed them!

m3h:
@deivid-rodriguez  thanks for the comments! I think I've applied them all now.

jeffwidman (approved):
Thanks for fixing this! Nice job on the debugging and code spelunking to figure out a clean solution.

jeffwidman:
I'm going to set auto-merge as I assume CI will pass and looks low risk to merge straight to `main` w/o a deploy, so this will probably get picked up as part of a deploy tomorrow.

PR #6159, opened 2022-11-17
Create a permanent root Gemfile

deivid-rodriguez:
For the same reasons this is useful in the development container, I'd like to also have it outside of the development container.

My use case is to quickly run RuboCop over the whole repo when making global changes across ecosystems.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -8,6 +8,7 @@ AllCops:
     - "*/bin/**/*"
     - "**/tmp/**/*"
     - "*/spec/fixtures/**/*"
+    - "dry-run/**/*"
   NewCops: enable
   SuggestExtensions: false
 Gemspec/DeprecatedAttributeAssignment:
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+eval_gemfile File.join(File.dirname(__FILE__), "omnibus/Gemfile")
```
landongrindheim, line 3: 👌 Nice solution!

.rubocop.yml
```diff
@@ -8,6 +8,7 @@ AllCops:
     - "*/bin/**/*"
     - "**/tmp/**/*"
     - "*/spec/fixtures/**/*"
+    - "dry-run/**/*"
```
landongrindheim, line 11: 🔎 Can you provide context on this change? It doesn't _seem_ related to the Gemfile.

landongrindheim (approved):

.rubocop.yml
deivid-rodriguez: Oh, absolutely, I should've at least extracted that to its own commit, let me do that. That's because `dry-run` is a common ignored folder at the repo root generated by our `dry-run.rb` script. If you run RuboCop from the repo root, you don't want to inspect all the repos in there.

PR #6163, opened 2022-11-18
Transform "No files changed!" unexpected error into expected one

deivid-rodriguez:
If a npm dependency is bundled, we can't really upgrade it unless the parent dependency is bumped.

This change should teach the update checker about that, so that unexpected security update errors show expected "upgrade is not possible" errors instead.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb
@@ -158,6 +158,7 @@ module Dependabot
         return version_resolver.latest_version_resolvable_with_full_unlock? if dependency.top_level?
 
         return false unless security_advisories.any?
+        return false unless subdependency_version_resolver.latest_version_resolvable_with_full_unlock?
 
         vulnerability_audit["fix_available"]
       end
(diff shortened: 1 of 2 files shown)


jeffwidman (approved):
This seems reasonable to me, but I know @mctofu @Nishnha @bdragon put a lot of time into NPM transitive-dependency work, including security updates, so would feel better if one of them could take a look and confirm this wasn't intentionally written this way.

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker.rb
```diff
@@ -158,6 +158,7 @@ def latest_version_resolvable_with_full_unlock?
         return version_resolver.latest_version_resolvable_with_full_unlock? if dependency.top_level?
 
         return false unless security_advisories.any?
+        return false unless subdependency_version_resolver.latest_version_resolvable_with_full_unlock?
```
mctofu, line 161: Ideally the vulnerability auditor would be able to propose updating the parent dependency (via `npm audit fix`) but if that's not currently the case I think it makes sense to bail early here.

Naming the method `latest_version_resolvable_with_full_unlock?` might be misleading because the vulnerability auditor isn't guaranteed to find a solution. For example, there might not be a parent update which allows the fixed version of the sub dependency. I think the method name could be more specific to bundled dependencies if we know those won't update.

deivid-rodriguez: Sorry I never got back to this!

What you say definitely make sense. A bundled dependency could in principle be updated just fine too by updating the parent. Also the naming does not seem accurate either as you point out.

Unfortunately I completely lost context of the original problem now 😞. I'll re-read the linked issue and see if I can still reproduce the error.

PR #6164, opened 2022-11-18
Refactor some common logic across ecosystems

deivid-rodriguez:
While working on #6071 I noticed some common logic across ecosystem that I think makes sense to be moved to common.

There's a lot of more that can be done here, but I didn't want to deviate too much from my original goal, so I stopped here which I believe is already an improvement.


jurre (approved):

deivid-rodriguez:
Thanks @jurre! I'll deploy this now to make sure it's all good!

PR #6165, opened 2022-11-18
need pull_request trigger for branch updates

jakecoffman:
#6155 removed the pull_request trigger, but we need it since it triggers when we update a branch from master, for instance, or push a small change. That way we don't need to ask for reapproval in those cases. 

I **think** types: edited is the correct thing to put. We don't need the workflow to run when the PR opens, or closes, etc.

I also fixed the names of the latest release workflow being the same as the branch release workflow, which then got me carried away to make all the names shorter and more accurate.

### DIFF
--- a/.github/workflows/docker-branch-releases.yml
+++ b/.github/workflows/docker-branch-releases.yml
@@ -1,16 +1,22 @@
-name: Push docker branch images
+name: Push branch images
 env:
   GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   BASE_IMAGE: "ubuntu:20.04"
   UPDATER_IMAGE: "dependabot/updater"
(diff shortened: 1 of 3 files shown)


.github/workflows/docker-branch-releases.yml
```diff
@@ -1,16 +1,22 @@
...
+    types:
+      - edited
   pull_request_review:
-    types: [submitted]
+    types:
+      - submitted
 
 jobs:
   push-updater-image:
-    name: Export dependabot-updater image to build artifacts
```
jakecoffman, line 13: This name was from a previous iteration where we uploaded to build artifacts.

deivid-rodriguez (approved):

jakecoffman:
I'll merge this after @pavera's merge so we don't have to update branches excessively.

PR #6166, opened 2022-11-18
Bitbucket improvements

stefangr:
### Broaden the functionality of the pull_requests method.

This way the code can also be used to fetch more than one pull request and the state of the pull requests to fetch can be specified.

### Add decline pull request method

With this function it's possible to decline a pull request with a provided comment (a default comment is provided as fallback).

### DIFF
--- a/common/lib/dependabot/clients/bitbucket.rb
+++ b/common/lib/dependabot/clients/bitbucket.rb
@@ -79,19 +79,24 @@ module Dependabot
         JSON.parse(response.body)
       end
 
-      def pull_requests(repo, source_branch, target_branch)
-        pr_path = "#{repo}/pullrequests"
-        # Get pull requests with any status
-        pr_path += "?status=OPEN&status=MERGED&status=DECLINED&status=SUPERSEDED"
+      def pull_requests(repo, source_branch, target_branch, status = %w(OPEN MERGED DECLINED SUPERSEDED))
+        pr_path = "#{repo}/pullrequests?"
+        # Get pull requests with given status
+        status.each { |n| pr_path += "status=#{n}&" }
         next_page_url = base_url + pr_path
         pull_requests = paginate({ "next" => next_page_url })
 
         pull_requests unless source_branch && target_branch
 
         pull_requests.select do |pr|
-          pr_source_branch = pr.fetch("source").fetch("branch").fetch("name")
+          if source_branch.nil?
+            source_branch_matches = true
+          else
+            pr_source_branch = pr.fetch("source").fetch("branch").fetch("name")
+            source_branch_matches = pr_source_branch == source_branch
+          end
           pr_target_branch = pr.fetch("destination").fetch("branch").fetch("name")
-          pr_source_branch == source_branch && pr_target_branch == target_branch
+          source_branch_matches && pr_target_branch == target_branch
         end
       end
 
--- a/common/lib/dependabot/clients/bitbucket.rb
+++ b/common/lib/dependabot/clients/bitbucket.rb
@@ -156,7 +156,7 @@ module Dependabot
         decline_path = "#{repo}/pullrequests/#{pr_id}/decline"
         post(base_url + decline_path, "")
 
-        comment = "Dependabot declined pull request" if comment.nil?
+        comment = "Dependabot declined the pull request." if comment.nil?
 
         content = {
           content: {
--- a/common/lib/dependabot/clients/bitbucket.rb
+++ b/common/lib/dependabot/clients/bitbucket.rb
@@ -12,6 +12,8 @@ module Dependabot
 
       class Forbidden < StandardError; end
 
+      class TimedOut < StandardError; end
+
       #######################
       # Constructor methods #
       #######################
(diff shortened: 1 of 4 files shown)


jeffwidman (commented):
Can you explain the use case for this?

We don't want to get into the business of being a generic BitBucket client, so want to limit the scope of our functionality to only the calls used within typical Dependabot workflows...

common/lib/dependabot/clients/bitbucket.rb
```diff
@@ -144,6 +151,23 @@ def create_pull_request(repo, pr_name, source_branch, target_branch,
       end
       # rubocop:enable Metrics/ParameterLists
 
+      def decline_pull_request(repo, pr_id, comment = nil)
+        # https://developer.atlassian.com/cloud/bitbucket/rest/api-group-pullrequests/
+        decline_path = "#{repo}/pullrequests/#{pr_id}/decline"
+        post(base_url + decline_path, "")
+
+        comment = "Dependabot declined pull request" if comment.nil?
```
jeffwidman, line 159:
```suggestion
        comment = "Dependabot declined the pull request." if comment.nil?
```

common/lib/dependabot/clients/bitbucket.rb
```diff
@@ -144,6 +151,23 @@ def create_pull_request(repo, pr_name, source_branch, target_branch,
       end
       # rubocop:enable Metrics/ParameterLists
 
+      def decline_pull_request(repo, pr_id, comment = nil)
+        # https://developer.atlassian.com/cloud/bitbucket/rest/api-group-pullrequests/
```
jeffwidman, line 155: Do you want to deep link to https://developer.atlassian.com/cloud/bitbucket/rest/api-group-pullrequests/#api-repositories-workspace-repo-slug-pullrequests-pull-request-id-decline-post ?

Either way is fine with me, shorter is also less likely to break.

jeffwidman (approved):
From a quick review looks fine to me, and thanks for including tests.

I'll merge it later today. Our current release process to RubyGems is unfortunately a little borked, and while we plan to fix that it likely means we won't be cutting a new tag for a bit so in the meantime you'll need to just pin to a SHA to pull this in.

common/lib/dependabot/clients/bitbucket.rb
stefangr: I applied the suggested change (also in the test).

common/lib/dependabot/clients/bitbucket.rb
stefangr: I choose just the page to keep it shorter.
And also to give a possible reference to other pull request related calls.
So I'd keep it like it is.

johnathanl:
Hi @stefangr,

Thank you for your contribution!

I was wondering what sort of Bitbucket access you need to give to the bot to be able to run the new code to create PRs. I have been getting error with our old PR Read/Write repo access token and stuck with this error. Here are the screenshot of the token generation and the error.

Screen Shot 2022-12-06 at 10 30 49

```
  - Updating sentry-sdk (from 1.11.0)…/home/dependabot/dependabot-script/.bundle/gems/dependabot-common-0.214.0/lib/dependabot/clients/bitbucket.rb:217:in `get': Dependabot::Clients::Bitbucket::Unauthorized (Dependabot::Clients::Bitbucket::Unauthorized)
        from /home/dependabot/dependabot-script/.bundle/gems/dependabot-common-0.214.0/lib/dependabot/clients/bitbucket.rb:172:in `current_user'
        from /home/dependabot/dependabot-script/.bundle/gems/dependabot-common-0.214.0/lib/dependabot/clients/bitbucket.rb:177:in `default_reviewers'
        from /home/dependabot/dependabot-script/.bundle/gems/dependabot-common-0.214.0/lib/dependabot/clients/bitbucket.rb:129:in `create_pull_request'
        from /home/dependabot/dependabot-script/.bundle/gems/dependabot-common-0.214.0/lib/dependabot/pull_request_creator/bitbucket.rb:79:in `create_pull_request'
        from /home/dependabot/dependabot-script/.bundle/gems/dependabot-common-0.214.0/lib/dependabot/pull_request_creator/bitbucket.rb:37:in `create'
        from /home/dependabot/dependabot-script/.bundle/gems/dependabot-common-0.214.0/lib/dependabot/pull_request_creator.rb:107:in `create'
        from ./generic-update-script.rb:233:in `block in <main>'
        from ./generic-update-script.rb:176:in `each'
        from ./generic-update-script.rb:176:in `<main>'
```

I need the new Python 3.10.7 support for one of our repo to run the dependabot-script, other's we have rolled back to 0.213.0 for the moment.

Thank you in advance!

stefangr:
@johnathanl
I use an [oauth consumer](https://github.com/dependabot/dependabot-script/compare/main...stefangr:dependabot-script:bitbucket-oauth-consumer) instead of a repository access token.

For that I need the following permissions:
- account:read
- pull requests:write

The call that is failing for you is not available if you do not use an oauth consumer.
So I guess there is need to catch the potential unauthorized response.

PR #6170, opened 2022-11-18
Respect existing precision of base docker images

### LINKED ISSUE #3933: Docker bump adds patch version
**Package ecosystem**
Docker

**Updated dependency**
Alpine from 3.13 to 3.14.0

I expected to be updated to 3.14, not 3.14.0

https://github.com/nginxinc/kubernetes-ingress/pull/1672/files

deivid-rodriguez:
Fixes #3933.


docker/spec/dependabot/docker/update_checker_spec.rb
```diff
@@ -451,7 +451,7 @@
 
       context "followed by numbers and with less components than other version but higher underscore part" do
         let(:version) { "11.0.16_8-jdk" }
-        it { is_expected.to eq("11.0.16.1_1-jdk") }
+        it { is_expected.to eq("11.0.16_8-jdk") }
```
deivid-rodriguez, line 454: This one update is probably expected so this needs more work.

I think we may want to choose to keep precision when we have to choose between several tags that all point to the same SHA (like, say, `php:8`, `php:8.1`, and `php:8.1.12`) , but allow precision changes other than that.

deivid-rodriguez: I amended the PR as suggested in the previous review comment. Happy with it now!

docker/spec/dependabot/docker/update_checker_spec.rb
```diff
@@ -960,4 +990,17 @@
       end
     end
   end
+
+  private
+
+  def stub_same_sha_for(*tags)
```
jeffwidman, line 996: Is it convention to put private test helpers at the bottom of the spec files? In other language ecosystems, I'm more used to them being at the top.

jeffwidman (approved):
👍 

I was a bit confused when initially reviewing this PR because the extra refactoring made it hard to follow the logic changes... but then I noticed your nice commit history and that made it _much_ easier to follow, so thank you for taking the time to do that. And nice refactors too!

docker/spec/dependabot/docker/update_checker_spec.rb
deivid-rodriguez: No, not really, I think my personal style here is pretty rare actually. Normally people would move these into a `let` at the top indeed. Should I change it?

deivid-rodriguez: Actually, since `let` does not take parameters, I'm not really sure where people put this kind of test helper 😅. I'll leave this as is.

PR #6172, opened 2022-11-18
submodule cloning may not be important to the job, ignore errors

jakecoffman:
In #5982 we enabled submodule cloning by default. However, sometimes a repo has a submodule that is inaccessible to Dependabot that is unrelated to dependencies.

We talked about adding a flag to disable it, but it's nice when Dependabot just works. 

So for this PR I took the approach to just ignore the submodule clone errors and continue with the update. That way if it was  not important, it will succeed in doing the update.

If it was important it will fail later, but I added an error log that says the nature of the failure during the fetch as a hint as to what might be wrong.

### DIFF
--- a/common/lib/dependabot/file_fetchers/base.rb
+++ b/common/lib/dependabot/file_fetchers/base.rb
@@ -26,6 +26,9 @@ module Dependabot
         Dependabot::Clients::CodeCommit::NotFound
       ].freeze
 
+      GIT_SUBMODULE_ERROR_REGEX =
+        /^fatal: unable to access '(?<url>.*)': The requested URL returned error: (?<code>\d+)$/
+
       def self.required_files_in?(_filename_array)
         raise NotImplementedError
       end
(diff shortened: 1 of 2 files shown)


common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -592,19 +595,34 @@ def _clone_repo_contents(target_directory:)
...
+            raise unless e.message.match?(GIT_SUBMODULE_ERROR_REGEX) && e.message.include?("submodule")
+
+            submodule_cloning_failed = true
+            match = e.message.match(GIT_SUBMODULE_ERROR_REGEX)
+            url = match.named_captures["url"]
+            code = match.named_captures["code"]
+
+            # Submodules might be in the repo but unrelated to dependencies,
+            # so ignoring this error to try the update anyway since the base repo exists.
+            Dependabot.logger.error("Cloning of submodule failed: #{url} error: #{code}")
```
landongrindheim, line 616: 👌 I can picture this being really helpful 😄

README.md
```diff
@@ -271,9 +271,7 @@ classes:
 
 The high-level flow looks like this:
 
-<p align="center">
-  <img src="https://s3.eu-west-2.amazonaws.com/dependabot-images/package-manager-architecture.svg" alt="Dependabot architecture">
-</p>
+![Dependabot high-level architecture diagram](architecture.svg)
```
landongrindheim, line 274: Was this added intentionally? Should it be a separate commit?

landongrindheim (approved):

README.md
jakecoffman: Oops, bad squash I guess. That was something I fixed this morning in another PR.

Issue #2337, opened 2020-07-19
Rust crate update versions with a plus in seem to be ignored


alecmocatta:
The dependency in Cargo.toml:
```toml
zstd = "0.4"
```
can currently be updated to 0.5. However it seems the version string, currently "0.5.3+zstd.1.4.5", breaks this:

```text
updater | INFO <job_40523263> Checking if zstd  needs updating
  proxy | 2020/07/17 06:59:20 GET https://crates.io:443/api/v1/crates/zstd
  proxy | 2020/07/17 06:59:20 200 https://crates.io:443/api/v1/crates/zstd
updater | INFO <job_40523263> Latest version is 0.5.3+zstd.1.4.5
  proxy | 2020/07/17 06:59:20 GET https://api.github.com:443/repos/rust-lang/crates.io-index/commits/master
  proxy | 2020/07/17 06:59:20 * authenticating github api request
  proxy | 2020/07/17 06:59:20 304 https://api.github.com:443/repos/rust-lang/crates.io-index/commits/master
updater | INFO <job_40523263> No update needed for zstd 
```

jeffwidman:
xref: https://github.com/dependabot/dependabot-core/pull/6183#issuecomment-1323978276

PR #6183, opened 2022-11-21
Allow cargo versions with hypens in build identifiers

deivid-rodriguez:
Cargo implements standard semantic versions as per https://doc.rust-lang.org/cargo/reference/manifest.html#the-version-field.

And standard semantic versioning allows hyphens in build identifiers as per https://semver.org/#spec-item-10.

### DIFF
--- a/cargo/lib/dependabot/cargo/version.rb
+++ b/cargo/lib/dependabot/cargo/version.rb
@@ -12,7 +12,7 @@ module Dependabot
     class Version < Gem::Version
       VERSION_PATTERN = '[0-9]+(?>\.[0-9a-zA-Z]+)*' \
                         '(-[0-9A-Za-z-]+(\.[0-9a-zA-Z-]+)*)?' \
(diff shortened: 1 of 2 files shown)


Nishnha (approved):

jeffwidman (approved):

jeffwidman:
Looks like `+` may also be supported? 
* https://github.com/dependabot/dependabot-core/issues/2337

Could include in this PR or break it out as a separate one...

jeffwidman:
Actually, looking at https://github.com/dependabot/dependabot-core/issues/2337 more, that looks like a combo version of rust crate + underlying packaged zstd version, so probably not something to include here... it's a different separator purpose.

deivid-rodriguez:
Actually the `+` operator is already parsed, this PR fixes parsing `+` operator with a right hand side including `-`.

#2337 seems not a parsing problem, but an issue in update logic.

PR #6189, opened 2022-11-22
Respect custom directory for actions updates

deivid-rodriguez:
While testing security updates for actions, I noticed that manual security update jobs are created with an explicit `directory` option passed with the `/.github/workflows` value.

But our updater currently does not support this.

Apparently, it used to, but that broke at some point.

Fixes https://github.com/dependabot/dependabot-core/issues/5047.

### DIFF
--- a/github_actions/lib/dependabot/github_actions/file_fetcher.rb
+++ b/github_actions/lib/dependabot/github_actions/file_fetcher.rb
@@ -41,16 +41,22 @@ module Dependabot
       def workflow_files
         return @workflow_files if defined? @workflow_files
 
-        @workflow_files = [fetch_file_if_present("action.yml"), fetch_file_if_present("action.yaml")].compact
+        @workflow_files = []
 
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
This won't fully restore the previous behavior.

Previously users could indicate `directory: /foo/bar` and get workflow files looked for at `/foo/bar/.github/workflows`. Now they'll need to indicate `directory: /foo/bar/.github/workflows`.

I could restore the previous behavior but then I have to special case both `/` and `/.github/workflows`. So I'll go with the current patch for now.

jeffwidman:
I think your current patch sounds fine, it's okay that they need to be more explicit. 

Honestly, we probably should have done that from the beginning because using `"/"` is confusing... it's just common enough as a recursive placeholder plus since it's a "magic" stand-in for `"/.github/workflows"` it just creates confusion like ☝️... I mean even the security updates are passing in `".github/workflows"` 😀 

Regardless, we can always tweak further later if needed.

PR #6191, opened 2022-11-22
Revert "Python: skip flaky poetry spec with issue comment"

jeffwidman:
The associated ticket on Poetry
(https://github.com/python-poetry/poetry/issues/3010) was closed because the underlying triggers were mostly resolved in `poetry` `1.2` using the new installer. In particular, see
https://github.com/python-poetry/poetry/issues/3010#issuecomment-1207683271.

We flipped to the new installer in
https://github.com/dependabot/dependabot-core/pull/5838, so this skip should no longer be necessary.

This reverts commit ceb57007bc2ca8c403e36bce7c4580a538366dc1.


deivid-rodriguez (approved):

jeffwidman:
I think the failure here might be because the test uses python 2, but we dropped support for python 2... so the test needs updating to use python 3.

Need to check if I should just update the `python_2.toml`/ `python_2.lock` fixtures completely, or if just update it to python 3... or re-use a different python 3 fixture.

python/spec/dependabot/python/file_updater/poetry_file_updater_spec.rb
```diff
@@ -80,11 +80,9 @@
         to start_with("8cea4ecb5b2230fbd4a33a67a4da004f1ccabad48352aaf040")
     end
 
-    # TODO: Fix flaky spec caused by an issue in poetry:
-    # https://github.com/python-poetry/poetry/issues/3010)
-    context "with a specified Python version" do
-      let(:pyproject_fixture_name) { "python_2.toml" }
-      let(:lockfile_fixture_name) { "python_2.lock" }
+    context "with a specified Python version", :slow do
```
jeffwidman, line 83: I annotated the test with `:slow` because it's explicitly testing using a version that isn't pre-installed...

jeffwidman: Actually, that's not the purpose of this test... this is just to test that poetry handles when a python version is specified... so I think we can use a pre-installed version w/o affecting the fidelity of this test?

jeffwidman: And actually over in https://github.com/dependabot/dependabot-core/pull/6079 we ignore the patch version, so it'll just use whatever 3.10 version is already pre-installed... 

Not sure if I should leave it at 3.10.7 as that's semi-realistic use case to ensure poetry handles it or switch to 3.10 to avoid confusing future test writers...

jeffwidman:
The remaining error is surfaced by the test, but not caused by the test:
https://github.com/dependabot/dependabot-core/actions/runs/3527394392/jobs/5916407220#step:6:1845

It is a bug that landed in https://github.com/dependabot/dependabot-core/pull/6079, Tom is working on a fix right now as this will impact production jobs.

python/spec/dependabot/python/file_updater/poetry_file_updater_spec.rb
pavera: I don't know what our threshold for `:slow` is, but the untar/unzip takes about 5s. I moved a couple other tests out of `:slow` with the pre-compiled python versions.

pavera (approved):

PR #6194, opened 2022-11-22
Re-enable two pending tests

jeffwidman:
Per the discussion in https://github.com/dependabot/dependabot-core/pull/3319#discussion_r599482825,
these only temporarily needed to be marked `pending` until
https://github.com/dependabot/dependabot-core/pull/3327 was merged.

From the first thread it looks like the intent was to remove the
`pending` marker, but that accidentally got overlooked.

So this removes the `pending` marker, and then fixes the failures:
1. The first was a straightforward change from method to hash value.
2. The second wasn't raising the expected error... However, this has
   been true since [this code was originally committed](https://github.com/dependabot/dependabot-core/pull/3319/files#r600526250)
   (once #3327 was merged to make the test work). So I deleted the test
   as it added no value.

### DIFF
--- a/bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
+++ b/bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
@@ -173,20 +173,8 @@ RSpec.describe Dependabot::Bundler::UpdateChecker::VersionResolver do
 
         let(:dependency_files) { bundler_project_dependency_files("requires_bundler") }
 
-        pending "resolves version" do
-          expect(subject.version).to eq(Gem::Version.new("3.0.0"))
-        end
-      end
-
-      context "when bundled with v2 and requesting a version that requires bundler v1", :bundler_v2_only do
-        let(:dependency_name) { "guard-bundler" }
-        let(:requirement_string) { "~> 2.2.0" }
-
-        let(:dependency_files) { bundler_project_dependency_files("requires_bundler") }
-
-        pending "raises a DependencyFileNotResolvable error" do
-          expect { subject }.
-            to raise_error(Dependabot::DependencyFileNotResolvable)
+        it "resolves version" do
+          expect(subject[:version]).to eq(Gem::Version.new("3.0.0"))
         end
       end
 


deivid-rodriguez:
@jeffwidman Not sure if this is useful? https://github.com/deivid-rodriguez/dependabot-core/commit/ec4f8b64151eb1da94df0f302c9056746be825d2

Apparently I worked on this when first contributing to dependabot-core, but never followed up on it.

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
```diff
@@ -184,9 +184,8 @@
 
         let(:dependency_files) { bundler_project_dependency_files("requires_bundler") }
 
-        pending "raises a DependencyFileNotResolvable error" do
-          expect { subject }.
-            to raise_error(Dependabot::DependencyFileNotResolvable)
```
jeffwidman, line 189: This hasn't raised since it was originally committed... it was only marked `pending` because waiting on a follow-on PR to land, but even then wasn't expected to raise per https://github.com/dependabot/dependabot-core/pull/3319/files#r600526250

Nishnha: this is a weird test since most dependencies have upgraded to use bundler v2, but since you got it to work I don't mind keeping it around

Nishnha (approved):
Approving since CI is passing. Tests lgtm

bundler/spec/dependabot/bundler/update_checker/version_resolver_spec.rb
jeffwidman: Yeah, the intent of the test makes sense to me--ensure no unexpected exceptions start getting thrown here... but I'm not certain the test setup is correct, ie, is the `guard-bundler ~> 2.2.0` pin actually requesting bundler 1? If so, we should keep this... if not, we should nuke this test case.

deivid-rodriguez: I don't see anything in there specifically pinning to Bundler 1, I'd say the spec can go.

jeffwidman: Nuked it.

Nishnha (approved):

PR #6199, opened 2022-11-23
Use git to fill in gemspec files

deivid-rodriguez:
This is the result of running `time bundle exec rubocop --version` on my machine with the code in main:

```
$ time bundle exec rubocop --version
1.39.0
bundle exec rubocop --version  9,92s user 0,73s system 96% cpu 11,075 total
```

10 seconds, wait, what??

This is the result with this branch:

```
$ time bundle exec rubocop --version
1.39.0
bundle exec rubocop --version  0,56s user 0,30s system 94% cpu 0,909 total
```

Half a second seems more sensitive 😅.

To figure out why this is, we can add a `puts` statement in one of the gemspecs:

```diff
diff --git a/bundler/dependabot-bundler.gemspec b/bundler/dependabot-bundler.gemspec
index d8fa4822e..147ec9e55 100644
--- a/bundler/dependabot-bundler.gemspec
+++ b/bundler/dependabot-bundler.gemspec
@@ -42,4 +42,7 @@ Gem::Specification.new do |spec|
       spec.files << path unless File.directory?(path)
     end
   end
+
+  puts spec.files.size
+  puts spec.files.take(10)
 end
```

And running the above command we can see this:

```
$ time bundle exec rubocop --version     
12742
helpers/v1/.bundle/bin/bundle
helpers/v1/.bundle/bin/htmldiff
helpers/v1/.bundle/bin/irb
helpers/v1/.bundle/bin/ldiff
helpers/v1/.bundle/bin/rdbg
helpers/v1/.bundle/bin/rspec
helpers/v1/.bundle/bundler/gems/business-a1b78a929dac/.git/FETCH_HEAD
helpers/v1/.bundle/bundler/gems/business-a1b78a929dac/.git/HEAD
helpers/v1/.bundle/bundler/gems/business-a1b78a929dac/.git/ORIG_HEAD
helpers/v1/.bundle/bundler/gems/business-a1b78a929dac/.git/config
1.39.0
bundle exec rubocop --version  9,81s user 0,68s system 98% cpu 10,616 total
```

We're adding 12K files to the gem, including a bunch of generated files 😬.

The problem is that we're tryin to implement the `.gitignore` syntax in Ruby using just a few lines of code and... well, it doesn't really work.

With this PR, we use `git` itself that obviously knows how to handle `.gitignore`, and we get the result we expect:

```
$ time bundle exec rubocop --version     
78
helpers/v1/.gitignore
helpers/v1/Gemfile
helpers/v1/build
helpers/v1/lib/functions.rb
helpers/v1/lib/functions/conflicting_dependency_resolver.rb
helpers/v1/lib/functions/dependency_source.rb
helpers/v1/lib/functions/file_parser.rb
helpers/v1/lib/functions/force_updater.rb
helpers/v1/lib/functions/lockfile_updater.rb
helpers/v1/lib/functions/version_resolver.rb
1.39.0
bundle exec rubocop --version  0,56s user 0,30s system 96% cpu 0,888 total
```

I don't think this has ever bitten us when releasing production gems because we build them on a clean CI environment when the tree that we release matches a fresh clone and has nothing generated.

But this should avoid future issues in addition to speeding up my dev env.

### DIFF
--- a/pub/dependabot-pub.gemspec
+++ b/pub/dependabot-pub.gemspec
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require "find"
-
 Gem::Specification.new do |spec|
   common_gemspec =
     Bundler.load_gemspec_uncached("../common/dependabot-common.gemspec")
--- a/bundler/dependabot-bundler.gemspec
+++ b/bundler/dependabot-bundler.gemspec
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require "find"
-
 Gem::Specification.new do |spec|
   common_gemspec =
     Bundler.load_gemspec_uncached("../common/dependabot-common.gemspec")
@@ -30,16 +28,5 @@ Gem::Specification.new do |spec|
 
   next unless File.exist?("../.gitignore")
 
-  ignores = File.readlines("../.gitignore").grep(/\S+/).map(&:chomp)
-
-  next unless File.directory?("lib")
-
-  prefix = "/" + File.basename(File.expand_path(__dir__)) + "/"
-  Find.find("lib", "helpers") do |path|
-    if ignores.any? { |i| File.fnmatch(i, prefix + path, File::FNM_DOTMATCH) }
-      Find.prune
-    else
-      spec.files << path unless File.directory?(path)
-    end
-  end
+  spec.files += `git -C #{__dir__} ls-files lib helpers -z`.split("\x0")
 end
--- a/cargo/dependabot-cargo.gemspec
+++ b/cargo/dependabot-cargo.gemspec
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require "find"
-
 Gem::Specification.new do |spec|
   common_gemspec =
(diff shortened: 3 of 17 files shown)


jurre (approved):

pub/dependabot-pub.gemspec
```diff
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require "find"
```
jeffwidman, line 3: I'm curious: Was this simply an unused import or was it necessary somehow?

If unused, I'm surprised we don't have a rubocop linter in place to catch this...

deivid-rodriguez: Yeah, unused.

I don't think RuboCop provides anything like that. I guess it could implement something based on convention, but it would have a lot of false positives. A `require` can bring in any top level constants or any other side effects, so static analysis does not seem like a good fit to detect this kind of thing.

jeffwidman: Gotcha. 

It's a common linter in the python and go ecosystems, both of which can have side effects as well, but there are conventions for signaling if you expect an import to have side effects.

PR #6227, opened 2022-11-28
Prevent `NoMethodError` when captures are missing

landongrindheim:
Co-authored-by: David Rodríguez <deivid.rodriguez@riseup.net>


Nishnha (approved):
Passes checks and lgtm

PR #6228, opened 2022-11-28
submodule cloning may not be important to the job, ignore errors

jakecoffman:
In #6172 I missed that having SCP-style URLs in the `.gitmodules` would have a different error message.

So this checks for both error messages and ignores so that the update can attempt to proceed in the event the submodules were not important to the update.

### DIFF
--- a/common/lib/dependabot/file_fetchers/base.rb
+++ b/common/lib/dependabot/file_fetchers/base.rb
@@ -26,11 +26,11 @@ module Dependabot
         Dependabot::Clients::CodeCommit::NotFound
       ].freeze
 
-      GIT_SUBMODULE_ERROR_REGEX1 =
+      GIT_SUBMODULE_INACCESSIBLE_ERROR =
         /^fatal: unable to access '(?<url>.*)': The requested URL returned error: (?<code>\d+)$/
-      GIT_SUBMODULE_ERROR_REGEX2 =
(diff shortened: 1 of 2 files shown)


common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -26,8 +26,11 @@ class Base
         Dependabot::Clients::CodeCommit::NotFound
       ].freeze
 
-      GIT_SUBMODULE_ERROR_REGEX =
+      GIT_SUBMODULE_ERROR_REGEX1 =
         /^fatal: unable to access '(?<url>.*)': The requested URL returned error: (?<code>\d+)$/
+      GIT_SUBMODULE_ERROR_REGEX2 =
+        /^fatal: clone of '(?<url>.*)' into submodule path '.*' failed$/
+      GIT_SUBMODULE_ERROR_REGEX = /(#{GIT_SUBMODULE_ERROR_REGEX1})|(#{GIT_SUBMODULE_ERROR_REGEX2})/
```
landongrindheim, line 33: I like you're approach here. Could we use constant names that reflect the underlying errors more directly? Maybe something like `GIT_SUBMODULE_INACCESSIBLE_ERROR` for `REGEX1` and `GIT_SUBMODULE_CLONE_ERROR` for `REGEX2`?

landongrindheim (approved):

PR #6229, opened 2022-11-28
NPM: skip blank "file:" references in path dependencies

jakecoffman:
In NPM when you add a path dependency it can result in a blank `"file:"` entry in the package-lock. The code here was taking that and calling `File.join` resulting in `/package.json` which later creates an EACCES error when it tries to write the package.json to the root directory.

It seems these blank `file:` sections are just for the parent dependency, so we can safely skip over them.


mctofu (approved):

jeffwidman:
Nice find!

PR #6240, opened 2022-12-01
Don't use compare URLs for deciding GitHub Actions upgrades

deivid-rodriguez:
When dealing with pinned SHAs, we have two cases:

* If the pinned SHA points to a version tag, or to the tip of a version branch, then we can use standard version logic to upgrade the commit sha. For example, if the SHA currently points to the v7 branch and there's a v10 branch, we should upgrade to the SHA of the tip of the v10 branch. Or if the SHA current points to the v1.1.0 tag, and there's a v3.1.0 tag, then we can upgrade to the commit SHA associated to the v3.1.0 tag.

* If the pinned SHA does not point to a version tag or branch, then we already clone the full repository to figure out which branch the current commit belongs to, and upgrade to the tip of that branch. So that logic can stay in place.

Using compare URLs does not seem to play nice with GitHub Actions, and I don't think we need it. The reason (I believe) is that sometimes release tags or branches are not tagged off the main stable branch, but can slightly diverge from it, so compare urls don't give us the results we want, but that does not mean users don't want upgrades.

With this change, a workflow pointing to the commit SHA of an old version of `tj-actions/changed-files` gets correct upgrades:

```
=== tj-actions/changed-files (14)
 => checking for updates 4/4
🌍 https://github.com/tj-actions/changed-files.git/info/refs
 => latest available version is 34
 => latest allowed version is 34
 => updating tj-actions/changed-files from 14 to 34

    ± .github/workflows/howie-on-issues.yml
    ~~~
    48c48
    <       - uses: tj-actions/changed-files@b59431bc7d44f9e8951a290fc7d48879f2ca1939
    ---
    >       - uses: tj-actions/changed-files@8a4cc4fbd67975557b6d85dd302f5f9400b9c92e
    ~~~
🌍 Total requests made: '8'
```

With this change, the `actions/checkout` example given at https://github.com/dependabot/dependabot-core/pull/6239 starts getting upgrades, but gets incorrect upgrades, because it interprets 3.1.2 as the latest version of the action and tries to upgrade to the tip of [the dependabot/npm_and_yarn/minimatch-3.1.2 branch](https://github.com/actions/checkout/tree/dependabot/npm_and_yarn/minimatch-3.1.2):

```
=== actions/checkout (1.1.0)
 => checking for updates 2/4
🌍 https://github.com/actions/checkout.git/info/refs
 => latest available version is 3.1.2
 => latest allowed version is 3.1.2
 => updating actions/checkout from 1.1.0 to 3.1.2

    ± .github/workflows/howie-on-issues.yml
    ~~~
    30c30
    <       - uses: actions/checkout@0b496e91ec7ae4428c3ed2eeb4c3a40df431f2cc
    ---
    >       - uses: actions/checkout@91a9694ef96c0b41986a5fe0dd9209c21fd54f6f
    ~~~
```

So this PR should be deployed after #6239 so that Dependabot does not start proposing incorrect upgrades in cases like the above.


pavera (approved):

github_actions/spec/dependabot/github_actions/update_checker_spec.rb
```diff
@@ -138,39 +138,50 @@
...
-              headers: { "Content-Type" => "application/json" }
-            )
+        it { is_expected.to be_falsey }
+      end
+
+      context "that is a git commit SHA pointing to the tip of a branch named like a version" do
+        let(:upload_pack_fixture) { "run-vcpkg" }
+
+        context "and there's a branch named like a higher version" do
+          let(:tip_of_v6) { "205a4bde2b6ddf941a102fb50320ea1aa9338233" }
```
jeffwidman, line 153: Thanks, the addition of these `tip_of_vN` vars make the tests so much more readable!

PR #6241, opened 2022-12-01
NPM: do not switch dependency source from git to public registry

jakecoffman:
When a dependency is pinned to a git ref that is a version pattern, and there is a newer version of the dependency with the same name in the NPM registry, Dependabot would create a PR that switched the source from git to the registry.

This could potentially lead to a name-squatting/Dependency confusion attack where an attacker publishes a package of the same name. This is somewhat unlikely since using a package source of git without being published would be a fairly rare occurrence.

We weren't sure what the usefulness of this feature was anyway so I went with removing the functionality entirely.


deivid-rodriguez (approved):
Funny how I'm also removing calls to `git_commit_checker.branch_or_ref_in_release?` at #6240 because they also happen to be problematic in the Actions ecosystem.

Thanks for fixing this!

npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker_spec.rb
```diff
@@ -471,26 +471,6 @@
           end
         end
 
-        context "that is behind the latest release" do
-          let(:commit_compare_response) do
-            fixture("github", "commit_compare_behind.json")
```
deivid-rodriguez, line 476: This fixture can now probably be removed too?

npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker_spec.rb
```diff
@@ -471,26 +471,6 @@
...
-          end
-
-          it "updates to the latest release" do
-            expect(checker.latest_version).to eq(Gem::Version.new("4.0.0"))
-          end
-
-          context "when the registry doesn't return a latest release" do
-            let(:registry_response) do
-              fixture("npm_responses", "no_latest.json")
```
deivid-rodriguez, line 485: And this one can be removed too?

jeffwidman:
It's a fairly common use case where someone needs to pickup some fix that isn't release yet, so they'll temp pin to a SHA, but want to revert back to to the latest version once it includes the fix.

I've done this myself multiple times for deps in the go, python, and github-actions ecosystems, and it's what we've even been recommending folks do to pickup the changes here in `dependabot-core` until we get a process in place to release more frequently.

jeffwidman (approved):
I hate to see the convenience of this disappear, but do agree the security risk trumps the convenience.

Perhaps down the road we could find a different way to detect brandjacking such as  check the package history to see if it's been present on the registry since before the version commit was made, but implementing that will get complex very quickly so not a rabbit hole to go down right now...

Issue #6076, opened 2022-11-08
Yarn Berry Cache not included in dependabot PRs where cache exists


wozzo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

yarn

### Package manager version

3.2.x

### What you expected to see, versus what you actually saw

Yarn's offline cache is not updated by dependabot when using the `nodeLinker: node-modules` setting from yarn.

This relates to [PR #6017](https://github.com/dependabot/dependabot-core/pull/6017) and [my comment](https://github.com/dependabot/dependabot-core/pull/6017#issuecomment-1303802118).

We are using [Yarn's offline cache](https://yarnpkg.com/features/offline-cache/#gatsby-focus-wrapper), so a zip file is stored for each dependency.
When updating the a dependency the old zip file is typically deleted and a new one added to the repository in the cache folder (default `.yarn/cache`).

However this fix assumed that the offline cache was always used with zero-installs and checks for the existence of the `.pnp.cjs` file to determine whether to update the cache.
The existence of this file relates to the [`nodeLinker` option's value](https://yarnpkg.com/configuration/yarnrc#nodeLinker)
We use `node-modules`, not `pnp` so the above file does not exist, but we still maintain an offline cache at `.yarn/cache`.
The location of the offline cache is configured with the [`cacheFolder` option](https://yarnpkg.com/configuration/yarnrc#cacheFolder).

I think dependabot should check for the existence of the cache folder in the repository instead of the `.pnp.cjs` file as the determining factor in whether to update the cache or not.

pavera:
The issue I ran into when we used the existence of the cache folder as a determinant was that on repositories which do not have that folder committed, once we run an update the cache folder is created and so suddenly we detect that we should be updating the cache, which is not what the user wants in the case where the cache folder is not committed. I'll be revisiting this in the next few days as the PR you mentioned also did not fix the permissions change on `.pnp.cjs`. Thanks for the heads up here, I think we might need to inspect `.gitignore` or maintain state about the original state of the repo so that changes induced by Dependabot don't change behavior mid-update.

wozzo:
@pavera has there been any progress on this?

pavera:
@wozzo unfortunately no, it's now next on my plate, I should be starting work on it tomorrow.

pavera:
[...]

PR #6242, opened 2022-12-01
Yarn Berry: Offline cache updates

pavera:
## Context
Reported here: https://github.com/dependabot/dependabot-core/issues/6076. We are not updating the yarn cache if the repo is not configured for zero install, there is a third mode of operation that uses the cache but not zero install.

## Approach
Here I attempt to detect a non-zero install but offline cache enabled repo by inspecting the yarn config looking for `node-modules` and a checked in `cacheFolder`. I've manually tested with the CLI and this appears to correctly manage the cache. I've added unit tests to confirm that a) if it is a offline cache repo we correctly manage the cache, as in we run with `--mode=skip-build` so we remove items from the `cacheFolder` b) Confirm we use the correct mode `--mode=update-lockfile` in the absence of a checked in cache dir


npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
```diff
@@ -40,12 +40,19 @@ def self.yarn_zero_install?
...
+
       def self.yarn_berry_args
         if yarn_major_version == 2
           ""
-        elsif yarn_major_version >= 3 && yarn_zero_install?
+        elsif yarn_major_version >= 3 && (yarn_zero_install? || yarn_offline_cache?)
           "--mode=skip-build"
         else
+          # We only want this mode if the cache is not being updated/managed
+          # as this improperly leaves old versions in the cache
```
landongrindheim, line 55: Thanks for this context 🙇

jurre (approved):

landongrindheim (approved):

Issue #5276, opened 2022-06-16
Dependabot NOT creating pull requests for parent pom version.


sri-al-cc:
Dependabot is NOT creating pull requests for parent pom version.

We are using maven 3.X as package manager. Our project uses a parent pom  version as below
     <parent>
         <groupId></groupId>
         <artifactId>xxxxx-base-java-webapp</artifactId>
         <version>2022.1.0</version>
		 <type>pom</type>
    </parent>

We have an several next versions of parent poms in nexus repository, but dependabot thinks an update is not needed. However it does work and create Pull requests  for regular dependencies specified with
    <dependencies>
    </dependencies>

dependabot.yml

version: 2
updates:
- package-ecosystem: maven
  directory: "/xxxx"
  target-branch: "develop"
  open-pull-requests-limit: 10

Log file
2022-06-14T18:42:49.0954590Z Parsing dependencies information
2022-06-14T18:42:49.1031289Z Checking if xxxxx-base-java-webapp  needs updating
2022-06-14T18:42:49.1031933Z No update needed for xxxxx-base-java-webapp

jeffwidman:
I'm not 100% positive, but I _think_ this is resolved by a combo of https://github.com/dependabot/dependabot-core/pull/6491 and https://github.com/dependabot/dependabot-core/pull/6248, so closing. If for some reason I'm misunderstanding, please comment and we can re-open.

Issue #4425, opened 2021-11-18
Support maven pom filenames other than pom.xml


ryandcarter:
Some Maven projects use non-standard pom file names such as bom.xml and parent-pom.xml. Is it possible to add an override of the top-level filename to look for?

jurre:
We currently don't have a good way to inject such overrides, so adding it would be a fair bit of work since we'd have to work out how to set that up. We likely won't be able to tackle it soon.

It would be super helpful if you could provide some context as to why projects would customize the filename, how common the approach is within the community and maybe link to some example projects that are set up that way. That info will help us build the right feature once we do get to it, and will also help find the right prioritization for it.

yeikel:
This definitely sounds very uncommon and I am also curious to hear why you're doing this

filip26:
using `pom_parent.xml' is very common, especially when using [project inheritance](https://maven.apache.org/guides/introduction/introduction-to-the-pom.html#project-inheritance).  

For real life example see https://github.com/filip26/titanium-json-ld

jeffwidman:
Was this fixed by https://github.com/dependabot/dependabot-core/pull/6491? 
cc @soulus13

soulus13:
@jeffwidman , partially. It has been fixed for the inheritance/aggregation cases, but the entry point for everything still must be a file named pom.xml. 

As a workaround, though, it should be possible to create a dummy root pom.xml file, which will then point inside to the non-standard named parent/children modules (and that is what we plan to do in our repo).

filip26:
@soulus13 dummy pom.xml is not a workaround, it's like recommending that you can create a dummy `package.json` anytime you want to check dependencies and then when you want to build it, replace the file with the correct one .... 
The easiest, and most transparent way, is to allow define the entry points in a workflow.

soulus13:
@filip26 sure, fully agree with that. Still, the PR mentioned above has been solving slightly different problem, to allow custom names for parent and child modules. By doing that it also allowed for such a workaround for this case. 

Allowing for a different root pom name, at least from what I see - requires more effort and would affect not only the internal implementation details, but would also require to provide a way to specify this custom name when configuring the Dependabot for the repo. Other option - scanning all .xml files in a folder to check if they are pom files or not - does not look like a viable option to me.

PR #6248, opened 2022-12-02
Add support for custom pom.xml names for the child maven modules

soulus13:
Extended on the multi-module review support, allowing for the custom pom.xml names be used for the pom-s in child modules. With that PR they would be able to have various names, with the only limitation being that they must end with .xml. 
This PR should, hopefully, address a part of the https://github.com/dependabot/dependabot-core/issues/4425 issue.

### DIFF
--- a/maven/lib/dependabot/maven/file_fetcher.rb
+++ b/maven/lib/dependabot/maven/file_fetcher.rb
@@ -11,7 +11,7 @@ module Dependabot
                         "profile > modules > module"
 
       def self.required_files_in?(filenames)
-        (%w(pom.xml) - filenames).empty?
+        filenames.include?("pom.xml")
       end
 
       def self.required_files_message
@@ -66,7 +66,7 @@ module Dependabot
           name_parts = [
             base_path,
             relative_path,
-            relative_path.end_with?("pom.xml") ? nil : "pom.xml"
+            relative_path.end_with?(".xml") ? nil : "pom.xml"
           ].compact.reject(&:empty?)
           path = Pathname.new(File.join(*name_parts)).cleanpath.to_path
 
@@ -126,7 +126,7 @@ module Dependabot
         name_parts = [
           File.dirname(pom.name),
           relative_parent_path,
-          relative_parent_path.end_with?("pom.xml", "pom_parent.xml") ? nil : "pom.xml"
+          relative_parent_path.end_with?(".xml") ? nil : "pom.xml"
         ].compact.reject(&:empty?)
 
         Pathname.new(File.join(*name_parts)).cleanpath.to_path
--- a/maven/lib/dependabot/maven/file_parser.rb
+++ b/maven/lib/dependabot/maven/file_parser.rb
@@ -272,7 +272,9 @@ module Dependabot
 
       def pomfiles
         @pomfiles ||=
-          dependency_files.select { |f| f.name.end_with?("pom.xml", "pom_parent.xml") }
+          dependency_files.select do |f|
+            f.name.end_with?(".xml") && !f.name.end_with?("extensions.xml")
+          end
       end
 
       def extensionfiles
--- a/maven/lib/dependabot/maven/file_updater.rb
+++ b/maven/lib/dependabot/maven/file_updater.rb
@@ -13,6 +13,7 @@ module Dependabot
       def self.updated_files_regex
         [
           /^pom\.xml$/, %r{/pom\.xml$},
+          /.*\.xml$/, %r{/.*\.xml$},
           /^extensions.\.xml$/, %r{/extensions\.xml$}
         ]
       end
@@ -31,7 +32,7 @@ module Dependabot
           )
         end
 
-        updated_files.select! { |f| f.name.end_with?("pom.xml", "pom_parent.xml", "extensions.xml") }
+        updated_files.select! { |f| f.name.end_with?(".xml") }
         updated_files.reject! { |f| dependency_files.include?(f) }
 
         raise "No files changed!" if updated_files.none?
--- a/maven/spec/dependabot/maven/file_fetcher_spec.rb
+++ b/maven/spec/dependabot/maven/file_fetcher_spec.rb
@@ -29,6 +29,40 @@ RSpec.describe Dependabot::Maven::FileFetcher do
     }]
   end
 
+  describe ".required_files_in?" do
+    subject { described_class.required_files_in?(filenames) }
+
+    context "with only a pom.xml" do
+      let(:filenames) { %w(pom.xml) }
+      it { is_expected.to eq(true) }
+    end
+
+    context "with pom.xml and any other valid .xml" do
+      let(:filenames) { %w(pom.xml othermodule.xml) }
+      it { is_expected.to eq(true) }
+    end
+
+    context "with only an extensions.xml" do
+      let(:filenames) { %w(extensions.xml) }
+      it { is_expected.to eq(false) }
+    end
+
+    context "with an extensions.xml and a valid pom.xml file" do
+      let(:filenames) { %w(extensions.xml pom.xml) }
+      it { is_expected.to eq(true) }
+    end
+
+    context "with a non .xml file" do
+      let(:filenames) { %w(nonxml.txt) }
+      it { is_expected.to eq(false) }
+    end
+
+    context "with no files passed" do
+      let(:filenames) { %w() }
+      it { is_expected.to eq(false) }
+    end
+  end
+
   before do
     allow(file_fetcher_instance).to receive(:commit).and_return("sha")
 
@@ -323,6 +357,90 @@ RSpec.describe Dependabot::Maven::FileFetcher do
           ).once
         end
       end
+    end
+  end
+
+  context "with a multimodule custom named child poms" do
+    before do
+      stub_request(:get, File.join(url, "pom.xml?ref=sha")).
+        with(headers: { "Authorization" => "token token" }).
+        to_return(
+          status: 200,
+          body: fixture("github", "contents_java_multimodule_different_pom_names.json"),
+          headers: { "content-type" => "application/json" }
+        )
+
+      stub_request(:get, File.join(url, "submodule-one/pom.xml?ref=sha")).
+        with(headers: { "Authorization" => "token token" }).
+        to_return(
+          status: 200,
+          body: fixture("github", "contents_java_basic_pom.json"),
+          headers: { "content-type" => "application/json" }
+        )
+
+      stub_request(:get, File.join(url, "submodule-three/some-other-name.xml?ref=sha")).
+        with(headers: { "Authorization" => "token token" }).
+        to_return(
+          status: 200,
+          body: fixture("github", "contents_java_basic_pom.json"),
+          headers: { "content-type" => "application/json" }
+        )
+
+      stub_request(:get, File.join(url, "submodule-two/notpom.xml?ref=sha")).
+        with(headers: { "Authorization" => "token token" }).
+        to_return(
+          status: 200,
+          body: fixture("github", "contents_java_basic_pom.json"),
+          headers: { "content-type" => "application/json" }
+        )
+
+      stub_request(:get, File.join(url, ".mvn/extensions.xml?ref=sha")).
+        with(headers: { "Authorization" => "token token" }).
+        to_return(
+          status: 404
+        )
+    end
+
+    it "fetches the child poms with different names" do
+      expect(file_fetcher_instance.files.count).to eq(4)
+      expect(file_fetcher_instance.files.map(&:name)).
+        to match_array(
+          %w(pom.xml submodule-one/pom.xml submodule-three/some-other-name.xml submodule-two/notpom.xml)
+        )
+    end
+  end
+
+  context "with a parent pom with custom name" do
+    before do
+      stub_request(:get, File.join(url, "pom.xml?ref=sha")).
+        with(headers: { "Authorization" => "token token" }).
+        to_return(
+          status: 200,
+          body: fixture("github", "contents_java_parent_pom_custom_name.json"),
+          headers: { "content-type" => "application/json" }
+        )
+
+      stub_request(:get, File.join(url, "parentpom.xml?ref=sha")).
+        with(headers: { "Authorization" => "token token" }).
+        to_return(
+          status: 200,
+          body: fixture("github", "contents_java_basic_pom.json"),
+          headers: { "content-type" => "application/json" }
+        )
+
+      stub_request(:get, File.join(url, ".mvn/extensions.xml?ref=sha")).
+        with(headers: { "Authorization" => "token token" }).
+        to_return(
+          status: 404
+        )
+    end
+
+    it "fetches the parent pom" do
+      expect(file_fetcher_instance.files.count).to eq(2)
+      expect(file_fetcher_instance.files.map(&:name)).
+        to match_array(
+          %w(pom.xml parentpom.xml)
+        )
     end
   end
 end
--- a/maven/spec/dependabot/maven/file_parser_spec.rb
+++ b/maven/spec/dependabot/maven/file_parser_spec.rb
@@ -765,5 +765,227 @@ RSpec.describe Dependabot::Maven::FileParser do
         end
       end
     end
+
+    context "with a multimodule custom named child poms" do
+      let(:files) do
+        [
+          multimodule_custom_pom, submodule_one_pom, submodule_two_pom, submodule_three_pom
+        ]
+      end
+      let(:multimodule_custom_pom) do
+        Dependabot::DependencyFile.new(
+          name: "pom.xml",
+          content: fixture("poms", "multimodule_custom_modules.xml")
+        )
+      end
+      let(:submodule_one_pom) do
+        Dependabot::DependencyFile.new(
+          name: "submodule-one/pom.xml",
+          content: fixture("poms", "multimodule_custom_modules_submodule_one_pom.xml")
+        )
+      end
+      let(:submodule_two_pom) do
+        Dependabot::DependencyFile.new(
+          name: "submodule-two/notpom.xml",
+          content: fixture("poms", "multimodule_custom_modules_submodule_two_pom.xml")
+        )
+      end
+      let(:submodule_three_pom) do
+        Dependabot::DependencyFile.new(
+          name: "submodule-three/some-other-name.xml",
+          content: fixture("poms", "multimodule_custom_modules_submodule_three_pom.xml")
+        )
+      end
+
+      it "gets the right dependencies" do
+        expect(dependencies.map(&:name)).
+          to match_array(
+            %w(
+              net.sf.ehcache:ehcache
+              org.apache.httpcomponents:httpclient
+              org.springframework:spring-aop
+              org.springframework:spring-core
+            )
+          )
+      end
+
+      describe "the standard pom dependency" do
+        subject(:dependency) { dependencies.first }
+
+        it "has the right details" do
+          expect(dependency).to be_a(Dependabot::Dependency)
+          expect(dependency.name).
+            to eq("org.apache.httpcomponents:httpclient")
+          expect(dependency.version).to eq("4.0")
+          expect(dependency.requirements).to eq(
+            [{
+              requirement: "4.0",
+              file: "submodule-one/pom.xml",
+              groups: [],
+              source: nil,
+              metadata: {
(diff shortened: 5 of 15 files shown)


soulus13:
Fixed the codestyle issues, now should be all green

soulus13:
Hopefully now that CI has been fixed - everything will pass 😃

deivid-rodriguez (commented):
This sounds like an improvement over what we have, and it seems we're already special casing `parent_pom.xml` and `extensions.xml` anyways.

Can you elaborate a bit on the exceptions? The idea is that those two will never be top-level pom's but only included from other pom's?(Sorry if I use wrong terminology, I know nothing about maven). Could there be other exceptions (say, `parent-pom.xml`)? How would we handle those now, and how would we handle them after this PR?

soulus13:
> This sounds like an improvement over what we have, and it seems we're already special casing `parent_pom.xml` and `extensions.xml` anyways.
> 
> Can you elaborate a bit on the exceptions? The idea is that those two will never be top-level pom's but only included from other pom's?(Sorry if I use wrong terminology, I know nothing about maven). Could there be other exceptions (say, `parent-pom.xml`)? How would we handle those now, and how would we handle them after this PR?

Hi David, for sure.

With `extensions.xml` it is easy. Extensions are used to affect, in some way, the artifact/app build process or how maven itself work. There are core extensions which are loaded early and build extensions which are loaded late (https://maven.apache.org/guides/mini/guide-using-extensions.html#loading-extensions):
 - Core extensions: defined in `extensions.xml` file, stored under the `${maven.projectBasedir}/.mvn/` directory (as per the convention), allows to define relevant such extensions for your project. Their dependencies are currently gathered within the separate `extensionfiles` property, and then checked separately from the pom ones.
 - Build-time extensions: defined in `pom.xml`, mostly in the `build->extensions` path, or in plugins. Both places are already checked (they are included in `DEPENDENCY_SELECTOR`)

Now compared to that, the paths that we add within the `modules` part of the `pom.xml` file define the paths to the child pom modules with their own dependencies. There is also a convention there (https://maven.apache.org/pom.html#aggregation-or-multi-module), that if the path for a module is a directory - it will lookup the `pom.xml` file inside. But if you specify a path to the file - it will assume that within that file you have your pom structure defined, no matter what name it has (and that is the part added within this PR). Which, in reality, means, that you could, theoretically, specify that you child module is named `my-child-module/extension.xml`, and this would even properly work with maven... but would bring a lot of confusion overall, as this kind-of mixes two different concepts, and also would be harder to properly support within the dependabot architecture (i.e. Dependabot::Maven::FileFetcher.required_files_in should check, that we have some proper pom files inside, not only `extensions` files), thus they have been excluded from the places where we expect pom files only. 

Switching to the `pom_parent.xml` file - it is not a standard name in maven and is not mentioned anywhere in the documentation there. From what i get from the code (might also be wrong) is the following:
 - `pom_parent.xml` has been used just as a "constant" to "label" a local pom file which has been retrieved as a parent of the base POM file (and then recursively repeated for each fetched parent) 
 - in order for the parent pom to be retrieved by the dependabot - it has to be named `pom.xml` as well, so no other names are really supported there, even the `pom_parent.xml` 😃 
 - these files, for the reasons unknown to me, have been assumed to be "read-only" for dependabot purposes, which results in the following:
   - these files are not checked for the vulnerable dependencies, consequently should not be updated by dependabot in any way
   - but the information from these files is still used to populate the repositories from which to fetch the dependencies, as well as to populate the available properties files (from which a dependency version is often taken)
  
So now getting to the question on what changes after the PR for those two cases the answer is simple - nothing 😃 Just to sum it up (as my text above is a bit too long):
 - Core extensions have already been properly handled and nothing to be changed here
 - Build-time extensions are properly handled as well, nothing changes here
 - Parent poms, if they are local poms, and not an artefact from some repo are not being updated on purpose (even though i don't know the reasoning behind this decision), and this also stays as is
 - Thus the only thing that is changing with this PR is the way child pom files are handled, in order to allow having there wider variety of supported pom file names based on the maven specification

deivid-rodriguez:
Thanks for the explanation @soulus13, very helpful!

I had a closer look here. Made some bare testing and this seems to work as expected. Also the implementation looks good 🎉.

I think this PR will fix https://github.com/dependabot/dependabot-core/issues/5276 too?

I did notice one difference, on a repo using `parent_pom.xml`.

With main branch:

```
$ bin/dry-run.rb maven filip26/titanium-json-ld
=> fetching dependency files
=> dumping fetched dependency files: ./dry-run/filip26/titanium-json-ld/
=> parsing dependency files
=> updating 1 dependencies: com.apicatalog:titanium

=== com.apicatalog:titanium (1.3.2-SNAPSHOT)
 => checking for updates 1/1
🌍 --> GET https://repo.maven.apache.org/maven2/com/apicatalog/titanium/1.3.2-SNAPSHOT/maven-metadata.xml
🌍 <-- 404 https://repo.maven.apache.org:443/maven2/com/apicatalog/titanium/1.3.2-SNAPSHOT/maven-metadata.xml
🌍 --> GET https://repo.maven.apache.org/maven2/com/apicatalog/titanium/maven-metadata.xml
🌍 <-- 404 https://repo.maven.apache.org:443/maven2/com/apicatalog/titanium/maven-metadata.xml
 => latest available version is 
 => latest allowed version is 1.3.2-SNAPSHOT
    (no update possible 🙅‍♀️)
🌍 Total requests made: '2'
```

With this PR:

```
$ bin/dry-run.rb maven filip26/titanium-json-ld
=> fetching dependency files
=> dumping fetched dependency files: ./dry-run/filip26/titanium-json-ld/
=> parsing dependency files
=> updating 0 dependencies: 
🌍 Total requests made: '0'
```

The new behavior seems better since we'll never find any updates for locally pulled dependencies anyways.

I also looked into why we're using the `pom_parent.xml` trick, and found the PR but not much of an explanation: https://github.com/dependabot/dependabot-core/pull/356. However, it seems like the current implementation relies on this particular name to be excluded and it would be a non trivial refactoring to get rid of it even if we deemed it as not necessary. 

I guess we could rename the "magic file" to something less common like `pom_parent_just_a_trick_to_properly_parse_parent_poms_recursively.xml` 😆. Anyways, not for this PR (or for any PR, probably, but just an idea that came to mind).

Do you have a particular public repository to run more tests against this PR? It looks mostly ready but I'd like it to give it another pass.

maven/lib/dependabot/maven/file_fetcher.rb
```diff
@@ -11,7 +11,9 @@ class FileFetcher < Dependabot::FileFetchers::Base
                         "profile > modules > module"
 
       def self.required_files_in?(filenames)
-        (%w(pom.xml) - filenames).empty?
+        !filenames.empty? && filenames.all? do |f|
+          f.name.end_with?(".xml") && !f.name.end_with?("pom_parent.xml", "extensions.xml")
+        end
```
deivid-rodriguez, line 16: This one looks incorrect I think, because from the previous implementation I think we should receive an array of `String`, not an array of `DependencyFile`. Can you check? I guess `filenames.any? {|f| f.end_with?(".xml") && f.end_with?("pom_parent.xml", "extensions.xml") }` would do?

If it is indeed incorrect, can we cover the fix with a spec?

soulus13: Thanks for spotting that, @deivid-rodriguez, fixed & added some proper specs for it! And yep, the line you have provided surely worked fine, as we need at least one file to be a proper pom file, and not an extension or a parent. 
One thing i have spotted though... Looks like the maven implementation of the `FileFetcher` don't really invoke this method anywhere, or use it to really validate the fetched files, otherwise the previous implementation would have failed as well, as it required all the files to be only pom.xml, whereas `fetch_files` could already produce and worked with `extensions.xml` dependencies 😃 Guess `required_files_in?` has to be ideally added to the `fetch_files` method, like in some other languages' implementations of `FileFetcher` (but then again, looks like maven is not the only one missing that check 😃 )

deivid-rodriguez: Nice catch! It seems only Bundler uses this 😬. I guess we can leave that for another time, I opened a ticket about it: https://github.com/dependabot/dependabot-core/issues/6487.

soulus13: Now this comment that you left, @deivid-rodriguez, has really made me think a lot on how that works again, and study a bit deeper the overs logic there, and now i'm ready to discuss that further on!

> I think this PR will fix #5276 too?
Yes and no, at the moment 😃 Will get to this part down the line

> The new behavior seems better since we'll never find any updates for locally pulled dependencies anyways.
Correct, and this is due to now properly searching for the parent files as well, which i have totally forgotten since the time i made it. So my statement in the comment above was a bit incorrect - this PR indeed adds a logic to allow both alternative names for the child modules and for the parent ones, as i recall that fixing just one part seemed not right to me.

> I also looked into why we're using the `pom_parent.xml` trick, and found the PR but not much of an explanation: #356. However, it seems like the current implementation relies on this particular name to be excluded and it would be a non trivial refactoring to get rid of it even if we deemed it as not necessary.
And that is the main question left, if we still need this exclude. I would really assume that for any `pom.xml` file we fetch, no matter parent it is or not - we generally have to check its dependencies and bump them, if possible. 
But now we come to a fun part 😃 Even though i have tried to preserve the non-update-parent logic (as i did not know if that was for a purpose)  - it now allows parent files to slip through it, in case they have custom names and are not named `pom_parent.xml`. I have updated my test repo with some additional pom-s to demonstrate it, so feel free to dry run:
`bin/dry-run.rb maven soulus13/dependabot-maven-test`
There you will see that 2 out of 4 parent poms i have created there have been bumped, and 2 have not been. 

So now it would be great if you could, somehow, dig up why were the parent poms blocked from the dependency checker in the first place, if that is possible, for sure! And anyway, depending on what you think would be the best option - i can either go ahead and completely remove the parent pom exclusion from dependency checking (and only then would the #5276 be fully addressed), or i can adjust a bit the logic within Dependabot::Maven::FileFetcher#recursively_fetch_relative_path_parents to properly substitute the `parent_pom.name` with our "magic file" name, to fall back to the previous behavior for parent poms. My preference would be, for sure, to remove all that "magic" from the code and keep parent poms updated as well, as requested in #5276

> Do you have a particular public repository to run more tests against this PR? It looks mostly ready but I'd like it to give it another pass.
I have been mostly playing with my custom repo above, where i have added all the scenarios i could think of, but here is another real-life sample i have found with custom submodule names: https://github.com/EsotericSoftware/kryo. And for the parent custom module name - https://github.com/filip26/titanium-json-ld is still good enough. Especially if you do locally bulk replace of `pom_parent.xml` to some other magic name, just in case.

soulus13:
Now this comment that you left, @deivid-rodriguez, has really made me think a lot on how that works again, and study a bit deeper the overs logic there, and now i'm ready to discuss that further on!

> I think this PR will fix #5276 too?
Yes and no, at the moment 😃 Will get to this part down the line

> The new behavior seems better since we'll never find any updates for locally pulled dependencies anyways.
Correct, and this is due to now properly searching for the parent files as well, which i have totally forgotten since the time i made it. So my statement in the comment above was a bit incorrect - this PR indeed adds a logic to allow both alternative names for the child modules and for the parent ones, as i recall that fixing just one part seemed not right to me.

> I also looked into why we're using the `pom_parent.xml` trick, and found the PR but not much of an explanation: #356. However, it seems like the current implementation relies on this particular name to be excluded and it would be a non trivial refactoring to get rid of it even if we deemed it as not necessary.
And that is the main question left, if we still need this exclude. I would really assume that for any `pom.xml` file we fetch, no matter parent it is or not - we generally have to check its dependencies and bump them, if possible. 
But now we come to a fun part 😃 Even though i have tried to preserve the non-update-parent logic (as i did not know if that was for a purpose)  - it now allows parent files to slip through it, in case they have custom names and are not named `pom_parent.xml`. I have updated my test repo with some additional pom-s to demonstrate it, so feel free to dry run:
`bin/dry-run.rb maven soulus13/dependabot-maven-test`
There you will see that 2 out of 4 parent poms i have created there have been bumped, and 2 have not been. 

So now it would be great if you could, somehow, dig up why were the parent poms blocked from the dependency checker in the first place, if that is possible, for sure! And anyway, depending on what you think would be the best option - i can either go ahead and completely remove the parent pom exclusion from dependency checking (and only then would the #5276 be fully addressed), or i can adjust a bit the logic within Dependabot::Maven::FileFetcher#recursively_fetch_relative_path_parents to properly substitute the `parent_pom.name` with our "magic file" name, to fall back to the previous behavior for parent poms. My preference would be, for sure, to remove all that "magic" from the code and keep parent poms updated as well, as requested in #5276

> Do you have a particular public repository to run more tests against this PR? It looks mostly ready but I'd like it to give it another pass.
I have been mostly playing with my custom repo above, where i have added all the scenarios i could think of, but here is another real-life sample i have found with custom submodule names: https://github.com/EsotericSoftware/kryo. And for the parent custom module name - https://github.com/filip26/titanium-json-ld is still good enough. Especially if you do locally bulk replace of `pom_parent.xml` to some other magic name, just in case.

deivid-rodriguez:
That sounds like a good plan. Let me first remove the `pom_parent.xml` magic and start supporting the first half of this. We'll support `pom.xml` and `pom_parent.xml` after this first step.

Then you can rebase this PR and add support for generic names.

soulus13:
> That sounds like a good plan. Let me first remove the `pom_parent.xml` magic and start supporting the first half of this. We'll support `pom.xml` and `pom_parent.xml` after this first step.
> 
> Then you can rebase this PR and add support for generic names.

Sounds perfect for me, @deivid-rodriguez! If you would like any support from me in this - always willing to contribute into these improvements with whatever i can 😃

deivid-rodriguez:
Thanks! I tried to patch this locally and it seems fairly simple. I'll open a PR.

I have one more question though. We have learnt so far, and this PR adds support for updating:

* Directly linked child poms with arbitrary names via `<module>custom/childpom.xml</module>`.
* Directly linked parent moms with arbitrary names via `<parent><relativePath>custom/parentpom.xml</relativePath>`.

Does that maven support a "root pom" (not linked from any other file) not named `pom.xml`. If not, does that mean we still need a file named `pom.xml` to have a sane maven setup?

soulus13:
> Thanks! I tried to patch this locally and it seems fairly simple. I'll open a PR.
> 
> I have one more question though. We have learnt so far, and this PR adds support for updating:
> 
>     * Directly linked child poms with arbitrary names via `<module>custom/childpom.xml</module>`.
> 
>     * Directly linked parent poms with arbitrary names via `<parent><relativePath>custom/parentpom.xml</relativePath>`.
> 
> 
> Does that maven support a "root pom" (not linked from any other file) not named `pom.xml`. If not, does that mean we still need a file named `pom.xml` to have a sane maven setup?

In maven, when running it - you can provide any `pom` file name as an input (i.e. run it like that `mvn -f <path>/mycustomfile.xml` 😃 Thus the root pom does not have to be named `pom.xml`. And adding the ability to have a custom-named root pom was exactly the request in the https://github.com/dependabot/dependabot-core/issues/4425. 

Unfortunately, in the current dependabot implementation i could not see any easy approach with a reasonable amount of effort to make it customizable, as this name(s) then will have to be provided as a part of dependabot yaml configuration https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#maven-repository, so that users can specify either one or an array of file names with or without paths as the entrance points to their maven project(s) within the repo, so that we know what to look for. And after that - refactor the code to not require pom.xml file to be present in all the places. 

Thus to sum up, i would, indeed, agree, that within the existing limitations of the current dependabot checker implementation for maven - repo must have a file named `pom.xml` to be treated as a valid maven project.

deivid-rodriguez:
Great, I created #6491 as the first step, what do you think?

soulus13:
> Great, I created #6491 as the first step, what do you think?

Have checked it - looks good to me! And then after rebase i will get rid of all the leftover magic names, and will, maybe, look a bit in the specks, to see if i can add any additional valuable checks for dependency checking of parent poms, as that surely would not hurt :)

soulus13:
Rebased the branch on the updated main, and will follow-up with some extra tests within a day or two, to cover how the parent validation now works after the magic has been removed 😃

soulus13:
Added extra tests for the parent pom updates (including parent pom-s with custom name), so should be ready for the final review, @deivid-rodriguez!

maven/lib/dependabot/maven/file_fetcher.rb
deivid-rodriguez: After the clarification in https://github.com/dependabot/dependabot-core/pull/6248#issuecomment-1398715068, I wonder if (even if currently unused) we should leave this method as is. No strong opinion though.

deivid-rodriguez (approved):
Looks awesome, just a minor comment, but maybe it's fine as is!

maven/lib/dependabot/maven/file_fetcher.rb
soulus13: Nope, I do agree that it should be updated to reflect the real logic - one of the files for the dependency inspection must be "pom.xml". As i have force-pushed the branch several times to look nicer after the merge - this thread has now been detached from the source, so i think the current version reflects the reality best:
`filenames.include?("pom.xml")`

deivid-rodriguez (approved):

soulus13:
Guess it is good to merge now? :)

deivid-rodriguez:
Yes, I'll merge this soon :)

jeffwidman (commented):
Quick question since I  don't know much about the `maven` ecosystem and also haven't deep-dived in this code... 

Since this PR will now treat any file that ends in `.xml` as a potential manifest file, what will happen to other `.xml` files in these paths that are not POM files? I didn't see any new test cases covering this... 

I suspect we previously didn't worry about them because `pom.xml` indicates what it's about... Does the code do any sanity checking of the file contents to see if it looks like a manifest file? And if contents turn out to not be a manifest file, will it get silently ignored, or proceed to a point in the code where it blows up with an "unparsable manifest" exception?

soulus13:
> Quick question since I don't know much about the `maven` ecosystem and also haven't deep-dived in this code...
> 
> Since this PR will now treat any file that ends in `.xml` as a potential manifest file, what will happen to other `.xml` files in these paths that are not POM files? I didn't see any new test cases covering this...
> 
> I suspect we previously didn't worry about them because `pom.xml` indicates what it's about... Does the code do any sanity checking of the file contents to see if it looks like a manifest file? And if contents turn out to not be a manifest file, will it get silently ignored, or proceed to a point in the code where it blows up with an "unparsable manifest" exception?

Hi @jeffwidman, luckily it does not get that broad with this PR 😃 Summing up the below lines - nothing changes at all with regards to the "rogue" .xml files, compared to the current implementation.

What we have now is that the entry point must be a valid `pom.xml` file, which can have inside both path to its parent (must be a 'pom.xml' named file) and any number of "child" modules' paths (which also must be named `pom.xml`). And surely that can be repeated for its children recursively. And these are the only xml files that are validated by Dependabot.

Now what changes within this PR - is just adding the support for custom parent/child names, as per the maven specification, where `pom.xml` name is the default convention (you may even omit such name and just specify the path - that would suffice), yet there are no restrictions on the naming of parent/children modules. 
This, basically, means that these `.xml` files still have to be listed within the valid "root" `pom.xml` file as either parent or children modules. So it is really unlikely that in any direction we will meet non-pom.xml file, and in case we do - this would mean that the maven project in this repo would hardly work as well.

So both with this PR and before it could crash in the same way in different places in case of the bad file structure:
If "root" `pom.xml` is not a valid XML or not a valid `pom.xml`: could crash in `Dependabot::Maven::FileFetcher#fetch_files` (when parsing the XML itself or fetching the child modules or parent), `Dependabot::Maven::FileParser#pomfile_dependencies` (mostly raise `DependencyFileNotEvaluatable` exception) and i guess could crash somewhere else as well.
If its parent or child modules are not valid XML-s or not valid `pom.xml`-s: as the logic for them is more-or-less the same with the "root" one, and they are processed close to each other - would crush in the same places and in the same way.
And i guess in case of malformed XML - the whole process would just blow up anyway, or at least i don't see any handling for that on the module level here 😃

jeffwidman (approved):
Oh that makes perfect sense, thank you for the detailed reply... covers all the bases. 🎉 

I'll go ahead with the rebase / deploy / merge dance. Probably will deploy with a few other PR's later today.

Issue #5819, opened 2022-09-30
Update container image references in GitHub Action workflows


lorengordon:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

I've used Dependabot to update the FROM argument in a Dockerfile, but now I'm starting to use images directly in GitHub Actions, which does not use a Dockerfile. I would still like Dependabot to update the image reference in the workflow file though. Is there a way to do this today? (I did search the docs and issues but couldn't find anything.)

See: https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container

Syntax in the workflow files looks like this:

```
name: CI
on:
  push:
    branches: [ main ]
jobs:
  container-test-job:
    runs-on: ubuntu-latest
    container:
      image: node:14.16
```

jeffwidman:
This _may_ be what https://github.com/dependabot/dependabot-core/pull/2307 is trying to get at, although that code seems slightly different since it's looking for the `docker` keyword, not the `container` keyword... so I may be misunderstanding.

For now, adding new ecosystems / new places to search for updates is on pause... we hope to have a better story there in the not too distant future, but we're just not ready to provide the level of support required for a new ecosystem.

lorengordon:
That's an alternate syntax, applies to the `uses` argument... That would work also, though `container` is nice because you can control more of the docker options.

* https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-a-docker-hub-action
* https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontainer

lorengordon:
Looks like I could use Renovate instead, they recently released support for this, https://github.com/renovatebot/renovate/pull/16770

jeffwidman:
Thanks for clarifying and included the docs links.

Glad Renovate can help you get unblocked... hopefully in the not-too-distant future we can support this here as well.

ssbarnea:
I have the same need, https://github.com/ansible-community/ansible-lint-action/blob/main/action.yml#L26 does not update by itself.

thepwagner:
👋 The weird workaround for this is to add:
```yaml
apiVersion: lol
kind: trollololol
```

[...]

PR #6250, opened 2022-12-02
Ignore URLs in Actions

Nishnha:
Dependabot currently supports version updates for Actions that are referenced as a GitHub repository. There is a [feature request](https://github.com/dependabot/dependabot-core/issues/5819) to support image/container URLs in GitHub Actions, which will require checking which versions of the Action are available on the registry.

In the mean time, Dependabot should stop trying to parse these URLs as a reference to a GitHub repo and ignore them.

This PR accomplishes this by anchoring the `GITHUB_REPO_REFERENCE` regex to the start of the line.

### DIFF
--- a/github_actions/lib/dependabot/github_actions/file_parser.rb
+++ b/github_actions/lib/dependabot/github_actions/file_parser.rb
@@ -17,7 +17,7 @@ module Dependabot
       require "dependabot/file_parsers/base/dependency_set"
 
       GITHUB_REPO_REFERENCE = %r{
-        (?<owner>[\w.-]+)/
+        ^(?<owner>[\w.-]+)/
         (?<repo>[\w.-]+)
         (?<path>/[^\@]+)?
         @(?<ref>.+)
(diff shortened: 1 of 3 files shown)


deivid-rodriguez (approved):
This patch works and seems correct.

My approach would have been to go with something more verbose, because I think it could make it easier for future implementors of docker and path references:

```diff
-          # TODO: Support Docker references and path references
+          # TODO: Support Docker references
+          next if string.start_with?("docker://")
+
+          # TODO: Support path references
+          next if string.start_with?("./")
+
```

But I don't have a strong opinion on which one is best, so happy to introduce this way.

Thanks for improving this 💪

Nishnha:
> My approach would have been to go with something more verbose, because I think it could make it easier for future implementors of docker and path references:

I thought that actions could be referenced from any URL, but it seems like it's only `docker://` URLs. 
I'll leave the implementation as-is though in case this changes. That way we won't have to come back and update the regex if Actions does start supporting more than just docker hub.

PR #6253, opened 2022-12-03
Suppress Dependabot::Updater::SubprocessFailed

Nishnha:
Let's stop logging these errors to Sentry. 
We explicitly leave out error details because they may contain sensitive information, so continuing to post these errors won't help us learn anything new.

### DIFF
--- a/updater/lib/dependabot/updater.rb
+++ b/updater/lib/dependabot/updater.rb
@@ -843,14 +843,8 @@ module Dependabot
           }
         when Dependabot::SharedHelpers::HelperSubprocessFailed
           # If a helper subprocess has failed the error may include sensitive
-          # info such as file contents or paths. This information is already
-          # in the job logs, so we send a breadcrumb to Sentry to retrieve those
-          # instead.
-          msg = "Dependency update process failed, please check the job logs"
-          Raven.capture_exception(
-            SubprocessFailed.new(msg),
-            raven_context
-          )
+          # info such as file contents or paths, which is already available in
+          # the job logs.
 
           { "error-type": "unknown_error" }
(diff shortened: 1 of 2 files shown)


jeffwidman (approved):

updater/lib/dependabot/updater.rb
```diff
@@ -843,14 +843,8 @@ def handle_dependabot_error(error:, dependency:)
...
-          # info such as file contents or paths. This information is already
-          # in the job logs, so we send a breadcrumb to Sentry to retrieve those
-          # instead.
-          msg = "Dependency update process failed, please check the job logs"
-          Raven.capture_exception(
-            SubprocessFailed.new(msg),
-            raven_context
-          )
+          # info such as file contents or paths, which is already available in
+          # the job logs.
```
deivid-rodriguez, line 847: I believe these Sentry errors could actually be useful for us if they logged which command failed, and were split by command.

And I believe this is actually what the current code intends to do!

https://github.com/dependabot/dependabot-core/blob/0459eba45bea8bc7d6195b857eac1091e4a8269e/common/lib/dependabot/shared_helpers.rb#L55-L69

My understanding is that the `fingerprint` key in `#raven_context` is what makes errors be aggregated together as instances of the same error or appear as different errors.

I think the issue may be that `raven_context` of the exception is not being merged with the explicit context being passed to `Raven.capture_exception`, because we are passing a different exception type (`SubprocessFailed` vs `HelperSubprocessFailed`). I think if we copy `HelperSubprocessFailed#raven_context` to `SubprocessFailed#raven_context`, we may actually get something useful off these errors.

Show we try it?

PR #6254, opened 2022-12-03
Pass subprocess failed errors to the backend

Nishnha:
Let's pass subprocess failed errors to the Api backend so we can aggregate information about the errors better.

30c0639bb27185067a47e7bbd2a62e76355d141c re-raises a subprocess failed error in the npm vulnerability auditor that was previously only being logged to the job log.


jeffwidman:
Similar to #6253, should this be closed in favor of the approach taken in https://github.com/dependabot/dependabot-core/pull/6268?

PR #6257, opened 2022-12-04
Bump `ruby` to `3.1.3` and `bundler` to `2.3.26`

jeffwidman:
Saw this warning in the logs, which reminded me it's time to bump `ruby` to `3.1.3`:
```
warning: parser/current is loading parser/ruby31, which recognizes3.1.3-compliant syntax, but you are running 3.1.2.
```

Fix https://github.com/dependabot/dependabot-core/issues/6333


Dockerfile
```diff
@@ -70,9 +70,10 @@ RUN if ! getent group "$USER_GID"; then groupadd --gid "$USER_GID" dependabot ;
 ### RUBY
 
 # When bumping Ruby minor, need to also add the previous version to `bundler/helpers/v{1,2}/monkey_patches/definition_ruby_version_patch.rb`
-ARG RUBY_VERSION=3.1.2
+ARG RUBY_VERSION=3.1.3
 ARG RUBY_INSTALL_VERSION=0.8.5
 # Generally simplest to pin RUBYGEMS_SYSTEM_VERSION to the version that default ships with RUBY_VERSION.
+# TODO check CI to see what's being installed here before beign overwritten
```
jeffwidman, line 76: also bump it here to match:
https://github.com/dependabot/dependabot-core/blob/dd2c5dba914782e3acbd2b9f052af1b973347c88/common/dependabot-common.gemspec#L21

jeffwidman: Actually, looks like `ruby` `3.1.3` still bundles `rubygems 3.3.7`, at least per the output from CI:
```
#7 381.6 >>> Successfully installed ruby 3.1.3 into /usr/local
#7 382.5 Updating rubygems-update
#7 382.5 Successfully installed rubygems-update-3.3.7
#7 382.5 Installing RubyGems 3.3.7
#7 383.0   Successfully built RubyGem
#7 383.0   Name: bundler
#7 383.0   Version: 2.3.7
#7 383.0   File: bundler-2.3.7.gem
#7 383.0 Bundler 2.3.7 installed
#7 383.0 RubyGems 3.3.7 installed
#7 383.0 Regenerating binstubs
#7 383.0 Regenerating plugins
#7 383.0 
#7 383.0 
#7 383.0 
#7 383.0 ------------------------------------------------------------------------------
#7 383.0 
#7 383.0 RubyGems installed the following executables:
#7 383.0 	/usr/local/bin/gem
#7 383.0 	/usr/local/bin/bundle
#7 383.0 	/usr/local/bin/bundler
#7 383.0 
#7 383.1 RubyGems system software updated
```

bundler/helpers/v2/monkey_patches/definition_ruby_version_patch.rb
```diff
@@ -10,7 +10,7 @@ def source_requirements
         Gem::Specification.new("Ruby\0", requested_version)
     end
 
-    %w(2.5.3 2.6.10 2.7.6 3.0.4).each do |version|
+    %w(2.5.3 2.6.10 2.7.6 3.0.4, 3.1.2).each do |version|
```
jeffwidman, line 13: I copied the existing format, but I'm unclear:
1. Why does the `bundler` `v1` version list specify the patch level? But `bundler` `v2` version list doesn't?
2. Beyond adding the most recent version, should I be updating the patch versions for the previous versions as well? For example `3.0.4` --> [`3.0.5`](https://www.ruby-lang.org/en/news/2022/11/24/ruby-3-0-5-released/)?

@deivid-rodriguez maybe you know?

Dockerfile
deivid-rodriguez: I think that's because we install that specific version in the `Dockerfile`. I think we probably want to remove that and let the version that comes with the Ruby we're using be used.

bundler/helpers/v2/monkey_patches/definition_ruby_version_patch.rb
deivid-rodriguez: Regarding 1, I think that's "historical reasons". Up until Ruby 1.9.3, so like... 12 years ago, the Ruby version used to include a patch level and that was a meaningful part of the version, so Ruby would release versions like 1.9.3-p286, or 1.9.3-p327. Now we use only `x.y.z` versions without the `-p<n>` part.

The `-p<n>` part is still provided under the hood but is no longer meaningful. My guess is that Bundler 1.17.3 still needed that part of the version to be included even if not really necessary? Not sure, but I guess not worth touching that since we should be dropping Bundler 1 support soon-ish anyways.

Regarding 2, I think that should be ok. In the Ruby world, I've never seen any gems pinning up to the patch level of Ruby, normally they will specify things like `>= 3.0.0` or `~> 3.0`, so either should work for most cases.

jeffwidman:
Note to myself: Probably also need to bump bundler to `2.3.26`... that will likely fix the CI issue... 

No need to bump RubyGems as we're going to stop explicitly installing them:
* https://github.com/dependabot/dependabot-core/pull/6267

bundler/helpers/v2/monkey_patches/definition_ruby_version_patch.rb
deivid-rodriguez: By the way, I don't think you need to add this here. At least not as part of this PR. The idea here is to add the latest patch levels of previous supported versions of Ruby to the current Ruby version, which is already considered.

deivid-rodriguez: Suggestion per the above

```suggestion
    %w(2.5.3 2.6.10 2.7.6 3.0.4).each do |version|
```

bundler/helpers/v1/monkey_patches/definition_ruby_version_patch.rb
```diff
@@ -11,7 +11,7 @@ def index
           Gem::Specification.new("ruby\0", requested_version)
       end
 
-      %w(2.5.3p105 2.6.10p210 2.7.6p219 3.0.4p208).each do |version|
+      %w(2.5.3p105 2.6.10p210 2.7.6p219 3.0.4p208, 3.1.2p20).each do |version|
```
deivid-rodriguez, line 14: You don't need commas with `%w` and in fact this would unintentionally add a version named `3.0.4p208,` to the array

```suggestion
      %w(2.5.3p105 2.6.10p210 2.7.6p219 3.0.4p208 3.1.2p20).each do |version|
```

deivid-rodriguez: But as per https://github.com/dependabot/dependabot-core/pull/6257#discussion_r1043099175, I don't think this update is necessary, so the better suggestion would be

```suggestion
      %w(2.5.3p105 2.6.10p210 2.7.6p219 3.0.4p208).each do |version|
```

jeffwidman: Thanks @deivid-rodriguez much obliged!

bundler/helpers/v2/monkey_patches/definition_ruby_version_patch.rb
jeffwidman: Again, many thanks @deivid-rodriguez. For simplicity I bumped to `2.7.7` and `3.0.5`, as it doesn't hurt if they're on latest... also omitted the `3.1.2` as you suggested.

mctofu (approved):

jeffwidman:
Testing staging and looks okay, deploy to prod and not seeing any new/unexpected errors, so I think we're good to merge.

jeffwidman:
For whoever upgrades to ruby 3.2.0, you'll also want to bump this:
https://github.com/dependabot/dependabot-core/blob/a9892827194c4ea7c93f3f24479a480b76f44760/.devcontainer/devcontainer.json#L26

PR #6268, opened 2022-12-05
Pass `raven_context` to Sentry on failed subprocess exceptions

deivid-rodriguez:
So that we get proper differentiation of exceptions according to the subprocess that failed to run.

Sentry-raven will look into `#raven_context` of the passed exception and deep merge it into the passed context. However, the passed exception here is `SubprocessFailed` which does not define any `raven_context`. It's `HelperSubprocessFailed` defining it, so we should pass over that information to `SubprocessFailed`.

This is what I meant at https://github.com/dependabot/dependabot-core/pull/6253#discussion_r1040094395.

Completely untested.

### DIFF
--- a/common/lib/dependabot/shared_helpers.rb
+++ b/common/lib/dependabot/shared_helpers.rb
@@ -64,7 +64,7 @@ module Dependabot
       end
 
       def raven_context
-        { fingerprint: [@command], extra: @error_context }
+        { fingerprint: [@command], extra: @error_context.except(:stderr_output) }
       end
     end
 
--- a/updater/lib/dependabot/updater.rb
+++ b/updater/lib/dependabot/updater.rb
@@ -33,7 +33,15 @@ require "wildcard_matcher"
 # rubocop:disable Metrics/ClassLength
 module Dependabot
   class Updater
-    class SubprocessFailed < StandardError; end
+    class SubprocessFailed < StandardError
+      attr_reader :raven_context
+
+      def initialize(message, raven_context:)
+        super(message)
+
+        @raven_context = raven_context
+      end
+    end
 
     # These are errors that halt the update run and are handled in the main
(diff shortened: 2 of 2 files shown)


jeffwidman:
I haven't looked into this at all, but I know that `raven-sentry` has been deprecated for several years: https://github.com/getsentry/sentry-ruby#migrate-from-sentry-raven

And it looks like how context is passed around changed as part of that switchup: https://docs.sentry.io/platforms/ruby/migration/#context-interfaces-changed

Would it make sense to migrate to `sentry-ruby` before investing time in ensuring this works etc necessary to push this across the finish line?

landongrindheim (approved):
I think this is a great change! I'd love for us to be thinking about how we'll keep this seemless when we do migrate to `sentry-ruby` 😄

deivid-rodriguez:
> I think this is a great change! I'd love for us to be thinking about how we'll keep this seemless when we do migrate to sentry-ruby

Thank you! :) Sentry-ruby migration guide includes an explicit note about how to keep the current behavior: https://docs.sentry.io/platforms/ruby/migration/#exceptionraven_context.

A note about the expected result of deploying this PR:

Instead of a single Sentry exception with a lof of occurrences, we should now get a lot (or maybe a few, that's what I hope we will learn) of new exceptions, one per failed command. I think with this fix we will have the opposite problem, namely, too few aggregation, because if a subprocess command includes things like the repo name, or a dependency name, those will appear as different errors and won't be aggregated. But I still think we will be in a better place than we are now.

Nishnha:
This change is really helpful!

In Sentry I can now see the exact subprocess command that failed, and it seems like we have quite a few hidden failure cases to take a look at 😄 


I'll close out https://github.com/dependabot/dependabot-core/pull/6253 in favor of this

PR #6273, opened 2022-12-06
Aggregate more subcommand failed errors

deivid-rodriguez:
Follow up to #6272 to aggregate some more errors.

### DIFF
--- a/python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
+++ b/python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
@@ -76,8 +76,12 @@ module Dependabot
                 filenames_to_compile.each do |filename|
                   # Shell out to pip-compile.
                   # This is slow, as pip-compile needs to do installs.
+                  options = pip_compile_options(filename)
+                  options_fingerprint = pip_compile_options_fingerprint(options)
+
                   run_pip_compile_command(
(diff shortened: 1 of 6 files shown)


deivid-rodriguez:
@landongrindheim I added one more commit on top of your approval, because I had that extra change ready to go and CI was stuck so I took the change to push the extra change and get it to run. Can you have a look at the additional commit when you have some time? 🙇

python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
```diff
@@ -226,6 +236,16 @@ def new_resolver_supported?
           python_version >= Python::Version.new("3.7")
         end
 
+        def pip_compile_options_fingerprint(options)
+          options.sub(
+            /--output-file=\S+/, "--output-file=<output_file>"
```
landongrindheim, line 241: Do we need to be concerned with escaped white space in file names?

landongrindheim (approved):

python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
deivid-rodriguez: Yeah, this won't catch that. I decided to let that kind of thing be reported separately since it felt like a very edge case, so unlikely to get in the middle too much. We can merge it later if we feel it's best though!

PR #6276, opened 2022-12-07
Improve SubprocessFailed Sentry errors

deivid-rodriguez:
* Add a more meaningful title so they are distinguishable in the main index.
* Copy the original backtrace.


deivid-rodriguez:
Added one more commit to aggregate errors for a few more subprocesses.

landongrindheim (approved):
❤️

PR #6277, opened 2022-12-07
Remove no longer needed custom fingerprints

deivid-rodriguez:
This PR removes most of my recent changes 😆

Sentry groups events by stacktrace, and now we have a stacktrace for subprocess failed errors, so I believe Sentry should be able to group properly by default?

Also, I believe this is more correct because the same command may end up being run in different code locations, and those would be different errors.

### DIFF
--- a/updater/lib/dependabot/updater.rb
+++ b/updater/lib/dependabot/updater.rb
@@ -854,7 +854,7 @@ module Dependabot
           # info such as file contents or paths. This information is already
           # in the job logs, so we send a breadcrumb to Sentry to retrieve those
           # instead.
-          msg = "Subprocess #{error.raven_context[:fingerprint]} failed to run. Check the job logs for error messages"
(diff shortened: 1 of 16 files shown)


updater/lib/dependabot/updater.rb
```diff
@@ -854,7 +854,7 @@ def handle_dependabot_error(error:, dependency:)
           # info such as file contents or paths. This information is already
           # in the job logs, so we send a breadcrumb to Sentry to retrieve those
           # instead.
-          msg = "Subprocess #{error.raven_context[:fingerprint]} failed to run. Check the job logs for error messages"
+          msg = "Subprocess #{error.error_context[:command]} failed to run. Check the job logs for error messages"
```
landongrindheim, line 857: Good eye!

landongrindheim (approved):

PR #6282, opened 2022-12-08
Revert "Remove no longer needed custom fingerprints"

deivid-rodriguez:
This reverts commit 05f2c7c8ecfa4c222d59431d999833503f267cbd, reversing changes made to 5a4f0857df455c0d478a74cf9dda5e8a2b15b09d.

Sentry is getting confused about which stack trace frames are in-app vs outside-app. So it's merging unrelated exceptions together.

I think we should be able to tweak Sentry config to do the right thing but for now I'll revert the PR.


jeffwidman (approved):
Bummer, as it was starting to look like a very clean fix... this is def something that'd be nice to finish figuring out how to solve before you lose context on it, otherwise we are unlikely to remember we can remove all this fingerprint plumbing.

deivid-rodriguez:
Let's try [this](https://github.com/dependabot/dependabot-core/pull/6283).

PR #6283, opened 2022-12-08
Try fix in-app vs outside-app frame classification in Sentry

deivid-rodriguez:
Trying out an alternative to #6282.

### DIFF
--- a/updater/lib/dependabot/setup.rb
+++ b/updater/lib/dependabot/setup.rb
@@ -6,6 +6,31 @@ Bundler.settings.set_command_option(:frozen, "0")
 
 require "dependabot/sentry"
 Raven.configure do |config|
+  config.project_root = File.expand_path("../../..", __dir__)
+
+  config.app_dirs_pattern = %r{(
+    dependabot-updater/bin|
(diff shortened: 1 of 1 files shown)


brrygrdn (approved):

deivid-rodriguez:
Just wrapping things up here. I ended up merging #6282 because otherwise we won't get proper grouping on subcommand failed errors (not fully sure why).

Still, I think this PR is an improvement because it considers paths in ecosystem specific folders as "in-app" which means those frames in the stacktrace are displayed in Sentry by default.

If others agree, I'll merge this.

jeffwidman:
I honestly don't have full context here, but I trust your judgement. If you think it's an improvement, then let's merge it and try it out... can always revert if it's not...

deivid-rodriguez:
This seems to have done what I wanted, sawing the full backtrace (ecosystem + updater) by default. Hope you like it better too!

Nishnha:
The stack traces are really clean and only show dependabot-core traces, and they're properly grouped by the subprocess command. Thank you!!

PR #84, opened 2017-09-26
[WIP] Support transitioning Ruby git sources to Rubygems releases

greysteil:
Idea is to generate diffs as follows when the reference a git commit is pinned to is included in a release:

```diff
- gem "pry", git: "https://github.com/pry/pry", ref: "b12a7da"
+ gem "pry", "= 0.11.0"
```

Closes https://github.com/dependabot/feedback/issues/23.

**TODO:**
- [x] Create `GitCommitChecker` class that can tell whether a commit is included in a release
- [x] Use new `GitCommitChecker` class to generate correct requirements in `UpdateChecker`
- [x] Spec everything works in `FileUpdater` (it won't but can be fixed)
- [x] Spec everything works in `MetadataFinder` (it should...)
- [x] Update `PullRequestCreator` to explain what's happening for PRs like these
- [x] Double-check that UpdateChecker will only switch to released version when appropriate

PR #6284, opened 2022-12-08
Bundler: remove git-to-registry feature due to potential for exploitation

jakecoffman:
Like #6241 we're removing this feature since it could be used by an attacker to replace a dependency with a malicious one. 

This was originally added in #84.


Nishnha (approved):
Nice +/- LOC ratio

jeffwidman (approved):

PR #6286, opened 2022-12-09
[Bitbucket] Handle Unauthorized error on current user endpoint

stefangr:
If an access token is used that does not have access to the current_user endpoint, the default_reviewers method fails.

This PR prevents this failure.

See https://github.com/dependabot/dependabot-core/pull/6166#issuecomment-1338643047


jeffwidman (approved):
Thanks much @stefangr -- appreciate you fixing this up. ❤️

PR #6287, opened 2022-12-09
Do not convert the markdown PR description to HTML for bitbucket

stefangr:
Recently something changed in formatting the PR description, resulting in a mess of HTML where a nice description should be in bitbucket.

This PR changes the PR description from:
afbeelding

to:
afbeelding

### DIFF
--- a/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
@@ -34,7 +34,7 @@ module Dependabot
           @github_redirection_service = github_redirection_service
         end
 
-        def sanitize_links_and_mentions(text:, unsafe: false)
+        def sanitize_links_and_mentions(text:, unsafe: false, format_html: true)
           doc = CommonMarker.render_doc(
             text, :LIBERAL_HTML_TAG, COMMONMARKER_EXTENSIONS
           )
@@ -45,6 +45,8 @@ module Dependabot
           sanitize_nwo_text(doc)
 
           mode = unsafe ? :UNSAFE : :DEFAULT
+          return doc.to_commonmark([mode] + COMMONMARKER_OPTIONS) unless format_html
+
           doc.to_html(([mode] + COMMONMARKER_OPTIONS), COMMONMARKER_EXTENSIONS)
         end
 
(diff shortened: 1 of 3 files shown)


jeffwidman:
Sorry for the delay here.

So you're saying it used to work fine, then something changed and it stopped working? If so, do you know what change broke it here in `dependabot-core`?

I'm wondering if that change was unintentional, or maybe we upgraded a library that flipped a default setting w/o us realizing it. Because rather than special-casing BitBucket, in that case we might be able to set the value back to what it was previously which was working everywhere...

Also, can you squash history down to drop the changes from [5bc9df6](https://github.com/dependabot/dependabot-core/pull/6287/commits/5bc9df6d5f9b1a9fe09e37d1bd3b012e35329b4f)? Since it looks like you backed those out in favor of flipping a config.

stefangr:
I traced the change back to the upgrade of the dependabot-omnibus package from 0.212.0 to 0.214.0.

There I found this [little change](https://github.com/dependabot/dependabot-core/commit/9fd1c8d2a0617566ea6395930f263fdf526b2095). Which is fine of course and should stay.

jeffwidman:
Oh yeah that makes sense. 😀

I'm on my phone now but should be able to review the code in detail this coming week.

common/lib/dependabot/pull_request_creator/message_builder/metadata_presenter.rb
```diff
@@ -244,10 +244,14 @@ def source_provider_supports_html?
           !%w(azure bitbucket codecommit).include?(source.provider)
         end
 
+        def sanitize_to_html?
```
stefangr, line 247: I added a new method here for the sanitation, but perhaps using the `source_provider_supports_html` method is better.

jeffwidman (approved):
Oh man, you're making me glad I wasn't able to review this last week! 🤣 

This PR just got cleaner/tidier as time went on. Nice job!

common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
```diff
@@ -45,6 +45,8 @@ def sanitize_links_and_mentions(text:, unsafe: false)
           sanitize_nwo_text(doc)
 
           mode = unsafe ? :UNSAFE : :DEFAULT
+          return doc.to_commonmark([mode] + COMMONMARKER_OPTIONS) unless format_html
```
jeffwidman, line 48: Actually, if we do this, it'll affect all providers that don't support html:
https://github.com/dependabot/dependabot-core/blob/9df8aec3878641a139b9b63d94e43dbbb125c715/common/lib/dependabot/pull_request_creator/message_builder/metadata_presenter.rb#L244

That doesn't mean this is a bad way to go, but let me double-check if that list is still accurate... for example I already found #6453

jeffwidman: Let's go ahead and merge this, and then I'll open a follow-up PR to address #6453 and get input from Microsoft folks... I have a feeling that it now supports HTML, which would make our lives easy here.

Issue #6289, opened 2022-12-09
Stop hiding exception raised by composer


jeffwidman:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code improvement description

While debugging a customer issue, we ran into the problem that the error output from the `composer` helper was being swallowed:
https://github.com/dependabot/dependabot-core/blob/b911f9dfe1e11d19edcafb11e83d305758c97f85/composer/lib/dependabot/composer/update_checker/version_resolver.rb#L295-L304

The underlying error:
```
Your requirements could not be resolved to an installable set of packages. (Dependabot::SharedHelpers::HelperSubprocessFailed)
  Problem 1
    - php-amqplib/php-amqplib v2.11.3 conflicts with php[7.4.0].
    - php-amqplib/php-amqplib v2.11.3 conflicts with php[7.4.0].
    - Installation request for php 7.4.0 -> satisfiable by php[7.4.0].
    - Installation request for php-amqplib/php-amqplib (locked at v2.11.3, required as ^2.6) -> satisfiable by php-amqplib/php-amqplib[v2.11.3].
```

Swallowing the error makes it hard for users to figure what's going on.

The original motivation for swallowing was in the early days of Dependabot Preview, it opened issues for exceptions raised in the updater/core which has forced us to suppress certain errors that might be spammy or intermittent, e.g. resolvability errors. In version updates we no longer create issues so attaching errors to the last update job means we can attach any number of errors to highlight potential issues without spamming users.

A potential improvement would be to expose structured errors per checked dependency to make it clearer which dependency caused the error.

jeffwidman:
Hmm... digging into this more:
This was originally added by https://github.com/dependabot/dependabot-core/pull/165/ 
https://github.com/dependabot/dependabot-core/blob/a65ad2269efdc597ad01142ccc7a7e4664039f92/lib/dependabot/update_checkers/php/composer.rb#L132-L136

So the exception is the right one to raise... but the problem here is that PHP has native extensions that we don't install in the place where we're running `dependabot-core`... so a resolvability error may be because of install, or may be because we can't actually update... so we can't just blindly raise the exception w/o breaking everyone using native PHP extensions.

jeffwidman:
[...]

PR #6290, opened 2022-12-09
Expose composer error message when not resolvable

jeffwidman:
Expose the underlying composer error message when a dependency is not resolvable.

While debugging a customer issue, we ran into the problem that the error output from the `composer` helper was being swallowed: https://github.com/dependabot/dependabot-core/blob/b911f9dfe1e11d19edcafb11e83d305758c97f85/composer/lib/dependabot/composer/update_checker/version_resolver.rb#L295-L304

The underlying error:
```
Your requirements could not be resolved to an installable set of packages. (Dependabot::SharedHelpers::HelperSubprocessFailed)
  Problem 1
    - php-amqplib/php-amqplib v2.11.3 conflicts with php[7.4.0].
    - php-amqplib/php-amqplib v2.11.3 conflicts with php[7.4.0].
    - Installation request for php 7.4.0 -> satisfiable by php[7.4.0].
    - Installation request for php-amqplib/php-amqplib (locked at v2.11.3, required as ^2.6) -> satisfiable by php-amqplib/php-amqplib[v2.11.3].
```

Swallowing the error makes it hard for users to figure what's going on.

The original motivation for swallowing was in the early days of Dependabot Preview, it opened issues for exceptions raised in the updater/core which has forced us to suppress certain errors that might be spammy or intermittent, e.g. resolvability errors. In version updates we no longer create issues so attaching errors to the last update job means we can attach any number of errors to highlight potential issues without spamming users.

Fix https://github.com/dependabot/dependabot-core/issues/6289

### DIFF
--- a/composer/lib/dependabot/composer/update_checker/version_resolver.rb
+++ b/composer/lib/dependabot/composer/update_checker/version_resolver.rb
@@ -301,7 +301,7 @@ module Dependabot
             # If there *is* a lockfile we can't confidently distinguish between
             # cases where we can't install and cases where we can't update. For
             # now, we therefore just ignore the dependency.
-            nil
+            raise Dependabot::DependencyFileNotResolvable, sanitized_message
           elsif error.message.include?("URL required authentication") ||
                 error.message.include?("403 Forbidden")
             source = error.message.match(%r{https?://(?<source>[^/]+)/}).named_captures.fetch("source")


jeffwidman:
Putting this on ice as it'll need additional work to handle when missing PHP native extensions as mentioned in https://github.com/dependabot/dependabot-core/issues/6289#issuecomment-1344755822

jeffwidman:
We can probably switch this to logging the error message, so it's still visible in the job logs, but doesn't actually fail the job due to the problem mentioned in https://github.com/dependabot/dependabot-core/issues/6289#issuecomment-1344755822

GarryHurleyJr (approved):

jeffwidman:
Uh, thanks for the review/approve, but we don't want to merge this as-is... see the comment thread above.

robaiken:
I have opened a new PR to resolve this, so I am closing this PR: https://github.com/dependabot/dependabot-core/pull/9657

jeffwidman:
That's awesome you're tackling it! 

However, let's not actually close PR's until another PR is actually merged... because my experience is not all PR's get merged, and then we end up with a world of forgetting to re-open an original PR that may be halfway done. 

☝️ is a general best practice that I see followed in most open source repos, not specific just to this PR.

jeffwidman:
This was resolved by:
* https://github.com/dependabot/dependabot-core/pull/9657

PR #6291, opened 2022-12-09
Strip out URL credentials when reporting composer errors

jeffwidman:
These error messages should have their URL creds stripped before reporting them.

Also, from a GitHub security perspective, these creds never even make it to `dependabot-core` since we use a separate proxy for handling credentials... so this only affects users running `dependabot-core` on their own.

### DIFF
--- a/composer/lib/dependabot/composer/update_checker/version_resolver.rb
+++ b/composer/lib/dependabot/composer/update_checker/version_resolver.rb
@@ -285,12 +285,12 @@ module Dependabot
             raise MissingExtensions, [missing_extension]
           elsif error.message.include?("cannot require itself") ||
                 error.message.include?('packages.json" file could not be down')
-            raise Dependabot::DependencyFileNotResolvable, error.message
+            raise Dependabot::DependencyFileNotResolvable, sanitized_message
           elsif error.message.include?("No driver found to handle VCS") &&
                 !error.message.include?("@") && !error.message.include?("://")
             msg = "Dependabot detected a VCS requirement with a local path, " \
                   "rather than a URL. Dependabot does not support this " \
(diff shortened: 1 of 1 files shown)


composer/lib/dependabot/composer/update_checker/version_resolver.rb
```diff
@@ -285,12 +285,12 @@ def handle_composer_errors(error)
             raise MissingExtensions, [missing_extension]
           elsif error.message.include?("cannot require itself") ||
                 error.message.include?('packages.json" file could not be down')
-            raise Dependabot::DependencyFileNotResolvable, error.message
+            raise Dependabot::DependencyFileNotResolvable, sanitized_message
```
jurre, line 288: Don't we sanitize this downstream? https://github.com/dependabot/dependabot-core/blob/main/updater/lib/dependabot/sentry.rb#L6 if that's not working as expected, maybe we should fix it there?

Nishnha (approved):
Thanks

Unfortunately there are no tests to show what gets stripped out by `sanitized_message` but from the looks of it, it just replaces credentials with ***
https://github.com/dependabot/dependabot-core/blob/bb56ba30a00d14595696b2505e5603530057c366/composer/lib/dependabot/composer/update_checker/version_resolver.rb#L243
https://github.com/dependabot/dependabot-core/blob/bb56ba30a00d14595696b2505e5603530057c366/composer/lib/dependabot/composer/update_checker/version_resolver.rb#L525-L527

jeffwidman:
It turned out this contained a bug, which wasn't exposed until https://github.com/dependabot/dependabot-core/pull/6434 but furthermore wasn't even needed:
* https://github.com/dependabot/dependabot-core/pull/6435

Issue #3010, opened 2021-01-19
The PHP composer features should use the v2 metadata from Packagist


stof:
**Package manager/ecosystem**
composer
**Manifest contents prior to update**
n/a
**Updated dependency**
n/a
**What you expected to see, versus what you actually saw**

The code of the composer support is currently reading metadata from Packagist using the v1 metadata. This happens for instance in https://github.com/dependabot/dependabot-core/blob/0b286b2ac5a7a25059d158cc85f2613c3a30d937/composer/lib/dependabot/composer/update_checker/latest_version_finder.rb#L109 (there might be other places in the code reading such metadata too).
Packagist supports a new metadata format (introduced for composer 2 as composer 1 does not know how to use it) for better performance and reduced bandwidth. The composer team plans to deprecate the v1 metadata format and disable it on packagist.org in the future (the schedule is not defined yet as that depends on the adoption of composer 2).

It would be great if dependabot could switch to using the endpoint of metadata v2, so that it is ready for this upcoming deprecation. This endpoint is at `https://repo.packagist.org/p2/` instead of `https://repo.packagist.org/p/`

> Note that you currently use the `packagist.org` domain rather than the `repo.packagist.org` domain, which forbids you from using the official metadata mirrors. If dependabot runs in a US-based datacenter, using the `repo.packagist.org` domain would allow you to hit the mirror in Montreal instead of hitting a server in Europe, improving latency. Doing that switch is unrelated to the v1 vs v2 metadata change though (both are mirrored by packagist).

The differences between the v1 and v2 metadata formats are documented in https://github.com/composer/composer/blob/master/UPGRADE-2.0.md#for-composer-repository-implementors. Here are the main highlights:
- metadata are minified (this can be detected based on the `"minified": "composer/2.0"` top-level key in the file) to reduce the file size thanks to the fact that most releases don't change most of the metadata fields compared to the previous release. This will probably impact dependabot as it will need to un-minify the metadata
- only metadata for the package itself are in the file, not other packages providing the same name (should not be an issue for dependabot if it cares about the versions for the package name it loads)
- dev versions (corresponding to git branches rather than tags) are in a separate file (most projects don't allow dev versions of their dependencies and so don't need the metadata for git branches. Splitting them allowed a better caching as releases happen less often than pushes to branches)
- absence of a package is now always represented by a 404 response rather than sometimes returning a 200 response with an empty list in the `packages` key.

jeffwidman:
[...]

PR #6297, opened 2022-12-09
Switch to `repo.packagist.org` to allow mirrors

jeffwidman:
Per https://github.com/dependabot/dependabot-core/issues/3010:
> Note that you currently use the packagist.org domain rather than the
> repo.packagist.org domain, which forbids you from using the official
> metadata mirrors. If dependabot runs in a US-based datacenter, using
> the repo.packagist.org domain would allow you to hit the mirror in
> Montreal instead of hitting a server in Europe, improving latency.
> Doing that switch is unrelated to the v1 vs v2 metadata change though
> (both are mirrored by packagist).

So this switches the two places we call out to `packagist.org` to use the new domain.

### DIFF
--- a/composer/lib/dependabot/composer/metadata_finder.rb
+++ b/composer/lib/dependabot/composer/metadata_finder.rb
@@ -48,7 +48,7 @@ module Dependabot
       def packagist_listing
         return @packagist_listing unless @packagist_listing.nil?
 
-        response = Dependabot::RegistryClient.get(url: "https://packagist.org/p/#{dependency.name.downcase}.json")
+        response = Dependabot::RegistryClient.get(url: "https://repo.packagist.org/p/#{dependency.name.downcase}.json")
 
         return nil unless response.status == 200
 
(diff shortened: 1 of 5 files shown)


jeffwidman:
@stof @Seldaek can I trouble you for a quick review?

While here at GitHub this code primarily runs in a US datacenter, keep in mind that `dependabot-core` is an open source library, so there are also folks running it in their CI mashups from datacenters all over the world. So keep that in mind when deciding which domain you want us to hit.

Nishnha (approved):

Seldaek (commented):
This is good. Repo is the official domain which is using geodns, packagist.org works the same tho so technically this won't change a whole lot but it's more correct still.

jeffwidman:
Thanks for the quick reviews!

> Repo is the official domain which is using geodns, packagist.org works the same tho so technically this won't change a whole lot but it's more correct still.

FWIW, both users and SEO might be less confused if you `301` duplicate domain URLs to the cannonical domain. For example, since both `packagist.org` and `repo.packagist.org` show the same content, if I hadn't seen #3010, I would have zero idea which was the preferred domain.

Seldaek:
Yeah we do redirect for the website. We don't redirect old repo URLs on the main domain because that breaks dumb clients and also causes extra requests.

PR #6303, opened 2022-12-12
Upgrade to Bundler 2.4.0.dev

deivid-rodriguez:
Just getting a sense of how hard will be to upgrade Bundler to the upcoming 2.4.0 release.


deivid-rodriguez:
Quick update here. This is finally green. Next Bundler version will include a couple of extra improvements that should allow us to remove one monkeypatch and simplify (remove?) circular dependency errors detection, but the bulk of the work is done!

brcarp:
Curious, what's the backstory on `Specs / ci (pub, pub, 2, 0, pub) (pull_request)` running for 360 minutes and then getting cancelled?

(I'm following this PR because my Dependabot runs hit a timeout error, and I've been told that the Bundler upgrade might fix that.)

deivid-rodriguez:
I think that was just a transient issue, let me rebase this and retrigger CI.

brcarp:
> I think that was just a transient issue, let me rebase this and retrigger CI.

Excellent, thanks. Looks like you need some new VCR cassettes, by the way.

deivid-rodriguez:
I actually think this is an existing issue in main: #6834.

jakecoffman (approved):
🥳

PR #6304, opened 2022-12-12
fix hostname matching

jakecoffman:
The hostname matching `bare_uri.start_with?(c["host"])` could lead to vulnerabilities where an attacker could register a domain like `npm.registry.com.evil.com` and Dependabot would send credentials to it. 

This is not possible in production Dependabot as the updater image never has passwords or tokens. However we know some folks are running this on their own custom setup so we want to make sure they are secure as well.

There's a sufficient set of tests around this one so no new tests required.


jurre (approved):
That makes sense!

PR #6306, opened 2022-12-12
fix hostname matching to be exact to avoid security issues

jakecoffman:
More fixes like #6304. 

We want to do exact matches on hostname rather than substring match to avoid security issues.


honeyankit (approved):

jeffwidman (approved):
Did you want to add a test that would have matched under the old fuzzy match but not an exact match?

PR #6310, opened 2022-12-14
Group errors running `go list`

deivid-rodriguez:


jurre (approved):

jeffwidman (approved):
I may be wrong, but I suspect this won't throw that many exceptions, and the ones coming from [`go get`](https://github.com/dependabot/dependabot-core/blob/d8136f3c1008124baeccbfefa54f683f9a00c23f/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb#L175) are more likely to complain.

But I haven't looked at Sentry to see what's actually coming in...

deivid-rodriguez:
It does raise a few times. I think probably not too many but enough to be annoying since they are currently all raised as new errors.

PR #6315, opened 2022-12-15
Migrate to `packagist`'s `v2` metadata API

jeffwidman:
As explained in https://github.com/dependabot/dependabot-core/issues/3010:
> Packagist supports a new metadata format (introduced for composer 2 as composer 1 does not know how to use it) for better performance and reduced bandwidth. The composer team plans to deprecate the v1 metadat>
> ...
>The new endpoint is at https://repo.packagist.org/p2/ instead of https://repo.packagist.org/p/
>
> The differences between the v1 and v2 metadata formats are documented in https://github.com/composer/composer/blob/master/UPGRADE-2.0.md#for-composer-repository-implementors. Here are the main highlights:
>
> * metadata are minified (this can be detected based on the `"minified": "composer/2.0"` top-level key in the file) to reduce the file size thanks to the fact that most releases don't change most of the metada>
> * only metadata for the package itself are in the file, not other packages providing the same name (should not be an issue for dependabot if it cares about the versions for the package name it loads)
> * dev versions (corresponding to git branches rather than tags) are in a separate file (most projects don't allow dev versions of their dependencies and so don't need the metadata for git branches. Splitting >
> * absence of a package is now always represented by a 404 response rather than sometimes returning a 200 response with an empty list in the `packages` key.

So I did the following steps:
1. Update the two places we call this endpoing to hit `/p2/` rather than
   `/p/`
2. Update the code that parses those URL responses to handle the change
   from returning a hash of version listings to an array of version
   listings.
3. Research the minification impact--it turned out to not affect us
   because we were already searching for the newest source URL link,
   which is the one that doesn't get minified away.
4. Update the content of the packagist response test fixtures to match
   what's returned by the new endpoint. This was by far the most time
   consuming part as we had some test fixtures that tried to do
   double-duty in various ways, so I cleaned those up on a few separate
   PR's, which made the updates in this PR _much_ more straightforward.
5. Decide how to handle the move of dev versions (aka branches rather 
    than tags) to an outside channel. Since these are tied to a branch rather
    than a tag, it doesn't really make sense to support these, as currently
    I'm not aware of a way to know "this branch updates to this other branch".
    Additionally, pre-release version suffixes such as `BETA1` or
   `-RC1` are still returned on the main channel, so anyone who really wants
   pre-release versions that are upgradeable can still use those.
6. Investigate if any tests needed updating that were looking for `200`
   response and a missing package now that Packagist guarantees they'll
   return a `404` for missing packages. The only tests affected were a
   couple that check if Packagist returns an empty array, and I'd rather
   keep those for defensive coding purposes, as they may occur in the
   future if Packagist ever has bugs.

Fix https://github.com/dependabot/dependabot-core/issues/3010


composer/spec/fixtures/packagist_responses/path_dep--path_dep.json
```diff
@@ -1 +1,198 @@
...
+            };
+        </script>
+        <script nonce="ZMg6qEFJlrtA59ZjwkowDg==">
+    var trackPageload = false;
+</script>
+
+        <script nonce="ZMg6qEFJlrtA59ZjwkowDg==" src="/build/app.js?v=1664051059"></script>
+
+            </body>
+</html>
```
jeffwidman, line 198: I verified this is what `curl` sees, ie, the `html` response isn't caused by including a browser header.

I was a little surprised that the response went from the `json` `404` message to a `HTML` `404` message for a `.json` endpoint, but I suppose that makes sense if they simply changed `packagist.org` to not return anything custom so it defaults to the generic 404...

composer/spec/fixtures/packagist_response.json
```diff
@@ -1,2726 +1,2232 @@
...
-                "type": "library",
-                "uid": 1099516,
-                "version": "1.22.0",
-                "version_normalized": "1.22.0.0"
+                "support": {
+                    "issues": "https://github.com/Seldaek/monolog/issues",
+                    "source": "https://github.com/Seldaek/monolog/tree/1.22.0"
+                }
             },
-            "v1.22.1": {
```
jeffwidman, line 1703: note to self: this was a customization we added to ensure we parse `"v*"` versions correctly, so need to put it back... 

check for other customizations that need to be added here:
* https://github.com/dependabot/dependabot-core/commits/main/composer/spec/fixtures/packagist_response.json
* https://github.com/dependabot/dependabot-core/commits/a2c111565b61e371289ca0b90d7a3682ff2590c1/spec/fixtures/php/packagist_response.json?browsing_rename_history=true&new_path=composer/spec/fixtures/packagist_response.json&original_branch=main

composer/spec/fixtures/packagist_responses/path_dep--path_dep.json
jeffwidman: Interestingly, I noticed:
* https://repo.packagist.org/p2/pathdep/pathdep.json -> returns `"404 not found, no packages here"`
* https://repo.packagist.org/p2/path_dep/path_dep.json | https://repo.packagist.org/p2/pathdep/path_dep.json | https://repo.packagist.org/p2/path_dep/pathdep.json all return the HTML'ified 404 page.

Not sure why the addition of the `_` results in a HTML 404 page vs a simple plain text 404 error string.

composer/spec/fixtures/packagist_response.json
jeffwidman: These were handled in:
* https://github.com/dependabot/dependabot-core/pull/6332
* https://github.com/dependabot/dependabot-core/pull/6335

composer/spec/dependabot/composer/update_checker/latest_version_finder_spec.rb
```diff
@@ -165,7 +165,7 @@
             package_manager: "composer"
           )
         end
-        it { is_expected.to eq(Gem::Version.new("2.3.x.pre.dev")) }
+        it { is_expected.to eq(Gem::Version.new("2.3.0-RC4")) }
```
jeffwidman, line 168: This is the impact of dropping support for dev releases, but as I mentioned in the PR/commit description:
> Because pre-release version suffixes such as `BETA1` or
   `-RC1` are still returned on the main channel, I don't think we
   should invest any effort into supporting dev releases. They're not
   used much, they'd be another thing to support, and the pre-release
   suffixes provide a good-enough replacement.

composer/lib/dependabot/composer/metadata_finder.rb
```diff
@@ -28,27 +28,26 @@ def look_up_source_from_packagist
...
+        # NOTE: Each listing may not have all fields because they are minified to remove duplicate elements:
+        # * https://github.com/composer/composer/blob/main/UPGRADE-2.0.md#for-composer-repository-implementors
+        # * https://github.com/composer/metadata-minifier
+        version_listings.each do |i|
+          [i["homepage"], i.dig("source", "url")].each do |url|
+            source_url = Source.from_url(url)
+            return source_url unless source_url.nil?
+          end
+        end
+        return nil # rubocop:disable Style/RedundantReturn
```
jeffwidman, line 44: Like `rubocop`, I too was surprised I needed to explicitly return `nil` here... but if I didn't, then tests started failing and when I dug in it was because it was returning one of the previous function calls...  `version_listings` IIRC.

jeffwidman:
@stof @Seldaek can you take a quick review since this updates how we hit your API?

Once you give the okay, I can probably deploy it a day or two later.

composer/lib/dependabot/composer/metadata_finder.rb
```diff
@@ -28,27 +28,26 @@ def look_up_source_from_packagist
         return nil if packagist_listing&.fetch("packages", nil) == []
         return nil unless packagist_listing&.dig("packages", dependency.name.downcase)
 
-        version_listings =
-          packagist_listing["packages"][dependency.name.downcase].
-          select { |version, _| Composer::Version.correct?(version) }.
-          sort_by { |version, _| Composer::Version.new(version) }.
-          map { |_, listing| listing }.
-          reverse
```
jeffwidman, line 36: The v1 api returned them in the opposite order: oldest-> newest.

composer/lib/dependabot/composer/metadata_finder.rb
```diff
@@ -28,27 +28,26 @@ def look_up_source_from_packagist
...
-
-        potential_source_urls =
-          version_listings.
-          flat_map { |info| [info["homepage"], info.dig("source", "url")] }.
-          compact
-
-        source_url = potential_source_urls.find { |url| Source.from_url(url) }
```
jeffwidman, line 43: When I initially started working on this, I was afraid I'd have to navigate to the exact version of the dependency that we wanted, then extract the source URL from there. That'd probably be a more accurate way to handle it. But it'd be painful in minified code since we'd need to un-minify it to determine the correct source link that matched to that particular version.

But the previous code kept it simple, and merely iterated on the list of URLs from newest to oldest until it found the first viable candidate and then used that, regardless of whether it matched the version. So by continuing to keep things simple with the same behavior, we avoid any need to un-minify.

composer/lib/dependabot/composer/metadata_finder.rb
```diff
@@ -28,27 +28,26 @@ def look_up_source_from_packagist
...
-
-        Source.from_url(source_url)
```
jeffwidman, line 45: From a perf standpoint, not sure why this old code calls an expensive function (it uses regexes) to check if viable, and then if viable re-runs the same expensive function again to return the result... I chose instead to cache the result so that if not-nil we can just return the result rather than re-running it.

composer/lib/dependabot/composer/metadata_finder.rb
```diff
@@ -28,27 +28,26 @@ def look_up_source_from_packagist
...
-          packagist_listing["packages"][dependency.name.downcase].
-          select { |version, _| Composer::Version.correct?(version) }.
-          sort_by { |version, _| Composer::Version.new(version) }.
-          map { |_, listing| listing }.
-          reverse
-
-        potential_source_urls =
-          version_listings.
-          flat_map { |info| [info["homepage"], info.dig("source", "url")] }.
-          compact
```
jeffwidman, line 41: I also didn't understand why the old code built two lists, only to throw away most of the list once it found the first match... I chose instead to evaluate the URLs as I encountered them, and then `return` as soon as a match was found.

composer/spec/dependabot/composer/update_checker_spec.rb
```diff
@@ -437,11 +437,15 @@
       # This unit test is testing that a dependency located on https://pear.horde.org is still correctly
       # handled by composer. So ignore the fact that this package actually exists on packagist, and
       # pretend it just 404's.
-      let(:packagist_response) { '{"error":{"code":404,"message":"Not Found"}}' }
+      let(:packagist_response) { "404 not found, no packages here" }
```
jeffwidman, line 440: The `404` error message returned by packagist changed slightly from `v1` to `v2`... this is the correct `v2` error message.

composer/lib/dependabot/composer/metadata_finder.rb
Seldaek: Un-minify here would really only mean while you iterate through versions you keep the last good source_url you found, once you find the correct version field you return the last good url and done. But it also seems reasonable as is because in 99% of cases URLs will be the same for all versions

composer/lib/dependabot/composer/metadata_finder.rb
```diff
@@ -28,27 +28,26 @@ def look_up_source_from_packagist
         return nil if packagist_listing&.fetch("packages", nil) == []
         return nil unless packagist_listing&.dig("packages", dependency.name.downcase)
```
Seldaek, line 29: My ruby knowledge is pretty basic so apologies if I'm missing something but are you not calling packagist_listing several times here, which would result in several requests happening to us? Even if `Dependabot::RegistryClient.get` below is caching the network bits, it still seems kinda wasteful vs just calling it once here and working with the result instead of doing 3 calls.

composer/lib/dependabot/composer/metadata_finder.rb
```diff
@@ -28,27 +28,26 @@ def look_up_source_from_packagist
...
+          end
+        end
+        return nil # rubocop:disable Style/RedundantReturn
       end
 
       def packagist_listing
         return @packagist_listing unless @packagist_listing.nil?
 
-        response = Dependabot::RegistryClient.get(url: "https://repo.packagist.org/p/#{dependency.name.downcase}.json")
+        response = Dependabot::RegistryClient.get(url: "https://repo.packagist.org/p2/#{dependency.name.downcase}.json")
```
Seldaek, line 50: One little catch here, the v2 api does separate tagged releases and branch releases. So for packages which only have dev versions like https://packagist.org/packages/roave/security-advisories for example you will find no version in this file https://repo.packagist.org/p2/roave/security-advisories.json and instead you should look up in `p2/#{dependency.name.downcase}~dev.json` i.e. https://repo.packagist.org/p2/roave/security-advisories~dev.json

You could also first look up the `~dev` file if this is only used to find the source url. It'll typically be a much smaller file as there are often less dev versions than releases, and on packagist.org at least every package should have at least one dev version, but not every package has a release.

composer/lib/dependabot/composer/update_checker/latest_version_finder.rb
```diff
@@ -143,7 +143,23 @@ def parse_registry_response(response, url)
...
+          # * "packages": {<package name>: {<version number>: {hash of metadata for a particular release version}}}
+          #
+          # The updated v2 Metadata API is of the form:
+          # * "packages": {<package name>: [{hash of metadata for a particular release version}]}
+          #
+          # More details:
+          # * https://github.com/dependabot/dependabot-core/issues/3010
+          # * https://github.com/composer/composer/blob/master/UPGRADE-2.0.md#for-composer-repository-implementors
+          if version_listings.is_a?(Array) # default to v2 format
+            version_listings.map { |i| i.fetch("version") }
```
Seldaek, line 157: I think you could simply do this for all registries as even those with a hash keyed by versions must contain the version in the metadata hash, so this should be compatible with everything (tho a tiny bit slower than simply returning `.keys` I am sure)

composer/spec/fixtures/packagist_responses/path_dep--path_dep.json
Seldaek: This discrepancy is because `_` is not allowed in package names, so we route `/p2?/[A-Za-z0-9-]+/[A-Za-z0-9-]+(~dev)?\.json$` directly to filesystem to serve the static metadata fast (with a static json response for 404s) and everything else hits the application and thus returns a html 404.

Seldaek:
Alright I have reviewed everything and commented where I had something to contribute. Everything else looks reasonable to me 👍🏻

composer/lib/dependabot/composer/metadata_finder.rb
jurre: `packagist_listing` is memoized in the method itself already, so this is ~the same performance wise 👍

composer/lib/dependabot/composer/update_checker/latest_version_finder.rb
stof: Note that composer itself completely ignores the keys in the version listing. So it would be safer to always read it from the metadata hash as repositories other than packagist.org could be returning different values.

composer/lib/dependabot/composer/metadata_finder.rb
deivid-rodriguez: Oh, I think this is because in the last line of a method, `return <something>` is redundant. You can just write `<something>`. Because Ruby by default uses the result of the last line executed as the return value.

composer/lib/dependabot/composer/update_checker/latest_version_finder.rb
jeffwidman: 👍 switched to coercing v1 results to an array of metadata hashes to match v2 format.

composer/lib/dependabot/composer/metadata_finder.rb
jeffwidman: oh, good call, I didn't even think about writing a plain `nil` here as I was only considering omitting it entirely... and indeed `nil` passes both CI and rubocop. 👍

composer/lib/dependabot/composer/metadata_finder.rb
jeffwidman: Hmm... this is a bit of a conundrum... 

1. The `latest_version_finder.rb` code is used when we query packagist for the latest version. So if someone has a `composer.json` that includes `"roave/security-advisories": "dev-latest"`, then when we query for newest version, because we don't query the branch release  `*~dev.json` file, Dependabot will get an empty version list and simply short-circuit. Initially I was thinking this meant we needed to add support for `*~dev.json` files, but on second thought since branch releases by definition don't have "newer versions" that Dependabot can update, probably we don't actually need this?
2. The `metadata_finder.rb` code only gets called to populate the metadata of a PR _after_ we know we have an update. So once again, since dev/branch releases by definition cannot get PR updates, we'll never hit the metadata finder code... so from a correctness standpoint this should be okay?
3. Within `metadata_finder.rb`, there's a small optimization possible to use the branch release file for finding source URL even for normal tagged releases since it's a smaller file to download, but IIUC that's an optimization, not a correctness issue?

@Seldaek Can you confirm my understanding ☝️ is correct?

jeffwidman: Thanks! 

I'll probably ship this as-is then and do step 3 as a further optimization in a follow-on PR once I confirm this is working as expected.

composer/lib/dependabot/composer/metadata_finder.rb
```diff
@@ -28,27 +28,26 @@ def look_up_source_from_packagist
...
-
-        Source.from_url(source_url)
+        version_listings = packagist_listing["packages"][dependency.name.downcase]
+        # Packagist returns an array of version listings sorted newest to oldest.
+        # So iterate until we find the first URL that appears to be a source URL.
+        #
+        # NOTE: Each listing may not have all fields because they are minified to remove duplicate elements:
+        # * https://github.com/composer/composer/blob/main/UPGRADE-2.0.md#for-composer-repository-implementors
+        # * https://github.com/composer/metadata-minifier
+        version_listings.each do |i|
```
honeyankit, line 38: Instead of using two nested loops, we could use a single loop to iterate through all the URLs in the version_listings array. This will eliminate the need to loop through the homepage and source.url keys separately, and it will also allow you to return the first non-nil source_url value. 

```ruby
version_listings.each do |i|
  url = i["homepage"] || i.dig("source", "url")
  source_url = Source.from_url(url)
  return source_url unless source_url.nil?
end


```

honeyankit (approved):

composer/lib/dependabot/composer/metadata_finder.rb
jeffwidman: The code is effectively identical from a perf standpoint... Because both are iterating through the exact same list of things, and then breaking on the first non-nil match. 

The only difference I can think of is if Ruby does something funky under our feet at the compiler level of setting up a loop instead of an `||` statement... but I doubt it.

I'll update to this though, as it strikes me as a little more readable... 👍

jeffwidman: Regardless form a perf perspective, the nested for-loop isn't a bad thing because it will _immediately_ `return` on the first match, so even if we flatten the loop (which is what ☝️ is suggesting) then it's still the same runtime complexity... because we still have to evaluate the same number of things.

honeyankit: The below should fix the bug. But your logic seems to be fine as well. Since it is tested. I will go with it. 

```ruby
source_url = version_listings.find do |i|
  url = i["homepage"] || i.dig("source", "url")
  source = Source.from_url(url)
  !source.nil?
end
return source_url unless source_url.nil?


```

jeffwidman:
Deployed this about 30 mins ago and so far everything looks hunky-dory, so merging.

composer/lib/dependabot/composer/metadata_finder.rb
jeffwidman: I explored this a bit in a follow-on PR, but I think for now the optimization isn't worth shipping--see my comments on the PR for details:
* https://github.com/dependabot/dependabot-core/pull/6360

Issue #6131, opened 2022-11-14
Validate SHA256Sum for Dart


jeffwidman:
As of Dart 2.18 (not sure about earlier) the Dart project publishes SHA256 sums, so we should start validating those as part of our Dart installs: 
* https://dart.dev/get-dart/archive
* https://github.com/dependabot/dependabot-core/pull/6129
* https://github.com/dependabot/dependabot-core/pull/6129/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R263

This would make a great first issue if anyone is interested in starting to contribute to Dependabot.

PR #6319, opened 2022-12-16
Validate SHA256Sum for Dart

blue2cat:
Addresses #6131
This PR introduces changes to the Dart installation process. In addition to downloading the `zip` with the Dart SDK, the installation script now also downloads a `SHA256` hash file from the Dart project and verifies the downloaded zip. 

 - [x] Validate `SHA256` sums published by the Dart project. 
 - [x] Bump Dart from `1.8.5` -> `1.8.6`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -271,15 +271,20 @@ ENV PUB_CACHE=/opt/dart/pub-cache \
 
 # https://dart.dev/get-dart/archive
 ARG DART_VERSION=2.18.5
-# TODO Dart now publishes SHA256 checksums for their releases, we should validate against those.
+
 RUN DART_ARCH=${TARGETARCH} \
   && if [ "$TARGETARCH" = "amd64" ]; then DART_ARCH=x64; fi \
-  && curl --connect-timeout 15 --retry 5 "https://storage.googleapis.com/dart-archive/channels/stable/release/${DART_VERSION}/sdk/dartsdk-linux-${DART_ARCH}-release.zip" > "/tmp/dart-sdk.zip" \
+  && DART_EXE="dartsdk-linux-${DART_ARCH}-release.zip" \
+  && DOWNLOAD_URL="https://storage.googleapis.com/dart-archive/channels/stable/release/${DART_VERSION}/sdk/${DART_EXE}" \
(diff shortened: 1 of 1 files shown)


jeffwidman (approved):
This looks nice and clean with the inlined-vars 👍

Dockerfile
```diff
@@ -270,16 +270,21 @@ ENV PUB_CACHE=/opt/dart/pub-cache \
...
-# TODO Dart now publishes SHA256 checksums for their releases, we should validate against those.
+ARG DART_VERSION=2.18.6
+
 RUN DART_ARCH=${TARGETARCH} \
   && if [ "$TARGETARCH" = "amd64" ]; then DART_ARCH=x64; fi \
-  && curl --connect-timeout 15 --retry 5 "https://storage.googleapis.com/dart-archive/channels/stable/release/${DART_VERSION}/sdk/dartsdk-linux-${DART_ARCH}-release.zip" > "/tmp/dart-sdk.zip" \
+  && DART_EXE="dartsdk-linux-${DART_ARCH}-release.zip" \
+  && DOWNLOAD_URL="https://storage.googleapis.com/dart-archive/channels/stable/release/${DART_VERSION}/sdk/${DART_EXE}" \
+  && curl --connect-timeout 15 --retry 5 "${DOWNLOAD_URL}" > "/tmp/${DART_EXE}" \
+  && curl --connect-timeout 15 --retry 5 "${DOWNLOAD_URL}.sha256sum" > "/tmp/${DART_EXE}.sha256sum" \
```
jeffwidman, line 280: Something I overlooked when merging this is that typically SHA sums are hardcoded into source so that if there's a MITM attack against the code, it's not pulling both the download file and the SHA over the same network connection if an attacker is trying to send a malformed payload... compare how we do it for `go`: https://github.com/dependabot/dependabot-core/blob/da1be808185659f8956836babd26c76acd65bdc6/go_modules/Dockerfile#L8-L10

Doing it this way with an automated download isn't useful against an attacker, but _is_ useful against network / local disk corruption, so this is still useful. Also I personally feel the risk is pretty low, and this is certainly more convenient... so I'm fine leaving as-is for now. If this starts to be a possible risk, it's trivial to switch to hardcoding the SHAs.

Issue #5756, opened 2022-09-20
Terraform modules in Azure Devops repos 404 due to incorrect fetch path


AurimasNav:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

terraform

### What you expected to see, versus what you actually saw

I've expected to have PR for terraform module version update created, instead I get an error.
```json
  :host          => "dev.azure.com"
  :local_address => "172.17.0.2"
  :local_port    => 33882
  :path          => "/aurimasnavardauskas/terraform/_git/tf-modules.git/info/refs"
  :port          => 443
  :reason_phrase => "Not Found"
  :remote_ip     => "13.107.42.20"
  :status        => 404
  :status_line   => "HTTP/1.1 404 Not Found\r\n"
```
the correct path in azure devops repos should be `/aurimasnavardauskas/terraform/_git/tf-modules/info/refs` without the appended `.git` to the `tf-modules` repository name.

I am aware that this project is not responsible for azure devops dependabot task implementation, however I've already reported this issue there and after investigation it seems that the issue is in dependabot-core rather than the azure devops task implementation.
The related issue and investigation: https://github.com/tinglesoftware/dependabot-azure-devops/issues/347

### Images of the diff or a link to the PR, issue, or logs

build logs:
https://dev.azure.com/aurimasnavardauskas/dependabot/_build/results?buildId=6&view=logs&j=42a276bf-7749-5ba3-b33d-570bc5e297a3&t=07ac1df1-2752-5bc5-77e1-e8fdbfbf1b81

### Smallest manifest that reproduces the issue

1. `terraform` project contains:
    - Teraform "module" repository:  https://dev.azure.com/aurimasnavardauskas/terraform/_git/tf-modules
2.  `dependabot` project contains:
    - Repo scanned by dependabot, terraform "deployment" code which utilizes the module from repo mentioned above: https://dev.azure.com/aurimasnavardauskas/dependabot/_git/iac-tf
    - The repo where dependabot pipeline is defined:  https://dev.azure.com/aurimasnavardauskas/dependabot/_git/dependabot

jeffwidman:
Thanks providing a reproducible example in a public repo in your report.

Looks like @mburumaxwell found [some useful debugging breadcrumbs](https://github.com/tinglesoftware/dependabot-azure-devops/issues/347#issuecomment-1251154358):
[...]

PR #6320, opened 2022-12-18
Added support for Azure in git_submodules

mburumaxwell:
Support `git_submodules` from Azure DevOps. All the code was present but not set up.

Changes in #6304 and #6306 helped a great deal too in handling submodule URLs for Azure DevOps:

```bash
[submodule "another-repo"]
	path = another-repo
	url = https://contoso@dev.azure.com/contoso/MyProject/_git/another-repo
```

The suffix `.git` does not in Azure DevOps. For GitHub and GitLab, it works with and without the suffix hence removed it just when getting the `git-upload-pack` in `service_pack_uri(...)`. For `dev.azure.com` and `*.visualstudio.com`, this suffix is skipped.

Fixes: #6340 
Fixes: #5756

### DIFF
--- a/common/lib/dependabot/git_metadata_fetcher.rb
+++ b/common/lib/dependabot/git_metadata_fetcher.rb
@@ -106,7 +106,7 @@ module Dependabot
 
     def fetch_raw_upload_pack_with_git_for(uri)
       service_pack_uri = uri
-      service_pack_uri += ".git" unless service_pack_uri.end_with?(".git")
+      service_pack_uri += ".git" unless service_pack_uri.end_with?(".git") || skip_git_suffix(uri)
 
       env = { "PATH" => ENV.fetch("PATH", nil) }
       command = "git ls-remote #{service_pack_uri}"
@@ -158,8 +158,25 @@ module Dependabot
     def service_pack_uri(uri)
       service_pack_uri = uri_with_auth(uri)
       service_pack_uri = service_pack_uri.gsub(%r{/$}, "")
-      service_pack_uri += ".git" unless service_pack_uri.end_with?(".git")
+      service_pack_uri += ".git" unless service_pack_uri.end_with?(".git") || skip_git_suffix(uri)
       service_pack_uri + "/info/refs?service=git-upload-pack"
+    end
+
+    def skip_git_suffix(uri)
+      # TODO: Unlike the other providers (GitHub, GitLab, BitBucket), as of 2023-01-18 Azure DevOps does not support the
+      # ".git" suffix. It will return a 404.
+      # So skip adding ".git" if looks like an ADO URI.
+      # There's no access to the source object here, so have to check the URI instead.
+      # Even if we had the current source object, the URI may be for a dependency hosted elsewhere.
+      # Unfortunately as a consequence, urls pointing to Azure DevOps Server will not work.
+      # Only alternative is to remove the addition of ".git" suffix since the other providers
+      # (GitHub, GitLab, BitBucket) work with or without the suffix.
+      # That change has other ramifications, so it'd be better if Azure started supporting ".git"
+      # like all the other providers.
+      uri = "https://#{uri.split('git@').last.sub(%r{:/?}, '/')}" if uri.start_with?("git@")
+      uri = URI(uri)
+      hostname = uri.hostname.to_s
+      hostname == "dev.azure.com" || hostname.end_with?(".visualstudio.com")
     end
 
     # Add in username and password if present in credentials.
--- a/common/spec/dependabot/metadata_finders/base/commits_finder_spec.rb
+++ b/common/spec/dependabot/metadata_finders/base/commits_finder_spec.rb
@@ -702,7 +702,7 @@ RSpec.describe Dependabot::MetadataFinders::Base::CommitsFinder do
 
     context "with a azure repo" do
       let(:service_pack_url) do
-        "https://dev.azure.com/contoso/MyProject/_git/business.git/info/refs" \
+        "https://dev.azure.com/contoso/MyProject/_git/business/info/refs" \
           "?service=git-upload-pack"
       end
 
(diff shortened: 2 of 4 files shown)


mburumaxwell:
@jeffwidman I realize this will not pass CI because of the removal of the `.git` suffix. It seems to affect quite much.
This affects the URLs generated in #6321 . Fetching in #6343 would also be affected (though it gracefully failsover).

How you suggest going about that?

mburumaxwell:
@jeffwidman 
Seeing as to how complex the removal of the `.git` suffix is, I have opted to exclude it only for `dev.azure.com` and `*.visualstudio.com` hostnames as there is no other way to tell if we should. This will not work for Azure DevOps Server but maybe that is a fight for another day.

jeffwidman:
👋thanks for digging into this. I'll probably be able to take a look this coming week. One of my focuses the next few weeks is catching up on our backlog of PRs.

jeffwidman (commented):
So the code looks reasonable given that today Azure doesn't respond to `*.git` URI's, unlike all other source control providers.

However, before merging this, I'd like to reach out to the ADO team and get their perspective... they might be willing to add support for these URI's given that all the other platforms already have this support so they're essentially lagging behind.

I realize that even if that gets on their roadmap, it won't solve your problem today, so we'll still probably want to merge this, but I'd feel better if we first have a conversation with them... they may also know some setting that we're unaware of to handle this.

Given that Microsoft owns GitHub, it shouldn't be too difficult for me to track down who I need to speak with there... I've pinged a few people so will keep you posted.

common/lib/dependabot/git_metadata_fetcher.rb
```diff
@@ -158,10 +158,25 @@ def parse_refs_for_upload_pack
...
 
+    def skip_git_suffix(uri)
+      # Azure DevOps does not support the ".git" suffix. It will return a 404.
+      # Why does it have to be so different?
+      # Doing this check here because there is no access to the source object.
+      # Even if we had the current source object, the URI may be for a dependency hosted elsewhere.
+      # Unfortunately as a consequence, urls pointing to Azure DevOps Server will not work.
+      # Only alternative is to remove the addition of ".git" suffix since the other providers
+      # (GitHub, GitLab, BitBucket) work with or without the suffix.
+      # That change is massive and needs its own PR given the impact.
```
jeffwidman, line 173:
```suggestion
      # TODO: Unlike the other providers (GitHub, GitLab, BitBucket), as of 2023-01-18 Azure DevOps does not support the ".git" suffix. It will return a 404.
      # So skip adding ".git" if looks like an ADO URI. 
      # There's no access to the source object here, so have to check the URI instead.
      # Even if we had the current source object, the URI may be for a dependency hosted elsewhere.
      # Unfortunately as a consequence, urls pointing to Azure DevOps Server will not work.
      # Only alternative is to remove the addition of ".git" suffix since the other providers
      # (GitHub, GitLab, BitBucket) work with or without the suffix.
      # That change has other ramifications, so it'd be better if Azure started supporting ".git" like all the other providers.
```

jeffwidman:
@lachie83 do you happen to know who on the ADO team I can chat with about ☝️? Happy to chat privately, my GitHub email/slack handle is same as my GitHub username.

mburumaxwell:
> So the code looks reasonable given that today Azure doesn't respond to `*.git` URI's, unlike all the other popular source control providers.
> 
> Before merging this, I'd like to reach out to the ADO team and get their perspective... they might be willing to add support to get to feature parity. Or they might know some setting/URL that we're unaware of to handle this.
> 
> I realize that even if that gets on their roadmap, it won't solve your problem today, so after an initial conversation to ensure we're not overlooking something obvious then we'll still probably move forward with merging this and then open a tracking ticket in our backlog to remind us to check down the road if they added support.
> 
> Given that Microsoft owns GitHub, it shouldn't be too difficult for me to track down who I need to speak with there... I've pinged a few people so will keep you posted.

I am also not too thrilled about the `.git` suffix not being supported on Azure DevOps while everyone else seems to do it. I would be really glad if they actually did add support for it but I will keep my expectations in check. It should ideally be easier for you to tell. According to https://docs.antora.org/antora/latest/playbook/git-suffix/#ensure-git-suffix-key, other people are also affected.

OTOH, [GitHub](https://github.com/dependabot/dependabot-core), [Gitlab](https://gitlab.com/dependabot-gitlab/dependabot), and [Bitbucket](https://bitbucket.org/atlassian_tutorial/helloworld) support URLs without the `.git` suffix. I haven't gotten to study the implications of not having the suffix as regards the internal workings of this repo but could not having the suffix for all actually be an option/solution?

deivid-rodriguez:
If `.git`-less URLs work everywhere, we may do that as well. It's a higher risk change, sure, but better in terms of future maintenance. Thoughts @jeffwidman?

deivid-rodriguez:
> But I do care about the code/UX experience being intuitive... and to me these URI's are more predictable/expected with .git suffixes... But I'm only n=1, so others may have a different perspective.

I would hope we can make the code good enough so that its readability does not depend on the specific shape of this URI. And its complexity should be about the same as now (less than the approach in this PR). Not sure how it affects the UX either, you mean in logs and so on?

Anyways, let's definitely hear from ADO people first!

mburumaxwell:
This also fixes #5756

waltervos:
I'm not sure if it's appreciated, but I just wanted to let you know that this change would mean a lot to my team :)

jeffwidman (approved):
Sorry for the delay, got buried here.

I just emailed the ADO team, but will probably take a few weeks to sync up with them.

In the meantime, my vote is we should ship this as-is for now... my thinking is that it's a small self-contained change, easy to back out later if ADO ends up adding support for `.git` suffix down the road.

Until we talk to them though, I hesitate to spend further time exploring dropping the `.git` suffix. @mburumaxwell already looked into it a bit and says it touches a number of things. And right now the only thing we're forgoing is Azure DevOps Server support. Plus, I actually _like_ the signaling of the `.git` suffix... it feels intuitive to me. 😄 

So that's why I vote we punt for now on that and figure out next steps after that ADO conversation happens.

What say you @deivid-rodriguez ?

deivid-rodriguez:
Ok!

mburumaxwell:
I'm in total agreement too.

PR #6321, opened 2022-12-19
Generate compare path and fetch commits for Azure sources

mburumaxwell:
Handle generation of compare path and fetching commits for Azure sources in the metadata finder for commits.

These were previously marked as `TODO`.

Fixes: #6341


jeffwidman (approved):
Thanks! I'll wait for CI to pass and then merge.

PR #6334, opened 2022-12-19
Update `monolog/monolog` fixture to latest packagist response

jeffwidman:
For a few upcoming PR's, it'll be helpful if the `monolog/monolog` packagist fixture is current to match what's _actually_ returned by https://repo.packagist.org/p/monolog/monolog.json.

Initially I updated all the other fixtures as well, but the `illuminate/*` fixtures ballooned to 43+ MB... and then when I switched them to the `v2` metadata endpoint (https://github.com/dependabot/dependabot-core/issues/3010) they shrank back down due to the new minifications... so no point in adding that much cruft to our git history when I'm going to almost immediately drop it.

But the `monolog/monolog` change is necessary because I'll be updating all the tests that point at the generic `packagist_response.json` to point at this fixture... and since it already has a newer version than is in `packagist_response.json`, might as well first ensure the fixture is fully-up-to-date.

### DIFF
--- a/composer/spec/dependabot/composer/update_checker_spec.rb
+++ b/composer/spec/dependabot/composer/update_checker_spec.rb
@@ -207,7 +207,7 @@ RSpec.describe Dependabot::Composer::UpdateChecker do
 
     context "when the user is ignoring the latest version" do
       let(:ignored_versions) { [">= 1.22.0.a, < 3.0"] }
-      it { is_expected.to eq(Gem::Version.new("1.21.0")) }
+      it { is_expected.to eq(Gem::Version.new("2.8.0")) }
     end
 
     context "without a lockfile" do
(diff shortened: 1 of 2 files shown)


jakecoffman (approved):

composer/spec/dependabot/composer/update_checker_spec.rb
```diff
@@ -207,7 +207,7 @@
 
     context "when the user is ignoring the latest version" do
       let(:ignored_versions) { [">= 1.22.0.a, < 3.0"] }
-      it { is_expected.to eq(Gem::Version.new("1.21.0")) }
+      it { is_expected.to eq(Gem::Version.new("2.8.0")) }
```
jeffwidman, line 210: These should have stayed at `"1.21.0"`, given that we're ignoring `[">= 1.22.0.a, < 3.0"]`... Fixing that in https://github.com/dependabot/dependabot-core/pull/6611

Issue #6298, opened 2022-12-09
Drop support for PHP Composer v1


jeffwidman:
Today we support both Composer v1 and v2.

However, Composer v1 is [deprecated as of Feb 25, 2021](https://blog.packagist.com/deprecating-composer-1-support/).

At some point we'll stop supporting it here in `dependabot-core` as well. It's not urgent, as it's currently working fine and generally not a huge amount of overhead to maintain. But over time we plan to make bits of `dependabot-core` more modular by ecosystem, and doing that will be easier if we only have one version of composer to support.

So this is a placeholder ticket to track any discussions around timing, things that we'll need to update/remove, etc.

If you're a user of Dependabot who is still on Composer v1 and you cannot upgrade to Composer v2 for some reason, please chime in here explaining your use case.

jeffwidman:
@stof @Seldaek IIUC you two have back-end access to Packagist... do you have any metrics on current Composer v1 vs v2 usage rates? It'd be very helpful to see https://packagist.org/php-statistics but for composer versions as we gauge the timing of when to drop support for composer v1.

Seldaek:
@jeffwidman we are currently at 77% of installs (can be passive/ci/..) and 86% of updates (more active) done with composer v2, so v1 is well on its way out.

stof:
@jeffwidman by curiosity, do you have any stats about the ratio of dependabot updates being done using v1 vs v2 (for the GitHub-managed dependabot service) ?

jeffwidman:
Sorry for the delay here @Seldaek @stof, we didn't have this instrumented until #7323  

The most useful way of measuring this is by looking at the percentage of job configs that result in `composer` `1` vs `2`. Looking at total job percentages isn't as useful because some jobs are configured to run daily, vs other monthly, so some could have 30x the impact.

As of this morning, 92.2% of job configs result in `composer` `2`, so only 7.8% result in `composer` `1`.

This is lower than I expected, so I suspect that we'll drop `composer` `1` in the not too distant future:
* #7643

abdulapopoola:
tagging @jonjanego as fyi

stefangr:
Support for composer v1 in packagist.org has an official end date.
https://blog.packagist.com/shutting-down-packagist-org-support-for-composer-1-x/

abdulapopoola:
> Support for composer v1 in packagist.org has an official end date. https://blog.packagist.com/shutting-down-packagist-org-support-for-composer-1-x/

[...]

PR #6350, opened 2022-12-21
Clarify `composer` `v1` `PEAR` unit test

jeffwidman:
It took me a while to understand this unit test, because this package _is_ actually present on packagist: https://repo.packagist.org/p/pear-pear.horde.org/horde_date.json

However, from the associated `composer.json` fixture plus the existing fixture, it's clear that this test is checking what happens when a PEAR dependency is _not_ present on packagist.

So remove the fixture from the `packagist_responses` directory which is supposed to match what Packagist actually returns and instead hardcode the response for this single unit test to avoid confusion of future devs thinking they need to update it to match the actual packagist response.

This will make the https://github.com/dependabot/dependabot-core/pull/6315 a bit more straightforward because composer v1 doesn't know about the v2 metadata API so I will need to hardcode both v1 and v2 metadata responses (which return different HTTP status codes).

PEAR is mostly dead it seems, but it doesn't hurt to leave it around in case any old composer v1 projects still have PEAR deps listed... and this will all be gone soon enough once we remove support for `composer` `v1`:
* https://github.com/dependabot/dependabot-core/issues/6298

### DIFF
--- a/composer/spec/dependabot/composer/update_checker_spec.rb
+++ b/composer/spec/dependabot/composer/update_checker_spec.rb
@@ -434,6 +434,16 @@ RSpec.describe Dependabot::Composer::UpdateChecker do
         }]
       end
 
+      # This unit test is testing that a dependency located on https://pear.horde.org is still correctly
+      # handled by composer. So ignore the fact that this package actually exists on packagist, and
+      # pretend it just 404's.
+      let(:packagist_response) { '{"error":{"code":404,"message":"Not Found"}}' }
+      before do
+        v1_metadata_url = "https://repo.packagist.org/p/#{dependency_name.downcase}.json"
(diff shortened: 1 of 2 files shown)


jeffwidman:
This test has been a royal pain in https://github.com/dependabot/dependabot-core/pull/6315 because I have to ensure I stub both metadata API responses correctly, so I needed an additional fixture file, different HTTP error codes etc. 

Plus the test was just plain confusing since packagist does return this response and PEAR itself is mostly dead... eg this blog post from 2014: http://fabien.potencier.org/the-rise-of-composer-and-the-fall-of-pear.html

So this refactor makes stubbing both API's pretty straightforward so doesn't hurt to keep it around til we _actually_ drop composer v1 support...

honeyankit (approved):

Issue #6341, opened 2022-12-20
Commit diff URL and commits list missing for Azure-based sources/metadata


mburumaxwell:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

any

### What you expected to see, versus what you actually saw

The metadata finder for commits should produce a URL for the diff e.g. https://dev.azure.com/MyOrg/MyProj/_git/myrepo/branchCompare?baseVersion=GT1.0.0&targetVersion=GT1.0.2 for when a dependency is updated from 1.0.0 to 1.0.2 and a list of commits.

Instead, it generates https://dev.azure.com/MyOrg/MyProj/_git/myrepo/commits and no list of commits.

Having this is useful for packages hosted on Azure Artifacts and when updating git-submodules for Azure-based repositories.

PR #6351, opened 2022-12-22
Add CommitFinder tests/specs for Azure added in #6321

mburumaxwell:
Tests/specs that should have accompanied #6321

Related to #6341

### DIFF
--- a/common/spec/dependabot/metadata_finders/base/commits_finder_spec.rb
+++ b/common/spec/dependabot/metadata_finders/base/commits_finder_spec.rb
@@ -700,6 +700,116 @@ RSpec.describe Dependabot::MetadataFinders::Base::CommitsFinder do
       end
     end
 
+    context "with a azure repo" do
+      let(:service_pack_url) do
+        "https://dev.azure.com/contoso/MyProject/_git/business.git/info/refs" \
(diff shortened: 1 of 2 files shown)


jeffwidman (approved):
Thank you for this! I had checked the code attached to those PR's when I merged it and for some reason thought we didn't have BitBucket, GitLab etc unit tests so hadn't asked for any on Azure. 

So really appreciate you noticing the tests are off in a different place and then adding them!

Should make your life easier down the road too as the Azure wrapper maintainer as less likely a bug will creep in here in the future... 😄

mburumaxwell:
> Thank you for this! I had checked the code attached to those PR's when I merged it and for some reason thought we didn't have BitBucket, GitLab etc unit tests so hadn't asked for any on Azure.
> 
> So really appreciate you noticing the tests are off in a different place and then adding them!
> 
> Should make your life easier down the road too as the Azure wrapper maintainer as less likely a bug will creep in here in the future... 😄

Oh well! It was the first time writing tests/specs in Ruby, realized I would dodge many bullets by making sure these are present too. Especially, given that most Azure DevOps users will create an issue over at the extension repo first. Thanks for your guide.

PR #6357, opened 2022-12-22
fix OOM for large metadata responses and many dependencies

jakecoffman:
Dependabot was running out of memory when faced with updating many dependencies that had large metadata responses to download (releases, commits, etc).

I found the issue was that these classes that are memoized are holding onto the large responses, so the memory won't be cleared until the entire metadata generation step is complete. (A finder for each dependency is also memoized in the MessageBulder.) 

So by letting the class instances go out of scope here they can be GC'd along with their large response bodies, but we keep the strings we need for metadata.

### DIFF
--- a/common/lib/dependabot/metadata_finders/base.rb
+++ b/common/lib/dependabot/metadata_finders/base.rb
@@ -11,11 +11,39 @@ module Dependabot
 
       PACKAGE_MANAGERS_WITH_RELIABLE_DIRECTORIES = %w(npm_and_yarn pub).freeze
 
-      attr_reader :dependency, :credentials
+      attr_reader :dependency, :credentials, :changelog_url, :changelog_text, :upgrade_guide_url, :upgrade_guide_text,
+                  :releases_url, :releases_text, :commits_url, :commits
 
       def initialize(dependency:, credentials:)
         @dependency = dependency
         @credentials = credentials
+
+        # Purposefully not memoizing these class instances, as they're heavy to keep in memory for large responses.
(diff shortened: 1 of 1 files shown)


jakecoffman:
Oof! Looks like we rely on this implementation detail to mock out a lot of tests.

deivid-rodriguez:
Awesome find and clean solution 👏 👏

common/lib/dependabot/metadata_finders/base.rb
```diff
@@ -11,99 +11,51 @@ class Base
...
-      attr_reader :dependency, :credentials
+      attr_reader :dependency, :credentials, :changelog_url, :changelog_text, :upgrade_guide_url, :upgrade_guide_text,
+                  :releases_url, :releases_text, :commits_url, :commits
 
       def initialize(dependency:, credentials:)
         @dependency = dependency
         @credentials = credentials
-      end
-
-      def source_url
```
jeffwidman, line 21: We chatted about this internally, but wanted to mention here publicly as well... moving everything into the initialization just seems a bit off... suddenly instantiating an instance requires network I/O and if I want only a single method or two, I now have to get everything... I wasn't the only person with some hesitations about this path forward, so the problem/debugging is awesome, but we may end up pursuing a different/more flexible solution.

deivid-rodriguez: You're absolutely right!

jakecoffman:
Closing since this isn't the right solution.

Issue #6330, opened 2022-12-19
Metadata doesn't preserve newlines from plain text changelogs


jeffwidman:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

Any

### What you expected to see, versus what you actually saw

Metadata in https://github.com/dependabot/dependabot-core/pull/6327 is borked, it doesn't preserve the newlines from https://github.com/excon/excon/blob/2d1e2d9fbb6dbfd1e49e16e8aa41876220b3b318/changelog.txt#L4-L7 but instead strips them out making it difficult to read.

jeffwidman:
Might be a dupe of:
* https://github.com/dependabot/dependabot-core/issues/2367
?

PR #6358, opened 2022-12-22
Add `HARDBREAKS` to commonmarker config

blue2cat:
The commonmarker markdown parsing gem by default does not make all line breaks hard line breaks. This results in newlines being ignored in changelog summaries created by Dependabot. For example, in #6327 the changelog messages include no line breaks between items.

Original upstream changelog:
```
tweak readline code style
change read_nonblock to again limit to max_length
update bundled (test) certs
remove rdoc dependency to fix CI
```

Dependabot changelog before this commit:
```
tweak readline code style change read_nonblock to again limit to max_length update bundled (test) certs remove rdoc dependency to fix CI
```

Dependabot changelog after this commit:
```
tweak readline code style
change read_nonblock to again limit to max_length
update bundled (test) certs
remove rdoc dependency to fix CI
```

Fix https://github.com/dependabot/dependabot-core/issues/6330


deivid-rodriguez (commented):
Nice! Would you be willing to add a spec for this? Willing to help if needed :)

blue2cat:
@deivid-rodriguez  I've added a spec. There are still about 10 failing checks, I've left them in for now so it's easy to see the places where adding the `HARDBREAKS` flag to `commonmarker` has changed the output. The majority of failures are due to an extra `<br />` tag at the end of the output.

deivid-rodriguez:
Looking into those failures, I think we may need to scope the new option to changelogs in plain text? Otherwise it may introduce some artificial breaks for changelogs that render just fine at the moment, right?

jeffwidman:
>  I think we may need to scope the new option to changelogs in plain text?

That'd be brilliant if we could do that... I really like @blue2cat 's clean solution of modifying the upstream config. But didn't like how it changed PR metadata that didn't need it... totally didn't even cross my mind that we might be able to flip it on/off depending on the type of content. 👍

blue2cat:
That's a great idea! I'll work on only applying this parameter to plaintext changelogs.

blue2cat:
I've implemented logic that checks the format of the incoming changelog file. I used a regex that matches valid Markdown headers and links to determine if a given text block should have the `HARDBREAKS` config flag applied. 

Additionally, I also have a related security question. On line 37 of [link_and_mention_sanitizer.rb](https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb#L37), the default for the `unsafe` parameter is `false`. Would it be better if this defaulted to `true` for security reasons?

deivid-rodriguez:
@blue2cat Thanks for working on this! Can you check the CI failures? They seem definitely related to this PR, but not sure if they are expected or not. You can see the unexpected diff in the next job step after the one failing, labeled as "Diff".

jeffwidman:
@blue2cat still interested in working on this? Be great to ship it

blue2cat:
@jeffwidman I'm still interested in working on this. I'll try to dedicate some time to it soon.

common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
```diff
@@ -54,9 +58,23 @@ def sanitize_links_and_mentions(text:, unsafe: false, format_html: true)
           sanitize_nwo_text(doc)
 
           mode = unsafe ? :UNSAFE : :DEFAULT
-          return doc.to_commonmark([mode] + COMMONMARKER_OPTIONS) unless format_html
 
-          doc.to_html(([mode] + COMMONMARKER_OPTIONS), COMMONMARKER_EXTENSIONS)
+          # If the text does not contain markdown, we need to use the HARDBREAKS option
+          commonmarker_render_options = if text.length > 1000
```
jeffwidman, line 63: I assume that the check for text length is because you're trying to work around the CodeQL issue... but since we're not really worried about the CodeQL failure here, this code is overly complex.

Can you simplify it back to what you had earlier that simply checks for markdown and if not uses the `HARDBREAKS` option?

I think that will simplify stuff quite a bit here...

common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
```diff
@@ -54,9 +58,23 @@ def sanitize_links_and_mentions(text:, unsafe: false, format_html: true)
...
+                                          [COMMONMARKER_OPTIONS]
+                                        else
+                                          # If text length is manageable, use regex to determine rendering options
+                                          (unless text.match?(MARKDOWN_REGEX)
+                                             COMMONMARKER_OPTIONS + [:HARDBREAKS]
+                                           end) || [COMMONMARKER_OPTIONS]
+                                        end
+
+          options = [mode] + commonmarker_render_options
+          options = options.flatten.uniq
```
jeffwidman, line 73: Curious, why'd you have to do this? Since it wasn't there previously, just surprised to see it... Was the duplicate present previously and we should have been doing this all along? Or does this new code somehow result in duplicates that need to be cleaned up?

common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
blue2cat: Hey @jeffwidman, thanks for the review. I've simplified the logic back to my initial commit.

common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
blue2cat: This code was related to the regex timeout logic. Since the timeout is not needed, I've discarded lines 72 and 73.

jeffwidman (approved):
LGTM 👍 

I'm refraining from merging stuff right now as I'm on a 🏕️  trip and if there's any issues I may not be around to fixup/revert, so if no one else gets to it before then I'll 🚢 when I return.

Anyone else on the :dependabot: feel free to ship this in the meantime. 😄

PR #6359, opened 2022-12-22
Use source for GitHub client to support fetching metadata from GHES instances

mctofu:
We weren't providing the source here so we'd inadvertently load the metadata from github.com instead. The source provides the hostname & api endpoint to use when making requests.

codeql is flagging an issue in the changelog finder but since it's not related to this change I think we should address it in a separate PR.

### DIFF
--- a/common/lib/dependabot/metadata_finders/base/changelog_finder.rb
+++ b/common/lib/dependabot/metadata_finders/base/changelog_finder.rb
@@ -82,11 +82,12 @@ module Dependabot
           return unless suggested_changelog_url
 
           # TODO: Support other providers
-          source = Source.from_url(suggested_changelog_url)
-          return unless source&.provider == "github"
+          suggested_source = Source.from_url(suggested_changelog_url)
+          return unless suggested_source&.provider == "github"
 
-          opts = { path: source.directory, ref: source.branch }.compact
-          tmp_files = github_client.contents(source.repo, opts)
+          opts = { path: suggested_source.directory, ref: suggested_source.branch }.compact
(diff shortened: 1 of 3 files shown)


honeyankit (approved):

deivid-rodriguez (approved):
Nice! 💪

common/lib/dependabot/metadata_finders/base/changelog_finder.rb
```diff
@@ -346,7 +347,14 @@ def gitlab_client
 
         def github_client
           @github_client ||= Dependabot::Clients::GithubWithRetries.
-                             for_github_dot_com(credentials: credentials)
+                             for_source(source: source, credentials: credentials)
+        end
+
+        def github_client_for_source(client_source)
```
jeffwidman, line 353: Maybe add a code comment to this method explaining that in addition to cloud, GHES also uses the GitHub client... because if I didn't have the context of this PR, I'd probably forget about GHES and wonder why we need to specify a source for GitHub, since I'm so used to thinking primarily of cloud.

jeffwidman (approved):

common/lib/dependabot/metadata_finders/base/changelog_finder.rb
mctofu: I'd like to eventually remove all the usages of `Dependabot::Clients::GithubWithRetries.for_github_dot_com(credentials: credentials)` as there's only a few left after this and that would make it harder to ignore the source.

We should support source for the other clients as well but I'm not well equipped to test them all out.

PR #6361, opened 2022-12-23
[gomod] Support updating indirect dependencies

mctofu:
This is an improved take on an [earlier spike](https://github.com/dependabot/dependabot-core/pull/5555) I did to support updating indirect dependencies in a security update. This allows updating indirect dependencies that are listed in the `go.mod` file in either version or security updates. Dependencies that are only included in the `go.sum` and not the `go.mod` still can't be updated but using the go 1.17+ `go.mod` file any used indirect dependencies should be contained in the `go.mod`.

For version updates, the default allow filter is to only update direct dependencies so indirect dependencies will only be updated if a user configures their `dependabot.yml` to do so.

A benefit over my previous attempt is that this approach maintains the indirect dependency distinction so using the https://github.com/dependabot/fetch-metadata to inspect the commit metadata for the PR will correctly indicate an indirect dependency is being updated.

Another edge case I ran into with https://github.com/dependabot/dependabot-core/pull/6361/commits/bdb3ae3854f3fbee61b28955b0e5c0d8d428df6a is that we previously parsed "replaced" dependencies but removed their requirements. That caused us to treat them as indirect dependencies and not attempt to update them. I've preserved the behavior of not updating replaced dependencies (until we can implement a way to properly update them) by skipping them during the file parsing step.

The go smoke test is currently failing but that's expected. It needs to be updated to:
- Use the default allow condition so indirect deps aren't updated
- Update the expectation for the submitted dependency list which will now include indirect deps but exclude replaced deps.

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_parser.rb
+++ b/go_modules/lib/dependabot/go_modules/file_parser.rb
@@ -155,12 +155,10 @@ module Dependabot
       end
 
       def skip_dependency?(dep)
-        begin
-          path_uri = URI.parse("https://#{dep['Path']}")
-          !path_uri.host.include?(".")
-        rescue URI::InvalidURIError
-          false
-        end
+        path_uri = URI.parse("https://#{dep['Path']}")
+        !path_uri.host.include?(".")
+      rescue URI::InvalidURIError
+        false
       end
 
       def dependency_is_replaced(details)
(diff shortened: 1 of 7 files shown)


Marudhas (commented):
Ad

Marudhas (commented):
Af

jakecoffman (approved):
Makes sense to me!

jeffwidman:
> For version updates, the default allow filter is to only update direct dependencies so indirect dependencies will only be updated if a user configures their dependabot.yml to do so.

👍 Thanks for clarifying this, it was my one question after looking at the PR.

PR #6390, opened 2023-01-08
Pin codespace ruby version to match the rest of the repo

jeffwidman:
This was coming up as `3.2.0`, which was inconvenient when I was trying to regenerate the `Gemfile.lock` etc... life is simpler if we pin this to match the ruby version we use in the rest of `dependabot-core` and then update it in lockstep.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -23,7 +23,7 @@
     "github-cli": "latest",
     "node": "lts",
     "golang": "latest",
-    "ruby": "latest",
+    "ruby": "3.1.3",
     "rust": "latest",
     "dotnet": "latest",
     "ghcr.io/devcontainers/features/sshd:1": {


deivid-rodriguez (approved):

jeffwidman:
Agreed, but I'd looked and couldn't find a way to have it do that?

Nishnha:
> Agreed, but I'd looked and couldn't find a way to have it do that?

I couldn't find any feature that would let us do that either. 

Closest thing I could come up with is to [not specify any version](https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/docs/ruby.md#syntax) in the devcontainer (so no ruby version is installed, but rvm is), and then installing the ruby version we want with rvm during [the on-create script](https://github.com/dependabot/dependabot-core/blob/main/.devcontainer/on-create.sh)

If that's too extra, maybe we can just document that this field also needs to be bumped when we update ruby?

Either way, these are just nits and shouldn't be considered blocking

deivid-rodriguez:
Either of those seem fine to me 👍.

jeffwidman:
I'll just leave this as-is honestly, when I bump Ruby, I generally grep for the version string and it should pop up easily that way if someone greps for `3.1.3`

PR #6391, opened 2023-01-08
Install `cmake` by default in codespace

jeffwidman:
For some of the Dependabot PR's that bump gems within the Updater, we have to run `bundle install` within the `updater` subfolder in order to regenerate the updater's `Gemfile.lock`.

Unfortunately, doing that within the codespace was failing due to a missing `cmake`. So let's install it by default for convenience.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -5,7 +5,8 @@ ARG VARIANT="jammy"
 FROM mcr.microsoft.com/vscode/devcontainers/base:0-${VARIANT}
 
 # [Optional] Uncomment this section to install additional OS packages.
-# RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
-#     && apt-get -y install --no-install-recommends <your-package-list-here>
+RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+    && apt-get -y install --no-install-recommends \
+      cmake  # required for `bundle install` within the updater
 
 


.devcontainer/Dockerfile
```diff
@@ -5,7 +5,8 @@ ARG VARIANT="jammy"
 FROM mcr.microsoft.com/vscode/devcontainers/base:0-${VARIANT}
 
 # [Optional] Uncomment this section to install additional OS packages.
-# RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
-#     && apt-get -y install --no-install-recommends <your-package-list-here>
+RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+    && apt-get -y install --no-install-recommends \
+      cmake  # required for `bundle install` within the updater
```
jeffwidman, line 10: hmm... opening this PR branch in a codespace shows that `cmake` still isn't installed... 

I suspect it's because at this point we aren't root in the container: https://github.com/dependabot/dependabot-core/blob/c5557f9a15e54033350cb68e70adea92c86bec63/.devcontainer/devcontainer.json#L20

But not sure if/where the error output from this command might get logged in order to prove that?

Or is the problem that the codespace even if run on a branch isn't loading the `.devcontainer/Dockerfile` from that branch but instead from `main`? Dunno how well it tracks if you change the codespace configuration on a branch and then load a codespace on that branch...

Nishnha: you should be able to open a codespace, run `git pull`, checkout this branch, and then run `script/setup` to test whether cmake installs correctly

deivid-rodriguez: Would it make sense to use `Dockerfile.updater` as the base `Dockerfile`, like we do for core?

jeffwidman: Thanks to some help from @dedac, I realized I was wrong and `cmake` is installed by this PR!!

```shell
@jeffwidman ➜ /workspaces/dependabot-core (install-cmake-by-default-within-codespace ✗) $ which make
/usr/bin/make
@jeffwidman ➜ /workspaces/dependabot-core (install-cmake-by-default-within-codespace ✗) $ make --version
GNU Make 4.3
Built for x86_64-pc-linux-gnu
Copyright (C) 1988-2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

jeffwidman: > Would it make sense to use Dockerfile.updater as the base Dockerfile, like we do for core?

I'd be leery of making `core` dependent upon `updater` in any fashion given that the former is a library and the latter is an app so some of our external consumers might not be using it in any way. But let's see how the work from @jakecoffman / @pavera shakes out...

deivid-rodriguez: I guess I misunderstood this. I thought this devcontainer was for developing `updater/` and the core specific one was https://github.com/dependabot/dependabot-core/blob/c5557f9a15e54033350cb68e70adea92c86bec63/.devcontainer/core-dev/devcontainer.json.

jeffwidman: Oh, maybe I misunderstood! 

Hmm... if it's for dev'ing updater, then yeah, we probably _should_ make it dependant upon `Dockerfile.updater`... dunno...

@jakecoffman what say you? Since you currently have the most context since split these up earlier and also working on the Dockerfile stuff??

jakecoffman: I think Core is just a leftover concept from before we open sourced the updater image. Maybe we didn't do a great job of cleaning up that concept because we didn't need to.

Now that updater is in here we can just treat the updater image as where we do all our testing and developing since it's what we run in production.

And once we break all the updater images up there won't be a core anymore at all, how could there be?

Nishnha: I think we would actually want to make a Docker image for codespaces that has every ecosystem's updater image installed, plus devtools

Nishnha: > I guess I misunderstood this. I thought this devcontainer was for developing `updater/` and the core specific one was https://github.com/dependabot/dependabot-core/blob/c5557f9a15e54033350cb68e70adea92c86bec63/.devcontainer/core-dev/devcontainer.json.

`.devcontainer/devcontainer.json` is the devcontainer build for the Dependabot CLI I think? So when you run the Debugger that's the container it drops you in

`.devcontainer/core-dev/devcontainer.json` should be the codespace devcontainer

jakecoffman: `.devcontainer/devcontainer.json` is supposed to be the "normal" Codespace experience where it drops you into a workspace. It does try to install the CLI too which I think I need to fix.

I think if we want to continue to have a devcontainer that drops you inside the updater container, we're going to need one for each updater. Otherwise, we're going to have to maintain a Dockerfile that will get out of sync with the smaller image Dockerfiles.

Personally I think it makes more sense to use the "normal" Codespace and use scripts to run tests, etc.

PR #6375, opened 2023-01-04
Support the latest Python patch versions as of 2023-01-04

phillipuniverse:
3.11.1, 3.10.9, 3.9.16, 3.8.16, 3.7.16

PR #6404, opened 2023-01-10
Use latest Python patch releases for builds

phillipuniverse:
Follow-up to #6375 this activates those latest patch releases during runtime


jeffwidman:
Oh I didn't realize that, thanks for following up.

jeffwidman (approved):

PR #6409, opened 2023-01-10
Metadata fetching: Don't fallback to `git` if not installed

jeffwidman:
This git fallback is being hit whenever there's a non-`200` response code. However, if `git` isn't installed, this will result in throwing:
```shell
Errno::ENOENT with No such file or directory - git
```

So only fallback to `git` if installed.

### DIFF
--- a/common/lib/dependabot/git_metadata_fetcher.rb
+++ b/common/lib/dependabot/git_metadata_fetcher.rb
@@ -72,8 +72,14 @@ module Dependabot
       response = fetch_raw_upload_pack_for(uri)
       return response.body if response.status == 200
 
-      response_with_git = fetch_raw_upload_pack_with_git_for(uri)
-      return response_with_git.body if response_with_git.status == 200
+      # if git is installed, try fetching with it in case something we need is configured in .gitconfig (like creds)
+      env = { "PATH" => ENV.fetch("PATH", nil) }
+      command = SharedHelpers.escape_command("git --version")
+      stdout, stderr, process = Open3.capture3(env, command)
+      if process.success?
+        response = fetch_raw_upload_pack_with_git_for(uri)
+        return response.body if response.status == 200
+      end
 
       raise Dependabot::GitDependenciesNotReachable, [uri] unless uri.match?(KNOWN_HOSTS)
 
@@ -109,15 +115,20 @@ module Dependabot
       service_pack_uri += ".git" unless service_pack_uri.end_with?(".git")
 
       env = { "PATH" => ENV.fetch("PATH", nil) }
-      command = "git ls-remote #{service_pack_uri}"
-      command = SharedHelpers.escape_command(command)
 
+      command = SharedHelpers.escape_command("git --version") # Is git present?
       stdout, stderr, process = Open3.capture3(env, command)
-      # package the command response like a HTTP response so error handling
-      # remains unchanged
+      if process.success?
+        command = "git ls-remote #{service_pack_uri}"
+        command = SharedHelpers.escape_command(command)
+        stdout, stderr, process = Open3.capture3(env, command)
+      end
+
+      # package the command response like a HTTP response so error handling remains unchanged
       if process.success?
         OpenStruct.new(body: stdout, status: 200)
       else
+        # this picks up the response from either `git --version` or `git ls-remote`, whichever fails first
         OpenStruct.new(body: stderr, status: 500)
       end
     end


common/lib/dependabot/git_metadata_fetcher.rb
```diff
@@ -72,8 +72,14 @@ def fetch_upload_pack_for(uri)
       response = fetch_raw_upload_pack_for(uri)
       return response.body if response.status == 200
 
-      response_with_git = fetch_raw_upload_pack_with_git_for(uri)
-      return response_with_git.body if response_with_git.status == 200
```
jeffwidman, line 76: Since this is `response_with_git` rather than `response`, all the subsequent error handling is only handling the response from the initial excon request... I'm not sure if that was intentional, given the later comment `"package the command response like a HTTP response so error handling remains unchanged"` so I flipped it to `response`...

but either way could cause confusion...

common/lib/dependabot/git_metadata_fetcher.rb
```diff
@@ -72,8 +72,14 @@ def fetch_upload_pack_for(uri)
...
+      # if git is installed, try fetching with it in case something we need is configured in .gitconfig (like creds)
```
jeffwidman, line 75: Approach 1: gate the fallback at the callsite... it's more clear exactly what/why we're doing this

common/lib/dependabot/git_metadata_fetcher.rb
```diff
@@ -72,8 +72,14 @@ def fetch_upload_pack_for(uri)
...
+      env = { "PATH" => ENV.fetch("PATH", nil) }
+      command = SharedHelpers.escape_command("git --version")
+      stdout, stderr, process = Open3.capture3(env, command)
+      if process.success?
+        response = fetch_raw_upload_pack_with_git_for(uri)
+        return response.body if response.status == 200
```
jeffwidman, line 81: this may be better left as `response_with_git` so that the response doesn't hit the later error handling, I wasn't sure...

common/lib/dependabot/git_metadata_fetcher.rb
```diff
@@ -109,15 +115,20 @@ def fetch_raw_upload_pack_with_git_for(uri)
       service_pack_uri += ".git" unless service_pack_uri.end_with?(".git")
 
       env = { "PATH" => ENV.fetch("PATH", nil) }
-      command = "git ls-remote #{service_pack_uri}"
-      command = SharedHelpers.escape_command(command)
 
+      command = SharedHelpers.escape_command("git --version") # Is git present?
```
jeffwidman, line 119: Approach 2: gate internally within the method... this way if we ever call this method from somewhere else, all callers will inherit the gate...

deivid-rodriguez (commented):
Maybe it's simpler to rescue `Errno::ENOENT` directly from running `git ls-remote`? I don't think that error can be raised in any situation other than `git` not being in `PATH`.

common/lib/dependabot/git_metadata_fetcher.rb
```diff
@@ -109,15 +115,20 @@ def fetch_raw_upload_pack_with_git_for(uri)
...
+        command = "git ls-remote #{service_pack_uri}"
+        command = SharedHelpers.escape_command(command)
+        stdout, stderr, process = Open3.capture3(env, command)
+      end
+
+      # package the command response like a HTTP response so error handling remains unchanged
       if process.success?
         OpenStruct.new(body: stdout, status: 200)
       else
+        # this picks up the response from either `git --version` or `git ls-remote`, whichever fails first
```
jeffwidman, line 131: the fact that I needed to add a code comment may indicate this is overly clever, not sure...

jeffwidman (commented):
annotating the changes

jeffwidman:
Switched to rescuing `Errno::ENOENT` as suggested by @deivid-rodriguez 

This is ready for review. 

I'm not sure:

1. how to write a unit test, as not sure how to mock out the underlying `git` is missing? 
2. If this actually needs a unit test, as it should be fairly rare/low impact, we only saw this happen briefly internally when we experimented with using a different base container for running :dependabot:... my guess is most users who use :dependabot: open source will be running on systems that have `git` installed. But happy to add if reviewers think it's worth the time cost of figuring out how to do it...

deivid-rodriguez (approved):

PR #6418, opened 2023-01-12
remove Dockerfile.ci as it is redundant

jakecoffman:
Dockerfile.updater does the same work as Dockerfile.ci to copy source into the container. The benefit of this change is we have one less Dockerfile to think about!


honeyankit (approved):
Less code is better :-)

deivid-rodriguez (approved):
😍

PR #5625, opened 2022-09-05
Use a standard filesystem structure in all docker images

deivid-rodriguez:
And make it the same as in our repository.

While working on improving #5617, I wanted to change the way we install native helpers to be like this:

```diff
diff --git a/Dockerfile b/Dockerfile
index 566941768..0b858d95d 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -267,32 +267,32 @@ USER dependabot
 
 ENV DEPENDABOT_NATIVE_HELPERS_PATH="/opt"
 
-COPY --chown=dependabot:dependabot composer/helpers /opt/composer/helpers
-RUN bash /opt/composer/helpers/v1/build \
-  && bash /opt/composer/helpers/v2/build
+COPY --chown=dependabot:dependabot composer/helpers /home/dependabot/dependabot-core/composer/helpers
+RUN bash /home/dependabot/dependabot-core/composer/helpers/v1/build \
+  && bash /home/dependabot/dependabot-core/composer/helpers/v2/build
 
-COPY --chown=dependabot:dependabot bundler/helpers /opt/bundler/helpers
-RUN bash /opt/bundler/helpers/v1/build \
-  && bash /opt/bundler/helpers/v2/build
+COPY --chown=dependabot:dependabot bundler/helpers /home/dependabot/dependabot-core/bundler/helpers
+RUN bash /home/dependabot/dependabot-core/bundler/helpers/v1/build \
+  && bash /home/dependabot/dependabot-core/bundler/helpers/v2/build
 
-COPY --chown=dependabot:dependabot go_modules/helpers /opt/go_modules/helpers
-RUN bash /opt/go_modules/helpers/build
+COPY --chown=dependabot:dependabot go_modules/helpers /home/dependabot/dependabot-core/go_modules/helpers
+RUN bash /home/dependabot/dependabot-core/go_modules/helpers/build
 
-COPY --chown=dependabot:dependabot hex/helpers /opt/hex/helpers
-ENV MIX_HOME="/opt/hex/mix"
-RUN bash /opt/hex/helpers/build
+COPY --chown=dependabot:dependabot hex/helpers /home/dependabot/dependabot-core/hex/helpers
+ENV MIX_HOME="/home/dependabot/dependabot-core/hex/mix"
+RUN bash /home/dependabot/dependabot-core/hex/helpers/build
 
-COPY --chown=dependabot:dependabot pub/helpers /opt/pub/helpers
-RUN bash /opt/pub/helpers/build
[...]

PR #6420, opened 2023-01-12
Try fix CI

deivid-rodriguez:


jakecoffman:
I'm guessing #6418 only appeared to pass because it skipped all the builds? 🤦

bin/ci-test
```diff
@@ -35,7 +35,7 @@ function build() {
   export DOCKER_BUILDKIT=1
   export DOCKER_SCAN_SUGGEST=false
   export CORE=dependabot/dependabot-core
-  export CI=dependabot/dependabot-core-ci
+  export CI=dependabot/dependabot-updater
 
   set -x
   docker build -qt $CORE .
```
jakecoffman, line 41: The next line down from here references Dockerfile.ci which was removed in #6418. I think if we change it to Dockerfile.updater that should fix things?

deivid-rodriguez: This script is not actually used in CI, but I changed it because I noticed the out of date reference. I will fix the one you mention too!

I think the problem may be related to this old PR I created a while ago (from which I lost context now but feels like it could be related, looking at the failures we get now:

> cd: /home/dependabot/dependabot/python: No such file or directory

jakecoffman: Ah yeah the path is wrong, there's two dependabots in /home/dependabot/dependabot/python, should only be one!

deivid-rodriguez:
I rebased this on top of #5625 just to see if that fixes anything.

deivid-rodriguez:
Oh, I thought that invalid path was generated by the script and that the issue was more complicated. I saw now that there was just a typo in the path with that duplicated component? (@jakecoffman I think I ate your previous review comment through a force-push, sorry 😳).

jakecoffman:
No problem, trying a fix over here: https://github.com/dependabot/dependabot-core/pull/6421

deivid-rodriguez:
I think we are using the same fix :)

deivid-rodriguez:
It seems to be building fine now 👍.

deivid-rodriguez:
I removed the commit I had cherry-picked for forcing a rebuild for Python. So should be ready now!

jakecoffman:
I bumped common to get all the CI to run in https://github.com/dependabot/dependabot-core/pull/6421 and now seeing what looks like linter errors, so maybe we revert https://github.com/dependabot/dependabot-core/pull/6421 instead so I can get this fixed?

deivid-rodriguez:
Yeah, that makes sense, better to not introduce more mixups if we're going to need to revert everything 😅. Those linter errors again remind me of how I got into #5625, so I will have a closer look at this!

deivid-rodriguez:
All sorted out now. There's some cleanup in this PR that we can still introduce. I'll rebase it next week.

deivid-rodriguez:
Actually I prefer to do ☝️ as a separate PR. Closing.

PR #6421, opened 2023-01-12
fix CI

jakecoffman:
Broke things in #6418 but didn't notice because the filters skipped all CI tests. So I've poked the updater image to force CI on all ecosystems and common.


jakecoffman:
Due to the Bundler and common failures I will revert in https://github.com/dependabot/dependabot-core/pull/6422 and then reopen with a proper fix!

PR #6422, opened 2023-01-12
Revert "remove Dockerfile.ci as it is redundant"

jakecoffman:
Reverts dependabot/dependabot-core#6418

The CI only passed because it skipped everything and now I've got it running them there are some linter failures that might be due to rubocop config being in the wrong place perhaps.

Either way this will get our CI working again.


deivid-rodriguez (approved):
👍 Thanks @jakecoffman!

PR #6423, opened 2023-01-12
Rebuild all if the workflow itself changes

deivid-rodriguez:
Editing `.github/workflows/ci.yml` itself could affect specific ecosystems, or all of them. So for correctness we should rebuild everything like we do when `Dockerfile` or `common/` change.


jakecoffman (approved):
Good idea!

PR #6424, opened 2023-01-12
remove Dockerfile.ci since it is redundant

jakecoffman:
Second attempt at #6418 now with CI running!


jakecoffman:
I can reproduce the lint failures locally. The filesystem looks correct, .rubocop.yml looks right, really weird! I'll keep digging at this.

deivid-rodriguez:
Have a look at #5609 @jakecoffman!

deivid-rodriguez:
I think our approach is good to fix lint issues, but according to CI we still may need to also restore this bit from the `Dockerfile.ci`?

```Dockerfile
RUN git config --global user.name dependabot-ci \
  && git config --global user.email no-reply@github.com
```

I guess we should move it to the specific specs that need it.

jakecoffman:
I'm going to briefly look into why it needs the name and email set in the tests but not the updates.

jakecoffman:
Ah we have `git commit` in the test setup itself: https://github.com/dependabot/dependabot-core/blob/73fa7e8bb6955536d821d2206ed688d2dbaaa04e/common/spec/spec_helper.rb#L122

There's also at least one place in production code where we use `git commit`: https://github.com/dependabot/dependabot-core/blob/e3ce198a2fb571df97016353db72650c27092f20/go_modules/lib/dependabot/go_modules/file_updater.rb#L83-L89

It sets name and email right before it, so I don't think it's unreasonable to place it in the tests where needed too.

deivid-rodriguez:
Looks great, all that'd be still needed would be to delete the file to give some meaning to the PR title 😆.

May I merge #5609 and then you can rebase this and wrap this up?

jakecoffman:
@deivid-rodriguez yep, thanks for the help getting this sorted!

deivid-rodriguez:
No, thanks to you, really excited about the work you and @pavera are doing!

pavera (approved):

PR #6430, opened 2023-01-13
Run RuboCop as a monorepo

deivid-rodriguez:
Currently, if you introduce a style issue in one ecosystem, you'll get the same failure for all jobs of that ecosystem. I think it's better to get the failure in the lint job instead.

This PR also removes omnibus jobs, since `omnibus` has no tests and changes very rarely, so I think (even with Docker caching), they're just overhead. Instead, those few times when it does changes, run everything.


jakecoffman (approved):
I like how this gives more information during a failure instead of the lint hiding any test failures.

Nishnha:
Thanks! Does this mean we should only get 1 Dependabot PR for rubocop updates?

Should we close out https://github.com/dependabot/dependabot-core/pull/6445 and see what happens?

deivid-rodriguez:
This only affects how we run Dependabot, so it should not affect. However, I _think_ #5974 should do the trick for that. Let me deploy that one and see what happens!

PR #6431, opened 2023-01-13
fix git config not present for certain tests

jakecoffman:
In #6424 I moved the git config into the tests, but didn't realize there are some tests that run `git commit` directly like this one:

https://github.com/dependabot/dependabot-core/blob/c2b0cf8cd7d58f01565128ac68c72c4818d7edf3/common/spec/dependabot/file_fetchers/base_spec.rb#L1338-L1343

This is resulting in racy failures in CI where the git config hasn't been set globally yet.

By moving this out of the function it will run as soon as it's required which should fix that problem. It's a bit of a naive solution though so I am not opposed to other ideas.

### DIFF
--- a/common/spec/spec_helper.rb
+++ b/common/spec/spec_helper.rb
@@ -34,6 +34,9 @@ if ENV["COVERAGE"]
   end
 end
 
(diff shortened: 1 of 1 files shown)


jeffwidman (approved):
Makes sense to me.

PR #6434, opened 2023-01-13
Mark test `pending` for `composer` `Dependabot::PrivateSourceTimedOut`

jeffwidman:
This was commented out because it had to wait til `composer` timed out, which defaults to 300 seconds (5 mins).

However, `composer` supports setting a custom timeout using the env var `COMPOSER_PROCESS_TIMEOUT`.

So set a much lower timeout and re-enable the test. When I first re-enabled the test, I discovered a code bug. Furthermore, once that was fixed by https://github.com/dependabot/dependabot-core/pull/6435, I discovered a bug in the test itself, so we should keep this enabled if we have any hope of deriving value from it.

### DIFF
--- a/composer/spec/dependabot/composer/update_checker/version_resolver_spec.rb
+++ b/composer/spec/dependabot/composer/update_checker/version_resolver_spec.rb
@@ -274,21 +274,23 @@ RSpec.describe Dependabot::Composer::UpdateChecker::VersionResolver do
       end
     end
 
-    # This test is extremely slow, as it needs to wait for Composer to time out.
-    # As a result we currently keep it commented out.
-    # context "with an unreachable private registry" do
-    #   let(:project_name) { "unreachable_private_registry" }
-    #   let(:dependency_name) { "dependabot/dummy-pkg-a" }
-    #   let(:dependency_version) { nil }
-    #   let(:string_req) { "*" }
-    #   let(:latest_allowable_version) { Gem::Version.new("6.0.0") }
+    context "with an unreachable private registry" do
+      let(:project_name) { "unreachable_private_registry" }
+      let(:dependency_name) { "dependabot/dummy-pkg-a" }
+      let(:dependency_version) { nil }
+      let(:string_req) { "*" }
+      let(:latest_allowable_version) { Gem::Version.new("6.0.0") }
 
-    #   it "raises a Dependabot::PrivateSourceTimedOut error" do
-    #     expect { resolver.latest_resolvable_version }.
-    #       to raise_error(Dependabot::PrivateSourceTimedOut) do |error|
-    #         expect(error.source).to eq("https://composer.dependabot.com")
-    #       end
-    #   end
-    # end
+      before { ENV["COMPOSER_PROCESS_TIMEOUT"] = "1" }
+      after { ENV.delete("COMPOSER_PROCESS_TIMEOUT") }
+
+      it "raises a Dependabot::PrivateSourceTimedOut error" do
+        pending("TODO: this URL has no DNS record post GitHub acquisition, so switch to a routable URL that hangs")
+        expect { resolver.latest_resolvable_version }.
+          to raise_error(Dependabot::PrivateSourceTimedOut) do |error|
+            expect(error.source).to eq("https://composer.dependabot.com")
+          end
+      end
+    end
   end
 end


composer/spec/dependabot/composer/update_checker/version_resolver_spec.rb
```diff
@@ -274,21 +274,22 @@
...
+      let(:string_req) { "*" }
+      let(:latest_allowable_version) { Gem::Version.new("6.0.0") }
+
+      before { ENV["COMPOSER_PROCESS_TIMEOUT"] = "1" }
+      after { ENV.delete("COMPOSER_PROCESS_TIMEOUT") }
+
+      it "raises a Dependabot::PrivateSourceTimedOut error" do
+        expect { resolver.latest_resolvable_version }.
+          to raise_error(Dependabot::PrivateSourceTimedOut) do |error|
+            expect(error.source).to eq("https://composer.dependabot.com")
```
jeffwidman, line 290: Rather than hanging and eventually returning a timeout error, this returns:
```ruby
expected Dependabot::PrivateSourceTimedOut, got #<Dependabot::SharedHelpers::HelperSubprocessFailed: curl error 6 while downloading https://composer.dependabot.com/packages.json: Could not resolve host: composer.dependabot.com>
```

A few related changes since this test was originally written:
* `https://composer.dependabot.com` no longer has a DNS record now that Dependabot has been acquired by GitHub
* This was originally written against `composer` `v1`, and we're now using `v2` so may return the resolving failure differently


But in order to test this, we need to mock a routable endpoint that simply hangs... while I'm sure an easy way to do this exists in Ruby, I'm not sure how?

jeffwidman:
Blocked by https://github.com/dependabot/dependabot-core/pull/6435

composer/spec/dependabot/composer/update_checker/version_resolver_spec.rb
jeffwidman: There's some good suggestions here: https://stackoverflow.com/questions/100841/artificially-create-a-connection-timeout-error

jeffwidman: I'm going to handle ☝️ in a follow-up PR, as I expect once I fix the test we'll have to also update the code... so for now let's get the test marked pending so it's at least on our radar as something that needs fixing and not hiding commented out.

jeffwidman: https://github.com/dependabot/dependabot-core/pull/6776

Nishnha (approved):
👍🏾 thanks for the easy-to-follow comments

jeffwidman:
Marking as draft til I resolve https://github.com/dependabot/dependabot-core/pull/6775#issuecomment-1452891709

composer/spec/dependabot/composer/update_checker/version_resolver_spec.rb
```diff
@@ -274,21 +274,23 @@
...
+      let(:latest_allowable_version) { Gem::Version.new("6.0.0") }
+
+      before { ENV["COMPOSER_PROCESS_TIMEOUT"] = "1" }
+      after { ENV.delete("COMPOSER_PROCESS_TIMEOUT") }
+
+      it "raises a Dependabot::PrivateSourceTimedOut error" do
+        pending("TODO: this URL has no DNS record post GitHub acquisition, so switch to a routable URL that hangs")
+        expect { resolver.latest_resolvable_version }.
+          to raise_error(Dependabot::PrivateSourceTimedOut) do |error|
+            expect(error.source).to eq("https://composer.dependabot.com")
```
Nishnha, line 291: ah actually, these 2 `expect` clauses will probably also need to be changed to `pending`

jeffwidman: CI passes w/o need to, because as long as one expect fails, then the entire test is considered to fail I think??

composer/spec/dependabot/composer/update_checker/version_resolver_spec.rb
```diff
@@ -274,21 +274,23 @@
...
+      let(:dependency_name) { "dependabot/dummy-pkg-a" }
+      let(:dependency_version) { nil }
+      let(:string_req) { "*" }
+      let(:latest_allowable_version) { Gem::Version.new("6.0.0") }
+
+      before { ENV["COMPOSER_PROCESS_TIMEOUT"] = "1" }
+      after { ENV.delete("COMPOSER_PROCESS_TIMEOUT") }
+
+      it "raises a Dependabot::PrivateSourceTimedOut error" do
+        pending("TODO: this URL has no DNS record post GitHub acquisition, so switch to a routable URL that hangs")
```
jeffwidman, line 288: `pending()` means the test will still run... but is expected to fail.

It's okay to not mark this one as `skip()` (unlike #6775 ) because this already has a very short custom timeout of 1 second, so it's okay that it runs in CI.

PR #6435, opened 2023-01-13
Remove duplicate sanitization from `composer`

jeffwidman:
While working on https://github.com/dependabot/dependabot-core/pull/6434, I discovered that my prior change https://github.com/dependabot/dependabot-core/pull/6291/files#diff-82bee76077e5dbd4f3073948aa2ad29eec23333f9b4e06550fbf54c259875a24R343 was blowing up because `error.message` cannot be directly modified.

When I started to look how to change that, I realized that we _already_ sanitize all URL credentials from any message at the global level:
* https://github.com/dependabot/dependabot-core/pull/2828

So there's no need to sanitize here.

I compared the two regexes:
https://github.com/dependabot/dependabot-core/blob/9a941db3517eb407f00c5832c902a57a8762f1cb/common/lib/dependabot/errors.rb#L7 https://github.com/dependabot/dependabot-core/blob/9a941db3517eb407f00c5832c902a57a8762f1cb/composer/lib/dependabot/composer/update_checker/version_resolver.rb#L526

and while they differ, they both strip username/password from the `sensitive_error` test case:
https://github.com/dependabot/dependabot-core/blob/9a941db3517eb407f00c5832c902a57a8762f1cb/common/spec/helpers/test/run.rb#L11

I also grep'ed around, and we don't do creds sanitization for URLs in any other ecosystem... so it's best to remove this. If we do hear reports of unredacted creds getting into logs, then we certainly would rather fix at the global level regex, and not solely here in composer.

Also, from a GitHub security perspective, these creds never even make it to `dependabot-core` since we use a separate proxy for handling credentials... so this only affects users running `dependabot-core` on their own.

### DIFF
--- a/composer/lib/dependabot/composer/update_checker/version_resolver.rb
+++ b/composer/lib/dependabot/composer/update_checker/version_resolver.rb
@@ -240,8 +240,6 @@ module Dependabot
         # rubocop:disable Metrics/CyclomaticComplexity
         # rubocop:disable Metrics/MethodLength
         def handle_composer_errors(error)
-          sanitized_message = remove_url_credentials(error.message)
-
           # Special case for Laravel Nova, which will fall back to attempting
           # to close a private repo if given invalid (or no) credentials
           if error.message.include?("github.com/laravel/nova.git")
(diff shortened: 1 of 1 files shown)


Nishnha (approved):
> I realized that we already sanitize all URL credentials from any message at the global level

TIL!

If you did want to sanitize the error message for some reason, you should be able to do `error.message.dup`

PR #6436, opened 2023-01-14
Bump `ruby` to `3.2.0`

blue2cat:
Bumps `ruby` to `3.2.0` ~~and bundler to 2.4.3~~

Ruby `3.2.0` introduces two features that improve Regex expression parsing. 

From the `3.2.0` [release notes](https://www.ruby-lang.org/en/news/2022/12/25/ruby-3-2-0-released/): 
```
# This match takes 10 sec. in Ruby 3.1, and 0.003 sec. in Ruby 3.2
/^a*b?a*$/ =~ "a" * 50000 + "x"
```
> The improved matching algorithm allows most Regexp matching (about 90% in our experiments) to be completed in linear time.


blue2cat:
**Note**: Ruby `3.2.0` removed [taint checking methods](https://bugs.ruby-lang.org/issues/16131) from the codebase, which appears to be causing the installation of Bundler `v1` to fail when building the `Dockerfile` at step 25/39. 

image

deivid-rodriguez:
Bundler will be updated in #6303.

Regarding Ruby, those regexp improvements do look appealing for us :) One issue is indeed what to do with Bundler 1. I see several options:

* Drop support for Bundler 1 before upgrading to Ruby 3.2. That effectively blocks the Ruby upgrade until we decide we're ready to drop Bundler 1.

* Monkey patch Bundler 1 to remove all taint/untaint. This seems doable but messy, specially since I think this is used in quite a few places.

* Let Bundler 2 also update lockfiles created with Bundler 1. This sounds a bit like cheating but I think it should be fine because the lockfile format has not really changed between Bundler 1 and Bundler 2.

I think the latter may be the best course of action because it allows us to effectively get rid of Bundler 1 internally, while still being able to provide updates to lockfiles locked to Bundler 1.

jeffwidman:
Thanks for opening this PR.

I think the Bundler v1 discussion could easily become larger than just this PR, so I opened:
* #6457

So let's continue the discussion there about both short and long-term plans for Bundler v1?

deivid-rodriguez:
Sure @jeffwidman, makes total sense to split that out, you beat me to that :)

Dockerfile
```diff
@@ -70,7 +70,7 @@ RUN if ! getent group "$USER_GID"; then groupadd --gid "$USER_GID" dependabot ;
 ### RUBY
 
 # When bumping Ruby minor, need to also add the previous version to `bundler/helpers/v{1,2}/monkey_patches/definition_ruby_version_patch.rb`
-ARG RUBY_VERSION=3.1.3
+ARG RUBY_VERSION=3.2.0
```
jeffwidman, line 73: We restructured the dockerfiles, so this now needs to be bumped over here instead:
https://github.com/dependabot/dependabot-core/blob/3313530fa05c482dc0dace53af7dfffee185e2fd/Dockerfile.updater-core#L40

Can you rebase to pickup that change?

blue2cat: @jeffwidman done.

jeffwidman:
TBH, the metrics I've seen internally for `bundler 1` usage are _so_ high that they don't smell right... Like they're insanely high compared to what I was expected. Especially because as repos move to `ruby` `3.2`, they're going to hit this issue themselves and have to bump to `bundler 2` anyway. 

So that number I'd expect to start moving more than it has, which makes me suspect something might be amiss with how we've instrumented that data. I know both myself and Deivid have looked at the code and it looks correct, but as a triple-check I've been meaning to dig a bit deeper and find some open source repos that are emitting `bundler 1` to the metrics and manually inspect them to see if they're truly using `bundler 1`, or if our metrics code is misattributing something.

Anyway, this is probably a conversation better had internally but wanted to mention it on this thread so we don't forget and prematurely invest a bunch of time into supporting `bundler 1` / other Rubies until we're truly really confident in the data we're seeing.

jeffwidman:
Another PR managed to workaround updating Ruby w/o updating bundler:
* https://github.com/dependabot/dependabot-core/pull/9597

PR #6451, opened 2023-01-17
ecosystem-images: speed up base build

jakecoffman:
- Remove apt dependencies that are not needed for the ruby-install build
- Pass the `-j` flag to ruby-install which will build with multiple cores (I set to 4 but we might need to use 2 to match the number of cores in the Actions runner)

Also a couple of minor clean-ups.

### DIFF
--- a/Dockerfile.updater-core
+++ b/Dockerfile.updater-core
@@ -14,17 +14,9 @@ RUN apt-get update \
   && apt-get upgrade -y \
   && apt-get install -y --no-install-recommends \
     build-essential \
-    dirmngr \
-    git \
-    git-lfs \
-    bzr \
-    mercurial \
-    gnupg2 \
     ca-certificates \
     curl \
-    file \
     zlib1g-dev \
-    liblzma-dev \
     libyaml-dev \
     libgdbm-dev \
     bison \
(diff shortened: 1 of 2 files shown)


pavera (approved):

jakecoffman:
The uncached build is 2-3 minutes faster in Actions now!

Dockerfile.updater-core
```diff
@@ -181,10 +153,9 @@ ENV DEPENDABOT_HOME /home/dependabot
 COPY --chown=dependabot:dependabot LICENSE $DEPENDABOT_HOME
 COPY --chown=dependabot:dependabot updater $DEPENDABOT_HOME/dependabot-updater
 
-ENV PATH="$HOME/bin:/opt/go/bin:/opt/go_modules/bin:$PATH"
+ENV PATH="$HOME/bin:$PATH"
 ENV DEPENDABOT_NATIVE_HELPERS_PATH="/opt"
 
-USER dependabot
```
Nishnha, line 187: Won't removing this line run CI as root?

jakecoffman: Oh CI doesn't run on this container at the moment? We might want to run common on this one, so perhaps we can do that as `dependabot`.

PR #6454, opened 2023-01-17
Catch up with some changes in upstream pub

deivid-rodriguez:
It's possible that we'll need to pick up a fix in upstream pub at dart-lang/pub#3736. This PR gets ready for that by catching up with some recent changes in upstream pub.

### DIFF
--- a/pub/helpers/bin/dependency_services.dart
+++ b/pub/helpers/bin/dependency_services.dart
@@ -14,7 +14,7 @@ import 'package:pub/src/log.dart' as log;
 class _DependencyServicesCommandRunner extends CommandRunner<int>
     implements PubTopLevel {
   @override
-  String? get directory => argResults['directory'];
+  String get directory => argResults['directory'];
 
   @override
   bool get captureStackChains => argResults['verbose'];
--- a/pub/helpers/bin/dependency_services.dart
+++ b/pub/helpers/bin/dependency_services.dart
@@ -40,6 +40,7 @@ class _DependencyServicesCommandRunner extends CommandRunner<int>
             usageLineLength: lineLength) {
     argParser.addFlag('verbose',
         abbr: 'v', negatable: false, help: 'Shortcut for "--verbosity=all".');
+    PubTopLevel.addColorFlag(argParser);
     argParser.addOption(
       'directory',
       abbr: 'C',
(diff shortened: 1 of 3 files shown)


jeffwidman:
This would benefit from a review from @jonasfj or @sigurdm

pub/helpers/bin/dependency_services.dart
```diff
@@ -14,7 +14,7 @@ import 'package:pub/src/log.dart' as log;
 class _DependencyServicesCommandRunner extends CommandRunner<int>
     implements PubTopLevel {
   @override
-  String? get directory => argResults['directory'];
```
jeffwidman, line 17: why is this needed?

jeffwidman (approved):

pub/helpers/bin/dependency_services.dart
sigurdm: The CommandRunner class was updated: https://github.com/dart-lang/pub/pull/3587

sigurdm:
One thing this could use is a test of our hashes in pubspec.lock.

They should get added if they are already there, and left out if not present.

This was coded in the dependency_services here:
https://github.com/dart-lang/pub/pull/3482/files#diff-73780674d4f6139a45323ccb874a95106765df0d47fc0b217aad4b22011f0bc3

deivid-rodriguez:
Nice, thanks for pointing that out.

> They should get added if they are already there, and left out if not present.

This is indeed the ideal behavior for us ❤️.

I'll try to add some specs for that!

sigurdm:
In the light of https://github.com/flutter/flutter/issues/119091 we should do a double check of the behavior around the hosted url.

I'm afraid (but haven't tested) that with this PR dependabot would update the hosted url from 'pub.dartlang.org' to 'pub.dev'

Ideally this transition should be done by pub itself IMO. We probably need special handling in the dependency_services.

sigurdm:
> I'm afraid (but haven't tested) that with this PR dependabot would update the hosted url from 'pub.dartlang.org' to 'pub.dev'

Sorry, I never got back to this.

https://github.com/dart-lang/pub/pull/3846 should fix this issue, then we can forward all the way to that.

sigurdm:
The PR has landed now, we should be able to forward the pub ref to: https://github.com/dart-lang/pub/commit/29a7c0990cf070aa118570554f89402c37b54cf6

palisadoes:
Hello everyone, when could this be merged? We have to manually do all dependabot suggested updates to our packages and fixing the underlying issue would help greatly.

deivid-rodriguez:
Closing in favor of #7265.

PR #6459, opened 2023-01-18
Log full request path and responses in dry-run.rb

mburumaxwell:
When exciting `dry-run.rb` script, it would be ideal to log the full URL (built using `Excon::Utils.request_uri(payload)`) for requests that share the path but different query parameters, especially in Azure DevOps.

Logging the response is also useful when debugging.

This is a result of https://github.com/excon/excon/pull/803 hence the bump in excon version

Sample log with this change:

```txt
🌍 --> GET https://dev.azure.com/tingle/dependabot/_apis/git/repositories/repro-366
🌍 <-- 200 https://dev.azure.com:443/tingle/dependabot/_apis/git/repositories/repro-366
🌍 --> GET https://dev.azure.com/tingle/dependabot/_apis/git/repositories/repro-366/stats/branches?name=main
🌍 <-- 200 https://dev.azure.com:443/tingle/dependabot/_apis/git/repositories/repro-366/stats/branches?name=main
🌍 --> GET https://dev.azure.com/tingle/dependabot/_apis/git/repositories/repro-366/items?path=.github/dependabot.yml&versionDescriptor.versionType=commit&versionDescriptor.version=484e0dccc2d7b6a3bbb886a01c1da23544de8c36
🌍 <-- 200 https://dev.azure.com:443/tingle/dependabot/_apis/git/repositories/repro-366/items?path=.github/dependabot.yml&versionDescriptor.versionType=commit&versionDescriptor.version=484e0dccc2d7b6a3bbb886a01c1da23544de8c36
=> fetching dependency files
🌍 --> GET https://dev.azure.com/tingle/dependabot/_apis/git/repositories/repro-366
```

When a requests return 404, we can check if the path is correct or where an issue could be. In scenarios where the body is necessary, `EXCON_DEBUG=1` can be used.


.gitignore
```diff
@@ -30,3 +30,4 @@ coverage/
 .tool-versions
 .rspec_status
 .rdbg_history
+.idea
```
deivid-rodriguez, line 33: Is this intentional?

deivid-rodriguez (commented):
This looks like a nice improvement, thanks!

I think we may want to hold this PR for a bit until we're sure we're ready to jump to the most recent excon version since we had to pin ourselves to 0.93.1 recently. See https://github.com/dependabot/dependabot-core/commit/6d59a5ae80465db698e4cdc29270957ecd7bfc2c.

.gitignore
mburumaxwell: It would seem the https://github.com/excon/excon/issues/800 is resolved in in version 0.95 onwards. When the version is fixed in https://github.com/dependabot/dependabot-core/commit/6d59a5ae80465db698e4cdc29270957ecd7bfc2c by preventing 0.94 onwards, I see the range has been expanded to 0.98 recently. See https://github.com/dependabot/dependabot-core/pull/6403. I run a whole update for a repository without any errors since I skipped that version 0.94.0 version with issues.

Will remove this the ignore.

deivid-rodriguez: Yep, the issue seems already fixed, we just need to validate that upgrading causes no issues for us. We need to validate this on our internal systems, I will try to move that forward.

mburumaxwell: Thanks. Looking forward to that

geemus: Just let me know if you see any issues related to excon/excon#800 still lingering, it seemed like we squished the related issues, but they were a bit hard to pin down in the first place...

deivid-rodriguez: We will, thanks for your prompt effort on fixing that ❤️

jeffwidman (approved):
I think all you really need here is the `dry-run.rb` change? The `gemspec` / `Gemfile.lock` aren't needed for this PR IIUC??

mburumaxwell:
> I think all you really need here is the `dry-run.rb` change? The `gemspec` / `Gemfile.lock` aren't needed for this PR IIUC??

How about the changes in https://github.com/excon/excon/pull/803 without which the response doesn't have crucial aspects of the URI?

jeffwidman:
Ah, that's why you're bumping the minimum version... makes sense. I was only thinking about how it will default to the max version.

jeffwidman:
Can you fix merge conflicts so we can merge?

mburumaxwell:
> Can you fix merge conflicts so we can merge?

Done

jeffwidman:
Thanks again!

PR #6464, opened 2023-01-18
publish updater-core image

jakecoffman:
This publishes the latest tag for our new base image for all the ecosystem-specific images. Publishing this helps with caching.


deivid-rodriguez (approved):

deivid-rodriguez:
It does not seem to have worked https://github.com/dependabot/dependabot-core/actions/runs/3951709904/jobs/6765903776. Should we revert and work on this on a branch until it works?

jakecoffman:
Yep, I will do that and also make a separate workflow so it can be triggered and not deploy everything!

PR #6470, opened 2023-01-18
workflow to deploy updater-core:latest

jakecoffman:
This deploys updater-core:latest mostly for caching purposes. It's the base for all the ecosystem images.


pavera (approved):
👍

PR #6476, opened 2023-01-18
add a Maven ecosystem image

jakecoffman:
This creates the Maven-specific Dependabot image.

There are a few others that are somewhat trivial to implement, doing this one first by itself to set the pattern.


Dockerfile.maven
```diff
@@ -0,0 +1,4 @@
+FROM ghcr.io/dependabot/dependabot-updater-core
+
+USER dependabot
+COPY --chown=dependabot:dependabot maven /home/dependabot/maven
```
jeffwidman, line 4: Since there's nothing customized in here, could this be parameterized as a basic dockerfile that gets used in any ecosystem that doesn't need customization?

jeffwidman (approved):
Given that there's nothing custom in that dockerfile right now...

Dockerfile.maven
jakecoffman: Possibly, but I would like to not refactor as we break things out to avoid extra risk.

PR #5820, opened 2022-09-30
Run shellcheck against `./script/*`

jeffwidman:
A few more places we can run shellcheck against...

PR #6478, opened 2023-01-19
No need to check if env var set

jeffwidman:
Per the Docker [docs](https://docs.docker.com/engine/reference/commandline/run/#-set-environment-variables--e---env---env-file):
> When running the command, the Docker CLI client checks the value the variable has in your local environment and passes it to the container. If no `=` is provided and that variable is not exported in your local environment, the variable won’t be set in the container.

So there's no need to check if an env var is set in the environment before passing it to the `docker`  command.

Conveniently, this removes some of the code throwing `shellcheck` warnings over in:
* https://github.com/dependabot/dependabot-core/pull/5820

### DIFF
--- a/bin/ci-test
+++ b/bin/ci-test
@@ -1,5 +1,5 @@
 #!/bin/bash -e
-#/ usage: ci-test SUITE
+#/ usage: ci-test SUITE_NAME
 #/ simulate the process followed by ~/.github/workflows/ci.yml
 
 function print_usage() {
@@ -7,8 +7,8 @@ function print_usage() {
 }
 
 function handle_args() {
-  export SUITE=$1
-  case $SUITE in
+  export SUITE_NAME=$1
+  case $SUITE_NAME in
     bundler1 | bundler2)
       export MODULE=bundler
       ;;
@@ -17,7 +17,7 @@ function handle_args() {
       exit 1
       ;;
     *)
-      export MODULE=$SUITE
+      export MODULE=$SUITE_NAME
   esac
 
   if ! [ -d "$MODULE" ]; then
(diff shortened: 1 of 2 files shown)


script/_common
```diff
@@ -25,13 +25,8 @@ function docker_exec() {
 function docker_bundle_exec() {
   docker_build
 
-  VCR_ARGS=""
-  if [ -n "$VCR" ]; then
-    VCR_ARGS="--env \"VCR=$VCR\""
-  fi
```
jeffwidman, line 31: This code is simply trying to set `--env VCR=$VCR` only when `$VCR` is set... and can be replaced with simply `--env VCR`...

bin/ci-test
```diff
@@ -37,7 +37,9 @@ function build() {
   export CORE=dependabot/dependabot-core
   export UPDATER=dependabot/dependabot-updater
 
+  # shellcheck disable=SC2034  # shellcheck isn't smart enough to realize `docker build` uses `TARGETARCH`
   local TARGETARCH;
+  # shellcheck disable=SC2034  # shellcheck isn't smart enough to realize `docker build` uses `TARGETARCH`
```
jeffwidman, line 42: I had to repeat this line because I couldn't find a way to disable/re-enable shellcheck for a block of code... it was only the line or the file AFAICT.

deivid-rodriguez: Technically `TARGETARCH` should be local to this function, but I don't think using a standard global variable (by removing the `local` line) should cause any issues, and would save us from repeating this line which is a bit confusing.

Alternatively, we could add a blank line above this line just to emphasize that each directive applies to the line below it and this one is not just a mistyped `# shellcheck enable=SC2034` (which I just checked it's not valid, but it could!).

jeffwidman: Yeah, I'm not too worried once it's in the code about being mistaking for an accident... if someone tries to "clean it up", then CI will squawk at them once #5820 lands.

deivid-rodriguez: > Yeah, I'm not too worried once it's in the code about being mistaking for an accident... if someone tries to "clean it up", then CI will squawk at them once https://github.com/dependabot/dependabot-core/pull/5820 lands.

Even if our CI would catch it, if we can spare people the time to create that wrong change, we should try!

deivid-rodriguez: I know I left this comment after approving, because I noticed the duplicated line late, but if we can add that blank line or remove the `local TARGETARCH` line I'd go with that. But no blocker at all.

jeffwidman: Sorry, just seeing this. I misread your comment earlier and didn't realize I could potentially remove the `local TARGETARCH` line. Let me tag this into #5820

PR #6461, opened 2023-01-18
Fix flaky spec failure in updater

deivid-rodriguez:
Authenticate updater spec to by-pass rate limiting, otherwise it may fail randomly. For example: https://github.com/dependabot/dependabot-core/actions/runs/3949825265/jobs/6761544223.

```
Failures:

  1) bin/run fetch_files completes the job successfully and persists the files
     Failure/Error: expect(result).to include("Finished job processing")

       expected "I, [2023-01-18T14:31:25.371276 #292]  INFO -- sentry: ** [Raven] Raven 3.1.2 configured not to captu...s:\nDependabot encountered '2' error(s) during execution, please check the logs for more details.\n" to include "Finished job processing"
       Diff:
       @@ -1,26 +1,51 @@
       -Finished job processing
       +I, [2023-01-18T14:31:25.371276 #292]  INFO -- sentry: ** [Raven] Raven 3.1.2 configured not to capture errors: DSN not set
       +INFO <job_1> Starting job processing
       +ERROR <job_1> Repository is rate limited, attempting to retry in 734.285586435s
       +ERROR <job_1> <?xml version="1.0" encoding="iso-8859-1"?>
       +<job_1> <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
       +<job_1>          "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
       +<job_1> <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
       +<job_1> 	<head>
       +<job_1> 		<title>404 - Not Found</title>
       +<job_1> 	</head>
       +<job_1> 	<body>
       +<job_1> 		<h1>404 - Not Found</h1>
       +<job_1> 		<script type="text/javascript" src="//wpc.75674.betacdn.net/0075674/www/ec_tpm_bcon.js"></script>
       +<job_1> 	</body>
       +<job_1> </html>
       +
       +ERROR <job_1> /home/dependabot/dependabot-updater/lib/dependabot/api_client.rb:98:in `record_update_job_error'
       +ERROR <job_1> /home/dependabot/dependabot-updater/lib/dependabot/service.rb:38:in `record_update_job_error'
       +ERROR <job_1> /home/dependabot/dependabot-updater/lib/dependabot/file_fetcher_job.rb:179:in `record_error'
[...]

PR #6480, opened 2023-01-19
Fix flaky spec failure in updater (take 2)

deivid-rodriguez:
In https://github.com/dependabot/dependabot-core/pull/6461, I attempted to inject environment credentials in the fixture file that `bin_run_spec.rb` uses so that the spec's requests don't get rate limited.

However, turns out credentials were not being properly picked up by the fetcher.

I think the reason is that "credentials" were located at the wrong position in the fixture file, so were getting ignored. I also corrected other fixtures with the same problem.

### DIFF
--- a/updater/spec/bin_run_spec.rb
+++ b/updater/spec/bin_run_spec.rb
@@ -11,7 +11,7 @@ RSpec.describe "bin/run" do
         job_path = File.join(tempdir, "job.json")
 
         job_info = JSON.parse(File.read("spec/fixtures/jobs/job_with_credentials.json"))
-        job_info["job"]["credentials"][0]["password"] = test_access_token
+        job_info["credentials"][0]["password"] = test_access_token
 
         File.write(job_path, JSON.dump(job_info))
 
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
The hints of why I think the right place for credentials is a top level "credentials" key, and not one scoped to "job" is the place where credentials are passed:

https://github.com/dependabot/dependabot-core/blob/fe24ab9e9914ed2e4ed415b2d1cd4828fee07791/updater/lib/dependabot/file_fetcher_job.rb#L96

https://github.com/dependabot/dependabot-core/blob/fe24ab9e9914ed2e4ed415b2d1cd4828fee07791/updater/lib/dependabot/file_fetcher_job.rb#L175

And the examples used in the CLI: https://github.com/dependabot/cli/blob/main/README.md#job-description-file.

landongrindheim (approved):
Makes sense to me! 🚀

deivid-rodriguez:
Thanks for your reviews @landongrindheim!

PR #6485, opened 2023-01-19
adding more ecosystem images

jakecoffman:
This adds all of the ecosystem images that don't require additional dependencies.


jakecoffman:
Oops, cargo does have dependencies, I was a bit distracted while copy pasting this together. I'll go with it!

Dockerfile.updater-core
```diff
@@ -105,6 +105,8 @@ bundle install
 
 FROM ubuntu:20.04
 
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
```
jeffwidman, line 108: ??

jeffwidman: Ah, NVM, just saw the commit description... 
> Also added pipefail to the base image so that piped failures will fail to build the Docker image.

this is kinda why I dislike commits (and oftentimes by extension PR's) that do multiple things, it's too easy to accidentally miss...

jakecoffman: Comes from https://github.com/dependabot/dependabot-core/blob/2b5748470c41ed3bc42caec5184fc9c852eaf73c/Dockerfile#L5

In Bash if you don't set `pipefail` then lines with pipes can pass even though they failed, which was happening to me.

pavera (approved):

PR #6486, opened 2023-01-19
Breakout npm_and_yarn, Move some dockerfiles

pavera:
This breaks out npm_and_yarn and also moves the previously broken out ecosystem's dockerfiles into their respective subdirs to avoid clutter at the top level.


out.yml
```diff
@@ -0,0 +1,897 @@
+input:
```
jakecoffman, line 1: Did you mean to commit this file?

pavera: no...

pavera:
yeah, I'm digging to try to figure it out, I'll get it soon I'm sure.

jakecoffman (approved):
Nice work!

PR #6492, opened 2023-01-20
Break out bundler.

pavera:
This breaks the bundler updater out to its own image.

### DIFF
--- a/bundler/Dockerfile
+++ b/bundler/Dockerfile
@@ -1,10 +1,10 @@
 FROM ghcr.io/dependabot/dependabot-updater-core
-
+ARG BUNDLER_V1_VERSION=1.17.3
 RUN apt-get update \
   && apt-get upgrade -y \
   && apt-get install -y --no-install-recommends \
-    curl
-
+    curl \
+  && gem install bundler -v $BUNDLER_V2_VERSION --no-document
 USER dependabot
 
 COPY --chown=dependabot:dependabot bundler/helpers /opt/bundler/helpers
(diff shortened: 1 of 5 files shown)


bundler/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+  && apt-get install -y --no-install-recommends \
+    curl
+
+USER dependabot
+
+COPY --chown=dependabot:dependabot bundler/helpers /opt/bundler/helpers
+RUN bash /opt/bundler/helpers/v1/build \
+  && bash /opt/bundler/helpers/v2/build
+
+COPY --chown=dependabot:dependabot bundler /home/dependabot/bundler
```
jakecoffman, line 14: I think we need to install bundler v1 somewhere since we didn't transfer it into the updater-core image, since it's only needed by bundler ecosystem.

```
gem install bundler -v $BUNDLER_V1_VERSION --no-document
```

jakecoffman: Hmmm the bundler v1 tests all passed 🤔 .

deivid-rodriguez: I think global Bundler 1 installation is no longer needed. The native helper builder takes care of it.

jakecoffman (approved):

bundler/Dockerfile
pavera: interesting that the ci passes without it..

deivid-rodriguez: I created #6493 to verify my claim :)

PR #6493, opened 2023-01-20
We no longer need to install Bundler 1 globally

deivid-rodriguez:
Since https://github.com/dependabot/dependabot-core/pull/5513 I think it's no longer needed.

For running updater, Bundler 1 is not used anyways, and for running native helpers, the Bundler 1 native helper installer installs Bundler 1 to an isolated location already.

Just opening a PR so that CI can prove me wrong.


jeffwidman (approved):
🎉 

Might want to get input from @pavera / @jakecoffman on when to merge this, given the refactoring they're currently doing with all the dockerfiles.

jeffwidman:
Actually we should be okay to merge this now that https://github.com/dependabot/dependabot-core/pull/6492/ is merged, as this simply retains parity...

PR #6495, opened 2023-01-20
Breakout the remaining "helper" based ecosystems

pavera:
This breaks out:

- composer
- elm
- hex
- pub
- terraform

It also pulls a couple commonly used dependencies into updater-core.


jakecoffman (approved):
Looks good! I'll go ahead and merge it so we don't have to wait for CI on Monday.

PR #6501, opened 2023-01-23
rename hex image to match YAML value

jakecoffman:
We're using the documented YAML values as the GHCR image names since they are the official names in dependabot.yml. This one was mistakingly named "hex" so renaming to "mix."


landongrindheim (approved):

landongrindheim:
Got a 500 on my first approval. Sorry for the noise (which I'm adding to with my apology 🙃)

PR #6502, opened 2023-01-23
update scripts to take an ecosystem parameter

jakecoffman:
By passing the name of the ecosystem (specifically the module/directory name in core) we can run all of our scripts that we're used to using while developing and testing core.

The updater tests are going to be a bit of a monkey-wrench in this plan as it runs integration tests against multiple ecosystems. So I'm still trying to figure out how we'll fix that. Would it make sense to remove some integration tests since we have smoke tests now? If not I think we need to tag them and run them in multiple containers built for the right ecosystem.


jakecoffman:
I'm not going to merge this until we've fully released this project as it would make it difficult to test the current updater image 😄

jeffwidman (commented):
95% of the time if I'm building the dockerfiles locally rather than pulling from the pre-built one, it's because I'm working on a native helper... 

So we probably will want to update the [readme dockerfile section](https://github.com/dependabot/dependabot-core/#running-with-docker) to say:
```
docker pull <base image>  # you can skip this if pulling a pre-built ecosystem image
docker pull <ecosystem image>  # pull the pre-built ecosystem image
bin/docker-dev-shell  # Running this w/o docker pull will build missing images locally
```

We can also do this in a later PR if out of scope for this PR...

script/test
```diff
@@ -4,4 +4,4 @@ set -e
 cd "$(dirname "$0")/.."
 source script/_common
 
-docker_bundle_exec rspec "$@"
+docker_bundle_exec bundler rspec "$@"
```
jeffwidman, line 7: I don't understand why `bundler` is needed here?

Since `docker_bundle_exec` already calls `docker bundle exec`, this ends up as `docker bundle exec bundler rspec...` so isn't the extra `bundler` one too many?

jakecoffman: Ha, that's a bit confusing isn't it? The first parameter to docker_bundle_exec is now the ecosystem image to build/use. So it's saying build the bundler ecosystem, then run rspec in it.

Dockerfile.development
```diff
@@ -1,4 +1,5 @@
-FROM dependabot/dependabot-core
+ARG FROM=dependabot/dependabot-core
```
jeffwidman, line 1: args and keys with the same name are so easily confusing... I'd prefer to call this something slightly more verbose to reduce confusion... open to ideas, or just do

```suggestion
ARG FROM_IMAGE=dependabot/dependabot-core
```

Dockerfile.development
```diff
@@ -1,4 +1,5 @@
...
+FROM $FROM
```
jeffwidman, line 2: args and keys with the same name are so easily confusing... I'd prefer to call this something slightly more verbose to reduce confusion/mental overhead... open to ideas, or just do

```suggestion
ARG FROM_IMAGE=dependabot/dependabot-core
FROM $FROM_IMAGE
```

jeffwidman (commented):
Everything else looks fine except the one confusing from from...

.github/workflows/ci.yml
```diff
@@ -213,10 +213,24 @@ jobs:
...
+        run: |
+          docker build \
+            -f Dockerfile.updater \
+            -t "dependabot-updater:latest" \
+            --build-arg OMNIBUS_VERSION=latest \
+            --build-arg BUILDKIT_INLINE_CACHE=1 \
+            --cache-from ghcr.io/dependabot/dependabot-updater \
+            .
       - name: Run updater tests
-        run: ./script/ci-test-updater
```
deivid-rodriguez, line 219: I observed that we're no longer running the `./script/ci-test-updater` in CI, but also that script has been modified to take an ecosystem parameter, and similarly with `script/build` above.

I would prefer to keep delegating to scripts in CI, because they make it much easier to try reproduce things locally, but if this is not possible, could/should we remove any of these scripts?

deivid-rodriguez (commented):
Excited about this work!

.github/workflows/ci.yml
jakecoffman: I agree, I will take another pass at this after merging and remove a lot of the copy-pasting with script calls. 

The reason I did this here is I wanted to change the script, but not break current CI. So just another copy-paste for the short term...

jeffwidman (approved):
Thanks! 

So it sounds like there's two follow-up tasks for later PRs:
1. Fixup readme (and any other docs) to reflect this new API
2. A pass to reduce copy/paste.

Really looking forward to using this, particularly for the faster debug cycles on native helpers.

bin/ci-test
```diff
@@ -47,15 +47,16 @@ function build() {
 
   set -x
   docker build \
-    --build-arg TARGETARCH \
+    --build-arg TARGETARCH="$TARGETARCH" \
```
deivid-rodriguez, line 50: This reverted part of https://github.com/dependabot/dependabot-core/commit/89df66560c3192871d3c41a34a2fa151a4fef674, was it intentional?

jakecoffman: Oh sorry, I thought I had typo'd that. I *think* I might have been getting a build error with it on my ARM system. I'm happy to try again in a moment.

deivid-rodriguez: No problem, it's a very unknown docker feature, I didn't know it until @jeffwidman introduced it. I can see how you thought you typo'ed it!

jakecoffman: Confirmed it doesn't work with just `--build-arg TARGETARCH`.

jakecoffman: Follow up PR: https://github.com/dependabot/dependabot-core/pull/6549

Issue #5720, opened 2022-09-14
Upgrade `composer` to `2.4`


jeffwidman:
`composer` `2.4` provides a couple of new options:
* `audit` is enabled by default for `composer update`, resulting in a bunch more API calls. For what we do this is unnecessary and we'll need to ensure it's disabled from the get go. This could be done using the `COMPOSER_NO_AUDIT` env var.
* `bump` may be a feature we can leverage. That can be done later though, so I broke that out separate in #5719 

Details:
* https://blog.packagist.com/composer-2-4/
* https://github.com/shivammathur/setup-php/issues/635
* https://github.com/composer/composer/releases/tag/2.4.0

See also:
* #5719 
* #5721

jeffwidman:
Fixed by https://github.com/dependabot/dependabot-core/pull/5577.

Disabling audit was handled by https://github.com/dependabot/dependabot-core/pull/6503.

PR #6503, opened 2023-01-23
Add compatibility with composer v2.4+

stefangr:
When running the dependabot-script with `DEBUG_HELPERS=true` I saw this error:
```
PHP Deprecated:  Installer::setIgnorePlatformRequirements is deprecated since Composer 2.2, use setPlatformRequirementFilter instead. in /opt/composer/v2/vendor/composer/composer/src/Composer/Installer.php on line 1289
```

In this PR I work around the deprecations (there was one more, which was not verbose) and also added a forward compatible change to [disable the audit](https://github.com/dependabot/dependabot-core/pull/6385#issuecomment-1372922640).

Applied changes:
- Use `eventDispatcher::setRunScripts` instead of [deprecated `Installer::setRunScripts`](https://github.com/composer/composer/blob/2.5.1/src/Composer/Installer.php#L1241)
- Use `Installer::setPlatformRequirementFilter` instead of [deprecated `Installer::setIgnorePlatformRequirements`](https://github.com/composer/composer/blob/2.5.1/src/Composer/Installer.php#L1297)
- Use `setAudit(false)` if Installer has that method

### DIFF
--- a/composer/helpers/v2/src/UpdateChecker.php
+++ b/composer/helpers/v2/src/UpdateChecker.php
@@ -6,6 +6,7 @@ namespace Dependabot\Composer;
 
 use Composer\DependencyResolver\Request;
 use Composer\Factory;
+use Composer\Filter\PlatformRequirementFilter\PlatformRequirementFilterFactory;
 use Composer\Installer;
 use Composer\Package\PackageInterface;
 
@@ -59,6 +60,8 @@ final class UpdateChecker
             $composer->getAutoloadGenerator()
         );
 
+        $composer->getEventDispatcher()->setRunScripts(false);
+
         // For all potential options, see UpdateCommand in composer
         $install
             ->setUpdate(true)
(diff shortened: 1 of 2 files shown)


jeffwidman (approved):
Oh perfect timing, I was planning to work on #5720 / #6385 / https://github.com/dependabot/dependabot-core/pull/6385#issuecomment-1372922640 later this week!

I suspect life is simpler if we just keep on the latest `composer` and don't worry about backwards compatibility for options... I assume most folks run `dependabot` within the context of our bundled dependabot-core image, so they'll be using whatever version of `composer` is bundled with that. If you think differently for some reason, please let me know.

For simplicity of working across timezones / PR rebases, does the following sound reasonable?
1. Merge this
2. Rebase the Dependabot PR to bump `composer` to `2.5.x`
3. Add a commit remove the optional checks, such that `setAudit` is always called.
4. Merge... so hopefully we'll be back up to `composer` `latest` by EOD 👍

composer/helpers/v2/src/UpdateChecker.php
```diff
@@ -59,15 +60,20 @@ public static function getLatestResolvableVersion(array $args): ?string
...
         // For all potential options, see UpdateCommand in composer
         $install
             ->setUpdate(true)
             ->setInstall(false)
             ->setDevMode(true)
             ->setUpdateAllowTransitiveDependencies(Request::UPDATE_LISTED_WITH_TRANSITIVE_DEPS)
             ->setDumpAutoloader(false)
-            ->setRunScripts(false)
-            ->setIgnorePlatformRequirements(false);
+            ->setPlatformRequirementFilter(PlatformRequirementFilterFactory::fromBoolOrList(false));
```
jeffwidman, line 72: 😍 for fixing this, as a non-PHP expert, I would have muddled for a while before realizing it needed another import...

stefangr:
> 



> For simplicity of working across timezones / PR rebases, does the following sound reasonable?
> 
>     1. Merge this
> 
>     2. Rebase the Dependabot PR to bump `composer` to `2.5.x`
> 
>     3. Add a commit remove the optional checks, such that `setAudit` is always called.
> 
>     4. Merge... so hopefully we'll be back up to `composer` `latest` by EOD +1

That sounds completely reasonable. :+1:

PR #6504, opened 2023-01-23
add missing CI tests for common and updater

jakecoffman:
I'm dropping a dummy Dockerfile in common to make it fit with our other tests. I might break out a separate job instead since it's kind of silly to do that. Or we could push the `Dockerfile.updater-core` down into common, but then we'd need to update all the CI to point there. 

Either way, this adds a couple CI tests we missed.


jakecoffman:
Here's the updater failures in question:

```
rspec ./spec/dependabot/integration_spec.rb:583 # Dependabot::EndToEndJob composer updates dependencies correctly
rspec ./spec/dependabot/integration_spec.rb:647 # Dependabot::EndToEndJob composer summarizes the changes
rspec ./spec/dependabot/integration_spec.rb:517 # Dependabot::EndToEndJob JavaScript summarizes the changes
rspec ./spec/dependabot/integration_spec.rb:442 # Dependabot::EndToEndJob JavaScript updates dependencies correctly
rspec ./spec/dependabot/job_spec.rb:300 # Dependabot::Job#allowed_update? with dev dependencies during a security update while allowed: production is in effect 
rspec ./spec/dependabot/job_spec.rb:412 # Dependabot::Job#clone? for ecosystems that always clone is expected to eq true
rspec ./spec/dependabot/file_fetcher_job_spec.rb:166 # Dependabot::FileFetcherJob#perform_job when package ecosystem always clones clones the repo
rspec ./spec/dependabot/file_fetcher_job_spec.rb:184 # Dependabot::FileFetcherJob#perform_job when package ecosystem always clones when the fetcher raises a BranchNotFound error while cloning tells the backend about the error (and doesn't re-raise it)
rspec ./spec/npm_and_yarn_config_spec.rb:7 # npm and yarn config contains a valid .npmrc config file
rspec ./spec/npm_and_yarn_config_spec.rb:19 # npm and yarn config contains a valid .yarnrc config file
```

I built a bundler updater to run the integration tests, so it fails on all the other ecosystem integration tests.

jakecoffman:
I moved the npm and yarn rc file checks out to the npm_and_yarn suite, and I mounted some directories, so we're just left with:

```
rspec ./spec/dependabot/integration_spec.rb:647 # Dependabot::EndToEndJob composer summarizes the changes
rspec ./spec/dependabot/integration_spec.rb:583 # Dependabot::EndToEndJob composer updates dependencies correctly
rspec ./spec/dependabot/integration_spec.rb:442 # Dependabot::EndToEndJob JavaScript updates dependencies correctly
rspec ./spec/dependabot/integration_spec.rb:517 # Dependabot::EndToEndJob JavaScript summarizes the changes
```

jakecoffman:
I removed the two integration tests that remain since the smoke tests cover them and there's no way to test them without building a container for each, which is what the smoke tests already do.

Dockerfile.updater-core
```diff
@@ -125,6 +125,8 @@ RUN apt-get update \
     git-lfs \
     bzr \
     mercurial \
+    # needed to sign commits
```
jeffwidman, line 128: Oh this works??!

I thought `# comments` in the middle of single-line bash scripts that are made multi-line via `\` typically broke things... But I guessdocker is smart enough to strip the comment out before passing the command through? TIL!

common/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM ghcr.io/dependabot/dependabot-updater-core
```
jeffwidman, line 1: Might want to add a comment saying "placeholder docker file used only for running unit tests here in common"

otherwise it'll be less clear to anyone who randomly discovers this file while code spelunking...

jeffwidman (approved):

Dockerfile.updater-core
jakecoffman: Yeah it's a Docker thing!

PR #6512, opened 2023-01-24
Fallback to github.com for actions source

pavera:
## Context
There have been a few reported issues where Dependabot is attempting to source actions incorrectly from non-github.com sources. 

## Approach
Following @mctofu's advice, this implements a check and fallback so non-github sources will work, but will default to github.com if the non-github source is incorrect/inaccessible.

### DIFF
--- a/github_actions/lib/dependabot/github_actions/file_parser.rb
+++ b/github_actions/lib/dependabot/github_actions/file_parser.rb
@@ -53,20 +53,16 @@ module Dependabot
       end
 
       def build_github_dependency(file, string)
-        details = string.match(GITHUB_REPO_REFERENCE).named_captures
-        name = "#{details.fetch('owner')}/#{details.fetch('repo')}"
         unless source.hostname == "github.com"
-          potential_url = "https://#{source.hostname}/#{name}"
-          dep = github_dependency(file, string, potential_url)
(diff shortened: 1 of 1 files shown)


mctofu (approved):

landongrindheim (approved):

github_actions/lib/dependabot/github_actions/file_parser.rb
```diff
@@ -55,8 +55,20 @@ def workfile_file_dependencies(file)
       def build_github_dependency(file, string)
         details = string.match(GITHUB_REPO_REFERENCE).named_captures
         name = "#{details.fetch('owner')}/#{details.fetch('repo')}"
```
deivid-rodriguez, line 57: The above two lines are duplicated in the `github_dependency` method. I think you can remove them by changing `github_dependency`'s third parameter to receive a hostname, instead of a URI, and building the URI inside the `github_dependency` method, right when instantiating the `Dependency`:

```ruby
Dependency.new(
  # ...
  url: "https://#{hostname}/#{name}"
  # ...
)
```

deivid-rodriguez (approved):

jeffwidman:
This improves the behavior that first landed in:
* #5469

Issue #6270, opened 2022-12-06
Lockfile not updated for NPM workspaces using v3 lockfiles


Mr0grog:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

npm

### Package manager version

9.x

### Language version

Node.js 18.x

### Manifest location and content before the Dependabot update

`/package.json`, `/package-lock.json`, `/package-a/package.json`, `/package-b/package.json`

### dependabot.yml content

```yml
version: 2
updates:
  - package-ecosystem: "npm"
    directory: "/"
    schedule:
      interval: "weekly"
    versioning-strategy: increase
    open-pull-requests-limit: 10
```

### What you expected to see, versus what you actually saw

I recently updated a repo from having multiple separate packages to having multiple package joined under an NPM workspace (that is, there’s a `package.json` in the root directory with a `workspaces` property that points to the subdirectories with actual packages that have code and dependencies). I also happened to update the `package-lock.json` file from lockfile version 2 (NPM 8 and earlier) to version 3 (NPM 9).

When dependabot creates PRs against the new repository, it no longer updates the `package-lock.json` file in the root directory. However, it does so just fine when the lockfile uses the version 2 format.

For a live example, see these two repos, which are identical except for what lockfile format they use:
- https://github.com/Mr0grog/test-dependabot-npm-lockfile-v2-workspace uses the version 2 lockfile, and the PRs in it update the `package-lock.json` file along with the relevant `package.json` file(s). Example PR: https://github.com/Mr0grog/test-dependabot-npm-lockfile-v2-workspace/pull/2/files
- https://github.com/Mr0grog/test-dependabot-npm-lockfile-v3-workspace uses the version 3 lockfile, and the PRs in it *do not* update the `package-lock.json` file. Example PR: https://github.com/Mr0grog/test-dependabot-npm-lockfile-v3-workspace/pull/1/files

This all seems to work fine with version 3 lockfiles in non-workspaces repositories. For example: https://github.com/Mr0grog/test-dependabot-npm-lockfile-v3

### Images of the diff or a link to the PR, issue, or logs

For a live example, see these two repos, which are identical except for what lockfile format they use:
[...]

PR #6518, opened 2023-01-25
Fix updating v3 npm lockfiles

deivid-rodriguez:
Fixes https://github.com/dependabot/dependabot-core/issues/6270.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
@@ -155,30 +155,36 @@ module Dependabot
         def recursively_fetch_npm_lock_dependencies(object_with_dependencies)
           dependency_set = Dependabot::NpmAndYarn::FileParser::DependencySet.new
 
-          object_with_dependencies.
-            fetch("dependencies", {}).each do |name, details|
-              next unless semver_version_for(details["version"])
+          dependencies = object_with_dependencies["dependencies"]
+          dependencies ||= object_with_dependencies.fetch("packages", {}).transform_keys do |name|
+            name.delete_prefix("node_modules/")
+          end
 
-              dependency_args = {
-                name: name,
-                version: semver_version_for(details["version"]),
-                package_manager: "npm_and_yarn",
-                requirements: []
-              }
+          dependencies.each do |name, details|
+            next if name.empty? # v3 lockfiles include an empty key holding info of the current package
 
-              if details["bundled"]
-                dependency_args[:subdependency_metadata] =
-                  [{ npm_bundled: details["bundled"] }]
-              end
+            next unless semver_version_for(details["version"])
 
-              if details["dev"]
-                dependency_args[:subdependency_metadata] =
-                  [{ production: !details["dev"] }]
-              end
+            dependency_args = {
+              name: name,
+              version: semver_version_for(details["version"]),
+              package_manager: "npm_and_yarn",
+              requirements: []
+            }
 
-              dependency_set << Dependency.new(**dependency_args)
-              dependency_set += recursively_fetch_npm_lock_dependencies(details)
+            if details["bundled"]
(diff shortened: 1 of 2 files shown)


jurre (approved):

deivid-rodriguez:
I [tried this on a real repo](https://github.com/dsp-testing/test-dependabot-npm-lockfile-v3-workspace/pull/1) and it worked fine, except for hitting this npm issue: https://github.com/npm/cli/issues/2700. I think the new PRs are still in a much better place that without updating the lockfile at all, so good to go I think!

deivid-rodriguez:
https://github.com/npm/cli/issues/5470 also seems related. In any case, the better lockfile seems the one that includes "name" fields, and that's what Dependabot is generating. Whatever it is, it seems like a package manager issue.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
```diff
@@ -155,31 +155,37 @@ def shrinkwrap_dependencies
...
+          end
 
-              dependency_args = {
-                name: name,
-                version: semver_version_for(details["version"]),
-                package_manager: "npm_and_yarn",
-                requirements: []
-              }
+          dependencies.each do |name, details|
+            next if name.empty? # v3 lockfiles include an empty key holding info of the current package
```
jeffwidman, line 164: So if I'm reading correctly the code that changed is really only the lines above... below is just whitespace changes due to nesting within the `each do` block...??

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
```diff
@@ -155,31 +155,37 @@ def shrinkwrap_dependencies
         def recursively_fetch_npm_lock_dependencies(object_with_dependencies)
           dependency_set = Dependabot::NpmAndYarn::FileParser::DependencySet.new
 
-          object_with_dependencies.
-            fetch("dependencies", {}).each do |name, details|
-              next unless semver_version_for(details["version"])
+          dependencies = object_with_dependencies["dependencies"]
+          dependencies ||= object_with_dependencies.fetch("packages", {}).transform_keys do |name|
+            name.delete_prefix("node_modules/")
```
jeffwidman, line 160: Why does this get deleted? Also why is is fetching `"packages"` when before it didn't?

jeffwidman (commented):
Can you provide a little more context for those unfamiliar with the changes that landed in the new `npm` lockfile format?

I read the linked issue, but afraid I still don't quite follow...

> I think the new PRs are still in a much better place that without updating the lockfile at all, so good to go I think!

Yeah, as long as the logic is correct for what we do, doesn't seem like we should block this on upstream bugs unless they impact our behavior...

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
deivid-rodriguez: Yes, this is one PR to review with `w=1` :)

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
deivid-rodriguez: This is the v3 format, which no longer provides `dependencies` key, and prepends `node_modules/` to packages names.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
jeffwidman: oh right, you can do that in the github UI, awesome! makes it much, much simpler

jeffwidman (approved):

deivid-rodriguez:
Merging since missing lockfile updates is worse than some potential flip-flop on lockfile format due to npm issues.

Issue #6517, opened 2023-01-25
Can't use dependabot with a jruby project


ylecuyer:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

bundler

### Language version

jruby 9.3.8.0

### What you expected to see, versus what you actually saw

Dependabot-core can't be installed on jruby because the commonmarker dependency has some native code not compatible with jruby.

I came across this issue because when running dependabot with ruby on a jruby project, it will update the Gemfile.lock using the ruby platform and not the jruby platform. I wanted to run dependabot with jruby on a jruby project to have it using the correct platform but it isn't working

I also tried to change the native helper to call jruby instead of ruby (as hardcoded here: https://github.com/dependabot/dependabot-core/blob/v0.202.0/bundler/lib/dependabot/bundler/native_helpers.rb#L18) but I'm hitting the same bug as in: https://github.com/dependabot/dependabot-core/issues/5670#issuecomment-1243475874

deivid-rodriguez:
Thanks for reporting this. Platforms are quite tricky for Bundler (and transitively for Dependabot too). I'll try to make the dev experience issues reported at #5670 better so that you can dig into this more.

deivid-rodriguez:
@ylecuyer Just some heads up that #5670 was fixed, so you may be able to dig deeper into this yourself now!

ylecuyer:
@deivid-rodriguez thanks for the reminder, I'll give it another try at my second solution to this issue

ylecuyer:
with #5760 fixed I was able to proceed with the test.

Here is what I use to have on a simple example:

```diff
diff --git a/Gemfile b/Gemfile
index 9d18ad2..f551879 100644
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,3 @@
 source "https://rubygems.org"
 
-gem "nokogiri", "~> 1.12.0"
+gem "nokogiri", "~> 1.14.2"
diff --git a/Gemfile.lock b/Gemfile.lock
index 14a116b..43a846b 100644
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,15 +1,19 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    nokogiri (1.12.5-java)
+    nokogiri (1.14.2-java)
       racc (~> 1.4)
+    nokogiri (1.14.2-x86_64-linux)
+      racc (~> 1.4)
+    racc (1.6.2)
[...]

Issue #6457, opened 2023-01-17
Drop support for Bundler 1


jeffwidman:
Today we support both Bundler v1 and v2.

At some point we'll stop supporting Bundler v1 here in `dependabot-core` entirely as we're a small team so can't maintain deprecated ecosystems forever.

Re: timing... 
Although I couldn't find an official deprecation notice for Bundler v1, upgrading to Ruby `3.2` breaks Bundler v1: 
* https://github.com/dependabot/dependabot-core/pull/6436#issuecomment-1382661456

A potential workaround is using Bundler v2 to handle updating Bundler v1 configs:
* https://github.com/dependabot/dependabot-core/pull/6436#issuecomment-1382697809

If the workaround is straightforward, then it's not urgent, so may stick around for a while, depending on whether our metrics show it's a commonly used ecosystem or not.

So this is a placeholder ticket to track any discussions around timing, things that we'll need to update/remove, etc.

If you're a user of Dependabot who is still on Bundler v1 and you cannot upgrade to Bundler v2 for some reason, please chime in here explaining your use case.

jeffwidman:
From @deivid-rodriguez in https://github.com/dependabot/dependabot-core/pull/6436#issuecomment-1382697809:
> One issue is indeed what to do with Bundler 1. I see several options:
> 
> 1. Drop support for Bundler 1 before upgrading to Ruby 3.2. That effectively blocks the Ruby upgrade until we decide we're ready to drop Bundler 1.
> 
> 2. Monkey patch Bundler 1 to remove all taint/untaint. This seems doable but messy, specially since I think this is used in quite a few places.
> 
> 3. Let Bundler 2 also update lockfiles created with Bundler 1. This sounds a bit like cheating but I think it should be fine because the lockfile format has not really changed between Bundler 1 and Bundler 2.
> 
> I think the latter may be the best course of action because it allows us to effectively get rid of Bundler 1 internally, while still being able to provide updates to lockfiles locked to Bundler 1.

Option 3 does sound like the best course of action if it's straightforward. That said, if it's very much work we should probably first collect metrics on how much Bundler v1 is actually even used before investing much time into it... 

[...]

PR #6520, opened 2023-01-25
Decouple Bundler native helper specs from standard runtime

### LINKED ISSUE #5670: Bundler requires test gems when parsing dependencies

### Package ecosystem

Bundler

### Package manager version

bundler 2.4.0.dev

### Language version

ruby 2.7.6p219

_No response_

### dependabot.yml content

_No response_

### Updated dependency

_No response_

Hello, when running this:

```ruby
gitlab_bundler_parser = Dependabot::FileParsers.for_package_manager('bundler').new(dependency_files: files, source: gitlab_source, credentials: credentials)
dependencies = gitlab_bundler_parser.parse
```

I get an error, after adding `DEBUG_HELPERS=true` I see this is because this line:

```
{"BUNDLER_VERSION"=>"2.3.18", "BUNDLE_GEMFILE"=>"/home/ylecuyer/.rbenv/versions/2.7.6/lib/ruby/gems/2.7.0/gems/dependabot-bundler-0.202.0/helpers/v2/Gemfile", "GEM_HOME"=>"/home/ylecuyer/.rbenv/versions/2.7.6/lib/ruby/gems/2.7.0/gems/dependabot-bundler-0.202.0/helpers/v2/.bundle"}
bundle exec ruby /home/ylecuyer/.rbenv/versions/2.7.6/lib/ruby/gems/2.7.0/gems/dependabot-bundler-0.202.0/helpers/v2/run.rb
```

Is trying to fetch the dependencies listed here: https://github.com/dependabot/dependabot-core/blob/main/bundler/helpers/v2/Gemfile

Afaiu this gemfile is usefull only for spec and shouldn't error when running outside of specs.

I tried removing those lines: https://github.com/dependabot/dependabot-core/blob/main/bundler/lib/dependabot/bundler/native_helpers.rb#L48-L50 and it worked for me

It looks like this has been changed last year with https://github.com/dependabot/dependabot-core/commit/1535a204faf63ef331dd5ef48d9ed45f29173bc7#diff-833398d42331e02a2fd4bc36f042c10f78764d39a6d6a6b537e80e024cc55e0cR19 it is strange that I am the first one encountering the issue, maybe I'm doing something wrong.

_No response_

_No response_

_No response_

deivid-rodriguez:
Currently native helpers run in the context of a Gemfile with no gems. This always felt wrong to me.

My understanding is that there are two motivations for this:

* Run in presence of a lockfile with `BUNDLED WITH` set to either Bundler 1 or Bundler 2, so that RubyGems can properly activate the proper version according to which native helpers need to run.
* Optionally add test gems to the "empty by default" Gemfile, so that native helper specs can run.

Turns out none of these are needed, because:
* The proper version of Bundler can be also activated through `BUNDLER_VERSION`, which we already set.
* Native helper specs have the same dependencies as regular specs, so we can use the same Bundler context for running those.

The previous approach felt a bit overcomplicated, but also caused some issues when dependabot-core is used as a library, because native helpers would not run unless:

* Gems needed only for Dependabot to run their own tests are explicitly installed
OR
* Bundler is explicit configured as `bundle config without test`, which is what we do ourselves to avoid this problem in production.

An example of this last issue is explained at https://github.com/dependabot/dependabot-core/issues/5670, but has also recently caused issues for potential contributors to dependabot, for example, at https://github.com/dependabot/dependabot-core/issues/6517.

Closes #5670.


jurre (approved):
Yeah this makes sense and seems like a nice cleanup to me! It might be good to test this in staging on a real-world repo before deploying though just to make sure everything still works as expected?

deivid-rodriguez:
Absolutely, I think this is the kind of PR that deserves some manual testing in staging. I'll do that!

jeffwidman (approved):
Makes sense. I look forward to when we can move forward with https://github.com/dependabot/dependabot-core/issues/6457 which will simplify this even further.

deivid-rodriguez:
Yeah, the weirdness of running updates in an empty `bundler/setup` context would still be there, so this improvement should be independent from supporting Bundler 1 and Bundler 2 or just once. But I admittedly needed a few hacks for Bundler 1 to make this work :)

PR #6526, opened 2023-01-26
make gradle display_name property use full dependency name

### LINKED ISSUE #3353: Allow option to use full dependency name for gradle `display_name` property
As seen [here](https://github.com/dependabot/dependabot-core/blob/main/gradle/lib/dependabot/gradle.rb#L24), Dependabot uses the last artifact of a colon delimited string as the `display_name` for Gradle dependancies which ultimately make it to the commit message/PR title. For some of our Android dependencies this can end up being a very similar name or even the same in some cases.

This causes a small inconvenience when parsing through git commits in command line or release logs as they make it harder to understand at a glance which dependency was updated.

If the full name was used, the commit message would provide more useful information at a glance. 

**Example dependancies:**
com.amplitude:android-sdk
com.optimizely.ab:android-sdk

**Current PR titles for these:**
Bump android-sdk from 2.30.1 to 2.30.2
Bump android-sdk from 3.10.0 to 3.10.1 

**PR titles after proposed change:**
Bump com.amplitude:android-sdk from 2.30.1 to 2.30.2
Bump com.optimizely.ab:android-sdk from 3.10.0 to 3.10.1

mrwhoknows55:
closes #3353

### DIFF
--- a/gradle/lib/dependabot/gradle.rb
+++ b/gradle/lib/dependabot/gradle.rb
@@ -22,6 +22,6 @@ Dependabot::Dependency.
     "gradle",
     lambda { |name|
       artifact_id = name.split(":").last
-      %w(bom library).include?(artifact_id) ? name : artifact_id
+      name.length <= 100 ? name : artifact_id
     }
   )
--- a/gradle/spec/dependabot/gradle_spec.rb
+++ b/gradle/spec/dependabot/gradle_spec.rb
@@ -20,8 +20,8 @@ RSpec.describe Dependabot::Gradle do
     it { is_expected.to eq("group.com:dep") }
 
     context "with a 100+ character name" do
-          let(:name) { "com.long-domain-name-that-should-be-replaced-by-ellipsis.this-is-longer-group-id:the-longest-artifact-id" }
-          it { is_expected.to eq("the-longest-artifact-id") }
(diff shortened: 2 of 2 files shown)


jeffwidman:
@mrwhoknows55 want to pick [one of these solutions](https://github.com/dependabot/dependabot-core/issues/3353#issuecomment-1405417513) and run with it?:

> So there's a lot of variables there, but if we just cap max length of the full dependency name at 100 chars that's probably good enough.
> 
> For when it's over 100 chars, then if we want to get fancy, truncate the path, ie `full-pa../package-name`. But given that will be rare, the simpler option is to punt and if longer then 100 chars revert back to current behavior of only displaying package name. I'm personally fine either way, we can always tweak / improve it later.

jeffwidman (approved):
Looks great, thanks!

jeffwidman:
Can you fix the lint failures? https://github.com/dependabot/dependabot-core/actions/runs/4025933944/jobs/6925118702

For the line too long, I'm fine if you disable that check for that line, IMO the point is to have a long package name...
```
method(argument) # rubocop:disable Layout/LineLength
```

I suspect the smoke test failure is because the smoke test needs updating to match the new expected behavior--I can handle that post merge.

jeffwidman:
I went ahead and fixed the lints so we could land this.

jeffwidman:
I fixed the failing smoke test in https://github.com/dependabot/smoke-tests/pull/27.

PR #6530, opened 2023-01-27
remove the monolithic image

jakecoffman:
The ecosystems each have their own Docker image now. The Dockerfiles live under the module path, so `bundler/Dockerfile` is the Dockerfile for the Bundler updater. So there's no longer a reason to keep publishing dependabot-core and dependabot-updater.

We'll be replacing a few places where we use the dependabot-core image with ecosystem-specific images soon, like dependabot-script.

Note: This is a stacked PR, parent is https://github.com/dependabot/dependabot-core/pull/6502.


.github/workflows/ci.yml
```diff
@@ -195,45 +188,46 @@ jobs:
...
             --rm \
             -v "$(pwd)/updater/spec/fixtures/vcr_cassettes:/home/dependabot/dependabot-updater/spec/fixtures/vcr_cassettes" \
-            "dependabot-updater:latest" bundle exec rspec
-        env:
-          SKIP_BUILD: true
-          DEPENDABOT_TEST_ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+            -v "$(pwd)/npm_and_yarn:/home/dependabot/npm_and_yarn" \
+            -v "$(pwd)/go_modules:/home/dependabot/go_modules" \
+            -v "$(pwd)/composer:/home/dependabot/composer" \
+            "dependabot/dependabot-updater-bundler:latest" bundle exec rspec
```
jeffwidman, line 230: Can you add a code comment clarifying why _these_ four ecosystems in particular have to be mounted in here? I know why, but if I lacked that context from other PR's, it'd be confusing why these and not any others...

jeffwidman (approved):
It's too bad that this is going to make `git blame` a lot more painful. Would have been nice to preserve that, particularly for the `Dockerfile`. But that ship has mostly sailed, oh well, at least the new image files are pretty well commented.

.github/workflows/ci.yml
```diff
@@ -9,47 +9,40 @@ on:
...
+          - { path: npm_and_yarn, name: npm_and_yarn, ci_node_total: 3, ci_node_index: 0, ecosystem: npm }
+          - { path: npm_and_yarn, name: npm_and_yarn, ci_node_total: 3, ci_node_index: 1, ecosystem: npm }
+          - { path: npm_and_yarn, name: npm_and_yarn, ci_node_total: 3, ci_node_index: 2, ecosystem: npm }
+          - { path: nuget, name: nuget, ecosystem: nuget }
+          - { path: pub, name: pub, ci_node_total: 2, ci_node_index: 0, ecosystem: pub }
+          - { path: pub, name: pub, ci_node_total: 2, ci_node_index: 1, ecosystem: pub }
+          - { path: python, name: python, ci_node_total: 2, ci_node_index: 0, ecosystem: pip }
+          - { path: python, name: python, ci_node_total: 2, ci_node_index: 1, ecosystem: pip}
+          - { path: python, name: python_slow, ci_node_total: 2, ci_node_index: 0, ecosystem: pip }
+          - { path: python, name: python_slow, ci_node_total: 2, ci_node_index: 1, ecosystem: pip }
```
jeffwidman, line 44: Oh, one more question: Why is the number of python runs cut from 4 to 2?

jakecoffman: @pavera any reason for that?

deivid-rodriguez: I suspect it's because @pavera sped up a lot the Python suite when he started pre-building python versions at `docker build` time, so now these are fast enough to use just two nodes.

PR #6531, opened 2023-01-27
Renable golang tests

deivid-rodriguez:
When we introduced parallelization they were disabled due to a typo.


jakecoffman (approved):
d'oh, thanks!

deivid-rodriguez:
You're actually fixing this at #6530 😄. I'm surprised they all still pass since they have not run for months. I guess the go ecosystem has not changed a lot recently. Always, I'll close in favor of #6530!

deivid-rodriguez:
Oh, just saw you canceled the "Merge & Squash" at #6530. We can still introduce this PR first I guess, since it's a separate fix after all. Up to you!

PR #6538, opened 2023-01-29
Fix running tests in devcontainer

### LINKED ISSUE #6534: Error running terraform tests in devcontainer

Hello,

this issue is rather about some personal support. 

I am trying to contribute to one of the issues here in this repo but I am simply not getting the devcontainer running correctly. I oriented on the [Contributing Guidelines](https://github.com/dependabot/dependabot-core/blob/main/CONTRIBUTING.md) and the [dependabot-script](https://github.com/dependabot/dependabot-script) repo for setting up the helpers.

The devcontainer is build, I run the `bundle install` in the container, all Gems are correctly setup ( at least I think so). As I want to work on a terraform language feature I have also setup the terraform helper + the required hcl2json for the tests cases.

I try to execute the "Debug dry run" or "Debug tests" (on terraform - spec) but I always get the following error message:

```
/usr/local/rvm/gems/ruby-3.1.3/gems/bundler-2.4.5/lib/bundler/rubygems_integration.rb:304:in `block in replace_bin_path': can't find executable rdebug-ide for gem ruby-debug-ide. ruby-debug-ide is not currently included in the bundle, perhaps you meant to add it to your Gemfile? (Gem::Exception)
	from /usr/local/rvm/gems/ruby-3.1.3/gems/bundler-2.4.5/lib/bundler/rubygems_integration.rb:332:in `block in replace_bin_path'
	from /usr/local/rvm/gems/ruby-3.1.3/bin/rdebug-ide:32:in `<main>'
```

The error still occurs even after I explictily ran `gem install ruby-debug-ide`. The test execution works if I run `cd terraform && bundle exec rspec spec`, sadly I am not sure how to attach a debugger with this command. The execution still fails with 210 examples, 35 failures on testcases so even this setup is not working 100%.

Other packages (like maven) throws a different error:

```
/usr/local/rvm/gems/ruby-3.1.3/gems/ruby-debug-ide-0.7.3/lib/ruby-debug-ide.rb:9:in `require': cannot load such file -- debase (LoadError)
	from /usr/local/rvm/gems/ruby-3.1.3/gems/ruby-debug-ide-0.7.3/lib/ruby-debug-ide.rb:9:in `<top (required)>'
	from /usr/local/rvm/gems/ruby-3.1.3/gems/ruby-debug-ide-0.7.3/bin/rdebug-ide:8:in `require_relative'
	from /usr/local/rvm/gems/ruby-3.1.3/gems/ruby-debug-ide-0.7.3/bin/rdebug-ide:8:in `<top (required)>'
	from /usr/local/rvm/gems/ruby-3.1.3/bin/rdebug-ide:25:in `load'
	from /usr/local/rvm/gems/ruby-3.1.3/bin/rdebug-ide:25:in `<main>'
	from /usr/local/rvm/gems/ruby-3.1.3/bin/ruby_executable_hooks:22:in `eval'
	from /usr/local/rvm/gems/ruby-3.1.3/bin/ruby_executable_hooks:22:in `<main>'
```

This definetly blocks me from getting deeper into the project so I am happy about any help!

deivid-rodriguez:
I run into similar issues to #6534.

I investigated and found the culprit to be that `BUNDLE_PATH` is no longer set in the devcontainer so gems get installed globally, not at the path they got previously installed. So `bundle exec rspec` no longer works.

The reason is that the base Dockerfile is now multistage, and the `ENV BUNDLE_PATH` line was left behind in the builder stage, but it's no longer present in the final image.

This PR fixes the issue plus other little things I found while investigating. More details in commit messages.

Fixes #6534.

### DIFF
--- a/Dockerfile.updater-core
+++ b/Dockerfile.updater-core
@@ -42,9 +42,8 @@ ARG RUBY_INSTALL_VERSION=0.8.5
 
 # When bumping Bundler, need to also regenerate `updater/Gemfile.lock` via `bundle update --bundler`
 # Generally simplest to match the bundler version to the one that comes by default with whatever Ruby version we install.
-# This way other projects that import this library don't have to futz around with installing new /unexpected bundler versions.
+# This way other projects that import this library don't have to futz around with installing new / unexpected bundler versions.
 ARG BUNDLER_V2_VERSION=2.3.26
-ENV BUNDLE_SILENCE_ROOT_WARNING=1
 
 # Install Ruby, update RubyGems, and install Bundler
 RUN mkdir -p /tmp/ruby-install \
@@ -64,9 +63,6 @@ WORKDIR ${HOME}
 # Place a git shim ahead of git on the path to rewrite git arguments to use HTTPS.
 ARG SHIM="https://github.com/dependabot/git-shim/releases/download/v1.4.0/git-v1.4.0-linux-amd64.tar.gz"
 RUN curl -sL $SHIM -o git-shim.tar.gz && mkdir -p ~/bin && tar -xvf git-shim.tar.gz -C ~/bin && rm git-shim.tar.gz
-
-ENV DEPENDABOT_HOME /home/dependabot
-WORKDIR ${DEPENDABOT_HOME}
 
 COPY updater/Gemfile updater/Gemfile.lock dependabot-updater/
 COPY .ruby-version .ruby-version
@@ -96,7 +92,7 @@ RUN for ecosystem in git_submodules terraform github_actions hex elm docker nuge
       touch $ecosystem/lib/dependabot/$ecosystem.rb; \
     done
 
-WORKDIR $DEPENDABOT_HOME/dependabot-updater
+WORKDIR $HOME/dependabot-updater
 
 RUN bundle config set --local path 'vendor' && \
 bundle config set --local frozen 'true' && \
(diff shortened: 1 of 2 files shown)


jeffwidman (approved):
This seems reasonable to me, but for the `ENV` removals you might want to get a review from @jakecoffman or @pavera since they so recently worked on this area.

Dockerfile.updater-core
```diff
@@ -42,9 +42,6 @@ ARG RUBY_INSTALL_VERSION=0.8.5
 
 # When bumping Bundler, need to also regenerate `updater/Gemfile.lock` via `bundle update --bundler`
 ARG BUNDLER_V2_VERSION=2.3.25
-ENV BUNDLE_SILENCE_ROOT_WARNING=1
```
jeffwidman, line 45: Does this need to be added in a later stage or is that already happening? (On my phone so can't check source)

deivid-rodriguez: My rationale was since no longer setting it in the updater image has not broken anything, it was not needed. This makes sense since I don't think we run Bundler as root. Maybe we did this in the past and that's why we needed this to tell Bundler: "We know this is bad, but we know what we're doing". But I think we're using better practices now.

jakecoffman (approved):
Thanks for cleaning this up!

pavera (approved):

PR #6539, opened 2023-01-30
Re-bump Ruby to `3.1.3` to fix regression

jeffwidman:
During the recent Dockerfile refactoring to split into per-ecosystem, the ruby version regressed from `3.1.3` to `3.1.2`... probably because some of the changes from https://github.com/dependabot/dependabot-core/pull/6257 weren't pulled in.

So this re-applies them. I checked the other files change in that PR and they all still looked correct, this was the only file from that PR that regressed.


jakecoffman (approved):
Sorry about that!

PR #6549, opened 2023-01-30
allow targetarch to be set by Docker buildkit

jakecoffman:
According to the [Docker docs](https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope), we shouldn't have to set the TARGETARCH if we simply declare it in the Dockerfiles as `ARG TARGETARCH` when using Docker buildkit. 

This simplifies things a bit and allows any other architectures that our scripts don't explicitly handle to build the containers.

I tested this on my ARM machine and it works.


Dockerfile.updater-core
```diff
@@ -1,7 +1,5 @@
 FROM ubuntu:20.04 as builder
 
-ARG TARGETARCH=amd64
```
jakecoffman, line 3: This wasn't used in this file so I removed it.

Nishnha:
Could this close out https://github.com/dependabot/dependabot-core/pull/5536?
That PR also adds a build matrix for the new target arch but I don't think that's necessary since we have codespaces

jakecoffman:
@Nishnha no we still will need to add a workflow to build and publish for ARM to do that. This is just to clean up the scripts a bit. 

Also I had reverted some changes @jeffwidman had done in https://github.com/dependabot/dependabot-core/pull/6502 and while researching it a bit more realized we don't have to set that value at all.

jeffwidman (approved):
Oh nice, this is even cleaner than what I'd done!

PR #6555, opened 2023-01-31
Build per ecosystem dev containers

deivid-rodriguez:
I built the python dev container with

```
bin/docker-dev-shell python --rebuild
```

Then I built the golang one with

```
bin/docker-dev-shell golang --rebuild
```

The I wanted to use the python one again, so I run

```
bin/docker-dev-shell python
```

No rebuild because I had already built it.

But I got "no pyenv in path" error.

I figure it was because rebuilding for golang overwrote the image.

Should we built dev separate dev containers for each ecosystem? That's what I did here. I think could've passed the `IMAGE_NAME` and `CONTAINER_NAME` variables, but we probably want this to be the default and no need to make it configurable.

### DIFF
--- a/bin/docker-dev-shell
+++ b/bin/docker-dev-shell
@@ -2,9 +2,6 @@
 
 set -e
 
-IMAGE_NAME="${IMAGE_NAME:=dependabot/dependabot-core-development}"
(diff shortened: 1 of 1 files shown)


jakecoffman (approved):
👍 Yeah I was going this direction but I think I usually run it with `--rebuild` always so didn't notice this.

PR #6491, opened 2023-01-20
Allow updating pom's named `pom_parent.xml`

deivid-rodriguez:
During review of #6248, we noticed that Dependabot special cases the `pom_parent.xml` name as some trick to fetch parent poms. #6248 tries to preserve that, resulting in that with that PR, Dependabot can now update any pom names, except for `pom_parent.xml`.

We concluded that this "magic name" is not necessary. So this PR removes the magic and instead adds the `pom_parent.xml` to the list of "allowed names", so repositories like https://github.com/filip26/titanium-json-ld can now get updates.

I'm hesitant to add any specs over manual testing because those specs would be exercising a special case that should completely go way after we merge #6248, since that will allow any names.

PR #6557, opened 2023-01-31
Don't fetch poms outside of configured dir

deivid-rodriguez:
When Dependabot needs some files to perform an update, but those files should not be updated, we normally use the `#support_file` dependency attribute. However, in this case, I think users normally configure directories for completely disjoint maven apps, so we can avoid fetching files outside of directory altogether.

### DIFF
--- a/maven/lib/dependabot/maven/file_fetcher.rb
+++ b/maven/lib/dependabot/maven/file_fetcher.rb
@@ -92,18 +92,9 @@ module Dependabot
       def recursively_fetch_relative_path_parents(pom, fetched_filenames:)
         path = parent_path_for_pom(pom)
 
-        return [] if fetched_filenames.include?(path)
-
-        full_path_parts =
-          [directory.gsub(%r{^/}, ""), path].reject(&:empty?).compact
-
-        full_path = Pathname.new(File.join(*full_path_parts)).
-                    cleanpath.to_path
-
-        return [] if full_path.start_with?("..")
+        return [] if fetched_filenames.include?(path) || path.start_with?("..")
 
         parent_pom = fetch_file_from_host(path)
-        parent_pom.support_file = true
 
         [
           parent_pom,
--- a/maven/spec/dependabot/maven/file_fetcher_spec.rb
+++ b/maven/spec/dependabot/maven/file_fetcher_spec.rb
@@ -325,9 +325,7 @@ RSpec.describe Dependabot::Maven::FileFetcher do
         let(:directory) { "/util/util" }
 
         it "fetches the relevant poms" do
-          expect(file_fetcher_instance.files.count).to eq(3)
-          expect(file_fetcher_instance.files.map(&:name)).
-            to match_array(%w(pom.xml ../pom.xml ../../pom.xml))
+          expect(file_fetcher_instance.files.map(&:name)).to eq(%w(pom.xml))
         end
       end
 


jeffwidman:
Documenting this tries to fix the bug mentioned in https://github.com/dependabot/dependabot-core/pull/6491#issuecomment-1409276060.

jeffwidman (approved):

deivid-rodriguez:
Mmmm I think what fixes that bug is... this PR 😃.

We can add some docs for this, but I did also try the effect of configuring Dependabot to update directory `/foo` where some parent poms in `/foo` live in the root directory, and I got the expected behavior (no updates to root poms). So maybe it's ok without any extra docs.

jeffwidman:
Sorry, there's a misunderstanding here.

I didn't mean add docs, I meant my comment was simply noting/documenting this PR fixes that bug mentioned in the comments of that PR, since before the cross-link wasn't as obvious as a clear comment.

As per normal we are completely on the same page. 👍 😆

deivid-rodriguez:
🤦 Sorry, totally misread your message 😆

soulus13:
Hi @deivid-rodriguez , @jeffwidman , just accidentally stumbled upon this PR, and i would propose to correct that fix a little. Even though i do agree this fixes the problem raised - it is not fully aligned now with the expected behavior on the Maven projects' side and the previous improvements made 😃 And once again - sorry for my long reads below 😃 

**Short version:**
I would propose to revert this fix and instead add one more line in `Dependabot::Maven::FileFetcher#parent_path_for_pom` to check whether the XML path `project->parent` is present at all, and if not - stop the flow right away. And also stop the processing if `project->parent-relativePath` tag is present but empty. 
What it currently does - it goes right away for the `relativePath` (with a valid fallback to `..`), which is indeed incorrect when no parent has been specified. That should, basically, fix the issue raised in https://github.com/dependabot/dependabot-core/pull/6491#issuecomment-1409276060, while still preserving the previous functionality.

**Long version:**
Here is the spec that properly describes how maven works with the relativePath, understanding whether there is a parent there or not - https://maven.apache.org/ref/3.2.3/apidocs/org/apache/maven/model/Parent.html#getRelativePath(). Just citing it here to go through that:
> Get the relative path of the parent pom.xml file within the check out. If not specified, it defaults to ../pom.xml. Maven looks for the parent POM first in this location on the filesystem, then the local repository, and lastly in the remote repo. relativePath allows you to select a different location, for example when your structure is flat, or deeper without an intermediate parent POM. However, the group ID, artifact ID and version are still required, and must match the file in the location given or it will revert to the repository for the POM. This feature is only for enhancing the development in a local checkout of that project. Set the value to an empty string in case you want to disable the feature and always resolve the parent POM from the repositories.

So it is pretty natural for Maven `pom.xml` file to have his parent specified like that, without the `relevantPath`
```
<parent>
    <groupId>org.example</groupId>
    <artifactId>maven-test-parent-present</artifactId>
    <version>1.0</version>
</parent>
```
Now what maven reads from there - probably there is a parent `pom.xml` file present in the path `../pom.xml`, which it will try to look up. Now there are two cases possible:

- Maven has found this file and it's `groupId`, `artifactId` and `version` fields correspond to the ones defined above - it is deemed as a valid `pom.xml` parent file, and all the data from it is used going forward in any maven tasks. In this case it would generally be expected that Dependabot checks for any dependencies within this file and creates the PR to update any of its dependencies.
- Maven has found this file and it's `groupId`, `artifactId` and `version` don't match to the specified ones OR the file was not found at all - then this parent file is skipped and not used, and maven will try to lookup the parent in the maven central repository (or any other maven repos specified) instead. In this case it would be expected that Dependabot just checks if there is any newer version of this file in maven central (or other specified repositories) and creates the PR to bump this version if needed.

Now getting back to the example in this repo - the behavior demonstrated by Dependabot on https://github.com/jtgrohn/maven-dependabot-test repo before this PR was incorrect, but not because it tried to lookup and update the parent pom per se - that is generally fine, the Dependabot logic handling the case when the `relevantPath` is not specified in the "root" `pom.xml` was a correct one (within `Dependabot::Maven::FileFetcher#parent_path_for_pom`): 
- Check if relative path node is present, if not - default to `..`:
```
relative_parent_path = doc.at_xpath("/project/parent/relativePath")&.content&.strip || "..
```
- and then adds to it `pom.xml`, as the resulting path is a dir, not a file:
```
        name_parts = [
          File.dirname(pom.name),
          relative_parent_path,
          relative_parent_path.end_with?(".xml") ? nil : "pom.xml"
        ].compact.reject(&:empty?)
```
- all that results in saying - let's check if there is a parent on the `../pom.xml` path. 

This logic above fine, and everything that followed it before in `Dependabot::Maven::FileFetcher#recursively_fetch_relative_path_parents` also seems to be good:
- either we stop when we are trying to escape the repo boundaries (`path.start_with?("..")` part)
- or we stop when the parent file is not found and the exception `Dependabot::Maven::FileFetcher::Dependabot::DependencyFileNotFound` is thrown, which also makes sense

What is really missing - before all that it is mandatory to check if the `parent` node is present in the file at all (and that the `relativePath` was not intentionally set to be empty), as otherwise it is trying to find the file which noone has instructed to look for 😃 And this, indeed, would cause the issue https://github.com/dependabot/dependabot-core/pull/6491#issuecomment-1409276060

**Extra note:**
In addition to all the above, when fetching the parent - it would be really nice to add a check that the fetched parent file's `groupId`, `artifactId` and `version` equal to those of it 'child' pom, from which the parent is searched - otherwise such a parent pom should be deemed invalid as well. But i believe it could be looked at with a lower priority, as such cases I bet would be pretty rare 😃

deivid-rodriguez:
Wow @soulus13, thanks so much for explaining all that, really nice to have someone more familiar with Maven helping us out! I shoul've actually pinged you before merging this pR.

Would you be up for creating a PR with the revert + suggested fixup?

soulus13:
> Wow @soulus13, thanks so much for explaining all that, really nice to have someone more familiar with Maven helping us out! I shoul've actually pinged you before merging this pR.
> 
> Would you be up for creating a PR with the revert + suggested fixup?

@deivid-rodriguez , for sure! Will fire it within the weekends :)

PR #6569, opened 2023-02-01
Mount `/updater` files into `docker-dev-shell`

jeffwidman:
This way we can run those unit tests as well from within the dev shell.

They'll need to be run from within the `bundler` ecosystem.


jeffwidman:
I think this is all we need from https://github.com/dependabot/dependabot-core/tree/a2ae2480cbb584aaca104c945c2bfdc05894f82a/updater, as I don't forsee a need for `config` or `licenses` directories...

But I wasn't sure if the `bin` dir might be useful?

deivid-rodriguez (approved):
Let's mount this for now and if anything else turns out necessary, we add it later. ❤️

PR #6570, opened 2023-02-01
Map `updater` ecosystem to `bundler`

jeffwidman:
Technically every ecosystem results in building a per-ecosystem updater image. But currently running the `updater` unit tests requires the `bundler` ecosystem image.

So this is a convenience to make it so we don't have to rememeber that
and can just run `bin/docker-dev-shell updater`.

### DIFF
--- a/bin/docker-dev-shell
+++ b/bin/docker-dev-shell
@@ -21,7 +21,14 @@ fi
 if [[ -z "$1" ]]; then
   HELP=true
 fi
-ECOSYSTEM="$1"
+
+# Technically every ecosystem results in building a per-ecosystem updater image.
+# But currently running the `updater` unit tests requires the `bundler` ecosystem image.
+if [ "$1" == "updater" ]; then
+  ECOSYSTEM="bundler"
+else
+  ECOSYSTEM="$1"
+fi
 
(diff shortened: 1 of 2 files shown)


jeffwidman:
I am of two minds on this one... it's convenient, but it may perpetuate confusion if someone starts to think "updater always = bundler"... and it's not hard to remember to type `docker-dev-shell bundler` instead of `docker-dev-shell updater`

I also considered adding a `updater/script/ci-test` ala:
https://github.com/dependabot/dependabot-core/blob/eeca17f20c5d3308e1600900387bf822069bab48/cargo/script/ci-test#L1-L6

but it's kinda weird...

deivid-rodriguez (approved):
I'm good with this! I prefer `docker-dev-shell updater` for running updater tests 👍.

To slightly reduce the confusion of "updater" not being an ecosystem, you may want to instead do something like

```bash
if [ "$1" == "updater" ]; then
  ECOSYSTEM="bundler"
else
  ECOSYSTEM="$1"
fi
```

brrygrdn (approved):
Yeah, this makes sense to me 👍🏻

PR #6572, opened 2023-02-01
Remove `EndToEndJob`

jeffwidman:
This is an old artifact from before `FileFetcherJob` and `UpdateFilesJob`
existed...

We've long wanted to get rid of these, there are at least two PR's I
found in our internal repo from before we open sourced this code.

But the blocker was that we didn't have integration tests that tested
the combination of `FileFetcherJob` + `UpdateFilesJob`, so we wanted
to refactor the integration tests to test the handoff between the two.

To be clear: We _do_ have unit tests that test each of those individual
components, so we can still refactor internally knowing that we're
protected by individual unit tests for a fast dev/test cycle. It's just
the combination of the two that's not tested.

However, we now have the full `smoke` test suite which tests everything
end to end. So rather than refactoring the integration tests, it's now
safe to simply remove them.


updater/spec/dependabot/integration_spec.rb
```diff
@@ -1,379 +0,0 @@
...
-      end_to_end_job.run
-    end
-
-    it "summarizes the changes" do
-      expect(Dependabot.logger).to receive(:info).with(/Changes to Dependabot Pull Requests/) do |log_message|
-        expect(log_message).to include("created", "dummy-pkg-b ( from 1.1.0 to 1.2.0 )")
-      end
-
-      end_to_end_job.run
-    end
```
mctofu, line 163: Is this tested elsewhere? I don't think this is checked in the smoke test? If not it would be worth preserving somehow.

updater/spec/dependabot/integration_spec.rb
```diff
@@ -1,379 +0,0 @@
...
-        before do
-          allow(Dependabot::Environment).to receive(:github_actions?) { "true" }
-        end
-
-        it "raises an exception" do
-          expect { end_to_end_job.run }.to raise_error(Dependabot::RunFailure)
-        end
-      end
-    end
-  end
```
mctofu, line 241: Is this tested elsewhere? I don't think this is checked in the smoke test? If not it would be worth preserving somehow.

mctofu (commented):
I'm in favor of removing this but I saw a few tests that I don't think are covered by the smoke tests. Would like to confirm we have tests for them somewhere else or see if we can move them.

jeffwidman:
Go for it @brrygrdn !

jeffwidman:
Switch to draft based on discussion in:
* https://github.com/dependabot/dependabot-core/pull/6760#issuecomment-1450644241

jeffwidman:
Closing as https://github.com/dependabot/dependabot-core/pull/6760 handled everything we need here.

PR #6574, opened 2023-02-02
fix: use Regexp.escape for YAML tag replacement

lyoung-confluent:
dependabot doesn't use `Regexp.escape` before injecting the docker tag about the be replaced into a regex. As these tags are typically versions containing `.` (Regex wildcard), this results in an incorrect replacement regex that matches too broadly.

### DIFF
--- a/docker/lib/dependabot/docker/file_updater.rb
+++ b/docker/lib/dependabot/docker/file_updater.rb
@@ -168,7 +168,7 @@ module Dependabot
         modified_content = file.content
 
         old_tags.each do |old_tag|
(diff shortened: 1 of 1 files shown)


jeffwidman (commented):
Makes sense, thanks!

Can you add a test with a version that includes a `.` to ensure no regressions in the future?

If you need pointers on how to do that, just let us know. I'd start by looking at the other tests in https://github.com/dependabot/dependabot-core/blob/7354dd2492adcd6388a77bab39d8c09e1d45ee96/docker/spec/dependabot/docker/file_updater_spec.rb

deivid-rodriguez:
Yeah, my understanding is that we'd a test case with versions similar enough to get confused by the absence of dot escaping, for example, "1.2.3" vs "122.3".

PR #6575, opened 2023-02-02
Update README to link to list of package managers.

markdorison:
When I was working with the Docker image locally, it took me a minute to parse what the acceptable arguments were for the package manager. I added a link to the documentation, but I am not sure that is the best place, as I see `gomod` listed there for Go as opposed to `go_modules` used here. I see the [list with names in some of the GitHub actions workflows](https://github.com/dependabot/dependabot-core/blob/82906a986a5d09e31e27e038618f49d1cba940cf/.github/workflows/images-latest.yml#L36-L52). 

Is there a better, more canonical place that I could link this documentation to with the appropriate values?


jeffwidman (commented):
Thanks Mark, and I feel your pain, I hit the same thing yesterday.

Unfortunately, the naming is a bit confusing here because it's actually the ruby module name in this context... ie, `go_modules` rather than `gomod`.

This is a horrid user experience, and probably one we should cleanup sooner rather than later. 

If it's okay with you, let's float this PR for a bit as a reminder to me to look into what it'd take to start to standardize some of this to reduce confusion... I don't think saying "go_modules" one place and "gomod" is intuitive for anyone.

markdorison:
@jeffwidman That makes sense to me. There is probably more elegant way to solve this in the long run, but in the meantime, should we turn [this](https://github.com/dependabot/dependabot-core/blob/82906a986a5d09e31e27e038618f49d1cba940cf/.github/workflows/images-latest.yml#L36-L52) into a table for the README similar to what I did in https://github.com/dependabot/dependabot-core/issues/6577#issuecomment-1414354486?

robaiken (approved):

PR #6576, opened 2023-02-02
Go: update to 1.20

jakecoffman:
I don't think there were any significant changes to Go modules, but the release is something like 300 MB smaller since they stopped shipping pre-compiled library archives. So our [gomod image](https://github.com/dependabot/dependabot-core/pkgs/container/dependabot-updater-gomod) will be smaller too! 🎉


landongrindheim (approved):
🤏

Nishnha (approved):

deivid-rodriguez (approved):
🎉

Issue #2025, opened 2020-02-24
Properties used in Maven dependencyManagement not picked up correctly


wollefitz:
Dependabot doesn't correctly pick up versions that are stored in Maven properties and used in `dependencyManagement` sections. I created a [sample repository](https://github.com/wollefitz/dependabot-maven-properties) to analyze the issue a little more in-depth.

- When letting Dependabot run on the [repository's master branch](https://github.com/wollefitz/dependabot-maven-properties/blob/master/pom.xml) (no effective dependencies, just a `dependencyManagement` block) I'm getting the following error:

```
Dependabot can't evaluate your Java dependency files.

As a result, Dependabot couldn't check whether any of your dependencies are out-of-date.

The error Dependabot encountered was:

Property not found: project.version.spring-boot
```

See also [this issue Dependabot created](https://github.com/wollefitz/dependabot-maven-properties/issues/1)

- When letting Dependabot run on [a branch where I added a dependency on `org.springframework.boot:spring-boot-starter-web`](https://github.com/wollefitz/dependabot-maven-properties/tree/direct-dependency) I'm not getting the above error anymore but Dependabot doesn't create a Pull Request to bump `project.version.spring-boot` even though it understands that currently the latest version is `2.2.4.RELEASE`:

```
updater | INFO <job_24269133> Checking if org.springframework.boot:spring-boot-starter-web  needs updating
  proxy | 2020/02/24 11:41:40 GET https://repo1.maven.org:443/maven2/org/springframework/boot/spring-boot-starter-web/maven-metadata.xml
  proxy | 2020/02/24 11:41:40 200 https://repo1.maven.org:443/maven2/org/springframework/boot/spring-boot-starter-web/maven-metadata.xml
  proxy | 2020/02/24 11:41:40 GET https://repo.maven.apache.org:443/maven2/org/springframework/boot/spring-boot-starter-web/maven-metadata.xml
  proxy | 2020/02/24 11:41:40 200 https://repo.maven.apache.org:443/maven2/org/springframework/boot/spring-boot-starter-web/maven-metadata.xml
  proxy | 2020/02/24 11:41:40 HEAD https://repo1.maven.org:443/maven2/org/springframework/boot/spring-boot-starter-web/2.2.4.RELEASE/spring-boot-starter-web-2.2.4.RELEASE.jar
[...]

Issue #1859, opened 2020-05-14
[java:maven] Dependency versions defined through a property are not respected by if the property starts with 'project.'   


ingogriebsch:
It looks like Dependabot is not checking dependencies if the version of the dependency is defined through a property AND the property starts with 'project'.

This is working:
```
<properties>
	<guava.version>28.1-jre</guava.version>
</properties>

<dependencies>
        <dependency>
            <groupId>com.google.guava</groupId>
            <artifactId>guava</artifactId>
            <version>${guava.version}</version>
        </dependency>
</dependencies>
```

This is not working:
```
<properties>
	<project.guava.version>28.1-jre</guava.version>
</properties>

<dependencies>
        <dependency>
            <groupId>com.google.guava</groupId>
            <artifactId>guava</artifactId>
            <version>${project.guava.version}</version>
        </dependency>
</dependencies>
```

We created a repository explaining the issue in more detail. There you can have a look at the specific commits and the corresponding outputs of the Dependabot check. You can have a look at it here:
https://github.com/gbtec-ag/dependabot-version-property-issue

A follow up problem is that Dependabot closes the corresponding PR (if one was created) but never opens the PR again after the property is corrected. Means one needs to reopen the PRs by hand (if available).

jeffwidman:
👋 Hey there! Sorry for the glacially slow response... we've been a bit buried the past few years, and only just now starting to dig through some of our backlog.

This issue is pretty old, and we've pushed a lot of bugfixes since then... is this still a valid issue?

jeffwidman:
Keeping open despite lack of user response because this one has a public repo so should be easy to check if reproducible or not.

PR #6581, opened 2023-02-03
Fix updating properties with name starting with "project"

deivid-rodriguez:
Fixes https://github.com/dependabot/dependabot-core/issues/1859.
Fixes https://github.com/dependabot/dependabot-core/issues/2025.

### DIFF
--- a/maven/lib/dependabot/maven/file_parser/property_value_finder.rb
+++ b/maven/lib/dependabot/maven/file_parser/property_value_finder.rb
@@ -36,8 +36,8 @@ module Dependabot
             loop do
               candidate_node =
                 doc.at_xpath("/project/#{nm}") ||
-                doc.at_xpath("/project/properties/#{nm}") ||
-                doc.at_xpath("/project/profiles/profile/properties/#{nm}")
+                doc.at_xpath("/project/properties/#{property_name}") ||
+                doc.at_xpath("/project/profiles/profile/properties/#{property_name}")
               break candidate_node if candidate_node
               break unless nm.match?(DOT_SEPARATOR_REGEX)
 
--- a/maven/spec/dependabot/maven/file_parser/property_value_finder_spec.rb
+++ b/maven/spec/dependabot/maven/file_parser/property_value_finder_spec.rb
@@ -36,6 +36,12 @@ RSpec.describe Dependabot::Maven::FileParser::PropertyValueFinder do
         its([:value]) { is_expected.to eq("0.0.2-RELEASE") }
       end
 
+      context "and the property name starts with 'project' but not an attribute of the project" do
+        let(:base_pom_fixture_name) { "property_name_starts_with_project_pom.xml" }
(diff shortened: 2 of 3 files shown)


jeffwidman (approved):

jeffwidman:
Merging since CI is green, and we've got other PR's we want to keep moving so no sense merging those and then re-running CI here again.

deivid-rodriguez:
I forgot to ask for a review from @soulus13 in case he'd be fine with having a quick look 🙏.

jeffwidman:
Oh, good thought, I'd definitely appreciate your input @soulus13 since you know more about this ecosystem than myself + Deivid combined. 😄

soulus13:
Hi @deivid-rodriguez , @jeffwidman , yep, this specific fix addresses the issue nicely and should not break anything 😃 There has been indeed no need to carve out the `pom` and `project` prefixes there when looking for value in the properties' paths (and not doing lookup in xml structure). A little bit on these "custom" cases for the properties could found in https://github.com/cko/predefined_maven_properties/blob/master/README.md and https://books.sonatype.com/mvnref-book/reference/resource-filtering-sect-properties.html.

Still - i believe there is one more bug in the current implementation even after the fix:
- what `Dependabot::Maven::FileParser::PropertyValueFinder#sanitize_property_name` does - it removes the `pom` and `project` prefixes, as these are reserved prefixes to do a lookup on maven file - ok, that starts good and does make sense
- next it tries to do the lookup in the loop, slowly replacing the dots in the property names by slashes, one-by-one. That overcomplicates the things a bit, i think. You could see the pom structure here https://maven.apache.org/ref/3.0.3/maven-model/maven.html, and none of the xml's nodes have dots inside - thus it makes sense to replace them all with slashes right away for the xml structure lookup part (which is only this one `doc.at_xpath("/project/#{nm}")`)
- and then, if the replacement would be done only once - no need  for the loop to exist here at all 😃As in the current implementation we either find the result right away - or find it in the very last run of the loop anyway, when all the substitution is done, so it calls out for the simplification 😃 Unless i have missed something, for sure

soulus13:
Ah, and one more thing i have forgotten to mention - current logic of looking up the value in profiles part (`doc.at_xpath("/project/profiles/profile/properties/#{nm}")`) is generally prone to errors, as profiles could have multiple nodes with properties, each of which could have the specific property defined, and it could have different versions inside:
```
    <profiles>
        <profile>
            <id>profileOne</id>
            <properties>
                <property>
                    <spring.aop.version>1.0.1</spring.aop.version>
                </property>
            </properties>
        </profile>
        <profile>
            <id>profileTwo</id>
            <properties>
                <property>
                    <property>
                        <spring.aop.version>2.0.2</spring.aop.version>
                    </property>
                </property>
            </properties>
        </profile>
    </profiles>
```

Current implementation would only find `1.0.1` version and be done with that, whereas `2.0.2` version could be vulnerable as well, leaving the project vulnerable. Not sure how often that really is a case - but technically it could be.

PR #6583, opened 2023-02-03
Fix test failure on main

### LINKED ISSUE #6573: Unit test failure present on `main`

I can reproduce https://github.com/dependabot/dependabot-core/actions/runs/4069620621/jobs/7009563558 on `main`:
```ruby
Failures:

  1) Dependabot::NpmAndYarn::FileFetcher with .yarn data stored in git-lfs pulls files from lfs after cloning
     Failure/Error:
       expect(
         File.read(
           File.join(repo_contents_path, ".yarn", "releases", "yarn-3.2.4.cjs")
         )
       ).to start_with("#!/usr/bin/env node")

       expected "version [https://git-lfs.github.com/spec/v1\noid](https://git-lfs.github.com/spec/v1/noid) sha256:ca1a380d840c676152fbf8635a82d838113fe77e1665466699a4e106df4930d2\nsize 2194904\n" to start with "#!/usr/bin/env node"
     # ./spec/dependabot/npm_and_yarn/file_fetcher_spec.rb:85:in `block (3 levels) in <top (required)>'
     # /home/dependabot/common/spec/spec_helper.rb:63:in `block (2 levels) in <top (required)>'

Finished in 2 minutes 28.5 seconds (files took 0.95993 seconds to load)
303 examples, 1 failure

Failed examples:

rspec ./spec/dependabot/npm_and_yarn/file_fetcher_spec.rb:82 # Dependabot::NpmAndYarn::FileFetcher with .yarn data stored in git-lfs pulls files from lfs after cloning

Randomized with seed 50820
```

deivid-rodriguez:
Test that we delegate to `git lfs`, but not how lfs works

We're getting under the hood

```
batch response: This repository is over its data quota. Account responsible for LFS bandwidth should purchase more data packs to restore access.
error: failed to fetch some objects from 'https://github.com/dependabot-fixtures/dependabot-yarn-lfs-fixture.git/info/lfs'
```

so `git lfs` is not able to checkout the files, and the spec fails.

Closes #6573.

### DIFF
--- a/npm_and_yarn/spec/dependabot/npm_and_yarn/file_fetcher_spec.rb
+++ b/npm_and_yarn/spec/dependabot/npm_and_yarn/file_fetcher_spec.rb
@@ -79,21 +79,16 @@ RSpec.describe Dependabot::NpmAndYarn::FileFetcher do
       described_class.new(source: source, credentials: credentials, repo_contents_path: repo_contents_path)
     end
 
+    before do
+      allow(Dependabot::SharedHelpers).to receive(:run_shell_command).and_call_original
+    end
+
     it "pulls files from lfs after cloning" do
       # Calling #files triggers the clone
       expect(file_fetcher_instance.files.map(&:name)).to contain_exactly("package.json", "yarn.lock", ".yarnrc.yml")
-      expect(
-        File.read(
-          File.join(repo_contents_path, ".yarn", "releases", "yarn-3.2.4.cjs")
-        )
-      ).to start_with("#!/usr/bin/env node")
-
-      # LFS files not needed by dependabot are not pulled
-      expect(
-        File.read(
-          File.join(repo_contents_path, ".pnp.cjs")
-        )
-      ).to start_with("version https://git-lfs.github.com/spec/v1")
+      expect(Dependabot::SharedHelpers).to have_received(:run_shell_command).with(
+        %r{\Agit lfs pull --include .yarn,./yarn/cache}
+      )
     end
   end
 


deivid-rodriguez:
Regardless of this, the failure surfaces a new issue, since we're proceeding if the `git lfs` file cannot be checked out. We should probably detect the situation and raise a proper error.

jeffwidman (commented):
Thanks for digging into this. I'd started the other night, but it was 11PM so I gave up and went to bed. 🙃 

I'm of two minds on this route of fixing... I dislike mocking what we don't own (the `lfs` API), esp when the fixture is already setup. And we could easily get more quota, I doubt our unit test actually pulls that much data, we probably just happened to hit the quota because of CI, or if the quota is based on number of pulls rather than actual bytes transferred (more likely).

But I also realize this lets us get away from having a test depend on network I/O, which is nice. And the `lfs` API is unlikely to change so it's probably safe.

I know nothing about how LFS works under the covers, but I wonder how hard it'd be to make it point directly at a local fixture that we already ship? Or is the problem that we need a specific `git` repo external to `dependabot-core`?

Also, good catch that we could use another test when `lfs` fails... maybe file an issue for that so it's at least in our backlog?

deivid-rodriguez:
> I know nothing about how LFS works under the covers, but I wonder how hard it'd be to make it point directly at a local fixture that we already ship? Or is the problem that we need to point at a git repo external to dependabot-core?

Not sure if I understood this ideal well, like creating a fixture consisting of a git repository using lfs vendored as a submodule or something, and then somehow make Dependabot pull from it instead of from a real Github repo? If it's something like that, it feels overkill to me.

> Also, good catch that we could use another test when lfs fails... maybe file an issue for that so it's at least in our backlog?

Of course, I had not decided whether to fix it straight away or create an issue. I was planning to decided once this PR is completed.

jeffwidman:
I just checked, and we transferred 1.5 GB over the last 27 days. The plan resets in 3 days. That's higher than I'd have expected, until I remembered we were running double the CI suite for a while during the project to split docker by ecosystem, because we were running both monolith CI and per-ecosystem CI. Let me ask internally to see how hard it'd be to get our quota bumped up slightly.

jeffwidman:
Here's a discussion on handling errors in the original PR: https://github.com/dependabot/dependabot-core/pull/5833/files#r1007040120

deivid-rodriguez:
The cases discussed there are a bit a different, because in this case, there's no exception raised at all. We'd need to look at checked out files and verify whether they look like "unchecked git lfs files", or alternatively parse stderr and look for this particular error.

npm_and_yarn/spec/dependabot/npm_and_yarn/file_fetcher_spec.rb
```diff
@@ -79,21 +79,16 @@
...
     it "pulls files from lfs after cloning" do
       # Calling #files triggers the clone
       expect(file_fetcher_instance.files.map(&:name)).to contain_exactly("package.json", "yarn.lock", ".yarnrc.yml")
-      expect(
-        File.read(
-          File.join(repo_contents_path, ".yarn", "releases", "yarn-3.2.4.cjs")
-        )
-      ).to start_with("#!/usr/bin/env node")
-
-      # LFS files not needed by dependabot are not pulled
```
jeffwidman, line 91: I'd kinda hate to lose this test, it's nice to ensure we are only getting what we want if someone refactors that call to `git lfs` in some way

deivid-rodriguez: If someone refactored the `git lfs` call and forgot to use `--include` with explicit arguments somehow, they'd break the new test too. But yeah, a realworld test always covers you better. I'm fine with anything that puts the spec back to green.

jeffwidman:
> I just checked, and we transferred 1.5 GB over the last 27 days. The plan resets in 3 days.
> 
> That's higher than I'd have expected, until I remembered we were running double the CI suite for a while during the project to split docker by ecosystem, because we were running both monolith CI and per-ecosystem CI. And many of those changes were at the repo script level, so they triggered the full CI run rather than filtering the `npm` test out... I suspect we don't get anywhere near that limit normally.
> 
> Let me ask internally to see how hard it'd be to get our quota bumped up slightly.

The GitHub billing team increased the monthly bandwidth allowance of our test fixture org from the default 1 GB per month to 250 GB. So spec is now green and we are unlikely to run into bandwidth problems on this for a _long_ time.

So I would _slightly_ lean toward not going this route of mocking, preferring to retain the integration test. This isn't something I feel strongly on, so if you really want to, I'm fine with merging it. Just let me know and I can approve if desired.

deivid-rodriguez:
I don't have a strong opinion either, so let's close this.

Thanks for moving forward the quota increase!

PR #6590, opened 2023-02-03
Parallelize tests

deivid-rodriguez:
This is just a WIP, but wanted to share the work since I've been intermittently working on this for the past week.

It all started while reading over git documentation looking for something else and finding about [GIT_CONFIG_GLOBAL](https://git-scm.com/docs/git-config#ENVIRONMENT). It occurred to me that we could leverage that feature to parallelize our specs.

We currently do parallelization in CI only by splitting tests and running each slice on different CI jobs.

This PR allows truly parallelizing things and easily run all tests for an ecosystem locally  by running `script/ci-test`. For the ecosystems I have tried, it shows 50% speed up over `bundle exec rspec`.

Unfortunately still needs more work and even when getting to green CI, we probably to run it fully a few times to ensure no random failures are still present.

But as I said, I wanted to share the work.

Fixes https://github.com/dependabot/dependabot-core/issues/4909.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,36 +18,25 @@ jobs:
       fail-fast: false
       matrix:
         suite:
-          - { path: bundler, name: bundler1, ci_node_total: 2, ci_node_index: 0, ecosystem: bundler }
-          - { path: bundler, name: bundler1, ci_node_total: 2, ci_node_index: 1, ecosystem: bundler }
-          - { path: bundler, name: bundler2, ci_node_total: 2, ci_node_index: 0, ecosystem: bundler }
-          - { path: bundler, name: bundler2, ci_node_total: 2, ci_node_index: 1, ecosystem: bundler }
+          - { path: bundler, name: bundler1, ecosystem: bundler }
+          - { path: bundler, name: bundler2, ecosystem: bundler }
           - { path: cargo, name: cargo, ecosystem: cargo }
           - { path: common, name: common, ecosystem: common}
-          - { path: composer, name: composer, ci_node_total: 2, ci_node_index: 0, ecosystem: composer }
-          - { path: composer, name: composer, ci_node_total: 2, ci_node_index: 1, ecosystem: composer }
+          - { path: composer, name: composer, ecosystem: composer }
           - { path: docker, name: docker, ecosystem: docker }
           - { path: elm, name: elm, ecosystem: elm }
           - { path: git_submodules, name: git_submodules, ecosystem: gitsubmodule }
           - { path: github_actions, name: github_actions, ecosystem: github-actions }
-          - { path: go_modules, name: go_module, ci_node_total: 2, ci_node_index: 0, ecosystem: gomod }
-          - { path: go_modules, name: go_module, ci_node_total: 2, ci_node_index: 1, ecosystem: gomod }
+          - { path: go_modules, name: go_module, ecosystem: gomod }
           - { path: gradle, name: gradle, ecosystem: gradle }
-          - { path: hex, name: hex, ci_node_total: 2, ci_node_index: 0, ecosystem: mix }
-          - { path: hex, name: hex, ci_node_total: 2, ci_node_index: 1, ecosystem: mix }
+          - { path: hex, name: hex, ecosystem: mix }
           - { path: maven, name: maven, ecosystem: maven }
-          - { path: npm_and_yarn, name: npm_and_yarn, ci_node_total: 3, ci_node_index: 0, ecosystem: npm }
-          - { path: npm_and_yarn, name: npm_and_yarn, ci_node_total: 3, ci_node_index: 1, ecosystem: npm }
-          - { path: npm_and_yarn, name: npm_and_yarn, ci_node_total: 3, ci_node_index: 2, ecosystem: npm }
+          - { path: npm_and_yarn, name: npm_and_yarn, ecosystem: npm }
           - { path: nuget, name: nuget, ecosystem: nuget }
-          - { path: pub, name: pub, ci_node_total: 2, ci_node_index: 0, ecosystem: pub }
-          - { path: pub, name: pub, ci_node_total: 2, ci_node_index: 1, ecosystem: pub }
-          - { path: python, name: python, ci_node_total: 2, ci_node_index: 0, ecosystem: pip }
-          - { path: python, name: python, ci_node_total: 2, ci_node_index: 1, ecosystem: pip }
-          - { path: python, name: python_slow, ci_node_total: 2, ci_node_index: 0, ecosystem: pip }
-          - { path: python, name: python_slow, ci_node_total: 2, ci_node_index: 1, ecosystem: pip }
-          - { path: swift, name: swift, ci_node_total: 2, ci_node_index: 0, ecosystem: swift }
-          - { path: swift, name: swift, ci_node_total: 2, ci_node_index: 1, ecosystem: swift }
+          - { path: pub, name: pub, ecosystem: pub }
+          - { path: python, name: python, ecosystem: pip }
+          - { path: python, name: python_slow, ecosystem: pip }
+          - { path: swift, name: swift, ecosystem: swift }
           - { path: terraform, name: terraform, ecosystem: terraform }
 
     steps:
@@ -212,8 +201,6 @@ jobs:
             --env "RAISE_ON_WARNINGS=true" \
             --env "DEPENDABOT_TEST_ACCESS_TOKEN=${{ secrets.GITHUB_TOKEN }}" \
             --env "SUITE_NAME=${{ matrix.suite.name }}" \
-            --env "CI_NODE_TOTAL=${{ matrix.suite.ci_node_total }}" \
-            --env "CI_NODE_INDEX=${{ matrix.suite.ci_node_index }}" \
             --rm ghcr.io/dependabot/dependabot-updater-${{ matrix.suite.ecosystem }} bash -c \
             "cd /home/dependabot/${{ matrix.suite.path }} && ./script/ci-test"
 
--- a/bundler/spec/spec_helper.rb
+++ b/bundler/spec/spec_helper.rb
@@ -69,8 +69,14 @@ RSpec.configure do |config|
   config.after do
     # Cleanup side effects from cloning git gems, so that they don't interfere
     # with other specs.
-    helper_path = Dependabot::Bundler::NativeHelpers.versioned_helper_path(PackageManagerHelper.bundler_version)
-    FileUtils.rm_rf File.join(helper_path, ".bundle", "bundler")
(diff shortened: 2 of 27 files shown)


deivid-rodriguez:
Unfortunately I found another issue other that safety of global git configuration.

Dependabot also change the pwd to a temporary directory in order to run native helpers, and that's also not thread safe because it changes the pwd for all threads. Normally Dependabot does this:

1. Create a temporary directory.
2. Change to this new temporary directory.
3. Write some dependency files on this directory.
4. Shell out to a native helper on this directory.
5. Change back to the original directory.

* 3. does not really need changing directories, you could use full paths and write without moving folder.
* 4. could leverage [the `:chdir` parameter to `Process.spawn`](https://rubyapi.org/3.2/o/process#method-c-spawn), which is thread-safe. 

If we implement ☝️, then we could stop changing the pwd back and forth and things would be thread-safe.

But this is non-trivial work. Still I'll try to loop back to this at some point!

jeffwidman:
Despite the heavier lift, I definitely think the value is worth the time investment here... you'd certainly get some ✨ for it. 😁

deivid-rodriguez:
This seems consistently green now 🎉

Individual jobs seem approximately as fast as before, but we have way less jobs now!

I suspect some issues could eventually come up in particular since there are still several `Dir.chdir` around. But tests seem happy and I don't think this affects production since we don't run anything in parallel there, so I think this could be tested!

If we go with #5030 first, then we can avoid installing git from a ppa, since Ubuntu 22.04 provides a modern enough git version for this to work.

I'll set this as ready in case someone want to start reviewing.

jakecoffman (approved):
Happy to get rid of the stashing of the git config in the shared helpers. Great work!

abdulapopoola:
Happy to see that the Ubuntu update will come in handy. @jeffwidman if we can get that PR merged, that would unlock a ton more value like this.

abdulapopoola (approved):

.github/workflows/ci.yml
```diff
@@ -18,36 +18,25 @@ jobs:
...
-          - { path: composer, name: composer, ci_node_total: 2, ci_node_index: 0, ecosystem: composer }
-          - { path: composer, name: composer, ci_node_total: 2, ci_node_index: 1, ecosystem: composer }
+          - { path: composer, name: composer, ecosystem: composer }
           - { path: docker, name: docker, ecosystem: docker }
           - { path: elm, name: elm, ecosystem: elm }
           - { path: git_submodules, name: git_submodules, ecosystem: gitsubmodule }
           - { path: github_actions, name: github_actions, ecosystem: github-actions }
-          - { path: go_modules, name: go_module, ci_node_total: 2, ci_node_index: 0, ecosystem: gomod }
-          - { path: go_modules, name: go_module, ci_node_total: 2, ci_node_index: 1, ecosystem: gomod }
+          - { path: go_modules, name: go_module, ecosystem: gomod }
```
jeffwidman, line 30: How many different names can we have for the same thing?! Not the fault of this PR! 😄

bundler/spec/spec_helper.rb
```diff
@@ -69,8 +69,9 @@ def suppress_output
   config.after do
     # Cleanup side effects from cloning git gems, so that they don't interfere
     # with other specs.
-    helper_path = Dependabot::Bundler::NativeHelpers.versioned_helper_path(PackageManagerHelper.bundler_version)
-    FileUtils.rm_rf File.join(helper_path, ".bundle", "bundler")
-    FileUtils.rm_rf File.join(helper_path, ".bundle", "cache", "bundler", "git")
+    helper_path = Dependabot::Utils::BUMP_TMP_DIR_PATH
+    base_path = File.join(helper_path, ".bundle", "ruby", RbConfig::CONFIG["ruby_version"])
```
jeffwidman, line 73: Is `helper_path` used anywhere else? If not, maybe collapse them?

I actually find the naming slightly confusing, I'd expect `helper_path` to be built from `base_path`, not the other way around... Maybe I misunderstand how it works though?

common/lib/dependabot/utils.rb
```diff
@@ -1,13 +1,14 @@
...
+require "tmpdir"
 require "set"
 
 # TODO: in due course, these "registries" should live in a wrapper gem, not
 #       dependabot-core.
 module Dependabot
   module Utils
     BUMP_TMP_FILE_PREFIX = "dependabot_"
-    BUMP_TMP_DIR_PATH = "tmp"
+    BUMP_TMP_DIR_PATH = File.expand_path(Dir::Tmpname.create("tmp", "tmp") { nil })
```
jeffwidman, line 11: The double `"tmp"` is confusing, any way you could rename it slightly to clarify how they're different? I see one is the directory to create the other in, but it's not clear from the code which is which, so I had to lookup the signature which is annoying...

npm_and_yarn/helpers/patches/npm++pacote+9.5.12.patch
```diff
@@ -0,0 +1,14 @@
...
+index 7642eb2..7bb3324 100644
+--- a/node_modules/npm/node_modules/pacote/lib/util/git.js
++++ b/node_modules/npm/node_modules/pacote/lib/util/git.js
+@@ -25,7 +25,8 @@ const GOOD_ENV_VARS = new Set([
+   'GIT_SSH',
+   'GIT_SSH_COMMAND',
+   'GIT_SSL_CAINFO',
+-  'GIT_SSL_NO_VERIFY'
++  'GIT_SSL_NO_VERIFY',
++  'GIT_CONFIG_GLOBAL'
```
jeffwidman, line 11: Really a shame upstream isn't more interested in this patch...

jeffwidman (approved):
Fantastic work.

Looking at the commit log, I see a bunch of different changes, and only some seem related to the scope of paralleliz'ing tests:
1. be great if you merge via merge and not squash
2. If we have to revert for any reason, we'd want to only revert specific commits, and not all of these
3. Speaking of, do you want to extract the unrelated commits into separate PR's to make it more clear / easier to revert/`git blame` spelunk? Up to you, and not worth it if it's hard to untangle, it just looks like several of them could be easily cherry-picked out.

bundler/spec/spec_helper.rb
deivid-rodriguez: I agree it's confusing, I only extracted `base_path` because of a RuboCop LineLength (my least preferred RuboCop rule 😆).

I'll collapse them and "fix" the offense in other way.

common/lib/dependabot/utils.rb
deivid-rodriguez: I can try, but I can't do any anything against `Dir::Tmpname.create` signature being not very clear...

jeffwidman: No big deal, I was just thinking if you're creating `"tmp"` within `"/tmp"` maybe name the one you're creating something else? But if that has other side effects then just ignore my comment and move on.

npm_and_yarn/helpers/patches/npm++pacote+9.5.12.patch
deivid-rodriguez: Actually I did not propose this patch at all.

Apparently pacote now uses https://www.npmjs.com/package/@npmcli/git which I don't think resets the environment anymore.

So this is just temporary until we drop NPM6 support, which is the only thing relying on this old pacote.

common/lib/dependabot/utils.rb
deivid-rodriguez: I'm not creating "tmp" inside "/tmp", I'm creating "tmp_<unique-identifier>" inside (relative to current dir) "tmp". Maybe not passing anything as dir prefix would make things less confusing? Or any other suggested prefix?

deivid-rodriguez:
Regarding commits and merge method. Yes, generally I only merge & squash single commit PRs. I think we agreed on this some time ago.

Regarding extracting separate PRs, ok I can do that for a few changes not strictly related.

bundler/spec/spec_helper.rb
deivid-rodriguez: Done!

common/lib/dependabot/utils.rb
deivid-rodriguez: Passing the first parameter empty seems to work and should be less confusing since there's no double "tmp".

deivid-rodriguez:
I decided that the only commit that makes sense to extract is the one about composer, so I extracted that to #7879.

deivid-rodriguez:
Thanks for the feedback @jeffwidman, not only because of the good suggestions, but also because I think it's good to let this PR go through a few amend and rebase rounds, in case some flaky failures surface. Nothing so far fortunately! 💪

deivid-rodriguez:
ℹ️ I will be editing required CI checks to match the new sets of jobs right before merging this PR. After that, other open PRs will need to rebase on top of (or merge) main to get their CI checks green again.

deivid-rodriguez:
Yes, I had that concern too but my empiric observation is that, for each ecosystem, the slowest job of the new set of jobs is not a lot slower than the slowest job of the previous set of jobs.

Take for example Python, our slowest ecosystem.

Spec times on this PR:

* 10m 41s for the standard run: https://github.com/dependabot/dependabot-core/actions/runs/5949449089/job/16135268116?pr=6590
* 5m 54s for the slow run: https://github.com/dependabot/dependabot-core/actions/runs/5949449089/job/16135268300?pr=6590.

On a recent main run:

* 10m 7s for the standard run (slice 1): https://github.com/dependabot/dependabot-core/actions/runs/5948989928/job/16133884985
* 4m 25s for the standard run (slice 2): https://github.com/dependabot/dependabot-core/actions/runs/5948989928/job/16133885229
* 5m 41s for the slow run (slice 1): https://github.com/dependabot/dependabot-core/actions/runs/5948989928/job/16133885484
* 3m 42s for the slow run (slice 2): https://github.com/dependabot/dependabot-core/actions/runs/5948989928/job/16133885741

deivid-rodriguez:
I think even if we saw some degradation, this is still worth doing, since particularly for CI runs that trigger a full CI, we'll probably stop seeing queued jobs waiting to be picked up by a runner, which does happen now sometimes.

Also, after this PR, there's more room for improvement since we can now pretty easily speed up jobs by building parallel groups of specs using previous runtimes instead of file size.

jurre:
@deivid-rodriguez interesting, well that addresses my concern then! 👍 We could also look into getting large runners set up for these

jeffwidman:
> EDIT: By using previous runtimes we'll also be able to get rid of the `python_slow` job since that's just a less reliable way of splitting jobs by previous runtimes.

Is there any followup cleanup that needs to be done here such as removing the `:slow` designation / `rspec` helper in the `python` ecosystem code?

deivid-rodriguez:
I created a follow up issue at #7889.

PR #6595, opened 2023-02-04
Skip invalid docker image references in helm charts

lyoung-confluent:
It is possible to include variables/references in a helm chart, ex:
```yaml
image:
  repository: foo/bar
  tag: "${bar.foo}"
```
Currently, the docker parser will extract this and construct the docker reference as the string `foo/bar:${bar.foo}`.

Because this reference contains invalid characters (`{`, `}`, `$`) for a docker tag, the regex match against `IMAGE_SPEC` will (correctly) result in nil:
```
lib/dependabot/docker/file_parser.rb:172:in `block (2 levels) in workfile_file_dependencies': undefined method `named_captures' for nil:NilClass (NoMethodError)
``` 
By adding a `next unless` check we can skip over references like this without crashing allowing the rest of the file/directory to be upgraded.

### DIFF
--- a/docker/lib/dependabot/docker/file_parser.rb
+++ b/docker/lib/dependabot/docker/file_parser.rb
@@ -168,7 +168,9 @@ module Dependabot
 
           images.each do |string|
             # TODO: Support Docker references and path references
-            details = string.match(IMAGE_SPEC).named_captures
+            match = string.match(IMAGE_SPEC)
+            next unless match
+            details = match.named_captures
             details["registry"] = nil if details["registry"] == "docker.io"
 
             version = version_from(details)


jeffwidman (commented):
Nice find!

Can you add a test?

markhallen:
Thank you for the contribution @lyoung-confluent. We really appreciate your effort in finding the exact cause of this problem.

It seems that a recent change addressed this issue and you can see the updated code here:

https://github.com/dependabot/dependabot-core/blob/b9ea83bcad97700d007e5e5b57609df0aee756c1/docker/lib/dependabot/docker/file_parser.rb#L104-L115

I'm going to close this PR. Please let me know if the problem is resolved for you or if you continue to have issues. Thanks

PR #6344, opened 2022-12-20
Allow optional single and double quotes in yaml versions/tags

mburumaxwell:
Fixes: #6313

PR #6596, opened 2023-02-04
Allow helm/k8s image tag to be double/single quoted

lyoung-confluent:
Similar to #6344, allow the `image: "..."` YAML tag to be quoted.


jeffwidman (commented):
Makes sense, but can you add a test to prevent future regressions?

Should be trivial to extend the tests added in #6344 with an additional test case.

abdulapopoola:
@lyoung-confluent , can you please add tests and resolve the conflicts?

PR #6597, opened 2023-02-04
Don't require whitespace before image tag in helm charts

lyoung-confluent:
A helm `values.yaml` can be a complex file with nested structures, ex:
```yaml
---
service_a:
  image: foo/bar:v1.2.3
````
However, a very simple values files could also contain `image` as a top level tag, ex:
```yaml
---
image: foo/bar:v1.2.3
```
Dependabot assumes there will be whitespace `\s+` before the `image:` tag, but this is not always true.


jeffwidman (approved):
Good catch, thanks!

Don't think this one needs a test.

PR #6599, opened 2023-02-05
Adjust parent pom fetching logic

soulus13:
Reverted the fix in #6557 
Updated the parent fetching logic:
 - added a check that we only try to fetch parent `pom.xml` when the `parent` node is defined in the file
 - added additional checks when we fetch parent file to validate its `groupId`, `artifactId` and `versionId` against the ones provided in its "child" `pom.xml`
 - additionally added the cases for the empty `groupId` in case the `groupId` of the parent `pom.xml` is inherited from its own parent
 - covered the logic with tests and amended the old tests which have demonstrated incorrect behavior but worked fine before the logic adjustment

### DIFF
--- a/maven/lib/dependabot/maven/file_fetcher.rb
+++ b/maven/lib/dependabot/maven/file_fetcher.rb
@@ -92,9 +92,18 @@ module Dependabot
       def recursively_fetch_relative_path_parents(pom, fetched_filenames:)
         path = parent_path_for_pom(pom)
 
-        return [] if fetched_filenames.include?(path) || path.start_with?("..")
+        return [] if fetched_filenames.include?(path)
+
+        full_path_parts =
+          [directory.gsub(%r{^/}, ""), path].reject(&:empty?).compact
+
+        full_path = Pathname.new(File.join(*full_path_parts)).
+                    cleanpath.to_path
+
+        return [] if full_path.start_with?("..")
 
         parent_pom = fetch_file_from_host(path)
+        parent_pom.support_file = true
 
         [
           parent_pom,
--- a/maven/lib/dependabot/maven/file_fetcher.rb
+++ b/maven/lib/dependabot/maven/file_fetcher.rb
@@ -92,18 +92,18 @@ module Dependabot
       def recursively_fetch_relative_path_parents(pom, fetched_filenames:)
         path = parent_path_for_pom(pom)
 
-        return [] if fetched_filenames.include?(path)
+        return [] if path.nil? || fetched_filenames.include?(path)
 
         full_path_parts =
           [directory.gsub(%r{^/}, ""), path].reject(&:empty?).compact
 
-        full_path = Pathname.new(File.join(*full_path_parts)).
-                    cleanpath.to_path
+        full_path = Pathname.new(File.join(*full_path_parts)).cleanpath.to_path
 
         return [] if full_path.start_with?("..")
 
         parent_pom = fetch_file_from_host(path)
-        parent_pom.support_file = true
+
+        return [] unless fetched_pom_is_parent(pom, parent_pom)
 
         [
           parent_pom,
@@ -120,6 +120,8 @@ module Dependabot
         doc = Nokogiri::XML(pom.content)
         doc.remove_namespaces!
 
+        return unless doc.at_xpath("/project/parent")
+
         relative_parent_path =
           doc.at_xpath("/project/parent/relativePath")&.content&.strip || ".."
 
@@ -130,6 +132,28 @@ module Dependabot
         ].compact.reject(&:empty?)
 
         Pathname.new(File.join(*name_parts)).cleanpath.to_path
+      end
+
+      def fetched_pom_is_parent(pom, parent_pom)
+        pom_doc = Nokogiri::XML(pom.content).remove_namespaces!
+        pom_artifact_id, pom_group_id, pom_version = fetch_pom_unique_ids(pom_doc, true)
+
+        parent_doc = Nokogiri::XML(parent_pom.content).remove_namespaces!
+        parent_artifact_id, parent_group_id, parent_version = fetch_pom_unique_ids(parent_doc, false)
+
+        if parent_group_id.nil?
+          [parent_artifact_id, parent_version] == [pom_artifact_id, pom_version]
+        else
+          [parent_group_id, parent_artifact_id, parent_version] == [pom_group_id, pom_artifact_id, pom_version]
+        end
+      end
+
+      def fetch_pom_unique_ids(doc, check_parent_node)
+        parent = check_parent_node ? "/parent" : ""
+        group_id = doc.at_xpath("/project#{parent}/groupId")&.content&.strip
+        artifact_id = doc.at_xpath("/project#{parent}/artifactId")&.content&.strip
+        version = doc.at_xpath("/project#{parent}/version")&.content&.strip
+        [artifact_id, group_id, version]
       end
     end
   end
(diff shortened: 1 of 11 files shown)


soulus13:
@deivid-rodriguez, feel free to look😃 I have reverted the PR #6557 , added some additional logic to properly fetch parent pom-s and covered all that with tests. Any suggestions on the formatting/styling/improving are welcome, as the fight with the Perceived complexity check might have made the resulting code a bit less readable

jeffwidman:
Thanks! We'll both probably try to take a look.

> Any suggestions on the formatting/styling/improving are welcome, as the fight with the Perceived complexity check might have made the resulting code a bit less readable

You can also disable these checks for a particular function / section of code if you think it results in less readable code, for example see here as well as the `enable` commands at the top of the preceding function:
https://github.com/dependabot/dependabot-core/blob/4b000ace1ebcbf2b812a85522da6b1ae13a57367/bin/dry-run.rb#L374-L377

Linters are here to be helpful, not something we need to militantly follow (at least not in this repo 😁).

maven/lib/dependabot/maven/file_fetcher.rb
```diff
@@ -122,6 +133,30 @@ def parent_path_for_pom(pom)
 
         Pathname.new(File.join(*name_parts)).cleanpath.to_path
       end
+
+      def fetched_pom_is_parent(pom, parent_pom)
+        pom_doc = Nokogiri::XML(pom.content)
+        pom_doc.remove_namespaces!
```
jeffwidman, line 139: nit: To keep the length of the method more skimmable, I'd probably chain these on a single line... 
```suggestion
        pom_doc = Nokogiri::XML(pom.content).remove_namespaces
```

jeffwidman (commented):
From a quick look this all seems reasonable to me, although I lack familiarity with Java so no idea if the logic changes are correct.

I'm really impressed you caught the problems in the tests esp since they were passing and not yelling at you, nice job!

Since Deivid has been working in this area I'll wait for him to approve, or get a second sanity check from someone else on the team who knows more about this ecosystem.

maven/lib/dependabot/maven/file_fetcher.rb
soulus13: Good call! This indeed looks better to me 😃

maven/lib/dependabot/maven/file_fetcher.rb
```diff
@@ -122,6 +133,28 @@ def parent_path_for_pom(pom)
 
         Pathname.new(File.join(*name_parts)).cleanpath.to_path
       end
+
+      def fetched_pom_is_parent(pom, parent_pom)
+        pom_doc = Nokogiri::XML(pom.content).remove_namespaces!
```
jeffwidman, line 138: Oh weird, `remove_namespaces!` actually returns something? I'm used to the exclamation point variant does an in-place transform and there's a similarly named method with no exclamation point that must be called when wanting to return something... Otherwise`pom_doc` ends up `nil`.

I'm on my phone so can't look up the docs, but wanted to comment so at least one of us remembers to double check... Although I assume this is covered by a test in at least some manner.

But I'm also not a ruby expert so may be misunderstanding something...

soulus13: That is the same reaction i had (but i'm surely not a ruby expert as well) 😃 And apparently there is no `remove_namespaces` method there as well. But then i have checked the sources for initila method here https://www.rubydoc.info/github/sparklemotion/nokogiri/Nokogiri%2FXML%2FDocument:remove_namespaces! and it indeed looks like it returns self in the end. And the tests would have indeed failed if it did not return the XML doc back.

deivid-rodriguez: This suprised me too!

maven/spec/dependabot/maven/file_fetcher_spec.rb
```diff
@@ -441,4 +441,161 @@
...
+    end
+
+    context "when asked to fetch a pom which local parent has different groupId" do
+      let(:directory) { "/parent_modules_project/different_group_id" }
+      it "fetches only this pom" do
+        expect(file_fetcher_instance.files.map(&:name)).to eq(%w(pom.xml))
+      end
+    end
+
+    context "when asked to fetch a pom which local parent has different parentId" do
```
deivid-rodriguez, line 594:
```suggestion
    context "when asked to fetch a pom which local parent has different artifactId" do
```

maven/spec/dependabot/maven/file_fetcher_spec.rb
```diff
@@ -441,4 +441,161 @@
...
+      let(:directory) { "/parent_modules_project/pom_with_implicit_parent_path" }
+      it "fetches only both this pom and its implicit parent" do
+        expect(file_fetcher_instance.files.count).to eq(2)
+        expect(file_fetcher_instance.files.map(&:name)).
+          to match_array(%w(pom.xml ../pom.xml))
+      end
+    end
+
+    context "when asked to fetch a pom which local parent has different versionId" do
+      let(:directory) { "/parent_modules_project/different_version_id" }
```
deivid-rodriguez, line 581: How does maven behave here? It seems like a bad place to be for a repo, right? In any case, good with not messing with this kind of mismatch 👍.

deivid-rodriguez: As a minor nit on top of this, I would it be `version` rather than `versionId` here?

maven/lib/dependabot/maven/file_fetcher.rb
```diff
@@ -122,6 +133,28 @@ def parent_path_for_pom(pom)
...
+        pom_artifact_id, pom_group_id, pom_version = fetch_pom_unique_ids(pom_doc, true)
+
+        parent_doc = Nokogiri::XML(parent_pom.content).remove_namespaces!
+        parent_artifact_id, parent_group_id, parent_version = fetch_pom_unique_ids(parent_doc, false)
+
+        if parent_group_id.nil?
+          [parent_artifact_id, parent_version] == [pom_artifact_id, pom_version]
```
deivid-rodriguez, line 145: If the parent does not have a `groupId` defined, but the child does, do we still want to fetch the pom? I guess we do because `groupId` is not compulsory and if only a child defined it, that'd mean no conflict?

deivid-rodriguez (commented):
This looks very good, thorughly tested, and easy to read. Awesome PR! Just left a few minor questions.

maven/spec/dependabot/maven/file_fetcher_spec.rb
soulus13: > How does maven behave here? It seems like a bad place to be for a repo, right? In any case, good with not messing with this kind of mismatch 👍.

Generally - it would throw a warning and then try to lookup in remote repositories - basically for any of these 3 cases when it fails to find anything 😃

maven/lib/dependabot/maven/file_fetcher.rb
soulus13: If the parent does not have a `groupId` defined - it would generally mean that it should have own `parent` node defined as well, and inherit the `groupId` from there. Otherwise it would be an incorrect parent `pom.xml` at all, in this case. 

Thus i have skipped here the check that "when there is no `groupId` on the parent - it must have own `parent`", just to simplify the things a bit  and not try to recursively pull all parents for this check, assuming we check the valid projects here on Github 😃

deivid-rodriguez: I see, sounds good, thanks for clarifying!

deivid-rodriguez (approved):
👏 👏

Issue #4445, opened 2021-11-23
Log when running `go mod tidy` fails


pete-woods:
**Package ecosystem**
Go mod
**Package manager version**
Go 1.17
**Language version**
Go 1.17
**Manifest location and content prior to update**

**dependabot.yml content**
```
version: 2

updates:
  - package-ecosystem: "gomod"
    directory: "/"
    schedule:
      interval: "weekly"
```
**Updated dependency**

**What you expected to see, versus what you actually saw**

**Native package manager behavior**

**Images of the diff or a link to the PR, issue or logs**

**🕹 Bonus points: Smallest manifest that reproduces the issue**

Somewhere between the 5th Oct and the 21st Oct, dependabot seems to have stopped running `go mod tidy` for PRs to our repos. I can't see anything significant that changed in our repos. It appears to be happening for both internal projects and public ones.

Last PR where it behaved properly (5th Oct): https://github.com/circleci/ex/pull/92
First PR where it stopped tidying (21st Oct):  https://github.com/circleci/ex/pull/114
Whole bunch of PRs where I had to manually go mod tidy:
- https://github.com/circleci/ex/pull/122
- https://github.com/circleci/ex/pull/126
- https://github.com/circleci/ex/pull/133
- https://github.com/circleci/ex/pull/134
- https://github.com/circleci/ex/pull/135

jurre:
Thanks for reporting this @pete-woods, my first guess is that dependabot encounters an error on these repo's when running go mod tidy, which we [currently ignore](https://github.com/dependabot/dependabot-core/blob/350703e2e6982401c732b007e6275b22ae0aed87/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb#L139-L149), let me see if that's the case and if I can find out what the errors are

pete-woods:
I'm not seeing any errors in the public-facing logs, FWIW: https://github.com/circleci/ex/network/updates/235635214

jurre:
> I'm not seeing any errors in the public-facing logs, FWIW: https://github.com/circleci/ex/network/updates/235635214

Yeah we currently swallow those entirely, it would probably be nice to at least log those when they occur

pete-woods:
I guess it must do (or at least use the latest Go), as we have successfully generated PRs while we were using Go 1.17 in the mod file.

jurre:
We currently run dependabot on go 1.17 fwiw

pete-woods:
[...]

PR #6521, opened 2023-01-25
Improve error when native helper gives empty output

deivid-rodriguez:
If native helper output is empty, an empty string is unparseable JSON and we end up getting an error about that, but no information about the underlying culprit.

In this case, show the native error output, which is likely to contain something useful.

#### Before

```
Dependabot::Bundler::UpdateChecker::VersionResolver#latest_resolvable_version_details with a gemspec and a Gemfile with an upper bound that is lower than the current req 
Failure/Error: raise Dependabot::DependencyFileNotEvaluatable, msg

Dependabot::DependencyFileNotEvaluatable:
  Error evaluating your dependency files:
# ./lib/dependabot/bundler/update_checker/shared_bundler_helpers.rb:82:in `handle_bundler_errors'
# ./lib/dependabot/bundler/update_checker/version_resolver.rb:119:in `rescue in fetch_latest_resolvable_version_details'
# ./lib/dependabot/bundler/update_checker/version_resolver.rb:75:in `fetch_latest_resolvable_version_details'
# ./lib/dependabot/bundler/update_checker/version_resolver.rb:44:in `latest_resolvable_version_details'
# ./spec/dependabot/bundler/update_checker/version_resolver_spec.rb:52:in `block (3 levels) in <top (required)>'
# ./spec/dependabot/bundler/update_checker/version_resolver_spec.rb:414:in `block (5 levels) in <top (required)>'
# ./spec/spec_helper.rb:50:in `block (2 levels) in <top (required)>'
# /home/dependabot/common/spec/spec_helper.rb:63:in `block (2 levels) in <top (required)>'
# ./.bundle/ruby/3.1.0/gems/webmock-3.18.1/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'
# ------------------
# --- Caused by: ---
# JSON::ParserError:
#   unexpected token at ''
#   ./.bundle/ruby/3.1.0/gems/json-2.6.3/lib/json/common.rb:216:in `parse'
```

#### After

```
Dependabot::Bundler::UpdateChecker::VersionResolver#latest_resolvable_version_details with a gemspec and a Gemfile with an upper bound that is lower than the current req 
Failure/Error: raise Dependabot::DependencyFileNotEvaluatable, msg

Dependabot::DependencyFileNotEvaluatable:
[...]

PR #6601, opened 2023-02-06
Log when `go mod tidy` fails

JohnTitor:
Fixes https://github.com/dependabot/dependabot-core/issues/4445

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
@@ -2,6 +2,7 @@
 
 require "dependabot/shared_helpers"
 require "dependabot/errors"
+require "dependabot/logger"
 require "dependabot/go_modules/file_updater"
 require "dependabot/go_modules/native_helpers"
 require "dependabot/go_modules/replace_stubber"
@@ -139,10 +140,11 @@ module Dependabot
           command = "go mod tidy -e"
 
           # we explicitly don't raise an error for 'go mod tidy' and silently
-          # continue here. `go mod tidy` shouldn't block updating versions
-          # because there are some edge cases where it's OK to fail (such as
-          # generated files not available yet to us).
-          Open3.capture3(environment, command)
+          # continue with an info log here. `go mod tidy` shouldn't block
+          # updating versions because there are some edge cases where it's OK to fail
+          # (such as generated files not available yet to us).
+          _, stderr, status = Open3.capture3(environment, command)
+          Dependabot.logger.info "Failed to `go mod tidy`: #{stderr}" unless status.success?
         end
 
         def run_go_vendor


go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -139,10 +140,11 @@ def run_go_mod_tidy
...
           # we explicitly don't raise an error for 'go mod tidy' and silently
-          # continue here. `go mod tidy` shouldn't block updating versions
-          # because there are some edge cases where it's OK to fail (such as
-          # generated files not available yet to us).
-          Open3.capture3(environment, command)
+          # continue with an info log here. `go mod tidy` shouldn't block
+          # updating versions because there are some edge cases where it's OK to fail
+          # (such as generated files not available yet to us).
+          _, stderr, status = Open3.capture3(environment, command)
+          Dependabot.logger.info "Failed to `go mod tidy`: #{stderr}" unless status.success?
```
jeffwidman, line 147: Unfortunately, we _probably_ can't log the `stderr` output for fear it might contain creds or leak PII data into our logs. I say _probably_ because I'm not 100% sure, give me a day or two to double-check internally as we are currently working through the same question for:
* https://github.com/dependabot/dependabot-core/pull/6521

iff we can't do this, then we'll need to do something like this... but again, wait to hear from me before changing it:
```suggestion
          Dependabot.logger.info "Unexpected error while running `go mod tidy`, continuing without tidying" unless status.success?
```

jeffwidman (commented):
Thanks for this!

I just kickstarted CI for you.

Give me a day or two to inquire about the security risk of logging the actual error.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
JohnTitor: IIUC the above message will not help much in debugging as `go mod tidy` can fail for a variety of reasons. Let's consider another way if it's sensitive to reveal.

jeffwidman: The fact that it failed at all is actually very useful, as historically the problem has typically been something in the underlying repo, not the changes that Dependabot made. So the user can simply try running `go mod tidy` locally against their repo and see what it spits out. But do agree that showing the full error would be even more useful. Give me a day or two to sync up internally on this.

jeffwidman: So the upshot was that it's fine to expose the error in the job logs. 

The concern that hasn't been resolved internally revolves around raising exceptions since they get propagated to a different system, with a different set of requirements on what can/can't be stored. But since this only logs and doesn't raise any error, it won't matter.

jeffwidman (approved):
Let's ship this! 🎉 

Thanks again.

JohnTitor:
Awesome, thanks for checking/reviewing, too! It'd be great if you could let me know when it's deployed :)

jeffwidman:
I'll probably deploy it tomorrow, Friday at the very latest.

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
yeikel: I am curious to hear why we picked `info` here when this is an error?

JohnTitor: Because it can happen when a user does nothing wrong.

yeikel: That does not justify the level, in my opinion, as it is still an error

JohnTitor: I guess I don't understand your argument, I'd recommend opening a new issue or submitting a PR directly. Note that commenting on a merged PR isn't a good practice as it easily gets stalled.

PR #6609, opened 2023-02-06
Refactor python version detection

pavera:
## Context
The 4 python package managers had very nearly copy/pasted code to detect the project python version and install the correct python version. This copy/pasted code had drifted over time in some package managers (pipenv and poetry had gained some additional error handling around unsupported python versions among others). This aims to clean up the copy/pasted code and provide a new mixin module that can hold cross package manager methods to increase code reuse going forward

## Approach
I implemented a new mixin module in `python/helpers.rb` and attempted to unify the behavior of detecting a project python version, verifying support for said python version in Dependabot, and getting that python version installed. Going forward this will hopefully simplify the task of changing our python installation strategy to possibly utilize a process similar to Actions "setup/python" method.

### DIFF
--- a/python/lib/dependabot/python/file_updater/pip_compile_file_updater.rb
+++ b/python/lib/dependabot/python/file_updater/pip_compile_file_updater.rb
@@ -547,14 +547,16 @@ module Dependabot
 
         def python_requirement_parser
           @python_requirement_parser ||=
-            FileParser::PythonRequirementParser.
-            new(dependency_files: dependency_files)
+            FileParser::PythonRequirementParser.new(
+              dependency_files: dependency_files
+            )
         end
 
         def language_version_manager
           @language_version_manager ||=
-            LanguageVersionManager.
-            new(python_requirement_parser: python_requirement_parser)
+            LanguageVersionManager.new(
+              python_requirement_parser: python_requirement_parser
+            )
         end
 
         def setup_files
--- a/python/lib/dependabot/python/file_updater/pip_compile_file_updater.rb
+++ b/python/lib/dependabot/python/file_updater/pip_compile_file_updater.rb
@@ -18,7 +18,7 @@ module Dependabot
     class FileUpdater
       # rubocop:disable Metrics/ClassLength
       class PipCompileFileUpdater
-      include Helpers
+        include Helpers
         require_relative "requirement_replacer"
         require_relative "requirement_file_updater"
         require_relative "setup_file_sanitizer"
--- a/python/lib/dependabot/python/file_updater/pip_compile_file_updater.rb
+++ b/python/lib/dependabot/python/file_updater/pip_compile_file_updater.rb
@@ -7,7 +7,7 @@ require "dependabot/python/file_fetcher"
 require "dependabot/python/file_parser/python_requirement_parser"
 require "dependabot/python/file_updater"
 require "dependabot/shared_helpers"
-require "dependabot/python/helpers"
+require "dependabot/python/language_version_manager"
 require "dependabot/python/native_helpers"
 require "dependabot/python/python_versions"
 require "dependabot/python/name_normaliser"
@@ -18,7 +18,6 @@ module Dependabot
     class FileUpdater
       # rubocop:disable Metrics/ClassLength
       class PipCompileFileUpdater
-        include Helpers
         require_relative "requirement_replacer"
         require_relative "requirement_file_updater"
         require_relative "setup_file_sanitizer"
@@ -67,7 +66,7 @@ module Dependabot
         def compile_new_requirement_files
           SharedHelpers.in_a_temporary_directory do
             write_updated_dependency_files
-            install_required_python
+            language_version_manager.install_required_python
 
             filenames_to_compile.each do |filename|
               # Shell out to pip-compile, generate a new set of requirements.
@@ -177,7 +176,7 @@ module Dependabot
 
         def run_pip_compile_command(command, allow_unsafe_shell_command: false, fingerprint:)
           run_command(
-            "pyenv local #{python_major_minor}",
+            "pyenv local #{language_version_manager.python_major_minor}",
             fingerprint: "pyenv local <python_major_minor>"
           )
 
@@ -211,7 +210,7 @@ module Dependabot
           end
 
           # Overwrite the .python-version with updated content
-          File.write(".python-version", python_major_minor)
+          File.write(".python-version", language_version_manager.python_major_minor)
 
           setup_files.each do |file|
             path = file.name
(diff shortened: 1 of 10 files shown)


python/lib/dependabot/python/file_updater/pipfile_file_updater.rb
```diff
@@ -196,6 +198,12 @@ def updated_generated_files
             SharedHelpers.in_a_temporary_directory do
               SharedHelpers.with_git_configured(credentials: credentials) do
                 write_temporary_dependency_files(prepared_pipfile_content)
+                # Initialize a git repo to appease pip-tools
+                begin
+                  run_command("git init") if setup_files.any?
+                rescue Dependabot::SharedHelpers::HelperSubprocessFailed
+                  nil
```
jeffwidman, line 205: Unrelated to this PR:
What happens if `git` isn't present? 

We've seen this over in https://github.com/dependabot/dependabot-core/pull/6409... something to keep in mind I guess for the future to ensure on GHES and cloud we have `git` available to `python` ecosystem... 

If an error in creating a git repo results in `pip-tools` updates completely breaking, then it'd be nice to at least have an `info` log for easier debugging... don't have to capture the full subprocess, but instead we could log saying "couldn't create git repo, all pip-tools jobs will fail"...

And we probably should just outright bail here, no point in continuing the job if we know `pip-tools` will fail.

Again, unrelated to the scope of this PR, so I can take a stab at this in a subsequent PR if it's an issue.

python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
```diff
@@ -541,7 +495,6 @@ def setup_cfg_files
           dependency_files.select { |f| f.name.end_with?("setup.cfg") }
         end
       end
-      # rubocop:enable Metrics/ClassLength
```
jeffwidman, line 544: 🧹

python/lib/dependabot/python/update_checker/pipenv_version_resolver.rb
```diff
@@ -78,6 +79,12 @@ def fetch_latest_resolvable_version_string(requirement:)
             SharedHelpers.in_a_temporary_directory do
               SharedHelpers.with_git_configured(credentials: credentials) do
                 write_temporary_dependency_files(updated_req: requirement)
+                # Initialize a git repo to appease pip-tools
```
jeffwidman, line 82: If we take action to handle git in my other comment, than that action needs to change here too...

jeffwidman (approved):
This looks fantastic to me. 👍 

I know some folks don't see a lot of value in this type of cleanup because it's not that huge LOC savings, but this is _so valuable_ for eliminating the chance of drift down the road between all the copy/pasted variants... well worth the time IMO.

Might want to wait for @deivid-rodriguez 's review in case he knows some Ruby idioms related to mixins that you and I are ignorant of.

deivid-rodriguez (approved):
Looks good to me!

My only feedback is the generally I prefer to extract meaningful classes if possible in this kind of refactoring, rather that generic helper mixins. I think it makes the code easier to read and test.

In this case, I think we could try extract a `Dependabot::Python::LanguageVersionManager` class that takes the `python_requirement_parser` as a parameter on instantiation.

But I'd be happy (if you agree) to try this as a follow up.

python/lib/dependabot/python/file_updater/pipfile_file_updater.rb
pavera: Yes, I think this probably deserves a subsequent PR. Currently we use git during the python ecosystem docker image build, so I think we're pretty well assured that it is available. However, this could change in the future and then we'd want an early failure to tell us we need to re-add git.

pavera:
> In this case, I think we could try extract a `Dependabot::Python::LanguageVersionManager` class that takes the `python_requirement_parser` as a parameter on instantiation.

I can have a go at that here, no sense incurring tech debt cause I didn't originally notice the common factor of `python_requirement_parser`

python/lib/dependabot/python/file_updater/pip_compile_file_updater.rb
```diff
@@ -545,50 +545,16 @@ def requirement_map
...
-          end
-        end
-
         def python_requirement_parser
           @python_requirement_parser ||=
-            FileParser::PythonRequirementParser.new(
-              dependency_files: dependency_files
-            )
+            FileParser::PythonRequirementParser.
+            new(dependency_files: dependency_files)
```
deivid-rodriguez, line 551: Incredibly minor nitpick. Is this style change intentional? I generally prefer the previous style, that's why I ask.

python/lib/dependabot/python/file_updater/pipfile_file_updater.rb
```diff
@@ -196,7 +197,13 @@ def updated_generated_files
...
               SharedHelpers.with_git_configured(credentials: credentials) do
                 write_temporary_dependency_files(prepared_pipfile_content)
-                install_required_python
+                # Initialize a git repo to appease pip-tools
+                begin
+                  run_command("git init") if setup_files.any?
+                rescue Dependabot::SharedHelpers::HelperSubprocessFailed
+                  nil
+                end
+                language_version_manager.install_required_python
```
deivid-rodriguez, line 206: Previously I think you had to inline this logic because mixin in the previous module would result in a duplicate method name? Now I think we can leave the previous helper method `install_required_python`, and just change `Helpers.install_required_python(python_version)` to `language_version_manager.install_required_python` inside that method?

python/lib/dependabot/python/update_checker/pipenv_version_resolver.rb
```diff
@@ -78,7 +78,13 @@ def fetch_latest_resolvable_version_string(requirement:)
...
               SharedHelpers.with_git_configured(credentials: credentials) do
                 write_temporary_dependency_files(updated_req: requirement)
-                install_required_python
+                # Initialize a git repo to appease pip-tools
+                begin
+                  run_command("git init") if setup_files.any?
+                rescue Dependabot::SharedHelpers::HelperSubprocessFailed
+                  nil
+                end
+                language_version_manager.install_required_python
```
deivid-rodriguez, line 87: Same thing about potentially restoring the previous `install_required_python` method to reduce diff size.

deivid-rodriguez (commented):
Just a couple minor nits, looks great in any case!

python/lib/dependabot/python/file_updater/pip_compile_file_updater.rb
pavera: I originally copied this method from `pip_version_resolver.rb` the 1 place it's formatted this way.. then copied it from the Helper back to everywhere else.. yay copy/paste and keeping things consistent :(

jeffwidman (approved):
I noticed four files changed in `updater_checker` dir, but only three in `file_updater`... 

So is `file_updater/pipfile_file_updater.rb` used for both `pip` and `pipenv`?

pavera:
> So is `file_updater/pipfile_file_updater.rb` used for both `pip` and `pipenv`?

No, plain pip is handled by `file_updater/requirement_file_updater.rb` which didn't have any references to the code in question, so no update needed.

python/lib/dependabot/python/file_updater/pipfile_file_updater.rb
jeffwidman: I opened https://github.com/dependabot/dependabot-core/issues/6617 to track this.

PR #6611, opened 2023-02-07
Fix broken `composer` unit tests

jeffwidman:
`monolog/monolog` released a new `2.9.x` version which broke these unit
tests: https://github.com/Seldaek/monolog/releases/tag/2.9.1

A couple of things going on here:
- the native helper calls out to Packagist, and since it's not Ruby,
  it can't be easily stubbed, so that's why it suddenly started picking
  up the 2.9.1 release.
- However, this illustrated that I actually introduced a bug in the
  tests back in https://github.com/dependabot/dependabot-core/pull/6334 / 34216d42d099c60d9d999664e4dcb956721e1c89
  because anything above `1.22.0.a` should have been ignored.
- The tests were still broken because there's been a 3.0 release that
  isn't resolvable because it requires PHP 8, but we're still running
  PHP 7, so it exposed the bug described in: 
  https://github.com/dependabot/dependabot-core/blob/3846a60c18074d6b52f6d05c33a554e6a1b6ff31/composer/lib/dependabot/composer/update_checker/version_resolver.rb#L219-L226
- So to workaround the bug, I bumped the top of the unit test ignored
  version up to `4.0`, which is unreleased. This will also prevent this
  test failing again when we bump to PHP 8.

So this fixes the unit test back to a range that includes the latest
release, and also back to the expected value.

### DIFF
--- a/composer/spec/dependabot/composer/update_checker_spec.rb
+++ b/composer/spec/dependabot/composer/update_checker_spec.rb
@@ -198,8 +198,8 @@ RSpec.describe Dependabot::Composer::UpdateChecker do
     end
 
     context "when the user is ignoring the latest version" do
-      let(:ignored_versions) { [">= 1.22.0.a, < 3.0"] }
-      it { is_expected.to eq(Gem::Version.new("2.8.0")) }
+      let(:ignored_versions) { [">= 1.22.0.a, < 4.0"] }
+      it { is_expected.to eq(Gem::Version.new("1.21.0")) }
     end
 
     context "without a lockfile" do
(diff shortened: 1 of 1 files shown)


composer/spec/dependabot/composer/update_checker_spec.rb
```diff
@@ -198,8 +198,8 @@
     end
 
     context "when the user is ignoring the latest version" do
-      let(:ignored_versions) { [">= 1.22.0.a, < 3.0"] }
-      it { is_expected.to eq(Gem::Version.new("2.8.0")) }
+      let(:ignored_versions) { [">= 1.22.0.a, < 4.0"] }
```
jeffwidman, line 201: This is weird... when this is at `< 3.0`, the test returns `Gem::Version.new("2.9.1")`

But when I bump it up to `< 4.0`, the test returns `Gem::Version.new("1.2.1")`

??!

jeffwidman:
To me the subtlety of this bug illustrates that it'd be _really_ nice if we can start working with upstream package managers to pass them the list of ignored versions so they can simply filter them before doing their resolving.

jurre (approved):

Issue #6579, opened 2023-02-03
Guidance for Building Docker Image


pinkfloydx33:
My apologies if this is not the correct place to ask this.... 

We run a customized version of the dependabot-script inside of our Azure Devops instance. The script requires the dependabot-core image. Due to some restrictions in AzDO, we need to alter the UID/GID in the dependabot image to match those of our Azure Devops build agent user. 

Up until last week we would clone this repo and then build `./Dockerfile`  passing `--build-arg USER_UID=1234 --build-arg USER_GID=1234`. We'd then publish the new image to our container registry and use that in the actual run of dependabot. 

However this file is now gone, so at the very least the file to build has changed. But.. I've reviewed the replacement file as well as the PR that introduced the changes and it is not clear to me if/how I can continue to build the image from source. I'm not a docker expert and the person who originally created this process is no longer with us. I'm struggling to figure out what the right thing to do is.

Any help/guidance once be appreciated.

jeffwidman:
👋 Sorry for the pain here!

We actually should document how to do this for others as well, so I'll give you some high-level instructions and let me know what else you need, and then I'll probably turn this into a help doc long term over in the https://github.com/dependabot/dependabot-script repo.

Instead of building the root Dockerfile, you'll want to build the dockerfile(s) for the ecosystems you want to support. Example: [`nuget/Dockerfile`](https://github.com/dependabot/dependabot-core/blob/main/nuget/Dockerfile)

And then use that. That should actually build much faster.

If you have multiple ecosystems that you need to support, you'll need to build multiple docker images, and run each separately against the different ecosystems... ie, the java image against Java manifests, the php image against composer manifests etc.

Also, you may find https://github.com/tinglesoftware/dependabot-azure-devops useful if you haven't seen it already, although I'm not sure if @mburumaxwell has had a chance yet to update it to support the new per-ecosystem docker image workflows.

pinkfloydx33:
So that's interesting. We actually used to build a "slim" version of the image which involved pre-processing the dockerfile, reading it line by line and detecting the ecosystems we didn't want, outputting the lines to keep into a new file and building that.

Having images dedicated per ecosystem definitely eases that a bit. Unfortunately I need several ecosystems (nuget, npm, go, docker). On its face I don't mind building four images, I'm just not sure its going to work with our dependabot setup, at least not without a refactoring of the job--but that's my problem :) 

[...]

PR #6614, opened 2023-02-07
fixes docker-dev-shell not using the cache due to wrong tag

jakecoffman:
The `script/_common` scripts have a small section where it translates certain ecosystem names (like `npm_and_yarn`) to their official image name (`npm` in this example). So if we hook it up here in the `docker-dev-shell` script it will then use the cache since the tag matches. 

Thanks to @pavera for pointing this out.

### DIFF
--- a/bin/docker-dev-shell
+++ b/bin/docker-dev-shell
@@ -1,6 +1,8 @@
 #!/bin/bash
 
 set -e
+cd "$(dirname "$0")/.."
+source script/_common
 
 HELP=false
 REBUILD=false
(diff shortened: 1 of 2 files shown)


bin/docker-dev-shell
```diff
@@ -43,26 +45,12 @@ build_image() {
   export BUILT_IMAGE=true
   echo "$(tput setaf 2)=> building image from Dockerfile$(tput sgr0)"
 
-  docker build \
-    --build-arg BUILDKIT_INLINE_CACHE=1 \
-    --build-arg "USER_UID=$(id -u)" \
-    --build-arg "USER_GID=$(id -g)" \
```
deivid-rodriguez, line 49: Are these ARGs no longer necessary? If not, should they be removed from the base image too?

deivid-rodriguez: Just recalled https://github.com/dependabot/dependabot-core/issues/6579. Not sure if that's the original motivation, but some people are (were 😅) using these parameters for the custom builds of our images.

jakecoffman: Sounds like we can't remove it then!

The original context of the UID/GID setting: https://github.com/dependabot/dependabot-core/pull/3398

> The ability for bin/docker-dev-shell --rebuild to synchronize the dependabot UID/GID for consistency across volume mounts is retained

Sounds like an issue of Docker volume mounting permissions, which I haven't had an issue with using `script/dependabot`. But due to #6579 I wonder if we should add it back via environment variables in `script/_common`.

deivid-rodriguez: I guess that'd be the nicest thing to do yeah.

jeffwidman: Yeah, env vars that the user can override to pick whatever they want seems better TBH than hardcoding `id -u` / `id -g`... they may want something other than that.

jakecoffman: How does it look now?

deivid-rodriguez: Looks good!

deivid-rodriguez (approved):

PR #6615, opened 2023-02-07
Fix upgrades when docker registry API does not provide digests

### LINKED ISSUE #6612: Dependabot doesn't recognize updated Docker images.

### Package ecosystem

Docker

### Package manager version

No response

### Language version

_No response_

/bundle.Dockerfile
/oauth.Dockerfile

### dependabot.yml content
```yaml
version: 2
updates:
    directory: "/"
    schedule:
      # Check for updates to GitHub Actions every weekday
      interval: daily
  - package-ecosystem: "docker"
    directory: "/"
    schedule:
      interval: daily
  - package-ecosystem: "gomod"
    directory: "/"
    schedule:
      interval: daily
```
### Updated dependency

from registry.access.redhat.com/ubi8/ubi-minimal:8.7-923.1669829893
to registry.access.redhat.com/ubi8/ubi-minimal:8.7-1049

Expect - registry.access.redhat.com/ubi8/ubi-minimal:8.7-1049
SAW- no updates

No response

No response

```
FROM registry.access.redhat.com/ubi8/ubi-minimal:8.7-923.1669829893 as spi-oauth
```

deivid-rodriguez:
Some Docker registry API implementations don't provide a "content digest" header. For example, `https://registry.access.redhat.com/v2`.

In this case, we would incorrectly miss updates because the digests for any two images would be considered equal (`nil == nil`), and that would trigger our logic to not update to latest since we assumed it was actually the same version, just a change in version precision.

This PR should fix the problem.

Closes #6612.


docker/lib/dependabot/docker/update_checker.rb
```diff
@@ -152,17 +156,21 @@ def fetch_latest_version(version)
         candidate_tags = sort_tags(candidate_tags, version)
 
         latest_tag = candidate_tags.last
+        return version unless latest_tag
```
mctofu, line 159: ⚡ Nice improvement to this section!

mctofu (approved):

docker/lib/dependabot/docker/update_checker.rb
deivid-rodriguez: Thank you! :)

jeffwidman:
Related discussion in:
* https://github.com/dependabot/dependabot-core/pull/6748

Redhat is planning to provide digests in the not too distant future... although other registries may still have this behavior.

Issue #5945, opened 2022-10-20
Infinite loop when resolving npm dependency 


leoamigood:
### Package ecosystem

npm

### What you expected to see, versus what you actually saw

Somehow when there is an infinite loop when dependabot tries to resolve [this repo](https://github.com/rodcisal/meteor-react-native-simple-boilerplate) dependencies.

Looks like it's stuck in a loop here:
https://github.com/dependabot/dependabot-core/blob/e7c93f9d89d191cfd75bf0fe4288c51d3010ae13/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb#L184

leoamigood:
Same issue with [this repo](https://github.com/zhwei820/rnNavExample)

jeffwidman:
We should try to repro this using the dry-run script against both those repos... if it doesn't reproduce, then close for the reasons outlined here:
* https://github.com/dependabot/dependabot-core/issues/5824#issuecomment-1302396988

deivid-rodriguez:
It does reproduce, will try to fix it at #6616.

PR #6616, opened 2023-02-07
Detect path dependencies falling outside of repository and error out

deivid-rodriguez:
Otherwise we hang infinitely.

Fixes https://github.com/dependabot/dependabot-core/issues/5945.

Needs tests and I haven't even run npm CI locally either, so just opening a WIP for now.

### DIFF
--- a/npm_and_yarn/spec/dependabot/npm_and_yarn/file_fetcher_spec.rb
+++ b/npm_and_yarn/spec/dependabot/npm_and_yarn/file_fetcher_spec.rb
@@ -241,6 +241,23 @@ RSpec.describe Dependabot::NpmAndYarn::FileFetcher do
           to raise_error(Dependabot::PathDependenciesNotReachable)
       end
     end
+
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
Spec added!

npm_and_yarn/spec/dependabot/npm_and_yarn/file_fetcher_spec.rb
```diff
@@ -241,6 +241,23 @@
           to raise_error(Dependabot::PathDependenciesNotReachable)
       end
     end
+
+    context "with a path dependency that is an absolute path" do
```
jeffwidman, line 245:
```suggestion
    context "with a path dependency that is a relative path pointing outside of the repo" do
```

I get that it's effectively an absolute path, but calling that felt slightly confusing...

jeffwidman (approved):
nice find!

PR #6621, opened 2023-02-08
Order lists in docker scripts lexicographically

jeffwidman:
I was working on another PR and kept getting annoyed that these lists were unordered, made it painful to track down what was already there vs what was missing.


honeyankit (approved):
Thank you this tedious work!

PR #1463, opened 2019-10-24
IAM role and AWS cred file support for Docker bumps

JimNero009:
The AWS SDK has built-in default behaviour for credential finding. Therefore, if we just don't force the client to accept a custom-made credentials object, it will execute down that chain to find credentials. This means that this change will allow for developers to get AWS access for Docker ECR bumping by providing credentials via:
1. Environment variables
2. .aws/credentials files
3. Instance roles (IAM)

In order to not break the existing flow, this is an option that has to be opted into -- though ideally, I think it should probably be the default. WDYT?

PR #6622, opened 2023-02-08
Enable AWS ECR client to look for creds in default fallback locations

### LINKED ISSUE #1009: AWS ECR credentials not fetched properly
Hi,

I just tried out the ECR support and found out that you expect hardcoded AWS credentials.
While that is a valid use case if you are running on-prem and need to access AWS resources, it is considered a bad practice when running on AWS.

https://github.com/dependabot/dependabot-core/blob/master/docker/lib/dependabot/docker/utils/credentials_finder.rb#L39

I'd suggest (to stay backwards compatible) that if neither username and password are set (or empty) the default SDK client should be used directly, so it tries to either use ENV variables or the instance role.
WIth that on AWS the need of storing static credentials goes away.

jeffwidman:
This enables the AWS ECR client to look for creds in default fallback locations if they are not explicitly provided.

NOTE: This will _only_ affect folks running `dependabot-core` by themselves outside of the Dependabot Service offered by GitHub.

According to the [AWS ECR SDK Client docs](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/ECR/Client.html#initialize-instance_method):
> When `:credentials` are not configured directly, the following locations will be searched for credentials:
> * `Aws.config[:credentials]`
> * The `:access_key_id`, `:secret_access_key`, and `:session_token` options.
> * `ENV['AWS_ACCESS_KEY_ID']`, `ENV['AWS_SECRET_ACCESS_KEY']`
> * `~/.aws/credentials`
> * `~/.aws/config`

The flow here of picking what creds to prioritize is important to get correct. We have to cover all the following scenarios:
  1. users on Dependabot Service at GitHub... our proxy will manage the creds, so `dependabot-core` never sees them
  2. users who are hardcoding creds for `dependabot-script`
  3. users who have creds in `~/.aws/`
  4. users accessing public ECR that doesn't require creds

For 1️⃣ , it'll be the same as 4️⃣ , and happens in a context where 2️⃣ / 3️⃣ won't be populated, so should be straightforward as long as there's not a big perf hit for the ECR SDK library to check the filesystem for 3️⃣.

99% of traffic through this library probably happens through the Dependabot Service on GitHub, so 99% of the time the checks will be pointless... but the naive flow is far more intuitive, so if we do need a flag for perf reasons, we should make it opt-out of credential checking rather than opt-in. And actually we'd be able to re-use this across the `dependabot-core` library anywhere we use creds.

This is an edited version of @JimNero009's original work over in https://github.com/dependabot/dependabot-core/pull/1463, and includes him as a co-author on the commit.

Fix #1009

### DIFF
--- a/bin/docker-dev-shell
+++ b/bin/docker-dev-shell
@@ -107,6 +107,7 @@ CODE_DIR="/home/dependabot/dependabot-core"
 touch .core-bash_history
 mkdir -p dry-run tmp
 docker run --rm -ti \
+  -v "$HOME/.aws:$CODE_DIR/../.aws" \
   -v "$(pwd)/.core-bash_history:/home/dependabot/.bash_history" \
   -v "$(pwd)/.rubocop.yml:$CODE_DIR/.rubocop.yml" \
   -v "$(pwd)/bin:$CODE_DIR/bin" \
--- a/docker/lib/dependabot/docker/utils/credentials_finder.rb
+++ b/docker/lib/dependabot/docker/utils/credentials_finder.rb
@@ -47,29 +47,33 @@ module Dependabot
           # If credentials have been generated from AWS we can just return them
           return registry_details if registry_details["username"] == "AWS"
 
-          # If we don't have credentials, we might get them from the proxy
-          return registry_details if registry_details["username"].nil?
-
-          # Otherwise, we need to use the provided Access Key ID and secret to
-          # generate a temporary username and password
+          # Build a client either with explicit creds or default creds
+          registry_hostname = registry_details.fetch("registry")
+          region = registry_hostname.match(AWS_ECR_URL).named_captures.fetch("region")
           aws_credentials = Aws::Credentials.new(
             registry_details["username"],
             registry_details["password"]
           )
 
-          registry_hostname = registry_details.fetch("registry")
-          region = registry_hostname.match(AWS_ECR_URL).
-                   named_captures.fetch("region")
+          ecr_client =
+            if aws_credentials.set?
+              Aws::ECR::Client.new(region: region, credentials: aws_credentials)
+            else
+              # Let the client check default locations for credentials
+              Aws::ECR::Client.new(region: region)
+            end
 
+          # If the client still lacks credentials, we might be running within GitHub's
+          # Dependabot Service, in which case we might get them from the proxy
+          return registry_details if ecr_client.config.credentials.nil?
+
+          # Otherwise, we need to use the provided Access Key ID and secret to
+          # generate a temporary username and password
           @authorization_tokens ||= {}
           @authorization_tokens[registry_hostname] ||=
-            Aws::ECR::Client.new(region: region, credentials: aws_credentials).
-            get_authorization_token.authorization_data.first.
-            authorization_token
-
+            ecr_client.get_authorization_token.authorization_data.first.authorization_token
           username, password =
             Base64.decode64(@authorization_tokens[registry_hostname]).split(":")
-
           registry_details.merge("username" => username, "password" => password)
         rescue Aws::Errors::MissingCredentialsError,
                Aws::ECR::Errors::UnrecognizedClientException,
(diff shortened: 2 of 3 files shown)


bin/docker-dev-shell
```diff
@@ -115,6 +115,7 @@ CODE_DIR="/home/dependabot/dependabot-core"
 touch .core-bash_history
 mkdir -p dry-run tmp
 docker run --rm -ti \
+  -v "$HOME/.aws:$CODE_DIR/../.aws" \
```
jeffwidman, line 118: We'll eventually need to ensure this is plumbed through over in `dependabot-script` as well if it's not already, but we can worry about that later...

docker/lib/dependabot/docker/utils/credentials_finder.rb
```diff
@@ -47,29 +47,33 @@ def build_aws_credentials(registry_details)
...
 
-          registry_hostname = registry_details.fetch("registry")
-          region = registry_hostname.match(AWS_ECR_URL).
-                   named_captures.fetch("region")
+          ecr_client =
+            if aws_credentials.set?
+              Aws::ECR::Client.new(region: region, credentials: aws_credentials)
+            else
+              # Let the client check default locations for credentials
+              Aws::ECR::Client.new(region: region)
```
jeffwidman, line 63: This is where we take a small perf hit in the Dependabot-on-GitHub service since it looks for config files in a number of places that will never exist... but I think those calls should be quite quick such that it doesn't really matter?

jeffwidman (commented):
Annotating changes

docker/lib/dependabot/docker/utils/credentials_finder.rb
brrygrdn: I think this is fine, I guess the only mild risk is that it encounters a file-system related issue trying to access a certain path. It does make me think we should have some kind of environment flag that explicitly tells code we are running in the service and to skip certain behaviours like this.

I don't think of that as a performance requirement so much as reducing logic a job could 'snag' on due to unforeseen circumstances ( e.g. for some weird reason we end up needing to have .aws dotfiles in our image, or they get added by accident, etc, etc ).

jeffwidman: > I don't think of that as a performance requirement so much as reducing logic a job could 'snag' on due to unforeseen circumstances ( e.g. for some weird reason we end up needing to have .aws dotfiles in our image, or they get added by accident, etc, etc ).

That's a good point.

I'd prefer to ship this as-is for now as I don't expect any breakage and we can test-deploy before merge. And then have a follow-up ticket to add that flag... Also because I'm interested in doing a DRY audit across updater, dry-run, dependabot-script etc to tease out the common bits so adding that flag / env var would be easier to handle as part of that audit/re-think.

mctofu (approved):

bin/docker-dev-shell
deivid-rodriguez: I think `bin/docker-dev-shell` should behave as closely as possible as "the service", and this will introduce non-deterministic behavior in the dev container, depending on whether the developer has AWS credentials configured or not, right? I feel this could lead to confusing behavior/weird spec failures, on certain developer machines, which defeats the point of a developer container a bit. Is this necessary for the feature implemented by this PR?

jeffwidman: Hmm... good point. 

It's not necessary for `docker-dev-shell`, and you're right that's currently primarily intended for dev work, not prod, so let me pull this out.

I do think we should still plumb it through for `dependabot-script` style stuff where it's intended to be run on prod.

But anyone using `docker-dev-shell` as the entrypoint for prod is doing it wrong IMO.

jeffwidman: https://github.com/dependabot/dependabot-core/pull/6633

deivid-rodriguez: Agreed on plumbing this for `dependabot-script`, yeah!

Issue #3597, opened 2021-04-29
Don't force the use of a .github file for this feature


ucirello:
![image](https://user-images.githubusercontent.com/138827/116576893-d70de300-a8c4-11eb-9ad5-698faff9a68e.png)

My issue was closed without due consideration at https://github.com/dependabot/dependabot-core/issues/3596#event-4663513506

@jurre linked to topics regarding the automerge and completely ignored the other topics, the one related about the need of having a github file in my repository. 

Please restore the behavior in which was possible to configure the tool through an interface and not a file.

asciimike:
Admittedly, the decision to use a `dependabot.yml` predates me, but I think that config files have a few concrete advantages over just configuration UI:
 - it's versioned, so it's easy to roll back to previous versions if a change broke something
 - related, that history is effectively an audit log (you can tell who made the change when)
 - it's tied to the auth settings for the repo (must be a writer, potentially must be in `CODEOWNERS`)

Is the desire a UI for easier configuration (basically, if we skinned this file with a UI is that sufficient), or do you truly not want a file at all (if so, why not), or some other reason?

ucirello:
@asciimike -- thanks for enumerating all these non-advantages. They are fantasy advantages that makes total sense in the out-of-touch product managers. If you believe that the only way to manage this work is inside the git repository itself, then you may have a deeper problem than you realize.

> do you truly not want a file at all (if so, why not), or some other reason?

I truly don't want a file at all. I want a host-agnostic git repository. I am already very annoyed by the fact that I need to keep CD/CI files (as a matter of fact, I am preparing to offboard from Travis CI and run it elsewhere, in a place where the configuration sits OUT of the repository).

I keep mirrors of my repositories in many places. Essentially you are forcing me to contaminate my mirrors with Github specific files. But we all know why, right? It's the Microsoftization of Github and you have to make sure that we, the consumers, are tied to you and dependent on you as much as possible.

In any case, I decided to strip dependabot from my repositories. With that said, feel free to close this issue as I am not going to be using dependabot anymore.

asciimike:
I think that there is an interesting use case here around vendor agnostic dependency updates, and that's definitely a supported use case for running your own version of dependabot-core. 

That said, the tone used here is unacceptable in an open-source project, hence why this thread has been locked. We welcome feature requests and even disagreement with our decisions, but we do not welcome it in this way.

PR #6626, opened 2023-02-08
Update README and helper scripts for the updater

brrygrdn:
The Updater maintains its own README which needs some tweaks now we have leaned out our Docker images 🎉 

As part of this, I've created a `script/updater/build` helper which abstracts away the need to know that the Updater builds against the Bundler base image.

I've also moved `script/ci-test-updater` into `updater/script/test` as the test requirements for CI and developers are functionally identical.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -181,11 +181,11 @@ jobs:
         uses: actions/checkout@v3
       # using bundler as the test updater
       - name: Build ecosystem image
-        run: script/build bundler
+        run: updater/script/build
       - name: Run updater tests
         env:
           DEPENDABOT_TEST_ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: script/ci-test-updater
+        run: updater/script/test
 
   lint:
     name: Lint
--- a/updater/README.md
+++ b/updater/README.md
@@ -11,30 +11,33 @@ GitHub network, and so is not generally accessible.
 
 ## Setup
 
-You will need to provide the build a Personal Access Token to access the GitHub Package Registry to retrieve
-dependency containers.
+To work with the updater, you will need to build the bundler image using the
+build script in this directory:
 
-[Create a token](https://github.com/settings/tokens/new) with the `packages:read` scope and set it in your environment
-as `GPR_TOKEN`
-
-Run the setup script:
-
-```
-script/setup
+```bash
+script/build
 ```
 
 ## Tests
 
 We run [rspec](https://rspec.info/) tests inside a Docker container for this project:
 
-```
+```bash
 script/test
 ```
 
 You can run an individual test file like so:
 
-```
+```bash
 script/test spec/dependabot/integration_spec.rb
+```
+
+A small number of tests hit the GitHub API, so you will need to set the envvar
+`DEPENDABOT_TEST_ACCESS_TOKEN` with a Personal Access Token with the full `repo`
+scope.
+
+```bash
+export DEPENDABOT_TEST_ACCESS_TOKEN=ghp_xxx
 ```
 
 ### VCR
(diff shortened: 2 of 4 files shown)


jakecoffman (approved):
Looks good, thanks for cleaning this up!

updater/script/test
```diff
@@ -16,4 +16,4 @@ docker run --env DEPENDABOT_TEST_ACCESS_TOKEN \
   -v "$(pwd)/npm_and_yarn:/home/dependabot/npm_and_yarn" \
   -v "$(pwd)/go_modules:/home/dependabot/go_modules" \
   -v "$(pwd)/composer:/home/dependabot/composer" \
-  "ghcr.io/dependabot/dependabot-updater-bundler:latest" bundle exec rspec
+  "ghcr.io/dependabot/dependabot-updater-bundler:latest" bundle exec rspec "${@}"
```
deivid-rodriguez, line 19: We can also run specific updater specs easily now :)

deivid-rodriguez (approved):
Nice improvements!

.github/workflows/ci.yml
```diff
@@ -181,11 +181,11 @@ jobs:
         uses: actions/checkout@v3
       # using bundler as the test updater
```
jeffwidman, line 182: this comment should be deleted now that the implementation detail is buried within `updater/script/build`

updater/script/build
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+set -e
+cd "$(dirname "$0")/../.."
+# For simplicity, we just use the bundler base image for the updater environment
+script/build bundler
```
jeffwidman, line 6: I personally am 🤷‍♂️ on whether this adds value... I get that it encapsulates the "use the bundler image as a stand-in" hack, but OTOH, I often find the additional layer of indirection from small scripts like this add more mental overhead than value... and here if I'm looking at this, it's probably because I'm trying to get into the docker dev shell to debug something in the updater, so it's unclear to someone new to this repo how this plays with that.

updater/README.md
```diff
@@ -11,32 +11,35 @@ GitHub network, and so is not generally accessible.
...
-```
+```bash
 script/test
 ```
 
 You can run an individual test file like so:
 
-```
+```bash
 script/test spec/dependabot/integration_spec.rb
```
jeffwidman, line 32: I'm not sure I'm in love with going this route, since for failures the `rspec` output spits out
```
rspec <filepath>:<line>
```

so for our drive-by contributors who are unfamiliar with `rspec` or even ruby in general (that was me a year ago!) seeing this in the docs leaves me confused on how do I run that particular test because I'm supposed to run `script/test`, not `rspec`... resulting in "??!" followed by half an hour poking through scripts to see how things are wired up.

I've also found it a little slow when each test run involves spinning up `docker run`...

For all the ecosystem-specific tests, we've gone the route of recommending dropping into the `docker-dev-shell`, then `cd <folder> && rspec` because then it's easy to see how to pass flags directly to `rspec`. Ideally we could continue that pattern for `updater`, even though it's not technically an ecosystem, as that's what our drive-by contributors will be used to.

Is there a reason we can't go that route? The requesite folders are already mounted in the `docker-dev-shell`:
* https://github.com/dependabot/dependabot-core/pull/6569

updater/script/test
```diff
@@ -1,13 +1,13 @@
 #!/bin/bash
```
jeffwidman, line 1: If we went the route of suggesting devs run tests within the `docker-dev-shell`, then maybe this should be renamed to `updater/script/ci-test` to be consistent with the scripts used in the ecosystem folders?

jeffwidman (commented):
I had some questions about going this route... we def need to fix the broken docs, but everywhere else we recommend developing using the `docker-dev-shell` and it'd be ideal to keep that development API consistent even for the `updater` as it ensures a consistent dev experience for our many drive-by contributors.

I'll reach out to you on slack for a quick call to chat about some options here.

jeffwidman:
Also, to be clear, really appreciate you working on this! It def is broken and needs fixing...

updater/README.md
deivid-rodriguez: You got a point @jeffwidman, I like the idea of keeping the `docker-dev-shell` pattern.

brrygrdn: That's a fair point, I think I approached this as a 'fix what is broken' based on my muscle memory of working on updater when it lived separately.

I'll try that workflow today and see how I get on and if there are any caveats for the updater.

brrygrdn:
I discussed this with @jeffwidman and I agree we should be trying to avoid having multiple entry points for contributors, I ran into some issues bundling the `updater/` project on the `docker-dev-shell` and have set it aside for now to finish my WIP, but I'm going to hold on merging this until we can look at that as a preferred alternative.

jeffwidman:
I'm going to mark as a draft while I try to get the `updater` working from within `docker-dev-shell`... then will tweak this slightly... def want to pull in most of this, just with some slight modifications.

brrygrdn:
#3597 is the preferred path to make the updater dev workflow align with other projects

PR #6628, opened 2023-02-09
Don't raise all versions ignored error if digest can be updated

### LINKED ISSUE #1971: [Docker] stay on specific tag and update digest only
```
FROM registry:2@sha256:3a8eef8d0a818b9bbb4bd17667253473e2d99935ccbbd37649af6bcaa064cf0d
```

I think that dependabot will upgrade base image tag for me once there is `registry:3` available.

I want to stay at registry:2 and update digest only. Is there any configuration available for this case?

deivid-rodriguez:
User may have all three ignores `["version-update:semver-major", "version-update:semver-minor", "version-update:semver-patch"]` configured, but will still want digest updates if pinned to specific digests.

This PR fixes that.

Closes #1971.

### DIFF
--- a/docker/lib/dependabot/docker/update_checker.rb
+++ b/docker/lib/dependabot/docker/update_checker.rb
@@ -404,7 +404,10 @@ module Dependabot
             version = comparable_version_from(tag)
             ignore_requirements.any? { |r| r.satisfied_by?(version) }
           end
-        if @raise_on_ignored && filter_lower_versions(filtered).empty? && filter_lower_versions(candidate_tags).any?
+        if @raise_on_ignored &&
+           filter_lower_versions(filtered).empty? &&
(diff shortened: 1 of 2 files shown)


jeffwidman (approved):

yob:
Thanks!

How long does it take for dependabot changes to reach GitHub? Once this has shipped I'm happy to test it and report back.

jeffwidman:
Thanks for the reminder, I meant to deploy this morning and got busy with other things. I just deployed, so it's live now.

arnested:
Very cool and useful!

I have a project where we have a multistage build that uses the same image in three different versions, and where I also prefer to only update the digests.

With this merged, I can get updates for the first of them but not the remaining two.

github com_arnested_dep-php-test_pull_1_files

PR #6582, opened 2023-02-03
Pass on setup docs for contributing

deivid-rodriguez:
Not mentioning `bin/dry-run.rb` or the CLI yet, just focusing on removing out of date information and replacing it with how you get setup to run specs for a specific ecosystem.

Issue #6577, opened 2023-02-02
It's confusing when to use `go_modules` vs `gomod`


jeffwidman:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code improvement description

`gomod` is a leftover from when `go` didn't have a cannonical package manager, so there was the `go dep` project and several others.

We've also now got `go_modules` and it's confusing which to use where... for example:
* https://github.com/dependabot/dependabot-core/pull/6575

We should standardize on a cannonical name for go modules. We could use `go`, since I doubt there will be another package manager in that ecosystem, but that also feels risky in case I'm wrong... so `go_modules` is probably the safest solution.

So this is a reminder to me to investigate what it'd take to standardize on that.

markdorison:
This is not just true for Go, but also for a number of other ecosystems. The example below from one of this repo's workflows shows the following items differ:

<table>
<tr>
	<td>NAME</td>
	<td>ECOSYSTEM</td>
<tr>
	<td>git_submodules </td>
	<td>gitsubmodule </td>
<tr>
	<td>github_actions</td>
    <td>github-actions</td>
<tr>
	<td>go_modules</td>
    <td>gomod</td>
<tr>
	<td>hex</td>
    <td>mix</td>
<tr>
    <td>npm_and_yarn</td>
    <td>npm</td>
<tr>
    <td>python</td>
    <td>pip</td>
</table>

https://github.com/dependabot/dependabot-core/blob/82906a986a5d09e31e27e038618f49d1cba940cf/.github/workflows/images-latest.yml#L36-L52

Should this issue cover all of these?

krokofant:
This was especially confusing when wanting to maintain the ecosystem-label when there is only one ecosystem then dependabot only adds the "dependency" label, but for easier search I'd like to have the ecosystem-label everywhere. It was a bit hard to find which label is actually added.

Via this issue I know that I should add the label "github_actions" instead of the ecosystem name "github-actions".

PR #6657, opened 2023-02-13
update readme to include new ecosystem image info

jakecoffman:
We've broken the large updater image out into smaller images and thus some of the scripts have changed. 

Here's a first pass at updating the README to capture those changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,18 @@ $ bin/docker-dev-shell go_modules
 [dependabot-core-dev] ~/dependabot-core $
 ```
 
+The ecosystem parameter is the top-level directory name of the ecosystem in this project.
+
+If you're running the amd64 architecture, you have the option to pull the image instead
+of building it directly:
+
+```shell
+$ docker pull ghcr.io/dependabot/dependabot-updater-gomod
+```
+
+The updater image name uses the [YAML name](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem)
+to specify the ecosystem.
+
 ### Dry run script
 
 You can use the "dry-run" script to simulate a dependency update job, printing
--- a/README.md
+++ b/README.md
@@ -80,8 +80,8 @@ $ bin/docker-dev-shell go_modules
 
 The ecosystem parameter is the top-level directory name of the ecosystem in this project.
 
-If you're running the amd64 architecture, you have the option to pull the image instead
-of building it directly:
+If you're running the amd64 architecture, you have the option to pull the updater image instead
+of building it from scratch:
 
 ```shell
 $ docker pull ghcr.io/dependabot/dependabot-updater-gomod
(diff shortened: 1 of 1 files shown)


honeyankit (approved):

honeyankit:
Thank you for the doc!

README.md
```diff
@@ -70,12 +70,10 @@ best way to work with the project.
 
 ## Running with Docker
 
-Start by pulling the developer image from the [GitHub Container Registry][ghcr-core-dev] and then start the developer shell:
+Start by running the developer shell:
 
 ```shell
-$ docker pull ghcr.io/dependabot/dependabot-core-development:latest
-$ docker tag ghcr.io/dependabot/dependabot-core-development dependabot/dependabot-core-development
```
jeffwidman, line 77: I'm very hesitant to remove these, even with the updater broken apart, it is still a _significant_ speedup to pull these rather than building locally, and most folks reading the Readme aren't going to be working in a codespace.

Why can't we instead recommend they pull `docker pull ghcr.io/dependabot/dependabot-core-development-<ecosystem>:latest`??

jakecoffman: We aren't currently publishing development images since they don't add much on top of the updater image they're built on, which is why I removed this.

We could suggest pulling the ecosystem image `ghcr.io/dependabot/dependabot-<ecosystem>` but... this README was written in simpler times when we all were running the same architecture. I think until we get an arm64 image published it would be unwise for folks running arm64 to run the amd64 image as emulation is quite slow. 

I could mention that caveat here.

abdulapopoola: I ran the existing commands and I got an error:

`invalid argument "ghcr.io/dependabot/dependabot-updater-NPM" for "-t, --tag flag: invalid reference format: repository name must be lowercase`. Had errors with lowercase npm and yarn too. 

Trying go_modules now; these might be newbie mistakes so open to hearing more thougthts.

jakecoffman: @abdulapopoola it should be `bin/docker-dev-shell npm_and_yarn`

The ecosystem is the directory in this project. I can add a blurb about that here as well. @jeffwidman also created an issue to track fixing this confusion: https://github.com/dependabot/dependabot-core/issues/6577

abdulapopoola: `go_modules` worked; so I guess it is a case of knowing the right params to pass? Is there a list somewher to smoothen the onboarding process for new devs?

deivid-rodriguez: Yeah, I also documented this at `CONTRIBUTING.md` [recently](https://github.com/dependabot/dependabot-core/pull/6582), but the README still needed an update. I went with this to document how to get the name of the ecosystem:

> Then, assuming you're working on a single language, you'll need to start a development container for that language through
>
> ```
> bin/docker-dev-shell <ecosystem>
> ```
>
> The name of the ecosystem should be one of the top level root folders in this repo.

jakecoffman: I pushed an update that addresses this, let me know what you all think!

deivid-rodriguez (approved):

PR #6658, opened 2023-02-13
Go: instrument mod version

jakecoffman:
We have a lot of code handling conditions around Go modules when they first were released. A lot has changes since then. 

This will help us get a better understanding of what version of Go modules is being used by Dependabot users, so we can clean up our code.

A smoke test failure is expected since this adds a new call.


jakecoffman:
I'm a bit confused why the Updater spec is failing. It works locally when I run it directly.

jeffwidman (approved):
Oh interesting, I didn't realize we collected this for some of the other ecosystems.

PR #6660, opened 2023-02-13
Allow updating GitHub reusable workflows

### LINKED ISSUE #4327: Dependabot to update reusable github-actions workflow
Hello,

I'd love for dependabot to also update reusable github actions.

I've got a repository with reusable actions created and started adding these to my repositories. I do pin them by semantic-version tag (v1.0.0, v1.1.0) and it would be really great if dependabot apart from custom actions could also update references to reusable actions.

dwc0011:
This PR fixes #4327

It changes how the "uses" key is searched for and found.  Reusable workflows do not have the "steps" key, this fix no longer make the assumption that a "steps" key exists, however it still processes the "steps" key if it does exist.  

The fix also allows for a mix of reusable or standard workflows to exist in the same directory.

New logic to find and update "uses":
Check for "jobs" key and provide that section to the search.
Recursively search for "uses" key.
If it is a hash and has "uses" key, then add the "uses" value to the `found_uses` array.
If it is a hash and has "steps" key, then send the "steps" value recursive search function.
If neither "steps" or "uses" are found, then send each item in the current hash/array into the recursive search function.

Once the entire jobs section has been searched return the `found_uses` and build a dependency set out of them.

Adds the following tests:
empty File
semver Reusable Workflow
standard Reusable Workflow

### DIFF
--- a/github_actions/spec/dependabot/github_actions/file_parser_spec.rb
+++ b/github_actions/spec/dependabot/github_actions/file_parser_spec.rb
@@ -167,6 +167,16 @@ RSpec.describe Dependabot::GithubActions::FileParser do
       end
     end
 
+    describe "with empty" do
+      subject(:dependency) { dependencies.first }
+      let(:workflow_file_fixture_name) { "empty.yml" }
+
+      it "has no dependencies" do
+        expect(dependencies.count).to be(0)
+        expect(dependency).to be_nil
+      end
+    end
+
     context "with a bad Ruby object" do
       let(:workflow_file_fixture_name) { "bad_ruby_object.yml" }
 
--- a/github_actions/spec/dependabot/github_actions/file_parser_spec.rb
+++ b/github_actions/spec/dependabot/github_actions/file_parser_spec.rb
@@ -140,6 +140,33 @@ RSpec.describe Dependabot::GithubActions::FileParser do
       end
     end
 
+    describe "with reusable workflow" do
+      subject(:dependency) { dependencies.first }
+      let(:workflow_file_fixture_name) { "workflow_reusable.yml" }
+
+      let(:expected_requirements) do
+        [{
+          requirement: nil,
+          groups: [],
+          file: ".github/workflows/workflow.yml",
+          source: {
+            type: "git",
+            url: "https://github.com/actions/checkout",
+            ref: "v2.1.0",
+            branch: nil
+          },
+          metadata: { declaration_string: "actions/checkout/.github/workflows/test.yml@v2.1.0" }
+        }]
+      end
+
+      it "has the right details" do
+        expect(dependency).to be_a(Dependabot::Dependency)
+        expect(dependency.name).to eq("actions/checkout")
+        expect(dependency.version).to eq("2.1.0")
+        expect(dependency.requirements).to eq(expected_requirements)
+      end
+    end
+
     context "with a bad Ruby object" do
       let(:workflow_file_fixture_name) { "bad_ruby_object.yml" }
 
@@ -165,6 +192,54 @@ RSpec.describe Dependabot::GithubActions::FileParser do
       it "ignores the Docker url reference" do
         expect(dependencies.count).to be(0)
         expect(dependency).to be_nil
+      end
+    end
+
+    context "with a semver tag pinned to a reusable workflow commit" do
+      let(:workflow_file_fixture_name) { "workflow_semver_reusable.yml" }
+      let(:service_pack_url) do
+        "https://github.com/actions/checkout.git/info/refs" \
+          "?service=git-upload-pack"
+      end
+      before do
+        stub_request(:get, service_pack_url).
+          to_return(
+            status: 200,
+            body: fixture("git", "upload_packs", "checkout"),
+            headers: {
+              "content-type" => "application/x-git-upload-pack-advertisement"
+            }
+          )
+      end
+
+      its(:length) { is_expected.to eq(1) }
+
+      describe "the first dependency" do
+        subject(:dependency) { dependencies.first }
+        let(:expected_requirements) do
+          [{
+            requirement: nil,
+            groups: [],
+            file: ".github/workflows/workflow.yml",
+            source: {
+              type: "git",
+              url: "https://github.com/actions/checkout",
+              ref: "01aecccf739ca6ff86c0539fbc67a7a5007bbc81",
+              branch: nil
+            },
+            metadata: {
+              declaration_string:
+              "actions/checkout/.github/workflows/test.yml@01aecccf739ca6ff86c0539fbc67a7a5007bbc81"
+            }
+          }]
+        end
+
+        it "has the right details" do
+          expect(dependency).to be_a(Dependabot::Dependency)
+          expect(dependency.name).to eq("actions/checkout")
+          expect(dependency.version).to eq("2.1.0")
+          expect(dependency.requirements).to eq(expected_requirements)
+        end
       end
     end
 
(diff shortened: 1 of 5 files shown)


dwc0011:
In addition to the added specs, I also ran against a test repo.
Two files define non-reusable actions (lint and release yml files)
Reusable workflow defined in test.yml
An empty.yml was also created, to ensure that empty files do not cause an exeception

yml files here:
https://github.com/dwc0011/test-old-actions-wf/tree/master/.github/workflows

PRs created:
https://github.com/dwc0011/test-old-actions-wf/pulls

github_actions/spec/dependabot/github_actions/file_parser_spec.rb
```diff
@@ -140,6 +140,43 @@
...
+        expect(dependency.version).to eq("2.1.0")
+        expect(dependency.requirements).to eq(expected_requirements)
+      end
+    end
+
+    describe "with empty" do
+      subject(:dependency) { dependencies.first }
+      let(:workflow_file_fixture_name) { "empty.yml" }
+
+      it "has the no dependencies" do
```
deivid-rodriguez, line 174:
```suggestion
      it "has no dependencies" do
```

deivid-rodriguez (commented):
This looks correct, thanks for adding this feature!

Only comment is that there are a few style changes mixed with the addition to the feature, so while reviewing this and trying to understand the PR, I removed those and simplified it a bit (I think) to try to make the implementation more clear. I'll be pushing some slight modifications tomorrow that's ok with you!

github_actions/spec/dependabot/github_actions/file_parser_spec.rb
dwc0011: No complaints from me, I am not super familiar with specs and am still learning ruby :)

deivid-rodriguez: It's a great PR, well covered by tests and implementation a very requested feature :) Thanks 😍.

I just noticed that the final diff was a bit bigger than it could, due to the different commits refining things and iteratively getting this to work properly. I pushed a slightly cleaner version that I made while reviewing!

deivid-rodriguez (approved):

dwc0011:
Thanks @deivid-rodriguez for the style changes and the approval.

Any thoughts on when this will be merged and released?

deivid-rodriguez:
I'll try to merge and deploy this soon™️.

deivid-rodriguez:
I do have one question. In the linked issue I referenced this piece of code:

https://github.com/dependabot/dependabot-core/blob/3846a60c18074d6b52f6d05c33a554e6a1b6ff31/github_actions/lib/dependabot/github_actions/file_fetcher.rb#L69-L72

My understanding now is that implementing this would be a separate feature that would allow dependabot to also update actions' dependencies of reusable workflows that live in the same repository being updated but in non-standard locations. Is this correct?

lorengordon:
> My understanding now is that implementing this would be a separate feature that would allow dependabot to also update actions' dependencies of reusable workflows that live in the same repository being updated but in non-standard locations. Is this correct?

Are reusable workflows _allowed_ to live in "non-standard" locations in a repo? All the examples I see in the docs point at paths under `.github/workflows`...

Ref: https://docs.github.com/en/actions/using-workflows/reusing-workflows

deivid-rodriguez:
I have no idea, I assumed since they are pointed to directly by file path, that they could live anywhere. But you're totally right that the docs seem to indicate the opposite. Well, that's great then, less work for us. Should we remove the TODO then?

dwc0011:
> I have no idea, I assumed since they are pointed to directly by file path, that they could live anywhere. But you're totally right that the docs seem to indicate the opposite. Well, that's great then, less work for us. Should we remove the TODO then?

I would think removing it would be appropriate.  The files are correctly being fetched from the ".github/workflows" directory, the only issue is it was looking for the "steps" hash for "uses", where with the new reusable actions the "uses" key can be many places under "jobs" including "steps".

jtgrohn:
> Are reusable workflows allowed to live in "non-standard" locations in a repo? All the examples I see in the docs point at paths under .github/workflows...

Yes. Reusable workflows work from any location.

deivid-rodriguez:
🤣 Ok, I'll remove the latest commit and reticket support that then.

dwc0011:
> > Are reusable workflows allowed to live in "non-standard" locations in a repo? All the examples I see in the docs point at paths under .github/workflows...
> 
> Yes. Reusable workflows work from any location.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_iduses
Do you have a reference for this? the docs indicate differently, if there is documentation that says otherwise please link.

Calling a reusable workflow
You call a reusable workflow by using the uses keyword. Unlike when you are using actions within a workflow, you call reusable workflows directly within a job, and not from within job steps.

[jobs.<job_id>.uses](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_iduses)

You reference reusable workflow files using one of the following syntaxes:

{owner}/{repo}/.github/workflows/{filename}@{ref} for reusable workflows in public and private repositories.
./.github/workflows/{filename} for reusable workflows in the same repository.

jtgrohn:
> > > Are reusable workflows allowed to live in "non-standard" locations in a repo? All the examples I see in the docs point at paths under .github/workflows...
> > 
> > 
> > Yes. Reusable workflows work from any location.
> 
> https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_iduses Do you have a reference for this? the docs indicate differently, if there is documentation that says otherwise please link.
> 
> Calling a reusable workflow You call a reusable workflow by using the uses keyword. Unlike when you are using actions within a workflow, you call reusable workflows directly within a job, and not from within job steps.
> 
> [jobs.<job_id>.uses](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_iduses)
> 
> You reference reusable workflow files using one of the following syntaxes:
> 
> {owner}/{repo}/.github/workflows/{filename}@{ref} for reusable workflows in public and private repositories. ./.github/workflows/{filename} for reusable workflows in the same repository.

Tested it out and turns out I'm wrong. I swear I had done this previously, but I guess not.

image

dwc0011:
> > > > Are reusable workflows allowed to live in "non-standard" locations in a repo? All the examples I see in the docs point at paths under .github/workflows...
> > > 
> > > 
> > > Yes. Reusable workflows work from any location.
> > 
> > 
> > https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_iduses Do you have a reference for this? the docs indicate differently, if there is documentation that says otherwise please link.
> > Calling a reusable workflow You call a reusable workflow by using the uses keyword. Unlike when you are using actions within a workflow, you call reusable workflows directly within a job, and not from within job steps.
> > [jobs.<job_id>.uses](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_iduses)
> > You reference reusable workflow files using one of the following syntaxes:
> > {owner}/{repo}/.github/workflows/{filename}@{ref} for reusable workflows in public and private repositories. ./.github/workflows/{filename} for reusable workflows in the same repository.
> 
> Tested it out and turns out I'm wrong. I swear I had done this previously, but I guess not.
> 
> image

Thanks for testing it and letting us know.  @deivid-rodriguez I think we are good and due diligence has been done at this point.

deivid-rodriguez:
Alright, I will move this forward in the following days.

deivid-rodriguez:
Alright, this is now live! Let me know if you find any issues and thanks for the fantastic contribution ❤️.

dwc0011:
> Alright, this is now live! Let me know if you find any issues and thanks for the fantastic contribution ❤️.

Thank you for the quick reviews, refactors and merge!

PR #6662, opened 2023-02-14
remove bundler cache after installing

jakecoffman:
The Bundler ecosystem is evading the smoke test cache after a while, causing failures. This is probably due to how Bundler caches dependency data. When we run `bundle install` during a build, it stores a cache of the versions it fetched under `~/.bundle`. After a day or so this is out of date and Bundler attempts to get new data, probably sending new headers to do so, avoiding the cache.

So by removing the Bundler cache during image build, our tests should always make the same call to get fresh data instead of starting with a snapshot at the time of build.

An outstanding question is whether this is a good idea outside of testing. With this change, updates become more deterministic, so I think it's worth the extra download cost it incurs. Also we don't need to make a special case in tests to handle this one-off discrepancy. Other opinions on this are welcome.

### DIFF
--- a/Dockerfile.updater-core
+++ b/Dockerfile.updater-core
@@ -96,7 +96,8 @@ WORKDIR $HOME/dependabot-updater
 RUN bundle config set --local path 'vendor' && \
 bundle config set --local frozen 'true' && \
 bundle config set --local without 'development' && \
-bundle install
+bundle install && \
+rm -rf ~/.bundle
 
 FROM ubuntu:20.04
 


deivid-rodriguez:
Just to clarify Bundler behavior.

Bundler uses a `/versions` file for resolution that includes the list of all releases of all gems. So in case new gems were released since you last downloaded this file, then Bundler will only request "new information" and append that to the existing local file. For this it uses `Range` and `If-None-Match` request headers. This is so that you don't need to download a huge text file everytime you use Bundler.

So, yes, if we remove this cached file the updates will be technically more deterministic but the result should not change anyways since all Bundler resolutions will make sure the `versions` file is up to date first (regardless of whether a full download or an incremental one is necessary). And if we remove it, all updates will need to download the full file (I just checked and it's 18Mb, I thought it was more actually).

jakecoffman:
@deivid-rodriguez thanks for the clarification! The determinism will be in the calls that are made during the update, and not the change that is created by Dependabot. 👍

deivid-rodriguez:
This failed because the smoke tests cache needs to be refreshed after this PR?

jakecoffman:
@deivid-rodriguez if I refresh it now it will go green too, I can do that.

jakecoffman:
@deivid-rodriguez Fixed! Also the fresh download from rubygems is 6MB compressed, so I think I'm good with going forward with this if it fixes our caching problems.

deivid-rodriguez (approved):
I was a bit concerned with making a change that degrades our real updater only to fix a test specific issue, hence my suggestion to run smoke tests with an empty mount on the versions file. But I guess you're right and this won't be really noticiable.

jakecoffman:
We can always revisit in the future if anything changes. Also this made the image smaller, so for Actions and the Smoke Tests that download the updater image each time, this probably saves a 🤏  small amount of bandwidth for the case where the updater image has an out-of-date cache built-in.

deivid-rodriguez:
Sure, thanks for fixing this! 🎉

PR #5867, opened 2022-10-10
basic first pass at creating a grouped PR

mctofu:
WIP exploration of updating multiple deps in a single PR. This converts the dry-run.rb to produce a single PR for all dependency updates.

2 main changes:
- The update of each dependency in the dry-run script works against the data in $files. After each dependency is updated I needed to apply those changes to $files so the next update would use that as a starting base. So there shouldn't be any merge conflicts but we definitely could end up in a case where the 2nd dependency can't update due to the updates to the 1st dependency. I think that will just be a limitation of this approach rather than something we need to try to solve.
- I keep an overall list of which dependencies were updated & which files were updated (with their most recent content) and that gets passed to the PR creator
```
bin/dry-run.rb npm_and_yarn github/dependabot-action --dep axios,dockerode --pull-request
To use retry middleware with Faraday v2.0+, install `faraday-retry` gem
=> fetching dependency files
=> dumping fetched dependency files: ./dry-run/github/dependabot-action/
=> parsing dependency files
=> updating 2 dependencies: axios, dockerode

=== axios (0.23.0)
 => checking for updates 1/2
🌍 https://registry.npmjs.org/axios
🌍 https://registry.npmjs.org/axios/1.1.2
 => latest available version is 1.1.2
 => latest allowed version is 1.1.2
 => requirements to unlock: own
 => requirements update strategy: bump_versions
 => updating axios from 0.23.0 to 1.1.2

=== dockerode (3.3.1)
 => checking for updates 2/2
🌍 https://registry.npmjs.org/dockerode
🌍 https://registry.npmjs.org/dockerode/3.3.4
 => latest available version is 3.3.4
🌍 https://registry.npmjs.org/@types%2Fdockerode
🌍 https://registry.npmjs.org/@types%2Fdockerode/3.3.11
 => latest allowed version is 3.3.1
 => requirements to unlock: all
 => requirements update strategy: bump_versions
🌍 https://registry.npmjs.org/@types%2Fdockerode
🌍 https://registry.npmjs.org/@types%2Fdockerode/3.3.11
🌍 https://registry.npmjs.org/dockerode
🌍 https://registry.npmjs.org/dockerode/3.3.4
 => updating dockerode, @types/dockerode
🌍 https://registry.npmjs.org/axios/latest
🌍 https://registry.npmjs.org/dockerode/latest
🌍 https://registry.npmjs.org/@types%2Fdockerode/latest
🌍 https://github.com/axios/axios.git/info/refs
[...]

PR #6663, opened 2023-02-14
[Prototype] Generating grouped update PRs

brrygrdn:
### ⚠️ This is an experimental spike ⚠️ 

The purpose of this is spike is to create an isolated code line to reimplement `Dependabot::Update#run`  without taking on the need to verify our prototype work as strictly as the production implementation.

It is important to fail-fast on our proposed strategy and test our service's assumptions about PRs without investing more time in refactoring the existing code to make way for a new architecture that we can't be sure will actually work out.

**This should not be considered indicative of the final architecture**

### The Approach

This spike explores grouping in a naive, ecosystem-agnostic way by using the `Dependabot::Updater` class to drive collecting up the Dependency changes required by a particular group rule - for now we assume a rule of '*', i.e. group everything - and then executing a single `FileUpdater` and PR creation invocation on the aggregate changes.

For purposes of simplicity, we have disabled concerns around managing the lifecycle of PRs, i.e. superseding and rebasing existing PR, as well as set aside Security-only update strategies since they are already special-cased.

### Goals

1. Flush out general refactors required in the `updater/` code to improve quality, test coverage, etc
2. Determine architectural changes required to `Dependabot::Updater` to better isolate the `UpdateChecker` and `FileUpdater` steps to break the assumption they are always 1:1
3. Provide a walking skeleton we can attach further experiments within `common/lib` to in order to test our hypothesis without waiting on (1) and (2) to ship

### Non-goals

1. Write production quality-code
    - This is a highly experimental spike, if we do eventually merge it for continued testing it will remain behind an experiment flag and remain isolated in such a way it cannot impact existing code.
    - This should also be considered a throw-away prototype, so using this in our own scripts is **highly discouraged** and will not be supported.
2. Address the `dry-run` script
    - https://github.com/dependabot/dependabot-core/pull/5867 spikes out this strategy in the `dry-run` script as part of an earlier experiment
    - We might further the #5867 spike to do some more exploration, but it should be considered a feed-in to this work which will take priority
    - Once we have settled the `Dependabot::Updater` architecture we will figure out how to best maintain the `dry-run` script as the two overlap heavily

### DIFF
--- a/updater/lib/dependabot/experimental_grouped_updater.rb
+++ b/updater/lib/dependabot/experimental_grouped_updater.rb
@@ -234,7 +234,7 @@ module Dependabot
       #
       #        It is also worth noting that it mutates the `dependency_files`
       #        ivar which I *think* is acceptable in interests of memory management
-      group_updated_files = update_batch.inject(dependency_files) do |files, (lead_dep, updated_deps)|
+      group_updated_files = update_batch.inject(dependency_files) do |files, (lead_dep_name, updated_deps)|
         # FIXME: Move the creation of the all_updated_deps list into this loop
         #
         # If we fail to generate a diff, the PR body and Service will still get the `updated_deps` that
@@ -244,6 +244,10 @@ module Dependabot
         # which might include supressing 'downgrades' if a previous step moved a given dep to a higher
         # value already.
         if updated_deps&.any?
+          lead_dep_name = lead_dep_name.downcase
+          lead_dep = updated_deps.find do |dep|
+            dep.name.downcase == lead_dep_name
+          end
           generate_dependency_files_for(lead_dep, updated_deps, files) if updated_deps&.any?
         else
           files # pass on the existing if there are no updated dependencies for this lead dependency
--- a/updater/lib/dependabot/experimental_grouped_updater.rb
+++ b/updater/lib/dependabot/experimental_grouped_updater.rb
@@ -21,32 +21,32 @@ module Dependabot
       end
 
       logger_info("[Experimental] Starting grouped update job for #{job.source.repo}")
-      # Establish collection
-      # Do the check and diff
-      dependencies.each { |dep| check_and_create_pr_with_error_handling(dep) }
-      # PR everything
+      # We should log the rule being executed, let's just hard-code wildcard for now
+      # since the prototype makes best-effort to do everything in one pass.
+      logger_info("Starting batch for Group Rule: '*'")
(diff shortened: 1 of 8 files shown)


updater/lib/dependabot/experimental_grouped_updater.rb
```diff
@@ -0,0 +1,276 @@
...
+      #
+      #        In a follow-up iteration, a 'shared workspace' could provide the
+      #        filtering for us assuming we iteratively make file changes for
+      #        each Array of dependencies in the batch and the FileUpdater tells
+      #        us which cannot be applied.
+      all_updated_deps = update_batch.values.flatten
+      logger_info("Updated #{all_updated_deps.count} dependencies")
+      updated_files = generate_dependency_files_for(all_updated_deps)
+
+      create_pull_request(all_updated_deps, updated_files, pr_message(all_updated_deps, updated_files))
```
brrygrdn, line 237: This passes existing tests but is too naive to work under most circumstances, this is where we need to replicate the iterative file changes from the [dry-run prototype](https://github.com/dependabot/dependabot-core/pull/5867/files#diff-095eccc30d572b6b06c61c5b23d291ddcdaca2b3255650fd303590a86781d964R780-R788).

@bdragon I think this is essentially where the control plane for the underlying file system would be, with `generate_dependency_files_for` being executed per row of `update_batch.values` and the outcome resulting in either a `git commit` or `git clean`.

updater/spec/dependabot/experimental_grouped_updater_spec.rb
```diff
@@ -0,0 +1,1986 @@
...
+
+# FIXME: This file is a copy-paste of `spec/dependabot/updater_spec.rb`
+#
+#       The intent is to run the tests for existing behaviour without the churn
+#       of refactoring the existing tests into shared examples until we know
+#       how the code is going to decompose
+#
+#       Tests that don't apply or work properly for grouped updates are skipped
+#       rather than deleted, and new group-specific tests are added under the
+#       `run_grouped` block.
```
brrygrdn, line 11: As noted, this test is a fork of the existing tests as just acts as a plumb-line for how much coverage we have on the new code, what is currently side-stepped, etc as a way to heatmap where we need to bring up coverage before we start changing the 'real' implementation.

landongrindheim (approved):
Approving to unblock image building workflows 🔓

brrygrdn:
Closing this out as superseded by https://github.com/dependabot/dependabot-core/pull/6663

PR #6667, opened 2023-02-14
remove activesupport dependency by creating package_manager_version

jakecoffman:
Sparked by my troubles in https://github.com/dependabot/dependabot-core/pull/6658 to get the Updater test working, and remembering that updating activesupport to the latest version broke things in https://github.com/dependabot/dependabot-core/pull/6522, I asked why we need it at all?

It seems more straight forward to elevate a method (`package_manager_version`) to be implemented in the base, and have file_fetchers implement it than to use Notifications. That and we get to delete a dependency that is giving us trouble!


common/lib/dependabot/shared_helpers.rb
```diff
@@ -151,7 +150,6 @@ def self.excon_defaults(options = nil)
       options ||= {}
       headers = options.delete(:headers)
       {
-        instrumentor: ActiveSupport::Notifications,
```
jakecoffman, line 154: I'm not sure what this was gaining us since I don't see us subscribing to any other notifications?

jakecoffman: Ah I see in dry-run we do have a subscribe.

jakecoffman: I'll see if there's another way to achieve what we had without ActiveSupport.

jakecoffman:
Looks like this is mostly working now, just some Active Support extensions we could either monkey-patch back in or rewrite with plain-ole-Ruby. The dry-run HTTP request output is nice so we'll definitely want to add that back in somehow.

I will pause now for any input. Is this a good idea?

deivid-rodriguez:
ActiveSupport is helpful is you use a significant part of the functionality it provides. In our case, we seem to be using just minimal functionality but we get the disadvantages of globally monkeypatching a lot of core classes. If we can do this, I'm all for it.

jakecoffman:
I think I got everything. I opted to replace ActiveSupport extensions instead of monkey patching them back in. Also made the executive decision to delete bin_run_spec.rb since we have a suite of test that use bin/run with more thorough assertions.

bin/dry-run.rb
```diff
@@ -804,7 +800,6 @@ def security_fix?(dependency)
 StackProf.results("tmp/stackprof-#{Time.now.strftime('%Y-%m-%d-%H:%M')}.dump") if $options[:profile]
 
 puts "🌍 Total requests made: '#{$network_trace_count}'"
-puts "🎈 Package manager version log: #{$package_manager_version_log.join('\n')}" if $package_manager_version_log.any?
```
deivid-rodriguez, line 807: Should we replace this with the information now provided by the fetcher?

maven/lib/dependabot/maven/update_checker/version_finder.rb
```diff
@@ -58,7 +58,7 @@ def versions
             repositories.map do |repository_details|
               url = repository_details.fetch("url")
               xml = dependency_metadata(repository_details)
-              next [] if xml.blank?
+              next [] if xml.nil? || xml.blank?
```
deivid-rodriguez, line 61: The `nil?` part can be undone in the end I guess?

deivid-rodriguez (commented):
Looks great to me, awesome job!

I guess this could be logically divided into two commits, one to remove the end to end fetcher specs, the other one to remove the activesupport dependency? Happy to rebase and squash this to do that, just to get git history a bit cleaner.

bin/dry-run.rb
jakecoffman: Ah yes, of course!

maven/lib/dependabot/maven/update_checker/version_finder.rb
jakecoffman: Uh yeah, what happened here? 😆

jakecoffman: Oh ok, so ActiveSupport seems to have been providing a nil-safe `blank?`, but there's also a Nokogiri `blank?` that isn't nil-safe. So I think we can just just for `nil?` here.

deivid-rodriguez: Mmmm, I guess if we need to deal with [blank? XML nodes](https://www.rubydoc.info/github/sparklemotion/nokogiri/Nokogiri%2FXML%2FNode:blank%3F) here, then the current version adding back nil safety seems fine?

jakecoffman: I don't think it matters since it only calls `css` with a selector that it won't find, should have the same result either way.

```ruby
irb(main):006:0> Nokogiri.XML("not even valid").css("no > exist")
=> []
```

jakecoffman:
@deivid-rodriguez rebased

deivid-rodriguez (approved):
Really happy about this, thanks for the nice work!

PR #6671, opened 2023-02-15
Respect inclusive upper bound style in requirements.txt files

### LINKED ISSUE #5149: Dependabot increase to latest available
_tbh, not sure if this is the right place to ask as it is about the GH bot, not action._

I wonder if there is a way to make the dependency update safe/reliable and update to the latest available package version so it can be tested and not increase to a future version that is not released yet...

for example, if we have requirements such as
```
packageABC >= 0.1, <=0.8.6
```
and there is a new release `0.8.7`, so I want to increase the version to `<=0.8.7` not to `<0.9` as I can't know if eventually next `0.8.8` is also safe... see the sample PR from dependable bellow which bumped version from actual 12.x to even 13.0 which is unacceptable :(

https://github.com/PyTorchLightning/pytorch-lightning/pull/13047/files
also sked on https://github.community/t/dependabot-increase-to-latest-available/250979

deivid-rodriguez:
Fixes #5149.

### DIFF
--- a/python/spec/dependabot/python/update_checker/requirements_updater_spec.rb
+++ b/python/spec/dependabot/python/update_checker/requirements_updater_spec.rb
@@ -134,7 +134,7 @@ RSpec.describe Dependabot::Python::UpdateChecker::RequirementsUpdater do
                 let(:requirement_txt_req_string) { "<=1.9.2,>=1.9" }
                 let(:latest_resolvable_version) { "1.10" }
 
-                its([:requirement]) { is_expected.to eq(">=1.9,<1.11.0") }
+                its([:requirement]) { is_expected.to eq(">=1.9,<=1.10") }
               end
             end
           end
--- a/python/lib/dependabot/python/update_checker/requirements_updater.rb
+++ b/python/lib/dependabot/python/update_checker/requirements_updater.rb
@@ -275,8 +275,10 @@ module Dependabot
             next r.to_s if r.satisfied_by?(latest_resolvable_version)
 
             case op = r.requirements.first.first
-            when "<", "<="
-              "<" + update_greatest_version(r.to_s, latest_resolvable_version)
+            when "<"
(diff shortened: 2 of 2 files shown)


python/spec/dependabot/python/update_checker/requirements_updater_spec.rb
```diff
@@ -134,7 +134,7 @@
                 let(:requirement_txt_req_string) { "<=1.9.2,>=1.9" }
                 let(:latest_resolvable_version) { "1.10" }
 
-                its([:requirement]) { is_expected.to eq(">=1.9,<1.11.0") }
+                its([:requirement]) { is_expected.to eq(">=1.9,<=1.10") }
```
jeffwidman, line 137: Given that the user was requesting `1.9.2`, we're losing specificity here since we're now only specifying `1.10`, not `1.10.0`... 

I think that's okay in this case since the upstream version is `1.10`? 🤷‍♂️

jeffwidman (approved):

jeffwidman:
I think we should ship this as-is, and can iterate on it further down the road based on user feedback.

python/spec/dependabot/python/update_checker/requirements_updater_spec.rb
jeffwidman: Yeah, on second thought this is probably fine--we're losing specificity in our explicit pin because the upstream package chose not to preserve the specificity of their previous releases. 

The alternative of trying to guess when to increase the specificity of upstream versioning seems like it has some edge cases where some percentage of users will always say we picked wrong.

deivid-rodriguez: Yeah, that was my thinking too. I think it should be fine as is but as you said we could further iterate based on feedback 👍.

PR #6674, opened 2023-02-15
fix some smoke tests skipping due to filters

jakecoffman:
I noticed some smoke tests were skipping incorrectly due to their name not matching up with the filter name. This brings it in-line with `ci.yml`.


deivid-rodriguez (approved):

jakecoffman:
D'oh the name of the test in smoke tests is now not matching up! I'll get them all to agree one way or another.

PR #6684, opened 2023-02-16
Try fix images fork workflow

deivid-rodriguez:
I believe we should be pushing the same TAG for all ecosystem, just like we do for regular PRs.


jakecoffman:
This is a bit confusing, but in the regular workflow we're using `{{ github.sha}}` which for the `pull_request` trigger type is the [SHA of the merge branch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request). So GitHub merges the branch to main and then takes the SHA which is different than the SHA of the last branch commit. 

So I was attempting to simulate that here with forks by merging them to main via cherry-pick because forks can't be merged directly.

I'm not opposed to changing it though, and I might always be wrong about the above statements 😄

deivid-rodriguez:
@jakecoffman Sure, but does it make any difference for us? Isn't this just a unique name to the image that we build, but not linked to the source code other than that?

deivid-rodriguez:
My initial approach was splitting the job into two, once that does this dance and actually pushes the `docker-branch-release-workflow` branch, and that does not have a matrix of ecosystems. Then another one that `needs:` the first one that checks out the temporary branch and builds each images using this tag. That way all images reuse the same tag.

But then I thought, do we really need this?

jakecoffman (approved):
Not sure this was necessary, so I'm good with removing it 👍

deivid-rodriguez:
Ok, I added one more commit to reject non-rebased PRs (previously we were automatically rebasing so it was not necessary), but I think it'd be good to check it now to make sure we don't actually revert any functionality.

jakecoffman (approved):
Makes sense 👍

deivid-rodriguez:
Oh my god, that was tough, but I finally managed to test that last version I pushed does what I expected:

* Push all ecosystem images tagged with the latest commit in the PR.
* Reject not fully rebased PRs.

PR #6690, opened 2023-02-17
Fix updating composite actions

deivid-rodriguez:
I think #6660 has regressed this case. I noticed it thanks to https://github.com/dependabot/dependabot-core/issues/4178#issuecomment-1432287375.

I'll try to add a spec and ship this soon.


Nishnha (approved):
This LGTM 
Only thing I would consider adding is checking that the file parser returned the correct number of dependencies (does `docker/setup-qemu-action@v2.1.0` count as an action, for example?)

Unsure if it's already checked elsewhere

deivid-rodriguez:
Thanks @Nishnha! Yup, all `uses:` should count as dependencies. I added an extra spec to check that.

PR #6693, opened 2023-02-17
Add workspace experiment to maintain state between updates and capture success/failure of each

bdragon:
Over the course of a version update, Dependabot will typically use the [`in_a_temporary_repo_directory`](https://github.com/dependabot/dependabot-core/blob/ef20a02f81275d0ac49232b0a7117eefcf125373/common/lib/dependabot/shared_helpers.rb#L26) and [`in_a_temporary_directory`](https://github.com/dependabot/dependabot-core/blob/ef20a02f81275d0ac49232b0a7117eefcf125373/common/lib/dependabot/shared_helpers.rb#L42) shared helpers (depending on whether the current ecosystem has been configured to [`always_clone_for_package_manager?`](https://github.com/dependabot/dependabot-core/blob/ef20a02f81275d0ac49232b0a7117eefcf125373/common/lib/dependabot/utils.rb#L40) or not, respectively) to make changes to manifest files using the native package manager.

This PR introduces the concept of a shared, or persistent workspace as an experiment behind the `shared_workspace` feature flag. When the experiment is enabled, it will only be used by always-clone package managers (in practice: `go_modules`, `npm_and_yarn`, `pub`, and `terraform`) or when vendoring dependencies. This is because the implementation is backed by git and requires a git repo to work. In the future, we may simply `git init` ([like this](https://github.com/dependabot/dependabot-core/blob/d6175dec361f9aedeac213919c56ac22028aa02d/go_modules/lib/dependabot/go_modules/file_updater.rb#L70-L91)) before using the `in_a_temporary_directory` shared helper so that the concept can be used across all ecosystems regardless of whether or not they clone.

A workspace is simply a git repo: successful changes are commits and failed changes are stashes. When the experiment is enabled and `in_a_temporary_repo_directory` is invoked, its block argument will be passed to the workspace's `#change` method. If the block exits without failure, any changes to the repo contents will be captured as a successful change and committed to the repo; if the block raises, the state of the repo directory (and the exception that was raised) will be captured as a failed change attempt and stashed to the repo.

This concept allows us to iteratively make changes to manifest files, each change starting from the state of the previous successful change, rather than starting from scratch each time. This has obvious potential benefits for grouped updates, in which several updates are made in succession over the course of a single version update job. As of this PR, if a change raises, the exception will be re-raised to mirror current behavior. In the future we may want to inspect and potentially suppress the failure and carry on with the next update in order to allow partially-successful grouped updates.

While this experiment may ultimately support grouped updates, it is being released separately so that we can start evaluating its viability independent of grouped updates work.


common/spec/dependabot/workspace_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+          index 0000000..3a30967
+          --- /dev/null
+          +++ b/Gemfile
+          @@ -0,0 +1 @@
+          +gem "timecop", "~> 0.9.6", group: :test
+          DIFF
+        )
+        expect(workspace.change_attempts.first.error).to be_nil
+        expect(workspace.change_attempts.first.error?).to be_falsy
+        expect(workspace.change_attempts.first.success?).to be_truthy
```
brrygrdn, line 70: 👏🏻 I _really_ love this interface

brrygrdn (commented):
This looks great! I really like the `workspace.change {}` call signature, and I can see the methods you've created being something we can plug into logging to better surface errors.

common/lib/dependabot/shared_helpers.rb
```diff
@@ -25,14 +26,19 @@ module SharedHelpers
 
     def self.in_a_temporary_repo_directory(directory = "/",
                                            repo_contents_path = nil,
+                                           **options,
```
brrygrdn, line 29: ❓ Should we avoid changing the method signature of this class?

It doesn't look like we are actually setting `options[:memo]` currently and the workspace tolerates this being `nil` by using placeholders.

I think we should ideally try to pass more meaningful strings in eventually so there's no need to change the use of `memo` inside the `Dependabot::Workspace`, but perhaps we should remove it from here for now until we have concrete plans to pass in some information about the change from the `FileUpdater`?

common/lib/dependabot/shared_helpers.rb
```diff
@@ -25,14 +26,19 @@ module SharedHelpers
...
-        path = Pathname.new(File.join(repo_contents_path, directory)).
-               expand_path
+        path = Pathname.new(File.join(repo_contents_path, directory)).expand_path
         reset_git_repo(repo_contents_path)
         # Handle missing directories by creating an empty one and relying on the
         # file fetcher to raise a DependencyFileNotFound error
         FileUtils.mkdir_p(path)
+
+        if (workspace = Dependabot::Workspace.active_workspace)
+          return workspace.change(options[:memo], &block)
```
brrygrdn, line 39: As above, I think we can just pass the block for now and start setting a memo in the `FileUpdater` once we've verified our experiment and wired it up into metrics.

updater/lib/dependabot/updater/operations/group_update_all_versions.rb
```diff
@@ -39,6 +39,7 @@ def perform
           # We should log the rule being executed, let's just hard-code wildcard for now
           # since the prototype makes best-effort to do everything in one pass.
           Dependabot.logger.info("Starting update group for '#{GROUP_NAME_PLACEHOLDER}'")
+          prepare_workspace
```
brrygrdn, line 42: 👍🏻

brrygrdn (commented):
I have one question about the `options` hash as I _think_ it is unused for now so it might make sense to just omit it for now and bring it back once we have proven things out more.

common/lib/dependabot/shared_helpers.rb
bdragon: That makes sense. Let's get things working and we can figure out how to pass context later.

common/lib/dependabot/shared_helpers.rb
```diff
@@ -27,12 +28,16 @@ def self.in_a_temporary_repo_directory(directory = "/",
...
       if repo_contents_path
-        path = Pathname.new(File.join(repo_contents_path, directory)).
-               expand_path
+        path = Pathname.new(File.join(repo_contents_path, directory)).expand_path
         reset_git_repo(repo_contents_path)
         # Handle missing directories by creating an empty one and relying on the
         # file fetcher to raise a DependencyFileNotFound error
         FileUtils.mkdir_p(path)
+
+        if (workspace = Dependabot::Workspace.active_workspace)
```
brrygrdn, line 37:
```suggestion
        # If a workspace object has been initialised for us, we should prefer to use it so we commit 
        # any changes we make with additional debugging, otherwise we just use the bare commands.
        if (workspace = Dependabot::Workspace.active_workspace)
```

I think this is worth a comment so there's a breadcrumb to look for it being assigned, etc.

common/lib/dependabot/shared_helpers.rb
```diff
@@ -26,7 +26,6 @@ module SharedHelpers
 
     def self.in_a_temporary_repo_directory(directory = "/",
                                            repo_contents_path = nil,
-                                           **options,
```
brrygrdn: Thanks for removing this for now 👍🏻

updater/lib/dependabot/updater.rb
```diff
@@ -9,6 +9,7 @@
 require "dependabot/logger"
 require "dependabot/security_advisory"
 require "dependabot/update_checkers"
+require "dependabot/workspace"
```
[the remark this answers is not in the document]
bdragon: This makes sense. I've backed out the changes that would initialize a workspace for now.

brrygrdn (approved):
Thanks for the updates, lets :shipit:

PR #6703, opened 2023-02-21
fix: improve yarn@^1 subdependency update success

MarshallOfSound:
Previously the logic attempted fresh installs of package@version and copied the generated lockfile contents across to force a dependency realignment.  This was sometimes flawed as "version" could somehow (didn't figure this out) sometimes be a pinned version, resulting in a noop lockfile change for a dependency bump that was definitely possible.

The new logic is much more straightfoward IMO, we take the existing lockfile.  Remove all entries that map to the given package that also resolve to the suspect version and then just run "yarn" again and take the resultant lockfile.

This logic is mirroring my implementation in `uuaw` --> https://github.com/MarshallOfSound/uuaw which aimed to automatically fix `yarn audit` security alerts which dependabot failed to solve automatically.  I figured it would be better for everyone if I could somehow take the logic that worked in `uuaw` and mirror it into dependabot-core so here we are.

There is currently one unsolved low audit output on `electron/website` that I was using to test this change.  Specifically dependabot currently reports it is unable to fix it, with this change it successfully bumps the version.  Command:

```bash
bin/dry-run.rb --dep=@sideway/formula npm_and_yarn electron/website
```

Notably as well this new logic almost exactly aligns with the `npm6` logic in `updateDependencyFile` so consistency is good too 👍 

Tests all pass locally for `npm_and_yarn`

### DIFF
--- a/npm_and_yarn/helpers/lib/yarn/subdependency-updater.js
+++ b/npm_and_yarn/helpers/lib/yarn/subdependency-updater.js
@@ -5,7 +5,7 @@ const Config = require("@dependabot/yarn-lib/lib/config").default;
 const { EventReporter } = require("@dependabot/yarn-lib/lib/reporters");
 const Lockfile = require("@dependabot/yarn-lib/lib/lockfile").default;
 const fixDuplicates = require("./fix-duplicates");
-const { LightweightAdd, LightweightInstall } = require("./helpers");
+const { LightweightInstall, LOCKFILE_ENTRY_REGEX } = require("./helpers");
 const { parse } = require("./lockfile-parser");
 const stringify =
   require("@dependabot/yarn-lib/lib/lockfile/stringify").default;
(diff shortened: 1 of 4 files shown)


deivid-rodriguez (approved):
Thanks so much for working this and sharing your fix to the whole community 😍. Really great when bug fixes also result in simpler code.

One question, are the spec changes needed because the expectations were incorrect before, and that was surfaced by the bug fix?

MarshallOfSound:
> One question, are the spec changes needed because the expectations were incorrect before, and that was surfaced by the bug fix?

Yes, the old code ensured that the sub dependency was updated specifically to `version` and only exactly `version`, this isn't how berry or npm6 work and was actually partially the cause of the underlying issue where some subdependencies weren't ever updated.  For instance one of these tests is shared across npm6, berry and yarn and the update to the test aligned the "expected result" with npm6 and berry, when yarn was for some reason different

deivid-rodriguez:
Great, I'll try to ship this in the next few days!

jeffwidman (approved):

PR #6705, opened 2023-02-21
Rename Updater entrypoints from -Job to -Command

brrygrdn:
We implement a Dependabot::Job class as part of the Updater which is a value class that represents a single unit of work within the Dependabot Service and is used in a few places.

The entrypoint classes we have bump up against this with their current naming and it creates some confusion, especially as we'd like to pass the entrypoint object to the updater in order to  act as a binding to the client/service and other variables which should persist between runs.

In future, this will allow us to firm up the entrypoint/command class pattern so we can start to push more of the updater logic into the ecosystem gems in order to allow them to operate more independently where it makes sense ( i.e. they have native support for grouping/vulnerability data handling/etc )


bdragon (approved):
Nice 👍

PR #6707, opened 2023-02-21
Properly support the Updater project in the dev shell

brrygrdn:
Supersedes https://github.com/dependabot/dependabot-core/pull/6626

This updates the `Dockerfile.development` with the additional requirements needed to bundle and test the `updater/` project inside the `bin/docker-dev-shell` and aligns the README with this workflow.

### DIFF
--- a/updater/README.md
+++ b/updater/README.md
@@ -11,30 +11,37 @@ GitHub network, and so is not generally accessible.
 
 ## Setup
 
-You will need to provide the build a Personal Access Token to access the GitHub Package Registry to retrieve
-dependency containers.
+To work on the Updater, you will need to start a Docker dev shell. The bundler
+shell will suffice
 
-[Create a token](https://github.com/settings/tokens/new) with the `packages:read` scope and set it in your environment
-as `GPR_TOKEN`
-
-Run the setup script:
-
-```
-script/setup
+```zsh
+➜ bin/docker-dev-shell bundler
+[dependabot-core-dev] ~/dependabot-core $ cd updater/
+[dependabot-core-dev] ~/dependabot-core/updater $ bundle
 ```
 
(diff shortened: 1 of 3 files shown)


updater/README.md
```diff
@@ -11,30 +11,37 @@ GitHub network, and so is not generally accessible.
...
+[dependabot-core-dev] ~/dependabot-core/updater $ bundle exec rspec spec/dependabot/integration_spec.rb
 ```
-script/test spec/dependabot/integration_spec.rb
+
+A small number of tests hit the GitHub API, so you will need to set the envvar
+`DEPENDABOT_TEST_ACCESS_TOKEN` with a Personal Access Token with the full `repo`
+scope.
+
+```zsh
+export DEPENDABOT_TEST_ACCESS_TOKEN=ghp_xxx
```
landongrindheim, line 42: Since we're suggesting using a PAT, it makes sense to me to suggest keeping secrets out of shell history here. What do you think?

```suggestion
# keep secrets from being stored in shell history by prefixing with a space
export HISTCONTROL=ignorespace
export  DEPENDABOT_TEST_ACCESS_TOKEN=ghp_xxx
```

landongrindheim (approved):

updater/README.md
brrygrdn: Yeah, that's a good idea 👍🏻

brrygrdn:
There is one test that is prompting for a git username/password combo as the dev shell is running with a tty unlike CI, but that is an existing problem we may not have seen in the past as it may just have silently used your git config ( which is maybe worse! ) - the creds aren't required for the test to pass so I'm going to punt on finding a way to stub that as part of this PR and handle it later.

PR #6713, opened 2023-02-22
Go: update psuedo-versions to released versions

jakecoffman:
Currently Dependabot doesn't touch Go pseudo-versions, they're handled as if they were pinned on a revision like other ecosystems.

However, the Go tooling generates pre-release identifiers that are semver compliant. That means Dependabot does have enough information to bump a pseudo-version like `v1.0.1-<date>-<hash>` to a tagged `v1.0.1`. 

This PR allows the update to continue, and adds the pseudo-version to the list of candidates to avoid downgrades, in the event the user is pinned to a newer version than is currently released.

This will *not* update the dependency to a non-released pseudo-version, only pseudo to released.

Note: I will need to update the Go smoke test.

### DIFF
--- a/go_modules/spec/dependabot/go_modules/update_checker/latest_version_finder_spec.rb
+++ b/go_modules/spec/dependabot/go_modules/update_checker/latest_version_finder_spec.rb
@@ -71,7 +71,9 @@ RSpec.describe Dependabot::GoModules::UpdateChecker::LatestVersionFinder do
         let(:goprivate) { "" }
 
         it "returns the latest minor version for the dependency's current major version" do
-          expect(finder.latest_version).to eq(Dependabot::GoModules::Version.new("1.1.0"))
+          # v1.2.5 isn't a tag on the repo, but it exists in proxy.golang.org, so it
+          # gets listed by `go list` since goprivate (and $GOPROXY) are blank.
+          expect(finder.latest_version).to eq(Dependabot::GoModules::Version.new("1.2.5"))
         end
       end
 
@@ -132,8 +134,11 @@ RSpec.describe Dependabot::GoModules::UpdateChecker::LatestVersionFinder do
     context "for a Git pseudo-version with releases available" do
       let(:dependency_version) { "1.0.0-20181018214848-ab544413d0d3" }
 
-      it "doesn't return the releases, currently" do
-        expect(finder.latest_version).to eq(dependency_version)
+      it "returns the latest pre-release" do
+        # Since a pseudo-version is always a pre-release, those aren't filtered.
+        # Here there was the choice to go to v1.1.0 or v1.2.0 pre-release, and it chose
+        # the largest since being on a pre-release means all pre-releases are considered.
+        expect(finder.latest_version).to eq(Dependabot::GoModules::Version.new("1.2.0-pre2"))
       end
     end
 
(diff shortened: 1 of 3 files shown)


jakecoffman:
Actually it's not the `latest_version`'s job to filter out versions that are lower, so I think this is actually the correct behavior.

deivid-rodriguez:
Just run into a similar failure on an unrelated PR: https://github.com/dependabot/dependabot-core/pull/6714, so at least that test is probably flaky and the failure unrelated to this changeset?

jakecoffman:
@deivid-rodriguez Ah, I was wondering if it was relying on some value being cached in https://proxy.golang.org/ and my messing with it broke things. So that confirms it. We should probably just remove it.

deivid-rodriguez:
You mean completely remove the test? I don't know how important this particular test is but in principle I'd rather find the root cause of the flakiness and fix it.

jakecoffman:
The important bit is we're setting the environment variable, not necessarily testing what value the proxy currently holds. So I'm looking at mocking the SharedHelper to assert the environment is correct and not check the version that's returned.

go_modules/spec/dependabot/go_modules/version_spec.rb
```diff
@@ -112,4 +112,20 @@
...
+      expect(described_class.new("v1.0.0-20231231120000-abcdefabcdef")).to be < described_class.new("v1.0.0")
+    end
+
+    it "sorts pre-release versions correctly" do
+      expect(described_class.new("v1.0.0-pre2.0.20231231120000-abcdefabcdef")).to be >
+                                                                                  described_class.new("v1.0.0-pre2")
+    end
+
+    it "sorts minor versions correctly" do
+      expect(described_class.new("v1.0.1-20231231120000-abcdefabcdef")).to be < described_class.new("v1.0.1")
```
mctofu, line 128:
```suggestion
      expect(described_class.new("v1.0.1-0.20231231120000-abcdefabcdef")).to be < described_class.new("v1.0.1")
```

Not positive here but from the doc I think we'd expect a `-0` in there?

> vX.Y.(Z+1)-0.yyyymmddhhmmss-abcdefabcdef is used when the base version is a release version like vX.Y.Z. For example, if the base version is v1.2.3, a pseudo-version might be v1.2.4-0.20191109021931-daa7c04131f5.

go_modules/spec/dependabot/go_modules/update_checker/latest_version_finder_spec.rb
```diff
@@ -144,16 +144,19 @@ module foobar
     context "for a Git pseudo-version with releases available" do
       let(:dependency_version) { "1.0.0-20181018214848-ab544413d0d3" }
 
-      it "doesn't return the releases, currently" do
-        expect(finder.latest_version).to eq(dependency_version)
+      it "returns the latest pre-release" do
```
mctofu, line 147: Could we add a test where it upgrades to a release (`1.2.0`)? I think that will be a common use case when an older dependency transitions to using modules.

mctofu (approved):

go_modules/spec/dependabot/go_modules/update_checker/latest_version_finder_spec.rb
jakecoffman: I added some more realistic tests for this case, both version updates and security updates, based on golang.org/x/net.

PR #6714, opened 2023-02-22
Strip BOM from file contents

### LINKED ISSUE #5118: Dependencies not found in Docker when Dockerfile uses UTF-8 BOM encoding
**Package ecosystem**
docker

**Package manager version**
n/a

**Language version**
n/a

**Manifest location and content before the Dependabot update**
[Dockerfile1](https://github.com/jblazek/dependabot-test/blob/main/Dockerfile1)
[Dockerfile2](https://github.com/jblazek/dependabot-test/blob/main/Dockerfile2)

**dependabot.yml content**

https://github.com/jblazek/dependabot-test/blob/main/.github/dependabot.yaml

**Updated dependency**
not applicable, updates not recognized.

When using docker package manager, I noticed one repo Dockerfile was not updating while others were. I expect to see _Checking if dotnet/aspnet 5.0.7-alpine3.13 needs updating_, but did not and just saw _finished job processing_ 

Upon investigation, the file which was not being updated was using UTF-8 BOM encoding while others which were getting PR updates were just using UTF-8.

n/a

_Dockerfile1_

```text
updater | INFO <job_362848288> Starting job processing
updater | INFO <job_362848288> Starting update job for jblazek/dependabot-test
updater | INFO <job_362848288> Checking if dotnet/aspnet 5.0.7-alpine3.13 needs updating

...other log entries checking dependency version...

updater | INFO <job_362848288> Finished job processing
```

_Dockerfile2_

```text
updater | INFO <job_362848288> Starting job processing

...other log entries, no dependency is detected...

updater | INFO <job_362848288> Finished job processing
```

PR for file not using BOM encoding: https://github.com/jblazek/dependabot-test/pull/1

Smallest manifest is [Dockerfile2](https://github.com/jblazek/dependabot-test/blob/main/Dockerfile2), only a FROM directive is needed and save the file with UTF-8 BOM encoding.

### LINKED ISSUE #6711: Dependabot can't parse your global.json with an UTF-8 BOM

### Package ecosystem

nuget

### Package manager version

_No response_

### Language version

.NET 7

[/global.json](https://github.com/Revolutionary-Games/RevolutionaryGamesCommon/blob/3e2973d1c72807e1bb17061295a26b2ba3191240/global.json)

### dependabot.yml content

https://github.com/Revolutionary-Games/RevolutionaryGamesCommon/blob/3e2973d1c72807e1bb17061295a26b2ba3191240/.github/dependabot.yml

### Updated dependency

_No response_

I expect dependabot to keep working even after adding a `global.json` file to specify which .NET SDK version I want to use with my project. Instead what happens is that dependabot fails with:
```
Dependabot can't parse your global.json

Dependabot failed to update your dependencies because there was an error parsing the global.json found at /global.json.

Dependabot encountered the following error:

Dependabot::DependencyFileNotParseable
```

And at least my eyes can't spot any of the error messages that are supposedly in the log that Github shows as extra details.

Update: a while later I realized the problem is that if there is an UTF-8 BOM in the file, *then* dependabot fails. So the failure reason is not the JSON data but the file starting with a BOM marker.

~~Experimenting randomly, I found out that if I remove the "allowPrerelease" part from my `global.json` file it starts working again. However, that should not be required as that is part of the `global.json` file specification since .NET Core 3.0: https://learn.microsoft.com/en-us/dotnet/core/tools/global-json~~

I have not seen any `dotnet` commands failing since adding my `global.json` file. So I think I got the syntax right. I tried switching `latestMajor` to `latestMinor` but that did not make dependabot work. Only after removing the `allowPrerelease` part did it suddenly start working.

_No response_

_No response_

deivid-rodriguez:
Fixes #5118.
Fixes #6711.

Redo of #5129.


deivid-rodriguez:
Need to a spec here, but this seems to be working great. Surprising to see at least three separate open issues affected by this.

deivid-rodriguez:
I added some specs, setting this as ready for a review!

PR #6715, opened 2023-02-22
Go: fix flaky test due to proxy remembering deleted tag

jakecoffman:
While messing with the Go fixtures on another test, the Go proxy suddenly started returning v1.2.5 as the latest version on https://github.com/dependabot-fixtures/go-modules-lib. That tag doesn't exist, so this test is failing. 

To make this less flaky, I hooked up a spy to ensure the GOPRIVATE environment variable is being set correctly rather than inferring it from the version returned.


jakecoffman:
I should note: I'm assuming v1.2.5 is a deleted tag, I didn't delete any tags, it's just not there. So it must have been deleted at some point, or something else is going on with the proxy?

honeyankit (approved):

PR #6719, opened 2023-02-23
Use latest Python 3.10 and 3.11 as of 2023-02-23

phillipuniverse:
Updates default versions to 3.10.10 and 3.11.2. Things moved around a bit since I opened #6375 and #6404 but I think these changes cover it.


deivid-rodriguez (approved):
Thanks!

PR #6721, opened 2023-02-24
fix Go "no longer vulnerable" due to bad SHA detection

jakecoffman:
I noticed one of my personal Go repos was still not getting the security update for golang.org/x/net after #6713. 

The issue is that the GitCommit checker is identifying some Go pseudo-versions as Git SHAs (like `v0.0.0-20180826012351-8a410e7b638d`) but others are not Git SHAs (like `v0.0.0-20201021035429-f5854403a974`). There must be a bug in this code identifying dependency sources as git: 

https://github.com/dependabot/dependabot-core/blob/cddd02aa8b8dc44ab0580b139c8ceafb45e97aa2/go_modules/lib/dependabot/go_modules/file_parser.rb#L38-L42

That causes `checker.vulnerable?` to return false here: 

https://github.com/dependabot/dependabot-core/blob/7196bb00a2e52d0d2ed167d1964ddab7e88dc1fe/common/lib/dependabot/update_checkers/base.rb#L137-L138

Which then leads to the "no longer vulnerable" message here:

https://github.com/dependabot/dependabot-core/blob/85342df046ed66fa6258d4002e7ec6b209e27918/updater/lib/dependabot/updater.rb#L220-L228

In Go, a Git dependency can't appear in `go.mod`, it's always a pseudo-version which is comparable, so the comment above 

> Can't (currently) detect whether git dependencies are vulnerable

doesn't apply to Go.

By returning false in `existing_version_is_sha?` we ensure the updater code will not assume the dependency is no longer vulnerable due to being on a Go pseudo-version.

I think there's also the case to remove all of the "git source" code in file_parser.rb since, again, Git dependencies in Go are always comparable. But I'll leave that for another day!

### DIFF
--- a/go_modules/lib/dependabot/go_modules/update_checker.rb
+++ b/go_modules/lib/dependabot/go_modules/update_checker.rb
@@ -84,7 +84,6 @@ module Dependabot
       def default_source
         { type: "default", source: dependency.name }
       end
-
     end
   end
 end
--- a/go_modules/lib/dependabot/go_modules/update_checker.rb
+++ b/go_modules/lib/dependabot/go_modules/update_checker.rb
@@ -68,11 +68,9 @@ module Dependabot
         raise NotImplementedError
       end
 
-      # Override the base class's check for whether this is a git dependency,
-      # since not all dep git dependencies have a SHA version (sometimes their
-      # version is the tag)
+      # Go only supports semver and semver-compliant pseudo-versions, so it can't be a SHA.
       def existing_version_is_sha?
-        git_dependency?
+        false
       end
 
       def version_from_tag(tag)
@@ -83,23 +81,10 @@ module Dependabot
         tag&.fetch(:tag)
       end
 
-      def git_dependency?
-        git_commit_checker.git_dependency?
-      end
-
       def default_source
         { type: "default", source: dependency.name }
       end
 
-      def git_commit_checker
-        @git_commit_checker ||=
-          GitCommitChecker.new(
-            dependency: dependency,
-            credentials: credentials,
-            ignored_versions: ignored_versions,
-            raise_on_ignored: raise_on_ignored
-          )
-      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


go_modules/lib/dependabot/go_modules/update_checker.rb
```diff
@@ -83,23 +81,9 @@ def version_from_tag(tag)
         tag&.fetch(:tag)
       end
 
-      def git_dependency?
-        git_commit_checker.git_dependency?
```
deivid-rodriguez, line 87: The base class implements `existing_version_is_sha?`  by delegating to `git_dependency?`. Would it make more sense to remove `existing_version_is_sha?` and leave `git_dependency?` implemented as `false`?

I guess not technically correct since from your explanation, go _does_ have git dependencies, they just map to "standard versions". But just concerned that we could copy some code from another ecosystem in the future that has branching according to `git_dependency?`, and we unintentionally reintroduce this kind of bug.

ANyways, no big deal, this is great!

deivid-rodriguez (approved):

go_modules/lib/dependabot/go_modules/update_checker.rb
jakecoffman: Yeah it's confusing, it does have Git dependencies... but they aren't SHAs. Since they're comparable, the reasoning why we treated Git dependencies as pins in other ecosystems doesn't apply here. 

The base class doesn't use `git_dependency?` though? 

https://github.com/dependabot/dependabot-core/blob/7196bb00a2e52d0d2ed167d1964ddab7e88dc1fe/common/lib/dependabot/update_checkers/base.rb#L207-L211

I did consider removing `git_dependency?` completely since the base class implementation will always be false, but I liked putting a comment above the method. 🤷

deivid-rodriguez: Disregard my previous comment 😳. I greped for the base class implementation, and I confused the previous implementation in the go ecosystem, with the base class implementation 😆

jeffwidman:
> There must be a bug in this code identifying dependency sources as git:
> https://github.com/dependabot/dependabot-core/blob/cddd02aa8b8dc44ab0580b139c8ceafb45e97aa2/go_modules/lib/dependabot/go_modules/file_parser.rb#L38-L42

To close the loop, no need to investigate the bug since the code is simply removed in https://github.com/dependabot/dependabot-core/pull/6723/files#diff-5473cd4fd4f2f0f6b13968468fc3a0109a7769f9c729e4ad36a03ac8f419a660R36

PR #6723, opened 2023-02-24
Go: remove git-type sources since pseudo-versions are comparable

### LINKED ISSUE #4448: Cleanup duplicate code by standardizing on one way - maybe eliminate native helper for `go`?

jakecoffman:
In #6713 and #6721 I started pulling the thread around Go pseudo-versions, and in this PR I complete the motion!

In Go Modules we can have a dependency like `github.com/mattn/go-isatty v0.0.4` and we can have a dependency like `github.com/mattn/go-isatty v0.0.4-0.20171107045240-f4b977dc7396`. To Dependabot, these are both valid semvers and can be compared. Previously we treated the pseudo-version like it was pinned to a SHA. In other ecosystems we've not touched SHA dependencies because they aren't comparable, and figuring out where they fall in a tree is non-trivial. 

In Go we don't have that problem, so there's no reason to designate a source as a Git type. We can compare the versions directly.

However, it is also possible to use vanity URLs like `golang.org/x/text`. This isn't a problem until we get to the metadata gathering. We need to be able to gather tags, and release notes, and golang.org/x/text doesn't have those. The Go tooling makes a request to `https://golang.org/x/text?go-get=1` which responds with some `<meta>` tags that have a link to `github.com`. 

That is also what Dependabot does, but previously it was happening in the file_parser (for "git sources"), and it was trickling down through the source's `url` to the MetadataFinder. So I've hooked up the native helper in the MetadataFinder code, and removed the Ruby implemented version. I think it's better to use the native version, and the code comments agree! 

Fix #4448


jakecoffman:
I'm not in a rush to ship this because it doesn't really fix anything, more of a code-cleanup.

mctofu (approved):

PR #6729, opened 2023-02-24
Inline updater specs

landongrindheim:
### 🖼️ Context
While trying to make changes to the updater logic, it's become apparent that our tests have become difficult to work with. This is a first pass at inlining setup so that tests can be reasoned about as individual units.

The approach here is generally:
* Remove `let` statements which are not overridden
* Stub calls to `Dependabot.logger`
* Exract setup of `job`, `service` and `updater` objects into callable methods
* Use :point_up: to bring setup into individual specs

### DIFF
--- a/updater/spec/dependabot/updater_spec.rb
+++ b/updater/spec/dependabot/updater_spec.rb
@@ -21,8 +21,6 @@ RSpec.describe Dependabot::Updater do
     )
   end
 
-  let(:logger) { double(Logger) }
-
   def service
     @service ||=
       instance_double(
@@ -38,10 +36,9 @@ RSpec.describe Dependabot::Updater do
   end
 
   before do
+    allow(Dependabot.logger).to receive(:info)
+    allow(Dependabot.logger).to receive(:error)
     allow_any_instance_of(Dependabot::ApiClient).to receive(:record_package_manager_version)
-    allow(Dependabot).to receive(:logger).and_return(logger)
-    allow(logger).to receive(:info)
-    allow(logger).to receive(:error)
 
     allow(Dependabot::Environment).to receive(:token).and_return("some_token")
     allow(Dependabot::Environment).to receive(:job_id).and_return(1)
@@ -259,10 +256,10 @@ RSpec.describe Dependabot::Updater do
     end
 
     it "logs the current and latest versions" do
-      expect(logger).
+      expect(Dependabot.logger).
         to receive(:info).
         with("<job_1> Checking if dummy-pkg-b 1.1.0 needs updating")
-      expect(logger).
+      expect(Dependabot.logger).
         to receive(:info).
         with("<job_1> Latest version is 1.2.0")
       updater.run
@@ -276,10 +273,10 @@ RSpec.describe Dependabot::Updater do
       end
 
       it "logs the update requirements and strategy" do
-        expect(logger).
+        expect(Dependabot.logger).
           to receive(:info).
           with("<job_1> Requirements to unlock own")
-        expect(logger).
+        expect(Dependabot.logger).
           to receive(:info).
           with("<job_1> Requirements update strategy bump_versions")
         updater.run
@@ -290,7 +287,7 @@ RSpec.describe Dependabot::Updater do
       let(:allowed_updates) { [{ "dependency-name" => "typoed-dep-name" }] }
 
       it "logs the current and latest versions" do
-        expect(logger).
+        expect(Dependabot.logger).
           to receive(:info).
           with("<job_1> Found no dependencies to update after filtering " \
                "allowed updates")
@@ -347,7 +344,7 @@ RSpec.describe Dependabot::Updater do
               }
             }
           )
-          expect(logger).
+          expect(Dependabot.logger).
             to receive(:info).with(
               "<job_1> Dependabot can't update vulnerable dependencies for " \
               "projects without a lockfile or pinned version requirement as " \
@@ -423,7 +420,7 @@ RSpec.describe Dependabot::Updater do
               }
             }
           )
-          expect(logger).
+          expect(Dependabot.logger).
             to receive(:info).with(
               "<job_1> The latest possible version that can be installed is " \
               "1.2.0 because of the following conflicting dependency:\n" \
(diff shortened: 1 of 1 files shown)


updater/spec/dependabot/updater_spec.rb
```diff
@@ -10,77 +10,97 @@
...
-      job: job,
-      dependency_files: dependency_files,
-      base_commit_sha: "sha",
-      repo_contents_path: repo_contents_path
-    )
+  # rubocop:disable Metrics/MethodLength
+  def build_job(requested_dependencies: nil, allowed_updates: default_allowed_updates,
+                existing_pull_requests: [], ignore_conditions: [], security_advisories: [],
+                experiments: {}, updating_a_pull_request: false, security_updates_only: false)
+    @build_job ||=
```
deivid-rodriguez, line 17: I think this memoization doesn't make sense since the method has parameters, so multiple invocations will be ignored in favor of the first one. I'd remove it.

updater/spec/dependabot/updater_spec.rb
```diff
@@ -10,77 +10,97 @@
...
+          "prefix-development" => "[bump-dev]",
+          "include-scope" => true
+        },
+        security_updates_only: security_updates_only
+      )
+  end
+  # rubocop:enable Metrics/MethodLength
+
+  def build_updater(service: build_service, job: build_job, dependency_files: default_dependency_files)
+    @build_updater ||=
```
deivid-rodriguez, line 63: Same issue with memoizing the result.

updater/spec/dependabot/updater_spec.rb
```diff
@@ -10,77 +10,97 @@
...
-    allow(service).to receive(:record_update_job_error)
-    allow_any_instance_of(Dependabot::ApiClient).to receive(:record_package_manager_version)
-    allow(Dependabot).to receive(:logger).and_return(logger)
-    allow(logger).to receive(:info)
-    allow(logger).to receive(:error)
-
-    allow(Dependabot::Environment).to receive(:token).and_return("some_token")
-    allow(Dependabot::Environment).to receive(:job_id).and_return(1)
+  def build_service(job: build_job)
+    @build_service ||=
```
deivid-rodriguez, line 75: And here too!

deivid-rodriguez (commented):
Nice, I'm good with this! Just left some comments about some memoizations.

updater/spec/dependabot/updater_spec.rb
landongrindheim: Good call! That was a relic of refactoring that I became blind to 😅

Rubocop was enforcing the name, which is really awkward (IMO), so it's extra great to get rid of this 😄

updater/spec/dependabot/updater_spec.rb
landongrindheim: Addressed in [9f50771](https://github.com/dependabot/dependabot-core/pull/6729/commits/9f50771aaaf4e3bf8478d5b6dc199359b60c3284)

updater/spec/dependabot/updater_spec.rb
landongrindheim: Addressed in [9f50771](https://github.com/dependabot/dependabot-core/pull/6729/commits/9f50771aaaf4e3bf8478d5b6dc199359b60c3284)

updater/spec/dependabot/updater_spec.rb
landongrindheim: Addressed in [9f50771](https://github.com/dependabot/dependabot-core/pull/6729/commits/9f50771aaaf4e3bf8478d5b6dc199359b60c3284)

deivid-rodriguez (commented):
I had a look at the FIXMEs about errors supposed to be raised, and they do seem correct to me? 

The updater does not raise errors but reports them to Sentry/Backend instead and that's what the specs seem to be checking. The setup is a bit cumbersome through stubbing `checker#can_update?` to raise the different errors but I'd say it's working fine.

landongrindheim:
> I had a look at the FIXMEs about errors supposed to be raised, and they do seem correct to me?
> 
> The updater does not raise errors but reports them to Sentry/Backend instead and that's what the specs seem to be checking. The setup is a bit cumbersome through stubbing `checker#can_update?` to raise the different errors but I'd say it's working fine.

Ah, you're right! I had mistakenly thought that we were relying on [this line](https://github.com/dependabot/dependabot-core/pull/6729/files#diff-f4781b2315ce8443ff37419b75b5af931a59345bd047ffed4317435ab9f673f0L1384) to raise errors, and that was happening in a separate context. But it's happening [in a before block](https://github.com/dependabot/dependabot-core/pull/6729/files#diff-f4781b2315ce8443ff37419b75b5af931a59345bd047ffed4317435ab9f673f0R1757) via `raise error`.

I'll go ahead and remove those FIXMEs 😄 

---

Separately, I left two FIXMEs because specs were failing due to the file mode being changed for me locally. That doesn't seem to affect CI. Do you know why that happens? I can remove those separately if it's expected behavior.

deivid-rodriguez (approved):
Nice, these tests look much easier to work with now!

bdragon (approved):
This is so much more readable. Thank you for taking this on 👏

brrygrdn (approved):

updater/spec/dependabot/updater_spec.rb
```diff
@@ -1512,18 +1970,30 @@ def expect_update_checker_with_ignored_versions(versions)
...
           updater.run
         end
 
         it "tells the main backend" do
+          job = build_job
+          service = build_service(job: job)
+          updater = build_updater(service: service, job: job)
+
+          # allow(updater).to receive(:dependency_files).and_raise(error)
+
```
deivid-rodriguez, line 1988: I just noticed this stray comment that's also probably a leftover of the previous FIXMEs that we ended removing?

updater/spec/dependabot/updater_spec.rb
```diff
@@ -1492,18 +1938,30 @@ def expect_update_checker_with_ignored_versions(versions)
...
```
deivid-rodriguez, line 1956: Another one of these.

deivid-rodriguez (approved):
@landongrindheim I just noticed two more leftovers from the previous FIXME notes that we ended up removing. Other than that please merge away!

PR #6743, opened 2023-02-27
Don't ever update GitHub Actions unless they are actually pinned 

### LINKED ISSUE #6739: Github Actions: Unexpected change of branch
### Package ecosystem

Github Actions

### Updated dependency

dtolnay/rust-toolchain

Dependabot sent a PR to change `dtolnay/rust-toolchain@stable` to `dtolnay/rust-toolchain@master` in a Github Actions workflow. These are two different branches doing different things and this PR would break the workflow.
The same setup worked fine for many months now, so it must be a recent regression, presumably introduced in https://github.com/dependabot/dependabot-core/pull/6082.

cc @deivid-rodriguez

https://github.com/rust-lang/cargo-bisect-rustc/pull/259

deivid-rodriguez:
The change in Dependency combine logic at #6082 surfaced a bug where if a Github Action is referencing two different branches in the same workflow, Dependabot would create a PR unifying it to a single branch, while is not what the user wants.

The fix is to ignore any `uses:` references which include a reference that should not be updated, like a branch name not named like a version (like `master` or `stable`).

In addition to fixing this bug, this change also allows Dependabot to create PRs for mixed situations like the following, where the version is updated and the others are left alone:

```yaml
uses: dtolnay/rust-toolchain@master
# ...
uses: dtolnay/rust-toolchain@stable
# ...
uses: dtolnay/rust-toolchain@1.42
```

or the following, where the SHA is updated and the others are left alone:

```yaml
uses: dtolnay/rust-toolchain@master
# ...
uses: dtolnay/rust-toolchain@stable
# ...
uses: dtolnay/rust-toolchain@df6c627e2064a425669c08e51ed613bb3f013c69
```

There's a change in behavior for other mixed cases like

```yaml
uses: dtolnay/rust-toolchain@master
# ...
uses: dtolnay/rust-toolchain@1.42
# ...
uses: dtolnay/rust-toolchain@df6c627e2064a425669c08e51ed613bb3f013c69
```

Before Dependabot would find no updates, while now it will update both the SHA-pinned and the version-pinned lines to the latest _version_. I'm unclear about why users would pin some workflows and not others, so I'm fine with making this change and listen to any users that don't actually want this. My expectation is that this will help users catch situations where they were unintentionally using multiple styles and help them unify those.

Fixes #6739.

### DIFF
--- a/github_actions/lib/dependabot/github_actions/file_parser.rb
+++ b/github_actions/lib/dependabot/github_actions/file_parser.rb
@@ -46,7 +46,13 @@ module Dependabot
 
         uses_strings.each do |string|
           # TODO: Support Docker references and path references
-          dependency_set << build_github_dependency(file, string) if string.match?(GITHUB_REPO_REFERENCE)
+          next unless string.match?(GITHUB_REPO_REFERENCE)
+
+          dep = build_github_dependency(file, string)
+          git_checker = Dependabot::GitCommitChecker.new(dependency: dep, credentials: credentials)
+          next unless git_checker.pinned?
+
+          dependency_set << dep
         end
 
         dependency_set
--- a/github_actions/spec/dependabot/github_actions/file_parser_spec.rb
+++ b/github_actions/spec/dependabot/github_actions/file_parser_spec.rb
@@ -30,8 +30,24 @@ RSpec.describe Dependabot::GithubActions::FileParser do
     )
   end
 
+  def mock_service_pack_request(nwo)
+    stub_request(:get, "https://github.com/#{nwo}.git/info/refs?service=git-upload-pack").
+      to_return(
+        status: 200,
+        body: fixture("git", "upload_packs", "checkout"),
+        headers: {
+          "content-type" => "application/x-git-upload-pack-advertisement"
+        }
+      )
+  end
+
   describe "parse" do
     subject(:dependencies) { parser.parse }
+
+    before do
+      mock_service_pack_request("actions/checkout")
+      mock_service_pack_request("actions/setup-node")
+    end
 
     its(:length) { is_expected.to eq(2) }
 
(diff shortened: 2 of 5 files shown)


mctofu (approved):
This makes sense to me. I've seen that come up in support a few times where Dependabot didn't update actions because of a stray reference to a branch in one workflow so this seems like an improvement. Do we have any test cases that cover mixed usage of branch/sha/version?

deivid-rodriguez:
I guess we don't have tests for those cases, since they were not working up until now and tests were still passing.

It'd be good to add those, but didn't want to delay the fix to this regression. Should I ship this, and then try add some more specs as a follow up?

LingMan:
There should be no update in this scenario:
```yaml
uses: dtolnay/rust-toolchain@master
# ...
uses: dtolnay/rust-toolchain@stable
# ...
uses: dtolnay/rust-toolchain@1.42
```
The `rust-toolchain` action uses branches like `1.42` to make available older compiler versions. Rust projects then usually test against both `stable` and against a minimal version they wish to support.

deivid-rodriguez:
> The rust-toolchain action uses branches like 1.42 to make available older compiler versions. Rust projects then usually test against both stable and against a minimal version they wish to support.

You mean that you'll never want any updates to a `uses: dtolnay/rust-toolchain@1.42`, or that you don't want updates when also included in combination with `uses: dtolnay/rust-toolchain@stable` or similar? Either way, proposing an update for an out of date version number is what Dependabot does with every other action, and seems like the right thing to do to me. If this particular action is special for your case, then you can always ignore updates for it.

LingMan:
There should never be an update to a `uses: dtolnay/rust-toolchain@1.42` line. The `1.42` (and also the `stable`) refers to the version of the compiler the action should provide. It does not refer to the version of the action itself (`master` is really the all that's available for this action).

Updating a `v1.42` to a `v1.50` (notice the extra `v`) would be fine for me. That matches the way actions are usually versioned. If the `v` is missing, please just treat it as a string and not as a version number.

Ignoring this particular action is of course possible, but it would be a big pain. Large parts of the Rust ecosystem use `dtolnay/rust-toolchain` and ensure compatibility with older compilers in this way.

deivid-rodriguez:
> There should never be an update to a uses: `dtolnay/rust-toolchain@1.42` line. The `1.42` (and also the `stable`) refers to the version of the compiler the action should provide. It does not refer to the version of the action itself (`master` is really the all that's available for this action).

> Updating a v1.42 to a v1.50 (notice the extra v) would be fine for me. That matches the way actions are usually versioned. If the v is missing, please just treat it as a string and not as a version number.

That deviates from how Dependabot works for every other action, so I don't think we will be changing this behavior. In any case, it would be a separate feature request. Dependabot has been creating PRs updating `@1.42` lines for a long time, it just didn't do it when `@stable` or `@master` were also used in the same workflow, due to some bugs.

LingMan:
Welp, looks like Dependabot and parts of `dtolnay/rust-toolchain`'s intended usage are then incompatible. Let's hope most projects test their minimally supported Rust version as part of a matrix and and pass it in over the `toolchain` input instead of relying on the branch name.

PR #6748, opened 2023-02-28
Don't propose updates to a more precise version of the same docker tag

### LINKED ISSUE #6746: Docker tag pointed to the same image causes the update

### Package ecosystem

Docker

### Package manager version

_No response_

### Language version

_No response_

/probe/Dockerfile

### dependabot.yml content

```
version: 2
updates:
  - package-ecosystem: 'gomod'
    directory: '/'
    schedule:
      interval: 'weekly'
  - package-ecosystem: 'gomod'
    directory: '/tools'
    schedule:
      interval: 'weekly'
  - package-ecosystem: "docker"
    directory: "/"
    schedule:
      interval: "weekly"
  - package-ecosystem: "docker"
    directory: "/probe"
    schedule:
      interval: "weekly"
```

### Updated dependency

From: `registry.access.redhat.com/ubi8/ubi-minimal:8.7`
To: `registry.access.redhat.com/ubi8/ubi-minimal:8.7-1085`

### Expected behavior:
No PR should be raised since tags are pointing to the same image. PR should be raised only if minor version is updated, i.e. `8.7` -> `8.8` (without hyphen).
### Actual behavior:
PR has been raised.

```
$ docker pull registry.access.redhat.com/ubi8/ubi-minimal:8.7
8.7: Pulling from ubi8/ubi-minimal
c7e856e03741: Pull complete 
Digest: sha256:dc06ba83c6f47fc0a2bca516a9b99f1cf8ef37331fd460f4ca55579a815ee6cb
Status: Downloaded newer image for registry.access.redhat.com/ubi8/ubi-minimal:8.7
registry.access.redhat.com/ubi8/ubi-minimal:8.7
$ docker pull registry.access.redhat.com/ubi8/ubi-minimal:8.7-1085
8.7-1085: Pulling from ubi8/ubi-minimal
Digest: sha256:dc06ba83c6f47fc0a2bca516a9b99f1cf8ef37331fd460f4ca55579a815ee6cb
Status: Downloaded newer image for registry.access.redhat.com/ubi8/ubi-minimal:8.7-1085
registry.access.redhat.com/ubi8/ubi-minimal:8.7-1085
```

stackrox/acs-fleet-manager#854

```
FROM registry.access.redhat.com/ubi8/ubi-minimal:8.7
```

deivid-rodriguez:
This change allows us to avoid unwanted PRs like the one reported at #6746.

I took the chance to clean things up a bit and extract a `Dependabot::Docker::Tag` class that holds information about a Docker tag and how it maps to numeric versions.

Fixes #6746.


docker/lib/dependabot/docker/update_checker.rb
```diff
@@ -105,12 +95,15 @@ def version_up_to_date?
       end
 
       def version_tag_up_to_date?(version)
```
jeffwidman, line 97: `version` --> `current_version`?? to disambiguate from `latest_version` inside of this method...

docker/lib/dependabot/docker/update_checker.rb
```diff
@@ -105,12 +95,15 @@ def version_up_to_date?
...
 
-        latest_version = fetch_latest_version(version)
+        latest_tag = fetch_latest_version(version_tag)
 
-        old_v = numeric_version_from(version)
-        latest_v = numeric_version_from(latest_version)
+        old_v = version_tag.numeric_version
+        latest_v = latest_tag.numeric_version
 
         version_class.new(latest_v) <= version_class.new(old_v)
```
jeffwidman, line 108: Maybe delete these two intermediate vars? I think it'd be fine to inline stuff into the next line:
```suggestion
        version_class.new(latest_tag.numeric_version) <= version_class.new(current_tag.numeric_version)
```

Note that my suggestion includes a renaming of `version_tag` to `current_tag` so won't work as-is...

docker/lib/dependabot/docker/update_checker.rb
```diff
@@ -163,69 +156,55 @@ def fetch_latest_version(version)
...
       end
 
-      def comparable_tags_from_registry(version)
-        original_prefix = prefix_of(version)
-        original_suffix = suffix_of(version)
-        original_format = format_of(version)
+      def comparable_tags_from_registry(original_tag)
+        original_prefix = original_tag.prefix
+        original_suffix = original_tag.suffix
+        original_format = original_tag.format
```
jeffwidman, line 161: These three intermediate vars look like they could all be deleted and actually improve readability through conciseness?

jeffwidman (approved):
Nice cleanup. Had a few trivial suggestions that are optional.

docker/lib/dependabot/docker/tag.rb
```diff
@@ -0,0 +1,73 @@
...
+      end
+
+      def canonical?
+        return false unless numeric_version
+        return true if name == numeric_version
+
+        # .NET tags are suffixed with -sdk
+        return true if name == numeric_version + "-sdk"
+
+        name == "jdk-" + numeric_version
```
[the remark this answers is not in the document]
deivid-rodriguez: Definitely weird! The code was already there and I wanted to focus on the refactoring of extracting a `Dependabot::Docker::Tag` class, so I decided to stop there and leave existing logic untouched.

But I can improve this as a follow, since I also just realized that we can remove the `-sdk` special case now since these images [have been removed](https://github.com/dotnet/dotnet-docker/issues/3052) 🎉

docker/lib/dependabot/docker/update_checker.rb
deivid-rodriguez: This didn't seem particularly confusing to me, I felt it was clear that `version` was the parameter to the method. So I decided not to touch this parameter name. I did change the ones that changed types to add `_tag` to avoid confusion.

docker/lib/dependabot/docker/update_checker.rb
deivid-rodriguez: Yeah, specially those `old_v` names seem unnecessary. I can do this as a followup PR since I will also probably memoize this values when first called to not run the same regex over an over again after inlining the code.

docker/lib/dependabot/docker/update_checker.rb
deivid-rodriguez: Same as with the other comment, I can do this as a follow up to not add more complexity to this PR.

docker/lib/dependabot/docker/update_checker.rb
jeffwidman: Makes sense. 👍 for shipping as-is, but let's leave these threads about future cleanups unresolved so they're easy to spot later when you come back for the follow-on.

deivid-rodriguez:
You're right. I think it's best to ship the fix ourselves now to not depend on upstream plans/fixes/improvements, but we can leave a `TODO` for a future logic removal, once we verify that this particular registry API implementation is finally providing proper digests that we can use?

deivid-rodriguez:
Another consideration is that last time I looked at Registry API v2 docs, providing content digests was optional, so maybe we can keep the logic regardless to satisfy registries that choose to not implement it.

jeffwidman:
Maybe just add a `NOTE` pointing to this discussion and to the Redhat announcment so that future devs have the relevant context if they're working on a bugfix of some kind in that section of code that they're aware they may be able to delete it altogether, and then call it good for now?

PR #6760, opened 2023-03-01
[Updater] Remove the EndToEndCommand, Refactor the integration test to cover more production code

brrygrdn:
See: https://github.com/dependabot/dependabot-core/pull/6572

Follows up on removing the binary that used the `EndToEndCommand` class in https://github.com/dependabot/dependabot-core/pull/6752 by removing the class itself while retaining the integration tests by refactoring them to use the `FileFetcherCommand` and `UpdateFilesCommand` back-to-back as functionally equivalent.

### Rationale

This PR has two goals:
- Remove the `EndToEndCommand` as dead code which we don't need to maintain or update to incorporate grouped updates
- Maintain the existing integration test against as much of a cross-section of the Updater to provide a regression test for deeper refactors of the code
  - While we now have an end to end smoke test suite that is a more representative regression and acceptance suite for the Updater, it exists outside the 'inner loop' and this test still has value as a fast local validation check **if** it tests the right things

### Changes

This pass makes no behavioural changes to the implementation of the Updater, nor does it change the inputs or expectations of the integration test, I've kept everything focused on modifying the test setup:

- The integration test now runs `Dependabot::FileFetcherCommand.new.run` and `Dependabot::UpdateFilesCommand.new.run` back to back instead of `Dependabot::EndToEndCommand.new.run`
- All "self-stubs" on the Dependabot::EndToEndCommand instance are removed in preference of stubbing the `Dependabot::Environment` as a proxy for reading input from the file system / envvars
- To support this, the mechanics of reading the "job_definition" from the file system have been moved unto `Dependabot::Environment`
  - This is somewhat opinionated, but I think it is a fairly common idiom in several frameworks to use an "environment" object to allow portability between CLI / Service interfaces and handle context-specific concerns
- The `Dependabot::Environment::job_definition` stub is used to pass in job parameters as hashes, as well as to inject base64 encoded `Dependabot::DependencyFile` so it covers input validation and deserialisation in both the fetch and update steps


jeffwidman:
This makes sense as a top-level lift... one of the things that came out of investigating https://github.com/dependabot/dependabot-core/pull/6572 is several folks pointing out many of the integration tests were redundant now that we have smoke... It might be worth taking a pass through those to delete any integration tests that are obviously superfluous so we don't have to maintain that code down the road? Albeit in a follow-on PR, not in this PR.

brrygrdn:
> one of the things that came out of investigating https://github.com/dependabot/dependabot-core/pull/6572 is several folks pointing out many of the integration tests were redundant now that we have smoke

I think that's possible, these integration tests are also something of a misnomer as they stub out large sections of real behaviour which undermines their value.

As part of lining up to add grouped updates, I'm mostly concerned about cutting away dead code and leaving the tests in situ just to confirm anything we do doesn't modify behaviour without getting into the fine grain detail of which tests are wasted yet, similar to the approach in #6729.

I'd rather shunt deleting tests until we've gotten a little further in the transform and have more confidence about what we can remove.

jeffwidman:
Makes sense, I'll switch https://github.com/dependabot/dependabot-core/pull/6572 to draft since it's got some context from _mctofu on which tests probably shouldn't be removed and then down the road a bit can take another look at it after rebasing on top of these changes...

brrygrdn:
> Makes sense, I'll switch https://github.com/dependabot/dependabot-core/pull/6572 to draft since it's got some context from _mctofu on which tests probably shouldn't be removed and then down the road a bit can take another look at it after rebasing on top of these changes...

As I surfaced on the last few passes on this, the question of the value of the integration test vs the smoke tests is something I turned around on. I think right now, I value having this as a quick regression test I can run without leaving the context of a Ruby workflow.

I'm not sure we need it long term, so I'll be mindful about cleaning it more than I have, but it is proving useful as a forcing function to try to shove things into clear inputs/outputs. I suspect the right answer is we keep fleshing out the unit tests to the point we feel they are readable and thorough and this kind of integration test adds little to no value - we have too much daylight between the smoke tests and our existing units right now I suspect 🤔

jeffwidman:
Yeah, I definitely see the value of a few integration tests that are locally available within ruby dev flow... but shouldn't need more than a handful to get the gist of "is this wired up right across class boundaries?" IE, no need to extensive edge case integration tests...

updater/lib/dependabot/environment.rb
```diff
@@ -31,6 +31,10 @@ def self.github_actions?
       @github_actions ||= environment_variable("GITHUB_ACTIONS", false)
     end
 
+    def self.job_definition
+      @job_definition ||= JSON.parse(File.read(job_path))
+    end
```
landongrindheim, line 36: 💡 Not sure if this would be supported by our current code, but I think it'd be a big benefit to deep freeze this object.

landongrindheim (approved):
I'd love to see someone more familiar with this code add their review, but this looks good to me 😄

jakecoffman (approved):
Makes sense and LGTM!

updater/lib/dependabot/environment.rb
brrygrdn: Yeah, that's a fair call out. I'd like to find a more durable home for the read from/write to file methods in general, so I'll keep this in mind.

updater/lib/dependabot/file_fetcher_command.rb
```diff
@@ -7,6 +7,11 @@
 
 module Dependabot
   class FileFetcherCommand < BaseCommand
+    # BaseCommand does not implement this method, so we should expose
+    # the instance variable for error handling to avoid raising a
+    # NotImplementError if it is referenced
```
jeffwidman, line 12:
```suggestion
    # NotImplementedError if it is referenced
```

jeffwidman (approved):
I wasted two hours confused by EndToEndJob a while back, so I'm glad to see this source of confusion getting eliminated...

PR #6761, opened 2023-03-01
Include summary of errors at the end of job output

mctofu:
This should make it a little easier to diagnose failures without needing to also consult the Dependabot error UIs. It's also handy for figuring out which dependencies failed when theres a long list of dependencies to update.

Adds two potential summaries at the end of the job:

For errors that aren't specific to a dependency:
```
     +--------------------+
     |    Errors          |
     +--------------------+
     | job_repo_not_found |
     +--------------------+
```

For errors updating a particular dependency:
```
+-------------------------------------+
|    Dependencies failed to update    |
+---------------------+---------------+
| best_dependency_yay | unknown_error |
+---------------------+---------------+
```

### DIFF
--- a/updater/lib/dependabot/service.rb
+++ b/updater/lib/dependabot/service.rb
@@ -95,7 +95,7 @@ module Dependabot
     # +--------------------+
     def job_error_type_summary
       job_error_types = errors.filter_map { |error_type, dependency| [error_type] if dependency.nil? }
-      return unless job_error_types.any?
+      return if job_error_types.none?
 
       Terminal::Table.new do |t|
         t.title = "Errors"
@@ -114,7 +114,7 @@ module Dependabot
       dependency_errors = errors.filter_map do |error_type, dependency|
         [dependency.name, error_type] unless dependency.nil?
       end
-      return unless dependency_errors.any?
+      return if dependency_errors.none?
 
       Terminal::Table.new do |t|
         t.title = "Dependencies failed to update"
--- a/updater/lib/dependabot/service.rb
+++ b/updater/lib/dependabot/service.rb
@@ -34,8 +34,8 @@ module Dependabot
       @pull_requests << [dependency_name, "closed: #{reason}"]
     end
 
-    def record_update_job_error(job_id, error_type:, error_details:)
-      @errors << error_type.to_s
+    def record_update_job_error(job_id, error_type:, error_details:, dependency: nil)
+      @errors << [error_type.to_s, dependency]
       client.record_update_job_error(job_id, error_type: error_type, error_details: error_details)
     end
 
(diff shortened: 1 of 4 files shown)


jakecoffman (approved):

updater/lib/dependabot/service.rb
```diff
@@ -84,6 +86,42 @@ def error_summary
...
+    # +-------------------------------------+
+    # |    Dependencies failed to update    |
+    # +---------------------+---------------+
+    # | best_dependency_yay | unknown_error |
+    # +---------------------+---------------+
+    def dependency_error_summary
+      dependency_errors = errors.filter_map do |error_type, dependency|
+        [dependency.name, error_type] unless dependency.nil?
+      end
+      return unless dependency_errors.any?
```
landongrindheim, line 117: 💅 Personal preference: _I_ find `unless` harder to think about. Another option would be to use `#none?` with `if` 👇

```suggestion
      return if dependency_errors.none?
```

If you see this similarly, and feel like making the change, `return unless job_error_types.any?` could use the same approach 😄

updater/spec/dependabot/service_spec.rb
```diff
@@ -268,6 +296,27 @@
...
+      it "includes an error count" do
+        expect(service.summary).
+          to include("Dependabot encountered '1' error(s) during execution")
+      end
+
+      it "includes an error summary" do
+        expect(service.summary).
+          to include("unknown_error")
+        expect(service.summary).
+          to include("dependabot-cobol")
```
landongrindheim, line 318: 🔎 Are you working on adding a new ecosystem? 😉

updater/spec/dependabot/service_spec.rb
```diff
@@ -302,7 +352,16 @@
...
+          to include("Dependabot encountered '2' error(s) during execution")
+      end
+
+      it "includes an error summary" do
+        expect(service.summary).
+          to include("epoch_error")
+        expect(service.summary).
+          to include("unknown_error")
+        expect(service.summary).
+          to include("dependabot-fortran")
```
landongrindheim, line 364: 😱 *ANOTHER?*

landongrindheim (approved):

updater/lib/dependabot/service.rb
```diff
@@ -84,6 +86,42 @@ def error_summary
...
+    # Example output:
+    #
+    # +--------------------+
+    # |    Errors          |
+    # +--------------------+
+    # | job_repo_not_found |
+    # +--------------------+
+    def job_error_type_summary
+      job_error_types = errors.filter_map { |error_type, dependency| [error_type] if dependency.nil? }
+      return unless job_error_types.any?
```
mctofu, line 98:
```suggestion
      return if job_error_types.none?
```

jeffwidman (approved):
👍 this is huge, great idea

PR #6763, opened 2023-03-02
Go: fix NilClass error when directory doesn't have a go.mod file

jakecoffman:
Fixes NoMethodError error introduced just now in #6658

```
NoMethodError
undefined method `content' for nil:NilClass

            "gomod" => go_mod.content.match(/^go\s(\d+\.\d+)/)&.captures&.first || "unknown"
```

This is being caused by misconfigured dependabot.yml where `go.mod` is not present in the directory. I'm kind of surprised we don't raise an exception sooner since it's listed in `required_files_in?`.

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_fetcher.rb
+++ b/go_modules/lib/dependabot/go_modules/file_fetcher.rb
@@ -15,6 +15,8 @@ module Dependabot
       end
 
       def package_manager_version
+        return nil unless go_mod
+
         {
           ecosystem: "gomod",
           package_managers: {


go_modules/lib/dependabot/go_modules/file_fetcher.rb
```diff
@@ -15,6 +15,8 @@ def self.required_files_message
       end
 
       def package_manager_version
+        return nil unless go_mod
```
landongrindheim, line 18: Verified that we're handling `nil` in [the FileFetcherCommand](https://github.com/dependabot/dependabot-core/blob/5e1be21dbffd749203af347478e6d5a15a180181/updater/lib/dependabot/file_fetcher_command.rb#L20)

mctofu (approved):

landongrindheim (approved):

PR #6769, opened 2023-03-02
Make setup local to individual specs

landongrindheim:
### 🖼️ Context

This represents another set of movements toward inlining spec setup so that specs are self-contained. The intention is that specs will be duplicatable/modifiable without affecting surrounding specs.

The majority of this change is concerned with the `UpdateChecker`, which had been aggressively stubbed in order to get the behavior we want.

Things which have not been done _yet_, which should be possible now:
- [ ] Removing deep context nesting
- [x] Re-ordering methods and examples to improve flow/readability


brrygrdn (approved):
Looks good to me!

I appreciate moving the test object and stubbing into single methods a lot 👍🏻

PR #6770, opened 2023-03-02
[Updater] Remove `job_id` from the Service/ApiClient's method signatures

brrygrdn:
I'm in the process of rewriting the Service/ApiClient's methods to accept a value class rather than a set of arguments so it is easier to slot in a "GroupRule" as a new optional argument without having to modify the whole way up and down the call chain as I iterate on it.

Unfortunately, this means modifying the whole way up and down the call chain one last time! 😅 

Since I'm doing this, I'm going to cross off a bit of technical debt - we currently instantiate the `ApiClient` with the Job's token, but then pass the Job's ID on every method call.

This makes our stubbing and method signatures very verbose and since a token is scoped to a Job ID, it creates the impression the client could be used to reach across jobs within a single context when it can't.

Let's clean this up and consolidate the id/token as a matched credential pair.

### Rationale

This is a noisy change so I've pulled it out as a separate PR to fly ahead of the rest of my changes to make it easier to review/test in isolation.

### Other changes

As part of this, I've made the opinionated change to rename `token` to `job_token` just to avoid confusion between it and any of the credentials we handle in some places.


brrygrdn:
Glad to have the smoke tests as they seem to indicate I've missed a few touch points

updater/lib/dependabot/job.rb
```diff
@@ -4,6 +4,16 @@
...
+# Describes a single Dependabot workload within the GitHub-integrated Service
+#
+# This primarily acts as a value class to hold inputs for various Core objects
+# and is an approximate data structure for the 'job description file' used by
+# the CLI tool.
+#
+# See: https://github.com/dependabot/cli#job-description-file
+#
+# This class should evenually be promoted to common/lib and augmented to
+# validate job description files.
```
landongrindheim, line 16: I'm appreciating these class-level comments 😄

landongrindheim (approved):
This makes a lot of sense to me. Great job 🥇

PR #6771, opened 2023-03-02
Annotate `pending` tests with comments

jeffwidman:
Most of our `pending` tests have associated code comments. However, they're only visible when viewing the raw source, not when running `rspec`.

So this makes it visible when running tests why a particular test is pending.


landongrindheim (approved):
Looks good to me!

Are we getting any value from having pending tests? Could we delete them if we can't re-enable them?

jeffwidman:
> Are we getting any value from having pending tests? Could we delete them if we can't re-enable them?

Good Q... I'm not in a hurry to drop them as I've been slowly wading through some of them and cleaning them up... that's how I ran into this PR actually, wondering if they could be cleaned up or not only to discover the code comments... the bundler ones I haven't looked at yet, they may be droppable if/when we drop bundler 1. The others are essentially self-documenting of bugs, so that no one is like "we missed test coverage"... and not harming anything in there for now.

PR #6774, opened 2023-03-03
WIP - does composer internally handle the `v` prefix?

jeffwidman:
in middle of other code and saw a note about composer internally handling the v prefix just fine, so throwing this up to see what CI thinks... since I know we have test cases that check against v/non-v...


jeffwidman:
CI shows that we still need the `v` prefix... The open question in my mind is that because our code doesn't handle the `v` prefix so we pre-emptively strip it off in the native helper? Ie, if `composer` thinks with `v` is fine, should we be allowing that within the rest of our code?

However, not worth worrying about right now, the bigger priority in our PHP support would be dropping `composer 1` and switching to `php` `8` support. After that we could circle back on this I suppose.

PR #6775, opened 2023-03-03
Switch commented out test to pending

jeffwidman:
This should be `pending` rather than commented out.

Long term we can probably [stub in a super short timeout via a custom `.yarnrc` file](https://azureossd.github.io/2022/09/10/fix-yarn-ESOCKETTIMEDOUT-with-.yarnrc-configuration-file/) but this is good enough for the first iteration here.


Nishnha (approved):
makes sense to use pending since it shows up in the test suite output. 
if we did `xit` then we wouldn't see that this test exists (and was skipped over) when we run the test suite

jeffwidman:
> if we did xit then we wouldn't see that this test exists (and was skipped over) when we run the test suite

From my local testing, this isn't true... `xit`, `skip()`, and `pending()` all show up in the test output.

What's more interesting is that a test that's marked `pending()` but passes will fail the test suite... so the fact that this test didn't fail the suite indicates it's currently broken. 🤔

jeffwidman:
Indeed, the test is running/failing: 
* https://github.com/dependabot/dependabot-core/issues/6778

Issue #6778, opened 2023-03-03
Test that was expected to pass but marked as pending because slow isn't actually passing


jeffwidman:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code improvement description

In https://github.com/dependabot/dependabot-core/pull/6775/files#diff-2af3f7bf669aa660ac397125615d548476086dab856e6f89041d4ebe2cbe7f6aR166, I switched a test from commented out to `pending()` because I thought that meant it'd be skipped. The original impetus for being commented out was because it tests a timeout, so it is very slow.

However, I just realized that `pending()` still runs tests... and if they pass, it will fail the test suite so you can unpend the test.

I double-checked and sure enough, this test is running/failing: https://github.com/dependabot/dependabot-core/actions/runs/4319740144/jobs/7539244341#step:5:188

```
2) Dependabot::NpmAndYarn::FileUpdater::YarnLockfileUpdater errors with a registry that times out raises a helpful error
     # This test is extremely slow (1m45s) so only run locally. TODO: stub a custom timeout value.
     Failure/Error:
       expect { updated_yarn_lock_content }.
         to raise_error(Dependabot::PrivateSourceTimedOut) do |error|
           expect(error.source).to eq("timeout.cv/repository/mirror")
         end

       expected Dependabot::PrivateSourceTimedOut, got #<Dependabot::SharedHelpers::HelperSubprocessFailed: https://timeout.cv/repository/mirror/fetch-factory: getaddrinfo ENOTFOUND timeout.cv> with backtrace:
         # /home/dependabot/common/lib/dependabot/shared_helpers.rb:123:in `run_helper_subprocess'
         # ./lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb:[191](https://github.com/dependabot/dependabot-core/actions/runs/4319740144/jobs/7539244341#step:5:192):in `run_yarn_top_level_updater'
         # ./lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb:114:in `block (2 levels) in run_yarn_updater'
         # ./lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb:107:in `chdir'
         # ./lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb:107:in `block in run_yarn_updater'
         # /home/dependabot/common/lib/dependabot/shared_helpers.rb:169:in `with_git_configured'
[...]

PR #6779, opened 2023-03-03
Use `skip()` instead of `pending()` where appropriate

jeffwidman:
In `rspec`, `skip()` completely skips the test. Whereas `pending()` runs the test, and expects it to fail. In fact, if it doesn't fail, `pending()` will fail the test suite so that the developer realizes whatever the failure mode was is now fixed.

For tests that we _never_ expect to pass without developer time, there's no point in marking them as `pending()`, we should mark them as `skip()` instead.

I grep'd for `pending(` and switched to `skip(` where appropriate.


npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater_spec.rb
```diff
@@ -163,7 +163,7 @@
       let(:files) { project_dependency_files("yarn/simple_with_registry_that_times_out") }
 
       it "raises a helpful error" do
-        pending("This test is extremely slow (1m45s) so only run locally. TODO: stub a custom timeout value.")
```
jeffwidman, line 166: This is not only a slow test, it's also a failing test, otherwise `pending()` wouldn't have worked:
* #6778

deivid-rodriguez (approved):

PR #6780, opened 2023-03-03
Allow passing autoCompleteIgnoreConfigIds value for azure

mburumaxwell:
Autocomplete functionality added in #4064 is missing a variable to set autoCompleteIgnoreConfigIds. This PR adds it.

### DIFF
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -194,7 +194,7 @@ module Dependabot
 
       def autocomplete_pull_request(pull_request_id, auto_complete_set_by, merge_commit_message,
                                     delete_source_branch = true, squash_merge = true, merge_strategy = "squash",
-                                    trans_work_items = true)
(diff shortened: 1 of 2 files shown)


jeffwidman (commented):
What does this field represent? Should there be a unit test for passing a non-default value?

mburumaxwell:
According to the docs, "List of any policy configuration Id's which auto-complete should not wait for. Only applies to optional policies"

I don't believe there is a need to test for non-default values given:
1. This is just for Azure DevOps and not any other provider
2. The method is only invoked in unofficial usage e.g. an extension or custom usage
3. The value is only serialized and not used for anything else like making another object.

mburumaxwell:
@jeffwidman your thoughts?

mburumaxwell:
@jeffwidman the test is in. Okay now?

jeffwidman (approved):
My apologies, I lost track of this.

Looks great!

PR #6782, opened 2023-03-03
Attempt to detect OutOfDisk errors when cloning repos

mctofu:
Failures from cloning a repo assume a repo access error by default. This causes confusion when the error is something else. I suspect OutOfDisk is the culprit for an issue I'm looking into so this adds detection for that.


honeyankit (approved):
Thank you for this PR.

Nishnha (approved):
Nice improvement. I didn't realize catching a new error type could be this straightforward!

PR #6784, opened 2023-03-03
Introduce group rule to branch namer

landongrindheim:
As we move toward grouping updates, we'll need an alternative approach to naming. This pull request represents the structure of this change. Its details have been omitted intentionally as I believe we'll have better information as we get closer to fully supporting that workflow.


common/lib/dependabot/pull_request_creator/branch_namer/solo_strategy.rb
```diff
@@ -0,0 +1,208 @@
+# frozen_string_literal: true
+
+require "digest"
+
+require "dependabot/metadata_finders"
+
+module Dependabot
+  class PullRequestCreator
+    class BranchNamer
+      class SoloStrategy
```
landongrindheim, line 10: This object represents our current approach to naming. I don't think `Solo` necessarily captures that well, but it stands as a complement to a grouping. I'm open to this changing now or later 🙂

common/lib/dependabot/group_rule.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Dependabot
+  class GroupRule
```
landongrindheim, line 4: I'm a little unclear on the interfact for a `GroupRule`. The `rule` portion signifies a policy-like approach, where criteria must be met. I'm open to this changing, or to refining the concept as we move forward.

common/lib/dependabot/pull_request_creator/branch_namer/group_rule_strategy.rb
```diff
@@ -0,0 +1,28 @@
...
+          @files         = files
+          @target_branch = target_branch
+          @group_rule    = group_rule
+          @separator     = separator
+          @prefix        = prefix
+          @max_length    = max_length
+        end
+
+        def new_branch_name
+          group_rule.name
```
brrygrdn, line 19: We should probably expect a group rule name to be a humanized string, as we'll need to either:
- Validate it strictly as being a valid branch name in the config file
- Accept any string and 'slugify' it here replacing invalid characters/spaces etc

I'd tend to prefer the latter as while it might end up making sense to validate it in the config for other reasons, this component probably shouldn't trust its inputs?

brrygrdn (approved):
👍🏻 This makes sense to me, I think sinking the existing logic into a discrete class is a good call vs just adding more to `BranchNamer` 😅 - `SoloStrategy` makes sense name-wise, we can revisit it later as we figure out how the grouped vs non-grouped behaviours play out.

Issue #4979, opened 2022-04-08
[pub] Reconsider the default `versioning-strategy`


jonasfj:
Reconsider the default `versioning-strategy`.

We have the following _versioning-strategies_:
 * `widen`, extend only the upper bound to include the new version.
 * `increase`, always update the constraint lower bound to match the new version.
 * `increase-if-necessary`, leave the constraint if the original constraint allows the new version, otherwise, bump the constraint.

<table>
    <thead>
        <tr>
            <th align="center">Constraint</th>
            <th align="center">Current Version</th>
            <th align="center">New Version</th>
            <th>Strategy</th>
            <th>New Constraint</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td rowspan="3" align="center"><code>^1.0.0</code></td>
            <td rowspan="3" align="center"><code>1.0.0</code></td>
            <td rowspan="3" align="center"><code>1.2.0</code></td>
            <td align="center"><code>'widen'</code></td>
            <td><code>^1.0.0</code></td>
        </tr>
        <tr>
            <td align="center"><code>'increase'</code></td>
            <td><code>^1.2.0</code></td>
        </tr>
        <tr>
            <td align="center"><code>'increase-if-necessary'</code></td>
            <td><code>^1.0.0</code></td>
        </tr>
      <tr>
            <td rowspan="3" align="center"><code>^1.0.0</code></td>
            <td rowspan="3" align="center"><code>1.0.0</code></td>
            <td rowspan="3" align="center"><code>2.0.0</code></td>
            <td align="center"><code>'widen'</code></td>
            <td><code>&gt;=1.0.0 &lt;3.0.0</code></td>
        </tr>
        <tr>
            <td align="center"><code>'increase'</code></td>
            <td><code>^2.0.0</code></td>
        </tr>
        <tr>
            <td align="center"><code>'increase-if-necessary'</code></td>
            <td><code>^2.0.0</code></td>
        </tr>
    </tbody>
</table>
(side note: maybe we should include tables like this in the documentation, feel free to steal)

### Considerations

**For packages** using the `widen` strategy is attractive because increasing the lower-bound constraint can cause resolution conflicts (as we can only have one version of each package).

**For applications** using the `increase` strategy is attractive because it makes accidental downgrades unlikely.

[...]

PR #6786, opened 2023-03-04
[pub] use 'increase-if-necessary' as the default strategy for pub dependencies

devoncarew:
- change the default versioning strategy for pub dependencies to `increase-if-necessary`
- as a side-effect, no longer use the pubspec.yaml 'version' field to try and identify an app vs a package (they'll both get the same default strategy)

I believe this is generally in-line with what's discussed in https://github.com/dependabot/dependabot-core/issues/4979.

### DIFF
--- a/pub/lib/dependabot/pub/update_checker.rb
+++ b/pub/lib/dependabot/pub/update_checker.rb
@@ -134,21 +134,8 @@ module Dependabot
           [nil, "widen_ranges", "bump_versions", "bump_versions_if_necessary"].include? requirements_update_strategy
 
         if requirements_update_strategy.nil?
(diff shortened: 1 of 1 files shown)


devoncarew:
> Before merging, I want to take one more gander at #4979 and think carefully about the choices there... at first glance, the arguments presented there could just as easily apply to most other ecosystems... not saying it's a bad idea, just that I want to think on it overnight. I'll post in that issue if I have any concerns.

No worries - I'm fine with taking some time to think through the changes.

mafreud:
@devoncarew, Could you resume fixing the conflicts? I agree with that default versioning strategy for ```pub``` is ```increase-if-necessary```.

PR #6792, opened 2023-03-06
Introduce Dependabot::DependencyChange as a wrapper for updater output plus metadata

brrygrdn:
This PR introduces a new [Dependabot::DependencyChange](https://github.com/dependabot/dependabot-core/pull/6792/files#diff-1df7c2253c7a813b8528a05529986fe5d2f7ac95cf2e9d5d6c65db463fc28aba) class which wraps a set of `Dependabot::Dependency` and `Dependabot::DependencyFile` objects which constitute the outcome of a `Dependabot::Updater` operation.

### Rationale

As part of the [prototype](https://github.com/dependabot/dependabot-core/pull/6663/files#diff-1f8c3901647148d6f9fbda3a5c943490beb28c1291f21a3b6713471e0c39902bR42-R62) effort, Rubocop flagged the growing number of parameters being passed down into the API Client - effectively the 'output' class for the Updater - when we started adding in new placeholders for the concept of a "Group Rule".

This problem indicates we have a missing abstraction between the operation to generate a change and the components when then transform it for output via Dependabot API.

Our next major planned refactor is to turn the `Dependabot::Updater` operations into concrete classes so each combination of major job branches is represented by a single, well-defined class for a fresh Version Update, a rebase of an existing Version PR, a new Security Update, etc, etc.

The `Dependabot::DependencyChange` class is a bottom-up refactor that adds a new abstraction which reduces the amount of parameters we need to pass while also providing a starting point for the output of these new operation classes to force a wedge between the part of the Updater which constructs change sets and the part which processes them into PRs so we can start to expose the injection site for grouping behaviours.

### Changes

Originally I intended the `Dependabot::DependencyChange` to be built up over the course of the `Dependabot::Updater#check_and_create_pull_request` and `Dependabot::Updater#check_and_update_pull_request` methods, allowing more of the logic which combs and filters the "updated_deps" list to be moved into the class itself.

I still think this might be worth doing, but I walked it back as it resulted in a substantial number of test stub rewrites which made me lose confidence in the change.

For now, the only major responsibility the `Dependabot::DependencyChange` class extracts from the Updater is responsibility for constructing the Pull Request message object, which it does by being aware of the `Dependabot::Job`. 

I think there are some logical threads that follow on from this to incorporate things like the `@created_pull_requests` and `@errors` sets unto the job so the change set is capable of answering questions about wither it duplicates or supersedes existing work - but I've left that for now to de-risk introducing this interface and avoid delaying splitting the `Dependabot::Updater` out further.

### DIFF
--- a/updater/lib/dependabot/api_client.rb
+++ b/updater/lib/dependabot/api_client.rb
@@ -73,7 +73,7 @@ module Dependabot
       body = {
         data: {
           "dependency-names": dependency_change.dependencies.map(&:name),
-          "updated-dependency-files": dependency_change.updated_dependency_files,
+          "updated-dependency-files": dependency_change.updated_dependency_files_hash,
           "base-commit-sha": base_commit_sha
         }
       }
--- a/updater/lib/dependabot/api_client.rb
+++ b/updater/lib/dependabot/api_client.rb
@@ -52,10 +52,9 @@ module Dependabot
       Job.new(job_data.merge(token: job_token))
     end
 
-    def create_pull_request(dependencies, updated_dependency_files,
-                            base_commit_sha, pr_message)
+    def create_pull_request(dependency_change, base_commit_sha)
       api_url = "#{base_url}/update_jobs/#{job_id}/create_pull_request"
-      data = create_pull_request_data(dependencies, updated_dependency_files, base_commit_sha, pr_message)
+      data = create_pull_request_data(dependency_change, base_commit_sha)
       response = http_client.post(api_url, json: { data: data })
       raise ApiError, response.body if response.code >= 400
     rescue HTTP::ConnectionError, OpenSSL::SSL::SSLError
@@ -67,12 +66,12 @@ module Dependabot
     end
 
     # TODO: Determine if we should regenerate the PR message within core for updates
-    def update_pull_request(dependencies, updated_dependency_files, base_commit_sha)
+    def update_pull_request(dependency_change, base_commit_sha)
       api_url = "#{base_url}/update_jobs/#{job_id}/update_pull_request"
       body = {
         data: {
-          "dependency-names": dependencies.map(&:name),
-          "updated-dependency-files": updated_dependency_files,
+          "dependency-names": dependency_change.dependencies.map(&:name),
+          "updated-dependency-files": dependency_change.updated_dependency_files,
           "base-commit-sha": base_commit_sha
         }
       }
@@ -192,9 +191,9 @@ module Dependabot
       sleep(rand(3.0..10.0)) && retry
     end
 
-    def create_pull_request_data(dependencies, updated_dependency_files, base_commit_sha, pr_message)
+    def create_pull_request_data(dependency_change, base_commit_sha)
       data = {
-        dependencies: dependencies.map do |dep|
+        dependencies: dependency_change.dependencies.map do |dep|
           {
             name: dep.name,
             "previous-version": dep.previous_version,
@@ -205,14 +204,14 @@ module Dependabot
             removed: dep.removed? ? true : nil
           }.compact)
         end,
-        "updated-dependency-files": updated_dependency_files,
+        "updated-dependency-files": dependency_change.updated_dependency_files_hash,
         "base-commit-sha": base_commit_sha
       }
-      return data unless pr_message
+      return data unless dependency_change.pr_message
 
-      data["commit-message"] = pr_message.commit_message
-      data["pr-title"] = pr_message.pr_name
-      data["pr-body"] = pr_message.pr_message
+      data["commit-message"] = dependency_change.pr_message.commit_message
+      data["pr-title"] = dependency_change.pr_message.pr_name
+      data["pr-body"] = dependency_change.pr_message.pr_message
       data
     end
   end
--- a/updater/lib/dependabot/api_client.rb
+++ b/updater/lib/dependabot/api_client.rb
@@ -52,6 +52,7 @@ module Dependabot
       Job.new(job_data.merge(token: job_token))
     end
 
+    # TODO: Make `base_commit_sha` part of Dependabot::DependencyChange
     def create_pull_request(dependency_change, base_commit_sha)
       api_url = "#{base_url}/update_jobs/#{job_id}/create_pull_request"
       data = create_pull_request_data(dependency_change, base_commit_sha)
@@ -65,6 +66,7 @@ module Dependabot
       sleep(rand(3.0..10.0)) && retry
     end
 
+    # TODO: Make `base_commit_sha` part of Dependabot::DependencyChange
     # TODO: Determine if we should regenerate the PR message within core for updates
     def update_pull_request(dependency_change, base_commit_sha)
       api_url = "#{base_url}/update_jobs/#{job_id}/update_pull_request"
--- a/updater/lib/dependabot/api_client.rb
+++ b/updater/lib/dependabot/api_client.rb
@@ -208,7 +208,17 @@ module Dependabot
         end,
         "updated-dependency-files": dependency_change.updated_dependency_files_hash,
         "base-commit-sha": base_commit_sha
-      }
+      }.merge({
+        # TODO: Replace this flag with a group-rule object
+        #
+        # In future this should be something like:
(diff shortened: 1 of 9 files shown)


updater/lib/dependabot/api_client.rb
```diff
@@ -192,9 +193,9 @@ def fetch_job_details_from_backend
       sleep(rand(3.0..10.0)) && retry
     end
 
-    def create_pull_request_data(dependencies, updated_dependency_files, base_commit_sha, pr_message)
+    def create_pull_request_data(dependency_change, base_commit_sha)
```
brrygrdn, line 196: This method has gotten big enough I think it's probably worth extracting it into a serialiser class which presents a `dependency_change` for the API in a neater way.

I've punted on this for now as I want to land the concept of `DependencyChange` and I'm conscious it might change during code review.

updater/lib/dependabot/api_client.rb
```diff
@@ -52,10 +52,10 @@ def fetch_job
       Job.new(job_data.merge(token: job_token))
     end
 
-    def create_pull_request(dependencies, updated_dependency_files,
-                            base_commit_sha, pr_message)
+    # TODO: Make `base_commit_sha` part of Dependabot::DependencyChange
```
brrygrdn, line 55: I've punted on this for now because how we obtain the `base_commit_sha` is a little unusual. I think for the purposes of `Dependabot::Updater` it should always be present on the `job`, but I need to verify this.

updater/lib/dependabot/api_client.rb
```diff
@@ -205,14 +206,24 @@ def create_pull_request_data(dependencies, updated_dependency_files, base_commit
...
+        #
+        # This will allow us to pass back the rule id and other parameters
+        # to allow Dependabot API to augment PR creation and associate it
+        # with the rule for rebasing, etc.
+        "grouped-update": dependency_change.grouped_update? ? true : nil
+      }.compact)
+      return data unless dependency_change.pr_message
+
+      data["commit-message"] = dependency_change.pr_message.commit_message
+      data["pr-title"] = dependency_change.pr_message.pr_name
```
abdulapopoola, line 225: newbie q: can these (e.g. "pr-title") be constants?

updater/lib/dependabot/dependency_change.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+
+    def humanized
+      dependencies.map do |dependency|
+        "#{dependency.name} ( from #{dependency.previous_version} to #{dependency.version} )"
+      end.join(", ")
+    end
+
+    def updated_dependency_files_hash
+      updated_dependency_files.map(&:to_h)
```
landongrindheim, line 51: I assume this is simply porting over current behavior, in which case we can return to this conversation later 🔖

1. Are we certain these files are in the required shape to have `#to_h` called on them?
2. Are there concerns around file name collisions?
  * concern: If we have two `Gemfile` entries, one would be lost when being set as a Hash key.

updater/lib/dependabot/dependency_change.rb
```diff
@@ -0,0 +1,61 @@
...
+        files: updated_dependency_files,
+        credentials: job.credentials,
+        commit_message_options: job.commit_message_options,
+        # This ensures that PR messages we build replace github.com links with
+        # a redirect that stop markdown enriching them into mentions on the source
+        # repository.
+        #
+        # TODO: Promote this value to a constant or similar once we have
+        # updated core to avoid surprise outcomes if this is unset.
+        github_redirection_service: "github-redirect.dependabot.com"
```
landongrindheim, line 40: 👌 This makes sense to me. We could make it a constant and provide it as the default argument 👍

updater/lib/dependabot/dependency_change.rb
```diff
@@ -0,0 +1,61 @@
...
+# determined by a Dependabot operation.
+#
+# It includes a list of changed Dependabot::Dependency objects, an array of
+# Dependabot::DependencyFile objects which contain the changes to be applied
+# along with any Dependabot::GroupRule that was used to generate the change.
+#
+# This class provides methods for presenting the change set which can be used
+# by adapters to create a Pull Request, apply the changes on disk, etc.
+module Dependabot
+  class DependencyChange
```
landongrindheim, line 13: I'm really digging this extraction! 👏 on noticing the need and finding a sane way to address it 🙇 

I can tell you've spent time thinking about what this object should include/exclude. I find myself wondering whether `job` is part of a dependency change. My thinking is that a job might _have a_ dependency change, while the rest of the parameters are _parts of_ a dependency change. 🤔

I see `job` is only referenced in `#pr_message`. Given that, we _could_ let it be an argument to that method. I kinda like that: `dependency_change.pr_message(for: job)`

I'll trust your judgment here. I think this is a big improvement, and points us in a really positive direction 🥇

updater/lib/dependabot/updater.rb
```diff
@@ -189,13 +196,13 @@ def check_and_update_pull_request(dependencies)
         # The dependencies being updated have changed. Close the existing
         # multi-dependency PR and try creating a new one.
         close_pull_request(reason: :dependencies_changed)
-        create_pull_request(updated_deps, updated_files, pr_message(updated_deps, updated_files))
+        create_pull_request(updated_deps, updated_files)
```
landongrindheim, line 199: 🙌

updater/spec/dependabot/dependency_change_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+      allow(job).to receive(:source).and_return(github_source)
+      allow(job).to receive(:credentials).and_return(job_credentials)
+      allow(job).to receive(:commit_message_options).and_return(commit_message_options)
+      allow(Dependabot::PullRequestCreator::MessageBuilder).to receive(:new).and_return(message_builder_mock)
+    end
+
+    context "when the change is for a new pull request" do
+      before do
+        allow(job).to receive(:updating_a_pull_request?).and_return(false)
+      end
```
landongrindheim, line 98: 🤙 Callback to an earlier comment, I think being able to pass in `job` to `#pr_message` would make this test really simple.

updater/spec/dependabot/integration_spec.rb
```diff
@@ -145,7 +145,7 @@
 
     it "updates dependencies correctly" do
       expect(api_client).
-        to receive(:create_pull_request) do |deps, files, commit_sha|
+        to receive(:create_pull_request) do |dependency_change, commit_sha|
```
landongrindheim, line 148: 🤤

landongrindheim (approved):
I ❤️ this change!

updater/lib/dependabot/dependency_change.rb
brrygrdn: I'm moving the existing comment in this case, I would address this but I've forgotten what this means since I wrote the original comment:

```
once we have  updated core to avoid surprise outcomes if this is unset
```

I'm not sure that has happened yet?

updater/lib/dependabot/dependency_change.rb
brrygrdn: Yeah, I'm 50/50 on this class knowing about the Job - I think of it as "I know the context in which this change was made" as the `base_commit_sha` could conceivably be retrieved from it as well and the `Updater#existing_pull_request` could sink into `DependencyChange` as well if the `Job` maintained the list of Pull Requests created so far.

Essentially I think this feels untidy as it is an unfinished piece; I think this class should know about the 'context' in which it is created, the job is _part_ of that context and good enough to use for now.

The thing I'm not sure about is, should the Job _be_ the context or should the Job and some other things be split out into a new class representing just context?

updater/spec/dependabot/dependency_change_spec.rb
brrygrdn: Yeah, that's a good call out. The API client currently doesn't know about the job, but it does know the `job_id` and `job_token` - it might make sense to instantiate it with the Job so it can use what it needs? 

This might violate principal of least knowledge a little if Job becomes the overall "context" class as I mentioned above, maybe this is a good reason to split job and context?

PR #6796, opened 2023-03-06
Bump latest version to reflect current release

landongrindheim:
This being out of date caused a [CI failure](https://github.com/dependabot/dependabot-core/actions/runs/4346810196/jobs/7594010245) in another branch. It'd be great to make this a static target, but I'm not sure what our process is for locking the version down.


jeffwidman (approved):

jeffwidman:
We've mocked this kind of thing in `core` before, but IIUC, these `updater` tests tend to be sanity checks that function more like lightweight integration tests, so maybe better not to mock... in general, this particularly library tends not to change that often, and not a big deal when it does to bump it. 🤷

PR #6798, opened 2023-03-06
Detect when pip compile hits memory or disk limits

mctofu:
These currently end up in the unknown_error category so this provides easier visibility to the user.

### DIFF
--- a/python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
+++ b/python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
@@ -173,6 +173,10 @@ module Dependabot
             raise GitDependenciesNotReachable, url
           end
 
+          raise Dependabot::OutOfDisk if error.message.end_with?("[Errno 28] No space left on device")
+
+          raise Dependabot::OutOfMemory if error.message.end_with?("MemoryError")
+
           raise
         end
         # rubocop:enable Metrics/AbcSize
--- a/python/spec/dependabot/python/update_checker/pip_compile_version_resolver_spec.rb
+++ b/python/spec/dependabot/python/update_checker/pip_compile_version_resolver_spec.rb
@@ -387,5 +387,43 @@ RSpec.describe namespace::PipCompileVersionResolver do
         end
       end
     end
+
+    context "that fails to resolve due to resource limits" do
+      context "because it ran out of disk space" do
+        before do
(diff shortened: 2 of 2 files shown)


python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
```diff
@@ -173,6 +173,10 @@ def handle_pip_compile_errors(error)
             raise GitDependenciesNotReachable, url
           end
 
+          raise Dependabot::OutOfDisk if error.message.end_with?("[Errno 28] No space left on device")
```
honeyankit, line 176: This is fine or we can also just check for error message contains `No space left on device`.

honeyankit (approved):

jeffwidman (approved):

python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
abdulapopoola: should the message be a constant?

Also purely curious; do we know what is filling up the disk space? I.E., why are the machines running out of disk space?

mctofu: > should the message be a constant?

IMO, not really since that adds an extra layer of indirection when reading the code. Things that would nudge me there are:
- We used this specific message in more than one place with the python handling code
- The match was a complicated regex or had a non-obvious connection to the error so would benefit from a more descriptive constant name.

> Also purely curious; do we know what is filling up the disk space? I.E., why are the machines running out of disk space?

Primary usage of diskspace:
- The repo being updated, especially in ecosystems where we need to clone the repo
- Dependencies, especially when the dependencies are vendored so we'll have old + new versions of the dependencies on disk somewhere.

There's also potential for inefficiency when we update multiple dependencies. If the package manager is leaving behind files for each update that may also contribute to running out of space towards the end of a job. I think we'll be able to better analyze once we start reporting the error.

abdulapopoola: Thanks; nbd with the constant if this is a localized single value; I raised it because I saw a couple scenarios wherein the strings were used as constants.

I'll be excited to see what the data shows up once the reports start coming in! Thanks for fixing this again.

PR #4441, opened 2021-11-23
Use redirect.github.com for redirect service

lseppala:
`github-redirect.dependabot.com` has a single CNAME record pointing to `redirect.github.com`. 

```
❯ dig +all github-redirect.dependabot.com
; <<>> DiG 9.10.6 <<>> +all github-redirect.dependabot.com

;; QUESTION SECTION:
;github-redirect.dependabot.com.	IN	A

;; ANSWER SECTION:
github-redirect.dependabot.com.	3600 IN	CNAME	redirect.github.com.
redirect.github.com.	60	IN	A	140.82.113.18
```

This change updates the `DEFAULT_GITHUB_REDIRECTION_SERVICE` constant from `github-redirect.dependabot.com` to `redirect.github.com`. This DNS entry is a stronger candidate for long-term support as dependabot.com records are decommissioned. 

It wasn't strictly necessary to update tests as `github_redirection_service` is explicitly set in tests, but references to `github-redirect.dependabot.com` were replaced to avoid future confusion.

PR #6801, opened 2023-03-07
Rely on previously-defined constant

landongrindheim:
The DEFAULT_GITHUB_REDIRECTION_SERVICE has been the default argument for PullRequestCreator, but I think it can also serve the same purpose in PullRequestCreator::MessageBuilder.

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -207,7 +207,7 @@ jobs:
         gh release download --repo dependabot/cli -p "*linux-amd64.tar.gz"
         tar xzvf *.tar.gz >/dev/null 2>&1
         ./dependabot --version
-        URL=https://api.github.com/repos/dependabot/smoke-tests/contents/tests/smoke-${{ matrix.suite.name }}.yaml
+        URL=https://api.github.com/repos/dependabot/smoke-tests/contents/tests/smoke-${{ matrix.suite.name }}.yaml?ref=use-redirect-dot-github
         curl $(gh api $URL --jq .download_url) -o smoke.yaml
 
     # Download the Proxy cache. The job is ideally 100% cached so no real calls are made.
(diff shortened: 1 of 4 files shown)


.github/workflows/smoke.yml
```diff
@@ -186,7 +186,7 @@ jobs:
         gh release download --repo dependabot/cli -p "*linux-amd64.tar.gz"
         tar xzvf *.tar.gz >/dev/null 2>&1
         ./dependabot --version
-        URL=https://api.github.com/repos/dependabot/smoke-tests/contents/tests/smoke-${{ matrix.suite.name }}.yaml
+        URL=https://api.github.com/repos/dependabot/smoke-tests/contents/tests/smoke-${{ matrix.suite.name }}.yaml?ref=use-redirect-dot-github
```
landongrindheim, line 189: This should **not** make it into the default branch. It's pointing to [a branch](https://github.com/dependabot/smoke-tests/pull/40) which holds the expected output of this branch's concerns. I'm open to input re: whether it should remain as part of commit history.

landongrindheim: ☝️ was a proof-of-concept wherein I updated the redirect URL in the smoke-tests repo. This is the resulting build result 👇 
image

---

I'll be removing that ref now.

Nishnha:
`DEFAULT_GITHUB_REDIRECTION_SERVICE` is defined as `redirect.github.com` in the `PullRequestCreator`, but the redirect URL that's being replaced in this PR is `github-redirect.dependabot.com`.

Does that matter?

https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/pull_request_creator.rb#L26

Nishnha:
> `DEFAULT_GITHUB_REDIRECTION_SERVICE` is defined as `redirect.github.com` in the `PullRequestCreator`, but the redirect URL that's being replaced in this PR is `github-redirect.dependabot.com`.
> 
> What's the difference between the two URLs?
> 
> https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/pull_request_creator.rb#L26

Ah, looks like `github-redirect.dependabot.com` is an outdated URL. It was updated in https://github.com/dependabot/dependabot-core/pull/4441

Nishnha (approved):
Thanks for showing that e2e tests are passing on the `use-redirect-dot-github` branch.

This LGTM

PR #6808, opened 2023-03-08
Extract log helper methods into a Logger::Formatter, Push the `job_id` into the Job object

brrygrdn:
This branch extracts a couple of tidy ups from some ongoing work to prepare the Dependabot::Updater to receive our [prototype grouped updating strategy](https://github.com/dependabot/dependabot-core/pull/6663).

While splitting the updater up into discrete classes, there were two cross-cutting pain points:
- Having pass _both_ a Job object and the job_id is counter productive and it leaves us reaching back to the `Dependabot::Environment` object as we start to extract code. 
  - This is likely going to make compartmentalisation worse as the `Environment` object should never be part of Core but some of the Updater strategies we are extracting _should_ become Core-bound eventually so ecosystems can override them
  - Solution: Let's just allow a Job to know its id?
- Writing to the `Dependabot.logger` is wrapped in `logger_info` and `logger_error` methods in two levels of abstraction, the "Command" class and the Updater itself to inject a `job_id` prefix
  - Breaking the code out means they need access to a global logger with the prefix behaviour
  - We also have some limitations in our current logging layout that means we don't use debug level message which would _really_ help both contributors and users a lot
  - Solution: Let's use Logger::Formatters to attach a presenter to the global logger we setup at the start of a job
  - Bonus points: Let's teach `Dependabot::Environment` about a debug flag so we can default to :info, but optionally invoke :debug

### Service Changes

- This branch introduces the use of a `DEPENDABOT_DEBUG` envvar, when 'true' it will set the log level to :debug.
- The job.json, aka Job Definition File, can now also pass a `job.debug = true` value to set the log level on a job-by-job basis
- The formatter is aware of github.com/dependabot/cli's [placeholder Job ID](https://github.com/dependabot/cli/blob/8f18d3681cafd6bbd5078ddfb7893ae8c227837b/internal/infra/updater.go#L25) and omits the prefix in that environment

### DIFF
--- a/updater/lib/dependabot/environment.rb
+++ b/updater/lib/dependabot/environment.rb
@@ -10,9 +10,12 @@ module Dependabot
       @job_token ||= environment_variable("DEPENDABOT_JOB_TOKEN")
     end
 
+    def self.debug_enabled?
+      @debug_enabled ||= job_debug_enabled? || environment_debug_enabled?
+    end
+
     def self.api_url
-      default = "http://localhost:3001"
-      @api_url ||= environment_variable("DEPENDABOT_API_URL", default)
+      @api_url ||= environment_variable("DEPENDABOT_API_URL", "http://localhost:3001")
     end
 
     def self.job_path
@@ -35,7 +38,7 @@ module Dependabot
       @job_definition ||= JSON.parse(File.read(job_path))
     end
 
-    def self.environment_variable(variable_name, default = :_undefined)
+    private_class_method def self.environment_variable(variable_name, default = :_undefined)
       return ENV.fetch(variable_name, default) unless default == :_undefined
 
       ENV.fetch(variable_name) do
@@ -43,6 +46,12 @@ module Dependabot
       end
     end
 
-    private_class_method :environment_variable
+    private_class_method def self.job_debug_enabled?
+      !!job_definition.dig("job", "debug")
+    end
+
+    private_class_method def self.environment_debug_enabled?
+      !!environment_variable("DEPENDABOT_DEBUG", false)
+    end
   end
 end
--- a/updater/lib/dependabot/environment.rb
+++ b/updater/lib/dependabot/environment.rb
@@ -14,6 +14,10 @@ module Dependabot
       @debug_enabled ||= job_debug_enabled? || environment_debug_enabled?
     end
 
+    def self.log_level
+      debug_enabled? ? :debug : :info
+    end
+
     def self.api_url
       @api_url ||= environment_variable("DEPENDABOT_API_URL", "http://localhost:3001")
     end
--- a/updater/spec/dependabot/environment_spec.rb
+++ b/updater/spec/dependabot/environment_spec.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+require "dependabot/environment"
+
+RSpec.describe Dependabot::Environment do
+  subject(:environment) { described_class }
+
+  describe "::debug_enabled?" do
+    after do
+      # Reset the memoisation after each test
+      environment.remove_instance_variable(:@debug_enabled)
+    end
+
+    it "is false by default" do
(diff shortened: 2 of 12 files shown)


updater/spec/dependabot/updater_spec.rb
```diff
@@ -2441,10 +2441,10 @@ def expect_update_checker_with_ignored_versions(versions)
 
         expect(Dependabot.logger).
           to have_received(:info).
-          with("<job_1>   version-update:semver-patch - from .github/dependabot.yaml")
+          with("  version-update:semver-patch - from .github/dependabot.yaml")
         expect(Dependabot.logger).
           to have_received(:info).
-          with("<job_1>   version-update:semver-minor - from .github/dependabot.yaml")
+          with("  version-update:semver-minor - from .github/dependabot.yaml")
```
brrygrdn, line 2447: I believe this double-spacing is working as intended as the intent is to tab the content in from the right margin, but I'll check

brrygrdn: [Yep](https://github.com/dependabot/dependabot-core/blob/a60d501d79965fe739b039a598cea9b3ddc74376/updater/lib/dependabot/updater.rb#L546)

updater/spec/dependabot/environment_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    after do
+      # Reset the memoisation after each test
+      environment.remove_instance_variable(:@debug_enabled)
+    end
+
+    it "is false by default" do
+      allow(environment).to receive(:job_definition).and_return({})
+      allow(ENV).to receive(:fetch).with("DEPENDABOT_DEBUG", false).and_return(false)
+
+      expect(environment).not_to be_debug_enabled
```
Nishnha, line 20: where is `be_debug_enabled` defined?

updater/spec/dependabot/updater_spec.rb
Nishnha: I think you should be able to match with regex - instead of `with("  some message")` you could use `with(/some message/)` to ignore whitespace

Can't tell if capturing whitespace is intentional here

Nishnha (approved):
Love the idea of a global level logger.

This all makes sense so far. I think you'll also need to define a way for the `debug` flag to be set on a job, but that's probably an API change?

I find it kind of weird that the CLI uses a job ID of `cli` instead of another flag to indicate its environment, but I guess that's a separate issue.

updater/spec/dependabot/environment_spec.rb
brrygrdn: Ah, this is an RSpec-isim, `be_debug_enabled` is translated to `environment.debug_enabled? to be true` under the hood.

updater/lib/dependabot/environment.rb
```diff
@@ -35,14 +42,20 @@ def self.job_definition
...
       return ENV.fetch(variable_name, default) unless default == :_undefined
 
       ENV.fetch(variable_name) do
         raise ArgumentError, "Missing environment variable #{variable_name}"
       end
     end
 
-    private_class_method :environment_variable
+    private_class_method def self.job_debug_enabled?
+      !!job_definition.dig("job", "debug")
```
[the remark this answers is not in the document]
brrygrdn: The "double-bang operator" is an idiom in Ruby to force any type into Boolean, I do always feel odd about using it over the `present?` helper - but that's a Rails-isim we don't have in Core as we don't want to include ActiveSupport anymore.

brrygrdn:
> This all makes sense so far. I think you'll also need to define a way for the debug flag to be set on a job, but that's probably an API change?

Yeah, this should be everything we need in Core, we can add debug flag setters to the various entry points like the `dry-run.rb` script. I've punted on that for now as we haven't started adding any `.debug` calls so we can probably just start using it as a dev tool and see how we get on for now

PR #6811, opened 2023-03-10
Diff step should only run when smoke tests have actually run

deivid-rodriguez:


landongrindheim (approved):

jakecoffman:
But won't this cause it to not run if the previous step failed? We might have to add something like `&& (success() || failure())`.

deivid-rodriguez:
Oh, you're right! I think we can merge both steps, then? Capture the smoke test status. If it failed, print the diff. And then exit with the original smoke test status.

deivid-rodriguez:
Some questions? Do we want the diff to run in the success case? Also, should printing a diff on failure when the "test" command fails actually be a CLI feature?

jakecoffman:
> Do we want the diff to run in the success case?

I did like this as it gives an independent sanity check that the CLI assertions are working correctly.

> Also, should printing a diff on failure when the "test" command fails actually be a CLI feature?

We had this as a feature, but the diff style with the library we used was not great for certain changes. We could add it back in to the CLI if we can generate a nice diff. I just haven't prioritized that at all.

deivid-rodriguez:
> I did like this as it gives an independent sanity check that the CLI assertions are working correctly.

What would be a case where this would help? If smoke tests succeeds, but the diff is not empty, should we know about that? I _think_ that may be already happening, but it's invisible to us because `continue-on-error: true` shows a green check regardless (I wish it showed as orange or something).

deivid-rodriguez:
I'm closing this since it's unclear to me whether it was a good change or not. If I run into issues again I can revisit!

PR #6812, opened 2023-03-10
Remove unnecessary stderr capturing

deivid-rodriguez:
I think specs pass without this, and I don't see any unwanted output logged when
tests are run which is another reason I could think of why this is needed.


landongrindheim:
@deivid-rodriguez I wasn't expecting to see [specs failing](https://github.com/dependabot/dependabot-core/actions/runs/4385005319/jobs/7677171543) as a result of this diff. Do you know if they're unrelated failures, or if we were swallowing errors without realizing it?

deivid-rodriguez:
I have no idea, I wasn't expecting it either. I can't repro the failures locally 🤔.

deivid-rodriguez:
I confirmed that common tests are already broken, not sure why/how, but shouldn't be related to this patch.

Nishnha:
> I confirmed that common tests are already broken, not sure why/how, but shouldn't be related to this patch.

This should be fixed now if you want to rebase the branch

landongrindheim (approved):
That's a little more like it 😄

Issue #3723, opened 2021-05-14
Support for private terraform modules


wagnst:
Please support private terraform modules hosted in registries that need authentication:

Running the latest dependabot (which now supports HCL 2 - #1176), reveals:
- `terraform-google-modules/project-factory/google` is public, working fine
- `lhsystems/serviceaccount/google` is a private module hosted on Terraform Enterprise 

Its called like this:

```
module "svc" {
  source  = "app.terraform.io/lhsystems/serviceaccount/google"
  version = "0.0.28"
}
```

The dependabot logs show:
```
updater | INFO <job_134426693> Checking if terraform-google-modules/project-factory/google  needs updating
  proxy | 2021/05/14 22:26:51 [034] GET https://registry.terraform.io:443/v1/modules/terraform-google-modules/project-factory/google/versions
  proxy | 2021/05/14 22:26:51 [034] 200 https://registry.terraform.io:443/v1/modules/terraform-google-modules/project-factory/google/versions
updater | INFO <job_134426693> Latest version is 10.3.2
updater | INFO <job_134426693> Requirements to unlock update_not_possible
updater | INFO <job_134426693> Requirements update strategy 
updater | INFO <job_134426693> No update possible for terraform-google-modules/project-factory/google 
updater | INFO <job_134426693> Checking if lhsystems/serviceaccount/google  needs updating
updater | INFO <job_134426693> Latest version is 
updater | INFO <job_134426693> No update needed for lhsystems/serviceaccount/google 
```

Dependabot needs to authenticate against the private module registry e.g. via `terraform login` or via the API token (`TF_API_TOKEN`) which can be supplied via ENV var, some example is also described [here](https://learn.hashicorp.com/tutorials/terraform/github-actions)

feelepxyz:
@wagnst 💯  we're working on adding support right now 👍

xlgmokha:
We shipped support for Terraform private registries. You can find out more in https://github.com/github/docs/pull/6808, https://github.com/dependabot/dependabot-core/pull/3790, https://github.com/dependabot/dependabot-core/pull/3821, https://github.com/dependabot/dependabot-core/pull/3811, https://github.com/dependabot/dependabot-core/pull/3790, https://github.com/dependabot/dependabot-core/pull/3756.

Please give it a try and let us know what you think. https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#terraform-registry

[...]

PR #6828, opened 2023-03-13
Terraform: fix NoMethodError due to no versions in module

jakecoffman:
We're seeing this error quite a bit in the Terraform ecosystem. 

```
NoMethodError: undefined method `fetch' for nil:NilClass

          fetch("modules").first.fetch("versions").
                                ^^^^^^
  from terraform/lib/dependabot/terraform/registry_client.rb:83:in `all_module_versions'
  from terraform/lib/dependabot/terraform/update_checker.rb:73:in `all_module_versions'
  from terraform/lib/dependabot/terraform/update_checker.rb:64:in `latest_version_for_registry_dependency'
  from terraform/lib/dependabot/terraform/update_checker.rb:18:in `latest_version'
  from dependabot-updater/lib/dependabot/updater.rb:525:in `all_versions_ignored?'
  from dependabot-updater/lib/dependabot/updater.rb:222:in `check_and_create_pull_request'
  from dependabot-updater/lib/dependabot/updater.rb:103:in `check_and_create_pr_with_error_handling'
  from dependabot-updater/lib/dependabot/updater.rb:79:in `block in run'
  from dependabot-updater/lib/dependabot/updater.rb:79:in `each'
  from dependabot-updater/lib/dependabot/updater.rb:79:in `run'
  from dependabot-updater/lib/dependabot/update_files_command.rb:17:in `perform_job'
  from dependabot-updater/lib/dependabot/base_command.rb:50:in `run'
  from bin/update_files.rb:23:in `<main>'
```

It seems to be due to the module response returning no modules: `"modules": []`.

So I've added a specific nil check to throw a better error.

### DIFF
--- a/terraform/lib/dependabot/terraform/registry_client.rb
+++ b/terraform/lib/dependabot/terraform/registry_client.rb
@@ -80,7 +80,10 @@ module Dependabot
         response = http_get!(URI.join(base_url, "#{identifier}/versions"))
 
         JSON.parse(response.body).
-          fetch("modules").first.fetch("versions", []).
+          fetch("modules").
+          first.
+          tap { |mod| raise error("Registry returned no modules") if mod.nil? }.
+          fetch("versions", []).
           map { |release| version_class.new(release.fetch("version")) }
       end
 
--- a/terraform/lib/dependabot/terraform/registry_client.rb
+++ b/terraform/lib/dependabot/terraform/registry_client.rb
@@ -80,7 +80,7 @@ module Dependabot
         response = http_get!(URI.join(base_url, "#{identifier}/versions"))
 
         JSON.parse(response.body).
-          fetch("modules").first.fetch("versions").
+          fetch("modules").first.fetch("versions", []).
           map { |release| version_class.new(release.fetch("version")) }
       end
 
--- a/terraform/spec/dependabot/terraform/registry_client_spec.rb
+++ b/terraform/spec/dependabot/terraform/registry_client_spec.rb
@@ -112,7 +112,7 @@ RSpec.describe Dependabot::Terraform::RegistryClient do
     expect(response.max).to eq(Gem::Version.new("0.10.1"))
   end
 
-  it "doesn't raise when the module has no versions" do
+  it "raises an exception when there are no modules" do
     hostname = "registry.terraform.io"
     stub_request(:get, "https://#{hostname}/.well-known/terraform.json").and_return(
       status: 200,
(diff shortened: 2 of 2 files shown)


Nishnha (approved):
That seems like an appropriate fix to me.

jakecoffman:
D'oh, I read the stack trace wrong. `first` is returning `nil` so the response is `modules: []`. In this case I think raising an exception seems appropriate?

terraform/lib/dependabot/terraform/registry_client.rb
```diff
@@ -80,7 +80,7 @@ def all_module_versions(identifier:)
         response = http_get!(URI.join(base_url, "#{identifier}/versions"))
 
         JSON.parse(response.body).
-          fetch("modules").first.fetch("versions").
+          fetch("modules").first.fetch("versions", []).
```
landongrindheim, line 83: Depending on what you're after, you could return the object you would expect to reference 👇 
```suggestion
          fetch("modules", [{}]).first.fetch("versions", []).
```

If this is something we shouldn't encounter/handle, you could raise an error 👇  

```ruby
  fetch("modules") { raise NoModulesError }
  .first
  .fetch("versions")

jakecoffman: Actually that doesn't work here because the `modules` property does exist, it's just an empty list. So I used `tap` to check for `nil`.

jeffwidman:
@HorizonNet any thoughts since you've been helping curate Terraform in here? (which we really appreciate BTW! 👍 )

HorizonNet:
@jeffwidman I didn't run into this one yet. What's the precondition for it? Do you have an example for a request where you run into this one? If tried it with https://registry.terraform.io/v1/modules/cloudposse/label/null/versions and few others, but couldn't find a fitting example.

With the public registry I'm not even sure if it's possible to have a module without a version in the registry. What would happen with `versions: []`? Kind of assuming that you're already on the latest version of the module?

jakecoffman:
I tried finding an open source example, but now realize all of the errors are coming from a specific org and the dependencies are private and going through https://app.terraform.io/api/registry/v1/modules. So we can't see the response from the server. We can infer that it's returning a `modules: []` because it's causing `.first` to return `nil`. I originally interpreted that wrong but updated the description.

So, somehow this seems to be possible to get the registry to return this, but probably the best thing we can do is raise a more explicit error. The one posted above makes it seem like a problem in Dependabot, when it's actually the registry is in an odd state.

HorizonNet:
This looks to me like there's a module referenced from a Terraform Cloud private registry, which doesn't exist or the credentials passed to Dependabot don't have access to it, meaning that the login works (and that's why there's no 401), but access to the module is missing (that's why it returns empty instead of a 401).

I, as a user, would expect Dependabot to do the following in that case

1. Assume that I'm already on the latest version.
2. Add a warning in the logs like "Couldn't update module/abc as no modules have been returned from registry.xyz".
3. Add a point in the documentation pointing people how to test it on their own and that they eventually should reach out to the module maintainer or to the registry provider.

In that case Dependabot cannot do much about it. That's why I also would go with a warning instead of an error.

I also saw that the [documentation](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#terraform-registry) is a little bit misleading when it comes to private registries. Probably would be good to add an example for Terraform Cloud private registries like [this one](https://github.com/dependabot/dependabot-core/issues/3723#issuecomment-858602012). I don't think that this is the reasone here as then there should be a 401 flying around somewhere.

jakecoffman:
I think if the credentials were valid but they didn't have access it would return a 404 which is what I get when I try to reach their module with an account I made. 

Since this is affecting just one customer and the response from the server being surprising I am going to wait until I have more information before proceeding. 

Thanks for helping with this @HorizonNet!

PR #6830, opened 2023-03-14
Don't depend on flake8 at runtime

deivid-rodriguez:
It should avoid avoidable resolution issues when installing old Python's
dinamically


python/script/ci-test
```diff
@@ -2,6 +2,8 @@
 
 set -e
 
-bundle install
+pyenv exec pip --disable-pip-version-check install --use-pep517 -r "dev-requirements.txt"
 pyenv exec flake8 helpers/. --count --exclude=./.*,./python/spec/fixtures --show-source --statistics
+
+bundle install
```
jeffwidman, line 8: This move is unrelated and just aesthetic cleanup, right?

python/helpers/dev-requirements.txt
```diff
@@ -0,0 +1,2 @@
+pip>=21.3.1,<23.1.0  # Range maintains py36 support TODO: Review python 3.6 support in April 2023 (eol ubuntu 18.04)
```
jeffwidman, line 1: Kinda hate to see this duplicated since it's got so much custom specification / comment metadata...easy for that to go out of sync.

Another way to go is via a top-level `requirements.txt` that installs `pip`, which is then inherited by `dev-requirements.txt` / `prod-requirements.txt`: https://stackoverflow.com/a/20720019/770425

python/script/ci-test
```diff
@@ -2,6 +2,8 @@
 
 set -e
 
-bundle install
+pyenv exec pip --disable-pip-version-check install --use-pep517 -r "dev-requirements.txt"
```
jeffwidman, line 5: Could we simply do:
```shell
pyenv exec pip --disable-pip-version-check install --use-pep517 flake8==5.0.4
```

That'd remove the need for `dev-requirements.txt` unless/until we start adding additional packages there...

jeffwidman (commented):
I hadn't realized we were installing this in our prod requirements.

python/helpers/dev-requirements.txt
deivid-rodriguez: Yeah, I thought of that but since this is a single dependency that should be kept in sync by Dependabot anyways, I figured it was not worth the trouble. I can do it though if you feel strongly about it.

python/script/ci-test
deivid-rodriguez: Yep, I also though of that, but then that dependency won't be watched but Dependabot and will fall out of date.

python/helpers/dev-requirements.txt
deivid-rodriguez: I included this because I don't want a version of flake8 that does not support our oldest supported Python to be installed, but maybe this concern does not make sense in Python or is overkill. And we already already doing it for flake==5.0.4 anyways, which does not support Python 3.6.0 and below. But my idea was to set the example for _after_ we drop Python 3.6 support.

The concern is that if we install a flake8 version that does not support our minimum supported Python, that could potentially suggest some syntax for our native helpers that's only compatible with newer Pythons and would break when native helpers are run by older Pythons.

In any case, even if the concern makes sense, I don't think a requirement on pip makes sense to prevent that? I'm pushing an alternative approach.

python/script/ci-test
deivid-rodriguez: I reverted to the previous order 👍.

python/script/ci-test
jeffwidman: good thought, probably better to keep it in a `*requirements.txt` file then...

python/helpers/dev-requirements.txt
jeffwidman: > The concern is that if we install a flake8 version that does not support our minimum supported Python, that could potentially suggest some syntax for our native helpers that's only compatible with newer Pythons and would break when native helpers are run by older Pythons.

Oh, good point! Given that, it's not a bad thing that this is triggering a failure, it's just we need to realize/fix the failure sooner.

I'd probably consider even just lowering our flake8 pin to one that works with `python 3.6.0` and call it good until we drop 3.6.0...

mctofu (approved):

mctofu (approved):

PR #6839, opened 2023-03-14
Adjust poetry/pyproject python version to match installed in version resolver

mctofu:
Resolves this error when checking for updates when a specific python version is pinned in pyproject.toml. We want it to match the version that pre-installed with the updater rather than dynamically installing a different patch release.

```
Dependabot::SharedHelpers::HelperSubprocessFailed: The currently activated Python version 3.11.2 is not supported by the project (3.11.1). Trying to find and use a compatible version.

Poetry was unable to find a compatible version. If you have one, you can explicitly use it via the "env use" command.
 ./lib/dependabot/python/update_checker/poetry_version_resolver.rb:349:in `run_poetry_command'
 ./lib/dependabot/python/update_checker/poetry_version_resolver.rb:167:in `run_poetry_update_command'
```

### DIFF
--- a/python/lib/dependabot/python/update_checker/poetry_version_resolver.rb
+++ b/python/lib/dependabot/python/update_checker/poetry_version_resolver.rb
@@ -227,6 +227,7 @@ module Dependabot
         def updated_pyproject_content(updated_requirement:)
           content = pyproject.content
           content = sanitize_pyproject_content(content)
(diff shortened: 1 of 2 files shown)


jeffwidman (approved):
Nice catch!

pavera (approved):

PR #6843, opened 2023-03-15
Fix retrieval of private registry metadata from the Job definition

brrygrdn:
Fixes a bug introduced in #6810 

The modified code in this PR attempted to retrieve the `credential-metadata` using the wrong key name by confuting the pre- and post-standardisation use of kebab- and snake-case respectively, i.e. `credential_metadata` instead.

This change reverts to the original call order to ensure that the retrieval happens after standardisation to guard against any future changes to key manipulation leaving the credential_metadata clause out of alignment with the rest of the payload.

### DIFF
--- a/updater/lib/dependabot/job.rb
+++ b/updater/lib/dependabot/job.rb
@@ -58,10 +58,11 @@ module Dependabot
     end
 
     def self.new_update_job(job_id:, job_definition:, repo_contents_path: nil)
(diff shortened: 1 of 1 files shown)


jakecoffman (approved):

brrygrdn:
This match does not seem to have fully solved the issue, taking another look at it now

PR #6844, opened 2023-03-15
Fix the call order when pulling through cred metadata

brrygrdn:
Fast-follow on https://github.com/dependabot/dependabot-core/pull/6843

The previous fix only remediated for the wrong key case being used, but by regressing the call order it didn't account for the fact that the `slice` operation no longer includes the `credentials` key.

This is a split-the-difference fix which uses the new call order so the `credentials` are injected to a filtered attribute list that uses a single `PERMITTED_KEYS` list but are pulled from a version of the hash that is pre-filtered **and** standardised.

### DIFF
--- a/updater/spec/dependabot/job_spec.rb
+++ b/updater/spec/dependabot/job_spec.rb
@@ -63,6 +63,30 @@ RSpec.describe Dependabot::Job do
   let(:commit_message_options) { nil }
   let(:vendor_dependencies) { false }
 
+  describe "::new_update_job" do
+    let(:job_json) { fixture("jobs/job_with_credentials.json") }
+
+    let(:new_update_job) do
+      described_class.new_update_job(
+        job_id: anything,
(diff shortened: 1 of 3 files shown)


updater/spec/dependabot/job_spec.rb
```diff
@@ -63,6 +63,30 @@
   let(:commit_message_options) { nil }
   let(:vendor_dependencies) { false }
 
+  describe "::new_update_job" do
+    let(:job_json) { fixture("jobs/job_with_credentials.json") }
```
brrygrdn, line 67: [Fixture](https://github.com/dependabot/dependabot-core/blob/ce38f0d9b5e6004eb97041a25157ce6332bfc7c7/updater/spec/fixtures/jobs/job_with_credentials.json)

This file includes both `credentials` and `credentials-metadata` which isn't representative of the kind of files we actually inject, but it is a good way to make sure we use the metadata by precedent if both are present.

PR #6846, opened 2023-03-15
[Updater] Prefer to use `Dependabot::Service` directly and as an application error reporter

brrygrdn:

Previous PR: https://github.com/dependabot/dependabot-core/pull/6810

The `Dependabot::Updater` class implements quite a lot of private methods to assist error handling. As we start to break the updater out into individual strategy classes which handle discrete operations, many of them need to share these methods.

This PR clears the way to start breaking the logic apart primarily via three changes:
- Stop maintaining a cache of errors inside `Dependabot::Updater`. This isn't used in any significant way, we can just use `Dependabot::Service#errors` as a substitute in the one place we need to check if any errors have happened.
- Prefer to use `service. record_update_job_error` _directly_ instead of creating an "error_detail" hash to use the `Dependabot::Updater#record_error` private method, which has been removed
- Refactor error capturing via Sentry/Raven unto the `Dependabot::Service` class so it has an explicit receiver that binds it to the backend service and accepts `Dependabot` objects as arguments

This will allow each individual strategy class to be passed the service so the various `record_foo_error` helper private methods can be moved without dragging more duplicate private methods in.

### Other Changes

There are a few small tidy ups made alongside these changes:
- Remove an unused `error_context` method on `Dependabot::Updater`
- Remove unused accessors on `Dependabot::Service`, notably disallow using its `Dependabot::ApiClient` directly

### DIFF
--- a/updater/lib/dependabot/service.rb
+++ b/updater/lib/dependabot/service.rb
@@ -42,6 +42,25 @@ module Dependabot
       client.record_update_job_error(error_type: error_type, error_details: error_details)
     end
 
+    # This method wraps the Raven client as the Application error tracker
+    # the service uses to notice errors.
+    #
+    # This should be called as an alternative/in addition to record_update_job_error
+    # for cases where an error could indicate a problem with the service.
+    def capture_exception(error:, job: nil, dependency: nil, tags: {}, extra: {})
+      Raven.capture_exception(
+        error,
+        {
+          tags: tags,
+          extra: extra.merge({
+            update_job_id: job&.id,
+            package_manager: job&.package_manager,
+            dependency_name: dependency&.name
+          }.compact)
+        }
+      )
+    end
+
     def noop?
       pull_requests.empty? && errors.empty?
     end
--- a/updater/lib/dependabot/service.rb
+++ b/updater/lib/dependabot/service.rb
@@ -3,13 +3,16 @@
 require "terminal-table"
 require "dependabot/api_client"
 
-# Wraps an API client with the current state of communications with the Dependabot Service
-# and provides an interface to summarise all actions taken.
+# This class provides an output adapter for the Dependabot Service which manages
+# communication with the private API as well as consolidated error handling.
+#
+# Currently this is the only output adapter available, but in future we may
+# support others for use with the dependabot/cli project.
 #
 module Dependabot
   class Service
     extend Forwardable
-    attr_reader :client, :events, :pull_requests, :errors
+    attr_reader :pull_requests, :errors
 
     def initialize(client:)
       @client = client
@@ -70,6 +73,8 @@ module Dependabot
     end
 
     private
+
+    attr_reader :client
 
     def pull_request_summary
       return unless pull_requests.any?
--- a/updater/lib/dependabot/updater.rb
+++ b/updater/lib/dependabot/updater.rb
@@ -786,6 +786,7 @@ module Dependabot
       service.close_pull_request(job.dependencies, reason)
     end
 
+    # rubocop:disable Metrics/AbcSize
     # rubocop:disable Metrics/MethodLength
     def handle_dependabot_error(error:, dependency:)
       error_details =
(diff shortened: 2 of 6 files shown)


updater/lib/dependabot/service.rb
```diff
@@ -39,6 +42,25 @@ def record_update_job_error(error_type:, error_details:, dependency: nil)
...
+    def capture_exception(error:, job: nil, dependency: nil, tags: {}, extra: {})
+      Raven.capture_exception(
+        error,
+        {
+          tags: tags,
+          extra: extra.merge({
+            update_job_id: job&.id,
+            package_manager: job&.package_manager,
+            dependency_name: dependency&.name
+          }.compact)
```
landongrindheim, line 59: 🔎 I understand the impulse to send only those key/value pairs for which we have data, but I wonder if the absence of values is valuable for understanding context.

As a 🤏 cheap argument, I'd be able to glean much more insight from:

```ruby
{
  update_job_id: 1,
  dependency_name: "dependabot-fortran",
  package_manager: nil
}
```

than
```ruby
{
  update_job_id: 1,
  dependency_name: "dependabot-fortran"
}
```

updater/lib/dependabot/updater.rb
```diff
@@ -60,8 +60,7 @@ def initialize(service:, job:, dependency_files:, base_commit_sha:)
       @job = job
       @dependency_files = dependency_files
       @base_commit_sha = base_commit_sha
-      # TODO: Collect @created_pull_requests and @errors on the Job object?
-      @errors = []
+      # TODO: Collect @created_pull_requests on the Job object?
```
landongrindheim, line 63: ✅ 👏

updater/lib/dependabot/updater.rb
```diff
@@ -81,18 +80,20 @@ def run
           "Dependabot::AllVersionsIgnored was unexpectedly raised for a non-security update job"
         )
         error.set_backtrace(e.backtrace)
-        Raven.capture_exception(error, raven_context)
+        service.capture_exception(error: error, job: job)
```
landongrindheim, line 83: 📈 This is *such* a nice improvement!

updater/lib/dependabot/updater.rb
```diff
@@ -81,18 +80,20 @@ def run
...
         return
       end
 
       # OOM errors are special cased so that we stop the update run early
-      error = { "error-type": RUN_HALTING_ERRORS.fetch(e.class) }
-      record_error(error)
+      service.record_update_job_error(
+        error_type: RUN_HALTING_ERRORS.fetch(e.class),
```
landongrindheim, line 89: 🔎 I see we weren't handling this before, so no need to change anything. Should we provide a fallback value for `Hash#fetch` here?

updater/lib/dependabot/updater.rb
```diff
@@ -885,11 +891,16 @@ def handle_dependabot_error(error:, dependency:)
...
     # rubocop:enable Metrics/MethodLength
+    # rubocop:enable Metrics/AbcSize
+
     # rubocop:disable Metrics/MethodLength
-    # rubocop:disable Metrics/CyclomaticComplexity
     def handle_parser_error(error)
+      # This happens if the repo gets removed after a job gets kicked off.
+      # The service will handle the removal without any prompt from the updater,
+      # so no need to add an error to the errors array
+      return if error.is_a? Dependabot::RepoNotFound
```
landongrindheim, line 902: This feels like the right place for this 👍

landongrindheim (approved):
🥇 Again, I'm loving the direction in which you're pushing our components to work together 😍

updater/lib/dependabot/service.rb
brrygrdn: That's fair, I think we should actually expect `dependency_name` to be passed/present move often -but- I'm err'ing on the side of maintaining existing behaviour for now so I'd prefer to pass on this to avoid test churn for now.

updater/lib/dependabot/updater.rb
brrygrdn: I think we probably want it to fail loud as we, ideally, should never enter this method unless we already checked the subject `error` was present in `RUN_HALTING_ERRORS` to begin with.

This code might be able to fold up into the call site in another pass?

PR #6847, opened 2023-03-15
[Updater] Start splitting the Updater by extracting DependencySnapshot as an input class

brrygrdn:

Depends on https://github.com/dependabot/dependabot-core/pull/6846

A Dependabot update process has three broad steps:

1. Fetch the files we need from the repository
2. Parse the fetched files into a set of dependency information Dependabot can work with
3. Iterate over the dependency update to make changes to the fetched files which are pushed back to the repo

Currently, step(1) is performed as a single command so we do not need to provide repository credentials to the Update step and it hands files over to step(2) by encoding them into a 'job definition' along with the workload description it used to fetch them.

As we start to break the `Dependabot::Updater` class out into a set of components we can compose to create grouped updates, one difficulty is that step(2) is performed lazily inside `Dependabot::Updater#dependencies` which ostensibly functions like a getter method but in reality mixes the responsibility of parsing the files with filtering them according to the update strategy to be performed.

### Changes

This PR introduces a new `DependencySnapshot` object which encapsulates the responsibility of retrieving and decoding the files from step(2) and correctly hydrating them into `Dependabot::Dependency` objects to result in a value class which describes the starting state of the update.

### Rationale

This value class is used like a shim right now, with the parsing still being performed lazily inside `Dependabot::Updater#dependencies` in order to minimise the blast radius of refactoring the code, so initially it only provides value by reducing the number of arguments `Dependabot::Updater` requires, but it sets us up for two fast-follows:
- We can now start splitting out "operation" classes which take on a subset of the overall `Dependabot::Updater` class responsibilities which accept this object as an argument and proactively evaluate it to start refactoring the call site and flushing out assumptions about performing the parse/filter in a single method.
- Once we've figured out how to remove some implicit assumptions around the parse/filter method call order, we can push the responsibility of handling parser errors entirely out of the Updater into a discrete precondition which will extract a test surface out of `Dependabot::Updater` and make it easier to reason about.

### DIFF
--- a/updater/lib/dependabot/dependency_snapshot.rb
+++ b/updater/lib/dependabot/dependency_snapshot.rb
@@ -1,0 +1,81 @@
+# frozen_string_literal: true
+
+require "base64"
+require "dependabot/file_parsers"
+
+# This class describes the dependencies obtained from a project at a specific commit SHA
+# including both the Dependabot::DependencyFile objects at that reference as well as
+# means to parse them into a set of Dependabot::Dependency objects.
+#
+# This class is the input for a Dependabot::Updater process with Dependabot::DependencyChange
+# representing the output.
+module Dependabot
+  class DependencySnapshot
+    # TODO: Enforce non-nil values for job_definition["base64_dependency_files"]
+    #       and job_definition["base_commit_sha"]
+    #
+    # We historically tolerate nil values for both these keys from the `job_definition`
+    # but it doesn't seem like we should. Rather than introduce a behaviour change
+    # as part of the change introducing this class, let's do it as a follow-up.
+    def self.create_from_job_definition(job:, job_definition:)
+      decoded_dependency_files = job_definition.fetch("base64_dependency_files", []).map do |a|
+        file = Dependabot::DependencyFile.new(**a.transform_keys(&:to_sym))
+        file.content = Base64.decode64(file.content).force_encoding("utf-8") unless file.binary? && !file.deleted?
+        file
+      end
+
+      new(
+        job: job,
+        base_commit_sha: job_definition.fetch("base_commit_sha", nil),
+        dependency_files: decoded_dependency_files
+      )
+    end
+
+    attr_reader :base_commit_sha, :dependency_files
+
+    def initialize(job:, base_commit_sha:, dependency_files:)
+      @job = job
+      @base_commit_sha = base_commit_sha
+      @dependency_files = dependency_files
+    end
+
+    def dependencies
+      return @dependencies if defined?(@dependencies)
+
+      parse_files!
+    end
+
+    private
+
+    attr_reader :job
+
+    # TODO: Parse files during instantiation?
+    #
+    # To avoid having to re-home Dependabot::Updater#handle_parser_error,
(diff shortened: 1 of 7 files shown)


updater/lib/dependabot/service.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "raven"
```
landongrindheim, line 3: 🔎 Is this a missed import from a previous PR?

updater/lib/dependabot/dependency_snapshot.rb
```diff
@@ -0,0 +1,81 @@
...
+require "dependabot/file_parsers"
+
+# This class describes the dependencies obtained from a project at a specific commit SHA
+# including both the Dependabot::DependencyFile objects at that reference as well as
+# means to parse them into a set of Dependabot::Dependency objects.
+#
+# This class is the input for a Dependabot::Updater process with Dependabot::DependencyChange
+# representing the output.
+module Dependabot
+  class DependencySnapshot
```
landongrindheim, line 13: First of all, I really dig the clear lines that are being drawn in the updater 🎉 I think these recent changes just make sense 😄 

💅 I wonder if we can come up with a stronger signal for this being an input. The `DependencyChange` name suggests that its responses will represent output, but the `DependencySnapshot` name doesn't resonate _with me_ as an input per se. Maybe that's the way it should be, and the names of variables should be holding meaning.

No need to update this, but if that resonates with you, I'm happy to try to work with you to come up with a name that reflects the before state of the dependency file 🙂

landongrindheim (approved):
🚀

updater/lib/dependabot/service.rb
brrygrdn: Yeah, it should have moved in a previous PR. This file is always included after `require "raven"` in the [binstub](https://github.com/dependabot/dependabot-core/blob/main/updater/bin/update_files.rb#L7) so it didn't cause any issues.

updater/lib/dependabot/dependency_snapshot.rb
brrygrdn: Yeah, I'm not completely happy with this name but I also don't think the responsibilities of this class are fully final so I'd prefer to revisit this after a few more transforms once the parsing is properly contained

brrygrdn:
I've added one additional commit which uses the new class to push the `update_dependency_list` private method out the Updater unto the Service.

The Service being responsible for serialising a business object ( DependencySnapshot ) for the ApiClient feels more correct and moves another private method unto an explicit receiver so operation classes can access it without duplicating it. It makes more sense to do that as part of this PR than in a follow up.

updater/lib/dependabot/updater.rb
```diff
@@ -608,7 +608,7 @@ def dependencies
       all_deps = dependency_snapshot.dependencies
 
       # Tell the backend about the current dependencies on the target branch
-      update_dependency_list(all_deps)
+      service.update_dependency_list(dependency_snapshot: dependency_snapshot)
```
landongrindheim, line 611: ⏩

landongrindheim (approved):
Looks good to me! 🚀

PR #6853, opened 2023-03-16
fix Faraday find_proxy slowness

jakecoffman:
I noticed significant delay at the start of gathering data for the PR. I tracked this down to Faraday's feature of finding a proxy in the environment. For whatever reason it's very slow for us in our smoke tests, and possibly production. 

To fix it, I've set the proxy explicitly from the environment, if it's set there. If it's not, it will again fall back to the old behavior. So if anyone is relying on that, it should still work.

### DIFF
--- a/common/lib/dependabot/clients/github_with_retries.rb
+++ b/common/lib/dependabot/clients/github_with_retries.rb
@@ -89,6 +89,12 @@ module Dependabot
         access_tokens << nil if access_tokens.empty?
         access_tokens.uniq!
 
+        # Explicitly set the proxy if one is set in the environment
+        # as Faraday's find_proxy is very slow.
+        Octokit.configure do |c|
+          c.proxy = ENV["HTTPS_PROXY"] if ENV["HTTPS_PROXY"]
+        end
+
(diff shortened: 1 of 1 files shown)


jakecoffman:
Oh wow, that seems to have made quite a difference. Comparing this [with](https://github.com/dependabot/dependabot-core/actions/runs/4440144288/jobs/7793570864) and [without](https://github.com/dependabot/dependabot-core/actions/runs/4439770410/jobs/7792691253):

| test      | before | after|
| ----------- | ----------- |-------------|
| bundler      | 5:39       | 1:55 |
| go   | 6:54        | 2:34 |
| terraform | 9:12 | 2:32 |

Nishnha:
For reference this is where we use find_proxy https://github.com/dependabot/dependabot-core/blob/a2afea4f8d4a6bf52f3e46f0603d31dbbd56452b/updater/lib/dependabot/api_client.rb#L146

Nishnha (approved):

mctofu (approved):

deivid-rodriguez:
This us soooooooo awesome! Thank you @jakecoffman! 😍

PR #6857, opened 2023-03-16
run all CI on Gemfile or Gemfile.lock change

jakecoffman:
One of our tests starting failing out of nowhere and slowed us down a bit. We found this was due to a fairly minor Octokit change in https://github.com/dependabot/dependabot-core/pull/6826. 

To catch this sort of thing sooner, we should be running all CI whenever the updater Gemfile or lock changes. 

I wasn't sure if `updater/Gemfile*` would pick up both, so I opted for `updater/Gemfil*`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,58 +55,68 @@ jobs:
             bundler:
               - Dockerfile.updater-core
               - 'common/**'
+              - 'updater/Gemfil*'
               - 'omnibus/**'
               - '.github/workflows/ci.yml'
               - 'bundler/**'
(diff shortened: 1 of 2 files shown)


Nishnha (approved):

Nishnha:
The filters managed by this action: https://github.com/dorny/paths-filter

Not sure if the `*` glob is supported but might as well try it out?

pavera (approved):

Nishnha:
Rebased the branch so it includes the changes from https://github.com/dependabot/dependabot-core/pull/6855

PR #6859, opened 2023-03-17
avoid find_proxy which is slow

jakecoffman:
This is the same change as in #6853. @Nishnha noted that there was another `find_proxy` I missed. Turned out that it's probably the same issue, and by avoiding it we save even more time!

On my machine it made the Go test go from 1:35 to :25! 🤯


mctofu (approved):
🚤

honeyankit (approved):

PR #6861, opened 2023-03-17
add timestamp to logger

jakecoffman:
To help track down freezes and performance issues it's helpful if every line has a consistent timestamp. This matches the timestamp format that the proxy uses. 

Before:

```
updater | I, [2023-03-17T11:39:57.488735 #911]  INFO -- sentry: ** [Raven] Raven 3.1.2 configured not to capture errors: DSN not set
updater | INFO Starting job processing
```

After:

```
updater | 2023/03/17 11:43:56 INFO Raven 3.1.2 configured not to capture errors: DSN not set
updater | 2023/03/17 11:43:58 INFO Starting job processing
```


landongrindheim (approved):
📆⌚⏱️

PR #6865, opened 2023-03-17
[Updater] Break out handler methods for Dependabot::Error classes into a small utility class

brrygrdn:
Follows up on https://github.com/dependabot/dependabot-core/pull/6847 which must merge first

This is small and somewhat imperfect extract-to-class refactor to move aside some long methods we use for presenting `Dependabot::Errors` via the `Dependabot::Service` and/or logging.

As we split out distinct operator classes, duplicating these methods is really undesirable as they have a lot of specific behaviour that would be hard to keep in alignment.

I think there's a deeper refactor to be had here since methods ultimately derive an "error detail" hash based on the subclass of `Dependabot::Error` being handled, e.g.

```ruby
{
  "error-type": "private_source_timed_out",
  "error-detail": { source: error.source }
}
```

I think this could just become a method on each `Dependabot::Error` subclass, but that would require a significant change in core with substantial testing to be confident.

Likewise, I think the `handle_dependabot_error` and `handle_parser_error` could be DRY'd out, but the parsing itself should be pulled up out of the Updater so doing a deeper change at this time might be counter-productive.

### DIFF
--- a/updater/lib/dependabot/updater/error_handler.rb
+++ b/updater/lib/dependabot/updater/error_handler.rb
@@ -1,0 +1,246 @@
+# frozen_string_literal: true
+
+# This class is responsible for determining how to present a Dependabot::Error
+# to the Service and Logger.
+#
+# TODO: Iterate further on leaner error handling
+#
+# This class is a coarse abstraction of some shared logic that has several flags
+# against it from Rubocop we aren't addressing right now.
+#
+# It feels like this concern could be slimmed down if each Dependabot::Error
+# class implemented a "presenter" method to generate it's own `error-type` and
+# `error-detail` since this never draws attributes from the Updater context.
+#
+# For now, let's just extract it and set it aside as a tangent from the critical
+# path.
+module Dependabot
+  class Updater
+    class ErrorHandler
+      # These are errors that halt the update run and are handled in the main
+      # backend. They do *not* raise a sentry.
+      RUN_HALTING_ERRORS = {
+        Dependabot::OutOfDisk => "out_of_disk",
+        Dependabot::OutOfMemory => "out_of_memory",
+        Dependabot::AllVersionsIgnored => "all_versions_ignored",
+        Dependabot::UnexpectedExternalCode => "unexpected_external_code",
+        Errno::ENOSPC => "out_of_disk",
+        Octokit::Unauthorized => "octokit_unauthorized"
+      }.freeze
+
+      def initialize(service:, job:)
+        @service = service
+        @job = job
+      end
+
+      # rubocop:disable Metrics/AbcSize
+      # rubocop:disable Metrics/MethodLength
+      def handle_dependabot_error(error:, dependency:)
+        # If the error is fatal for the run, we should re-raise it rather than
+        # pass it back to the service.
+        raise error if RUN_HALTING_ERRORS.keys.any? { |err| error.is_a?(err) }
+
+        error_details =
+          case error
+          when Dependabot::DependencyFileNotResolvable
+            {
+              "error-type": "dependency_file_not_resolvable",
+              "error-detail": { message: error.message }
+            }
+          when Dependabot::DependencyFileNotEvaluatable
+            {
+              "error-type": "dependency_file_not_evaluatable",
+              "error-detail": { message: error.message }
+            }
+          when Dependabot::GitDependenciesNotReachable
+            {
+              "error-type": "git_dependencies_not_reachable",
+              "error-detail": { "dependency-urls": error.dependency_urls }
+            }
+          when Dependabot::GitDependencyReferenceNotFound
+            {
+              "error-type": "git_dependency_reference_not_found",
+              "error-detail": { dependency: error.dependency }
+            }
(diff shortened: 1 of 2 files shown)


updater/lib/dependabot/dependency_snapshot.rb
```diff
@@ -0,0 +1,81 @@
...
+    # returning an empty dependency array in Dependabot::Updater#dependencies
+    # in order to 'fall through' to an error outcome elsewhere in the class.
+    #
+    # Given this uncertainity, and the need to significantly refactor tests,
+    # it makes sense to introduce this shim and then deal with the call
+    # site once we've split out the downstream behaviour in the updater.
+    #
+    def parse_files!
+      @dependencies = dependency_file_parser.parse
+    end
```
landongrindheim, line 68: 🤔 Would it make sense to expose a method that signals that parsing has completed? I understand defining `@dependencies` signals this internally, and that we may not want to have collaborators checking this externally. Maybe something like a result object would be appropriate, given your note about handling parse errors.

None of this is necessary, but _maybe_ it would be useful? I'm interested in your thoughts as we're still paving the path for further changes.

updater/lib/dependabot/update_files_command.rb
```diff
@@ -2,19 +2,43 @@
...
+      # and decode them into an object.
+
+      # TODO: Parse the dependency files when instantiated
+      #
+      # We can pull the error handling for parser exceptions up into this class to
+      # completely remove the concern from Dependabot::Updater.
+      #
+      # This should happen separately to introducing the class as a shim.
+      #
+      # See: updater/lib/dependabot/dependency_snapshot.rb:52
```
landongrindheim, line 22: I think a no-op of sorts, probably in the shape of error forwarding, makes sense here.

updater/lib/dependabot/updater/error_handler.rb
```diff
@@ -0,0 +1,246 @@
...
+          error: error,
+          error_type: error_details.fetch(:"error-type"),
+          error_detail: error_details.fetch(:"error-detail", nil)
+        )
+      end
+      # rubocop:enable Metrics/MethodLength
+      # rubocop:enable Metrics/AbcSize
+
+      # rubocop:disable Metrics/MethodLength
+      def handle_parser_error(error)
```
landongrindheim, line 149: 👍 This feels like a solid place for this logic 🏡

updater/lib/dependabot/updater/error_handler.rb
```diff
@@ -0,0 +1,246 @@
...
+#
+# It feels like this concern could be slimmed down if each Dependabot::Error
+# class implemented a "presenter" method to generate it's own `error-type` and
+# `error-detail` since this never draws attributes from the Updater context.
+#
+# For now, let's just extract it and set it aside as a tangent from the critical
+# path.
+module Dependabot
+  class Updater
+    class ErrorHandler
```
landongrindheim, line 19: I'm appreciating this as a place to put error handling. The reasons for it to change will be limited to new errors and new requirements to handle those errors. That feels right to me, especially compared to the previous location 🙂

updater/spec/dependabot/updater_spec.rb
```diff
@@ -1292,7 +1265,8 @@ def expect_update_checker_with_ignored_versions(versions)
               service = build_service
               updater = build_updater(service: service, job: job)
 
-              allow(updater).to receive(:dependency_files).
+              # TODO: Move this stub unto Dependabot::DependencySnapshot once it is better integrated
+              allow(Dependabot::FileParsers).to receive_message_chain(:for_package_manager, :new, :parse).
```
landongrindheim, line 1269: 👍 

It's gonna be a chore to tackle all these TODOs, but I'm glad you're leaving them so we can take care of them.

landongrindheim (approved):
🤩

updater/lib/dependabot/dependency_snapshot.rb
brrygrdn: > we may not want to have collaborators checking this externally.

I think the call order we have right now means we _can_ proceed if parsing fails, but from reasoning through the code I don't think we ever actually should? i.e. there shouldn't be any collaboration if parsing fails, it's always a sign to stop the job.

https://github.com/dependabot/dependabot-core/blob/8f9e4734020e0029a9ec53e1d38163bbcc6235c7/updater/lib/dependabot/updater.rb#L71-L77

In the existing behaviour, if parsing fails, we get an empty value for "dependencies" at the code above.

For a new update, `[]` means we don't iterate and the job silently ends.

For updating an existing PR, we end up here:
https://github.com/dependabot/dependabot-core/blob/8f9e4734020e0029a9ec53e1d38163bbcc6235c7/updater/lib/dependabot/updater.rb#L134-L146

This will nearly always result in `[] != [$at_least_one_dep]`, so we exit closing the existing PR.

This is feels an awful lot like cleanly closing the job accident/on a weak signal. We do always place a parser error on the UpdateJob, so the user should at least see the job failed due to unparsable files, but I think it also means we are likely closing "rebased" PRs on parse errors and that might be a bug.

PR #6866, opened 2023-03-17
[Updater] Establish the Updater::Operations pattern, Extract UpdateAllVersions as the first strategy class

brrygrdn:
This change is downstream of https://github.com/dependabot/dependabot-core/pull/6865 and https://github.com/dependabot/dependabot-core/pull/6847 which should both merge first

In order to introduce Grouped Updates, we need to clean up the distinct business objects involved in a single Dependabot "Operation" in order to avoid hard-to-reason about branching at different points in the code line.

This PR introduces the `Dependabot::Operations` module which acts as a way to isolate each distinct type of Update we do to avoid us switching from a Version- to Security-update strategy within methods.

It then uses this pattern to extract the logic required to do _just_ a "fresh" Version Update as a distinct operation class, failing over to the "legacy" implementation for all other jobs to help us de-risk the change.

The guts of this change are best looked reviewed via these specific commits:
- 018fad2d6f1b7e0fa14a7b7b3336614ec4864421 This copy-pastes **all** of the methods from `Dependabot::Updater` that are required to pass the test suite with no modifications
- 3f6ac3ad443b85dbb5d70db61e91b28292719494 In this commit, the `dependencies` method which parses and filters the job's dependency files has all logic related to Security updates and rebasing existing PRs removed
- 3f6ac3ad443b85dbb5d70db61e91b28292719494 Removes all remaining Security-update specific code

I ran the full test suite at each step to maintain confidence I wasn't changing behaviour.

### Notes

This PR is not as DRY as it could be, there are likely some more copy-pasted methods that could be extracted to classes that the `Dependabot::Updater` can pass into the `Operation::UpdateAllVersions` class.

I decided not to go too far with the de-duplication until we've:
- De-risked shipping the coarse cut-over to verify there's no surprises from removing the Security- and rebase-specific logic
- Extracted at least one more Operation to give us a better sense of what is/isn't common between them in order to avoid adding indirection on anything we don't actually use in more than one case

### DIFF
--- a/updater/lib/dependabot/updater/operations.rb
+++ b/updater/lib/dependabot/updater/operations.rb
@@ -30,10 +30,10 @@ module Dependabot
         UpdateAllVersions
       ]
 
-      def self.operation_for(job:)
+      def self.class_for(job:)
         raise ArgumentError, "Expected Dependabot::Job, got #{job.class}" unless job.is_a?(Dependabot::Job)
 
-        OPERATIONS.first { |op| op.applies_to?(job: job) }
+        OPERATIONS.find { |op| op.applies_to?(job: job) }
       end
     end
   end
--- a/updater/lib/dependabot/updater/operations.rb
+++ b/updater/lib/dependabot/updater/operations.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "dependabot/updater/operations/update_all_versions"
+
 # This module is responsible for determining which Operation a Job is requesting
 # the Updater to perform.
 #
@@ -21,8 +23,17 @@
 module Dependabot
   class Updater
     module Operations
-      def self.operation_for(_job:)
-        nil
+      # We check if each operation ::applies_to? a given job, returning the first
+      # that does, so these Operations should be ordered so that those with most
+      # specific preconditions go before those with more permissive checks.
+      OPERATIONS = [
+        UpdateAllVersions
+      ]
+
+      def self.operation_for(job:)
+        raise ArgumentError, "Expected Dependabot::Job, got #{job.class}" unless job.is_a?(Dependabot::Job)
+
+        OPERATIONS.first { |op| op.applies_to?(job: job) }
       end
     end
   end
--- a/updater/lib/dependabot/updater/operations.rb
+++ b/updater/lib/dependabot/updater/operations.rb
@@ -11,9 +11,6 @@ require "dependabot/updater/operations/update_all_versions"
 #
 # Consider the following guidelines when working on Operation classes:
 #
-# - Operations act on the principal of least knowledge prefering to cherry pick
-#   arguments from the Dependabot::Job so their contract is explicit.
-#
 # - Operations *should not have optional parameters*, prefer to create a new
 #   class instead of adding branching to an existing one.
 #
--- a/updater/lib/dependabot/updater/operations.rb
+++ b/updater/lib/dependabot/updater/operations.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+# This module is responsible for determining which Operation a Job is requesting
+# the Updater to perform.
+#
+# The design goal for this module is to make these classes easy to understand,
+# maintain and extend so we can eventually support community-contributed
+# alternatives or ecosystem-specific implementations.
+#
+# Consider the following guidelines when working on Operation classes:
+#
+# - Operations act on the principal of least knowledge prefering to cherry pick
+#   arguments from the Dependabot::Job so their contract is explicit.
+#
+# - Operations *should not have optional parameters*, prefer to create a new
+#   class instead of adding branching to an existing one.
+#
+# - Operations should prefer to share logic by composition, there is no base
+#   class. We want to avoid implicit or indirect behaviour as much as possible.
+#
+module Dependabot
+  class Updater
+    module Operations
+      def self.operation_for(_job:)
+        nil
+      end
+    end
+  end
+end
--- a/updater/lib/dependabot/updater/operations/update_all_versions.rb
+++ b/updater/lib/dependabot/updater/operations/update_all_versions.rb
@@ -23,6 +23,7 @@ module Dependabot
 
         def perform
           Dependabot.logger.info("Starting update job for #{job.source.repo}")
+          Dependabot.logger.info("Checking all dependencies for version updates...")
           dependencies.each { |dep| check_and_create_pr_with_error_handling(dep) }
         end
 
--- a/updater/lib/dependabot/updater/operations/update_all_versions.rb
+++ b/updater/lib/dependabot/updater/operations/update_all_versions.rb
@@ -34,7 +34,6 @@ module Dependabot
                     :error_handler,
                     :created_pull_requests
 
-        # rubocop:disable Metrics/PerceivedComplexity
         def dependencies
           all_deps = dependency_snapshot.dependencies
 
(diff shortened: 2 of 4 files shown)


updater/lib/dependabot/updater/operations.rb
```diff
@@ -0,0 +1,40 @@
...
+#
+# The design goal for this module is to make these classes easy to understand,
+# maintain and extend so we can eventually support community-contributed
+# alternatives or ecosystem-specific implementations.
+#
+# Consider the following guidelines when working on Operation classes:
+#
+# - Operations act on the principal of least knowledge prefering to cherry pick
+#   arguments from the Dependabot::Job so their contract is explicit.
+#
```
brrygrdn, line 16: I staked out these rules before I started writing any code and I wasn't able to keep to this one.

I'm very concerned at the `Dependabot::Job` becoming too broad an AI, it already provides vulnerability checking, etc, so I think there is work to be done in narrowing it into a few smaller interfaces.

I think solving this is fairly large and off the critical path, so I'm leaning towards backlogging this until we get a bit further with other priorities.

landongrindheim (approved):
I hard started reviewing this at the end of the day on Friday. Since you returned this to a draft state, I'm submitting and will return to review when we're ready to see this ship 😄

updater/lib/dependabot/updater/operations.rb
```diff
@@ -0,0 +1,40 @@
...
+# - Operations *should not have optional parameters*, prefer to create a new
+#   class instead of adding branching to an existing one.
+#
+# - Operations should prefer to share logic by composition, there is no base
+#   class. We want to avoid implicit or indirect behaviour as much as possible.
```
landongrindheim, line 21: 😄 I'm appreciating these design principles. A comment seems right. I wonder if, _in the future_, we could enforce this with linting so that someone would have to be very intentional about going against these principles.

updater/lib/dependabot/updater/operations/update_all_versions.rb
```diff
@@ -0,0 +1,357 @@
...
+          @dependency_snapshot = dependency_snapshot
+          @error_handler = error_handler
+          # TODO: Collect @created_pull_requests on the Job object?
+          @created_pull_requests = []
+        end
+
+        def perform
+          Dependabot.logger.info("Starting update job for #{job.source.repo}")
+          dependencies.each { |dep| check_and_create_pr_with_error_handling(dep) }
+        end
```
landongrindheim, line 27: I'd appreciate it if someone more deeply familiar with this repo, like @jakecoffman, could verify this workflow 🙏

updater/spec/dependabot/updater/operations_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      # We always expect jobs that update a pull request to specify their
+      # existing dependency changes, a job with this set of conditions
+      # should never exist.
+      job = instance_double(Dependabot::Job,
+                            security_updates_only?: false,
+                            updating_a_pull_request?: true,
+                            dependencies: [],
+                            is_a?: true)
+
+      expect(described_class.class_for(job: job)).to be_nil
```
landongrindheim, line 20: 🤔 Would it make sense to have a default for this?

I could see an argument being made that a version or (maybe preferably?) security update of an individual dependency would be the default behavior. We already raise an error if it's not a `Job`, so it makes sense to me to make sure we have an `Operation` class for all jobs. I could also see an argument being made for raising an error for the case you described above:

> a job with this set of conditions should never exist.

If we can't act on it, maybe it's best to bail early.

landongrindheim (approved):

updater/lib/dependabot/updater/operations/update_all_versions.rb
brrygrdn: FWIW, this code is a copy-paste from the existing updater with some specific clauses removed - I think it's probably worth reviewing-by-commit: 

> The guts of this change are best looked reviewed via these specific commits:
> - 018fad2d6f1b7e0fa14a7b7b3336614ec4864421 This copy-pastes **all** of the methods from `Dependabot::Updater` that are required to pass the test suite with no modifications
> - 3f6ac3ad443b85dbb5d70db61e91b28292719494 In this commit, the `dependencies` method which parses and filters the job's dependency files has all logic related to Security updates and rebasing existing PRs removed
> - 3f6ac3ad443b85dbb5d70db61e91b28292719494 Removes all remaining Security-update specific code

updater/lib/dependabot/updater/operations/update_all_versions.rb
```diff
@@ -0,0 +1,357 @@
...
+module Dependabot
+  class Updater
+    module Operations
+      class UpdateAllVersions
+        def self.applies_to?(job:)
+          return false if job.security_updates_only?
+          return false if job.updating_a_pull_request?
+          return false if job.dependencies&.any?
+
+          true
```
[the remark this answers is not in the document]
brrygrdn: I did think about using an experiment to roll this out, but since it is just extracting code and clipping off branches of if statements that are essentially `false` when the `applies_to?` method is true, I'm of fairly high confidence this is a refactor with no expected behaviour change.

I think smoke testing + staging is sufficient to roll into production and watch the first few jobs run vs a gradual rollout via feature flag - the gradual rollout might actually give us false confidence as any problems may not be statistically significant in a sample of 10% or 25% of jobs.

updater/lib/dependabot/updater/operations/update_all_versions.rb
```diff
@@ -0,0 +1,357 @@
+# frozen_string_literal: true
+
+module Dependabot
+  class Updater
+    module Operations
+      class UpdateAllVersions
+        def self.applies_to?(job:)
+          return false if job.security_updates_only?
+          return false if job.updating_a_pull_request?
+          return false if job.dependencies&.any?
```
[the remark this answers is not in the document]
brrygrdn: Yeah, for a fresh Version Update, we never pre-fill the `dependencies` array, it is only prepopulated for Security-Updates or rebases.

If we were to do "fresh", targeted Version Updates with a pre-filled set of `dependencies` with the existing code I think they might "work by accident" - we've had some experiments around 'click to update this one Dependency' in the past that benefited from this.

I think the spirit of this Operation refactor is that we should not have a strategy for a scenario we currently do not explicitly support, so essentially if we wanted to do "fresh", targeted Version Updates in future, we would define a new class with:

```
def self.applies_to?(job:)
  return false if job.security_updates_only?
  return false if job.dependencies.empty?
end
```

It would then implement a version of this logic that _explicitly_ resolved the job.dependencies with the parsed Dependencies.

updater/lib/dependabot/updater/operations/update_all_versions.rb
jakecoffman: Looks good!

I haven't been watching the changes so forgive my lack of context in my questions, maybe you're already headed in this direction:

Would it make sense to make the Operation per-dependency? In `check_and_create_pr_with_error_handling` it calls `if pr_exists_for_latest_version?` and `if (existing_pr = ...`. In the job definition passed in, we can identify that a PR exists beforehand and so it should be straight forward to separate a `CreatePR` and `UpdatePR` Operations.

brrygrdn: > Would it make sense to make the Operation per-dependency?

This is a good question, the "fresh" Version Update is a funny special case as most of the other job types tend to focus on a single ( or handful ) of Dependencies.

I've been thinking more of Operations being a job-level abstraction for now, which might be a bit of a misnomer, but I definitely do think as we start to work in grouping we should start to split the create/update pr responsibilities out - I'm just not sure how to align that breakdown right now.

PR #6879, opened 2023-03-21
allow Ruby 3.2.0 updates to work

jakecoffman:
While we're [deliberating on how to proceed with Ruby 3.2.0](https://github.com/dependabot/dependabot-core/pull/6436), we're seeing some updates failing due to requirements on it. So to fix that we can add that version here to get PRs flowing again.

### DIFF
--- a/bundler/helpers/v2/monkey_patches/definition_ruby_version_patch.rb
+++ b/bundler/helpers/v2/monkey_patches/definition_ruby_version_patch.rb
@@ -10,7 +10,7 @@ module BundlerDefinitionRubyVersionPatch
         Gem::Specification.new("Ruby\0", requested_version)
     end
 
-    %w(2.5.3 2.6.10 2.7.7 3.0.5 3.2.0 3.2.1).each do |version|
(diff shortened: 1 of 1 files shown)


bundler/helpers/v2/monkey_patches/definition_ruby_version_patch.rb
```diff
@@ -10,7 +10,7 @@ def source_requirements
         Gem::Specification.new("Ruby\0", requested_version)
     end
 
-    %w(2.5.3 2.6.10 2.7.7 3.0.5 3.2.0).each do |version|
+    %w(2.5.3 2.6.10 2.7.7 3.0.5 3.2.0 3.2.1).each do |version|
```
deivid-rodriguez, line 13: We normally keeping just the latest patch version of each minor release, so I'd follow that pattern.

An exhaustive list would be more correct but nobody pins to exact rubies through library dependencies, and applications pinning to exact rubies are already handled by the lines above this one.

jakecoffman: Thanks 🙂

PR #6883, opened 2023-03-21
fix Yarn 1 erroring on parent npmrc files with undefined vars

jakecoffman:
If a repo has multiple .npmrc files, Dependabot only rewrites one of them to remove variables. NPM doesn't seem to have an issue with this, I assume it's just expanding undefined vars to `""`, but Yarn 1 is not ok with that and errors.

To fix this, I've added some code to rewrite all of the .npmrc files and remove the variables. 

Still trying to figure out if there's a way to test this in this project, I have a CLI test that is passing with this change.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
@@ -367,6 +367,10 @@ module Dependabot
             File.write(file.name, updated_content)
           end
 
+          clean_npmrc_in_path(yarn_lock)
+        end
+
+        def clean_npmrc_in_path(yarn_lock)
           # Berry does not read npmrc files.
           return if Helpers.yarn_berry?(yarn_lock)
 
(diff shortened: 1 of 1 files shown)


Nishnha:
Npm handles this a bit differently - instead of removing any variables, it only includes in the lines it needs into an .npmrc
https://github.com/dependabot/dependabot-core/blob/9404097d8b244e55183c744d3982df29c7922173/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npmrc_builder.rb#L25-L36

and then the content is written out into a temporary directory when performing updates
https://github.com/dependabot/dependabot-core/blob/9404097d8b244e55183c744d3982df29c7922173/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb#L482

And it looks like Yarn v1 does the same thing
https://github.com/dependabot/dependabot-core/blob/9404097d8b244e55183c744d3982df29c7922173/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb#L566-L571
https://github.com/dependabot/dependabot-core/blob/9404097d8b244e55183c744d3982df29c7922173/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb#L346

You may need to update the way we add these these lockfiles to the temporary directory so that all of the .yarnrc files are added

jakecoffman:
@Nishnha Yarn is using `in_a_temporary_repo_directory` which is cloned to, so there's no need to write more .npmrc files, they're already there! Which is actually why this problem exists. We need to clean those up too.

jakecoffman:
I created a Yarn 1 smoke test which also tests this issue since it required a clone to have certain files on disk, and I wasn't sure how I'd achieve that in the rspec tests: https://github.com/dependabot/smoke-tests/pull/45

pavera (approved):

PR #6884, opened 2023-03-21
[Updater] Incorporate our first pass at grouped updates as an experiment we can ship

brrygrdn:
This branch supersedes our original spike in https://github.com/dependabot/dependabot-core/pull/6663

Having established and shipped the "operation" pattern as a way to separate out the code paths for specific types of `Dependabot::Job` objects to avoid a lot of low-level branching, we can now package our experimental spike into a code structure we can safely ship without risking an impact on production jobs.

This PR introduces the `Dependabot:: Updater::Operations::GroupUpdateAllVersions` class which applies to all fresh, Version Updates **which have a `Dependabot::Experiment flag enabled**.

This code is currently very lightly tested to confirm that it behaves identically to the existing adapter for a simple unit test case. I plan to add some smoke tests to demonstrate the correct grouping behaviour as a guard against regressions on this basic walking skeleton, but there is a job to be done in fleshing out the tests properly as a follow up.


updater/lib/dependabot/updater/operations/group_update_all_versions.rb
```diff
@@ -0,0 +1,350 @@
...
+          )
+          [] # return an empty set
+        rescue StandardError => e
+          error_handler.handle_dependabot_error(error: e, dependency: dependency)
+          [] # return an empty set
+        end
+
+        def filter_unrelated_and_unchanged(updated_dependencies, checker)
+          updated_dependencies.reject do |d|
+            next false if d.name == checker.dependency.name
```
abdulapopoola, line 173: any concerns with casing?

brrygrdn: I think it's ok at this point as we are using `Dependabot::Dependency` objects and the name should be normalised, plus this is currently replicating [existing behaviour](https://github.com/dependabot/dependabot-core/pull/6884/files#diff-be79fd5c595a9f52f3ab443d0487b65c13d3b2b65d5b8793fb85ff91601862b5R133-R136) that hasn't been DRY'd out yet.

Case checking is something we have a few inconsistencies on though, so it's a good spot.

updater/lib/dependabot/updater/operations/group_update_all_versions.rb
```diff
@@ -0,0 +1,370 @@
...
+          all_deps = dependency_snapshot.dependencies
+
+          # Tell the backend about the current dependencies on the target branch
+          service.update_dependency_list(dependency_snapshot: dependency_snapshot)
+
+          allowed_deps = all_deps.select { |d| job.allowed_update?(d) }
+          # Return dependencies in a random order, with top-level dependencies
+          # considered first so that dependency runs which time out don't always hit
+          # the same dependencies
+          allowed_deps = allowed_deps.shuffle unless ENV["UPDATER_DETERMINISTIC"]
```
brrygrdn, line 80: I think we should probably remove shuffling _within_ groups and instead shuffle the groups when we have more than one, but for now I think shuffling has the benefit that it might flush out issues compiling changes together if we run them in a random order - if we made this deterministic now it could result in false confidence.

jakecoffman (approved):
LGTM

PR #5436, opened 2022-07-27
Bump docker_registry2 to fix #3989

noorul:
The latest version of docker_registry2 has the fix for #3989.

PR #6885, opened 2023-03-21
Bump `docker_registry2` to `1.13`

jeffwidman:
Bump `docker_registry2` to `1.13` to pickup the fix in https://github.com/deitch/docker_registry2/pull/68.

This pin isn't strictly necessary since the upstream fix is mostly cosmetic making a stack trace more legible, so I originally wasn't going to force it.

However, I removed the `>=` pin as it's redundant given the `~>` pin. It crept in originally as part of forcing a newer patch release in https://github.com/dependabot/dependabot-core/commit/e991abb4aee96bbd9d114a0e899f03f796eb4424, but should have been removed by https://github.com/dependabot/dependabot-core/commit/0299755916386b23df8fc8de58a05052a1c827a3 / https://github.com/dependabot/dependabot-core/pull/5436.

And since I was doing that, I figured I might as well bump it too.


jeffwidman:
ah yeah, good ol' `cd updater && bundle install` probably... I have to do that routinely on the Dependabot PR's...

jeffwidman:
I think this is good to go for now.

I'm not worried about bumping the license and would rather ignore it for now--we can pick that up later, when we add a CI check enforcing license update compliance, otherwise it'll just fall out of date again...

Nishnha (approved):

PR #6864, opened 2023-03-17
Link docker image to repo via OCI source annotation

mctofu:
Allows us to include richer metadata about the release when creating Docker update PRs (commits, release notes, repo link, etc) to address https://github.com/dependabot/dependabot-core/issues/2085. To enable support, maintainers of docker images need to add the [`org.opencontainers.image.source`](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys) label to their `Dockerfile` with the url of the source repo. Additionally, they need to tag the repo with the same tags as the published docker images. This allows Dependabot to understand which repo & commit is associated each version/tag of a docker image.

This example repo demonstrates the setup: https://github.com/dependabot-fixtures/docker-with-source

There isn't a simple registry API to read the source annotation. Instead you need to locate and download the correct blob which contains it. We could implement this ourselves but the [regctl](https://github.com/regclient/regclient) tool made this pretty easy so I've used that instead. It might be an option to replace our native registry calls in the future?

### notes
I did not implement authentication support for regctl. On the GitHub platform or using the CLI the authentication will be handled by the proxy so it isn't needed. This could be added later by anyone that needs this feature and runs dependabot outside of those methods.

It is also possible for the OCI annotations to indicate the commit associated with a docker image. If we used this, maintainers wouldn't need to add equivalent tags to their repo. I wanted to support this initially but our metadata finders don't seem to provide a paved path for this (unless I've missed something?). It would take some additional work to wire that up which we could evaluate in the future.

PR #6887, opened 2023-03-21
Add OCI source annotation to updater core image

mctofu:
This will help Dependabot include relevant change details when it opens PRs against this image. Support was added in https://github.com/dependabot/dependabot-core/pull/6864.

I'm not sure if we also need to add this label to each ecosystem image. I'll followup with that change if this doesn't have the expected result.

### DIFF
--- a/Dockerfile.updater-core
+++ b/Dockerfile.updater-core
@@ -101,6 +101,8 @@ rm -rf ~/.bundle
 
 FROM ubuntu:20.04
 
+LABEL org.opencontainers.image.source="https://github.com/dependabot/dependabot-core"
+
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 ENV DEBIAN_FRONTEND="noninteractive" \


Nishnha (approved):
❤️

mctofu:
Confirmed this label is present in the ecosystem images:
```
% regctl image inspect ghcr.io/dependabot/dependabot-updater-bundler:9b526b7214f5b03a6d9378efb03b1cc55d258762 | jq .config
{
  "User": "dependabot",
  "Env": [
    "PATH=/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
    "DEBIAN_FRONTEND=noninteractive",
    "LC_ALL=en_US.UTF-8",
    "LANG=en_US.UTF-8",
    "GIT_LFS_SKIP_SMUDGE=1",
    "DEPENDABOT_HOME=/home/dependabot",
    "DEPENDABOT_NATIVE_HELPERS_PATH=/opt"
  ],
  "Cmd": [
    "bin/run"
  ],
  "WorkingDir": "/home/dependabot/dependabot-updater",
  "Labels": {
    "org.opencontainers.image.ref.name": "ubuntu",
    "org.opencontainers.image.source": "https://github.com/dependabot/dependabot-core",
    "org.opencontainers.image.version": "20.04"
  },
  "ArgsEscaped": true
}
```

Labels are also inherited from our base ubuntu image. That means if maintainers don't set the source annotation in their own image Dependabot can pick up the change logs from a base image instead which may not be desired behavior.

PR #6904, opened 2023-03-24
Cleanup Bundler lockfile updater

deivid-rodriguez:
Just checking what CI thinks about removing this code.

### DIFF
--- a/bundler/helpers/v2/lib/functions/lockfile_updater.rb
+++ b/bundler/helpers/v2/lib/functions/lockfile_updater.rb
@@ -171,12 +171,7 @@ module Functions
       end
 
       # If there are specific dependencies we can unlock, unlock them
-      return dependencies_to_unlock.append(*potential_deps) if potential_deps.any?
-
-      # Fall back to unlocking *all* sub-dependencies. This is required
-      # because Bundler's SolveFailure objects don't include enough
-      # information to chart the full path through all conflicts unwound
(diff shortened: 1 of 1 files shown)


jeffwidman (approved):
CI seems happy and we're now on Bundler 2.4, so I think this is ready to go?

deivid-rodriguez:
Even if CI didn't fail, I was not able to fully convince myself this code is not actually needed anymore, so I'd rather close this. We can open a tech-debt issue to figure this out and either get the code covered, or deleted? Or add a TODO?

jeffwidman:
Makes sense.

We can probably get by without a new issue since GitHub project boards allow tracking PR's, so we can treat this PR itself as the open issue...

v-kbukum1:
@jeffwidman , @deivid-rodriguez do you think we should add an issue for this PR in our backlog or close it. I think it will remain here like that if we don't create issue and add it in our backlog?

jeffwidman:
No idea.

deivid-rodriguez:
No idea either. You could try shipping this gradually under a feature flag, and if it does not cause issues, consolidate the removal. Or feel free to close if you don't think it's worth it!

v-kbukum1:
I am closing it for now. As I understand we don't have related issue so it is hard to predict if that is going to cause an issue as well. So closing it may be better for now.

PR #6906, opened 2023-03-24
[Updater] Perform Dependency file parsing early, and fail fast if there's a problem

brrygrdn:
### Background

In the Updater, if file parsing fails, we get an empty value for "dependencies" at the code [here](https://github.com/dependabot/dependabot-core/blob/main/updater/lib/dependabot/updater.rb#L621-L624).

```ruby
    rescue StandardError => e
      error_handler.handle_parser_error(e)
      []
    end
```

We lazily evaluate `dependencies` in the `Dependabot::Updater#run` method which results in us quietly exiting the job in two difference places based [on this if statement](https://github.com/dependabot/dependabot-core/blob/main/updater/lib/dependabot/updater.rb#L59-L77):

```ruby
    def run
      return unless job


      if job.updating_a_pull_request?
        Dependabot.logger.info("Starting PR update job for #{job.source.repo}")
        check_and_update_existing_pr_with_error_handling(dependencies)
      else
        Dependabot.logger.info("Starting update job for #{job.source.repo}")
        dependencies.each { |dep| check_and_create_pr_with_error_handling(dep) }
      end
    rescue *ErrorHandler::RUN_HALTING_ERRORS.keys => e
      if e.is_a?(Dependabot::AllVersionsIgnored) && !job.security_updates_only?
        error = StandardError.new(
          "Dependabot::AllVersionsIgnored was unexpectedly raised for a non-security update job"
        )
        error.set_backtrace(e.backtrace)
        service.capture_exception(error: error, job: job)
        return
      end
```

For a new update, `[]` means we don't iterate and the job silently ends on the line:

```ruby
dependencies.each { |dep| check_and_create_pr_with_error_handling(dep) }
```

For updating an existing PR, we end up [here](https://github.com/dependabot/dependabot-core/blob/e5bb0997b216cd7bf70b12f921318442804a3036/updater/lib/dependabot/updater.rb#L121-L133):

```ruby
      if dependencies.count != job.dependencies.count
        # TODO: Remove the unless statement
        #
        # This check is to determine if there was an error parsing the dependency
        # dependency file.
        #
        # For update existing PR operations we should early exit after a failed
        # parse instead, but we currently share the `#dependencies` method
        # with other code paths. This will be fixed as we break out Operation
        # classes.
        close_pull_request(reason: :dependency_removed) unless service.errors.any?
        return
      end
```

This will nearly always result in `[] != [$at_least_one_dep]`, so we exit. We don't close the PR as the parser failure has placed an error on the job, but we could have the same outcome by stopping if the parse fails.

### The Change

This branch yoinks all file parsing concerns to the very start of the `UpdateFileCommand` so we retrieve the files from the job definition, decode them and parse them before starting into any actual update logic.

This means that all exception handling around parsing failures can happen immediately and clearly around the instantiation of the `Dependabot::DependencySnapshot`, removing a significant concern from the `Dependabot::Updater` class and clarifying that we only expect __update time__ errors to be handled inside it, not parse errors.

### Rationale

The main reason to do this now is we still need to split out another 3 "Operation" classes in the style of https://github.com/dependabot/dependabot-core/pull/6866.

This work will break the Updater down further, allowing us to pull apart its responsibilities and break the test cases across some concrete classes. In its current state, file parsing would end up becoming a concern of every responsibility, but a code walk tells us that we can short-circuit this and turn it into a precondition.

This will significantly reduce the remaining work by allowing the rest of the class breakdown to be focused on just generating updates and posting PRs.


updater/lib/dependabot/dependency_snapshot.rb
```diff
@@ -11,58 +11,34 @@
...
     def initialize(job:, base_commit_sha:, dependency_files:)
       @job = job
       @base_commit_sha = base_commit_sha
       @dependency_files = dependency_files
-    end
-
-    def dependencies
-      return @dependencies if defined?(@dependencies)
 
       parse_files!
```
landongrindheim, line 37: 💅 I think it makes sense to explicitly set `@dependencies` here.

updater/lib/dependabot/updater.rb
```diff
@@ -133,16 +133,10 @@ def check_and_update_existing_pr_with_error_handling(dependencies)
...
-        # dependency file.
-        #
-        # For update existing PR operations we should early exit after a failed
-        # parse instead, but we currently share the `#dependencies` method
-        # with other code paths. This will be fixed as we break out Operation
-        # classes.
-        close_pull_request(reason: :dependency_removed) unless service.errors.any?
+        # If the job dependencies mismatch the parsed dependencies, then
+        # we should close the PR as at least one thing we changed has been
+        # removed from the project.
```
landongrindheim, line 138: Thanks for this context 😄

updater/lib/dependabot/updater.rb
```diff
@@ -591,8 +585,6 @@ def existing_pull_request(updated_dependencies)
         created_pull_requests.find { |pr| Set.new(pr) == new_pr_set }
     end
 
-    # rubocop:disable Metrics/AbcSize
-    # rubocop:disable Metrics/PerceivedComplexity
```
landongrindheim, line 595: 🤩

updater/spec/dependabot/dependency_snapshot_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+  end
+
+  let(:job) do
+    instance_double(Dependabot::Job,
+                    package_manager: "bundler",
+                    repo_contents_path: nil,
+                    credentials: [],
+                    reject_external_code?: false,
+                    source: source,
+                    experiments: { large_hadron_collider: true })
```
landongrindheim, line 26: 😆

landongrindheim (approved):

updater/lib/dependabot/dependency_snapshot.rb
landongrindheim: The diff doesn't really allow for this, but FWIW, I was referring to setting `@dependencies` in the initializer (and removing the setting of it from `#parse_files!`

updater/lib/dependabot/update_files_command.rb
```diff
@@ -10,20 +10,19 @@ class UpdateFilesCommand < BaseCommand
...
-      )
+      # them, decode and parse them into an object that knows the current state
+      # of the project's dependencies.
+      begin
+        dependency_snapshot = Dependabot::DependencySnapshot.create_from_job_definition(
+          job: job,
+          job_definition: Environment.job_definition
+        )
+      rescue StandardError => e
+        handle_parser_error(e)
```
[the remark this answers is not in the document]
brrygrdn: I'm split on this as I originally created that class to contain these two long methods, but because it is namespaced within the Updater it felt right to remove the concern from it as being out of bounds.

I'm still a bit uncertain of the long term place for that helper, ideally we'd push the need to do the dance to get from `DependabotError` class to `error-details` hash to be a convention where every error implemented the "presentation" hash directly but that's a big lift and pretty outside the critical path.

PR #6910, opened 2023-03-24
Minor: Allow to change ruby used to run native helpers #6517

ylecuyer:
Fix for #6517

I have added a `ruby_cmd` option to the `FileParser` class so that updating the gem is done with the context of jruby for example (thus installing proper native extensions)

It can be used like this:

```ruby
@parser = Dependabot::FileParsers.for_package_manager('bundler').new(dependency_files: @files, source: @source, credentials: @credentials, options: { ruby_cmd: '/home/ylecuyer/.rbenv/versions/jruby-9.3.8.0/bin/jruby' })
```


Nishnha:
This implementation makes sense to me. Are you self-hosting Dependabot with JRuby installed? 
GitHub hosted Dependabot doesn't ship with JRuby, so I would expect [the custom `ruby_cmd`](https://github.com/dependabot/dependabot-core/pull/6910/files#diff-0b1c0887c8f0e4e5821508313c257f84aa81a78b9ba9c47d07cff0bdc09d71e3R27-R40) to fail

ylecuyer:
I do have dependabot self-hosted indeed. It isn't running with jruby but instead with ruby.

Here I need custom jruby because for jruby project the diff created by dependabot brings dependencies with the ruby platform and not Java platform. In https://github.com/dependabot/dependabot-core/issues/6517#issuecomment-1483581368 I give an example diff of the Gemfile.lock with this change.

> GitHub hosted Dependabot doesn't ship with JRuby, so I would expect [the custom ruby_cmd](https://github.com/dependabot/dependabot-core/pull/6910/files#diff-0b1c0887c8f0e4e5821508313c257f84aa81a78b9ba9c47d07cff0bdc09d71e3R27-R40) to fail

The default isn't changed and will keep using ruby unless you explicitly override it, so in the end the change should still be compatible with github hosting.

Nishnha (dismissed):
This LGTM. 

Note to maintainers: we should look into fixing the situations where we replace platform-specific gems with the default ruby ones. We can leverage the `--platform` option when installing gems

jeffwidman (commented):
I like the idea, but I wonder if this is the right API?

We have the same problem for _all_ native helpers... but typically self-hosted folks don't run into this because they're already running :dependabot: in the context of a runtime that already has the support they need.

I don't know a lot about `jruby`, but I'm curious why you don't run Dependabot entirely using Jruby? That'd solve this for you.

Again though, I _do_ think we will need an API for passing in the desired runtime for _all_ ecosystems that use a native helper... ie, be able to run `go` or `php` with custom flags or a custom compiled version that you're only using for Dependabot.

We could ship this as a short-term thing, or we could figure out that API a little more. I know we're planning to do more investigation/thought on ecosystem version/runtime handling later this quarter, and this may fall into that bucket?

ylecuyer:
> I don't know a lot about jruby, but I'm curious why you don't run Dependabot entirely using Jruby? That'd solve this for you.

We have multiple projects some running ruby and some jruby, our dependabot setup runs from one place only and is configured to scan all projects and create associated PRs.

Contrary to other dependabot users, we do not update all dependencies when an update is available. We only update dependencies having vulnerabilities. Thus when we receive a CVE alert (like nokogiri in the example which led to this PR) we run the dependabot pipeline which will patch all projects. That's why we do not have some automated process running from every projects.

> We could ship this as a short-term thing

That would be nice, I'm already using my fork with the patch with good results so far.

PR #6914, opened 2023-03-26
pip-tools: Honour `--resolver` option for existing requirements. 

gmcrocetti:
## Problem it solves

Dependabot looks into existing `requirements.txt` files to grasp the original flags passed to `pip-compile`. `--resolver` is not considered in current version. In order to replicate flags options, we gotta add it :)

## Description
Detects and add this header in case the requirements file was generated with it.
Create new fixtures files representing requirements generated with both '--resolver=legacy' and '--resolver=backtracking'

I'm not a ruby expert, in fact, this is my first time touching it. Please let me know if anything can be written in a more "ruby-ish" way.

BTW, I couldn't find any issue related


deivid-rodriguez:
Thanks for working on this @gmcrocetti! ❤️ 

The patch looks good but unfortunately the test does not cover the fix. This is because we overwrite the header in the generated file to be the original one here: https://github.com/dependabot/dependabot-core/blob/bccb2b23450e3fd322d1af366f30451a476de37e/python/lib/dependabot/python/file_updater/pip_compile_file_updater.rb#L280. So those specs are going to pass (and already pass without your changes), no matter which flags we pass to `pip-compile` under the hood.

To figure out a spec that we can write to cover the fix, I wonder what's the original problem that prompted you to create this PR? Maybe you actually need `--resolver=backtracking` because otherwise Dependabot is unable to create a PR, and we're still using `--resolver=legacy` under the hood since that's still the default?

deivid-rodriguez:
> I may have misunderstood but this means the "final" header is misleading in the sense dependabot may have ignored some of these flags ? I mean... it will copy --resolver=backtracking from the original file but doesn't mean the command was called with it ? Interesting 🤔

Yep. Ideally we'll end up running the same command recorded in the header, but there could be cases (like this bug) where we don't. Either way don't want to update the header in our PRs regardless, because users never expect Dependabot updating anything other than version numbers.

deivid-rodriguez:
> Dependabot is able to run but the dependencies are not what we expect :/. I indeed need this flag in order to generate the proper set of dependencies.

Can you share a public repository where this issue can be reproduced? Ideally we'd be able to record the situation in a test case, and instead of asserting that the correct header is written, assert that the correct dependencies are resolved.

gmcrocetti:
Okay !
Fair enough. It's indeed the expected user experience. Thanks for explaining :)

I don't have one right now (it's privte) but will work towards that :). I'll let you know

deivid-rodriguez (approved):
👏 👏

deivid-rodriguez:
Thanks so much @gmcrocetti, this is live now!

PR #6921, opened 2023-03-27
fix elm tests

jakecoffman:
Elm tests started failing due to a new release of https://github.com/mercurymedia/elm-datetime-picker. 

We should probably write the test so this doesn't happen, but for now I will just update the fixture.


deivid-rodriguez (approved):
Thank you!

Issue #6308, opened 2022-12-13
Support generic interface for new source control providers


jjmanton:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

I have a need currently to integrate Dependabot with an on-premise bitbucket server. I was able to get this working by modifying the code, but I feel there is an opportunity here to allow for a pluggable model.

I found this comment for context - https://github.com/dependabot/dependabot-core/pull/1450#issuecomment-821039074

The above seems to indicate support for a generic interface. I would be happy to help build and maintain the Bitbucket Server gem, however I would require some assistance on the dependabot side to ensure it can be consumed and conforms to an interface that works for current and future needs.

With some initial investigation, it seems the gem would need to implement roughly three files - 
- Client (Dependabot::Client::CLIENT_NAME)
- Pull Request Creator (Dependabot::PullRequestCreator::CLIENT_NAME)
- Pull Request Updater (Dependabot::PullRequestUpdater::CLIENT_NAME)

After making the changes, we could envision having Bitbucket Cloud, Bitbucket Server, GitLab SaaS, Azure DevOps, and CodeCommit all converted to gems to be supported by the community.

jeffwidman:
👋 Personally, I have long wanted to do exactly what you suggest of having a more generic interface for clients, and then move the implementation-specific code outside of `dependabot-core`.

Basically tighten up the abstractions and also make it easier for community to maintain. We've had some internal discussions about different ways to do this (there's several paths forward), but mostly it's just been bottlenecked by not being high enough priority. Hopefully something we can tackle down the road here.

One follow-up question--Bitbucket Cloud vs Bitbucket Server... I'm surprised you can't just use the BitBucket cloud client, but pointed at a custom URL for your Bitbucket server instance... are there differences beyond a custom domain name that necessitate a full-blown separate client?

jjmanton:
@jeffwidman That's great to hear! If you have any notes that you can share, I'd be happy to start hacking on it a bit. I am not necessarily a Ruby developer, but can probably pick it up. If you have any preference on the overall structure or how the gem might plugin (and where) I can start putting some things together.

On the Bitbucket side, unfortunately the APIs are JUST different enough to warrant separate code. It's quite frustrating!

PR #6924, opened 2023-03-27
PROPOSAL: Allow Injection of Git Provider

tiedec:
This is a proposal for a mechanism to allow Git providers to be given by the caller. This allows dependabot-core to be more generic and extends its use beyond the handful of currently supported providers. 

With this proposal, the caller may optionally pass an "external provider" object that implements the details needed to support their Git client to dependabot-core. The Dependabot::Source object stores the external provider and invokes functionality from it if it was given. Fundamentally, this extends the case statements that act on the Dependabot::Source.provider attribute to include the external provider. Note that when no external provider is given, dependabot-core will work exactly as it does today. There is no other change in logic.

Notes:
- The external_files directory is not intended to be added to dependabot-core. This is merely a place where I put the example code that is the responsibility of other teams. (dependabot-script was the base for this.) It was included for the sole purpose of providing context and a usage example.
- This is not intended to be a complete solution but is a proof of concept of the approach. Based on the feedback on this PR, a complete solution can be implemented.
- This is a followup to https://github.com/dependabot/dependabot-core/issues/6308.


adc-cjewett:
Hi @tiedec. Sorry if this isn't related to the pull-request specifically but I noticed the files in here suggest the addition of Bitbucket Server support. Is the existing Bitbucket functionality only compatible with Bitbucket Cloud? We were looking to integrate with this soon and between the dependabot-script issues and difficulty building all of this locally I figured I'd just ask instead before going down the rabbit hole.

tiedec:
@adc-cjewett, yes, the de facto dependabot-core supports Bitbucket Cloud (bitbucket.org) but not the Enterprise Bitbucket Server, aka Bitbucket Data Center. My company uses Bitbucket Server so this tool did not meet our needs. The purpose of this PR is to all us to add support for Bitbucket Server via a plug in.  This separates the Git provider from the dependabot implementation.

Your wording is a little unclear to me. If Bitbucket Cloud is your Git provider, then there is no need for you to follow this PR. Just use the "bitbucket" provider and you should be good to go with the existing dependabot tool. On the other hand, if it is Bitbucket Server you are using then perhaps you can help me get this PR approved. :)

adc-cjewett:
@tiedec Thanks for the info! To clear up the confusion we're in a position where we use Bitbucket Data Center and not Bitbucket Cloud. There is a growing need for a tool like this to help improve our developer experience which is why I've been digging through this project a little more in the last week.

tiedec:
@adc-cjewett, thanks for the clarification; sounds like we are in the same boat.

While we are waiting for progress on this proposal, we have decided to move ahead by monkey patching the dependabot classes that are touched by this proposal. Basically, instead of modifying the gem itself we are using the released gem and altering specific details at runtime. We found the monkey patch approach to be easier to maintain than producing our own version of the gem.

PR #6930, opened 2023-03-28
decrease chance of Docker cache invalidation due to common change

jakecoffman:
I noticed we're copying all of common in before doing the `bundle install`. This means for any change in common we're having to wait for bundler to run again when it's not required unless the Gemfile or gemspec changes.

This should result in faster builds!

### DIFF
--- a/hex/spec/dependabot/hex/file_updater/lockfile_updater_spec.rb
+++ b/hex/spec/dependabot/hex/file_updater/lockfile_updater_spec.rb
@@ -372,6 +372,7 @@ RSpec.describe Dependabot::Hex::FileUpdater::LockfileUpdater do
       end
 
       it "updates the dependency version in the lockfile" do
+        pending "dependabot-private.fly.dev is offline"
         expect(updated_lockfile_content).to include %({:hex, :jason, "1.1.0")
         expect(updated_lockfile_content).not_to include(
           "0f7cfa9bdb23fed721ec05419bcee2b2c21a77e926bce0deda029b5adc716fe2"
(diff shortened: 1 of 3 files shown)


deivid-rodriguez (approved):
💪 💯

jakecoffman:
The hex tests are failing because they rely on https://dependabot-private.fly.dev which is currently down. I'll mark them as pending and file an issue to circle back and fix them so they don't rely on a third party server to be running to succeed.

deivid-rodriguez (approved):

hex/spec/dependabot/hex/update_checker_spec.rb
```diff
@@ -359,7 +359,7 @@
       end
     end
 
-    context "with a dependency from a private repo" do
+    pending "with a dependency from a private repo" do
```
jeffwidman, line 362: Can you please add an explanatory code comment when making stuff like this pending?

I recently went through some pain on old tests that were marked pending w/o any comments trying to figure out why they were pending... and the test had been that way for several years, so the files had been restructured multiple times so it was extremely painful trying to navigate `git blame` to figure out why they were disabled... particularly when part of the file had been moved and part hadn't... and I wanted to ensure I understood why it was pending to ensure i didn't misunderstand what it was trying to test. So leaving the code comment in-line ensures it's clear to future devs why it's temp marked pending as that comment will follow the test even if it gets moved to a different file.

hex/spec/dependabot/hex/file_updater/lockfile_updater_spec.rb
```diff
@@ -372,6 +372,7 @@
       end
 
       it "updates the dependency version in the lockfile" do
+        pending "dependabot-private.fly.dev is offline"
```
jeffwidman, line 375:
```suggestion
    pending "test depends on dependabot-private.fly.dev which went offline, so rewrite w/o that dependency"
```

PR #5231, opened 2022-06-06
Yarn: support recursive paths expansion

FunkyloverOne:
So now it can successfully expand a glob like `packages/*/*`.
Resolves https://github.com/dependabot/dependabot-core/issues/1777

~~**NOTE**: this PR is based on top of https://github.com/dependabot/dependabot-core/pull/5220~~

**UPDATE**: the previously mentioned PR is already merged, so this one is rebased now.

Issue #6923, opened 2023-03-27
FileFetcher specs (at least the ones in `npm_and_yarn`) don't exercise real usage


deivid-rodriguez:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code improvement description

I noticed that our current file fetcher specs for `npm_and_yarn` instantiate a `FileFetcher` with `repo_contents_path: nil`. That essentially means that [the fetcher](https://github.com/dependabot/dependabot-core/blob/4840ffeab7bda6d1ad9bd291c9abbc1c77bab4e9/common/lib/dependabot/file_fetchers/base.rb#L43C2-L52) by default does not clone the target repository but fetches files through the files API.

However, that's not how `npm_and_yarn` works [by default](https://github.com/dependabot/dependabot-core/blob/4840ffeab7bda6d1ad9bd291c9abbc1c77bab4e9/npm_and_yarn/lib/dependabot/npm_and_yarn.rb#L28), since it always clones.

I assume this is also the case for other top level folders that always clone full repositories.

For these cases, file fetcher specs feel quite useless since they're not exercising the real code actually run by Dependabot.

Should we migrate these specs to actually clone real repositories? For example, some test repositories under the `dependabot-fixtures` organization? We currently do that for a single spec, the one that checks fetching git-lfs files.

PR #6932, opened 2023-03-28
Implement double star glob workspaces for npm & yarn

deivid-rodriguez:
They were only supported on `lerna.json` files, but not on `package.json` files. Now they are supported everywhere.

This is a follow up to https://github.com/dependabot/dependabot-core/pull/5231 and an alternative to https://github.com/dependabot/dependabot-core/pull/4549.


deivid-rodriguez:
This should actually close https://github.com/lensapp/lens/pull/7435!

deivid-rodriguez:
From that PR I realized that they just moved to a monorepo setup, so probably not a lot of people has copied their `**/*` configuration. But still feels like a pretty common thing to do in monorepos, so good to properly support it!

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -367,19 +366,7 @@ def fetch_lerna_packages_from_path(path, nested = false)
             fetch_file_if_present(File.join(path, "npm-shrinkwrap.json"))
           ].compact
         rescue Dependabot::DependencyFileNotFound
-          matches_double_glob =
```
jakecoffman, line 370: Was this removed because it was a fallback, but now it should work correctly the first time?

deivid-rodriguez: Yes, this cleanup should come for free after this fix since globs should be now properly expanded before they are passed to this method.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -419,11 +405,12 @@ def find_directories(glob)
...
 
-        results = paths.select { |filename| File.fnmatch?(glob, filename) }
+        results = paths.select { |filename| File.fnmatch?(glob, filename, File::FNM_PATHNAME) }
         return results unless ignored_glob
 
-        results.reject { |filename| File.fnmatch?(ignored_glob, filename) }
+        results.reject { |filename| File.fnmatch?(ignored_glob, filename, File::FNM_PATHNAME) }
       end
 
       def recursive_find_directories(glob, prefix = "")
```
jakecoffman, line 416: Is this doing what `Dir.glob` does? Is there an opportunity to simplify this now that npm_and_yarn is a cloned ecosystem?

jakecoffman (approved):
LGTM, one suggestion about simplifying by using `Dir.glob` that we can take a pass at later if it makes sense.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
deivid-rodriguez: Yeah, this is definitely possible now indeed.

I thought of that but I wasn't clear on how to apprach tests so went with fixing (instead of changing) the current logic.

I think whenever we tackle #6923 would be a good opportunity to simplify this.

PR #6934, opened 2023-03-29
[Updater] Update the service's dependency metadata in one place

brrygrdn:
Follows up on #6906 

Now we've established the fast-fail approach to Dependency parsing in production, this PR makes a small follow-up change to relay the outcome of the parsing to the service as metadata we use to maintain an understanding of the current state of the project to immediately after the parsing takes place.

This will remove the need to perform this call in each operation class, avoiding repetition.


landongrindheim (approved):
🙌

PR #6937, opened 2023-03-29
Document the credential proxy

jeffwidman:
Add a section to the Readme documenting the credential proxy.

[rendered](https://github.com/dependabot/dependabot-core/tree/document-credential-proxy#private-registry-credential-management)

This provides three benefits:
1. Users are reassured that GitHub keeps their secrets safe.
2. Security researchers who manage to hack `dependabot-core` aren't surprised when we tell them that doesn't mean they hacked the entire Dependabot Service at GitHub.
3. Clarifies why adding support for private registries to `dependabot-core` isn't enough for those registries to be supported by the Dependabot service that GitHub runs. We still have to plumb those auth schemes into the proxy.

Perhaps down the road we may be able to open source the proxy so others can benefit from the increased security, but that's a much larger conversation that we're not quite ready to have.

For now I tacked this onto the Readme... long term the Readme would probably benefit from getting diced up into shorter docs... but that's out of scope for this PR, and I'd rather land all the content before looking at how to dice it up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,6 +293,39 @@ This is a "meta" gem, that simply depends on all the others. If you want to
 automatically include support for all languages, you can just include this gem
 and you'll get all you need.
 
+### Private Registry Credential Management
+
+For many ecosystems, Dependabot Core supports private registries. Sometimes this happens by passing the private registry
+credentials directly to the native package managers (`npm`, `pip`, `bundler`, etc), other times it happens within the
+Dependabot Core Ruby code.
+
+```mermaid
+sequenceDiagram
+    Private Registry Credentials->>Dependabot Core:<br />
+    Dependabot Core->>Native Package Managers:<br />
+    Native Package Managers->>Package Registries:<br />
+    Dependabot Core->>Package Registries:<br />
+```
+
+While simple and straightforward, this is a security risk for ecosystems that allow running untrusted code within their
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -293,6 +293,39 @@ This is a "meta" gem, that simply depends on all the others. If you want to
...
 
+### Private Registry Credential Management
+
+For many ecosystems, Dependabot Core supports private registries. Sometimes this happens by passing the private registry
+credentials directly to the native package managers (`npm`, `pip`, `bundler`, etc), other times it happens within the
+Dependabot Core Ruby code.
+
+```mermaid
+sequenceDiagram
+    Private Registry Credentials->>Dependabot Core:
```
jeffwidman, line 304: This diagram failed to parse unless these lines had some text after the `:`, which is why there's a trailing space. That'd be fine if i wanted to add text to the arrows, but I'd prefer to keep things simple.

However, my editor wants to strip all trailing whitespace upon save.

I couldn't find an alternative syntax in the mermaid docs for signaling that no text was needed? Only thing I could come up with is the `flowchart LR`, but that wouldn't be consistent with the second diagram where I _do_ want to document the arrows... 🤷‍♂️ 

Any tips from the mermaid experts on how to make this work without a trailing space?

jeffwidman: Discovered that adding `<br />` works. `&nbsp;` doesn't--it results in adding `"&nbsp;"` as a visible label on the arrows, which is def not what we want.

PR #6939, opened 2023-03-29
[Updater] Extract RefreshVersionPullRequest as an Operation class

brrygrdn:
Follows on from https://github.com/dependabot/dependabot-core/pull/6866, #6884 

This branch peels out the Updater code required to accept a job which is "refreshing" an existing Dependabot PR for a version update using the same approach as before:
- Remove any code related to non-concerns for this operation type, in this case security-updates primarily

### Why Refresh?

We use different terminology for this operation in some places depending on the trigger, but it isn't correct to think of it as a "rebase" despite commands like `@dependabot rebase` existing.

The actual behaviour is one of three outcomes:
- If the project still needs the same dependency at the same version, then the PR is updates on top of the current head of the target branch; this is effectively a rebase
- If the project still needs the same dependency updated, but there is a newer version, then a new PR is created to supersede the existing one
- If a number of other circumstances have changed, such as the Dependency now being removed, ignored, updated by another commit, etc, then the PR is closed.

The term "update" is fairly saturated in Dependabot for obvious reasons, so I went with "refresh" as a good-enough catch all term.

## Notes

As with #6866, this code is intentionally not DRY'd out as much as possible. We've de-risked this approach and this branch completes the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)), so the next step will be push the shared code into an interface mixin that is common to all Operation classes before starting to peel out the two Security Operations.

This PR is also branched off https://github.com/dependabot/dependabot-core/pull/6934 which should ship first as an enabling change.

### DIFF
--- a/updater/lib/dependabot/updater/operations/refresh_version_pull_request.rb
+++ b/updater/lib/dependabot/updater/operations/refresh_version_pull_request.rb
@@ -70,12 +70,7 @@ module Dependabot
                                    "updates")
           end
 
-          # Consider updating vulnerable deps first. Only consider the first 10,
-          # though, to ensure they don't take up the entire update run
-          deps = allowed_deps.select { |d| job.vulnerable?(d) }.sample(10) +
-                 allowed_deps.reject { |d| job.vulnerable?(d) }
-
-          deps
+          allowed_deps
         end
 
         def check_and_update_existing_pr_with_error_handling(dependencies)
@@ -100,24 +95,6 @@ module Dependabot
             # we should close the PR as at least one thing we changed has been
             # removed from the project.
             close_pull_request(reason: :dependency_removed)
-            return
-          end
-
-          # NOTE: Prevent security only updates from turning into latest version
-          # updates if the current version is no longer vulnerable. This happens
-          # when a security update is applied by the user directly and the existing
-          # pull request is rebased.
-          if job.security_updates_only? &&
-             dependencies.none? { |d| job.allowed_update?(d) }
-            lead_dependency = dependencies.first
-            if job.vulnerable?(lead_dependency)
-              Dependabot.logger.info(
-                "Dependency no longer allowed to update #{lead_dependency.name} #{lead_dependency.version}"
-              )
-            else
-              Dependabot.logger.info("No longer vulnerable #{lead_dependency.name} #{lead_dependency.version}")
-            end
-            close_pull_request(reason: :up_to_date)
             return
           end
 
@@ -213,7 +190,7 @@ module Dependabot
         end
 
         def raise_on_ignored?(dependency)
-          job.security_updates_only? || ignore_conditions_for(dependency).any?
+          ignore_conditions_for(dependency).any?
         end
 
         def ignore_conditions_for(dep)
@@ -230,7 +207,7 @@ module Dependabot
           end
           Dependabot::Config::UpdateConfig.
             new(ignore_conditions: ignore_conditions).
-            ignored_versions_for(dep, security_updates_only: job.security_updates_only?)
+            ignored_versions_for(dep, security_updates_only: false)
         end
 
         def update_checker_for(dependency, raise_on_ignored:)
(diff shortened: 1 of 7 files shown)


updater/lib/dependabot/updater/operations/refresh_version_pull_request.rb
```diff
@@ -0,0 +1,346 @@
...
+# that updates a dependnency to the latest permitted version.
+#
+# It will determine if the existing diff is still relevant, in which case it
+# functions similar to a "rebase", but in the case where the project folder's
+# dependencies have changed or a newer version is available, it will supersede
+# the existing pull request with a new one for clarity.
+module Dependabot
+  class Updater
+    module Operations
+      class RefreshVersionPullRequest
```
landongrindheim, line 13: 🤔 While `Update` is inherent in this name, I wonder if it should be included to make a stronger distinction between it and a security update. I haven't spent time with it, so this is more of a knee-jerk reaction.

Since you spent time with this, what are your thoughts on `RefreshVersionUpdatePullRequest` vs `RefreshVersionPullRequest`?

landongrindheim (approved):

updater/lib/dependabot/updater/operations/refresh_version_pull_request.rb
```diff
@@ -0,0 +1,346 @@
...
+                                   "updates")
+          end
+
+          allowed_deps
+        end
+
+        # rubocop:disable Metrics/AbcSize
+        # rubocop:disable Metrics/PerceivedComplexity
+        #
+        # TODO: Push checks on dependencies into Dependabot::DependencyChange
```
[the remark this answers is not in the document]
brrygrdn: I'm copying this over from the Updater where the TODO was originally marked out. I'm not sure it applies in the way I thought it might, I was thinking that as we filtered and set the dependencies, we could do some of the work in that class as a kind of batch compiler, so we'd set @dependencies, then verify!, then generate the change all on that object.

I'm not sure that works anymore as there's a pattern emerging between "single" and "batch" operations I hadn't really thought much about - i.e. security updates, and all kinds of rebases are single, version updates (and groups) are batches - so this might make more sense fitting into that.

updater/lib/dependabot/updater/operations/refresh_version_pull_request.rb
brrygrdn: Yeah, I think I've started biasing towards avoiding the word `Update` as it feels really implicit, but I also think the 'nouns' are a little week in terms of a PR that updates to the latest version vs a PR that updates to a secure version, etc.

We have very fixed languages/ideas in the service and I'm trying not to bleed them into Core/the Updater, but it probably makes sense to just use `VersionUpdate` `SecurityUpdate` to be honest?

landongrindheim: > We have very fixed languages/ideas in the service and I'm trying not to bleed them into Core/the Updater, but it probably makes sense to just use VersionUpdate SecurityUpdate to be honest?

I think it does make sense. I think security updates are a semi-abstract idea in Core, so naming them might make that more concrete. In all of this, I trust your judgment and defering until later is 👍 by me 🙂

brrygrdn: I went ahead and renamed it as SecurityUpdate/VersionUpdate are the features we talk about in our documentation and for better or worse, the abstraction is through into the Updater.

I think by splitting this up we can start to gain nuance in future, but for now, one naming convention is a little easier to parse.

brrygrdn:
@landongrindheim I've added a last commit removing some dead code from the `def dependencies` method for the new operation - this operation cannot be invoked if `job.dependencies` is empty, so we will always early return the outcome of the `dependency_snapshot.dependencies.select` statement.

landongrindheim (approved):

PR #6942, opened 2023-03-30
Revert disabling hex private registry unit test

jeffwidman:
The private registry server is back up, so these are now failing (yay for `pending` showing when this happens!)

This reverts
https://github.com/dependabot/dependabot-core/pull/6930/commits/1c4cc2de36e0f502c99e041af1be2b812454d17f.

Long term it'd still be convenient if these didn't depend on an external private registry; however, they do serve as a nice integration test (or a surprising one if it gets upgraded unexpectedly!)

For now though this unblocked `main`.


jakecoffman (approved):
We should get some VCR cassettes around this.

jakecoffman:
True, VCRs won't help us if Hex is making the calls so we need to figure that out.

PR #6944, opened 2023-03-30
Refactor npm ecosystem lockfile parsing

deivid-rodriguez:
As I start looking into pnpm, I noticed that some classes have mixed concerns including code specific to yarn and npm, and are going to get much worse when adding pnpm.

This PR tries to improve this situation by extracting package manager specific classes to deal with parsing lockfiles.


jeffwidman (approved):
✨

PR #6950, opened 2023-03-30
Fix "no files were updated!" errors in workspaces

deivid-rodriguez:
When updating a dependency not included in the root `package.json` file, and that only requires `package-lock.json` changes, we ended up running `npm install --lockfile-only` (without explicit `package@version` arg), and that would result in no lockfile changes and the above error.

We need to force the update through a `package@version` CLI argument. However, as it was already explained on a comment, in this case this causes a new dependency to be added to the top level `package.json` file, which is not desirable. To workaround that, we save the original `package.json` content, and restore it after the upgrade and re-run `npm install --lockfile-only` again to bring the lockfile in sync.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
@@ -186,7 +186,7 @@ module Dependabot
           end
 
           # TODO: Update the npm 6 updater to use these args as we currently
-          # do the same in the js updater helper, we've kept it seperate for
+          # do the same in the js updater helper, we've kept it separate for
           # the npm 7 rollout
           install_args = top_level_dependencies.map { |dependency| npm_install_args(dependency) }
 
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
@@ -241,8 +241,8 @@ module Dependabot
         #
         # Other npm flags:
         # - `--force` ignores checks for platform (os, cpu) and engines
-        # - `--dry-run=false` the updater sets a global .npmrc with dry-run:
-        #   true to work around an issue in npm 6, we don't want that here
+        # - `--dry-run=false` the updater sets a global .npmrc with `dry-run: true`
+        #   to work around an issue in npm 6, we don't want that here
         # - `--ignore-scripts` disables prepare and prepack scripts which are
         #   run when installing git dependencies
         def run_npm_install_lockfile_only(*install_args)
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
@@ -176,48 +176,28 @@ module Dependabot
             dependency_in_package_json?(dependency)
           end
 
-          # NOTE: When updating a dependency in a nested workspace project we
-          # need to run `npm install` without any arguments to update the root
-          # level lockfile after having updated the nested packages package.json
-          # requirement, otherwise npm will add the dependency as a new
-          # top-level dependency to the root lockfile.
-          install_args = ""
-          if dependencies_in_current_package_json
-            # TODO: Update the npm 6 updater to use these args as we currently
-            # do the same in the js updater helper, we've kept it seperate for
-            # the npm 7 rollout
-            install_args = top_level_dependencies.map { |dependency| npm_install_args(dependency) }
+          unless dependencies_in_current_package_json
+            # NOTE: When updating a dependency in a nested workspace project, npm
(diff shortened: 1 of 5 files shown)


npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -176,48 +176,28 @@ def run_npm8_top_level_updater(top_level_dependencies:)
...
-            "install",
-            "<install_args>",
-            "--force",
-            "--dry-run",
-            "false",
-            "--ignore-scripts",
-            "--package-lock-only"
-          ].join(" ")
+          # TODO: Update the npm 6 updater to use these args as we currently
+          # do the same in the js updater helper, we've kept it seperate for
```
jeffwidman, line 189:
```suggestion
          # do the same in the js updater helper, we've kept it separate for
```

I realize you're copy/pasting, but might as well clean it up...

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -256,6 +236,41 @@ def flattenend_manifest_dependencies
             end
         end
 
+        # Runs `npm install` with `--package-lock-only` flag to update the
+        # lockfiile.
+        #
+        # Other npm flags:
+        # - `--force` ignores checks for platform (os, cpu) and engines
+        # - `--dry-run=false` the updater sets a global .npmrc with dry-run:
+        #   true to work around an issue in npm 6, we don't want that here
```
jeffwidman, line 245: I'd treat `dry-run: true` as a single word and not break across lines in the middle of that to reduce risk of misreading.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
```diff
@@ -256,6 +236,41 @@ def flattenend_manifest_dependencies
...
+          fingerprint = [
+            "npm",
+            "install",
+            install_args.empty? ? "" : "<install_args>",
+            "--force",
+            "--dry-run",
+            "false",
+            "--ignore-scripts",
+            "--package-lock-only"
+          ].join(" ")
```
jeffwidman, line 269: I realize this is copy/pasted, but doesn't seem a little weird this is copy/pasted vs an inline function (does Ruby allow those?)  where the fingerprint feeds in a different value than the command call...  nbd though.

jeffwidman (approved):
Makes sense to me!

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
deivid-rodriguez: You mean quoting it, right?

jeffwidman: I was thinking:
```
        # - `--dry-run=false` the updater sets a global .npmrc with
        #    dry-run: true to work around an issue in npm 6, we don't want that here
```

But quoting it is probably even better
```
        #    `dry-run: true` to work around an issue in npm 6, we don't want that here
```

Regardless of quoting or not, I still wouldn't break across lines... too easy for the eye to miss the quote if partially split across lines.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
jeffwidman: Yeah, I dunno if this is doable in Ruby, but in python or Go I'd do something like:

```
        def run_npm_install_lockfile_only(*install_args)
           def generate_command_string(install_args)      
            [
              "npm",
              "install",
              install_args,
              "--force",
              "--dry-run",
              "false",
              "--ignore-scripts",
              "--package-lock-only"
            ].join(" ")

          command = generate_command_string((*install_args).join(" "))
          # Make `install_args` generic to Sentry for easier grouping and avoid risk of exposing PII
          fingerprint = generate_command_string((install_args.empty? ? "" : "<install_args>"))
```

No idea if it's idiomatic in Ruby though...

deivid-rodriguez: Ahhh, I see. You'd use a lambda for that, see https://rubyapi.org/3.1/o/proc. Agree on doing it separately since it should be happening everywhere and would deviate too much from the point of this PR.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb
deivid-rodriguez: Aaah I see, fixed!

riski2995: # - [ ] ~~````

- - [ ] # > ````~~~~@

 ~~

PR #6954, opened 2023-03-30
Discourage filing issues about Alerts / Dependency Graph

jeffwidman:
We get the occasional issue about Alerts / Dependency Graph, and they're out of scope for both this issue tracker and the team that watches the issue tracker... ie, there's nothing we can do.

So let's point folks in the right direction from the get-go.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -5,8 +5,10 @@ body:
   - type: markdown
     attributes:
       value: |
+        This issue-tracker is meant for issues related to Dependabot's updating logic, a good rule of thumb is that if you have questions about the _diff_ in a PR, it belongs here.
+
+        Issues related to security alerts or Dependency Graph should instead be filed as a [Code Security discussion](https://github.com/orgs/community/discussions/categories/code-security).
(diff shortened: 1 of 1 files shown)


abdulapopoola (approved):

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -5,8 +5,10 @@ body:
   - type: markdown
     attributes:
       value: |
+        This issue-tracker is meant for issues related to Dependabot's updating logic, a good rule of thumb is that if you have questions about the _diff_ in a PR, it belongs here.
+
+        Issues related to security alerts or Dependency Graph should instead be filed as a [Code Security discussion](https://github.com/orgs/community/discussions/categories/code-security).
```
jeffwidman, line 10: @courtneycl Is `Code Security` the right category for Dependency Graph? or would you prefer users file these discussions under a different category / issue tracker?

courtneycl: yes it is!

jeffwidman: Confirmed via slack that this is the correct category.

PR #6961, opened 2023-03-31
[Updater] Extract UpdateVulnerableVersion as an Operation class

brrygrdn:
Follows on from https://github.com/dependabot/dependabot-core/pull/6866, https://github.com/dependabot/dependabot-core/pull/6884, https://github.com/dependabot/dependabot-core/pull/6939

This branch peels out the Updater code required to accept a job which is generating a new Dependabot PR for a security update using the typical approach of:
- Remove any code related to non-concerns for this operation type, in this case version updates and rebases

### Notes

The one significant deviation from the previous PRs is that this does make test changes. In order to pass the existing tests, some setups need to be completed by ensuring there is a value for `job.dependencies`.

In real-world conditions, we would never start a Security Update job without at least one, and normally only one, specific target dependency to fix.

This should not realistically result in a change of behaviour, it is essentially removing an unsupported "working by accident" codepath where the Dependency files contain exactly the insecure dependency to be fixed.

### DIFF
--- a/updater/lib/dependabot/updater/operations/create_security_update_pull_request.rb
+++ b/updater/lib/dependabot/updater/operations/create_security_update_pull_request.rb
@@ -70,7 +70,7 @@ module Dependabot
 
           log_checking_for_update(dependency)
 
-          return if all_versions_ignored?(dependency, checker)
+          Dependabot.logger.info("Latest version is #{checker.latest_version}")
 
           # If the dependency isn't vulnerable or we can't know for sure we won't be
           # able to know if the updated dependency fixes any advisories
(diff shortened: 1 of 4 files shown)


brrygrdn:
Before merging this, there is some cleanup work I'd like to push forward on the existing three Operation classes to avoid continuing the heavy duplicate since Security Updates have so many more helper methods.

I've created this as a reference to make sure my cleanup aligns with the "next" step in the refactor without breaking the tests.

brrygrdn:
This PR shouldn't merge until the smoke tests have been updated: https://github.com/dependabot/smoke-tests/pull/51

Currently, the security update smoke test goes into the `legacy_run` codeline by accident.

updater/lib/dependabot/updater/operations/create_security_update_pull_request.rb
```diff
@@ -0,0 +1,401 @@
...
+              "#{latest_allowed_version} because of the following " \
+              "conflicting #{dependencies_pluralized}:\n\n#{dep_messages}"
+          else
+            "The latest possible version of #{checker.dependency.name} that can " \
+              "be installed is #{latest_allowed_version}"
+          end
+        end
+        ### END: Security Update Helpers
+
+        def requirements_to_unlock(checker)
```
deivid-rodriguez, line 357: We should be able to fully move this method to `UpdateChecker` after #5902 and not keep it duplicated for all operations.

deivid-rodriguez (approved):
First time I review these refactorings and I love them!

PR #6963, opened 2023-03-31
Add links to issue tracker / changelog to gemspecs

jeffwidman:
In particular adding the changelog metadata makes Dependabot PR's show the changelog contents when it creates PRs.

For now I pointed them all at the same changelog because we don't maintain separate changelogs/issue trackers for the individual sub-gems. Easy enough to change later if needed.

### DIFF
--- a/common/dependabot-common.gemspec
+++ b/common/dependabot-common.gemspec
@@ -14,6 +14,11 @@ Gem::Specification.new do |spec|
   spec.homepage     = "https://github.com/dependabot/dependabot-core"
   spec.license      = "Nonstandard" # License Zero Prosperity Public License
 
+  spec.metadata = {
+    "issue_tracker_uri" => "https://github.com/dependabot/dependabot-core/issues",
+    "changelog_uri" => "https://github.com/dependabot/dependabot-core/blob/main/CHANGELOG.md"
+  }
+
   spec.require_path = "lib"
   spec.files        = []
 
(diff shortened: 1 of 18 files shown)


deivid-rodriguez (approved):
🎉

common/dependabot-common.gemspec
```diff
@@ -7,13 +7,18 @@ Gem::Specification.new do |spec|
   spec.version      = Dependabot::VERSION
   spec.summary      = "Shared code used between Dependabot package managers"
   spec.description  = "Automated dependency management for Ruby, JavaScript, " \
-                      "Python, PHP, Elixir, Rust, Java, .NET, Elm and Go"
+                      "Python, Docker, PHP, Elixir, Rust, Java, .NET, Elm and Go"
```
abdulapopoola, line 10: Dart too?

abdulapopoola: Also, can we maybe sort this list?

abdulapopoola (approved):

common/dependabot-common.gemspec
jeffwidman: oof, I didn't mean to ship this with this PR, as I've got another I'm working on right now to rewrite these descriptions... they're super confusion on RubyGems (that was actually what prompted me wanting to include fixing this as part of our internal OKR around contributor docs)

jeffwidman: Removed from this PR.

PR #6965, opened 2023-03-31
Use the correct author email

jeffwidman:
Use the correct email for open source projects at GitHub.

The previous one is a leftover from before Dependabot was acquired and is no longer monitored.

### DIFF
--- a/common/dependabot-common.gemspec
+++ b/common/dependabot-common.gemspec
@@ -10,7 +10,7 @@ Gem::Specification.new do |spec|
                       "Python, PHP, Elixir, Rust, Java, .NET, Elm and Go"
 
   spec.author       = "Dependabot"
-  spec.email        = "support@dependabot.com"
+  spec.email        = "opensource@github.com"
   spec.homepage     = "https://github.com/dependabot/dependabot-core"
   spec.license      = "Nonstandard" # License Zero Prosperity Public License
 


common/dependabot-common.gemspec
```diff
@@ -10,7 +10,7 @@ Gem::Specification.new do |spec|
                       "Python, PHP, Elixir, Rust, Java, .NET, Elm and Go"
 
   spec.author       = "Dependabot"
-  spec.email        = "support@dependabot.com"
+  spec.email        = "opensource@github.com"
```
abdulapopoola, line 13: it seems we have a lot of spec files that need to be touched?

jeffwidman: Thanks for the thorough review to doublecheck for other usages! ✨ 

Spec files are test files, so it's okay to leave them as-is.

I did the same search, and the only one I saw that isn't spec/fixtures is:
* https://github.com/dependabot/dependabot-core/issues/6968

But that's out of scope for this particular PR, so IMO we should merge this one as-is and handle that as a separate issue.

jeffwidman: Also, feel free to correct me if you see other usages that aren't tests / fixtures.

abdulapopoola (approved):

PR #6966, opened 2023-03-31
Remove checkbox for searching for duplicates

jeffwidman:
Tech Debt issues are rarely filed by n00bs, but almost always by folks already familiar with the project.

Because they tend to be more knowledgable, they generally are already aware of duplicates, so this just gets in the way.

### DIFF
--- a/.github/ISSUE_TEMPLATE/tech-debt.yml
+++ b/.github/ISSUE_TEMPLATE/tech-debt.yml
@@ -2,14 +2,6 @@ name: ⚙ Tech Debt
 description: Tech debt is for code improvements that do not change the user-facing behavior (ie, neither adding features nor fixing bugs).
 labels: "T: tech-debt ⚙️"
 body:
-  - type: checkboxes
-    attributes:
-      label: Is there an existing issue for this?
-      description: Please search existing issues to avoid creating duplicates.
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/tech-debt.yml
```diff
@@ -2,14 +2,6 @@ name: ⚙ Tech Debt
 description: Tech debt is for code improvements that do not change the user-facing behavior (ie, neither adding features nor fixing bugs).
 labels: "T: tech-debt ⚙️"
 body:
```
abdulapopoola, line 4: We still have dupes even with this check; I believe our triage process can take care of this going forward

abdulapopoola (approved):

deivid-rodriguez (approved):

.github/ISSUE_TEMPLATE/tech-debt.yml
jeffwidman: Yeah, ☝️ is probably a good argument for removing it from the templates for bugs/feature requests as well... but I'm not touching those until I finish the readme reorg as I plan to cross link a bunch of stuff back to readme sections...

PR #6971, opened 2023-03-31
Clarify `summary`/`description` for RubyGems

jeffwidman:
One of the things I noticed a long time ago on RubyGems is it's not really clear how all the Dependabot gems fit together... which is mostly because their `description` metadata field all inherit from `common`.

If you're trying to get going with the Dependabot gem, it's unclear which one you actually want.

So this fixes that.

### DIFF
--- a/omnibus/dependabot-omnibus.gemspec
+++ b/omnibus/dependabot-omnibus.gemspec
@@ -5,23 +5,23 @@ Gem::Specification.new do |spec|
     Bundler.load_gemspec_uncached("../common/dependabot-common.gemspec")
 
   spec.name         = "dependabot-omnibus"
-  spec.summary      = "Meta-package that depends on all dependabot package " \
-                      "managers"
-  spec.version      = common_gemspec.version
-  spec.description  = common_gemspec.description
+  spec.summary      = "Meta-package that provides all the gems included in Dependabot"
+  spec.description  = "Dependabot-Omnibus provides all the gems included in Dependabot. " \
+                      "Dependabot provides automated dependency updates for multiple package managers."
 
   spec.author       = common_gemspec.author
   spec.email        = common_gemspec.email
   spec.homepage     = common_gemspec.homepage
   spec.license      = common_gemspec.license
 
-  spec.required_ruby_version = common_gemspec.required_ruby_version
-  spec.required_rubygems_version = common_gemspec.required_ruby_version
-
   spec.metadata = {
     "issue_tracker_uri" => common_gemspec.metadata["issue_tracker_uri"],
     "changelog_uri" => common_gemspec.metadata["changelog_uri"]
   }
+
+  spec.version = common_gemspec.version
+  spec.required_ruby_version = common_gemspec.required_ruby_version
+  spec.required_rubygems_version = common_gemspec.required_ruby_version
 
   spec.require_path = "lib"
   spec.files        = ["lib/dependabot/omnibus.rb"]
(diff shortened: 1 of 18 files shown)


omnibus/dependabot-omnibus.gemspec
```diff
@@ -5,24 +5,24 @@ Gem::Specification.new do |spec|
...
-  spec.required_rubygems_version = common_gemspec.required_ruby_version
-
   spec.metadata = {
     "issue_tracker_uri" => common_gemspec.metadata["issue_tracker_uri"],
     "changelog_uri" => common_gemspec.metadata["changelog_uri"]
   }
 
+  spec.version = common_gemspec.version
+  spec.required_ruby_version = common_gemspec.required_ruby_version
+  spec.required_rubygems_version = common_gemspec.required_ruby_version
```
jeffwidman, line 24: I moved these under the `spec.metadata` for consistency of ordering with the other files, and then on the other files that resulted in moving `spec.require_path`/`spec.files` for consistency... but no actual changes in any of the values.

jeffwidman:
The only values that changed in this PR are `summary` / `description`... the others are just re-location for ordering consistency across all the files.

omnibus/dependabot-omnibus.gemspec
Samboun01: only Solution for this is to make sure that the url you have passed is legal, with a proper protocol. The best way to do it is validating process

deivid-rodriguez (approved):
Definitely an improvement from inheriting summary & description from core.

I think the

> Dependabot provides automated dependency updates for multiple package managers.

part is slightly confusing, since I initially though it was referred to the specific package managers of each ecosystem (so, for example, it made sense to me on `dependabot-npm_and_yarn`, but not on `dependabot-docker`, since there's no really such thing as a package manager for Docker).

Then I realized it refers to the different ecosystem specific gems, and it's kind of an introductory sentence to recommend `dependabot-omnibus`?

To reduce this confusion a bit, I'd maybe go with "languages" rather than "package managers", although that's not fully accurate either. Maybe something a long the lines of

> Dependabot provides automated updates for multiple files (either dependency manifests or dependency lock files) used to keep track of software dependencies.

Anyways, definitely not a blocker since this is an improvement over what we have now! Thank you!!

jeffwidman:
Hmm... good point re: confusion. 

TBH, I'm not in love with any of the options on the table, I think the problem might be that it wasn't immediately clear that it was an introductory sentence about "overall Dependabot", and making that clear would sidestep any questions about the scope of languages/package managers.

Overall I expect anyone who hits this gem will already be somewhat familiar with Dependabot, and if not it's easy to Google--so I don't think the tricky bit is answering "what is dependabot"... instead I expect a user wants to grab "the dependabot gem" but it's not clear which gem to use... so IMO our goal with these descriptions should focus on answering "how does this particular gem fit within Dependabot?"...

Currently it's:
```
"Dependabot-Bundler provides support for bumping Ruby (bundler) gems via Dependabot Core. " \
"Dependabot provides automated dependency updates for multiple package managers. " \
"You probably want the meta-gem dependabot-omnibus."
```

Would  this clarify that this is a piece of a larger whole instead?

```
"Dependabot-Bundler provides support for bumping Ruby (bundler) gems via Dependabot Core. " \
"Dependabot Core provides automated dependency updates for multiple package managers. " \
"You probably want the Dependabot Core meta-gem dependabot-omnibus."
```

jeffwidman:
Perfecto!

deivid-rodriguez (approved):

PR #6978, opened 2023-04-02
build(deps): bump GOLANG_VERSION to 1.20.2

yeikel:


go_modules/Dockerfile
```diff
@@ -4,10 +4,13 @@ ARG TARGETARCH
 USER root
 
 # Install Go
-ARG GOLANG_VERSION=1.20
-# You can find the sha here: https://storage.googleapis.com/golang/go${GOLANG_VERSION}.linux-amd64.tar.gz.sha256
-ARG GOLANG_AMD64_CHECKSUM=5a9ebcc65c1cce56e0d2dc616aff4c4cedcfbda8cc6f0288cc08cda3b18dcbf1
-ARG GOLANG_ARM64_CHECKSUM=17700b6e5108e2a2c3b1a43cd865d3f9c66b7f1c5f0cec26d3672cc131cc0994
+ARG GOLANG_VERSION=1.20.2
+
+# You can find the sha here: https://dl.google.com/go/go${GOLANG_VERSION}.linux-amd64.tar.gz.sha256
```
yeikel, line 9: Downloading from `https://dl.google.com` and validating from `https://storage.googleapis.com` felt wrong

Also, the comment was misleading as the URL changes depending on the architecture

jeffwidman: Makes sense... what do you think about switching both the comment and download links to `https://go.dev/dl/`?

Since it looks like the go project is now using that as their recommended domain for downloads? (IIRC, that wasn't true historically when this was originally setup)

then could be simply:

```dockerfile
ARG GOLANG_VERSION=1.20.2
# You can find the SHA's here: https://go.dev/dl/
ARG GOLANG_AMD64_CHECKSUM=4eaea32f59cde4dc635fbc42161031d13e1c780b87097f4b4234cfce671f1768
ARG GOLANG_ARM64_CHECKSUM=78d632915bb75e9a6356a47a42625fd1a785c83a64a643fedd8f61e31b1b3bef
```

NBD either way, I'm fine with whatever you prefer.

jeffwidman (approved):
Thanks!

Looks fine, one optional suggestion, totally up to you. 

Will merge once you get a chance to reply.

yeikel:
> Thanks!
> 
> Looks fine, one optional suggestion, totally up to you.
> 
> Will merge once you get a chance to reply.

Thank you. I applied your feedback. 

I had to enable redirection (--location) because `https://go.dev/dl/` ultimately redirects to `https://dl.google.com` and curls does not follow redirects by default

jeffwidman:
> I had to enable redirection (`--location`) because `https://go.dev/dl/` ultimately redirects to `https://dl.google.com` and `curl` does not follow redirects by default

Ah that's right, that's why we did it that way originally... This means it'll cost us an extra http call on every CI download/build, but I think that's probably fine... However, I don't know for sure, so I'm going to get a second opinion from the team before merging.

yeikel:
> > I had to enable redirection (`--location`) because `https://go.dev/dl/` ultimately redirects to `https://dl.google.com` and `curl` does not follow redirects by default
> 
> Ah that's right, that's why we did it that way originally... This means it'll cost us an extra http call on every CI download/build, but I think that's probably fine... However, I don't know for sure, so I'm going to get a second opinion from the team before merging.

That's fair. It should be a very cheap operation but I can revert it if needed

Keep me posted

deivid-rodriguez (approved):
I don't have a strong opinion either way.

mctofu:
I'd lean towards the `go.dev/dl` links since those are listed as the official download links at https://go.dev/dl/. Assume those would be less likely to break if they make changes to the download hosting backends.

jeffwidman:
Sounds good, I'll merge this later today.

PR #6989, opened 2023-04-03
[Updater] Absorb Security Advisory and Ignore Conditions into Job

brrygrdn:
Before adding two more "Operation" classes for Security Updates, we should ensure that any code that is starting to be carried into _every_ Operation class so far is de-duplicated as much as possible.

## Changes to Dependabot::Job

This PR follows up the observation that the `Dependabot::Job` already implements a `security_advisories_for` private method which is duplicated and used by the Updater and now the Operation classes. By making this method public, we can just call it from the Job, removing the need to duplicate the method as well as the `name_match?` helper.

Another source of duplicate is filtering the `job.ignore_conditions` by a given Dependency name and then hydrating them into a `Dependabot::Config::UpdateConfig` so they can be passed into core classes as `Dependabot::Config::IgnoreCondition` object. By preparing the `Dependabot::Config::UpdateConfig` within the `Dependabot::Job` once, we can use it to provide this functionality as required.

Finally, there's a duplicated log helper which needs knowledge internal to the Job that isn't otherwise required by the Updater anymore. I've moved it into the job for simplicity, but this isn't an ideal home for it. I've added a note on making this better in future but it requires a deeper change.

## Changes to Dependabot::Snapshot

Rather than repeat logic to filter the parsed dependencies by either those that are permitted by `allow_conditions` or those which intersect with the `job.dependencies` within the Updater, I've just moved this logic unto `Dependabot::Snapshot` and slimmed down the `dependencies` method(s) in the various classes.


updater/lib/dependabot/updater/operations/update_all_versions.rb
```diff
@@ -185,55 +168,11 @@ def file_updater_for(dependencies)
...
-            end
-          end
-        end
-
-        def name_match?(name1, name2)
-          WildcardMatcher.match?(
-            job.name_normaliser.call(name1),
-            job.name_normaliser.call(name2)
-          )
+          job.log_ignore_conditions_for(dependency)
```
landongrindheim, line 175: I see this is a refactor, where behavior shouldn't change at all. I'm still going to take the opportunity to ask about the code that's being updated 😇 

It feels 🤏 strange to me to have an operations class tell the job to log something. I wonder if in the longer term it would make sense for the job to return the ignore conditions in a format that can be logged. I have a feeling this is related to your comment in the job class, which I commented on.

landongrindheim (approved):

updater/lib/dependabot/updater/operations/update_all_versions.rb
brrygrdn: Yeah, this is definitely clunky, if it wasn't for the size of the method required to parse the 'raw' `ignore_conditions` I wouldn't mind it being duplicated.

I did also think about having a mixin that collects up all the common logging methods which might be an adequate middle ground but I punted on it as part of this as there's actually fewer shared logging methods that it seems.

My first preference would be that we effectively only have one notation of ignore conditions, as objects, available on the Job and then have a single log-presentation method but that's annoying non-trivial.

PR #6991, opened 2023-04-03
[Updater] Move filtering of parsed dependencies by allow rules or job dependencies

brrygrdn:
Before adding two more "Operation" classes for Security Updates, we should ensure that any code that is starting to be carried into every Operation class so far is de-duplicated as much as possible.

Rather than repeat logic to filter the parsed dependencies by either those that are permitted by allow_conditions or those which intersect with the job.dependencies within the Updater, I've just moved this logic unto `Dependabot::Snapshot` and slimmed down the dependencies method(s) in the various classes.

### DIFF
--- a/updater/lib/dependabot/dependency_snapshot.rb
+++ b/updater/lib/dependabot/dependency_snapshot.rb
@@ -27,6 +27,32 @@ module Dependabot
 
     attr_reader :base_commit_sha, :dependency_files, :dependencies
 
+    # Returns the subset of all project dependencies which are permitted
+    # by the project configuration.
+    def allowed_dependencies
+      @allowed_dependencies ||= dependencies.select { |d| job.allowed_update?(d) }
+    end
+
+    # Returns the subset of all project dependencies which are specifically
+    # requested to be updated by the job definition.
+    def job_dependencies
+      return [] unless job.dependencies&.any?
+      return @job_dependencies if defined? @job_dependencies
+
+      # Gradle, Maven and Nuget dependency names can be case-insensitive and
+      # the dependency name in the security advisory often doesn't match what
+      # users have specified in their manifest.
+      #
+      # It's technically possibly to publish case-sensitive npm packages to a
+      # private registry but shouldn't cause problems here as job.dependencies
+      # is set either from an existing PR rebase/recreate or a security
+      # advisory.
+      job_dependency_names = job.dependencies.map(&:downcase)
+      @job_dependencies = dependencies.select do |dep|
+        job_dependency_names.include?(dep.name.downcase)
+      end
+    end
+
     private
 
     def initialize(job:, base_commit_sha:, dependency_files:)
--- a/updater/lib/dependabot/job.rb
+++ b/updater/lib/dependabot/job.rb
@@ -134,6 +134,14 @@ module Dependabot
       @reject_external_code
     end
 
+    # TODO: Remove vulnerability checking
+    #
+    # This method does too much, let's make it focused on _just_ determining
+    # if the given dependency is within the configurations allowed_updates.
+    #
+    # The calling operation should be responsible for checking vulnerability
+    # separately, if required.
+    #
     # rubocop:disable Metrics/PerceivedComplexity
     def allowed_update?(dependency)
       allowed_updates.any? do |update|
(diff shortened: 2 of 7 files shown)


updater/lib/dependabot/job.rb
```diff
@@ -134,6 +134,14 @@ def reject_external_code?
       @reject_external_code
     end
 
+    # TODO: Remove vulnerability checking
+    #
+    # This method does too much, let's make it focused on _just_ determining
+    # if the given dependency is within the configurations allowed_updates.
+    #
+    # The calling operation should be responsible for checking vulnerability
+    # separately, if required.
```
landongrindheim, line 143: 👍

updater/lib/dependabot/dependency_snapshot.rb
```diff
@@ -27,6 +27,32 @@ def self.create_from_job_definition(job:, job_definition:)
...
+    # by the project configuration.
+    def allowed_dependencies
+      @allowed_dependencies ||= dependencies.select { |d| job.allowed_update?(d) }
+    end
+
+    # Returns the subset of all project dependencies which are specifically
+    # requested to be updated by the job definition.
+    def job_dependencies
+      return [] unless job.dependencies&.any?
+      return @job_dependencies if defined? @job_dependencies
```
landongrindheim, line 40: 💅 Not significant!

Are there cases where `job.dependencies` would change? It seems 🤏 surprising to have a guard clause that checks state before the `return @ivar if defined? @ivar` clause. I would typically expect that to be a part of the memoization.

updater/lib/dependabot/updater.rb
```diff
@@ -69,6 +69,7 @@ def run
         error_handler: error_handler
       ).perform
     rescue *ErrorHandler::RUN_HALTING_ERRORS.keys => e
+      # TODO: Drop this into Security-specific operations
```
landongrindheim, line 72: 🤤

updater/lib/dependabot/updater.rb
```diff
@@ -581,34 +582,19 @@ def existing_pull_request(updated_dependencies)
...
-        end
+      return dependency_snapshot.job_dependencies if job.dependencies
+
+      if dependency_snapshot.dependencies.any? && dependency_snapshot.allowed_dependencies.none?
+        Dependabot.logger.info("Found no dependencies to update after filtering allowed updates")
+        return []
       end
 
-      allowed_deps = all_deps.select { |d| job.allowed_update?(d) }
+      allowed_deps = dependency_snapshot.allowed_dependencies
```
landongrindheim, line 593: I'm loving the ergonomics of having `dependency_snapshot`! 👏

updater/lib/dependabot/updater.rb
```diff
@@ -581,34 +582,19 @@ def existing_pull_request(updated_dependencies)
...
       # Return dependencies in a random order, with top-level dependencies
       # considered first so that dependency runs which time out don't always hit
       # the same dependencies
-      allowed_deps = allowed_deps.shuffle unless ENV["UPDATER_DETERMINISTIC"]
-
-      if all_deps.any? && allowed_deps.none?
-        Dependabot.logger.info("Found no dependencies to update after filtering allowed " \
-                               "updates")
-      end
+      allowed_deps = allowed_deps.shuffle unless Environment.deterministic_updates?
```
landongrindheim, line 597: I dig the naming. It adds helpful context to a variable name that I had to think about 🙂

updater/lib/dependabot/updater/operations/refresh_version_update_pull_request.rb
```diff
@@ -44,18 +44,7 @@ def perform
...
-          #
-          # It's technically possibly to publish case-sensitive npm packages to a
-          # private registry but shouldn't cause problems here as job.dependencies
-          # is set either from an existing PR rebase/recreate or a security
-          # advisory.
-          job_dependencies = job.dependencies.map(&:downcase)
-          dependency_snapshot.dependencies.select do |dep|
-            job_dependencies.include?(dep.name.downcase)
-          end
+          dependency_snapshot.job_dependencies
```
landongrindheim, line 47: 💥

landongrindheim (approved):

PR #6996, opened 2023-04-04
Generate proper PR summary table for docker digest updates

deivid-rodriguez:
Before:

```
$ script/dependabot update docker dsp-testing/dependabot-core-5758
# ...
updater | 2023/04/04 13:11:52 INFO Results:
updater | +---------------------------------------------------------------------+
updater | |                 Changes to Dependabot Pull Requests                 |
updater | +---------+-----------------------------------------------------------+
updater | | created | docker/library/ruby ( from 2.7.6-buster to 3.2.2-buster ) |
updater | | created | ubuntu ( from focal to focal )                            |
updater | +---------+-----------------------------------------------------------+
proxy | time="2023-04-04T13:11:52Z" level=info msg="4/28 calls cached (14%)"
```

After:

```
$ script/dependabot update docker dsp-testing/dependabot-core-5758
# ...
updater | 2023/04/04 13:09:40 INFO Results:
updater | +---------------------------------------------------------------------+
updater | |                 Changes to Dependabot Pull Requests                 |
updater | +---------+-----------------------------------------------------------+
updater | | created | docker/library/ruby ( from 2.7.6-buster to 3.2.2-buster ) |
updater | | created | ubuntu ( from `450e066` to `24a0df4` )                    |
updater | +---------+-----------------------------------------------------------+
proxy | time="2023-04-04T13:09:40Z" level=info msg="4/28 calls cached (14%)"
```

The description is now consistent with PR titles and bodies.

### DIFF
--- a/common/lib/dependabot/dependency.rb
+++ b/common/lib/dependabot/dependency.rb
@@ -110,6 +110,65 @@ module Dependabot
       display_name_builder.call(name)
     end
 
+    def humanized_previous_version
+      # If we don't have a previous version, we *may* still be able to figure
+      # one out if a ref was provided and has been changed (in which case the
+      # previous ref was essentially the version).
+      if previous_version.nil?
+        return ref_changed? ? previous_ref : nil
+      end
+
+      if previous_version.match?(/^[0-9a-f]{40}$/)
+        return previous_ref if ref_changed? && previous_ref
+
+        "`#{previous_version[0..6]}`"
+      elsif version == previous_version &&
+            package_manager == "docker"
+        digest = docker_digest_from_reqs(previous_requirements)
+        "`#{digest.split(':').last[0..6]}`"
+      else
+        previous_version
+      end
+    end
+
+    def humanized_version
+      if version.match?(/^[0-9a-f]{40}$/)
+        return new_ref if ref_changed? && new_ref
+
+        "`#{version[0..6]}`"
+      elsif version == previous_version &&
+            package_manager == "docker"
+        digest = docker_digest_from_reqs(requirements)
+        "`#{digest.split(':').last[0..6]}`"
+      else
+        version
+      end
+    end
+
+    def docker_digest_from_reqs(requirements)
+      requirements.
+        filter_map { |r| r.dig(:source, "digest") || r.dig(:source, :digest) }.
+        first
+    end
+
+    def previous_ref
+      previous_refs = previous_requirements.filter_map do |r|
+        r.dig(:source, "ref") || r.dig(:source, :ref)
+      end.uniq
+      return previous_refs.first if previous_refs.count == 1
+    end
+
+    def new_ref
+      new_refs = requirements.filter_map do |r|
+        r.dig(:source, "ref") || r.dig(:source, :ref)
+      end.uniq
+      return new_refs.first if new_refs.count == 1
+    end
+
+    def ref_changed?
+      previous_ref != new_ref
+    end
+
     # Returns all detected versions of the dependency. Only ecosystems that
     # support this feature will return more than the current version.
     def all_versions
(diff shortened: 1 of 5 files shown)


common/lib/dependabot/dependency.rb
```diff
@@ -110,6 +110,67 @@ def display_name
...
+      if previous_version.nil?
+        return ref_changed? ? previous_ref : nil
+      end
+
+      if previous_version.match?(/^[0-9a-f]{40}$/)
+        return previous_ref if ref_changed? && previous_ref
+
+        "`#{previous_version[0..6]}`"
+      elsif version == previous_version &&
+            package_manager == "docker"
```
mctofu, line 126: I'm worried it's going to get pretty messy if we start to include package manager specific logic in this class. Is there another way to lookup or inject this logic?

Also, just a heads up that this class is also used in our backend service and we expect to be able to serialize it over the apis. I don't see anything that would break that currently but just something to be aware of!

deivid-rodriguez: > I'm worried it's going to get pretty messy if we start to include package manager specific logic in this class. Is there another way to lookup or inject this logic?

Another way would be to implement something like [display_name](https://github.com/dependabot/dependabot-core/blob/97722615459c6a08b83aa58226523aedbbe88f18/gradle/lib/dependabot/gradle.rb#L20-L27). I don't love that approach though.

mctofu: Ah, got it, didn't notice this check already existed in the common gem before. In that case I think it's OK to leave it in for now.

If we were to improve it we'll probably want to avoid a subclass as that will make the serialization more complicated. The display_name approach wouldn't have that same concern and would let us use code from the docker gem.

updater/spec/dependabot/integration_spec.rb
```diff
@@ -398,8 +398,8 @@
         expect(log_message).to include(
           "created",
           "dummy-git-dependency",
-          "from 20151f9b67c8a04461fa0ee28385b6187b86587b",
-          "to c0e25c2eb332122873f73acb3b61fb2e261cfd8f"
+          "from v1.0.0",
```
mctofu, line 401: Could you explain why this is populating versions now instead of shas? That doesn't match the example in the PR so I'm not following this part.

common/lib/dependabot/dependency.rb
```diff
@@ -110,6 +110,67 @@ def display_name
...
+    def humanized_previous_version
+      # If we don't have a previous version, we *may* still be able to figure
+      # one out if a ref was provided and has been changed (in which case the
+      # previous ref was essentially the version).
+      if previous_version.nil?
+        return ref_changed? ? previous_ref : nil
+      end
+
+      if previous_version.match?(/^[0-9a-f]{40}$/)
+        return previous_ref if ref_changed? && previous_ref
```
deivid-rodriguez, line 122: This line checks whether there's a reference available instead in the raw commit sha and use that if available (and has been changed).

updater/spec/dependabot/integration_spec.rb
deivid-rodriguez: Oh, sorry, I only found about this after creating the PR when I saw this spec failure in the updater specs.

This is just another nice side effect of using a "humanized version" instead of the "raw version". In the case of `git` dependencies, the new logic will checker whether there's a reference available instead in the raw commit sha and use that if available (and has been changed). It's this logic here: https://github.com/dependabot/dependabot-core/pull/6996/files#r1157787124.

deivid-rodriguez: As an example, if a Gemfile is being changed from 

```ruby
gem "foo", gitHub: "user/repo", tag: "v1.0.0"
```

to

```ruby
gem "foo", gitHub: "user/repo", tag: "v1.1.0"
```

we'll now use those in the summary instead of the locked commit SHAs.

mctofu (approved):
Thanks for the additional explanation! Nice improvement!

PR #6995, opened 2023-04-03
bump Cargo version

jakecoffman:
Bump rust, use the new [CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse](https://blog.rust-lang.org/2023/03/09/Rust-1.68.0.html#cargos-sparse-protocol) option.

PR #6999, opened 2023-04-04
No need to explicitly configure option now that it's the default

jeffwidman:
Blocked until Rust 1.70 lands, that's when they plan to change this (which we'll need to confirm).
--

We added this in Rust `1.68` via https://github.com/dependabot/dependabot-core/pull/6995 as we needed it earlier, but the `rust` project has since made this the default behavior.

So this removes the explicit configuration of the option.

### DIFF
--- a/cargo/Dockerfile
+++ b/cargo/Dockerfile
@@ -5,7 +5,6 @@ USER root
 # Install Rust
 ENV RUSTUP_HOME=/opt/rust \
   CARGO_HOME=/opt/rust \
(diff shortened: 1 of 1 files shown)


jakecoffman:
I think we're going to always want this set because in Rust you can specify what version of the toolchain you're pinned to with a `rust-toolchain` file. So if they select `channel="1.68.0"` we'd want to continue using the sparse protocol. https://rust-lang.github.io/rustup/overrides.html#channel

The toolchain we install in the Dockerfile only serves as the default if one isn't set explicitly.

PR #7009, opened 2023-04-05
[Updater] Inform the backend metric service which operation class is used for each update

brrygrdn:
In order to verify we can turn down the legacy code path soon, we need to make sure all production jobs are routing through an Operation class before we remove it.

This adds metric gathering to the service and API endpoint and sends a metric to the backend at the start of each update.


brrygrdn:
Smoke tests are failing as the cli doesn't know how to mock these new calls, this should be fixed by https://github.com/dependabot/cli/pull/93

updater/lib/dependabot/api_client.rb
```diff
@@ -137,6 +137,21 @@ def record_package_manager_version(ecosystem, package_managers)
...
+        data: {
+          metric: metric,
+          tags: tags
+        }
+      }
+      response = http_client.post(api_url, json: body)
+      # We treat metrics as fire-and-forget, so just warn if they fail.
+      Dependabot.logger.warn("Unable to report metric '#{metric}'.") if response.code >= 400
+    rescue HTTP::ConnectionError, OpenSSL::SSL::SSLError
+      Dependabot.logger.warn("Unable to report metric '#{metric}'.")
```
brrygrdn, line 152:
```suggestion
      Dependabot.logger.debug("Unable to report metric '#{metric}'.") if response.code >= 400
    rescue HTTP::ConnectionError, OpenSSL::SSL::SSLError
      Dependabot.logger.debug("Unable to report metric '#{metric}'.")
```

Per @mctofu's suggestion, let's use debug rather than warn so metric failures don't appear in the logs by default as they may end up being a red herring when debugging real issues.

landongrindheim (approved):

PR #7011, opened 2023-04-06
chore(ee): register custom docker filenames

yeikel:
This helps with code highlighting in Github.com

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,3 @@
 * text=auto
+Dockerfile.development linguist-language=Dockerfile
+Dockerfile.updater-core linguist-language=Dockerfile


yeikel:
This is a little strange. I am sure that yesterday it was not highlighting the syntax, but it is working today somehow. 

Closing...but it is weird

jeffwidman:
Thanks, and clever, I was unaware of this feature.

yeikel:
Github is acting for me 

@jeffwidman 

What do you see when you open the files in your browsers? Today...the syntax is not being highlighted for me. So maybe this is needed after all? 


image

yeikel:
Re-opening because I think that this is actually still needed

yeikel:
> This seems to be working randomly for me, and this change does not make a difference.
> 
> Reloading the page sometimes highlights properly, sometimes doesn't, regardless of this change.

How did you test this change to say that it is not working? Because as per the docs, it takes some time to apply it in a "low priority job"

yeikel:
Docs : https://github.com/github/linguist/blob/5a0c74277548122267d84283910abd5e0b89380e/docs/how-linguist-works.md?plain=1#L8

deivid-rodriguez:
Ah ok!

I reloaded [a permalink from current main](https://github.com/dependabot/dependabot-core/blob/ba77fd507513e91366eb214a58ea8d6a47098808/Dockerfile.development), and [a permalink from your commit](https://github.com/dependabot/dependabot-core/blob/0dc2fab8558a6e13e6773483a95d6728a32aa3a2/Dockerfile.development) several times, and I got inconsistent results in both cases.

If the file needs to be in main, then we can only try it! Let's do it!

jeffwidman:
Sorry for the slow reply here. I wonder if there's some internal bug/changing behavior within GitHub, because I'd expect to _just work_ w/o needing to call it out specifically.

Let me try asking internally within GitHub.

deivid-rodriguez:
Yeah, I don't see any changes here. Highlighting this file keeps happening randomly. I 👍 to revert this.

yeikel:
I do see more consistent rendering now. Is it still inconsistent for you? 

I reloaded the page > 20 times and in every single time it rendered as expected

yeikel:
> Sorry for the slow reply here. I wonder if there's some internal bug/changing behavior within GitHub, because I'd expect to _just work_ w/o needing to call it out specifically.
> 
> Let me try asking internally within GitHub.
> 
> If, as I suspect, this is supposed to be working without explicitly specifying, then I'd personally vote to remove this just to keep life simple...

I do not believe that this supposed to be working as is given that what is supported is a different pattern 

https://github.com/github/linguist/blob/5a0c74277548122267d84283910abd5e0b89380e/lib/linguist/languages.yml#L1528-L1538

If I understanding the document document above, we would need to rename the docker files so that they end in .dockerfile

For example Dockerfile.development => development.Dockerfile 

This issue might be relevant : https://github.com/github/linguist/issues/4566

I do not have enough information though, so please let me know what you find internally

Issue #6912, opened 2023-03-25
API rate limit exceeded - describe how to perform authenticated request


laeubi:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

I'm try to do some coding with dependabot and followed the instructions on the page, but performing the "dry run" instantly lets me run into "API rate limit exceeded"

I found an old issue, but that was about running dependabot in production and its quite vague:
- https://github.com/dependabot/dependabot-core/issues/3024

so I wonder if it would be possible to enhance the current readme with a section how one has to configure the development dev docker container with a user token e.g. 

- what permissions must be granted?
- where to configure the token so it is used by dependabot dry_run script?

jeffwidman:
Thanks @laeubi great questions.

I put up https://github.com/dependabot/dependabot-core/pull/7020 which addresses how to pass the token to the script.

As far as what permissions are needed, that depends on what repositories you're trying to access. For public repos, the base PAT with no additional permissions will suffice for getting a higher rate limit. For accessing private repositories, you'll need to ensure the PAT is authorized to read from those repositories.

Dry run doesn't actually create PR's, it emits the output it would use to create the PR, so no need for write permissions.

I thought ☝️ was a little more detail than needed in the README given that it just _works_ with the default perms, so for now only documented how to pass in the token, along with links to more docs on configuring PATs.

laeubi:
@jeffwidman thanks I think this already helps, and most of the time a public repository is fine for testing anyways :+1:

Issue #5758, opened 2022-09-20
Dependabot not ignoring major semver changes


lcooper01:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

Docker

### dependabot.yml content

```
version: 2
updates:
  - package-ecosystem: "docker" # See documentation for possible values
    directory: "/" # Location of package manifests
    schedule:
      interval: "daily"
    ignore:
      - dependency-name: "*"
        update-types: ["version-update:semver-major"]
```

### Updated dependency

```eclipse-temurin``` from ```11.0.14.1_1-jre-alpine``` to ```11.0.16.1_1-jre-alpine```

### What you expected to see, versus what you actually saw

Expected

```eclipse-temurin``` from ```11.0.14.1_1-jre-alpine``` to ```11.0.16.1_1-jre-alpine```

Actual

```eclipse-temurin``` from ```11.0.14.1_1-jre-alpine``` to ```17.0.4_8-jre-alpine```

### Images of the diff or a link to the PR, issue, or logs

```
  proxy | time="2022-09-20T15:23:11Z" level=info msg="proxy starting" commit=b031647dc5f52d8120800fc16337727989cb9be0
  proxy | 2022/09/20 15:23:11 Listening (:1080)
updater | 2022-09-20T15:23:11.959085124 [anonymous-instance:main:WARN:src/firecracker/src/main.rs:370] You are using a deprecated parameter: --seccomp-level 2, that will be removed in a future version.
updater | 2022-09-20T15:23:12.024873453 [465689105:main:WARN:src/devices/src/legacy/serial.rs:432] Detached the serial input due to peer close/error.
updater | time="2022-09-20T15:23:15Z" level=info msg="guest starting" commit=d97478b458e198f9b9a6cb546d902ee2e6651286
updater | time="2022-09-20T15:23:15Z" level=info msg="starting job..." fetcher_timeout=5m0s job_id=465689105 updater_timeout=45m0s updater_version=614fafdba436f058e32b92f1fc0a6ea940fe5c01
updater | I, [2022-09-20T15:23:18.755226 #7]  INFO -- sentry: ** [Raven] Raven 3.1.2 ready to catch errors
updater | To use retry middleware with Faraday v2.0+, install `faraday-retry` gem
updater | INFO <job_465689105> Starting job processing
  proxy | 2022/09/20 15:23:24 [002] GET https://api.github.com:443/repos/elcoops99/test-dependabot
[...]

PR #7020, opened 2023-04-07
Document some other useful options for the dry-run script

jeffwidman:
I use these options regularly when testing.

I'm not sure which are also applicable to the CLI, so I didn't document them, but if they are supported we should add mention of them either here or in the CLI readme.

Fix https://github.com/dependabot/dependabot-core/issues/6912

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ However, please do not leave comments that contribute nothing new to the discuss
 
 ### Don't file issues about Security Alerts or Dependency Graph
 
-The issue-tracker is meant solely for issues related to Dependabot's updating logic. Issues about security alerts or Dependency Graph should instead be filed as a [Code Security discussion](https://github.com/orgs/community/discussions/categories/code-security).
+The issue-tracker is meant solely for issues related to Dependabot's updating logic. Issues about [security alerts](https://docs.github.com/en/code-security/dependabot/dependabot-alerts/about-dependabot-alerts) or [Dependency Graph](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-the-dependency-graph) should instead be filed as a [Code Security discussion](https://github.com/orgs/community/discussions/categories/code-security).
 
 A good rule of thumb is that if you have questions about the _diff_ in a PR, it belongs here.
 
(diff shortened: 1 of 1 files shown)


deivid-rodriguez (approved):
Thank you!

Something else it may be worth documenting is how to deal with ignores, like here:

https://github.com/dependabot/dependabot-core/issues/5758#issuecomment-1278209859

jeffwidman:
> Something else it may be worth documenting is how to deal with ignores, like here:
> https://github.com/dependabot/dependabot-core/issues/5758#issuecomment-1278209859

Nice, I didn't even know we could do that! Will add.

README.md
```diff
@@ -208,22 +208,22 @@ $ bin/dry-run.rb bundler dependabot/demo --dir="/ruby"
...
+## CLI tool
 
 The [Dependabot CLI](https://github.com/dependabot/cli) is a newly-developed tool that incorporates the [GitHub
 Credentials Proxy](#private-registry-credential-management) to more realistically simulate what's happening within
 the Dependabot-at-GitHub service when talking to private registries.
 
 It has a [dedicated debugging guide](https://github.com/dependabot/cli#debugging-with-the-cli), including support for dropping into the Ruby debugger.
 
-### Dry-Run Script
+## Dry-Run Script
```
jeffwidman, line 226: I noticed in the preview these previously didn't _pop_ as much as I wanted when scrolling quickly through the document (we've got a long'ish `README`!) so I made the a little larger.

README.md
```diff
@@ -318,15 +339,26 @@ not synced to the development container. So you have two choices for editing the
...
+
+The env var `IGNORE_CONDITIONS` allows passing one or more ignore conditions to the to the [dry-run script](#dry-run-script):
+
+```bash
+IGNORE_CONDITIONS='[{"dependency-name":"*","update-types": ["version-update:semver-major"]}]' \
+bin/dry-run.rb docker test_org/test-dependabot`
+```
 
-### Debugging Security Alerts / Advisories
+### Debugging Security Updates
```
jeffwidman, line 354: I realized "Alerts / Advisories" was a misnomer because we're actually testing the security "update", which is triggered by the alerts, but not the alerts themselves. This is also more consistent with our stated policy that issues about alerts do not belong in this repo.

deivid-rodriguez (approved):
💯

Issue #7008, opened 2023-04-05
YAML file with dot in their name aren't detected by Dependabot::Docker::FileFetcher


negrel:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

Docker

### Package manager version

NA

### Language version

NA

### Manifest location and content before the Dependabot update

/deployment.abc.yaml

### dependabot.yml content

NA

### Updated dependency

NA

### What you expected to see, versus what you actually saw

FileFetcher.files doesn't contain `deployment.abc.yaml`

File and any file containing a dot doesn't match the following YAML regex:
```ruby
      YAML_REGEXP = /^[^\.]+\.ya?ml$/i.freeze
```
https://github.com/dependabot/dependabot-core/blob/HEAD/docker/lib/dependabot/docker/file_fetcher.rb#L11

### Native package manager behavior

NA

### Images of the diff or a link to the PR, issue, or logs

NA

### Smallest manifest that reproduces the issue

NA

jeffwidman:
Makes sense.

Want to submit a PR fixing it?

negrel:
Yes sure !

PR #7022, opened 2023-04-07
fix k8s files with dot in their name are ignored

negrel:
Related to #7008

### DIFF
--- a/docker/lib/dependabot/docker/file_fetcher.rb
+++ b/docker/lib/dependabot/docker/file_fetcher.rb
@@ -8,7 +8,7 @@ require "dependabot/file_fetchers/base"
 module Dependabot
   module Docker
     class FileFetcher < Dependabot::FileFetchers::Base
-      YAML_REGEXP = /^[^\.]+\.ya?ml$/i
+      YAML_REGEXP = /\.ya?ml$/i
       DOCKER_REGEXP = /dockerfile/i
 
       def self.required_files_in?(filenames)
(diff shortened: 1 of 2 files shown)


landongrindheim (approved):

jeffwidman:
Could you add a test to ensure we don't regress in the future?

If you're not sure how let me know and I can find an example when I'm at my computer (on my phone right now)

negrel:
Yes, it depends on the type of test you want. An example would be great!

jeffwidman:
I suspect it's as simple as adding some additional test cases this array?

https://github.com/dependabot/dependabot-core/blob/a4556cf9956dc765930a75b707d8d877d9e5e215/docker/spec/dependabot/docker/file_fetcher_spec.rb#L272-L286

jeffwidman (changes requested):
Marking as requested changes so it's clear form PR overview page that this is blocked on needing the test (which looks straightforward)

jeffwidman:
@negrel @mctofu should this be closed now that https://github.com/dependabot/dependabot-core/pull/8862 is merged?

Looking at the diff, it looks like that PR has a superset of what's tested here.

Nishnha:
> @negrel @mctofu should this be closed now that #8862 is merged?
> 
> Looking at the diff, it looks like that PR has a superset of what's tested here.

Looks like it. Thanks for the heads up!

negrel:
LGTM

PR #7023, opened 2023-04-07
Fix reinstalling `npm_and_yarn` native helper dependencies

deivid-rodriguez:
During image build, this works because `dry-run` is not yet set in configuration. But if you re-run it inside the development container after, for example, changing the version of a native dependencies, it no longer installs anything.

Avoid `dry-run` mode explicitly to prevent this.


jeffwidman (approved):
Oof, nice catch!

PR #7024, opened 2023-04-07
Fix update checking in Yarn repositories using workspaces

deivid-rodriguez:
There were two issues:

* We were actually running `npm` to find the version to update to, instead of running `yarn` in this kind of repository. Not surprised that fails many times 😅.
* In addition to that, the error message for peer dependency conflicts when using workspaces is slightly different and our previous regex did not catch it.

The former issue was happening for both Yarn classic and Yarn Berry monorepos, while the latter affected Yarn classic only.


npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -519,7 +521,7 @@ def run_yarn_berry_checker(path:, version:)
           end
         end
 
-        def run_yarn_checker(path:, version:)
+        def run_yarn_classic_checker(path:, version:)
```
jeffwidman, line 530: Is `yarn_classic` Yarn v1? If so, a code comment might be helpful context for those less familiar with the ecosystem...

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -486,8 +486,14 @@ def lockfiles_for_path(lockfiles:, path:)
         def run_checker(path:, version:)
```
jeffwidman, line 486: I'm actually kind of surprised there were no tests to catch this?

Not by peering into the internals, but if this is failing a bunch in prod currently, then maybe could add a unit test that mimics that by running `run_checker` on yarn monorepo and then ensuring it works rather than throwing that error we're seeing?

If this is really hard to write, it may not be worth doing, but if it's straightforward might be worth tacking on?

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -33,7 +33,7 @@ class VersionResolver
         # peer dependency "react@>=0.14.0 <16.0.0".
```
jeffwidman, line 33: It looks like this code comment documents example error strings... does it need updating as well? I'm unclear if this is a new error string or a slight tweak on an existing one...

jeffwidman (approved):
Nice catch!

A few small comments, I trust your judgement on whether they need addressing vs ship as-is.

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
deivid-rodriguez: Yes, it is. Since Yarn v3 is referred to as yarn berry, I thought I'd use yarn classic here. And also because I'm not sure where yarn v2 falls, or whether it exists at all 😆. I'm still figuring out yarn versioning but I read "classic" a few times when meaning "yarn before berry", that's why I used it here because it seemed a common terminology for Yarn people.

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
deivid-rodriguez: I was struggling how to write a spec for this, I suspect because `npm install` actually succeeds on many yarn repos, it just ignores the yarn lockfile. I guess there are some cases where it does not, but I couldn't easily find an easy case.

Actually while trying to write a spec for this, I discovered the second issue fixed here, for which a test is easier to write.

So I figured this was good with only a test for the latter fix.

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
deivid-rodriguez: Yes, let me update it!

deivid-rodriguez: Updated!

jeffwidman: Oh that is so much better, I hadn't even realized the `\` was intended to signal multi-line break. 😁 

❤️

jeffwidman (approved):

Issue #1697, opened 2020-02-19
Support for .net core local tool files


xt0rted:
In newer versions of .net core global tools can be installed locally to a project. I'm not sure where the documentation is for this, it seems to be missing from the site right now.

The tools are installed from nuget.org and are saved to a `.config/dotnet-tools.json` file in the root of your project. This is the file format:

```json
{
  "version": 1,
  "isRoot": true,
  "tools": {
    "dotnet-format": {
      "version": "3.1.37601",
      "commands": [
        "dotnet-format"
      ]
    }
  }
}
```

xt0rted:
The documentation for this is now available at https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools#install-a-local-tool

ejball:
Would love to see this implemented. `dotnet-tools.json` also works at the root of the project (i.e. not in `.config`).

xt0rted:
Now that I'm using local tools more I put some action workflows together that look for local tools and update them if needed. A demo of this can be found in [xt0rted/dotnet-tool-update-test](https://github.com/xt0rted/dotnet-tool-update-test).

PR #7028, opened 2023-04-09
Support for nuget tools in .config/dotnet-tools.json

mburumaxwell:
Local dotnet tools such as [`dotnet-ef`](https://learn.microsoft.com/en-us/ef/core/cli/dotnet) are distributed [as NuGet packages](https://www.nuget.org/packages/dotnet-ef/). The tools and their versions are stored in `.config/dotnet-tools.json` as decribed in the [docs](https://learn.microsoft.com/en-us/dotnet/core/tools/local-tools-how-to-use).

This PR adds support for updating them using the existing infrastructure for NuGet. Basically similar to the support for `global.json` but works around directory nesting and multiline regex match for the JSON.

Fixes: #1697

### DIFF
--- a/nuget/lib/dependabot/nuget/file_updater.rb
+++ b/nuget/lib/dependabot/nuget/file_updater.rb
@@ -15,6 +15,7 @@ module Dependabot
           %r{^[^/]*\.[a-z]{2}proj$},
           /^packages\.config$/i,
           /^global\.json$/i,
+          /^dotnet-tools\.json$/i,
           /^Directory\.Build\.props$/i,
           /^Directory\.Build\.targets$/i,
           /^Packages\.props$/i
@@ -56,6 +57,10 @@ module Dependabot
 
       def global_json
         dependency_files.find { |f| f.name.casecmp("global.json").zero? }
+      end
+
+      def dotnet_tools_json
+        dependency_files.find { |f| f.name.casecmp(".config/dotnet-tools.json").zero? }
       end
 
       def check_required_files
@@ -126,6 +131,13 @@ module Dependabot
             global_json.content.match(
               /"#{Regexp.escape(dependency.name)}"\s*:\s*
                "#{Regexp.escape(dependency.previous_version)}"/x
+            ).to_s
+          ]
+        elsif requirement.fetch(:file).casecmp(".config/dotnet-tools.json").zero?
+          [
+            dotnet_tools_json.content.match(
+              /"#{Regexp.escape(dependency.name)}"\s*:\s*{\s*"version"\s*:\s*
+               "#{Regexp.escape(dependency.previous_version)}"/xm
             ).to_s
           ]
         else
--- a/nuget/spec/dependabot/nuget/file_updater_spec.rb
+++ b/nuget/spec/dependabot/nuget/file_updater_spec.rb
@@ -361,5 +361,105 @@ RSpec.describe Dependabot::Nuget::FileUpdater do
         end
       end
     end
+
+    context "with a dotnet-tools.json" do
+      let(:dependency_files) { [csproj_file, dotnet_tools_json] }
+      let(:dotnet_tools_json) do
+        Dependabot::DependencyFile.new(
+          content: fixture("dotnet_tools_jsons", "dotnet-tools.json"),
+          name: ".config/dotnet-tools.json"
+        )
+      end
+      let(:dependency) do
+        Dependabot::Dependency.new(
+          name: "dotnetsay",
+          version: "2.1.7",
+          previous_version: "1.0.0",
+          requirements: [{
+            file: ".config/dotnet-tools.json",
+            requirement: "2.1.7",
+            groups: ["dependencies"],
+            source: nil
+          }],
+          previous_requirements: [{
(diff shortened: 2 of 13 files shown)


mburumaxwell:
@jeffwidman @deivid-rodriguez 
Does this kind of change need to follow specific guidelines?

nuget/lib/dependabot/nuget/file_updater.rb
```diff
@@ -128,6 +133,13 @@ def original_declarations(dependency, requirement)
                "#{Regexp.escape(dependency.previous_version)}"/x
             ).to_s
           ]
+        elsif requirement.fetch(:file).casecmp(".config/dotnet-tools.json").zero?
+          [
+            dotnet_tools_json.content.match(
+              /"#{Regexp.escape(dependency.name)}"(?:.*):(?:.*){(?:.*)"version"(?:.*)
```
deivid-rodriguez, line 139: I think this may be matching too much, because `.*` is greedy by default?

Like, in the example `dotnet-tools.json` file added as a test case, if both `botsay` and `dotnetsay` are locked at the same version, they would both get updates when updating the first dependency in the file (`botsay`)?

mburumaxwell: The first double quotation would match one and both the other. E.g. "dotnetsay" : { "version" : "1.0.0" } would not conflict with "botsay" : { "version" : "1.0.0" }

As for the greedy match, i wouldn't get it to work with `\s` or `\S`. Do you have a suggestion?

deivid-rodriguez: This is what I mean: 
https://rubular.com/r/xD4D1x1SlqDRkP. This regex would match both dependencies, and then when replacing I'm afraid it would update the versions of both. Does that make sense?

I would indeed try to use `\s`. Maybe something like

```suggestion
              /"#{Regexp.escape(dependency.name)}"(?:\s*):(?:\s*){(?:[^}]*)"version"(?:\s*):(?:\s*)
```

Also, can these be quoted using single quotes?

mburumaxwell: Thanks for the insights here. I will make the RegEx change shortly. It seems to solve the conflict problem in the process? https://rubular.com/r/8owRHxVYfwXM4y

deivid-rodriguez: If you can add an extra spec to cover the tricky case, that'd be awesome!

mburumaxwell: Done.

deivid-rodriguez: Thanks! Let me have a look!

nuget/spec/dependabot/nuget/file_updater_spec.rb
```diff
@@ -361,5 +361,53 @@
...
+      let(:dependency_files) { [csproj_file, dotnet_tools_json] }
+      let(:dotnet_tools_json) do
+        Dependabot::DependencyFile.new(
+          content: fixture("dotnet_tools_jsons", "dotnet-tools.json"),
+          name: ".config/dotnet-tools.json"
+        )
+      end
+      let(:dependency) do
+        Dependabot::Dependency.new(
+          name: "dotnetsay",
```
deivid-rodriguez, line 375: Since we're updating the last dependency in the file, I don't think this catches the conflict we discuss, we'd need to update the _first_ dependency in the file so that the conflict could happen.

deivid-rodriguez (commented):
Just a nitpick on the modified spec, other than that looks good to me!

nuget/spec/dependabot/nuget/file_updater_spec.rb
mburumaxwell: I changed the test files to make both tools have the same version the confirm in the spec that one one is updated and not both. Did I miss somthing?

deivid-rodriguez: If you revert the changes in the regex, I think the current spec would still pass.

Because the match starts from `"#{dependency.name}"` all the way to `"#{dependency.previous_version}"` and the problem of accidentally matching versions of two dependencies because the version of the second dependency version is the same as the first one can only happen when updating the first dependency in the file (`botsnay`, not `dotnetsay`).

Let me know if I didn't explain this properly, regexps are hard 😅.

deivid-rodriguez: Looks good to me now!

deivid-rodriguez (approved):
👏 👏

mburumaxwell:
@deivid-rodriguez anything else remaining?

deivid-rodriguez:
No, sorry for dropping the ball here. Merging!

alexrp:
This PR seems to lack support for `dotnet-tools.json` in the root of a repo. There isn't actually any requirement that the file must be under `.config`; `dotnet tool restore` will look for it in the current directory too.

deivid-rodriguez:
Happy to accept a PR to also look at the current directory!

PR #7030, opened 2023-04-10
fix(npm): registry inferring should include the full registry path

### LINKED ISSUE #6804: [NPM] Registry inferring ignores the registry resource path

### Package ecosystem

npm

### Package manager version

_No response_

### Language version

6

_No response_

### dependabot.yml content

I am using dependabot core directly

### Updated dependency

_No response_

When a dependency is declared in a private registry such as : 

```json
"cors": {
			"version": "2.8.5",
			"resolved": "https://myRegistry/api/npm/npm/cors/-/cors-2.8.5.tgz",
			"integrity": "sha1-6sEdpRWS3Ya58G9uesKTs9+HXSk=",
			"requires": {
				"object-assign": "^4",
				"vary": "^1"
			}
		},
```

The registry detection detects the registry as `https://myRegistry/` when it should detect the full URI `https://myRegistry/api/npm/npm/`

That results in 404 errors when the updater tries to query `https://myRegistry/` instead of the expected `https://myRegistry/api/npm/npm/`

Likely offending code: 

https://github.com/dependabot/dependabot-core/blob/30f9e404c832253c2da6f61d73f6c51a6d96984e/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb#L104

One solution for now, is to explicitly declare the registry using `.npmrc`

_No response_

{"error":"404 Not Found - GET https://myRegistry/cors"}

Run the registry detection using a private registry like this: 

```json
"cors": {
			"version": "2.8.5",
			"resolved": "https://myRegistry/api/npm/npm/cors/-/cors-2.8.5.tgz",
			"integrity": "sha1-6sEdpRWS3Ya58G9uesKTs9+HXSk=",
			"requires": {
				"object-assign": "^4",
				"vary": "^1"
			}
		}
```

yeikel:
Fixes #6804

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
@@ -10,7 +10,7 @@ require "dependabot/npm_and_yarn/file_parser/lockfile_parser"
 
 module Dependabot
   module NpmAndYarn
-    class FileFetcher < Dependabot::FileFetchers::Base
+    class FileFetcher < Dependabot::FileFetchers::Base # rubocop:disable Metrics/ClassLength
       require_relative "file_fetcher/path_dependency_builder"
 
       # Npm always prefixes file paths in the lockfile "version" with "file:"
@@ -22,6 +22,7 @@ module Dependabot
       # "yarn link", e.g. "link:react"
       PATH_DEPENDENCY_STARTS = %w(file: link:. link:/ link:~/ / ./ ../ ~/).freeze
       PATH_DEPENDENCY_CLEAN_REGEX = /^file:|^link:/
+      DEFAULT_NPM_REGISTRY = "https://registry.npmjs.org"
 
       def self.required_files_in?(filenames)
         filenames.include?("package.json")
@@ -86,25 +87,39 @@ module Dependabot
 
       # If every entry in the lockfile uses the same registry, we can infer
       # that there is a global .npmrc file, so add it here as if it were in the repo.
-      def inferred_npmrc
+
+      def inferred_npmrc # rubocop:disable Metrics/PerceivedComplexity
         return @inferred_npmrc if defined?(@inferred_npmrc)
         return @inferred_npmrc = nil unless npmrc.nil? && package_lock
 
         known_registries = []
-        JSON.parse(package_lock.content).fetch("dependencies", {}).each do |_name, details|
-          resolved = details.fetch("resolved", "https://registry.npmjs.org")
+        JSON.parse(package_lock.content).fetch("dependencies", {}).each do |dependency_name, details|
+          resolved = details.fetch("resolved", DEFAULT_NPM_REGISTRY)
+
           begin
             uri = URI.parse(resolved)
           rescue URI::InvalidURIError
             # Ignoring non-URIs since they're not registries.
-            # This can happen if resolved is false, for instance.
             next
           end
-          # Check for scheme since path dependencies will not have one
-          known_registries << "#{uri.scheme}://#{uri.host}" if uri.scheme && uri.host
+
+          next unless uri.scheme && uri.host
+
+          known_registry = "#{uri.scheme}://#{uri.host}"
+          path = uri.path
+
+          next unless path
+
+          index = path.index(dependency_name)
+          if index
+            registry_base_path = path[0...index].delete_suffix("/")
+            known_registry << registry_base_path
+          end
+
+          known_registries << known_registry
         end
 
-        if known_registries.uniq.length == 1 && known_registries.first != "https://registry.npmjs.org"
+        if known_registries.uniq.length == 1 && known_registries.first != DEFAULT_NPM_REGISTRY
           Dependabot.logger.info("Inferred global NPM registry is: #{known_registries.first}")
           return @inferred_npmrc = Dependabot::DependencyFile.new(
             name: ".npmrc",
--- a/npm_and_yarn/spec/dependabot/npm_and_yarn/file_fetcher_spec.rb
+++ b/npm_and_yarn/spec/dependabot/npm_and_yarn/file_fetcher_spec.rb
@@ -1787,7 +1787,36 @@ RSpec.describe Dependabot::NpmAndYarn::FileFetcher do
       expect(file_fetcher_instance.files.map(&:name)).
         to eq(%w(package.json package-lock.json .npmrc))
       expect(file_fetcher_instance.files.find { |f| f.name == ".npmrc" }.content).
-        to eq("registry=https://npm.fury.io")
+        to eq("registry=https://npm.fury.io/dependabot")
+    end
+  end
+
+  context "with no .npmrc but package-lock.json contains a artifactory repository" do
+    before do
+      allow(file_fetcher_instance).to receive(:commit).and_return("sha")
+
+      stub_request(:get, File.join(url, "package.json?ref=sha")).
+        with(headers: { "Authorization" => "token token" }).
+        to_return(
+          status: 200,
+          body: fixture_to_response("projects/npm6/private_artifactory_repository", "package.json"),
+          headers: json_header
+        )
+
+      stub_request(:get, File.join(url, "package-lock.json?ref=sha")).
+        with(headers: { "Authorization" => "token token" }).
+        to_return(
+          status: 200,
+          body: fixture_to_response("projects/npm6/private_artifactory_repository", "package-lock.json"),
+          headers: json_header
+        )
+    end
+
+    it "infers an npmrc file" do
+      expect(file_fetcher_instance.files.map(&:name)).
+        to eq(%w(package.json package-lock.json .npmrc))
+      expect(file_fetcher_instance.files.find { |f| f.name == ".npmrc" }.content).
+        to eq("registry=https://myRegistry/api/npm/npm")
     end
   end
 end
--- a/npm_and_yarn/spec/fixtures/projects/npm6/private_artifactory_repository/package-lock.json
+++ b/npm_and_yarn/spec/fixtures/projects/npm6/private_artifactory_repository/package-lock.json
@@ -1,0 +1,175 @@
+{
+  "name": "private_artifactory_repository",
+  "version": "1.0.0",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
+    "": {
+      "version": "1.0.0",
+      "dependencies": {
+        "fetch-factory": "0.0.1"
+      },
+      "devDependencies": {
+        "etag": "1.0.0"
+      }
+    },
+    "node_modules/encoding": {
+      "version": "0.1.13",
+      "resolved": "https://myRegistry/api/npm/npm/encoding/-/encoding-0.1.13.tgz",
+      "integrity": "sha512-ETBauow1T35Y/WZMkio9jiM0Z5xjHHmJ4XmjZOq1l/dXz3lr2sRn87nJy20RupqSh1F2m3HHPSp8ShIPQJrJ3A==",
+      "dependencies": {
+        "iconv-lite": "^0.6.2"
+      }
+    },
+    "node_modules/es6-promise": {
+      "version": "3.3.1",
+      "resolved": "https://myRegistry/api/npm/npm/es6-promise/-/es6-promise-3.3.1.tgz",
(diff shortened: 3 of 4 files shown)


npm_and_yarn/spec/dependabot/npm_and_yarn/file_fetcher_spec.rb
```diff
@@ -1787,7 +1787,37 @@
       expect(file_fetcher_instance.files.map(&:name)).
         to eq(%w(package.json package-lock.json .npmrc))
       expect(file_fetcher_instance.files.find { |f| f.name == ".npmrc" }.content).
-        to eq("registry=https://npm.fury.io")
+        to eq("registry=https://npm.fury.io/dependabot/")
```
yeikel, line 1790: I don't think that this test was correct

yeikel: I created a test org to double check this 

image

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -85,25 +86,25 @@
...
-          begin
-            uri = URI.parse(resolved)
-          rescue URI::InvalidURIError
-            # Ignoring non-URIs since they're not registries.
-            # This can happen if resolved is false, for instance.
-            next
+        JSON.parse(package_lock.content).fetch("dependencies", {}).each do |dependency_name, details|
+          resolved = details.fetch("resolved", DEFAULT_NPM_REGISTRY)
+
+          next if !resolved.is_a?(String) || resolved.start_with?(DEFAULT_NPM_REGISTRY)
```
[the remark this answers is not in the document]
yeikel: I understand this rule, but I do not think it can be exploited? 

I can remove the check and it should still work though. I can push that if needed. I think think that it'd be wasteful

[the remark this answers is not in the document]
yeikel: I pushed [Remove check](https://github.com/dependabot/dependabot-core/pull/7030/commits/0904dfa793c061450b4cf3bd0a0c69e659b9a934)

npm_and_yarn/spec/dependabot/npm_and_yarn/file_fetcher_spec.rb
deivid-rodriguez: Thanks for trying this in real life! 💪

deivid-rodriguez (commented):
Nice fix!

I do agree that the issue is probably not exploited, but I think we can probably keep the previous approach of properly parsing URIs to avoid it. I can suggest something tomorrow.

npm_and_yarn/spec/fixtures/projects/npm6/private_artifactory_repository/package-lock.json
```diff
@@ -0,0 +1,75 @@
+{
+  "name": "test",
+  "version": "1.0.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+      "encoding": {
+          "version": "0.1.12",
+          "resolved": "https://myRegistry/api/npm/npm/cors/-/cors-2.8.5.tgz",
```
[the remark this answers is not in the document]
yeikel: Great catch! 

No. This was the result of copying an existing example. I re-generated the lockfile and update the PR

Thanks!

deivid-rodriguez:
I think it's fine in CRuby. But more importantly, I think it's best to focus on the bug fix rather than mixing in performance related patches. If the problem you mention is an issue, we could handle it separately.

Let me suggest this alternative patch that keeps the existing approach of parsing registries into well structured URIs:

```diff
diff --git a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
index 4ef88467c..10f75aa33 100644
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
@@ -90,7 +90,7 @@ module Dependabot
         return @inferred_npmrc = nil unless npmrc.nil? && package_lock
 
         known_registries = []
-        JSON.parse(package_lock.content).fetch("dependencies", {}).each do |_name, details|
+        JSON.parse(package_lock.content).fetch("dependencies", {}).each do |dependency_name, details|
           resolved = details.fetch("resolved", "https://registry.npmjs.org")
           begin
             uri = URI.parse(resolved)
@@ -99,8 +99,24 @@ module Dependabot
             # This can happen if resolved is false, for instance.
             next
           end
+
+          next unless uri.scheme && uri.host
+
           # Check for scheme since path dependencies will not have one
-          known_registries << "#{uri.scheme}://#{uri.host}" if uri.scheme && uri.host
+          known_registry = "#{uri.scheme}://#{uri.host}"
+
+          path = uri.path
+          if path
+            index = path.index(dependency_name)
+
+            if index
+              registry_base_path = path[0...index].delete_suffix("/")
+
+              known_registry << registry_base_path
+            end
+          end
+
+          known_registries << known_registry
         end
 
         if known_registries.uniq.length == 1 && known_registries.first != "https://registry.npmjs.org"
```

Using this approach is how I discovered yesterday that the test lockfile was incorrect, because the tests you added do not pass with this patch due to the dependency name vs registry url discrepancy.

npm_and_yarn/spec/dependabot/npm_and_yarn/file_fetcher_spec.rb
```diff
@@ -1787,7 +1787,37 @@
...
+        with(headers: { "Authorization" => "token token" }).
+        to_return(
+          status: 200,
+          body: fixture_to_response("projects/npm6/private_artifactory_repository", "package-lock.json"),
+          headers: json_header
+        )
+    end
+
+    it "infers an npmrc file" do
+      expect(file_fetcher_instance.files.count).to eq(3)
```
deivid-rodriguez, line 1816: Can we remove this assertion? Since we are checking for an exact array in the next line, if the array was incorrect, it's best to let the exact check fail since it will give more helpful information, rather than just the incorrect size of the array.

yeikel:
> I think it's fine in CRuby. But more importantly, I think it's best to focus on the bug fix rather than mixing in performance related patches. If the problem you mention is an issue, we could handle it separately.
> 
> Let me suggest this alternative patch that keeps the existing approach of parsing registries into well structured URIs:
> 
> ```diff
> diff --git a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
> index 4ef88467c..10f75aa33 100644
> --- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
> +++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
> @@ -90,7 +90,7 @@ module Dependabot
>          return @inferred_npmrc = nil unless npmrc.nil? && package_lock
>  
>          known_registries = []
> -        JSON.parse(package_lock.content).fetch("dependencies", {}).each do |_name, details|
> +        JSON.parse(package_lock.content).fetch("dependencies", {}).each do |dependency_name, details|
>            resolved = details.fetch("resolved", "https://registry.npmjs.org")
>            begin
>              uri = URI.parse(resolved)
> @@ -99,8 +99,24 @@ module Dependabot
>              # This can happen if resolved is false, for instance.
>              next
>            end
> +
> +          next unless uri.scheme && uri.host
> +
>            # Check for scheme since path dependencies will not have one
> -          known_registries << "#{uri.scheme}://#{uri.host}" if uri.scheme && uri.host
> +          known_registry = "#{uri.scheme}://#{uri.host}"
> +
> +          path = uri.path
> +          if path
> +            index = path.index(dependency_name)
> +
> +            if index
> +              registry_base_path = path[0...index].delete_suffix("/")
> +
> +              known_registry << registry_base_path
> +            end
> +          end
> +
> +          known_registries << known_registry
>          end
>  
>          if known_registries.uniq.length == 1 && known_registries.first != "https://registry.npmjs.org"
> ```
> 
> Using this approach is how I discovered yesterday that the test lockfile was incorrect, because the tests you added do not pass with this patch due to the dependency name vs registry url discrepancy.

Thank you for your feedback. I applied most of it but still left the check I added about the String check. There is no need to try parsing a URI that we know in advance that will always fail

I also had to disable `Metrics/PerceivedComplexity`  and `Metrics/PerceivedComplexity` because the class is now too large and the method a little complicated as per the rules. Let me know if you have any concerns about that

Thanks!

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -85,25 +86,43 @@ def fetch_files
...
         return @inferred_npmrc = nil unless npmrc.nil? && package_lock
 
         known_registries = []
-        JSON.parse(package_lock.content).fetch("dependencies", {}).each do |_name, details|
-          resolved = details.fetch("resolved", "https://registry.npmjs.org")
+        JSON.parse(package_lock.content).fetch("dependencies", {}).each do |dependency_name, details|
+          resolved = details.fetch("resolved", DEFAULT_NPM_REGISTRY)
+
+          # skips values that are not strings. For example, when it is `false` as it will always fail to parse
+          next unless resolved.is_a?(String)
```
yeikel, line 99: This is not needed given the rescue below, but I think that it is a good addition

We could add a log in the rescue clause in the future. As I'd be interested to hear in what situations the parsing fails when the input is a string

yeikel:
@deivid-rodriguez Let me know if you have any additional feedback :)

Thanks again

yeikel:
@jeffwidman I would appreciate your input here as well

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
deivid-rodriguez: The `false` case, we do need it since I think it's due to an [upstream npm bug](https://github.com/npm/cli/issues/1138). But we should be able to remove it once we drop npm v6 support.

The other cases I guess they're there to handle corrupted lockfiles including malformed URIs. Not sure if that's worth supporting though.

In any case, all this seems unrelated to this fix so I'd prefer if we handled it separately.

So, my preference would be to leave this as is and let the diff in this PR and commit focus on the bug being fixed. You can extract this change to a separate PR and we can figure it out separately.

What do you think?

deivid-rodriguez:
Sorry for the delay @yeikel, this looks good to me, but I'd like to extract the malformed URI handling changes/discussion to a separate PR/issue if that's ok.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
yeikel: That's fair. I updated the pull request removing that check. I'll send a follow up PR later today for the specific check

Thanks!

deivid-rodriguez:
@yeikel This looks ready from my side and I'm pretty confident about it since you have taken the time to actually try this on a real registry, but I'm not super familiar with how private registries are setup in general, so let me ask for a second review here from another teammate just in case. Thanks!

pavera (approved):
LGTM

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
jeffwidman: I am behind on my notifications due to a head cold followed by food poisoning, so am not spending time to dig into all the details here, but @deivid-rodriguez wherever this npm6 workaround is, if it's not already let's try to make sure it's commented as such in the code? That way when we drop support for it, we can simply grep the codebase for "npm 6" to easily find these cases to delete them? Because if they're only discoverable via `git blame`, they'll get overlooked during the upgrade...

deivid-rodriguez: Thanks for noticing @jeffwidman, I see now that this PR has actually removed the `# This can happen if resolved is false, for instance.` comment. That's my oversight since I was not expecting that comment to be removed and planned to handle it in the follow up PR including the change that I asked to be removed.

@yeikel Can you create a follow up PR to restore that comment and also link to the upstream issue that forces us to special case that?

yeikel: I will create one PR later today

Thanks!

yeikel: Submitted https://github.com/dependabot/dependabot-core/pull/7106

PR #7042, opened 2023-04-10
Fix Python CI

deivid-rodriguez:
We've need to update this spec twice so far, and will have to keep doing this every time blinker releases new versions.

I fixed the spec permanently this time by only checking for presence of the dependency.


dwc0011:
PR #7040 is failing because this Specs / ci (python, python, 2, 0, pip) (pull_request)  which is unrelated to anything in the PR.


Failed examples:

rspec ./spec/dependabot/python/file_updater/pipfile_file_updater_spec.rb:300 # Dependabot::Python::FileUpdater::PipfileFileUpdater#updated_dependency_files the updated Pipfile.lock with a subdependency from an extra doesn't remove the subdependency

deivid-rodriguez:
Yes, I was just opening a PR with a dummy change to make sure this was not fault on any PR, but just CI failing on our main branch.

I fixed the issue now!

jeffwidman (approved):

PR #7044, opened 2023-04-10
Label PRs by the top level folder they update

deivid-rodriguez:


.github/labeler.yml
```diff
@@ -0,0 +1,47 @@
+"L: ruby:bundler":
+  - bundler
```
landongrindheim, line 2: All the examples in the [actions/labeler README](https://github.com/actions/labeler#basic-examples) include wildcard matchers. This is my first encounter with this Action. Do we need to use `bundler/*`?

landongrindheim (approved):
I left a question around how configuration will be honored, but I see no reason to not give this a try 😄

.github/labeler.yml
deivid-rodriguez: You're right, it's my first encounter with this action too!

I don't think this is going to work as is, but maybe it does and the labeler README needs better docs?

I'll do some tests to figure it out.

deivid-rodriguez: I wasn't sure how to test this without merging it. And I'm pretty sure the folder only approach was not going to work, so I changed this to use a proper glob.

yeikel: For future references, you can always test this(and any other automation) using a fork. That'd be less destructible in case there is any problem as you can recreate the fork anytime

deivid-rodriguez: Thanks for the tip. I tried it on my fork and it worked as expected!

deivid-rodriguez:
Let's try this!

yeikel:
It seems to be failing in some scenarios. Example https://github.com/dependabot/dependabot-core/actions/runs/4695403461/jobs/8324540326?pr=7030

deivid-rodriguez:
Ugg, my bad, typo! Thanks for noticing!

yeikel:
> Ugg, my bad, typo! Thanks for noticing!

It's working now :) 

https://github.com/dependabot/dependabot-core/actions/runs/4700405646/jobs/8335038927?pr=7030

Issue #6938, opened 2023-03-29
Use core `go` lib/tooling for identify package metadata URLs instead of the `Masterminds/VCS` fork


jeffwidman:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code improvement description

Today we fetch metadata information using `github.com/Masterminds/vcs` library:
https://github.com/dependabot/dependabot-core/blob/642f064cc2126996cfaa31f61255e49a451da6cf/go_modules/helpers/importresolver/main.go#L14-L33

However, that is a copy/paste of some code in core `go`, so we'd rather get direct access to the actual source. [Further context](https://github.com/dependabot/dependabot-core/issues/4448#issuecomment-1247317691):

> I think the code is trying to map a Go import path back to a repository so we can e.g. link to the repo in the Dependabot PR, find the CHANGELOG, link to the diff between versions, etc, etc.
> It’s not totally trivial with Go because:
> 1. they’ve got that whole vanity URL thing going on, where you have to pass `go-get=1` in to make it act like a fancy redirect to a repo
> 2. a Go import path doesn’t necessarily actually match a repo URL. For instance, I might be able to import https://github.com/go-kit/kit/tracing, but that’s not a valid URL on GitHub. The repo is [github.com/go-kit/kit](http://github.com/go-kit/kit), and the package is browsable at https://github.com/go-kit/kit/tree/master/tracing. So Go bakes in some [logic](https://github.com/golang/go/blob/95c125a44ad1a0c3e441c3214160cd7b4483e79c/src/cmd/go/internal/vcs/vcs.go#L1437) to handle that for a series of SCMs, and the Masterminds/vcs package does a [similar thing](https://github.com/Masterminds/vcs/blob/master/vcs_remote_lookup.go) too.
> 
> Ideally, we’d just use whatever golang/go does, but IIRC that wasn’t importable so the next best thing was the Masterminds/vcs implementation. The Ruby code you linked to predates our usage of that Go library, and I’d guess is just a really incomplete Ruby implementation of the same thing. 
> Perhaps these days a) there’s a good Ruby implementation, or b) we think the VCS list is stable enough we’d be happy to translate it to Ruby, or c) we could somehow codegen the Ruby from the canonical Go implementation? Or we just stick with that one-off Go helper?

I dug into this a little, and there's already an upstream request:
* https://github.com/golang/go/issues/18387

Where someone [noted](https://github.com/golang/go/issues/18387#issuecomment-277001616): 
[...]

PR #7045, opened 2023-04-10
Replace custom `go` helper with `go list`

jeffwidman:
Over time, we've managed to thin out our custom `go` helper in favor of native package manager tooling via `go list`, `go get`, and `go mod`.

The final remaining usage was for translating package import URLs into the actual repository URL for fetching metadata.

The original impetus for the `github.com/Masterminds/vcs` library was that it was a copy/paste of some code in core `go`:

 [Further context](https://github.com/dependabot/dependabot-core/issues/4448#issuecomment-1247317691):

> I think the code is trying to map a Go import path back to a repository so we can e.g. link to the repo in the Dependabot PR, find the CHANGELOG, link to the diff between versions, etc, etc.
> It’s not totally trivial with Go because:
> 1. they’ve got that whole vanity URL thing going on, where you have to pass `go-get=1` in to make it act like a fancy redirect to a repo
> 2. a Go import path doesn’t necessarily actually match a repo URL. For instance, I might be able to import https://github.com/go-kit/kit/tracing, but that’s not a valid URL on GitHub. The repo is [github.com/go-kit/kit](http://github.com/go-kit/kit), and the package is browsable at https://github.com/go-kit/kit/tree/master/tracing. So Go bakes in some [logic](https://github.com/golang/go/blob/95c125a44ad1a0c3e441c3214160cd7b4483e79c/src/cmd/go/internal/vcs/vcs.go#L1437) to handle that for a series of SCMs, and the Masterminds/vcs package does a [similar thing](https://github.com/Masterminds/vcs/blob/master/vcs_remote_lookup.go) too.
>
> Ideally, we’d just use whatever golang/go does, but IIRC that wasn’t importable so the next best thing was the Masterminds/vcs implementation. The Ruby code you linked to predates our usage of that Go library, and I’d guess is just a really incomplete Ruby implementation of the same thing.
> Perhaps these days a) there’s a good Ruby implementation, or b) we think the VCS list is stable enough we’d be happy to translate it to Ruby, or c) we could somehow codegen the Ruby from the canonical Go implementation? Or we just stick with that one-off Go helper?

However, I discovered after doing some digging in https://github.com/dependabot/dependabot-core/issues/6938 that there _is both_ an upstream copy (https://github.com/golang/go/issues/18387#issuecomment-277001616, which may potentially get deprecated https://github.com/golang/go/issues/57051) and an actual `go list` command that provides enough of what we need:
* https://github.com/golang/go/issues/44742

```
$ go list -m -f '{{.Origin}}' golang.org/x/tools@latest
{git https://go.googlesource.com/tools    refs/tags/v0.3.0 502c634771c4ba335286d55fc24eeded1704f592 }
```

While this doesn't fully resolve the upstream issue:
* https://github.com/golang/go/issues/18387

For our purposes it provides enough of what we need since we already have the module path extracted from `go.mod` (https://github.com/golang/go/issues/18387#issuecomment-1492664354).

Performing this switch allows us to completely delete the native go helper altogether.

I debated leaving the framework in place in case we later wish to extend it for other features, but decided for now to rip it out as it's easy enough to add back later. And in general we're trying to move toward leveraging native package manager tooling whenever possible. So if we later needed an additional feature, we're probably start with a discussion with the `go` team to see if they'd build it into the `go` cmd tooling.

Fix https://github.com/dependabot/dependabot-core/issues/6938


go_modules/spec/dependabot/go_modules/metadata_finder_spec.rb
```diff
@@ -63,4 +63,46 @@
       it { is_expected.to eq("https://github.com/satori/go.uuid") }
     end
   end
+
+  describe "#look_up_source_using_go_list" do
```
jeffwidman, line 67: These tests are a direct move via copy/paste from go_modules/spec/dependabot/go_modules/path_converter_spec.rb

go_modules/lib/dependabot/go_modules/update_checker.rb
```diff
@@ -4,7 +4,6 @@
 require "dependabot/update_checkers/base"
 require "dependabot/shared_helpers"
 require "dependabot/errors"
-require "dependabot/go_modules/native_helpers"
```
jeffwidman, line 7: I suspect this is an unused import... probably never cleaned up when we previously thinned out other commands from the native helpers...

go_modules/lib/dependabot/go_modules/file_updater/go_mod_updater.rb
```diff
@@ -4,7 +4,6 @@
 require "dependabot/errors"
 require "dependabot/logger"
 require "dependabot/go_modules/file_updater"
-require "dependabot/go_modules/native_helpers"
```
jeffwidman, line 7: I suspect this is an unused import... probably never cleaned up when we previously thinned out other commands from the native helpers...

jeffwidman:
Thanks for pointing that out @jakecoffman ... I'd been only working on this locally so didn't even notice the smoke tests were failing!

I poked around the interwebs a bit, but this one looks tricky.

jeffwidman:
Given that the `go` team declined my feature request to expose the `go-source` meta tag content (https://github.com/golang/go/issues/59566), I don't see a path forward here unfortunately.

PR #7055, opened 2023-04-12
chore(ee): ignore `.idea` folder

yeikel:
I use [IDEA products ](https://www.jetbrains.com/products/) , and I would appreciate if this is merged as currently I need to ensure that I don't commit it accidentally with each pull request

Creating a separate pull request for this as suggested by this comment: 

https://github.com/dependabot/dependabot-core/pull/7029#discussion_r1163093575

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,5 @@ coverage/
 .tool-versions
 .rspec_status
 .rdbg_history
+# Ignore Jetbrains IntelliJ project-config directory
+.idea/


.gitignore
```diff
@@ -30,3 +30,4 @@ coverage/
 .tool-versions
 .rspec_status
 .rdbg_history
+.idea/
```
jeffwidman, line 33: Can you add a comment of what this is intended to ignore? Something like

```
# ignore Jetbrains IntelliJ project-config directory
.idea/
```

That way folks who don't use IDEA have some context w/o having to run `git blame` on it

yeikel: Thanks. I updated it

deivid-rodriguez (approved):
Thanks!

PR #7066, opened 2023-04-12
cargo: Fix sparse registry error on update

jakecoffman:
We're seeing some errors like this:

```
error: failed to get `arrow` as a dependency of package `geoengine-datatypes v0.7.0 (/home/dependabot/dependabot-updater/dependabot_tmp_dir/datatypes)`
 
Caused by: 
 failed to load source for dependency `arrow`
 
Caused by:
 Unable to update registry `crates-io`
 
Caused by:
  usage of sparse registries requires `-Z sparse-registry`
```

This is due to the sparse registry we enabled in https://github.com/dependabot/dependabot-core/pull/6995. It seems this setting is required for the update to work as well as the environment variable.

### DIFF
--- a/cargo/lib/dependabot/cargo/file_updater/lockfile_updater.rb
+++ b/cargo/lib/dependabot/cargo/file_updater/lockfile_updater.rb
@@ -32,7 +32,7 @@ module Dependabot
             SharedHelpers.with_git_configured(credentials: credentials) do
               # Shell out to Cargo, which handles everything for us, and does
               # so without doing an install (so it's fast).
-              run_shell_command("cargo update -p #{dependency_spec}", fingerprint: "cargo update -p <dependency_spec>")
+              run_shell_command("cargo -Z sparse-registry update -p #{dependency_spec}", fingerprint: "cargo update -p <dependency_spec>")
             end
 
             updated_lockfile = File.read("Cargo.lock")
--- a/cargo/lib/dependabot/cargo/file_updater/lockfile_updater.rb
+++ b/cargo/lib/dependabot/cargo/file_updater/lockfile_updater.rb
@@ -6,6 +6,7 @@ require "dependabot/git_commit_checker"
 require "dependabot/cargo/file_updater"
 require "dependabot/cargo/file_updater/manifest_updater"
 require "dependabot/cargo/file_parser"
+require "dependabot/cargo/toolchain_parser"
 require "dependabot/shared_helpers"
 module Dependabot
   module Cargo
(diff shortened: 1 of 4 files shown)


jakecoffman:
This isn't a complete solution, we also need to consider a range of `unstable` channel releases. But this should improve our success rate with Cargo updates, so we can ship incrementally.

The `-Z` flags only apply to unstable channels so we can ignore older stable toolchains. They will continue to use the non-sparse protocol.

jeffwidman:
How frequently is this problem occurring?

Given it's only cropping up on nightly/unstable, and anyone using nightly/unstable generally is quick to update their toolchains, I suspect most folks affected by this will be updating soon if they haven't already...

I'm not objecting to merging since this is already written, but any code we add is also code we have to maintain later... so I'd be inclined to simultaneously open a draft PR to remove this workaround, and consider it blocked for another 6-12 months and then pull it out.

This also gets into the territory that we've discussed for later this quarter of setting clear policy guidelines around what versions we support within different ecosystems (cc @abdulapopoola for visibility as we discuss that initiative). Eg, for Rust Cargo, in general we probably don't want to be trying to support workarounds required for nightly/unstable for very long once a feature has been merged into stable.

Two other ideas:
1. don't try to use the sparse protocol at all for nightly/unstable that are older than the stable version where the sparse protocol landed... this has a perf hit for us internally, but again I suspect the frequency is pretty darn low so it'd be okay.
2. throw an error message to the user of "you are on an older nightly/unstable, please upgrade to a stable >= <version where sparse landed in stable>"

Any reason we can't go with 1️⃣ ??

Also, my apologies if I misunderstand completely how channels work in Cargo/Rust, I've read the docs but haven't worked directly with them so I may be misunderstanding something here that limits the implementation options.

jakecoffman:
The errors are hidden behind the HelperSubprocessFailed exception, so we can't tell exactly how many updates this is affecting.

I created a PR to gather toolchain channel data that should be helpful going forward: https://github.com/dependabot/dependabot-core/pull/7080

Consider that this feature only stabilized in January, so anyone who pinned to a nightly last year are now broken. I don't want to disable the sparse protocol and risk an incident unless we know more about who is using what channels. So, do we want to wait on this and proceed with https://github.com/dependabot/dependabot-core/pull/7080?

jakecoffman:
Initial numbers indicate <7% of our jobs involve a toolchain, and even less of that are toolchains before 1.68, maybe ~1%? 

Based on that I think we can go with recategorizing the error as a resolution error and recommend upgrading toolchains to >= 1.68.

jeffwidman:
@ilya-bobyr Yep, see https://github.com/dependabot/dependabot-core/pull/7122.

Is there a reason you can't update your toolchain to `1.68.0`?

ilya-bobyr:
We saw failures in some dynamically loaded code.
And no one spend enough time fixing it yet.

jakecoffman:
Sorry for the trouble! Dependabot was putting too much pressure on GitHub without the sparse registry so it forced us to update. Usually we're not this aggressive about forcing recent versions.

ilya-bobyr:
> Sorry for the trouble! Dependabot was putting too much pressure on GitHub without the sparse registry so it forced us to update. Usually we're not this aggressive about forcing recent versions.

No worries, and thank you for the explanation!
After you merge the warning, it would be even easier for the users to understand that it is time to update :))

PR #7070, opened 2023-04-12
Bundle the updater when bumping versions

brrygrdn:
This updates the bin/bump-versions.rb script to run `bundle` in the Updater directory to ensure the Gemfile.lock picks up the new version as part of the staged changes.

### Notes

This does assume the host machine has ruby/bundler installed and is ok bundling the updater gems locally but on the upside it verifies the bundling goes as expected before the changes leave the machine.


jakecoffman (approved):

deivid-rodriguez:
Alternatively `bundle lock` (just brings lockfile is sync but does not install gems) may work?

brrygrdn:
> Alternatively bundle lock (just brings lockfile is sync but does not install gems) may work?

Aha, yes, that's much better, I always forget that exists!

bin/bump-version.rb
```diff
@@ -75,6 +75,9 @@ def proposed_changes(version, _new_version)
   File.write(version_path, new_version_contents)
   puts "☑️  common/lib/dependabot.rb updated"
 
+  # Bump the updater's Gemfile with the new version
```
deivid-rodriguez, line 78:
```suggestion
  # Bump the updater's Gemfile.lock with the new version
```

deivid-rodriguez (approved):
Nice!

PR #7074, opened 2023-04-13
GroupRule -> DependencyGroup

Nishnha:
Rename GroupRule to DependencyGroup

PR #7075, opened 2023-04-13
Run Dependency Group updates

Nishnha:
This PR is stacked on top of https://github.com/dependabot/dependabot-core/pull/7074

I have tried to make each commit represent a single file change for this PR.

Things to note:

The main contribution of this PR is that it introduces a new class, the `DependencyGroupEngine`, which registers `DependencyGroups` based on values passed in from an Update Job. 

The engine is not called directly, but dependency groups are made available to the Updater through 2 methods:
- `DependencySnapshot#groups` 
- `DependencySnapshot#ungrouped_dependencies`

Here is what the job summary for 2 dependency groups looks like:
```
+------------------------------------------------------------------------------------------------------------------------------------+
|                                                Changes to Dependabot Pull Requests                                                 |
+---------+--------------------------------------------------------------------------------------------------------------------------+
| created | @emotion/react ( from 11.10.5 to 11.10.6 ), @emotion/styled ( from 11.10.5 to 11.10.6 )                                  |
| created | gatsby ( from 5.6.0 to 5.8.1 ), gatsby-cli ( from 5.6.0 to 5.8.0 ), gatsby-plugin-emotion ( from 8.6.0 to 8.8.0 ), ga... |
+---------+--------------------------------------------------------------------------------------------------------------------------+
```

This implementation will only run dependencies that belong to a dependency group. It does not try to generate any PRs for ungrouped dependencies.


updater/lib/dependabot/updater/operations/group_update_all_versions.rb
```diff
@@ -37,44 +35,48 @@ def initialize(service:, job:, dependency_snapshot:, error_handler:)
...
-                    :service,
-                    :dependency_snapshot,
-                    :error_handler,
-                    :dependency_group
+          # Let's not worry about this for now, but eventually we'd try to do single updates
+          # for anything that didn't appear in _at least one_ group.
+          dependency_snapshot.ungrouped_dependencies.each do |dependency|
+            # dependency_change = create_dependency_change_for(dependency)
+            # create_pull_request(dependency_change)
+          end
```
brrygrdn, line 77:
```suggestion
          # FIXME: Let's not worry about this for now but eventually we'd try to do single updates
          # for anything that didn't appear in _at least one_ group.
          # dependency_snapshot.ungrouped_dependencies.each do |dependency|
          #   dependency_change = create_dependency_change_for(dependency)
          #   create_pull_request(dependency_change)
          # end
```

We might as well comment out the entire code block to avoid running the empty loop, I've been using `FIXME` as a search anchor for any pseduocode or notes I've been dropping in to make these things easy to find again.

updater/lib/dependabot/dependency_snapshot.rb
```diff
@@ -53,6 +53,18 @@ def job_dependencies
       end
     end
 
+    # A dependency snapshot will always have the same set of dependencies since it only depends
+    # on the Job and dependency groups, which are static for a given commit.
+    def groups
+      # The DependencyGroupEngine registers dependencies when the Job is created
+      # and it will memoize the dependency groups
+      Dependabot::DependencyGroupEngine.dependency_groups(allowed_dependencies)
```
brrygrdn, line 61:
```suggestion
      @groups ||= Dependabot::DependencyGroupEngine.dependency_groups(allowed_dependencies)
```

Should we memoize this to avoid naively recalculating it?

updater/lib/dependabot/dependency_snapshot.rb
```diff
@@ -53,6 +53,18 @@ def job_dependencies
...
+    end
+
+    def ungrouped_dependencies
+      Dependabot::DependencyGroupEngine.ungrouped_dependencies(allowed_dependencies)
```
brrygrdn, line 65:
```suggestion
      @ ungrouped_dependencies ||= Dependabot::DependencyGroupEngine.ungrouped_dependencies(allowed_dependencies)
```

Same as above

updater/lib/dependabot/job.rb
```diff
@@ -233,6 +238,18 @@ def security_advisories_for(dependency)
...
+        Dependabot::DependencyGroupEngine.register(group["name"], group["rules"])
+      end
+    end
+
+    def belongs_to_dependency_group?(dependency)
+      return false unless dependency_groups.any?
+
+      Dependabot::DependencyGroupEngine.groups_for(dependency).any?
+    end
+
```
brrygrdn, line 252: It doesn't look like we use this method?

```suggestion
```

common/spec/dependabot/dependency_group_spec.rb
```diff
@@ -1,14 +1,105 @@
...
+      expect(dependency_group.rules).to eq(rules)
+    end
+  end
+
+  describe "#dependencies" do
+    context "when calculate_dependency_groups! has not been run" do
+      it "returns an empty list" do
+        expect(dependency_group.dependencies).to eq([])
+      end
+    end
```
brrygrdn, line 62: Not blocking, I think we can avoid drawing in the language of the `DependencyGroupEngine` class and just test this was 
```suggestion
    context "when no dependencies are assigned to the group" do
      it "returns an empty list" do
        expect(dependency_group.dependencies).to eq([])
      end
    end
```

common/spec/dependabot/dependency_group_spec.rb
```diff
@@ -1,14 +1,105 @@
...
+
+    context "after calculate_dependency_groups! has been run" do
```
brrygrdn, line 64:
```suggestion
    context "when dependencies have been assigned" do
```

updater/lib/dependabot/dependency_group_engine.rb
```diff
@@ -0,0 +1,83 @@
...
+# Its methods should only be called with Dependabot::DependencyGroupEngine
+#
+# Groups are only calculated once after the Job has registered its dependencies
+# All allowed dependencies should be passed in to the calculate_dependency_groups! method
+#
+# **Note:** This is currently an experimental feature which is not supported
+#           in the service or as an integration point.
+#
+module Dependabot
+  module DependencyGroupEngine
```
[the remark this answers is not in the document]
brrygrdn: That said, I think we can leave this as-is since it is working for this PR and refactor it in a follow-up, overall how this works is nice and clear 👍🏻

updater/lib/dependabot/dependency_group_engine.rb
```diff
@@ -0,0 +1,83 @@
...
+        @ungrouped_dependencies << dependency if groups.empty?
+
+        groups.each do |group|
+          group.dependencies.push(dependency)
+          @dependency_groups[group.name.to_sym] = group
+        end
+      end
+
+      true
+    end
```
brrygrdn, line 81: This remaining methods would effectively just become instance methods I think?

updater/lib/dependabot/updater/operations/group_update_all_versions.rb
```diff
@@ -37,44 +35,48 @@ def initialize(service:, job:, dependency_snapshot:, error_handler:)
...
-              raise if ErrorHandler::RUN_HALTING_ERRORS.keys.any? { |err| e.is_a?(err) }
-
-              error_handler.handle_dependabot_error(error: e, dependency: group_dependency)
+          dependency_snapshot.groups.each do |_group_hash, group|
+            Dependabot.logger.info("[Experimental] Starting grouped update job for #{job.source.repo}")
+            # We should log the rule being executed, let's just hard-code wildcard for now
+            # since the prototype makes best-effort to do everything in one pass.
+            # This should be replaced with the actual Dependabot::DependencyGroup instances that have names
+            # dependency_snapshot.groups will need to be updated to return a list of
+            # Dependabot::DependencyGroup instances
```
brrygrdn, line 48:
```suggestion
```

brrygrdn (commented):
This looks great! I really like the split out responsibilities between `DependencyGroup` and the `DependencyGroupEngine`.

I have a few minor notes, but my longer thought on making the `DependencyGroupEngine` an instance rather than a 'singleton-like' class is probably worth doing separately. Ruby class variables have a lot of gotchas so while they work it's generally not the preferred idiom and I think we get a side benefit that when we start to do more rule validation we can throw any parsing errors at a nice and clear point in the code later.

updater/lib/dependabot/dependency_snapshot.rb
brrygrdn: Disregard this, we should only call this once and its effectively hitting a singleton class so memoization isn't really useful.

updater/lib/dependabot/dependency_snapshot.rb
brrygrdn: Disregard this, we should only call this once and its effectively hitting a singleton class so memoization isn't really useful.

updater/lib/dependabot/dependency_change.rb
```diff
@@ -13,7 +13,7 @@ module Dependabot
   class DependencyChange
     attr_reader :job, :updated_dependencies, :updated_dependency_files
 
-    def initialize(job:, dependencies:, updated_dependency_files:, dependency_group: nil)
+    def initialize(job:, updated_dependencies:, updated_dependency_files:, dependency_group: nil)
```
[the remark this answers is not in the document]
brrygrdn: I meant to check back in on this, but looking at [main](https://github.com/dependabot/dependabot-core/blob/main/updater/lib/dependabot/dependency_change.rb#L16-L21): https://github.com/dependabot/dependabot-core/blob/9a721d9f1923227a1219ef33f030e4541f080588/updater/lib/dependabot/dependency_change.rb#L16-L21

It seems this change is already in master, so there's maybe something weird with the diff calculation?

updater/spec/dependabot/dependency_change_builder_spec.rb
```diff
@@ -105,9 +105,9 @@
       end
     end
 
-    context "when the source is a group rule" do
+    context "when the source is a dependency group" do
       let(:change_source) do
-        Dependabot::GroupRule.new(name: "dummy-pkg-*")
+        Dependabot::DependencyGroup.new(name: "dummy-pkg-*", rules: ["dummy-pkg-*"])
```
brrygrdn, line 110: I've had to refer back to the API, and I believe the actual data structure we get here would be more like:

```suggestion
        Dependabot::DependencyGroup.new(name: "dummy-pkg-*", rules: { "patterns" => "dummy-pkg-*"]})
```

i.e the configuration file would have the form:

```yaml
version: 2
updates:
  - package-ecosystem: "bundler"
    directory: "/"
    schedule:
      interval: "daily"
    groups:
      - dummy-pkg-*:
          patterns:
            - "dummy-pkg-*"
```

We landed on this after a few passed in order to allow for rules which used things other than pattern-based matching as we iterate.

For purposes of this iteration in Core/Updater, I think it is fine to keep our code-as is and just strip out the value of `group["rules]["patterns"]` as the only thing we care about right now.

Nishnha: Thanks I actually got that [wrong](https://github.com/dependabot/cli/pull/94) in the CLI too. I'll propose a fix there and make an adjustment here

Nishnha:
This PR is in kind of a weird state because I had it stacked on top of https://github.com/dependabot/dependabot-core/pull/7074/commits which I then squashed into main.

The [merge commit](https://github.com/dependabot/dependabot-core/pull/7075/commits/62319d6e46265010a3ec87629d63e4be9c3c816d) brings this branch back up to date with `main`, but there are a couple of commits, namely ff792bcbb07c4fc116fe9a5625943153ded95048 and e8a2b4b0f8deb2e1a2a367fa6e120cd0a33461a9, that were also present in the stacked PR that I squashed.

So those 2 commits are kind of unnecessary? If you diff them with main they look (mostly) empty.

Anyhow, I'm going to continue rolling forward with commits this branch. If we want to have a cleaner history I'd be happy to re-open this branch as a new PR but I don't want to rebase and lose @brrygrdn's comments here.

updater/lib/dependabot/dependency_group_engine.rb
```diff
@@ -0,0 +1,83 @@
...
+# groups that are defined by users in their dependabot config file.
+#
+# Each UpdateJob registers its own DependencyGroupEngine which calculates
+# the grouped and ungrouped dependencies for a DependencySnapshot
+#
+# This is a static class tied to the lifecycle of a Job
+# Its methods should only be called with Dependabot::DependencyGroupEngine
+#
+# Groups are only calculated once after the Job has registered its dependencies
+# All allowed dependencies should be passed in to the calculate_dependency_groups! method
```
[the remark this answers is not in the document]
Nishnha: Thanks, those bullet points read easier to me too.

For "Its methods should only be called with Dependabot::DependencyGroupEngine" my original intention was to signal that this class follows a singleton pattern (for now) and can only be accessed statically.

I think it's fine to leave that line out though since I plan on refactoring this class to one that can be instantiated.

updater/lib/dependabot/dependency_change.rb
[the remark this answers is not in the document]
Nishnha: I think that was an artifact from this originally being a stacked PR, and the PR it was stacked on top of was out-of-date with main. 

It does look fixed now, somehow 🤷🏾

Nishnha:
Based on some of the comments from @brrygrdn, after this PR lands I'm going to follow up with a refactor of the DependencyGroupEngine with the goals being:

- make DependencyGroupEngine a [class](https://github.com/dependabot/dependabot-core/pull/7075#discussion_r1165384710) that is instantiated by a DependencySnapshot
- attempt a more [performant](https://github.com/dependabot/dependabot-core/pull/7075#discussion_r1165394446) check over dependency group rules
- [handle errors](https://github.com/dependabot/dependabot-core/pull/7075#pullrequestreview-1383161229) that result during dependency group rules checks

common/lib/dependabot/dependency_group.rb
```diff
@@ -2,10 +2,30 @@
...
-    def initialize(name)
+    def initialize(name, rule)
       @name = name
+      @rules = rule
+      @dependencies = []
     end
+
+    def contains?(dependency)
+      @dependencies.include?(dependency) if @dependencies.any?
+      rules.any? { |rule| WildcardMatcher.match?(rule, dependency.name) }
```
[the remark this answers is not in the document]
Nishnha: Thanks I'll take a look at this in a follow up PR
re: https://github.com/dependabot/dependabot-core/pull/7075#issuecomment-1507571882

updater/spec/dependabot/dependency_change_builder_spec.rb
Nishnha: Adjusted in 9ab1df5a11c9a74d48d22e5b73eeb9df1c002f2f

Nishnha: Fixed for the CLI in https://github.com/dependabot/cli/pull/95

common/lib/dependabot/dependency_group.rb
```diff
@@ -1,11 +1,20 @@
...
   class DependencyGroup
-    attr_reader :name
+    attr_reader :name, :rules, :dependencies
 
-    def initialize(name)
+    def initialize(name, rule)
       @name = name
+      @rules = rule
+      @dependencies = []
+    end
```
brrygrdn, line 13:
```suggestion
    def initialize(name, rules)
      @name = name
      @rules = rules
      @dependencies = []
    end
```

brrygrdn: I was just double-checking we were using positional arguments here, I'm going to update this to use keyword arguments in a PR I'd like to get through to the API layer so we don't have to change the contract later, but it's likely to conflict with this.

I'll link shortly.

brrygrdn: CRF: https://github.com/dependabot/dependabot-core/pull/7084

We are inconsistently using the group name as a keyword arg in some places, so let's consolidate on using keyword arguments as we start to upstream this code into the service.

Nishnha: fixed in 99a5c63423f5c15282a2414554538958b829bb2c

updater/lib/dependabot/updater/operations/group_update_all_versions.rb
```diff
@@ -36,43 +34,47 @@ def initialize(service:, job:, dependency_snapshot:, error_handler:)
...
-              #
-              #        We could use all_updated_deps.first, but that could be misleading. It may
-              #        make more sense to handle the dependency group as a Dependancy-ish object
-              group_dependency = OpenStruct.new(name: "group-all")
-              raise if ErrorHandler::RUN_HALTING_ERRORS.keys.any? { |err| e.is_a?(err) }
-
-              error_handler.handle_dependabot_error(error: e, dependency: group_dependency)
+          # FIXME: This preserves the default behavior of grouping all updates into a single PR
+          # but we should figure out if this is the default behavior we want.
+          register_all_dependencies_group if job.dependency_groups.empty?
```
brrygrdn, line 43: The smoke test seems to be failing on a method missing, we can tolerate nil by flipping the logic to do an `unless &.any?`

```suggestion
          register_all_dependencies_group unless job.dependency_groups&.any?
```

I think this is probably work keeping for now to avoid any co-ordination problems between this PR, the API and smoke tests 👍🏻

Nishnha:
I see this [test failure](https://github.com/dependabot/dependabot-core/actions/runs/4724151172/jobs/8381011696?pr=7075) in the updater spec but I can't recreate it locally ☹️ 

```ruby
Failures:

  1) Dependabot::Updater#run with the grouped experiment enabled updates multiple dependencies in a single PR correctly
     Failure/Error:
       expect(service).to receive(:create_pull_request) do |dependency_change, base_commit_sha|
         expect(dependency_change.updated_dependencies.first).to have_attributes(name: "dummy-pkg-b")
         expect(dependency_change.updated_dependency_files_hash).to eql(
           [
             {
               "name" => "Gemfile",
               "content" => fixture("bundler/updated/Gemfile"),
               "directory" => "/",
               "type" => "file",
               "mode" => "100644",

       (#<Dependabot::Service:0x00007ff524fb8d28 @client=#<InstanceDouble(Dependabot::ApiClient) (anonymous)>, @pull_requests=[], @errors=[]>).create_pull_request(*(any args))
           expected: 1 time with any arguments
           received: 2 times with any arguments
     # ./spec/dependabot/updater_spec.rb:2206:in `block (3 levels) in <top (required)>'
     # ./vendor/ruby/3.1.0/gems/webmock-3.18.1/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'
```

I was thinking of changing the test to be `expect(service).to receive(:create_pull_request).at_least(:once) do ...` but I'm not sure why `:create_pull_request` is getting called more than once in the first place?

updater/lib/dependabot/updater/operations/group_update_all_versions.rb
```diff
@@ -83,20 +85,35 @@ def dependencies
...
+        attr_reader :job,
+                    :service,
+                    :dependency_snapshot,
+                    :error_handler
+
+        def register_all_dependencies_group
+          all_dependencies_group = { "name" => "group-all", "rules" => { "patterns" => ["*"] } }
+          Dependabot::DependencyGroupEngine.register(all_dependencies_group["name"],
+                                                     all_dependencies_group["rules"]["patterns"])
+        end
```
brrygrdn, line 99: 👍🏻

updater/spec/dependabot/dependency_change_builder_spec.rb
```diff
@@ -107,6 +107,9 @@
 
     context "when the source is a dependency group" do
       let(:change_source) do
+        # FIXME: rules are actually a hash but for the purposes of this pass we can leave it as a list
+        # Once this is refactored we should create a DependencyGroup like so
+        # Dependabot::DependencyGroup.new(name: "dummy-pkg-*", rules: { "patterns" => ["dummy-pkg-*"] })
```
brrygrdn, line 112: 👍🏻

brrygrdn:
> I was thinking of changing the test to be expect(service).to receive(:create_pull_request).at_least(:once) do ... but I'm not sure why :create_pull_request is getting called more than once in the first place?

Yeah, that's odd - it seems like it should only be called once. It might be worth just adding a `puts` statement to throw out the dependency_group name/identity at the start of the `Service#create_pull_request` method just in case there is a mystery guest in the test setup?

I'm thinking that due to the `DependencyGroupEngine` being a singleton, it might mean that rules declared in other tests are leaking when ran in CI as part of a full suite? We could also put a before to reset the group rule engine to check?

Nishnha:
> I'm thinking that due to the DependencyGroupEngine being a singleton, it might mean that rules declared in other tests are leaking when ran in CI as part of a full suite? 

Looks like it was a leaky dependency group! I fixed this in 9ac2f7e0a4c982187c23b2a0c6d0c7026d4df674 by no longer registering a DependencyGroup for the job spec. The test this was used for checked whether the `:register_dependency_groups` method was called on job creation, but we don't _actually_ need to have a dependency group registered for this test to pass.

PR #7078, opened 2023-04-13
Fix gradle file fetcher when included build is in submodule

patari:
If gradle included build is in submodule, dependabot gives error `dependency_file_not_found` from request to `https://api.github.com:443/repos/<owner>/<repo>/contents/<submodule_path>?ref=<ref>`.

This fixes it by ignoring 404 errors from included build files (gradle subprojects already has similar ignore).

### DIFF
--- a/gradle/lib/dependabot/gradle/file_fetcher.rb
+++ b/gradle/lib/dependabot/gradle/file_fetcher.rb
@@ -41,7 +41,13 @@ module Dependabot
         files += subproject_buildfiles(root_dir)
         files += dependency_script_plugins(root_dir)
         files + included_builds(root_dir).
-                flat_map { |dir| all_buildfiles_in_build(dir) }
+                flat_map do |dir|
+                  all_buildfiles_in_build(dir)
+                rescue Dependabot::DependencyFileNotFound
+                  # included build can be in a submodule
+                  nil
+                end.
+                compact
       end
 
       def included_builds(root_dir)
--- a/gradle/spec/dependabot/gradle/file_fetcher_spec.rb
+++ b/gradle/spec/dependabot/gradle/file_fetcher_spec.rb
@@ -274,6 +274,22 @@ RSpec.describe Dependabot::Gradle::FileFetcher do
             ))
         end
       end
+
+      context "when the included build can't be found" do
+        before do
+          stub_content_request("?ref=sha", "contents_java_with_settings.json")
+          stub_content_request("settings.gradle?ref=sha", "contents_java_settings_1_included_build.json")
+          stub_content_request("build.gradle?ref=sha", "contents_java_basic_buildfile.json")
+          stub_content_request("app/build.gradle?ref=sha", "contents_java_basic_buildfile.json")
+          stub_no_content_request("included?ref=sha")
+        end
+
+        it "fetches the main buildfile" do
+          expect(file_fetcher_instance.files.count).to eq(3)
+          expect(file_fetcher_instance.files.map(&:name)).
+            to match_array(%w(build.gradle settings.gradle app/build.gradle))
+        end
+      end
     end
 
     context "only a settings.gradle" do


jeffwidman:
👋 Thanks for putting this up.

> This fixes it by ignoring 404 errors from included build files (gradle subprojects already has similar ignore).

Can you link to the code containing this ignore or explain this more? Do you mean upstream native Gradle java code already has this, or that Dependabot already has a similar ignore in a different section of our codebase?

patari:
> wave Thanks for putting this up.
> 
> > This fixes it by ignoring 404 errors from included build files (gradle subprojects already has similar ignore).
> 
> Can you link to the code containing this ignore or explain this more? Do you mean upstream native Gradle java code already has this, or that Dependabot already has a similar ignore in a different section of our codebase?

https://github.com/dependabot/dependabot-core/blob/d8840c00ab2e184c891ca146d5ff4fda59505501/gradle/lib/dependabot/gradle/file_fetcher.rb#L83-L85

jeffwidman:
Thanks!

I am not familiar enough with Gradle to even know the difference between Gradle subprojects vs submodules so I'm not able to evaluate this properly w/o some further digging. So gonna "phone-a-friend(ly-contributor)"... 😄 

@soulus13 @yeikel I don't recall if you both are primarily Maven users or also familiar with Gradle... your thoughts on this?

patari:
Yeah, gradle seems to have many ways to define dependencies. There is also something similar in Dependabot for gradle dependency script plugins: https://github.com/dependabot/dependabot-core/blob/d8840c00ab2e184c891ca146d5ff4fda59505501/gradle/lib/dependabot/gradle/file_fetcher.rb#L113-L117

This is just a quick fix to make Dependabot work with gradle included builds in git submodules. This might require some refactoring later but maybe wait for https://github.com/dependabot/dependabot-core/issues/1164?

gradle/lib/dependabot/gradle/file_fetcher.rb
```diff
@@ -41,7 +41,13 @@ def all_buildfiles_in_build(root_dir)
         files += subproject_buildfiles(root_dir)
         files += dependency_script_plugins(root_dir)
         files + included_builds(root_dir).
-                flat_map { |dir| all_buildfiles_in_build(dir) }
+                flat_map do |dir|
+                  all_buildfiles_in_build(dir)
+                rescue Dependabot::DependencyFileNotFound
+                  # included build can be in a submodule
+                  nil
```
jeffwidman, line 48: Just realized this is going to catch _all_ failures to find a file, which seems overly broad.

Is there a way to definitively identify that it's in a git submodule? (since it sounds like Gradle Submodules refer to files in git submodules?)

You pointed out the `file_exists_in_submodule?` method, could we just use that here too?
https://github.com/dependabot/dependabot-core/blob/4cab0ee282f9d774e65bad0c9858ade9b74b45d0/gradle/lib/dependabot/gradle/file_fetcher.rb#L130-L135

soulus13: I'm more of a Maven guy, but indeed, this solution sounds more logical and aligned with the existing implementation, if we only ignore the File Not Found from submodule

jakecoffman:
For a submodule, GitHub doesn't return a 404, but instead a 200 with `"type": "submodule"`. Are you seeing a 404 in the logs? That might be more of an indicator that the included build directory doesn't exist at all. 

I think for submodules the reason it's failing is we don't pass `fetch_sumodules: true` into `fetch_file_if_present` call in the FileFetcher. I assume that is because we won't be updating the dependencies in those build files, so they shouldn't affect the dependencies of the local files? Not sure if that is a correct assumption.

So I think I'm with @jeffwidman, it would be good to check if it's caused by a submodule vs a completely missing directory.

yeikel:
Another reason why a 404 might be returned can be due to permission denied(401) error. For some reason, the Github API was designed to behave that way

I don't know if that would apply to Dependabot though

gradle/lib/dependabot/gradle/file_fetcher.rb
patari: Fixed!

patari:
> For a submodule, GitHub doesn't return a 404, but instead a 200 with `"type": "submodule"`. Are you seeing a 404 in the logs? That might be more of an indicator that the included build directory doesn't exist at all.
> 
> I think for submodules the reason it's failing is we don't pass `fetch_sumodules: true` into `fetch_file_if_present` call in the FileFetcher. I assume that is because we won't be updating the dependencies in those build files, so they shouldn't affect the dependencies of the local files? Not sure if that is a correct assumption.
> 
> So I think I'm with @jeffwidman, it would be good to check if it's caused by a submodule vs a completely missing directory.

The 404 comes from request to `https://api.github.com:443/repos/<owner>/<repo>/contents/<submodule>?ref=<ref>` (`file_fetchers/base#repo_contents`). If we pass `fetch_submodules: true` Dependabot finds all files from the git submodule and starts to find updates for those files also.

I added a commit to ignore included build only if its in a git submodule.

jakecoffman:
I've setup what I think the situation is here: https://github.com/dsp-testing/gradle-simple/tree/f43d406d67a96305fe5f0adce7a104d541ff5acd

If I dry-run this with the CLI `dependabot update gradle dsp-testing/gradle-simple` I get:

```
  proxy | 2023/04/21 12:49:59 [016] GET https://api.github.com:443/repos/dsp-testing/gradle-simple/contents/include?ref=f43d406d67a96305fe5f0adce7a104d541ff5acd
  proxy | 2023/04/21 12:49:59 [016] * authenticating github api request
  proxy | 2023/04/21 12:49:59 [016] 200 https://api.github.com:443/repos/dsp-testing/gradle-simple/contents/include?ref=f43d406d67a96305fe5f0adce7a104d541ff5acd
updater | 2023/04/21 12:49:59 ERROR Error during file fetching; aborting
  proxy | 2023/04/21 12:49:59 [017] POST http://host.docker.internal:58592/update_jobs/cli/record_update_job_error
  proxy | 2023/04/21 12:49:59 [017] 200 http://host.docker.internal:58592/update_jobs/cli/record_update_job_error
  proxy | 2023/04/21 12:49:59 [018] PATCH http://host.docker.internal:58592/update_jobs/cli/mark_as_processed
  proxy | 2023/04/21 12:49:59 [018] 200 http://host.docker.internal:58592/update_jobs/cli/mark_as_processed
updater | 2023/04/21 12:49:59 INFO Finished job processing
updater | 2023/04/21 12:49:59 INFO Results:
updater | Dependabot encountered '1' error(s) during execution, please check the logs for more details.
updater | +---------------------------+
updater | |          Errors           |
updater | +---------------------------+
updater | | dependency_file_not_found |
updater | +---------------------------+
```

Hitting https://api.github.com/repos/dsp-testing/gradle-simple/contents/include?ref=f43d406d67a96305fe5f0adce7a104d541ff5acd I get a 200 with `"type": "submodule"`.

404 due to not authorized isn't possible because it's getting the submodule directory from the current project, not trying to get the contents of the submodule from another repo. 

If you're getting a 404 still from this I'm interested in what your setup is to reproduce that.

patari:
Oh yeah, my included build is in subfolder of a git submodule so requests are little different. The 404 comes from request to `/repos/<owner>/<repo>/contents/<submodule>/<path>`, e.g. https://api.github.com/repos/dsp-testing/gradle-simple/contents/include/path?ref=f43d406d67a96305fe5f0adce7a104d541ff5acd.

I updated spec to match original case (=included build in root of a submodule), the fix works for both cases.

Thomas-Vos:
Thanks for creating this PR, I am running into the same issue. There is an issue #7201 with some details. I checked the logs and for me it's a 404 error. My included build is also in subfolder of a git submodule, and private as well.

vitorhugods:
For more people that may find this issue through the PR, there is a work-around suggested by @patari here:

- https://github.com/dependabot/dependabot-core/issues/7201#issuecomment-1571319655

jakecoffman:
This should be fixed with https://github.com/dependabot/dependabot-core/pull/9278, since we're cloning and fetching submodules up front the file should be present. Let me know if it's working for you without this PR.

PR #7080, opened 2023-04-13
cargo: gather toolchain statistics

jakecoffman:
We want to gather info on the optional Cargo overrides file to find out how many people are using them, and what versions are commonly pinned. That will help us make decisions about what versions of the toolchain we should continue to support.

### DIFF
--- a/cargo/lib/dependabot/cargo/file_fetcher.rb
+++ b/cargo/lib/dependabot/cargo/file_fetcher.rb
@@ -20,6 +20,21 @@ module Dependabot
         "Repo must contain a Cargo.toml."
       end
 
+      def package_manager_version
+        channel = if rust_toolchain
+                    TomlRB.parse(rust_toolchain.content).fetch("toolchain", nil)&.fetch("channel", nil)
+                  else
+                    "default"
+                  end
+
+        {
+          ecosystem: "cargo",
+          package_managers: {
+            "channel" => channel
+          }
+        }
+      end
+
       private
 
       def fetch_files
--- a/cargo/spec/dependabot/cargo/file_fetcher_spec.rb
+++ b/cargo/spec/dependabot/cargo/file_fetcher_spec.rb
@@ -82,6 +82,13 @@ RSpec.describe Dependabot::Cargo::FileFetcher do
       expect(file_fetcher_instance.files.map(&:name)).
         to eq(["Cargo.toml"])
     end
+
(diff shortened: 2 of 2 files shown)


honeyankit (approved):

honeyankit:
Smoke test failing.

jakecoffman:
Smoke test is failing successfully! That's the new call that this adds, I will have to regenerate the test to add the new call:

```diff
+    - type: record_package_manager_version
+      expect:
+        data:
+            ecosystem: cargo
+            package-managers:
+                channel: default
```

cargo/lib/dependabot/cargo/file_fetcher.rb
```diff
@@ -20,6 +20,21 @@ def self.required_files_message
...
+        channel = if rust_toolchain
+                    TomlRB.parse(rust_toolchain.content).fetch("toolchain", nil)&.fetch("channel", nil)
+                  else
+                    "default"
+                  end
+
+        {
+          ecosystem: "cargo",
+          package_managers: {
+            "channel" => channel
```
jeffwidman, line 33: Thinking aloud: 
Looking at https://rust-lang.github.io/rustup/overrides.html#channel, it looks like this will result in values like `"nightly-2020-07-10"`... 

I wonder if we'll ever want to query by "nightly" vs "stable"? For example, the "nightly" will have a super long tail so difficult to group by it... 

I suppose we can just do whatever extraction / grouping in the data consumer later, and "nightly" w/o the associated date is often somewhat meaningless.

So probably good to just go with this for now and call it good.

cargo/spec/dependabot/cargo/file_fetcher_spec.rb
```diff
@@ -123,7 +137,7 @@
         with(headers: { "Authorization" => "token token" }).
         to_return(
           status: 200,
-          body: fixture("github", "contents_cargo_lockfile.json"),
+          body: JSON.dump({ content: Base64.encode64("[toolchain]\nchannel = \"1.2.3\"") }),
```
jeffwidman, line 140: Oh interesting, I was wondering if this value comes in as a version string, based on [their docs](https://rust-lang.github.io/rustup/overrides.html#channel) I was thinking it might show as `"stable-year-day-month"` so it's nice to see this!

jeffwidman (approved):
Looks great! Thanks for doing this!

Issue #1736, opened 2020-03-14
Feature request: support  the `pnpm` package manager


jbergstroem:
[`pnpm`](https://github.com/pnpm/pnpm) is an alternative to npm and yarn that has been around a fairly long time. It has its own lockfile in yaml and should be relatively straightforward to support. Would PRs/other types of help be accepted or are adding more pr's a "can of worms"?

---

[Edit] April 2022 - since the thread is growing: there is still no update on when this will be implemented. If you need this now, your best bet is to use renovatebot: https://docs.renovatebot.com/javascript/

[Edit] November 2022 - for now, github staff suggests using [a workaround using the dependency submission api](https://github.com/dependabot/dependabot-core/issues/1736#issuecomment-1272772849)

brunoparga:
Now that Microsoft own both GitHub and npm, the odds of them supporting pnpm are slim at best.

jbergstroem:
> Now that Microsoft own both GitHub and npm, the odds of them supporting pnpm are slim at best.

My interpretation is that this wouldn't necessarily be political with yarn being supported and all. It's likely due to the smaller adoption of `pnpm` vs `npm` and `yarn`.

didinele:
Would absolutely love this, currently pretty awkward being forced into alternatives like renovate, which are defenitely fine, but nowhere near as satisfying.

hacknug:
`pnpm` become a first class citizen with the last public VSCode release ([changelog](https://code.visualstudio.com/updates/v1_47#_pnpm-package-manager-support) & [PR](https://github.com/microsoft/vscode/pull/100654)). Any chance this can be revisit?

GiriB:
@feelepxyz  Are contributions welcome for this? I see the following work items for this 

- [ ] Update [FileFetcher](https://www.phase2technology.com/blog/open-source-licensing-part-2-software-licensing-is-a-continuum) to fetch the pnpm lockfiles
- [ ]  Update [FileParser](https://github.com/dependabot/dependabot-core/blob/f4cf8cfdcc44fa4f309af2df264a3dabe7ee5b7b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb#L14) to fetch dependencies from the lockfile (Not fully sure how this works. Still need to deep dive. Why do we parse lock files? Just to get the list of dependencies? Or is there more to it?) 
[...]

PR #7081, opened 2023-04-13
Add PNPM support

deivid-rodriguez:
Just sharing some initial work!

### DIFF
--- a/npm_and_yarn/Dockerfile
+++ b/npm_and_yarn/Dockerfile
@@ -1,6 +1,7 @@
 FROM ghcr.io/dependabot/dependabot-updater-core
 ENV DEPENDABOT_HOME /home/dependabot
 ### JAVASCRIPT
+ARG PNPM_VERSION=8.3.1
 ARG YARN_VERSION=3.5.0
 
 # Install Node and npm
@@ -15,6 +16,9 @@ RUN curl -sL https://deb.nodesource.com/setup_16.x | bash - \
 RUN corepack enable
 
 USER dependabot
+
+# Install pnpm and set it to a stable version
+RUN corepack prepare pnpm@$PNPM_VERSION --activate
 
 # Install yarn berry and set it to a stable version
 RUN corepack prepare yarn@$YARN_VERSION --activate
--- a/npm_and_yarn/helpers/lib/pnpm/lockfile-parser.js
+++ b/npm_and_yarn/helpers/lib/pnpm/lockfile-parser.js
@@ -1,0 +1,77 @@
+/* PNPM-LOCK.YAML PARSER
+ *
+ * Inputs:
+ *  - directory containing a pnpm-lock.yaml file
+ *
+ * Outputs:
+ *  - JSON formatted information of dependencies (name, version, dependency-type)
+ */
+const { readWantedLockfile } = require("@pnpm/lockfile-file");
+const dependencyPath = require("@pnpm/dependency-path");
+
+async function parse(directory) {
+  const lockfile = await readWantedLockfile(directory, {
+    ignoreIncompatible: true
+  });
+
+  return Object.entries(lockfile.packages ?? {})
+    .map(([depPath, pkgSnapshot]) => nameVerDevFromPkgSnapshot(depPath, pkgSnapshot, Object.values(lockfile.importers)))
+}
+
+function nameVerDevFromPkgSnapshot(depPath, pkgSnapshot, projectSnapshots) {
+  let name;
+  let version;
+
+  if (!pkgSnapshot.name) {
+    const pkgInfo = dependencyPath.parse(depPath);
+    name = pkgInfo.name;
+    version = pkgInfo.version;
+  } else {
+    name = pkgSnapshot.name;
+    version = pkgSnapshot.version;
+  }
+
+  let specifiers = [];
+  let aliased = false;
+
+  projectSnapshots.every(projectSnapshot => {
+    const projectSpecifiers = projectSnapshot.specifiers;
+
+    if (Object.values(projectSpecifiers).some(specifier => specifier.startsWith(`npm:${name}@`) || specifier == `npm:${name}`)) {
+      aliased = true;
+      return false;
+    }
+
+    currentSpecifier = projectSpecifiers[name];
+
+    if (!currentSpecifier) {
+      return true;
+    }
+
+    let specifierVersion = currentSpecifier.version;
+
+    if (!currentSpecifier.version) {
+      specifierVersion = projectSnapshot.dependencies?.[name] || projectSnapshot.devDependencies?.[name] || projectSnapshot.optionalDependencies?.[name]
+    }
+
+    if (
+      specifierVersion == version ||
+      specifierVersion.startsWith(`${version}_`) || // lockfileVersion 5.4
+      specifierVersion.startsWith(`${version}(`)    // lockfileVersion 6.0
+    ) {
+      specifiers.push(currentSpecifier.specifier || currentSpecifier);
+    }
+
+    return true;
+  });
+
+  return {
+    name: name,
+    version: version,
+    dev: pkgSnapshot.dev,
+    specifiers: specifiers,
+    aliased: aliased
+  }
+}
+
+module.exports = { parse };
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
@@ -56,6 +56,7 @@ module Dependabot
 
         package_managers["npm"] = Helpers.npm_version_numeric(package_lock.content) if package_lock
         package_managers["yarn"] = yarn_version if yarn_version
+        package_managers["pnpm"] = pnpm_version if pnpm_version
         package_managers["shrinkwrap"] = 1 if shrinkwrap
         package_managers["unknown"] = 1 if package_managers.empty?
 
@@ -167,6 +168,18 @@ module Dependabot
         return unless yarn_lock
 
         Helpers.yarn_version_numeric(yarn_lock)
+      end
+
+      def pnpm_version
+        return @pnpm_version if defined?(@pnpm_version)
+
+        @pnpm_version = package_manager.locked_version("pnpm") || guess_pnpm_version
+      end
+
+      def guess_pnpm_version
+        return unless pnpm_lock
+
+        Helpers.pnpm_major_version
       end
 
       def package_manager
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "json"
+require "dependabot/experiments"
 require "dependabot/logger"
 require "dependabot/file_fetchers"
 require "dependabot/file_fetchers/base"
@@ -73,7 +74,7 @@ module Dependabot
         fetched_files << package_json
         fetched_files += npm_files
         fetched_files += yarn_files
-        fetched_files += pnpm_files
+        fetched_files += pnpm_files if Experiments.enabled?(:pnpm_updates)
         fetched_files += lerna_files
         fetched_files += workspace_package_jsons
         fetched_files += path_dependencies(fetched_files)
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
@@ -101,6 +101,8 @@ module Dependabot
       def pnpm_files
         fetched_pnpm_files = []
         fetched_pnpm_files << pnpm_lock if pnpm_lock
+        fetched_pnpm_files << pnpm_workspace_yaml if pnpm_workspace_yaml
+        fetched_pnpm_files += pnpm_workspace_package_jsons
         fetched_pnpm_files
       end
 
@@ -257,6 +259,11 @@ module Dependabot
                        tap { |f| f.support_file = true }
       end
 
+      def pnpm_workspace_yaml
+        @pnpm_workspace_yaml ||= fetch_file_if_present("pnpm-workspace.yaml")&.
+                                tap { |f| f.support_file = true }
+      end
+
       def lerna_json
         @lerna_json ||= fetch_file_if_present("lerna.json")&.
                         tap { |f| f.support_file = true }
@@ -268,6 +275,10 @@ module Dependabot
 
       def lerna_packages
         @lerna_packages ||= fetch_lerna_packages
+      end
+
+      def pnpm_workspace_package_jsons
+        @pnpm_workspace_package_jsons ||= fetch_pnpm_workspace_package_jsons
       end
 
       # rubocop:disable Metrics/PerceivedComplexity
@@ -407,6 +418,14 @@ module Dependabot
         end.compact
       end
 
+      def fetch_pnpm_workspace_package_jsons
+        return [] unless parsed_pnpm_workspace_yaml["packages"]
+
+        workspace_paths(parsed_pnpm_workspace_yaml["packages"]).filter_map do |workspace|
+          fetch_package_json_if_present(workspace)
+        end
+      end
+
       def fetch_lerna_packages_from_path(path)
         package_json = fetch_package_json_if_present(path)
         return unless package_json
@@ -514,6 +533,14 @@ module Dependabot
         JSON.parse(shrinkwrap.content)
       rescue JSON::ParserError
         {}
+      end
+
+      def parsed_pnpm_workspace_yaml
+        return {} unless pnpm_workspace_yaml
+
+        YAML.safe_load(pnpm_workspace_yaml.content)
+      rescue Pysch::SyntaxError
+        raise Dependabot::DependencyFileNotParseable, pnpm_workspace_yaml.path
       end
 
       def skip_package_lock?
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
@@ -72,6 +72,7 @@ module Dependabot
         fetched_files << package_json
         fetched_files += npm_files
         fetched_files += yarn_files
+        fetched_files += pnpm_files
         fetched_files += lerna_files
         fetched_files += workspace_package_jsons
         fetched_files += path_dependencies(fetched_files)
@@ -94,6 +95,12 @@ module Dependabot
         fetched_yarn_files << yarnrc if yarnrc
         fetched_yarn_files << yarnrc_yml if yarnrc_yml
         fetched_yarn_files
+      end
+
+      def pnpm_files
+        fetched_pnpm_files = []
+        fetched_pnpm_files << pnpm_lock if pnpm_lock
+        fetched_pnpm_files
       end
 
       def lerna_files
@@ -180,6 +187,12 @@ module Dependabot
         return @yarn_lock if defined?(@yarn_lock)
 
         @yarn_lock = fetch_file_if_present("yarn.lock")
+      end
+
+      def pnpm_lock
+        return @pnpm_lock if defined?(@pnpm_lock)
+
+        @pnpm_lock = fetch_file_if_present("pnpm-lock.yaml") unless skip_pnpm_lock?
       end
 
       def shrinkwrap
@@ -494,6 +507,12 @@ module Dependabot
         return false unless npmrc
 
         npmrc.content.match?(/^package-lock\s*=\s*false/)
+      end
+
+      def skip_pnpm_lock?
+        return false unless npmrc
+
+        npmrc.content.match?(/^lockfile\s*=\s*false/)
       end
 
       def build_unfetchable_deps(unfetchable_deps)
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
@@ -9,6 +9,7 @@ module Dependabot
     class FileParser
       class LockfileParser
         require "dependabot/npm_and_yarn/file_parser/yarn_lock"
+        require "dependabot/npm_and_yarn/file_parser/pnpm_lock"
         require "dependabot/npm_and_yarn/file_parser/json_lock"
 
         def initialize(dependency_files:)
@@ -22,7 +23,7 @@ module Dependabot
           # end up unique by name. That's not a perfect representation of
           # the nested nature of JS resolution, but it makes everything work
           # comparably to other flat-resolution strategies
-          (yarn_locks + package_locks + shrinkwraps).each do |file|
+          (yarn_locks + pnpm_locks + package_locks + shrinkwraps).each do |file|
             dependency_set += lockfile_for(file).dependencies
           end
 
@@ -50,10 +51,10 @@ module Dependabot
         def potential_lockfiles_for_manifest(manifest_filename)
           dir_name = File.dirname(manifest_filename)
           possible_lockfile_names =
-            %w(package-lock.json npm-shrinkwrap.json yarn.lock).map do |f|
+            %w(package-lock.json npm-shrinkwrap.json pnpm-lock.yaml yarn.lock).map do |f|
               Pathname.new(File.join(dir_name, f)).cleanpath.to_path
             end +
-            %w(yarn.lock package-lock.json npm-shrinkwrap.json)
+            %w(yarn.lock pnpm-lock.yaml package-lock.json npm-shrinkwrap.json)
 
           possible_lockfile_names.uniq.
             filter_map { |nm| dependency_files.find { |f| f.name == nm } }
@@ -67,8 +68,10 @@ module Dependabot
           @lockfiles ||= {}
           @lockfiles[file.name] ||= if [*package_locks, *shrinkwraps].include?(file)
                                       JsonLock.new(file)
-                                    else
+                                    elsif yarn_locks.include?(file)
                                       YarnLock.new(file)
+                                    else
+                                      PnpmLock.new(file)
                                     end
         end
 
@@ -76,6 +79,12 @@ module Dependabot
           @package_locks ||=
             dependency_files.
             select { |f| f.name.end_with?("package-lock.json") }
+        end
+
+        def pnpm_locks
+          @pnpm_locks ||=
+            dependency_files.
+            select { |f| f.name.end_with?("pnpm-lock.yaml") }
         end
 
         def yarn_locks
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/pnpm_lock.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/pnpm_lock.rb
@@ -1,0 +1,68 @@
+# frozen_string_literal: true
+
+require "dependabot/errors"
+
+module Dependabot
+  module NpmAndYarn
+    class FileParser
+      class PnpmLock
+        def initialize(dependency_file)
+          @dependency_file = dependency_file
+        end
+
+        def parsed
+          @parsed ||= SharedHelpers.in_a_temporary_directory do
+            File.write("pnpm-lock.yaml", @dependency_file.content)
+
+            SharedHelpers.run_helper_subprocess(
+              command: NativeHelpers.helper_path,
+              function: "pnpm:parseLockfile",
+              args: [Dir.pwd]
+            )
+          rescue SharedHelpers::HelperSubprocessFailed
+            raise Dependabot::DependencyFileNotParseable, @dependency_file.path
+          end
+        end
+
+        def dependencies
+          dependency_set = Dependabot::NpmAndYarn::FileParser::DependencySet.new
+
+          parsed.each do |details|
+            next if details["aliased"]
+
+            name = details["name"]
+            version = details["version"]
+
+            dependency_args = {
+              name: name,
+              version: version,
+              package_manager: "npm_and_yarn",
+              requirements: []
+            }
+
+            if details["dev"]
+              dependency_args[:subdependency_metadata] =
+                [{ production: !details["dev"] }]
+            end
+
+            dependency_set << Dependency.new(**dependency_args)
+          end
+
+          dependency_set
+        end
+
+        def details(dependency_name, requirement, _manifest_name)
+          details_candidates = parsed.select { |info| info["name"] == dependency_name }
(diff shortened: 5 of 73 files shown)


npm_and_yarn/Dockerfile
```diff
@@ -16,6 +16,10 @@ RUN curl -sL https://deb.nodesource.com/setup_16.x | bash - \
 RUN corepack enable \
   && corepack prepare yarn@$YARN_VERSION --activate
 
+# Install pnpm and set it to a stable version
+RUN corepack enable \
+  && corepack prepare pnpm@8.2.0 --activate
```
raulfdm, line 21: I know this PR is in the early stage, but just to mention, `pnpm` major versions are most often not backward compatible.

If you have a project running pnpm v7 and tries to install deps with v8, it won't work. They use the `pnpm-lock.yaml` file to control this.

deivid-rodriguez: Thanks! Yes, we'll carefully consider the different lockfile formats and PNPM versions. Of course, we'll limit support to some reasonable range of lockfile formats and PNPM versions.

In any case, I'd recommend using the [packageManager](https://nodejs.org/api/packages.html#determining-package-manager) field in `package.json` for best support, since that gives Dependabot a hint of what version of PNPM will produce the most compatible result.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -418,6 +429,24 @@ def fetch_lerna_packages
...
+
+        package_json_files = []
+
+        workspace_paths(parsed_pnpm_workspace_yaml["packages"]).each do |path|
+          file = File.join(path, "package.json")
+
+          begin
+            package_json_files << fetch_file_from_host(file)
+          rescue Dependabot::DependencyFileNotFound
+            nil
```
jeffwidman, line 444: Is it expected this will happen sometimes?

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
```diff
@@ -67,6 +68,8 @@ def lockfile_for(file)
           @lockfiles ||= {}
           @lockfiles[file.name] ||= if [*package_locks, *shrinkwraps].include?(file)
                                       JsonLock.new(file)
+                                    elsif pnpm_locks.include?(file)
+                                      PnpmLock.new(file)
                                     else
                                       YarnLock.new(file)
```
jeffwidman, line 74: What about when these cases conflict such that a user has a lock file for both yarn and pnpm... we don't actually know what they prefer us to do, so should we throw an error in that case?

This is probably true other places in this code as well, and while the issue existed previously, it'll be even more true now that there's three options and `pnpm` being such a close cousin of `npm`... if multiple lockfiles are present, we can't definitively say "npm should win, then pnpm, then yarn"...

npm_and_yarn/spec/dependabot/npm_and_yarn/file_parser/lockfile_parser_spec.rb
```diff
@@ -70,6 +70,68 @@
...
+          expect(dependencies.map(&:name)).to_not include("my-fetch-factory")
+        end
+      end
+
+      context "that contain multiple dependencies" do
+        let(:dependency_files) { project_dependency_files("pnpm/no_lockfile_change") }
+
+        its(:length) { is_expected.to eq(370) }
+
+        describe "a repeated dependency" do
```
jeffwidman, line 106: 👍 for this test case

npm_and_yarn/spec/dependabot/npm_and_yarn/file_parser/lockfile_parser_spec.rb
```diff
@@ -70,6 +70,68 @@
...
+          subject { dependencies.find { |d| d.name == "async" } }
+
+          its(:version) { is_expected.to eq("1.5.2") }
+        end
+      end
+
+      context "that contains specified locked to versions with peer disambiguation suffix" do
```
jeffwidman, line 113: typo? reads like there's a missing word, but I'm actually not sure what it might be...

npm_and_yarn/spec/dependabot/npm_and_yarn/file_parser/lockfile_parser_spec.rb
```diff
@@ -70,6 +70,68 @@
...
+          expect(dependencies.map(&:name)).to include("@typescript-eslint/parser")
+        end
+      end
+
+      context "that contain bad lockfile" do
+        let(:dependency_files) { project_dependency_files("pnpm/broken_lockfile") }
+
+        it "raises a DependencyFileNotParseable error" do
+          expect { dependencies }.
+            to raise_error(Dependabot::DependencyFileNotParseable) do |error|
```
jeffwidman, line 128: Oh here you are already expecting the `DependencyFileNotParseable` that I suggested elsewhere... but what's raising it then since the exception is swallowed at that other location?

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/pnpm_lock.rb
```diff
@@ -0,0 +1,68 @@
...
+            dependency_args = {
+              name: name,
+              version: version,
+              package_manager: "npm_and_yarn",
+              requirements: []
+            }
+
+            if details["dev"]
+              dependency_args[:subdependency_metadata] =
+                [{ production: !details["dev"] }]
```
jeffwidman, line 45: this stanza took a moment to understand what's happening... I wonder if clearer naming of variables might make it more obvious this refers to `dependency-type`'s? `dev`/`production` don't hint at it very strongly... that said, I'm also fine to leave as-is given the helper key name is also `dev`, so viewed through that lens `dev` makes more sense.

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -332,22 +332,22 @@ def fetch_peer_dependency_errors(version:)
           []
         end
 
-        def handle_peer_dependency_errors(error)
+        def handle_peer_dependency_errors(message)
```
jeffwidman: nice cleanup. This one could perhaps be extracted to its own PR to keep the `pnpm` PR smaller?

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -46,6 +46,15 @@ class VersionResolver
             YN0060:\s|\s.+\sprovides\s(?<required_dep>.+?)\s\((?<info_hash>\w+)\).+what\s(?<requiring_dep>.+?)\srequests
           /x
 
+        # Error message from pnpm add:
```
jeffwidman, line 49: Is this line stating that the error message from pnpm will look like the next two lines? Or that all three lines are what pnpm spits out?

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -483,17 +498,21 @@ def lockfiles_for_path(lockfiles:, path:)
         end
 
         def run_checker(path:, version:)
-          # If there are both yarn lockfiles and npm lockfiles only run the
-          # yarn updater
           yarn_lockfiles = lockfiles_for_path(lockfiles: dependency_files_builder.yarn_locks, path: path)
           return run_yarn_checker(path: path, version: version, lockfile: yarn_lockfiles.first) if yarn_lockfiles.any?
 
+          pnpm_lockfiles = lockfiles_for_path(lockfiles: dependency_files_builder.pnpm_locks, path: path)
+          return run_pnpm_checker(path: path, version: version) if pnpm_lockfiles.any?
```
jeffwidman, line 520: Why does yarn win over pnpm if both are present? Why not check for presence of all three, and if more than one found, raise an error that we aren't sure what we should do?

There may be a good logical reason to still do what we do today, but if so, a code comment would be helpful to explain why.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -566,6 +566,12 @@ def ignore_package_lock?
         npmrc.content.match?(/^package-lock\s*=\s*false/)
       end
 
+      def ignore_pnpm_lock?
```
jeffwidman, line 570: The word `ignore` confused me for a sec, as I thought we somehow supported that in https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore

Looks like this is instead actually trying to respect this setting: https://pnpm.io/npmrc#lockfile

Maybe call it `skip_pnpm_lock?` to avoid confusion?

I'm also fine as-is...

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -302,26 +304,34 @@ def updated_types_dependencies
         end
 
         def peer_dependency_errors
-          return @peer_dependency_errors if @peer_dependency_errors_checked
+          resolution_errors.reject { |error| error == "NOT_RESOLVABLE" }
```
jeffwidman, line 307: Where does this `"NOT_RESOLVABLE"` come from? pnpm itself, or something within Dependabot?

jeffwidman (commented):
Took an initial review pass. It was very helpful that you'd kept a clean commit history so I could go through commit-by-commit to understand what the intent was.

Looks really solid so far, great job with the test cases covering a lot of potential error conditions.

A few comments, and some questions about stuff I didn't fully understand.

Also, I merged a bunch of your other PR's, so if you rebase it might slim this down a little?

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
deivid-rodriguez: I think it is yeah.

Users may specify globs in their `pnpm-workspace.yml`, but it could happen that one of the folders that result from resolving the glob does not have a `package.json` file. That would be what this is rescuing. This is essentially copied from the logic of parsing the `workspaces` declaration in the `package.json` file which for `pnpm` lives in a separate file.

To be noted though, is that, we could probably use `Dir.glob` here directly (because we have the full repo cloned locally in this case). If we did that, there'd be no need for this rescue I think.

This was mentioned by @jakecoffman at https://github.com/dependabot/dependabot-core/pull/6932, and I suggested #6923 as a good moment to tackle it.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
deivid-rodriguez: Definitely.

As you mentioned, there are quite a few places were we establish a package manager priority, unfortunately.

In my opinion, it'd be probably be best to infer a package manager in the file fetcher, and error out there if there's a conflict. And then stop looking at different possibilities everywhere else and then just follow the chosen package manager logic through all the update assuming only the proper files for that package manager are present.

But that'd be probably a separate refactoring task?

npm_and_yarn/spec/dependabot/npm_and_yarn/file_parser/lockfile_parser_spec.rb
deivid-rodriguez: Yes, I think I meant

```suggestion
      context "locked to versions with peer disambiguation suffix" do
```

Will update!

npm_and_yarn/spec/dependabot/npm_and_yarn/file_parser/lockfile_parser_spec.rb
deivid-rodriguez: This is the error when failing to parse the lockfile (`pnpm-lock.yaml`). Your suggestion was to also raise it when failing to parse the `pnpm-workspace.yaml`, which makes sense to me and I will implement it!

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
deivid-rodriguez: Sure.

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
deivid-rodriguez: Yeah, same conversation as above. I think we should simplify this. I just tried to do something that follows the current pattern, but I think I also failed to make it consistent anyways.

I'm going to dedicate a small amount of time to try clean this up.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
deivid-rodriguez: I just copied it from the analogous `npm` setting, but I can rename both.

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
deivid-rodriguez: The next two lines. It keeps the comment format of the constants for Yarn and NPM, but I agree it's not very clear. I think we could remove the `Error message from yarn add:`, `Error message from pnpm add:`, `Error message from npm install` headers? It also makes the comments easier to maintain because I actually stopped using `pnpm add` in favor of `pnpm update` and forgot to update this comment.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -172,6 +172,25 @@ def get_yarn_version_from_package_json(package_manager)
         version_match&.named_captures&.fetch("version", nil)
       end
 
+      def pnpm_version
+        return @pnpm_version if defined?(@pnpm_version)
+
+        if (package_manager = parsed_package_manager)
+          get_pnpm_version_from_package_json(package_manager)
+        elsif pnpm_lock
+          8
```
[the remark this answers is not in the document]
jeffwidman: Can you add a code comment saying that?

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -523,6 +552,14 @@ def parsed_shrinkwrap
         {}
       end
 
+      def parsed_pnpm_workspace_yaml
+        return {} unless pnpm_workspace_yaml
+
+        YAML.safe_load(pnpm_workspace_yaml.content)
+      rescue Pysch::SyntaxError
+        {}
```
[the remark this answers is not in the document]
jeffwidman: Out of scope for this PR: your explanation for why to _not_ raise at this point for npm/yarn is great, and would be a very useful code comment in the relevant sections for npm/yarn...

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
jeffwidman: If it's going away, then sounds fine... but if it's sticking around, explaining the expected context for why `nil` might be encountered is a useful code comment... that way if someone is debugging and hits nil, they can compare it to their scenario to see if it's an expected vs unexpected nil

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
jeffwidman: Sounds reasonable on all counts.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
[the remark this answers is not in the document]
deivid-rodriguez: Agreed, will do!

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
jeffwidman: I'm fine with whatever, titles can be helpful context but also understand dropping it as long as it's clear contextually where this error arises... I'm speaking with lots of scar tissue because over in the PHP checker we've got a bunch of error messages, but many of them are outdated with the switch to `composer 2`, only it's not clear which ones... if you choose to keep the title, do something to make it clear :
```suggestion
# Error message returned by `pnpm update` for `pnpm` version 8:
```

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
jeffwidman: I'm cool with also just filing an issue, cross-linking it to our upcoming "ecosystem clarity" effort and moving on for now.

The other way to solve this long term is if we get the user to explicitly tell us which package manager they want us to use... `pnpm`, `yarn`, `npm`, rather than us guessing... I prefer that, although we'll need to figure out how that works in the context of security updates where a repo may not have a `dependabot.yaml` config file.

But if there's easy enough cleanup/tidying where it clarifies the code without getting all the way into redoing our config files, then yeah, that makes sense too... I trust your judgement.

npm_and_yarn/helpers/lib/pnpm/lockfile-parser.js
```diff
@@ -0,0 +1,73 @@
+/* PNPM-LOCK.YAML PARSER
+ *
+ * Inputs:
+ *  - directory containing a pnpm-lock.yaml file
+ *
+ * Outputs:
+ *  - JSON formatted information of dependencies (name, version, type)
```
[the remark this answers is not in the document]
deivid-rodriguez: Added to 876e9d74854f2d09d2d46440309aa943f609064e!

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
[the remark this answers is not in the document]
deivid-rodriguez: I changed 876e9d74854f2d09d2d46440309aa943f609064e to raise a `DependencyFileNotParseable` error. I was going to create a separate PR to comment on why we swallow errors when parsing lockfiles here, but I'm not sure it's the right thing to do vs raising directly, so I decided to pass and move on to other things.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
deivid-rodriguez: I did change the priority so that it's at least consistent for now in the different places.

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
deivid-rodriguez: I opened a ticket! https://github.com/dependabot/dependabot-core/issues/7184

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
deivid-rodriguez: Done at #7178!

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
deivid-rodriguez: I added a comment a refactored this a bit at #7183.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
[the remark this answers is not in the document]
jeffwidman: Nice! Best code comments are well-named method/var names. 😁

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
deivid-rodriguez: I removed this commit. It was an improvement for an edge case (trying to update to a version with yanked deps), but it's not specific to PNPM, so I'll propose it separately if needed.

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
deivid-rodriguez: I ended up updating the message and aligning the other headers at #7195.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
deivid-rodriguez: Renamed and aligned analogous method for NPM at #7197.

deivid-rodriguez:
This is ready for another review round @jeffwidman!

AntoineKM (commented):
Impressive work, well done, love it !

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -496,6 +509,12 @@ def skip_package_lock?
         npmrc.content.match?(/^package-lock\s*=\s*false/)
       end
 
+      def skip_pnpm_lock?
+        return false unless npmrc
+
+        npmrc.content.match?(/^lockfile\s*=\s*false/)
```
jeffwidman, line 556: Could this `lockfile` key have whitespace in front of it and still work?

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -169,6 +170,18 @@ def guess_yarn_version
         Helpers.yarn_version_numeric(yarn_lock)
       end
 
+      def pnpm_version
+        return @pnpm_version if defined?(@pnpm_version)
+
+        @pnpm_version = package_manager.locked_version("pnpm") || guess_pnpm_version
+      end
+
+      def guess_pnpm_version
```
jeffwidman, line 182: I really like the refactor/rename you did with this method, it's so much more clear the intent. 👍

jeffwidman (approved):
Looks great! 🎉 

Had a few trivial comments, they do not block shipping, so can either be addressed now or in a follow-on PR since I know we're trying to get this live so beta users can start hammering on it.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
deivid-rodriguez: I don't know, I followed the pattern of `skip_package_lock?`. This is an INI file, and I'm not sure if that's valid.

That said, this feature is questionable, since it'd be only useful if someone is committing a lockfile and configuring it to not be used. Maybe overkill to add a special case for this ourselves, so we could alternatively remove `skip_package_lock?` and not add this.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser/lockfile_parser.rb
```diff
@@ -78,6 +81,12 @@ def package_locks
             select { |f| f.name.end_with?("package-lock.json") }
         end
 
+        def pnpm_locks
+          @pnpm_locks ||=
+            dependency_files.
+            select { |f| f.name.end_with?("pnpm-lock.yaml") }
```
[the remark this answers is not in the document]
deivid-rodriguez: Yep, it's exactly that `== <name>` is used for root files, `.end_with?(<name>)` for any file with a particular name.

We could probably improve this but it's a pattern used all over, so I just followed the existing pattern.

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -46,6 +46,15 @@ class VersionResolver
             YN0060:\s|\s.+\sprovides\s(?<required_dep>.+?)\s\((?<info_hash>\w+)\).+what\s(?<requiring_dep>.+?)\srequests
           /x
 
+        # Error message returned by `pnpm update`:
```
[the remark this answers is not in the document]
deivid-rodriguez: I have not observed any variations across versions, so I went with optimistically assuming it's stable (at least so far).

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/pnpm_lockfile_updater.rb
```diff
@@ -0,0 +1,172 @@
...
+          handle_pnpm_lock_updater_error(e, pnpm_lock)
+        end
+
+        private
+
+        attr_reader :dependencies, :dependency_files, :repo_contents_path, :credentials
+
+        IRRESOLVABLE_PACKAGE = "ERR_PNPM_NO_MATCHING_VERSION"
+        INVALID_REQUIREMENT = "ERR_PNPM_SPEC_NOT_SUPPORTED_BY_ANY_RESOLVER"
+        MISSING_PACKAGE = /(?<package_req>.*?) is not in the npm registry, or you have no permission to fetch it/
```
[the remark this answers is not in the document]
deivid-rodriguez: That'd be any supported version I think, since I haven't seen any differences between versions. Also, I believe in particular the `ERR_PNPM` codes are there to allow changing the error wording without changing the error code, so those should stay stable. Unfortunately, not all errors have codes like that.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
jeffwidman: I'm also cool with shipping as-is, and we can always later on add extra handling for whitespace if that becomes a problem. My guess is it never will.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/pnpm_lockfile_updater.rb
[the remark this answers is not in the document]
jeffwidman: Ah, that's brilliant of the `pnpm` folks to ship well-known error codes!

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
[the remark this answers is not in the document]
jeffwidman: Sounds fine, I'm not so confident it'll always be stable, but there's always the `git blame` time machine for figuring this out...

deivid-rodriguez:
Alright, let's try this!

AntoineKM:
Hi, I just wanted to know when it will be published on @dependabot ?

BrianLovelace128:
Also interested in the answer to AntoineKM's question.

deivid-rodriguez:
Hello, sorry for the delay. The enablement of this feature has been delayed slightly, but will be made generally available soon. I will post a more specific update once I know all the details!

Stanzilla:
@deivid-rodriguez looks like it is out now? https://github.com/dependabot/dependabot-core/releases/tag/v0.218.0

PeterMK85:
>it doesn't work yet.

Same here.

```bash
The property '#/updates/0/package-ecosystem' value "pnpm" did not match one of the following values: npm, bundler, composer, maven, mix, cargo, gradle, nuget, gomod, docker, elm, gitsubmodule, github-actions, pip, terraform, pub
```

`.github/dependabot.yml`

```yaml
version: 2
enable-beta-ecosystems: true
updates:
    - package-ecosystem: 'pnpm'
```

mhogeveen:
@PeterMK85 For the beta testing the `package-ecosystem` key needed to be set to `npm`, not `pnpm`. Maybe then it will work for you?

PeterMK85:
@mhogeveen 

Okay so adjusting `pnpm` to `npm` started to work 

_but_ 

There is a major issue:
Only `package.json` is being updated but `pnpm-lock.yaml` does not in a monorepo pnpm setup.
So it will fail on the next installation:
`ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with package.json`

Rawnly:
Same problem here, monorepo with pnpm, lock file is ignored

mhogeveen:
@PeterMK85

All beta testers I checked do not have the problem you are having. The key differences I've noticed are the use of the `packageManager` key in the `package.json` and the fact that they were/are beta testers.

You can check out this issue (https://github.com/dependabot/dependabot-core/issues/1736) to find the beta testers repo's for examples of dependabot working on pnpm monorepo's.

Might be worth adding the `packageManager` key to see if that works. Otherwise I'm out of ideas 🤷‍♂️

Rawnly:
Hey @mhogeveen i have the same problem as Peter but I'm currently using the package manager key in my package json. It's currently pnpm 8.5.1 and I'm in the beta. 

All the PRs created by dependabot are failing checks and can't be merged due to the lockfile not being up to date. 

Let me know if you need any other information. 

PS: we're currently on a monorepo managed by turbo.

PeterMK85:
@mhogeveen 

Is it still required to send an email with the repo to @carogalvin ? 
I'm not part of the Beta I thought it was publicly rolled out.

mhogeveen:
Just a quick disclaimer: I'm not affiliated to the project! Just a beta tester trying to help out. If none of the options I've mentioned above seem to work, it's probably best to wait until there is an official message from staff or a contributor.

Rawnly:
> Is it still required to send an email with the repo to @carogalvin ?

this is what we've done so far. Try to reach her out and see if it's still needed

Nishnha:
This feature is still in beta. Unfortunately, we are not accepting more beta testers at the moment.

We will post a larger announcement when pnpm support is generally available, so please be patient and thank you for your support!

PeterMK85:
@Nishnha 

Thanks for the update, by and chance can you share the planned public roll out date(range)?

If it's a matter of weeks / months, might affect the technical approach to resolve the issue.

Huge thanks again for the effort

AntoineKM:
@PeterMK85 if you're looking for an alternative in the meantime you might try @renovate

carogalvin:
Hey everyone! Sorry for the current frustration with the rollout, we're in a company-wide change freeze so we were unfortunately halted on enabling this when we wanted to originally. We are ready to go to GA immediately when it lifts, and we'll let you know on the [issue](https://github.com/dependabot/dependabot-core/issues/1736) when that happens. I can't really give you a date because the target date keeps shifting and I don't want to accidentally lie to you 😅 As a rough estimate, I expect more than a week but less than a month.

We are also investigating the problem with lockfiles!

PR #7084, opened 2023-04-14
[Updater] Wire up and flesh out branch naming for Grouped Update PRs

brrygrdn:
We have spiked out using `DependencyGroup` in branch naming but in our internal tests we are overriding the `prefix` value for projects in order to work around potential naming collisions since the prototype cannot yet manage superseding of existing pull requests.

Before we go into the next round of testing, we should make our branch naming for grouped PRs more compatible with existing automation and also avoid problems of the branch name overflowing due to a larger-than-normal number of dependencies being involved.

This PR finishes off our grouped naming strategy so it incorporates the normal prefix, the package manager, directory and target branch along with the actual group name being used.

### DIFF
--- a/updater/lib/dependabot/updater/operations/group_update_all_versions.rb
+++ b/updater/lib/dependabot/updater/operations/group_update_all_versions.rb
@@ -37,7 +37,7 @@ module Dependabot
           @dependency_snapshot = dependency_snapshot
           @error_handler = error_handler
           # This is a placeholder for a real rule object obtained from config in future
-          @dependency_group = Dependabot::DependencyGroup.new(name: GROUP_NAME_PLACEHOLDER)
+          @dependency_group = Dependabot::DependencyGroup.new(name: GROUP_NAME_PLACEHOLDER, rules: [])
         end
 
         def perform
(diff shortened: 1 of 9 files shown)


updater/lib/dependabot/updater/operations/group_update_all_versions.rb
```diff
@@ -37,7 +37,7 @@ def initialize(service:, job:, dependency_snapshot:, error_handler:)
           @dependency_snapshot = dependency_snapshot
           @error_handler = error_handler
           # This is a placeholder for a real rule object obtained from config in future
-          @dependency_group = Dependabot::DependencyGroup.new(name: GROUP_NAME_PLACEHOLDER)
+          @dependency_group = Dependabot::DependencyGroup.new(name: GROUP_NAME_PLACEHOLDER, rules: [])
```
Nishnha, line 40: nit: I think it would make sense to update GROUP_NAME_PLACEHOLDER here to "all-dependencies". Otherwise branch names will pop up with a `*`

brrygrdn: Good spot, yep, we should be consistent

brrygrdn: Ah, I've actually realised we don't use this name to generate the branch name as API is the one that _actually_ generates the object that gets used.

I'm going to leave this as-is for now to avoid CI churn, etc as we should yoink this placeholder out once #7075 merges

PR #7085, opened 2023-04-14
Fix incorrect detection of top level gemspecs

deivid-rodriguez:
When dealing with libraries, Dependabot sets an extra requirement on Ruby to be the lowest version supported by the gem.

However, it was also incorrectly doing so when dealing with vendored path gems, because all gemspecs were being considered when adding this extra requirement, not just the one in the top level.


landongrindheim (approved):

brrygrdn (approved):
Nice find, thanks for fixing this!

PR #7087, opened 2023-04-14
Improve npm package manager instrumentation

deivid-rodriguez:
This is definitely not perfect, but I think better than what we have.

There's two improvements:

* If given a yarn Berry lockfile, we were still incorrectly instrumenting yarn 1 (unless `packageManager` set in `package.json`).
* If we can't resolve a specific package manager, we'll still run the update job and provide updates. We'll use the npm update checker but we could potentially switch to any other package manager without issues, so I think it's helpful to log separately.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
@@ -16,6 +16,14 @@ module Dependabot
         6
       end
 
+      def self.yarn_version_numeric(lockfile_content)
+        if yarn_berry?(lockfile_content)
+          3
+        else
+          1
+        end
+      end
+
       def self.fetch_yarnrc_yml_value(key, default_value)
         if File.exist?(".yarnrc.yml") && (yarnrc = YAML.load_file(".yarnrc.yml"))
           yarnrc.fetch(key, default_value)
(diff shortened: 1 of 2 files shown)


jakecoffman (approved):

npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
```diff
@@ -16,6 +16,14 @@ def self.npm_version_numeric(lockfile_content)
         6
       end
 
+      def self.yarn_version_numeric(lockfile_content)
+        if yarn_berry?(lockfile_content)
+          3
+        else
+          1
+        end
```
landongrindheim, line 24: 🔎 Can we use [yarn_major_version](https://github.com/dependabot/dependabot-core/pull/7087/files#diff-85ec3865f2fdb48045a2536b37b847156ed3bd3960afdc8bd900ae568de9acd7R42) for this?

deivid-rodriguez: This is tricky stuff. `yarn --version` would give us the version Dependabot is running (which will always be Yarn Berry I believe). I think it's more interesting to know the version the project needs, and the `yarn_berry?` helper is the criteria used in the update checker and other places to decide whether to run Yarn Berry, or Yarn 1 native helpers. So, I think `yarn_berry?` is more reliable/informative actually.

PR #7090, opened 2023-04-14
use tags in error reporting

jakecoffman:
Tags are searchable, extra is not. This improves our ability to figure out what is going wrong with a job ID.


landongrindheim (approved):
🏷️

deivid-rodriguez:
Here too I think:

https://github.com/dependabot/dependabot-core/blob/b7e1352c8c6c8284ba5fd689bf7a13eeb8baccfa/common/lib/dependabot/shared_helpers.rb#L67

jakecoffman:
@deivid-rodriguez I could be wrong, but I think it's not needed there because it gets handled here:

https://github.com/dependabot/dependabot-core/blob/8f3342a4028a31eee80cc0982884eabeae0189d0/updater/lib/dependabot/updater/error_handler.rb#L103-L111

deivid-rodriguez:
Now, you're right! I thought you wanted to replace every "extra" with "tags", but I see now that you wanted it only for the job ID, which makes sense!

deivid-rodriguez (approved):

PR #7091, opened 2023-04-14
[Updater] Grouped updates will not include ignored dependencies

brrygrdn:
We originally bypassed ignore conditions for the grouped updates prototype as we were not sure they would interact predictably at first.

After some further testing, we are good to restore this capability when creating group PRs.

### DIFF
--- a/updater/spec/dependabot/updater_spec.rb
+++ b/updater/spec/dependabot/updater_spec.rb
@@ -18,8 +18,12 @@ RSpec.describe Dependabot::Updater do
       to_return(status: 200, body: fixture("rubygems-index"))
     stub_request(:get, "https://index.rubygems.org/info/dummy-pkg-a").
       to_return(status: 200, body: fixture("rubygems-info-a"))
+    stub_request(:get, "https://rubygems.org/api/v1/versions/dummy-pkg-a.json").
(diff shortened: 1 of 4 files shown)


brrygrdn:
Ops, it looks like I have some bad copypasta in the grouped update smoke test, I'll fix that now

updater/spec/dependabot/updater_spec.rb
```diff
@@ -2234,6 +2236,29 @@ def expect_update_checker_with_ignored_versions(versions)
 
       updater.run
     end
+
+    it "does not include ignored dependencies in the group PR" do
```
Nishnha, line 2240: I think it would be helpful to also include a test checking the length of `dependency_change.updated_dependencies` to show it only updates 1 dependency, so if it's ignored then there is no PR created.

Nishnha (approved):

brrygrdn:
Something where happened with the automerge resulting in the PR being stuck - I pulled main and merged it back into this branch as the PR said there were conflicts but couldn't show them.

It seems the branch had merged into main, but the PR had been left dangling which is why it is now showing as closed rather than merged.

deivid-rodriguez:
Weird!

PR #7094, opened 2023-04-15
chore(gradle): remove unused pom fixtures

yeikel:
These files are not used within the gradle specs


deivid-rodriguez (approved):
Wooow, awesome catch! Thanks for cleaning this up!

PR #7108, opened 2023-04-18
Improve handling missing package metadata

deivid-rodriguez:
While working on PNPM I found the npm `faker` library declares a private repository as the source for its metadata. See https://registry.npmjs.org/faker/latest.

This results in Dependabot hanging on a prompt when trying to access metadata.

So I created da03a7a82bd96ec3866e978fc1c183f6ac550ca7 to fix that.

But then I noticed that the PR it creates, while including a correct diff and version bump, it includes links to an inaccessible repository.

So I created 09565494c938287041a00822597c1232a5563836 to instead build an "empty PR" without metadata.

But it still felt we could do a bit better and try to give a hint on why we failed to find metadata. So I created 7702c47983107d52b78b6e742a45be07a1dc3cae.

This still needs work to make tests pass but I wanted to ask for opinions.


v-kbukum1:
@deivid-rodriguez , is that problem going on? Do we want to fix the problem or it is ok to close this PR?

deivid-rodriguez:
Sorry @kbukum1, I don't know but I'd assume it's still an issue since the package pointing to a private repo in its metadata is still there. It's just an edge case, feel free to close it if you'd rather not work on completing this PR. I'm not planning to work on this myself. Thanks!

v-kbukum1:
I think we still can keep that. I will find time to later maybe to fix conflicts and see if I can merge that into the dependabot.

cceneag:
👍

deivid-rodriguez:
I'll close since it does not seem that this fix will get landed. Feel free to reopen separately if still interested, though!

PR #7112, opened 2023-04-19
Fix parsing some YAMLs due to outdated libyaml

deivid-rodriguez:
While working on PNPM, I was going crazy because Dependabot could not parse the following YAML file:

https://github.com/ice-lab/icestark/blob/ec2d44244a2b4ebd7af0e0ff0a10f528b33b8a64/pnpm-lock.yaml

However, it would be parsed just fine on my dev machine.

On my dev machine:

```
$ ruby -rpsych -e 'Psych.load_file("tmp/ice-lab/icestark/pnpm-lock.yaml")'
# ok
```

On the development container:

```
$ ruby -rpsych -e 'Psych.load_file("tmp/ice-lab/icestark/pnpm-lock.yaml")'
/usr/local/lib/ruby/gems/3.1.0/gems/psych-4.0.4/lib/psych.rb:455:in `parse': (tmp/ice-lab/icestark/pnpm-lock.yaml): did not find expected ',' or '}' while parsing a flow mapping at line 1308 column 17 (Psych::SyntaxError)
	from /usr/local/lib/ruby/gems/3.1.0/gems/psych-4.0.4/lib/psych.rb:455:in `parse_stream'
	from /usr/local/lib/ruby/gems/3.1.0/gems/psych-4.0.4/lib/psych.rb:399:in `parse'
	from /usr/local/lib/ruby/gems/3.1.0/gems/psych-4.0.4/lib/psych.rb:323:in `safe_load'
	from /usr/local/lib/ruby/gems/3.1.0/gems/psych-4.0.4/lib/psych.rb:369:in `load'
	from /usr/local/lib/ruby/gems/3.1.0/gems/psych-4.0.4/lib/psych.rb:671:in `block in load_file'
	from /usr/local/lib/ruby/gems/3.1.0/gems/psych-4.0.4/lib/psych.rb:670:in `open'
	from /usr/local/lib/ruby/gems/3.1.0/gems/psych-4.0.4/lib/psych.rb:670:in `load_file'
	from -e:1:in `<main>'

```

After investigating, turns out I was using a newer version of libyaml locally, which must've corrected some bugs.

Unfortunately, updating the underlying libyaml is painful, but I think it can be done, so this is my take at it.

An alternative approach would be to avoid parsing YAML. For my purposes is doable, since I just wanted to detect the version used by the PNPM lockfile and that can be figured out pretty reliably without parsing YAML. But I was worried that this could be biting us in other places so I'm trying to upgrade libyaml.

### DIFF
--- a/Dockerfile.updater-core
+++ b/Dockerfile.updater-core
@@ -95,13 +95,24 @@ COPY --from=ruby:3.1.4 --chown=dependabot:dependabot /usr/local /usr/local
 # This way other projects that import this library don't have to futz around with installing new / unexpected bundler versions.
 ARG BUNDLER_V2_VERSION=2.4.11
 
+# We had to explicitly bump this as the bundled version `0.2.2` in ubuntu 20.04 has a bug.
+# Once Ubuntu base image pulls in a new enough yaml version, we may not need to
+# explicitly manage this. However, if we do opt to pull it back out, see all changes
+# required in https://github.com/dependabot/dependabot-core/pull/7112
+ARG LIBYAML_VERSION=0.2.5
+RUN curl -sL https://pyyaml.org/download/libyaml/yaml-$LIBYAML_VERSION.tar.gz -o libyaml.tar.gz && \
+  mkdir -p $DEPENDABOT_HOME/src/libyaml && \
+  tar -xvf libyaml.tar.gz -C $DEPENDABOT_HOME/src/libyaml && \
+  rm libyaml.tar.gz
+
 RUN gem install bundler -v $BUNDLER_V2_VERSION --no-document && \
  rm -rf /var/lib/gems/*/cache/* && \
+ bundle config set --global build.psych --with-libyaml-source-dir=$DEPENDABOT_HOME/src/libyaml/yaml-$LIBYAML_VERSION && \
  bundle config set --local path 'vendor' && \
  bundle config set --local frozen 'true' && \
  bundle config set --local without 'development' && \
  bundle install && \
- rm -rf ~/.bundle
+ rm -rf ~/.bundle/cache
 
 COPY --chown=dependabot:dependabot LICENSE $DEPENDABOT_HOME
 
--- a/common/dependabot-common.gemspec
+++ b/common/dependabot-common.gemspec
@@ -38,6 +38,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "nokogiri", "~> 1.8"
   spec.add_dependency "octokit", ">= 4.6", "< 7.0"
   spec.add_dependency "parser", ">= 2.5", "< 4.0"
+  spec.add_dependency "psych", "~> 5.0"
   spec.add_dependency "toml-rb", ">= 1.1.2", "< 3.0"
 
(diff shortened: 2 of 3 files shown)


brrygrdn (commented):
This upgrade makes sense to me!

Regarding the test failures, the Updater spec failure looks like we might have a new flake potentially introduced by #7075 - it's worth rebasing/merging main to see if it clears up just in case but we might need to go look at that.

deivid-rodriguez:
Well, that didn't do the trick. Hopefully it's fine after #7114.

Dockerfile.updater-core
```diff
@@ -95,13 +95,20 @@ COPY --from=ruby:3.1.4 --chown=dependabot:dependabot /usr/local /usr/local
 # This way other projects that import this library don't have to futz around with installing new / unexpected bundler versions.
 ARG BUNDLER_V2_VERSION=2.4.11
 
+ARG LIBYAML_VERSION=0.2.5
```
jeffwidman, line 98:
```suggestion
# We had to explicitly bump this as the bundled version in ubuntu 20.04 has a bug.
# Once Ubuntu base image pulls in a new enough yaml version, we may not need to
# explicitly manage this. However, if we do opt to pull it back out, see all changes
# required in https://github.com/dependabot/dependabot-core/pull/7112
ARG LIBYAML_VERSION=0.2.5
```

Dockerfile.updater-core
```diff
@@ -95,13 +95,20 @@ COPY --from=ruby:3.1.4 --chown=dependabot:dependabot /usr/local /usr/local
...
+
 RUN gem install bundler -v $BUNDLER_V2_VERSION --no-document && \
  rm -rf /var/lib/gems/*/cache/* && \
+ bundle config set --global build.psych --with-libyaml-source-dir=$DEPENDABOT_HOME/src/libyaml/yaml-$LIBYAML_VERSION && \
  bundle config set --local path 'vendor' && \
  bundle config set --local frozen 'true' && \
  bundle config set --local without 'development' && \
  bundle install && \
- rm -rf ~/.bundle
+ rm -rf ~/.bundle/cache
```
jeffwidman, line 111: Should this change happen regardless of the libyaml bump? Or is does somehow nuking the entire `~/.bundle` dir help with docker image caching more than nuking `~/.bundle/cache`?

jeffwidman (approved):
Seems fine to me.

It looks like Focal (which we're on now) and Jammy (which we'll be moving to in https://github.com/dependabot/dependabot-core/pull/5030) both pull in `0.2.2` but starting in Kinetic (`22.10`) `libyaml-dev` defaults to `0.2.5`: https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=libyaml-dev&searchon=names

Dockerfile.updater-core
deivid-rodriguez: It's because I need to configure Bundler globally to use the downloaded sources and the previous `rm -rf ~/.bundle` would wipe out the configuration, which is also needed when installing gems from `omnibus/`.

jeffwidman (approved):

PR #7113, opened 2023-04-19
Fix flaky specs in updater

deivid-rodriguez:
Reproducer:

```
$ script/ci-test-updater './spec/dependabot/dependency_group_engine_spec.rb[1:4:1]' './spec/dependabot/updater_spec.rb[1:2:1]' --seed 24926
```

There was one spec causing side effects that was missing cleanup.

Also I changed cleaning up side effects to be done after the specs that create them, not before.


landongrindheim (approved):

deivid-rodriguez:
Thanks ❤️

jeffwidman:
> Also I changed cleaning up side effects to be done after the specs that create them, not before.

Oof, nice catch! 😍

PR #7114, opened 2023-04-19
Fix more order dependent spec failures

deivid-rodriguez:
We were also leaking the grouped updates experiment.

Failures can be reproduced with:

```
$ script/ci-test-updater './spec/dependabot/dependency_group_engine_spec.rb:129' './spec/dependabot/update_files_command_spec.rb:52' --order=defined
```


deivid-rodriguez:
Thanks again @landongrindheim!

PR #7118, opened 2023-04-19
Don't show empty releases section

deivid-rodriguez:
When the updated dependency has no GitHub releases, it's pointless to link to an empty page.

If accepted, this should go together with a PR updating the smoke tests.


landongrindheim (approved):

jeffwidman (approved):

deivid-rodriguez:
This one should be ready now, but I don't recall the process to move forward a PR that requires smoke tests changes. Do I merge the PR with red CI, and then regenerate smoke tests at https://github.com/dependabot/smoke-tests?

common/lib/dependabot/metadata_finders/base/release_finder.rb
```diff
@@ -20,9 +20,16 @@ def initialize(source:, dependency:, credentials:)
         def releases_url
           return unless source
 
+          # Azure does not provide tags via API, so we can't check whether
+          # there are any releases. So, optimistically return the tags location
+          return "#{source.url}/tags" if source.provider == "azure"
```
deivid-rodriguez, line 25: Getting the specs to pass revealed that Azure does not have a releases API that I could mock, so I couldn't get the existing test to pass.

Instead, I kept the existing behavior, which I think it makes sense. Since Azure does not have a releases API, we cannot check in advance whether there are any releases available, so optimistically including the link sounds reasonable to me.

deivid-rodriguez:
Verified that the updated smoke tests pass, so I'll be merging this and the smoke tests PR shortly!

PR #7119, opened 2023-04-19
Bundler: handle SolveFailures as expected errors

jakecoffman:
This should give the user a nice message on why Dependabot couldn't solve the dependencies given, and stop showing as "unknown errors" for our reporting.

### DIFF
--- a/bundler/lib/dependabot/bundler/file_updater/lockfile_updater.rb
+++ b/bundler/lib/dependabot/bundler/file_updater/lockfile_updater.rb
@@ -79,6 +79,10 @@ module Dependabot
               )
             end
           post_process_lockfile(lockfile_body)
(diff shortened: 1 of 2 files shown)


brrygrdn (approved):
This makes sense to me 👍🏻 We can still track the trend of `DependencyFileNotResolvable` without it setting off alarms which seems like a happy medium.

jakecoffman:
Ooo test failure, looks like this will be a little more complicated as the SolveFailure is handled further up the chain to update multiple dependencies.

jakecoffman:
I've scoped it down to just when the lockfile_updater throws it, and now all is well. There might be a case to call `handle_bundler_errors(e)` here in the future.

PR #7120, opened 2023-04-19
Cargo: handle TOML parsing errors

jakecoffman:
Per the [docs](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file), the toolchain files should be TOML since rustup 1.23. We're seeing a fair amount of projects using the old style, which won't work with what version we're on, so going to turn that unknown error into a DependencyFileNotParseable error.

### DIFF
--- a/cargo/lib/dependabot/cargo/file_fetcher.rb
+++ b/cargo/lib/dependabot/cargo/file_fetcher.rb
@@ -34,8 +34,10 @@ module Dependabot
           }
         }
       rescue TomlRB::ParseError
-        raise Dependabot::DependencyFileNotParseable.new(rust_toolchain.path,
-                                                         "only rust-toolchain files formatted as TOML are supported, the non-TOML format was deprecated by Rust")
+        raise Dependabot::DependencyFileNotParseable.new(
+          rust_toolchain.path,
+          "only rust-toolchain files formatted as TOML are supported, the non-TOML format was deprecated by Rust"
+        )
       end
 
       private
--- a/cargo/lib/dependabot/cargo/file_fetcher.rb
+++ b/cargo/lib/dependabot/cargo/file_fetcher.rb
@@ -33,6 +33,8 @@ module Dependabot
             "channel" => channel
           }
         }
+      rescue TomlRB::ParseError
+        raise Dependabot::DependencyFileNotParseable, rust_toolchain.path
       end
 
       private
--- a/cargo/lib/dependabot/cargo/file_fetcher.rb
+++ b/cargo/lib/dependabot/cargo/file_fetcher.rb
@@ -35,7 +35,7 @@ module Dependabot
         }
       rescue TomlRB::ParseError
         raise Dependabot::DependencyFileNotParseable.new(rust_toolchain.path,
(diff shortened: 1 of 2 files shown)


cargo/lib/dependabot/cargo/file_fetcher.rb
```diff
@@ -33,6 +33,8 @@ def package_manager_version
             "channel" => channel
           }
         }
+      rescue TomlRB::ParseError
+        raise Dependabot::DependencyFileNotParseable, rust_toolchain.path
```
jeffwidman, line 37: I wish we could pass an optional custom error message in addition to the path, because the rust docs say they still support the legacy format, so we're likely to get customers complaining "`rustc` works with my config file, why doesn't :dependabot: ??"

So this would be a perfect place to say "Rust switched to TOML format since XYZ, so we only support TOML versions"...

Not a big deal though...

jeffwidman (approved):
Another nice find Jake! 🎉

cargo/lib/dependabot/cargo/file_fetcher.rb
deivid-rodriguez: We can! `Dependabot::DependencyFileNotParseable.new(rust_toolchain.path, <custom_message>)`

deivid-rodriguez: See https://github.com/dependabot/dependabot-core/pull/7003 (wrong) and https://github.com/dependabot/dependabot-core/pull/7068 (fixed).

deivid-rodriguez: Actually those PR didn't really fix this, they just give a more sensible message if you don't pass a custom message. But just wanted to say that this should be possible!

jeffwidman: ah right! I was looking only at https://github.com/dependabot/dependabot-core/blob/009b8f2c34ada42b4a2659b7fd9bd3054049aa78/common/lib/dependabot/errors.rb#L97 but forgot to check the underlying base class.

jakecoffman: I added a friendlier message. How does that look?

cargo/lib/dependabot/cargo/file_fetcher.rb
```diff
@@ -33,6 +33,9 @@ def package_manager_version
             "channel" => channel
           }
         }
+      rescue TomlRB::ParseError
+        raise Dependabot::DependencyFileNotParseable.new(rust_toolchain.path,
+                                                         "only rust-toolchain files with TOML syntax are supported")
```
jeffwidman, line 38:
```suggestion
                                                         "only rust-toolchain files formatted as TOML are supported, the non-TOML format was deprecated by Rust")
```

jeffwidman (approved):
Looks great!

I added an optional tweak that clarifies this isn't something that "we" are driving, but rather an upstream Rust-lang thing...

jeffwidman:
Merging to keep the CI train moving so we don't get too backlogged...

PR #7121, opened 2023-04-19
🛑 Stop clobbering Octokit middleware

landongrindheim:
If `dependabot-core` is used in another application which has a separate dependency on Octokit, requests made through `dependabot-core` will rewrite the middleware stack. This is an attempt to prevent that behavior and make the setting of middleware specific to the client being built.


jeffwidman (approved):
So IIUC, `Octokit.middleware` is a global? So you're changing the scope of this setting to only the local client?

deivid-rodriguez (approved):
It makes sense to me too!

landongrindheim:
> So IIUC, Octokit.middleware is a global? So you're changing the scope of this setting to only the local client?

Correct 😄 

I probably could have described that better. Thanks for the question/making it clearer 🙇

PR #7122, opened 2023-04-19
Cargo: handle unsupported toolchain versions

jakecoffman:
Follow-up to #7066, this will display a helpful error when doing an update on a toolchain that is not supported by Dependabot.


cargo/lib/dependabot/cargo/file_updater/lockfile_updater.rb
```diff
@@ -145,6 +145,10 @@ def run_shell_command(command, fingerprint:)
           # returns a non-zero status
           return if process.success?
 
+          if stdout.include?("usage of sparse registries requires `-Z sparse-registry`")
```
jeffwidman, line 148: Interesting, I was expecting you to check for the specific version, but I guess that'd force having to check stable vs nightly separately, this is much cleaner!

However, what happens if someone runs with a version that is so old it isn't even aware of the `sparse-registry` option at all? Will it quietly ignore the `CARGO_REGISTRIES_CRATES_IO_PROTOCOL` env var value altogether or loudly complain that it's an unknown value?

https://github.com/dependabot/dependabot-core/blob/009b8f2c34ada42b4a2659b7fd9bd3054049aa78/cargo/Dockerfile#L8

jeffwidman (approved):
I'm 👍 to ship this as-is, although this may not fully protect against _really_ old versions... 🤷‍♂️ 

That said, really old versions are probably a really low % of our users, as  folks who lag that far on Rust are unlikely to be :dependabot: power users.

cargo/lib/dependabot/cargo/file_updater/lockfile_updater.rb
jeffwidman: Also, FTR, here's more info on when this error is thrown:
* https://github.com/rust-lang/cargo/issues/11965

jakecoffman: Yes this is the Python way, just try it and see if it works, handle the exception. Saves a lot of code. 

I think with much older versions they'd just work as they do now. It seems like very few users at this point though, looking at the stats. Probably not worth the effort of parsing dates, versions, and figuring out which are affected. Some of our [pause work](https://github.blog/changelog/2023-01-12-dependabot-pull-requests-pause-for-inactivity/) may take care of those anyway.

PR #7126, opened 2023-04-20
CodeQL: ignore purposefully invalid ruby files

jeffwidman:
CodeQL is complaining that these two files are unparsable Ruby:
```
[2023-04-20 07:34:33] [build-stdout] [2023-04-20 07:34:33] [build-stdout]  WARN /home/runner/work/dependabot-core/dependabot-core/bundler/spec/fixtures/projects/bundler1/invalid_ruby/Gemfile:2: A parse error occurred. Check the syntax of the file. If the file is invalid, correct the error or exclude the file from analysis.
[2023-04-20 07:34:33] [build-stdout] [2023-04-20 07:34:33] [build-stdout]  WARN /home/runner/work/dependabot-core/dependabot-core/bundler/spec/fixtures/projects/bundler2/invalid_ruby/Gemfile:2: A parse error occurred. Check the syntax of the file. If the file is invalid, correct the error or exclude the file from analysis.
```

These two files are test specs and intentionally unparsable, so ignoring them.

Docs on configuring ignores are here:
* https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/customizing-code-scanning#specifying-directories-to-scan
* https://github.com/github/codeql-action#configuration-via-config-input

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -51,6 +51,10 @@ jobs:
         languages: ${{ matrix.language }}
         # TODO: Unpin when https://github.com/github/codeql/issues/13103 fixed.
         tools: https://github.com/github/codeql-action/releases/download/codeql-bundle-20230418/codeql-bundle.tar.gz
+        config: |
+          paths-ignore:
+            - 'bundler/spec/fixtures/projects/bundler1/invalid_ruby/Gemfile'
+            - 'bundler/spec/fixtures/projects/bundler2/invalid_ruby/Gemfile'
         # If you wish to specify custom queries, you can do so here or in a config file.
         # By default, queries listed here will override any specified in a config file.
         # Prefix the list here with "+" to use these queries and those in the config file.


.github/workflows/codeql-analysis.yml
```diff
@@ -50,6 +50,9 @@ jobs:
       uses: github/codeql-action/init@v2
       with:
         languages: ${{ matrix.language }}
+        paths-ignore:
+          - 'bundler/spec/fixtures/projects/bundler1/invalid_ruby/Gemfile'
+          - 'bundler/spec/fixtures/projects/bundler2/invalid_ruby/Gemfile'
```
jeffwidman, line 55: Their docs show this goes in a separate config file that then gets configured in the workflow file, but I was hoping I could get away with putting it directly into the workflow file... no dice, at least not this way: https://github.com/dependabot/dependabot-core/actions/runs/4751969153

jeffwidman: Actually, looks like there's a PR close to being merged that'd let us pass config directly in the workflow file rather than the indirection of _yet-another-config-file_...
* https://github.com/github/codeql-action/pull/1590

jeffwidman: This should be unblocked now: https://github.com/github/codeql-action/blob/main/CHANGELOG.md#233---04-may-2023

jeffwidman:
```shell
  [2023-05-12 08:45:39] [build-stdout] [2023-05-12 08:45:39] [build-stdout]  WARN /home/runner/work/dependabot-core/dependabot-core/common/spec/dependabot/dependency_file_spec.rb:81: A parse error occurred. Check the syntax of the file. If the file is invalid, correct the error or exclude the file from analysis.
  [2023-05-12 08:45:39] [build-stdout] [2023-05-12 08:45:39] [build-stdout]  WARN /home/runner/work/dependabot-core/dependabot-core/common/spec/dependabot/dependency_file_spec.rb:83: A parse error occurred. Check the syntax of the file. If the file is invalid, correct the error or exclude the file from analysis.
```

🤔 hmm, these look okay to me:
https://github.com/dependabot/dependabot-core/blob/f103d051a0e4b1a1df4a88b644c4c9bae5887f58/common/spec/dependabot/dependency_file_spec.rb#L81-L83

deivid-rodriguez (approved):

PR #7128, opened 2023-04-20
[Updater] Extract RefreshSecurityUpdatePullRequest into an Operation class

brrygrdn:
Follows on from https://github.com/dependabot/dependabot-core/pull/6866, https://github.com/dependabot/dependabot-core/pull/6884, https://github.com/dependabot/dependabot-core/pull/6939, https://github.com/dependabot/dependabot-core/pull/6961

This branch is the last of four class extractions to peel out the distinct scenarios the Updater _actually_ handles today in clear terms.

#### Changes

This branch peels out the Updater code required to accept a job which is refreshing a Dependabot PR for a security update using the typical approach of:
- Remove any code related to non-concerns for this operation type, in this case version updates and rebases

The one deviation is that in the initial commit, I've extracted all Security-related helpers into a mixin to encapsulate and avoid duplicating them since they are identical between the "Create" and "Refresh" scenarios. This also affords removing them from the legacy codeline and adding the mixin instead.

For reference, this class is basically a remix of those introduced in https://github.com/dependabot/dependabot-core/pull/6939 and https://github.com/dependabot/dependabot-core/pull/6939

#### What's next

With this PR, every job we run should now be going through one of the four updater classes instead of the `Updater#legacy_run` method.

I intend to create a PR that branches from this to delete that code and mop up any test cases that are falling into implicit cases we don't actually run in the real world, but I plan to ship this to verify that no production jobs are leaking into the legacy code first as that could change the remaining scope as it might mean our analysis has some gaps.

### DIFF
--- a/updater/lib/dependabot/updater/operations/refresh_security_update_pull_request.rb
+++ b/updater/lib/dependabot/updater/operations/refresh_security_update_pull_request.rb
@@ -1,0 +1,254 @@
+# frozen_string_literal: true
+
+# This class implements our strategy for 'refreshing' an existing Pull Request
+# that updates an insecure dependency.
+#
+# TODO: copyedit
+#
+# It will determine if the existing diff is still relevant, in which case it
+# functions similar to a "rebase", but in the case where the project folder's
+# dependencies have changed or a newer version is available, it will supersede
+# the existing pull request with a new one for clarity.
+module Dependabot
+  class Updater
+    module Operations
+      class RefreshSecurityUpdatePullRequest
+        include SecurityUpdateHelpers
+
+        def self.applies_to?(job:)
+          return false unless job.security_updates_only?
+          # If we haven't been given metadata about the dependencies present
+          # in the pull request, this strategy cannot act.
+          return false if job.dependencies&.none?
+
+          job.updating_a_pull_request?
+        end
+
+        def self.tag_name
+          :update_security_pr
+        end
+
+        def initialize(service:, job:, dependency_snapshot:, error_handler:)
+          @service = service
+          @job = job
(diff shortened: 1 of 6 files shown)


updater/lib/dependabot/updater/operations/refresh_security_update_pull_request.rb
```diff
@@ -0,0 +1,233 @@
...
+
+        def initialize(service:, job:, dependency_snapshot:, error_handler:)
+          @service = service
+          @job = job
+          @dependency_snapshot = dependency_snapshot
+          @error_handler = error_handler
+        end
+
+        def perform
+          dependency = dependencies.last
```
jakecoffman, line 39: Why not `.first` which is the lead dependency? I think we refer to it in other places, might make sense to be consistent with error messaging. Or was this how it has always been?

jakecoffman (approved):
LGTM!

updater/lib/dependabot/updater/operations/refresh_security_update_pull_request.rb
brrygrdn: > Or was this how it has always been?

Yes, for some reason we have always used first for `lead_dependency` but last here. There are three places we select 1 of N dependencies, 2 are `last` and 1 is `first` and I'm not 100% sure why.

I will scan back through the code and add some TODOs on this though as I think I earmarked it in the prototype way back and didn't leave a note in the actual implementation.

jakecoffman: No harm in leaving it then, don't want to introduce a new bug somehow!

Issue #6202, opened 2022-11-23
Remove / fix remaining hardcoded branch references to `"master"`


jeffwidman:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code improvement description

https://github.com/dependabot/dependabot-core/pull/6149 got me curious, and indeed we have two more places we hardcode a reference to `"master"`:

https://github.com/dependabot/dependabot-core/blob/0ac4e37da9bb6c054ab75116a4e891404490a342/npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser.rb#L289

https://github.com/dependabot/dependabot-core/blob/0ac4e37da9bb6c054ab75116a4e891404490a342/hex/helpers/lib/parse_deps.exs#L96

These are the only two remaining hardcodings that I found.

We should fix those to fallback to the default branch, which may _not_ be `"master"`.

Related fix:
* https://github.com/dependabot/dependabot-core/pull/6149

jeffwidman:
Looks like this crept back in when `bun` was added... 
* https://github.com/dependabot/dependabot-core/pull/12655

My guess is that's because @kbukum1 copied from `npm` ecosystem, and we hadn't yet removed `"master"` from there, as this PR was only merged two days ago:
* https://github.com/dependabot/dependabot-core/pull/7131

But once I get bun handled, we should be good to go once and for all.

PR #7131, opened 2023-04-21
Remove more hardcoded `master` references

jeffwidman:
Follows up on https://github.com/dependabot/dependabot-core/pull/6149 by removing two more hardcoded references to `master` that I discovered.

Originally I assumed we'd need to flip these to `HEAD`, but in https://github.com/dependabot/dependabot-core/pull/6149 we were able to get away with simply removing the hardcoded reference altogether in the `bundler` ecosystem and we haven't seen any ill effects the past 6 months...

So let's try removing this and see if it works as smoothly in these ecosystems as it did in the `bundler` ecosystem.

Fix https://github.com/dependabot/dependabot-core/issues/6202


jeffwidman:
And two code failures... since these are two separate ecosystems, I'll likely break this into two separate PR's...

I suspect this indicates we'll need to still hardcode some value, but perhaps we can get away with `HEAD`, otherwise we'll have to have the `master` vs `main` conversation... 😩

abdulapopoola (approved):

abdulapopoola (approved):

hex/helpers/lib/parse_deps.exs
```diff
@@ -109,7 +109,7 @@ defmodule Parser do
     %{
       type: "git",
       url: repo_url,
-      branch: opts[:branch] || "master",
```
jeffwidman, line 112: @TylerWitt any idea why this might be failing? You seem to be a lot more knowledgable than me about Hex/Erlang... 

If we can figure out a fix, should probably split this out to it's own dedicated PR...

rickreyhsig (approved):

hex/helpers/lib/parse_deps.exs
TylerWitt: Sorry for the delay!

This might have been a temporary hex outage or other transient error. I cherry-picked this commit and the build worked fine. I'd retry CI.

jeffwidman: No worries, thanks for trying--your timing is impeccable... you probably tried it just a few days after it was fixed 😁 :
* https://github.com/dependabot/dependabot-core/pull/12109

jeffwidman:
Now that the Hex fix was handled in a separate PR, I've rebased this and updated the PR / commit description to reflect that the scope of this is now only for `npm`.

v-kbukum1:
@jeffwidman , 

Do we want to merge this PR? If we want, I can approve it and deploy/check/merge it.

jeffwidman:
Yes, this one is ready to go. Thank you!

kbukum1 (approved):
LGTM

v-kbukum1:
@jeffwidman , 

After pipeline succeed, I will deploy that and if there is no issue then merge it.

kbukum1 (approved):

PR #7134, opened 2023-04-21
Improve corepack setup

deivid-rodriguez:
~We're already using corepack to activate the version of Yarn used by each project through the `packageManager` field in their `package.json` file.~

~However, downloading and activating the project's version on the fly could fail sometimes, for example, in some firewalled environments.~

~This PR takes a more explicit approach of explicitly activating the project's version with `corepack`, but if that fails, proceed with the global version.~

~NOTE: I included a few other refactorings and improvements with the PR.~

This PR has been repurposed. Its original purpose was 👆, however, we have never heard any reports of corepack failing to activate versions, so I don't want to complicate our implementation fixing a non issue.

However, the other corepack improvements I made here seemed nice, so I kept them and iterated on them. This is a summary of what's in here right now:

* Improve our corepack setup, and make corepack be used consistently by all package managers.
* Make Dependabot use PNPM 7 for lockfile 5.4 format, since that will result in better diffs. We use corepack to activate that version when necessary.
* Whenever a package manager is explicitly configured through `packageManager` setting in the `package.json` file, we only consider dependency files for that package manager and ignore the others if present. This fixes errors like `bin/dry-run.rb npm_and_yarn "internet-identity-labs/nfid"`, which results in `Usage Error: This project is configured to use npm` when trying to run `pnpm` on it.

### DIFF
--- a/npm_and_yarn/Dockerfile
+++ b/npm_and_yarn/Dockerfile
@@ -1,5 +1,8 @@
 FROM ghcr.io/dependabot/dependabot-updater-core
 
+# Check for updates at https://github.com/nodejs/corepack/releases
+ARG COREPACK_VERSION=0.24.0
+
 # Check for updates at https://github.com/pnpm/pnpm/releases
 ARG PNPM_VERSION=8.11.0
 
@@ -22,6 +25,7 @@ RUN mkdir -p /etc/apt/keyrings \
   && apt-get install -y --no-install-recommends \
   nodejs \
   && rm -rf /var/lib/apt/lists/* \
+  && npm install -g corepack@$COREPACK_VERSION \
   && npm install -g npm@$NPM_VERSION \
   && rm -rf ~/.npm
 
--- a/npm_and_yarn/Dockerfile
+++ b/npm_and_yarn/Dockerfile
@@ -32,10 +32,10 @@ RUN mkdir -p /etc/apt/keyrings \
 USER dependabot
 
 # Install pnpm and set it to a stable version
-RUN corepack prepare pnpm@$PNPM_VERSION --activate
+RUN corepack install pnpm@$PNPM_VERSION --global
 
 # Install yarn berry and set it to a stable version
-RUN corepack prepare yarn@$YARN_VERSION --activate
+RUN corepack install yarn@$YARN_VERSION --global
 
 ENV DEPENDABOT_NATIVE_HELPERS_PATH="/opt"
 COPY --chown=dependabot:dependabot npm_and_yarn/helpers /opt/npm_and_yarn/helpers
--- a/npm_and_yarn/Dockerfile
+++ b/npm_and_yarn/Dockerfile
@@ -26,7 +26,6 @@ RUN mkdir -p /etc/apt/keyrings \
   nodejs \
   && rm -rf /var/lib/apt/lists/* \
   && npm install -g corepack@$COREPACK_VERSION \
-  && npm install -g npm@$NPM_VERSION \
   && rm -rf ~/.npm
 
 USER dependabot
@@ -36,6 +35,9 @@ RUN corepack install pnpm@$PNPM_VERSION --global
 
 # Install yarn berry and set it to a stable version
 RUN corepack install yarn@$YARN_VERSION --global
+
+# Install npm and set it to a stable version
+RUN corepack install npm@$NPM_VERSION --global
 
 ENV DEPENDABOT_NATIVE_HELPERS_PATH="/opt"
 COPY --chown=dependabot:dependabot npm_and_yarn/helpers /opt/npm_and_yarn/helpers
--- a/npm_and_yarn/Dockerfile
+++ b/npm_and_yarn/Dockerfile
@@ -29,8 +29,6 @@ RUN mkdir -p /etc/apt/keyrings \
   && npm install -g npm@$NPM_VERSION \
   && rm -rf ~/.npm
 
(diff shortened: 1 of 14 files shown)


npm_and_yarn/Dockerfile
```diff
@@ -35,9 +37,6 @@ ENV NODE_EXTRA_CA_CERTS=/etc/ssl/certs/ca-certificates.crt
 
 # END: HACKY WORKAROUND FOR NPM GIT INSTALLS SPAWNING CHILD PROCESS
 
-# Our native helpers pull in yarn 1, so we need to reset the version globally
```
jeffwidman, line 38: Is this statement no longer true? Because the new ordering in this PR never resets the version after the native helpers run...

jeffwidman (approved):
Given the unrelated changes, if it were me, I'd probably extract into a couple of different PR's,  but I also understand the convenience of doing a single PR... 

But for merging, this one should _not_ be squashed, too much going on in here, better to keep the commits separate.

npm_and_yarn/Dockerfile
deivid-rodriguez: Yeah, I think this statement was never true.

Only thing that should override a `corepack prepare <package_manager>@<version> --activate` should be a different `corepack prepare <package_manager>@<version> --activate` command, which I don't think installing native helpers does.

I think the original problem here was that the first `corepack prepare <package_manager>@<version> --activate` was being run as root, so the corepack binstubs were being installed in a location that the dependabot user won't use. That's why we needed to rerun the command again as the dependabot user.

I think I can extract this change to a separate PR to make sure it's good.

abdulapopoola (approved):

npm_and_yarn/Dockerfile
jeffwidman: This landed in https://github.com/dependabot/dependabot-core/pull/7160

coolbaluk:
👋 

Do you expect to merge this one ?

I'm struggling to get dependabot to run with yarn v4 as it seems to fallback to v3, wondering if these changes would help.

deivid-rodriguez:
Hi @coolbaluk, sorry for the delay in answering.

We should already be running yarn 4 and by your comment at https://github.com/dependabot/dependabot-core/issues/7962#issuecomment-1727969074, you already fixed your problem. But just clarify that this PR should keep existing behavior regarding yarn.

npm_and_yarn/lib/dependabot/npm_and_yarn/package_manager.rb
```diff
@@ -1,19 +1,59 @@
...
+
+        version = requested_version(name)
+
+        if version
+          install(name, version)
+        else
+          version = guessed_version(name)
+
+          if version && name == "pnpm"
+            if Version.new(version.to_s) < Version.new("7")
```
jurre, line 26: Is there any way the requested_version could be < 7? Should we account for this there as well?

deivid-rodriguez: Yes, and I think this set of changes is unintentionally restoring support for that. We should definitely be guarding and raising in both cases, I'll fix this.

bdragon (approved):
Sorry if this is misplaced, but since we're making improvements to the nodejs ecosystem, and specifically the `npm_version_numeric` helper method, this feels at least relevant.

I've never loved the fact that our treatment of npm versions has evolved to reflect the history of our codebase rather than that of the tool and its lockfile formats. For historical reasons our support has evolved to have an npm6/npm8 split, and we use the `lockfileVersion` as a proxy to determine whether we should use one version or the other. The real pivot in the lockfile format occurred in npm7.

`lockfileVersion` 1 indicates the lockfile was generated by npm5 or npm6.

`lockfileVersion` 2 or 3 indicates the lockfile was generated by npm7 or later, the difference being that `lockfileVersion` 2 is backwards compatible to `lockfileVersion` 1, while `lockfileVersion` 3 is not backwards compatible.

Currently `lockfileVersion` 3 is only used in the [hidden `node_modules/.package-lock.json` lockfile](https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json#hidden-lockfiles) that is generated as of npm7. In a future version of npm, once support for npm6 has been dropped, the `lockfileVersion` will likely change from 2 to 3.

In places where we need to pivot handling depending on whether we're dealing with one lockfile format or another, we use the npm version as a proxy and we call out the npm version with an implicit bias for one version or the other, for example `run_npm_updater` (assumes npm6) vs. `run_npm8_updater`, or with the organization of our helper scripts under `npm/` (assumes npm8) and `npm6/`.

I'd love to see the pivots in our code reflect the real differences that necessitate the pivot like `lockfileVersion` 1 vs. `lockfileVersion` 2, rather than superficial ones like npm6 vs. npm8. Or if we're going to pivot on npm versions, maybe call out the legacy ones by version and treat the latest bundled version as the default.

To be clear, I don't see any reason not to make the improvements proposed here.

deivid-rodriguez:
Thanks for the input, @bdragon, your ideas make total sense to me.

Here I just tried to keep things as they work today, because I didn't really had any insights on the history & evolution of NPM versions & lockfile formats until now :)

deivid-rodriguez:
To be clear, this PR does change the NPM version to be used if explicitly requested by users (through `packageManager` field in `package.json` file).

But also fixing our pivoting around lockfile formats when no explicit version is requested would make sense to me!

PR #7136, opened 2023-04-21
[Updater] Ensure we no longer test the legacy code path

brrygrdn:
Follows on from https://github.com/dependabot/dependabot-core/pull/7128

Now that all well-formed job definitions are routed to an Operation class, the only jobs that can possibly hit the `Updater#legacy_run` path are those where the attributes **do not match real-world configurations**.

This PR makes a change to disable the `Updater#legacy_run` method in our test suite so any existing or new tests which hit the legacy code line will fail.

Having done this, it then addresses two cohorts of failing tests:
- A set of ignore condition tests which are creating new Version PRs but are using `job.dependencies` as a convenience in their test setup to avoid having to account for N dependencies
  - This fix for these is to adjust the expectations so they only examine the target dependency without using this hook
- A set of tests which target creating a new PR for a _specific_ dependency that were not security updates
  - For this group the tests were mostly checking the same behaviour for updating as for grouping, so I updated them to be security updates to reflect real world conditions and they continue to pass
  - In one case, testing that a nominated subdependency would still result in an update, I moved this into the "Updating a PR" test as this is the scenario we are most likely to encounter and handle this problem
  - Finally, one test outright checks we can create a fresh PR for a nominated single Dependency, which is behaviour we do not actually use in the real world. I've allowed this to continue passing by enabling the legacy code, but it will be removed when we clean up.

### Rationale

The overall goal of the Operation classes is to only have concrete code lines for the job configurations we run in the real world, this ultimately means removing a single capability expressed in our tests at present which is to create a fresh Version Update PR for a set of specific dependencies.

This capability definitely seems like something we might want in future, but in keeping with our intent to make the code paths as explicit as possible I think it makes sense to remove it until we need it rather than leave ambiguity.

As a follow-on from the grouped update introduction and as part of the cleanup from this task the `updater_spec` should be broken up to more thoroughly test the concrete classes which will make it much easier to bring this back with _exhaustive_ testing rather than having one example in the file currently touch this path.

### DIFF
--- a/updater/lib/dependabot/updater.rb
+++ b/updater/lib/dependabot/updater.rb
@@ -95,6 +95,8 @@ module Dependabot
     # This is the original logic within run, we currently fail over to this if
     # no Operation class exists for the given job.
     def legacy_run
+      raise Dependabot::NotImplemented unless Environment.legacy_run_enabled?
+
       service.increment_metric("updater.started", tags: { operation: "Legacy" })
       if job.updating_a_pull_request?
         Dependabot.logger.info("Starting PR update job for #{job.source.repo}")
--- a/updater/spec/dependabot/updater_spec.rb
+++ b/updater/spec/dependabot/updater_spec.rb
@@ -1383,6 +1383,11 @@ RSpec.describe Dependabot::Updater do
       # This will be deprecated when `legacy_run` is removed but should be
       # fairly trivial to bring back if required.
       context "and the job is create a version PR" do
+        before do
+          # Permit the legacy_run method to be used
+          allow(Dependabot::Environment).to receive(:legacy_run_enabled?) { true }
+        end
+
         it "only attempts to update dependencies on the specified list" do
           stub_update_checker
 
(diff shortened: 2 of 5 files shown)


updater/spec/dependabot/updater_spec.rb
```diff
@@ -1361,184 +1407,157 @@ def expect_update_checker_with_ignored_versions(versions)
...
+                }
+              ],
+              security_updates_only: true,
               updating_a_pull_request: false
             )
             service = build_service
             updater = build_updater(service: service, job: job)
 
-            expect(service).to_not receive(:close_pull_request)
+            expect(service).to receive(:create_pull_request)
```
jakecoffman, line 1431: Seems like a good idea!

updater/lib/dependabot/updater.rb
```diff
@@ -95,6 +95,8 @@ def run
     # This is the original logic within run, we currently fail over to this if
     # no Operation class exists for the given job.
     def legacy_run
+      raise Dependabot::NotImplemented unless Environment.legacy_run_enabled?
```
jakecoffman, line 98: The other day I was wondering why we had this exception. Nice use of it!

jakecoffman (approved):
Make sense!

PR #7137, opened 2023-04-21
[Updater] DependencyChange passes any group to the PR message builder

brrygrdn:
In order to ensure the PR title and message body are built using the right strategy, we need to pass any group assigned to the `Dependabot::DependencyChange` on to the `Dependabot::PullRequestCreator::MessageBuilder` class.

I've also made a small agreement change to the name of the placeholder group we use if no configuration is supplied to use the same placeholder name we use elsewhere.

### DIFF
--- a/updater/spec/dependabot/dependency_change_spec.rb
+++ b/updater/spec/dependabot/dependency_change_spec.rb
@@ -99,10 +99,36 @@ RSpec.describe Dependabot::DependencyChange do
           files: updated_dependency_files,
           dependencies: updated_dependencies,
           credentials: job_credentials,
-          commit_message_options: commit_message_options
+          commit_message_options: commit_message_options,
+          dependency_group: nil
         )
 
       expect(dependency_change.pr_message).to eql("Hello World!")
+    end
+
+    context "when a dependency group is assigned" do
+      it "delegates to the Dependabot::PullRequestCreator::MessageBuilder with the group included" do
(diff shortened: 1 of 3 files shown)


brrygrdn:
Smoke test re-alignment here: https://github.com/dependabot/smoke-tests/pull/57

updater/spec/dependabot/dependency_change_spec.rb
```diff
@@ -99,11 +99,37 @@
...
+          to receive(:new).with(
+            source: github_source,
+            files: updated_dependency_files,
+            dependencies: updated_dependencies,
+            credentials: job_credentials,
+            commit_message_options: commit_message_options,
+            dependency_group: group
+          )
+
+        expect(dependency_change.pr_message).to eql("Hello World!")
```
landongrindheim, line 130: 🔎 When I read this, I see that the assertion is being made against the mock in the [test setup](https://github.com/dependabot/dependabot-core/pull/7137/files#diff-6254ba183349892acb9556d882f751be30988df63e1949923569dcbb3bb2219cR84-R93). Am I reading that right?

If I am reading that correctly, I'm concerned that we'd be getting a false sense of confidence from the test. I think we might be better off just asserting that the instance was created with the correct arguments. What do you think?

landongrindheim (approved):

bdragon (approved):

PR #7138, opened 2023-04-21
Adding the version catalog file to the update file regex

### LINKED ISSUE #7105: PR isn't closed after cherry-picking Dependabot commit

### Package ecosystem

Gradle

### Package manager version

8.1

### Language version

_No response_

`build.gradle.kts`:

```kotlin
plugins {
    alias(libs.plugins.kotlin)
}

version = "1.0.0"
group = "bar"

repositories {
    mavenCentral()
}
```

`gradle/libs.versions.toml`:

```toml
[versions]
kotlin = "1.8.10"

[plugins]
kotlin = { id = "org.jetbrains.kotlin.jvm", version.ref = "kotlin" }
```

### dependabot.yml content

```
version: 2
updates:
  - package-ecosystem: "gradle"
    directory: "/"
    schedule:
      interval: "daily"
```

### Updated dependency

_No response_

I have initiated the repository with a single Kotlin plugin dependency in the `libs.versions.toml` file with an outdated version set: `1.8.10`.
After receiving the PR, I have cherry-picked Dependabot's commit with `1.8.20` version and pushed it to the `main` branch.

Unfortunately, Dependabot did not close the PR.

_No response_

https://github.com/hsz/dependabot-pr-not-closed-after-satisfying-update-request/pull/1

commit already cherry-picked to `main` here:
https://github.com/hsz/dependabot-pr-not-closed-after-satisfying-update-request/commit/8b45a0136195fcf2b53bde0d87e2ebfe01dc236b

https://github.com/hsz/dependabot-pr-not-closed-after-satisfying-update-request

pavera:
Fixes #7105.

We aren't closing PRs when they involve the version catalog file. We use this list of regex to compare against changes in push events to determine if we should close a PR.


deivid-rodriguez (approved):

jeffwidman (approved):
Is there an existing unit test of this code that should also be updated? Otherwise how hard would it be to add one?

hsz:
Out of the curiosity — you consider the `gradle.properties` file too, right?
Because I didn't see it in the modified line.

jeffwidman:
I'm not sure if we look at the `gradle.properties` file for anything... https://github.com/dependabot/dependabot-core/issues?q=is%3Aissue+is%3Aopen+gradle.properties shows an open ticket related to that file... and I don't know if there are any other dep versions configured in that file that would be relevant from a `dependabot-core` perspective?

But if we do (and you would probably know more than me), then we should also include it in this regex of "triggering checking whether we should close an open PR"... If so, let us know (preferably via a new ticket so that we don't lose track of it)

hsz:
I've updated my [test repo](https://github.com/hsz/dependabot-pr-not-closed-after-satisfying-update-request) to use `gradle.properties` and read the dependency version from it with:

https://github.com/hsz/dependabot-pr-not-closed-after-satisfying-update-request/blob/63827d3775eb818f998d4828e4bfe6822290ad0d/build.gradle.kts#L1

Dependabot ignored a dependency that relies on a variable:

```
implementation("junit:junit:${junitVersionProvider.get()}")
```

I'm not sure if that's expected, but that makes considering the `gradle.properties` file obsolete. 🙃

PR #7141, opened 2023-04-21
Don't run update checker unnecessarily on all javascript packages in a monorepo

deivid-rodriguez:
If a monorepo provides multiple packages, but a single root lockfile, we just need to run the update checker in the root lockfile, since that one should be managing all dependencies.

We were running the update checker once per package, making things painfully slow.

Also, we were running the commands to check for the update inside each folder, and I'm not sure that actually works as expected in presence of a root lockfile.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/dependency_files_filterer.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/dependency_files_filterer.rb
@@ -14,6 +14,10 @@ module Dependabot
         @updated_dependencies = updated_dependencies
       end
 
+      def paths_requiring_update_check
+        @paths_requiring_update_check ||= fetch_paths_requiring_update_check
+      end
+
       def files_requiring_update
         @files_requiring_update ||=
           dependency_files.select do |file|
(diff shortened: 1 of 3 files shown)


npm_and_yarn/lib/dependabot/npm_and_yarn/dependency_files_filterer.rb
```diff
@@ -34,6 +38,14 @@ def package_files_requiring_update
 
       attr_reader :dependency_files, :updated_dependencies
 
+      def fetch_paths_requiring_update_check
+        return [File.dirname(root_lockfile.name)] if lockfiles == [root_lockfile]
```
jeffwidman, line 42:
```suggestion
        # if a root lockfile exists, it tracks all dependencies
        return [File.dirname(root_lockfile.name)] if lockfiles == [root_lockfile]
```

jeffwidman (commented):
> If a monorepo provides multiple packages, but a single root lockfile, we just need to run the update checker in the root lockfile, since that one should be managing all dependencies.

I lack context on this scenario... if a root lockfile exists, why would subfolders even have lockfiles at all?

And won't it cause problems if the root lockfile exists but the subfolder lockfiles become woefully out of date?

Is this the expectation of users in these ecosystems, that they just ignore these sub-lockfiles altogether?

Or will somehow if we check/update the deps in the root lockfile that the native helper (`npm`, `yarn`, etc) will handle updating the sub-lockfiles in the subfolders to stay in sync with the root lockfile?

npm_and_yarn/lib/dependabot/npm_and_yarn/dependency_files_filterer.rb
deivid-rodriguez: I'll add a comment, but I'll use "if ONLY a lockfile exists, it tracks all dependencies" to make it a bit more clear.

jeffwidman (approved):
I missed that this is the only lockfile listed in that array. And the comment will be helpful to clarify for others like me who are moving too fast and missed the details. 😄 

Makes perfect sense. 👍 

🚢

PR #7142, opened 2023-04-21
Cargo: handle dependencies that are not in the index

jakecoffman:
Currently when Dependabot encounters an error from Cargo when it can't find a dependency it raises a HelperSubprocessFailed error. 

This recategorizes that case to a DependencyFileNotResolvable error which shows a better error including the dependency that was not found.

I randomly generated the dependency name in the test so no one can register the name and break our tests 😄

### DIFF
--- a/cargo/spec/dependabot/cargo/file_updater/lockfile_updater_spec.rb
+++ b/cargo/spec/dependabot/cargo/file_updater/lockfile_updater_spec.rb
@@ -105,6 +105,31 @@ RSpec.describe Dependabot::Cargo::FileUpdater::LockfileUpdater do
       end
     end
 
+    context "when the dependency doesn't exist" do
+      random_unlikely_package_name = (0...255).map { ("a".."z").to_a[rand(26)] }.join
+      content = <<~CONTENT
+        [package]
+        name = "foo"
+        version = "0.1.0"
+        authors = ["me"]
+
+        [dependencies]
+        #{random_unlikely_package_name} = "99.99.99"
+      CONTENT
+
+      let(:manifest) do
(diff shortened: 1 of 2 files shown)


cargo/spec/dependabot/cargo/file_updater/lockfile_updater_spec.rb
```diff
@@ -105,6 +105,31 @@
...
+    context "when the dependency doesn't exist" do
+      random_unlikely_package_name = (0...255).map { ("a".."z").to_a[rand(26)] }.join
+      content = <<~CONTENT
+        [package]
+        name = "foo"
+        version = "0.1.0"
+        authors = ["me"]
+
+        [dependencies]
+        #{random_unlikely_package_name} = "99.99.99"
```
jeffwidman, line 117: clever way to do this!

jeffwidman (approved):
This is great.

One question:
Will this block the rest of the job? If so, that's fine if `cargo` blocks/fails the whole run--we can't do more than they do... 

but if `cargo` will throw this error and then continues processing the rest of the packages, we'd want to check if this error results in :dependabot: mirroring the `cargo` behavior...

jakecoffman:
@jeffwidman I think you're asking if updating other dependencies will also result in this same error? I'm not sure if that is the case.

Unfortunately there's only a handful of [halting errors](https://github.com/dependabot/dependabot-core/blob/8f3342a4028a31eee80cc0982884eabeae0189d0/updater/lib/dependabot/updater/error_handler.rb#L22-L29), 
so there's not a way to end a job once it has got to the part where it is updating dependencies.

So I haven't really looked into it because there's no way to stop the update without throwing a different error yet. We might revisit in the future and add a `should_halt?` to all of the errors.

PR #7143, opened 2023-04-21
Cargo: handle packages that don't exist at Git repo

jakecoffman:
In Cargo if you specify the package name, it validates that the Git repo you point to has a package at that location with the same name. If it doesn't it displays a message "no matching package found."

Dependabot was throwing a HelperSubprocessFailed message for this case. This changes it to a better message to point the user to the issue, and cleans up our error reporting.

### DIFF
--- a/composer/helpers/v1/composer.lock
+++ b/composer/helpers/v1/composer.lock
@@ -1920,16 +1920,16 @@
         },
         {
             "name": "phpstan/phpstan",
-            "version": "1.10.13",
+            "version": "1.10.14",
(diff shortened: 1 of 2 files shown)


brrygrdn (approved):
🚀

jakecoffman:
@brrygrdn Mind taking a quick look at https://github.com/dependabot/dependabot-core/pull/7142? It's basically the same PR, different error message.

jeffwidman (approved):

PR #7110, opened 2023-04-18
Run UpdateAllVersions with ungrouped dependencies

Nishnha:
Run UpdateAllVersions with ungrouped dependencies if any are present.

Here is what the output for a grouped update looks like when it also creates PRs for its ungrouped dependencies:
```
+------------------------------------------------------------------------------------------------------------------------------------+
|                                                Changes to Dependabot Pull Requests                                                 |
+---------+--------------------------------------------------------------------------------------------------------------------------+
| created | @emotion/react ( from 11.10.5 to 11.10.6 ), @emotion/styled ( from 11.10.5 to 11.10.6 )                                  |
| created | gatsby ( from 5.6.0 to 5.9.0 ), gatsby-cli ( from 5.6.0 to 5.9.0 ), gatsby-plugin-emotion ( from 8.6.0 to 8.9.0 ), ga... |
| created | @reduxjs/toolkit ( from 1.9.2 to 1.9.5 )                                                                                 |
| created | axios ( from 1.3.2 to 1.3.5 )                                                                                            |
| created | graphql-http ( from 1.14.0 to 1.18.0 )                                                                                   |
| created | react-router ( from 6.8.1 to 6.10.0 )                                                                                    |
| created | react-router-dom ( from 6.8.1 to 6.10.0 )                                                                                |
| created | web-vitals ( from 2.1.4 to 3.3.1 )                                                                                       |
| created | @babel/cli ( from 7.20.7 to 7.21.0 )                                                                                     |
| created | @babel/core ( from 7.20.12 to 7.21.4 )                                                                                   |
| created | @babel/eslint-parser ( from 7.19.1 to 7.21.3 )                                                                           |
| created | @babel/parser ( from 7.20.15 to 7.21.4 )                                                                                 |
| created | @babel/preset-env ( from 7.20.2 to 7.21.4 )                                                                              |
| created | @babel/preset-typescript ( from 7.18.6 to 7.21.4 )                                                                       |
| created | @testing-library/react ( from 13.4.0 to 14.0.0 )                                                                         |
| created | @types/node ( from 17.0.45 to 18.15.11 )                                                                                 |
| created | @types/react ( from 18.0.27 to 18.0.37 )                                                                                 |
| created | @types/react-dom ( from 18.0.10 to 18.0.11 )                                                                             |
| created | @typescript-eslint/eslint-plugin ( from 5.51.0 to 5.59.0 )                                                               |
[...]

PR #7159, opened 2023-04-24
v0.217.0

pavera:
## v0.217.0, 24 April 2023

- Run UpdateAllVersions with ungrouped dependencies [#7110](https://github.com/dependabot/dependabot-core/pull/7110)
- [Updater] Ensure we no longer test the legacy code path [#7136](https://github.com/dependabot/dependabot-core/pull/7136)
- [Updater] Extract RefreshSecurityUpdatePullRequest into an Operation class [#7128](https://github.com/dependabot/dependabot-core/pull/7128)
- [Updater] DependencyChange passes any group to the PR message builder [#7137](https://github.com/dependabot/dependabot-core/pull/7137)

### DIFF
--- a/updater/Gemfile.lock
+++ b/updater/Gemfile.lock
@@ -322,4 +322,4 @@ DEPENDENCIES
   webmock (~> 3.18)
 
 BUNDLED WITH
-   2.3.26
+   2.4.11
--- a/updater/Gemfile.lock
+++ b/updater/Gemfile.lock
@@ -201,7 +201,7 @@ GEM
       rake (~> 13.0)
     mime-types (3.4.1)
       mime-types-data (~> 3.2015)
-    mime-types-data (3.2023.0218.1)
+    mime-types-data (3.2022.0105)
     mini_mime (1.1.2)
     mini_portile2 (2.8.1)
     multi_xml (0.6.0)
--- a/updater/Gemfile.lock
+++ b/updater/Gemfile.lock
@@ -1,19 +1,19 @@
 PATH
   remote: ../bundler
   specs:
-    dependabot-bundler (0.216.2)
-      dependabot-common (= 0.216.2)
+    dependabot-bundler (0.217.0)
+      dependabot-common (= 0.217.0)
 
 PATH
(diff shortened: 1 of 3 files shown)


updater/Gemfile.lock
```diff
@@ -322,4 +322,4 @@ DEPENDENCIES
   webmock (~> 3.18)
 
 BUNDLED WITH
-   2.4.11
+   2.3.26
```
jeffwidman, line 325: This looks off?

* https://github.com/dependabot/dependabot-core/pull/7046

Even though you ran this from a `2.3.26` `bundler` version, I suspect it'll be okay if you simply manually put it back to `2.4.11` w/o having to re-run with the newer `bundler` version... 🙉 🙈

pavera: Good catch :)

updater/Gemfile.lock
```diff
@@ -201,7 +201,7 @@ GEM
       rake (~> 13.0)
     mime-types (3.4.1)
       mime-types-data (~> 3.2015)
-    mime-types-data (3.2022.0105)
+    mime-types-data (3.2023.0218.1)
```
jeffwidman, line 204: Why did this change? 🤔 

None of the changes landed in this release bump any deps AFAICT.

Unless it's the issue that the updater script doesn't actually pickup all the PRs in the changelog... part of why switching to GitHub releases... 

(and 🙇‍♂️ for cutting this release to get it out, even though a little more painful right now)

jeffwidman (approved):
Approving to unblock, as the one thing I spotted seems harmless, just surprising.

jeffwidman:
Also, can you please merge via squash on this? Since these two commits really should land as one...

updater/Gemfile.lock
pavera: I'm going to revert this change as well, I suspect this is due to me running the update on my mac and not in a codespace.

pavera: I did merge a number of dependabot PRs, but I didn't think any to do with mime-types.

PR #7160, opened 2023-04-24
Prepare yarn berry just once

deivid-rodriguez:
I think the problem was not that native helpers pull Yarn 1, but that the version was being activated for the root user, not for the dependabot user. I don't think we need it for root.

(Extracted from #7134)


jeffwidman (approved):
Nice cleanup and thanks for extracting this out.

PR #7162, opened 2023-04-24
Cleanup leftover require

deivid-rodriguez:
I think this is a leftover from c0f7545d441045ce81ce6f74608d274272841812.


jeffwidman (approved):
Thank you for cleaning up what I missed. 😄

Issue #7132, opened 2023-04-21
Publish the Dependabot ruby gems more frequently (weekly??)


jeffwidman:
### Code improvement description

This ticket is a planning/tracking ticket for publishing the Dependabot gems more frequently to RubyGems.

### Context:
We used to publish releases about once a week, which were then deployed to the GitHub Dependabot service. A few months ago, we switched to deploying straight from PR's to the GitHub service... this allowed us to test code live in production before merging to `main`, ensuring that released code has already been validated in production.

However, because we no longer had a forcing function for publishing the gems to RubyGems, we never got around to it... and several months passed between releases. 

We want to get back to a regular publishing cadence, as we've got many users who use the `dependabot-core` library outside of the GitHub service, and they are some of our most active contributors.

### Plan:
In order to do this, we need to streamline the release process so it requires less maintainer time.

My current thought is the following:
1. Streamline the script for bumping the PR versions. Today it has some unnecessary complexity making it brittle (https://github.com/dependabot/dependabot-core/issues/5777).
2. Have a GitHub action that leverages this script to automate opening a PR to bump the version. This could happen on the cadence we plan to cut the releases, or it could open the next one as soon as the previous one is merged.
3. Switch the release notes from a Changelog text file to using GitHub releases. Otherwise we have a race condition where if another PR gets merged before the version bump PR, it won't be included in the changelog...
4. When a maintainer is ready, they merge the version bump PR, then tag the release using GitHub releases.
5. A GitHub action watches for new tags and pushes them to RubyGems. (this already happens today)

#### Edge cases:
* Need to cut an emergency release: A human can either run the script for bumping versions, or run the github action that normally runs on cron.
* Want a major version bump: we'd probably just do this manually.

### Decisions:
* What frequency to release? I think weekly is probably about right, but anywhere from daily to monthly could work. We're open to input from the community.
* What versioning scheme to use? Calver, semver, or a hybrid? We're open to input, but we also have some internal constraints related to the Dependabot service at GitHub that aren't fully visible that may influence the outcome here.
* Do we fully automate cutting releases, or should we keep a human involved with merging the version bump PR and double-checking the changelog? This influences the versioning conversation.

### Other notes:
* If we stick with full semver, then probably we'd need the github action in 2️⃣ above to take an input arg of "major/minor/patch" and a human would need to trigger it.
[...]

PR #7163, opened 2023-04-24
Switch from `Changelog.md` to GitHub releases

jeffwidman:
We'd like to publish release more often, and part of that is making the release process more streamlined so that the maintainers don't dread the busywork of cutting a release.

Part of that is automating creating PR's that bump the version... but there's a race condition because we don't keep the changelog routinely updated with every PR... so if the version bump PR is opened, and then another PR is merged, that other PR won't be mentioning in the release notes.

The easiest way to sidestep that is to switch to managing the release notes via GitHub releases and stop using a Changelog text file.

We've had this discussion of "flat file vs git tags / GitHub releases" on a number of open source projects that I help maintain, and the pros/cons typically boil down to GitHub releases are more convenient, but they are more difficult to grep offline and less portable for projects that are concerned about lock-in to GitHub.

:dependabot: is already tightly connected to GitHub so I don't think that's a concern. :joy:

And :dependabot: is built around the concept of internet access, so I don't foresee much need to be grep'ing the changelog while offline... Searching for a release right from the releases page ([example](https://github.com/vishvananda/netns/releases?q=watch+github&expanded=true)) will generally be good enough, and advanced use cases can still grep the git tags content.

There's more streamlining to come, but this is a good incremental step to move us forward.

Related:

### DIFF
--- a/bin/bump-version.rb
+++ b/bin/bump-version.rb
@@ -8,50 +8,6 @@ end
 component = ARGV[0].to_sym
 dry_run = ARGV[1] == "--dry-run"
 
-# rubocop:disable Lint/LiteralAsCondition
-unless `which gh` && $?.success?
-  puts "Please install the gh cli: brew install gh"
-  exit 1
-end
-
-unless `gh auth status -h github.com > /dev/null 2>&1` && $?.success?
-  puts "Please login to GitHub first: gh auth login"
-  exit 1
-end
-# rubocop:enable Lint/LiteralAsCondition
-
-CHANGELOG_PATH = File.join(__dir__, "..", "CHANGELOG.md")
-CHANGELOG_CONTENTS = File.read(CHANGELOG_PATH)
-
-def proposed_changes(version, _new_version)
-  dependabot_team = `gh api -X GET 'orgs/dependabot/teams/maintainers/members' --jq '.[].login'`
-  dependabot_team = dependabot_team.split("\n").map(&:strip) + ["dependabot"]
-
-  commit_subjects = `git log --pretty="%s" v#{version}..HEAD`.lines
-  merge_subjects = commit_subjects.select do |s|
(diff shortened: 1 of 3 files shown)


bin/bump-version.rb
```diff
@@ -80,37 +36,23 @@ def proposed_changes(version, _new_version)
...
-    proposed_changes.join("\n") + "\n",
-    CHANGELOG_CONTENTS
-  ].join("\n")
-
-  File.write(CHANGELOG_PATH, new_changelog_contents)
-  puts "☑️  CHANGELOG.md updated"
-end
-
 unless dry_run
+  puts "Now, create the PR"
```
jeffwidman, line 40: To be clear: I'll put up a PR shortly to automate creation of this PR, but wanted to keep that separate from this PR, which simply switches us to GitHub releases.

jakecoffman (approved):
LGTM

One thing I am curious about is with `gh release create v1.X.X --generate-notes --draft` does the draft release create the tag immediately or does that happen after the draft is published? If it happens after publish, it might be nice to automatically run `gh release create` with the correct version argument to avoid accidentally typing the wrong tag version. 

If I am understanding correctly we're going to have to manually type out the version ID. 

Otherwise let's try it!

bin/bump-version.rb
jakecoffman: Ope I missed this.

jeffwidman:
Good point... I'll try to consider that when automating it.

I personally prefer automation to only create a draft release instead of a real release because that way the human can edit the changelog before publishing... and yes, we can do so after publishing, but it's cleaner to do so before hand.

But we could pre-generate the `gh` command to use including the correct version in the GitHub Action workflow summary... 

I wonder if the URL for creating a new release can deep-link to having the tag included as a HTTP query param? I'll look into it.

PR #7164, opened 2023-04-24
No need for `--dry-run`

jeffwidman:
Now that this script no longer generates the changelog, it's so simple that there's no need for `--dry-run`. Any file changes can be reverted using `git restore`.

I thought about leaving it in for development purposes, but anyone working on this can hack what they need by commenting code out.

So might as well simplify it, esp since this will be used within a GitHub action workflow soon, simpler is better.


jakecoffman (approved):
Simple is better!

PR #7166, opened 2023-04-25
Pass the serialized group name when creating a PR for grouped updates

Nishnha:
Pass the serialized group name to PR creator (and the API client) when creating a PR for grouped updates.

We no longer pass the `grouped-update` key to the PR creator and instead pass in the `dependency-group` name

For now we are transforming the group name to downcase, snake_case, and to a symbol (although the test checks the name after it's been serialized to JSON, so the DependencyChange#id is converted to a string)

### DIFF
--- a/common/lib/dependabot/dependency_group.rb
+++ b/common/lib/dependabot/dependency_group.rb
@@ -4,17 +4,28 @@ require "wildcard_matcher"
 
 module Dependabot
   class DependencyGroup
-    attr_reader :name, :rules, :dependencies
+    attr_reader :name, :rules, :dependencies, :id
 
     def initialize(name:, rules:)
       @name = name
       @rules = rules
       @dependencies = []
+      @id = id_hash
     end
 
     def contains?(dependency)
       @dependencies.include?(dependency) if @dependencies.any?
       rules.any? { |rule| WildcardMatcher.match?(rule, dependency.name) }
+    end
+
+    private
+
+    def id_hash
+      { "name" => serialized_group_name }
+    end
+
+    def serialized_group_name
+      name.downcase.gsub("-", "_").to_sym
     end
   end
 end
--- a/common/lib/dependabot/dependency_group.rb
+++ b/common/lib/dependabot/dependency_group.rb
@@ -4,13 +4,12 @@ require "wildcard_matcher"
 
 module Dependabot
   class DependencyGroup
-    attr_reader :name, :rules, :dependencies, :id
+    attr_reader :name, :rules, :dependencies
 
     def initialize(name:, rules:)
       @name = name
       @rules = rules
       @dependencies = []
-      @id = id_hash
     end
 
     def contains?(dependency)
@@ -18,11 +17,11 @@ module Dependabot
       rules.any? { |rule| WildcardMatcher.match?(rule, dependency.name) }
     end
 
-    private
-
-    def id_hash
+    def to_h
       { "name" => serialized_group_name }
     end
+    
+    private
 
     def serialized_group_name
       name.downcase.gsub("-", "_").to_sym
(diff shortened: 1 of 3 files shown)


common/lib/dependabot/dependency_group.rb
```diff
@@ -4,17 +4,28 @@
...
     def contains?(dependency)
       @dependencies.include?(dependency) if @dependencies.any?
       rules.any? { |rule| WildcardMatcher.match?(rule, dependency.name) }
     end
+
+    private
+
+    def id_hash
+      { "name" => serialized_group_name }
+    end
```
brrygrdn, line 25:
```suggestion
    def to_h
      { "name" => serialized_group_name }
    end
    
    private
```

I think it would be more idiomatic to just make this a type-casting/serialise method and avoid bringing in the term `id` at all in the event we find we do need it future.

updater/spec/dependabot/api_client_spec.rb
```diff
@@ -175,22 +175,25 @@
...
 
     context "grouped updates" do
-      it "does not include the grouped_update key by default" do
+      it "does not include the grouped-update or dependency-group keys by default" do
         client.create_pull_request(dependency_change, base_commit)
 
         expect(WebMock).
           to(have_requested(:post, create_pull_request_url).
              with do |req|
                expect(req.body).not_to include("grouped-update")
```
brrygrdn, line 184:
```suggestion
```

I think we can just replace all knowledge of this key with `dependency-group`

updater/spec/dependabot/api_client_spec.rb
```diff
@@ -175,22 +175,25 @@
     end
 
     context "grouped updates" do
-      it "does not include the grouped_update key by default" do
+      it "does not include the grouped-update or dependency-group keys by default" do
```
brrygrdn, line 178:
```suggestion
      it "does not include the dependency-group key by default" do
```

updater/spec/dependabot/api_client_spec.rb
```diff
@@ -199,7 +202,11 @@
           to(have_requested(:post, create_pull_request_url).
              with do |req|
                data = JSON.parse(req.body)["data"]
-               expect(data["grouped-update"]).to be true
+               # FIXME: Once we fully remove the `group-all` prototype behavior, we can also remove the
+               # `grouped-update` expectation since dependency-groups will require rules and only
+               # return the `dependency-group` key
+               expect(data["grouped-update"]).to be nil
```
brrygrdn, line 208: Hmm, this comment suggests should we still be sending this key as `true` if a group is involved?

I don't think we need to care about this key anymore even if the placeholder is involved though. I believe the API behaviour will set the `grouped-update` flag on its side if a Dependency Group is sent, even if that group isn't in the database. This should result in jobs carried out with a placeholder that was either injected by the API or by the Updater itself having the same behaviour.

brrygrdn (commented):
A couple of questions about removing the `grouped_update` key from the payload ( which I think is fine ) and moving away from using the term `id`.

Nishnha:
I'll update the smoke-tests to use the new dependency-group key instead of grouped-update

Nishnha:
Opened https://github.com/dependabot/smoke-tests/pull/61 for updating the smoke tests. The CLI changes need to be deployed first though

common/lib/dependabot/dependency_group.rb
```diff
@@ -4,17 +4,28 @@
...
+
+    def serialized_group_name
+      name.downcase.gsub("-", "_").to_sym
+    end
```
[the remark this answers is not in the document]
brrygrdn: > in the API we use #parameterize and here we do something similar but not exactly the same.

Yeah, good catch, removing this is the right way to go thanks for the fix!

brrygrdn (approved):

PR #7168, opened 2023-04-25
Cargo: handle unsupported toolchain versions (part 2)

jakecoffman:
This is the same PR as #7122 but against the version resolver. 

It would probably be good in the future to create a native helper that both the version resolver and lockfile updater use since the errors seem like they are the same between the two.

### DIFF
--- a/cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
+++ b/cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
@@ -237,6 +237,10 @@ module Dependabot
             return nil
           end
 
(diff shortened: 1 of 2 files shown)


jeffwidman (approved):

jeffwidman:
> It would probably be good in the future to create a native helper that both the version resolver and lockfile updater use since the errors seem like they are the same between the two.

Why not wrap the calls to `cargo` in a custom ruby class / function that handles common errors? And use that both places?

Native helpers can be useful some places, but IMO the simplicity of using the package manager's CLI is tough to beat, especially because users are generally using it as the reference for expected behavior and we don't have to futz around with building, keeping internal API's updated, etc.

Issue #6993, opened 2023-04-03
Update the description on the confusing `dependabot-core` gem from RubyGems


jeffwidman:
### Code improvement description

This gem currently exists, but hasn't been updated since 2019: https://rubygems.org/gems/dependabot-core

And the description currently says:
> Automated dependency management for Ruby, JavaScript, Python, PHP, Elixir, Rust, Java, .NET, Elm and Go

However, this is NOT the gem we actually use, instead we use:
* https://rubygems.org/gems/dependabot-omnibus
* https://rubygems.org/gems/dependabot-common
* https://rubygems.org/gems/dependabot-<ecosystem-specific>

So the presence of the `dependabot-core` gem creates confusion for users.

When I raised the issue in our weekly team sync, most of the team didn't even know this existed, and everyone was strongly in favor of deleting it to reduce confusion.

I'm creating this ticket to document that we're nuking it.

jeffwidman:
Actually, reading the docs at https://help.rubygems.org/kb/gemcutter/removing-a-published-rubygem reminds me that we don't want to leave this namespace unoccupied or we risk someone name-squatting on it.

Better to put up an empty gem with a placeholder description pointing at `dependabot-omnibus`.

PR #7171, opened 2023-04-25
Add a placeholder `dependabot-core.gemspec`

jeffwidman:
While auditing [our gems](https://rubygems.org/profiles/dependabot), I noticed there was a `dependabot-core` gem that hadn't been updated in years: https://rubygems.org/gems/dependabot-core/versions/0.94.13

This was probably the original gem before it was all split apart into the meta gem `dependabot-omnibus` and the subgems.

This creates confusion because the name matches the `dependabot-core` repo.

At first I planned to delete it, but then I saw the note on Rubygems that deleting a gem frees up the namespace. We would much rather own this namespace to prevent namesquatting, so I've created a placeholder gemspec that has no associated source code.

I wasn't sure if `gem` would allow building/publishing with no source code, so tested, and it all worked:

```shell
$ gem build dependabot-core.gemspec
WARNING:  no files specified
WARNING:  description and summary are identical
WARNING:  See https://guides.rubygems.org/specification-reference/ for help
  Successfully built RubyGem
  Name: dependabot-core
  Version: 0.95.1
  File: dependabot-core-0.95.1.gem
$ gem push dependabot-core-0.95.1.gem
Pushing gem to https://rubygems.org...
Successfully registered gem: dependabot-core (0.95.1)
```

You can see the updated gem here: https://rubygems.org/gems/dependabot-core/versions/0.95.1

Which is much less confusing than it used to be.

I think we should keep the gemspec around in case anyone needs to make small changes at some point, no need for them to recreate the whole file.

This repo seemed like the best fit, I wasn't sure whether to float it in `./common`, so I stuck it in the root dir... seems harmless enough.

Fix https://github.com/dependabot/dependabot-core/issues/6993

### DIFF
--- a/dependabot-core.gemspec
+++ b/dependabot-core.gemspec
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+# This is a placeholder gem to prevent namesquatting on https://rubygems.org/gems/dependabot-core
+# Any updates must be manually published to RubyGems.
+# It's excluded from the normal release process because it's not expected to change.
+
+Gem::Specification.new do |spec|
+  spec.name = "dependabot-core"
+  spec.summary      = "This is a placeholder gem to prevent namesquatting. " \
+                      "You probably want the gem 'dependabot-omnibus'."
+  spec.description  = "This is a placeholder gem to prevent namesquatting. " \
+                      "You probably want the gem 'dependabot-omnibus'."
+  spec.authors     = "Dependabot"
+  spec.email       = "opensource@github.com"
+  spec.files       = [] # intentionally empty, this is a placeholder gem to prevent namesquatting
+  spec.homepage    = "https://github.com/dependabot/dependabot-core"
+  spec.license = "Nonstandard" # License Zero Prosperity Public License
+
+  spec.metadata = {
+    "bug_tracker_uri" => "https://github.com/dependabot/dependabot-core/issues"
+  }
+
+  spec.version = "0.95.1"
+  # Since a placeholder gem, no need to keep `required_ruby_version` up to date with `.ruby-version`
+  spec.required_ruby_version = ">= 3" # rubocop:disable Gemspec/RequiredRubyVersion
+end
--- a/Rakefile
+++ b/Rakefile
@@ -10,6 +10,8 @@ require "bundler"
 require "./common/lib/dependabot"
 require "yaml"
 
+# ./dependabot-core.gemspec is purposefully excluded from this list
+# because it's an empty gem as a placeholder to prevent namesquatting.
 GEMSPECS = %w(
   common/dependabot-common.gemspec
   go_modules/dependabot-go_modules.gemspec


dependabot-core.gemspec
```diff
@@ -0,0 +1,20 @@
+# This is a placeholder gem to prevent namesquatting on https://rubygems.org/gems/dependabot-core
+# Any updates must be manually published to RubyGems.
+# It's excluded from the normal release process because it's not expected to change.
+
+Gem::Specification.new do |spec|
+    spec.name        = "dependabot-core"
+    spec.version     = "0.95.1"
```
jeffwidman, line 7: The last published version before this was `0.94.13`, so this probably should be `1.0.0` since it's a breaking change to drop all source.

However, I wanted to test to see if this would actually work w/o `gem build`/`gem publish` complaining about lack of source files, so I used this version and well, it worked and it's live. 🤷‍♂️

dependabot-core.gemspec
jeffwidman, line 1: We could put this in a subfolder, but the only one that might make sense is `common`, and that seems a bit buried? Plus confusing... today it wouldn't muck with our release automation: https://github.com/dependabot/dependabot-core/blob/17ab50090e14ffdf8efc3da01c0a4814d3d116dd/Rakefile#L13

But that could change in the future... so root dir this seems like the least bad place to put it... 

That or could simply not commit it at all, but I think it's better to have a placeholder file here so we sorta don't forget about it and so someone doesn't have to recreate this down the road...

abdulapopoola:
Great find!

dependabot-core.gemspec
```diff
@@ -0,0 +1,25 @@
...
+  spec.files       = [] # intentionally empty, this is a placeholder gem to prevent namesquatting
+  spec.homepage    = "https://github.com/dependabot/dependabot-core"
+  spec.license = "Nonstandard" # License Zero Prosperity Public License
+
+  spec.metadata = {
+    "bug_tracker_uri" => "https://github.com/dependabot/dependabot-core/issues"
+  }
+
+  spec.version = "0.95.1"
+  spec.required_ruby_version = ">= 3.1.0"
```
jeffwidman, line 24: It's dumb to include this `required_ruby_version`, but:
1. `rubocop` complains if I don't. 
2. So I tried setting it to `3`, but then `rubocop` complained that it wasn't as high as https://github.com/dependabot/dependabot-core/blob/17ab50090e14ffdf8efc3da01c0a4814d3d116dd/.ruby-version#L1
3. I tried disabling just that particular `rubocop` warning, but I couldn't figure out the magic incantation despite [checking the docs](https://docs.rubocop.org/rubocop/configuration.html#disabling-cops-within-source-code).

The unfortunate bit is that we'll end up having to bump this line _every_ time we bump `ruby`, which is trivial, but it makes the file look more active in `git blame` than it really is... and may lead to someone thinking they need to publish an updated version each time.

So I'd really rather figure out how to silence `rubocop` on this...

jeffwidman: Got it with a little more googling... basically I have to keep the key, but I can prevent it from needing to be bumped to match `.ruby-version`:
```ruby
spec.required_ruby_version = ">= 3" # rubocop:disable Gemspec/RequiredRubyVersion
```

dependabot-core.gemspec
```diff
@@ -0,0 +1,26 @@
...
+  spec.email       = "opensource@github.com"
+  spec.files       = [] # intentionally empty, this is a placeholder gem to prevent namesquatting
+  spec.homepage    = "https://github.com/dependabot/dependabot-core"
+  spec.license = "Nonstandard" # License Zero Prosperity Public License
+
+  spec.metadata = {
+    "bug_tracker_uri" => "https://github.com/dependabot/dependabot-core/issues"
+  }
+
+  spec.version = "0.95.1"
```
landongrindheim, line 23: I understand this gemspec will not really do much, but I think it makes sense to point to the current version. Note: 👇 will necessitate adding `require "./lib/dependabot"` to the top of this file.

```suggestion
  spec.version = Dependabot::VERSION
```

landongrindheim (approved):

dependabot-core.gemspec
jeffwidman: I don't mind doing this at all, but curious why you think it's an improvement?

And thank you for the Ruby pointer on the `require`!

abdulapopoola (approved):

dependabot-core.gemspec
landongrindheim: > I'm happy to do this, but curious why you think it's an improvement? Since it's disconnected from the rest of our code... and also won't be published regularly to RubyGems, so the version bumps would be "jarring" perhaps?

> Only benefit I see is being consistent with how we handle versions in the other gems.

My thoughts are that it signals the existence of a newer version. I'm positive you've thought more deeply about this than me, so please feel free to proceed as you see fit 😄

jeffwidman: Thanks Landon, I appreciate that.

Circling back on this with fresh eyes. I personally think as a drive-by dev, if I see a hardcoded version, and it maps to the latest version on https://rubygems.org/gems/dependabot-core, then that's a lot easier to understand than if we've got a version here in core that is on `0.200'ish` but rubygems is still way back on `0.95.1`... Obviously we could then publish the new one, but they'll still inevitably drift over time since we don't plan to publish this gem on every release of `dependabot-core`.

So as the versions drift, I start to wonder if we simply haven't published the gem and wonder how much drift there's been in the code from what's on rubygems. So making it clear they're in sync via the hardcoded version seems like the least confusing option for now.

So I'm going to merge as-is for now.

Obviously if we ever start using this gem, we should immediately switch to using `Dependabot::VERSION`.

PR #7177, opened 2023-04-26
Improve npm instrumentation

deivid-rodriguez:
I'm still not totally sure that we should introduce the approach in #7134, but the little refactorings there should be good to introduce separately.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
@@ -40,6 +40,10 @@ module Dependabot
       end
 
       def self.yarn_major_version
+        @yarn_major_version ||= fetch_yarn_major_version
+      end
+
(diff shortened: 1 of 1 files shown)


npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -153,11 +153,13 @@ def inferred_npmrc # rubocop:disable Metrics/PerceivedComplexity
...
 
-        if locked_version = package_manager.locked_version("yarn")
-          locked_version
-        elsif yarn_lock
-          Helpers.yarn_version_numeric(yarn_lock)
-        end
+        package_manager.locked_version("yarn") || guess_yarn_version
+      end
+
+      def guess_yarn_version
```
jeffwidman, line 159: now that is a function name that gives me confidence in what it does! 😁

jeffwidman (approved):

PR #7178, opened 2023-04-26
Pass exception message directly to peer dependency error handler

deivid-rodriguez:
Extracted from #7081 as suggested by @jeffwidman!


jeffwidman:
Merging this to keep it moving forward so #7081 can be rebased on top of it...

PR #7179, opened 2023-04-26
Clarify the `images-latest` workflow

jeffwidman:
This took me a little longer than it should have to understand what was going on, so here's a few cleanup that I'd have found helpful:
1. Rename `TAG` to `COMMIT_SHA` as that's what it is... and tag is a bit ambiguous because the image is actually getting tagged with multiple things.
2. Organize all the tagging steps together.
3. Move the login step right next to publish... total nit, but typical workflows are build, tag, publish, and login is  really part of publishing.


jeffwidman:
Oops, meant to push this against my fork for testing...

PR #7180, opened 2023-04-26
Clarify the `images-latest` workflow

jeffwidman:
This took me a little longer than it should have to understand what was going on, so here's a few cleanups that would have helped me grok things faster:
1. Rename `TAG` to `COMMIT_SHA` because tag is a bit ambiguous because the image is actually getting tagged with multiple things.
2. Organize all the tagging steps together.
3. Move the login step right next to publish... total nit, but typical workflows are build, tag, publish, and login is really part of publishing.
4. Clarify that we're _only_ building the ecosystem images, as `dependabot-updater` is ambiguous since there's both the base core image and the ecosystem images.

### DIFF
--- a/.github/workflows/images-latest.yml
+++ b/.github/workflows/images-latest.yml
@@ -51,34 +51,32 @@ jobs:
           - { name: python, ecosystem: pip }
           - { name: terraform, ecosystem: terraform }
     env:
-      TAG: ${{ github.sha }}
+      COMMIT_SHA: ${{ github.sha }}
       NAME: ${{ matrix.suite.name }}
       ECOSYSTEM: ${{ matrix.suite.ecosystem }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
 
-      - name: Build dependabot-updater image
+      - name: Build the dependabot-updater-<ecosystem> image
+        # despite the script input being $NAME, the resulting image is dependabot-updater-${ECOSYSTEM}
         run: script/build ${NAME}
 
+      - name: Tag the images with the SHA, `latest`, and the date version
+        run: |
+          docker tag "${UPDATER_IMAGE}${ECOSYSTEM}" "${UPDATER_IMAGE}${ECOSYSTEM}:$COMMIT_SHA"
+          docker tag "${UPDATER_IMAGE}${ECOSYSTEM}" "${UPDATER_IMAGE}${ECOSYSTEM}:latest"
+          docker tag "${UPDATER_IMAGE}${ECOSYSTEM}" "${UPDATER_IMAGE}${ECOSYSTEM}:${{ needs.date-version.outputs.date }}"
+
       - name: Log in to GHCR
-        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/images-latest.yml
```diff
@@ -51,34 +51,32 @@ jobs:
...
-      TAG: ${{ github.sha }}
+      COMMIT_SHA: ${{ github.sha }}
       NAME: ${{ matrix.suite.name }}
       ECOSYSTEM: ${{ matrix.suite.ecosystem }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
 
-      - name: Build dependabot-updater image
+      - name: Build the dependabot-updater-<ecosystem> image
```
jeffwidman, line 61: Unfortunately step names don't have access to the env vars, otherwise I'd have used `dependabot-updater-${ECOSYSTEM}` in the name... I tried. 😄

.github/workflows/images-latest.yml
```diff
@@ -51,34 +51,32 @@ jobs:
...
 
-      - name: Log in to GHCR
+      - name: Tag the images with the SHA, `latest`, and the date version
         run: |
-          echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
+          docker tag "${UPDATER_IMAGE}${ECOSYSTEM}" "${UPDATER_IMAGE}${ECOSYSTEM}:$COMMIT_SHA"
+          docker tag "${UPDATER_IMAGE}${ECOSYSTEM}" "${UPDATER_IMAGE}${ECOSYSTEM}:latest"
+          docker tag "${UPDATER_IMAGE}${ECOSYSTEM}" "${UPDATER_IMAGE}${ECOSYSTEM}:${{ needs.date-version.outputs.date }}"
 
-      - name: Push branch image
```
jeffwidman, line 68: I suspect this allusion to branch name was a copy/paste leftover since we're _not_ on the branch, but rather on `main`...

jakecoffman (approved):

.github/workflows/images-latest.yml
```diff
@@ -51,34 +51,32 @@ jobs:
...
-        run: |
-          docker tag "${UPDATER_IMAGE}${ECOSYSTEM}" "${UPDATER_IMAGE}${ECOSYSTEM}:$TAG"
-          docker push "${UPDATER_IMAGE}${ECOSYSTEM}:$TAG"
+      - name: Log in to GHCR
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
 
-      - name: Push latest on main
-        run: |
-          docker tag "${UPDATER_IMAGE}${ECOSYSTEM}:$TAG" "${UPDATER_IMAGE}${ECOSYSTEM}:latest"
```
jeffwidman, line 75: I don't understand why this was using `"${UPDATER_IMAGE}${ECOSYSTEM}:$TAG"` rather than `"${UPDATER_IMAGE}${ECOSYSTEM}"` when declaring the image to tag... since they're effectively the same thing here. 🤷‍♂️

PR #7190, opened 2023-04-28
Prevent links in PR description that link to redirect.redirect.github.com

stefangr:
The different sections of the metadata go through the LinkAndMentionSanitizer twice.

Once per sub section and once as the complete message.

In case of source providers that support HTML that is not a problem. The Markdown is converted to HTML and after that left alone.

In case of a source provider that does not support HTML (like bitbucket) there is a problem.
The links in the markdown are converted from github.com to redirect.github.com. And in the second pass the links are converted from redirect.github.com to redirect.redirect.github.com.

This PR contains a fix so only the second pass will convert the links in case the source provider does not support HTML.

### DIFF
--- a/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb
@@ -134,6 +134,8 @@ module Dependabot
         end
 
         def replace_github_host(text)
+          return text if !github_redirection_service.nil? && text.include?(github_redirection_service)
+
           text.gsub(
             /(www\.)?github.com/, github_redirection_service || "github.com"
           )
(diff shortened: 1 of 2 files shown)


jeffwidman (changes requested):
What you're trying to do makes complete sense. 👍 

Unfortunately, this solution is tracking state across function calls, and that can be hard to reason about and gets brittle quickly.

From a quick look at `LinkAndMentionSanitizer`, it looks like the `github_redirection_service` is the only URL that gets rewritten.

Would it be possible to instead modify the method that does the rewrite so that if the URL already matches the rewritten URL that it won't attempt another rewrite?

https://github.com/dependabot/dependabot-core/blob/221a6e5599ea1e8c30f6167b0a1bbe48c0b8e15f/common/lib/dependabot/pull_request_creator/message_builder/link_and_mention_sanitizer.rb#L136-L140

That would be a bit more robust and keep all the "rewriting" logic constrained within that method.

jeffwidman (approved):
This looks so much more maintainable, thanks!

PR #7192, opened 2023-04-28
[Updater] Implement an Operation capable of refreshing a Grouped Update PR.

brrygrdn:
"Refreshing" a Dependabot pull request essentially has three distinct outcomes, we recompute the DependencyChange on the current head of the target branch and:

- determine that all the same dependencies change to the same versions as the existing PR; in this case we update the existing PR
- determine that one or more dependencies are now involved or removed; in this case we close the existing PR and create a new one
- determine that all the dependencies are the same, but one or more are now updating to a different version; in this case we create a new PR and leave it to the backend service to detect the new PR supersedes and existing one

This strategy is consistent with our normal single-dependency PRs with one exception; instead of looking at _all_ existing PRs to determine whether it is updating/replacing/superseding it only considers PRs for the group rule that job is targeting.

### Example output

```
updater | +---------------------------------------------------------------------------------------------------------------------------------------------------------+
updater | |                                                           Changes to Dependabot Pull Requests                                                           |
updater | +------------------------------+--------------------------------------------------------------------------------------------------------------------------+
updater | | closed: dependencies_changed | com.bugsnag:bugsnag,com.google.inject:guice,io.github.lukehutch:fast-classpath-scanner,net.kyori:text,net.lingala.zip... |
updater | | created                      | com.google.inject:guice ( from 4.2.0 to 5.1.0 ), net.lingala.zip4j:zip4j ( from 1.3.2 to 2.11.5 ), net.kyori:text ( f... |
updater | +------------------------------+--------------------------------------------------------------------------------------------------------------------------+
```

### Significant changes

This PR introduces two new attributes to the "job definition"

- `existing_group_pull_requests`: this serialises any grouped PRs separately to 'normal' PRs so we don't have unexpected interactions between single- and group-updates
  - In future, we may merge this into the `existing_pull_requests` key for simplicity, but it would be a breaking change so we should do it incrementally.
- `dependency-group-to-refresh`: the name of the dependency group to be refreshed
  - As part of this process, we confirm that a group by this name still exists in the configuration for the repository, failing if not


updater/lib/dependabot/dependency_snapshot.rb
```diff
@@ -53,6 +53,19 @@ def job_dependencies
       end
     end
 
+    def job_group_name
+      job.dependency_group_to_refresh
+    end
```
brrygrdn, line 58: My thinking is that in future we might have a `dependency_group_to_create` key to allow us to just build a single group if we want to change the workload profile of group/version updates.

For now it might make sense to remove this indirection but I wanted to shorten the reference in a couple of places

updater/lib/dependabot/updater/group_update_creation.rb
```diff
@@ -0,0 +1,239 @@
...
+#
+# When included in an Operation it expects the following to be available:
+# - job: the current Dependabot::Job object
+# - dependency_snapshot: the Dependabot::DependencySnapshot of the current
+#   repo state
+# - error_handler: a Dependabot::UpdaterErrorHandler to report
+#
+module Dependabot
+  class Updater
+    module GroupUpdateCreation
```
brrygrdn, line 16: I considered two alternative approaches:
- Let's make a GroupUpdate base class!
- Let's create a new GroupUpdaterBuilder object we instantiate and call!

I ended up 👎🏻 on both of those as I'd rather not add inherence into the Updater given the problems we've had with branching code and I wasn't totally sold on the value of a `GroupUpdaterBuilder` object as a separate component we might want to add tests on specifically.

I ended up defaulting to a coarse composition approach with a view that we can revisit this later as there's still some DRYing out and test improvement to be done on the other Operations and the right pattern is likely to become clearer with more examples.

brrygrdn:
PR to make the CLI compatible: https://github.com/dependabot/cli/pull/105

updater/lib/dependabot/updater/group_update_creation.rb
```diff
@@ -0,0 +1,239 @@
...
+      # Returns a Dependabot::DependencyChange object that encapsulates the
+      # outcome of attempting to update every dependency iteratively which
+      # can be used for PR creation.
+      def compile_all_dependency_changes_for(group)
+        all_updated_dependencies = []
+        updated_files = []
+        group.dependencies.each do |dependency|
```
brrygrdn, line 23: @Nishnha I just want to call out this change so it doesn't get washed out in the move, I changed this line from

```ruby
dependencies.each do |dependency|
            next unless group.contains?(dependency)
```

In my testing, the two sets were always the same so it semantically made more sense to me to use `group.dependencies` and it also made this method agnostic of whether we just have a target group or several groups.

Nishnha: Yep this would be identical. After the dependency groups are calculated (which is a side effect of calling DependencySnapshot#job_groups) we can just refer to `group.dependencies`

updater/lib/dependabot/dependency_change.rb
```diff
@@ -46,5 +46,51 @@ def updated_dependency_files_hash
...
+    # rather than supersede it as the new changes don't necessarily follow
+    # from the previous ones; dependencies could have been removed from the
+    # project, or pinned by other changes.
+    def should_replace_existing_pr?
+      return false unless job.updating_a_pull_request?
+
+      # NOTE: Gradle, Maven and Nuget dependency names can be case-insensitive
+      # and the dependency name injected from a security advisory often doesn't
+      # match what users have specified in their manifest.
+      updated_dependencies.map(&:name).map(&:downcase) != job.dependencies.map(&:downcase)
```
Nishnha, line 63: job.dependencies is a list of Dependabot::Dependency objects, so I think you would also need to map across the job dependency names

updater/spec/fixtures/job_definitions/bundler/version_updates/group_update_refresh.yaml
```diff
@@ -0,0 +1,45 @@
...
+    prefix:
+    prefix-development:
+    include-scope:
+  security-updates-only: false
+  dependency-groups:
+  - name: everything-everywhere-all-at-once
+    rules:
+      patterns:
+        - "*"
+  dependency-group-to-refresh: everything-everywhere-all-at-once
```
Nishnha, line 45: <3

updater/lib/dependabot/dependency_change.rb
```diff
@@ -46,5 +46,51 @@ def updated_dependency_files_hash
...
+    def existing_pull_request
+      if grouped_update?
+        # We only want PRs for the same group that have the same versions
+        job.existing_group_pull_requests.find do |pr|
+          pr["dependency-group-name"] == dependency_group.name &&
+            Set.new(pr["dependencies"]) == updated_dependencies_set
+        end
+      else
+        job.existing_pull_requests.find { |pr| Set.new(pr) == updated_dependencies_set }
+      end
```
Nishnha, line 81: We may want to break out the "existing grouped update PR?" check into its own method that we can call in `#upsert_pull_request`

Users could define a group that has a single dependency and matches an existing ungrouped PR, causing `dependency_change.matches_existing_pr?` to return true.
If that happens, then we would close the existing ungrouped dependency PR in favor of a grouped one (which could be okay?)

If we introduce a `#matches_existing_grouped_pr?` method then we can make sure we only upsert grouped updates in the refresh grouped update PR operation

updater/spec/dependabot/updater/operations/refresh_group_update_pull_request_spec.rb
```diff
@@ -0,0 +1,234 @@
...
+      updated_bundler_files
+    end
+
+    before do
+      stub_rubygems_calls
+    end
+
+    it "closes the pull request" do
+      expect(mock_error_handler).not_to receive(:handle_dependabot_error)
+      expect(mock_service).to receive(:close_pull_request).with(["dummy-pkg-b"], :up_to_date)
```
Nishnha, line 99: I doubled checked this and dummy-pkg-b is the only dependency in the "updated" gemfiles, but when using the "original" gemfiles dummy-pkg-c also updates as a transitive dependency 👍🏾 

https://github.com/dependabot/dependabot-core/blob/ded459a52682161823aa9b94904790fb156f6af9/updater/spec/fixtures/bundler2/updated/Gemfile

updater/spec/dependabot/updater/operations/refresh_group_update_pull_request_spec.rb
```diff
@@ -0,0 +1,234 @@
...
+      expect(mock_service).to receive(:capture_exception).with(
+        error: an_instance_of(Dependabot::DependabotError),
+        job: job
+      )
+
+      group_update_all.perform
+    end
+  end
+
+  context "when the target dependency group has stale values for its rules" do
```
Nishnha, line 211: What makes the dependency group in this job stale?

updater/lib/dependabot/dependency_change.rb
brrygrdn: This caught me out too, `dependency_snapshot.job_dependencies` are hydrated `Dependabot::Dependency` objects, but `job.dependencies` is just the [raw array from the job definition file](https://github.com/dependabot/dependabot-core/blob/9faebe9751245ea96a801007d86b96264eecd027/updater/spec/fixtures/job_definitions/bundler/version_updates/group_update_refresh.yaml#L10-L11).

We should probably do a naming cleanup pass on this to make that more obvious as I mocked tests incorrectly and then had a surprise when I tried it with the CLI.

updater/spec/dependabot/updater/operations/refresh_group_update_pull_request_spec.rb
brrygrdn: Ah, this is a rogue test I can remove - this was necessary when `dependency-group-to-refresh` sent the full dependency group object to ensure we used the right `rules` value.

Since we only send a name, it doesn't matter anymore so we can drop this test.

updater/lib/dependabot/dependency_change.rb
brrygrdn: > Users could define a group that has a single dependency and matches an existing ungrouped PR, causing dependency_change.matches_existing_pr? to return true. 

This shouldn't happen at present as `grouped_updated?` on line 73 would be true for the grouped update meaning it wouldn't check the `existing_pull_requests` set for any matches, so it should keep both job types in their own lane.

We might want to consider doing a one-time check between strategies when users change their config to add a group ( i.e. close any old single PRs that match the new group ), but I think we can leave that for now and allow it to be managed manually so it's ok not to do any cross-checking and do a hard fork on `grouped_updated?` being true/false for the current job.

jeffwidman:
> In future, we may merge this into the `existing_pull_requests` key for simplicity, but it would be a breaking change so we should do it incrementally.

FWIW, I wouldn't be worried about any breaking changes in the `updater`, no one uses that beyond us. And for us, keep in mind grouped updates are still alpha/beta... breaking isn't the end of the world. Only if break normal "single" pull requests would this really matter IMO...

Issue #4072, opened 2021-07-22
`.npmrc` with `"engine-strict=true"` rule throws `EBADENGINE` error


websolutions-hamburg:
**Package ecosystem**
npm
**Package manager version**
npm 7.20.0
**Language version**
node 16.4.0
**dependabot.yml content**
```yml
version: 2
updates:
  - package-ecosystem: 'npm'
    directory: '/'
    schedule:
      interval: 'daily'
      time: '05:00'
      timezone: 'Europe/Berlin'
    commit-message:
      prefix: 'npm'
      include: 'scope'
    labels:
      - 'Dependabot :robot:'
    versioning-strategy: 'increase'
```
**What you expected to see, versus what you actually saw**
In our `package.json` we use the engines entry:

```json
{
  "engines": {
    "node": ">=16.4.0",
    "npm": ">=7.20.0"
  }
}
```

And in our `.npmrc` we set `engine-strict=true`.

The problem is now, that Dependabot currently uses `npm v7.19.1` and `node v14.17.2` and runs into an `EBADENGINE` error.

Is it possible to tell Dependabot to ignore the `engine-strict=true` flag?

**Log Output**

```
updater | INFO <job_174034416> Updating globby from 11.0.4 to 12.0.0
updater | I, [2021-07-22T10:00:09.990507 #9]  INFO -- sentry: ** [Raven] Sending event 3809c73b99ca4e1a922a494b9a850108 to Sentry
  proxy | 2021/07/22 10:00:10 [164] POST https://sentry.io:443/api/1451818/store/
  proxy | 2021/07/22 10:00:10 [164] 200 https://sentry.io:443/api/1451818/store/
updater | ERROR <job_174034416> Error processing globby (Dependabot::SharedHelpers::HelperSubprocessFailed)
updater | ERROR <job_174034416> npm WARN using --force Recommended protections disabled.
updater | <job_174034416> npm ERR! code EBADENGINE
updater | <job_174034416> npm ERR! engine Unsupported engine
updater | <job_174034416> npm ERR! engine Not compatible with your version of node/npm: undefined
updater | <job_174034416> npm ERR! notsup Not compatible with your version of node/npm: undefined
updater | <job_174034416> npm ERR! notsup Required: {"node":">=16.4.0","npm":">=7.20.0"}
updater | <job_174034416> npm ERR! notsup Actual:   {"npm":"7.19.1","node":"v14.17.2"}
```

ApolloLB:
[...]

PR #7194, opened 2023-04-28
NPM: handle EBADENGINE better

jakecoffman:
For additional context see https://github.com/dependabot/dependabot-core/issues/4072#issuecomment-1413222679

When `engine-strict` is set to true, Dependabot can error if the currently installed version doesn't match the range set in package.json. 

This gives the error in the UI more substance to help the customer figure out why it happened and solve the problem if they choose. Also helps clean up our error reporting!


abdulapopoola (approved):
Great find and fix! Love the fact that we are cleaning up our errors with actionable info for our users!! Thanks.

PR #7195, opened 2023-04-28
Clarify comments on peer dependency error regexes

deivid-rodriguez:
This should make it more clear that these lines are not part of the errors themselves.

Suggested by @jeffwidman at https://github.com/dependabot/dependabot-core/pull/7081#discussion_r1176951467.


jeffwidman (approved):
❤️

PR #7199, opened 2023-04-28
Defer opening duplicate grouped PRs

Nishnha:
For Grouped Updates, if a PR already exists for the group we should just skip it.
It will be left up to the "rebase" job for that group to refresh it if required

For PRs that were created as part of a group, metadata about the PR in addition to the dependencies that it updated are accessible through the new `existing_group_pull_requests` attribute which serializes like so:

```
[
  { # PR 1
    dependency_group: { name: "production" },
    dependencies: [
      { dependency-name: excon, dependency-version: 0.99.0 },
      { dependency-name: factory_girl, dependency-version: 4.9.0 },
      { dependency-name: rails, dependency-version: 7.0.4.2 }
    ]
  },
  { # PR 2
    dependency_group: { name: "dev" },
    dependencies: [
      { dependency-name: rspec, dependency-version: 3.12.0 }
      { dependency-name: rubocop, dependency-version: 1.46.0 }
    ]
  },
]
```

For reviewers:
- does the test coverage look appropriate?
- does the shape of `existing_group_pull_request`, including the key type (symbol vs string), look correct?
- does the way the `existing_group_pull_request` hash is accessed make sense? Should we consider using `WithIndifferentAccess` OR standardizing on a single key type instead?


Nishnha:
I committed fc6d93a36b16de997cc733b7f8a637d21860fcb1 because I saw the bundler smoke test fail with a message like

```bash
updater | 2023/05/01 14:05:57 ERROR undefined method `each' for nil:NilClass
updater | 
updater |             each.
updater |             ^^^^
```

I think this is because `existing_group_pull_request` is not defined for the smoke test yet, and although the value should default to an empty list, the smoke test would still need to have the key available to default on.

Once this PR lands I will update the grouped smoke tests to include the `existing_group_pull_request` key

updater/lib/dependabot/updater/operations/group_update_all_versions.rb
```diff
@@ -105,6 +109,12 @@ def run_ungrouped_dependency_updates
           ).perform
         end
 
+        def pr_exists_for_dependency_group?(group)
+          job.existing_group_pull_requests.
+            each&.
+            any? { |pr| pr.dig(:dependency_group, "name") == group.name }
```
bdragon, line 115: I think this `each&` is extra—could be simplified to `job.existing_group_pull_requests.any? { |pr| ...`

Good callout about the hash keys! (One of ruby's warts IMO.) I _think_ only the top-level hash keys will be symbols and the nested ones will be strings. According to the [serializer in dependabot-api](https://github.com/github/dependabot-api/blob/87a3b46341558caa2bead51472a4d0b84932aba4/app/serializers/update_job_serializer_private.rb#L106-L116), I believe this would need to be `pr.dig("dependency_group_name")`, though.

updater/spec/dependabot/updater_spec.rb
```diff
@@ -2310,6 +2310,26 @@ def expect_update_checker_with_ignored_versions(versions, dependency_matcher: an
       expect(service).not_to receive(:create_pull_request)
       updater.run
     end
+
+    it "does not create a new pull request for a group if one already exists" do
+      job = build_job(
+        existing_group_pull_requests: [
+          {
+            dependency_group: { "name" => "group-b" },
```
bdragon, line 2318: Same comment as above about `dependency_group_name`.

brrygrdn (approved):
This looks good to me, we might need to hold on merging this until we get the rebase trigger wired in as it'll cause us to avoid creating any new grouped PRs for internal test repos right now 🤔 .

PR #7208, opened 2023-05-01
Catch `bundle lock` failures

jeffwidman:
If `bundle lock` failed for some reason, then the exit code wasn't preserved and the message about the gemfile being updated successfully was incorrectly printed.


mctofu (approved):

bin/bump-version.rb
```diff
@@ -28,6 +28,10 @@
 
 # Bump the updater's Gemfile.lock with the new version
 `cd updater/ && bundle lock`
+if $?.exitstatus
```
jeffwidman, line 31: oof: https://github.com/dependabot/dependabot-core/pull/7209

PR #7211, opened 2023-05-02
Create version bump PRs using a custom action

jeffwidman:
This continues the process of streamlining our releases by adding a custom GitHub Action workflow to automate creating version bump PR's.

For now, I intentionally left it as manually triggered to keep the review straightforward and let us try it for a week or two.

Once we're sure we're happy with this, then we plan to add a workflow trigger of a regular cron task to ensure we release frequently... probably on a weekly cadence.

Related:

### DIFF
--- a/.github/workflows/gems-bump-version.yml
+++ b/.github/workflows/gems-bump-version.yml
@@ -1,0 +1,71 @@
+name: Gems - Bump Version
+on:
+  workflow_dispatch:
+    inputs:
+      version_type:
+        description: "Version Type?"
+        required: true
+        type: choice
+        options:
+          - minor
+          - patch
+        default: "minor"
+
+jobs:
+  Create-PR-To-Bump-Dependabot-Gems-Version:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ruby/setup-ruby@v1 # bump-version.rb needs bundler
+
+      - name: Bump the version
+        run: |
+          NEW_VERSION=$(bin/bump-version.rb ${{ github.event.inputs.version_type }})
+          echo "New version is: $NEW_VERSION"
+          echo "NEW_VERSION=$NEW_VERSION" >> $GITHUB_ENV
+
+      - name: Configure the git user
+        run: |
+          git config user.name "github-actions[bot]"
+          # Specifying the full email allows the avatar to show up: https://github.com/orgs/community/discussions/26560
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+      - name: Create a branch and commit the changes
+        run: |
+          # Using an idempotent branch name ensures no duplicate PR's are created
+          # if the action is re-run before the previous PR is merged.
+          # The branch name is purposefully different from the release tag to
+          # avoid ambiguity when selecting git refs.
+          git checkout -b "bump-to-v${{ env.NEW_VERSION }}"
+          git add common/lib/dependabot.rb updater/Gemfile.lock
+          echo "Creating commit / PR linking to the releases notes URL."
+          echo "This URL will 404 until the release is actually tagged, which you should do as soon as the PR is merged."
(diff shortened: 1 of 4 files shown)


.github/workflows/gems-bump-version.yml
```diff
@@ -0,0 +1,71 @@
...
+  workflow_dispatch:
+    inputs:
+      version_type:
+        description: "Version Type?"
+        required: true
+        type: choice
+        options:
+          - minor
+          - patch
+        default: "minor"
```
jeffwidman, line 12: When triggered via Cron, it'll run using `minor` with the assumption that new features got added since the most recent release. If that's not true, at that point we can always manually trigger a `patch` release instead. But I doubt we'll even pay attention to it much.

.github/workflows/gems-bump-version.yml
```diff
@@ -0,0 +1,71 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      version_type:
+        description: "Version Type?"
+        required: true
+        type: choice
+        options:
+          - minor
+          - patch
```
jeffwidman, line 11: `major` is omitted simply because the `bump-version.rb` script doesn't currently support it.

TBH, I think that's just fine for now, major version bumps have never actually happened yet, so we can wait to add support until we actually need it.

jeffwidman (commented):
Annotating the changes. ✍️

.github/workflows/gems-bump-version.yml
```diff
@@ -0,0 +1,71 @@
...
+          echo "PR_URL=$PR_URL" >> $GITHUB_ENV
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set summary
+        run: |
+          echo ":rocket: PR created at URL: ${{ env.PR_URL }}" >> $GITHUB_STEP_SUMMARY
+          echo "" >> $GITHUB_STEP_SUMMARY
+          echo "After the PR is approved/merged, create a new release tagged as \`v${{ env.NEW_VERSION }}\`, _making sure to point it at the merge commit_:" >> $GITHUB_STEP_SUMMARY
+          echo "* You can do this via the web UI - use the \`Generate release notes\` button and then edit as needed: https://github.com/${{ github.repository }}/releases/new?tag=v${{ env.NEW_VERSION }}&title=v${{ env.NEW_VERSION }}" >> $GITHUB_STEP_SUMMARY
```
jeffwidman, line 64: I wish I could add a query param like `generate-notes=true`, but unfortunately that's [not one of the supported options](https://docs.github.com/en/enterprise-cloud@latest/repositories/releasing-projects-on-github/automation-for-release-forms-with-query-parameters).

.github/workflows/gems-bump-version.yml
```diff
@@ -0,0 +1,71 @@
...
+        run: |
+          # Using an idempotent branch name ensures no duplicate PR's are created
+          # if the action is re-run before the previous PR is merged.
+          # The branch name is purposefully different from the release tag to
+          # avoid ambiguity when selecting git refs.
+          git checkout -b "bump-to-v${{ env.NEW_VERSION }}"
+          git add common/lib/dependabot.rb updater/Gemfile.lock
+          echo "Creating commit / PR linking to the releases notes URL."
+          echo "This URL will 404 until the release is actually tagged, which you should do as soon as the PR is merged."
+          git commit -m "v${{ env.NEW_VERSION }}" -m "Release notes: https://github.com/${{ github.repository }}/releases/tag/v${{ env.NEW_VERSION }}"
```
deivid-rodriguez, line 43: I didn't know you could pass multiple `-m` flags, neat!

deivid-rodriguez (approved):
Looks good!

pavera (approved):
👍 Excited for these changes!

.github/workflows/gems-bump-version.yml
jeffwidman: I checked with the Repos team about potentially adding this, and it turns out that trying to implement it is more complex than it looks at first... it'd probably also slow down the speed of the web form, not to mention needing to handle a number of logic edge cases, etc. 

So not something they plan to do.

PR #7212, opened 2023-05-02
Remove Updater#legacy_run

brrygrdn:
We haven't seen any jobs hit this code path in over one week of observation, so let's complete the migration and remove this code.

This PR does not address refactoring the specs into per-operation files since that is off the critical path and we've generally been treating gaps tests as no-harm, no-foul.

It does however remove a test for one "operation" that doesn't actually exist in production: creating a brand new PR for one or more specific Dependencies.

If we need this behaviour in future, it'll be relatively trivial to compose a new Operation class and test to bring it back so let's remove it for clarity for now.

Finally, I've added some notes on the plan to split and migrate the tests as well as a warning to avoid adding new tests to the old interface. I've added an issue to the backlog to circle back to do this as we get time.

### DIFF
--- a/updater/spec/dependabot/updater_spec.rb
+++ b/updater/spec/dependabot/updater_spec.rb
@@ -10,6 +10,35 @@ require "dependabot/file_fetchers"
 require "dependabot/updater"
 require "dependabot/service"
 
+### DO NOT ADD NEW TESTS TO THIS FILE
+#
+# This file tests all of our specific Dependabot::Updater::Operations via the
+# top-level Dependabot::Updater interface as it predates us breaking the class
+# up.
+#
+# Any tests should be added to the relevant file in spec/dependabot/operations,
+# if it does not exist it should be created, for an example see:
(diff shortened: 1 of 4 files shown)


jakecoffman (approved):
🗑️

updater/spec/dependabot/updater_spec.rb
```diff
@@ -10,10 +10,37 @@
...
+# - Copy any _passing_ tests to your new spec/dependabot/operations file
+# - Check which of the failing tests should apply to the target Operation
+# - Copy them and adjust their setup so they pass
+# - Repeat for the next Operation
+# - Consider breaking out shared_example groups for any tests which are the same
+#   for each Operation
+#
+# Once this process has been completed, this test should be repurposed to ensure
+# that the Updater delegates to the right Operation class and handles halting
+# errors in an expected way.
```
jeffwidman, line 41: ❤️ great comment!

PR #7213, opened 2023-05-02
[Updater] Fix brittle test for grouped updates

brrygrdn:
We added a new Grouped Update spec which uses VCR and actually executes bundler to verify a behaviour, but the test had a mass expectation that the entire lockfile matches the fixture.

This isn't realistic as even with the fixtures pinning the target dependencies to a fixed range, downstream dependencies may shift slightly between executions.

Let's update the spec to just verify the Gemfile.lock has been updated to match the gemspec as a more stable test that still verifies the outcome.

### DIFF
--- a/updater/spec/dependabot/updater/operations/group_update_all_versions_spec.rb
+++ b/updater/spec/dependabot/updater/operations/group_update_all_versions_spec.rb
@@ -132,16 +132,22 @@ RSpec.describe Dependabot::Updater::Operations::GroupUpdateAllVersions do
       expect(mock_error_handler).not_to receive(:handle_dependabot_error)
       expect(mock_service).to receive(:create_pull_request) do |dependency_change|
         expect(dependency_change.dependency_group.name).to eql("everything-everywhere-all-at-once")
-        expect(dependency_change.updated_dependency_files_hash.length).to eql(2)
 
-        gemfile_lock = dependency_change.updated_dependency_files.find { |file| file.path == "/Gemfile.lock" }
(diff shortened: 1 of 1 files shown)


jurre (approved):

updater/spec/dependabot/updater/operations/group_update_all_versions_spec.rb
```diff
@@ -132,16 +132,22 @@
...
+        expect(dependency_change.updated_dependency_files_hash.length).to eql(2)
+
+        gemspec = dependency_change.updated_dependency_files.find { |file| file.path == "/library.gemspec" }
+        expect(gemspec.content).to eql(fixture("bundler_gemspec/updated/library.gemspec"))
+
+        gemfile_lock = dependency_change.updated_dependency_files.find { |file| file.path == "/Gemfile.lock" }
+        # Since we are actually running bundler for this test, let's just check the Gemfile.lock has been updated
+        # with the same ranges as the library.gemspec rather than expecting the entire lockfile to match.
+        expect(gemfile_lock.content).to include("rack (>= 2.1.4, < 3.1.0)")
+        expect(gemfile_lock.content).to include("rubocop (>= 0.76, < 1.51)")
```
jeffwidman, line 150: Won't this still be brittle/fail as soon as `rubocop` ships `1.51`?

One way around this is adding an `ignore` condition to limit the top-end, example:
https://github.com/dependabot/dependabot-core/pull/7016/files#diff-f431e04235a6a13d962956d7be22db4aacd5ee96a9cf996c42ac62f26dc3d4ccR158

PR #7215, opened 2023-05-02
Fix updating GitHub Actions pinned to mixed references

### LINKED ISSUE #7214: GitHub Actions with mixed versions not getting updated

### Package ecosystem

GitHub Actions

### Package manager version

2

### Language version

_No response_

 - `.github/workflows/build.yml` - https://github.com/hashicorp/terraform-ls/blob/main/.github/workflows/build.yml
 - `.github/workflows/test.yml` - https://github.com/hashicorp/terraform-ls/blob/main/.github/workflows/test.yml

### dependabot.yml content

```yml
version: 2
updates:
  - package-ecosystem: "gomod"
    directory: "/"
    schedule:
      interval: "daily"
    labels: ["dependencies"]

    directory: "/"
    schedule:
      interval: "daily"
    labels: ["dependencies"]
```

### Updated dependency

No updates detected/suggested.

 - Expected `actions/upload-artifact` to be upgraded from `v3.1.1` to `v3.1.2`.
   - `build.yml` has this Action pinned to `v3.1.1`, e.g. https://github.com/hashicorp/terraform-ls/blob/410c012bc4a46e644bad8333ba2ab2a77efba055/.github/workflows/build.yml#L54
   - There is another workflow `benchmarks.yml` with this Action [is already pinned](https://github.com/hashicorp/terraform-ls/blob/410c012bc4a46e644bad8333ba2ab2a77efba055/.github/workflows/benchmarks.yml#L70) to `v3.1.2`
 - Expected `actions/checkout` to be upgraded from `v3.5.0` to `v3.5.2`
   - One job in the `test.yml` workflow has this Action pinned to `v3.5.0` https://github.com/hashicorp/terraform-ls/blob/410c012bc4a46e644bad8333ba2ab2a77efba055/.github/workflows/test.yml#L20
   - The same Action within the same workflow [is already pinned](https://github.com/hashicorp/terraform-ls/blob/410c012bc4a46e644bad8333ba2ab2a77efba055/.github/workflows/test.yml#L38) to `v.3.5.2`

I found https://github.com/dependabot/dependabot-core/pull/6082 and https://github.com/dependabot/dependabot-core/issues/3750 which probably should have addressed this issue. Could this be just a caching problem somewhere then?

deivid-rodriguez:
When workflow files are pinned to mixed SHA references (some out of date, and some not), things would not work unless an out of date reference would be the first reference parsed.

There were two issues:

* When parsing, once we resolve the tag for a pinned SHA, we would not resolve others, and thus not let our "dependency merging logic" set the lowest tag as the version for the entire dependency.

* When figuring out updated sources, we were assuming there was a single pinned SHA. Instead, we need to update all pinned SHAs.

Fixes #7214.


github_actions/spec/dependabot/github_actions/update_checker_spec.rb
```diff
@@ -381,11 +384,11 @@
       end
     end
 
-    context "given a dependency with multiple git refs", :vcr do
```
jeffwidman, line 384: Where these `vcr` cassettes actually unused, or just outdated?

I'm personally fine dropping them, but if I'm reading this correctly it means these tests now require internet access to run whereas previously they didn't... 

Again, I'm fine with this change, just wanted clarification since commit message says dropping _unused_ cassettes...

jeffwidman (approved):
This all looks fine, but TBH I don't have a lot of confidence that I'm fully grok'ing this one because there's enough subtle edge cases that I think we'll just have to trust CI for no regressions...

github_actions/spec/dependabot/github_actions/update_checker_spec.rb
deivid-rodriguez: Extracted and improved this particular change as #7233.

jeffwidman (approved):

PR #7217, opened 2023-05-02
[Updater] Add a test for two overlapping groups

brrygrdn:
This is a quick unit test that verifies two things:
- That we correctly parse N `dependency_groups` to perform two group update attempts
- That if two groups overlap, we will PR the same dependencies in both by design

### DIFF
--- a/updater/lib/dependabot/updater/operations/group_update_all_versions.rb
+++ b/updater/lib/dependabot/updater/operations/group_update_all_versions.rb
@@ -71,6 +71,8 @@ module Dependabot
         def run_grouped_dependency_updates
           Dependabot.logger.info("Starting grouped update job for #{job.source.repo}")
 
+          Dependabot.logger.info("Found #{dependency_snapshot.groups.count} group(s).")
+
           dependency_snapshot.groups.each do |_group_hash, group|
             Dependabot.logger.info("Starting update group for '#{group.name}'")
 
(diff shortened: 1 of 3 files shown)


updater/spec/dependabot/updater/operations/group_update_all_versions_spec.rb
```diff
@@ -81,6 +81,40 @@
...
+    let(:job_definition) do
+      job_definition_fixture("bundler/version_updates/group_update_all_overlapping_groups")
+    end
+
+    let(:dependency_files) do
+      original_bundler_files
+    end
+
+    before do
+      stub_rubygems_calls
```
jeffwidman, line 94: Does this need to be cleared/reset after the tests?

Or does it already register a cleanup callback internally?

updater/lib/dependabot/updater/operations/group_update_all_versions.rb
```diff
@@ -45,6 +45,8 @@ def perform
 
           Dependabot.logger.info("Starting grouped update job for #{job.source.repo}")
 
+          Dependabot.logger.info("Found #{dependency_snapshot.groups.count} group(s).")
```
jeffwidman, line 48: This will also be useful in prod debugging. 👍

jeffwidman (approved):

brrygrdn:
Ah, the test failure is a known issue, this can merge downstream of https://github.com/dependabot/dependabot-core/pull/7216 and the problem will go away.

updater/spec/dependabot/updater/operations/group_update_all_versions_spec.rb
brrygrdn: Good spot, it is ultimately just `webmock` stubs which are plugged into RSpec to clear afterwards automagically.

PR #7224, opened 2023-05-03
Pin docker_registry2 to 1.14.0 to fix CI

deivid-rodriguez:
I think we may need some changes to adapt to version 1.15.0, in particular, to https://github.com/deitch/docker_registry2/pull/85.

In tests we stub a HEAD request which 1.15.0 no longer performs, so hopefully we just need to adapt the specs, but the new version means no behavioral changes for us.

For now I'll pin it to get CI green.


jurre (approved):

deivid-rodriguez:
I was hoping Dependabot to create a PR to upgrade to docker_registry2 so that we can track the update there, but it failed. I'll look into why.

deivid-rodriguez:
I think I figured it out, will try to open a fix shortly.

deivid-rodriguez:
PR fixing this is at #7227! After that is merged, a PR updating docker_registry2 to 1.15.0 should pop up.

PR #7227, opened 2023-05-03
Fix Dependabot failing to create PRs when updates are needed on path gemspecs

deivid-rodriguez:
Checking CI for now, I will add a spec if it passes.


deivid-rodriguez:
Spec added a I set this as ready!

pavera (approved):

PR #7228, opened 2023-05-03
Add tag to indicate if repo is private in sentry reports

jurre:
Currently finding a way to reproduce issues found in Sentry can be challenging, usually the easiest way is to find a public project that we can run our code against. Today this is hard, because we need to manually comb the errors in sentry until we find one that's public.

With this change, that should no longer be an issue.

### DIFF
--- a/updater/lib/dependabot/job.rb
+++ b/updater/lib/dependabot/job.rb
@@ -41,6 +41,7 @@ module Dependabot
       vendor_dependencies
       dependency_groups
       dependency_group_to_refresh
+      repo_private
     )
 
     attr_reader :allowed_updates,
@@ -80,7 +81,7 @@ module Dependabot
 
     # NOTE: "attributes" are fetched and injected at run time from
     # dependabot-api using the UpdateJobPrivateSerializer
-    def initialize(attributes)
+    def initialize(attributes) # rubocop:disable Metrics/AbcSize
       @id                             = attributes.fetch(:id)
       @allowed_updates                = attributes.fetch(:allowed_updates)
       @commit_message_options         = attributes.fetch(:commit_message_options, {})
@@ -111,6 +112,7 @@ module Dependabot
       @vendor_dependencies            = attributes.fetch(:vendor_dependencies, false)
       @dependency_groups              = attributes.fetch(:dependency_groups, [])
       @dependency_group_to_refresh    = attributes.fetch(:dependency_group_to_refresh, nil)
+      @repo_private                   = attributes.fetch(:repo_private, nil)
 
       register_experiments
       register_dependency_groups
@@ -128,6 +130,10 @@ module Dependabot
       return nil unless clone?
 
       @repo_contents_path
+    end
+
+    def repo_private?
+      @repo_private
     end
 
     def updating_a_pull_request?
--- a/updater/lib/dependabot/service.rb
+++ b/updater/lib/dependabot/service.rb
@@ -72,7 +72,8 @@ module Dependabot
         {
           tags: tags.merge({
             update_job_id: job&.id,
-            package_manager: job&.package_manager
+            package_manager: job&.package_manager,
+            repo_private: job&.repo_private?
           }.compact),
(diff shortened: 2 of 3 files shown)


jakecoffman (approved):

updater/lib/dependabot/job.rb
```diff
@@ -108,6 +109,7 @@ def initialize(attributes)
       @vendor_dependencies            = attributes.fetch(:vendor_dependencies, false)
       @dependency_groups              = attributes.fetch(:dependency_groups, [])
       @dependency_group_to_refresh    = attributes.fetch(:dependency_group_to_refresh, nil)
+      @repo_private                   = attributes.fetch(:repo_private, nil)
```
landongrindheim, line 112: 🔎 Since this is meant to be a boolean, what do you think of defaulting it to `false`?

```suggestion
      @repo_private                   = attributes.fetch(:repo_private, false)
```

landongrindheim (approved):

updater/lib/dependabot/job.rb
jeffwidman: Curious of your thinking here Landon... to me, `nil`/`false` communicate different things, and since this _describes something_, rather than being an _option we choose to set_, I'd be hesitant to set a default value which may actually be incorrect if we ever had a bug... I think we'd be more likely to discover a bug if we saw `nil` and wondered why it wasn't getting set... 

But you're an experienced engineer, so maybe I'm overlooking something?

jeffwidman (approved):

updater/lib/dependabot/job.rb
landongrindheim: > Curious of your thinking here

It's only used in `#repo_private?` (👇). Having predicate methods only return `true`/`false` is a thing I believe in/practice. `repo_private? => nil` kinda breaks my brain. I'd also be 👍 to using `!!@repo_private` in the above method. 

```ruby
    def repo_private?
      @repo_private
    end
```

If `nil` being present is indicative of a bug, we probably shouldn't provide a default value of `#fetch`. Instead, we should let it blow up so we can catch/address the issue earlier.

jurre: Yeah, so my reasoning for not defaulting to false here has to do with how we'd deploy the different components that make up dependabot. And by allowing `nil` here, it communicates that the value was not passed in. This way, we can roll out the change to core, without worrying if the change in the internal components have been deployed or not. Having it return `false` in this case, is in my opinion misleading, because we'd end up with repo's that most definitely _are_ private getting that value.

What we can do is once everything has been rolled out in all environments is come back and make this value non-optional and remove the default value and have it blow up if no value is passed in. But given that our deployments on GHES instances aren't managed by us it carries some risk and I'm not sure that it's worth it.

To me there is no inherent value in ensuring that this tag in sentry is never nil, if it's nil, the tag is omitted, which seems desirable in this case. We could change the `repo_private?` method name to omit the `?` if you feel strongly about predicate methods not returning nil, I personally do not feel strongly about this.

landongrindheim: I'm definitely not interested in forcing my preferences on others, so please feel free to ship this as is 🙂 

---

Without the context of the commit/PR, I wouldn't have understood this to be specifically for error reporting. The predicate method reads as a uitility that will be used throughout the library, but it's not actually used.

I'm not familiar with how these are parsed and sent to Sentry. Is there something iterating over the instance variables? Do we need `#repo_private?`?

abdulapopoola (approved):

updater/lib/dependabot/job.rb
jurre: It's called [here](https://github.com/dependabot/dependabot-core/pull/7228/files#diff-72d65f40e823f48b30f91700de1178de9a24ad4530b76aab448c7342ea56d494R76)

landongrindheim: 🤦 Thanks for pointing that out!

PR #7230, opened 2023-05-03
Use `latest` `bundler` for generating `updater/Gemfile.Lock`

jeffwidman:
Previously, the version of bundler defaults to the one shipped with Ruby.

However, we custom pin to newer versions in `Dockerfile.updater-core`, so when we then run `bundle lock`, within the context of `docker-dev-shell`, it bumps the `BUNDLED WITH` version: https://github.com/dependabot/dependabot-core/pull/7229#discussion_r1184064978

The `ruby/setup-ruby` action does support reading the bundler version from `Gemfile.lock`: https://github.com/ruby/setup-ruby/blob/d2b39ad0b52eca07d23f3aa14fdf2a3fcc1f411c/action.yml#L20

But our lockfile is in a subdirectory. They do support specifying the working directory to look into: https://github.com/ruby/setup-ruby/blob/d2b39ad0b52eca07d23f3aa14fdf2a3fcc1f411c/action.yml#L27-L28

But if we do that, then the action doesn't pickup the `.ruby-version` file, since that's located in the root directory.

I really _didn't_ want to manage yet another pin, so defaulted this to `latest` with the assumption that generally in `dependabot-core` we'll be on the latest version of `bundler`.

This is an alternative to:
* https://github.com/dependabot/dependabot-core/pull/7231

### DIFF
--- a/.github/workflows/gems-bump-version.yml
+++ b/.github/workflows/gems-bump-version.yml
@@ -17,6 +17,11 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: ruby/setup-ruby@v1 # bump-version.rb needs bundler
+        with:
+          # We pin bundler version in Dockerfile.updater-core BUNDLER_V2_VERSION
+          # which also ends up in `updater/Gemfile.lock`. To avoid having to
+          # keep the pin in-sync here, assume that core is generally on 'latest'.
+          bundler: 'latest'
 
       - name: Bump the version
         run: |


jeffwidman:
Alternatively, what we could do here is stop running `bundle lock` for `updater/Gemfile.lock` using the version of `bundler` in `Dockerfile.updater-core`.

The version in the dockerfile is used for the native helper, where we _do_ want the newer versions in order to pickup bugfixes when running against user repos.

But IIUC (and I'm not 100% confident), that doesn't require `updater/Gemfile.lock` to use that version of `bundler`... and in fact makes our life more difficult because it's yet another pin we have to manage compared to if we just used the default version shipping with Ruby.

I'm going to get input from others on the team who know Ruby / bundler / updater better.

deivid-rodriguez:
I think it should be enough to set `working-directory: updater`, so that the version in the lockfile is respected? See https://github.com/ruby/setup-ruby/blob/d2b39ad0b52eca07d23f3aa14fdf2a3fcc1f411c/action.yml#L27-L28C12.

jeffwidman:
Unfortunately doing that creates an inverse problem, as I mentioned in my PR description:
> But our lockfile is in a subdirectory. They do support specifying the working directory to look into: https://github.com/ruby/setup-ruby/blob/d2b39ad0b52eca07d23f3aa14fdf2a3fcc1f411c/action.yml#L27-L28
>
> But if we do that, then the action doesn't pickup the `.ruby-version` file, since that's located in the root directory.

the `.ruby-version` file needs to be in root to be picked up by Rubocop, and probably other tooling as well like IDEs... 

🐔 and 🥚... unless I'm overlooking something?

deivid-rodriguez:
Oh, sorry I missed that in the PR description, good point.

I'm afraid using latest may update the `BUNDLED WITH` version to latest? Maybe symlink a `.ruby-version` file at `updater/` to the root one?

jeffwidman:
Closing in favor of:
* https://github.com/dependabot/dependabot-core/pull/7239

The problem with this approach is that it will silently change the version of `bundler` used in the native helper, but only in situations where we run `bundle install` against `updater/Gemfile.lock`... ie, production at GitHub, but not `dry-run.rb` so it could lead to very difficult to debug behavior.

PR #7231, opened 2023-05-03
Use `ruby`'s default version of `bundler` for `updater/Gemfile.lock`

jeffwidman:
See problem here: https://github.com/dependabot/dependabot-core/pull/7229#discussion_r1184064978

If we keep the version of `bundler` in `updater/Gemfile.lock` in sync with the version used in our native helper, then we not only have to remember to bump it when we bump the native helper version, we also will need to start pinning it in `.github/workflows/gems-bump-version.yml` file.

The simpler thing is to simply stop keeping them in sync.

This is an alternative to:
* https://github.com/dependabot/dependabot-core/pull/7230


jeffwidman:
Closing in favor of:
* https://github.com/dependabot/dependabot-core/pull/7239

jeffwidman:
The problem with this approach is that it will silently change the version of `bundler` used in the native helper, but only in situations where we run `bundle install` against `updater/Gemfile.lock`... ie, production at GitHub, but not `dry-run.rb` so it could lead to very difficult to debug behavior.

PR #7233, opened 2023-05-04
Remove unnecessary usages of VCR

deivid-rodriguez:
I extracted this from #7215 and went all the way to remove all usages of VCR from this file.


jurre (approved):
Nice

PR #7237, opened 2023-05-04
Remove `gem install bundler`

deivid-rodriguez:
We can use the default Bundler that comes with Ruby, and that should automatically install the version locked at `BUNDLED WITH` inside the lockfile and use that.


deivid-rodriguez:
Didn't work and I won't dig deeper into why.

PR #7242, opened 2023-05-05
Raise when a sub dependency is already up-to-date in a grouped update

Nishnha:
WIP

For now we raise a new error when the a sub-dependency is already up-to-date in a grouped update.
In the future we should avoid filtering out the lockfile in #filter_sub_dependency_files for grouped updates

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater.rb
@@ -25,6 +25,17 @@ module Dependabot
         end
       end
 
+      class GroupedSubDependencyUpToDate < StandardError
+        def initialize(message:, error_context:)
+          super(message)
+          @error_context = error_context
+        end
+
+        def raven_context
+          { extra: @error_context }
+        end
+      end
+
       def self.updated_files_regex
         [
           /^package\.json$/,
@@ -40,6 +51,16 @@ module Dependabot
 
         updated_files += updated_manifest_files
         updated_files += updated_lockfiles
+
+        # FIXME: For grouped dependency updates, it is possible that a transitive dependency was updated
+        # by a previous dependency update. We should avoid filtering sub dependency files when
+        # performing grouped updates.
+        if updated_files.none? && filtered_sub_dependency_files.none?
+          raise GroupedSubDependencyUpToDate.new(
+            message: "Subdependency is already up-to-date",
+            error_context: error_context(updated_files: updated_files)
+          )
+        end
 
         if updated_files.none?
           raise NoChangeError.new(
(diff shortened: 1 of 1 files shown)


npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater.rb
```diff
@@ -25,6 +25,17 @@ def raven_context
...
+      class GroupedSubDependencyUpToDate < StandardError
+        def initialize(message:, error_context:)
+          super(message)
+          @error_context = error_context
+        end
+
+        def raven_context
+          { extra: @error_context }
+        end
+      end
```
brrygrdn, line 37: I think my concern with adding this is that it makes npm specifically semi-aware of groups but in a way where it is kind of guessing the problem is due to a grouped update being performed 🤔 

This could lead to confusion where due to a bug elsewhere this error starts appearing in non-grouped updates?

brrygrdn:
> Alternatively, we can tolerate these errors or try and detect when the dependency is already up-to-date before we filter out the lockfile.

Yeah, this is a tricky one - I think we should consider tolerating these errors locally within the Updater's grouped update operations for now - potentially just rescuing them and logging a warning - while we investigate the right way to handle this.

brrygrdn:
>  This can happen when an earlier update has already updated the transitive dependency update being attempted.

I don't fully understand why this is happening - we pass the modified files from the previous step into each update. So I'm not following why we'd re-attempt to update something a previous step had handled.

That suggests to me that our file hand-over between steps is lossy? i.e. we aren't passing everything that has actually been modified.

Do we need to revise this method?
https://github.com/dependabot/dependabot-core/blob/f4165e7ca5b03694cddb31a4c7496f7951b91339/updater/lib/dependabot/updater/group_update_creation.rb#L148-L157

I note that it assumes that `dependency_snapshot.dependency_files` contains all the files that need to be carried over between steps, so if the `updated_files` from an NPM update expanded this list to include other files we would drop them 🤔

Nishnha:
> > This can happen when an earlier update has already updated the transitive dependency update being attempted.
> 
> I don't fully understand why this is happening - we pass the modified files from the previous step into each update. So I'm not following why we'd re-attempt to update something a previous step had handled.
> 
> That suggests to me that our file hand-over between steps is lossy? i.e. we aren't passing everything that has actually been modified.
> 
> Do we need to revise this method?
> 
> https://github.com/dependabot/dependabot-core/blob/f4165e7ca5b03694cddb31a4c7496f7951b91339/updater/lib/dependabot/updater/group_update_creation.rb#L148-L157
> 
> I note that it assumes that `dependency_snapshot.dependency_files` contains all the files that need to be carried over between steps, so if the `updated_files` from an NPM update expanded this list to include other files we would drop them 🤔

No, I don't believe the file hand-over step is lossy.

My understanding is that the #filter_sub_dependency_files method is checking whether any dependencies are able to update, but since the dependency update being attempted is already up-to-date from an earlier parent dependency update, the method returns an empty list.

We could also skip over filtering the sub dependency files for now but it would result in duplicate work

deivid-rodriguez:
So, is the problem here that dependency parsing happens just once before each group update step, so the originally parsed versions for sub dependencies can become inaccurate by intermediate sub dependency updates in intermediate steps?

Nishnha:
Closing in favor of https://github.com/dependabot/dependabot-core/pull/7276, which implements the fix described in https://github.com/dependabot/dependabot-core/pull/7242#issuecomment-1536378426

PR #7244, opened 2023-05-05
Less truncation in RSpec errors

deivid-rodriguez:
This decreases the chances that useful parts of an error message are hidden, making it harder to see what the problem is.

For example, the error fixed by #7243 now would show like this:

```
Failures:

  1) Dependabot::Cargo::FileUpdater::LockfileUpdater#updated_lockfile_content when updating the lockfile fails because an existing requirement is no good raises a helpful error
     Failure/Error:
       expect { updater.updated_lockfile_content }.
         to raise_error do |error|
           expect(error).to be_a(Dependabot::DependencyFileNotResolvable)
           expect(error.message).
             to include("version for the requirement `regex = \"^99.0.0\"`")
         end
     
       expected #<Dependabot::SharedHelpers::HelperSubprocessFailed: Updating crates.io index
       error: failed to select a version for the requirement `time = "=99.0.0"`
       candidate versions found which didn't match: 0.3.20, 0.3.19, 0.3.18, ...
       location searched: crates.io index
       required by package `dependabot v0.1.0 (/home/dependabot/dependabot-core/cargo/dependabot_tmp_dir)`
       perhaps a crate was updated and forgotten to be re-vendored?> to be a kind of Dependabot::DependencyFileNotResolvable
     # ./spec/dependabot/cargo/file_updater/lockfile_updater_spec.rb:98:in `block (5 levels) in <top (required)>'
     # /home/dependabot/dependabot-core/common/spec/spec_helper.rb:64:in `block (2 levels) in <top (required)>'
     # /home/dependabot/dependabot-core/omnibus/.bundle/ruby/3.1.0/gems/webmock-3.18.1/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'

Finished in 0.65461 seconds (files took 3.2 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spec/dependabot/cargo/file_updater/lockfile_updater_spec.rb:97 # Dependabot::Cargo::FileUpdater::LockfileUpdater#updated_lockfile_content when updating the lockfile fails because an existing requirement is no good raises a helpful error
```

The useful bit that makes the culprit easier to discover (`time = "=99.0.0"`) is now displayed by default.


landongrindheim (approved):
💯0️⃣

PR #7245, opened 2023-05-05
Refactor preparing `package.json` files

deivid-rodriguez:
I found a fair bit of duplicated logic when updating NPM and Yarn lockfiles that's already implemented in `PackageJsonPreparer`. So I'm refactoring things to reuse it.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
@@ -17,6 +17,7 @@ module Dependabot
       class YarnLockfileUpdater
         require_relative "npmrc_builder"
         require_relative "package_json_updater"
+        require_relative "package_json_preparer"
 
         def initialize(dependencies:, dependency_files:, repo_contents_path:, credentials:)
           @dependencies = dependencies
@@ -357,13 +358,7 @@ module Dependabot
                 file.content
               end
 
-            updated_content = replace_ssh_sources(updated_content)
-
-            # A bug prevents Yarn recognising that a directory is part of a
-            # workspace if it is specified with a `./` prefix.
-            updated_content = remove_workspace_path_prefixes(updated_content)
-
-            updated_content = sanitized_package_json_content(updated_content)
+            updated_content = package_json_preparer(updated_content).prepared_content
             File.write(file.name, updated_content)
           end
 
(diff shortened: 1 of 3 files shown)


npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
```diff
@@ -574,18 +537,6 @@ def yarnrc_content
...
 
-        def sanitized_package_json_content(content)
-          updated_content =
-            content.
-            gsub(/\{\{[^\}]*?\}\}/, "something"). # {{ nm }} syntax not allowed
-            gsub(/(?<!\\)\\ /, " ").          # escaped whitespace not allowed
-            gsub(%r{^\s*//.*}, " ")           # comments are not allowed
-
-          json = JSON.parse(updated_content)
-          json["name"] = json["name"].delete(" ") if json["name"].is_a?(String)
```
deivid-rodriguez, line 585: This cleanup is no longer done after this refactoring. I think it's no longer needed but I'll extract its removal to a separate PR.

deivid-rodriguez: This was extracted into https://github.com/dependabot/dependabot-core/pull/7246 and deployed a couple of weeks ago.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
```diff
@@ -545,6 +492,22 @@ def updated_package_json_content(file)
             ).updated_package_json.content
         end
 
+        def prepared_package_json_content(content)
+          @prepared_package_json_content ||= {}
+          @prepared_package_json_content[content] ||=
+            package_json_preparer(
+              content
+            ).prepared_content
+        end
```
Nishnha, line 501: Adding this as a memoized function seems extra compared to using `package_json_preparer(content).prepared_content` explicitly?

I don't feel strongly about this though so it is also fine to leave in.

Nishnha (approved):
Thanks!

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
deivid-rodriguez: I removed the extra memoization and also found a similar one which I removed too at https://github.com/dependabot/dependabot-core/pull/7245/commits/81da5a1b40699053079e26730822daa94845160c.

PR #7246, opened 2023-05-05
Revert sanitization of package name

deivid-rodriguez:
I think this was a mistake from the beginning. The specs added after reverting the patch still pass without the change, and the commit message reads: "JS: Sanitize spaces in filenames", while there's no filename involved here whatsoever, just the package name.

This commit reverts c5b1fe6c958b8dbfaeeaca5c96ffddca60a5076d and 2864404857f4918a323a4109cd38cbb8ae0285a4.

References: https://github.com/dependabot/dependabot-core/pull/7245#discussion_r1186445982.


deivid-rodriguez:
The removed tests still pass after the revert. Of course, because test npm and the removed code is for Yarn. Something got weirdly out of sync here. In any case, just to make sure I migrated the specs to use Yarn, and they still pass.

In any case, I think it's best to remove them to avoid any confusion, since there's nothing invalid (I believe) about a package name with spaces?

jakecoffman (approved):

deivid-rodriguez:
That's a good point, but yeah if that was needed the place would be wrong. And why only for Yarn? 🤷‍♂️

jeffwidman (approved):

jeffwidman:
Checked with Deivid on slack and he thought this was good to go, so going to merge/deploy it.

PR #7247, opened 2023-05-05
Generate PR intro for grouped update PRs

bdragon:
This PR adds handling to generate more user friendly PR intros for grouped update PRs (similar to the way in which we generate PR titles for grouped update PRs).


common/lib/dependabot/pull_request_creator/message_builder.rb
```diff
@@ -357,14 +372,17 @@ def dependency_set
...
-        dependencies.map do |dependency|
-          if source_url(dependency)
-            "[#{dependency.display_name}](#{source_url(dependency)})"
-          elsif homepage_url(dependency)
-            "[#{dependency.display_name}](#{homepage_url(dependency)})"
-          else
-            dependency.display_name
-          end
+        uniq_deps = dependencies.each_with_object({}) { |dep, memo| memo[dep.name] ||= dep }.values
+        uniq_deps.map { |dep| dependency_link(dep) }
```
jurre, line 376: Not sure if any of the work done here is potentially expensive, but seems like we could memoize this. If it's always a small collection and the work is not expensive then keeping it as is seems fine 👍

jurre (approved):
Overall approach looks great to me, noticed there's an e2e test failing, which looks like it's because a generated grouped PR has changed, as expected 👍 Also needs a `rubocop -A` to fix some string formatting, but that should not lead to significant changes to the PR

bdragon:
Thanks @jurre 

I think the way this needs to be deployed is:
* comment out the failing e2e test and merge this PR
* regenerate expectations for failing e2e test with the new output
* re-enable commented-out e2e test, which should then be passing

PR #7262, opened 2023-05-09
Revert lockfile-only fix again

### LINKED ISSUE #7257: Gems that require Ruby 3.0+ are giving updates when on Ruby 2.7.8 that don't run or work

### Package ecosystem

bundler

### Package manager version

2.4.12

### Language version

2.7.8

/Gemfile

### dependabot.yml content

```
package-ecosystem: "bundler"
    directory: "/"
    schedule:
      interval: "daily"
    versioning-strategy: increase
    open-pull-requests-limit: 10
    insecure-external-code-execution: deny
    commit-message:
      prefix: "[bundler] "
    allow:
      - dependency-type: "all"
 ```

### Updated dependency

plaid, for example ---> from 16.0.0 to anything beyond

same as selenium-webdriver (4.9.0) to 4.9.1 ( see https://github.com/SeleniumHQ/selenium/commit/cd3216e68dbc4ae61c0efae6e2969d5e18776bdb)

Dependabot should either:

* Not propose a gem update that conflicts with ruby version (and perhaps a config for which ruby version to use, or read .ruby-version)
* have a separate error message in the Dependabot.yml to suggest ruby upgrade

Note: I am aware Ruby 2.7.8 is EOL (or soon to be), but dependabot can handle this better, in my opinion

it doesn't suggest an update is available when running `bundler update`

n/a

.ruby_version of 2.7.8 (or any other way to set based on your environment.

A Gemfile with
```
gem "plaid",          "~> 16.0.0"
gem "selenium-webdriver", "~> 4.9.0"
```

### PR DESCRIPTION
**Pull request by deivid-rodriguez**
This reverts commit c1fa6d81df62c07142b6e7806765d299bb50bebc.

Once again, this change seems to be causing issues with incorrect PRs being created.

I managed to reproduce this at https://github.com/dsp-testing/brew.

There's several confusing things in the logs in there, but the one that stands out to me is that sometimes Bundler seems to be falling back to the old dependency API (`https://index.rubygems.org/api/v1/dependencies`), which does not provide Ruby version constraints. That's why Dependabot thinks the update is correct and creates a PR.

This does not point out at a bug in c1fa6d81df62c07142b6e7806765d299bb50bebc, but at some problem in either Bundler (client) or rubygems.org (server).

Interestingly, I can't reproduce the problem either locally with dry-run or the CLI (the new API is always used), it seems to only happen in production, that's why the issue went unnoticed again.  If we ever try to introduce the fix again, we should make sure to completely forbid `/api/v1/dependencies`. We might not need to do anything here though since that's being [sunset](https://blog.rubygems.org/2023/02/22/dependency-api-deprecation.html). The date has been extended due to some issues which may or may not be related to this problem.

Anyways, it seems that the issue has been surfaced by our change, so let's revert it again 😢.

Fixes #7257 (hopefully, will test staging before merging).
Reopens #5569 (once again).

### CODE REVIEW

**Comment by Bo98** [comment]
Looks to me that it stops receiving a response from `/versions` (seemingly permanently until the end of the run - though that's only about a minute later):

```
  proxy | 2023/05/08 19:01:06 [536] GET https://index.rubygems.org:443/versions
  proxy | 2023/05/08 19:01:06 [536] 304 https://index.rubygems.org:443/versions
  proxy | 2023/05/08 19:01:22 [538] GET https://index.rubygems.org:443/versions
  proxy | 2023/05/08 19:01:32 [540] GET https://index.rubygems.org:443/versions
  proxy | 2023/05/08 19:01:42 [542] GET https://index.rubygems.org:443/api/v1/dependencies
```

Exactly 10 seconds between requests screams timeout to me. Not sure if there's been an increase in the volume of requests or what.

> We might not need to do anything here though since that's being [sunset](https://blog.rubygems.org/2023/02/22/dependency-api-deprecation.html).

It'll probably fall back to full index once the dependencies API is gone.

---

I appreciate your efforts on that original lockfile issue fix - looks like you're quite close now!

deivid-rodriguez:
Thanks for the research. I'm thinking an alternative solution could be to monkeypatch Bundler to never try neither the old API not the full index when talking to rubygems.org. If the compact index is having issues, just let it fail, or maybe add some retries.

deivid-rodriguez:
> I suppose there's still an issue somewhere causing compact index timeouts, but there's always the risk that something could happen anyway (as is the nature of the internet), so it seems like a good idea to be improving the fallback scenario anyway.

Yeah, I'm going to check with the RubyGems.org team, but I think these may be due to some intermittent rubygems.org issues.

deivid-rodriguez:
@Bo98 I think I found a [proper fix](https://github.com/dependabot/dependabot-core/pull/7269) for this. I just deployed it, so hopefully if you now comment `@dependabot recreate` on the bad PRs, Dependabot should hopefully close them with a comment like: "Looks like <depenedency> is no longer updatable, so this is no longer needed.".

deivid-rodriguez:
After all, the original issue seems to have been caused by rubygems.org Compact Index API issues, and not directly by the `lockfile-only` fix, since #7257 is not using `versioning-strategy: lockfile-only` at all. So I'm going to close this and hope #7269 plus fixes at rubygems.org will do the trick here!

Bo98:
Thank you again David!

deivid-rodriguez:
Thanks to you for the great feedback and patience! ❤️

PR #7263, opened 2023-05-09
Do not check only unlocked gems for pruning

jurre:
We're seeing an issue with bundler updates not correctly pruning all the old cached dependencies that have been removed.

This PR removes a bit of code that digs into bundler internals to find gems that have been "unlocked" (meaning to have their constraints removed so they _may_ be updated by bundler), and only remove `.gem`s that match this list. This code existed in an attempt to support cases where you had a conditional dependency in your Gemfile (for example in order to test against multiple versions of some dependency). However, this logic seems flawed, as it does not take into account any transitive dependencies, which is also the issue that we've observed.

Since this code seems to cause more problems than it tries to solve, I think removing it is the best step forward for now. We could potentially update the check to take all the old deps, compare them with the new ones, and only prune `.gem`s from the cache that were actively changed in this PR, but that has us deviating from bundlers native behavior, which I'm not sure is worth the hassle.

This also removes a bit of temporary code that was in place in order to support an internal project at GitHub, which is no longer needed. Since these were entangled, removing them together is easiest, but if there's any objections I'm happy to split this up.


deivid-rodriguez:
@jurre I got curious about this and experimented with the fix I had in mind. Can you check whether this patch fixes the issues with some gems not being pruned?

```diff
diff --git a/bundler/helpers/v2/lib/functions/lockfile_updater.rb b/bundler/helpers/v2/lib/functions/lockfile_updater.rb
index ca619ccbb..c49603c75 100644
--- a/bundler/helpers/v2/lib/functions/lockfile_updater.rb
+++ b/bundler/helpers/v2/lib/functions/lockfile_updater.rb
@@ -64,8 +64,12 @@ module Functions
     def cache_vendored_gems(definition)
       # Dependencies that have been unlocked for the update (including
       # sub-dependencies)
-      unlocked_gems = definition.instance_variable_get(:@unlock).
-                      fetch(:gems).reject { |gem| __keep_on_prune?(gem) }
+      resolve = definition.resolve
+
+      updated_gems = resolve.reject do |spec|
+        lockfile_specs.include?(spec) || __keep_on_prune?(spec.name)
+      end.map(&:name).uniq
+
       bundler_opts = {
         cache_all: true,
         cache_all_platforms: true,
@@ -79,8 +83,7 @@ module Functions
         # Only prune unlocked gems (the original implementation is in
         # Bundler::Runtime)
         cache_path = Bundler.app_cache
-        resolve = definition.resolve
-        prune_gem_cache(resolve, cache_path, unlocked_gems)
+        prune_gem_cache(resolve, cache_path, updated_gems)
         prune_git_and_path_cache(resolve, cache_path)
       end
     end
@@ -95,14 +98,14 @@ module Functions
     end
 
     # Copied from Bundler::Runtime: Modified to only prune gems that have
-    # been unlocked
-    def prune_gem_cache(resolve, cache_path, unlocked_gems)
+    # been updated
+    def prune_gem_cache(resolve, cache_path, updated_gems)
       cached_gems = Dir["#{cache_path}/*.gem"]
 
       outdated_gems = cached_gems.reject do |path|
         spec = Bundler.rubygems.spec_from_gem path
 
-        !unlocked_gems.include?(spec.name) || resolve.any? do |s|
+        !updated_gems.include?(spec.name) || resolve.any? do |s|
           s.name == spec.name && s.version == spec.version &&
             !s.source.is_a?(Bundler::Source::Git)
         end
@@ -147,8 +150,7 @@ module Functions
     end
 
     def unlock_blocking_subdeps(dependencies_to_unlock, error)
-      all_deps =  Bundler::LockfileParser.new(lockfile).
-                  specs.map(&:name).map(&:to_s)
+      all_deps = lockfile_specs.map(&:name).map(&:to_s)
       top_level = build_definition([]).dependencies.
                   map(&:name).map(&:to_s)
       allowed_new_unlocks = all_deps - top_level - dependencies_to_unlock
@@ -224,6 +226,10 @@ module Functions
       sources.all? { |s| s&.fetch("type", nil) == "git" }
     end
 
+    def lockfile_specs
+      @lockfile_specs ||= ::Bundler::LockfileParser.new(lockfile).specs
+    end
+
     def lockfile
       @lockfile ||= File.read(lockfile_name)
     end
```

I think it should not break any existing specs.

jurre:
@deivid-rodriguez that does not seem to pass the test-case we had in place for this unfortunately

deivid-rodriguez:
Oh too bad! You mean any existing test cases in our test suite or the realworld example where we found that not all previous `.gem` files were being pruned?

jurre:
I was just running the tests cases that we have here around conditional deps, I assume it will work as expected on the real-world scenario that doesn't have these

deivid-rodriguez:
@jurre Just to make sure I pushed the change to trigger CI and it seems to be passing: https://github.com/dependabot/dependabot-core/pull/7295.

jurre:
> @jurre Just to make sure I pushed the change to trigger CI and it seems to be passing: #7295.

Nice! My dev env must have not picked up the changes somehow, that's brilliant, let me pull those changes in here, I agree keeping the not updated gems in the cache is better, thanks @deivid-rodriguez!

PR #7266, opened 2023-05-09
Only talk to compact index API

deivid-rodriguez:
Experimenting with an alternative to #7262.


deivid-rodriguez:
This does not seem to work and I think I found a proper solution at https://github.com/dependabot/dependabot-core/pull/7269. Closing.

PR #7268, opened 2023-05-09
[Updater] Fix a bug in how we handle 'peer' manifests

brrygrdn:
For ecosystems such as docker, it is very common for each step in the Updater to modify a _different_ manifest file rather than them all to sequentially change the same files.

Our existing logic was adequate for ecosystems (and projects) where we were making iterative changes to the same files, but in cases where updates modified a subset, we were losing those _not_ modified in the previous step and reverting them back to the initial state.

This PR adds a failing test case for this and then solves the issue by introducing a small utility class which wraps a hash map of the dependency files at their most recent state with some metadata on the changes made.

### Debug logging

As part of making this change, I've added debug logging into the `Dependabot::Updater::GroupDependencyFileBatch` so it verbosely explains the files after each change is applied, e.g.

```
updater | 2023/05/09 16:28:10 DEBUG Dependency files updated:
updater | 2023/05/09 16:28:10 DEBUG   - /docker/Dockerfile.bundler ( Changed 1 times )
updater | 2023/05/09 16:28:10 DEBUG   - /docker/Dockerfile.cargo ( Changed 1 times )
updater | 2023/05/09 16:28:10 DEBUG   - /docker/Dockerfile.composer ( Changed 1 times )
updater | 2023/05/09 16:28:10 DEBUG   - /docker/Dockerfile.docker ( Changed 1 times )
updater | 2023/05/09 16:28:10 DEBUG   - /docker/Dockerfile.elm ( Changed 1 times )
updater | 2023/05/09 16:28:10 DEBUG   - /docker/Dockerfile.github-actions ( Changed 1 times )
updater | 2023/05/09 16:28:10 DEBUG   - /docker/Dockerfile.gitsubmodule
updater | 2023/05/09 16:28:10 DEBUG   - /docker/Dockerfile.gomod
updater | 2023/05/09 16:28:10 DEBUG   - /docker/Dockerfile.gradle
updater | 2023/05/09 16:28:10 DEBUG   - /docker/Dockerfile.maven
updater | 2023/05/09 16:28:10 DEBUG   - /docker/Dockerfile.mix
updater | 2023/05/09 16:28:10 DEBUG   - /docker/Dockerfile.npm
updater | 2023/05/09 16:28:10 DEBUG   - /docker/Dockerfile.nuget
updater | 2023/05/09 16:28:10 DEBUG   - /docker/Dockerfile.pip
updater | 2023/05/09 16:28:10 DEBUG   - /docker/Dockerfile.proxy
updater | 2023/05/09 16:28:10 DEBUG   - /docker/Dockerfile.pub
updater | 2023/05/09 16:28:10 DEBUG   - /docker/Dockerfile.terraform
```


updater/lib/dependabot/updater/group_dependency_file_batch.rb
```diff
@@ -0,0 +1,61 @@
...
+
+      # Replaces the existing files
+      def merge(updated_files)
+        updated_files.each do |updated_file|
+          existing_file = @dependency_file_batch[updated_file.path]
+
+          change_count = if existing_file
+                           existing_file.fetch(:change_count, 0)
+                         else
+                           # Let's warn about this in debug mode, but otherwise tolerate this.
```
mctofu, line 39: Could updates to vendored dependencies trigger this or do they get handled somewhere else?

mctofu (approved):

updater/lib/dependabot/updater/group_dependency_file_batch.rb
brrygrdn: I think the adapters do pass them back so this will warn on vendored files - I will make sure this is the case as it will be worth calling out a new file being injected into the set in debug mode.

PR #7269, opened 2023-05-09
Fix some bad PRs getting created when compact index API is having issues

### LINKED ISSUE #7257: Gems that require Ruby 3.0+ are giving updates when on Ruby 2.7.8 that don't run or work

deivid-rodriguez:
If Bundler needs to fallback to fetching dependency information from the old API, we need to double that the target version Bundler said it's compatible, it's actually compatible with the locked Ruby version.

This is because this API ignores requirements on the Ruby version.

This was already checked when figuring out the latest resolvable version, but it was not checked when falling back to trying to unlock all dependencies.

This commit should fix that.

Fixes #7257.


deivid-rodriguez:
> To clarify, are you referring here to the old API or the compact API?

To the old API. The CompactIndex API should be always correct, however, it's been having issues for the last few days, so Bundler fallback to the old API, which mostly works, but not for us. 

> It is surprising that we have to handle this edge case in our custom code, and not in bundler nor our custom native helper for bundler, but if anyone would understand the intricacies here, it'd be you...

I first tried to monkeypatch Bundler to not use the dependency API, but then if the CompactIndex API is having issues like now... nothing works. Then I found that we were already explicitly handling this case, just not everywhere where needed, so I just fixed the existing approach. By the way, I _think_ CI is failing because.... the CompactIndex API is having issue, I believe we're missing some VCR cassettes in there.

PR #7271, opened 2023-05-09
Fix outdated reference to DockerHub

jeffwidman:
When the images were split, this was updated to push to GHCR rather than Dockerhub.

In fact, you can see here the last image was pushed 3 months ago: https://hub.docker.com/r/dependabot/dependabot-core/tags

But this reference to dockerhub wasn't updated.

Note:
This pushes the `dependabot-updater-core` image to GHCR, but not the ecosystem-specific images... that was likely an oversight. Correcting that is outside the scope of this PR though.

### DIFF
--- a/.github/workflows/images-updater-core.yml
+++ b/.github/workflows/images-updater-core.yml
@@ -9,7 +9,7 @@ on:
       - v[0-9]+.[0-9]+.[0-9]+
 jobs:
   push-updater-core-image:
-    name: Push dependabot-updater-core image to docker hub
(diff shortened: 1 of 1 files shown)


jeffwidman:
I've also removed the `DOCKER_USERNAME` / `DOCKER_PASSWORD` encrypted secrets from this repo, they are not currently used and have not been for a while, but they are likely still valid creds... We have another copy of them internally, so if we later decide to start re-publishing to DockerHub and need them again we can re-add them.

deivid-rodriguez (approved):

PR #7272, opened 2023-05-10
Switch to watching releases rather than tags

jeffwidman:
Watching for tags works, but introduces the risk that an unexpected tag name that happens to match the regex will trigger a push.

Additionally, a tag on a branch other than `main` could trigger the workflow as well.

By limiting only to actual releases, we ensure that we only push to RubyGems when we actually publish a release.

Related:

### DIFF
--- a/.github/workflows/gems-release-to-rubygems.yml
+++ b/.github/workflows/gems-release-to-rubygems.yml
@@ -1,8 +1,7 @@
 name: Gems - Release to RubyGems
 on:
-  push:
-    tags:
-      - v[0-9]+.[0-9]+.[0-9]+
+  release:
+    types: [published]
 permissions:
   contents: read
 jobs:


Nishnha:
I think we'll want to update our scripts/docs to generate releases instead of adding a tag? We currently don't have any releases for Core https://github.com/dependabot/dependabot-core/releases

jeffwidman:
> I think we'll first want to update our scripts/docs to generate releases instead of adding a tag?

They were already updated in https://github.com/dependabot/dependabot-core/pull/7211:

https://github.com/dependabot/dependabot-core/blob/d9519cd2c4c901bc60d26c3b93ebed65049d8f4c/README.md#L535-L545

https://github.com/dependabot/dependabot-core/blob/d9519cd2c4c901bc60d26c3b93ebed65049d8f4c/.github/workflows/gems-bump-version.yml#L67-L77

You can see the generated output here: 

https://github.com/dependabot/dependabot-core/actions/runs/4888068150

> We currently don't have any releases for Core https://github.com/dependabot/dependabot-core/releases

Yeah, that's because I ran into the issue that the [version bump PR's need a PAT to trigger CI](https://github.com/dependabot/dependabot-core/pull/7240)... I'm working through some security questions around that with a few other folks.

I'm fine to wait, but there's no code change (other than the PAT perhaps) that blocks this... both our docs and our scripts are already updated.

deivid-rodriguez (approved):

PR #7275, opened 2023-05-10
Pin CodeQL action

deivid-rodriguez:
The current version is so slow.


deivid-rodriguez:
Yeah, this one seems like the last fast version (3 minutes vs half an hour). The idea would be to merge this and let Dependabot open PRs to upgrade to newer versions until the performance regression is fixed.

Nishnha:
I think we could also let the codeQL team know and ask them to update the v2 tag to match the latest version?

Nishnha (approved):

deivid-rodriguez:
I asked about it at https://github.com/github/codeql/pull/13028, maybe they already have a potential fix!

deivid-rodriguez:
Oh wait, I'm confused now, I swear I saw the Ruby analysis finishing fast, but it seems to be still running. What happened? 🤣

deivid-rodriguez:
I can see from the logs that the action is still using CodeQL-2.13.0 under the hood, which I believe is the slow version. I'm not sure how to configure an older version to be used.

Issue #7223, opened 2023-05-03
Dependabot frequently failing to raise PR for pub packages


cpswan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

pub

### Package manager version

NA

### Language version

2.19.6

### Manifest location and content before the Dependabot update

[/packages/at_secondary_server/pubspec.yaml](https://github.com/atsign-foundation/at_server/blob/trunk/packages/at_secondary_server/pubspec.yaml)

### dependabot.yml content

```yaml
version: 2
enable-beta-ecosystems: true
updates:
  # Maintain dependencies for GitHub Actions
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "daily"
  - package-ecosystem: "docker"
    directory: "/tools/build_secondary"
    schedule:
      interval: "daily"
  - package-ecosystem: "docker"
    directory: "/tools/build_virtual_environment/ve"
    schedule:
      interval: "daily"
  - package-ecosystem: "docker"
    directory: "/tools/build_virtual_environment/ve_base"
    schedule:
      interval: "daily"     
  - package-ecosystem: "docker"
    directory: "/packages/at_root_server"
    schedule:
      interval: "daily"
  - package-ecosystem: "pub"
    directory: "/packages/at_secondary_server"
    schedule:
      interval: "daily"
  - package-ecosystem: "pip"
    directory: "/tools" # Location of package manifests
    schedule:
      interval: "daily"
      
```

### Updated dependency

| Dependency | From version | To version |
|---|---|---|
| collection | 1.17.0 | 1.17.1 |
| basic_utils | 5.4.2 | 5.5.4 |
| meta | 1.8.0 | 1.9.1 |
| args | 2.3.1 | 2.4.0 |

### What you expected to see, versus what you actually saw

I'd expect to see a PR for the dependencies that need to be updated. Instead the [logs](https://gist.github.com/cpswan/9884347c0cddc7e29b6bb5bf9113fd45) show that a PR couldn't be raised:

They conclude with:

```
updater | 2023/05/03 10:40:54 INFO Results:
updater | Dependabot encountered '4' error(s) during execution, please check the logs for more details.
updater | +-------------------------------+
updater | | Dependencies failed to update |
updater | +---------------+---------------+
updater | | collection    | unknown_error |
updater | | basic_utils   | unknown_error |
updater | | meta          | unknown_error |
updater | | args          | unknown_error |
[...]

PR #7277, opened 2023-05-10
fix NilClass error on credential-less runs

jakecoffman:
Noticed in https://github.com/dependabot/dependabot-core/issues/7223, when running a CLI job without credentials can result in a NilClass error due to `credentials_metadata` not existing.

Defaulting the value to `[]` fixes it. 

This isn't something that can happen in production, more of a user-friendly feature for folks who haven't set up a LOCAL_GITHUB_ACCESS_TOKEN.


deivid-rodriguez (approved):
I'm pretty sure I run into this while experimenting with the CLI and I did get confused!

PR #7281, opened 2023-05-10
Add basic smoke test for PNPM

deivid-rodriguez:


deivid-rodriguez:
Ready for a review, will remove the last commit before merging which I added just to make sure the spec is actually passing.

Nishnha (approved):

PR #7288, opened 2023-05-10
Go: fix grouped updates missing updated dependencies in manifests

jakecoffman:
The Go updater code has two code paths. The more common one is it uses the cloned repo to do updates, first doing a `git reset`, not using the manifests that were fetched.

The other path only happens when a clone fails and there's no vendoring (and other specific conditions) where it will write out a dummy project that imports all dependencies so that the update can proceed with imperfect information.

The problem is when Grouped Updates comes along, the Go code needs to write the manifests to disk each time since they get wiped out between updates of the dependencies with that `git reset`.

So this PR passes in the dependency_files to the GoModUpdater so it can overwrite what is on disk each time, and make the Grouped Update successful.

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_updater.rb
+++ b/go_modules/lib/dependabot/go_modules/file_updater.rb
@@ -117,6 +117,7 @@ module Dependabot
         @file_updater ||=
           GoModUpdater.new(
             dependencies: dependencies,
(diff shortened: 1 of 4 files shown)


bdragon (approved):

brrygrdn (approved):
Nice find! Thanks for fixing this!

PR #7295, opened 2023-05-11
Fix missing cache prunes

deivid-rodriguez:
Just pushing this change to see which specs fail.


deivid-rodriguez:
I rebased this PR now that https://github.com/dependabot/dependabot-core/pull/7296 has been shipped. Happily no complaints so far about #7296 by the way 🎉.

jeffwidman:
Took a look, but I have minimal context on this one... @jurre / @deivid-rodriguez should we ship this or hold off?

jurre (approved):

PR #7296, opened 2023-05-11
Remove `persistent_gems_after_clean` workaround

jurre:
We put this check in place as a workaround for an internal project at GitHub that was using some custom extensions to bundler. This has since been removed, and this check is no longer relevant, so let's clean it up.


deivid-rodriguez (commented):
I think the same code is present on Bundler 1 helper and that can go too?

deivid-rodriguez (approved):

jeffwidman (approved):

PR #7298, opened 2023-05-11
Allow updating the version of PNPM itself

deivid-rodriguez:
This is just an experiment, no tests yet.

It's the first example of us updating the version of a package manager, I think it would be a good practice to start doing this more, since it helps ecosystems move forward and makes it easier for us to drop support for old versions.


Stanzilla:
Would love to see this, also for other package managers!

pboling:
pnpm v9 still does not work _at all_. ✨ 

It has been months now, with only minimal interest or response.  We need something like your solution @deivid-rodriguez  so the community is not dependent on non-existent dependabot maintenance.

# Related Issues:

- https://github.com/dependabot/dependabot-core/issues/9684#issuecomment-2137899341
- https://github.com/dependabot/dependabot-core/issues/9682
- https://github.com/dependabot/dependabot-core/issues/9684
- https://github.com/dependabot/dependabot-core/issues/9321

# Related PRs

- https://github.com/dependabot/dependabot-core/pull/7298
- https://github.com/dependabot/dependabot-core/pull/8936
- https://github.com/dependabot/dependabot-core/pull/9006
- https://github.com/dependabot/dependabot-core/pull/9687#issuecomment-2139150623
- https://github.com/dependabot/dependabot-core/pull/9668
- https://github.com/dependabot/dependabot-core/pull/9808

deivid-rodriguez:
I'm not sure if the almost identical message you posted in most of those other linked issues is relevant there, but I don't think it's relevant here at all. I don't think this PR will fix any of those issues.

You're welcome you to try for yourself and verify though, and also to contribute a fix for the problem with PNPM 9. That would be helpful.

pboling:
I am too overloaded by other work to be able to help support a marketing bullet point for Microsoft.  My hope in posting a summary of the issue is to get Microsoft to become aware of the mess and cleanup the house, but that isn't working.

This is an excellent example of why we can't allow this space to become monopolized. I will instead switch to Depfu, which supports pnpm.

deivid-rodriguez:
I'm going to close this since I never managed to complete it.

PR #7301, opened 2023-05-12
No need to specify read perms

jeffwidman:
I've changed the default perms of the `GITHUB_TOKEN` to read-only to match the [new
default](https://github.blog/changelog/2023-02-02-github-actions-updating-the-default-github_token-permissions-to-read-only/).

As a result, there's no need to explicitly downgrade the permissions to `read`, and it creates confusion/visual distraction.

For the CodeQL change, they explicitly document that for public repos only the `security-events` scope needs to be explicitly specified for _public_ repos, such as this one.


jeffwidman:
Actually @mctofu pointed out to me privately:
> There's still value in setting this because it prevents the token from having package:read and metadata:read by default 
> https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

So closing.

Issue #7258, opened 2023-05-08
Issues when updating git dependencies in PNPM


deivid-rodriguez:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

npm

### Package manager version

PNPM

### Manifest location and content before the Dependabot update

I don't have a repro for this yet, but we're seeing issues in private repositories related to upgrading dependencies coming from git sources.

### What you expected to see, versus what you actually saw

Updates should happen correctly without errors.

In particular, we see errors like the following:

```
 ERROR  Command failed with exit code 128: git ls-remote git+ssh://git@github.com/<repo>.git HEAD
error: cannot run ssh: No such file or directory
fatal: unable to fork

This error happened while installing the dependencies of <name>@<version>

pnpm: Command failed with exit code 128: git ls-remote git+ssh://git@github.com/<repo>.git HEAD
error: cannot run ssh: No such file or directory
fatal: unable to fork
    at makeError (/home/dependabot/.cache/node/corepack/pnpm/8.3.1/dist/pnpm.cjs:24230:17)
    at handlePromise (/home/dependabot/.cache/node/corepack/pnpm/8.3.1/dist/pnpm.cjs:24801:33)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async getRepoRefs (/home/dependabot/.cache/node/corepack/pnpm/8.3.1/dist/pnpm.cjs:110651:23)
    at async resolveRef (/home/dependabot/.cache/node/corepack/pnpm/8.3.1/dist/pnpm.cjs:110663:20)
    at async resolveGit (/home/dependabot/.cache/node/corepack/pnpm/8.3.1/dist/pnpm.cjs:110614:24)
    at async Object.resolve (/home/dependabot/.cache/node/corepack/pnpm/8.3.1/dist/pnpm.cjs:113185:173)
    at async run (/home/dependabot/.cache/node/corepack/pnpm/8.3.1/dist/pnpm.cjs:124970:23)
/home/dependabot/common/lib/dependabot/shared_helpers.rb:329:in `run_shell_command'
/home/dependabot/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/pnpm_lockfile_updater.rb:58:in `run_pnpm_updater'
[...]

PR #7303, opened 2023-05-12
Do we need git-shim?

deivid-rodriguez:
More of a question PR, do we need it?

### DIFF
--- a/Dockerfile.updater-core
+++ b/Dockerfile.updater-core
@@ -50,10 +50,6 @@ WORKDIR $DEPENDABOT_HOME
 # This avoids downloading large files not necessary for the dependabot scripts
 ENV GIT_LFS_SKIP_SMUDGE=1
 
-# Place a git shim ahead of git on the path to rewrite git arguments to use HTTPS.
-ARG SHIM="https://github.com/dependabot/git-shim/releases/download/v1.4.0/git-v1.4.0-linux-amd64.tar.gz"
-RUN curl -sL $SHIM -o git-shim.tar.gz && mkdir -p ~/bin && tar -xvf git-shim.tar.gz -C ~/bin && rm git-shim.tar.gz
-
 COPY --chown=dependabot:dependabot .rubocop.yml .rubocop.yml
 COPY --chown=dependabot:dependabot .ruby-version .ruby-version
 COPY --chown=dependabot:dependabot omnibus omnibus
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
So the rationale here is that the motivation explained in [git-shim's](https://github.com/dependabot/git-shim) README:

> This is useful in a corporate environment where an HTTP Proxy is used. You could use the command git config --global url."https://".insteadOf git:// but that doesn't work for SCP-style style URLs: git@github.com:org/repo.git due to the colon :.

I don't think that's true, since we're actually using this git feature with SCP-style URIs? 🤔

https://github.com/dependabot/dependabot-core/blob/6252098c225778707f257ce77a6bcdd3be03b8c4/common/lib/dependabot/shared_helpers.rb#L261-L268C8

I guess the one case where insteadOf doesn't work is when using inlined `user:pass` in URLs, like `ssh://user:pass@github.com/foo/bar`, maybe that's the reason?

Insights appreciated!

jakecoffman:
See https://github.com/dependabot/dependabot-core/pull/5310 for more info on why we went with the shim.

We should add that glommio dependency to the Cargo smoke tests so any future attempts to fix it will not regress!

deivid-rodriguez:
Mmmm, I see, thanks for the pointer.

So we keep the git shim to fix the edge case of "not knowing the host in advance + SCP-style uri being used", correct? However, that realworld example there does not fit that case, right? Both URIs mentioned there (ssh://git@github.com/DataDog/glommio.git and git://git.kernel.dk/liburing) and not SCP-style URIs?

deivid-rodriguez:
I would agree normally, but SCP-style URIs feel like maybe too big of an edge case to warrant keeping our own shim? Specially if we have not really seen any issue in real life that cannot be fixed tweaking other stuff we already do.

In the case of PNPM, URL rewriting confused me. PNPM keeps track and detects the best protocol based on the shape of the URIs and accessibility of the repo. If it detects ssh as the best protocol, and we still rewrite URIs to https, then the requests may succeed, but PNPM will still save a ssh URI to the lockfile. If I had changed the shim or the our `insteadOf` configurations to rewrite URIs in the case of https://github.com/dependabot/dependabot-core/issues/7258, we would've introduced a bug probably worse than https://github.com/dependabot/dependabot-core/issues/7258, because we would've generated incorrect PRs.

Overall, I think all approaches have upsides and downsides, so I'm good with sticking to the shim. We should probably update the README of https://github.com/dependabot/git-shim to clarify the use case a bit though?

deivid-rodriguez:
I'll close this since it was just a way to learn, thanks for the info and insights @jakecoffman :)

PR #7145, opened 2023-04-22
fix(gradle): update dependencies only if group and artifact ids are the same

yeikel:
With the current implementation, any line matching the requirement is upgraded. That leads to undesired side effects as described in the issue

Fixes #7002

PR #7304, opened 2023-05-12
Revert "fix(gradle): update dependencies only if both group and artifact ids are the same"

deivid-rodriguez:
This reverts commit 6252098c225778707f257ce77a6bcdd3be03b8c4.

Turns out it broke [something related to version catalogs](https://github.com/dependabot/dependabot-core/pull/7145#issuecomment-1545596624), reverting to figure that out without blocking other stuff.


deivid-rodriguez:
Also added the version catalog smoke test so that we can catch issues properly next time.

PR #7308, opened 2023-05-12
🤫 Silence patch-version warnings for consumers

landongrindheim:
We've been doing this in our specs, but consumers of this library end up with warnings if they're not in perfect compliance (which we've deemed unnecessary). This is a bet that good citizenship leans more on protecting downstream workflows than patch-level exactness.


jeffwidman:
What warning is this spitting out?

landongrindheim:
> Can you provide an example of the warning this is spitting out?

@jeffwidman We've gotten this output when using `dependabot-core` from an app running Ruby 3.1.2:

```
warning: parser/current is loading parser/ruby31, which recognizes3.1.3-compliant syntax, but you are running 3.1.2.
```

Patch-level enforcement isn't something _I_ think we should be in the game of, so I'd like to see this relaxed. That said, I haven't touched this in two months and I think leaving it open provides little to no value. I'm going to close this for now.

PR #7310, opened 2023-05-12
[Updater] GroupUpdateAllVersions now performs refreshes for existing PRs instead of deferring them to a separate job

brrygrdn:
For our initial iteration of grouped update scheduled jobs we choose the following strategy:
- The scheduled `GroupUpdateAllVersions` job only runs for groups which do not have a pre-existing, open pull request
- Separately, the service elects to enqueue a `RefreshGroupUpdatePullRequest` for these existing open pull requests

This approach optimised for avoiding a single Operation doing too much and prefer to run smaller jobs when we can, **but** a major drawback is it means that the primarily scheduled update job logs do not show these refreshes.

This is a pretty significant drawback as we currently do not do a great job of surfacing refresh/rebase logs in the UI.

### Changes

This branch adds a small interface to allow the `RefreshGroupUpdatePullRequest` to have data about an existing pull request to be injected as an 'overlay' on the `Dependabot::Job` in order to allow `GroupUpdateAllVersions` to delegate to it to perform refresh operations in-line without having to incorporate the refresh logic directly.

### DIFF
--- a/updater/lib/dependabot/dependency_change.rb
+++ b/updater/lib/dependabot/dependency_change.rb
@@ -46,51 +46,5 @@ module Dependabot
     def grouped_update?
       !!dependency_group
     end
-
-    # This method combines checking the job's `updating_a_pull_request` flag
-    # with verification the dependencies involved remain the same.
-    #
-    # If the dependencies involved have changed, we should close the old PR
-    # rather than supersede it as the new changes don't necessarily follow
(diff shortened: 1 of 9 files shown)


mctofu (approved):

brrygrdn:
We are parking this change for now as it requires some API changes that we need to hammer out the right direction on to move to this model.

PR #7323, opened 2023-05-15
Turn `composer` version into a metric

jeffwidman:
We'd like to know how many jobs are running using `composer 1` vs
`composer 2`, so this exposes it as a metric.

Long term, this can be further refined to report `composer` versions
with more granularity, for example `2.3`, `2.4`, `2.5`, etc.

But right now we don't have active questions that'd be answered by
knowing that granularity, and our metrics provider charges based on the
number of metrics we emit, so `2` is an appropriate bucket for now.

Related:
* https://github.com/dependabot/dependabot-core/issues/6298


composer/lib/dependabot/composer/helpers.rb
```diff
@@ -16,13 +16,13 @@ module Helpers
       def self.composer_version(composer_json, parsed_lockfile = nil)
         if parsed_lockfile && parsed_lockfile["plugin-api-version"]
           version = Composer::Version.new(parsed_lockfile["plugin-api-version"])
```
jeffwidman, line 18: We could directly return the version here, then let callers decide their desired granularity... ie, metrics probably `2.4`, `2.5`, etc.

But right now we don't need that, and I expect by the time we do we'll have dropped support for composer v1 (https://github.com/dependabot/dependabot-core/issues/6298), so the implementation will be a bit simpler.

So will leave that for future PR.

I did add the necessary unit tests so that if/when that happens, they'll flag the callers that need changing.

composer/spec/dependabot/composer/file_fetcher_spec.rb
```diff
@@ -59,6 +59,13 @@
       to match_array(%w(composer.json composer.lock))
   end
 
+  it "provides the composer version" do
+    expect(file_fetcher_instance.package_manager_version).to eq({
+      ecosystem: "composer",
+      package_managers: { "composer" => "2" }
```
jeffwidman, line 65: If we return the full version, this test will complain, which will remind us to truncate the patch version rather than fully reporting it.

composer/spec/dependabot/composer/file_fetcher_spec.rb
```diff
@@ -76,6 +83,12 @@
     it "fetches the composer.json" do
       expect(file_fetcher_instance.files.map(&:name)).to eq(["composer.json"])
     end
+
+    it "provides the composer version" do
+      expect(file_fetcher_instance.package_manager_version).to eq({
+        ecosystem: "composer",
+        package_managers: { "composer" => "2" }
```
jeffwidman, line 90: Vs this test will doublecheck that if we don't have `composer.lock` so we can't know the full version and are defaulting to `2` that it still continues to default even if we return the full version normally.

jeffwidman (commented):
Annotating the changes.

composer/spec/dependabot/composer/file_fetcher_spec.rb
```diff
@@ -59,6 +59,13 @@
       to match_array(%w(composer.json composer.lock))
   end
 
+  it "provides the composer version" do
+    expect(file_fetcher_instance.package_manager_version).to eq({
+      ecosystem: "composer",
+      package_managers: { "composer" => "1" }
```
jeffwidman, line 65: Unfortunately, both of these tests are against a fixture that was generated using `composer 1`.

I thought about fixing, but we'll need to update this fixture anyway when we drop support for `composer 1`, so no sense partially doing it here, only to completely refactor later.

We _do_ have some UT test coverage of `"2"` because the code that actually generates the version number is used to accessing the native helper... if it wasn't spitting out `"2"`, all those unit tests would be failing.

composer/lib/dependabot/composer/helpers.rb
```diff
@@ -16,13 +16,13 @@ module Helpers
       def self.composer_version(composer_json, parsed_lockfile = nil)
         if parsed_lockfile && parsed_lockfile["plugin-api-version"]
```
pavera, line 17: Should we use `.fetch` here to protect against a key error if there is a missing `plugin-api-version`?

composer/lib/dependabot/composer/native_helpers.rb
```diff
@@ -3,8 +3,8 @@
 module Dependabot
   module Composer
     module NativeHelpers
-      def self.composer_helper_path(composer_version: "v2")
-        File.join(composer_helpers_dir, composer_version, "bin/run")
+      def self.composer_helper_path(composer_version: "2")
+        File.join(composer_helpers_dir, "v" + composer_version, "bin/run")
```
pavera, line 7:
```suggestion
        File.join(composer_helpers_dir, "v#{ composer_version }", "bin/run")
```

composer/lib/dependabot/composer/helpers.rb
jeffwidman: I had to look this one up as I didn't know 😁 

For a missing key:
* `.fetch` raises a `KeyError`, which we'd need to catch...
* `[key]` returns `nil`, which is "falsey", so a missing key will result in this `if` condition failing.

So IIUC, we want it to stay like it is actually...

pavera (approved):
Of course my suggestion failed linting...

PR #7324, opened 2023-05-16
Add App Token to release workflow so CI jobs run

jeffwidman:
Adding an App Token to the release workflow so CI jobs run.

This App Token is currently stored as an actions secret. Actions prevents forks from reading secrets, so we'd have to explicitly merge a PR in order to leak the secret. Actions also redacts secrets, so we should be mostly safe from typos.

If we later decide we want to protect it further, we have two options:

1. Migrate it to a deployment secret for increased security. This way _any_ workflow run accessing the secret requires explicit approval from the maintainers. Unfortunately, the workflow will just hang in a "waiting" state, but not easily notify the user that it needs approval. I prototyped the workflow, and it was pretty clunky--we had to trigger the workflow, remember to manually approve it, and only then the PR shows up... vs we'd like to use the PR showing up every week automatically as a reminder to cut a release. Vs here the PR just shows up and waits for us to merge it, CI will already have run.
2. Migrate the entire workflow to a private repo. This way we reduce the odds of leaking the secret, but the workflows are a little complicated because a workflow in another repo can't trigger "on release". In other words, it'd be easy to migrate this "generate PR" workflow, but the secret we care more about is the RubyGems API key and that'd still be in the clear unless we migrated it. So we'd end up having to have a small trigger workflow in this repo that watches for releases, then triggers the workflow in the private repo.

Both of the above options are technically more secure, but it's primarily defense-in-depth as actions will already keep the current state secure unless we really screw up a PR review or fat-finger a major major typo. So after discussion we decided the inconvenience of the above two options aren't worth the increased defense-in-depth they provide.

We went with an application token rather than a PAT so that the token isn't tied to a particular developer's account. The app is scoped to only this specific repo, so if it's compromised it won't allow privilege escalation to other repos.

### DIFF
--- a/.github/workflows/gems-bump-version.yml
+++ b/.github/workflows/gems-bump-version.yml
@@ -15,8 +15,16 @@ jobs:
   Create-PR-To-Bump-Dependabot-Gems-Version:
     runs-on: ubuntu-latest
     steps:
+      - name: Generate token
+        id: generate_token
+        uses: tibdex/github-app-token@b62528385c34dbc9f38e5f4225ac829252d1ea92 # v1.8.0
+        with:
+          app_id: ${{ secrets.DEPENDABOT_CORE_ACTION_AUTOMATION_APP_ID }}
+          private_key: ${{ secrets.DEPENDABOT_CORE_ACTION_AUTOMATION_PRIVATE_KEY }}
+
       - uses: actions/checkout@v3
         with:
+          token: ${{ steps.generate_token.outputs.token }}
           # Ensure we start from main in case the workflow is run from a branch
           ref: "main"
 
@@ -64,7 +72,7 @@ jobs:
           echo "PR created at URL: $PR_URL"
           echo "PR_URL=$PR_URL" >> $GITHUB_ENV
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ steps.generate_token.outputs.token }}
 
       - name: Set summary
         run: |


mctofu (approved):

.github/workflows/gems-bump-version.yml
```diff
@@ -15,7 +15,17 @@ jobs:
...
+        id: generate_token
+        uses: tibdex/github-app-token@b62528385c34dbc9f38e5f4225ac829252d1ea92 # v1.8.0
+        with:
+          app_id: ${{ secrets.DEVELOPER_AUTOMATION_APP_ID }}
+          private_key: ${{ secrets.DEVELOPER_AUTOMATION_PRIVATE_KEY }}
+
       - uses: actions/checkout@v3
+        with:
+          token: ${{ steps.generate_token.outputs.token }}
+
```
yeikel, line 28: I believe that checkout step can continue to use the generated token

jeffwidman: It can, but a few steps later we have:

1. Create a branch, commit the changes
2. Push the branch

I'm not actually sure if 1️⃣ requires a write token, since it's a local branch/commit, but I am sure that 2️⃣ needs the token... 

And I couldn't find a way to pass the token to `git` in a non-clunky way... rather than adding something like https://stackoverflow.com/a/57932145/770425, I thought it was simpler to just `checkout` the repo with the write permissions and then I didn't need to doublecheck if 1️⃣ needed the token and I didn't need to configure custom tokens in the URLs... 

Technically it might be slightly less secure, but there's only two steps that get elevated privileges beyond what they need, and that's bumping the version and configuring the git user... so I thought it was fine to take the shortcut in the name of visual simplicity. 😀

.github/workflows/gems-bump-version.yml
```diff
@@ -15,7 +15,17 @@ jobs:
   Create-PR-To-Bump-Dependabot-Gems-Version:
     runs-on: ubuntu-latest
     steps:
+      - name: Generate token
+        id: generate_token
+        uses: tibdex/github-app-token@b62528385c34dbc9f38e5f4225ac829252d1ea92 # v1.8.0
+        with:
+          app_id: ${{ secrets.DEVELOPER_AUTOMATION_APP_ID }}
+          private_key: ${{ secrets.DEVELOPER_AUTOMATION_PRIVATE_KEY }}
```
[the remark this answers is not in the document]
jeffwidman: Oh, good catch 👍 

We started out going for a PAT and then switch to an app token, let me update the PR/commit descriptions...

.github/workflows/gems-bump-version.yml
yeikel: What I meant is that the checkout step specifically does not seem to need the elevated token and you're passing the `GH_TOKEN` to the GH cli manually. 

You can still generate it and use it for the other step unless I am missing something. It does not change anything in terms of security, but I was just curious

yeikel: I see what you mean, I missed the steps where you call `git` directly

Thank you for explaining!

.github/workflows/gems-bump-version.yml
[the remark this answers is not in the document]
yeikel: No problem :) 

If you squash the changes you can do it at that time :)

jeffwidman:
I was able to trigger a successful run here: https://github.com/dependabot/dependabot-core/actions/runs/5000268607

Note that this required the change from #7335

.github/workflows/gems-bump-version.yml
jeffwidman, line 1: I supposed could change the "configure git user" step to run with this new actions user instead of
```
git config user.name "github-actions[bot]"
# Specifying the full email allows the avatar to show up: https://github.com/orgs/community/discussions/26560
git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```

See for example this PR where the PR is created by one bot, but the commit is created by a different bot:
* https://github.com/dependabot/dependabot-core/pull/7336

I don't _think_ we could easily make it so the PR gets created by the GitHub Actions bot instead of our custom bot...

PR #7328, opened 2023-05-16
docs: improve documentation

yeikel:


npm_and_yarn/Dockerfile
```diff
@@ -1,16 +1,25 @@
 FROM ghcr.io/dependabot/dependabot-updater-core
 ENV DEPENDABOT_HOME /home/dependabot
-### JAVASCRIPT
```
yeikel, line 3: This comment did not add any value in my opinion. I am guessing that this was relevant when we had a single image

jeffwidman (approved):
Thank you!

PR #7342, opened 2023-05-17
Look in parent directories for nuget.config files

jmarolf:
technically nuget.config files don't need to be next to project or solution files. Nuget traverses the parent directories as well. This change allows us to look for the nuget.config files in the same way the nuget client does


jurre:
Thanks! The code change looks good to me and the new functionality makes sense. We'll need to hold off with merging this for a bit as we're prioritizing some stability work internally. I'll update here once we're ready to accept this.

jurre (approved):

Issue #1245, opened 2019-07-04
Swift Package Manager Support?


satishbabariya:
Is there any plans for spm support?, some of the following details may help

1. use `Package.swift` in `file_fetcher`
2. use `.package(url: "https://github.com/abc/abc.git", from: "1.0.0)` for dependency in `file_parser`

Listing Dependency using Command Line (exec)

```
swift package dump-package --type json
```

For Package Metadata

```
swift package describe --type json
```

3. use git tag for checking new version update

I don't know much ruby, but I can help in creating specific dependency updater for package file.

greysteil:
We'd love to add spm support. Are there any docs and/or example `Package.swift` files you can link to?

satishbabariya:
here is the documentation on spm [https://github.com/apple/swift-package-manager/tree/master/Documentation](https://github.com/apple/swift-package-manager/tree/master/Documentation)

and here is the example of package file [https://github.com/vapor/vapor/blob/master/Package.swift](https://github.com/vapor/vapor/blob/master/Package.swift)

satishbabariya:
also here is the dockerfile in case
```dockerfile
FROM ubuntu:18.04

RUN export DEBIAN_FRONTEND=noninteractive DEBCONF_NONINTERACTIVE_SEEN=true && apt-get -q update && \
    apt-get -q install -y \
    libatomic1 \
    libbsd0 \
    libcurl4 \
    libxml2 \
    tzdata \
    && rm -r /var/lib/apt/lists/*

ARG SWIFT_PLATFORM=ubuntu18.04
ARG SWIFT_BRANCH=swift-5.0.1-release
ARG SWIFT_VERSION=swift-5.0.1-RELEASE

ENV SWIFT_PLATFORM=$SWIFT_PLATFORM \
    SWIFT_BRANCH=$SWIFT_BRANCH \
    SWIFT_VERSION=$SWIFT_VERSION

RUN SWIFT_URL=https://swift.org/builds/$SWIFT_BRANCH/$(echo "$SWIFT_PLATFORM" | tr -d .)/$SWIFT_VERSION/$SWIFT_VERSION-$SWIFT_PLATFORM.tar.gz \
    && curl -fSsL $SWIFT_URL -o swift.tar.gz \
    && export GNUPGHOME="$(mktemp -d)" \
    && tar -xzf swift.tar.gz --directory / --strip-components=1 \
    && rm -r "$GNUPGHOME" swift.tar.gz \
[...]

PR #5562, opened 2022-08-21
Add support for Swift Package Manager

rhwood:
This is #3772 but rebased against cafef062f5bb155e0735d6421a872f36d0240c19 (head of main as of 2022-08-21) because I couldn't figure out how to create a PR against the source of #3772 that didn't have merge conflicts.

PR #3772, opened 2021-05-21
Add support for Swift Package Manager

mattt:
Initial work to add Swift support to Dependabot.

This is currently a work-in-progress, but I'd welcome any feedback on what we have so far.

PR #7344, opened 2023-05-17
[WIP] Add Support for Swift Package Manager

0xTim:
⚠️ This PR is a work in progress ⚠️ 

This PR adds support for [Swift](https://www.swift.org) via the [Swift Package Manager](https://www.swift.org/package-manager/). This PR is heavily based upon #3772 and #5562 but updated for the latest SwiftPM changes and to fit in with the current requirements for language support in Dependabot.

Fix: https://github.com/dependabot/dependabot-core/issues/1245


jurre:
Awesome, thanks @0xTim! Looking forward to digging into this PR, it might take a few days before we get to it, but just wanted to let you know we're excited about this.

0xTim:
@jurre great! There's still a fair bit to do so I'll ping you when it's in a reasonable state!

lseppala:
This is an awesome addition! Excited to see it come together

One thing I'll note: XCode projects will use a `Package.resolved` file _without_ a `Package.swift`. The dependencies are directly managed in the XCode UI.

https://developer.apple.com/documentation/xcode/adding-package-dependencies-to-your-app#Coordinate-package-versions-across-your-team

While Swift libraries and often "Swift on Server" projects will use `Package.swift` to define Swift dependency requirements, XCode projects won't.

Apologies if you've already considered this and chose to [require Package.swift](https://github.com/dependabot/dependabot-core/pull/7344/files#diff-79aae8570fdc2a360d5752abfffb3c61b319eb7eb6618c36957df9d0e0dd7239R59) for other reasons. I wanted to raise it because this "2 ways of specifying dependencies" model has bitten me before!

0xTim:
@lseppala I would say Xcode projects are out of scope for the time being. I could look into it once the initial support is added but a) it's not a priority for me personally and b) trying to reverse engineer the Xcode project that can change with different Xcode versions might cause more problems. (I'm also half expecting in a couple of years that the Swift Playground app manifest will also be used for Xcode apps)

0xTim:
@deivid-rodriguez just checked and I have several unfinished spikes but nothing worth committing! Do you want to take this over wholesale or is the plan to collaborate and guide the PR through with contributions from me? (Either work to be clear!)

deivid-rodriguez:
Planning to take this over wholesale, but I'm sure I'll come back with questions about Swift is that's ok with you!

0xTim:
Ok great, reach out and I'll help where I can!

deivid-rodriguez:
One question @0xTim.

Current implementation resolves each package name from its url, using the `Dependabot::Swift::Package::Identifier` class, into a `<scope>.<name>` identifier (for example, `mona.linkedlist`).

However, I'm observing that the `Package.resolved` includes a explicit package name too:

```json
{
  "object": {
    "pins": [
      {
        "package": "LinkedList",
        "repositoryURL": "https://github.com/mona/LinkedList.git",
        "state": {
          "branch": null,
          "revision": "79b05a5da89557a4fa50647cc92488a3d9c3f478",
          "version": "1.2.2"
        }
      },
      ...
    ]
  }
}
````

Do you think it would be more preferable to use that instead (or just the full URL if there's no lockfile)? It seems simpler to me.

My assumption is that Swift either does not allow conflicting package names, or resolves them somehow, saving an unambiguous package name to the lockfile.

Thoughts?

deivid-rodriguez:
Thanks!

Another question. How do Swift users normally update dependencies?

What I gathered so far is:

* There's no `swift` command that changes the `Package.swift` file (for example to relax/update a requirement range to allow a new version). So when you need to update a dependency requirement, you manually edit this file.
* There's a `swift package update [package-name]` command that updates the `Package.resolved` file to the latest version that match the requirement in the `Package.swift` file.
* There's a `swift package resolve` command that brings the `Package.resolved` file up to date with the dependency specifications in `Package.swift`, trying to respect existing versions in `Package.resolved` as much as possible. I guess that's what you'd normally run after changing dependencies in `Package.swift` file.

Are the above correct? Is that how users normally manage swift dependencies?

stevapple:
> * There's no `swift` command that changes the `Package.swift` file (for example to relax/update a requirement range to allow a new version). So when you need to update a dependency requirement, you manually edit this file.

Correct for the time. There was [an accepted proposal](https://github.com/apple/swift-evolution/blob/main/proposals/0301-package-editing-commands.md) to add such functionality, but unfortunately it’s [not implemented](https://github.com/apple/swift-package-manager/pull/5802) yet. Until that lands developers have to edit the file manually.

> * There's a `swift package update [package-name]` command that updates the `Package.resolved` file to the latest version that match the requirement in the `Package.swift` file.

Correct. And there’s also a `swift package update` command that tries to update the resolved versions to latest for all packages. The result should be identical to running `swift package resolve` without `Package.resolved`. IMO this is likely to be the most popular way for updating dependencies during development?

> * There's a `swift package resolve` command that brings the `Package.resolved` file up to date with the dependency specifications in `Package.swift`, trying to respect existing versions in `Package.resolved` as much as possible. I guess that's what you'd normally run after changing dependencies in `Package.swift` file.

Exactly👍 SwiftPM will check if the resolved packages in `Package.resolved` matches the dependency graph of `Package.swift`, and if not it will run the resolver.

deivid-rodriguez:
Hei!

I'm working on swift updates at #7525, so let's close this PR and move the discussion there. I will add some attribution there to previous contributors of the initial skeleton :)

@steveapple I have some other questions in case you're still happy to help 😄.

stevapple:
> @steveapple I have some other questions in case you're still happy to help 😄.

Feel free to mention me:)

PR #7348, opened 2023-05-19
Upgrade Node.js to active LTS version (18.x)

christianvuerings:
# Changes

Upgrades the Node.js version that dependabot uses to the latest active LTS (Long Term Support) version which is 18.x. https://nodejs.org/en
image

# Why

* Dependabot currently uses a `Maintenance LTS` version of Node.js instead of a `Active LTS` version. https://nodejs.dev/en/about/releases/
image
* We run into the following issue since we have `engine-strict` enabled and want to avoid 
```
Dependabot uses Node.js v16.20.0
 and NPM 8.19.4
. Due to the engine-strict setting, the update will not succeed.
```

# Related issues / threads

* https://github.com/dependabot/dependabot-core/issues/4072#issuecomment-1413222679
* https://github.com/dependabot/dependabot-core/pull/7194 
* https://zenn.dev/risu729/articles/dependabot-engine-strict

### DIFF
--- a/npm_and_yarn/Dockerfile
+++ b/npm_and_yarn/Dockerfile
@@ -7,11 +7,11 @@ ARG PNPM_VERSION=8.6.4
 ARG YARN_VERSION=3.6.0
 
 # See https://github.com/nodesource/distributions#installation-instructions
-ARG NODEJS_VERSION=16.x
+ARG NODEJS_VERSION=18.x
 
 # Check for updates at https://github.com/npm/cli/releases
 # This version should be compatible with the Node.js version declared above. See https://nodejs.org/en/download/releases as well
-ARG NPM_VERSION=8.19.4
+ARG NPM_VERSION=9.5.1
 
 # Install Node and npm
 RUN curl -sL https://deb.nodesource.com/setup_$NODEJS_VERSION | bash - \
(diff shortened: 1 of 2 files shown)


npm_and_yarn/Dockerfile
```diff
@@ -8,11 +8,11 @@ ARG PNPM_VERSION=8.3.1
 ARG YARN_VERSION=3.5.0
 
 # See https://github.com/nodesource/distributions#installation-instructions
-ARG NODEJS_VERSION=16.x
+ARG NODEJS_VERSION=18.x
 
 # Check for updates at https://github.com/npm/cli/releases
 # This version should be compatible with the Node.js version declared above. See https://nodejs.org/en/download/releases as well
-ARG NPM_VERSION=8.19.4
+ARG NPM_VERSION=9.5.1
```
christianvuerings, line 15: See https://nodejs.org/en/download/releases
image

jurre:
Thanks, it probably makes sense to use that latest stable version 👍 It looks like you remove the package lock (by accident?), would you mind taking a look at that?

jurre:
It looks like this is breaking one of our tests, would you mind taking a look at that? 🙇

andrejleitner:
`npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater_spec.rb:2525`
```diff
- to eq("is-number@jonschlinkert/is-number#semver:^4.0.0")
+ to eq("is-number@github:jonschlinkert/is-number#semver:^4.0.0")
```

christianvuerings:
@jurre @andrejleitner I updated the PR - all tests are passing now

lizthegrey:
This is required to support the latest Yarn versions, which have dropped 16.x support.

jurre:
Thanks for the headsup, let me roll this out 👍

jurre (approved):

PR #7351, opened 2023-05-22
Fetching GitLab repo contents correctly uses the ref argument

maciej-gol:
Dependabot can properly list GitLab repo contents on a given branch/commit.

This is problematic when you are, for example, testing dependabot config and the requirements files are not present on the main repository branch, but on the changed branch only. In such case, dependabot won't see expected requirements, resulting in `Dependabot::DependencyFileNotFound`.

# Context
According to the GitLab API (https://docs.gitlab.com/ee/api/repositories.html#list-repository-tree), you have to use the `ref` argument to list the repository at a given commit/branch.

The GitLab client blindly passes all the arguments to the API itself, including the `ref_name` argument. GitLab ignores extra GET parameters, thus it wasn't erroring out.

### DIFF
--- a/common/lib/dependabot/file_fetchers/base.rb
+++ b/common/lib/dependabot/file_fetchers/base.rb
@@ -375,7 +375,7 @@ module Dependabot
 
       def _gitlab_repo_contents(repo, path, commit)
         gitlab_client.
-          repo_tree(repo, path: path, ref_name: commit, per_page: 100).
+          repo_tree(repo, path: path, ref: commit, per_page: 100).
           map do |file|
             # GitLab API essentially returns the output from `git ls-tree`
             type = case file.type


maciej-gol:
Mind having a look into this?

jurre:
The change looks good to me, we have some internal stability work that the team is prioritizing so I'll need to hold off merging this for a bit. One thing that I would love to see is some sort of test that would catch this in the future.

maciej-gol:
Will try to have a look into adding tests, Im not really proficient with ruby

jeffwidman (approved):
While a test would be nice, I'm personally okay shipping this as-is w/o any tests because the impact of a regression is relatively small:

* Won't impact our cloud service at GitHub at all
* Impact for GitLab users is primarily when their testing on a branch other than `main`
* It clearly hasn't been a blocker for folks so far, so it's more of a nice-to-have
* The Gitlab PR creator is very much community maintained.
* Long term, we _may_ eventually move away from Ruby-based PR creators to instead doing something with the [CLI](https://github.com/dependabot/cli) where it takes job output and has pluggable PR creators... that's a ways off, but it seems reasonable this code may not live forever.

Anyway, all that to say, I think this is fine to ship.

Thanks for tracking it down.

Issue #7071, opened 2023-04-12
Gemspec updating does not respect symlinks


mlarraz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

Bundler

### What you expected to see, versus what you actually saw

After https://github.com/dependabot/dependabot-core/pull/7051, path-based gems that are installed via symlink are having the symlink to their gemspec deleted and replaced with the a file with the updated contents.

Instead, Dependabot should respect the symlink, not delete it, and update the original gemspec at its source.

deivid-rodriguez:
Thanks for the notice @mlarraz!

I suspect this is an issue symlinks in general, just showed up for you  after we introduced #7051 because you symlink gemspecs.

Can you post an example public project reproducing the issue, so that I can use it to reproduce this and fix it?

laboone:
Hi, I am also experiencing the same issue.
Since there is no public repository that reproduces the problem, I will post if I am able to create a repository that can reproduce the issue.

laboone:
@deivid-rodriguez @jakecoffman 
I have created a repository that reproduces the issue: https://github.com/laboone/monorepo-for-dependabot
In this repository, Dependabot creates a PR that removes the symlink, as shown in https://github.com/laboone/monorepo-for-dependabot/pull/1.

I have also created a PR https://github.com/dependabot/dependabot-core/pull/7352 that reverts the changes made in https://github.com/dependabot/dependabot-core/pull/7051.
I would appreciate it if you could merge https://github.com/dependabot/dependabot-core/pull/7352 :pray:

laboone:
@deivid-rodriguez 
I believe this fix (https://github.com/dependabot/dependabot-core/pull/7411) is necessary. Could you please merge and release it?

deivid-rodriguez:
Sorry for the delay @laboone. Fix is live now, let me know how it's working!

PR #7051, opened 2023-04-11
Allow updating gemspecs loaded from a Gemfile

deivid-rodriguez:
This patch would need careful consideration since I guess there was a reason to forbid this in the first place.

But it makes Dependabot generate correct PRs for our development dependencies in /common.

Without this PR:

```
$ bin/dry-run.rb bundler dependabot/dependabot-core --dir /updater --dep rubocop-performance 
=> fetching dependency files
=> dumping fetched dependency files: ./dry-run/dependabot/dependabot-core/updater
=> parsing dependency files
=> updating 1 dependencies: rubocop-performance

=== rubocop-performance (1.16.0)
 => checking for updates 1/1
🌍 --> GET https://rubygems.org/api/v1/versions/rubocop-performance.json
🌍 <-- 200 https://rubygems.org/api/v1/versions/rubocop-performance.json
 => latest available version is 1.17.1
 => latest allowed version is 1.16.0
 => requirements to unlock: update_not_possible
 => requirements update strategy: bump_versions
    (no update possible 🙅‍♀️)
🌍 Total requests made: '1'
🎈 Package manager version log: {:ecosystem=>"bundler", :package_managers=>{"bundler"=>"2"}}
```

With this PR:

```
$ bin/dry-run.rb bundler dependabot/dependabot-core --dir /updater --dep rubocop-performance 
=> fetching dependency files
=> dumping fetched dependency files: ./dry-run/dependabot/dependabot-core/updater
=> parsing dependency files
=> updating 1 dependencies: rubocop-performance

=== rubocop-performance (1.16.0)
 => checking for updates 1/1
🌍 --> GET https://rubygems.org/api/v1/versions/rubocop-performance.json
🌍 <-- 200 https://rubygems.org/api/v1/versions/rubocop-performance.json
 => latest available version is 1.17.1
 => latest allowed version is 1.17.1
 => requirements to unlock: own
 => requirements update strategy: bump_versions
🌍 --> GET https://rubygems.org/api/v1/gems/rubocop-performance.json
🌍 <-- 200 https://rubygems.org/api/v1/gems/rubocop-performance.json
🌍 --> GET https://github.com/rubocop/rubocop-performance.git/info/refs?service=git-upload-pack
🌍 <-- 200 https://github.com/rubocop/rubocop-performance.git/info/refs?service=git-upload-pack
 => bump rubocop-performance from 1.16.0 to 1.17.1 in /updater

    ± Gemfile.lock
    ~~~
[...]

PR #7352, opened 2023-05-22
Revert "Allow updating gemspecs loaded from a Gemfile"

laboone:
### Issue
- https://github.com/dependabot/dependabot-core/issues/7071

### Description
When there is a Gem installed via a symlink, Dependabot removes the symlink.
I have created a repository that reproduces this issue: https://github.com/laboone/monorepo-for-dependabot
In this repository, Dependabot creates a PR that removes the symlink, as shown in https://github.com/laboone/monorepo-for-dependabot/pull/1.

This functionality seems to have been introduced in https://github.com/dependabot/dependabot-core/pull/7051.
This PR is a revert of that change.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,12 @@ updates:
     schedule:
       interval: "weekly"
 
+  # No ecosystem folders are watched because they roll up to the omnibus watcher
+  - package-ecosystem: "bundler"
+    directory: "/omnibus"
+    schedule:
+      interval: "weekly"
+
   # Watch the per-ecosystem native helpers
   - package-ecosystem: "composer"
     directory: "/composer/helpers/v1"
--- a/bundler/lib/dependabot/bundler/file_fetcher.rb
+++ b/bundler/lib/dependabot/bundler/file_fetcher.rb
@@ -142,7 +142,7 @@ module Dependabot
 
         raise Dependabot::PathDependenciesNotReachable, unfetchable_gems if unfetchable_gems.any?
 
-        gemspec_files
+        gemspec_files.tap { |ar| ar.each { |f| f.support_file = true } }
       end
 
       def path_gemspec_paths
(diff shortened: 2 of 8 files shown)


jeffwidman:
👋 The bug report makes sense, but we're in the middle of a deploy freeze for a little bit so I have to hold off shipping this to GitHub's service until that finishes.

Given we're already waiting on☝️, I'd also like @deivid-rodriguez to take a peek at this since he put up the original PR so may have ideas for a way to handle the symlink issue in a way that solves the original issue but also doesn't introduce new problems. He's traveling for the next two weeks, so I'll try to remember to remind him to look at this when he returns, but if I forget please ping in two weeks.

.github/dependabot.yml
```diff
@@ -5,6 +5,12 @@ updates:
     schedule:
       interval: "weekly"
 
+  # No ecosystem folders are watched because they roll up to the omnibus watcher
```
jurre, line 8: The commit that introduced this change seems unrelated, can we remove it?

deivid-rodriguez:
I had an initial look. I don't think this is an issue with #7051. I believe that bug fix is valid, just made this issue appeared.

I believe the problem is that we don't have logic to properly fetch files inside symlinked folders from the Github API. We only have logic to fetch directly symlinked files. I'm experimenting with a fix at #7411.

.github/dependabot.yml
deivid-rodriguez: Basically the bug fix in #7051 allowed us to no longer have to watch `/omnibus`. So technically, if we revert the bug fix, we should start watching that folder again.

But I don't think we should be reverting that bug fix, so I'm looking into properly fixing this bug, which seems related to symlinked folders.

deivid-rodriguez:
I think #7411 is the right fix, so I'd close this PR in favor of #7411. @laboone would you be able to try out that PR?

laboone:
@deivid-rodriguez 
Thank you for your investigation.
Based on your comment in https://github.com/dependabot/dependabot-core/pull/7411#issuecomment-1580767676, it seems to be working correctly.
I will close this PR:+1:

PR #7354, opened 2023-05-22
[Updater] Extract creation of new group Pull Requests into a discrete class

brrygrdn:
~~Follows on from https://github.com/dependabot/dependabot-core/pull/7310~~

This extracts `CreateGroupUpdatePullRequest` as a sibling class to `RefreshGroupUpdatePullRequest` and removes all responsibility for actually processing groups into PullRequests from `GroupUpdateAllVersions` so its sole responsibility is co-ordination of which strategy to use for each group.

### DIFF
--- a/updater/lib/dependabot/updater/operations/create_group_update_pull_request.rb
+++ b/updater/lib/dependabot/updater/operations/create_group_update_pull_request.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+
+require "dependabot/updater/group_update_creation"
+
+# This class implements our strategy for creating a single Pull Request which
+# updates all outdated Dependencies within a specific project folder that match
+# a specificed Dependency Group.
+#
+# This will always post a new Pull Request to Dependabot API and does not check
+# to see if any exists for the group or any of the dependencies involved.
+#
+module Dependabot
+  class Updater
+    module Operations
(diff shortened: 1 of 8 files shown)


updater/lib/dependabot/updater/operations/create_group_update_pull_request.rb
```diff
@@ -0,0 +1,71 @@
...
+#
+module Dependabot
+  class Updater
+    module Operations
+      class CreateGroupUpdatePullRequest
+        include GroupUpdateCreation
+
+        # We do not invoke this class directly for any jobs, so let's return false in the event this
+        # check is called.
+        def self.applies_to?(_job:)
```
bdragon, line 20: If the caller were to use `applies_to?(job:)` it would fail with an ArgumentError due to the `_job:` kwarg, which might be a roundabout way of discovering that this class should not be invoked directly.

bdragon (approved):

updater/lib/dependabot/updater/operations/create_group_update_pull_request.rb
brrygrdn: Good point, I was treating this as an unused positional argument, I'll fix this as I'm ok with this being called by accident and just returning false.

Nishnha (approved):
LGTM

PR #7358, opened 2023-05-22
Ensure updated dependencies are correctly included when building dependency change instance

bdragon:
Ensures that (a) a dependency whose version has changed but whose requirements have not changed is still included in the generated dependency change instance, and (b) a dependency whose requirements have changed but whose version has not changed is still included in the generated dependency change instance.

Should unblock https://github.com/dependabot/smoke-tests/pull/72

### DIFF
--- a/updater/lib/dependabot/dependency_change_builder.rb
+++ b/updater/lib/dependabot/dependency_change_builder.rb
@@ -38,7 +38,8 @@ module Dependabot
       updated_deps = updated_dependencies.reject do |d|
         # Avoid rejecting the source dependency
         next false if source_dependency_name && d.name == source_dependency_name
-        next true if d.top_level? && d.requirements == d.previous_requirements
+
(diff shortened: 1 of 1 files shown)


Nishnha (commented):
Nice! 

I had to walk through this with an example to really understand the change:

- Previously, for an update like `lodash ( from 4.17.15 to 4.17.21 )` with the requirement `^4.17.15`, the guard clause would reject the updated dependency because the requirement has not changed. 
- Now, with this patch, the guard clause will \*fail\* since the requirements have not changed. So instead of rejecting the dependency update early, the next line of code can execute.
- The next line of code, `d.version == d.previous_version` sees that the version did change and the dependency update is no longer rejected!

The (now passing) smoke test + the existing unit tests give me some confidence that this won't break anything, but I think we should monitor our dashboards closely when we roll this out

I think this also resolves https://github.com/dependabot/dependabot-core/issues/7356 ?

Nishnha (approved):

PR #7359, opened 2023-05-22
[Updater] Avoid mis-representing a Dependency Group as a Dependency in error handling

brrygrdn:
This resolves some `FIXME` statements where we were silently substituting the `DependencyGroup#name` for `Dependency#name` as a 'good enough' workaround for the `handle_dependabot_error` method requiring a dependency to be passed.

This PR splits the `handle_dependabot_error` into two flavours:
- handle_dependency_error - this is effectively the same as the existing `handle_dependabot_error` method
- handle_job_error - this is a separate signature for cases where the context does not include a specific Dependency which can be used in places where the `FIXME` was noted

As part of this I've also split the `log_error` method into `log_dependency_error` and `log_job_error` flavours for the same reason.

### Scope

For now, I've settled for sending the `dependency_group` as part of the information we sent to Sentry for unknown errors, omitting it if no group is involved. In a follow-up PR we will revise how the job summary is built up to make sure we explain which errors relate to groups as well as improve the PR summary.

### DIFF
--- a/updater/lib/dependabot/service.rb
+++ b/updater/lib/dependabot/service.rb
@@ -66,7 +66,7 @@ module Dependabot
     #
     # This should be called as an alternative/in addition to record_update_job_error
     # for cases where an error could indicate a problem with the service.
-    def capture_exception(error:, job: nil, dependency: nil, tags: {}, extra: {})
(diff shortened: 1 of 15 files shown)


Nishnha (approved):
🐝🐝🐝

PR #7365, opened 2023-05-23
[Grouped Updates] Generate deterministic branch names based on content

brrygrdn:
Grouped update Pull Requests will generally have more dependencies that we can tolerate in the branch name compared to solo dependency PRs, but we rely on deterministic branch names to avoid race conditions where we retry pushing a branch that has already been successfully pushed.

The current behaviour of appending a timestamp is resulting in us opening the odd duplicate when something causes us to backoff and partially retry opening the PR, see:
- https://github.com/github/dependabot-action/pull/949
- https://github.com/github/dependabot-action/pull/950

This fixes the bug by updating the branch naming strategy for groups to use a short ( 10 character ) digest hash of the dependency changes.

This will ensure that if we open a new PR for the same group with different versions or dependencies, it will still avoid clashing with any lingering old PRs that haven't been cleaned up while giving us back the guarantee of at-most-once pushes.

### DIFF
--- a/common/lib/dependabot/pull_request_creator/branch_namer/dependency_group_strategy.rb
+++ b/common/lib/dependabot/pull_request_creator/branch_namer/dependency_group_strategy.rb
@@ -21,7 +21,7 @@ module Dependabot
         # fixed-length name, so we can punt on handling truncation until
         # we determine the strict validation rules for names
         def new_branch_name
-          File.join(prefixes, timestamped_group_name).gsub("/", separator)
+          File.join(prefixes, group_name_with_dependency_digest).gsub("/", separator)
         end
 
         private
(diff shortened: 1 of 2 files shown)


common/lib/dependabot/pull_request_creator/branch_namer/dependency_group_strategy.rb
```diff
@@ -37,11 +37,20 @@ def prefixes
...
+        # given set of dependency changes.
+        #
+        # Let's append a short hash digest of the dependency changes so that we can
+        # meet this guarantee.
+        def group_name_with_dependency_digest
+          "#{dependency_group.name}-#{dependency_digest}"
+        end
+
+        def dependency_digest
+          @dependency_digest ||= Digest::MD5.hexdigest(dependencies.map do |dependency|
```
jakecoffman, line 51: Is the order of the dependencies deterministic? If not the hash will be different each run.

jakecoffman (approved):
LGTM

common/lib/dependabot/pull_request_creator/branch_namer/dependency_group_strategy.rb
brrygrdn: Fixed in https://github.com/dependabot/dependabot-core/pull/7365/commits/ae54458d5e371ffc9a3c5a187271c095b05c1b52

Nishnha (approved):

Issue #7319, opened 2023-05-15
Grouped bundler update fails with "undefined method `force_encoding' for nil:NilClass"


jdelStrother:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

Bundler

### Package manager version

2.4.12

### Language version

Ruby 3.2.2

### Manifest location and content before the Dependabot update

/Gemfile

### dependabot.yml content

```
version: 2

registries:
  contribsys: # Define access for Sidekiq/contribsys registry
    type: rubygems-server
    url: https://enterprise.contribsys.com/
    username: ${{secrets.SIDEKIQ_USERNAME}}
    password: ${{secrets.SIDEKIQ_PASSWORD}}

  ruby-github:
    type: rubygems-server
    url: https://rubygems.pkg.github.com/audioboom/
    token: ${{ secrets.GH_PACKAGE_REGISTRY_TOKEN }}

updates:
  - package-ecosystem: bundler
    directory: "/"
    schedule:
      interval: daily
    open-pull-requests-limit: 99
    # This seems bad, but is the default for rubygems.org, as far as I can tell.
    # So it's only necessary because we added contribsys for Sidekiq,
    # which I think we have to assume is probably fine.
    insecure-external-code-execution: allow
    registries:
      - contribsys
      - ruby-github
    vendor: true

    groups:
      rubocop-dependencies:
        patterns:
          - "rubocop*"
      rspec-dependencies:
        patterns:
          - "rspec*"
      cloud-service-dependencies:
        # Both AWS & Google Cloud have lots of auto-generated releases with not-very-helpful release notes.
        # Just group them into one big update
        patterns:
          - "aws-*"
          - "google-*"
      flipper-dependencies:
        patterns:
          - "flipper*"
      sidekiq-dependencies:
        patterns:
          - "sidekiq*"
      sentry-dependencies:
        patterns:
          - "sentry*"

```

### What you expected to see, versus what you actually saw

Since enabling grouped updates, our dependabot updates fail with - 

```
...
  proxy | 2023/05/14 07:59:44 [216] GET https://index.rubygems.org:443/versions
  proxy | 2023/05/14 07:59:44 [216] 304 https://index.rubygems.org:443/versions
  proxy | 2023/05/14 08:00:14 [218] GET https://rubygems.org:443/gems/rubocop-1.51.0.gem
[...]

PR #7372, opened 2023-05-25
Fix an issue where vendored gems are mistakenly parsed as Gemfiles

brrygrdn:
Potential fix for https://github.com/dependabot/dependabot-core/issues/7319

When updating using groups and vendored gems, we saw the following issue in the logs:

```
updater | 
updater |       string = string.dup.force_encoding(encoding)
updater |                          ^^^^^^^^^^^^^^^
updater | 2023/05/14 08:00:20 ERROR <job_662289885> /home/dependabot/dependabot-updater/vendor/ruby/3.1.0/gems/parser-3.2.2.0/lib/parser/base.rb:98:in `setup_source_buffer'
updater | 2023/05/14 08:00:20 ERROR <job_662289885> /home/dependabot/dependabot-updater/vendor/ruby/3.1.0/gems/parser-3.2.2.0/lib/parser/base.rb:32:in `parse'
updater | 2023/05/14 08:00:20 ERROR <job_662289885> /home/dependabot/bundler/lib/dependabot/bundler/file_parser/gemfile_declaration_finder.rb:45:in `parsed_gemfile'
up
```

The issue was that a `DependencyFile` with a `nil` value for `content` was passed into the `gemfile_declaration_finder` unexpectedly **after** we made the first change to the dependency files which debug showed us was:

```
2023/05/25 17:40:34 DEBUG Dependency files updated:
2023/05/25 17:40:34 DEBUG   - /Gemfile 
2023/05/25 17:40:34 DEBUG   - /Gemfile.lock ( Changed 1 times )
2023/05/25 17:40:34 DEBUG   - /vendor/cache/[redacted]-old-version ( Changed 1 times )
2023/05/25 17:40:34 DEBUG   - /vendor/cache/[redacted]-new-version ( Changed 1 times )
```

Upon investigating further, I realised that we currently handle all `DependencyFile` objects passed back from the updater in all ecosystems as being dependency **manifests**, but this isn't true when vendoring is involved and in this case by including the vendored file change in the subsequent step it was causing the Bundler implementation to try to parse it as if it was a Gemfile.

I was able to fix the problem using this change by removing the vendored files from being passed to subsequent steps **but I'm not sure this is the right approach** - it seems like something else is wrong inside the bundler library that we are finding a vendored file instead of the Gemfile as we should be looking through the set rather than using an arbitrary file, e.g. first or last in array.

I need to add a simple test case that captures the problem to iterate on this a little more.

### Tasks

- [ ] Add a test case for Gemfile.lock-only changes with vendored files
- [ ] Verify it is red with existing behaviour due to the nil content error
- [ ] Determine _why_ the wrong file is being used as a Gemfile
- [ ] Figure out if we should partition file changes like this or drop that part of the change

### DIFF
--- a/updater/lib/dependabot/updater/group_dependency_file_batch.rb
+++ b/updater/lib/dependabot/updater/group_dependency_file_batch.rb
@@ -10,6 +10,8 @@ module Dependabot
           hash[file.path] = { file: file, changed: false, changes: 0 }
         end
 
+        @other_updated_files = {}
+
         Dependabot.logger.debug("Starting with '#{@dependency_file_batch.count}' dependency files:")
         debug_current_state
       end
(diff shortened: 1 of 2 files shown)


brrygrdn:
This has been superseded by two alternative fixes:
- https://github.com/dependabot/dependabot-core/pull/7404
- https://github.com/dependabot/dependabot-core/pull/7405

PR #7381, opened 2023-05-31
Set file encoding in GitLab commits

mikaellanger:
request_{creator,updater}/gitlab.rb doesn't pass the DependencyFile content_encoding to the GitLab API. This causes vendored files (in our case yarn pnp zip files) to be uploaded as text.

### DIFF
--- a/common/lib/dependabot/pull_request_creator/gitlab.rb
+++ b/common/lib/dependabot/pull_request_creator/gitlab.rb
@@ -109,7 +109,7 @@ module Dependabot
             action: file_action(file),
             file_path: file.type == "symlink" ? file.symlink_target : file.path,
             content: file.content,
-            **file_encoding(file)
+            encoding: file.content_encoding
           }
         end
       end
@@ -122,15 +122,6 @@ module Dependabot
           "create"
         else
           "update"
-        end
-      end
-
-      # @param [DependencyFile] file
-      def file_encoding(file)
-        if file.content_encoding == Dependabot::DependencyFile::ContentEncoding::BASE64
-          { encoding: "base64" }
-        else
-          {}
         end
       end
 
--- a/common/lib/dependabot/pull_request_creator/gitlab.rb
+++ b/common/lib/dependabot/pull_request_creator/gitlab.rb
@@ -128,7 +128,7 @@ module Dependabot
       # @param [DependencyFile] file
       def file_encoding(file)
         if file.content_encoding == Dependabot::DependencyFile::ContentEncoding::BASE64
-          {encoding: "base64"}
+          { encoding: "base64" }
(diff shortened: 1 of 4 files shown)


common/lib/dependabot/pull_request_creator/gitlab.rb
```diff
@@ -124,6 +125,15 @@ def file_action(file)
         end
       end
 
+      # @param [DependencyFile] file
+      def file_encoding(file)
+        if file.content_encoding == Dependabot::DependencyFile::ContentEncoding::BASE64
```
jurre, line 130:
```suggestion
        if file.binary?
```

common/lib/dependabot/pull_request_updater/gitlab.rb
```diff
@@ -94,6 +95,15 @@ def file_action(file)
           "update"
         end
       end
+
+      # @param [DependencyFile] file
+      def file_encoding(file)
+        if file.content_encoding == Dependabot::DependencyFile::ContentEncoding::BASE64
```
jurre, line 101:
```suggestion
        if file.binary?
```

jurre (commented):
Overall changes look good, I made a few suggestions and had one question, lmk your thoughts.

Internally we have a few stability improvements that we need to focus on first before we can accept any contributions, so it'll take a bit of time before we can merge this, just FYI.

jeffwidman (approved):

mikaellanger:
> Overlooked that Jurre's suggestions on binary weren't fully implemented... please do that first.

@jeffwidman Not sure I understand what you mean? That code was  removed in favor of setting encoding directly to file.content_encoding per the other comment by Jurre https://github.com/dependabot/dependabot-core/pull/7381#discussion_r1218309197

common/lib/dependabot/pull_request_updater/gitlab.rb
```diff
@@ -94,6 +95,7 @@ def file_action(file)
           "update"
         end
       end
+
```
jeffwidman, line 98:
```suggestion
```

jeffwidman (approved):
Thanks!

fadedDexofan:
@jeffwidman GitLab API docs says that encoding should be either text or base64, not utf-8. Text is default. Setting `encoding` to utf-8 breaks dependabot-updater.
```/builds/services-devops/kira-dependabot-script/vendor/ruby/3.1.0/gems/gitlab-4.19.0/lib/gitlab/request.rb:71:in `validate': Server responded with code 400, message: actions[0][encoding] does not have a valid value. Request URI: <REPO_PATH>/repository/commits (Gitlab::Error::BadRequest)```

PR #7400, opened 2023-06-05
Allow parsing symbols, time and date values in workflow yaml

jurre:
In github_actions, it's valid to use symbol, date and time objects in your workflow yaml files. Up until now, Dependabot was failing parsing on these, as the classes are by default not allowed by `YAML.safe_load`, this changes that to match the behavior in GitHub.

### DIFF
--- a/github_actions/lib/dependabot/github_actions/file_parser.rb
+++ b/github_actions/lib/dependabot/github_actions/file_parser.rb
@@ -38,7 +38,7 @@ module Dependabot
       def workfile_file_dependencies(file)
         dependency_set = DependencySet.new
 
-        json = YAML.safe_load(file.content, aliases: true)
+        json = YAML.safe_load(file.content, aliases: true, permitted_classes: [Date, Time, Symbol])
         return dependency_set if json.nil?
 
         uses_strings = deep_fetch_uses(json.fetch("jobs", json.fetch("runs", nil))).uniq
--- a/github_actions/spec/fixtures/workflow_files/workflow.yml
+++ b/github_actions/spec/fixtures/workflow_files/workflow.yml
@@ -112,3 +112,16 @@ jobs:
     steps:
       - name: Set up node without quotes
         uses: actions/setup-node@master
+
+  set-up-node-with-colon:
+    name: :set-up-node-with-colon
+
+  set-up-node-with-date:
+    name: "Set up node with date"
+    env:
+      START_DATE: 2016-01-01
+
+  set-up-node-with-time:
+    name: "Set up node with time"
+    env:
+      NYC_BALL_DROP: 2022-01-01T05:00:00Z


github_actions/lib/dependabot/github_actions/file_parser.rb
```diff
@@ -38,7 +38,7 @@ def parse
       def workfile_file_dependencies(file)
         dependency_set = DependencySet.new
 
-        json = YAML.safe_load(file.content, aliases: true)
+        json = YAML.safe_load(file.content, aliases: true, permitted_classes: [Date, Time, Symbol])
```
landongrindheim, line 41: 👍 This makes sense to me! My understanding is that we're broadening what we will accept. Is that right?

In the event that an invalid value is provided, are readable errors pushed up the stack?

landongrindheim (approved):

github_actions/lib/dependabot/github_actions/file_parser.rb
jurre: What sort of invalid value are you thinking here, and which stack?

landongrindheim: > What sort of invalid value are you thinking here
Sorry, my question has less to do with reviewing this PR and more to do with understanding how `YAML.safe_load` works. I'm questioning in order to learn, not to refine the change. For invalid values, let's use a `Symbol` _before_ this change was applied.

> which stack?
I probably used the wrong terminology here. I'm wondering who/what needs to handle the errors from `YAML.safe_load`. Again, not a question to refine the PR, but a question to learn.

jurre: > For invalid values, let's use a Symbol before this change was applied.

You'd end up with a ParseError:

```bash
~ % cat symbols.yaml
set-up-node-with-colon:
  name: :set-up-node-with-colon
~ % pwd
/Users/jurre
~ % irb
irb(main):001:0> require "YAML"
=> true
irb(main):002:0> YAML.safe_load(File.open("/Users/jurre/symbols.yaml"))
	... 19 levels...
	16: from /Users/jurre/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/psych-3.3.0/lib/psych/visitors/visitor.rb:30:in `visit'
	15: from /Users/jurre/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/psych-3.3.0/lib/psych/visitors/to_ruby.rb:166:in `visit_Psych_Nodes_Mapping'
	14: from /Users/jurre/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/psych-3.3.0/lib/psych/visitors/to_ruby.rb:341:in `revive_hash'
	13: from /Users/jurre/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/psych-3.3.0/lib/psych/visitors/to_ruby.rb:341:in `each_slice'
	12: from /Users/jurre/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/psych-3.3.0/lib/psych/visitors/to_ruby.rb:341:in `each'
	11: from /Users/jurre/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/psych-3.3.0/lib/psych/visitors/to_ruby.rb:343:in `block in revive_hash'
	10: from /Users/jurre/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/psych-3.3.0/lib/psych/visitors/to_ruby.rb:34:in `accept'
	 9: from /Users/jurre/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/psych-3.3.0/lib/psych/visitors/visitor.rb:6:in `accept'
	 8: from /Users/jurre/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/psych-3.3.0/lib/psych/visitors/visitor.rb:30:in `visit'
	 7: from /Users/jurre/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/psych-3.3.0/lib/psych/visitors/to_ruby.rb:127:in `visit_Psych_Nodes_Scalar'
	 6: from /Users/jurre/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/psych-3.3.0/lib/psych/visitors/to_ruby.rb:64:in `deserialize'
	 5: from /Users/jurre/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/psych-3.3.0/lib/psych/scalar_scanner.rb:74:in `tokenize'
	 4: from /Users/jurre/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/psych-3.3.0/lib/psych/class_loader.rb:84:in `symbolize'
	 3: from /Users/jurre/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/psych-3.3.0/lib/psych/class_loader.rb:32:in `symbolize'
	 2: from (eval):2:in `symbol'
	 1: from /Users/jurre/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/psych-3.3.0/lib/psych/class_loader.rb:28:in `load'
/Users/jurre/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/psych-3.3.0/lib/psych/class_loader.rb:99:in `find': Tried to load unspecified class: Symbol (Psych::DisallowedClass)
```

As for error handling, this ends up being an "not parseable" error as far as Dependabot is concerned, it treats the workflow file as if it was malformed, it's not a great experience.

PR #7404, opened 2023-06-05
[Grouped Updates] Avoid passing non-manifest file changes between group updates

brrygrdn:
Historically, our libraries receive a list of `Dependabot::DependencyFile` objects at the start of an operation which contains the files we have explicitly fetched from the repository which represent the manifest declaration.

When we update files, we return a new list of `Dependabot::DependencyFile` objects which includes the manifest files we have modified along with any vendored files we have changed.

Grouped Updates modifies this contract by starting to pass the `Dependency::DependencyFile` list returned from one update as input into the next, but this has the unintended consequence that we now start to inject vendored file changes into this input list for the first time.

### The Problem

When we create a `Dependabot::DependencyFile` object for a vendored dependency we encode the content in Base64 so if it is naively copied into the file system without decoding the content it creates a junk file that will either cause an error or be removed by the package manager.

Additionally, this file maybe actually represent a deletion.

### The Fix

This change reverts to our current contract that the **input dependency_files list will never include vendored files**.

This will resolve the immediate problem with Grouped Updates + Bundler + Vendoring we are seeing in https://github.com/dependabot/dependabot-core/issues/7319.

The downside of this fix is that it means each step will need to redo the changes to any vendored files within each step which will result in significant rework for large dependency groups.

### Next Steps

We could ship this fix in isolation but we'd be potentially trading problems in Bundler for an inefficient update process as well as possible unforeseen consequences in other ecosystems that are less graceful about the vendor files being stale.

The full remediation will require an additional change that integrates the library code added in https://github.com/dependabot/dependabot-core/pull/6693 which I will open downstream of this to simplify code review.


brrygrdn:
We're going to go with shipping https://github.com/dependabot/dependabot-core/pull/6693 to make environments retail their vendored file changes in git to avoid churning the ecosystem interfaces.

Once that PR has merged, I'll update this to trigger a workspace per group and rewrite the description to reflect the final state.

updater/spec/dependabot/updater/operations/group_update_all_versions_spec.rb
```diff
@@ -289,4 +289,72 @@
...
+      job_definition_fixture("bundler/version_updates/group_update_all_with_vendoring")
+    end
+
+    let(:dependency_files) do
+      # Let's use the already up-to-date files
+      original_bundler_files(fixture: "bundler_vendored")
+    end
+
+    before do
+      stub_rubygems_calls
```
Nishnha, line 312: I think you can update [#stub_rubygems_calls](https://github.com/dependabot/dependabot-core/blob/f602c6372fb127a4470481ed3702be82530b53d8/updater/spec/support/dummy_pkg_helpers.rb#L8) to also stub out the vendored gem. The VCR returns a `401 repository not found` right now, so you probably don't need it?

updater/spec/dependabot/updater/operations/group_update_all_versions_spec.rb
```diff
@@ -289,4 +289,72 @@
...
+        new_dummy_pkg_b = dependency_change.updated_dependency_files.find do |file|
+          file.path == "/vendor/cache/dummy-pkg-b-1.2.0.gem"
+        end
+        expect(new_dummy_pkg_b.operation).to eql("create")
+
+        # We've deleted the old version of the vendored git dependency
+        old_git_dependency_files = dependency_change.updated_dependency_files.select do |file|
+          file.path.start_with?("/vendor/cache/ruby-dummy-git-dependency-20151f9b67c8")
+        end
+        expect(old_git_dependency_files.map(&:operation)).to eql(%w(delete delete))
```
Nishnha, line 348: for other reviewers: the two `delete` operations here correlate to the bundlecache and gemspec files being deleted.

Nishnha:
Couple questions:

> The downside of this fix is that it means each step will need to redo the changes to any vendored files within each step

Can you point out where this happens? From my read through the vendored dependencies seem to only update once.

Did you add the `change_count` to track how many times the vendored files' changes are redone? Or does it serve some other purpose?

updater/spec/dependabot/updater/operations/group_update_all_versions_spec.rb
brrygrdn: Yeah, that's a bug that the smoke tests are tripping into, I need to update the test to use a 'real' gem instead.

brrygrdn:
> > The downside of this fix is that it means each step will need to redo the changes to any vendored files within each step
>
> Can you point out where this happens? From my read through the vendored dependencies seem to only update once.

I did a better job of explaining this in the spike to check in the vendored file changes: https://github.com/dependabot/dependabot-core/pull/7469

Essentially, it happens within the native tool itself, we reset the files on disk between updates - removing any vendored file changes and forcing it to replace the files from all previous steps. This is essentially opaque to us in terms of how each ecosystem copes, so the best way to avoid random behaviour is to ensure we keep the files-on-disk up to date using git.

brrygrdn:
> Did you add the change_count to track how many times the vendored files' changes are redone? Or does it serve some other purpose?

It helps us count the total number of changes to the manifests - so if we tried to update N dependencies, we should expect a manfest to change N times in most cases, but we might find this isn't true so it helps us figure out which files change in each step when debug is enabled in cases where we are dealing with things like nested or peer manifests ( e.g. Docker )

updater/spec/dependabot/updater/operations/group_update_all_versions_spec.rb
brrygrdn: Smoke test output for reference here: https://github.com/dependabot/smoke-tests/pull/84

deivid-rodriguez (approved):
This makes sense to me @brrygrdn. Also nice job with the test 👏

updater/spec/dependabot/updater/operations/group_update_all_versions_spec.rb
brrygrdn: I've investigated this a bit more and realised it's not what I thought - I'm going to leave the stubs for this alone for now since we're tolerating it via VCR and it is kind of inconsequential to the test for now.

Nishnha (approved):
This makes sense to me. I did have some trouble grasping the gist of the issue but the description in https://github.com/dependabot/dependabot-core/pull/7469 was indeed helpful. What made it really click was looking at how the `DependencyGroupChangeBatch` was actually being used in [GroupUpdateCreation](https://github.com/dependabot/dependabot-core/blob/main/updater/lib/dependabot/updater/group_update_creation.rb#L21C9-L23)

I like that this interface makes the handling of vendored dependencies invisible to the caller.

PR #7405, opened 2023-06-05
[Grouped Updates] Bundler handles previous vendor changes gracefully for grouped updates

brrygrdn:
Potential fix for https://github.com/dependabot/dependabot-core/issues/7319

This branch makes our bundler implementation aware that the input `dependency_files` array may now contain changes to `vendor/cache` during grouped updates and avoids treating them as "evaled_gemfiles" by filtering them out within the various places we implement this.

This is only a partial solution at time of writing as it competes with https://github.com/dependabot/dependabot-core/pull/7404 but has the disadvantage that it does not prepare the `vendor/cache` directory properly with previous changes resulting in rework on each subsequent update.


brrygrdn:
Closing this out as we are not going to pursue this remediation path.

Issue #7048, opened 2023-04-11
Requires an environment variable to indicate that dependabot is now running on github.


shu-mutou:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

My dependabot check needs to prevent running `postinstall` script of `yarn` when dependabot check. Because my dependabot check occurs error during postinstall. Also, no postinstall is required during my dependabot check.

If there is an environment variable that indicate running dependabot, we will be able to skip contents of postinstall like follow:

```
   postinstall: test -n "${DEPENDABOT_RUNNING}" || <original scripts>
```

Thanks!

shu-mutou:
Any other way to know dependabot is running please let me know. Thanks!

shu-mutou:
I found `DEPENDABOT_HOME` is useful for this issue.

I created following script. and call it from `postinstall`. This workaround seems to work well!! 🎉 

```
if [ -z "${DEPENDABOT_HOME}" ]; then
  # TODO: if dependabot will support yarn postinstall properly,
  #       move following line into .yarnrc.yml.
  npx ngcc && node ./hack/version.mjs && cd ../../ && npx husky install modules/web/.husky
  exit $?
fi

exit 0
```

PR #7407, opened 2023-06-06
Add `DEPENDABOT` environment variable for users

shu-mutou:
Describe usage of `DEPENDABOT_HOME` environment variable to determine whether Dependabot is running or not.

in favor of https://github.com/dependabot/dependabot-core/issues/7048

### DIFF
--- a/Dockerfile.updater-core
+++ b/Dockerfile.updater-core
@@ -46,6 +46,9 @@ USER dependabot
 ENV DEPENDABOT_HOME="/home/dependabot"
 WORKDIR $DEPENDABOT_HOME
 
+# For users to determine if dependabot is running
+ENV DEPENDABOT=true
+
 # Disable automatic pulling of files stored with Git LFS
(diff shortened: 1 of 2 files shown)


jurre:
This definitely works, but it's more of an implementation detail on how we've set up Dependabot and it could change going forward 🤔 

If i understand correctly, you have some `yarn` plugins that are ran when dependabot envokes yarn during it's update process, and you would like to prevent those scripts from executing in this scenario, right? I think that's a very reasonable request, but rather than reusing an implementation detail like the `DEPENDABOT_HOME` path that we set for internal use, what if we added a new `export DEPENDABOT=true` environment variable specifically for this, and document that as such?

shu-mutou:
Thank you for your perfect understanding.
As you said, I would like to add a new environment variable to a location similar to DEPENDABOT_HOME. I'll update PR, thanks!

shu-mutou:
@jurre PTAL, thanks!!

Issue #7384, opened 2023-06-01
[github-actions] dependabot not updating https://github.com/Swatinem/rust-cache


tamird:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

github-actions

### Manifest location and content before the Dependabot update

There are a few affected manifests in aya-rs/aya: https://github.com/search?q=repo%3Aaya-rs%2Faya%20Swatinem%2Frust-cache&type=code, and one in aya-rs/bpf-linker: https://github.com/aya-rs/bpf-linker/blob/7847c0b2959ca517779ff569ef0d231e6734fad8/.github/workflows/build-test.yml#L39-L40

### dependabot.yml content

https://github.com/aya-rs/aya/blob/main/.github/dependabot.yml

### Updated dependency

https://github.com/Swatinem/rust-cache (v1 vs v2)

### What you expected to see, versus what you actually saw

I expected an update, but none happened.

Drarig29:
We have a similar issue, where we recently set up dependabot to work on 2 repositories with the exact same `dependabot.yml` file, and only one repository had dependabot pull requests opened.

This file was recently updated with `v3` for `checkout` and `setup-node`:

https://github.com/DataDog/datadog-ci-azure-devops/blob/25703a397130d0ff50bac161eb0f49574fe1aac5/.github/workflows/unit-tests.yml

But this file (in a different repository) did not, although they are very similar:

https://github.com/DataDog/synthetics-ci-github-action/blob/dd7049d88a128ad2c01fe567792ab2a25f918eef/.github/workflows/unit-tests.yml

Is there a way to see dependabot's debug logs?

tamird:
@Drarig29 I think your issues are caused by a separate problem (which I also have seen in https://github.com/aya-rs/bpf-linker/pull/47): your repo contains multiple different workflows, and some of them reference `checkout@v2` while others reference `checkout@v3`: https://github.com/search?q=repo%3ADataDog%2Fsynthetics-ci-github-action%20actions%2Fcheckout&type=code.

You can see the dependabot logs for your repo at https://github.com/DataDog/datadog-ci-azure-devops/network/updates - click on the "last checked ..." link.

It's probably worth finding that log (it will contain some errors) and reporting a separate bug. I should have done that - now that I manually bumped the dep in https://github.com/aya-rs/bpf-linker/pull/47, my old log is gone.

Drarig29:
[...]

PR #7410, opened 2023-06-06
Fix edge case when updating Actions with mixed versions

deivid-rodriguez:
Previously we would avoid updating requirements if the commit for the dependency does not match the commit for the updated reference.

The commit for the dependency is guessed by the `GitCommitChecker` using the _first_ source as the dependency source. However, this can be inaccurate when updating dependencies with sources with mixed versions, since they may point to different commits.

I think this is an issue in the `GitCommitChecker` but I also noticed that these checks should not be necessary, we can unconditionally update the source, even if the commits don't match.

Fixes https://github.com/dependabot/dependabot-core/issues/7384.

### DIFF
--- a/github_actions/lib/dependabot/github_actions/update_checker.rb
+++ b/github_actions/lib/dependabot/github_actions/update_checker.rb
@@ -182,15 +182,13 @@ module Dependabot
 
         # Return the git tag if updating a pinned version
         if git_commit_checker.pinned_ref_looks_like_version? &&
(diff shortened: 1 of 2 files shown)


jurre:
> However, this can be inaccurate when updating dependencies with sources with mixed versions, since they may point to different commits.

I'm not quite sure that I understand why this is the case @deivid-rodriguez, could you expand on it a bit so I can learn?

PR #7412, opened 2023-06-07
Target latest Python versions - 3.11.4, 3.10.12, 3.9.17, 3.8.17, 3.7.17

phillipuniverse:
Same as #7013

Waiting on:

- [x] Python 3.11.4 released
- [x] Pyenv 2.3.19 released


phillipuniverse:
Hm I think with the pip upgrade that happened with the Python upgrade, we are now running into https://github.com/jazzband/pip-tools/issues/1825#issue-1634660340 from `pip-tools`. I'll investigate a bit more.

phillipuniverse:
Looks like that was the issue! When we go and update the dependencies in requirements.txt when dropping Python 3.6 compatibility, I think they should be updated like this:

```
pip>=21.3.1
pip-tools>=6.13.0
```

Not 100% sure the `pip-tools` pip versions of compatibilities, could also update to:

```
pip>=23.1.0
pip-tools>=6.13.0
```

phillipuniverse:
The pip-tools incompatibility got resolved in #7034. This should not be good to go!!

jurre (approved):

phillipuniverse:
Smoke tests will be fixed by https://github.com/dependabot/smoke-tests/pull/75

Issue #7389, opened 2023-06-03
Poetry >=1.5.0 removes category from poetry.lock


esev:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

pip (poetry)

### Package manager version

poetry >=1.5.0

### Language version

python 3.11

### Manifest location and content before the Dependabot update

https://github.com/pywemo/pywemo/blob/main/poetry.lock

Previous format: 

```toml
[[package]]
name = "cryptography"
version = "40.0.2"
description = "cryptography is a package which provides cryptographic recipes and primitives to Python developers."
category = "dev"
optional = false
python-versions = ">=3.6"
files = [
...
```

Format after Poetry 1.5.0 (lacks a `category`): https://github.com/pywemo/pywemo/pull/415/commits/65969c987ea6e79124fe639ea79a71b4e1f09967 (poetry.lock)

```toml
[[package]]
name = "cryptography"
version = "40.0.2"
description = "cryptography is a package which provides cryptographic recipes and primitives to Python developers."
optional = false
python-versions = ">=3.6"
files = [
...
```

### dependabot.yml content

```yaml
version: 2
updates:
  - package-ecosystem: "pip"
    directory: "/"
    schedule:
      # Check for updates to Poetry dependencies every week
      interval: "weekly"

  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      # Check for updates to GitHub Actions every week
      interval: "weekly"
```

### Updated dependency

cryptography

### What you expected to see, versus what you actually saw

Expected to see dependabot update for cryptography.

What happened:

Logs: https://github.com/pywemo/pywemo/security/dependabot/6/update-logs/336414827
https://github.com/dependabot/dependabot-core/blob/d097b116b78cadc36ec75d5e898c81ca3a72b28a/python/lib/dependabot/python/update_checker/poetry_version_resolver.rb#L300-L304

```
updater | 2023/06/02 20:34:03 INFO Raven 3.1.2 ready to catch errors
updater | 2023/06/02 20:34:04 INFO <job_672681628> Starting job processing
updater | 2023/06/02 20:34:08 INFO <job_672681628> Starting update job for pywemo/pywemo
updater | 2023/06/02 20:34:08 INFO <job_672681628> Checking if cryptography 40.0.2 needs updating
  proxy | 2023/06/02 20:34:08 [026] GET https://pypi.org:443/simple/cryptography/
[...]

Issue #6659, opened 2023-02-13
Poetry +1.2 is not updating group.dev.dependencies when using new group syntax vs dev-dependencies


magame95:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

pip (poetry)

### Package manager version

Poetry 1.2

### Language version

Python 3.9

### Manifest location and content before the Dependabot update

Location: `/pyproject.toml`
Content:
```
[tool.poetry]
name = "dependabot-poetry13"
version = "0.1.0"
description = ""
authors = ["magame95 <galavix.mendez@gmail.com>"]
readme = "README.md"
packages = [{include = "dependabot_poetry13"}]

[tool.poetry.dependencies]
python = "^3.9"
django = "^4.1.6"


[tool.poetry.group.dev.dependencies]
flake8 = "^5.0.4"
pre-commit = "^2.21.0"

[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```

### dependabot.yml content

```
version: 2
updates:
  - package-ecosystem: "pip"
    directory: "/"
    schedule:
      interval: "daily"
```

### Updated dependency

flake8 5.0.4 to 6.0.0
pre-commit 2.21.0 to 3.0.4

### What you expected to see, versus what you actually saw

Expected:
Dependabot creates 2 PRs
`Bump flake8 from 5.0.4 to 6.0.0`
`Bump pre-commit from 2.21.0 to 3.0.4`

Resulted:
Dependabot determined that updated was not possible (from logs).

### Images of the diff or a link to the PR, issue, or logs

I created 2 test repos to demonstrate de difference between dependabot usage between poetry new group and old dev-dependencies syntax.
https://github.com/magame95/dependabot-poetry1.3
https://github.com/magame95/dependabot-poetry12 (Note how dependabot opened 2 PRs for the dependencies in question)

Also, if you see at the [Insights](https://github.com/magame95/dependabot-poetry12/network/dependencies) for dev-dependencies, its displaying development dependencies under Dependencies defined in [pyproject.toml](https://github.com/magame95/dependabot-poetry12/blob/main/pyproject.toml) in constrast to the group syntax repo [Insights](https://github.com/magame95/dependabot-poetry1.3/network/dependencies)

new group.dev.dependencies syntax
![image](https://user-images.githubusercontent.com/88177250/218582880-5895aead-7417-42b0-8640-5034f0484859.png)
old dev-dependencies syntax
[...]

PR #7418, opened 2023-06-09
Support Poetry 1.5.1 poetry.lock with absence of category key

phillipuniverse:
Follow-up to the original Poetry 1.5 merge at #7350 and the smoke test updates at https://github.com/dependabot/smoke-tests/pull/75. Removal of the category key happened in the 1.5.0 release at https://github.com/python-poetry/poetry/releases/tag/1.5.0.

Fixes https://github.com/dependabot/dependabot-core/issues/7389
Fixes https://github.com/dependabot/dependabot-core/issues/7641


phillipuniverse:
Another thought I had looking at this. In `requirements.txt` we specify that we want to install a flexible version of Poetry:

https://github.com/dependabot/dependabot-core/blob/edd181f894d4642cf71972df7f46a738738fd4a6/python/helpers/requirements.txt#L6

But Dependabot isn't _actually_ operating against any version other than 1.5.1, so this `requirements.txt` is a little misleading. I think it would make more sense to have this line be:

```
poetry==1.5.1
```

To make it clear which version Dependabot is operating with. Then update PRs like #7350 don't need to widen, they just need to update this static version.

Thoughts?

phillipuniverse:
The test failures are what I was afraid of. Looks like Dependabot has special logic on the `category` key existing in the `poetry.lock` file and Poetry has removed that functionality:

https://github.com/dependabot/dependabot-core/blob/edd181f894d4642cf71972df7f46a738738fd4a6/python/lib/dependabot/python/file_updater/poetry_file_updater.rb#L167-L174

So I think that the current Dependabot code will not work with a Poetry 1.5 lockfile and more needs to be done to make it compatible.

phillipuniverse:
I’m not exactly sure how this needs to change. It seems like this was designed to work in the absence of a pyproject.toml file, but now the dependency group information is only managed there. Should we move the check to just look at pyproject.toml instead?

phillipuniverse:
I looked a bit more into this and we are butting up against multiple tech debt issues here with Poetry changing things. Since the category is removed from `poetry.lock`, the only way we can obtain this information is via `pyproject.toml`. The current code paths I believe allow only a `poetry.lock`, so making changes to that core assumption could break other workflows.

Determining the category from `pyproject.toml` is also a little tricky considering older Poetry versions. Older versions of Poetry (< 1.2.0) had you specify dependencies as `dev-dependencies`, whereas Poetry > 1.2.0 uses an arbitrary group. 

Here is a good example that illustrates the different ways to specify dependencies (`dev-dependencies` is deprecated but not removed in Poetry yet):

https://github.com/dependabot/dependabot-core/blob/19f4402040452622cfd65095ceffc133b0cdcfbf/python/spec/fixtures/pyproject_files/poetry_group_dependencies.toml#L1-L20

Support for group dependencies [was added in this PR](https://github.com/dependabot/dependabot-core/pull/6417) but works around the issue by continuing to rely on the `category` in `poetry.lock`. That will also need to be reworked a bit to rely on `pyproject.toml`.

Finally, there is some core GitHub integration at play to be able to parse out dependencies in the 'Insights' tab. This was called out at https://github.com/dependabot/dependabot-core/issues/6659#issuecomment-1431632688 and something might have to change there  depending on the reliance on `poetry.lock` to get through those insights.

---

This PR needs additional guidance from the Dependabot maintainers for how to proceed.

pavera:
As far as the `requirements.txt` specification the range is there to allow support under python 3.6. If dependabot detects that a given repository is using python 3.6 it will install a python 3.6 virtualenv and then `pip install` the `requirements.txt`. The result is that on python 3.6 poetry 1.1.15 will be installed. We are hoping to remove this soonish as we look to remove python 3.6 support.

As to the issue with the `category` key being removed, the dilemma is that we can only really support 1 version of the lockfile currently. If we update our code to support the 1.5 lockfile format users still on < 1.5 will see updates that remove the `category` key and locally their tooling will add the key back in. That being said, we would like to move in the direction of supporting the latest package manager version more quickly, even if that means breaking backwards compatibility. With that in mind, if you could take a stab at moving the category detection to using `pyproject.toml` on this PR I think that would be the preferred direction from our side.

phillipuniverse:
@pavera thanks for the additional information! Specifically the Python 3.6 nuance, I had not considered how that worked but it makes perfect sense now how someone would get Poetry 1.1.15.

> With that in mind, if you could take a stab at moving the category detection to using pyproject.toml on this PR I think that would be the preferred direction from our side.

Definitely, I think that is the overall more stable approach. I can take this on.

One clarifying question - how does Dependabot generally support package managers that have many categories? Since we were relying on the 'category' behavior before this treated all group dependencies as "dev" dependencies. Do we want to keep that behavior or is there a better way to mark different group dependencies as different categories in the Dependabot language?

phillipuniverse:
Started in on a candidate solution and ran into a problem - transitive dependencies. Transitive dependencies are in the `poetry.lock` file but are not in the `pyproject.toml` file (by definition). I think in this case we'll need to traverse up the dependency tree until we find a dependency within a group in pyproject.toml,

jonathangreen:
@phillipuniverse I think the right solution is that you propose, to actually work out the category from the transitive dependencies. However maybe a first pass solution to this is to put all transitive dependencies into the `main` category, with a further todo to do a more sophisticated transitive dep resolution? 

According to the [comment here](https://github.com/python-poetry/poetry/pull/7637#issuecomment-1494272266) on the PR that removed the category key, defaulting to `main` is what poetry itself is doing.

allanlewis:
If you want to know which category a dependency is in, could `poetry export` help? I'm thinking you could export the prod and dev dependencies and see which are only dev.

phillipuniverse:
@jonathangreen thanks for additional info. I pushed up a relatively naive solution as you suggested, simply checking the existence of the dependency in `pyproject.toml`. If we didn't find it (meaning, transitive) I defaulted treating them as main dependencies, as indicated by the PR you linked.

As an aside, I think once the Python 3.6 (and potentially Python 3.7?) support is dropped in #7610, it's a good opportunity to look at the dependency category handling overall for Poetry support. There is still a lot of reliance on `dev-dependencies` (even in my latest changes) and the "group" support was bolted on. Since it's such an easy migration to go from `dev-dependencies` -> `group.dev.dependencies`, I think it would simplify some code paths to only rely on groups.

Remaining TODOs on this PR
- [ ] Fix the few failing tests
- [ ] Clean up the changes to all the `.lock` files that I rewrote for Poetry 1.5, I'll put this back where they were
- [ ] Make sure we have tests around a 1.5 lock file. I'm thinking I will modify the pyproject.toml added to test group support at https://github.com/dependabot/dependabot-core/blob/8919de6bed26d61cb55e4f05a92c0aca41972233/python/spec/dependabot/python/file_parser/pyproject_files_parser_spec.rb#L232-L245, generate a 1.5 lockfile from here and do verifications in the updator

The code as written is ready to review, I'm not planning on making many modifications. I'm pretty new to Ruby so I would appreciate readability suggestions!

bdragon (commented):
@phillipuniverse thanks for raising this!

Here is my understanding:
- if the dependency has a `category` key in the lockfile, it was generated with Poetry <v1.5 and we use its value to determine the dependency group
- if the dependency is in the main dependency group (i.e., `tool.poetry.dependencies`) in `pyproject.toml`, it stays that way
- if the dependency is in the [legacy] dev dependency group (i.e., `tool.poetry.dev-dependencies`) in `pyproject.toml`, it stays that way
- if the dependency is included in a dependency group other than the main one (e.g., `tool.poetry.group.test.dependencies`) in `pyproject.toml`, it goes in the dev dependencies group
- if the dependency is transitive, it goes in the main dependencies group

`tool.poetry.dev-dependencies` has been deprecated since Poetry v1.2 and `tool.poetry.group.dev.dependencies` is preferred, but this is a separate concern.

Based on the conversation here, the changes in https://github.com/python-poetry/poetry/pull/7637, and the Poetry docs—and given that we need to straddle Poetry versions 1.1 and 1.5—I think the solution proposed here looks pretty good.

I think we'd prefer to avoid traversing the dependency tree for every transitive dependency if possible, so defaulting to `dependencies` seems reasonable.

I went ahead and updated this branch with the latest main. It looks like you have one failing spec—if you don't mind getting that to pass, it'd be great to see this merged. Thanks!

jeffwidman:
> and given that we need to straddle Poetry versions 1.1 and 1.5

We may not need to do that once we drop support for Python 3.6/3.7... 
* #7702 
* https://github.com/dependabot/dependabot-core/pull/7716

I am planning to look at this once August 17th rolls around and I can merge my long string of Python PR's... Can we hold off on merging this until then?

There's a broader discussion we're starting to have internally about how old of versions we'll promise to support for package managers. So we'll also be trying to get some clarity there on how much we choose to support multiple versions vs telling users "we're not going to support older than N versions of `poetry` because it's not hard to update to the newer version of `poetry`, please do that instead"... again, a broader conversation that we're working through internally.

So let's just put this on pause til August 17th and then re-open the discussion.

G-Rath:
> So let's just put this on pause til August 17th and then re-open the discussion.

keeping in mind that without this dependabot just flat out falls over on new `poetry.lock` files - we've already had a few patch updates on our projects that've needed to be done manually due to this; is it possible get something out to fix the current hard error in parallel to your longer term discussions?

phillipuniverse:
@G-Rath @jeffwidman today, all automated security updates in GitHub fail for projects that use Poetry >=1.5 without this PR. This has been exacerbated by Dependabot being updated to use Poetry 1.5 when emitting normal, non-security-related updates.

Now that I have a bit more clarity on the final state, I can address the feedback that @bdragon suggested, we can fix the breakage in the GitHub/Dependabot ecosystem that exists today, then go for broader sweeping changes in tandem with the Python 3.8+ change.

@jeffwidman does that change your mind at all or still think we need to wait?

jeffwidman:
Thanks for the feedback. I didn't realize this was breaking everything for you folks, that sucks. 

Yeah, given that I'd also lean towards shipping this sooner, and then later cleaning up any cruft it introduces for supporting older versions of poetry. 

Let me take a deeper look at this on Monday before committing for sure though.

python/spec/dependabot/python/file_parser/pyproject_files_parser_spec.rb
```diff
@@ -136,7 +136,9 @@
           subject(:dep) { dependencies.find { |d| d.name == "atomicwrites" } }
 
           its(:subdependency_metadata) do
-            is_expected.to eq([{ production: false }])
+            # This is how Poetry treats transitive dev dependencies, see discussion at https://github.com/python-poetry/poetry/pull/7637#issuecomment-1494272266
+            # and https://github.com/dependabot/dependabot-core/pull/7418#issuecomment-1644012926
+            is_expected.to eq([{ production: true }])
```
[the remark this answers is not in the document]
phillipuniverse: Based on the latest reviw from @bdragon, this changed test is the expected behavior.

python/lib/dependabot/python/update_checker/poetry_version_resolver.rb
```diff
@@ -298,12 +298,32 @@ def update_dependency_requirement(toml_node, requirement)
         end
 
         def subdep_type
```
phillipuniverse, line 300: I just happened to find this function unexpectedly searching around for `subdep_type`. Looks like there is a decent amount of logic shared in the file updater, but probably small enough such that it made more sense to copy into here vs extract into a utility.

Since this is getting larger, maybe it makes sense to refactor to a utility? Another set of options:

1. Refactor into a utility function and call from both places. Opinion on where this should go?
2. Remove this modification since it's not specifically tackling the core issue and we are trying to be pretty targeted per other conversations
3. Keep as-is

phillipuniverse (commented):
Latest changes has tests around the 1.5 lockfile and should be compatible with changes requested from @bdragon

python/lib/dependabot/python/update_checker/poetry_version_resolver.rb
deivid-rodriguez: You forgot to pass `poetry_object` around, and a real update seems to be still failing because of that. So this needs to be fixed before merging, and also signals some lack of coverage here.

I'll look into a good place for factoring out this method.

deivid-rodriguez: I proposed https://github.com/dependabot/dependabot-core/pull/7418 to refactor this. We can rebase and adapt this PR to it, and should be all good.

deivid-rodriguez:
We just drop old Python support, and will now unconditionally run Poetry 1.5.

However, I guess we'll need to keep the compatibility code in case we find an old lockfile anyways? We may not generate an ideal diff but that's an existing problem and I'm not sure completely dropping support for lockfiles generated with old Poetry versions would be an improvement. So I'd lean towards keeping it?

deivid-rodriguez:
Ok, so in the end I decided to go with a different approach at #7834, which I find cleaner and I believe should work better.

@phillipuniverse I'm trying to add attribution to the commit but it just won't pick it up. Not sure what I'm doing wrong.

phillipuniverse:
> We may not generate an ideal diff but that's an existing problem and I'm not sure completely dropping support for lockfiles generated with old Poetry versions would be an improvement. So I'd lean towards keeping it?

Hm maybe... in the current situation, Dependabot will read the old version and then rewrite with 1.5+ anyway. Seems about the same as just erroring out with a message about only supporting Poetry 1.5+ 🤷 

> Ok, so in the end I decided to go with a different approach at https://github.com/dependabot/dependabot-core/pull/7834, which I find cleaner and I believe should work better.

Definitely agree. I looked it over and this looks to be a more sustainable solution than what I had here. I really like the `poetry show --only main` part of it, very simple solution for discovering dev vs non-dev.

> I'm trying to add attribution to the commit but it just won't pick it up. Not sure what I'm doing wrong.

@deivid-rodriguez I appreciate the attempt at attribution but honestly, nbd on my side. I'm just excited for my automated security updates to start working again!!

Thanks for the care and attention! I'm going to go ahead and close this in favor of #7834.

deivid-rodriguez:
> Hm maybe... in the current situation, Dependabot will read the old version and then rewrite with 1.5+ anyway. Seems about the same as just erroring out with a message about only supporting Poetry 1.5+ 🤷

That's a good point, not necessarily better to keep opening PRs changing the lockfile format to 1.5+. However, that's an existing issue so I prefer to keep any fix focused on adding support 1.5+ lockfile format but keep the current behavior other than that. I also have the feeling that many users are aware of this issue but may be using Dependabot PRs as a way to upgrade their Poetry version to stop the format from changing back and forth.

> Definitely agree. I looked it over and this looks to be a more sustainable solution than what I had here. I really like the poetry show --only main part of it, very simple solution for discovering dev vs non-dev.

Thanks for the kind words!

> @deivid-rodriguez I appreciate the attempt at attribution but honestly, nbd on my side. I'm just excited for my automated security updates to start working again!!

That's great, I'm still confused since it's the first time adding `Co-authored-by:` information to a commit message is not picked up by GitHub 🤷.

jeffwidman:
👋 I just want to say thanks again to everyone who chimed in on this issue providing additional context. I spent my afternoon reading poetry docs and learning it, and also understanding the issue here, and it was super helpful to have all the pointers everyone provided to historical context.

I'm deploying #7834 shortly, and as long as that looks green I'll merge it to `main`. So your `poetry` :dependabot: PR's are hopefully fixed by the time you read this.

phillipuniverse:
@deivid-rodriguez @jeffwidman - just wanted to report that thanks to you guys, all of my automated security updates have started working again!! Thank you so much!!

jeffwidman:
Awesome thanks for circling back to confirm!

esev:
Same here. Thank you!

Though I noticed one odd indirect dependency update where a package was downgraded. I hadn't seen that before, so just checking, is that expected?
https://github.com/pywemo/pywemo/pull/538/commits/d2d03e894809c79cb76bfc854e546a703d401e18 (rstcheck: 6.1.2 -> 3.5.0)

PR #7430, opened 2023-06-14
Ignore tags not matching prefix, when workflow is pinned to SHAs

deivid-rodriguez:
When updating explicit tags that include version numbers in their name in GitHub Actions workflows, we only update to tags that respect the existing format.

For example, if a workflow is pinned to the v2.3.6 tag through `github/codeql-action@v2.3.6` and the action's author tags `codeql-bundle-v2.13.4`, then Dependabot won't create updates.

However, when the workflow is pinned to the specific commit for that tag, for example like this `github/codeql-action@83f0fe6c4988d98a455712a27f0255212bba9bd4 # v2.3.6`, then Dependabot will still create an update to `github/codeql-action@cdcdbb579706841c47f7063dda365e292e5cad7a # v2.13.4`.

This PR fixes the problem by ensuring the format is respected, not only for explicit tags, but also for the tags corresponding to explicit SHA pins.


deivid-rodriguez:
Oops, I broke common CI apparently. Back to WIP.

deivid-rodriguez:
I think it should be fixed now but I'd like to confirm that the smoke test failure is not related to this PR.

jurre (approved):
That failure looks unrelated to me, not sure how it could be caused by these changes

PR #7433, opened 2023-06-14
[Grouped Updates] The VendorUpdater class watermarks DependencyFile objects it creates

brrygrdn:
Groundwork to fix https://github.com/dependabot/dependabot-core/issues/7319

Grouped Updates need to manage the aggregation of `DependencyFile` objects return from updaters in two ways:
- They need the maintain an overall list of every file that changed during each step to create the PR
- The need to maintain a list of the **manifest files** to pass into subsequent updaters so the changes are performed iteratively

In the case of vendor files, we need them to be present in the overall file list for (1), but we **do not** want them to be passed into subsequent updaters as their interface does not expect to have to filter them out and retrofitting this filtering across all our ecosystems isn't the most practical approach.

### The Changes

This PR does the following
- Adds a `vendored_file` attribute to `Dependabot::DependencyFile` which is false by default
- Splits out a base class from the `VendorUpdater` class to provide support for managing arbitrary files within an Updater that don't fall within the remit of the `FileUpdater` itself, i.e. they are not a manifest file, nor are they a vendor file
- Use this new base class to manage `.pnp.*` and install state files for yarn berry within our NPM library
- Updater the concrete `VendorUpdater` to set `vendored_file: true` on any `DependecyFile` objects it creates

### Rationale

This change introduces the `vendor_file` watermarking without changing any behaviour or modifying the Group Update processes to use it to isolate and de-risk this from the larger effort to handle grouped updates + vendoring.

### DIFF
--- a/common/lib/dependabot/file_updaters/vendor_updater.rb
+++ b/common/lib/dependabot/file_updaters/vendor_updater.rb
@@ -22,6 +22,12 @@ module Dependabot
       end
 
       alias updated_vendor_cache_files updated_files
+
+      private
+
+      def create_dependency_file(parameters)
+        Dependabot::DependencyFile.new(**parameters.merge({ vendored_file: true }))
+      end
     end
   end
 end
--- a/common/lib/dependabot/file_updaters/vendor_updater.rb
+++ b/common/lib/dependabot/file_updaters/vendor_updater.rb
@@ -1,20 +1,20 @@
 # frozen_string_literal: true
 
 require "dependabot/dependency_file"
-require "dependabot/file_updaters/supplement_updater"
+require "dependabot/file_updaters/artifact_updater"
 
-# This class is a specialisation of the SupplementUpdater which should be used
-# for vendored files so any files it creates are properly flagged as relating
-# to vendoring of dependencies.
+# This class is a specialisation of the ArtifactUpdater which should be used
+# for vendored files so any DependencyFile objects it creates are properly
+# flagged.
 #
-# This flagging ensures that the Updater and other components while handle them
-# correctly when compiling grouped updates.
+# This flagging ensures that the Updater will handle them correctly when
+# compiling grouped updates.
 module Dependabot
   module FileUpdaters
-    class VendorUpdater < SupplementUpdater
+    class VendorUpdater < ArtifactUpdater
       # This provides backwards compatability for anyone who used this class
-      # before the base class was introduced and aligns the method's public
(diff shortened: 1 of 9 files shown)


brrygrdn:
👋🏻 @pavera I just wanted to check that my changes around the npm file updater made sense for the yarn files being managed

common/lib/dependabot/file_updaters/vendor_updater.rb
```diff
@@ -1,78 +1,32 @@
...
+# This class is a specialisation of the ArtifactUpdater which should be used
+# for vendored files so any DependencyFile objects it creates are properly
+# flagged.
+#
+# This flagging ensures that the Updater will handle them correctly when
+# compiling grouped updates.
 module Dependabot
   module FileUpdaters
-    class VendorUpdater
+    class VendorUpdater < ArtifactUpdater
```
brrygrdn, line 14: I considered an alternative approach of adding a new argument to `VendorUpdater` rather than creating a 'generic' `ArtifactUpdater`, something like 

```
def initialise(repo_contents_path:, vendor_dir:, create_vendored_files: true)
```

Although this would work had result in less churn to the API, I felt this didn't make the intent that vendor files need some special handling clear enough and it would extremely easy for a subtle bug if we were to develop a new use for this utility class similar to the calls in the NPM `FileUpdater`.

The split of 'utility class to grab any artefact outside the manifest files' and 'specific class to manage vendored dependencies' seemed like a safer, more intentional interface.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater.rb
```diff
@@ -63,11 +64,11 @@ def updated_dependency_files
 
       def vendor_updated_files(updated_files)
         base_dir = updated_files.first.directory
-        pnp_updater.updated_vendor_cache_files(base_directory: base_dir).each do |file|
-          updated_files << file if file.name == ".pnp.cjs" || file.name == ".pnp.data.json"
+        pnp_updater.updated_files(base_directory: base_dir, only_paths: [".pnp.cjs", ".pnp.data.json"]).each do |file|
+          updated_files << file
```
brrygrdn, line 68: This was an opportunistic change to allow us to optimise very slightly to avoid reading any files we don't plan on using I've made in passing.

I'm ok with backing this out as it _is_ scope creep, but it was a fairly quick chance to make in passing.

bdragon (approved):
Looks great!

Issue #5391, opened 2022-07-18
[pub] do a smallest update heuristic for security updates


jonasfj:
When there is a security update to a package dependabot should be able to make a smallest possible update that gets the security fix.

For `pub` we'll probably need to apply a heuristic, @sigurdm we should continue the discussion on this.

jeffwidman:
@jonasfj @sigurdm 👋 I'm planning to wire up security updates for Pub in the next few weeks.

Do you have interest/time to create a PR adding support for the smallest update / minimal version required to bump to a non-vulnerable version?

My focus will primarily be on wiring up the plumbing so that security alerts can trigger Dependabot PR's, so it's doubtful I'll have time to add a new version strategy (esp as I'm not super familiar with the pub ecosystem).

If you can't get to it, no problem, I can just default to using the latest version strategy for now. But thought I'd check, as it'd be more efficient to wire it all up correctly from the get go rather than have to circle back to update it later.

jonasfj:
Hmm, I'll chat with sigurdm@ when he is back tomorrow.

I think we had a plan to do sometime. But there is lots of balls in the air -- as always. And its been some time.
I don't recall if we figured out how we were supposed to integrate/expose a new upgrade strategy.

@sigurdm as I recall we were thinking of increasing the lower bound constraint in-memory and then doing a downgrade solve.. or something like that. Because we want a newer version than the advisory affects, and we don't want older versions of anything we have. But otherwise we're not interested in upgrading anything beyond that -- hence, a downgrade solve with those lower bound constraints.

sigurdm:
> @sigurdm as I recall we were thinking of increasing the lower bound constraint in-memory and then doing a downgrade solve..

Yeah I suggested that - but after we discussed a bit I think we found out that won't work if there are more affected versions coming after the current.

I think we really need a way for dependabot to communicate all the affected versions to pub dependency_services such that we can forbid all of them in a solve.

jeffwidman:
That should be simple enough... at the point that dependency_services are called, the full array of `security_advisories` is available... here's a quick example of what's available using a fake advisory I injected into the `dry-run` script:
```shell
SECURITY_ADVISORIES='[{"dependency-name":"retry","patched-versions":[],"unaffected-versions":[],"affected-versions":["<= 3.0.2"]}]' bin/dry-run.rb pub test_org/test_repo --cache=files --dep=retry
```

results in:
```ruby
(ruby) security_advisories
[...]

PR #7446, opened 2023-06-16
Pub smallest update

sigurdm:
(two first commits are a42800e5a2f539dd5d86fdc3a6f3beefc971c753)

This fixes: https://github.com/dependabot/dependabot-core/issues/5391


sigurdm:
cc @jonasfj

pub/lib/dependabot/pub/update_checker.rb
```diff
@@ -42,26 +42,32 @@ def lowest_resolvable_security_fix_version
...
+        # If the current version is not vulnerable, we stay on it.
+        return version_unless_ignored dependency.version unless vulnerable?
+
+        e = dependency_services_smallest_update
+        return nil if e.nil?
+
+        upgrade = e.find { |u| u["name"] == dependency.name }
+
+        version = upgrade["version"]
+        # TODO: Should we do this upgrade even if the found version is ignored?
```
jurre, line 56: Ignore conditions also apply to security updates, as per the [docs here](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore):

> Setting this option will also affect pull requests for security updates to the manifest files of this package manager, unless you use `target-branch` to check for version updates on a non-default branch.

(The `target-branch` part is handled outside of core, we only pass in ignore conditions relevant to the branch that's getting updated)

For reference you can also see that we do this for npm [here](https://github.com/dependabot/dependabot-core/blob/main/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/latest_version_finder.rb#L71C1-L72)

sigurdm: Thanks for the info. Will remove the TODO :)

jonasfj (commented):
LGTM as far as I can see.

jurre:
Thanks! With this change, does [this](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#pub) exception for pub still apply? 🤔

sigurdm:
> Thanks! With this change, does [this](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#pub) exception for pub still apply? 🤔

I believe it still applies. This PR only touches security updates, not handling of ignored versions.

jurre (approved):

pub/lib/dependabot/pub/update_checker.rb
```diff
@@ -42,26 +42,31 @@ def lowest_resolvable_security_fix_version
...
       end
 
       def updated_requirements
         # Requirements that need to be changed, if obtain:
-        # latest_resolvable_version
-        entry = current_report["singleBreaking"].find { |d| d["name"] == dependency.name }
+        # latest_resolvable_version or lowest_security_fix_version
+        entry = if vulnerable?
+                  updates = dependency_services_smallest_update
+                  raise "Cannot upgrade from vulnerability without unlock." if update.size > 1
```
deivid-rodriguez, line 64: Tried this against a repo and run into an error due to a typo here:

```suggestion
                  raise "Cannot upgrade from vulnerability without unlock." if updates.size > 1
```

I guess we should get this covered, maybe through a smoke test at https://github.com/dependabot/smoke-tests.

sigurdm: Thanks for catching this!

Will you be able to set up a smoke test?

deivid-rodriguez: Not right now, but I'll try set that up!

deivid-rodriguez:
I retried the update after the typo fix and (in retrospective, as expected) I got the error:

> Cannot upgrade from vulnerability without unlock.

This was unexpected to me since the "highest version fix" worked fine. One potential cause is that the update I'm trying is from https://pub.dev/packages/http/versions/0.13.2 to https://pub.dev/packages/http/versions/0.13.3. That actually _removes_ one subdependency, so maybe that's showing as two updates, but it's a case that should be fine?

sigurdm:
> That actually removes one subdependency, so maybe that's showing as two updates, but it's a case that should be fine?

Yeah, that is a good point.

Should we make it that we only complain when the updates contains more than one change to the **manifest** (as opposed to changes in transitive dependencies that affect only the lockfile)? Or should we count the number of non-deleting changes?

I guess the reason we put a limit, is that security updates should be somewhat minimal...

deivid-rodriguez:
We try minimal security updates in terms of version moves, that's why we implement "lowest fixed version" logic, but it's more important that the security update can be fixed, even if it involves unlocking other dependencies. For example, I believe for NPM we have some logic to unlock parent dependencies if that's necessary to get out of a security vulnerability in a transitive dependency.

With that in mind, I'd go with whatever solution allows us to fix more security alerts.

sigurdm:
Ah, now I remember why I inserted the check - the api of `updated_requirements` only allows us to return a single update, as opposed to `updated_dependencies_after_full_unlock` that returns a list of upgraded requirements.

I'll make the check "smarter" to only count breaking changes...

deivid-rodriguez:
By breaking changes, you mean major bumps in transitive dependencies? If that's necessary to fix a vulnerability, I think it's not a problem.

I believe it's always best to show users a path forward (even if the proposed PR needs more work in terms of breaking changes) than just saying "this alert cannot be fixed"?

sigurdm:
> By breaking changes, you mean major bumps in transitive dependencies? If that's necessary to fix a vulnerability, I think it's not a problem.

No, I meant bumps that require an unlock in the manifest - sorry that was unclear.

deivid-rodriguez:
Interesting. So Pub does have the ability to unlock other direct dependencies to get a security upgrade done. We should eventually try to make that work too then, but this seems like a good start! Will give the updated PR a try, thanks for your work!

deivid-rodriguez:
Sorry for dropping the ball (again). I'll try this again and hopefully ship it tomorrow.

sigurdm:
NP - just didn't want this to fall between the cracks :)

deivid-rodriguez:
Hei @sigurdm!

I tried again and I'm still getting the error "Cannot upgrade from vulnerability without unlocking other packages." on an alert where we can now generate a proper PR. I'll have a closer look when I can.

It's just a `pubspec.lock` that has `http` locked to "0.13.2" (vulnerable to https://github.com/advisories/GHSA-4rgh-jx4f-qfcq). Right now we create an update to "0.13.5", the expected outcome of this PR is to generate an update to "0.13.3".

sigurdm:
Hmm, not sure what goes wrong.

I created a pubspec.lock locked to http 0.13.2, and ran:

```
SECURITY_ADVISORIES='[{"dependency-name":"http","affected-versions":["0.13.2"]}]' bin/dry-run.rb pub sigurdm/private_repo2 --security-updates-only
....
 => bump http from 0.13.2 to 0.13.3
....
```

Is there a way I can replicate what you see?

deivid-rodriguez:
I guess the repo I was using is not as simple as I said! Try with https://github.com/dsp-testing/security-updates-for-pub.

PR #7453, opened 2023-06-19
Remove pnpm experiment flag

mctofu:
Pnpm is now available to everyone so we can remove the experiment flag.


abdulapopoola (approved):
Thanks!

PR #7455, opened 2023-06-20
Remove duplicated ENV

deivid-rodriguez:
It's already set on the base image:

https://github.com/dependabot/dependabot-core/blob/e92409d72d55c436c4a66f75b404a3924e62dd84/Dockerfile.updater-core#L46


landongrindheim (approved):
✂️

Issue #7457, opened 2023-06-20
Dockerfile.development default FROM_IMAGE is not correct


dwc0011:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

.devcontainer/core-dev/devcontainer.json

### What you expected to see, versus what you actually saw

Creating devcontainer fails on pulling the Dockerfile.development default ARG FROM_IMAGE=dependabot/dependabot-core, as it does not exist in docker.io but in ghcr.io.

Update to ARG FROM_IMAGE=ghcr.io/dependabot/dependabot-core

PR incoming.

PR #7458, opened 2023-06-20
Update the image repo to ghcr.io

dwc0011:
Fixes https://github.com/dependabot/dependabot-core/issues/7457


deivid-rodriguez (commented):
🤔 But that's an old image prior to the per-ecosystem image refactoring I think? That development image, even if it builds, it's unlikely to be useful.

I think it'd make more sense to remove the default and make the `FROM_IMAGE` build arg required?

dwc0011:
> 🤔 But that's an old image prior to the per-ecosystem image refactoring I think? That development image, even if it builds, it's unlikely to be useful.
> 
> I think it'd make more sense to remove the default and make the `FROM_IMAGE` build arg required?

It does work as it is just bringing in the environment necessary to run Dependabot, i.e. Ruby and a few other tools.  That said, if it is no longer maintained, then yes, I think it should be a ARG without a default value or possibly to the updater-core image.

I will have to familiarize myself with this new per ecosystem image and how it is intended to be run with a script, as that is how the devcontainer launches and debugs via vscode.  It may be that the updater core image is a better fit for this, but that is something I will need to explore.

In the meantime, I am actually addressing another issue and just noticed this and a few other issues with the devcontainer that I wanted to address.

Dockerfile.development
```diff
@@ -1,4 +1,4 @@
-ARG FROM_IMAGE=dependabot/dependabot-core
+ARG FROM_IMAGE=ghcr.io/dependabot/dependabot-core
```
jeffwidman, line 1: Yeah, as documented in https://github.com/dependabot/dependabot-core#building-images-from-scratch the new base image at the bottom of our docker stack is now `dependabot-updater-core`: 
https://github.com/dependabot/dependabot-core/blob/4c9072d8a5575c7b0d7cd29c32c6d4658687f9cc/Dockerfile.updater-core

```suggestion
ARG FROM_IMAGE=ghcr.io/dependabot/dependabot-updater-core
```

jeffwidman (changes requested):
Great catch. We aren't publishing that image anymore, but most all the same stuff is now in the base updater image

dwc0011:
> Makes sense, let us know what you find. I guess defaulting to https://github.com/dependabot/dependabot-core/pkgs/container/dependabot-updater-core would be another option if that works.

I tested the updater core image last week and had no issues, updated the change requested to that just now :)

dwc0011:
As a result of testing, I noticed that rubocp and solargraph were not working correctly, added the proper configuration to get them evaluating properly.

dwc0011:
Actually, there is an issue.  There is no terraform in this image.

dwc0011:
Working on converting to a FROM_ARG

dwc0011:
@jeffwidman  @deivid-rodriguez deivid-rodriguez 
Is there a list of to ecosystem app version to package tags?  

For example: I want to use my FROM_IMAGE to be an image that has terraform version 1.4.5 or npm version 5.0.1 etc etc
What is the best way to determine this?

I want to create instructions on how to use the ARG but also want to add instructions on how to find the image tag they want to reference as well.  If there isn't a version available that matches their need then reference them to the developer build section.

Thanks

jeffwidman:
There's some bigger questions on this PR, and they're good ones. But in the meantime, we're still referencing an old/outdated docker image on `main`, so I extracted a subset of these commits into:
* https://github.com/dependabot/dependabot-core/pull/7565

This way having the broader discussion doesn't block removing old references from `main`.

dwc0011:
I added comments too late to the other PR but since we are moving forward on this one.  I need some guidance on preferred implementation.

Options:
1.  As this current PR does, add a ARG that accepts an ecosystem image and tag and make devs figure out which one supports a version of the helpers they are addressing.
2.  Use updater-core as base image and then update Dockerfile.development to install all helper tools
3.  Create a new updater-core-dev image that uses the updater-core image and installs all the helper tools and users can derive from it.

Once a preference is made, if 1 or 2 I can address, if 3 then that will be on the dependabot team.
If 1. is selected I need to know if there is a manifest for tool versions or is it a hunt through the commits to find what version it supports.  I think this would be good info to know for anybody hosting their own dependabot as many people are not running the latest version of helper tools.

PR #7459, opened 2023-06-20
Instantiate less dependencies

deivid-rodriguez:
While working on #7344 I was experimenting with adding some metadata to dependencies during file parsing that I would then use in subsequent steps. But I noticed that metadata is not being passed around.

I _think_ this is a bug introduced by 3a0f9950396b46421eb5009836f87809d6f88388, but I'm opening this PR and running CI to make sure I'm not missing anything.


bdragon (approved):
👍

PR #7462, opened 2023-06-21
Fix actions updates when inconsistent casing is used

deivid-rodriguez:
Fixes a small GitHub Actions edge case where we'll get a `Multiple sources!` error instead of proper updates.

### DIFF
--- a/github_actions/spec/dependabot/github_actions/file_parser_spec.rb
+++ b/github_actions/spec/dependabot/github_actions/file_parser_spec.rb
@@ -296,6 +296,53 @@ RSpec.describe Dependabot::GithubActions::FileParser do
       end
     end
 
+    context "with actions using inconsistent case" do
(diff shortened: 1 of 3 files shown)


github_actions/spec/dependabot/github_actions/file_parser_spec.rb
```diff
@@ -428,5 +475,6 @@ def mock_service_pack_request(nwo)
         end
       end
     end
+
```
landongrindheim, line 478: 💅 Looks like this stray newline found its way into the diff. Feel free to ignore 🙂 

```suggestion
```

landongrindheim (approved):
🚀

github_actions/spec/dependabot/github_actions/file_parser_spec.rb
deivid-rodriguez: Nice catch, it sounds like RuboCop also complained. Fixing!

PR #7464, opened 2023-06-21
Refactor dependency source management

deivid-rodriguez:
Normally the Swift package manager dependencies come from git repositories, so I'm planning to heavily use `GitCommitChecker` for checking for Swift updates.

However, `GitCommitChecker` is only useful for checking updates for direct dependencies, since it looks at `dependency.requirements` to find the git repo source, and only direct dependencies have explicit requirements.

I'm looking into refactoring this class so that it accepts a custom git source and can be potentially used for checking updates for indirect dependencies too.


deivid-rodriguez:
I expanded this PR to add support in `GitCommitChecker` for checking subdependencies. We did not need so far because in other package managers, git dependencies can only be defined at the top level, otherwise they are registry dependencies. But for Swift, most times subdependencies are also git dependencies.

`GitCommitChecker` now looks for a git source, not only in `Dependency#requirements` (which are specific to top level dependencies), but also in `Dependency#subdependency_metadata`.

cargo/lib/dependabot/cargo/update_checker.rb
```diff
@@ -279,12 +274,7 @@ def git_subdependency?
       end
 
       def path_dependency?
-        sources = dependency.requirements.
-                  map { |r| r.fetch(:source) }.uniq.compact
-
-        raise "Multiple sources! #{sources.join(', ')}" if sources.count > 1
-
-        sources.first&.fetch(:type) == "path"
+        dependency.source_details&.fetch(:type) == "path"
```
jurre, line 277:
```suggestion
        dependency.source_type == "path"
```

jurre (approved):
That makes sense to me!

jakecoffman (approved):

PR #7467, opened 2023-06-22
Configure git with ENV

deivid-rodriguez:
This makes tests have less side effects.

For example, if you run tests locally (outside of the development image), you'll end up with your git user set as `dependabot-ci`. That actually leaked into a PR we merged recently:

Captura de pantalla 2023-06-22 a las 16 05 33


deivid-rodriguez:
This seems to have worked, so I set it as ready!

jurre (approved):

PR #7469, opened 2023-06-22
[Grouped Updates] Utilise a git-backed workspace for each group so we retain vendor file changes properly

brrygrdn:
Merges into https://github.com/dependabot/dependabot-core/pull/7404

We've updated the group update strategy to avoid polluting the manifest files passed into each instance of Dependabot Core objects with changed vendored files, but the problem with this is that we also revert the project-on-disk to the state it was in before the previous update **and only copy the manifest back in**.

This means it is up to the native package manager to re-download the vendor files that changed in the previous step as well as the current step, which over time results in us replacing the same files over and over. Whether this is cached or we end up fully re-downloading binaries can end up being an implementation detail of the tools we shell out to, so we can't really rely on this long-term.

This branch solves the problem by using a workspace library to allow the Updater to orchestrate commits to the files on disk so each subsequent step has the previous step's changes persisted. We will still copy in the manifests as we need to modify their content before running native tooling to make the subsequent update, but this avoids risk of double-work.

### DIFF
--- a/updater/lib/dependabot/updater/group_update_creation.rb
+++ b/updater/lib/dependabot/updater/group_update_creation.rb
@@ -2,6 +2,7 @@
 
 require "dependabot/dependency_change_builder"
 require "dependabot/updater/dependency_group_change_batch"
+require "dependabot/workspace"
 
 # This module contains the methods required to build a DependencyChange for
 # a single DependencyGroup.
@@ -18,6 +19,8 @@ module Dependabot
       # outcome of attempting to update every dependency iteratively which
       # can be used for PR creation.
       def compile_all_dependency_changes_for(group)
+        prepare_workspace
+
         group_changes = Dependabot::Updater::DependencyGroupChangeBatch.new(
           initial_dependency_files: dependency_snapshot.dependency_files
         )
@@ -50,6 +53,7 @@ module Dependabot
 
           # Store the updated files for the next loop
           group_changes.merge(dependency_change)
+          store_changes(dependency)
         end
 
         # Create a single Dependabot::DependencyChange that aggregates everything we've updated
@@ -60,6 +64,8 @@ module Dependabot
           updated_dependency_files: group_changes.updated_dependency_files,
           dependency_group: group
         )
+      ensure
+        cleanup_workspace
       end
 
       def dependency_file_parser(dependency_files)
@@ -231,6 +237,32 @@ module Dependabot
             update_checker_for(original_peer_dep, dependency_files, raise_on_ignored: false).
               can_update?(requirements_to_unlock: :own)
           end
+      end
+
+      def prepare_workspace
+        return unless job.clone? && job.repo_contents_path
+
+        # TODO: Remove the directory parameter
+        #
+        # We should defer calculation of the `path` to the call site in the shared_helper
+        # so it is impossible to calculate different values for workspace and non-workspace
+        # calls to the helper.
+        Dependabot::Workspace.setup(
+          repo_contents_path: job.repo_contents_path,
+          directory: Pathname.new(job.source.directory || "/").cleanpath
+        )
+      end
+
+      def store_changes(dependency)
+        return unless job.clone? && job.repo_contents_path
+
+        Dependabot::Workspace.store_change(memo: "Updating #{dependency.name}")
+      end
+
+      def cleanup_workspace
+        return unless job.clone? && job.repo_contents_path
+
+        Dependabot::Workspace.cleanup!
       end
     end
   end
--- a/updater/lib/dependabot/updater/group_update_creation.rb
+++ b/updater/lib/dependabot/updater/group_update_creation.rb
@@ -242,11 +242,6 @@ module Dependabot
       def prepare_workspace
         return unless job.clone? && job.repo_contents_path
 
-        # TODO: Remove the directory parameter
-        #
-        # We should defer calculation of the `path` to the call site in the shared_helper
-        # so it is impossible to calculate different values for workspace and non-workspace
-        # calls to the helper.
         Dependabot::Workspace.setup(
           repo_contents_path: job.repo_contents_path,
           directory: Pathname.new(job.source.directory || "/").cleanpath
--- a/common/lib/dependabot/shared_helpers.rb
+++ b/common/lib/dependabot/shared_helpers.rb
@@ -24,21 +24,19 @@ module Dependabot
                  "(+https://github.com/dependabot/dependabot-core)"
     SIGKILL = 9
 
-    def self.in_a_temporary_repo_directory(directory = "/",
-                                           repo_contents_path = nil,
-                                           &block)
+    def self.in_a_temporary_repo_directory(directory = "/", repo_contents_path = nil, &block)
       if repo_contents_path
-        path = Pathname.new(File.join(repo_contents_path, directory)).expand_path
-        reset_git_repo(repo_contents_path)
-        # Handle missing directories by creating an empty one and relying on the
-        # file fetcher to raise a DependencyFileNotFound error
-        FileUtils.mkdir_p(path)
+        if Dependabot::Workspace.active_workspace
+          Dependabot::Workspace.active_workspace.change(&block)
+        else
+          path = Pathname.new(File.join(repo_contents_path, directory)).expand_path
+          reset_git_repo(repo_contents_path)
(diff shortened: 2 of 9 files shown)


brrygrdn:
I expect common lib to fail on this for now as I've made some breaking changes to the API for the workspace I want to socialise for review before fixing tests /cc @bdragon

brrygrdn:
This branch can be checked using this smoke test: https://github.com/dependabot/smoke-tests/pull/84

updater/lib/dependabot/updater/group_update_creation.rb
```diff
@@ -232,6 +238,32 @@ def peer_dependency_should_update_instead?(dependency_name, dependency_files, up
...
       end
+
+      def prepare_workspace
+        return unless job.clone? && job.repo_contents_path
+
+        # TODO: Remove the directory parameter
+        #
+        # We should defer calculation of the `path` to the call site in the shared_helper
+        # so it is impossible to calculate different values for workspace and non-workspace
+        # calls to the helper.
```
bdragon, line 249: Good callout 👍

bdragon (approved):
This looks good! Being able to stage changes incrementally before committing makes for a more flexible API

updater/lib/dependabot/updater/group_update_creation.rb
brrygrdn: This might be a bit of a 🐔 and 🥚 having looked at it more as we rely on the path to set the git configuration within the repo folder - setting them globally seems like bad form and I'd prefer not to - but it's mostly a side effect of us passing the `repo_contents_path` and `directory` right down the stack to the native helper.

They should always be the same in both call-sites as they do not change for the entire context of a run[^1], so I think while it feels odd as they could in principal be out of sync that shouldn't ever happen without a bug being introduced.

[^1]: Unless we support cross-folder grouping jobs in future 🤔

brrygrdn: I didn't say the important part in my last comment - I'm going to back off changing this for now

updater/lib/dependabot/updater/group_update_creation.rb
```diff
@@ -18,6 +19,8 @@ module GroupUpdateCreation
       # outcome of attempting to update every dependency iteratively which
       # can be used for PR creation.
       def compile_all_dependency_changes_for(group)
```
brrygrdn, line 21: We now get an overall debug spiel that looks like this for each step:

```
updater | 2023/06/23 13:42:33 DEBUG File deleted: '/bundler-vendored/vendor/cache/rack-2.1.4.gem'
updater | 2023/06/23 13:42:33 DEBUG File created: '/bundler-vendored/vendor/cache/rack-3.0.8.gem'
updater | 2023/06/23 13:42:33 DEBUG Dependencies updated:
updater | 2023/06/23 13:42:33 DEBUG  - rubocop ( 0.76.0 to 1.53.0 )
updater | 2023/06/23 13:42:33 DEBUG  - rack ( 2.1.4 to 3.0.8 )
updater | 2023/06/23 13:42:33 DEBUG Dependency files updated:
updater | 2023/06/23 13:42:33 DEBUG   - /bundler-vendored/Gemfile ( Changed 1 times )
updater | 2023/06/23 13:42:33 DEBUG   - /bundler-vendored/Gemfile.lock ( Changed 1 times )
updater | 2023/06/23 13:42:33 DEBUG Vendored dependency changes:
updater | 2023/06/23 13:42:33 DEBUG   - /bundler-vendored/vendor/cache/jaro_winkler-1.5.4.gem ( Changed 1 times )
updater | 2023/06/23 13:42:33 DEBUG   - /bundler-vendored/vendor/cache/parallel-1.22.1.gem ( Changed 1 times )
updater | 2023/06/23 13:42:33 DEBUG   - /bundler-vendored/vendor/cache/parser-3.1.2.0.gem ( Changed 1 times )
updater | 2023/06/23 13:42:33 DEBUG   - /bundler-vendored/vendor/cache/rubocop-0.76.0.gem ( Changed 1 times )
updater | 2023/06/23 13:42:33 DEBUG   - /bundler-vendored/vendor/cache/ruby-progressbar-1.11.0.gem ( Changed 1 times )
updater | 2023/06/23 13:42:33 DEBUG   - /bundler-vendored/vendor/cache/unicode-display_width-1.6.1.gem ( Changed 1 times )
updater | 2023/06/23 13:42:33 DEBUG   - /bundler-vendored/vendor/cache/json-2.6.3.gem ( Changed 1 times )
updater | 2023/06/23 13:42:33 DEBUG   - /bundler-vendored/vendor/cache/language_server-protocol-3.17.0.3.gem ( Changed 1 times )
updater | 2023/06/23 13:42:33 DEBUG   - /bundler-vendored/vendor/cache/parallel-1.23.0.gem ( Changed 1 times )
updater | 2023/06/23 13:42:33 DEBUG   - /bundler-vendored/vendor/cache/parser-3.2.2.3.gem ( Changed 1 times )
updater | 2023/06/23 13:42:33 DEBUG   - /bundler-vendored/vendor/cache/racc-1.7.1.gem ( Changed 1 times )
updater | 2023/06/23 13:42:33 DEBUG   - /bundler-vendored/vendor/cache/regexp_parser-2.8.1.gem ( Changed 1 times )
updater | 2023/06/23 13:42:33 DEBUG   - /bundler-vendored/vendor/cache/rexml-3.2.5.gem ( Changed 1 times )
updater | 2023/06/23 13:42:33 DEBUG   - /bundler-vendored/vendor/cache/rubocop-1.53.0.gem ( Changed 1 times )
updater | 2023/06/23 13:42:33 DEBUG   - /bundler-vendored/vendor/cache/rubocop-ast-1.29.0.gem ( Changed 1 times )
updater | 2023/06/23 13:42:33 DEBUG   - /bundler-vendored/vendor/cache/ruby-progressbar-1.13.0.gem ( Changed 1 times )
updater | 2023/06/23 13:42:33 DEBUG   - /bundler-vendored/vendor/cache/unicode-display_width-2.4.2.gem ( Changed 1 times )
updater | 2023/06/23 13:42:33 DEBUG   - /bundler-vendored/vendor/cache/rack-2.1.4.gem ( Changed 1 times )
updater | 2023/06/23 13:42:33 DEBUG   - /bundler-vendored/vendor/cache/rack-3.0.8.gem ( Changed 1 times )
updater | 2023/06/23 13:42:33 DEBUG Storing change to workspace: Updating rack
updater | 2023/06/23 13:42:33 DEBUG [workspace] store_change - before: 11b45e5 Updating rubocop
updater | 2023/06/23 13:42:33 DEBUG [workspace] store_change - after: 1118b5c Updating rack
```

brrygrdn: I've definitely messed up the `change_count` though, I'll look at that separately.

updater/lib/dependabot/updater/group_update_creation.rb
```diff
@@ -232,6 +238,27 @@ def peer_dependency_should_update_instead?(dependency_name, dependency_files, up
               can_update?(requirements_to_unlock: :own)
           end
       end
+
+      def prepare_workspace
+        return unless job.clone? && job.repo_contents_path
```
Nishnha, line 243: Can you explain why these checks are here? I looked at [job.clone?](https://github.com/dependabot/dependabot-core/blob/c9b38ced0d36e66339e92852030a1f9f9750b0ce/updater/lib/dependabot/job.rb#L122-L125) and it would be true if `vendor_dependencies?` is true. 

I think that we would want to use a workspace when `vendor_dependencies?` is true though?

bdragon: I think the intent of the checks is to make sure we only utilize the workspace in cases where we know a git repo is available. Since it's a `return unless`, I think it's doing what we want.

brrygrdn: We generally must clone as a precondition for `vendor_dependencies: true` to the best of my knowledge since we are manipulating content that is beyond the scope of the `FileFetcher` involved - there are some languages in which we always clone for this reason so the workspace will be invoked for those as well, but that's relatively useful for the extra debugging information even if we do extract/inject only manifest changes.

Over time we should look at expanding this so we always use the workspace for grouped updates so we have access to per-set diff - and the potential to experiment with extracting individual commits to open PRs with them instead of just the aggregate diff.

PR #7473, opened 2023-06-23
Remove simplecov

deivid-rodriguez:
There were a few leftovers but I'm pretty sure we're not actually using it.


jurre (approved):

jeffwidman:
Nice find!

Issue #6976, opened 2023-04-01
AWS Codecommit pull request description max size


joshua22s:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

bundler

### Language version

ruby 3.1

### Manifest location and content before the Dependabot update

`# frozen_string_literal: true

source "https://rubygems.org"

gem "irb"
gem "dependabot-omnibus", "~> 0.215.0"
gem "optimist", "~> 3.0.1"`

### What you expected to see, versus what you actually saw

I am configuring Dependabot with an Angular project in a AWS codecommit project. I am using ruby in my AWS Codebuild project for this. I am getting stuck on the creation of the pull request in the repository. The error I get says that the size of the pull request description is to large for AWS Codecommit. According to the AWS Codecommit documentation the max size of a pull request description is 10240 characters. The description created by Dependabot seems to exceed that.

jeffwidman:
Makes sense.

Support for CodeCommit is community-driven, we're thrilled folks are using Dependabot and happy to review PR's, but not something the core team will be tackling.

It'd probably be pretty straightforward to plumb in a max description length and truncate anything longer... In fact, I think we already do that for some PR creation clients. But CodeCommit support was contributed somewhat recently, and it probably just was never plumbed through for that client.

The CodeCommit client lives here: https://github.com/dependabot/dependabot-core/blob/fa8af4e545e57b2e6d181a91d1b49effc2a0d8bd/common/lib/dependabot/pull_request_creator/codecommit.rb

dwc0011:
Submitted PR #7487 to address this

PR #7487, opened 2023-06-26
Add truncate_pr_description and align rescue properly in fetch_file_c…

dwc0011:
…ontents method

This PR fixes issue #6976 by creating truncate pr method similar to that in azure.rb

In addition to satisfy the linter I aligned the rescue on line 73


yeikel (commented):
Please add tests

dwc0011:
@deivid-rodriguez @jeffwidman if you have chance to review I would appreciate it.  There is  major change for terraform providers that causes a PR description that exceeds the codecommit limit and is blowing up on us.  Thanks!

dwc0011:
@deivid-rodriguez @jeffwidman Any estimate on when you all will be able to look at this?

common/lib/dependabot/clients/codecommit.rb
```diff
@@ -193,6 +196,19 @@ def create_pull_request(pr_name, target_branch, source_branch,
...
       end
 
+      def truncate_pr_description(pr_description)
+        # Codecommit only supports descriptions up to 10240 chars in length
+        # https://docs.aws.amazon.com/codecommit/latest/APIReference/API_PullRequest.html
+        pr_description = pr_description.to_s
+        if pr_description.length > MAX_PR_DESCRIPTION_LENGTH
+          truncated_msg = "...\n\n_Description has been truncated_"
+          truncate_length = MAX_PR_DESCRIPTION_LENGTH - truncated_msg.length
+          pr_description = (pr_description[0..truncate_length] + truncated_msg)
```
jeffwidman, line 206: This whole section of var names + logic is really clean/readable, well done. 💯

jeffwidman (approved):
Looks great!

Nice thorough test cases too (🎩 💁‍♂️  to @yeikel for the nudge).

common/lib/dependabot/clients/codecommit.rb
jeffwidman: Oh, haha, this is copy/pasted from Azure... kudos to @jurre on the cleanness of his original implementation.

jeffwidman:
Hmm, I just noticed this _already_ exists in 2 other places:

https://github.com/dependabot/dependabot-core/blob/221a6e5599ea1e8c30f6167b0a1bbe48c0b8e15f/common/lib/dependabot/pull_request_creator/github.rb#L351-L362

https://github.com/dependabot/dependabot-core/blob/221a6e5599ea1e8c30f6167b0a1bbe48c0b8e15f/common/lib/dependabot/clients/azure.rb#L378-L389

And this PR will add a third... And all the other platforms (BitBucket, GitLab, etc) also have PR limits:
* https://jira.atlassian.com/browse/BSERV-14135
* https://docs.gitlab.com/ee/administration/instance_limits.html#size-of-comments-and-descriptions-of-issues-merge-requests-and-epics

This is going to become a mess quickly if each client has their own implementation.

So could we instead pull this truncation helper into the common PR description creator functionality? 

Azure may need to remain a special case since they calculate character limits based on the UTF-16 case, but the others can all be DRY'd up.

A quick peek looks like all the clients in https://github.com/dependabot/dependabot-core/blob/221a6e5599ea1e8c30f6167b0a1bbe48c0b8e15f/common/lib/dependabot/pull_request_creator.rb build their PR descriptions using `MessageBuilder.pr_message`: 

https://github.com/dependabot/dependabot-core/blob/221a6e5599ea1e8c30f6167b0a1bbe48c0b8e15f/common/lib/dependabot/pull_request_creator/message_builder.rb#L50-L56

So could you take this `truncate_message()` method and:
1. move it to the `MessageBuilder` class
2. give it a configurable param for the length at which to truncate. And maybe if `nil` then no truncation... 
3. Then within each client we'd pass in the param for the length for that platform... 

You can see a similar idea in this PR that wires up a common param for truncating the length of the PR branch name:
* https://github.com/dependabot/dependabot-core/pull/5338

The one bit I'm a little unclear on is how to set a default value, because there's no point in each caller of this code having to manually set the value... most everyone will simply want the default for their platform.

My guess is this call:
https://github.com/dependabot/dependabot-core/blob/221a6e5599ea1e8c30f6167b0a1bbe48c0b8e15f/common/lib/dependabot/pull_request_creator.rb#L210

Would look more like:
```ruby
pr_description: message.pr_message(CODECOMMIT.MAX_PR_DESCRIPTION_LENGTH),
```

Or you may need to hook it into the memo'ized MessageBuilder, I'm not sure:
https://github.com/dependabot/dependabot-core/blob/221a6e5599ea1e8c30f6167b0a1bbe48c0b8e15f/common/lib/dependabot/pull_request_creator.rb#L218-L232

Overall though this doesn't look like too much more work here... your existing set of tests should be fine, just move them over to the common code.

If you want to just wire it up for CodeCommit, then I can handle the cleanup of moving the GitHub truncation code to using this new common param.

common/lib/dependabot/clients/codecommit.rb
jeffwidman: And this also exists here too: https://github.com/dependabot/dependabot-core/blob/221a6e5599ea1e8c30f6167b0a1bbe48c0b8e15f/common/lib/dependabot/pull_request_creator/github.rb#L351-L362

oof, this will get out of hand quickly.

jeffwidman:
As illustrated by https://github.com/dependabot/dependabot-core/issues/7564#issuecomment-1636869709, some users will probably have valid reasons for truncating the PR description even shorter than their platform's limit. So ideally we both set the per-platform default and also make it a configurable param.

common/lib/dependabot/clients/codecommit.rb
jurre: Could extract a little helper? Although these are unlikely to ever change

dwc0011:
I will take a look at this later today or tomorrow.  Thanks!

dwc0011:
@jeffwidman I have moved the truncate and updated the calls. 

"my Ruby foo isn't strong enough to know the best way to do this"  - My Ruby fu was all gained contributing here to dependabot, I am learning as I go 😁

dwc0011:
@jeffwidman I also tested this using the dry run script against code commit to see if the default values were properly selected and then if overriding worked, as well as with passing encoding param.  After a couple of tweaks it did exactly what I expected it would.

Excited to get this one merged.  Hopefully this passes muster.

common/lib/dependabot/pull_request_creator/azure.rb
```diff
@@ -10,6 +10,11 @@ class Azure
                   :files, :commit_message, :pr_description, :pr_name,
                   :author_details, :labeler, :reviewers, :assignees, :work_item
 
+      MAX_PR_DESCRIPTION_LENGTH = 3999 # Azure has a max length of 4000 (0 based count)
+      ENCODING = Encoding::UTF_16 # Azure DevOps only support descriptions up to 4000 characters in UTF-16
+      # encoding.
+      # https://developercommunity.visualstudio.com/content/problem/608770/remove-4000-character-limit-on-pull-request-descri.html
```
jeffwidman, line 16:
```suggestion
      # Azure DevOps has a max description length of 4,000 characters in UTF-16 encoding:
      # https://developercommunity.visualstudio.com/content/problem/608770/remove-4000-character-limit-on-pull-request-descri.html
      PR_DESCRIPTION_MAX_LENGTH = 3_999 # zero-based indexing
      PR_DESCRIPTION_ENCODING = Encoding::UTF_16
```

The slight renames of the constants was intentional for improved clarity and consistency... for example encoding could be different for PR description vs other bits like titles, labels, etc...

common/lib/dependabot/pull_request_creator.rb
```diff
@@ -216,19 +220,32 @@ def codecommit_creator
...
-          pr_message_footer: pr_message_footer,
-          vulnerabilities_fixed: vulnerabilities_fixed,
-          github_redirection_service: github_redirection_service,
-          dependency_group: dependency_group
-        )
+      return @message unless @message.nil?
+
+      case source.provider
+      when "github"
+        @pr_message_max_length = Github::MAX_PR_DESCRIPTION_LENGTH if @pr_message_max_length.nil?
```
jeffwidman, line 227: Let's call all of these constants `PR_DESCRIPTION_MAX_LENGTH` as I think "pr description" is the parent in the taxonomy... it also matches better with `@pr_message_max_length`

common/lib/dependabot/pull_request_creator/github.rb
```diff
@@ -9,7 +9,11 @@ module Dependabot
   class PullRequestCreator
     # rubocop:disable Metrics/ClassLength
     class Github
-      MAX_PR_DESCRIPTION_LENGTH = 65_536 # characters (see #create_pull_request)
```
jeffwidman, line 12: I realize this is technically a slight breaking change to rename this to `PR_DESCRIPTION_MAX_LENGTH`, but I think that's fine for this... although technically part of the public API I highly doubt it's used by anyone outside the scope of this library.

common/lib/dependabot/pull_request_creator/github.rb
```diff
@@ -9,7 +9,11 @@ module Dependabot
...
+      MAX_PR_DESCRIPTION_LENGTH = 65_536 # Limit PR description to MAX_PR_DESCRIPTION_LENGTH (65,536) characters
+      # and truncate with message if over. The API limit is 262,144 bytes
+      # (https://github.community/t/maximum-length-for-the-comment-body-in-issues-and-pr/148867/2).
+      # As Ruby strings are UTF-8 encoded, this is a pessimistic limit: it
+      # presumes the case where all characters are 4 bytes.
```
jeffwidman, line 16:
```suggestion
      # GitHub limits PR descriptions to 262,144 bytes. As Ruby strings are UTF-8 encoded, 
      # pessimistically assuming that all characters are 4 bytes results in a max length of 65,536 characters.
        # https://github.community/t/maximum-length-for-the-comment-body-in-issues-and-pr/148867/2
      PR_DESCRIPTION_MAX_LENGTH = 65_536

File: common/lib/dependabot/pull_request_creator.rb [hunk 6]
```diff
@@ -88,6 +90,8 @@ def initialize(source:, base_commit:, dependencies:, files:, credentials:,
       @provider_metadata          = provider_metadata
       @message                    = message
       @dependency_group           = dependency_group
+      @pr_message_max_length      = pr_message_max_length
```
jeffwidman, line 93: Any reason why you picked `pr_message...` rather than `pr_description...`?

My thinking is there are multiple messages tied to a PR, for example comments, reviews, etc but only one main description tied to it... but I may be too GitHub centric as that's the main platform I work with (both professionally and my open source projects)

common/lib/dependabot/pull_request_creator/message_builder.rb
```diff
@@ -39,22 +42,45 @@ def initialize(source:, dependencies:, files:, credentials:,
...
+
+        msg = msg.dup
+        msg = msg.force_encoding(pr_message_encoding) unless pr_message_encoding.nil?
+
+        if msg.length > pr_message_max_length
+          tr_msg = pr_message_encoding.nil? ? TRUNCATED_MSG : (+TRUNCATED_MSG).dup.force_encoding(pr_message_encoding)
+          trunc_length = pr_message_max_length - tr_msg.length
+          msg = (msg[0..trunc_length] + tr_msg)
+        end
+        msg.force_encoding(Encoding::UTF_8) unless pr_message_encoding.nil?
```
jeffwidman, line 80:
```suggestion
        # if we used a custom encoding for calculating length, then we need to force back to UTF-8
        msg.force_encoding(Encoding::UTF_8) unless pr_message_encoding.nil?
```

common/lib/dependabot/clients/codecommit.rb
```diff
@@ -68,8 +68,8 @@ def fetch_file_contents(repo, commit, path)
           commit_specifier: commit,
           file_path: path
         ).file_content
-        rescue Aws::CodeCommit::Errors::FileDoesNotExistException
-          raise NotFound
+      rescue Aws::CodeCommit::Errors::FileDoesNotExistException
+        raise NotFound
```
jeffwidman, line 72: Should this get pulled out to a separate PR? I realize it's trivial but I'll be merging via squash so if for some reason we have to revert I'd rather not revert this change as it's unrelated to truncation AFAICT

jeffwidman (commented):
This looks really solid to me! Nice job. I appreciate you going the extra mile to make the encoding arg generic so we can use it beyond Azure.

A few small nits, feel free to push back on any of them if you think I'm missing something or just plain wrong. 😁 

What do you think about also setting default values for BitBucket and GitLab?
* https://docs.gitlab.com/ee/administration/instance_limits.html#size-of-comments-and-descriptions-of-issues-merge-requests-and-epics
* https://jira.atlassian.com/browse/BSERV-14135#:~:text=Currently%2C%20the%20max%20length%20of,property%20to%20increase%20this%20limit.

It'd be super simple to tack onto this PR since it's just setting a few more values / case statements. And I'm not too particular about our inability to test them because if they happen to be wrong other folks who are using :dependabot: on those platforms can always come along and correct it since this is OSS.

jeffwidman:
Just remembered that @mburumaxwell had noticed that `force_encoding` has a problem:
* https://github.com/dependabot/dependabot-core/pull/7140

My inclination is to first land this PR as-is, as it's semi-generic so will have a larger impact, then circle back and rebase/try to his PR after this is merged.

dwc0011:
> This looks really solid to me! Nice job. I appreciate you going the extra mile to make the encoding arg generic so we can use it beyond Azure.
> 
> A few small nits, feel free to push back on any of them if you think I'm missing something or just plain wrong. 😁
> 
> What do you think about also setting default values for BitBucket and GitLab?
> 
> * https://docs.gitlab.com/ee/administration/instance_limits.html#size-of-comments-and-descriptions-of-issues-merge-requests-and-epics
> * https://jira.atlassian.com/browse/BSERV-14135#:~:text=Currently%2C%20the%20max%20length%20of,property%20to%20increase%20this%20limit.
> 
> It'd be super simple to tack onto this PR since it's just setting a few more values / case statements. And I'm not too particular about our inability to test them because if they happen to be wrong other folks who are using :dependabot: on those platforms can always come along and correct it since this is OSS.

I was going to do this, but I actually use bitbucket and was able to create a PR with a description longer than that value in that ticket, so I was not sure what the real limit was. Since gitlab and bitbucket didnt have values before I just figured leave it as is and if someone comes across an issue then they would add the limit.

common/lib/dependabot/pull_request_creator.rb
dwc0011: I named it that as the method called is pr_message in MessageBuilder

common/lib/dependabot/pull_request_creator.rb
dwc0011: I had used the naming convention that was already in place, but based on your suggestion I have updated the names to your proposed convention.

common/lib/dependabot/pull_request_creator.rb
jeffwidman: Ah, gotcha, yeah, it probably wouldn't hurt to rename that to DescriptionBuilder (totally out of scope of this PR)... I'll look into that after this is merged...  let's leave this comment unresolved as a reminder to me.

common/lib/dependabot/clients/codecommit.rb
jeffwidman: I see this resolved, but I don't see a separate PR for it? Don't want to lose track of it if it's something that needs doing...

common/lib/dependabot/clients/azure.rb
```diff
@@ -22,7 +22,7 @@ class TagsCreationForbidden < StandardError; end
 
       RETRYABLE_ERRORS = [InternalServerError, BadGateway, ServiceNotAvailable].freeze
 
-      MAX_PR_DESCRIPTION_LENGTH = 3999
+      PR_DESCRIPTION_MAX_LENGTH = 3999
```
jeffwidman, line 25: I'm confused... why is this both here and also in `common/lib/dependabot/pull_request_creator/azure.rb`??

I'm guessing it should be removed from this file?

common/lib/dependabot/pull_request_creator/azure.rb
```diff
@@ -10,6 +10,11 @@ class Azure
                   :files, :commit_message, :pr_description, :pr_name,
                   :author_details, :labeler, :reviewers, :assignees, :work_item
 
+      PR_DESCRIPTION_MAX_LENGTH = 3999 # Azure has a max length of 4000 (0 based count)
+      PR_DESCRIPTION_ENCODING = Encoding::UTF_16 # Azure DevOps only support descriptions up to 4000 characters in UTF-16
+      # encoding.
+      # https://developercommunity.visualstudio.com/content/problem/608770/remove-4000-character-limit-on-pull-request-descri.html
```
jeffwidman, line 16:
```suggestion
      # Azure DevOps limits PR descriptions to a max of 4,000 characters in UTF-16 encoding:
      # https://developercommunity.visualstudio.com/content/problem/608770/remove-4000-character-limit-on-pull-request-descri.html
      PR_DESCRIPTION_MAX_LENGTH = 3_999 # 0 based count
      PR_DESCRIPTION_ENCODING = Encoding::UTF_16 
```

common/lib/dependabot/clients/codecommit.rb
dwc0011: I can add one in, I just haven't had the time, yet.  I hit this as time allows. I don't think it is high priority as it has not been causing issues (that I know of) the linter was just complaining about it, I don't think it needs to hold this one up.

common/lib/dependabot/pull_request_creator/github.rb
```diff
@@ -9,7 +9,11 @@ module Dependabot
   class PullRequestCreator
     # rubocop:disable Metrics/ClassLength
     class Github
-      MAX_PR_DESCRIPTION_LENGTH = 65_536 # characters (see #create_pull_request)
+      PR_DESCRIPTION_MAX_LENGTH = 65_536 # Limit PR description to PR_DESCRIPTION_MAX_LENGTH (65,536) characters
+      # and truncate with message if over. The API limit is 262,144 bytes
+      # (https://github.community/t/maximum-length-for-the-comment-body-in-issues-and-pr/148867/2).
+      # As Ruby strings are UTF-8 encoded, this is a pessimistic limit: it
+      # presumes the case where all characters are 4 bytes.
```
jeffwidman, line 16:
```suggestion
      # GitHub limits PR descriptions to a max of 65,536 characters:
      # https://github.com/orgs/community/discussions/27190#discussioncomment-3726017
      PR_DESCRIPTION_MAX_LENGTH = 65_536
```

common/lib/dependabot/pull_request_creator/codecommit.rb
```diff
@@ -10,6 +10,9 @@ class Codecommit
                   :files, :commit_message, :pr_description, :pr_name,
                   :author_details, :labeler
 
+      PR_DESCRIPTION_MAX_LENGTH = 10_239 # CodeCommit has a max length of 10240 (0 based count)
+      # https://docs.aws.amazon.com/codecommit/latest/APIReference/API_PullRequest.html
```
jeffwidman, line 14:
```suggestion
     # CodeCommit limits PR descriptions to a max length of 10,240 characters:
     # https://docs.aws.amazon.com/codecommit/latest/APIReference/API_PullRequest.html
      PR_DESCRIPTION_MAX_LENGTH = 10_239 # 0 based count
```

common/lib/dependabot/clients/azure.rb
jeffwidman, line 25:
```suggestion
```

common/lib/dependabot/pull_request_creator/github.rb
```diff
@@ -9,7 +9,9 @@ module Dependabot
   class PullRequestCreator
     # rubocop:disable Metrics/ClassLength
     class Github
-      MAX_PR_DESCRIPTION_LENGTH = 65_536 # characters (see #create_pull_request)
+      # GitHub limits PR descriptions to a max of 65,536 characters:
+      # https://github.com/orgs/community/discussions/27190#discussioncomment-3726017
+      PR_DESCRIPTION_MAX_LENGTH = 65_536
```
jeffwidman, line 14:
```suggestion
      PR_DESCRIPTION_MAX_LENGTH = 65_535 # 0 based count
````

**Review by jeffwidman — COMMENTED** [review]
I applied a few more mostly cosmetic changes, we've worked together enough that I was pretty sure you wouldn't mind if it helps this land sooner. 😁 

This is my last remaining question--once you address this I'll approve/merge:
https://github.com/dependabot/dependabot-core/pull/7487/files#r1273880291

File: common/lib/dependabot/clients/azure.rb (hunk 9 above)
Comment (dwc0011) [reply common/lib/dependabot/clients/azure.rb:25]: Yes that was a miss.  It should no longer be in the Client.

**Review by jeffwidman — APPROVED** [review]
Thank you again!

File: common/lib/dependabot/pull_request_creator/codecommit.rb [hunk 14]
```diff
@@ -10,6 +10,10 @@ class Codecommit
                   :files, :commit_message, :pr_description, :pr_name,
                   :author_details, :labeler
 
+     # CodeCommit limits PR descriptions to a max length of 10,240 characters:
+     # https://docs.aws.amazon.com/codecommit/latest/APIReference/API_PullRequest.html
+      PR_DESCRIPTION_MAX_LENGTH = 10_239 # 0 based count
```
jeffwidman, line 15:
```suggestion
      # CodeCommit limits PR descriptions to a max length of 10,240 characters:
      # https://docs.aws.amazon.com/codecommit/latest/APIReference/API_PullRequest.html
      PR_DESCRIPTION_MAX_LENGTH = 10_239 # 0 based count
```

common/lib/dependabot/pull_request_creator/azure.rb
```diff
@@ -10,6 +10,11 @@ class Azure
                   :files, :commit_message, :pr_description, :pr_name,
                   :author_details, :labeler, :reviewers, :assignees, :work_item
 
+      # Azure DevOps limits PR descriptions to a max of 4,000 characters in UTF-16 encoding:
+      # https://developercommunity.visualstudio.com/content/problem/608770/remove-4000-character-limit-on-pull-request-descri.html
+      PR_DESCRIPTION_MAX_LENGTH = 3_999 # 0 based count
+      PR_DESCRIPTION_ENCODING = Encoding::UTF_16
```
jeffwidman, line 16:
```suggestion
      PR_DESCRIPTION_ENCODING = Encoding::UTF_16
```

jeffwidman:
Yeah, I screwed up the whitespace using the GitHub UI ```suggestion tool... I just fixed it.

PR #7491, opened 2023-06-27
Update README.md

jmartens:
Add warning to dependabot-scripts section along the lines of the warning in dependabot-scripts mentioning it is broken after the monolith refactoring


jeffwidman (approved):
Good catch, and sorry about the breakage. 😢

PR #7492, opened 2023-06-27
Stop recording the `ecosystem` param

jeffwidman:
I'm starting a broader project to re-think how we instrument the ecosystem (language * package manager) versions.

As part of this, the shape of the data that is sent to the internal API is changing, and the `ecosystem` param is no longer a separate k/v param.

Although the cloud version of the API no longer uses/requires this param, customers running GHES could upgrade their version of `dependabot-core` but against an older version of the API which still requires/uses this param. So I continued to send the param, but as a placeholder string of `"deprecated"`, so that nothing breaks.

Again, this param is no longer used by the cloud's version of the API, so it will never see the `ecosystem` placeholder value.


cargo/lib/dependabot/cargo/file_fetcher.rb
```diff
@@ -28,9 +28,8 @@ def package_manager_version
                   end
 
         {
-          ecosystem: "cargo",
           package_managers: {
-            "channel" => channel
+            "cargo" => channel
```
jeffwidman, line 32: I consciously renamed this from `"channel"` to `"cargo"` because that seemed more consistent with the pattern in other ecosystems.

In the long term, it probably won't matter because the entire shape of the data is changing, but in the short term it just seemed more straightforward this way.

jeffwidman:
Changing to `draft` as it may be simpler to do all this work as a one-time flip to the new data format rather than pull it off piecemeal and then adding it back...

jeffwidman:
This is ready for review.

Smoke test failures are expected, TBH it might be simplest to fix them after merging the related PR's so that I don't have to keep re-doing them in lockstep.

But keeping the PR's here in `core` separate for easier bite-size review.

brrygrdn (approved):

PR #7494, opened 2023-06-29
Fix Github Actions dependency parsing edge case

deivid-rodriguez:
When the action changed from "branch versions" to "tag versions", we would fail to detect updates.

For example, in https://github.com/Swatinem/rust-cache, "v1" is a branch, while "v2" is a tag.

Fixes https://github.com/dependabot/dependabot-core/issues/7384#issuecomment-1609704958.


jurre (approved):

tamird:
Is this ready to merge? This bug remains at large.

deivid-rodriguez:
Sorry for the delay @tamird, this is now live!

PR #7497, opened 2023-06-29
Bump Bundler to 2.4.15

deivid-rodriguez:
New version has a working `bundle lock --update --bundler` command, so we can bump the Bundler version locally without having to actually install any gems.


Nishnha (approved):

jeffwidman (approved):

jeffwidman:
Actually we should let @deivid-rodriguez merge/deploy this one, as he will best know if any bugs have cropped up preventing it...

deivid-rodriguez:
Both 2.4.15 and 2.4.16 had some regressions over previous versions, so I'd go straight to 2.4.17. If that passes CI, I think it's good to go as is, since 2.4.17 did not have any regressions reported.

Issue #4948, opened 2022-04-04
Improve backtrace for spec failures caused by empty native helper output 


mattt:
Dependabot native helpers use JSON for inter-process communication. When a test fails because the helper executable fails to produce any output, the log message looks like this:

```
1) Dependabot::Bundler::FileUpdater#updated_dependency_files the updated lockfile when the Gemfile.lock didn't have a BUNDLED WITH line doesn't add in a BUNDLED WITH
   Failure/Error:
     SharedHelpers.run_helper_subprocess(
       command: command,
       function: function,
       args: args,
       env: {
         # Bundler will pick the matching installed major version
         "BUNDLER_VERSION" => bundler_version,
         "BUNDLE_GEMFILE" => File.join(helpers_path, "Gemfile"),
         "GEM_PATH" => File.join(helpers_path, "vendor/cache")
       }
   Dependabot::SharedHelpers::HelperSubprocessFailed:
   # /common/lib/dependabot/shared_helpers.rb:128:in `rescue in run_helper_subprocess'
   # /common/lib/dependabot/shared_helpers.rb:77:in `run_helper_subprocess'
   # ./lib/dependabot/bundler/native_helpers.rb:44:in `block in run_bundler_subprocess'
   # ./lib/dependabot/bundler/native_helpers.rb:40:in `run_bundler_subprocess'
   # ./lib/dependabot/bundler/file_updater/lockfile_updater.rb:68:in `block in build_updated_lockfile'
   # /common/lib/dependabot/shared_helpers.rb:48:in `block in in_a_temporary_directory'
   # /common/lib/dependabot/shared_helpers.rb:48:in `chdir'
   # /common/lib/dependabot/shared_helpers.rb:48:in `in_a_temporary_directory'
   # /common/lib/dependabot/shared_helpers.rb:37:in `in_a_temporary_repo_directory'
   # ./lib/dependabot/bundler/file_updater/lockfile_updater.rb:62:in `build_updated_lockfile'
   # ./lib/dependabot/bundler/file_updater/lockfile_updater.rb:46:in `updated_lockfile_content'
   # ./lib/dependabot/bundler/file_updater.rb:157:in `updated_lockfile_content'
   # ./lib/dependabot/bundler/file_updater.rb:41:in `updated_dependency_files'
   # ./spec/dependabot/bundler/file_updater_spec.rb:54:in `block (3 levels) in <top (required)>'
[...]

PR #7498, opened 2023-06-29
Dwc0011/improve native helper errors

dwc0011:
Addresses #4948 by enhancing the error message while not divulging secrets.

This is a work in progress, but I want to run the specs and begin the conversation as to the solution.


Mroczek1992:
Ufvjh iig

PR #7502, opened 2023-06-30
Detect interpolation in terragrunt sources and skip if present

### LINKED ISSUE #5841: Terragrunt source results in error

### Package ecosystem

Terraform

### Package manager version

_No response_

### Language version

_No response_

_No response_

### dependabot.yml content

_No response_

### Updated dependency

_No response_

What I expected: The terragrunt source to be evaluated and if unable to, an error message would be logged and dependabot would continue processing the remainder of the reposititory.
What I saw: The path was incorrectly thought to be a registry source and found invalid.  Dependabot threw an error and stopped processing the remainder of the repository.

```
/home/dependabot/dependabot-script/vendor/ruby/2.7.0/gems/dependabot-terraform-0.211.0/lib/dependabot/terraform/file_parser.rb:213:in `registry_source_details_from': Invalid registry source specified: '${path_relative_from_include()}/..//.ci/${path_relative_to_include()}' (Dependabot::DependencyFileNotEvaluatable)
--
from /home/dependabot/dependabot-script/vendor/ruby/2.7.0/gems/dependabot-terraform-0.211.0/lib/dependabot/terraform/file_parser.rb:180:in `source_from'
from /home/dependabot/dependabot-script/vendor/ruby/2.7.0/gems/dependabot-terraform-0.211.0/lib/dependabot/terraform/file_parser.rb:145:in `build_terragrunt_dependency'
from /home/dependabot/dependabot-script/vendor/ruby/2.7.0/gems/dependabot-terraform-0.211.0/lib/dependabot/terraform/file_parser.rb:71:in `block (2 levels) in parse_terragrunt_files'
from /home/dependabot/dependabot-script/vendor/ruby/2.7.0/gems/dependabot-terraform-0.211.0/lib/dependabot/terraform/file_parser.rb:68:in `each'
from /home/dependabot/dependabot-script/vendor/ruby/2.7.0/gems/dependabot-terraform-0.211.0/lib/dependabot/terraform/file_parser.rb:68:in `block in parse_terragrunt_files'
from /home/dependabot/dependabot-script/vendor/ruby/2.7.0/gems/dependabot-terraform-0.211.0/lib/dependabot/terraform/file_parser.rb:66:in `each'
from /home/dependabot/dependabot-script/vendor/ruby/2.7.0/gems/dependabot-terraform-0.211.0/lib/dependabot/terraform/file_parser.rb:66:in `parse_terragrunt_files'
from /home/dependabot/dependabot-script/vendor/ruby/2.7.0/gems/dependabot-terraform-0.211.0/lib/dependabot/terraform/file_parser.rb:34:in `parse'
from /home/dependabot/dependabot-script/updater.rb:81:in `update'
```

Terragrunt will evaluate the path from the method and use it as a local path.

_No response_

```
terraform {
 source          = "${path_relative_from_include()}/..//.ci/${path_relative_to_include()}"
 }
```

dwc0011:
Fixes #5841 by using a regex to determine if interpolation syntax is used anywhere in a source.

### DIFF
--- a/terraform/lib/dependabot/terraform/file_parser.rb
+++ b/terraform/lib/dependabot/terraform/file_parser.rb
@@ -68,7 +68,11 @@ module Dependabot
           modules.each do |details|
             next unless details["source"]
 
-            dependency_set << build_terragrunt_dependency(file, details)
+            source = source_from(details)
+            # Cannot update local path modules, skip
+            next if source.nil? || source[:type] == "path"
+
+            dependency_set << build_terragrunt_dependency(file, source)
           end
         end
       end
@@ -141,15 +145,8 @@ module Dependabot
         details.is_a?(String)
       end
 
-      def build_terragrunt_dependency(file, details)
-        source = source_from(details)
-        dep_name =
-          if Source.from_url(source[:url])
-            Source.from_url(source[:url]).repo
-          else
-            source[:url]
-          end
-
+      def build_terragrunt_dependency(file, source)
+        dep_name = Source.from_url(source[:url]) ? Source.from_url(source[:url]).repo : source[:url]
         version = version_from_ref(source[:ref])
 
         Dependency.new(
@@ -178,6 +175,8 @@ module Dependabot
             git_source_details_from(bare_source)
           when :registry
             registry_source_details_from(bare_source)
+          when :interpolation
+            return nil
           end
 
         source_details[:proxy_url] = raw_source if raw_source != bare_source
@@ -261,6 +260,7 @@ module Dependabot
 
       # rubocop:disable Metrics/PerceivedComplexity
       def source_type(source_string)
+        return :interpolation if source_string.match?(/\$\{[^}]+\}/)
         return :path if source_string.start_with?(".")
         return :github if source_string.start_with?("github.com/")
         return :bitbucket if source_string.start_with?("bitbucket.org/")
--- a/terraform/lib/dependabot/terraform/file_parser.rb
+++ b/terraform/lib/dependabot/terraform/file_parser.rb
@@ -260,7 +260,7 @@ module Dependabot
 
       # rubocop:disable Metrics/PerceivedComplexity
       def source_type(source_string)
-        return :interpolation if source_string.match?(%r{\$\{[^{}/]+\}})
+        return :interpolation if source_string.include?("${")
         return :path if source_string.start_with?(".")
         return :github if source_string.start_with?("github.com/")
         return :bitbucket if source_string.start_with?("bitbucket.org/")
--- a/terraform/lib/dependabot/terraform/file_parser.rb
+++ b/terraform/lib/dependabot/terraform/file_parser.rb
@@ -69,7 +69,7 @@ module Dependabot
             next unless details["source"]
 
             source = source_from(details)
-            # Cannot update local path modules, skip
+            # Cannot update nil (interpolation sources) or local path modules, skip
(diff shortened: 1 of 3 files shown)


dwc0011:
@deivid-rodriguez @jeffwidman Any estimate on when you all will be able to look at this?

jeffwidman (approved):
Makes sense, thanks for all the helpful context in https://github.com/dependabot/dependabot-core/issues/5841.

terraform/lib/dependabot/terraform/file_parser.rb
```diff
@@ -68,7 +68,11 @@ def parse_terragrunt_files(dependency_set)
           modules.each do |details|
             next unless details["source"]
 
-            dependency_set << build_terragrunt_dependency(file, details)
+            source = source_from(details)
+            # Cannot update nil (interpolation sources) or local path modules, skip
+            next if source.nil? || source[:type] == "path"
```
jeffwidman, line 73: I don't know how this code is used, and don't know Terragrunt/Terraform very well.

Obviously we can't update local path modules as deps to the current terragrunt files, but is there any risk that this prevents updating a local path module that's sitting in the same monorepo? Ie, because this stops :dependabot: from parsing the local module, it also prevent Dependabot from updating any remote deps listed in _that_ local module?

I highly doubt that's a risk, so I'm going to merge, but wanted to doublecheck with you @dwc0011 since you're the expert.

lorengordon: I don't think it would _prevent_ it, exactly... There simply is no way for Dependabot to know how the Terragrunt interpolation resolves. The resolution depends on the Terragrunt config at runtime, and may not even be _consistent_. So when the user is using Terragrunt interpolation this way, Dependabot simply _can't_ use its recursive feature to update anything in that other path anyway.

The workaround for the user would be just to specify both paths in the Dependabot config, so then Dependabot processes both. Which honestly is my preference anyway; the recursive feature in Dependabot has been a bit of a pain and I'd rather just disable it. (Though now with grouping it is becoming of more interest...)

dwc0011: For terraform local path modules, dependabot does not support updating local path modules, so it makes sense that terragrunt local path modules would not be updated as well. 
Local Path Module Files are found and traversed by the file fetcher, not the file parser, all files including those where local path modules are found are then passed to the file parser. That said Terragrunt files do not have a fetch local path module at this time. (This could be added in a separate PR)

As mentioned, the file fetcher is responsible for gathering all the files to process including local source module files.  
For terraform files see: [fetch local path modules](https://github.com/dependabot/dependabot-core/blob/2530fd02738ddfd6b6251a736953c18946f7ec71/terraform/lib/dependabot/terraform/file_fetcher.rb#L54C11-L54C34 ). 


Since the process is fetch the files, parse, check updatable, and then update; not returning terragrunt local path dependency in the file_parser makes the most sense.  We do this for other types that are not supported as well. i.e. local path modules for terraform, interpolation for terragrunt (added by this PR), 


Even if the dependency was added in the file parser, it would be removed later in the [update checker](https://github.com/dependabot/dependabot-core/blob/2530fd02738ddfd6b6251a736953c18946f7ec71/terraform/lib/dependabot/terraform/update_checker.rb#L193)

jeffwidman: Thanks both of you. That's very helpful context, and I appreciate the extra details... Will be useful for any future `git blame` spelunkers.

PR #7516, opened 2023-07-03
Only record ecosystem versions when flag set

jeffwidman:
There's no point in recording the versions on GHES / non-cloud instances because we can't access that data for telemetry purposes. So put it behind a feature flag.

This will also make it so later on when we do the rename of `record_package_manager_version` to `record_ecosystem_versions`, there is less to worry about for backwards incompatible changes on GHES.

Before merging this, first need to merge the internal change that sets this feature flag, so that we don't lose visibility on cloud after we merge it.

Related:
* https://github.com/dependabot/dependabot-core/pull/7492
* https://github.com/dependabot/dependabot-core/pull/7517

### DIFF
--- a/updater/lib/dependabot/file_fetcher_command.rb
+++ b/updater/lib/dependabot/file_fetcher_command.rb
@@ -21,8 +21,14 @@ module Dependabot
         @base_commit_sha = file_fetcher.commit
         raise "base commit SHA not found" unless @base_commit_sha
 
-        version = file_fetcher.package_manager_version
-        api_client.record_package_manager_version(version[:package_managers]) unless version.nil?
+        # We don't set this flag in GHES because there's no point in recording versions since we can't access that data.
+        # TODO: The flag is named `record_ecosystem_versions` because `package_manager_version` is getting renamed to
+        # to that shortly... but splitting into separate PR's for lower risk/easiery testability. Once the follow-on PR
+        # with the rename lands, the name confusion will disappear.
+        if Experiments.enabled?(:record_ecosystem_versions)
+          version = file_fetcher.package_manager_version
+          api_client.record_package_manager_version(version[:package_managers]) unless version.nil?
+        end
 
         dependency_files
       rescue StandardError => e


jeffwidman:
Smoke test failures are expected, because it doesn't pass the flag.

Since this is a flag we'd want to run for _all_ smoke tests, I _think_ we need to update the [cli](https://github.com/dependabot/cli) to always pass the flag, rather than updating the individual the smoke tests.

jeffwidman:
Before merging this, first need to merge the internal change that sets this feature flag, so that we don't lose visibility on cloud after we merge it.

updater/lib/dependabot/file_fetcher_command.rb
```diff
@@ -21,8 +21,14 @@ def perform_job
         @base_commit_sha = file_fetcher.commit
         raise "base commit SHA not found" unless @base_commit_sha
 
-        version = file_fetcher.package_manager_version
-        api_client.record_package_manager_version(version[:ecosystem], version[:package_managers]) unless version.nil?
+        # We don't set this flag in GHES because there's no point in recording versions since we can't access that data.
+        # TODO: The flag is named `record_ecosystem_versions` because `package_manager_version` is getting renamed to
+        # to that shortly... but splitting into separate PR's for lower risk/easiery testability. Once the follow-on PR
+        # with the rename lands, the name confusion will disappear.
+        if Experiments.enabled?(:record_ecosystem_versions)
```
brrygrdn, line 28: Nothing to do with this change, but I wonder if we should eventually split `Experiments` into `Features` for long-lived switches and `Experiments` for short-term ones 🤔

brrygrdn (approved):
👍🏻

updater/lib/dependabot/file_fetcher_command.rb
jeffwidman: Totally agree. I think the internal repo that controls setting these flags will be the best place to discuss a potential refactoring (the result of which would then flow out here into core).

jeffwidman:
As mentioned elsewhere, going to merge despite the failing smoke tests as there's a chain of connected PR's around this ecosystem metrics refactor, split apart for easier review, and then will update the smoke tests at the end once they land.

PR #7517, opened 2023-07-04
Switch to using the new `record_ecosystem_versions` endpoint.

jeffwidman:
This flips the `updater` to calling the new `record_ecosystem_versions`
endpoint.

Additionally, this flips the per-ecosystem implementations of
`package_manager_version` to this new `ecosystem_versions` hash.

Note that we expect the shape of the data to be slightly different in
the long term for these metrics... ie, we plan to record a
`min`/`max`/`raw` or something along those lines for the range of
versions supported by the manifest.

However, changing that will be done in separate PR's for easier
review/discussion/audit of those individual ecosystems.

Related:
* https://github.com/dependabot/dependabot-core/pull/7492
* https://github.com/dependabot/dependabot-core/pull/7516

### DIFF
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -799,8 +799,8 @@ StackProf.stop if $options[:profile]
 StackProf.results("tmp/stackprof-#{Time.now.strftime('%Y-%m-%d-%H:%M')}.dump") if $options[:profile]
 
 puts "🌍 Total requests made: '#{$network_trace_count}'"
-package_manager = fetcher.package_manager_version
-puts "🎈 Package manager version log: #{package_manager}" unless package_manager.nil?
+ecosystem_versions = fetcher.ecosystem_versions
+puts "🎈 Ecosystem Versions log: #{ecosystem_versions}" unless ecosystem_versions.nil?
 
 # rubocop:enable Metrics/BlockLength
 
--- a/common/lib/dependabot/file_fetchers/base.rb
+++ b/common/lib/dependabot/file_fetchers/base.rb
@@ -102,7 +102,7 @@ module Dependabot
         raise Dependabot::RepoNotFound, source
       end
 
-      def package_manager_version
+      def ecosystem_versions
         nil
       end
 
--- a/updater/lib/dependabot/api_client.rb
+++ b/updater/lib/dependabot/api_client.rb
@@ -119,13 +119,10 @@ module Dependabot
       sleep(rand(3.0..10.0)) && retry
     end
 
-    def record_package_manager_version(package_managers)
-      api_url = "#{base_url}/update_jobs/#{job_id}/record_package_manager_version"
+    def record_ecosystem_versions(ecosystem_versions)
+      api_url = "#{base_url}/update_jobs/#{job_id}/record_ecosystem_versions"
       body = {
-        data: {
-          ecosystem: "deprecated",
-          "package-managers": package_managers
-        }
+        data: { ecosystem_versions: ecosystem_versions }
       }
       response = http_client.post(api_url, json: body)
       raise ApiError, response.body if response.code >= 400
--- a/updater/lib/dependabot/file_fetcher_command.rb
+++ b/updater/lib/dependabot/file_fetcher_command.rb
@@ -22,12 +22,9 @@ module Dependabot
         raise "base commit SHA not found" unless @base_commit_sha
 
         # We don't set this flag in GHES because there's no point in recording versions since we can't access that data.
-        # TODO: The flag is named `record_ecosystem_versions` because `package_manager_version` is getting renamed to
-        # to that shortly... but splitting into separate PR's for lower risk/easiery testability. Once the follow-on PR
-        # with the rename lands, the name confusion will disappear.
         if Experiments.enabled?(:record_ecosystem_versions)
-          version = file_fetcher.package_manager_version
-          api_client.record_package_manager_version(version[:package_managers]) unless version.nil?
+          ecosystem_versions = file_fetcher.ecosystem_versions
+          api_client.record_ecosystem_versions(ecosystem_versions) unless ecosystem_versions.nil?
         end
 
         dependency_files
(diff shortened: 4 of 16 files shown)


jeffwidman:
This is stacked on top of https://github.com/dependabot/dependabot-core/pull/7516 and only the final commit is new / needs review.

updater/lib/dependabot/api_client.rb
```diff
@@ -119,13 +119,10 @@ def update_dependency_list(dependencies, dependency_files)
       sleep(rand(3.0..10.0)) && retry
     end
 
-    def record_package_manager_version(ecosystem, package_managers)
-      api_url = "#{base_url}/update_jobs/#{job_id}/record_package_manager_version"
+    def record_ecosystem_versions(ecosystem_versions)
```
brrygrdn, line 122: Just making sure I've thought about this properly: this endpoint is new, so if someone pulls down an image with this change to GHES the service might not have that endpoint but that's fine as we've de-fanged the call unless an experiment is enabled in the [previous PR](https://github.com/dependabot/dependabot-core/pull/7516) and it will never been enabled for any version of GHES.

bin/dry-run.rb
```diff
@@ -799,8 +799,8 @@ def security_fix?(dependency)
 StackProf.results("tmp/stackprof-#{Time.now.strftime('%Y-%m-%d-%H:%M')}.dump") if $options[:profile]
 
 puts "🌍 Total requests made: '#{$network_trace_count}'"
-package_manager = fetcher.package_manager_version
-puts "🎈 Package manager version log: #{package_manager}" unless package_manager.nil?
+e_versions = fetcher.ecosystem_versions
+puts "🎈 Ecosystem Versions log: #{e_versions}" unless e_versions.nil?
```
brrygrdn, line 803:
```suggestion
ecosystem_versions = fetcher.ecosystem_versions
puts "🎈 Ecosystem Versions log: #{ecosystem_versions}" unless ecosystem_versions.nil?
```

Style nit: Variable contractions always feel a little un-Ruby to me

updater/lib/dependabot/file_fetcher_command.rb
```diff
@@ -22,12 +22,9 @@ def perform_job
...
 
         # We don't set this flag in GHES because there's no point in recording versions since we can't access that data.
-        # TODO: The flag is named `record_ecosystem_versions` because `package_manager_version` is getting renamed to
-        # to that shortly... but splitting into separate PR's for lower risk/easiery testability. Once the follow-on PR
-        # with the rename lands, the name confusion will disappear.
         if Experiments.enabled?(:record_ecosystem_versions)
-          version = file_fetcher.package_manager_version
-          api_client.record_package_manager_version(version[:ecosystem], version[:package_managers]) unless version.nil?
+          e_versions = file_fetcher.ecosystem_versions
+          api_client.record_ecosystem_versions(e_versions) unless e_versions.nil?
```
brrygrdn, line 27:
```suggestion
ecosystem_versions = file_fetcher.ecosystem_versions
api_client.record_ecosystem_versions(ecosystem_versions) unless ecosystem_versions.nil?
```

Sorry, it's really a really minor but deeply engrained nitpick 😅

brrygrdn (approved):
This makes sense to me, 👍🏻

updater/lib/dependabot/api_client.rb
jeffwidman: Exactly.

updater/lib/dependabot/file_fetcher_command.rb
jeffwidman: hah, I had that originally then decided it was too long. But I didn't want `versions` like it used to be as that felt too ambiguous. 👍 for this from me.

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -102,7 +102,7 @@ def clone_repo_contents
         raise Dependabot::RepoNotFound, source
       end
 
-      def package_manager_version
+      def ecosystem_versions
```
jeffwidman, line 105: This needs a code comment explaining the expected shape of the data, or possibly a doc somewhere for devs, but we're still noodling internally on what we want this to look like, so punting on those docs until later. 

Right now just want to get it swapped over to this new endpoint so can start prototyping actual code to see how my design ideas hold up to reality--ie, do we want to record more/different info then I expect.

updater/spec/dependabot/file_fetcher_command_spec.rb
```diff
@@ -109,8 +109,9 @@
         allow_any_instance_of(Dependabot::Bundler::FileFetcher).
           to receive(:files).
           and_raise(exception)
+        # TODO this may be deletable since the ecosystem_versions call is now buried behind a flag...
```
jeffwidman, line 112: TODO before merge: try this to see if i can safely delete it... may need to first get some working ecosystems ported over to this though... like bundler, npm_and_yarn, etc.

updater/spec/dependabot/integration_spec.rb
```diff
@@ -51,14 +51,15 @@
...
+                    record_ecosystem_versions: nil,
                     increment_metric: nil)
   end
   let(:file_fetcher) do
     instance_double(Dependabot::FileFetchers::Base,
                     files: dependency_files,
                     commit: "sha",
-                    package_manager_version: nil)
+                    # TODO this may be deletable since the ecosystem_versions call is now buried behind a flag...
+                    ecosystem_versions: nil)
```
jeffwidman, line 62: See https://github.com/dependabot/dependabot-core/pull/7517/files#r1252257873

updater/spec/dependabot/integration_spec.rb
```diff
@@ -79,8 +80,6 @@
 
     # Stub Dependabot object with instance doubles
     allow(Dependabot::ApiClient).to receive(:new).and_return(api_client)
-    # Recording the package manager happens via an observer so the instantiated `api_client` does not receive this call
-    allow_any_instance_of(Dependabot::ApiClient).to receive(:record_package_manager_version)
```
jeffwidman, line 83: See https://github.com/dependabot/dependabot-core/pull/7517/files#r1252257873

jeffwidman:
These test failures indicate may need to port the callers of this as part of this PR as well... or bury the package manager calls in private methods since the `ecosystem_versions()` will likely call both `package_manager_versions` and `programming_language_versions` :
```ruby
rspec './spec/dependabot/bundler/file_fetcher_spec.rb[1:1:1:5]' # Dependabot::Bundler::FileFetcher behaves like a dependency file fetcher the class doesn't define any additional public instance methods
```

PR #7520, opened 2023-07-06
[Updater] Spike out a possible implementation for grouping by dependency-type

brrygrdn:
This is an early exploration in going from 'there is 1 way to group dependencies' to 'there are N ways to group dependencies', with splitting by development/production types as the first new rule.

This is very much a sketch as there's some things we need to figure out around how rules interact and how we perform validation when multiple rules are involved which can interact with or conflict with higher-order configuration points.


brrygrdn:
This is superseded by https://github.com/dependabot/dependabot-core/pull/7581

PR #7525, opened 2023-07-06
Support for Swift package manager

deivid-rodriguez:
This PR implements #1245.

For now just a draft but most of the functionality I intend to ship initially should be already there.


deivid-rodriguez:
One question for @stevapple (maybe @0xTim knows too?).

So, our updater runs under Linux, and since we're running Swift directly, I've encountered a problem where we're adding Linux specific dependencies to some manifests.

For example, whenever we find [this dependency](https://github.com/stephencelis/SQLite.swift/blob/d34f63a90a153a8c54ea743f88978faf83d01a91/Package.swift#L41-L44).

Can you think of any workaround? I was thinking we could maybe ask upstream for some feature that allows to resolve under a specific platform, like `swift package update --platform macOS` or something. But regardless of that, it'd be nice to figure out a way to somehow override this `#if os()` macro.

That said, what I've been able to see from some forum research is that this may be an old hack and not what's recommended right now. Maybe the right way now is [conditional dependencies](https://developer.apple.com/documentation/packagedescription/target/dependency/target(name:condition:)). If that's the case, maybe it's best to let libraries using this old hack fix themselves.

stevapple:
The only way to deal with such hack is to syntactically parse the file with SwiftSyntax, IIRC. It’s hard and not worth the effort. 
Some packages may rely on this hack because their upstream used it and some because they’re directly using platform-specific APIs in the manifest, so removing or bypassing can be likely breaking.

I would suggest to leave it alone and let the library author fix it. These packages are also not likely to have `Package.resolved` committed, because it could differ from platform to platform.

> I've encountered a problem where we're adding Linux specific dependencies to some manifests.

I don’t think it’s a big problem… SwiftPM will remove unnecessary packages and add necessary ones during resolution.

deivid-rodriguez:
Thanks for the input, that's reassuring!

deivid-rodriguez:
Not fully done yet, but happy to get some early feedback here!

swift/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+    libedit2 \
+    libgcc-9-dev \
+    libpython2.7 \
+    libsqlite3-0 \
+    libstdc++-9-dev \
+    libxml2 \
+    libz3-dev \
+    pkg-config \
+    tzdata \
+    uuid-dev \
```
jakecoffman, line 21: There's some dependencies I'm curious about, like sqlite3 and libpython2.7. Are these all needed?

jakecoffman (approved):
Looks good, maybe we can start some smoke tests too? 👏

swift/Dockerfile
deivid-rodriguez: I got no idea. I copied this from the table at https://www.swift.org/download/#using-downloads (Linux section).

deivid-rodriguez:
Yes, I'll create some smoke tests tomorrow!

swift/Dockerfile
stevapple: `libpython2.7` looks suspicious, as Swift already moves to Python 3. Also, Python is only required for LLDB, so as long as we don’t use the debugger it should be unnecessary.

`sqlite3` is required for SwiftPM.

0xTim: What version of Ubuntu is this targeting? 22.04 uses Python 3

0xTim: 🤦 I see below it's using 20.04 - why not use Jammy here?

deivid-rodriguez: Unfortunately our base image still uses 20.04 so we cannot move on just yet. See https://github.com/dependabot/dependabot-core/pull/5030.

deivid-rodriguez:
Smoke test created at https://github.com/dependabot/smoke-tests/pull/106.

My understanding is that we need to merge this and let an initial updater image for Swift be published so that the smoke test can pass.

deivid-rodriguez:
Actually I just pushed a small trick to verify smoke tests before merging.

deivid-rodriguez:
The smoke test is passing now! And it actually detected a real issue! I will next merge the smoke test PR, regenerate smoke tests cache, and then edit this PR to run against the default branch of the smoke tests repo.

PR #7534, opened 2023-07-08
Track unknown errors

Nishnha:
Dependabot currently treats `unknown_errors` the same as any other error type.

However, `unknown_errors` are a bit of a special case because they are mostly caused by a HelperSubprocessFailure when Dependabot shells out to native package managers. The helper subprocess failures are harder to debug because they can be specific to a dependency set or some combination of a package manager version and project file features.

In order to better investigate unknown errors, we should send the error details to the backend service so we can look for trends.

This PR makes a few changes to the way unknown errors are logged:

1. unknown errors can only be generated from a few sources:
    a. helper subprocess failures
    b. file fetcher command errors
    c. update files command errors 
2. helper subprocess failure commands are now logged to the Dependabot job log so users can investigate issues themselves (https://github.com/dependabot/dependabot-core/pull/6521)
3. octokit 500 errors are treated as a new error type, `server_error`, since they are well-known
4. unknown errors also increment an `updater.unknown_error` metric with tags for the package manager

~Still needs a feature flag for GHES support and tests for the new endpoint.~

EDIT: Check has been added in Dependabot-api to to return immediately if in GHES or Proxima environment

### DIFF
--- a/common/lib/dependabot/shared_helpers.rb
+++ b/common/lib/dependabot/shared_helpers.rb
@@ -64,7 +64,7 @@ module Dependabot
 
       def initialize(message:, error_context:, error_class: nil, trace: nil)
         super(message)
-        @error_class = error_class || ""
+        @error_class = error_class || "HelperSubprocessFailed"
         @error_context = error_context
         @fingerprint = error_context[:fingerprint] || error_context[:command]
         @trace = trace
--- a/common/lib/dependabot/shared_helpers.rb
+++ b/common/lib/dependabot/shared_helpers.rb
@@ -130,14 +130,14 @@ module Dependabot
       return response["result"] if process.success?
 
       raise HelperSubprocessFailed.new(
-        message: response["error"],
+        message: "Command \"#{error_context['command']}\" failed with #{response['error']}",
         error_class: response["error_class"],
         error_context: error_context,
         trace: response["trace"]
       )
     rescue JSON::ParserError
       raise HelperSubprocessFailed.new(
-        message: stdout || "No output from command",
+        message: stdout.empty? ? "No output from command \"#{error_context['command']}\"" : stdout,
         error_class: "JSON::ParserError",
         error_context: error_context
       )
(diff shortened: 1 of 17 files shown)


updater/lib/dependabot/updater/error_handler.rb
```diff
@@ -90,9 +90,9 @@ def handle_job_error(error:, dependency_group: nil)
 
       # Provides logging for errors that occur outside of a dependency context
       def log_job_error(error:, error_type:, error_detail: nil)
-        if error_type == "unknown_error"
+        if error_type == "unknown_error" && !Dependabot.enterprise?
```
Nishnha, line 93: Instead of checking if we are on `enterprise?` here, we can just always post to the `unknown_error` endpoint and check if we are on enterprise on the backend. If we are, then we can return 200 and discard the error message since we can't use that info.

honeyankit:
```suggestion
        if error_type == "unknown_error"
```

common/lib/dependabot/shared_helpers.rb
```diff
@@ -130,19 +131,20 @@ def self.run_helper_subprocess(command:, function:, args:, env: nil,
       return response["result"] if process.success?
 
       raise HelperSubprocessFailed.new(
-        message: response["error"],
+        message: "#{response['error']} failed with command (#{error_context[:command]})",
```
jurre, line 134: Could there be any PII logged in the command that we're running that we should avoid logging?

jurre (approved):
The changes look good to me, but I am slightly worried we could be loggin PII when logging the commands we're running natively, so it might be good to have a look at that. Going to approve as to unblock this once we're confident about that, though 👍

common/lib/dependabot/shared_helpers.rb
honeyankit: @jurre , All these logs will be going through the new endpoint and in the new endpoint, we are configuring failbot to not send the capture logs to sentry. The logs will be going to splunk and GitHub data warehouse. 
https://github.com/github/dependabot-api/pull/4213/files#diff-8988e11e084c4f99d07033e93421b1119a84ee3d9f9e3c9a94be43959bc91420R74-R77

PR #7548, opened 2023-07-11
[Grouped Updates] Refactor the DependencyGroupEngine into an object, Improved logging for empty groups

brrygrdn:
Addresses https://github.com/dependabot/dependabot-core/pull/7075#discussion_r1165384710

We currently implement the `DependencyGroupEngine` as a module which uses class variables to act a little like a singleton as we weren't sure how it would be injected into the Updater when we created it so we defaulted to global scope.

With much of the architecture now settled, it makes sense to move away from this global object to instantiate it immediately after we parse the project's manifest files when creating a the `DependencySnapshot`

This ensures that any fatal issues parsing the rules can be handled here:
https://github.com/dependabot/dependabot-core/blob/779657e58e1d9d969b4d084e5179505f102595c8/updater/lib/dependabot/update_files_command.rb#L15-L25

## Improved handling of empty groups

Right now we quietly swallow any groups which are defined but do not correspond to any dependencies which means we fail over to doing single updates without calling out **why** the group wasn't attempted.

As part of this work, it became much easier to warn during the job startup when we find an empty group, which results in us logging a message like:

```
Please check your configuration as there are groups no dependencies match:
- group-a
- group-b

This can happen if:
- the group's 'pattern' rules are misspelled
- your configuration's 'allow' rules do not permit any of the dependencies that match the group
- the dependencies that match the group rules have been removed from your project
```

At runtime, we will now acknowledge the empty groups exist and specifically log that we are skipping them like so:

```
Skipping update group for 'group-a' as it does not match any allowed dependencies.
```

### Rebasing 

It is possible for us to find when rebasing a grouped PR that the group definition has now become empty due to all dependencies matching the group having been removed.

Rather than the job being a silent no-op, we will now **close** the pull request with the following comment:

```
None of your dependencies match this group anymore, you may need to update your configuration file to remove it or change its rules.
```

### TODO

- [x] Ship a change to the Dependabot service so it is aware of the `dependency_group_empty` close reason


updater/lib/dependabot/dependency_group_engine.rb
```diff
@@ -5,81 +5,80 @@
...
 
-    def self.reset!
-      @groups_calculated = false
-      @registered_groups = []
-
-      @dependency_groups = {}
-      @ungrouped_dependencies = []
-    end
+    def self.from_job_config(job:)
+      groups = job.dependency_groups.map do |group|
```
jakecoffman, line 23: Looks like the smoke tests are upset that `dependency_groups` is `nil`. Do we need to default that to `[]`?

jakecoffman:
LGTM other than the smoke test failures, makes sense 👍

brrygrdn:
The remaining smoke test failure is actually an issue with the test rig, which I've fixed here: https://github.com/dependabot/smoke-tests/pull/103

This does make me 🤔 about the decision to always parse the group configuration immediately after parsing dependencies in every job as some jobs may not actually involve using the groups like rebases but I think the problem here was ultimately a mandatory array being nil which is a defect we won't see in the real world.

We could push the evaluation of groups later but since it only warns it still seems better to me to perform it at the same call sight as evaluating the manifests and parsing dependencies. In future it might make sense to move the operation split up so we don't bother parsing the files when we are told to update a specific dependency/and or rebase a PR?

updater/lib/dependabot/dependency_group_engine.rb
brrygrdn: Fixed!

brrygrdn:
@jakecoffman This should be good to go now

jakecoffman (approved):

PR #7553, opened 2023-07-12
Migrate from `sentry-raven` to `sentry-ruby`

jeffwidman:
The Sentry team deprecated `sentry-raven` nearly three years ago:
* https://github.com/getsentry/sentry-ruby/pull/1125

So this migrates to their replacement `sentry-ruby` gem.

Beyond the namespace changes, the major logical changes are:
1. Events now default to sending asynchronously using a thread pool whereas raven used to send them synchronously.
2. The `raven_context` attached to exceptions is no longer automatically picked up. It must be explicitly wired in.
3. Processor classes were removed. This didn't matter much to us as we had only a single `ExceptionSanitizer`, which is easily wired into the `before_send()` interface.
4. The `config.logger` param was removed.


updater/lib/dependabot/base_command.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require "raven"
+require "sentry-ruby"
```
jeffwidman, line 3: I'm unclear if we actually need this given we already `require "sentry-ruby"` in  `fetch_files.rb`/`update_files.rb`.

updater/lib/dependabot/sentry.rb
```diff
@@ -1,15 +1,28 @@
...
-class ExceptionSanitizer < Raven::Processor
+class ExceptionSanitizer
   REPO = %r{[\w.\-]+/([\w.\-]+)}
   PATTERNS = {
     auth_token: /(?:authorization|bearer):? (\w+)/i,
     repo: %r{api\.github\.com/repos/#{REPO}|github\.com/#{REPO}}
   }.freeze
 
+  def process_sentry(event, hint)
+    # TODO migrate these payloads from the old "data" to the new "event"/"hint"
```
jeffwidman, line 12: This should be quick if I can get into the debugger to see the payload, but for some reason when I drop a `debugger` statement here, isn't firing during unit tests... Maybe they've mocked the actual Sentry call so the caller of this (`before_send()`) never fires?

updater/lib/dependabot/service.rb
```diff
@@ -61,26 +61,22 @@ def update_dependency_list(dependency_snapshot:)
...
+      tags = tags.merge({
+        update_job_id: job&.id,
+        package_manager: job&.package_manager,
+        repo_private: job&.repo_private?
+      }.compact)
+      extra = extra.merge({
+        dependency_name: dependency&.name,
+        dependency_group: dependency_group&.name
+      }.compact)
+      Sentry.capture_exception(error, tags: tags, extra: extra)
```
jeffwidman, line 79: The underlying signature changed from:
```ruby
def self.from_exception(exc, options = {}, &block) # old

def capture_exception(error, **options, &block) # new
```

updater/lib/dependabot/setup.rb
```diff
@@ -10,8 +10,8 @@
 end
 
 require "dependabot/sentry"
-Raven.configure do |config|
-  config.logger = Dependabot.logger
+Sentry.init do |config|
+  # config.logger = Dependabot.logger # TODO
```
jeffwidman, line 14: TODO: still need to research if this should be replaced with anything, or simply deleted...

updater/lib/dependabot/setup.rb
```diff
@@ -37,7 +37,31 @@
     pub
   )}x
 
-  config.processors += [ExceptionSanitizer]
+  config.before_send = lambda do |event, hint|
+    debugger
```
jeffwidman, line 41: TODO: need to get this to fire so I can debug this section of code, and see what the payloads look like...

updater/lib/dependabot/setup.rb
```diff
@@ -37,7 +37,31 @@
...
+    if !hint[:exception].nil?
+      ExceptionSanitizer.process_sentry(event, hint)
+
+      # TODO integrate these two...
+      # TODO copied from migration guide, I don't understand how this assignment works though?
+      # https://docs.sentry.io/platforms/ruby/migration/#exceptionraven_context
+      if exception = hint[:exception]
```
jeffwidman, line 48: Ruby n00b question: What does this construct do?

I see inline assignment, does it also check if `hint[:exception]` is nil or does it check if it's false?

updater/lib/dependabot/setup.rb
```diff
@@ -37,7 +37,31 @@
...
+        exception.raven_context.each do |key, value|
```
jeffwidman, line 49: I'll need to look into this... at a minimum, I'll probably want to rename this to `sentry_context`

updater/lib/dependabot/setup.rb
```diff
@@ -37,7 +37,31 @@
...
+        end
+      end
+
+    else
+      event
+    end
+  end
+
+  # to send events synchronously like sentry-raven does, set it to 0
+  # config.background_worker_threads = 0
```
jeffwidman, line 60: note for myself: I'm guessing we should send synchronously, both for reduced risk by minimizing change and to ensure a job that's blowing up doesn't fire exception, then exit immediately before the thread pool has a chance to send the exception... I haven't looked into if there's a `flush()` style method, but IMO probably simplest to sidestep any complexity with a synchronous call... esp since the context where we call this is already a background job, no human is waiting on a low-latency response.

updater/spec/dependabot/sentry_spec.rb
```diff
@@ -5,6 +5,7 @@
 
 RSpec.describe ExceptionSanitizer do
   let(:message) { "kaboom" }
+  # TODO migrate this from "data" to "event"/"hint"
```
jeffwidman, line 8: If I can figure out how to get access to the debugger this should be straightforward...

jeffwidman (commented):
This was a quick spike I started on last night for fun, but it's not top priority so annotating the changes in case I don't get back to it for a bit.

updater/lib/dependabot/setup.rb
deivid-rodriguez: It checks whether `hint[:exception]` is either `nil` or `false`, that's usually called "falsy".

updater/lib/dependabot/sentry.rb
```diff
@@ -1,29 +1,31 @@
...
-    data[:exception][:values] = data[:exception][:values].map do |e|
+    event.exception.values = event.exception.values.map do |e|
       PATTERNS.each do |key, regex|
-        next unless (matches = e[:value].scan(regex))
+        next unless (matches = e.value.scan(regex))
 
         matches.flatten.compact.each do |match|
-          e[:value] = e[:value].gsub(match, "[FILTERED_#{key.to_s.upcase}]")
+          # TODO: Discuss with upstream, because e.value is intended to be read-only, so have to hack around it with `instance_variable_set()`
+          e.instance_variable_set(:@value, e.value.gsub(match, "[FILTERED_#{key.to_s.upcase}]"))
```
jeffwidman, line 23: Opened upstream feature / pull requests:
* https://github.com/getsentry/sentry-ruby/issues/2071
* https://github.com/getsentry/sentry-ruby/pull/2072

updater/lib/dependabot/sentry.rb
```diff
@@ -1,29 +1,29 @@
 # frozen_string_literal: true
```
jeffwidman, line 1: Since this file is no longer using `RavenProcessor`,  I should actually further move out the sentry-specific bits to the caller and keep this as a generic exception message processing class.

That will also make testing it a bit easier since I can test solely against exception strings.

This also means the looping through multiple exceptions should get moved to the caller, but I already have to do that anyway because of the other processing we need to do in `before_send`, like applying `raven_context`.

updater/lib/dependabot/sentry.rb
jeffwidman: It's merged, so I updated the PR, but can't resolve this TODO until it lands it an actual release and that release gets pulled into this PR...

jeffwidman:
Closing in favor of @JamieMagee 's great work in #8818 .

My work wasn't entirely in vain, as we're benefiting from https://github.com/getsentry/sentry-ruby/pull/2072 which I discovered while working on this draft PR.

PR #7561, opened 2023-07-13
ignore peer dependencies that are in a grouped update group

jakecoffman:
When updating a dependency, Dependabot considers peer dependency requirements as restrictions on what versions it can update the current dependency to. That makes sense for single dependency updates, but for grouped updates when the peer is also in the group, we shouldn't consider its peer requirements as it's also going to be updated with the rest of the group.

Here's an example. Say we have two dependencies, `A ^1.0` and `B ^1.0`. If B has a peer dependency requirement on A of `A ^1.0` and A 2.0 comes out, Dependabot can't upgrade A to 2.0 because of the peer dependency.

However, if A and B are in a group, then B is most likely also going to be upgraded since they're related. So an A 2.0 will probably have a matching B 2.0.

It **is possible** that B 2.0 is not available though, so the CI tests on the repo would need to catch that, but it seems less likely as often the group is written such that all dependencies are upgradable together.


common/lib/dependabot/update_checkers/base.rb
```diff
@@ -10,12 +10,12 @@ class Base
       attr_reader :dependency, :dependency_files, :repo_contents_path,
                   :credentials, :ignored_versions, :raise_on_ignored,
                   :security_advisories, :requirements_update_strategy,
-                  :options
+                  :group, :options
```
brrygrdn, line 13: Nit: should we prefer `dependency_group` as a hint to the base class? 

I'm conscious that some package managers have their own concept of a group so we might eventually do more do have `dependency_group` look at `native_group` or `package_manager_group` in future.

brrygrdn (commented):
This approach makes sense to me and I agree with the reasoning around the trade off we are making.

The one thing I've having a slightly harder time reasoning out is how this would integrate if we were to group by SemVar - let's say that in order to bump `dependency_a` from 0.3 to 0.4, we need `dependency_b` to be updated from 1.9 to 2.0.

I think this might result in us creating an update that may not resolve until `dependency_b` was updated to 2.0 in a separate PR[^1].

This is probably ok as long as CI would catch that the PR isn't mergeable, I just wonder if it would increase the frequency of this edge-case enough to be a problem 🤔 

[^1]: That may require unlocking and updating `dependency_a` as part of the single PR, which we should do I think

brrygrdn (approved):
👍🏻 on shipping this as a solid improvement for our current lossy NPM groups

deivid-rodriguez:
Even without groups, some ecosystems still have some auto-grouping behavior.

If the latest version is not resolvable by only unlocking the dependency being updated, Dependabot tries to also unlock all other dependencies and then groups automatically the other dependencies that got updated with this trick with the original dependency. This is what the `latest_version_resolvable_with_full_unlock?` and sibling methods do.

Not sure how this plays with group updates, and not sure if it's implemented for NPM at all, but just wanted to mention it since if that plays nice with group updates, it could also be a solution to implement that fallback here and make these cases work regardless of groups.

Anyways, this change seems like a reasonable tradeoff for now 👍.

jakecoffman:
Documenting the thought around [d481863](https://github.com/dependabot/dependabot-core/pull/7561/commits/d481863e5fcabba6c481e7d0f787cbbe9ec63d98).

The code that is removed was copied from the single dependency update, and makes sense in that context. Because Dependabot is smart enough to update multiple dependencies that would have resulted in two PRs.

For grouped updates, kind of the whole point is to group things that are similar. So if peer dependencies span groups, or some in and some out of the group, it's kind of asking for multiple PRs with the same dependency in it? 

If this leads to unexpected behavior we can add it back.

PR #7563, opened 2023-07-14
testing bundler smoke tests

jakecoffman:
I think the caching is not right, testing with this Draft PR.


jakecoffman:
It's good!

PR #7565, opened 2023-07-15
Use the new base image

jeffwidman:
We split the monolith docker image into per-ecosystem images. The `dependabot/dependabot-core` image was replaced by `dependabot/dependabot-updater-core`.

Additionally, we don't publish this new image to Dockerhub, but only to GHCR.

So this corrects the references to it.

This is mostly the great work of @dwc0011 from:
* https://github.com/dependabot/dependabot-core/pull/7458

However, the scope of that PR is starting to get a bit larger, and we'll need to have more conversations about the best way forward. It's unclear if we should even have a default `FROM` value vs always forcing it to be set. But that's a bigger conversation.

So for now I extracted this smaller set of changes out so we can at least stop referencing an old Dockerfile that we no longer update.

### DIFF
--- a/.devcontainer/core-dev/devcontainer.json
+++ b/.devcontainer/core-dev/devcontainer.json
@@ -2,7 +2,7 @@
   "name": "core-dev",
   "build": {
     "dockerfile": "../../Dockerfile.development",
-    "cacheFrom": "ghcr.io/dependabot/dependabot-core-development"
+    "cacheFrom": "ghcr.io/dependabot/dependabot-updater-core"
   },
 
   "workspaceFolder": "/home/dependabot/dependabot-core",
--- a/Dockerfile.development
+++ b/Dockerfile.development
@@ -1,4 +1,4 @@
-ARG FROM_IMAGE=dependabot/dependabot-core
+ARG FROM_IMAGE=ghcr.io/dependabot/dependabot-updater-core
 FROM $FROM_IMAGE
 
 # Temporarily switch to root user in order to install packages


jeffwidman:
The original PR https://github.com/dependabot/dependabot-core/pull/7458 mentions fixing solargraph, rubocop etc.

Those may indeed need to be pulled in, but I wasn't sure and didn't want to go down that rabbit hole just yet. Mostly I want to remove all mentions of the old dockerfile / dockerhub presence. And then from there we can start to figure out next steps (and may in fact have more urgency to do so if tooling gets broken).

deivid-rodriguez (approved):

dwc0011:
> The original PR #7458 mentions fixing solargraph, rubocop etc.
> 
> Those may indeed need to be pulled in, but I wasn't sure and didn't want to go down that rabbit hole just yet. Mostly I want to remove all mentions of the old dockerfile / dockerhub presence. And then from there we can start to figure out next steps (and may in fact have more urgency to do so if some tooling gets broken). To be clear though, most of our tooling would be fine, because our normal scripts set a custom `FROM` value, it's only if someone is using this image from scratch w/o our helper scripts that it'd be broken.

Yeah, I added in the rubocop and solargraph changes because they were totally broken and have been for some time.  I should have pulled them into another PR or changed the title of this one, to make devcontainer work again, but I was focused on getting a clean devcontainer set up and those changes were made as a result of that.

jeffwidman:
@dwc0011 nw at all, and thanks for finding this issue.

I'm 🚢 this so you can rebase #7458 and keep it moving forward.

dwc0011:
> @dwc0011 nw at all, and thanks for finding this issue.
> 
> I'm 🚢 this so you can rebase #7458 and keep it moving forward.

Sorry I went back and reviewed what I was doing and remembered why the updater-core would not work.  Which is why I made a placeholder for an ecosystem build in the .devcontainer and removed the default image in the Dockerfile

dwc0011:
In my PR there are two lines in the devcontainer that address SolarGraph ("solargraph.useBundler": false,) and Rubocop ("editor.defaultFormatter": "LoranKloeze.ruby-rubocop-revived") .  

Everything else is related to the devcontainer image.  The questions I had were related to is there a manifest that says what container image containes what version of the helper tool, ie how to know which terraform ecosystem build contains 1.4.x vs 1.5.x etc etc.  So I could update the instructions to help people chose the proper ecosystem build for development.

Developers will either need to provide an ecosystem build or if  they are to use the updater-core image and then in the Dockerfile.development it should install all the helper tools.  Another option is to support a developer image again that has all the helper tools installed.

Not sure what the preference is as this ecosystem build is pretty new.

PR #7566, opened 2023-07-15
Document how to `require` the `debugger`

jeffwidman:
Realized our readme assumed some ruby foo that not all folks know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,8 +274,10 @@ Here's an example of how to string all these together
 You can add a `debugger` statement anywhere in the ruby code, for example:
 
 ```ruby
+      require 'debug' # at the top of the file
+
       def latest_resolvable_version
-        debugger
+        debugger # within the actual code
         latest_version_finder.latest_version
       end
 ```


README.md
```diff
@@ -274,8 +274,10 @@ Here's an example of how to string all these together
 You can add a `debugger` statement anywhere in the ruby code, for example:
 
 ```ruby
+      require 'debug' # at the top of the file
```
jurre, line 277: Do we need to require it manually? We already require it in the dry-run script at the top-level: https://github.com/dependabot/dependabot-core/blob/2530fd02738ddfd6b6251a736953c18946f7ec71/bin/dry-run.rb#L73

jeffwidman: Hmm... Good point, and CLI also sets it via `--debug` IIRC...

I'm trying to remember how I got into the situation where I needed to do it manually... pretty sure it's because I was running `updater/bin/fetch_files.rb`, which we don't expect open source collaborators to ever touch.

jeffwidman:
For now we can probably skip this since as @jurre points out it's not needed for the normal case. Can always revisit later if it comes up again.

README.md
Nishnha: I've had to add this to my code while writing unit tests.

PR #7567, opened 2023-07-16
Mount `updater/bin` into the docker dev shell

jeffwidman:
I found it handy to have access to these while debugging some `updater` code.


jakecoffman (approved):
The scripts might not work since there's no API running, but no harm in mounting them if you find value in it.

jeffwidman:
Yeah, they don't work, but I was investigating an issue in the updater setup, before it even hits the API. I just needed the files present so I could call them within the docker dev shell.

PR #7580, opened 2023-07-18
fix flaky test that fails when new versions are released

jakecoffman:
To permanently fix this, I've switch from matching a fixture as the result to using a regex to decide if the beginning of the line starts with the expected non-standard prefix. The content of the update isn't what is under test here, so I think this is good.


jurre (approved):

npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater_spec.rb
```diff
@@ -1416,9 +1416,8 @@
 
       context "with non-standard indentation" do
         it "preserves indentation in the package-lock.json" do
-          expect(updated_npm_lock.content).to eq(
-            fixture("updated_projects", "npm6", "simple", "package-lock.json")
-          )
+          # Ensure lines start with { or } or tabs. The actual content of the update isn't under test here.
+          expect(updated_npm_lock.content.split("\n")).to all(match(/^{|}|\t+["}]/))
```
deivid-rodriguez, line 1420: If we `JSON.parse` a package.json file and then `JSON.dump` back, I think we end up with a single line of JSON that would satisfy this assertion. Should we verify that too?

deivid-rodriguez (commented):
Nice! Just added a minor comment to potentially improve the new assertion.

npm_and_yarn/spec/dependabot/npm_and_yarn/file_updater_spec.rb
jakecoffman: Is the concern that if the updated_npm_lock.content is accidentally turned into a single line that the regex won't catch it?

jakecoffman: OK I can put another PR up with an assertion that the lockfile is more than a few lines to check for that case. Thanks for mentioning it!

deivid-rodriguez: I just realized that this is checking the `package-lock.json` content, not the `package.json` content as I initially thought.

Looking at the [PR](https://github.com/dependabot/dependabot-core/pull/1125) that introduced this, the new test seems still not enough to cover it, though?

We could alternatively check that no leading whitespace changes are introduced on any lines. In pseudocode, check that `(updated_lines - original_lines).size == (updated_lines_with_leading_whitespace_stripped - original_lines_with_leading_whitespace_stripped).size`.

jakecoffman: I'm not sure how the regex doesn't cover it, can you explain more how it doesn't? 

The way I figured it, we happen to know that we're using tabs in the fixture, so the only possibilities for the updated lines are they start with either:
- `{` for line 1
- `}` for the last line
- Tabs followed by a `"` or `}`

deivid-rodriguez: Sorry @jakecoffman, I misunderstood what you were doing here and thought you were expecting with `{`, `}`, or generic whitespace. I didn't notice you were explicitly checking for tabs.

This is definitely good as is, sorry for the distraction!

jakecoffman: No need for apology! Always good to talk through things. 😃

PR #7581, opened 2023-07-18
[Grouped Updates] Implement experimental Semantic Versioning rule

brrygrdn:
This branch introduces a new experiment flag: `grouped_updates_experimental_rules`

With this experiment enabled the follow changes happen by default:
- Dependency Groups **will no longer include Major version upgrades**
- Any Major upgrades available will be opened in a separate PR.

It also allows a new configuration rule:
```yaml
# Custom SemVer groups
groups:
  patches:
    highest-semver-allowed: "patch"
  small-upgrades:
    highest-semver-allowed: "minor"
  big-upgrades:
    highest-semver-allowed: "major"
```

***These rules are not final syntax***, we still need to do a naming pass and also resolve how existing rules interplay with them so this may churn.

### Approach

The implementation of update-type filtering is fairly trivial, `Dependabot::Dependency` objects implement a `production?` method so we do this pretty much the same way we filter on dependency name.

The trickier case is Semantic Versioning which we do by **injecting ignore conditions** for any versions of a Dependency outside the highest patch level allowed.

This strategy means the group does the right thing with very few snags - the tricky part is that assuming that the group(s) only cover up to `minor` patch levels we will then want to provide PRs for any `major` versions available separately.

For this spike, I've only gone so far into the logic required to do this efficiently so the heuristic is:
> If any dependencies are in **only** groups with `highest-semvar-allowed` below `major`, attempt a separate update

This is a good enough strategy for now, but for large projects it would be crushingly slow as we'd currently call out to the package manager twice for everything when we already know the versions available.

This spike has already gotten pretty long so I'd rather put a cut line here as I'd like to test this in some real-world conditions on projects where we can control the size/tolerate the added runtime to validate the work so far.

### Next Steps

The two main things that need to happen with this approach are:
- Look at the global ignore conditions to determine the maximum semver the job permits for `*` dependencies and start using that as the cap on whether we try a separate update
- Start comparing the `latest_version` and `latest_resolvable_version` to determine if there are actually any upgrades ahead of the group to further avoid double work
- Look at how we can start implementing caching of the version lists we look up, this will likely need some per-ecosystem work


updater/lib/dependabot/dependency_group_engine.rb
```diff
@@ -56,6 +56,10 @@ def assign_to_groups!(dependencies:)
       @groups_calculated = true
     end
 
+    def ungrouped_dependencies
+      @ungrouped_dependencies + dependencies_with_ungrouped_semvar_levels
```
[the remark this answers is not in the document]
brrygrdn: Geez, yeah, I thought I caught the places I'd done that - thanks for spotting this

jurre:
> Dependency Groups will not longer include Major version upgrades

@brrygrdn and I had an interesting chat about this, I get the decision for this default, but I think it would be interesting to explore a way we can make this config more explicit, I could imagine it being a confusing experience otherwise. No need to block on that while this is behind a feature flag though

common/lib/dependabot/dependency_group.rb
```diff
@@ -1,25 +1,53 @@
...
+      end
+    end
+
     attr_reader :name, :rules, :dependencies
 
     def initialize(name:, rules:)
       @name = name
       @rules = rules
       @dependencies = []
+      @ignore_condition = generate_ignore_condition!
```
jurre, line 32: Pedantic, but maybe we can make the naming explicitly say what these ignore conditions are for?

common/lib/dependabot/dependency_group.rb
```diff
@@ -32,5 +60,81 @@ def to_config_yaml
...
+
+      Dependabot.logger.debug("The #{name} group has set ignores for update-type(s): #{ignored_update_types}")
+
+      Dependabot::Config::IgnoreCondition.new(
+        dependency_name: ANY_DEPENDENCY_NAME,
+        update_types: ignored_update_types
+      )
+    end
+
+    def ignored_update_types_for_rules
```
jurre, line 117: Do we also foresee a need for folks to configure only updates for minor or major version bumps? Right now you'd always get everything _up to_ the specified range. It's a good starting point, but have we thought about the other scenario's?

common/lib/dependabot/config/ignore_condition.rb
```diff
@@ -4,8 +4,8 @@ module Dependabot
   module Config
     # Filters versions that should not be considered for dependency updates
     class IgnoreCondition
-      PATCH_VERSION_TYPE = "version-update:semver-patch"
-      MINOR_VERSION_TYPE = "version-update:semver-minor"
+      PATCH_VERSION_TYPE = "version-update:semver-patch",
+      MINOR_VERSION_TYPE = "version-update:semver-minor",
       MAJOR_VERSION_TYPE = "version-update:semver-major"
```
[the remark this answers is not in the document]
brrygrdn: Yeah, my bad, I only partially walked back a change to this - will fix 🤦🏻

common/lib/dependabot/dependency_group.rb
brrygrdn: I originally started out with this, but this can be achieved using ignores, if the group had a highest semver value of `major`, you could add:

```
ignore:
  - dependency-name: *
     update-types: ["version-updates:semver-patch"]
```

The semantics of allowing something like "I only want patch-and-major updates in this group" is hard to reason about if we make it too broad, so I'd ended up leaning on our existing configuration options to keep the group semantics simple.

brrygrdn: I've been thinking about this more and implementing as a list might end up being the right answer if someone wants the following:

```
groups
  patches:
    update-types: ["patches"]
  small-bumps:
    update-types: ["minor"]
```

With the `highest-semver-allowed` approach, this would _almost_ work, but it would be surprising if you only had patch versions available for one or more dependencies - you end up with the `patches` and `small-bumps` PRs containing both containing the patch version.

jakecoffman (approved):

PR #7590, opened 2023-07-19
Cancel previous jobs in the same branch when repushing

deivid-rodriguez:
Sometimes when working on fixing CI you may need to push several times. For these cases, I think it makes sense to automatically cancel previous runs in the same branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,10 @@ on:
   schedule:
     - cron: "0 0 * * *"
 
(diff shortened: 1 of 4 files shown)


jurre (approved):

deivid-rodriguez:
Let's try it!

jeffwidman:
Nice! Didn't realize this was even an option. I shudder to think how many CI minutes I've accidentally wasted with repeated force-pushes tweaking things.

PR #7591, opened 2023-07-19
Don't try to diff if there's nothing to diff

deivid-rodriguez:
On a passing smoke test workflow, you still get some Red Cross annotations.

They don't make the build fail, but can be distracting.

Trying to not run them when not necessary.


deivid-rodriguez:
Just tried this and it works!

jakecoffman (approved):
This has bothered me for a while, thanks for taking care of it!

PR #7594, opened 2023-07-19
Aggressively group prod and dev dependencies for NPM

abdulapopoola:
Use wildcard matching to aggregate prod and dev updates into two separate groups

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -53,12 +53,14 @@ updates:
     groups:
       dependencies:
         patterns:
-          - "*yarn*"
-          - "*arborist*"
-          - "*nock*"
-          - "*npm*"
-          - "*semver*"
+          - "@dependabot/yarn-lib"
+          - "@npmcli/arborist"
           - "detect-indent"
+          - "nock"
+          - "npm"
+          - "@pnpm/lockfile-file"
+          - "@pnpm/dependency-path"
+          - "semver"
       dev-dependencies:
         patterns:
           - "*eslint*"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -50,6 +50,20 @@ updates:
       interval: "weekly"
       day: "sunday"
       time: "16:00"
+    groups:
+      dependencies:
+        patterns:
+          - "*yarn*"
+          - "*arborist*"
+          - "*nock*"
+          - "*npm*"
+          - "*semver*"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -50,6 +50,20 @@ updates:
...
       time: "16:00"
+    groups:
+      dependencies:
+        patterns:
+          - "*yarn*"
+          - "*arborist*"
+          - "*nock*"
+          - "*npm*"
+          - "*semver*"
+          - "detect-indent"
```
landongrindheim, line 61: 🔎 I don't know how likely it is that new dependencies will be introduced and catch us by surprise, but I think I'd be more comfortable with the patterns being the [list of dependencies](https://github.com/dependabot/dependabot-core/blob/cd0af0cf66dc9488c30bf07203cf644e6c271e1d/npm_and_yarn/helpers/package.json#L12-L19) as they curently exist (i.e. hard-coding them). That would make the kind of errors that _I'm_ susceptible to harder to introduce 😅  What do you think of that?

If that's something you're agreeable to, I think 👇 would do the trick 😄 
```suggestion
          - "@dependabot/yarn-lib"
          - "@npmcli/arborist"
          - "detect-indent"
          - "nock"
          - "npm"
          - "@pnpm/lockfile-file"
          - "@pnpm/dependency-path"
          - "semver"
```

landongrindheim (approved):

.github/dependabot.yml
abdulapopoola: Good suggestion; agree with being explicit over implicit. Thank you!

.github/dependabot.yml
```diff
@@ -50,6 +50,22 @@ updates:
...
+          - "@npmcli/arborist"
+          - "detect-indent"
+          - "nock"
+          - "npm"
+          - "@pnpm/lockfile-file"
+          - "@pnpm/dependency-path"
+          - "semver"
+      dev-dependencies:
+        patterns:
+          - "*eslint*"
```
abdulapopoola, line 66: @landongrindheim would you want the same 'explicit' configuration to apply to dev-dependencies too? I can make the update too.

landongrindheim: I'm less concerned with the dev dependencies 😄

deivid-rodriguez (approved):
Happy to try it!

PR #7609, opened 2023-07-22
Point at the correct `Dockerfile` for python

jeffwidman:
We broke the monolithic docker image into per-language images. So point the link at the Python one.

Also update to a PR that bumped python using this new dockerfile so that we have a better example. It's not perfect as it doesn't bump the `pyenv global` var, but it's the best we've got right now. We can replace it with the [`3.12` PR in October](https://peps.python.org/pep-0693/).

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -7,23 +7,25 @@ Python support for [`dependabot-core`][core-repo].
 We rely on `pyenv` to manage Python's versions.
 
 Updating the list of known versions might be tricky, here are the steps:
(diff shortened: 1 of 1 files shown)


jeffwidman:
Let's first land:
* https://github.com/dependabot/dependabot-core/pull/7611

As that will simplify the readme slightly by removing one of the steps completely.

deivid-rodriguez (approved):

Issue #6299, opened 2022-12-09
Drop support for Python 3.6


jeffwidman:
Python 3.6 was [EOL'd in Dec 2021](https://endoflife.date/python). However, we've continued to support it because the Ubuntu LTS `18.04` release installed it by default and `18.04` isn't EOL until April 2023:
https://github.com/dependabot/dependabot-core/blob/cf448a99d35d7956f4c64b8249029324c9627a91/python/lib/dependabot/python/python_versions.rb#L18-L19

But we're starting to run into various issues where `3.6` is forcing us to include hacky workarounds:
* latest versions of most of the python package managers break on `3.6`, including `poetry`, `pip`, `pipenv` and `pip-tools`.
* [Bumping to `22.04` is blocked because `3.6` conflicts with its `openssl` version](https://github.com/dependabot/dependabot-core/pull/5030)

So this is a placeholder ticket to track any discussions around timing, things that we'll need to update/remove, etc.

If you're a user of Dependabot who is still on Python 3.6 and you cannot upgrade for some reason, please chime in here explaining your use case.

jeffwidman:
Before we commit to a timeline on this, we'd like to measure how frequently users are running Dependabot against `python 3.6` repos.

However, the new ["pause for inactivity" feature](https://github.blog/changelog/2023-01-12-dependabot-pull-requests-pause-for-inactivity/) is still gradually rolling out... and that will impact these metrics pretty drastically because abandoned repos that are still running Dependabot are the most likely to have never moved beyond `3.6`.

We expect that rollout to finish in the next few weeks, after which we plan to gather metrics and then work with some other internal stakeholders to decide on the actual date we'll drop support.

edgarrmondragon:
Are there any updates on this? I think this is blocking support for Poetry 1.4.x

grski:
This is a necessary change for https://github.com/dependabot/dependabot-core/pull/6758 AFAIK, or updating Poetry to 1.4.x 
ANy update here or how can one help to make it move faster?

jeffwidman:
Thanks for letting us know you want it, we've been a little hesitant to drop it because we know some users are probably still on `3.6` and we haven't yet had time to pull metrics to see how many users will get left behind when we drop it... so knowing we've got users that are blocked on us upgrading is good to know as a counterpoint.

[...]

PR #7610, opened 2023-07-22
Drop `python` `3.6`

jeffwidman:
This drops `python` `3.6` which has been EOL'd since Dec 2021.

This does _not_ tackle any of the follow-on work that dropping `3.6` unlocks... ie, bumping to newer versions of `pip-tools`, `pip`, `poetry` etc. 

This cannot be merged until August 17th: https://github.blog/changelog/2023-08-02-deprecation-notice-dependabot-to-drop-support-for-python-3-6-and-3-7/

Fix: https://github.com/dependabot/dependabot-core/issues/6299


deivid-rodriguez (approved):
Seems straightforward!

abdulapopoola (approved):

PR #7611, opened 2023-07-22
Delete leftover build script

jeffwidman:
This script is a leftover from when we had a monolithic image for all ecosystems.

Now that we use a specific docker image just for python, that dockerfile uses the parameterized version of this script, called `build_for_version`.

You can see more historical context in this commit: https://github.com/dependabot/dependabot-core/commit/3e999356cbaae522f1cec7dfae88a45a1e480d98

I grep'd around and couldn't find any remaining usages of the `build` script, so let's nuke it.


deivid-rodriguez (commented):
If you change a dependency and then manually run the build script, will you want all native helpers reinstalled for tests to pass? I guess that's why the previous `build` script installed for all pythons?

jeffwidman:
>  I guess that's why the previous build script installed for all pythons?

I don't think so... [this commit](https://github.com/dependabot/dependabot-core/commit/3e999356cbaae522f1cec7dfae88a45a1e480d98) explains that the `build` script was left-in for the monolithic image... probably it was _not_ removed when the multi-stage build was added in order to reduce risk during initial rollout, and only later once everything was proven stable was it removed but removing this script was overlooked.

> If you change a dependency and then manually run the build script, will you want all native helpers reinstalled for tests to pass?

Technically that's true and this would be a convenience, but TBH it's a pain to have to keep the list of python versions updated here... Historically the frequency that we need to re-run the build script for testing seems like it's smaller than the frequency of bumping python releases.

It also feels dangerous to have a `build` script that someone might use during dev flow that's different than the `build-for-helpers` script that is actually used in prod.

pavera (approved):

deivid-rodriguez (approved):

PR #7613, opened 2023-07-22
Stop manually installing python

jeffwidman:
Depends on:

In the python dockerfile, we pre-install python 3.7-3.11, then `tar` it
to save space.

Now that we're not supporting python <= 3.6, we no longer have any
reason to run `pyenv install`, so we can dop that and simplify our code.

### DIFF
--- a/python/lib/dependabot/python/language_version_manager.rb
+++ b/python/lib/dependabot/python/language_version_manager.rb
@@ -14,22 +14,9 @@ module Dependabot
         # The leading space is important in the version check
         return if SharedHelpers.run_shell_command("pyenv versions").include?(" #{python_major_minor}.")
 
-        if File.exist?("/usr/local/.pyenv/#{python_major_minor}.tar.gz")
-          SharedHelpers.run_shell_command(
-            "tar xzf /usr/local/.pyenv/#{python_major_minor}.tar.gz -C /usr/local/.pyenv/"
-          )
-          return if SharedHelpers.run_shell_command("pyenv versions").
-                    include?(" #{python_major_minor}.")
-        end
-
-        Dependabot.logger.info("Installing required Python #{python_version}.")
-        start = Time.now
-        SharedHelpers.run_shell_command("pyenv install -s #{python_version}")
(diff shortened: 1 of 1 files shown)


python/lib/dependabot/python/language_version_manager.rb
```diff
@@ -14,22 +14,9 @@ def install_required_python
...
-        start = Time.now
-        SharedHelpers.run_shell_command("pyenv install -s #{python_version}")
-        SharedHelpers.run_shell_command("pyenv exec pip install --upgrade pip")
-        SharedHelpers.run_shell_command("pyenv exec pip install -r" \
-                                        "#{NativeHelpers.python_requirements_path}")
-        time_taken = Time.now - start
-        Dependabot.logger.info("Installing Python #{python_version} took #{time_taken}s.")
+        SharedHelpers.run_shell_command(
+          "tar xzf /usr/local/.pyenv/#{python_major_minor}.tar.gz -C /usr/local/.pyenv/"
+        )
```
deivid-rodriguez, line 19: What happens if we find a python version _higher_ than the latest that we support, for example, right after a new Python release.

deivid-rodriguez (approved):
Seems straightforward but I did have a question!

python/lib/dependabot/python/language_version_manager.rb
jeffwidman: Great question!

The behavior should remain the same as today:
1. Newer patch releases are ignored and we use the matching `major.minor` pre-installed, until someone bumps the patch release we use.
2. Newer minor versions (eg `3.12` that is getting released in a few months) will throw an error about an unsupported Python version.

This is a reasonable case where we'd want to manually install, but OTOH the time/network cost of doing that dynamically on every job that needs it is quite expensive. So far, it's worked out okay that whenever a new minor release happens, someone from the community pretty quickly puts up an upgrade PR. Ideally we actually start scheduling this into our ecosystem calendar so that we're ready with a PR the day it gets released. But right now we don't have the maintainer bandwidth to stay on top of that, so I think we're okay for now.

Issue #5809, opened 2022-09-28
Updater failing with "Setting experimental.system-git-client does not exist"


diguid:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

pip

### Package manager version

Poetry 1.2.1

### Language version

Python 3.7.5

### Manifest location and content before the Dependabot update

/poetry.lock

### dependabot.yml content

```
version: 2
registries:
  python-index-pypi-hub-legiti-io-simple:
    type: python-index
    url: https://pypi.hub.legiti.io/simple/
    username: "${{secrets.PYTHON_INDEX_PYPI_HUB_LEGITI_IO_SIMPLE_USERNAME}}"
    password: "${{secrets.PYTHON_INDEX_PYPI_HUB_LEGITI_IO_SIMPLE_PASSWORD}}"

updates:
- package-ecosystem: pip
  directory: "/"
  insecure-external-code-execution: allow
  schedule:
    interval: daily
    time: "18:00"  # UTC
  open-pull-requests-limit: 10
  target-branch: develop
  registries:
  - python-index-pypi-hub-legiti-io-simple
```

### What you expected to see, versus what you actually saw

Expected: PRs with updated versions

Actual: Errors in the dependabot run.

```
updater | ERROR <job_471376381> Error processing pytest-cov (Dependabot::SharedHelpers::HelperSubprocessFailed)
updater | ERROR <job_471376381> ValueError
updater | <job_471376381> 
updater | <job_471376381>   Setting experimental.system-git-client does not exist
updater | <job_471376381> 
updater | <job_471376381>   at /usr/local/.pyenv/versions/3.7.13/lib/python3.7/site-packages/poetry/console/commands/config.py:248 in handle
updater | <job_471376381>       244│                 raise ValueError("You must pass exactly 1 value")
updater | <job_471376381>       245│ 
updater | <job_471376381>       246│             return 0
updater | <job_471376381>       247│ 
updater | <job_471376381>     → 248│         raise ValueError("Setting {} does not exist".format(self.argument("key")))
updater | <job_471376381>       249│ 
updater | <job_471376381>       250│     def _handle_single_value(self, source, key, callbacks, values):
updater | <job_471376381>       251│         validator, normalizer, _ = callbacks
updater | <job_471376381>       252│
updater | ERROR <job_471376381> /home/dependabot/python/lib/dependabot/python/update_checker/poetry_version_resolver.rb:353:in `run_poetry_command'
```

[...]

PR #7614, opened 2023-07-22
Remove `3.6` guard when setting `poetry config experimental.system-git-client`

jeffwidman:
Depends on:

We no longer support Python 3.6, so this now always needs to happen.

I also bumped the minimum version of `poetry` up to `1.2` as that's when
the flag was added. It shouldn't matter, as python will default to the
highest version allowed by that version of python, and now that we only
support `3.7` onwards that will be the upper bound.


deivid-rodriguez (approved):
Cool! This "fixes" https://github.com/dependabot/dependabot-core/issues/5809.

jeffwidman:
Good memory! I'd completely forgotten about that issue/PR.

PR #7615, opened 2023-07-22
Stop explicitly specifying python patch versions

jeffwidman:
Depends on:
 
Now that we've dropped Python 3.6 support, we no longer need to
explicitly list all python versions down to the `patch` level because
the following are true:

1. All python versions that Dependabot supports have a matching
   major.minor python version that is already pre-installed in the
   Dockerfile.
2. We completely ignore the patch level when matching the user's desired
   python with the python versions we support.

So we only need to track the explicit version that is
pre-downloaded/installed to the Dockerfile.


python/lib/dependabot/python/language_version_manager.rb
```diff
@@ -33,8 +42,12 @@ def install_required_python
...
 
       def python_major_minor
-        @python ||= Python::Version.new(python_version)
-        "#{@python.segments[0]}.#{@python.segments[1]}"
+        @python ||= extract_python_major_minor(python_version)
+      end
+
+      def self.extract_python_major_minor(python_version_string)
+        full_version = Python::Version.new(python_version_string)
+        "#{@full_version.segments[0]}.#{@full_version.segments[1]}"
```
jeffwidman, line 50: I extracted this to a class method as a helper as there are multiple places we could use this... but Rubocop is complaining about it. I'm not fully understanding why... or rather, the bits that I do understand after reading up on it leave me a little frustrated...

This may require refactoring first such that `@python` is memoized separately outside of this method. 🤔

jeffwidman: https://github.com/dependabot/dependabot-core/pull/7703

python/lib/dependabot/python/file_updater/poetry_file_updater.rb
```diff
@@ -7,7 +7,6 @@
 require "dependabot/python/language_version_manager"
 require "dependabot/python/version"
 require "dependabot/python/requirement"
-require "dependabot/python/python_versions"
```
jeffwidman, line 10: These `require`'s are probably dead on `main` leftover from when `_pavera` did that super useful refactoring to DRY the python version selection logic into a single class.

deivid-rodriguez (approved):

PR #7617, opened 2023-07-22
Instrument Python version metric

jeffwidman:
We want to know what versions of python are required by our users, so this is an initial stab at collecting some data.

Ideally we capture the minimum version, the maximum version, and the raw data from each manifest file that allowed us to calculate the min/max... this won't be quite so clean/polished, but we need to start somewhere and even a primitive form of this metric will provide enough information for us to understand the user impact of dropping python 3.6.

### DIFF
--- a/python/lib/dependabot/python/file_fetcher.rb
+++ b/python/lib/dependabot/python/file_fetcher.rb
@@ -37,6 +37,27 @@ module Dependabot
           "or a Pipfile."
       end
 
+      def ecosystem_versions
+        # Hmm... it's weird that this calls file parser methods, but here we are in the file fetcher... for all
+        # ecosystems our goal is to extract the user specified versions, so we'll need to do file parsing... so should
+        # we move this `ecosystem_versions` metrics method to run in the file parser for all ecosystems? Downside is if
+        # file parsing blows up, this metric isn't emitted, but reality is we have to parse anyway... as we want to know
+        # the user-specified range of versions, not the version Dependabot chose to run.
+        python_requirement_parser = FileParser::PythonRequirementParser.new(dependency_files: files)
+        language_version_manager = LanguageVersionManager.new(python_requirement_parser: python_requirement_parser)
+        {
+          languages: {
+            "python": {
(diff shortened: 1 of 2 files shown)


pavera (commented):
This should do the trick at least initially. I think the most interesting question this raises is how many repos actually have multiple python versions defined in different manifest files?

pavera (approved):

python/lib/dependabot/python/file_fetcher.rb
```diff
@@ -37,6 +37,28 @@ def self.required_files_message
           "or a Pipfile."
       end
 
+      def ecosystem_versions
+        # Hmm... it's weird that this calls file parser methods, but here we are in the file fetcher... for all
+        # ecosystems our goal is to extract the user specified versions, so we'll need to do file parsing... so should
+        # we move this `ecosystem_versions` metrics method to run in the file parser for all ecosystems? Downside is if
+        # file parsing blows up, this metric isn't emitted, but reality is we have to parse anyway... as we want to know
+        # the user-specified range of versions, not the version Dependabot chose to run.
```
jeffwidman, line 45: I'm planning to have follow-on PR's that move this to the `FileParser` class, but that refactor will require more work, and want this live now so it starts collecting metrics, so I'm going to merge as-is and then clean it up.

python/lib/dependabot/python/file_fetcher.rb
```diff
@@ -37,6 +37,28 @@ def self.required_files_message
...
+        python_requirement_parser = FileParser::PythonRequirementParser.new(dependency_files: files)
+        language_version_manager = LanguageVersionManager.new(python_requirement_parser: python_requirement_parser)
+        {
+          languages: {
+            python: {
+              # TODO: alternatively this could use `python_requirement_parser.user_specified_requirements` which
+              # returns an array... which we could flip to return a hash of manifest name => version
+              # string and then check for min/max versions... today it simply defaults to
+              # array.first which seems rather arbitrary.
```
jeffwidman, line 54: `user_specified_requirements` is a private method, so the best way to do that is move this to `FileParser`, but again, we want metrics soon, so ship this first then clean that up.

PR #7621, opened 2023-07-24
fix Gemfile not updating for peer dependencies

jakecoffman:
This fixes an issue in Bundler updates where if you have a peer dependency on the dependency being updated, the peer gets updated in the lockfile, but not the Gemfile.

The problem can be reproduced with this Gemfile:

```ruby
# frozen_string_literal: true

source "https://rubygems.org"

git_source(:github) {|repo_name| "https://github.com/#{repo_name}" }

gem "octicons", "~> 19.2"
gem "octicons_helper", "~> 19.2"
```

With this PR, `octicons` and `octicons_helper` are correctly updated together in the Gemfile. I checked against [this branch](https://github.com/dependabot/smoke-tests/compare/brrygrdn/octicon-bug) of the smoke tests.

I'm not as knowledgeable about Bundler, so let me know if something I did here doesn't make sense!


jakecoffman:
Oh I will add this case as a fixture.

jurre (approved):
That makes sense to me 👍

deivid-rodriguez (approved):
I was wondering if it'd make more sense to fix this inside the native helper? I think we have more direct access to direct dependencies in there.

That said, anything that fixes this and passes all tests seems good!

jakecoffman:
Ok if this works we can circle back and see if it makes more sense in the native helper. Personally I don't like developing in the helpers, it's a bit of a pain to debug them!

PR #7634, opened 2023-07-25
Tweak `ecosystem_versions` schema to be consistent

jeffwidman:
We want to know three things for every ecosystem component:

1. The raw version string and the name of the manifest file that it was pulled from. This provides great fidelity, especially if there are multiple sources of versions (looking at you Python with `.python-version` + `setup.py` etc in the same repo)
2. "rounded" / summarized versions of the above that are easy to write comparison queries against. Ie, `max`/`min` supported versions that are rounded to decimals rather than semver so that in a SQL-like query language we can easily write comparison queries.

So this starts the process of transforming our existing metrics to this format.

### DIFF
--- a/bundler/lib/dependabot/bundler/file_fetcher.rb
+++ b/bundler/lib/dependabot/bundler/file_fetcher.rb
@@ -26,7 +26,11 @@ module Dependabot
       def ecosystem_versions
         {
           package_managers: {
-            "bundler" => Helpers.detected_bundler_version(lockfile)
+            bundler: {
+              # TODO: Today this method returns only the major version, but long term it'll return the full string, at
+              # which point we'll need to move that to "raw" key and then coerce it to major.minor for "max"/"min".
+              "max" => Helpers.detected_bundler_version(lockfile)
+            }
           }
         }
       end
--- a/cargo/lib/dependabot/cargo/file_fetcher.rb
+++ b/cargo/lib/dependabot/cargo/file_fetcher.rb
@@ -29,7 +29,15 @@ module Dependabot
 
         {
           package_managers: {
-            "cargo" => channel
+            cargo: {
+              # ecosystem_versions data gets turned into metrics dashboards. We want those dashboard queries to be
+              # generic across ecosystems and other ecosystems allow specifying a range of supported versions, so
+              # specify max/min even though `Cargo.toml` only specifies a single version, ie max == min.
+              #
+              # TODO: need to round the channel from raw version to `major.minor` for easier grouping in dashboards
+              "max" => channel,
+              "min" => channel
+            }
           }
         }
       rescue TomlRB::ParseError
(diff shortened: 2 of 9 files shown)


cargo/lib/dependabot/cargo/file_fetcher.rb
```diff
@@ -29,7 +29,13 @@ def ecosystem_versions
 
         {
           package_managers: {
-            "cargo" => channel
+            cargo: {
+              # ecosystem_versions data gets turned into metrics dashboards. We want those dashboard queries to be
+              # generic across ecosystems and other ecosystems allow specifying a range of supported versions, so
+              # specify max/min even though `Cargo.toml` only specifies a single version, ie max == min.
+              "max" => channel
```
jeffwidman, line 36: Need to round the channel from semver to a decimal for both min/max...

composer/lib/dependabot/composer/file_fetcher.rb
```diff
@@ -20,7 +20,9 @@ def self.required_files_message
       def ecosystem_versions
         {
           package_managers: {
-            "composer" => Helpers.composer_version(parsed_composer_json, parsed_lockfile) || "unknown"
+            composer: {
+              "max" => Helpers.composer_version(parsed_composer_json, parsed_lockfile)
```
jeffwidman, line 24: Today this method returns only the major version, but long term it'll return the full string, so we'll need to coerce it then from semver to a decimal of major.minor

go_modules/lib/dependabot/go_modules/file_fetcher.rb
```diff
@@ -16,10 +16,16 @@ def self.required_files_message
 
       def ecosystem_versions
         return nil unless go_mod
-
+        go_version_string = go_mod.content.match(/^go\s(\d+\.\d+)/)&.captures&.first || "unknown"
```
jeffwidman, line 19: No need to coerce as this is already only major.minor, patch isn't supported in `go.mod` files.

bundler/lib/dependabot/bundler/file_fetcher.rb
```diff
@@ -26,7 +26,9 @@ def self.required_files_message
       def ecosystem_versions
         {
           package_managers: {
-            "bundler" => Helpers.detected_bundler_version(lockfile)
+            bundler: {
+              "max" => Helpers.detected_bundler_version(lockfile)
```
jeffwidman, line 30: Today this only returns the major version, but long term this should be coerced as well to major.minor... given that long term once we're only running Bundler 2 we'll be returning the full `bundler` version extracted from the `BUNDLED_WITH` statement.

jeffwidman:
I'm no longer on the :dependabot: team and this PR changes metrics, which requires accompanying changes on the Dependabot internal back-end, so not something I can move forward with on my own.

I'll leave it up to the current team to decide whether to proceed with this or close it out.

jeffwidman:
@kbukum1 did a bunch of work to update the metrics, so I suspect this is no longer relevant.

PR #7638, opened 2023-07-26
Implement security update support for Swift

deivid-rodriguez:
For now the logic is super simple.

Pick the lowest fixed version. If it works for the project, then use that. Otherwise fail.

While testing resolvability, I tried to create a real situation where the security update would fail.

It's in this repo: https://github.com/dsp-testing/swift-security-updates-2. It has the following manifest:

```swift
// swift-tools-version:5.2
// The swift-tools-version declares the minimum version of Swift required to build this package.
import PackageDescription

let package = Package(
    name: "Foo",
    platforms: [
        .macOS(.v10_13), .iOS(.v11), .tvOS(.v11), .watchOS(.v4)
    ],
    dependencies: [
        .package(url: "https://github.com/vapor/vapor.git", from: "3.0.0"),
        .package(url: "https://github.com/dependabot-fixtures/swift-example-vapor3", from: "1.0.0"),
    ],
    swiftLanguageVersions: [.v5]
)
```

I also created https://github.com/dependabot-fixtures/swift-example-vapor3, which has a single v1.0.0 tag and depends on Vapor 3.

Vapor 3 is vulnerable to a handful of vulnerabilities, so Dependabot and Swift will try to update it to Vapor 4. My expectation is that they would fail to do so since the package at https://github.com/dependabot-fixtures/swift-example-vapor3 only supports Vapor 3.

However, it updates to Vapor 4 happily.

So, there's something I'm missing about how SwiftPM resolves dependencies.

The issue above is why I did not add a test for the "failed to resolve the lowest fixed version" case, because I couldn't find how to reproduce it.


deivid-rodriguez:
Summoning @stevapple for SwiftPM knowledge. My expectation is that in the above situation SwiftPM would never resolve to `swift-example-vapor3` at `v1` and `vapor` at `v4`. In other words, I'd expect SwiftPM to fail in that case. Yet it seems to be working just fine. Any ideas?

stevapple:
Have you checked that the `Package.resolved` file points to Vapor 4 even after running `swift package resolve` again?

SwiftPM always prioritizes `Package.swift` requirement, and `4.x.x` doesn't match the `from: "3.0.0"` requirement. SwiftPM will silently revert the `Package.resolved` file to locally resolved versions in such case.

deivid-rodriguez:
Sorry, I didn't explain correctly. I tried to create an update with Dependabot and I ended up with the following manifest (and Vapor 4 in the `Package.resolved` file)

```swift
// swift-tools-version:5.2
// The swift-tools-version declares the minimum version of Swift required to build this package.
import PackageDescription

let package = Package(
    name: "Foo",
    platforms: [
        .macOS(.v10_13), .iOS(.v11), .tvOS(.v11), .watchOS(.v4)
    ],
    dependencies: [
        .package(url: "https://github.com/vapor/vapor.git", from: "4.77.2"),
        .package(url: "https://github.com/dependabot-fixtures/swift-example-vapor3", from: "1.0.0"),
    ],
    swiftLanguageVersions: [.v5]
)
```

Swift seemed happy with it and that surprised me.

stevapple:
I think I’ve got it. Since the Vapor dependency in https://github.com/dependabot-fixtures/swift-example-vapor3 is unused, SwiftPM will treat it as something like "local development dependency", which won’t be propagated to its dependents. If you really add it to a product, SwiftPM will emit the error.

deivid-rodriguez:
I see, thanks. I will look into that!

deivid-rodriguez:
@stevapple Indeed making those adjustments made SwiftPM raise resolvability errors. Writing this spec also help finding a few issues. Thanks!

This should be ready.

swift/lib/dependabot/swift/file_updater/lockfile_updater.rb
```diff
@@ -31,6 +28,18 @@ def updated_lockfile_content
...
 
+        def try_lockfile_update(dependencies)
+          SharedHelpers.run_shell_command(
+            "swift package update #{dependencies.map(&:name).join(' ')}",
+            fingerprint: "swift package update <dependency_name>"
+          )
+        rescue SharedHelpers::HelperSubprocessFailed
+          # We could check for resolvability errors in the error message.
+          # However, that's brittle and if there's other pre-existing errors in
+          # the project, they should be surfaced during dependency parsing.
```
jakecoffman, line 39: Do we want to at least log the message so folks can see why it didn't update?

jakecoffman (approved):
LGTM

stevapple:
> Sorry, I didn't explain correctly. I tried to create an update of the above example repo with Dependabot and I ended up with the following manifest (and Vapor 4 in the `Package.resolved` file)
> 
> ```swift
> // swift-tools-version:5.2
> // The swift-tools-version declares the minimum version of Swift required to build this package.
> import PackageDescription
> 
> let package = Package(
>     name: "Foo",
>     platforms: [
>         .macOS(.v10_13), .iOS(.v11), .tvOS(.v11), .watchOS(.v4)
>     ],
>     dependencies: [
>         .package(url: "https://github.com/vapor/vapor.git", from: "4.77.2"),
>         .package(url: "https://github.com/dependabot-fixtures/swift-example-vapor3", from: "1.0.0"),
>     ],
>     swiftLanguageVersions: [.v5]
> )
> ```
> 
> Swift seemed happy with it and that surprised me.

Did you mean Dependabot will update the `Package.swift` across major versions automatically? This is very likely to break existing codebase if CI is not set up properly.

deivid-rodriguez:
Yeah, that's how Dependabot works for every ecosystem: it updates to the latest version by default if it can. Dependabot could indeed break your CI when proposing updates, and merging version upgrades without CI is likely to break your programs.

We have ignore options though: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore.

I think the default behavior is fine because people also find useful that Dependabot does the work of updating dependency manifests for you, and signaling that there's a new version available through a PR, even if the PR needs more work and can't be merged right away.

PR #7643, opened 2023-07-27
Remove `composer` `v1` support

jeffwidman:
This removes support for `composer` `v1`.

I chose to do this by removing all version-selection code paths as it
results in simpler code. Down the road, if the `composer` team releases
`v3`, we will need to add this back, but I suspect that will be a _long_
time until it happens... v1 grew somewhat organically, and then v2 was
an attempt to cleanup the warts, so they'll probably stay with v2 for a
long time. So no point in us maintaining version selection code for
several years in the meantime.

`composer` `2` doesn't support PEAR, so we're implicitly dropping
support for it when we drop `composer 1`. However, PEAR usage has been
mostly supplanted by Packagist.

This is confirmed by our metrics which show low usage of `composer 1`.
And of those, I'm sure only a fraction will be using PEAR, so this
should have hardly any user impact.

Under `composer` `v1` package names were allowed to have
uppercase letters. These now throw an error in `composer` `v2`. I
updated the code / unit tests in this PR to assert the error is handled
as a well-known user error.

Fix: https://github.com/dependabot/dependabot-core/issues/6298


composer/lib/dependabot/composer/update_checker/version_resolver.rb
```diff
@@ -33,7 +33,7 @@ def initialize(extensions)
         MISSING_IMPLICIT_PLATFORM_REQ_REGEX =
           %r{
             (?<!with|for|by)\sext\-[^\s\/]+\s.*?\s(?=->)|
-            (?<=requires\s)php(?:\-[^\s\/]+)?\s.*?\s(?=->)| # composer v1
+            (?<=requires\s)php(?:\-[^\s\/]+)?\s.*?\s(?=->)| # composer v1 - TODO possibly this should be nuked?
```
jeffwidman, line 36: address before merge

composer/spec/dependabot/composer/update_checker/version_resolver_spec.rb
```diff
@@ -183,22 +183,40 @@
...
+    context "with a dependency name that is only valid in composer v1" do
+      let(:project_name) { "invalid_v2_requirement" }
       let(:dependency_name) { "monolog/Monolog" }
       let(:latest_allowable_version) { Gem::Version.new("1.25.1") }
       let(:dependency_version) { "1.0.2" }
 
-      it { is_expected.to eq(Dependabot::Composer::Version.new("1.25.1")) }
+      # TODO: currently this test fails because raises a Dependabot::SharedHelpers::HelperSubprocessFailed error. We
+      # need to catch/handle this as a DependencyFileNotParseable so that it's treated as a user-caused error for
+      # display/alert monitoring purposes.
```
jeffwidman, line 210: address before merge

composer/spec/dependabot/composer/update_checker_spec.rb
```diff
@@ -427,39 +421,6 @@
       it { is_expected.to be_nil }
     end
 
-    context "with a PEAR dependency (composer v1)" do
```
jeffwidman, line 430: `composer` `2` doesn't support PEAR, so we're implicitly dropping support for it when we drop `composer 1`. However, PEAR usage has been mostly supplanted by Packagist. 

This is confirmed by our metrics which show low usage of `composer 1`. And of those, I'm sure only a fraction will be using PEAR, so this should have hardly any user impact.

jeffwidman (commented):
Annotating the changes + marking down the TODOs that still need addressing.

composer/spec/dependabot/composer/update_checker/version_resolver_spec.rb
jeffwidman: Unclear to me if this should be `DependencyFileNotParseable` or `DependencyFileNotEvaluatable`... for now I went with `DependencyFileNotParseable`

composer/Dockerfile
```diff
@@ -53,8 +51,7 @@ RUN mkdir /tmp/composer-cache \
   && rm -rf /tmp/composer-cache
 
 COPY --chown=dependabot:dependabot composer/helpers /opt/composer/helpers
-RUN bash /opt/composer/helpers/v1/build \
-  && bash /opt/composer/helpers/v2/build
+RUN bash /opt/composer/helpers/build
```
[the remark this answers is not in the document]
jeffwidman: @Seldaek @stof Is this a reasonable assumption? 

Because if I'm wrong and you expect to ship `composer` `v3` with breaking changes in the not too distant future, then I'll keep the version-selection logic around...

composer/lib/dependabot/composer/helpers.rb
```diff
@@ -13,6 +13,9 @@ module Helpers
         |composer-(?:plugin|runtime)-api)$
       /x
 
+      # TODO: the only caller of this is `ecosystem_versions`, so extend/update this to actually return the full version
+      # string from `plugin-api-version` string
+      # also, this may be better moved to the FileParser rather than sitting here in Helpers...
```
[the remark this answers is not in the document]
jeffwidman: I'm actually going to address this after merge in a separate PR as there's a few things going on here, and easier to keep separate... the TODO note captures the gist of it.

jeffwidman:
This is basically ready, just blocked on whenever the :dependabot: team is comfortable with the customer impacts of dropping `composer v1`.

jeffwidman:
We've dropped `composer` `v1` support:
* https://github.blog/changelog/2024-11-06-deprecation-dependabot-no-longer-supports-composer-v1/
* https://github.com/dependabot/dependabot-core/discussions/10900

sachin-sandhu:
Closing this PR as the work related to drop support for V1 is complete now. https://github.com/dependabot/dependabot-core/pull/10934

PR #7648, opened 2023-07-27
Normalize Swift package names

deivid-rodriguez:
In other places in Github, like in Security Advisories, or the Dependency Graph page, a different naming schema is used.

This means that when we try to find security updates to resolve a certain alert for a dependency, we won't be able to match the dependency with the advisory in the alert, and Dependabot will see it as non vulnerable.

So, normalize the schema to what the rest of GitHub uses.


deivid-rodriguez:
This will require an update to the smoke tests but other than that I think it's ready.

Open questions:

* In the second commit I made an extra change to keep current PR titles using the short name, but we may want to keep the one starting with `github.com/`.
* I used metadata to pass the "identifier" around, since we need to pass that to the `swift package update` command to update specific packages, and swift does not know about our naming scheme. An alternative to using metadata would be to use a `display_name` builder like [the one for Gradle](https://github.com/dependabot/dependabot-core/blob/20930c0c92dd63328f9e21bde4aebd7b83e52edd/gradle/lib/dependabot/gradle.rb#L20-L27) for example, and deduce it from the long name.

Thoughts?

deivid-rodriguez:
I believe changing the naming schema here is also going to force users to use the `github.com/reactivecocoa/reactiveswift` naming schema in their ignore conditions?

I'm not sure that's ideal, but if we go that route, we should also use `github.com/reactivecocoa/reactiveswift` in PR titles and bodies to not make things more confusing.

deivid-rodriguez:
We believe the best path forward is to stick to full names, at least for now.

Last open question is regarding the implementation:

I used metadata to pass around the "identifier" dumped by `swift package show-dependencies --format join` around, since we need to pass identifier to the `swift package update` and `swift package resolve` commands to update specific packages, and swift does not know about our naming scheme.

An alternative to using metadata would be to use add some custom logic right before passing it to the package manager (like `name.split("/").last`).

Letting the package manager "decide" the logic to build the "identifier" for a dependency feels better, but it does require more work to allow passing that metadata around everywhere.

deivid-rodriguez:
This should go green once https://github.com/dependabot/smoke-tests/pull/112 is merged.

jakecoffman:
Oh your last comment says exactly the same, so yes we agree!

PR #7654, opened 2023-07-27
Show ignore conditions in a request  body

honeyankit:
## Context

To remove ignore conditions on the dependencies on a given pull request user needs to have option to view those conditions. This task is to implement that for the user.  The PR body should look something like below:

**PR Body**
Dependabot Ignore Conditions Applied on the Pull Request
|Dependency Name| Ignore Condition|
|-|-|
business| [< 1.9, >1.8.0]|
business| [< 1.6.0, >1.3.0]|
statesmen| [< 1.2.0, >1.1.0]|

The user will use the ignore condition mentioned in the PR body and will use the upcoming new [unignore command](https://github.com/github/dependabot-updates/issues/4424) to remove those conditions on a specific dependency.

Fixes: https://github.com/github/dependabot-updates/issues/4478

## Approach

In the PR body, displaying top 20 lasted ignore conditions (if any) sorted by `created_at` or `updated_at` for dependencies in the pull request. 

The reason for the limiting to show only 20 ignore conditions on the PR is body is because at this moment "GitHub limits PR/comment descriptions to a max of 65,536 characters". More detailed discussion on this topic is [here](https://github.com/github/dependabot-updates/issues/4478#issuecomment-1650558908)

### DIFF
--- a/common/lib/dependabot/pull_request_creator/message_builder.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder.rb
@@ -526,18 +526,23 @@ module Dependabot
           "| #{ic[:dependency_name]} | [#{ic[:version_requirement]}] |"
         end
       
+        summary = "Most Recent Ignore Conditions Applied to This Pull Request"
         # Define the table structure
         table_header = "| Dependency Name | Ignore Conditions |"
         table_divider = "| --- | --- |"
         table_body = table_rows.join("\n")
 
-        "\n#{[table_header, table_divider, table_body].join("\n")}\n"
+        body = "\n#{[table_header, table_divider, table_body].join("\n")}\n"
 
-        # Build the collapsible section
-        details = "<details>\n<summary>Most Recent Ignore Conditions Applied to This Pull Request</summary>\n" \
-                  "#{[table_header, table_divider, table_body].join("\n")}\n\n</details>"
+        if !%w(azure bitbucket codecommit).include?(source.provider)
+          # Build the collapsible section
+          msg = "<details>\n<summary>#{summary}</summary>\n\n" \
+                    "#{[table_header, table_divider, table_body].join("\n")}\n</details>"
 
-        "\n\n#{details}\n\n"
+          "\n#{msg}\n"
+        else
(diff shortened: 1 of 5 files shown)


jakecoffman (approved):
Looks good, nice work!

common/lib/dependabot/pull_request_creator/message_builder.rb
```diff
@@ -504,6 +523,46 @@ def metadata_cascades_for_dep(dependency)
         ).to_s
       end
 
+      def ignore_conditions_table
+        # Return an empty string if ignore_conditions is empty
+        return "" if @ignore_conditions.empty?
+
+        # Filter out the conditions where from_config_file is false and dependency is in @dependencies
+        valid_ignore_conditions = @ignore_conditions.select do |ic|
+          !ic[:from_config_file] && dependencies.any? { |dep| dep.name == ic[:dependency_name] }
```
jurre, line 532: @honeyankit where does this `from_config_file` value come from? I don't see it in https://github.com/github/dependabot-api/blob/9fc5f8ad58c303bb9a1df9e460ee812d5350e797/app/models/update_job.rb#L184-L187

honeyankit: @jurre: We also need to modify this function. My earlier assumption was wrong about the attributes of ignore_conditions, We have to replace `from_config_file` with `source`

honeyankit: I had implemented `@dependbot show <dep -name> ignore conditions` and implement something [similar](https://github.com/github/dependabot-api/pull/4311/files#diff-c2b744d74da5d1462a2bd142ab16ff464fa3ed91f0310f248eb650f5db506245R42-R75). We have to change the `ignore_conditions_table` method entirely.

honeyankit: I have added the field `created_at` and `updated_at` in update job on this PR: https://github.com/github/dependabot-api/pull/4319

PR #7658, opened 2023-07-28
Update docker_registry2

NautiluX:
* Need to change digest function used to retain functionality


NautiluX:
cc @jurre

jurre (approved):
This looks good to me, thanks!

Context for others, when we previously tried to update to v1.17 it changed the request we made to the registry to perform a GET instead of a HEAD request, which resulted in dockerhub aggressively rate limiting us. This version will perform a HEAD, and only fall back to use a GET if the registry does not support `HEAD` requests

jurre:
@NautiluX I'm not sure if I'll be able to deploy this today, I have a fairly full day, but I'll try to sneak it in, otherwise I'll try do it on Monday

NautiluX:
Thanks @jurre.

deivid-rodriguez:
@NautiluX Independent from this fix but, do you know if the API provides an easy way to find all tags pointing to a given digest? Or alternatively, to find all tags pointing to the same digest as a given tag? I think we're essentially doing that here, in the hard way (with a lot of requests):

https://github.com/dependabot/dependabot-core/blob/4dd41de922d8534ee67d0e0111f0a82807b54b19/docker/lib/dependabot/docker/update_checker.rb#L191-L197

NautiluX:
Sorry, I'm not a docker registry expert, just trying to get Dependabot to update our base images again. 😅
Dockerhub seems to have an [undocumented API endpoint](https://forums.docker.com/t/fetching-docker-image-tags-with-created-time-and-digest/85357) `/tags` (without `/list`) that does it, but it's not part of the registry v2 API so probably shouldn't be used here.

deivid-rodriguez:
Thank you! That's useful, we may want to consider it still, just for dockerhub if it's going to make a big difference.

Curious, why are updates broken in the older version of this library?

NautiluX:
the pagination through tags doesn't work for the Red Hat registry registry.access.redhat.com in the old version. Some new base image versions aren't detected as they may be on page 2-n. I opened https://github.com/deitch/docker_registry2/pull/92 to fix it in the lib, and it is merged, but need to propagate this change to Dependabot (this PR).

deivid-rodriguez:
Aaah, thanks for the context! And for contributing fixes :heart:. We'll take care of this PR soon.

NautiluX:
Thanks @deivid-rodriguez.

jurre:
I really wanted to get this out today, but it's getting pretty late over here for me, so I'll deploy this on Monday

jeffwidman (approved):

jeffwidman:
Deployed and monitored for a bit, and seems okay, so merging. 🤞 no other surprises.

PR #7664, opened 2023-07-28
[Grouped Updates] Refactor how SemVer groups are configured

brrygrdn:
This merges into #7581 and changes the way this rule is configured:

```yaml
groups:
   rails:
     patterns:
     - "rails*"
     update-types: ["minor", "patch"]
```

If `update-types` is not specified, we default to `["major", "minor", "patch"]`, i.e. all versions.

### DIFF
--- a/common/lib/dependabot/dependency_group.rb
+++ b/common/lib/dependabot/dependency_group.rb
@@ -12,10 +12,17 @@ module Dependabot
     ANY_DEPENDENCY_NAME = "*"
     SECURITY_UPDATES_ONLY = false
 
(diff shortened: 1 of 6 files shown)


jakecoffman (approved):

jakecoffman:
The swift failure is unrelated, looks like ignores don't work and the test isn't cached.

jakecoffman:
Updated smoke tests for this here: https://github.com/dependabot/smoke-tests/pull/110

LGTM

jurre (approved):

PR #7665, opened 2023-07-28
Nishnha/revert docker version regexp

Nishnha:
reverts https://github.com/dependabot/dependabot-core/pull/6080 to see what tests fail. We may have been overmatching regexpessions earlier, but now some common docker tags are not being detected correctly, like

`org/image-name:3.10-master-999`

which is a combination of `<name-with-owner>:<version>-<branch>-<build>`


deivid-rodriguez:
Uggh, nice catch @Nishnha! Embarrassed to have introduced this kind of bug just to fix a CodeQL alert. We should add a regression spec so this is not reintroduced in the future.

deivid-rodriguez:
Closing in favor of #7687!

Issue #4660, opened 2022-01-24
Dependabot environment variables


trent-abc:
I would like to specify an environment variable during dependabot runs.

I'm attempting to authenticate against a private cloudsmith python repository.  I arrived at a working configuration [here](https://github.com/dependabot/dependabot-core/issues/4539).  However, this working configuration means my Pipfile looks like so:

```
[[source]]
url = "https://dl.cloudsmith.io/basic/company/repo/python/index/"
verify_ssl = true
name = "cloudsmith"
```

While this works for dependabot given the yml in the aforementioned ticket, it doesn't work for local development.  For local development our pipfile looks like so:

```
[[source]]
url = "https://dl.cloudsmith.io/${CLOUDSMITH_KEY}/company/repo/python/index/"
verify_ssl = true
name = "cloudsmith"
```

So if I'm just able to set the environment variable ${CLOUDSMITH_KEY} to "basic" my problem is solved.

```
version: 2
updates:
  - package-ecosystem: "pip"
  ...
    env:
      CLOUDSMITH_KEY: basic
  ...
```

Unfortunately it's unsupported:

<img width="911" alt="Screen Shot 2022-01-24 at 3 28 54 PM" src="https://user-images.githubusercontent.com/82410520/150882352-20ccc5f1-9ff3-4a21-9644-63d7e42d3295.png">

Is there another way around this, or am I missing something?  I would have thought setting some environment variables was a common use-case.

nwalsh1995:
I have the same issue. The local paths in our dependency files are relative to the top level folder and not the project folder. Dependabot assumes the paths are relative to the project folder so it cannot follow dependencies. We can solve this issue because Python pip `requirements.txt` supports environment variables. So we specify the environment variable different for local dev and another for dependabot execution. This maintains compatibility with the two workflows if environment variables could be supported.

davidblewett:
As another point: we're hitting an issue where Dependabot can't correctly handle a Rust dependency via Cargo's Git implementation. We need to be able to set this: https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli , but can't due to missing environment variables for Dependabot.

Vadorequest:
In my case, I'd like to grant access to some of our ENV variables that are already set in the GitHub repo, that would be even better than duplicating those for Dependabot specifically.

Vadorequest:
I found this article: https://danielabaron.me/blog/dependabot-secrets/#dependabot-secrets

[...]

Issue #2001, opened 2020-04-02
Support using custom PHP environment variables to pass API key to paid plugins


Gandhi11:
Hi,

I'm using Bedrock with wordpress to be able to install my plugins using composer. For premium plugins, I am using [private-composer-installer](https://github.com/ffraenz/private-composer-installer) to install them. I am injecting multiple environment variables which are the keys to download the plugins.

Currently there are only two variables supported: **ACF_PRO_KEY** and **GRAVITY_FORMS_KEY**. Is there any way to add other ones?

For example, I am using these for ACF Pro and WPML.
- **PLUGIN_ACF_KEY** (I could change this one to **ACF_PRO_KEY**)
- **PLUGIN_WPML_SUBSCRIPTION_KEY**
- **PLUGIN_WPML_USER_ID**

viktorbijlenga:
We also use [Bedrock by Roots](https://github.com/roots/bedrock) to manage WordPress. We then use Dependabot to manage our WordPress plugins, with great help by the excellent [WordPress Packagist](https://wpackagist.org/).

Currently our main use of PHP environment variables is to manage the plugin Advanced Custom Fields Pro, which depends on a PHP environment variable to be updated as explained by @Gandhi11 above. 

Our composer solution for ACF is using [PiventIT/ACF-pro-installer](https://github.com/pivvenit/acf-pro-installer) by @Qrious. A lot of people is also using [PhilippBaschke /acf-pro-installer](https://github.com/PhilippBaschke/acf-pro-installer). On top of using ACF this way, we also have other plugins which depends on a similar workflow.

A week ago we started see a new pull reuqest from Dependabot across our repos, with information regarding the future of Dependabot. The information has the following part which is related to custom PHP enciroment variables in the future.

> Your repo is using PHP environment variables. There is no support for this registry type in GitHub-native Dependabot, so these registries were not added to the new config file. We recommend fetching these dependencies via GitHub Actions.

I would love to see a future in which it's possible to still update dependencies that need a PHP env variable in a good way. Anyone who have ideas on a good way forward that would work for a multitude of plugins (maybe using Github Actions as mentioned) would be a great addition to the PHP/WordPress community.

I would love if @asciimike or someone from the Dependabot team would pitch in here, since it would be great to hear their ideas in more detail, on how those who depends on this, could move this forward together.

Qrious:
This situation is undesired in my opinion, certainly because it is a significant step backwards compared to the preview. 

[...]

PR #7675, opened 2023-07-31
Stop explicitly catching missing env-var error in `composer`

jeffwidman:
This removes explicitly catching errors about missing env vars.

This is likely a bit controversial, so here's my thinking:

1. Dependabot at GitHub doesn't support injecting custom env vars. And for security reasons is unlikey to anytime soon: https://github.com/dependabot/dependabot-core/issues/4660
2. The primary use case was for `ACF Pro`, but that is no longer relevant since they now support private `composer` auth natively: https://github.com/dependabot/dependabot-core/issues/2001#issuecomment-1426401494
3. In general we've told users that if they need custom env vars for authentication reasons, they should instead ask their upstream providers to use the standard `composer` tooling, rather than a workaround. We aren't going to go out of our way to support non-standard tooling practices.
4. Setting env vars is still relevant for anyone who runs Dependabot standalone, but those are typically power users who are very comfortable debugging / reading the logs to realize the error is the lack of an env var.
5. Technically this could cause more internal alerts because removing this error changes those updates from being seen as "well-known user errors" to "unknown service errors" (which we internally alert on). However, I strongly suspect that very few users actually hit this problem... so if just a couple of repos are hitting it, that shouldn't be unduly affecting our metrics.
6. Currently unknown `composer` errors are hidden, but we should resolve that with a systemic solution rather than one-off-solutions for specific packages, which turns into a game of whack-a-mole: https://github.com/dependabot/dependabot-core/pull/6290
7. Composer error handling is a big mess in our code... it's copy/pasted several places, with `TODO` notes saying we should refactor/cleanup our error handling code. And we don't even know what's still relevant once we drop `composer` `1` as the error messages changed a bit. So any cleanup / simplification we can do of the error handling code makes our future refactoring easier.

If we needed this generic error for other ecosystems, that'd make sense, but so far it's `composer` specific so until we need that, let's just remove it.


deivid-rodriguez:
Turns out I need this error for other ecosystems. For example, this repo includes a `.yarnrc.yml` file that uses an environment variable, and that prevents dependabot from running. This user error is perfect for explaining this, so I'm closing this and will be opening a PR to catch this Yarn error and raise it as `Dependabot::MissingEnvironmentVariable`.

Let me know if you disagree though!

jeffwidman:
Makes sense to me to keep it. 👍

deivid-rodriguez:
https://github.com/dependabot/dependabot-core/pull/8446 will start using this error class 👍.

PR #7679, opened 2023-07-31
Move `ecosystem_versions` to `FileParser` class

jeffwidman:
The intent of `ecosystem_versions` is to report the range of versions supported by the target repo, _not the version that Dependabot chose to use_.

In order to do that, we need to parse the manifest files for any lines indicating the range of supported versions.

Intuitively, this sounds like `FileParser`, not `FileFetcher`.

Historically though, the code has lived in `FileFetcher`, where it sorta grew organically. So as we systematize this, it makes more sense to move it to the `FileParser`.

### DIFF
--- a/python/lib/dependabot/python/file_parser.rb
+++ b/python/lib/dependabot/python/file_parser.rb
@@ -46,6 +46,28 @@ module Dependabot
         dependency_set.dependencies
       end
 
+      def ecosystem_versions
+        # Hmm... it's weird that I have to instantiate a PythonRequirementParser rather than just use this existing
+        # FileParser... that kinda defeats the whole purpose of moving this to fileparser... maybe instead I should move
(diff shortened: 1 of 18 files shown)


jeffwidman:
Still need to move the unit tests to the right location, but first putting this up as a POC for discussion with some teammates... as I'm not quite sure this is the best way to go... theoretically the line between `FileFetcher`/`FileParser` is pretty clear, but working through this showed that in reality that line is a lot more muddled.

python/lib/dependabot/python/file_parser.rb
```diff
@@ -46,6 +46,28 @@ def parse
         dependency_set.dependencies
       end
 
+      def ecosystem_versions
```
jeffwidman, line 49: This method is one I could use a second opinion on...

jeffwidman:
@kbukum1 did a bunch of work to update the metrics, so I suspect this is no longer relevant.

Issue #5723, opened 2022-09-14
Need to update python fixtures to reflect new Warehouse API responses


jeffwidman:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code improvement description

Now that PyPI updated their backend to the warehouse implementation, the API's have been updated...
So we should update our other python test fixtures.

For example, I noticed some changes while working on https://github.com/dependabot/dependabot-core/pull/5445.

PR #7681, opened 2023-08-01
Stop checking deprecated `bugtrack_url`

jeffwidman:
Per the [Warehouse PyPI docs](https://warehouse.pypa.io/api-reference/json.html#project):
> The `bugtrack_url` key on this response should be considered deprecated.
>
> It currently always returns `null` and in the future, the `bugtrack_url` key may be removed from this response.

### DIFF
--- a/python/spec/fixtures/pypi/pypi_response_pendulum.json
+++ b/python/spec/fixtures/pypi/pypi_response_pendulum.json
@@ -2,7 +2,7 @@
     "info": {
         "author": "Sébastien Eustace",
         "author_email": "sebastien@eustace.io",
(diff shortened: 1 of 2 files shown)


python/spec/fixtures/pypi/pypi_response_pendulum.json
```diff
@@ -2,7 +2,7 @@
     "info": {
         "author": "Sébastien Eustace",
         "author_email": "sebastien@eustace.io",
-        "bugtrack_url": "",
```
jeffwidman, line 5: This is another example of why we need:
* #5723

PR #7682, opened 2023-08-01
Refresh PyPI response mocks

jeffwidman:
Our PyPI mocks were originally created in 2017 before they replaced the back end with Warehouse, which was a complete rewrite:

* https://github.com/dependabot/dependabot-core/commits/697d7f90fc0bf97c1c9d6004de7e8629483a101a/spec/fixtures/python/pypi_simple_response.html?browsing_rename_history=true&new_path=python/spec/fixtures/pypi/pypi_simple_response.html&original_branch=main
* https://pyfound.blogspot.com/2018/03/warehouse-all-new-pypi-is-now-in-beta.html

Ever since I realized this, I've felt a bit nervous about our tests.

Warehouse has mostly retained backwards compatibility in their APIs, but there were a few small changes, and also new fields.

So here I simply refreshed the mock by pulling up the current response from PyPI, then updated any unit tests that needed fixing.

While working on this, I noticed that the raw `simple` API that returns HTML is now offered with a JSON schema, but migrating to that it out of scope of this PR.


jeffwidman:
I still think this is a useful idea, but it's not something I'll have time to tackle for the foreseeable future.

PR #7688, opened 2023-08-01
Teach DependencyGroup to calculate the ranges we should ignore when looking for versions that fall outside the group

brrygrdn:
This is a follow up on our experiment with SemVer grouping in #7581 

The current implementation is _very alpha_ where we will:
1. Correctly prevent the group from updating beyond the desired SemVer range(s) configured
2. Attempt a second ungrouped update on any dependencies where we haven't addressed the highest possible version in-group

The behaviour for (2) is currently not fully implemented so it essentially results in us creating the highest version available **as well** as the group PR.

The next step is to improve how we process the separate updates in two ways:
- We should make best effort _not_ to attempt a second upgrade if we know the highest version is within the group for a dependency
- We should explicitly ignore any version ranges that fall **within** the group when doing this separate update.

This Pull Request is a baby step towards the latter which adds a new method to `DependencyGroup`:

```ruby
dependency_group.ignored_version_ranges_for_ungrouped_versions_of(dependency)
```

This method is essentially the opposite of `DependencyGroup#ignored_version_ranges_for` providing ignored ranges that fall within the group.

This branch does not wire this new behaviour up yet as we need a new way to collect and perform the dependencies that need to be updated which is aware of the DependencyGroup object so it can collect this information as part of the job.

### DIFF
--- a/common/spec/dependabot/dependency_group_spec.rb
+++ b/common/spec/dependabot/dependency_group_spec.rb
@@ -205,7 +205,7 @@ RSpec.describe Dependabot::DependencyGroup do
     end
   end
 
-  describe "#ignored_versions_for with experimental rules enabled" do
+  describe "#ignored_version_ranges_for with experimental rules enabled" do
     let(:dependency) do
       Dependabot::Dependency.new(
         name: "business",
@@ -227,7 +227,7 @@ RSpec.describe Dependabot::DependencyGroup do
 
     context "the group has not defined an update-types rule" do
       it "returns an empty array as nothing should be ignored" do
-        expect(dependency_group.ignored_versions_for(dependency)).to be_empty
+        expect(dependency_group.ignored_version_ranges_for(dependency)).to be_empty
       end
     end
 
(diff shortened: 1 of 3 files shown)


common/spec/dependabot/dependency_group_spec.rb
```diff
@@ -331,7 +331,131 @@
...
+        }
+      end
+
+      it "returns ranges which ignore minor versions" do
+        expect(dependency_group.ignored_version_ranges_for_ungrouped_versions_of(dependency)).to eql([
+          ">= 1.9.a, < 2"
+        ])
+      end
+    end
+  end
```
brrygrdn, line 430: I'd draw attention to this case as something non-intuitive but I've err'd on the side of doing exactly what the config tells us over guessing at what users want.

jakecoffman (approved):

brrygrdn:
I'm going to hold on merging this while we discuss some implications of this approach, we might not need this depending on what we decide 🤔

PR #7692, opened 2023-08-01
Test that unsupported Python versions raise the expected error

jeffwidman:
Since we'll be dropping 3.6/3.7 soon, we want to ensure that users on unsupported python versions get a clearly understandable error and not a random traceback.

### DIFF
--- a/python/spec/dependabot/python/update_checker_spec.rb
+++ b/python/spec/dependabot/python/update_checker_spec.rb
@@ -218,9 +218,21 @@ RSpec.describe Dependabot::Python::UpdateChecker do
 
         it { is_expected.to eq(Gem::Version.new("3.2.4")) }
 
-        context "that disallows the latest version" do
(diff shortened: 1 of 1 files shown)


python/spec/dependabot/python/update_checker_spec.rb
```diff
@@ -218,10 +218,22 @@
...
 
-        context "that disallows the latest version" do
-          let(:python_version_content) { "3.5.3\n" }
+        context "that is set to the oldest version of python supported by Dependabot" do
+          let(:python_version_content) { "3.6.0\n" }
           it { is_expected.to eq(Gem::Version.new("2.2.24")) }
         end
+
+        context "that is set to a python version no longer supported by Dependabot" do
+          let(:python_version_content) { "3.4.0\n" }
```
abdulapopoola, line 227: should we include a link to the deprecation notice in the original message?

abdulapopoola (approved):

PR #7703, opened 2023-08-03
Add `truncate_to_major_minor` helper method

jeffwidman:
Add a major/minor helper. Technically this may change behavior slightly if we start seeing versions that were successfully parsed by the regex but can't be parsed by `Python::Version.new()`, but in that case we should fix `Python::Version.new()` so that our string-> version parsing logic is consistent everywhere.

But as long as CI is happy than we should go with this unless we start seeing oddball python version strings showing up in our error logs.

### DIFF
--- a/python/lib/dependabot/python/language_version_manager.rb
+++ b/python/lib/dependabot/python/language_version_manager.rb
@@ -33,8 +33,7 @@ module Dependabot
       end
 
       def python_major_minor
-        @python ||= Python::Version.new(python_version)
-        "#{@python.segments[0]}.#{@python.segments[1]}"
+        @python_major_minor ||= Python::Version.new(python_version).truncate_to_major_minor
       end
 
       def python_version
@@ -67,12 +66,10 @@ module Dependabot
           find { |v| requirement.satisfied_by?(Python::Version.new(v)) }
         return version if version
 
-        # If not, and we're dealing with a simple version string
-        # and changing the patch version would fix things, we do that
-        # as the patch version is unlikely to affect resolution
+        # If not, and we're dealing with a simple version string and changing the patch version would fix things,
+        # we do that as the patch version is unlikely to affect resolution
         if requirement_string.start_with?(/\d/)
-          requirement =
-            Python::Requirement.new(requirement_string.gsub(/\.\d+$/, ".*"))
+          requirement = Python::Version.new(requirement_string).truncate_to_major_minor
           version =
             PythonVersions::SUPPORTED_VERSIONS_TO_ITERATE.
             find { |v| requirement.satisfied_by?(Python::Version.new(v)) }
--- a/python/lib/dependabot/python/version.rb
+++ b/python/lib/dependabot/python/version.rb
@@ -66,6 +66,10 @@ module Dependabot
         local_version_comparison(other)
       end
 
+      def truncate_to_major_minor
+        Python::Version.new(segments[0..1].join("."))
+      end
+
       private
 
       def epoch_comparison(other)
--- a/python/spec/dependabot/python/version_spec.rb
+++ b/python/spec/dependabot/python/version_spec.rb
@@ -149,6 +149,40 @@ RSpec.describe Dependabot::Python::Version do
         it { is_expected.to eq(false) }
       end
     end
+
+    describe "#truncate_to_major_minor" do
+      subject { version.truncate_to_major_minor }
+
+      context "with a prerelease" do
+        let(:version_string) { "1.0.0alpha" }
+        it { is_expected.to eq("1.0") }
+      end
+
+      context "with a normal release" do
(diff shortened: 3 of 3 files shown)


python/lib/dependabot/python/language_version_manager.rb
```diff
@@ -33,8 +33,12 @@ def install_required_python
...
 
       def python_major_minor
-        @python ||= Python::Version.new(python_version)
-        "#{@python.segments[0]}.#{@python.segments[1]}"
+        @python_major_minor ||= self.class.extract_python_major_minor(python_version)
+      end
+
+      def self.extract_python_major_minor(python_version_string)
+        full_version = Python::Version.new(python_version_string)
+        "#{full_version.segments[0]}.#{full_version.segments[1]}"
```
jeffwidman, line 41: actually, this should return a `Python::Version`, hang on a sec...

python/lib/dependabot/python/language_version_manager.rb
```diff
@@ -33,8 +33,12 @@ def install_required_python
       end
 
       def python_major_minor
-        @python ||= Python::Version.new(python_version)
-        "#{@python.segments[0]}.#{@python.segments[1]}"
+        @python_major_minor ||= self.class.truncate_version_to_major_minor(python_version)
```
mctofu, line 36: Does anything else rely on this method? This is changing the return value from a string to `Python::Version`

jeffwidman:
I need to play around with this a little rather than rushing it...

Ruby's `Gem::Version` class may always represent a truncated string as `"1.2.0"` rather than `"1.2", in which case it might be better if this helper returned a string of `"1.2"` as rounding down but giving `"1.2.0"` could cause confusion...

python/lib/dependabot/python/version.rb
```diff
@@ -66,6 +66,10 @@ def <=>(other)
         local_version_comparison(other)
       end
 
+      def truncate_to_major_minor
```
jeffwidman, line 69: This could be moved all the way to `Dependabot::Version` as we could probably use it elsehwere, but I thought that might be a little heavy-handed given how much version-like strings vary across ecosystems... So for now sticking with Python.

python/lib/dependabot/python/language_version_manager.rb
jeffwidman: True, I could slap a `.to_s` on the end, but I audited all the usages of it and from what I can tell they should all be fine with duck typing... ie, they're called from within a string, or they're used as values in a hash that is passed to `TomlRB.dump()` which I expect will call `.to_s` on it.

So I think we'll be fine with retaining the typing here and only coercing to string at the point of use.

The primary drawback I see of this is that it potentially ties us tighter to the `Gem::Version` class, when in many ways we'd rather treat the versions as opaque strings wherever possible for passing to native helpers. That could be a reasonable argument for having `python_major_minor` continue to return a string.

Thoughts?

mctofu: Got it! No strong preference, just wanted to make sure it was expected & works.

python/lib/dependabot/python/language_version_manager.rb
```diff
@@ -67,12 +66,10 @@ def python_version_from_supported_versions
...
         return version if version
 
-        # If not, and we're dealing with a simple version string
-        # and changing the patch version would fix things, we do that
-        # as the patch version is unlikely to affect resolution
+        # If not, and we're dealing with a simple version string and changing the patch version would fix things,
+        # we do that as the patch version is unlikely to affect resolution
         if requirement_string.start_with?(/\d/)
-          requirement =
-            Python::Requirement.new(requirement_string.gsub(/\.\d+$/, ".*"))
```
jeffwidman, line 75: Oh interesting, this isn't actually truncating to `major.minor`, it's actually setting a wildcard/range:
```ruby
(ruby) z = Python::Version.new(requirement_string).truncate_to_major_minor
Gem::Version.new("3.6")
(ruby) zz = Python::Requirement.new(requirement_string.gsub(/\.\d+$/, ".*"))
Gem::Requirement.new(["~> 3.6.0.a"])
(ruby) requirement_string.gsub(/\.\d+$/, ".*")
"3.6.*"
```

jeffwidman:
Closing in favor of:
* https://github.com/dependabot/dependabot-core/pull/7705

Due to https://github.com/dependabot/dependabot-core/pull/7703#discussion_r1282621774.

I am so thankful for thorough unit tests.

python/lib/dependabot/python/language_version_manager.rb
jeffwidman: Oh silly me, `Python::Requirement` != `Python::Version` 🤦 

There goes two hours of my life. No wonder the truncate didn't work here.

PR #7708, opened 2023-08-03
Enable `--verbose` when running specs

deivid-rodriguez:
That prints the actual RSpec command that gets run under the hood. If the current CI flakies are order dependent, it should allow us to reproduce them.

### DIFF
--- a/bundler/script/ci-test
+++ b/bundler/script/ci-test
@@ -3,7 +3,7 @@
 set -e
 
 bundle install
-bundle exec parallel_test spec/ -n "$CI_NODE_TOTAL" --only-group "$CI_NODE_INDEX" --group-by filesize --type rspec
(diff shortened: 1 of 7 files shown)


deivid-rodriguez:
Ok. So this did not help reproducing exactly but provided some clues. If I run the command now suggested in the error output, everything still passes. BUT, if I run that command, with my internet connection switched off, then the same set of specs fail, just with a different error message. Well, not exactly the same, I believe some more, but all the specs that fail in CI here also fail without an internet connection. I suspect this is some change in rubygems.org.

deivid-rodriguez:
This was not the proper hint, but I think it does not hurt to have it.

Nishnha (approved):
I don't have a strong opinion on this so I think it's fine to enable verbose output

deivid-rodriguez:
I think this is not too invasive, and quite helpful for troubleshooting CI, so let's do it. If somebody finds it too much we can switch back!

PR #7711, opened 2023-08-03
Upgrade `pip-tools` to `7.2.0`

jeffwidman:
Relevant breaking changes: https://github.com/jazzband/pip-tools/releases/tag/7.0.0

> Backwards Incompatible Changes:
> 
> - Default to --resolver=backtracking (https://github.com/jazzband/pip-tools/pull/1897). 
> - Drop support for Python 3.7 (https://github.com/jazzband/pip-tools/pull/1879).

So this is blocked by:
* #7702

### DIFF
--- a/python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
+++ b/python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
@@ -250,9 +250,6 @@ module Dependabot
           # TODO: Stop explicitly specifying `allow-unsafe` once it becomes the default:
           # https://github.com/jazzband/pip-tools/issues/989#issuecomment-1661254701
           options += ["--allow-unsafe"]
-          # TODO: This is the default as of https://github.com/jazzband/pip-tools/releases/tag/7.0.0
-          # so stop explicitly specifying it as soon as we upgrade
-          options += ["--resolver backtracking"]
 
           if (requirements_file = compiled_file_for_filename(filename))
             options << "--output-file=#{requirements_file.name}"
--- a/python/helpers/requirements.txt
+++ b/python/helpers/requirements.txt
@@ -1,5 +1,5 @@
 pip>=21.3.1,<23.3.0  # Range maintains py36 support TODO: Review python 3.6 support in April 2023 (eol ubuntu 18.04)
-pip-tools>=6.4.0,<=6.14.0  # Range maintains py36 support TODO: Review python 3.6 support in April 2023 (eol ubuntu 18.04)
+pip-tools==7.2.0
 hashin==0.17.0
 pipenv==2022.4.8
 pipfile==0.0.2


python/lib/dependabot/python/update_checker/pip_compile_version_resolver.rb
```diff
@@ -253,8 +248,9 @@ def pip_compile_options_fingerprint(options)
         def pip_compile_options(filename)
           options = @build_isolation ? ["--build-isolation"] : ["--no-build-isolation"]
           options += pip_compile_index_options
+          # TODO: Stop explicitly specifying `allow-unsafe` once it becomes the default:
+          # https://github.com/jazzband/pip-tools/issues/989#issuecomment-1661254701
           options += ["--allow-unsafe"]
-          options += ["--resolver backtracking"] if new_resolver_supported?
```
jeffwidman, line 257: Interestingly, just a few months ago this option was made inheritable from the conf file:
* https://github.com/dependabot/dependabot-core/pull/6914

https://github.com/dependabot/dependabot-core/blob/a29231615dbfa6ba5d063aaee82dc948a3b3006e/python/lib/dependabot/python/file_updater/pip_compile_file_updater.rb#L444-L445

But only over in `pip_compile_file_updater`, not here in `pip_compile_version_resolver`... @gmcrocetti curious why you only added it in one file?

jeffwidman: Also, out of scope for this PR, but I'd be personally inclined once this PR lands bumping `pip-tools` past `7.x` to drop support for the `legacy` resolver... it's no longer the default, it's days are numbered, so I suspect not that many folks use it... probably the original impetus for #6914 was actually to use the new `backtracking` resolver before it was the default.

 So if we removed support for `legacy`... probably be throwing a user-facing error if we encounter the option, then we wouldn't have to worry about supporting multiple paths and would also give a heads up to users that they need to switch.

gmcrocetti: Hey @jeffwidman .

> But only over in pip_compile_file_updater, not here in pip_compile_version_resolver... @gmcrocetti curious why you only added it in one file?

Sorry but I can't help you much here, I'm not well versed with the codebase. What I do know and call tell you for sure is that `--resolver=backtracking` never worked before #6914. Also, line `257` checked (should) true for my private repository (We started with python 3.9.*) - LMK if I can provide more information.

> probably the original impetus for https://github.com/dependabot/dependabot-core/pull/6914 was actually to use the new **backtracking resolver before it was the default**.

In fact the idea was to let users choose whatever resolver they wanted to.

>  probably be throwing a user-facing error if we encounter the option, then we wouldn't have to worry about supporting multiple paths

I do agree "not that many folks" are using `legacy` but I would follow Hyrum's Law take on that and let users decide until of course pip-tools removes the support.

jeffwidman: Thanks for getting back to me. 

> What I do know and call tell you for sure is that --resolver=backtracking never worked before https://github.com/dependabot/dependabot-core/pull/6914.

Yes, because the `backtracking` resolver wasn't the default until `7.0.0` release.

I guess on second thought, there may be other resolvers down the road too, who knows... in which case we should properly implement the resolver checks both places.

jeffwidman: I took a further look at this, prompted by:
* https://github.com/dependabot/dependabot-core/pull/7738

TBH, there's fundamentally more here than just the missing `resolver` option... we simply don't look at the options at all specified in the header of the compiled `requirements.txt`. Furthermore, there's a lot of cruft in there of copy/pasted code... so IMO it's not worth hacking in just this single option... instead, better to look at refactoring the whole "parsing" code so that its DRY'd up and reusable from both places:
* https://github.com/dependabot/dependabot-core/issues/7740

deivid-rodriguez (approved):

PR #7713, opened 2023-08-03
Use dockerignore rules that play better with recent docker versions

deivid-rodriguez:
Apparently new versions on Docker (at least on v23) no longer include Gemfile.lock files inside `bundler/spec/fixtures` in the build image.
    
Reading our current `.dockerignore`, it seems that the new behaviour prioritizes more explicit rules (not ending with `*`) even if they appear earlier in the file.
    
I'm not sure if this is expected, or a bug, but we can change the entries to not be ambiguous and simplify them.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -13,11 +13,10 @@
 /tmp
 **/.bundle
 **/coverage
-**/Gemfile.lock
+/Gemfile.lock
+*/Gemfile.lock
 !updater/Gemfile.lock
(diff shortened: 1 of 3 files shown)


jeffwidman (approved):

deivid-rodriguez:
This should fix Bundler's CI, currently broken due to the recent upgrade of the docker client in Actions base image.

PR #7737, opened 2023-08-05
Delete unused test fixture

jeffwidman:
This file appears unused. It was added in
https://github.com/dependabot/dependabot-core/commit/b6ae9d3bbb32647e012f9cef0b20e4af131c6387 but I don't see any reference to it in that commit's `pip-compile` spec files.

I do not see any accompanying `.in` file, so only this needs to be removed.

So let's see what CI thinks.


Nishnha (approved):
Looks unreferenced to me https://github.com/search?q=repo%3Adependabot%2Fdependabot-core%20pip_compile_unresolvable_if_unpinned&type=code

PR #7739, opened 2023-08-05
Delete unused method `error_certainly_bad_python_version?`

jeffwidman:
This method is not called anywhere in our codebase.

`git blame` shows it was added 4 years ago and used at that time, but the python version detection code has been heavily refactored since then so I didn't bother to trace where it's usage got removed.

So let's nuke it.


Nishnha (approved):
Looks unused 👍🏾

PR #7741, opened 2023-08-05
Switch from pipfile to plette lib

jeffwidman:
`plette` requires `python 3.7`, so blocked on:
* #6299 

The main user of `pipfile` is `pipenv`, and _not_ any of the other
python package managers.

However, `pipfile` library has been pretty much unmaintained, so
`pipenv` switched to using `plette` for parsing/validation of
`Pipfile`'s:
* https://github.com/pypa/pipenv/issues/5310
* https://github.com/pypa/pipenv/pull/5339

So let's switch our usage as well. Today we only use `pipfile` for
generating hashes, so this is effectively a silent no-op. However, down
the road we could leverage `plette` for `Pipfile` parsing/validation...
for example see how it's flagging things here:
https://github.com/dependabot/dependabot-core/pull/6104#issuecomment-1356403336

### DIFF
--- a/python/helpers/lib/hasher.py
+++ b/python/helpers/lib/hasher.py
@@ -1,6 +1,6 @@
 import hashin
 import json
-import pipfile
+import plette
 from poetry.factory import Factory
 
 
@@ -15,9 +15,10 @@ def get_dependency_hash(dependency_name, dependency_version, algorithm):
 
 
 def get_pipfile_hash(directory):
-    p = pipfile.load(directory + '/Pipfile')
+    with open(directory + '/Pipfile') as f:
+        pipfile = plette.Pipfile.load(f)
 
-    return json.dumps({"result": p.hash})
+    return json.dumps({"result": pipfile.get_hash().value})
 
 
 def get_pyproject_hash(directory):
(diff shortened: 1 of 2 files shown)


jeffwidman:
@matteius @oz123 What the current state of `plette`? I see `pipenv` uses it instead of `pipfile`, but then I saw a comment by you two saying that `plette` is also in maintenance/only critical fixes mode. Curious why that is? Is there a different library planned for the future?

oz123:
I'm actively working on palette. It's pretty much complete, hence there is not much to do.
Given to much free time , I might replace Cerberus with dataclass validation.

jeffwidman:
Superseded by (since I am not working on the :dependabot: team, I no longer have write access to this repo):
* https://github.com/dependabot/dependabot-core/pull/8627

The failing test is solved by:
* https://github.com/dependabot/dependabot-core/pull/8626

oz123:
Hey Jeff, headsup, I am about to release a new version of plette some when in January 2024. I am just waiting from the review from @matteius. Then new release is a complete rewrite of the core without cerberus.
The new code has the same old API.

jeffwidman:
Thanks for the heads up, maybe we can have the conversation over in #8627 though since I am no longer on the Dependabot team I had to re-open this PR from my own fork... 

But really though, as long as the public API for hash generation doesn't change, then it won't really matter. Can merge this now and then :dependabot: will send a bump PR whenever you release the new version.

PR #7744, opened 2023-08-06
Add `name` key to `sources` in `Pipfile`s

jeffwidman:
Newer versions of `pipenv` require any specified `sources` to be
explicitly `name`'d.

More context here: https://github.com/pypa/pipenv/discussions/5370#discussioncomment-3701061

This has been true since at least Sept 2022. Our version of `pipenv` is
from April of 2022, so it doesn't complain. But it will as soon as we upgrade.

Also, for sources that :dependabot: dynamically injects into the
`Pipfile`, they need a name. These sources are stripped from the final
`Pipfile` / `Pipfile.lock` during the `FileUpdater#post_process_lockfile`
method, so all we need is a placeholder `name` to placate `pipenv`.

Long term, we may want to add custom error handling to flag this
missing key as a `Dependabot::DependencyFileNotResolvable` error.

But I decided that was out of scope for now as this PR does not generate
the error... that will not happen until we upgrade to newer `pipenv`.
And even at that point, our first priority will be upgrading, and then
from there handling any new errors that start popping up.

And even then, most of the active users of `pipenv` are unlikely to see this
error because they're likely running a newer version of `pipenv`.

### DIFF
--- a/python/lib/dependabot/python/file_updater/pipfile_preparer.rb
+++ b/python/lib/dependabot/python/file_updater/pipfile_preparer.rb
@@ -130,9 +130,12 @@ module Dependabot
 
         def config_variable_sources(credentials)
           @config_variable_sources ||=
-            credentials.
-            select { |cred| cred["type"] == "python_index" }.
-            map { |c| { "url" => AuthedUrlBuilder.authed_url(credential: c) } }
+            credentials.select { |cred| cred["type"] == "python_index" }.map.with_index do |c, i|
+              {
+                "name" => "dependabot-inserted-index-#{i}",
+                "url" => AuthedUrlBuilder.authed_url(credential: c)
+              }
+            end
         end
       end
     end
--- a/python/spec/fixtures/pipfile_files/version_not_specified.lock
+++ b/python/spec/fixtures/pipfile_files/version_not_specified.lock
@@ -1,12 +1,13 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "c402ea48092e9d467af51a483bb8dd8ad0620e11c94f009dcd433f97a99d45db"
+            "sha256": "e76ae491d793d659f05c7f7eab261fd6167dc062efcba08f17be68e73eb87665"
         },
         "pipfile-spec": 6,
         "requires": {},
         "sources": [
             {
+                "name": "pypi",
                 "url": "https://pypi.org/simple",
                 "verify_ssl": true
             }
--- a/python/spec/dependabot/python/file_updater/pipfile_file_updater_spec.rb
+++ b/python/spec/dependabot/python/file_updater/pipfile_file_updater_spec.rb
@@ -179,6 +179,34 @@ RSpec.describe Dependabot::Python::FileUpdater::PipfileFileUpdater do
       end
     end
 
+    context "with a source not included in the original Pipfile" do
+      let(:credentials) do
+        [
+          {
+            "type" => "git_source",
+            "host" => "github.com",
+            "username" => "x-access-token",
+            "password" => "token"
+          },
+          {
+            "type" => "python_index",
+            "index-url" => "https://pypi.posrip.com/pypi/"
+          }
+        ]
+      end
+
+      it "the source is not included in the final updated files" do
+        expect(updated_files.map(&:name)).to eq(%w(Pipfile.lock)) # because Pipfile shouldn't have changed
+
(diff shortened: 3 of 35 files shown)


jeffwidman:
Likely there will be hash mismatches in the lockfiles, so they'll need regenerating... if so, check one or two to see what's actually changing... as no point in regenerating here using an older version of `pipenv` only to have to regenerate them again under the newer version of `pipenv`...

python/spec/fixtures/pipfile_files/version_not_specified.lock
```diff
@@ -1,12 +1,13 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "c402ea48092e9d467af51a483bb8dd8ad0620e11c94f009dcd433f97a99d45db"
+            "sha256": "e76ae491d793d659f05c7f7eab261fd6167dc062efcba08f17be68e73eb87665"
```
jeffwidman, line 4: This was the only lockfile that the tests required to have an updated hash.

I strongly suspect the hashes are outdated for the other fixture lockfiles, but again, we can regenerate them all after updating to latest `pipenv` once we drop Python 3.7.

python/lib/dependabot/python/file_updater/pipfile_preparer.rb
```diff
@@ -130,9 +130,15 @@ def sub_auth_url(source, credentials)
...
           @config_variable_sources ||=
-            credentials.
-            select { |cred| cred["type"] == "python_index" }.
-            map { |c| { "url" => AuthedUrlBuilder.authed_url(credential: c) } }
+            credentials.select { |cred| cred["type"] == "python_index" }.
+            map do |c|
+              {
+                "url" => AuthedUrlBuilder.authed_url(credential: c),
+                # Random suffix ensure names are unique. This entire source is stripped out of the final Pipfile.lock in
+                # FileUpdater#post_process_lockfile so it's okay that the names are non-deterministic.
```
jeffwidman, line 138: Actually lemme add a spec that ensures this is stripped out of the final `Pipfile[.lock]` files

python/lib/dependabot/python/file_updater/pipfile_preparer.rb
```diff
@@ -130,9 +130,15 @@ def sub_auth_url(source, credentials)
...
+                "name" => "dependabot-inserted-index-#{SecureRandom.alphanumeric(5)}"
+              }
+            end
```
deivid-rodriguez, line 141: Looks fine to me. It'd be nice to try a real update and make sure that these never end up leaking into a PR. I briefly read over the code, and I agree these should be stripped out but just in case.

Regarding implementation, I think I'd go with something more deterministic, like switching to `map.with_index |c, i|`, and appending the index. But doesn't matter too much I guess.

deivid-rodriguez (approved):

python/lib/dependabot/python/file_updater/pipfile_preparer.rb
deivid-rodriguez: Oh I missed your comment where you say you'll add a spec to make sure it's stripped. That's even better!

jeffwidman:
We got a report via a support ticket that :dependabot: opened a PR that included `index: dependabot-inserted-index-0 ` in `Pipfile.lock`. So this PR introduced some bug... 

Unfortunately, I can't repro and the affected user's repo is a private company so I can't access their repo and they are migrating to a different python package manager so decided it wasn't worthwhile for them to try to create a reproducible example.

The one observation they made:
> the bad index seems to be added while updating the sub-dependencies of our private libs.

If anyone else encounters this, please feel free to file a _new_ issue, preferably here in `dependabot-core` and linking to reproducible example and I'd be happy to take a look.

juanitosvq:
We are having the same issue but unfortunately also in a private repository :disappointed: 

It looks like older versions of `pipenv` are happy with that new `index`, but certainly the newest version `pipenv-2023.8.28` is not:
```
Successfully uninstalled pipenv-2023.2.4

..............................................................................................................................

Installing dependencies from Pipfile.lock (923e0b)...
--
125 | Unable to find dependabot-inserted-index-0 in sources, please check
126 | dependencies:
```

We also have private libraries, so not sure I'll be able to create a reproducible example.

jeffwidman:
Can you at least create a new issue describing what you're seeing? The `pipenv` maintainers are great folks and we have good lines of communication open with them, so maybe between all of us we can get it figured out...

Issue #7743, opened 2023-08-06
npm grouping never creates a PR, but creates a branch with commit


KyleSanderson:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

npm

### Package manager version

pnpm

### Language version

18

### Manifest location and content before the Dependabot update

https://github.com/autobrr/autobrr/commit/e0459d5a297f8aaad5f9f33557fe4006f3cddcb9

### dependabot.yml content

https://github.com/autobrr/autobrr/pull/1028

### What you expected to see, versus what you actually saw

Records there's 35 updates - where's the PR?

### Native package manager behavior

Reasonable, working behaviour.

jakecoffman:
Thanks for reporting this. A recent PR changed truncation behavior, so large PR bodies were not getting truncated. I made a change in the Dependabot service that added truncation back and I was able to get a PR from a fork: https://github.com/jakecoffman/autobrr/pull/4

KyleSanderson:
Thank you very much

PR #7753, opened 2023-08-07
fix PR unable to create with grouped updates

jakecoffman:
fixes https://github.com/dependabot/dependabot-core/issues/7743

We're seeing large updates that have large PR bodies fail to create on GitHub with a 422. I think this is due to [this change](https://github.com/dependabot/dependabot-core/pull/7487) moving the truncation from the `PullRequestCreator.create` to the `MessageBuilder`, and also we started recently using the `MessageBuilder` here in the `DependencyChange` class. It didn't get the new truncation settings passed in.

For now I copy-pasted the PullRequestCreator's new switch case to add in the limit to get PRs flowing again. This code seems like it could use a refactor.

### DIFF
--- a/updater/lib/dependabot/dependency_change.rb
+++ b/updater/lib/dependabot/dependency_change.rb
@@ -33,6 +33,8 @@ module Dependabot
         pr_message_max_length = Dependabot::PullRequestCreator::Codecommit::PR_DESCRIPTION_MAX_LENGTH
       when "bitbucket"
         pr_message_max_length = Dependabot::PullRequestCreator::Bitbucket::PR_DESCRIPTION_MAX_LENGTH
+      else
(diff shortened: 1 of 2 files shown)


honeyankit (approved):

jakecoffman:
Actually this won't fix it because the dependabot-api appends additional data to the end of the message. I'll leave this open though, we might want to ship truncation at this level anyway as we're going to discard the additional message data either way. Might as well shed some sooner?

jeffwidman (approved):
I vote ship this as well because it provides a code tie-in to the other code here in `core`... ie, if someone is grep'ing for usages they'll see it's used in the updater/cloud/grouped updates work...

jurre (approved):

PR #7757, opened 2023-08-07
Don't cancel full CI runs on main

deivid-rodriguez:
Since CI on PRs only runs depending on changed files, I think it's good to have an after-merge full CI for every PR.

Solution taken from https://stackoverflow.com/a/70972844/143243.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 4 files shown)


jeffwidman (approved):
Nice!

This would have prevented this failure/cancellation: https://github.com/dependabot/dependabot-core/commit/58edfc36a134528a12972c2019a618dcf9b50f05

jeffwidman:
Let's land this one ASAP so we stop getting cancellations on `main`.

PR #7760, opened 2023-08-08
More swift requirement parsing fixes

deivid-rodriguez:
Switches to a more explicit regex that should be more robust.


swift/spec/fixtures/projects/double_parentheses/Sources/swift-package-monitored-by-dependabot/swift_package_monitored_by_dependabot.swift
```diff
@@ -0,0 +1,2 @@
+// The Swift Programming Language
+// https://docs.swift.org/swift-book
```
jurre, line 2: What are these new files used for?

jurre (approved):

Issue #6487, opened 2023-01-19
Review `required_files_in?` necessity


deivid-rodriguez:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code improvement description

We have a `required_files_in?` as part of our public fetcher API: https://github.com/dependabot/dependabot-core/blob/5024751db27968f00c769ed93e34243e268b780f/common/lib/dependabot/file_fetchers/README.md#public-api

But only the Bundler ecosystem seems to be actually using it? Ecosystems seem to normally override `fetch_files` (a private method called by the public `files` method) and raise an explicit error from there when a required file is not found.

Keeping these while being unused causes additional burden to both contributors and maintainers. See https://github.com/dependabot/dependabot-core/pull/6248#discussion_r1081860748.

We should evaluate whether we should deprecate this method, or start actually using it.

PR #7763, opened 2023-08-08
Unify file fetchers

deivid-rodriguez:
WIP for https://github.com/dependabot/dependabot-core/issues/6487.


deivid-rodriguez:
I'm not sure what the best course of action is, so I won't push this further for now.

PR #7764, opened 2023-08-08
Replace deprecated `pipenv lock` with `pipenv requirements`

jeffwidman:
In the versions between what we are currently on in `dependaobt-core` and the latest versions of `pipenv`, they deprecated and then completely removed the `pipenv lock [-d]` command in favor of `pipenv requirements [--dev]` command.

Since our current version of `pipenv` supports the `pipenv requirements` command, let's proactively switch to it now. That will make the future upgrade in https://github.com/dependabot/dependabot-core/pull/7715 easier.

More context:
* https://github.com/pypa/pipenv/pull/5069
* https://github.com/pypa/pipenv/pull/5091
* https://github.com/pypa/pipenv/issues/5198
* https://github.com/pypa/pipenv/pull/5200

### DIFF
--- a/python/lib/dependabot/python/file_updater/pipfile_file_updater.rb
+++ b/python/lib/dependabot/python/file_updater/pipfile_file_updater.rb
@@ -92,7 +92,7 @@ module Dependabot
 
           # Find any requirement files that list the same dependencies as
           # the (old) Pipfile.lock. Any such files were almost certainly
-          # generated using `pipenv lock -r`
+          # generated using `pipenv requirements`
           requirements_files.select do |req_file|
             deps = []
             req_file.content.scan(regex) { deps << Regexp.last_match }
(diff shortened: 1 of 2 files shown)


honeyankit (approved):

jeffwidman:
BTW @matteius it was a little painful to track this down. 

I'm all for deprecations and even removing functionality, but maybe leave the deprecation message around a little longer? 

Our current version is less than two years old and in that time it was deprecated, but then the warning was also removed... would have been nicer if `pipenv lock -r` became a no-op with a warning message for another 12-18 months so that I didn't struggle trying to figure out if there was a replacement command or if the functionality was completely removed.

The other way this is typically done is with Semver where I can look at the major version release notes to see the breaking changes. But that wasn't possibly since `pipenv` is using semver.

Just a suggestion.

matteius:
Thanks for the feedback @jeffwidman -- I think that is a reasonable request we can try to accommodate going forward.  CC @oz123

PR #7772, opened 2023-08-09
Unify memoization

deivid-rodriguez:
We seem to be using different techniques to memoize results of methods.

I tried to unify this around:

* Use standard `||=` for methods that should never return nil like `updated_dependency_files`.
* Use a `defined?` guard for methods that may return nil, like the ones for fetching files if they are present.

### DIFF
--- a/common/lib/dependabot/metadata_finders/base.rb
+++ b/common/lib/dependabot/metadata_finders/base.rb
@@ -117,9 +117,8 @@ module Dependabot
       end
 
       def source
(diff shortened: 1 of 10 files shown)


jurre (approved):
Looks good to me 👍 is there a downside to always defaulting to `if defined?(@ivar)`?

deivid-rodriguez:
I guess just that it's maybe a bit too verbose over `||=`?

landongrindheim (approved):
😍

PR #7773, opened 2023-08-09
Fix incorrect memoizations

deivid-rodriguez:
The `fetch_file_if_present` method returns `nil` if the file is not present. That means we can't use `||=` for memoizing its result, and we'll end up retrying every time.


landongrindheim (approved):

deivid-rodriguez:
@landongrindheim I pushed more changes so that now all instances of this issue should be fixed!

PR #7774, opened 2023-08-09
`fetch_file_if_present` should ignore all "Not Found" errors

deivid-rodriguez:
The `fetch_file_if_present` method is used for fetching optional files. So it should tolerate "Not found" errors. Currently it does tolerate them, except when they are caused by a missing folder. For example, when looking for `.mvn/extensions.xml` and a `.mvn` directory is not there.

This PR makes it consistently ignore all "Not found" errors.

### DIFF
--- a/common/lib/dependabot/file_fetchers/base.rb
+++ b/common/lib/dependabot/file_fetchers/base.rb
@@ -132,8 +132,7 @@ module Dependabot
 
         fetch_file_from_host(filename, fetch_submodules: fetch_submodules)
       rescue *CLIENT_NOT_FOUND_ERRORS
-        path = Pathname.new(File.join(directory, filename)).cleanpath.to_path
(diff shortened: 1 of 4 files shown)


common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -132,8 +132,7 @@ def fetch_file_if_present(filename, fetch_submodules: false)
 
         fetch_file_from_host(filename, fetch_submodules: fetch_submodules)
       rescue *CLIENT_NOT_FOUND_ERRORS
-        path = Pathname.new(File.join(directory, filename)).cleanpath.to_path
-        raise Dependabot::DependencyFileNotFound, path
+        nil
```
landongrindheim, line 135: This new behavior matches the method name, so it makes sense to me 👍

landongrindheim (approved):

PR #7776, opened 2023-08-09
implement semver grouping and individual PRs

jakecoffman:
This modifies semver grouping to make it put a dependency into either a group or an individual PR, not both. 

So if you have a dependency that is potentially grouped with `update-types: ["patch", "minor"]` and a major version is available, it will create an individual PR. If the desire is to have the dependency in the group, an ignore should be added to the job for the major version of that dependency.

I updated the smoke tests [here](https://github.com/dependabot/smoke-tests/pull/110) to reflect the current state of this branch.


updater/lib/dependabot/dependency_snapshot.rb
```diff
@@ -69,9 +69,16 @@ def groups
...
+        all_grouped_changes.none? { |change| change.name == dep.name }
+      end
+    end
+
     def ungrouped_dependencies
       # If no groups are defined, all dependencies are ungrouped by default.
       return allowed_dependencies unless groups.any?
+      return @ungrouped_dependencies if defined?(@ungrouped_dependencies)
 
       @dependency_group_engine.ungrouped_dependencies
```
jurre, line 83: Can we not always store the `upgrouped_dependencies` on the `dependency_group_engine` 🤔

updater/lib/dependabot/updater/group_update_creation.rb
```diff
@@ -207,6 +197,31 @@ def all_versions_ignored?(dependency, checker)
         true
       end
 
+      # if the latest update is greater than the update-types, then it should not be in the group, but
+      # be an individual PR, or another group that fits it
+      def include_in_group?(group, dependency, checker)
+        return true unless group.rules["update-types"]
```
jurre, line 203: If there's no `rules["update-types"]` it always goes in the group? Why is that?

jurre (approved):
Looks good to me, I left some small questions in-line, mostly for my own understanding, but nothing blocking, nice work on this!

updater/lib/dependabot/dependency_snapshot.rb
jakecoffman: I can try pushing it down into the engine, but there's no reference to dependency_group_engine in GroupUpdateAllVersions, so it will pass through the dependency_snapshot anyway or I could add it to attr_reader. 

Not sure if that's better 🤷

updater/lib/dependabot/updater/group_update_creation.rb
jakecoffman: Then they didn't specify a semver grouping. The prod/dev and pattern checks happen before we get to this point, so there's no need to check anything else here. 

Should I change the method name to something more semver specific for clarity? `is_semver_allowed?`

updater/lib/dependabot/dependency_snapshot.rb
jurre: I was thinking that way we could get rid of the `return @ungrouped_dependencies if defined?(@ungrouped_dependencies)`. I'm also not sure if it's better, I'm also fine keeping things as is

updater/lib/dependabot/updater/group_update_creation.rb
jurre: Should we pull this check out of `include_in_group?` maybe? Feels like it's a different concern. Idk, I don't feel strongly, we can also leave a comment explaining it

updater/lib/dependabot/dependency_snapshot.rb
jakecoffman: Let's leave it for now, I'm considering some refactorings that we can try after this goes in that might improve the situation. Probably good to leave that for a new PR.

updater/lib/dependabot/updater/group_update_creation.rb
jakecoffman: Like so?

```ruby
return [] unless group.rules["update-types"] && semver_allows_grouping?(group, dependency, checker)
```

I kind of like it in the method, I added comments and made the method name more obvious that it's applying semver  grouping rules.

PR #7778, opened 2023-08-09
Improve `fetch_file_if_present` to make a single API request instead of two

deivid-rodriguez:
I noticed this file does two requests: One to check whether the file is present inside it's folder, then other one to actually fetch the file.

I wonder why it can't just fetch the file directly.

I believe this would result in:

* Simpler code.
* More performant fetching.
* Easier fetcher specs since there's less requests to mock.
* Less cluttered logs.

Opening a PoC PR to see what's I'm missing.


deivid-rodriguez:
A second thought after looking at CI reveals that the current approach is probably better, since the "directory listing" API calls can be reused to avoid trying to fetch files if we know there are not there. I'm going to close this.

PR #7784, opened 2023-08-10
Stop pinning `wheel`

jeffwidman:
We've gone back and forth on this repeatedly:
* https://github.com/dependabot/dependabot-core/pull/5597
* https://github.com/dependabot/dependabot-core/pull/7748#issuecomment-1670974893

As @yeikel pointed out, if we're going to keep pinning this, we really ought to have CI that checks it in some way (although that'd potentially be tricky as we want to not only test on latest python, but also oldest python).

As Deivid pointed out though, it's not really providing a lot of benefit for us to pin... simpler to just let `pip` pick whatever it needs and keep going.

If we observe breakage, we can start pinning again. Although probably (hopefully) that'd be very infrequent, and it'd be only a temporary thing until upstream fixes their bug and releases a new version then we can drop the pin. Or in that case I'd probably actually expect `pip-tools` to handle the work of temp-pinning as they're the ones who need it.


yeikel (commented):
This should be fine as that all our tests are passing. If any regression is introduced the plan that you mentioned should work

deivid-rodriguez (approved):

PR #7786, opened 2023-08-10
Fix private source authentication error message

deivid-rodriguez:
Recent Bundler versions include `bundle config set --global <source> username:password` in the error message, instead of `bundle config <source> username:password`.

But our regexp had not been updated, so the source was getting parsed as `set --global <source>`.


landongrindheim (approved):
📈

jeffwidman (approved):

PR #7787, opened 2023-08-10
Update ignore condition table

jurre:
Previously the ignore conditions weren't showing, because we were assuming symbol keys, when these keys are strings and formatted in kebab-case.

### DIFF
--- a/common/lib/dependabot/pull_request_creator/message_builder.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder.rb
@@ -529,7 +529,7 @@ module Dependabot
 
         # Filter out the conditions where from_config_file is false and dependency is in @dependencies
         valid_ignore_conditions = @ignore_conditions.select do |ic|
-          ic["source"] == "@dependabot ignore command" && dependencies.any? { |dep| dep.name == ic["dependency-name"] }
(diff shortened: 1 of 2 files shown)


common/lib/dependabot/pull_request_creator/message_builder.rb
```diff
@@ -529,18 +529,18 @@ def ignore_conditions_table
 
         # Filter out the conditions where from_config_file is false and dependency is in @dependencies
         valid_ignore_conditions = @ignore_conditions.select do |ic|
-          !ic[:from_config_file] && dependencies.any? { |dep| dep.name == ic[:dependency_name] }
+          ic["source"] == "@dependabot ignore command" && dependencies.any? { |dep| dep.name == ic["dependency-name"] }
```
landongrindheim, line 532: ❓ Would it make sense to make this 🤏 more broad? It seems unlikely to be falsely triggered, and using a Regex could make this a little more resilient to change. Probably YAGNI, though.

```ruby
ic["source"] =~ /@dependabot ignore/
```

landongrindheim (approved):

common/lib/dependabot/pull_request_creator/message_builder.rb
jurre: Yeah I suppose we could do that

PR #7791, opened 2023-08-11
Copy clone logic in dry-run.rb from the updater

deivid-rodriguez:
So that dry-run matches real behavior better.


jurre (approved):
With all this behavior copied over, it's not a DRY run 😸 

I will show myself out.

PR #7817, opened 2023-08-15
Do not attempt to group git dependencies as semver

jurre:
By definitions, git dependencies do not follow semver, as their version number is a hash, so these should never be included in the semver group

### DIFF
--- a/updater/lib/dependabot/updater/group_update_creation.rb
+++ b/updater/lib/dependabot/updater/group_update_creation.rb
@@ -205,6 +205,9 @@ module Dependabot
         # There are no group rules defined, so this dependency can be included in the group.
         return true unless group.rules["update-types"]
 
+        # git dependencies are not SemVer compatible so we cannot include them in the group
+        return false if git_dependency?(dependency)
+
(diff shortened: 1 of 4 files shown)


deivid-rodriguez (approved):
Looks good!

I only noticed the small inconsistency that contexts which use `vcr: { allow_unused_http_interactions: true }` are also tagged as `:vcr`. I think it's not necessary, so we could clean that up in a separate PR.

jakecoffman (approved):

updater/lib/dependabot/updater/group_update_creation.rb
```diff
@@ -238,6 +241,13 @@ def requirements_to_unlock(checker)
         end
       end
 
+      def git_dependency?(dependency)
+        GitCommitChecker.new(
+          dependency: dependency,
+          credentials: job.credentials
+        ).git_dependency?
```
jurre, line 248: I did not love having to instantiate this here, since _most_ `UpdateChecker`s have an instance of this memoized, but it is private and not all of them have it, so sorting through that became a bit of a mess, and I opted to just instantiate a new one, since instantiating objects is cheap, this is not a hot path.

Issue #5572, opened 2022-08-22
Run `yamllint` on PR's


jeffwidman:
We should run `yamllint` on PR's.

This may come for free if we started using the GitHub super linter, I'm not sure as I haven't looked into that.

abu2rki:
> We should run `yamllint` on PR's.
> 
> This may come for free if we started using the GitHub super linter, I'm not sure as I haven't looked into that.

PR #7818, opened 2023-08-16
Add `yamllint` to linters

jeffwidman:
We've got a lot of YAML in this repo... as this repo gets more popular, we'll benefit from having some consistency / linter in place.

Fix https://github.com/dependabot/dependabot-core/issues/5572


abdulapopoola (approved):

.github/workflows/smoke.yml
```diff
@@ -41,291 +41,291 @@ jobs:
...
-            - 'updater/**'
-            - 'npm_and_yarn/**'
-          'yarn-berry-workspaces':
-            - .github/workflows/smoke.yml
-            - .dockerignore
-            - Dockerfile.updater-core
-            - 'common/**'
-            - 'updater/**'
-            - 'npm_and_yarn/**'
+      - uses: actions/checkout@v3
```
jeffwidman, line 60: I configured `yamllint` to not try to enforce a particular style for indentation, as long as the file was internally consistent.

So here for example, the problem is that the indentation varied depending on the section of the file.

.yamllint.yaml
```diff
@@ -0,0 +1,26 @@
...
+    # We could downgrade the line length error to a warning, but for now it clutters the output too much.
+    # max: 120
+    # level: warning
+    disable
+
+ignore:
+  - "composer/helpers/v2/vendor/*"
+  - "github_actions/spec/fixtures/*"
+  - go_modules/spec/fixtures/projects/nested_vendor/nested/vendor/github.com/pkg/errors/appveyor.yml
+  - go_modules/spec/fixtures/projects/vendor/vendor/github.com/pkg/errors/appveyor.yml
```
jeffwidman, line 24: Where possible, I'd rather disable within the yaml file itself by adding: `# yamllint disable-file` to the top of the file. However, for both of these files, since they were vendored from upstream, I thought better to explicitly ignore here... that way we can easily rebuild the fixture w/o worrying about comment diffs.

.github/workflows/ci.yml
```diff
@@ -1,5 +1,5 @@
 name: Specs
-on:
+on:  # yamllint disable-line rule:truthy
```
jeffwidman, line 2: > `# yamllint disable-line rule:truthy`

This was annoying to disable on a case-by-case basis, but the rule looks actually helpful, so I left it enabled and just added inline comments.

PR #7819, opened 2023-08-16
Support security updates for NPM with exact requirements and no lockfile

deivid-rodriguez:


jurre (approved):
Yes that makes a lot of sense to me, thanks for tackling this!

deivid-rodriguez:
I fixed the failing test! 💪

PR #7826, opened 2023-08-17
Refactor poetry logic to parse subdependency types

deivid-rodriguez:
To unblock #7418.

### DIFF
--- a/python/lib/dependabot/python/file_parser/subdependency_type_parser.rb
+++ b/python/lib/dependabot/python/file_parser/subdependency_type_parser.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+require "toml-rb"
+require "dependabot/python/file_parser"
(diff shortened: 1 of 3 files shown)


python/lib/dependabot/python/file_parser/subdependency_type_parser.rb
```diff
@@ -0,0 +1,34 @@
...
+      class SubdependencyTypeParser
+        def initialize(lockfile:)
+          @lockfile = lockfile
+        end
+
+        def subdep_type(dep)
+          category =
+            TomlRB.parse(lockfile.content).fetch("package", []).
+            find { |dets| normalise(dets.fetch("name")) == dep.name }.
+            fetch("category")
```
jurre, line 19: Do we not also run into a key not found error on old lockfiles here?

jurre (approved):

PR #7829, opened 2023-08-17
fix ungrouped PRs being created due to errors during grouped update

jakecoffman:
This fixes an issue where when an error occurs during a grouped update, Dependabot considers the dependency as ungrouped. This is because for semver grouping we need to make external calls to decide if the dependency is in the group. When not using semver groups, we already know the grouping.

Also it was considering dependencies that didn't have updates to be ungrouped as well, and tried to make an individual PR in that case too.


jakecoffman:
Fixes a portion of https://github.com/dependabot/dependabot-core/issues/7830, but there's still some work to be done for semver updates to fully resolve the issue.

deivid-rodriguez (approved):

PR #7832, opened 2023-08-17
Fix yanked library problems in Poetry not detected when lockfile is present

deivid-rodriguez:
The tests updated in this PR were using a legacy `pyproject.lock` file, which is currently ignored by poetry 1.5. So the tests were unintentionally exercising the no-lockfile case and Dependabot was not behaving properly for the lockfile case.

This PR only fixes this bug but I will follow up with another PR to completely remove code dealing with `pyproject.lock` files.


abdulapopoola (approved):

jeffwidman:
> The tests updated in this PR were using a legacy `pyproject.lock` file, which is currently ignored by poetry 1.5

I trust ya, but when I tried to verify I couldn't find a commit/PR removing support for this... is this something that landed in `1.5` or has it been true for a while and we just didn't catch it when we updated `poetry`?

If you have a reference close at hand be great to include it, if not, no worries, I gave up after 5 mins of spelunking and decided it wasn't worth digging further.

jeffwidman (approved):

deivid-rodriguez:
Same as https://github.com/dependabot/dependabot-core/pull/7833#issuecomment-1683543411.

I didn't really check poetry upstream at all, just observed the current behavior of Poetry 1.5.1 which is the only version we use at the moment.

PR #7833, opened 2023-08-17
Remove code handling pyproject.lock files

deivid-rodriguez:
Poetry no longer supports them.


abdulapopoola (approved):
Thanks for cleaning all these up!

deivid-rodriguez:
This should get green once rebased over #7832.

jeffwidman (approved):
This is AWESOME that we can clean this up.

Be great to document (if you happen to know) a reference from upstream `poetry` that they've dropped support for this.

deivid-rodriguez:
I don't really know it.

The way I discovered this is when I started trying to run `poetry show --only main` for #7834, and seeing that for some specs this command would complain with an error "No poetry.lock was found". Then I saw that all the specs were actually running with a `pyproject.lock` file and that poetry was simply ignoring that and running all tests as if there was no lockfile. And that's only what made me find about the bug fixed by #7832.

PR #7834, opened 2023-08-17
Add support for Poetry 1.5 lockfiles

### LINKED ISSUE #7389: Poetry >=1.5.0 removes category from poetry.lock

### LINKED ISSUE #7641: Dependabot chokes on missing field in poetry.lock with poetry v1.5.0
Poetry 1.5.0 [removed](https://github.com/python-poetry/poetry/blob/e8ca4b32ceeb1648eaba6bed49e8b36dcfbf0cc0/CHANGELOG.md?plain=1#L62) the `category` field from `package` records in `poetry.lock`.

Dependabot still [expects](https://github.com/dependabot/dependabot-core/blob/919af326f7613ab39ae8362cef8a1b7890546d32/python/lib/dependabot/python/update_checker/poetry_version_resolver.rb#L304) this field to be present (`Hash#fetch` without a default value) and raises an exception if it's not.

deivid-rodriguez:
Since we upgraded to Poetry 1.5, we're failing to provide updates to any projects including Poetry 1.5 specific `poetry.lock` files, which have removed the "category" attribute from locked dependencies.

There's an open PR to fix this at #7418, however, I'm not fully convinced it's the right solution because:

* It no longer gets right the type (development/production) of sub dependencies.
* It still keeps the approach of looking into the lockfile directly, so we're subject to suffer from similar changes in the lockfile format in the future.
* It forces us to keep compatibility code if we don't want to hard fail on lockfiles created by Poetry >1.5, and it's complicated in general.

I think this PR provides a better solution which applies one of the principles we're trying to apply recently: delegate to the package manager as much as possible. So, we let `poetry show --only main` let us find the list of "production" dependencies.

Also, the logic is moved to the parser. If we get the dependency type right there, then the proper information is already carried out as `Dependency#production?`.

Fixes #7389.
Fixes #7641.
Supersedes #7418.


phillipuniverse:
@deivid-rodriguez I think the email might be the issue:

> Co-authored-by: Phillip Verheyden <phillip@Phillips-MBP.localdomain>

I think that should be:

> Co-authored-by: Phillip Verheyden <pverheyden@gmail.com>

deivid-rodriguez:
There you go, thanks!

python/lib/dependabot/python/helpers.rb
```diff
@@ -0,0 +1,34 @@
...
+module Dependabot
+  module Python
+    module Helpers
+      def self.run_poetry_command(command, fingerprint: nil)
+        start = Time.now
+        command = SharedHelpers.escape_command(command)
+        stdout, process = Open3.capture2e(command)
+        time_taken = Time.now - start
+
+        # Raise an error with the output from the shell session if Pipenv
```
jeffwidman, line 18: Hmm... I think `Pipenv` was a typo here, and it should actually be `poetry`...

However, nothing in this logic (that I can see) is `poetry` specific, so maybe this should be `run_python_package_manager_command()`??

As we may want to use this for `pipenv`, `pip-tools`, `pip`, etc?

python/lib/dependabot/python/file_parser/pyproject_files_parser.rb
```diff
@@ -154,21 +159,49 @@ def lockfile_dependencies
...
+
+            begin
+              output = Helpers.run_poetry_command("pyenv exec poetry show --only main")
+
+              output.split("\n").map { |line| line.split.first }
+            rescue SharedHelpers::HelperSubprocessFailed
+              # Sometimes, we may be dealing with an old lockfile that our
+              # poetry version can't show dependency information for. Other
+              # commands we use like `poetry update` are more resilient and
+              # automatically heal the lockfile. So we rescue the error an make
```
jeffwidman, line 196:
```suggestion
              # automatically heal the lockfile. So we rescue the error and make
```

python/lib/dependabot/python/file_parser/pyproject_files_parser.rb
```diff
@@ -154,21 +159,49 @@ def lockfile_dependencies
...
+          @production_dependency_names ||= parse_production_dependency_names
+        end
+
+        def parse_production_dependency_names
+          SharedHelpers.in_a_temporary_directory do
+            File.write(pyproject.name, pyproject.content)
+            File.write(lockfile.name, lockfile.content)
+
+            begin
+              output = Helpers.run_poetry_command("pyenv exec poetry show --only main")
```
jeffwidman, line 189: 👍 for using the poetry CLI rather than inspecting the file directly.

However, the docs for [`show` indicate that `--only main` and `--without dev` are both workable here](https://python-poetry.org/docs/cli/#show)... 

At first I was thinking that `--without dev` might be better here, as `main` is the default group but there may be other groups, but then I saw this note [here](https://python-poetry.org/docs/managing-dependencies/):
> Dependency groups, other than the implicit `main` group, must only contain dependencies you need in your development process. Installing them is only possible by using Poetry.

So I'm guessing docs/tests, etc are other groups that won't be considered part of `main`, but aren't part of `dev` either... 

The :dependabot: concept of dep categories isn't quite so sophisticated, we tend to simply have "prod" vs "dev", and in this method solely want prod deps, not "non-dev deps", so `--only main` makes sense to me!

jeffwidman (approved):
👍 Looks good to me other than possibly the helper can be made more generic. But that can be handled in a follow-on PR, if appropriate.

I'm going to deploy and merge this to unblock all the folks affected by it.

Sorry for the delay here, many thanks to @phillipuniverse for pushing on this repeatedly.

I didn't have bandwidth to look at this til we got through dropping `3.6`/`3.7`, but it does simplify things a bit now that we pin to solely to `poetry==1.5.1`.

BTW, researching all the background on this issue (I hadn't spent much time looking at `poetry` before today) resulted in some interesting dinner conversation:
> My wife: "Honey, what'd you do today?"
> Me: "Read a bunch about poetry"
> Her:  ???...

python/lib/dependabot/python/helpers.rb
jeffwidman:
```suggestion
        # Raise an error with the output from the shell session if Poetry
```

jeffwidman:
I tested in our staging environment using this new code against an old-style lockfile with the `category` key and a new-style without the `category` key and they both worked flawlessly. So I went ahead and pushed this live to production. If you hit problems, feel free to file an issue and reference this PR in that issue.

MJA645 (commented):
Does this affect me?

nantiferov:
Hi,

nantiferov:
Hi,

I'm not sure if I should open new issue for this, but seems that even after this released, Dependabot still removes `category` from lock file.

I tried it today on one of the repos, and PR has all categories removed, example:
image

phillipuniverse:
@nantiferov this is expected behavior. Dependabot does not have a way to specify a  reskin of Poetry to use (see #1556) and is currently using 1.5. When poetry.lock is re-created with Poetry 1.5, the category is always removed. The only fix really is to make sure that you are using Poetry 1.5 everywhere (locally, CI, etc) to ensure the category key stays removed.

This PR was about resolving an issue where if that category key was removed, Dependabot could not parse it correctly.

nantiferov:
Ah, ok, now it's clear. Thank you!

python/lib/dependabot/python/file_parser/pyproject_files_parser.rb
deivid-rodriguez: Yeah, in general I tend to consider "dev", everything that's not production/runtime, that's why I went with `--only main`.

PR #7835, opened 2023-08-17
Revert "Don't copy ruby version file into the updater image (#7802)"

deivid-rodriguez:
This reverts commit a89b3acef9b2965df6a3006851e8a5fc6a20fa41.

Turns out this is also used by rubocop in order to decide whether to apply cops that depend on specific ruby versions or not. Without this, running RuboCop inside a development shell does not run cleanly.


jeffwidman (approved):
I just spent 10 mins tracking this down myself and then saw your PR 😁 

May be helpful to have an inline comment saying `# used by rubocop`, but OTOH it wasn't that hard to track down the problem to this...

deivid-rodriguez:
Was going to add a comment, but ended up moving this again at https://github.com/dependabot/dependabot-core/pull/7841 😆

PR #7837, opened 2023-08-17
Fix `ecosystem_version` blowing up on deprecated python version

jeffwidman:
`LanguageVersionManager#python_major_minor` internally calls `#python_version_from_supported_versions` which meant that it blew up if the Python version is deprecated/unsupported.

So instead copy the major.minor truncation code into `ecosystem_version`.

I considered creating a helper and re-using it both places, but decided that since it's a relatively short/simple 1-liner that copy/pasting was fine. Jake would be proud of me for not forcing DRY. :grin:


deivid-rodriguez:
I _think_ this is no longer happening, so let me close this! Feel free to reopen if I got this wrong though, Jeff.

Issue #7416, opened 2023-06-09
Dependabot tries to bump PHP 8.2.6 to 8.3.0alpha1-fpm / 8.3.0beta2-fpm-bullseye


ruudk:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

Docker

### Language version

8.3

### Manifest location and content before the Dependabot update

8.2.6

### dependabot.yml content

```yaml
version: 2
updates:
- package-ecosystem: docker
  directory: "/etc/php-fpm"
  rebase-strategy: disabled
  ignore:
    - dependency-name: php
      versions: [">= 8.3.0"]
```

### What you expected to see, versus what you actually saw

8.2.7 (because we're on 8.2.6 and 8.2.7 is newly released)

8.3.0 we have excluded, but somehow that doesn't work for alpha releases?

ruudk:
Today it proposed this again:
> Bump php from 8.2.7-fpm-bullseye to 8.3.0alpha3-fpm-bullseye in /etc/php-fpm

This doesn't make any sense.

ruudk:
Today I received another bump:
> Bump php from 8.2.8-fpm-bullseye to 8.3.0beta2-fpm-bullseye in /etc/php-fpm

deivid-rodriguez:
Hei! Thanks for sticking with us and logging the different issues you're getting.

I tried to reproduce this at https://github.com/dsp-testing/dependabot-core-7416/pull/2, but I think I'm late because https://hub.docker.com/layers/library/php/8.2.9-fpm-bullseye/images/sha256-45b0a2a588f2d0589aff73d092c461019d39710787d7e658aff0a2d6a3511049?context=explore was pushed a few hours ago and things are working as expected again.

I'm adding some logging at #7842 to make this easier to figure out next time it happens.

Can you share the update logs and the exact configuration you are using? The configuration you posted does not seem valid because it doesn't define a `schedule`.

Thanks!

ruudk:
Thanks @deivid-rodriguez 💙 

Here's the full config:
```yaml
version: 2
updates:
- package-ecosystem: "github-actions"
  directory: "/"
  open-pull-requests-limit: 20
  schedule:
    interval: "daily"
    time: "06:00"
    timezone: Europe/Amsterdam
- package-ecosystem: docker
  directory: "/etc/php-fpm"
  schedule:
    interval: daily
    time: "06:00"
    timezone: Europe/Amsterdam
[...]

PR #7842, opened 2023-08-18
Debug issues with docker prereleases

deivid-rodriguez:
This doesn't fix anything but I expect it to make it easier to investigate issues like https://github.com/dependabot/dependabot-core/issues/7416.


Nishnha (approved):
Thanks for making this assumption more explicit and logging it

PR #7843, opened 2023-08-18
Simplify development images

deivid-rodriguez:
Right now, if you jump into a development shell, you'll notice this:

```
[dependabot-core-dev] ~/dependabot-core $ ls ../docker/
dependabot-docker.gemspec  Dockerfile  Gemfile  lib  README.md  script  spec
[dependabot-core-dev] ~/dependabot-core $ ls docker/
dependabot-docker.gemspec  Gemfile  lib  script  spec
```

This seems a bit strange.

I think we can use the same structure as in the production updater image for development, simplifying also the Dockerfile and making the image a bit thinner (~ 50Mb).

### DIFF
--- a/Dockerfile.development
+++ b/Dockerfile.development
@@ -1,6 +1,8 @@
 ARG FROM_IMAGE=ghcr.io/dependabot/dependabot-updater-core
 FROM $FROM_IMAGE
 
+ARG ECOSYSTEM
+
 # Temporarily switch to root user in order to install packages
 USER root
 RUN apt-get update \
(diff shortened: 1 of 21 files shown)


landongrindheim (approved):
Neat! 😄

deivid-rodriguez:
Let me know if you find any issues!

Nishnha:
There was an issue on my machine where some ecosystems worked correctly and contained the `lib/` and `spec/` directories while other ecosystems only contained the `Gemfile` and `.gemspec` directories.

It turns out I had existing `dependabot/dependabot-core-development-*` images on my machine, causing the docker-dev-shell not to rebuild them.

After deleting the image with `docker image rm dependabot/dependabot-core-development-bundler`, I was able to create a new, working image with `bin/docker-dev-shell updater`.

Hope this helps anyone else running into a similar issue

PR #7850, opened 2023-08-19
Fix encoding option value for gitlab commit creation

andrcuns:
This fixes broken commit creation for gitlab that was introduced by: https://github.com/dependabot/dependabot-core/pull/7381

The issue here is that gitlab api only supports 2 encoding values, `base64` and `text`. Current implementation will try to pass `utf-8` as indicated by unit tests which will raise an error like: `Server responded with code 400, message: actions[0][encoding] does not have a valid value, actions[1][encoding] does not have a valid value`.

Gitlab api doc for refernce: https://docs.gitlab.com/ee/api/commits.html#create-a-commit-with-multiple-files-and-actions

### DIFF
--- a/common/lib/dependabot/pull_request_creator/gitlab.rb
+++ b/common/lib/dependabot/pull_request_creator/gitlab.rb
@@ -99,30 +99,8 @@ module Dependabot
           source.repo,
           branch_name,
           commit_message,
-          file_actions
+          files
         )
-      end
-
-      def file_actions
-        files.map do |file|
-          {
-            action: file_action(file),
-            file_path: file.type == "symlink" ? file.symlink_target : file.path,
-            content: file.content,
-            encoding: file.content_encoding == Dependabot::DependencyFile::ContentEncoding::BASE64 ? "base64" : "text"
-          }
-        end
-      end
-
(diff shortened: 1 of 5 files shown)


common/lib/dependabot/pull_request_creator/gitlab.rb
```diff
@@ -109,7 +109,7 @@ def file_actions
             action: file_action(file),
             file_path: file.type == "symlink" ? file.symlink_target : file.path,
             content: file.content,
-            encoding: file.content_encoding
+            encoding: file.content_encoding == Dependabot::DependencyFile::ContentEncoding::BASE64 ? "base64" : "text"
```
andrcuns, line 112: I chose to not pass constant directly and hardcode the string instead because technically the `BASE64` constant could be renamed in `dependabot-core` or `gitlab` api might introduce a differently named option in next version of api for example.

It would be better to have common method to return encoding option in both creator and updater class, but not sure where to place it, we don't seem to have a common module just for gitlab stuff.

andrcuns: That makes sense 👍 I moved the file action mapping to client proxy and removed duplication

PR #7863, opened 2023-08-21
Bump `elixir` to `1.15.4`

jeffwidman:
Release notes:
* https://github.com/elixir-lang/elixir/releases/tag/v1.15.0
* https://github.com/elixir-lang/elixir/releases/tag/v1.15.4

### DIFF
--- a/hex/Dockerfile
+++ b/hex/Dockerfile
@@ -19,8 +19,8 @@ RUN apt-get update \
 
 # Install Elixir
 # https://github.com/elixir-lang/elixir/releases
-ARG ELIXIR_VERSION=v1.14.4
-ARG ELIXIR_CHECKSUM=b5705275d62ce3dae172d25d7ea567fffdabb45458abeb29c0189923b907367c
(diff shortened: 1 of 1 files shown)


jeffwidman:
While on the surface this seems relatively low priority, this is actually the current end of a string I've been tugging at the last few hours... 
1. Ubuntu 22.04 upgrade requires us to switch away from the ESOL PPA
2. The RabbitMQ Erlang PPA looks like a solid alternative
3. It only has the latest Erlang / Elixir versions... while we will probably stick with installing Elixir straight from GitHub releases, I'd feel better if we first updated to the current Elixir version so that if anything starts going awry it's easier to track down the source of the problem and not stuck wondering if it's a version bump vs PPA change...

jeffwidman:
Test failures are because the ESOL PPA is being flaky...

jeffwidman:
Smoke test failure looks a little concerning... worth digging into.

jeffwidman:
Superseded by:
* https://github.com/dependabot/dependabot-core/pull/8622

PR #7865, opened 2023-08-22
Switch to RabbitMQ Erlang PPA

jeffwidman:
This is mostly WIP for now, it has a few CI errors that would need to be worked through. Putting up for now mostly as a record of how it could look to go this direction.

But we may not need to use the full apt-get style stuff, since these are throwaway docker images, and don't have a lot of dependencies, we may be able to get away with directly downloading the underlying file and not needing the full blown `apt` stuff...

### DIFF
--- a/hex/Dockerfile
+++ b/hex/Dockerfile
@@ -3,22 +3,56 @@ RUN apt-get update \
   && apt-get upgrade -y \
   && apt-get install -y --no-install-recommends \
     gnupg2
-# Install Erlang and Elixir
-ENV PATH="$PATH:/usr/local/elixir/bin"
+
+# Install Erlang
+ARG ERLANG_MAJOR_VERSION=24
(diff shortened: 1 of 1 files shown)


jeffwidman:
There are a couple of failing specs that would be better if someone more familiar with Erlang dug into... instead, for now to unblock the Ubuntu 22.04 upgrade, I switched to the default Erlang installs from the 22.04 PPA:
* https://github.com/dependabot/dependabot-core/pull/5030

Hopefully an Erlang user from the open source :dependabot: community can swing by and put up a PR taking this across the finish line for us.

Alternatively, if we wait a few more months, the new Erlang Solutions PPA that they're setting up may start working, or they may get it donated to the Erlang foundation: https://erlangforums.com/t/the-eef-is-looking-for-volunteers-to-take-over-esls-build-packages/2238

PR #7862, opened 2023-08-21
Regenerate some lockfiles with Poetry 1.5

deivid-rodriguez:
Otherwise some specs fail depending on order:

```
[dependabot-core-dev] ~/python $ rspec spec/dependabot/python/file_parser/pyproject_files_parser_spec.rb:145
Run options: include {:locations=>{"./spec/dependabot/python/file_parser/pyproject_files_parser_spec.rb"=>[145]}}

Randomized with seed 52409
F

Failures:

  1) Dependabot::Python::FileParser::PyprojectFilesParser parse poetry files with a lockfile with a path dependency excludes the path dependency
     Failure/Error: dep.name if dep.production?

     RuntimeError:
       Unsupported package_manager pip
     # /home/dependabot/common/lib/dependabot/dependency.rb:15:in `production_check_for_package_manager'
     # /home/dependabot/common/lib/dependabot/dependency.rb:97:in `production?'
     # ./lib/dependabot/python/file_parser/pyproject_files_parser.rb:199:in `block (2 levels) in parse_production_dependency_names'
     # ./lib/dependabot/python/file_parser/pyproject_files_parser.rb:198:in `each'
     # ./lib/dependabot/python/file_parser/pyproject_files_parser.rb:198:in `filter_map'
     # ./lib/dependabot/python/file_parser/pyproject_files_parser.rb:198:in `rescue in block in parse_production_dependency_names'
     # ./lib/dependabot/python/file_parser/pyproject_files_parser.rb:188:in `block in parse_production_dependency_names'
     # /home/dependabot/common/lib/dependabot/shared_helpers.rb:55:in `block in in_a_temporary_directory'
     # /home/dependabot/common/lib/dependabot/shared_helpers.rb:55:in `chdir'
     # /home/dependabot/common/lib/dependabot/shared_helpers.rb:55:in `in_a_temporary_directory'
     # ./lib/dependabot/python/file_parser/pyproject_files_parser.rb:184:in `parse_production_dependency_names'
     # ./lib/dependabot/python/file_parser/pyproject_files_parser.rb:180:in `production_dependency_names'
     # ./lib/dependabot/python/file_parser/pyproject_files_parser.rb:171:in `block in lockfile_dependencies'
     # ./lib/dependabot/python/file_parser/pyproject_files_parser.rb:159:in `each'
     # ./lib/dependabot/python/file_parser/pyproject_files_parser.rb:159:in `lockfile_dependencies'
[...]

PR #7866, opened 2023-08-22
Update poetry version to 1.6.1

noorul:
The latest version of poetry is 1.6.1. Since we already dropped support for Python 3.7, we can move to this version.


jeffwidman (commented):
LGTM!

Two test failures look like you need to regenerate those lockfiles... in this case, I think it'd be fine to manually change those two lines by hand as the release notes don't indicate any other changes happened.

Otherwise, see https://github.com/dependabot/dependabot-core/pull/7862#issuecomment-1687364394

jeffwidman (approved):
Thanks again!

PR #7867, opened 2023-08-22
fix edge cases during semver grouping creating single PRs erroneously

jakecoffman:
This pushes the "handled" logic into the group so that the `compile_updates_for` method can specify that the dependency is handled immediately when it is found to be in the semver grouping. This way, if there is an error or the update is not possible, it will still be considered handled and not attempt to create a single PR for it.

Also, if an exception is thrown, even during version fetching before we can know whether it's in the grouping or not, we consider the dependency handled as well since it's not known what versions are available. Customers can request a rebase or recreate to try again and not have superfluous PRs kicking about.


common/lib/dependabot/dependency_group.rb
```diff
@@ -9,12 +9,21 @@
...
 
     def initialize(name:, rules:)
       @name = name
       @rules = rules
       @dependencies = []
+      @handled_dependencies = Set.new
+    end
+
+    def add_to_handled(*dependencies)
+      @handled_dependencies += dependencies.map(&:name)
```
jurre, line 22: 🤔 is it sufficient to store just the `name`? Some package managers, like npm, allow having multiple versions of the same dependency installed.

In that case, if they're defined in the `package.json`, I do you need to name them explicitly, and that distinct name is likely what we get here, for multiple versions in the `package-lock.json` (transitive dependencies), this likely ends up not being an issue?

updater/lib/dependabot/dependency_snapshot.rb
```diff
@@ -69,16 +69,14 @@ def groups
...
     def ungrouped_dependencies
       # If no groups are defined, all dependencies are ungrouped by default.
       return allowed_dependencies unless groups.any?
-      return @ungrouped_dependencies if defined?(@ungrouped_dependencies)
 
-      @dependency_group_engine.ungrouped_dependencies
+      # Otherwise return dependencies that haven't been handled during the group update portion.
+      all_handled_dependencies = Set.new
+      groups.each { |group| all_handled_dependencies += group.handled_dependencies }
+      allowed_dependencies.reject { |dep| all_handled_dependencies.include?(dep.name) }
```
jurre, line 79: Would this work?

```suggestion
      allowed_dependencies.reject { |dep| groups.any? { |group| group.handled_dependencies.include?(dep.name) } }
```

common/lib/dependabot/dependency_group.rb
jakecoffman: Hmmm good question. Is it currently possible for Dependabot to handle (create, update) multiple PRs for the same dependency name? I would think not, at least based on how the [job dependencies](https://github.com/dependabot/dependabot-core/blob/23de1c7583117bd15f26ab33482383ed57208f14/updater/lib/dependabot/job.rb#L89) property for rebasing contains only the name and not some kind of unique identifier.

common/lib/dependabot/dependency_group.rb
```diff
@@ -9,12 +9,21 @@
...
+    end
+
+    def add_all_to_handled
+      @handled_dependencies += dependencies.map(&:name)
     end
```
jurre, line 27: I don't feel strongly, but what if we had a `contains_or_handled?` method that we could call on the groups? Then we don't need to explicitly `add_to_handled` if we already add the dependency to the group?

common/lib/dependabot/dependency_group.rb
jurre: Yep, I think you're right 👍

common/lib/dependabot/dependency_group.rb
jakecoffman: But we only want to consider all the dependencies in the group as handled when there are no semver groups, or if a PR already exists for the group. 

Essentially we're marking which dependencies that are _potentially_ in the group as _actually_ in the group as we proceed through this process.

I'm tempted to rename `group.dependencies` to `group.potentially_grouped_dependencies` to make that clear.

updater/lib/dependabot/dependency_snapshot.rb
jakecoffman: I wrote it this way first pass, but it seemed a bit unclear. I liked being able to name `all_handled_dependencies`. Now that I'm looking at it again, I kind of like this instead:

```ruby
      all_handled_dependencies = Set.new(groups.flat_map(&:handled_dependencies))
      allowed_dependencies.reject { |dep| all_handled_dependencies.include?(dep.name) }
```

I'm trying not to code golf it, but also keep it brief. Also the Set prevents any kind of silly `O(n^2)` for our monorepo friends.

jurre: Yep, makes sense 👍

jakecoffman:
I'm a bit confused now. The specs have shown an issue and now I'm seeing both groups and dependency_snapshots have `dependencies`. It seems like the snapshot has the dependencies that are already filtered on by patterns, and the group dependencies are populated as the update proceeds. So I need to switch things up a bit 🚧

jakecoffman:
I need to add a test and then this will be ready to go.

updater/spec/dependabot/updater/operations/group_update_all_versions_spec.rb
```diff
@@ -326,6 +326,29 @@
     end
   end
 
+  context "when there are semver rules but an error occurs gathering versions" do
+    before do
+      allow_any_instance_of(Dependabot::Bundler::UpdateChecker).
+        to receive(:latest_version).
+          and_raise(StandardError.new("Test error while getting latest version"))
```
jurre, line 333:
```suggestion
        and_raise(StandardError.new("Test error while getting latest version"))
```

The linter is upset about this

PR #7869, opened 2023-08-22
Bump to `ruby` `3.2`

jeffwidman:
Bump to `ruby` `3.2` now that we're on a more modern Ubuntu version.

I followed the example of:
* https://github.com/dependabot/dependabot-core/pull/6257


bundler/helpers/v2/monkey_patches/definition_ruby_version_patch.rb
```diff
@@ -10,7 +10,7 @@ def source_requirements
         Gem::Specification.new("Ruby\0", requested_version)
     end
 
-    %w(2.5.3 2.6.10 2.7.7 3.0.5 3.2.1).each do |version|
+    %w(2.5.3 2.6.10 2.7.7 3.0.5 3.1.4 3.2.2).each do |version|
```
jeffwidman, line 13: I realize technically we only need the latest patch version of the prior releases, but here we'd actually needed the version `3.2` previously when we were on `3.1`, so I thought simplest to just treat it as "note all ruby versions"` as it keeps life simple.

I do wonder if we should be dropping the old Ruby 2 versions... but that's out of scope for this PR.

jeffwidman:
Throwing this up to see what CI thinks...

common/dependabot-common.gemspec
```diff
@@ -20,8 +20,8 @@ Gem::Specification.new do |spec|
   }
 
   spec.version = Dependabot::VERSION
-  spec.required_ruby_version = ">= 3.1.0"
-  spec.required_rubygems_version = ">= 3.3.7"
+  spec.required_ruby_version = ">= 3.2.0"
```
jeffwidman, line 23: I realize this is a little opinionated given this is a library, but still, I think it's not too unreasonable... `3.2` has been out for a while and wasn't that large of a jump from `3.1`, so shouldn't be that hard for folks to upgrade their toolchains unless they're airgapped/really monolithic environments with build systems that are super coupled... but I generally doubt folks run Dependabot tightly coupled, since they can run Dependabot as a sidecar script using it's own mini environment.

common/dependabot-common.gemspec
```diff
@@ -20,8 +20,8 @@ Gem::Specification.new do |spec|
...
+  spec.required_rubygems_version = ">= 3.3.7" # TODO: bump this to whatever version is bundled with ruby 3.2
```
jeffwidman, line 24: Not sure of the fastest way to figure this out? Not seeing anything obvious in the CI logs when I grep them for "rubygem"

jeffwidman:
Oh, I forgot we already had:
* https://github.com/dependabot/dependabot-core/pull/6436

@blue2cat is a friend of mine, and I happen to know he's been pretty busy, so will leave this open for a little while to work through CI although I'll be sure to credit him as co-author on the commit...

jeffwidman:
Closing as I forgot this is all waiting on getting through the Bundler 1 conflict... we can either re-open and fix later, or work through it all over in #6436

PR #7876, opened 2023-08-23
Don't double-install packages required for building Python.

jeffwidman:
I don't see a reason to double-install these apt packages. If we instead change the base for `python` `3.11` to be `python-core`, then we pick this up for free. It won't be cheaper in wall clock time since the jobs are parallel, but it saves network bandwidth/CPU/IOPS.

More importantly, it reduces the chance of developer mistakes if we accidentally add or remove a package from one place and forget to copy/paste that change.

### DIFF
--- a/python/Dockerfile
+++ b/python/Dockerfile
@@ -55,11 +55,11 @@ USER dependabot
 ENV DEPENDABOT_NATIVE_HELPERS_PATH="/opt"
 RUN git -c advice.detachedHead=false clone https://github.com/pyenv/pyenv.git --branch $PYENV_VERSION --single-branch --depth=1 /usr/local/.pyenv
 
-FROM python-core as python-3.10
-RUN pyenv install $PY_3_10 \
-  && bash /opt/python/helpers/build $PY_3_10 \
+FROM python-core as python-3.8
+RUN pyenv install $PY_3_8 \
+  && bash /opt/python/helpers/build $PY_3_8 \
   && cd /usr/local/.pyenv \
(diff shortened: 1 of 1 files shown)


python/Dockerfile
```diff
@@ -14,6 +14,8 @@ ARG PY_3_8
 ARG PYENV_VERSION
 USER root
 
+# Install *only* the apt packages required for this builder image to build Python.
+# C-libs needed by users to build their Python packages should be installed down below in the final docker image.
```
jeffwidman, line 18: This adds a c-lib here:
* https://github.com/dependabot/dependabot-core/pull/7875

Did it as a separate PR to prove CI is okay with this first step individually, separate from the additional install.

python/Dockerfile
```diff
@@ -55,69 +55,25 @@ USER dependabot
 ENV DEPENDABOT_NATIVE_HELPERS_PATH="/opt"
 RUN git -c advice.detachedHead=false clone https://github.com/pyenv/pyenv.git --branch $PYENV_VERSION --single-branch --depth=1 /usr/local/.pyenv
 
-FROM python-core as python-3.10
-RUN pyenv install $PY_3_10 \
-  && bash /opt/python/helpers/build $PY_3_10 \
+FROM python-core as python-3.8
```
jeffwidman, line 58: This was a simple re-ordering as `3.11` was already at the bottom, so easier to go oldest to newest, rather than newest to oldest, and then back to newest.

PR #7881, opened 2023-08-23
Use dependency-type and semver grouping for dev dependencies

jurre:
Instead of specifying individual dependencies, we can now leverage the dependency-type config option for these, and while we're at it, let's ensure we don't include major updates in these groups, as they tend to be harder to merge and addressing them separately is usually the better option.


deivid-rodriguez (approved):
Nice to not have to hardcode dependency names anymore!

PR #7884, opened 2023-08-23
fixes toolchain directive getting into go.mod

### LINKED ISSUE #7868: gomod: Bumping a dependency that sets "go 1.21" in their go.mod sets "toolchain 1.21" in mine

### Package ecosystem

gomod

### Package manager version

1.20

### Language version

1.20

go.mod. Link to original module file: https://github.com/boinkor-net/tsnsrv/blob/469a5104d7914816d185fec37a031cc47b67f032/go.mod

### dependabot.yml content

https://github.com/boinkor-net/tsnsrv/blob/469a5104d7914816d185fec37a031cc47b67f032/.github/dependabot.yml

### Updated dependency

"tailscale" from v1.46.1 to v1.48.1

I was expecting this PR to only contain the version bump; however, it also added this line: https://github.com/boinkor-net/tsnsrv/pull/25/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R4

in diff form:

```diff
diff --git a/go.mod b/go.mod
index a6d61d9..73b1321 100644
--- a/go.mod
+++ b/go.mod
@@ -1,13 +1,14 @@
 module github.com/antifuchs/tsnsrv
 
 go 1.20
+toolchain go1.21.0
 
 require (
 	github.com/peterbourgon/ff/v3 v3.4.0
 	github.com/prometheus/client_golang v1.16.0
 	github.com/stretchr/testify v1.8.4
-	golang.org/x/exp v0.0.0-20230425010034-47ecfdc1ba53
-	tailscale.com v1.46.1
+	golang.org/x/exp v0.0.0-20230725093048-515e97ebf090
+	tailscale.com v1.48.1
 )
 
 require (
@@ -32,6 +33,7 @@ require (
 	github.com/cespare/xxhash/v2 v2.2.0 // indirect
 	github.com/coreos/go-iptables v0.6.0 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/dblohm7/wingoes v0.0.0-20230803162905-5c6286bb8c6e // indirect
 	github.com/fxamacker/cbor/v2 v2.4.0 // indirect
 	github.com/go-ole/go-ole v1.2.6 // indirect
 	github.com/godbus/dbus/v5 v5.1.0 // indirect
```

note the "toolchain" line that directly contradicts the `go` line above it. Go 1.20 does not understand the "toolchain" keyword, and so the resulting PR build fails.

(aside, yep, apparently it's fine to link the tailscale library into a go 1.20 library, but yeah, I should upgrade that version as soon as my toolchain supports it.)

_No response_

Find the PR at https://github.com/boinkor-net/tsnsrv/pull/25

_No response_

jakecoffman:
Fixes #7868

I had to read https://tip.golang.org/doc/toolchain#select a few times to understand this so let me summarize:

With `GOTOOLCHAIN` unset, Go 1.21 will use the locally installed toolchain (1.21.0 currently) if the go directive is <= 1.21. This means it was injecting the new toolchain directive and updating the go directive to 1.21 in older `go.mod`s, but Dependabot code reverts the change to the go directive, which is how we ended up with #7868. This only affected go.mod files that had dependencies that were on 1.21. 

By switching to `GOTOOLCHAIN="go1.20+auto"` we revert to using Go 1.20 to retain the current behavior for `go.mod` with go directive <=1.20, but the `auto` means any `go.mod` that are on newer versions will automatically select the right Go toolchain and download it if not installed locally.

Thus technically this is the last bump we ever need to do with the Go version. Although to prevent each job from downloading the Go toolchain on-the-fly we can pre-install multiple popular versions.


go_modules/spec/dependabot/go_modules/file_updater_spec.rb
```diff
@@ -127,6 +127,15 @@ module declares its path as: go.etcd.io/bbolt
       end
     end
 
+    context "pre 1.21 go.mod that uses a 1.21 dependency" do
+      let(:project_name) { "toolchain" }
+      let(:files) { [go_mod] }
+
+      it "doesn't add a toolchain directive" do
+        expect(updated_files.first&.content).to_not include("toolchain")
```
landongrindheim, line 135: 🔎 I think we should omit `&.` here as it will pass `nil` to the `include` matcher. I think the resulting `NoMethodError` for `#content` would be a little easier to reason about as it's closer to the code.

```suggestion
        expect(updated_files.first.content).to_not include("toolchain")
```

landongrindheim (approved):

PR #7888, opened 2023-08-23
ensure required arrays always exists

jakecoffman:
Small cleanup: This data is always sent from the service and set by the CLI so we can remove some TODOs.


honeyankit (approved):

landongrindheim (approved):

jakecoffman:
D'oh looks like the integration tests don't have the key. I'll have to add them.

jakecoffman:
I'll do this some other time. It's currently working, why mess with it?

PR #7890, opened 2023-08-23
feat: Updates Hex to 2.0.6

isaacsanders:
Closes https://github.com/dependabot/dependabot-core/issues/6009


hex/helpers/lib/run.exs
```diff
@@ -118,7 +118,7 @@ defmodule DependencyHelper do
   end
 
   defp fetch_public_key(repo, repo_url, auth_key, fingerprint) do
-    case Hex.Repo.get_public_key(repo_url, auth_key) do
+    case Hex.Repo.get_public_key(%{trusted: true, url: repo_url, auth_key: auth_key}) do
```
jeffwidman, line 121: What does `trusted: true` mean?

I see these params are used similarly here: https://github.com/hexpm/hex/blob/1881f9fe8e0571ba7fdcfc86ecf484913125dc37/lib/mix/tasks/hex.repo.ex#L218-L220

I'm not an Erlang/Hex expert, but typically in a package manager an arg like "trusted" implies a private/insecure registry of some kind...

jeffwidman (commented):
Thanks for this, I'm super excited you were willing to figure this out and get it resolved.

However, I'm a bit surprised that the only fix needed to get the tests passing was the change to `version_resolver.rb`, because when I attempted the upgrade, the breaking change in Hex 2 to [print out the detailed error message was breaking the message passing across `stdio`](https://github.com/dependabot/dependabot-core/pull/6026#issuecomment-1336276793).

How were you able to workaround that? Was the problem simply that the public key in `run.exs` needed to be configured differently?

What was I misunderstanding?

hex/helpers/lib/run.exs
landongrindheim: I _think_ Elixir will pattern match this, so `%{trusted: false` would cause this to not match.

isaacsanders: Trusted isn’t used in a clear way. I will need more time to dissect it in the code to give a better answer.

isaacsanders:
> Thanks for this, I'm super excited you were willing to figure this out and get it resolved.
> 
> However, I'm a bit surprised that the only fix needed to get the tests passing was the change to `version_resolver.rb`, because when I attempted the upgrade, the breaking change in Hex 2 to [print out the detailed error message was breaking the message passing across `stdio`](https://github.com/dependabot/dependabot-core/pull/6026#issuecomment-1336276793).
> 
> How were you able to workaround that? Was the problem simply that the public key in `run.exs` needed to be configured differently?
> 
> What was I misunderstanding?

There was an exception occurring due to a pattern mismatch. It caused the io at the end to not occur.

isaacsanders:
@jeffwidman Can we get this merged?

hex/helpers/lib/run.exs
isaacsanders: `trusted: true` appears to indicate that there is a fingerprint that is being used to verify the key.

isaacsanders: Since we are using a fingerprint to verify the key, that should mean that our use of the `trusted` key here matches the intended use in the code.

jakecoffman (approved):
Sorry for the delay, thanks for the fix!

PR #7891, opened 2023-08-23
Produce smaller Python builds

tvalenta:
This PR seeks to reduce the size of the Python image:
* LD Flags prevent symbols from being added
* The helper script deletes the extra pyc and .a files

This doesn't entirely produce identical layers because the tar files still have differing timestamps. Thoughts for changing that behaviour include passing a consistent date to `--mtime` as an argument to `tar(1)` or changing the output itself, in the build helper
```
find /usr/local/.pyenv -exec touch -d '2000-01-01' -h {} +
```

rspec tests passed:
```

Finished in 11 minutes 50 seconds (files took 1.1 seconds to load)
1149 examples, 0 failures, 19 pending

Randomized with seed 25745
```

### DIFF
--- a/python/Dockerfile
+++ b/python/Dockerfile
@@ -45,32 +45,28 @@ RUN git -c advice.detachedHead=false clone https://github.com/pyenv/pyenv.git --
 
 FROM python-core as python-3.8
 # GCC - Do not add symbols to objects
-ENV LDFLAGS="-Wl,--strip-all"
-RUN pyenv install $PY_3_8 \
+RUN env LDFLAGS_NODIST="-Wl,--strip-all" pyenv install $PY_3_8 \
   && bash /opt/python/helpers/build $PY_3_8 \
   && cd /usr/local/.pyenv \
   && tar czf 3.8.tar.gz versions/$PY_3_8
 
 FROM python-core as python-3.9
 # GCC - Do not add symbols to objects
-ENV LDFLAGS="-Wl,--strip-all"
-RUN pyenv install $PY_3_9 \
+RUN env LDFLAGS_NODIST="-Wl,--strip-all" pyenv install $PY_3_9 \
   && bash /opt/python/helpers/build $PY_3_9 \
   && cd /usr/local/.pyenv \
   && tar czf 3.9.tar.gz versions/$PY_3_9
 
 FROM python-core as python-3.10
 # GCC - Do not add symbols to objects
-ENV LDFLAGS="-Wl,--strip-all"
-RUN pyenv install $PY_3_10 \
+RUN env LDFLAGS_NODIST="-Wl,--strip-all" pyenv install $PY_3_10 \
   && bash /opt/python/helpers/build $PY_3_10 \
   && cd /usr/local/.pyenv \
   && tar czf 3.10.tar.gz versions/$PY_3_10
 
 FROM python-core
 # GCC - Do not add symbols to objects
-ENV LDFLAGS="-Wl,--strip-all"
-RUN pyenv install $PY_3_11 \
+RUN env LDFLAGS_NODIST='-Wl,--strip-all' pyenv install $PY_3_11 \
   && pyenv global $PY_3_11 \
   && bash /opt/python/helpers/build $PY_3_11
 
--- a/python/Dockerfile
+++ b/python/Dockerfile
@@ -44,24 +44,32 @@ ENV DEPENDABOT_NATIVE_HELPERS_PATH="/opt"
 RUN git -c advice.detachedHead=false clone https://github.com/pyenv/pyenv.git --branch $PYENV_VERSION --single-branch --depth=1 /usr/local/.pyenv
 
 FROM python-core as python-3.8
+# GCC - Do not add symbols to objects
+ENV LDFLAGS="-Wl,--strip-all"
 RUN pyenv install $PY_3_8 \
   && bash /opt/python/helpers/build $PY_3_8 \
   && cd /usr/local/.pyenv \
   && tar czf 3.8.tar.gz versions/$PY_3_8
 
 FROM python-core as python-3.9
+# GCC - Do not add symbols to objects
+ENV LDFLAGS="-Wl,--strip-all"
 RUN pyenv install $PY_3_9 \
   && bash /opt/python/helpers/build $PY_3_9 \
   && cd /usr/local/.pyenv \
   && tar czf 3.9.tar.gz versions/$PY_3_9
 
 FROM python-core as python-3.10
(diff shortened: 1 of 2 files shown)


tvalenta:
@jeffwidman I see you've been working to shrink the size of the Python image. Wondering if this PR fits in with your efforts. If not, I can close it for another time. Or, if you would wish for potentially consistent tarfiles (with the timestamps being consistent across builds) I could put in some extra effort to make that work.

Nishnha:
Thanks @tvalenta. Do you happen to know how much this would shrink the python builds/dependabot python image by?

tvalenta:
> 

I've been testing with podman on macOS arm, so the sizes are slightly different, but the most recent image with the adjustments:
```
dependabot@aa95902c26d7:/usr/local/.pyenv$ du -shc versions *gz
108M	versions
31M	3.10.tar.gz
33M	3.8.tar.gz
30M	3.9.tar.gz
201M	total
```

Today's image:
```
dependabot@36d68fa332eb:/usr/local/.pyenv$ du -shc versions *gz
415M	versions
89M	3.10.tar.gz
86M	3.8.tar.gz
85M	3.9.tar.gz
674M	total
```

If I extract all of the tar files, I'm still ahead:
```
dependabot@aa95902c26d7:/usr/local/.pyenv$ for name in *gz; do tar xf $name; done
dependabot@aa95902c26d7:/usr/local/.pyenv$ du -shc versions/*
105M	versions/3.10.12
108M	versions/3.11.4
106M	versions/3.8.17
103M	versions/3.9.17
421M	total
```

Nishnha:
That's a big win!

I found an active issue where these flags can leak from pyenv and cause collisions https://github.com/pyenv/pyenv/issues/2204

I'm not sure how/if this would affect Dependabot, but we use pyenv in a few other places so I'd err on the side of caution. There is an alternative flag, `LDFLAGS_NODIST`, that we can set instead to avoid the issue.

tvalenta:
I can try those variables and retest.

tvalenta:
```
dependabot@7a99005e7999:/usr/local/.pyenv$ for name in *gz; do tar xf $name; done
dependabot@7a99005e7999:/usr/local/.pyenv$ for name in versions/*; do echo $name ;  $name/bin/python3 -m sysconfig|grep strip ; done
versions/3.10.13
	CONFIGURE_LDFLAGS_NODIST = "-Wl,--strip-all"
	PY_CORE_LDFLAGS = "-L/usr/local/.pyenv/versions/3.10.13/lib -Wl,-rpath,/usr/local/.pyenv/versions/3.10.13/lib -L/usr/local/.pyenv/versions/3.10.13/lib -Wl,-rpath,/usr/local/.pyenv/versions/3.10.13/lib -Wl,--strip-all -Wl,--strip-all"
	PY_LDFLAGS_NODIST = "-Wl,--strip-all -Wl,--strip-all"
versions/3.11.5
	CONFIGURE_LDFLAGS_NODIST = "-Wl,--strip-all"
	PY_CORE_LDFLAGS = "-L/usr/local/.pyenv/versions/3.11.5/lib -Wl,-rpath,/usr/local/.pyenv/versions/3.11.5/lib -L/usr/local/.pyenv/versions/3.11.5/lib -Wl,-rpath,/usr/local/.pyenv/versions/3.11.5/lib -Wl,--strip-all -Wl,--strip-all"
	PY_LDFLAGS_NODIST = "-Wl,--strip-all -Wl,--strip-all"
	PY_LDFLAGS_NOLTO = "-L/usr/local/.pyenv/versions/3.11.5/lib -Wl,-rpath,/usr/local/.pyenv/versions/3.11.5/lib -L/usr/local/.pyenv/versions/3.11.5/lib -Wl,-rpath,/usr/local/.pyenv/versions/3.11.5/lib  -Wl,--strip-all"
versions/3.8.18
	CONFIGURE_LDFLAGS_NODIST = "-Wl,--strip-all"
	PY_CORE_LDFLAGS = "-L/usr/local/.pyenv/versions/3.8.18/lib -Wl,-rpath,/usr/local/.pyenv/versions/3.8.18/lib -L/usr/local/.pyenv/versions/3.8.18/lib -Wl,-rpath,/usr/local/.pyenv/versions/3.8.18/lib -Wl,--strip-all -Wl,--strip-all"
	PY_LDFLAGS_NODIST = "-Wl,--strip-all -Wl,--strip-all"
versions/3.9.18
	CONFIGURE_LDFLAGS_NODIST = "-Wl,--strip-all"
	PY_CORE_LDFLAGS = "-L/usr/local/.pyenv/versions/3.9.18/lib -Wl,-rpath,/usr/local/.pyenv/versions/3.9.18/lib -L/usr/local/.pyenv/versions/3.9.18/lib -Wl,-rpath,/usr/local/.pyenv/versions/3.9.18/lib -Wl,--strip-all -Wl,--strip-all"
	PY_LDFLAGS_NODIST = "-Wl,--strip-all -Wl,--strip-all"
dependabot@7a99005e7999:/usr/local/.pyenv$
```
I noticed the duplication of the string `Wl,--strip-all` mirrors duplication elsewhere that exists in the current image as well

Nishnha (approved):
Thank you for updating the variables. I'll look into deploying this tomorrow so I can monitor for any new errors. I'll also need to cut a new release of Dependabot Core so we can actually benefit from these smaller builds on our backend.

jeffwidman:
Thanks @tvalenta for your hard work here!

I was planning to look at this after you tagged me, and then @Nishnha beat me to it. 😁 

However, yesterday I got super frustrated with how slow it was compiling from source, and went on a little code spelunking trip and resulted in:
* https://github.com/dependabot/dependabot-core/pull/7928
* https://github.com/dependabot/dependabot-core/pull/7934

I think that might be a better way to go since it completely sidesteps compilation entirely? It also looks like upstream already has some of these flags set, so we get the space benefits as well as the speed benefits.

tvalenta:
Not compiling is, I'd agree, the better way to go. I've seen a number of sources still have the bytecode pyc files and unstripped binaries, but I'd be of the opinion that those would be easy enough to deal with separately.  In which case, I'll close this and should there be an opportunity to clean up cruft that lingers in the prebuilt packages, offer to do that later.

From all my trials, each version should only be about 110MB total.

jeffwidman:
@tvalenta for the 110MB, is that total including after running the python native helper build script?

I found that added the majority of size as it downloads poetry, pipenv, etc and all their deps... Typical sizes I observed were more like 0.5GB.

Def interested in any optimizations you want to put up once this other PR lands, although probably most helpful if you first see if they can be contributed back up to the source docker image as then more folks beyond :dependabot: will benefit from them.

Issue #7889, opened 2023-08-23
Balance parallel tests by runtime


deivid-rodriguez:
### Code improvement description

We now use [parallel_tests](https://github.com/grosser/parallel_tests/tree/v3.8.0) in the "standard" way, since #6590.

However, we could improve CI speed more if we balanced parallel groups of specs by previous runtimes. See https://github.com/grosser/parallel_tests/tree/v3.8.0#even-test-group-run-times.

We could save previous runtimes on CI runs in the main branch as artifacts, and then download and use those in our regular CI runs for more efficient test group split.

If we do that, we can remove the `python_slow` job and the `:slow` tags in python specs, since we will automatically distribute slow tests as efficiently as possible.

PR #7925, opened 2023-08-29
Remove `slow` test designation from Python tests.

jeffwidman:
These were originally split out separate from the other Python tests because they are extra slow... typically because they shell out to the underlying package manager which may have to do a buncho of work such as downloading/compiling/installing packages.

However, with the recent switch to `parallel_tests` in https://github.com/dependabot/dependabot-core/pull/6590, this may no longer be necessary.

Ideally we actually record the slow tests dynamically and then use that for future runs
(https://github.com/dependabot/dependabot-core/issues/7889) but putting up this to test what happens to runtimes if we simply stop splitting out the slow tests... I imagine things will slow down, but may not be as much as we anticipate?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,6 @@ jobs:
           - { path: nuget, name: nuget, ecosystem: nuget }
           - { path: pub, name: pub, ecosystem: pub }
           - { path: python, name: python, ecosystem: pip }
-          - { path: python, name: python_slow, ecosystem: pip }
           - { path: swift, name: swift, ecosystem: swift }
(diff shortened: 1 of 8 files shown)


ailadas:
@dependabot rebase

jeffwidman:
Doing this resulted in the python test suite taking ~14 mins... ie, doing this is the same as adding normal python test run time + slow python test run time... which is a bummer as I haven't dug into the parallel tests enough to understand the internals, so was holding out hope that it internally would parallelize the run and see not much slowdown... 

Closing for now, this will be blocked by a fair bit more work to update to `pipenv upgrade` / `pip --dry-run` which should speedup the invocations of native package managers, which tend to be the slow ones.

PR #7928, opened 2023-08-30
Speed up Python docker builds using pre-compiled python

jeffwidman:
Our Python docker builds are slow... a typical CI test run for Python
takes ~21 minutes, of which the first ~10 minutes are spent building the
image.

This results in slower local development, slow CI test suites, and slow deployments.

The main culprit is `pyenv install` which under the covers downloads the
python source and then compiles it locally. Profiling showed that the
download was quick, so even though `pyenv` supports `aria2c`, there's
not much to be gained there. Unfortunately, a quick look at the `pyenv`
issue tracker showed [there's no way to to pass pre-compiled artifacts to
`pyenv`](https://github.com/orgs/pyenv/discussions/1872).

For a long time we've bandied about the idea of switching from `pyenv`
to downloading pre-compiled Pythons. In fact, the GitHub Actions
publishes versions that we could use here:
https://github.com/actions/python-versions/releases

However, we use `pyenv local` + `pyenv exec` throughout our Ruby code
for switching to different Python versions. So we thought that it'd take
a week or more to fully migrate away from `pyenv`.

Today I had to rebuild the python image multiple times, and got so
annoyed that I decided to poke at it a bit.

It turns out that `pyenv` is simply a shim layer, and as long as
`/usr/local/.pyenv/versions/<x.y.z>/bin` exists, it will happily pass
commands to anything in that folder.

So I was able to come up with an intermediate solution that speeds the
builds up drastically without requiring a large code refactor.

Running this locally results in the Python download/install/build
step going from ~500 seconds all the way down to ~75 seconds, a savings
of 7 minutes. Given that a full CI run of the python test suite
previously took ~21 minutes, this cuts it by 1/3.

Related but pulling the pre-compiled Python from a different source:
* https://github.com/dependabot/dependabot-core/pull/7934

### DIFF
--- a/python/Dockerfile
+++ b/python/Dockerfile
@@ -16,6 +16,8 @@ USER root
 
 # Install *only* the apt packages required for this builder image to build Python.
 # C-libs needed by users to build their Python packages should be installed down below in the final docker image.
+# TODO: not all these packages may be needed now that we've switched from `pyenv install` which compiled from source to
+# downloading / copying pre-compiled python
 RUN apt-get update \
   && apt-get upgrade -y \
   && apt-get install -y --no-install-recommends \
@@ -32,10 +34,15 @@ RUN apt-get update \
     tk-dev \
     xz-utils \
     zlib1g-dev \
+    # jq isn't necessary to build Python from source, it's used below for locating the download URLs
+    jq \
   && rm -rf /var/lib/apt/lists/*
 
 COPY --chown=dependabot:dependabot python/helpers /opt/python/helpers
 USER root
+# TODO: Now that switched from `pyenv install` which compiled from source to downloading / copying a pre-compiled python
+# we could entirely drop pyenv if we change our ruby code that calls `pyenv exec` to track which version of python to
+# call and uses the full python paths.
 ENV PYENV_ROOT=/usr/local/.pyenv \
   PATH="/usr/local/.pyenv/bin:$PATH"
 RUN mkdir -p "$PYENV_ROOT" && chown dependabot:dependabot "$PYENV_ROOT"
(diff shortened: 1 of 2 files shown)


jeffwidman:
Whoa, the benefits are even better on the server!
1. Before: https://github.com/dependabot/dependabot-core/actions/runs/6015676784/job/16318038486?pr=7905 👉  11m 39s
2. After: https://github.com/dependabot/dependabot-core/actions/runs/6023271199/job/16339680992?pr=7928 👉 2m 4s

python/Dockerfile
```diff
@@ -16,6 +16,8 @@ USER root
 
 # Install *only* the apt packages required for this builder image to build Python.
 # C-libs needed by users to build their Python packages should be installed down below in the final docker image.
+# TODO: not all these packages may be needed now that we've switched from `pyenv install` which compiled from source to
+# downloading / copying pre-compiled python
```
jeffwidman, line 20: There's likely further speed/space improvements possibly by dropping some of these packages, but I'd rather do that in a separate PR after letting this one prove itself in production.

python/Dockerfile
```diff
@@ -32,43 +34,89 @@ RUN apt-get update \
...
     zlib1g-dev \
+    # jq isn't necessary to build Python from source, it's used below for locating the download URLs
+    jq \
   && rm -rf /var/lib/apt/lists/*
 
 COPY --chown=dependabot:dependabot python/helpers /opt/python/helpers
 USER root
+# TODO: Now that switched from `pyenv install` which compiled from source to downloading / copying a pre-compiled python
+# we could entirely drop pyenv if we change our ruby code that calls `pyenv exec` to track which version of python to
+# call and uses the full python paths.
```
jeffwidman, line 45: A quick grep of `pyenv ` commands showed this actually should be pretty straightforward, since we already have to keep track of when to tell `pyenv` to switch to a different python version.

Dropping `pyenv` completely would simplify the Dockerfile and be one less dep we pull in. But again, rather handle that in a separate PR after letting this one marinate a bit.

python/Dockerfile
```diff
@@ -32,43 +34,89 @@ RUN apt-get update \
...
+  | xargs -i -- curl -L {} \
+  | tar xz \
+  && rm build_output.txt Python-$PY_3_8.tgz tools_structure.txt setup.sh \
+  && mkdir /usr/local/.pyenv/versions/ \
+  && ln -s /opt/hostedtoolcache/Python/$PY_3_8/x64 /usr/local/.pyenv/versions/$PY_3_8 \
   && bash /opt/python/helpers/build $PY_3_8 \
-  && cd /usr/local/.pyenv \
-  && tar czf 3.8.tar.gz versions/$PY_3_8
+  && cd /opt/hostedtoolcache/Python \
+  && tar czf $PY_3_8.tar.gz $PY_3_8
```
jeffwidman, line 69: These are tar'd for space reasons... Since this changes the compile source, I double-checked the size, and each one still averaged ~0.5GB uncompressed. Since majority of our runs use the default `python` (currently `3.11`), it's still worth compressing the non-default versions.

python/Dockerfile
```diff
@@ -32,43 +34,89 @@ RUN apt-get update \
...
+  | xargs -i -- curl -L {} \
+  | tar xz \
+  && rm build_output.txt Python-$PY_3_10.tgz tools_structure.txt setup.sh \
+  && mkdir /usr/local/.pyenv/versions/ \
+  && ln -s /opt/hostedtoolcache/Python/$PY_3_10/x64 /usr/local/.pyenv/versions/$PY_3_10 \
   && bash /opt/python/helpers/build $PY_3_10 \
-  && cd /usr/local/.pyenv \
-  && tar czf 3.10.tar.gz versions/$PY_3_10
+  && cd /opt/hostedtoolcache/Python \
+  && tar czf $PY_3_10.tar.gz $PY_3_10
```
jeffwidman, line 97: I hate the lack of DRY across all these, but AFAIK there's no clean way to loop something like this in a multi-stage docker build.

I didn't duplicate the comments across each one... decided that doing it solely on the default `3.11` install sufficed.

python/lib/dependabot/python/language_version_manager.rb
```diff
@@ -23,7 +23,12 @@ def install_required_python
         return if SharedHelpers.run_shell_command("pyenv versions").include?(" #{python_major_minor}.")
 
         SharedHelpers.run_shell_command(
-          "tar xzf /usr/local/.pyenv/#{python_major_minor}.tar.gz -C /usr/local/.pyenv/"
+          "tar xzf /opt/hostedtoolcache/Python/#{python_version}.tar.gz -C /opt/hostedtoolcache/Python/"
+        )
+        # pyenv expects the python installation files in the `versions` folder, but the pre-compiled python3 / pip3
+        # expect to reside in the /opt/hostedtoolcache/Python/x.y.z/x64 dir, so need a symlink to make them play nice.
+        SharedHelpers.run_shell_command(
+          "ln -s /opt/hostedtoolcache/Python/#{python_version}/x64 /usr/local/.pyenv/versions/#{python_version}"
```
jeffwidman, line 31: This is the one place where it may be more painful to swap architectures, since I'm not sure how we communicate to the Ruby code that it needs `x64` vs `arm64`?

python/Dockerfile
```diff
@@ -32,43 +34,89 @@ RUN apt-get update \
...
 
 FROM python-core
-RUN pyenv install $PY_3_11 \
+# The pre-compiled Python expects to be installed to this dir
+RUN mkdir -p /opt/hostedtoolcache/Python/$PY_3_11/x64/ \
+  && cd /opt/hostedtoolcache/Python/$PY_3_11/x64/ \
+  # TODO: Add support for arm64 on Ubuntu whenever actions/python-versions adds support for it. Currently not available.
```
jeffwidman, line 103: I expect support for this to be added at some point, eg this upstream PR:
* https://github.com/actions/python-versions/pull/230

At that point, it'll be easy enough to make the `x64` vs `arm64` value dynamic here using `TARGETARCH`...

jeffwidman (commented):
Annotating the changes.

python/Dockerfile
jakecoffman: I can't build this locally on the M1 chip. Did you consider copying out the built Python from the official Docker images like how we're installing Ruby? That way we get the architectures for free without extra scripting, should simplify this part.

jeffwidman: I actually _did_ consider it, but for some reason thought that we were trying to move away from copying from official docker images... but a little digging and I see it's the opposite and the Ruby flip was to move away from `apt-get` towards docker images.

It's a great idea, let me try it.

jeffwidman: Turned out to be pretty straightforward:
* https://github.com/dependabot/dependabot-core/pull/7934

And it's also a bit cleaner as it allowed copying the files directly into `/usr/local/.pyenv/versions`, no need for any confusing symlinks.

jeffwidman:
Closing in favor of https://github.com/dependabot/dependabot-core/pull/7934 which is a very similar approach, but pulls the sources from the official docker image... benefits of that approach:
* Builds are faster by about 15 seconds for some reason
* Cleaner because the linking to the C-libs uses more relative links somehow, so I can fully copy it into `/usr/local/.pyenv/versions` so avoids confusing symlinks
* Supports `arm` out of the box

PR #7914, opened 2023-08-28
Target latest Python versions - 3.11.5, 3.10.13, 3.9.18, 3.8.18

phillipuniverse:

PR #7934, opened 2023-08-31
Use pre-compiled Python from official Docker image

jeffwidman:
Our Python docker builds are slow... a typical CI test run for Python takes ~21 minutes, of which the first ~10 minutes are spent building the image.

This results in slower local development, slow CI test suites, and slow deployments.

The main culprit is `pyenv install` which under the covers downloads the python source and then compiles it locally. Profiling showed that the download was quick, so even though `pyenv` supports `aria2c`, there's not much to be gained there. Unfortunately, a quick look at the `pyenv` issue tracker showed [there's no way to to pass pre-compiled artifacts to `pyenv`](https://github.com/orgs/pyenv/discussions/1872).

For a long time we've bandied about the idea of switching from `pyenv` to downloading pre-compiled Pythons.However, we use `pyenv local` + `pyenv exec` throughout our Ruby code for switching to different Python versions. So we thought that it'd take a week or more to fully migrate away from `pyenv`.

Today I had to rebuild the python image multiple times, and got so annoyed that I decided to poke at it a bit.

It turns out that `pyenv` is simply a shim layer, and as long as `/usr/local/.pyenv/versions/<x.y.z>/bin` exists, it will happily pass commands to anything in that folder.

So I was able to come up with an intermediate solution that speeds the builds up drastically without requiring a large code refactor.

Running this locally results in the Python download/install/build step going from ~500 seconds all the way down to ~33 seconds, a savings of nearly 8 minutes. Given that a full CI run of the python test suite previously took ~21 minutes, this cuts it by 1/3.

Related but pulling the pre-compiled Python from a different source:
* https://github.com/dependabot/dependabot-core/pull/7928

### DIFF
--- a/python/Dockerfile
+++ b/python/Dockerfile
@@ -16,6 +16,8 @@ USER root
 
 # Install *only* the apt packages required for this builder image to build Python.
 # C-libs needed by users to build their Python packages should be installed down below in the final docker image.
+# TODO: not all these packages may be needed now that we've switched from `pyenv install` which compiled from source to
+# downloading / copying pre-compiled python
 RUN apt-get update \
   && apt-get upgrade -y \
   && apt-get install -y --no-install-recommends \
@@ -36,6 +38,9 @@ RUN apt-get update \
 
 COPY --chown=dependabot:dependabot python/helpers /opt/python/helpers
 USER root
+# TODO: Now that switched from `pyenv install` which compiled from source to downloading / copying a pre-compiled python
+# we could entirely drop pyenv if we change our ruby code that calls `pyenv exec` to track which version of python to
+# call and uses the full python paths.
 ENV PYENV_ROOT=/usr/local/.pyenv \
   PATH="/usr/local/.pyenv/bin:$PATH"
 RUN mkdir -p "$PYENV_ROOT" && chown dependabot:dependabot "$PYENV_ROOT"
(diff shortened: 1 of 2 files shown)


jeffwidman:
This seems like a better way forward than #7928 for the reasons explained here:
* https://github.com/dependabot/dependabot-core/pull/7928#issuecomment-1700036744

In particular, I was surprised by the 15s speedup in build times compared to that.

python/Dockerfile
```diff
@@ -36,39 +38,96 @@ RUN apt-get update \
...
+# We used to use `pyenv install 3.x.y` but it's really slow because it compiles from source (~500s). So instead, we hack
+# around that by downloading pre-compiled versions, then placing them where pyenv expects it in the `versions` subfolder.
+# In the future, we should consider dropping pyenv completely, as it's mostly used here for legacy reasons...
+# Although it is convenient when debugging to be able to quickly flip through environments.
+RUN mkdir "${PYENV_ROOT}/versions"
+
+## 3.8
+# Docker doesn't support parametrizing `COPY --from:python:$PY_1_23-bookworm`, so work around it using an alias.
+# TODO: If upstream adds support for Ubuntu, use that instead of Debian as the base suffix: https://github.com/docker-library/python/pull/791
+FROM python:$PY_3_8-bookworm as upstream-python-3.8
```
jeffwidman, line 60: @phillipuniverse will be happy to see that this is fully parametrized so it doesn't have to be updated when we bump to latest python patch releases. 😉 
* https://github.com/dependabot/dependabot-core/pull/7914#issue-1870109012

python/Dockerfile
```diff
@@ -36,39 +38,96 @@ RUN apt-get update \
...
+RUN bash /opt/python/helpers/build $PY_3_11
+
+RUN pyenv global $PY_3_11
 
-COPY --from=python-3.10 /usr/local/.pyenv/3.10.tar.gz /usr/local/.pyenv/3.10.tar.gz
-COPY --from=python-3.9 /usr/local/.pyenv/3.9.tar.gz /usr/local/.pyenv/3.9.tar.gz
-COPY --from=python-3.8 /usr/local/.pyenv/3.8.tar.gz /usr/local/.pyenv/3.8.tar.gz
+COPY --from=python-3.8 $PYENV_ROOT/versions/$PY_3_8.tar.gz $PYENV_ROOT/versions/$PY_3_8.tar.gz
+COPY --from=python-3.9 $PYENV_ROOT/versions/$PY_3_9.tar.gz $PYENV_ROOT/versions/$PY_3_9.tar.gz
+COPY --from=python-3.10 $PYENV_ROOT/versions/$PY_3_10.tar.gz $PYENV_ROOT/versions/$PY_3_10.tar.gz
```
jeffwidman, line 130: I realized these should probably get moved below the below `apt-get install` so that if for some reason one of them is slow to build, the `3.11` image can proceed a little further before it gets blocked waiting on it... 

Will do in a follow-on PR...

python/Dockerfile
```diff
@@ -36,39 +38,96 @@ RUN apt-get update \
 
 COPY --chown=dependabot:dependabot python/helpers /opt/python/helpers
 USER root
+# TODO: Now that switched from `pyenv install` which compiled from source to downloading / copying a pre-compiled python
+# we could entirely drop pyenv if we change our ruby code that calls `pyenv exec` to track which version of python to
+# call and uses the full python paths.
 ENV PYENV_ROOT=/usr/local/.pyenv \
   PATH="/usr/local/.pyenv/bin:$PATH"
```
Nishnha, line 45: I think you can remove this line that updates the PATH env

python/Dockerfile
phillipuniverse: Nice!! I am indeed happy to see this!

Also, good notes overall on optimizing the Docker build with pyenv. I have struggled with long build times myself in other projects for the same reasons!

python/Dockerfile
```diff
@@ -36,39 +38,96 @@ RUN apt-get update \
...
+COPY --from=upstream-python-3.8 --chown=dependabot:dependabot /usr/local/include $PYTHON_INSTALL_LOCATION/include
+COPY --from=upstream-python-3.8 --chown=dependabot:dependabot /usr/local/lib $PYTHON_INSTALL_LOCATION/lib
+# `pip` and other scripts need their shebangs rewritten for the new location
+RUN find $PYTHON_INSTALL_LOCATION/bin -type f -exec sed -i "1s|^#!/usr/local/bin/python|#!$PYTHON_INSTALL_LOCATION/bin/python|" {} +
+# Ensure pyenv works and it's the python version we expect
+RUN PYENV_VERSION=$PY_3_8 pyenv exec python --version | grep "Python $PY_3_8" || exit 1
+RUN bash /opt/python/helpers/build $PY_3_8
+# This python environment occupies ~0.5 GB and gets used for a fraction of jobs, so store it compressed.
+RUN cd $PYENV_ROOT/versions \
+  && tar czf $PY_3_8.tar.gz $PY_3_8
```
Nishnha, line 73: When would these get uncompressed?

python/lib/dependabot/python/language_version_manager.rb
```diff
@@ -23,7 +23,7 @@ def install_required_python
         return if SharedHelpers.run_shell_command("pyenv versions").include?(" #{python_major_minor}.")
 
         SharedHelpers.run_shell_command(
-          "tar xzf /usr/local/.pyenv/#{python_major_minor}.tar.gz -C /usr/local/.pyenv/"
+          "tar xzf /usr/local/.pyenv/versions/#{python_version}.tar.gz -C /usr/local/.pyenv/versions"
```
Nishnha, line 26: I think this has been brought up in conversation elsewhere, but this change would mean some projects using an older patch version would be run using the latest matching minor version

That should be okay though as long as the patch updates are actually non-breaking...

jeffwidman: Not sure I follow... I think this only changes an implementation detail of the name of the tar file? The result of the untarring doesn't actually change.

Also, the matching logic for older patch getting coerced to newer patch doesn't happen here, it happens over in:
https://github.com/dependabot/dependabot-core/blob/78bd848e8881fc0bec480ed1ca809b0c40efab5d/python/lib/dependabot/python/language_version_manager.rb#L54-L61

And we already  have been doing that for months.

python/Dockerfile
jeffwidman: Don't we need this so the `pyenv` shims take precedence on the `PATH`?

https://github.com/pyenv/pyenv#set-up-your-shell-environment-for-pyenv

python/lib/dependabot/python/language_version_manager.rb
Nishnha: Understood. I guess I had mixed the two up

python/Dockerfile
```diff
@@ -36,39 +38,96 @@ RUN apt-get update \
...
-  && cd /usr/local/.pyenv \
-  && tar czf 3.8.tar.gz versions/$PY_3_8
+ARG PYTHON_INSTALL_LOCATION="$PYENV_ROOT/versions/$PY_3_8"
+COPY --from=upstream-python-3.8 --chown=dependabot:dependabot /usr/local/bin $PYTHON_INSTALL_LOCATION/bin
+COPY --from=upstream-python-3.8 --chown=dependabot:dependabot /usr/local/include $PYTHON_INSTALL_LOCATION/include
+COPY --from=upstream-python-3.8 --chown=dependabot:dependabot /usr/local/lib $PYTHON_INSTALL_LOCATION/lib
+# `pip` and other scripts need their shebangs rewritten for the new location
+RUN find $PYTHON_INSTALL_LOCATION/bin -type f -exec sed -i "1s|^#!/usr/local/bin/python|#!$PYTHON_INSTALL_LOCATION/bin/python|" {} +
+# Ensure pyenv works and it's the python version we expect
+RUN PYENV_VERSION=$PY_3_8 pyenv exec python --version | grep "Python $PY_3_8" || exit 1
```
Nishnha, line 69: What happens if this command fails and returns `exit 1`? Does the whole Docker build fail?

python/lib/dependabot/python/language_version_manager.rb
jeffwidman: It's super confusing, my memory just happens to be fresh since I rewrote that code a few weeks ago as we could simplify the logic a bit after we dropped python 3.6/3.7...

python/Dockerfile
jeffwidman: Yes!

IMO we should be doing this test on `main` regardless of whether we go with this approach or not, as a safeguard against copy/pasting a change from one version to another and forgetting to update the version number... although i suppose in that case having it parametrized kinda defeats the purpose :-D But I figured this guard against at least some breakage, or if the upstream dockerfile or pyenv changed behavior in some way...

python/Dockerfile
Nishnha: Yep you're right we do. I thought this was adding the actual python installations to the PATH

Nishnha (approved):
I really like this approach of using the official docker images. The faster CI build step is a huge win 🏅

Also, merging this PR would mean https://github.com/dependabot/dependabot-core/pull/7891 is no longer necessary.

Your comments helped a lot in understanding the changes, thanks! You may want to open this as a separate issue though

> TODO: Now that switched from `pyenv install` which compiled from source to downloading / copying a pre-compiled python we could entirely drop pyenv if we change our ruby code that calls `pyenv exec` to track which version of python to call and uses the full python paths.

python/Dockerfile
Nishnha: > as a safeguard against copy/pasting a change from one version to another and forgetting to update the version number... although I suppose in that case having it parametrized kinda defeats the purpose 😁.

Agreed, this is a good idea. It would also reveal if pyenv changes where/how it recognizes valid python installs

> But I figured this guards against at least some breakage, or if the upstream python docker image or pyenv changed behavior / file structure in some way...

Yeah I think this method of installing python packages could be seen as temperamental if upstream changes their behavior (especially with swapping out the shebang path), but we can always revert if we need to.

jeffwidman: Yeah, I wasn't super excited about needing to mung the shebang path, and for compiling ourselves we could even tune the compiler for our use cases, but since our use case is mostly CI / ephemeral jobs that run just a few lines of Python, rather than a long-lived service, it makes a _lot_ more sense to trade faster build times for slower runtimes since it's likely a few `ms` runtime improvement at most...

python/Dockerfile
jeffwidman: https://github.com/dependabot/dependabot-core/pull/7941

PR #7935, opened 2023-08-31
Collapse run command experiment

jeffwidman:
Experiment on https://github.com/dependabot/dependabot-core/pull/7934 to see if collapsing the `RUN` instructions to a single `RUN` command is more performant...


jeffwidman:
Threw this up as a quick experiment because I was curious, but doesn't look like any change: https://github.com/dependabot/dependabot-core/pull/7935/commits/a1c147472659dc932299381d8dd132d00a9a8221

One build was 1:40, and the other 2:03, vs 1:45 previously for two builds w/o this optimization, so both are within spitting distance of one another from a time perf perspective...

Also, I realized that one other side-effect of splitting by layers is how it impacts cache hits when doing "cache from" in other images... for our use case there's zero impact since we never cache from these images.

PR #7941, opened 2023-08-31
Move copying the other pythons to the end of the Dockerfile

jeffwidman:
This is a multi-stage build, which means that building/tarring these files happens in concurrent builds.

Previously, if one of those builds took longer for some unexpected reason--perhaps a slow HTTP call, then it'd block the pipeline.

By moving them to the bottom, it allows the main image to proceed with `apt-get update && apt-get install ...` which takes a little bit of time and gives any slow concurrent builds plenty of time to finish.


jakecoffman (approved):
Nice!

PR #7943, opened 2023-08-31
Stop installing apt packages for compiling Python

jeffwidman:
Now that we've stopped compiling Python from source and instead download/copy a pre-compiled version, there's no need to install these apt packages.

A few may be needed for users to build python deps written in native C, but those shouldn't be listed in this section, they should be added later in the the Dockerfile where we already list other packages needed for that:
https://github.com/dependabot/dependabot-core/blob/51b268a81a983c6547af4d95768218a03bd8e751/python/Dockerfile#L128-L143

Also, it's unclear which of these would be needed, so let's start by removing these, and then as needed can add back the couple that we actually discover are needed.

### DIFF
--- a/python/Dockerfile
+++ b/python/Dockerfile
@@ -14,30 +14,8 @@ ARG PY_3_8
 ARG PYENV_VERSION
 USER root
 
-# Install *only* the apt packages required for this builder image to build Python.
-# C-libs needed by users to build their Python packages should be installed down below in the final docker image.
-# TODO: not all these packages may be needed now that we've switched from `pyenv install` which compiled from source to
-# downloading / copying pre-compiled python
-RUN apt-get update \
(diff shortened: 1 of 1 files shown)


Nishnha (approved):
LGTM I compared this to the suggested build environment list for pyenv https://github.com/pyenv/pyenv/wiki#suggested-build-environment

The only one not listed here was curl but I guess that's installed elsewhere

jeffwidman:
I'm going to try deploying this in prod and see what Sentry shows for user errors... but again, probably we should move forward with this, and then add back missing packages as needed in a follow-on PR to https://github.com/dependabot/dependabot-core/blob/51b268a81a983c6547af4d95768218a03bd8e751/python/Dockerfile#L128-L143

PR #7946, opened 2023-08-31
Python tar consistency

tvalenta:
When creating tar archives of Python versions:
* sort directory entries by name instead of OS order
* Use the timestamp of libpython3.so as the mtime of all added files.

This should increase the chance that archives created from consistent
sources will be uniform and reproducible, being more likely to match
cached layers.

On one Linux instance:
```
dependabot@492fc87f7481:/usr/local/.pyenv/versions$ md5sum *gz
e4a8931b35ccbd103fb194aa477f0f10  3.10.13.tar.gz
6f9123a5ef7b5f6a8c83d566a840cc6c  3.8.18.tar.gz
2e9d8dc9c22ff8809afd3f26265b0d08  3.9.18.tar.gz
```
On a different server:
```
dependabot@492fc87f7481:/usr/local/.pyenv/versions$ md5sum *gz
e4a8931b35ccbd103fb194aa477f0f10  3.10.13.tar.gz
6f9123a5ef7b5f6a8c83d566a840cc6c  3.8.18.tar.gz
2e9d8dc9c22ff8809afd3f26265b0d08  3.9.18.tar.gz
```

### DIFF
--- a/python/Dockerfile
+++ b/python/Dockerfile
@@ -70,7 +70,7 @@ RUN PYENV_VERSION=$PY_3_8 pyenv exec python --version | grep "Python $PY_3_8" ||
 RUN bash /opt/python/helpers/build $PY_3_8
 # This python environment occupies ~0.5 GB and gets used for a fraction of jobs, so store it compressed.
 RUN cd $PYENV_ROOT/versions \
-  && tar czf $PY_3_8.tar.gz $PY_3_8
+  && tar --mtime=./$PY_3_8/lib/libpython3.so --sort=name -czf $PY_3_8.tar.gz $PY_3_8
 
 ## 3.9
 # Docker doesn't support parametrizing `COPY --from:python:$PY_1_23-bookworm`, so work around it using an alias.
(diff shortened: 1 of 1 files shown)


python/Dockerfile
```diff
@@ -70,7 +70,7 @@ RUN PYENV_VERSION=$PY_3_8 pyenv exec python --version | grep "Python $PY_3_8" ||
 RUN bash /opt/python/helpers/build $PY_3_8
 # This python environment occupies ~0.5 GB and gets used for a fraction of jobs, so store it compressed.
 RUN cd $PYENV_ROOT/versions \
-  && tar czf $PY_3_8.tar.gz $PY_3_8
+  && tar --mtime=./$PY_3_8/lib/libpython3.so --sort=name -czf $PY_3_8.tar.gz $PY_3_8
```
jeffwidman, line 73:
```suggestion
  # using a consistent mtime and sort order increases the odds this layer will be pulled from the docker image cache
  && tar --mtime=./$PY_3_8/lib/libpython3.so --sort=name -czf $PY_3_8.tar.gz $PY_3_8
```

Can you add a code comment like ☝️ to all of these for consistency? I know there's `git blame`, but this one is non-obvious enough that an inline comment would be helpful for anyone who comes along later wondering why this was done and if changing it will break anything.

tvalenta: I can, and I will!

jeffwidman:
👍 makes sense to me, with a minor nit about commenting it so that later devs know that if they need to change this it won't break anything other than cache hit rate.

tvalenta:
Since there have been modifications to the Dockerfile, I'll just close this and redo for future consideration. The changes are minor enough that it would be an easier path.

PR #7949, opened 2023-09-01
Make building the default python concurrent rather than sequential

jeffwidman:
Previously we'd build all the non-default pythons concurrently with building the default python, and then sequence installing the `apt` packages behind building the default Python.

However, these `apt` packages can take a little while to fetch/install, and they don't require the default Python, so instead parallelize building the default python at the same time as installing the apt packages.

This also makes the flow consistent across all pythons of building in a separate stage and then copying to the final image.

### DIFF
--- a/python/Dockerfile
+++ b/python/Dockerfile
@@ -90,7 +90,7 @@ RUN cd $PYENV_ROOT/versions \
 # Docker doesn't support parametrizing `COPY --from:python:$PY_1_23-bookworm`, so work around it using an alias.
 # TODO: If upstream adds support for Ubuntu, use that instead of Debian as the base suffix: https://github.com/docker-library/python/pull/791
 FROM python:$PY_3_11-bookworm as upstream-python-3.11
-FROM python-core
+FROM python-core as python-3.11
 ARG PYTHON_INSTALL_LOCATION="$PYENV_ROOT/versions/$PY_3_11"
 COPY --from=upstream-python-3.11 --chown=dependabot:dependabot /usr/local/bin $PYTHON_INSTALL_LOCATION/bin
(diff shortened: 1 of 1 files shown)


jeffwidman:
Thought of this while I was driving down the road running an errand and couldn't resist throwing it up... 😁

jeffwidman:
Hmm... will debug more later:
```shell
[dependabot-core-dev] ~/python $ DEBUG_HELPERS=true rspec ./spec/dependabot/python/file_updater_spec.rb:125
Run options: include {:locations=>{"./spec/dependabot/python/file_updater_spec.rb"=>[125]}}

Randomized with seed 8995
pyenv exec python3 /opt/python/run.py
get_pipfile_hash

/usr/local/.pyenv/versions/3.11.5/bin/python3: can't open file '/opt/python/run.py': [Errno 2] No such file or directory
```

python/Dockerfile
```diff
@@ -100,9 +100,9 @@ RUN find $PYTHON_INSTALL_LOCATION/bin -type f -exec sed -i "1s|^#!/usr/local/bin
...
 RUN PYENV_VERSION=$PY_3_11 pyenv exec python --version | grep "Python $PY_3_11" || exit 1
 RUN bash /opt/python/helpers/build $PY_3_11
+# This is the default Python, so no need to tar it
 
-RUN pyenv global $PY_3_11
-
+FROM python-core
 # Install C-libs needed to build users' Python packages. Please document why each package is needed.
 USER root
 RUN apt-get update \
```
jeffwidman, line 108: This is the list of package installs that we now do concurrently with the python builds... It shouldn't take _that_ long, but nevertheless it's faster to do this concurrently, plus I expect this list to grow over time as users report C-libs that they need installed to build their deps.

Issue #3759, opened 2021-05-20
Dependabot errors on gradle dependency with prefix version range


J3173:
**Package ecosystem**
gradle

**Package manager version**
gradle 6.5.1

**Language version**
Java 11

**Manifest location and content prior to update**
https://github.com/openEHR/archie/blob/51a8c40f6741a4ca33befae80489537054a24ae5/build.gradle#L75

**dependabot.yml content**
https://github.com/openEHR/archie/blob/51a8c40f6741a4ca33befae80489537054a24ae5/.github/dependabot.yml

**Updated dependency**
junit:junit:4.+

**What you expected to see, versus what you actually saw**
Dependabot status shows "Dependabot encountered an unknown error" when you have a gradle dependency with a [prefix version range](https://docs.gradle.org/current/userguide/single_versions.html), i.e. a version string ending in a `+` sign, e.g.: `testCompile 'junit:junit:4.+'`. Changing the dependency to `testCompile 'junit:junit:4.13.2'` makes the Dependabot run complete normally.

**Native package manager behavior**
N/A

**Images of the diff or a link to the PR, issue or logs**
Log snippet from Insights > Dependency graph > Dependabot:
```
updater | I, [2021-05-20T11:12:06.586733 #7]  INFO -- sentry: ** [Raven] Sending event 3470f01421054b2c925703a1ed699d1a to Sentry
updater | ERROR <job_136399394> Error processing junit:junit (RuntimeError)
updater | ERROR <job_136399394> No files changed!
updater | ERROR <job_136399394> /home/dependabot/dependabot-updater/vendor/ruby/2.6.0/gems/dependabot-gradle-0.148.3/lib/dependabot/gradle/file_updater.rb:35:in `updated_dependency_files'
updater | ERROR <job_136399394> /home/dependabot/dependabot-updater/lib/dependabot/updater.rb:691:in `generate_dependency_files_for'
updater | ERROR <job_136399394> /home/dependabot/dependabot-updater/lib/dependabot/updater.rb:281:in `check_and_create_pull_request'
updater | ERROR <job_136399394> /home/dependabot/dependabot-updater/lib/dependabot/updater.rb:89:in `check_and_create_pr_with_error_handling'
updater | ERROR <job_136399394> /home/dependabot/dependabot-updater/lib/dependabot/updater.rb:56:in `block in run'
updater | ERROR <job_136399394> /home/dependabot/dependabot-updater/lib/dependabot/updater.rb:56:in `each'
[...]

PR #7975, opened 2023-09-05
Gradle: fix comparison of the prefix version range

jakecoffman:
Fixes https://github.com/dependabot/dependabot-core/issues/3759

This change causes the `+` to always compare higher than any version in the semver position. That means when the dependency specifies `4.+` with no lockfile, it won't attempt to update to version 4.11 because 4.+ already covers that.

The code in `gradle/version.rb` is particularly challenging. If there's a better way to write this let me know.

### DIFF
--- a/gradle/lib/dependabot/gradle/version.rb
+++ b/gradle/lib/dependabot/gradle/version.rb
@@ -153,6 +153,10 @@ module Dependabot
       end
 
       def compare_prefixed_token(prefix:, token:, other_prefix:, other_token:)
+        return 1 if token == "+" && other_token != "+"
+        return -1 if other_token == "+" && token != "+"
+        return 0 if token == "+" && other_token == "+"
+
         token_type = token.match?(/^\d+$/) ? :number : :qualifier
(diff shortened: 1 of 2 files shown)


gradle/lib/dependabot/gradle/version.rb
```diff
@@ -153,6 +153,10 @@ def pad_for_comparison(prefixed_tokens, other_prefixed_tokens)
       end
 
       def compare_prefixed_token(prefix:, token:, other_prefix:, other_token:)
+        return 1 if token == "+" && other_token != "+"
+        return -1 if other_token == "+" && token != "+"
+        return 0 if token == "+" && other_token == "+"
+
```
jurre, line 159: This is probably a bit gnarly, and not recommended, but I think `other_token <=> token` will produce the desired results, it relies on `"1" > "+"` being true

jurre (approved):

gradle/lib/dependabot/gradle/version.rb
jakecoffman: Yeah that might be misleading, although a part of me likes it. I'll stick with the long-winded version for now.

PR #7980, opened 2023-09-06
Maven: fix classifier being part of the dependency name

jakecoffman:
In https://github.com/dependabot/dependabot-core/pull/1924 the classifier was added to take account of the different URL pattern needed to pull the dependency, however it was added to the dependency name and this has caused some issues around `allow` and `ignore`. 

I've put up a repro here: https://github.com/dsp-testing/maven-no-make-pr

Since it is appending the classifier to the end of the dependency name, the allow is preventing the update.

To fix this I moved the classifier into the dependency metadata, which didn't exist when #1924 was created, but now is a perfect place for this data to live!

### DIFF
--- a/maven/lib/dependabot/maven/file_parser.rb
+++ b/maven/lib/dependabot/maven/file_parser.rb
@@ -89,9 +89,6 @@ module Dependabot
         return unless (name = dependency_name(dependency_node, pom))
         return if internal_dependency_names.include?(name)
 
-        classifier = dependency_classifier(dependency_node, pom)
-        name = "#{name}:#{classifier}" if classifier
-
         build_dependency(pom, dependency_node, name)
       end
 
@@ -109,7 +106,7 @@ module Dependabot
             property_source: property_source(dependency_node, pom)
           }.compact
 
-        Dependency.new(
+        dependency = Dependency.new(
           name: name,
           version: dependency_version(pom, dependency_node),
           package_manager: "maven",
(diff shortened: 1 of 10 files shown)


maven/lib/dependabot/maven/file_parser.rb
```diff
@@ -123,6 +120,10 @@ def build_dependency(pom, dependency_node, name)
             }.merge(property_details)
           }]
         )
+
+        classifier = dependency_classifier(dependency_node, pom)
+        dependency.requirements.first[:metadata][:classifier] = classifier if classifier
+        dependency
```
landongrindheim, line 126: 🔎 What do you think of doing this inline similarly to `groups: dependency_groups(pom, dependency_node)`? That'd look something like:

```ruby
        Dependency.new(
          name: name,
          version: dependency_version(pom, dependency_node),
          package_manager: "maven",
          requirements: [{
            requirement: dependency_requirement(pom, dependency_node),
            file: pom.name,
            groups: dependency_groups(pom, dependency_node),
            source: nil,
            metadata: {
              packaging_type: packaging_type(pom, dependency_node)
              classifier: dependency_classifier(dependency_node, pom)
            }.merge(property_details).compact
          }]
        )
```

(with an added `Hash#compact` to take care of the `nil` case)

My thinking is that we can rely on the implicit return value and conform the surrounding style (for `version`, `requirements` and `groups`)

landongrindheim (approved):

maven/lib/dependabot/maven/file_parser.rb
jakecoffman: Ah, I had started doing this but the `nil` was failing tests. Good idea adding `compact`!

jakecoffman:
This fixes the issue in the repro, merging.

Issue #7915, opened 2023-08-28
Same dependencies showing up in multiple groups (bundler)


Drowze:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

bundler

### Manifest location and content before the Dependabot update

It's a private project with a lot of private gems (private gems being hosted on a self-hosted Sonatype Nexus Repository), but roughly:

Gemfile
```ruby
source "https://rubygems.org"

git_source(:github) do |repo_name|
  repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?("/")
  "https://github.com/#{repo_name}.git"
end

ruby File.read(File.join(__dir__, ".ruby-version")).strip

source "https://nexus.example.com/repository/my-gems/" do
  gem "gem_a-client"
  gem "gem_b-client"
  gem "rack-custom_middleware", require: "rack/custom_middleware"
end

gem "puma", "~> 5.6.6"
gem "puma-plugin-statsd"
gem "ddtrace", require: ENV["DATADOG_ENABLED"] == "true"
gem "unleash"
gem "karafka", "~> 2.1.4"
gem "karafka-rdkafka", "~> 0.13.4"
gem "waterdrop", "~> 2.6.6"

group :test 
  gem "karafka-testing"
end
```

### dependabot.yml content

# For updates options, check https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#configuration-options-for-updates
# Validate the file at https://dependabot.com/docs/config-file/validator/

version: 2
registries:
  my-nexus:
    type: rubygems-server
    url: https://nexus.example.com/repository/my-gems/
    username: nexus_read_only
    password: ${{secrets.NEXUS_READ_ONLY_PASSWORD}}

updates:
  - package-ecosystem: "bundler"
    directory: "/"
    insecure-external-code-execution: allow
    registries:
      - my-nexus
    schedule:
      interval: "weekly"
    commit-message:
      prefix: "[Dependabot]"
    pull-request-branch-name:
      separator: "-"
    groups:
      karafka:
        patterns:
          - "karafka"
          - "karafka-core"
          - "karafka-rdkafka"
          - "karafka-testing"
          - "waterdrop"
      internal-libraries:
        patterns:
          - "gem_a-client"
          - "gem_b-client"
          - "rack-custom_middleware", require: "rack/custom_middleware"
      dev-dependencies:
[...]

PR #7981, opened 2023-09-07
Dependencies only match with one group

Nishnha:
Fixes https://github.com/dependabot/dependabot-core/issues/7915

Right now, dependencies will show up in every group they match with. This PR changes that so dependencies will only match one group.

Dependencies will try to match against each group in the same order, which is the order that they are listed in the `dependabot.yml` file.

### DIFF
--- a/updater/lib/dependabot/dependency_group_engine.rb
+++ b/updater/lib/dependabot/dependency_group_engine.rb
@@ -9,12 +9,6 @@ require "dependabot/dependency_group"
 # any groups from the job's configuration before assigning the dependency list to
 # the groups.
 #
-# We permit dependencies to be in more than one group and also track those which
-# have zero matches so they may be updated individuall.
-#
-# **Note:** This is currently an experimental feature which is not supported
-#           in the service or as an integration point.
-#
 module Dependabot
   class DependencyGroupEngine
     class ConfigurationError < StandardError; end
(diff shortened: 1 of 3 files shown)


jurre (approved):

jakecoffman:
Another complexity of this is we skip the update [here](https://github.com/dependabot/dependabot-core/blob/1e1702652859e893b1e7fd1302518a9eb59fb3c6/updater/lib/dependabot/updater/operations/group_update_all_versions.rb#L79-L87) if a group PR exists for that group and add all the dependencies we think are in it. But I noticed now this code is actually wrong. We need to add all of the dependencies that are in the group in the job definition, not call `add_all_to_handled` which will add dependencies that MIGHT be in the group. 

Confusing, I know!

thobiaskarlsson:
Is this issue related to what we are experiencing. We just enabled `groups` (with `*`) this morning to a (internal) repository. Around 50 minutes ago Dependabot created a (very nice) PR with 10 updates. However, 45 minutes ago (~5 minutes later) Dependabot _also_ created individual PRs for each update, ending up with the grouped PR being auto-closed due to: 

> Looks like these dependencies are updatable in another way, so this is no longer needed.

Shouldn't it be the _other way_ around as the entire purpose of `groups` is to minimize the number of PRs being created. :confused:

jakecoffman:
Fixed in https://github.com/dependabot/dependabot-core/pull/8106

PR #8005, opened 2023-09-11
Update semver

jurre:
This transitive version of semver has a vulnerability, let's patch it


landongrindheim (approved):

yeikel:
Can we configure dependabot to take care of this? Or we don't do that intentionally?

Issue #7782, opened 2023-08-09
Sorbet type checking


JamieMagee:
### Code improvement description

[Sorbet][1] is a type checker for Ruby. It allows for gradual addition of types to an existing codebase and can work for both static and runtime type checking.

Currently, this issue is mainly a placeholder.

- [x] Migrate `Layout/DotPosition` from `trailing` to `leading`

    Sorbet made 1 intentional compatibility break with Ruby syntax relating to multiline method calls which match Ruby keywords. Unfortunately, this happens quite often for RSpec's matchers. For example `and`, `or`, etc.

    The fix is to switch from enforcing `trailing` dot position to `leading` with Rubocop, to ensure compatibility with Sorbet. This can be done automatically, but is going to result in a very large diff.

    - https://srb.help/2001
    - https://github.com/sorbet/sorbet/pull/1993
    - #7789 

- [x] The above change is going to result in a rather large git diff. To preserve the usefulness of `git blame` a `.git-blame-ignore-revs` file should be added at the root of the repository with the git object ID of the above commit.

    - https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/
    - #7992 

- [x] Install Sorbet and its dependencies.

    - https://sorbet.org/docs/adopting#step-1-install-dependencies
    - #8007 

- [x] Initialise Sorbet

    - https://sorbet.org/docs/adopting#step-2-initialize-sorbet-in-our-project
    - #8012

- [x] Fix any initial errors

    - https://sorbet.org/docs/adopting#step-4-fix-constant-resolution-errors

- [x] Enforce static type checking as part of the build process

    - https://sorbet.org/docs/adopting#step-5-enable-type-checking
    - #8017

- [x] Install and enable the Sorbet RuboCop rules

    - https://github.com/Shopify/rubocop-sorbet
    - #8016 

- [x] Add the Sorbet Visual Studio Code extension to the list of recommended extensions, as well as the Dev Container configuration

    - https://marketplace.visualstudio.com/items?itemName=sorbet.sorbet-vscode-extension
    - #8018

[1]: https://sorbet.org/

JamieMagee:
At this point, I think all of the initialization and setup is complete. Now the work of adding types begins.

yeikel:
Something I noticed is that the Sorbet scan always runs. Perhaps it should only run if there are changes to Ruby files? 

[...]

PR #8007, opened 2023-09-11
Add sorbet dependencies

JamieMagee:
Part of #7782 and step 1 of the [adopting Sorbet in an existing codebase][1] guide.

This will be followed with step 2 [initializing sorbet][2]

[1]: https://sorbet.org/docs/adopting#step-1-install-dependencies
[2]: https://sorbet.org/docs/adopting#step-2-initialize-sorbet-in-our-project

### DIFF
--- a/updater/Gemfile.lock
+++ b/updater/Gemfile.lock
@@ -27,6 +27,7 @@ PATH
       octokit (>= 4.6, < 7.0)
       parser (>= 2.5, < 4.0)
       psych (~> 5.0)
+      sorbet-runtime (~> 0.5)
       toml-rb (>= 1.1.2, < 3.0)
 
 PATH
@@ -282,6 +283,7 @@ GEM
       faraday (>= 0.17.3, < 3)
     sentry-raven (3.1.2)
       faraday (>= 1.0)
+    sorbet-runtime (0.5.11011)
     stringio (3.0.6)
     terminal-table (3.0.2)
       unicode-display_width (>= 1.1.1, < 3)
--- a/common/dependabot-common.gemspec
+++ b/common/dependabot-common.gemspec
@@ -39,6 +39,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "octokit", ">= 4.6", "< 7.0"
   spec.add_dependency "parser", ">= 2.5", "< 4.0"
   spec.add_dependency "psych", "~> 5.0"
+  spec.add_dependency "sorbet-runtime", "~> 0.5"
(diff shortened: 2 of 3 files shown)


Nishnha (commented):
Hmm not sure what the difference between adding these gems here versus https://github.com/dependabot/dependabot-core/blob/main/common/dependabot-common.gemspec is.

JamieMagee:
Yeah, I was torn too. But given the VSCode integration gems are installed here, it seemed like the more appropriate place for the Sorbet gems as well.

Nishnha (commented):
Well the `omnibus` uses the `common` gemspec so I guess it doesn't really matter where we install them as long as these tools are accessible from a `bin/docker-dev-shell` container

mctofu:
> Hmm not sure what the difference between adding these gems here versus https://github.com/dependabot/dependabot-core/blob/main/common/dependabot-common.gemspec is.

If we will need it to be available when running/testing the individual ecosystem gems it should go in dependabot-common.

JamieMagee:
Yes, they'll be needed to power both IDE integrations, as well as in individual ecosystems. I'll move them to common.

Nishnha:
Hmm maybe it's me but that didn't work. Did I run it right?

```
$ git checkout jamiemagee/add-sorbet
> branch 'jamiemagee/add-sorbet' set up to track 'jamie/jamiemagee/add-sorbet'.
> Switched to a new branch 'jamiemagee/add-sorbet'

$ script/build bundler
> [+] Building 42.0s (43/43) FINISHED

[dependabot-core-dev] ~ $ cd bundler
[dependabot-core-dev] ~/bundler $ bundle install
> Resolving dependencies...

[dependabot-core-dev] ~/bundler $ bundle exec sorbet
> bundler: command not found: sorbet
> Install missing gem executables with `bundle install`
```

Nishnha (approved):

updater/Gemfile.lock
```diff
@@ -27,6 +27,7 @@ PATH
       octokit (>= 4.6, < 7.0)
       parser (>= 2.5, < 4.0)
       psych (~> 5.0)
+      sorbet-runtime (~> 0.5)
```
jurre, line 30: Why are these (only) in the _updater_ lockfile? 🤔 I'd expect the to be in the `common` lockfile instead?

JamieMagee: `Gemfile.lock` is explicitly ignored in `.gitignore` _except_ for in `updater`:

https://github.com/dependabot/dependabot-core/blob/ff4b17e9b2c742987a252099c26d182ca2408eda/.gitignore#L9-L11

PR #8013, opened 2023-09-12
raise exceptions when PR creation fails

jakecoffman:
We sometimes see reports of PRs failing to create, but the logs have no information in them. I suspect part of this is due to these errors that are being swallowed up as expected.

So I've removed that and replaces a few blank `return`s with `raise` so we can see when those cases are happening.


common/lib/dependabot/pull_request_creator/github.rb
```diff
@@ -42,8 +42,8 @@ def initialize(source:, branch_name:, base_commit:, credentials:,
       end
 
       def create
-        return if branch_exists?(branch_name) && unmerged_pull_request_exists?
-        return if require_up_to_date_base? && !base_commit_is_up_to_date?
+        raise "Unmerged PR exists" if branch_exists?(branch_name) && unmerged_pull_request_exists?
+        raise "Base commit is not up to date" if require_up_to_date_base? && !base_commit_is_up_to_date?
```
abdulapopoola, line 46: [newbie q] will these errors get surfaced to the user or will they be currently visible in the logs?

jakecoffman: These will be in the logs, but there is another effort to show the errors to the users as well.

mctofu (approved):

PR #8014, opened 2023-09-12
Make sorbet and tapioca optional

JamieMagee:
Sorbet currently doesn't support `aarch64-linux`. Unfortunately, this means that it doesn't work on Docker on macOS, unless you force `--platform=linux/amd64` or `DOCKER_DEFAULT_PLATFORM=linux/amd64`.

This change makes `sorbet` and `tapioca` only install on supported platforms for now. This means that typechecking with `srb tc` will not work on Docker on macOS. However, while it's an optional, non-enforced, part of the project this is an acceptable trade-off.

Thankfully, it looks like official support for `aarch64-linux` is extremely close to being completed. So I anticipate this workaround will be shortlived.

References:
- https://github.com/sorbet/sorbet/issues/4119
- https://github.com/sorbet/sorbet-build-image/pull/8

### DIFF
--- a/common/Gemfile
+++ b/common/Gemfile
@@ -3,3 +3,8 @@
 source "https://rubygems.org"
 
 gemspec
+
+install_if -> { RUBY_PLATFORM.include?("darwin") || RUBY_PLATFORM.include?("x86_64") } do
+  gem "sorbet", group: :development
+  gem "tapioca", require: false, group: :development
+end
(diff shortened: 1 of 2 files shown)


jurre:
Hmm, that's a bummer, I believe the whole core team is on OSX using Docker :( our codespaces setup is currently not really sufficient, so we might need to invest there a little bit depending on how quickly support for this lands

JamieMagee:
@jurre do you know how it's split between ARM macs and x86_64 macs? Cause it's only Docker on ARM macOS that would be unable to run `srb tc` locally.

jakecoffman (approved):
I verified this builds for me on ARM. 

I suspect my editor will still have better completion even though I can't run Sorbet directly but we need a typed file first to test 😄

PR #8026, opened 2023-09-13
Actions: skip unsupported uses strings

jakecoffman:
This fixes a bug where if you name the local reusable workflow with an `@` then the path is parsed wrong.

For example with `./.github/workflows/test.yml@v2.1.0`. The code expects either `<owner>/<repo>@<version>` or `./.github/<path>`, but this parsed as `./.github/<path>@<version>` which is wrong since the `@v2.1.0` is just a part of the path. Thus we ended up with a dependency of `owner: '.github'` which broke the update when fetching the remote file.

Since we currently don't support updating the dependencies of local paths in this way, and also we don't support updating `docker://` workflows either, I'm explicitly skipping them for now to avoid any other latent issues. 

Conveniently there was a TODO that I could put this skip right after 😄.

### DIFF
--- a/github_actions/spec/dependabot/github_actions/file_parser_spec.rb
+++ b/github_actions/spec/dependabot/github_actions/file_parser_spec.rb
@@ -215,6 +215,14 @@ RSpec.describe Dependabot::GithubActions::FileParser do
       end
     end
 
+    describe "with a local reusable workflow that has an @ in the file name" do
+      let(:workflow_file_fixture_name) { "local_workflow_with_at.yml" }
+
+      it "does not treat the path like a dependency" do
+        expect(dependencies).to eq([])
+      end
+    end
+
     describe "with composite actions" do
       let(:workflow_file_fixture_name) { "composite_action.yml" }
       let(:workflow_files) do
--- a/github_actions/lib/dependabot/github_actions/file_parser.rb
+++ b/github_actions/lib/dependabot/github_actions/file_parser.rb
@@ -46,6 +46,7 @@ module Dependabot
 
         uses_strings.each do |string|
           # TODO: Support Docker references and path references
+          next if string.start_with?(".") || string.start_with?("docker://")
           next unless string.match?(GITHUB_REPO_REFERENCE)
 
           dep = build_github_dependency(file, string)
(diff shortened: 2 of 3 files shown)


abdulapopoola (approved):

github_actions/spec/dependabot/github_actions/file_parser_spec.rb
```diff
@@ -215,6 +215,14 @@ def mock_service_pack_request(nwo)
       end
     end
 
+    describe "with a local reusable workflow that has an @ in the file name" do
+      let(:workflow_file_fixture_name) { "local_workflow_with_at.yml" }
```
landongrindheim, line 219: I'm not clear on why the `@` has significance to us in this scenario. My understanding of the change, and the behavior it's fixing, is that the locality (i.e. the `.`) is more significant. We wouldn't attempt to update `./.github/workflows/test.yml@2.1.0` because it's co-located in the repo. Am I reading that correctly? If so, I think it makes sense to leave off the `@` detail as it suggests the crux of the issue is with the version. What do you think?

landongrindheim (approved):

github_actions/spec/dependabot/github_actions/file_parser_spec.rb
jakecoffman: Fair enough, it's just that the `@` is what was causing it to be an issue.

landongrindheim: > Fair enough, it's just that the @ is what was causing it to be an issue.

@jakecoffman By this, do you mean that the `@` was an issue for this particular user? Or that it's an issue that's widely seen? I definitely don't want to push you away from your instincts/what you think is right!

I guess another way to ask this would be: what would Dependabot do if the `@2.1.0` was missing?

jakecoffman: If there was no @ then there was no bug, it just happened to work I think. The @ made the bug come out because of [this regex](https://github.com/dependabot/dependabot-core/blob/742ba48c115295042564ba97bfd5d3af82d38033/github_actions/lib/dependabot/github_actions/file_parser.rb#L23-L24). 

In the end though, the @ isn't important. The important thing is we're ignoring paths as dependencies because Dependabot can't update them.

PR #8028, opened 2023-09-13
Add types to clients

JamieMagee:
Mainly adding `typed: true` to clients, but also includes:

- Making `dependabot.rb` and `wildcard_matcher.rb` `typed: strong` mostly as a learning experience to see what is required at that type enforcement level
- Define stubs of `to_json` for `Array` and `Hash`
- Define stubs for `AWS` errors which Sorbet is unable to determine (See [The TODO RBI file][1] for more details)
- Wrap `*_with_retries` clients with `T.unsafe` as they explicitly rely on `method_missing` which Sorbet can't deal with

#7782

[1]: https://sorbet.org/docs/rbi#the-todo-rbi-file

### DIFF
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -3,6 +3,7 @@
 
 require "dependabot/shared_helpers"
 require "excon"
+require "sorbet-runtime"
 
 module Dependabot
   module Clients
--- a/common/lib/dependabot/clients/azure.rb
+++ b/common/lib/dependabot/clients/azure.rb
@@ -1,4 +1,4 @@
-# typed: false
+# typed: true
 # frozen_string_literal: true
 
 require "dependabot/shared_helpers"
@@ -7,6 +7,8 @@ require "excon"
 module Dependabot
   module Clients
     class Azure
+      extend T::Sig
+
       class NotFound < StandardError; end
 
       class InternalServerError < StandardError; end
@@ -253,8 +255,9 @@ module Dependabot
         JSON.parse(response.body).fetch("value")
       end
 
+      sig { params(url: String).returns(Excon::Response) }
       def get(url)
-        response = nil
+        response = T.let(nil, T.nilable(Excon::Response))
 
         retry_connection_failures do
           response = Excon.get(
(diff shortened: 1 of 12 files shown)


common/lib/dependabot/clients/azure.rb
```diff
@@ -253,8 +255,9 @@ def compare(previous_tag, new_tag, type)
         JSON.parse(response.body).fetch("value")
       end
 
+      sig { params(url: String).returns(Excon::Response) }
       def get(url)
-        response = nil
+        response = T.let(nil, T.nilable(Excon::Response))
```
jurre, line 260: Are these `T.let`s and `T.must`s checked at runtime as well, or are these also just compile-time checks?

jurre:
This one seems like a change that we should test in staging, and deploy before merging, because it's actually changing code ran in prod, as opposed to the previous PRs that were mainly no-ops

common/lib/dependabot/clients/azure.rb
JamieMagee: `T.let` and `T.must` are both static and runtime checks. Unfortunately, it's not possible to satisfy Sorbet's type checks with just static checks in this specific instance.

Sorbet infers `response = nil` is of type `nil`, and then complains that `Excon::Result` cannot be assigned to type `nil`. So `response` has to be defined as `T.nilable(Excon::Response)`.

Later we make calls to `response.status` and `response.body`. I made the `response.status` calls use the safe navigation operator `&.`. But before returning from the method, I assert that `response` is not `nil` by calling `T.must(response)`. The alternative is propagating the nil all the way up the stack, which will require a lot of condditionals where the result of a method call _might_ be `nil`.

jurre: I see, thanks for explaining! It might make sense to enable the runtime checking only in tests and not in prod yet if there's a way to make that configurable? That way we reduce the chance of introducing any new issues, and also I believe there's a (small?) cost incurred when enabling runtime checks?

JamieMagee: Yeah, I can add that configuration somewhere in the root of `common`. There is a small performance penalty for runtime type checks, but it's somewhere in the region of 5%[^1]

> we have runtime type checking turned on even for our most performance-sensitive request paths, and the slowdown from runtime type checking amounts to about 5%.

And if we ever end up using the Sorbet compiler to AOT compile `dependabot-core` we'll see huge performance _increases_ with runtime type checks[^2]

[^1]: https://blog.jez.io/runtime-type-checking/
[^2]: https://sorbet.org/blog/2021/07/30/open-sourcing-sorbet-compiler

JamieMagee:
Superseded by #8038 (as I created this from the `main` branch of my fork 🤦)

PR #8029, opened 2023-09-13
fix docker-dev-shell on ARM

jakecoffman:
Fixes the freeze on `bundle install` when running `bin/docker-dev-shell common --rebuild ` on ARM. 

Not sure why `install_if` worked in common but doesn't work in omnibus. Plain ole `if` seems ok though!


JamieMagee:
I don't have permissions to approve, but this works locally for me, and it works in CI too. Thanks!

mctofu (approved):

PR #8032, opened 2023-09-13
Add sig to dependency injection containers

JamieMagee:
Related to #7782

### DIFF
--- a/common/lib/dependabot/file_fetchers.rb
+++ b/common/lib/dependabot/file_fetchers.rb
@@ -2,6 +2,7 @@
 # frozen_string_literal: true
 
 require "sorbet-runtime"
+require "dependabot/file_fetchers/base"
 
 module Dependabot
   module FileFetchers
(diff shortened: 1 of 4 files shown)


Nishnha:
It might be a good idea to create a package_manager enum for type hinting? I don't think we have a canonical list in Dependabot Core outside of https://github.com/dependabot/dependabot-core/blob/1b6f1e4cdff6f6d8e2a9ab76d87cb16b20ecf55d/updater/lib/dependabot/updater.rb#L14-L31

jakecoffman (approved):
Looks good!

@Nishnha The updater only calls method on the base so I don't think that's necessary?

JamieMagee:
I think a package manager enum is a good idea, but I'm focusing on adding `sig`s and less on general refactoring

Nishnha (approved):

jakecoffman:
Starting to deploy this now.

PR #8037, opened 2023-09-14
fix: call 'split' on string-type object, not on version-type object

fredrikaverpil:
This hopefully fixes #8003 which exhibits the following error:

```
updater | 2023/09/14 09:33:53 ERROR <job_721897648> Error processing pytz (NoMethodError)
updater | 2023/09/14 09:33:53 ERROR <job_721897648> undefined method `split' for #<Dependabot::Python::Version 2023.3.post1>
updater | 
updater |         version, @local_version = version.split("+")
updater |                                          ^^^^^^
```

The error message is indicating that the `split` method is being called on an instance of `Dependabot::Python::Version`, not a string. The `split` method is a string method, so it cannot be called on an object of a different type unless it implements its own "split" method, which this "Version" type doesn't seem to do.

cc @carogalvin 👋 😅 sorry for pinging you here but this seems to be a showstopper for grouped dependabot PRs with Python at the moment. I started seeing this error as I added grouped dependencies like so:

```diff
version: 2

updates:
  - package-ecosystem: "github-actions"
    directory: "/"
    registries: "*"
    schedule:
      interval: "weekly"
      day: "tuesday"
  - package-ecosystem: "pip"
+     groups:
+       production-dependencies:
+         dependency-type: "production"
+         update-types:
+           - minor
+           - patch
+       development-dependencies:
+         dependency-type: "development"
+         update-types:
+           - minor
+           - patch
    insecure-external-code-execution: allow
    directory: "/"
    registries: "*"
    schedule:
      interval: "weekly"
      day: "tuesday"
    ignore:
      - dependency-name: "dr"
    open-pull-requests-limit: 50
    allow:
      - dependency-name: "*"
        dependency-type: all
    commit-message:
      prefix: "chore"
      include: "scope"
    versioning-strategy: increase-if-necessary

registries:
  github:
    type: git
    url: https://github.com/
    username: x-access-token
    password: ${{secrets.REDACTED_TOKEN}}

  cloudsmith:
    type: python-index
    url: https://dl.cloudsmith.io/<REDACTED>/<ORG>/<REPO>/python/simple/
    replaces-base: true
```

I'm no expert in ruby, so take this PR as just an indication as to what might be wrong 😅

PR #8039, opened 2023-09-14
Add types to requirement classes

JamieMagee:
Part of #7782 

I stopped adding safe navigation operators (`&.`) and type assertions ([`T.must`][1]) and started using the escape hatch ([`T.unsafe`][2]). Mainly because I am unsure what the expected behaviour is supposed to be in. My main goal with this project is to get as much as possible to as high a type checking level as possible. `T.unsafe` can be dealt with at a later date when attempting to get to [`strong`][3].

[1]: https://sorbet.org/docs/type-assertions#tmust
[2]: https://sorbet.org/docs/troubleshooting#tunsafe
[3]: https://sorbet.org/docs/static#file-level-granularity-strictness-levels

### DIFF
--- a/bundler/lib/dependabot/bundler/requirement.rb
+++ b/bundler/lib/dependabot/bundler/requirement.rb
@@ -1,4 +1,4 @@
-# typed: true
+# typed: strong
 # frozen_string_literal: true
 
 require "dependabot/utils"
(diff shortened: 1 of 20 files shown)


jakecoffman:
Nice, we just had a bug yesterday that types could have caught: https://github.com/dependabot/dependabot-core/pull/8037

JamieMagee:
Will split this into smaller PRs

PR #8043, opened 2023-09-15
raise if the reference already exists

jakecoffman:
This was missed in #8013.

Currently it returns `nil` but then raises an UnexpectedError.


common/lib/dependabot/pull_request_creator/github.rb
```diff
@@ -248,7 +248,7 @@ def create_branch(commit)
           branch
         rescue Octokit::UnprocessableEntity => e
           # Return quietly in the case of a race
```
jurre, line 250: This now no longer applies, right?

```suggestion
```

jakecoffman: Yeah now we return loudly.

jurre (approved):

PR #8094, opened 2023-09-26
Properly infer `.npmrc` for PNPM

deivid-rodriguez:
There are two issues:

* PNPM lockfile updater is not using `NpmrcBuilder` to properly configure private registries. This was an oversight from the initial PNPM implementation since all other lockfile updaters (NPM and Yarn) do it.

This issue is fixed by cherry-picking #8091 from @mburumaxwell.

* Even if an npmrc file is generated, the logic for generating it is incomplete. The `NpmrcBuilder` only looks in the lockfile for finding the proper registry vs scope configuration, but can't work if the lockfile does not include registries or if parsing a specific lockfile is not implemented. However, we have a smarter class, `RegistryFinder`, that can also infer the proper registry by directly querying it and checking which dependencies it serves. This class is already used by the `UpdateChecker` and there's a TODO note on top of `NpmrcBuilder` to refactor it to use `RegistryFinder`: https://github.com/dependabot/dependabot-core/blob/0d64ef28006599f59147c107266719bfc3edc0cb/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npmrc_builder.rb#L9-L11.

The second commit refactors `NpmrcBuilder` to use `RegistryFinder`, but just for PNPM. Once I validate that this works properly, I can migrate NPM & Yarn too and completely remove lockfile parsing code from this class.

This PR should fix #7731.

PR #8095, opened 2023-09-26
Bypass review decision for explicitly requested builds

deivid-rodriguez:
I'd like to try #8094 in staging, but currently I'd need someone to "dummy-approve" it for me so that the updater images are uploaded. If I do this, the review loses its point.

Would it make sense to not require an approval if I explicitly trigger the workflow to build updater images?


honeyankit (approved):

deivid-rodriguez:
Allowing ourselves to deploying unreviewed PRs, even if only to staging, is something not everybody is comfortable with. I'll test the PR in staging once reviewed.

PR #8068, opened 2023-09-19
Respect style of each action when mixed styles are used

deivid-rodriguez:
Fixes #7971.

For example, on https://github.com/getlogbook/logbook:

#### Before

    ± .github/workflows/wheel-builder.yml
    ~~~
    --- /tmp/original20230919-234-sgikwf	2023-09-19 17:36:17.144755893 +0000
    +++ /tmp/updated20230919-234-cdm607	2023-09-19 17:36:17.144755893 +0000
    @@ -25,7 +25,7 @@
         name: Build sdist
         runs-on: ubuntu-latest
         steps:
    -      - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
    +      - uses: actions/checkout@v4 # v3.5.3
             with:
               # The tag to build or the tag received by the tag event
               ref: ${{ github.event.inputs.version || github.ref }}
    @@ -50,7 +50,7 @@
             os: [ubuntu-20.04, windows-2019, macos-11]
     
         steps:
    -      - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
    +      - uses: actions/checkout@v4 # v3.5.3
             with:
               # The tag to build or the tag received by the tag event
               ref: ${{ github.event.inputs.version || github.ref }}
    ~~~
    3 insertions (+), 3 deletions (-)

#### After

    ± .github/workflows/wheel-builder.yml
    ~~~
    --- /tmp/original20230919-213-acjvhd	2023-09-19 17:35:11.844596815 +0000
    +++ /tmp/updated20230919-213-juz0p3	2023-09-19 17:35:11.844596815 +0000
    @@ -25,7 +25,7 @@
         name: Build sdist
         runs-on: ubuntu-latest
         steps:
    -      - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
    +      - uses: actions/checkout@72f2cec99f417b1a1c5e2e88945068983b7965f9 # v4.5.4
             with:
               # The tag to build or the tag received by the tag event
               ref: ${{ github.event.inputs.version || github.ref }}
    @@ -50,7 +50,7 @@
             os: [ubuntu-20.04, windows-2019, macos-11]
     
         steps:
    -      - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
[...]

PR #8098, opened 2023-09-27
Fix some github actions version comments not getting updated

deivid-rodriguez:
See https://github.com/rubygems/rubygems/pull/6995 for an example.

Since there is an old action pinned to `actions/checkout@v3.6.0`, that's the version the dependency takes, and the one that tries to be replaced. However, all comments are using `# v4.0.0`, since that's what the other actions are using, so the comment replace fails.

Instead, respect the version each action is pinned to when performing the replace, keeping the spirit of #8068.


jurre (approved):

deivid-rodriguez:
It worked for the RubyGems PR, so merging!

PR #8104, opened 2023-09-28
Raise unsupported Python version error as an expected error

deivid-rodriguez:


jurre (approved):

common/lib/dependabot/errors.rb
```diff
@@ -76,6 +76,20 @@ def initialize(source, msg = nil)
...
+  class ToolVersionNotSupported < DependabotError
+    attr_reader :tool_name, :detected_version, :supported_versions
+
+    def initialize(tool_name, detected_version, supported_versions)
+      @tool_name = tool_name
+      @detected_version = detected_version
+      @supported_versions = supported_versions
+
+      msg = "Dependabot detected the following #{tool_name} requirement for your project: '#{detected_version}'." \
+            "\n\nCurrently, the following Python versions are supported in Dependabot: #{supported_versions}."
```
abdulapopoola, line 88: should this be `tool` instead of the hardcoded Python?

deivid-rodriguez: Oh wow, nice catch! Let me update it!

jeffwidman:
Just curious what the motivation was here as the PR says "as an expected error" but that was already true... ? So this PR instead makes it a new class of error. Is that for internal metrics or greater clarity to the user? 

To be clear, I like the change as `DependencyFileNotResolvable` always felt weird here since it probably_is_resolvable, just not supported by :Dependabot:.

deivid-rodriguez:
Hei! Good to hear from you :)

This error was raised from the fetcher step, not the updater step, and the code in `/updater` does not re-raise `DependencyFileNotResolvable` errors as expected errors (only the updater does). So it was being raised as an internal error instead and this error was showing up in our error tracker instead of being displayed the users in the UI.

deivid-rodriguez:
An alternative fix would've been to add a block [here](https://github.com/dependabot/dependabot-core/blob/22f94ac0044a3db135cc7449bf23b6437dce9647/updater/lib/dependabot/file_fetcher_command.rb#L123-L131) for `DependencyFileNotResolvable` like the one I added for the new error.

But I decided to create a separate error class since it felt cleaner.

jeffwidman:
Hah, that was exactly going to be my follow-up question. Thanks for the clarification!

jeffwidman:
Also, I missed you folks! Last night was the first night I slept longer than four hours at a stretch since the baby arrived, so that's why I finally had the energy to see what's up around these parts.

PR #8105, opened 2023-09-28
Fix CI

deivid-rodriguez:
A new version of RuboCop was released, and broke our CI.

We generally want to proactively update RuboCop, and not have to react to it breaking our CI. So I'd rather keep the version we run locked.

To keep it locked, we have to use the `updater/Gemfile`, since that's the one that includes a lockfile.

To do that we configure `BUNDLE_GEMFILE` in CI to use that.

I decided to also add a rubocop binstub to make it easier to run rubocop either outside (like CI does) or inside our development image.


deivid-rodriguez:
Pushed one extra fix and this should be now green!

PR #8127, opened 2023-10-02
Fix version comments after quoted strings

### LINKED ISSUE #8125: YAML string quotes prevent version comment updates

### Package ecosystem

GitHub Actions

### Package manager version

_No response_

### Language version

_No response_

_No response_

### dependabot.yml content

_No response_

### Updated dependency

_No response_

I tried using git SHAs with version comments and expected Dependabot to keep the version comment updated. However, when YAML string quotes are present, the version comment doesn't get updated. Easy example:

```yaml
- uses: "actions/checkout@01aecccf739ca6ff86c0539fbc67a7a5007bbc81" # v2.1.0
```

Dependabot will submit a PR to update only the git SHA, but not the comment.

_No response_

_No response_

```yaml
on: ["push"]

name: "Integration"
jobs:
  chore:
    steps:
      - uses: "actions/checkout@01aecccf739ca6ff86c0539fbc67a7a5007bbc81" # v2.1.0
```

kurtmckee:
When YAML strings are quoted, Dependabot fails to update trailing version comments. For example, Dependabot will update the SHA but not the "v2.1.0" comment below:

```yaml
- uses: "actions/checkout@01aecccf739ca6ff86c0539fbc67a7a5007bbc81" # v2.1.0
```

This PR fixes the bug in the associated regex.

In addition, this PR removes a test case using a short SHA (`"01aecc"`) that is no longer updated (a problem that the test suite doesn't notice) and isn't supported by GitHub Actions anymore.

It also updates a short-SHA-based test case (`"01aecc#v2.1.0"`) to use a full-length SHA so that it is more likely to catch bugs that might be introduced.

Fixes #8125

### DIFF
--- a/github_actions/lib/dependabot/github_actions/file_updater.rb
+++ b/github_actions/lib/dependabot/github_actions/file_updater.rb
@@ -78,7 +78,7 @@ module Dependabot
           updated_content =
             updated_content
             .gsub(
-              /(?<=\W|"|')#{Regexp.escape(old_declaration)}(?<comment>\s+#.*)?(?=\s|"|'|$)/
+              /(?<=\W|"|')#{Regexp.escape(old_declaration)}["']?(?<comment>\s+#.*)?(?=\s|$)/
             ) do |match|
               comment = Regexp.last_match(:comment)
               match.gsub!(old_declaration, new_declaration)
--- a/github_actions/spec/dependabot/github_actions/file_updater_spec.rb
+++ b/github_actions/spec/dependabot/github_actions/file_updater_spec.rb
@@ -374,11 +374,11 @@ RSpec.describe Dependabot::GithubActions::FileUpdater do
         it "updates SHA version" do
           old_sha = dependency.previous_requirements.first.dig(:source, :ref)
           expect(subject.content).to include "#{dependency.name}@#{dependency.requirements.first.dig(:source, :ref)}"
-          expect(subject.content).not_to match(/#{old_sha}\s+#.*#{dependency.previous_version}/)
+          expect(subject.content).not_to match(/#{old_sha}['"]?\s+#.*#{dependency.previous_version}/)
         end
         it "updates version comment" do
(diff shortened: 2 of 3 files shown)


deivid-rodriguez (approved):
Awesome, nice catch!

kurtmckee:
My commit verification settings are causing the commits to be flagged as "Unverified". This won't be an issue if this is squash-merged, but if not then I can rebase on `main` to ensure the branch is up-to-date _and_ has valid GPG signatures.

Let me know if that would be helpful, depending on how PRs are merged. :+1:

deivid-rodriguez:
I was planning to use standard Merge, so feel free to rebase, thanks! 🙏

kurtmckee:
Thanks for your patience! I've rebased on `main` and force-pushed. I'll monitor the test suite runs, but please let me know if this needs any additional work!

kurtmckee:
Test suite has passed. :partying_face:

deivid-rodriguez:
@kurtmckee I just deployed this, would you be able to confirm it's now working properly?

kurtmckee:
I don't know how to check that except to wait for Dependabot's scheduled checks.

Is there a way to trigger this against a test repository?

deivid-rodriguez:
That works!

Also, if you go to "Insights > Dependency Graph > Dependabot" of an existing repo with dependabot enabled, you should be able to manually check for updates.

Also, on an existing PR where this bug was triggered, commenting `@dependabot recreate` should retry using the new code.

kurtmckee:
[This appears to be working!](https://github.com/kurtmckee/listparser/pull/61/files) :partying_face: 

image

deivid-rodriguez:
Awesome! Can you rebase one last time? We require up to date PRs for merging and one other PR sneaked into main while you were testing this 😅

kurtmckee:
No worries at all. Thanks for sharing how to manually check for updates!

deivid-rodriguez:
Thanks so much for the fix!

PR #8128, opened 2023-10-02
grouped security updates

jakecoffman:
This is a grouped security updates prototype.

It detects a group by seeing multiple dependencies are passed into a security update job. Then it creates a grouping on-the-fly to generate a single PR instead of the usual multiple PRs. 

This is a feature in development, so there's no way to exercise it without flags set in the service.

### DIFF
--- a/updater/lib/dependabot/updater/group_update_creation.rb
+++ b/updater/lib/dependabot/updater/group_update_creation.rb
@@ -190,7 +190,7 @@ module Dependabot
           repo_contents_path: job.repo_contents_path,
           credentials: job.credentials,
           ignored_versions: job.ignore_conditions_for(dependency),
-          security_advisories: [], # FIXME: Version updates do not use advisory data for now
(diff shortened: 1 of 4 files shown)


jurre (approved):
Approving this so I can deploy it to staging

@jakecoffman could be nice to throw some feature flag checks in here so we could deploy it to prod if we wanted?

jakecoffman:
I can do that. I was wondering if it was necessary though? If a security job can only come through with a single dependency specified in the job then I think we're good without a feature flag.

jurre:
> I can do that. I was wondering if it was necessary though? If a security job can only come through with a single dependency specified in the job then I think we're good without a feature flag.

Yeah that's fair!

jakecoffman:
I queried all update jobs and verified security jobs never have more than one dependency. So this shouldn't interfere with existing logic, and doesn't need a feature flag.

jurre (approved):

PR #8134, opened 2023-10-03
Suppress for `Layout/MultilineMethodCallIndentation` offense

ydah:


deivid-rodriguez:
It sounds like https://github.com/rubocop/rubocop/issues/12199 from RuboCop 1.56.4 is catching these.

I'm curious, how did you find about these errors? Our project is using RuboCop 1.56.3 at the moment, which does not raise offenses for those.

Can you also upgrade RuboCop in `updater/Gemfile.lock` as part of this PR, so that our CI catches these?

deivid-rodriguez:
> Without thinking, I moved directly under the ./bundler directory, ran bundle install, and then ran rubocop.

Aaaah, I see. I will look into making all ecosystems use the lockfile in `updater` by default.

PR #8138, opened 2023-10-04
Make sure tests use the same gem versions we use in production

deivid-rodriguez:
In #8134 it was noticed that running `rubocop` from an ecosystem directory, will run the latest version of RuboCop, not the ones we have locked at `updater/Gemfile.lock`.

This PR is an experiment to try to make sure a consistent version of rubocop is run all around while also simplifying our setup.


Nishnha (approved):
These changes made a lot more sense after I saw your comment in https://github.com/dependabot/dependabot-core/pull/8134#issuecomment-1745396660

I think this will work!

deivid-rodriguez:
Thanks @Nishnha, yes, I think this will work.

However, turns out some specs depend on a `Gemfile` file being present at the root of an ecosystem. This seems unnecessary and confusing, and I've also been bitten by similar issues when running specs on my Linux laptop, but it requires some extra work to fix it.

deivid-rodriguez:
I think this is good to go now.

Doing this would prevent sudden introduction of broken tests by external releases, like it happened at #8509. That workaround will no longer be necessary with this PR, although I'm not reverting it, because we do use `json` directly, so it's a runtime dependency after all.

Also, note that in my opinion the best solution here would be for each ecosystem to have its own Gemfile & Gemfile.lock file, but without multidir version updates group PRs, it's going to get very noisy, so reusing `updater/` lockfile seems good enough for now?

jakecoffman (approved):

deivid-rodriguez:
Let's see how this setup works!

PR #8142, opened 2023-10-04
Catch up test lockfile with parser 3.2.2.4 release

deivid-rodriguez:
The new parser 3.2.2.4 release has caused a new test failure.


jurre (commented):
👍 for now. Anyway we can make this less dependent on latest released version, for example by specifying a range in the Gemfile?

deivid-rodriguez:
I thought of that but I imagined that if I specify a range, it will also be updated and still cause the issue? Maybe change the Gemfile to use a dev dependency that does not have transitive dependencies?

jakecoffman (approved):

deivid-rodriguez:
Thanks @jakecoffman!

Happy to make this test more robust but merging for now just to get CI back to green!

PR #8143, opened 2023-10-04
fix completely ignored dependencies querying for updates

### LINKED ISSUE #7651: Ignore doesn't ignore

### Package ecosystem

Docker

### dependabot.yml content

  - package-ecosystem: "docker"
    directory: "/"
    schedule:
      interval: "daily"
    ignore:
      - dependency-name: "xyz/*"

At the moment the dependencies are still checked and appear in the logs as
```
updater | 2023/07/19 09:16:17 INFO <job_697229887> Checking if xyz latest needs updating
updater | 2023/07/19 09:16:17 INFO <job_697229887> Ignored versions:
updater | 2023/07/19 09:16:17 INFO <job_697229887>   >= 0 - from .github/dependabot.yml
updater | 2023/07/19 09:16:17 INFO <job_697229887> Latest version is latest
 proxy | 2023/07/19 09:16:17 [020] HEAD xyz
```
As the check might fail intentionally (let's say because of a dependency in a private repository), the dependency would still fail the dependabot task (and count for the 30 trials until dependabot refuses to work).

I would expect them to be ignored without a check and error.

jakecoffman:
Fixes #7651

If you try to ignore all versions of a dependency that is otherwise allowed to update, Dependabot will still query the registry to get a list of versions.

This becomes a problem if the dependency source is an un-configured private registry because it will fail the HTTP request and cause an error on the job which is completely avoidable if Dependabot had skipped the check.

This PR adds a check for completely ignored dependencies in `allowed_update?` which will fix the issue for individual PRs and grouped updates. 

Future enhancements to this could include checking if the range provided means no update is possible, like if the dependency is at `1.0.0` and the ignored version range is `> 1.0.0`.

### DIFF
--- a/updater/lib/dependabot/job.rb
+++ b/updater/lib/dependabot/job.rb
@@ -174,7 +174,7 @@ module Dependabot
     def allowed_update?(dependency)
       # Ignoring all versions is another way to say no updates allowed
       if completely_ignored?(dependency) && !security_updates_only?
-        Dependabot.logger.info("All versions of #{dependency.name} are ignored")
+        Dependabot.logger.info("All versions of #{dependency.name} ignored, no update allowed")
         return false
       end
 
--- a/updater/lib/dependabot/job.rb
+++ b/updater/lib/dependabot/job.rb
@@ -171,6 +171,7 @@ module Dependabot
     # separately, if required.
     #
     # rubocop:disable Metrics/PerceivedComplexity
+    # rubocop:disable Metrics/CyclomaticComplexity
     def allowed_update?(dependency)
       # Ignoring all versions is another way to say no updates allowed
       if completely_ignored?(dependency) && !security_updates_only?
@@ -206,6 +207,7 @@ module Dependabot
       end
     end
     # rubocop:enable Metrics/PerceivedComplexity
+    # rubocop:enable Metrics/CyclomaticComplexity
 
     def vulnerable?(dependency)
       security_advisories = security_advisories_for(dependency)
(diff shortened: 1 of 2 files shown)


updater/lib/dependabot/job.rb
```diff
@@ -171,7 +171,14 @@ def reject_external_code?
     # separately, if required.
     #
     # rubocop:disable Metrics/PerceivedComplexity
+    # rubocop:disable Metrics/CyclomaticComplexity
     def allowed_update?(dependency)
+      # Ignoring all versions is another way to say no updates allowed
+      if completely_ignored?(dependency) && !security_updates_only?
```
abdulapopoola, line 177: should this condition be in the `completely_ignored` method? Also, for my curiosity, any reason why security_updates have to be false?

Finally; any correlation between this and the ongoing alert rules by @landongrindheim and @bdragon ?

abdulapopoola (approved):
approving with comments

updater/lib/dependabot/job.rb
deivid-rodriguez: I had the same thought as @abdulapopoola, but also `ignore_conditions_for` method is already passing around the `security_updates_only?` flag, so maybe whatever the correct behavior for security updates is, it is already being considered?

jakecoffman: Ah, didn't notice it was passed in `ignore_conditions_for` already, thanks @deivid-rodriguez I'll remove it.

jakecoffman: For alert rules, we shouldn't even get in the job, it will ignore before creating the job itself.

PR #8144, opened 2023-10-04
Added record update job error api back to capture unknown errors 

honeyankit:
## Context

- This PR adds the new feature flag named `record_update_job_unknown_error`. It has been created to be always enabled when running on our cloud infrastructure but remains inactive for GHES. The reason behind this differentiation is that older versions of GHES do not support the functionality of reporting unknown errors. This change ensures smooth operation across various environments without compromising on functionality where supported.

- This feature flag will be used to skip any `api_client.record_update_job_unknown_error` call to capture the unknown error when running in GHES environment.

- This new feature flag `record_update_job_unknown_error` is already created in Dependabot-api and deployed to production.

- The [PR](https://github.com/github/dependabot-updates/issues/3650#:~:text=EE-,Track%20unknown%20errors%20dependabot/dependabot%2Dcore%237534,-EE) which introduced the new api `record_update_job_unknown_error` for reference.

### DIFF
--- a/updater/lib/dependabot/service.rb
+++ b/updater/lib/dependabot/service.rb
@@ -49,8 +49,7 @@ module Dependabot
       client.record_update_job_error(error_type: error_type, error_details: error_details)
     end
 
-    def record_update_job_unknown_error(error_type:, error_details:, dependency: nil)
-      @errors << [error_type.to_s, dependency]
+    def record_update_job_unknown_error(error_type:, error_details:)
       client.record_update_job_unknown_error(error_type: error_type, error_details: error_details)
     end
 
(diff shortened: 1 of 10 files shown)


updater/lib/dependabot/service.rb
```diff
@@ -49,8 +49,7 @@ def record_update_job_error(error_type:, error_details:, dependency: nil)
       client.record_update_job_error(error_type: error_type, error_details: error_details)
     end
 
-    def record_update_job_unknown_error(error_type:, error_details:, dependency: nil)
-      @errors << [error_type.to_s, dependency]
+    def record_update_job_unknown_error(error_type:, error_details:)
```
deivid-rodriguez, line 52: I removed adding to `@errors` here because otherwise the PR summary displayed in job logs was showing two duplicated errors. And that was making one spec failing that checked for a "Dependabot encountered '1' error" log.

honeyankit: thank you so much. this was getting super hard to crack it for me atleast.

mctofu (approved):

PR #8146, opened 2023-10-06
Add Manual Trigger for Gem Release

honeyankit:
- Add Manual Trigger for Gem Release. The gem release action workflow is failing with error missing tag. The tag is already present. I do not have any other way to cleanly trigger this workflow so adding this PR to manually trigger the gem release.
- Add `edited` to Release event, which means when the release tag in Dependabot Core is edited this action workflow will be triggered as well.

### DIFF
--- a/.github/workflows/gems-release-to-rubygems.yml
+++ b/.github/workflows/gems-release-to-rubygems.yml
@@ -4,7 +4,7 @@ on:  # yamllint disable-line rule:truthy
     # It's fine to trigger on every release because if we tag a release w/o
     # bumping the Gem version, RubyGems will reject it with an error that the
     # version is already live.
-    types: [published]
+    types: [published, edited]
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/gems-release-to-rubygems.yml
```diff
@@ -5,6 +5,7 @@ on:  # yamllint disable-line rule:truthy
     # bumping the Gem version, RubyGems will reject it with an error that the
     # version is already live.
     types: [published]
```
honeyankit, line 7:
```suggestion
    types: [published, edited]
```

deivid-rodriguez:
@honeyankit I have an idea of the root cause for this. The `actions/checkout` action has a `fetch-tags` option that defaults to false. Maybe that's why the release workflow is not finding tags? I propose to close this PR since you now managed to release and try out that fix for the root cause to see if it works fine next time.

honeyankit:
@deivid-rodriguez : Thank you for suggestion. I will close this PR. How did you come to conclusion that fetch-tag was the issue because some of the gems were created successfully with the new tag but then in the end it was failing with error that tag not found?

honeyankit:
> here

I will open another PR with your purpose fix and see if it works! Thank you.

PR #8147, opened 2023-10-06
Raise expected errors when unsupported PNPM versions are used

### LINKED ISSUE #7500: Raise an error if we find unsupported PNPM versions

### Feature description

If we find an app locked to an unsupported PNPM version, or using an unsupported PNPM lockfile format, we should raise an error instead of letting "undefined behavior" happen.

We currently support PNPM 7 and 8, as per https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates#supported-repositories-and-ecosystems.

deivid-rodriguez:
Closes #7500.

I expect this to fix a bunch of errors in our internal board, but of course will also break for repos where it's actually working.

I don't think this needs a deprecation changelog since we have [documented](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem) v7 and v8 as the only supported versions from the beginning.

#7298 would make transition easier for users but I think pnpm v6 is old enough so that this PR does not need to block on that feature.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
@@ -184,7 +184,13 @@ module Dependabot
       def pnpm_version
         return @pnpm_version if defined?(@pnpm_version)
 
-        @pnpm_version = package_manager.requested_version("pnpm") || guess_pnpm_version
+        version = package_manager.requested_version("pnpm") || guess_pnpm_version
+
+        if version && Version.new(version.to_s) < Version.new("7")
+          raise ToolVersionNotSupported.new("PNPM", version.to_s, "7.*, 8.*")
+        end
+
+        @pnpm_version = version
       end
 
       def guess_pnpm_version
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
@@ -190,7 +190,7 @@ module Dependabot
       def guess_pnpm_version
         return unless pnpm_lock
 
-        Helpers.pnpm_major_version
+        Helpers.pnpm_version_numeric(pnpm_lock)
       end
 
       def package_manager
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
@@ -168,7 +168,7 @@ module Dependabot
       def yarn_version
         return @yarn_version if defined?(@yarn_version)
 
-        @yarn_version = package_manager.locked_version("yarn") || guess_yarn_version
+        @yarn_version = package_manager.requested_version("yarn") || guess_yarn_version
       end
 
       def guess_yarn_version
@@ -180,7 +180,7 @@ module Dependabot
       def pnpm_version
         return @pnpm_version if defined?(@pnpm_version)
 
-        @pnpm_version = package_manager.locked_version("pnpm") || guess_pnpm_version
+        @pnpm_version = package_manager.requested_version("pnpm") || guess_pnpm_version
       end
 
       def guess_pnpm_version
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
@@ -56,7 +56,7 @@ module Dependabot
       def ecosystem_versions
         package_managers = {}
 
-        package_managers["npm"] = Helpers.npm_version_numeric(package_lock.content) if package_lock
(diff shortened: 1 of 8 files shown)


npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -176,13 +180,19 @@ def guess_yarn_version
       def pnpm_version
         return @pnpm_version if defined?(@pnpm_version)
 
-        @pnpm_version = package_manager.locked_version("pnpm") || guess_pnpm_version
+        version = package_manager.requested_version("pnpm") || guess_pnpm_version
+
+        if version && Version.new(version.to_s) < Version.new("7")
+          raise ToolVersionNotSupported.new("PNPM", version.to_s, "7.x, 8.x")
```
abdulapopoola, line 186: should this be `7.*` to stick to our pattern?

jakecoffman (approved):

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
deivid-rodriguez: Does it work in `package.json`? If it works, that's fine with me, I just wanted to use something I know it's supported in this particular ecosystem.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
```diff
@@ -161,10 +161,14 @@ def inferred_npmrc # rubocop:disable Metrics/PerceivedComplexity
         @inferred_npmrc = nil
       end
 
+      def npm_version
+        Helpers.npm_version_numeric(package_lock.content)
```
abdulapopoola, line 165: newbie q: I'm assuming we don't guess the `npm` version like we do for `pnpm` and `yarn`?

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
abdulapopoola: sorry I meant the error message :) ; I think package.json allows semver.

deivid-rodriguez: Yes, I meant the error message two!

I'd like to include ranges that make sense for each ecosystem in the error messages, so that users can put that right in their manifests. So, for example in Python, if they want to upgrade to 3.11, they can drop `3.11.*` in their manifest and it will work, but `3.11.x` won't. By using `7.x` here I wanted to follow that approach since I'm not sure if `7.*` is a valid JS version range.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
deivid-rodriguez: We do some guessing based on the lockfile format, but we still don't respect the `packageManager` setting in `package.json`. We probably should!

npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb
abdulapopoola: Ahh; I think they follow SemVer too so I *assume* (feel free to correct me lol) 7.*.* should work.

deivid-rodriguez: Tried it and you're right, `7.*` works just the same. I'll change this, thanks for the suggestion!

npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
```diff
@@ -25,6 +25,14 @@ def self.yarn_version_numeric(yarn_lock)
         end
       end
 
+      def self.pnpm_version_numeric(pnpm_lock)
+        if pnpm_lockfile_version(pnpm_lock).to_f >= 5.4
+          8
+        else
```
yeikel, line 31: This is a bit strange, in particular the `>= 5.4` translation to 8

Could you please add some inline comment to explain this?

yeikel: Thank you for the explanation! 

In terms of what to do, I think that if we do not support an specific lockfile format, we should fail fast. We should not cause damage or receive reports for unsupported versions

deivid-rodriguez: After thinking about this a bit more, I think this is a strict improvement over what we have now.

This brings PNPM in line with what we already do for NPM and Yarn in terms of guessing the version from the lockfile format, and should transform unknown errors due to incompatibilities into user errors (even if the error message is not ideal).

If the new error message comes as confusing when a specific PNPM version is not explicitly requested in `package.json`, we can iterate and give a more specific error that points to the lockfile format instead, but I don't want to block this PR on that.

PR #8106, opened 2023-09-28
fix dependency duplication across multiple groups

jakecoffman:
fixes #7915 and possibly others

If you define multiple groups that all match the same dependencies, the dependencies will show up in each PR. 

This fix takes the dependencies that have already been handled and adds them to a Set in the DependencySnapshot that we can use to skip dependencies that have already been handled in previous groups. 

This also adds pre-processing of the existing pull requests. This is needed because imagine due to semver grouping, or a change in the manifest, a PR is created for a certain dependency and the PR is still open. That group may be further down the list, but now an earlier group wants to create a PR with that dependency. Unless we pre-process the existing PRs, then it will create a duplicate again. 

And finally when encountering an existing grouped PR, this now adds the dependencies listed in the existing PR record of the job definition, rather than the ones calculated from patterns, etc. This means if a group happens to also update a dependency outside of the group, like for instance because it has a peer-dependency on it, Dependabot won't open another PR bumping it.

PR #8150, opened 2023-10-06
fix refreshing a grouped PR causes dependency duplication

### LINKED ISSUE #8149: grouped refresh causes dependency duplication
> We're still seeing duplication [...]
> 
> For example, this grouped PR includes the `libcnb-data` and `libcnb-package` packages:
https://github.com/heroku/languages-github-actions/pull/152
> 
> ...even though they are covered by this (earlier in the Dependabot config definition) group:
https://github.com/heroku/languages-github-actions/pull/148
> 
> Our Dependabot config is here:
> https://github.com/heroku/languages-github-actions/blob/main/.github/dependabot.yml
> 
> And I reported this originally here:
> https://github.com/dependabot/dependabot-core/issues/7939
> 
> Which was then marked as a duplicate of #7915 (which this PR closed).

_Originally posted by @edmorley in https://github.com/dependabot/dependabot-core/issues/8106#issuecomment-1750275326_

jakecoffman:
fixes #8149

In #8106 we accounted for existing grouped pull requests dependencies when running a grouped update job to prevent a dependency from appearing in multiple PRs.

This PR adds that same logic to the refresh so when Dependabot rebases or recreates a PR it doesn't duplicate the dependencies.

### DIFF
--- a/updater/lib/dependabot/updater/group_update_creation.rb
+++ b/updater/lib/dependabot/updater/group_update_creation.rb
@@ -305,6 +305,16 @@ module Dependabot
 
         Dependabot::Workspace.cleanup!
       end
+
+      def pr_exists_for_dependency_group?(group)
+        job.existing_group_pull_requests&.any? { |pr| pr["dependency-group-name"] == group.name }
+      end
+
+      def dependencies_in_existing_pr_for_group(group)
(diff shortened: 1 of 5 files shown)


jurre (approved):
🎉

PR #8164, opened 2023-10-09
Teach Docker::Version about SHA and String versions

Nishnha:
Previous `Docker::Version#new` would always return a String. Then it was modified to try and return a `Gem::Version`, but Docker versions can also be a SHA or a String (like with `alpine:edge`).

This PR teaches Docker versions to handle SHA and String versions internally, and just treats them as Strings.

### DIFF
--- a/common/lib/dependabot/config/ignore_condition.rb
+++ b/common/lib/dependabot/config/ignore_condition.rb
@@ -83,6 +83,7 @@ module Dependabot
 
       def correct_version_for(dependency)
         version = dependency.version
+        version = version.to_s if version.is_a?(Gem::Version)
         return if version.nil? || version.empty?
 
         version_class = version_class_for(dependency.package_manager)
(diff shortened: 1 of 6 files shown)


jakecoffman:
I thought if it was a Version then it wasn't a SHA? Consider this updater code (and comment):

https://github.com/dependabot/dependabot-core/blob/c7734fd577960a385530c079fc5bd297946803b4/updater/lib/dependabot/job.rb#L220-L224

Nishnha:
> I thought if it was a Version then it wasn't a SHA? Consider this updater code (and comment):
> 
> https://github.com/dependabot/dependabot-core/blob/c7734fd577960a385530c079fc5bd297946803b4/updater/lib/dependabot/job.rb#L220-L224

Ah that's a good point. So in this case, SHAs are not valid versions and should return false for the `#correct?` method because we can't compare them for security updates?

I think that would also apply to version names that are Strings?

Nishnha:
Closing this PR in favor of https://github.com/dependabot/dependabot-core/pull/8165

Issue #6110, opened 2022-11-11
Review and unify shelling-out logic


deivid-rodriguez:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code improvement description

Our usage of `Open3.capture2e` which returns the full stdout+stderr of the subprocess can cause bugs where warnings from the package manager could leak into manifest files.

In #6103 I proposed a focused fix for the one instance of this bug we got reported, but this problem is likely to be happening in other places.

We should review the places where we shell out and unify the logic. I guess probably two helpers may suffice, one that returns just the output of the subprocess by using `Open3.capture3`, one that keeps using `Open3.capture2e` for the places where we're either not interested in the output, or at least won't be using it to write it into files.

This is a list of places with mostly duplicated logic for shelling out:

https://github.com/dependabot/dependabot-core/blob/aa78f70aef129043eb40412e30800e2a0345877c/python/lib/dependabot/python/file_updater/pip_compile_file_updater.rb#L140-L153
https://github.com/dependabot/dependabot-core/blob/c00613fdce7f13eb832d5cd5ae9f0ac116edb50c/python/lib/dependabot/python/file_parser/python_requirement_parser.rb#L104-L120
https://github.com/dependabot/dependabot-core/blob/aa78f70aef129043eb40412e30800e2a0345877c/python/lib/dependabot/python/file_updater/pipfile_file_updater.rb#L246-L264
https://github.com/dependabot/dependabot-core/blob/c00613fdce7f13eb832d5cd5ae9f0ac116edb50c/elm/lib/dependabot/elm/update_checker/elm_19_version_resolver.rb#L116-L133
https://github.com/dependabot/dependabot-core/blob/aa78f70aef129043eb40412e30800e2a0345877c/python/lib/dependabot/python/file_updater/poetry_file_updater.rb#L196-L214
https://github.com/dependabot/dependabot-core/blob/32f17db17a332248a99b155b8af763579a3faa79/common/lib/dependabot/shared_helpers.rb#L78-L134
https://github.com/dependabot/dependabot-core/blob/aa78f70aef129043eb40412e30800e2a0345877c/python/lib/dependabot/python/file_updater/pipfile_file_updater.rb#L246-L264
https://github.com/dependabot/dependabot-core/blob/8a70b5e719eb465e134d0167c5bf6513bd9ee77a/cargo/lib/dependabot/cargo/update_checker/version_resolver.rb#L135-L153
[...]

PR #8167, opened 2023-10-10
Refactor shelling out in Python

deivid-rodriguez:
There's a lot of duplicated code in the Python ecosystem that deals with shelling out to package managers. This PR refactors it to always use the same helper.

This fixes most of #6110.


python/lib/dependabot/python/file_updater/poetry_file_updater.rb
```diff
@@ -204,7 +203,7 @@ def run_poetry_update_command
         end
 
         def run_poetry_command(command, fingerprint: nil)
-          Helpers.run_poetry_command(command, fingerprint: fingerprint)
+          SharedHelpers.run_shell_command(command, fingerprint: fingerprint)
```
deivid-rodriguez, line 206: Same as the other comment, but reversed. This used to raise an error including only stderr, now we'll raise an error include stdout and stderr concatenated.

Seems fine to I guess!

python/lib/dependabot/python/update_checker/poetry_version_resolver.rb
```diff
@@ -309,7 +308,7 @@ def lockfile
         end
 
         def run_poetry_command(command, fingerprint: nil)
-          Helpers.run_poetry_command(command, fingerprint: fingerprint)
+          SharedHelpers.run_shell_command(command, fingerprint: fingerprint)
```
deivid-rodriguez, line 311: Same as the other comment in the other `run_poetry_command` method.

yeikel (commented):
It's always great to see these cleanups and more so when the tests are happy 🥇

Nishnha (approved):
This LGTM. I would maybe get another approval from the team though since I haven't used `time_taken` in my debugging, but maybe someone else has found it useful and doesn't want it removed?

deivid-rodriguez:
@Nishnha Not that this is not removing `time_taken`, the same code is present in the common helpers which is now used consistently:

https://github.com/dependabot/dependabot-core/blob/1b3368e3d7e4eecb5ccb9475151ddd09018f5221/common/lib/dependabot/shared_helpers.rb#L317-L348

Only changes in behavior should be the ones I noticed as inline comments in the PR.

PR #8168, opened 2023-10-10
Fix detection of compilation errors

deivid-rodriguez:
This is used to retry without build isolation, but the message has changed in recent pip versions and was no longer matched. Now, at least in some cases, the error reads like this:

```
pip._internal.exceptions.InstallationSubprocessError: Getting requirements to build wheel exited with 1
```


abdulapopoola (approved):

deivid-rodriguez:
After some more thought, and seeing a few other errors, I think it's probably best to replace the previous error message with the new one instead of relaxing it.

I'm seeing other `pip._internal.exceptions.InstallationSubprocessError` errors with different root causes unrelated to `--build-isolation`, and rerunning the command with `--no-build-isolation` probably won't work either and will only hide the real culprit even more.

deivid-rodriguez:
I updated this PR to use the more conservative approach 👍.

deivid-rodriguez:
This PR itself was not fixing any known issue, it needs extra fixes on top of it. So it may encrease runtime for no good reason, even if it's a bug fix. So I decided to introduce it together with other bug fixes over at #8189.

PR #8170, opened 2023-10-10
Ignore file dependencies when parsing requirement files

deivid-rodriguez:
These can't be updated.

If we try to update them, we end up running into errors later. Sample repro:

```
$ bin/dry-run.rb pip "bowtie-json-schema/bowtie" --dir="/." --updater-options=grouped_updates_experimental_rules,record_ecosystem_versions,record_update_job_unknown_error,unignore_commands --commit=a8fd4b1828d5a71f8bda669e8dc84daf19caa1c9 --cache=files --dep bowtie-json-schema
```

### DIFF
--- a/python/helpers/lib/parser.py
+++ b/python/helpers/lib/parser.py
@@ -99,6 +99,10 @@ def parse_requirements(directory):
                 if install_req.req is None:
                     continue
 
+                # Ignore file: requirements
+                if install_req.link is not None and install_req.link.is_file:
+                    continue
+
                 pattern = r"-[cr] (.*) \(line \d+\)"
                 abs_path = re.search(pattern, install_req.comes_from).group(1)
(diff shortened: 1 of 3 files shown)


Nishnha (approved):
I think ignoring path dependencies is the right move for now.
One suggestion I have is maybe we should log to the user that we are skipping the dependency because it's a path dependency?

deivid-rodriguez:
Thanks for having a look @Nishnha!

We skip many other cases in Python and we don't log anything (for example, dependencies without specific requirements, or with weird requirement ranges). I believe as long the user does not expect any Dependabot updates for the Dependency, it's fine to not long anything.

It'd also be a bit non standard to log from Python native helper.

So I'll ship this as is, but let me know if you feel strongly about logging this and we can iterate on this 👍.

PR #8171, opened 2023-10-10
remove unused code

jakecoffman:
This code is never invoked in the security update path, so it's a bit misleading. Removing it!


abdulapopoola:
[newbie question] Do the logs show if these paths ever got executed and stopped working?

jakecoffman:
@abdulapopoola No these logs are deleted after 14 days so we can't look back very far. 

If you go back to the PR that added these you can see they were never invoked: https://github.com/dependabot/dependabot-core/pull/6961

PR #8172, opened 2023-10-10
Raise user error when not finding path dependencies in Python

deivid-rodriguez:
[This repo](https://github.com/SylinaZhang/superset-3/) has an incorrect Dependabot/Python configuration, because:

* It configures Dependabot at `/requirements`.
* It has a file `requirements/base.in` including `-e file:.`.

So Dependabot and pip-compile are correctly expecting that a Python project (either a `setup.py` or a `pyproject.toml` file) is present a `/requirements`.

We currently get an unknown error registered and buried in the logs, raised by a `pip-compile` subprocess:

> pip._internal.exceptions.InstallationError: file:///. (from -r base.in (line 19)) does not appear to be a Python project: neither 'setup.py' nor 'pyproject.toml' found.

With these changes, we should be able to detect the problem and show a better error to the user, saying that we expected to find a `pyproject.toml` file at `/requirements`.


honeyankit (approved):

Nishnha:
I think the https://github.com/repos/SylinaZhang/superset-3 repo may be private. 
If the base64 content is important to testing this feature we may want to just obscure the repo name, otherwise we may want to remove the base64 content entirely.

Nishnha:
> @Nishnha The repo is public, I linked to it in the PR description. The URL you used has an errant `/repos` segment, that's why it's 404.
> 
> I used the sample public repo to create the tests by recording the responses from the [contents API](https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28).

Oh cool yeah I did have the wrong URL.

python/lib/dependabot/python/file_fetcher.rb
```diff
@@ -337,10 +337,7 @@ def fetch_path_setup_file(path, allow_pyproject: false)
             ).tap { |f| f.support_file = true }
           rescue Dependabot::DependencyFileNotFound
             # For projects with pyproject.toml attempt to fetch a pyproject.toml
-            # at the given path instead of a setup.py. We do not require a
-            # setup.py to be present, so if none can be found, simply return
-            return [] unless allow_pyproject
-
+            # at the given path instead of a setup.py.
```
Nishnha, line 340: Won't this cause python projects that don't use `pyproject.toml` to raise an error when they call `#fetch_path_setup_file`?

Do all python ecosystems that we support now use `pyproject.toml`?

deivid-rodriguez: If there's a "path dependency" (what Python calls "editable" dependency), then there needs to be either a setup.py or a pyproject.toml file at that path. If none of them are found, at pip-compile does not consider it a valid Python project and fails I believe.

In this case, we're in the rescue block from fetching a setup.py file, so this means none of them are found.

Nishnha: That makes sense

> If there's a "path dependency" (what Python calls "editable" dependency), then there needs to be either a setup.py or a pyproject.toml file at that path

Does your code support when a pyproject.toml is found but a setup.py is not?

I see your test at https://github.com/dependabot/dependabot-core/pull/8172/files#diff-ba69c800c882181677c61aa2162b09edfd1bb30422d39e5f11afe35dcf2df821R833-R846 only checks whether a `DependencyFileNotFound` error is raised when neither exist

Nishnha: Ah I misread the path.gsub 

Thanks for explaining this all to me!

Nishnha (approved):

deivid-rodriguez:
I substantially changed this PR since I wasn't fully happy with the error message. Now we should display something like:

> The following path based dependencies could not be retrieved:
>
> * `-e file:.` at /requirements/base.in (no setup.py or pyproject.toml found in /requirements)

The idea is that the user knows where she needs to go to fix the configuration.

Happy to get feedback on the copy.

Also, while at it, I included several renaming since in a world that's shifting towards `pyproject.toml`, the previous naming using "setup" everywhere was confusing.

deivid-rodriguez:
Ok, I hope this is it. Previous format did not render great in the error banner over dependabot logs, so I went with the following (which respects current formatting, so should render fine):

> The following path based dependencies could not be retrieved: `"-e file:." at /requirements/base.in`

Ready, once and for all!

Nishnha (approved):
LGTM. Helps to view the diff with hide whitespace enabled.

PR #8179, opened 2023-10-11
Contribute changes to NuGet updater from Azure team

### LINKED ISSUE #6710: Leaf level dependencies should be updated first

### Package ecosystem

NuGet

### Package manager version

_No response_

### Language version

_No response_

https://github.com/batkaevruslan/dependabot_issues/blob/main/DependabotRepro/A/A.csproj
https://github.com/batkaevruslan/dependabot_issues/blob/main/DependabotRepro/B/B.csproj

### dependabot.yml content

https://github.com/batkaevruslan/dependabot_issues/blob/main/.github/dependabot.yml

### Updated dependency

AspNetCore.HealthChecks.Rabbitmq from 5.0.2 to 6.0.2

I expect Dependabot to update dependency from the leaf level of dependency graph first: Microsoft.Extensions.Diagnostics.HealthChecks
image

Unfortunately, the root dependency AspNetCore.HealthChecks.Rabbitmq is updated first. That gives me a non-buildable solution:
image

In other words, I expect the Dependabot log for my repo to be:

> updater | |                       Changes to Dependabot Pull Requests                        |
> updater | +---------+------------------------------------------------------------------------+
> updater | | created | Microsoft.Extensions.Diagnostics.HealthChecks ( from 5.0.17 to 7.0.3 ) |
> updater | | created | AspNetCore.HealthChecks.Rabbitmq ( from 5.0.2 to 6.0.2 )               |
> updater | +---------+------------------------------------------------------------------------+

instead of the current:

> updater | |                       Changes to Dependabot Pull Requests                        |
> updater | +---------+------------------------------------------------------------------------+
> updater | | created | AspNetCore.HealthChecks.Rabbitmq ( from 5.0.2 to 6.0.2 )               |
> updater | | created | Microsoft.Extensions.Diagnostics.HealthChecks ( from 5.0.17 to 7.0.3 ) |
> updater | +---------+------------------------------------------------------------------------+

_No response_

https://github.com/batkaevruslan/dependabot_issues/pull/1

Main branch can be built: https://github.com/batkaevruslan/dependabot_issues

Dependabot's branch that fails to build: https://github.com/batkaevruslan/dependabot_issues/tree/dependabot/nuget/DependabotRepro/main/AspNetCore.HealthChecks.Rabbitmq-6.0.2

My branch with suggested update order (when Microsoft.Extensions.Diagnostics.HealthChecks is updated first): https://github.com/batkaevruslan/dependabot_issues/tree/chore/update-Microsoft.Extensions.Diagnostics.HealthChecks

### LINKED ISSUE #8183: NuGet: package update is offered that is incompatible with the target framework

### Package ecosystem

NuGet

### Package manager version

NuGet Command Line 6.7.0.127

### Language version

C# 10, .NET 6

Applicable snippet from `Directory.Build.props`, which defines the version property:
```xml
<Project>
  <PropertyGroup>
    <TargetFrameworkName>net6.0</TargetFrameworkName>
    <AspNetCoreVersion>6.0.*</AspNetCoreVersion>
  </PropertyGroup>
</Project>
```

Usage in .csproj file:
```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>$(TargetFrameworkName)</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="$(AspNetCoreVersion)" />
  </ItemGroup>
</Project>
```

### dependabot.yml content

```yml
version: 2
updates:
  directory: "/"
  schedule:
    interval: "weekly"
  pull-request-branch-name:
    separator: "-"
- package-ecosystem: nuget
  directory: "/"
  schedule:
    interval: daily
  pull-request-branch-name:
    separator: "-"
  open-pull-requests-limit: 25
  ignore:
  # Block updates to all exposed dependencies of the NuGet packages we produce, as updating them would be a breaking change.
  - dependency-name: 'Ben.Demystifier'
  - dependency-name: 'Humanizer*'
  - dependency-name: 'Microsoft.CodeAnalysis*'
  - dependency-name: 'Microsoft.EntityFrameworkCore*'
  # Block major updates of packages that require a matching .NET version.
  - dependency-name: 'Microsoft.AspNetCore*'
    update-types: ["version-update:semver-major"]
```

### Updated dependency

Updates the requirements on [Microsoft.AspNetCore.Mvc.Testing](https://github.com/dotnet/aspnetcore) to permit the latest version.

The PR updates `AspNetCoreVersion` in `Directory.Build.props` from 6.0.* to 7.0.*.

First of all, the update should not be offered because we have an ignore rule in `dependabot.yml` (possibly related to #8182):
```yml
- package-ecosystem: nuget
  directory: "/"
  ignore:
  # Block major updates of packages that require a matching .NET version.
  - dependency-name: 'Microsoft.AspNetCore*'
    update-types: ["version-update:semver-major"]
```

Assuming the ignore rule does not work, the package update should still not be offered. Because v6.0.* is compatible solely with .NET 6 (see [here](https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc.Testing/6.0.23#supportedframeworks-body-tab)), whereas v7.* is solely compatible with .NET 7 (see [here](https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc.Testing/7.0.12#supportedframeworks-body-tab)). Because our project does not target .NET 7, the update can never work.

Error from continuous integration build:
```
error NU1202: Package Microsoft.AspNetCore.Mvc.Testing 7.0.12 is not compatible with net6.0 (.NETCoreApp,Version=v6.0). Package Microsoft.AspNetCore.Mvc.Testing 7.0.12 supports: net7.0 (.NETCoreApp,Version=v7.0)
```

There are many more packages whose versions are tightly bound to the .NET framework version. They can only be updated when updating the target framework as well. For example:
- Microsoft.AspNetCore.Authentication.Certificate
- Microsoft.AspNetCore.Authentication.JwtBearer
- Microsoft.AspNetCore.Authentication.OpenIdConnect
- Microsoft.AspNetCore.Cryptography.KeyDerivation
- Microsoft.AspNetCore.DataProtection
- Microsoft.AspNetCore.DataProtection.StackExchangeRedis
- Microsoft.AspNetCore.TestHost
- Microsoft.Extensions.Caching.StackExchangeRedis
- Microsoft.Extensions.Diagnostics.HealthChecks

Unless dependabot also updates the target framework (which, fortunately, it does not), there is no point in offering PRs for major version changes.

_No response_

https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/1355

_No response_

brettfo:
Contribute NuGet changes back to the community
=====

Internally at Microsoft we've been making some updates to the NuGet dependabot updater and are now in a position to contribute those changes back to the community. We're excited to be able to do this and hope that it will help improve the experience for everyone.

This work was [announced](https://www.youtube.com/watch?v=zCKZMilzWUk&t=1202s) by Bryan Sullivan at Microsoft Build 2023 in May of this year.

# Summary of improvements

- Rewrite core updater logic in C#.  External contributors that want to make further improvements will likely be more comfortable working in a .NET environment with Visual Studio.  Some Ruby code still exists to take advantage of the functionality offered in the base updaters.
- Use code from the NuGet project (via a submodule) to handle updates for `packages.config` scenarios.  This ensures that we're using as close to the same logic as possible as the NuGet CLI to update the references.
- Use MSBuild to compute the complete list of `.props` and `.targets` files that are imported, regardless of any complicated path computations.  This ensures that for each project file being analyzed, we can better determine the exact file that declares the package version.  This can be particularly helpful in large monorepos.
- Propertly handle updates to implicit dependencies.  This is particularly useful for SDK-style projects that may not explicitly list all of their dependencies.
- Propertly handle updates to peer dependencies.  This will help reduce build failues in generated pull requests by ensuring that all applicable packages are updated together.
- Update binding redirect assembly versions in `app.config`/`web.config` files.  Without these changes the user's app will likely fail to load the assemblies at runtime, making these errors hard to detect.

# Notes

Part of the NuGet updater work (i.e., supporting `packages.config`) required code from the NuGet/NuGet.Client repo.  The specific code required only exists as a Windows console app targeting net472 (see nuget.org/downloads) and the NuGet team does not want contributions back to allow this to build on Linux/net7.0 (I asked) so to avoid copying the entire repo, a submodule was added tied to the latest public release.  To help make local development easier, the NuGet build script checks for a sentinel file in that submodule and if it's not present, will report the command the developer can use to fix the issue.

# Related PRs

There were a number of other internal improvements that we made that have been split into separate PRs:

- #8191
- #8192
- #8193
- #8194

# Caveats

There will be a few more changes from our side that I'll add to this PR as appropriate, but wanted to get the ball rolling on review and feedback.

I'll separately be working on cleaning up the git history to remove duplicate commits.  When that work is done I'll force-push to this branch to help ensure a clean merge when it's all complete.

As of internal commit `15e3b8d09374468e40d8d77bafc1c01178db444c`

# Closed issues

* Closes #6710.
* Closes #8183.


.devcontainer/post-create.sh
```diff
@@ -4,4 +4,10 @@ tar xzvf ./*.tar.gz >/dev/null 2>&1
 sudo mv dependabot /usr/local/bin
 rm ./*.tar.gz
 
+# The image comes loaded with 8.0 SDK, but we need a 7.0 runtime for running tests
```
yeikel, line 7: Can you add a clarifying comment explaining why 7.0 is needed for the tests? 

I assume that it is because you target more than one version but it would still be helpful to be explicit

Dockerfile.updater-core
```diff
@@ -88,7 +88,7 @@ WORKDIR $DEPENDABOT_HOME/dependabot-updater
 
 # Install Ruby from official Docker image
 # When bumping Ruby minor, need to also add the previous version to `bundler/helpers/v{1,2}/monkey_patches/definition_ruby_version_patch.rb`
-COPY --from=ruby:3.1.4 --chown=dependabot:dependabot /usr/local /usr/local
+COPY --from=ruby:3.1.4-bullseye --chown=dependabot:dependabot /usr/local /usr/local
```
yeikel, line 91: Why is this change needed?

npm_and_yarn/Dockerfile
```diff
@@ -1,15 +1,26 @@
...
+
+# Check for updates at https://github.com/yarnpkg/berry/releases
+ARG YARN_VERSION=3.5.0
+
+# See https://github.com/nodesource/distributions#installation-instructions
+ARG NODEJS_VERSION=16.x
+
+# Check for updates at https://github.com/npm/cli/releases
+# This version should be compatible with the Node.js version declared above. See https://nodejs.org/en/download/releases as well
+ARG NPM_VERSION=8.19.4
```
yeikel, line 15: It seems that this branch needs to be re based with the latest main because this is quite outdated

nuget/Dockerfile
```diff
@@ -1,7 +1,44 @@
...
 
-USER dependabot
+USER root
+
+ENV DEPENDABOT_NATIVE_HELPERS_PATH="/opt"
+
+# Install .NET SDK dependencies
+RUN apt-get update \
+  && apt-get install -y --no-install-recommends \
+  libicu-dev \
```
yeikel, line 10: We should considering pinning this version to make the image more reproducible

nuget/helpers/lib/NuGetUpdater/NuGetProjects/NuGet.CommandLine/AssemblyMetadataExtractor.cs
```diff
@@ -0,0 +1,203 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
yeikel, line 1: Does it make sense to License dependabot-core with two different licenses?

I am no lciense expert, but if `main` uses ``

.devcontainer/post-create.sh
brettfo: done, I'll push up the new comment shortly

Dockerfile.updater-core
brettfo: I originally created this merge from the point that we took an internal fork.  The latest iteration of my commits doesn't have this change.

npm_and_yarn/Dockerfile
brettfo: good catch, I've pushed up a better rebase; this file is no longer touched

nuget/Dockerfile
brettfo: done, will push changes shortly

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "nuget/helpers/lib/NuGet.Client"]
+	path = nuget/helpers/lib/NuGet.Client
+	url = https://github.com/NuGet/NuGet.Client
```
deivid-rodriguez, line 3: Can you elaborate on the need on using `NuGet.Client` through a submodule? Is it not distributed in a more standard way?

Dockerfile.updater-core
```diff
@@ -115,7 +115,6 @@ RUN gem install bundler -v $BUNDLER_V2_VERSION --no-document && \
  rm -rf /var/lib/gems/*/cache/* && \
  bundle config set --global build.psych --with-libyaml-source-dir=$DEPENDABOT_HOME/src/libyaml/yaml-$LIBYAML_VERSION && \
  bundle config set --local path 'vendor' && \
- bundle config set --local frozen 'true' && \
```
deivid-rodriguez, line 118: This is the kind of change that, if it's necessary, it's best to be discussed in isolation as a separate PR, since it has nothing to do with NuGet.

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -100,7 +100,7 @@ def clone_repo_contents
           raise Dependabot::OutOfDisk
         end
 
-        raise Dependabot::RepoNotFound, source
+        raise Dependabot::RepoNotFound.new(source, e.message)
```
deivid-rodriguez, line 103: Same thing about extracting changes unrelated to NuGet to separate PRs.

common/lib/dependabot/pull_request_creator/message_builder/metadata_presenter.rb
```diff
@@ -154,7 +154,7 @@ def build_details_tag(summary:, body:)
             msg += body
             msg + "</details>\n"
           else
-            "\n##{summary}\n\n#{body}"
+            "\n# #{summary}\n\n#{body}"
```
deivid-rodriguez, line 157: Can we get a separate PR with this change/fix?

docker/lib/dependabot/docker/file_parser.rb
```diff
@@ -24,12 +24,15 @@ class FileParser < Dependabot::FileParsers::Base
...
-      TAG = /:(?<tag>[\w][\w.-]{0,127})/
+      TAG_NO_PREFIX = /(?<tag>[\w][\w.-]{0,127})/
+      TAG = /:#{TAG_NO_PREFIX}/
       DIGEST = /(?<digest>[0-9a-f]{64})/
       NAME = /\s+AS\s+(?<name>[\w-]+)/
       FROM_LINE =
         %r{^#{FROM}\s+(#{PLATFORM}\s+)?(#{REGISTRY}/)?
           #{IMAGE}#{TAG}?(?:@sha256:#{DIGEST})?#{NAME}?}x
+      TAG_WITH_DIGEST =
+        %r{^#{TAG_NO_PREFIX}(?:@sha256:#{DIGEST})?}x
```
deivid-rodriguez, line 35: I was not expecting changes to the docker ecosystem in this PR. Can you elaborate? If it's just an unrelated change/fix, a separate PR would be greatly appreciated!

updater/lib/dependabot/updater/operations/update_all_versions.rb
```diff
@@ -99,6 +100,12 @@ def check_and_create_pull_request(dependency)
             requirements_to_unlock: requirements_to_unlock
           )
 
+          if updated_deps.empty?
+            return Dependabot.logger.info(
+              "No update possible for #{dependency.name} #{dependency.version}"
+            )
+          end
+
```
deivid-rodriguez, line 108: Seems like another tweak that's probably a good change for us, but probably unrelated to NuGet, so best to extract it to a separate PR.

deivid-rodriguez (commented):
I made a quick pass commenting on some stuff that seems general and not necessarily related to NuGet.

It'd be nice to discuss & merge those changes independently through separate PRs.

Dockerfile.updater-core
brettfo: Sure thing, I'll split this out along with the other non-NuGet changes into separate PRs.

.gitmodules
brettfo: I'll add a detailed comment to the PR description shortly, but the short version is that the functionality required isn't available as a NuGet package or otherwise that can be consumed by a Linux container (it's currently only available as a net472 Windows console app and the NuGet team doesn't want to support any other configuration), so to avoid having a complete copy of the NuGet code here or elsewhere, it was much simpler to use a submodule and only re-create the project files with the relevant changes.

Dockerfile.updater-core
brettfo: I've pulled out this commit, but now every Docker image fails to build because the Gemspec.lock files aren't present.  I'm not very familiar with the Ruby ecosystem, but how does this feature work in a regular PR?  What are the steps necessary to avoid the build errors I'm seeing?

deivid-rodriguez: Not sure if I understood. We use `frozen` since it's a good practice to ensure that `Gemfile` and `Gemfile.lock` are in sync, and `frozen` enforces that. Why do you need to remove this in the first place? Also, can I see the errors you are getting?

jakecoffman:
This is adding 520,000 lines of VCR Cassettes which seems a bit excessive. The next nearest ecosystem is 40,000 lines. Mind taking a look and see if more can be mocked out rather than recorded?

carogalvin:
[In our current documentation, we claim to support NuGet <= 4.8.](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem) Will that need to be updated?

brettfo:
> [In our current documentation, we claim to support NuGet <= 4.8.](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem) Will that need to be updated?

Yes, once this is merged that can be updated to `<= 6.7.0`.  We're using the NuGet client code that corresponds with their 6.7.0 release.

carogalvin:
@brettfo @deivid-rodriguez I'm working with our documentation team on this; we don't really have thorough docs on specific behaviour of different ecosystems. Are there any particular still-existing limitations in NuGet we should call out?

brettfo:
> @brettfo @deivid-rodriguez I'm working with our documentation team on this; we don't really have thorough docs on specific behaviour of different ecosystems. Are there any particular still-existing limitations in NuGet we should call out?

@carogalvin I think the only feature not supported is the SDK version inline in the XML attribute like so:

``` xml
<Project Sdk="SomeSdk/1.2.3">
  <!--               ^^^^^^ this isn't supported yet -->
  ...
</Project>
```

deivid-rodriguez:
Working on getting specs to pass with less cassettes at https://github.com/dependabot/dependabot-core/tree/nuget-pass-specs.

brettfo:
Thank you for doing the heavy lifting for this, it would have taken me 2 months.  I reviewed the changes in the other PR and everything looks good from my end.

I think the only thing we're waiting on is some internal email discussions about updating the README, etc.  As soon as those conversations resolve themselves I'll add to this.

carogalvin:
@brettfo you can remove the line "All contributions to Dependabot Core implicitly transfer the IP of that contribution to GitHub, Inc. where it will be licensed the same way as above." from the README and that will be sufficient to merge this

brettfo:
@carogalvin Thank you, I've pushed a commit with that line removed.
@deivid-rodriguez I think this is ready to merge.

deivid-rodriguez:
I still want to add a few cleanups and add a smoke test at ttps://github.com/dependabot/smoke-tests/. But not planning to add any changes that could affect production anyways. Will merge this on Friday 👍.

jakecoffman:
I built this image and ran with the CLI: `dependabot update nuget microsoft/clrmd`

Getting an error for all dependencies:

```
updater | 2023/11/16 18:14:38 ERROR Error processing Microsoft.Build.Utilities.Core (NoMethodError)
updater | 2023/11/16 18:14:38 ERROR undefined method `directory' for nil:NilClass
updater | 
updater |         return "" if files.first.directory == "/"
updater |                                 ^^^^^^^^^^
updater | 2023/11/16 18:14:38 ERROR /home/dependabot/common/lib/dependabot/pull_request_creator/message_builder.rb:176:in `pr_name_directory'
```

deivid-rodriguez:
Let me look into it!

deivid-rodriguez:
I looked into this.

I _believe_ this should be surfaced as a user error.

If I add this debugging print

```diff
diff --git a/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs b/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
index 4db75a078..aef777046 100644
--- a/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
+++ b/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
@@ -255,6 +255,10 @@ internal static partial class MSBuildHelper
 
             var (exitCode, stdout, stderr) = await ProcessEx.RunAsync("dotnet", $"build \"{projectPath}\" /t:_ReportDependencies");
             var lines = stdout.Split('\n').Select(line => line.Trim());
+            Console.WriteLine(exitCode);
+            Console.WriteLine(stdout);
+            Console.WriteLine(stderr);
+
             var pattern = PackagePattern();
             var allDependencies = lines
                 .Select(line => pattern.Match(line))
@@ -341,4 +345,4 @@ internal static partial class MSBuildHelper
 
     [GeneratedRegex("^\\s*NuGetData::Package=(?<PackageName>[^,]+), Version=(?<PackageVersion>.+)$")]
     private static partial Regex PackagePattern();
-}
+}
\ No newline at end of file
```

We can see that a dotnet command used for checking project dependencies is failing:

```
1
MSBuild version 17.6.3+07e294721 for .NET
  Determining projects to restore...
/tmp/package-dependency-resolution_10F0ar/Project.csproj : error NU1102: Unable to find package Microsoft.CodeAnalysis.CSharp.CodeStyle with version (= 4.6.0-1.23073.4)
/tmp/package-dependency-resolution_10F0ar/Project.csproj : error NU1102:   - Found 1051 version(s) in dotnet-tools [ Nearest version: 4.6.0-1.23109.10 ]
/tmp/package-dependency-resolution_10F0ar/Project.csproj : error NU1102:   - Found 52 version(s) in dotnet-public [ Nearest version: 4.6.0-1.final ]
/tmp/package-dependency-resolution_10F0ar/Project.csproj : error NU1102:   - Found 0 version(s) in dotnet8-transport
/tmp/package-dependency-resolution_10F0ar/Project.csproj : error NU1102:   - Found 0 version(s) in dotnet8
/tmp/package-dependency-resolution_10F0ar/Project.csproj : error NU1102:   - Found 0 version(s) in dotnet-eng
  Failed to restore /tmp/package-dependency-resolution_10F0ar/Project.csproj (in 3.18 sec).

Build FAILED.

/tmp/package-dependency-resolution_10F0ar/Project.csproj : error NU1102: Unable to find package Microsoft.CodeAnalysis.CSharp.CodeStyle with version (= 4.6.0-1.23073.4)
/tmp/package-dependency-resolution_10F0ar/Project.csproj : error NU1102:   - Found 1051 version(s) in dotnet-tools [ Nearest version: 4.6.0-1.23109.10 ]
/tmp/package-dependency-resolution_10F0ar/Project.csproj : error NU1102:   - Found 52 version(s) in dotnet-public [ Nearest version: 4.6.0-1.final ]
/tmp/package-dependency-resolution_10F0ar/Project.csproj : error NU1102:   - Found 0 version(s) in dotnet8-transport
/tmp/package-dependency-resolution_10F0ar/Project.csproj : error NU1102:   - Found 0 version(s) in dotnet8
/tmp/package-dependency-resolution_10F0ar/Project.csproj : error NU1102:   - Found 0 version(s) in dotnet-eng
    0 Warning(s)
    1 Error(s)

Time Elapsed 00:00:04.39
```

My interpretation of this error is that the project is locked to a version of a dependency that has been yanked and it's no longer available.

If this is the case, we'd need to add proper error handling to the native tool and ruby runner, so that this is surfaced as a user error. Probably raise if the status code here is not zero, and then catch some error code at the `UpdaterChecker`, and re-raise an appropriate user error (`DependencyFileNotResolvable` or whatever).

Alternatively, we may not be parsing the version of this dependency correctly. If that's the case, then this is a real bug that needs fixing.

Finally, this issue suggests were doing a lot of duplicate work because:

* We're first parsing dependencies manually from Ruby in the `FileParser` step.
* Then we're again parsing dependencies using native tools.

If this is due to a yanked version, and the native tool is able to detect this, it sounds like the native tool logic should be used from the `FileParser` and the Ruby logic removed. That way we have less code to maintain and can raise a proper error earlier.

None of this seems blocking to me, but I'm not sure how common this issue is, and I have not checked what our current behavior is. We probably don't currently error on this kind of thing since the current NuGet updater has no resolvability capabilities whatsoever, just bumps things manually.

deivid-rodriguez:
Small nit: I see you're committing your own `.editorconfig` file setting `insert_final_newline = false`. This is inconsistent with the rest of the project. Would you be opposed to setting it to true, and adding a final new line to all files?

deivid-rodriguez:
Also I pushed a few changes to make nuget more consistent with other ecosystems such as:

* Install dotnet to `/usr/local`, that's where we usually install "languages".
* Make native CLI tool be installed under `/opt/nuget`, previously it was installed directly to `/opt`.
* Allow building and installing native helpers by running `nuget/helpers/build` from the dev image.
* Also fixed some warnings about having a world writable tmp. For that I set the `NUGET_SCRATCH` variable to somewhere we can write to.


Let me know if these changes look ok!

jakecoffman (approved):

brettfo:
@deivid-rodriguez Thank you for looking into these issues, I was travelling all day yesterday.

As for the missing version, it's likely not a yanked version, but that specific version isn't available on the feed listed in `NuGet.Config`, and in this case you're correct in that it should be marked as a failure, because the version we're supposed to update to isn't available, so there's nothing that can be done.

Regarding `.editorconfig` feel free to set it to whatever best aligns with the rest of the code, we're not too picky.

As for the various path changes, etc., that's also just fine.  When I originally wrote this I basically rolled the dice on where to install stuff; I have to strong preference where it lives.

deivid-rodriguez:
> As for the missing version, it's likely not a yanked version, but that specific version isn't available on the feed listed in NuGet.Config, and in this case you're correct in that it should be marked as a failure, because the version we're supposed to update to isn't available, so there's nothing that can be done.

Yeah, I assumed it was yanked but it could also be a typo or whatever. Note though that this is not the version we are updating to, but that version being currently used by the project. And also note that this will prevent all updates to a project, not just updates to that specific "broken" dependency.

brettfo:
> > As for the missing version, it's likely not a yanked version, but that specific version isn't available on the feed listed in NuGet.Config, and in this case you're correct in that it should be marked as a failure, because the version we're supposed to update to isn't available, so there's nothing that can be done.
> 
> Yeah, I assumed it was yanked but it could also be a typo or whatever. Note though that this is not the version we are updating to, but that version currently used/locked in the project. And also note that this will prevent all updates to a project, not just updates to that specific "broken" dependency.

Ahh, I understand.  That's certainly a scenario we didn't consider, where the old package version couldn't be found.  I don't think that should prevent the upgrade, only the new version not being found should be a problem.  I'll file an issue to fix the old-version not found, because that ideally _should_ work.

deivid-rodriguez:
To summarize my understanding of what's happening, in case it helps you figuring this out:

* After figuring which dependencies and which versions you want to update to, you'll try to update each dependency by running your native tool here:
https://github.com/dependabot/dependabot-core/blob/d94992ab898b38356d49e08ddd3b2e055d5c698e/nuget/lib/dependabot/nuget/file_updater.rb#L65

* First thing your native tool does is finding current project dependencies: https://github.com/dependabot/dependabot-core/blob/d94992ab898b38356d49e08ddd3b2e055d5c698e/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Updater/SdkPackageUpdater.cs#L69. As I pointed out before, this seems like duplicated work from the Ruby FileParser. The native tool seems to do this in order to double check whether the dependency we're trying to update to is really a project dependency, and whether its version is already higher than the target version.

* In order to find current project dependencies, the native tool writes a dummy project with manually parsed top level dependencies. This dummy project defines a target to `dotnet build` to report project dependencies in a format that you later parse: https://github.com/dependabot/dependabot-core/blob/d94992ab898b38356d49e08ddd3b2e055d5c698e/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs#L218-L263

* Because of the error with this dependency, the `dotnet build` command fails and the output is not in the format you expect, so dependencies end up being parsed as empty.

* Because of that, the dependency is not found in the (empty) list of project dependencies, so the update is skipped: https://github.com/dependabot/dependabot-core/blob/d94992ab898b38356d49e08ddd3b2e055d5c698e/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Updater/SdkPackageUpdater.cs#L51-L56.

In general, if project dependencies are initialky broken, it's fine to raise a user error. I think this is what we normally do in other ecosystems, because package managers normally fail to run in this situation. And that makes sense to me.

But if the broken dependency can be ignored and the update provided anyways, I guess that's also fine. The user won't know things are broken, but at least they'll have broken, but more up to date, dependencies 😆.

I wonder how checking resolvability did not run into this problem and you only find out at the very end, when trying to actually update the files. It sounds like there's definitely a bug somewhere.

deivid-rodriguez:
I cleanup this commit history a bit more so that obsessive-me is happy and the smoke test that I added seems to be passing, so I'm going to deploy this!

nuget/helpers/lib/NuGetUpdater/NuGetProjects/NuGet.CommandLine/AssemblyMetadataExtractor.cs
deivid-rodriguez: This should've been addressed by 5078f14b92b177eaad0017df6565c4ef8323f9d6.

brettfo:
@deivid-rodriguez  The analysis in [this comment](https://github.com/dependabot/dependabot-core/pull/8179#issuecomment-1816887612) is mostly correct, just with one caveat.

The function `GetAllPackageDependenciesAsync` doesn't get the dependencies for a **project**, but given a single **package** (and version and target framework) it gets the full list of transitive dependencies all the way to the bottom.  This is a surprisingly involved process and rather than pull down a package, look at its dependencies, then repeat until no more are left (all the while resolving conflicting package versions), we cheat and make that fake project and tell MSBuild to do the hard part, then we intercept it and report it back out.

This function came about from the following scenario:

- SDK-style projects only list top level dependencies, not any of the transitive ones (because MSBuild/NuGet figures them out at build time.)
- We try to fix/update a package, but it's not a top level dependency, so it would get skipped.
- To fix this we resolve all packages all the way down to see if the one we're supposed to update appears in the list somewhere.
- If it appears in this list (and not at the top level), then we need to add a new top-level dependency with this version fix.

Eventually we'd like to remove most of the Ruby code and simply call into our CLI, but it gets kind of nasty.  Consider the case where we have `MyLibrary.csproj` which references `SomePackage/1.0` and nothing else, but `SomePackage` transitively brings in `SomeTransitiveDependency/0.5`.  I also have another project `MyConsoleApp.csproj` that references `MyLibrary.csproj`.  In the .NET ecosystem, the console app project doesn't list that it has a dependency on `SomePackage/1.0` (and by extension doesn't list that it also depends on `SomeTransitiveDependency/0.5`) because that's resolved at build time.  Now with all that, say we're trying to update `SomeTransitiveDependency/0.5` to `0.6` to fix some vulnerability.  When we initially report what dependencies are present, should it look like this:

```
// note, this is what would actually be listed in the files in the repo
MyLibrary.csproj => SomePackage/1.0
MyConsoleApp.csproj => MyLibrary.csproj
```

or like this

```
MyLibrary.csproj => SomePackage/1.0, SomeTransitiveDependency/0.5
MyConsoleApp.csproj => MyLibrary.csproj
```

or like this

```
MyLibrary.csproj => SomePackage/1.0, SomeTransitiveDependency/0.5
MyConsoleApp.csproj => MyLibrary.csproj, SomePackage/1.0, SomeTransitiveDependency/0.5
```

(many more possibilities)

Ultimately the "proper" fix is to just add a direct reference to `SomeTransitiveDependency/0.6` in `MyLibrary.csproj` because then  `SomePackage/1.0` gets the benefit and the console app project gets the update as well, but if we reported that `MyConsoleApp.csproj` has a dependency on `SomePackage/1.0` **and** `SomeTransitiveDependency/0.5` then (my understanding is that) the expectation is for _both_ `.csproj` files to get a reference to `0.6`, but the owner of a repo wouldn't accept that fix because it's redundant.

deivid-rodriguez:
I'm a bit confused. `nuget` does not have lockfiles, right? So is the expectation that in case of a security vulnerabilty in a "hidden" transitive dependency, Dependabot will add a new top level dependency to fix it? That would sound like going too far for me, I would go bug the package manager about lockfiles 😅.

deivid-rodriguez:
I'm also confused about the fact that if the dependency parser is not able to detect transitive dependencies, how come you even try to update a transitive dependency? Or is the `GetAllPackageDependenciesAsync` also used in the `FileParser` step?

JoeRobich:
NuGet does have lockfiles, although they may wish it didn't. =p

The dependency_finder.rb is able to identify transitive dependencies. Long term we would want to move away from adding them as top-level dependencies, if the package that brought in the problematic version has a fixed version. With CentralPackageManagement and TransitiveVersionPinning the story is a little better, as it doesn't require changes to the project files, the potential issue of polluting the Directory.Packages.props with PackageVersion items that may no longer be necessary exists.

deivid-rodriguez:
Ok, so

> SDK-style projects only list top level dependencies, not any of the transitive ones (because MSBuild/NuGet figures them out at build time.)

So from this I assume at least SDK-style projects don't have lockfiles? Or else why would you need to build a dummy project to figure out which version of transitive dependencies is being used?

I guess it's starting to become apparent that I know nothing about this ecosystem 🤣 .

deivid-rodriguez:
> The `dependency_finder.rb` is able to identify transitive dependencies.

Ah, ok. So in that case the question becomes, why `GetAllPackageDependenciesAsync` chokes when finding transitive dependencies in https://github.com/microsoft/clrmd, while `DependencyFinder` seems to work just fine? I guess it's because of the project vs package distinction mentioned earlier?

Are lockfiles are shared for a whole project? Or are they package-specific? What's the name or an example of Nuget lockfiles so I can find out more?

Also, each package inside a project could be using a different version of the same dependency, right?

JoeRobich:
> why GetAllPackageDependenciesAsync chokes when finding transitive dependencies

I imagine the difference is that dependency_finder isn't as strict as MSBuild here.

> Are lockfiles are shared for a whole project? Or are they package-specific? What's the name or an example of Nuget lockfiles so I can find out more?

It does not have strong adoption within or outside Microsoft. Lockfiles are named packages.lock.json. Here are some linked to documentation on NuGet lockfile support.

- https://devblogs.microsoft.com/nuget/enable-repeatable-package-restores-using-a-lock-file/
- https://learn.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#locking-dependencies

> Also, each package inside a project could be using a different version of the same dependency, right?

Yes, there could be mulitple dependencies which in turn depend on different version of the same dependency. The dependency_finder is coded to take the highest version and work as if that will be the resolved version.

Ultimately, I would like for the NuGetUpdater tool to have an `--analyze` command which will use MSBuild and NuGet to generate all the information that will be needed by the file_parser and update_checker on the ruby side. It will be the only way to ensure the dependency resolution behavior matches.

brettfo:
@JoeRobich explained it nicely, but I wanted to re-mention similar aspects here.  If the NuGet updater has been asked to update a transitive dependency, we absolutely want to do it to eliminate the vulnerability, even if it requires adding a new top-level dependency (but in some cases the ideal fix would be to pull in a newer top-level dependency that _also_ updates the transitive).  In this case we don't want to add a lock file to their code.  Lock files _do_ exist in NuGet, but they're not commonly used and not a focus (although PR #6031 was created to enable updating those, but that PR is based on the old updater model.  Relatedly, I reached out separately to the author of that PR discussing the changes we're currently making and they're on board with closing/re-doing that PR once this big change is in.)

deivid-rodriguez:
Sure, I agree with what you said, I'm definitely not suggesting that we add a lockfile to eliminate a vulnerability in a transitive dependency 😅.

What I meant to say is that our take is usually that direct dependencies specifications should be used for dependencies that your project uses directly, not for "dependencies that your project uses directly plus indirect dependencies that have ever had a security vulnerability".

I don't think GitHub is able to parse "invisible" indirect dependencies (I don't think they show up [here](https://github.com/microsoft/clrmd/network/dependencies), for example?), so Dependabot should never be asked to update an indirect "invisible" dependency. It may happen if dependencies are manually uploaded through the dependency submission API, although not sure if NuGet is supported there.

I have a question though, shouldn't NuGet/MsBuild resolve to the latest (hopefully secure) version of a dependency if compatible with the rest of the package automatically? And if the latest (secure) version of a hidden dependency is incompatible with the rest of the package, how does pinning to it through a direct dependency fixes that? 🤔

JoeRobich:
> I have a question though, shouldn't NuGet/MsBuild resolve to the latest (hopefully secure) version of a dependency if compatible with the rest of the package automatically? 

The package version resolves to the highest requested version (`4.0.0` and `4.0.5` would resolve to `4.0.5`) based on dependencies, not the highest published version. Although if the dependencies use wildcards (4.* for instance), it may actually try for the highest published matching version. Also, since we are talking about MSBuild and NuGet, I am sure there are more scenarios that don't follow either behavior.

> if the latest (secure) version of a hidden dependency is incompatible with the rest of the package, how does pinning to it through a direct dependency fixes that?

You are right it wouldn't fix anything. The version_finder would filter out the 'fixed' version as incompatible from the possible_versions and no fix would be attempted.

deivid-rodriguez:
> The package version resolves to the highest requested version (4.0.0 and 4.0.5 would resolve to 4.0.5) based on dependencies, not the highest published version. Although if the dependencies use wildcards (4.* for instance), it may actually try for the highest published, matching version. Also, since we are talking about MSBuild and NuGet, I am sure there are more scenarios that don't follow either behavior.

Yeah, that's what I meant by "compatible with the rest of the package". Although I don't get the example you gave. If transitive dependency A requires transitive dependency X at version 4.0.0, and transitive dependency B also requires transitive dependency X at version 4.0.5, would that not be a conflict? Does NuGet has some concept of "soft/recommended" requirements, maybe?

> You are right it wouldn't fix anything. The version_finder would filter out the 'fixed' version as incompatible from the possible_versions and no fix would be attempted.

Ok, so can we clarify under which circumstances Dependabot will add a new top level dependency explicitly?
 
Overall, I'm a bit confused about the behavior we're bringing regarding transitive dependencies. Not saying it's not ok for Nuget, just pointing out that explicitly pinning transitive dependencies as new top level dependencies is not something we currently do for any ecosystem, and I can't make much sense of it at the moment.

Maybe some examples in a dummy project would help making this more clear.

JoeRobich:
> Does NuGet have some concept of "soft/recommended" requirements, maybe?

The most common version string format `"4.0.0"` is sort of this soft recommended requirement. If a 4.0.0 package exists and no other dependencies require something higher, then the package will resolve with version `4.0.0`. If the package doesn't exist in the package feed but version `4.0.1` does, then `4.0.1` will be the resolved version. If the package exists but another dependency is set to use `"4.0.5"` then `4.0.5` will be the resolved version.

There is a strict version match syntax `"[4.0.0]"` which will cause build errors, if the package is not present in the feed or if it conflicts with another dependency that needs a different version.

> can we clarify under which circumstances Dependabot will add a new top-level dependency explicitly?

Just a note, when we were talking about "compatibility" earlier, I was thinking we meant with regards to a package's supported frameworks. As in, a fixed package dropped support for the TFM used by the vulnerable repo.

The following applies to projects which use PackageReference as packages.config will have all dependencies listed.

If a vulnerability is reported against a transitive dependency:
  1. The repo uses Central Package Management
    a. The repo does not have TransitiveVersionPinning enabled
       A top-level package reference item will be added to the projects that bring in the transitive dependency.
    b. The repo does use TransitiveVersionPinning
       A package version item will be added to the Directory.Packages.props.
  2. The repo does not use Central Package Management
    A top-level package reference item will be added to the projects that bring in the transitive dependency.

Say a repo uses an RPC package which depends on Newtonsoft.Json, but has no direct dependency on it. If a vulnerability were reported against Newtonsoft.Json and a fix is published, then the NuGet updater will attempt to add top-level dependencies where necessary to resolve the vulnerability. If, however, the fixed Newtonsoft.Json package dropped support for the TFM used in the repo, then no dependencies would be added.

deivid-rodriguez:
> The most common version string format "4.0.0" is sort of this soft recommended requirement. If a 4.0.0 package exists and no other dependencies require something higher, then the package will resolve with version 4.0.0. If the package doesn't exist in the package feed but version 4.0.1 does, then 4.0.1 will be the resolved version. If the package exists but another dependency is set to use "4.0.5" then 4.0.5 will be the resolved version.

> There is a strict version match syntax "[4.0.0]" which will cause build errors, if the package is not present in the feed or if it conflicts with another dependency that needs a different version.

I see, that clarifies things, thanks! I believe Gradle has something similar.

> 1. The repo uses Central Package Management
>   a. The repo does not have TransitiveVersionPinning enabled
>     A top-level package reference item will be added to the projects that bring in the transitive dependency.
>  b. The repo does use TransitiveVersionPinning
>    A package version item will be added to the Directory.Packages.props.
> 2. The repo does not use Central Package Management
>   a. A top-level package reference item will be added to the projects that bring in the transitive dependency.

Ok. So if I understand correct, central package management is essentially a `Directory.Packages.props` file that enables setting explicit versions of dependencies that you want all packages to use. And if you use that, there's also a `TransitiveVersionPinning` feature that allows you to also choose version for transitive dependencies.

In my opinion, users who care about this should either enable these features or use lockfiles, and then Dependabot would properly fix their vulnerabilities. In other words, I think 1b if great, but 1a and 2a implement something we don't have for any other ecosystem and feels overkill to me.

That said, you know your ecosystem best, so if you think NuGet users want this, I won't block it. As I explained before, GitHub can't currently detect "invisible transitive dependencies", so we should never get security alerts for these anyways. So these features should be "dead code" for Github.com anyways if I'm understanding this correctly.

caspChristian:
So this needs to be reverted or fixed, preferrably also with a unit test that makes sure wildcard version strings isn't broken again
https://github.com/dependabot/dependabot-core/issues/9442#issuecomment-2597102827

PR #8174, opened 2023-10-10
Expose a new directory_not_found user error

deivid-rodriguez:
If the user is configuring a directory for dependabot that no longer exists, and the ecosystem does not clone full repositories, we'll end up with unknown `Octokit::NotFound` errors when trying to fetch dependency files.

This PR creates a new error that will be displayed to the user so that she can fix her configuration.

PR #8181, opened 2023-10-12
Improve detection of unsupported cargo toolchains

deivid-rodriguez:
We're getting the following flaky in cargo:

https://github.com/dependabot/dependabot-core/actions/runs/6493735738/job/17635263354

```
Failures:

  1) Dependabot::Cargo::UpdateChecker::VersionResolver latest_resolvable_version when using a toolchain that is too old raises a helpful error
     Failure/Error:
       expect { subject }
         .to raise_error(Dependabot::DependencyFileNotEvaluatable)

       expected Dependabot::DependencyFileNotEvaluatable, got #<Dependabot::SharedHelpers::HelperSubprocessFailed: Blocking waiting for file lock on package cache
       error: Unable to update registry `crates-io`

       Caused by:
         usage of sparse registries requires `-Z sparse-registry`> with backtrace:
         # ./lib/dependabot/cargo/update_checker/version_resolver.rb:153:in `run_cargo_command'
         # ./lib/dependabot/cargo/update_checker/version_resolver.rb:137:in `run_cargo_update_command'
         # ./lib/dependabot/cargo/update_checker/version_resolver.rb:307:in `block (2 levels) in original_requirements_resolvable?'
         # /home/dependabot/common/lib/dependabot/shared_helpers.rb:195:in `with_git_configured'
         # ./lib/dependabot/cargo/update_checker/version_resolver.rb:306:in `block in original_requirements_resolvable?'
         # /home/dependabot/common/lib/dependabot/shared_helpers.rb:56:in `block in in_a_temporary_directory'
         # /home/dependabot/common/lib/dependabot/shared_helpers.rb:56:in `chdir'
         # /home/dependabot/common/lib/dependabot/shared_helpers.rb:56:in `in_a_temporary_directory'
         # ./lib/dependabot/cargo/update_checker/version_resolver.rb:303:in `original_requirements_resolvable?'
         # ./lib/dependabot/cargo/update_checker/version_resolver.rb:298:in `resolvability_error?'
         # ./lib/dependabot/cargo/update_checker/version_resolver.rb:245:in `handle_cargo_errors'
         # ./lib/dependabot/cargo/update_checker/version_resolver.rb:59:in `rescue in fetch_latest_resolvable_version'
         # ./lib/dependabot/cargo/update_checker/version_resolver.rb:41:in `fetch_latest_resolvable_version'
         # ./lib/dependabot/cargo/update_checker/version_resolver.rb:33:in `latest_resolvable_version'
         # ./spec/dependabot/cargo/update_checker/version_resolver_spec.rb:71:in `block (3 levels) in <top (required)>'
         # ./spec/dependabot/cargo/update_checker/version_resolver_spec.rb:177:in `block (5 levels) in <top (required)>'
         # ./spec/dependabot/cargo/update_checker/version_resolver_spec.rb:177:in `block (4 levels) in <top (required)>'
         # /home/dependabot/common/spec/spec_helper.rb:46:in `block (2 levels) in <top (required)>'
     # ./spec/dependabot/cargo/update_checker/version_resolver_spec.rb:177:in `block (4 levels) in <top (required)>'
     # /home/dependabot/common/spec/spec_helper.rb:46:in `block (2 levels) in <top (required)>'

Finished in 33.38 seconds (files took 1.24 seconds to load)
360 examples, 1 failure

Failed examples:

rspec ./spec/dependabot/cargo/update_checker/version_resolver_spec.rb:176 # Dependabot::Cargo::UpdateChecker::VersionResolver latest_resolvable_version when using a toolchain that is too old raises a helpful error
```

It seems confusing to me because we fail to categorize the error as matching ``usage of sparse registries requires `-Z sparse-registry` `` which is what determines that the cargo version is too old. But still, the command being raised is including that.

The problem is that we fail to match the string in the original error, but then when later trying to figure out whether the error was a resolvability issue, we end up running another cargo command to see if the original requirements were resolvable, and in that case, the ``usage of sparse registries requires `-Z sparse-registry` `` string is in the output.

So I'm wondering why the original error did not include the string.

To be able to verify this, we could return "nil" from `original_requirements_resolvable` if this happens, and let the original error be raised in that case.

### DIFF
--- a/cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
+++ b/cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
@@ -244,7 +244,7 @@ module Dependabot
 
           raise Dependabot::DependencyFileNotResolvable, error.message if resolvability_error?(error.message)
 
-          raise error
+          raise
         end
         # rubocop:enable Metrics/AbcSize
         # rubocop:enable Metrics/PerceivedComplexity
--- a/cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
+++ b/cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
@@ -238,7 +238,7 @@ module Dependabot
             return nil
           end
 
-          if error.message.include?("usage of sparse registries requires `-Z sparse-registry`")
+          if using_old_toolchain?(error.message)
             raise Dependabot::DependencyFileNotEvaluatable, "Dependabot only supports toolchain 1.68 and up."
           end
 
@@ -248,6 +248,15 @@ module Dependabot
         end
         # rubocop:enable Metrics/AbcSize
         # rubocop:enable Metrics/PerceivedComplexity
+
+        def using_old_toolchain?(message)
+          return true if message.include?("usage of sparse registries requires `-Z sparse-registry`")
+
+          version_log = /rust version (?<version>\d.\d+)/.match(message)
+          return false unless version_log
+
+          version_class.new(version_log[:version]) < version_class.new("1.68")
+        end
 
         def unreachable_git_urls
           return @unreachable_git_urls if defined?(@unreachable_git_urls)
--- a/cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
+++ b/cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
@@ -295,7 +295,11 @@ module Dependabot
           return true if message.match?(/feature `[^\`]+` is required/)
           return true if message.include?("unexpected end of input while parsing major version number")
 
-          !original_requirements_resolvable?
+          original_requirements_resolvable = original_requirements_resolvable?
+
+          return false if original_requirements_resolvable == :unknown
+
+          !original_requirements_resolvable
         end
 
         def original_requirements_resolvable?
@@ -310,13 +314,15 @@ module Dependabot
 
           true
         rescue SharedHelpers::HelperSubprocessFailed => e
-          raise unless e.message.include?("no matching version") ||
-                       e.message.include?("failed to select a version") ||
-                       e.message.include?("no matching package named") ||
-                       e.message.include?("failed to parse manifest") ||
-                       e.message.include?("failed to update submodule")
-
-          false
+          if e.message.include?("no matching version") ||
+             e.message.include?("failed to select a version") ||
+             e.message.include?("no matching package named") ||
+             e.message.include?("failed to parse manifest") ||
+             e.message.include?("failed to update submodule")
+            false
+          else
+            :unknown
+          end
         end
 
         def workspace_native_library_update_error?(message)
--- a/cargo/lib/dependabot/cargo/file_updater/lockfile_updater.rb
+++ b/cargo/lib/dependabot/cargo/file_updater/lockfile_updater.rb
@@ -146,7 +146,7 @@ module Dependabot
           # returns a non-zero status
           return if process.success?
 
-          if stdout.include?("usage of sparse registries requires `-Z sparse-registry`")
+          if using_old_toolchain?(stdout)
             raise Dependabot::DependencyFileNotEvaluatable, "Dependabot only supports toolchain 1.68 and up."
           end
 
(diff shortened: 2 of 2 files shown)


deivid-rodriguez:
That did the trick and I can see the real culprit now!

deivid-rodriguez:
By the way, this flakes seems to be happening consistently right now? At least I'm not been able to get #8174 green by retrying!

cargo/spec/dependabot/cargo/file_updater/lockfile_updater_spec.rb
```diff
@@ -61,21 +61,6 @@
...
-        )
-      end
-      let(:dependency_files) { [manifest, lockfile, toolchain_file] }
-
-      it "raises a helpful error" do
-        expect { updated_lockfile_content }
-          .to raise_error(Dependabot::DependencyFileNotEvaluatable)
-      end
-    end
-
```
jakecoffman, line 78: Can the test go back in now that it works?

deivid-rodriguez: This is me completely removing the detection from lockfile updater, because I believe this should never be triggered. We always go through `UpdateChecker` first, and we would detect it there before?

If this is really needed, then I need to duplicate the fix in the updater to get the test to pass.

cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
```diff
@@ -295,7 +304,13 @@ def resolvability_error?(message)
...
           return true if message.include?("unexpected end of input while parsing major version number")
 
-          !original_requirements_resolvable?
+          original_requirements_resolvable = original_requirements_resolvable?
+
+          if original_requirements_resolvable == true
+            false
+          elsif original_requirements_resolvable == false
+            true
+          end
```
jakecoffman, line 313: It's still early here, trying to understand this change. It looks the same as the code it replaces?

cargo/spec/dependabot/cargo/file_updater/lockfile_updater_spec.rb
jakecoffman: Maybe move to UpdateChecker?

cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
deivid-rodriguez: The previous code returns always either `true` or `false`. The new code returns either `true`, `false`, or `nil` if we couldn't find a definitive hint about a resolvability issue.

deivid-rodriguez: Satisfying RuboCop requests made it quite unreadable, I'll add a comment.

jakecoffman (approved):
Looks good, would be good to have a test to make sure this doesn't break in the future. I recall it was causing a lot of unknown errors at the time.

cargo/spec/dependabot/cargo/file_updater/lockfile_updater_spec.rb
deivid-rodriguez: It's already there!

jakecoffman: Oh ok, then we're good to go!

cargo/lib/dependabot/cargo/update_checker/version_resolver.rb
jakecoffman: Not a big deal, I was trying to figure out the significance of the change.

deivid-rodriguez: Have a look at e1c8dceb565d02129b44b571ee46cb5814ed066f, it should be easier to read!

cargo/spec/dependabot/cargo/file_updater/lockfile_updater_spec.rb
deivid-rodriguez: Actually my assumption was incorrect. I think in the case of security updates, we won't go through `VersionResolver` sometimes but pick the lowest fixed version directly from `LatestVersionFinder`. So the check in `LockfileUpdater` may be still needed. I'll port the fix and restore the test just in case.

deivid-rodriguez: I duplicated the fix in `LockfileUpdater` at https://github.com/dependabot/dependabot-core/pull/8181/commits/39bb742bd287e222881358ff772a33e84eaa7f1b, and restored the spec 👍.

PR #8188, opened 2023-10-13
Pass user info to Sentry

deivid-rodriguez:
It could be useful to be able to prioritize errors by the number of repos affected.

### DIFF
--- a/updater/lib/dependabot/service.rb
+++ b/updater/lib/dependabot/service.rb
@@ -109,7 +109,10 @@ module Dependabot
           extra: extra.merge({
             dependency_name: dependency&.name,
             dependency_group: dependency_group&.name
-          }.compact)
+          }.compact),
+          user: {
+            id: job&.repo_owner
+          }
         }
       )
     end
--- a/updater/lib/dependabot/job.rb
+++ b/updater/lib/dependabot/job.rb
@@ -142,6 +142,10 @@ module Dependabot
       @repo_private
     end
 
+    def repo_owner
+      source&.organization
+    end
+
     def updating_a_pull_request?
       @updating_a_pull_request
     end
--- a/updater/spec/dependabot/service_spec.rb
+++ b/updater/spec/dependabot/service_spec.rb
@@ -250,14 +250,24 @@ RSpec.describe Dependabot::Service do
       Dependabot::DependabotError.new("Something went wrong")
     end
 
-    it "delegates error capture to Sentry (Raven)" do
+    it "delegates error capture to Sentry (Raven), adding user info if any" do
       service.capture_exception(error: error, tags: { foo: "bar" }, extra: { baz: "qux" })
 
-      expect(Raven).to have_received(:capture_exception).with(error, tags: { foo: "bar" }, extra: { baz: "qux" })
+      expect(Raven).to have_received(:capture_exception)
+        .with(error,
+              tags: {
+                foo: "bar"
+              },
+              extra: {
+                baz: "qux"
+              },
+              user: {
+                id: nil
+              })
     end
 
     it "extracts information from a job if provided" do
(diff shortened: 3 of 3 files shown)


updater/lib/dependabot/service.rb
```diff
@@ -83,7 +84,10 @@ def capture_exception(error:, job: nil, dependency: nil, dependency_group: nil,
           extra: extra.merge({
             dependency_name: dependency&.name,
             dependency_group: dependency_group&.name
-          }.compact)
+          }.compact),
+          user: {
+            id: Base64.encode64(job&.source&.url&.to_s)
```
jurre, line 89: I don't believe `job.source` has a `url` property? 🤔 here's what we pass:

```
def source
    {
      provider: repo_source_resolver.provider,
      repo: repo_source_resolver.repo,
      directory: repo_source_resolver.directory,
      branch: repo_source_resolver.branch,
      api_endpoint: repo_source_resolver.api_endpoint,
      hostname: repo_source_resolver.hostname,
    }
  end
```

deivid-rodriguez: I only read the code diagonally, but I understood that would would instantiate a `Dependabot::Source` out of that info that does a have `#url` method.

jurre: Ah, I see! Is it somewhat weird to pass a repo URL as "user" here? Should we instead pass the owner (account)?

deivid-rodriguez: Good point. I'm not sure! I think both would be useful, but I'm not sure which one would be more useful. In that case, I think we can use whatever is not weird 😄.

Let me update!

deivid-rodriguez: Updated this to use the owner. I'm also ignoring "provider" (GitHub for use), since I assume this will only be used in the context of a single provider.

updater/lib/dependabot/service.rb
```diff
@@ -83,7 +84,10 @@ def capture_exception(error:, job: nil, dependency: nil, dependency_group: nil,
           extra: extra.merge({
             dependency_name: dependency&.name,
             dependency_group: dependency_group&.name
-          }.compact)
+          }.compact),
+          user: {
+            id: (Base64.encode64(job&.repo_owner) if job&.repo_owner)
```
yeikel, line 89: Why is this encoding necessary?

deivid-rodriguez: We consider the repository name as private identifiable information, so we prefer to not pass that on to a third party like Sentry. The purpose of this is to be able to aggregate errors per user, but the actual user is not important, so I thought of sending it encoded.

deivid-rodriguez: I wasn't too concerned about whether this can be reversed or not, and just used the first utility that I saw available that would hide this to the casual Sentry user. I'll use a real crypto func that's hard to reverse, though. Thanks!

yeikel: Thanks! I think that it is very easy to tell that a String is base64 and if the concern is about privacy, then it would definitely be better to hash it. 

Unless you do need to see the actual value for debugging?

deivid-rodriguez: Nah, I think properly hashing it like you suggest is better, I'll change it!

yeikel (commented):
It is looking great now 🚀

updater/lib/dependabot/service.rb
deivid-rodriguez: @mctofu pointed out that since we're passing only repo owner in the end (as opposed to also the repo), and organizations are public on GitHub, it's fine to not hash anything here. So I removed the hashing.

jakecoffman (approved):

PR #8192, opened 2023-10-13
port Docker updater improvements from Azure DevOps

brettfo:
As part of the work inside Azure DevOps to improve dependabot (see #8179) there were some updates to the Docker updater that needed to be split out into their own PR, specifically:

- Allow docker version prefixes to contain underscores
- Add support for digests when updating helm charts

(cc @brbayes-msft)

### DIFF
--- a/docker/lib/dependabot/docker/file_parser.rb
+++ b/docker/lib/dependabot/docker/file_parser.rb
@@ -24,12 +24,14 @@ module Dependabot
 
       FROM = /FROM/i
       PLATFORM = /--platform\=(?<platform>\S+)/
-      TAG = /:(?<tag>[\w][\w.-]{0,127})/
+      TAG_NO_PREFIX = /(?<tag>[\w][\w.-]{0,127})/
+      TAG = /:#{TAG_NO_PREFIX}/
       DIGEST = /(?<digest>[0-9a-f]{64})/
       NAME = /\s+AS\s+(?<name>[\w-]+)/
       FROM_LINE =
         %r{^#{FROM}\s+(#{PLATFORM}\s+)?(#{REGISTRY}/)?
           #{IMAGE}#{TAG}?(?:@sha256:#{DIGEST})?#{NAME}?}x
+      TAG_WITH_DIGEST = /^#{TAG_NO_PREFIX}(?:@sha256:#{DIGEST})?/x
 
       AWS_ECR_URL = /dkr\.ecr\.(?<region>[^.]+)\.amazonaws\.com/
 
@@ -168,18 +170,20 @@ module Dependabot
 
       def parse_helm(img_hash)
         repo = img_hash.fetch("repository", nil)
-        tag = img_hash.key?("tag") ? img_hash.fetch("tag", nil) : img_hash.fetch("version", nil)
+        tag_value = img_hash.key?("tag") ? img_hash.fetch("tag", nil) : img_hash.fetch("version", nil)
         registry = img_hash.fetch("registry", nil)
 
-        if !repo.nil? && !registry.nil? && !tag.nil?
-          ["#{registry}/#{repo}:#{tag}"]
-        elsif !repo.nil? && !tag.nil?
-          ["#{repo}:#{tag}"]
-        elsif !repo.nil?
-          [repo]
-        else
-          []
-        end
+        tag_details = tag_value.to_s.match(TAG_WITH_DIGEST).named_captures
+        tag = tag_details["tag"]
+        digest = tag_details["digest"]
+
+        return [] unless repo
+        return [repo] unless tag
+
+        image = "#{repo}:#{tag}"
+        image.prepend("#{registry}/") if registry
+        image.append("@sha256:#{digest}/") if digest
+        [image]
       end
     end
   end
--- a/docker/lib/dependabot/docker/file_updater.rb
+++ b/docker/lib/dependabot/docker/file_updater.rb
@@ -158,7 +158,7 @@ module Dependabot
         old_tags.each do |old_tag|
           old_tag_regex = /^\s+(?:-\s)?(?:tag|version):\s+["']?#{old_tag}["']?(?=\s|$)/
           modified_content = modified_content.gsub(old_tag_regex) do |old_img_tag|
-            old_img_tag.gsub(old_tag.to_s, new_yaml_tag(file).to_s)
+            old_img_tag.gsub(old_tag.to_s, new_helm_tag(file).to_s)
           end
         end
         modified_content
@@ -187,11 +187,6 @@ module Dependabot
         "#{prefix}#{dependency.name}#{tag}#{digest}"
       end
 
-      def new_yaml_tag(file)
-        element = dependency.requirements.find { |r| r[:file] == file.name }
-        element.fetch(:source)[:tag] || ""
-      end
-
       def old_yaml_images(file)
         previous_requirements(file).map do |r|
           prefix = r.fetch(:source)[:registry] ? "#{r.fetch(:source)[:registry]}/" : ""
@@ -203,8 +198,17 @@ module Dependabot
 
       def old_helm_tags(file)
         previous_requirements(file).map do |r|
-          r.fetch(:source)[:tag] || ""
+          tag = r.fetch(:source)[:tag] || ""
+          digest = r.fetch(:source)[:digest] ? "@sha256:#{r.fetch(:source)[:digest]}" : ""
+          "#{tag}#{digest}"
         end
+      end
+
+      def new_helm_tag(file)
+        element = dependency.requirements.find { |r| r[:file] == file.name }
+        tag = element.fetch(:source)[:tag] || ""
+        digest = element.fetch(:source)[:digest] ? "@sha256:#{element.fetch(:source)[:digest]}" : ""
+        "#{tag}#{digest}"
       end
 
       def requirements(file)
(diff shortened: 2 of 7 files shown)


abdulapopoola:
Thank you so much for this! Would it be possible to have some specs to avoid regressions in the future please?

docker/lib/dependabot/docker/tag.rb
```diff
@@ -9,8 +9,8 @@
       WORDS_WITH_BUILD = /(?:(?:-[a-z]+)+-[0-9]+)+/
       VERSION_REGEX = /v?(?<version>[0-9]+(?:\.[0-9]+)*(?:_[0-9]+|\.[a-z0-9]+|#{WORDS_WITH_BUILD}|-(?:kb)?[0-9]+)*)/i
       VERSION_WITH_SFX = /^#{VERSION_REGEX}(?<suffix>-[a-z][a-z0-9.\-]*)?$/i
-      VERSION_WITH_PFX = /^(?<prefix>[a-z][a-z0-9.\-]*-)?#{VERSION_REGEX}$/i
-      VERSION_WITH_PFX_AND_SFX = /^(?<prefix>[a-z\-]+-)?#{VERSION_REGEX}(?<suffix>-[a-z\-]+)?$/i
+      VERSION_WITH_PFX = /^(?<prefix>[a-z][a-z0-9.\-_]*-)?#{VERSION_REGEX}$/i
```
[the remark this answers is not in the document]
ryanbrandenburg: @abdulapopoola I'm not the greatest regex wizard but I think this would be a pre-existing condition (IE, this was not caused by our change to this REGEX but was already "inefficient" and just somehow grandfather-ed in). Do you have any advise for resolving this given that that's the case.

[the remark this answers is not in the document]
abdulapopoola: As a team, I think we accept RE as necessary and that codeQL flagging might be too sensitive; tagging @deivid-rodriguez and @jakecoffman to get their thoughts.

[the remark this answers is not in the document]
deivid-rodriguez: I think it's fine for CodeQL to flag these but I don't like it that it flags issues on _any_ changed lines, even if they were present also before the changes. It places unnecessary burden on contributions like here.

Anyways, I believe it's fine to ignore.

Still, I had a look and I think this change may fix it, and it does not seem to break any specs:

```diff
diff --git a/docker/lib/dependabot/docker/tag.rb b/docker/lib/dependabot/docker/tag.rb
index 65c21e267..8eb805614 100644
--- a/docker/lib/dependabot/docker/tag.rb
+++ b/docker/lib/dependabot/docker/tag.rb
@@ -8,7 +8,7 @@ module Dependabot
     class Tag
       WORDS_WITH_BUILD = /(?:(?:-[a-z]+)+-[0-9]+)+/
       VERSION_REGEX = /v?(?<version>[0-9]+(?:\.[0-9]+)*(?:_[0-9]+|\.[a-z0-9]+|#{WORDS_WITH_BUILD}|-(?:kb)?[0-9]+)*)/i
-      VERSION_WITH_SFX = /^#{VERSION_REGEX}(?<suffix>-[a-z][a-z0-9.\-]*)?$/i
+      VERSION_WITH_SFX = /^#{VERSION_REGEX}(?<suffix>-[a-z][a-z0-9.]*)?$/i
       VERSION_WITH_PFX = /^(?<prefix>[a-z][a-z0-9.\-]*-)?#{VERSION_REGEX}$/i
       VERSION_WITH_PFX_AND_SFX = /^(?<prefix>[a-z\-]+-)?#{VERSION_REGEX}(?<suffix>-[a-z\-]+)?$/i
       NAME_WITH_VERSION =
```

Just in case we want to change that separately.

docker/spec/dependabot/docker/file_parser_spec.rb
```diff
@@ -462,6 +462,34 @@
...
+              registry: "registry-host.io:5000",
+              tag: "17.04_"
+            }
+          }]
+        end
+
+        it "has the right details" do
+          expect(dependency).to be_a(Dependabot::Dependency)
+          expect(dependency.name).to eq("myreg/ubuntu")
+          expect(dependency.version).to eq("17.04_")
```
deivid-rodriguez, line 487: Is it this a real example, or just a simplified version for the sake of testing? Do you have an example of a realworld image tagged like this?

So far we've been trying to support to most widely used versioning schemes out there since there's no official spec for versioning docker tags. But arbitrarily putting an underscore after the version number does not make a lot of sense to me.

ryanbrandenburg: This is not a real example. The version we're thinking of is non-public, but it's a docker container for which the tag is shaped something like `someRepo_19700101.4`. I can replace the above with that now that I know what shape we're imagining.

ryanbrandenburg:
This should be ready for re-review.

deivid-rodriguez:
Would it be possible to split this PR into two commits, one for the helm charts feature, and another way for the detection of docker images with underscored suffixes? That'd help with the review!

docker/spec/dependabot/docker/file_parser_spec.rb
deivid-rodriguez: So, just to confirm, you need to support underscores as suffixes to the version, not as prefixes as originally stated, correct?

ryanbrandenburg:
I have rebased the commits as per your request.

docker/lib/dependabot/docker/file_updater.rb
```diff
@@ -158,7 +158,7 @@ def update_helm(file, content)
         old_tags.each do |old_tag|
           old_tag_regex = /^\s+(?:-\s)?(?:tag|version):\s+["']?#{old_tag}["']?(?=\s|$)/
           modified_content = modified_content.gsub(old_tag_regex) do |old_img_tag|
-            old_img_tag.gsub(old_tag.to_s, new_yaml_tag(file).to_s)
+            old_img_tag.gsub(old_tag.to_s, new_helm_tag(file).to_s)
```
deivid-rodriguez, line 161: The old method is dead now?

docker/lib/dependabot/docker/file_parser.rb
```diff
@@ -181,6 +193,8 @@ def parse_helm(img_hash)
           []
         end
       end
+      # rubocop:enable Metrics/PerceivedComplexity
+      # rubocop:enable Metrics/AbcSize
```
deivid-rodriguez, line 197: This got a bit out of hand, let me propose a simpler way:

```diff
-      # rubocop:disable Metrics/AbcSize
-      # rubocop:disable Metrics/PerceivedComplexity
       def parse_helm(img_hash)
         repo = img_hash.fetch("repository", nil)
         tag_value = img_hash.key?("tag") ? img_hash.fetch("tag", nil) : img_hash.fetch("version", nil)
@@ -178,22 +176,14 @@ module Dependabot
         tag = tag_details["tag"]
         digest = tag_details["digest"]
 
-        if !repo.nil? && !registry.nil? && !tag.nil? && !digest.nil?
-          ["#{registry}/#{repo}:#{tag}@sha256:#{digest}"]
-        elsif !repo.nil? && !registry.nil? && !tag.nil?
-          ["#{registry}/#{repo}:#{tag}"]
-        elsif !repo.nil? && !tag.nil? && !digest.nil?
-          ["#{repo}:#{tag}@sha256:#{digest}"]
-        elsif !repo.nil? && !tag.nil?
-          ["#{repo}:#{tag}"]
-        elsif !repo.nil?
-          [repo]
-        else
-          []
-        end
+        return [] unless repo
+        return [repo] unless tag
+
+        image = "#{repo}:#{tag}"
+        image.prepend("#{registry}/") if registry
+        image.append("@sha256:#{digest}/") if digest
+        [image]
       end
-      # rubocop:enable Metrics/PerceivedComplexity
-      # rubocop:enable Metrics/AbcSize
```

Also, I think we have no tests for parsing helm charts, not this PRs job but we should probably add some.

docker/spec/dependabot/docker/file_updater_spec.rb
```diff
@@ -1316,6 +1358,46 @@
...
+
+      describe "the updated helmfile" do
+        subject(:updated_helmfile) do
+          updated_files.find { |f| f.name == "values.yaml" }
+        end
+
+        its(:content) { is_expected.to include "image:\n  repository: 'nginx'\n  tag: 1.14.3\n" }
+      end
+    end
+
```
deivid-rodriguez, line 1400: If I completely revert the changes in `docker/lib` about updating heml charts, these specs still pass? Maybe they're not covering the right code?

deivid-rodriguez (commented):
This is looking good. The changes for supporting underscore prefixes look good to me, feel free to propose separately if you want those merged straightaway. For heml charts I had some questions.

docker/lib/dependabot/docker/file_updater.rb
brettfo: Correct, that method is no longer used, I'll remove it shortly.

docker/lib/dependabot/docker/file_parser.rb
brettfo: Good call!  Will update and push changes.

docker/spec/dependabot/docker/file_updater_spec.rb
brettfo: I think the test name was a copy/paste error.  I've renamed the test and pushed an update.

brettfo:
@deivid-rodriguez I've cleaned up the commits based on your feedback and rebased on the current `main`.

jakecoffman (approved):

PR #8193, opened 2023-10-13
stop processing if updated_deps is empty

brettfo:
As part of the work inside Azure DevOps to improve dependabot (see #8179) there were other updates that needed to be split out into their own PR, specifically not continuing with an update if the updated set was empty.


updater/lib/dependabot/updater/operations/update_all_versions.rb
```diff
@@ -99,6 +100,12 @@ def check_and_create_pull_request(dependency)
             requirements_to_unlock: requirements_to_unlock
           )
 
+          if updated_deps.empty?
+            return Dependabot.logger.info(
+              "No update possible for #{dependency.name} #{dependency.version}"
+            )
+          end
```
yeikel, line 107: Under what situation can this happen? A test would help

brettfo: We found an instance internally where the NuGet updater errored out and the result was an empty update set.  While we fixed the specific issue we had, there still remains the possibility that another updater could also error out and result in an empty set, and that's why we ultimately added this check as well, to prevent the pr creator from crashing.  If there is a better place for this check we can certainly move it.

deivid-rodriguez: I see.

I think it makes sense to avoid crashing randomly while trying to create a PR. However, I don't think this is an expected error condition, but a situation that's pointing at a Dependabot bug somewhere.

So, instead of just logging a message, we should be raising an error if this happens. Maybe something like:

```ruby
raise "Dependabot found some dependency requirements to unlock, yet it failed to update any dependencies"
```

ryanbrandenburg:
I have adjusted the change per your suggestion, added a test for this scenario, and updated the branch. The PR should be ready for re-review.

updater/spec/dependabot/updater/operations/update_all_versions_spec.rb
```diff
@@ -0,0 +1,127 @@
+# typed: true
+# frozen_string_literal: true
+
+require "spec_helper"
+require "support/dummy_pkg_helpers"
+require "support/dependency_file_helpers"
+
```
jakecoffman, line 7: I can't run this test directly with `script/ci-test-updater updater/spec/dependabot/updater/operations/update_all_versions_spec.rb`, there are some missing requires:

```suggestion

require "dependabot/service"
require "dependabot/dependency_snapshot"
require "dependabot/updater/error_handler"
```

updater/spec/dependabot/updater/operations/update_all_versions_spec.rb
```diff
@@ -0,0 +1,127 @@
...
+  let(:mock_error_handler) do
+    instance_double(Dependabot::Updater::ErrorHandler)
+  end
+
+  before do
+    allow(Dependabot.logger).to receive(:info)
+    stub_request(:get, "https://rubygems.org/api/v1/gems/business.json")
+      .to_return(status: 200, body: fixture("ruby/rubygems_response.json"))
+    stub_request(:get, "https://rubygems.org/api/v1/versions/business.json")
+      .to_return(status: 200, body: fixture("ruby/versions-business.json"))
```
jakecoffman, line 83: Since this is testing the updater, it shouldn't be necessary to mock out calls, it shouldn't be making any! Indeed if I delete these the test still passes.

```suggestion
```

updater/spec/dependabot/updater/operations/update_all_versions_spec.rb
```diff
@@ -0,0 +1,127 @@
...
+    let(:dependency_previous_version) { "1.5.0" }
+    let(:requirements) do
+      [{ file: "Gemfile", requirement: "~> 1.5.0", groups: [], source: nil }]
+    end
+    let(:previous_requirements) do
+      [{ file: "Gemfile", requirement: "~> 1.5.0", groups: [], source: nil }]
+    end
+
+    it "does not update any dependencies" do
+      expect { subject.send(:check_and_create_pull_request, dependency) }.to raise_error(
```
jakecoffman, line 122: We don't usually test private methods directly. That said it looks like if you take this suggestion you'll need to change this test a bit to get it to work. DependencySnapshot wants something it can parse, might be nice if there was a way to create one and inject a fake dependency.

```suggestion
      expect { subject.perform }.to raise_error(
```

updater/spec/dependabot/updater/operations/update_all_versions_spec.rb
jakecoffman: To verify switch this to an expectation in the test, like:

```
expect(WebMock).to have_requested(:get, "http://something.com/users/123").once
```

updater/spec/dependabot/updater/operations/empty_return_update_checker.rb
```diff
@@ -0,0 +1,40 @@
+# typed: true
+# frozen_string_literal: true
+
+module Dependabot
+  module EmptyReturn
+    class UpdateChecker < Dependabot::UpdateCheckers::Base
```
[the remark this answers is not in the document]
ryanbrandenburg: I don't follow what you mean about a "Bundler situation", could you clarify? I landed on using fake classes because the pattern seems to be to register a class (not an instance of the class) that will then be created to run the required items against. Did you maybe mean to suggest that we use `allow_any_instance_of` to override the behavior of the bundler `update_checker.rb` (and switch our test object to call bundler instead of the made up "emptyreturn"? If I do that in a `before` block does it stay contained to the relevant tests/context? I'm new to Ruby and unsure about how contexts/state is maintained (for example the requires above are no doubt missing because my instinct from C# is "if it compiles it has all the 'usings' it needs.").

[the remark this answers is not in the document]
jakecoffman: Sometimes it's easier to write some code than to explain. I think this should work: b078bec5619e0eb9ba7c2d435c7e444c1eafe5d4

> Did you maybe mean to suggest that we use allow_any_instance_of to override the behavior of the bundler update_checker.rb (and switch our test object to call bundler instead of the made up "emptyreturn")?

Yes!

> If I do that in a before block does it stay contained to the relevant tests/context?

Yes!

updater/spec/dependabot/updater/operations/update_all_versions_spec.rb
ryanbrandenburg: Yeah, I was originally trying to call `perform` (if only because I didn't know you could do `.send` for private methods in Ruby) and it ended up needing a lot of extra context/mocks that wasn't directly related to the change we had made.

In the absence of existing tests of this class or a similar example for parsing-mocks I figured it was better to do the simpler version of things, and when someone adds more tests to this area maybe it makes sense to switch it over. I can add a comment to that effect if it helps. Does that allay your concerns?

updater/spec/dependabot/updater/operations/update_all_versions_spec.rb
ryanbrandenburg: I'm not sure I follow the second comment, didn't you say in the first comment that there shouldn't be any http requests? What would I be setting the expectation for?

updater/spec/dependabot/updater/operations/empty_return_update_checker.rb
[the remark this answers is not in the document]
ryanbrandenburg: I integrated that commit in manually, (trying to make sure I understand it and I already had some changes from other feedback). Does that look good?

jakecoffman (approved):

PR #8204, opened 2023-10-16
Ensure Grouped Security Updates are rebased correctly

jurre:
Currently, when rebasing a grouped security update, we will default to creating individual PRs for each dependency, since the updater does not know how to refresh grouped updates.

This adds a basic strategy that reuses the refresh behavior for Version Update groups.


jakecoffman (approved):
LGTM

PR #8210, opened 2023-10-16
Add OpenTelemetry SDK

JamieMagee:
This PR roughly follows the [OpenTelemetry getting started documentation][1] for Ruby.

It adds the `opentelemetry-sdk` package to both `dependabot-updater`, for configuration, and `dependabot-common`, for use throughout the codebase.

Where I differ from the getting started guide, is instead of using the `opentelemetry-instrumentation-all` package, I install instrumentation packages individually. I've started with HTTP clients initially:

- [`opentelemetry-instrumentation-excon`][2]
- [`opentelemetry-instrumentation-faraday`][3] (Used internally by `octokit`)
- [`opentelemetry-instrumentation-http`][4]

Finally, all of this is gated behind an environment variable `OTEL_ENABLED` (This matches the [OpenTelemetry environment variable naming scheme][5]). OpenTelemetry packages are only `require`d if `OTEL_ENABLED` is set to `true`.

NB: by default OpenTelemetry exports in [OTLP format][6] to an [OpenTelemetry Collector][7] listening at `localhost:4318`. This can be overridden by setting the `OTEL_TRACES_EXPORTER` environment variable to `console`.

[1]: https://opentelemetry.io/docs/instrumentation/ruby/getting-started/
[2]: https://github.com/open-telemetry/opentelemetry-ruby-contrib/tree/main/instrumentation/excon#readme
[3]: https://github.com/open-telemetry/opentelemetry-ruby-contrib/tree/main/instrumentation/faraday#readme
[4]: https://github.com/open-telemetry/opentelemetry-ruby-contrib/tree/main/instrumentation/http#readme
[5]: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/
[6]: https://opentelemetry.io/docs/specs/otlp/
[7]: https://opentelemetry.io/docs/collector/


jurre:
We don't really have a good/safe way to provision Dependabot with secrets for an OTel provider, which is why don't have this or something like Datadog running in this environment currently. The problem here is that if we were to provide the tokens directly to dependabot-core, they could be exfiltrated by any ecosystem that is able to run arbitrary code in the isolated dependabot VMs (python, ruby, elixir to name a few)

JamieMagee:
I understand that you don't want to directly place Datadog tokens into Dependabot updaters, but I think OpenTelemetry has recommended architectures for collecting signals that take this into account.

I think the method you are describing most closely aligns with OpenTelemetry's ["No collector"][1] architecture. Where the application itself talks directly to the monitoring platform. For example:

image

While this would work, I'm not suggesting it for the reasons you listed above. Though, it does sound like Datadog's [client tokens][2] are designed for this sort of architecture.

The next standard architecture is the ["agent"][3] architecture, where the OpenTelemetry instrumented application exports signals in OTLP format to an [OpenTelemetry collector][4], by default at `localhost:4318`.

image

In this architecture the client token for Datadog would live in the OpenTelemetry collector, likely deployed as a Docker container, completely separate from Dependabot itself. OpenTelemetry also publish [security best practices][5] for the collector if you're interested in reading about their advice and some threat modelling.

The final recommended architecture is the ["gateway"][6] architecture, where a load balancer sits between the OpenTelemetry instrumented application and a several instances of OpenTelemetry collectors. This is an extension of the "agent" architecture, and mainly provides better scalability.

image

All that is to say, I think there is an architecture for OpenTelemetry that will work with Datadog. However, this PR doesn't even get that far, as OpenTelemetry is explicitly disabled by default. I can't see a way for any malicious actor to enable and exfiltrate telemetry, as it requires setting the `OTEL_ENABLED` environment variable early in the setup process, before any untrusted code is run.

[1]: https://opentelemetry.io/docs/collector/deployment/no-collector/
[2]: https://docs.datadoghq.com/account_management/api-app-keys/#client-tokens
[3]: https://opentelemetry.io/docs/collector/deployment/agent/
[4]: https://opentelemetry.io/docs/collector/
[5]: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/security-best-practices.md
[6]: https://opentelemetry.io/docs/collector/deployment/gateway/

JamieMagee:
I'm not sure that would work.

Yes, in this PR I've focused on HTTP calls, which in theory could be captured from the proxy (though they would have to be manually instrumented, instead of using the automatic instrumentation). But it would completely prevent any [manual instrumentation][1].

[1]: https://opentelemetry.io/docs/instrumentation/ruby/manual/

jakecoffman:
I guess that is over-engineering it, we could add the Otel SDK to the Proxy and make it the backend then we're in the "No collector" architecture. The Proxy can decide where to send the data.

JamieMagee:
The OpenTelemetry SDK is the part that allows us to instrument the codebase though. Without that we're basically building our own telemetry library. Plus, how do we export telemetry that is not an HTTP call to the proxy?

We can (and should) force the OpenTelemetry collector to talk via the proxy.

jakecoffman (approved):
Had a chat with @JamieMagee, basically the configuration will be going to a Collector running on the Dependabot infrastructure network as a job runs to collect the data and translate it to send it to the backend based on the config.

Any sensitive variables will not be in the Updater container, it's sending OTLP locally and the Collector exports it from there.

sorbet/rbi/shims/opentelemetry-sdk.rbi
```diff
@@ -0,0 +1,7 @@
+# typed: strong
+# frozen_string_literal: true
+
+module OpenTelemetry
+  module SDK
+  end
```
JamieMagee, line 6: NOTE: this can't be stubbed better, because Sorbet needs the block to be named. See:
- https://sorbet.org/docs/procs#annotating-methods-that-use-yield
- https://github.com/open-telemetry/opentelemetry-ruby/blob/afc7bc56b39adae925ebd465948a722bd64e1d2e/sdk/lib/opentelemetry/sdk.rb#L63-L66

PR #8214, opened 2023-10-17
Add types to `FileUpdaters::Base`

JamieMagee:
A couple of notable things I'd like to highlight about this PR:

- Reordered the parameters of `initialize` to fix [`Sorbet/KeywordArgumentOrdering`: Optional keyword arguments must be at the end of the parameter list][1]
- It's a little unclear what the types of `credentials` and `options` should be
	- I was more confident in `credentials` being `T::Hash[String, String]`
	- I was less confident in `options` so it's `T::Hash[String, T.untyped]`
- Methods marked with [`overridable`][2] should likely be marked as `abstract`, but `abstract` methods are required to have no body, and I am not sure if anything is depending on them raising a `NotImplementedError`
- Changes to other files are type system 'escape hatches' as they haven't been properly typed yet and I wanted to keep this PR small

Part of #7782

[1]: https://github.com/Shopify/rubocop-sorbet/blob/main/manual/cops_sorbet.md#sorbetkeywordargumentordering
[2]: https://sorbet.org/docs/override-checking

### DIFF
--- a/common/lib/dependabot/file_updaters/base.rb
+++ b/common/lib/dependabot/file_updaters/base.rb
@@ -12,12 +12,16 @@ module Dependabot
 
       sig { returns(T::Array[Dependabot::Dependency]) }
       attr_reader :dependencies
+
       sig { returns(T::Array[Dependabot::DependencyFile]) }
       attr_reader :dependency_files
+
       sig { returns(T.nilable(String)) }
       attr_reader :repo_contents_path
+
       sig { returns(T::Array[T::Hash[String, String]]) }
       attr_reader :credentials
+
       sig { returns(T::Hash[Symbol, T.untyped]) }
       attr_reader :options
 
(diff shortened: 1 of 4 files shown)


common/lib/dependabot/file_updaters/base.rb
```diff
@@ -1,18 +1,41 @@
...
+      sig { returns(T::Array[Dependabot::Dependency]) }
+      attr_reader :dependencies
+      sig { returns(T::Array[Dependabot::DependencyFile]) }
+      attr_reader :dependency_files
+      sig { returns(T.nilable(String)) }
+      attr_reader :repo_contents_path
+      sig { returns(T::Array[T::Hash[String, String]]) }
+      attr_reader :credentials
+      sig { returns(T::Hash[Symbol, T.untyped]) }
+      attr_reader :options
```
deivid-rodriguez, line 22: I'd find this easier to parse if we added a blank line between each sig+attribute duple. What do you think?

deivid-rodriguez (approved):

common/lib/dependabot/file_updaters/base.rb
JamieMagee: Absolutely! Can do.

PR #8220, opened 2023-10-17
Add more debugging info for an unknown error

deivid-rodriguez:


deivid-rodriguez:
Closing because I think I found the issue. Will open a separate PR!

PR #8224, opened 2023-10-17
Bump Bundler to 2.4.22

deivid-rodriguez:


deivid-rodriguez:
Finally found some time to wrap this up. Tricky upgrade, lots of things going on upstream!

deivid-rodriguez:
Ended up going all the way to Bundler 2.4.22, which was just released.

bundler/helpers/v2/lib/functions/version_resolver.rb
```diff
@@ -127,7 +127,7 @@ def lockfile
     def fetcher_class(dep)
       return unless dep.source.is_a?(::Bundler::Source::Rubygems)
 
-      dep.source.fetchers.first.fetchers.first.class.to_s
+      dep.source.fetchers.first.send(:fetchers).first.class.to_s
```
landongrindheim, line 130: 💅 I'm sure you've done your homework and this is safe, but `#send` always makes me nervous 😰 What do you think about using `#public_send` to pacify my neuroses?

```suggestion
      dep.source.fetchers.first.public_send(:fetchers).first.class.to_s
```

landongrindheim (approved):
I love the patch-by-patch approach you took in this upgrade ❤️ Looks great to me! 🚀

bundler/helpers/v2/lib/functions/version_resolver.rb
deivid-rodriguez: Actually the problem I "solved" here is that this method is now private, that's why I switched to `send`.

I know this sucks but I looked into using/providing a better interface, but nothing obvious came up, and I didn't want to block the upgrade on that.

Also considered reverting to a public method upstream, but I'm not sure dependabot-core was using it is a good justification to make it public again.

PR #8226, opened 2023-10-18
Make composer CI pass consistently

deivid-rodriguez:
It seems permanently broken right now.

### DIFF
--- a/composer/lib/dependabot/composer/file_fetcher.rb
+++ b/composer/lib/dependabot/composer/file_fetcher.rb
@@ -9,6 +9,7 @@ module Dependabot
   module Composer
     class FileFetcher < Dependabot::FileFetchers::Base
       require_relative "file_fetcher/path_dependency_builder"
+      require_relative "helpers"
 
       def self.required_files_in?(filenames)
         filenames.include?("composer.json")
--- a/composer/spec/spec_helper.rb
+++ b/composer/spec/spec_helper.rb
@@ -12,5 +12,9 @@ end
 require "#{common_dir}/spec/spec_helper.rb"
 
 RSpec.configure do |config|
+  config.before(:suite) do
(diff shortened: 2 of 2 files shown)


jurre (approved):

composer/lib/dependabot/composer/update_checker/version_resolver.rb
```diff
@@ -41,7 +41,7 @@ def initialize(extensions)
         SOURCE_TIMED_OUT_REGEX =
           /The "(?<url>[^"]+packages\.json)".*timed out/
         FAILED_GIT_CLONE_WITH_MIRROR = /Failed to execute git clone --(mirror|checkout)[^']*'(?<url>.*?)'/
-        FAILED_GIT_CLONE = /Failed to clone (?<url>.*?)/
+        FAILED_GIT_CLONE = /Failed to clone (?<url>[^ ]+)/
```
deivid-rodriguez, line 44: I'm confused how this was working. Since this was a non greedy match, I believe it would always match an empty URL. Yet we have some tests for this and they generally pass 🤔.

deivid-rodriguez:
I think my fix makes sense, but does not make CI pass consistently. I do have another change locally that feels more promising. At least it hasn't failed after 20 full runs :). I'll propose it separately.

composer/lib/dependabot/composer/update_checker/version_resolver.rb
deivid-rodriguez: I removed this change from the PR since it's not needed to get tests green and I want to investigate it a bit more.

deivid-rodriguez:
This is ready now!

deivid-rodriguez:
Merging to unblock other PRs.

I believe the fix makes sense, so that each parallel test process does not interfere with others. We'll see how it goes.

PR #8238, opened 2023-10-19
check types in the updater too

jakecoffman:
Running `bundle exec srb tc` was not checking the types in the updater, so adding `--dir updater` seems to fix that. 

I ran some other commands too and it did .rbi stuff.


jakecoffman:
@JamieMagee look ok? It seems to be working!

abdulapopoola (approved):

JamieMagee (approved):

PR #8239, opened 2023-10-19
add types to the ApiClient

jakecoffman:
A few of these types I am unsure about, but as we type the Operations files we'll find out and fix them!


jakecoffman:
Smoke tests showing me the way!

updater/lib/dependabot/api_client.rb
```diff
@@ -170,18 +193,25 @@ def increment_metric(metric, tags:)
 
     private
 
-    attr_reader :base_url, :job_id, :job_token
+    sig { returns(String) }
+    attr_reader :base_url
+    sig { returns(T.any(String, Integer)) }
+    attr_reader :job_id
+    sig { returns(String) }
+    attr_reader :job_token
```
jurre, line 201: Incredibly pedantic, but I find this quite hard to read, and think its a bit better with the spaces?

```suggestion
    sig { returns(String) }
    attr_reader :base_url

    sig { returns(T.any(String, Integer)) }
    attr_reader :job_id

    sig { returns(String) }
    attr_reader :job_token
```

updater/lib/dependabot/api_client.rb
```diff
@@ -170,18 +193,25 @@ def increment_metric(metric, tags:)
...
 
+    sig { returns(T.untyped) }
```
jurre, line 203: This doesn't return a `HTTP::Client`? 🤔

jurre (approved):

updater/lib/dependabot/api_client.rb
jakecoffman: Works for me, can we make a rubocop rule that enforces this?

updater/lib/dependabot/api_client.rb
jakecoffman: Oh yeah it could. I will have to add more to the todo.rb. 

I do wish it generated an rbi for the http gem but when I run `tapioca gem http` it doesn't do it. @JamieMagee any idea why that doesn't work?

jakecoffman: Well, since I'll be making fake types in todo.rb there's not really a point in typing it.

JamieMagee: I'm not sure about why the http gem isn't generating correctly, but you shouldn't be adding things to the todo file, only removing them.

See:
- https://sorbet.org/docs/rbi#the-todo-rbi-file
- https://sorbet.org/docs/rbi#hand-written-rbis-for-gems

PR #8247, opened 2023-10-20
Correct handling of `updater` directory in sorbet

JamieMagee:
The current configuration had 2 directories listed, `.` and `updater`. However, Sorbet only allows for a single directory. Additionally, we also had `--ignore=updater`. This configuration appeared to be accepted but hid some errors.

Also, a couple of points about `todo.rbi`:
- Things should only ever be removed from `todo.rbi`, never added. Either:
	- They should be migrated to a hand-written `.rbi` in `sorbet/rbi/shims`
	- A `spoom` generated `.rbi` should be added in `sorbet/rbi/gems`
- None of the `Dependabot::` classes or modules should appear in `todo.rbi`
	- This is an indication that those files should be migrated to `# typed: true`

### DIFF
--- a/bundler/lib/dependabot/bundler.rb
+++ b/bundler/lib/dependabot/bundler.rb
@@ -1,4 +1,4 @@
-# typed: true
+# typed: strict
 # frozen_string_literal: true
 
 # These all need to be required so the various classes can be registered in a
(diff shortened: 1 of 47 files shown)


JamieMagee:
Also subsequently ran:
- `spoom bump --from=false --to=true`
- `spoom bump --from=true --to=strict`
- `spoom bump --from=strict --to=strong`

To automatically update the type strength where possible, without any code changes.

jakecoffman (approved):
Nice catch, I guess I glossed over that `ignore`! 

I wonder if we should put together a `script/types` that runs a series of idempotent typing related commands to make it easier to keep things in shipshape?

PR #8261, opened 2023-10-23
add types to some of the Dependabot::Config classes

jakecoffman:


common/lib/dependabot/config/file.rb
```diff
@@ -69,6 +73,10 @@ def ignore_conditions(cfg)
         end
       end
 
+      sig do
+        params(cfg: T.nilable(T::Hash[Symbol,
+                                      T.untyped])).returns(Dependabot::Config::UpdateConfig::CommitMessageOptions)
```
jurre, line 78: Do we need to specify the full namespace like this, or can we rely on the fact that we're already in `Dependabot::Config` and omit those? I'm asking because then we can make this all fit on one line and the sig becomes slightly less ugly 💅 

Probably overly pedantic, feel free to ignore

jurre (approved):

common/lib/dependabot/config/file.rb
JamieMagee: Relative namespaces are totally fine. I usually use fully-qualified namespaces, but that's just a personal preference.

PR #8264, opened 2023-10-23
fix individual PRs being created that should be in the group

### LINKED ISSUE #8234: Individual PRs incorrectly opened (and then immediately closed) when grouped updates enabled

### Package ecosystem

Cargo

### Package manager version

1.73

### Language version

1.73

https://github.com/heroku/buildpacks-python/blob/76e1c6f822dfa9567ad0072b9e082547891c35e8/Cargo.toml
https://github.com/heroku/buildpacks-python/blob/76e1c6f822dfa9567ad0072b9e082547891c35e8/Cargo.lock

### dependabot.yml content

https://github.com/heroku/buildpacks-python/blob/76e1c6f822dfa9567ad0072b9e082547891c35e8/.github/dependabot.yml

### Updated dependency

_No response_

I have grouped updates enabled for this repo, so I expect single-dependency PRs not to be opened unless they are for dependencies not included in the group (or say for other edge cases, such as it being an out-of-scheduled security PR etc).

Prior to this issue occurring, there were two Dependabot PRs open in the repo:
- https://github.com/heroku/buildpacks-python/pull/118 (a grouped PR)
- https://github.com/heroku/buildpacks-python/pull/124 (an out-of-schedule security update PR)

At this point I then triggered a manual Dependabot run (thinking "well if we need to do the security update PR, I might as well update all the things at once"), using Insights -> Dependency graph -> Dependabot -> etc.

The new Dependabot run then proceeded to make two new single-dependency PR for dependencies that should be in a group:
https://github.com/heroku/buildpacks-python/pull/125
https://github.com/heroku/buildpacks-python/pull/126

Immediately after that, Dependabot created the grouped PR after all
https://github.com/heroku/buildpacks-python/pull/127

...and then proceeded to close one of the erroneously opened single update PRs, but not the other?

It seems like there is some race condition or other weirdness going on here.

_No response_

Dependabot run logs:
https://github.com/heroku/buildpacks-python/network/updates/737488004

_No response_

jakecoffman:
fixes #8234

I think the situation is:

1. A group PR exists with a subset of dependencies
2. During the next scheduled run, Dependabot kicks off a group rebase and a normal scheduled job
3. The scheduled job adds all of the dependencies in the existing PR to the handled group
   - It falls back into the individual PR logic and begins processing dependencies that should be in the group, but happen to not be at this time because there were no updates when the job that created it ran
   - One of the dependencies that isn't in the group yet but should be has an update, so an individual PR is created for it
4. The rebase job creates a PR which supersedes the previous grouped PR because it added the new dependency to it
5. The individual PR also gets superseded because a newer PR was created that contained a superset of its dependencies (1 in this case)

The complication here is if `update-types` is used, we are assuming it belongs in this group. Otherwise we'd have to create a checker and see if it should belong in it based on the versions available at the time the job is run.


edmorley:
Hi! #8234 is happening to us quite a bit - I don't suppose you know when you might have a chance to revisit this PR? :-)

deivid-rodriguez (approved):

PR #8266, opened 2023-10-24
Set Python to always clone

deivid-rodriguez:
While working on Python issues, I noticed that it's quite common for native package manager commands to fail because of missing files.

Let's make sure we always have everything we need.


jakecoffman (approved):

honeyankit:
Just queries, will this change increase the job time for python packages?

deivid-rodriguez:
Not necessarily, I think it depends on the repo. Sometimes manually fetching through http a list of specific files, one by one, may take longer than cloning the full repo.

abdulapopoola (approved):

PR #8267, opened 2023-10-24
Unlock related Gemfile dependencies, but not everything that changed

### LINKED ISSUE #8195: Update of sentry gems tries to update git sources

### Package ecosystem

Bundler

### Package manager version

2.4.10

### Language version

3.2.2

https://github.com/robbevp/dependabot-bundler-sentry-reproduction/blob/main/Gemfile

### dependabot.yml content

https://github.com/robbevp/dependabot-bundler-sentry-reproduction/blob/main/.github/dependabot.yml

### Updated dependency

* sentry-rails: 5.10 -> 5.12
* sentry-ruby: 5.10 -> 5.12

(I've also seen this happen for other versions of the sentry gems)

When updating the sentry gems (which happens together, regardless of whether a group was created), dependabot will also update the gems which have git sources. These other dependencies should not be touched or mentioned.

If the git source can also be updated, dependabot will update it (see mocha in my reproduction).
If the git source is up to date, dependabot will still mention it in the PR title and description.

I've noticed this problem in my actual projects (with a lot more dependencies) and:
* It consistently happens for the sentry gems
* Update the sentry gems will pull in all gems with a git source

_No response_

https://github.com/robbevp/dependabot-bundler-sentry-reproduction/pull/1
https://github.com/robbevp/dependabot-bundler-sentry-reproduction/pull/2

See this repo for the smallest possible reproduction: https://github.com/robbevp/dependabot-bundler-sentry-reproduction

deivid-rodriguez:
#7621 fixed an issue where some dependencies were not getting requirement updates in the Gemfile in situations where users would usually expect them.

However, the fix was too aggressive and it caused also unrelated dependencies to the update being unlocked under some situations.

This PR implements a hopefully safer approach.

Fixes #8195.

### DIFF
--- a/bundler/helpers/v2/lib/functions/force_updater.rb
+++ b/bundler/helpers/v2/lib/functions/force_updater.rb
@@ -21,13 +21,21 @@ module Functions
         definition = build_definition(dependencies_to_unlock: dependencies_to_unlock)
         definition.resolve_remotely!
         specs = definition.resolve
-        updates = ([dependency_name, *dependencies_to_unlock] - subdependencies).uniq.map { |name| { name: name } }
+        updates = ([dependency_name, *dependencies_to_unlock] - subdependencies + extra_top_level_deps(specs)).uniq
+
+        updates = updates.map do |name|
+          {
+            name: name
+          }
+        end
+
         specs = specs.map do |dep|
           {
             name: dep.name,
             version: dep.version
           }
         end
+
         [updates, specs]
       rescue Bundler::SolveFailure => e
         raise unless update_multiple_dependencies?
@@ -52,6 +60,15 @@ module Functions
                 :lockfile_name, :credentials,
                 :update_multiple_dependencies
     alias update_multiple_dependencies? update_multiple_dependencies
+
+    def extra_top_level_deps(specs)
+      top_level_dep_names.reject do |name|
+        original_version = original_specs.find { |s| s.name == name }&.version
+        new_version = specs[name].first&.version
+
+        original_version == new_version
+      end
+    end
 
     def new_dependencies_to_unlock_from(error:, already_unlocked:)
       names = [*already_unlocked, dependency_name]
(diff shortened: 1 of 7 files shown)


bundler/helpers/v2/lib/functions/force_updater.rb
```diff
@@ -53,6 +62,28 @@ def run
...
+      top_level_deps.each do |dep|
+        name = dep.name
+
+        original_version = original_specs.find { |s| s.name == name }&.version
+        new_version = specs[name].first&.version
+        next unless original_version != new_version
+
+        dependency_of_updated_dep = updates.any? do |updated_dep|
+          new_spec = specs[updated_dep].first
+          next false unless new_spec
```
Nishnha, line 75: How does a dependency end up in the updates list if it's not in the spec? Would this be a peer dependency, for example?

bundler/helpers/v2/lib/functions/force_updater.rb
```diff
@@ -53,6 +62,28 @@ def run
...
+
+          new_spec.dependencies.any? { |d| d.name == name }
+        end
+        next unless dependency_of_updated_dep
```
Nishnha, line 79:
```suggestion
        next unless updates.any? do |updated_dep|
          new_spec = specs[updated_dep].first
          next false unless new_spec

          new_spec.dependencies.any? { |d| d.name == name }
        end
```

bundler/helpers/v2/lib/functions/force_updater.rb
deivid-rodriguez: Nice catch, I don't think that can happen. One case where a transitve dependency would no longer be in the resolved array of specs is when it's removed, but since here we know it's a dependency that got explictly updated, I don't think it can happen. Let me simplify this.

bundler/helpers/v2/lib/functions/force_updater.rb
Nishnha: The name confused me more than the code itself xD
It's probably fine either way really

deivid-rodriguez: Haha, well if it didn't have the intended effect, let's kill it! 😄

bundler/helpers/v2/lib/functions/force_updater.rb
deivid-rodriguez: I pushed a simpler alternative that I believe should also be more correct. Technically we should be checking all matching specs, not just the first (for the same name and version, there could be different gems resolved for the different platforms, and they could even have different dependencies).

jakecoffman (approved):
There are a number of grouped updates issues in the Bundler ecosystem this may fix as well.

Nishnha (approved):
This makes sense to me!

deivid-rodriguez:
During testing I found that this approach was creating duplicated PRs sometimes. I changed the approach to one that's simpler and does not have this issue, namely, if any other top level dependencies changed versions during "force update", then add those too to the updated_dependencies array.

jurre (approved):

PR #8274, opened 2023-10-25
Types for Dependabot::Updater::Operations

jurre:
~This defines an Interface that declares what these operations should implement, avoiding us having to have them inherit from a base class.~

~Having a base-class does not seem that bad either though, so I'm also happy to go that route.~

Seems like Sorbet doesn't have that type of interfaces, all the classes have to inherit from it anyway, so that's what I did. Bit of a bummer but it works


jakecoffman:
Looks like RefreshVersionUpdatePullRequest needs to inherit from Operation too.

PR #8275, opened 2023-10-25
build(deps): bump node to v20

yeikel:
Release notes : 

https://nodejs.org/en/blog/announcements/v19-release-announce
https://nodejs.org/en/blog/announcements/v20-release-announce

### DIFF
--- a/npm_and_yarn/Dockerfile
+++ b/npm_and_yarn/Dockerfile
@@ -10,7 +10,7 @@ ARG PNPM_VERSION=8.15.2
 ARG YARN_VERSION=3.7.0
 
 # See https://github.com/nodesource/distributions#installation-instructions
-ARG NODEJS_VERSION=18.x
+ARG NODEJS_VERSION=20
 
 # Check for updates at https://github.com/npm/cli/releases
 # This version should be compatible with the Node.js version declared above. See https://nodejs.org/en/download/releases as well
@@ -20,7 +20,7 @@ ARG NPM_VERSION=9.6.5
 # Install Node and npm
 RUN mkdir -p /etc/apt/keyrings \
   && curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg \
-  && echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_$NODEJS_VERSION nodistro main" | tee /etc/apt/sources.list.d/nodesource.list \
+  && echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_${NODEJS_VERSION}.x nodistro main" | tee /etc/apt/sources.list.d/nodesource.list \
   && apt-get update \
   && apt-get install -y --no-install-recommends \
   nodejs \


yeikel:
@deivid-rodriguez What do you think about this change? 

We're still pinning the npm and yarn versions so it should be fine?

npm_and_yarn/Dockerfile
```diff
@@ -7,7 +7,7 @@ ARG PNPM_VERSION=8.9.2
 ARG YARN_VERSION=3.6.4
 
 # See https://github.com/nodesource/distributions#installation-instructions
-ARG NODEJS_VERSION=18.x
+ARG NODEJS_VERSION=20.x
```
JamieMagee, line 10: unrelated, but what do you think about moving the `.x` from `NODEJS_VERSION` to line 20 so it's `node_${NODEJS_VERSION}.x`

yeikel: That sounds reasonable, thanks!

yeikel:
> Currently we run a fixed version on nodejs, so that means we are not able to support applications with certain nodejs constraints. For example, if someone specifying "node": "^18.12.0" in their engines package.json section, my understanding is that this PR will break that

As far as I was able to test, it does not seem that NPM enforces this by default

Dependencies can also define what engines they support that can lead to the following warning at installation time  : 

> EBADENGINE Unsupported engine {
> npm WARN EBADENGINE   package: 'pretty-format@28.1.3',
> npm WARN EBADENGINE   required: { node: '^12.13.0 || ^14.15.0 || ^16.10.0 || >=17.0.0' },
> npm WARN EBADENGINE   current: { node: 'v16.1.0', npm: '8.19.2' }
> npm WARN EBADENGINE }

By default this is not enabled but it can be can be enforced via `.npmrc` setting `engine-strict=true which will lead to the following failure at installation time : 

```
npm ERR! code EBADENGINE
npm ERR! engine Unsupported engine
npm ERR! engine Not compatible with your version of node/npm: @jest/console@28.1.3
npm ERR! notsup Not compatible with your version of node/npm: @jest/console@28.1.3
npm ERR! notsup Required: {"node":"^12.13.0 || ^14.15.0 || ^16.10.0 || >=17.0.0"}
npm ERR! notsup Actual:   {"npm":"8.19.2","node":"v16.1.0"}
```

As far as I can tell, we are catching that already when the user enforces it : https://github.com/dependabot/dependabot-core/blob/db9cb4b21a34576e1bc9588ead3b2801018679d2/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npm_lockfile_updater.rb#L422-L426 


> ideally we would be able to use the most appropriate nodejs version for each repo, or even better, package manager would provide a way to ask "please resolve for nodejs version X, regardless of the running node"

I am not sure what you're thinking here, but perhaps we could include more than one version in the image with `nvm` and switch accordingly?

deivid-rodriguez (approved):

yeikel:
@deivid-rodriguez What changed that you decided to approve it? I am just curious as your points were valid and interesting 

Thank you!

deivid-rodriguez:
Going a bit back and forth with this, but the intention is to deploy this soon-ish, because the potential issues seem quite minor, and not deploying this implies having similar issues for users that require the latest NodeJS version.

I'm still working on going through some common related errors unknown errors and raising them as informative user errors, and I'll probably deploy this once that's done.

yeikel:
@abdulapopoola @jakecoffman Could you please help me to assign a point of contact now that David lost the necessary access to check the metrics (or that's what I understand) ?

yeikel:
@deivid-rodriguez Do you remember where you left off? Thanks!

deivid-rodriguez:
Hei!

I run out of time for working on better error handling for NPM and had to switch to other tasks. So my idea was to ship this as is.

Given your explanation in https://github.com/dependabot/dependabot-core/pull/8275#issuecomment-1785369140 that this is by default a warning and not an error, I think it's fine to move forward with the upgrade. We should not be penalizing users of modern nodejs versions in favor of users of old nodejs versions.

yeikel:
> Hei!
> 
> I run out of time for working on better error handling for NPM and had to switch to other tasks. So my idea was to ship this as is.
> 
> Given your explanation in [#8275 (comment)](https://github.com/dependabot/dependabot-core/pull/8275#issuecomment-1785369140) that this is by default a warning and not an error, I think it's fine to move forward with the upgrade. We should not be penalizing users of modern nodejs versions in favor of users of old nodejs versions.

Awesome, thank you for taking the time to explain. 

Do you recall what was missing in terms of error handling? If you have time at some point in the near future, I'd be nice to log them as tech deb issues 🙏.

 In any case, as you pointed out, this change should be ready to be merged as is. Thank you for taking the time to reply so quickly! 🙇‍♂️ 

@abdulapopoola Any chance you can help move this one forward?

deivid-rodriguez:
I didn't have time to materialize it into a proper issue, but I suspected from analyzing error messages logged in our internal tracker that sometimes we were not properly handling EBADENGINE errors. Because otherwise they should be showing up as user errors and not as unknown errors in our internal tracker.

I think it's fine to wait until someone reports any issues.

yeikel:
@jurre This one is ready as well

basil:
Possibly related to #8896?

xu3u4:
Any plan to merge this PR? 🙏

yeikel:
@jurre Can you please review/help to move this one forward?

smoookeeey:
> @jurre Can you please review/help to move this one forward? 
- 
Approved

jurre (approved):

jurre:
Thanks @yeikel! I've let this sit in prod for an hour and not observing any issues so far but we'll keep an eye out

yeikel:
> Thanks @yeikel! I've let this sit in prod for an hour and not observing any issues so far but we'll keep an eye out

Thank you! I think that the issues should be minimal because we're unfortunately still pinning to NPM 9.x

xu3u4:
Thank you all 🙇‍♀️🙇‍♀️🙇‍♀️

PR #8279, opened 2023-10-26
Avoid incorrectly downgrading top level deps

deivid-rodriguez:
If we unlock other top level dependencies when trying to update a dependency, that kind of auto-grouping is fine.

However, we should make sure we don't downgrade a top level dependency just to upgrade another.


bundler/spec/dependabot/bundler/update_checker/force_updater_spec.rb
```diff
@@ -306,5 +306,24 @@
...
+          file: "Gemfile",
+          requirement: "~> 7.1",
+          groups: [:default],
+          source: nil
+        }]
+      end
+
+      it "raises a resolvability error" do
+        expect { updater.updated_dependencies }
+          .to raise_error(Dependabot::DependencyFileNotResolvable)
```
jurre, line 325: Wondering if it makes sense to _raise_ in this scenario, wouldn't it be better to simply not perform the update and say we're at the latest possible version? I thought that's what we do in most of these sort of scenario's?

deivid-rodriguez: Yep, that's how it works.

I assert for that error since I'm unit testing the force updater, but that error is gracefully handled here:

https://github.com/dependabot/dependabot-core/blob/b39f6b0eeecf8140943e924e93838f65835f7572/bundler/lib/dependabot/bundler/update_checker.rb#L150-L170

jurre: Ah thanks for clarifying!

jurre (approved):

PR #8286, opened 2023-10-27
Simplify docker regex

deivid-rodriguez:
This is getting out of hand, trying to keep it a bit simpler.


yeikel:
The CI passes, so I guess that it is fine? These regexps are always a hassle to maintain and review

deivid-rodriguez:
Turns out the `kb-` part of the regex is currently necessary (unfortunately) because we currently special case `KB-` in `mcr.microsoft.com/windows/servercore` images. The specs in in #8442 will make these changes fail specs. Setting as WIP.

deivid-rodriguez:
I removed the `-kb` simplification from this PR. Back to ready.

jeffwidman (approved):

PR #8288, opened 2023-10-27
Revert python always clone registration

deivid-rodriguez:
The right key was actually "pip", not "python".

Enabling this was silently a noop. Properly enabling it requires more work, so revert for now.

In addition, I added some changes to make sure this doesn't happen again. I guess these changes could potentially get in the middle for people extending dependabot-core to add custom ecosystems, but I don't think anybody is reliably doing this, and if they are, they can probably add one more patch to workaround this. For now, I'd prioritize our own convenience.

### DIFF
--- a/common/lib/dependabot/utils.rb
+++ b/common/lib/dependabot/utils.rb
@@ -5,6 +5,7 @@ require "tmpdir"
 require "set"
 require "sorbet-runtime"
 require "dependabot/version"
+require "dependabot/config/file"
 
 # TODO: in due course, these "registries" should live in a wrapper gem, not
 #       dependabot-core.
@@ -22,11 +23,13 @@ module Dependabot
       version_class = @version_classes[package_manager]
       return version_class if version_class
 
-      raise "Unsupported package_manager #{package_manager}"
+      raise "Unregistered package_manager #{package_manager}"
     end
 
     sig { params(package_manager: String, version_class: T.class_of(Dependabot::Version)).void }
     def self.register_version_class(package_manager, version_class)
+      validate_package_manager!(package_manager)
+
       @version_classes[package_manager] = version_class
     end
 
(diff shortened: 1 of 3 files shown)


mctofu (approved):

common/lib/dependabot/utils.rb
```diff
@@ -54,7 +59,21 @@ def self.always_clone_for_package_manager?(package_manager)
...
     def self.register_always_clone(package_manager)
+      validate_package_manager!(package_manager)
+
       @cloning_package_managers << package_manager
     end
+
+    sig { params(package_manager: String).void }
+    def self.validate_package_manager!(package_manager)
+      # Official package manager
+      return if Config::File::PACKAGE_MANAGER_LOOKUP.invert.key?(package_manager)
```
JamieMagee, line 70: The Sorbet error is because this isn't typed:

https://github.com/dependabot/dependabot-core/blob/ba47d42101b843867e05f4c24d6e45e15b1a3076/common/lib/dependabot/config/file.rb#L49-L67

The fix is to wrap it in `T.let(..., T::Hash[String, String])`

deivid-rodriguez: Thank you!

JamieMagee: You're welcome! But I should have been more explicit. It should be:

```rb
PACKAGE_MANAGER_LOOKUP = T.let({
        "bundler" => "bundler",
        "cargo" => "cargo",
        "composer" => "composer",
        "docker" => "docker",
        "elm" => "elm",
        "github-actions" => "github_actions",
        "gitsubmodule" => "submodules",
        "gomod" => "go_modules",
        "gradle" => "gradle",
        "maven" => "maven",
        "mix" => "hex",
        "nuget" => "nuget",
        "npm" => "npm_and_yarn",
        "pip" => "pip",
        "pub" => "pub",
        "swift" => "swift",
        "terraform" => "terraform"
      }.freeze, T::Hash[String, String])
```

PR #8299, opened 2023-10-30
Composer: update expected plugin API version

fredden:
`plugin-api-version` was changed from "2.3.0" to "2.6.0" in Composer version 2.6.0. https://github.com/composer/composer/commit/16bdfe4dae0d6ef151d71441849e735f3b0a93c8

This pull request fixes tests to allow upgrading Composer to 2.6.0+. I've targeted https://github.com/dependabot/dependabot-core/pull/8159 here, but the base for this can be easily changed to the default branch if that's preferred.

### DIFF
--- a/composer/Dockerfile
+++ b/composer/Dockerfile
@@ -1,6 +1,6 @@
 FROM ghcr.io/dependabot/dependabot-updater-core
 ARG COMPOSER_V1_VERSION=1.10.26
-ARG COMPOSER_V2_VERSION=2.5.8
+ARG COMPOSER_V2_VERSION=2.6.5
 ENV COMPOSER_ALLOW_SUPERUSER=1
 RUN apt-get update \
   && apt-get upgrade -y \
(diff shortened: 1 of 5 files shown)


fredden:
There seems to be a circular dependency here, where this pull request and https://github.com/dependabot/smoke-tests/pull/133 need to be merged simultaneously in order for the tests to pass. How do we link them together?

Nishnha:
> There seems to be a circular dependency here, where this pull request and [dependabot/smoke-tests#133](https://github.com/dependabot/smoke-tests/pull/133) need to be merged simultaneously in order for the tests to pass. How do we link them together?

Thanks for the smoke-tests PR. I've merged that and re-ran the e2e test for composer on this PR so now it's passing CI! 
I'm going keep the base as-is so it merges into https://github.com/dependabot/dependabot-core/pull/8159. Then I'll rebase that entire PR and its CI should be passing thanks to you 😃

PR #8300, opened 2023-10-30
Upgrade Composer v1 to 1.10.27

fredden:
For some reason, dependabot did not complete the Composer v1 upgrade in https://github.com/dependabot/dependabot-core/pull/8114. This pull request completes that action, which should also make the tests pass there.


deivid-rodriguez:
Thank you!

PR #8302, opened 2023-10-30
Type `SharedHelpers`

JamieMagee:
This only brings `SharedHelpers` to `typed: true` because the changes here are different than the rest of the PRs I've been opening recently. There will be a lot more work to get `SharedHelpers` to `strict` and eventually `strong`.

### DIFF
--- a/common/lib/dependabot/shared_helpers.rb
+++ b/common/lib/dependabot/shared_helpers.rb
@@ -219,7 +219,6 @@ module Dependabot
       File.join(__dir__, "../../bin/git-credential-store-immutable")
     end
 
-    # rubocop:disable Metrics/AbcSize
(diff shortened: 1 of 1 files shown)


landongrindheim (approved):

common/lib/dependabot/shared_helpers.rb
```diff
@@ -82,6 +85,7 @@ def self.escape_command(command)
     end
 
     # rubocop:disable Metrics/MethodLength
+    # rubocop:disable Metrics/AbcSize
```
landongrindheim, line 88: Looks like this disablement [isn't necessary](https://github.com/dependabot/dependabot-core/actions/runs/6710089882/job/18234551734?pr=8302#step:4:12) 😎

Issue #8303, opened 2023-10-30
Unify cloning as the fetch strategy for all ecosystems


deivid-rodriguez:
### Code improvement description

We've come to the conclusion that cloning full repositories it's probably the most resilient approach to file fetching, since we make sure no files will be missing when shelling out to package managers.

We should eventually switch all ecosystems to clone full repositories.

This is a tech-debt ticket to track that.

For a given ecosystem, this is a non exhaustive list of things to be done:

* Enable full cloning through `Dependabot::Utils.register_always_clone(<package_ecosystem>)`.
* Make sure native package manager commands use the full clone. This will usually involve passing around `repo_contents_path` to the different classes and switching `SharedHelpers.in_temporary_repo` calls to `SharedHelpers.in_temporary_repo_directory` that also takes the path to the clone instead of a fresh temporary folder.

abdulapopoola:
This has been fixed thanks to @jakecoffman !!

PR #8309, opened 2023-10-31
Teach Updater to Invoke fetch files Based on Multi-Dir

honeyankit:
## Context
This pull request introduces enhancements to the FileFetcherCommand class, ensuring file fetching when dealing with repositories containing multiple directories. 

## Key Changes

Multi-Directory Support: The code now supports fetching files from multiple directories in a repository within same ecosystem


honeyankit:
My test is python, since all the tests are in updater are in bundler so changing my test to Bundler.

updater/lib/dependabot/file_fetcher_command.rb
```diff
@@ -64,6 +72,23 @@ def save_job_details
...
+          credentials: Environment.job_definition.fetch("credentials", []),
+          options: job.experiments
+        )
+        ff.files
+      end
+      all_dependency_files.flatten
+    rescue Octokit::BadGateway
+      @file_fetcher_retries ||= 0
+      @file_fetcher_retries += 1
+      @file_fetcher_retries <= 2 ? retry : raise
```
jakecoffman, line 89: Does the rescue need to go inside the map? It looks like if it does throw BadGateway it will retry from the start.

updater/lib/dependabot/file_fetcher_command.rb
```diff
@@ -64,6 +72,23 @@ def save_job_details
                                        ))
     end
 
+    def dependency_files_for_multi_directories
```
jakecoffman, line 75: It looks like the result of this method needs to be memoized since it gets called twice.

In `dependency_files` this happened because we used the same `file_fetcher` twice and the `files` call was memoized. Now since we're making a file_fetcher on demand we should memoize the result. It looks like in the smoke tests it gets all the files twice, memoizing should fix that.

jakecoffman: Another option is to save the result from the call and parameterize `base64_dependency_files`, up to you how to fix it!

honeyankit: Good catch! I will add the memoized to the `all_dependency_files` result

updater/lib/dependabot/file_fetcher_command.rb
honeyankit: Thanks for the review, I will adjust the `dependency_files_for_multi_directories` method to retry fetching only the directory that caused the BadGateway error, rather than restarting from the beginning

jakecoffman (approved):
LGTM

updater/lib/dependabot/file_fetcher_command.rb
```diff
@@ -64,6 +72,25 @@ def save_job_details
...
+            source: updated_dir_source,
+            credentials: Environment.job_definition.fetch("credentials", []),
+            options: job.experiments
+          )
+          ff.files
+        rescue Octokit::BadGateway => e
+          retries += 1
+          retry if retries <= 2
+          raise e
+        end
```
deivid-rodriguez, line 90: Sounds like retry logic is duplicated with `dependency_files`. Perhaps you can extract a `with_retries` helper that yields to the block it receives, and rescues and retries on BadGateway errors.

updater/lib/dependabot/file_fetcher_command.rb
```diff
@@ -64,6 +72,25 @@ def save_job_details
                                        ))
     end
 
+    def dependency_files_for_multi_directories
+      @dependency_files_for_multi_directories ||= job.source.directories.map do |dir|
+        retries = 0
+        begin
+          updated_dir_source = job.source.clone.tap { |s| s.directory = dir }
+          ff = FileFetchers.for_package_manager(job.package_manager).new(
```
deivid-rodriguez, line 80: If this is the same ecosystem, could we instead do `file_fetcher.class.new`?

honeyankit: The `FileFetchers.for_package_manager(job.package_manager)` is being called for the first time within `dependency_files_for_multi_directories` and so we do not have an existing `file_fetcher` object for that package manager,  so we will need to instantiate a new file_fetcher for each directory.

deivid-rodriguez: I see, do these fetchers need the `repo_contents_path` option like the main fetcher sets?

If that's the case, I would probably reuse all instantiation logic and keep a `@file_fetchers` instance variable as a hash of fetchers per directory.

updater/lib/dependabot/file_fetcher_command.rb
```diff
@@ -74,29 +74,33 @@ def save_job_details
...
+    end
+
+    def with_retries(max_retries: 2)
+      retries ||= 0
+      begin
+        yield
+      rescue Octokit::BadGateway => e
+        retries += 1
+        retry if retries <= max_retries
+        raise e
```
deivid-rodriguez, line 102: I prefer `raise` to `raise e`. The distinction is quite subtle, `raise` let the original exception be raised, while `raise e` creates a new exception that's a duplicate of the original.

Behavior is essentially the same but you get a less complicated backtrace with `raise`, since with `raise e` you'll get a full backtrace from this location, and then another one from the original exception's location as the exception's cause.

honeyankit: Thanks this make more sense. I will make the changes.

updater/lib/dependabot/file_fetcher_command.rb
honeyankit: Putting my discussion from Deivid for context here:

[Ankit]: I am able to get the correct  dependency_files  for all the directories even without passing the repo_contents_path as an argument in dependency_files_for_multi_directories method like we do in dependency_files  method. If I understand your concern correctly, we should have exact same way to instantiate the file-fetcher  irrespective of the whether it is being used in dependency_files_for_multi_directories  or dependency_files  method (this make sense to me as well) ? (edited) 


[Deivid]: Yes, you're able to all dependency files. The problem, I believe, is that if you don't pass around the repo_contents_path, then shelling out to package manager in subsequent steps (update checker, file updater) will happen from a temporary empty directory where dependency_files will be written. If repo_contents_path is passed around, package managers run in the folder of the repository clone. That's generally more resilient since we make sure we never miss any files that are not proper dependency_files, but that package managers need for context. See https://github.com/dependabot/dependabot-core/issues/8303 for more context on this.

honeyankit: @deivid-rodriguez: I took above suggestions, now section of code is blowing up. In python ecosystem, It was just using `dependency_files` output as input so I passed it as a parameter `file_fetcher.ecosystem_versions(dfiles)` but I was wrong for other ecosystems, the implementation are different.

```ruby
        # We don't set this flag in GHES because there's no point in recording versions since we can't access that data.
        if Experiments.enabled?(:record_ecosystem_versions)
          ecosystem_versions = file_fetcher.ecosystem_versions
          api_client.record_ecosystem_versions(ecosystem_versions) unless ecosystem_versions.nil?
        end
```

honeyankit: This issue is fixed now. @deivid-rodriguez feel free to review it again!

deivid-rodriguez (approved):
Looks good to me!

jakecoffman:
Two of the smoke tests ([go-security](https://github.com/dependabot/smoke-tests/blob/main/tests/smoke-go-security.yaml), and [go-update-pr](https://github.com/dependabot/smoke-tests/blob/main/tests/smoke-go-update-pr.yaml)) are now getting `mode` added which wasn't there before. I guess that isn't a problem, but just curious why that is.

jakecoffman:
Looking into the `mode` changes some more, on main it gets into [this rescue](https://github.com/dependabot/dependabot-core/blob/be96f0fb12c41a3a08ad2b5b466e0fa4562e93ad/common/lib/dependabot/dependency_file.rb#L55C1-L55C1) because the `go.mod` is not found, but the `go.sum` is. 

This appears to be because `Dir.pwd` is `"/home/dependabot/dependabot-updater"` for the `go.mod` fetch but `"/home/dependabot/dependabot-updater/repo"` for the `go.sum` fetch. 

The new code must have reordered some things so that the `go.mod` is checked in the right pwd. But this seems like a bug in the Go FileFetcher.

I'm good with proceeding with this and fixing the tests as it seems more correct now!

PR #8312, opened 2023-11-01
Replace `pipenv lock` with `pipenv upgrade`

### LINKED ISSUE #6836: Switch to leveraging `pipenv`'s new `pipenv upgrade` command
> I have another update for dependabot team though -- which is in the `main` branch of `pipenv` is a a new `pipenv upgrade` command that should be perfect for dependabot's use case of upgrading the Pipfile/Pipfile.lock (doesn't require package install) -- and doesn't affect the whole set of lock file dependencies.  
> 
> Since `--selective-upgrade` and `--keep-outdated` are deprecated for removal, it would be useful to try updating dependabot to use this new command (it will be in the next release, maybe in a week but you can try it out from the main branch today).
> 
> If you are experiencing issues related to the upgrade path, feel free to open new issue reports and we will prioritize looking into them.   Thank you!

_Originally posted by @matteius in https://github.com/pypa/pipenv/issues/2791#issuecomment-1468694507_

deivid-rodriguez:
The `pipenv lock` command had several problems:

* It upgrades all versions to the latest version, not just the
  dependency being upgraded. This required us to explicitly pin all top
  level dependencies to prevent them from being upgraded, and even so,
  we'd still give incorrect results sometimes, when unrelated transitive
  dependencies would also get unintendedly upgraded.

* It not only resolved versions, but also installed packags, with the
  potential problems like missing OS dependencies associated with that.

Closes #6836.


deivid-rodriguez:
This smoke test failure is a bug/missing feature in pipenv: https://github.com/pypa/pipenv/issues/6002.

PR #8314, opened 2023-11-01
Add types for `ExceptionSanitizer`

JamieMagee:


updater/lib/dependabot/sentry.rb
```diff
@@ -1,23 +1,33 @@
...
+    T::Hash[Symbol, Regexp]
+  )
 
+  sig do
+    params(data: T::Hash[Symbol, T.nilable(T::Hash[Symbol, T::Array[T::Hash[Symbol, String]]])])
+      .returns(T::Hash[Symbol, T.untyped])
+  end
   def process(data)
-    return data unless data[:exception] && data[:exception][:values]
+    return data unless data[:exception]&.[](:values)
```
jakecoffman, line 25: First time I've seen this syntax, might be more clear with `dig`.

```suggestion
    return data unless data.dig(:exception, :values)
```

jakecoffman (approved):

PR #8318, opened 2023-11-02
Add types for `Environment`

JamieMagee:
The types for the boolean `def`s is a little funky (`github_actions` and `deterministic_updates`) but it's a good start, and works for now.


updater/lib/dependabot/environment.rb
```diff
@@ -1,64 +1,93 @@
...
+      type_parameters(:T)
+        .params(variable_name: String, default: T.nilable(T.type_parameter(:T)))
+        .returns(T.any(String, T.type_parameter(:T)))
+    end
+    private_class_method def self.environment_variable(variable_name, default = nil)
+      return ENV.fetch(variable_name, default) if default
 
       ENV.fetch(variable_name) do
         raise ArgumentError, "Missing environment variable #{variable_name}"
       end
```
JamieMagee, line 82: I think the original code, `ENV.fetch(variable_name, default) unless default` only was called if `default` was `nil`. Which means the default value, if present, was never used.

Correct me if I'm wrong, but I think this might be a Sorbet win.

jakecoffman: Since it was `ENV.fetch(variable_name, default) unless default == :_undefined` it's being called when a second parameter isn't being passed at all. A fancy way to enforce some environment variables as required, and giving the others a fallback value.

JamieMagee: Generic types are erased at runtime which makes this difficult. I've refactored it as best I can.

updater/lib/dependabot/environment.rb
```diff
@@ -1,64 +1,96 @@
...
     def self.job_definition
-      @job_definition ||= JSON.parse(File.read(job_path))
+      @job_definition ||= T.let(JSON.parse(File.read(job_path)), T.nilable(T::Hash[String, T.untyped]))
     end
 
+    sig do
+      type_parameters(:T)
+        .params(variable_name: String, default: T.any(Symbol, T.type_parameter(:T)))
+        .returns(T.any(String, T.type_parameter(:T)))
+    end
```
jakecoffman, line 76: This is pretty hairy, what do you think about making `default` a String, the default value `"_undefined"` then in the 2 cases we return a Boolean above we do `== "false"`?

JamieMagee: Unfortunately Sorbet won't type narrow with a string of "_undefined". I think I will split out boolean values to a separate "environment_variable_boolean" method, which should simplify things a bit.

I know the type signature is a little complex, but it shouldn't need to be edited. It would be nicer if I could use the `:undefined` literal symbol, [like TypeScript][1], but [Sorbet doesn't support that][2].

[1]: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types
[2]: https://github.com/sorbet/sorbet/issues/2504

updater/lib/dependabot/environment.rb
```diff
@@ -1,64 +1,96 @@
...
     end
 
+    sig { returns(T.any(String, T::Boolean)) }
     def self.deterministic_updates?
-      @deterministic_updates ||= environment_variable("UPDATER_DETERMINISTIC", false)
+      @deterministic_updates ||= T.let(
+        environment_variable("UPDATER_DETERMINISTIC", false),
+        T.nilable(T.any(String, T::Boolean))
+      )
     end
```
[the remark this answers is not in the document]
JamieMagee: I extracted this into an `environment_variable_boolean` method instead.

jakecoffman (dismissed):

JamieMagee:
Closing in favour of #10201

PR #8319, opened 2023-11-02
strictly type `Updater` mixins

JamieMagee:
Because the mixins use properties available in the `Operation` classes, Sorbet needs to know that they exist. I created a common base class that contains the properties required by the mixins, and used Sorbet's [`require_ancestor`][1] to ensure that the properties are available.

[1]: https://sorbet.org/docs/requires-ancestor


JamieMagee:
Closing in favour of #10201

PR #8323, opened 2023-11-02
Properly aggregate `pipenv upgrade` errors in Sentry

deivid-rodriguez:
After #8312, `pipenv upgrade` errors are not being aggregated in Sentry. This commit should fix that.

### DIFF
--- a/python/lib/dependabot/python/pipenv_runner.rb
+++ b/python/lib/dependabot/python/pipenv_runner.rb
@@ -18,7 +18,7 @@ module Dependabot
         command = "pyenv exec pipenv upgrade #{dependency_name}#{constraint}"
         command << " --dev" if lockfile_section == "develop"
 
-        run(command)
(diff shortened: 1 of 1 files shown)


python/lib/dependabot/python/pipenv_runner.rb
```diff
@@ -18,7 +18,7 @@ def run_upgrade(constraint)
         command = "pyenv exec pipenv upgrade #{dependency_name}#{constraint}"
         command << " --dev" if lockfile_section == "develop"
 
-        run(command)
+        run(command, fingerprint: "pyenv exec pipenv upgrade <dependency_name><constraint>")
```
landongrindheim, line 21: 🔎 Is the hardcoded `<dependency_name><constraint>` string what's causing these to be aggregated? i.e. Would we get different traces for every combination if we used `#{dependency_name}#{constraint}`?

landongrindheim (approved):

deivid-rodriguez:
Thanks for your reviews Landon 😍!

PR #8324, opened 2023-11-02
Catch more 404 PNPM fetch errors and raise them as user errors

deivid-rodriguez:
Sometimes errors will have the following shape

```
ERR_PNPM_FETCH_404  GET https://registry.npmjs.org/@tiptap-pro/extension-mathematics/-/extension-mathematics-2.3.3.tgz: Not Found - 404
```

instead of the one we were capturing

```
ERR_PNPM_FETCH_404  GET https://registry.npmjs.org/@dependabot%2Ftotally-fake-dependency-soz: Not Found - 404

This error happened while installing a direct dependency of dependabot_tmp_dir

@dependabot/totally-fake-dependency-soz is not in the npm registry, or you have no permission to fetch it.
```

This PR changes the regex to capture both types of errors.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_parser.rb
@@ -11,6 +11,7 @@ require "dependabot/npm_and_yarn/helpers"
 require "dependabot/npm_and_yarn/native_helpers"
 require "dependabot/npm_and_yarn/version"
 require "dependabot/npm_and_yarn/requirement"
+require "dependabot/npm_and_yarn/registry_parser"
 require "dependabot/git_metadata_fetcher"
 require "dependabot/git_commit_checker"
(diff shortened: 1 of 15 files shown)


honeyankit (approved):
Thanks. Looks good to me. I have a question. what made you think to new register parser   class?

deivid-rodriguez:
Good question!

We already had logic in place to, given a registry url, extract the host.

For this, I needed logic to, given a registry url, extract the dependency.

I thought these two were very related and was hoping to reuse some logic. In the end, I ended up not reusing any existing logic, but I think it still makes sense to keep logic related like this together.

deivid-rodriguez:
Just some heads up, the spec added by this PR takes about 1m 15s to run, so has delayed quite a lot of NPM jobs.

The good news is that once https://github.com/pnpm/pnpm/pull/7276 is released and we upgrade our default PNPM version, then it should start running fast.

PR #8328, opened 2023-11-03
Nishnha/gsu multi dir python

Nishnha:
creates a different group for each directory

I plan on merging the different groups together by making the first directory a `DependencyChange` object and merging it with the next directory's `DependencyGroupChangeBatch` objects and so forth


Nishnha:
Closing out since https://github.com/dependabot/dependabot-core/pull/8331 seems like the better approach.

Issue #8242, opened 2023-10-20
Can't specify which registry pnpm should use


dancallaghan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

pnpm

### Package manager version

8.9.2

### Language version

18.17.1

### dependabot.yml content

Here is a minimal version of our `dependabot.yml`. I can provide more detail if required

```yaml
version: 2
registries:
  npm-github:
    type: npm-registry
    url: https://npm.pkg.github.com
    username: foo@bar.com
    password: ${{secrets.GH_TOKEN}}
updates:
  - package-ecosystem: npm
    directory: /
    open-pull-requests-limit: 10
    registries:
      - npm-github
```

### What you expected to see, versus what you actually saw

We are using `https://npm.pkg.github.com` as a private registry for our scoped private packages (e.g. `@org/logger`). We are running into issues because dependabot is checking `https://npm.pkg.github.com` registry first for _all_ public packages. In most cases this is 404ing and falling back to the global registry `https://registry.npmjs.org`.

There are situations where the public package exists in the GitHub registry, however it has stale information about the package. This is causing issues like this:

```
The latest release of @types/node is "17.0.18".
```

(Spoiler, this is not the latest release of `@types/node` 😅)

Unlike `npm` and `yarn` it looks like there is no way to specify a `locked_registry` with `pnpm` so we can tell dependabot to use a specific registry when resolving a dependency. With a pnpm lockfile, you can [include a tarball URL](https://pnpm.io/npmrc#lockfile-include-tarball-url), however from looking at the code this is not respected, unlike properties like `resolved`.

Not sure if there is another way to achieve this? For example, is there a way to tell dependabot only to use a scoped private registry for scoped dependencies?

### Images of the diff or a link to the PR, issue, or logs

```
proxy | 2023/10/19 21:54:50 [032] GET https://npm.pkg.github.com:443/@sentry%2Fnode
proxy | 2023/10/19 21:54:50 [032] * authenticating npm registry request (host: npm.pkg.github.com, basic auth)
proxy | 2023/10/19 21:54:50 [032] 404 https://npm.pkg.github.com:443/@sentry%2Fnode
proxy | 2023/10/19 21:54:51 [034] GET https://registry.npmjs.org:443/@sentry%2Fnode
proxy | 2023/10/19 21:54:51 [034] 200 https://registry.npmjs.org:443/@sentry%2Fnode
[...]

PR #8330, opened 2023-11-03
Fix PNPM issues with private registries

deivid-rodriguez:
This PR fixes two issues:

* Dependabot can't infer the proper registry for each dependency by looking at registry urls in pnpm-lock.yaml.

This is handy because developers don't need to commit a `.npmrc` file and matches what we already support for NPM and Yarn.

* Dependabot incorrectly tries to infer the proper registry for each dependency by looking at presence in the registry, even when registries & scopes are explicitly configured in `.npmrc`.

This is a good last resort fallback, but it should not be used if registries & scopes have been explicitly configured.

Fixes https://github.com/dependabot/dependabot-core/issues/8242.
Fixes performance issues introduced by https://github.com/dependabot/dependabot-core/pull/8094.
Likely to fix https://github.com/dependabot/dependabot-core/issues/8008.

### DIFF
--- a/npm_and_yarn/helpers/lib/pnpm/lockfile-parser.js
+++ b/npm_and_yarn/helpers/lib/pnpm/lockfile-parser.js
@@ -68,6 +68,7 @@ function nameVerDevFromPkgSnapshot(depPath, pkgSnapshot, projectSnapshots) {
   return {
     name: name,
     version: version,
+    resolved: pkgSnapshot.resolution.tarball,
     dev: pkgSnapshot.dev,
     specifiers: specifiers,
     aliased: aliased
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/registry_finder.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/registry_finder.rb
@@ -31,7 +31,7 @@ module Dependabot
         end
 
         def registry
-          @registry ||= locked_registry || first_registry_with_dependency_details
+          @registry ||= locked_registry || configured_registry || first_registry_with_dependency_details
         end
 
         def auth_headers
@@ -49,15 +49,19 @@ module Dependabot
         end
 
         def registry_from_rc(dependency_name)
-          return global_registry unless dependency_name.start_with?("@") && dependency_name.include?("/")
-
-          scope = dependency_name.split("/").first
-          scoped_registry(scope)
+          explicit_registry_from_rc(dependency_name) || global_registry
         end
 
         private
 
         attr_reader :dependency, :credentials, :npmrc_file, :yarnrc_file, :yarnrc_yml_file
+
+        def explicit_registry_from_rc(dependency_name)
+          return configured_global_registry unless dependency_name.start_with?("@") && dependency_name.include?("/")
+
+          scope = dependency_name.split("/").first
+          scoped_registry(scope)
+        end
 
         def first_registry_with_dependency_details
           @first_registry_with_dependency_details ||=
@@ -126,6 +130,13 @@ module Dependabot
           detailed_registry || lockfile_registry
         end
 
+        def configured_registry
+          configured_registry_url = explicit_registry_from_rc(dependency.name)
+          return unless configured_registry_url
+
+          normalize_configured_registry(configured_registry_url)
+        end
+
         def known_registries
           @known_registries ||=
             begin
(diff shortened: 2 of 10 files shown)


yeikel:
Thank you for tagging me! 

I am not too familiar with the ecosystem but I understand the problem. Any chance you can add a spec?

deivid-rodriguez:
Actually I meant to ping @mburumaxwell 🤦‍♂️, sorry about that. Thanks for chiming in though! Yes, this is a draft because I plan to add some specs 👍.

deivid-rodriguez:
@dancallaghan This is now ready, I'll deploy it next week!

jakecoffman (approved):

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/registry_finder.rb
```diff
@@ -49,16 +49,22 @@ def self.central_registry?(registry)
         end
 
         def registry_from_rc(dependency_name)
-          return global_registry unless dependency_name.start_with?("@") && dependency_name.include?("/")
-
-          scope = dependency_name.split("/").first
-          scoped_registry(scope)
+          explicit_registry_from_rc(dependency_name) || global_registry
```
deivid-rodriguez, line 52: I did this for backwards compatibility with other usages of `RegistryFinder#registry_from_rc`. I think it's used just once and could probably use the changed behavior too, I'll look into what would mean to change it.

It would make the code a bit simpler and may fix additional issues, so worth looking into it.

npm_and_yarn/helpers/lib/pnpm/lockfile-parser.js
```diff
@@ -33,6 +33,7 @@ function nameVerDevFromPkgSnapshot(depPath, pkgSnapshot, projectSnapshots) {
 
   let specifiers = [];
   let aliased = false;
+  let resolved = pkgSnapshot.resolution.tarball;
```
yeikel, line 36: nit: Is there any reason reason to use `let` as opposed to `const`?

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/registry_finder.rb
```diff
@@ -208,34 +215,40 @@ def unique_registries(registries)
...
+
+        # rubocop:disable Metrics/PerceivedComplexity
+        def configured_global_registry
+          return @configured_global_registry if defined? @configured_global_registry
+
           npmrc_file&.content.to_s.scan(NPM_GLOBAL_REGISTRY_REGEX) do
             next if Regexp.last_match[:registry].include?("${")
 
-            return @global_registry = Regexp.last_match[:registry].strip
+            return @configured_global_registry = Regexp.last_match[:registry].strip
```
yeikel, line 231: Would it make sense to extract `Regexp.last_match[:registry].strip` to a method?

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/registry_finder.rb
```diff
@@ -279,6 +292,12 @@ def parsed_yarnrc_yml
 
           @parsed_yarnrc_yml = YAML.safe_load(yarnrc_yml_file.content)
         end
+
+        def normalize_configured_registry(url)
+          url.sub(%r{/+$}, "")
+             .sub(%r{^.*?//}, "")
+             .gsub(/\s+/, "%20")
+        end
```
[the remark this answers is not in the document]
yeikel: If you have a spec testing this then please point me to it

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/registry_finder.rb
deivid-rodriguez: I had a look at this but decided it was not worth exploring further right now.

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/registry_finder.rb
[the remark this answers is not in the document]
deivid-rodriguez: I added this to respect the format currently returned by `RegistryFinder#registry`. The format is a URL with no protocol, no trailing slashes, and escaped espaces. However, they may appear in other format in rc files, so needs to be normalized. There are many specs that fail if you don't do this.

Note that I did not add this logic, it was already there, I just extracted it to method so that I can reuse it.

deivid-rodriguez:
Thanks for reviewing @yeikel, hopefully I addressed your comments.

@dancallaghan If you're able to try this again, I appreciate. I introduced a few refactoring so it'd be good to know I did not break anything!

dancallaghan:
Thanks @deivid-rodriguez

I've tested locally with the latest changes on our use case and it's working as expected 👍

BPScott:
@deivid-rodriguez, It seems like the changes to the lockfile in pnpm v9 mean these improvements no longer apply.

The performance issues caused by dependabot checking non-standard repositories for all packages rather than the ones within configured namespaces have resurfaced

deivid-rodriguez:
I no longer actively contribute to this repo, so I won't be able to help right now, but thanks for letting me know!

PR #8331, opened 2023-11-03
handle updating manifests across multiple directories

jakecoffman:
follow-on to #8309

This makes the Updater run multiple updates across multiple directories, then smashes all the changes together and creates a PR. Confirmed with some of the smoke tests: https://github.com/dependabot/smoke-tests/pull/135

### DIFF
--- a/updater/lib/dependabot/updater/dependency_group_change_batch.rb
+++ b/updater/lib/dependabot/updater/dependency_group_change_batch.rb
@@ -1,6 +1,8 @@
 # typed: true
 # frozen_string_literal: true
 
+require "pathname"
+
 # This class is responsible for aggregating individual DependencyChange objects
 # by tracking changes to individual files and the overall dependency list.
 module Dependabot
@@ -22,9 +24,9 @@ module Dependabot
       end
 
       # Returns an array of DependencyFile objects for the current state
-      def current_dependency_files
-        @dependency_file_batch.map do |_path, data|
-          data[:file]
+      def current_dependency_files(directory)
+        @dependency_file_batch.filter_map do |path, data|
+          data[:file] if Pathname.new(path).dirname.to_s == directory
         end
       end
 
--- a/updater/lib/dependabot/updater/dependency_group_change_batch.rb
+++ b/updater/lib/dependabot/updater/dependency_group_change_batch.rb
@@ -25,6 +25,7 @@ module Dependabot
 
       # Returns an array of DependencyFile objects for the current state
       def current_dependency_files(directory)
+        directory = Pathname.new(directory).cleanpath.to_s
         @dependency_file_batch.filter_map do |path, data|
           data[:file] if Pathname.new(path).dirname.to_s == directory
         end
--- a/updater/lib/dependabot/updater/dependency_group_change_batch.rb
+++ b/updater/lib/dependabot/updater/dependency_group_change_batch.rb
@@ -24,8 +24,12 @@ module Dependabot
       end
 
       # Returns an array of DependencyFile objects for the current state
-      def current_dependency_files(directory)
-        directory = Pathname.new(directory).cleanpath.to_s
+      def current_dependency_files(job)
+        directory = Pathname.new(job.source.directory).cleanpath.to_s
+
+        # GitHub Actions adds a special case where root directory scans the .github/workflows folder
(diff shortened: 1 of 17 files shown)


jakecoffman:
Look at https://github.com/dependabot/dependabot-core/pull/8331/commits/d84cd5cc61943f5e6f6c6432458e52313c2fd4e3 before I refactored to more easily see what I changed.

updater/lib/dependabot/updater/dependency_group_change_batch.rb
```diff
@@ -22,9 +22,9 @@ def initialize(initial_dependency_files:)
       end
 
       # Returns an array of DependencyFile objects for the current state
-      def current_dependency_files
-        @dependency_file_batch.map do |_path, data|
-          data[:file]
+      def current_dependency_files(directory)
+        @dependency_file_batch.filter_map do |path, data|
+          data[:file] if path.start_with?(directory)
```
jakecoffman, line 27: FYI This is not correct, need to do use proper path classes to figure out if the two are exactly the same, not `start_with`.

updater/lib/dependabot/updater/operations/create_group_security_update_pull_request.rb
```diff
@@ -81,6 +63,42 @@ def perform
...
+
+          if job.source.directories.nil?
+            @dependency_change = compile_all_dependency_changes_for(group)
+          else
+            dependency_changes = job.source.directories.map do |directory|
+              job.source.directory = directory
+              # Fixes not updating because it already updated in a previous group
+              dependency_snapshot.handled_dependencies.clear
+              compile_all_dependency_changes_for(group)
+            end
```
Nishnha, line 94: This is a cool approach! I think it makes a lot of sense. I do think we should make the logic of swapping the job directory more explicit though. I'll take a stab at refactoring it

updater/lib/dependabot/updater/dependency_group_change_batch.rb
deivid-rodriguez: So, if I'm understanding this correctly, the input `dependency_files` to `FileParser`, `UpdateChecker` and `FileUpdater` won't change, correct? In other words, we can still assume that list of files has the same `directory` for all of them?

jakecoffman: Correct! 🎯

updater/lib/dependabot/updater/operations/create_group_security_update_pull_request.rb
jakecoffman: I hear you, but I wouldn't worry too much about refactoring and polishing yet. We can do that in follow-on PRs. For now I think it's important to get this out and tested so we can make progress!

jakecoffman:
Our test fixtures had "/bundler" but then we were mocking the files out to be at directory "/" so I made those consistent to get the tests passing.

honeyankit (approved):
LGTM

jakecoffman:
Upon deploying I noticed an increase in errors which was caused by how security updates come in with directory `/.`. Since this wasn't being normalized it wasn't matching in the new filtering, and so Dependabot was throwing RuntimeError. 

I fixed this in https://github.com/dependabot/dependabot-core/pull/8331/commits/bc259c23958ec80731bbb2837f9d1505ffaaaf58 and added tests.

honeyankit (approved):

jakecoffman:
Looks like there's one more bug in the same area. When we process `github-actions` jobs we allow directory to be `/` even though the files are at `/.github/workflows` so by filtering out we're causing the same RuntimeError as above. We'll need to add an exception for Actions I think?

The reason our smoke tests didn't catch this is we added a separate directory `/actions` to avoid the smoke test checking our actual workflows for updates. However this kind of setup is extremely rare, so I'll start by making the smoke test for Actions more realistic.

jakecoffman:
I updated the smoke test in https://github.com/dependabot/smoke-tests/pull/139 and now it's correctly failing.

jakecoffman:
A little unfortunate but I had to add a special case in https://github.com/dependabot/dependabot-core/pull/8331/commits/0a4ac826aeeac49edd66974b2bc59832f5c1069b for github-actions where directory `/` actually means `.github/workflows` which is what that ecosystem does.

PR #8333, opened 2023-11-04
npm: Fall back on configured global registry for scoped packages

dancallaghan:
If a scoped registry is not configured for a scoped dependency, however a global registry is configured, it should fall back on the configured global registry.

As discussed in https://github.com/dependabot/dependabot-core/issues/8242#issuecomment-1793136986


deivid-rodriguez (approved):
Thank you! ❤️

PR #8344, opened 2023-11-06
Simplify group updates by dep type test to fix CI

deivid-rodriguez:
Just check that only prod (and not dev) dependencies are updated (and the other way around for the opposite situation), instead of asserting for exact lockfiles, which are brittle to releases of transitive dependencies.

I could do a finer check for subdependencies, but this seems enough to me?


updater/spec/dependabot/updater/operations/group_update_all_versions_spec.rb
```diff
@@ -233,7 +233,7 @@
         gemfile_lock = dependency_change.updated_dependency_files.find do |file|
           file.path == "/Gemfile.lock"
         end
-        expect(gemfile_lock.content).to eql(fixture("bundler_grouped_by_types/updated_development_deps/Gemfile.lock"))
+        expect(gemfile_lock.content).to include("rubocop (1.56.0)").and include("rack (2.1.3)")
```
landongrindheim, line 236: I like having the structure of the `Gemfile.lock` file as a reference, but the relationship between the input and output for this test aren't exactly clear.

By using the top-level declarations, rather than requirements, I think this is sound 👍

landongrindheim (approved):

PR #8310, opened 2023-11-01
Type `SharedHelpers` more thoroughly

JamieMagee:

PR #8347, opened 2023-11-06
Fix common CI

deivid-rodriguez:
Apparently #8310 broke some common specs. No idea why CI did not catch that, but can be reproduced with

```
rspec ./spec/dependabot/file_fetchers/base_spec.rb:1646
```

from common.

This PR removes the signature from `run_shell_command` until we can address the real culprit, since that seems to let the specs pass.

### DIFF
--- a/sorbet/config
+++ b/sorbet/config
@@ -1,6 +1,7 @@
 --dir=.
 --ignore=tmp/
 --ignore=vendor/
+--ignore=.bundle/
 --disable-watchman
 
 # Sorbet doesn't currently support RSpec very well, so we ignore all of our specs.
--- a/common/spec/dependabot/file_fetchers/base_spec.rb
+++ b/common/spec/dependabot/file_fetchers/base_spec.rb
@@ -1635,21 +1635,12 @@ RSpec.describe Dependabot::FileFetchers::Base do
     let(:repo_contents_path) { Dir.mktmpdir }
     let(:submodule_contents_path) { File.join(repo_contents_path, "examplelib") }
 
-    before do
-      allow(Dependabot::SharedHelpers)
-        .to receive(:run_shell_command).and_call_original
-    end
-
     after { FileUtils.rm_rf(repo_contents_path) }
 
     describe "#clone_repo_contents" do
       it "does not clone submodules by default" do
         file_fetcher_instance.clone_repo_contents
 
-        expect(Dependabot::SharedHelpers)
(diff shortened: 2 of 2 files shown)


deivid-rodriguez:
@JamieMagee I think this may be a fun one for you 😃

honeyankit (approved):

honeyankit:
> This PR removes the signature from run_shell_command
 
 Just checking, Is there any harm in removing the `signature`

deivid-rodriguez:
I think @JamieMagee may answer that better but my understanding is that it just means less strict typing. It's definitely not a permanent solution, just a way to get our CI green for now and unblock other PRs. We would readd the signature and reenable strict typing as soon as we find the culprit.

JamieMagee:
Can you share an example of the failure?

@deivid-rodriguez is right, it's just less strict type-checking

sorbet/config
```diff
@@ -1,6 +1,7 @@
 --dir=.
 --ignore=tmp/
 --ignore=vendor/
+--ignore=.bundle/
```
JamieMagee, line 4: What makes this change necessary?

deivid-rodriguez: Oh, nothing, this is just because I run sorbet locally, and since I had files at `bundler/helpers/v1/.bundle` and `bundler/helpers/v2/.bundle` (probably because of having run bundler specs), sorbet showed a bunch of failures in those installed gems.

deivid-rodriguez:
Yes, see https://github.com/dependabot/dependabot-core/actions/runs/6774474932/job/18411610576?pr=8337.

deivid-rodriguez:
I pushed a different change that allows specs to pass while keeping the strict typing. I guess this is just poor RSpec support.

JamieMagee:
> I pushed a different change that allows specs to pass while keeping the strict typing. I guess this is just poor RSpec support.

Poor RSpec support in Sorbet?

deivid-rodriguez:
Yes, that's what I meant.

Just realized my change is not working, not sure why I thought it was 😢

deivid-rodriguez:
Ok, I pushed yet another different approach. This one should be fine I believe.

deivid-rodriguez:
I guess [these](https://github.com/search?q=repo%3Asorbet%2Fsorbet+and_call_original&type=issues) could be related? In any case, I think it's best to not rely on these mocks and move on.

jakecoffman (approved):

PR #8355, opened 2023-11-07
Fix incorrect casing in some Dependabot PR titles

deivid-rodriguez:
When figuring out which case to use, we were looking at previous Dependabot commit messages.

However, our regexps were trying to match the whole commit messages, not just the commit title, and that was causing some false positives.


jakecoffman (approved):

steebchen:
Thanks for the swift fix :D

PR #8359, opened 2023-11-08
Remove feature flag gating dependency removal

mctofu:
We needed to gate this feature because older versions of GHES didn't have support for PRs that remove dependencies. All currently supported versions of GHES now support this so we can remove the flag.


abdulapopoola (approved):

deivid-rodriguez (approved):
Thanks for cleaning up! :)

PR #8361, opened 2023-11-08
Fix more test failures

deivid-rodriguez:
Probably due to sorbet and rspec's `and_call_original` not playing nice with each other.

See for example latest run in main branch's CI: https://github.com/dependabot/dependabot-core/actions/runs/6787005329/job/18448834031.


jurre:
> Probably due to sorbet and rspec's and_call_original not playing nice with each other.

Can you share more about how they don't play nice with each other?

jurre (approved):

deivid-rodriguez:
I don't have a lot of information. I just know that migrating away from `and_call_original` makes the tests pass.

I should've mentioned that this is a follow up to https://github.com/dependabot/dependabot-core/pull/8347, where more similar issues where found. In there I mentioned [this link](https://github.com/search?q=repo%3Asorbet%2Fsorbet+and_call_original&type=issues) showing a few issues in sorbet mentioning `and_call_original`.

PR #8371, opened 2023-11-13
Respect poetry explicit source

### LINKED ISSUE #7918: dependabot updater still check poetry `explicit` source

### Package ecosystem

poetry

### Package manager version

poetry 1.6

### Language version

python 3.10

`/pyproject.toml` and `/poetry.lock`

### dependabot.yml content

```
version: 2
updates:
  - package-ecosystem: "pip" # See documentation for possible values
    directory: "/" # Location of package manifests
    schedule:
      interval: "daily"
    open-pull-requests-limit: 5
    allow:
      - dependency-type: "all"
  - package-ecosystem: "docker" # See documentation for possible values
    directory: "/" # Location of package manifests
    schedule:
      interval: "daily"
    open-pull-requests-limit: 20
```

### Updated dependency

_No response_

Poetry 1.5.0 introduced explicit package source
https://python-poetry.org/docs/repositories/#explicit-package-sources

> If package sources are configured as explicit, these sources are only searched when a package configuration [explicitly indicates](https://python-poetry.org/docs/repositories/#package-source-constraint) that it should be found on this package source.

However, in latest dependabot, the updater will still check package against source marked as `explicit`. Which double the execution time. 

```
updater | 2023/08/29 06:15:39 INFO <job_715108988> Checking if protobuf 3.20.3 needs updating
  proxy | 2023/08/29 06:15:39 [044] GET https://pypi.org:443/simple/protobuf/
  proxy | 2023/08/29 06:15:39 [044] 200 https://pypi.org:443/simple/protobuf/
  proxy | 2023/08/29 06:15:40 [046] GET https://[my private pypi]/pypi/protobuf/
  proxy | 2023/08/29 06:15:40 [046] 404 https://[my private pypi]/pypi/protobuf/
```

will not check protobuf against my private pypi

_No response_

_No response_

lucemia:
## Context

- The issue #7918 is being addressed to incorporate support for Poetry's explicit package source.
- Following this update, the explicit package source will only be utilized if a package explicitly specifies it as the source.

Poetry 1.5.0 introduced the concept of an explicit package source, which can be found here: [Explicit Package Sources](https://python-poetry.org/docs/repositories/#explicit-package-sources).

> If package sources are configured as explicit, these sources are only searched when a package configuration [explicitly indicates](https://python-poetry.org/docs/repositories/#package-source-constraint) that it should be found on this package source.

## Summary of Modifications
- An update has been made to the pyproject_files_parser, which previously ignored package source configuration in pyproject.toml but will now correctly read the value.
- The index_finder has also been updated to ignore any source marked as explicit unless it is specified as the source of the package.

## Testing
- test with `./bin/dry-run.rb` and check explicit package source won't be consider unless it is defined as the source of a package

## User facing changes
- a source marked as explicitly, won't be check for package that not specified it as source
- other behavior should be the same

## To-Do:
- based on the document [package-source-constraint](https://python-poetry.org/docs/repositories/#package-source-constraint), If a package specifies a source, it should not check other sources. Currently all sources (except explicitly) will still be considered

> A repository that is configured to be the only source for retrieving a certain package can itself have any priority. In particular, it does not need to have priority "explicit". If a repository is configured to be the source of a package, it will be the only source that is considered for that package and the repository priority will have no effect on the resolution.

Fixes #7918.

### DIFF
--- a/python/spec/dependabot/python/file_parser/pyproject_files_parser_spec.rb
+++ b/python/spec/dependabot/python/file_parser/pyproject_files_parser_spec.rb
@@ -251,6 +251,15 @@ RSpec.describe Dependabot::Python::FileParser::PyprojectFilesParser do
         expect(dependency_names).to include("sphinx")
       end
     end
+
+    context "with package specify source" do
+      let(:pyproject_fixture_name) { "package_specify_source.toml" }
+      subject(:dependency) { dependencies.find { |f| f.name == "black" } }
+
+      it "specifies a package source" do
+        expect(dependency.requirements[0][:source]).to eq("custom")
+      end
+    end
   end
 
   describe "parse standard python files" do
--- a/python/spec/dependabot/python/update_checker/index_finder_spec.rb
+++ b/python/spec/dependabot/python/update_checker/index_finder_spec.rb
@@ -9,7 +9,8 @@ RSpec.describe Dependabot::Python::UpdateChecker::IndexFinder do
   let(:finder) do
     described_class.new(
       dependency_files: dependency_files,
-      credentials: credentials
+      credentials: credentials,
+      dependency: dependency
     )
   end
   let(:credentials) do
@@ -21,6 +22,19 @@ RSpec.describe Dependabot::Python::UpdateChecker::IndexFinder do
     }]
   end
   let(:dependency_files) { [requirements_file] }
+  let(:dependency) do
+    Dependabot::Dependency.new(
+      name: "requests",
+      version: "2.4.1",
+      requirements: [{
+        requirement: "==2.4.1",
+        file: "requirements.txt",
+        groups: ["dependencies"],
+        source: nil
+      }],
+      package_manager: "pip"
+    )
+  end
 
   before do
     stub_request(:get, pypi_url).to_return(status: 200, body: pypi_response)
@@ -302,6 +316,43 @@ RSpec.describe Dependabot::Python::UpdateChecker::IndexFinder do
               "https://pypi.org/simple/",
               "https://some.internal.registry.com/pypi/"
             ]
+          )
+        end
+      end
+
+      context "when set in a pyproject.toml file and marked as explicit" do
+        let(:pyproject_fixture_name) { "extra_source_explicit.toml" }
+        let(:dependency_files) { [pyproject] }
+
+        it "gets the right index URLs" do
+          expect(index_urls).to match_array(
+            [
+              "https://pypi.org/simple/"
+            ]
+          )
+        end
+      end
+
+      context "when set in a pyproject.toml file and marked as explicit and specify with source" do
+        let(:pyproject_fixture_name) { "extra_source_explicit_and_package_specify_source.toml" }
+        let(:dependency_files) { [pyproject] }
+        let(:dependency) do
+          Dependabot::Dependency.new(
+            name: "requests",
+            version: "2.4.1",
+            requirements: [{
+              requirement: "==2.4.1",
+              file: "requirements.txt",
+              groups: ["dependencies"],
+              source: "custom"
+            }],
+            package_manager: "pip"
+          )
+        end
+
+        it "gets the right index URLs" do
+          expect(index_urls).to match_array(
+            ["https://pypi.org/simple/", "https://some.internal.registry.com/pypi/"]
           )
         end
       end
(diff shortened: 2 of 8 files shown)


python/spec/dependabot/python/file_parser/pyproject_files_parser_spec.rb
```diff
@@ -251,6 +251,14 @@
         expect(dependency_names).to include("sphinx")
       end
     end
+
+    context "with package specify source" do
+      let(:pyproject_fixture_name) { "package_specify_source.toml" }
+      subject(:dependency) { dependencies.find { |f| f.name == "black" } }
+      it "package source is specified" do
```
deivid-rodriguez, line 258: I'd normally add a blank line before this one, and try to make the description read well in English, for example, "`it "specifies a package source" do`.

python/spec/dependabot/python/update_checker/index_finder_spec.rb
```diff
@@ -306,6 +320,43 @@
         end
       end
 
+      context "set in a pyproject.toml file and marked as explicit" do
```
deivid-rodriguez, line 323: You'll normally start context with `when` or `with`. For example, `context "when set in a pyproject.toml file and marked as explicit" do`. If you do this, when you run `rspec --format documentation`, it reads nicer :)

python/spec/dependabot/python/update_checker/index_finder_spec.rb
```diff
@@ -306,6 +320,43 @@
...
+        it "gets the right index URLs" do
+          expect(index_urls).to match_array(
+            [
+              "https://pypi.org/simple/"
+            ]
+          )
+        end
+      end
+
+      context "set in a pyproject.toml file and marked as explicit and specify with source" do
```
deivid-rodriguez, line 336: Same nitpick about spec wording.

python/spec/fixtures/pyproject_files/extra_source_explicit.toml
```diff
@@ -0,0 +1,17 @@
...
+description = "Various small python projects."
+
+[tool.poetry.dependencies]
+python = "^3.7"
+requests = "2.18.0"
+
+[[tool.poetry.source]]
+name = "custom"
+url = "https://some.internal.registry.com/pypi/"
+priority = "explicit"
```
deivid-rodriguez, line 17: Can we add a EOF character to this fixture?

Also, small question, my understanding is that this is not a common "realworld" file because the explicit source is not used by any dependency, but it's enough for our tests because the source information is already included in the `Dependency` instance that you are using for testing?

deivid-rodriguez (approved):
This looks great.

I added a few nitpick comments, nothing that blocks merging this as is, really, but feel free to address!

python/spec/fixtures/pyproject_files/extra_source_explicit.toml
lucemia: > Can we add a EOF character to this fixture?
> 

Sure, will added

> Also, small question, my understanding is that this is not a common "realworld" file because the explicit source is not used by any dependency, but it's enough for our tests because the source information is already included in the `Dependency` instance that you are using for testing?

Indeed, this is not a real-world scenario. The test case utilizes the Dependency instance like other tests in index_finder_spec.rb. 

`extra_source_explicit.toml` has been intentionally crafted as a minimal example to replicate the bug #7918, and it is being tested in the unit test named **when set in a pyproject.toml file and marked as explicit.**

It might cause some confusion to reuse the TOML, so I will add a distinct TOML configuration in the unit test **when set in a pyproject.toml file and marked as explicit and specify with source**

python/spec/dependabot/python/update_checker/index_finder_spec.rb
```diff
@@ -333,8 +333,8 @@
         end
       end
 
-      context "set in a pyproject.toml file and marked as explicit and specify with source" do
-        let(:pyproject_fixture_name) { "extra_source_explicit.toml" }
+      context "when set in a pyproject.toml file and marked as explicit and specify with source" do
+        let(:pyproject_fixture_name) { "extra_source_explicit_and_pacakge_specify_source.toml" }
```
deivid-rodriguez, line 337:
```suggestion
        let(:pyproject_fixture_name) { "extra_source_explicit_and_package_specify_source.toml" }
```

deivid-rodriguez (commented):
Looks great now, could you squash all commits into one to simplify git history?

lucemia:
@deivid-rodriguez done!

deivid-rodriguez (approved):
Awesome, I will merge and deploy this later today! Thanks for the great work 💪

deivid-rodriguez:
Thanks so much for the nice contribution!

PR #8378, opened 2023-11-13
fix: Use portable Bash shebang

l0b0:
Makes it possible to use Dependabot CLI on distros where the `bash` executable is not in `/bin`, such as NixOS.


terraform/spec/fixtures/registry_responses/hashicorp_consul_aws_0.3.8.json
JamieMagee, line 1: Revert this change please.

l0b0: Done.

l0b0:
There seems to be some flaky tests unrelated to this change: https://github.com/dependabot/dependabot-core/actions/runs/6856999542/job/18645260774 and https://github.com/dependabot/dependabot-core/actions/runs/6857109312/job/18645578043

l0b0:
Ping?

JamieMagee (approved):
Changes look good, but I don't have merge rights.

l0b0:
> Changes look good, but I don't have merge rights.

Welp. I can't request a review either.

JamieMagee:
That I can do 😁

jakecoffman (approved):
Looks good.

We're running shellcheck in our lint workflow. Is there an optional check we can enable to make sure these stay portable?

l0b0:
> We're running shellcheck in our lint workflow. Is there an optional check we can enable to make sure these stay portable?

ShellCheck does not have any [tests](https://gist.github.com/nicerobot/53cee11ee0abbdc997661e65b348f375#file-_shellcheck-md) for a portable Bash shebang line. In fact, they have rejected it several times in the past, based on these arguments:

- [A malicious user could hijack script runs by inserting a `bash` executable before the "real" Bash on the `PATH`](https://github.com/koalaman/shellcheck/issues/880#issuecomment-298032886). If the system is that compromised already, the only way to stay "safe" would be for you and every script you run to refer to every command by its full path, which nobody does. So I don't see why this is valid.
- [`/usr/bin/env` might not exist, or Bash could not be on the `PATH`](https://github.com/koalaman/shellcheck/issues/1047#issuecomment-357368441). At that point we're dealing with an obscure platform. Anyone running such a platform is probably capable of understanding the relevant error message and correct the shebang themselves. On the other hand, they would also be capable of adding the ShellCheck `ignore` directive themselves.

All that said, I can understand why the ShellCheck people don't want to deal with anything to do with the outside of the shell itself.

PR #8386, opened 2023-11-14
Nuget overhaul

deivid-rodriguez:
Just trying to get specs at #8179 green. Opening a WIP PR to check CI.


nuget/lib/dependabot/nuget/update_checker/nupkg_fetcher.rb
```diff
@@ -0,0 +1,121 @@
...
+          package_id_downcased = package_id.downcase
+          "#{base_url}/package/#{package_id_downcased}/#{package_version}"
+        end
+
+        def self.fetch_stream(stream_url, auth_header, max_redirects = 5)
+          current_url = stream_url
+          current_redirects = 0
+
+          loop do
+            connection = Excon.new(current_url, persistent: true)
```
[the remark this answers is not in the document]
deivid-rodriguez: Ultimately, the base url here gets picked up from user registry configuration. That's exactly the intended behavior here, I don't know which validation we could be adding. I'll dismiss this one unless someone has other ideas.

nuget/spec/dependabot/nuget/update_checker/version_finder_spec.rb
```diff
@@ -288,15 +327,15 @@
 
           custom_v2_nuget_versions_url =
             "https://www.nuget.org/api/v2/FindPackagesById()?id=" \
-            "'Microsoft.Extensions.DependencyModel'"
+            "'#{dependency_name}'"
```
brettfo, line 330: I was looking at this test locally and I think part of this was a really old copy/paste bug; there is no `4.8.1` of `Microsoft.Extensions.DependencyModel`, but there _is_ of `Autofac`, so using `VCR=all` might not work.

brettfo:
This is absolutely fantastic!  I reviewed the changes you've made and so far it all looks good.  I added a comment about one instance I found when testing locally that might make things easier (`Microsoft.Extensions.DependencyModel` vs `Autofac`), but that's all I think I need to mention.

Also, I was pointed to your comments on Slack, but I'm away from my computer that can access Slack until next week, so you can contact me via GitHub comments or email (brettfo@microsoft.com).

deivid-rodriguez:
Awesome @brettfo, let's discuss here publicly.

Is it ok to force-push this changes to #8179? I can keep a local copy of the current state just in case, but if the work in this PR seems solid I'd like to incorporate it there and close this PR. I'd also like to do some basic commit message cleanup to remove references to your issue tracker and make life of future git spelunkers easier.

nuget/spec/dependabot/nuget/update_checker/version_finder_spec.rb
deivid-rodriguez: Interesting!

These tests are no longer end-to-end, since I removed VCR from them, because:
* VCR was causing problems in our CI which does not allow `VCR=all` mode.
* There were too many cassettes committed to the repo.

So I changed these to be unit specs for each class, which is what other ecosystems normally do. Because of this, it no longer really matters if the dependency matches a real dependency or not.

For end to end tests we use https://github.com/dependabot/smoke-tests, so it'd be good to move a end to end test there.

brettfo:
Yes, by all means you can force-push to the other PR.  Do I need to grant you contributor status on my fork?

deivid-rodriguez:
I'm not sure, but I think unless you explicitly opted out to disallow maintainers from editing it when creating the PR, I should be able to do it. I'll try and let you know if I find issues.

deivid-rodriguez:
Yeah, I could force push without issues. Closing in favor of #8179.

PR #8389, opened 2023-11-14
yarn:update add a handled error for missing tags

pavera:
In the case of git dependencies, if the version is not tagged correctly the error is not caught as a resolvability error. This adds handling for the error message if a release is not tagged as expected.


jurre:
Any way to capture this in a test? Would that be valuable?

pavera:
> Any way to capture this in a test? Would that be valuable?

I'll take a look at existing tests and see what we have for other resolvability errors.

jurre (approved):

deivid-rodriguez (commented):
The way you explained this, it sounds like maybe `Dependabot::GitDependencyReferenceNotFound` is a better fit for this error?

Regarding testing this, my initial approach is normally:

* Introduce a dummy change in a existing analogous regex/string.
* Rerun tests and see what breaks.
* Use the failing test as a starting point for a new test.

deivid-rodriguez:
@pavera Any comments on my previous feedback?

deivid-rodriguez:
I'm good with merging this without tests if this is hard to test, but it sounds like the user error could be a better one if I understood your explanation correctly?

PR #8390, opened 2023-11-14
Use blobless clones

deivid-rodriguez:
I guess this could be a minor performance improvement.

This was extracted from #8179 since it affects all ecosystems, so should be introduced more carefully.

### DIFF
--- a/common/lib/dependabot/file_fetchers/base.rb
+++ b/common/lib/dependabot/file_fetchers/base.rb
@@ -753,7 +753,7 @@ module Dependabot
           FileUtils.mkdir_p(path)
 
           clone_options = StringIO.new
-          clone_options << "--no-tags --depth 1"
+          clone_options << "--no-tags --depth 1 --filter=blob:none"
           clone_options << if recurse_submodules_when_cloning?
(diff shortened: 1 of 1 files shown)


jurre:
Wouldn't this introduce issues when trying to grab things like `.gem` or other vendored files? Not super familiar with this git feature so might be misunderstanding

deivid-rodriguez:
My understanding from looking at the diagrams at https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/ is that this is redundant.

`--blob:none` only fetches blobs lazily when doing checkout. But we're already passing `--depth=1` we'll be checking cloning a single reference, which we _do_ checkout. So, `--blob:none` saves nothing in this context.

I would close this and remove it from #8179.

deivid-rodriguez:
For now I removed it from the NuGet PR.

brettfo:
Tagging @JamieMagee since he has more context around this feature.

deivid-rodriguez:
Closing for now. We can keep discussing in this closed PR and eventually reopen if needed.

PR #8423, opened 2023-11-17
Capture dependencies groups with inline comments in pyproject files

dsuleimenov:
Fix for the following issue:
https://github.com/dependabot/dependabot-core/issues/8366

### DIFF
--- a/python/spec/dependabot/python/file_updater/poetry_file_updater_spec.rb
+++ b/python/spec/dependabot/python/file_updater/poetry_file_updater_spec.rb
@@ -501,6 +501,155 @@ RSpec.describe Dependabot::Python::FileUpdater::PoetryFileUpdater do
         end
       end
 
+      context "with inline comments in the dependencies groups" do
+        let(:pyproject_fixture_name) { "inline_comments.toml" }
+        let(:dependency) do
+          Dependabot::Dependency.new(
+            name: dependency_name,
+            version: "1.27.2",
+            previous_version: "1.18.1",
+            package_manager: "pip",
+            requirements: requirements,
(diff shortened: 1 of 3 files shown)


python/spec/dependabot/python/file_updater/poetry_file_updater_spec.rb
```diff
@@ -501,6 +501,137 @@
...
+              file: "pyproject.toml",
+              source: nil,
+              groups: ["dev"]
+            }, {
+              requirement: ">=3.8.2",
+              file: "pyproject.toml",
+              source: nil,
+              groups: ["test"]
+            }]
+          )
```
deivid-rodriguez, line 544: In real life, all `requirements` here will be for the same dependency, so to better represent real behavior, each dependency should configure the requirements array for the dependency it's testing (similarly to how you customized `dependency_name`. Does that make sense? Basically, I think you'd need to add a `let(:requirements)` to each spec, and use that here.  Does that make sense?

deivid-rodriguez (commented):
This is great, just left a minor comment on the specs!

python/spec/dependabot/python/file_updater/poetry_file_updater_spec.rb
dsuleimenov: Certainly, I agree with your suggestion.
Thanks for the review, done 👍

dsuleimenov:
Hi @deivid-rodriguez
Could you please review my PR again when you have time?

deivid-rodriguez:
Thanks so much for the fix! This is now deployed to the Github service, let us know if you run into any issues!

Issue #7320, opened 2023-05-15
Dependabot does not locate new alpine image versions for Ruby Alpine docker images


mitchdawson1982:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

Docker

### Manifest location and content before the Dependabot update

  - package-ecosystem: "docker"
    directory: "/dns-service"
    schedule:
      interval: "daily"

### dependabot.yml content

version: 2
updates:
  - package-ecosystem: "bundler"
    directory: "/dns-service/metrics"
    schedule:
      interval: "daily"
  - package-ecosystem: "docker"
    directory: "/dns-service"
    schedule:
      interval: "daily"
  - package-ecosystem: "docker"
    directory: "/nginx"
    schedule:
      interval: "daily"

### Updated dependency

Update logs
  proxy | 2023/05/13 00:00:02 proxy starting, commit: d719b9b9871e853c5fd63bf3552f3f857d979f3c
  proxy | 2023/05/13 00:00:02 Listening (:1080)
updater | 2023-05-13T00:00:03.453347814 [662241047:main:WARN:src/devices/src/legacy/serial.rs:222] Detached the serial input due to peer close/error.
updater | time="2023-05-13T00:00:08Z" level=info msg="guest starting" commit=8ab4a20db815b67034070152643b9878c12b051d
updater | time="2023-05-13T00:00:08Z" level=info msg="starting job..." fetcher_timeout=10m0s job_id=662241047 updater_timeout=45m0s updater_version=49704c16cb0893c0ab8c5f884471c324baf39b83-docker
updater | 2023/05/13 00:00:13 INFO Raven 3.1.2 ready to catch errors
updater | 2023/05/13 00:00:18 INFO <job_662241047> Starting job processing
  proxy | 2023/05/13 00:00:18 [002] GET https://api.github.com:443/repos/ministryofjustice/staff-device-dns-server
  proxy | 2023/05/13 00:00:18 [002] * authenticating github api request with token for api.github.com
  proxy | 2023/05/13 00:00:19 [002] 200 https://api.github.com:443/repos/ministryofjustice/staff-device-dns-server
  proxy | 2023/05/13 00:00:19 [004] GET https://api.github.com:443/repos/ministryofjustice/staff-device-dns-server/git/refs/heads/main
  proxy | 2023/05/13 00:00:19 [004] * authenticating github api request with token for api.github.com
[...]

Issue #5708, opened 2022-09-13
Dependabot doesn't create a PR for docker updates where a multi-version image tag is used


Moser-ss:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

Docker

### Manifest location and content before the Dependabot update

self-host/runner/Dockerfile

```Dockerfile
FROM ghcr.io/actions-runner-controller/actions-runner-controller/actions-runner:v2.296.1-ubuntu-20.04
```

### dependabot.yml content

```yaml
version: 2
updates:
    - package-ecosystem: 'docker'
      directory: /self-host/runner
      schedule:
          interval: 'daily'
          time: '10:00'
      open-pull-requests-limit: 10
```

### Updated dependency

ghcr.io/actions-runner-controller/actions-runner-controller/actions-runner 

```json
{
    "name": "actions-runner-controller/actions-runner-controller/actions-runner",
    "tags": [
        "v2.285.1-ubuntu-20.04",
        "v2.285.1-ubuntu-20.04-2bd6d63",
        "v2.286.0-ubuntu-20.04-8a7720d",
        "v2.286.0-ubuntu-20.04",
        "v2.286.0-ubuntu-20.04-ad48851",
        "v2.286.1-ubuntu-20.04",
        "v2.286.1-ubuntu-20.04-5e86881",
        "v2.287.0-ubuntu-20.04",
        "v2.287.0-ubuntu-20.04-81b2c5a",
        "v2.287.1-ubuntu-20.04-715e6a4",
        "v2.287.1-ubuntu-20.04-f09a974",
        "v2.287.1-ubuntu-20.04-cc25dd7",
        "v2.287.1-ubuntu-20.04",
        "v2.287.1-ubuntu-20.04-6f591ee",
        "v2.288.0-ubuntu-20.04",
        "v2.288.0-ubuntu-20.04-f0fa99f",
        "v2.288.1-ubuntu-20.04-e42db00",
        "v2.288.1-ubuntu-20.04-c221b6e",
        "v2.288.1-ubuntu-20.04",
        "v2.288.1-ubuntu-20.04-b25a0fd",
        "v2.289.1-ubuntu-20.04-4cbbcd6",
        "v2.289.1-ubuntu-20.04-2cb04dd",
        "v2.289.1-ubuntu-20.04",
        "v2.289.1-ubuntu-20.04-debf53c",
        "v2.289.2-ubuntu-20.04",
[...]

PR #8432, opened 2023-11-18
Fix the new alpine image versions for docker

honeyankit:
## Context

Support issue where Dependabot logs show creation of PR, but no PR is created for the alpine image `node:20.8.1-alpine3.18`. 

Related issues: 
- https://github.com/dependabot/dependabot-core/issues/7320
- https://github.com/dependabot/dependabot-core/issues/5708

Root Cause:
The `version_requirement` is getting created correctly. But the issue is that during the checking of if 20.9.0-alpine3.18 is included in the version_requirement: "> 20.8.1.alpine3.18, < 20.9", it returns true which leads to Dependabot not creating PR and ignoring it. In the [version](https://github.com/dependabot/dependabot-core/blob/main/docker/lib/dependabot/docker/version.rb#L22) class of docker, the [split](https://github.com/dependabot/dependabot-core/blob/main/docker/lib/dependabot/docker/version.rb#L16) between release_part and the update_part is not happening for the version 20.9.0.alpine3.18 because of that, in the requirement class, the method [satisfied?](https://github.com/dependabot/dependabot-core/blob/main/docker/lib/dependabot/docker/requirement.rb#L17) is returning the whole version (20.9.0.alpine3.18) and the check for the [ignore_condition](https://github.com/github/dependabot-api/blob/main/app/models/ignore_condition.rb#L33) is returning true and PR being not created.

Main testing and validation changes:

* <a href="diffhunk://#diff-968e6fb682aaef429a9620f7531b5e7dc1739db00396eb560efeb68ae8e5fbb7R19-R35">`docker/spec/dependabot/docker/requirement_spec.rb`</a>: Added a new test case to test the behavior of the `Gem::Requirement.new` method when given a comma-separated string as the requirement string. <a href="diffhunk://#diff-968e6fb682aaef429a9620f7531b5e7dc1739db00396eb560efeb68ae8e5fbb7R19-R35">[1]</a> <a href="diffhunk://#diff-968e6fb682aaef429a9620f7531b5e7dc1739db00396eb560efeb68ae8e5fbb7R6">[2]</a>
* <a href="diffhunk://#diff-abf87b8e365d9d17077c9bec7059cfa59f938f297e92a2a30b7e49c87a720cadR38">`docker/spec/dependabot/docker/version_spec.rb`</a>: Added a new test case to check if the `check_version_for_correctness?` method correctly classifies a version string with an Alpine version as correct.
* <a href="diffhunk://#diff-fc43c2354fbec2f5957f7cd35c7b6e7f87e3372fe8cf2b9d50bddc512545889dR6">`docker/lib/dependabot/docker/version.rb`</a>: Modified the `initialize` method in the `Docker::Version` class to validate the version string when converting the image version to semver. <a href="diffhunk://#diff-fc43c2354fbec2f5957f7cd35c7b6e7f87e3372fe8cf2b9d50bddc512545889dR6">[1]</a> <a href="diffhunk://#diff-fc43c2354fbec2f5957f7cd35c7b6e7f87e3372fe8cf2b9d50bddc512545889dR17-R19">[2]</a>

## Anything to highlight to the reviewers
Initially added the `version = Tag.new(version).numeric_version` to the `file_parser` during building dependency to set the version via `.numeric_version`.  But the issue is that it changes the PR title and PR body and does not have the full version name (`20.8.1-alpine3.18`)

```bash
+----------------------------------------------------------------+
|              Changes to Dependabot Pull Requests               |
+---------+------------------------------------------------------+
| created | node ( from 20.8.1 to 20.9.0 ) |
+---------+------------------------------------------------------+
```


docker/spec/dependabot/docker/update_checker_spec.rb
```diff
@@ -344,7 +344,11 @@ def stub_tag_with_no_digest(tag)
 
       context "and we're using one" do
         let(:version) { "1803" }
-        it { is_expected.to eq("1903") }
+        it "expects another specific version" do
+          VCR.use_cassette("windows-servercore-1803") do
+            expect(subject).to eq("1903")
+          end
+        end
```
deivid-rodriguez, line 351: The second commit feels a bit suspicious, why does this change need extra network requests that need to be recorded? I think it would be the first case of us using VCR here and it would be good to know why, and also probably mock the requests instead of using VCR like the other specs do.

docker/lib/dependabot/docker/version.rb
```diff
@@ -14,9 +15,13 @@ module Docker
...
       def initialize(version)
         release_part, update_part = version.split("_", 2)
+        release_part = release_part.sub("v", "")
 
-        @release_part = Dependabot::Version.new(release_part.sub("v", "").tr("-", "."))
+        # The numeric_version is needed here to validate the version string (ex: 20.9.0-alpine3.18)
+        # when the call is made via Depenedabot Api to convert the image version to semver.
+        release_part = Tag.new(release_part).numeric_version
 
+        @release_part = Dependabot::Version.new(release_part.sub("v", "").tr("-", "."))
```
deivid-rodriguez, line 24: If we have subbed the "v" already, do we need to do it again?

deivid-rodriguez (commented):
I feel the docker ecosystem keeps getting out of hand as we fix things, I guess we need to attempt a proper refactor at some point. But if this works and fixes a user issue, let's do it I guess.

deivid-rodriguez:
To clarify, I think it's great to have a working solution for this tricky issue. The reason this approach feels a bit off is because it makes `Dependency::Docker::Version` dependent on an external class while normally `Version` classes are standalone classes. But this shouldn't necessarily be a problem and can be refactored later.

docker/spec/dependabot/docker/update_checker_spec.rb
deivid-rodriguez: I think I figured this out. It was a bug which #8442 should fix. After that PR, the second commit here should not be needed.

abdulapopoola (approved):

PR #8441, opened 2023-11-20
Strictly type `GitMetadataFetcher`

JamieMagee:


honeyankit (approved):
I liked the implementation of the GitRef class!

Issue #7000, opened 2023-04-04
Support for updating devcontainer.json files (for dev containers)


apgrucza:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

Dev containers are growing in popularity. They are used in VSCode and GitHub Codespaces, and support is being added to other tools such as Gitpod. Dependabot itself even uses them [here](https://github.com/dependabot/dependabot-core/blob/main/.devcontainer/devcontainer.json).

Dependabot should support updating versions in the `devcontainer.json` file (specification [here](https://containers.dev/implementors/json_reference/)). This should include updating both the `image` and `features` properties.

Image example:

```json
{
  "image": "mcr.microsoft.com/devcontainers/go:1"
}
```

Features example (note that both [feature versions and package versions](https://containers.dev/implementors/features/#devcontainer-json-properties) can appear in this object):

```json
"features": {
  "ghcr.io/user/repo/go": {},
  "ghcr.io/user/repo1/go:1": {},
  "ghcr.io/user/repo2/go:latest": {},
  "https://github.com/user/repo/releases/devcontainer-feature-go.tgz": { 
        "optionA": "value" 
  },
  "./myGoFeature": { 
        "optionA": true,
        "optionB": "hello",
        "version" : "1.0.0"
  }
}
```

jeffwidman:
👋 it's a good suggestion, we've had an internal team request this as well. 

No idea if/when we'll support this, given that we're (mostly) on pause for new ecosystems right now... long term we'd like to figure out a way to be more of a platform that others can build on top of so that if you wanted something you could build it yourself but that's still closer to a dream rather than reality.

For anyone else coming along, if adding support for this ecosystem is of interest please 👍 it as we use that as one signal of impact when prioritizing what to work on.

hfhbd:
Do you also have a configuration sample for the yaml file to enable the devcontainers ecosystem?

deivid-rodriguez:
Sorry, this is not yet fully released, let me reopen this and I'll share a link to documentation once it's working.

chris-bateman:
Might be too early to try this but I did anyway :)
Purposely mistyped devcontainer to generate the below error - 
![image](https://github.com/dependabot/dependabot-core/assets/12206103/6fc9d18e-cbd6-422a-a15a-2db13c98503a)

So devcontainers is in the package list.

I then update it to use that and get this error - 
[...]

PR #8445, opened 2023-11-21
Add `devcontainers` ecosystem

joshspicer:
This change provides initial support for a [devcontainers](https://containers.dev) ecosystem, leaning heavily on the [`devcontainers/cli`](https://github.com/devcontainers/cli) for ecosystem-specific implementation.

This updater will look in the provided `--directory` for a `devcontainer.json` (or `.devcontainer.json` if `directory === /`) and update all dev container [Features](https://containers.dev/implementors/features/) that are pinned to an out-of-date major version.  The [lockfile](https://github.com/devcontainers/spec/blob/main/docs/specs/devcontainer-lockfile.md) will also be updated if one is already present in the target directory.

I aim to add future updater granularity/modes in the future.

Requires at least dev container CLI [v0.54.0](https://github.com/devcontainers/cli/pull/689): 

## Demos (with the `dependabot/cli`)

```bash
LOCAL_GITHUB_ACCESS_TOKEN="$LOCAL_GITHUB_ACCESS_TOKEN"  dependabot  \
     --updater-image ghcr.io/dependabot/dependabot-updater-devcontainers  \
      update \
      devcontainers \
      joshspicer/versioning-project-two-configs \
      -o /tmp/updates.yaml \
      --directory /.devcontainer

updater | +-----------------------------------------------------------------------+
updater | |                  Changes to Dependabot Pull Requests                  |
updater | +---------+-------------------------------------------------------------+
updater | | created | ghcr.io/codspace/versioning/foo:1 ( from 1.1.0 to 2.11.1 )  |
updater | | created | ghcr.io/codspace/versioning/baz:1.0 ( from 1.0.0 to 2.0.0 ) |
updater | +---------+-------------------------------------------------------------+
```

```bash
LOCAL_GITHUB_ACCESS_TOKEN="$LOCAL_GITHUB_ACCESS_TOKEN"  dependabot  \
     --updater-image ghcr.io/dependabot/dependabot-updater-devcontainers  \
      update \
      devcontainers \
      joshspicer/versioning-project-two-configs \
      -o /tmp/updates.yaml \
      --directory /

updater | +----------------------------------------------------------------------+
updater | |                 Changes to Dependabot Pull Requests                  |
updater | +---------+------------------------------------------------------------+
updater | | created | ghcr.io/codspace/versioning/foo:1 ( from 1.1.0 to 2.11.1 ) |
updater | +---------+------------------------------------------------------------+
```

Closes https://github.com/dependabot/dependabot-core/issues/7000.

### DIFF
--- a/devcontainers/lib/dependabot/devcontainers/file_parser/feature_dependency_parser.rb
+++ b/devcontainers/lib/dependabot/devcontainers/file_parser/feature_dependency_parser.rb
@@ -65,6 +65,10 @@ module Dependabot
             # Skip sha pinned tags for now. Ideally the devcontainers CLI would give us updated SHA info
             next if name.end_with?("@sha256")
 
+            # Skip deprecated features until `devcontainer features info tag`
+            # and `devcontainer upgrade` work with them. See https://github.com/devcontainers/cli/issues/712
+            next unless name.include?("/")
+
             current = versions_object["current"]
 
             dep = Dependency.new(
--- a/devcontainers/lib/dependabot/devcontainers/file_parser/feature_dependency_parser.rb
+++ b/devcontainers/lib/dependabot/devcontainers/file_parser/feature_dependency_parser.rb
@@ -1,0 +1,100 @@
+# typed: true
+# frozen_string_literal: true
+
+require "dependabot/devcontainers/requirement"
+require "dependabot/file_parsers/base"
+require "dependabot/shared_helpers"
+require "dependabot/dependency"
+require "json"
+require "uri"
+
+module Dependabot
+  module Devcontainers
+    class FileParser < Dependabot::FileParsers::Base
+      class FeatureDependencyParser
+        def initialize(config_dependency_file:, repo_contents_path:, credentials:)
+          @config_dependency_file = config_dependency_file
+          @repo_contents_path = repo_contents_path
+          @credentials = credentials
+        end
+
+        def parse
+          SharedHelpers.in_a_temporary_repo_directory(base_dir, repo_contents_path) do
+            SharedHelpers.with_git_configured(credentials: credentials) do
+              parse_cli_json(evaluate_with_cli)
+            end
+          end
+        end
+
+        private
+
+        def base_dir
+          File.dirname(config_dependency_file.path)
+        end
+
+        def config_name
+          File.basename(config_dependency_file.path)
+        end
+
+        def config_contents
+          config_dependency_file.content
+        end
+
+        # https://github.com/devcontainers/cli/blob/9444540283b236298c28f397dea879e7ec222ca1/src/spec-node/devContainersSpecCLI.ts#L1072
+        def evaluate_with_cli
+          raise "config_name must be a string" unless config_name.is_a?(String) && !config_name.empty?
+
+          cmd = "devcontainer outdated --workspace-folder . --config #{config_name} --output-format json"
+          Dependabot.logger.info("Running command: #{cmd}")
+
+          json = SharedHelpers.run_shell_command(
+            cmd,
+            stderr_to_stdout: false
+          )
+
+          JSON.parse(json)
+        end
+
+        def parse_cli_json(json)
+          dependencies = []
+
+          features = json["features"]
+          features.each do |feature, versions_object|
+            name, requirement = feature.split(":")
+            current = versions_object["current"]
+            wanted = versions_object["wanted"]
+            latest = versions_object["latest"]
+            wanted_major = versions_object["wantedMajor"]
+            latest_major = versions_object["latestMajor"]
+
+            dep = Dependency.new(
+              name: name,
+              version: current,
+              package_manager: "devcontainers",
+              requirements: [
+                {
+                  requirement: requirement,
+                  file: config_dependency_file.name,
+                  groups: ["feature"],
+                  source: nil,
+                  metadata: {
+                    wanted: wanted,
+                    latest: latest,
+                    wanted_major: wanted_major,
+                    latest_major: latest_major
+                  }
+                }
+              ],
+              metadata: {}
+            )
+
+            dependencies << dep
+          end
+          dependencies
+        end
+
+        attr_reader :config_dependency_file, :repo_contents_path, :credentials
+      end
+    end
+  end
+end
--- a/devcontainers/lib/dependabot/devcontainers/file_parser/feature_dependency_parser.rb
+++ b/devcontainers/lib/dependabot/devcontainers/file_parser/feature_dependency_parser.rb
@@ -61,6 +61,10 @@ module Dependabot
           features = json["features"]
           features.each do |feature, versions_object|
             name, requirement = feature.split(":")
+
+            # Skip sha pinned tags for now. Ideally the devcontainers CLI would give us updated SHA info
+            next if name.end_with?("@sha256")
+
             current = versions_object["current"]
             wanted = versions_object["wanted"]
             latest = versions_object["latest"]
--- a/devcontainers/lib/dependabot/devcontainers/file_parser/feature_dependency_parser.rb
+++ b/devcontainers/lib/dependabot/devcontainers/file_parser/feature_dependency_parser.rb
@@ -66,10 +66,6 @@ module Dependabot
             next if name.end_with?("@sha256")
 
             current = versions_object["current"]
-            wanted = versions_object["wanted"]
-            latest = versions_object["latest"]
-            wanted_major = versions_object["wantedMajor"]
-            latest_major = versions_object["latestMajor"]
 
             dep = Dependency.new(
               name: name,
@@ -80,16 +76,9 @@ module Dependabot
                   requirement: requirement,
                   file: config_dependency_file.name,
                   groups: ["feature"],
-                  source: nil,
-                  metadata: {
-                    wanted: wanted,
-                    latest: latest,
-                    wanted_major: wanted_major,
-                    latest_major: latest_major
-                  }
+                  source: nil
                 }
-              ],
-              metadata: {}
+              ]
             )
 
             dependencies << dep
--- a/devcontainers/lib/dependabot/devcontainers/file_updater/config_updater.rb
+++ b/devcontainers/lib/dependabot/devcontainers/file_updater/config_updater.rb
@@ -10,9 +10,10 @@ module Dependabot
   module Devcontainers
     class FileUpdater < Dependabot::FileUpdaters::Base
       class ConfigUpdater
-        def initialize(feature:, requirement:, manifest:, repo_contents_path:, credentials:)
+        def initialize(feature:, requirement:, version:, manifest:, repo_contents_path:, credentials:)
           @feature = feature
           @requirement = requirement
+          @version = version
           @manifest = manifest
           @repo_contents_path = repo_contents_path
           @credentials = credentials
@@ -22,8 +23,8 @@ module Dependabot
           SharedHelpers.in_a_temporary_repo_directory(base_dir, repo_contents_path) do
             SharedHelpers.with_git_configured(credentials: credentials) do
               update_manifests(
-                target_requirement: requirement[:requirement],
-                target_version: requirement[:metadata][:latest]
+                target_requirement: requirement,
+                target_version: version
               )
 
               [File.read(manifest_name), File.read(lockfile_name)].compact
@@ -70,7 +71,7 @@ module Dependabot
           SharedHelpers.run_shell_command(cmd, stderr_to_stdout: false)
         end
 
-        attr_reader :feature, :requirement, :manifest, :repo_contents_path, :credentials
+        attr_reader :feature, :requirement, :version, :manifest, :repo_contents_path, :credentials
       end
     end
   end
--- a/devcontainers/lib/dependabot/devcontainers/file_updater/config_updater.rb
+++ b/devcontainers/lib/dependabot/devcontainers/file_updater/config_updater.rb
@@ -1,0 +1,75 @@
+# typed: true
+# frozen_string_literal: true
+
+require "dependabot/file_updaters/base"
+require "dependabot/shared_helpers"
+require "dependabot/logger"
+require "dependabot/devcontainers/utils"
+
+module Dependabot
+  module Devcontainers
+    class FileUpdater < Dependabot::FileUpdaters::Base
+      class ConfigUpdater
+        def initialize(feature:, requirement:, manifest:, repo_contents_path:, credentials:)
+          @feature = feature
+          @requirement = requirement
+          @manifest = manifest
+          @repo_contents_path = repo_contents_path
+          @credentials = credentials
+        end
+
+        def update
+          SharedHelpers.in_a_temporary_repo_directory(base_dir, repo_contents_path) do
+            SharedHelpers.with_git_configured(credentials: credentials) do
+              update_manifest(
+                target_requirement: requirement[:requirement],
+                target_version: requirement[:metadata][:latest]
+              )
+
+              [File.read(manifest_name), File.read(lockfile_name)].compact
+            end
+          end
+        end
+
+        private
+
+        def base_dir
+          File.dirname(manifest.name)
+        end
+
+        def manifest_name
+          File.basename(manifest.name)
+        end
+
+        def lockfile_name
+          Utils.expected_lockfile_name(manifest_name)
+        end
+
+        def update_manifest(target_requirement:, target_version:)
+          # First force target version to upgrade lockfile. I believe the CLI
+          # shoudl automatically upgrade the lockfile directly, but needs a
+          # explicit version at the moment. If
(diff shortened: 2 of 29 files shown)


updater/Gemfile.lock
```diff
@@ -381,4 +388,4 @@ DEPENDENCIES
   webmock (~> 3.18)
 
 BUNDLED WITH
-   2.4.22
+   2.3.26
```
deivid-rodriguez, line 391: You probably used Bundler 2.3.26 to regenerate the lockfile. Make sure you use Bundler 2.4.22 so this does not get downgraded. If you run `bundler` within the development image that should be the case already but if you run `BUNDLE_GEMFILE=updater/Gemfile bundle lock` outside of the development container, you'll need `gem install bundler:2.4.22` before that.

joshspicer: updated 👍

devcontainers/lib/dependabot/devcontainers/utils.rb
```diff
@@ -0,0 +1,24 @@
+# typed: false
+# frozen_string_literal: true
+
+module Dependabot
+  module Devcontainers
+    module Utils
+      def self.expected_config_basename(directory)
+        root_directory?(directory) ? ".devcontainer.json" : "devcontainer.json"
```
deivid-rodriguez, line 8: I believe directory is `/` by default is users don't configure it. That means, Dependabot won't just work for users using `.devcontainer/devcontainer.json` and not explicitly configuring a directory. Correct?

Should we instead, given a directory, look for both `.devcontainer.json`, and `.devcontainer/devcontainer.json`, relative to the configured directory?

devcontainers/spec/devcontainers/devcontainers/file_parser_spec.rb
```diff
@@ -0,0 +1,208 @@
...
+    end
+  end
+
+  context "with a .devcontainer.json in repo root" do
+    let(:project_name) { "config_in_root" }
+    let(:config_file_name) { ".devcontainer.json" }
+    let(:directory) { "/" }
+
+    let(:expectations) do
+      EXPECTED_PARSE_FROM_ROOT_DEV_CONTAINER
```
deivid-rodriguez, line 66: These top-level constants are a bit non standard, normally you'd put them inside a different global `let`.

devcontainers/spec/devcontainers/devcontainers_spec.rb
```diff
@@ -0,0 +1,10 @@
+# typed: false
+# frozen_string_literal: true
+
+require "spec_helper"
+require "dependabot/devcontainers"
+require_common_spec "shared_examples_for_autoloading"
+
+# RSpec.describe Dependabot::Devcontainers do
+#   it_behaves_like "it registers the required classes", "devcontainers"
+# end
```
deivid-rodriguez, line 10: Should this be enabled?

devcontainers/lib/dependabot/devcontainers/metadata_finder.rb
```diff
@@ -0,0 +1,22 @@
...
+
+module Dependabot
+  module Devcontainers
+    class MetadataFinder < Dependabot::MetadataFinders::Base
+      private
+
+      def look_up_source
+        # TODO: Make upstream changes to dev container CLI to point to docs.
+        #       Specifically, 'devcontainers features info' can be augmented to expose documentationUrl
+        nil
```
deivid-rodriguez, line 15: PRs would have no release or changelog info for now. Seems fine to get started but just noticing it.

devcontainers/lib/dependabot/devcontainers/file_updater/config_updater.rb
```diff
@@ -0,0 +1,71 @@
...
+
+          puts "Running command: #{cmd}"
+          SharedHelpers.run_shell_command(
+            "#{cmd} #{args}",
+            stderr_to_stdout: false
+          )
+      rescue SharedHelpers::HelperSubprocessFailed => e
+        puts "!! [UPDATER] Failed to run #{cmd} command:\n#{e.message}"
+        Dependabot.logger.info("Failed to run #{cmd} command:\n#{e.message}")
+        end
```
deivid-rodriguez, line 65: This is very weirdly indented, surprised rubocop did not catch it.

Also, it'd be better to not rescue this if it's not expected that it fails, so that we get issues reported in our internal tracker and don't leave bugs hidden in update job logs.

devcontainers/lib/dependabot/devcontainers/file_parser/feature_dependency_parser.rb
```diff
@@ -0,0 +1,105 @@
...
+          puts "Running command: #{cmd}"
+          begin
+            json = SharedHelpers.run_shell_command(
+              cmd,
+              stderr_to_stdout: false
+            )
+            JSON.parse(json)
+          rescue StandardError => e
+            puts "Error running shell command:\n#{e.message}"
+            Dependabot.logger.info("Failed to run 'devcontainer outdated' command:\n#{e.message}")
```
deivid-rodriguez, line 59: This is an interesting technique. It allows us to mostly skip the "update checker" step, which seems nice.

Using this approach only dependencies that have updates available get parsed, correct? So an update would never result in "no updates needed" because up to date dependencies are "invisible", correct?

deivid-rodriguez (commented):
Made a first review pass, seems pretty neat!

devcontainers/lib/dependabot/devcontainers/utils.rb
joshspicer: That's correct, this implementation requires setting `directory`.

I should not there's a third, more general option for where a config lives, `.devcontainer/<SOME_FOLDER>/devcontainer.json`, where `SOME_FOLDER` can be anything.  

I am on the fence on what makes the most sense.  I asked in Slack and sounded like this was a natural pattern (to use `directory` to point to the appropriate config).  If we don't require `directory` to be set when discovering configs, we'll need to check all folders nested in `.devcontainer` too (probably not a huge deal).  It also complicated some of the other components, but I could probably work around that

devcontainers/lib/dependabot/devcontainers/metadata_finder.rb
joshspicer: 👍  I have an idea of what to put here, but wasn't sure how to easily source it - I think making an upstream change to the devcontainers CLI to fetch this info would be the best approach (don't want to add too much devcontainer-specific logic here in core)

devcontainers/lib/dependabot/devcontainers/utils.rb
joshspicer: We chatted offline - sounds like this initial approach is OK. We can iterate if it becomes unclear.  

[ref](https://github.slack.com/archives/C03UT3Q6V34/p1701369938981659?thread_ts=1700257487.226549&cid=C03UT3Q6V34)

devcontainers/spec/devcontainers/devcontainers/file_parser_spec.rb
joshspicer: How would one go about testing different input variables (ie `directory`) without these lets?

devcontainers/spec/devcontainers/devcontainers/file_parser_spec.rb
```diff
@@ -0,0 +1,208 @@
...
+      name: config_file_name,
+      content: fixture("projects", project_name, directory, config_file_name)
+    )
+  end
+
+  let(:dependencies) { parser.parse }
+
+  shared_examples_for "parse" do
+    it "parses dependencies fine" do
+      expectations.each.with_index do |expected, _index|
```
deivid-rodriguez, line 45: You're not using `_index` so I think you can omit `with_index` altogether?

deivid-rodriguez:
I rebased the PR, addressed several issues, and implemented the new way of handling updates. Now we provide updates to all fetched files, relative to the configured directory (by default "/").

For example:

```
$ bin/dry-run.rb devcontainers joshspicer/versioning-project-two-configs 
=> cloning into /home/dependabot/tmp/joshspicer/versioning-project-two-configs
=> parsing dependency files
Running command: devcontainer outdated --workspace-folder . --config .devcontainer.json --output-format json
Running command: devcontainer outdated --workspace-folder . --config devcontainer.json --output-format json
=> updating 3 dependencies: ghcr.io/codspace/versioning/foo:1, ghcr.io/codspace/versioning/bar:1, ghcr.io/codspace/versioning/baz:1.0

=== ghcr.io/codspace/versioning/foo:1 (1.1.0)
 => checking for updates 1/3
 => latest available version is 2.11.1
 => latest allowed version is 2.11.1
Running command: devcontainer upgrade
Running command: devcontainer upgrade
 => bump ghcr.io/codspace/versioning/foo:1 from 1.1.0 to 2.11.1

    ± .devcontainer.json
    ~~~
    --- /tmp/original20231221-316-tawni8	2023-12-21 17:35:09.209309011 +0000
    +++ /tmp/updated20231221-316-8gbvq6	2023-12-21 17:35:09.209309011 +0000
    @@ -1,7 +1,7 @@
     {
         "image": "mcr.microsoft.com/devcontainers/typescript-node:18",
         "features": {
    -        "ghcr.io/codspace/versioning/foo:1": {},
    +        "ghcr.io/codspace/versioning/foo:2": {},
             "ghcr.io/codspace/versioning/bar:1": {}
         }
     }
    ~~~
    2 insertions (+), 2 deletions (-)

    ± .devcontainer/devcontainer.json
    ~~~
    --- /tmp/original20231221-316-6o3e3w	2023-12-21 17:35:09.212309011 +0000
    +++ /tmp/updated20231221-316-j9vt98	2023-12-21 17:35:09.212309011 +0000
    @@ -1,7 +1,7 @@
     {
         "image": "mcr.microsoft.com/devcontainers/typescript-node:18",
         "features": {
    -        "ghcr.io/codspace/versioning/foo:1": {},
    +        "ghcr.io/codspace/versioning/foo:2": {},
             "ghcr.io/codspace/versioning/bar:1": {},
             "ghcr.io/codspace/versioning/baz:1.0": {}
         }
    ~~~
    2 insertions (+), 2 deletions (-)

    ± .devcontainer/devcontainer-lock.json
    ~~~
    --- /tmp/original20231221-316-xio06s	2023-12-21 17:35:09.214309011 +0000
    +++ /tmp/updated20231221-316-ugjs8o	2023-12-21 17:35:09.214309011 +0000
    @@ -10,10 +10,10 @@
           "resolved": "ghcr.io/codspace/versioning/baz@sha256:37f36051adf6da0a43764b9669b945e0f06e11973e02c57ad261b03bb1057cb7",
           "integrity": "sha256:37f36051adf6da0a43764b9669b945e0f06e11973e02c57ad261b03bb1057cb7"
         },
    -    "ghcr.io/codspace/versioning/foo:1": {
    -      "version": "1.1.0",
    -      "resolved": "ghcr.io/codspace/versioning/foo@sha256:80d2d7b58afeaf907451c6f4e24de47b09a327a24a21a2d3323b7abf76d14be5",
    -      "integrity": "sha256:80d2d7b58afeaf907451c6f4e24de47b09a327a24a21a2d3323b7abf76d14be5"
    +    "ghcr.io/codspace/versioning/foo:2": {
    +      "version": "2.11.1",
    +      "resolved": "ghcr.io/codspace/versioning/foo@sha256:e98cdc5066cff85c5076dfec32058d53e5b9bbc75b125d84adcdf295674c14ee",
    +      "integrity": "sha256:e98cdc5066cff85c5076dfec32058d53e5b9bbc75b125d84adcdf295674c14ee"
         }
       }
     }
    \ No newline at end of file
    ~~~
    5 insertions (+), 5 deletions (-)

=== ghcr.io/codspace/versioning/bar:1 (1.0.0)
 => checking for updates 2/3
 => latest available version is 1.0.0

=== ghcr.io/codspace/versioning/baz:1.0 (1.0.0)
 => checking for updates 3/3
 => latest available version is 2.0.0
 => latest allowed version is 2.0.0
Running command: devcontainer upgrade
 => bump ghcr.io/codspace/versioning/baz:1.0 from 1.0.0 to 2.0.0

    ± .devcontainer/devcontainer.json
    ~~~
    --- /tmp/original20231221-316-44zv9n	2023-12-21 17:35:14.339309013 +0000
    +++ /tmp/updated20231221-316-e1khaj	2023-12-21 17:35:14.340309013 +0000
    @@ -3,6 +3,6 @@
         "features": {
             "ghcr.io/codspace/versioning/foo:1": {},
             "ghcr.io/codspace/versioning/bar:1": {},
    -        "ghcr.io/codspace/versioning/baz:1.0": {}
    +        "ghcr.io/codspace/versioning/baz:2": {}
         }
     }
    ~~~
    2 insertions (+), 2 deletions (-)

    ± .devcontainer/devcontainer-lock.json
    ~~~
    --- /tmp/original20231221-316-fk8v4b	2023-12-21 17:35:14.343309013 +0000
    +++ /tmp/updated20231221-316-f439qx	2023-12-21 17:35:14.343309013 +0000
    @@ -5,10 +5,10 @@
           "resolved": "ghcr.io/codspace/versioning/bar@sha256:0eb80a7a45ea6ac6d2057798608be4cacb3d3667d4818118e17acc5037d687d4",
           "integrity": "sha256:0eb80a7a45ea6ac6d2057798608be4cacb3d3667d4818118e17acc5037d687d4"
         },
    -    "ghcr.io/codspace/versioning/baz:1.0": {
    -      "version": "1.0.0",
    -      "resolved": "ghcr.io/codspace/versioning/baz@sha256:37f36051adf6da0a43764b9669b945e0f06e11973e02c57ad261b03bb1057cb7",
    -      "integrity": "sha256:37f36051adf6da0a43764b9669b945e0f06e11973e02c57ad261b03bb1057cb7"
    +    "ghcr.io/codspace/versioning/baz:2": {
    +      "version": "2.0.0",
    +      "resolved": "ghcr.io/codspace/versioning/baz@sha256:3420e9d222352d5bee4d7f2c99dd7492295f5518041650ade9e8ecd0d6ce49d8",
    +      "integrity": "sha256:3420e9d222352d5bee4d7f2c99dd7492295f5518041650ade9e8ecd0d6ce49d8"
         },
         "ghcr.io/codspace/versioning/foo:1": {
           "version": "1.1.0",
    ~~~
    5 insertions (+), 5 deletions (-)
🌍 Total requests made: '0'
```

I saw a few more issues that I'll be addressing next week:

* Don't parse the requirement as part of the dependency name: it should be `ghcr.io/codspace/versioning/foo` not `ghcr.io/codspace/versioning/foo:1`.
* Respect style of existing requirements for now: `ghcr.io/codspace/versioning/baz:1.0` should be updated to `ghcr.io/codspace/versioning/baz:2.0`, not to `ghcr.io/codspace/versioning/baz:2`.

Also found an issue in the CLI that I reported at https://github.com/devcontainers/cli/issues/712.

deivid-rodriguez:
I fixed all the issues mentioned previously and got CI to pass.

However, I was thinking about the smart approach taken here of asking `devcontainers outdated` directly in the `FileParser` step about the latest version and "bypassing" the `UpdateChecker` step and I believe it's going to come short for some common Dependabot features. For example:

* Users need to be able to skip certain versions from configuration. See https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore.
* Similarly (same doc as above), users may choose to ignore certain update types for a dependency, for example, patch and minor versions.

So I think we still need to implement a basic `UpdateChecker` that can access all available versions of the feature and do some filtering over them.

We'll need to work on this after holidays!

deivid-rodriguez:
Alternatively, and probably better since it would only require changes to the devcontainers CLI, we could add an `--ignore-range` flag to the `devcontainers outdated` CLI command to support the first use case, and `--ignore-major`, `--ignore-minor`, and `--ignore-patch` flags to support the latter.

deivid-rodriguez:
Found and fixed a few more issues, including one that I reported [upstream](https://github.com/devcontainers/cli/issues/715) (and added workaround for it here).

Also cleaned up git history to keep things a bit more tidy.

This is ready for a review!

deivid-rodriguez:
@joshspicer I looked at the issues you found:

* For SHA256 versions, I'm ignoring them for now, so we don't even try to update them. If someone is using those, I guess it's their preference and they have some motivation, so better for Dependabot to not be annoying to those people. By the way, I'm doing the same for deprecated features because the CLI doesn't handle them very well at the moment. I looked for a "/" in the feature name to detect non deprecated features, is that correct?
* As per the second issue, after I added support for ignores, it's now working "fine": no error, and it's updating the invalid tag to the latest version. I think that's acceptable.

I also fixed some other issues I found and added support for ignores!

abdulapopoola (approved):

jakecoffman (approved):

deivid-rodriguez:
I merged the first commit in this PR as #8858, so that I can build updater images for this PR (since workflow files are picked up from the main branch).

deivid-rodriguez:
Let's do this!

PR #8446, opened 2023-11-21
Sanitize `.yanrc.yml` when missing environment variables prevent yarn from running

deivid-rodriguez:
Sometimes we fail to run `yarn` because the user has configured an environment variable in their `.yarnrc.yml` file and Dependabot does not support that.

For example,

```
$ bin/dry-run.rb npm_and_yarn "dailypay/jest-circus-allure-environment" --dir="/." --updater-options=grouped_updates_experimental_rules,record_ecosystem_versions,record_update_job_unknown_error --commit=9582ee938e3e4adf96734a4739b201c2d10653f0 --cache=files --dep prettier
```

This PR raises those errors as expected errors.


mctofu (approved):

npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
```diff
@@ -51,12 +51,29 @@ def self.yarn_berry?(yarn_lock)
...
-        @yarn_major_version ||= fetch_yarn_major_version
-      end
-
-      def self.fetch_yarn_major_version
+        retries = T.let(0, Integer)
         output = SharedHelpers.run_shell_command("yarn --version")
         Version.new(output).major
+      rescue Dependabot::SharedHelpers::HelperSubprocessFailed => e
+        # Should never happen, can probably be removed once this settles
+        raise "Failed to replace ENV, not sure why" if retries.positive?
```
mctofu, line 59: One reason might be if there are multiple env vars in the file? Or potentially multiple files with envs?

Could we do a global cleanup of variables in the file to get around the 1st issue?

deivid-rodriguez: My idea was to let this error show up in Sentry and use it to learn about the edge cases. But I did push the fix you suggested to the first edge case!

mctofu (approved):

PR #8468, opened 2023-11-27
Bump `nuget` files type strictness

JamieMagee:
This change bumps the type strictness for the files added in #8179. That PR was started before Sorbet was added to the codebase, so it's expected that they started as `#typed: false`.

The bump was performed automatically with `bundle exec spoom bump --from=false --to=true nuget`.

### DIFF
--- a/nuget/lib/dependabot/nuget/cache_manager.rb
+++ b/nuget/lib/dependabot/nuget/cache_manager.rb
@@ -1,4 +1,4 @@
-# typed: false
+# typed: true
 # frozen_string_literal: true
 
 require "dependabot/file_fetchers"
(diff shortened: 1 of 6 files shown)


deivid-rodriguez:
Oh, nice catch @JamieMagee, thanks!

deivid-rodriguez (approved):

deivid-rodriguez:
Should we be detecting this automatically before merge?

jurre:
> bundle exec spoom bump --from=false --to=true

Yeah maybe we can run `bundle exec spoom bump --from=false --to=true` and assert it outputs nothing in CI?

JamieMagee:
> Should we be detecting this automatically before merge?

It might be possible if you were to run `spoom bump` as part of CI and exit if there are any files listed?

PR #8469, opened 2023-11-27
Run `spoom bump` as part of CI

JamieMagee:
This change should block a PR that doesn't have the appropriate type strictness level set on a file.


deivid-rodriguez (approved):

JamieMagee:
I am not sure why it is trying to bump `bundler/helpers/v1/run.rb` from `true` to `strict` when it clearly doesn't meet the requirements.

JamieMagee:
I worked around it by removing

```ruby
def output(obj)
  print JSON.dump(obj)
end
```

and just inlining it instead. It's a workaround, but IMO a small and acceptable one.

PR #8471, opened 2023-11-27
build(deps): bump pNPM to 8.11.0

yeikel:
Release notes

https://github.com/pnpm/pnpm/releases/tag/v8.11.0

Note to reviewers/users : 

This might lead to some confusion with _larger-than expected diffs_ in the rare situations described in https://github.com/pnpm/pnpm/pull/7318 but it is a necessary change


deivid-rodriguez (approved):

deivid-rodriguez:
I missed your note about that upstream change, but I did see it in the changelog notes. My understanding is that since we use corepack and respect PNPM if it's pinned by users, this should not be an issue and it's easily avoidable if it gets in the middle.

PR #8485, opened 2023-11-29
Fix failing test for NuGet updates

martincostello:
Fix test that fails due to a warning about an out-of-support TFM (`net5.0`) by adding a `NoWarn` for `NETSDK1138`.

I _think_ the difference as to why this fails for me and not in your CI tests is because I have the .NET 7.0.404 SDK installed.


brettfo (approved):

deivid-rodriguez (approved):

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core.Test/Update/UpdateWorkerTests.Sdk.cs
```diff
@@ -1560,6 +1560,7 @@ public async Task PinTransitiveDependencyByAddingPackageVersion()
                 <Project Sdk="Microsoft.NET.Sdk">
 
                   <PropertyGroup>
+                    <NoWarn>$(NoWarn);NETSDK1138</NoWarn>
```
yeikel, line 1563: I don't understand this change. From my little understanding of `Net`, `NoWarn` is used to ignore specific warnings, but is that what the goal is? 

I don't understand how a warning impacts that a test fails

Could you please clarify?

yeikel: Also, if the target is to ignore these, the [documentation suggests 
](https://learn.microsoft.com/en-us/dotnet/core/tools/sdk-errors/netsdk1138)

`<CheckEolTargetFramework>false</CheckEolTargetFramework>` 

Instead to skip the check all together rather than to hide the warnings

deivid-rodriguez: I guess it's asserting some output and the warning gets in the middle? Maybe something expects JSON or a well known format and the warning messes that up? We do have a way to ignore error output when shelling out in Ruby code, but this is native code, so not sure if there's a better more generic solution.

martincostello: So I was taking a look into #8483 and opened the new solution locally and ran the tests, and this one failed. When I looked at the test output it contained this warning and no other obvious output indicating a reason for the failure. I added this suppression, and then it passed.

Maybe that's a red herring, but that was the process I went through. There's definitely _something_ off somewhere.

PR #8486, opened 2023-11-29
Refactor error handling

deivid-rodriguez:
It's not straightforward to know where new errors need to be added, so that they are properly passed back to the server.

Also, error handling is duplicated in dry-run.rb script and updater commands.

This PR tries to improve both of the above things.


jakecoffman (approved):

bin/dry-run.rb
```diff
@@ -402,101 +402,26 @@ def fetch_files(fetcher)
...
   end
 rescue StandardError => e
-  error_details = handle_dependabot_error(error: e)
+  error_details = Dependabot.fetcher_error_details(e)
+  raise unless error_details
 
   puts " => handled error whilst fetching dependencies: #{error_details.fetch(:"error-type")} " \
        "#{error_details.fetch(:"error-detail")}"
 
   []
```
yeikel, line 411: Is there any reason why we cannot `raise` when there are error details? Is that because we only throw when there are unknown exceptions?

Also, rather than `puts`, is there any reason not to use the [Logger](https://github.com/dependabot/dependabot-core/blame/main/common/lib/dependabot/logger.rb) with `error` level?

common/lib/dependabot/errors.rb
```diff
@@ -5,6 +5,210 @@
...
+    when Dependabot::ToolVersionNotSupported
+      {
+        "error-type": "tool_version_not_supported",
+        "error-detail": {
+          "tool-name": error.tool_name,
+          "detected-version": error.detected_version,
+          "supported-versions": error.supported_versions
+        }
+      }
+    when Dependabot::BranchNotFound
```
yeikel, line 20: I think that it would be ideal if each of these exceptions produced these via their constructors and exposed a dedicated method to fetch them. If we had that, we would not need this mapping at all

bin/dry-run.rb
deivid-rodriguez: I followed the same strategy already used for fetching files. Same for using `puts`, I just did what we were already doing in other error handling blocks.

common/lib/dependabot/errors.rb
deivid-rodriguez: That's neat, I think it'd be a cool follow up to this PR!

deivid-rodriguez:
@yeikel I'm merging this because I have an enhancement I want to ship based on this PR (#8326). But your ideas look great, we can implement them as follow ups 👍.

PR #8492, opened 2023-11-30
Create abstract `Dependabot::Requirement` class

JamieMagee:


Nishnha (approved):
Thank you! This implementation makes sense to me. 

I think in a future PR we can try to assert the requirement string is a `T.must(String)`, but we'll also have to add nil checks to the package manager implementations. I also think we will see runtime errors with that change, so it's probably a good idea to do it separately.

PR #8494, opened 2023-11-30
Teach Dependabot how to present multi-directory PRs

Nishnha:
Fix up the title and PR body for mutli-directory PRs so they split the dependency updates by directory.

So https://github.com/dsp-testing/gsu-7/pull/1 would look like

Screenshot 2023-11-30 at 4 19 27 PM

### DIFF
--- a/common/lib/dependabot/pull_request_creator/message_builder.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder.rb
@@ -260,6 +260,8 @@ module Dependabot
       # rubocop:disable Metrics/PerceivedComplexity
       # rubocop:disable Metrics/AbcSize
       def version_commit_message_intro
+        return multi_directory_group_intro if dependency_group && job.source&.directories.count > 1
+
         return group_intro if dependency_group
 
         return multidependency_property_intro if dependencies.count > 1 && updating_a_property?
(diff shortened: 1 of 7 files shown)


jakecoffman (approved):

Nishnha:
This is what it looks like in practice https://github.com/dsp-testing/nish-gsu-7/pull/1

Not exactly what I was going for though - I wanted it to be:

```
"Bumps the go_modules group across 2 directories with 2 updates"

Updates github.com/consensys/gnark-crypto from 0.9.1 to 0.12.1
Updates golang.org/x/text from 0.0.0-20170915032832-14c0d48ead0c to 0.3.8

"Bumps the go_modules group across 2 directories with 2 updates"

Updates github.com/consensys/gnark-crypto from 0.9.1 to 0.12.1
Updates golang.org/x/text from 0.0.0-20170915032832-14c0d48ead0c to 0.3.8
```

But now I realize that if the PR body gets truncated, the summaries will be missing, so maybe it's okay that they're all at the top of the PR body?

I'm going to merge this PR and if we want to change it later we can. We still have to fix the branch name too.

carogalvin:
The summary is duplicated at the top, will that happen every time?

Nishnha:
> The summary is duplicated at the top, will that happen every time?

No it will change depending on the dependencies updated for that directory. I'm going to make a minor change so that the directory is included in the summary, and then I'll share a better example with different dependencies.

PR #8495, opened 2023-11-30
Add fingerprint to dynamic git commands

deivid-rodriguez:
So that errors are grouped in Sentry.

### DIFF
--- a/common/lib/dependabot/workspace/git.rb
+++ b/common/lib/dependabot/workspace/git.rb
@@ -31,7 +31,10 @@ module Dependabot
 
       sig { override.returns(String) }
       def to_patch
-        run_shell_command("git diff --patch #{@initial_head_sha}.. .")
+        run_shell_command(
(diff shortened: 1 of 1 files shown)


common/lib/dependabot/workspace/git.rb
```diff
@@ -124,14 +137,21 @@ def commit(memo = nil)
...
         [head_sha, diff]
       end
 
       sig { params(sha: String).returns(String) }
       def reset(sha)
-        run_shell_command("git reset --hard #{sha}")
+        run_shell_command(
+          "git reset --hard #{sha}",
+          fingerprint: "git reset --hard <sha>"
+        )
```
JamieMagee, line 154: I can't comment below here, but the Sorbet fix is to change `kwargs` on line 164 from `T::Boolean` to `T.any(T::Boolean, String)`

jurre (approved):

PR #8497, opened 2023-11-30
add support for refreshing a grouped security update

jakecoffman:
Forgot to add support for this in #8331


jakecoffman:
I just noticed using the CLI it's creating a new PR instead of updating the existing one because the data is coming as `existing_pull_requests` instead of `existing_group_pull_requests`. This is because currently for grouped security updates the group is ephemeral. 

It's a quick fix in `dependency_change` to look there instead when it's a group, so I'll do that.

honeyankit:
LGTM

PR #8502, opened 2023-12-01
don't attempt to update a package if no versions could be found

brettfo:
This partially fixes the issue:

```
ERROR undefined method `fetch' for nil:NilClass

                                                               ^^^^^^
```

This can happen if querying NuGet for the available package versions returns an empty set and can happen if the package source doesn't contain that package.

The fix is to only report dependencies where at least one of the package repositories reported at least one search result.


abdulapopoola (approved):

nuget/lib/dependabot/nuget/update_checker.rb
```diff
@@ -59,6 +59,9 @@ def up_to_date?
         # is up-to-date
         return true unless requirements_unlocked_or_can_be?
 
+        # If no preferred version can be found then the dependency is up-to-date
+        return true unless preferred_resolvable_version_details
+
```
jakecoffman, line 64: If a dependency can't be found in any of the registries supplied, then I agree with @deivid-rodriguez it should raise an exception that is a "user error" like maybe `DependencyFileNotResolvable` so we can tell the user why their dependency wasn't updated.

If a registry finds the dependency but all versions are filtered out, like it's the latest version, then saying it's up to date is correct. 

Would this suggestion achieve that?

```suggestion
        raise DependencyFileNotResolvable, "#{dependency.name} not found in any registry" if versions.none?

        # If no preferred version can be found then the dependency is up-to-date
        return true unless preferred_resolvable_version_details

```

deivid-rodriguez: Yeah, that's what I thought was happening here, that no versions could be found at all for the dependency.

However, I'm on the fence about raising this as a user error vs an unexpected error. I initially suggested unexpected error because this could be caused by a Dependabot bug like here. But in general we don't expect `FileParser` to validate that the Dependencies it parses exist, so it seems fine to raise a user error. If it turns out this is an invalid dependency, then we'll probably get a bug report about it.

I wonder if something like this should be pulled to common code, since it seems it could apply to all ecosystems. Not something for this PR though!

jakecoffman: I prefer a user error in this case because if this does happen and it's not a bug, there's not anything we can do to fix it and it will pollute our error reporting. 

With a user error, the risk is a bug but we will likely get a report for it, and we can still see elevated levels of specific errors in our metrics which we could track down.

deivid-rodriguez: Yeah, seems fine.

brettfo: I spoke with @deivid-rodriguez offline and here's what we came up with.

It's not uncommon for a package owner to publish a package then immediately de-list it; this might be used to produce dev or daily builds (I've done this in the past).  In this case the query will return nothing, but when the build requests that specific package it _will_ be returned, so if we start raising `DependencyFileNotResolvable` their entire update scenario will be broken.

The fix I'm going to work on is to update the initial dependency scanning and only report a dependency if one of the package repositories returned _something_ for that dependency.  This will make the dependency scanning pretty network heavy in the beginning, but should ensure that we don't report something that we can't find.

brettfo:
@deivid-rodriguez I re-worked this change to only report a `Dependency` if one of the package sources actually contains that package.  Since this moves a bunch of network calls into the file fetcher tests, the bulk of this PR is simply stubbing the search requests.

JoeRobich (approved):
Looks good. If RepositoryFinder understood nuget.configs \<packageSourceMapping\>, it could potentially reduce the number of dependeny urls that had to be checked. Future enhancement maybe.

brettfo:
> Looks good. If RepositoryFinder understood nuget.configs <packageSourceMapping>, it could potentially reduce the number of dependeny urls that had to be checked. Future enhancement maybe.

Good call, that'll improve the other lookups, too.  I'm going to add one or 2 more test scenarios, just to be safe, then I think this is ready for final approval.

brettfo:
Updated to stub out more network calls instead of pulling from a file.  Reduced changed lines from 36k to 300.

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
```diff
@@ -261,12 +266,53 @@ def build_dependency(name, req, version, prop_name, project_file, dev: false)
...
-          Dependency.new(
+          dependency = Dependency.new(
             name: name,
             version: version,
             package_manager: "nuget",
             requirements: [requirement]
           )
+
+          # only include dependency if one of the sources has it
+          return unless dependency_has_search_results?(dependency)
```
jakecoffman, line 277: Consider logging the dependency and why it was skipped. It might be hard to track down why a certain dependency is missing without a log message here.

jakecoffman (approved):
This is a bit different than how most of the other ecosystems do it, but let's give it a shot. I like that the calls are cached so they don't have to be made again during the UpdateChecker step.

jakecoffman:
I deployed this and a new exception popped up:

```
updater | 2023/12/05 13:01:59 ERROR key not found: :search_url
updater | 2023/12/05 13:01:59 ERROR /home/dependabot/nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb:308:in `fetch'
updater | 2023/12/05 13:01:59 ERROR /home/dependabot/nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb:308:in `execute_search_for_dependency_url'
updater | 2023/12/05 13:01:59 ERROR /home/dependabot/nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb:293:in `block in dependency_has_search_results?'
```

Reproducible with: `dependabot update nuget udap-tools/udap-dotnet`

jakecoffman:
Over the last 3 hours the NilClass error has dropped dramatically. Now the KeyError is frequent, but it's an order of magnitude less than the NilClass error was. This is definitely an improvement, I'll go ahead and merge this so we can keep moving forward.

brettfo:
@jakecoffman PR for `key not found: :search_url` here: #8534

PR #8512, opened 2023-12-02
only consider the `Include` attribute when parsing project files

brettfo:
This fixes one specific instance of the issue

```
ERROR undefined method `fetch' for nil:NilClass

                                                               ^^^^^^
```

The problem arises when there is a line like this in a project:

``` xml
...
<ProjectReference Update="PackageThatDoesNotExist" Version="1.2.3" />
...
```

That's perfectly valid MSBuild and at restore/build time results in a noop because there is no dependency with that name (marked with `Include=`) so there is nothing to update, but since that package doesn't exist, trying to later determine the latest version predictably fails because not only is there no latest version, but there is no version at all.

Even if that `<ProjectReference>` node was for a valid package, just the presence of the `Update` attribute isn't enough, because if it wasn't `Include`d to begin with, there is nothing to update so it shouldn't be reported as a dependency.

There is the possibility that just the presence of the `Update` attribute is sufficient to properly fix a dependency version, but that only happens if the package being `Update`d is a transitive dependency of an existing package, and that's a scenario that's now properly covered where an entirely new `<PackageReference Include="" />` line will be added, so updating just the `Update` attribute is superfluous.
 
There is another work item to not even consider packages that don't have any versions that will be addressed later in #8502.

### DIFF
--- a/nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
+++ b/nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
@@ -270,8 +270,7 @@ module Dependabot
         end
 
         def dependency_name(dependency_node, project_file)
-          raw_name = get_attribute_value(dependency_node, "Include") ||
-                     get_attribute_value(dependency_node, "Update")
+          raw_name = get_attribute_value(dependency_node, "Include")
           return unless raw_name
 
           # If the item contains @(ItemGroup) then ignore as it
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):
This makes sense to me in principle but ideally someone else from your team would double check this :)

One suggestion I'd have is to link PRs to open core issues, so they are automatically closed when PR merges, and also provide a CLI/dry-run.rb example of the issue being fixed, so that it's easier to verify/review.

brettfo:
Scenario covered by a better solution in #8502

PR #8524, opened 2023-12-04
Nuget file fetching simplifications

deivid-rodriguez:
I was about to submit my review to https://github.com/dependabot/dependabot-core/pull/8511 with some simplifications, but I got there late. It's best anyways to get have got the smoke tests back to green right aways, and then add whatever iterations, so proposing my suggestions separately!


brettfo (approved):
Should be good, when I originally made these changes I added tests that should catch anything that might be broken.

mctofu (approved):

PR #8525, opened 2023-12-04
clean up smoke.yml

jakecoffman:
I'm trying to come up with a way to discover smoke tests dynamically and run them in dependabot-core. While doing that I realized there was a simpler way to do the filters. So I thought I would ship this PR first to clean things up and make sure it worked.

The idea is that if any of the "common" files change we need to run all smoke tests, so there's no reason to list every test out in the filters, just the directories in core that changed. It gets rid of a lot of duplication.

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -23,6 +23,9 @@ jobs:
       fail-fast: false
       matrix:
         suite:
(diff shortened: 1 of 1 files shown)


honeyankit (approved):
This is nice! Thank you for cleaning it up!

deivid-rodriguez:
Yeah, nice cleanup!

PR #8529, opened 2023-12-04
Create feature flag for Grouped security updates

ryanbrandenburg:
This creates a feature flag for Grouped security updates.


updater/lib/dependabot/updater/operations/create_group_security_update_pull_request.rb
```diff
@@ -15,6 +15,7 @@ class CreateGroupSecurityUpdatePullRequest
         include GroupUpdateCreation
 
         def self.applies_to?(job:)
+          return false if Dependabot::Experiments.enabled?(:grouped_security_updates_disabled)
```
ryanbrandenburg, line 18: I recognize that it's a bit weird to phrase it like this, but I need the default case to be grouped_security_updates being on and the way Dependabot::Experiments works is it's falsy by default.

updater/spec/dependabot/integration_spec.rb
```diff
@@ -209,8 +209,12 @@
     context "when there is an exception that blocks PR creation (cloud)" do
       before do
         allow(api_client).to receive(:create_pull_request).and_raise(StandardError, "oh no!")
-        allow(Dependabot::Experiments).to receive(:enabled?).with(:record_ecosystem_versions).and_return(true)
```
ryanbrandenburg, line 212: As far as I can tell doing it this way really only worked while we had a single call against `enabled?`. Since we already had the reset! method I figured we could make use of it instead of mocking things so that later levels of abstraction have an easier time handling things.

updater/spec/dependabot/updater/error_handler_spec.rb
```diff
@@ -232,10 +236,6 @@
         end
       end
 
-      before do
-        allow(Dependabot::Experiments).to receive(:enabled?).with(:record_update_job_unknown_error).and_return(false)
```
ryanbrandenburg, line 236: These explicit mocks were not needed once we stop mocking (default to false), so I deleted them.

jakecoffman (approved):
Thanks!

PR #8535, opened 2023-12-05
fix group update creation failure when a dependency is ignored

jakecoffman:
When checking to see if all versions are ignored, grouped security updates need to call `lowest_security_fix_version`, otherwise it will continue and throw AllVersionsIgnored which halts the update.


Nishnha (approved):
Good catch!

PR #8536, opened 2023-12-05
only run suites that have changes

jakecoffman:
Another step towards dynamic discovery of smoke tests.

In this PR I split the filter and matrix out into separate files. The filter was just for vanity, makes it much easier to read the workflow. Having the matrix suites in a file allows us to run the dorny/paths-filter action first and then filter the matrix.json to only tests that needs to run. 

In other words, going forward when you change a file in `go_modules`, it won't run the `bundler` smoke test only to skip all steps. Instead it won't even run in the first place!

### DIFF
--- a/.github/smoke-filters.yml
+++ b/.github/smoke-filters.yml
@@ -1,0 +1,40 @@
+common:
+  - .github/workflows/smoke.yml
+  - .dockerignore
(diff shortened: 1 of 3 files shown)


deivid-rodriguez (approved):
Nice! Sounds like tricky `jq` trickery!

PR #8541, opened 2023-12-06
Teach updater how to do multi-directory version pull requests

brbayes-msft:


jakecoffman:
Would you mind adding a [smoke test](https://github.com/dependabot/smoke-tests#how-to-add-new-tests) for both of these? If you want to copy existing security multi-dir test and make it a version update, that would be enough.

brbayes-msft:
> Would you mind adding a [smoke test](https://github.com/dependabot/smoke-tests#how-to-add-new-tests) for both of these? If you want to copy existing security multi-dir test and make it a version update, that would be enough.

Smoke tests have been added: https://github.com/dependabot/smoke-tests/pull/156

updater/lib/dependabot/dependency_change.rb
```diff
@@ -81,7 +81,7 @@ def should_replace_existing_pr?
       # NOTE: Gradle, Maven and Nuget dependency names can be case-insensitive
       # and the dependency name injected from a security advisory often doesn't
       # match what users have specified in their manifest.
-      updated_dependencies.map { |x| x.name.downcase } != job.dependencies.map(&:downcase)
+      updated_dependencies.map { |x| x.name.downcase }.uniq.sort != job.dependencies.map(&:downcase).uniq.sort
```
jakecoffman, line 84: Could you add a spec for this change? I have a feeling we'll be revisiting it again would be good to lock it down with a test.

brbayes-msft: I added specs to cover this code.  Let me know if there's a better format to style them.

jzabroski:
@brbayes-msft Is https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/customizing-dependency-updates#grouping-dependabot-version-updates-into-one-pull-request updated with an example for how this works?

brbayes-msft:
> @brbayes-msft Is https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/customizing-dependency-updates#grouping-dependabot-version-updates-into-one-pull-request updated with an example for how this works?

Not at this time.  This pull request makes the requisite updater changes, but additional work has to be done beyond the updater before this could be enabled by a user.

PR #8546, opened 2023-12-06
Don't escape dependency names in tarball URLs since it doesn't always work

### LINKED ISSUE #8544: update private packages from npmjs suddenly stop working

### Package ecosystem

npm

### Package manager version

9.8.1

### Language version

18.18.2

_No response_

### dependabot.yml content

`version: 2
registries:
  npm-npmjs:
    type: npm-registry
    url: https://registry.npmjs.org
    token: ${{secrets.NPM_SECRET}
updates:
  - package-ecosystem: 'npm'
    directory: '/'
    registries:
      - npm-npmjs
    schedule:
      interval: 'daily'
      time: '04:30'
    open-pull-requests-limit: 20`

### Updated dependency

_No response_

Suddenly it stopped working for the dependencies from our private organization dependencies in npmjs.

auth token is fine, packages are fine, local install works, no config changes but from one day to another this week it stopped working.

` proxy | 2023/12/06 08:42:00 [110] GET https://registry.npmjs.org:443/@myorga%2Fmypackage
  proxy | 2023/12/06 08:42:00 [110] * authenticating npm registry request (host: registry.npmjs.org, token auth)
  proxy | 2023/12/06 08:42:01 [110] 200 https://registry.npmjs.org:443/@myorga%2Fmypackage
  proxy | 2023/12/06 08:42:01 [112] HEAD https://registry.npmjs.org:443/@myorga%2Fmypackaget/-/mypackage-0.385.0.tgz
  proxy | 2023/12/06 08:42:01 [112] * authenticating npm registry request (host: registry.npmjs.org, token auth)

  proxy | 2023/12/06 08:42:02 [112] 404 https://registry.npmjs.org:443/@myorga%2Fmypackage/-/mypackage-0.385.0.tgz`
  
 auth seems fine but get 404 while try downloading

_No response_

_No response_

_No response_

deivid-rodriguez:
~Apparently a `HEAD` request is incorrectly returning 404 sometimes. Downloading the full tgz may be heavy, but I think it won't happen often and correctness should come before speed. We can iterate later if this fixes the problem.~

After testing with OP, it was found that it was the %2F in tarball URLs that was causing this, so this PR was changed to not escape dependency names when building tarball URLs for yanked detection.  

Closes #8544.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/registry_finder.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/registry_finder.rb
@@ -39,13 +39,14 @@ module Dependabot
         end
 
         def dependency_url
-          "#{registry_url.gsub(%r{/+$}, '')}/#{escaped_dependency_name}"
+          "#{registry_url}/#{escaped_dependency_name}"
         end
 
         def tarball_url(version)
           version_without_build_metadata = version.to_s.gsub(/\+.*/, "")
 
-          "#{dependency_url}/-/#{scopeless_name}-#{version_without_build_metadata}.tgz"
+          # Dependency name needs to be unescaped since tarball URLs don't always work with escaped slashes
+          "#{registry_url}/#{dependency.name}/-/#{scopeless_name}-#{version_without_build_metadata}.tgz"
         end
 
         def self.central_registry?(registry)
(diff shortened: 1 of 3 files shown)


jakecoffman:
If I am understanding correctly, Dependabot does the listing of versions, and then after does a HEAD on the tarball to see if it's yanked, e.g.:

```
  proxy | 2023/12/06 14:54:16 [026] GET https://registry.npmjs.org:443/axios
  proxy | 2023/12/06 14:54:17 [026] 200 https://registry.npmjs.org:443/axios
  proxy | 2023/12/06 14:54:17 [028] HEAD https://registry.npmjs.org:443/axios/-/axios-1.4.0.tgz
  proxy | 2023/12/06 14:54:17 [028] 200 https://registry.npmjs.org:443/axios/-/axios-1.4.0.tgz
```

I'm curious, if a version was yanked, wouldn't it not appear in the listing in the first place?

deivid-rodriguez:
See https://github.com/dependabot/dependabot-core/pull/8489 for context. The unfortunate reality is no, yanked versions seems to appear in the listing, at least sometimes, so my conclusion was that the only reliable way to figure out whether a package is yanked is to directly check if it's there.

PR #8548, opened 2023-12-06
Bump golang from 1.21.4-bookworm to 1.21.5-bookworm in /go_modules

TomSellers:
Similar to PR #8374. This allows Dependabot to function when used against Go projects that require Go 1.21.5


jakecoffman (approved):
Thanks

PR #8551, opened 2023-12-06
dynamically discover smoke tests

jakecoffman:
When adding a new smoke test, we had to remember to also add it to this list of tests, which inevitably means we forgot.

This change pulls the list of tests from the smoke test so as long as it's on main it will be run!

It relies on convention that the "test" value in smoke-matrix.json matches the test name like so: `smoke-$test-.yaml`.

So I need to rename a few tests after merging this.

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -52,7 +52,7 @@ jobs:
           jq -c '.[]' filtered.json | while read -r i; do
             test=$(echo "$i" | jq -r '.test')
             jq --argjson i "$i" --arg test "$test" -r '.[] | select(.name | contains($test)) | {core: ($i | .core), ecosystem: ($i | .ecosystem), name: .name}' tests.json
-          done | jq -s . > suites.json
+          done | jq -cs . > suites.json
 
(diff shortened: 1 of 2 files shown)


jakecoffman:
Oh looks like there's a problem, the filter needs to be more strict: `go_modules, gomod, smoke-cargo.yaml`

jakecoffman:
It works!

The smoke test failures are expected due to a recent change in the PR body.

There's a few tests missing which will be fixed by https://github.com/dependabot/smoke-tests/pull/153

honeyankit (approved):

.github/workflows/smoke.yml
```diff
@@ -43,8 +43,19 @@ jobs:
           jq -c --argjson changes "$(cat changes.json)" '[.[] | select(.core as $p | $changes | index($p))]' .github/smoke-matrix.json > filtered.json
           cat filtered.json
 
+          # Curl the smoke-test tests directory to get a list of tests to run
```
honeyankit, line 46: ❤️

honeyankit (approved):

Issue #8505, opened 2023-12-01
Error updating NuGet - Package does not exist as a dependency


eduherminio:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

NuGet

### Language version

.NET 8

### Manifest location and content before the Dependabot update

https://github.com/eduherminio/AdventOfCode.Template/blob/main/AdventOfCode/AdventOfCode.csproj

### dependabot.yml content

https://github.com/eduherminio/AdventOfCode.Template/blob/main/.github/dependabot.yml

### Updated dependency

https://www.nuget.org/packages/AoCHelper from v3.0.0 to v3.1.0

### What you expected to see, versus what you actually saw

Please note that **last change in the repository was also performed by Dependabot, so no changes in configuration have taken place since then**

Dependabot is failing to update the single dependency he's monitoring in my repo

![image](https://github.com/dependabot/dependabot-core/assets/11148519/1fff1e31-a498-42e8-ab7b-98b557a0410e)

### Native package manager behavior

New packaged is restored correctly, as expected

### Images of the diff or a link to the PR, issue, or logs

<details>
<Summary>Dependabot logs</Summary>

```bash

  proxy | 2023/12/01 13:37:04 proxy starting, commit: 6cffd6fae1b2f713f2d837bc45fe916f855c821d
  proxy | 2023/12/01 13:37:04 Listening (:1080)
updater | 2023-12-01T13:37:04.666055833 [757147332:main:WARN:src/devices/src/legacy/serial.rs:222] Detached the serial input due to peer close/error.
updater | time="2023-12-01T13:37:06Z" level=info msg="guest starting" commit=17df297b9449ef5936111b658329653bcfc0c9bc
updater | time="2023-12-01T13:37:06Z" level=info msg="starting job..." fetcher_timeout=10m0s job_id=757147332 updater_timeout=45m0s updater_version=650e132a94478a92e282498de9510fb8396960cc-nuget
updater | 2023/12/01 13:37:08 INFO Raven 3.1.2 ready to catch errors
updater | 2023/12/01 13:37:09 INFO <job_757147332> Starting job processing
  proxy | 2023/12/01 13:37:09 [002] GET https://github.com:443/eduherminio/AdventOfCode.Template/info/refs?service=git-upload-pack
  proxy | 2023/12/01 13:37:09 [002] * authenticating git server request (host: github.com)
[...]

PR #8562, opened 2023-12-07
[NuGet] add dotnet 8.0 support

### LINKED ISSUE #8530: error that 7.0 .NET SDK is being used, not 8.0

### Package ecosystem

nuget

### Package manager version

whatever is in the 8.0 SDK

### Language version

_No response_

various in https://github.com/danmoseley/eShop/tree/dependabot

### dependabot.yml content

https://github.com/danmoseley/eShop/blob/dependabot/.github/dependabot.yml

### Updated dependency

_No response_

log: https://github.com/danmoseley/eShop/network/updates/758220044 

errors appear to start ere

```
updater |   Updating global.json files.
updater |     Dependency [Duende.IdentityServer.AspNetIdentity] not found in any global.json files.
updater |   No dotnet-tools.json files found.
updater | Running for project [/home/dependabot/dependabot-updater/repo/src/eShop.AppHost/eShop.AppHost.csproj]
updater |   Running for SDK-style project
updater | dotnet build in GetAllPackageDependenciesAsync failed. STDOUT: MSBuild version 17.6.3+07e294721 for .NET
updater |   Determining projects to restore...
updater | /usr/local/dotnet/current/sdk/7.0.305/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.TargetFrameworkInference.targets(151,5): error NETSDK1045: The current .NET SDK does not support targeting .NET 8.0.  Either target .NET 7.0 or lower, or use a version of the .NET SDK that supports .NET 8.0. [/tmp/package-dependency-resolution_66Ggzt/Project.csproj]
```

The .NET 8.0.100 SDK is shipped 3 weeks ago. As far as I know, my repo is calling for it -- the global.json has
```json
{
  "sdk": {
    "version": "8.0.100",
    "rollForward": "latestPatch",
    "allowPrerelease": true
  }
}
```

Is Dependabot using the 7.0 SDK still? That would cause it to fail when examining projects that target 8.0.

_No response_

_No response_

_No response_

brettfo:
Update the NuGet updater to `net8.0` and also allow updating projects with that `TargetFramework`.  Some project content refactoring was done to make future upgrades easier via `$(CommonTargetFramework)`.

I also did some cleanup by forcing each project's `bin/` and `obj/` directory into `artifacts/` at the root of the C# helpers.  This is to make local development easier because I commonly build/rebuild on Windows and WSL and that can cause conflicting changes in the intermediate directories and I didn't like manually clearing them out one project at a time so I put everything under one place to make deletion easier.  This is also inline with what the github.com/dotnet org does with all of their repos.

## **There is a caveat with redirecting to the `artifacts/` directory.**

If a user has previously built the C# NuGet updater tool and then update their sources and do a re-build, there will be multiple build failures because their machine will still contain the old `bin/` and `obj/` directories in each project.  The fix is to do a one-time clean and they'll be good to go again.  This issue will not affect new clones or CI.
 
Fixes #8530 and #8505.

### DIFF
--- a/nuget/helpers/lib/NuGetUpdater/Directory.Common.props
+++ b/nuget/helpers/lib/NuGetUpdater/Directory.Common.props
@@ -1,0 +1,16 @@
+<Project>
+  <PropertyGroup>
+    <!--
+      When updating the common target framework below, you will also need to:
+
+      1. Copy the latest `SupportedFrameworks.cs` from
+            https://github.com/NuGet/NuGetGallery/blob/main/src/NuGetGallery.Core/Frameworks/SupportedFrameworks.cs
+         to
+            NuGetUpdater\NuGetUpdater.Core\FrameworkChecker\SupportedFrameworks.cs
+      2. Update tests as needed at `NuGetUpdater\NuGetUpdater.Core.Test\FrameworkChecker\CompatibilityCheckerFacts.cs`
+    -->
+    <CommonTargetFramework>net8.0</CommonTargetFramework>
+    <UseArtifactsOutput>true</UseArtifactsOutput>
+    <ArtifactsPath>$(MSBuildThisFileDirectory)artifacts</ArtifactsPath>
+  </PropertyGroup>
+</Project>
--- a/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/FrameworkChecker/SupportedFrameworks.cs
+++ b/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/FrameworkChecker/SupportedFrameworks.cs
@@ -3,148 +3,144 @@
 
 using System;
 using System.Collections.Generic;
-
 using NuGet.Frameworks;
-
 using static NuGet.Frameworks.FrameworkConstants;
 using static NuGet.Frameworks.FrameworkConstants.CommonFrameworks;
 
-namespace NuGetUpdater.Core.FrameworkChecker;
-
-/// <summary>
-/// This class contains documented supported frameworks.
-/// </summary>
-/// <remarks>
-/// All these frameworks were retrieved from the following sources:
-/// dotnet documentation: https://docs.microsoft.com/en-us/dotnet/standard/frameworks.
-/// nuget documentation: https://docs.microsoft.com/en-us/nuget/reference/target-frameworks.
-/// nuget client FrameworkConstants.CommonFrameworks: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Frameworks/FrameworkConstants.cs.
-/// Deprecated frameworks are not included on the list https://docs.microsoft.com/en-us/dotnet/standard/frameworks#deprecated-target-frameworks.
-/// </remarks>
-public static class SupportedFrameworks
+namespace NuGetGallery.Frameworks
 {
-    public static readonly Version Version8 = new Version(8, 0, 0, 0);
-
-    public static readonly NuGetFramework MonoAndroid = new NuGetFramework(FrameworkIdentifiers.MonoAndroid, EmptyVersion);
-    public static readonly NuGetFramework MonoTouch = new NuGetFramework(FrameworkIdentifiers.MonoTouch, EmptyVersion);
-    public static readonly NuGetFramework MonoMac = new NuGetFramework(FrameworkIdentifiers.MonoMac, EmptyVersion);
-    public static readonly NuGetFramework Net3 = new NuGetFramework(FrameworkIdentifiers.Net, new Version(3, 0, 0, 0));
-    public static readonly NuGetFramework Net48 = new NuGetFramework(FrameworkIdentifiers.Net, new Version(4, 8, 0, 0));
-    public static readonly NuGetFramework Net481 = new NuGetFramework(FrameworkIdentifiers.Net, new Version(4, 8, 1, 0));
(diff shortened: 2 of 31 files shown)


nuget/helpers/lib/NuGetUpdater/Directory.Common.props
```diff
@@ -0,0 +1,8 @@
+<Project>
+  <PropertyGroup>
+    <CommonTargetFramework>net8.0</CommonTargetFramework>
+    <ArtifactsDir>$(MSBuildThisFileDirectory)artifacts</ArtifactsDir>
+    <BaseOutputPath>$(ArtifactsDir)\bin\$(MSBuildProjectName)</BaseOutputPath>
+    <BaseIntermediateOutputPath>$(ArtifactsDir)\obj\$(MSBuildProjectName)</BaseIntermediateOutputPath>
```
martincostello, line 6: You can get this sort of behaviour with just the new `UseArtifactsOutput` property added in the .NET 8 SDK: https://learn.microsoft.com/dotnet/core/sdk/artifacts-output

brettfo: I didn't know this existed, thank you!

nuget/helpers/lib/NuGetUpdater/Directory.Common.props
```diff
@@ -0,0 +1,8 @@
+<Project>
+  <PropertyGroup>
+    <CommonTargetFramework>net8.0</CommonTargetFramework>
```
JoeRobich, line 3: Can we add a note that when updating this property, we should also ensure that we also have updated the [SupportedFrameworks](https://github.com/dependabot/dependabot-core/blob/main/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/FrameworkChecker/SupportedFrameworks.cs) used be the framework compatibility check? It should at least follow the changes made to [NuGetGallery](https://github.com/NuGet/NuGetGallery/blob/main/src/NuGetGallery.Core/Frameworks/SupportedFrameworks.cs), although ideally be ahead.

JoeRobich (approved):
Changes look good to me. 

Thinking ahead to net9 preview support. Would it be possible to either preload a .net9 [validated or preview SDK](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script#:~:text=for%20public%20use.-,preview,-%3A%20The%20monthly%20public) when building the Dockerfile or install a local version which matches the [global.json when a version higher than the baked in version is pinned](https://github.com/dotnet/aspnetcore/blob/main/global.json)? Also, should we get ahead of net9 packages and add them to our SupportedFrameworks file?

nuget/helpers/lib/NuGetUpdater/Directory.Common.props
brettfo: Good catch, I missed updating the supported frameworks check.  I'll update that and add a comment.

brettfo:
Force-pushed with feedback.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/FrameworkChecker/SupportedFrameworks.cs
```diff
@@ -3,148 +3,144 @@
...
+        public static readonly NuGetFramework Net70Android = new NuGetFramework(FrameworkIdentifiers.NetCoreApp, Version7, "android", EmptyVersion);
+        public static readonly NuGetFramework Net70Ios = new NuGetFramework(FrameworkIdentifiers.NetCoreApp, Version7, "ios", EmptyVersion);
+        public static readonly NuGetFramework Net70MacOs = new NuGetFramework(FrameworkIdentifiers.NetCoreApp, Version7, "macos", EmptyVersion);
+        public static readonly NuGetFramework Net70MacCatalyst = new NuGetFramework(FrameworkIdentifiers.NetCoreApp, Version7, "maccatalyst", EmptyVersion);
+        public static readonly NuGetFramework Net70TvOs = new NuGetFramework(FrameworkIdentifiers.NetCoreApp, Version7, "tvos", EmptyVersion);
+        public static readonly NuGetFramework Net70Windows = new NuGetFramework(FrameworkIdentifiers.NetCoreApp, Version7, "windows", EmptyVersion);
+        public static readonly NuGetFramework Net80 = new NuGetFramework(FrameworkIdentifiers.NetCoreApp, Version8); // https://github.com/NuGet/Engineering/issues/5112
```
yeikel, line 45: At this issue is behind a private repository, it would be helpful for people like me if we had an explanation stating why this is here/what is the problem with this assignment

martincostello:
> Installing an SDK at runtime would be a pretty heavy operation, but I don't see a problem with keeping the Dockerfile up to date with public releases, like `-preview1`, `-rc1`, etc. I don't think it's possible for us to be able to support every internal/daily build, but the few major releases should be pretty easy.

You could potentially do something similar [to this](https://github.com/martincostello/alexa-london-travel/blob/d9bf6a43cc4e5b859e2d27b7730bebde9ca3319d/Dockerfile#L11) where if the repository has a `global.json` file and the SDK version in it isn't the same as the one already installed in the Dockerfile, then pull down the right version using the .NET installation scripts. AFAIK those work with any build of the SDK, even daily ones, so would be the most flexible.

```dockerfile
RUN DOTNET_INSTALL_DIR="/usr/share/dotnet" && curl -sSL https://dot.net/v1/dotnet-install.sh | bash /dev/stdin --jsonfile global.json --install-dir $DOTNET_INSTALL_DIR
```

Issue #8573, opened 2023-12-08
Dependabot throws `dependency_file_not_resolvable` related to GitHub Packages NuGet feed


austindrenski:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

NuGet

### Language version

dotnet 8.0

### Manifest location and content before the Dependabot update

```console
$ tree
.
├── Directory.Build.props
├── Directory.Packages.props
├── SomeSolution.sln
├── global.json
└── src
    ├── Directory.Build.props
    └── SomeProject
        └── SomeProject.csproj
└── tests
    ├── Directory.Build.props
    └── SomeProject.Tests
        └── SomeProject.Tests.csproj

3 directories, 10 files
```

```csproj
<!-- Directory.Build.props -->
<Project>

  <PropertyGroup>
    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
  </PropertyGroup>

</Project>
```

```csproj
<!-- Directory.Packages.props -->
<Project>

  <ItemGroup>
    <PackageVersion Include="AWSSDK.Core" Version="3.7.300.16" />
    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.8.0" />
    <PackageVersion Include="xunit" Version="2.6.2" />
    <PackageVersion Include="xunit.runner.visualstudio" Version="2.5.4" />
  </ItemGroup>

</Project>
```

```csproj
<!-- src/Directory.Packages.props -->
<Project>

  <Import Project="../Directory.Build.props" />

</Project>
```

```csproj
<!-- src/SomeProject/SomeProject.csproj -->
<Project Sdk="Microsoft.NET.Sdk">

  <ItemGroup>
    <PackageReference Include="AWSSDK.Core" />
  </ItemGroup>

</Project>
```

```csproj
<!-- tests/Directory.Packages.props -->
<Project>

  <Import Project="../Directory.Build.props" />

  <ItemGroup>
    <PackageReference Include="Microsoft.NET.Test.Sdk" />
    <PackageReference Include="xunit" />
    <PackageReference Include="xunit.runner.visualstudio" />
  </ItemGroup>

</Project>
```

```csproj
<!-- tests/SomeProject.Tests/SomeProject.Tests.csproj -->
<Project Sdk="Microsoft.NET.Sdk">

  <ItemGroup>
    <ProjectReference Include="../../src/SomeProject/SomeProject.csproj" />
  </ItemGroup>

[...]

PR #8575, opened 2023-12-08
resolve nupkg download link from repository base address and type

brettfo:
Previously the shape of the repository's URL was used to figure out the `.nupkg` download URL (e.g., looking for the string `"/v3"`, etc.)  This PR checks the `:repository_type` field instead, then properly queries the `PackageBaseAddress` parameter in v3 APIs and the `base` attribute in v2 APIs to construct the proper download URL.  In either case if the base address cannot be determined, only then do we fall back to what we had before of trying to divine it.

Should fix the URL generation with #8573, but there may be some underlying auth issues that are not being checked here.


nuget/lib/dependabot/nuget/update_checker/nupkg_fetcher.rb
```diff
@@ -19,23 +19,31 @@ def self.fetch_nupkg_buffer(dependency_urls, package_id, package_version)
...
+                        elsif repository_type == "v3"
+                          get_nuget_v3_package_url(repository_details, package_id, package_version)
                         else
                           raise Dependabot::DependencyFileNotResolvable, "Unexpected NuGet feed format: #{feed_url}"
                         end
 
+          package_url
+        end
+
+        def self.fetch_nupkg_buffer_from_repository(repository_details, package_id, package_version)
```
JoeRobich, line 42: I like that you broke this bit out.

JoeRobich (approved):

brettfo:
Dismissed CodeQL alert because that line only differs in whitespace and isn't relevant to this PR.

jakecoffman (approved):

PR #8594, opened 2023-12-11
Terraform supports having an optional prefix of 'v' for version

BlakeBosin:
Example of this would be v1.0.0 and this is also equal to 1.0.0.  The RubyGems Version Class will throw an exception when given a string of "v1.0.0" stating that the version is malformed.  This PR extends terraform/lib/dependabot/terraform/version.rb to allow a versions with a string containing "v" and removing it via regex if present.  Terraform Docs can be found here for versioning: https://developer.hashicorp.com/terraform/registry/modules/publish#releasing-new-versions.

### DIFF
--- a/terraform/lib/dependabot/terraform/version.rb
+++ b/terraform/lib/dependabot/terraform/version.rb
@@ -22,7 +22,7 @@ module Dependabot
         # See terraform specs: https://developer.hashicorp.com/terraform/registry/modules/publish#releasing-new-versions
         @version_string = @version_string.gsub(/^v/, "")
 
-        super(@version_string)
+        super
       end
 
       sig { override.params(version: VersionParameter).returns(Dependabot::Terraform::Version) }
--- a/terraform/lib/dependabot/terraform/version.rb
+++ b/terraform/lib/dependabot/terraform/version.rb
@@ -22,7 +22,7 @@ module Dependabot
         # See terraform specs: https://developer.hashicorp.com/terraform/registry/modules/publish#releasing-new-versions
         @version_string = @version_string.gsub(/^v/, "")
 
-        super
+        super(@version_string)
       end
 
       sig { override.params(version: VersionParameter).returns(Dependabot::Terraform::Version) }
--- a/terraform/lib/dependabot/terraform/version.rb
+++ b/terraform/lib/dependabot/terraform/version.rb
@@ -19,7 +19,9 @@ module Dependabot
       sig { override.params(version: VersionParameter).void }
       def initialize(version)
         @version_string = T.let(version.to_s, String)
-        super
+        @version_string = @version_string.gsub(/^v/, '')
+
+        super(@version_string)
       end
 
       sig { override.params(version: VersionParameter).returns(Dependabot::Terraform::Version) }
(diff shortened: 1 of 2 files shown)


abdulapopoola (dismissed):

BlakeBosin:
Hello @abdulapopoola and @deivid-rodriguez, do ether of you have additional questions on the PR?  My guess is most people do not include the letter "v" in there terraform module release version.  If this will take a while longer we will most likely start the process to start yanking our old modules and republish new ones.   That said I still think theres value to the PR as it will allow the "v" to be optional and work if its present, which based on the terraform spec is allowed.

abdulapopoola:
Thanks a lot @BlakeBosin for this and sorry about the delay; I can act on getting this rolled out this week.

terraform/lib/dependabot/terraform/version.rb
```diff
@@ -30,7 +30,9 @@ class Version < Dependabot::Version
       end
       def initialize(version)
         @version_string = T.let(version.to_s, String)
-        super
+        @version_string = @version_string.gsub(/^v/, "")
```
abdulapopoola, line 33: @BlakeBosin could you please add a link to the 
```suggestion
       # See terraform specs: https://developer.hashicorp.com/terraform/registry/modules/publish#releasing-new-versions
        @version_string = @version_string.gsub(/^v/, "")
```

BlakeBosin: tooke your suggestion on the comment.

i did run the test without passing @version_string but it will be required because when super gets called without a vars being passed it assumes "version" would be passed but we modified "version" input to make the version without "v" if it was passed in.  See failure from test below.  So because we are modifying version passed in and maintaining the original value that was passed in "@version_string" replaces "version".

`Failures:

  1) Dependabot::Terraform::Version#to_s allow optional v on release 
     Failure/Error: super

     ArgumentError:
       Malformed version number string v1.0.0`

BlakeBosin: @abdulapopoola - let me know if you would like other changes.  ill try to watch my email over the next few days.  thank you for your current responses on this.

jeffwidman (approved):
This looks good to me and thanks for incorporating the code comment - very helpful for the next engineer.

I've asked @abdulapopoola to take a quick look to ensure the rest of his question was resolved.

Sorry for the couple of month delay on this, and feel free to tag myself or Abdul if we get distracted and don't get back to you on this... normally I'm not a fan of tagging maintainers for unresolved questions--most of the time there's a reason we're waiting on something--but this one is pretty straightforward...

terraform/lib/dependabot/terraform/version.rb
jeffwidman: I talked with @abdulapopoola and sounds like it's resolved from his point of view.

jeffwidman:
@abdulapopoola let me know via slack this is good to go from his perspective.

I see a linting error now, but I'm not actually sure we want that `RSpec/ContextWording` linter... let me try to dig into that a little with the team.

sachin-sandhu:
HI @BlakeBosin , thank you for the contribution!!, We are merging the feature requests now , I see some conflicts in files, can you please resolve from your side so that we can take up the latest changes

sachin-sandhu:
@BlakeBosin , I reviewed the conflicting code. it looks like this functionality was added by https://github.com/dependabot/dependabot-core/pull/10222 recently. We are closing the PR , Please feel free to suggest any improvements . Thanks a lot for contributing to dependabot. 

cc @abdulapopoola , @jeffwidman , @amazimbe

PR #8595, opened 2023-12-11
Fix docker attempting to append digest for helm

brbayes-msft:
When attempting to append a digest for helm charts, the docker updater is failing because the `.append` method does not exist.  This corrects that behavior to use the proper append syntax.


deivid-rodriguez (approved):
😮 Where did I get this from? Thanks for fixing my mistake!

abdulapopoola (approved):

PR #8604, opened 2023-12-13
break out the ci paths filter to a file

jakecoffman:
This makes it easier to read the workflow, and I used an anchor to reduce duplication. Makes it easier to spot mistakes. 

I also noticed that our CI was running gradle when maven changes, this is due to some code sharing. So I had our smoke test do the same.


.github/ci-filters.yml
```diff
@@ -0,0 +1,62 @@
...
+  - 'common/**'
+  - 'updater/Gemfil*'
+  - 'omnibus/**'
+  - '.github/workflows/ci.yml'
+bundler:
+  - *shared
+  - 'bundler/**'
+cargo:
+  - *shared
+  - 'cargo/**'
```
landongrindheim, line 13: Alphabetized for easy reading/searching 😍

landongrindheim (approved):

PR #8557, opened 2023-12-07
Strictly type `BranchNamer::Base`

JamieMagee:
This can be marked as `strong` once https://github.com/sorbet/sorbet/pull/7523 is merged.

PR #8606, opened 2023-12-13
fix Sorbet type error

jakecoffman:
The last commit on main is green, but this started failing on branches. Something must not be pinned correctly in a Gemfile or gemspec somewhere.


jakecoffman:
Nice, there's a flaky test as well. I'll have to fix it here too since the type check and flaky test are both required.

landongrindheim (approved):

jakecoffman:
@deivid-rodriguez FYI I marked the failing test pending for now.

JamieMagee:
This is likely related to a fix I contributed upstream: https://github.com/sorbet/sorbet/pull/7523

It was needed for #8557

PR #8611, opened 2023-12-14
consider shrinkwrap when inferring an npmrc file

jakecoffman:
npm-shrinkwrap.json may be in a repository without a lockfile, so Dependabot needs to look at it to infer a private registry in the npmrc file.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npmrc_builder.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npmrc_builder.rb
@@ -151,18 +151,13 @@ module Dependabot
             return @dependency_urls
           end
 
-          if package_lock
+          npm_lockfile = package_lock || shrinkwrap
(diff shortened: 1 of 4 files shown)


npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npmrc_builder.rb
```diff
@@ -155,6 +157,12 @@ def dependency_urls
                           .flatten
                           .select { |url| url.is_a?(String) }
                           .reject { |url| url.start_with?("git") }
+          elsif shrinkwrap
+            @dependency_urls +=
+              shrinkwrap.content.scan(/"resolved"\s*:\s*"(.*)"/)
```
abdulapopoola, line 162: seems to be the same code pattern as the preceding block; an opportunity to consolidate?

abdulapopoola (approved):

jurre (approved):

PR #8619, opened 2023-12-15
Bump Bundler to 2.5

deivid-rodriguez:


Dockerfile.updater-core
```diff
@@ -109,7 +109,7 @@ WORKDIR $DEPENDABOT_HOME/dependabot-updater
 # When bumping Bundler, need to also regenerate `updater/Gemfile.lock` via `bundle lock --update --bundler`
 # Generally simplest to match the bundler version to the one that comes by default with whatever Ruby version we install.
 # This way other projects that import this library don't have to futz around with installing new / unexpected bundler versions.
-ARG BUNDLER_V2_VERSION=2.4.22
+ARG BUNDLER_V2_VERSION=2.5.0
```
simi, line 112: Any reason to not use 2.5.3?

landongrindheim (approved):

deivid-rodriguez:
I debugged the smoke tests failure.

The problem is that for smoke tests, Dependabot uses cached network requests to provide deterministic behavior, since we assert for the upgraded lockfile, and don't want a transitive dependency release to cause a different result. This cache is based off request path and headers.

However, I'm observing that now the headers are slightly different:

> In Bundler 2.4: /versions ({"If-None-Match"=>"\"df4beda2ed9cc8f4741f5447f4e5259a\"", "Range"=>"bytes=19118384-"})
> In Bundler 2.5: /versions ({"Range"=>"bytes=19118384-", "If-None-Match"=>"df4beda2ed9cc8f4741f5447f4e5259a"})

The dropped quoting is probably invalidating Dependabot's cache, causing a full redownload of the versions file, and the difference we don't want to show up to appear.

This seems harmless so I think what we need to do is:

* Let the diff change happen, so manually regenerate smoke tests cache and make a PR there with the new expected lockfile diff.
* Once that's merged, this PR should get back to green and can be merged.
* Then next time smoke tests cache is regenerated automatically, it will be generated using Bundler 2.5, and the expected diffs for Bundler smoke tests will stay stable.

@jakecoffman Does this make sense?

jakecoffman:
Yep makes sense. This is why the smoke tests are not a required check. You can also merge this first and regenerate the smoke tests, whichever order you like.

deivid-rodriguez:
Alright, I regenerated the smoke tests bundler cache, and created a PR to the smoke tests repo at https://github.com/dependabot/smoke-tests/pull/164 👍. Once that lands, this should be good to ship!

Dockerfile.updater-core
deivid-rodriguez: Done now!

jakecoffman (approved):

PR #8620, opened 2023-12-15
support Composer artifact repositories

jakecoffman:
With Composer it's possible to have artifact repositories which are a local directory of zipped dependencies: https://getcomposer.org/doc/05-repositories.md#artifact

Since Dependabot wasn't fetching these archives, or even the empty directories, Composer was throwing errors saying the file wasn't found:

```
RecursiveDirectoryIterator::__construct(path/to/zipped/packages): failed to open dir: No such file or directory
```

This PR tries downloading the zipped files if they exist. If not a fake `.gitkeep` is added so it will create the directory to appease Composer.

### DIFF
--- a/composer/lib/dependabot/composer/file_fetcher.rb
+++ b/composer/lib/dependabot/composer/file_fetcher.rb
@@ -72,7 +72,12 @@ module Dependabot
                           .map { |file| File.join(url, file.name) }
 
               zip_files.map do |zip_file|
-                fetch_file_with_root_fallback(zip_file)
+                DependencyFile.new(
+                  name: zip_file,
+                  content: _fetch_file_content(zip_file),
+                  directory: directory,
+                  type: "file"
+                )
               end
             end
           end.flatten
@@ -82,7 +87,7 @@ module Dependabot
           DependencyFile.new(
             name: File.join(url, ".gitkeep"),
             content: "",
-            directory: Pathname.new(File.join(directory, url)).cleanpath.to_path,
+            directory: directory,
             type: "file"
           )
         end
--- a/composer/lib/dependabot/composer/file_fetcher.rb
+++ b/composer/lib/dependabot/composer/file_fetcher.rb
@@ -37,6 +37,7 @@ module Dependabot
         fetched_files << composer_json
         fetched_files << composer_lock if composer_lock
         fetched_files << auth_json if auth_json
+        fetched_files += artifact_dependencies
         fetched_files += path_dependencies
         fetched_files
       end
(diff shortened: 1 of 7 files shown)


jakecoffman:
It would probably be better if we cloned the whole repo and did the `in_a_temporary_repo_directory` but that's a pretty big change.

jakecoffman:
I tried this out with the CLI on a repo with artifacts, and a repo that specified an artifact but the directory was empty. Both worked.

composer/lib/dependabot/composer/file_fetcher.rb
```diff
@@ -90,17 +126,24 @@ def path_dependencies
           end
       end
 
+      def artifact_sources
```
abdulapopoola, line 129: seems only the 'type' param differs between both methods? Deferring to your decision though

composer/lib/dependabot/composer/file_fetcher.rb
```diff
@@ -90,17 +126,24 @@ def path_dependencies
...
           begin
             repos = parsed_composer_json.fetch("repositories", [])
             if repos.is_a?(Hash) || repos.is_a?(Array)
               repos = repos.values if repos.is_a?(Hash)
               repos = repos.select { |r| r.is_a?(Hash) }
 
               repos
-                .select { |details| details["type"] == "path" }
-                .map { |details| details["url"] }
+                .select { |details| details["type"] == "path" || details["type"] == "artifact" }
```
abdulapopoola, line 146: seems like the `artifact_sources` method?

composer/lib/dependabot/composer/update_checker/version_resolver.rb
```diff
@@ -471,6 +479,11 @@ def path_dependency_files
             dependency_files.select { |f| f.name.end_with?("/composer.json") }
         end
 
+        def zipped_path_dependency_files
+          @zipped_path_dependency_files ||=
+            dependency_files.select { |f| f.name.end_with?(".zip", ".gitkeep") }
```
abdulapopoola, line 484: Should we make these constants? Ditto for .gitkeep and others

abdulapopoola (approved):

composer/lib/dependabot/composer/file_fetcher.rb
jakecoffman: This seems more like a Ruby pattern of having `artifact_sources` and `path_sources` rather than `sources(type: "artifact")` which is why I went with this.

composer/lib/dependabot/composer/file_fetcher.rb
jakecoffman: It is, but this method is memoized so we don't have to parse the composer.json twice. 

There might be a better way to do this but this seemed ok.

composer/lib/dependabot/composer/update_checker/version_resolver.rb
jakecoffman: I don't think that would add much for safety, and would detract from readability depending on how that is done.

composer/lib/dependabot/composer/file_fetcher.rb
abdulapopoola: Thanks; TIL; I didn't know that!

PR #8624, opened 2023-12-16
Detect version constraint violations in NuGet

ryanbrandenburg:
Right now it's possible for Dependabot to send PRs which we know will fail because the version we attempt to update to violates the constraints of another package (which may not be automatically update-able). Rather than sending bad PRs, lets check for this condition right before we update and stop the process if it would cause an issue.


nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/NuGetUpdater.Core.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
JoeRobich, line 1: Looks like a BOM change perhaps. Should we specify a particular encoding in our .editorconfig?

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
```diff
@@ -211,27 +213,52 @@ public static IEnumerable<Dependency> GetTopLevelPackageDependenyInfos(Immutable
...
-            var projectDirectory = Path.GetDirectoryName(projectPath);
-            projectDirectory ??= repoRoot;
-            var topLevelFiles = Directory.GetFiles(repoRoot);
-            var nugetConfigPath = PathHelper.GetFileInDirectoryOrParent(projectDirectory, repoRoot, "NuGet.Config", caseSensitive: false);
-            if (nugetConfigPath is not null)
-            {
-                File.Copy(nugetConfigPath, Path.Combine(tempDirectory.FullName, "NuGet.Config"));
-            }
+            var tempProjectPath = await CreateTempProjectAsync(tempDirectory, repoRoot, projectPath, targetFramework, packages);
+            var (exitCode, stdOut, stdErr) = await ProcessEx.RunAsync("dotnet", $"build \"{tempProjectPath}\"");
```
JoeRobich, line 222: Would calling `restore` instead of `build` work here?

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
```diff
@@ -211,27 +213,52 @@ public static IEnumerable<Dependency> GetTopLevelPackageDependenyInfos(Immutable
...
+        finally
+        {
+            tempDirectory.Delete(recursive: true);
+        }
+    }
+
+    private static ProjectRootElement CreateProjectRootElement(ProjectBuildFile buildFile)
+    {
+        var xmlString = buildFile.Contents.ToFullString();
+        using var xmlStream = new MemoryStream(Encoding.UTF8.GetBytes(xmlString));
```
JoeRobich, line 237: Wonder if it would be more efficient to create a StringReader here instead of a MemoryStream.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
```diff
@@ -211,27 +213,52 @@ public static IEnumerable<Dependency> GetTopLevelPackageDependenyInfos(Immutable
...
 
-            var packageReferences = string.Join(
-                Environment.NewLine,
-                packages
-                    .Where(p => !string.IsNullOrWhiteSpace(p.Version)) // empty `Version` attributes will cause the temporary project to not build
-                    .Select(static p => $"<PackageReference Include=\"{p.Name}\" Version=\"[{p.Version}]\" />"));
+    private static async Task<string> CreateTempProjectAsync(DirectoryInfo tempDir, string repoRoot, string projectPath, string targetFramework, Dependency[] packages)
+    {
+        var projectDirectory = Path.GetDirectoryName(projectPath);
+        projectDirectory ??= repoRoot;
```
JoeRobich, line 247: I know this is moved code, but seems like this could just be inlined above.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
```diff
@@ -211,27 +213,52 @@ public static IEnumerable<Dependency> GetTopLevelPackageDependenyInfos(Immutable
...
+        var topLevelFiles = Directory.GetFiles(repoRoot);
```
JoeRobich, line 248: This appears to be unused.

JoeRobich (approved):
Ideally, I think the UpdateChecker would be identifying this in advance and either not recommending the update or attempting to update the dependency with the constraint also.

Changes look good to stop us from generating failing PRs.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
ryanbrandenburg: I tried to force the editorconfig to show me a warning about this but was not successful :/

ryanbrandenburg:

bdragon (approved):

bdragon:
This change will be included in Thursday's release.

PR #8626, opened 2023-12-16
Make `toml` an explicit requirement

jeffwidman:
I couldn't figure out why the tests were failing for:
* https://github.com/dependabot/dependabot-core/pull/7741 (which was superseded by https://github.com/dependabot/dependabot-core/pull/8627)

until I realized that `pipfile` imports `toml`:
https://github.com/pypa/pipfile/blob/4706d2cbd35e0b47a05a6421fa17f93827bc454f/setup.py#L44

which then gets used over in the unrelated file `parser.py`: https://github.com/dependabot/dependabot-core/blob/89ebc55dac8630574301a10917425f80a56e4763/python/helpers/lib/parser.py#L24

So let's make the import of `toml` explicit so that we aren't relying on the side effects of importing `pipfile`. The `toml` requirement from `pipfile` isn't pinned, so I simply pinned to the latest release.

Python `3.11` added a native `tomllib` library, so once we drop support for `Python 3.10` we can drop this 3p lib entirely.

### DIFF
--- a/python/helpers/lib/parser.py
+++ b/python/helpers/lib/parser.py
@@ -13,6 +13,8 @@ from pip._internal.req.constructors import (
 )
 
 from packaging.requirements import InvalidRequirement, Requirement
+# TODO: Replace 3p package `toml` with 3.11's new stdlib `tomllib` once we drop
+# support for Python 3.10.
 import toml
 
 # Inspired by pips internal check:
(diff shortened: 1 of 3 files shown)


jeffwidman:
Can I get a review on this? It's good to go.

jeffwidman:
@deivid-rodriguez @pavera can one of you take a look at this?

To my eyes it's extremely straightforward, so I suspect it just got missed in the Christmas time busyness.

jeffwidman:
@abdulapopoola this is ready for review...

abdulapopoola (approved):

jeffwidman:
Thanks!

PR #8649, opened 2023-12-19
pin Sorbet's version so it doesn't randomly break

jakecoffman:
The CI on `main` is randomly breaking because we don't have `sorbet` pinned.

This will make it so we update our types in a more controlled manor when Dependabot bumps the dependency.


jurre (approved):
Idea: some sort of automation to keep the version up to date on a schedule or something 🤔

PR #8659, opened 2023-12-21
improve how the CI workflow skips

jakecoffman:
This pre-filters the paths so that we don't have to start the matrix job at all when skipping. 

However I just realized this won't work because all the CI runs are required. I don't think there's a way to avoid having to run a job just to skip if it's required.


jakecoffman:
Yep they're all just in `Expected — Waiting for status to be reported` so we can't do this 😢

PR #8661, opened 2023-12-22
Handle missing files

ryanbrandenburg:
- If a file is missing (because the path contains an MSBuild property, or actually doesn't exist) don't fail the whole update.
- More correctly handle property replacement. The previous string-replace didn't work in scenarios like `[$(Property)]` and the property "realizer" was kind of confusing. I condensed all that into some helper functions that I think are more correct, and at least we can fix them all in one place in the future.


nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core.Test/Update/UpdateWorkerTestBase.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace NuGetUpdater.Core.Test.Update;
 
-public class UpdateWorkerTestBase
+public abstract class UpdateWorkerTestBase
```
ryanbrandenburg, line 11: Not really related to the rest of the PR but I noticed this and wanted to make sure we were explicit that you should only be extending this class.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
```diff
@@ -211,6 +200,50 @@ public static IEnumerable<Dependency> GetTopLevelPackageDependenyInfos(Immutable
...
+                }
+            }
+        } while (modified is true);
+
+        return msbuildString;
+    }
+
+    public static bool HasPropertyReplace(string versionContent)
+    {
+        return versionContent.Contains("$(") && versionContent.Contains(')');
```
JoeRobich, line 236: As a sanity check you could get the index of both and ensure that they are ordered properly.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
```diff
@@ -211,6 +200,50 @@ public static IEnumerable<Dependency> GetTopLevelPackageDependenyInfos(Immutable
...
+    public static string GetRootValue(string msbuildString, Dictionary<string, string> propertyInfo)
+    {
+        bool modified;
+        do
+        {
+            modified = false;
+            if (HasPropertyReplace(msbuildString))
+            {
+                modified = false;
+                foreach (var property in propertyInfo)
```
JoeRobich, line 218: Should we just get the property name and perform the replacement instead of spinning through all the properties?

ryanbrandenburg: We must spin through all the properties unfortunately because there could be nested property definition (`NewtonsoftJsonVersion`'s value is defined as the value of `NewtonsoftJsonV6Version` or some such).

JoeRobich: But we need to loop while property name is not null already, right?

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
```diff
@@ -211,6 +201,62 @@ public static IEnumerable<Dependency> GetTopLevelPackageDependenyInfos(Immutable
...
+    /// <summary>
+    /// Given an MSBuild string and a set of properties, returns our best guess at the final value MSBuild will evaluate to.
+    /// </summary>
+    /// <param name="msbuildString"></param>
+    /// <param name="propertyInfo"></param>
+    /// <returns></returns>
+    public static string GetRootValue(string msbuildString, Dictionary<string, string> propertyInfo)
+    {
+        bool modified;
+        var seenProperties = new HashSet<string>(StringComparer.Ordinal);
```
JoeRobich, line 213:
```suggestion
        var seenProperties = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
```diff
@@ -343,6 +389,10 @@ internal static async Task<ImmutableArray<ProjectBuildFile>> LoadBuildFiles(stri
             });
             buildFileList.AddRange(project.Imports.Select(i => i.ImportedProject.FullPath.NormalizePathToUnix()));
         }
+        catch(InvalidProjectFileException)
```
JoeRobich, line 392: nit:
```suggestion
        catch (InvalidProjectFileException)
```

JoeRobich: ignore if the current formatting satisfies the .editorconfig

JoeRobich (approved):
Looks good. Just some suggestions.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
```diff
@@ -71,15 +73,9 @@ public static string[] GetTargetFrameworkMonikers(ImmutableArray<ProjectBuildFil
...
+            if (HasPropertyReplace(tfms))
             {
-                var propertyName = tfms.Substring(2, tfms.Length - 3);
-                while (propertyName is not null)
-                {
-                    propertyName = propertyInfo.TryGetValue(propertyName, out tfms) && tfms.StartsWith("$(") && tfms.EndsWith(")")
-                        ? tfms.Substring(2, tfms.Length - 3)
-                        : null;
-                }
+                tfms = GetRootValue(tfms, propertyInfo);
```
JoeRobich, line 78: Could always call `GetRoolValue` here. Then all use of `HasPropertyReplace` and `GetPropertyName` will be done in pairs making a good case for them to become a single `TryGetPropertyName` method.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
ryanbrandenburg: Cleanup on save got turned off somehow, turned it back on, thanks!

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
ryanbrandenburg: Love that, good catch.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
```diff
@@ -211,6 +198,58 @@ public static IEnumerable<Dependency> GetTopLevelPackageDependenyInfos(Immutable
...
+    /// <param name="propertyInfo"></param>
+    /// <returns></returns>
+    public static string GetRootedValue(string msbuildString, Dictionary<string, string> propertyInfo)
+    {
+        bool modified;
+        var seenProperties = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+        do
+        {
+            modified = false;
+            if (TryGetPropertyName(msbuildString, out var propertyName))
```
JoeRobich, line 214: This can now become a while loop and we can remove the `modified` variable.
```suggestion
        while (TryGetPropertyName(msbuildString, out var propertyName))
        {
```

brettfo (approved):

jakecoffman (approved):

PR #8668, opened 2023-12-26
Discontinue `dependabot-omnibus` gem

deivid-rodriguez:
The `dependabot-omnibus` was a shortcut for standalone Dependabot usage, back from the times when we had a monolithic image.

These days I think there's no reason to keep shipping a gem that includes updaters for all ecosystems bundled with core.


deivid-rodriguez:
Not sure if relevant anymore, I'll just close.

PR #8678, opened 2023-12-29
Start removing updater coupling on specific ecosystems

deivid-rodriguez:
This PR:

* Removes some duplicated ecosystem requires from updater.
* Moves requires of all ecosystems from updater spec_helper to requires of specific stuff needed for each spec.
* Migrates updater specs to not depend on the docker ecosystem.

More work should follow to remove updater specs coupling on:

* [x] npm_and_yarn
* [x] go_modules
* [ ] bundler

After that, work to remove runtime coupling on all ecosystems could proceed.


jakecoffman:
@deivid-rodriguez Do you need help finishing this up? We want to make some dummy Updaters!

deivid-rodriguez:
I think this is mergeable as is, but needs more work to completely rely on dummy updaters in updater. Help appreciated, sure!

jakecoffman (approved):

deivid-rodriguez:
Oh, thanks for fixing things up, I actually thought this was already passing CI.

jakecoffman:
@deivid-rodriguez No problem! This is a great approach, thanks for putting it together.

PR #8679, opened 2023-12-29
NuGet efficiency

ryanbrandenburg:
I saw some problems related to timeouts when running Dependabot for NuGet packages. It seems that we were using the query http API in places where we already knew exactly which package it was we wanted. The query API can be quite slow on some nuget registry implementations which can exceed the 20 second timeout for that call. Rather than extend the timeout and tax NuGet registries I'm trying to use the Registration API which returns faster because it already knows which package it's working against. This has the downside that it returns a lot more JSON that we need to parse, but given the slowness of the query API I've found it to be a significant savings.

While working through this I found some other performance problems which might affect particularly large or nested repositories, such as the index call for a NuGet repo being made repeatedly even though it doesn't vary, and instances where caching results can keep us from walking transitive dependencies multiple times.


nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
```diff
@@ -99,6 +114,17 @@ def base_url_from_v3_metadata(metadata)
             &.fetch("@id")
         end
 
+        def registration_url_from_v3_metadata(metadata)
+          allowed_registration_types = %w(
+            RegistrationsBaseUrl/3.0.0-beta
+            RegistrationsBaseUrl
+          )
```
brettfo, line 121: According to [this](https://learn.microsoft.com/en-us/nuget/api/registration-base-url-resource#versioning) I'd say set the search order to:

```
RegistrationsBaseUrl/3.6.0
RegistrationsBaseUrl/3.4.0
RegistrationsBaseUrl/3.0.0-rc
RegistrationsBaseUrl/3.0.0-beta
RegistrationsBaseUrl
```

brettfo (changes requested):

ryanbrandenburg:
This needs a test of the paging system once I've got the existing tests working.

nuget/spec/dependabot/nuget/file_parser/project_file_parser_spec.rb
```diff
@@ -870,8 +918,9 @@ def dependencies_from(dep_info)
             end
 
             it "has the right details" do
-              query_stub = stub_search_results_with_versions_v3("microsoft.extensions.dependencymodel_cached",
-                                                                ["1.1.1", "1.1.0"])
+              versions_stub = stub_versions_v3("microsoft.extensions.dependencymodel_cached", ["1.1.1", "1.1.0"])
+              stub_registry_v3("microsoft.extensions.dependencymodel_cached", ["1.1.1", "1.1.0"])
```
JoeRobich, line 922: Does it need to be stubbed twice?

brettfo (approved):

ryanbrandenburg:
> This needs a test of the paging system once I've got the existing tests working.

Turns out the smoke-tests already handle this case for us.

nuget/spec/dependabot/nuget/file_parser/project_file_parser_spec.rb
ryanbrandenburg: This shouldn't stub anything twice, it stubs the call to the versions API and then the call to the registry API. the Versions API is used to quickly and cheaply establish if the package exists at all, and the registry API is used to generate the list of actual versions we need (the versions API includes unlisted versions which we need to exclude).

nuget/lib/dependabot/nuget/file_fetcher.rb
```diff
@@ -32,6 +33,7 @@ def self.required_files_message
 
       sig { override.returns(T::Array[DependencyFile]) }
       def fetch_files
+        @files_fetched = {}
```
JoeRobich, line 36: Would the CacheManager be a good fit for this?

JoeRobich (approved):
LGTM

jakecoffman (approved):

jakecoffman:
I deployed this but getting a new stack trace:

```
KeyError: key not found: :versions_url
  from nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb:310:in `fetch'
  from nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb:310:in `dependency_url_has_matching_result_v3?'
  from nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb:301:in `dependency_url_has_matching_result?'
  from nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb:294:in `block in dependency_has_search_results?'
  from nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb:293:in `any?'
  from nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb:293:in `dependency_has_search_results?'
  from nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb:278:in `build_dependency'
  from nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb:98:in `block in parse_dependencies'
  from nokogiri/xml/node_set.rb:235:in `block in each'
  from nokogiri/xml/node_set.rb:234:in `upto'
  from nokogiri/xml/node_set.rb:234:in `each'
  from nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb:91:in `parse_dependencies'
  from nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb:58:in `dependency_set'
  from nuget/lib/dependabot/nuget/file_parser.rb:39:in `block in project_file_dependencies'
  from nuget/lib/dependabot/nuget/file_parser.rb:37:in `each'
  from nuget/lib/dependabot/nuget/file_parser.rb:37:in `project_file_dependencies'
  from nuget/lib/dependabot/nuget/file_parser.rb:25:in `parse'
  from dependabot-updater/lib/dependabot/dependency_snapshot.rb:101:in `parse_files!'
  from dependabot-updater/lib/dependabot/dependency_snapshot.rb:92:in `initialize'
  from dependabot-updater/lib/dependabot/dependency_snapshot.rb:24:in `new'
  from dependabot-updater/lib/dependabot/dependency_snapshot.rb:24:in `create_from_job_definition'
  from dependabot-updater/lib/dependabot/update_files_command.rb:21:in `perform_job'
  from dependabot-updater/lib/dependabot/base_command.rb:53:in `run'
  from bin/update_files.rb:24:in `<main>'
```

ryanbrandenburg:
Talked to @jakecoffman offline. I think this happens when a NuGet Source varies from the specification by omitting the [required `PackageBaseAddress` resource](https://learn.microsoft.com/en-us/nuget/api/overview#resources-and-schema). I believe the source in question here is Artifactory, as supported by [this issue on their repo](https://jfrog.atlassian.net/browse/RTFACT-20469) which seems to have been filed based on [this issue from Paket](https://github.com/fsprojects/Paket/issues/3638).

Because this scenario DID used to work and a work-around (using the registration API) is relatively easy and cheap I'm going to try to make this work. Obviously if a NuGet Source is not following the specifications we're in dangerous territory, but as long as the only spec deviation is not listing PackageBaseAddress I think we can get there.

brettfo:
> This may very well be the fix for [tinglesoftware/dependabot-azure-devops#921](https://github.com/tinglesoftware/dependabot-azure-devops/issues/921)

@mburumaxwell This PR probably won't fix the issue you linked, **but** #8703 in combination with a separate internal change that went in yesterday should fix the issue.  The internal issue was that we originally only injected authorization for the `SearchQueryService` NuGet endpoint, but some feeds don't return that, they return `SearchQueryService/3.0.0-beta`, `SearchQueryService/3.0.0-rc`, etc. so we just expanded the endpoints where we inject auth tokens.

nuget/lib/dependabot/nuget/nuget_helpers.rb
```diff
@@ -0,0 +1,108 @@
+# typed: true
+# frozen_string_literal: true
+
+require "dependabot/nuget/cache_manager"
+require "dependabot/nuget/update_checker/repository_finder"
+
+module Dependabot
+  module Nuget
+    class NugetHelpers
```
JoeRobich, line 9: Like how you gathered a lot of the v3 api usage together here. Would love to see us do the same for v2.

JoeRobich (approved):

nuget/lib/dependabot/nuget/nuget_helpers.rb
ryanbrandenburg: I almost pulled V2 in in this PR, but it's already quite large and we're wanting to get it out to people soon. If anyone were to move the v2 APIs onto here in the future (and even better combine them so the repository_details decides which one to use) I'd be super here for it.

bdragon (approved):
This looks good to me!

The only thing I would recommend double-checking is the use of `Hash#[]` vs. `Hash#fetch()` without a default, probably mainly in the response-parsing code in `Dependabot::Nuget::NugetClient`. `fetch()` without a default will raise if the field is omitted, so we just want to make we always expect those fields to exist.

ryanbrandenburg:
@bdragon I double-checked the NugetClient usage of `.fetch` and I believe that all those usages are required fields (in their given contexts). I don't have much of a Ruby background, is it considered better form to use `hash[key]` or `.fetch(key, nil)`? I had assumed `hash[key]` was cleaner when you weren't sure if you have the KVP (and I also am under the impression that the two are interchangeable).

bdragon:
@ryanbrandenburg 

In general they will do the same thing, but there is a subtle difference. When you create a hash, you have the option to provide a block that will be used to specify the return value when a caller tries to access a key that doesn't exist with `hash[key]` (this is the `Hash#default_proc`). For example, you could do this:

```ruby
irb(main):001:0> a = {}
=> {}
irb(main):002:0> a[:missing]
=> nil
irb(main):003:0> b = Hash.new { |h, k| "unknown key '#{k}'" }
=> {}
irb(main):004:0> b[:missing]
=> "unknown key 'missing'"
```

The default behavior for `#default_proc` is to just return nil, but as you can see the behavior of `hash[key]` kind of depends on how it was created.

When a key might be missing `hash[key]` feels cleaner to me, too, but I guess technically it's not guaranteed to return nil if the key is missing if the hash has a custom `default_proc` for some reason. In contrast, `hash.fetch(key, nil)` is always guaranteed to return nil if the key is missing.

The main thing I look out for is `hash.fetch(key)` without a default value, since that will raise a `KeyError` if the key is missing and this has bitten us before.

ryanbrandenburg:
Ah, I'm glad I asked! So it sounds like if we were taking Hash's from outside callers (as a library for usage by other projects) we might use `hash.fetch(key, nil)` to guarantee behavior but for an internally contained project like this `hash[key]` should be fine when nil-return is desired and `hash.fetch(key)` should be fine when a result is "known" to exist. That being the case I think the PR adheres to those guidelines so we're good there.

nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
```diff
@@ -62,7 +64,9 @@ def build_url_for_details(repo_details)
           body = remove_wrapping_zero_width_chars(response.body)
           base_url = base_url_from_v3_metadata(JSON.parse(body))
           resolved_base_url = base_url || repo_details.fetch(:url).gsub("/index.json", "-flatcontainer")
-          search_url = search_url_from_v3_metadata(JSON.parse(body))
+          parsed_json = JSON.parse(body)
```
JoeRobich, line 67: We should move this up so we can use it when getting the base_url from the response body.

nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
```diff
@@ -54,6 +55,7 @@ def find_dependency_urls
             end.compact.uniq
         end
 
+        # rubocop:disable Metrics/AbcSize
         def build_url_for_details(repo_details)
```
JoeRobich, line 59: Not for this PR, but I think this method would is a good candidate for some cleanup. Doing a bit more than building a url. Would we want to encapsulate parsing the repo metadata response somewhere else?

JoeRobich (approved):

bdragon:
Is this ready to deploy (once up-to-date with the base branch)?

ryanbrandenburg:
> Is this ready to deploy (once up-to-date with the base branch)?

I believe so.

ryanbrandenburg:
Looks like there was some unreliability in the smoke tests? It passed on re-run and doesn't seem related to my changes.

bdragon:
I've deployed the updater with this change so I'm going to merge this PR. I'll cut a release of the dependabot gems on Thursday and it will include this change.

Issue #8633, opened 2023-12-16
NuGet update failes when project path contains spaces.


TomW-Skyline:
The command arguments need to be surrounded by quotes, because otherwise there are problems when they contain spaces.

https://github.com/dependabot/dependabot-core/blob/89ebc55dac8630574301a10917425f80a56e4763/nuget/lib/dependabot/nuget/native_helpers.rb#L57

abdulapopoola:
Sorry to hear that; will reopen the issue and tagging @brettfo for awareness

brettfo:
@TomW-Skyline Do you have a recent log file you can share?  There are likely more places we need to surround arguments.

brettfo:
@TomW-Skyline Added PR #9092 to allow proper execution of the tool with spaces in the command line.

brettfo:
@TomW-Skyline I created a repo with a space in the path and did a full dependabot update run to verify the path fix.  I've also just submitted #9104 which will fail a unit test on a malformed path so we can detect this faster.

PR #8695, opened 2024-01-04
Surround command line arguments with quotes

TomW-Skyline:
Fixes https://github.com/dependabot/dependabot-core/issues/8633


nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Cli/Program.cs
```diff
@@ -20,6 +21,9 @@ internal static async Task<int> Main(string[] args)
         };
         command.TreatUnmatchedTokensAsErrors = true;
 
+        // trim quotes
+        args = args.Select(x => x.Trim('"')).ToArray();
```
brettfo, line 25: Are there double quotes coming through here?  I thought the dotnet argument parsing code and/or OS would trim them off before they get here.

brettfo (approved):

jakecoffman:
Is there a way we can test this? Should we modify one of the [NuGet smoke tests](https://github.com/dependabot/smoke-tests/tree/main/tests) to have a space in a name?

brbayes-msft (approved):

PR #8698, opened 2024-01-04
Normalize folder names

deivid-rodriguez:
I find it confusing that we use different naming in source code ("updater/") and in the updater image file structure ("Dependabot-updater/").

This PR tries to normalize that.


jakecoffman:
We'll have to change these in all the runners: https://github.com/dependabot/cli/blob/f50317a4a56527afc8f575f19f39e78e0c9c2303/internal/infra/updater.go#L33-L35

I think they have to be absolute paths.

I wonder if dropping a symbolic link in the Dockerfile from `dependabot-updater` to `updater` would ease the transition?

deivid-rodriguez:
Yeah, correct!

I added some code for backwards compatibility but honestly a symlink in the Dockerfile sounds like a better and safer approach.

jakecoffman (dismissed):
This is one of those PRs I'd make sure to use it with staging before going to prod 😄

deivid-rodriguez:
Haha, yes, I was planning to do that. It's an all works or nothing works PR 😅.

deivid-rodriguez:
Not sure if this PR makes sense anymore. It seems too old, so I'm closing it, since it seems it's in nobody's radar.

PR #8702, opened 2024-01-04
fix unsupported Go modules preventing updates

### LINKED ISSUE #8700: Unsupported Go modules cause an exception

- [x] I have searched the existing issues

### Package ecosystem

Go

### Package manager version

1.21.5

### Language version

_No response_

_No response_

### dependabot.yml content

_No response_

### Updated dependency

_No response_

I expect to see an error in the logs, but the exception should be caught.

The exception is because [`rsc.io/script v0.0.1`][1] only supports Go v1.22, but the [Dependabot updater images uses Go v1.21.5][2].

The fix is to add the `-e` flag to the [`go list` command that is used][3] which will print a JSON object with errors to stdout, instead of an error message to stderr, which causes an uncaught exception.

An example of the error JSON is

```json
{
        "Path": "rsc.io/script",
        "Error": {
                "Err": "loading module retractions for rsc.io/script@v0.0.1: module rsc.io/script@v0.0.1 requires go \u003e= 1.22 (running go 1.21.5)"
        }
}
```

See [the documentation for more information][4]

[1]: https://github.com/rsc/script/blob/v0.0.1/go.mod
[2]: https://github.com/dependabot/dependabot-core/blob/a4b13fb2735ef4dd119d394f2d14f6847d488d43/go_modules/Dockerfile#L1
[3]: https://github.com/dependabot/dependabot-core/blob/a4b13fb2735ef4dd119d394f2d14f6847d488d43/go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb#L97
[4]: https://pkg.go.dev/cmd/go/internal/list

_No response_

https://github.com/dependabot/cli/network/updates/767958040

<details>
<summary>Logs</summary>

```
Error processing rsc.io/script (Dependabot::SharedHelpers::HelperSubprocessFailed)
go: loading module retractions for rsc.io/script@v0.0.1: module rsc.io/script@v0.0.1 requires go >= 1.22 (running go 1.21.5; GOTOOLCHAIN=local)
/home/dependabot/common/lib/dependabot/shared_helpers.rb:427:in `run_shell_command'
/home/dependabot/dependabot-updater/vendor/ruby/3.1.0/gems/sorbet-runtime-0.5.11142/lib/types/private/methods/call_validation.rb:153:in `bind_call'
/home/dependabot/dependabot-updater/vendor/ruby/3.1.0/gems/sorbet-runtime-0.5.11142/lib/types/private/methods/call_validation.rb:153:in `validate_call_skip_block_type'
/home/dependabot/dependabot-updater/vendor/ruby/3.1.0/gems/sorbet-runtime-0.5.11142/lib/types/private/methods/call_validation.rb:95:in `block in create_validator_slow_skip_block_type'
/home/dependabot/go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb:96:in `block (2 levels) in fetch_available_versions'
/home/dependabot/common/lib/dependabot/shared_helpers.rb:264:in `with_git_configured'
/home/dependabot/dependabot-updater/vendor/ruby/3.1.0/gems/sorbet-runtime-0.5.11142/lib/types/private/methods/call_validation.rb:256:in `bind_call'
/home/dependabot/dependabot-updater/vendor/ruby/3.1.0/gems/sorbet-runtime-0.5.11142/lib/types/private/methods/call_validation.rb:256:in `validate_call'
/home/dependabot/dependabot-updater/vendor/ruby/3.1.0/gems/sorbet-runtime-0.5.11142/lib/types/private/methods/call_validation.rb:177:in `block in create_validator_slow'
/home/dependabot/go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb:82:in `block in fetch_available_versions'
/home/dependabot/common/lib/dependabot/shared_helpers.rb:80:in `block in in_a_temporary_directory'
/home/dependabot/common/lib/dependabot/shared_helpers.rb:80:in `chdir'
/home/dependabot/common/lib/dependabot/shared_helpers.rb:80:in `in_a_temporary_directory'
/home/dependabot/dependabot-updater/vendor/ruby/3.1.0/gems/sorbet-runtime-0.5.11142/lib/types/private/methods/call_validation.rb:256:in `bind_call'
/home/dependabot/dependabot-updater/vendor/ruby/3.1.0/gems/sorbet-runtime-0.5.11142/lib/types/private/methods/call_validation.rb:256:in `validate_call'
/home/dependabot/dependabot-updater/vendor/ruby/3.1.0/gems/sorbet-runtime-0.5.11142/lib/types/private/methods/call_validation.rb:177:in `block in create_validator_slow'
/home/dependabot/go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb:81:in `fetch_available_versions'
/home/dependabot/go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb:77:in `available_versions'
/home/dependabot/go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb:56:in `fetch_latest_version'
/home/dependabot/go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb:44:in `latest_version'
/home/dependabot/go_modules/lib/dependabot/go_modules/update_checker.rb:16:in `latest_resolvable_version'
/home/dependabot/go_modules/lib/dependabot/go_modules/update_checker.rb:23:in `latest_version'
/home/dependabot/dependabot-updater/lib/dependabot/updater/operations/update_all_versions.rb:177:in `all_versions_ignored?'
/home/dependabot/dependabot-updater/lib/dependabot/updater/operations/update_all_versions.rb:80:in `check_and_create_pull_request'
/home/dependabot/dependabot-updater/lib/dependabot/updater/operations/update_all_versions.rb:60:in `check_and_create_pr_with_error_handling'
/home/dependabot/dependabot-updater/lib/dependabot/updater/operations/update_all_versions.rb:35:in `block in perform'
/home/dependabot/dependabot-updater/lib/dependabot/updater/operations/update_all_versions.rb:35:in `each'
/home/dependabot/dependabot-updater/lib/dependabot/updater/operations/update_all_versions.rb:35:in `perform'
/home/dependabot/dependabot-updater/lib/dependabot/updater/operations/group_update_all_versions.rb:128:in `run_ungrouped_dependency_updates'
/home/dependabot/dependabot-updater/lib/dependabot/updater/operations/group_update_all_versions.rb:63:in `perform'
/home/dependabot/dependabot-updater/lib/dependabot/updater.rb:64:in `run'
/home/dependabot/dependabot-updater/lib/dependabot/update_files_command.rb:43:in `perform_job'
/home/dependabot/dependabot-updater/lib/dependabot/base_command.rb:53:in `run'
bin/update_files.rb:24:in `<main>'
```

</details>

_No response_

jakecoffman:
fix #8700

As @JamieMagee mentions, without the -e flag unsupported versions of modules breaks updates for the entire run since `go list` errors out.

It might be worth throwing this into the Go smoke test to prevent regressions.

### DIFF
--- a/go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb
+++ b/go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb
@@ -23,7 +23,10 @@ module Dependabot
           /unrecognized import path/,
           /malformed module path/,
           # (Private) module could not be fetched
-          /module .*: git ls-remote .*: exit status 128/m
+          /module .*: git ls-remote .*: exit status 128/m,
+          # The module was retracted from the proxy
+          # OR the version of Go required is greater than what Dependabot supports
+          /go: loading module retractions for/
         ].freeze
         INVALID_VERSION_REGEX = /version "[^"]+" invalid/m
         PSEUDO_VERSION_REGEX = /\b\d{14}-[0-9a-f]{12}$/
@@ -94,8 +97,8 @@ module Dependabot
               env = { "GOPRIVATE" => @goprivate }
 
               versions_json = SharedHelpers.run_shell_command(
-                "go list -e -m -versions -json #{dependency.name}",
-                fingerprint: "go list -e -m -versions -json <dependency_name>",
+                "go list -m -versions -json #{dependency.name}",
+                fingerprint: "go list -m -versions -json <dependency_name>",
                 env: env
               )
               version_strings = JSON.parse(versions_json)["Versions"]
--- a/go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb
+++ b/go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb
@@ -94,8 +94,8 @@ module Dependabot
               env = { "GOPRIVATE" => @goprivate }
 
               versions_json = SharedHelpers.run_shell_command(
-                "go list -m -versions -json #{dependency.name}",
-                fingerprint: "go list -m -versions -json <dependency_name>",
+                "go list -e -m -versions -json #{dependency.name}",
+                fingerprint: "go list -e -m -versions -json <dependency_name>",
                 env: env
               )
               version_strings = JSON.parse(versions_json)["Versions"]
--- a/go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb
+++ b/go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb
@@ -23,7 +23,7 @@ module Dependabot
           /unrecognized import path/,
           /malformed module path/,
           # (Private) module could not be fetched
-          /module .*: git ls-remote .*: exit status 128/m,
+          /module .*: git ls-remote .*: exit status 128/m
         ].freeze
         # The module was retracted from the proxy
         # OR the version of Go required is greater than what Dependabot supports
--- a/go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb
+++ b/go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb
@@ -24,11 +24,11 @@ module Dependabot
           /malformed module path/,
           # (Private) module could not be fetched
           /module .*: git ls-remote .*: exit status 128/m,
-          # The module was retracted from the proxy
-          # OR the version of Go required is greater than what Dependabot supports
-          /go: loading module retractions for/
         ].freeze
-        INVALID_VERSION_REGEX = /version "[^"]+" invalid/m
+        # The module was retracted from the proxy
+        # OR the version of Go required is greater than what Dependabot supports
+        # OR other go.mod version errors
+        INVALID_VERSION_REGEX = /(go: loading module retractions for)|(version "[^"]+" invalid)/m
         PSEUDO_VERSION_REGEX = /\b\d{14}-[0-9a-f]{12}$/
 
         def initialize(dependency:, dependency_files:, credentials:,
--- a/go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb
+++ b/go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb
@@ -114,17 +114,7 @@ module Dependabot
           retry_count += 1
           retry if transitory_failure?(e) && retry_count < 2
 
-          handle_subprocess_error(e)
-        end
-
-        def handle_subprocess_error(error)
-          if RESOLVABILITY_ERROR_REGEXES.any? { |rgx| error.message =~ rgx }
-            ResolvabilityErrors.handle(error.message, goprivate: @goprivate)
-          elsif INVALID_VERSION_REGEX.match?(error.message)
-            raise Dependabot::DependencyFileNotResolvable, error.message
-          end
-
-          raise
+          ResolvabilityErrors.handle(e.message, goprivate: @goprivate)
         end
 
         def transitory_failure?(error)
--- a/go_modules/spec/dependabot/go_modules/update_checker/latest_version_finder_spec.rb
+++ b/go_modules/spec/dependabot/go_modules/update_checker/latest_version_finder_spec.rb
@@ -62,6 +62,10 @@ RSpec.describe Dependabot::GoModules::UpdateChecker::LatestVersionFinder do
     )
   end
 
+  before do
+    ENV["GOTOOLCHAIN"] = "local"
+  end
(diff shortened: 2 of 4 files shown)


go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb
```diff
@@ -94,8 +94,8 @@ def fetch_available_versions
               env = { "GOPRIVATE" => @goprivate }
 
               versions_json = SharedHelpers.run_shell_command(
-                "go list -m -versions -json #{dependency.name}",
-                fingerprint: "go list -m -versions -json <dependency_name>",
+                "go list -e -m -versions -json #{dependency.name}",
+                fingerprint: "go list -e -m -versions -json <dependency_name>",
                 env: env
               )
               version_strings = JSON.parse(versions_json)["Versions"]
```
JamieMagee, line 101: What do you think about adding an `error = JSON.parse(versions_json)["Error"]["Err"]` here and log that as a warning?

go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb
```diff
@@ -94,8 +94,8 @@ def fetch_available_versions
               env = { "GOPRIVATE" => @goprivate }
 
               versions_json = SharedHelpers.run_shell_command(
-                "go list -m -versions -json #{dependency.name}",
-                fingerprint: "go list -m -versions -json <dependency_name>",
+                "go list -e -m -versions -json #{dependency.name}",
```
mctofu, line 97: Will this prevent `SharedHelpers::HelperSubprocessFailed` from being raised for the conditions handled in https://github.com/dependabot/dependabot-core/pull/8702/files#diff-9536ab82fee1fffd8e6fe35594bcddae3979d6de56792638b57d551856fff01eR109-R115?

If so, we'd need to convert that handling to use the form of error output or handle this particular error via exception instead?

jakecoffman: You're right, I think the short term solution is just to add it as another resolvability error instead.

go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb
jakecoffman: I wonder if this would be a better longer term fix. If the error comes back in the JSON we know it was a resolvability error and don't need to maintain regex for error messages. If it is a HelperSubprocessFailed error it's probably something on our side we need to fix.

JamieMagee (approved):

go_modules/lib/dependabot/go_modules/update_checker/latest_version_finder.rb
jakecoffman: Actually now that I've looked at the error handling, I'm wondering why we're going to so much trouble? If we use `-e` then take any errors out of the JSON and raise them as DependencyFileNotResolvable that's much simpler. Then we don't have to keep updating this code every time a new error message pops up.

What do you think?

mctofu: Makes sense to me. Trying to identify errors based on the message isn't ideal.

mctofu (approved):

mctofu:
Could we add a fixture/test for this? I think we can reproduce it by adding a test module to https://github.com/dependabot-fixtures that requires a future go version?

jakecoffman:
I tried to use the `-e` flag but it turns out it still returns an error code so it doesn't really help much.

I ended up just removing the regex check so we can stop chasing error message changes with each new Go release. Also added a test for this scenario.

go_modules/spec/dependabot/go_modules/update_checker/latest_version_finder_spec.rb
```diff
@@ -62,6 +62,10 @@ module foobar
     )
   end
 
+  before do
+    ENV["GOTOOLCHAIN"] = "local"
```
jakecoffman, line 66: Prevent flakiness from other tests that might have set the GOTOOLCHAIN to something else.

jakecoffman:
Upon further investigation, I don't think this was preventing updates from happening. We were looking at the dependabot/cli repo and it was setup to only update direct dependencies which is why it wasn't updating the indirect ones.

This is still a good change and improved the error message:

Screenshot 2024-01-05 at 10 06 08 AM

JamieMagee:
Sorry for misreading the logs!

PR #8704, opened 2024-01-04
Check if RBIs are up-to-date

JamieMagee:
The previous RBI check, `bin/tapioca && git add . && git diff --cached --exit-code` would always succeed because `bin/tapioca` prints the help message, but doesn't generate the RBIs.

The updated script runs `tapioca gem --verify` instead.

Related to #8670


JamieMagee:
Turns out I have Bundler 2.4.22 installed locally, which resolves the dependency tree differently than Bundler 2.5.4, which is installed by default in CI.

deivid-rodriguez (approved):
🙈 Thanks for catching my mistake!

deivid-rodriguez:
I'm afraid this is not going to work. Since there's no `Gemfile.lock` file, any release of any dependency will cause the check to fail. Should we add a lockfile?

JamieMagee:
> I'm afraid this is not going to work. Since there's no `Gemfile.lock` file, any release of any dependency will cause the check to fail. Should we add a lockfile?

Yeah, I did always wonder why there was no `Gemfile.lock`. Do you know?

deivid-rodriguez:
No idea.

We should eventually merge this Gemfile with the one in updater (once aarch64 is supported by sorbet) and manage everything together, but for now we can add a root lockfile. Mind if I add that to this PR?

JamieMagee:
I don't think Sorbet Linux + aarch64 support is a huge blocker? It just means you won't be able to typecheck locally in a devcontainer. CI will still typecheck, and it will still work in mac + aarch64.

Go ahead with the lockfile. Which Bundler version are you going to use?

deivid-rodriguez:
I think the Bundler version should not change anything but I'll lock the latest. The Ruby version could make a difference since it's common that gems bring different requirements on the Ruby version so different rubies could lead to different resolutions. The Ruby version used will be 3.1.4 as per the root `.ruby-version` file.

Regarding Mac + aarch64, it's not that you cannot run type check, but that Bundler won't be able to resolve the updater Gemfile at all in macOS if it includes sorbet-static.

deivid-rodriguez:
I pushed a lockfile and also added Dependabot commit to bump all gems in it weekly. That weekly PR will need an amend to run `bin/tapioca gem`, but that seems fine?

deivid-rodriguez:
Good to go @JamieMagee?

JamieMagee:
Yep!

PR #8227, opened 2023-10-18
chore(python): target latest python versions 3.12.0, 3.11.6, 3.10.13, 3.9.18, 3.8.18

dshick-nydig:
Last time done at https://github.com/dependabot/dependabot-core/pull/7914 by @phillipuniverse

PR #8705, opened 2024-01-05
Update pyenv and pythons

jurre:
Python 3.12 now exists, so this updates pyenv to a version that knows about it and does the dance in the docker file to install that version

### DIFF
--- a/python/Dockerfile
+++ b/python/Dockerfile
@@ -1,12 +1,14 @@
 # This list must match the versions specified in
 # python/lib/dependabot/python/language_version_manager.rb: PRE_INSTALLED_PYTHON_VERSIONS
-ARG PY_3_11=3.11.5
+ARG PY_3_12=3.12.1
+ARG PY_3_11=3.11.7
 ARG PY_3_10=3.10.13
 ARG PY_3_9=3.9.18
 ARG PY_3_8=3.8.18
-ARG PYENV_VERSION=v2.3.25
+ARG PYENV_VERSION=v2.3.35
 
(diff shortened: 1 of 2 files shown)


python/Dockerfile
```diff
@@ -1,16 +1,16 @@
...
-ARG PY_3_8=3.8.18
-ARG PYENV_VERSION=v2.3.25
+ARG PYENV_VERSION=v2.3.35
 
 FROM ghcr.io/dependabot/dependabot-updater-core as python-core
+ARG PY_3_12
 ARG PY_3_11
 ARG PY_3_10
 ARG PY_3_9
-ARG PY_3_8
```
landongrindheim, line 13: 🤔 We published [a changelog post](https://github.blog/changelog/2023-08-02-deprecation-notice-dependabot-to-drop-support-for-python-3-6-and-3-7/) for dropping 3.6/3.7 support. What do you think about dropping support in another PR, perhaps with a similar approach (or publishing some kind of signal that states we'll continually drop support for outdated versions)?

jurre: We're not dropping support for it, it's just not installed by default so when running a job that needs this version, we'll first install it

jurre: Oh looks like we're not installing stuff on the fly anymore, I'll put it back

deivid-rodriguez:
There's also https://github.com/dependabot/dependabot-core/pull/8227, which I think it's the same thing but also started some work on test failures.

jurre:
> There's also #8227, which I think it's the same thing but also started some work on test failures.

Oh no, I'd totally missed that, I did try searching for it 🤦

PR #8712, opened 2024-01-05
Improve root gemfile handling

deivid-rodriguez:
* I think updating all dependencies in Gemfile is too much as explained in https://github.com/dependabot/dependabot-core/pull/8709#issuecomment-1879156849.
* Is anyone using `reek` and `solargraph`? They've been there for ages but I believe they're not being used by anyone?


jakecoffman (approved):

deivid-rodriguez:
I guess if we were to use some editor integration, it should be Sorbet-based! Since @jakecoffman is fine with it, I'll merge this, we can always revert if someone was actually using this!

PR #8722, opened 2024-01-05
properly locate `nuget.config` when it's further up-directory from a project

brettfo:
If dependabot is started from a deep directory, a `NuGet.config` file further up the tree (e.g., at the root) wasn't properly found.  This fixes the search behavior to look further up when the file name might not be `NuGet.config`, but `../../NuGet.config`.


JoeRobich (approved):
lgtm

jakecoffman (approved):

PR #8727, opened 2024-01-07
Strong type `Dependabot::Version`

JamieMagee:


common/lib/dependabot/version.rb
```diff
@@ -1,21 +1,36 @@
...
   class Version < Gem::Version
+    extend T::Sig
+    extend T::Helpers
+
+    abstract!
+
+    sig { overridable.params(version: T.any(String, Dependabot::Version)).void }
     def initialize(version)
-      @original_version = version
+      @original_version = T.let(version.to_s, String)
```
JamieMagee, line 15: The only use of this is `to_semver`, and the only use of that is `ignore_condition`, which expects a `String`:

https://github.com/dependabot/dependabot-core/blob/389cb8a49af6a4c499da6322a6c138f734ccc310/common/lib/dependabot/config/ignore_condition.rb#L60-L73

Gemfile
```diff
@@ -22,7 +22,7 @@ gemspec path: "swift"
 gemspec path: "terraform"
 
 # Sorbet
-gem "sorbet", "0.5.11156", group: :development
+gem "sorbet", "0.5.11178", group: :development
```
JamieMagee, line 25: This upgrade is necessary because `sorbet` version `0.5.11158` shipped with more accurate types for `Gem::Version`[^1]. Specifically, Dependabot uses `String`, `Gem::Version` and `Integer` interchangeably throughout the codebase.

[^1]: https://github.com/sorbet/sorbet/commit/8e8ac5509099153963e38c3c94b8a3ea27515df5

common/lib/dependabot/version.rb
```diff
@@ -1,21 +1,75 @@
...
+            Float,
+            Gem::Version,
+            NilClass
+          )
+        )
+        .returns(Dependabot::Version)
+    end
+    def self.new(version)
+      T.cast(super, Dependabot::Version)
     end
```
JamieMagee, line 49: This override is required because we do `version_class.new`, and without the override, `super` calls `Gem::Version.new` which returns `Gem::Version` which doesn't have `self.correct?`.

composer/lib/dependabot/composer/update_checker/version_resolver.rb
```diff
@@ -218,7 +219,7 @@ def updated_version_requirement_string
                           .select { |req_string| req_string.match?(VERSION_REGEX) }
                           .map { |req_string| req_string.match(VERSION_REGEX) }
                           .select { |version| requirement_valid?(">= #{version}") }
-                          .max_by { |version| Composer::Version.new(version) }
+                          .max_by { |version| Composer::Version.new(version.to_s) }
```
JamieMagee, line 222: Without `to_s` `version` is `MatchData`.

docker/lib/dependabot/docker/version.rb
```diff
@@ -41,7 +41,7 @@ def self.correct?(version)
 
         release_part, = parsed_version[:version].split("_", 2)
         release_part = Tag.new(release_part.chomp(".").chomp("-").chomp("_")).numeric_version || parsed_version
-        super(release_part)
+        super(release_part.to_s)
```
JamieMagee, line 44: Without `to_s` `version` is `MatchData`.

common/lib/dependabot/version.rb
bdragon: Weird but makes sense given that `Gem::Version` [overrides `::new`](https://github.com/rubygems/rubygems/blob/8ffa73df3a250623b13c5bf7ab48ed8b8e84e46f/lib/rubygems/version.rb#L203)

npm_and_yarn/lib/dependabot/npm_and_yarn/version.rb
```diff
@@ -42,29 +55,58 @@ def self.semver_for(version)
...
+          .returns(Dependabot::NpmAndYarn::Version)
+      end
+      def self.new(version)
+        T.cast(super, Dependabot::NpmAndYarn::Version)
       end
 
       sig { returns(Integer) }
       def major
-        @major ||= T.let(segments[0] || 0, T.nilable(Integer))
+        @major ||= T.let(segments[0].to_i, T.nilable(Integer))
```
bdragon, line 99: Here and in `#minor`, `#patch`, I'm wondering why the `T.nilable`? Seems like these values will always be an integer.

bdragon (commented):
Nice work @JamieMagee! Thanks for annotating the diff with comments 🙂

LGTM. I just had a question about some `T.nilable`s.

npm_and_yarn/lib/dependabot/npm_and_yarn/version.rb
JamieMagee: It's because of the use of `||=` syntax. It's shorthand for `@major || major = segments[0].to_i`, which implies that `@major` can be `nil`, so Sorbet forces us to make it `T.nilable(Integer)`.

It's accurate though, because `@major` is `nil` until we first call `major`. At which point we set `@major` and return it.

bdragon: Ok, thanks for the explanation. I think I'm just still getting the hang of sorbet. I was reading that as the type for the `segments[0].to_i` expression, which will always be an integer because `nil.to_i => 0` and `"xyz".to_i => 0`.

bdragon (approved):

abdulapopoola (approved):

PR #8729, opened 2024-01-08
Strong type `Dependabot::PullRequestCreator`

JamieMagee:


abdulapopoola (approved):

common/lib/dependabot/pull_request_creator.rb
```diff
@@ -281,15 +411,19 @@ def labeler
...
 
+    sig { returns(T::Boolean) }
     def includes_security_fixes?
       vulnerabilities_fixed.values.flatten.any?
     end
 
+    sig { params(dependency: Dependabot::Dependency).returns(T::Boolean) }
     def requirements_changed?(dependency)
-      (dependency.requirements - dependency.previous_requirements).any?
+      (dependency.requirements - T.must(dependency.previous_requirements)).any?
```
bdragon, line 426: This isn't new but seeing  the`T.must` now is making me wonder, what if `dependency.previous_requirements` is nil?

bdragon (approved):

PR #8730, opened 2024-01-08
Strong type `Dependabot::PullRequestUpdater`

JamieMagee:


common/lib/dependabot/pull_request_updater.rb
```diff
@@ -27,6 +68,9 @@ def initialize(source:, base_commit:, old_commit:, files:,
       @provider_metadata   = provider_metadata
     end
 
+    # TODO: Each implementation returns a client-specific type.
+    #       We should standardise this to return a `Dependabot::Branch` type instead.
```
abdulapopoola, line 72: please do you mind creating an issue for this and tagging it with the EE tag so it gets done?

abdulapopoola (approved):

common/lib/dependabot/pull_request_updater.rb
JamieMagee: #8735

bdragon (approved):

PR #8742, opened 2024-01-09
grouped security updates: use the group if one is defined

jakecoffman:
Previously we assumed that a grouped security update would just group all of the dependencies passed in the job. 

This PR will cause grouped security updates to use the one defined in the job instead.

<s>We can't merge this until the service starts sending groups for these types of jobs.</s>

I've updated this to fallback to the old behavior, so it should be mergeable now.


jakecoffman:
Also good to go, but the devcontainers smoke test is trying to tell me something is wrong.

updater/lib/dependabot/dependency_change_builder.rb
```diff
@@ -28,7 +28,12 @@ def self.create_from(**kwargs)
 
     def initialize(job:, dependency_files:, updated_dependencies:, change_source:)
       @job = job
-      @dependency_files = dependency_files
+
+      dir = Pathname.new(job.source.directory).cleanpath
+      @dependency_files = dependency_files.select { |f| Pathname.new(f.directory).cleanpath == dir }
+
+      raise "Missing directory in dependency files: #{dir}" unless @dependency_files.any?
```
abdulapopoola, line 35: Should we have a well defined exception here? That way it's easier to track in Sentry and other places?

updater/lib/dependabot/dependency_group_engine.rb
```diff
@@ -21,6 +21,22 @@ class DependencyGroupEngine
     class ConfigurationError < StandardError; end
 
     def self.from_job_config(job:)
+      if job.security_updates_only? && job.dependencies && job.dependencies.count > 1 && job.dependency_groups.empty?
```
abdulapopoola, line 24: should this be refactored into a function: `is_grouped_security_update`?

updater/lib/dependabot/updater/operations/group_update_all_versions.rb
```diff
@@ -21,9 +21,10 @@ class GroupUpdateAllVersions
         include GroupUpdateCreation
 
         def self.applies_to?(job:)
-          return false if job.security_updates_only?
           return false if job.updating_a_pull_request?
-          return false if job.dependencies&.any?
+          if Dependabot::Experiments.enabled?(:grouped_security_updates_disabled) && job.security_updates_only?
+            return false
+          end
```
abdulapopoola, line 27: this is a bit hard to read; can it be simplified with an `unless` check (maybe?)

updater/lib/dependabot/dependency_group_engine.rb
jakecoffman: It's kind of janky because this is a class method. I'd have to define it on `Job` for it to make sense, but that seems really unclear to me: `Job.grouped_security_update?` will have a changing meaning as we move forward. I think we should leave this ugly and then remove it when the API starts sending groups.

updater/lib/dependabot/updater/operations/group_update_all_versions.rb
jakecoffman: I can't tell how `unless` would make this less confusing.

This flag was added in https://github.com/dependabot/dependabot-core/pull/8529, I'm only moving it over in this PR. We could add an experiment and turn it on which will clean this logic up. I can do that in a followup PR.

updater/lib/dependabot/dependency_change_builder.rb
jakecoffman: Raising like this will result in a RuntimeError exception. Just checked and we have 161 places in the code that raises in the same way. 

For unique scenarios and sanity checks I think raising like this is ok. It would quickly become unmanageable to create a new exception class for each one. Sentry is good about grouping by stack trace and not the exception class name so I don't think it will make a difference, just more code to manage.

jurre: When exactly does this happen @jakecoffman?

jakecoffman: If the file_fetcher didn't set the directory (meaning the ecosystem didn't use the `fetch_file_if_present`), or the directory wasn't cleaned and so the comparison fails.

It's more of a sanity check. It happened once while I was working on this PR and resulted in a misleading error so I added this raise to point us in the right direction if it happens again.

jurre (approved):

jakecoffman:
I deployed this PR and saw exceptions where a grouped security update with 1 dependency wasn't getting picked up by the GroupUpdateAllVersions operation. So in https://github.com/dependabot/dependabot-core/pull/8742/commits/06673bf7bcc89dbb1e6843d759a2fd5232b5f993 I swapped the check for an implicit group to be if `directories` is present which works better.

This will all be cleaned up once the API sends groups! 😓

jakecoffman:
Redeployed and it looks good!

PR #8746, opened 2024-01-09
Test that file_parser reads .proj files

ryanbrandenburg:
Right now if Dependabot is checking a directory which contains only .proj files (but that .proj file may reference a csproj higher up) it will fail with  `No project file or packages.config!`. That's because the regex used in this case is `/\.[a-z]{2}proj$/`, if we make the 2 characters optional we get the .proj file we were expecting.

### DIFF
--- a/nuget/lib/dependabot/nuget/file_parser.rb
+++ b/nuget/lib/dependabot/nuget/file_parser.rb
@@ -74,7 +74,7 @@ module Dependabot
       end
 
       def project_files
-        projfile = /\.[a-z]{2}proj$/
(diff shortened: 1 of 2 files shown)


brettfo (approved):

ryanbrandenburg:

bdragon (approved):
Looks good. I was just looking at the FileUpdater and this seems consistent with what it's doing: https://github.com/dependabot/dependabot-core/blob/763f4449c2898c6ed3bd6c9c36beabde6f33f32b/nuget/lib/dependabot/nuget/file_updater.rb#L161-L163

PR #8747, opened 2024-01-09
Raise a proper user error when `package.json` includes just a dummy string

deivid-rodriguez:


abdulapopoola:
on 2nd thoughts; I think this error might be common across ecosystems; should it be moved to the FileParser helper?

deivid-rodriguez:
🤔, the fact that the parsed `package.json` file should be a Ruby Hash seems quite specific to NPM ecosystem to me! I guess we have similar situations in other ecosystems but it's not clear to me how to generalize this error control.

jeffwidman (approved):

PR #8748, opened 2024-01-09
patch user-level `NuGet.Config` before invoking dotnet/nuget tools

### LINKED ISSUE #8721: honor `nuget-feed` sources from `dependabot.yml`
The NuGet updater doesn't consider feed sources from `dependabot.yml` which means if no `NuGet.config` is present, it always falls back to querying `api.nuget.org` which might not be the desired behavior.

E.g., consider this `dependabot.yml` snippet:

``` yaml
version: 2
registries:
  public-nuget:
    type: nuget-feed
  github-nuget:
    type: nuget-feed
    url: https://nuget.pkg.github.com/ExampleOrganization/index.json
    username: some-user
    password: ${{secrets.NUGET_SECRET_GITHUB}}
# ...
```

In this case if no `NuGet.config` file is present, the GitHub feed should still be used.

These custom feeds should also be passed to the C# tool.  Probably the easiest/cleanest way to do this is to modify the file `~/.nuget/NuGet/NuGet.Config` so that the real NuGet and dotnet tools can use it.

brettfo:
When dependabot is launched, private NuGet feed credentials might be passed in via the `credentials` object.  These credentials are used to query NuGet, but when the C# command line tool is launched, these credentials are not passed through.  Since the command line tool may start other dotnet or NuGet processes, the proper way to propagate credentials is to patch the user-level `NuGet.Config` file located at `~/.nuget/NuGet/NuGet.Config`.  This patching is handled in a wrapper function that ensures the original config file is restored after calling the command line tool.  In production this file restore shouldn't matter, because a single dependabot run launches a new container then it's discarded, but when running locally it's best to restore this file.

Fixes #8721.

### DIFF
--- a/nuget/helpers/build
+++ b/nuget/helpers/build
@@ -38,3 +38,8 @@ dotnet clean
 
 echo "verifying NuGetUpdater tool"
 "$install_dir/NuGetUpdater/NuGetUpdater.Cli" --version
+
+if [ ! -f "$HOME/.nuget/NuGet/NuGet.Config" ]; then
+  echo "user-level NuGet.Config not found; credential patching will not work"
+  exit 1
+fi
--- a/nuget/lib/dependabot/nuget/file_updater.rb
+++ b/nuget/lib/dependabot/nuget/file_updater.rb
@@ -62,8 +62,9 @@ module Dependabot
 
           next unless project_dependencies.any? { |dep| dep.name.casecmp(dependency.name).zero? }
 
-          NativeHelpers.run_nuget_updater_tool(repo_contents_path, proj_path, dependency, !dependency.top_level?,
-                                               credentials)
+          NativeHelpers.run_nuget_updater_tool(repo_root: repo_contents_path, proj_path: proj_path,
+                                               dependency: dependency, is_transitive: !dependency.top_level?,
+                                               credentials: credentials)
           update_ran = true
         end
 
@@ -78,8 +79,9 @@ module Dependabot
           project_file = project_files.first
           proj_path = dependency_file_path(project_file)
 
-          NativeHelpers.run_nuget_updater_tool(repo_contents_path, proj_path, dependency, !dependency.top_level?,
-                                               credentials)
+          NativeHelpers.run_nuget_updater_tool(repo_root: repo_contents_path, proj_path: proj_path,
+                                               dependency: dependency, is_transitive: !dependency.top_level?,
+                                               credentials: credentials)
           return true
         end
 
--- a/nuget/lib/dependabot/nuget/file_updater.rb
+++ b/nuget/lib/dependabot/nuget/file_updater.rb
@@ -62,7 +62,8 @@ module Dependabot
 
           next unless project_dependencies.any? { |dep| dep.name.casecmp(dependency.name).zero? }
 
-          NativeHelpers.run_nuget_updater_tool(repo_contents_path, proj_path, dependency, !dependency.top_level?)
+          NativeHelpers.run_nuget_updater_tool(repo_contents_path, proj_path, dependency, !dependency.top_level?,
+                                               credentials)
           update_ran = true
         end
 
@@ -77,7 +78,8 @@ module Dependabot
           project_file = project_files.first
           proj_path = dependency_file_path(project_file)
 
-          NativeHelpers.run_nuget_updater_tool(repo_contents_path, proj_path, dependency, !dependency.top_level?)
+          NativeHelpers.run_nuget_updater_tool(repo_contents_path, proj_path, dependency, !dependency.top_level?,
+                                               credentials)
           return true
         end
 
--- a/nuget/lib/dependabot/nuget/native_helpers.rb
+++ b/nuget/lib/dependabot/nuget/native_helpers.rb
@@ -48,7 +48,7 @@ module Dependabot
       end
 
       # rubocop:disable Metrics/MethodLength
-      def self.run_nuget_updater_tool(repo_root, proj_path, dependency, is_transitive, credentials)
+      def self.run_nuget_updater_tool(repo_root:, proj_path:, dependency:, is_transitive:, credentials:)
         exe_path = File.join(native_helpers_root, "NuGetUpdater", "NuGetUpdater.Cli")
         command = [
           exe_path,
--- a/nuget/lib/dependabot/nuget/native_helpers.rb
+++ b/nuget/lib/dependabot/nuget/native_helpers.rb
@@ -86,10 +86,10 @@ module Dependabot
 
         puts "running NuGet updater:\n" + command
 
-        NuGetConfigCredentialHelpers.patch_nuget_config_for_action(credentials, lambda {
+        NuGetConfigCredentialHelpers.patch_nuget_config_for_action(credentials) do
           output = SharedHelpers.run_shell_command(command, fingerprint: fingerprint)
           puts output
-        })
+        end
       end
       # rubocop:enable Metrics/MethodLength
     end
--- a/nuget/lib/dependabot/nuget/native_helpers.rb
+++ b/nuget/lib/dependabot/nuget/native_helpers.rb
@@ -1,6 +1,8 @@
 # typed: true
 # frozen_string_literal: true
 
+require_relative "nuget_config_credential_helpers.rb"
+
 module Dependabot
   module Nuget
     module NativeHelpers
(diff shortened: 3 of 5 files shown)


nuget/lib/dependabot/nuget/file_updater.rb
```diff
@@ -62,7 +62,8 @@ def try_update_projects(dependency)
 
           next unless project_dependencies.any? { |dep| dep.name.casecmp(dependency.name).zero? }
 
-          NativeHelpers.run_nuget_updater_tool(repo_contents_path, proj_path, dependency, !dependency.top_level?)
+          NativeHelpers.run_nuget_updater_tool(repo_contents_path, proj_path, dependency, !dependency.top_level?,
+                                               credentials)
```
deivid-rodriguez, line 66: Not necessarily for this PR, just general advice:

Ruby methods with more than 2 or 3 positional parameters get out of hand easily, because it's not easy to rememeber the order of parameters so you need to check signature every time and makes it easy to make mistakes.

I recommend to use keyword parameters, like `credentials: credentials`.

nuget/lib/dependabot/nuget/nuget_config_credential_helpers.rb
```diff
@@ -0,0 +1,70 @@
+# typed: true
+# frozen_string_literal: true
+
+module Dependabot
+  module Nuget
+    module NuGetConfigCredentialHelpers
+      def self.user_nuget_config_path
+        home_directory = Dir.home
+        File.join(home_directory, ".nuget", "NuGet", "NuGet.Config")
```
edgarrs, line 9: Does the casing works in Linux? Windows is case insensitive so we won't have a problem.

nuget/lib/dependabot/nuget/nuget_config_credential_helpers.rb
```diff
@@ -0,0 +1,70 @@
...
+      end
+
+      def self.temporary_nuget_config_path
+        user_nuget_config_path + "_ORIGINAL"
```
edgarrs, line 13: should this be a "random" name or prefix in case a previous run fail and didn't rename the file back?

nuget/lib/dependabot/nuget/nuget_config_credential_helpers.rb
```diff
@@ -0,0 +1,70 @@
...
+          source_name = "credentialed_source_#{i + 1}"
+          package_sources << "    <add key=\"#{source_name}\" value=\"#{c.fetch('url')}\" />"
+          package_source_credentials << "    <#{source_name}>"
+          package_source_credentials << "      <add key=\"Username\" value=\"user\" />"
+          package_source_credentials << "      <add key=\"ClearTextPassword\" value=\"#{c.fetch('token')}\" />"
+          package_source_credentials << "    </#{source_name}>"
+        end
+
+        nuget_config = <<~NUGET_XML
+          <?xml version="1.0" encoding="utf-8"?>
```
edgarrs, line 36: Is it possible that the original nuget.config file had different sections than packagesSources and packageSourceCredentials that are needed for the repo to work?

nuget/helpers/build
```diff
@@ -38,3 +38,8 @@ dotnet clean
 
 echo "verifying NuGetUpdater tool"
 "$install_dir/NuGetUpdater/NuGetUpdater.Cli" --version
+
+if [ ! -f "$HOME/.nuget/NuGet/NuGet.Config" ]; then
+  echo "user-level NuGet.Config not found; credential patching will not work"
```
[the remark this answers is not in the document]
brettfo: It should always be there, it's really just to ensure the container was built correctly.  If the file isn't at that location, then the credential patching won't work and it'll be really difficult to figure out why.

This change is very similar to one nearer the top of the file that ensures the nuget.sln file is present; really it's just a proxy to make sure the git submodule was properly initialized, because the errors you see if not are too difficult to figure out.

nuget/lib/dependabot/nuget/nuget_config_credential_helpers.rb
brettfo: This is the exact casing used in Linux and this file is placed during the dotnet install.

nuget/lib/dependabot/nuget/nuget_config_credential_helpers.rb
brettfo: It shouldn't matter.  When actually running, we'll only use this container for a single job.  I considered patching this file as a part of the job starting, but then it would have been impossible to test, so I instead use the wrapper below.

nuget/lib/dependabot/nuget/nuget_config_credential_helpers.rb
brettfo: No, this file has a set content that's put down during the install and these are the only things we should need.  We're editing this file because the repo itself (cloned to a different directory) might have its own nuget.config files, but the behavior of nuget is to also patch in the contents of this file which we completely controll.

nuget/lib/dependabot/nuget/file_updater.rb
brettfo: Added commit with the change.

JoeRobich (approved):
lgtm

nuget/lib/dependabot/nuget/nuget_config_credential_helpers.rb
```diff
@@ -0,0 +1,69 @@
...
+      def self.patch_nuget_config_for_action(credentials, action)
+        add_credentials_to_nuget_config(credentials)
+        begin
+          action.call
+        rescue StandardError
+        ensure
+          restore_user_nuget_config
+        end
+      end
+      # rubocop:enable Lint/SuppressedException
```
bdragon, line 66: It might be a bit more idiomatic/ruby-ish to have this method accept a block ([this example](https://github.com/dependabot/dependabot-core/blob/763f4449c2898c6ed3bd6c9c36beabde6f33f32b/common/lib/dependabot/shared_helpers.rb#L254) comes to mind) by tossing an `&` in front of the second parameter

```ruby
def self.with_patched_nuget_config(credentials, &action)
  add_credentials_to_nuget_config(credentials)
  begin
    action.call
  rescue StandardError
  ensure
    restore_user_nuget_config
  end
end
```

which would change the calling code [above](https://github.com/dependabot/dependabot-core/pull/8748/files#diff-ed307fbccf4b647ac64a7cde4b3449b7a0fd61fe90247280449ea2fbcb29404dR89) to this:

```ruby
NuGetConfigCredentialHelpers.with_patched_nuget_config(credentials) do
  output = SharedHelpers.run_shell_command(command, fingerprint: fingerprint)
  puts output
end
```

nuget/lib/dependabot/nuget/file_updater.rb
```diff
@@ -62,7 +62,9 @@ def try_update_projects(dependency)
 
           next unless project_dependencies.any? { |dep| dep.name.casecmp(dependency.name).zero? }
 
-          NativeHelpers.run_nuget_updater_tool(repo_contents_path, proj_path, dependency, !dependency.top_level?)
+          NativeHelpers.run_nuget_updater_tool(repo_root: repo_contents_path, proj_path: proj_path,
+                                               dependency: dependency, is_transitive: !dependency.top_level?,
+                                               credentials: credentials)
```
bdragon, line 67: Just noting that the config patching/restoring is happening in a loop here. I'm guessing the number of project files is tiny in practice so no reason for concern.

nuget/lib/dependabot/nuget/nuget_config_credential_helpers.rb
brettfo: Thank you, this is great!  I didn't know Ruby had this feature.  I'll update this shortly.

nuget/lib/dependabot/nuget/file_updater.rb
brettfo: Correct, there usually aren't a ton of project files.  If it becomes a bottleneck we can move the `NuGet.Config` patching up a level, but for now I wanted to scope it as closely to the CLI tool as possible.

nuget/lib/dependabot/nuget/nuget_config_credential_helpers.rb
brettfo: Done.

nuget/lib/dependabot/nuget/native_helpers.rb
```diff
@@ -1,6 +1,8 @@
 # typed: true
 # frozen_string_literal: true
 
+require_relative "nuget_config_credential_helpers.rb"
```
bdragon, line 4: The `.rb` extension isn't necessary for `require_relative`. I don't think it causes any harm but since we don't do it elsewhere, might as well leave it off

```suggestion
require_relative "nuget_config_credential_helpers"
```

bdragon (approved):

Issue #8642, opened 2023-12-17
NuGet updates convert csproj files from CRLF to LF


adamralph:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

NuGet

### Manifest location and content before the Dependabot update

https://github.com/adamralph/minver/blob/main/MinVer.Lib/MinVer.Lib.csproj

### dependabot.yml content

https://github.com/adamralph/minver/blob/main/.github/dependabot.yml

### Updated dependency

NuGet.Versioning from 6.7.0 to 6.8.0

### What you expected to see, versus what you actually saw

Expected: Only the dependency updated, with line endings left the same.

Actual: The dependency updated, but all line endings change from CRLF to LF.

### Native package manager behavior

Only the dependency update, with line endings left the same.

### Images of the diff or a link to the PR, issue, or logs

https://github.com/adamralph/minver/pull/936/files

adamralph:
More examples:

- https://github.com/adamralph/bullseye/pull/932/files
- https://github.com/adamralph/minver/pull/960/files

fuzzzerd:
I have had this happen in as well, unfortunately they're private repositories so I can't share a link to the pull, but the behavior is the same as the linked pulls here.

fuzzzerd:
This issue just cropped up again on another project of mine as well:

![diff with line endings changed](https://github.com/dependabot/dependabot-core/assets/711755/df4b8f44-9e7d-4a86-a71d-af1bef5decd9)

aaron-kruse:
Same issue here also in a private repo, I believe it started a few months ago. It might kind of go without saying, but this is on a Windows machine.

If it helps, I also found this in the update logs:
```
updater | Update complete.
updater | Fixing mismatched Unix line endings for [Web/Web.csproj].
```

With main checked out:
```
git ls-files --eol Web.csproj
i/crlf  w/crlf  attr/text=auto          Web.csproj
```

With a Dependabot branch checked out where line endings have been replaced:
```
git ls-files --eol Web.csproj
i/lf    w/crlf  attr/text=auto          Web.csproj
```

Content of `.gitattributes` file:
```
*		text=auto

# Scripts - override since bash throws errors if crlf is used
*.sh	text eol=lf
```

Note that with either of the two branches above checked out, when I open the file in Notepad++, the line endings are always crlf (I remember reading that the Git client does some "magic" using `.gitattributes` among other things to auto-convert line endings during checkouts/commits).

[...]

PR #8749, opened 2024-01-09
No encoding conversions on cloned files

deivid-rodriguez:
There's this repo where the `.npmrc` file has `UTF16_LE` encoding. Dependabot fails to run on it:

```
$ SECURITY_ADVISORIES='[{"dependency-name":"follow-redirects","patched-versions":[],"unaffected-versions":[],"affected-versions":["< 1.14.7"]},{"dependency-name":"follow-redirects","patched-versions":[],"unaffected-versions":[],"affected-versions":["< 1.14.8"]},{"dependency-name":"follow-redirects","patched-versions":[],"unaffected-versions":[],"affected-versions":["< 1.15.4"]}]'  bin/dry-run.rb npm_and_yarn "Stuart-Wilcox/gameroom" --dir="/client" --dep="follow-redirects" --security-updates-only --updater-options=grouped_updates_experimental_rules,record_ecosystem_versions,record_update_job_unknown_error --commit=5a02652ca0840b728ddd96dee9afa0f935829dad
=> cloning into /home/dependabot/tmp/Stuart-Wilcox/gameroom
=> checking out commit 5a02652ca0840b728ddd96dee9afa0f935829dad
/home/dependabot/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb:560:in `match?': invalid byte sequence in UTF-8 (ArgumentError)
	from /home/dependabot/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb:560:in `skip_package_lock?'
	from /home/dependabot/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb:96:in `npm_files'
	from /home/dependabot/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb:78:in `fetch_files'
	from /home/dependabot/dependabot-updater/vendor/ruby/3.1.0/gems/sorbet-runtime-0.5.11142/lib/types/private/methods/call_validation.rb:256:in `bind_call'
	from /home/dependabot/dependabot-updater/vendor/ruby/3.1.0/gems/sorbet-runtime-0.5.11142/lib/types/private/methods/call_validation.rb:256:in `validate_call'
	from /home/dependabot/dependabot-updater/vendor/ruby/3.1.0/gems/sorbet-runtime-0.5.11142/lib/types/private/methods/_methods.rb:275:in `block in _on_method_added'
	from /home/dependabot/common/lib/dependabot/file_fetchers/base.rb:108:in `files'
	from /home/dependabot/dependabot-updater/vendor/ruby/3.1.0/gems/sorbet-runtime-0.5.11142/lib/types/private/methods/call_validation_2_7.rb:660:in `bind_call'
	from /home/dependabot/dependabot-updater/vendor/ruby/3.1.0/gems/sorbet-runtime-0.5.11142/lib/types/private/methods/call_validation_2_7.rb:660:in `block in create_validator_method_medium0'
	from bin/dry-run.rb:402:in `fetch_files'
	from bin/dry-run.rb:487:in `<main>'
```

This change fixes the problem.

It's a bit risky because it affects all files for all ecosystems that clone, but I _think_ it may work fine.

The idea is to read the raw content, without any automatic conversions to our default encoding (UTF-8).

It may still fail for files with non standard encoding at write time, but in this case, the `.npmrc` is just a support file that's just read, so this allows that case to work.

### DIFF
--- a/common/lib/dependabot/file_fetchers/base.rb
+++ b/common/lib/dependabot/file_fetchers/base.rb
@@ -184,7 +184,8 @@ module Dependabot
         repo_path = File.join(clone_repo_contents, path)
         raise Dependabot::DependencyFileNotFound, path unless File.exist?(repo_path)
 
-        content = File.read(repo_path)
+        content = File.binread(repo_path)
+
         type = if File.symlink?(repo_path)
                  symlink_target = File.readlink(repo_path)
                  "symlink"


abdulapopoola (approved):

abdulapopoola:
@JamieMagee this sounds quite similar to your fix [here](https://github.com/dependabot/dependabot-core/pull/8929). Thoughts on this PR; should we switch to this upstream fix to get it everywhere?

Tagging @jakecoffman , @jurre and @mctofu too in case they have thoughts.

JamieMagee:
@abdulapopoola I think my approach was different. I forced a conversion to UTF-8 and dropped any characters that couldn't be converted. This approach seems to defer the conversion to later.

deivid-rodriguez:
Yeah, my approach was basically to not do any encoding (or line ending conversions), just read the raw content. It fixes the crash I posted in the PR body, but should be tried carefully. If it works fine, it seems like a simpler approach that may result in less similar errors down the line!

deivid-rodriguez:
Also I think there's a possibility that this will fix https://github.com/dependabot/dependabot-core/issues/8642?

deivid-rodriguez:
This did not get much attention even if it had a clear repro. Should I close?

deivid-rodriguez:
I'll close.

PR #8750, opened 2024-01-09
Handle unexpected kub image shapes

ryanbrandenburg:
If the `image` field didn't match our regex we would throw when `.match` returned nil. It's possible we should handle a wider variety of image specs, but when all that fails we need to not throw nil exceptions.


ryanbrandenburg:

JoeRobich (approved):

bdragon (approved):

PR #8751, opened 2024-01-10
Fix potential null ref when files is empty

abdulapopoola:
This error happens because the files array is empty


jakecoffman:
This should be "fixed" now: https://github.com/dependabot/dependabot-core/pull/8782

If there are no files in the update it will raise before getting into the PR creation routine. So now we can chase down why there were no files in the update.

PR #8752, opened 2024-01-10
Fix broken nuget CI

deivid-rodriguez:
Nuget CI is currently broken.

Issues are order dependent (although seem to be happening consistently), and can be reproduced with:

```
$ rspec './spec/dependabot/nuget/update_checker/version_finder_spec.rb[1:1:15:1:1:1,1:1:17:1:1]' --seed 17384
```

The issue is dependency caching.

This PR fixes the problem by making sure specs always avoid caching, except when the caching itself is being explicitly tested.

In addition to this, this PR also introduces some changes to make it possible to actually reproduce the failure, because running specs in isolation currently fails with another error, namely:

```
$ rspec ./spec/dependabot/nuget/update_checker/version_finder_spec.rb:407

An error occurred while loading ./spec/dependabot/nuget/update_checker/version_finder_spec.rb.
Failure/Error:
        require_relative "update_checker/version_finder"
        require_relative "update_checker/property_updater"
        require_relative "update_checker/requirements_updater"
        require_relative "update_checker/dependency_finder"

        def latest_version
          # No need to find latest version for transitive dependencies unless they have a vulnerability.
          return dependency.version if !dependency.top_level? && !vulnerable?

TypeError:
  superclass mismatch for class UpdateChecker
# ./lib/dependabot/nuget/update_checker.rb:10:in `<module:Nuget>'
# ./lib/dependabot/nuget/update_checker.rb:9:in `<module:Dependabot>'
# ./lib/dependabot/nuget/update_checker.rb:8:in `<top (required)>'
# ./lib/dependabot/nuget/update_checker/version_finder.rb:7:in `require'
# ./lib/dependabot/nuget/update_checker/version_finder.rb:7:in `<top (required)>'
# ./spec/dependabot/nuget/update_checker/version_finder_spec.rb:7:in `require'
# ./spec/dependabot/nuget/update_checker/version_finder_spec.rb:7:in `<top (required)>'
```

### DIFF
--- a/nuget/lib/dependabot/nuget/file_parser/packages_config_parser.rb
+++ b/nuget/lib/dependabot/nuget/file_parser/packages_config_parser.rb
@@ -26,8 +26,6 @@ module Dependabot
         end
 
         def dependency_set
-          return parse_dependencies if CacheManager.caching_disabled?
-
           key = "#{packages_config.name.downcase}::#{packages_config.content.hash}"
           cache = PackagesConfigParser.dependency_set_cache
 
--- a/nuget/lib/dependabot/nuget/file_parser/packages_config_parser.rb
+++ b/nuget/lib/dependabot/nuget/file_parser/packages_config_parser.rb
@@ -32,10 +32,6 @@ module Dependabot
           cache = PackagesConfigParser.dependency_set_cache
 
           cache[key] ||= parse_dependencies
-
-          dependency_set = Dependabot::FileParsers::Base::DependencySet.new
-          dependency_set += cache[key]
-          dependency_set
         end
 
         private
--- a/nuget/lib/dependabot/nuget/cache_manager.rb
+++ b/nuget/lib/dependabot/nuget/cache_manager.rb
@@ -13,6 +13,8 @@ module Dependabot
       end
 
       def self.cache(name)
+        return {} if caching_disabled?
+
         @cache ||= {}
         @cache[name] ||= {}
         @cache[name]
--- a/nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
+++ b/nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
@@ -50,8 +50,6 @@ module Dependabot
         end
 
         def dependency_set(project_file:)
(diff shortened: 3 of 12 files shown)


jurre (approved):

deivid-rodriguez:
Ready for a second look @jurre!

nuget/lib/dependabot/nuget/file_parser/packages_config_parser.rb
```diff
@@ -26,16 +26,10 @@ def initialize(packages_config:)
...
-          return parse_dependencies if CacheManager.caching_disabled?
-
           key = "#{packages_config.name.downcase}::#{packages_config.content.hash}"
           cache = PackagesConfigParser.dependency_set_cache
 
           cache[key] ||= parse_dependencies
-
-          dependency_set = Dependabot::FileParsers::Base::DependencySet.new
-          dependency_set += cache[key]
-          dependency_set
```
jurre, line 38: Are we removing the caching behavior entirely with this change? I'm not quite following how removing this code doesn't alter behavior 🤔

deivid-rodriguez: I should've extracted this to a separate commit. My understanding is that this is just redundant code. The result of the (cached or not) `parse_dependencies` method should be already a `Dependabot::FileParsers::Base::DependencySet`, unless I'm missing something.

deivid-rodriguez: I believe the caching is done in the above line, which I kept!

jurre: Thanks for clarifying!

jurre (approved):

nuget/lib/dependabot/nuget/file_parser/packages_config_parser.rb
deivid-rodriguez: Awesome, I did split that commit out so the diff of the actual fix for "caching disabled mode" is easier to understand.

Also added one more missing require that running specs locally found.

PR #8755, opened 2024-01-10
Branch names use the job_directory if the directory is nil

Nishnha:
WIP

We are seeing an error around creating the branch name for grouped security updates where the `directory` is nil for some [DependencyFiles](https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/dependency_file.rb#L18).

> undefined method `directory' for nil:NilClass files.first.directory.tr(" ", "-")

The default value is [actually `"/"`](https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/dependency_file.rb#L79), so it's explicitly being set to nil somewhere.

We could pass in an empty string for the directory, or we could use the `job_directory`, which is the directory that triggered the security update job. I chose the latter approach in this PR.

### DIFF
--- a/common/lib/dependabot/pull_request_creator/branch_namer/dependency_group_strategy.rb
+++ b/common/lib/dependabot/pull_request_creator/branch_namer/dependency_group_strategy.rb
@@ -64,7 +64,7 @@ module Dependabot
         end
 
         def directory
-          T.must(files.first).directory.tr(" ", "-")
+          T.must(files.first).directory.&tr(" ", "-") || T.must(files.first).job_directory.tr(" ", "-")
         end
       end
     end
--- a/updater/lib/dependabot/dependency_change.rb
+++ b/updater/lib/dependabot/dependency_change.rb
@@ -58,10 +58,7 @@ module Dependabot
     end
 
     def updated_dependency_files_hash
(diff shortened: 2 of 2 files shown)


abdulapopoola:
@jakecoffman is this related to the [files being empty issue](https://github.slack.com/archives/C03UT3Q6V34/p1704849441188919) we were discussing?

jakecoffman:
Yeah seems like too much of a coincidence.

Nishnha:
Yeah it is for the same issue. I didn't realize this was mainly affecting nuget - I thought it was related to grouped security updates. It should let these erroring PRs get created and the `job_directory` should have the same value as `directory`

jakecoffman:
This PR made me realize a DependencyFile's directory *is* the job_directory, so I made a PR to test that theory: #8762

Indeed all the tests pass. So I don't think this PR is the right approach. We need to figure out how a DependencyFile's directory could be blank, because I don't see how that could happen:

```ruby
irb(main):006> Dependabot::DependencyFile.new(name: "requirements.txt", content:" ", directory: "").directory
=> "/"
```

It must be getting reassigned somewhere in the NuGet code.

PR #8758, opened 2024-01-11
Strict type `Dependabot::PullRequestCreator::Labeler`

JamieMagee:


common/lib/dependabot/pull_request_creator/labeler.rb
```diff
@@ -1,32 +1,54 @@
-# typed: true
+# typed: strict
```
bdragon, line 1: Just checking if this is intentionally strict and not strong like some of the other recently-typed files.

bdragon (approved):

common/lib/dependabot/pull_request_creator/labeler.rb
JamieMagee: Good catch, you're right. Thanks! I'll fix it in the morning.

JamieMagee: @bdragon I had a look, and I can't use strong because of the use of `T.unsafe` 😢

I'll have to come back to this once I properly type all the HTTP clients and can get rid of the `T.unsafe`.

PR #8405, opened 2023-11-15
fix broken updates due to incorrect directory filtering

jakecoffman:
fixes #8398

In #8331 I introduced filtering files by directory so we could achieve multi-directory updates. This filtering was too naive though as it has broken updates for a few ecosystems that have special handling for directories:

- In NuGet, [src](https://github.com/dependabot/dependabot-core/blob/641de673efbcfb0b92641f7051a7325b473dd44e/nuget/lib/dependabot/nuget/file_fetcher.rb#L98-L102) is used at in some cases
- In GitHub Actions, `.github/workflows` is automatically used if the directory is `/`

And there may be others!

These special cases resulted in the directory of the file being different than the directory of the update (the job's configured directory).

By marking the files fetched with a `job_directory`, we can remove the special cases and the filtering works as expected.

PR #8762, opened 2024-01-11
directory is the job directory

jakecoffman:
In #8405 I was under the impression that `DependencyFile`'s `directory` was the directory of the file, and name was just the basename, but that's not the case. A `DependencyFile`'s `directory` should always be the job (source) directory, and the `name` contains the relative path from the `directory` and also the filename. 

Not sure what problem I was running into before, but removing `job_directory` and filtering on directory alone is passing the smoke tests, so this is confirmed!

### DIFF
--- a/common/lib/dependabot/dependency_file.rb
+++ b/common/lib/dependabot/dependency_file.rb
@@ -14,6 +14,8 @@ module Dependabot
     sig { returns(T.nilable(String)) }
     attr_accessor :content
 
+    # This is the directory of the job source, not the directory of the file itself.
+    # The name actually contains the relative path from the job directory.
     sig { returns(String) }
     attr_accessor :directory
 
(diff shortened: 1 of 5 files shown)


deivid-rodriguez (approved):
So if I'm understanding correctly, even in a multidir group update, each dependency file will have the `directory` attribute set to the proper directory (from the ones included in the job).

jakecoffman:
@deivid-rodriguez correct! For multi-directory fetches it loops over the directories: https://github.com/dependabot/dependabot-core/blob/b05d6edd6797cbb15a28c12dbb74008618d3fe38/updater/lib/dependabot/file_fetcher_command.rb#L100-L108

For each directory it sets the job source as that directory: https://github.com/dependabot/dependabot-core/blob/b05d6edd6797cbb15a28c12dbb74008618d3fe38/updater/lib/dependabot/file_fetcher_command.rb#L79

Nishnha:
I see, so the directory wasn't being set on DependencyFiles during grouped updates because we were referencing the `job_directory` instead? Glad you spotted this!

Nishnha (approved):

PR #8763, opened 2024-01-11
Track errored dependencies in a DependencyChange

Nishnha:
During a grouped (security) update refresh, we have no way of knowing whether the `updated_dependencies` list is empty because all of the dependencies during the update errored, or if the users have manually updated the dependencies themselves, so Dependabot closes out a PR during a rebase/recreate when it should leave the PR open and explain any errors.

This PR adds in an `errored_dependencies` array to a DependencyChange. Alternatively, we could create a flag instead of an array if we don't want to do anything with these errors (other than know they occurred).

This isn't a great approach imo, because we would need to call `FileUpdater#add_errored_dependency` for every error across all ecosystems. It might be a nicer implementation if we intercept/re-raise Dependabot all FileUpdater errors and make errored_dependencies a flag. We wouldn't know which dependencies errored without checking the logs, but that might be okay.

### DIFF
--- a/common/lib/dependabot/file_updaters/base.rb
+++ b/common/lib/dependabot/file_updaters/base.rb
@@ -16,6 +16,9 @@ module Dependabot
       sig { returns(T::Array[Dependabot::DependencyFile]) }
       attr_reader :dependency_files
 
+      sig { returns(T::Array[String]) }
+      attr_reader :errored_dependencies
+
       sig { returns(T.nilable(String)) }
       attr_reader :repo_contents_path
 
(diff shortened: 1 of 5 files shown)


jakecoffman:
I think we can avoid changing all of the ecosystems since the updater typically knows why there was no update.

For example, on this line we know the dependency was up-to-date: https://github.com/dependabot/dependabot-core/blob/f8648e51faa6c410fe54af7a63747de0c715f0d8/updater/lib/dependabot/updater/group_update_creation.rb#L160-L163

Down here we know there was an error: https://github.com/dependabot/dependabot-core/blob/f8648e51faa6c410fe54af7a63747de0c715f0d8/updater/lib/dependabot/updater/group_update_creation.rb#L178-L193

PR #8765, opened 2024-01-11
defensive programming around missing dependency files during an update

### LINKED ISSUE #6487: Review `required_files_in?` necessity

jakecoffman:
We trust that the ecosystems are returning an array of dependencies but we never really check if they are and we continue with the update. It's possible due to an error that the array is empty, so let's check for that.

Also during a grouped update we filter out based on the directory, which should never result in an empty array but just in case it does let's check for that. 

Edit: Also noticed a while back that the Updater never calls `reqiured_files_in?`. I think some ecosystems have setup explicit checks that the required files were fetched, but others assume the Updater code calls `required_files_in?`. So I went ahead and called it in the Base FileFetcher's `files` method so we can simplify that check across the board. fixes #6487


github_actions/lib/dependabot/github_actions/file_fetcher.rb
```diff
@@ -11,10 +11,8 @@ class FileFetcher < Dependabot::FileFetchers::Base
       extend T::Sig
       extend T::Helpers
 
-      FILENAME_PATTERN = /^(\.github|action.ya?ml)$/
-
       def self.required_files_in?(filenames)
-        filenames.any? { |f| f.match?(FILENAME_PATTERN) }
```
abdulapopoola, line 17: any reason to remove the constant?

jakecoffman: No reason, I put it back.

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -104,7 +108,16 @@ def target_branch
 
       sig { returns(T::Array[DependencyFile]) }
       def files
-        @files ||= T.let(fetch_files, T.nilable(T::Array[DependencyFile]))
+        return T.must(@files) if defined? @files
```
deivid-rodriguez, line 111: Are both this and the `T.must` in the last line of the method needed?

deivid-rodriguez (approved):
This is really great! It makes file fetchers easier and with a better default behavior!

Also this PR closes https://github.com/dependabot/dependabot-core/issues/6487 😃.

common/lib/dependabot/file_fetchers/base.rb
jakecoffman: Yes but I think I can clean this up if I move the `@files` creation to the constructor.

jakecoffman: It doesn't seem to be the Ruby style (at least the Dependabot style?) to initialize all variables in the constructor but it makes using types easier.

brbayes-msft (approved):

rvelusamy:
This PR is introducing an issue for https://github.com/dependabot/dependabot-core/issues/8867. Could someone check this?

Issue #8593, opened 2023-12-11
Helm Values error undefined method `named_captures' for nil:NilClass


caleblloyd:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

Docker (Helm)

### Manifest location and content before the Dependabot update

helm/charts/nats/values.yaml

### dependabot.yml content

https://github.com/nats-io/k8s/blob/c2df016039b443cef65d2bbacc2947ec7ed6dd4c/.github/dependabot.yml

### Updated dependency

Docker image `nats` should be updated from `2.10.5-alpine` to `2.10.6-alpine`

### What you expected to see, versus what you actually saw

On November 30, 2023 the Dependabot job ran fine.

On December 1, 2023, the Dependabot job started having errors with the Ruby exception in the logs.

There were no code changes committed to the repository on either of these dates.

### Images of the diff or a link to the PR, issue, or logs

```

  proxy | 2023/12/01 18:11:32 proxy starting, commit: 6cffd6fae1b2f713f2d837bc45fe916f855c821d
  proxy | 2023/12/01 18:11:32 Listening (:1080)
updater | 2023-12-01T18:11:32.904176528 [757272134:main:WARN:src/devices/src/legacy/serial.rs:222] Detached the serial input due to peer close/error.
updater | time="2023-12-01T18:11:34Z" level=info msg="guest starting" commit=17df297b9449ef5936111b658329653bcfc0c9bc
updater | time="2023-12-01T18:11:34Z" level=info msg="starting job..." fetcher_timeout=10m0s job_id=757272134 updater_timeout=45m0s updater_version=e83cd566eb627cd83ed43645e38ef9caff47ed79-docker
updater | 2023/12/01 18:11:36 INFO Raven 3.1.2 ready to catch errors
updater | 2023/12/01 18:11:37 INFO <job_757272134> Starting job processing
  proxy | 2023/12/01 18:11:38 [002] GET https://api.github.com:443/repos/nats-io/k8s
  proxy | 2023/12/01 18:11:38 [002] * authenticating github api request with token for api.github.com
  proxy | 2023/12/01 18:11:38 [002] 200 https://api.github.com:443/repos/nats-io/k8s
  proxy | 2023/12/01 18:11:38 [004] GET https://api.github.com:443/repos/nats-io/k8s/git/refs/heads/main
  proxy | 2023/12/01 18:11:38 [004] * authenticating github api request with token for api.github.com
[...]

PR #8766, opened 2024-01-11
Handle non-matching tags

landongrindheim:
Fixes https://github.com/dependabot/dependabot-core/issues/8593

### DIFF
--- a/docker/lib/dependabot/docker/file_parser.rb
+++ b/docker/lib/dependabot/docker/file_parser.rb
@@ -171,11 +171,15 @@ module Dependabot
       end
 
       def parse_helm(img_hash)
-        repo = img_hash.fetch("repository", nil)
         tag_value = img_hash.key?("tag") ? img_hash.fetch("tag", nil) : img_hash.fetch("version", nil)
+        return [] unless tag_value
+
+        repo = img_hash.fetch("repository", nil)
         registry = img_hash.fetch("registry", nil)
 
-        tag_details = tag_value.to_s.match(TAG_WITH_DIGEST).named_captures
+        tag_match = tag_value.to_s.match(TAG_WITH_DIGEST)
+        tag_details = tag_match&.named_captures || {}
+
         tag = tag_details["tag"]
         digest = tag_details["digest"]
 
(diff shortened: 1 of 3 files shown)


abdulapopoola (approved):

docker/lib/dependabot/docker/file_parser.rb
```diff
@@ -171,11 +171,15 @@ def manifest_files
...
         tag_value = img_hash.key?("tag") ? img_hash.fetch("tag", nil) : img_hash.fetch("version", nil)
+        return [] unless tag_value
+
+        repo = img_hash.fetch("repository", nil)
         registry = img_hash.fetch("registry", nil)
 
-        tag_details = tag_value.to_s.match(TAG_WITH_DIGEST).named_captures
+        tag_match = tag_value.to_s.match(TAG_WITH_DIGEST)
+        tag_details = tag_match&.named_captures || {}
+
```
deivid-rodriguez, line 182: I think the fallback to `{}` if no named captures are found is not needed? I'm also not sure which effect it would have. If the tag content did not match `TAG_WITH_DIGEST`, we'd probably want to either raise a user error, or ignore the dependency, but I'm not sure if this change achieves that? I think I'd leave it as is for now.

deivid-rodriguez (commented):
Just a small comment on part of the fix, but the other part and the test look good!

docker/lib/dependabot/docker/file_parser.rb
landongrindheim: Good call! I botched the amending of my commit. I'll fix this and re-request a review 😄

docker/lib/dependabot/docker/file_parser.rb
```diff
@@ -171,17 +171,19 @@ def manifest_files
       end
 
       def parse_helm(img_hash)
-        repo = img_hash.fetch("repository", nil)
         tag_value = img_hash.key?("tag") ? img_hash.fetch("tag", nil) : img_hash.fetch("version", nil)
-        registry = img_hash.fetch("registry", nil)
+        return [] unless tag_value
+
+        repo = img_hash.fetch("repository", nil)
+        return [] unless repo
```
deivid-rodriguez, line 178: Nice restructure!

deivid-rodriguez (approved):

PR #8778, opened 2024-01-12
Fix vscode rdbg launch command

JamieMagee:
The `omnibus` directory doesn't include `.bundle/bin/rspec` anymore, so I use the `rspec` command from `$PATH` instead. I also changed it to run the current file by default, instead of asking you to input the file path.

I also moved the tests command to the top of the list, so it runs by default when  hitting F5. I think it's more common than debugging `dry-run.rb`.


ryanbrandenburg (approved):

abdulapopoola (approved):

deivid-rodriguez (approved):
Not familiar with vscode, but the changes here look sensible!

PR #8793, opened 2024-01-14
Fix type definition for approvers in pull request creator

andrcuns:
Current type for the main PullRequestCreator interface prevents adding reviewers for Gitlab configuration because it is a bit more elaborate and accepts hash with different types of reviewers.


common/lib/dependabot/pull_request_creator/gitlab.rb
```diff
@@ -150,7 +150,7 @@ def add_approvers_to_merge_request(merge_request)
       end
 
       def approvers_hash
-        @approvers_hash ||= approvers&.transform_keys(&:to_sym) || {}
```
andrcuns, line 153: Since technically this regression is already released, might make sense to just force using specific type and not have to perform this conversion.

abdulapopoola: Thanks @andrcuns ; I assume this regression impacts GitLab's interfaces specifically (where the approvers can be a hash versus array of strings?

Is this accurate?

Also tagging @JamieMagee , @deivid-rodriguez , and @jakecoffman

JamieMagee: @andrcuns could you share the error message you got?

@abdulapopoola it is likely correct, but we don't test dependabot-core against GitLab so it's hard for us to see.

andrcuns: @JamieMagee Yes, this is specifically for the Gitlab's interface. Maybe other implementations could be impacted as well, but I did not go through other vendors. The error would look something like this when trying to set gitlab specific reviewers/approvers.

```
Parameter 'reviewers': Expected type T.nilable(T::Array[String]), got type Hash with value {:approvers=>[3], :reviewers=>[2]}
Caller: /home/dependabot/app/vendor/bundle/ruby/3.1.0/gems/sorbet-runtime-0.5.11190/lib/types/private/methods/call_validation.rb:217
Definition: /home/dependabot/app/vendor/bundle/ruby/3.1.0/gems/dependabot-common-0.240.0/lib/dependabot/pull_request_creator.rb:174
```

This is because in GitLab, you can set `approvers`, `reviewers` and `approver_groups`.

JamieMagee: Thanks! Yeah, this looks good for now.

In the long run, I would like to refactor this so we're not passing around platform-specific responses to the platform-agnostic `PullRequestCreator`. we likely need some common interface that all platform-specific responses get translated to, with a `platformSpecificMetadata` property. Or something like that.

JamieMagee (approved):

jakecoffman (approved):

PR #8807, opened 2024-01-16
allow recusive fetches for python and rust

jubrad:
### Purpose:
Resolve https://github.com/dependabot/dependabot-core/issues/8428

Following [this comment](https://github.com/dependabot/dependabot-core/issues/8428#issuecomment-1822719969) in issue 8428, I'm adding additional functions to python and rust file fetchers to enable recursive submodules. 
>Hi!
>We recently started cloning full python repositories when running updates, however we don't yet clone submodules.
>We'd need to add [this method](https://github.com/dependabot/dependabot->core/blob/77bd32723fb202b9ec4861574347d258c446162f/npm_and_yarn/lib/dependabot/npm_and_yarn/file_fetcher.rb#L90-L92) to the python fetcher, like we do for Javascript.
>Would you be interested in creating a PR to fix this?

### Notes:
I don't see any language specific tests for these, I'd be happy to add one but I'm sure where to do so, guidance welcome. 

This is also my first commit here, and I'm not super clear on environment setup and testing, but I was able to spin up a python docker container and run dry run on one of my repositories that was hitting this issue and it seems to work.

### Questions:
Is there a good reason this isn't in `common/lib/dependabot/file_fetchers/base.rb`? 

cc @deivid-rodriguez


abdulapopoola:
Thanks for getting this started; here is an example [commit](https://github.com/dependabot/dependabot-core/commit/2f8a624d74de0ee214954894e5a64ac5f1ce0e05) by @bdragon that might be useful

jakecoffman:
I made all ecosystems recurse submodules in https://github.com/dependabot/dependabot-core/pull/9278 so you should be good now!

jubrad:
@jakecoffman you rock!! thanks

PR #8812, opened 2024-01-17
Handle malformed Global.json

ryanbrandenburg:
There exist some repos which have files named "global.json" which are not formatted as "a global.json file". We must also be resilient against repos which have malformed JSON. Let's be more resilient to that kind of scenerio.


nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Files/JsonBuildFile.cs
```diff
@@ -27,6 +27,18 @@ public void UpdateProperty(string[] propertyPath, string newValue)
...
     {
-        Node = new Lazy<JsonNode?>(() => JsonHelper.ParseNode(Contents));
+        Node = new Lazy<JsonNode?>(() =>
+        {
+            try
+            {
+                return JsonHelper.ParseNode(Contents);
+            }
+            catch (System.Text.Json.JsonException)
+            {
```
abdulapopoola, line 37: Please is it possible to surface an error here: "Invalid configuration file". We do have this pattern in other parts of the code and I wonder if we can continue with that practice.

brettfo (approved):

JoeRobich (approved):

jakecoffman (approved):

PR #8815, opened 2024-01-17
Use correct global.json discovery logic when updating.

JoeRobich:
This change was brought up by https://github.com/dependabot/dependabot-core/pull/8812

We were locating an invalid Global.json and attempting an upgrade. However, this global.json was outside of the locations that would have it included in the build. An underlying issue was how we were discovering global.json path in the updater. This change moves us to locating global.json using appropriate discovery logic. 

> The .NET SDK looks for a global.json file in the current working directory (which isn't necessarily the same as the project directory) or one of its parent directories.
https://learn.microsoft.com/en-us/dotnet/core/tools/global-json


brettfo (approved):

JoeRobich:
@JamieMagee Are you authorized to merge?

JamieMagee:
@JoeRobich unfortunately not 😢

JoeRobich:
@bdragon Hi, could I get a merge when you feel good about this change? Or, if someone else is on rotation, could you tag them in?

jakecoffman (approved):
I'll get this deployed, thanks!

PR #8818, opened 2024-01-17
Migrate from `sentry-raven` to `sentry-ruby`

JamieMagee:
This migrates from the deprecated [`sentry-raven`][1] gem to the replacement [`sentry-ruby`][2] gem. `sentry-raven` was deprecated in March 2021.

The changes mostly follow the [Sentry migration guide][3]. The largest changes were around event processors. Previously, `sentry-raven` allowed you to inherit from `Raven::Processor` and pass it to `processors` during configuration. In `sentry-ruby` that has been replaced with `before_send` which accepts a single lambda. I've recreated the same behaviour by using a reducer to apply processors consecutively. See the ["scrubbing sensitive data"][4] and ["Filtering for Ruby"][5] documentation for more information.

Once this is merged, it opens up the possibility of using the [`sentry-opentelemetry`][6] gem as well.

[1]: https://rubygems.org/gems/sentry-raven
[2]: https://rubygems.org/gems/sentry-ruby
[3]: https://docs.sentry.io/platforms/ruby/migration/
[4]: https://docs.sentry.io/platforms/ruby/data-management/sensitive-data/
[5]: https://docs.sentry.io/platforms/ruby/configuration/filtering/
[6]: https://docs.sentry.io/platforms/ruby/performance/instrumentation/opentelemetry/


deivid-rodriguez:
Awesome, it was about time! See https://github.com/dependabot/dependabot-core/pull/7553 for previous art. This PR can supersede that one but maybe you find something useful in the code/comments there!

updater/lib/dependabot/sentry/sentry_context_processor.rb
```diff
@@ -0,0 +1,25 @@
...
+        hint: T::Hash[Symbol, T.untyped]
+      )
+      .returns(Sentry::Event)
+  end
+  def process(event, hint)
+    if (exception = hint[:exception])
+      exception.raven_context.each do |key, value|
+        event.send("#{key}=", value)
+      end
+    end
```
JamieMagee, line 22: Copied from the migration guide here: https://docs.sentry.io/platforms/ruby/migration/#exceptionraven_context

updater/lib/dependabot/setup.rb
```diff
@@ -39,7 +41,8 @@
     swift
   )}x
 
-  config.processors += [ExceptionSanitizer]
+  config.before_send = ->(event, hint) { Dependabot::Sentry::Processor.process_chain(event, hint) }
+  config.propagate_traces = false
```
JamieMagee, line 45: This prevents Sentry adding `Sentry-Trace` and `Baggage` headers by default to all outgoing requests. It's designed to make it easier to do distributed tracing between microservices or frontend/backend. As we don't have that architecture, and it currently breaks our smoke test cache, I've chosen to disable it by default.

If we want to get this distributed trace information, we'll need to ignore these headers in our updater proxy.

References:
- https://github.com/getsentry/sentry-ruby/blob/7666c94f1f2671e4c42cd8c4a6991e3ce7da56f9/sentry-ruby/lib/sentry/configuration.rb#L189-L191
- https://www.w3.org/TR/baggage/

updater/lib/dependabot/service.rb
```diff
@@ -100,21 +101,19 @@ def update_dependency_list(dependency_snapshot:)
...
+          "gh.dependabot_api.update_job.id": job&.id,
+          "gh.dependabot_api.update_config.package_manager": job&.package_manager,
+          "gh.repo.is_private": job&.repo_private?
+        }.compact),
+        extra: extra.merge({
+          dependency_name: dependency&.name,
+          dependency_group: dependency_group&.name
+        }.compact),
+        user: {
+          id: job&.repo_owner
```
JamieMagee, line 116: `capture_exception` now expects `kwargs` instead of a `Hash`. See https://github.com/getsentry/sentry-ruby/blob/7666c94f1f2671e4c42cd8c4a6991e3ce7da56f9/sentry-ruby/lib/sentry-ruby.rb#L394-L401

sorbet/rbi/shims/sentry-ruby.rbi
```diff
@@ -0,0 +1,55 @@
+# typed: strong
+# frozen_string_literal: true
+
+module Sentry
```
JamieMagee, line 4: `spoom gem sentry-ruby` generates an empty file, so this is minimal handwritten types for our uses.

updater/lib/dependabot/setup.rb
```diff
@@ -39,7 +41,8 @@
     swift
   )}x
 
-  config.processors += [ExceptionSanitizer]
+  config.before_send = ->(event, hint) { Dependabot::Sentry::Processor.process_chain(event, hint) }
```
JamieMagee, line 44: `config.processors` has been removed in favour of a lambda `before_send`. I've replicated the behaviour by using a reducer to call processors in order.

I had to migrate `raven_context` to a custom processor as well, because `raven_context` is no longer attached to events by default.

deivid-rodriguez (approved):
Fantastic job! ❤️

jeffwidman:
Thanks @JamieMagee for tackling this, one less hanging TODO that I feel bad about not having the time to finish 😊 plus I was always annoyed that Dependabot was relying on an EOL'd library.

Reading through this PR was a lot of fun 🎉  because it answered several of my questions from https://github.com/dependabot/dependabot-core/pull/7553.

In case you didn't see this note from my PR:
> Events now default to sending asynchronously using a thread pool whereas raven used to send them synchronously.

I suspect this change is harmless, but still, might be worth double-checking when this gets deployed, just in case. IIRC, the async behavior isn't actually something that really matters in the Dependabot case since these events fire as part of a job queue, as opposed to a webpage where we need to get back to the human as fast as possible.

More details [here](https://docs.sentry.io/platforms/ruby/migration/#added), including how to preserve the old behavior through configuration if necessary.

PR #8819, opened 2024-01-17
Group Python helper updates

abdulapopoola:
Group Python helper updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -132,6 +132,10 @@ updates:
       interval: "weekly"
       day: "sunday"
       time: "16:00"
+    groups:
+      all-python-helpers:
+        patterns:
+          - ["*"]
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Normally package managers (pipenv, pip-tools, poetry) require specific work when upgraded. I don't think they should block each other. Also upgrading pip requires work sometimes too, and we've had specific regressions due to the cython version. I would suggest to start with the following groups:

```yaml
common:
  patterns:
   - pip
   - cython
   - flake8

poetry:
  patterns:
    - poetry
    - hashin

pip-tools:
  patterns:
    - pip-tools

pipenv:
  patterns:
    - pipenv
    - pipfile
```

abdulapopoola:
Thanks! Updated

deivid-rodriguez (approved):

PR #8827, opened 2024-01-17
Handle Update PackageReferences more correctly

ryanbrandenburg:
The existing code includes all packageReferences the TempProject as `<PackageReference Include`, regardless of how they were found in the original context. If (for a given project) a PackageReference exists only as `<PackageReference Update` MSBuild does not load that reference and it should not appear in the Dependencies list. The way I saw this manifesting was that packages which don't get a `<PackageReference Include` in the csproj were getting manually included due to `PR Update`s, which caused `NU1605` errors (package downgrade). I assume it could have other downstream effects that don't occur to me ATM as well.

My fix is that we start keeping track of if we have only seen `Update` versions of particular `PackageReference` items, and if so we should add them to our temp project only as `Update` references.

### DIFF
--- a/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Dependency.cs
+++ b/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Dependency.cs
@@ -1,3 +1,3 @@
 namespace NuGetUpdater.Core;
 
-public sealed record Dependency(string Name, string? Version, DependencyType Type, bool IsDevDependency = false, bool IsOverride = false, bool UpdateOnly = false);
+public sealed record Dependency(string Name, string? Version, DependencyType Type, bool IsDevDependency = false, bool IsOverride = false, bool IsUpdate = false);
--- a/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Dependency.cs
+++ b/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Dependency.cs
@@ -1,3 +1,3 @@
 namespace NuGetUpdater.Core;
 
-public sealed record Dependency(string Name, string? Version, DependencyType Type, bool IsDevDependency = false, bool IsOverride = false);
+public sealed record Dependency(string Name, string? Version, DependencyType Type, bool IsDevDependency = false, bool IsOverride = false, bool UpdateOnly = false);
--- a/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
+++ b/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
@@ -136,7 +136,7 @@ internal static partial class MSBuildHelper
 
     public static IEnumerable<Dependency> GetTopLevelPackageDependenyInfos(ImmutableArray<ProjectBuildFile> buildFiles)
     {
-        Dictionary<string, string> packageInfo = new(StringComparer.OrdinalIgnoreCase);
+        Dictionary<string, (string, bool)> packageInfo = new(StringComparer.OrdinalIgnoreCase);
         Dictionary<string, string> packageVersionInfo = new(StringComparer.OrdinalIgnoreCase);
         Dictionary<string, string> propertyInfo = new(StringComparer.OrdinalIgnoreCase);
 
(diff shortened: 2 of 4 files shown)


nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Dependency.cs
```diff
@@ -1,3 +1,3 @@
 namespace NuGetUpdater.Core;
 
-public sealed record Dependency(string Name, string? Version, DependencyType Type, bool IsDevDependency = false, bool IsOverride = false);
+public sealed record Dependency(string Name, string? Version, DependencyType Type, bool IsDevDependency = false, bool IsOverride = false, bool UpdateOnly = false);
```
JoeRobich, line 3: Suggestion, rename to be consistent with other flags that are of the form 'IsX'.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
```diff
@@ -154,7 +154,20 @@ public static IEnumerable<Dependency> GetTopLevelPackageDependenyInfos(Immutable
                 {
                     if (!string.IsNullOrWhiteSpace(attributeValue))
                     {
-                        packageInfo[attributeValue] = versionSpecification;
+                        if (packageInfo.TryGetValue(attributeValue, out var existingInfo))
+                        {
+                            var existingVersion = existingInfo.Item1;
+                            var existingUpdate = existingInfo.Item2;
+                            var vSpec = string.IsNullOrEmpty(versionSpecification) ? existingVersion : versionSpecification;
```
JoeRobich, line 161: If the package reference is an update, should its version be the one that ends up in the packageInfo?

JoeRobich (commented):
Good catch with this fix.

brettfo (approved):

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
```diff
@@ -154,7 +154,20 @@ public static IEnumerable<Dependency> GetTopLevelPackageDependenyInfos(Immutable
...
+                            var existingUpdate = existingInfo.Item2;
+                            var vSpec = string.IsNullOrEmpty(versionSpecification) || existingUpdate ? existingVersion : versionSpecification;
+
+                            var isUpdate = existingUpdate && string.IsNullOrEmpty(packageItem.Include);
+                            packageInfo[attributeValue] = (vSpec, isUpdate);
+                        }
+                        else
+                        {
+                            var isUpdate = string.IsNullOrEmpty(packageItem.Update);
+                            packageInfo[attributeValue] = (versionSpecification, !string.IsNullOrEmpty(packageItem.Update));
```
JoeRobich, line 169:
```suggestion
                            var isUpdate = !string.IsNullOrEmpty(packageItem.Update);
                            packageInfo[attributeValue] = (versionSpecification, isUpdate);
```

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
```diff
@@ -154,7 +154,20 @@ public static IEnumerable<Dependency> GetTopLevelPackageDependenyInfos(Immutable
                 {
                     if (!string.IsNullOrWhiteSpace(attributeValue))
                     {
-                        packageInfo[attributeValue] = versionSpecification;
+                        if (packageInfo.TryGetValue(attributeValue, out var existingInfo))
+                        {
+                            var existingVersion = existingInfo.Item1;
+                            var existingUpdate = existingInfo.Item2;
+                            var vSpec = string.IsNullOrEmpty(versionSpecification) || existingUpdate ? existingVersion : versionSpecification;
```
JoeRobich, line 161: Maybe a comment along these lines to explain the reasoning for choosing one version over the other.

```suggestion
                            // Retain the version from the Update reference since the intention
                            // would be to override the version of the Include reference.
                            var vSpec = string.IsNullOrEmpty(versionSpecification) || existingUpdate ? existingVersion : versionSpecification;
```

JoeRobich (approved):

jakecoffman (approved):

PR #8835, opened 2024-01-18
Always use .ruby-version for Bundler dependency resolution

etiennebarrie:
Currently dependabot-bundler bundles (resolves dependencies) in the context of the Ruby version that Dependabot runs by default, which is [3.1.4](https://github.com/dependabot/dependabot-core/blob/54ab93eb30e4ef732dace694df0a04674b5cafa0/Dockerfile.updater-core#L55).

It's problematic when some gems have different platforms available for a given version. For example, the latest version of `google-protobuf` has a `x86_64-linux` version available that's only compatible with Ruby versions before 3.3, [see it here](https://rubygems.org/gems/google-protobuf/versions/3.25.2-x86_64-linux). It means that bundling in the context of 3.1.4 will bring this version (since it's compatible with Ruby 3.1.4), but once on CI or on developers machines, that version is not compatible, and the bundle is resolved differently.

It's possible to have Dependabot use another Ruby version by simply using [`ruby` from the Gemfile DSL](https://bundler.io/v2.5/man/gemfile.5.html#RUBY). This works thanks to [the monkey patch in `bundler/helpers/v2/monkey_patches/definition_ruby_version_patch.rb`](https://github.com/dependabot/dependabot-core/blob/54ab93eb30e4ef732dace694df0a04674b5cafa0/bundler/helpers/v2/monkey_patches/definition_ruby_version_patch.rb#L24), which changes the `metadata_dependencies` to include the version of Ruby used by the Gemfile.

It's also possible to use the version from `.ruby-version` by using `ruby file: ".ruby-version"`. Dependabot detects that [the file is mentioned in the Gemfile](https://github.com/dependabot/dependabot-core/blob/54ab93eb30e4ef732dace694df0a04674b5cafa0/bundler/lib/dependabot/bundler/file_fetcher.rb#L101), downloads the file, and then lets Bundler do its work with the help of the monkey-patch.

This PR aims to enable the use of `.ruby-version` even when it's not declared in the Gemfile. The `.ruby-version` file is found in the same directory where the Gemfile is.

This may cause Dependabot to resolve gem dependencies differently which may be lead to compatibility issues, but on the other hand it will solve problems we have today where Dependabot leads Bundler to resolve incorrectly.


deivid-rodriguez (approved):

etiennebarrie:
I've rebased, is there anything blocking I can help with?

deivid-rodriguez:
From my side nothing missing! In case any maintainer has time to look into this, this has my blessing :)

abdulapopoola:
Thanks, I'll try to get this merged since it has @deivid-rodriguez 's blessings :); thanks for all the contributions!

abdulapopoola (approved):

PR #8838, opened 2024-01-18
Fix reviewers, assignees types in PullReqeustCreator

andrcuns:
This is a follow up to https://github.com/dependabot/dependabot-core/pull/8793

I'm a bit new to sorbet and completely missed that the type is also defined on initializer params.

Also updated `assignees` attribute because in case of GitLab it is array of integers not strings.


andrcuns:
cc: @JamieMagee since you recently reviewed the previous pr

JamieMagee (approved):

JamieMagee:
@andrcuns Thanks for this. Approved!

Everyone on the team is a bit new to Sorbet 😅

jakecoffman (approved):

PR #8840, opened 2024-01-18
Handle STDOUT more correctly

ryanbrandenburg:
There are cases where we are failing to look at the entire result of a ProcessExtensions call because we return our result as soon as `process.Exited` is Invoked. But `process.Exited` does not indicate that all the STDOUT and STDERR events have been handled, only that the process has entered the Exited state. So I took a look at [what they do in dotnet/format](https://github.com/dotnet/format/blob/main/src/Utilities/ProcessRunner.cs) and made some changes. Of particular note here is the call to `WaitForExit` which guarantees that all events currently in-flight returns before returning.

This could express itself in a lot of ways, but what I observed was that not all transitive dependencies of a project were being correctly reported (we shell that work out to MSBuild). I was unable to write a unit-test which replicated that behavior (unsurprising since it's technically a timing issue), but did find that my changes fixed local runs which were exhibiting this behavior. If anyone can come up with a Unit Test of `ProcessExtensions.RunAsync` which could reliably test this I'm all ear, otherwise I think we'll have to rely on the practical test I did and the fact that we're cribbing off of dotnet/format and roslyn.

Huge thanks to @JoeRobich for pairing with me on this one.

### DIFF
--- a/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/ProcessExtensions.cs
+++ b/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/ProcessExtensions.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Diagnostics;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace NuGetUpdater.Core;
@@ -11,6 +12,7 @@ public static class ProcessEx
     {
         var tcs = new TaskCompletionSource<(int, string, string)>();
 
+        var redirectInitiated = new ManualResetEventSlim();
         var process = new Process
         {
             StartInfo =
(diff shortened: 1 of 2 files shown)


abdulapopoola (approved):

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/ProcessExtensions.cs
```diff
@@ -37,8 +39,20 @@ public static Task<(int ExitCode, string Output, string Error)> RunAsync(string
 
         process.Exited += (sender, args) =>
         {
-            tcs.TrySetResult((process.ExitCode, stdout.ToString(), stderr.ToString()));
-            process.Dispose();
+            // We must call WaitForExit to make sure we've received all OutputDataReceived/ErrorDataReceived calls
```
JoeRobich, line 42: Expand comment to give context about why using the reset event is necessary.

```suggestion
            // It is necessary to wait until we have invoked 'BeginXReadLine' for our redirected IO. Then,
            // we must call WaitForExit to make sure we've received all OutputDataReceived/ErrorDataReceived calls
```

JoeRobich (commented):
This was a good discovery.

JoeRobich (approved):

PR #8841, opened 2024-01-18
Group prod dependencies for Ruby

abdulapopoola:


.github/dependabot.yml
```diff
@@ -15,7 +15,11 @@ updates:
         update-types:
           - "minor"
           - "patch"
-
+      prod-dependencies:
+        dependency-type: "production"
```
abdulapopoola, line 19: @deivid-rodriguez any recommendations for improving our grouping of Ruby PRs? I'd love to have as many things grouped together as possible.

deivid-rodriguez:
We don't usually get many regressions when updating updater dependencies, so this may be fine and allow us to stay more up to date.

Unfortunately since I introduced the root Gemfile & Gemfile.lock, Dependabot PRs for `updater/` require manual amendment because both `Gemfile` and `updater/Gemfile` use dependencies in `dependabot-common`. However, those get currently bumped in `updater/Gemfile.lock`, but not in the root `Gemfile.lock`, and CI complains about that. So one needs to run `bundle lock` and amend Dependabot PRs with the updates that creates.

This is a limitation of Dependabot that will be fixed once multi-directory version-updates are supported.

deivid-rodriguez (approved):
Just to clarify, I'm in favor of trying this! Just wanted to explain in my other comment that updater PRs currently require a bit of extra work and are not mergeable as is.

yeikel:
> Just to clarify, I'm in favor of trying this! Just wanted to explain in my other comment that updater PRs currently require a bit of extra work and are not mergeable as is.

Can this be automated with a dedicated workflow?

jeffwidman:
> This is a limitation of Dependabot that will be fixed once multi-directory version-updates are supported.

Is this no longer a blocker now that we've shipped https://github.blog/changelog/2024-06-25-simplified-dependabot-yml-configuration-with-multi-directory-key-directories-and-wildcard-glob-support/ ?

deivid-rodriguez:
Yep, I think it should be better now with multi-directory updates 👍.

PR #8842, opened 2024-01-18
use common helper to look for `Directory.Packages.props`

brettfo:
In PR #8722 I added a helper to properly locate directory-relative files when the updater is started in a subdirectory, but I forgot to use it to find `Directory.Packages.props`; this simply uses the new helper and deletes the old function.

The other directory-relative files (`Directory.Build.props` and `Directory.Build.targets`) already have tests to ensure they can be found when starting from a subdirectory (see line 399 in the updated `file_fetcher_spec.rb`.)


nuget/lib/dependabot/nuget/file_fetcher.rb
```diff
@@ -191,28 +193,6 @@ def proj_files
...
-        full_project_dir.ascend.each do |base|
-          break if found_directory_packages_props_file
-
-          candidate_file_path = Pathname.new(base).join("Directory.Packages.props").cleanpath.to_path
-          candidate_directory = Pathname.new(File.dirname(candidate_file_path))
-          relative_candidate_directory = candidate_directory.relative_path_from(directory_path)
-          candidate_file = repo_contents(dir: relative_candidate_directory).find do |f|
-            f.name.casecmp?("Directory.Packages.props")
-          end
-          found_directory_packages_props_file = fetch_file_from_host(candidate_file.name) if candidate_file
```
JoeRobich, line 210: I see now, when fetching the file this code didn't join the relative path to the filename.

JoeRobich (approved):

abdulapopoola (approved):

PR #8849, opened 2024-01-19
Centralize more Nuget calls

ryanbrandenburg:
Trying to split up my earlier PR into more manageable chunks, easy parts first.


JoeRobich (approved):
Love this consolidation.

brettfo (approved):

Gemfile
```diff
@@ -28,12 +28,17 @@ gem "tapioca", "0.11.14", require: false, group: :development
...
   gpgme
   rake
+  rspec-its
+  rspec-sorbet
   rubocop
   rubocop-performance
   rubocop-sorbet
   stackprof
+  turbo_tests
+  vcr
```
jurre, line 41: It's not quite clear to me why these changes are needed, I would expect these to already be pulled in via the gemspec?

JamieMagee: They are unfortunately not. Running `bundle exec turbo_tests` in any directory currently gives

```
bundler: failed to load command: turbo_tests (/home/jamie/.asdf/installs/ruby/3.1.4/bin/turbo_tests)
/home/jamie/.asdf/installs/ruby/3.1.4/lib/ruby/gems/3.1.0/gems/bundler-2.5.3/lib/bundler/rubygems_integration.rb:265:in `block in replace_bin_path': can't find executable turbo_tests for gem turbo_tests. turbo_tests is not currently included in the bundle, perhaps you meant to add it to your Gemfile? (Gem::Exception)
```

I think it might actually be easier to get rid of `deps_shared_with_common` and just load the same development dependencies below

```diff
Dir.chdir(File.dirname(common_gemspec)) do
  Gem::Specification.load(common_gemspec).development_dependencies.each do |dep|
-     next unless deps_shared_with_common.include?(dep.name)
-
    gem dep.name, *dep.requirement.as_list
  end
end
```

jurre (approved):

PR #8859, opened 2024-01-22
Bump to Bundler 2.5.5

deivid-rodriguez:


JamieMagee (approved):
The failures are related to rate-limiting. I've noticed it happens when we queue too many PR builds.

deivid-rodriguez:
This should be ready now. Merge order would be:

* First manually regenerate bundler tests cache.
* Then restart CI in https://github.com/dependabot/smoke-tests/pull/173 and see it become green.
* Merge https://github.com/dependabot/smoke-tests/pull/173.
* Then restart CI here and see it become green.
* Then merge this PR.

The reason we need some Bundler tests regenerated is that Bundler 2.5.5 brings https://github.com/rubygems/rubygems/pull/7352 which due to the Etag format change will make Bundler's smoke tests cache not be hit one more time.

deivid-rodriguez:
The upgrade 2.5.5 is a bit complicated but following the steps detailed in my previous message should do the trick. Let me know if any help from my side is needed, and feel free to ping me whenever Bundler upgrades cause issues!

jakecoffman (approved):

PR #8863, opened 2024-01-22
Disable analyzers in MSBuild temp projects

ryanbrandenburg:
We want to disable analyzers from running when we run coherency checks and other MSBuild tasks to minimize the chances for things to go wrong. I saw an instance of analyzers running and causing issues in the wild, and though this wouldn't have fixed that issue (Analyzer in question doesn't respect this property) it will at minimum keep us from wasting cycles on analyzers we don't care about.


brettfo (approved):

abdulapopoola (approved):

ryanbrandenburg:
Failure looks like flaky docker stuff, re-running.

PR #8871, opened 2024-01-23
add a fake ecosystem for updater integration tests

jakecoffman:
This PR creates a fake ecosystem to make integration testing the updater simpler, and also introduces a new way to write integration tests.

The motivation here was grouped updates and especially multi-directory updates is going to require some fundamental refactoring, and our current tests use the existing classes directly. By writing tests at a higher level, we ensure that our changes aren't breaking anything.

## The Silent Ecosystem

One of the challenges of our smoke tests is managing the proxy cache so tests aren't constantly failing. A challenge of rspec tests is maintaining the VCR cassettes. A challenge of both of these approaches is we can't easily tell what versions were available at the time the test ran without examining the cache or cassettes. 

To solve this problem, I created the Silent Ecosystem which doesn't make any network requests! In order to find available versions it looks in the cloned repo for a file matching the name of the dependency. That file contains a JSON record with the versions, and potentially anything else required to make an interesting test.

An additional benefit to having a minimal ecosystem, just enough to test the updater code, is it shows us what is required for each ecosystem to implement. For instance, I noticed that for the UpdateChecker, we require the ecosystem to filter ignore rules, but we already have that information in the Updater. It seems we could move that into the Updater and then make ecosystem implementations simpler across the board.

## Script-based Integration Tests

I was encouraged by the results of using script-based testing in the Dependabot CLI and so I tried to apply it here. It seems useful so I thought we could give it a go.

One of the challenges of integration testing is we can't easily see all of the files involved on one screen. In the rspec, we have to look in fixture files for the manifests, another fixture file for the YAML potentially, and then possibly the VCR Cassettes for the actual responses from the registry.

With the script-based approach in combination with Silent's design, we get it all in one place:

```
dependabot update -f input.yml --local . --updater-image ghcr.io/dependabot/dependabot-updater-silent
stderr 'created \| dependency-a \( from 1.2.3 to 1.2.4 \)'
pr-created expected.json

-- manifest.json --
{
  "dependency-a": { "version": "1.2.3" },
  "dependency-b": { "version": "2.3.4" }
}

-- expected.json --
{
  "dependency-a": { "version": "1.2.4" },
  "dependency-b": { "version": "2.3.4" }
}

-- dependency-a --
{
  "versions": [
    "1.2.3",
    "1.2.4",
    "1.2.5"
  ]
}

-- dependency-b --
{
  "versions": [
    "2.3.4",
    "2.3.5"
  ]
}

-- input.yml --
job:
  package-manager: "silent"
  dependencies:
    - dependency-a
  source:
    directory: "/"
    provider: example
    hostname: example.com
    api-endpoint: https://example.com/api/v3
    repo: dependabot/smoke-tests
  security-advisories:
    - dependency-name: dependency-a
      affected-versions:
        - < 1.2.4
      patched-versions: []
      unaffected-versions: []
  security-updates-only: true

```

## TODOs

To write the tests, I leveraged the Dependabot CLI for it's input and output capabilities. However, this is overkill for these tests as the Silent ecosystem makes no network calls and is written entirely in Ruby. So the next step will be to execute the Ruby directory, but to do that will take a bit more work and decisions if that needs to go into dry-run or whip it up in the Go code.

## Naming

Naming is hard, if someone has another name for this ecosystem I am all ears. I tried not to call it a common word like Dummy or Integration as those are already well used in the codebase.


Nishnha:
Any suggestions on how to review this PR since it has so many line changes?

silent/lib/dependabot/silent.rb
```diff
@@ -0,0 +1,23 @@
...
+require "dependabot/silent/requirement"
+require "dependabot/silent/version"
+
+require "dependabot/pull_request_creator/labeler"
+Dependabot::PullRequestCreator::Labeler
+  .register_label_details("silent", name: "silent_package_manager", colour: "000000")
+
+require "dependabot/dependency"
+Dependabot::Dependency
+  .register_production_check("silent", ->(groups) { groups.empty? || groups.include?("prod") })
```
Nishnha, line 20: I don't think we would want to add this to list of production ecosystems?

Nishnha (approved):
I like this idea a lot and it definitely improves the readability of some of these tests! I think the name silent makes sense. I would appreciate some docs on how to write these integration tests or just explaining what the parts of the test files are.

I'm slightly concerned about removing existing rspec tests because I don't fully understand what's covered and what's not by these integration tests. I would feel more okay with the change if we introduced the silent ecosystem and removed the rspec tests as 2 separate PRs.

silent/lib/dependabot/silent.rb
jakecoffman: I agree, luckily that's not what this does. This is what makes `Dependency::production?` work, so we know whether this is a dev vs production dependency.

https://github.com/dependabot/dependabot-core/blob/9130f77d409395cbce16001a7597718467cfba4b/common/lib/dependabot/dependency.rb#L168-L177

PR #8874, opened 2024-01-23
Revert "Migrate from `sentry-raven` to `sentry-ruby`"

jakecoffman:
Reverts dependabot/dependabot-core#8818

We stopped receiving errors in Sentry when this was deployed, so reverting it and we can try to figure it out.


deivid-rodriguez (approved):

jakecoffman:
@JamieMagee you can put this PR back up again and we'll give it another go when we figure out what went wrong.

PR #8879, opened 2024-01-24
try to perform environment variable expansion in `NuGet.Confing`

### LINKED ISSUE #8876: NuGet update checker doesn't honor environment variables in `NuGet.Config`
When a `NuGet.Config` file is parsed, the contents of the file can contain environment variables as referenced [here](https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file#using-environment-variables) in the spec.

The [current implementation](https://github.com/dependabot/dependabot-core/blob/e693d3c0162e4cda6e6baba8330bf2e14bc7038b/nuget/lib/dependabot/nuget/update_checker/repository_finder.rb#L290) doesn't do any variable expansion, but should.

Any variables that couldn't be expanded should be reported, as that means package updates might not happen.  E.g., consider this example `NuGet.Config`:

``` xml
<configuration>
  <packageSources>
    <add key="SomePackageSource" value="https://nuget.example.com/index.json" />
  </packageSources>
  <packageSourceCredentials>
    <SomePackageSource>
      <add key="Username" value="user" />
      <add key="ClearTextPassword" value="%SECRET_INJECTED_BY_CI%" />
    </SomePackageSource>
  </packageSourceCredentials>
</configuration>
```

In this case, the environment variable `SECRET_INJECTED_BY_CI` is not present in the updater, so that package source cannot be queried.

brettfo:
Expand environment variables in `NuGet.Config` as per [the documentation](https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file#using-environment-variables).

There are several use cases around this, some which will work in the updater, others which will not.  One case where this won't help the updater, but is still interesting is the following:

``` xml
...
<add key="ClearTextPassword" value="%TOKEN_FROM_CI%" />
...
```

In this case the public dependabot updater won't have access to that token/environment variable, but if a user runs dependabot in their own isolated instance, they can inject that variable and it _will_ be replaced.

In the case where an environment variable couldn't be expanded, a warning is issued to help make it obvious to the user why their update may have failed.

To make this directly testable, the `known_repositories` field had to be exposed on the repository finder.

Fixes #8876.

### DIFF
--- a/nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
+++ b/nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
@@ -220,6 +220,7 @@ module Dependabot
             else
               key = node.attribute("key")&.value&.strip || node.at_xpath("./key")&.content&.strip
               url = node.attribute("value")&.value&.strip || node.at_xpath("./value")&.content&.strip
+              url = expand_windows_style_environment_variables(url) if url
               sources << { url: url, key: key }
             end
           end
--- a/nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
+++ b/nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
@@ -314,8 +314,8 @@ module Dependabot
           # NuGet.Config files can have Windows-style environment variables that need to be replaced
           # https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file#using-environment-variables
           string.gsub(/%([^%]+)%/) do
-            environment_variable_name = ::Regexp.last_match(1)
-            environment_variable_value = ENV.fetch(T.must(environment_variable_name), nil)
+            environment_variable_name = T.must(::Regexp.last_match(1))
+            environment_variable_value = ENV.fetch(environment_variable_name, nil)
             if environment_variable_value
               environment_variable_value
             else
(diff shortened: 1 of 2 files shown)


edgarrs (approved):

nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
```diff
@@ -307,6 +310,24 @@ def add_config_file_credentials(sources:, doc:)
         end
         # rubocop:enable Metrics/PerceivedComplexity
 
+        def expand_windows_style_environment_variables(string)
+          # NuGet.Config files can have Windows-style environment variables that need to be replaced
+          # https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file#using-environment-variables
+          string.gsub(/%([^%]+)%/) do
+            environment_variable_name = ::Regexp.last_match(1)
+            environment_variable_value = ENV.fetch(T.must(environment_variable_name), nil)
```
JamieMagee, line 318: Move `T.must()` up to the caller, otherwise the type is `T.nilable(String)` everywhere else

```suggestion
            environment_variable_name = T.must(::Regexp.last_match(1))
            environment_variable_value = ENV.fetch(environment_variable_name, nil)
```

JamieMagee (approved):

PR #8878, opened 2024-01-23
Migrate from `sentry-raven` to `sentry-ruby`

JamieMagee:
Reverts #8874 (Cherry-picked 85251aa60d7e72ae1aee8b1647a231bb5fdcdd4e from #8818)

~NOTE: Do not merge this yet.~

Verified in staging. Ready to merge.

PR #8885, opened 2024-01-24
Report release to sentry

deivid-rodriguez:
To better catch regressions.


deivid-rodriguez:
Sounds like the updater SHA is not getting passed correctly, needs some debugging.

JamieMagee:
Can I merge #8878 first please?

deivid-rodriguez:
Of course, I actually tried this in staging and saw "development" being reported as the release, which is unexpected! So it's currently marked as WIP and I'll hold working on it at least until your migration is done 👍.

JamieMagee:
Thanks! You are good to go.

abdulapopoola (approved):

JamieMagee (approved):

Nishnha:
It looks like it's using the default `development` value for the release on new sentry issues 

Screenshot 2024-02-22 at 12 57 00 PM

I think it's because the prepare tag and prepare tag (fork) steps are both getting skipped and that's where we add the value to the GITHUB_ENV. We are missing the case where the workflow was triggered by a pull request review

Screenshot 2024-02-22 at 12 55 33 PM

deivid-rodriguez:
Yep @Nishnha, I observed the same thing and marked this PR as WIP because of that. Awesome that you managed to fix the root cause :muscle: :heart:

PR #8890, opened 2024-01-24
Fix milestone type for PullRequestCreator

andrcuns:
Should be the last fix regarding types of PullRequestCreator and GitLab implementation.

This updates `milestone` type which in case of GitLab is just a single integer.

I did ran gem built from this branch through my workflow to make sure some other thing isn't missing again and I think it should be it now.


andrcuns:
cc: @JamieMagee

JamieMagee (approved):

Nishnha (approved):

PR #8901, opened 2024-01-25
search all candidate packages for compatibility in descending version order

brettfo:
The previous implementation did a binary search of available packages and it was possible for it to return nothing which would then later cause this:

```
undefined method `fetch' for nil:NilClass

                                                              ^^^^^^
```

The reason a binary search is not appropriate is because the list of package versions isn't sorted in a way that binary search can handle, because there can be any number of incompatible packages in the list.  The test mimics a real-world example where there is a project with the `netcoreapp3.1` TFM with a reference to `nunit, 3.14.0`.  Due to the order of packages to test, the binary search algorithm checked against package version 3.4.0 which only supports the earlier .NET Framework so when that compatibility check returned `false`, the binary search kept going lower and never found a match, eventually resulting in the `nil` value causing the error above.

The binary search approach would work if we first filtered the list to only contain packages that are compatible with the TFM, but doing all of those checks would be really expensive, so the cheaper approach is to check each package in descending order until we find one that is compatible.  In the new test case, this results in the same version that's already installed, resulting in no update.


ryanbrandenburg (approved):

JamieMagee (approved):

abdulapopoola (approved):

edgarrs (approved):

brettfo:
The smoke test failure seems to be a different ordering of the dependency files.  The old order was `.csproj, .props, .targets` and the new is `.csproj, .targets, .props`.  I have no idea how this change could have caused that, but I'll start investigating.

nuget/lib/dependabot/nuget/update_checker/version_finder.rb
```diff
@@ -83,7 +83,9 @@ def find_compatible_version(sorted_versions)
           return first_version unless version_compatible?(dependency.version)
           return first_version if version_compatible?(first_version.fetch(:version))
```
JoeRobich, line 84: This line can be removed as it will now be the run twice as part of the `.find(`

brettfo: Good catch; fixed.

JoeRobich (approved):

PR #8912, opened 2024-01-26
clean directory at job start

jakecoffman:
I noticed the new devcontainers smoke test has a directory that does not start with `/`. https://github.com/dependabot/smoke-tests/blob/5b012ee504a3ec30eb05af415d600aeff5cc00fd/tests/smoke-devcontainers.yaml#L16

It just happened to work but is now breaking as I [implement multi-dir support](https://github.com/dependabot/dependabot-core/pull/8742). 

I think we can be a little more flexible here and just add the `/` and also clean the path. 

In the future I'd like to explore changing the source directory type to a `Pathname` instead of a `String` to encourage comparisons to be for the cleaned path.

### DIFF
--- a/updater/lib/dependabot/job.rb
+++ b/updater/lib/dependabot/job.rb
@@ -332,6 +332,11 @@ module Dependabot
     end
 
     def build_source(source_details)
+      # Immediately normalize the source directory, ensure it starts with a "/"
+      directory = source_details["directory"]
+      directory = Pathname.new(directory).cleanpath.to_s
+      source_details["directory"] = "/#{directory}" unless directory.start_with?("/")
+
       Dependabot::Source.new(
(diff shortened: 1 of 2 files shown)


abdulapopoola (approved):

deivid-rodriguez:
I think I created it following https://github.com/dependabot/smoke-tests/blob/5b012ee504a3ec30eb05af415d600aeff5cc00fd/README.md#how-to-add-new-tests, so we may need to fix those docs too!

Way easier to add new smoke tests thanks to that by the way :heart:.

jakecoffman:
@deivid-rodriguez no problem, I'm fixing it here rather than in the smoke test or CLI because it would make a better experience to not worry about it anymore!

PR #8913, opened 2024-01-26
Handle MSBuild property conditions that have a property wrapped in single quotes in NuGetUpdater

bording:
It is very common for conditions in MSBuild to have the property values wrapped in single quotes. This PR updates `MSBuildHelper` to understand how to deal with these.

There were tests around the condition handling, so I added variations of the test data that includes quotes.


brettfo (approved):

abdulapopoola:
Thank you so much @bording and @DavidBoike for this!

PR #8926, opened 2024-01-29
Retry transient git clone errors

JamieMagee:
This change allows for Dependabot to retry transient git errors with an exponential backoff. The errors below are either ones we've seen from telemetry, or possible errors in git source code.

The current backoff strategy is configured to retry 5 times, with sleep times of 3, 6, 12, 24, and 48 seconds.

### DIFF
--- a/common/lib/dependabot/file_fetchers/base.rb
+++ b/common/lib/dependabot/file_fetchers/base.rb
@@ -809,7 +809,11 @@ module Dependabot
             if GIT_RETRYABLE_ERRORS.any? { |error| error.match?(e.message) } && retries < 5
               retries += 1
               # 3, 6, 12, 24, 48, ...
-              sleep((2 ^ (retries - 1)) * 3)
+              sleep_seconds = (2 ^ (retries - 1)) * 3
+              Dependabot.logger.warn(
+                "Failed to clone repo #{source.url} due to #{e.message}. Retrying in #{sleep_seconds} seconds..."
+              )
+              sleep(sleep_seconds)
               retry
             end
             raise unless e.message.match(GIT_SUBMODULE_ERROR_REGEX) && e.message.downcase.include?("submodule")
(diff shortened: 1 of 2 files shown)


abdulapopoola (approved):

ryanbrandenburg (approved):

jurre (approved):

jakecoffman (approved):

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -51,6 +51,24 @@ class Base
       GIT_SUBMODULE_CLONE_ERROR =
         /^fatal: clone of '(?<url>.*)' into submodule path '.*' failed$/
       GIT_SUBMODULE_ERROR_REGEX = /(#{GIT_SUBMODULE_INACCESSIBLE_ERROR})|(#{GIT_SUBMODULE_CLONE_ERROR})/
+      GIT_RETRYABLE_ERRORS =
+        T.let(
+          [
+            /remote error: Internal Server Error/,
+            /fatal: Couldn\'t find remote ref/,
```
abdulapopoola, line 58: are these all guaranteed retryable errors? Some seem to be terminal errors and maybe we shouldn't retry on them. 

On the flip side, maybe a max 3 min retry interval is not so expensive?

JamieMagee: I agree, it's hard to say whether or not all of these are retryable without retrying them. If we merge this, and telemetry shows no noticeable difference, we can re-evaluate?

In researching this, I did find that there is significant overlap with a tool the Chromium team uses[^1].

[^1]: https://chromium.googlesource.com/infra/infra/+/HEAD/go/src/infra/tools/git/retry_regexp.go#17

abdulapopoola: Makes sense; agree with you. And thanks for the Chromium link too!

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -786,6 +806,12 @@ def _clone_repo_contents(target_directory:)
 
             @submodules = find_submodules(path) if recurse_submodules_when_cloning?
           rescue SharedHelpers::HelperSubprocessFailed => e
+            if GIT_RETRYABLE_ERRORS.any? { |error| error.match?(e.message) } && retries < 5
+              retries += 1
+              # 3, 6, 12, 24, 48, ...
+              sleep((2 ^ (retries - 1)) * 3)
+              retry
```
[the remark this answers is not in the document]
JamieMagee: Added a warning

abdulapopoola (approved):

PR #8928, opened 2024-01-29
Do not swallow exception, print the message

trejjam:
Debugging of the native helper is extremely painful when you do not provide the context of failure.


abdulapopoola (approved):

nuget/lib/dependabot/nuget/nuget_config_credential_helpers.rb
```diff
@@ -55,17 +55,16 @@ def self.restore_user_nuget_config
...
       end
 
-      # rubocop:disable Lint/SuppressedException
       def self.patch_nuget_config_for_action(credentials, &_block)
         add_credentials_to_nuget_config(credentials)
         begin
           yield
-        rescue StandardError
+        rescue StandardError => e
+          puts "block causes an exception #{e}: #{e.message}"
```
JamieMagee, line 63: Should this use `Dependabot.logger` instead of `puts`?

trejjam: Logger, in general, sounds good. I just did not see it used anywhere (I am monkey writing Ruby I do not know the language or tools) hence I did not use it :)

brettfo:
In production this should never fail, but it might when running locally; is that what you're seeing?

trejjam:
Yes, I had a problem with the configuration of my environment.

The first was that dotnet was not discoverable, and the second was that the native tool was not authorized against a private repository, which is relatively easy to introduce in many places (ruby, dotnet, remote repository, invalid scope of credentials, ...).

brettfo:
The reason for the NuGet.Config mangling is specifically to pass through feed credentials that dependabot knows about, but the updater doesn't.

The most common way for dependabot to be run with a custom feed is via a `dependabot.yml` file in the repo, but if you need to run the tool locally for testing and use a private feed you can do this (I do this from the Docker development shell):

``` bash
LOCAL_CONFIG_VARIABLES='[{"type":"nuget_feed","token":"MY_SECRET_TOKEN","url":"https://nuget.example.com/index.json"}]' # add feeds as necessary
bin/dry-run.rb nuget dependabot/dependabot-core # ... other arguments
```

brettfo (approved):

brettfo:
Thanks for indulging me in questions, but yes, doing something with an exception rather than silently swallowing it can only be a good thing.

trejjam:
@brettfo https://github.com/dependabot/dependabot-core/pull/8928#issuecomment-1920177073 a different PR :)

trejjam:
About "In production, it should never fail." It did because dependabot does not handle well private repositories with native updater (I experience issue with Azure DevOps - it's handled in a different PR)

brettfo:
> @brettfo [#8928 (comment)](https://github.com/dependabot/dependabot-core/pull/8928#issuecomment-1920177073) a different PR :)

I have too many tabs open to keep track.  :)

> About "In production, it should never fail." It did because dependabot does not handle well private repositories with native updater (I experience issue with Azure DevOps - it's handled in a different PR)

Can you share more information about these scenarios and how the `NuGet.Config` is failing to get patched?  I can't think of a production scenario that would pass in one instance, but not in the other, because that file is always in the same place.

trejjam:
It's not about the file not being patched. The try-catch is not guarding `Nuget.config` patch.

It guards only calling native updater (passed block).

PR #8933, opened 2024-01-29
always clone all the ecosystems

jakecoffman:
This enables cloning for all ecosystems. It looks like most of them were cloning already at this point.

There's more code that can be cleaned up if you pull the `job.clone?` thread, but this is good for now.


bin/dry-run.rb
```diff
@@ -474,10 +474,7 @@ def log_conflicting_dependencies(conflicting_dependencies)
   commit: $options[:commit]
 )
 
-always_clone = Dependabot::Utils
-               .always_clone_for_package_manager?($package_manager)
```
abdulapopoola, line 478: wondering if we would ever need an 'opt-out' option (the inverse of our current state). On 2nd thoughts; I'll default to YAGNI. And as @jakecoffman mentioned, we do have Actions as a solution.

abdulapopoola (approved):

JamieMagee (approved):

jakecoffman:
I deleted the few remaining integration tests as they rely on some amount of mocking to work, and should be covered by smoke tests and the new silent tests. 

The github_actions smoke test is failing, but the data is the same just reordered, so I can regenerate it after merging.

So I think this is good to go.

PR #8940, opened 2024-01-30
Don't recursively update projects which have already been evaluated

ryanbrandenburg:
Imagine a directory structure like the following:
```text
dirs.proj
Level1\
    dirs.proj
    Level1a\
        Level1a.csproj
    Level1b\
        Level1b.csproj
```

With the current code it seems that we'll call the NuGetUpdater.CLI for `dirs.proj`, `Level1\dirs.proj`, `Level1\Level1a\Level1a.csproj`, and `Level1\Level1B\Level1b.csproj`, despite the fact that the first dirs.proj contained references to `Level1\dirs.proj`, which contains references to the csproj's, so they've all already been checked for the needed updates. This PR is attempting to make sure that we only make the initial call to dirs.proj (and any project file underneath it which was not included by dirs.proj, accidentally or otherwise).

I'm sending this one a bit early so that people can check my logic (and to use the CI for testing since my local instance was being annoying).


nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
```diff
@@ -14,6 +14,7 @@
 module Dependabot
   module Nuget
     class FileParser
+      # rubocop:disable Metrics/ClassLength
```
ryanbrandenburg, line 17: I'm not planning to leave this as-is, but I didn't want to spend a bunch of time and energy refactoring it smaller until I was confident that my fix works.

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
```diff
@@ -131,25 +169,18 @@ def add_transitive_dependencies(project_file, doc, dependency_set)
...
-          project_file_directory = File.dirname(project_file.name)
-          is_rooted = project_file_directory.start_with?("/")
-          # Root the directory path to avoid expand_path prepending the working directory
-          project_file_directory = "/" + project_file_directory unless is_rooted
-
           # Look for regular project references
-          doc.css(PROJECT_REFERENCE_SELECTOR).each do |reference_node|
+          project_refs = doc.css(PROJECT_REFERENCE_SELECTOR)
+          # Look for ProjectFile references (dirs.proj)
+          project_files = doc.css(PROJECT_FILE_SELECTOR)
```
ryanbrandenburg, line 175: This scenerio doesn't seem to be affecting the repo I was testing against for some reason, but it's possible for the system to ignore dirs.proj and just run the projects under it individually because it didn't pick up any of their dependencies.

nuget/lib/dependabot/nuget/file_updater.rb
```diff
@@ -62,9 +66,12 @@ def try_update_projects(dependency)
 
           next unless project_dependencies.any? { |dep| dep.name.casecmp(dependency.name).zero? }
 
-          NativeHelpers.run_nuget_updater_tool(repo_root: repo_contents_path, proj_path: proj_path,
-                                               dependency: dependency, is_transitive: !dependency.top_level?,
-                                               credentials: credentials)
+          call_nuget_updater_tool(dependency, proj_path) unless checked_files.include?(project_file.name)
```
ryanbrandenburg, line 69: This is the primary fix in this PR. If we've already considered a project_file because it's a transitive dependency of one we already ran then we need to skip it. Otherwise we can be running the same projects over and over.

nuget/spec/dependabot/nuget/file_parser/project_file_parser_spec.rb
```diff
@@ -6,119 +6,7 @@
 require "dependabot/source"
 require "dependabot/nuget/cache_manager"
 require "dependabot/nuget/file_parser/project_file_parser"
-
-module NuGetSearchStubs
```
ryanbrandenburg, line 10: Needed this elsewhere and it was crowding this file anyway.

nuget/spec/dependabot/nuget/file_updater_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+                       "microsoft.extensions.dependencymodel/1.0.0/" \
+                       "microsoft.extensions.dependencymodel.nuspec")
+      .to_return(status: 200, body: fixture("nuspecs", "Microsoft.Extensions.DependencyModel.1.0.0.nuspec"))
+  end
+
+  describe "#updated_dependency_files" do
+    subject(:updated_files) { file_updater_instance.updated_dependency_files }
+
+    context "with a dirs.proj" do
+      it "does not repeatedly update the same project", focus: true do
```
ryanbrandenburg, line 70: I just added a test for the scenario I'm dealing with, but it would be prudent for a future PR(s) to add more tests here. It's a complicated class which is in control of our most performance-critical item (NugetUpdater.Cli is the slowest part of runs for repos large enough to care about their perf).

nuget/spec/fixtures/nuspecs/Microsoft.Extensions.DependencyModel.1.0.0.nuspec
```diff
@@ -0,0 +1,28 @@
...
+    <owners>Microsoft.Extensions.DependencyModel</owners>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <licenseUrl>https://github.com/dotnet/core-setup/blob/master/LICENSE.TXT</licenseUrl>
+    <projectUrl>https://dot.net/</projectUrl>
+    <description>Abstractions for reading `.deps` files.</description>
+    <copyright>© Microsoft Corporation.  All rights reserved.</copyright>
+    <serviceable>true</serviceable>
+    <repository type="git" url="git://github.com/dotnet/core-setup" />
+    <dependencies>
+      <group targetFramework=".NETFramework4.5.1">
```
ryanbrandenburg, line 16: I trimmed dependencies of this package out because they were causing a lot of extra nuget calls which didn't matter to our scenario.

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
```diff
@@ -131,25 +166,18 @@ def add_transitive_dependencies(project_file, doc, dependency_set)
...
 
-            # normalize path separators
-            relative_path = relative_path.tr("\\", "/")
-            # path is relative to the project file directory
-            relative_path = File.join(project_file_directory, relative_path)
-
-            # get absolute path
-            full_path = File.expand_path(relative_path)
-            full_path = full_path[1..-1] unless is_rooted
+            full_path = full_path(project_file, relative_path)
```
JoeRobich, line 180: I would suggest changing either the variable or method name here to make it easier to read.

nuget/lib/dependabot/nuget/file_updater.rb
```diff
@@ -62,9 +66,12 @@ def try_update_projects(dependency)
...
+
+          checked_files.add(project_file.name)
```
JoeRobich, line 71: This key will need to include at least the dependency name being updated. `try_update_projects` will run once for each dependency in `dependencies`. 

It would be worth looking at updating the NuGetUpdater to take in an array of dependency updates to avoid the cost of evaluating the build files multiple times.

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
JoeRobich: Do dirs.proj files make it through this bit of code or do we "evaluate" the dirs.proj to get a list of referenced .csproj files?

nuget/lib/dependabot/nuget/file_updater.rb
ryanbrandenburg: Good point about the dependency, I think it's also technically the version that needs to be checked as well. I've updated the key.

I think you're right, reducing the number of build evaluations is definitely our biggest perf savings available. I filed https://dev.azure.com/mseng/1ES/_workitems/edit/2145689 to follow up on that.

nuget/spec/dependabot/nuget/file_updater_spec.rb
ryanbrandenburg: I created https://dev.azure.com/mseng/1ES/_workitems/edit/2145693 to track follow-up.

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
ryanbrandenburg: dirs.proj files make it through here, and we do call NugetUpdater.CLI directly on dirs.proj files. NugetUpdater.CLI then finds all the csproj's an checks them for the dependency we want to update.

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
ryanbrandenburg: Apparently I ended up keeping it under the limit, though I could swear it's longer now than when rubocop complained.

ryanbrandenburg:
As @JoeRobich and @brettfo pointed out offline, it's not just `dirs.proj`, `tests.proj` is just as valid etc. I don't THINK any of my changes are dirs.proj exclusive, but someone let me know if you spot something to the contrary.

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
```diff
@@ -131,27 +166,20 @@ def add_transitive_dependencies(project_file, doc, dependency_set)
...
-            # path is relative to the project file directory
-            relative_path = File.join(project_file_directory, relative_path)
-
-            # get absolute path
-            full_path = File.expand_path(relative_path)
-            full_path = full_path[1..-1] unless is_rooted
+            full_project_path = full_path(project_file, relative_path)
 
-            referenced_file = dependency_files.find { |f| f.name == full_path }
+            referenced_file = dependency_files.find { |f| f.name == full_project_path }
```
JoeRobich, line 182: @brettfo Has this been subtly wrong for a while? the `dependency_file.name` will be relative to the `dependency_file.directory` not an absolute path, right?

ryanbrandenburg: If that were the case wouldn't this test fail? https://github.com/dependabot/dependabot-core/blob/ec0349e8c0793612b46e60021f816709759b71ff/nuget/spec/dependabot/nuget/file_parser/project_file_parser_spec.rb#L169

I did some local testing and it seem like "full_path" in this case is actually working relative to the repo-root instead of the drive root, which is the same thing that the DepsFiles are doing so this works.

JoeRobich: ah ok. It was `expand_path` which was causing the question. Reading the docs answers it.

ryanbrandenburg:
I would be very surprised if my PR was actually related to that error. I didn't make a change to Nuget.config handling, the error apparently only happened once, and it sounds like the error is "Read-Only File System" which can sometimes indicate file system corruption (linux makes the FS read-only to keep it from doing more damage). My best bet would be a random flaky docker problem.

PR #8803, opened 2024-01-16
handle dependencies incidentally updated

jakecoffman:
Fixes #8540

During a grouped update, if a package manager updates two dependencies at the same time, perhaps because they depend on each other, then Dependabot would produce an individual PR for it.

This PR checks each dependency in the group to see if it has already been updated by a previous dependency. If so it adds it to the "handled" list so Dependabot won't bump it again.

PR #8941, opened 2024-01-30
add tests around incidental updates

jakecoffman:
Documenting the behavior of the updater with tests!

When a dependency depends on a specific version of another dependency they will get bumped together, but the interesting part is what Dependabot does next. 

Currently there's a bug with grouped updates using semver grouping, it will create an extra PR, so I've captured that behavior here which should be fixed by https://github.com/dependabot/dependabot-core/pull/8803.


abdulapopoola:
Really nice to see the new test format being used!!!

jurre (approved):

PR #8943, opened 2024-01-30
add sorbet types to Dependabot::Job

jakecoffman:
I probably got some things wrong, lets see what the tests say.


jakecoffman:
Our Updater specs are passing in types that may not be possible (e.g. Job ID comes from an environment variable so it should be a String but the tests are using an Integer) so I'll have to do a pass on them.

JamieMagee:
> Our Updater specs are passing in types that may not be possible (e.g. Job ID comes from an environment variable so it should be a String but the tests are using an Integer) so I'll have to do a pass on them.

You might be able to do a `T.any(String, Integer)` and add a `TODO` comment as a quick workaround?

updater/lib/dependabot/job.rb
```diff
@@ -43,87 +45,132 @@ class Job
...
-                :token,
-                :vendor_dependencies,
-                :dependency_groups,
-                :dependency_group_to_refresh
+    ).freeze, T::Array[Symbol])
+
+    sig { returns(T::Array[T::Hash[String, T.untyped]]) }
+    attr_reader :allowed_updates
+
+    sig { returns(T::Array[T.untyped]) }
```
JamieMagee, line 53:
```suggestion
    sig { returns(T::Array[T::Hash[String, String]) }
```

I've written this _way_ too many times. I need to write a type alias for it.

updater/lib/dependabot/job.rb
```diff
@@ -43,87 +45,132 @@ class Job
...
+    attr_reader :ignore_conditions
+
+    sig { returns(String) }
+    attr_reader :package_manager
+
+    sig { returns(T.nilable(String)) }
+    attr_reader :requirements_update_strategy
+
+    sig { returns(T::Array[T.untyped]) }
```
JamieMagee, line 77:
```suggestion
    sig { returns(T::Array[Dependabot::SecurityAdvisory]) }
```

updater/lib/dependabot/job.rb
```diff
@@ -43,87 +45,132 @@ class Job
...
+    sig { returns(Dependabot::Source) }
+    attr_reader :source
+
+    sig { returns(T.nilable(String)) }
+    attr_reader :token
 
+    sig { returns(T::Boolean) }
+    attr_reader :vendor_dependencies
+
+    sig { returns(T::Array[T.untyped]) }
```
JamieMagee, line 92:
```suggestion
    sig { returns(T::Array[Dependabot::DependencyGroup]) }
```

updater/lib/dependabot/job.rb
```diff
@@ -43,87 +45,132 @@ class Job
...
+    attr_reader :existing_pull_requests
+
+    sig { returns(T::Array[T.untyped]) }
+    attr_reader :existing_group_pull_requests
+
+    sig { returns(T.any(Integer, String)) }
+    attr_reader :id
+
+    sig { returns(T::Array[T.untyped]) }
```
JamieMagee, line 68:
```suggestion
    sig { returns(T::Array[Dependabot::Config::IgnoreCondition]) }
```

updater/lib/dependabot/job.rb
```diff
@@ -43,87 +45,132 @@ class Job
...
+    end
     def self.new_fetch_job(job_id:, job_definition:, repo_contents_path: nil)
-      attrs = standardise_keys(job_definition["job"]).slice(*PERMITTED_KEYS)
+      attrs = standardise_keys(job_definition["job"]).select { |k, _| PERMITTED_KEYS.include?(k) }
 
       new(attrs.merge(id: job_id, repo_contents_path: repo_contents_path))
     end
 
+    sig do
+      params(job_id: T.untyped, job_definition: T.untyped, repo_contents_path: T.untyped).returns(T.attached_class)
```
JamieMagee, line 108: `attached_class` is overkill here, unless you're planning to split this into a class hierarchy later?

jakecoffman: I don't remember doing that so probably once when I ran `bundle exec srb tc -a` it added it. That command can do some crazy things sometimes!

updater/lib/dependabot/job.rb
jakecoffman: Unfortunately that is not what this is, although I do wish it was.

This class takes the job input which is a big hash of everything, and sets the properties individually as variables and usually doesn't make them into a proper class. So they're still arrays of hashes of strings or nilable strings, etc etc. 😢

updater/lib/dependabot/job.rb
jakecoffman: Same as below, it's not constructed into a class yet.

updater/lib/dependabot/job.rb
jakecoffman: Best I can do is `sig { returns(T::Array[T::Hash[String, T.any(Boolean, String)]) }` because replaces-base exists. Not sure if there are any other cases.

JamieMagee (approved):

updater/lib/dependabot/job.rb
```diff
@@ -306,64 +373,81 @@ def log_ignore_conditions_for(dependency)
...
       )
     end
 
     # Provides a Dependabot::Config::UpdateConfig objected hydrated with
     # relevant information obtained from the job definition.
     #
     # At present we only use this for ignore rules.
+    sig { returns(Dependabot::Config::UpdateConfig) }
     def update_config
-      return @update_config if defined? @update_config
```
jurre, line 356: Is it intentional that we're no longer memoizing this? The operation doesn't look necessarily expensive but maybe we shouldn't lump in logical changes with the Sorbet ones, in case we hit any issues in prod it'll be easier to diagnose?

landongrindheim (approved):

updater/lib/dependabot/job.rb
jurre: Ah, looks like we're assigning it in the initializer and then no longer calling this method, is that something Sorbet wants us to do? We could make the change smaller by calling this `def calculate_update_config` or similar and creating an `attr_reader :update_config`, that way the calling code remains unchanged

updater/lib/dependabot/job.rb
```diff
@@ -43,124 +45,180 @@ class Job
...
+    sig { returns(T::Array[T::Hash[String, T.untyped]]) }
+    attr_reader :allowed_updates
+
+    sig { returns(T::Array[T::Hash[String, T.any(T::Boolean, String)]]) }
+    attr_reader :credentials
+
+    sig { returns(T.nilable(T::Array[String])) }
+    attr_reader :dependencies
+
+    sig { returns(T::Array[T.untyped]) }
```
jurre, line 59: Should this be something like 
```suggestion
    sig { returns(T::Array[T::Hash[T::Symbol, T.any(T::String, T::Bool)]]) }
```

updater/lib/dependabot/job.rb
jakecoffman: It does gently push you towards doing things in the initializer because using `defined?` means that it needs to be `nilable`.

updater/lib/dependabot/job.rb
jurre: On the one hand it seems tedious to type all of that out and read it, but on the other hand, I suppose it's just the price we pay for the benefits of having types

updater/lib/dependabot/job.rb
```diff
@@ -43,124 +45,180 @@ class Job
...
+      @ignore_conditions              =  T.let(attributes.fetch(:ignore_conditions), T::Array[T.untyped])
+      @package_manager                =  T.let(attributes.fetch(:package_manager), String)
+      @reject_external_code           =  T.let(attributes.fetch(:reject_external_code, false), T::Boolean)
+      @repo_contents_path             =  T.let(attributes.fetch(:repo_contents_path, nil), T.nilable(String))
+
+      @requirements_update_strategy   = T.let(build_update_strategy(
+                                                **attributes.slice(:requirements_update_strategy, :lockfile_only)
+                                              ), T.nilable(String))
+
+      @security_advisories            = T.let(attributes.fetch(:security_advisories), T::Array[T.untyped])
```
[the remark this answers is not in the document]
jakecoffman: Well... for example security_advisories is something like `T::Array[T::Hash[String,T.any(String, T::Array[String]])` which is kind of difficult for me to parse...

It would bring a lot of clarity to construct the actual `Dependabot::SecurityAdvisory` in the initializer and present it as `security_advisories` (`T::Array[Dependabot::SecurityAdvisory]`) instead of presenting the raw data and only using `Dependabot::SecurityAdvisory` when calling `security_advisories_for`.

The others are kind of similar, we have proper classes that have types we could use but we'd need to construct it here and then find all the places that currently use the array of hashes, so I am defering that to a future PR.

updater/lib/dependabot/job.rb
jakecoffman: I mentioned above in response to @landongrindheim I think it would be good to construct classes here instead of leaving them as raw arrays of hashes. `ExistingPullRequest` doesn't exist yet but it should! I think it would bring much needed clarity to the codebase.

jakecoffman: I wonder if it would be a better approach to leave this file as untyped, but then slowly type each of the attrs with the actual backing class?

jurre: Maybe `T::Array[T::Hash[T::Symbol, T.untyped]]` as a middle ground?

jakecoffman: Sure, but I think `existing_pull_requests` is actually a `T::Array[T::Array[T::Hash[String, T.untyped]]]` 

https://github.com/dependabot/dependabot-core/blob/f6f2548580dab0ad1b5b6fca2568c14470e6e157/updater/spec/dependabot/updater_spec.rb#L868-L875

jurre: Ah apologies your comments hadn't yet loaded when I suggested that, I think your suggestion of adding types to the backing class is good!

updater/lib/dependabot/job.rb
```diff
@@ -270,8 +336,9 @@ def security_advisories_for(dependency)
       end
     end
 
+    sig { params(dependency: Dependabot::Dependency).returns(T::Array[String]) }
     def ignore_conditions_for(dependency)
-      update_config.ignored_versions_for(
+      @update_config.ignored_versions_for(
```
jurre, line 341:
```suggestion
      update_config.ignored_versions_for(
```

jurre (approved):
Approving, but might be good to test this out on staging to ensure it doesn't break anything?

updater/lib/dependabot/job.rb
jakecoffman: Ah, so `def update_config` was actually private so there's no `attr_reader` for it which is why I used the `@` syntax. We could expose it as an `attr_reader` but it's not needed currently. Should we do that?

jakecoffman:
@jurre for sure, I am willing to bet I got one of these wrong.

updater/lib/dependabot/job.rb
jurre: You can make it a private `attr_reader` by defining after declaring `private`!

jakecoffman: :woah:

jakecoffman:
It's been deployed for 20 minutes and no TypeErrors so far, so I think this is good.

PR #8945, opened 2024-01-31
fix smoke tests failing because Dir.entries order is not deterministic

jakecoffman:
After https://github.com/dependabot/dependabot-core/pull/8933 I'm noticing some smoke tests are failing to pass and the depednency files are coming back in different order depending on what system is running the test.

I think I tracked it down to `Dir.entries` which doesn't guarantee any order, so by sorting we can have more deterministic tests.

Most likely this will break some smoke tests and I'll have to regenerate them.


abdulapopoola (approved):

jakecoffman:
It actually fixed the actions smoke test 🙌

Issue #8886, opened 2024-01-24
NuGet project file parser doesn't handle wildcards


brettfo:
It's not uncommon to find, e.g., a `dirs.proj` file that looks like the following:

``` xml
<Project Sdk="Microsoft.Build.Traversal">
  <ItemGroup>
    <ProjectReference Include="**\*.proj" />
  </ItemGroup>
</Project>
```

Those wildcards aren't currently expanded, so no project traversal can continue.

PR #8956, opened 2024-02-01
Expand wildcards in nuget project references

sebasgomez238:
First time contributing to the repo so all feedback is welcome.

This PR addresses #8886 where the nuget project file parser doesn't expand wildcards. This PR changes the behavior to place all paths that match the wildcard into an array and has the existing logic iterate over the array. If the given path is not a wildcard then the array still be initialized and contain the single path.

As a part of this fix I had to pass in the repo_path into the class in order to get access to the actual full path otherwise Dir.glob would be unable to expand the wildcard.


nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
```diff
@@ -151,21 +152,35 @@ def add_transitive_dependencies_from_project_references(project_file, doc, depen
...
+                  package_manager: dep.package_manager,
+                  requirements: []
+                )
+                dependency_set << dependency
+              end
             end
           end
         end
 
+        def handle_wildcard_path_in_project_references(full_path, _dependency_set)
```
ryanbrandenburg, line 175: Maybe `expand_wildcards_in_project_reference_path`? Also it looks like you can omit `_dependency_set` (the ruby tooling adds `_` to the front of unused parameters to disable the "you don't use this" warning).

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
```diff
@@ -151,21 +152,35 @@ def add_transitive_dependencies_from_project_references(project_file, doc, depen
...
+          return [full_path] unless full_path.include?("*")
```
ryanbrandenburg, line 176: Technically [`?` is also a wild-card](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-items?view=vs-2022#use-wildcards-to-specify-items) (though I don't believe it's common).

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
```diff
@@ -151,21 +152,35 @@ def add_transitive_dependencies_from_project_references(project_file, doc, depen
...
+
+          full_path = T.let(File.join(@repo_path, full_path), T.nilable(String))
```
ryanbrandenburg, line 178: I would have thought this `T.let` was redundant since I assume File.join is already typed?

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
```diff
@@ -151,21 +152,35 @@ def add_transitive_dependencies_from_project_references(project_file, doc, depen
...
+          expanded_wildcard = Dir.glob(T.must(full_path))
```
ryanbrandenburg, line 179: Is the Glob behavior of Ruby.Dir and MSBuild the same? It seems like it might be, and even if it's not exactly the same it's probably a "good enough" approximation but it's worth considering and leaving a comment about.

ryanbrandenburg (approved):
Approval with the priviso that I would like to see a test of this since it's a more complex problem ("*", "**", "?", %2A, etc. Don't feel the need to cover all those at once but let's add one so that someone who has a problem with other scenerios in the area in the future has an easier time adding coverage for that).  https://github.com/dependabot/dependabot-core/pull/8940 adds a test for project_file_parser.rb so once that gets merged you can build on it.

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
JoeRobich: With `@repo_path` being optional, the `File.join` would error if it were `nil`.

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
```diff
@@ -44,9 +44,10 @@ def self.dependency_url_search_cache
           CacheManager.cache("dependency_url_search_cache")
         end
 
-        def initialize(dependency_files:, credentials:)
+        def initialize(dependency_files:, credentials:, repo_path: nil)
```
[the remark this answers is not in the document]
ryanbrandenburg: I believe attr_reader enforcement is for `typed: strict`, and this file is `typed: true`.

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
JamieMagee, line 175: Can you a method signature please? I think it's:
```
sig { params(full_path: T.nilable(String)).returns(T.nilable(String)) }
```

But double check if `T.nilable` is required.

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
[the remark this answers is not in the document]
ryanbrandenburg: I think the idea is "If there's no wildcards then the correct expansion is a single result, the original path".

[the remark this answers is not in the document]
JamieMagee: Ah, I misunderstood what `Dir.glob` does.

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
ryanbrandenburg: If you turn `# typed: strict` on (even for just a minute) it will give you an error here and you can just use the quick-fix (I just did a bunch of them and they're pretty reliable on "normal" methods).

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
sebasgomez238: As you mentioned, it looks like any valid MSBuild glob is a valid Ruby.Dir glob. Ruby's Dir.Glob seems a bit more flexible but I don't think that will cause any issues

[MSBuild Glob](https://learn.microsoft.com/en-us/dotnet/api/microsoft.build.globbing.msbuildglob?view=msbuild-17-netcore)

[Dir.Glob](https://ruby-doc.org/3.2.2/Dir.html#method-c-glob)

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
sebasgomez238: I didn't know about the `_`. Thanks!

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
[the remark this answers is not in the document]
sebasgomez238: I renamed the variable. But for making it required I ended up having to pass it around many different classes that previously didn't have access to it, nor would they make use of it. Do you still think it should be made required?

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
```diff
@@ -114,8 +117,47 @@ public static IEnumerable<string> GetProjectPathsFromProject(string projFilePath
                     continue;
                 }
 
+                var fullPath = PathHelper.GetFullPathFromRelative(folderPath, projectPath);
+                if (fullPath.Contains('*'))
```
JoeRobich, line 121: You may want to see if we can just plug in the Microsoft.Extensions.FileGlobbing library to get back the matching files. See https://learn.microsoft.com/en-us/dotnet/core/extensions/file-globbing#get-all-matching-files

sebasgomez238: Thank you! I was not aware of this. This allowed. me to clean this up these changes quite a bit.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
```diff
@@ -97,6 +99,9 @@ public static IEnumerable<string> GetProjectPathsFromSolution(string solutionPat
         return solution.ProjectsInOrder.Select(p => p.AbsolutePath);
     }
 
+    //Create a hash set to store the processed project files
+    private static readonly HashSet<string> processedProjectFiles = new HashSet<string>();
```
sebasgomez238, line 103: I'm not 100% sure about making this static but if I didn't keep track of processed files then I would end up in an infinite loop.

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
```diff
@@ -46,16 +48,20 @@ def self.dependency_url_search_cache
...
 
-        def initialize(dependency_files:, credentials:)
+        def initialize(dependency_files:, credentials:, repo_contents_path: nil)
           @dependency_files       = dependency_files
           @credentials            = credentials
+          @repo_contents_path     = repo_contents_path
         end
 
-        def dependency_set(project_file:)
+        def dependency_set(project_file:, visited_project_files: Set.new)
```
sebasgomez238, line 57: Similar issue here where we would end up in an infinite loop. Each iteration had a clear cache so we would end up looping over the same set of files after expanding the wildcard. Could this cause problems in large projects?

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
```diff
@@ -14,7 +14,9 @@
 module Dependabot
   module Nuget
     class FileParser
-      class ProjectFileParser
+      class ProjectFileParser # rubocop:disable Metrics/ClassLength
```
sebasgomez238, line 17: Class was over the length limit but I feel the logic belongs in the same file.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
JoeRobich: I would move this to be a local variable in the `GetProjectPathsFromProject` method. It should only be run once in a typical run. However, having it be static could lead to test failures in the case there are multiple tests using the same project paths.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
```diff
@@ -114,21 +119,37 @@ public static IEnumerable<string> GetProjectPathsFromProject(string projFilePath
                     continue;
                 }
 
-                projectPath = PathHelper.GetFullPathFromRelative(folderPath, projectPath);
+                Matcher matcher = new Matcher();
+                matcher.AddInclude(projectReference.Include);
+                
+                var fullPath = PathHelper.GetFullPathFromRelative(folderPath, projectPath);
```
JoeRobich, line 124: This appears to be unused.

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
```diff
@@ -179,21 +192,53 @@ def add_transitive_dependencies_from_project_references(project_file, doc, depen
...
+              end
             end
           end
         end
 
+        sig { params(full_path: T.untyped).returns(T::Array[T.nilable(String)]) }
+        def expand_wildcards_in_project_reference_path(full_path)
+          return [full_path] unless full_path.include?("*") || full_path.include?("?")
+
+          full_path = T.let(File.join(@repo_contents_path, full_path), T.nilable(String))
```
JoeRobich, line 223: `File.join` will throw an error if `@repo_contents_path` is `nil`. Should we make `@repo_contents_path` a required constructor parameter?

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
```diff
@@ -114,21 +118,37 @@ public static IEnumerable<string> GetProjectPathsFromProject(string projFilePath
...
+
+                string searchDirectory = folderPath;
 
-                var projectExtension = Path.GetExtension(projectPath).ToLowerInvariant();
-                if (projectExtension == ".proj")
+                IEnumerable<string> files = matcher.GetResultsInFullPath(searchDirectory);
```
JoeRobich, line 128: We need a test in UpdateWorker.DirsProj.cs for this behavior.

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
```diff
@@ -179,21 +192,53 @@ def add_transitive_dependencies_from_project_references(project_file, doc, depen
...
+                )
+                dependency_set << dependency
+              end
             end
           end
         end
 
+        sig { params(full_path: T.untyped).returns(T::Array[T.nilable(String)]) }
+        def expand_wildcards_in_project_reference_path(full_path)
+          return [full_path] unless full_path.include?("*") || full_path.include?("?")
```
brettfo, line 221: You can skip this shortcut.  `Dir.glob()` will return just a single file if that's all that was specified.  E.g., calling it with `some/file.txt` will return `['some/file.txt']` and calling it with `some/*.txt` will return `['some/file.txt', 'some/other-file.txt']`.

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
```diff
@@ -179,21 +192,53 @@ def add_transitive_dependencies_from_project_references(project_file, doc, depen
...
+
+          full_path = T.let(File.join(@repo_contents_path, full_path), T.nilable(String))
+          expanded_wildcard = Dir.glob(T.must(full_path))
```
brettfo, line 224: [Call](https://ruby-doc.org/core-2.6.3/Dir.html#method-c-glob) with the flag parameter `File::FNM_CASEFOLD` to make it case-insensitive, since `dirs.proj` is a Windows-like experience.  I've seen repos that have `SomeProject.CSProj`, etc.

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
brettfo, line 223: Since `dirs.proj` is a very Windows-like experience, we should probably ensure that all `\` are replaced with `/`.

sebasgomez238: Paths are normalized before passing them into this function by calling https://github.com/dependabot/dependabot-core/blob/6a5f72e01743587260ce98f4eeea2296e85f7cba/nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb#L119 just on the path just before expanding.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
sebasgomez238: Moved this to be a local variable inside `GetProjectPathsFromProject()`

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
sebasgomez238: Added a couple of tests.

JoeRobich (approved):

michal-plebanski-ideals:
It seems that this change introduced a regression, my error stack: 
```
[2024-03-28 21:11:28 +0000] ERROR -- no implicit conversion of nil into String
[2024-03-28 21:11:28 +0000] ERROR -- /home/dependabot/app/vendor/bundle/ruby/3.1.0/gems/dependabot-nuget-0.245.0/lib/dependabot/nuget/file_parser/project_file_parser.rb:221:in `join'
/home/dependabot/app/vendor/bundle/ruby/3.1.0/gems/dependabot-nuget-0.245.0/lib/dependabot/nuget/file_parser/project_file_parser.rb:221:in `expand_wildcards_in_project_reference_path'
/home/dependabot/app/vendor/bundle/ruby/3.1.0/gems/sorbet-runtime-0.5.11266/lib/types/private/methods/call_validation.rb:270:in `bind_call'
/home/dependabot/app/vendor/bundle/ruby/3.1.0/gems/sorbet-runtime-0.5.11266/lib/types/private/methods/call_validation.rb:270:in `validate_call'
/home/dependabot/app/vendor/bundle/ruby/3.1.0/gems/sorbet-runtime-0.5.11266/lib/types/private/methods/_methods.rb:277:in `block in _on_method_added'
```

and my .csproj file content: 

```
        <PackageReference Include="<redacted>" CustomXmlVersion="$(SharedLibrary)">
            <CopyToOutputDirectory>lib/$(TargetFramework)/*.xml</CopyToOutputDirectory>
        </PackageReference>
```

seems to breaking things. @sebasgomez238 could you please tell me if this kind of wildcard usage is something that could be breaking dependabot?

FYI @honeyankit

PR #8957, opened 2024-02-01
fix security updates getting into grouped code

jakecoffman:
In https://github.com/dependabot/dependabot-core/pull/8742 the `applies_to?` method was too permissive so singular security updates are trying to create in groups.


mctofu (approved):

jakecoffman:
Smoke test failures are from invalid inputs, should be fixed by: https://github.com/dependabot/smoke-tests/pull/179

PR #8958, opened 2024-02-01
Expand Sorbet usage

ryanbrandenburg:
I was working in a couple of these classes recently and having type assertions would have made my experience easier so I figured I'd do a quick cleanup for them.

### DIFF
--- a/nuget/lib/dependabot/nuget/file_parser.rb
+++ b/nuget/lib/dependabot/nuget/file_parser.rb
@@ -1,4 +1,4 @@
-# typed: true
+# typed: strict
 # frozen_string_literal: true
 
 require "nokogiri"
@@ -72,12 +72,15 @@ module Dependabot
         DotNetToolsJsonParser.new(dotnet_tools_json: dotnet_tools_json).dependency_set
       end
 
+      sig { returns(Dependabot::Nuget::FileParser::ProjectFileParser) }
       def project_file_parser
-        @project_file_parser ||=
+        @project_file_parser ||= T.let(
           ProjectFileParser.new(
             dependency_files: dependency_files,
             credentials: credentials
-          )
+          ),
+          T.nilable(Dependabot::Nuget::FileParser::ProjectFileParser)
+        )
       end
 
       sig { returns(T::Array[Dependabot::DependencyFile]) }
--- a/nuget/lib/dependabot/nuget/file_parser.rb
+++ b/nuget/lib/dependabot/nuget/file_parser.rb
@@ -12,6 +12,8 @@ require "dependabot/file_parsers/base"
 module Dependabot
   module Nuget
     class FileParser < Dependabot::FileParsers::Base
+      extend T::Sig
+
       require "dependabot/file_parsers/base/dependency_set"
       require_relative "file_parser/project_file_parser"
       require_relative "file_parser/packages_config_parser"
@@ -20,6 +22,7 @@ module Dependabot
 
       PACKAGE_CONF_DEPENDENCY_SELECTOR = "packages > packages"
 
+      sig { override.returns(T::Array[Dependabot::Dependency]) }
       def parse
         dependency_set = DependencySet.new
         dependency_set += project_file_dependencies
@@ -31,6 +34,7 @@ module Dependabot
 
       private
 
+      sig { returns(Dependabot::FileParsers::Base::DependencySet) }
       def project_file_dependencies
         dependency_set = DependencySet.new
 
@@ -42,6 +46,7 @@ module Dependabot
         dependency_set
       end
 
+      sig { returns(Dependabot::FileParsers::Base::DependencySet) }
       def packages_config_dependencies
         dependency_set = DependencySet.new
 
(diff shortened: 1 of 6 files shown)


nuget/lib/dependabot/nuget/nuget_client.rb
```diff
@@ -49,38 +58,32 @@ def self.get_package_versions(dependency_name, repository_details)
...
         return unless body
 
         pages = body.fetch("items")
-        versions = Set.new
+        versions = T.let(Set.new, T::Set[String])
         pages.each do |page|
           items = page["items"]
           if items
             # inlined entries
-            items.each do |item|
```
ryanbrandenburg, line 69: I split this up because rubocop was complaining about perceived complexity after I added an extra assertion to deal with a missing nil-check.

nuget/lib/dependabot/nuget/nuget_client.rb
```diff
@@ -92,8 +95,27 @@ def self.get_package_versions(dependency_name, repository_details)
...
+        end
+      end
+
+      sig do
+        params(repository_details: T::Hash[Symbol, String], dependency_name: String)
+          .returns(T.nilable(T::Set[String]))
+      end
       private_class_method def self.get_versions_from_search_url_v3(repository_details, dependency_name)
-        search_url = repository_details[:search_url]
+        search_url = repository_details.fetch(:search_url)
```
ryanbrandenburg, line 118: Where I've changed `hash.[thing]` to `hash.fetch(thing)` it's because fetch assures a non-nil result but `[]` does not, and we should already be sure that the value exists in those places.

nuget/lib/dependabot/nuget/version.rb
```diff
@@ -11,45 +11,54 @@
...
         compare_prerelease_part(other)
       end
 
+      sig { params(other: T.untyped).returns(Integer) }
       def compare_release(other)
         release_str = @version_string.split("-").first || ""
         other_release_str = other.to_s.split("-").first || ""
 
-        Gem::Version.new(release_str) <=> Gem::Version.new(other_release_str)
+        T.must(Gem::Version.new(release_str) <=> Gem::Version.new(other_release_str))
```
ryanbrandenburg, line 57: Turns out `<=>` can return nil if the two values are incomparable. We know they're comparable in our cases so I did `T.must`.

common/lib/dependabot/file_parsers/base.rb
```diff
@@ -49,7 +49,7 @@ def initialize(dependency_files:, source:, repo_contents_path: nil,
         check_required_files
       end
 
-      sig { abstract.returns(Dependabot::DependencyFile) }
+      sig { abstract.returns(T::Array[Dependabot::Dependency]) }
```
JamieMagee, line 52: Looks like it might actually be:

```suggestion
      sig { abstract.returns(Dependabot::DependencySet) }
```

nuget/lib/dependabot/nuget/file_parser.rb
JamieMagee, line 1: Is it possible to upgrade this to `# typed: strict`?

nuget/lib/dependabot/nuget/version.rb
```diff
@@ -11,45 +11,54 @@
...
       end
 
       # rubocop:disable Metrics/PerceivedComplexity
+      sig { params(other: T.untyped).returns(Integer) }
```
JamieMagee, line 61: I think you can be a little more specific with `other` here. Maybe `BasicObject`?

nuget/lib/dependabot/nuget/version.rb
```diff
@@ -11,45 +11,54 @@
...
       def to_s
         @version_string
       end
 
+      sig { returns(String) }
       def inspect # :nodoc:
         "#<#{self.class} #{@version_string}>"
       end
 
+      sig { params(other: T.untyped).returns(Integer) }
```
JamieMagee, line 44: Maybe `BasicObject` instead of `T.untyped`?

nuget/lib/dependabot/nuget/version.rb
```diff
@@ -11,45 +11,54 @@
...
       ANCHORED_VERSION_PATTERN = /\A\s*(#{VERSION_PATTERN})?\s*\z/
 
+      sig { override.params(version: T.nilable(T.any(String, Integer, Float, Gem::Version))).returns(T::Boolean) }
       def self.correct?(version)
         return false if version.nil?
 
         version.to_s.match?(ANCHORED_VERSION_PATTERN)
       end
 
+      sig { override.params(version: T.nilable(T.any(String, Integer, Float, Gem::Version))).void }
```
JamieMagee, line 26: Is `T.any` necessary? Or can we be more specific? I know `T.any(String, Gem::Version)` is common, but `Integer` and `Float` are unusual.

JamieMagee:
Don't forget `require "sorbet-runtime"` in all files.

common/lib/dependabot/file_parsers/base.rb
ryanbrandenburg: The ones I looked at all used `DependencySet`, but actually were returning `Array[Dependabot::Dependency]` because they called `dependency_set.dependencies` (see [elm](https://github.com/dependabot/dependabot-core/blob/main/elm/lib/dependabot/elm/file_parser.rb#L22) or [python](https://github.com/dependabot/dependabot-core/blob/81ac70fde4c400110310f1b3804b09edfe76bcec/python/lib/dependabot/python/file_parser.rb#L48)).  If there's an ecosystem(s) that returns DependencySet you know of maybe we need to do `T.any(DependencySet, Array[Dependency])`?

ryanbrandenburg: We talked offline and it looks like they're all returning Array[Dependency].

JamieMagee (approved):

abdulapopoola (approved):

Issue #8949, opened 2024-02-01
Gomod dependency gets downgraded after manual update


achilleas-k:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

gomod

### Package manager version

go 1.19

### Language version

go 1.19

### dependabot.yml content

https://github.com/osbuild/images/blob/main/.github/dependabot.yml

```yaml
version: 2
updates:

  # Maintain dependencies for GitHub Actions
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
      time: "04:00"
    open-pull-requests-limit: 5
    rebase-strategy: "disabled"

  # Maintain dependencies for Go
  - package-ecosystem: "gomod"
    directory: "/"
    schedule:
      interval: "daily"
      time: "04:00"
    groups:
      go-deps:
        patterns:
          - "*"  # group all dependency updates into one PR
    open-pull-requests-limit: 1
    rebase-strategy: "auto"
```

### Updated dependency

github.com/containers/image/v5

### What you expected to see, versus what you actually saw

**What I expect:** dependabot should not be downgrading a dependency 
**What happens:** dependabot PRs downgrade a dependency every time it runs

**Description**

About a week ago, a dependabot PR pulled in updates that were incompatible with a specific dependency.  I manually updated the dependency from v5.22.1 to v5.29.1 and included it in the PR.
- dependabot PR: https://github.com/osbuild/images/pull/390
- Breaking commit: https://github.com/osbuild/images/commit/c8d9c5eff6928931faf7ff9e92d09d4f53572d1e
- Manual dependency update commit: https://github.com/osbuild/images/commit/13622d01ba62ec9a4cb946c5c82ac97bf1bdf0b7

Since then, every dependabot PR reverts the change to v5.22.1 of github.com/containers/image/v5, but fails because of the incompatibility.

Most recent example: https://github.com/osbuild/images/pull/416 (see: https://github.com/osbuild/images/pull/416/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L17)
See also:
- https://github.com/osbuild/images/pull/397
- https://github.com/osbuild/images/pull/401
- https://github.com/osbuild/images/pull/403
- https://github.com/osbuild/images/pull/406

### Images of the diff or a link to the PR, issue, or logs

- https://github.com/osbuild/images/pull/397
[...]

Issue #8847, opened 2024-01-19
Dependabot downgrades instead of stated upgrade (related to double tagging?)


dtrudg:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

Go

### Language version

1.21.6

### Manifest location and content before the Dependabot update

https://github.com/sylabs/singularity/blob/637e3eeeaf0265f11814006f271cc00c33532e64/go.mod

### dependabot.yml content

```
version: 2
updates:
  - package-ecosystem: "gomod" # See documentation for possible values
    directory: "/" # Location of package manifests
    schedule:
      interval: "daily"
    target-branch: main
  - package-ecosystem: "gomod" # See documentation for possible values
    directory: "/" # Location of package manifests
    schedule:
      interval: "daily"
    target-branch: release-4.1
  - package-ecosystem: "gomod" # See documentation for possible values
    directory: "/" # Location of package manifests
    schedule:
      interval: "daily"
    target-branch: release-4.0
  - package-ecosystem: "gomod" # See documentation for possible values
    directory: "/" # Location of package manifests
    schedule:
      interval: "daily"
    target-branch: release-3.11
```

### Updated dependency

github.com/opencontainers/image-spec from 1.1.0-rc5 to 1.1.0-rc.6

### What you expected to see, versus what you actually saw

The github.com/opencontainers/image-spec project tagged a new RC.

It appears that this was tagged as both `1.1.0-rc.6` and `1.1.0-rc6` ~(in error?)~, as queried with them here: https://github.com/opencontainers/image-spec/issues/1159

Dependabot opened a PR at https://github.com/sylabs/singularity/pull/2557 which announces that it is updating `github.com/opencontainers/image-spec from 1.1.0-rc5 to 1.1.0-rc.6`. Instead it is downgrading to the `1.1.0-rc3` tag, and causing a large number of other dependencies to be downgraded also.

See - https://github.com/sylabs/singularity/pull/2557/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6

### Native package manager behavior

Native `go get -u` updates to `v1.1.0-rc6`, with no other `go.mod` changes.

Note that it picked `v1.1.0-rc6` and not the `v1.1.0-rc.6` that dependabot reported attempting.

```
02:26 pm $ go get -u github.com/opencontainers/image-spec
go: downloading github.com/opencontainers/image-spec v1.1.0-rc6
[...]

PR #8962, opened 2024-02-02
fix Go prerelease ordering

jakecoffman:
A popular project [released](https://github.com/opencontainers/image-spec/releases) both a v1.1.0-rc6 and a v1.1.0-rc.6 which is causing downgrades because Dependabot sorts pre-releases differently than Go.

Dependabot thinks v1.1.0-rc.6 is the latest and so it runs `go get` with that version, but Go sees that as a downgrade, it's less than `-rc0`.

So to prevent that I've ported Go's [comparePrerelease](https://github.com/golang/mod/blob/fa1ba4269bda724bb9f01ec381fbbaf031e45833/semver/semver.go#L333).

Fixes:
- https://github.com/dependabot/dependabot-core/issues/8949
- https://github.com/dependabot/dependabot-core/issues/8847


mctofu (approved):

go_modules/lib/dependabot/go_modules/version.rb
```diff
@@ -39,6 +42,73 @@ def inspect # :nodoc:
...
         @version_string
       end
+
+      def <=>(other)
+        result = super(other)
+        return if result.nil?
+        return result unless result.zero?
+
+        other = self.class.new(other) unless other.is_a?(Version)
+        compare_prerelease(@prerelease || "", T.unsafe(other).prerelease || "")
```
JamieMagee, line 52: I think this is because of the signature on `def <=>` of `Dependabot::Version`. If this needs to be overridden in every `Dependabot::ECOSYSTEM::Version` class then it needs to be a generic. See https://sorbet.org/docs/override-checking#what-if-i-really-want-the-child-method-to-narrow-the-type

I'm not asking you to fix it, more of in case you're interested.

JamieMagee (approved):

jakecoffman:
We had a weird PR in the Dependabot CLI that this seems to have fixed: https://github.com/dependabot/cli/pull/265

PR #8963, opened 2024-02-02
make DependencySnapshot aware of multiple directories

jakecoffman:
This PR refactors DependencySnapshot to be aware of multiple directories.

## Why

Previously DependencySnapshot would take all of the dependency files and attempt to parse them to discover the dependencies. This breaks down with multi-dir, as really what we need to do is group the files by directory and then parse them, storing the dependencies for each directory separately.

Also, we need to keep track of `handled_dependencies` per directory as well or we end up thinking we've handled it in one directory, but when we encounter the same dependency in another directory then we've missed an update. Previously I was clearing the `handled_dependencies` between directories, but then the problem is we are dropping individual PRs that should have been created.

## Approach

For now I've encapsulated multi-dir support inside of DependencySnapshot by using `@current_directory` as a way to set which directory is currently being operated on. This prevents multi-dir from rippling out into the other non-group operations.

The group operations will set `dependency_snapshot.current_directory(dir)` which then makes all of the pre-existing methods continue to work as they always have. 

If I had made all of the methods take a directory param instead I would have had to touch many many more files and specs (I started down this path at first and it was overwhelming!)

Admittedly this looks a bit complex, but I hope getting integration tests down outside of Ruby/rspec will allow us to take a bold refactor to make it all much simpler in the end.


jakecoffman:
Looking at the smoke test failures, it looks like this fixes some issues, and the setup was wrong on two of them so https://github.com/dependabot/smoke-tests/pull/185 fixes that up.

I think this is good to go!

updater/lib/dependabot/dependency_snapshot.rb
```diff
@@ -37,26 +37,82 @@ def self.create_from_job_definition(job:, job_definition:)
     attr_reader :base_commit_sha
 
     sig { returns(T::Array[Dependabot::DependencyFile]) }
-    attr_reader :dependency_files
+    def all_dependency_files
+      @dependency_files
+    end
+
+    sig { returns(T::Array[Dependabot::DependencyFile]) }
+    def dependency_files
```
jurre, line 45: Minor:

```suggestion
    def dependency_files_for_current_directory
```

For clarity maybe? 🤷

jakecoffman: Outside of the class it's only used in `update_all_versions` which doesn't have the concept of a current directory. I'd say leave this for now and we'll see how future refactors feel about it 🤷 .

jakecoffman:
I'll fix the smoke tests once I start deploying, all the failures look correct, if that makes sense.

jurre (approved):

PR #8967, opened 2024-02-02
make a Credential class

jakecoffman:
credentials was incorrectly typed as a `T::Array[T::Hash[String,String]]` but because of `replaces-base` it's actually a `T::Array[T::Hash[String,T.any(T::Boolean,String)]]`. 

I thought since I have to touch all these places to change the type I could introduce a class instead to make it nicer.


jakecoffman:
Ok maybe now that I have to change all the specs I might regret it.

common/lib/dependabot/credential.rb
```diff
@@ -0,0 +1,30 @@
+# typed: strict
```
JamieMagee, line 1: Can this be `strong`?

jakecoffman: strong gets mad that `def_delegators` brings in untyped methods. I decided to punt on figuring that out or reimplementing them like with `[]` for now. I think the PR got large enough 😬

jurre (approved):

lorengordon:
@jakecoffman Just a heads up that this also breaks anyone using the [setup defined in dependabot-script](https://github.com/dependabot/dependabot-script/blob/main/update-script.rb#L54-L59). Any suggestions on how to set things up now, using the class?

lorengordon:
I hope it's this easy...

```diff
diff --git a/dependabot-codecommit/dependabot.rb b/dependabot-codecommit/dependabot.rb
index 79c3430..b32cc46 100644
--- a/dependabot-codecommit/dependabot.rb
+++ b/dependabot-codecommit/dependabot.rb
@@ -6,6 +6,7 @@ require_relative "updater"
 require_relative "client"
 require_relative "utils"

+require "dependabot/credential"
 require "dependabot/file_fetchers"
 require "logger"

@@ -18,12 +19,12 @@ logger.level = Logger::INFO

 error_set = Set[]
 credentials = [
-  {
+  Dependabot::Credential.new({
     "type" => "git_source",
     "host" => "github.com",
     "username" => "x-access-token",
     "password" => ENV["GITHUB_ACCESS_TOKEN"] # A GitHub access token with read access to public repos
-  }
+  })
 ]
```

lorengordon:
Yes, looks like that did it!

PR #8970, opened 2024-02-03
Strict type `Dependabot::GitSubmodules`

JamieMagee:


git_submodules/lib/dependabot/git_submodules/update_checker.rb
```diff
@@ -10,20 +12,30 @@
...
+          )
       end
 
+      sig { override.returns(T.nilable(T.any(String, Gem::Version))) }
       def latest_resolvable_version
         # Resolvability isn't an issue for submodules.
         latest_version
       end
 
+      sig { override.returns(T.nilable(T.any(String, Gem::Version))) }
```
jakecoffman, line 32: Should we be using `Dependabot::Version`? 🤔 

```suggestion
      sig { override.returns(T.nilable(T.any(String, Dependabot::Version))) }
```

jakecoffman (approved):

git_submodules/lib/dependabot/git_submodules/update_checker.rb
JamieMagee: Ideally, yes (Actually, I think it should be a generic that accepts `Dependabot::ECOSYSTEM::Version`). But `Gem::Version` and `Dependabot::Version` are used interchangeably throughout the codebase at the moment. Once it's fully typed it will be much easier to refactor.

PR #8986, opened 2024-02-05
add types to DependencySnapshot

jakecoffman:
I changed a few of the memoized methods into methods that calculate during initialization as the types work out much more easily this way.

### DIFF
--- a/updater/lib/dependabot/api_client.rb
+++ b/updater/lib/dependabot/api_client.rb
@@ -170,7 +170,7 @@ module Dependabot
       span&.finish
     end
 
-    sig { params(dependencies: T::Array[T::Hash[Symbol, T.untyped]], dependency_files: T::Array[DependencyFile]).void }
+    sig { params(dependencies: T::Array[T::Hash[Symbol, T.untyped]], dependency_files: T::Array[String]).void }
     def update_dependency_list(dependencies, dependency_files)
       span = ::Dependabot::OpenTelemetry.tracer&.start_span("update_dependency_list", kind: :internal)
       span&.set_attribute(::Dependabot::OpenTelemetry::Attributes::JOB_ID, job_id)
--- a/updater/lib/dependabot/dependency_snapshot.rb
+++ b/updater/lib/dependabot/dependency_snapshot.rb
@@ -1,4 +1,4 @@
-# typed: true
+# typed: strict
 # frozen_string_literal: true
 
 require "base64"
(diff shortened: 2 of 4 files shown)


updater/lib/dependabot/api_client.rb
```diff
@@ -170,7 +170,7 @@ def mark_job_as_processed(base_commit_sha)
       span&.finish
     end
 
-    sig { params(dependencies: T::Array[T::Hash[Symbol, T.untyped]], dependency_files: T::Array[DependencyFile]).void }
+    sig { params(dependencies: T::Array[T::Hash[Symbol, T.untyped]], dependency_files: T::Array[String]).void }
```
jakecoffman, line 173: I guess runtime validation isn't working for this part of the code because it's been wrong and I don't see any TypeErrors for it. It would happen for each job so we'd definitely notice.

updater/lib/dependabot/dependency_snapshot.rb
```diff
@@ -89,25 +75,34 @@ def ungrouped_dependencies
...
+    end
     def initialize(job:, base_commit_sha:, dependency_files:)
       @job = job
       @base_commit_sha = base_commit_sha
       @dependency_files = dependency_files
-      @handled_dependencies = Set.new
+      @handled_dependencies = T.let(Set.new, T::Set[String])
 
-      @dependencies = parse_files!
+      @dependencies = T.let(parse_files!, T::Array[Dependabot::Dependency])
```
jakecoffman, line 87: Kind of interesting that Sorbet doesn't infer the types for these methods that are defined in the same file.

JamieMagee (approved):

updater/lib/dependabot/api_client.rb
jurre: How come we're exposing this now, and can we verify that it's actually always taking a string?

jurre (approved):

updater/lib/dependabot/api_client.rb
jakecoffman: You can see in all the smoke tests, the data for the `update_dependency_list` endpoint's `dependency_files` is always an array of strings: https://github.com/dependabot/smoke-tests/blob/258269368c3c68116cf569bfed8bbafa5fcdc8e5/tests/smoke-bundler.yaml#L97

> How come we're exposing this now

Now that we've got more and more files typed, things are falling into place and mistakes made earlier are getting fixed. If there was no runtime error then no problem I guess 🤷

PR #8987, opened 2024-02-05
type DependencyGroupEngine

jakecoffman:
stacked on #8986

I'd like to merge this into that PR and then deploy them both. This was to make it easier to review. 

I can't use `typed: strong` yet because the dependency group passed in is untyped. In a followup refactor I'll try to address that.

### DIFF
--- a/updater/lib/dependabot/dependency_group_engine.rb
+++ b/updater/lib/dependabot/dependency_group_engine.rb
@@ -1,4 +1,4 @@
-# typed: true
+# typed: strict
 # frozen_string_literal: true
 
 require "dependabot/dependency_group"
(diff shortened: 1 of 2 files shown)


updater/lib/dependabot/dependency_group_engine.rb
```diff
@@ -44,12 +47,18 @@ def self.from_job_config(job:)
       new(dependency_groups: groups)
     end
 
-    attr_reader :dependency_groups, :groups_calculated, :ungrouped_dependencies
+    sig { returns(T::Array[Dependabot::DependencyGroup]) }
+    attr_reader :dependency_groups
+
+    sig { returns(T::Array[Dependabot::Dependency]) }
+    attr_reader :ungrouped_dependencies
```
Nishnha, line 54: Dunno if it matters here but these could both be nil I think

Nishnha (approved):
Had one question about the attr_reader variables but the types generally LGTM

updater/lib/dependabot/dependency_group_engine.rb
jakecoffman: `dependency_groups` is initialized to be an empty array here: https://github.com/dependabot/dependabot-core/blob/89737ae7b29468822ae3fc81544c31db2650d7d1/updater/lib/dependabot/job.rb#L171

`ungrouped_dependencies` is initialized to an empty array here: https://github.com/dependabot/dependabot-core/blob/7260c518278f115c25807c4f3b7cbed389ccdab6/updater/lib/dependabot/dependency_group_engine.rb#L90

So I think we're good.

JamieMagee (approved):

PR #8990, opened 2024-02-05
escape nuget feed urls before querying

brettfo:
An internal repo was failing to query packages because one of the NuGet feeds had a space resulting in the following error:

```
URI::InvalidURIError:
       bad URI(is not URI?): "https://www.myget.org/F/exceptionless/api with spaces/v3/index.json"
```

The fix is to simply escape the feed URLs.  The escaping is done right after consolidating all feeds from NuGet.Config files as well as feeds from `dependabot.yml`.


edgarrs (approved):

JoeRobich (commented):
👍

PR #8995, opened 2024-02-06
fix TypeError: no implicit conversion of Credential into Hash

jakecoffman:
Introduced by https://github.com/dependabot/dependabot-core/pull/8967, this is some untested and untyped code which is failing. 

I added the missing functionality required by this code to the Credential class and added a test for good measure.


jakecoffman:
Some other tests broke, let me fix that 🔨

PR #8998, opened 2024-02-06
Report Sorbet issue to sentry without raising

jurre:
Previously if Sorbet found runtime typecheck failures, these would result in an exception, halting further job execution.

While we should be informed of those failures, there is no reason why this should end up failing the job for our customers.

This PR configures Sorbet to report the exceptions to Sentry but continue processing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 <h1 align="center">
+
     <picture>
         <source media="(prefers-color-scheme: light)" srcset="https://user-images.githubusercontent.com/7659/174594540-5e29e523-396a-465b-9a6e-6cab5b15a568.svg">
         <source media="(prefers-color-scheme: dark)" srcset="https://user-images.githubusercontent.com/7659/174594559-0b3ddaa7-e75b-4f10-9dee-b51431a9fd4c.svg">
--- a/updater/lib/dependabot/setup.rb
+++ b/updater/lib/dependabot/setup.rb
@@ -8,6 +8,7 @@ require "dependabot/logger"
 require "dependabot/logger/formats"
 require "dependabot/opentelemetry"
 require "dependabot/sentry"
+require "dependabot/sorbet/runtime"
 
 Dependabot.logger = Logger.new($stdout).tap do |logger|
   logger.level = Dependabot::Environment.log_level
@@ -50,6 +51,7 @@ Sentry.init do |config|
 end
 
 Dependabot::OpenTelemetry.configure
+Dependabot::Sorbet::Runtime.silently_report_errors!
 
 # Ecosystems
 require "dependabot/python"
--- a/updater/lib/dependabot/sorbet/runtime.rb
+++ b/updater/lib/dependabot/sorbet/runtime.rb
@@ -15,7 +15,7 @@ module Dependabot
           error = InformationalError.new(opts[:pretty_message])
           error.set_backtrace(caller.dup)
 
-          Sentry.capture_exception(error)
+          ::Sentry.capture_exception(error)
         end
       end
     end
--- a/updater/lib/dependabot/sorbet/runtime.rb
+++ b/updater/lib/dependabot/sorbet/runtime.rb
@@ -1,13 +1,15 @@
-# typed: true
+# typed: strict
 # frozen_string_literal: true
-#
+
 require "sorbet-runtime"
 
 module Dependabot
   module Sorbet
     module Runtime
       class InformationalError < StandardError; end
+      extend T::Sig
 
+      sig { void }
       def self.silently_report_errors!
         T::Configuration.call_validation_error_handler = lambda do |sig, opts|
           error = InformationalError.new(opts[:pretty_message])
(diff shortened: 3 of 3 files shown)


updater/lib/dependabot/setup.rb
```diff
@@ -50,6 +51,7 @@
 end
 
 Dependabot::OpenTelemetry.configure
+Dependabot::Sorbet::Runtime.silently_report_errors!
```
jurre, line 54: `dependabot/setup` is only required in prod, so this wouldn't affect our test setup

updater/lib/dependabot/setup.rb
```diff
@@ -8,6 +8,7 @@
 require "dependabot/logger/formats"
 require "dependabot/opentelemetry"
 require "dependabot/sentry"
+require "dependabot/service"
```
landongrindheim, line 11: 🔎 I'm curious why this was needed. Was it missing before?

landongrindheim (approved):
❤️🤫

updater/lib/dependabot/sorbet/runtime.rb
```diff
@@ -0,0 +1,29 @@
+require "sorbet-runtime"
+
+module Dependabot
+  module Sorbet
+    module Runtime
+      class InformationalError < StandardError; end
+
+      def self.silently_report_errors!
+        T::Configuration.call_validation_error_handler = lambda do |sig, opts|
+          error = InformationalError.new(build_message(sig, opts))
```
JamieMagee, line 10: This is the default handler: https://github.com/sorbet/sorbet/blob/799c78d6d04231f10062c773a58f4bbc8c6bb7b3/gems/sorbet-runtime/lib/types/configuration.rb#L295-L297

Looks like we can just extract `:pretty_message` and use that instead of building our own?

updater/lib/dependabot/sorbet/runtime.rb
```diff
@@ -0,0 +1,29 @@
+require "sorbet-runtime"
```
JamieMagee, line 1: You might also need:

```suggestion
# typed: true
# frozen_string_literal: true

require "sentry-ruby"
require "sorbet-runtime"
```

updater/lib/dependabot/setup.rb
JamieMagee, line 11:
```suggestion
require "dependabot/service"
require "dependabot/sorbet/runtime"
```

updater/lib/dependabot/sorbet/runtime.rb
jurre: Since we already require Sentry before this file is required I don't think we need to require it here again?

updater/lib/dependabot/setup.rb
jurre: Oh no, not needed just an oversight from something else I tried, let me clean it up :+1:

updater/lib/dependabot/sorbet/runtime.rb
```diff
@@ -0,0 +1,21 @@
...
+  module Sorbet
+    module Runtime
+      class InformationalError < StandardError; end
+
+      def self.silently_report_errors!
+        T::Configuration.call_validation_error_handler = lambda do |sig, opts|
+          error = InformationalError.new(opts[:pretty_message])
+          error.set_backtrace(caller.dup)
+
+          Sentry.capture_exception(error)
```
JamieMagee, line 16: Sorbet is getting confused between `Dependabot::Sentry` and `Sentry`.

```suggestion
          ::Sentry.capture_exception(error)
```

updater/lib/dependabot/sorbet/runtime.rb
```diff
@@ -0,0 +1,21 @@
...
+# frozen_string_literal: true
+#
+require "sorbet-runtime"
+
+module Dependabot
+  module Sorbet
+    module Runtime
+      class InformationalError < StandardError; end
+
+      def self.silently_report_errors!
```
JamieMagee, line 11:
```suggestion
# typed: strict
# frozen_string_literal: true

require "sorbet-runtime"

module Dependabot
  module Sorbet
    module Runtime
      class InformationalError < StandardError; end
      extend T::Sig

      sig { void }
      def self.silently_report_errors!
```

JamieMagee (approved):

abdulapopoola:
Any blockers with merging this @jurre and @JamieMagee ?

JamieMagee:
I think this is ready to go

jurre:
No blockers other than time available in the day

README.md
```diff
@@ -1,4 +1,5 @@
 <h1 align="center">
+
```
davidstosik, line 2: Hello @jurre! 👋🏻 
I'm coming from this repo's README page to let you know the addition of this empty line in the README file broke how the image displays.
You'll find more details in my PR that attempts to fix that: https://github.com/dependabot/dependabot-core/pull/9095.

PR #8999, opened 2024-02-06
add types to DependencyChange

jakecoffman:


updater/lib/dependabot/api_client.rb
```diff
@@ -290,8 +290,6 @@ def create_pull_request_data(dependency_change, base_commit_sha)
         "base-commit-sha": base_commit_sha
       }.merge(dependency_group_hash(dependency_change))
 
-      return data unless dependency_change.pr_message
```
jakecoffman, line 293: One of the more interesting things Sorbet caught was that pr_message is never nil because it's actually a class instance that always can be created.

updater/spec/dependabot/dependency_change_spec.rb
```diff
@@ -86,7 +86,14 @@
     end
 
     let(:message_builder_mock) do
-      instance_double(Dependabot::PullRequestCreator::MessageBuilder, message: "Hello World!")
```
jakecoffman, line 89: This test was being setup wrong, it returns a class instance not a String.

abdulapopoola (approved):

updater/lib/dependabot/dependency_change.rb
JamieMagee, line 1: Do you need to `require "sorbet-runtime"`?

updater/lib/dependabot/service.rb
```diff
@@ -19,7 +19,7 @@ class Service
     extend T::Sig
     extend Forwardable
 
-    sig { returns(T::Array[T::Array[String]]) }
+    sig { returns(T::Array[T.untyped]) }
```
JamieMagee, line 22: Is this a client specific response? Too difficult to type without a common interface?

pavera (approved):

updater/lib/dependabot/dependency_change.rb
jakecoffman: I'm not sure, do I? I can try adding it to see if anything changes.

updater/lib/dependabot/service.rb
jakecoffman: `pull_requests` is something used only in this file and it's tricky to type, so I think we can circle back eventually.

jakecoffman:
One of the smoke tests might have caught a regression, I changed them since the workflow last run, so I'll look into that.

Issue #8910, opened 2024-01-26
Keep `username` and `password` separate from `token` for NuGet feeds


brettfo:
From #8887.

Artifactory NuGet feeds need the `username` and `password` fields kept separate and not pre-emptively joined into the `token` field for authorization to work properly.

Relevant files that will need to be changed:
- [`repository_finder.rb`](https://github.com/dependabot/dependabot-core/blob/51e0f78c1fbfacfb748a27c57c95a17082983785/nuget/lib/dependabot/nuget/update_checker/repository_finder.rb)
- [`metadata_finder.rb`](https://github.com/dependabot/dependabot-core/blob/51e0f78c1fbfacfb748a27c57c95a17082983785/nuget/lib/dependabot/nuget/metadata_finder.rb)
- [`nuget_config_credential_helpers.rb`](https://github.com/dependabot/dependabot-core/blob/51e0f78c1fbfacfb748a27c57c95a17082983785/nuget/lib/dependabot/nuget/nuget_config_credential_helpers.rb)

PR #9004, opened 2024-02-07
handle both `token` and `username`/`password` for NuGet feeds

### LINKED ISSUE #8887: Artifactory private nuget package source authentication

### Package ecosystem

nuget

### Package manager version

_No response_

### Language version

_No response_

_No response_

### dependabot.yml content

```
registries:
  nuget-public:
    type: nuget-feed

  my-artifactory:
    type: nuget-feed
    url: https://xyz.jfrog.io/artifactory/api/nuget/v3/nuget-local
    username: ${{secrets.ART_USER}}
    password: ${{secrets.ART_PASS}}
```

### Updated dependency

_No response_

artifactory private nuget package source authentication is not working. I think part of the reason could be that artifactory requires username/password, but it looks like `nuget_config_credential_helpers.rb` uses hardcoded user `user` and `token` instead of password:

image

is there a way to support username/password, so that artifactory authentication starts working? Otherwise none of the dependencies get updated, event the ones coming from public nuget feed.

By the way public package updates were working fine until something changed in the nuget implementation. Now none of the packages get updates in projects that also use private packages.

_No response_

_No response_

_No response_

brettfo:
Private NuGet feed information can be specified in the `dependabot.yml` file either with a `token` field, or with both `username` and `password`.  Previously we only honored the `token` field and even then, this wasn't properly handled when we patched the local `NuGet.Config` file; we hard-coded the username to `user` and left the token intact for the password.

This PR fixes the issue in 2 ways:

1. When reading feed data from the `credentials` object, we first use the `token` field if present, otherwise we construct a token from `<username>:<password>`, and finally fall back to `nil` (as before).  This value is then stored in the `token` property of the repository details.
2. When patching the local `NuGet.Config`, instead of simply pulling out the `token` field and using it as a password, we first try to split back out the `username` and `password` parts and apply those as appropriate.  If we couldn't split it apart, we then treat it like a normal token as before with a hard-coded username of `user`.

There were 2 places where this token creation occurred in the past so I introduced `credential_helpers.rb` to consolidate this into one location.

Aside from unit tests, I also ran a manual test against a private GitHub NuGet feed with the following values:

``` yaml
registries:
  some-private-feed:
    type: nuget-feed
    url: https://nuget.example.com/index.json
    token: "myuser:mypass"
```

and

``` yaml
registries:
  some-private-feed:
    type: nuget-feed
    url: https://nuget.example.com/index.json
    username: myuser
    password: mypass
```

Fixes #8887 and #8910.

### DIFF
--- a/nuget/lib/dependabot/nuget/credential_helpers.rb
+++ b/nuget/lib/dependabot/nuget/credential_helpers.rb
@@ -1,0 +1,19 @@
+# typed: true
+# frozen_string_literal: true
+
+module Dependabot
+  module Nuget
+    module CredentialHelpers
+      def self.get_token_from_credentials(credentials)
+        token = credentials["token"]
+        username = credentials["username"]
+        password = credentials["password"]
+
+        return token if token
+        return "#{username}:#{password}" if username && password
+
+        nil
+      end
+    end
+  end
+end
--- a/nuget/lib/dependabot/nuget/credential_helpers.rb
+++ b/nuget/lib/dependabot/nuget/credential_helpers.rb
@@ -4,10 +4,10 @@
 module Dependabot
   module Nuget
     module CredentialHelpers
-      def self.get_token_from_credentials(credentials)
-        token = credentials["token"]
-        username = credentials["username"]
-        password = credentials["password"]
+      def self.get_token_from_credential(credential)
+        token = credential["token"]
+        username = credential["username"]
+        password = credential["password"]
 
         return token if token
         return "#{username}:#{password}" if username && password
--- a/nuget/lib/dependabot/nuget/metadata_finder.rb
+++ b/nuget/lib/dependabot/nuget/metadata_finder.rb
@@ -145,12 +145,12 @@ module Dependabot
                            .find { |r| r.fetch(:source) }&.fetch(:source)
         url = source&.fetch(:url, nil) || source&.fetch("url")
 
-        matching_credentials = credentials
-                               .select { |cred| cred["type"] == "nuget_feed" }
-                               .find { |cred| cred["url"] == url }
-        return {} unless matching_credentials
+        matching_credential = credentials
+                              .select { |cred| cred["type"] == "nuget_feed" }
+                              .find { |cred| cred["url"] == url }
+        return {} unless matching_credential
 
-        token = CredentialHelpers.get_token_from_credentials(matching_credentials)
+        token = CredentialHelpers.get_token_from_credential(matching_credential)
         return {} unless token
 
         if token.include?(":")
--- a/nuget/lib/dependabot/nuget/metadata_finder.rb
+++ b/nuget/lib/dependabot/nuget/metadata_finder.rb
@@ -6,6 +6,7 @@ require "sorbet-runtime"
 require "dependabot/metadata_finders"
 require "dependabot/metadata_finders/base"
 require "dependabot/registry_client"
+require "dependabot/nuget/credential_helpers"
 
 module Dependabot
   module Nuget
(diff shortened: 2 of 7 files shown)


nuget/lib/dependabot/nuget/metadata_finder.rb
```diff
@@ -144,11 +145,12 @@ def auth_header
                            .find { |r| r.fetch(:source) }&.fetch(:source)
         url = source&.fetch(:url, nil) || source&.fetch("url")
 
-        token = credentials
-                .select { |cred| cred["type"] == "nuget_feed" }
-                .find { |cred| cred["url"] == url }
-                &.fetch("token", nil)
+        matching_credentials = credentials
```
JoeRobich, line 148: nit: name this `matching_credential` because I was expecting an array.

nuget/lib/dependabot/nuget/credential_helpers.rb
```diff
@@ -0,0 +1,19 @@
+# typed: true
+# frozen_string_literal: true
+
+module Dependabot
+  module Nuget
+    module CredentialHelpers
+      def self.get_token_from_credentials(credentials)
```
JoeRobich, line 7: nit: use singular `credential` in the method name and as the parameter.

nuget/spec/dependabot/nuget/nuget_config_credential_helpers_spec.rb
```diff
@@ -32,8 +32,13 @@
     context "with non-empty credential set" do
       let(:credentials) do
         [
-          { "type" => "nuget_feed", "url" => "https://private.nuget.example.com/index.json",
+          # private feed with a token
+          { "type" => "nuget_feed", "url" => "https://private1.nuget.example.com/index.json",
             "token" => "secret_token" },
+          # private feed with a username and password
+          { "type" => "nuget_feed", "url" => "https://private2.nuget.example.com/index.json", "username" => "my",
+            "password" => "passw0rd" },
```
JoeRobich, line 40: Any value in a test with all three set that ensures we take the `token`?

JoeRobich (approved):
Looks good!

nuget/lib/dependabot/nuget/metadata_finder.rb
brettfo: Done.

nuget/lib/dependabot/nuget/credential_helpers.rb
brettfo: Done.

nuget/spec/dependabot/nuget/nuget_config_credential_helpers_spec.rb
brettfo: Done.

brettfo:
N.b., failing CodeQL check is not a concern.  The Docker image running the update already has the relevant feed tokens and passwords in memory and we _have_ to write them out where we do, otherwise the C# NuGet updater won't be able to access it.  If a user's codebase _was_ read from that file during an update operation, they'd only be able to see their own tokens, not anybody else's.

abdulapopoola (approved):

jakecoffman:
Credentials are not given to the Updater when running on GitHub or the Dependabot CLI, they're passed to the Proxy and then a `credentials-metadata` is given to the Updater. 

You can see that being done in the CLI [here](https://github.com/dependabot/cli/blob/11b2fc430be318aad7c1f43e28757d1f39603d6d/cmd/dependabot/internal/cmd/update.go#L379-L382).

Confusingly in Core we rename `credentials-metadata` to `credentials` early on: https://github.com/dependabot/dependabot-core/blob/293b93dc6eba205a0ed821219e93e16c41cec6f2/updater/lib/dependabot/job.rb#L116

This may have been done in the past because [dependabot-script](https://github.com/dependabot/dependabot-script) doesn't use the Proxy and really does inject credentials which is why in some places you can still see password and tokens being used.

But the current approach is to keep the credentials away from the Updater since some of them can execute arbitrary code, but it also normalizes how credentials are handled across the board.

So unfortunately I don't think this PR will fix anything!

abdulapopoola:
@brettfo should this be abandoned now?

brettfo:
@abdulapopoola Yes, thank you, closing now.

PR #9017, opened 2024-02-07
only consider a package a development dependency if it doesn't have any other regular dependencies

### LINKED ISSUE #9016: `.nuspec` `developmentDependency` element not fully reliable
A package's `.nuspec` file may contain an element `<developmentDependency>true</developmentDependency>` which normally indicates that it's a package used by the compiler and can always be upgraded, however some packages specify that as well as regular `targetFramework` dependencies, which means these packages _might_ be able to be upgraded.  E.g., `Microsoft.EntityFrameworkCore.Design.nuspec` is as follows:

``` xml
<package xmlns="http://schemas.microsoft.com/packaging/2013/05/nuspec.xsd">
  <metadata>
    <id>Microsoft.EntityFrameworkCore.Design</id>
    <version>8.0.1</version>
    <authors>Microsoft</authors>
    <developmentDependency>true</developmentDependency>
    <requireLicenseAcceptance>true</requireLicenseAcceptance>
    <license type="expression">MIT</license>
    <licenseUrl>https://licenses.nuget.org/MIT</licenseUrl>
    <icon>Icon.png</icon>
    <projectUrl>https://docs.microsoft.com/ef/core/</projectUrl>
    <description>Shared design-time components for Entity Framework Core tools.</description>
    <copyright>© Microsoft Corporation. All rights reserved.</copyright>
    <tags>Entity Framework Core entity-framework-core EF Data O/RM EntityFramework
      EntityFrameworkCore EFCore</tags>
    <serviceable>true</serviceable>
    <repository type="git" url="https://github.com/dotnet/efcore"
      commit="423b1a0d38036efcefdbf1dfb80955449eae1238" />
    <dependencies>
      <group targetFramework="net8.0">
        <dependency id="Microsoft.EntityFrameworkCore.Relational" version="8.0.1"
          exclude="Build" />
        <dependency id="Humanizer.Core" version="2.14.1" exclude="Build,Analyzers" />
        <dependency id="Microsoft.CodeAnalysis.CSharp.Workspaces" version="4.5.0"
          exclude="Build,Analyzers" />
        <dependency id="Microsoft.Extensions.DependencyModel" version="8.0.0"
          exclude="Build,Analyzers" />
        <dependency id="Mono.TextTemplating" version="2.2.1" exclude="Build,Analyzers" />
      </group>
    </dependencies>
  </metadata>
</package>
```

N.b., both `<developmentDependency>true</developmentDependency>` _and_ `<group targetFramework="net8.0">`.  This indicates that the normal TFM checks still need to be performed.

`compatibility_checker.rb` needs to be updated to allow this.

brettfo:
A package's `.nuspec` file can self-identify as a development dependency which normally means it can always be upgraded, but some packages exist that identify as development dependencies but also have regular `targetFramework`-specific dependencies.  The fix is to only allow development dependency upgrades if they're purely a development dependency and don't contain any other dependencies _unless_ the specified target framework is still compatible.

Fixes #9016

### DIFF
--- a/nuget/spec/dependabot/nuget/update_checker/compatibility_checker_spec.rb
+++ b/nuget/spec/dependabot/nuget/update_checker/compatibility_checker_spec.rb
@@ -74,6 +74,96 @@ RSpec.describe Dependabot::Nuget::CompatibilityChecker do
   context "#compatible?" do
     subject(:compatible) { checker.compatible?(version) }
 
+    before do
+      stub_request(:get, "https://api.nuget.org/v3/registration5-gz-semver2/microsoft.appcenter.crashes/index.json")
+        .to_return(
+          status: 200,
+          body: {
+            items: [
+              items: [
+                {
+                  catalogEntry: {
+                    listed: true,
+                    version: "5.0.2"
+                  }
+                },
+                {
+                  catalogEntry: {
+                    listed: true,
+                    version: "5.0.3"
+                  }
+                }
+              ]
+            ]
+          }.to_json
+        )
+    end
+
+    context "when the `.nuspec` reports itself as a development dependency, but still has regular dependencies" do
+      let(:csproj_body) do
+        <<~XML
+          <Project Sdk="Microsoft.NET.Sdk">
+            <PropertyGroup>
+              <TargetFramework>net6.0</TargetFramework>
+            </PropertyGroup>
+            <ItemGroup>
+              <PackageReference Include="Microsoft.AppCenter.Crashes" Version="5.0.2" />
+            </ItemGroup>
+          </Project>
+        XML
+      end
+
+      before do
+        nuspec502 =
+          <<~XML
+            <package xmlns="http://schemas.microsoft.com/packaging/2013/05/nuspec.xsd">
+              <metadata>
+                <id>Microsoft.AppCenter.Crashes</id>
+                <version>5.0.2</version>
+                <developmentDependency>true</developmentDependency>
+                <dependencies>
+                  <group targetFramework="net6.0">
+                  </group>
+                </dependencies>
+              </metadata>
+            </package>
+          XML
+        nuspec503 = nuspec502.gsub("5.0.2", "5.0.3")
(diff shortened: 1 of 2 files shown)


nuget/spec/dependabot/nuget/update_checker/compatibility_checker_spec.rb
```diff
@@ -74,41 +74,109 @@
   context "#compatible?" do
     subject(:compatible) { checker.compatible?(version) }
 
-    context "when the `.nuspec` has groups without a `targetFramework` attribute" do
-      let(:version) { "5.0.3" }
+    before do
+      stub_request(:get, "https://api.nuget.org/v3/registration5-gz-semver2/microsoft.appcenter.crashes/index.json")
```
brettfo, line 78: The test diff looks worse than it really is.  This section got moved up a level because it's common to both `context`s below it, otherwise it's a new test case.

nuget/spec/dependabot/nuget/update_checker/compatibility_checker_spec.rb
```diff
@@ -74,41 +74,109 @@
...
+            status: 200,
+            body: nuspec601
+          )
+      end
+
+      context "with a targetFramework compatible version" do
+        let(:version) { "5.0.3" }
+
+        it "returns the correct data" do
+          expect(compatible).to be_truthy
```
brettfo, line 154: Verifies that we can update from `5.0.2` to `5.0.3`.  Both self-identify as development dependencies and the target frameworks match.

nuget/spec/dependabot/nuget/update_checker/compatibility_checker_spec.rb
```diff
@@ -74,41 +74,109 @@
...
+        end
+      end
+
+      context "with a targetFramework non-compatible version" do
+        let(:version) { "6.0.1" }
+
+        it "returns the correct data" do
+          expect(compatible).to be_falsey
```
brettfo, line 162: Verifies that we _can't_ update from `5.0.2` to `6.0.1`.  Both self-identify as development dependencies, but the target frameworks are incompatible.

JoeRobich (approved):

PR #9029, opened 2024-02-11
Strict type `Dependabot::MetadataFinders::Base::ChangelogFinder`

JamieMagee:
This has a lot of:
- `T.unsafe` because the source platform clients (GitHub, Azure DevOps, etc.) all use `method_missing` to inject credentials or handle retries. Rewriting them is going to be difficult 😖
- `OpenStruct` and `Sawyer::Resource` return types, again because of source platform clients
- `T.untyped` response types because of (you guessed it!) source platform clients

### DIFF
--- a/common/lib/dependabot/metadata_finders/base/changelog_finder.rb
+++ b/common/lib/dependabot/metadata_finders/base/changelog_finder.rb
@@ -1,4 +1,4 @@
-# typed: true
+# typed: strict
 # frozen_string_literal: true
 
 require "excon"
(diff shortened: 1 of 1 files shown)


common/lib/dependabot/metadata_finders/base/changelog_finder.rb
```diff
@@ -1,4 +1,4 @@
-# typed: true
+# typed: strict
```
abdulapopoola, line 1: nice!

abdulapopoola (approved):

PR #9039, opened 2024-02-12
test more of the security error scenarios

jakecoffman:
Improving test coverage of the error conditions unique to security updates.


silent/tests/testdata/su-err-pr-exists-security.txt
```diff
@@ -0,0 +1,41 @@
...
+    directory: "/"
+    provider: example
+    hostname: example.com
+    api-endpoint: https://example.com/api/v3
+    repo: dependabot/smoke-tests
+  security-advisories:
+    - dependency-name: dependency-a
+      affected-versions:
+        - <= 1.2.3
+      patched-versions: []
```
bdragon, line 36: I don't think it's material to the test working, so feel free to ignore: I think realistically `patched-versions` would include `1.2.4`.

bdragon (approved):
Great idea to add these!

silent/tests/testdata/su-err-pr-exists-security.txt
jakecoffman: I've never seen that field set to a value!

abdulapopoola: Interesting; do you know why that is the case?

jakecoffman: I don't, would need to dig into that.

PR #9040, opened 2024-02-12
support group configs specifically for security updates or version updates

jakecoffman:
This adds support for property `applies-to` which allows developers to specify that the group configuration should apply to only version updates, or only security updates.

For backward compatibility, if `applies-to` is not set then it applies only to version updates.

### DIFF
--- a/common/lib/dependabot/dependency_group.rb
+++ b/common/lib/dependabot/dependency_group.rb
@@ -28,12 +28,12 @@ module Dependabot
     sig do
       params(
         name: String,
-        applies_to: T.nilable(String),
-        rules: T::Hash[String, T.untyped]
+        rules: T::Hash[String, T.untyped],
+        applies_to: T.nilable(String)
       )
         .void
     end
-    def initialize(name:, applies_to:, rules:)
+    def initialize(name:, rules:, applies_to: "version-updates")
       @name = name
       # For backwards compatibility, if no applies_to is provided, default to "version-updates"
       @applies_to = T.let(applies_to || "version-updates", String)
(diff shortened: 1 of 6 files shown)


common/lib/dependabot/dependency_group.rb
```diff
@@ -22,15 +22,21 @@ class DependencyGroup
...
+        applies_to: T.nilable(String),
         rules: T::Hash[String, T.untyped]
       )
         .void
     end
-    def initialize(name:, rules:)
+    def initialize(name:, applies_to:, rules:)
       @name = name
+      # For backwards compatibility, if no applies_to is provided, default to "version-updates"
+      @applies_to = T.let(applies_to || "version-updates", String)
```
landongrindheim, line 39: 👍

landongrindheim (approved):

common/lib/dependabot/dependency_group.rb
abdulapopoola: Consider using as a default parameter in the function signature itself?

abdulapopoola (approved):

common/lib/dependabot/dependency_group.rb
jakecoffman: I do need to set a default, but I also have to have this here because in Ruby the default parameter only applies when the value isn't set, so a `nil` value will still get through:

```ruby
irb(main):001* def f(v: "default")
irb(main):002*   v
irb(main):003> end
=> :f
irb(main):004> f()
=> "default"
irb(main):005> f(v: nil)
=> nil
irb(main):006> f(v: "hello")
=> "hello"
```

abdulapopoola: minor: could Sorbet help with this?

jurre: I think Sorbet is doing all it can here, there just are scenario's where we will pass `nil`, for example in GHES where the API has not been updated, we will send job definitions without an `applies-to` value and this is the entry point where we normalize the data to make it backwards compatible

jurre (approved):

PR #9041, opened 2024-02-12
use boolean for clearer signal of a grouped update

jakecoffman:
In https://github.com/dependabot/dependabot-core/pull/8963 I am running into a problem where using `directory` and `directories` is not a clear enough signal that this is supposed to be a grouped update, so switching to the `grouped-update` boolean.

### DIFF
--- a/updater/lib/dependabot/job.rb
+++ b/updater/lib/dependabot/job.rb
@@ -43,6 +43,7 @@ module Dependabot
       update_subdependencies
       updating_a_pull_request
       vendor_dependencies
+      grouped_update
       dependency_groups
       dependency_group_to_refresh
       repo_private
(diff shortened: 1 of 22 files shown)


jakecoffman:
Smoke tests should be fixed by this PR: https://github.com/dependabot/smoke-tests/pull/184

updater/lib/dependabot/job.rb
```diff
@@ -168,6 +169,7 @@ def initialize(attributes) # rubocop:disable Metrics/AbcSize
       # We will need to do a pass updating the CLI and smoke tests before this is possible,
       # so let's consider it optional for now. If we get a nil value, let's force it to be
       # an array.
```
jurre, line 171: I think this comment applies to the `@dependency_groups` variable, so let's move things around so it stays close?

jurre (approved):

jakecoffman:
Now that we did https://github.com/dependabot/dependabot-core/pull/9040 I'm inclined to not do this and continue to let the updater code figure out if it's grouped based on the presence of a group that applies in the job definition.

The main factor for this before was that the API was sending groups with security updates, so it wasn't possible to decide if the updater was supposed to use it. Now that it is explicit with the `applies-to` we don't need this boolean.

PR #9043, opened 2024-02-13
Fix docker variant with matching digests not updating correctly

brbayes-msft:
There is an issue where the text replacement when a duplicate digest was present for dockerfile did not look for the dependency name or tag, only the old digest.  This updates the code to look for all of the components based on what was present in the original file.


honeyankit (approved):
LGTM

PR #9064, opened 2024-02-16
Surface out of disk/memory error message for easier visibility

honeyankit:
## Context

These currently end up in the `unknown_error category` so this provides easier visibility to the user.

- Added out of disk error message to the go module updater.
- Added out of disk/memory error message to the yarn file parser.

### DIFF
--- a/common/lib/dependabot/workspace/git.rb
+++ b/common/lib/dependabot/workspace/git.rb
@@ -162,6 +162,10 @@ module Dependabot
       sig { params(args: String, kwargs: T.any(T::Boolean, String)).returns(String) }
       def run_shell_command(*args, **kwargs)
         Dir.chdir(path) { T.unsafe(SharedHelpers).run_shell_command(*args, **kwargs) }
+      rescue Dependabot::SharedHelpers::HelperSubprocessFailed => e
+        raise Dependabot::OutOfDisk, e.message if e.message.end_with?("No space left on device")
+        raise Dependabot::OutOfDisk, e.message if e.message.end_with?("Out of diskspace")
+        raise Dependabot::OutOfMemory, e.message if e.message.end_with?("MemoryError")
       end
 
       sig { params(message: String).void }
--- a/common/lib/dependabot/workspace/git.rb
+++ b/common/lib/dependabot/workspace/git.rb
@@ -162,10 +162,6 @@ module Dependabot
       sig { params(args: String, kwargs: T.any(T::Boolean, String)).returns(String) }
       def run_shell_command(*args, **kwargs)
         Dir.chdir(path) { T.unsafe(SharedHelpers).run_shell_command(*args, **kwargs) }
-      rescue Dependabot::SharedHelpers::HelperSubprocessFailed => e
-        raise Dependabot::OutOfDisk, e.message if e.message.end_with?("No space left on device")
-        raise Dependabot::OutOfDisk, e.message if e.message.end_with?("Out of diskspace")
-        raise Dependabot::OutOfMemory, e.message if e.message.end_with?("MemoryError")
       end
 
       sig { params(message: String).void }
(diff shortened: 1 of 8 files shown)


jakecoffman:
I'd bet the error message `write fatal: sha1 file '/home/dependabot/dependabot-updater/repo/.git/index.lock' write error.` is not coming from Go and Yarn CLIs, but our workspace/git.rb operations.

honeyankit:
> I'd bet the error message `write fatal: sha1 file '/home/dependabot/dependabot-updater/repo/.git/index.lock' write error.` is not coming from Go and Yarn CLIs, but our workspace/git.rb operations.

There are two errors in the sentry:
https://github.sentry.io/discover/deltaforce:b9cc2996e6104f6dbf95937df5aa4f68/?field=title&field=event.type&field=project&field=user.display&field=timestamp&field=replayId&homepage=true&name=All+Events&project=1451818&query=disk&sort=-timestamp&statsPeriod=90d&yAxis=count%28%29

One is from `workspace/git.rb` and another one from `No space left on device @ io_write - yarn.lock`.

abdulapopoola (approved):

jakecoffman:
`write fatal: sha1 file '/home/dependabot/dependabot-updater/repo/.git/index.lock' write error` appears to be coming from `workspace/git.rb` so I would catch it there, it won't be coming from Go or Yarn code that won't fix the issue.

I agree that `No space left on device @ io_write - yarn.lock` is coming from the Yarn file parser, that looks good 👍

abdulapopoola:
Tagging @bdragon since he has context on the `git.rb` file

abdulapopoola:
Took a peek - it appears the source of the failure is the [reset](https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/workspace/git.rb#L151) command

honeyankit:
> Took a [peek](https://github.sentry.io/discover/deltaforce:b9cc2996e6104f6dbf95937df5aa4f68/?field=title&field=event.type&field=project&field=user.display&field=timestamp&field=replayId&homepage=true&name=All+Events&project=1451818&query=disk&sort=-timestamp&statsPeriod=90d&yAxis=count%28%29) - it appears the source of the failure is the [reset](https://github.com/dependabot/dependabot-core/blob/main/common/lib/dependabot/workspace/git.rb#L151) command

Yes, but I want to add exception to the common place where `run_shell_command` is defined. In the git.rb, the `run_shell_command` is called many git operations and adding it there will make more sense. I have already pushed the change, just adding test and we are good!

abdulapopoola (approved):

common/lib/dependabot/workspace/git.rb
```diff
@@ -162,6 +162,10 @@ def clean
       sig { params(args: String, kwargs: T.any(T::Boolean, String)).returns(String) }
       def run_shell_command(*args, **kwargs)
         Dir.chdir(path) { T.unsafe(SharedHelpers).run_shell_command(*args, **kwargs) }
+      rescue Dependabot::SharedHelpers::HelperSubprocessFailed => e
+        raise Dependabot::OutOfDisk, e.message if e.message.end_with?("No space left on device")
+        raise Dependabot::OutOfDisk, e.message if e.message.end_with?("Out of diskspace")
+        raise Dependabot::OutOfMemory, e.message if e.message.end_with?("MemoryError")
```
bdragon, line 168: This is a good idea but to me this feels like the wrong layer to add it (sounds like you might already be working on a change to this anyways). We already do something like this to check for out of memory errors in [`SharedHelpers.run_helper_subprocess`](https://github.com/dependabot/dependabot-core/blob/dfe50fe71cab4523af0aa1945e483470f639ca15/common/lib/dependabot/shared_helpers.rb#L183), so it seems like maybe this should live in [`SharedHelpers.run_shell_command`](https://github.com/dependabot/dependabot-core/blob/dfe50fe71cab4523af0aa1945e483470f639ca15/common/lib/dependabot/shared_helpers.rb#L401).

honeyankit: I think what you saying is correct. Let me change it. Thanks.

bdragon (approved):

PR #9067, opened 2024-02-16
Fix hardcoded amd64 arch for git-shim

andrcuns:
Adds `TARGETARCH` argument similar to updater images to prevent having `amd64` binary of `git-shim` when building arm image

### DIFF
--- a/Dockerfile.updater-core
+++ b/Dockerfile.updater-core
@@ -1,5 +1,7 @@
 FROM docker.io/library/ubuntu:22.04
 
+ARG TARGETARCH
+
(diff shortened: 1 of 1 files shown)


jakecoffman:
Huh, that's interesting as building a local ARM image has been working for me. I wonder how!

jakecoffman (approved):

andrcuns:
>Huh, that's interesting as building a local ARM image has been working for me. I wonder how!

It might be something to do with docker-desktop on macos and it's virtualisation + rosetta. The issue that caused me to see this happened during runtime when executing on linux arm environment.

PR #9072, opened 2024-02-17
Fix regression in docker credential helper caused by credential refactor

noorul:


jakecoffman:
Pulled changed into #9091, thanks!

PR #9092, opened 2024-02-21
nuget updater command is already space-enabled; allow unsafe execution

### LINKED ISSUE #8633: NuGet update failes when project path contains spaces.

brettfo:
The NuGet updater is invoked with already-escaped command line options, so we can simply call it with `allow_unsafe_shell_command: true`, otherwise the arguments are double escaped.

Adding that parameter was the real fix, but I also refactored the code a bit to make testing the command line generation easier.

N.b., the helper function `run_shell_command` only takes a string which it tries to escape and that's where the problem arose; we have to pre-escape our strings to allow for a path with spaces in it, etc., and then tell `run_shell_command` that we've already escaped everything.

Fixes #8633

### DIFF
--- a/nuget/lib/dependabot/nuget/native_helpers.rb
+++ b/nuget/lib/dependabot/nuget/native_helpers.rb
@@ -113,6 +113,7 @@ module Dependabot
         (command, fingerprint) = get_nuget_updater_tool_command(repo_root: repo_root, proj_path: proj_path,
                                                                 dependency: dependency, is_transitive: is_transitive)
 
+        command = T.must(command)
         puts "running NuGet updater:\n" + command
 
         NuGetConfigCredentialHelpers.patch_nuget_config_for_action(credentials) do
--- a/nuget/lib/dependabot/nuget/native_helpers.rb
+++ b/nuget/lib/dependabot/nuget/native_helpers.rb
@@ -45,7 +45,7 @@ module Dependabot
 
         puts "running NuGet updater:\n" + command
 
-        output = SharedHelpers.run_shell_command(command, fingerprint: fingerprint)
+        output = SharedHelpers.run_shell_command(command, allow_unsafe_shell_command: true, fingerprint: fingerprint)
         puts output
 
         # Exit code == 0 means that all project frameworks are compatible
@@ -55,17 +55,11 @@ module Dependabot
         false
       end
 
-      # rubocop:disable Metrics/MethodLength
       sig do
-        params(
-          repo_root: String,
-          proj_path: String,
-          dependency: Dependency,
-          is_transitive: T::Boolean,
-          credentials: T::Array[T.untyped]
-        ).void
+        params(repo_root: String, proj_path: String, dependency: Dependency,
+               is_transitive: T::Boolean).returns(T::Array[String])
       end
-      def self.run_nuget_updater_tool(repo_root:, proj_path:, dependency:, is_transitive:, credentials:)
+      def self.get_nuget_updater_tool_command(repo_root:, proj_path:, dependency:, is_transitive:)
         exe_path = File.join(native_helpers_root, "NuGetUpdater", "NuGetUpdater.Cli")
         command_parts = [
           exe_path,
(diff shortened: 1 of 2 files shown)


nuget/lib/dependabot/nuget/native_helpers.rb
```diff
@@ -55,17 +55,11 @@ def self.run_nuget_framework_check(project_tfms, package_tfms)
...
       sig do
-        params(
-          repo_root: String,
-          proj_path: String,
-          dependency: Dependency,
-          is_transitive: T::Boolean,
-          credentials: T::Array[T.untyped]
-        ).void
+        params(repo_root: String, proj_path: String, dependency: Dependency,
+               is_transitive: T::Boolean).returns(T::Array[String])
```
ryanbrandenburg, line 60: Should this be the ruby equivilant of a tuple (whatever that is) instead of an array since it will only ever have 2 elements in defined positions?

nuget/lib/dependabot/nuget/native_helpers.rb
```diff
@@ -103,14 +97,30 @@ def self.run_nuget_updater_tool(repo_root:, proj_path:, dependency:, is_transiti
...
+        command = T.must(command)
         puts "running NuGet updater:\n" + command
 
         NuGetConfigCredentialHelpers.patch_nuget_config_for_action(credentials) do
-          output = SharedHelpers.run_shell_command(command, fingerprint: fingerprint)
+          output = SharedHelpers.run_shell_command(command, allow_unsafe_shell_command: true, fingerprint: fingerprint)
           puts output
         end
       end
-      # rubocop:enable Metrics/MethodLength
```
ryanbrandenburg, line 113: Yay for this going away!

ryanbrandenburg (approved):

nuget/lib/dependabot/nuget/native_helpers.rb
JamieMagee: Tuples are still experimental in Sorbet, but you can definitely try them: https://sorbet.org/docs/tuples

brettfo: Ruby doesn't have tuples; it's recommended (at least in the documentation I found) to do it this way.

nuget/lib/dependabot/nuget/native_helpers.rb
```diff
@@ -103,14 +97,30 @@ def self.run_nuget_updater_tool(repo_root:, proj_path:, dependency:, is_transiti
...
+        [command, fingerprint]
+      end
+
+      sig do
+        params(
+          repo_root: String,
+          proj_path: String,
+          dependency: Dependency,
+          is_transitive: T::Boolean,
+          credentials: T::Array[T.untyped]
```
JamieMagee, line 109: Can you double check if this typing works now?

```suggestion
          credentials: T::Array[Dependabot::Credential]
```

I think whenever I added this initially, types were still a little up in the air.

nuget/lib/dependabot/nuget/native_helpers.rb
```diff
@@ -103,14 +97,30 @@ def self.run_nuget_updater_tool(repo_root:, proj_path:, dependency:, is_transiti
...
+        ).void
+      end
+      def self.run_nuget_updater_tool(repo_root:, proj_path:, dependency:, is_transitive:, credentials:)
+        (command, fingerprint) = get_nuget_updater_tool_command(repo_root: repo_root, proj_path: proj_path,
+                                                                dependency: dependency, is_transitive: is_transitive)
+
+        command = T.must(command)
```
JamieMagee, line 116: nit: you can directly put this as part of the method call below. But it's a stylistic choice, so feel free to ignore this.

nuget/lib/dependabot/nuget/native_helpers.rb
brettfo: @JamieMagee I didn't know this; just added and got rid of the `T.must(...)`

PR #9095, opened 2024-02-22
Fix README image header

davidstosik:
The block of HTML code that displays a picture header at the top of
the README file is indented with 4 spaces after an empty line.

GitHub Markdown sees that as a code block (see [Indented code
blocks](https://github.github.com/gfm/#indented-code-blocks) in the
GitHub Flavored Markdown Spec).

Removing the blank line should fix it as the HTML code won't be seen as
a code block anymore and will instead be unescaped and inserted as [raw HTML](https://github.github.com/gfm/#raw-html).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
 <h1 align="center">
-
     <picture>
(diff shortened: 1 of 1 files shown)


davidstosik:
The diff shows it worked.

jurre (approved):

jurre:
Thanks!

landongrindheim:
The header is now renering like 👇 I've opened #9100 to revert this
image

PR #9097, opened 2024-02-22
Update referenced projects during a run of NuGetUpdater.

JoeRobich:
There was an assumption being made in the Ruby side of the NuGet updater that runs of the native updater were updating the target project and referenced projects. The ruby side would then mark the referenced projects as processed and not attempt to update them separately. Whereas this assumption was true for dirs.proj files it was not true for .csproj, .vbproj, etc... This PR bring the two different ways of processing into alignment and makes the assumption true for all project files.

- Changed native NuGetUpdater to test against project files instead of always wrapping them with a solution. This brings the tests in line with how the ruby side invokes the native updater. A later PR will add test coverage around updating solutions.
- Renamed the dirs.proj unit tests to align it with the other UpdateWorkerTests.
- Refactors the SdkPackageUpdater by extracting logic into methods.


nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Updater/SdkPackageUpdater.cs
```diff
@@ -20,132 +20,124 @@ internal static class SdkPackageUpdater
...
+        Logger logger)
     {
         // SDK-style project, modify the XML directly
         logger.Log("  Running for SDK-style project");
-        var buildFiles = await MSBuildHelper.LoadBuildFiles(repoRootPath, projectPath);
-
-        var newDependencyNuGetVersion = NuGetVersion.Parse(newDependencyVersion);
 
-        // update all dependencies, including transitive
+        var buildFiles = await MSBuildHelper.LoadBuildFiles(repoRootPath, projectPath);
```
JoeRobich, line 28: The changes in this file are mostly just grouping logic into methods to make this more readable.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
```diff
@@ -101,10 +101,17 @@ public static IEnumerable<string> GetProjectPathsFromSolution(string solutionPat
...
-        var projectRootElement = ProjectRootElement.Open(projFilePath);
         var processedProjectFiles = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
-        projectStack.Push((Path.GetFullPath(Path.GetDirectoryName(projFilePath)!), projectRootElement));
+        try
+        {
+            var projectRootElement = ProjectRootElement.Open(projFilePath);
+            projectStack.Push((Path.GetFullPath(Path.GetDirectoryName(projFilePath)!), projectRootElement));
+        }
+        catch (InvalidProjectFileException)
```
JoeRobich, line 111: Needed to guard against invalid projects in this method because we are now looking for project paths before processing the project.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Updater/UpdaterWorker.cs
```diff
@@ -101,7 +104,40 @@ public async Task RunAsync(string repoRootPath, string workspacePath, string dep
...
+            return;
+        }
+
+        var projectFilePaths = MSBuildHelper.GetProjectPathsFromProject(projectPath);
+        foreach (var projectFullPath in projectFilePaths.Concat([projectPath]))
+        {
+            // If there is some MSBuild logic that needs to run to fully resolve the path skip the project
+            if (File.Exists(projectFullPath))
+            {
+                await RunUpdaterAsync(repoRootPath, projectFullPath, dependencyName, previousDependencyVersion, newDependencyVersion, isTransitive);
```
JoeRobich, line 120: This is the heart of the change. We now collect referenced projects and update them as we do for dirs.proj.

ryanbrandenburg (approved):

PR #9103, opened 2024-02-22
Allow a role to be assumed before checking Docker images in AWS ECR

### LINKED ISSUE #6152: Support IAM roles for ECR

### Feature description

Dependabot can be integrated with ECR using keys, but it doesnt support using IAM roles

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#docker-registry

It would be a good addition to allow access to ECR via IAM role instead of authenticating with keys

dgholz:
This is to resolve #6152

Very much a work-in-progress. I've tested this code with:
* An ECR repository that gives read permissions to the user via a resource policy
* A user with an identity policy to read the ECR repository
(both of the existing use-cases)

And a new use case:
* A user with permission to assume a role which has an identity policy to read the ECR repository

I still have a few challenges:
* Writing the tests to verify this behaviour
* Documenting how to use this feature
  * Mostly "which repo has the user-facing docs that need updating?"
* Understanding how the Credentials Proxy gets used from the Docker package updater
  * And if this change affects it

Any advice or guidance on these would be greatly appreciated!

I spotted a bug with `bin/dry-run` when credentials get set: they're not converted to `Dependabot::Credential`s. A follow-up to #8967?
I also saw that `regctl` assumes it's already logged in, which isn't always true.

### DIFF
--- a/bin/dry-run.rb
+++ b/bin/dry-run.rb
@@ -156,6 +156,16 @@ unless ENV["LOCAL_AZURE_ACCESS_TOKEN"].to_s.strip.empty?
   }
 end
 
+unless ENV["LOCAL_ECR_REGISTRY"].to_s.strip.empty?
+  $options[:credentials] << {
+    "type" => "docker_registry",
+    "registry" => ENV["LOCAL_ECR_REGISTRY"],
+    "username" => ENV["AWS_ACCESS_KEY_ID"],
+    "password" => ENV["AWS_SECRET_ACCESS_KEY"],
+    "session_token" => ENV["AWS_SESSION_TOKEN"]
+  }
+end
+
 unless ENV["LOCAL_CONFIG_VARIABLES"].to_s.strip.empty?
   # For example:
   # "[{\"type\":\"npm_registry\",\"registry\":\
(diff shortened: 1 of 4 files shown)


jurre:
Unfortunately, I don't think this is going to work, in production the VM that runs dependabot does not have access to credential data. We run a short-lived job specific proxy that intercepts requests and attaches any credentials that are configured for requests that match the configured registry. So while this might work in dry-runs, in production it would fail.

Unfortunately said proxy is not publicly available today

dgholz:
Thanks for the explanation, I was worried about that. I see that the  [Dependabot CLI](https://github.com/dependabot/cli) can set up a local environment with the credential proxy, but I guess that it's aimed at internal team members?

jurre:
> Thanks for the explanation, I was worried about that. I see that the [Dependabot CLI](https://github.com/dependabot/cli) can set up a local environment with the credential proxy, but I guess that it's aimed at internal team members?

Yeah the proxy container can be pulled publicly and contains an executable version of the proxy, but the source code is not available today. We've talked about opening this up as well to allow contributions such as yours but it's not something we have concrete plans for right now, as there's a bit of red tape involved and so many competing priorities. I wish I had a more satisfying answer to be honest.

dgholz:
No worries, thanks for confirming that. I'll drop this PR: the actual AWS SDK code change is pretty trivial, so no big deal to recreate later when/if the proxy gets opened up.

PR #9116, opened 2024-02-23
Set the dependabot_updater_version docker env from the build arg

Nishnha:
Fixes https://github.com/dependabot/dependabot-core/pull/8885#issuecomment-1959982179 where our sentry releases are still reporting the default value of "development" instead of the build SHA.

We were passing in the DEPENDABOT_UPDATER_VERSION env as a build arg but referencing it as an ENV in docker https://docs.docker.com/reference/cli/docker/container/run/#env

Docker buildx doesn't support the --env flag, but we can still pass the variable in as a build arg and then use it to set a docker environment variable


jurre:
How can we verify that this change resolves the issue @Nishnha ?

Nishnha:
> How can we verify that this change resolves the issue @Nishnha ?

We should be able to check what the value of `$DEPENDABOT_UPDATER_IMAGE` is on an image build from this branch.
We should also start seeing the correct release values in sentry

I also realized that docker buildx doesn't support the `--env` flag, but we can still pass the variable in as a build arg and then use it to set a docker environment variable
https://docs.docker.com/reference/cli/docker/buildx/build/

abdulapopoola (approved):

Nishnha:
This looks right!

I pulled the build image and ran docker inspect on it

```
$ docker pull ghcr.io/dependabot/dependabot-updater-bundler:44819e7652f0f213e0129c25509a6bfeb1dc5e53
$ docker images

REPOSITORY                                                               TAG                                        IMAGE ID       CREATED          SIZE
ghcr.io/dependabot/dependabot-updater-bundler                            44819e7652f0f213e0129c25509a6bfeb1dc5e53   ef0479026bcd   17 minutes ago   724MB

$ docker inspect ef0479026bcd | grep DEPENDABOT_UPDATER_VERSION
              "DEPENDABOT_UPDATER_VERSION=44819e7652f0f213e0129c25509a6bfeb1dc5e53"
```

Nishnha:
That worked!
Screenshot 2024-02-23 at 1 08 49 PM

PR #9117, opened 2024-02-23
find .nupkg URL without PackageBaseAddress

### LINKED ISSUE #8887: Artifactory private nuget package source authentication

brettfo:
The NuGet updater uses the `PackageBaseAddress` endpoint to generate URLs to download `.nupkg` packages, but some feed providers don't supply that endpoint.  This is a workaround that allows `PackageBaseAddress` to be missing/`nil` and if it is, it makes a few extra network hops to discover the `.nupkg` download URL.

This PR is also dependent on a separate internal PR that injects authentication to the proper URL endpoints.

Fixes #8887.

### DIFF
--- a/nuget/lib/dependabot/nuget/update_checker/nupkg_fetcher.rb
+++ b/nuget/lib/dependabot/nuget/update_checker/nupkg_fetcher.rb
@@ -4,6 +4,7 @@
 require "nokogiri"
 require "zip"
 require "stringio"
+require "dependabot/nuget/http_response_helpers"
 
 module Dependabot
   module Nuget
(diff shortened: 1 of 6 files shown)


nuget/lib/dependabot/nuget/update_checker/nupkg_fetcher.rb
```diff
@@ -43,11 +44,48 @@ def self.fetch_nupkg_buffer_from_repository(repository_details, package_id, pack
...
+        search_results = JSON.parse(search_response_body)
+
+        # find matching package and version
+        package_search_result = search_results&.[]("data")&.find { |d| package_id.casecmp?(d&.[]("id")) }
+        version_search_result = package_search_result&.[]("versions")&.find do |v|
+          package_version.casecmp?(v&.[]("version"))
+        end
+        registration_leaf_url = version_search_result&.[]("@id")
+
+        registration_leaf_response = fetch_url(registration_leaf_url, repository_details)
```
JoeRobich, line 77: I see a lot of conditional access prior to this point. Will `fetch_url` fail gently in the case that `registration_leaf_url` is `nil`?

JoeRobich (approved):
Looks good. Just one concern about possible nil reference.

nuget/lib/dependabot/nuget/update_checker/nupkg_fetcher.rb
brettfo: Good catch, added some early returns.

PR #9148, opened 2024-02-27
support multi-directory update with no groups

jakecoffman:
The GroupUpdateAllVersions operation needs to handle multi-directory updates regardless of whether there are any groups or not since it's the operation which understands directories.

It works because it will put all dependencies into the ungrouped collection since there are no groups defined, and then create them all as individual PRs.


updater/lib/dependabot/updater/operations/group_update_all_versions.rb
```diff
@@ -20,12 +20,14 @@ module Operations
...
         include GroupUpdateCreation
 
-        def self.applies_to?(job:)
+        def self.applies_to?(job:) # rubocop:disable Metrics/PerceivedComplexity
           return false if job.updating_a_pull_request?
           if Dependabot::Experiments.enabled?(:grouped_security_updates_disabled) && job.security_updates_only?
             return false
           end
 
+          return true if job.source.directories && job.source.directories.count > 1
```
jurre, line 29: Do we really care that a single entry in `directories` goes through this flow? We could simplify and just have it handle the single directory?

jurre (approved):
Yeah that makes sense as a first pass!

Next up what about if we simplify by just always giving the updater a job with `directories` and never `directory`, just sometimes it's an array of 1 entry that happens to be defined as `directory` in the config file?

updater/lib/dependabot/updater/operations/group_update_all_versions.rb
jakecoffman: Nope, we don't care. I will start refactoring next, operations have basically converged at this point.

PR #9149, opened 2024-02-27
handle ungrouped security updates with multi-dir version updates

jakecoffman:
If multi-directory version updates is enabled, but grouped security updates is not, then the API sends kind of an edge-casey job to the Updater.

This test passes, but it updated `frontend` and not `backend`, it really doesn't know which dependency to look at and seems to just pick the first one.

It should probably update both in individual PRs.


jakecoffman:
I merged this test into #9148.

Issue #9151, opened 2024-02-27
Property expansion doesn't work in `Directory.Packages.props`


brettfo:
The current behavior doesn't support property expansion in `Directory.Packages.props`.  Consider the following example which is acceptable MSBuild, but doesn't work in this repo:

`project.csproj`:

``` xml
<Project Sdk="Microsoft.NET.Sdk">
  ...
  <ItemGroup>
    <PackageReference Include="Some.Package" />
  </ItemGroup>
</Project>
```

`Directory.Packages.props`

``` xml
<Project>
  <PropertyGroup>
    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
  </PropertyGroup>
  <ItemGroup>
    <PackageVersion Include="Some.Package" Version="$(SomePackageVersion)" />
  </ItemGroup>
</Project>
```

`Directory.Build.props`

``` xml
<Project>
  <PropertyGroup>
    <SomePackageVersion>1.2.3</SomePackageVersion>
  </PropertyGroup>
</Project>
```

danmoseley:
Just noting here that the above (using an SDK) is a common and standard MSBuild pattern.

Meanwhile for folks hitting this a possible workaround in this specific case might be to add explicit import/s to the directory.package.props that will only kick in for dependabot, eg.,
```xml
<Import Project="directory.build.props" Condition="'$(somepackageversion)' == ''"/>
```
as the case may be.

JamieMagee:
This works now. MSBuild evaluates properties during discovery (`SdkProjectDiscovery.cs`), and `XmlFileWriter.cs` detects `$(PropertyName)` patterns during updates and modifies the property definition rather than the version attribute.

Issue #8490, opened 2023-11-30
After latest release dependabot consistently fails with `Malformed version number string...`


danmoseley:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

nuget

### dependabot.yml content

config
https://github.com/dotnet/aspire/blob/main/.github/dependabot.yml
where property is set
https://github.com/dotnet/aspire/blob/c6545a2b60213d3095865b9dda60553aaef63aed/eng/Versions.props#L37
where property is consumed
https://github.com/dotnet/aspire/blob/c6545a2b60213d3095865b9dda60553aaef63aed/Directory.Packages.props#L26  

### Updated dependency

n/a

### What you expected to see, versus what you actually saw

I appear to be hitting something similar (not resolving variables) to #8470. It has broken dependabot in  https://github.com/dotnet/aspire. Bug seems to be related to the changes to how property values and imports are followed in MSBuild format files.

Detail: it seems to be not correctly discovering that versions.props (link above) is getting imported into the project files along with directory.packages.props (link above). I don't see any relevant changes in our repo. So I came to look here and I see a lot of changes in the last few hours in what seems like relevant code
https://github.com/dependabot/dependabot-core/commits/main/nuget/lib/dependabot/nuget
and then I found this issue.

here are the github ID's if that is useful - not sure whether you can access or not.
Version update 755234367 Errored with the message "Dependabot encountered an unknown error" // soon after last passing one
Rebase update 754804010 passed // this was at ~7am PST or so on 11/27, the last time it passed
Version update 754804004 passed

```
updater | 2023/11/28 03:01:48 ERROR <job_755234367> Malformed version number string $MicrosoftAspNetCoreOpenApiPackageVersion
updater | 2023/11/28 03:01:48 ERROR <job_755234367> /usr/local/lib/ruby/3.1.0/rubygems/version.rb:223:in `initialize'
updater | 2023/11/28 03:01:48 ERROR <job_755234367> /home/dependabot/common/lib/dependabot/version.rb:9:in `initialize'
updater | 2023/11/28 03:01:48 ERROR <job_755234367> /home/dependabot/nuget/lib/dependabot/nuget/version.rb:27:in `initialize'
updater | 2023/11/28 03:01:48 ERROR <job_755234367> /usr/local/lib/ruby/3.1.0/rubygems/version.rb:204:in `new'
[...]

PR #9153, opened 2024-02-27
don't fail completely if package version cannot be parsed

brettfo:
If a package version property can't be expanded, silly things happen like querying NuGet for `$SomePackageVersion`.  This PR is 2 related fixes rolled into one.

1. Pre-filter package versions by passing them through `Version.parse()`.  If that fails, there's nothing meaningful that can be done, so we can quit early and, more importantly, cleanly.  Package versions that failed to parse are elevated to warnings.
2. Don't assume package versions in `Directory.Packages.props` are a version number; they might be a variable expansion.  Currently this PR just removes the `Version` requirement to allow the raw string to pass through, but I've filed #9151 to track future work.  The raw string pass-through will prevent a bunch of crashes if a variable is used in the property field.

This should fix the most egregious errors of #8490.

### DIFF
--- a/nuget/lib/dependabot/nuget/file_parser.rb
+++ b/nuget/lib/dependabot/nuget/file_parser.rb
@@ -30,7 +30,21 @@ module Dependabot
         dependency_set += packages_config_dependencies
         dependency_set += global_json_dependencies if global_json
         dependency_set += dotnet_tools_json_dependencies if dotnet_tools_json
-        dependency_set.dependencies
+
+        (dependencies, deps_with_unresolved_versions) = dependency_set.dependencies.partition do |d|
+          # try to parse the version; don't care about result, just that it succeeded
+          _ = Version.new(d.version)
+          true
+        rescue ArgumentError
+          # version could not be parsed
+          false
+        end
+
+        deps_with_unresolved_versions.each do |d|
+          Dependabot.logger.warn "Dependency '#{d.name}' excluded due to unparsable version: #{d.version}"
+        end
+
+        dependencies
       end
 
       private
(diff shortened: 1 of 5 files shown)


danmoseley (commented):
I can't read Ruby but the intent matches what I think we need! thanks.

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
```diff
@@ -408,10 +408,6 @@ def package_versions
                 version = dependency_version(package_node, file)
                 next unless name && version
 
-                version = Version.new(version)
-                existing_version = package_versions[name]
-                next if existing_version && existing_version > version
-
```
JoeRobich, line 414: Is the thinking that last version wins here?

nuget/lib/dependabot/nuget/file_parser.rb
```diff
@@ -30,7 +30,20 @@ def parse
         dependency_set += packages_config_dependencies
         dependency_set += global_json_dependencies if global_json
         dependency_set += dotnet_tools_json_dependencies if dotnet_tools_json
-        dependency_set.dependencies
+        (dependencies, deps_with_unresolved_versions) = dependency_set.dependencies.partition do |d|
```
JoeRobich, line 33: nit: maybe add a newline above to separate this.

JoeRobich (approved):

nuget/lib/dependabot/nuget/file_parser/project_file_parser.rb
JoeRobich: nevermind. I see the comment in the description.

brettfo: Correct, similar to multiple property definitions, last wins.

PR #9154, opened 2024-02-28
Strict type `Dependabot::PullRequestCreator::GitHub`

JamieMagee:


common/lib/dependabot/pull_request_creator.rb
```diff
@@ -101,7 +101,9 @@ def initialize(cause, pull_request)
     sig { returns(T::Hash[String, String]) }
     attr_reader :vulnerabilities_fixed
 
-    sig { returns(T.nilable(T.any(T::Array[String], T::Hash[Symbol, T::Array[Integer]]))) }
+    sig do
+      returns(T.nilable(T.any(T::Array[String], T::Hash[Symbol, T::Array[Integer]], T::Hash[String, T::Array[String]])))
+    end
```
bdragon, line 106: 1. This doesn't match the type signature of the `reviewers:` argument to the initializer, which is `T.nilable(T.any(T::Array[String], T::Hash[Symbol, T::Array[Integer]]))`. Curious if you know why?

2. I know that in this first pass we're just trying to assign types to what we already have, so this type signature is probably revealing some of the complexity in what we're doing here. One thing I've wondered when reading these PRs—can we define our own type aliases? I know we're benefitting from static type checking already but doing so could also help with readability:

```ruby
AzureReviewersT = T.type_alias { T::Array[String] }
GitHubReviewersT = T.type_alias { T::Hash[String, T::Array[String]] }
GitlabReviewersT = T.type_alias { T::Array[String] }
sig { returns(T.nilable(T.any(AzureReviewersT, GitHubReviewersT, GitlabReviewersT))) }
```

Not suggesting we necessarily do this here, just a thought.

JamieMagee: > 1. This doesn't match the type signature of the reviewers: argument to the initializer, which is T.nilable(T.any(T::Array[String], T::Hash[Symbol, T::Array[Integer]])). Should it?

Yeah, you're right. I just missed that and the `T.cast` hid the error. Thanks!

> 2. I know that in this first pass we're just trying to assign types to what we already have, so this type signature is probably revealing some of the complexity in what we're doing here. One thing I've wondered when reading these PRs—can we define our own type aliases? I know we're benefitting from static type checking already but doing so could also help with readability:

I agree. The types are just making the implicit complexity explicit. Aliases can definitely help here.

common/lib/dependabot/pull_request_creator.rb
```diff
@@ -101,7 +101,12 @@ def initialize(cause, pull_request)
     sig { returns(T::Hash[String, String]) }
     attr_reader :vulnerabilities_fixed
 
-    sig { returns(T.nilable(T.any(T::Array[String], T::Hash[Symbol, T::Array[Integer]]))) }
+    AzureReviewers = T.type_alias { T.nilable(T::Array[String]) }
+    GithubReviewers = T.type_alias { T.nilable(T::Hash[String, T::Array[String]]) }
+    GitLabReviewers = T.type_alias { T.nilable(T::Hash[Symbol, T::Array[Integer]]) }
+    Reviewers = T.type_alias { T.any(AzureReviewers, GithubReviewers, GitLabReviewers) }
```
bdragon, line 107: ❤️

bdragon (approved):

PR #9162, opened 2024-02-29
Filter out NuGet files where lines were only deleted

JamieMagee:
We've seen behaviour when we attempt to use NuGet to perform an update, where the remote dependency tree is not complete. For example, imagine we have the following dependency tree:

```mermaid
flowchart LR
    MyProject -->|Depends on| A[A 1.0]
    A -->|Depends on|B[B 1.0]
```

If package A version 2.0 is published, and it has a dependency on package B version 2.0, but that package isn't yet published to the NuGet feed, the upgrade will fail. NuGet will give the error `WARNING: Install failed. Rolling back...` and remove the existing dependencies on package A version 1.0 and package B version 1.0. This rollback behaviour of NuGet is not configurable.

To try and catch this, I'm comparing the number of lines in the original content versus the updated content. If the updated content has less lines than the original content, then a package has been removed somewhere and we likely don't want to create a pull request.

### DIFF
--- a/nuget/lib/dependabot/nuget/file_updater.rb
+++ b/nuget/lib/dependabot/nuget/file_updater.rb
@@ -42,6 +42,8 @@ module Dependabot
           normalized_content = normalize_content(f, updated_content)
           next if normalized_content == f.content
 
+          next if only_deleted_lines?(f.content, normalized_content)
+
           puts "The contents of file [#{f.name}] were updated."
 
           updated_file(file: f, content: normalized_content)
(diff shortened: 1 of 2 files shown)


JoeRobich:
Although this is certainly going to be faster, could we use the project_file_parser to get an updated set of dependencies to check for removal?

JoeRobich (approved):

JamieMagee:
Couldn't get the approach of counting dependencies to work, due to an issue with how we are attempting to update a dependency that we eventually can't.

See:
- https://github.com/dependabot/dependabot-core/actions/runs/8164760806/job/22354963387?pr=9162
- https://github.com/dependabot/dependabot-core/blob/d781f4e47705a967d35585663f0b43e6fd2b996e/updater/lib/dependabot/dependency_change_builder.rb#L69

I reverted back to the 'line counting' approach after discussion with @brettfo

brettfo (approved):

PR #9166, opened 2024-02-29
Use the project's version of a dependency to determine transitive or not

JoeRobich:
Currently the NuGet updater uses the workspace version of a dependency when determining whether it is transitive or not. However, when updating multiple project files, whether it is transitive or not is based on the dependency as that project sees it. This change has the file_updater locate the dependency from the project being processed and using its `.top_level?` to determine whether it is transitive.


brettfo:
currently re-working nuget transitive dependency handling

PR #9172, opened 2024-02-29
improve nuget v2 handling for non- nuget.org sources

### LINKED ISSUE #9160: NuGet v2 APIs not working on non-`nuget.org` endpoints
Also need to consider the situation where the v2 API returns a different endpoint URL.

E.g., Artifactory feeds expose a v2 endpoint looks like `https://SOME-NAME.jfrog.io/artifactory/api/nuget/SOME-FEED`, but it's also possible to request `https://SOME-NAME.jfrog.io/SOME-NAME/api/nuget/SOME-FEED` and in that case the base URL returned in the payload will be the original, `https://SOME-NAME.jfrog.io/artifactory/api/nuget/SOME-FEED`

Will need to authenticate against `base` feed as well as `base + Packages` entry.

brettfo:
Testing against NuGet v2 APIs revealed that some sources return results in a slightly different shape and follow different mechanisms for providing the `.nupkg` download URL.

This PR does primarily 2 things:

1. When parsing the results of `FindPackagesById()`, don't rely on the `<Id>` element under `<properties>`; it's not always there.  Instead rely on the `<title>` element directly on the `<entry>`.  All tested v2 APIs provided the same value (NuGet. MyGet, Artifactory).
2. Don't assume the `.nupkg` download URL because different sources build it differently.  Instead query the `Packages(Id=...,Version=...)` endpoint and check its `<content>` element.  All tested v2 APIs provided the same value (NuGet, MyGet, Artifactory).  The only difference between them was that NuGet and Artifactory had a `type` attribute of `application/zip` and MyGet had `binary/octet-stream`, but the end result was the same.

Fixes #9160.


nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
```diff
@@ -171,7 +171,7 @@ def build_v2_url(response, repo_details)
           base_url: base_url,
           repository_url: base_url,
           versions_url: File.join(
-            base_url,
+            base_url.delete_suffix("/"),
```
JoeRobich, line 174: Would it be better to do this where we build up the repository_details? or are there valid uses for the trailing slash?

JoeRobich (approved):

nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
brettfo: I wasn't sure if it would ever be necessary in `base_url`, but I'm explicitly removing it here because we're calling `.join()` and I was seeing instances of a double slash.

bdragon (approved):

PR #9177, opened 2024-03-01
Resolve errors from Sorbet `todo.rbi`

JamieMagee:


jurre (approved):
Looks like we can bump some sigils with this, and wow I'm shocked these made it through but happy we (well, you) found them with Sorbet!

jakecoffman:
Great work @JamieMagee! 🎉

PR #9179, opened 2024-03-01
`requirements_update_strategy` is `String` not `Symbol`

JamieMagee:
From an example stack trace

```
Parameter 'requirements_update_strategy': Expected type T.nilable(Symbol), got type String with value "widen_ranges"
Caller: /home/dependabot/dependabot-updater/vendor/ruby/3.1.0/gems/sorbet-runtime-0.5.11193/lib/types/private/methods/call_validation.rb:135
Definition: /home/dependabot/common/lib/dependabot/update_checkers/base.rb:61
```

### DIFF
--- a/bundler/lib/dependabot/bundler/update_checker.rb
+++ b/bundler/lib/dependabot/bundler/update_checker.rb
@@ -1,11 +1,13 @@
 # typed: true
 # frozen_string_literal: true
 
-require "dependabot/update_checkers"
-require "dependabot/update_checkers/base"
 require "dependabot/bundler/file_updater/requirement_replacer"
 require "dependabot/bundler/version"
 require "dependabot/git_commit_checker"
+require "dependabot/requirements_update_strategy"
+require "dependabot/update_checkers"
+require "dependabot/update_checkers/base"
+
 module Dependabot
   module Bundler
     class UpdateChecker < Dependabot::UpdateCheckers::Base
@@ -75,7 +77,7 @@ module Dependabot
 
       def requirements_unlocked_or_can_be?
         return true if requirements_unlocked?
-        return false if requirements_update_strategy == "lockfile_only"
+        return false if requirements_update_strategy == RequirementsUpdateStrategy::LockfileOnly
 
         dependency.specific_requirements
                   .all? do |req|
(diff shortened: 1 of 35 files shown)


abdulapopoola (approved):

jakecoffman:
I like this approach of leaving it as a String instead of needing to remember to turn it into a symbol!

Looks like there are some Python spec failures still?

bundler/lib/dependabot/bundler/update_checker.rb
```diff
@@ -75,7 +75,7 @@ def updated_requirements
 
       def requirements_unlocked_or_can_be?
         return true if requirements_unlocked?
-        return false if requirements_update_strategy == :lockfile_only
+        return false if requirements_update_strategy == "lockfile_only"
```
abdulapopoola, line 78: I know this already exists and you are helping out; could you please consider making these constants? That'll minimize the possibility of this regressing and help us standardize on known update strategies too (one place to look for them and change them)

bdragon (approved):
Looks good to me! I did a search and from what I can tell you got everything.

python/spec/dependabot/python/update_checker/requirements_updater_spec.rb
```diff
@@ -14,7 +14,7 @@
     )
   end
 
-  let(:update_strategy) { :bump_versions }
+  let(:update_strategy) { "bump_versions" }
```
bdragon, line 17: On L391 of this file you need to change

```diff
- %i(bump_versions bump_versions_if_necessary).each do |update_strategy|
+ %w(bump_versions bump_versions_if_necessary).each do |update_strategy|
```

and that should fix the python spec failures in CI.

JamieMagee: Thanks for the fix!

bundler/lib/dependabot/bundler/update_checker.rb
JamieMagee: I added the `RequirementsUpdateStrategy` enum and used it everywhere instead of strings or symbols.

abdulapopoola (approved):

bdragon (approved):

bdragon (approved):

jpinz (approved):

ryanbrandenburg (approved):

honeyankit (approved):

PR #9197, opened 2024-03-04
fix type of requirements_update_strategy

jakecoffman:
Seeing `Dependabot::Sorbet::Runtime::InformationalError` saying this should be a String which makes sense since it comes in from job.json. Those values all should be Strings I would think.


jurre (approved):

jakecoffman:
Looks like it can be either!

JamieMagee:
Thanks for fixing this @jakecoffman 

> Looks like it can be either!

I tried to address this in #9179

PR #9159, opened 2024-02-28
Only use credentials which have `registry` configured

JamieMagee:
We're seeing exceptions like this:

```
ERROR <job_f1193657-1f45-4e15-a1c9-f51909c517e2> Error processing @babel/traverse (NoMethodError)
ERROR <job_f1193657-1f45-4e15-a1c9-f51909c517e2> undefined method `gsub' for nil:NilClass
 
               url = "#{details['registry'].gsub(%r{/+$}, '')}/#{escaped_dependency_name}"
                                           ^^^^^
ERROR <job_f1193657-1f45-4e15-a1c9-f51909c517e2> /home/dependabot/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/registry_finder.rb:78:in `block in first_registry_with_dependency_details'
```

This should prevent us selecting credentials which are missing the `registry` key.

PR #9218, opened 2024-03-05
Only use credentials that have `registry` key when building `.npmrc`

JamieMagee:
It looks like #9159 and https://github.com/dependabot/cli/pull/290 were only partial fixes. We are still seeing these stack traces:

```
Error processing ... (KeyError)
key not found: "registry"
/usr/local/lib/ruby/3.1.0/forwardable.rb:238:in `fetch'
/usr/local/lib/ruby/3.1.0/forwardable.rb:238:in `fetch'
/home/dependabot/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/npmrc_builder.rb:227:in `block in credential_lines_for_npmrc'


JamieMagee:
Abandoning this, as it's fixed by https://github.com/dependabot/cli/pull/291

PR #9223, opened 2024-03-05
Set `branch` for `NuGet.Client` submodule

JamieMagee:
By setting `branch` in the `.gitmodules` definition, it will allow Dependabot to create pull requests when new tags are created in the `NuGet.Client` repository. Unfortunately, this will only work for patch updates, as a new branch is created to track every minor version.

Related #9222


bdragon (approved):
The only thing I'd mention is we'll need to remember to update this when upgrading minor versions

JamieMagee:
> The only thing I'd mention is we'll need to remember to update this when upgrading minor versions

Absolutely. But that's no worse than the position we're in now. At least with this change Dependabot will create PRs for patch versions.

PR #9224, opened 2024-03-06
Strict type some of `nuget`

JamieMagee:


nuget/lib/dependabot/nuget/cache_manager.rb
```diff
@@ -1,21 +1,27 @@
...
   module Nuget
     class CacheManager
+      extend T::Sig
+
+      sig { returns(T::Boolean) }
       def self.caching_disabled?
         ENV["DEPENDABOT_NUGET_CACHE_DISABLED"] == "true"
       end
 
+      sig { params(name: String).returns(T.untyped) }
```
JoeRobich, line 20: Should the returns be `T::Hash[T.untyped, T.untyped]` or is that not useful?

JoeRobich (approved):

PR #9227, opened 2024-03-06
prevent both directory and directories from being in the job definition

jakecoffman:
There's been a number of times when we accidentally defined both directory and directories in a job, and we get a strange result or a test might quietly be invalid. 

Rather than allowing an invalid job to proceed, let's validate immediately and fail loudly. 

This normally is impossible to happen in production, but if it is happening we need to know about it to fix it instead of it hiding behind another error.


bdragon (approved):

updater/lib/dependabot/job.rb
```diff
@@ -399,6 +400,12 @@ def register_experiments
       end
     end
 
+    sig { void }
+    def validate_job
+      raise "Either directory or directories must be provided" if source.directory.nil? && source.directories.nil?
```
abdulapopoola, line 405: Would it help to add more logs about what was actually set? And this also might be simpler as the XOR operation

abdulapopoola (approved):

updater/lib/dependabot/job.rb
jakecoffman: >Would it help to add more logs about what was actually set?

Nah, we log the job first thing before using this class and can also look it up via chat-op.

PR #9228, opened 2024-03-06
Fix Nuget grouped PR's

sebasgomez238:
Issue occured when because we were resetting repo files when we shouldn't. Using this helper method allows us to reset only when appropriate.


sebasgomez238:
@jakecoffman The behavior I observed when removing the line that resets the changes was that a single grouped update would work as expected but if multiple individual updates were applied in a single job then the updates would become cumulative where in the end the last update contained all of the previous updates. 

Wouldn't this mean that the modified dependency files are already being written to the disk? (Please correct me if I'm wrong)

In any case, a working solution would be to have the updater only reset if we're not conducting a grouped update which brings a lot of edge cases such as handling multiple groups or a group + an individual update.

jakecoffman:
> a working solution would be to have the updater only reset if we're not conducting a grouped update 

We already have that with this helper 👇 

https://github.com/dependabot/dependabot-core/blob/c676f5980bcd905fbb1367b3de5e9181b8f89b7d/common/lib/dependabot/shared_helpers.rb#L44-L63

For non-grouped updates it goes into this path:

https://github.com/dependabot/dependabot-core/blob/c676f5980bcd905fbb1367b3de5e9181b8f89b7d/common/lib/dependabot/shared_helpers.rb#L52-L58

For a grouped update it defers to the active workspace, where it resets at the end of all updates:

https://github.com/dependabot/dependabot-core/blob/1e066546b40e22bc588b32443e826df4f0a5a3ce/updater/lib/dependabot/updater/group_update_creation.rb#L92

Here's probably the clearest example of how to use it:

https://github.com/dependabot/dependabot-core/blob/86e8577a32223786ce7a6ff1414494a0d398de4f/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb#L56-L70

`write_temporary_dependency_files` is where it takes the in-memory files and writes them to the repo directory which may or may not have been reset, depending on if this is a grouped update or not.

brettfo (approved):

jakecoffman (approved):

PR #9229, opened 2024-03-06
Allow `on` as a YAML key

landongrindheim:
This is meant to clear out warnings from YAMLlint on our GitHub Actions workflows. Currently, `on` is being flagged. We should allow it since we use GitHub Actions for CI.

| Before | After |
|--------|--------|
| image | image|


honeyankit (approved):

JamieMagee:
I'd prefer to disable the warning for that line explicitly, with `# yamllint disable-line rule:truthy`, rather than make `on` globally truthy. It's explicitly shown in the GitHub Actions example: https://github.com/adrienverge/yamllint/blob/master/docs/integration.rst#integration-with-github-actions

JamieMagee (approved):

PR #9230, opened 2024-03-06
always directly download nupkg and cache the tfms

brettfo:
Internal logs were showing `Zip::Error` when trying to unpack the `.nupkg` inside `compatibility_checker.rb`.  The `.nupkg` contents were being cached, but when those contents are passed to `Zip::File.open_buffer()` the contents were changed which meant that subsequent attempts to open the zip stream might fail because it suddenly wasn't a valid zip archive.  The size increase didn't always result in a failure to open the zip, only sometimes.  I wasn't able to find any information about `open_buffer()` changing the buffer contents/increasing the size.

I was able to manually verify this against a test repo (100% repro before the fix, 0% after), but I couldn't figure out why or when to add a unit test and it didn't seem appropriate to check in a binary zip file to this repo.  Instead I don't cache the `.nupkg` contents and will always fetch a fresh copy, but _do_ cache the TFM results so in practice this should be the same behavior as before.

Since this behavior came out of another bug modifying how we download `.nupkg` contents (see #9204), I decided it would be beneficial to explicitly state what's going on in the downloader with a comment and a direct call to the `Excon` download API.  This way if the behavior of the previous method ever changes, this functionality won't be broken.


nuget/lib/dependabot/nuget/update_checker/compatibility_checker.rb
```diff
@@ -66,23 +66,31 @@ def project_tfms
       end
 
       def fetch_package_tfms(dependency_version)
-        nupkg_buffer = NupkgFetcher.fetch_nupkg_buffer(dependency_urls, dependency.name, dependency_version)
```
brettfo, line 69: The only change to this function is wrapping the body in `begin`/`end` so the result can be cached.  No change otherwise.

nuget/lib/dependabot/nuget/update_checker/nupkg_fetcher.rb
```diff
@@ -111,7 +111,13 @@ def self.fetch_stream(stream_url, auth_header, max_redirects = 5)
         current_redirects = 0
 
         loop do
-          response = fetch_url_with_auth(current_url, auth_header)
```
brettfo, line 114: `fetch_url_with_auth` was caching the result, but apparently repeatedly passing the result to `Zip::File.open_buffer()` will result in a corrupted buffer, so the fix is to always download and to not cache anything.

ryanbrandenburg (approved):

edgarrs (approved):

PR #9232, opened 2024-03-06
Simplify type parameters for `Gem::Version`

JamieMagee:
The upstream types for `Gem::Version.initialize`[^1] and `Gem::Version.new`[^2] are quite complex. They accept `T.any(String, Integer, Float, Gem::Version, NilClass)`.
For our use cases, these are too broad, and we see some issues with `SystemStackError` due to Sorbet's runtime type checking inserted at each level of our class hierarchy i.e. `Ecosystem::Version` -> `Dependabot::Version` -> `Gem::Version`.

Sorbet does have documentation on how to [narrow the types of a child method][1]. The documentation suggests altering the parent class to be generic, and specifying the generic type in child classes. Unfortunately, it's not feasible to alter the root class, `Gem::Version`.

Another approach might be to use one of Sorbet's escape hatches `override(allow_incompatible: true)`[^3]. But this disables the type checking entirely, and might hide other issues.

The approach I'm taking in this change is to instead write our own shim to override the accepted types.

[^1]: https://github.com/sorbet/sorbet/blob/15df026a4ffe45871809885a26142bddac5cbdde/rbi/stdlib/rubygems.rbi#L2635-L2646
[^2]: https://github.com/sorbet/sorbet/blob/15df026a4ffe45871809885a26142bddac5cbdde/rbi/stdlib/rubygems.rbi#L2738-L2749
[^3]: https://sorbet.org/docs/override-checking#use-overrideallow_incompatible-true

[1]: https://sorbet.org/docs/override-checking#what-if-i-really-want-the-child-method-to-narrow-the-type

### DIFF
--- a/sorbet/rbi/shims/rubygems.rbi
+++ b/sorbet/rbi/shims/rubygems.rbi
@@ -1,0 +1,34 @@
+# typed: strong
+# frozen_string_literal: true
+
+module Gem
+  class Version
+    sig do
+      params(
+        version: T.nilable(
+          T.any(
+            String,
+            Integer,
+            Gem::Version
+          )
+        )
+      )
+        .returns(Gem::Version)
+    end
+    def self.new(version); end
+
+    sig do
+      params(
+        version: T.nilable(
+          T.any(
+            String,
+            Integer,
+            Gem::Version
+          )
+        )
+      )
+        .void
+    end
+    def initialize(version); end
+  end
+end
--- a/common/lib/dependabot/version.rb
+++ b/common/lib/dependabot/version.rb
@@ -10,59 +10,22 @@ module Dependabot
 
     abstract!
 
-    sig do
-      override
-        .overridable
-        .params(
-          version: T.any(
-            String,
-            Integer,
-            Float,
-            Gem::Version,
-            NilClass
-          )
-        )
-        .void
-    end
+    VersionParameter = T.type_alias { T.nilable(T.any(String, Integer, Gem::Version)) }
+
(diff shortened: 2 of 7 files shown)


brettfo (approved):

abdulapopoola (approved):

sorbet/rbi/shims/rubygems.rbi
```diff
@@ -0,0 +1,34 @@
+# typed: strong
+# frozen_string_literal: true
+
+module Gem
+  class Version
```
bdragon, line 5: optional: Might be worth adding a comment about the fact that we're overriding/narrowing the underlying signature (essentially what you have in the PR description)?

sorbet/rbi/shims/rubygems.rbi
```diff
@@ -0,0 +1,34 @@
...
+        version: T.nilable(
+          T.any(
+            String,
+            Integer,
+            Gem::Version
+          )
+        )
+      )
+        .returns(Gem::Version)
+    end
```
bdragon, line 17: Since we're overriding the underlying signature... do we have valid use cases for integer and nil? Guessing a bunch of stuff breaks if we remove those without first adding types in some places?

bdragon (approved):

sorbet/rbi/shims/rubygems.rbi
JamieMagee: Integer seems to come from the `docker` updater. Let me see if I can coerce that into a string to simplify this even further.

PR #9234, opened 2024-03-06
report the current version as latest if nothing can be found

### LINKED ISSUE #9233: NuGet update checker can return `nil` for `latest_version`
When calling NuGet's `UpdateChecker.latest_version`, that can be `nil` if none of the package sources has the package.  Eventually this can result in:

```
undefined method `fetch' for nil:NilClass

                                                              ^^^^^^
```

brettfo:
When checking for updated packages, it's possible that none of the queried sources has that package.  The result is that `UpdateChecker.latest_version` can be `nil` which leads to all sorts of problems.

The fix is to report the currently installed package version as latest if nothing could be found.  This situation was found where a user is running dependabot on their repo with only their own private NuGet feed, specifically so they could always be up-to-date with their internal organization.  During this update process, the dependency `Newtonsoft.Json` version `13.0.3` was found, but since that's a public package, it wasn't found on their private feed.  The result was `undefined method 'fetch' for nil:NilClass`.  That is obviously incorrect, so in the instance where the package could not be found, it is simply reported as being up-to-date so no other action is attempted.

Fixes #9233

### DIFF
--- a/nuget/lib/dependabot/nuget/update_checker.rb
+++ b/nuget/lib/dependabot/nuget/update_checker.rb
@@ -17,7 +17,8 @@ module Dependabot
         # No need to find latest version for transitive dependencies unless they have a vulnerability.
         return dependency.version if !dependency.top_level? && !vulnerable?
 
-        @latest_version = latest_version_details&.fetch(:version)
+        # if no update sources have the requisite package, then we can only assume that the current version is correct
+        @latest_version = latest_version_details&.fetch(:version) || dependency.version
       end
 
       def latest_resolvable_version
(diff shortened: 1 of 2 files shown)


honeyankit (approved):

honeyankit:
Few CI errors to fix

brettfo:
There appear to be issues with the CI in general, I can't see how this change would cause failures in the e.g., swift updater.  I'll pull/rebase once `main` has some more commits to try to get a clean CI.

brettfo:
Ah, the CI is failing due to API rate limits.

edgarrs (approved):

PR #9240, opened 2024-03-07
Fix the number of updated directories in a group update

Nishnha:
Fix the number of updated directories in a group update.

Previously we were listing the number of directories that Dependabot would try to update instead of the number of directories that were _actually_ updated.

This only affects the PR title and the commit message (since it uses the PR title)


honeyankit (approved):

honeyankit:
Sorbet is failing

abdulapopoola (approved):

honeyankit:
This PR should fix the failing test: https://github.com/dependabot/smoke-tests/pull/194

PR #9251, opened 2024-03-08
📏 Standardize Error Keys

landongrindheim:
We're using a mix of dashes and underscores, which I find confusing. I've introduced some failures trying to get these right, so I'm making all string keys dasherized.

### DIFF
--- a/updater/lib/dependabot/base_command.rb
+++ b/updater/lib/dependabot/base_command.rb
@@ -64,7 +64,7 @@ module Dependabot
         ErrorAttributes::PACKAGE_MANAGER => job.package_manager,
         ErrorAttributes::JOB_ID => job.id,
         ErrorAttributes::DEPENDENCIES => job.dependencies,
-        ErrorAttributes::DEPENDENCY_GROUP => job.dependency_groups
+        ErrorAttributes::DEPENDENCY_GROUPS => job.dependency_groups
       }.compact
       service.record_update_job_unknown_error(error_type: "updater_error", error_details: error_details)
       service.increment_metric("updater.update_job_unknown_error", tags: {
--- a/updater/lib/dependabot/base_command.rb
+++ b/updater/lib/dependabot/base_command.rb
@@ -2,6 +2,7 @@
 # frozen_string_literal: true
 
 require "dependabot/api_client"
+require "dependabot/errors"
 require "dependabot/service"
 require "dependabot/logger"
 require "dependabot/logger/formats"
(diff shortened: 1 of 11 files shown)


JamieMagee:
What would you say to making these constants in `common/lib/dependabot/errors.rb`?

updater/lib/dependabot/base_command.rb
```diff
@@ -63,7 +63,7 @@ def handle_unknown_error(err)
         "package-manager" => job.package_manager,
         "job-id" => job.id,
         "job-dependencies" => job.dependencies,
-        "job-dependency_group" => job.dependency_groups
+        "job-dependency-group" => job.dependency_groups
```
jeffwidman, line 66: Given the value is plural, should the key be plural as well? ie `"job-dependency-groups"`...

This would be more consistent with the `"job-dependencies"` pairing above.

updater/lib/dependabot/base_command.rb
```diff
@@ -63,7 +63,7 @@ def handle_unknown_error(err)
         "package-manager" => job.package_manager,
```
jeffwidman, line 63: Thinking aloud: It's a little weird that everything else extracted from `job.*` has a key of `job-*`... but this doesn't. It's probably nbd but thought I'd mention it...

updater/lib/dependabot/base_command.rb
landongrindheim: That wasn't on my radar, but I think it makes sense. It will take a change elsewhere, but I'll try to keep this suggestion in scope for this PR 🙂

JamieMagee (approved):

updater/lib/dependabot/file_fetcher_command.rb
```diff
@@ -173,13 +174,14 @@ def handle_file_fetcher_error(error)
...
-          "error-class" => error.class.to_s,
-          "error-message" => error.message,
-          "error-backtrace" => error.backtrace.join("\n"),
-          "package-manager" => job.package_manager,
-          "job-id" => job.id,
-          "job-dependencies" => job.dependencies,
-          "job-dependency_group" => job.dependency_groups
+          ErrorAttributes::CLASS => error.class.to_s,
+          ErrorAttributes::MESSAGE => error.message,
+          ErrorAttributes::BACKTRACE => error.backtrace.join("\n"),
```
abdulapopoola, line 179: along the lines of creating the ERROR_ATTRIBUTES constant; would it be useful to create a method to handle generating the `unknown_error_details` which accepts and `error` and `dependencies`? That would facilitate reuse and standardization

landongrindheim: I considered this, but didn't pursue it because [Service#capture_exception](https://github.com/dependabot/dependabot-core/blob/d07d71042fb77662571077b56e189d292b059bc9/updater/lib/dependabot/service.rb#L101-L115) accepts other arguments for `job-dependencies` and `job-dependency-groups`. These are being provided in [ErrorHandler#error_details_for](https://github.com/dependabot/dependabot-core/blob/d07d71042fb77662571077b56e189d292b059bc9/updater/lib/dependabot/updater/error_handler.rb#L144-L149). IMO, it'd be preferable to standardize usage, but I think that stretches the task I'm trying to complete a little too far.

abdulapopoola: Thanks, please for my learning; from this code, the `dependency` and the `dependency-groups` seem to come from the same `job` value. If so, what are the tradeoffs being considered and the costs involved?

landongrindheim: I'm not 100% sure I understand the question, but I'll speak to what I think I understand.

Within [Service#capture_exception](https://github.com/dependabot/dependabot-core/blob/d07d71042fb77662571077b56e189d292b059bc9/updater/lib/dependabot/service.rb#L101-L115), we prioritize the values passed in as `dependency` and `dependency_group` (ex `dependency&.name || job&.dependencies`). There's [one place](https://github.com/dependabot/dependabot-core/blob/d07d71042fb77662571077b56e189d292b059bc9/updater/lib/dependabot/updater/error_handler.rb#L144-L149) where these values are being provided 👇

```ruby
          service.capture_exception(
            error: error,
            job: job,
            dependency: dependency,
            dependency_group: dependency_group
          )
```

This method is being delegated to by other methods, which appear at pretty high levels 👇 I count seven places that would need to be updated/refactored to make this a re-usable method that only fetches the information from `job`. I don't love the current state, but I _do_ see the value that's gained by getting reports about individual dependencies (vs all dependencies for the job).

| `#handle_job_error` | `#handle_dependency_error` |
|--------|--------|
|image | image |

jakecoffman (approved):

PR #9252, opened 2024-03-08
when resolving MSBuild properties, don't throw if it can't be resolved

brettfo:
This should allow other potential updates to flow instead of erroring out the whole process.


JamieMagee (approved):

JoeRobich (approved):

danmoseley (commented):
thanks!

abdulapopoola (approved):

danmoseley:
@brettfo now I have another error. Rather than keep going back and forth can you please fork dotnet/aspire and interpret the errors? If we can get it to work on that repo, we'll have made progress. 😄 

I opened https://github.com/dependabot/dependabot-core/issues/9297 for this.

Issue #8543, opened 2023-12-06
Problems with Local Package Source NU1301


Dave-EMIS:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

nuget

### What you expected to see, versus what you actually saw

I've had a look at the existing issues and I don't think this is covered (or at least not reported elsewhere).

We are using a Local Package source as part of a `nuget.config` file in the repository root.

```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
	<packageSources>
        <!-- Remove the machine Nuget.config that force using Packages 
             This causes problems with dotnet restore.
        -->
        <clear/>
		<!-- Some Packages are stored locally to avoid problems with getting them from ADO -->
		<add key="Local" value="packages" />
        <!-- Add the standard package source back in. -->
        <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
    </packageSources>
</configuration>
```

I have 2 solutions, both are failing with similar errors. 
1 Solution uses the package store, the other does not. This bug report will focus on the one that does not.

The solution is a simple .NET 6 / ASP.NET 6 website, single project.

I have tried adding an `ignore:` to the yml to ignore the packages that I store locally in the repo, but to no avail.

This is a private repo, so I can't share links, let me know if there's more you need.

The main error is here from the logs:

```plain
updater |   No global.json files found.
updater |   No dotnet-tools.json files found.
updater | Running for project [/home/dependabot/dependabot-updater/repo/FakeService/FakeService.csproj]
updater |   Running for SDK-style project
updater | dotnet build in GetAllPackageDependenciesAsync failed. STDOUT: MSBuild version 17.6.3+07e294721 for .NET
updater |   Determining projects to restore...
updater | /tmp/package-dependency-resolution_xrY0iA/Project.csproj : error NU1301: The local source '/tmp/package-dependency-resolution_xrY0iA/packages' doesn't exist.
updater | /tmp/package-dependency-resolution_xrY0iA/Project.csproj : error NU1301: The local source '/tmp/package-dependency-resolution_xrY0iA/packages' doesn't exist.
updater | /tmp/package-dependency-resolution_xrY0iA/Project.csproj : error NU1301: The local source '/tmp/package-dependency-resolution_xrY0iA/packages' doesn't exist.
updater |   Failed to restore /tmp/package-dependency-resolution_xrY0iA/Project.csproj (in 216 ms).
updater | 
[...]

PR #9253, opened 2024-03-09
Handle local nuget repositories

ryanbrandenburg:
The goal here is to handle local nuget feeds when they're specified in the nuget.config.

Everything under the `fixutres` path is added just for testing since we need packages which aren't in the SDK in order to verify they're being picked up from the local repo.


ryanbrandenburg:
`GetAllPackageDependencies_LocalNuGetRepos_AreCopiedToTempProject` turns out to be sort of tricky. It keeps going out to nuget.org even when I clear the `NuGet.config` first.

nuget/lib/dependabot/nuget/nuget_client.rb
```diff
@@ -21,11 +21,30 @@
...
+        params(dependency_name: String, repository_details: T::Hash[Symbol, String])
+          .returns(T.nilable(T::Set[String]))
+      end
+      private_class_method def self.get_package_versions_local(dependency_name, repository_details)
+        url = repository_details.fetch(:base_url)
+        raise "Local repo #{url} doesn't exist or isn't a directory" unless File.exist?(url) && File.directory?(url)
+
+        package_dir = File.join(url, dependency_name)
+        versions = Set.new
+        Dir.each_child(package_dir) do |child|
```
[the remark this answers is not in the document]
ryanbrandenburg: I think I'll have to do something like Dir.child(dependency_name) to make this check happy.

ryanbrandenburg:
> `GetAllPackageDependencies_LocalNuGetRepos_AreCopiedToTempProject` turns out to be sort of tricky. It keeps going out to nuget.org even when I clear the `NuGet.config` first.

The trick here was that the appropriate NuGet.config was not being found. I fixed that.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
```diff
@@ -341,10 +343,34 @@ private static ProjectRootElement CreateProjectRootElement(ProjectBuildFile buil
         var projectDirectory = Path.GetDirectoryName(projectPath);
         projectDirectory ??= repoRoot;
         var topLevelFiles = Directory.GetFiles(repoRoot);
-        var nugetConfigPath = PathHelper.GetFileInDirectoryOrParent(projectDirectory, repoRoot, "NuGet.Config", caseSensitive: false);
+        var nugetConfigPath = PathHelper.GetFileInDirectoryOrParent(projectPath, repoRoot, "NuGet.Config", caseSensitive: false);
         if (nugetConfigPath is not null)
         {
+            var nugetConfigDir = Path.GetDirectoryName(nugetConfigPath);
```
JoeRobich, line 349: This seems very NuGet-y. Should we extract this if body to the NuGetHelpers class?

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core.Test/NuGetUpdater.Core.Test.csproj
```diff
@@ -4,6 +4,7 @@
     <TargetFramework>$(CommonTargetFramework)</TargetFramework>
     <OutputType>Exe</OutputType>
     <Nullable>enable</Nullable>
+    <IsTestProject>true</IsTestProject>
```
JoeRobich: What do we get by adding this?

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core.Test/Utilities/MSBuildHelperTests.cs
```diff
@@ -319,6 +319,115 @@ public async Task AllPackageDependencies_DoNotIncludeUpdateOnlyPackages()
...
+        var nugetPackagesDirectory = Environment.GetEnvironmentVariable("NUGET_PACKAGES");
+        var nugetHttpCacheDirectory = Environment.GetEnvironmentVariable("NUGET_HTTP_CACHE_PATH");
+
+        try
+        {
+            using var temp = new TemporaryDirectory();
+
+            // It is important to have empty NuGet caches for this test, so override them with temp directories.
+            var tempNuGetPackagesDirectory = Path.Combine(temp.DirectoryPath, ".nuget", "packages");
+            Environment.SetEnvironmentVariable("NUGET_PACKAGES", tempNuGetPackagesDirectory);
```
JoeRobich, line 334: With us using this pattern more. Wonder if we could encapsulate this in an IDisposable?

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core.Test/NuGetUpdater.Core.Test.csproj
ryanbrandenburg: I was throwing stuff at the wall trying to figure out why I couldn't run tests locally. The answer ended up being that I needed to use VS Preview instead of  2022.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core.Test/Utilities/MSBuildHelperTests.cs
ryanbrandenburg: I was thinking about doing an `IDisposable`, but it gets a bit weird in this case because we only actually care about restoring some of the Environment variables. I think given that I was supposed to finish this last week I'll hold off on doing that until I'm back in the area.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
ryanbrandenburg: We spoke offline and agree that if we split this up it would be within MSBuildHelper, and only if another iteration is needed anyway.

JoeRobich (approved):

nuget/lib/dependabot/nuget/nuget_client.rb
```diff
@@ -21,11 +21,34 @@ def self.get_package_versions(dependency_name, repository_details)
...
       end
 
+      sig do
+        params(dependency_name: String, repository_details: T::Hash[Symbol, String])
+          .returns(T.nilable(T::Set[String]))
+      end
+      private_class_method def self.get_package_versions_local(dependency_name, repository_details)
+        url = repository_details.fetch(:base_url)
+        raise "Local repo #{url} doesn't exist or isn't a directory" unless File.exist?(url) && File.directory?(url)
+        raise "Local repo #{url} uses '..' which could cause it to escape the Git repository" if url.include? ".."
```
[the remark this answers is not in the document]
ryanbrandenburg: I actually think we can lose this check entirely.

kaylumah:
Awesome @ryanbrandenburg hope this can be merged soon :D it fixes #8543

brettfo (approved):

PR #9267, opened 2024-03-12
[NuGet] Add NuGetUpdater Discover command 

JoeRobich:
This moves the file_parser logic from Ruby to C#. It makes use of MSBuild and NuGet to get an accurate view of build files and dependencies. It collects a broad set of information that will be useful in later stages such as update_checker and file_updater. This information is written to a temp folder and consumed on the Ruby side.

The goal is that the discovery will run once and be the expensive operation. This will allow updating to be cheap as we already have the necessary information to directly update dependencies.


nuget/lib/dependabot/nuget/discovery/discovery_json_reader.rb
```diff
@@ -0,0 +1,94 @@
...
+        workspace_result.projects.each do |project|
+          dependency_set += project.dependency_set
+        end
+        if workspace_result.directory_packages_props
+          dependency_set += T.must(workspace_result.directory_packages_props).dependency_set
+        end
+        if workspace_result.dotnet_tools_json
+          dependency_set += T.must(workspace_result.dotnet_tools_json).dependency_set
+        end
+        dependency_set += T.must(workspace_result.global_json).dependency_set if workspace_result.global_json
```
ryanbrandenburg, line 71: It's a bummer that sorbet isn't better at detecting when a nilable type has already been confirmed as non-nil.

nuget/lib/dependabot/nuget/discovery/workspace_discovery.rb
```diff
@@ -0,0 +1,80 @@
+# typed: strong
+# frozen_string_literal: true
+
+require "dependabot/nuget/discovery/dependency_file_discovery"
+require "dependabot/nuget/discovery/directory_packages_props_discovery"
+require "dependabot/nuget/discovery/project_discovery"
```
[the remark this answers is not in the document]
JoeRobich: Let me tell you that you get it already. Went back and forth with the CI a bit last night. Happy to add the requires statements tho.

nuget/lib/dependabot/nuget/discovery/discovery_json_reader.rb
JoeRobich: Yeah, it doesn't seem to track nilable information about members only the object.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core.Test/Utilities/AssertEx.cs
```diff
@@ -0,0 +1,250 @@
+#nullable disable
```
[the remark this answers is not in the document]
JoeRobich: Good call. This wasn't as complex as I thought.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Discover/DiscoveryWorker.cs
```diff
@@ -0,0 +1,214 @@
...
+            _logger.Log("No project files found.");
+            return [];
+        }
+
+        return await RunForProjectPathsAsync(repoRootPath, projectPaths);
+    }
+
+    private static ImmutableArray<string> FindProjectFiles(string workspacePath)
+    {
+        return Directory.EnumerateFiles(workspacePath, "*.csproj", SearchOption.AllDirectories)
```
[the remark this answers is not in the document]
JoeRobich: Good call. Will go with "*.??proj" for now.

JoeRobich:
Smoke tests are failing due to new, more correct behavior. Previously, we would look at dependency nuspec files to determine the transitive dependencies, but did not take into account the TFM of the projects themselves. Dependencies in the smoke tests are built for netstandard1.6 and are pulling in the NETStandard.Library 1.6.x. However, the projects themselves are netstandard2.0 which means when MSBuild/NuGet evaluate them they bring in NETStandard.Library 2.x, which has many fewer sub-dependencies.

nuget/lib/dependabot/nuget/discovery/discovery_json_reader.rb
JamieMagee: Here's the documentation on [flow-sensitive typing][1], specifically on its limitations:

> Flow-sensitive type checking only works for local variables, not for values returned from method calls. Why? Sorbet can’t know that if a method is called twice in a row that it returns the same thing each time. Put another way, Sorbet never assumes that a method call is pure.

[1]: https://sorbet.org/docs/flow-sensitive#limitations-of-flow-sensitivity

common/spec/spec_helper.rb
```diff
@@ -85,6 +85,32 @@ def fixture(*name)
   File.read(File.join("spec", "fixtures", File.join(*name)))
 end
 
+# Creates a temporary directory and writes the provided files into it.
+#
+# @param files [DependencyFile] the files to be written into the temporary directory
+def write_tmp_repo(files,
```
JoeRobich, line 91: Since the NuGet updater needs files on disk for parsing and updating, this was super useful to convert existing tests where the test file contents were defined inline.

nuget/Dockerfile
```diff
@@ -24,7 +24,8 @@ RUN cd /tmp \
   && chmod +x dotnet-install.sh \
   && mkdir -p "${DOTNET_INSTALL_DIR}" \
   && ./dotnet-install.sh --version "${DOTNET_SDK_VERSION}" --install-dir "${DOTNET_INSTALL_DIR}" \
-  && rm dotnet-install.sh
+  && rm dotnet-install.sh \
+  && chown -R dependabot:dependabot "${DOTNET_INSTALL_DIR}/sdk"
```
JoeRobich, line 28: The NuGet file parser looks at all build files including those shipped in the SDK. This was necessary so the file could be accessed.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core.Test/Files/GlobalJsonBuildFileTests.cs
```diff
@@ -55,6 +53,7 @@ public void GlobalJson_GetDependencies_ReturnsDependencies()
     {
         var expectedDependencies = new List<Dependency>
         {
+            new("Microsoft.NET.Sdk", "6.0.405", DependencyType.MSBuildSdk),
```
JoeRobich, line 56: We report .NET SDK version info now but remove it on the Ruby side, since it has not been supported by dependabot.

nuget/spec/dependabot/nuget/update_checker/compatibility_checker_spec.rb
```diff
@@ -3,19 +3,29 @@
...
+require "dependabot/nuget/file_parser"
 require "dependabot/nuget/update_checker/compatibility_checker"
 require "dependabot/nuget/update_checker/repository_finder"
 require "dependabot/nuget/update_checker/tfm_finder"
 
 RSpec.describe Dependabot::Nuget::CompatibilityChecker do
   subject(:checker) do
+    Dependabot::Nuget::FileParser.new(dependency_files: dependency_files,
+                                      source: source,
+                                      repo_contents_path: repo_contents_path).parse
```
JoeRobich, line 15: Since several of the update_checker classes relied on information which is now supplied from the discovery.json, I run the FileParser prior to returning the class under test.

nuget/spec/dependabot/nuget/file_parser_spec.rb
```diff
@@ -54,13 +58,7 @@ def dependencies_from_info(deps_info)
...
 
     context "with a .proj file" do
-      let(:files) { [proj_file] }
-      let(:proj_file) do
-        Dependabot::DependencyFile.new(
-          name: "proj.proj",
-          content: fixture("csproj", "basic2.csproj")
-        )
-      end
+      let(:project_name) { "file_parser_proj" }
```
JoeRobich, line 61: Moved these tests from picking and choosing file to test against to having project folders containing the expected mix of files.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Discover/DiscoveryWorker.cs
```diff
@@ -0,0 +1,218 @@
+using System.Collections.Immutable;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+using NuGetUpdater.Core.Utilities;
+
```
ryanbrandenburg, line 6: Nit: Extra space

ryanbrandenburg (approved):

.github/workflows/smoke.yml
```diff
@@ -44,7 +44,7 @@ jobs:
           cat filtered.json
 
           # Curl the smoke-test tests directory to get a list of tests to run
-          URL=https://api.github.com/repos/dependabot/smoke-tests/contents/tests
+          URL=https://api.github.com/repos/dependabot/smoke-tests/contents/tests?ref=jorobich/update-nuget-tests
```
JoeRobich, line 47: Update smoke tests PR - https://github.com/dependabot/smoke-tests/pull/201

Nishnha (approved):

PR #9269, opened 2024-03-12
improve update-not-possible logging

jakecoffman:
When doing an update, you might get an update_not_possible error with the following logs:

```
updater | Latest version is 1.0.0
updater | Requirements to unlock update_not_possible
updater | Requirements update strategy bump_versions
updater | No update possible for my-dependency 1.0.0
```

One of the reasons this can happen is that all versions were filtered out while trying to update a specific dependency. 

This PR aims to improve the logging for ecosystems to give users a hint of why the update is not possible:

```
updater | Latest version is 1.0.0
updater | Requirements to unlock update_not_possible
updater | Requirements update strategy bump_versions
updater | Filtered out 156 unsupported Python 3.11 versions
updater | Filtered out 12 pre-release versions
updater | Filtered out 256 ignored versions
updater | No update possible for my-dependency 1.0.0
```

### DIFF
--- a/bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
+++ b/bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
@@ -68,7 +68,11 @@ module Dependabot
         def filter_prerelease_versions(versions_array)
           return versions_array if wants_prerelease?
 
-          versions_array.reject(&:prerelease?)
+          filtered = versions_array.reject(&:prerelease?)
+          if versions_array.count > filtered.count
+            Dependabot.logger.info("Filtered out #{versions_array.count - filtered.count} pre-release versions")
+          end
+          filtered
         end
 
         def filter_ignored_versions(versions_array)
@@ -76,6 +80,10 @@ module Dependabot
                      .reject { |v| ignore_requirements.any? { |r| r.satisfied_by?(v) } }
           if @raise_on_ignored && filter_lower_versions(filtered).empty? && filter_lower_versions(versions_array).any?
             raise AllVersionsIgnored
+          end
+
+          if versions_array.count > filtered.count
+            Dependabot.logger.info("Filtered out #{versions_array.count - filtered.count} ignored versions")
           end
 
           filtered
--- a/bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
+++ b/bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
@@ -10,11 +10,14 @@ require "dependabot/shared_helpers"
 require "dependabot/errors"
 require "dependabot/bundler/update_checker/latest_version_finder/" \
         "dependency_source"
+require "sorbet-runtime"
 
 module Dependabot
   module Bundler
     class UpdateChecker
       class LatestVersionFinder
+        extend T::Sig
+
         def initialize(dependency:, dependency_files:, repo_contents_path: nil,
                        credentials:, ignored_versions:, raise_on_ignored: false,
                        security_advisories:, options:)
@@ -65,6 +68,7 @@ module Dependabot
           relevant_versions.min
         end
 
+        sig { params(versions_array: T::Array[Gem::Version]).returns(T::Array[Gem::Version]) }
         def filter_prerelease_versions(versions_array)
           return versions_array if wants_prerelease?
 
@@ -75,6 +79,7 @@ module Dependabot
           filtered
         end
 
+        sig { params(versions_array: T::Array[Gem::Version]).returns(T::Array[Gem::Version]) }
         def filter_ignored_versions(versions_array)
           filtered = versions_array
                      .reject { |v| ignore_requirements.any? { |r| r.satisfied_by?(v) } }
--- a/cargo/lib/dependabot/cargo/update_checker/latest_version_finder.rb
+++ b/cargo/lib/dependabot/cargo/update_checker/latest_version_finder.rb
@@ -55,7 +55,10 @@ module Dependabot
         def filter_prerelease_versions(versions_array)
           return versions_array if wants_prerelease?
 
-          versions_array.reject(&:prerelease?)
+          filtered = versions_array.reject(&:prerelease?)
+          if versions_array.count > filtered.count
+            Dependabot.logger.info("Filtered out #{versions_array.count - filtered.count} pre-release versions")
+          end
         end
 
         def filter_ignored_versions(versions_array)
@@ -63,6 +66,10 @@ module Dependabot
                      .reject { |v| ignore_requirements.any? { |r| r.satisfied_by?(v) } }
           if @raise_on_ignored && filter_lower_versions(filtered).empty? && filter_lower_versions(versions_array).any?
             raise Dependabot::AllVersionsIgnored
+          end
+
+          if versions_array.count > filtered.count
+            Dependabot.logger.info("Filtered out #{versions_array.count - filtered.count} ignored versions")
           end
 
           filtered
--- a/cargo/lib/dependabot/cargo/update_checker/latest_version_finder.rb
+++ b/cargo/lib/dependabot/cargo/update_checker/latest_version_finder.rb
@@ -5,11 +5,14 @@ require "excon"
 require "dependabot/cargo/update_checker"
 require "dependabot/update_checkers/version_filters"
 require "dependabot/registry_client"
+require "sorbet-runtime"
 
 module Dependabot
   module Cargo
     class UpdateChecker
       class LatestVersionFinder
+        extend T::Sig
+
         def initialize(dependency:, dependency_files:, credentials:,
                        ignored_versions:, raise_on_ignored: false,
                        security_advisories:)
@@ -52,6 +55,7 @@ module Dependabot
           versions.min
         end
 
+        sig { params(versions_array: T::Array[Gem::Version]).returns(T::Array[Gem::Version]) }
         def filter_prerelease_versions(versions_array)
           return versions_array if wants_prerelease?
 
@@ -59,8 +63,10 @@ module Dependabot
           if versions_array.count > filtered.count
             Dependabot.logger.info("Filtered out #{versions_array.count - filtered.count} pre-release versions")
           end
+          filtered
         end
 
+        sig { params(versions_array: T::Array[Gem::Version]).returns(T::Array[Gem::Version]) }
         def filter_ignored_versions(versions_array)
           filtered = versions_array
                      .reject { |v| ignore_requirements.any? { |r| r.satisfied_by?(v) } }
--- a/common/lib/dependabot/update_checkers/version_filters.rb
+++ b/common/lib/dependabot/update_checkers/version_filters.rb
@@ -9,11 +9,20 @@ module Dependabot
       extend T::Sig
 
       sig do
-        params(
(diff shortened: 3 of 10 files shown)


jakecoffman:
Rather than try and test all of these new log messages, I added Sorbet types to make sure I didn't typo anything. That should be enough to verify this won't break things when deploying.

Some of the types were tricky though as a version array could be an array of Version, or a Hash of Symbols to stuff, including a Version. Rather than go down that rabbit hole, I left them as `T::Array[T.untyped]` since that's enough to verify I didn't break anything!

bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
```diff
@@ -65,19 +68,29 @@ def fetch_lowest_security_fix_version
...
+            Dependabot.logger.info("Filtered out #{versions_array.count - filtered.count} pre-release versions")
+          end
+          filtered
         end
 
+        sig { params(versions_array: T::Array[Gem::Version]).returns(T::Array[Gem::Version]) }
         def filter_ignored_versions(versions_array)
           filtered = versions_array
                      .reject { |v| ignore_requirements.any? { |r| r.satisfied_by?(v) } }
           if @raise_on_ignored && filter_lower_versions(filtered).empty? && filter_lower_versions(versions_array).any?
```
abdulapopoola, line 86: do we want to add a log here in case it doesn't raise? Would that help with debugging going forward too?

common/lib/dependabot/update_checkers/version_filters.rb
```diff
@@ -9,11 +9,20 @@ module VersionFilters
...
-          .returns(T::Array[T.any(Gem::Version, T::Hash[Symbol, Gem::Version])])
+        # Tricky generics explanation:
+        # There's a type T that is either a Gem::Version or a Hash with a :version key
+        # The method returns an array of T
+        # So whichever is provided as input, the output will be an array of the same type.
+        type_parameters(:T)
+          .params(
+            versions_array: T::Array[
+              T.any(
+                T.all(T.type_parameter(:T), Gem::Version),
```
abdulapopoola, line 20: can these types be named? Tagging @bdragon and @JamieMagee to weigh in

cargo/lib/dependabot/cargo/update_checker/latest_version_finder.rb
```diff
@@ -52,19 +55,29 @@ def fetch_lowest_security_fix_version
           versions.min
         end
 
+        sig { params(versions_array: T::Array[Gem::Version]).returns(T::Array[Gem::Version]) }
         def filter_prerelease_versions(versions_array)
           return versions_array if wants_prerelease?
 
-          versions_array.reject(&:prerelease?)
+          filtered = versions_array.reject(&:prerelease?)
```
abdulapopoola, line 62: just wondering; is this something that can be raised higher in the inheritance hierarchy so we don't have to implement it in every downstream implementation?

bundler/lib/dependabot/bundler/update_checker/latest_version_finder.rb
jakecoffman: I did, it's just below 👇

gradle/lib/dependabot/gradle/update_checker/version_finder.rb
```diff
@@ -76,24 +79,39 @@ def versions
...
+          filtered
         end
 
+        sig { params(possible_versions: T::Array[T.untyped]).returns(T::Array[T.untyped]) }
         def filter_date_based_versions(possible_versions)
           return possible_versions if wants_date_based_version?
 
-          possible_versions
-            .reject { |v| v.fetch(:version) > version_class.new(1900) }
+          filtered = possible_versions.reject { |v| v.fetch(:version) > version_class.new(1900) }
```
abdulapopoola, line 97: [purely curious so feel free to ignore] Do you know what the 1900 means?

gradle/lib/dependabot/gradle/update_checker/version_finder.rb
```diff
@@ -76,24 +79,39 @@ def versions
...
         end
 
+        sig { params(possible_versions: T::Array[T.untyped]).returns(T::Array[T.untyped]) }
         def filter_version_types(possible_versions)
-          possible_versions
-            .select { |v| matches_dependency_version_type?(v.fetch(:version)) }
+          filtered = possible_versions.select { |v| matches_dependency_version_type?(v.fetch(:version)) }
+          if possible_versions.count > filtered.count
+            diff = possible_versions.count - filtered.count
+            Dependabot.logger.info("Filtered out #{diff} versions not matching the type")
```
abdulapopoola, line 109: not matching the type? Can that be explained some more?

cargo/lib/dependabot/cargo/update_checker/latest_version_finder.rb
jakecoffman: Yes there are a few things like this and also filtering out versions that are lower than the current version that we could potentially pull up into the common updater code.

abdulapopoola: And if that's prohibitively expensive or not feasible, please can we fix it for all ecosystems in one go so we don't have it linger :)

gradle/lib/dependabot/gradle/update_checker/version_finder.rb
jakecoffman: If it's a date jammed into a version class then that line will filter it out? That's my guess.

gradle/lib/dependabot/gradle/update_checker/version_finder.rb
jakecoffman: Just looked up what these suffixes are actually called, they're artifact classifiers, so I updated the log message. Thanks for catching that.

cargo/lib/dependabot/cargo/update_checker/latest_version_finder.rb
jakecoffman: It's definitely out of scope for this PR, I can file an issue to track that work.

gradle/lib/dependabot/gradle/update_checker/version_finder.rb
```diff
@@ -106,7 +106,7 @@ def filter_version_types(possible_versions)
           filtered = possible_versions.select { |v| matches_dependency_version_type?(v.fetch(:version)) }
           if possible_versions.count > filtered.count
             diff = possible_versions.count - filtered.count
-            Dependabot.logger.info("Filtered out #{diff} versions not matching the type")
+            Dependabot.logger.info("Filtered out #{diff} versions with a different classifier")
```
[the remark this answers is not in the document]
jakecoffman: In context with the other log messages around it, I think it would look like:

```
updater | Latest version is 1.0.0-android
updater | Requirements to unlock update_not_possible
updater | Requirements update strategy bump_versions
updater | Filtered out 156 versions with a different classifier
updater | No update possible for my-dependency 1.0.0-android
```

I think it's clear in context assuming folks working in the ecosystem know it's called a classifier, but we can add it if you think that would help.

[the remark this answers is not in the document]
jakecoffman: I know the folks in Maven ecosystem know, this is how it would be specified in the pom:

```xml
<dependency>
   <groupId>org.example</groupId>
   <artifactId>stuff</artifactId>
   <version>1.0.0</version>       
   <classifier>android</classifier>
</dependency>
```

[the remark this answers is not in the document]
abdulapopoola: My 2c: the more verbose we are in logging; the more useful it is when debugging.

JoeRobich (approved):
NuGet changes look good to me.

gradle/lib/dependabot/gradle/update_checker/version_finder.rb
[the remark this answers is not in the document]
jakecoffman: I added the classifier name that we're filtering on to the log message.

common/lib/dependabot/update_checkers/version_filters.rb
jakecoffman: Generic type parameters across many languages use T, U, V in that order for type names.

Commonly the Java docs are referenced as the convention:
- https://docs.oracle.com/javase/tutorial/java/generics/types.html

Here's an example in the Go source code:
- https://github.com/golang/go/blob/a3f69c778fa6c082d38f9cdfacab27714f08a343/src/cmd/cgo/internal/test/typeparam.go#L10

Example in the Cargo book using `<T>` and the other cases:
- https://doc.rust-lang.org/book/ch10-00-generics.html

C++ using <T>:
- https://en.cppreference.com/w/cpp/language/templates

abdulapopoola: Sorry I wasn't clear - I understand the T, V, U naming convention for generics (I've used it too).

I meant to ask about `type aliases` like [this](https://github.com/dependabot/dependabot-core/pull/9232#discussion_r1516598410) example.

jakecoffman: Ah ok, I can make type aliases for `T::Array[Gem::Version]` and `T::Array[T.untyped]` if you think that would be helpful.

abdulapopoola: Only if it helps with readability :) Your call.

jakecoffman: Worth noting, I think `T::Hash[Symbol, Gem::Version]` is actually incorrect, it's a Hash that for the key `:version` it's a `Gem::Version`, but all the other keys are not `Gem::Version` so might make sense to introduce a proper class instead which would eliminate the need for a type alias.

gradle/lib/dependabot/gradle/update_checker/version_finder.rb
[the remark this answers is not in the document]
jurre: Yeah I think this is good, thanks! Tbh I am not sure if everyone involved knows what the classifier is, tbh, I didn't really know what it was either.

common/lib/dependabot/update_checkers/version_filters.rb
```diff
@@ -9,11 +9,20 @@ module VersionFilters
...
       sig do
-        params(
-          versions_array: T::Array[T.any(Gem::Version, T::Hash[Symbol, Gem::Version])],
-          security_advisories: T::Array[SecurityAdvisory]
-        )
-          .returns(T::Array[T.any(Gem::Version, T::Hash[Symbol, Gem::Version])])
+        # Tricky generics explanation:
+        # There's a type T that is either a Gem::Version or a Hash with a :version key
+        # The method returns an array of T
+        # So whichever is provided as input, the output will be an array of the same type.
```
JamieMagee, line 15: nit: add a link to the documentation here https://sorbet.org/docs/generics#placing-bounds-on-generic-methods

common/lib/dependabot/update_checkers/version_filters.rb
jakecoffman: Captured this info here so we can come back to it after this PR merges: https://github.com/dependabot/dependabot-core/issues/9283

JamieMagee (approved):

PR #9292, opened 2024-03-14
build(deps): bump corepack to 0.26.0

yeikel:
See https://github.com/nodejs/corepack/releases for release notes

The only change that might break some things is https://github.com/nodejs/corepack/pull/375 but just like corepack, we should not support EOL node versions anyways


yeikel:
https://github.com/dependabot/dependabot-core/pull/9324/files

PR #9293, opened 2024-03-14
Strict type some more `nuget`

JamieMagee:


nuget/lib/dependabot/nuget/update_checker/nupkg_fetcher.rb
```diff
@@ -90,31 +134,47 @@ def self.get_nuget_v3_package_url_from_search(repository_details, package_id, pa
...
+        params(
+          stream_url: String,
+          auth_header: T::Hash[String, String],
+          max_redirects: Integer
+        )
+          .returns(T.nilable(StringIO))
+      end
       def self.fetch_stream(stream_url, auth_header, max_redirects = 5)
-        current_url = stream_url
+        current_url = T.let(stream_url, T.nilable(String))
```
ryanbrandenburg, line 170: Can we avoid making this nilable  (and the T.must below) if we do T.must on `response.headers["Location"]`? I assume it's quarenteed to exist for redirect status'.

ryanbrandenburg (approved):

nuget/lib/dependabot/nuget/update_checker/nupkg_fetcher.rb
JamieMagee: Good point. Will do!

PR #9294, opened 2024-03-14
make dependency file not found message more specific

brettfo:
When an appropriate NuGet dependency file could not be found, return a more specific error message.  This also explicitly limits project files to the extensions `.csproj`, `.vbproj`, and `.fsproj`.  The earlier mechanism would allow other things like `.sfproj`, but the C# updater tool doesn't allow that, so we're tighter on what we'll accept/scan.

### DIFF
--- a/nuget/lib/dependabot/nuget/file_fetcher.rb
+++ b/nuget/lib/dependabot/nuget/file_fetcher.rb
@@ -25,7 +25,7 @@ module Dependabot
         return true if filenames.any? { |f| f.match?("^src$") }
         return true if filenames.any? { |f| f.end_with?(".proj") }
 
-        filenames.any? { |name| name.match?(/\.[a-z]{2}proj$/) }
+        filenames.any? { |name| name.match?(/\.(cs|vb|fs)proj$/) }
       end
 
       sig { override.returns(String) }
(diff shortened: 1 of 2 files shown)


abdulapopoola (approved):

JoeRobich (approved):

nuget/lib/dependabot/nuget/file_fetcher.rb
```diff
@@ -102,9 +99,16 @@ def project_files
...
+        raise_dependency_file_not_found
+      end
+
+      sig { returns(T.noreturn) }
+      def raise_dependency_file_not_found
         raise(
-          Dependabot::DependencyFileNotFound,
-          File.join(directory, "<anything>.(cs|vb|fs)proj")
+          Dependabot::DependencyFileNotFound.new(
+            File.join(directory, "*.(sln|csproj|vbproj|fsproj|proj)"),
```
abdulapopoola, line 109: any reason why we don't search for `.sln` files in the lookup above?

brettfo: Those are separately pulled in via `sln_file_names` on line 132.

PR #9302, opened 2024-03-15
Update stalebot.yml

jonjanego:
updating the user message based on feedback. end users to this repo cannot remove a label.


abdulapopoola:
what was the feedback?

jonjanego:
> what was the feedback?

users in the repo cannot remove a label, like the stalebot message suggested to do. i updated the message to remove that statement

abdulapopoola (approved):

PR #9323, opened 2024-03-18
Support Poetry non-package mode

onlined:
Starting with version 1.8, Poetry has a non-package mode for projects using Poetry not for building and publishing packages but only for dependency management. `name`, `version`, `description` and `authors` fields are optional in non-package mode.


yeikel (commented):
Please add a spec demonstrating this

onlined:
@yeikel Added the spec

yeikel:
> @yeikel Added the spec.

Thank you for taking the time. Let's wait for a review from the core team

abdulapopoola:
Thanks @onlined ! This should get reviewed and if it passes that, should be deployed during the week, thanks again.

abdulapopoola (approved):

abdulapopoola (approved):

PR #9324, opened 2024-03-18
Update corepack to fix new error after pnpm upgrade

abdulapopoola:
The PNPM bump to v8.15.5 resulted in new Sentry errors; trying to bump corepack to see if it resolves the issues.
```
 ERR_PNPM_UNSUPPORTED_PLATFORM  Unsupported platform for registry.npmjs.org/@swc/core-linux-arm-gnueabihf/1.4.8: wanted {"cpu":["arm"],"os":["linux"],"libc":["any"]} (current: {"os":"linux","cpu":"x64","libc":"glibc"})
```

### DIFF
--- a/npm_and_yarn/Dockerfile
+++ b/npm_and_yarn/Dockerfile
@@ -1,7 +1,7 @@
 FROM ghcr.io/dependabot/dependabot-updater-core
 
 # Check for updates at https://github.com/nodejs/corepack/releases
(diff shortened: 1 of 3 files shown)


jeffwidman:
I noticed Corepack is now on `28.2`(https://github.com/nodejs/corepack/releases).

Any idea if we should bump all the way to that? Or does this need to stay in lockstep with some other NPM/PNPM versions?

PR #9338, opened 2024-03-20
Cover parts of `go_modules` code with Sorbet

ByAgenT:
This PR adds `strict` Sorbet annotation in some files in `go_modules` and enables Sorbet Runtime.

### DIFF
--- a/go_modules/lib/dependabot/go_modules/file_fetcher.rb
+++ b/go_modules/lib/dependabot/go_modules/file_fetcher.rb
@@ -1,4 +1,4 @@
-# typed: strict
+# typed: strong
 # frozen_string_literal: true
 
 require "sorbet-runtime"
--- a/go_modules/lib/dependabot/go_modules/file_fetcher.rb
+++ b/go_modules/lib/dependabot/go_modules/file_fetcher.rb
@@ -38,7 +38,7 @@ module Dependabot
           directory,
           clone_repo_contents
         ) do
-          fetched_files = [T.must(go_mod)]
+          fetched_files = go_mod ? [go_mod] : []
           # Fetch the (optional) go.sum
           fetched_files << T.must(go_sum) if go_sum
           fetched_files
--- a/go_modules/lib/dependabot/go_modules/file_fetcher.rb
+++ b/go_modules/lib/dependabot/go_modules/file_fetcher.rb
@@ -1,4 +1,4 @@
-# typed: true
+# typed: strict
 # frozen_string_literal: true
 
 require "sorbet-runtime"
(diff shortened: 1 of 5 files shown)


go_modules/lib/dependabot/go_modules/file_fetcher.rb
```diff
@@ -37,23 +40,21 @@ def fetch_files
...
       end
 
       private
 
+      sig { returns(Dependabot::DependencyFile) }
       def go_mod
-        return @go_mod if defined?(@go_mod)
-
-        @go_mod = fetch_file_if_present("go.mod")
+        @go_mod ||= T.let(T.must(fetch_file_if_present("go.mod")), T.nilable(Dependabot::DependencyFile))
```
ByAgenT, line 52: @JamieMagee can you please confidence check me if this code is equivalent to what was before?

JamieMagee: The `if_present` suffix on the end of this makes is sounds like `nil` might actually be a valid value for `@go_mod`. So the `T.must` may be unnecessary here.

JamieMagee: Just checked, and the test failure is due to the `T.must` here.

go_modules/lib/dependabot/go_modules/file_fetcher.rb
```diff
@@ -35,25 +38,23 @@ def fetch_files
           directory,
           clone_repo_contents
         ) do
-          fetched_files = [go_mod]
+          fetched_files = go_mod ? [go_mod] : []
```
ByAgenT, line 41: Added conditional here to match the type

go_modules/lib/dependabot/go_modules/file_fetcher.rb
ByAgenT: Thanks, I figured out this call tree by adding `nil` check to `fetch_files` method, tests are passing now

go_modules/lib/dependabot/go_modules/file_parser.rb
```diff
@@ -85,56 +95,63 @@ def required_packages
 
             stdout, stderr, status = Open3.capture3(command)
             handle_parser_error(path, stderr) unless status.success?
-            JSON.parse(stdout)["Require"] || []
-          end
+            T.let(JSON.parse(stdout)["Require"] || [], T::Array[T::Hash[String, T.untyped]])
```
JamieMagee, line 98: Is this necessary? Or is it also caught by the outer `T.let`?

JamieMagee (approved):

go_modules/lib/dependabot/go_modules/file_parser.rb
ByAgenT: Good catch, doesn't look like it's necessary!

abdulapopoola (approved):

go_modules/lib/dependabot/go_modules/file_fetcher.rb
ryanbrandenburg, line 41: Oh ruby DOES have ternary expressions, dono why I couldn't find that.

ryanbrandenburg (approved):

PR #9347, opened 2024-03-21
only report dependencies from project files that have a target framework

brettfo:
If a `<TargetFramework>` cannot be parsed from a `.csproj`/`.vbproj`/`.fsproj`, there's no meaningful update that can be performed, so we explicitly check for that.  This also means that we can't pull dependencies directly from `packages.config`, etc. because we still depend on the project file's contents, so the tests were updated to reflect that.  `global.json` and `dotnet-tools.json` are exceptions to requiring a project file and are still handled appropriately.

Update to smoke-tests
=====
This PR depends on a PR dependabot/smoke-tests#197 and they must be taken in parallel.

Temporary commit has been removed.  ~~**N.b.**, this PR has a temporary commit that points to the smoke-tests PR branch to show the smoke-tests passing.  That commit will need to be removed before this is merged.~~

### DIFF
--- a/nuget/lib/dependabot/nuget/file_parser.rb
+++ b/nuget/lib/dependabot/nuget/file_parser.rb
@@ -63,9 +63,18 @@ module Dependabot
       def project_file_dependencies
         dependency_set = DependencySet.new
 
-        (project_files + project_import_files).each do |file|
-          parser = project_file_parser
-          dependency_set += parser.dependency_set(project_file: file)
+        project_files.each do |project_file|
+          tfms = project_file_parser.target_frameworks(project_file: project_file)
+          unless tfms.any?
+            Dependabot.logger.warn "Excluding project file '#{project_file.name}' due to unresolvable target framework"
+            next
+          end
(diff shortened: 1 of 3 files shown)


brettfo:
Smoke tests are failing for expected reasons, going to start on those now.  The short version is that a dependency is now only reported for its initiating `.csproj` where before it was reported for the `.csproj` as well as `Directory.Packages.props`, etc.

JoeRobich (approved):

brettfo:
The failing smoke test is unrelated to this PR.  Error message/diff looks like:

``` diff
--- smoke.yaml
+++ result.yaml
@@ -100,7 +97,7 @@
                   support_file: false
                   type: file
                 - content: |
-                    # This file is automatically @generated by Poetry 1.7.1 and should not be changed by hand.
+                    # This file is automatically @generated by Poetry 1.8.2 and should not be changed by hand.
 
```

abdulapopoola (approved):

PR #9353, opened 2024-03-22
Allow `package_version` to be nilable in `Dependabot::Nuget::NuspecFetcher`

JamieMagee:
Fix from #9337

Seen in production stack traces

```
Parameter 'package_version': Expected type String, got type NilClass
Caller: /home/dependabot/nuget/lib/dependabot/nuget/update_checker/dependency_finder.rb:186
Definition: /home/dependabot/nuget/lib/dependabot/nuget/update_checker/nuspec_fetcher.rb:25
```


nuget/lib/dependabot/nuget/update_checker/nuspec_fetcher.rb
```diff
@@ -18,7 +18,7 @@ class NuspecFetcher
         params(
           dependency_urls: T::Array[T::Hash[Symbol, String]],
           package_id: String,
-          package_version: String
+          package_version: T.nilable(String)
```
JoeRobich, line 21: @brettfo What should the behavior be here when `package_version` is `nil`?

abdulapopoola:
Addressed in #9365

PR #9365, opened 2024-03-25
`package_version` may be nil

bdragon:
Updates type signature to accommodate the possibility of a nil `package_version` argument in `NuspecFetcher::fetch_nuspec`.


abdulapopoola (approved):

JamieMagee:
This is a duplicate of #9353

PR #9366, opened 2024-03-26
Prevent `NoMethodError` in group update creation

bdragon:
We are seeing runtime exceptions due to calling a method on an argument that may be nil. This change properly handles nil arguments.

### DIFF
--- a/updater/lib/dependabot/updater/group_update_creation.rb
+++ b/updater/lib/dependabot/updater/group_update_creation.rb
@@ -340,6 +340,7 @@ module Dependabot
       end
 
       def deduce_updated_dependency(dependency, original_dependency)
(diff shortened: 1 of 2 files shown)


jeffwidman (approved):
Code change looks good, although I wonder how hard it'd be to add a unit test ensuring no future regressions?

bdragon:
> I wonder how hard it'd be to add a unit test ensuring no future regressions?

`Dependabot::Updater::GroupUpdateCreation` is a module and it looks like it doesn't have any tests. I'm guessing it wasn't really intended to be treated as a public API, the idea was probably that the functionality would be covered by more e2e-like updater tests or by tests for the classes it gets included in. It's included in three classes and one of those classes has tests, so I added a test there.

abdulapopoola (approved):

PR #9367, opened 2024-03-26
Prevent comparison of Integer with String in group update creation

bdragon:
We are seeing runtime exceptions due to comparing the semver components of two versions of a dependency when the components are not of the same type.

I considered two approaches:
1. Coerce the semver components of both versions to integers before comparing
2. Skip the comparison unless the semver components of both versions are of the same type

This is updater code so it's supposed to be ecosystem agnostic. Since each ecosystem brings its own version class (which inherits from `Gem::Version` and whose major/minor/patch components could be a number of different things), I opted to avoid making assumptions about the version instances. Instead, if two instances of an ecosystem's version class have major/minor/patch components of different types, we avoid the comparison altogether.


jurre:
Can we add a test that verifies this works as intended and avoids it from slipping back in?

landongrindheim (approved):

PR #9374, opened 2024-03-26
don't allow `global.json` from repo to affect MSBuild discovery

### LINKED ISSUE #9287: repo with `global.json` pinned to an older SDK can cause update failures
As discussed in [this comment](https://github.com/dependabot/dependabot-core/issues/9249#issuecomment-1995486806), a repo with a `global.json` with a pinned 6.0.x version can fail to update.

A potential fix would be to install additional SDKs in the Dockerfile _or_ do something similar to [this](https://github.com/dependabot/dependabot-core/blob/5a02df46f4e55855cc2755f42e0687f36c941f01/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs#L460) to temporarily sideline the file.

brettfo:
When the C# tool is invoked to process updates, the repo's root is set as the working directory.  It's common practice for a repo to contain a `global.json` file that specifies the .NET SDK to use, but when the C# tool is launched, it picks up that `global.json` file and tries to use that SDK version when locating MSBuild and if the SDK version specified in the file doesn't match what's installed in the updater (or it's not upgradable via `"rollForward": "latestMinor"` or similar), then an exception is thrown.  The updater (currently) has SDK 8.0.202, but it's not uncommon for repos in the wild to use 3, 5, 6, or 7; this is what causes the failure.

The desired behavior is that the updater uses the version of MSBuild that's installed in the updater and to accomplish this, the user's `global.json` is temporarily renamed so that it won't be picked up by the MSBuild locator.  Once MSBuild has been located and loaded, `global.json` is restored and the rest of the update proceeds as desired.

N.b., a common scenario is that the updater has version 8.0.202 of the SDK installed, but a repo's `global.json` specifies the last LTS like 6.0.100.  We _could_ fix this scenario by installing the major SDK versions in the updater, but installing all of the past and present versions would make the updater image >6GB and it still wouldn't work in all situations.  It's not uncommon for a repo to specify a newer SDK than the updater, includidng preview releases, etc.  It would be a never ending task to try and keep everything updated, but by doing it this way, we only have to do some creative file handling once.

Another possibility is to install the required SDK when the updater starts.  That method is probably no easier or difficult than what's happening in this PR, but the downside is that any repo that requires a different SDK will incur a large network and compute hit; each SDK installer is >200MB in size and takes 30s - 1m to install.

There is the possibility that some future version of the .NET SDK will change how MSBuild resolves imports and dependencies, but that would be such a major breaking change that we'd have to revisit the behavior anyway.

The nature of the fix is:

1. Temporarily displace `global.json` when trying to locate and load MSBuild.
2. When running the `dotnet` CLI against a temporary project (used quite frequently), ensure the working directory is properly set.

Fixes #9287

### DIFF
--- a/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Cli.Test/EntryPointTests.Update.cs
+++ b/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Cli.Test/EntryPointTests.Update.cs
@@ -3,6 +3,8 @@ using System.IO;
 using System.Text;
 using System.Threading.Tasks;
 
+using NuGetUpdater.Core;
+using NuGetUpdater.Core.Test;
 using NuGetUpdater.Core.Test.Update;
 
 using Xunit;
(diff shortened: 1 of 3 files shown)


brettfo:
@jurre I've been experimenting this morning and the method of pre-installing the common SDKs won't work in the way I had originally hoped (and I checked, the final image is 5.7G).  The updater tool is built with the 8.0 SDK and when it's launched if it sees the repo's `global.json` and thinks it's being forced to use the 6.0 SDK, but even if the 6.0 SDK is installed, the version of the DLLs it loads doesn't match what it expects so it'll fail anyway.

So we're back to the only reliable method being this; to temporarily displace the `global.json` and count on the MSBuild behavior remaining the same (or very similar), but nothing substantial has changed in over a decade in a way that would matter.  Plus there's still the issue of repos that use new preview SDKs; the only way to work with those is also the `global.json` displacement.

brbayes-msft (approved):

PR #9381, opened 2024-03-27
honor `packageSourceMapping` from `NuGet.Config`

### LINKED ISSUE #9380: NuGet updater doesn't honor `packageSourceMapping` from NuGet.Config

brettfo:
Update `repository_finder.rb` to filter NuGet sources to those with a matching `packageSourceMapping` entry.  If no mapping is found, no filtering is performed which means the regular method of scanning all sources is preserved.

Fixes #9380

### DIFF
--- a/nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
+++ b/nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
@@ -315,6 +315,23 @@ module Dependabot
           known_urls.include?(s.fetch(:url))
         end
 
+        # filter out based on packageSourceMapping
+        package_mapping_elements = doc.xpath("/configuration/packageSourceMapping/packageSource/package[@pattern]")
+        matching_package_elements = package_mapping_elements.select do |package_element|
+          pattern = package_element.attribute("pattern").value
+
+          # reusing this function for a case insensitive GLOB pattern patch (e.g., "Microsoft.Azure.*")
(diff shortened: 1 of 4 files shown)


nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
```diff
@@ -315,6 +315,21 @@ def repos_from_config_file(config_file)
           known_urls.include?(s.fetch(:url))
         end
 
+        # filter out based on packageSourceMapping
+        package_mapping_elements = doc.xpath("/configuration/packageSourceMapping/packageSource/package[@pattern]")
+        matching_package_elements = package_mapping_elements.select do |package_element|
+          pattern = package_element.attribute("pattern").value
+          File.fnmatch(pattern, @dependency.name, File::FNM_CASEFOLD) # case insensitive check
```
ryanbrandenburg, line 322: Am I reading it right that this isn't REALLY a file operation, it just uses the `File.fnmatch` method because it uses the same globbing patterns? If so you might leave a comment clarifying that, because I thought this was going out to disk at first.

ryanbrandenburg (approved):

nuget/lib/dependabot/nuget/update_checker/repository_finder.rb
brettfo: Good call, comment added.

PR #9385, opened 2024-03-28
don't fail loading build files that don't exist

brettfo:
When loading projects, we use the flag `IgnoreMissingImports` but if a project referencing the Aspire workload is loaded when that workload is not installed will fail because an attempt is made to load a generated project that isn't on disk.  The fix is to simply ensure all files exist.

### DIFF
--- a/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core.Test/Update/UpdateWorkerTests.Sdk.cs
+++ b/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core.Test/Update/UpdateWorkerTests.Sdk.cs
@@ -2497,5 +2497,36 @@ public partial class UpdateWorkerTests
                     """
                 );
         }
+
+        [Fact]
+        public async Task ProcessingProjectWithAspireDoesNotFailEvenThoughWorkloadIsNotInstalled()
+        {
+            // enumerating the build files will fail if the Aspire workload is not installed; this test ensures we can
+            // still process the update
+            await TestUpdateForProject("Newtonsoft.Json", "7.0.1", "13.0.1",
+                projectContents: """
+                    <Project Sdk="Microsoft.NET.Sdk">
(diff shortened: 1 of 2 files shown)


nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core.Test/Update/UpdateWorkerTests.Sdk.cs
```diff
@@ -2497,5 +2497,36 @@ public async Task NoChange_IfTargetFrameworkCouldNotBeEvaluated()
...
+        public async Task ProcessingProjectWithAspireDoesNotFailEvenThoughWorkloadIsNotInstalled()
+        {
+            // enumerating the build files will fail if the Aspire workload is not installed; this test ensures we can
+            // still process the update
+            await TestUpdateForProject("Newtonsoft.Json", "7.0.1", "13.0.1",
+                projectContents: """
+                    <Project Sdk="Microsoft.NET.Sdk">
+                      <PropertyGroup>
+                        <TargetFramework>net8.0</TargetFramework>
+                        <IsAspireHost>true</IsAspireHost>
```
danmoseley, line 2511: it doesn't need the aspire package reference to make this property do anything? (idk)

brettfo: Correct.  If I remove the `File.Exists` check and run this test it fails with the expected error and the Docker image where the unit tests ultimately run doesn't have the Aspire workload installed.

danmoseley: ah yes there's a little stub in the sdk that is aware that aspire is a workload.

JamieMagee (approved):

PR #9387, opened 2024-03-28
only report dependencies whose version numbers can be resolved

brettfo:
If a package version is a property that can't be resolved, don't throw because it fails the entire update process.  Instead exclude that dependency.

### DIFF
--- a/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
+++ b/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
@@ -264,7 +264,8 @@ internal static partial class MSBuildHelper
             var evaluationResult = GetEvaluatedValue(packageVersion, propertyInfo);
             if (evaluationResult.ResultType != EvaluationResultType.Success)
             {
-                throw new InvalidDataException(evaluationResult.ErrorMessage);
+                // if we can't resolve the package version, don't report the dependency
+                continue;
             }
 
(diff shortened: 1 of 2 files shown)


nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
```diff
@@ -264,7 +264,8 @@ public static IEnumerable<Dependency> GetTopLevelPackageDependencyInfos(Immutabl
             var evaluationResult = GetEvaluatedValue(packageVersion, propertyInfo);
             if (evaluationResult.ResultType != EvaluationResultType.Success)
             {
-                throw new InvalidDataException(evaluationResult.ErrorMessage);
+                // if we can't resolve the package version, don't report the dependency
+                continue;
```
JamieMagee, line 268: Should we log anything? Or is it going to get logged by the Ruby code regardless?

JamieMagee (approved):

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/MSBuildHelper.cs
brettfo: This doesn't necessarily mean anything so I don't think we need to log it.  We already log dependencies and properties that we could resolve, so if we're trying to figure out why something wasn't found, we can look there.

JoeRobich (approved):

jakecoffman (approved):

PR #9398, opened 2024-04-01
Handle requests for review from dependabot

landongrindheim:
This isn't allowed by the GitHub API, so we should handle it inline.


common/lib/dependabot/pull_request_creator/github.rb
```diff
@@ -396,6 +396,7 @@ def invalid_reviewer?(message)
         return true if message.include?("Could not resolve to a node")
         return true if message.include?("not a collaborator")
         return true if message.include?("Could not add requested reviewers")
+        return true if message.include?("Review cannot be requested from pull request author")
```
landongrindheim, line 399: We have tests covering the `not a collaborator` case. My calculation is that the behavior we're putting under test is covered, and that there isn't much value in adding a separate test case for this string.

jeffwidman (approved):

PR #9404, opened 2024-04-01
Resolve some Sorbet errors in Terraform

JamieMagee:
I removed the `FileSelector` mixin from `FileFetcher` as the mixin expects `dependency_files` to be defined. Instead, I extracted the 2 methods that are used in `FileFetcher`.

See [the Sorbet documentation][1] for the advice I followed

[1]: https://sorbet.org/docs/abstract#interfaces-and-the-included-hook

### DIFF
--- a/terraform/lib/dependabot/terraform/file_fetcher.rb
+++ b/terraform/lib/dependabot/terraform/file_fetcher.rb
@@ -1,7 +1,8 @@
-# typed: false
+# typed: strict
 # frozen_string_literal: true
 
 require "sorbet-runtime"
+
 require "dependabot/file_fetchers"
 require "dependabot/file_fetchers/base"
 require "dependabot/terraform/file_selector"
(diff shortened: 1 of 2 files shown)


terraform/lib/dependabot/terraform/file_fetcher.rb
```diff
@@ -37,22 +38,45 @@ def fetch_files
...
+
+      sig { params(file_name: String).returns(T::Boolean) }
+      def terragrunt_file?(file_name)
+        !lockfile?(file_name) && file_name.end_with?(".hcl")
       end
 
+      sig { params(filename: String).returns(T::Boolean) }
+      def lockfile?(filename)
+        filename == ".terraform.lock.hcl"
+      end
```
JamieMagee, line 69: These are the 2 methods I copied from `FileSelector`.

jurre: It'd be nice to keep one implementation of these methods, because I can definitely see the logic around identifying if a file is a terragrunt file or lockfile change over time, and now we'd have to track down all the implementations. Probably the easiest thing to do here is to extract a new module for these and include it in both places that need it.

JamieMagee: I moved them to a `FileFilter` module

JamieMagee:
I lost write permissions in this repository 😢  In the meantime I'll recreate this PR from a fork.

PR #9405, opened 2024-04-02
Update version_resolver to support yarn v4

jpandersen87:
This PR makes the needed updates to version_resolver so that npm_and_yarn test projects that use yarn v4 will not fail due to the new peer dependency message format. Currently only the yarn_berry peer_dependency project falls under this scenario, as that project assumes 4.1.1 when ran (with docker yarn using v4) due to not specifying a packageManager field in package.json. If desired, the docker yarn version can be kept at main branch version and the scenario tested for with a new project that specifies yarn v4 in it's package.json.

- npm_and_yarn docker yarn version bumped to 4.1.1
- new regex for yarn v4 peer dependency message format
- yarn v4 regex added to handle_peer_dependency_errors chain
- pipe decoration removed from original yarn_berry regex


npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -44,7 +44,15 @@ class VersionResolver
         # what ts-jest requests\n
         YARN_BERRY_PEER_DEP_ERROR_REGEX =
           /
-            YN0060:\s|\s.+\sprovides\s(?<required_dep>.+?)\s\((?<info_hash>\w+)\).+what\s(?<requiring_dep>.+?)\srequests
+            YN0060:\s│\s.+\sprovides\s(?<required_dep>.+?)\s\((?<info_hash>\w+)\).+what\s(?<requiring_dep>.+?)\srequests
```
deivid-rodriguez, line 47: Great eye catching the incorrect symbol here! How did you find about it? I wonder if we should remove the symbol from the regexp and relax it a bit instead, to avoid this kind of confusion (and to potentially avoid breakage if yarn decides to format errors using a different ASCII char or something). I think removing `\s│` from the regexp altogether may work.

deivid-rodriguez (commented):
Nice job!

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
jpandersen87: Yeah, removing the `\s│` makes sense as the error code matching in the beginning ensures we're dealing with the intended error string.

deivid-rodriguez (commented):
Nice!

nikolaigut:
When can we expect a release of this feature?

thavaahariharangit (approved):

PR #9411, opened 2024-04-02
Allow repo with only a dotnet-tools.json file

### LINKED ISSUE #9408: Support .NET Tools updates in a repository that only contains a tools manifest

### Feature description

I have a GitHub workflow that uses some .NET global tools to do things where the versions were pinned using environment variables. I had the idea that if I used a `.config/dotnet-tools.json` file to specify the versions instead, I could leverage dependabot to keep the tools up-to-date without me needing to manually bump the versions (see https://github.com/martincostello/github-automation/commit/64b1cd7dfcc01c8586fcffaa1799d04a1db619fc).

However, it turns out that this won't work as-is because my repository doesn't contain any .NET _code_:

> Errored with the message "Dependabot couldn't find a *.(sln|csproj|vbproj|fsproj|proj)"

image

It would be beneficial to update the relevant code to support this scenario by checking for the presence of a `.config/dotnet-tools.json` file in the root of the repository so that the job does not bail-out early with an error.

martincostello:
Allow a repository to only contain a `dotnet-tools.json` file for .NET dependabot updates.

I wasn't sure where was best to put an end-to-end test for this scenario.

Resolves #9408.

### DIFF
--- a/nuget/lib/dependabot/nuget/file_fetcher.rb
+++ b/nuget/lib/dependabot/nuget/file_fetcher.rb
@@ -24,13 +24,14 @@ module Dependabot
         return true if filenames.any? { |f| f.end_with?(".sln") }
         return true if filenames.any? { |f| f.match?("^src$") }
         return true if filenames.any? { |f| f.end_with?(".proj") }
+        return true if filenames.any? { |f| f.match?(/^dotnet\-tools\.json$/i) }
 
         filenames.any? { |name| name.match?(/\.(cs|vb|fs)proj$/) }
       end
 
       sig { override.returns(String) }
       def self.required_files_message
-        "Repo must contain a .proj file, .(cs|vb|fs)proj file, or a packages.config."
+        "Repo must contain a .proj file, .(cs|vb|fs)proj file, a packages.config, or a dotnet-tools.json."
       end
 
       sig do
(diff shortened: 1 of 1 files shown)


nuget/lib/dependabot/nuget/file_fetcher.rb
```diff
@@ -24,13 +24,14 @@ def self.required_files_in?(filenames)
         return true if filenames.any? { |f| f.end_with?(".sln") }
         return true if filenames.any? { |f| f.match?("^src$") }
         return true if filenames.any? { |f| f.end_with?(".proj") }
+        return true if filenames.any? { |f| f.match?(/^dotnet\-tools\.json$/i) }
```
martincostello, line 27: Not being a ruby person, I'm not sure what the best way to refactor this to make the linter happy about the perceived complexity.

JoeRobich: This change may allow the file_fetcher to run but it doesn't change how the updater works and it currently is expecting a project file to work against.

martincostello: Could you point me at the rough area of code I need to go and touch? I started off by searching for the error message I got in the GitHub UI.

JoeRobich: May have to hold off for a bit as we are trying to move more of the updater logic from ruby to C# for the NuGet ecosystem. The new `discover` command in the NuGetUpdater.Core project is now responsible for parsing dependencies. It currently uses project files as its starting point but could be updated to simply look for global.json and .config/dotnet-tools.json in the target directory. The `update` command in NuGetUpdater.Core is responsible for performing the update and will be going through some changes this month.

Issue #7840, opened 2023-08-18
Grouped dependencies: The title for updating a single dependency should be as before


huehnerlady:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

Gradle

### Package manager version

8.3

### Language version

Java/Groovy

### What you expected to see, versus what you actually saw

When there is only 1 dependency updated, I expect the Title to be the same as the PR titles without the grouping feature, so you see the dependency itself in the title.
eg: 	`Bump org.jetbrains:annotations from 23.0.0 to 24.0.1`

The title however shows
`Bump the dependencies group with 1 update`

yeikel:
Can you please share at least  : 

1. Smallest manifest that reproduces the issue
2. Images of the diff or a link to the PR, issue, or logs

With the little information you provided, it is a bit challenging to help you

jakecoffman:
I'm going to close this as a duplicate of #7662 since it was filed first, but thanks for reproducing the problem! We'll use it to test when we get around to it.

jakecoffman:
Reopening as this request is to change a grouped update title to look like a non-grouped update title if there is only 1 dependency updated. This is different than the bug identified in #7662.

huehnerlady:
@ghjm that is an interesting idea and I would argue would also resolve my issue as if I then say `minimum_updates=2` it should have the expected effect for the single dependabot PRs. maybe even a good default value for that parameter?

PR #9416, opened 2024-04-02
If only 1 dep in group is updated, use solo title

pavera:
Fixes https://github.com/dependabot/dependabot-core/issues/7840

I believe I updated the tests to cover both single update and multi-update groups correctly, but this would be something @Nishnha, @jurre or @jakecoffman could hopefully confirm? The biggest question is that in the case of the "directory" test, I just updated the test to be a multi-update because I thought the main point there was the presence of "directory" in the title, and we have other tests that prove the correct behavior of single vs multi updates in groups so I didn't test single update groups with directory.


honeyankit:
@pavera one test is failing in common. You just need to replace the multi-dir test with single dep test and test will pass.

jurre:
I'd like to take a minute and think through wether this is actually a good change to make. My worry is that with this change it's not apparent that the PR is actually a grouped PR, which behave slightly differently from ungrouped PRs. I understand the desire to get more actionable information into the PR title, but we should think about the consequences of that as well.

pavera:
As far as being able to tell if it is a grouped PR, the body still indicates that it is a bump in the group with 1 update. I did have the same concern about hiding all evidence of a group update and made sure the body still had that info. However, maybe we append or prepend `grouped` somehow in the title?

pavera:
I like `Bump business from 1.0.0 to 2.0.0 in the all-the-things group` best of those 2 options, let me attempt to make that happen.

jurre (approved):

landongrindheim (approved):

PR #9423, opened 2024-04-03
allow flamegraph gathering

jakecoffman:
To debug customer performance issues, this adds the ability to request a flamegraph be generated.

The CLI PR that will enable this is https://github.com/dependabot/cli/pull/299.

I tested this with the command `dependabot update bundler dependabot/dependabot-core --directory updater --flamegraph --dep vcr` and it produced a nice flamegraph!

### DIFF
--- a/common/lib/dependabot/pull_request_creator/message_builder.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder.rb
@@ -478,7 +478,7 @@ module Dependabot
                        "`#{dep.humanized_version}`"
                      ]
                    end
-                   "\n\n#{table([header] + rows)}"
+                   "\n\n#{table([header] + rows)}\n"
                  elsif update_count > 1
                    dependency_links_in_directory = dependency_links_for_directory(directory)
                    " #{T.must(T.must(dependency_links_in_directory)[0..-2]).join(', ')}" \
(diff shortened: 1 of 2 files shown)


honeyankit (approved):

pavera:
When I was looking into this it seemed stackprof could generate flamegraphs itself, though that required a command run on the shell against a profile. I wonder if it might be useful though to get both the raw profile and the flamegraph? I'm not sure what else we could do with raw profile data so it probably makes sense to just get the flamegraph for now.

pavera (approved):

updater/Gemfile
```diff
@@ -35,6 +35,9 @@ gem "sentry-opentelemetry", "~> 5.16"
 gem "sentry-ruby", "~> 5.16"
 gem "terminal-table", "~> 3.0.2"
 
+gem "flamegraph", "~> 0.9.5"
+gem "stackprof", "~> 0.2.16"
```
jurre, line 39: FYI this is already in common, since we have the functionality in the dry-run script: https://github.com/dependabot/dependabot-core/blob/1fc9a86809361f0f8a121d2b8cd31d77b80e03f4/common/dependabot-common.gemspec#L57

jakecoffman: Do you recommend I remove it here and switch that line in common to be `spec.add_dependency`?

PR #9424, opened 2024-04-03
remove redundant parsing of original files

jakecoffman:
In #8963 I made DependencySnapshot aware of multiple directories, so we can remove this TODO now.

This should speed up all grouped updates a small amount because it was re-parsing the files an extra time for each group.

Now we can simply do `dependency_snapshot.dependencies` which will return what DependencySnapshot parsed at the beginning of the job run.


jakecoffman:
I should mention that the code I'm removing was added in #8803, and we have a test for that case which is still passing. So this should be good to go!

honeyankit (approved):

pavera (approved):

PR #9425, opened 2024-04-03
speed up bundler updates by parsing gemspecs concurrently

jakecoffman:
I'm debugging slowness in the bundler ecosystem and generated a flame graph for our own updater since it had a lot of path dependencies using the command `dependabot update bundler dependabot/dependabot-core --directory updater --dep vcr --flamegraph`.

I noticed this section of parsing each gemspec could easily run concurrently.

Running locally this saved about 5 seconds! This could have much bigger payoffs for an even larger repo, and the fact grouped updates will reparse after each dependency updates to see what happened.


honeyankit (approved):

jakecoffman:
@honeyankit There's no queries at this point. It's just parsing each gemspec, but the amount of setup required for each gemspec was rather IO heavy, each was setting up it's own git workspace, and shelling out to the native helper. 

So I pulled up the git workspace one level, and running each native helper, anytime it gets stuck on IO it's going to pick up work on the next gemspec.

bdragon (commented):
Just adding a couple questions:

- On a multicore machine all threads still end up running on the same core, right? And it just multiplexes CPU time as threads are blocked/unblocked?
- All four threads are mutating the method-local `dependencies` variable, which is a `DependencySet` and I think does some data merging/munging under the hood when multiple versions of the same dependency are added. Is this fine or should we use a mutex? Or maybe make this a map operation that combines the result of each thread once all of them have returned?
- Each of the four threads spawns a process (`NativeHelpers.run_bundler_subprocess` -> `SharedHelpers.run_helper_subprocess` -> `Open3.capture3`). Do we have any concerns about this? I imagine if a `HelperSubprocessFailed` is raised in a thread it will bubble up normally and the other threads/processes will be stopped and cleaned up?

Issue #8634, opened 2023-12-16
Grouped version updates don't have changelogs/commits for Bundler


enescakir:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

Bundler

### Package manager version

2.4.10

### Language version

3.2.2

### Manifest location and content before the Dependabot update

https://github.com/ubicloud/ubicloud/blob/main/Gemfile

### dependabot.yml content

https://github.com/ubicloud/ubicloud/blob/main/.github/dependabot.yml

### What you expected to see, versus what you actually saw

I expected to see a changelogs/commits for the updated packages. However, the PR description doesn't contain any information about these updates.
The PR with this issue: https://github.com/ubicloud/ubicloud/pull/1024

Previously, it was putting changelogs/commits for the updated package. We have a PR with details from two weeks ago: https://github.com/ubicloud/ubicloud/pull/961

It look like that dependabot can't generate the PR description for `addressable` gem.

It appears that when dependabot cannot generate a PR description for a dependency in the group, it prevents dependabot from generating PR descriptions for all packages.

### Images of the diff or a link to the PR, issue, or logs

https://github.com/ubicloud/ubicloud/pull/1024

### Smallest manifest that reproduces the issue

https://github.com/enescakir/ruby-dependabot-test/pulls

enescakir:
Here's the dependabot logs.
[dependabot-logs.log](https://github.com/dependabot/dependabot-core/files/13693939/dependabot-logs.log)

These lines appear suspicious to me:

```
  proxy | 2023/12/16 21:01:43 [293] 200 https://api.github.com:443/repos/sporkmonger/addressable/contents/CHANGELOG.md?ref=main
updater | 2023/12/16 21:01:43 ERROR <job_762978551> Error while generating PR message: undefined method `name' for [:name, "CHANGELOG.md"]:Array
updater | 
updater |             tmp_files.find { |f| f.name == filename }
updater |                                   ^^^^^
  proxy | 2023/12/16 21:01:43 [295] GET https://roda.jeremyevans.net:443/status
  proxy | 2023/12/16 21:01:43 [295] 404 https://roda.jeremyevans.net:443/status
  proxy | 2023/12/16 21:01:43 [297] GET https://roda.jeremyevans.net:443/status
  proxy | 2023/12/16 21:01:43 [297] 404 https://roda.jeremyevans.net:443/status
[...]

PR #9429, opened 2024-04-03
Better support around bundler changelogs

Nishnha:
Bundler implements [#suggested_changelog_url](https://github.com/dependabot/dependabot-core/blob/36e67c6711c43cda2364cd78d7e38d5e2134ac0b/bundler/lib/dependabot/bundler/metadata_finder.rb#L40) which returns the `changelog_url` for gems hosted on rubygems like https://rubygems.org/gems/rack

This `changelog_url` can point to a changelog, a specific section of a changelog (e.g. github.com/myorg/myrepo/blob/main/CHANGELOG.md#v1.4.0), or just the repo. It can also include a reference to the latest tag or branch.

When Dependabot receives this `changelog_url` it looks up the resource on GitHub and receives the response as a `Sawyer::Resource`. This resource can either be a Hash if the `changelog_url` points to a file, or it can be an Array of files if the `changelog_url` is ambiguous.

Prior to this PR Dependabot would treat all responses as an Array and would raise an error if the response was a Hash. This caused the PR body to be missing for grouped updates!

This PR teaches Dependabot how to handle changelog responses that are Hashes. 

I was able to recreate this issue at https://github.com/dsp-testing/investigate-dependabot-core-8634/pull/2 and tested it against my branch with the CLI and it successfully generated the PR body and didn't raise an error! (see https://github.com/github/dependabot-updates/issues/5578#issuecomment-2037514371 for the output)

Fixes https://github.com/dependabot/dependabot-core/issues/8634 and https://github.com/github/dependabot-updates/issues/5578

### DIFF
--- a/common/lib/dependabot/pull_request_creator/message_builder.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder.rb
@@ -233,22 +233,41 @@ module Dependabot
 
       sig { returns(String) }
       def group_pr_name
+        if source.directories
+          grouped_directory_name
+        else
+          grouped_name
+        end
+      end
+
+      sig { returns(String) }
+      def grouped_name
+        updates = dependencies.map(&:name).uniq.count
+        if dependencies.count == 1
+          "#{solo_pr_name} in the #{T.must(dependency_group).name} group"
+        else
+          "bump the #{T.must(dependency_group).name} group#{pr_name_directory} " \
+            "with #{updates} update#{'s' if updates > 1}"
+        end
+      end
+
+      sig { returns(String) }
+      def grouped_directory_name
+        updates = dependencies.map(&:name).uniq.count
+
         directories_from_dependencies = dependencies.to_set { |dep| dep.metadata[:directory] }
 
         directories_with_updates = source.directories&.filter do |directory|
           directories_from_dependencies.include?(directory)
         end
 
-        updates = dependencies.map(&:name).uniq.count
-
-        if source.directories
+        if dependencies.count == 1
+          "#{solo_pr_name} in the #{T.must(dependency_group).name} group across " \
+            "#{T.must(directories_with_updates).count} directory"
+        else
           "bump the #{T.must(dependency_group).name} group across #{T.must(directories_with_updates).count} " \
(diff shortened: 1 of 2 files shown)


common/spec/dependabot/metadata_finders/base/changelog_finder_spec.rb
```diff
@@ -165,44 +173,72 @@
...
+
+              suggested_github_url_with_tag =
+                "https://api.github.com/repos/mperham/sidekiq/contents/?ref=v1.4.0"
+              stub_request(:get, suggested_github_url_with_tag)
+                .to_return(status: 404)
+            end
+
+            it "returns nil for the changelog url" do
+              expect(subject).to eq(nil)
+            end
```
Nishnha, line 240: I thought about showing that the ChangelogFinder will still try to find the changelog through other means, but there are other tests for that ("but with a changelog on the tag"), and I decided that returning nil instead of throwing an error was the important thing to test

common/lib/dependabot/metadata_finders/base/changelog_finder.rb
```diff
@@ -128,6 +128,16 @@ def changelog_from_suggested_url
           tmp_files = T.unsafe(suggested_source_client).contents(suggested_source&.repo, opts)
 
           filename = T.must(T.must(suggested_changelog_url).split("/").last).split("#").first
+
+          # If the suggested source points to a specific directory
+          # then we will receive a hash for just the changelog file
+          if suggested_source&.directory && tmp_files[:name] == filename
+            @changelog_from_suggested_url = tmp_files
+            return @changelog_from_suggested_url
```
jurre, line 136: If we're going to return this immediately, no need to first assign it

```suggestion
            return tmp_files
```

Nishnha: Hmm but if we don't assign it to `@changelog_from_suggested_url` then the method won't be memoized

I could change it to use `||= begin ... end` ?

jurre: Ah i'd missed that 😄 `return @changelog_from_suggested_url = tmp_files` will do the trick

jurre (approved):

Issue #7695, opened 2023-08-02
Listing unchanged dependency twice in the update PR


iesahin:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

bundler

### Package manager version

Bundler version 2.4.10

### Language version

ruby 3.2.2 (2023-03-30 revision e51014f9c0) [arm64-darwin21]

### Manifest location and content before the Dependabot update

/Gemfile

The relevant part is

```
# JWT auth servr/provider for the frontend to authenticate with rails
# Need to use this specific commit, which is unofficially the 2.2 release,
#   because the new version hasn't been released to RubyGems yet.
# See https://davidgay.org/programming/jwt-auth-rails-6-knock/
# and https://github.com/nsarno/knock/issues/250
gem "knock", github: "nsarno/knock", branch: "master" # rubocop:disable Bundler/GemVersion
```

and 

/Gemfile.lock

```
GIT
  remote: https://github.com/nsarno/knock.git
  revision: 8e8b3e8d29eccb83a83ea2449e006250f025632a
  branch: master
  specs:
    knock (2.2.0)
      bcrypt (~> 3.1)
      jwt (~> 2.2.1)
      rails (>= 5)

...

knock!

```

### dependabot.yml content

```
version: 2
updates:
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
      day: "tuesday"
    # https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#commit-message
    commit-message:
      # Use our PR title naming convention
      prefix: "UPGRADE - "
    reviewers:
      - "bmulholland"
      - "kmewhort"
      - "nikasvan"
      - "iesahin"

  - package-ecosystem: "bundler"
    directory: "/"
    allow:
      # By default, dependabot updates only direct dependencies. This updates
      # sub-/transitive dependencies
      - dependency-type: "all"
    schedule:
      interval: "weekly"
      day: "tuesday"
      # Emre prefers these at noon his time
      timezone: "Turkey"
      time: "12:00"
    commit-message:
      # Use our PR title naming convention
      prefix: "UPGRADE - "
    groups:
      # Dependabot grouped updates:
      # Docs at https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups
      # This is the name of your group, it will be used in PR titles and branch
      # names
[...]

PR #9436, opened 2024-04-04
Fix Duplicate Dependencies Showing in PR summary table

honeyankit:
## Context

This PR resolves the issue of duplicated entries in the group update PR's summary table by ensuring that only unique dependencies are considered in the message builder. These are then displayed in the PR summary table.

Fixes: https://github.com/dependabot/dependabot-core/issues/7695

### DIFF
--- a/common/lib/dependabot/pull_request_creator/message_builder.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder.rb
@@ -518,7 +518,7 @@ module Dependabot
       def group_intro
         update_count = dependencies.map(&:name).uniq.count
         # Ensure dependencies are unique by name
-unique_dependencies = dependencies.uniq(&:name)
+        unique_dependencies = dependencies.uniq(&:name)
 
         msg = "Bumps the #{T.must(dependency_group).name} group#{pr_name_directory} " \
               "with #{update_count} update#{update_count > 1 ? 's' : ''}:"
--- a/common/lib/dependabot/pull_request_creator/message_builder.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder.rb
@@ -518,7 +518,7 @@ module Dependabot
       def group_intro
         update_count = dependencies.map(&:name).uniq.count
         # Ensure dependencies are unique by name
-        unique_dependencies = dependencies.uniq { |dep| [dep.name] }
+unique_dependencies = dependencies.uniq(&:name)
 
         msg = "Bumps the #{T.must(dependency_group).name} group#{pr_name_directory} " \
               "with #{update_count} update#{update_count > 1 ? 's' : ''}:"
--- a/common/lib/dependabot/pull_request_creator/message_builder.rb
+++ b/common/lib/dependabot/pull_request_creator/message_builder.rb
@@ -517,13 +517,15 @@ module Dependabot
       sig { returns(String) }
       def group_intro
         update_count = dependencies.map(&:name).uniq.count
+        # Ensure dependencies are unique by name
+        unique_dependencies = dependencies.uniq { |dep| [dep.name] }
 
         msg = "Bumps the #{T.must(dependency_group).name} group#{pr_name_directory} " \
               "with #{update_count} update#{update_count > 1 ? 's' : ''}:"
 
         msg += if update_count >= 5
                  header = %w(Package From To)
-                 rows = dependencies.map do |dep|
+                 rows = unique_dependencies.map do |dep|
                    [
                      dependency_link(dep),
                      "`#{dep.humanized_previous_version}`",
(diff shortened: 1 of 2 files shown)


common/lib/dependabot/pull_request_creator/message_builder.rb
```diff
@@ -498,13 +498,15 @@ def multi_directory_group_intro
       sig { returns(String) }
       def group_intro
         update_count = dependencies.map(&:name).uniq.count
+        # Ensure dependencies are unique by name
+        unique_dependencies = dependencies.uniq { |dep| [dep.name] }
```
jurre, line 502:
```suggestion
unique_dependencies = dependencies.uniq(&:name)
```

jurre:
Can we add a test to prevent regression?

honeyankit:
> Can we add a test to prevent regression?

I have tested against the test repo. I am adding the test.

Nishnha:
There's also a mention of the table showing an update from a commit to a version, which I think is why they're showing up even though they didn't update: https://github.com/dependabot/dependabot-core/issues/7695#issuecomment-1662267570

Is a fix for that going to be part of this PR?

honeyankit:
> There's also a mention of the table showing a dependency updating from a commit to a version, even though they didn't actually update: [#7695 (comment)](https://github.com/dependabot/dependabot-core/issues/7695#issuecomment-1662267570)
> 
> Is a fix for that going to be part of this PR?

No. This will fix the duplicate entry. I can look into that as well. But first will let this ship

landongrindheim (approved):

common/spec/dependabot/pull_request_creator/message_builder_spec.rb
```diff
@@ -2309,6 +2309,116 @@ def commits_details(base:, head:)
...
+            Dependabot::Dependency.new(
+              name: "business5",
+              version: "0.17.0",
+              previous_version: "0.16.2",
+              package_manager: "dummy",
+              requirements: [],
+              previous_requirements: []
+            )
+          end
+          let(:dependencies) { [dependency, dependency2, dependency3, dependency4, dependency5, dependency6] }
```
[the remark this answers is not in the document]
honeyankit: I like this. Adding it now

Nishnha:
Since we're only uniq'ing by name, wouldn't the dependencies that gets removed from the PR body be random? I wonder if the "to" and "from" version can change for a duplicate dependency if you run the same job twice

honeyankit:
@Nishnha : For less then 5 dependencies, similar approach is used [here](https://github.com/dependabot/dependabot-core/blob/005dd7314eb6f4dfc2f1505631f01e19f3138c73/common/lib/dependabot/pull_request_creator/message_builder.rb#L610-L615) where by the dependency name the unique dependencies are filtered. So I have just refer that for more then 5 dependencies.

Nishnha:
> @Nishnha : For less then 5 dependencies for group PR, similar approach is used [here](https://github.com/dependabot/dependabot-core/blob/005dd7314eb6f4dfc2f1505631f01e19f3138c73/common/lib/dependabot/pull_request_creator/message_builder.rb#L610-L615) where by the dependency name the unique dependencies are filtered. So I have just refer that for more then 5 dependencies.

Right but the dependency version isn't displayed for <5 dependencies

jurre:
We could check for unique by name, from and to to be thorough?

honeyankit:
@jurre and @Nishnha : I have updated the code to remove duplicates based on name, to and from versions. 

1. Both Dep A will be shown
```bash
Dep A - previous version 1.2.4 latest version 1.2.5 
Dep A - previous version 1.2.4 latest version 1.2.6 
```

2. Both Dep A will be shown
```bash
Dep A - previous version 1.2.1 latest version 1.2.5 
Dep A - previous version 1.2.4 latest version 1.2.5
```

honeyankit:
All the smoke test errors are due to rate limit exceeded.

common/spec/dependabot/pull_request_creator/message_builder_spec.rb
```diff
@@ -2422,6 +2422,142 @@ def commits_details(base:, head:)
           end
         end
 
+        context "with five or more dependencies with same name, differet to and from versions", :vcr do
```
Nishnha: this line has a type for different
do we need the vcr? I don't see one added here

honeyankit: Removed all the unnecessary vcr from all the tests. I had copied it from other test which already had the vcr along with manual stub. Good catch.

honeyankit: Typo was fixed already: https://github.com/dependabot/dependabot-core/pull/9436/commits/4d241b2d02d95988a25f8e1d248bb1080edc3d92

Nishnha (approved):
Nice

edmorley:
@honeyankit @Nishnha @landongrindheim Could this have caused the regression seen in #9457?

PR #9443, opened 2024-04-05
bundler: avoid repeat lockfile parsing with caching

jakecoffman:
The bundler ecosystem repeatedly parsed the customer's lockfile during file fetching, file parsing, and file updating.

To speed things up I've introduced a cache which will return the pre-parsed lockfile.

### DIFF
--- a/bundler/lib/dependabot/bundler/cached_lockfile_parser.rb
+++ b/bundler/lib/dependabot/bundler/cached_lockfile_parser.rb
@@ -1,4 +1,4 @@
-# typed: strict
+# typed: strong
 # frozen_string_literal: true
 
 require "sorbet-runtime"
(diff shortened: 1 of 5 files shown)


Nishnha (approved):
Nice! This looks right to me. I noticed that the version resolvers ([v1](https://github.com/dependabot/dependabot-core/blob/4ff425713c8de7fd8c136f8e0414e751e2cc8029/bundler/helpers/v1/lib/functions/version_resolver.rb#L87), [v2](https://github.com/dependabot/dependabot-core/blob/4ff425713c8de7fd8c136f8e0414e751e2cc8029/bundler/helpers/v2/lib/functions/version_resolver.rb#L87)) also use the lockfile parser. You may want to consider using the cached lockfile parser there too

PR #9454, opened 2024-04-08
Handle more branch protection errors

jurre:
As a follow-up to #9439, I noticed a couple more exceptions around branch protection rules, which I've listed here.

I am not sure that adding a new test for each permutation of this error is of a lot of value, but open to feedback on this

### DIFF
--- a/common/lib/dependabot/pull_request_updater/github.rb
+++ b/common/lib/dependabot/pull_request_updater/github.rb
@@ -213,17 +213,20 @@ module Dependabot
         )
       end
 
-      BRANCH_PROTECTION_ERROR_MESSAGES = [
-        /protected branch/i,
-        /not authorized to push/i,
-        /must not contain merge commits/i,
-        /required status check/i,
-        /cannot force-push to this branch/i,
-        /pull request for this branch has been added to a merge queue/i,
-        # Unverified commits can be present when PR contains commits from other authors
-        /commits must have verified signatures/i,
-        /changes must be made through a pull request/i,
-      ]
+      BRANCH_PROTECTION_ERROR_MESSAGES = T.let(
+        [
+          /protected branch/i,
+          /not authorized to push/i,
+          /must not contain merge commits/i,
+          /required status check/i,
+          /cannot force-push to this branch/i,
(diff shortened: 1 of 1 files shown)


jeffwidman (approved):

common/lib/dependabot/pull_request_updater/github.rb
```diff
@@ -226,13 +241,7 @@ def update_branch(commit)
...
         return nil if e.message.match?(/Reference cannot be updated/i)
 
-        if e.message.match?(/protected branch/i) ||
-           e.message.match?(/not authorized to push/i) ||
-           e.message.include?("must not contain merge commits") ||
-           e.message.match?(/required status check/i) ||
-           e.message.match?(/cannot force-push to this branch/i)
-          raise BranchProtected
-        end
+        raise BranchProtected if BRANCH_PROTECTION_ERROR_MESSAGES.any? { |msg| e.message.match?(msg) }
```
landongrindheim, line 244: 🔎 What do you think of including the message in `BranchProtected`? I think there's a strong value add since a couple of the messages we're handling are related to protections that are more recent (merge queue usage, repository rules).

```suggestion
        raise BranchProtected, e.message if BRANCH_PROTECTION_ERROR_MESSAGES.any? { |msg| e.message.match?(msg) }
```

landongrindheim (approved):

common/lib/dependabot/pull_request_updater/github.rb
jurre: In general I'm a fan of that, but here specifically we just use it as a signal in the code that runs this (api in our case) to perform a different action. We're essentially using errors for control flow 🙄 so I'm not sure we'd benefit?

landongrindheim: I trust your understanding 😄 If the running code does anything with the signal (reporting to an error tracker), it'd be great to have the extra context. If not, no need.

jurre: It does not, but I suppose there's no harm in adding it either

PR #9458, opened 2024-04-09
bundler: avoid reparsing same gemspec

jakecoffman:
This is a slight performance improvement that will help with grouped updates. 

I noticed in a flamegraph that `gemspec_dependencies` is called multiple times for each call of `parse`, so memoizing the call improves the performance.


jurre:
For grouped updates, we'll create a new FileParser after every update right, so this will just prevent re-work during the update of a single dependency?

jakecoffman:
@jurre yep, exactly

jurre (approved):

PR #9462, opened 2024-04-09
Update how we record Bundler versions

landongrindheim:
We rely on this information for metrics, not application logic. I believe we're receiving poor information and making decisions based on it. I'm updating this to see if we're indeed supporting as many Bundler v1 projects as it seems.

### DIFF
--- a/bundler/lib/dependabot/bundler/helpers.rb
+++ b/bundler/lib/dependabot/bundler/helpers.rb
@@ -30,9 +30,9 @@ module Dependabot
         return "unknown" unless lockfile
 
         if (matches = lockfile.content.match(BUNDLER_MAJOR_VERSION_REGEX))
-          matches[:version]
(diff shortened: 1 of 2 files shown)


jakecoffman (approved):

bundler/lib/dependabot/bundler/helpers.rb
```diff
@@ -30,9 +30,9 @@ def self.detected_bundler_version(lockfile)
         return "unknown" unless lockfile
 
         if (matches = lockfile.content.match(BUNDLER_MAJOR_VERSION_REGEX))
-          matches[:version]
+          matches[:version].to_i
         else
-          "1"
+          "unknown"
```
jakecoffman, line 35: Maybe to differentiate from having no lockfile?

```suggestion
          "unspecified"
```

PR #9466, opened 2024-04-09
replace credentials with dummy creds

jakecoffman:
Customers are setting `npmAlwaysAuth` in their `.yarmrc.yml` which due to our sanitation results in `No authentication configured for request`, even if they set a fallback value with `${TOKEN:-fallback}` because we clear out the entire thing.

This works around that by setting a default value ourselves.

For the life of me I cannot reproduce this issue locally, but it should be completely harmless assuming all the tests pass.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
@@ -372,8 +372,9 @@ module Dependabot
         end
 
         def sanitize_yarnrc_content(content)
-          # Replace all "${...}" and ${...} occurrences with empty strings
-          content.gsub(/\"\$\{.*?\}\"/, '""').gsub(/\$\{.*?\}/, '""')
+          # Replace all "${...}" and ${...} occurrences with dummy strings. We use
+          # dummy strings instead of empty strings to prevent issues with npmAlwaysAuth
+          content.gsub(/"\$\{.*?}"/, '"DUMMYCREDS"').gsub(/\$\{.*?}/, '"DUMMYCREDS"')
         end
 
         def clean_npmrc_in_path(yarn_lock)


jakecoffman:
I verified this does fix the issue.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
```diff
@@ -372,8 +372,9 @@ def write_temporary_dependency_files(yarn_lock, update_package_json: true)
         end
 
         def sanitize_yarnrc_content(content)
-          # Replace all "${...}" and ${...} occurrences with empty strings
-          content.gsub(/\"\$\{.*?\}\"/, '""').gsub(/\$\{.*?\}/, '""')
+          # Replace all "${...}" and ${...} occurrences with dummy strings. We use
+          # dummy strings instead of empty strings to prevent issues with npmAlwaysAuth
+          content.gsub(/"\$\{.*?}"/, '"DUMMYCREDS"').gsub(/\$\{.*?}/, '"DUMMYCREDS"')
```
mitchellroberts11, line 377: I think this regex is no longer catching the closing brace. We're getting this error:

> updater | 2024/04/11 00:22:38 ERROR <job_812946518> bad URI(is not URI?): "https://our-npm-registry.us.online/registry}/api/npm/our-npm/express"

mitchellroberts11: I'm not particularly familiar with regex sorry, but I think it might be a ruby regex thing where `}` still needs to be escaped.

I think if the regex check is just reverted to the fully escaped version it will be fixed.

```suggestion
          content.gsub(/\"\$\{.*?\}\"/, '"DUMMYCREDS"').gsub(/\$\{.*?\}/, '"DUMMYCREDS"')
```

jakecoffman: I can't see how this PR would have caused that as job 812946518 is failing during the UpdateChecker step which happens before the FileUpdater step where I made this change.

mitchellroberts11: Yep, looks like that was an issue on our end sorry.

Our repo was trying to do some odd conditional assignment in a `.yarnrc.yml`:

```
npmRegistryServer: '${REMOTE_URL:-https://our.url.us.online/remote}/api/npm/our-npm/'
 ```

This fixed it:
```
npmRegistryServer: "https://our.url.us.online/remote/api/npm/our-npm/"
 ```

I'll look up the yml syntax guide in the morning and figure out what we were doing wrong. Thanks for your help, greatly appreciated.

apgrucza: The [GitHub documentation](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#npm-registry) says this:

> When accessing environment variables in your `.yarnrc.yml` file, you should provide a fallback value such as `${ENV_VAR-fallback}` or `${ENV_VAR:-fallback}.`

But with this change, the fallback value provided is no longer used because it's stripped out by `sanitize_yarnrc_content` (resulting in the above problem with `npmRegistryServer`). Is this intentional? If so, then the GitHub documentation should be updated because it gives the impression that you can provide a fallback value, when in fact the fallback is always `DUMMYCREDS`.

jakecoffman: I don't think it was originally intended when it was added in #9322, but if it simplifies the docs then I think it's a win.

@pavera I see you added the blurb about adding the fallback value. Do you recall it done because of `npmAlwaysAuth`?

PR #9471, opened 2024-04-10
Avoid passing empty strings as versions, part 2

### LINKED ISSUE #9430: ArgumentError: blank strings must not be provided as versions
```
ArgumentError: blank strings must not be provided as versions
  common/lib/dependabot/dependency.rb:357:in `check_values'
  dependabot-updater/vendor/ruby/3.1.0/gems/sorbet-runtime-0.5.11193/lib/types/private/methods/call_validation_2_7.rb:652:in `bind_call'
  dependabot-updater/vendor/ruby/3.1.0/gems/sorbet-runtime-0.5.11193/lib/types/private/methods/call_validation_2_7.rb:652:in `block in create_validator_procedure_fast0'
  common/lib/dependabot/dependency.rb:129:in `initialize'
  dependabot-updater/vendor/ruby/3.1.0/gems/sorbet-runtime-0.5.11193/lib/types/private/methods/call_validation.rb:169:in `bind_call'
...
(26 additional frame(s) were not displayed)
```

bdragon:
The issue described/addressed in #9461 is still causing runtime exceptions. Instead of dancing around the issue by trying to prevent empty string versions everywhere, this change updates `Dependabot::Dependency` by initializing its `version`/`previous_version` to nil if the given value for either resolves to an empty string.

Fixes #9430


jurre (approved):

common/lib/dependabot/update_checkers/base.rb
```diff
@@ -241,7 +241,7 @@ def updated_dependency_without_unlock
       sig { returns(Dependabot::Dependency) }
       def updated_dependency_with_own_req_unlock
         version = preferred_resolvable_version.to_s
-        previous_version = latest_resolvable_previous_version(version)&.to_s
+        previous_version = latest_resolvable_previous_version(version)
```
jurre, line 244: I'm actually a little surprised, because:

```ruby
irb(main):001:0> nil&.to_s
=> nil
irb(main):002:0> nil.to_s
=> ""
```

bdragon: Thanks @jurre I'm also a bit stumped... This would suggest that `dependency.version` is returning `""`, but `Dependency` will raise if you try to create one with `previous_version = "", so I'm not sure how we could get into this state.

bdragon: My first thought was this is a base class so maybe a subclass is overriding `latest_resolvable_previous_version` and returning an empty string, but the only ecosystem that overrides it is npm and we're seeing this error for composer, nuget, and others.

jurre: It looks like you can create a `Dependabot::Dependency` with a version of nil, then we call `to_s` on that?

https://github.com/dependabot/dependabot-core/blob/0ec3fe82cdc08567755ba023267c1b4fe062dde1/common/lib/dependabot/version.rb#L17

So we end up with it being `""` through that path?

bdragon: Yep, I think that's it!

bdragon: Ecosystems implement their own version classes and it seems most call `#to_s` on the version argument, plus `Gem::Version` is going to call `#to_s` anyways, so we'd need to override `Version#to_s` and `Version#version` and methods on ecosystem version classes that return the stringified version.

Instead of going down that rabbit hole, I think the most sensible thing to do is, instead of raising, just have `Dependency` set its `version` or `previous_version` to nil if either is an empty string.

PR #9472, opened 2024-04-10
5856: nokogiri stack level is too deep error

thavaahariharangit:


thavaahariharangit:
Need work : Latest can be an expression, But that expression should not be pointing to self. I am working on this.

maven/lib/dependabot/maven/file_parser/property_value_finder.rb
```diff
@@ -69,6 +73,16 @@ def property_details(property_name:, callsite_pom:)
...
         attr_reader :dependency_files
 
+        def extract_value_from_expression(expression_value:, property_name:, callsite_pom:)
+          # and the expression is pointing to self then raise the error
+          if expression_value.eql?("${#{property_name}}")
+            raise Dependabot::DependencyFileNotParseable, "[ERROR] Resolving expression: '#{expression_value}'"
+          end
+
+          # and the expression is pointing to another tag, then get the value of that tag
+          property_details(property_name: expression_value.slice!(2..-2), callsite_pom: callsite_pom)
```
jurre, line 83:
```suggestion
          property_details(property_name: expression_value.slice(2..-2), callsite_pom: callsite_pom)
```

`slice!` with a `!` will mutate the value in place, there is no need to mutate the value in place so we can use the non-mutating form. While in this case it won't change the behavior of the code, in general it's considered good practice to use non-mutating forms when possible, as mutating data can be a source of bugs.

thavaahariharangit:
Features added,
1. and in case of duplicate tags then read the latest.
2. and the latest tag is pointing to self then raise the error.
3. and the latest tag is pointing to another tag, then get the value of that tag.

maven/lib/dependabot/maven/file_parser/property_value_finder.rb
```diff
@@ -69,6 +73,16 @@ def property_details(property_name:, callsite_pom:)
 
         attr_reader :dependency_files
 
+        def extract_value_from_expression(expression_value:, property_name:, callsite_pom:)
+          # and the expression is pointing to self then raise the error
+          if expression_value.eql?("${#{property_name}}")
+            raise Dependabot::DependencyFileNotParseable, "[ERROR] Resolving expression: '#{expression_value}'"
```
jurre, line 79: We don't need to preface the message with `[ERROR]`, the logging code will already make sure to format it in a way where that's clear, also, at this point we know that it is a recursive cycle I think so maybe we can say that, something like:

```suggestion
            raise Dependabot::DependencyFileNotParseable, "Error trying to resolve recursive expression '#{expression_value}'."
```

thavaahariharangit:
Problem : 
mvn build was getting success and dependabot build is failing with message : 'nokogiri stack level is too deep error:'
Found the root cause: Duplicate properties tags are present in given pom file. latest tag having the expected value dependabot reading the first tag which is having an expression pointing to latest.

Solution provided: 
1. Updated dependabot to always read the latest.
2. Updated if the latest tag value is an expression and pointing to self throw meaningful error
3. Update if the latest tag value is an expression and pointing to another tag, then get that value.

Note to the reviewer:
I have added new fixture pom.file, 3 additional tests each checking above 3 scenarios respectively.

maven/lib/dependabot/maven/file_parser/property_value_finder.rb
```diff
@@ -63,6 +73,19 @@ def property_details(property_name:, callsite_pom:)
...
+          # and the expression is pointing to self then raise the error
+          if expression_value.eql?("${#{property_name}}")
+            raise Dependabot::DependencyFileNotParseable.new(
+              callsite_pom.name,
+              "Error trying to resolve recursive expression '#{expression_value}'."
+            )
+          end
+
+          # and the expression is pointing to another tag, then get the value of that tag
+          property_details(property_name: expression_value.slice(2..-2), callsite_pom: callsite_pom)
```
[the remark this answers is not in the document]
thavaahariharangit: I agree we need handle that scenario.
But as per this issue, customer is saying that maven is not giving any build error but dependabot does, That's the main focus here.  (So I fixed that error).

@bdragon and @jurre 
So I will raise a separate PR for the Malformed expression, does that sounds good

jurre (approved):

abdulapopoola:
How is the error being surfaced to our users?

abdulapopoola:
@thavaahariharangit ; I assume there will be a new error class raised in this PR; that new error will need to be handled in API. [See this PR](https://github.com/github/dependabot-api/pull/4535)

thavaahariharangit:
> @thavaahariharangit ; I assume there will be a new error class raised in this PR; that new error will need to be handled in API. [See this PR](https://github.com/github/dependabot-api/pull/4535).
> 
> Nevermind, you are not raising a new error but can you please double check and confirm that the `DependencyFileNotParseable` error actually shows up in the Dependabot Logs page once you test and deploy it?

@abdulapopoola  sure

PR #9474, opened 2024-04-10
Actions: allow less precise semver updates

jakecoffman:
If an Action is pinned to v3.5, and a v4 and v4.1.0 are released, Dependabot won't bump the Action.

This is because it tries to match the same precision. So v3.5 is precision 2, v4 is precision 1, v4.1.0 is precision 3.

It makes sense to not allow a v4 -> v4.1.0 bump because in the Action world, v4 and v4.1.0 SHOULD be pointed to the same commit.

However in the case above, v3.5 -> v4 does make sense, but we'd prefer a v4.X if available.

So this code will prefer a matching precision, but if one doesn't exist it will become less precise.


pavera (approved):
LGTM.

PR #9484, opened 2024-04-11
faster bundler lockfile parser

jakecoffman:
I extracted a simplified portion of how Bundler parses lockfiles. This should improve performance for customers with very large lockfiles. The performance improvements I made are:
- It only parses specs
- It only cares about the name because we were doing `.map { |x| x.name.to_s }` anyway
- It reads the file line-by-line which should reduce GC pressure
- Using a set means we don't have to uniq

Testing against a very large lockfile I see about 1 minute saved per update.

To verify that the parsing is correct, I put our updater lockfile into a fixture and compared that to what we were doing before and it matches.


jakecoffman:
Yeah good point. This is why it's always a good choice to go with a well-known format like json or yaml.

Nothing here can be upstreamed it hinged on the fact we only needed one bit of data from the file.

jeffwidman:
@deivid-rodriguez I just wonder if any of this could be upstreamed to have a "parse-specs-only" mode flag for bundler that is 🏎️ ? You'd know better than any of us the intersection of `bundler` + :dependabot: and what's actually sustainably maintainable long term?

deivid-rodriguez:
@jeffwidman I think the approach here would be mostly ok. It is true that relying on Bundler is ideal but the lockfile format does not change often (I don't recall any change ever in the format that would make parsing specs like done here not work).

It could potentially also be upstreamed, but it seems like a very specific use case to me and not straightforward to upstream?

I'm surprised though the parsing lockfiles is so costly, perhaps there's some performance issue upstream? Or maybe the same lockfile is being parsed over and over unnecessarily?

PR #9485, opened 2024-04-12
asynchronously create PRs

jakecoffman:
The PR creation process requires gathering a significant amount of metadata which is IO heavy. Rather than wait for that to finish, we can pop that in a thread and continue on with the next updates. 

Testing locally against dependabot-core this lead to a 30% reduction in the job runtime, but it's very dependent on if there's a lot of PRs to create.

The major downside of this is that the logs become more difficult to read because the metadata gathering overlaps with the output for the next dependency, so I've put this behind an experiment and we can enable it on customers that are hitting the timeout.

### DIFF
--- a/updater/lib/dependabot/service.rb
+++ b/updater/lib/dependabot/service.rb
@@ -31,6 +31,7 @@ module Dependabot
       @client = client
       @pull_requests = T.let([], T::Array[T.untyped])
       @errors = T.let([], T::Array[T.untyped])
+      @threads = T.let([], T::Array[T.untyped])
     end
 
     def_delegators :client,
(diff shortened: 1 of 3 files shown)


jakecoffman:
@jeffwidman that might work, or maybe it's not as big of a deal as I think. If we could make the logs clear to read while doing this is would be great to roll out to all updates.

jeffwidman:
Yeah, I've leveraged the thread prefix idea before in both Python and Go very effectively for debugging async bugs in other projects, and every time it was _so_ helpful, especially since I could just grep through the logfile to jump past all the noise when following one specific thread...

PR #9487, opened 2024-04-12
fix smoke tests not running

jakecoffman:
This branch doesn't exist so the smoke tests weren't running.


landongrindheim (approved):

Nishnha:
Sorry! I thought I had fixed this in https://github.com/dependabot/dependabot-core/pull/9473 but it seems like after the base changed back to main https://github.com/dependabot/dependabot-core/pull/9473/commits/05992c0f7d1ad73057a86ed03fba402ea0ae5fa5 was dropped from the PR?

Not sure how that happened though

PR #9513, opened 2024-04-17
Harry/sorbet 9512

thavaahariharangit:


jurre:
Why'd you open up a new PR? Would be easier to just push that commit to Dependabot's PR, that way we get all of the information in the PR description etc

PR #9517, opened 2024-04-17
hh170424: Fixing the sorbet type errors.

thavaahariharangit:
Fixed Sorbet errors

### DIFF
--- a/common/lib/dependabot/pull_request_creator/codecommit.rb
+++ b/common/lib/dependabot/pull_request_creator/codecommit.rb
@@ -202,7 +202,7 @@ module Dependabot
 
         codecommit_client_for_source.create_commit(
           branch_name,
-          author,
+          author.to_s,
           base_commit,
           commit_message,
           files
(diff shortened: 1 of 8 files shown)


common/lib/dependabot/pull_request_creator/codecommit.rb
```diff
@@ -202,7 +202,7 @@ def create_commit
 
         codecommit_client_for_source.create_commit(
           branch_name,
-          author,
+          author.to_s,
```
jurre, line 205: This change would result in something like this being passed:

```
{ author: "jurre", email: "jurre@example.com", date: "2024-04-17" }.to_s
=> "{:author=>\"jurre\", :email=>\"jurre@example.com\", :date=>\"2024-04-17\"}"
```

Or, if it's `nil` it would pass `nil.to_s` which will be an empty string.

That's probably not the intent, it might be better to change the interface to `T.nilable(String)` and then pick off the first value from the hash if it's not nil, so something like:

```ruby
author = author_details&.slice(:name, :email, :date)&.values&.first
```

That will return a nillable String, so we can then get rid of the `author = nil unless author&.any?` line

PR #9519, opened 2024-04-17
add functionality to ignore comments in .python-version file and fetch minimum Python version successfully

### LINKED ISSUE #6650: Minimum Python version not detected if .python-version contains comments

### Package ecosystem

pip

### Package manager version

_No response_

### Language version

_No response_

https://github.com/canonical/pycloudlib/blob/b02ba24b214ddf922491ee711b4ffe382c450da7/dev-requirements.txt

### dependabot.yml content

https://github.com/canonical/pycloudlib/blob/b02ba24b214ddf922491ee711b4ffe382c450da7/.github/dependabot.yml

### Updated dependency

_No response_

I expected `dependabot` to correctly detect the minimum Python version defined in `.python-version` and issue upgrades with correct python dependencies' versions.

If `.python-version` contains comments, then the suggested dependecies' versions were always the newest one.

Not working configuration: https://github.com/canonical/pycloudlib/blob/b02ba24b214ddf922491ee711b4ffe382c450da7/

Working configuration: https://github.com/canonical/pycloudlib/pull/275

_No response_

This PR tried to bump `flake8` to a version that was not compatible with python 3.6 (defined in `.python-version`):

https://github.com/canonical/pycloudlib/pull/274

`dependabot.yml`:

```yaml
version: 2
updates:
  - package-ecosystem: "pip"
    directory: "/"
```

`.python-version`:

```
3.6.5
# Specifies the python version that dependabot should use
# Note: The version must specify major.minor.patch components
# and be supported on:
# https://github.com/dependabot/dependabot-core/blob/7822f07e76cab9be6eca86a91aafc8d72bc42536/python/lib/dependabot/python/python_versions.rb
```

mohammadalizabihitari:
Fixes #6650 
Adds functionality to ignore comments in .python-version file, fetches the minimum Python version and unit test it.

### DIFF
--- a/python/lib/dependabot/python/file_parser/python_requirement_parser.rb
+++ b/python/lib/dependabot/python/file_parser/python_requirement_parser.rb
@@ -83,15 +83,12 @@ module Dependabot
         def python_version_file_version
           return unless python_version_file
 
-          # read the content, split into lines and remove any lines starting with '#'
-          content_lines = python_version_file.content.each_line.reject do |line|
-            line.strip.start_with?('#')
-          end
+          # read the content, split into lines and remove any lines with '#'
+          content_lines = python_version_file.content.each_line.map do |line|
+            line.sub(/#.*$/, " ").strip
+          end.reject(&:empty?)
 
-          # file_version = python_version_file.content.strip
-
-          # join the cleaned lines back into a single string and strip whitesapces
-          file_version = content_lines.join.strip
+          file_version = content_lines.first
           return if file_version&.empty?
           return unless pyenv_versions.include?("#{file_version}\n")
 
--- a/python/lib/dependabot/python/file_parser/python_requirement_parser.rb
+++ b/python/lib/dependabot/python/file_parser/python_requirement_parser.rb
@@ -83,7 +83,15 @@ module Dependabot
         def python_version_file_version
           return unless python_version_file
 
-          file_version = python_version_file.content.strip
+          # read the content, split into lines and remove any lines starting with '#'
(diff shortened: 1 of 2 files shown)


python/lib/dependabot/python/file_parser/python_requirement_parser.rb
```diff
@@ -83,7 +83,15 @@ def pip_compile_python_requirement
         def python_version_file_version
           return unless python_version_file
 
-          file_version = python_version_file.content.strip
+          # read the content, split into lines and remove any lines starting with '#'
+          content_lines = python_version_file.content.each_line.reject do |line|
+            line.strip.start_with?('#')
+          end
```
jurre, line 89: What if someone did this:

```
3.6.2 # We're using this version of python because of reasons
```

python/lib/dependabot/python/file_parser/python_requirement_parser.rb
```diff
@@ -83,7 +83,12 @@ def pip_compile_python_requirement
         def python_version_file_version
           return unless python_version_file
 
-          file_version = python_version_file.content.strip
+          # read the content, split into lines and remove any lines with '#'
+          content_lines = python_version_file.content.each_line.map do |line|
```
jurre, line 87: I don't think we need to call `.map` on each_line, it already takes a block that gives us a string, so this could be simplified to:

```suggestion
          content_lines = python_version_file.content.each_line do |line|
```

jurre (approved):

PR #9530, opened 2024-04-18
Harry/#9512 b

thavaahariharangit:
Ran bundle install in local


jurre:
@thavaahariharangit can you please push these changes to the original dependabot PR so that we have all the context about the dependencies being updated handy?

thavaahariharangit:
> @thavaahariharangit can you please push these changes to the original dependabot PR so that we have all the context about the dependencies being updated handy?

Sure

PR #9538, opened 2024-04-19
Maven: Handle Malformed expressions or Invalid values present in properties section of the pom.xml file

thavaahariharangit:
Problem:
How to handle the Malformed expression?

Solution provided : 
As per the maven documentation, properties can contain any values. So if it is an expression then we should treat it as an expression and find the value of it, else we need to treat that as regular value.

Property value can accept
1. Expression ${exp}
2. Booleans
3. Version numbers in any format
4. Date Time formats and etc.

Ref : 
https://maven.apache.org/pom.html#Properties
https://maven.apache.org/guides/introduction/introduction-to-the-pom.html

### DIFF
--- a/maven/lib/dependabot/maven/file_parser/property_value_finder.rb
+++ b/maven/lib/dependabot/maven/file_parser/property_value_finder.rb
@@ -49,7 +49,7 @@ module Dependabot
             end
 
           # and value is an expression
-          if node && node.content.strip.start_with?("${")
+          if node && /\$\{(?<expression>.+)\}/.match(node.content.strip)
             return extract_value_from_expression(
               expression: node.content.strip,
               property_name: property_name,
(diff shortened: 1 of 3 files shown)


jurre:
Maven itself seems to raise an error when finding an invalid expression, should we maybe do the same? The downside of doing it is that there might be some repo's where Dependabot can currently open PRs with updates whereas when we start raising an error, we won't. But on the other hand, since those project's won't be valid for Maven anyway, does it really matter?

abdulapopoola (approved):

thavaahariharangit:
> Maven itself seems to raise an error when finding an invalid expression, should we maybe do the same? The downside of doing it is that there might be some repo's where Dependabot can currently open PRs with updates whereas when we start raising an error, we won't. But on the other hand, since those project's won't be valid for Maven anyway, does it really matter?

If it is lombok.version, then maven is identifying lombok.version by going checking the lombok version. so the error from maven is `ERROR]   The project com.dependabot.testing:dependabot-demo:0.0.1-SNAPSHOT (/Users/hariharanthavachelvam/Documents/Sites/Demo/dependabot-demo/pom.xml) has 1 error
[ERROR]     'dependencies.dependency.version' for org.projectlombok:lombok:jar must be a valid version but is '${tag.version'. @ line 35, column 13`

but when I use simply tag.version(totally custom tag then ) `<tag.version>${tag.version</tag.version>` even the maven build is succeeding. 

So how maven validating is checking that tag name and looking available version on their repository.

But in general term Properties are allowed to have any values. 

That is why I left as it is.

PR #9542, opened 2024-04-19
allow updating package without assemblies from `packages.config`

brettfo:
Some packages (e.g., `jQuery`) don't contain assemblies, but if they're in a `packages.config` scenario, that can pose 2 issues with updating.

1. We can't determine the local packages directory.
2. The update process will fail unless `restore` has been run first.

To correct the first issue, we fall back to a relative packages path of `../packages` which is by far the most common value used.  Since the given package has no assemblies, this fall back value won't be written back to the `.csproj`; it's just temporary.

To correct the second issue it was just a case of expanding our detection logic of when we need to run `restore` before retrying `update`.

Part of this fix uncovered an issue where NuGet package are sometimes checked for a `content` directory with a case-sensitive search.  The test package used, however (`jQuery`), names that directory `Content` which doesn't match.  The fix was to copy a single file from the submodule and make the check case-insensitive.

### DIFF
--- a/nuget/helpers/lib/NuGetUpdater/NuGetProjects/NuGet.Packaging/PackageFolderReader.cs
+++ b/nuget/helpers/lib/NuGetUpdater/NuGetProjects/NuGet.Packaging/PackageFolderReader.cs
@@ -1,0 +1,270 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+using NuGet.Common;
+using NuGet.Frameworks;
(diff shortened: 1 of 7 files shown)


JoeRobich (approved):

nuget/helpers/lib/NuGetUpdater/NuGetProjects/NuGet.Packaging/PackageFolderReader.cs
```diff
@@ -0,0 +1,269 @@
...
+                // For sub folders verify it exists
+                // The root is expected to exist and should throw if it does not
+                getFiles = searchFolder.Exists;
+
+                // try a case-insensitive search
+                if (!getFiles)
+                {
+                    searchFolder = _root.GetDirectories().FirstOrDefault(d => d.Name.Equals(folder, StringComparison.OrdinalIgnoreCase));
+                    getFiles = searchFolder?.Exists == true;
+                }
```
brettfo, line 134: This was the case-insensitive patch that needed to be applied.

thavaahariharangit (approved):

Issue #9203, opened 2024-03-04
Measure test coverage


JamieMagee:
### Code improvement description

Test coverage is a simple measure of how much of the codebase has been tested. I think that setting specific coverage levels is not very helpful beyond a certain point, but it can help identify parts of the codebase that are poorly tested and may require more attention.

The most popular library for measuring code coverage in Ruby is [simplecov][1].

[1]: https://github.com/simplecov-ruby/simplecov

yeikel:
Can you assign this to me?

JamieMagee:
@yeikel done. Thank you!

smoookeeey:
Can you assign this to me?

JamieMagee:
@smoookeeey I assigned you too, but @yeikel asked to be assigned first. Please make sure you don't duplicate any work @yeikel might have already done.

smoookeeey:
Yes I will make sure not too do that

yeikel:
I haven't started yet, and I won't have capacity until sometime next week. @smoookeeey, if you pick it up before that, feel free to create the PR and link it here

smoookeeey:
I will look into it brother

abdulapopoola:
Thanks a lot @yeikel and @smoookeeey !

JamieMagee:
I just discovered that test coverage was previously added in #2576 to prevent a regression. Originally line coverage was set to 80% and branch coverage to 70%, with `refuse_coverage_drop` enabled to prevent any drops from occuring.

It was later gated behind the `COVERAGE` environment variable in #3339. However, because that environment variable wasn't set in GitHub Actions, code coverage was allowed to deteriorate. Subsequently, test coverage was removed in #7473.

yeikel:
I think that rather than enforcing certain %, we can start reporting it for awareness

JamieMagee:
I agree. Reporting-only is a good start. Then something like [`refuse_coverage_drop`][1] to prevent backslides and act like [a ratchet][2]

[1]: https://www.rubydoc.info/gems/simplecov/SimpleCov%2FConfiguration:refuse_coverage_drop
[2]: https://qntm.org/ratchet

yeikel:
@smoookeeey Are you still working on this? Feel free to create a draft PR whenever you're ready

JamieMagee:
@yeikel I think you can go ahead with this if you want.

PR #9595, opened 2024-04-24
Add `simplecov` to track test coverage

JamieMagee:
This is an initial addition of `simplecov` to Dependabot. Currently no level of coverage is enforced, but a report is generated.

From here, we can at least track coverage, and decide how and at what level we want to enforce code coverage.

Related to #9203


jurre:
I think currently this will dump a report to stdout for each test run, but since the testsuite is split up heavily it will be hard to get an overall sense of where we stand? It would be nice to think about a way we can track coverage across the entire codebase

JamieMagee:
It looks like we'd need a final step after all the unit test runs to collate the results. There are [some instructions in the simplecov documentation][1], but I'd need to investigate more on this.

[1]: https://github.com/simplecov-ruby/simplecov#merging-test-runs-under-different-execution-environments

yeikel:
> It looks like we'd need a final step after all the unit test runs to collate the results. There are [some instructions in the simplecov documentation][1], but I'd need to investigate more on this.
> 
> [1]: https://github.com/simplecov-ruby/simplecov#merging-test-runs-under-different-execution-environments

Relevant article that might help 

https://rubyyagi.com/merge-simplecov-circleci/

common/spec/spec_helper.rb
```diff
@@ -1,14 +1,32 @@
...
 
+# SimpleCov _must_ be started before any dependabot code is loaded
+SimpleCov.start do
+  add_filter "/spec/"
+  if ENV["CI"]
+    formatter SimpleCov::Formatter::SimpleFormatter
+  else
+    formatter SimpleCov::Formatter::MultiFormatter.new([
+      SimpleCov::Formatter::SimpleFormatter,
+      SimpleCov::Formatter::HTMLFormatter
```
yeikel, line 22: I understand the sentiment here, but I think that we should still generate the HTML report during the CI build and consider adding them as a build artifact for easy access

yeikel:
Also, I was wondering if we could see if there is any integration we could add to include this as a pull request comment with a link to the HTML report as that will be helpful to raise awareness. Can we leverage Github pages for this?

JamieMagee:
> Also, I was wondering if we could see if there is any integration we could add to include this as a pull request comment with a link to the HTML report as that will be helpful to raise awareness. Can we leverage Github pages for this?

I'm not sure about GitHub pages, but I have seen the [`joshmfrankel/simplecov-check-action`][1] and [Codecov supports simplecov] as well. I'll need to look more into it.

[1]: https://github.com/joshmfrankel/simplecov-check-action
[2]: https://about.codecov.io/tool/simplecov/

common/spec/spec_helper.rb
JamieMagee: Fair point. I'll change that and use `actions/updload-artifact`.

JamieMagee: @yeikel looks like this was merged. I'll leave #9203 open and handle it in a follow-up PR.

PR #9597, opened 2024-04-24
Bump Ruby to 3.3.1

jurre:
This PR upgrades Ruby from 3.1.4 to 3.3.1 and Rubygems from 2.3.1 to 2.5.9.

Changelogs:
- [Ruby](https://www.ruby-lang.org/en/news/2023/12/25/ruby-3-3-0-released/)
- [Rubygems](https://github.com/rubygems/rubygems/blob/master/bundler/CHANGELOG.md#259-april-12-2024)

These releases bring a lot of exciting performance improvements, including the ability to enable Ruby's YJIT, which seems promising, even for the relatively short lifetime of our jobs.

## Bundler v1
We’d been blocked on adopting Ruby 3.2+ for quite some time, so let me address what I’ve done to unblock us from that.

In Ruby 3.2, support for `Object#untaint` was removed. [Here](https://bugs.ruby-lang.org/issues/16131) is some context on that, but the important bit for this PR is, that method has been a no-op since Ruby 2.7. It hasn’t done anything for years.

The initial versions of Bundler 2 would ship with a special case to [omit the method on newer versions of Ruby](https://github.com/rubygems/bundler/blob/32a4159325bc181b19e33c3806ea05430f4bbbc7/lib/bundler/shared_helpers.rb#L16).

So, it seems safe to say that we do not care about `untaint` in the context of bundler.

What I’ve done here to make it work is include a patch onto `Object` to stub out `#untaint` and load it into Bundler v1, both in our native helpers and when we run `bundle install`, by wrapping the command in a new script.

## Native helper specs

In https://github.com/rubygems/rubygems/pull/6793 `Net::HTTP` was vendored and the namespace changed to `::Gem::Net::HTTP`. This caused our native helpers to no longer have network requests intercepted by WebMock. I've added a copy of WebMock's `Net::HTTP` adapter for the new namespace.

One thing of note is I've found our bundler v1 specs run against the latest Bundler version. Production code and tests from dependabot-core _do_ use Bundler 1.17 in that context, since `bundler/helpers/v1/run.rb` loads it via `gem "bundler", "~> 1.17"`. Our native helper specs are loaded via `BUNDLER_VERSION=1.17.3 bundle exec rspec` and this seems to _not_ load that version of bundler. This is surprising to me and something we might address, but maybe we'll remove support for Bundler 1 before we care about this.


JamieMagee:
FYI Ruby 3.3.1 was recently released: https://www.ruby-lang.org/en/news/2024/04/23/ruby-3-3-1-released/

JamieMagee:
I opened an issue upstream about the new Sorbet warning with `bigdecimal`: https://github.com/sorbet/sorbet/issues/7854

jurre:
Something slightly concerning that I found while working through these test failures:

Our _tests_ for the helpers in `bundler/helpers/v1` run against whatever the is latest version of Bundler that's installed, and _not_ against `1.17.3`, but our production code and the tests that call the native helpers from dependabot-core _do_ use bundler `1.17.3`. So these native helper tests are not reliable right now.

The reason this happens is because we call `gem "bundler", "~> 1.17.3"` in `run.rb`, and our test setup bypasses this and uses `bundle exec rspec` to invoke the tests directly, bypassing the specific bundler setup we're using. We are setting `BUNDLER_VERSION=1.17.3` when executing these tests, but it does not appear to be altering the version of bundler the tests are running with.

I verified this by setting a debugger in this environment and grabbing `Bundler::VERSION` when running the tests natively, and I modified the helpers to write the bundler version to a file when running in integration mode.

I've verified that this happens both before and after the Ruby upgrade, so it's not being introduced in this PR, this has been the case for a long time, probably since the beginning or when we reshuffled how bundler is loaded in this setup.

jurre:
I believe the native tests are failing because WebMock is no longer able to stub the requests because of https://github.com/rubygems/rubygems/pull/6793 and so the tests are making real network requests now.

landongrindheim (approved):

deivid-rodriguez (commented):
I left a small comment.

Nice work on this upgrade, it seems very tricky, and I love that you managed to keep Bundler v1 support, allowing the decision to drop that to not block this upgrade! 💪

JamieMagee:
Is YJIT enabled by default in Ruby 3.3? The release notes only mention that it is enabled by default for Rails 7.2.

landongrindheim:
> Is YJIT enabled by default in Ruby 3.3? The release notes only mention that it is enabled by default for Rails 7.2.

[The docs](https://github.com/ruby/ruby/blob/master/doc/yjit/yjit.md#command-line-options) suggest it's disabled by default. I'll try to get this PR out this week, then once we're sure of its stability, we can enable YJIT 😄

JamieMagee:
@landongrindheim I created #9633 to track it.

bundler/helpers/v2/monkey_patches/definition_ruby_version_patch.rb
```diff
@@ -26,7 +26,7 @@ def source_requirements
         Gem::Specification.new("Ruby\0", requested_version)
     end
 
-    %w(2.5.3 2.6.10 2.7.7 3.0.5 3.2.1).each do |version|
+    %w(2.5.3 2.6.10 2.7.8 3.0.7 3.1.5 3.2.4).each do |version|
```
landongrindheim, line 29: @deivid-rodriguez We upgraded this list to include the latest Rubies. I _think_ this is what we're supposed to do, but I'm not 100% confident that I understand the monkey patch. Can you give a 👍/👎 re: whether this lines up with the intent?

bundler/helpers/v1/monkey_patches/definition_ruby_version_patch.rb
```diff
@@ -12,7 +12,7 @@ def index
           Gem::Specification.new("ruby\0", requested_version)
       end
 
-      %w(2.5.3p105 2.6.10p210 2.7.6p219 3.0.4p208).each do |version|
+      %w(2.5.3p105 2.6.10p210 2.7.6p219 3.0.4p208 3.1.4p223).each do |version|
```
landongrindheim, line 15: @deivid-rodriguez I've included the _current_ Ruby in this (Bundler v1) list. There's a more recent release (Ruby 3.1.5), but we're not running it yet. Does this list seem reasonable to you?

bundler/helpers/v2/monkey_patches/definition_ruby_version_patch.rb
deivid-rodriguez: Yes, it seems correct to me, the list needs to complete the list of "supported rubies". Since Dependabot now uses Ruby 3.3, it seems correct :+1:.

I think basically this is needed so that Dependabot also works even in situations where the application being updated is not compatible with the Ruby version Dependabot is using internally. I'd like to support this upstream, and we have [an issue](https://github.com/rubygems/rubygems/issues/7399) tracking it but not sure when/if it will happen.

deivid-rodriguez (commented):
:+1: :tada:

bundler/helpers/v1/monkey_patches/definition_ruby_version_patch.rb
deivid-rodriguez: I'm not sure if the patch part (`p105` for 2.5.3, etc) is needed or not. I suspect it was some Bundler 1 thing that made that necessary so just in case I'd keep them. Unfortunately I don't know of a good way of figuring out the patchlevel other than installing each ruby and checking the value of `RUBY_PATCHLEVEL`.

landongrindheim: We've had this out for about four hours now without issues. I'm going to merge it. I'll have to do a followup PR to add the patch part 😄 Thanks for all your input 🙇

landongrindheim: I've opened https://github.com/dependabot/dependabot-core/pull/9645 to add the patch levels 🙂

PR #9600, opened 2024-04-24
Ignoring the fixture files in workflow (dependency review and code QL) runs

thavaahariharangit:
Adding ignore helpet test fixtures to both code QL and dependency review jobs.

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -1,5 +1,12 @@
 name: Dependency Review
-on: [pull_request]  # yamllint disable-line rule:truthy
+on:
+  pull_request:
+    # yamllint disable-line rule:truthy
+    branches: [ main ]
+    paths-ignore:
+      - '*/spec/fixtures/**'
+      - '*/helpers/test/*/fixtures/**'
+      - 'common/lib/dependabot.rb'
 
 permissions:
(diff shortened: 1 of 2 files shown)


thavaahariharangit:
This PR is related to the https://github.com/dependabot/dependabot-core/pull/9562

There I found dependency-review.yml is running against the fixtures. And the job is failing.

in order to ensure the dependency-review.yml files to ignore the fixtures (test build files.) , I have added this changes.

Deploying and merging this will ensure that in future dependency-review.yml will not run against the fixtures.

abdulapopoola (approved):
Thanks

.github/workflows/dependency-review.yml
```diff
@@ -1,5 +1,12 @@
 name: Dependency Review
-on: [pull_request]  # yamllint disable-line rule:truthy
+on:
+  pull_request:
+    # yamllint disable-line rule:truthy
```
landongrindheim, line 4: 🔎 This will need to stay on the same line as `on:` in order to be honored by YAMLLint 🙂 

```suggestion
on: # yamllint disable-line rule:truthy
  pull_request:
```

thavaahariharangit: Thanks, and done

PR #9601, opened 2024-04-24
Store exceptions to be reported later

jurre:
Store exceptions to be reported later. This is useful for things like the message builder which runs in a separate process from the rest of the PR creation.

### DIFF
--- a/updater/spec/dependabot/update_files_command_spec.rb
+++ b/updater/spec/dependabot/update_files_command_spec.rb
@@ -4,6 +4,8 @@
 require "spec_helper"
 require "dependabot/exception_capturer"
 require "dependabot/update_files_command"
+require "dependabot/file_parsers"
+require "dependabot/bundler"
 require "tmpdir"
 
 RSpec.describe Dependabot::UpdateFilesCommand do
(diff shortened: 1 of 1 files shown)


updater/bin/update_files.rb
```diff
@@ -31,7 +32,12 @@ class UpdaterKilledError < StandardError; end
       Dependabot::Environment.job_id,
       Dependabot::Environment.job_token
     )
-  Dependabot::Service.new(client: api_client).capture_exception(error: error, tags: tags)
+  service = Dependabot::Service.new(client: api_client)
+
+  Dependabot::ExceptionCapturer.handle_captured_exceptions do |exception|
+    service.capture_exception(error: exception, tags: tags)
+  end
+  service.capture_exception(error: error, tags: tags)
```
jakecoffman, line 40: I think we want to put this down below 👇 We're eating the exception in the MessageBuilder so it won't reach this code in the trap.

What might be simpler would be give the ExceptionCapturer the service here (outside of the trap), and have it log to the API immediately when `capture_exception` is called.

jurre:
I'm going to close this out, feel free to revive if anyone needs this level of observability

PR #9610, opened 2024-04-25
fix handled dependencies for multidir

jakecoffman:


jakecoffman:
Closing, I think this has been handled by https://github.com/dependabot/dependabot-core/pull/9988 and https://github.com/dependabot/dependabot-core/pull/9938

PR #9611, opened 2024-04-25
#9508: Recreating and Handling  "Name contains illegal characters" in package.json

thavaahariharangit:
Our sentry has raised a new error, 
"Name contains illegal characters" error was not handled.

This PR
1. Recreates the error in both helper (js level) and lib (ruby level)
2. Exception handled in Ruby level
3. Test added in both Ruby and js level.


npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
```diff
@@ -201,6 +201,8 @@ def run_yarn_top_level_updater(top_level_dependency_updates:)
               top_level_dependency_updates
             ]
           )
+        rescue SharedHelpers::HelperSubprocessFailed => e
```
abdulapopoola, line 204: is this the only failure when we run `yarn update`; I assume there are other possible failures too and we should only catch the fileNotParseable specifically.

thavaahariharangit: Yes ensured, that we are catching only file not parseable error.

`[dependabot-core-dev] ~ $ bin/dry-run.rb npm_and_yarn Sotatek-VanPham/colend-utilities`
Original repository will generate helpful error saying file contains illegal charcter.

I have created new repo with same code base by removing the @ character from the name
`[dependabot-core-dev] ~ $ bin/dry-run.rb npm_and_yarn thavaahariharangit/colend-utilities`
This works with out an erorr generated new updated package.json with new dependency suggestions.

npm_and_yarn/helpers/test/yarn/fixtures/updater/illegal_character/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "@colend-contract-helpers",
+  "dependencies": {
+    "@commitlint/cli": "^15.0.0",
```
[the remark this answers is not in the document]
thavaahariharangit: Added comments in both Ruby spec and Js test.

abdulapopoola (approved):

Issue #9442, opened 2024-04-05
NuGet Package With wildcard version throws error


martin-shields-sage:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

nuget

### Package manager version

latest

### Language version

.NET 6

### Manifest location and content before the Dependabot update

https://github.com/martin-shields-sage/TestProject/blob/master/TestProject.csproj

### dependabot.yml content

```
# To get started with Dependabot version updates, you'll need to specify which
# package ecosystems to update and where the package manifests are located.
# Please see the documentation for all configuration options:
# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

version: 2
updates:
  - package-ecosystem: "nuget" # See documentation for possible values
    directory: "/" # Location of package manifests
    schedule:
      interval: "daily"
```

### Updated dependency

https://www.nuget.org/packages/Newtonsoft.Json

12. -> 13.0.3

### What you expected to see, versus what you actually saw

I expect to get a PR for it to update newtonsoft.json to version 13...
(If i update my .csproj with `12.0.3` for example it will run successfully and provide the PR)
what I got was an error in dependabot and no PR made.

### Native package manager behavior

NuGet resolves 12.* to the latest version which at the time of writing is `12.0.3`

### Images of the diff or a link to the PR, issue, or logs

```
 proxy | 2024/04/05 13:15:18 proxy starting, commit: cf8623577dad71c128f219df2b27df6de35b909d
  proxy | 2024/04/05 13:15:18 Listening (:1080)
updater | 2024-04-05T13:15:19.275575323 [810630670:main:WARN:src/devices/src/legacy/serial.rs:222] Detached the serial input due to peer close/error.
updater | time="2024-04-05T13:15:22Z" level=info msg="guest starting" commit=6d38d40701a6d7284d26225f78dd2c6b85c16229
updater | time="2024-04-05T13:15:22Z" level=info msg="starting job..." fetcher_timeout=10m0s job_id=810630670 updater_timeout=45m0s updater_version=5fcf148988e84e7e03871d44106032320cf7c61d-nuget
updater | 2024/04/05 13:15:27 INFO <job_810630670> Starting job processing
[...]

PR #9613, opened 2024-04-25
Nuget - Update version parser to include wildcard version

martin-shields-sage:


JamieMagee:
This only includes test changes. Are wildcard versions already supported?

martin-shields-sage:
> This only includes test changes. Are wildcard versions already supported?

I just pushed a change to do the actual fix, wildcard versions arnt supported as of yet as highlighted here: https://github.com/dependabot/dependabot-core/issues/9442

I have not done Ruby before but was keen to give this a try anyways, if you do have any advice would be appreciated thank you.

Issue #9261, opened 2024-03-11
Nuget update timeout caused by too many requests to https://dc.services.visualstudio.com:443/v2/track


TomW-Skyline:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

Nuget

### What you expected to see, versus what you actually saw

Dependabot updates fail on our solution with "Dependabot timed out during its update". In the logging I see thousands of POST requests to "https://dc.services.visualstudio.com:443/v2/track". These request take a lot of time. What are they used for, and how can this be solved?

 proxy | 2024/03/11 02:01:37 [623] POST https://dc.services.visualstudio.com:443/v2/track
  proxy | 2024/03/11 02:01:37 [623] 200 https://dc.services.visualstudio.com:443/v2/track
  proxy | 2024/03/11 02:01:37 [625] POST https://dc.services.visualstudio.com:443/v2/track
  proxy | 2024/03/11 02:01:37 [625] 200 https://dc.services.visualstudio.com:443/v2/track
  proxy | 2024/03/11 02:01:37 [627] POST https://dc.services.visualstudio.com:443/v2/track
  proxy | 2024/03/11 02:01:37 [627] 200 https://dc.services.visualstudio.com:443/v2/track
  proxy | 2024/03/11 02:01:37 [629] POST https://dc.services.visualstudio.com:443/v2/track
  proxy | 2024/03/11 02:01:37 [629] 200 https://dc.services.visualstudio.com:443/v2/track
  proxy | 2024/03/11 02:01:39 Posting metrics to remote API endpoint
  proxy | 2024/03/11 02:01:39 Successfully posted metrics data via api client
  proxy | 2024/03/11 02:01:41 [632] POST https://dc.services.visualstudio.com:443/v2/track
  proxy | 2024/03/11 02:01:41 [633] POST https://dc.services.visualstudio.com:443/v2/track
  proxy | 2024/03/11 02:01:41 [633] 200 https://dc.services.visualstudio.com:443/v2/track
  proxy | 2024/03/11 02:01:41 [632] 200 https://dc.services.visualstudio.com:443/v2/track
  proxy | 2024/03/11 02:01:43 [635] POST https://dc.services.visualstudio.com:443/v2/track
  proxy | 2024/03/11 02:01:43 [635] 200 https://dc.services.visualstudio.com:443/v2/track
  proxy | 2024/03/11 02:01:43 [637] POST https://dc.services.visualstudio.com:443/v2/track
[...]

Issue #9375, opened 2024-03-27
Timeout running job when updating NuGet packages after updater logic was re-written in C#


batkaevruslan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

NuGet

### What you expected to see, versus what you actually saw

Hi! Since the release of https://github.blog/changelog/2023-11-28-improvements-to-nuget-support-for-dependabot/
Dependabot stopped working in our non-tiny mono repository. Before that, it had been working for more than half a year.

```
proxy | 2024/03/26 05:40:57 proxy starting, commit: cf8623577dad71c128f219df2b27df6de35b909d
...
updater | 2024/03/26 05:41:06 INFO <job_805408216> Starting job processing
...
updater |     file: Directory.Packages.props, metadata: 
updater |   name: Microsoft.Build.Traversal, version: 4.1.0
updater |     file: Build.IntegrationTests.proj, metadata: 
updater | 2024/03/26 05:48:58 INFO <job_805408216> Checking if prometheus-net 8.2.1 needs updating
updater | 2024/03/26 05:48:59 INFO <job_805408216> Filtered out 128 pre-release versions
  proxy | 2024/03/26 05:48:59 [227] GET https://nuget.***.dev:443/nuget/Packages(Id='prometheus-net',Version='8.2.1')
  proxy | 2024/03/26 05:48:59 [227] * authenticating nuget feed request (host: nuget.***.dev, basic auth)
  proxy | 2024/03/26 05:48:59 [227] 404 https://nuget.***.dev:443/nuget/Packages(Id='prometheus-net',Version='8.2.1')
  proxy | 2024/03/26 05:48:59 [229] GET https://api.nuget.org:443/v3-flatcontainer/prometheus-net/8.2.1/prometheus-net.nuspec
  proxy | 2024/03/26 05:48:59 [229] 200 https://api.nuget.org:443/v3-flatcontainer/prometheus-net/8.2.1/prometheus-net.nuspec
  proxy | 2024/03/26 05:49:57 Posting metrics to remote API endpoint
  proxy | 2024/03/26 05:49:57 Successfully posted metrics data via api client
updater | time="2024-03-26T06:28:30Z" level=info msg="task complete" container_id=job-805408216-updater exit_code=137 job_id=805408216 step=updater
updater | time="2024-03-26T06:28:32Z" level=warning msg="timeout running job" error="waiting for updater: waiting for container: context deadline exceeded" job_id=805408216
```
[...]

PR #9614, opened 2024-04-25
don't sent dotnet telemetry

brettfo:
The `dotnet` CLI by default sends a bunch of telemetry that's not necessary (and isn't really what they're looking for anyway.)  It shows up in the log as a bunch of `POST` requests to `https://dc.services.visualstudio.com/v2/track`.  Setting an environment variable turns this off.  I've confirmed this change locally.


JamieMagee:
Closes #9261
Also potentially #9375

JamieMagee (approved):
I went down a rabbit hole with NuGet on this issue. Can't believe it was the .NET CLI the whole time!

brettfo:
@JamieMagee that second issue won't be closed by this, it's been running on my machine for 4 hours now, but the first one might.

PR #9571, opened 2024-04-23
Sign branch images with cosign

Nishnha:
Local branch of https://github.com/dependabot/dependabot-core/pull/9547

An initial implementation of https://github.com/dependabot/dependabot-core/issues/9546 for branch images

cc/ @JamieMagee

PR #9615, opened 2024-04-26
Sign remaining images with cosign

JamieMagee:
This is a follow up to #9571, which verified the signing of branch images. I'm now extending this to the production images that we publish from this repository.

### DIFF
--- a/.github/workflows/images-updater-core.yml
+++ b/.github/workflows/images-updater-core.yml
@@ -1,7 +1,7 @@
 name: Updater-Core image
 env:
   UPDATER_CORE_IMAGE: "ghcr.io/dependabot/dependabot-updater-core"
-on:  # yamllint disable-line rule:truthy
+on: # yamllint disable-line rule:truthy
   push:
     branches:
       - main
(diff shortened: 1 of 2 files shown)


.github/workflows/images-updater-core.yml
```diff
@@ -28,6 +32,8 @@ jobs:
       - name: Push latest image
         run: |
           docker push "$UPDATER_CORE_IMAGE:latest"
+          cosign sign --yes $(cosign triangulate --type=digest "$UPDATER_CORE_IMAGE:latest")
```
JamieMagee, line 35: The tag below should also resolve to the same digest, so we only need to sign once.

yeikel: Using `latest` is a bit dangerous. I'd prefer if we could use a hash

JamieMagee: Don't worry, `cosign triangulate` prints the hash behind a tag.

```
$ cosign triangulate --help
Outputs the located cosign image reference. This is the location where cosign stores the specified artifact type.

$ cosign triangulate --type=digest ghcr.io/dependabot/dependabot-updater-core:latest
ghcr.io/dependabot/dependabot-updater-core@sha256:5401993f730399e809a7ae35156674acb39a7aece707fddb0b0427c938c381d6
```

bdragon (approved):

Issue #9546, opened 2024-04-20
Sign Dependabot container images


JamieMagee:
### Code improvement description

Signing generated containers allows us (and any external users) to verify that a container image actually came from us.  With GitHub Actions support for OIDC [^1] it's possible to use cosign [^2] to sign containers in GitHub Actions without any human interaction or storing long-lived keys as secrets.

Here's a GitHub blog post on the same topic: https://github.blog/2021-12-06-safeguard-container-signing-capability-actions/

[^1]: https://github.blog/changelog/2021-10-27-github-actions-secure-cloud-deployments-with-openid-connect/
[^2]: https://github.com/sigstore/cosign

JamieMagee:
It worked for the branch images. Here's the logs for the Bundler image from #9571

```
Generating ephemeral keys...
Retrieving signed certificate...
Successfully verified SCT...

	The sigstore service, hosted by sigstore a Series of LF Projects, LLC, is provided pursuant to the Hosted Project Tools Terms of Use, available at [https://lfprojects.org/policies/hosted-project-tools-terms-of-use/.](https://lfprojects.org/policies/hosted-project-tools-terms-of-use/)
	Note that if your submission includes personal data associated with this signed artifact, it will be part of an immutable record.
	This may include the email address associated with the account with which you authenticate your contractual Agreement.
	This information will be used for signing this artifact and will be stored in public transparency logs and cannot be removed later, and is subject to the Immutable Record notice at [https://lfprojects.org/policies/hosted-project-tools-immutable-records/.](https://lfprojects.org/policies/hosted-project-tools-immutable-records/)

By typing 'y', you attest that (1) you are not submitting the personal data of any other person; and (2) you understand and agree to the statement and the Agreement terms at the URLs listed above.
tlog entry created with index: 88201941
Pushing signature to: ghcr.io/dependabot/dependabot-updater-bundler
```

I can view that record on Rekor: https://search.sigstore.dev/?logIndex=88201941

And I can also verify the claims and certificates using cosign

```
$ cosign verify \
  ghcr.io/dependabot/dependabot-updater-bundler@sha256:32137fcccb1700db2225587464ad9ece4c3e0dc0cb0f50540df830eb49956513 \
  --certificate-oidc-issuer https://token.actions.githubusercontent.com \
  --certificate-identity https://github.com/dependabot/dependabot-core/.github/workflows/images-branch.yml@refs/pull/9571/merge

[...]

PR #9616, opened 2024-04-26
Enable Docker content trust

JamieMagee:
This change enabled [Docker content trust][1], which verifies the signatures of container images used by docker.

Why are we using Docker content trust to verify signatures, when I am talking about signing container images with cosign in #9546 I hear you ask? I hope this explains things: https://xkcd.com/927/

[1]: https://docs.docker.com/engine/security/trust/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,9 @@ concurrency:
 jobs:
   ci:
     runs-on: ubuntu-latest
+    env:
+      DOCKER_BUILDKIT: 1
+      DOCKER_CONTENT_TRUST: 1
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -14,6 +14,9 @@ concurrency:
 jobs:
   ci:
     runs-on: ubuntu-latest
+    env:
+      DOCKER_BUILDKIT: 1
```
JamieMagee, line 18: I also enabled Docker buildkit as the default builder here. It was enabled everywhere else, so I guessed that it missing here was an oversight.

PR #9634, opened 2024-04-29
Dependabot core issue npe composer fileparser

GarryHurleyJr:


composer/lib/dependabot/composer/file_parser.rb
```diff
@@ -165,9 +165,18 @@ def git_dependency_details(package_details, requirement)
       end
 
       def lockfile_details(name:, type:)
-        key = lockfile_key(type)
-        parsed_lockfile.fetch(key, []).find { |d| d["name"] == name }
-      end
+		begin
```
abdulapopoola, line 168: seems the whitespace formatting can be aligned?

composer/lib/dependabot/composer/file_parser.rb
```diff
@@ -165,9 +165,18 @@ def git_dependency_details(package_details, requirement)
...
+           key = lockfile_key(type)
```
abdulapopoola, line 169: do we know why this can throw?

composer/lib/dependabot/composer/file_parser.rb
GarryHurleyJr: yeah, I did an edit in a code editor that has its indent settings configured wrong.  I will fix the cosmetics.

composer/lib/dependabot/composer/file_parser.rb
GarryHurleyJr: well, what I found is that the file_parser is actually looking at all files in a folder.  It ignores the name of the file. when looking for the key fields.  My thought is that the reason we are seeing the issue is that we are not filtering the filees we parse based on name or extension.  If we did that, it would limit us.  so, type could be nil, so could key, parsed_lockfile, parsed_lockfile.fetch(key,[])  or d.  The problem is that we are not getting the nil in lockfile because any file can be parsed - just not properly.

composer/lib/dependabot/composer/file_parser.rb
```diff
@@ -165,10 +165,19 @@ def git_dependency_details(package_details, requirement)
...
-        key = lockfile_key(type)
-        parsed_lockfile.fetch(key, []).find { |d| d["name"] == name }
-      end
-
+        begin
+          key = lockfile_key(type)
+        rescue
+          key = nil
+        end
+
```
abdulapopoola, line 173: can we rescue everything in here instead of doing all the checks? Or maybe move up to the callers of `lockfile_details` and throw the `DependencyFileNotParseable` there?

GarryHurleyJr:
Overthought the issue.  reworked code

Issue #9612, opened 2024-04-25
Closing .NET PRs as "no longer needed" when dependency is still present and out-of-date


DavidBoike:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

.NET

### Manifest location and content before the Dependabot update

https://github.com/Particular/ServiceControl/blob/master/src/Directory.Packages.props#L17

### dependabot.yml content

https://github.com/Particular/ServiceControl/blob/master/.github/dependabot.yml

### Updated dependency

Microsoft.AspNetCore.Mvc.Testing from 8.0.3 to 8.0.4

### What you expected to see, versus what you actually saw

Dependabot PR was approved, and auto-merge was enabled. Instead of merging, Dependabot [left the comment](https://github.com/Particular/ServiceControl/pull/4066#issuecomment-2056729474):

> Looks like Microsoft.AspNetCore.Mvc.Testing is no longer a dependency, so this is no longer needed.

Except it's definitely still used. A [code search of the repo](https://github.com/search?q=repo%3AParticular%2FServiceControl%20Microsoft.AspNetCore.Mvc.Testing&type=code) shows the version defined in https://github.com/Particular/ServiceControl/blob/da452ea82a2cf6aca420d73029c7861cc009c764/src/Directory.Packages.props#L17 as 8.0.3 and the package being referenced in several projects in the solution, while the actual current version of the package on NuGet is [8.0.4](https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc.Testing/8.0.4).

### Native package manager behavior

N/A

### Images of the diff or a link to the PR, issue, or logs

https://github.com/Particular/ServiceControl/pull/4066

### Smallest manifest that reproduces the issue

Not sure, but we have only seen this in:

1. Repos with NuGet central package management via a Directory.Packages.props file.
    * The content above is one example
    * Another is [Particular/Particular.EndpointThroughputCounter: Bump Azure.Messaging.ServiceBus from 7.17.2 to 7.17.5](https://github.com/Particular/Particular.EndpointThroughputCounter/pull/747) where the [package version in the repo is still ](https://github.com/Particular/Particular.EndpointThroughputCounter/blob/main/src/Directory.Packages.props#L13) and [current is 7.17.5](https://www.nuget.org/packages/Azure.Messaging.ServiceBus/7.17.5)
2. Private repos. I assume this is less helpful, but I can find examples of this too if necessary.
[...]

Issue #9495, opened 2024-04-13
Unhandled exception: System.ArgumentException: An item with the same key has already been added.


martincostello:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package ecosystem

NuGet

### Manifest location and content before the Dependabot update

https://github.com/martincostello/alexa-london-travel/commit/1b47c2e5b810a23206718c3904c5c257974ee4c8

### dependabot.yml content

[dependabot.yml](https://github.com/martincostello/alexa-london-travel/blob/1b47c2e5b810a23206718c3904c5c257974ee4c8/.github/dependabot.yml)

### Updated dependency

None

### What you expected to see, versus what you actually saw

Dependencies should have been updated, but instead nothing is updated an an error is logged:

```console
updater | 2024/04/13 09:39:39 ERROR <job_814225123> Block argument of NuGetConfigCredentialHelpers::patch_nuget_config_for_action causes an exception Discovering build files in workspace [/home/dependabot/dependabot-updater/repo].
updater |   No dotnet-tools.json file found.
updater |   Discovered [global.json] file.
updater |   Discovering projects beneath [.].
updater |   No packages.config file found.
updater | Unhandled exception: System.ArgumentException: An item with the same key has already been added. Key: coverlet.msbuild
updater |    at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
updater |    at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
updater |    at System.Linq.Enumerable.ToDictionary[TSource,TKey](IEnumerable`1 source, Func`2 keySelector, IEqualityComparer`1 comparer)
updater |    at NuGetUpdater.Core.Discover.SdkProjectDiscovery.DiscoverAsync(String repoRootPath, String workspacePath, String projectPath, Logger logger) in /opt/nuget/lib/NuGetUpdater/NuGetUpdater.Core/Discover/SdkProjectDiscovery.cs:line 32
updater |    at NuGetUpdater.Core.Discover.DiscoveryWorker.RunForProjectPathsAsync(String repoRootPath, String workspacePath, IEnumerable`1 projectPaths) in /opt/nuget/lib/NuGetUpdater/NuGetUpdater.Core/Discover/DiscoveryWorker.cs:line 160
[...]

PR #9642, opened 2024-04-30
Allow NuGet dependencies to exist with duplicate names but different dependency types

DavidBoike:
* Fixes https://github.com/dependabot/dependabot-core/issues/9612 and https://github.com/dependabot/dependabot-core/issues/9495

`NuGetUpdater.Cli.exe discover` was throwing "An item with teh same key has already been added" in the situation where a centrally-managed Directory.Build.props file would have 2 references for the same package based on a condition such as the repository using central package management.

This is an example that would make the tool throw from my company's centrally-managed [Directory.Build.props file](https://github.com/Particular/RepoStandards/blob/master/src/Directory.Build.props):

```xml
<ItemGroup Condition="'$(ManagePackageVersionsCentrally)' == 'true'">
  <GlobalPackageReference Include="Particular.Analyzers" Version="$(ParticularAnalyzersVersion)" />
</ItemGroup>

<ItemGroup Condition="'$(ManagePackageVersionsCentrally)' != 'true'">
  <PackageReference Include="Particular.Analyzers" Version="$(ParticularAnalyzersVersion)" PrivateAssets="All" />
</ItemGroup>
```

If central package management is enabled, then our company-managed Roslyn analyzers package is included as a `GlobalPackageReference`. If not, it's added to the project as a regular `PackageReference`.

But the duplicate name caused a problem with the discover code tried to do `.ToDictionary(…)` on the name alone, so this change uses a lookup to allow multiple values.

A test has been added that demonstrates the desired result.

### DIFF
--- a/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Cli.Test/EntryPointTests.Discover.cs
+++ b/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Cli.Test/EntryPointTests.Discover.cs
@@ -214,6 +214,77 @@ public partial class EntryPointTests
             });
         }
 
+        [Fact]
+        public async Task WithDuplicateDependenciesOfDifferentTypes()
+        {
+            var projectPath = "path/to/my.csproj";
+            var directoryBuildPropsPath = "path/Directory.Build.props";
+            await RunAsync(path =>
+            [
+                "discover",
+                "--repo-root",
+                path,
+                "--workspace",
+                path,
+            ],
+            new[]
+            {
+                (projectPath, """
+                    <Project ToolsVersion="15.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+                      <PropertyGroup>
+                        <TargetFramework>net8.0</TargetFramework>
+                      </PropertyGroup>
+                      <ItemGroup>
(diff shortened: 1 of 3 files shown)


DavidBoike:
I see that one of the tests in `Specs / ci (nuget, nuget, nuget)` is failing. I don't grok Ruby or RSpec very well, unfortunately, but I do see that the most recent run on `main` [also had a failure](https://github.com/dependabot/dependabot-core/actions/runs/8887088104) in that test suite, although it's a different test. I don't know if the RSpec tests even run NuGetUpdater.Cli (I can't find any evidence that it does?) but I did take the failure I saw (getting 5 dependencies out of a 9-dependency packages.config file) and set up a different test in the .NET project tests, and the updater was still coming up with 9 dependencies there, so I don't know what's up with that.

DavidBoike:
Must have been a heisenbug, there was a new commit to main, and after I rebased all the tests came up green.

brettfo (approved):

brettfo:
@DavidBoike If you curious for the future, this repo only invokes the Ruby unit tests so to include the C# tests we got sneaky and run `dotnet test ...` [here](https://github.com/dependabot/dependabot-core/blob/41313ee8b79af48f8368a96123179a4a4959b843/nuget/spec/dependabot/nuget/native_helpers_spec.rb#L59-L95).

PR #9646, opened 2024-04-30
experimental glob support

jakecoffman:
This adds support for globs in the directories field. 

## Approach

The FileFetcher for each ecosystem knows what files are required in the directory to perform an update.

So we use Ruby's built-in support for globs `Dir.glob` and then iterate over all of the directories with a new FileFetcher. When the FileFetcher doesn't find manifests it throws DependencyFileNotFound, which we catch and eat. 

Now, the work to find the directories that have manifests is done, and the `directory` field of each DependencyFile is set to the `job.source.directory` that it was found in. That means in the DependencySnapshot we can gather up all those `dependency` fields, and replace the `job.source.directories` with that actual list of directories that need updates.


updater/lib/dependabot/file_fetcher_command.rb
```diff
@@ -250,5 +282,10 @@ def github_connectivity_client(job)
         }
       })
     end
+
+    def glob?(directory)
+      # We could tighten this up, but it's probably close enough.
+      directory.include?("*") || directory.include?("?") || (directory.include?("[") && directory.include?("]"))
```
Nishnha, line 288: Can you explain how ? [ and ] correlate to a glob directory?

jakecoffman: It's a common supported syntax: https://en.wikipedia.org/wiki/Glob_(programming)#Syntax

```
irb(main):001> File.fnmatch?("[ab]c", "c", File::FNM_PATHNAME)
=> false
irb(main):002> File.fnmatch?("[ab]c", "ac", File::FNM_PATHNAME)
=> true
irb(main):003> File.fnmatch?("[ab]c", "bc", File::FNM_PATHNAME)
=> true
irb(main):004> File.fnmatch?("[ab]c", "cc", File::FNM_PATHNAME)
=> false
```

silent/tests/testdata/vu-glob.txt
```diff
@@ -0,0 +1,78 @@
...
+  ]
+}
+
+-- input.yml --
+job:
+  package-manager: "silent"
+  source:
+    directories:
+      - "**/*"
+      - "/"
```
Nishnha, line 72: Should this be only the globstar entry?

jakecoffman: Normally`**/*` won't match dotfiles or the current working directory `.` which is why I added it here.

We could set `File::FNM_DOTMATCH` which would match `.` but also every directory that starts with `.` will be matched as well. 

I can't think of a reason that would be a bad idea other than it's going to search more directories. It's probably better we set it since it's easy to have glob not match files and directories starting with `.`: `**/[!.]*`

So I'll make that change.

Nishnha (approved):
This looks good to me

Issue #4202, opened 2021-09-06
Missing  ext-pcov 


oleg-andreyev:
**Package ecosystem**
composer
**Package manager version**
composer 2.x
**Language version**
PHP 7.3
**Manifest location and content prior to update**
```json
{"require": {
    "php": "^7.3",
    "ext-pdo": "*",
    "ext-curl": "*",
    "ext-iconv": "*",
    "ext-json": "*",
    "ext-zlib": "*",
    "doctrine/common": "~2.7",
    "google/apiclient": "^2.0",
    "jms/serializer-bundle": "~0.13|~1.1|^2.0|^3.7",
    "symfony/console": "^2.8|^3.4|^4.4|^5.1",
    "symfony/dependency-injection": "^2.7|^3.4|^4.0|^5.1",
    "symfony/monolog-bundle": "^2.11|^3.1",
    "symfony/stopwatch": "^2.6|^3.4|^4.4|^5.1",
    "twig/twig": "1.42.3|^2.4",
    "webmozart/assert": "~1.8.0"
  },
  "require-dev": {
    "ext-pcov": "*",
    "infection/infection": "^0.18.0",
    "doctrine/orm": "^2.7",
    "phpmd/phpmd": "^2.10",
    "phpmetrics/phpmetrics": "^2.7",
    "phpunit/phpunit": "^9.2",
    "psalm/phar": "~4.1.0",
    "qossmic/deptrac-shim": "^0.11.1",
    "symfony/framework-bundle": "^5.1",
    "symfony/dotenv": "^5.1",
    "symfony/event-dispatcher": "^5.1",
    "symfony/yaml": "^5.1",
    "symfony/proxy-manager-bridge": "^5.1",
    "sensiolabs-de/deptrac-shim": "^0.10"
  },}
```

**Updated dependency**

```
Your requirements could not be resolved to an installable set of packages.
Problem 1
    - Root composer.json requires PHP extension ext-pcov * but it is missing from your system. Install or enable PHP's pcov extension.

To enable extensions, verify that they are enabled in your .ini files}
```

**Native package manager behavior**
-

**Images of the diff or a link to the PR, issue or logs**
-

jeffwidman:
👋 Hey there! Sorry for the glacially slow response... we've been a bit buried the past few years, and only just now starting to dig through some of our backlog.

This issue is pretty old, and we've pushed a lot of bugfixes since then... is this still a valid issue?

oleg-andreyev:
[...]

PR #9657, opened 2024-05-02
Stop hiding exception raised by composer

robaiken:
Previously, errors from Composer were suppressed, which made it difficult for users to understand what caused the issue. With this change, we will now display all the errors without hiding them. This allows users to identify and investigate potential problems more easily.


jeffwidman:
I was surprised to see you went after this by raising rather than logging the error... 

Did you see my comment in https://github.com/dependabot/dependabot-core/pull/6290#issuecomment-1535218293 ?

composer/lib/dependabot/composer/update_checker/version_resolver.rb
```diff
@@ -308,6 +308,7 @@ def handle_composer_errors(error)
             # If there *is* a lockfile we can't confidently distinguish between
             # cases where we can't install and cases where we can't update. For
             # now, we therefore just ignore the dependency.
+            Dependabot.logger.error(error.message)
```
jeffwidman, line 311: I noticed that elsehwere when logging `error.message` we generally also log the backtrace: https://github.com/search?q=repo%3Adependabot%2Fdependabot-core%20Dependabot.logger.error&type=code

I'm not sure what is in `error.message`... does the user have enough information in this situation to skip the backtrace?

Just want to be sure that the user has enough information to understand what triggered the error since it can be due to two very different reasons:
1. lockfile can't be checked because install fails due to a missing native extension, ie "can't install"
2. lockfile has conflicting version requirements, ie "can't update"

composer/spec/dependabot/composer/update_checker_spec.rb
```diff
@@ -528,6 +528,12 @@
 
       it { is_expected.to be_nil }
 
+      it "logs an error" do
+        allow(Dependabot.logger).to receive(:error)
+        handle_composer_errors
+        expect(Dependabot.logger).to have_received(:error).once
+      end
```
jeffwidman, line 535: If you look at the `context` of this test, it's for an update that can't resolve due to `version_conflict_on_update`... it's great that we're logging the error in that case so that the user can see the failure is a malformed lockfile. This is the "can't update" scenario.

Another scenario that can trigger this error is `missing_native_extension`... that's the "can't install" scenario.

I don't know if that scenario already exists in the test suite, but if so I'd copy this test to that as well so that we are testing both scenarios that we expect to throw errors.

jeffwidman (commented):
The approach makes sense to me, just two questions/suggestions about the implementation.

composer/spec/dependabot/composer/update_checker_spec.rb
```diff
@@ -528,6 +528,12 @@
 
       it { is_expected.to be_nil }
 
+      it "version conflict logs an error" do
```
jeffwidman, line 531: The `version conflict` part should be part of the context, not added here... this `it` statement describes the result

ie, modify the `context "with an update that can't resolve" do` to say something like
`context "with an update that can't resolve due to a version conflict" do`

composer/spec/dependabot/composer/update_checker_spec.rb
```diff
@@ -541,6 +547,27 @@
       end
     end
 
+    context "missing native extension logs an error" do
+      let(:dependency_name) { "robaiken/missing-ext" }
```
jeffwidman, line 551: Is this a real extension on packagist? Or a completely fake one? 

If real, this shouldn't be under the `robaiken` namespace... we have a dedicated namespace on packagist for test fixtures: https://packagist.org/packages/dependabot/

If this is a fake extension, then composer will throw an error about unable to find the extension on packagist... ie, a 404-related error... Which is a different error than "unable to find the extension due to compilation errors"... 

A more realistic use case would be to use a PHP extension that _is_ present on packagist, but not popular enough for us to pull in: 
* https://github.com/dependabot/dependabot-core/issues/4202
* https://packagist.org/packages/phpunit/php-code-coverage

jeffwidman:
The skeleton of where you're going on this looks great! Just need a few details tweaked and then this should be good to go...

composer/spec/dependabot/composer/update_checker_spec.rb
robaiken: I have removed the old fixture file and replaced it witht he suggested one

jeffwidman (approved):
👍

composer/spec/dependabot/composer/update_checker_spec.rb
```diff
@@ -528,6 +528,12 @@
 
       it { is_expected.to be_nil }
 
+      it "logs an error" do
+        allow(Dependabot.logger).to receive(:error)
+        Dependabot.logger.error
```
jurre, line 533: Isn't _this_method call fulfilling the expectation set on the line above? We should instead run the actual code under test here I think @robaiken

jeffwidman: Oof, good catch Jurre, I totally missed that.

robaiken: nice, thanks for spotting this

composer/spec/dependabot/composer/update_checker_spec.rb
```diff
@@ -528,10 +528,13 @@
...
-        allow(Dependabot.logger).to receive(:error)
-        Dependabot.logger.error
-        expect(Dependabot.logger).to have_received(:error).once
+      context "logs an error" do
+        before { allow(Dependabot.logger).to receive(:error) }
+
+        it do
+          is_expected.to be_nil
+          expect(Dependabot.logger).to have_received(:error).at_least(:twice)
+        end
```
jeffwidman: I'm curious why you broke this out as a separate section? Why not blend this with the existing test in line 529? 

The spacing of this suggestion is off, so please don't take it verbatim, but this illustrates the general concept... 

```suggestion
      before { allow(Dependabot.logger).to receive(:error) }
      it { is_expected.to be_nil }
      expect(Dependabot.logger).to have_received(:error).once
```

composer/spec/dependabot/composer/update_checker_spec.rb
```diff
@@ -547,27 +550,6 @@
...
-
-      it { is_expected.to be_nil }
-
-      it "logs an error" do
-        allow(Dependabot.logger).to receive(:error)
-        Dependabot.logger.error
-        expect(Dependabot.logger).to have_received(:error).once
-      end
-    end
-
```
jeffwidman: Why delete this test? It was originally added to specifically test a _different_ scenario than the previous test because the `let` part of the setup is different... this is necessary to test the "missing native extension" scenario which is different than the "conflicting version requirements" scenario...

composer/spec/fixtures/packagist_responses/phpunit--php-code-coverage.json
jeffwidman, line 1: As it stand now, this text fixture is unused because the test that you added that used it was deleted... you'll want to either re-add that test or delete this fixture... my vote is for re-adding that test as I think it's important to communicate to the develeoper/test the scenario where Dependabot can't do anything because the native compiled extension is missing.

robaiken: The `handle_composer_errors` has a condition that checks for missing Native Extensions in the composer error message. If there is a missing Native extension, then it will be caught by this check:

https://github.com/dependabot/dependabot-core/blob/78cf509e9df3455af9acb6d7b6edc18fc641425a/composer/lib/dependabot/composer/update_checker/version_resolver.rb#L266-L275

So it shouldn't be possible for this to happen

composer/spec/dependabot/composer/update_checker_spec.rb
robaiken: https://github.com/dependabot/dependabot-core/pull/9657#discussion_r1600238557

jurre (approved):

Issue #9633, opened 2024-04-29
Enable YJIT


JamieMagee:
### Code improvement description

Once #9597 is complete, we should enable YJIT and monitor the performance.

References:
- https://github.com/ruby/ruby/blob/master/doc/yjit/yjit.md#command-line-options

JamieMagee:
@landongrindheim is there enough data to decide whether or not to enable YJIT globally?

landongrindheim:
@JamieMagee Thanks for checking in on this! Here's a graph tracking job durations (the thing we were specifically hoping to impact) 👇 I don't think there's an appreciable difference, though the max values are 🤏 improved. I suspect YJIT would make a bigger difference for longer-running processes.

![Image](https://github.com/user-attachments/assets/5b253c6a-44de-431b-a69b-99ba567c8aa3)

---

Here's a split view of the same data, in case the overlay is hard to read.


![Image](https://github.com/user-attachments/assets/7a2d0e11-dd4e-4abc-a58a-f81f7f827e27)

JamieMagee:
@landongrindheim thanks for the update. If that's kusto, could you regenerate the graphs with [`percentiles`][1] instead of `avg`. That should better show if the max values are improving.

I'd recommend using the [68-95-99.7 rule][2] to show 1, 2, and 3 standard deviations from the mean i.e. `summarize percentiles(..., 68, 95, 99.7) by yjit_enabled`

[1]: https://learn.microsoft.com/en-us/azure/data-explorer/kusto/query/percentiles-aggregation-function
[2]: https://en.wikipedia.org/wiki/68%E2%80%9395%E2%80%9399.7_rule

landongrindheim:
It's Datadog. And I wrote that before thinking about what it was actually reporting 🤦 The max would be the timing of an individual job, which isn't especially meaningful in this context.

landongrindheim:
YJIT was enabled. I think @JamieMagee's suggestion for monitoring via percentiles is the right call. However, I no longer have access to those metrics.

PR #9663, opened 2024-05-03
Enable YJIT for even-numbered update jobs

landongrindheim:
We primarily upgraded to Ruby 3.3 so that we could enable YJIT. We did this in order to hopefully mitigate some of the performance issues some of our larger customers have faced.

This change enables YJIT for _some_ update jobs. The reason for the incremental enablement is that we would like to measure the impact that YJIT has before adopting it fully (which we intend to do).

### DIFF
--- a/updater/bin/fetch_files.rb
+++ b/updater/bin/fetch_files.rb
@@ -32,9 +32,7 @@ trap("TERM") do
 end
 
 begin
-  if Dependabot::Environment.job_id.to_i.even?
-    RubyVM::YJIT.enable
-  end
+  RubyVM::YJIT.enable if Dependabot::Environment.job_id.to_i.even?
 
   Dependabot::FileFetcherCommand.new.run
 rescue Dependabot::RunFailure
--- a/updater/bin/fetch_files.rb
+++ b/updater/bin/fetch_files.rb
@@ -32,6 +32,10 @@ trap("TERM") do
 end
 
 begin
+  if Dependabot::Environment.job_id.to_i.even?
+    RubyVM::YJIT.enable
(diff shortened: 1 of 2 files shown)


jakecoffman (approved):

updater/bin/fetch_files.rb
```diff
@@ -32,6 +32,10 @@ class UpdaterKilledError < StandardError; end
 end
 
 begin
+  if Dependabot::Environment.job_id.to_i.even?
```
abdulapopoola, line 35: would it help to have a feature flag here? That'll make rollback easier (toggling a flag versus a code change)

landongrindheim: We rollback by specifying the sha via chatops, so the rollback ergonomics are very similar 🙂 YJIT is native to Ruby (and a default for Rails 7.2 apps), so I don't anticipate needing to rollback.

abdulapopoola: I assume we don't have any flighting support for core? Not asking you to change anything, mostly trying to build context.

landongrindheim: I don't know what flighting support is. Can you explain?

abdulapopoola: Oops; I meant feature flags (used that interchangeably)

landongrindheim: We do support feature flags, or something very similar. They're called Experiments in this codebase. They're registered/wired up via the service.

bdragon (approved):
Nice idea

JamieMagee:
Related to #9633

PR #9668, opened 2024-05-03
Use pnpm@9 for lockfile >= 9.0

tusbar:
Related: https://github.com/dependabot/dependabot-core/issues/9522

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
@@ -38,7 +38,9 @@ module Dependabot
 
       sig { params(pnpm_lock: DependencyFile).returns(Integer) }
       def self.pnpm_version_numeric(pnpm_lock)
-        if pnpm_lockfile_version(pnpm_lock).to_f >= 6.0
+        if pnpm_lockfile_version(pnpm_lock).to_f >= 9.0
+          9
+        elsif pnpm_lockfile_version(pnpm_lock).to_f >= 6.0
           8
         elsif pnpm_lockfile_version(pnpm_lock).to_f >= 5.4
           7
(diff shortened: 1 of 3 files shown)


jeffwidman (approved):

bdragon (approved):

npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
```diff
@@ -38,7 +38,9 @@ def self.yarn_version_numeric(yarn_lock)
 
       sig { params(pnpm_lock: DependencyFile).returns(Integer) }
       def self.pnpm_version_numeric(pnpm_lock)
-        if pnpm_lockfile_version(pnpm_lock).to_f >= 6.0
+        if pnpm_lockfile_version(pnpm_lock).to_f >= 9.0
+          9
+        elsif pnpm_lockfile_version(pnpm_lock).to_f >= 6.0
           8
         elsif pnpm_lockfile_version(pnpm_lock).to_f >= 5.4
           7
```
pboling, line 46: It might be better to invert the checks:
```
let pnpm_version = pnpm_lockfile_version(pnpm_lock).to_f;

if (typeof(pnpm_version) === 'undefined' || pnpm_version  < 5.4)
  6
elsif pnpm_version < 6
  7
elsif pnpm_version < 9
  8
else
  parseInt(pnpm_version, 10)
end
```

Then you'll likely never have to fix it again. ;). Just be done fixing it.

pboling (commented):
Yes, but perhaps perma-fix it?

fflaten:
Is pnpm 9 supported at github immediately or are we waiting on a release?

robaiken:
@fflaten The update to support pnpm 9 is live, Dependabot is ready to handle your pnpm projects. If you have any other issues, please raise them in the [issues](https://github.com/dependabot/dependabot-core/issues) tab.

tusbar:
@robaiken – although lockfiles v9.0 can be parsed, pnpm 9 won’t be used as long as this is not released.

Lockfiles will get re-generated using v6.0 format.

jeffwidman:
We deploy `main` at GitHub before it goes into an official tagged release. 

Deploys are invisible to the public, but we often follow a deploy-then-merge strategy, and if we don't, we generally merge a couple of low-risk PR's and then do a fast-follow deploy which will pick up the whole 🚂 .

So this is already live on GitHub.com.

pboling:
It still does not work _at all_

- https://github.com/dependabot/dependabot-core/pull/9687#issuecomment-2139150623
- https://github.com/dependabot/dependabot-core/issues/9684#issuecomment-2137899341

PR #9687, opened 2024-05-08
Improve support for PNPM 9

robaiken:


tusbar:
@robaiken it seems that the corepack version is always the version getting used and the `pnpm_version_numeric` method doesn’t matter?

npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
```diff
@@ -38,14 +38,16 @@ def self.yarn_version_numeric(yarn_lock)
...
+
+        if pnpm_version < 5.4
+          6
+        elsif pnpm_version < 6
           7
+        elsif pnpm_version < 9
+          8
         else
-          6
+          pnpm_version.to_i
```
pboling, line 50: Much better! :)

pboling: OK, but that's the current state of things, right?  This is a better version of what is currently already "unsafe and probably wrong", right?

wJoenn: What's unsafe or wrong with this exactly ?
```ruby
if pnpm_lockfile_version(pnpm_lock).to_f >= 9.0
  9
elsif pnpm_lockfile_version(pnpm_lock).to_f >= 6.0
  8
elsif pnpm_lockfile_version(pnpm_lock).to_f >= 5.4
  7
else
  6
end
```

tusbar:
@robaiken hello, can this get more attention please? pnpm support has been broken for almost a month now.

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/pnpm_lockfile_updater.rb
```diff
@@ -66,8 +66,8 @@ def run_pnpm_updater
           end.join(" ")
 
           Helpers.run_pnpm_command(
-            "install #{dependency_updates} --lockfile-only --ignore-workspace-root-check",
-            fingerprint: "install <dependency_updates> --lockfile-only --ignore-workspace-root-check"
+            "install #{dependency_updates} --lockfile-only",
+            fingerprint: "install <dependency_updates> --lockfile-only"
```
jurre, line 70: The failing tests suggest we should add this back

```suggestion
            "install #{dependency_updates} --lockfile-only --ignore-workspace-root-check",
            fingerprint: "install <dependency_updates> --lockfile-only --ignore-workspace-root-check"
```

npm_and_yarn/lib/dependabot/npm_and_yarn/helpers.rb
pboling: > What's unsafe or wrong with this exactly ?

- Given `pnpm_lockfile_version(pnpm_lock).to_f == 10`, then it will use version `9`
- Given `pnpm_lockfile_version(pnpm_lock).to_f == 42`, then it will use version `9`
- Given `pnpm_lockfile_version(pnpm_lock).to_f == 8576309`, then it will use version `9`

Invert the logic, and you can future proof it.  It makes no sense to constrain it the way it is.

robaiken:
this has been replaced by: https://github.com/dependabot/dependabot-core/pull/10073

PR #9694, opened 2024-05-09
restrict C# nuget unit tests to local package feeds

brettfo:
Comments have been added inline with the code where the most attention should be paid.

In short, if a unit test is written as it was before, the behavior will remain the same; network access to `api.nuget.org` will be allowed.  _However_, as soon as the `packages` parameter is specified, a test-only `NuGet.Config` is written to the test directory which disallows **all** network access and only allows the packages created in a special directory to be queried.  Since we don't use any custom NuGet API handling, this is perfectly fine as we're letting the "real" NuGet handle its sources however it sees fit (which in this case is all from a local feed.)

The vast majority of this PR was changing things like `Newtonsoft.Json` to `Some.Package` and can be ignored.  In some tests the specific package dependencies that made the test case interesting have been explicitly laid out which should help with readability quite a bit.

### DIFF
--- a/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core.Test/MockNuGetPackage.cs
+++ b/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core.Test/MockNuGetPackage.cs
@@ -1,0 +1,419 @@
+using System.IO.Compression;
+using System.Security.Cryptography;
+using System.Text;
+using System.Text.Json.Nodes;
+using System.Text.RegularExpressions;
+using System.Xml;
+using System.Xml.Linq;
+using System.Xml.XPath;
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Emit;
+
+namespace NuGetUpdater.Core.Test
+{
+    public record MockNuGetPackage(
+        string Id,
+        string Version,
+        XElement[]? AdditionalMetadata = null,
+        (string? TargetFramework, (string Id, string Version)[] Packages)[]? DependencyGroups = null,
+        (string Path, byte[] Content)[]? Files = null)
+    {
+        private static readonly XNamespace Namespace = "http://schemas.microsoft.com/packaging/2013/05/nuspec.xsd";
+        private static readonly XmlWriterSettings WriterSettings = new()
+        {
+            Encoding = Encoding.UTF8,
+            Indent = true,
+        };
+
+        private XDocument? _nuspec;
+        private Stream? _stream;
+
+        public void WriteToDirectory(string localPackageSourcePath)
+        {
+            string cachePath = Path.Join(localPackageSourcePath, "_nupkg_cache");
+            string nupkgPath = Path.Join(cachePath, $"{Id}.{Version}.nupkg");
+            Directory.CreateDirectory(cachePath);
+            Stream stream = GetZipStream();
+            using (FileStream fileStream = new(nupkgPath, FileMode.Create))
+            {
+                stream.CopyTo(fileStream);
+            }
+
+            // add the package to the local feed; this is equivalent to running
+            //   nuget add <nupkgPath> -source <localPackageSourcePath>
+            // but running that in-process locks the files, so we have to do it manually
+            // the end result is 4 files:
+            //   .nupkg.metadata                // a JSON object with the package's content hash and some other fields
+            //   <id>.<version>.nupkg           // the package itself
+            //   <id>.<version>.nupkg.sha512    // the SHA512 hash of the package
+            //   <id>.nuspec                    // the package's nuspec file
+            string expandedPath = Path.Join(localPackageSourcePath, Id.ToLowerInvariant(), Version);
+            Directory.CreateDirectory(expandedPath);
+            File.Copy(nupkgPath, Path.Join(expandedPath, $"{Id}.{Version}.nupkg".ToLowerInvariant()));
+            using XmlWriter writer = XmlWriter.Create(Path.Join(expandedPath, $"{Id}.nuspec".ToLowerInvariant()), WriterSettings);
+            GetNuspec().WriteTo(writer);
+            using SHA512 sha512 = SHA512.Create();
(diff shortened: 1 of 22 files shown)


nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core.Test/Update/UpdateWorkerTestBase.cs
```diff
@@ -201,6 +223,42 @@ public abstract class UpdateWorkerTestBase : TestBase
         AssertContainsFiles(expectedResult, actualResult);
     }
 
+    public static async Task MockNuGetPackagesInDirectory(MockNuGetPackage[]? packages, string temporaryDirectory)
```
brettfo, line 226: This is where the fake NuGet packages are written to a local feed and an appropriate `NuGet.Config` is created.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core.Test/MockNuGetPackage.cs
```diff
@@ -0,0 +1,419 @@
...
+        private static readonly XmlWriterSettings WriterSettings = new()
+        {
+            Encoding = Encoding.UTF8,
+            Indent = true,
+        };
+
+        private XDocument? _nuspec;
+        private Stream? _stream;
+
+        public void WriteToDirectory(string localPackageSourcePath)
```
brettfo, line 33: This is the interesting part about the fake NuGet packages; a directory is populated with some specific files (comments inline) that allow the directory to be used as a NuGet package source.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core.Test/Update/UpdateWorkerTestBase.cs
brettfo: If no `packages` parameter is specified, no local feed and no `NuGet.Config` file will be created, which will still allow tests to hit the network if they're still required.  As soon as a test sets `packages: [...]` then all network access it cut off.  I did this to allow for the most flexibility, but also as a way to ensure that a test that specifies `packages` won't accidentally pull something from `api.nuget.org`.

brbayes-msft (approved):

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/PathHelper.cs
```diff
@@ -111,4 +111,22 @@ public static void CopyDirectory(string sourceDirectory, string destinationDirec
             CopyDirectory(subDir.FullName, newDestinationDir);
         }
     }
+
+    public static bool IsSubdirectoryOf(string parentDirectory, string childDirectory)
+    {
+        var parentDirInfo = new DirectoryInfo(parentDirectory);
+        var childDirInfo = new DirectoryInfo(childDirectory);
+
+        while (childDirInfo.Parent is not null)
```
JoeRobich, line 120: I like this more thorough implementation, but would `childDirInfo.FullName.StartsWith(parentDirInfo.FullName);` work? This is me assuming DirectoryInfo normalized the paths.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core.Test/Update/UpdateWorkerTestBase.cs
```diff
@@ -201,6 +223,42 @@ public abstract class UpdateWorkerTestBase : TestBase
...
+            MockNuGetPackage[] allPackages = packages.Concat(MockNuGetPackage.CommonPackages).ToArray();
+
+            // write all packages to disk
+            foreach (MockNuGetPackage package in allPackages)
+            {
+                package.WriteToDirectory(localFeedPath);
+            }
+
+            // override various nuget locations
+            foreach (var envName in new[] { "NUGET_PACKAGES", "NUGET_HTTP_CACHE_PATH", "NUGET_SCRATCH", "NUGET_PLUGINS_CACHE_PATH" })
```
JoeRobich, line 241: Do these values need to be reverted or is the expectation that every test will end up mocking their packages?

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core.Test/Update/UpdateWorkerTestBase.cs
JoeRobich: I wonder if this method should be extracted somewhere more central, since it isn't Updater specific.

JoeRobich (approved):
This change makes me so happy!

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Utilities/PathHelper.cs
brettfo: `.StartsWith()` doesn't work for things like `C:\temp` and `C:\temporary-documents\file.txt`.  In this case `.StartsWith` will return `true`, but it's not under the same path.  I considered ensuring a trailing directory separator, but that ends up with a bunch of edge cases that I didn't like.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core.Test/Update/UpdateWorkerTestBase.cs
brettfo: They're re-set in every test and the only negative effect I could think of is if a test _doesn't_ use the fake NuGet packages and actually wants to hit `api.nuget.org`, it's possible that these values will cause that test to break, but since the goal of this is to isolate the tests, that kind of test breaking is probably a good thing.

nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core.Test/Update/UpdateWorkerTestBase.cs
brettfo: I made it `public` so I could use it elsewhere (in fact, the `discover` tests use it), I just didn't have a good place to put it and I wanted to get this huge PR off of my box :)

PR #9709, opened 2024-05-13
Generate TODO file for rubocop rspec rules

jurre:
This breaks out all the RSpec related rubocop rules into a new `.rubocop_todo.yml` file, listing each individual file that violates the rule, for most rules.

This allows us to target violations on a per rule / per file basis, and also lists out if the rule can be safely autocorrected for.


thavaahariharangit (approved):
As I needed this files to be merged, to be get started with https://github.com/github/dependabot-updates/issues/5988

PR #9727, opened 2024-05-14
#6017: Address RSpec/ContextWording Rubocop violations.

thavaahariharangit:
Updating context messages when it's not starting with when, without or without


abdulapopoola (approved):

JamieMagee (approved):

bundler/helpers/v1/spec/functions/dependency_source_spec.rb
```diff
@@ -47,7 +47,7 @@
       ])
     end
 
-    context "specified as the default source" do
+    context "with specified as the default source" do
```
jurre, line 50: @thavaahariharangit same thing on a lot of these as well, this should really read

`when specified as the default source`

PR #9730, opened 2024-05-14
Enable RSpec/DescribeClass

robaiken:
https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecdescribeclass

### DIFF
--- a/npm_and_yarn/spec/npm_and_yarn_config_spec.rb
+++ b/npm_and_yarn/spec/npm_and_yarn_config_spec.rb
@@ -3,7 +3,11 @@
 
 require "spec_helper"
 
+# NOTE: This test does not have a corresponding class. It is testing the npm and yarn configuration.
+# rubocop:disable RSpec/DescribeClass
 RSpec.describe "npm and yarn config" do
+  # rubocop:enable RSpec/DescribeClass
+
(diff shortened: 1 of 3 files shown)


npm_and_yarn/spec/npm_and_yarn_config_spec.rb
```diff
@@ -3,7 +3,11 @@
 
 require "spec_helper"
 
+# NOTE: This test does not have a corresponding class. It is testing the npm and yarn configuration.
+# rubocop:disable RSpec/DescribeClass
 RSpec.describe "npm and yarn config" do
+  # rubocop:enable RSpec/DescribeClass
```
jurre, line 9:
```suggestion
RSpec.describe "npm and yarn config" do # rubocop:disable RSpec/DescribeClass
```

Feel entirely free to ignore this, but I always find this a bit more ergonomic to write, I don't think it's well documented that this is possible, but it disables the rule for the block

jurre (approved):

npm_and_yarn/spec/npm_and_yarn_config_spec.rb
robaiken: That is so much better, thanks

PR #9731, opened 2024-05-14
Enable RSpec/EmptyExampleGroup

robaiken:
https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecemptyexamplegroup

### DIFF
--- a/composer/spec/dependabot/composer/file_updater/lockfile_updater_spec.rb
+++ b/composer/spec/dependabot/composer/file_updater/lockfile_updater_spec.rb
@@ -691,31 +691,6 @@ RSpec.describe Dependabot::Composer::FileUpdater::LockfileUpdater do
       end
     end
 
-    context "when there are patches (composer v2)" do
-      let(:project_name) { "patches" }
-
-      let(:dependency) do
-        Dependabot::Dependency.new(
-          name: "ehime/hello-world",
-          version: "1.0.5",
-          requirements: [{
-            file: "composer.json",
-            requirement: "1.0.5",
-            groups: [],
-            source: nil
-          }],
-          previous_version: "1.0.4",
-          previous_requirements: [{
-            file: "composer.json",
-            requirement: "1.0.4",
-            groups: [],
-            source: nil
-          }],
-          package_manager: "composer"
-        )
-      end
-    end
-
     context "regression spec for media-organizer" do
       let(:project_name) { "media_organizer" }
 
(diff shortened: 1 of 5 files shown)


robaiken:
Do we want to keep these tests or are we happy with deleting them?

jurre:
@robaiken could you dig out why they were disabled? As it stands, I'd rather delete them then keep them commented out, but if it's easy to fix them and make them reliable/fast, that seems better?

composer/spec/dependabot/composer/file_updater/lockfile_updater_spec.rb
```diff
@@ -691,31 +691,6 @@
       end
     end
 
-    context "when there are patches (composer v2)" do
```
robaiken, line 694: patch data isn't going to be re-applied to composer.lock

https://github.com/dependabot/dependabot-core/pull/6620

robaiken:
@jurre I couldn't find the context about why the NuGet tests were skipped. Both tests seemed to have issues related to VCR

JamieMagee (approved):

nuget/spec/dependabot/nuget/file_fetcher_spec.rb
```diff
@@ -324,10 +324,11 @@
         )
     end
 
-    # it "fetches the packages.config" do
-    #   expect(file_fetcher_instance.files.map(&:name)).
-    #     to match_array(%w(NuGet.Config packages.config))
-    # end
+    it "fetches the packages.config" do
+      skip "This test was commented out and does not work at the moment"
```
jurre, line 328: Do we expect this will realistically ever be fixed?

nuget/spec/dependabot/nuget/update_checker/version_finder_spec.rb
```diff
@@ -456,10 +456,10 @@
       let(:dependency_name) { "NuGet.Protocol" }
       let(:dependency_version) { "6.3.0" }
 
-      # skipped
-      # it "returns the expected version" do
-      #   expect(subject[:version]).to eq(version_class.new("6.5.0"))
-      # end
+      it "returns the expected version" do
+        skip "This test was commented out and does not work at the moment"
```
jurre, line 460: Same here, is there are plan to address this, or will this realistically just remain skipped?

robaiken:
@jurre I think it's up to the Azure team if they want to keep the tests or not. I have sent them a message, so they are aware of the skipped tests.

PR #9733, opened 2024-05-14
Bump golang from 1.22.2-bookworm to 1.22.3-bookworm in /go_modules

TomSellers:
Similar to PRs #8374 and #8548, this allows Dependabot to function when used against Go projects that require Go 1.22.3.

Go 1.22.3 addresses some security issues. I have verified that the docker image exists.

Reference:
- Go 1.22.3 Milestones: https://github.com/golang/go/issues?q=milestone%3AGo1.22.3+label%3ACherryPickApproved
- Golang docker images: https://hub.docker.com/_/golang


JamieMagee (approved):

rickardgranberg:
Any way this can be expedited? Don't like being without a working dependabot for this long...
Also, I think there's a dupe: #9701

TomSellers:
Note: This will need to be done again next Tuesday when Go 1.22.4 is released.

https://groups.google.com/g/golang-announce/c/ZMLFbgVXBI4

jakecoffman:
@rickardgranberg @TomSellers What error are you seeing? 

In #9435 I put back the auto-download behavior in the Go CLI so it should be working even though Dependabot isn't running the latest Go. We're seeing Dependabot PRs flowing on the Dependabot CLI even though it's on Go 1.22.3: https://github.com/dependabot/cli/pull/326

So if you're seeing an error I'd love to fix the root cause, keeping up with Go releases is not a great long-term solution.

jakecoffman:
https://github.com/dependabot/dependabot-core/pull/9701 is now merged, but again if you're seeing an error because of this please file an issue. I'd love to make it not necessary to keep up with the releases so closely!

rickardgranberg:
I know this is closed, but this is the error I'm seeing:

```
Dependabot failed to update your dependencies because there was an error resolving your Go dependency files.

Dependabot encountered the following error:

go: loading module retractions for <redacted>@v0.1.0: module <redacted-dependency>@v0.34.0 requires go >= 1.22.3 (running go 1.22.2; GOTOOLCHAIN=local+auto)
[Troubleshoot Dependabot errors](https://docs.github.com/github/managing-security-vulnerabilities/troubleshooting-dependabot-errors)
```

The error is seen in a repo where there's a dependency on a module that has `go 1.22.3` in it's go.mod. Running dependabot in a repo that on it's own requires `go 1.22.3` works fine. So it's only in the dependency resolution.

Hope this helps.

PR #9738, opened 2024-05-15
Enabling RSpec/VoidExpect

robaiken:
https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecvoidexpect

### DIFF
--- a/npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker/dependency_files_builder_spec.rb
+++ b/npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker/dependency_files_builder_spec.rb
@@ -75,7 +75,7 @@ RSpec.describe(Dependabot::NpmAndYarn::UpdateChecker::DependencyFilesBuilder) do
         expect(Dir.children(".")).to match_array(
           %w(package.json .npmrc)
         )
-        expect(File.empty?(".npmrc"))
+        expect(File.read(".npmrc")).to be_empty
       end
     end
   end
@@ -90,7 +90,7 @@ RSpec.describe(Dependabot::NpmAndYarn::UpdateChecker::DependencyFilesBuilder) do
         expect(Dir.children(".")).to match_array(
           %w(package.json .npmrc .yarnrc)
         )
-        expect(File.empty?(".npmrc"))
+        expect(File.read(".npmrc")).not_to be_empty
       end
     end
   end
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -339,8 +339,3 @@ RSpec/VerifiedDoubles:
     - 'updater/spec/dependabot/file_fetcher_command_spec.rb'
     - 'updater/spec/dependabot/sentry/sentry_context_processor_spec.rb'
     - 'updater/spec/dependabot/update_files_command_spec.rb'
-
-# Offense count: 2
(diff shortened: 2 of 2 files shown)


npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker/dependency_files_builder_spec.rb
```diff
@@ -87,7 +87,7 @@ def project_dependency_file(file_name)
         expect(Dir.children(".")).to match_array(
           %w(package.json .npmrc .yarnrc)
         )
-        expect(File.empty?(".npmrc"))
+        expect(File.read(".npmrc")).not_to be_empty
```
jurre, line 90: This inverts the expectation doesn't it?

jurre: Very confused why this doesn't break any of the tests 😬

robaiken: Yea, @jurre , it shouldn't work, but I don't want to delay this epic for a broken test. I can either skip the test or leave it as it is, but I will raise an issue.

jurre: I'm just surprised that it was passing before. Is there a subtle difference between `File.empty?` and `File.read.empty?` maybe?

landongrindheim: I assume `.npmrc` has a zero size. `File.read` won't blow up if the file exists (which it does, as shown in the expectation a few lines up), but is empty.

```plaintext
 dependabot@codespaces-a9309d:/app$ touch .npmrc
dependabot@codespaces-a9309d:/app$ irb
irb(main):001> File.read(".npmrc").empty?
=> true
```

landongrindheim: > Is this expected behaviour?

There are enough references in this thread that I'm not 100% sure which behavior you're referring to 🙈 

My hope is that the (perceived) intent of the test will remain. The expectation could be updated to maintain the current behavior with something like 👇 
```suggestion
        expect(File.empty?(".npmrc")).to be_truthy
```

robaiken: @landongrindheim it looks like `File.empty?(".npmrc")` returns `false`. 

> There are enough references in this thread that I'm not 100% sure which behavior you're referring to 🙈
Lol, sorry, you are right. The behavior I am referring to is whether .npmrc should exist in this scenario. It seems like the original test was not set up correctly as it did not have an assertion, so it passes every time. If that's the case, then it's weird that RSpec doesn't log that the test has no assertions.

landongrindheim: @robaiken 👍 In that case, I think it makes sense to make the test honest and ensure that the assertions are inline with the intent of the tests. Good find!

GarryHurleyJr (approved):

PR #9743, opened 2024-05-15
Enabling RSpec/MultipleDescribes

robaiken:
https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecmultipledescribes


JamieMagee (approved):

common/spec/dependabot/dependabot_error_spec.rb
```diff
@@ -0,0 +1,217 @@
+# typed: false
+# frozen_string_literal: true
```
abdulapopoola, line 2: Any reason why the file is renamed?

jurre: I'd guess because of https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Naming/FileName

common/spec/dependabot/dependabot_error_spec.rb
```diff
@@ -0,0 +1,217 @@
...
+
+require "spec_helper"
+require "dependabot/errors"
+
+RSpec.describe Dependabot::DependabotError do
+  describe Dependabot::DependabotError do
```
jurre, line 8: Since this is testing the top-level `Dependabot::DependabotError` which contains the other namespaces, I think we should drop the nesting for these tests, having two describes with the same class is a little uncommon, this should instead just follow the namespace hierarchy 1:1.

common/spec/dependabot/dependabot_error_spec.rb
GarryHurleyJr: thanks Jurre

common/spec/dependabot/dependabot_error_spec.rb
GarryHurleyJr: resolved this issue

jurre:
There seem to be unrelated changes in this PR?

jurre:
> > There seem to be unrelated changes in this PR? just ones made necessary by merges to avoid conflicts
> 
> 
> 
> I merged in main to keep the code fresh.  Since there were conflicts, I fixed those.

If they were fixed, the unnecessary changes would not be showing

GarryHurleyJr:
> > > There seem to be unrelated changes in this PR? just ones made necessary by merges to avoid conflicts
> > 
> > 
> > I merged in main to keep the code fresh.  Since there were conflicts, I fixed those.
> 
> If they were fixed, the unrelated changes would not be showing.
> 
> It might be easiest to do a hard reset to main and re-apply the changes.

Pulled out unrelated code into a separate PR https://github.com/dependabot/dependabot-core/pull/10041 to avoidconfusion.  WIll rebase after that PR is merged.

GarryHurleyJr:
This PR is superseded by https://github.com/dependabot/dependabot-core/pull/10045

PR #9749, opened 2024-05-16
#6017: file_updter_spec context_worrding updates.

thavaahariharangit:


.rubocop_todo.yml
```diff
@@ -60,15 +60,14 @@ RSpec/ContainExactly:
   Exclude:
     - 'npm_and_yarn/spec/dependabot/npm_and_yarn/update_checker_spec.rb'
 
-# Offense count: 30
+# Offense count: 8
 # Configuration parameters: Prefixes, AllowedPatterns.
 # Prefixes: when, with, without
 RSpec/ContextWording:
   Exclude:
     - 'common/spec/dependabot/pull_request_creator/message_builder_spec.rb'
```
abdulapopoola, line 68: can all the files be fixed in one go?

bundler/spec/dependabot/bundler/file_updater_spec.rb
```diff
@@ -118,7 +118,7 @@
             .to include("\"statesman\", \"~> 1.2.0\"")
         end
 
-        context "for a gems.rb setup" do
+        context "when for a gems.rb setup" do
```
jurre, line 121: Let's please try to make these sentences flow nicely, the idea of these context blocks is that, stringed together they explain in English what the test is trying to address, and those sentences should flow well, it makes it easier to understand the tests.

So in this case: `when using a gems.rb setup` for example, that way the entire test will read:

> Dependabot::Bundler::FileUpdater##updated_dependency_files the updated gemfile when using a gems.rb setup, it delegates to the GemfileUpdater

PR #9754, opened 2024-05-17
True type `Dependabot::NpmAndYarn::UpdateChecker::VersionResolver`

JamieMagee:


npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -715,7 +723,7 @@ def requirement_class
         end
 
         def version_regex
-          version_class::VERSION_PATTERN
+          Dependabot::NpmAndYarn::Version::VERSION_PATTERN
```
JamieMagee, line 726: I don't think this is _actually_ dynamic. And Sorbet requires constants to be static.

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/version_resolver.rb
```diff
@@ -650,7 +658,7 @@ def version_install_arg(version:)
           git_source = dependency.requirements.find { |req| req[:source] && req[:source][:type] == "git" }
 
           if git_source
-            "#{dependency.name}@#{git_req[:source][:url]}##{version}"
+            "#{dependency.name}@#{git_source[:source][:url]}##{version}"
```
JamieMagee, line 661: I think this may have been a typo, but I'd like someone to double check my change.

landongrindheim: @JamieMagee Agreed that it was likely a typo. I've grepped around and not found any definition of `git_req` in the NPM ecosystem code 🙂

landongrindheim (approved):

PR #9755, opened 2024-05-17
Pin actions in CI workflow to commit SHA

JamieMagee:
Recommended as part of [GitHub's own GitHub Actions security hardening guide][1] and [the OSSF Scorecards checks][2]

> **Pin actions to a full length commit SHA**
>
> Pinning an action to a full length commit SHA is currently the only way to use an action as an immutable release. Pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository, as they would need to generate a SHA-1 collision for a valid Git object payload. When selecting a SHA, you should verify it is from the action's repository and not a repository fork.

[1]: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions
[2]: https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,11 +42,11 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
         with:
           submodules: recursive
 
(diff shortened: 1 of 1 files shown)


jurre (approved):

jurre:
Thanks Jamie! 👍

PR #9756, opened 2024-05-17
Revert "#6017: Address RSpec/ContextWording Rubocop violations. (#9727)"

jurre:
This reverts commit 0a333d26d334eaf602f4d73269eb9959dfadbdef.

Unfortunately the PR in question only addresses the cop at the surface level, but in many cases does not address the cops intent.

The intent of this rule is for specs, when reading the full context, to be easily readable, so that when rspec gives you the full contextualized name of the spec, it makes sense in plain english and explains what the test is trying to verify.

Many of these examples now have simply prepended `when` or `with`, without making the descriptions grammatically make sense. For example:

```
Dependabot::Bundler::FileFetcher::ChildGemfileFinder#child_gemfile_paths when the file does include a child Gemfile when that includes a variable raises a helpful error
```

This unfortunately leaves us in an arguably worse position, because now rspec can't point us at the examples that need to be updated anymore.

I'm going to revert this change, and we can bring back the changes that make sense, and fix up those that don't.


thavaahariharangit (approved):
Reverting and providing more meaningful context wording.

PR #9760, opened 2024-05-17
Fix #4999 supports vendoring for JavaScript

### LINKED ISSUE #4999: Support NuGet package version overrides
In a forthcoming version of the .NET SDK/Visual Studio/NuGet, support will be added for projects using NuGet [Central Package Management](https://devblogs.microsoft.com/nuget/introducing-central-package-management/) (i.e. with `Directory.Packages.props`) to override the version of a specific package reference using the new [`VersionOverride`](https://devblogs.microsoft.com/nuget/introducing-central-package-management/#overriding-package-versions) property.

As far as I can tell from looking at the code (I'm by no means a Ruby expert so I may have misinterpreted it), dependabot-core would not currently support offering updates for such declarations.

Dependabot already supports package updates for usage of `<PackageVersion>` elements in `Directory.Packages.props`, so I would have thought this should also be supported to complete the support of this feature when it becomes stable in Visual Studio 17.2/.NET SDK 6.0.300/NuGet 6.2.

mohammadalizabihitari:
It is a fix for issue [#4999](https://github.com/github/dependabot-updates/issues/4999) but instead of raising a friendly error provides support for vendoring npm_module


abdulapopoola:
Can the rbi changes be in a separate change?

PR #9766, opened 2024-05-18
Pushing changes in progress for git-lfs support adding to entire ecosystem.

GarryHurleyJr:
Draft pull request.  This is to keep my work in progress.  The branch name will change before this PR is finalized.  I will be adding code to this as time goes on.  This will fix [Issue 6039](https://github.com/github/dependabot-updates/issues/6039), which I created as part of the resolution for [FR issue 5945](https://github.com/github/dependabot-updates/issues/5945)


common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -861,19 +861,35 @@ def decode_binary_string(str)
...
-          <<~CMD
-            git -C #{path} ls-files --stage
-          CMD
-        ).split("\n").filter_map do |line|
+        lfsEnabled = true
+        commandString = getCommandString(path,lfsEnabled)
+        /debugger
+        p commandString/
+        SharedHelpers.run_shell_command(commandString
+         ).split("\n").filter_map do |line|
```
bdragon, line 869: I think this could be simplified a bit without the helper methods and by using `Dir.chdir`

```suggestion
        lfs_enabled = # ...
        git_bin = lfs_enabled ? "git-lfs" : "git"
        Dir.chdir(path) do
          SharedHelpers.run_shell_command("#{git_bin} ls-files --stage")
        end.split("\n").filter_map do |line|
```

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -861,19 +861,35 @@ def decode_binary_string(str)
...
+debugger
           next path if type == DependencyFile::Mode::SUBMODULE
         end
       end
+
+      def getCommandString(path,lfsEnabled)
+	      #the HEREDOC command will see any stray spaces.
+        return "git -C #{path} ls-files --stage" unless lfsEnabled
+        Dependabot.logger.warn("LFS is enabled in this repo.  Please use an LFS enabled client")
+        commandString = "CWD=\"#{__dir__}\";cd #{path};git-lfs ls-files --stage;cd $CWD"
```
bdragon, line 883: I suggested something a little simpler above, but just adding a couple notes:
* [`__dir__`](https://docs.ruby-lang.org/en/3.3/Kernel.html#method-i-__dir__) is the name of the directory that contains this source code file, not the current working directory of the process
* Rather than setting an environment variable for the directory, this is a good use case for [`Dir.chdir`](https://docs.ruby-lang.org/en/3.3/Dir.html#method-c-chdir). This method changes into the given directory, runs whatever is inside block, and then changes back to the original directory.

bdragon (commented):
I added some inline suggestions and here's some general Ruby feedback. I've added links to the docs for our linter, RuboCop, which provide some examples and more detail. Not trying to nitpick, just trying to save you some headache with the linter and code reviews 🙂
* Naming conventions: [method names](https://docs.rubocop.org/rubocop/cops_naming.html#namingmethodname), [method parameter names](https://docs.rubocop.org/rubocop/cops_naming.html#namingmethodparametername), [block parameter names](https://docs.rubocop.org/rubocop/cops_naming.html#namingblockparametername), and [variable names](https://docs.rubocop.org/rubocop/cops_naming.html#namingvariablename) use `snake_case`.
* [Redundant `return`](https://docs.rubocop.org/rubocop/cops_style.html#styleredundantreturn). In Ruby the last expression in a block or method is the return value of that block or method. In these cases the explicit `return` should be omitted. It is common/idiomatic to use guard clauses in methods to reduce if/else nesting, so the `return` is obviously needed for those cases.
* `unless !!value` is unnecessary. Just let the value evaluate as truthy/falsy. Hint: in Ruby `false` and `nil` are falsy and literally all other values are truthy. If you need to explicitly check for `nil`, use `value.nil?`.
* You probably haven't gotten to this point in the implementation yet, but if you end up adding methods to a file that has the `# typed: strict` annotation you're going to want to add Sorbet `sig` blocks to the methods so the type checker passes. The syntax is weird so I would just look for other examples in the file/codebase and follow what they're doing.

GarryHurleyJr:
With some discussion, I have a path towards getting this back on track.
detecting lfs 

- clone shallow
- check for .gitattributes file
- if not found, no lfs
- if found and does not contain "filter=lfs", no lfs
- else lfs

if lfs detected, use git-lfs
if no lfs use git

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -890,10 +890,10 @@ def find_submodules(path)
       sig { params(path: String).returns(T::Boolean) }
       def isLfsEnabled(path)
         filepath = File.join(path,".gitattributes")
-        lfsEnabled = File.exist?(filepath) && File.readable?(filepath) && SharedHelpers.run_shell_command("cat #{filepath} | grep \"filter=lfs\"").include? "#{filepath}"
+        lfsEnabled = T.let(true, T::Boolean) if File.exist?(filepath) && File.readable?(filepath) && SharedHelpers.run_shell_command("cat #{filepath} | grep \"filter=lfs\"").include? "filter=lfs"
```
abdulapopoola, line 893: any reason for copying the same code pattern in 3 different files? Can we centralize it in one location?

common/spec/dependabot/file_fetchers/base_spec.rb
```diff
@@ -304,7 +304,7 @@ def fetch_files
     before do
       allow(file_fetcher_instance).to receive(:commit).and_return("sha")
     end
-    #start of lfs testing
+
```
abdulapopoola, line 307: Can we have unit tests covering these new scenarios?

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -861,18 +858,40 @@ def decode_binary_string(str)
...
+
+      sig { params(path: String).returns(T.nilable(T::Boolean)) }
+      def is_lfs_enabled(path)
+        filepath = File.join(path, ".gitattributes")
+        lfs_enabled = T.let(true, T::Boolean) if File.exist?(filepath) && File.readable?(filepath) &&
+          SharedHelpers.run_shell_command("cat #{filepath} | grep \"filter=lfs\"").include?("filter=lfs")
+      rescue Exception => e
+        # this should not be needed, but I don't trust 'should'
+        lfs_enabled = T.let(false, T::Boolean)
+        raise e.exception("Message: #{e.message}")
```
jurre, line 886: We're rescuing the exception, setting `lfs_enabled` and then immediately re-raising the exception. What's the point of doing all of that? The exception will halt the code from executing, so all of this can just be removed and the end result is the same: an exception is raised.

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -861,18 +858,40 @@ def decode_binary_string(str)
 
       sig { params(path: String).returns(T::Array[String]) }
       def find_submodules(path)
-        SharedHelpers.run_shell_command(
-          <<~CMD
-            git -C #{path} ls-files --stage
-          CMD
-        ).split("\n").filter_map do |line|
+        lfs_enabled = is_lfs_enabled(path) if lfs_enabled.nil?
```
jurre, line 861:
```suggestion
        lfs_enabled = is_lfs_enabled(path) if lfs_enabled.nil?
```

Looks like we're calling `.nil?` on `lfs_enabled` before it is defined, what exactly are we trying to do here?

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -861,18 +858,40 @@ def decode_binary_string(str)
...
+      rescue SharedHelpers::HelperSubprocessFailed => e
+        Dependabot.logger.warn("LFS is enabled in this repo.  Please use an LFS enabled client") if lfs_enabled
+        Dependabot.logger.error(e.message)
+        raise e.exception("Message: #{e.message}")
+      end
+
+      sig { params(path: String).returns(T.nilable(T::Boolean)) }
+      def is_lfs_enabled(path)
+        filepath = File.join(path, ".gitattributes")
+        lfs_enabled = T.let(true, T::Boolean) if File.exist?(filepath) && File.readable?(filepath) &&
```
jurre, line 881:
```suggestion
        T.let(true, T::Boolean) if File.exist?(filepath) && File.readable?(filepath) &&
```

No need to assign anything to this variable, we're just returning the results of the method calls

common/lib/dependabot/file_fetchers/base.rb
```diff
@@ -861,18 +858,40 @@ def decode_binary_string(str)
...
+          SharedHelpers.run_shell_command("cat #{filepath} | grep \"filter=lfs\"").include?("filter=lfs")
+      rescue Exception => e
+        # this should not be needed, but I don't trust 'should'
+        lfs_enabled = T.let(false, T::Boolean)
+        raise e.exception("Message: #{e.message}")
+      end
```
jurre, line 887: I think all of this can be simplified to this:

```suggestion
sig { params(path: String).returns(T::Boolean) }
  def lfs_enabled?(path)
    gitattributes_path = File.join(path, '.gitattributes')

    return false unless File.exist?(gitattributes_path)

    File.readlines(gitattributes_path).each do |line|
      return true if line =~ /filter=lfs/
    end

    false
  end```

File: common/lib/dependabot/shared_helpers.rb [hunk 9]
```diff
@@ -374,12 +374,34 @@ def self.configure_git_to_use_https(host)
 
     sig { params(path: String).void }
     def self.reset_git_repo(path)
-      Dir.chdir(path) do
-        run_shell_command("git reset HEAD --hard")
-        run_shell_command("git clean -fx")
+      if lfs_enabled?(path)
```
jurre, line 377: Why do we need 3 different methods in 3 different places to check if lfs is enabled?

GarryHurleyJr: Because we have three different files in common that each call the git command in a different way and none of them refer to the others.  To make matters worse, some code never even bothers to use any of them.

jurre: Seems like the could all still share a single, shared way to check if lfs is enabled?

common/lib/dependabot/shared_helpers.rb
```diff
@@ -374,12 +374,35 @@ def self.configure_git_to_use_https(host)
...
       end
     end
 
+    sig { params(path: String).returns(T.nilable(T::Boolean)) }
+    def self.lfs_enabled?(path)
+      filepath = File.join(path, ".gitattributes")
+      T.let(true, T::Boolean) if File.exist?(filepath) && File.readable?(filepath) &&
+                                        SharedHelpers.run_shell_command("cat #{filepath} | grep \"filter=lfs\"")
+                                                     .include?("filter=lfs")
+    rescue StandardError => e
```
bdragon, line 400: Jurre's [suggestion](https://github.com/dependabot/dependabot-core/pull/9766#discussion_r1620398131) is more idiomatic and won't incur the cost of shelling out, so I think you should incorporate that.

Just to call attention to it, shelling out to grep like this will return a non-zero exit if there's no match. `SharedHelpers.run_shell_command` will [raise](https://github.com/dependabot/dependabot-core/blob/87d4dfe741e2297d65776a5825dfe40bf31f460d/common/lib/dependabot/shared_helpers.rb#L421-L435) a `HelperSubprocessFailed` exception on a non-zero exit, but here we're swallowing virtually _any_ error that could occur and carrying on. The problem with this is it assumes the only thing that could go wrong is grep not finding a match, when there are other legitimate exceptions that could occur that we would want to propagate.

If you find yourself writing `rescue StandardError`, it might be a good idea to ask if it's really necessary for what you're trying to do. If you truly want to check for the `HelperSubprocessFailed` exception, rescue that class specifically so that other, unexpected exceptions are propagated correctly.

jeffwidman:
Garry has moved on from the :dependabot: team. The underlying issue still exists, but I suspect whoever tackles this will likely start from a fresh PR, so I'm closing this.

Note that there's also this related PR:
* https://github.com/dependabot/dependabot-core/pull/10052

Issue #9745, opened 2024-05-15
Add support for NuGet package scanning of .wixproj files


tippmar-nr:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

The latest version of the Wix installer uses [NuGet packages](https://www.nuget.org/profiles/wixtoolset) to bring in required dependencies (previous versions required a separate local install of the Wix toolset). 

Wix installer projects have a `.wixproj` extension and won't be scanned by the NuGet ecosystem in Dependabox. Please consider adding `.wixproj` as a valid project file extension.

Issue #9711, opened 2024-05-13
Support `.vcxproj` for NuGet dependencies


renefritze:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

We're pulling in native C++ dependencies via NuGets into `.vcxproj`

AFAICT, at present `vcxproj`projects are not considered when discovering which projects' dependencies to consider for an update. 

We currently use `packages.config` files for each solution and import targets files manually.

Is there either a way to "force" dependabot to consider them or can this functionality be added?

Would moving to PackageReference and central version management help?

leeter:
This previously worked as recently as a few months ago and broke somewhat recently. It was explicitly broken in #9294, the author seems to have assumed that nuget only supplied .net dependencies

abhijeetnarvekar:
Can we please get this regression fixed for C++ (.vcxproj) files? We know it worked perfectly fine and loved ❤️  dependabot updating our C++ project nuget dependencies automatically.

zarif98:
How has no one picked up this up even now? This would be a really nice feature to have!

Paper-Folding:
I got a C++/CLI solution mixed with C# `.csproj` and C++ `.vcxproj` projects, and due to `.vcxproj` projects, dependabot update check failed with:  
> **Dependabot couldn't find a Microsoft.Cpp.Default.props**  
> Dependabot requires a Microsoft.Cpp.Default.props to evaluate your .NET dependencies. It had expected to find one at the path: $(VCTargetsPath)/Microsoft.Cpp.Default.props.

PR #9785, opened 2024-05-21
improve NuGet discovery by expanding `*.proj` and `*.sln` files

brettfo:
This also removes a recursive search that wasn't correct; the _only_ entry points allowed from any give directory are the following:

- `*.sln` // n.b., this will be expanded to regular project files
- `*.proj` // n.b., this will be expanded to regular project files
- `*.(cs|vb|fs)proj` // regular project files

This fixes some internal issues where a `.sln` file might be in a different directory than the projects, e.g., dependabot is started in the `/solutions` directory and the following files exist: `/solutions/the-solution.son` and `/src/project.csproj`.  Previously, dependabot wouldn't have been able to navigate from the `/solutions` directory over to `/src` but now as long as the project listing in the `.sln` is correct, then we'll pull in the correct projects.

The same applies to `*.proj` files as well as `*.(cs|vb|fs)proj` files that might have had a `<ProjectReference>` element to another directory.


nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Cli.Test/EntryPointTests.Discover.cs
```diff
@@ -106,14 +105,13 @@ public async Task WithSolution()
...
         {
-            var projectPath = "path/to/my.csproj";
             await RunAsync(path =>
                 [
                     "discover",
                     "--repo-root",
                     path,
                     "--workspace",
-                    path,
+                    "path/to",
```
samtrion, line 114: It would be useful to also paths with `whitespaces`. Something like `path/to project`

brettfo: Good call, added a separate test for specifically this.

samtrion:
@brettfo It would certainly make sense to also consider the open feature requests during this refactoring.

#9711
#9745

brettfo:
@samtrion There's more than just this PR to enable anything that's not .csproj/.vbproj/.fsproj so I'll leave that for the future just to keep this PR focused.

JamieMagee (approved):

PR #9787, opened 2024-05-21
Limit top-level permissions for `labeler` workflow

JamieMagee:
By default the top-level `GITHUB_TOKEN` has all permissions[^1]. This change removes all permissions by default, as recommended by the GitHub Actions security hardening guide[^2].

This change is a low risk test. I plan to modify all other workflows if this works okay.

[^1]: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token
[^2]: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#restricting-permissions-for-tokens


JamieMagee:
@jurre Yes, the `triage` job in the workflow has the following permissions:

```
permissions:
  contents: read
  pull-requests: write
```

In general, the guide asks you to only put common `read` permissions at the top level, and scope `write` permissions to the specific jobs that require them. This workflow only has a single job, so it makes no real difference splitting the `content: read` to the top-level, and the `pull-requests: write` to the job level.

jurre (approved):

PR #9790, opened 2024-05-21
Enable RSpec/DescribedClass

robaiken:
https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecdescribedclass


jurre (approved):

JamieMagee (approved):

abdulapopoola (approved):

hex/spec/dependabot/hex/update_checker_spec.rb
```diff
@@ -437,45 +457,48 @@
...
+      #       "repo" => "dependabot",
+      #       "auth_key" => "d6fc2b6n6h7katic6vuq6k5e2csahcm4",
+      #       "url" => "https://dependabot-private.fly.dev"
+      #     })]
+      #   end
+
+      # describe "Skipping the failing test", pending: "https://github.com/dependabot/dependabot-core/issues/9884" do
+      # it { is_expected.to eq(Dependabot::Hex::Version.new("1.1.0")) } # skipped test
+      # end
+      #      end
```
landongrindheim, line 484: I see several code blocks have been commented out in this commit. If we skip these, we'll at least get _some_ feedback from our test suite. As is, the commented out tests put us at risk of being falsey confident.

I'd love it if, when we skipped a test, the commit message reflected that. When this is ready to be merged, it'd be great if the message was something like "Skip Hex test due to loss of registry access"

landongrindheim (changes requested):
Requesting changes as I see this has been approved, but there have since been comments which will cause tests to not be run. We should address these before merging.

hex/spec/dependabot/hex/update_checker_spec.rb
GarryHurleyJr: I am trying to rework this issue.  These tests are supposed to be skipped, and I am working on skipping them the correct way rather than commenting them out.

GarryHurleyJr: this change has been reverted

hex/spec/dependabot/hex/update_checker_spec.rb
```diff
@@ -9,6 +9,10 @@
 require_common_spec "update_checkers/shared_examples_for_update_checkers"
 
 RSpec.describe Dependabot::Hex::UpdateChecker do
+  before do
+    stub_request(:get, hex_url).to_return(status: 200, body: hex_response)
+  end
+
```
landongrindheim, line 15: 🔎 For next time: these changes seem to be unrelated to the focus of this PR. It'd be great to keep unrelated changes separate 🙂 That said, the functionality remains, so I think we're okay to go.

PR #9792, opened 2024-05-21
handle duplicate directories due to globs

jakecoffman:
If duplicate directories are specified in the job definition, especially when using globs, it's possible that a directory is listed multiple times. This either causes duplicate work, or strange PRs because the DependencySnapshot already thinks those directories have been handled.

By uniqing the list after the globs are applied we ensure we're only processing each directory once.


updater/lib/dependabot/file_fetcher_command.rb
```diff
@@ -120,9 +120,9 @@ def dependency_files_for_globs
 
           has_glob = true
           dir = dir.delete_prefix("/")
-          Dir.glob(dir, File::FNM_DOTMATCH).select { |d| File.directory?(d) }
+          Dir.glob(dir, File::FNM_DOTMATCH).select { |d| File.directory?(d) }.map { |d| "/#{d}" }
```
Nishnha, line 123: why prepend a `/` ? is that just what the updater expects?

jakecoffman: The API sends all directories with a starting `/` (at least it tries to) so we need to match that in order to properly uniq the directories.

Nishnha (approved):
lgtm

PR #9796, opened 2024-05-21
raj/Composer_Dependabot_not_respecting_ignore_minor/patch_updates

raj-meka:
…also note at version_resolver need to understand.


composer/lib/dependabot/composer/update_checker/version_resolver.rb
raj-meka, line 1: The ignore conditions are being passed to Dependabot correctly. However, it seems like the ignore conditions are not being properly applied in the update process. This could be due to the way the VersionResolver class is handling these conditions.  In the VersionResolver class, the latest_resolvable_version method is responsible for fetching the latest resolvable version of a dependency. However, it does not currently consider the ignore conditions when determining the latest version.  To fix this, i have modify the latest_resolvable_version method to check the ignore conditions before returning the version as below.

raj-meka:
Due to technical glitches i am deleting this PR and raised a new PR on behalf of it #9848

Issue #9781, opened 2024-05-20
Dependabot relies on the @unlock[:gems] value in Bundler::Definition, which is now removed at initialize time


honeyankit:
## Context

To solve lazy specification error, we need to update rubygem in Dependabot core with the release having the fix in the below comment. But before we do that, we also need to make few changes into Dependabot core to support the release as `Dependabot relies on the @unlock[:gems] value in Bundler::Definition, which is now removed at initialize time`

More detailed context: https://github.com/rubygems/rubygems/pull/7659#issuecomment-2114690972

This task will unblock the cargo private registry support for GA

sachin-sandhu:
Closing issue as monkey patch was applied [Bundler::materialize_for_installation for resolve Bundler::LazySpecification issue #9807](https://github.com/dependabot/dependabot-core/pull/9807)

PR #9807, opened 2024-05-22
Monkey patch on Bundler::materialize_for_installation for resolve Bundler::LazySpecification issue

sachin-sandhu:
### Draft

**Context** : https://github.com/dependabot/dependabot-core/issues/9781

"To solve lazy specification error, we need to update rubygem in Dependabot core with the release having the fix in the below comment. But before we do that, we also need to make few changes into Dependabot core to support the release as Dependabot relies on the @unlock[:gems] value in Bundler::Definition, which is now removed at initialize time

More detailed context: https://github.com/rubygems/rubygems/pull/7659#issuecomment-2114690972

This task will unblock the cargo private registry support for GA"

**Fix applied:**  Applied monkey patch on Bundler::materialize_for_installation

### DIFF
--- a/bundler/helpers/v2/monkey_patches/bundler_spec_set_patch.rb
+++ b/bundler/helpers/v2/monkey_patches/bundler_spec_set_patch.rb
@@ -1,0 +1,23 @@
+# typed: false
+# frozen_string_literal: true
+
+require "bundler/definition"
+
+# description needs update
+#
+module BundlerSpecSetPatch
+  def materialized_for_all_platforms
+    @specs.map do |s|
+      next s unless s.is_a?(LazySpecification)
+
+      s.source.cached!
+      s.source.remote!
+      spec = s.materialize_for_installation
+      raise GemNotFound, "Could not find #{s.full_name} in any of the sources" unless spec
+
+      spec
+    end
+  end
+end
+
+Bundler::SpecSet.prepend(BundlerSpecSetPatch)
--- a/bundler/helpers/v2/run.rb
+++ b/bundler/helpers/v2/run.rb
@@ -17,6 +17,7 @@ end
 require "definition_ruby_version_patch"
 require "definition_bundler_version_patch"
 require "git_source_patch"
+require "bundler_spec_set_patch"
 
 require "functions"
 
--- a/bundler/helpers/v2/run.rb
+++ b/bundler/helpers/v2/run.rb
@@ -17,7 +17,7 @@ end
 require "definition_ruby_version_patch"
 require "definition_bundler_version_patch"
 require "git_source_patch"
-require "bundler_spec_set_patch"
+require "definition_bundler_spec_set_patch"
 
 require "functions"
 
--- a/bundler/helpers/v2/monkey_patches/definition_bundler_spec_set_patch.rb
+++ b/bundler/helpers/v2/monkey_patches/definition_bundler_spec_set_patch.rb
@@ -3,8 +3,8 @@
 
 require "bundler/spec_set"
 
-# description needs update
-#
+# monkey patch materialized_for_all_platforms for lazy specification issue resolution
+# https://github.com/dependabot/dependabot-core/pull/9807
 module BundlerSpecSetPatch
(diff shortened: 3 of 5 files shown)


bundler/helpers/v2/monkey_patches/bundler_spec_set_patch.rb
```diff
@@ -0,0 +1,23 @@
...
+# frozen_string_literal: true
+
+require "bundler/definition"
+
+# description needs update
+#
+module BundlerSpecSetPatch
+  def materialized_for_all_platforms
+    @specs.map do |s|
+      next s unless s.is_a?(LazySpecification)
```
jurre, line 11:
```suggestion
      next s unless s.is_a?(Bundler::LazySpecification)
```

bundler/helpers/v2/monkey_patches/bundler_spec_set_patch.rb
```diff
@@ -0,0 +1,23 @@
...
+
+      s.source.cached!
+      s.source.remote!
+      spec = s.materialize_for_installation
+      raise GemNotFound, "Could not find #{s.full_name} in any of the sources" unless spec
```
jurre, line 16:
```suggestion
      raise Bundler::GemNotFound, "Could not find #{s.full_name} in any of the sources" unless spec
```

bundler/helpers/v2/monkey_patches/bundler_spec_set_patch.rb
```diff
@@ -0,0 +1,23 @@
+# typed: false
+# frozen_string_literal: true
+
+require "bundler/definition"
```
jurre, line 4:
```suggestion
require "bundler/spec_set"
```

bundler/helpers/v2/run.rb
```diff
@@ -17,6 +17,7 @@
 require "definition_ruby_version_patch"
 require "definition_bundler_version_patch"
 require "git_source_patch"
+require "bundler_spec_set_patch"
```
jurre, line 20: Noticed we're calling the other ones `definition_bundler_spec_set_patch`, let's match that naming

bundler/helpers/v2/monkey_patches/bundler_spec_set_patch.rb
jurre: Maybe we need to require `bundler/lazy_specification` and whatever defines `Bundler::GemNotFound` also

jurre:
If there is any way that we can verify that this fixes the original issue, that would be valuable @sachin-sandhu

sachin-sandhu:
> If there is any way that we can verify that this fixes the original issue, that would be valuable @sachin-sandhu

Ill check with @honeyankit today, good news is, CI, rspec and smoke passes are good.

honeyankit:
@jurre Correct me if I am wrong but this monkey patch should also be applied to bundler v1?

jurre:
@honeyankit I don't think so, it was a regression due to an update in bundler, and v1 hasn't seen an update in years

sachin-sandhu:
@jurre , @honeyankit 

Applied the patch and added test cases.

Tested/ran dependabot CLI (_dependabot update bundler bradfeehan/treehouse_) against the public repo _bradfeehan/treehouse_ having lazy spec issue .

### **Results:** 
**Without patch :** Got lazy spec error : `ERROR undefined method 'default_gem?' for an instance of Bundler::LazySpecification` , full logs : 
[without_patch.txt](https://github.com/dependabot/dependabot-core/files/15440633/without_patch.txt)

**With patch** : No `ERROR undefined method 'default_gem?' for an instance of Bundler::LazySpecification` found while updating . full logs : 
[with_patch.txt](https://github.com/dependabot/dependabot-core/files/15440640/with_patch.txt)

honeyankit (approved):
Thank you for putting this PR

deivid-rodriguez:
Just for your information, I finally released Bundler 2.5.11 yesterday. This monkey patch can go once Dependabot upgrades Bundler.

PR #9815, opened 2024-05-23
TICtxtWrd-github_actions:Test Improvements, Context Wording, ecosystem: github_actions.

thavaahariharangit:
Test Improvements, Context Wording, ecosystem: github_actions.


github_actions/spec/dependabot/github_actions/file_fetcher_spec.rb
```diff
@@ -81,7 +81,7 @@
         )
     end
 
-    context "and an explicit directory given" do
+    context "when an explicit directory given" do
```
jurre, line 84:
```suggestion
    context "when an explicit directory is given" do
```

abdulapopoola (approved):

PR #9824, opened 2024-05-23
fix nuget ignored package version handling

brettfo:
The current handling of `ignored_versions` in a `job.yml` file is incorrect.  Consider an example from the unit tests; a `job.yml` could list that for a certain package, `ignored_versions` should be `< 1.1.1, > 1.1.1` which should be read as "ignore anything below 1.1.1 and ignore anything above 1.1.1" which boils down to "only accept version 1.1.1".  The previous implementation did a slightly different thing of "negate (version is less than 1.1.1 AND version is greater than 1.1.1)" which is impossible; this was a wrong application of our good friend [De Morgan](https://en.wikipedia.org/wiki/De_Morgan%27s_laws).

The proper fix here is to turn that check into "negate (version is less than 1.1.1 OR version is greater than 1.1.1)".  Thankfully the Maven updater does the correct thing so I could almost directly copy what was done there.

Some additional changes had to be made, namely ensuring that the `ignored_versions` property was passed around everywhere and fixing some unit tests that weren't correct.

### DIFF
--- a/nuget/spec/dependabot/nuget/update_checker/version_finder_spec.rb
+++ b/nuget/spec/dependabot/nuget/update_checker/version_finder_spec.rb
@@ -236,7 +236,7 @@ RSpec.describe Dependabot::Nuget::UpdateChecker::VersionFinder do
     end
 
     context "when a version range is specified using Ruby syntax" do
-      let(:ignored_versions) { [">= 2.a, < 3.0.0"] }
+      let(:ignored_versions) { [">= 2.a"] }
       let(:expected_version) { "1.1.2" }
 
       its([:version]) { is_expected.to eq(version_class.new("1.1.2")) }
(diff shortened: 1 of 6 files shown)


nuget/spec/dependabot/nuget/update_checker/version_finder_spec.rb
```diff
@@ -226,7 +226,7 @@
     end
 
     context "when a version range is specified using Ruby syntax" do
-      let(:ignored_versions) { [">= 2.a, < 3.0.0"] }
```
brettfo, line 229: This test was incorrect because it originally said to exclude anything greater than or equal to 2.a and less than 3, which is an empty set, and given that this test was really only checking the Ruby-specific syntax of `2.a` I simplified it.

nuget/spec/dependabot/nuget/update_checker/version_finder_spec.rb
```diff
@@ -671,7 +677,7 @@ def create_nupkg(nuspec_name, nuspec_content)
     its([:version]) { is_expected.to eq(version_class.new("2.0.0")) }
 
     context "when the user is ignoring the lowest version" do
-      let(:ignored_versions) { [">= 2.a, <= 2.0.0"] }
```
brettfo, line 674: This test was originally incorrect for the same reason as the one above so it has been simplified.

sebasgomez238 (approved):

honeyankit (approved):
LGTM

PR #9826, opened 2024-05-23
update to 8.0.300 SDK

brettfo:
The behavior around handling `Directory.Packages.props` has subtly changed in the `*.300` SDK which meant for unit tests, an empty `Directory.Build.props` had to be written and in other cases writing an empty `Directory.Packages.props` wasn't the correct thing to do any more.

More details about the difference with handling `Directory.Packages.props` can be found at NuGet/Home#13453.

### DIFF
--- a/nuget/Dockerfile
+++ b/nuget/Dockerfile
@@ -11,7 +11,7 @@ RUN apt-get update \
   && rm -rf /var/lib/apt/lists/*
 
 # Install .NET SDK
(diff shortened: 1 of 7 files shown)


brettfo:
Codespell CI failure is coming from the `python` module.

brbayes-msft (approved):

JamieMagee (approved):

PR #9800, opened 2024-05-22
Strict type `Dependabot::NpmAndYarn::FileUpdater::PackageJsonUpdater`

JamieMagee:

PR #9827, opened 2024-05-24
True type remainder of `npm_and_yarn`

JamieMagee:
This, alongside #9800, fixes all the remaining type errors in the `npm_and_yarn` ecosystem.

Once both of these are merged I can add

```yml
Sorbet/TrueSigil:
  Enabled: true
```

to `.rubocop.yml`.

### DIFF
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
@@ -1,4 +1,4 @@
-# typed: false
+# typed: true
 # frozen_string_literal: true
 
 require "uri"
@@ -147,7 +147,8 @@ module Dependabot
           retry_count += 1
           raise if retry_count > 2
 
-          sleep(rand(3.0..10.0)) && retry
+          sleep(rand(3.0..10.0))
+          retry
         end
 
         # rubocop:enable Metrics/PerceivedComplexity
--- a/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/registry_finder.rb
+++ b/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/registry_finder.rb
@@ -1,4 +1,4 @@
-# typed: false
+# typed: true
 # frozen_string_literal: true
 
 require "excon"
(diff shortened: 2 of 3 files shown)


npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
```diff
@@ -147,7 +147,8 @@ def run_yarn_updater(path:, yarn_lock:, top_level_dependency_updates:)
           retry_count += 1
           raise if retry_count > 2
 
-          sleep(rand(3.0..10.0)) && retry
+          sleep(rand(3.0..10.0))
+          retry
```
JamieMagee, line 151: This never worked. `sleep` returns an integer, which is coerced into a truthy value, so the right hand side of the `&&` is never evaluated.

npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/registry_finder.rb
```diff
@@ -174,14 +174,14 @@ def npmrc_registries
 
           registries = []
           npmrc_file.content.scan(NPM_AUTH_TOKEN_REGEX) do
-            next if Regexp.last_match[:registry].include?("${")
+            next if Regexp.last_match&.[](:registry)&.include?("${")
```
JamieMagee, line 177: Unfortunately I can't use `fetch` here like I do in `Hash` i.e. `Regexp.last_match&.fetch(:registry)`. `last_match` returns [a `MatchData`][1], not a `Hash`.

[1]: https://docs.ruby-lang.org/en/3.3/MatchData.html

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
landongrindheim: When considering typing, I definitely think this is the right change 💯

Separately: noting that the right-hand side _will_ be evaluated in this situation:

```plaintext
irb(main):001> sleep(1) && 42
=> 42
```

landongrindheim (approved):

npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater/yarn_lockfile_updater.rb
JamieMagee: Interesting... Sorbet led me to believe that the right-hand side was dead code

PR #9839, opened 2024-05-28
TICtxtWrd-nuget: Test Improvement, Context Wording, for the ecosystem nuget

thavaahariharangit:
Test Improvement, Context Wording, for the ecosystem nuget

### DIFF
--- a/nuget/spec/dependabot/nuget/file_parser_spec.rb
+++ b/nuget/spec/dependabot/nuget/file_parser_spec.rb
@@ -217,7 +217,6 @@ RSpec.describe Dependabot::Nuget::FileParser do
       end
 
       context "that is nested" do
-        let(:directory) { "/dir" }
         let(:packages_config) do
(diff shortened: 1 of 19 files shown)


nuget/spec/dependabot/nuget/file_parser_spec.rb
```diff
@@ -216,8 +216,7 @@
         end
       end
 
-      context "that is nested" do
-        let(:directory) { "/dir" }
```
brettfo, line 220: Could we keep this line as well as revert the next 2 changes that look like this:

``` diff
-     file: "packages.config",
+     file: "dir/packages.config",
```

Part of the PR that made those changes was to verify when an update job was started not in the root directory and I'd like to maintain the test coverage.

thavaahariharangit:
Feature covered in https://github.com/dependabot/dependabot-core/pull/9883

PR #9847, opened 2024-05-28
Enforce `true` sigil in `npm_and_yarn` ecosystem

JamieMagee:
With #9800 and #9827, all of the `npm_and_yarn` ecosystem is now `# typed: true` with Sorbet. This change enforces this requirement with RuboCop to prevent regressions in type safety.


sachin-sandhu (approved):
Approving based on comments.

PR #9848, opened 2024-05-29
raj-meka/Composer_ignore_condition_issue_fix_where_Composer_Dependabot_not_respecting_ignore_minor/patch_updates

raj-meka:
…_resolve and updateChecker.

### DIFF
--- a/composer/helpers/v2/src/UpdateChecker.php
+++ b/composer/helpers/v2/src/UpdateChecker.php
@@ -14,7 +14,7 @@ final class UpdateChecker
 {
     public static function getLatestResolvableVersion(array $args): ?string
     {
-        [$workingDirectory, $dependencyName, $gitCredentials, $registryCredentials] = $args;
+        [$workingDirectory, $dependencyName, $gitCredentials, $registryCredentials, $latestAllowableVersion] = $args;
 
         $httpBasicCredentials = [];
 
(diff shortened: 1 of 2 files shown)


raj-meka:
Problem Statement : Dependabot not respecting, ignore minor/patch updates

Run a job at composer with ignore condition as ignore semver-minor and smever-patch the dependabot should ignore the updates but it is creating a PR for it.

So after analysis of issue,  i am able to see that Dependabot is receiving the ignore condition when the job starts. So, it is mishandling them in the updater.

As a fix for the above problem, i am passing the "latest allowable version" to the PHP native helper.

@latest_allowable_version is set when we initialize the version_resolver class.

I have assign "latest allowable version" as parameter in PHP class.

In PHP class instead of just passing the dependency, i have pass the dependency and the latest allowable version.

So instead of passing laravel/farmework i am passing laravel/framework:11.8.0

With the way the client's composer file is setup, upgrading to version 11 cause an error with other dependencies like symfony. This is an expected failure due to the following reasons.

1. Laravel 11 needs a new version of symfony
2. Dependabot is only upgrading Laravel, not symfony
3. The client has their version of symfony pinned to v6 in their composer file, making it harder to upgrade Laravel to 11.

Now, with this fix i am facing issue at the moment seems to be related to private registries and loading a custom package from a VCS. 

Unfortunately, i am unsure how to fix that other than updating the version is expected to be in the required spec files.

@jurre

thavaahariharangit:
covered in https://github.com/dependabot/dependabot-core/pull/9967

PR #9857, opened 2024-05-30
Update bundler from 2.5.9 to 2.5.11

jordan-brough:
See https://github.com/rubygems/rubygems/blob/master/bundler/CHANGELOG.md

FYI, a side effect of this update:
After merging this PR, dependabot will strip credentials from the Gemfile.lock if there are any. E.g. for sidekiq-pro. This is because as of Bundler 2.5.10 gem credentials are never written to Gemfile.lock. See this PR:
https://github.com/rubygems/rubygems/pull/7560

Our app is using Bundler 2.5.10+, so we're currently having to fight with dependabot because when we run `bundle install` the credentials get stripped and then when dependabot runs it puts them back in, since it's using an older version of Bundler.


jordan-brough:
Thanks @deivid-rodriguez. I'll close this PR in favor of yours.

deivid-rodriguez:
Thanks to you!

PR #9860, opened 2024-05-30
Ignore 422 Validation Failed responses when adding assignees to a PR

Nishnha:
Ignore 422 Validation Failed responses when adding assignees to a PR

We are sometimes seeing 422 Validation Failed responses when we add assignees to a PR https://docs.github.com/en/rest/using-the-rest-api/troubleshooting-the-rest-api?apiVersion=2022-11-28#validation-failed

We don't check that the assignees are real users on the repo so that may be the culprit. For now let's ignore these errors so we can continue to create the PR.

### DIFF
--- a/common/lib/dependabot/pull_request_creator/github.rb
+++ b/common/lib/dependabot/pull_request_creator/github.rb
@@ -439,10 +439,12 @@ module Dependabot
           pull_request.number,
           assignees
         )
-      rescue Octokit::UnprocessableEntity, Octokit::NotFound
-        # Octokit::UnprocessableEntity - This can happen if an invalid assignee was passed
-        # Octokit::NotFound            - This can happen if a passed assignee login is now an org account
+      rescue Octokit::NotFound
+        # This can happen if a passed assignee login is now an org account
         nil
+      rescue Octokit::UnprocessableEntity => e
+        # This can happen if an invalid assignee was passed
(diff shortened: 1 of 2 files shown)


common/lib/dependabot/pull_request_creator/github.rb
```diff
@@ -439,8 +439,9 @@ def add_assignees_to_pull_request(pull_request)
           pull_request.number,
           assignees
         )
-      rescue Octokit::NotFound
-        # This can happen if a passed assignee login is now an org account
+      rescue Octokit::UnprocessableEntity, Octokit::NotFound
+        # Octokit::UnprocessableEntity - This can happen if an invalid assignee was passed
```
landongrindheim, line 443: 🔎 What do you think about narrowing in on the error? That could be done by checking the contents of message before bailing, and re-raising if it's not what we intend to swallow.

Nishnha: Unfortunately I haven't been able to reproduce this issue locally, so I don't have a fixture file for it. I came up with 2e60d50076bf5a1d97a1403deabeb7d00ad5d2a3 based on the sentry event 
Screenshot 2024-05-31 at 12 45 24 PM

landongrindheim (approved):

abdulapopoola (approved):

PR #9862, opened 2024-05-30
Upgrade Bundler to 2.5.11 and RubyGems to 3.5.11

deivid-rodriguez:
It needs the tweak I mentioned in https://github.com/rubygems/rubygems/pull/7659#issuecomment-2114690972.

Two of the failing specs I think are just because Bundler got smarter (and Dependabot should get smarter too as a side effect), so I updated expectations.

There's one more failing that I'm not sure about and I'd need to dig in more.

Opening a WIP PR to check smoke tests too.


deivid-rodriguez:
This should be ready now!

Some expectations changed, but I think it's just because Dependabot will be smarter now and able to create more PRs. So I mostly updated the expectations.

jurre:
Thanks @deivid-rodriguez! Can we also get rid of our recent bundler patch with this upgrade?

deivid-rodriguez:
Oh good point, let me remove it!

deivid-rodriguez:
I think I may have pushed too many times and reached some rate limit, apologies! This should be ready now I believe.

abdulapopoola (approved):

abdulapopoola:
Thanks @deivid-rodriguez !

PR #9870, opened 2024-05-31
Strict type `Dependabot::Docker::FileFetcher`

raj-meka:


docker/lib/dependabot/docker/file_fetcher.rb
```diff
@@ -40,63 +42,72 @@ def fetch_files
...
         end
       end
 
       private
 
+      sig { returns(T::Array[DependencyFile]) }
       def dockerfiles
-        @dockerfiles ||=
+        @dockerfiles ||= T.let([], T.nilable(T::Array[T.untyped]))
+        @dockerfiles +=
```
jurre, line 60: Won't this now always append to the array, since the previous line sets it to an empty array if unset, but next time the method is called it will not set it to nil, but still execute this line, appending to the array?

jurre: I believe you can use the following pattern instead:

```ruby
sig { returns(Foo) }
def foo
  @foo ||= T.let(Foo.new, T.nilable(Foo))
end

```

JamieMagee: Yeah, you should be able to wrap the whole `||=` statement in `T.let`, and not have to add the `+=` statement.

docker/lib/dependabot/docker/file_fetcher.rb
```diff
@@ -40,63 +42,72 @@ def fetch_files
...
       def incorrectly_encoded_dockerfiles
-        dockerfiles.reject { |f| f.content.valid_encoding? }
+        dockerfiles.reject { |f| T.must(f.content).valid_encoding? }
       end
 
+      sig { returns(T::Array[DependencyFile]) }
       def yamlfiles
-        @yamlfiles ||=
+        @yamlfiles ||= T.let([], T.nilable(T::Array[T.untyped]))
+        @yamlfiles +=
```
JamieMagee, line 79: Same here as above.

docker/lib/dependabot/docker/file_fetcher.rb
```diff
@@ -40,63 +42,72 @@ def fetch_files
...
           repo_contents(raise_errors: false)
           .select { |f| f.type == "file" && f.name.match?(YAML_REGEXP) }
           .map { |f| fetch_file_from_host(f.name) }
       end
 
+      sig { params(resource: T.untyped).returns(T.nilable(T::Boolean)) }
```
JamieMagee, line 85: Can you be more specific here? Even `BasicObject`?

docker/lib/dependabot/docker/file_fetcher.rb
```diff
@@ -40,63 +42,72 @@ def fetch_files
...
+        @dockerfiles +=
           repo_contents(raise_errors: false)
           .select { |f| f.type == "file" && f.name.match?(DOCKER_REGEXP) }
           .map { |f| fetch_file_from_host(f.name) }
       end
 
+      sig { returns(T::Array[DependencyFile]) }
       def correctly_encoded_dockerfiles
-        dockerfiles.select { |f| f.content.valid_encoding? }
+        dockerfiles.select { |f| T.must(f.content).valid_encoding? }
```
JamieMagee, line 68: Does `f.content&.valid_encoding?` work here?

docker/lib/dependabot/docker/file_fetcher.rb
raj-meka: This is fixed as per review comments

docker/lib/dependabot/docker/file_fetcher.rb
raj-meka: This is fixed as per review comment

docker/lib/dependabot/docker/file_fetcher.rb
raj-meka: This is fixed as per review comment

docker/lib/dependabot/docker/file_fetcher.rb
```diff
@@ -40,63 +42,76 @@ def fetch_files
...
           end
         rescue ::Psych::Exception
           false
         end
       end
 
+      sig { returns(T::Array[Dependabot::DependencyFile]) }
       def incorrectly_encoded_yamlfiles
-        yamlfiles.reject { |f| f.content.valid_encoding? }
+        yamlfiles.reject { |f| T.must(f.content).valid_encoding? }
```
JamieMagee, line 114: Does this work?

```suggestion
        yamlfiles.reject { |f| f.content&.valid_encoding? }
```

docker/lib/dependabot/docker/file_fetcher.rb
JamieMagee: https://www.rubydoc.info/stdlib/core/Object:is_a%3F

```suggestion
      sig { params(resource: Object).returns(T.nilable(T::Boolean)) }
```

JamieMagee (approved):

abdulapopoola (approved):

GarryHurleyJr (approved):

docker/lib/dependabot/docker/file_fetcher.rb
```diff
@@ -40,63 +42,76 @@ def fetch_files
...
         end
       end
 
       private
 
+      sig { returns(T::Array[DependencyFile]) }
       def dockerfiles
-        @dockerfiles ||=
-          repo_contents(raise_errors: false)
+        @dockerfiles ||= T.let(fetch_dockerfiles, T.nilable(T::Array[DependencyFile]))
```
abdulapopoola, line 59: this seems like a change that should not be there?

docker/lib/dependabot/docker/file_fetcher.rb
```diff
@@ -40,63 +42,76 @@ def fetch_files
...
+        dockerfiles.select { |f| f.content&.valid_encoding? }
       end
 
+      sig { returns(T::Array[DependencyFile]) }
       def incorrectly_encoded_dockerfiles
-        dockerfiles.reject { |f| f.content.valid_encoding? }
+        dockerfiles.reject { |f| f.content&.valid_encoding? }
       end
 
+      sig { returns(T::Array[DependencyFile]) }
```
abdulapopoola, line 79: should this be `nilable` too?

jurre: No, since we `T.let(` the expression that we return. It reads a little weird, because the `@yamlfiles` variable _is_ nilable, since if the method is not called, it is nil, but then the value returned from `repo_contents(raise_errors: false).select ....` always returns a value, and so by the time the method returns we know it is not `nil`, even though `@yamlfiles` _is_ `T.nilable(T::Array[DependencyFile])`.

PR #9871, opened 2024-05-31
Strict type `Dependabot::Docker::FileParser`

raj-meka:


abdulapopoola (approved):

docker/lib/dependabot/docker/file_parser.rb
```diff
@@ -75,15 +79,17 @@ def parse
 
       private
 
+      sig { returns(T::Array[Dependabot::DependencyFile]) }
       def dockerfiles
         # The Docker file fetcher fetches Dockerfiles and yaml files. Reject yaml files.
         dependency_files.reject { |f| f.type == "file" && f.name.match?(YAML_REGEXP) }
       end
-
+      sig { params(parsed_from_line: T.untyped).returns(T.untyped) }
```
JamieMagee, line 87: Can you be more specific with the types here? Looks like it should be `params(parsed_from_line: T::Hash[String, String]).returns(String)`

docker/lib/dependabot/docker/file_parser.rb
```diff
@@ -75,15 +79,17 @@ def parse
...
       def version_from(parsed_from_line)
         parsed_from_line.fetch("tag") || parsed_from_line.fetch("digest")
       end
 
+      sig { params(parsed_from_line: T.untyped).returns(T::Hash[T.untyped, T.untyped]) }
```
JamieMagee, line 92: Same as above, be more specific.

JamieMagee (changes requested):
IMO there are too many uses of `T.untyped` and `T.anything` that you should be able to be more specific about.

docker/lib/dependabot/docker/file_parser.rb
raj-meka: Fixed this as per review comments

docker/lib/dependabot/docker/file_parser.rb
raj-meka: Fixed this as per review comments

docker/lib/dependabot/docker/file_parser.rb
```diff
@@ -7,10 +7,13 @@
 require "dependabot/file_parsers"
 require "dependabot/file_parsers/base"
 require "dependabot/errors"
+require "sorbet-runtime"
 
 module Dependabot
   module Docker
     class FileParser < Dependabot::FileParsers::Base
+      extend T::Sig
+      extend T::Helpers
```
JamieMagee, line 16: I'm not sure that this extend is required?

docker/lib/dependabot/docker/file_parser.rb
```diff
@@ -167,11 +182,13 @@ def deep_fetch_images_from_hash(json_object)
         images + json_object.values.flat_map { |obj| deep_fetch_images(obj) }
       end
 
+      sig { returns(T::Array[Dependabot::DependencyFile]) }
       def manifest_files
         # Dependencies include both Dockerfiles and yaml, select yaml.
         dependency_files.select { |f| f.type == "file" && f.name.match?(YAML_REGEXP) }
       end
 
+      sig { params(img_hash: T.untyped).returns(T::Array[T.untyped]) }
```
JamieMagee, line 191: Can we be more specific here?
```suggestion
      sig { params(img_hash: T::Hash[String, T.nilable(String)]).returns(T::Array[String]) }
```

docker/lib/dependabot/docker/file_parser.rb
```diff
@@ -152,6 +166,7 @@ def deep_fetch_images(json_obj)
         end
       end
 
+      sig { params(json_object: T.untyped).returns(T::Array[T.untyped]) }
```
JamieMagee, line 169: Can we be more specific than `T.untyped`?

JamieMagee (approved):

abdulapopoola (approved):

PR #9886, opened 2024-06-03
Prevent github actions with same name ending after hyphen to be updated

### LINKED ISSUE #9885: Wrong action with same name ending gets updated (julia-actions/cache vs actions/cache)
This dependabot PR (https://github.com/abelsiqueira/COPIERTemplate.jl/pull/202/) updated the `actions/cache` to v4, but also the `julia-actions/cache` to v4 (which doesn't exits).

Maybe the following substitution is the problem?

https://github.com/dependabot/dependabot-core/blob/e98046d3aaab433f737848c98d942bb8e206726b/github_actions/lib/dependabot/github_actions/file_updater.rb#L90-L91

abelsiqueira:

Prevent the case where one action has the name of another with a prefix followed by a - (hyphen). For instance, julia-actions/cache and actions/cache.
This is done by updating the regex that matches the action names. `\W` used before does not match `-`, so I changed to the explicit `[^a-zA-Z_-]`.
I have also created a test and a fixture to use in the test.

Fix #9885

The new test should capture the use case presented in #9885. I tested it before and after the fix, so it does appear to be testing what I propose.

### Checklist

### DIFF
--- a/github_actions/lib/dependabot/github_actions/file_updater.rb
+++ b/github_actions/lib/dependabot/github_actions/file_updater.rb
@@ -78,7 +78,7 @@ module Dependabot
             old_declaration
             .gsub(/@.*+/, "@#{new_ref}")
 
-          # Replace the old declaration that's preceded by a non-word character
+          # Replace the old declaration that's preceded by a non-word character (unless it's a hyphen)
           # and followed by a whitespace character (comments) or EOL.
           # If the declaration is followed by a comment that lists the version associated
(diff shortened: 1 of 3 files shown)


abdulapopoola:
Thank you so much for the fix and tests @abelsiqueira

abdulapopoola (approved):

abelsiqueira:
Thanks for the review

abdulapopoola:
Thanks @abelsiqueira for the contribution and help!

PR #9888, opened 2024-06-03
fail sooner when PR creation data is incomplete

jakecoffman:

The PullRequestCreator refuses to create PRs when either the `previous_version` is absent or the `previous_requirements` reveal there were no changes to a dependency.

That code is here: https://github.com/dependabot/dependabot-core/blob/75b48123458652e6572c1ac6ab0934b63412f371/common/lib/dependabot/pull_request_creator.rb#L224-L231

This results in what looks like successful runs ending with a PR not being created. 

To improve this, we need to move the validation check closer to where the issue is. In this PR, I added behind an experiment the same check right before sending the PR data to the client.

This will give us, and customers, more information about the nature of the problem, since it will tell us which dependency is failing and why. Then we can target which ecosystems need help and see if we can fix them.

We should start seeing the errors from the Updater side and track them in Sentry along with the dependency name and ecosystem.

### Checklist

### DIFF
--- a/updater/spec/dependabot/service_spec.rb
+++ b/updater/spec/dependabot/service_spec.rb
@@ -10,6 +10,7 @@ require "dependabot/dependency_snapshot"
 require "dependabot/errors"
 require "dependabot/pull_request_creator"
 require "dependabot/service"
+require "dependabot/experiments"
 
 RSpec.describe Dependabot::Service do
   subject(:service) { described_class.new(client: mock_client) }
@@ -94,8 +95,6 @@ RSpec.describe Dependabot::Service do
             commit_message: "Commit message"
           )
         )
-
-      service.create_pull_request(dependency_change, base_sha)
     end
   end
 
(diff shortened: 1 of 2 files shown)


updater/spec/dependabot/service_spec.rb
```diff
@@ -221,15 +220,70 @@
...
+    context "when the change is missing a requirements change" do
+      let(:dependencies) do
+        [
+          Dependabot::Dependency.new(
+            name: "dependabot-fortran",
+            package_manager: "bundler",
+            version: "1.8.0",
+            previous_version: "1.7.0",
+            requirements: [
+              { file: "Gemfile", requirement: "~> 1.8.0", groups: [], source: nil }
```
Nishnha, line 273: hmm these requirements look normal to me why would it raise an error?

Nishnha: oh because the requirement version didn't change 🙈

Nishnha (approved):

PR #9894, opened 2024-06-04
Init

sachin-sandhu:

Issue : Sentry issue https://github.sentry.io/issues/5028346197/?project=1451818&referrer=github_integration
:DependabotError - dependabot-updater/lib/dependabot/dependency_change_builder.rb in run
FileUpdater failed

Preface : FileUpdater failed is a raised exception (Nuget ecosystem) when file updater is unable to update dependencies.

Example :   
Package [Plugin.Maui.Audio] Does not exist as a dependency in [/home/dependabot/dependabot-updater/repo/examples/MAUI/MAUI.csproj].
2024/06/03 22:58:37 ERROR Error processing Plugin.Maui.Audio (Dependabot::DependabotError)
2024/06/03 22:58:37 ERROR FileUpdater failed

### Checklist

- [ ] I have run the complete test suite to ensure all tests and linters pass.
- [ ] I have thoroughly tested my code changes to ensure they work as expected, including adding additional tests for new functionality.
- [ ] I have written clear and descriptive commit messages.
- [ ] I have provided a detailed description of the changes in the pull request, including the problem it addresses, how it fixes the problem, and any relevant details about the implementation.
- [ ] I have ensured that the code is well-documented and easy to understand.

### DIFF
--- a/updater/lib/dependabot/dependency_change_builder.rb
+++ b/updater/lib/dependabot/dependency_change_builder.rb
@@ -66,6 +66,7 @@ module Dependabot
     sig { returns(Dependabot::DependencyChange) }
     def run
       updated_files = generate_dependency_files
+      Dependabot.logger.error("Generated Dependency list" + updated_files.to_s)
       raise DependabotError, "FileUpdater failed" unless updated_files.any?
 
       # Remove any unchanged dependencies from the updated list


sachin-sandhu:
@dependabot/azure-dev-ops , as mentioned in the issue description, this issue is specific in case directory.build.prop has package references which are commented out but referred in projects. **for now** , the solution is to uncomment the package reference while directory.build.prop file contents are passed to native helper.

nuget/lib/dependabot/nuget/file_fetcher.rb
```diff
@@ -150,6 +150,16 @@ def sln_file_names
       sig { returns(T::Array[Dependabot::DependencyFile]) }
       def directory_build_files
         @directory_build_files ||= T.let(fetch_directory_build_files, T.nilable(T::Array[Dependabot::DependencyFile]))
+
+        prop_file_content = @directory_build_files[0].instance_variable_get(:@content).to_s
+
+        if prop_file_content.include? "<!--<ItemGroup"
```
abdulapopoola, line 156: We should not be parsing commented files, can we find out which part of the code is actually parsing these commented out dependencies?

sachin-sandhu: @abdulapopoola , **Directory.Build.props** file is used as a central management for some settings (including dependencies) in .Net solution. The file is usually placed in root of solution that is applicable to all projects within the solution (csproj, vsproj, Directory.Packages.props etc). So in case a dependency is listed in **Directory.Build.props** file (i.e. PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers" PrivateAssets="all"/>) , all projects who have same dependency listed in project files (csproj, vsproj) will refer to settings (for particular dependency) in **Directory.Build.props** file. 
So in case, the dependency is commented out from **Directory.Build.props** file as this is the case, the nuget should take the setting and omit that particular dependency from update in projects. But in this case it is throwing an exception. In case, the dependency is uncommented(intended fix), the relation between **Directory.Build.props** and project files is established to be good, the update is success and :dependabot: is able to generate a PR. Here are possible solutions to this.
 
**Solution 1:** Throw a more informed exception. In case similar pattern is found and dependabot is unable to update the dependency, a more detailed exception can be thrown.

**Solution 2:** (current) Parse the **Directory.Build.props** file and update file content info before sending it to native helper. The update in this case is to uncomment dependency information so that the update can be made.

**Solution 3:** Remove the particular dependency from projects files (csproj, vsproj) update list before sending it to native helper. This would involve parsing the files and updating content before sending to native helper. However while testing this solution, the projects faced issue with build after no consistent update.

**Answer to question:** The files are passed as a json to native helper . For example, if there are 3 project files in solution and a **Directory.Build.props** file, The content is sent as json to native helper. So only way to modify the content is to parse (already happening) and update the contents as intended (as the fix is trying to do, needs improvement). A more straightaway solution would be to throw a meaningful exception.

_JSON content_

INFO Discovery JSON content: {
  "FilePath": "",
  "IsSuccess": true,
  "Projects": [
    {
      "FilePath": "Directory.Build.props",
      "Dependencies": [],
      "IsSuccess": true,
      "Properties": [],
      "TargetFrameworks": [],
      "ReferencedProjectPaths": []
    },
    {
      "FilePath": "src/Wanted.Bus/Wanted.Bus.csproj",
      "Dependencies": [
        {
          "Name": "AutoMapper",
          "Version": "13.0.1",...

sachin-sandhu:
closing as not needed.

PR #9895, opened 2024-06-04
Strict type `Dependabot::Docker::Requirement::FileUpdater`

raj-meka:


docker/lib/dependabot/docker/file_updater.rb
```diff
@@ -47,33 +52,40 @@ def updated_dependency_files
...
         raise "No Dockerfile!"
       end
 
+      sig { params(file: Dependabot::DependencyFile).returns(String) }
       def updated_dockerfile_content(file)
         old_sources = previous_sources(file)
         new_sources = sources(file)
 
-        updated_content = file.content
+        updated_content = T.let(file.content, T.untyped)
```
abdulapopoola, line 74: can this be typed please?

docker/lib/dependabot/docker/file_updater.rb
```diff
@@ -47,33 +52,40 @@ def updated_dependency_files
...
         end
 
         raise "Expected content to change!" if updated_content == file.content
 
         updated_content
       end
 
+      sig do
+        params(previous_content: String, old_source: T::Hash[Symbol, T.nilable(String)],
+               new_source: T::Hash[Symbol, T.nilable(String)]).returns(String)
```
abdulapopoola, line 87: Can we have a type alias for this? `T::Hash[Symbol, T.nilable(String)]`

Tagging @JamieMagee in case this already exists

docker/lib/dependabot/docker/file_updater.rb
```diff
@@ -109,46 +121,55 @@ def update_digest_and_tag(previous_content, old_source, new_source)
         end
       end
 
+      sig { params(source: T::Hash[Symbol, T.nilable(String)]).returns(T::Boolean) }
```
abdulapopoola, line 124: Wondering if we can have a reuseable alias for this too?

docker/lib/dependabot/docker/file_updater.rb
```diff
@@ -109,46 +121,55 @@ def update_digest_and_tag(previous_content, old_source, new_source)
...
       def specified_with_tag?(source)
-        source[:tag]
+        !!source[:tag]
```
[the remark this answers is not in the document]
jurre: Not really, it's not unheard of, but in typical Ruby code you don't care if something is explicitly true, rather that it's "truthy". In this case for sorbet it's probably the best option if we need this to return a bool

docker/lib/dependabot/docker/file_updater.rb
JamieMagee: We don't currently have an alias for it.

Sometimes we use `T::Hash[Symbol, T.nilable(String)]`  and sometimes we use `T::Hash[Symbol, T.untyped]` depending on the exact key we extract from the `Hash`. If we absolutely know it's going to be a String, we can use the slightly stricter version, but if we're extracting multiple keys, or nested keys in the same method, `T.untyped` is the best approach right now.

In the long term, we should have a class that defines all the possible properties this can have. Similar to what we did for `Credential` in #8967.

docker/lib/dependabot/docker/file_updater.rb
```diff
@@ -109,46 +121,55 @@ def update_digest_and_tag(previous_content, old_source, new_source)
...
         requirements(file)
           .map { |r| r.fetch(:source)[:tag] }
       end
 
+      sig { params(file: Dependabot::DependencyFile).returns(T::Array[String]) }
       def old_tags(file)
-        previous_requirements(file)
-          .map { |r| r.fetch(:source)[:tag] }
+        T.must(previous_requirements(file))
+         .map { |r| r.fetch(:source)[:tag] }
```
JamieMagee, line 143: Does this work?

```suggestion
        previous_requirements(file)
         &.map { |r| r.fetch(:source)[:tag] }
```

docker/lib/dependabot/docker/file_updater.rb
```diff
@@ -109,46 +121,55 @@ def update_digest_and_tag(previous_content, old_source, new_source)
...
 
+      sig { params(file: Dependabot::DependencyFile).returns(T::Array[T::Hash[Symbol, T.nilable(String)]]) }
       def sources(file)
         requirements(file).map { |r| r.fetch(:source) }
       end
 
+      sig { params(file: Dependabot::DependencyFile).returns(T::Array[T::Hash[Symbol, T.nilable(String)]]) }
       def previous_sources(file)
-        previous_requirements(file).map { |r| r.fetch(:source) }
+        T.must(previous_requirements(file)).map { |r| r.fetch(:source) }
```
JamieMagee, line 158: Same as above.

docker/lib/dependabot/docker/file_updater.rb
```diff
@@ -109,46 +121,55 @@ def update_digest_and_tag(previous_content, old_source, new_source)
...
       end
 
+      sig { params(file: Dependabot::DependencyFile).returns(String) }
       def updated_yaml_content(file)
         updated_content = file.content
-        updated_content = update_helm(file, updated_content) if Utils.likely_helm_chart?(file)
-        updated_content = update_image(file, updated_content)
+        updated_content = update_helm(file, T.must(updated_content)) if Utils.likely_helm_chart?(file)
+        updated_content = update_image(file, T.must(updated_content))
```
JamieMagee, line 165: Not sure this is the correct approach here. `file.content` can be `nil`. Can you try and allow the signatures of the called methods to accept that?

docker/lib/dependabot/docker/file_updater.rb
abdulapopoola: Thank you so much for clarifying.

docker/lib/dependabot/docker/file_updater.rb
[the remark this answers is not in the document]
landongrindheim: I'd say it's common, but [not agreed upon](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/DoubleNegation).

My personal preference is to ensure that a method ending in `?` always returns `true` or `false`. To Jurre's point, Ruby itself only cares whether the value is "truthy" (i.e. not `false` or `nil`).

> Is `.nil?` not a more common approach?

`!source[:tag].nil?` might be more common. My sample size is too small to say 🙈 The caveat is that if `source[:tag] == false`, the `!source[:tag].nil?` would return `true`.

JamieMagee (approved):

docker/lib/dependabot/docker/file_updater.rb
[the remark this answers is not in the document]
GarryHurleyJr: yeah, I found that one as well, but was discouraged from using it because in the scenario that I was using it, a nil also evaluated to 'not true' so it worked just as well.
If you want to deal witht eh nil for source, though, I would use something like 
**if source[:tag].nil? raise [some exception] 
rescue [some exception]
/ handle it**

abdulapopoola (approved):

abdulapopoola (approved):

PR #9910, opened 2024-06-05
RSpec enable predicate matcher

GarryHurleyJr:
The Predicate Matcher rubocop rule is disabled for some files.  This Pull request will re-enable it.

Please review for duplicated code blocks which escaped the autocorrect.

This rubocop rule will work as desired

### Checklist

### DIFF
--- a/common/spec/dependabot/dependency_group_spec.rb
+++ b/common/spec/dependabot/dependency_group_spec.rb
@@ -117,17 +117,17 @@ RSpec.describe Dependabot::DependencyGroup do
       context "when no dependencies are assigned to the group" do
         it "returns true if the dependency matches a pattern" do
           expect(dependency_group.dependencies).to eq([])
-          expect(dependency_group.contains?(test_dependency1)).to be_truthy
+          expect(dependency_group.contains?(test_dependency1)).to be(true)
         end
 
         it "returns false if the dependency is specifically excluded" do
           expect(dependency_group.dependencies).to eq([])
-          expect(dependency_group.contains?(test_dependency2)).to be_falsey
+          expect(dependency_group.contains?(test_dependency2)).to be(false)
         end
 
         it "returns false if the dependency does not match any patterns" do
           expect(dependency_group.dependencies).to eq([])
-          expect(dependency_group.contains?(production_dependency)).to be_falsey
+          expect(dependency_group.contains?(production_dependency)).to be(false)
         end
       end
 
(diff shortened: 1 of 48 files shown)


common/spec/dependabot/dependency_group_spec.rb
```diff
@@ -117,17 +117,17 @@
       context "when no dependencies are assigned to the group" do
         it "returns true if the dependency matches a pattern" do
           expect(dependency_group.dependencies).to eq([])
-          expect(dependency_group.contains?(test_dependency1)).to be_truthy
+          expect(dependency_group).to be_contains(test_dependency1)
```
JamieMagee, line 120: `to be_contains` reads a little odd. Does `to include` work?

Same for below.

jurre: Since this is a `Dependabot::DependencyGroup` we'd want to alias `contains?` to `include?` for it to work, but I agree that this reads weird as is so I'd recommend we make that change :+1:

jurre:
Agree with Jamie that `expect(dependency_group).to be_contains(test_dependency1)` reads pretty weird, I've left a suggestion as to how to resolve that, other than that it looks good other than that there seem to be a bunch of weird merge conflict type commits in this branch.

If you could please fix the git history so there's just 1 relevant commit that enables the rule, and one commit that addresses the `be_contains` issue, it'll be good to go 👍

common/spec/dependabot/dependency_group_spec.rb
GarryHurleyJr: to be honest, I was weirded out by the whole syntax being there at all.  I think I understand the concept, that we want the test to pass only if it is true and not be skipped if it is nil.  Actually, I also see that I somehow got the **.contains?** call away from the **dependency_group** which actually defines the method. I changed it to .to be(true) - or .to be(false) in case there is a 'not' present. Committed change.

common/lib/dependabot.rb
```diff
@@ -2,5 +2,5 @@
 # frozen_string_literal: true
 
 module Dependabot
-  VERSION = "0.260.0"
+  VERSION = "0.259.0"
```
jurre, line 5: This downgrades the version, seems like this is on the wrong end of a merge conflict

jurre (changes requested):

common/spec/dependabot/dependency_group_spec.rb
```diff
@@ -117,17 +117,17 @@
       context "when no dependencies are assigned to the group" do
         it "returns true if the dependency matches a pattern" do
           expect(dependency_group.dependencies).to eq([])
-          expect(dependency_group.contains?(test_dependency1)).to be_truthy
+          expect(dependency_group.contains?(test_dependency1)).to be(true)
```
jurre, line 120: This changes the expectation from looking for a truthy value to checking for exactly `true`. This is not what I was suggesting; I suggested aliasing the `contains?` method to `include?` so that we can write this test in the form that Jamie suggested

GarryHurleyJr: well, the predicate matcher rubocop rule actually suggested the changes.  **.contains?** is a method on the dependency_group object.  I guess we could rewrite that method to only respond with true or false, and never nil and get the same effect.  No matter what, the "to be_truthy" and "to be_falsey" will fail this cop.


https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/PredicateMatcher

common/lib/dependabot.rb
GarryHurleyJr: fixed.  Goodd catch.

jurre: It does not seem to be fixed

common/spec/dependabot/dependency_group_spec.rb
jurre: That's not what I'm suggesting. Maybe read what I wrote again carefully

jurre:
@GarryHurleyJr let's leave the tests in the format they're at now, and just focus on getting the PR to succeed CI and remove the downgraded gem version for the project.

GarryHurleyJr:
The code on this is so far out of date that it is riddled with issues.  I am going to start this PR over again.

PR #9914, opened 2024-06-06
True type `Dependabot::Bundler::UpdateChecker::SharedBundlerHelpers`

JamieMagee:

Upgrade `Dependabot::Bundler::UpdateChecker::SharedBundlerHelpers` from `typed: false` to `typed: true`.

### Checklist

### DIFF
--- a/bundler/lib/dependabot/bundler/update_checker/shared_bundler_helpers.rb
+++ b/bundler/lib/dependabot/bundler/update_checker/shared_bundler_helpers.rb
@@ -1,7 +1,8 @@
-# typed: false
+# typed: true
 # frozen_string_literal: true
 
 require "excon"
+require "sorbet-runtime"
 require "uri"
 
 require "dependabot/bundler/update_checker"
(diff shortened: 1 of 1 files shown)


bundler/lib/dependabot/bundler/update_checker/shared_bundler_helpers.rb
```diff
@@ -217,6 +228,9 @@ def jfrog_source
           end
         end
 
+        sig { abstract.returns(String) }
+        def bundler_version; end
```
robaiken, line 232: Is it worth raising an error if this method it call there is no body?

```suggestion
        dsig { abstract.returns(String) }
         def bundler_version
                  raise NotImplementedError, "method not implemented because sorbet demands it exists"
         end
```

robaiken (approved):

bundler/lib/dependabot/bundler/update_checker/shared_bundler_helpers.rb
JamieMagee: There's no need. That approach is only necessary before we started using Sorbet. Once we mark a method as abstract, Sorbet enforces all implementing classes to have a concrete version of the abstract method. [See the documentation][1] for more info:

> Any method marked abstract must have no body. sorbet-runtime will take care to raise an exception if an abstract method is called at runtime.

[1]: https://sorbet.org/docs/abstract#additional-requirements

jurre (approved):

PR #9923, opened 2024-06-06
Give better error when security update not possible due to "lockfile_only" strategy

deivid-rodriguez:

I receive the following error when Dependabot tries to create a security update for an advisory:

Captura de pantalla 2024-06-06 a las 19 58 18

This PR tries to change the above error to instead read (note the "4.0.0.beta6" in the latest possible version instead of "4.0.0.beta7"):

> Dependabot cannot update activeadmin to a non-vulnerable version
>
> The latest possible version of activeadmin that can be installed is **4.0.0.beta6**.
>
> The earliest fixed version is 4.0.0.beta7.

Which makes more sense.

### Checklist


deivid-rodriguez:
This should be ready now!

bundler/spec/dependabot/bundler/update_checker/force_updater_spec.rb
```diff
@@ -16,7 +16,7 @@
     described_class.new(
       dependency: dependency,
       dependency_files: dependency_files,
-      target_version: target_version,
+      target_version: Gem::Version.new(target_version),
```
jurre, line 19: Were these tests previously using the wrong type in the setup and stubs?

deivid-rodriguez: I think in the realworld a `Gem::Version` is always passed, so when I saw tests failing, I normalized the tests to also pass that.

I believe the `force_update` native helper (which this class mainly delegates to) works with a string or a `Gem::Version` indifferently, but my changes have broken that.

So, I could make this work for both `String` and `Gem::Version`, but since in the realworld a `Gem::Version` is passed, I figured it was passed to normalize the test to also pass that and let the class take a consistent type.

deivid-rodriguez: Any opinions @jurre? I think the current version is fine, it will also make `ForceUpdater` easier to type I guess. Let me know if you prefer that I change it, though!

jurre: No, seems like a good change

jurre (approved):

PR #9924, opened 2024-06-06
don't fail when a dependency doesn't have a previous version

jakecoffman:

In https://github.com/dependabot/dependabot-core/pull/9888 I moved the failure out of the service and into the Updater.

In this PR, I prevent the failure by removing the questionable update, and logging to tell us what's happening.

It seems there are various reasons why an update with no previous version will happen, but rather than failing the job or a group, we should log it and continue on.

We should stop seeing the error being reported.

### Checklist

### DIFF
--- a/updater/lib/dependabot/dependency_change.rb
+++ b/updater/lib/dependabot/dependency_change.rb
@@ -139,7 +139,7 @@ module Dependabot
     end
 
     sig { returns(T::Boolean) }
-    def previous_version?
+    def all_have_previous_version?
       return true if updated_dependencies.all?(&:requirements_changed?)
       return true if updated_dependencies.all?(&:previous_version)
 
@@ -148,8 +148,7 @@ module Dependabot
 
     sig { void }
     def check_dependencies_have_previous_version
-      return if updated_dependencies.all?(&:requirements_changed?)
-      return if updated_dependencies.all?(&:previous_version)
+      return if all_have_previous_version?
 
       deps_no_previous_version = updated_dependencies.reject(&:previous_version)
       deps_no_change = updated_dependencies.reject(&:requirements_changed?)
--- a/updater/lib/dependabot/dependency_change.rb
+++ b/updater/lib/dependabot/dependency_change.rb
@@ -145,7 +145,7 @@ module Dependabot
       false
     end
 
-    sig { params(updated_dependencies: T::Array[Dependabot::Dependency]).void }
+    sig { void }
     def check_dependencies_have_previous_version
       return if updated_dependencies.all?(&:requirements_changed?)
       return if updated_dependencies.all?(&:previous_version)
--- a/updater/lib/dependabot/dependency_change.rb
+++ b/updater/lib/dependabot/dependency_change.rb
@@ -16,6 +16,21 @@ module Dependabot
   class DependencyChange
     extend T::Sig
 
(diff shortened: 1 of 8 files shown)


updater/lib/dependabot/dependency_change.rb
```diff
@@ -122,6 +138,27 @@ def merge_changes!(dependency_changes)
...
+    def previous_version?
+      return true if updated_dependencies.all?(&:requirements_changed?)
+      return true if updated_dependencies.all?(&:previous_version)
+
+      false
+    end
+
+    sig { void }
+    def check_dependencies_have_previous_version
+      return if updated_dependencies.all?(&:requirements_changed?)
```
abdulapopoola, line 151: wondering if this is already covered by the `previous_version?` method above?

return if previous_version?

updater/lib/dependabot/updater/operations/create_group_update_pull_request.rb
```diff
@@ -69,7 +69,7 @@ def dependency_change
           if job.source.directories.nil?
             @dependency_change = compile_all_dependency_changes_for(group)
           else
-            dependency_changes = job.source.directories.map do |directory|
+            dependency_changes = job.source.directories.filter_map do |directory|
```
Nishnha, line 72: I think we need to make this same change in refresh_group_update_pull_request

jakecoffman: Yeah, that's true! Seems much less likely to happen, but if it did it would be an exception.

updater/lib/dependabot/service.rb
```diff
@@ -64,9 +49,7 @@ def wait_for_calls_to_finish
 
     sig { params(dependency_change: Dependabot::DependencyChange, base_commit_sha: String).void }
     def create_pull_request(dependency_change, base_commit_sha)
-      if Experiments.enabled?("dependency_change_validation")
-        check_dependencies_have_previous_version(dependency_change.updated_dependencies)
-      end
+      dependency_change.check_dependencies_have_previous_version if Experiments.enabled?("dependency_change_validation")
```
abdulapopoola, line 52: I assume this will still throw the exception and I think the plan was to swallow the error, log it silently and continue?

jakecoffman: Yes, but I am leaving it in for now to verify that what I am changing is effective. If it was the exception will stop happening.

updater/lib/dependabot/updater/operations/create_group_update_pull_request.rb
abdulapopoola: what exception would it be if it happens in that case?

Nishnha (approved):

updater/lib/dependabot/updater/operations/create_group_update_pull_request.rb
jakecoffman: NilClass. I am working on adding some Sorbet sigs to catch this.

jakecoffman: This Sorbet change is getting big so I'll do a follow-up PR with those changes so this PR doesn't get difficult to read.

jakecoffman:
I reran a job that previously errored and it didn't error this time, so this is working.

PR #9929, opened 2024-06-07
True type `updater`

JamieMagee:

Fixes the remaining type errors in `updater` and enforces the `true` sigil with RuboCop

While I've added sigs to most of the methods, I wasn't able to set `strict` due to `attr_reader`s.

Reviewing this commit-by-commit might be easiest.

Sorbet tests pass

### Checklist

### DIFF
--- a/updater/lib/dependabot/updater/group_update_creation.rb
+++ b/updater/lib/dependabot/updater/group_update_creation.rb
@@ -38,6 +38,7 @@ module Dependabot
       # rubocop:disable Metrics/AbcSize
       # rubocop:disable Metrics/MethodLength
       # rubocop:disable Metrics/PerceivedComplexity
+      sig { params(group: Dependabot::DependencyGroup).returns(Dependabot::DependencyChange) }
       def compile_all_dependency_changes_for(group)
         prepare_workspace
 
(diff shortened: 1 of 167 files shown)


Lovebugpaul50 (commented):
ITISWHATITIS

Lovebugpaul50 (commented):
ITISWHATITIS

updater/lib/dependabot/updater/group_update_creation.rb
```diff
@@ -93,9 +113,9 @@ def compile_all_dependency_changes_for(group)
       ensure
         cleanup_workspace
```
JamieMagee, line 114: I think I might be hitting a bug with Sorbet, where this `ensure` block is returning `T::Private::Types::Void::VOID` instead of `dependency_change`

abdulapopoola (approved):

updater/lib/dependabot/updater/security_update_helpers.rb
```diff
@@ -50,16 +63,17 @@ def record_dependency_file_not_supported_error(checker)
...
       end
 
+      sig { params(checker: Dependabot::UpdateCheckers::Base).void }
       def record_security_update_not_possible_error(checker)
         latest_allowed_version =
           (checker.lowest_resolvable_security_fix_version ||
            checker.dependency.version)&.to_s
         lowest_non_vulnerable_version =
-          checker.lowest_security_fix_version&.to_s
+          checker.lowest_security_fix_version.to_s
```
jeffwidman, line 72: This can be `nil` if we found a non-vulnerable version candidate, but then filtered it out as a pre-release etc:
* https://github.com/dependabot/dependabot-core/issues/11948

PR #9934, opened 2024-06-07
Strict type `Dependabot::Swift::MetaDataFinder`

raj-meka:
Sorbet Implementation for Swift MetaDataFinder ruby file.

Please pay attention to the fixes in the Swift MetaDataFinder ruby file.

The sorbet rules will fire as needed.

### Checklist

- [x]  I have run the complete test suite to ensure all tests and linters pass.
- [x]  I have thoroughly tested my code changes to ensure they work as expected, including adding additional tests for new functionality.
- [x]  I have written clear and descriptive commit messages.
- [x]  I have provided a detailed description of the changes in the pull request, including the problem it addresses, how it fixes the problem, and any relevant details about the implementation.
- [x]  I have ensured that the code is well-documented and easy to understand.

### DIFF
--- a/swift/lib/dependabot/swift/metadata_finder.rb
+++ b/swift/lib/dependabot/swift/metadata_finder.rb
@@ -9,6 +9,7 @@ module Dependabot
   module Swift
     class MetadataFinder < Dependabot::MetadataFinders::Base
       extend T::Sig
+
       private
 
       sig { override.returns(T.nilable(T.any(Dependabot::Source, T.untyped))) }
--- a/swift/lib/dependabot/swift/metadata_finder.rb
+++ b/swift/lib/dependabot/swift/metadata_finder.rb
@@ -1,14 +1,17 @@
-# typed: true
+# typed: strict
 # frozen_string_literal: true
 
 require "dependabot/metadata_finders"
 require "dependabot/metadata_finders/base"
+require "sorbet-runtime"
 
 module Dependabot
   module Swift
     class MetadataFinder < Dependabot::MetadataFinders::Base
+      extend T::Sig
       private
 
+      sig { override.returns(T.nilable(T.any(Dependabot::Source, T.untyped))) }
       def look_up_source
         case new_source_type
         when "git" then find_source_from_git_url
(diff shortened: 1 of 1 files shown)


abdulapopoola (approved):

swift/lib/dependabot/swift/metadata_finder.rb
```diff
@@ -1,14 +1,18 @@
...
+require "sorbet-runtime"
 
 module Dependabot
   module Swift
     class MetadataFinder < Dependabot::MetadataFinders::Base
+      extend T::Sig
+
       private
 
+      sig { override.returns(T.nilable(T.any(Dependabot::Source, T.untyped))) }
```
JamieMagee, line 15: `T.any(Dependabot::Source, T.untyped)` is equivalent to `T.untyped`. I don't think you need the union with `T.untyped` here?

```suggestion
      sig { override.returns(T.nilable(Dependabot::Source)) }
```

JamieMagee: @raj-meka I spoke with you offline about this.

The error is due to `find_source_from_registry` returning `void`. I thought this method came from `Dependabot::MetadataFinders::Base`, but it doesn't. It looks like support for `registry` source types was never implemented.

@deivid-rodriguez I was wondering if you remember the intention here?

If not, I'd suggest changing the return type of `find_source_from_registry` to `returns(T.noreturn)` works ([See the documentation here][1]).

[1]: https://sorbet.org/docs/noreturn#declaring-that-a-method-always-raises

raj-meka: I have done change as per review comment

JamieMagee (approved):

swift/lib/dependabot/swift/metadata_finder.rb
deivid-rodriguez: I left it unimplemented, yes. When we first added Swift support, registry source support had been implemented in SwiftPM, but most people seemed to not be using it and relying on git sources instead. So I didn't tackle registry support initially. It could be implemented if registry sources have gained any traction in Swift.

